PR #1655, opened 2019-04-01
Removed logging error messages for OE_INVALID_PARAMETER, OE_BUFFER_TOO_SMALL

soccerGB:
This change is to remove unnecessary logging to avoid confusion when examining logging produced from test runs.


johnkord:
Do you really think this is necessary for 0.5? And is this necessary at all? Those log messages are only hit during testing, right?

soccerGB:
I don't think this need to make to 0.5.0 release

PR #1668, opened 2019-04-03
Remove global cmake flags and add them to CheckDevFlows

oprinmarius:
Currently we have alot of CMake warnings generating hundreds of MB of Jenkins logs
Removed global "--warn-uninitialized and --warn-unused-vars" CMake flags and add them only to the CheckDevFlows stage


oprinmarius:
bors try

PR #1673, opened 2019-04-04
Optimize docker images build

oprinmarius:
Optimize building of the docker images


oprinmarius:
bors try

ionutbalutoiu:
Just had a discussion with @oprinmarius about this PR. The changes from here are still prone to have race conditions. We'll come up with another approach. Closing this.

PR #1676, opened 2019-04-04
Fix language support assumptions in the CMake package

andschwa:
The CMake package checks if the Spectre mitigation flags are supported by the compilers, but to do so it uses a pair of modules that only work in the CMake languages if C and CXX are enabled. This commit makes these checks more robust by not assuming either language is enabled, and only checking if possible. If either check passes (instead of both), the Spectre mitigations will be applied.


andschwa:
This is a really funny bug that cropped up in Paul's sample because it only enables C as a language.

anakrish:
@andschwa I don't understand this issue. OE works only in C/C++. So either one or both must be enabled no?

PR #1682, opened 2019-04-08
[Ansible] [CI/CD] Removed clang-7 pinned package version & added platform specific validation

ionutbalutoiu:
* Removed clang-7 pinned packages versions. The `clang-7` APT packages were recently updated, and the previous packages versions were removed from the upstream LLVM repository.

  This caused us [CI failures](https://oe-jenkins.eastus.cloudapp.azure.com/blue/organizations/jenkins/OpenEnclave-sandbox/detail/OpenEnclave-sandbox/1017/pipeline) because we use Ansible when building Docker images.

  Thus, due to the fact that only a single `clang-7` package version will be available in the upstream LLVM repository, it doesn't make any sense to pin to a specific version.

* Add platform specific validation. Instead of having global variables for the packages versions
used at the validation role, we use platform specific role variables. These will contain the expected versions to be found on each platform.

Signed-off-by: Ionut Balutoiu <ibalutoiu@cloudbasesolutions.com>

### DIFF
--- a/scripts/ansible/inventory/group_vars/linux-agents
+++ b/scripts/ansible/inventory/group_vars/linux-agents
@@ -3,12 +3,6 @@
 
 ansible_ssh_user: azureuser
 
-cmake_target_version: "3.13"
-docker_target_version: "18.09"
-dockerd_target_version: "18.09"
-clang_target_version: "7.0"
-ocaml_target_version: "4.0"
-
 validation_directories:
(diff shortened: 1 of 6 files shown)


ionutbalutoiu:
Merging this into master as it fixes some CI failures when setting up Docker images with Ansible.

@johnkord @CodeMonkeyLeet We might want to have it included in `0.5` as well.

bors r+

PR #1685, opened 2019-04-09
Remove clean/checkout commands from Openenclave.groovy common library

### LINKED ISSUE #1664: Openenclave.groovy shouldn't clean/checkout
Forcing a clean/checkout before the build prevents Jenkinsfiles that include this Openenclave.groovy library from modifying the build (for example, with a "git apply PATCHFILE") before the build begins. This issue is a recommendation to remove the clean/checkout commands from the Openenclave.groovy library and put those into individual Jenkinsfiles instead.

EmilAlexandruStoica:
This PR fixes #1664 by removing the clean/checkout commands from the Openenclave.groovy library.
The clean/checkout commands will instead be used inside each Jenkinsfile.


ionutbalutoiu:
We tested this. It's all good. Merging

bors r+

PR #1613, opened 2019-03-27
Bump master to v0.6.x

andschwa:
This PR serves as the announcement that v0.5.0 has branched, and that all commits to master are now for v0.6.0, unless they are additionally backported.

Accordingly, this bumps the version in master to v0.6.x and updates the changelog.

Also fixes v0.4.1 link in CHANGELOG since the links were never updated.

PR #1687, opened 2019-04-09
Updating CHANGELOG.md and VERSION to reflect the new release number.

EMumau:


CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [Unreleased]
 ------------
 
+[v0.5.0] - 2019-4-9
```
CodeMonkeyLeet: A couple of additional fixups:

* Anand's note for the fix of the CVE should be moved under the **Fixed** section instead of **Added**

* There's also a typo in his text for "... will now write to the supplied argument if it _lies_ in host memory."

* In the **Changed** section, there's an outdated link regarding the QE changes: "To resolve this issue, please install the latest version of the [Intel SGX DCAP packages](https://01.org/intel-software-guard-extensions/downloads) (1.0.1 or newer) on the system that generates the remote report.~~, which as of the time of this change is version 1.0.1 and can be found here: **https://download.01.org/intel-sgx/dcap-1.0.1/dcap_installer/ubuntuServer1604/**~~"

* Mark both the v0.4.0 and v0.4.1 versions as `[DEPRECATED]` since we're not yanking them, but folks should be aware they should not be using them anymore.

* Folks haven't been correctly updating the commit ranges for each update. If you look at the bottom, you'll see that the tagging for the versions in the titles are intended to be links:

    `[Unreleased]: https://github.com/microsoft/openenclave/compare/v0.4.0...HEAD`

So we should have:
```
  [Unreleased]: https://github.com/microsoft/openenclave/compare/v0.5.0...HEAD
  [v0.5.0] https://github.com/microsoft/openenclave/compare/v0.4.1...v0.5.0
  [v0.4.1] https://github.com/microsoft/openenclave/compare/v0.4.0...0.4.1
```

docs/Releasing.md
```diff
@@ -18,7 +18,7 @@ The initial announcement of the upcoming release will be a PR by the release
 manager to the `master` branch with the commit to bump the [VERSION
 file](../VERSION) to _next_ pre-release, e.g. `v0.6.x`, and the commit to update
 the [CHANGELOG](#changelog-updates) "Unreleased" section to `v0.6.0` (and change
-the previous to v0.5.0).
+the previous to `v0.6.0`).
```
CodeMonkeyLeet: Actually this change is incorrect, so you should drop it (the previous version relative to unreleased being v0.6.0 should be v0.5.0). Reading through the doc, it's confusing in what it posits to be the "current" state when it uses each of the version numbers and requires a more extensive rewrite. You should just ignore this for now and we'll fix in master later.

CHANGELOG.md
EMumau: @CodeMonkeyLeet - Will you take another look to see if addressed all of these issues?

docs/Releasing.md
EMumau: @CodeMonkeyLeet @andschwa - This was a change from #1613 which is now out of date.  I just took this change.  @andschwa - Could you please clarify?

andschwa:
Why this and not #1613?

Issue #1659, opened 2019-04-02
Issues with container in Jenkins CI


johnkord:
We've hit a few issues with the container stages in Jenkins CI in the last week or so. We're not sure what's going on, but here is example output:

https://oe-jenkins.eastus.cloudapp.azure.com/job/Bors/job/staging/741/display/redirect

johnkord:
#1624

johnkord:
Here's an example of another weird issue that pops up every so often: https://oe-jenkins.eastus.cloudapp.azure.com/blue/organizations/jenkins/Bors/detail/staging/755/pipeline

```
At global scope:
cc1plus: error: unrecognized command line option '-Wno-implicit-fallthrough' [-Werror]
cc1plus: all warnings being treated as errors
```

ionutbalutoiu:
This issue might be caused by a recently discovered race condition (that got fixed in https://github.com/Microsoft/openenclave/pull/1719).

I'll be monitoring the CI, and if this doesn't reproduce, we can close this issue.

ionutbalutoiu:
I have run ~20 sandbox runs only with simulation stages enabled and none gave any error. Seems like https://github.com/Microsoft/openenclave/pull/1719 was the main reason of this failure.

Closing this until further notice.

PR #1689, opened 2019-04-09
[Docs] Update outdated items from Edger8rGettingStarted.md

ionutbalutoiu:
Found outdated references found by inspection:

* Fix typo `--untrusted-die` (instead of `--untrusted-dir`) for the bash code snippet
* Use `oe_result_t` instead of `oe_return_t` (since this type is not available anymore)
* Remove `public` from the `untrusted` edl definition, otherwise the parser will fail
* Add `[user_check]` to the pointers from the edl definition, otherwise the parser will fail
* Use `OE_OK` instead of the `OE_SUCCESS` (since this is not available anymore)
* Use `oe_create_hello_enclave` instead of `oe_create_enclave` to match the example described in the document

### DIFF
--- a/docs/GettingStartedDocs/Edger8rGettingStarted.md
+++ b/docs/GettingStartedDocs/Edger8rGettingStarted.md
@@ -68,14 +68,14 @@ A simple example of an enclave method and host method are as follows, lets call 
 ```edl
 enclave {
     trusted {
-        public oe_return_t enclave_hello(
-            char *this_is_a_string
+        public oe_result_t enclave_hello(
+            [user_check] char *this_is_a_string
             );
     };
 
     untrusted {
-        public oe_return_t host_hello(
-            char *this_is_a_string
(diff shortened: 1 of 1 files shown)


andschwa:
@johnkord if you ask me, it looks like the behavior of the default compiler on Ubuntu 16.04 changed (like maybe it updated and so is emitting more warnings), hence the CI failures.

johnkord:
@andschwa That's what I thought too, but I've tried to repro this issue on a 16.04 system, and I simply can't when using the distro's latest gcc compiler! There's an open issue that covers this problem: #1659

ionutbalutoiu:
> @andschwa That's what I thought too, but I've tried to repro this issue on a 16.04 system, and I simply can't when using the distro's latest gcc compiler! There's an open issue that covers this problem: #1659

Yes. It looks like an intermittent Docker failure again, we're trying to solve this.

PR #1695, opened 2019-04-11
Update Jenkinsfile to use registry images

### LINKED ISSUE #1672: [CI/CD] Testing Docker images should be built serially at the beginning of the Jenkinsfile
At the moment, the CI builds the testing Docker images at the beginning of the pipeline stages using [this function](https://github.com/Microsoft/openenclave/blob/05ea7ea49682e32e45c22e412c41a6392b874b19/src/jenkins/common/Openenclave.groovy#L15-L18) from the shared library. This workflow applies only to the pipelines stages using Docker containers.

We have multiple pipeline stages using Docker containers, and they are executed in parallel. When the Docker images are already built, containers use the cached image, and everything is fine.

The problem appears when the Docker image is initially built. We have two nodes configured with multiple workers. Two or more workers might attempt to build the same image at the same time, and there's the possibility of a race condition.

It's likely that the transient issues https://github.com/Microsoft/openenclave/issues/1624, https://github.com/Microsoft/openenclave/issues/1659 might come up due to a race condition described above.

oprinmarius:
Use pre-pushed Docker images for the containerized CI testing stages.

This is an attempt to eliminate the race conditions when trying to build the Docker images on the fly with the CI.

Fixes #1672

### DIFF
--- a/.jenkins/Jenkinsfile
+++ b/.jenkins/Jenkinsfile
@@ -46,7 +46,7 @@ def simulationTest(String version, String platform_mode, String build_type) {
                            ninja -v
                            ctest --output-on-failure
                            """
-                oe.oetoolsImage(version, "clang-7", task)
+                oe.ContainerRun("oetools-full-${version}", "clang-7", task)
             }
         }
     }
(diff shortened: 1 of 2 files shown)


oprinmarius:
bors try

oprinmarius:
bors try

oprinmarius:
bors try

PR #1696, opened 2019-04-11
Fix emitted whitespace errors

andschwa:
This fixes the edger8r from emitting trailing whitespace on lines.


andschwa:
I noticed trailing whitespace in generated files. No reason to do that, so this is a trivial fix.

PR #1699, opened 2019-04-15
[CI/CD] Fix nightly packages testing

ionutbalutoiu:
Starting with v0.5 release, the samples directory contains more
samples bundled.

To run every sample bundled with the debian packages, we have to
execute `make build` and `make run` in each sample subdirectory.


vlaza:
bors try

PR #1703, opened 2019-04-16
oe_policy_tool

jiria:
This PR makes some changes to attestation functions and structures based on changes needed for AAS. It also introduces oe_policy_tool for capturing CyReS certificate chains and attestation policy authoring.


.gitmodules
```diff
@@ -17,3 +17,6 @@
 	path = 3rdparty/OpteeCalls
 	url = https://github.com/ms-iot/OpteeCalls.git
 	branch = master
+[submodule "3rdparty/json11"]
+	path = 3rdparty/json11
```
dthaler: Ok for the moment, but in the future, I'd much prefer our projects use https://github.com/akheron/jansson instead.  That's because it's used by the GitHub jose library, and I previously ported the jose library and jansson to work in an enclave, for the OTrP project.

include/openenclave/bits/report.h
```diff
@@ -98,30 +98,24 @@ typedef struct _oe_identity
...
     /** The Product ID for the enclave.
-     * For SGX enclaves, this is the ISVPRODID value. */
+     * For SGX enclaves, this is the ISVPRODID value. 
+     * For CyReS components, this is a binary encoding of a GUID based on the component name. */
     uint8_t product_id[OE_PRODUCT_ID_SIZE];
 
-    /** The unique ID for the device. 
-     *  For CyReS TAs, this is a U-Boot identifier. */
-    uint8_t device_unique_id[OE_UNIQUE_ID_SIZE];
+    /* For CyReS components, this is ECC public key issued for the CyReS component by the caller component. */
```
dthaler: nit: insert "the" before ECC

new_platforms/samples/oe_policy_tool/enc/enc.c
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <openenclave/enclave.h>
+#include <openenclave/bits/stdio.h>
+
+// Include the trusted oe_policy_tool header that is generated
+// during the build. This file is generated by calling the
+// sdk tool sgx_edger8r against the oe_policy_tool.edl file.
```
dthaler: shouldn't it use oe_edger8r instead of sgx_edger8r?

new_platforms/samples/oe_policy_tool/host/host.cpp
```diff
@@ -0,0 +1,477 @@
...
+    }
+#endif
+}
+
+std::vector<oe_report_t> extract_reports(const std::string& source_directory)
+{
+    std::vector<oe_report_t> reports;
+
+    auto action = [&reports, &source_directory](const std::string& filename) {
+        const std::string fullpath = source_directory + "\\" + filename;
```
dthaler: In line 381 you have Linux code, but Linux doesn't use "\\" as the path separator

.gitmodules
jiria: Ok, will keep in mind. Dropping for now from here, as this was only needed by oe_policy_tool, which is moved to a different repo.

include/openenclave/bits/report.h
jiria: Fixed.

new_platforms/samples/oe_policy_tool/enc/enc.c
jiria: Removed from here, fixed in the other repo.

new_platforms/samples/oe_policy_tool/host/host.cpp
jiria: Changed to "/" in the other repo.

include/openenclave/bits/report.h
```diff
@@ -98,30 +98,24 @@ typedef struct _oe_identity
...
-     * For SGX enclaves, this is the ISVPRODID value. */
+     * For SGX enclaves, this is the ISVPRODID value. 
+     * For CyReS components, this is a binary encoding of a GUID based on the component name. */
     uint8_t product_id[OE_PRODUCT_ID_SIZE];
 
-    /** The unique ID for the device. 
-     *  For CyReS TAs, this is a U-Boot identifier. */
-    uint8_t device_unique_id[OE_UNIQUE_ID_SIZE];
+    /* For CyReS components, this is the ECC public key issued for the CyReS component by the caller component. */
+    uint8_t public_key[OE_ECC_PUBLIC_KEY_SIZE];
```
soccerGB: Does this field contain a ECC public key? If it is, it's probably better to name it accordingly.
Or it could be any other type of key?
Another question is:  is it an option to have an union for covering CyRes specific fields?

include/openenclave/bits/report.h
```diff
@@ -98,30 +98,24 @@ typedef struct _oe_identity
...
-    /** The signer ID for the device. 
-     *  For CyReS TAs, this is a hash of the public key that signed the U-Boot image. */
-    uint8_t device_signer_id[OE_SIGNER_ID_SIZE];
-
-    /** The public key for the device.
-     *  For CyReS TAs, this is ECC public key which corresponding private key is held by U-Boot
-     * image. */
-    uint8_t device_public_key[OE_ECC_PUBLIC_KEY_SIZE];
+    /* For CyReS components, this points to an identity of a parent CyReS component that started execution of the current CyReS component. */
+    struct _oe_identity* caller;
```
soccerGB: Is this new field, "caller", used for chaining together the identity information from each cert in the chain?

jiria: That is correct. I thought this was the least intrusive method. Let me know if you have any suggestions. This is also the reason why the free method for parsed report has been added.

include/openenclave/bits/report.h
jiria: Currently ECC only, happy to rename. As for the union, we could, but I think it would not give us much memory wise, as there would no be other fields to balance it, as I was using all the other fields for CyRes as well. The only exception being security_version, but that does not seem worth it on its own. Your call.

soccerGB: For the naming, I was trying to confirm my understanding of this change.  I am ok if you keep it the same.

I don't feel strong about adding union for it. Thanks for explanation.

PR #1497, opened 2019-02-26
Bump LLVM libcxx to 8.0.0

vlaza:
Fixes part **LLVM libcxx 8.0.0** of https://github.com/Microsoft/openenclave/issues/1456

PR #1705, opened 2019-04-16
Update to clang version 8

oprinmarius:
This PR updates the project Clang version from 7 to 8

Fixes https://github.com/Microsoft/openenclave/issues/1688


anakrish:
While we **need** clang-7 or above for spectre mitigation, do we really **need** clang-8?
Yes, we should support clang-8; but should we require it?
Are there any security benefits that clang-8 brings on top of clang-7?

CodeMonkeyLeet:
Should this be timed with #1497 to keep our choice of tooling and LLVM library in sync?

ionutbalutoiu:
> Should this be timed with #1497 to keep our choice of tooling and LLVM library in sync?

I agree!

anakrish:
bors try

ionutbalutoiu:
> bors try

We need to get https://github.com/Microsoft/openenclave/pull/1497 merged first in order to have meaningful Bors runs on this PR.

For the time being, we cherry-pick the changes from https://github.com/Microsoft/openenclave/pull/1497 and execute the CI into a sandbox run.

[This](https://oe-jenkins.eastus.cloudapp.azure.com/blue/organizations/jenkins/OpenEnclave-sandbox/detail/OpenEnclave-sandbox/1179/pipeline) is a recent CI run using `clang-8` with changes from https://github.com/Microsoft/openenclave/pull/1497 and this PR.

anakrish:
> 
> We need to get #1497 merged first in order to have meaningful Bors runs on this PR.
> 
@ionutbalutoiu Why do we need updated libcxx for clang-8?

Some background into why I did a bors try...
I am playing around with using OE an Arch Linux derivative. Since it is a rolling distro, it had relatively latest GCC (8.2.1) and clang. (8.0)

OE built fine and all tests passed with GCC  8.2.1 (two tests needed to be tweaked to disable warnings).

For clang 8.0, there was a single line compile error in type_traits header due to something to do with _Float16 not being supported in the platforms. It commented it out and then everything built fine.
But all tests fail due to SEGV.

ionutbalutoiu:
@anakrish I was referring to the following comment from @CodeMonkeyLeet, that I believe is reasonable:

> Should this be timed with #1497 to keep our choice of tooling and LLVM library in sync?

But we might be able separately merge this if we want.

yakman2020:
The problem is that alpine linux doesn't have clang 8. IF we went ahead with this we would have to build clang 8 from source on alpine containers.  We want to use alpine containers for enclave build as it gives us a musl-natural environment.

achamayou:
@yakman2020 it seems unfortunate to have this constaint coming from Alpine Linux when the only supported distributions are Ubuntu 16.04 and 18.04, and the both benefit from packages provided and maintained by LLVM themselves.

CodeMonkeyLeet:
@ionutbalutoiu @oprinmarius One additional observation: The [GettingStarted.Windows.md](https://github.com/microsoft/openenclave/blob/master/docs/GettingStartedDocs/GettingStarted.Windows.md) instructions right now are currently based off clang 7.0.1 install on Windows. @jxyang recently tried to use the latest 8 version and that introduced build breaks when compiling on Windows. Can we investigate if the linux-windows parity is broken by moving up to v8?

jxyang:
The issue tracking the build error is https://github.com/microsoft/openenclave/issues/1801

ionutbalutoiu:
> @ionutbalutoiu @oprinmarius One additional observation: The [GettingStarted.Windows.md](https://github.com/microsoft/openenclave/blob/master/docs/GettingStartedDocs/GettingStarted.Windows.md) instructions right now are currently based off clang 7.0.1 install on Windows. @jxyang recently tried to use the latest 8 version and that introduced build breaks when compiling on Windows. Can we investigate if the linux-windows parity is broken by moving up to v8?

Yes, we'll take the Linux-Windows parity into consideration before moving on with this PR.

yakman2020:
This should be an issue if we don't actually build in clang 8. We should also create a work item to make clear what specific  toolschains we support and flags to indicate which toolchain to use rather than hard coding.

PR #1709, opened 2019-04-16
Update SGX driver link in SDK getting started docs

CodeMonkeyLeet:


CodeMonkeyLeet:
@johnkord @EMumau We need to cherry pick this to v0.5 branch as well, since the openenclave.io links to the version of the docs there, which is now broken.

CodeMonkeyLeet:
@anitagov They do, but we don't take a packaged install in that legacy configuration in the OE repo. See the prereqs scripts for syncing and rebuilding the latest isgx driver from source.

CodeMonkeyLeet:
bors retry

PR #1711, opened 2019-04-16
Make generated code compatible with strict-aliasing

### LINKED ISSUE #1541: oeedger8r-generated code not compatible with -Werror=strict-aliasing
```
interface_u.c:179:5: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
     *(uint8_t**)&_pargs_out = _output_buffer;
```

This is using gcc 7.3.0 and a recent build of OE (e504559e9ea03dfd).

andschwa:
Resolves #1541.

### DIFF
--- a/include/openenclave/edger8r/common.h
+++ b/include/openenclave/edger8r/common.h
@@ -80,10 +80,10 @@ done:
  * Compute and set the pointer value for the given parameter within the input
  * buffer. Make sure that the buffer has enough space.
  */
-#define OE_SET_IN_POINTER(argname, argsize)                                  \
+#define OE_SET_IN_POINTER(argname, argsize, argtype)                         \
     if (pargs_in->argname)                                                   \
     {                                                                        \
-        *(uint8_t**)&pargs_in->argname = input_buffer + input_buffer_offset; \
+        pargs_in->argname = (argtype)(input_buffer + input_buffer_offset); \
         OE_ADD_SIZE(input_buffer_offset, (size_t)(argsize));                 \
         if (input_buffer_offset > input_buffer_size)                         \
         {                                                                    \
@@ -98,10 +98,10 @@ done:
  * Compute and set the pointer value for the given parameter within the output
  * buffer. Make sure that the buffer has enough space.
  */
-#define OE_SET_OUT_POINTER(argname, argsize)                                   \
+#define OE_SET_OUT_POINTER(argname, argsize, argtype)                          \
     if (pargs_in->argname)                                                     \
     {                                                                          \
-        *(uint8_t**)&pargs_in->argname = output_buffer + output_buffer_offset; \
+        pargs_in->argname = (argtype)(output_buffer + output_buffer_offset); \
         OE_ADD_SIZE(output_buffer_offset, (size_t)(argsize));                  \
         if (output_buffer_offset > output_buffer_size)                         \
         {                                                                      \
@@ -116,11 +116,11 @@ done:
  * Also copy the contents of the corresponding in-out pointer in the input
  * buffer.
  */
-#define OE_COPY_AND_SET_IN_OUT_POINTER(argname, argsize)                       \
+#define OE_COPY_AND_SET_IN_OUT_POINTER(argname, argsize, argtype)                 \
     if (pargs_in->argname)                                                     \
     {                                                                          \
-        uint8_t* _p_in = (uint8_t*)pargs_in->argname;                          \
-        *(uint8_t**)&pargs_in->argname = output_buffer + output_buffer_offset; \
+        argtype _p_in = (argtype)pargs_in->argname;                          \
+        pargs_in->argname = (argtype)(output_buffer + output_buffer_offset); \
         OE_ADD_SIZE(output_buffer_offset, (size_t)argsize);                    \
         if (output_buffer_offset > output_buffer_size)                         \
         {                                                                      \
@@ -133,12 +133,12 @@ done:
 /**
  * Copy an input parameter to input buffer.
  */
-#define OE_WRITE_IN_PARAM(argname, size)                                   \
+#define OE_WRITE_IN_PARAM(argname, argsize, argtype)                       \
     if (argname)                                                           \
     {                                                                      \
-        *(uint8_t**)&_args.argname = _input_buffer + _input_buffer_offset; \
-        OE_ADD_SIZE(_input_buffer_offset, (size_t)(size));                 \
-        memcpy((void*)_args.argname, argname, (size_t)(size));             \
+        _args.argname = (argtype)(_input_buffer + _input_buffer_offset); \
+        OE_ADD_SIZE(_input_buffer_offset, (size_t)(argsize));                 \
+        memcpy((void*)_args.argname, argname, (size_t)(argsize));             \
     }
 
 #define OE_WRITE_IN_OUT_PARAM OE_WRITE_IN_PARAM
@@ -146,14 +146,14 @@ done:
 /**
  * Read an output parameter from output buffer.
  */
-#define OE_READ_OUT_PARAM(argname, size)                    \
+#define OE_READ_OUT_PARAM(argname, argsize)                 \
     if (argname)                                            \
     {                                                       \
         memcpy(                                             \
             (void*)argname,                                 \
             _output_buffer + _output_buffer_offset,         \
-            (size_t)(size));                                \
-        OE_ADD_SIZE(_output_buffer_offset, (size_t)(size)); \
+            (size_t)(argsize));                             \
+        OE_ADD_SIZE(_output_buffer_offset, (size_t)(argsize)); \
     }
 
 #define OE_READ_IN_OUT_PARAM OE_READ_OUT_PARAM
--- a/include/openenclave/edger8r/common.h
+++ b/include/openenclave/edger8r/common.h
@@ -80,16 +80,16 @@ done:
  * Compute and set the pointer value for the given parameter within the input
  * buffer. Make sure that the buffer has enough space.
  */
-#define OE_SET_IN_POINTER(argname, argsize, argtype)                         \
-    if (pargs_in->argname)                                                   \
-    {                                                                        \
+#define OE_SET_IN_POINTER(argname, argsize, argtype)                       \
+    if (pargs_in->argname)                                                 \
+    {                                                                      \
         pargs_in->argname = (argtype)(input_buffer + input_buffer_offset); \
-        OE_ADD_SIZE(input_buffer_offset, (size_t)(argsize));                 \
-        if (input_buffer_offset > input_buffer_size)                         \
-        {                                                                    \
-            _result = OE_BUFFER_TOO_SMALL;                                   \
-            goto done;                                                       \
-        }                                                                    \
+        OE_ADD_SIZE(input_buffer_offset, (size_t)(argsize));               \
+        if (input_buffer_offset > input_buffer_size)                       \
+        {                                                                  \
+            _result = OE_BUFFER_TOO_SMALL;                                 \
+            goto done;                                                     \
+        }                                                                  \
     }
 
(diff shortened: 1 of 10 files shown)


andschwa:
bors try

include/openenclave/edger8r/common.h
```diff
@@ -80,10 +80,10 @@ OE_INLINE oe_result_t oe_add_size(size_t* total, size_t size)
  * Compute and set the pointer value for the given parameter within the input
  * buffer. Make sure that the buffer has enough space.
  */
-#define OE_SET_IN_POINTER(argname, argsize)                                  \
+#define OE_SET_IN_POINTER(argname, argsize, type)                            \
```
anakrish: Passing in the type is a great idea.

include/openenclave/edger8r/common.h
```diff
@@ -80,10 +80,10 @@ OE_INLINE oe_result_t oe_add_size(size_t* total, size_t size)
...
     if (pargs_in->argname)                                                   \
     {                                                                        \
-        *(uint8_t**)&pargs_in->argname = input_buffer + input_buffer_offset; \
+        *(type*)&(pargs_in->argname) = (type)(input_buffer + input_buffer_offset); \
```
anakrish: One of our goals has been that the generated code is C++ compatible since one day we would like to generate C++ files.
Can you make sure that this code is C++ compatible (I don't see any issues with it).
You can test C++ compatibility directly #including the generates C files in a C++ file

tests/oeedger8r/edl/aliasing.edl
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
anakrish: Do you want to have separate edl files or just a single edl file for locking down bug fixes?

andschwa: This one should be separate because we have to build with specific flags to repro the bug.

anakrish: Ah ok.

andschwa:
bors try

tests/oeedger8r/host/teststring.cpp
```diff
@@ -71,8 +71,8 @@ oe_result_t ecall_string_no_null_terminator_modified(
     *(uint8_t**)&_pargs_in = _input_buffer;
     OE_ADD_SIZE(_input_buffer_offset, sizeof(*_pargs_in));
 
-    OE_WRITE_IN_PARAM(s1, _args.s1_len * sizeof(char));
-    OE_WRITE_IN_OUT_PARAM(s2, _args.s2_len * sizeof(char));
+    OE_WRITE_IN_PARAM(s1, _args.s1_len * sizeof(char), char*);
```
andschwa: @anakrish These are silly; why are we using these macros directly still?

tools/oeedger8r/Emitter.ml
```diff
@@ -327,6 +326,27 @@ let get_cast_to_mem_expr (ptype, decl)=
...
+      int
+      double*
+      char*
+    ]}. NOTE: This does not yet handle multi-dimensional arrays correctly. *)
+let get_parameter_type_str (pt: Ast.parameter_type) (declr: Ast.declarator) =
+  match pt with
+  | Ast.PTPtr (atype, ptr_attr) ->
+    let tystr = Ast.get_tystr atype in
+    (* TODO: We probably need to handle multi-dimensional arrays here. *)
+    tystr ^ if Ast.is_array declr || ptr_attr.pa_isary then "*" else ""
```
andschwa: This is what broke everything else after I'd initially fixed my single repro yesterday 😆 Had to handle both normal arrays and "foreign" arrays.

tools/oeedger8r/Emitter.ml
```diff
@@ -327,6 +326,27 @@ let get_cast_to_mem_expr (ptype, decl)=
...
+  | Ast.PTVal (atype) -> Ast.get_tystr atype
```
andschwa: Probably not a case our current uses will hit...

tools/oeedger8r/Emitter.ml
```diff
@@ -379,9 +399,10 @@ let oe_prepare_input_buffer (os:out_channel) (fd:Ast.func_decl) (alloc_func:stri
       | Ast.PTPtr (atype, ptr_attr) ->
         if ptr_attr.Ast.pa_chkptr then
           let size = oe_get_param_size (ptype, decl, "_args.") in
+          let tystr = get_parameter_type_str ptype decl in
```
andschwa: Bah all this code is called so many times and it's sooo similar.

include/openenclave/edger8r/common.h
andschwa: Thanks 😄

include/openenclave/edger8r/common.h
andschwa:
```
#include "aliasing_t.c"

add_enclave(TARGET edl_aliasing_enc CXX SOURCES aliasing.cpp ${aliasing_t}pp)

[303/754] : && /usr/bin/c++  -g   tests/oeedger8r/enc/CMakeFiles/edl_aliasing_enc.dir/aliasing.cpp.o tests/oeedger8r/enc/CMakeFiles/edl_aliasing_enc.dir/aliasing_t.cpp.o  -o tests/oeedger8r/enc/edl_aliasing_enc ... suceeded
```

Looks like it worked fine.

andschwa:
@anakrish What was the other bug this was also fixing, from the IOT team?

tools/oeedger8r/Emitter.ml
anakrish: Given that now the type is being used in the LHS, how do const arguments work?

tools/oeedger8r/Emitter.ml
anakrish: Feel free to refactor :)

tools/oeedger8r/Emitter.ml
anakrish: Do we assert in our tests that the foreign arrays values are correctly marshalled?

andschwa: To the last question, it sure looks like it to me:

andschwa: To the first question, this function will never emit a `const` qualifier as `get_tystr` does not do that, and we don't explicitly check `pa_rdonly` to add it.

We could, if we needed to. Checking if that's a possibility...

include/openenclave/edger8r/common.h
```diff
@@ -80,16 +80,16 @@ OE_INLINE oe_result_t oe_add_size(size_t* total, size_t size)
...
-        OE_ADD_SIZE(input_buffer_offset, (size_t)(argsize));                 \
-        if (input_buffer_offset > input_buffer_size)                         \
-        {                                                                    \
-            _result = OE_BUFFER_TOO_SMALL;                                   \
-            goto done;                                                       \
-        }                                                                    \
+#define OE_SET_IN_POINTER(argname, argsize, argtype)                       \
+    if (pargs_in->argname)                                                 \
+    {                                                                      \
+        pargs_in->argname = (argtype)(input_buffer + input_buffer_offset); \
```
andschwa: @anakrish Upon reflection, I simplified this quite a bit 😆

include/openenclave/edger8r/common.h
```diff
@@ -116,44 +116,44 @@ OE_INLINE oe_result_t oe_add_size(size_t* total, size_t size)
...
-        if (output_buffer_offset > output_buffer_size)                         \
-        {                                                                      \
-            _result = OE_BUFFER_TOO_SMALL;                                     \
-            goto done;                                                         \
-        }                                                                      \
-        memcpy(pargs_in->argname, _p_in, (size_t)(argsize));                   \
+#define OE_COPY_AND_SET_IN_OUT_POINTER(argname, argsize, argtype)            \
+    if (pargs_in->argname)                                                   \
+    {                                                                        \
+        argtype _p_in = (argtype)pargs_in->argname;                          \
```
anakrish: Are we reaching here for foreign arrays.
Given:
typedef int my_array_t[5];

void foo([in, isary] my_array_t input);

We don't know what the root type is; and the foreign array is embedded within the mashalling struct.
So we shouln't reach here.

andschwa:
@anakrish I combined the test I added with the other tests after enabling -fno-strict-aliasing for them, which prompted several more fixes. About to add tests to verify the behavior of foreign types.

tools/oeedger8r/Emitter.ml
andschwa: I had to dig into this quite a bit, should be good to go now.

include/openenclave/edger8r/common.h
andschwa: We don't reach here exactly, but there were several places where we had to cast to `void *` when marshalling foreign arrays that we weren't yet doing correctly.

andschwa:
bors try

andschwa:
Windows issues...

bors try

tests/oeedger8r/enc/CMakeLists.txt
```diff
@@ -47,3 +48,7 @@ if (CMAKE_CXX_COMPILER_ID MATCHES GNU OR CMAKE_CXX_COMPILER_ID MATCHES Clang OR
 endif()
 
 target_include_directories(edl_enc PRIVATE ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/..)
+if (NOT WIN32)
```
anakrish: For enclaves, is there some issue with enabling this on Windows?

tools/oeedger8r/Emitter.ml
```diff
@@ -315,18 +314,35 @@ let oe_gen_args_header (ec: enclave_content) (dir:string)=
     {[
       int a[10][20]
     ]}
-    needs to be cast to [int *] . *)
+    needs to be cast to [int *].
+
+    NOTE: Foreign arrays are marshalled as [void *], but foreign pointers
```
anakrish: Got it. Thanks for the clarification.

include/openenclave/edger8r/common.h
anakrish: @HernanGatta I remember that you ran into an issue with GCC code generation for ARM and had to fix it in your copy of these macros. With this PR we are going to pass in the type to the macros so that the cast is correctly typed.
Hopefully this should fix the issue that you ran into as well.

tests/oeedger8r/enc/CMakeLists.txt
andschwa: Yeah it doesn't like the compiler flag -Werror; do you know the right way to do it on Windows? I didn't see any gain spending time on it since we have the check on Linux.

andschwa:
@anakrish The log for the Windows compilation failure is [here](https://oe-jenkins.eastus.cloudapp.azure.com/blue/organizations/jenkins/Bors/detail/trying/839/pipeline). Feel free to take a look.

tests/oeedger8r/enc/CMakeLists.txt
anakrish: Based on this: https://oe-jenkins.eastus.cloudapp.azure.com/blue/organizations/jenkins/Bors/detail/trying/839/pipeline/
the error is happening on the host.

We should not be passing in the flag for the host. But we should be passing in the flag for enclaves.

andschwa: I mean there's no reason not to enable strict-aliasing for of our code.

andschwa: As this code will disappear shortly with the resolution of #1717 (which this PR is blocking), I am going to punt this issue. We will have strict-aliasing enabled for all of our code in a Windows and Linux compatible manner.

PR #1714, opened 2019-04-17
Use correct offset for debug and simulate flags.

### LINKED ISSUE #1708: oegdb incorrectly prints that an enclave has been loaded in simulation mode
oegdb prints that an enclave has been loaded in simulation mode even for enclaves loaded in hardware mode. This is just a diagnostic message and should not affect debugging.
We need to make sure that the field offsets in enclave_t structure is consistent with what the debugger expects.

CC: @radhikaj

anakrish:
Fixes #1708 

oe_enclave_t offsets here:
https://github.com/Microsoft/openenclave/blob/113a8bf22768b3152d57b31085e1cb981556ed99/host/sgx/enclave.h#L136-L139

### DIFF
--- a/tests/debugger/oegdb/CMakeLists.txt
+++ b/tests/debugger/oegdb/CMakeLists.txt
@@ -9,6 +9,7 @@ add_test(
     COMMAND
         ${OE_BINDIR}/oegdb --batch
         --command=${CMAKE_CURRENT_SOURCE_DIR}/commands.gdb
+        --return-child-result # This fails the test in case of any error.
         -arg host/oe_gdb_test_host enc/oe_gdb_test_enc
 )
 
@@ -17,5 +18,6 @@ add_test(
     COMMAND
         ${OE_BINDIR}/oegdb --batch
         --command=${CMAKE_CURRENT_SOURCE_DIR}/commands.gdb
+        --return-child-result # Tnis fails the test in case of any error.
         -arg host/oe_gdb_test_host enc/oe_gdb_test_enc --simulation-mode
 )
--- a/tests/debugger/oegdb/CMakeLists.txt
+++ b/tests/debugger/oegdb/CMakeLists.txt
@@ -18,6 +18,6 @@ add_test(
     COMMAND
         ${OE_BINDIR}/oegdb --batch
         --command=${CMAKE_CURRENT_SOURCE_DIR}/commands.gdb
-        --return-child-result # Tnis fails the test in case of any error.
+        --return-child-result # This fails the test in case of any error.
         -arg host/oe_gdb_test_host enc/oe_gdb_test_enc --simulation-mode
 )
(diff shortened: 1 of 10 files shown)


anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

tests/debugger/oegdb/CMakeLists.txt
```diff
@@ -17,5 +18,6 @@ add_test(
     COMMAND
         ${OE_BINDIR}/oegdb --batch
         --command=${CMAKE_CURRENT_SOURCE_DIR}/commands.gdb
+        --return-child-result # Tnis fails the test in case of any error.
```
CodeMonkeyLeet: Typo: "# _This_ fails the test …"

tests/debugger/oegdb/oe_gdb_test.edl
```diff
@@ -4,6 +4,7 @@
 enclave {
     trusted {
         public int enc_add(int a, int b);
+	public void enc_assert_debugger_binary_contract();
```
CodeMonkeyLeet: Style: Convert tabs to spaces.

anakrish:
bors try

tests/debugger/oegdb/enc/contract.c
```diff
@@ -0,0 +1,46 @@
...
+#include <openenclave/internal/sgxtypes.h>
+#include <openenclave/internal/tests.h>
+#include <stdio.h>
+#include <string.h>
+#include "../../../../enclave/core/sgx/td.h"
+
+// These values are filled by the debugger.
+// The variables are marked volatile so that the compiler does not hard-code
+// the initialization value of the variables in comparision.
+// Since 0 could be a legal value for some of these items, they are
```
CodeMonkeyLeet: This could probably use some additional elaboration as to why they need to be illegal values. (i.e. expect the test evaluation of the variables to fail if commands.gdb could not overwrite the variables in the debugger)

tests/debugger/oegdb/CMakeLists.txt
anakrish: Fixed.

tests/debugger/oegdb/oe_gdb_test.edl
anakrish: Fixed.

tests/debugger/oegdb/enc/contract.c
anakrish: Fixed.

anakrish:
@paulcallen 
Formalizing/rewrite of the debugger contract is planned for v0.6 and would take most of the release.
The proposed contract incorporates your suggestions where applicable.

The fix in this PR is more appropriate for keeping the master not broken.
I have added tests to rigorously lock down the contract. This test is not throwaway and will be applicable for v0.6 as well when the contract is rewritten.

Can you take another pass at reviewing this?

PR #1719, opened 2019-04-22
[CI/CD] Fix CI shared lib race condition

ionutbalutoiu:
Variables declared in Jenkins groovy pipeline logic without any
type or the keyword `def` are considered global by default.
Every global variable is shared among parallel stages with all
the nodes.

The `image` variable from the CI shared library was meant to be
a local variable, but due to the fact that we lacked `def` or
any type when declaring the variable, this was considered global
and caused a lot of race conditions, and random transient failures.


ionutbalutoiu:
This might potentially solve the issue https://github.com/Microsoft/openenclave/issues/1659

PR #1720, opened 2019-04-22
Fix error message in the data sealing sample

rgerganov:


andschwa:
Thanks for the PR @rgerganov! Getting this triaged for review.

PR #1724, opened 2019-04-23
[CMAKE] Validate CMAKE_BUILD_TYPE at top level

aserdean:
CMAKE_BUILD_TYPE should be either empty, DEBUG, RELEASE, RELWITHDEBINFO or
MINSIZEREL.

Add a validation step to warn the user if he is using a nonstandard build
type.

Signed-off-by: Alin Gabriel Serdean <aserdean@ovn.org>
Fixes: https://github.com/Microsoft/openenclave/issues/1722

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,13 @@ endif ()
 # ignored.
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+# Validate CMAKE_BUILD_TYPE
+string(TOUPPER "${CMAKE_BUILD_TYPE}" uppercase_CMAKE_BUILD_TYPE)
+list(APPEND uppercase_build_type "" "DEBUG" "RELEASE" "RELWITHDEBINFO")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -59,6 +59,13 @@ endif ()
 # ignored.
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+# Validate CMAKE_BUILD_TYPE
+string(TOUPPER "${CMAKE_BUILD_TYPE}" uppercase_CMAKE_BUILD_TYPE)
+list(APPEND uppercase_build_type "" "DEBUG" "RELEASE" "RELWITHDEBINFO" "MINSIZEREL")
```
andschwa: I don't see us actually testing `MINSIZEREL` so I'd suggest we remove that and just assert agains the other three.

aserdean: I took the basic types from CMAKE :). Indeed we don't use `MINSIZEREL` so I removed it.

PR #1727, opened 2019-04-24
Added checks to CMake find_library for crypto and dl

EmilAlexandruStoica:
This pull request fixes https://github.com/Microsoft/openenclave/issues/1563 by adding checks to CMake find_library for crypto and dl


cmake/sdk_cmake_targets_readme.md
```diff
@@ -32,11 +32,8 @@ The remaining targets in `openenclave-targets.cmake` are automatically included
 when needed by the above.
 
 The package will also search for (and requires) the
-[`Threads` package](https://cmake.org/cmake/help/latest/module/FindThreads.html).
-
-However, the libraries `crypto` and `dl` will also need to be available on the
-system, but due to an outstanding issue, the failure will happen at link-time if
-they are unavailable.
+[`Threads` package](https://cmake.org/cmake/help/latest/module/FindThreads.html),
+`CryptoPP` package and `dlib` package.
```
andschwa: Can we link to both of these modules too?

host/CMakeLists.txt
```diff
@@ -109,18 +109,8 @@ add_library(oehost STATIC
...
-  add_library(crypto SHARED IMPORTED)
-  set_target_properties(crypto PROPERTIES IMPORTED_LOCATION ${CRYPTO_LIB})
-
-  find_library(DL_LIB NAMES dl)
-  add_library(dl SHARED IMPORTED)
-  set_target_properties(dl PROPERTIES IMPORTED_LOCATION ${DL_LIB})
-endif ()
-
+find_package(CryptoPP REQUIRED)
+find_package(dlib REQUIRED)
```
andschwa: The line:

```
  target_link_libraries(oehost PRIVATE crypto dl Threads::Threads)
```

Must be adjusted to use the imported targets.

EmilAlexandruStoica: After the last commit changes, the line looks like this:

`target_link_libraries(oehost PRIVATE CryptoPP:CryptoPP dlib::dlib Threads::Threads)`

cmake/sdk_cmake_targets_readme.md
EmilAlexandruStoica: Links added. Please take a look at them and tell me if they are ok (I couldn't find any documentation for these two libraries on [cmake.org](https://cmake.org/) page).

ionutbalutoiu:
@andschwa when you get the chance, please take a look again the latest changes from PR.

andschwa:
Looks like these aren't included by default?

```
  By not providing "FindCryptoPP.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "CryptoPP",
  but CMake did not find one.

  Could not find a package configuration file provided by "CryptoPP" with any
  of the following names:

    CryptoPPConfig.cmake
    cryptopp-config.cmake
```

If they're not [default modules](https://cmake.org/cmake/help/latest/manual/cmake-modules.7.html#find-modules), then we're using a third-party's module and will need to commit it, which means redistributing it (which means checking the license is appropriate and adding a third-party notice).

andschwa:
I am seeing 2,713 new files and 896,276 new lines...adding the upstream modules should only be a few files each (the module, its license, and a readme)...

EmilAlexandruStoica:
@andschwa I added the licences and also the libraries as submodules. Let me know if this what you asked for.

ionutbalutoiu:
bors try

andschwa:
No, I’m sorry, it wasn’t. We should commit only the minimum code required to use find_package(dlib)/cryptopp. This should not require committing entire projects. Each library should have a corresponding CMake config file, which if it’s not installed by the package on the system or part of the CMake distribution itself, we could add it here to use it, or write our own.

host/CMakeLists.txt
```diff
@@ -112,19 +112,19 @@ target_link_libraries(oehost PUBLIC oe_includes)
 # TODO: Replace these with `find_package` and add as dependencies to
 # the CMake package.
 if (UNIX)
-  find_library(CRYPTO_LIB NAMES crypto)
-  add_library(crypto SHARED IMPORTED)
-  set_target_properties(crypto PROPERTIES IMPORTED_LOCATION ${CRYPTO_LIB})
+  find_library(CRYPTO_LIB NAMES cryptopp)
```
andschwa: I'm sorry, I'm lost, why did we change the library from crypto to cryptopp?

If there isn't a ready made CMake module for finding `crypto` that's fine, the only thing we're really trying to get out of this is a configuration time check for the `crypto` library (right now it's link-time, because we don't use `find_library` with `REQUIRED` nor are we using a dedicated CMake module for it). Whatever we use we need to duplicate for our installed CMake package too, which is why a simple module which finds the crypto library we're expecting would be idea, but this PR actually looks like it's changing the library to this different cryptopp implementation instead?

debugger/ptraceLib/CMakeLists.txt
```diff
@@ -1,9 +1,9 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-find_library(DL_LIB NAMES dl)
-add_library(dl SHARED IMPORTED)
-set_target_properties(dl PROPERTIES IMPORTED_LOCATION ${DL_LIB})
+find_library(DL_LIB NAMES dlib)
```
andschwa: Since the `dl` library should probably always exist (can we verify this?) we can probably just check if this results in Not Found and error out, and make sure we do the same in our CMake package, we should definitely not be swapping to another implementation.

andschwa:
I've gotten really confused as to the approach taken here. With respect to the original issue https://github.com/microsoft/openenclave/issues/1563, we're just trying to solidify our logic for finding our dependencies.

For instance, we don't check if the calls to `find_library()` for `crypto` and `dl` fail, which means if they happen to fail, then the error will be at link time, and this results in a poor user experience. This is true for both building from source and from against our package.

I believe that the `dl` library is unlikely to ever be missing, but we should double check, and the same for `crypto`.

Often CMake ships a default module for finding a library and bringing them in as imported targets, which is nice because this would give us error checking for free, and sometimes if they don't then there's a readily available third party package to do so, but this should just be a single file with a little bit of find_library and check for errors logic. I think this is what got confused here, as this PR started adding build code from `cryptopp` and `dlib` (notably not `crypto` nor `dl`), implying they were third-party modules for finding `crypto` and `dl`, but looking at the actual packages added, I'm not even sure what they are.

Can we restart this and just check for errors from `find_library` (after verifying that CMake doesn't have a module to find either of these libraries as CMake packages)?

EmilAlexandruStoica:
@andschwa

There is no CMake module for crypto or dl (see [here](https://cmake.org/cmake/help/v3.14/manual/cmake-modules.7.html)).

The error checking was done and this is the result:
```
-- Spectre 1 mitigations supported
-- Valgrind not found
-- Looking for crypto library - found
-- Looking for dl library - found
-- Looking for pthread.h
```
Please see the updated PR changes. Is this what you are looking for ?

host/CMakeLists.txt
```diff
@@ -145,12 +145,22 @@ add_dependencies(oehost posix_untrusted_edl)
 # the CMake package.
 if (UNIX)
   find_library(CRYPTO_LIB NAMES crypto)
-  add_library(crypto SHARED IMPORTED)
-  set_target_properties(crypto PROPERTIES IMPORTED_LOCATION ${CRYPTO_LIB})
-
+  if (NOT CRYPTO_LIB)
```
andschwa: This is great. Can you copy this into the CMake package logic too? https://github.com/microsoft/openenclave/blob/1922dfbfc37c2d78361e662de45cd313af12f486/cmake/openenclave-config.cmake.in#L16

EmilAlexandruStoica: I can add the crypto and dl as dependecies too. From what I know the find_dependency macro also checks for errors, so there is no need to also add conditions here. ([find dependecy macro](https://cmake.org/pipermail/cmake-developers/2016-February/027876.html))

andschwa: `find_dependency` only wraps `find_package`, which you said there weren't any for dl or crypto, so I think we'd have to use `find_library` here like you did outside of the package...

EmilAlexandruStoica: I understand. I did the changes to the ```openenclave-config.cmake.in```.

ionutbalutoiu:
bors try

ionutbalutoiu:
bors try

PR #1733, opened 2019-04-26
Remove `:PATH` from CMake variables

andschwa:
It is unnecessary, and while not incorrect, it is confusing.


andschwa:
@anitagov I grepped for and replaced all the rest of the instances.

bors try

andschwa:
@anitagov Whoops but oh well. I guess that's where this came from to begin with 😆

PR #1734, opened 2019-04-26
Refactored enclave crypto to oecrypto and added ability to rename mbedtls symbols

gupta-ak:


gupta-ak:
@andschwa We ended up changing what we wanted for the oecrypto refactor, so I ended up not needing the mbedtls CMake build refactor that I did 😅 .

anitagov:
@gupta-ak - Please ensure that you do a git mv of existing files so that the previous history is available easily with GitHub history, etc.

andschwa:
@anitagov Looks to me like he did.

3rdparty/mbedtls/rename-mbedtls-symbols
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Explanation:
```
andschwa: Really appreciate the explanation 👍

samples/CMakeLists.txt
```diff
@@ -15,5 +15,7 @@ install(DIRECTORY helloworld file-encryptor data-sealing
 install(FILES README.md
   DESTINATION ${CMAKE_INSTALL_DATADIR}/openenclave/samples)
 
-add_test(NAME samples
-  COMMAND ${CMAKE_COMMAND} -DUSE_LIBSGX=${USE_LIBSGX} -DSOURCE_DIR=${CMAKE_CURRENT_SOURCE_DIR} -DBUILD_DIR=${PROJECT_BINARY_DIR} -DPREFIX_DIR=${CMAKE_INSTALL_PREFIX} -P ${CMAKE_CURRENT_SOURCE_DIR}/test-samples.cmake)
+if (NOT OE_PREFIX_MBEDTLS_SYMBOLS)
```
andschwa: Why does this break? Can we fix it?

enclave/crypto/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+    key.c
+    random.c
+    rsa.c
+    sha.c
+    ${PLATFORM_SRC})
+
+maybe_build_using_clangw(oecrypto)
+
+target_include_directories(oecrypto PRIVATE
+    ${PROJECT_SOURCE_DIR}/include/openenclave/corelibc)
```
andschwa: If this is the case, should this actually link against oelibc (or even just oelibc_includes)?

(I'd appreciate an update to the documented dependency graph to go along with this PR so we know what we're intending to change here.)

enclave/crypto/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+
+target_link_libraries(oecrypto PUBLIC
+    oe_includes
+    mbedcrypto)
+
+if (OE_PREFIX_MBEDTLS_SYMBOLS)
```
andschwa: This logic is a little dangerous and is probably going to result in a "caching" issue. At some point someone is going to change the value of this option and not do a clean build and things won't work. It would be a major pain probably to support that scenario, so I'm not suggesting we should, just maybe leave a comment the optoin that if it's changed a clean build needs to happen. Otherwise we need to setup up some dependencies really carefully.

samples/CMakeLists.txt
gupta-ak: Some of the samples use mbedtls directly, so those won't work. If we have each sample as a different test, then we can fix it. I'd like this in general, because it would allow testing the installation of the `OE_PREFIX_MBEDTLS_SYMBOLS` path without manually running the samples.

enclave/crypto/CMakeLists.txt
gupta-ak: oelibc isn't corelibc. corelibc is the minimal set of standard C functions used by oecore. oelibc is the musl C standard library that is built from oecore/oecrypto/etc.

andschwa: Gotcha, that makes sense. But let's still update the graph since we're adding and changing dependencies.

samples/CMakeLists.txt
andschwa: Can they be updated to use oecrypto instead of mbedtls? I don't recall them using mbedtls directly (at least from a build standpoint), they just picked up whatever was necessary from `oeenclave`/`oelibc` and `oehost`.

The samples test is an integration test of the package; they're not each a different test because the context switches from our build to the installed build when that script is called. They're actually built and ran from an installed copy of the SDK (`build/install/opt/openenclave/...`) (which was installed on the fly by that script). I don't think it's viable to make each sample build and run as a separate test in our build because then each sample would have to install the SDK.

It would require a lot of hacking up of the script. Is there another way?

enclave/crypto/CMakeLists.txt
gupta-ak: Yup, ill update the graph.

andschwa: Thanks 🙏

samples/CMakeLists.txt
gupta-ak: oecrypto is a private API and not available for the installation. The samples use mbedtls directly, since our default OE SDK installation packages mbedtls libraries + headers, so they can call it. (e.g. https://github.com/Microsoft/openenclave/blob/master/samples/data-sealing/common/keys.cpp). 

The idea for `OE_PREFIX_MBEDTLS_SYMBOLS` is that you don't want mbedtls packaged with OE SDK, because your application is using it's own copy of mbedtls which would conflict with the OE one. Since the OESDK still needs crypto, we rename the mbedtls symbols to avoid the symbol issues.

Can we export the  `OE_PREFIX_MBEDTLS_SYMBOLS` for the samples? or maybe some hack that checks if mbedtls is available and only run some of the samples if it is?

enclave/crypto/CMakeLists.txt
anakrish: > oelibc isn't corelibc. corelibc is the minimal set of standard C functions used by oecore. oelibc is the musl C standard library that is built from oecore/oecrypto/etc. To use a kernel comparision, oelibc is the user-space libc, while corelibc is the libc used internally by the kernel.

@mikbras @CodeMonkeyLeet @paulcallen 
What Akash describes about corelibc in this issue was more or less my understanding as well: corelibc is the minimal set of **standard C functions** used by oecore. As per our discussion in #1732, this description is not exactly accurate. I would suggest that we consider a more appropriate name for corelibc since it is acting like a platform abstraction layer.

samples/CMakeLists.txt
andschwa: > Can we export the OE_PREFIX_MBEDTLS_SYMBOLS for the samples?

Can you expand on this?

>  or maybe some hack that checks if mbedtls is available and only run some of the samples if it is?

That's certainly doable, and that might even be something we want to make not a hack at all and instead expose as a a user visible value in the CMake package so that if you're using a build of the SDK without mbedtls, it's very obvious (and easy to test for).

enclave/crypto/CMakeLists.txt
gupta-ak: Would it work if the library file name changes?
like this:
```
if (OE_PREFIX_MBEDTLS_SYMBOLS)
   set_target_properties(oecrypto PROPERTIES OUTPUT_NAME "oecrypto2")
   # objcopy, etc
endif()

install(TARGETS oecrypto EXPORT openenclave-targets
  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}/openenclave/enclave
  RENAME liboecrypto.a)
```

samples/CMakeLists.txt
gupta-ak: > Can you expand on this?

Like in `samples/CMakeLists.txt`: the `add_test` line can have:

```${CMAKE_COMMAND} -DOE_PREFIX_MBEDTLS_SYMBOLS =${OE_PREFIX_MBEDTLS_SYMBOLS } ...```.

Then, use the variable in `test-samples.cmake`.

> That's certainly doable, and that might even be something we want to make not a hack at all and instead expose as a a user visible value in the CMake package so that if you're using a build of the SDK without mbedtls, it's very obvious (and easy to test for).

That sounds reasonable. Is it possible to add a CMake variable to the `openenclave-targets` export set?

**Comment by radhikaj** [comment]
Maybe the title can be modified to reflect the refactoring work as well?

File: enclave/crypto/CMakeLists.txt (hunk 4 above)
Comment (radhikaj) [reply enclave/crypto/CMakeLists.txt]: Should the target be linking against oe_includes?

File: samples/CMakeLists.txt (hunk 2 above)
Comment (andschwa) [reply samples/CMakeLists.txt]: That'd the the hacky way to do it just for the samples. We don't need do that.

If this is something that users might need to check when they're using the SDK (which it seems like it is), let's define it on package creation in [openenclave-config.cmake.in](https://github.com/Microsoft/openenclave/blob/master/cmake/openenclave-config.cmake.in) instead and use it directly.

> That sounds reasonable. Is it possible to add a CMake variable to the openenclave-targets export set?

 I mean kinda, just add it to that file like the other variables. They'll be filled in on configuration when the file is [configured](https://github.com/Microsoft/openenclave/blob/a6eea72472a798798d1001fc80787b1b92d811c1/cmake/package_settings.cmake#L31).

Then the samples (which use our SDK via `find_package`) will have any and all variables from that file set.

File: enclave/crypto/CMakeLists.txt (hunk 4 above)
Comment (andschwa) [reply enclave/crypto/CMakeLists.txt]: @gupta-ak I don't think so, the main problem is conditionally adding the post-build step. If that script gets executed, then you turn OFF `OE_PREFIX_MBEDTLS_SYMBOLS`, you wouldn't _expect_ the script to have been executed, but I _don't think_ CMake is going to detect the removal of the post-build step and rebuild oecrypto without running the script. I think it'll just not update it, and then it'll have the symbols renamed despite it being OFF.

I'm not 100% certain this is going to be a problem, but I needed to point it out because sometimes CMake does weird things and it's usually around conditional logic like this.

Comment (gupta-ak) [reply enclave/crypto/CMakeLists.txt]: @radhikaj The target technically doesn't need to. `mbedcrypto` links against `oecore`, which exports `oe_includes` as a public, so it already has it. I can remove it, so the cmake graph is cleaner.

@andschwa Turns out that CMake is smart enough and it actually rebuilds!

**Comment by gupta-ak** [comment]
> 
> 
> Maybe the title can be modified to reflect the refactoring work as well?

Done.

File: enclave/crypto/CMakeLists.txt (hunk 3 above)
Comment (gupta-ak) [reply enclave/crypto/CMakeLists.txt]: @andschwa I updated the graph and cleaned it up a bit.

File: 3rdparty/mbedtls/rename-mbedtls-symbols [hunk 5]
```diff
@@ -0,0 +1,12 @@
...
+#   readelf -sW: dump symbol table (-s) with the full/wide names (-W).
+#   awk '...': If the symbol is global and begins with mbedtls, then output
+#              the pair $SYMBOL oe_$SYMBOL to feed to objcopy. The seen
+#              array prevents printing duplicates.
+#   objcopy: --redefine-syms will take each $SYMBOL replace it with oe_$SYMBOL.
+readelf -sW "$1" | awk '$5 == "GLOBAL" && $NF ~ /^mbedtls/ && !seen[$NF] {print $NF,"oe_"$NF; seen[$NF]++} {}' > symbols.txt
```
anakrish: As written, this script won't work on Windows. 

1. Is there no plans to support this on Windows?
2. On Windows, we have bash, llvm-objdump. I think llvm also has an llvm-objcopy. Can this script be rewritten to use objdump and objcopy?

enclave/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+add_subdirectory(crypto)
```
CodeMonkeyLeet: @andschwa `enclave/core` is currently getting built by root CMakeLists.txt but now `enclave` has a dependency on `enclave/crypto` as well and has to build it as part of its CMakeList.txt? This is getting increasingly awkward; should this be refactored out into a subfolder instead of keeping it at the root of /enclave? (e.g. even `enclave/enclave` works)

enclave/crypto/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+
+add_library(oecrypto STATIC
+    ../../common/asn1.c
+    ../../common/cert.c
+    ../../common/kdf.c
+    asn1.c
+    cert.c
+    crl.c
+    ec.c
+    entropy.c
```
CodeMonkeyLeet: Why pull entropy and the RDRAND functions out of oecore into oecrypto? These should be part of the PAL (which is fulfilled by oecore today, and possibly Mike's propose oeposix lib in the future):

 - This avoids the weird circular dependency introduced between mbed*.lib builds and the oecrypto.lib

 - With the introduction of device abstractions and a more fully featured syscall handling, it's likely we would want `/dev/random` or `/dev/urandom` to point to hardware provided entropy fundamental impls, not upwards to oecrypto. 

- It needs to be part of our interface for supporting other crypto libs (i.e. the Coco scenario where they want to do a full mbedtls replacement, or for COSIINE to produce an enclave compatible OpenSSL impl to be used by the OE runtime)

tests/CMakeLists.txt
```diff
@@ -35,7 +35,9 @@ if (UNIX OR ADD_WINDOWS_ENCLAVE_TESTS OR USE_CLANGW)
             add_subdirectory(cppException)
             add_subdirectory(ecall)
             add_subdirectory(file)
-            add_subdirectory(mbed)
+            if (NOT OE_PREFIX_MBEDTLS_SYMBOLS)
```
CodeMonkeyLeet: If the user scenario is that a dev can clone our repo and rebuild using this flag so they can quarantine our use of mbedtls, then there should be an expectation that the resulting mbedtls library is tested as well. Can we modify the tests/mbed build to also do the symbol rename if OE_PREFIX_MBEDTLS_SYMBOLS?

3rdparty/mbedtls/CMakeLists.txt
```diff
@@ -73,21 +73,45 @@ ExternalProject_Add(mbedtls-wrap
...
+  set(MBEDX509_LIB_NAME liboembedx509.a)
+  set(MBEDTLS_LIB_NAME liboembedtls.a)
+else()
+  set(MBEDCRYPTO_LIB_NAME libmbedcrypto.a)
+  set(MBEDX509_LIB_NAME libmbedx509.a)
+  set(MBEDTLS_LIB_NAME libmbedtls.a)
+endif()
+
+# Prefix the mbedtls symbols if desired.
+if (OE_PREFIX_MBEDTLS_SYMBOLS)
```
CodeMonkeyLeet: Instead of building `oembed*.lib` and `oecrypto.lib` as substitutes to the regular version, why not have a flag that creates objcopies with the prefixes as an additional set of libs? 

- That way it doesn't disrupt the tests or sample builds, the unprefixed versions always get to run

- It opens up the possibility of us distributing these additional libraries as part of the SDK package (or as an add-on package) instead of requiring users rebuild the SDK themselves. 
   - The primary complication to this is doubling the number of pkgconfig files published just to accommodate the choice of prefixes for oecrypto/mbed* libs.

enclave/crypto/CMakeLists.txt
gupta-ak: `entropy.c` has the `mebdtls_hardware_poll` function, so it's specific to mbedtls. However, I can take the implementation + rd_rand and move it back to oecore. So, theres a platform entropy function. It won't solve the circular dependency however.

anakrish:
Can we clarify the goal of this PR in terms of what benefits it begets the end-user?

If the goal is to allow the user to use their own version of mbedtls library, then I would recommend that we do away with a separate library for crypto routines used internally by the OE SDK.
The crypto routines used by OE SDK ought to have the oe_ prefix, and the object files for those crypto routines ought to be just sucked into liboeenclave.a. Then, any crypto library (as in a separate .a file) packed by OE would be purely for external use or for building high level functionality like tls.

enclave/CMakeLists.txt
andschwa: 🤷‍♂

andschwa: I thought oecore was built by `enclave/core/CMakeLists.txt`...

CodeMonkeyLeet: @andschwa I'm referring to the fact that for apparent ordering reasons we end up doing `add_subdirectory(enclave/core)` in /CMakeLists.txt rather than /enclave/CMakeLists.txt (which is weird because it also has to build oeenclave.

enclave/crypto/CMakeLists.txt
CodeMonkeyLeet: Specifically for `mbedtls_hardware_poll`, @mikbras and I had a discussion a year back about making mbedtls a fully swappable library without having to change OE runtime code, and one of the options there was to make `mbedtls_hardware_poll` part of the mbedtls build (like we do for oelibc) instead of OE code. i.e. it's the responsibility of the swappable library to integrate the use of entropy from "internal 3rdparty public interfaces" OE provides from the PAL. 

The primary advantage of having `mbedtls_hardware_poll` already available in one of our libraries is only for the case where someone wants to do specifically a direct mbedtls replacement, and I would prioritize a more general encapsulation without the circular reference over catering to that one case.

enclave/CMakeLists.txt
andschwa: Oh that is weird. I don't remember doing that.

andschwa: Honestly I'd just try changing that; it's probably not necessary.

gupta-ak: since `3rdparty` depends on `oecore`, doesn't the `oecore` target need to be available before CMake goes in the third party directories?

andschwa: 😬 Only sometimes...CMake has differing rules for when things need to be declared for its different entities. I'd have to test it to see.

gupta-ak:
I'm closing this PR and creating a new one to reflect the changes we discussed internally.

PR #1735, opened 2019-04-26
Fix whitespace in generated args headers

andschwa:
This also removes the `Ast.` qualifier since we `open Ast` (and use it extensively). I don't believe this is something we could have automated with a code formatter, and the rest of my work in progress is based on top of it.

### DIFF
--- a/tests/oeedger8r/enc/CMakeLists.txt
+++ b/tests/oeedger8r/enc/CMakeLists.txt
@@ -9,6 +9,9 @@ oeedl_file(
     --edl-search-dir ../moreedl
 )
 
+# Dummy target used for generating from EDL on demand.
+add_custom_target(edl_gen DEPENDS ${all_t})
+
 oeedl_file(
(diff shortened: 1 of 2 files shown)


tests/oeedger8r/enc/CMakeLists.txt
```diff
@@ -9,6 +9,9 @@ oeedl_file(
     --edl-search-dir ../moreedl
 )
 
+# Dummy target used for generating from EDL on demand.
+add_custom_target(edl_gen DEPENDS ${all_t})
```
anakrish: Not clear what this target does.

andschwa: From the commit:

> This is useful to force the (re)generation of the generated files for
the edger8r test EDL files without having to rebuild anything, extra
useful on systems where the rest of the project doesn't build, such as
my MacBook.

PR #1737, opened 2019-04-26
Apply ocamlformat to Emitter.ml

andschwa:
This applies the default formatting provided by `ocamlformat` to our existing code. We do not yet have this setup in CI and `format-code`, but I would like that to happen. Meanwhile, developers (read: me) can run it automatically via editor integrations.

The default profile by the tool is very close to the style I was applying manually while refactoring, and it is _much_ better to automate it. This tool is suggested by the [OCamlverse Best Practices](https://ocamlverse.github.io/content/best_practices.html) we already ought to be following. I suggest we just go with it.

### DIFF
--- a/tools/oeedger8r/README.md
+++ b/tools/oeedger8r/README.md
@@ -1,63 +1,75 @@
-# Open Enclave Edger8r POC
+The Open Enclave SDK's oeedger8r
+================================
 
-This folder contains Proof Of Concept of adapting Intel Edger8r to generate Open Enclave edge routines.
+The `oeedger8r` tool is a plugin to Intel SGX's Edger8r. We support the same
+Enclave Definition Language syntax, but generate edge routines for use with Open
+Enclave.
 
----
-## Building and Running
+Building and Running
+--------------------
 
-### Build
+The `oeedger8r` is built as part of the Open Enclave SDK CMake build process,
+and shipped in our package.
 
-`$ make`
+To build from source, please follow
+[Advanced Build Info](../../docs/GettingStartedDocs/Contributors/AdvancedBuildInfo.md).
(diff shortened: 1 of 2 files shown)


tools/oeedger8r/README.md
```diff
@@ -1,63 +1,75 @@
-# Open Enclave Edger8r POC
+The Open Enclave SDK's oeedger8r
+================================
 
-This folder contains Proof Of Concept of adapting Intel Edger8r to generate Open Enclave edge routines.
+The `oedger8r` tool is a plugin to Intel SGX's Edger8r. We support the same
```
anakrish: Type oe**e**dger8r.
Here and elsewhere.

tools/oeedger8r/README.md
```diff
@@ -1,63 +1,75 @@
...
+   types in addition to the abstract syntax tree types defined in `Ast.ml`. If
+   `enclave_content` is defined only in `CodeGen.ml`, then it would lead to a
+   cyclic dependency between `CodeGen.ml` and `Plugin.ml`. This is solved by
+   defining the `enclave_content` record in `Ast.ml` and redefining it as an
+   equivalent type in `CodeGen.ml`.
+
+### Edge Routine Emitter
+
+The edge routine emitter for Open Enclave is implemented in `Emitter.ml`. It
+generates code for all the test EDL files. It is work in progress. There is also
```
anakrish: We can remove 'work in progress'. It is quite solid now, extensively tested, and does the job.

andschwa:
@anakrish typos fixed and adding oedgr8r as an abbrev to Emacs to auto fix it in the future 😅 
bors r+

PR #1740, opened 2019-04-30
Refactor oe_cert_verify to remove error string return

CodeMonkeyLeet:
- Remove optional arg for return error string in oe_cert_verify

   - This was not being percolated to user anyway as part of public
     API (e.g. revocation check) and only checked as part of tests.

   - Tests that use this mechanism were not robust and cannot be
     extended easily to accomodate BCrypt equivalents on Windows,
     which does not return error strings.

- Introduce OE_VERIFY_CRL_MISSING and OE_VERIFY_REVOKED error codes
  as specific error codes that can be checked by tests instead.

- Trace underlying verification errors string with logging methods.

- Update cert.c for host and enclave to consistently return
  OE_VERIFY_FAILED for non-specific failures as part of the
  oe_cert_verify call.

- Consolidate calls to openssl/mbedTLS library for cert verification
  for consistent error handling.

- Fix read_mixed_chain in crypto tests to return a valid buffer.

- Fix openssl version of oe_cert_chain_free to return OE_INVALID_PARAMETER
      when passed an invalid cert chain instead of a valid one.

### DIFF
--- a/common/sgx/revocation.c
+++ b/common/sgx/revocation.c
@@ -160,14 +160,12 @@ oe_result_t oe_enforce_revocation(
     oe_cert_chain_t* pck_cert_chain)
 {
     oe_result_t result = OE_FAILURE;
-    oe_result_t r = OE_FAILURE;
     ParsedExtensionInfo parsed_extension_info = {{0}};
     oe_get_revocation_info_args_t revocation_args = {0};
     oe_cert_chain_t tcb_issuer_chain = {0};
     oe_cert_chain_t crl_issuer_chain[3] = {{{0}}};
     oe_parsed_tcb_info_t parsed_tcb_info = {0};
     oe_tcb_level_t platform_tcb_level = {{0}};
-    oe_verify_cert_error_t cert_verify_error = {0};
     char* intermediate_crl_url = NULL;
     char* leaf_crl_url = NULL;
     oe_crl_t crls[2] = {{{0}}};
@@ -240,13 +238,7 @@ oe_result_t oe_enforce_revocation(
     // constraint. If the crl_issuer_chain was different from the certificate
     // chain, then verification would fail because the CRLs will not be found
     // for certificates in the chain.
-    r = oe_cert_verify(
-        leaf_cert, &crl_issuer_chain[0], crl_ptrs, 2, &cert_verify_error);
-    if (r != OE_OK)
-    {
-        OE_RAISE_MSG(
-            r, "oe_cer_verify failed with error = %s\n", cert_verify_error.buf);
-    }
+    OE_CHECK(oe_cert_verify(leaf_cert, &crl_issuer_chain[0], crl_ptrs, 2));
 
     for (uint32_t i = 0; i < OE_COUNTOF(platform_tcb_level.sgx_tcb_comp_svn);
          ++i)
--- a/common/sgx/revocation.c
+++ b/common/sgx/revocation.c
@@ -238,7 +238,8 @@ oe_result_t oe_enforce_revocation(
     // constraint. If the crl_issuer_chain was different from the certificate
     // chain, then verification would fail because the CRLs will not be found
     // for certificates in the chain.
-    OE_CHECK(oe_cert_verify(leaf_cert, &crl_issuer_chain[0], crl_ptrs, 2));
+    OE_CHECK(oe_cert_verify(
+        leaf_cert, crl_issuer_chain, crl_ptrs, OE_COUNTOF(crl_ptrs)));
 
     for (uint32_t i = 0; i < OE_COUNTOF(platform_tcb_level.sgx_tcb_comp_svn);
          ++i)
@@ -273,7 +274,7 @@ oe_result_t oe_enforce_revocation(
     for (uint32_t i = 0; i < OE_COUNTOF(crls); ++i)
     {
         OE_CHECK(oe_crl_get_update_dates(
-            &crls[0], &crl_this_update_date, &crl_next_update_date));
+            &crls[i], &crl_this_update_date, &crl_next_update_date));
 
         _trace_datetime("crl this update date ", &crl_this_update_date);
         _trace_datetime("crl next update date ", &crl_next_update_date);
--- a/enclave/cert.c
+++ b/enclave/cert.c
@@ -205,16 +205,10 @@ OE_INLINE bool _cert_chain_is_valid(const CertChain* impl)
 /*
 **==============================================================================
 **
-** Location helper functions:
+** Local helper functions:
 **
 **==============================================================================
 */
-
-static void _set_err(oe_verify_cert_error_t* error, const char* str)
-{
-    if (error)
-        oe_strlcpy(error->buf, str, sizeof(error->buf));
-}
 
 static bool _x509_buf_equal(
     const mbedtls_x509_buf* x,
@@ -306,7 +300,7 @@ static mbedtls_x509_crt* _sort_certs_by_issue_date(mbedtls_x509_crt* chain)
         p = &chain->next;
         while (*p)
         {
-            //  // Update oldest if a new oldest cert is found.
+            // Update oldest if a new oldest cert is found.
             if (_mbedtls_x509_time_is_before_or_equal(
                     &(*p)->valid_from, &(*p_oldest)->valid_from))
             {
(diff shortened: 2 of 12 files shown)


tests/crypto/read_file.c
```diff
@@ -228,12 +228,8 @@ oe_result_t read_mod(char* filename, uint8_t* mod, size_t* mod_size)
 
 oe_result_t read_mixed_chain(char* chain, char* chain1, char* chain2)
 {
-    char chain_temp[max_cert_chain_size * 2];
-    OE_UNUSED(chain);
-
-    strcat(chain_temp, chain1);
-    strcat(chain_temp, chain2);
-    chain = chain_temp;
+    strcat(chain, chain1);
```
anakrish: The earlier code seems like a bug. chain_temp was not being initialized to a null terminated string.
And the input pointer was just being assigned to a local variable.

What is read_mixed_chain testing?

host/crypto/openssl/cert.c
```diff
@@ -294,49 +274,108 @@ static STACK_OF(X509) * _clone_chain(STACK_OF(X509) * chain)
...
+    if (crls && num_crls)
+    {
+        X509_VERIFY_PARAM* verify_param = NULL;
+
+        for (size_t i = 0; i < num_crls; i++)
+        {
+            crl_t* crl_impl = (crl_t*)crls[i];
+
+            /* X509_STORE_add_crl manages its own addition refcount */
+            if (!X509_STORE_add_crl(store, crl_impl->crl))
```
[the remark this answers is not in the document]
CodeMonkeyLeet: It was a bug. I was tracing it in the debugger and it was leaving the function with more refs on the CRLs than it was entering with since the X509_STORE_add_crl was already managing its own refcount increment. I suspect the original author was not aware of this.

tests/crypto/read_file.c
CodeMonkeyLeet: The earlier code was definitely returning garbage, which causes the `test_mixed_chain` in rsa_tests.c to fail when updated to check for a more specific error code. `read_mixed_chain` is just a helper function for combining two chains, and as used by `test_mixed_chain`; from the code, it is supposed to check that we reject a cert chain presented as `{ca1, root1, ca2, root2}`.

host/crypto/openssl/cert.c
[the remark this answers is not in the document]
anakrish: Great!

andschwa:
From triage meeting, @mikbras would you please review this? Thanks!

common/sgx/revocation.c
```diff
@@ -240,13 +238,7 @@ oe_result_t oe_enforce_revocation(
...
     // chain, then verification would fail because the CRLs will not be found
     // for certificates in the chain.
-    r = oe_cert_verify(
-        leaf_cert, &crl_issuer_chain[0], crl_ptrs, 2, &cert_verify_error);
-    if (r != OE_OK)
-    {
-        OE_RAISE_MSG(
-            r, "oe_cer_verify failed with error = %s\n", cert_verify_error.buf);
-    }
+    OE_CHECK(oe_cert_verify(leaf_cert, &crl_issuer_chain[0], crl_ptrs, 2));
```
mikbras: I know you didn't introduce it, but could you use **OE_COUNTOF(crl_pts)** instead the hard-coded **2**?

Also, I think **crl_issuer_chain** is clearer than **&crl_issuer_chain[0]**.

enclave/cert.c
```diff
@@ -327,19 +321,51 @@ static mbedtls_x509_crt* _sort_certs_by_issue_date(mbedtls_x509_crt* chain)
...
+    mbedtls_x509_crt* ca_cert_chain,
+    mbedtls_x509_crl* ca_crls)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    uint32_t flags = 0;
+    if (mbedtls_x509_crt_verify(
+            leaf_cert, ca_cert_chain, ca_crls, NULL, &flags, NULL, NULL) != 0)
+    {
+        char error[1024] = {0};
+        mbedtls_x509_crt_verify_info(error, sizeof(error), "", flags);
```
mikbras: Check the return value of **mbedtls_x509_crt_verify()**?

enclave/cert.c
```diff
@@ -327,19 +321,51 @@ static mbedtls_x509_crt* _sort_certs_by_issue_date(mbedtls_x509_crt* chain)
...
+        if (flags & MBEDTLS_X509_BADCERT_REVOKED)
+            result = OE_VERIFY_REVOKED;
+        else if (flags & MBEDTLS_X509_BADCRL_EXPIRED)
+            result = OE_VERIFY_CRL_EXPIRED;
+        else
+            result = OE_VERIFY_FAILED;
```
mikbras: Maybe there are other MBEDTLS errors worth catching (like mismatch?), but probably good for now since you will at least see the error message in the log.

enclave/cert.c
```diff
@@ -327,19 +321,51 @@ static mbedtls_x509_crt* _sort_certs_by_issue_date(mbedtls_x509_crt* chain)
...
+
+        OE_RAISE_MSG(
+            result,
+            "mbedtls_x509_crt_verify failed with %s (flags=0x%x)\n",
```
mikbras: I might be helpful to also print the return value from the mbedtls function.

enclave/cert.c
```diff
@@ -731,31 +745,23 @@ oe_result_t oe_cert_verify(
...
-    /* Initialize error */
-    if (error)
-        *error->buf = '\0';
-
     /* Reject invalid certificate */
     if (!_cert_is_valid(cert_impl))
     {
-        _set_err(error, "invalid cert parameter");
-        OE_RAISE(OE_INVALID_PARAMETER);
+        OE_RAISE_MSG(OE_INVALID_PARAMETER, "Invalid cert parameter", NULL);
```
mikbras: The final NULL parameter is not needed.

enclave/cert.c
```diff
@@ -731,31 +745,23 @@ oe_result_t oe_cert_verify(
...
     }
 
     /* Reject invalid certificate chain */
     if (!_cert_chain_is_valid(chain_impl))
     {
-        _set_err(error, "invalid chain parameter");
-        OE_RAISE(OE_INVALID_PARAMETER);
+        OE_RAISE_MSG(OE_INVALID_PARAMETER, "Invalid chain parameter", NULL);
```
mikbras: Again, the final NULL is not needed.

host/crypto/openssl/cert.c
```diff
@@ -294,49 +274,108 @@ static STACK_OF(X509) * _clone_chain(STACK_OF(X509) * chain)
...
+        {
+            case X509_V_ERR_CRL_HAS_EXPIRED:
+                verify_result = OE_VERIFY_CRL_EXPIRED;
+                break;
+            case X509_V_ERR_UNABLE_TO_GET_CRL:
+                verify_result = OE_VERIFY_CRL_MISSING;
+                break;
+            case X509_V_ERR_CERT_REVOKED:
+                verify_result = OE_VERIFY_REVOKED;
+                break;
```
mikbras: Should there be a default case for **OE_VERIFY_FAILED**?

tests/crypto/read_file.c
```diff
@@ -228,12 +228,8 @@ oe_result_t read_mod(char* filename, uint8_t* mod, size_t* mod_size)
...
+    strcat(chain, chain2);
```
mikbras: Consider using **strncat()**?

include/openenclave/bits/result.h
```diff
@@ -265,10 +265,20 @@ typedef enum _oe_result
     OE_UNSUPPORTED_ENCLAVE_IMAGE,
 
     /**
-     * The CRL for a PCK certification expired
+     * The CRL for a PCK certificate has expired.
```
dthaler: I think "PCK" should be deleted from line 268.  The name in 270 doesn't have it, and PCK is very SGX specific unlike the name in line 270.

include/openenclave/bits/result.h
```diff
@@ -265,10 +265,20 @@ typedef enum _oe_result
...
      */
     OE_VERIFY_CRL_EXPIRED,
 
+    /**
+     * The CRL for a PCK certificate could not be found.
```
dthaler: Ditto, "PCK" should be removed from line 273 unless it's added to 275

enclave/cert.c
CodeMonkeyLeet: Technically true because it currently only compiles on linux, but MSVC flags all such instances on the host side and so I've gotten into the habit of requiring explicit NULL for the macro.

enclave/cert.c
CodeMonkeyLeet: There's a fairly long list of errors that it could return and I opted not to inject all of them into our error codes. Even the intersection set with OpenSSL is pretty big (10+). I don't want to reserve error codes for them unless we actually exercise the code paths that test for them, which we don't explicitly verify today.

enclave/cert.c
CodeMonkeyLeet: Considered this, but opted to retain the existing behavior moved from `_verify_whole_chain`; mbedTLS headers do this unusual thing where they define their error codes as a _negation_ of a hex value, so you need to do an extra step just to create string value that's text searchable in the headers. The error string being returned is already more useful in that regard.

enclave/cert.c
CodeMonkeyLeet: This call is for mbedtls_x509_crt_verify_**info** which returns the length of the written error string (which is null terminated regardless of success). Just maintaining the existing logic of printing the OE error message with whatever error info can be salvaged from the message, even if that is the empty string. (This is also how mbedtls internally invokes this function)

host/crypto/openssl/cert.c
CodeMonkeyLeet: I had one originally but took it out because all the switch block does is remap `OE_VERIFY_FAILED` to a more specific error and the default case was just a reassignment of `OE_VERIFY_FAILED ` to the internal `verify_result` that was already initialized to it.

include/openenclave/bits/result.h
CodeMonkeyLeet: Removed

include/openenclave/bits/result.h
CodeMonkeyLeet: Removed

tests/crypto/read_file.c
CodeMonkeyLeet: Good suggestion, switched to using `oe_strlcat()` as a better semantic fit for the crypto tests more broadly.

enclave/cert.c
mikbras: This seems like a flaw with **OE_RAISE_MSG()**. Maybe we can fix as a separate issue.

host/crypto/openssl/cert.c
mikbras: Sounds fine.

tests/crypto/read_file.c
mikbras: Thanks!

enclave/cert.c
mikbras: Sounds good.

enclave/cert.c
mikbras: Yeah, I think we can expand as needed for clarification.

enclave/cert.c
mikbras: Yeah, besides, this funciton probably cannot fail.

CodeMonkeyLeet:
bors retry

Issue #1707, opened 2019-04-16
Enforce integrity of secure channel after remote attestation


AjayBrahmakshatriya:
The remote_attestation sample implements a "proprietary" secure channel (as mentioned in #1325). The algorithm exchanges public keys for both the enclaves and uses these public keys to encrypt the messages to each other. 
But the public keys are visible to the external world including the host. This compromises the integrity of the messages sent over the channel. The host can fake messages on behalf of the enclaves. 

As a fix, the enclaves should use the established ephemeral private keys to perform a signed Diffie-Hellman key exchange and use symmetric key cryptography to communicate after that point. This will also improve the communication overhead. 

Diffie-Hellman key exchange would be the ideal implementation. But an alternate implementation could just generate an ephemeral symmetric key in one of the enclave, sign it with its own private key and then encrypt it with the public key of the other enclave to be sent to the other enclave. This will ensure that the symmetric key is known only to the two enclaves and the root of trust is in the remote attestation.

soccerGB:
Ajay, thanks for pointing out this issue. 
What you pointed out is correct that the channel established in the existing remote attestation sample is susceptible to fake messages originated by any untrusted hosts that have access to the public key.
Given the attested tls feature (#1325) is coming in a couple weeks, what I am going to do is to update the remote attestation to limit it to mutual attestation and point out additional session exchange logics (such as Diffie-Hellman algorithm) in order to establish a really "secure" channel, instead of adding session exchange logics there, which will be soon replaced by the Attested TLS feature.
Let me know what you think

AjayBrahmakshatriya:
Hi Cheng,
When you say point out additional session exchange logics do you mean you would add instructions to the README? Also does this mean you will remove the `ecall_dispatcher::generate_encrypted_message` function and the corresponding `ecall_dispatcher::process_encrypted_msg` function from the sample?
I agree, Attested TLS would solve this completely. Looking forward to that feature.
Thanks

anitagov:
@soccerGB - Can you please clarify if you want to go ahead with the plan of action that you mentioned here.

CodeMonkeyLeet:
Reassigned to @anitagov to implement the reduced scope fix

anitagov:
My PR #1944 addresses this and modifies the remote attestation to remove claims of having a secure channel. The additional session exchange routines after mutual attestation has been completed have been removed. 

There will be other issues/PRs for samples with TLS, etc.
Closing t his as #1944 has been merged to master.

[...]

Issue #1325, opened 2019-01-13
Need to add support for establishing secure channel between enclaves : TLS with remote attestation


soccerGB:
# Integrate attestation to TLS (Transport Layer Security) between Open Enclave enclaves

## Prerequisites

  The audience is assumed to be familiar:
   
  - [Transport Layer Security (TLS)](https://en.wikipedia.org/wiki/Transport_Layer_Security) a cryptographic protocol designed to provide communications security over a computer network.

  - [Open Enclave Remote attestation](https://github.com/Microsoft/openenclave/tree/master/samples/remote_attestation#what-is-attestation): Remote Attestation is the concept of a HW entity or of a
combination of HW and SW gaining the trust of a remote provider or producer.

## Abstract

  As Open Enclave SDK getting adopted into more realistic scenarios, we are receiving requests from OE SDK developers for adding support to establish a TLS channel between two attested enclaves.

  We do have the remote_attestation sample that shows how to conduct remote attestation between two enclaves and establish a `proprietary` channel based on asymmetric keys exchanged during the attestation process. It demonstrates how to conduct mutual attestation but it does not go all the way to show how to establish a fully secure channel. Most of the real world software uses TLS-like standard protocol through popular TLS APIs (OpenSSL, WolfSSL, Mbedtls...) for establishing secure channels.

  Instead of inventing a new communication protocol, the document proposes a solution for addressing above customer need by adding a mechanism into OE SDK to seamlessly integrate remote attestation into the TLS's handshake process without changing any of existing TLS APIs.

### Overview

  The attestation feature that comes with TEE (such as Intel SGX or ARM's TrustZone enclave) could significantly improve a TLS endpoint (client or server) trustworthiness for a TLS connection starting or terminating inside an enclave. 

  By taking advantage of the fact that TLS involving parties use public-key cryptography for identity authentication during the TLS handshaking process, we propose to use a self-signed X509.V3 certificate to represent a TLS endpoint's identity. In this certificate, we make it cryptographically bound to this specific enclave instance by adding a custom certificate extension (called quote extension) with this enclave's attestation quote that has the certificate's public key information embedded. (a new OE API , oe_gen_tls_cert(), will be added to generate this certificate)

  The receiving TLS connection peer will perform the following "Custom extended certificate validation" inside the peer certificate verification callback (all TLS APIs support this callback for custom certificate validation)

  1. Validate certificate
  2. Validate quote extension
  3. Extract this quote extension and perform quote validation
[...]

PR #1744, opened 2019-05-01
Enhanced remote_attestation sample to enforce integrity of secure channel

anitagov:
Addresses issue #1707 

After the remote_attestation succeeds, added the following (high level):
Created an ephemeral symmetric key in one of the enclaves and exchanged it with the other enclave using PKI to establish a secure channel.
Used the symmetric key to encrypt/decrypt the secret message using AES_GCM to communicate after this point for confidentiality and authentication. 
**GCM parameters used**: Random IV, sequence number as additional authenticated data to prevent replay attacks and secret message was the input data to the authenticated encryption function. The output tag, a cryptographic checksum on the data was verified during decryption phase.


samples/remote_attestation/common/crypto.cpp
```diff
@@ -310,3 +303,202 @@ bool Crypto::get_rsa_modulus_from_pem(
...
+    unsigned char iv_str[12];
+
+    bool result = false;
+    int res = -1;
+    size_t iv_len, tag_len = 16;
+
+    mbedtls_gcm_init(&gcm_context);
+
+    memset(iv_str, 0x00, sizeof(iv_str));
+    memcpy(iv_str, "1234567890ab", sizeof("1234567890ab"));
```
CodeMonkeyLeet: **Do not use a fixed IV for AES-GCM.** This is such a common security issue that it has prompted the design of nonce-reuse resistant variants of AES-GCM.

You've also provided a 12-byte string with a length of 16 for some reason?

samples/remote_attestation/common/dispatcher.cpp
```diff
@@ -182,10 +182,173 @@ int ecall_dispatcher::verify_report_and_set_pubkey(
...
+        TRACE_ENCLAVE("ecall_dispatcher initialization failed to generate "
+                      "ephemeral symmetric key.");
+        goto exit;
+    }
+
+    TRACE_ENCLAVE("enclave: establish_secure_channel: Generated random "
+                  "symmetric key successfully");
+
+    // Step 2- Encrypt the symmetric key with the other enclave's public key
+    // TO DO - PREFIX with counter/seq number to prevent replay attacks
```
CodeMonkeyLeet: If you want to do it this way, this is not optional and needs to be part of this changeset.

samples/remote_attestation/common/crypto.cpp
```diff
@@ -194,7 +187,7 @@ bool Crypto::Encrypt(
  * decrypt the given data using current enclave's private key.
  * Used to receive encrypted data from another enclave.
  */
-bool Crypto::decrypt(
+bool Crypto::Decrypt(
```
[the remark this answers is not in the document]
anitagov: This sample has naming violations which will need to be fixed.

samples/remote_attestation/common/crypto.cpp
anitagov: That is a bug in my code. 
I will be changing the fixed IV to be a random one as mentioned in my TODO in Description.  The recommended default length for a IV is 96 bytes as suggested by the GCM spec.  

Started this PR to get valuable feedback. So, thanks a lot.

samples/remote_attestation/common/dispatcher.cpp
anitagov: Yep, agree and plan to implement.

samples/remote_attestation/common/dispatcher.cpp
```diff
@@ -182,10 +182,173 @@ int ecall_dispatcher::verify_report_and_set_pubkey(
...
+            0)
+        {
+            goto exit;
+        }
+
+        TRACE_ENCLAVE("enclave: establish_secure_channel: Computed SHA hash "
+                      "of the encrypted key");
+
+        // Step 4 - Sign this SHA hash with my enclave's private key
+        if (m_crypto->Sign(digest, 32, signature, &signature_size) != 0)
```
gupta-ak: Doing it this way (rsa encrypt -> sign) means that the receiver cannot verify that the signed message actually came from the your enclave.

For example, if Alice, Bob and Charlie are trust each other and Alice sends an encrypted and signed message to Bob (i.e `Sign(Encrypt(Bob_pub, Message), Alice_private)`), then Charlie can intercept the message and replace the signature with his own (i.e `Sign(Encrypt, Bob_pub, Message), Charlie_private)`). They're no way for Alice to verify who actually sent the message without that info embedded in the message.

samples/remote_attestation/common/crypto.cpp
gupta-ak: To add to Simon's point on why not to use a fixed iv:

GCM encryption essentially takes a counter, encrypts it and XORs it with the plain text. (i.e `ciphertext = plaintext XOR Encrypt(Counter)`).  Reusing the IV with the same key means that you can just XOR the ciphertexts together to get the XOR'd plaintext. This might be enough to figure out the plaintext outright.

There's also more complicated attacks that allow the attacker to gain a lot more information on the keys.

samples/remote_attestation/common/dispatcher.cpp
anitagov: You are right that Alice will not know who sent the original message if the signature gets switched by Charlie but the decryption of the message will fail as Charlie's private key will be used to decrypt. _One thing to note is that Charlie will need to first block Bob's receipt of Alice's original message, before sending the re-signed ciphertext._ 

There are issues with both methods of Sign & Encrypt vs Encrypt & Sign. 

So, best approach would be to use the Diffie-Hellman key exchange to establish the shared secret.

samples/remote_attestation/common/crypto.cpp
anitagov: @gupta-ak - Implemented random IV already.

samples/remote_attestation/common/dispatcher.cpp
anitagov: @gupta-ak - You are right that Alice will not know who sent the original message. One thing to note is that Charlie will need to first block Bob's receipt of Alice's original message, before sending the re-signed ciphertext.

There are issues with both methods of Sign & Encrypt vs Encrypt & Sign.
Sign-Encrypt-Sign or Encrypt-Sign-Encrypt are suggested ways to resolve this.

So, best approach would be to use the Diffie-Hellman key exchange to establish the shared secret.

anitagov: @gupta-ak - Wanted to add that in this specific sample, Alice is only expecting message from Bob and will not proceed to decrypt the message if the signature is from some one else, including another trusted party. So, this suffices for this sample.

**Considering adding the name in the encrypted message so that in a general case, Alice can additionally verify that the signature and the encrypted was sent by the same trusted source.** 
Thanks.

samples/remote_attestation/common/crypto.cpp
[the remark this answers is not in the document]
anitagov: Fixed

anitagov:
@anakrish - The goal of this PR is to modify existing remote attestation sample to establish a secure channel after mutual attestation without TLS. The existing sample is also contrived as both enclaves are on the same host.
Another sample using TLS will be added once the TLS feature is in where the enclaves will communicate across different VMs using sockets.

CodeMonkeyLeet:
@anitagov @anakrish @soccerGB I don't think it's a priority for the remote attestation sample to illustrate the establishment of a properly secured channel given the upcoming TLS feature. It is sufficient to demonstrate the use of remote attestation in performing Diffie-Hellman key exchange to reduce the size and scope of the sample.

anakrish:
I agree. It is better to restrict reduce the scope of the attestation sample.
From issue #1707 this PR is supposed to fix, @soccerGB says:

> Given the attested tls feature (#1325) is coming in a couple weeks, what I am going to do is to update the remote attestation to limit it to mutual attestation and point out additional session exchange logics (such as Diffie-Hellman algorithm) in order to establish a really "secure" channel, instead of adding session exchange logics there, which will be soon replaced by the Attested TLS feature.

That indicates that the plan was to _limit the scope of the sample to just mutual attestation and just have pointers to session exchange logics instead of adding session exchange logics_.
This PR seems to do the opposite and tries to implement a key exchange algorithm and establish a secure channel.
There seems to be disconnect between the plan of action in #1707 and this PR. Can anyone provide more insight into this disconnect?

samples/remote_attestation/common/dispatcher.cpp
```diff
@@ -182,10 +182,174 @@ int ecall_dispatcher::verify_report_and_set_pubkey(
...
+            m_enclave_config->sym_key,
+            32,
+            encrypted_key_buf,
+            &encrypted_key_size))
+
+    {
+        // Reason we allocate memory in the host is so that the encrypted data
+        // can be accessed by the hosts - aka real world scenario where we have
+        // 2 hosts and two enclaves
+        uint8_t* host_buf = (uint8_t*)oe_host_malloc(total_size);
```
anakrish: I don't understand this comment. In Trust Zone, which is a real world scenario, enclaves cannot allocate memory on the host and then use the memory like a pointer.
@EMumau took a pass through even our test code and removed calls to oe_host_malloc (as much as possible) since it is bad programming practice.

anitagov: The existing sample still uses oe_host_malloc and so this would need to be removed (probably addressed by another PR).

anitagov: @anakrish - With regards to your earlier comment w.r.t. plan of action specified in #1707 , we (Cheng-mean or I) will need to update the PR with the reasoning for the change (if this gets in). Otherwise, further action will still need to be taken to fix the issues.

samples/remote_attestation/common/dispatcher.cpp
anitagov: Implemented

anitagov:
Thanks for all your valuable feedback and comments.
@CodeMonkeyLeet @soccerGB  and all - I have my alternate implementation coded and running fine. AES-GCM with random IV, sequence number (as additional authentical data) to prevent replay attacks, secret message as the input data to the authenticated encryption function and the output tag, a cryptographic checksum on the data which is verified during decryption phase.

**Issues with the DH key exchange** - The DH key exchange is also vulnerable to the man-in-the-middle attack and is by itself not authenticated. The best part is that there is no guarantee that the person on the other end that you are deriving keys together is whoever they say they are. DH can be used in combination with RSA (or ECDSA) to get a communication channel that are both authenticated and forward secure. Replay attacks will still be an issue. To get a final solution, it will require almost implementing what TLS 1.2/1.3 is doing (non-trivial but can be done). So, my suggestion is to go with my current implementation as the TLS solution is around the corner and will guarantee the secure channel.

My current implementation can serve as a specific use case where there is no transport layer such as TLS.

@soccerGB - Please let me know your thoughts. Thanks.

samples/remote_attestation/common/dispatcher.cpp
```diff
@@ -182,10 +182,177 @@ int ecall_dispatcher::verify_report_and_set_pubkey(
     return ret;
 }
 
+int ecall_dispatcher::establish_secure_channel(uint8_t** key, size_t* key_size)
+{
+    uint8_t encrypted_key_buf[512];
+    uint8_t signature[256];
+    uint8_t digest[32];
+    size_t encrypted_key_size = 256;
+    size_t signature_size = 256;
```
soccerGB: Does this signature_size has to be 256? 
Would 32 be enough?

samples/remote_attestation/README.md
```diff
@@ -43,7 +43,7 @@ There are two types of attestation:
 
 ### Secure Communication Channel
```
soccerGB: It would be great to add explanation to cover the mechanism of secure establishment.

samples/remote_attestation/common/dispatcher.cpp
```diff
@@ -182,10 +182,177 @@ int ecall_dispatcher::verify_report_and_set_pubkey(
...
+        goto exit;
+    }
+
+    ret = 0;
+exit:
+    return ret;
+}
+
 int ecall_dispatcher::generate_encrypted_message(uint8_t** data, size_t* size)
 {
```
soccerGB: This routine seems to assume that a secure channel was already established, which might not be true in all cases.
m_initialized could be re-purpose for it.

samples/remote_attestation/common/dispatcher.cpp
soccerGB: also, it would be great to define constants for those numbers (32, 256, 512 used above) to make it easier to read the code.

samples/remote_attestation/common/dispatcher.cpp
```diff
@@ -182,10 +182,177 @@ int ecall_dispatcher::verify_report_and_set_pubkey(
...
-    uint8_t encrypted_data_buf[1024];
+    uint8_t encrypted_data_buf[ENCLAVE_SECRET_DATA_SIZE];
+    uint8_t tag_str[ENCLAVE_SECRET_DATA_SIZE];
     size_t encrypted_data_size;
+    size_t total_size = ENCLAVE_SECRET_DATA_SIZE * 2 + IV_SIZE;
```
soccerGB: it would be great if you could explain why there is a *2 in the total_size allocation.

samples/remote_attestation/common/dispatcher.cpp
```diff
@@ -195,36 +362,66 @@ int ecall_dispatcher::generate_encrypted_message(uint8_t** data, size_t* size)
...
-            &encrypted_data_size))
+            &encrypted_data_size,
+            tag_str))
     {
-        uint8_t* host_buf = (uint8_t*)oe_host_malloc(encrypted_data_size);
+        // Reason we allocate memory in the host is so that the encrypted data
+        // can be accessed by the hosts - aka real world scenario where we have
+        // 2 hosts and two enclaves. Allocate space for iv_str & tag_str to be
+        // appended after encrypted data
+        total_size = ENCLAVE_SECRET_DATA_SIZE * 2 + IV_SIZE;
```
soccerGB: Not sure why this total_size is being calculated again.

anakrish:
> Overall, the secure channel mechanism looks good. We can add this sample to the extra sample repo (https://github.com/openenclave/openenclave-extra-samples) first after this PR review is completed. (I added a few comments)

@soccerGB I still don't understand the value that such a sample would provide, for the following reasons:

- We will have proper tls implementation soon. That would be the standardized way of establishing a secure channel between two remote enclaves.
- It is still not clear what the plan is to address the original customer reported issue (#1707) and we seem to be keeping that issue open even though that issue has a stated plan of action to reduce the scope of the sample.
- If we really want to demonstrate the establishment of a secure channel that is a not tls, then why not demonstrate how to one enclave can securely make an ecall on another enclave via the host? That is a genuine use case and would help bring feature parity with Intel.

anitagov:
Thanks for your comments @soccerGB and @anakrish 
@soccerGB - Can you please update issue #1707 if you want to go ahead with the reduced scope of the sample as you described it OR
an alternate reduced scope like requested by Simon i.e. demonstrate the use of remote attestation in performing Diffie-Hellman key exchange. This, I believe, is adding the DH key to reportData during mutual Attestation.
I can implement the reduced scope once we agree on what it should entail. Thanks.

This is an interesting PR and I see a new use-case/request with every update. We should capture the proposed use-cases for samples in a new issue and address them separately as this PR is getting overloaded.

samples/remote_attestation/common/dispatcher.cpp
anitagov: ENCLAVE_SECRET_DATA_SIZE is 16 bytes and same as tag length. Yes, makes sense to define a separate constant for TAG_LEN.

samples/remote_attestation/common/dispatcher.cpp
anitagov: Good catch. I meant to remove the first calculation during declaration of total_sense.

samples/remote_attestation/common/dispatcher.cpp
anitagov: Fixed

samples/remote_attestation/common/dispatcher.cpp
anitagov: Done

samples/remote_attestation/common/dispatcher.cpp
anitagov: Yes, good point. Will incorporate.

samples/remote_attestation/README.md
```diff
@@ -43,7 +43,7 @@ There are two types of attestation:
...
 
-Remote Attestation alone is not enough for the remote party to be able to securely deliver their secrets to the requesting enclave. Securely delivering services requires a secure communication channel. Using Asymmetric/Public-Key encryption is a mechanism for establishing such a channel.
+Remote Attestation alone is not enough for the remote party to be able to securely deliver their secrets to the requesting enclave. Securely delivering services requires a secure communication channel which is usually guaranteed by Transport Layer Security (TLS). Using Asymmetric/Public-Key encryption in conjunction with generating a symmetric key based on the sealing key provides an alternate mechanism for establishing such a channel in the absence of TLS support.
```
dthaler: Change "usually guaranteed" to "often guaranteed" (which makes no assertion that it's the majority).
Also change "channel in the absence of TLS support" to "non-TLS channel" since one can choose to not use TLS even if there is TLS support.

samples/remote_attestation/common/dispatcher.cpp
dthaler: You should at least replace the comment in 225 with a TODO saying this won't work and needs to be fixed.  However, since this is apparently new code, we shouldn't add (i.e., merge) new code that won't work.

samples/remote_attestation/common/dispatcher.cpp
```diff
@@ -195,36 +362,66 @@ int ecall_dispatcher::generate_encrypted_message(uint8_t** data, size_t* size)
...
             m_enclave_config->enclave_secret_data,
             ENCLAVE_SECRET_DATA_SIZE,
+            iv_str,
             encrypted_data_buf,
-            &encrypted_data_size))
+            &encrypted_data_size,
+            tag_str))
     {
-        uint8_t* host_buf = (uint8_t*)oe_host_malloc(encrypted_data_size);
+        // Reason we allocate memory in the host is so that the encrypted data
```
dthaler: At least fix grammar...
"We allocate memory in the host so that..."
And add a TODO to fix this.

samples/remote_attestation/common/dispatcher.cpp
```diff
@@ -234,16 +431,37 @@ int ecall_dispatcher::process_encrypted_msg(
...
+
+    memcpy(iv_str, &encrypted_data[ENCLAVE_SECRET_DATA_SIZE], IV_SIZE);
+    memcpy(
+        tag_str,
+        &encrypted_data[ENCLAVE_SECRET_DATA_SIZE + IV_SIZE],
+        ENCLAVE_SECRET_DATA_SIZE);
+
+    // Convert sequence number to 4 character bytes
+    m_enclave_config->sequence_number++;
+    add_str[0] = m_enclave_config->sequence_number & 0xff;
```
dthaler: Do you need this 4-line approach?  Maybe just use htonl?  E.g. something along the lines of
*((uint32_t*)add_str) = htonl(m_enclave_config->sequence_number);
I believe that allows the compiler the opportunity to optimize the byte swap to a single assembly instruction.

samples/remote_attestation/enclave_a/ecalls.cpp
```diff
@@ -56,6 +56,25 @@ int verify_report_and_set_pubkey(
...
+ * Create ephemeral symmetric key, encrypt using other enclave's public key
+ * and sign using private key
+ */
+int establish_secure_channel(uint8_t** data, size_t* size)
+{
+    return dispatcher.establish_secure_channel(data, size);
+}
+
+/**
+ * Verify signature, decrypt using private ke
```
dthaler: typo "ke" at end

samples/remote_attestation/enclave_b/ecalls.cpp
```diff
@@ -56,6 +56,25 @@ int verify_report_and_set_pubkey(
...
+ * Create ephemeral symmetric key, encrypt using other enclave's public key
+ * and sign using private key
+ */
+int establish_secure_channel(uint8_t** data, size_t* size)
+{
+    return dispatcher.establish_secure_channel(data, size);
+}
+
+/**
+ * Verify signature, decrypt using private ke
```
dthaler: typo "ke" at end of line

samples/remote_attestation/remoteattestation.edl
```diff
@@ -18,6 +18,17 @@ enclave {
...
+        // Create ephemeral symmetric key, encrypt using other enclave's public key
+        // and sign using private key
+        public int establish_secure_channel(  [out] uint8_t** data,
+                                              [out] size_t*  size);
+
+        // Verify signature, decrypt using private key
+        // Send back symmetric key after encrypting using other enclave's public key
+        // and sign using private key
+        public int acknowledge_secure_channel(  [in, count=size] uint8_t* data,
+                                               size_t  size);
```
dthaler: spacing looks all messed up since size_t doesn't line up under anything

samples/remote_attestation/common/dispatcher.cpp
anitagov: @dthaler - The existing samples do this and probably will not work for Trust Zone. Let's address this with a new issue. oe_host_malloc needs to be deprecated.

samples/remote_attestation/README.md
anitagov: ok, will incorporate.

samples/remote_attestation/common/dispatcher.cpp
anitagov: Reason I added the 4-lines because it was clear which byte-order was being used. htonl is a good alternative which I hadn't considered. Thanks. Will incorporate.

samples/remote_attestation/enclave_a/ecalls.cpp
anitagov: Saw this and fixed it at my end already. Thanks for flagging all the places that I copied this line over.

samples/remote_attestation/remoteattestation.edl
anitagov: clang-format modified this. Will try to fix this.

anitagov: ok, will fix

samples/remote_attestation/common/dispatcher.cpp
anitagov: Will rephrase.

samples/remote_attestation/common/dispatcher.cpp
anitagov: Established a state machine to make this possible. Definitely cleaner.

samples/remote_attestation/common/dispatcher.cpp
anitagov: Reason I added the 4-lines because it was clear which byte-order was being used and the exact bytes that were being sent. Will consider your suggestion.

anitagov:
bors try

CodeMonkeyLeet:
This should be going into extra-samples repo

samples/remote_attestation/common/dispatcher.cpp
anitagov: Done

samples/remote_attestation/common/dispatcher.cpp
anitagov: Rephrased and also filed an issue to deprecate oe_host_malloc in tests and samples

samples/remote_attestation/common/dispatcher.cpp
anitagov: I am not doing a byte swap but just converting the sequence number to 4 character bytes. Improved the code to just do a memcpy. This should work as both the enclaves are on the same machine. 
Adding htonl, etc. will complicate this sample and require including the socket libraries, etc.

anitagov:
Thanks for all your feedback.
@CodeMonkeyLeet @dthaler - Incorporated your suggestions. Please review.

@soccerGB - Please review the code changes. Only pending thing is fixing the documentation. Will also change the name of this sample to something more appropriate so that it does not conflict with the existing sample.

This is still planned to get into extra-samples.

anitagov:
Closing this as this sample is being renamed to proprietary_secure_channel and will reside in 
https://github.com/openenclave/openenclave-extra-samples. The following PR addresses the checkin to this repository.
https://github.com/openenclave/openenclave-extra-samples/pull/2

PR #1747, opened 2019-05-01
POSIX extensions - host-side POSIX feature

mikbras:
This is the second PR in a series.  It defines the host-side of the POSIX feature, which includes an EDL interface (see **posix.edl**) and the host-side implementation of that interface (see **host/linux/posix.c** and **host/windows/posix.c**). 

Host-side POSIX Interface
-----------------------------

The **host-side POSIX feature** implements OCALLs for the following features.
- **files** - open, read, write, close, stat, fcntl, lseek, access, link, unlink, rename, truncate
- **directories** - opendir, readdir, rewinddir, closedir, mkdir, rmdir
- **sockets** - socket, listen, bind, accept, connect, accept, recv, send, etc.
- **netdb** - getaddrinfo, getnameinfo
- **polling** - epoll_create1, epoll_wait, epoll_ctl
- **unistd** - getuid, getgid, etc.

The Linux implementation is complete. The Windows implementation is ongoing and will be part of a later PR.

Internal EDL Extensions
--------------------------

@andschwa and @anakrish -- please let me know if you see a better way to do this (and thanks Andy for writing our cmake EDL rules). The requirements include:
- Use EDL internally without requiring publication (the EDL functions pass internal structures that cannot be published).
- Provide a way for add-on modules (**liboeposix**) to register their own function tables.
- An **edger8r** option would be nice to support table-ids making the wrappers unnecessary.

This PR also extends the ECALL and OCALL dispatching layer to support alternative function tables. This adds the following functions (see **include/openenclave/internal/calls.h**). 
- **oe_register_ecall_function_table()** -- registers an alternative ECALL function table.
- **oe_register_ocall_function_table()** -- registers an alternative OCALL function table.
- **oe_call_host_function_by_table_id()** -- calls the host function in the table given by a table-id.
- **oe_call_enclave_function_by_table_id()** -- calls the enclave function in the table given by a table-id.

Since the **edger8r** emitter only supports a single ecall/ocall function table, the following wrapper files were developed:

- **host/posix_u_wrapper.c** (includes **posix_u.c**)
- **posix/posix_t_wrapper.c** (includes **posix_t.c**) -- will be in a future PR.


common/posix.edl
```diff
@@ -0,0 +1,395 @@
...
+            [in] oe_host_fd_t sv[2],
+            [out, count=1] int* err);
+
+        int oe_posix_connect_ocall(
+            oe_host_fd_t sockfd,
+            [in, size=addrlen] const struct oe_sockaddr* addr,
+            oe_socklen_t addrlen,
+            [out, count=1] int* err);
+
+        oe_host_fd_t oe_posix_accept_ocall(
```
mikbras: @andschwa and @anakrish: This function defines **addrlen_in** and **addrlen_out**. Is there a way to get by with only one parameter. The following does not seem to work since addlen is a pointer and not an integer.

```
oe_host_fd_t oe_posix_accept_ocall(
     oe_host_fd_t sockfd,
     [in, out, size=addrlen] struct oe_sockaddr* addr,
     [in, out, count=1] oe_socklen_t* addrlen,
     [out, count=1] int* err);
```

common/posix.edl
```diff
@@ -0,0 +1,395 @@
...
+
+        public int oe_posix_polling_notify_ecall(
+            [in, count=num_notifications]
+            struct _oe_device_notifications* notifications,
+            size_t num_notifications);
+    };
+
+    untrusted {
+
+        int oe_posix_uname_ocall(
```
radhikaj: @mikbras Is the err over her obtained from errno. I thought there was a way to annotate edl functions to propagate errno. Would that be useful here?

common/posix.edl
```diff
@@ -0,0 +1,395 @@
...
+            oe_host_fd_t sockfd,
+            int backlog,
+            [out, count=1] int* err);
+
+        ssize_t oe_posix_recvmsg_ocall(
+            oe_host_fd_t sockfd,
+            [out, size=msg_namelen] void* msg_name,
+            oe_socklen_t msg_namelen,
+            [out, count=1] oe_socklen_t* msg_namelen_out,
+            [out, size=msg_buflen] void* msg_buf,
```
radhikaj: I just have a question here, is  struct iovec *msg_iov, int msg_iovlen etc serialized into msg_buf and deserialized later?

common/posix.edl
anakrish: Yes. @crisl-ms implemented the **propagate_errno** EDL feature. Untrusted functions can be annotated with this attribute and the errno values will be automatically propagated back to the enclave.

common/posix.edl
anakrish: @mikbras Your alternative using in/out parameter should work. Most likely an edger8r bug if it does not.
I don't see how your example is different to:
https://github.com/Microsoft/openenclave/blob/ada28f14b4b8f5fdd1450c6e41e12db2cb09248e/tests/oeedger8r/edl/foreign.edl#L15-L19

host/posix_u_wrapper.c
```diff
@@ -0,0 +1,67 @@
...
+#include "hostthread.h"
+
+#if defined(_MSC_VER)
+#define st_atime st_atime
+#define st_mtime st_mtime
+#define st_ctime st_ctime
+#endif
+
+/* Override oe_call_enclave_function() calls with _call_enclave_function(). */
+#define oe_call_enclave_function _call_enclave_function
```
anakrish: @mikbras @andschwa 
I implemented a similar functionality in the branch: https://github.com/Microsoft/openenclave/tree/anakrish/edger8r/internal.
@dthaler had very similar requirements for trustzone/optee of generating marshaling code for internal libraries (sockets, files etc).
My recommendation would be the accept the implementation of "internal" EDL libraries as in this PR with the understanding that we enhance oeedger8r to support internal libraries as first class concept. At that point, we will rework this PR's way of implementing internal libraries.

host/sgx/calls.c
```diff
@@ -684,18 +745,19 @@ oe_result_t oe_ecall(
 /*
 **==============================================================================
 **
-** oe_call_enclave_function()
+** oe_call_enclave_function_by_table_id()
 **
-** Call the enclave function specified by the given function-id.
+** Call the enclave function specified by the given table-id and function-id.
 ** Note: Currently only SGX style marshaling is supported. input_buffer contains
```
anakrish: Actually, this comment is outdated. We support marshalling for both SGX and TrustZone.

host/windows/posix.c
```diff
@@ -0,0 +1,580 @@
...
+**==============================================================================
+*/
+
+#include <io.h>
+#include <stdint.h>
+
+#include "posix_u.h"
+
+/* Definitions in <openenclave/corelibc/errno.h> with Windows headers. */
+#ifdef _OE_ERRNO_H
```
anakrish: I don't quite follow the errno implementation in this PR. errno is standard C. So it ought to be available in Windows too.
The EDL propagate_errno feature might simplify the errno handling and propagation.

include/openenclave/internal/posix.h
```diff
@@ -0,0 +1,95 @@
...
+
+/*
+**==============================================================================
+**
+** oe_device_notifications_t:
+**
+**==============================================================================
+*/
+
+OE_PACK_BEGIN
```
anakrish: Can you describe why the struct ought to be packed?

anakrish:
@mikbras This PR looks good to me. I would have to take a deeper look at the posix stuff and comment on those.
As it stands:
1. I am totally fine with the internal edl tables in this PR. We implemented a similar functionality for TrustZone in a separate branch. Now that we have the same requirements, we ought to add --internal as a first class concept to oeedger8r so that it will satisfy both SGX and TrustZone. However, it does not have to be part of this PR.
2. EDL supports propagate_errno as a first class concept. I would recommend, investigating if that can be used to simplify some of the errno handling.

anakrish:
@mikbras @CodeMonkeyLeet Do you think it would be work documenting what posix functions support internally somewhere? We have often had to spend quite some time trying to keep https://github.com/Microsoft/openenclave/blob/master/docs/LibcSupport.md updated.

common/posix.edl
mikbras: It is a bug. I created issue #1748 to track.

What is different is that the size argument is a pointer to a size, not an actual size.

common/posix.edl
mikbras: That's great news. :) I will reword to use that feature.

Will this work on Windows too?

anakrish: It ought to work on Windows; but hasn't been locked down with a test yet.
In any case, this is a high priority bug fix for edger8r if it does not work on Windows.
CC: @andschwa

CodeMonkeyLeet:
@mikbras @anakrish 

> Do you think it would be work documenting what posix functions support internally somewhere?

Yes, it would be work to do that and I would expect it as part of this series of work, in particular any partial semantics we support for the POSIX APIs. It would be easier to spin up a new doc and convert the existing [LibcSupport.md](https://github.com/Microsoft/openenclave/blob/master/docs/LibcSupport.md) to capture only the standard C headers (which would also need to be updated to reflect any changes in the degree of support due to #1732)

common/posix.edl
mikbras: Yes, that is correct. This is to work around the lack of a deep-copy feature in EDL. Instead of passing an array of iovs (not supported by EDL), we flatten the iov array into a single buffer and pass it to the host. The host constructs an iov array with a single element referring to msg_buf and msg_buflen.

host/posix_u_wrapper.c
mikbras: Yes, I agree. I'm amenable to a better design later on.

common/posix.edl
anakrish: Ah ok. I see it now. It is currently not supported by edger8r. This requirement of an in-out size parameter has cropped up a few times in the past. Would be worth considering if EDL ought to be enhanced to support this scenario.

host/sgx/calls.c
mikbras: Thanks Anand. I pushed a change to remove that outdated comment.

host/windows/posix.c
mikbras: Thanks Anand. I removed that block of code since it is no longer necessary. I added it because the standard errno names conflict with similar names in Windows. I have since restored the OE_ prefixes in <openenclave/corelibc/errno.h> to resolve the issue.

Note that in this module we need to raise the POSIX errnos but still handle the Windows errnos (the two are overlapping).

include/openenclave/internal/posix.h
mikbras: Thanks. We cleaned this up and added a comment to explain this is an overlay of a system struct.

mikbras:
> @mikbras @anakrish
> 
> Do you think it would be work documenting what posix functions support internally somewhere?
> 
> Yes, it would be work to do that and I would expect it as part of this series of work, in particular any partial semantics we support for the POSIX APIs. It would be easier to spin up a new doc and convert the existing LibcSupport.md to capture only the standard C headers (which would also need to be updated to reflect any changes in the degree of support due to #1732)

Agreed. This will come in a later PR in this series. I am holding off pending review of the first few pieces, in case of change and randomization.

common/posix.edl
mikbras: It works fine in EDL. It is the emitter that generates incorrect code.

anakrish: Yes.It is syntactically valid in EDL.
Just like _int float a;_ is syntactically valid but semantically invalid C code. 

EDL currently does not claim to support our example; however it does not emit an _unsupported construct_ error message either. @andschwa we can start off by erroring out and following up with intel if we want to enhance the semantics.

common/posix.edl
mikbras: Yes, I verified it will be fine in windows. Thanks Anand.

andschwa: Oh, we need a unit test around `propagate_errno` for the edger8r too 😄 I'll probably be adding one tomorrow to ensure my upcoming PR doesn't break it.

common/posix.edl
andschwa: @mikbras the upstream EDL parser we use from Intel does not really do much in the way of validation, so there is quite a lot of "valid EDL" code you can write that will parse fine and result in silent bugs. We have some additional warnings around count/size parameters implemented in our edger8r that I'd like to ship back upstream to Intel, and checking if pointers are being used erroneously would be a great extra warning to add until if/when this is better supported.

common/posix.edl
andschwa: Deep-copy will come...!

common/posix.edl
mikbras: Ah I see. Thanks Andy for the update.

common/posix.edl
mikbras: Nice idea. Thanks Andy.

common/posix.edl
mikbras: Yaay! :)

host/windows/posix.c
```diff
@@ -0,0 +1,520 @@
...
+    int flags,
+    oe_mode_t mode)
+{
+    PANIC;
+}
+
+ssize_t oe_posix_read_ocall(oe_host_fd_t fd, void* buf, size_t count)
+{
+    /* ATTN: casting 64-bit fd to 32-bit fd */
+    return _read((int)fd, buf, (uint32_t)count);
```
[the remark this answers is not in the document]
mikbras: Thanks. The Windows implementation is still under development but for now I added overflow checks and raised EBADF on overflow of the file descriptor. More to come.

host/windows/posix.c
```diff
@@ -0,0 +1,520 @@
...
+    uint16_t* d_reclen,
+    uint8_t* d_type,
+    char* d_name,
+    size_t d_namelen)
+{
+    PANIC;
+}
+
+void oe_posix_rewinddir_ocall(uint64_t dirp)
+{
```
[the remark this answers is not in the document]
mikbras: Good catch! Thanks.

host/linux/posix.c
```diff
@@ -0,0 +1,1133 @@
...
+
+typedef struct _wait_args
+{
+    int64_t enclaveid;
+    oe_host_fd_t epfd;
+    int maxevents;
+    struct epoll_event events[];
+} wait_args_t;
+
+static void* epoll_wait_thread(void* arg_)
```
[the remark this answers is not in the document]
mikbras: Thanks for finding this. I missed it during my review of the code.

host/linux/posix.c
```diff
@@ -0,0 +1,1133 @@
...
+    int64_t* d_off,
+    uint16_t* d_reclen,
+    uint8_t* d_type,
+    char* d_name,
+    size_t d_namelen)
+{
+    int ret = -1;
+    struct dirent* ent;
+
+    errno = 0;
```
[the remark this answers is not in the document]
mikbras: The plan is remap Windows codes produced by functions and to rework <openenclave/corelibc/errno.h> to define OE_ prefixed names and then use those constants on the Windows side.

host/linux/posix.c
```diff
@@ -0,0 +1,1133 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
mikbras: Good idea. I will hold this until the next PR so that Bruce will not be randomized (he has windows/posix.c checked out with many changes).

host/sgx/create.c
```diff
@@ -639,6 +640,9 @@ oe_result_t oe_create_enclave(
         (flags & OE_ENCLAVE_FLAG_RESERVED) || config || config_size > 0)
         OE_RAISE(OE_INVALID_PARAMETER);
 
+    /* Install the POSIX ocall function table. */
+    oe_register_posix_ocall_function_table();
```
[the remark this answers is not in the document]
mikbras: I plan to use a different table for those calls since this table (on the enclave side) is registered by a separate add-on library (oeposix) and so it cannot be shared with the table within oecore. This table will most likely have the following table ids:

```
OE_INTERNAL_ECALL_FUNCTION_TABLE_ID 1
OE_INTERNAL_OCALL_FUNCTION_TABLE_ID 1
```
And will be registered with:

```
oe_register_ecall_function_table()
oe_register_ocall_function_table()
```

mikbras:
Thank you everyone for your feedback. I've addressed all comments. I will break posix.c into smaller pieces in the next PR after Bruce checks in **host/windows/posix.c**.

host/linux/posix.c
```diff
@@ -0,0 +1,1133 @@
...
+    errno = 0;
+
+    return access(pathname, mode);
+}
+
+int oe_posix_link_ocall(const char* oldpath, const char* newpath)
+{
+    errno = 0;
+
+    return link(oldpath, newpath);
```
anakrish: @gupta-ak Can edlgen be used to autogen these forwarding functions? Not saying that w should right now, just curious.
CC: @andschwa for future ideas. Another team has expressed interest in having forwarding function auto-generated as well.

host/linux/posix.c
```diff
@@ -0,0 +1,1133 @@
...
+    int64_t enclaveid,
+    oe_host_fd_t epfd,
+    size_t maxevents)
+{
+    int ret = -1;
+    size_t eventsize;
+    pthread_t thread = 0;
+    wait_args_t* args = NULL;
+
+    eventsize = sizeof(struct oe_epoll_event) * maxevents;
```
anakrish: @EMumau Seems like epoll implementation here could benefit from switchless calling.

host/linux/posix.c
```diff
@@ -0,0 +1,1133 @@
...
+    {
+        OE_STATIC_ASSERT(sizeof(args->fds[0]) == sizeof(fds[0]));
+        memcpy(&args->fds[fd_idx], &fds[fd_idx], sizeof(fds[fd_idx]));
+    }
+
+    // We lose the wait thread when we exit the func, but the thread will die
+    // on its own copy args then spawn pthread to do the waiting. That way we
+    // can ecall with notification. the thread args are freed by the thread
+    // func.
+    if (pthread_create(&thread, NULL, _poll_wait_thread, args) < 0)
```
anakrish: @mikbras Shouldn't poll be a blocking call? Right now the poll happens in the created _poll_wait_thread, while the current thread would continue execution and return control back to the enclave. Is the enclave going to spin on some variable till the _poll_wait_thread completes?

@EMumau This seems relevant to switchless calling.
@andschwa Your ideas in this regard also appreciated.

host/linux/posix.c
mikbras: Interesting future idea. :)

host/linux/posix.c
mikbras: Yes, that would be nice. At the very least we should cache some host memory.

host/linux/posix.c
mikbras: Good observation Anand. I brought this up also and I too would like to get rid of this thread. I discussed this with Bruce and it is on my list.

Here is how it works now:

1. The enclave calls **oe_posix_epoll_wait_async_ocall()**, which spins up this thread that watches for events.
2. The enclave returns and then enters **oe_wait_device_notification()** where it waits (on a condition variable) for event notions from (1) the host thread, and (2) other enclave-side devices.

Note that there are at least two event sources: host and enclave.

I proposed an alternative approach.

1. Enclave calls **oe_posix_epoll_wait()** and waits in the host.
2. Control returns to the enclave when events are ready or when another enclave thread wakes the host out of the wait with a function that will be called **oe_posix_epoll_wake()**.

Do you think this is something we should change now, or later?

host/windows/posix.c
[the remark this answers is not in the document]
CodeMonkeyLeet: @mikbras The change looks fine although I'm not sure I got an answer to my question: the truncation technically also happens on the [Linux side](https://linux.die.net/man/2/read) which is why I asked why `oe_host_fd_t` is defined by to be wider than `int`.

host/linux/posix.c
anakrish: @mikbras 
In current approach, if the enclave is going to wait on a condition variable, the variable in turn would transition back to the host perform a thread wait. Not sure what that is buying us.

The alternative approach is straight-forward and feels more natural.
You can implement the alternative approach as appropriate,  in this PR or the next where the enclave side gets checked in.

host/linux/posix.c
gupta-ak: Yeah, it can generate the C calls on the enclave+host side and the edl as well. I made it also do things like make it work properly on `accept`, so like `accept(int, struct sockaddr*, socklen_t*) -> accept(int, [in,out, size=socklen_in] struct sockaddr*, [in, out] socklen_t*, size_t socklen_in)`.

Since it was a prototype, I didn't make it templatized, so the generated code looks different, but if there is an ask from another team, we could add to it and make it a proper tool.

host/windows/posix.c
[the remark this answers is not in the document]
mikbras: File descriptors are uniformly 64-bits across all POSIX functions. This is to allow Windows options for how to implement files. Today we are using integers, tomorrow we might use HANDLE, depending on the underlying implementation. Also, note that Windows defines 64-bit socket descriptors. This was not my choice. I would have preferred to use int throughout the interface and to map ints to handles on the Windows side. This is a change I intend to make in a later PR.

host/linux/posix.c
mikbras: I cannot defend the original approach since it is not mine and I had the same observation that you did. Therefore: I reworked polling to wait on the host side.

host/linux/posix.c
```diff
@@ -0,0 +1,1099 @@
...
+
+    if (nfds < 0)
+        goto done;
+
+    /* Remove the dummy event for the wakefd. */
+    for (int i = 0; i < nfds; i++)
+    {
+        if (events[i].data.u64 == WAKEFD_MAGIC)
+        {
+            events[i] = events[nfds - 1];
```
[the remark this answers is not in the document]
mikbras: :)

host/linux/posix.c
```diff
@@ -0,0 +1,1099 @@
...
+
+        /* Treat as an interrupt if no other descriptors are read. */
+        if (nfds == 0)
+        {
+            errno = EINTR;
+            goto done;
+        }
+    }
+
+    ret = nfds;
```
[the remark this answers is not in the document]
mikbras: Agreed. We will review that with enclave side.

host/linux/posix.c
```diff
@@ -0,0 +1,1099 @@
...
+        }
+    }
+
+    /* Read the word that oe_posix_epoll_wake_ocall() wrote. */
+    if (found_wake_event)
+    {
+        int fd = -1;
+        uint64_t c;
+
+        pthread_mutex_lock(&_epolls_lock);
```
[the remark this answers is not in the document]
mikbras: Switched to spinlock.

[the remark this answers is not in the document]
mikbras: Only one thread may call epoll_wait() on a given epoll instance at  a time. There is no problem with the epolls[] array, but I don't think this scenario is meaningful.

host/linux/posix.c
```diff
@@ -0,0 +1,1099 @@
...
+
+    if (!buf)
+    {
+        errno = EINVAL;
+        goto done;
+    }
+
+    if ((ret = uname(&uts)) != -1)
+    {
+        oe_strlcpy(buf->sysname, uts.sysname, sizeof(buf->sysname));
```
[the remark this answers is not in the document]
mikbras: Good find. I am erroring out now. For example:

```
        /* sysname: */
        {
            if (strlen(uts.sysname) >= sizeof(buf->sysname))
            {
                errno = ENAMETOOLONG;
                goto done;
            }

            strcpy(buf->sysname, uts.sysname);
        }
```

mikbras:
bors try

PR #1751, opened 2019-05-02
Add phony targest sample makefiles

aserdean:
[samples][Makefile]: Add phony targets for data-sealing/Makefile
[samples][Makefile]: Add phony targets for file-encryptor/Makefile
[samples][Makefile]: Add phony targets for helloworld/Makefile
[samples][Makefile]: Add phony targets for local_attestation/Makefile
[samples][Makefile]: Add phony targets for remote_attestation/Makefile

Fixes: https://github.com/Microsoft/openenclave/issues/1750

### DIFF
--- a/samples/local_attestation/Makefile
+++ b/samples/local_attestation/Makefile
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+.PHONY: all build clean run
+
 all: build
 
 build:
--- a/samples/data-sealing/Makefile
+++ b/samples/data-sealing/Makefile
@@ -1,6 +1,10 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+.PHONY: all build clean run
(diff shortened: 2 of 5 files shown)


samples/local_attestation/Makefile
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+.PHONY: all build clean run
```
anitagov: @aserdean - Shouldn't the order be clean build run?

aserdean: @anitagov the actual order does not matter, I just chose an alphabetical one 😄 .
https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

This PR mainly is to fix the issue if someone is using CMake and creates a `build` directory while also trying to use the GNU makefiles. For that case we needed a phony target `build`. I added the rest for the theoretical performance gain.

anitagov: OK, great. Thanks for the interesting link.

PR #1761, opened 2019-05-03
Added cert generation and verification APIs for attested TLS

soccerGB:
This is certification generation (oe_gen_tls_cert) and verification (oe_verify_tls_cert) part of attested tls feature.
https://github.com/Microsoft/openenclave/issues/1325

I am hoping to preliminary feedback whether the implementation of those two new APIs impacts other features/refactoring activities under construction.

End-2-end test and samples will come in a separate PR once its depending components (socket) are ready.


3rdparty/mbedtls/mbedtls/library/x509write_crt.c
```diff
@@ -308,7 +308,7 @@ int mbedtls_x509write_crt_der( mbedtls_x509write_cert *ctx, unsigned char *buf,
     unsigned char *c, *c2;
     unsigned char hash[64];
     unsigned char sig[MBEDTLS_MPI_MAX_SIZE];
-    unsigned char tmp_buf[2048];
+    unsigned char tmp_buf[2048*4];
```
johnkord: Oh I remember you talking about this a few months ago, that mbedtls doesn't support large enough asn1 tags to hold a quote. That's unfortunate! It would mean that a user that wants to replace the mbedtls library at the upcoming oecrypto interface with a different version of mbedtls, they would need to make sure their mbedtls library is patched like this. I'm fine with that, as long as we make it clear to the user replacing this! :)

soccerGB: Yes. That's true.
I am in a process of trying to get those mbedtls buffer size changes to upstream.
(openssl does not have this limitation)

johnkord: Nice!

enclave/cert.c
```diff
@@ -1047,3 +1135,137 @@ oe_result_t oe_cert_chain_get_leaf_cert(
...
+
+    // Write a built up certificate to a X509 DER structure Note: data
+    // is written at the end of the buffer! Use the return value to
+    // determine where you should start using the buffer.
+    *bytes_written = (size_t)mbedtls_x509write_crt_der(
+        &x509cert, buff, cert_buf_size, mbedtls_ctr_drbg_random, &ctr_drbg);
+    if (*bytes_written <= 0)
+        OE_RAISE_MSG(OE_FAILURE, "bytes_written = 0x%x ", *bytes_written);
+
+    oe_memcpy_s(
```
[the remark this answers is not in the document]
radhikaj: There are a couple more places with oe_memcpy_s

common/sgx/tlsverifier.c
```diff
@@ -0,0 +1,156 @@
...
+
+oe_result_t verify_report_user_data(
+    uint8_t* key_buff,
+    size_t key_buff_size,
+    uint8_t* report_data);
+
+static const char* oid_oe_report = X509_OID_FOR_QUOTE_STRING;
+
+// verify report user data against peer certificate
+oe_result_t verify_report_user_data(
```
radhikaj: Thanks Chengmean. This is inline with the oecrypto philosophy and looks good.

3rdparty/mbedtls/mbedtls/library/x509write_crt.c
anakrish: We have avoided editing 3rd party code as much as possible. Can we instead make these changes via a patch that gets applied as part of the build system? That way these changes will not be lost when we upgrade mbedtls.

anakrish: Allocating 8 K variable on the stack would increase the risk of stack overflow.
Is this a good crypto practice?
Is that how OpenSSL handles large data?

common/sgx/tlsverifier.c
```diff
@@ -0,0 +1,156 @@
...
+    // find the extension
+    report = (uint8_t*)oe_malloc(report_size);
+    if (!report)
+        OE_RAISE(OE_OUT_OF_MEMORY);
+
+    OE_CHECK(oe_cert_find_extension(
+        &cert, (const char*)oid_oe_report, report, &report_size));
+    OE_TRACE_VERBOSE("extract_x509_report_extension() succeeded");
+
+#ifdef _OE_ENCLAVE_H
```
anakrish: Use OE_BUILD_ENCLAVE instead.

enclave/tls_cert.c
```diff
@@ -0,0 +1,215 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// clang-format off
```
anakrish: Why clang format needs to be turned off?

enclave/tls_cert.c
```diff
@@ -0,0 +1,215 @@
...
+#include "key.h"
+#include "rsa.h"
+
+// Todo: consider set CN with enclave's MRENCLAVE values
+#define ISSUER_NAME "CN=Open Enclave SDK,O=OESDK TLS,C=UK"
+#define SUBJECT_NAME ISSUER_NAME
+#define DATE_NOT_VALID_BEFORE "20190501000000"
+#define DATE_NOT_VALID_AFTER "20501231235959"
+
+static unsigned char _cert_buf[MAX_CERT_SIZE] = {
```
anakrish: Is this thread-safe?

common/sgx/tlsverifier.c
```diff
@@ -0,0 +1,156 @@
...
+    // set enclave to NULL because we are dealing only with remote report now
+    //------------------------------------------------------------------------
+
+    // determine the size of the extension
+    if (oe_cert_find_extension(
+            &cert, (const char*)oid_oe_report, NULL, &report_size) !=
+        OE_BUFFER_TOO_SMALL)
+        OE_RAISE(OE_FAILURE);
+
+    // find the extension
```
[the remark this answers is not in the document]
soccerGB: It's a wrong spot.  It should be been moved down a couple lines

enclave/cert.c
```diff
@@ -1047,3 +1135,137 @@ oe_result_t oe_cert_chain_get_leaf_cert(
...
+        0,
+        (const uint8_t*)config->ext_data_buf,
+        config->ext_data_buf_size);
+    if (ret)
+        OE_RAISE_MSG(OE_FAILURE, "ret = 0x%x ", ret);
+
+    // Write a built up certificate to a X509 DER structure Note: data
+    // is written at the end of the buffer! Use the return value to
+    // determine where you should start using the buffer.
+    *bytes_written = (size_t)mbedtls_x509write_crt_der(
```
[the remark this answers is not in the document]
soccerGB: cert_buf_size is greater or equal to byte_written

3rdparty/mbedtls/mbedtls/library/x509write_crt.c
soccerGB: This is a very good point. I will make it's done as suggested.

enclave/cert.c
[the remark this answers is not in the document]
soccerGB: thanks for pointing this issue. They are all addressed.

enclave/tls_cert.c
soccerGB: It was added from earlier prototype work, it;s not needed anymore.

common/sgx/tlsverifier.c
soccerGB: Thanks for the suggestion. Addressed

enclave/tls_cert.c
soccerGB: The header reordering from running clang formatting causes compiler errors on the cert related type defines.

enclave/core/string.c
```diff
@@ -138,4 +138,22 @@ char* oe_strstr(const char* haystack, const char* needle)
...
+
+    /* Copy at most n bytes. Terminate when src is exhausted. */
+    while (n-- && *src)
+        *dest++ = *src++;
+
+    /* If there is room left, then inject zero-terminator. */
+    if (n)
+        *dest = '\0';
+
+    return ret;
```
anakrish: @mikbras @soccerGB @CodeMonkeyLeet @paulcallen @gupta-ak 
The need to implement and review oe_strncpy is one of the reasons why corelibc is more trouble than it is worth.

This implementation of oe_strncpy  looks correct - copy at most n characters from src to dst - a _bounded string-copy_.

However this implementation is incorrect because, counterintuitively, strncpy has different semantics according to the C standard. strncpy requires that the destination be padded with null characters if the src contains less that n characters. 

From [strncpy C++ reference](http://www.cplusplus.com/reference/cstring/strncpy/)


> Copies the first num characters of source to destination. If the end of the source C string (which is signaled by a null-character) is found before num characters have been copied, **destination is padded with zeros until a total of num characters have been written to it.**

From [Ansi C Rationale](http://www.lysator.liu.se/c/rat/d11.html)

>  strncpy was initially introduced into the C library to deal with fixed-length name fields in structures such as directory entries.  Such fields are not used in the same way as strings: the trailing null is unnecessary for a maximum-length field, and setting trailing bytes for shorter names to null assures efficient field-wise comparisons.  strncpy is not by origin a ``bounded strcpy,'' and the Committee has preferred to recognize existing practice rather than alter the function to better suit it to such use. 

C libraries take great pains to implement the standard; no matter how counter-intuitive the standard is. Take for example MUSL's implementation of   which strncpy forwards to:
https://github.com/microsoft/openenclave/blob/fe0c6af59d1d3ec038c47cda52555cdf351b10f8/3rdparty/musl/musl/src/string/stpncpy.c#L10-L29

MUSL's strncpy follows the C standard by null-padding the destination bytes. Additionally, when used as a glibc replacement, it does glibc compatible optimizations.

Since crypto is a key component in writing secure code, we need to shy away from writing new code in the hot path. Instead we need to rely on battle tested code like MUSL or GLIBC (along with their security fixes).
If we start implementing libc functions, when we are already using a full fledged libc implementation like MUSL, we ought to lock down each C function we implement rigorously with tests and prove to ourselves that we are standards compliant. At this moment I don't see rigorous tests for corelibc.

soccerGB:
bors  try

enclave/tls_cert.c
soccerGB: It is NOT! :-(  Good catch. 
Global memory block was replaced with dynamically allocated local memory.

soccerGB:
bors try

enclave/core/string.c
anitagov: Agree with Anand here.

include/openenclave/enclave.h
```diff
@@ -724,6 +724,55 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * @param cert If not NULL, the buffer to free.
+ */
+void oe_free_tls_cert(uint8_t* cert);
+
+/**
+ * oe_verify_tls_cert
+ *
+ * This function preform a custom validation on the input certificate. This
+ * validation includes exacting a quote extension from the certificate before
+ * performing a quote validation on it. An optional enclave_identity_callback
```
CodeMonkeyLeet: The callback should not be optional if the default behavior means allowing a TLS connection to any arbitrary enclave.

include/openenclave/host.h
```diff
@@ -384,6 +384,25 @@ void oe_free_key(
     uint8_t* key_info,
     size_t key_info_size);
 
+/**
+ * oe_verify_tls_cert
```
CodeMonkeyLeet: It's not obvious that this is not general support for TLS certs; even in the enclave we expect that they will want to auth against non-enclave servers and so will need deal with normal TLS cert approval.

We might want to consider qualifying this as an `oe_verify_enclave_tls_cert` or something with a special term indicating our unique cert format e.g. `oe_verify_attestation_cert`

include/openenclave/internal/cert.h
```diff
@@ -134,6 +153,22 @@ oe_result_t oe_cert_verify(
...
+ *
+ * @param cert verify this certificate
+ * @param error Optional. Holds the error message if this function failed.
+ *
+ * @return OE_OK verify ok
+ * @return OE_VERIFY_FAILED
+ * @return OE_INVALID_PARAMETER
+ * @return OE_FAILURE
+ */
+oe_result_t oe_verify_self_signed_cert(
```
CodeMonkeyLeet: Do we need a separate method? I'm a little uncomfortable with this, since it doesn't actually even validate that the cert is self-signed. I would prefer making `chain` an optional parameter in `oe_cert_verify` and have the impl do the self-signed check instead of the cert chain validation.

include/openenclave/internal/report.h
```diff
@@ -161,6 +161,13 @@ OE_STATIC_ASSERT(sizeof(oe_report_header_t) == 16);
 OE_STATIC_ASSERT(
     OE_OFFSETOF(oe_report_header_t, report) == sizeof(oe_report_header_t));
 
+// ISO(1).ANSI(2).USA(840).Microsoft(113556).ACC(10).Classes(103).Subclass(1)
```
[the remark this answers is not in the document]
soccerGB: Mismatched class id fixed.
Still figuring out how to register this OID

include/openenclave/host.h
```diff
@@ -384,6 +384,25 @@ void oe_free_key(
...
+ *
+ * This function preform a custom validation on the input certificate. This
+ * validation includes exacting a quote extension from the certificate before
```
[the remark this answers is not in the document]
soccerGB: good catch!

fixed.

include/openenclave/host.h
```diff
@@ -384,6 +384,25 @@ void oe_free_key(
...
+ * performing a quote validation on it. An optional enclave_identity_callback
+ * could be passed in for a calling client to further validate the identity of
+ * the enclave creating the quote.
+ * @retval OE_OK on a successful validation
+ * @retval appropriate error code when failed.
```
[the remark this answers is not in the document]
soccerGB: API comment updated

include/openenclave/enclave.h
soccerGB: When the connection is between a non-sgx host client and a sgx enclave server,  the server might not request any certificate from the client, in this case, it does not need to register this callback

include/openenclave/enclave.h
```diff
@@ -724,6 +724,55 @@ oe_enclave_t* oe_get_enclave(void);
...
+    size_t subject_key_size,
+    uint8_t** output_cert,
+    size_t* output_cert_size);
+/**
+ * Free the given cert
+ * @param cert If not NULL, the buffer to free.
+ */
+void oe_free_tls_cert(uint8_t* cert);
+
+/**
```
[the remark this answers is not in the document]
soccerGB: addressed

include/openenclave/enclave.h
```diff
@@ -724,6 +724,55 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * This function preform a custom validation on the input certificate. This
+ * validation includes exacting a quote extension from the certificate before
+ * performing a quote validation on it. An optional enclave_identity_callback
+ * could be passed in for a calling client to further validate the identity of
+ * the enclave creating the quote.
+ * @retval OE_OK on a successful validation
+ * @retval appropriate error code when failed.
+ */
+typedef oe_result_t (
+    *oe_enclave_identity_verify_callback_t)(oe_identity_t* identity, void* arg);
```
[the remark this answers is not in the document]
soccerGB: Suggestion taken.

include/openenclave/internal/cert.h
```diff
@@ -314,6 +349,33 @@ oe_result_t oe_get_crl_distribution_points(
     uint8_t* buffer,
     size_t* buffer_size);
 
+#ifdef _OE_ENCLAVE_H
```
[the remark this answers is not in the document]
soccerGB: They are specific to enclaves.
Are you suggesting to create a new cert.h under enclave?

include/openenclave/internal/cert.h
```diff
@@ -314,6 +349,33 @@ oe_result_t oe_get_crl_distribution_points(
...
+    unsigned char* issuer_name;
+    unsigned char* date_not_valid_before;
+    unsigned char* date_not_valid_after;
+    uint8_t* ext_data_buf;
+    size_t ext_data_buf_size;
+    char* ext_oid;
+    size_t ext_oid_size;
+} oe_cert_config_t;
+
+#define MAX_CERT_SIZE 8 * 1024
```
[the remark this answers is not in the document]
soccerGB: good idea.

enclave/tls_cert.c
```diff
@@ -0,0 +1,205 @@
...
+    result = oe_get_public_key_by_policy(
+        OE_SEAL_POLICY_UNIQUE,
+        &params,
+        public_key,
+        public_key_size,
+        NULL,
+        NULL);
+    OE_CHECK(result);
+
+    result = oe_get_private_key_by_policy(
```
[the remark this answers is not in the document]
soccerGB: It's the caller of oe_gen_tls_cert to provide keys.
This generate_key_pair function is no longer part of api implementation after the unused generate_key_pair function is removed.

enclave/tls_cert.c
```diff
@@ -0,0 +1,205 @@
...
+    result = OE_OK;
+    cert_buf = NULL;
+done:
+    oe_free(cert_buf);
+    return result;
+}
+
+// input: input_data and input_data_len
+// output: key, key_size
+oe_result_t generate_key_pair(
```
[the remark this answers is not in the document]
soccerGB: This was missed during the code refactoring cleanup process.
Thanks for pointing this out.

common/sgx/tlsverifier.c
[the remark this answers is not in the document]
soccerGB: Yeah, it's only used in the this file, "static" works.

tests/tls_apis/tls.edl
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+enclave {
+    trusted {
+        public oe_result_t get_TLS_cert([out] unsigned char** data, [out] size_t* data_size);
```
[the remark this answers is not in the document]
soccerGB: renamed

tls_apis/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
soccerGB: This is a unittest for those new added APIs.
README.md added

The duplicated copy under root is removed. (thanks for spotting this)

3rdparty/mbedtls/mbedtls/library/x509write_crt.c
dthaler: Even better would be to NOT copy 3rd party code in the first place, but instead use a git submodule, and then you can submit into a branch (fork) of mbedtls that you depend on, until the branch gets merged.

dthaler: Yes stack space can be a problem. Doing a malloc is safer than consuming lots of stack space.

enclave/cert.c
```diff
@@ -634,6 +635,55 @@ oe_result_t oe_cert_read_pem(
...
+    /* Clear the implementation */
+    if (impl)
+        memset(impl, 0, sizeof(Cert));
+
+    /* Check parameters */
+    if (!der_data || !der_size || !cert)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+    /* Allocate memory for the certificate */
+    if (!(crt = mbedtls_calloc(1, sizeof(mbedtls_x509_crt))))
```
dthaler: Personally I think it's bad for a core OE file to directly hard code mbedTLS specific API calls instead of going through an abstraction layer.   I thought that was being addressed.

enclave/core/string.c
dthaler: +1.  If we have a POSIX name we must adhere to the standard.  strncpy as specified in the standard is dangerous since it does not guarantee null termination (as explained at the link Anand cites).   As such, if we make oe_strncpy be public, then it should probably be marked as deprecated or unsafe or something, so it will generate warnings if an app tries to use it.  strncpy_s is the safer API to use.

enclave/tls_cert.c
```diff
@@ -0,0 +1,164 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// clang-format off
+ #include <openenclave/bits/defs.h>
```
dthaler: remove initial whitespace in 5-6, for consistency with subsequent lines

include/openenclave/enclave.h
```diff
@@ -724,6 +724,55 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * @param subject_key: a public key used to in the certificate
+ * @param subject_key_size The size of the subject_key buffer.
+ *
+ * @param output_cert: a pointer to buffer pointer, where the buffer is
+ * allocated by a client
+ * @param output_cert_size: size of the buffer above.
+ *
+ * @return OE_OK on success
+ */
+oe_result_t oe_gen_tls_cert(
```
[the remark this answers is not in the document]
dthaler: Thanks Simon.  Strong +1 to using full verbs not abbreviations (acronyms like tls are ok, just not abbreviations).   The only exception is to match an ANSI or POSIX standard (e.g., strcmp uses abbreviations).

include/openenclave/enclave.h
```diff
@@ -724,6 +724,61 @@ oe_enclave_t* oe_get_enclave(void);
...
 
+/**
+ * oe_gen_tls_cert.
+ *
+ * This function generates a self-signed certificate with an embedded quote from
+ * the underlying enclave.
+ *
+ * @param issuer_key: a private key used to sign this certificate
+ * @param issuer_key_size The size of the issuer_key buffer.
+ * @param subject_key: a public key used to in the certificate
```
dthaler: can't parse "used to in".  Also a nit that some lines end in . and some don't

include/openenclave/enclave.h
```diff
@@ -724,6 +724,61 @@ oe_enclave_t* oe_get_enclave(void);
  */
 oe_result_t oe_random(void* data, size_t size);
 
+/**
+ * oe_gen_tls_cert.
+ *
+ * This function generates a self-signed certificate with an embedded quote from
+ * the underlying enclave.
+ *
+ * @param issuer_key: a private key used to sign this certificate
```
dthaler: Can we please use the direction arguments [in] or [out] or [in,out] to param to better document the syntax in any public headers?  See http://www.doxygen.nl/manual/commands.html#cmdparam

include/openenclave/enclave.h
```diff
@@ -724,6 +724,61 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * @return OE_OK on success
+ */
+oe_result_t oe_gen_tls_cert(
+    uint8_t* issuer_key,
+    size_t issuer_key_size,
+    uint8_t* subject_key,
+    size_t subject_key_size,
+    uint8_t** output_cert,
+    size_t* output_cert_size);
+/**
```
dthaler: insert blank line between 750 and 751

include/openenclave/enclave.h
```diff
@@ -724,6 +724,61 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * Free the given cert
+ * @param cert If not NULL, the buffer to free.
+ */
+void oe_free_tls_cert(uint8_t* cert);
+
+/**
+ * identity validation callback type
+ * @param identity: a pointer to an enclave's identity information
```
dthaler: Incorrect doxygen syntax.  Remove colon.  See http://www.doxygen.nl/manual/commands.html#cmdparam

include/openenclave/enclave.h
```diff
@@ -724,6 +724,61 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * @param arg: caller defined context
+ */
+typedef oe_result_t (
+    *oe_identity_verify_callback_t)(oe_identity_t* identity, void* arg);
+
+/**
+ * oe_verify_tls_cert
+ *
+ * This function preform a custom validation on the input certificate. This
```
dthaler: typo: "perform" -> "performs"

include/openenclave/enclave.h
```diff
@@ -724,6 +724,61 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * validation includes exacting a quote extension from the certificate before
```
dthaler: Did you mean "extracting" instead of "exacting"?

common/sgx/tlsverifier.c
```diff
@@ -63,7 +58,7 @@ oe_result_t verify_report_user_data(
 oe_result_t oe_verify_tls_cert(
```
radhikaj: Chengmean, this API can be called from within an enclave and outside an enclave, correct? In both cases does it have to be compiled with SGX=ON?

tests/tls_apis/enc/enc.cpp
```diff
@@ -0,0 +1,179 @@
...
+        OE_TRACE_ERROR(" failed with %s\n", oe_result_str(result));
+        goto done;
+    }
+
+    OE_TRACE_INFO("output_cert_size = 0x%x", output_cert_size);
+    // validate cert inside the enclave
+    result = oe_verify_tls_cert(
+        output_cert, output_cert_size, enclave_identity_verifier, NULL);
+    OE_TRACE_INFO(
+        "\nFrom inside encalve: verifying SGX certificate extensions... %s\n",
```
dthaler: Delete "SGX", this API should not be SGX specific.  Also "encalve" is misspelled

tests/tls_apis/enc/enc.cpp
```diff
@@ -0,0 +1,179 @@
...
+    if (identity->security_version < 1)
+    {
+        OE_TRACE_ERROR(
+            "identity->security_version checking failed (%d)\n",
+            identity->security_version);
+        goto done;
+    }
+
+    // Dump an enclave's unique ID, signer ID and Product ID. They are
+    // MRENCLAVE, MRSIGNER and ISVPRODID for SGX enclaves In a real scenario,
```
dthaler: nit: Fix punctuation (add period after "enclaves").  It's correct in host.cpp but enc.cpp here doesn't match host.cpp.

tests/tls_apis/enc/enc.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+done:
+
+    if (private_key)
+        oe_free_key(private_key, private_key_size, NULL, 0);
+    if (public_key)
+        oe_free_key(public_key, public_key_size, NULL, 0);
+
+    // free certificate buffer
+    if (output_cert)
```
dthaler: Delete line 167, since oe_free_tls_cert already handles null pointers.  Probably same for lines 161 and 163.

tests/tls_apis/host/host.cpp
```diff
@@ -0,0 +1,113 @@
...
+        oe_put_err("oe_create_enclave(): result=%u", result);
+
+    OE_TRACE_INFO("Host: get tls certificate from an enclave \n");
+    result = get_tls_cert(enclave, &ecall_result, &cert, &cert_size);
+    if ((result != OE_OK) || (ecall_result != OE_OK))
+        oe_put_err("get_tls_cert() failed: result=%u", result);
+
+    fflush(stdout);
+
+#if 1
```
dthaler: Why is there "#if 1" code in here?  Remove?

tests/tls_apis/host/host.cpp
```diff
@@ -0,0 +1,113 @@
...
+    }
+#endif
+
+    // validate cert
+    OE_TRACE_INFO("Host: Verifying tls certificate\n");
+    OE_TRACE_INFO("Host: cert = %p cert_size = %d\n", cert, cert_size);
+    result =
+        oe_verify_tls_cert(cert, cert_size, enclave_identity_verifier, NULL);
+    OE_TRACE_INFO(
+        "Host: Verifying SGX certificate extensions from host ... %s\n",
```
dthaler: Remove "SGX"

3rdparty/mbedtls/mbedtls/library/x509write_crt.c
soccerGB: replaced with dynamic memory allocation.

tests/tls_apis/enc/enc.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include <openenclave/internal/tests.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "tls_t.h"
+
+//#define printf oe_host_printf
+
+// This is the identity validation callback. An TLS connecting party (client or
```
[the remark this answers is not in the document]
soccerGB: Thanks for offline explanation.

Fixed as suggested.

enclave/tls_cert.c
```diff
@@ -0,0 +1,205 @@
...
+
+// Todo: consider set CN with enclave's MRENCLAVE values
+#define ISSUER_NAME "CN=Open Enclave SDK,O=OESDK TLS,C=UK"
+#define SUBJECT_NAME ISSUER_NAME
+#define DATE_NOT_VALID_BEFORE "20190501000000"
+#define DATE_NOT_VALID_AFTER "20501231235959"
+
+static const unsigned char oid_oe_report[] = X509_OID_FOR_QUOTE_EXT;
+
+oe_result_t calc_sha256(uint8_t* buf, size_t buf_size, OE_SHA256* sha256)
```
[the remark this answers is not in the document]
soccerGB: addressed

include/openenclave/enclave.h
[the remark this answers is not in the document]
soccerGB: addressed

include/openenclave/host.h
soccerGB: Thanks for the great feedback.

I have renamed them to the following :
oe_generate_attestation_cert
oe_free_attestation_cert
oe_verify_attestation_cert

enclave/tls_cert.c
soccerGB: removed

tests/tls_apis/enc/enc.cpp
soccerGB: Fixed.

tests/tls_apis/enc/enc.cpp
soccerGB: fixed

tests/tls_apis/host/host.cpp
soccerGB: removed

include/openenclave/enclave.h
soccerGB: fixed

include/openenclave/enclave.h
soccerGB: addressed for those new APIs in this PR.
https://github.com/microsoft/openenclave/issues/1804 is opened to track the work for updating all other existing OE SDK API

include/openenclave/enclave.h
soccerGB: done

include/openenclave/enclave.h
soccerGB: fixed

include/openenclave/enclave.h
soccerGB: fixed

include/openenclave/enclave.h
soccerGB: It should be extracting. Corrected

tests/tls_apis/enc/enc.cpp
soccerGB: deleted

tests/tls_apis/host/host.cpp
soccerGB: removed

include/openenclave/internal/cert.h
soccerGB: Very good suggestion.
I replaced it with or_cert_verify as suggested.

3rdparty/mbedtls/mbedtls/library/x509write_crt.c
soccerGB: I opened a mbetls issue https://github.com/ARMmbed/mbedtls/issues/2631 for tracking this mbedtls limitation. And an PR with a clean in-place solution (https://github.com/ARMmbed/mbedtls/pull/2632) was posted an ARM developer shortly after my posting. 

With that said,  once we pick up the mbedtls release with above PR, we do not need to maintain the private mbedtls changes I included in this PR.

include/openenclave/internal/cert.h
```diff
@@ -56,6 +56,25 @@ oe_result_t oe_cert_read_pem(
...
+ * The caller is responsible for releasing the certificate by passing it to
+ * oe_cert_free().
+ *
+ * @param cert initialized certificate handle upon return
+ * @param der_data DER data
+ * @param der_size size of the DER data
+ *
+ * @return OE_OK load was successful
+ */
+oe_result_t oe_cert_read_der(
```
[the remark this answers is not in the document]
soccerGB: A new test case, _test_der_cert_verify_good, was added for exercising this newly added internal API.

common/sgx/tlsverifier.c
soccerGB: Thanks for pointing out this concern, which will be addressed when the following PR is merged.

https://github.com/microsoft/openenclave/pull/1575

enclave/cert.c
soccerGB: This cert.c is part of the crypto implementation for the enclave, which depends solely on the mbedtls library. 
Before that dependency was replaced/removed, we will have to keep it this way.

enclave/tls_cert.c
```diff
@@ -0,0 +1,205 @@
...
+// clang-format on
+
+#include <stdio.h>
+#include "../common/common.h"
+#include "ec.h"
+#include "key.h"
+#include "rsa.h"
+
+// Todo: consider set CN with enclave's MRENCLAVE values
+#define ISSUER_NAME "CN=Open Enclave SDK,O=OESDK TLS,C=UK"
```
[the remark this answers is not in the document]
soccerGB: Thanks for bringing this up.
I added a subject name parameter to the oe_generate_attestation_cert() api, which would give the caller a chance to customize their cert.

include/openenclave/enclave.h
```diff
@@ -724,6 +724,76 @@ oe_enclave_t* oe_get_enclave(void);
  */
 oe_result_t oe_random(void* data, size_t size);
 
+/**
+ * oe_generate_attestation_cert.
```
dthaler: Don't use abbreviations ("cert") in non-POSIX API names.

include/openenclave/enclave.h
```diff
@@ -724,6 +724,76 @@ oe_enclave_t* oe_get_enclave(void);
...
+ *
+ * This function generates a self-signed x509 certificate with an embedded
```
dthaler: x509 -> X.509

include/openenclave/enclave.h
```diff
@@ -724,6 +724,76 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * quote from the underlying enclave.
+ *
+ * @param[in] subject_name a string contains contain an X.509 distinguished
```
dthaler: "contains contain" -> "containing"

include/openenclave/enclave.h
```diff
@@ -724,6 +724,76 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * name (DN) for customizing the generated certificate. This name is also used
+ * as the issuer name because this is a self-signed certificate
+ * See RFC5280 (https://tools.ietf.org/html/rfc5280) specification for details.
+ * Example value "CN=Open Enclave SDK,O=OESDK TLS,C=US"
+ *
+ * @param[in] private_key a private key used to sign this certificate
```
dthaler: nit: some lines end in period (740, 742) and some don't (739, 741).  Be consistent throughout this file.

include/openenclave/enclave.h
```diff
@@ -724,6 +724,76 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * @param[in] identity a pointer to an enclave's identity information
+ * @param[in] arg caller defined context
+ */
+typedef oe_result_t (
+    *oe_identity_verify_callback_t)(oe_identity_t* identity, void* arg);
+
+/**
+ * oe_verify_attestation_cert
+ *
+ * This function preform a custom validation on the input certificate. This
```
dthaler: "preform" -> "perform"

include/openenclave/enclave.h
```diff
@@ -724,6 +724,76 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * validation includes extracting a quote extension from the certificate before
+ * performing a quote validation on it. An optional enclave_identity_callback
+ * could be passed in for a calling client to further validate the identity of
+ * the enclave creating the quote.
+ * @param[in] cert_in_der a pointer to buffer holding certificate contents
```
dthaler: "... in DER format"?

include/openenclave/enclave.h
```diff
@@ -724,6 +724,76 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * @param[in] cert_in_der_len size of certificate buffer above
+ * @param[in] enclave_identity_callback callback routine for custom identity
+ * checking
+ * @param[in] arg optional argument
```
dthaler: what is "arg" used for?  "optional argument" is not a helpful description

include/openenclave/enclave.h
```diff
@@ -724,6 +724,76 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * @param[in] arg caller defined context
+ */
+typedef oe_result_t (
+    *oe_identity_verify_callback_t)(oe_identity_t* identity, void* arg);
+
+/**
+ * oe_verify_attestation_cert
+ *
+ * This function preform a custom validation on the input certificate. This
+ * validation includes extracting a quote extension from the certificate before
```
dthaler: I think the term "quote" is SGX specific and does not belong in the doxygen comments at all, unless used in an example where it's clear it's using SGX as the example

include/openenclave/host.h
```diff
@@ -384,6 +384,39 @@ void oe_free_key(
...
+ * @param[in] identity a pointer to an enclave's identity information
+ * @param[in] arg caller defined context
+ */
+typedef oe_result_t (
+    *oe_identity_verify_callback_t)(oe_identity_t* identity, void* arg);
+
+/**
+ * oe_verify_attestation_cert
+ *
+ * This function preform a custom validation on the input certificate. This
```
dthaler: Same comments in host.h as I added into enclave.h

enclave/core/string.c
soccerGB: This is strncpy was added for the purpose getting mbedtls built on top of corelibc, where once we starting building mbedtls on top of the musl, this direct corelibc dependency will be removed thus the concern raised will be gone because this strncpy will become no longer needed.

include/openenclave/host.h
soccerGB: addressed

include/openenclave/enclave.h
soccerGB: rephrased to 
 * @param[in] arg an optional context pointer argument specified by the
 * application when setting callback

include/openenclave/enclave.h
soccerGB: added

include/openenclave/enclave.h
soccerGB: thanks for spotting this. Fixed.

include/openenclave/enclave.h
soccerGB: thanks for spotting this. Fixed.

include/openenclave/enclave.h
soccerGB: thanks for spotting this. Fixed.

include/openenclave/enclave.h
soccerGB: thanks for spotting this. Fixed.

include/openenclave/enclave.h
soccerGB: thanks for spotting this. 
Renamed to 
oe_generate_attestation_certificate
oe_free_attestation_certificate
oe_verify_attestation_certificate

include/openenclave/enclave.h
soccerGB: good point. 
I replaced "quote" with "attestation evidence".

tests/tls_apis/enc/enc.cpp
```diff
@@ -0,0 +1,337 @@
...
+        int res = -1;
+        mbedtls_ctr_drbg_context ctr_drbg_contex;
+        mbedtls_entropy_context entropy_context;
+        mbedtls_pk_context pk_context;
+        size_t local_public_key_size = 512;
+        size_t local_private_key_size = 2048;
+
+        mbedtls_ctr_drbg_init(&ctr_drbg_contex);
+        mbedtls_entropy_init(&entropy_context);
+        mbedtls_pk_init(&pk_context);
```
gupta-ak: This doesn't get freed with `mbedtls_pk_free`, which causes the memory leak in CI.

soccerGB: good catch! Thanks.
Fixed

include/openenclave/internal/cert.h
```diff
@@ -305,6 +344,33 @@ oe_result_t oe_get_crl_distribution_points(
     uint8_t* buffer,
     size_t* buffer_size);
 
+#ifdef _OE_ENCLAVE_H
+// TODO: add comments
```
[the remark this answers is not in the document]
soccerGB: addressed

include/openenclave/internal/cert.h
```diff
@@ -169,6 +188,26 @@ oe_result_t oe_cert_get_ec_public_key(
     const oe_cert_t* cert,
     oe_ec_public_key_t* public_key);
 
+/**
+ * Get the public key (in PRM frfrom a certificate.
+ *
+ * This function gets the EC public key from the given certificate. If the
```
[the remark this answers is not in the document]
soccerGB: fixed

include/openenclave/internal/cert.h
```diff
@@ -169,6 +188,26 @@ oe_result_t oe_cert_get_ec_public_key(
     const oe_cert_t* cert,
     oe_ec_public_key_t* public_key);
 
+/**
+ * Get the public key (in PRM frfrom a certificate.
```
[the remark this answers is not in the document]
soccerGB: corrected

include/openenclave/host.h
```diff
@@ -384,6 +384,41 @@ void oe_free_key(
...
+ * @param[in] arg an optional context pointer argument specified by the caller
+ * when setting callback
+ * @retval OE_OK on a successful validation
+ * @retval OE_VERIFY_FAILED on quote failure
+ * @retval OE_INVALID_PARAMETER At least one parameter is invalid
+ * @retval OE_FAILURE general failure
+ * @retval other appropriate error code
+ */
+oe_result_t oe_verify_attestation_certificate(
+    uint8_t* der_crt,
```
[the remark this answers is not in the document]
soccerGB: fixed

include/openenclave/enclave.h
```diff
@@ -724,6 +724,78 @@ oe_enclave_t* oe_get_enclave(void);
...
+ * @param[in] private_key_size The size of the private_key buffer
+ * @param[in] public_key a public key used as the certificate's subject key
+ * @param[in] public_key_size The size of the public_key buffer.
+ *
+ * @param[out] output_cert a pointer to buffer pointer
+ * @param[out] output_cert_size size of the buffer above
+ *
+ * @return OE_OK on success
+ */
+oe_result_t oe_generate_attestation_certificate(
```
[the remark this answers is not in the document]
soccerGB: clarified by adding function description before this function

enclave/tls_cert.c
```diff
@@ -0,0 +1,159 @@
...
+        public_key,
+        public_key_size,
+        remote_report_buf,
+        remote_report_buf_size,
+        output_cert,
+        output_cert_size);
+    OE_CHECK_MSG(
+        result, "generate_x509_cert failed : %s", oe_result_str(result));
+
+    OE_TRACE_VERBOSE("self-signed certificat size = %d", *output_cert_size);
```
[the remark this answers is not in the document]
soccerGB: typo fixed

enclave/tls_cert.c
```diff
@@ -0,0 +1,159 @@
...
+
+    // generate quote with hash(cert's subject key) and set it as report data
+    OE_TRACE_VERBOSE(
+        "generate quote with hash from public_key_size=%d public_key key "
+        "=\n[%s]\n",
+        public_key_size,
+        public_key);
+    oe_memset_s(sha256.buf, OE_SHA256_SIZE, 0, OE_SHA256_SIZE);
+    OE_CHECK(oe_sha256_init(&sha256_ctx));
+    OE_CHECK(oe_sha256_update(&sha256_ctx, public_key, public_key_size));
```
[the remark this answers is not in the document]
soccerGB: Two supported types are : RSA and EC

enclave/tls_cert.c
```diff
@@ -0,0 +1,159 @@
...
+#define SUBJECT_NAME "CN=Open Enclave SDK,O=OESDK TLS,C=US"
+#define DATE_NOT_VALID_BEFORE "20190501000000"
+#define DATE_NOT_VALID_AFTER "20501231235959"
+
+static const unsigned char oid_oe_report[] = X509_OID_FOR_QUOTE_EXT;
+
+// Input: an issuer and subject key pair
+// Output: a self-signed certificate embedded critical extension with quote
+// information as its content
+oe_result_t generate_x509_cert(
```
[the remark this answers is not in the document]
soccerGB: addressed

enclave/cert.c
```diff
@@ -649,6 +653,55 @@ oe_result_t oe_cert_read_pem(
...
+        OE_RAISE(OE_OUT_OF_MEMORY);
+
+    /* Initialize the certificate structure */
+    mbedtls_x509_crt_init(crt);
+
+    /* Parse a single DER formatted certificate and add it to the chained list
+     */
+    rc = mbedtls_x509_crt_parse_der(crt, (const uint8_t*)der_data, der_size);
+    if (rc != 0)
+        OE_RAISE(OE_FAILURE, "mbedtls_x509_crt_parse rc= 0x%x\n", rc);
```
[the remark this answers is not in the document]
soccerGB: replaced

enclave/cert.c
```diff
@@ -1019,3 +1078,155 @@ oe_result_t oe_cert_chain_get_leaf_cert(
...
+    if (ret)
+        OE_RAISE_MSG(OE_FAILURE, "ret = 0x%x ", ret);
+
+    ret = mbedtls_x509write_crt_set_issuer_name(
+        &x509cert, (const char*)config->issuer_name);
+
+    if (ret)
+        OE_RAISE_MSG(OE_FAILURE, "ret = 0x%x ", ret);
+
+    ret = mbedtls_mpi_read_string(&serial, 10, "1");
```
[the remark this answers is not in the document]
soccerGB: It should be ok because this field is not used in the validation.

enclave/cert.c
```diff
@@ -1019,3 +1078,155 @@ oe_result_t oe_cert_chain_get_leaf_cert(
...
+        &issuer_key,
+        (const unsigned char*)config->private_key_buf,
+        config->private_key_buf_size,
+        NULL,
+        0);
+    if (ret)
+        OE_RAISE_MSG(OE_FAILURE, "ret = 0x%x ", ret);
+
+    // init certificate
+    mbedtls_x509write_crt_init(&x509cert);
```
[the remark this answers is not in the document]
soccerGB: good suggestion. updated

enclave/cert.c
```diff
@@ -1019,3 +1078,155 @@ oe_result_t oe_cert_chain_get_leaf_cert(
...
+    mbedtls_x509write_cert x509cert = {0};
+    mbedtls_pk_context subject_key;
+    mbedtls_pk_context issuer_key;
+    mbedtls_ctr_drbg_context ctr_drbg;
+    mbedtls_entropy_context entropy;
+    unsigned char* buff = NULL;
+    const char* pers = "OE SDK cert";
+    int ret = 0;
+
+    if ((buff = oe_malloc(cert_buf_size)) == NULL)
```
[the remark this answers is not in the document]
soccerGB: fixed

enclave/cert.c
```diff
@@ -1019,3 +1078,155 @@ oe_result_t oe_cert_chain_get_leaf_cert(
...
+        OE_RAISE(OE_OUT_OF_MEMORY);
+
+    mbedtls_pk_init(&subject_key);
+    mbedtls_pk_init(&issuer_key);
+    mbedtls_mpi_init(&serial);
+    mbedtls_ctr_drbg_init(&ctr_drbg);
+    mbedtls_entropy_init(&entropy);
+
+    ret = mbedtls_ctr_drbg_seed(
```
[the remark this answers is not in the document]
soccerGB: good suggestion, fixed.

common/sgx/tlsverifier.c
```diff
@@ -0,0 +1,147 @@
...
+    result = oe_verify_report(NULL, report, report_size, &parsed_report);
+#endif
+    OE_CHECK(result);
+    OE_TRACE_VERBOSE("quote validation succeeded");
+
+    // verify report data: hash(public key)
+    // extract public key from the cert
+    oe_memset_s(pub_key_buf, sizeof(pub_key_buf), 0, sizeof(pub_key_buf));
+    result =
+        oe_cert_write_public_key_pem(&cert, pub_key_buf, &pub_key_buf_size);
```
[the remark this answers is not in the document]
soccerGB: You are right. I replaced it with dynamic memory allocation (2k).

soccerGB:
bors try

tests/tls_apis/enc/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+include(oeedl_file)
+
+oeedl_file(../tls.edl enclave gen)
+
+add_executable(tls_enc enc.cpp ${gen})
+
+target_include_directories(tls_enc PRIVATE
+                          ${CMAKE_CURRENT_BINARY_DIR}
+                          ${PROJECT_SOURCE_DIR}/include/openenclave/corelibc)
```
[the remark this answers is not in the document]
soccerGB: This is an internal test case, where a couple of internal oe_* api, defined in oecorelibc, were used. That's why corelibc path was included. 
In the upcoming external same code, it will solely base on the musl headers.

3rdparty/mbedtls/mbedtls/library/x509write_crt.c
```diff
@@ -406,48 +422,67 @@ int mbedtls_x509write_crt_der( mbedtls_x509write_cert *ctx, unsigned char *buf,
...
+     * making sure that it doesn't underflow
+     * into the CRT buffer. */
     c2 = buf + size;
-    MBEDTLS_ASN1_CHK_ADD( sig_and_oid_len, mbedtls_x509_write_sig( &c2, buf,
+    MBEDTLS_ASN1_CHK_ADD( sig_and_oid_len, mbedtls_x509_write_sig( &c2, c,
                                         sig_oid, sig_oid_len, sig, sig_len ) );
 
-    if( len > (size_t)( c2 - buf ) )
-        return( MBEDTLS_ERR_ASN1_BUF_TOO_SMALL );
+    /*
```
anakrish: @soccerGB The previous version returned MBEDTLS_ERR_ASN1_BUF_TOO_SMALL. 
I cannot spot where the new version returns the same error code.

enclave/cert.c
```diff
@@ -649,6 +654,55 @@ oe_result_t oe_cert_read_pem(
...
+    const void* der_data,
+    size_t der_size)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    Cert* impl = (Cert*)cert;
+    mbedtls_x509_crt* crt = NULL;
+    int rc = 0;
+
+    /* Clear the implementation */
+    if (impl)
```
anakrish: If an impl already exists, then shouldn't we be cleaning up contents of the impl rather than just doing memset?

enclave/core/string.c
anakrish: Can you open an issue to keep track of removing this strncpy implementation or fixing it to be libc compliant before the next release? Otherwise, we would increase the risk of unpredictable behavior/potential security vulnerability.

host/crypto/openssl/cert.c
```diff
@@ -665,7 +707,7 @@ oe_result_t oe_cert_verify(
     }
 
     /* Check for invalid chain parameter */
-    if (!_cert_chain_is_valid(chain_impl))
+    if (chain && !_cert_chain_is_valid(chain_impl))
```
anakrish: Should this be `chain_impl &&` ?

3rdparty/mbedtls/mbedtls/library/x509write_crt.c
```diff
@@ -298,25 +298,25 @@ static int x509_write_time( unsigned char **p, unsigned char *start,
...
     size_t sub_len = 0, pub_len = 0, sig_and_oid_len = 0, sig_len;
     size_t len = 0;
     mbedtls_pk_type_t pk_alg;
 
     /*
-     * Prepare data to be signed in tmp_buf
+     * Prepare data to be signed at the end of the target buffer
      */
-    c = tmp_buf + sizeof( tmp_buf );
+    c = buf + size;
```
dthaler: These look like changes to mbedtls itself.  Are they being contributed back upstream to the master mbedtls?  If not, why not?  If so, what's the link to the pull request or commit?

enclave/cert.c
dthaler: I can't parse the last sentence in your reply.   Do you mean "Until that is done"?  meaning it's future work to fix this?  Or what?   I still maintain its bad to call mbedtls APIs directly from OE core code.

enclave/tls_cert.c
```diff
@@ -0,0 +1,205 @@
...
+#include "../common/common.h"
+#include "ec.h"
+#include "key.h"
+#include "rsa.h"
+
+// Todo: consider set CN with enclave's MRENCLAVE values
+#define ISSUER_NAME "CN=Open Enclave SDK,O=OESDK TLS,C=UK"
+#define SUBJECT_NAME ISSUER_NAME
+#define DATE_NOT_VALID_BEFORE "20190501000000"
+#define DATE_NOT_VALID_AFTER "20501231235959"
```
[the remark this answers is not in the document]
dthaler: I think it's bad to hard code any timestamps in code, as opposed to in admin-configurable policy.

include/openenclave/enclave.h
dthaler: Nit is not yet fixed.  See line 742

include/openenclave/enclave.h
```diff
@@ -724,6 +724,78 @@ oe_enclave_t* oe_get_enclave(void);
...
+    uint8_t* public_key,
+    size_t public_key_size,
+    uint8_t** output_cert,
+    size_t* output_cert_size);
+
+/**
+ * Free the given cert
+ * @param[in] cert If not NULL, the buffer to free.
+ */
+void oe_free_attestation_certificate(uint8_t* cert);
```
dthaler: style comment: In Microsoft public headers and sample code, we try to avoid abbreviations not just in method names, but also variable/parameter names.  "cert" should be expanded to "certificate" in 755, 756, 760, 762, etc., for consistency with the method name.  (Acronyms on the other hand are fine, so "der" is perfectly fine in 794.)   There are a very small number of exceptions where abbreviations are ok.  Those include "app" and "arg" (which is used in this file and is ok.)

tests/attestation_cert_apis/README.md
```diff
@@ -0,0 +1,13 @@
+oe_generate_attestation_cert, oe_free_attestation_cert, and oe_verify_attestation_cert API tests
+=====================
+
+Test scenario:
+
+- **Host Side**
+  1. Create an enclave
+  2. Issue an ecall (get_tls_cert) into enclave for getting a self-signed certificate embedded with an quote of the enclave
+  3. Once the certificate is received, call oe_verify_attestation_cert to verify the certificate and quote
```
dthaler: cert -> certificate (and same in oe_generate_attestation_cert on line 12)

tests/attestation_cert_apis/host/host.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "tls_u.h"
+
+#define TEST_EC_KEY 0
+#define TEST_RSA_KEY 1
+#define SKIP_RETURN_CODE 2
+
+// This is the identity validation callback. An TLS connecting party (client or
```
dthaler: grammar nit: "An TLS" -> "A TLS"

tests/attestation_cert_apis/host/host.cpp
```diff
@@ -0,0 +1,145 @@
...
+        // for testing purpose, output the whole cer in DER format
+        char filename[80];
+        FILE* file = NULL;
+
+        sprintf(
+            filename,
+            "./cert_%s.der",
+            test_type == TEST_RSA_KEY ? "rsa" : "ec");
+        OE_TRACE_INFO(
+            "Host: Log quote embedded certificate to file: [%s]\n", filename);
```
dthaler: I believe the term "quote" is SGX specific.  Use platform agnostic terms in trace output from non-SGX-specific code

PR #1763, opened 2019-05-04
Refactor the edger8r

andschwa:
This refactoring pays back the technical debt of the edger8r having previously been a prototype. The main goal of the work were to make the edger8r easier to understand. To that end, most functions now read "forwards" instead of "backwards," that is, you can look at the list of strings and see skeleton of the code it is supposed to generate. Most formatting has been removed from the string content and instead applied to the lists of strings, again making it easier to parse the embedded C code.

All code was tested for current usage, with a fair amount of dead code identified and removed. As the refactoring progressed, better patterns were identified and implemented. Notably, embedded pattern matching to identify different pointer types was mostly replaced with maps onto lists instead filtered by predicates. Some code generation was identified as duplicate and so refactored into reusable functions; conversely some code which had no business being independent functions was "unfactored" to read more plainly.

The generated files now (mostly) pass `format-code`, with the exception of whitespace on empty lines and extra-long lines, both problems not being worth fixing for generated code. Also of note, when there is nothing to emit, a comment to that effect is now produced instead of an extra blank line, making it obvious to the reader that indeed no code should exist there. All generated comments and headers are now consistent and easy to identify, with more section headers introduced for easier reading of the edge routines.

Due to the nature of the work, this code is best reviewed by commit instead of by the whole diff. Most commits in the series should build without further commits, except where noted.

Still TODO: I identified a bad pattern I'd introduced in the predicates where I'm matching `(ptype, decl)` and throwing away the `decl`, as these functions became increasingly reused it was made obvious their signature should change to just `ptype`, but it was too late to do this retroactively. Expect another commit or two on Monday, with that change, the addition of a unit test for errno propagation, a cleanup of the member decl generation logic (note internal TODOs), and maybe a commit to group functions logically (now that it is possible to do so).

### DIFF
--- a/tools/oeedger8r/Emitter.ml
+++ b/tools/oeedger8r/Emitter.ml
@@ -616,15 +616,19 @@ let oe_gen_ecall_functions (tfs : trusted_func list) =
   if tfs <> [] then List.flatten (List.map oe_gen_ecall_function tfs)
   else ["/* There were no ecalls. */"]
 
-let oe_gen_ecall_table (os : out_channel) (ec : enclave_content) =
-  fprintf os "\n\n/****** ECALL function table  *************/\n" ;
-  fprintf os "oe_ecall_func_t __oe_ecalls_table[] = {\n" ;
-  List.iter
-    (fun f -> fprintf os "    (oe_ecall_func_t) ecall_%s,\n" f.tf_fdecl.fname)
-    ec.tfunc_decls ;
-  fprintf os "};\n\n" ;
-  fprintf os
-    "size_t __oe_ecalls_table_size = OE_COUNTOF(__oe_ecalls_table);\n\n"
+(** Generate ECALL function table. *)
+let oe_gen_ecall_table (tfs : trusted_func list) =
+  if tfs <> [] then
+    [ "oe_ecall_func_t __oe_ecalls_table[] = {"
+    ; "    "
+      ^ String.concat ",\n    "
+          (List.map
+             (fun f -> sprintf "(oe_ecall_func_t) ecall_%s" f.tf_fdecl.fname)
+             tfs)
+    ; "};"
+    ; ""
+    ; "size_t __oe_ecalls_table_size = OE_COUNTOF(__oe_ecalls_table);" ]
+  else ["/* There were no ecalls. */"]
 
 let gen_fill_marshal_struct (os : out_channel) (fd : func_decl) (args : string)
     =
--- a/tools/oeedger8r/Emitter.ml
+++ b/tools/oeedger8r/Emitter.ml
@@ -491,20 +491,6 @@ let get_cast_from_mem_expr (ptype, decl) =
         sprintf "(const %s) " (get_tystr t)
       else ""
 
-let oe_copy_members_to_enclave (os : out_channel) (fd : func_decl) =
-  let is_primitive ptype =
-    match ptype with
-    | PTPtr (atype, ptr_attr) -> not ptr_attr.pa_chkptr
-    | _ -> true
-  in
-  let gen_copy_member (ptype, decl) =
-    if is_primitive ptype then
-      fprintf os "    enc_args.%s = args.%s;\n" decl.identifier decl.identifier
-  in
-  fprintf os "    /* Copy primitive properties to enc_args */\n" ;
-  List.iter gen_copy_member fd.plist ;
-  fprintf os "\n"
-
 let oe_gen_allocate_buffers (os : out_channel) (fd : func_decl) =
   let gen_allocate_buffer (ptype, decl) =
     match ptype with
--- a/tools/oeedger8r/Emitter.ml
+++ b/tools/oeedger8r/Emitter.ml
@@ -47,16 +47,6 @@ let gen_parm_str (p : pdecl) =
 
 let retval_declr = {identifier= "_retval"; array_dims= []}
 
-let get_ret_tystr (fd : func_decl) = get_tystr fd.rtype
-
-let get_plist_str (fd : func_decl) =
-  if fd.plist = [] then ""
-  else
-    List.fold_left
-      (fun acc pd -> acc ^ ",\n        " ^ gen_parm_str pd)
-      (gen_parm_str (List.hd fd.plist))
-      (List.tl fd.plist)
-
 (** [conv_array_to_ptr] is used to convert Array form into Pointer form.
     {[
       int array[10][20] => [count = 200] int* array
@@ -240,21 +230,25 @@ let oe_get_param_size (ptype, decl, argstruct) =
 (** Generate the prototype for a given function. Optionally add an
     [oe_enclave_t*] first parameter. *)
 let oe_gen_prototype (fd : func_decl) =
-  let params_str = if fd.plist = [] then "void" else get_plist_str fd in
-  sprintf "%s %s(%s)" (get_ret_tystr fd) fd.fname params_str
+  let get_plist_str (plist : pdecl list) =
+    if List.length plist = 0 then "void"
+    else
+      (if List.length plist = 1 then "" else "\n    ")
+      ^ String.concat ",\n    " (List.map gen_parm_str plist)
+  in
+  sprintf "%s %s(%s)" (get_tystr fd.rtype) fd.fname (get_plist_str fd.plist)
 
 let oe_gen_wrapper_prototype (fd : func_decl) (is_ecall : bool) =
-  let plist_str = get_plist_str fd in
-  let retval_str =
-    if fd.rtype = Void then "" else sprintf "%s* _retval" (get_ret_tystr fd)
-  in
+  let args = List.map gen_parm_str fd.plist in
   let args =
-    if is_ecall then ["oe_enclave_t* enclave"; retval_str; plist_str]
-    else [retval_str; plist_str]
+    if fd.rtype <> Void then sprintf "%s* _retval" (get_tystr fd.rtype) :: args
+    else args
   in
-  let args = List.filter (fun s -> s <> "") args in
-  sprintf "oe_result_t %s(\n        %s)" fd.fname
-    (String.concat ",\n        " args)
+  let args = if is_ecall then "oe_enclave_t* enclave" :: args else args in
+  sprintf "oe_result_t %s(%s)" fd.fname
+    ( if List.length args = 0 then ""
+    else if List.length args = 1 then List.hd args
+    else "\n    " ^ String.concat ",\n    " args )
 
 (** Emit [struct], [union], or [enum]. *)
 let emit_composite_type =
--- a/tools/oeedger8r/Emitter.ml
+++ b/tools/oeedger8r/Emitter.ml
@@ -820,6 +820,11 @@ let oe_gen_ocall_function (uf : untrusted_func) =
   ; "}"
   ; "" ]
 
+(** Generate all OCALL functions, if any. *)
+let oe_gen_ocall_functions (ufs : untrusted_func list) =
+  if ufs <> [] then List.flatten (List.map oe_gen_ocall_function ufs)
+  else ["/* There were no ocalls. */"]
+
 (** Generate the OCALL function table. Note that this is always present. *)
 let oe_gen_ocall_table (ufs : untrusted_func list) (name : string) =
   [ sprintf "static oe_ocall_func_t __%s_ocall_function_table[] = {" name
@@ -1054,25 +1059,6 @@ let gen_t_c (ec : enclave_content) (ep : edger8r_params) =
   fprintf os "%s" (String.concat "\n" content) ;
   close_out os
 
-let oe_emit_create_enclave_defn (os : out_channel) (ec : enclave_content) =
-  fprintf os "oe_result_t oe_create_%s_enclave(const char* path,\n"
-    ec.enclave_name ;
-  fprintf os "                                 oe_enclave_type_t type,\n" ;
-  fprintf os "                                 uint32_t flags,\n" ;
-  fprintf os "                                 const void* config,\n" ;
-  fprintf os "                                 uint32_t config_size,\n" ;
-  fprintf os "                                 oe_enclave_t** enclave)\n" ;
-  fprintf os "{\n" ;
-  fprintf os "    return oe_create_enclave(path,\n" ;
-  fprintf os "               type,\n" ;
-  fprintf os "               flags,\n" ;
-  fprintf os "               config,\n" ;
-  fprintf os "               config_size,\n" ;
-  fprintf os "               __%s_ocall_function_table,\n" ec.enclave_name ;
-  fprintf os "               %d,\n" (List.length ec.ufunc_decls) ;
-  fprintf os "               enclave);\n" ;
-  fprintf os "}\n\n"
-
 let gen_u_h (ec : enclave_content) (ep : edger8r_params) =
   let oe_gen_tfunc_wrapper_prototypes (tfs : trusted_func list) =
     if tfs <> [] then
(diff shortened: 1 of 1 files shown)


andschwa:
This will make implementing deep-copy (and other edger8r changes) _worlds_ easier (for me).

andschwa:
If you care to, take a look at the generated code of e.g. the edger8r tests before and after this chain. It's so much better.

anakrish:
Snippet from build/tests/echo/enc/echo_args.h
```c
/* No errno support. */

/**** User includes. ****/
/* There were no user includes. */

/**** User defined types in EDL. ****/
/* There were no user defined types. */
/**** ECALL marshalling structs. ****/
``` 

A newline before ECALL marshalling structs would make this more readable.
Not clear what 'No errno support' means.

anakrish:
```c
/**** Trusted function IDs ****/
enum
{
    fcn_id_enc_echo = 0,
    fcn_id_trusted_call_id_max = OE_ENUM_MAX
};

/**** Untrusted function IDs. ****/
enum
{
    fcn_id_host_echo = 0,
    fcn_id_untrusted_call_max = OE_ENUM_MAX
};
```

Does refactoring make it easy to prefix the enums with the edl file name? Making enums globally unique allows the host to include multiple edl _u.h files easily.

andschwa:
> A newline before ECALL marshalling structs would make this more readable.

Ah, thanks! Must have missed that.

> Not clear what 'No errno support' means.

As in, `propogate_errno` was not true. I can make this comment more explicit.

anakrish:
>> The generated files now (mostly) pass format-code

Though not critical, I can see some code that is not format-code compatible.
Inconsistent spacing, comments longer than 80 chars etc.
```c
 host_echo_args_t _args, *_pargs_in = NULL, *_pargs_out=NULL;
```

andschwa:
> ```c
> ```c
>  host_echo_args_t _args, *_pargs_in = NULL, *_pargs_out=NULL;
> ```
> 
> 
>       
> ```

Are you sure you generated? I specifically remember fixing `*_pargs_out=NULL`.

Also, I don't believe line length limits are worth fixing:

> with the exception of whitespace on empty lines and extra-long lines

The other notable exception is `if (foo) blah` on one line, where one line is better even though it's too long, because it's repetitive, generated code.

anakrish:
Yes you are right RE _pargs_out. I had gotten my master/PR tabs mixed up.

anakrish:
There seems to be some indentation issue in some cases:

```c
    /* Compute input buffer size. Include in and in-out parameters. */
    OE_ADD_SIZE(_input_buffer_size, sizeof(host_echo_args_t));
    if (in) OE_ADD_SIZE(_input_buffer_size, _args.in_len * sizeof(char));
if (str1) OE_ADD_SIZE(_input_buffer_size, _args.str1_len * sizeof(char));
if (str3) OE_ADD_SIZE(_input_buffer_size, sizeof(char[100]));
```

andschwa:
> ```c
> /**** Trusted function IDs ****/
> enum
> {
>     fcn_id_enc_echo = 0,
>     fcn_id_trusted_call_id_max = OE_ENUM_MAX
> };
> 
> /**** Untrusted function IDs. ****/
> enum
> {
>     fcn_id_host_echo = 0,
>     fcn_id_untrusted_call_max = OE_ENUM_MAX
> };
> ```
> 
> Does refactoring make it easy to prefix the enums with the edl file name? Making enums globally unique allows the host to include multiple edl _u.h files easily.

Do you mean something like (for `foo.edl`):

```
enum
{
    foo_fcn_id_host_echo = 0,
    foo_fcn_id_untrusted_call_max = OE_ENUM_MAX
};
```

Or does the `enum` need a name too?

anakrish:
> 
> Do you mean something like (for `foo.edl`):
> 
> ```
> enum
> {
>     foo_fcn_id_host_echo = 0,
>     foo_fcn_id_untrusted_call_max = OE_ENUM_MAX
> };
> ```

Yes. This would be ideal.

anakrish:
We could space the following a bit better
```c
    /* Unmarshal return value and out, in-out parameters. */
    *_retval = _pargs_out->_retval;
    OE_READ_OUT_PARAM(out, (size_t)(sizeof(char[100])));
    /* Propagate errno. */
    /* Errno not enabled. */
    _result = OE_OK;
```

andschwa:
> ```c
> OE_ADD_SIZE
> ```

Hunting for this...have a feeling I broke it in my last commit.

andschwa:
> There seems to be some indentation issue in some cases:
> 
> ```c
>     /* Compute input buffer size. Include in and in-out parameters. */
>     OE_ADD_SIZE(_input_buffer_size, sizeof(host_echo_args_t));
>     if (in) OE_ADD_SIZE(_input_buffer_size, _args.in_len * sizeof(char));
> if (str1) OE_ADD_SIZE(_input_buffer_size, _args.str1_len * sizeof(char));
> if (str3) OE_ADD_SIZE(_input_buffer_size, sizeof(char[100]));
> ```

Fixed. Thanks for doing all that scrolling. This broke when I refactored this into `oe_compute_input_buffer_size`; since now I'm returning a single string back to a list that has its indentation applied when it's concatted. Still looking for a better way to handle that, but it's fixed with a note now.

andschwa:
> We could space the following a bit better
> 
> ```c
>     /* Unmarshal return value and out, in-out parameters. */
>     *_retval = _pargs_out->_retval;
>     OE_READ_OUT_PARAM(out, (size_t)(sizeof(char[100])));
>     /* Propagate errno. */
>     /* Errno not enabled. */
>     _result = OE_OK;
> ```

Like:

```c
    /* Unmarshal return value and out, in-out parameters. */
    *_retval = _pargs_out->_retval;
    OE_READ_OUT_PARAM(out, (size_t)(sizeof(char[100])));

    /* Propagate errno. */
    /* Errno not enabled. */

    _result = OE_OK;
```
?

anakrish:
> Like:
> 
> ```c
>     /* Unmarshal return value and out, in-out parameters. */
>     *_retval = _pargs_out->_retval;
>     OE_READ_OUT_PARAM(out, (size_t)(sizeof(char[100])));
> 
>     /* Propagate errno. */
>     /* Errno not enabled. */
> 
>     _result = OE_OK;
> ```
> 
> ?
Looks good.

anakrish:
I noticed that the generated files no longer have a terminating newline character.

ISO C requires a newline character at the end of the file. 
If you put "int foo() { return 1; }" in a file foo.c and compile it using clang -c -Wpedantic foo.c, it will emit a warning:
```bash
$ clang -c   -Wpedantic foo.c
foo.c:4:2: warning: no newline at end of file [-Wnewline-eof]
}
 ^
1 warning generated.
```

Any reason why we want to make the generated code non ISO C compatible?
@CodeMonkeyLeet If I remember correctly, at one point in time we had an automatic check to prevent check-in of non newline-terminated files.

andschwa:
> I noticed that the generated files no longer have a terminating newline character.
> 
> ISO C requires a newline character at the end of the file.
> If you put "int foo() { return 1; }" in a file foo.c and compile it using clang -c -Wpedantic foo.c, it will emit a warning:
> 
> ```shell
> $ clang -c   -Wpedantic foo.c
> foo.c:4:2: warning: no newline at end of file [-Wnewline-eof]
> }
>  ^
> 1 warning generated.
> ```
> 
> Any reason why we want to make the generated code non ISO C compatible?
> @CodeMonkeyLeet If I remember correctly, at one point in time we had an automatic check to prevent check-in of non newline-terminated files.

Oh I forgot about that. Yes, I need to make sure they all end with newlines. Thanks!

andschwa:
> > Do you mean something like (for `foo.edl`):
> > ```
> > enum
> > {
> >     foo_fcn_id_host_echo = 0,
> >     foo_fcn_id_untrusted_call_max = OE_ENUM_MAX
> > };
> > ```
> 
> Yes. This would be ideal.

Yeah, I'd say pretty easy considering I have a commit implementing this and passing tests right now. That said, I'd like to clean it up to pass through a curried version of get_function_id instead of passing through the enclave name itself, seems like it'd be cleaner.

andschwa:
> I noticed that the generated files no longer have a terminating newline character.
> 
> ISO C requires a newline character at the end of the file.
> If you put "int foo() { return 1; }" in a file foo.c and compile it using clang -c -Wpedantic foo.c, it will emit a warning:
> 
> ```shell
> $ clang -c   -Wpedantic foo.c
> foo.c:4:2: warning: no newline at end of file [-Wnewline-eof]
> }
>  ^
> 1 warning generated.
> ```
> 
> Any reason why we want to make the generated code non ISO C compatible?
> @CodeMonkeyLeet If I remember correctly, at one point in time we had an automatic check to prevent check-in of non newline-terminated files.

Fixed, thanks for the catch. It was the C files I'd forgotten to newline-terminate.

andschwa:
> We could space the following a bit better
> 
> ```c
>     /* Unmarshal return value and out, in-out parameters. */
>     *_retval = _pargs_out->_retval;
>     OE_READ_OUT_PARAM(out, (size_t)(sizeof(char[100])));
>     /* Propagate errno. */
>     /* Errno not enabled. */
>     _result = OE_OK;
> ```

Fixed.

andschwa:
> A newline before ECALL marshalling structs would make this more readable.

Fixed (I actually have a TODO to make all these newlines a bit cleaner here, but worked around it for now.)

> Not clear what 'No errno support' means.

Fixed. I clarified all the messages; let me know what you think.

andschwa:
@gupta-ak I don't know how you did any work in the edger8r before this 😅

anakrish:
Couple more comments re generated code. After this I am going to look at the Ocaml implementation:

1. The include guard for _args.h needs to have the EDGER8R_ prefix like _t.h and _u.h.
2. Errono propagation not enabled makes sense on a per function basis and adds value when reading each function's generated implementation. However, it feels out of place in _args.h

tools/oeedger8r/Emitter.ml
```diff
@@ -217,102 +231,143 @@ let oe_get_param_size (ptype, decl, argstruct) =
...
+    else if List.length args = 1 then List.hd args
+    else "\n    " ^ String.concat ",\n    " args )
 
 (** Emit [struct], [union], or [enum]. *)
-let emit_composite_type (os : out_channel) = function
-  | StructDef s -> emit_struct_or_union os s false
-  | UnionDef u -> emit_struct_or_union os u true
-  | EnumDef e -> emit_enum os e
+let emit_composite_type =
+  let emit_struct_or_union (s : struct_def) (union : bool) =
```
anakrish: I like the way this looks.

tools/oeedger8r/Emitter.ml
```diff
@@ -217,102 +231,143 @@ let oe_get_param_size (ptype, decl, argstruct) =
...
+    [ sprintf "typedef enum %s" e.enname
+    ; "{"
+    ; String.concat ",\n"
+        (List.map
+           (fun (name, value) ->
+             sprintf "    %s%s" name
+               ( match value with
+               | EnumVal (AString s) -> " = " ^ s
+               | EnumVal (ANumber n) -> " = " ^ string_of_int n
+               | EnumValNone -> "" ) )
```
anakrish: Can you take a follow up item to see if we need to add a final enumeration member whose value equals to OE_ENUM_MAX?

tools/oeedger8r/Emitter.ml
```diff
@@ -325,153 +380,139 @@ let oe_gen_args_header (ec : enclave_content) (dir : string) =
...
+        sprintf "if (%s) OE_ADD_SIZE(_output_buffer_size, %s);" decl.identifier
+          size )
+      (List.filter (fun p -> is_out_ptr p || is_inout_ptr p) plist)
+  in
+  (* Note that the indentation for the first line is applied by the
+     parent function. *)
+  if params <> [] then String.concat "\n    " params
+  else "/* There were no out nor in-out parameters. */"
+
+let oe_serialize_buffer_inputs (plist : pdecl list) =
```
anakrish: It feels like these duplication in these 3 functions can be avoided.
In future.

tools/oeedger8r/Emitter.ml
```diff
@@ -325,153 +380,139 @@ let oe_gen_args_header (ec : enclave_content) (dir : string) =
...
+        ""
+      ; "/* Setup output arg struct pointer. */"
+      ; sprintf "_pargs_out = (%s_args_t*)_output_buffer;" fd.fname
+      ; "OE_ADD_SIZE(_output_buffer_offset, sizeof(*_pargs_out));"
+      ; ""
+      ; "/* Check if the call succeeded. */"
+      ; "if ((_result = _pargs_out->_result) != OE_OK)"
+      ; "    goto done;"
+      ; ""
+      ; "/* Currently exactly _output_buffer_size bytes must be written. */"
```
anakrish: In the future we can tweaks this to support variable length outputs.

tools/oeedger8r/Emitter.ml
```diff
@@ -483,435 +524,371 @@ let get_cast_from_mem_expr (ptype, decl) =
...
   in
-  let call_str = ret_str ^ fd.fname ^ params_str in
-  fprintf os "    /* Call user function */\n" ;
-  fprintf os "    %s;\n" call_str
+  "    "
+  ^ String.concat "\n    "
+      [ "/* Set out and in-out pointers. */"
+      ; "/* In-out parameters are copied to output buffer. */"
+      ; ( if params <> [] then String.concat "\n    " params
+        else "/* There were no out nor in-out parameters. */" ) ]
```
anakrish: Can you double check the grammar here? 
We were taught that it was "neither..nor"

tools/oeedger8r/Emitter.ml
```diff
@@ -483,435 +524,371 @@ let get_cast_from_mem_expr (ptype, decl) =
...
+             (if is_wstr_ptr (ptype, decl) then "_WIDE" else "")
+             decl.identifier decl.identifier )
+         (List.filter
+            (fun p ->
+              (is_str_ptr p || is_wstr_ptr p) && (is_in_ptr p || is_inout_ptr p)
+              )
+            fd.plist)
+     in
+     if params <> [] then String.concat "\n" params
+     else "    /* There were no in nor in-out string parameters. */")
```
anakrish: Double-check 'nor' usage.

tools/oeedger8r/Emitter.ml
```diff
@@ -110,13 +101,34 @@ let mk_ms_member_decl (pt : parameter_type) (declr : declarator)
...
+  | PTVal _ -> false
+  | PTPtr (_, a) -> a.pa_chkptr && a.pa_direction = PtrOut
+
+let is_inout_ptr (ptype, _) =
+  match ptype with
+  | PTVal _ -> false
+  | PTPtr (_, a) -> a.pa_chkptr && a.pa_direction = PtrInOut
+
+let is_str_ptr (ptype, _) =
+  match ptype with PTVal _ -> false | PTPtr (_, a) -> a.pa_isstr
```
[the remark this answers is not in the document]
andschwa: I'm just using `ocamlformat` with the defaults; we can adjust it if you really want, but it seemed easier to take the decision out of our hands 😄

tools/oeedger8r/Emitter.ml
```diff
@@ -149,21 +162,22 @@ let oe_gen_marshal_struct_impl (fd : func_decl) (errno : string)
...
+  let struct_name = fd.fname ^ "_args_t" in
   match fd.rtype with
   | Void -> oe_mk_struct_decl member_list_str struct_name
   | _ ->
       let rv_str = mk_ms_member_decl (PTVal fd.rtype) retval_declr isecall in
       oe_mk_struct_decl (rv_str ^ member_list_str) struct_name
 
 let oe_gen_ecall_marshal_struct (tf : trusted_func) =
-  oe_gen_marshal_struct_impl tf.tf_fdecl "" true
+  oe_gen_marshal_struct_impl tf.tf_fdecl "" true ^ "\n"
```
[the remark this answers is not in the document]
andschwa: `ocamlformat`...

tools/oeedger8r/Emitter.ml
```diff
@@ -217,102 +231,143 @@ let oe_get_param_size (ptype, decl, argstruct) =
 (** Generate the prototype for a given function. Optionally add an
     [oe_enclave_t*] first parameter. *)
 let oe_gen_prototype (fd : func_decl) =
-  let params_str =
-    if List.length fd.plist = 0 then "void" else get_plist_str fd
+  let get_plist_str (plist : pdecl list) =
+    if List.length plist = 0 then "void"
```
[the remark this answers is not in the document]
andschwa: You know, I noticed some funny stuff with checking `Void`, I'd punted it for a later fix.

andschwa:
> * The include guard for _args.h needs to have the EDGER8R_ prefix like _t.h and _u.h.

I checked master and this is an existing bug, fixed now though.

tools/oeedger8r/Emitter.ml
andschwa: Filed #1767.

tools/oeedger8r/Emitter.ml
andschwa: Me too, minus the use of bool for union/struct...but apparently types aren't that strongly represented in OCaml. There's a lot of code I found could be simplified if I could simply pass a type constructor _as_ an argument.

tools/oeedger8r/Emitter.ml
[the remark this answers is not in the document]
andschwa: (Literally just tried putting parens around it, on save my hook that runs `ocamlformat` removed them 😅)

tools/oeedger8r/Emitter.ml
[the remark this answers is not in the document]
andschwa: Oh, no, this is not equivalent to `fd.rtype = Void`, as it is not the return type. This is checking the function takes zero arguments (plist), and if so, generates the correct *C* (not C++) code to say `foo(void)` instead of `foo()`.

tools/oeedger8r/Emitter.ml
andschwa: I know, I felt that way while I was writing them, but just enough is different that it makes it difficult to deduplicate further. Their filter predicates are different (so their comments are different) and what they're printing is different. I opted for readability, but certainly it can be improved.

tools/oeedger8r/Emitter.ml
andschwa: That'd be cool.

tools/oeedger8r/Emitter.ml
andschwa: Yeah it [seems](https://www.wikihow.com/Use-nor) to [be](https://thewritepractice.com/how-to-use-either-neither-or-and-nor-correctly/) correct. It's not as common, but you can use 'nor' without 'neither' if it's paired with another negative (in this case, 'no').

andschwa:
> 2\. Errono propagation not enabled makes sense on a per function basis and adds value when reading each function's generated implementation. However, it feels out of place in _args.h

It's in place of `#include <errno.h>`. Let me include that string in the comment itself to make it more obvious.

andschwa:
@anakrish Do you want to include #1765 in this? I didn't get a chance to add an explicit test for it, but it is also doesn't break anything so we certainly could chunk it since this is under development anyway.

andschwa:
bors try

anakrish:
> @anakrish Do you want to include #1765 in this? I didn't get a chance to add an explicit test for it, but it is also doesn't break anything so we certainly could chunk it since this is under development anyway.

Let us hold off on that one.  This PR is a net positive. Bringing in #1765 without a test feels like we are taking away from this PR. There is no immediate urgency for #1765

anakrish:
> 2. Errono propagation not enabled makes sense on a per function basis and adds value when reading each function's generated implementation. However, it feels out of place in _args.h
> 
> It's in place of #include <errno.h>. Let me include that string in the comment itself to make it more obvious.

Ah ok. It never occurred to me that the message was taking the place of the errno header.

andschwa:
Noting that my TODOs in the PR description are future TODOs and don't block this, merging!

bors r+

Thanks for the reviews Anand!

PR #1765, opened 2019-05-07
Make function IDs globally unique

andschwa:
This makes the enums globally unique which allows the host to include multiple generated untrusted headers more easily.

It also does a small refactor of functions which did not need to be globally scoped.


andschwa:
@anakrish We should probably add an example / test with this.

PR #1769, opened 2019-05-08
[host/cmake] _FORTIFY_SOURCE should not be used on debug builds

aserdean:
As per: https://github.com/microsoft/openenclave/issues/1725 , `_FORTIFY_SOURCE`
should be used only on the targets: release, relwithdebinfo, minsizerel.

Fixes: https://github.com/microsoft/openenclave/issues/1725

### DIFF
--- a/host/CMakeLists.txt
+++ b/host/CMakeLists.txt
@@ -176,7 +176,7 @@ if (UNIX)
      PUBLIC -fstack-protector-strong)
   target_compile_definitions(oehost
      PRIVATE _GNU_SOURCE
-     PUBLIC _FORTIFY_SOURCE=2)
+     PUBLIC $<$<NOT:$<CONFIG:debug>>:_FORTIFY_SOURCE=2>)
 endif ()
 
 if (CMAKE_C_COMPILER_ID MATCHES GNU)


host/CMakeLists.txt
```diff
@@ -174,9 +174,15 @@ if (UNIX)
   target_compile_options(oehost
      PRIVATE -Wno-attributes -Wmissing-prototypes -fPIC ${PLATFORM_FLAGS}
      PUBLIC -fstack-protector-strong)
-  target_compile_definitions(oehost
-     PRIVATE _GNU_SOURCE
-     PUBLIC _FORTIFY_SOURCE=2)
+  string(TOUPPER "${CMAKE_BUILD_TYPE}" CMAKE_BUILD_TYPE_SUFFIX)
```
andschwa: This is really confusing to read. Try changing it to:

```
target_compile_definitions(oehost
    PRIVATE _GNU_SOURCE
    PUBLIC $<$<CONFIG:debug>:_FORTIFY_SOURCE=2)>)
```

The `$<CONFIG:cfg>` generator expression is case insensitive, so it should be used wherever possible instead of the hack with `TOUPPER`.

host/CMakeLists.txt
```diff
@@ -176,7 +176,7 @@ if (UNIX)
      PUBLIC -fstack-protector-strong)
   target_compile_definitions(oehost
      PRIVATE _GNU_SOURCE
-     PUBLIC _FORTIFY_SOURCE=2)
+     PUBLIC $<$<NOT:$<CONFIG:debug>>:_FORTIFY_SOURCE=2>))
```
andschwa: Heh, I forgot the `not`. Thanks!

host/CMakeLists.txt
aserdean: @andschwa thanks for the suggestion.

I updated the commit, can you take please another look?

PR #1771, opened 2019-05-08
 Prefix function IDs with enclave name

andschwa:
This makes the enums globally unique which allows the host to include multiple generated untrusted headers more easily. Note the addition of the enclave `edl_other_enc` which can be loaded via the edger8r tests host, meaning it includes both `all_u.h` _and_ `other_u.h`.

### DIFF
--- a/tests/oeedger8r/CMakeLists.txt
+++ b/tests/oeedger8r/CMakeLists.txt
@@ -8,5 +8,6 @@ if (BUILD_ENCLAVES)
 endif()
 
 add_enclave_test(tests/oeedger8r edl_host edl_enc)
+add_enclave_test(tests/oeedger8r_other edl_host edl_other_enc)
 
 add_subdirectory(behavior)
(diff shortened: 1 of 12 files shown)


andschwa:
@soccerGB Please test this, it should support your scenario now.

This replaces @anakrish's [WIP](https://github.com/Microsoft/openenclave/compare/anakrish/fix/name-conflict). @anakrish, I noticed I had to rename significantly fewer things (only the function ID enums) to make this work, yet your branch prefixed all the args structs too; why was that?

anakrish:
> @soccerGB Please test this, it should support your scenario now.
> This replaces @anakrish's WIP. @anakrish, I noticed I had to rename significantly fewer things (only the function ID enums) to make this work, yet your branch prefixed all the args structs too; why was that?

Most likely to scope all identifiers from an EDL file into a "namespace" by consistently prefixing them.
However, I don't think it is a necessity.

PR #1773, opened 2019-05-09
Move crypto APIs called from within enclave code into internal/crypto directory

radhikaj:
Putting all oe crypto functions into oecrypto.h and placing that in oecrypto directory. This is the first step towards providing developers guidance to develop their own implementation of oecrypto built over their choice of crypto library(openssl, mbedtls etc).


radhikaj:
Note that there is a single header. This is the plan as per the oecrypto discussion and was summarized as such.

anakrish:
@radhikaj 
Will the new header be published?
How can a end user write their own crypto implementations and hook it up to the OE SDK?

anitagov:
I am not sure of the advantage of putting all the crypto header files in one file. But the cons is that it will dramatically increase compile and link time. Also a small change in one of the header files will cause everything to get rebuilt. So, wondering why this choice was made.

radhikaj:
> I am not sure of the advantage of putting all the crypto header files in one file. But the cons is that it will dramatically increase compile and link time. Also a small change in one of the header files will cause everything to get rebuilt. So, wondering why this choice was made.
@anitagov I am following the document that was prepared. It said to make a single header. If you think this is not the right way, please comment on document. Until the doc is changed, that is the plan.

radhikaj:
> @radhikaj
> Will the new header be published?
> How can a end user write their own crypto implementations and hook it up to the OE SDK?

adfsadfasdf

radhikaj:
> @radhikaj
> Will the new header be published?
> How can a end user write their own crypto implementations and hook it up to the OE SDK?

@anakrish This header will be published to github(openenclave organization) along with guidance and a sample. The guidance will be provided in conjunction with that sample.

anakrish:
@radhikaj 
To save yourself rework, in case you haven't already done so, can you follow up with @CodeMonkeyLeet about a unified header? From what I remember he had some concerns. 
Note: I don't have any issue either way; but would simply like to avoid rework.

I agree with @johnkord that there won't be significant changes in compilation times either way.

anakrish:
> > @radhikaj
> > Will the new header be published?
> > How can a end user write their own crypto implementations and hook it up to the OE SDK?
> 
> @anakrish This header will be published to github(openenclave organization) along with guidance and a sample. The guidance will be provided in conjunction with that sample.

@radhikaj Sounds good to me. Maybe we want to change the location of the header? Internal in your branch has both stuff we want an advanced user to change (oecrypto) as well as stuff that we don't expence any one to change (say atomic.h)

radhikaj:
> > > @radhikaj
> > > Will the new header be published?
> > > How can a end user write their own crypto implementations and hook it up to the OE SDK?
> > 
> > 
> > @anakrish This header will be published to github(openenclave organization) along with guidance and a sample. The guidance will be provided in conjunction with that sample.
> 
> @radhikaj Sounds good to me. Maybe we want to change the location of the header? Internal in your branch has both stuff we want an advanced user to change (oecrypto) as well as stuff that we don't expence any one to change (say atomic.h)

@anakrish Thanks. What would be a good place to put this header? We dont want it to be published with OESDK.

anakrish:
Just moving oecrypto to outside of internal itself will separate it from other internal (currently everything used internally subject to change) stuff.

CodeMonkeyLeet:
@radhikaj I haven't seen a compelling reason for consolidating all our crypto defs into a single header other than "the doc says so". I've also expressed in this in said doc. This seems counterproductive for a number of reasons:

- It reduces readability. It's much easier to navigate a set of well-partitioned header files than search through a single massive header. It also makes clearer at a glance the set of functionality that's covered by looking at the set of headers well-named headers rather than a single one.

- It makes all source files pull in more typedefs/definitions/structs than likely needed, which makes it harder to maintain a separation of concerns and componentization. (e.g. does everything in core want to include header declarations for TLS methods? What happens if we decide to factor that out as a separate optional lib?) 

Moving the set of crypto headers into their own folder is probably sufficient for facilitating further documentation around creating alternative implementations of `liboecrypto`. 

A couple of other issues with the simple refactor: 

- There's an existing implicit assumption that the crypto headers under internal are uniformly shared between the host and enclave implementations. If oecrypto.h is expected to represent what an enclave crypto library needs to implement, then it should be scoped as such.

- On the same note, there are a number of methods are only used in tests and are not actually used in the enclave runtime. (e.g. `oe_rsa_generate_key_pair()`) Some of these I'm going to prune out given that they cannot be supported with the existing semantics on Windows host.

Also regarding this:

> @anakrish This header will be published to github(openenclave organization) along with guidance and a sample. The guidance will be provided in conjunction with that sample.

To clarify: this change is going into the OE SDK master branch, but I would recommend that anything (samples/documentation) dealing with non-public headers (e.g. oelibc replacement, crypto lib replacement) be published under separate GitHub projects under the Open Enclave org.

radhikaj:
> @radhikaj I haven't seen a compelling reason for consolidating all our crypto defs into a single header other than "the doc says so". I've also expressed in this in said doc. This seems counterproductive for a number of reasons:
> 
> * It reduces readability. It's much easier to navigate a set of well-partitioned header files than search through a single massive header. It also makes clearer at a glance the set of functionality that's covered by looking at the set of headers well-named headers rather than a single one.
> * It makes all source files pull in more typedefs/definitions/structs than likely needed, which makes it harder to maintain a separation of concerns and componentization. (e.g. does everything in core want to include header declarations for TLS methods? What happens if we decide to factor that out as a separate optional lib?)
> 
> Moving the set of crypto headers into their own folder is probably sufficient for facilitating further documentation around creating alternative implementations of `liboecrypto`.
> 
> A couple of other issues with the simple refactor:
> 
> * There's an existing implicit assumption that the crypto headers under internal are uniformly shared between the host and enclave implementations. If oecrypto.h is expected to represent what an enclave crypto library needs to implement, then it should be scoped as such.
> * On the same note, there are a number of methods are only used in tests and are not actually used in the enclave runtime. (e.g. `oe_rsa_generate_key_pair()`) Some of these I'm going to prune out given that they cannot be supported with the existing semantics on Windows host.
> 
> Also regarding this:
> 
> > @anakrish This header will be published to github(openenclave organization) along with guidance and a sample. The guidance will be provided in conjunction with that sample.
> 
> To clarify: this change is going into the OE SDK master branch, but I would recommend that anything (samples/documentation) dealing with non-public headers (e.g. oelibc replacement, crypto lib replacement) be published under separate GitHub projects under the Open Enclave org.

@CodeMonkeyLeet From my point of view, having a single header makes it easy for a developer to figure out which methods would need to be implemented and I believe that is why it was put in the doc. I see that making TLS libs optional would be difficult with a single header. 
I would like to keep the same approach as we had before, with APIs common to enclave and host.  If there are APIs that are specific to enclaves, we can move them into oecrypto/oecrypto_enclave.h . Please let me know your thoughts on this. 
Since this feature has seen lots of inputs and design iterations, if we could agree on the specifics in the doc and then execute, it would be great.

mikbras:
> @radhikaj I haven't seen a compelling reason for consolidating all our crypto defs into a single header other than "the doc says so". I've also expressed in this in said doc. This seems counterproductive for a number of reasons:
> 
> It reduces readability. It's much easier to navigate a set of well-partitioned header files than search through a single massive header. It also makes clearer at a glance the set of functionality that's covered by looking at the set of headers well-named headers rather than a single one.
> It makes all source files pull in more typedefs/definitions/structs than likely needed, which makes it harder to maintain a separation of concerns and componentization. (e.g. does everything in core want to include header declarations for TLS methods? What happens if we decide to factor that out as a separate optional lib?)
> 
> Moving the set of crypto headers into their own folder is probably sufficient for facilitating further documentation around creating alternative implementations of liboecrypto.
> A couple of other issues with the simple refactor:
> 
> There's an existing implicit assumption that the crypto headers under internal are uniformly shared between the host and enclave implementations. If oecrypto.h is expected to represent what an enclave crypto library needs to implement, then it should be scoped as such.
> On the same note, there are a number of methods are only used in tests and are not actually used in the enclave runtime. (e.g. oe_rsa_generate_key_pair()) Some of these I'm going to prune out given that they cannot be supported with the existing semantics on Windows host.
> 
> Also regarding this:
> 
> @anakrish This header will be published to github(openenclave organization) along with guidance and a sample. The guidance will be provided in conjunction with that sample.
> 
> To clarify: this change is going into the OE SDK master branch, but I would recommend that anything (samples/documentation) dealing with non-public headers (e.g. oelibc replacement, crypto lib replacement) be published under separate GitHub projects under the Open Enclave org.
> 
> @CodeMonkeyLeet From my point of view, having a single header makes it easy for a developer to figure out which methods would need to be implemented and I believe that is why it was put in the doc. I see that making TLS libs optional would be difficult with a single header.
> I would like to keep the same approach as we had before, with APIs common to enclave and host.  If there are APIs that are specific to enclaves, we can move them into oecrypto/oecrypto_enclave.h . Please let me know your thoughts on this.
> Since this feature has seen lots of inputs and design iterations, if we could agree on the specifics in the doc and then execute, it would be great.

Yes, my preference too is to maintain the separate headers but move them into single directory called oecrypto.

mikbras:
Actually, I think we should just use "crypto" like this:
```
#include <openenclave/crypto/cert.h>
```

The "openenclave" is qualification enough I think.

common/asn1.h
```diff
@@ -4,7 +4,7 @@
 #ifndef _OE_COMMON_ASN1_H
 #define _OE_COMMON_ASN1_H
 
-#include <openenclave/internal/asn1.h>
+#include <openenclave/internal/oecrypto/oecrypto.h>
```
mikbras: I guess I would stick with the original header names and use:

```
#include <openenclave/internal/asn1.h>
```

Maybe eventually we would publish and move up to:

```
#include <openenclave/crypto/asn1.h>
```

The use of the "oe" prefix seems redundant to me.

radhikaj:
Per consensus, I have moved the crypto headers into a crypto directory. Further splitting between enclave and host as well as removing unnecessary APIs will be done in future iterations.

gupta-ak:
#1782 is in so feel free to debase and merge this one in @radhikaj

radhikaj:
Per discussion
1.	There are APIs that are specific to enclave. These headers will live in internal/crypto. There are some files that have a mix of host only APIs and enclave only APIs. These would need to be split.
2.	There are APIs that are common to host and enclave. These headers will be in internal/crypto
3.	There are APIs that are specific to host. These will live in internal.
4.     There are APIs that are called only from tests. These will live in internal.
This ensures that anyone who wants to replace oecrypto will only need to implement APIs in internal/crypto

There are APIs in files such kdf.c that are not mbedtls or openssl dependent. We will provide these files with our github project demonstrating how to replace oecrypto. This way the developer would not need to rewrite the code in kdf.c .

radhikaj:
[cyrpto APIs.xlsx](https://github.com/microsoft/openenclave/files/3238810/cyrpto.APIs.xlsx)

include/openenclave/internal/crypto/cert.h
```diff
@@ -0,0 +1,380 @@
...
+} oe_cert_t;
+
+typedef struct _oe_cert_chain
+{
+    /* Internal private implementation */
+    uint64_t impl[4];
+} oe_cert_chain_t;
+
+/**
+<<<<<<< e801377375205f7f45208fdfa35bee794f28da94
```
radhikaj: Get rid of hash

radhikaj:
> Overall, looks good.
> 
> The `rdrand.h` doesn't belong in the crypto API. It's implemented in oecore and is used for oecore's platform entropy function `oe_get_entropy` for x86_64. A crypto library, like mbedtls will use the function to seed their random number generator, so it's a lower level / syscall-like function.

Thanks Akash. I have moved rdrand.h out of crypto.

include/openenclave/internal/crypto/keys.h
```diff
@@ -7,7 +7,7 @@
 #include <openenclave/bits/defs.h>
 #include <openenclave/bits/result.h>
 #include <openenclave/bits/types.h>
-#include "sgxtypes.h"
```
CodeMonkeyLeet: keys.h is an SGX internal helper method for `EGETKEY`, not a crypto API implemented by the crypto lib.

Extra credit: we should probably rename keys.h to reflect that e.g. `sgxkeys.h` like the existing `sgxsign.h` or `sgxproperties.h`

tests/crypto/host/main.c
```diff
@@ -3,12 +3,12 @@
 
 #include <assert.h>
 #include <openenclave/internal/cert.h>
+#include <openenclave/internal/crypto/sha.h>
```
CodeMonkeyLeet: Nit: It doesn't look like this file needs anything except the tests.h headers and stdio.h, this could probably be cleaned up.

include/openenclave/internal/crypto/keys.h
radhikaj: Done. Including extra credit.

tests/crypto/host/main.c
radhikaj: Done

radhikaj:
Opened issue #1852  to track crypto refactoring work

radhikaj:
Addressed Simon's concerns and followed consensus during discussion. Can move location of crypto headers around as needed

PR #1774, opened 2019-05-09
Add `propagate_errno` oedger8r test

andschwa:
This adds a very basic test asserting that `propagate_errno` works correctly to propagate the value of errno from an ocall to the enclave. It does not test any sort of complicated threading scenario, which we may want to add support for later.

### DIFF
--- a/tests/oeedger8r/edl/all.edl
+++ b/tests/oeedger8r/edl/all.edl
@@ -7,6 +7,7 @@ enclave  {
     from "array.edl"    import *;
     from "basic.edl"    import *;
     from "enum.edl"     import *;
+    from "errno.edl"    import *;
     from "foreign.edl"  import *;
     from "pointer.edl"  import *;
     from "string.edl"   import *;
--- a/tests/oeedger8r/edl/errno.edl
+++ b/tests/oeedger8r/edl/errno.edl
@@ -1,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+enclave {
+  trusted {
+    public void test_errno_edl_ocalls();
+  };
+
+  untrusted {
+    void ocall_errno() propagate_errno;
+  };
+};
--- a/tests/oeedger8r/enc/CMakeLists.txt
+++ b/tests/oeedger8r/enc/CMakeLists.txt
@@ -26,6 +26,7 @@ add_enclave(TARGET edl_enc CXX
     testarray.cpp
     testbasic.cpp
     testenum.cpp
+    testerrno.cpp
     testforeign.cpp
     testpointer.cpp
     teststring.cpp
--- a/tests/oeedger8r/enc/testerrno.cpp
+++ b/tests/oeedger8r/enc/testerrno.cpp
@@ -1,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "../edltestutils.h"
+
+#include <errno.h>
+#include <openenclave/enclave.h>
+#include <openenclave/internal/tests.h>
+#include <stdio.h>
+#include "all_t.h"
+
+void test_errno_edl_ocalls()
+{
+    errno = 0;
+    OE_TEST(errno == 0);
+
+    OE_TEST(ocall_errno() == OE_OK);
+    OE_TEST(errno == 0x12345678);
+
+    printf("=== test_errno_edl_ocalls passed\n");
+}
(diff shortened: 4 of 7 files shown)


andschwa:
This [sandbox](https://oe-jenkins.eastus.cloudapp.azure.com/blue/organizations/jenkins/OpenEnclave-sandbox/detail/OpenEnclave-sandbox/1208/pipeline/76) run shows it working on both Linux and Windows.

andschwa:
Changed the number from 1 to 0x12345678 to avoid any accidental passes. New [sandbox](https://oe-jenkins.eastus.cloudapp.azure.com/blue/organizations/jenkins/OpenEnclave-sandbox/detail/OpenEnclave-sandbox/1209/pipeline) run.

andschwa:
Note that EDL propagate_errno support does not handle any sort of mapping of error values, it literally just copies the value of an int called `errno` from the host to the enclave.

andschwa:
> It is not clear to me why we we don't want to test the multi-thread scenario which is a fundamental requirement of C errno.

We can, I just hadn't in this yet. This was to test that it works as expected on Windows.

anakrish:
Given that this is a technical debt since Jan, I would prefer that we pay it off in full when we are looking at it. Otherwise, we may move on to other features and not get time to add the full test.

mikbras:
> Given that this is a technical debt since Jan, I would prefer that we pay it off in full when we are looking at it. Otherwise, we may move on to other features and not get time to add the full test.

Thanks for adding Anand.

mikbras:
> Given that this is a technical debt since Jan, I would prefer that we pay it off in full when we are looking at it. Otherwise, we may move on to other features and not get time to add the full test.

In my opinion, this is hardly worth the effort. We already have a test that confirms that errno propagation works. A multi-threaded test for errno propagation would only confirm that the implementation of errno in Windows and Linux is correct, which I am willing to accept on faith. The edge routines contain the following assignment:

```
    pargs_out->_ocall_errno = errno;
```
If this fails when multi-threading is added, then there is a bug in Windows or Linux, not with the edge routines.

anakrish:
@mikbras  

Given that edger8r lives in the boundary of trusted and untrusted code, I would like edger8r to be rigorously locked down with tests.

Adding thorough tests for errno was in my todo list in Jan; but unfortunately, I got pulled into other things.

> > A multi-threaded test for errno propagation would only confirm that the implementation of errno in Windows and Linux is correct, which I am willing to accept on faith. The edge routines contain the following assignment:
> 
> ```
>     pargs_out->_ocall_errno = errno;
> ```
> 

Keep in mind that switchless calls are coming and that scenario involves a many-to-1 bindings between host and enclave threads. Also we would like play with different kinds of code-generation in the future and optimize the generated code.

mikbras:
> @mikbras
> Given that edger8r lives in the boundary of trusted and untrusted code, I would like edger8r to be rigorously locked down with tests.
> Adding thorough tests for errno was in my todo list in Jan; but unfortunately, I got pulled into other things.
> 
> 
> A multi-threaded test for errno propagation would only confirm that the implementation of errno in Windows and Linux is correct, which I am willing to accept on faith. The edge routines contain the following assignment:
> 
>     pargs_out->_ocall_errno = errno;
> 
> 
> Keep in mind that switchless calls are coming and that scenario involves a many-to-1 bindings between host and enclave threads. Also we would like play with different kinds of code-generation in the future and optimize the generated code.

anakrish:
This conversation got me thinking and let to uncovering an interesting scenario.
What should the initial value of errno be in an enclave thread? Should it be 0 or should it be the value of errno in the host prior to making the ecall?

I have captured this and other thoughts in #1775.
To unblock folks I am accepting this PR.

mikbras:
> @mikbras
> Given that edger8r lives in the boundary of trusted and untrusted code, I would like edger8r to be rigorously locked down with tests.
> Adding thorough tests for errno was in my todo list in Jan; but unfortunately, I got pulled into other things.
> A multi-threaded test for errno propagation would only confirm that the implementation of errno in Windows and Linux is correct, which I am willing to accept on faith. The edge routines contain the following assignment:
> pargs_out->_ocall_errno = errno;
> 
> Keep in mind that switchless calls are coming and that scenario involves a many-to-1 bindings between host and enclave threads. Also we would like play with different kinds of code-generation in the future and optimize the generated code.

I think you just broadened the scope to include multi-threaded testing of edge routines in general, and I am in favor of that. But I hold to my original point about **errno**, that it is an aspect of Windows or Linux and there is little to be gained by targeting **errno**. Also, you have to look at all this in terms of priorities.

anakrish:
I think I may have clicked the wrong button and closed this PR.
github does not allow me to unclose it.

PR #1780, opened 2019-05-10
Statically decouple liboecore and liboeenclave

mikbras:
This PR has three parts.

1. It removes the static dependency of **liboecore** on **liboenclave** through the use of dynamically installed upcalls.

2. It provides the new **liboestart** library with an alternative **_start** symbol.

3. It implements the **mintcb** test, which creates an enclave with no **liboeenclave** dependency.


tests/mintcb/enc/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+
+oeedl_file(../mintcb.edl enclave gen)
+
+add_executable(mintcb_enc enc.c ${gen})
```
radhikaj: Would mintcb_enc be a CMake exported target?

enclave/sgx/start.S
```diff
@@ -13,28 +13,21 @@
...
+    pushq %rdi
+    pushq %rsi
+    /* Dynamically link liboecore to liboeenclave (install upcalls). */
+    call oe_link_enclave
+    /* Restore the registers passed by EENTER. */
+    popq %rsi
+    popq %rdi
+    popq %rcx
+    popq %rbx
+    popq %rax
```
anitagov: I double-checked that the registers that you saved/restored are passed by EENTER.

tests/mintcb/README.md
```diff
@@ -0,0 +1,8 @@
+mintcb
+======
+
+This test shows how to create a minimal-TCB enclave that only links
+**liboecore** (thereby avoiding **liboeenclave** and its dependencies).
+
+Ordinarily, the **_start** entry point is linked from **liboeenclave**. This
+test, links this entry point defined in **liboestart**.
```
anitagov: I like the test but wondering if you can print the TCB size. We can then compare this with regular TCB enclave size (may need a similar test that links from liboeenclave).

soccerGB:
Changes look good.

It would be great to see this merged into the master soon

yakman2020:
bors try

enclave/start/sgx/start.S
```diff
@@ -0,0 +1,20 @@
...
+//==============================================================================
+//
+// _start():
+//
+//     The default entry point for the enclave image. This entry point jumps
+//     directly to the oe_enter() function in liboecore.
+//
+//==============================================================================
+
+.globl _start
```
jxyang: I am a little lost: why do we have to define `_start` again? The other one is in enclave/sgx/start.S.

enclave/sgx/start.S
```diff
@@ -13,28 +13,21 @@
...
 _start:
 .cfi_startproc
+    /* Save the registers passed by EENTER. */
+    pushq %rax
+    pushq %rbx
+    pushq %rcx
+    pushq %rdi
+    pushq %rsi
+    /* Dynamically link liboecore to liboeenclave (install upcalls). */
+    call oe_link_enclave
```
jxyang: >call oe_link_enclave [](start = 4, length = 20)

Can we move the call to oe_enter, so that we can potentially do away with those register save/restores?

enclave/start/sgx/start.S
BRMcLaren: I was curious too and started looking around.

I believe what has happened is that it's due to a subdirectory inclusion in the CMake files, so it is defined twice if this is an SGX machine, and once if not. The first inclusion also seems to do some registry pushing.

Looking further, however, I believe this has to do with scope limitations in CMake. Looking at sgx/CMakeLists.txt, we can see the subdirectory enclusion of /start if this is an SGX machine.

If we take a look at the library dependencies of start/CMakeLists.txt, we see the library dependency of Start.S. 

Disclaimer: I am not _overly familiar with assembly_, so please take this with a grain of salt. 

But it seems that the `cfi_startproc` needs to be different when being used on an sgx machine, which entails not using the registry pushing when called from the `oestart` library. This is might be the reason why it was chosen to not encapsulate the registry pushing when redefining this.

Which might have resulted in the need to keep `oe_link_enclave` out of the `oe_enter` call, as commented above. If this is true, a slight refactor may be needed but it seems to be the correct way of handling this.

kuanggang-git:

jhand2:
liboecore no longer depends on liboeenclave. Can this be closed?

PR #1781, opened 2019-05-11
Pull upstream edger8r code from Intel

andschwa:
This is copied from https://github.com/intel/linux-sgx at tag `sgx_2.5`.

It is a straight copy, meaning any and all of our modifications are now
gone, but this is intentional as patches should be applied on top only
as needed.

Really the only change we need to deal with immediately is that `struct`
and `union` types are now represented slightly differently (in order to
support nesting types in structs). This gets us back to the status quo,
and enables us to start developing deep copy support.


andschwa:
bors try

PR #1782, opened 2019-05-11
Added oecryptombed target

gupta-ak:
This PR refactors the crypto code from oeenclave and puts it into a oecryptombed target. It removes `mbedtls_hardware_poll` in oecore and replaces it with an `oe_get_entropy` function. The `mbedtls_hardware_poll` function itself is now bundled with mbedtls.

I'm breaking up the crypto refactor into multiple parts for easier readability.


3rdparty/mbedtls/CMakeLists.txt
```diff
@@ -44,9 +45,18 @@ endif ()
 list(JOIN MBEDTLS_WRAP_CFLAGS_LIST " " MBEDTLS_WRAP_CFLAGS)
 configure_file(config.h config.h) # This copies from source to binary folders.
 
+# Since we define mbedtls to use an alternate entropy source, it uses an
+# undefined mebdtls_hardware_poll function. We define it here and add it to
+# the mbedcrypto library, so that we avoid link errors and also circular
+# library dependecies.
```
gupta-ak: FYI, this file gets entirely rewritten in a future PR since imported library targets don't work properly in cmake with install targets.

3rdparty/mbedtls/CMakeLists.txt
```diff
@@ -73,6 +83,11 @@ ExternalProject_Add(mbedtls-wrap
 ExternalProject_Get_property(mbedtls-wrap SOURCE_DIR)
 ExternalProject_Get_property(mbedtls-wrap BINARY_DIR)
 
+# Add the mbedtls_hardware_poll function to the mbedcrypto library.
+add_custom_command(TARGET mbedtls-wrap
+    POST_BUILD
+    COMMAND ${CMAKE_AR} r ${BINARY_DIR}/library/libmbedcrypto.a $<TARGET_OBJECTS:oe_mbedtls_hardware_poll>)
```
radhikaj: Can this be ${LLVM_AR}?

enclave/crypto/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+    ../../common/cert.c
+    ../../common/kdf.c
+    asn1.c
+    cert.c
+    crl.c
+    ec.c
+    cmac.c
+    hmac.c
+    key.c
+    random.c
```
radhikaj: Random.c is in this lib as well as in oeenclave. Is that intentional?

include/openenclave/internal/rdrand.h
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Generate a sequence of random bytes using Intel RDRAND instruction
+ *
+ * This function generates 8 random bytes using direct call to Intel's RDRAND
+ * instruction. This method will block if there is insufficient hardware
+ * entropy to provide the full 64-bits of randomness.
+ *
+ * @return OE_OK on success
+ */
+uint64_t oe_rdrand(void);
```
radhikaj: For my understanding, why does this need to be in a new header instead of random.h?

enclave/core/sgx/entropy.c
```diff
@@ -50,10 +40,7 @@ int mbedtls_hardware_poll(
             *p++ = *q++;
```
radhikaj: What is q here? Is it a global variable? I am trying to understand this and dont see q defined in this function.

andschwa: The GitHub diff is bad and cut it out, it's a few lines above:

> const unsigned char* q = (const unsigned char*)&x;

pkgconfig/CMakeLists.txt
```diff
@@ -54,6 +54,7 @@ set(ENCLAVE_CLIBS_1
   -Wl,-z,now
   -L\${libdir}/openenclave/enclave
   -loeenclave
+  -loecryptombed
```
radhikaj: Would the sample cmake files need to be updated to link with oecryptombed?

enclave/core/sgx/entropy.c
radhikaj: @andschwa  Thanks :)

enclave/crypto/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+add_library(oecryptombed STATIC
+    ../../common/asn1.c
+    ../../common/cert.c
+    ../../common/kdf.c
```
radhikaj: Since kdf.c does not have any mbedtls (or openssl) specific calls, does it need to be here? All the other .c files that you are including here require calls unique to the specific crypto that are used.

3rdparty/mbedtls/CMakeLists.txt
gupta-ak: `${LLVM_AR}` is the path to the llvm archiver on clang on windows.  See the code that line 27:
```
  find_program(LLVM_AR "llvm-ar.exe")
  set(MBEDTLS_LINKER "-DCMAKE_AR=${LLVM_AR}")
  set(MBEDTLS_LINKER "-DCMAKE_AR=${LLVM_AR}")
  set(MBEDTLS_COMPILER "")
  set(MBEDTLS_COMPILER "")
  set(CMAKE_AR ${LLVM_AR})
```
So, `${CMAKE_AR} == ${LLVM_AR}` on clang-on-windows and the correct archiver in other systems.

enclave/crypto/CMakeLists.txt
gupta-ak: Yeah. The random.c in `oeenclave` defines the `oe_rand` function, which basically just calls `oe_random_internal` and the random.c in `oecrypto` has all the mbedtls specific stuff, along with `oe_random_internal`. I'm not sure why there is an `oe_random_internal` in the first space, since all `oe_rand` does is call it.

include/openenclave/internal/rdrand.h
gupta-ak: This isn't part of the crypto API. It's specific to oecore and x86_64 for generating a entropy seed for the platform.

pkgconfig/CMakeLists.txt
gupta-ak: The samples use pkg-config, so updating this file automatically updates them.

enclave/crypto/CMakeLists.txt
gupta-ak: @CodeMonkeyLeet @mikbras We need a discussion on what is part of the OE crypto API. The API was originally made as a way to share code with a common API for the host and the enclave, but now we're using the API for pluggable crypto, which is a different use case.

gupta-ak:
bors try

3rdparty/mbedtls/mbedtls_hardware_poll.c
```diff
@@ -0,0 +1,32 @@
...
+ * for MBEDTLS. All other MBEDTLS entropy sources are disabled since they don't
+ * work within enclaves.
+ */
+int mbedtls_hardware_poll(
+    void* data,
+    unsigned char* output,
+    size_t len,
+    size_t* olen)
+{
+    int ret = -1;
```
anakrish: Can you follow up with @mikbras if the convention is to always initialize out parameters to known values, even in case of failures?
If so, in this case, we would initialize *olen to 0, for a non null olen.

enclave/crypto/CMakeLists.txt
anakrish: Maybe we could rename one of the files to appropriately reflect the function it defines.

enclave/core/sgx/entropy.c
```diff
@@ -3,24 +3,14 @@
...
-    unsigned char* p = output;
+#include <openenclave/internal/entropy.h>
+#include <openenclave/internal/rdrand.h>
 
-    OE_UNUSED(data);
+/* TODO: This should use RDSEED instead. See issue #242. */
+oe_result_t oe_get_entropy(void* output, size_t len)
+{
+    oe_result_t ret = OE_UNEXPECTED;
+    unsigned char* p = (unsigned char*)output;
```
anakrish: Can you also take a pass at cleaning up this function?
Do we really need to use memcpy_s here or just do a uint64_t volatile write?
We don't do any memcpy_s for the left over bytes.

enclave/crypto/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+    sha.c)
+
+maybe_build_using_clangw(oecryptombed)
+
+target_include_directories(oecryptombed PRIVATE
+    ${PROJECT_SOURCE_DIR}/include/openenclave/corelibc)
+
+target_link_libraries(oecryptombed PUBLIC
+    oe_includes
+    mbedcrypto)
```
anakrish: So there will be 4 crypto libraries? liboecryptombed.a, libmbedtls.a, ….
Is there a plan to merge all of them to a single library?

enclave/crypto/random.c
```diff
@@ -0,0 +1,81 @@
...
+static oe_result_t _seed_result = OE_UNEXPECTED;
+static oe_once_t _seed_once = OE_ONCE_INIT;
+
+/* Wrapper to set file-scope _seed_result */
+static void _seed_entropy_source_once()
+{
+    _seed_result = _seed_entropy_source();
+}
+
+mbedtls_ctr_drbg_context* oe_mbedtls_get_drbg()
```
anakrish: This function does not check the seed result. Is that by design?

3rdparty/mbedtls/mbedtls_hardware_poll.c
gupta-ak: @mikbras can you comment on this?

enclave/core/sgx/entropy.c
gupta-ak: i'm not sure what you mean by a uint64_t volatile write. If you mean something like:
```C
while (n--)
{
    uint64_t x = oe_rdrand();
    *((uint64_t*) p) = x;
    p += sizeof(uint64_t);
}
```
That's really dangerous since p isn't guaranteed to be aligned to 8 bytes, so it can crash. You need a memcpy-like function like `memcpy_(s)` or the `while(...) { *p++ = *q++ }` loop.

enclave/crypto/CMakeLists.txt
gupta-ak: mbedtls is actually 3 libraries (mbedtls, libmbedx509, libmbedcrypto). Since we're saying that you can use the crypto library we use for your enclave applications, I don't think we should combine them, since users would expect those three mbedtls libraries to exist.

enclave/crypto/CMakeLists.txt
gupta-ak: I'll name the crypto one `random_internal.c`

enclave/crypto/random.c
gupta-ak: That's how the function was implemented. @mikbras do you know why the seed result isn't checked.

enclave/core/sgx/entropy.c
anakrish: Good point about alignment.
We need to change line 40 to use memcpy_s

enclave/crypto/random.c
CodeMonkeyLeet: Regardless of the intent, this looks like a bug; the expectation is that this function should return NULL on failure (see enclave/ec.c or rsa.c), but `&_drbg` is always a valid pointer, regardless of whether it was ever initialized or successfully seeded.

gupta-ak: Ok. Sounds good. It looked like a bug to me too, so I wasn't sure why it was implemented that way.

enclave/crypto/CMakeLists.txt
gupta-ak: So, all the code in common (`asn1.c`, `cert.c`, `kdf.c`) can be purely through the oecrypto APIs. However, if you take these files and move them out to `oeenclave`, you run into linking problems, since the enclave specific code in `oecryptombed` might refer to a symbol that's defined in `common/[asn1.c|cert.c|kdf.c]`.

Speficially, the issue I hit was that `common/asn1.c` defines a functions that is used by `enclave/crypto/asn1.c` so the linking fails.

radhikaj: We can merge this pull request and continue refactoring incrementally from then on.

andschwa: Yeah, a similar discussion just came up in #1825. When there is further incremental work to be done, we should capture that in an issue(s) for the current release's milestone, and allow the partial implementation to be merged in order to avoid long-lived branches.

3rdparty/mbedtls/CMakeLists.txt
gupta-ak: I filed issue #1836

PR #1784, opened 2019-05-13
Disable simulation mode on Windows because of sporadic crashes

### LINKED ISSUE #1753: Need to disable simulation mode on Windows
We need to disable simulation mode Windows because it crashes sporadically. Linux uses the FS register for pthreads and leaves the GS register unused. Simulation mode on Linux takes over the GS register. But Windows uses the FS and GS register and these are corrupted when calling into the enclave in simulation mode.

Suggestion: oe_create_enclave() should fail on Windows whit the simulation mode flag.

jxyang:
Fixes #1753 and description from issue is given below:

We need to disable simulation mode Windows because it crashes sporadically. Linux uses the FS register for pthreads and leaves the GS register unused. Simulation mode on Linux takes over the GS register. But Windows uses the FS and GS register and these are corrupted when calling into the enclave in simulation mode.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -187,6 +187,7 @@ if (USE_DEBUG_MALLOC AND WIN32)
 endif ()
 
 option(ADD_WINDOWS_ENCLAVE_TESTS "Build Windows enclave tests" OFF)
+# Warning: turning on simulation mode on Windows may cause test failures and random crashes
 option(WIN32_SIMULATION "Windows Simulation Mode" OFF)
 
 find_program(VALGRIND "valgrind")
--- a/host/sgx/create.c
+++ b/host/sgx/create.c
@@ -647,6 +647,9 @@ oe_result_t oe_create_enclave(
         OE_RAISE(OE_OUT_OF_MEMORY);
 
 #if defined(_WIN32)
+    /* Disable simulation mode on windows */
+    if (flags & OE_ENCLAVE_FLAG_SIMULATE)
+        OE_RAISE(OE_INVALID_PARAMETER);
 
     /* Create Windows events for each TCS binding. Enclaves use
      * this event when calling into the host to handle waits/wakes


anitagov:
Please fix the description as it seems like an incomplete cut and paste from the issue. Suggestion to rephrase in the PR appropriately (I think the referenced issue may get duplicated in the log but you may want to verify that). Thanks.

anitagov:
Please issue the command "bors try" (without the quotes) in this PR as this will fire the CI run. Trust this helps.

jxyang:
bors try

anitagov:
bors try

jxyang:
bors try

jxyang:
bors try

anitagov:
Observed that the WIN32_SIMULATION knob is set to OFF by default and this mode is not run on our Jenkins CI servers. So, we are good.

Please modify the documentation to reflect that Windows simulation mode is currently unsupported due to #1784 . This seems to be the only outstanding issue right now.

host/sgx/create.c
```diff
@@ -647,6 +647,9 @@ oe_result_t oe_create_enclave(
         OE_RAISE(OE_OUT_OF_MEMORY);
 
 #if defined(_WIN32)
+    /* Disable simulation mode on windows */
+    if (flags & OE_ENCLAVE_FLAG_SIMULATE)
+        OE_RAISE(OE_INVALID_PARAMETER);
```
anitagov: Nice

CMakeLists.txt
```diff
@@ -187,6 +187,7 @@ if (USE_DEBUG_MALLOC AND WIN32)
 endif ()
 
 option(ADD_WINDOWS_ENCLAVE_TESTS "Build Windows enclave tests" OFF)
+# Warning: turning on simulation mode on Windows may cause test failures and random crashes
 option(WIN32_SIMULATION "Windows Simulation Mode" OFF)
```
anitagov: Right now, enabling Windows simulation mode will result in failure to create enclave. So your comment is not correct. Consider rephrasing appropriately.

Anyways, my request was to make a note about simulation mode being currently disabled in the GettingStarted.Windows.md guide and also reference the issue number.

anitagov: This is fine.

Anyways, my request was to make a note about simulation mode being currently disabled in the GettingStarted.Windows.md guide and also reference the issue number. Can you make that change please?

jxyang: Sure.

anakrish:
@jxyang Can you update the GettingStarted.Windows.md as @anitagov suggests? With this PR, that document is no longer accurate.

anitagov:
Thanks for your feedback @anakrish . @jxyang has addressed doc changes/fixes with his PR #1802

PR #1785, opened 2019-05-13
Fix EDL dependencies in oeedger8r tests

andschwa:
This removes the use of the (buggy) `oeedl_file` CMake script in favor
of using CMake custom commands directly. The script does not support
setting more dependencies easily, and abstracts too much making the
build code difficult to follow. Instead we use commands directly, which
means we can explicitly set our build dependencies to include the EDL
files imported in `all.edl`. This fixes the bug where the edger8r was
not rerun if an imported file was changed.


andschwa:
This was really annoying; I was editing an imported EDL file and kept looking for my changes to appear...

andschwa:
bors try

PR #1791, opened 2019-05-15
Add (ignored) oeedger8r tests for deep copy scenarios

andschwa:
This adds (and ignores the expected failures of) unit tests for the edger8r covering a set of scenarios on an internal document.

As we implement deep-copy support, we can enable these tests by changing `OE_TEST_IGNORE` to `OE_TEST`.

### DIFF
--- a/include/openenclave/internal/tests.h
+++ b/include/openenclave/internal/tests.h
@@ -21,7 +21,7 @@
 
 OE_EXTERNC_BEGIN
 
-#define OE_TEST(COND)                           \
+#define OE_TEST_IF(COND, OE_TEST_ABORT)         \
     do                                          \
     {                                           \
         if (!(COND))                            \
@@ -33,9 +33,14 @@ OE_EXTERNC_BEGIN
                 __LINE__,                       \
                 __FUNCTION__,                   \
                 #COND);                         \
-            OE_ABORT();                         \
+            if (OE_TEST_ABORT)                  \
+                OE_ABORT();                     \
         }                                       \
     } while (0)
+
+#define OE_TEST(COND) OE_TEST_IF(COND, true)
+
+#define OE_TEST_IGNORE(COND) OE_TEST_IF(COND, false)
 
 /*
  * Return flags to pass to oe_create_enclave() based on the OE_SIMULATION
--- a/tests/oeedger8r/behavior/CMakeLists.txt
+++ b/tests/oeedger8r/behavior/CMakeLists.txt
@@ -73,3 +73,7 @@ set_tests_properties(edger8r_count_signedness_warning PROPERTIES
 add_test(NAME edger8r_size_and_count_warning COMMAND edger8r ${EDGER8R_ARGS} size_and_count.edl)
 set_tests_properties(edger8r_size_and_count_warning PROPERTIES
   PASS_REGULAR_EXPRESSION "Function 'size_and_count': simultaneous 'size' and 'count' parameters 'size' and 'count' are not supported by oeedger8r.")
+
+add_test(NAME edger8r_deepcopy_value_warning COMMAND edger8r ${EDGER8R_ARGS} deepcopy_value.edl)
+set_tests_properties(edger8r_deepcopy_value_warning PROPERTIES
+  PASS_REGULAR_EXPRESSION "error: the structure declaration \"MyStruct\" specifies a deep copy is expected. Referenced by value in function \"deepcopy_value\" detected.")
--- a/tests/oeedger8r/enc/testdeepcopy.cpp
+++ b/tests/oeedger8r/enc/testdeepcopy.cpp
@@ -45,7 +45,7 @@ void deepcopy_count(CountStruct* s)
     OE_TEST(s->size == 64);
     for (size_t i = 0; i < 3; ++i)
         OE_TEST(s->ptr[i] == data[i]);
-    OE_TEST(oe_is_within_enclave(s->ptr, 3 * sizeof(uint64_t)));
+    OE_TEST_IGNORE(oe_is_within_enclave(s->ptr, 3 * sizeof(uint64_t)));
 }
 
 // Assert that the struct is deep-copied such that `s->ptr` has a copy
@@ -56,7 +56,7 @@ void deepcopy_countparam(CountParamStruct* s)
     OE_TEST(s->size == 64);
     for (size_t i = 0; i < s->count; ++i)
         OE_TEST(s->ptr[i] == data[i]);
-    OE_TEST(oe_is_within_enclave(s->ptr, s->count * sizeof(uint64_t)));
+    OE_TEST_IGNORE(oe_is_within_enclave(s->ptr, s->count * sizeof(uint64_t)));
 }
 
 // Assert that the struct is deep-copied such that `s->ptr` has a copy
@@ -67,7 +67,7 @@ void deepcopy_sizeparam(SizeParamStruct* s)
     OE_TEST(s->size == 64);
     for (size_t i = 0; i < s->size / sizeof(uint64_t); ++i)
         OE_TEST(s->ptr[i] == data[i]);
-    OE_TEST(oe_is_within_enclave(s->ptr, s->size));
+    OE_TEST_IGNORE(oe_is_within_enclave(s->ptr, s->size));
 }
 
 // Assert that the struct is deep-copied such that `s->ptr` has a copy
@@ -78,7 +78,7 @@ void deepcopy_countsizeparam(CountSizeParamStruct* s)
     OE_TEST(s->size == 4);
     for (size_t i = 0; i < (s->count * s->size) / sizeof(uint64_t); ++i)
         OE_TEST(s->ptr[i] == data[i]);
-    OE_TEST(oe_is_within_enclave(s->ptr, s->count * s->size));
+    OE_TEST_IGNORE(oe_is_within_enclave(s->ptr, s->count * s->size));
 }
 
 // Assert that the struct array is deep-copied such that each
@@ -90,11 +90,13 @@ void deepcopy_countparamarray(CountParamStruct* s)
     OE_TEST(s[0].size == 64);
     for (size_t i = 0; i < s[0].count; ++i)
         OE_TEST(s[0].ptr[i] == data[i]);
-    OE_TEST(oe_is_within_enclave(s[0].ptr, s->count * sizeof(uint64_t)));
+    OE_TEST_IGNORE(
+        oe_is_within_enclave(s[0].ptr, s[0].count * sizeof(uint64_t)));
 
     OE_TEST(s[1].count == 3);
     OE_TEST(s[1].size == 32);
     for (size_t i = 0; i < s[1].count; ++i)
         OE_TEST(s[1].ptr[i] == data[4 + i]);
-    OE_TEST(oe_is_within_enclave(s[1].ptr, s->count * sizeof(uint64_t)));
+    OE_TEST_IGNORE(
+        oe_is_within_enclave(s[1].ptr, s[1].count * sizeof(uint64_t)));
 }
--- a/tests/oeedger8r/enc/testdeepcopy.cpp
+++ b/tests/oeedger8r/enc/testdeepcopy.cpp
@@ -69,3 +69,14 @@ void deepcopy_sizeparam(SizeParamStruct* s)
         OE_TEST(s->ptr[i] == data[i]);
     OE_TEST(oe_is_within_enclave(s->ptr, s->size));
 }
+
+// Assert that the struct is deep-copied such that `s->ptr` has a copy
+// of `s->count * s->size` bytes of `data` in enclave memory.
+void deepcopy_countsizeparam(CountSizeParamStruct* s)
+{
+    OE_TEST(s->count == 8);
+    OE_TEST(s->size == 4);
+    for (size_t i = 0; i < (s->count * s->size) / sizeof(uint64_t); ++i)
+        OE_TEST(s->ptr[i] == data[i]);
+    OE_TEST(oe_is_within_enclave(s->ptr, s->count * s->size));
+}
--- a/tests/oeedger8r/enc/testdeepcopy.cpp
+++ b/tests/oeedger8r/enc/testdeepcopy.cpp
@@ -118,3 +118,21 @@ void deepcopy_sizeparamarray(SizeParamStruct* s)
         OE_TEST(s[1].ptr[i] == data[4 + i]);
     OE_TEST_IGNORE(oe_is_within_enclave(s[1].ptr, s[1].size));
 }
+
+// Assert that the struct array is deep-copied such that each
+// element's `ptr` has a copy of its `count * size` bytes of `data` in
+// enclave memory.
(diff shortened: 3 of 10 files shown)


tests/oeedger8r/enc/testdeepcopy.cpp
```diff
@@ -0,0 +1,136 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "../edltestutils.h"
+
+#include <openenclave/enclave.h>
+#include <openenclave/internal/tests.h>
+#include "all_t.h"
+
+static uint64_t data[8] = {0x1112131415161718,
```
andschwa: Comment from John: move to shared header.

tests/oeedger8r/behavior/CMakeLists.txt
```diff
@@ -73,3 +73,12 @@ set_tests_properties(edger8r_count_signedness_warning PROPERTIES
...
   PASS_REGULAR_EXPRESSION "Function 'size_and_count': simultaneous 'size' and 'count' parameters 'size' and 'count' are not supported by oeedger8r.")
+
+add_test(NAME edger8r_deepcopy_value_warning COMMAND edger8r ${EDGER8R_ARGS} deepcopy_value.edl)
+set_tests_properties(edger8r_deepcopy_value_warning PROPERTIES
+  PASS_REGULAR_EXPRESSION "error: the structure declaration \"MyStruct\" specifies a deep copy is expected. Referenced by value in function \"deepcopy_value\" detected.")
+
+# TODO: This test won't pass until support is added.
+add_test(NAME edger8r_deepcopy_oddsize_warning COMMAND edger8r ${EDGER8R_ARGS} deepcopy_oddsize.edl)
+# set_tests_properties(edger8r_deepcopy_oddsize_warning PROPERTIES
+  # PASS_REGULAR_EXPRESSION "TODO: Choose an appropriate error message.")
```
anitagov: Remove lines 83/84 if you don't need them.

tests/oeedger8r/host/testdeepcopy.cpp
```diff
@@ -0,0 +1,87 @@
...
+        auto s = init_structs<CountSizeParamStruct>();
+        s[0].count = 8;
+        s[0].size = 4;
+        s[1].count = 3;
+        s[1].size = 8;
+        OE_TEST(deepcopy_countsizeparamarray(enclave, s.data()) == OE_OK);
+    }
+
+    printf("=== test_deepcopy_edl_ecalls failures ignored!\n");
+}
```
anitagov: Great if you can comment the reason for the chosen values for count and size when possible.

include/openenclave/internal/tests.h
```diff
@@ -33,10 +33,15 @@ OE_EXTERNC_BEGIN
...
                 #COND);                         \
-            OE_ABORT();                         \
+            if (OE_TEST_ABORT)                  \
+                OE_ABORT();                     \
         }                                       \
     } while (0)
 
+#define OE_TEST(COND) OE_TEST_IF(COND, true)
+
+#define OE_TEST_IGNORE(COND) OE_TEST_IF(COND, false)
```
anitagov: Nice

tests/oeedger8r/behavior/CMakeLists.txt
andschwa: We do need them, but the test isn't enable-able yet (see comment).

tests/oeedger8r/host/testdeepcopy.cpp
andschwa: The reason is that their the scenarios presented in an internal document, which this PR implements as tests.

andschwa:
bors try

tests/oeedger8r/behavior/CMakeLists.txt
```diff
@@ -73,3 +73,12 @@ set_tests_properties(edger8r_count_signedness_warning PROPERTIES
 add_test(NAME edger8r_size_and_count_warning COMMAND edger8r ${EDGER8R_ARGS} size_and_count.edl)
 set_tests_properties(edger8r_size_and_count_warning PROPERTIES
   PASS_REGULAR_EXPRESSION "Function 'size_and_count': simultaneous 'size' and 'count' parameters 'size' and 'count' are not supported by oeedger8r.")
+
+add_test(NAME edger8r_deepcopy_value_warning COMMAND edger8r ${EDGER8R_ARGS} deepcopy_value.edl)
+set_tests_properties(edger8r_deepcopy_value_warning PROPERTIES
+  PASS_REGULAR_EXPRESSION "error: the structure declaration \"MyStruct\" specifies a deep copy is expected. Referenced by value in function \"deepcopy_value\" detected.")
```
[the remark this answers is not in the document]
anakrish: _Referenced by value in function deepcopy_value detected._ is bit hard to parse. Not clear what the error message means.

tests/oeedger8r/behavior/deepcopy_oddsize.edl
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+enclave {
+  struct MyStruct {
+    [size=13] int* ptr;
```
anakrish: Can you add a comment why size is 13?
Is it intentionally not a multiple of sizeof(int)?

tests/oeedger8r/behavior/deepcopy_oddsize.edl
```diff
@@ -0,0 +1,14 @@
...
+  };
+
+  trusted {
+    // This should error because the size parameter attached to
+    // `s.ptr` is not a multiple of the size of `int`.
+    public void deepcopy_oddsize([in, count=1] struct MyStruct* s);
```
anakrish: Ah ok. Are we doing the validation? If so can the validation be done at the struct declaration rather than the use?

tests/oeedger8r/enc/testdeepcopy.cpp
```diff
@@ -0,0 +1,136 @@
...
+    OE_TEST(s->size == 64);
+    OE_TEST(s->ptr == ptr);
+    OE_TEST(oe_is_outside_enclave(s->ptr, sizeof(uint64_t)));
+}
+
+// Assert that the struct is deep-copied such that `s->ptr` has a copy
+// of three elements of `data` in enclave memory.
+void deepcopy_count(CountStruct* s)
+{
+    OE_TEST(s->count == 7);
```
anakrish: Do these tests (and fields) here add any value? I was about to suggest that the oe_is_within_enclave ought to use size or count and then realized that those fields weren't being used as annotations.

tests/oeedger8r/enc/testdeepcopy.cpp
```diff
@@ -0,0 +1,136 @@
...
+    OE_TEST_IGNORE(oe_is_within_enclave(s->ptr, s->size));
+}
+
+// Assert that the struct is deep-copied such that `s->ptr` has a copy
+// of `s->count * s->size` bytes of `data` in enclave memory.
+void deepcopy_countsizeparam(CountSizeParamStruct* s)
+{
+    OE_TEST(s->count == 8);
+    OE_TEST(s->size == 4);
+    for (size_t i = 0; i < (s->count * s->size) / sizeof(uint64_t); ++i)
```
anakrish: Can you double-check the upper bound? Is that what sgx_edger8r does?

tests/oeedger8r/enc/testdeepcopy.cpp
```diff
@@ -0,0 +1,136 @@
...
+// Assert that the struct array is deep-copied such that each
+// element's `ptr` has a copy of its `count` elements of `data` in
+// enclave memory.
+void deepcopy_countparamarray(CountParamStruct* s)
+{
+    OE_TEST(s[0].count == 7);
+    OE_TEST(s[0].size == 64);
+    for (size_t i = 0; i < s[0].count; ++i)
+        OE_TEST(s[0].ptr[i] == data[i]);
+    OE_TEST_IGNORE(oe_is_within_enclave(s[0].ptr, s->count * sizeof(uint64_t)));
```
anakrish: Nice.

tests/oeedger8r/behavior/deepcopy_oddsize.edl
andschwa: Yup.

tests/oeedger8r/behavior/deepcopy_oddsize.edl
andschwa: We are not (yet). It's a pending test that's not enabled, because we ought to validate, but don't, and Intel doesn't.

tests/oeedger8r/enc/testdeepcopy.cpp
andschwa: What do you mean?

tests/oeedger8r/enc/testdeepcopy.cpp
andschwa: They're just being really explicit based on the documented scenarios (which also state what each value should be before and after).

tests/oeedger8r/behavior/CMakeLists.txt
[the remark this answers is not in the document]
andschwa: It's a warning emitted by Intel's part of the edger8r, and they chose to use "error:"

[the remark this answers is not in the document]
andschwa: @anakrish I agree, I think that was even in my feedback 😆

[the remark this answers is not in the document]
anakrish: Ah ok. Can you spin up an email-thread with intel asking them to change this/other suggestions you have, now that you are taking a deeper look at the deep-copy feature?

tests/oeedger8r/enc/testdeepcopy.cpp
anakrish: Ok.

tests/oeedger8r/enc/testdeepcopy.cpp
anakrish: Is the upper bound of the loop supposed to be count*size?
(I don't yet see a use case for this edger8r feature that allows specifying both size and count)

tools/oeedger8r/Emitter.ml
```diff
@@ -141,43 +106,48 @@ let open_file (filename : string) (dir : string) =
...
-      "" new_param_list
+let oe_gen_marshal_struct =
+  let gen_member_decl (ptype : parameter_type) (decl : declarator) =
+    let tystr = get_tystr (get_param_atype ptype) in
+    let tystr =
+      if is_foreign_array ptype then
+        sprintf "/* foreign array of type %s */ void*" tystr
+      else tystr
+    in
+    (* TODO: We shouldn't have to emit a strlen variable for ocalls,
```
anakrish: It is simpler to have the marshaling strategy be consistent for ecalls and ocalls.
Is the concern about the length variables for ocalls in regards to space?

tools/oeedger8r/Emitter.ml
```diff
@@ -141,43 +106,48 @@ let open_file (filename : string) (dir : string) =
...
+       but currently we require it. *)
+    let need_strlen p =
+      (is_str_ptr p || is_wstr_ptr p) && (is_in_ptr p || is_inout_ptr p)
+    in
+    [ sprintf "%s %s;" tystr decl.identifier
+    ; ( if need_strlen ptype then sprintf "size_t %s_len;" decl.identifier
```
anakrish: It would be preferable to use _uint64_t_ instead of _size_t_ for variables we introduce.

tests/oeedger8r/enc/testdeepcopy.cpp
andschwa: Oh, yes, that's my understanding from the scenarios document, but I'll double-check.

tests/oeedger8r/enc/testdeepcopy.cpp
andschwa: Thanks!

tools/oeedger8r/Emitter.ml
andschwa: Sorry, my comment wasn't very clear. Intel states [here](https://github.com/microsoft/openenclave/blob/6b5f1612ab124d05cba596f48b80558c4da36864/tools/oeedger8r/intel/CodeGen.ml#L324) a reason not to do this for ocalls, but I (briefly) tested removing it for us, and it didn't work, it's just a TODO to revisit.

tools/oeedger8r/Emitter.ml
andschwa: Why's that?

andschwa:
> From reading the code, it seems that most if not all of the Ocaml changes have nothing to do with the tests being added.
> If so, I would prefer submitting these unrelated changes as separate PRs.

Oh, yeah, I branched off a little late. I can remove those changes. They were originally intended as "not quite deep-copy related and not breaking changes so let's get them in early" but my prototype went quicker than expected so I'll move them out of here.

tools/oeedger8r/Emitter.ml
anakrish: I don't think that reason is valid. Strings can be **in-out** parameters for ocalls, which is almost like an in parameter for ecalls.

tools/oeedger8r/Emitter.ml
andschwa: We are using `size_t` in a lot of other places in the generated code already...should we address all of their uses? It looks like it's not as consistent a type as it probably should have been.

That said, this is used to compare against the [sizeof operator](https://en.cppreference.com/w/c/language/sizeof) which the standard states explicitly returns a `size_t`.

dthaler: size_t is correct is many cases.  It's just not correct any time you need cross-machine interop, such as defining fields that are transmitted between machines, or stored in files that may be read by different machines.

andschwa:
Removed OCaml changes from this PR.

tests/oeedger8r/behavior/CMakeLists.txt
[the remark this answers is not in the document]
andschwa: Will do.

tests/oeedger8r/behavior/deepcopy_oddsize.edl
andschwa: I'm going to drop the scenario 6 yet because we don't actually enable its test (even in an ignored state).

tests/oeedger8r/enc/testdeepcopy.cpp
andschwa: Not gonna do for now.

tests/oeedger8r/enc/testdeepcopy.cpp
andschwa: I checked, and yes, the expected before if both count and size are specified that the product of count and size is the number of bytes to marshal. Dividing that by the size of the element then yields the number elements marshalled (this upper bound).

There's an open question on what to do if `count * size` is not an even multiple of the size of the elements, which is pending work for us to add checks around, but is specifically not tested here.

PR #1792, opened 2019-05-16
Set correct ARM build options based on selected board

dthaler:

### DIFF
--- a/new_platforms/vsextension/ProjectWizard/WizardImplementation.cs
+++ b/new_platforms/vsextension/ProjectWizard/WizardImplementation.cs
@@ -83,6 +83,13 @@ namespace OpenEnclaveSDK
             string unixPath = "/mnt/" + drive + "/" + relativeFolder;
 
             replacementsDictionary.Add("$OETADevKitPath$", unixPath);
+
+            string buildFlavor = board.Split('\\')[0];
+            replacementsDictionary.Add("$OpteeBuildFlavor$", buildFlavor);
+
+            string opteeCompilerFlavor = (buildFlavor == "vexpress-qemu_virt") ? "arm-linux-gnueabihf-" : "aarch64-linux-gnu-";
+            replacementsDictionary.Add("$OpteeCompilerFlavor$", opteeCompilerFlavor);
+
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


new_platforms/vsextension/ProjectWizard/WizardImplementation.cs
```diff
@@ -83,6 +83,13 @@ private bool SetOETADevKitPath(Dictionary<string, string> replacementsDictionary
             string unixPath = "/mnt/" + drive + "/" + relativeFolder;
 
             replacementsDictionary.Add("$OETADevKitPath$", unixPath);
+
+            string buildFlavor = board.Split('\\')[0];
+            replacementsDictionary.Add("$OpteeBuildFlavor$", buildFlavor);
```
HernanGatta: Why is `Split('\\')[0]` necessary?

Also, if users pick their own dev kit, `board` will contain `"Other"` so the replacement of `OpteeBuildFlavor` will yield a path into the NuGet package that won't exist (in `optee\sub.mk`). Come to think of it, if users pick their own dev kit, they should also provide the OE libs, since those are potentially dev kit-specific. This would not have worked before this PR, so in that sense it's not a regression.

new_platforms/vsextension/ProjectWizard/WizardImplementation.cs
```diff
@@ -83,6 +83,13 @@ private bool SetOETADevKitPath(Dictionary<string, string> replacementsDictionary
...
+
+            string opteeCompilerFlavor = (buildFlavor == "vexpress-qemu_virt") ? "arm-linux-gnueabihf-" : "aarch64-linux-gnu-";
```
HernanGatta: This is also an issue if users pick their own dev kit. We can't assume which compiler it's correct for their board.

new_platforms/vsextension/ProjectWizard/WizardImplementation.cs
dthaler: the split is necessary because the actual strings look like "ls-ls1012grapeboard\\export-ta_arm64" and I only want the first part if there's other path segments.  The strings are in BoardPickerPage.cs starting at line 26.

dthaler: Good point on the "Other", will do that in a future changelist.

new_platforms/vsextension/ProjectWizard/WizardImplementation.cs
dthaler: true, but not a regression, at least by default we have a valid answer here (aarch64 being the default).  Will cover as part of future changelist for "Other" boards

new_platforms/vsextension/ProjectWizard/WizardImplementation.cs
HernanGatta: OK, I'll mark this PR as approved then.

dthaler:
> Perhaps we should disable allowing users to pick their own dev kit until we figure out how to do it correctly.

If it's easy, then disabling it is extra throwaway work.  So I'd rather do it correctly in a subsequent changelist, or disable it if it's found to be too hard.

PR #1796, opened 2019-05-16
Fix VS extension compiler options when user picks a custom ARM board

dthaler:

### DIFF
--- a/new_platforms/vsextension/ProjectWizard/WizardImplementation.cs
+++ b/new_platforms/vsextension/ProjectWizard/WizardImplementation.cs
@@ -6,6 +6,7 @@ using EnvDTE;
 using System.IO;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio;
 
 namespace OpenEnclaveSDK
 {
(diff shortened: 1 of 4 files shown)


new_platforms/vsextension/ProjectWizard/WizardImplementation.cs
```diff
@@ -33,6 +33,31 @@ public void RunFinished()
...
+                        continue;
+                    }
+
+                    return line.TrimEnd().EndsWith("y");
+                }
+                return false;
+            }
+            catch (IOException e)
+            {
+                return false;
```
HernanGatta: Is it possible to write out to the Output window a message saying that the extension failed to detect hard/soft float support? The rationale being that the if processor in question does indeed support hard float but we fail to detect it, performance via soft float will be worse than it could be.

dthaler: Done

new_platforms/vsextension/ProjectWizard/WizardImplementation.cs
```diff
@@ -72,22 +117,62 @@ private bool SetOETADevKitPath(Dictionary<string, string> replacementsDictionary
...
+                if (hardwareFloatSupported)
+                {
+                    opteeCompilerFlavor = "arm-linux-gnueabihf-";
+                }
+                else
+                {
+                    opteeCompilerFlavor = "arm-linux-gnueabi-";
 
-            string opteeCompilerFlavor = (buildFlavor == "vexpress-qemu_virt") ? "arm-linux-gnueabihf-" : "aarch64-linux-gnu-";
+                    // Output a warning so the developer knows they won't get hardware float support.
```
HernanGatta: Consider moving the warning only if we fail to detect hardware float support, on line 66. If the board being targeted does not have hardware float support, the user will get a false warning.

new_platforms/vsextension/ProjectWizard/WizardImplementation.cs
```diff
@@ -84,10 +109,35 @@ private bool SetOETADevKitPath(Dictionary<string, string> replacementsDictionary
...
+            // From this path, determine the compiler from the last path segment, and
+            // the build flavor from the next-to-last path segment.
+
+            string[] pathComponents = folder.Split('\\');
+            if (pathComponents.Length < 2)
+            {
+                return false;
+            }
+            string buildFlavor = pathComponents[pathComponents.Length - 2];
             replacementsDictionary.Add("$OpteeBuildFlavor$", buildFlavor);
```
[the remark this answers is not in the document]
dthaler: This comment is obsolete since the code is significantly different in the Friday PR

new_platforms/vsextension/ProjectWizard/WizardImplementation.cs
dthaler: There's no way to check whether the board has hardware float support, only whether the user has configured it in conf.mk, right?   That's why I didn't consider the warning to be "false".  I can move it if you still feel strongly.

dthaler: Updated per our discussion

PR #1799, opened 2019-05-16
changed e_report_header_t.report_size's type from size_t to uint64_t

soccerGB:
The changes in this PR address the following issue:

https://github.com/microsoft/openenclave/issues/1798

Basically, this change makes the size fixed for an OE SDK quote header.

### DIFF
--- a/enclave/core/sgx/report.c
+++ b/enclave/core/sgx/report.c
@@ -314,7 +314,7 @@ oe_result_t oe_get_report_v1(
     header->report_type = (flags & OE_REPORT_FLAGS_REMOTE_ATTESTATION)
                               ? OE_REPORT_TYPE_SGX_REMOTE
                               : OE_REPORT_TYPE_SGX_LOCAL;
-    header->report_size = (uint64_t)*report_buffer_size;
+    header->report_size = *report_buffer_size;
     OE_CHECK(oe_safe_add_u64(
         *report_buffer_size, sizeof(oe_report_header_t), report_buffer_size));
     result = OE_OK;
--- a/enclave/core/sgx/report.c
+++ b/enclave/core/sgx/report.c
@@ -314,7 +314,7 @@ oe_result_t oe_get_report_v1(
     header->report_type = (flags & OE_REPORT_FLAGS_REMOTE_ATTESTATION)
                               ? OE_REPORT_TYPE_SGX_REMOTE
                               : OE_REPORT_TYPE_SGX_LOCAL;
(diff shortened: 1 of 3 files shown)


CodeMonkeyLeet:
I believe @mikbras previously pushed to standardize on `size_t` for consistency with Linux APIs for buffer/structure sizing so he should review this.

It would probably be helpful to elaborate on the scenario this is intended to support. We currently do not support non-64-bit platforms and I'm pretty sure there are many place in the code where this has been previously assumed. Adding support for 32-bit platforms is a non-trivial decision at that point, and likely something we will need to do more comprehensively than this spot-fix.

dthaler:
32-bit support is required for TrustZone support for i.MX6.
All fields in protocol structs must be fixed-length, you should never use size_t inside any protocol struct, e.g., IP header, TCP header, whatever else.

enclave/core/sgx/report.c
```diff
@@ -314,7 +314,7 @@ oe_result_t oe_get_report_v1(
     header->report_type = (flags & OE_REPORT_FLAGS_REMOTE_ATTESTATION)
                               ? OE_REPORT_TYPE_SGX_REMOTE
                               : OE_REPORT_TYPE_SGX_LOCAL;
-    header->report_size = *report_buffer_size;
+    header->report_size = (uint64_t)*report_buffer_size;
```
mikbras: Optional: consider removing this cast. Either **sizeof(size_t)** <= **sizeof(uint64_t)**, in which case you don't need a cast, or **sizeof(size_t)** > **sizeof(uint64_t)**, in which case the cast is erroneous. The latter case is impossible on a 64-bit platform but will be a real problem with 128-bit platforms (future).

host/sgx/report.c
```diff
@@ -204,7 +204,7 @@ oe_result_t oe_get_report_v1(
     header->report_type = (flags & OE_REPORT_FLAGS_REMOTE_ATTESTATION)
                               ? OE_REPORT_TYPE_SGX_REMOTE
                               : OE_REPORT_TYPE_SGX_LOCAL;
-    header->report_size = *report_buffer_size;
+    header->report_size = (uint64_t)(*report_buffer_size);
```
mikbras: Same here.

enclave/core/sgx/report.c
soccerGB: Good point. I will remove this unnecessary casting.
Thanks for the great feedback.

PR #1813, opened 2019-05-22
Introduce OE_CRYPTO_ERROR code to replace OE_FAILURE in crypto wrappers

CodeMonkeyLeet:
- Add the new OE_CRYPTO_ERROR code to result.h and update existing OE crypto
  to return it instead of OE_FAILURE. This is intended to support debugging
  when failures in OE crypto wrappers are due to an underlying library error.
- Split sha.c into openssl & BCrypt-specific source files.
- Add BCrypt implementation for random.c.

### DIFF
--- a/host/crypto/bcrypt/random.c
+++ b/host/crypto/bcrypt/random.c
@@ -1,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <openenclave/internal/raise.h>
+#include <openenclave/internal/random.h>
+#include "bcrypt.h"
+
+oe_result_t oe_random_internal(void* data, size_t size)
+{
+    oe_result_t result = OE_UNEXPECTED;
+
+    /* For consistency with OpenSSL, keep this to int max */
+    if (size > OE_INT_MAX)
+        return OE_INVALID_PARAMETER;
+
+    NTSTATUS status = BCryptGenRandom(
+        NULL, data, (ULONG)size, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
+
+    if (!BCRYPT_SUCCESS(status))
+        OE_RAISE_MSG(
+            OE_CRYPTO_ERROR, "BcryptGenRandom failed (err=%#x)\n", status);
+
+    result = OE_OK;
+
+done:
+    return result;
+}
--- a/host/crypto/bcrypt/rsa.c
+++ b/host/crypto/bcrypt/rsa.c
@@ -81,7 +81,7 @@ static oe_result_t _rsa_pem_to_der(
 
     /* With a null buffer, CryptStringToA returns true and sets the size. */
     if (!success)
-        OE_RAISE(OE_FAILURE);
+        OE_RAISE(OE_CRYPTO_ERROR);
 
     der_local = (uint8_t*)malloc(der_local_size);
     if (der_local == NULL)
@@ -97,7 +97,7 @@ static oe_result_t _rsa_pem_to_der(
         NULL);
 
     if (!success)
-        OE_RAISE(OE_FAILURE);
+        OE_RAISE(OE_CRYPTO_ERROR);
 
     *der_data = der_local;
     *der_size = der_local_size;
@@ -146,7 +146,7 @@ oe_result_t oe_rsa_private_key_read_pem(
             &rsa_blob_size);
 
         if (!success)
-            OE_RAISE(OE_FAILURE);
+            OE_RAISE(OE_CRYPTO_ERROR);
     }
 
     /* Step 3: Convert the Crypt object to a BCrypt Key. */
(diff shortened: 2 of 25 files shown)


CodeMonkeyLeet:
@anitagov FYI, this pulls 28f4352c into the master branch directly.

host/crypto/bcrypt/rsa.c
```diff
@@ -161,7 +161,7 @@ oe_result_t oe_rsa_private_key_read_pem(
             0);
 
         if (status != STATUS_SUCCESS)
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Thanks for the catch, I needed a subsequent fix from the source branch. Fixed now.

host/crypto/bcrypt/sha.c
```diff
@@ -0,0 +1,83 @@
...
+#include <openenclave/internal/defs.h>
+#include <openenclave/internal/raise.h>
+#include <openenclave/internal/sha.h>
+#include <stdio.h>
+#include <string.h>
+#include "bcrypt.h"
+
+typedef struct _oe_sha256_context_impl
+{
+    BCRYPT_HASH_HANDLE handle;
```
anakrish: Do we need a _magic_ prefix here?
As done for
https://github.com/microsoft/openenclave/blob/bfd18e10c64c3287bbbf25238e594fc9797f4999/enclave/cert.c#L119-L124

host/crypto/bcrypt/random.c
```diff
@@ -0,0 +1,27 @@
...
+    /* For consistency with OpenSSL, keep this to int max */
+    if (size > OE_INT_MAX)
+        return OE_INVALID_PARAMETER;
+
+    NTSTATUS status = BCryptGenRandom(
+        NULL, data, (ULONG)size, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
+
+    if (!BCRYPT_SUCCESS(status))
+        OE_RAISE_MSG(
+            OE_CRYPTO_ERROR, "BcryptGenRandom failed (err=%#x)\n", status);
```
anakrish: Is %#x specifier better than the 0x%x specifiers used for other OE_CRYPTO_ERROR instances in say enclave/key.c?

host/crypto/openssl/rsa.c
```diff
@@ -490,35 +490,35 @@ oe_result_t oe_rsa_get_public_key_from_private(
...
     if (!(rsa_private = EVP_PKEY_get1_RSA(private_key_temp->pkey)))
-        OE_RAISE(OE_FAILURE);
+        OE_RAISE(OE_CRYPTO_ERROR);
 
     RSA_get0_key(rsa_private, &private_e, &private_n, NULL);
 
     /* Check if it's possible to get the public key. */
-    if (!private_n || !private_n)
-        OE_RAISE(OE_FAILURE);
+    if (!private_e || !private_n)
```
anakrish: Good catch. Is there some way for us to be confident that there are no such errors lurking around.
Especially in cryptographic code.
Would code-coverage be enough or do we need a more thorough testing strategy?

host/crypto/bcrypt/random.c
CodeMonkeyLeet: It's in the realm of personal taste. IIRC, `%#x` will omit the '0x' if the value is 0 which I find more readable. It also keeps the casing of the prefix consistent with the casing of the hex value.

host/crypto/openssl/rsa.c
CodeMonkeyLeet: Code coverage would be the strategy for general testing of implementation; in this example though, it's really a question of incorrect/inadequate return value validation and such things are usually caught through static analysis. I don't think there's a catch-all strategy here.

host/crypto/bcrypt/sha.c
CodeMonkeyLeet: I don't think so. The use of magic numbers is usually to allow for quick identification of variant types, but the use case introduced seems restricted to distinguishing host vs. enclave implementations* of potentially portable data types (e.g. keys, certs). Context objects (e.g. sha, hmac) in the existing implementation have not traditionally done this, I'm assuming because it's never been kosher to pass these kinds of things between computation boundaries anyway.

I'm not going update the structural use of magic numbers as part of this change set, in particular since @gupta-ak has already called for a broader evaluation of how we are structuring the crypto API in terms of:
* APIs defined for common use between host & enclave
* APIs that need to be implemented inside the enclave to satisfy the OE runtime. 
* APIs that need to be implemented for use in the host library/tools across OS platforms. 

I suspect we'll revisit how we deal with magic numbers as part of that discussion.

\* _Note for example that RSA uses the same magic numbers in both BCrypt & OpenSSL implementations today._

anakrish: Sounds good to me.

host/crypto/openssl/rsa.c
anakrish: Ok. It would be good if we can start investing in static-analysis on OE code base.

PR #1815, opened 2019-05-22
Make pkgconfig relocatable

### LINKED ISSUE #1654: Make pkg-config relocatable
In #1649 it was (re-)discovered that the pkg-config files are incompatible with `make DESTDIR` types of relocatable installations. While we have a work-around in place for the tests, this needs to be fixed.

oprinmarius:
Make pkgconfig relocatable

Fixes #1654


oprinmarius:
bors try

oprinmarius:
bors -

oprinmarius:
bors try-

oprinmarius:
bors try

oprinmarius:
bors try

andschwa:
@oprinmarius Can you explain how this made the pkg-config files relocatable?

oprinmarius:
@andschwa this does not make pkg-config relocatable. I changed the SYSROOT dir for the samples , since we are using code which is not installed in the default location.

PKG_CONFIG_SYSROOT_DIR
Modify -I and -L to use the directories located in target sysroot. this option is usefull when crosscompiling package that use pkg-config to determine CFLAGS anf LDFLAGS. -I and -L are modified to point to the new system root. this means that a -I/usr/include/libfoo will become -I/var/target/usr/include/libfoo with a PKG_CONFIG_SYSROOT_DIR equal to /var/target (same rule apply to -L)

I added the PKG_CONFIG_SYSROOT_DIR variable to the samples ENV.

I didn't manage to make the pkg-config files change whenever we use a different DESTDIR for the make install target. DESTDIR variable is not present in the environment at the moment we build the pkg-config files, and they won't change unless we change the CMAKE_INSTALL_PREFIX

I will sync with @aserdean  today to see if he can suggest a better solution to make the pkgconfig files relocatable.

aserdean:
@andschwa I'm confused about what `relocatable` means in this use case. Do we need to reconfigure the `*.pc` files as per user input or make the `*.pc` files themselves relocatable depending on `pcfiledir` i.e.:
https://github.com/aserdean/openenclave/commit/5c45923701377b05d900972e66e67bf275e7609b

andschwa:
@aserdean Right now, the CMake build configures the pkg-config files at _configure_ time, with hard-coded paths, dependency on `CMAKE_INSTALL_PREFIX`. This is fine is you only wish to install to that location after you've built, but we state we support using `make DESTDIR=foo install`, which happens at _build_ time, and right now that would place pkg-config files in the right location, but with bad paths in them. So instead of hard-coding paths in the pkg-config files at configuration time, we need them to work if they've been moved around (probably this means relative paths or logic similar to the CMake package, if that's possible). @anakrish might remember more specifics here.

aserdean:
@andschwa thanks so much for the explanation.

> So instead of hard-coding paths in the pkg-config files at configuration time, we need them to work if they've been moved around (probably this means relative paths or logic similar to the CMake package, if that's possible).

I think going through the route of `pcfiledir` would be better in this case then:
```
pcfiledir
The installed location of the .pc file. This can be used to query the location of the .pc file for a particular module, but it can also be used to make .pc files relocatable. For instance:
prefix=${pcfiledir}/../..
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include
```
since setting prefix to relative path based on `pcfiledir` should achieve what we are looking for.

From the commit I would previously pointing, this would be the most relevant [line](https://github.com/aserdean/openenclave/commit/5c45923701377b05d900972e66e67bf275e7609b#diff-f65dd8d9f9ac8774fb22a5cbd9b11c32R99) (FWIW commit has been tested if you want to give it a spin).

Please let us know if you agree or if you want us to explore a different option.

andschwa:
@aserdean that sounds great to me! I will defer to you here, as I don't know much about pkg-config (other than our implementation was broken 😆 ) Thank you!

oprinmarius:
bors try

Issue #1777, opened 2019-05-10
Log ECALLs/OCALLs


letmaik:
It would be very useful to optionally log all ECALLs/OCALLs for debug enclaves using the existing logging mechanism. I could imagine that a new log level would be useful for this, e.g. `OE_LOG_LEVEL=VERBOSE`.

For this to be useful, e.g. for security analysis via automatic checks of logs during development, the function name and, where useful, arguments have to be included in the log message. I imagine that this is known for all OE-implicit OCALLs (e.g. printf), but there may be more metadata necessary for any EDL-defined functions, however this should be doable.

Which pieces in OE would have to be touched to make this happen?

anakrish:
@soccerGB @andschwa would be the right folks to follow up on this.

johnkord:
This may be able to be done with code generated from EDL, but let's discuss the best way to implement this.

jxyang:
My intuition is to put function names in ECALL/OCALL function tables that are populated by edger8r. Would be it a potential security vulnerability?

letmaik:
EDL is just one area. This also applies to all implicit OCALLs that OE does, e.g. when printing to stdout.

achamayou:
This would be particularly useful for "implicit" OCALLs, not specified in the EDL.  To elaborate on the use case, we anticipate situations where users building applications will import libraries that are not written with enclaves in mind, and which occasionally may contain code that results in an implicity OCALL (such as writing to stdout).
Even with full initial audit, this could slip in during an upgrade etc.

This is neither good for performance nor for security, and so it would be useful to at least log them, so they can be caught in testing. Even better would be a mechanism to forbid it entirely, but logging will still be useful to find out where the calls come from.

@johnkord if you can give us some pointers on where implicit ocalls are generated, we'd be happy to PR this.

jxyang:
@achamayou  The OCALLs are processed in `_handle_ocall`. The user defined host functions (in the .edl file) are handled by `_handle_call_host_function`, the implicit OCALLs are handled in the switch statement case by case. The supported implicit OCALLs are defined in `enum _oe_func`, for example:

```c
typedef enum _oe_func
{
    OE_OCALL_GET_QE_TARGET_INFO,
    OE_OCALL_GET_QUOTE,
    OE_OCALL_GET_REVOCATION_INFO,
    OE_OCALL_GET_QE_ID_INFO,
    ...
}
```
[...]

PR #1816, opened 2019-05-22
Log ocalls

achamayou:
As discussed in #1777, this would be helpful when trying to understand what OCalls happen and when.

### DIFF
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -432,17 +432,11 @@ static oe_result_t _handle_ocall(
         *arg_out = 0;
 
     if (func != OE_OCALL_LOG)
-    {
-        oe_log_args_t log_args;
-        log_args.level = OE_LOG_LEVEL_INFO;
-        snprintf(
-            log_args.message,
-            OE_LOG_MESSAGE_LEN_MAX,
-            "Enclave %s OE_OCALL: %s\n",
+        oe_log(
+            OE_LOG_LEVEL_INFO,
+            "%s OE_OCALL: %s\n",
             enclave->path,
             oe_ocall_str(func));
-        oe_handle_log(enclave, (uint64_t)&log_args);
-    }
 
     switch ((oe_func_t)func)
     {
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -438,7 +438,8 @@ static oe_result_t _handle_ocall(
         snprintf(
             log_args.message,
             OE_LOG_MESSAGE_LEN_MAX,
-            "OE_OCALL: %s\n",
+            "Enclave %s OE_OCALL: %s\n",
+            enclave->path,
             oe_ocall_str(func));
         oe_handle_log(enclave, (uint64_t)&log_args);
     }
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -415,12 +415,11 @@ static oe_result_t _handle_ocall(
     if (arg_out)
         *arg_out = 0;
 
-    if (func != OE_OCALL_LOG)
-        oe_log(
-            OE_LOG_LEVEL_INFO,
-            "%s OE_OCALL: %s\n",
-            enclave->path,
-            oe_ocall_str(func));
+    oe_log(
+        OE_LOG_LEVEL_INFO,
+        "%s OE_OCALL: %s\n",
+        enclave->path,
+        oe_ocall_str(func));
 
     switch ((oe_func_t)func)
     {
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -365,9 +365,7 @@ done:
     return result;
 }
 
-const char* oe_ocall_str(oe_func_t ocall);
-
-const char* oe_ocall_str(oe_func_t ocall)
+static const char* oe_ocall_str(oe_func_t ocall)
 {
     switch (ocall)
     {
@@ -405,7 +403,7 @@ const char* oe_ocall_str(oe_func_t ocall)
             return "OE_OCALL_LOG";
         default:
             return "UNKNOWN";
-     }
+    }
 };
 
 /*
@@ -437,7 +435,11 @@ static oe_result_t _handle_ocall(
     {
         oe_log_args_t log_args;
         log_args.level = OE_LOG_LEVEL_INFO;
-        snprintf(log_args.message, OE_LOG_MESSAGE_LEN_MAX, "OE_OCALL: %s\n", oe_ocall_str(func));
+        snprintf(
+            log_args.message,
+            OE_LOG_MESSAGE_LEN_MAX,
+            "OE_OCALL: %s\n",
+            oe_ocall_str(func));
         oe_handle_log(enclave, (uint64_t)&log_args);
     }
 
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -367,43 +367,27 @@ done:
 
 static const char* oe_ocall_str(oe_func_t ocall)
 {
-    switch (ocall)
-    {
-        case OE_OCALL_CALL_HOST_FUNCTION:
-            return "OE_OCALL_CALL_HOST_FUNCTION";
-        case OE_OCALL_GET_QE_TARGET_INFO:
-            return "OE_OCALL_GET_QE_TARGET_INFO";
-        case OE_OCALL_GET_QUOTE:
-            return "OE_OCALL_GET_QUOTE";
-        case OE_OCALL_GET_REVOCATION_INFO:
-            return "OE_OCALL_GET_REVOCATION_INFO";
-        case OE_OCALL_GET_QE_ID_INFO:
-            return "OE_OCALL_GET_QE_ID_INFO";
-        case OE_OCALL_THREAD_WAKE:
-            return "OE_OCALL_THREAD_WAKE";
-        case OE_OCALL_THREAD_WAIT:
-            return "OE_OCALL_THREAD_WAIT";
-        case OE_OCALL_THREAD_WAKE_WAIT:
-            return "OE_OCALL_THREAD_WAKE_WAIT";
-        case OE_OCALL_MALLOC:
-            return "OE_OCALL_MALLOC";
-        case OE_OCALL_REALLOC:
-            return "OE_OCALL_REALLOC";
-        case OE_OCALL_FREE:
-            return "OE_OCALL_FREE";
-        case OE_OCALL_WRITE:
-            return "OE_OCALL_WRITE";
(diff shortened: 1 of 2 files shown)


host/sgx/calls.c
```diff
@@ -390,6 +431,19 @@ static oe_result_t _handle_ocall(
     if (arg_out)
         *arg_out = 0;
 
+    if (func != OE_OCALL_LOG)
+    {
+        oe_log_args_t log_args;
```
letmaik: Why not `oe_log`? Isn't that the more high-level function to use?

host/sgx/calls.c
```diff
@@ -365,6 +365,47 @@ static oe_result_t _handle_call_host_function(
     return result;
 }
 
+static const char* oe_ocall_str(oe_func_t ocall)
+{
+    switch (ocall)
```
jxyang: We can simplify here with a string array, e.g.:
```
    static const char* func_names[] = 
    {
        "OE_OCALL_CALL_HOST_FUNCTION",
        "OE_OCALL_GET_QE_TARGET_INFO",
        ...
    };
     if (ocall < __OE_FUNC_MAX)
       return func_names[ocall - OE_OCALL_BASE];
```
This way future OCALLs can be more easily added.

host/sgx/calls.c
```diff
@@ -390,6 +431,19 @@ static oe_result_t _handle_ocall(
     if (arg_out)
         *arg_out = 0;
 
+    if (func != OE_OCALL_LOG)
```
jxyang: Why don't we log OE_OCALL_LOG calls?

letmaik: I'm assuming endless recursive loop. You shouldn't log something in your log function.

letmaik:
> @letmaik @achamayou Are you also interested in logging AEP exits from the enclave as well. Say the 3rd part code invokes CPUID instruction and this technically transfers control to the (untrusted) host.
> Would you be interested in logging that as well?

@anakrish Good point, I doubt it's high on the priority list (right, @achamayou?), so we should probably do that in a separate PR. In principle I'm in favour of logging these events as well.

host/sgx/calls.c
achamayou: No reason other than my lack of familiarity with the codebase, will use oe_log

host/sgx/calls.c
achamayou: Ok, will do. I’d deliberately copied what’s there for oe_result_t to be consistent but I agree a string array is more elegant.

host/sgx/calls.c
achamayou: There is no recursive loop here because this is on the host side, it won’t cause another OCall.
It’s two lines for each log line though, it seemed a bit excessive, but I don’t feel strongly either way, I am happy to put it back if we think it’s preferable.

host/sgx/calls.c
```diff
@@ -365,6 +365,47 @@ static oe_result_t _handle_call_host_function(
...
+            return "OE_OCALL_WRITE";
+        case OE_OCALL_SLEEP:
+            return "OE_OCALL_SLEEP";
+        case OE_OCALL_GET_TIME:
+            return "OE_OCALL_GET_TIME";
+        case OE_OCALL_BACKTRACE_SYMBOLS:
+            return "OE_OCALL_BACKTRACE_SYMBOLS";
+        case OE_OCALL_LOG:
+            return "OE_OCALL_LOG";
+        default:
```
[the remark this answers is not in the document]
achamayou: It depends on the flags we pass. I need to check.

achamayou:
@letmaik @anakrish yes that does sound useful too, if you have any pointers to where that is done, I will gladly look into that.

I will address the feedback tomorrow morning.

anakrish:
@achamayou 
The following function is where the host traps the signal (linux) or structured-exception(windows) resultant from an Asynchronous Exit from an enclave:

https://github.com/microsoft/openenclave/blob/ed417f47f5d27f0f48b9c822e77f56af659d46cc/host/sgx/exception.c#L21

You'd have to log before the  control is transferred back to the enclave in line 50:
https://github.com/microsoft/openenclave/blob/ed417f47f5d27f0f48b9c822e77f56af659d46cc/host/sgx/exception.c#L50-L51 

Note that by design, not much about the exception is known to the host.

host/sgx/calls.c
achamayou: Done

host/sgx/calls.c
achamayou: @jxyang how do you feel about this? I'm happy either way.

host/sgx/calls.c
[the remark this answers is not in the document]
achamayou: @anakrish so the answer is "not currently, no".  We would need to set -Wswitch-enum https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html. But that's irrelevant once this is turned into a string array lookup anyway,

host/sgx/calls.c
achamayou: Done

host/sgx/calls.c
```diff
@@ -390,6 +431,19 @@ static oe_result_t _handle_ocall(
...
 
+    if (func != OE_OCALL_LOG)
+    {
+        oe_log_args_t log_args;
+        log_args.level = OE_LOG_LEVEL_INFO;
+        snprintf(
+            log_args.message,
+            OE_LOG_MESSAGE_LEN_MAX,
+            "Enclave %s OE_OCALL: %s\n",
+            enclave->path,
```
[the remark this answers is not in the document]
achamayou: @anakrish not for our purposes no, but let me know if you want something else in there (the base address?)

achamayou:
@anakrish except for the question of where we should log the logging calls themselves, I believe all feedback has been addressed now.

Having had a look at exception.c, I agree that there is so little information available about the exception and what caused it that it probably does not make a lot of sense to log it.

host/sgx/calls.c
jxyang: We should cover OE_OCALL_LOG as well for the sake of completeness. Like @achamayou said, there is no risk of recursive loop.

achamayou: Done

host/sgx/calls.c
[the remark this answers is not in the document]
anakrish: The base address would work for now.

anakrish:
> @anakrish except for the question of where we should log the logging calls themselves, I believe all feedback has been addressed now.
> 
> Having had a look at exception.c, I agree that there is so little information available about the exception and what caused it that it probably does not make a lot of sense to log it.
@achamayou Sounds good.

host/sgx/calls.c
```diff
@@ -365,6 +365,31 @@ static oe_result_t _handle_call_host_function(
...
+                                       "OE_OCALL_MALLOC",
+                                       "OE_OCALL_REALLOC",
+                                       "OE_OCALL_FREE",
+                                       "OE_OCALL_WRITE",
+                                       "OE_OCALL_SLEEP",
+                                       "OE_OCALL_GET_TIME",
+                                       "OE_OCALL_BACKTRACE_SYMBOLS"
+                                       "OE_OCALL_LOG"};
+
+    if (ocall >= OE_OCALL_BASE && ocall < __OE_FUNC_MAX)
```
anakrish: As written, the compiler does not know the relationship between OE_COUNTOF(func_names) and __OE_FUNC_MAX. We could add the following assertion:
OE_STATIC_ASSERT(OE_OCALL_BASE + OE_COUNTOF(func_names) == __OE_FUNC_MAX);

jxyang: That's a nice way to catch we add some OCALL in the enum but not in func_names.

jxyang:
Other than @anakrish's suggestion of adding an assertion, it looks good to me.

host/sgx/calls.c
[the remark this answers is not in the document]
anakrish: @jxyang Can you please investigate if we could add -Wswitch-enum to the set of flags that we compile the OE SDK with? That would be a nice way to get into our cmake build system.

host/sgx/calls.c
achamayou: Sounds good, will add!

host/sgx/calls.c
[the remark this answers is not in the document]
jxyang: I tried that. The thing is, we have many switch statements which don't handle the last enum value (the MAX), for example `__OE_EC_TYPE_MAX`. It will spur quite a few compiler errors if we add this flag.

host/sgx/calls.c
achamayou: So that won't work, because __OE_FUNC_MAX = OE_ENUM_MAX, which is 0xffffffff. I was a bit puzzled by that and thought this might be a mistake, but that seems to be done consistently across enums. I don't understand why, I would have thought having the max of each actual enum would be preferable.

Anyway, let me know if you want me to do something else instead, I cannot think of a good alternative without removing the assignment on the max value.

host/sgx/calls.c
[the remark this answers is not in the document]
achamayou: Done

achamayou:
@anakrish @jxyang added the base addr, but as described in the relevant comment, the additional assertion cannot work.

achamayou:
bors try

host/sgx/calls.c
jxyang: It looks like @mikbras made all enums to be 4 bytes by forcing the MAX values to be 0xffffffff in [this commit](https://github.com/microsoft/openenclave/commit/480b29ad6c6a1a918d6c8808ecdd2738f1bdbba2). Please check with him what was the motivation and whether we could make an exception for `__OE_FUNC_MAX`.

achamayou:
bors try

host/sgx/calls.c
anakrish: @jxyang The motivation was to ensure a consistent integer size for all enums rather than have the compiler pick the integer type.

@achamayou 
We could add another enum value OE_OCALL_MAX that contains the desired value and use that in the assertion instead.

achamayou: @anakrish Done

achamayou:
bors try

achamayou:
@anakrish the Windows cross compile tests seem to suffer from OE_PLATFORM_ERROR, I suspect this is transient and will kick off the CI again if that's ok.

achamayou:
bors try

anakrish:
@achamayou Yes that is a transient error.

jxyang:
Thanks for the PR, @achamayou

PR #1575, opened 2019-03-16
Enable host side quote verification on non-SGX machines

gupta-ak:
To build the `open-enclave-host` package:
1) build with `USE_LIBSGX=OFF` and the debian package generator, (i.e. `cmake .. -DUSE_LIBSGX=OFF -DCPACK_GENERATOR=DEB`)
2) compile the code
3) make the package with `cpack -D CPACK_DEB_COMPONENT_INSTALL=ON -D CPACK_COMPONENTS_ALL=OEQUOTE`

PR #1818, opened 2019-05-23
Remove OE_USE_LIBSGX on verifying report.

gupta-ak:
This allows remote quote verification without `OE_USE_LIBSGX`. It's essentially a piece of PR #1575 to unblock Cheng-mean's PR #1761.

A way to test this is to run the report test with a pre-generated report. That test should now pass.

### DIFF
--- a/common/sgx/qeidentity.c
+++ b/common/sgx/qeidentity.c
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#ifdef OE_USE_LIBSGX
 #include "qeidentity.h"
 #include <openenclave/internal/raise.h>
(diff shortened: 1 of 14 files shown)


gupta-ak:
bors try

gupta-ak:
> 
> 
> Looks good to me. Gets rid of unnecessary code.
> Is there a test that locks down the specific scenario that we are trying to support?

The end goal is a library that can verify quotes on any non-SGX machine. This is a highly requested feature from different customers. It's hard to automate a test for this since you need to generate a report with a SGX machine and then validate it on a non-SGX machine.

gupta-ak:
> 
> 
> Looks good to me. Gets rid of unnecessary code.
> Is there a test that locks down the specific scenario that we are trying to support?

Filed issue #1819 for this.

PR #1786, opened 2019-05-14
Update libunwind to v1.3-stable

HernanGatta:
This PR updates `libunwind` to version `v1.3-stable`.

This is the first in a series of upcoming PR's to merge ARM TrustZone support into `master`. Currently, ARM TrustZone support lives in the `feature.new_platforms` branch within a re-implementation of the SDK in the `new_platforms` folder. The intent of this series of PR's is to integrate that functionality alongside support for Intel SGX in a single, cohesive SDK.

Version `v1.3-stable` of `libunwind` properly supports AArch64, whereas the version currently in `master` does not. This PR only updates `libunwind` and makes no changes specific to ARM TrustZone. Those will be included in a subsequent PR.

PR #1820, opened 2019-05-23
Generate OP-TEE Trusted Applications (TAs)

HernanGatta:
This PR contains an initial set of changes to begin moving ARM TrustZone support into `master` as well as additional modifications to CI to ensure that the SDK continues to build for `AArch64`. The changes introduced in this PR will be used in subsequent PR's as building blocks. As such, the changes ought not to be considered final.

The major contribution in this PR is the ability to create proper OP-TEE OS-compatible Trusted Application (TA) binaries (i.e., the equivalent of Intel SGX enclaves). The resulting binaries have extension `.ta`. Producing TA binaries unblocks beginning work on the host-side API's, which in turn will unblock further development of the enclave-side once TA's can load and begin execution.

**Details**

This PR:
1. Updates the AArch64 build step in Jenkins to:
1.1. Download pre-built TA Dev Kits from Azure blob storage;
1.2. Configure the build accordingly.
2. Updates Ansible configuration to install additional dependencies needed to build and sign TA binaries;
3. Adds Microsoft's fork of OP-TEE OS a submodule;
4. Adds logic to build `libutee`, which contains a user-mode interface for OP-TEE OS system calls;
5. Adds logic to build TA binaries as required by OP-TEE OS (CMake-only for now):
5.1. Updates existing targets to use UUIDs.
6. Adds initial bits of OP-TEE OS support into `oecore`, `oeenclave` and `oelibc`.

*Updates*

7. Adds stubs for concurrency functions (spinlocks, mutexes, etc.):
7.1. Note that OP-TEE supports neither concurrency nor reentrancy in user-mode TAs.
8. Adds stubs for functions required by the GlobalPlatform specifications implemented by OP-TEE OS (`TA_*EntryPoint`).

**Note:** This does *not* depend on #1786.


.jenkins/Jenkinsfile
```diff
@@ -58,9 +58,16 @@ def AArch64GNUTest(String version, String build_type) {
             cleanWs()
             checkout scm
             def task = """
-                        cmake ${WORKSPACE} -G Ninja -DCMAKE_TOOLCHAIN_FILE=${WORKSPACE}/cmake/arm-cross.cmake -DCMAKE_BUILD_TYPE=${build_type} -DUSE_LIBSGX=OFF -Wdev
-                        ninja -v
-                        """
+                       curl ${DEVKITS_URI} | tar xvz -C ${WORKSPACE} --no-same-permissions
```
andschwa: This is liable to fail builds when the network has issues. Can we cache whatever exists at this URI?

3rdparty/optee/libutee/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+  ${OE_TZ_LIBUTEE_SRC}/tee_api_objects.c
+  ${OE_TZ_LIBUTEE_SRC}/tee_api_operations.c
+  ${OE_TZ_LIBUTEE_SRC}/tee_tcpudp_socket.c
+  ${OE_TZ_LIBUTEE_SRC}/tee_socket_pta.c
+  ${OE_TZ_LIBUTEE_SRC}/arch/arm/user_ta_entry.c
+  ${OE_TZ_LIBUTILS_EXT_SRC}/snprintk.c
+  PROPERTIES
+    COMPILE_FLAGS -Wno-conversion)
+
+set_property(TARGET oeutee PROPERTY
```
andschwa: I know this is still common throughout the build, but it's unnecessary and should be removed when files are being modified (and not added in new files).

3rdparty/optee/libutee/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+set(LIBUTEE_S_SRC
+  ${OE_TZ_LIBUTEE_SRC}/arch/arm/utee_syscalls_a64.S
```
andschwa: Where's `OE_TZ_LIBUTEE_SRC` set and whey do we need to reference it over and over, can we just use a short relative path maybe?

3rdparty/optee/libutee/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+  ${OE_TZ_LIBUTEE_SRC}/tee_api_se.c
+  ${OE_TZ_LIBUTEE_SRC}/tee_api_panic.c
+  ${OE_TZ_LIBUTEE_SRC}/tee_tcpudp_socket.c
+  ${OE_TZ_LIBUTEE_SRC}/tee_socket_pta.c
+  ${OE_TZ_LIBUTEE_SRC}/arch/arm/user_ta_entry.c
+  ${OE_TZ_LIBUTEE_SRC}/arch/arm/utee_misc.c
+  ${OE_TZ_LIBUTILS_EXT_SRC}/snprintk.c
+  ${OE_TZ_LIBUTILS_EXT_SRC}/buf_compare_ct.c)
+
+add_library(oeuteeasm OBJECT ${LIBUTEE_S_SRC})
```
andschwa: I would encourage just defining the sources directly in the two `add_library` calls, otherwise a reader has to pick up on the two different variables and libraries where the difference is only `S` vs `C` (and that's easy to miss at a glance 😆).

3rdparty/optee/libutee/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+target_include_directories(oeuteeasm PRIVATE
+  ${OE_TZ_LIBUTEE_INC}
+  ${OE_TZ_LIBUTILS_EXT_INC})
+
+add_library(oeutee OBJECT ${LIBUTEE_C_SRC})
+set_property(TARGET oeutee PROPERTY C_STANDARD 99)
+target_compile_options(oeutee PUBLIC
+  -include ${OE_TZ_TA_DEV_KIT_CONF})
+target_include_directories(oeutee PRIVATE
+  ${OE_TZ_LIBUTEE_SRC}
```
andschwa: Does `OE_TZ_LIBUTEE_SRC` really need to be added to the includes? Usually if the includes use quotes, we don't have to explicitly add it to the list as the including file's directory is checked first by default.

3rdparty/optee/libutee/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+  ${OE_TZ_LIBUTEE_SRC}/base64.c
+  ${OE_TZ_LIBUTEE_SRC}/tee_api.c
+  ${OE_TZ_LIBUTEE_SRC}/tee_api_objects.c
+  ${OE_TZ_LIBUTEE_SRC}/tee_api_operations.c
+  ${OE_TZ_LIBUTEE_SRC}/tee_tcpudp_socket.c
+  ${OE_TZ_LIBUTEE_SRC}/tee_socket_pta.c
+  ${OE_TZ_LIBUTEE_SRC}/arch/arm/user_ta_entry.c
+  ${OE_TZ_LIBUTILS_EXT_SRC}/snprintk.c
+  PROPERTIES
+    COMPILE_FLAGS -Wno-conversion)
```
andschwa: 😬

cmake/add_enclave.cmake
```diff
@@ -1,50 +1,96 @@
...
+# that layout is indeed followed, adding a `<target>.ta` target to generate
+# the final TA binary. Additionally, TA binaries must be named with a UUID.
+# As such, while the CMake target names remain in sync with <target>, the
+# resulting binaries use <uuid> plus their corresponding extension as their
+# on-disk name.
+#
 # The target is always linked to `oeenclave`, and if the optional flag
 # `CXX` is passed, it is also linked to `oelibcxx`
 #
+# NOTE: This must be a macro! To generate TA binaries, a custom linker
```
andschwa: Thanks for documenting and explaining this.

cmake/add_enclave.cmake
```diff
@@ -101,3 +147,91 @@ function(add_enclave)
...
+     "${multiValueArgs}"
+     ${ARGN})
+
+  # Set up the linker flags exactly as we need them such that the resulting
+  # binary be compatible with OP-TEE's loader.
+  set(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS)
+  set(CMAKE_SHARED_LIBRARY_LINK_CXX_FLAGS)
+  set(CMAKE_EXE_EXPORTS_C_FLAG)
+
+  string(REPLACE "gcc" "ld" LINKER ${CMAKE_C_COMPILER})
```
andschwa: Seems easy to break...

cmake/add_enclave.cmake
```diff
@@ -101,3 +147,91 @@ function(add_enclave)
...
+  set(CMAKE_C_LINK_EXECUTABLE "${LINKER} <CMAKE_CXX_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> <LINK_LIBRARIES> -o <TARGET>")
+  set(CMAKE_CXX_LINK_EXECUTABLE "${LINKER} <CMAKE_CXX_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> <LINK_LIBRARIES> -o <TARGET>")
+
+  # Generate linker script from template.
+  string(REPLACE "gcc" "cpp" C_PREPROCESSOR ${CMAKE_C_COMPILER})
+  set(TA_LINKER_SCRIPT ${CMAKE_CURRENT_BINARY_DIR}/ta.ld)
+  set(TA_LINKER_SCRIPT ${TA_LINKER_SCRIPT} PARENT_SCOPE)
+  add_custom_target(${ENCLAVE_TARGET}.ld
```
andschwa: Should their be a `DEPENDS` for this (and the other custom targets)? I suspect `SOURCES` was used with the intention of setting up a dependency, but thanks to CMake being CMake, this actually only has the side effect of grouping the files for IDEs parsing CMake projects. `DEPENDS` is what will actually set the dependencies.

cmake/add_enclave.cmake
```diff
@@ -101,3 +147,91 @@ function(add_enclave)
...
+        --strip-unneeded $<TARGET_FILE:${ENCLAVE_TARGET}>
+        $<TARGET_FILE_DIR:${ENCLAVE_TARGET}>/${ENCLAVE_UUID}.stripped.elf
+    BYPRODUCTS $<TARGET_FILE_DIR:${ENCLAVE_TARGET}>/${ENCLAVE_UUID}.stripped.elf)
+  add_dependencies(${ENCLAVE_TARGET}.stripped.elf ${ENCLAVE_TARGET})
+
+  # Sign the TA with the given key, or with the default key if none was given.
+  if(NOT ENCLAVE_KEY)
+    set(ENCLAVE_KEY ${OE_TZ_TA_DEV_KIT_DEFAULT_SIGNING_KEY})
+  endif()
+  add_custom_target(${ENCLAVE_TARGET}.ta ALL
```
andschwa: Should you have a dependency on `OE_TZ_TA_DEV_KIT_SIGN_TOOL` or does that variable resolve to an executable target in the CMake dependency graph?

cmake/add_enclave.cmake
```diff
@@ -101,3 +147,91 @@ function(add_enclave)
...
+        --uuid ${ENCLAVE_UUID}
+        --version 0
+        --in $<TARGET_FILE_DIR:${ENCLAVE_TARGET}>/${ENCLAVE_UUID}.stripped.elf
+        --out $<TARGET_FILE_DIR:${ENCLAVE_TARGET}>/${ENCLAVE_UUID}.ta
+    BYPRODUCTS $<TARGET_FILE_DIR:${ENCLAVE_TARGET}>/${ENCLAVE_UUID}.ta)
+  add_dependencies(${ENCLAVE_TARGET}.ta ${ENCLAVE_TARGET}.stripped.elf)
+
+  # Set linker options.
+  # NOTE: This has to be at the end, apparently:
+  #       https://gitlab.kitware.com/cmake/cmake/issues/17210
```
andschwa: 😆

cmake/arm-cross.cmake
```diff
@@ -6,3 +6,71 @@ set(CMAKE_SYSTEM_PROCESSOR aarch64)
...
+    -nostartfiles
+    -fno-builtin-memcpy
+    -fno-builtin-memset
+    -ffreestanding
+    -funwind-tables
+    -fpie
+    -fPIC
+    -gdwarf)
+
+string(REPLACE ";"
```
andschwa: Maybe leave a comment as to what we're doing here for non-CMake people (converting a CMake list variable to a string variable for direct use by a compilation tool, also, I think I have an example of how to do this without `REPLACE` somewhere that I'll try to dig up...)

cmake/add_enclave.cmake
```diff
@@ -101,3 +147,91 @@ function(add_enclave)
...
+   set(oneValueArgs TARGET UUID KEY)
+   set(multiValueArgs SOURCES)
+   cmake_parse_arguments(ENCLAVE
+     "${options}"
+     "${oneValueArgs}"
+     "${multiValueArgs}"
+     ${ARGN})
+
+  # Set up the linker flags exactly as we need them such that the resulting
+  # binary be compatible with OP-TEE's loader.
```
dthaler: fix grammar ("be" -> "is")

cmake/arm-cross.cmake
```diff
@@ -6,3 +6,71 @@ set(CMAKE_SYSTEM_PROCESSOR aarch64)
...
+
+# Path to required bits of the TA Dev Kit.
+set(OE_TZ_TA_DEV_KIT_CONF                   ${OE_TA_DEV_KIT_DIR}/host_include/conf.h)
+set(OE_TZ_TA_DEV_KIT_LINKER_SCRIPT_TEMPLATE ${OE_TA_DEV_KIT_DIR}/src/ta.ld.S)
+set(OE_TZ_TA_DEV_KIT_HEADER_SOURCE          ${OE_TA_DEV_KIT_DIR}/src/user_ta_header.c)
+set(OE_TZ_TA_DEV_KIT_DEFAULT_SIGNING_KEY    ${OE_TA_DEV_KIT_DIR}/keys/default_ta.pem)
+set(OE_TZ_TA_DEV_KIT_SIGN_TOOL              ${OE_TA_DEV_KIT_DIR}/scripts/sign.py)
+
+# Path to OP-TEE OS.
+set(OE_TZ_OPTEE_SRC ${PROJECT_SOURCE_DIR}/3rdparty/optee/optee_os)
```
andschwa: Hey I found it.

3rdparty/optee/libutee/CMakeLists.txt
andschwa: (Based on where I found it, looks like this _could_ just be `optee/os` here.)

libc/optee/abort.c
```diff
@@ -3,4 +3,6 @@
 
 void abort(void)
 {
+    while (1)
```
dthaler: won't this cause a hang instead of an abort?

libc/optee/stack.c
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+void* __stack_chk_guard = (void*)0x00000aff;
+
+void __attribute__((noreturn)) __stack_chk_fail(void);
+
+void __stack_chk_fail(void)
+{
+    while (1)
```
dthaler: Why hang?

tests/SampleApp/enc/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 
 oeedl_file(../SampleApp.edl enclave gen)
 
-add_enclave(TARGET SampleApp SOURCES SampleApp.cpp ${gen})
+add_enclave(TARGET SampleApp UUID 25419627-14f6-4625-9329-cf5f10a57fea SOURCES SampleApp.cpp ${gen})
```
andschwa: Can we leave an explanation in a local readme as to what the UUID here is for?

libc/optee/sysconf.c
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <errno.h>
+#include <openenclave/enclave.h>
+
+long sysconf(int name)
+{
+    OE_UNUSED(name);
+    return 8;
```
dthaler: What does the number "8" mean here?  (add a comment)

.jenkins/Jenkinsfile
HernanGatta: I'm admittedly not familiar with Jenkins. If I were to download this file outside of the container as part of the Jenkins step, will it remain on the system for subsequent runs? If so, I could download and extract the file only if it does not yet exist, then share it into the container.

Did you have something else in mind?

3rdparty/optee/libutee/CMakeLists.txt
HernanGatta:
```
OE_TZ_OPTEE_SRC = ${PROJECT_SOURCE_DIR}/3rdparty/optee/optee_os
OE_TZ_LIBUTEE_SRC = ${OE_TZ_OPTEE_SRC}/lib/libutee

=> OE_TZ_LIBUTEE_SRC = <repo root>/3rdparty/optee/optee_os/lib/libutee
```
I followed the pattern here [this](https://github.com/microsoft/openenclave/blob/master/libc/CMakeLists.txt#L79).

cmake/add_enclave.cmake
HernanGatta: No problem!

cmake/add_enclave.cmake
HernanGatta: It's the subjunctive mood!

cmake/add_enclave.cmake
HernanGatta: If we ever wish to build TA's with a toolchain other than GCC, then yes. I have yet to see anyone try to build TA's with clang, so I would argue that this solution is sufficient in the short-term, seeing as GCC binaries follow the pattern `<toolchain-prefix>-<tool-name>`.

cmake/add_enclave.cmake
HernanGatta: `OE_TZ_TA_DEV_KIT_SIGN_TOOL` is a Python script in the OP-TEE OS submodule; it's always there.

libc/optee/abort.c
HernanGatta: Yes. Now that TA's are being built in master, the plan is to add the host-side functionality to load TA's from Linux. Once TA's load, I'll add the remaining missing functionality inside the TA. I've added an infinite loop here so that I can catch it during debugging if it ever hits. The proper implementation will be to call into OP-TEE's abort handler.

libc/optee/stack.c
HernanGatta: Same as with `abort()`.

tests/SampleApp/enc/CMakeLists.txt
HernanGatta: Sure. We'll have to overhaul the documentation in any case to cover TrustZone, so this will be in that changeset, once we reach that point.

3rdparty/optee/libutee/CMakeLists.txt
andschwa: Ah that makes sense. That pattern is the way it is because the CMakeLists.txt is in `src/libc` while the the 3rdparty code is `../../3rdparty/musl/musl`.

Since this file exists at `3rdparty/optee/libutee/CMakeLists.txt` and the code exists at `3rdparty/optee/optee_os/lib/libutee`...you're fine. Looking at those two paths you'd have to `../optee_os/lib/libutee` and that's just as annoying.

.jenkins/Jenkinsfile
andschwa: Roping in @aserdean here...

andschwa: And @johnkord

libc/optee/abort.c
dthaler: Please add a TODO comment inline

libc/optee/stack.c
dthaler: Please add a TODO comment inline

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "3rdparty/optee/optee_os"]
```
andschwa: Oh I skipped right past this. We have so far avoided adding submodules to OE for several very good reasons.

1) They make using Git much more difficult for those not familiar with submodules.
2) They introduce a network based dependency for every "clean" build (like done from CI), which can cause CI failures.
3) Git uses SHA1 for its hashes, and since SHA1 is known to be quite broken, there's no actual security assurance that the submodule checkout is exactly the code the developers said it should be.
4) It adds another location to monitor for security breaches etc.

So we have opted for all other 3rdparty code to commit and redistribute (with appropriate notices of course) the code directly within our repo. Obviously #3 still applies but since it's coming from a single trusted location, it's easier to be certain about.

How opposed would your team be to following this route instead of a submodule?

/cc @CodeMonkeyLeet @mikbras

3rdparty/optee/libutee/CMakeLists.txt
HernanGatta: I'm afraid so, the build of the library fails with missing includes otherwise.

.gitmodules
dthaler: Very opposed.  I don't consider any of those reasons to be "very good" reasons (i.e., the supposed benefit is not worth all the problems, and the benefits of submodules outweigh them).   Instead, submodules:
1) make it easier/safer to incorporate external bugfixes in submodules, since code won't diverge, which can lead to security vulnerabilities
2) encourage contributing back code in submodules rather than encouraging maintaining separate copies
3) reduce the footprint of the specific repo, making code reviews easier
4) make licensing easier since the entire OE repo can have a single license

I would strongly recommend all the other copies of 3rdparty code to be removed and replaced with submodules.

cmake/add_enclave.cmake
HernanGatta: Good catch. The other two custom targets are followed by a call to `add_dependencies`, which CMake's documentation suggests should be used for custom targets to depend on other CMake targets.

cmake/arm-cross.cmake
HernanGatta: It turns out that this variable isn't used anywhere, so I've removed it.

libc/optee/stack.c
HernanGatta: Removed the file altogether.

libc/optee/sysconf.c
HernanGatta: File removed. This was a workaround from when I ported master's C++ support into our branch.

libc/optee/abort.c
HernanGatta: Added.

.gitmodules
HernanGatta: I back Dave on this one. Additionally, our fork of OP-TEE OS picks up changes fairly often, which we would then have to replicate here.

andschwa: Do you want to teach everyone how to use submodules? 🤣 I laugh, but that is a very high cost that is dangerous to underestimate (I've been down this road several times).

HernanGatta: I'd be interested in hearing what problems you've had with submodules. I've found that they work just fine once you know what to look out for. We can set up a short meeting and discuss the pros and cons if you'd like.

dthaler: All of the other OSS projects (which are not Microsoft-specific by the way) I've ever worked on use submodules.  So: yes.

.jenkins/Jenkinsfile
ionutbalutoiu: > I'm admittedly not familiar with Jenkins. If I were to download this file outside of the container as part of the Jenkins step that this line is embedded in, will it remain on the system for subsequent runs? If so, I could download and extract the file only if it does not yet exist, then share the extracted ones into the container. 

Tasks under `node` definitions in Jenkinsfiles (see `node("nonSGX")` in this example) are scheduled on a machine labelled `nonSGX` (there might be multiple machines with this label having free workers to execute the task).

There is no guarantee that subsequent runs will have the downloaded file. In fact, it's a good practice to start a CI stage from a clean workspace, and we try to do this. That's why we do `cleanWs()` above to clean the job workspace if it exists.

> Alternately, I could include a cached version inside the container image itself, and only download the file if a different one is specified via the corresponding pipeline argument.
> Did you have something else in mind?

I like this idea! I assume that most of the time the cached file will be used, and we will be fine. Whenever we want to update the cached file, we just update the `latest` CI Docker images.

Also, a simple retry logic on failure for the `curl` download command will work as well.

.gitmodules
johnkord: I think for the basic requirements (that we can develop our project while utilizing 3rdparty code), both submodules and checking in source (either unarchived or archived) will be satisfactory. There are some advantages and disadvantages to both styles of including 3rdparty code (as detailed above), and personally I don't have a strong opinion as they both accomplish my basic requirements. I have seen both styles used in the open source world, for example mesos utilizes a similar style to what we use in OE. (There's another style too, which is where users are told that they need to download/extract/(patch)/install 3rdparty source dependences, and then they specify those extracted 3rdparty files at configure time. In my opinion that can be a nightmare for developers, unless it's a pre-built package that can be pulled through an APT/yum repo. I see this style often in open source projects, and I really find it annoying!)

Even though I find both styles to be satisfactory, I think it would be awkward to mix & match styles, where we have some 3rdparty code that comes from a submodule and some 3rdparty code that has its source checked into our repository.  If we decide on making this change, we should probably do it for all of our 3rdparty dependencies out of consistency.

Maybe we should organize a meeting to come up with a plan for the future on what style of 3rdparty references we should maintaining? What do you all think?

jxyang: I second Andy on this. It's hard to comply to SDL with an open source project, it's even harder to achieve compliance with submodules pointing to other open source projects. SDL requires every version of every 3rd party library need to be on an "approved" list. So the #1 benefit of submodules, easier/safer to incorporate external bugfixes, is not practical, considering the cost of get approval for those bug fixes.

anakrish: **What I would prefer is the option where we can keep track of the changes we make to 3rd party code, and reapply them when we upgrade the 3rd party libraries.**
 
The key point to keep in mind is that **we don't use 3rd-party code as-is**. Since the code must  run within an enclave, we have to make changes to it.
 
Some 3rd party code like mbedtls have a nice config.h which can be edited to configure it they way we want. 

Some times the changes involve directly editing code deep in the guts of the 3rd-party module (see https://github.com/microsoft/openenclave/commit/5e8ea112910f9964666a684f110c1357f47e0c91#diff-d45)

Not all changes can be upstreamed since these changes are specific to an SGX enclave.

Which of the two approaches (or some other approach) would allow us to keep track of the changes we have done to 3rd party code and **automatically** make sure the changes are reapplied when we upgrade the 3rd party modules?

The issue I see with the current approach is that when a 3rd party module is upgraded as in https://github.com/microsoft/openenclave/pull/1786/files, there are a large number of files to look at, it is not possible to figure out which of those files contain manual edits. 
I would instead prefer the PR to just contain the set of our changes plus just a reference to the libunwind version we are upgrading to.

dthaler:
Re "Not all changes can be upstreamed since these changes are specific to an SGX enclave."

I disagree with this principle.   A good OSS library is platform agnostic, and we should choose dependencies that have platform agility as a goal.  SGX and TrustZone are just other types of platforms, and should be handled like ports to another other platform (OS, processor, whatever)   Any OSS library that refuses to be platform agile is probably not one we want to take a dependency on, since it runs counter to the goals of OE.

.jenkins/Jenkinsfile
HernanGatta: See #1887.

.gitmodules
paulcallen: I don't see this as a blocking issue if checked in after we do the 0.6 release branch off master, then create an issue to discuss this in a proper forum and come to a conclusion one way or another.

andschwa: I do see adding submodules right now as a blocking issue.

johnkord: I'm in agreement that we should go with submodules as our standard way of consuming 3rdparty source code, but I think we should make a transition plan before we quickly switch to using them, or we might introduce some problems. For example, our CI process may need to do additional cleanup for its submodules between builds or you might have sequential builds interacting with each other, unpredictably influencing the outcome of a CI run. There's several other things that need consideration that Andy brought up in our discussion earlier this week, like that people will now need to do a recursive clone or a submodule update --init, or that our build system should likely warn people if the submodules haven't been initialized yet, and so on.

I imagine the change to using submodules probably needs to be done in a separate PR. What do you all think about that? Tagging @ionutbalutoiu for communication with the Cloudbase team on possibly coordinating this upcoming task

paulcallen: I will drive this issue. I have created issue #1902

paulcallen:
@mikbras Can you review the libc bits of this pull request 
@andschwa please review the cmake bits

libc/optee/link.c
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
mikbras: This looks like duplicate code that could be shared across ELF64 architectures.

libc/optee/syscalls.c
```diff
@@ -3,20 +3,273 @@
```
mikbras: This file **optee/syscalls.c** is a cut-n-paste of **sgx/syscalls.c**. We should probably refactor **sgx/syscalls.c** into a common and non-common parts.

libc/optee/trace.c
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
mikbras: This **trace_printf** function looks like something new. What is it for? If it is for printing to the console, maybe we could consider reconciling it with **oe_host_printf()**?

tests/SampleAppCRT/enc/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 
 oeedl_file(../SampleAppCRT.edl enclave gen)
 
-add_enclave(TARGET SampleAppCRT SOURCES SampleAppCRT.cpp ${gen})
+add_enclave(TARGET SampleAppCRT UUID 31c4ecec-b94a-432a-ab76-4df401b6ee59 SOURCES SampleAppCRT.cpp ${gen})
```
mikbras: This seems like an undue burden for the test designers. Consider adding a file to the source directory called uuid.txt. The **add_enclave** cmake function can detect the existence of this file and do two things:
- Generate the appropriate C-header file (shown below).
- Name the executable accordingly.

tests/create-errors/enc/optee/user_ta_header_defines.h
```diff
@@ -0,0 +1,24 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved. */
+/* Licensed under the MIT License. */
+#pragma once
```
mikbras: The OE coding standard requires the following style.
```
#ifndef _USER_TA_HEADER_DEFINES_H
#define _USER_TA_HEADER_DEFINES_H
...
#endif /* _USER_TA_HEADER_DEFINES_H */
```

tests/create-rapid/enc/optee/user_ta_header_defines.h
```diff
@@ -0,0 +1,24 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved. */
```
mikbras: Auto-generate this file from the **uuid.txt** file in the same source directory and place it in **CMAKE_CURRENT_BINARY_DIR**. The auto-generation tool can be placed in **toosl/oeoptee** perhaps.

tests/lockless_queue/enc/optee/user_ta_header_defines.h
```diff
@@ -0,0 +1,24 @@
...
+#define TA_UUID                                            \
+    { /* eb99d409-3d52-439c-b374-87f664136434 */           \
+        0xeb99d409, 0x3d52, 0x439c,                        \
+        {                                                  \
+            0xb3, 0x74, 0x87, 0xf6, 0x64, 0x13, 0x64, 0x34 \
+        }                                                  \
+    }
+
+#define TA_FLAGS (TA_FLAG_EXEC_DDR)
+#define TA_STACK_SIZE (12 * 1024)      /* 12 KB */
```
mikbras: It seems like these should be generated from the enclave properties (oeinfo).

enclave/core/CMakeLists.txt
```diff
@@ -68,13 +66,21 @@ add_library(oecore STATIC
     __secs_to_tm.c
     __stack_chk_fail.c
     assert.c
+    atexit.c
+    backtrace.c
```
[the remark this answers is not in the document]
HernanGatta: A few points:

1. I have yet to implement backtrace support for OP-TEE TA's;
2. TA files are ELF files, so I presume we can share some logic with SGX;
3. #1786 picked up AArch64 support in libunwind;
4. OP-TEE has backtrace support built in to dump stack traces on TA panics, though that's orthogonal to the SDK.

enclave/core/optee/abort.c
```diff
@@ -1,6 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-
-void oe_abort(void)
```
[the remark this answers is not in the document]
HernanGatta: This is coming in #1892.

enclave/core/optee/calls.c
```diff
@@ -4,6 +4,14 @@
 #include <openenclave/bits/defs.h>
 #include <openenclave/edger8r/enclave.h>
 
+oe_result_t oe_ocall(uint16_t func, uint64_t arg_in, uint64_t* arg_out)
```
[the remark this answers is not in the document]
HernanGatta: I does indeed, coming in #1892.

enclave/core/optee/calls.c
```diff
@@ -21,13 +29,9 @@ oe_result_t oe_call_host_function(
...
 }
 
-void* oe_allocate_ocall_buffer(size_t size)
-{
-    OE_UNUSED(size);
-    return NULL;
-}
-
-void oe_free_ocall_buffer(void* buffer)
+void oe_abort(void)
```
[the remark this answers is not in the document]
HernanGatta: `abort()` calls `oe_abort()`.

enclave/core/optee/calls.c
```diff
@@ -21,13 +29,9 @@ oe_result_t oe_call_host_function(
...
 {
-    OE_UNUSED(buffer);
+    // TODO: Determine the appropriate call to make into OP-TEE on TA abort.
+    while (1)
```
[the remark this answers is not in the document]
HernanGatta: The proper implementation is coming in #1892, which calls into OP-TEE's panic handler.

enclave/core/CMakeLists.txt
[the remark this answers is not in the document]
mikbras: Oh great news about libunwind. Yeah, that makes sense. Thanks.

enclave/core/optee/abort.c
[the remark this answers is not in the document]
mikbras: Oh good. Thank you.

enclave/core/optee/calls.c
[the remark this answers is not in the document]
mikbras: Oh I see.

enclave/core/optee/calls.c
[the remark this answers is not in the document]
mikbras: Okay. Sounds good.

enclave/core/optee/printf.c
```diff
@@ -10,16 +10,3 @@ oe_result_t oe_log(log_level_t level, const char* fmt, ...)
     OE_UNUSED(fmt);
     return OE_UNSUPPORTED;
 }
-
-int oe_host_fprintf(int device, const char* fmt, ...)
```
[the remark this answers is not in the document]
HernanGatta: These are in `enclave/core/hostcalls.c`, where the implementation in shared with SGX.

enclave/core/optee/backtrace.c
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <openenclave/enclave.h>
+
+int oe_backtrace(void** buffer, int size)
```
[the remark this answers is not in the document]
HernanGatta: Added.

enclave/core/optee/reloc.c
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
HernanGatta: Agreed; refactored to share `init/fini` calls.

libc/optee/link.c
HernanGatta: Agreed; refactored so that `dl_iterate_phdr` is shared across OP-TEE and SGX.

libc/optee/trace.c
HernanGatta: `trace_printf` is a function used by `libutee` which we do not include in the SDK because doing so would pull in redefinitions of `printf` and friends. In the OP-TEE world, `trace_printf` is a wrapper around `printf` to pretty-print logging information. `printf` is in turn redirected to secure serial output, if available (i.e. a serial port may be reserved to OP-TEE during boot and used for logging).

enclave/core/optee/calls.c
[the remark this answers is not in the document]
HernanGatta: `oe_abort` was defined in `enclave/core/optee/abort.c`, where it was a no-op.

libc/optee/syscalls.c
HernanGatta: Indeed; `syscalls.c` is now shared.

Given the recent changes in master, there no longer is an issue with the fact that `SYS_open` does not exist in AArch64 and is instead replaced with `SYS_openat` (a difference that follows Linux).

tests/SampleAppCRT/enc/CMakeLists.txt
HernanGatta: Implemented `OE_SET_ENCLAVE_OPTEE`. See [this commit](https://github.com/openenclave/openenclave/pull/1820/commits/4f06e4da250098e8becb794fdb52e9e43e1f34aa].

tests/create-errors/enc/optee/user_ta_header_defines.h
HernanGatta: Missed it. -- File no longer exists.

tests/create-rapid/enc/optee/user_ta_header_defines.h
HernanGatta: Idem re. `OE_SET_ENCLAVE_OPTEE`.

tests/lockless_queue/enc/optee/user_ta_header_defines.h
HernanGatta: Idem re. `OE_SET_ENCLAVE_OPTEE`.

HernanGatta:
bors try

HernanGatta:
bors try

HernanGatta:
bors try

HernanGatta:
bors try

HernanGatta:
bors try

HernanGatta:
bors try

HernanGatta:
bors try

HernanGatta:
bors try

PR #1823, opened 2019-05-23
Disable qe identity fallback in OE for qe identity support in Azure DCAP client

johnkord:
This PR removes any fallback for qe identity validation when the DCAP quote provider fails to provide a valid qe identity json file when it is invoked to do so.

As of right now (2019-05-23) qe identity support isn't available in a release of the Azure DCAP client, but when support is available in a released version of the Azure DCAP client (hopefully soon), we can take these changes.

Before these changes get merged, but after a new version of the Azure DCAP client is available, we should validate this change actually works :) I've tried it manually on the update_qeid_url branch in the Azure-DCAP-Client repo, and seems to work great.

@soccerGB did I miss anything in this PR? Does anything else need to change?

### DIFF
--- a/common/sgx/qeidentity.c
+++ b/common/sgx/qeidentity.c
@@ -37,7 +37,7 @@ oe_result_t oe_enforce_qe_identity(sgx_report_body_t* qe_report_body)
         // because either get_qe_identity_info API was not supported or
         // unexpected error. Both cases are error conditions.
         OE_RAISE_MSG(
-            OE_VERIFY_FAILED,
+            OE_QEIDENTITY_NOT_RETRIEVABLE,
             "unable to retrieve qe identity from DCAP quote provider",
             NULL);
     }
--- a/common/sgx/qeidentity.c
+++ b/common/sgx/qeidentity.c
@@ -7,20 +7,6 @@
 #include "../common.h"
 #include "tcbinfo.h"
 
-// hardcoded property values used for validating quoting enclave when qe
-// identity info is not available
-// The mrsigner value of Intel's Production quoting enclave.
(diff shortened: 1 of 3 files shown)


common/sgx/qeidentity.c
```diff
@@ -50,28 +36,10 @@ oe_result_t oe_enforce_qe_identity(sgx_report_body_t* qe_report_body)
...
-        // Ensure that the QE is not a debug supporting enclave.
-        if (qe_report_body->attributes.flags & SGX_FLAGS_DEBUG)
-            OE_RAISE_MSG(OE_VERIFY_FAILED, "QE has SGX_FLAGS_DEBUG set!!");
-
-        result = OE_OK;
-        goto done;
+        // unexpected error. Both cases are error conditions.
+        OE_RAISE_MSG(
+            OE_VERIFY_FAILED,
+            "unable to retrieve qe identity from DCAP quote provider");
```
CodeMonkeyLeet: This needs to specify a NULL arg for the string to compile on Windows:
```c
OE_RAISE_MSG(OE_VERFIY_FAILED, "...", NULL);
```

See https://github.com/microsoft/openenclave/pull/1818/files for an example

johnkord:
@CodeMonkeyLeet That's a really important point, thanks Simon.

common/sgx/qeidentity.c
soccerGB: It would be great to define a new error code for this specific error case, instead of overloading OE_VERFIY_FAILED.

soccerGB:
Thanks for adding the new error code to make it easier to root cause issue from this.

johnkord:
Superseded by #2209, so I am closing this PR!

PR #1825, opened 2019-05-23
Implement deep-copy prototype

andschwa:
The deep-copy **prototype** is only exposed to users of `oeedger8r` if they specify the `--experimental` flag when generating their edge routines (note that this is done in our tests).

Notable lacking support is handling of deep-copied `out` structures. However, `in` structures with `count > 1` (essentially arrays) are supported.

Next steps: a lot more tests, refactoring, and proper error handling.

Please review both the edge8r changes and the generated code in `all_u.c`, `all_t.c`, and `all_args.h` (that is, the edge routines of the oeedger8r unit tests, specifically the functions prefixed with `deepcopy_`).

### DIFF
--- a/include/openenclave/edger8r/common.h
+++ b/include/openenclave/edger8r/common.h
@@ -133,12 +133,12 @@ done:
 /**
  * Copy an input parameter to input buffer.
  */
-#define OE_WRITE_IN_PARAM(argname, argsize, argtype)                     \
-    if (argname)                                                         \
-    {                                                                    \
-        _args.argname = (argtype)(_input_buffer + _input_buffer_offset); \
-        OE_ADD_SIZE(_input_buffer_offset, (size_t)(argsize));            \
-        memcpy((void*)_args.argname, argname, (size_t)(argsize));        \
+#define OE_WRITE_IN_PARAM(argname, argsname, argsize, argtype)            \
+    if (argname)                                                          \
+    {                                                                     \
+        _args.argsname = (argtype)(_input_buffer + _input_buffer_offset); \
+        OE_ADD_SIZE(_input_buffer_offset, (size_t)(argsize));             \
+        memcpy((void*)_args.argsname, argname, (size_t)(argsize));        \
     }
 
 #define OE_WRITE_IN_OUT_PARAM OE_WRITE_IN_PARAM
--- a/tests/oeedger8r/edl/deepcopy.edl
+++ b/tests/oeedger8r/edl/deepcopy.edl
@@ -34,6 +34,13 @@ enclave {
     [count=count, size=size] uint64_t* ptr;
   };
 
+  struct NestedStruct {
+    int plain_int;
+    [count=4] int* array_of_int;
+    ShallowStruct* shallow_struct;
+    [count=3] CountStruct* array_of_struct;
+  };
+
   trusted {
     // Since `s` is passed by value, `s.ptr` is not deep copied.
     public void deepcopy_value(ShallowStruct s, [user_check] uint64_t* ptr);
@@ -72,5 +79,8 @@ enclave {
     // Deep copy of two `CountSizeParamStruct`s each with an embedded
     // array and different counts should take place.
     public void deepcopy_countsizeparamarray([in, count=2] CountSizeParamStruct* s);
+
+    // Maybe test for recursive copying.
+    public void deepcopy_nested([in, count=1] NestedStruct* n);
   };
 };
--- a/tests/oeedger8r/enc/testdeepcopy.cpp
+++ b/tests/oeedger8r/enc/testdeepcopy.cpp
@@ -136,3 +136,21 @@ void deepcopy_countsizeparamarray(CountSizeParamStruct* s)
         OE_TEST(s[1].ptr[i] == data[4 + i]);
     OE_TEST_IGNORE(oe_is_within_enclave(s[1].ptr, s[1].count * s[1].size));
 }
+
+void deepcopy_nested(NestedStruct* n)
+{
+    OE_TEST(oe_is_within_enclave(n, sizeof(NestedStruct)));
+
+    OE_TEST(n->plain_int == 13);
+
+    OE_TEST_IGNORE(oe_is_within_enclave(n->array_of_int, 4 * sizeof(int)));
+    for (int i = 0; i < 4; ++i)
+        OE_TEST(n->array_of_int[i] == i);
+
+    OE_TEST(oe_is_outside_enclave(n->shallow_struct, sizeof(ShallowStruct)));
+
+    OE_TEST_IGNORE(
+        oe_is_within_enclave(n->array_of_struct, 3 * sizeof(CountStruct)));
+    for (size_t i = 0; i < 3; ++i)
+        deepcopy_count(&(n->array_of_struct[i]));
+}
--- a/tests/oeedger8r/enc/testdeepcopy.cpp
+++ b/tests/oeedger8r/enc/testdeepcopy.cpp
@@ -45,7 +45,7 @@ void deepcopy_count(CountStruct* s)
     OE_TEST(s->size == 64);
     for (size_t i = 0; i < 3; ++i)
         OE_TEST(s->ptr[i] == data[i]);
-    OE_TEST_IGNORE(oe_is_within_enclave(s->ptr, 3 * sizeof(uint64_t)));
+    OE_TEST(oe_is_within_enclave(s->ptr, 3 * sizeof(uint64_t)));
 }
 
 // Assert that the struct is deep-copied such that `s->ptr` has a copy
@@ -56,7 +56,7 @@ void deepcopy_countparam(CountParamStruct* s)
     OE_TEST(s->size == 64);
     for (size_t i = 0; i < s->count; ++i)
         OE_TEST(s->ptr[i] == data[i]);
-    OE_TEST_IGNORE(oe_is_within_enclave(s->ptr, s->count * sizeof(uint64_t)));
+    OE_TEST(oe_is_within_enclave(s->ptr, s->count * sizeof(uint64_t)));
 }
 
 // Assert that the struct is deep-copied such that `s->ptr` has a copy
@@ -67,7 +67,7 @@ void deepcopy_sizeparam(SizeParamStruct* s)
     OE_TEST(s->size == 64);
     for (size_t i = 0; i < s->size / sizeof(uint64_t); ++i)
         OE_TEST(s->ptr[i] == data[i]);
-    OE_TEST_IGNORE(oe_is_within_enclave(s->ptr, s->size));
+    OE_TEST(oe_is_within_enclave(s->ptr, s->size));
 }
 
 // Assert that the struct is deep-copied such that `s->ptr` has a copy
@@ -78,7 +78,7 @@ void deepcopy_countsizeparam(CountSizeParamStruct* s)
     OE_TEST(s->size == 4);
     for (size_t i = 0; i < (s->count * s->size) / sizeof(uint64_t); ++i)
         OE_TEST(s->ptr[i] == data[i]);
-    OE_TEST_IGNORE(oe_is_within_enclave(s->ptr, s->count * s->size));
+    OE_TEST(oe_is_within_enclave(s->ptr, s->count * s->size));
 }
 
 // Assert that the struct array is deep-copied such that each
@@ -90,15 +90,13 @@ void deepcopy_countparamarray(CountParamStruct* s)
     OE_TEST(s[0].size == 64);
     for (size_t i = 0; i < s[0].count; ++i)
         OE_TEST(s[0].ptr[i] == data[i]);
-    OE_TEST_IGNORE(
-        oe_is_within_enclave(s[0].ptr, s[0].count * sizeof(uint64_t)));
+    OE_TEST(oe_is_within_enclave(s[0].ptr, s[0].count * sizeof(uint64_t)));
 
     OE_TEST(s[1].count == 3);
     OE_TEST(s[1].size == 32);
     for (size_t i = 0; i < s[1].count; ++i)
         OE_TEST(s[1].ptr[i] == data[4 + i]);
-    OE_TEST_IGNORE(
-        oe_is_within_enclave(s[1].ptr, s[1].count * sizeof(uint64_t)));
+    OE_TEST(oe_is_within_enclave(s[1].ptr, s[1].count * sizeof(uint64_t)));
 }
 
 // Assert that the struct array is deep-copied such that each
@@ -110,13 +108,13 @@ void deepcopy_sizeparamarray(SizeParamStruct* s)
     OE_TEST(s[0].size == 64);
     for (size_t i = 0; i < s[0].size / sizeof(uint64_t); ++i)
         OE_TEST(s[0].ptr[i] == data[i]);
-    OE_TEST_IGNORE(oe_is_within_enclave(s[0].ptr, s[0].size));
+    OE_TEST(oe_is_within_enclave(s[0].ptr, s[0].size));
 
     OE_TEST(s[1].count == 3);
     OE_TEST(s[1].size == 32);
     for (size_t i = 0; i < s[1].size / sizeof(uint64_t); ++i)
         OE_TEST(s[1].ptr[i] == data[4 + i]);
-    OE_TEST_IGNORE(oe_is_within_enclave(s[1].ptr, s[1].size));
+    OE_TEST(oe_is_within_enclave(s[1].ptr, s[1].size));
 }
 
 // Assert that the struct array is deep-copied such that each
@@ -128,13 +126,13 @@ void deepcopy_countsizeparamarray(CountSizeParamStruct* s)
     OE_TEST(s[0].size == 4);
     for (size_t i = 0; i < (s[0].count * s[0].size) / sizeof(uint64_t); ++i)
         OE_TEST(s[0].ptr[i] == data[i]);
-    OE_TEST_IGNORE(oe_is_within_enclave(s[0].ptr, s[0].count * s[0].size));
+    OE_TEST(oe_is_within_enclave(s[0].ptr, s[0].count * s[0].size));
 
     OE_TEST(s[1].count == 3);
     OE_TEST(s[1].size == 8);
     for (size_t i = 0; i < (s[1].count * s[1].size) / sizeof(uint64_t); ++i)
         OE_TEST(s[1].ptr[i] == data[4 + i]);
-    OE_TEST_IGNORE(oe_is_within_enclave(s[1].ptr, s[1].count * s[1].size));
+    OE_TEST(oe_is_within_enclave(s[1].ptr, s[1].count * s[1].size));
 }
 
 void deepcopy_nested(NestedStruct* n)
@@ -143,14 +141,13 @@ void deepcopy_nested(NestedStruct* n)
 
     OE_TEST(n->plain_int == 13);
 
-    OE_TEST_IGNORE(oe_is_within_enclave(n->array_of_int, 4 * sizeof(int)));
+    OE_TEST(oe_is_within_enclave(n->array_of_int, 4 * sizeof(int)));
     for (int i = 0; i < 4; ++i)
         OE_TEST(n->array_of_int[i] == i);
 
     OE_TEST(oe_is_outside_enclave(n->shallow_struct, sizeof(ShallowStruct)));
 
-    OE_TEST_IGNORE(
-        oe_is_within_enclave(n->array_of_struct, 3 * sizeof(CountStruct)));
+    OE_TEST(oe_is_within_enclave(n->array_of_struct, 3 * sizeof(CountStruct)));
     for (size_t i = 0; i < 3; ++i)
         deepcopy_count(&(n->array_of_struct[i]));
 }
--- a/tests/oeedger8r/host/teststring.cpp
+++ b/tests/oeedger8r/host/teststring.cpp
@@ -71,8 +71,8 @@ oe_result_t ecall_string_no_null_terminator_modified(
     _pargs_in = (ecall_string_no_null_terminator_args_t*)_input_buffer;
     OE_ADD_SIZE(_input_buffer_offset, sizeof(*_pargs_in));
 
-    OE_WRITE_IN_PARAM(s1, _args.s1_len * sizeof(char), char*);
-    OE_WRITE_IN_OUT_PARAM(s2, _args.s2_len * sizeof(char), char*);
+    OE_WRITE_IN_PARAM(s1, s1, _args.s1_len * sizeof(char), char*);
+    OE_WRITE_IN_OUT_PARAM(s2, s2, _args.s2_len * sizeof(char), char*);
 
     /* Copy args structure (now filled) to input buffer */
     memcpy(_pargs_in, &_args, sizeof(*_pargs_in));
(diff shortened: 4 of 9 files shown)


andschwa:
bors try

include/openenclave/edger8r/common.h
```diff
@@ -133,12 +133,12 @@ OE_INLINE oe_result_t oe_add_size(size_t* total, size_t size)
 /**
  * Copy an input parameter to input buffer.
  */
-#define OE_WRITE_IN_PARAM(argname, argsize, argtype)                     \
-    if (argname)                                                         \
-    {                                                                    \
-        _args.argname = (argtype)(_input_buffer + _input_buffer_offset); \
-        OE_ADD_SIZE(_input_buffer_offset, (size_t)(argsize));            \
-        memcpy((void*)_args.argname, argname, (size_t)(argsize));        \
+#define OE_WRITE_IN_PARAM(argname, argsname, argsize, argtype)            \
```
andschwa: Oops, I can probably completely undo this change. I don't believe it's necessary any more.

tests/oeedger8r/edl/deepcopy.edl
```diff
@@ -34,6 +34,13 @@ enclave {
     [count=count, size=size] uint64_t* ptr;
   };
 
+  struct NestedStruct {
```
andschwa: This is the only test for recursive deep-copying (that is, more than the levels supported by Intel's edger8r). While I've not yet committed to supporting recursive deep-copying in v6, it's probably doable. However, it will take extra testing and error checking (as there are a lot more edge cases to consider, especially with respect to cycles).

tests/oeedger8r/enc/testdeepcopy.cpp
```diff
@@ -45,7 +45,7 @@ void deepcopy_count(CountStruct* s)
     OE_TEST(s->size == 64);
     for (size_t i = 0; i < 3; ++i)
         OE_TEST(s->ptr[i] == data[i]);
-    OE_TEST_IGNORE(oe_is_within_enclave(s->ptr, 3 * sizeof(uint64_t)));
+    OE_TEST(oe_is_within_enclave(s->ptr, 3 * sizeof(uint64_t)));
```
andschwa: Notable missing test is if `s == nullptr`, then the generated code will be incorrect, as it does not check `s && s->ptr` but only `s->ptr` (for now, this is obviously a TODO).

tests/oeedger8r/host/teststring.cpp
```diff
@@ -71,8 +71,8 @@ oe_result_t ecall_string_no_null_terminator_modified(
     _pargs_in = (ecall_string_no_null_terminator_args_t*)_input_buffer;
     OE_ADD_SIZE(_input_buffer_offset, sizeof(*_pargs_in));
 
-    OE_WRITE_IN_PARAM(s1, _args.s1_len * sizeof(char), char*);
-    OE_WRITE_IN_OUT_PARAM(s2, _args.s2_len * sizeof(char), char*);
+    OE_WRITE_IN_PARAM(s1, s1, _args.s1_len * sizeof(char), char*);
```
andschwa: I believe I can undo this (see above).

tools/oeedger8r/intel/Util.ml
```diff
@@ -69,6 +69,7 @@ let usage (progname: string) =
 --trusted             Generate trusted proxy and bridge\n\
 --untrusted-dir <dir> Specify the directory for saving untrusted code\n\
 --trusted-dir   <dir> Specify the directory for saving trusted code\n\
+--experimental        Enable experimental features\n\
```
andschwa: This is an upstream change we'll need to get back to Intel.

tools/oeedger8r/Emitter.ml
```diff
@@ -73,62 +71,49 @@ let conv_array_to_ptr (pd : pdecl) : pdecl =
...
-          | _ -> false
-        else false
-  in
-  let str_len =
-    if need_str_len_var pt then sprintf "    size_t %s_len;\n" field else ""
+(** ----- End code borrowed and tweaked from {!CodeGen.ml} ----- *)
+
+(* Borrowed from upstream OCaml implementation in
+   https://github.com/ocaml/ocaml/pull/2185 *)
+let filter_map f =
```
andschwa: TODO: I need to file a report with OSPO and a proper 3rdparty notice in order to redistribute this code. (Alternatively, we could try to pull in the [Base standard library for OCaml](https://github.com/janestreet/base) which would give me a much more complete FP library 😉)

tools/oeedger8r/Emitter.ml
```diff
@@ -73,62 +71,49 @@ let conv_array_to_ptr (pd : pdecl) : pdecl =
...
+  aux []
 
-(** ----- End code borrowed and tweaked from {!CodeGen.ml} ----- *)
+(* Helper to flatten and map at the same time. *)
+let flatten_map f l =
+  let l = List.map f l in
+  List.flatten l
 
-let is_in_ptr (ptype, _) =
+let is_in_ptr (ptype : parameter_type) =
```
andschwa: I can refactor this again to take advantage of `function` pattern matching.

tools/oeedger8r/Emitter.ml
```diff
@@ -73,62 +71,49 @@ let conv_array_to_ptr (pd : pdecl) : pdecl =
...
-let is_str_ptr (ptype, _) =
+let is_str_ptr (ptype : parameter_type) =
   match ptype with PTVal _ -> false | PTPtr (_, a) -> a.pa_isstr
 
-let is_wstr_ptr (ptype, _) =
+let is_wstr_ptr (ptype : parameter_type) =
   match ptype with PTVal _ -> false | PTPtr (_, a) -> a.pa_iswstr
 
+let gen_c_for i =
+  if i = "1" then [] else [sprintf "for (size_t _i = 0; _i < %s; ++_i)" i]
```
andschwa: I'd like to do something better than `i = "1"` here, but honestly, it's a true statement as `count = 1` is the assumed default unless `count` is explicitly set higher.

Also should probably use a less likely to collide variable name than `_i`.

tools/oeedger8r/Emitter.ml
```diff
@@ -228,6 +175,51 @@ let oe_get_param_size (ptype, decl, argstruct) =
       else pa_size
   | _ -> ""
 
+(** TODO: Refactor!  *)
```
andschwa: Seriously.

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
-
-(** Includes are emitted in [args.h]. Imported functions have already
-    been brought into function lists. *)
-let gen_t_h (ec : enclave_content) (ep : edger8r_params) =
-  let oe_gen_tfunc_prototypes (tfs : trusted_func list) =
-    if tfs <> [] then
-      List.map (fun f -> sprintf "%s;" (oe_gen_prototype f.tf_fdecl)) tfs
-    else ["/* There were no ecalls. */"]
+(** Generate the Enclave code. *)
+let gen_enclave_code (ec : enclave_content) (ep : edger8r_params) =
```
andschwa: This is now the "top-level" function, meaning we can eliminate the explicit passing of `ec` and `ep` to all the nested functions.

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
+      in
+      let struct_name = fd.fname ^ "_args_t" in
+      let retval_decl = {identifier= "_retval"; array_dims= []} in
+      let members =
+        [ ["oe_result_t _result;"]
+        ; ( if fd.rtype = Void then []
+          else gen_member_decl (PTVal fd.rtype, retval_decl) )
+        ; (if errno then ["int _ocall_errno;"] else [])
+        ; flatten_map gen_member_decl (List.map conv_array_to_ptr fd.plist) ]
+        |> List.flatten
```
andschwa: If you catch other good places for the use of `|> List.flatten`, let me know, as it makes the code a lot more readable.

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
+      in
+      [ sprintf "typedef struct _%s" struct_name
+      ; "{"
+      ; "    " ^ String.concat "\n    " members
+      ; sprintf "} %s;" struct_name
+      ; "" ]
+    in
+    let oe_gen_user_includes (includes : string list) =
```
andschwa: I think I ought to refactor this back to their calling sites, as they're now much simpler. I've also soured on the explicit comments emitted when they would otherwise return empty. Let me know what you think.

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
+      else ["/* There were no ecalls. */"; ""]
+    in
+    let oe_gen_ocall_marshal_structs (ufs : untrusted_func list) =
+      if ufs <> [] then
+        flatten_map
+          (fun uf -> oe_gen_marshal_struct uf.uf_fdecl uf.uf_propagate_errno)
+          ufs
+      else ["/* There were no ocalls. */"; ""]
+    in
+    let with_errno =
```
andschwa: This is probably repeated elsewhere now that these functions are scoped under the top-level function.

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
+      List.exists (fun uf -> uf.uf_propagate_errno) ec.ufunc_decls
+    in
+    let guard_macro =
```
andschwa: Ditto.

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
-  let oe_gen_ufunc_wrapper_prototypes (ufs : untrusted_func list) =
-    if ufs <> [] then
-      List.map
-        (fun f -> sprintf "%s;" (oe_gen_wrapper_prototype f.uf_fdecl false))
-        ufs
-    else ["/* There were no ocalls. */"]
+  (* We need to check Ptrs for Foreign or Struct types, then check
+     those against the user's structs, and then check if any members
+     should be deep copied. *)
+  let get_deepcopy a =
```
andschwa: The magic 🤯

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
-    ; ""
-    ; "#include <openenclave/enclave.h>"
-    ; ""
-    ; sprintf "#include \"%s_args.h\"" ec.file_shortnm
-    ; ""
-    ; "OE_EXTERNC_BEGIN"
+  (* Prepare [input_buffer]. *)
+  let oe_prepare_input_buffer (fd : func_decl) (alloc_func : string) =
+    let oe_compute_buffer_size (buffer : string)
+        (predicate : parameter_type * declarator -> bool) (plist : pdecl list)
```
andschwa: This function signature is a good example of why I've soured on the explicit typing of parameters. I think we should instead just use deduced typing (coupled with way more comments than I have currently).

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
+      ; "{"
+      ; "    _result = OE_FAILURE;"
+      ; "    goto done;"
+      ; "}"
+      ; ""
+      ; "/* Unmarshal return value and out, in-out parameters. */"
+      ; ( if fd.rtype <> Void then "*_retval = _pargs_out->_retval;"
+        else "/* No return value. */" ) ]
+    ; (* This does not use String.concat because the elements are multiple lines. *)
+      (* TODO: Fix for deep copy. *)
```
andschwa: This is why `out` params are not yet supported 😉

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
+    ; (* Buffer validation *)
+      "    /* Make sure input and output buffers lie within the enclave. */"
+    ; "    if (!input_buffer || !oe_is_within_enclave(input_buffer, \
+       input_buffer_size))"
+    ; "        goto done;"
     ; ""
-    ; "#include <stdlib.h>"
-    ; "#include <string.h>"
-    ; "#include <wchar.h>"
+    ; "    if (!output_buffer || !oe_is_within_enclave(output_buffer, \
```
andschwa: Uh GitHub, bad diff highlighting!

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
+      |> List.flatten
+    in
+    flatten_map (gen_assignment "" "1") fd.plist
   in
-  let guard = sprintf "EDGER8R_%s_U_H" (String.uppercase ec.file_shortnm) in
-  let content =
-    [ sprintf "#ifndef %s" guard
-    ; sprintf "#define %s" guard
+  (* Generate host ECALL wrapper function. *)
+  let oe_gen_host_ecall_wrapper (name : string) (tf : trusted_func) =
```
andschwa: The `name` parameter is ripe for refactoring after the nesting of these functions.

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
+    ; "done:"
+    ; "    if (_buffer)"
+    ; "        oe_free_ocall_buffer(_buffer);"
+    ; "    return _result;"
+    ; "}"
+    ; "" ]
   in
-  let content =
-    [ sprintf "#include \"%s_u.h\"" ec.file_shortnm
+  (* Generate ocall function. *)
```
andschwa: Bad comment 😅

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
   in
-  let ecalls_fname = ec.file_shortnm ^ "_u.c" in
-  let os = open_file ecalls_fname ep.untrusted_dir in
-  fprintf os "%s" (String.concat "\n" content) ;
-  close_out os
-
-(** Generate the Enclave code. *)
-let gen_enclave_code (ec : enclave_content) (ep : edger8r_params) =
+  (* Validate Open Enclave supported EDL features. *)
+  let validate_oe_support (ec : enclave_content) (ep : edger8r_params) =
```
andschwa: Oh, should probably move this function further down maybe. I haven't really looked at the ordering of the functions here yet.

HernanGatta:
I haven't looked at the changes to `oeedger8r` yet (I've got to brush up on my OCaml, it's been a while), but do your changes ever assume that the enclave can read host memory? The reason for the question is that this is not necessarily the case for OP-TEE.

tools/oeedger8r/Emitter.ml
```diff
@@ -228,6 +175,51 @@ let oe_get_param_size (ptype, decl, argstruct) =
...
+  let decl_size_to_string (ptype : parameter_type) (d : declarator) =
+    let dims = List.map string_of_int d.array_dims in
+    String.concat " * " dims
+  in
+  match ptype with
+  | PTPtr (_, ptr_attr) ->
+      let pa_size = pa_size_to_string ptr_attr.pa_size in
+      (* Compute declared size *)
+      let decl_size = decl_size_to_string ptype decl in
+      if ptr_attr.pa_isstr || ptr_attr.pa_iswstr then "1"
```
anakrish: Instead of "1" we could use a string option or a new variant type that is more descriptive.

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
+        ; ( if fd.rtype = Void then []
+          else gen_member_decl (PTVal fd.rtype, retval_decl) )
+        ; (if errno then ["int _ocall_errno;"] else [])
+        ; flatten_map gen_member_decl (List.map conv_array_to_ptr fd.plist) ]
+        |> List.flatten
+      in
+      [ sprintf "typedef struct _%s" struct_name
+      ; "{"
+      ; "    " ^ String.concat "\n    " members
+      ; sprintf "} %s;" struct_name
```
anakrish: I wonder if using the ^ operator would be more readable here:

`; "} " ^ struct_name ^ ";"`

tools/oeedger8r/Emitter.ml
```diff
@@ -967,274 +472,861 @@ let warn_size_and_count_params (fd : func_decl) =
...
+      ; "" ]
+    in
+    let oe_gen_user_includes (includes : string list) =
+      if includes <> [] then List.map (sprintf "#include \"%s\"") includes
+      else ["/* There were no user includes. */"]
+    in
+    let oe_gen_user_types (cts : composite_type list) =
```
anakrish: I like the nesting. However, it completely messes up the diff and it is hard to follow what is just a simple nesting vs what is a genuine code change.

tools/oeedger8r/Emitter.ml
anakrish: There are couple of options here. You could use a type variable in the declaration for verbose types and have the compiler figure it out.

`let oe_compute_buffer_size (buffer:string) (predicate : 'a_predicate) (plist : pdecl list)`

Alternatively you could a type alias.

tools/oeedger8r/Emitter.ml
andschwa: I personally believe the diff was already going to be difficult to go through due to the amount of churn, making this an ideal time to move these functions, but if you really feel necessary, I can undo the nesting, plumb the passing of enclave_params all the way down in order to support `--experimental`, update the PR with maybe a slightly better diff (again, I don't think it would be a huge improvement), and then redo the nesting and "unplumb" all the parameters.

tools/oeedger8r/Emitter.ml
andschwa: There's a lot more out of Base that would be nice to use though 😄

tools/oeedger8r/Emitter.ml
```diff
@@ -73,62 +71,49 @@ let conv_array_to_ptr (pd : pdecl) : pdecl =
...
   match ptype with PTVal _ -> false | PTPtr (_, a) -> a.pa_isstr
 
-let is_wstr_ptr (ptype, _) =
+let is_wstr_ptr (ptype : parameter_type) =
   match ptype with PTVal _ -> false | PTPtr (_, a) -> a.pa_iswstr
 
+let gen_c_for i =
+  if i = "1" then [] else [sprintf "for (size_t _i = 0; _i < %s; ++_i)" i]
+
+let gen_c_deref i = if i = "1" then "->" else "[_i]."
```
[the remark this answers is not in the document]
andschwa: Oh heh I confused myself re-reading this to the point that I need to do something about it. Will probably use the Some/None pervasives.

tools/oeedger8r/Emitter.ml
```diff
@@ -228,6 +175,51 @@ let oe_get_param_size (ptype, decl, argstruct) =
...
   | _ -> ""
 
+(** TODO: Refactor!  *)
+let oe_get_param_count (ptype, decl, argstruct) =
+  (* Get the base type of the parameter, that is, recursively
+     decompose the pointer. *)
+  let atype =
+    match get_param_atype ptype with
+    | Ptr at -> at
+    | _ -> get_param_atype ptype
```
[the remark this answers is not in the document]
andschwa: This is original code just copied for a second use. It's unfortunately one of the worst parts of our code where we have some refactoring / reconsidering to do, cause I can't answer your question easily here.

tools/oeedger8r/Emitter.ml
```diff
@@ -228,6 +175,51 @@ let oe_get_param_size (ptype, decl, argstruct) =
...
+  let attr_value_to_string av =
+    match av with
+    | None -> "1"
+    | Some (ANumber n) -> string_of_int n
+    | Some (AString s) -> sprintf "%s%s" argstruct s
+    (* another parameter name *)
+  in
+  let pa_size_to_string pa =
+    let c = attr_value_to_string pa.ps_count in
+    if c <> "" then sprintf "%s" c
```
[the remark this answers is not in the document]
andschwa: Note the TODO above 😆 this was copied from `oe_get_param_size` and minimally adapted, it's an area that needs a lot of work.

tools/oeedger8r/Emitter.ml
andschwa: I think i'm going to drop most argument types and instead rely on type deductions. After writing a lot of this code, I found that they add very little value (maybe none entirely).

andschwa:
> I haven't looked at the changes to `oeedger8r` yet (I've got to brush up on my OCaml, it's been a while), but do your changes ever assume that the enclave can read host memory? The reason for the question is that this is not necessarily the case for OP-TEE.

No, I do not believe so. We're simply expanding our existing marshaling structs to marshal all the data between enclave and host.

tools/oeedger8r/Emitter.ml
anakrish: You could just submit a separate PR with just the code-nesting stuff. That can be easily reviewed and merged to master. Then this PR will be much easier to review.

andschwa:
> I took an initial pass at this. Looks good. But quite hard to review since nesting/moving of functions around is being done with feature implementation. So it is very difficult to figure out which code is new and needs scrutiny vs which code is just code-nesting/movement that I can safely skim over.

I can try to undo the nesting and plumb through the parameters needed to implement --experimental, but I think it will only help a bit.

> I have a general question about the use of master branch; not for this PR per se, but for many PRs I see going in now.
> I am fine with using master branch as a dev branch where incomplete/in-progress code can be checked in. However, currently we just seem to be picking stuff from the master branch onto the release branch without any pull-request/code-review.
> So if I were to relax review bar and accept a particular PR with the understanding that it is in dev, what part of the development process ensures that only complete and shipping quality features are pulled into the release branch?

Master branch is the development branch.

We should not be picking things randomly into release branches, we should be following the process to branch off master when it is at a reasonable state (which means when items in the v6 milestone are complete or close to it), and do PRs to the release branch whenever cherry picking changes to it (I don't believe this was followed for the last release, which means we need to improve our processes here).

I believe the part of the development process (that I'm admittedly missing here for now) is the tracking of the parts of a feature by issues/checklists in the release milestone. For this PR, that's on me to create before this can be merged as partial progress.

anakrish:
Thanks for the explanation. For most part, it sounds good.

>PRs to the release branch whenever cherry picking changes to it (I don't believe this was followed for the last release, which means we need to improve our processes here).

Great. This makes me less nervous about the process. I think it is something we need to bring up to the team; since this was not mentioned so far in any of the release process discussions.

> , we should be following the process to branch off master when it is at a reasonable state (which means when items in the v6 milestone are complete or close to it)

I still have concerns about this. Master being a shared dev branch, ensures via CI that no regressions have taken place. However, what is the quality bar for each feature in master before we branch? How is the branch date determined? Is it a pre-determined date (if so, there is no guarantee that the in progress features meet the quality bar) or is it gonna be determined by evaluating the quality of each feature?

@CodeMonkeyLeet I would appreciate your input into this as well.

andschwa:
@anakrish I plumbed `ep` through to avoid the reindentation of the refactoring. Sorry, I thought we'd loosely agreed to do that in this PR, but I will redo it in a second PR.

TODOs for _this_ PR:

- [ ] Notice for 3rdparty code
- [ ] Check for `nullptr` for nested code
- [ ] Undo change to `OE_WRITE_IN_PARAM`

andschwa:
> I still have concerns about this. Master being a shared dev branch, ensures via CI that no regressions have taken place. However, what is the quality bar for each feature in master before we branch? How is the branch date determined? Is it a pre-determined date (if so, there is no guarantee that the in progress features meet the quality bar) or is it gonna be determined by evaluating the quality of each feature?

My understanding is that although a lot of this responsibility is on the chosen release manager (such as choosing a date), the quality evaluation and testing falls on all of us as a team to do while preparing a release.

andschwa:
Opened https://github.com/microsoft/openenclave/issues/1839 to track changes for next iteration(s).

andschwa:
I am waiting on permissions to propagate so I can properly record the use of code from upstream OCaml, plus an approval from a reviewer. I believe I captured all the requested changes (and more!) in #1839.

andschwa:
bors try

andschwa:
bors try

andschwa:
bors try

PR #1828, opened 2019-05-24
Set up CI with Azure Pipelines

johnkord:


johnkord:
I'm learning about azure devops and this was automatically created! Yikes!

PR #1838, opened 2019-05-29
lockless_queue implemented to support switchless calling

EMumau:
This is a requirement for the implementation of switchless calling.  Breaking the switchless calling pull request into smaller pieces to make it easier to review.

### DIFF
--- a/common/lockless_queue.c
+++ b/common/lockless_queue.c
@@ -1,0 +1,143 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. */
+
+#include <openenclave/internal/lockless_queue.h>
+#include <stdlib.h>
+
+#if _MSC_VER
+#include <intrin.h>
+#endif
+
+/* functions for oe_lockless_queue_node */
+/*---------------------------------------------------------------------------*/
+void oe_lockless_queue_node_init(oe_lockless_queue_node* p_node)
+{
+    p_node->p_link = NULL;
+} /* init_oe_lockless_queue_node */
+
+/* functions for oe_lockless_queue */
+/*---------------------------------------------------------------------------*/
+void oe_lockless_queue_init(oe_lockless_queue* p_queue)
+{
+#ifdef _MSC_VER
+    _InterlockedExchangePointer(&(p_queue->p_tail), NULL);
+    _InterlockedExchangePointer(&(p_queue->p_head), NULL);
+#elif defined __GNUC__
+    __atomic_store_n(&(p_queue->p_tail), NULL, __ATOMIC_RELAXED);
+    __atomic_store_n(&(p_queue->p_head), NULL, __ATOMIC_RELAXED);
+#endif /* __GNUC__ */
+} /* init_oe_lockless_queue */
+
+void oe_lockless_queue_push_back(
+    oe_lockless_queue* p_queue,
+    oe_lockless_queue_node* p_node)
+{
+    oe_lockless_queue_node* p_expected = NULL;
+#ifdef _MSC_VER
+    oe_lockless_queue_node* p_actual = NULL;
+#endif /* _MSC_VER */
+    do
+    {
+#ifdef _MSC_VER
+        p_expected = p_actual;
+        p_node->p_link = p_expected;
+        p_actual = (oe_lockless_queue_node*)_InterlockedCompareExchangePointer(
+            &(p_queue->p_tail), p_node, p_expected);
+    } while (p_actual != p_expected);
+#elif defined __GNUC__
+        p_node->p_link = p_expected;
+    } while (!__atomic_compare_exchange_n(
+        &(p_queue->p_tail),
+        &p_expected,
+        p_node,
+        1,
+        __ATOMIC_ACQ_REL,
+        __ATOMIC_ACQUIRE));
+#endif /* __GNUC__ */
+} /* oe_lockless_queue_push */
+
+oe_lockless_queue_node* oe_lockless_queue_pop_front(oe_lockless_queue* p_queue)
+{
+    /* try to take a node from the head */
+    oe_lockless_queue_node* popped_node = NULL;
+#ifdef _MSC_VER
+    popped_node = (oe_lockless_queue_node*)_InterlockedCompareExchangePointer(
+        &(p_queue->p_head), NULL, NULL);
+#elif defined __GNUC__
+    popped_node = __atomic_load_n(&(p_queue->p_head), __ATOMIC_ACQUIRE);
+#endif /* __GNUC__ */
+
+    if (NULL != popped_node)
+    {
+        /* there was a node at the head
+         * pop the node from the head and replace it with the node that it
+         *     points to
+         * remove the reference from the popped node to the next node */
+        oe_lockless_queue_node* next_node = popped_node->p_link;
+        popped_node->p_link = NULL;
+#ifdef _MSC_VER
+        _InterlockedExchangePointer(&(p_queue->p_head), next_node);
+#elif defined __GNUC__
+        __atomic_store_n(&(p_queue->p_head), next_node, __ATOMIC_RELEASE);
+#endif /* __GNUC__ */
+    }
+    else
+    {
+        /* there wasn't a node at the head
+         * so refill the head with the nodes from the tail */
+#ifdef _MSC_VER
+        popped_node =
+            (oe_lockless_queue_node*)_InterlockedCompareExchangePointer(
+                &(p_queue->p_tail), NULL, NULL);
+#elif defined __GNUC__
+        popped_node = __atomic_load_n(&(p_queue->p_tail), __ATOMIC_RELAXED);
+#endif /* __GNUC__ */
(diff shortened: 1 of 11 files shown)


common/lockless_queue.c
```diff
@@ -0,0 +1,143 @@
...
+        p_expected = p_actual;
+        p_node->p_link = p_expected;
+        p_actual = (oe_lockless_queue_node*)_InterlockedCompareExchangePointer(
+            &(p_queue->p_tail), p_node, p_expected);
+    } while (p_actual != p_expected);
+#elif defined __GNUC__
+        p_node->p_link = p_expected;
+    } while (!__atomic_compare_exchange_n(
+        &(p_queue->p_tail),
+        &p_expected,
```
anakrish: I don't follow this logic. 
I would expect that a successful push_back operation would have the following post-conditions:

1. p_node->p_link == NULL
2. p_queue->p_tail == p_node

Since __atomic_compare_exchange_n modifies p_expected in case of failure, it seems like we are constructing a (reverse) linked list starting at tail.

Is that the case?

common/lockless_queue.c
```diff
@@ -0,0 +1,143 @@
...
+        /* there was a node at the head
+         * pop the node from the head and replace it with the node that it
+         *     points to
+         * remove the reference from the popped node to the next node */
+        oe_lockless_queue_node* next_node = popped_node->p_link;
+        popped_node->p_link = NULL;
+#ifdef _MSC_VER
+        _InterlockedExchangePointer(&(p_queue->p_head), next_node);
+#elif defined __GNUC__
+        __atomic_store_n(&(p_queue->p_head), next_node, __ATOMIC_RELEASE);
```
anakrish: It is not clear to me what the purpose of 'head' is. Is this a deque?

common/lockless_queue.c
```diff
@@ -0,0 +1,143 @@
...
+    else
+    {
+        /* there wasn't a node at the head
+         * so refill the head with the nodes from the tail */
+#ifdef _MSC_VER
+        popped_node =
+            (oe_lockless_queue_node*)_InterlockedCompareExchangePointer(
+                &(p_queue->p_tail), NULL, NULL);
+#elif defined __GNUC__
+        popped_node = __atomic_load_n(&(p_queue->p_tail), __ATOMIC_RELAXED);
```
anakrish: Going by the logic of push_back, it seems like this would pop of the last added node which tail points to.
That would indicate a LIFO order; shouldn't this be a FIFO order?

common/lockless_queue.c
```diff
@@ -0,0 +1,143 @@
...
+                1,
+                __ATOMIC_ACQ_REL,
+                __ATOMIC_ACQUIRE))
+            {
+                continue;
+            }
+#endif /* __GNUC__ */
+
+            /* reverse the nodes from the tail */
+            oe_lockless_queue_node* prev_node = NULL;
```
anakrish: I don't understand why we need to pop all the nodes and re-insert them.
What happens to the order when a new node is added while we are in the process of re-adding all these nodes here?

common/lockless_queue.c
```diff
@@ -0,0 +1,143 @@
...
+
+/* functions for oe_lockless_queue */
+/*---------------------------------------------------------------------------*/
+void oe_lockless_queue_init(oe_lockless_queue* p_queue)
+{
+#ifdef _MSC_VER
+    _InterlockedExchangePointer(&(p_queue->p_tail), NULL);
+    _InterlockedExchangePointer(&(p_queue->p_head), NULL);
+#elif defined __GNUC__
+    __atomic_store_n(&(p_queue->p_tail), NULL, __ATOMIC_RELAXED);
```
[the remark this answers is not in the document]
EMumau: Very briefly, RELAXED operations enforce ordering but do not synchronize with other threads (no memory fence) where ACQUIRE and RELEASE enforce ordering and synchronize with other threads (using memory fence).  The specific uses of ACQUIRE and RELEASE in the code enforce the required memory ordering between threads while the RELAXED operations reduce the number of memory reads.
C++ Concurrency in Action - Anthony Williams - Section 5.3

common/lockless_queue.c
EMumau: Short summary:  The implementation of a lockless data structure is limited by the constraints of atomic operations.  The data structure needs to always be in a consistent stable state across threads.  The only guarantee is that individual atomic operations are consistent between threads.  There is no way to assure that multiple atomic operations happen without thread context changes.
Due to these constraints, this FIFO queue is implemented as two stacks.  This queue implementation is thread safe for multiple producers and a single consumer.
p_tail is the "push" stack.  The implementation of oe_lockless_queue_push_back enables multiple threads to call push_back concurrently.  The compare-exchange operator ensures that conflicts between threads are resolved in a manner that ensures the state of the data structure is stable.
p_head is the "pop" stack.  The implementation of oe_lockless_queue_pop_back is only safe for one thread to call concurrently, but it is safe to call concurrently with push_back. This method first checks to see if there are any nodes on the pop stack and if so it pops a node off and returns.  In the case that there aren't any nodes on the pop stack, it pops all of the nodes off of the push stack, reverses the order, and returns the new node at the top of the stack (front of the queue).

common/lockless_queue.c
EMumau: This is FIFO queue.  It is not a deque.  https://github.com/microsoft/openenclave/pull/1838#discussion_r288771644

common/lockless_queue.c
EMumau: This is a FIFO queue.  The tests cases provided in tests/lockless_queue ensure that the nodes are popped off in an order consistent with the thread that inserted them.

common/lockless_queue.c
EMumau: This is because of the two stack implementation.  https://github.com/microsoft/openenclave/pull/1838#discussion_r288771644  Two stacks are used because it is not possible to update more than one pointer with a single atomic operation.  It is also not possible to implement a double-linked list atomically.  However, this implementation is safe for pushing a new node while concurrently popping a node even in the case where all the nodes are being moved from the tail to the head.

Note:  I have implemented lockless FIFO queues before as a single-linked list.  While that makes it much easier to follow the logic, it is much less efficient because each pop needs to iterate through each node in the list (using full atomic synchronization for each node->p_next) to get to the head.  That implementation has O (n x n) complexity while the two stack implementation has O (n) complexity.

common/lockless_queue.c
anakrish: @EMumau Thanks for the explanation. Why not add these design/implementation decisions as comments in the code in the first place. Otherwise, the reader is left with deciphering the implementation details/guarantees.

anakrish: Based on your description, this queue handles _multiple-producers, single-consumer_ scenario (the pop can only be safely called only by one thread).
This is bit surprising for use in implementing switchless calls. How are the queues going to be used for implementing ecalls? Is it going to be 1 queue for each enclave worker thread?

common/lockless_queue.c
```diff
@@ -0,0 +1,143 @@
...
+         * so refill the head with the nodes from the tail */
+#ifdef _MSC_VER
+        popped_node =
+            (oe_lockless_queue_node*)_InterlockedCompareExchangePointer(
+                &(p_queue->p_tail), NULL, NULL);
+#elif defined __GNUC__
+        popped_node = __atomic_load_n(&(p_queue->p_tail), __ATOMIC_RELAXED);
+#endif /* __GNUC__ */
+
+        if (NULL != popped_node)
```
anakrish: If the head-stack is empty, we are always going to fill the head-stack with reversed tail-stack.
In that case, it seems like this entire logic can be simplified by atomically swapping p_tail and NULL:

```c
popped_node = NULL;
oe_lockless_queue_node* null_node = NULL;
__atomic_exchange(&(p_queue->p_tail), &null_node, &popped_node, __ATOMIC_ACQUIRE);
```
// Now that the list has been detached from tail, reverse it if not empty and install as head-stack.


Am I missing something?

EMumau: I noticed that too and intended to change that logic.  I must have gotten distracted.  The old gnu intrinsic libs didn't have a simple exchange (only compare exchange).  I have made this change which is now running in sandbox.

common/lockless_queue.c
EMumau: I have added comments to lockless_queue.h to clarify the usage.

To use this implementation of a lockless queue with *multiple-consumers* there would need to be some additional locking mechanism (i.e. spinlock) around the call to oe_lockless_queue_pop_front.  Every implementation of thread-safe queues that I have seen has been *single-consumer*.

anakrish:
@paulcallen Can you review this since you had objection to using our own thread-safe/lockless queue implementation?

anakrish:
After discussion with @lilyfang, switchless calls will not make the v0.6 release.
Due to queues being key data structures, we have created a task for continuing the investigation #1896 of open source queues/ringbuffers.
We will initially use this implementation; but will complete the investigation before shipping in v0.7 release.

PR #1841, opened 2019-05-30
Made mbedtls depend on oelibc.

gupta-ak:
The PR does the following:

- Make mbedtls depend on musl headers and musl libc instead of corelibc headers and oecore.
- Remove corelibc headers includes from oeenclave/oecrypto/mbedtls code.
- Rewrite the mbedtls CMake build so the dependencies between the mbedtls libraries, oelibc and oecore
are correctly defined when installed.
- Change the attestation test to use libc headers instead of corelibc.


enclave/crypto/cmac.c
```diff
@@ -1,13 +1,8 @@
...
 // Licensed under the MIT License.
 
-/* Nest mbedtls header includes with required corelibc defines */
-// clang-format off
-#include "mbedtls_corelibc_defs.h"
-#include <mbedtls/pk.h>
 #include <mbedtls/cmac.h>
-#include "mbedtls_corelibc_undef.h"
-// clang-format on
+#include <mbedtls/pk.h>
```
soccerGB: This is much cleaner!

soccerGB:
bors try

CMakeLists.txt
```diff
@@ -212,6 +212,7 @@ if (BUILD_ENCLAVES)
   add_subdirectory(enclave/core)
   add_subdirectory(3rdparty)
   add_subdirectory(libc)
+  add_subdirectory(3rdparty/mbedtls)
```
CodeMonkeyLeet: Why are we doing this? For build ordering? It deserves a comment about the dependency at least, but in the CMake idiom, shouldn't this just be an explicit `add_depenencies` on oelibc?

Also, for my understanding, doesn't the link dependency already force the CMake graph to handle this?

gupta-ak: oh interesting. I though that cmake wants the targets defined before you use them, but it looks like it works fine.

Yeah, this isn't needed. you can also remove the `add_subdirectory(enclave/core)` part and just have `enclave/CMakeLists.txt` do `add_subdirectory`.

3rdparty/mbedtls/CMakeLists.txt
```diff
@@ -1,146 +1,187 @@
...
+  set(CMAKE_AR ${LLVM_AR})
+endif()
 
-# NOTE: Because these libraries are `IMPORTED`, it is not possible
-# with CMake to install them as targets, so instead they must be
-# installed as files. This is what causes the above problem with the
-# repeated dependency on `oecore`.
-install(FILES ${BINARY_DIR}/library/libmbedtls.a
-  DESTINATION ${CMAKE_INSTALL_LIBDIR}/openenclave/enclave)
+if ("${CMAKE_BUILD_TYPE}" STREQUAL "DEBUG")
```
anakrish: What ought to be the flags for RelWithDebInfo?
Won't the optimization-level flags be inherited via oelibc dependency?

3rdparty/mbedtls/CMakeLists.txt
```diff
@@ -1,146 +1,187 @@
...
+string(REGEX MATCH "GNU" CMAKE_COMPILER_IS_GNU "${CMAKE_C_COMPILER_ID}")
+
+set(MBEDTLS_COMPILE_OPTS
+  -W
+  -Wdeclaration-after-statement
+  -Wwrite-strings
+  -Wshadow
+  # Disable conversion warnings inherited from OE.
+  -Wno-sign-conversion
+  -Wno-conversion)
```
anakrish: Can you confirm if mbedtls is being compiled with 'enclave-appropriate' flags?

- -fPIE  - needed for getting the compiler to generate code optimal for use in an executable/first loaded module. Without this there could be relocations that our loader may not be able to handle.
- -fvisibility=hidden - to keep the enclave self-contained and also helps keep the loader simple
- -fno-stack-protector - We don't support stack protector yet. Needs more work in our C library startup.
- -ftls-model=local-exec - This is the most optimal thread-local storage model - best for use in a single module that is loaded. All TLS variables are accessed via negative offsets off FS register.

Without these flags, there will be loading errors for some enclaves; but common enough, but hard to predictably reproduce.

They ought to get inherited from oelibc dependency; but can you double-check since cmake has surprised us in the past.

tests/attestation_cert_apis/enc/enc.cpp
```diff
@@ -202,10 +175,10 @@ oe_result_t generate_key_pair(
 
         *public_key = local_public_key;
         // plus one to make sure \0 at the end if counted
```
dthaler: typo: change "if" to "is"

gupta-ak:
bors try

tests/attestation_cert_apis/enc/enc.cpp
gupta-ak: thanks fixed!

3rdparty/mbedtls/CMakeLists.txt
gupta-ak: your right. no need to have this part.

3rdparty/mbedtls/CMakeLists.txt
gupta-ak: yup they are all there

PR #1846, opened 2019-05-31
Add SDK version to OE Environment preparation

george-caluta:
This change is required in order for the Azure DCAP build tests added in microsoft/Azure-DCAP-Client#65


anitagov:
Yes, we need this SDK to build the DCAP Client for Windows. Thanks.

PR #1847, opened 2019-05-31
Change ocall log level to VERBOSE from INFO.

johnkord:
This is a proposal to change the log level of the new ocall logging feature from INFO to VERBOSE. I'm proposing this change because there is important diagnostic information that can be gathered when setting the log level to INFO (for example, when diagnosing quote verification failures), and OCALL log messages add a lot of noise when they might not be important for diagnosing an issue related to core OE functionality.

Tagging @achamayou who first implemented the ocall log feature. Feel free to tag anyone else for review

### DIFF
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -420,7 +420,7 @@ static oe_result_t _handle_ocall(
         *arg_out = 0;
 
     oe_log(
-        OE_LOG_LEVEL_INFO,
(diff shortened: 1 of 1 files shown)


achamayou:
Sounds good to me, I agree that INFO was probably a little too optimistic. Initially I had an exception not to log logging OCalls, which reduced output somewhat, but we changed that so it’s definitely best at VERBOSE now.

Issue #1839, opened 2019-05-29
Deep-copy features


andschwa:
This is a "story" covering TODOs to finish for the deep-copy work:

- [x] Clean up `get_deepcopy` function(s)
- [x] Undo change to `OE_WRITE_IN_PARAM`
- [x] Add tests for and support `nullptr` structs and their fields
- [x] Fix indentation of `for` loops
- [x] Clean up `gen_c_for` parameter names and C variables
- [x] Fix up `gen_c_deref` parameter names
- [x] Use a `string option` instead of `count = "1"` for `attr_value_to_string` etc.
- [ ] Ship `--experimental` flag upstream to Intel
- [x] Add tests for and support `out` params
- [ ] Add Scenario 6 test
- [x] Refactor so `edger8r_params` are "global"
- [x] Refactor deeply plumbed arguments to use "global" `ep`
- [x] Refactor I/O side effects to one location
- [x] Refactor `oe_get_param_size` and `oe_get_param_count`
- [x] Use string concatenation over `sprintf` in places to enhance readability
- [ ] Initiate security review
- [ ] Document use / restrictions
- [x] Refactor `is_x_ptr` functions to use `function` pattern matching
- [x] Add `is_x_and_y_ptr` functions for common use cases

andschwa:
@yakman2020 By the way, much of this got into v6 under the experimental flag. I’m just loathe to remove it without more thorough testing.

mingweishih:
As deep-copy feature is supported by default now, closing this issue.

PR #1853, opened 2019-06-03
Minimum changes needed to refactor against `edger8r_params`

andschwa:
This is a basic refactor with no logic changes, that moves all code
depending on `edger8r_params` under `gen_enclave_code` so we can
refactor out all the deeply nested argument passing. This commit only
moves the code to enable these upcoming changes.


andschwa:
Part of #1839.

PR #1858, opened 2019-06-04
Log ECALLs

letmaik:
Same as #1816 but for ECALLs.

### DIFF
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -396,15 +396,15 @@ static const char* oe_ocall_str(oe_func_t ocall)
 
 static const char* oe_ecall_str(oe_func_t ecall)
 {
-    static const char* func_names[] = {"OE_ECALL_DESTRUCTOR",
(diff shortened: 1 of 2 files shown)


host/sgx/calls.c
```diff
@@ -742,6 +763,13 @@ oe_result_t oe_ecall(
     if (!(tcs = _assign_tcs(enclave)))
         OE_RAISE(OE_OUT_OF_THREADS);
 
+    oe_log(
+        OE_LOG_LEVEL_VERBOSE,
+        "%s 0x%x OE_ECALL: %s\n",
```
johnkord: We probably want to be consistent with the other PR which changes this line for OCALLs

letmaik: Done.

letmaik:
@johnkord I aligned it with the other PR.

johnkord:
Thanks Maik!

bors r+

PR #1859, opened 2019-06-04
OCall logging bugfixes

achamayou:
@letmaik spotted a couple of mistakes that found their way into the OCall logging PR:

- A missing comma (!) in the array of names, coupled with a mistaken assertion
- OE_OCALL_CALL_HOST_FUNCTION was labeled as OE_OCALL, when it should have been EDL_OCALL

### DIFF
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -367,22 +367,22 @@ done:
 
 static const char* oe_ocall_str(oe_func_t ocall)
 {
-    static const char* func_names[] = {"OE_OCALL_CALL_HOST_FUNCTION",
-                                       "OE_OCALL_GET_QE_TARGET_INFO",
-                                       "OE_OCALL_GET_QUOTE",
-                                       "OE_OCALL_GET_REVOCATION_INFO",
-                                       "OE_OCALL_GET_QE_ID_INFO",
-                                       "OE_OCALL_THREAD_WAKE",
-                                       "OE_OCALL_THREAD_WAIT",
-                                       "OE_OCALL_THREAD_WAKE_WAIT",
-                                       "OE_OCALL_MALLOC",
-                                       "OE_OCALL_REALLOC",
-                                       "OE_OCALL_FREE",
-                                       "OE_OCALL_WRITE",
-                                       "OE_OCALL_SLEEP",
-                                       "OE_OCALL_GET_TIME",
-                                       "OE_OCALL_BACKTRACE_SYMBOLS",
-                                       "OE_OCALL_LOG"};
+    static const char* func_names[] = {"CALL_HOST_FUNCTION",
+                                       "GET_QE_TARGET_INFO",
+                                       "GET_QUOTE",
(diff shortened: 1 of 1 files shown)


host/sgx/calls.c
```diff
@@ -421,9 +420,10 @@ static oe_result_t _handle_ocall(
 
     oe_log(
         OE_LOG_LEVEL_VERBOSE,
-        "%s 0x%x OE_OCALL: %s\n",
+        "%s 0x%x %s: %s\n",
         enclave->path,
         enclave->addr,
+        func == OE_OCALL_CALL_HOST_FUNCTION ? "EDL_OCALL" : "OE_OCALL",
```
letmaik: +1. And in a future PR we would extend `oe_ocall_str` to return the actual EDL function name instead of "OE_OCALL_CALL_HOST_FUNCTION".

host/sgx/calls.c
```diff
@@ -381,11 +381,10 @@ static const char* oe_ocall_str(oe_func_t ocall)
                                        "OE_OCALL_WRITE",
                                        "OE_OCALL_SLEEP",
                                        "OE_OCALL_GET_TIME",
-                                       "OE_OCALL_BACKTRACE_SYMBOLS"
+                                       "OE_OCALL_BACKTRACE_SYMBOLS",
```
letmaik: Since we're cleaning up, what do folks think of leaving out the "OE_OCALL_" prefix? It find it redundant and it makes log messages longer.

achamayou: I'm happy either way, I can also the value in keeping this transparent and greppable.

host/sgx/calls.c
achamayou: I had a look at this, there is no easy way to do it now, the names just aren't captured. We'd need to change the way generation works etc, this is definitely for another PR.

host/sgx/calls.c
jxyang: I am for removing them.

---
In reply to: [290203277](https://github.com/microsoft/openenclave/pull/1859#discussion_r290203277) [](ancestors = 290203277)

johnkord: I agree, I'm in favor of removing them.

jxyang: I think we could search based on the pattern "OE_OCALL: <func_name>". So the OE_CALL prefix here is not really needed to make the message greppable.

---
In reply to: [290226050](https://github.com/microsoft/openenclave/pull/1859#discussion_r290226050) [](ancestors = 290226050)

achamayou: Ok, I'll take them out then.

achamayou:
bors try

andschwa:
Hey, these commits don't seem to be linking to your GitHub account. Can you @achamayou please double-check your Git authorship and GitHub settings? Thanks!

achamayou:
@andschwa sorry, didn't see this before. Not sure what happened there, the commits were against a fork of the repo, I'll double check next time.

PR #1860, opened 2019-06-04
Enable backtrace on windows

mikbras:
This PR enables back-tracing of enclaves on Windows. Recall that this feature implements two Linux functions **backtrace** and **backtrace_symbols** (see manpage).

**oe_backtrace()** generates a back trace entirely within the enclave.
**oe_backtrace_symbols()** translates back-trace addresses to symbol names with the assistance of the host.
**oe_print_backtrace()** can be called by an enclave to print a back-trace.

This feature is a great help in debugging. For example, the **assert** macro **could** trivially be extended to call **oe_print_backtrace()**.

To enable debug-malloc, which uses this facility to print back-traces of unfreed blocks, use the **cmake** **-DUSE_DEBUG_MALLOC=1** option.


ionutbalutoiu:
@mikbras Looks like the previous runs had some transient errors. I'll keep an eye on the CI too see if they reproduce.

The PR is merged right now.

Issue #1843, opened 2019-05-30
Remove the assumption "hacks" from the add_enclave_test cmake function


johnkord:
Several people, while enabling or adding tests on Windows, have encountered issues with the add_enclave_test cmake function, many times due to two "hacks" that make (commonly) invalid assumptions. Moreover, the comments are out of date for these two hacks; we will continue to support the scenario where ELF enclaves are built on Linux and used on Windows.

Please refer to these two comment blocks and code below:
https://github.com/microsoft/openenclave/blob/de692d7565c99a90fda919b9b6f953ce73efa2a2/cmake/add_enclave_test.cmake#L36
https://github.com/microsoft/openenclave/blob/de692d7565c99a90fda919b9b6f953ce73efa2a2/cmake/add_enclave_test.cmake#L44

CodeMonkeyLeet:
Consider deprecating this build configuration instead (build on linux and copy tests to Windows)

anitagov:
As already pointed out in PR #1862 , I have found this feature very useful when enabling tests on Windows, especially 3rd party libraries when issues are encountered using the cross-compiler. Also great to be able to easily test if an ELF enclave will work on Windows.

CodeMonkeyLeet:
Per closed PR #1862, instead of removing the build flag entirely, it needs to be repurposed:

> My original thinking was that the USE_CLANGW flag would suffice, but that means that the only way the ctests get added is if you also did the cross compile build.
>
> Instead of reverting this though, I suggest that we add a new flag to only build the windows components:
>
> - It would be mutually exclusive to USE_CLANGW
> - It would skip test enclaves and anything else that relied on bash, such as the test/crypto/data)
> - It would require doing a pass through the CMakeLists.txt in the repo to respect this new flag.

PR #1862, opened 2019-06-04
Remove ADD_WINDOWS_ENCLAVE_TESTS build option.

johnkord:
This removes the stages from Jenkins where the ELF test enclaves are built on Linux and then copied to Windows to be run. It also removes the infrastructure for running these tests, which was used when ADD_WINDOWS_ENCLAVE_TESTS is set to ON.

Because of these changes, "add_enclave_test" is now just a basic wrapper for "add_test" in all cases. I'd simply remove that cmake function and replace all references of add_enclave_test to add_test, but I think it might complicate a lot of ongoing work. I suggest doing that in another PR. I'll update issue #1843 accordingly if this PR is accepted.


cmake/add_enclave_test.cmake
```diff
@@ -6,70 +6,9 @@
...
-        add_custom_target(${TEST_NAME_WITHOUT_SLASH}.windows ALL
-            DEPENDS ${TEST_NAME}_windows_include
-            )
-
-        add_test(NAME ${TEST_NAME} COMMAND $<TARGET_FILE:${HOST_FILE}> ${CMAKE_CURRENT_BINARY_DIR}/${TEST_HOSTSUBPATH}/${TEST_ENCFILE} ${ARGN})
-
-    elseif (UNIX OR USE_CLANGW)
-        add_test(NAME ${TEST_NAME} COMMAND $<TARGET_FILE:${HOST_FILE}> $<TARGET_FILE:${ENC_FILE}> ${ARGN})
-    endif()
+    add_test(NAME ${TEST_NAME} COMMAND $<TARGET_FILE:${HOST_FILE}> $<TARGET_FILE:${ENC_FILE}> ${ARGN})
```
CodeMonkeyLeet: Reading through this, I realized that @anitagov was correct and this would break the manual scenario she wanted to retain where the ability to build purely windows bits and then copy over linux enclave bits and run the tests.

My original thinking was that the `USE_CLANGW` flag would suffice, but that means that the only way the ctests get added is if you also did the cross compile build.

Instead of reverting this though, I suggest that we add a new flag to only build the windows components:
- It would be mutually exclusive to USE_CLANGW
- It would skip test enclaves and anything else that relied on bash, such as the test/crypto/data)
- It would require doing a pass through the CMakeLists.txt in the repo to respect this new flag.

anitagov: Thanks @CodeMonkeyLeet . This feature is very handy to enable ELF enclaves on Windows, especially the 3rd party libraries when issues are encountered using the cross-compiler.
Like the new flag suggested by @CodeMonkeyLeet

yakman2020:
@johnkord Status on this?

jhand2:
It seems we maybe want to retain this scenario. Can this PR be closed?

CodeMonkeyLeet:
@jhand2 Agreed. I'm closing out this PR and letting #1843 track the preferred alternative to change the build option to permit building linux components separately and performing a manual binary drop.

PR #1863, opened 2019-06-04
Refactor use of `get_deepcopy`

andschwa:
This removes the extraneous use of `structs` and `ep`, and refactors `get_deepcopy` to be a bit more readable (I hope).

### DIFF
--- a/tools/oeedger8r/Emitter.ml
+++ b/tools/oeedger8r/Emitter.ml
@@ -290,12 +290,6 @@ let emit_composite_type =
   | UnionDef u -> emit_union u
   | EnumDef e -> emit_enum e
 
-(** This transforms the [Struct] types embedded in a [composite_type
-   list] into an association list of the struct name to its parameter
-   list, which is much easier to deal with. *)
-let get_structs =
-  filter_map (function StructDef s -> Some (s.sname, s.smlist) | _ -> None)
-
 (** Generate a cast expression for a pointer argument. Pointer
     arguments need to be cast to their root type, since the marshalling
     struct has the root pointer. For example:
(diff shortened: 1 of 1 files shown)


tools/oeedger8r/Emitter.ml
```diff
@@ -451,30 +445,45 @@ let warn_size_and_count_params (fd : func_decl) =
...
+      | Ptr p ->
+          (* [ec.comp_defs] is a list of all composite types, but we're
+             only interested in the structs, so we filter out the rest
+             and unwrap them from [composite_type]. *)
+          let structs =
+            filter_map
+              (function StructDef s -> Some s | _ -> None)
+              ec.comp_defs
+          in
+          let is_match name s = s.sname = name in
```
andschwa: @anakrish Can this be cleaned up? Without `List.assoc_opt` or `List.find_opt`, this is about the best I could do.

tools/oeedger8r/Emitter.ml
```diff
@@ -655,15 +663,15 @@ let gen_enclave_code (ec : enclave_content) (ep : edger8r_params) =
...
+             (get_deepcopy (get_param_atype ptype))) ]
         |> List.flatten
       in
       let params =
         flatten_map (gen_serialize "" "1")
           (List.filter (fun (p, _) -> is_in_ptr p || is_inout_ptr p) plist)
       in
       (* Note that the indentation for the first line is applied by the
-     parent function. *)
+         parent function. *)
```
andschwa: Oof, `ocamlformat` doesn't fix comment indentations, but I accidentally triggered an auto-fill which went and fixed all the moved comments. Just gonna leave it.

tools/oeedger8r/Emitter.ml
anakrish: Reading the code, the 'structs' expression depends only on ec.comp_defs, so it could be moved outside of get_deepcopy and thus be computed only once.
There is no guarantee that Ocaml itself will move the evaluation of 'structs' outside get_deepcopy.

I feel moving it out would make it more readable.
Also have a function get_struct_by_name that looks up the structs given a name.

anakrish: With a get_struct_by_name function, it seems like should_deepcopy_a can be made more declarative:

let shoud_deepcopy_a = function
| Ptr (Struct n) -> get_struct_by_name n
| Ptr (Foreign n) -> get_struct_by_name n
| -> None

I would suggest renaming this function since the function name suggests a bool return type contrary to the implementation

andschwa: Omg I forgot about nesting pattern matching. TYVM.

andschwa: > Reading the code, the 'structs' expression depends only on ec.comp_defs, so it could be moved outside of get_deepcopy and thus be computed only once.
> There is no guarantee that Ocaml itself will move the evaluation of 'structs' outside get_deepcopy.
> 
> I feel moving it out would make it more readable.

I feel like this is just premature optimization though. We're certainly not running up against noticeable runtimes with the edger8r. I'll make some changes here, but going for readability over performance optimizations.

anakrish: I'm more interested in readability and scoping functions appropriately would make it more readable.

andschwa:
Part of #1839.

PR #1869, opened 2019-06-05
Fix extraneous uses of `sprintf` in edger8r

andschwa:
Substituting instead the concatenation operator where it makes the code
more readable.


andschwa:
Well, that's annoying.

andschwa:
Part of #1839.

PR #1871, opened 2019-06-05
Add support for recursive calls to oe_once/pthread_once.

johnkord:
This change addresses one scenario I encountered when integrating the OpenSSL library into Open Enclave: I noticed my enclave hanging during the call to oe_initialize_openssl (which calls pthread_once), where the hanging never terminated. This hang was partially due to this local static lock:

```C
static oe_spinlock_t _lock = OE_SPINLOCK_INITIALIZER;
```

This is an issue for OpenSSL because when invoking pthread_once to initialize OpenSSL's library, OpenSSL itself will also call pthread_once on several of its internal initialization functions. Our invocation of pthread_once will acquire the spinlock "_lock", and then when OpenSSL's pthread_once invocations attempt to acquire this lock, they will hang forever. Here is what a backtrace looks like without this change:

```gdb
(gdb) run
Starting program: /home/johnkord/openenclave/build/tests/crypto/enclave/host/cryptohost ./enc/cryptoenc
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
=== begin _test_asn1_parsing()
^C
Program received signal SIGINT, Interrupt.
oe_spin_lock (spinlock=0x7ffff25a464c <oe_once._lock>) at /home/johnkord/openenclave/enclave/core/sgx/spinlock.c:43
43              while (*spinlock)
(gdb) where
#0  oe_spin_lock (spinlock=0x7ffff25a464c <oe_once._lock>) at /home/johnkord/openenclave/enclave/core/sgx/spinlock.c:43
#1  0x00007ffff228f4f0 in oe_once (once=0x7ffff25a12f4 <base>, func=0x7ffff2120a80 <ossl_init_base_ossl_>) at /home/johnkord/openenclave/enclave/core/sgx/once.c:22
#2  0x00007ffff225f846 in oe_pthread_once (once=0x7ffff25a12f4 <base>, func=0x7ffff2120a80 <ossl_init_base_ossl_>) at /home/johnkord/openenclave/enclave/core/pthread.c:97
#3  0x00007ffff20cdf76 in pthread_once (once=0x7ffff25a12f4 <base>, func=0x7ffff2120a80 <ossl_init_base_ossl_>)
    at /home/johnkord/openenclave/include/openenclave/corelibc/bits/pthread_once.h:10
#4  0x00007ffff214a986 in CRYPTO_THREAD_run_once ()
#5  0x00007ffff212059d in OPENSSL_init_crypto ()
#6  0x00007ffff209ca78 in _initialize () at /home/johnkord/openenclave/enclave/init.c:13
#7  0x00007ffff228f57f in oe_once (once=0x7ffff259f3f0 <_once>, func=0x7ffff209ca60 <_initialize>) at /home/johnkord/openenclave/enclave/core/sgx/once.c:27
#8  0x00007ffff225f846 in oe_pthread_once (once=0x7ffff259f3f0 <_once>, func=0x7ffff209ca60 <_initialize>) at /home/johnkord/openenclave/enclave/core/pthread.c:97
#9  0x00007ffff20cdf76 in pthread_once (once=0x7ffff259f3f0 <_once>, func=0x7ffff209ca60 <_initialize>) at /home/johnkord/openenclave/include/openenclave/corelibc/bits/pthread_once.h:10
#10 0x00007ffff209ca4b in oe_initialize_openssl () at /home/johnkord/openenclave/enclave/init.c:20
#11 0x00007ffff208cf67 in oe_cert_read_pem (cert=0x7ffff2de0820, pem_data=0x7ffff2590980 <_CERT>, pem_size=1351) at /home/johnkord/openenclave/enclave/cert.c:464
#12 0x00007ffff2070acf in _test_asn1_parsing () at /home/johnkord/openenclave/tests/crypto/asn1_tests.c:337
#13 0x00007ffff20708f9 in TestASN1 () at /home/johnkord/openenclave/tests/crypto/asn1_tests.c:368
#14 0x00007ffff2082cb9 in TestAll () at /home/johnkord/openenclave/tests/crypto/tests.c:9
#15 0x00007ffff206f1fd in test () at /home/johnkord/openenclave/tests/crypto/enclave/enc/enc.c:47
#16 0x00007ffff208351c in ecall_test (input_buffer=0x7ffff25e0790 "", input_buffer_size=16, output_buffer=0x7ffff25e07a0 "", output_buffer_size=16, output_bytes_written=0x7ffff2de0c78)
    at /home/johnkord/openenclave/build/tests/crypto/enclave/enc/crypto_t.c:50
#17 0x00007ffff22675cf in _handle_call_enclave_function (arg_in=140737488347080) at /home/johnkord/openenclave/enclave/core/sgx/calls.c:326
#18 0x00007ffff22661b7 in _handle_ecall (td=0x7ffff2de6000, func=2, arg_in=140737488347080, output_arg1=0x7ffff2de0fc8, output_arg2=0x7ffff2de0fc0)
    at /home/johnkord/openenclave/enclave/core/sgx/calls.c:432
#19 0x00007ffff2265b59 in __oe_handle_main (arg1=281483566645248, arg2=140737488347080, cssa=0, tcs=0x7ffff2de2000, output_arg1=0x7ffff2de0fc8, output_arg2=0x7ffff2de0fc0)
    at /home/johnkord/openenclave/enclave/core/sgx/calls.c:861
#20 0x00007ffff229765b in oe_enter () at /home/johnkord/openenclave/enclave/core/sgx/enter.S:162
#21 0x0000000000428cb1 in __morestack () at /home/johnkord/openenclave/host/sgx/linux/enter.S:95
#22 0x000000000040af37 in _do_eenter (enclave=0x67f100, tcs=0x7ffff2de2000, aep=0x428b24 <OE_AEP>, code_in=OE_CODE_ECALL, func_in=2, arg_in=140737488347080, code_out=0x7fffffffdde0,
    func_out=0x7fffffffddde, result_out=0x7fffffffdddc, arg_out=0x7fffffffddd0) at /home/johnkord/openenclave/host/sgx/calls.c:231
#23 oe_ecall (enclave=0x67f100, func=2, arg=140737488347080, arg_out_ptr=0x7fffffffdfc0) at /home/johnkord/openenclave/host/sgx/calls.c:669
#24 0x000000000040bd65 in _call_enclave_function (is_internal_call=false, enclave=0x67f100, function_id=0, input_buffer=0x68ccf0, input_buffer_size=16, output_buffer=0x68cd00,
    output_buffer_size=16, output_bytes_written=0x7fffffffe1a8) at /home/johnkord/openenclave/host/sgx/calls.c:748
#25 0x000000000040bb3a in oe_call_enclave_function (enclave=0x67f100, function_id=0, input_buffer=0x68ccf0, input_buffer_size=16, output_buffer=0x68cd00, output_buffer_size=16,
    output_bytes_written=0x7fffffffe1a8) at /home/johnkord/openenclave/host/sgx/calls.c:775
#26 0x0000000000408011 in test (enclave=0x67f100) at /home/johnkord/openenclave/build/tests/crypto/enclave/host/crypto_u.c:67
#27 0x0000000000407aec in main (argc=2, argv=0x7fffffffe3d8) at /home/johnkord/openenclave/tests/crypto/enclave/host/host.c:58
```

The solution to this problem was designed by @anakrish and we both pair-programmed it earlier this year. In short, this implementation uses the oe_once_t (i.e. pthread_once_t) variable as the spinlock instead of a local static spinlock, and it uses the atomic function `__sync_val_compare_and_swap` to handle that variable safely.

This solution seems logical to me, but I'm mostly inexperienced when it comes to low level locks like this, and I've never used pthread_once before this change. So definitely, please point out if any invalid assumptions that have been made.

Questions I have:
1. Is OE_ATOMIC_MEMORY_BARRIER_RELEASE in the right location? Originally, I had it inside one of the conditions, but I don't know exactly what that operation does. It seems logical that if each potential thread invokes the ACQUIRE operation, it should probably also invoke the RELEASE, but I have no idea. Double check me here, please :)
2. Is there any harm in using a pthread_once_t variable like this? If we can't, then I'm not really sure how to solve this problem, as we can't extend on the pthread_once_t type with oe_once_t, because OpenSSL allocates/manages the memory of their own pthread_once_t variables. I assume other third party libraries would also have the same expectations
3. Are there any better atomic functions to call besides this `__sync_val_compare_and_swap` ? Here is the reference we used when designing/implementing this: https://gcc.gnu.org/onlinedocs/gcc-7.4.0/gcc/_005f_005fsync-Builtins.html . These atomic functions are referred to as "legacy" now, so there are probably modern functions we can invoke instead.

### DIFF
--- a/enclave/core/sgx/once.c
+++ b/enclave/core/sgx/once.c
@@ -14,6 +14,7 @@ oe_result_t oe_once(oe_once_t* once, void (*func)(void))
     if (!once)
         return OE_INVALID_PARAMETER;
 
+    oe_once_t status = *once;
     /* Double checked locking (DCLP). */
     /* DCLP Acquire barrier. */
     OE_ATOMIC_MEMORY_BARRIER_ACQUIRE();
@@ -25,7 +26,7 @@ oe_result_t oe_once(oe_once_t* once, void (*func)(void))
       function invocation to complete. Otherwise, this thread can try to take
       ownership of invoking the function
     */
-    if (*once != FUNC_INVOKED)
+    if (status != FUNC_INVOKED)
     {
         /*
           Multiple threads could reach here simultaneously after checking
--- a/enclave/core/sgx/once.c
+++ b/enclave/core/sgx/once.c
@@ -14,10 +14,10 @@ oe_result_t oe_once(oe_once_t* once, void (*func)(void))
     if (!once)
         return OE_INVALID_PARAMETER;
 
-    oe_once_t status = *once;
     /* Double checked locking (DCLP). */
     /* DCLP Acquire barrier. */
-    OE_ATOMIC_MEMORY_BARRIER_ACQUIRE();
+
+    oe_once_t status = __atomic_load_n(once, __ATOMIC_ACQUIRE);
 
     /*
       Use an atomic-acquire load operation to check if the function has not been
@@ -36,38 +36,42 @@ oe_result_t oe_once(oe_once_t* once, void (*func)(void))
           To determine who gets to invoke the function, each thread atomically
           compares the value of once to FUNC_NOT_INVOKED and if equal, sets the
           value to FUNC_BEING_INVOKED to signal to other threads that the
-          function is being invoked. The return value of
-          __sync_val_compare_and_swap determines which thread takes ownership.
-          If __sync_val_compare_and_swap returns FUNC_NOT_INVOKED, then that
-          means this thread successfully set once to FUNC_BEING_INVOKED, and can
-          now safely call func. If __sync_val_compare_and_swap returns
-          FUNC_BEING_INVOKED, this means another thread's
-          __sync_val_compare_and_swap succeeded first and therefore this thread
-          now has to wait for the other thread to complete (ie wait for once to
-          become FUNC_INVOKED. If __sync_val_compare_and_swap returns
-          FUNC_INVOKED, that means another thread has already called the
-          function and marked the once as complete. This thread can safely
-          proceed.
+          function is being invoked.
+
+          If the compare and exchange succeeds, then this thread owns the
+          responsibility of invoking the function.
+
+          If the compar and exchange fails, then another thread has taken
+          ownership of calling the function and therefore this thread must
+          wait for the other thread to complete the invocation.
+
         */
-        oe_once_t retval = __sync_val_compare_and_swap(
-            once, FUNC_NOT_INVOKED, FUNC_BEING_INVOKED);
-        if (retval == FUNC_NOT_INVOKED)
+        oe_once_t expected = FUNC_NOT_INVOKED;
+        bool retval = __atomic_compare_exchange_n(
+            once,
+            &expected, // If the current value of once if FUNC_NOT_INVOKED
+            FUNC_BEING_INVOKED, // take ownership of calling the function.
+            false,              // Use a strong compare exchange.
+            __ATOMIC_RELEASE,   // If exchange was successful, use release
+                                // barrier.
+            __ATOMIC_ACQUIRE);  // Otherwise use acquire barrier.
+
+        if (retval)
         {
             if (func)
                 func();
 
-            OE_ATOMIC_MEMORY_BARRIER_RELEASE();
-            *once = FUNC_INVOKED;
+            // Inform other threads that func has completed by setting it to
+            // FUNC_INVOKED. Use a release barrier.
+            __atomic_store_n(once, FUNC_INVOKED, __ATOMIC_RELEASE);
         }
-        else if (retval == FUNC_BEING_INVOKED)
+        else
         {
             /*
               Another thread is invoking the function. Wait for that thread to
-              finish the invocation and mark the once variable to FUNC_INVOKED.
+              finish the invocation.
             */
-            while (__sync_val_compare_and_swap(
-                       once, FUNC_BEING_INVOKED, FUNC_BEING_INVOKED) !=
-                   FUNC_BEING_INVOKED)
+            while (__atomic_load_n(once, __ATOMIC_ACQUIRE) != FUNC_INVOKED)
             {
                 // Relinquish CPU
                 asm volatile("pause");
--- a/enclave/core/sgx/once.c
+++ b/enclave/core/sgx/once.c
@@ -5,9 +5,9 @@
 #include <openenclave/internal/thread.h>
 #include <openenclave/internal/utils.h>
 
-#define SPINLOCK_NOT_USED 0
-#define SPINLOCK_ACTIVELY_HELD 1
-#define SPINLOCK_USED 2
+#define FUNC_NOT_INVOKED 0
+#define FUNC_BEING_INVOKED 1
+#define FUNC_INVOKED 2
 
 oe_result_t oe_once(oe_once_t* once, void (*func)(void))
 {
@@ -17,27 +17,61 @@ oe_result_t oe_once(oe_once_t* once, void (*func)(void))
     /* Double checked locking (DCLP). */
     /* DCLP Acquire barrier. */
     OE_ATOMIC_MEMORY_BARRIER_ACQUIRE();
-    if (*once != SPINLOCK_USED)
+
+    /*
+      Use an atomic-acquire load operation to check if the function has not been
+      invoked. If the function has already been invoked, there is nothing to do.
+      If the function is being invoked, then this thread must wait for the
+      function invocation to complete. Otherwise, this thread can try to take
+      ownership of invoking the function
+    */
+    if (*once != FUNC_INVOKED)
     {
+        /*
+          Multiple threads could reach here simultaneously after checking
+          whether the function has been invoked or not. Only one of them must
+          now invoke the function and others must wait for the function
+          invocation to complete.
+          To determine who gets to invoke the function, each thread atomically
+          compares the value of once to FUNC_NOT_INVOKED and if equal, sets the
+          value to FUNC_BEING_INVOKED to signal to other threads that the
+          function is being invoked. The return value of
+          __sync_val_compare_and_swap determines which thread takes ownership.
+          If __sync_val_compare_and_swap returns FUNC_NOT_INVOKED, then that
+          means this thread successfully set once to FUNC_BEING_INVOKED, and can
+          now safely call func. If __sync_val_compare_and_swap returns
+          FUNC_BEING_INVOKED, this means another thread's
+          __sync_val_compare_and_swap succeeded first and therefore this thread
+          now has to wait for the other thread to complete (ie wait for once to
+          become FUNC_INVOKED. If __sync_val_compare_and_swap returns
+          FUNC_INVOKED, that means another thread has already called the
+          function and marked the once as complete. This thread can safely
+          proceed.
+        */
         oe_once_t retval = __sync_val_compare_and_swap(
-            once, SPINLOCK_NOT_USED, SPINLOCK_ACTIVELY_HELD);
-        if (retval == SPINLOCK_NOT_USED)
+            once, FUNC_NOT_INVOKED, FUNC_BEING_INVOKED);
+        if (retval == FUNC_NOT_INVOKED)
         {
             if (func)
                 func();
 
-            *once = SPINLOCK_USED;
+            OE_ATOMIC_MEMORY_BARRIER_RELEASE();
+            *once = FUNC_INVOKED;
         }
-        else if (retval == SPINLOCK_ACTIVELY_HELD)
+        else if (retval == FUNC_BEING_INVOKED)
         {
+            /*
+              Another thread is invoking the function. Wait for that thread to
+              finish the invocation and mark the once variable to FUNC_INVOKED.
+            */
             while (__sync_val_compare_and_swap(
-                       once, SPINLOCK_ACTIVELY_HELD, SPINLOCK_ACTIVELY_HELD) !=
-                   SPINLOCK_ACTIVELY_HELD)
+                       once, FUNC_BEING_INVOKED, FUNC_BEING_INVOKED) !=
+                   FUNC_BEING_INVOKED)
             {
+                // Relinquish CPU
                 asm volatile("pause");
             }
         }
     }
-    OE_ATOMIC_MEMORY_BARRIER_RELEASE();
     return OE_OK;
 }
--- a/enclave/core/sgx/once.c
+++ b/enclave/core/sgx/once.c
@@ -5,34 +5,39 @@
 #include <openenclave/internal/thread.h>
 #include <openenclave/internal/utils.h>
 
+#define SPINLOCK_NOT_USED 0
+#define SPINLOCK_ACTIVELY_HELD 1
+#define SPINLOCK_USED 2
+
 oe_result_t oe_once(oe_once_t* once, void (*func)(void))
 {
     if (!once)
         return OE_INVALID_PARAMETER;
 
     /* Double checked locking (DCLP). */
-    oe_once_t o = *once;
-
     /* DCLP Acquire barrier. */
     OE_ATOMIC_MEMORY_BARRIER_ACQUIRE();
-    if (o == 0)
+    if (*once != SPINLOCK_USED)
     {
-        static oe_spinlock_t _lock = OE_SPINLOCK_INITIALIZER;
-
-        oe_spin_lock(&_lock);
-
-        if (*once == 0)
+        oe_once_t retval = __sync_val_compare_and_swap(
+            once, SPINLOCK_NOT_USED, SPINLOCK_ACTIVELY_HELD);
+        if (retval == SPINLOCK_NOT_USED)
         {
             if (func)
                 func();
 
-            /* DCLP Release barrier. */
-            OE_ATOMIC_MEMORY_BARRIER_RELEASE();
-            *once = 1;
+            *once = SPINLOCK_USED;
         }
-
-        oe_spin_unlock(&_lock);
+        else if (retval == SPINLOCK_ACTIVELY_HELD)
+        {
+            while (__sync_val_compare_and_swap(
+                       once, SPINLOCK_ACTIVELY_HELD, SPINLOCK_ACTIVELY_HELD) !=
+                   SPINLOCK_ACTIVELY_HELD)
+            {
+                asm volatile("pause");
+            }
+        }
     }
-
+    OE_ATOMIC_MEMORY_BARRIER_RELEASE();
     return OE_OK;
 }


enclave/core/sgx/once.c
```diff
@@ -5,34 +5,39 @@
 #include <openenclave/internal/thread.h>
 #include <openenclave/internal/utils.h>
 
+#define SPINLOCK_NOT_USED 0
```
anakrish: Consider renaming these to
FUNC_NOT_INVOKED,
FUNC_BEING_INVOKED,
FUNC_INVOKED

if that reads better.

enclave/core/sgx/once.c
```diff
@@ -5,34 +5,39 @@
...
     if (!once)
         return OE_INVALID_PARAMETER;
 
     /* Double checked locking (DCLP). */
-    oe_once_t o = *once;
-
     /* DCLP Acquire barrier. */
     OE_ATOMIC_MEMORY_BARRIER_ACQUIRE();
-    if (o == 0)
+    if (*once != SPINLOCK_USED)
```
anakrish: "Use an atomic-acquire load operation to check if the function has not been invoked.
If the function has already been invoked, there is nothing to do.
If the function is being invoked, then this thread must wait for the function invocation to complete.
Otherwise, this thread can try to take ownership of invoking the function".

enclave/core/sgx/once.c
```diff
@@ -5,34 +5,39 @@
...
     {
-        static oe_spinlock_t _lock = OE_SPINLOCK_INITIALIZER;
-
-        oe_spin_lock(&_lock);
-
-        if (*once == 0)
+        oe_once_t retval = __sync_val_compare_and_swap(
```
anakrish: "Multiple threads could reach here simultaneously after checking whether the function has been invoked or not.
Only one of them must now invoke the function and others must wait for the function invocation to complete.

To determine who gets to invoke the function, each thread atomically compares the value of once to FUNC_NOT_INVOKED and if equal, sets the value to FUNC_BEING_INVOKED to signal to other threads that the function is being invoked. There return value of __sync_val_compare_and_swap determines which thread takes ownership.
If __sync_val_compare_and_swap returns FUNC_NOT_INVOKED, then that means this thread successfully set once to FUNC_BEING_INVOKED, and can now safely call func.
If __sync_val_compare_and_swap returns FUNC_BEING_INVOKED, this means another thread's __sync_val_compare_and_swap succeeded first and therefore this thread now has to wait for the other thread to complete (ie wait for once to become FUNC_INVOKED.
If __sync_val_compare_and_swap returns FUNC_INVOKED, that means another thread has already called the function and marked the once as complete. This thread can safely proceed."

enclave/core/sgx/once.c
```diff
@@ -5,34 +5,39 @@
...
+            once, SPINLOCK_NOT_USED, SPINLOCK_ACTIVELY_HELD);
+        if (retval == SPINLOCK_NOT_USED)
         {
             if (func)
                 func();
 
-            /* DCLP Release barrier. */
-            OE_ATOMIC_MEMORY_BARRIER_RELEASE();
-            *once = 1;
+            *once = SPINLOCK_USED;
```
anakrish: This line ought to be preceded by 
OE_ATOMIC_MEMORY_BARRIER_RELEASE();

The concept to acquire-release is this:
Say, func all func does is set g to some value;

void func()
{
    g = MAGIC;
}

Eventhough, as written, the order is:
    g = MAGIC;
   *once = SPINLOCK_USED; // or FUNC_INVOKED
the compiler, could reorder this to:
   *once = SPINLOCK_USED;
    g = MAGIC
Even if compiler preserves the order, another thread could see it in the incorrect order when contents of one processor core is flushed and synced to another core. That is another core could see these variables being written out of order.
So, these ACQUIRE/RELESE macros insert special memory barriers so that different cores in a multi-core system see the writes in the correct order.

Note: x64 is strongly ordered, and these marcros are no-ops. But other processors are not.

enclave/core/sgx/once.c
```diff
@@ -5,34 +5,39 @@
...
+        }
+        else if (retval == SPINLOCK_ACTIVELY_HELD)
+        {
+            while (__sync_val_compare_and_swap(
```
anakrish: "Another thread is invoking the function. Wait for that thread to finish the invocation and mark the once variable to FUNC_INVOKED".

enclave/core/sgx/once.c
```diff
@@ -5,34 +5,39 @@
...
+                       once, SPINLOCK_ACTIVELY_HELD, SPINLOCK_ACTIVELY_HELD) !=
+                   SPINLOCK_ACTIVELY_HELD)
+            {
+                asm volatile("pause");
```
anakrish: "Relinquish CPU"

enclave/core/sgx/once.c
```diff
@@ -5,34 +5,39 @@
...
+            }
         }
-
-        oe_spin_unlock(&_lock);
     }
-
+    OE_ATOMIC_MEMORY_BARRIER_RELEASE();
```
anakrish: This seems like incorrect place for release.

johnkord:
Thanks so much, Anand. I'll incorporate your suggested changes!

enclave/core/sgx/once.c
johnkord: I just made your suggested change for this, but I'm going to leave this comment unresolved for the really great & important information you left here about concerns regarding ordering of operations in this scenario.

enclave/core/sgx/once.c
```diff
@@ -5,34 +5,73 @@
...
+    {
+        /*
+          Multiple threads could reach here simultaneously after checking
+          whether the function has been invoked or not. Only one of them must
+          now invoke the function and others must wait for the function
+          invocation to complete.
+          To determine who gets to invoke the function, each thread atomically
+          compares the value of once to FUNC_NOT_INVOKED and if equal, sets the
+          value to FUNC_BEING_INVOKED to signal to other threads that the
+          function is being invoked. There return value of
```
anitagov: Change "There return value of" to "The return value"

anitagov: >Are there any better atomic functions to call besides this __sync_val_compare_and_swap ? Here is the reference we used when designing/implementing this: https://gcc.gnu.org/onlinedocs/gcc-7.4.0/gcc/_005f_005fsync-Builtins.html 

This function call seems compatible to the Intel Itanium Processor-specific Application Binary Interface, section 7.4 and not sure if they conform to the Intel IA86/64 ABI. Will investigate some more and update later.

enclave/core/sgx/once.c
```diff
@@ -5,34 +5,73 @@
...
 oe_result_t oe_once(oe_once_t* once, void (*func)(void))
 {
     if (!once)
         return OE_INVALID_PARAMETER;
 
     /* Double checked locking (DCLP). */
-    oe_once_t o = *once;
-
     /* DCLP Acquire barrier. */
     OE_ATOMIC_MEMORY_BARRIER_ACQUIRE();
```
gupta-ak: What does the memory barrier do in the new code? The original code had a `oe_once_t o = *once` statement before so the memory barrier would prevent the statement reordering.

enclave/core/sgx/once.c
```diff
@@ -5,34 +5,73 @@
...
+        */
+        oe_once_t retval = __sync_val_compare_and_swap(
+            once, FUNC_NOT_INVOKED, FUNC_BEING_INVOKED);
+        if (retval == FUNC_NOT_INVOKED)
         {
             if (func)
                 func();
 
-            /* DCLP Release barrier. */
             OE_ATOMIC_MEMORY_BARRIER_RELEASE();
```
gupta-ak: The memory barrier in this case would ensure that the side effects from `func()` is observed before `*once = ...` is called right?

johnkord: If I understand correctly (and this is a very new concept to me, so I might be wrong), I think the memory barrier intends on making these two calls to be un-reorderable in a multithreaded scenario:
```
oe_once_t retval = __sync_val_compare_and_swap(once, FUNC_NOT_INVOKED, FUNC_BEING_INVOKED);
func();
```

Is it important that the first 'if' condition be in the memory barrier too? Specifically this condition:
```
if (*once != FUNC_INVOKED)
```

I imagine that the barrier probably does not need to be above that line. I'm not sure as I'm definitely not an expert on this topic!

enclave/core/sgx/once.c
anakrish: It makes sure that all the writes that happen in other threads before write to the *once variable happen (i.e observed by this thread), before the *once variable is read by this thread.

A good reference is C++11 memory ordering which describes the semantics more formally https://en.cppreference.com/w/cpp/atomic/memory_order:

> memory_order_acquire A load operation with this memory order performs the acquire operation on the affected memory location: no reads or writes in the current thread can be reordered before this load. All writes in other threads that release the same atomic variable are visible in the current thread (see Release-Acquire ordering below)

> memory_order_release : no reads or writes in the current thread can be reordered after this store. All writes in the current thread are visible in other threads that acquire the same atomic variable (see Release-Acquire ordering below) and writes that carry a dependency into the atomic variable become visible in other threads that consume the same atomic (see Release-Consume ordering below).

enclave/core/sgx/once.c
anakrish: @gupta-ak Good catch! We need to fix the implementation.

enclave/core/sgx/once.c
anitagov: @johnkord - Use the modern version i.e. __atomic_compare_exchange_n or  __atomic_compare_exchange. Here is the reference. https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html
Trust this helps.

enclave/core/sgx/once.c
```diff
@@ -5,34 +5,74 @@
...
+#define FUNC_NOT_INVOKED 0
+#define FUNC_BEING_INVOKED 1
+#define FUNC_INVOKED 2
+
 oe_result_t oe_once(oe_once_t* once, void (*func)(void))
 {
     if (!once)
         return OE_INVALID_PARAMETER;
 
+    oe_once_t status = *once;
```
anitagov: The optimizer will optimize **status** entirely out here and in line 29 and use the value passed on the stack. If you don't want that to happen, prefix this line with volatile.

gupta-ak: While it's true that volatile will prevent it being optimized out, there's a memory barrier right under the statement, so it already won't be optimized out.

Note on architectures with a weak memory model (i.e. ARM), volatile would not be sufficient, since you need to do a load-acquire., making the volatile redundant.

anakrish: @anitagov I agree with @gupta-ak

enclave/core/sgx/once.c
```diff
@@ -5,34 +5,74 @@
...
+            *once = FUNC_INVOKED;
+        }
+        else if (retval == FUNC_BEING_INVOKED)
+        {
+            /*
+              Another thread is invoking the function. Wait for that thread to
+              finish the invocation and mark the once variable to FUNC_INVOKED.
+            */
+            while (__sync_val_compare_and_swap(
+                       once, FUNC_BEING_INVOKED, FUNC_BEING_INVOKED) !=
```
gupta-ak: Based of the comment, shouldn't this be:
`while (!__sync_bool_compare_and_swap(once, FUNC_INVOKED, FUNC_INVOKED))`? (i.e CAS loop until it *once == FUNC_INVOKED).

It seems that the current logic would fail in the following condition:
```C
Thread1: else if (retval == FUNC_BEING_INVOKED)
Thread2: *once = FUNC_INVOKED
Thread1: while (__sync_val_compare_and_swap(once, FUNC_BEING_INVOKED, FUNC_BEING_INVOKED) != FUNC_BEING_INVOKED) // infinite loop since it will always return FUNC_INVOKED.
```

anakrish: @gupta-ak

Yes you are right.
 The intent was that if a thread sees retval == FUNC_BEING_INVOKED, then it needs to wait for the invocation to complete. This could be expressed as:

- Wait until *once == FUNC_INVOKED
- Or wait while *once == FUNC_BEING_INVOKED

I don't remember which one we were trying to code. The first one is more easier to read.
So we should use the loop you provided.

Is there a better way to express _loop until *once == FUNC_INVOKED_ ? 
Can we somehow hide the compare and swap and expresses it as an _atomic compare_ or _atomic load and then compare_?

enclave/core/sgx/once.c
```diff
@@ -5,34 +5,74 @@
...
+                   FUNC_BEING_INVOKED)
+            {
+                // Relinquish CPU
+                asm volatile("pause");
+            }
```
anakrish: @gupta-ak 
Thanks for catching this; it gave me an opportunity to re-scrutinize this.

The thread that sees _retval == FUNC_BEING_INVOKED_, infers that another thread is calling func and therefore decides to wait for *once to become FUNC_INVOKED (i.e for the invocation to complete).

It also needs to make sure it sees the side effects of the func too and not just the latest value of *once.
So it would need an _acquire barrier_ on line 75 just before exiting the scope.

What do you think?

enclave/core/sgx/once.c
anitagov: This is being optimized out with the Compiler explorer (https://godbolt.org/) and hence I brought it up.

anakrish: @anitagov To be clear, when you say "optimized out" what do you mean?

Consider the following two functions where the value of _*once_ is read twice and added.
The second function has an _acquire barrier_ where s the first does not.

```c
#include <atomic>

int read_value(int* once)
{
    int local = *once;
    return local + *once;
}
int read_value_with_barrier(int* once)
{
    int local = *once;
    std::atomic_thread_fence(std::memory_order_acquire);
    return local + *once;
}
```
In the first function, the compiler sees that *once has already been read and reuses the value without reading the value again:

```asm
read_value(int*): 
    movl (%rdi), %eax
    addl %eax, %eax    # Reuse value in eax
    retq
```

For the second function, the _acquire-barrier_ asks the compiler to NOT move any reads after the barrier to before the barrier. Therefore the compiler emits code to re-read the value of _*once_.

```asm
read_value_with_barrier(int*): 
     movl (%rdi), %eax
     addl (%rdi), %eax     # Reread the value of *rdi i.e *once
     retq
```
You can see this example here: https://godbolt.org/z/t9VWD7

x64 is a strongly ordered acrchitecture and the compiler does not need to emit any special assembly instructions for the _acquire-barrier_.

However, x64 is an exception and special instructions are needed for ARM, Power PC, RISC and other architectures to ensure the _acquire_ semantics.

In ARM64 for example, in addition to re-reading the value of *once, the compiler also emits a [Data Memory Barrier  Instruction](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0489c/CIHGHHIE.html)

```asm
read_value(int*):
    ldr w0, [x0]
    lsl w0, w0, 1   # Reuse the value in w0
    ret

read_value_with_barrier(int*):
    ldr w1, [x0]
    dmb ishld      # Acquire barrier 
    ldr w0, [x0]    # Reread *once
    add w0, w1, w0
    ret
```

You can see the example here: https://godbolt.org/z/AK72MU

If there is still confusion, can you share the link to what you are seeing in compiler explorer?(There is a share button).

johnkord:
bors r-

I'd like to wait for Anand and Akash to finish their discussion before merging this, in case they find anything that needs to change!

enclave/core/sgx/once.c
gupta-ak: I believe GCC's CAS already has an `acquire/release` barrier in the implementation. We would need to just check the disassembly.

enclave/core/sgx/once.c
gupta-ak: I think this works:
```
while (*((volatile oe_once_t*) once) != FUNC_INVOKED) {}
ACQUIRE_BARRIER
```

Just we have to make sure that the for loop doesn't get optimized and that there's an acquire barrier to synchronize with the release barrier.

yakman2020:
It needs to be decided today. If its not in it has to be out.

My impression is anands suggestions could easily be applied later.

yakman2020:
Is there a use case where this fails?

yakman2020:
@akash, could you resolve this this afternoon? We are trying to get a release out and I suspect we want this change to be in, rather than pushed to 0.7

enclave/core/sgx/once.c
EMumau: GCC has deprecated the __sync_* line of intrinsic functions and replaced them with the __atomic_* line of functions.  The __atomic_* functions line up closely with the C++11 standards and allow you to explicitly declare the acquire/release barrier operations to use with each atomic operation that you call.

anitagov: As I mentioned 5 days back for the replacement to the _sync legacy function -
> @johnkord - Use the modern version i.e. __atomic_compare_exchange_n or __atomic_compare_exchange. Here is the reference. https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html
> Trust this helps.

anakrish: @EMumau Great suggestion. I have taken a stab at using the __atomic functions.
Can you do a quick review and also see if the comments look good as well?

@gupta-ak Can you also do a quick review? Let us wrap this up ASAP.

enclave/core/sgx/once.c
```diff
@@ -5,34 +5,78 @@
...
 
-        if (*once == 0)
+        */
+        oe_once_t expected = FUNC_NOT_INVOKED;
+        bool retval = __atomic_compare_exchange_n(
+            once,
+            &expected, // If the current value of once if FUNC_NOT_INVOKED
+            FUNC_BEING_INVOKED, // take ownership of calling the function.
+            false,              // Use a strong compare exchange.
+            __ATOMIC_RELEASE,   // If exchange was successful, use release
```
EMumau: I would use __ATOMIC_ACQ_REL for the success case and __ATOMIC_ACQUIRE for the fail case.  That's my only nit.  The rest of this code LGTM.

enclave/core/sgx/once.c
anitagov: Thanks @anakrish . I verified again with Compiler explorer that things were fine.

PR #1875, opened 2019-06-05
Isolate all I/O side effects into one location

andschwa:
This changes `open_file` to `write_file` which is the only function that
performs file I/O. Then, all the 'gen' functions simply return their
list of strings (becoming variables instead of functions), and it is
only at the end of `gen_enclave_code` that conditional file I/O is
performed.

This has the side effect (heh) of the string lists moving up an
indentation level. This change was required in order to remove the
arguments to the 'gen' functions, because when they turned into
variables, the were unconditionally evaluated, and the embedded side
effects became a problem.


tools/oeedger8r/Emitter.ml
```diff
@@ -1269,65 +1249,62 @@ let gen_enclave_code (ec : enclave_content) (ep : edger8r_params) =
...
+    ; "               enclave);"
+    ; "}"
+    ; ""
+    ; "OE_EXTERNC_END"
+    ; "" ]
   in
   (* NOTE: The below code is all I/O side effects to emit warnings or
      generate files. *)
-  validate_oe_support ep ;
+  validate_oe_support ;
```
anakrish: This seems off. validate_oe_support is no longer a function has probably has type unit.
It seems best to retain it as a function since we seem to want to do validation first; alternatively we need to move its definition up so that it gets evaluated first.

andschwa:
> Additionally consider renaming functions; 'oe_gen_args_header' is no longer a verb.

I am planning to do a rename of some of these functions in the next pass. We're inconsistent about the `oe_` prefix too.

tools/oeedger8r/Emitter.ml
andschwa: Good catch.

andschwa:
@anakrish Fixed the validate_oe_support logic.

andschwa:
Windows is having intermediate issues.

bors r+

andschwa:
Part of #1839.

PR #1877, opened 2019-06-06
Remove no strict aliasing

aserdean:
Fixes: https://github.com/microsoft/openenclave/issues/1717

GCC 5 optimized build requires:
https://github.com/aserdean/openenclave/commit/b33787c4e4b121965e259bbb8d82e5be97c7c4e3

### DIFF
--- a/cmake/compiler_settings.cmake
+++ b/cmake/compiler_settings.cmake
@@ -73,8 +73,6 @@ if (CMAKE_CXX_COMPILER_ID MATCHES GNU OR CMAKE_CXX_COMPILER_ID MATCHES Clang)
   # to fix warnings as they arise, so they don't accumulate "to be fixed later".
   add_compile_options(-Wall -Werror -Wpointer-arith -Wconversion -Wextra -Wno-missing-field-initializers)
 
-  add_compile_options(-fno-strict-aliasing)
-
   # Enables XSAVE intrinsics
   if (OE_SGX)
       add_compile_options(-mxsave)
--- a/host/sgx/sgxload.c
+++ b/host/sgx/sgxload.c
@@ -331,14 +331,13 @@ static oe_result_t _get_sig_struct(
     sgx_sigstruct_t* sigstruct)
 {
     oe_result_t result = OE_UNEXPECTED;
+    sgx_sigstruct_t* p = (sgx_sigstruct_t*)properties->sigstruct;
 
     memset(sigstruct, 0, sizeof(sgx_sigstruct_t));
 
     /* If sigstruct doesn't have expected header, treat enclave as unsigned */
-    if (memcmp(
-            ((sgx_sigstruct_t*)properties->sigstruct)->header,
-            SGX_SIGSTRUCT_HEADER,
-            sizeof(SGX_SIGSTRUCT_HEADER)) != 0)
+    if (memcmp(p->header, SGX_SIGSTRUCT_HEADER, sizeof(SGX_SIGSTRUCT_HEADER)) !=
+        0)
     {
         /* Only debug-sign unsigned enclaves in debug mode, fail otherwise */
         if (!(properties->config.attributes & SGX_FLAGS_DEBUG))


CodeMonkeyLeet:
@anakrish Can you take a look at this? Jork added the global `no-strict-aliasing` flag a long time back and I don't recall a specific rationale for it.

On a related note, should we be dereferencing type-punned pointers to begin with? The option was added along with other changes to tighten compiler warnings, so I infer that it was originally added to prevent unsafe type aliasing behavior.

anakrish:
@CodeMonkeyLeet I am reading up on this.

gupta-ak:
Is removing fno-strict-aliasing safe to do? Aren't stuff like casting `oe_ec_private_key_t` to `oe_private_key_t` violating strict aliasing like here:
```C
typedef struct _oe_private_key
{
    uint64_t magic;
    mbedtls_pk_context pk;
} oe_private_key_t;

/* Opaque representation of a private EC key */
typedef struct _oe_ec_private_key
{
    /* Internal implementation */
    uint64_t impl[4];
} oe_ec_private_key_t;

oe_result_t oe_ec_public_key_from_coordinates(
    oe_ec_public_key_t* public_key,
    oe_ec_type_t ec_type,
    const uint8_t* x_data,
    size_t x_size,
    const uint8_t* y_data,
    size_t y_size)
{
    oe_public_key_t* impl = (oe_public_key_t*)public_key;

    // skip code to access with aliased pointer.
    rc = mbedtls_pk_setup(&impl->pk, info);
    if (rc != 0)
        OE_RAISE_MSG(OE_CRYPTO_ERROR, "rc = 0x%x", rc);
```

anakrish:
@gupta-ak Can you provide an example of how the code you describe can behave wrongly if the flag we to be removed?
If we have a concrete example and document it, it will make it easier to refer to in future if this topic were to come up again.

anakrish:
The following is the smallest snippet of code that I could come up with where strict-aliasing has an effect in semantics:

https://godbolt.org/z/YT6v2E
```c
int foo(int* a, long* b)
{
    *a = 1;
    // With strict aliasing a and b can never
    // point to the same object. The compiler emits
    // code to load b first, then write to a and then
    // add 1 (value of a) to pre-loaded value of b.
    // Since a and b don't alias each other, the compiler
    // could also have written to a first and then loaded b.
    return *a + *b;
}
```

When compiled with strict-aliasing (default):
```asm
foo:
        movq    (%rsi), %rax   ;; load b first
        movl    $1, (%rdi)     ;; write to a; assumes this has no effect on b
        addl    $1, %eax
        ret
```

However when the code is compiled without strict aliasing (-fno-strict-aliasing):
https://godbolt.org/z/MjD60n
```c
int foo(int* a, long* b)
{
    *a = 1;
    // Without strict aliasing a and b could
    // point to the same object. Therefore the compiler
    // emits the write to a first before loading b.
    return *a + *b;
}
```
the compiler assumes that a and b could alias each other and thus write to a could modify b.
Therefore the compiler emits the write to a first just in case it points to b.
```asm
foo:
        movl    $1, (%rdi)        ;; write to a first
        movq    (%rsi), %rax   ;; load b (b could be 1 if it pointed to same location as a)
        addl    $1, %eax         ;; compiler knows a is always 1 at this point
        ret
```

anakrish:
@gupta-ak @CodeMonkeyLeet I have created a small snippet where strict/no-strict aliasing has semantic effect.

I can see that with -fno-strict-aliasing, the compiler has to assume that any two memory locations can pretty much alias each other (unless it can prove that they cannot: e.g a local variable and an input pointer can never alias each other). Therefore the compiler could end up generating less efficient code.

However, it would be good to come up with an actual example in our code base where strict aliasing could cause issues. Yes, we do reinterpretcast oe_ec_public_key_t to oe_public_key_t; but we don't seem to be accessing contents via both the pointers; that could cause aliasing issues.

anakrish:
Capturing links:
https://blog.regehr.org/archives/1307

radhikaj:
Closing this PR since there is no activity on it currently. Please reopen when it is ready to merge.

PR #1865, opened 2019-06-05
Support for DCAP Libraries for Windows added to Openenclave

anitagov:
DCAP dependencies and libraries are added to the Open Enclave flows for Windows (addresses some bullets in #1043 ).
Default build for Windows is still OE_LIBSGX=OFF i.e. the AESM/non-FLC flows.
Installing and/or building Intel SGX DCAP and Azure DCAP client libraries/packages for Windows is documented in the Getting Started on Windows Guide.

PR #1878, opened 2019-06-06
Make openssl available to the command line with ansible windows.yml

george-caluta:
openssl is needed for certificate generation by a few tests (big-malloc, debug-mode, props, etc) and the build will fail on Windows ACC VMs systems configured by ansible/oe-engine. 
For Windows, we are using the openssl from Git and manually add this to the system PATH. 
This PR adds C:\Program Files\Git\mingw64\bin to the system PATH in ansible windows.yml

### DIFF
--- a/scripts/ansible/roles/windows/openenclave/vars/windows.yml
+++ b/scripts/ansible/roles/windows/openenclave/vars/windows.yml
@@ -33,6 +33,7 @@ packages_env_path:
   - 'C:\Program Files\OCaml\bin'
   - 'C:\Program Files\LLVM\bin'
   - 'C:\Program Files\Git\bin'
+  - 'C:\Program Files\Git\mingw64\bin'
   - 'C:\Program Files\shellcheck'
   - 'C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Auxiliary\Build'
 


anakrish:
Why does openssl need to be in the path?

Can you also update https://github.com/microsoft/openenclave/blob/master/docs/GettingStartedDocs/GettingStarted.Windows.md?

johnkord:
@anakrish this PR appears to be based on an email request that @anitagov sent to @ionutbalutoiu yesterday. I imagine the request was made because openssl is used at the command line to generate keys/CSRs/certs and other PKI assets for crypto tests, and Anita is currently enabling those tests

ionutbalutoiu:
> @anakrish this PR appears to be based on an email request that @anitagov sent to @ionutbalutoiu yesterday. I imagine the request was made because openssl is used at the command line to generate keys/CSRs/certs and other PKI assets for crypto tests, and Anita is currently enabling those tests

This is correct. Thanks John for the reply!

anakrish:
When creating PRs it would be good to specify the rationale in the PR description as well as in the comments for the changelist. That would help preserve a nice history that one can navigate at a later date.

In any case, https://github.com/microsoft/openenclave/blob/master/docs/GettingStartedDocs/GettingStarted.Windows.md needs to be updated.

anitagov:
@anakrish - This is the change being made to ansible yml file for windows as we need to do this currently in order for build on Windows to work.

My PR #1865 modified GettingStartedWindows guide to reflect this change. However, after reflecting on some comments from @CodeMonkeyLeet , I believe the best approach will be to establish the path to openssl and set OPENSSL at the start like we do for BASH. I will do this with my PR if that sounds ok.

In that case, this PR may not be needed.

ionutbalutoiu:
> @anakrish - This is the change being made to ansible yml file for windows as we need to do this currently in order for build on Windows to work.
> 
> My PR #1865 modified GettingStartedWindows guide to reflect this change. However, after reflecting on some comments from @CodeMonkeyLeet , I believe the best approach will be to establish the path to openssl and set OPENSSL at the start like we do for BASH. I will do this with my PR if that sounds ok.
> 
> In that case, this PR may not be needed.

Sounds fair to me. Feel free to close this PR if you go that route.

CodeMonkeyLeet:
The crypto/enclave tests have already been enabled in master and they haven't failed on the cross-compile configuration. Those tests have put the `openssl` operations in a bash script and as executed in the bash environment, already have access to the git mingw64 instance of openssl.

There are two classes of things that will fail in the local windows dev environment if openssl is not globally pathed, and these are existing issues:

1. CMake scripts that use openssl in a `COMMAND`
    - This includes add_enclave.cmake, which is used in the following tests for enclave signing keys:
        - bigmalloc
        - debug-mode
        - props
    - These tests do currently pass in CI/CD so ansible has configured those systems in some way that this is not an issue. @ionutbalutoiu/@johnkord would you know of anything in this space?

2. Samples that call openssl in the body of the makefile or CMakeLists.txt
    - Samples test does not currently run on Windows and so this is currently avoided, but will be an issue later.

@anitagov Explicitly pathing openssl in our CMake config is a good idea and will solve (1) but we will still have to come up with a different solution for (2) in the long run though.

anitagov:
> There are two classes of things that will fail in the local windows dev environment if openssl is not globally pathed, and these are existing issues:
> 
> 1. CMake scripts that use openssl in a `COMMAND`
>    
>    * This includes add_enclave.cmake, which is used in the following tests for enclave signing keys:
>      
>      * bigmalloc
>      * debug-mode
>      * props
>    * These tests do currently pass in CI/CD so ansible has configured those systems in some way that this is not an issue. @ionutbalutoiu /@johnkord would you know of anything in this space?

ansible does not configure the path to openssl and hence I filed this issue.
The path was most likely added manually on our CI/CD systems and hence these tests PASS.

> 2. Samples that call openssl in the body of the makefile or CMakeLists.txt
>    
>    * Samples test does not currently run on Windows and so this is currently avoided, but will be an issue later.
> 
> @anitagov Explicitly pathing openssl in our CMake config is a good idea and will solve (1) but we will still have to come up with a different solution for (2) in the long run though.

My original suggestion of adding mingw64\bin\ to the PATH will resolve both 1) and 2). This is what we are all doing manually and works. The Windows Starting Guide will need to be updated with this info and should be done with this PR

Will update the description for this PR too.

ionutbalutoiu:
> > There are two classes of things that will fail in the local windows dev environment if openssl is not globally pathed, and these are existing issues:
> > 
> > 1. CMake scripts that use openssl in a `COMMAND`
> >    
> >    * This includes add_enclave.cmake, which is used in the following tests for enclave signing keys:
> >      
> >      * bigmalloc
> >      * debug-mode
> >      * props
> >    * These tests do currently pass in CI/CD so ansible has configured those systems in some way that this is not an issue. @ionutbalutoiu /@johnkord would you know of anything in this space?
> 
> ansible does not configure the path to openssl and hence I filed this issue.
> The path was most likely added manually on our CI/CD systems and hence these tests PASS.
> 
> > 1. Samples that call openssl in the body of the makefile or CMakeLists.txt
> >    
> >    * Samples test does not currently run on Windows and so this is currently avoided, but will be an issue later.
> > 
> > @anitagov Explicitly pathing openssl in our CMake config is a good idea and will solve (1) but we will still have to come up with a different solution for (2) in the long run though.
> 
> My original suggestion of adding mingw64\bin\ to the PATH will resolve both 1) and 2). This is what we are all doing manually and works. The Windows Starting Guide will need to be updated with this info and should be done with this PR
> 
> Will update the description for this PR too.

That's correct. I manually updated the PATH on our 2x Windows ACC Jenkins systems to have Bors run Anita's PR codebase. If we dropped this PR, I would've reverted the change. So far, I see that this works fine.

@anitagov gave her approval. @CodeMonkeyLeet, if you agree as well, feel free to let Bors know!

PR #1889, opened 2019-06-06
Enabled MBEDTLS_NET_C component

soccerGB:
As part of the work for supporting attested TLS feature (https://github.com/microsoft/openenclave/issues/1325) , we need to enable this mbedtls


gupta-ak:
does this depend on Mike's PR to build?

PR #1892, opened 2019-06-06
Create, Terminate, ECALLs and OCALLs

HernanGatta:
*For review only*

This PR contains:

1. Implementation of `oe_create_enclave`;
2. Implementation of `oe_terminate_enclave`;
3. Implementation of `oe_is_[within|outside]_enclave`;
4. Implementation of `oe_abort`;
5. ECALL support;
6. OCALL support;
7. Sharing of additional code between SGX and OP-TEE support.

I've tried to keep each individual commit reasonable, so it is possible to follow each step in the changelist (see [commits](https://github.com/microsoft/openenclave/pull/1892/commits)).

There are dependencies on changes to:

1. OP-TEE OS:
1.1. The Remote Procedure Call Pseudo-TA (RPC PTA);
1.2. The RPC PTA implements OCALL support in OP-TEE OS and exposes an interface to user-mode TAs;
1.3. [Diff](https://github.com/ms-iot/optee_os/compare/dd4c992df826610e3bfdd756cc7a1685e7b0495b...5461f173a5a4846396ccca3e866e1bea90c4e8f9#diff-ee061305f8535978e87ae5ebed0a7c8b).
2. OP-TEE Client:
2.1. The host-side RPC interface to the RPC PTA;
2.2. It receives RPC requests from the RPC PTA and lets TA hosts handle them;
2.3. [Diff](https://github.com/ms-iot/optee_client/compare/ms-optee-ocalls...user/hegatta/openenclave-merger).

We have work pending to merge these changes into the release branches of both projects.

**Note:** This PR is against `user/hegatta/arm-trustzone-stage1`.


enclave/core/calls.c
```diff
@@ -0,0 +1,44 @@
+
```
dthaler: Delete blank line at top

host/calls.c
```diff
@@ -0,0 +1,134 @@
+
```
dthaler: Delete blank line before copyright

host/ocalls.c
```diff
@@ -1,9 +1,35 @@
...
+}
+
+void HandleRealloc(uint64_t arg_in, uint64_t* arg_out)
+{
+    oe_realloc_args_t* args = (oe_realloc_args_t*)arg_in;
+
+    if (args)
+    {
+        if (arg_out)
+            *arg_out = (uint64_t)realloc(args->ptr, args->size);
```
dthaler: Are you sure it's correct to skip the realloc if arg_out is NULL?  This looks suspicious

host/optee/linux/enclave.c
```diff
@@ -0,0 +1,472 @@
...
+#include <openenclave/edger8r/host.h>
+#include <openenclave/internal/calls.h>
+#include <openenclave/internal/defs.h>
+#include <openenclave/internal/raise.h>
+
+#include "../../calls.h"
+#include "../../ocalls.h"
+#include "enclave.h"
+
+#define PARAM_TYPES_IN_OUT      \
```
dthaler: Is this really all Linux-specific?

include/openenclave/bits/types.h
```diff
@@ -121,6 +121,11 @@ typedef enum _oe_enclave_type
      * other than x64 will not support this and will generate errors.
      */
     OE_ENCLAVE_TYPE_SGX = 2,
+    /** OE_ENCLAVE_TYPE_OPTEE will force the platform to use OP-TEE, but any
+     * platform other than AArch64 with ARM TrustZone will not support this
```
dthaler: Why?   If we will need separate types in this enum then or should be something like OE_ENCLAVE_TYPE_OPTEE_AARCH64, no?   Or do we expect both 32 and 64 to work with the same enum value in the future?

include/openenclave/internal/calls.h
```diff
@@ -460,9 +460,15 @@ oe_result_t oe_ecall(
  * not whether it was successful. The ECALL implementation must define its own
  * error reporting scheme based on its parameters.
  *
+ * The argument description parameters (*_size and *_is_pointer) are only used
```
dthaler: remove "only used for OP-TEE" since there may be other TEEs in the future that need them too and it's easy to miss this comment when adding another TEE.

host/ocalls.c
HernanGatta: Heh, I copy-pasted this from SGX-specific sources.

This is rather curious. It seems that the interface should force `arg_out` to never be `NULL`. Otherwise, If `realloc` returns a pointer to a place in memory other than that originally pointed to by `args->ptr`, the caller has no way of knowing that the memory block has moved.

host/optee/linux/enclave.c
HernanGatta: No, but there's a "but".

The GP TEE Client API specification states that the API may be implemented either in non-secure user-mode or in non-secure kernel mode.

In Linux, the API is implemented by `libteec` in user-mode. On Windows, the API is implemented in kernel-mode in the OP-TEE miniport driver and a custom interface is exposed to user-mode (`CallOpteeCommand`).

So to answer your question, the logic in question here belongs in `CallOpteeCommand` on Windows. This, in turn, makes me realize that the latter must also change, in addition to the driver.

We should consider exposing the TEE Client API in user-mode on Windows, too.

include/openenclave/bits/types.h
HernanGatta: We expect the same enum value to work on 32-bit ARM in the future, too.

include/openenclave/internal/calls.h
HernanGatta: Perhaps this should read "GlobalPlatform-compliant TEEs". That's really what we are adding support for in the SDK; the fact that the only GP-compliant TEEs work on ARM is orthogonal.

host/ocalls.c
HernanGatta: On second thought, there is no way to enforce the requirement that `arg_out` not be `NULL` within this function since `arg_out` is the only way to signal to the caller that an error occurred. So calling this with `arg_out = NULL` is a programming error. I'll look at how it is used.

dthaler: Right, you should delete line 24 and just dereference *arg_out so it will crash on a programming error.

include/openenclave/bits/types.h
dthaler: Great, so delete the "but any platform..." phrase from the comment.

host/ocalls.c
HernanGatta: I like it :) .

host/optee/linux/enclave.c
HernanGatta: As a follow-up, if there is anything we can share with Windows, I'll refactor this file at that time.

host/ocalls.c
HernanGatta: Done.

include/openenclave/bits/types.h
HernanGatta: Rephrased so that ARM32 is included in the meaning of `OE_ENCLAVE_TYPE_OPTEE`.

include/openenclave/internal/calls.h
HernanGatta: Rephrased.

PR #1899, opened 2019-06-07
Rename liboeposix to liboesyscall.

mikbras:
This PR is a follow-up to review comments on the POSIX work. By popular demand, I am renaming the **POSIX** feature to the **SYSCALL** feature. Several reviewers (Anand, Akash, Dave) noted that this work contains non-POSIX features and that a more comprehensive name would be better (Dave suggested the SYSCALL name for extensions but I think it's a good overall name). For the public API, this merely changes the name of the **liboeposix** library to **liboesyscall**. For the internal API, this PR renames 'posix' to 'syscall' in many places. Additionally, Anand suggested separating the new **corelibc** headers  into a different directory. The PR does that too, restoring **corelibc** to its original self.

**To summarize**: this PR changes two kinds of things:
(1) 'posix' becomes 'syscall' in many places (file names, directory names, function names).
(2) syscall-related headers are moved from the **corelibc** directory to **internal/syscall**.

Note that the only consumer of **liboesyscall** is MUSL (**liboelibc**).

The changes were mechanical naming changes and there were no functional changes whatsoever. Unfortunately, Github does not always correctly detect that headers have been renamed. Instead it shows them as deleted in one directory and added to another. So I apologize that this inflates the size of the PR.


include/openenclave/corelibc/endian.h
```diff
@@ -5,8 +5,8 @@
  * into oecore. It scopes down the determination of __BYTE_ORDER
  * relevant to memcpy avoiding GNUC/BSD specializations.
  */
-#ifndef _OE_ENDIAN_H
-#define _OE_ENDIAN_H
+#ifndef _OE_SYSCALL_ENDIAN_H
```
CodeMonkeyLeet: This header remains in corelibc and should not require an updated header guard name.

include/openenclave/internal/syscall/dirent.h
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
CodeMonkeyLeet: For files that were moved to the new header location, did you do it via `git mv`? If possible, it would be nice to preserve the lineage of the file history instead of having deletes and adds.

CodeMonkeyLeet:
> // Copyright (c) Microsoft Corporation. All rights reserved.

It looks like sys/signal.h was only included as a build shim before, but if the intent is that only liboelibc links against liboesyscall, then can this (and the references from host/linux/posix.c & syscall.c) be removed entirely?

---
Refers to: include/openenclave/internal/syscall/sys/signal.h:1 in 433c94f. [](commit_id = 433c94fd78b54deb47add1757be0b8109558d635, deletion_comment = False)

syscall/README.md
```diff
@@ -1,22 +1,22 @@
-liboeposix:
-===========
+liboesyscall
```
CodeMonkeyLeet: Nit: we're removing the colons in some cases but not others, just pick one for consistency.

include/openenclave/internal/syscall/fd.h
```diff
@@ -32,9 +31,9 @@ typedef struct _oe_fd_ops
 
     ssize_t (*write)(oe_fd_t* desc, const void* buf, size_t count);
 
-    ssize_t (*readv)(oe_fd_t* desc, const struct oe_iovec *iov, int iovcnt);
+    ssize_t (*readv)(oe_fd_t* desc, const struct oe_iovec* iov, int iovcnt);
```
CodeMonkeyLeet: >oe_iovec* [](start = 49, length = 9)

Thanks for standardizing these :)

CodeMonkeyLeet:
> add_subdirectory(cpio)

Consider sorting alphabetically for readability.

---
Refers to: tests/syscall/CMakeLists.txt:5 in 433c94f. [](commit_id = 433c94fd78b54deb47add1757be0b8109558d635, deletion_comment = False)

include/openenclave/corelibc/endian.h
mikbras: Thanks. Resolved.

include/openenclave/internal/syscall/dirent.h
mikbras: I'm sorry but this is a git glitch.

mikbras:
> // Copyright (c) Microsoft Corporation. All rights reserved.
> 
> It looks like sys/signal.h was only included as a build shim before, but if the intent is that only liboelibc links against liboesyscall, then can this (and the references from host/linux/posix.c & syscall.c) be removed entirely?
> 
> Refers to: include/openenclave/internal/syscall/sys/signal.h:1 in 433c94f. [](commit_id = 433c94f, deletion_comment = False)

I removed all vestiges of signal (as well as epoll).

mikbras:
> add_subdirectory(cpio)
> 
> Consider sorting alphabetically for readability.
> 
> Refers to: tests/syscall/CMakeLists.txt:5 in 433c94f. [](commit_id = 433c94f, deletion_comment = False)

Yeah, fixed (and I added a comment that cpio must come first).

include/openenclave/internal/syscall/fd.h
mikbras: My pleasure. :)

syscall/README.md
mikbras: Yeah, restored.

mikbras:
Thank you Simon for your review. I addressed all comments as requested. In addition to removing **signal**, I removed all **epoll** references as well.

CodeMonkeyLeet:
@mikbras FYI regarding:
> Yeah, fixed (and I added a comment that cpio must come first).

Akash had a similar thing about this in a different PR a while back and it turns out that the ordering in the CMakeLists.txt should not matter since the cmake graph will order it correctly. (see the changes to build enclave/core under the enclave/CMakeLists.txt again)

mikbras:
> Did you mean load_file_without_crlf? CRLs are often interpreted as certificate revocation lists in this context. (Also, it's stripping CR but not FS here, so load_file_without_cr?)

I checked this fix into the wrong branch. Backed out.

mikbras:
> FYI, the "b" doesn't appear to be propagated (see comment in tests/crypto/enclave/host/host.c) in the syscall hook mechanism.

Backed out this fix to crypto.

PR #1900, opened 2019-06-07
Disable Intel's edger8r from looking for GCC

andschwa:
This resurrects a patch we'd previously made and was lost when we pulled
upstream edger8r code from Intel. This patch should silence the warning
on Windows that the preprocessor was not found.


andschwa:
This change was resurrected from #1781.

andschwa:
Sorry @anitagov, I had no idea about this thing and should have looked further.

PR #1901, opened 2019-06-07
Handle null pointers in deep-copy logic

andschwa:
This adds some (admittedly barebones) tests for null pointers, and then handles them.

### DIFF
--- a/tools/oeedger8r/Emitter.ml
+++ b/tools/oeedger8r/Emitter.ml
@@ -106,8 +106,14 @@ let is_wstr_ptr = function PTVal _ -> false | PTPtr (_, a) -> a.pa_iswstr
 
 let is_str_or_wstr_ptr (p, _) = is_str_ptr p || is_wstr_ptr p
 
-let gen_c_for i =
-  if i = "1" then [] else [sprintf "for (size_t _i = 0; _i < %s; ++_i)" i]
+let gen_c_for count body =
+  if count = "1" then body
+  else
+    [ [sprintf "for (size_t _i = 0; _i < %s; _i++)" count]
+    ; ["{"]
+    ; List.map (( ^ ) "    ") body
+    ; ["}"] ]
+    |> List.flatten
 
 let gen_c_deref i = if i = "1" then "->" else "[_i]."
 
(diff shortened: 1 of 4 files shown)


tools/oeedger8r/Emitter.ml
```diff
@@ -764,26 +778,33 @@ let gen_enclave_code (ec : enclave_content) (ep : edger8r_params) =
...
+                (arg ^ gen_c_deref param_count)
+                param_count setter)
+             (get_deepcopy_members (get_param_atype ptype))) ]
+      |> List.flatten )
   in
   let oe_gen_in_and_inout_setters (plist : pdecl list) =
     let params =
       flatten_map
-        (oe_gen_set_pointers "" "1" (fun p ->
+        (oe_gen_set_pointers "" "" "1" (fun p ->
```
andschwa: @anakrish I have an internal TODO to refactor arg + arg^ parameters here, but it'll come later.

andschwa:
Part of #1839.

include/openenclave/edger8r/common.h
```diff
@@ -81,7 +81,7 @@ OE_INLINE oe_result_t oe_add_size(size_t* total, size_t size)
  * buffer. Make sure that the buffer has enough space.
  */
 #define OE_SET_IN_POINTER(argname, argsize, argtype)                       \
-    if (pargs_in->argname)                                                 \
+    if (pargs_in->argname) /* TODO: This check is now extraneous. */       \
```
anakrish: This header, though internal, is installed and is user visible when debugging.
 I am not comfortable with introducing a new TODO in the last week of v0.6 development.

andschwa: @anakrish This is literally just noting that we do `if(pargs_in->argname)` in the generated code now. There's no harm in doing it here, but it's superfluous. If you want I can just remove the TODOs until if/when we decide to remove the extra checks.

tools/oeedger8r/Emitter.ml
```diff
@@ -764,26 +778,33 @@ let gen_enclave_code (ec : enclave_content) (ep : edger8r_params) =
...
-         oe_get_param_count (ptype, decl, "pargs_in->" ^ prefix)
-       in
-       flatten_map
-         (oe_gen_set_pointers
-            (arg ^ gen_c_deref param_count)
-            param_count setter)
-         (get_deepcopy_members (get_param_atype ptype))) ]
-    |> List.flatten
+    let check_expr =
+      (* TODO: Also gross. *)
```
andschwa: @anakrish This is just a code-style comment of something I want to clean up eventually.

andschwa:
@anakrish this PR is opened separately in order to keep PRs to particular changes, I have two more pending sets of commits to open on top of this. We'll do a final clean up before branching for release.

andschwa:
@anakrish I removed the TODOs and replaced with NOTEs in the edger8r for potential cleanup.

tools/oeedger8r/Emitter.ml
```diff
@@ -637,21 +643,25 @@ let gen_enclave_code (ec : enclave_content) (ep : edger8r_params) =
...
+          String.concat " && " (List.filter (( <> ) "") [parent; arg])
+        in
+        gen_c_for count
+          ( [ [sprintf "if (%s)" check_expr]
+            ; [sprintf "    OE_ADD_SIZE(%s, %s);" buffer size]
+            ; (let param_count =
+                 oe_get_param_count (ptype, decl, "_args." ^ prefix)
+               in
+               flatten_map
+                 (gen_add_size arg (arg ^ gen_c_deref param_count) param_count)
```
anakrish: Overall looks good. Can you double-check if we need to pass check_expr to the children?
Intuitively i would expect that check_expr gets passed down each level and each level adds another condition to the check_expr

andschwa: Yup we did, thank you! Fixed.

andschwa:
We keep having CI issues.

bors r-

PR #1903, opened 2019-06-08
More VS work for Linux

dthaler:
Update nuget package
Add OE_SIMULATION=1 for Simulation configs
Fix binary name of enclave in Simulation config
Fixed bug where wrong TargetName was used since it was defined after
being used by RemoteTargetPath
Remove .signed suffix when copying enclave since OE doesn't expect it

### DIFF
--- a/new_platforms/vsextension/ProjectWizard/ImportEnclaveCommand.cs
+++ b/new_platforms/vsextension/ProjectWizard/ImportEnclaveCommand.cs
@@ -345,7 +345,7 @@ namespace OpenEnclaveSDK
                     // Add nuget package to project.
                     // See https://stackoverflow.com/questions/41803738/how-to-programmatically-install-a-nuget-package/41895490#41895490
                     // and more particularly https://docs.microsoft.com/en-us/nuget/visual-studio-extensibility/nuget-api-in-visual-studio
-                    var packageVersions = new Dictionary<string, string>() { { "openenclave", "0.2.0-CI-20190521-020719" } };
+                    var packageVersions = new Dictionary<string, string>() { { "openenclave", "0.2.0-CI-20190607-222616" } };
                     var componentModel = (IComponentModel)(await this.ServiceProvider.GetServiceAsync(typeof(SComponentModel)));
                     var packageInstaller = componentModel.GetService<IVsPackageInstaller2>();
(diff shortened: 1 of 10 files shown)


new_platforms/vsextension/ProjectWizard/ImportEnclaveCommand.cs
```diff
@@ -441,7 +441,12 @@ private async void Execute(object sender, EventArgs e)
                             if (gdbRule != null)
                             {
                                 // Configure GDB debugger settings.
-                                gdbRule.SetPropertyValue("PreLaunchCommand", "export PYTHONPATH=/opt/openenclave/lib/openenclave/debugger/gdb-sgx-plugin;export LD_PRELOAD=/opt/openenclave/lib/openenclave/debugger/liboe_ptrace.so");
+                                string gdbEnvironmentSettings = "export PYTHONPATH=/opt/openenclave/lib/openenclave/debugger/gdb-sgx-plugin;export LD_PRELOAD=/opt/openenclave/lib/openenclave/debugger/liboe_ptrace.so";
```
anakrish: These are oegdb internals. It would be preferable not to rely on the internals here. Is it possible to invoke the oegdb wrapper script  that hides all the internals.

anakrish: /opt/openenclave is the default install location for OE. However, an user could install OE to any location or even have multiple installs if say they are trying to debug an issue that occurs only in a particular release of OE. Can we avoid this assumption about OE install location?

dthaler: I don't believe VS lets you change the debugger executable name, you can only run scripts before invoking gdb, and commands sent to gdb.  So I could not find any way to invoke the oegdb wrapper.

dthaler: it is possible to avoid assumption about OE install location in the future by prompting the user.  This would be more work though so will not be in this changelist.

anakrish: > I don't believe VS lets you change the debugger executable name, you can only run scripts before invoking gdb, and commands sent to gdb. So I could not find any way to invoke the oegdb wrapper.

Created #1905 for future work.

PR #1910, opened 2019-06-10
Handle `count * size`

andschwa:
This fixes the deep-copy code to properly handle when both `count` and `size` attributes are specified, plus loosely associated cleanups.

### DIFF
--- a/tools/oeedger8r/Emitter.ml
+++ b/tools/oeedger8r/Emitter.ml
@@ -336,15 +336,12 @@ let oe_gen_call_user_function (fd : func_decl) =
          fd.plist)
     ^ ");" ]
 
-(** Check if any of the parameters or the return type has the given
-    root type. *)
-let uses_type (root_type : atype) (fd : func_decl) =
-  let param_match =
-    List.exists (fun (pt, decl) -> root_type = get_param_atype pt) fd.plist
-  in
-  if param_match then param_match else root_type = fd.rtype
-
 let warn_non_portable_types (fd : func_decl) =
+  (* Check if any of the parameters or the return type has the given
+     root type. *)
+  let uses_type (t : atype) =
+    t = fd.rtype || List.exists (fun (p, _) -> t = get_param_atype p) fd.plist
+  in
   let print_portability_warning ty =
     printf
       "Warning: Function '%s': %s has different sizes on Windows and Linux. \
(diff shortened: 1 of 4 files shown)


andschwa:
Part of #1839.

tools/oeedger8r/Emitter.ml
```diff
@@ -154,42 +154,41 @@ let attr_value_to_string argstruct = function
...
-      | None ->
-          let dims =
-            List.map (fun i -> "[" ^ string_of_int i ^ "]") d.array_dims
-          in
-          sprintf "sizeof(%s%s)" type_expr (String.concat "" dims) )
+  let get_ptr_or_decl_size (p : ptr_size) =
+    let size = attr_value_to_string argstruct p.ps_size
+    and count = attr_value_to_string argstruct p.ps_count in
+    match (size, count) with
+    | Some s, None -> s
```
anakrish: Nice!

tools/oeedger8r/Emitter.ml
```diff
@@ -154,42 +154,41 @@ let attr_value_to_string argstruct = function
...
-      match attr_value_to_string argstruct p.ps_size with
-      | Some s -> sprintf "(%s / sizeof(%s))" s type_expr
-      | None ->
-          let dims = List.map string_of_int d.array_dims in
-          String.concat " * " dims )
+  let get_ptr_or_decl_count (p : ptr_size) =
+    let size = attr_value_to_string argstruct p.ps_size
+    and count = attr_value_to_string argstruct p.ps_count in
+    match (size, count) with
+    | Some s, None -> sprintf "(%s / sizeof(%s))" s type_expr
```
anakrish: We would need a strategy when size is not a multiple of sizeof(type)

tools/oeedger8r/Emitter.ml
```diff
@@ -154,42 +154,41 @@ let attr_value_to_string argstruct = function
...
+    | None, Some c -> c
+    (* TODO: Check that this is an even multiple of the size of type. *)
+    | Some s, Some c -> sprintf "((%s * %s) / sizeof(%s))" s c type_expr
+    | None, None ->
+        let dims = List.map string_of_int decl.array_dims in
```
anakrish: Isn't this just get_array_dims like you changed above?

tools/oeedger8r/Emitter.ml
```diff
@@ -377,37 +376,34 @@ let warn_signed_size_or_count_types (fd : func_decl) =
...
+      let id = decl.identifier in
+      if List.mem id size_params then
         match ptype with
-        (* TODO: Combine these two patterns. *)
         | PTVal (Long s | LLong s) when s = Signed ->
-            print_signedness_warning name
-        | PTVal (Int i) when get_int_signedness i = Signed ->
-            print_signedness_warning name
+            print_signedness_warning id
+        | PTVal (Int i) when i.ia_signedness = Signed ->
```
anakrish: Not part of this PR, but shouldn't we also warn about int32, int16 etc?

tools/oeedger8r/Emitter.ml
andschwa: Isn't there a TODO for that right here?

tools/oeedger8r/Emitter.ml
andschwa: Whoops, that's what happens when I have multiple PRs in progress and have to reorder them 😅

tools/oeedger8r/Emitter.ml
andschwa: Oh my bad I only put it in the other function 😅

tools/oeedger8r/Emitter.ml
andschwa: Wait nope, this is doing something entirely different than get_array_dims. This is giving us the expression `2 * 4 * 3 * 5` etc., instead of `[2][4][3][5]`.

tools/oeedger8r/Emitter.ml
andschwa: Actually it was here, moved it up.

PR #1911, opened 2019-06-11
Add PS module to manage Hyper-V SGX settings.

HernanGatta:


new_platforms/scripts/VirtualMachineSgxSettings.psd1
```diff
@@ -0,0 +1,12 @@
+@{
+GUID="{5c36f9d9-e64f-4dbb-ae38-6ad1669954c4}"
+Author="Microsoft Corporation"
+CompanyName="Microsoft Corporation"
+Copyright="© Microsoft Corporation. All rights reserved."
+RootModule = "VirtualMachineSgxSettings.psm1"
```
dthaler: nit: other lines have no space around '='

new_platforms/scripts/VirtualMachineSgxSettings.psm1
```diff
@@ -0,0 +1,283 @@
...
+
+### -----------------------------------
+### Strings
+### -----------------------------------
+
+Data Strings
+{
+# culture="en-US"
+ConvertFrom-StringData @'
+    ERR_NO_MEM_SETTING_DATA_OBJECT = Could not retreive the Memory Setting Data object for VM '{0}' from WMI.
```
dthaler: typo: retrieve (I before E except after C)

new_platforms/scripts/VirtualMachineSgxSettings.psm1
```diff
@@ -0,0 +1,283 @@
...
+    ERR_NO_VMM = Could not retreive the instance of the Virtual System Management Service from WMI.
```
dthaler: typo: retrieve (line 170 at least has it right)

new_platforms/scripts/VirtualMachineSgxSettings.psm1
```diff
@@ -0,0 +1,283 @@
...
+    ERR_VM_STATE = "The VM '{0}' must be Off before its SGX settings may be modified."
```
dthaler: Why are double quotes needed on this line but not on line 22, 27, etc?

new_platforms/scripts/VirtualMachineSgxSettings.psm1
```diff
@@ -0,0 +1,283 @@
...
+    ERR_NO_SETTINGS_TO_MODIFY = Specify at least one SGX setting to modify.
+    ERR_NO_SGX_MEM_TO_ENABLE = To enable SGX, some SGX EPC memory is required (-SgxSize).
+    ERR_COULD_NOT_MODIFY_SETTINGS = Could not modify SGX settings for VM '{0}' via WMI with error value '{1}'.
+
+    WARN_GEN2_TITLE = The VM {0}' is not a Generation 2 or later VM.
```
dthaler: shouldn't there be a matching single quote before {0}?

new_platforms/scripts/VirtualMachineSgxSettings.psm1
```diff
@@ -0,0 +1,283 @@
...
+        # VM to configure.
+        [Parameter(Mandatory=$True, ParameterSetName="ByObject", Position=0)]
+        [ValidateNotNull()]
+        [Microsoft.HyperV.PowerShell.VirtualMachine]$Vm,
+
+        # Whether SGX is enabled in the VM.
+        [Parameter(Position=1)]
+        [Bool]$IsSgxEnabled = $False,
+
+        # Desired SGX EPC Memory Size for the VM (in MB).
```
dthaler: Question: is there any "help text" for the command that should indicate that the size is in MB?

new_platforms/scripts/VirtualMachineSgxSettings.psm1
```diff
@@ -0,0 +1,283 @@
...
+    }
+}
+
+### -----------------------------------
+### Helper Functions
+### -----------------------------------
+
+Function Get-VMSgxMemorySettingData([Microsoft.HyperV.PowerShell.VirtualMachine] $Vm)
+{
+    $Mem =  Get-WmiObject -Namespace root/virtualization/v2 -Class Msvm_MemorySettingData |
```
jiria: nit: double space

new_platforms/scripts/VirtualMachineSgxSettings.psm1
```diff
@@ -0,0 +1,283 @@
...
+        [String]$VmName,
+
+        # VM to configure.
+        [Parameter(Mandatory=$True, ParameterSetName="ByObject", Position=0)]
+        [ValidateNotNull()]
+        [Microsoft.HyperV.PowerShell.VirtualMachine]$Vm,
+
+        # Whether SGX is enabled in the VM.
+        [Parameter(Position=1)]
+        [Bool]$IsSgxEnabled = $False,
```
jiria: Would it make sense to make this into a switch?

new_platforms/scripts/VirtualMachineSgxSettings.psm1
```diff
@@ -0,0 +1,283 @@
...
+        [Parameter(Position=3)]
+        [String]$SgxLaunchControlDefault = [String]::Empty,
+
+        # Desired SGX Launch Control Mode for the VM.
+        [Parameter(Position=4)]
+        [UInt32]$SgxLaunchControlMode = 0,
+
+        # Ignore warnings.
+        [Parameter()]
+        [Switch]$Force = $False
```
jiria: Switch is $false by default.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
```diff
@@ -0,0 +1,283 @@
...
+        #      Note: Only Gen2 VMs support Secure Boot.
+        # 2. Checkpoints are not supported with SGX turned on.
+
+        If (!$Force -and $IsSgxEnabled)
+        {
+            # Check VM generation.
+            If ($Vm.Generation -lt 2)
+            {
+                Write-Warning ($Strings.WARN_GEN2_TITLE -f $Vm.Name)
+                Write-Warning $Strings.WARN_GEN2_TEXT
```
jiria: Given the above comment about SGX support, should we not abort/throw here?

new_platforms/scripts/VirtualMachineSgxSettings.psd1
HernanGatta: Ran file formatter on both files.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: Fixed.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: Fixed.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: They are not; removed.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: Added.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: Ran formatter on both files.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: I prefer to think of switches as the elements passed in through the command line and storage variables as boolean. Thoughts?

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: I added the explicit default for parity with the parameters above to express intent, but OK; removed.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: That's one of the issues I wanted your input on.

The WMI method invoked on line 247 does its own validation of the desired settings. However, it does not provide the user with any meaningful information in case of failure, other than a numeric error code. As a result, I decided to add some checking of my own. However, said checking duplicates the checking done by the WMI method, and they might, further down the road, diverge.

Hence, I decided to make the checks warn or inform rather than terminate, except when the error is obvious (i.e. enable SGX with no SGX memory, or when the VM is running).

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: Added help, accessible via:

```
Get-Help Get-VMSgx
Get-Help Set-VMSgx
```

new_platforms/scripts/VirtualMachineSgxSettings.psm1
jiria: Agreed, but I think I am missing something in this case. In this case, -IsSgxEnabled would be passed through command line or however the cmdlet is invoked. I would think about storage variable as a local function variable or a global variable. How are you distinguishing between storage boolean parameters and switches?

In my head, if the default was $true, then I would keep it as Bool. But if the default is $false, then it feels like a switch. Wanna enable it, add the switch.

jiria: Speaking of which, if this is an input parameter, Is... does not sound to me like the right name.

Thinking more, I think I understand better your original intent. You want to have a way to specify both true and false, but not false implicitly by ignoring (even though that is what currently happens). Ok, I guess bool makes sense here, but would still consider renaming it, perhaps SgxEnabled?

new_platforms/scripts/VirtualMachineSgxSettings.psm1
```diff
@@ -0,0 +1,379 @@
...
+            }
+
+            # Check Secure Boot settings.
+            If ($Vm.Generation -ge 2) {
+                $Fw = $Vm | Get-VMFirmware
+
+                If ($Fw.SecureBoot -eq [Microsoft.HyperV.PowerShell.OnOffState]::On) {
+                    Write-Information ($Strings.WARN_SECURE_BOOT_TITLE -f $Vm.Name)
```
jiria: Are you aware of this?

The $InformationPreference preference variable value determines whether the message you provide to Write-Information is displayed at the expected point in a script's operation. Because the default value of this variable is SilentlyContinue, by default, informational messages are not shown. 

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/write-information?view=powershell-6

new_platforms/scripts/VirtualMachineSgxSettings.psm1
jiria: One way is to run the check after the failure of the WMI call. When it fails, you can test for the familiar failure patterns to help user diagnose.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: I've changed it to `SgxEnabled` everywhere.

The function checks against `$PSBoundParameters`, so `SgxEnabled` isn't read and nothing is done if you don't specify it.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: Good idea.

HernanGatta: I've refactored the error checking:

For conditions that won't cause the modification command to fail, I check them before trying to change anything, unless the user passes in `-Force`. Those checks are useful to give the user an idea of what might go wrong when trying to start VMs.

For conditions that are likely to have caused the modification command to have failed, I check them only after the failure is detected.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: Yes, I noticed when the messages would not print.

The value of `$InformationPreference` is set at the top of the module.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
jiria: Given that you are checking against PSBoundParameters, does it make sense to have default values? Imho it does not, as they are not used. Seems confusing to me.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
jiria: I wonder if we are overloading -Force here. In my head, -Force is ignore warnings/possible conflicts and proceed. We are doing that even without -Force. Does it make sense to use -Force just to change warning output, without any real change in the code flow?

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: Removed the defaults.

new_platforms/scripts/VirtualMachineSgxSettings.psm1
HernanGatta: Agreed; removed the usage of `Force` altogether. The user can use PowerShell's built-in warning suppression mechanisms.

HernanGatta:
Merged as part of #2159 instead.

Issue #1919, opened 2019-06-11
oe_mount has an access violation when fed an invalid file system name


radhikaj:
In oe_mount mount_point_t is not initialized. 
If line 142 fails  "device = oe_device_table_find(filesystemtype, OE_DEVICE_TYPE_FILE_SYSTEM); "
then we go to done:
    if (mount_point.path)
        oe_free(mount_point.path);

But since mount_point.path was not initialized to null, it has an access violation.

anitagov:
Good find Radhika

PR #1922, opened 2019-06-11
Uninitialized mount_point causes segmentation fault during error cond…

radhikaj:
…itions

Issue #1919

### DIFF
--- a/posix/mount.c
+++ b/posix/mount.c
@@ -139,7 +139,7 @@ int oe_mount(
     bool locked = false;
     oe_posix_path_t source_path;
     oe_posix_path_t target_path;
-    mount_point_t mount_point;
+    mount_point_t mount_point = {0};
 
     if (!target || !filesystemtype)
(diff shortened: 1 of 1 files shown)


posix/mount.c
```diff
@@ -139,7 +139,7 @@ int oe_mount(
     bool locked = false;
     oe_posix_path_t source_path;
     oe_posix_path_t target_path;
-    mount_point_t mount_point;
+    mount_point_t mount_point = {0};
```
anitagov: This should work. But perhaps, initializing all the elements is more readable:
mount_point_t mount_point = {0, 0, 0, 0};

If we were using C99 (I believe we support C98), a better way is to use designated initialization:
mount_point_t mount_point = {.path=0, .path_size=0, .fs=0, flags=0};

radhikaj: @anitagov ={0} is a commonly accepted way to 0 out a structure.

johnkord:
CI should be fixed now :)

bors r+

PR #1926, opened 2019-06-12
Unblock CI by disabling the crypto test on Windows temporarily.

johnkord:
There have been some changes to the Windows CI systems recently to support DCAP on Windows, and somehow this test has begun failing. This test was passing, but we aren't sure why it's failing now. This PR will temporarily disable this test until we can understand why it is failing, so that other work won't be blocked by CI failures.


anitagov:
We noticed that the AESM service was running but the DCAP libraries were installed on that platform and the registry setting for DCAP was set.

anitagov:
John - You will also need to disable the crypto_crls_cert_chain tests for Windows.

anitagov:
@johnkord - The crypto test has just started failing on Windows and we saw this in a regular AESM Windows platform too. Can you please remove DCAP from the description and add one of the pipelines that failed? Thanks.

johnkord:
@anitagov I don't think removing DCAP from the description makes sense here, because I know that @ionutbalutoiu was making changes to some of the Windows CI systems that involved supporting DCAP. I will add one of the CI failures to the description here, though!

anitagov:
The problem is with new openssl version on Windows which generates PEM certificates with CRLF line endings on Windows. Mike's PR #1932 fixes this.

PR #1932, opened 2019-06-13
Fix regression in tests/crypto

mikbras:
This fixes a regression in **tests/crypto**. The new **openssl** command generates PEM certificates with **CRLF** line endings on Windows. This change modifies **read_pem_key()** so that it discards **CR** characters when reading certificates.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -46,6 +46,8 @@ if (UNIX OR ADD_WINDOWS_ENCLAVE_TESTS OR USE_CLANGW)
             add_subdirectory(thread_local_no_tdata)
         endif()
         add_subdirectory(bigmalloc)
+        add_subdirectory(crypto)
+        add_subdirectory(crypto_crls_cert_chains)
         add_subdirectory(debug-mode)
         add_subdirectory(echo)
         add_subdirectory(enclaveparam)
@@ -85,8 +87,6 @@ if (OE_SGX AND UNIX)
    # ecall_ocall enclave size cannot be handled by Windows ninja CI
    add_subdirectory(ecall_ocall)
    add_subdirectory(libunwind)
-   add_subdirectory(crypto)
-   add_subdirectory(crypto_crls_cert_chains)
 
    # Attestation supported only on Linux
    add_subdirectory(qeidentity)
--- a/tests/crypto/ec_tests.c
+++ b/tests/crypto/ec_tests.c
@@ -914,19 +914,32 @@ void TestEC()
             "../data/root.ec.cert.pem",
             _CHAIN,
             OE_COUNTOF(_CHAIN)) == OE_OK);
-    OE_TEST(read_key("../data/root.ec.key.pem", _PRIVATE_KEY) == OE_OK);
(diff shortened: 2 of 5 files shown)


mikbras:
bors+

mikbras:
bors try

anitagov:
@mikbras - The crypto_crl_cert_chains test fails on Windows when the ELF enclaves are created on Linux and then copied over to Windows.

mikbras:
I'm giving up on this regression. Apparently there are deeper problems than I anticipated.

anitagov:
@CodeMonkeyLeet

anitagov:
@mikbras - I believe there are multiple problems. This PR seems to address the problem with the crypto test failing due to the new version of openssl introducing the CR/LF with PEM certificates. Perhaps, we should address the crypto_crl_cert_chains failure with another PR.

CodeMonkeyLeet:
Based on the proposed cause of failure, it's not clear to me that the presented fix is the right one. Is this a test failure or a product failure?

1. For example, if the failure is due to formatting differences between and provided key and the same key being written out by mbedtls not containing CRLF, then fixing the test code would be appropriate, even if the fix also needed to be propagated to tests/crypto_crl_cert_chains as well.

2. If the failure is that mbedtls does not handle CR characters appropriately and interprets a different key/cert as a result, thereby causing a failure (e.g. failure to verify signature/decrypt) then clearly the OE crypto wrapper needs to take this into account instead of the test. 

In either case, I don't think we want to strip the characters before passing to the crypto implementation to handle; we actually want the crypto implementation to potentially deal with different line endings on Windows vs. Linux because I fully expect that OE needs to support that.

mikbras:
> Based on the proposed cause of failure, it's not clear to me that the presented fix is the right one. Is this a test failure or a product failure?
> 
> 
> For example, if the failure is due to formatting differences between and provided key and the same key being written out by mbedtls not containing CRLF, then fixing the test code would be appropriate, even if the fix also needed to be propagated to tests/crypto_crl_cert_chains as well.
> 
> 
> If the failure is that mbedtls does not handle CR characters appropriately and interprets a different key/cert as a result, thereby causing a failure (e.g. failure to verify signature/decrypt) then clearly the OE crypto wrapper needs to take this into account instead of the test.
> 
> 
> In either case, I don't think we want to strip the characters before passing to the crypto implementation to handle; we actually want the crypto implementation to potentially deal with different line endings on Windows vs. Linux because I fully expect that OE needs to support that.

The problem is this: the test checks that the input certificate (read by mbed ) is identical to the output certificate (written by mbed). The mbed implementation works fine. The test fails because the input and output certificates are identical except for differing line endings. A better fix would be to convert both the input and output certificates to a common format before comparing them. I recommend something like this:
```
strip_carriage_returns(input, &input_size);
strip_carriage_returns(output, &output_size);
/* Check that input_size == output_size */
/* Check that input and output are identical. */
```

anitagov:
@yakman2020 - I think this change should not have got into master. Please see final comments from @CodeMonkeyLeet  and @mikbras .
The test crl_cert_chains breaks on Windows with the newer openssl with these changes.

anitagov:
Filed issue #1955 to get the final suggested fix into the master.

Issue #1527, opened 2019-03-04
Provide script or packages list for porting to platforms where ansible may not work


anakrish:
While ansible works nicely, it does not work on all platforms.
To help compiling/porting OE to such platforms it would be good to:
1) Provide a shell-script that installs necessary packages
2) Document the list of packages needed somewhere so that they can be manually installed.
     Reading the scripts/ansible folder, it is not really obvious what packages I need to install manually 
     to get a system up and running.

@johnkord @CodeMonkeyLeet @ionutbalutoiu @aserdean

ionutbalutoiu:
@anakrish

1. As far as I know, we support Ubuntu 16.04, Ubuntu 18.04 and Windows Server 2016 ACC for the OE project.

    On Ubuntu 16.04 & 18.04 you can run Ansible locally just as any shell script ([this section](https://github.com/Microsoft/openenclave/blob/master/docs/GettingStartedDocs/SGX1FLCGettingStarted.md#install-project-requirements) has instructions for this).

    On Windows Server 2016 topic, unfortunately Ansible supports Windows only as a target remote machine to configured via WinRM. That means, you cannot run Ansible locally on Windows, and you need to target it from a Linux sandbox. I understand that this is not desirable, since the Windows devs will be happier to just run a local script, and get all their OE requirements ready. Do you think that a simple PowerShell script will do here ?

2. Currently, there is documentation in the [docs/GettingStartedDocs](https://github.com/Microsoft/openenclave/tree/master/docs/GettingStartedDocs) directory about this.
    So far, I noticed this:
    * Info to the packages to be installed on Ubuntu platform [here](https://github.com/Microsoft/openenclave/blob/master/docs/GettingStartedDocs/install_oe_sdk.md)
    * Info for all the requirements necessary to be setup on Windows [here](https://github.com/Microsoft/openenclave/blob/master/docs/GettingStartedDocs/GettingStarted.Windows.md)

    Maybe it's worth looking if anything is outdated there, but that's the place for official docs.

ionutbalutoiu:
Quick update on something related to the issue described here, and a discussing started at PR https://github.com/microsoft/openenclave/pull/1865#discussion_r291387496.

We are working on a PowerShell standalone script to configure a clean ACC Windows VM. This script would allow someone to quickly configure a Windows dev environment, and it will be called by Ansible as well.

We'll go with this approach to overcome the fact that Ansible doesn't run natively on Windows.

[...]

PR #1934, opened 2019-06-13
Add PowerShell script to configure a clean Windows SGX enabled machine

george-caluta:
* Add PowerShell script will install all the prerequisites needed to build the Open Enclave project, and updated the README with instructions to run the script. Fixes https://github.com/openenclave/openenclave/issues/1527
* Updated Ansible `oe-windows-acc-setup.yml` playbook to use this script


dthaler:
This change appears to be specific to Windows VMs not Linux VMs too.  Please update the title.

scripts/oe-env-prepare.ps1
```diff
@@ -0,0 +1,446 @@
...
+                    -InstallDirectory $installDir
+
+    $pswInstaller = Join-Path $installDir "Intel SGX PSW for Windows v2.3.100.49777\PSW_EXE_RS2_and_before\Intel(R)_SGX_Windows_x64_PSW_2.3.100.49777.exe"
+    $p = Start-Process -Wait -NoNewWindow -PassThru -FilePath $pswInstaller `
+                       -ArgumentList @("--extract-folder", "$installDir", "--x")
+    if($p.ExitCode -ne 0) {
+        Throw "Failed to extract the Intel SGX PSW bundle: $pswInstaller"
+    }
+    $p = Start-Process -Wait -NoNewWindow -PassThru -FilePath "$installDir\install.exe" `
+                       -ArgumentList @("install", "--eula=accept", "--output=$installDir\intel_install.log", "--components=all")
```
ionutbalutoiu: Please get the unattended parameters from our Ansible scripts:
https://github.com/microsoft/openenclave/blob/a982b46cf440def8fb66e94f2622a4f81e2b350b/scripts/ansible/roles/windows/az-dcap-client/tasks/environment-setup.yml#L64

We want to maintain exactly what's in Ansible.

scripts/oe-env-prepare.ps1
```diff
@@ -0,0 +1,446 @@
...
+        "--add Microsoft.VisualStudio.Workload.VCTools",
+        "--add Microsoft.VisualStudio.Component.VC.Tools.x86.x64",
+        "--add Microsoft.VisualStudio.Component.VC.140",
+        "--add Microsoft.VisualStudio.Component.Windows10SDK.16299.Desktop",
+        "--add Microsoft.VisualStudio.Component.Windows81SDK",
+        "--add Microsoft.VisualStudio.Component.VC.ATL"
+        "--add Microsoft.VisualStudio.Component.Windows10SDK.17134"
+        "--includeRecommended"
+    )
+    # VisualStudio install sometimes is throwing errors on first try.
```
ionutbalutoiu: I believe this was something related with `oe-engine` only.

IMO, I don't think this is necessary here. Please remove it and see if the problem persist.

scripts/oe-env-prepare.ps1
```diff
@@ -0,0 +1,446 @@
...
+    }
+    "ocaml" = @{
+        "url" = "http://www.ocamlpro.com/pub/ocpwin/ocpwin-builds/ocpwin64/20160113/ocpwin64-20160113-4.02.1+ocp1-msvc64.zip"
+        "local_file" = Join-Path $PACKAGES_DIRECTORY "ocpwin64.zip"
+    }
+    "clang7" = @{
+        "url" = "http://releases.llvm.org/7.0.1/LLVM-7.0.1-win64.exe"
+        "local_file" = Join-Path $PACKAGES_DIRECTORY "LLVM-7.0.1-win64.exe"
+    }
+    "sgx_drivers" = @{
```
ionutbalutoiu: I don't see this `sgx_drivers` declared anywhere in our Ansible scripts.

scripts/oe-env-prepare.ps1
```diff
@@ -0,0 +1,446 @@
...
+    $installDir = Join-Path $env:ProgramFiles "OCaml"
+    $tmpDir = Join-Path $PACKAGES_DIRECTORY "ocpwin64"
+    Install-ZipTool -ZipPath $PACKAGES["ocaml"]["local_file"] `
+                    -InstallDirectory $tmpDir `
+                    -EnvironmentPath @("$installDir\bin")
+    $extractDir = Get-ChildItem $tmpDir | Select-Object -Property Name
+    New-Directory -Path $installDir -RemoveExisting
+    Copy-Item  "$tmpDir\$($extractDir.Name)\*" -Destination $installDir -Recurse -Force
+    Push-Location "$installDir\bin"
+    & ocpwin -in
```
ionutbalutoiu: I don't believe this is needed.

Looking at the [documentation](https://www.typerex.org/ocpwin-manual-install.html), step 2, this seems to register OCaml in the Windows registry and update the PATH.

However, the Ansible logic didn't execute this and we already update the PATH via `Install-ZipTool` at the beginning of the function. Please maintain the same Ansible logic that we have:
https://github.com/microsoft/openenclave/blob/a982b46cf440def8fb66e94f2622a4f81e2b350b/scripts/ansible/roles/windows/openenclave/tasks/environment-setup.yml#L45-L55

scripts/oe-env-prepare.ps1
```diff
@@ -0,0 +1,446 @@
...
+    if ($LASTEXITCODE -ne 0) {
+        throw "Failed to install ocaml"
+    }
+    Pop-Location
+}
+function Install-LLVM {
+    Install-Tool -InstallerPath $PACKAGES["clang7"]["local_file"] `
+                 -ArgumentList "/S"
```
ionutbalutoiu: We need the `-EnvironmentPath` parameter to make sure that the LLVM bin directory is present in the PATH.

scripts/oe-env-prepare.ps1
```diff
@@ -0,0 +1,446 @@
...
+    if(Test-Path -Path $shellcheckDest) {
+        Remove-Item -Path $shellcheckDest -Force -Recurse
+    }
+    New-Item -ItemType Directory -Path $shellcheckDest
+    Move-Item "$PACKAGES_DIRECTORY\shellcheck.exe" $shellcheckDest -Force
+}
+
+function Install-DCAPDrivers {
+    Install-Tool -InstallerPath $PACKAGES["dcap"]["local_file"] `
+                 -ArgumentList "/auto $PACKAGES_DIRECTORY\Intel_SGX_DCAP"
```
ionutbalutoiu: Have the `-ArgumentList` as an array of strings.

Instead of `"/auto $PACKAGES_DIRECTORY\Intel_SGX_DCAP"`, use `@("/auto", "$PACKAGES_DIRECTORY\Intel_SGX_DCAP")`.

scripts/oe-env-prepare.ps1
```diff
@@ -0,0 +1,446 @@
...
+        if($inf.Count -gt 1) {
+            $inf
+            Throw "Multiple $driver.inf files found"
+        }
+        & "$PACKAGES_DIRECTORY\devcon.exe" install "$($inf.FullName)" $drivers[$driver]['location']
+        if($LASTEXITCODE) {
+            Throw "Failed to install $driver driver"
+        }
+    }
+    $OE_NUGET_DIR = "C:\openenclave\prereqs\nuget"
```
ionutbalutoiu: This should be a global variable moved at the top of the file.

scripts/oe-env-prepare.ps1
```diff
@@ -0,0 +1,446 @@
...
+    New-Directory -Path $TEMP_NUGET_DIR
+    Copy-Item "$PACKAGES_DIRECTORY\Intel_SGX_DCAP\Intel SGX DCAP for Windows *\nupkg\*\*.nupkg" -Destination $TEMP_NUGET_DIR -Force
+    Copy-Item $PACKAGES['azure_dcap_client_nupkg']['local_file'] -Destination $TEMP_NUGET_DIR -Force
+    & "$PACKAGES_DIRECTORY\nuget.exe" install 'EnclaveCommonAPI' -Source "$TEMP_NUGET_DIR" -OutputDirectory "$OE_NUGET_DIR" -ExcludeVersion
+    & "$PACKAGES_DIRECTORY\nuget.exe" install 'DCAP_Components' -Source "$TEMP_NUGET_DIR" -OutputDirectory "$OE_NUGET_DIR" -ExcludeVersion
+    & "$PACKAGES_DIRECTORY\nuget.exe" install 'Microsoft.Azure.DCAP.Client' -Source "$TEMP_NUGET_DIR;nuget.org" -OutputDirectory "$OE_NUGET_DIR" -ExcludeVersion
+
+    New-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\sgx_lc_msr\Parameters" -Name "SGX_Launch_Config_Optin" -Value "1" -PropertyType DWORD -Force
+
+    Add-ToSystemPath -Path 'C:\Program Files\shellcheck'
```
ionutbalutoiu: All the below `Add-ToSytemPath` shouldn't be in the `Install-DCAPDrivers` function as they are not related to `DCAPDrivers`.

Please move them to their own functions.

scripts/oe-env-prepare.ps1
```diff
@@ -0,0 +1,446 @@
...
+    Install-SGX
+    Install-PSW
+
+    Start-ExecuteWithRetry -ScriptBlock {
+        Start-Service "AESMService" -ErrorAction Stop
+    } -RetryMessage "Failed to start AESMService. Retrying"
+
+    Copy-Item -Path $PACKAGES["nuget"]["local_file"] -Destination "${AZUREDATA_BIN_DIRECTORY}\nuget.exe"
+
+    Install-VisualStudio
```
ionutbalutoiu: I believe Visual Studio is installed before installing the PSW software.

Please install and configure the prerequisites in the same order as Ansible is doing right now.

scripts/oe-env-prepare.ps1
```diff
@@ -0,0 +1,446 @@
...
+    Install-Ocaml
+    Install-LLVM
+    Install-Shellcheck
+    Install-DCAPDrivers
+    Write-Output 'Please reboot your computer for the configuration to complete.'
```
ionutbalutoiu: We should reboot the node at the end of the script (instead of `exit 0`).

And we should have the docs say that the script will reboot the node.

scripts/oe-env-prepare.ps1
```diff
@@ -0,0 +1,446 @@
...
+    "clang7" = @{
+        "url" = "http://releases.llvm.org/7.0.1/LLVM-7.0.1-win64.exe"
+        "local_file" = Join-Path $PACKAGES_DIRECTORY "LLVM-7.0.1-win64.exe"
+    }
+    "sgx_drivers" = @{
+        "url" = "http://download.windowsupdate.com/d/msdownload/update/driver/drvs/2018/01/af564f2c-2bc5-43be-a863-437a5a0008cb_61e7ba0c2e17c87caf4d5d3cdf1f35f6be462b38.cab"
+        "local_file" = Join-Path $PACKAGES_DIRECTORY "sgx_base.cab"
+    }
+    "psw" = @{
+        "url" = "https://oejenkins.blob.core.windows.net/oejenkins/Intel_SGX_PSW_for_Windows_v2.3.100.49777.exe"
```
CodeMonkeyLeet: Is https://oejenkins.blob.core.windows.net expected to be a publicly accessible endpoint for these packages? The expectation is that 3rd party devs will use this script to set this up on their local device as well, as documented in GettingStarted.Windows.md.

We should probably build parameterization of source paths into the script explicitly so Ansible and pull from our server caches but users get the default public endpoints.

docs/GettingStartedDocs/GettingStarted.Windows.md
```diff
@@ -103,6 +103,20 @@ C:\> where ocaml
 C:\Program Files\ocpwin64\4.02.1+ocp1-msvc64-20160113\bin\ocaml.exe
 ```
 
+Configuration script
```
CodeMonkeyLeet: This section needs to be moved up before we describe individually installing each of the components, and should be presented as an alternative. For example, move this under the Prerequisites header at the start of the doc:

> Prerequisites
> -------------
>
> The following are prerequisites for building and running Open Enclave on Windows.
> 
> ...
> 
> To deploy all the the prerequisites for building Open Enclave, you can run powershell script `scripts/oe-env-prepare.ps1`:
> ```powershell
> cd scripts
> .\oe-env-prepare.ps1
> ```
> 
> To deploy each prerequisite individually, refer to the sections below.

scripts/oe-env-prepare.ps1
```diff
@@ -0,0 +1,446 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
CodeMonkeyLeet: Naming convention: all scripts should use the _verb-noun_ pattern. For example: `install-windows-prereqs.ps1`

scripts/oe-env-prepare.ps1
george-caluta: Rebooting from the script makes ansible lose connection to the node:
`The error was: winrm.exceptions.WinRMTransportError: ('http', 'Bad HTTP response returned from server. Code 500')
`
I suggest using `win_reboot` from ansible and when run manually, instruct the user to reboot the node manually by displaying this message at the end of the script.

scripts/oe-env-prepare.ps1
ionutbalutoiu: > Is https://oejenkins.blob.core.windows.net expected to be a publicly accessible endpoint for these packages? The expectation is that 3rd party devs will use this script to set this up on their local device as well, as documented in GettingStarted.Windows.md.

@CodeMonkeyLeet Yes, the https://oejenkins.blob.core.windows.net endpoint comes from our OE Jenkins storage account, and the `oejenkins` container is public, so everything starting with https://oejenkins.blob.core.windows.net/oejenkins can be publicly accessible without any problem.

The Intel PSW installer is uploaded in our storage account on purpose, for caching purposes, in case Intel removes the installer from their web server (we encountered something like this with issue https://github.com/microsoft/openenclave/issues/1617).

> 
> We should probably build parameterization of source paths into the script explicitly so Ansible and pull from our server caches but users get the default public endpoints.

Instead of hard-coding the download URLs into the scripts itself, are you suggesting to have script parameters for every download URL ? I think this would be a good idea in case somebody uses the script and caches the script installer URLs into their internal cache web server.

scripts/oe-env-prepare.ps1
ionutbalutoiu: Hum, you're right. But I think we can still reboot with the script, and wait with an Ansible task until the node becomes reachable again.

Keep in mind that I found out recently that `win_reboot` from Ansible doesn't properly reboot a SGX Windows VM from Azure. The node never comes back online, and you have to manually reboot it via the Azure API.

I am preparing a new pull request to have this fixed. Basically, we reboot the node via a CMD shell command, and we wait the node to become reachable over WinRM.

I suggest that you adopt the same approach here. Execute the script, which does a reboot at the end, and in the following task just wait for the node to become reachable over WinRM.

See my changes: https://github.com/ionutbalutoiu/openenclave/commit/96a56dc8c9df8a835d38c3c24865eab4a4c7e44f.

ionutbalutoiu: Small update here.

Just synced offline with @george-caluta. It turns out that rebooting from the PowerShell script causes Ansible to sometimes error out the Ansible tasks that executes the PowerShell script:
```
TASK [Run the OE setup script] *******************************************************************************************************************************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: winrm.exceptions.WinRMTransportError: ('http', 'Bad HTTP response returned from server. Code 500')
fatal: [192.168.88.224]: FAILED! => {"msg": "Unexpected failure during module execution.", "stdout": ""}
```

Thus, we will be rebooting the machine outside the PowerShell script, and the script will only instruct that a reboot is needed.

scripts/oe-env-prepare.ps1
CodeMonkeyLeet: Yes, I am suggesting that the download URLs be parameterized inputs into the script. Our ansible use of it should provide our blob storage endpoint, and the script can default to a set of public endpoints from Intel for public users. I do not want us to serve our cache the the broader internet.

ionutbalutoiu: @CodeMonkeyLeet 

This sounds good. We'll be doing this. We have public endpoints for every file downloaded from the internet, except this one:
```
$AzureDCAPNupkgURL = 'https://oejenkins.blob.core.windows.net/oejenkins/Microsoft.Azure.DCAP.Client.1.0.0.nupkg'
```

The `Microsoft.Azure.DCAP.Client.1.0.0.nupkg` is a custom nupkg file with the `v1.0` release from [Microsoft/Azure-DCAP-Client](https://github.com/Microsoft/Azure-DCAP-Client), hosted in our storage account.

Do you know if there is an official public URL for this nupkg file ?

Ideally, I believe that this one should be attached to the official GitHub releases https://github.com/Microsoft/Azure-DCAP-Client/releases, and we can have the GitHub URL as default download URL in the script. But right now, only an archive with the library built is attached there.

dthaler:
Title says "VM" but the docs/GettingStartedDocs/GettingStarted.Windows.md page is not supposed to be VM specific, as the title is clearly generic.  All sections look ok except the PSW section where the first paragraph is confusing, and ambiguous as to whether/how it applies to physical machines.   (OE is not just for the cloud.)

scripts/oe-env-prepare.ps1
george-caluta: Added official download links for all software except ```Microsoft.Azure.DCAP.Client.1.0.0.nupkg```

CodeMonkeyLeet: @ionutbalutoiu Yes, I agree that it should be pulled from their GitHub releases.

@soccerGB @anitagov Can we work with the owners of Azure DCAP client to publish a copy of their client nupkg from their GitHub? I understand that we've contributed to help stabilize their package drop for Windows.

ionutbalutoiu:
> Title says "VM" but the docs/GettingStartedDocs/GettingStarted.Windows.md page is not supposed to be VM specific, as the title is clearly generic.

Thank-you for the feedback. I changed the pull request title to:
```
Add PowerShell script to configure a clean Windows SGX enabled machine
```
This should be clear enough, but please let me know if you have any other concerns.

> All sections look ok except the PSW section where the first paragraph is confusing, and ambiguous as to whether/how it applies to physical machines. (OE is not just for the cloud.)

Are you referring to [this section](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/GettingStarted.Windows.md#intel-sgx-platform-software-for-windows-psw) ? The first paragraph points out that this requirement is optional if you have:
1. At least Windows 10 with the Fall Creators Update installed 
2. You are using a Windows Server 2016 image with an Azure Confidential Compute (ACC) VM

I guess your confusion is about `1.` as it not clearly stated whether it applies to virtual machines or physical machines.

I believe that the statement applies to both, as you will get `AESMService` with Windows 10 (updated to at least 1709, Fall Creators Update) regardless of running in a virtual machine or a physical machine, as long as the SGX device is available to the operating system.

/cc @CodeMonkeyLeet @anitagov @johnkord to get a confirmation on this one.

scripts/oe-env-prepare.ps1
ionutbalutoiu: @CodeMonkeyLeet @soccerGB @anitagov 

I created the issue https://github.com/Microsoft/Azure-DCAP-Client/issues/74 to have this tracked.

In my opinion, we can merge this pull request with our internal storage account download link, and add a `TODO` comment in order to have this updated later with the official download link.

I'll make sure I follow up later and address the `TODO` once the `nupkg` file is published with the Azure-DCAP-Client releases.

ionutbalutoiu:
> LGTM. Can you track a couple of additional items:
> 
> * We should have a task in the openenclave repo to track fixing the replacement of the Azure-DCAP-client endpoint
> * We need to cache all package dependencies we depend on for setting up our configuration that's Azure controlled (i.e. not just the Intel DCAP and PSW packages) This is also going to be true for the Linux configurations, Docker containers etc. as well, so let's break that down in additional tasks.

Sounds good! These are my follow-up issue regarding these tasks:
* Issue https://github.com/openenclave/openenclave/issues/1988
* Issue https://github.com/openenclave/openenclave/issues/1989

ionutbalutoiu:
Merging this one. Thank-you everyone!

bors r+

PR #1935, opened 2019-06-13
Handle OE_SIMULATION appropriately in VS extension

dthaler:
Fixes issue reported by Brian Telfer in the bug bash

### DIFF
--- a/new_platforms/vsextension/ProjectWizard/ImportEnclaveCommand.cs
+++ b/new_platforms/vsextension/ProjectWizard/ImportEnclaveCommand.cs
@@ -393,11 +393,20 @@ namespace OpenEnclaveSDK
                     {
                         var config3 = config as VCConfiguration3;
                         string name = config.Name;
+
+                        var clRule = config.Rules.Item("CL") as IVCRulePropertyStorage;
+                        string preprocessorDefinitions = clRule.GetUnevaluatedPropertyValue("PreprocessorDefinitions");
+                        if (isWindows)
+                        {
+                            // Allow calling getenv() in the generated host code.
+                            preprocessorDefinitions = "_CRT_SECURE_NO_WARNINGS;" + preprocessorDefinitions;
+                            clRule.SetPropertyValue("PreprocessorDefinitions", preprocessorDefinitions);
+                        }
+
                         if (name.Contains("ARM"))
                         {
-                            var clRule = config.Rules.Item("CL") as IVCRulePropertyStorage;
-                            string value = clRule.GetUnevaluatedPropertyValue("PreprocessorDefinitions");
-                            clRule.SetPropertyValue("PreprocessorDefinitions", "_ARM_;" + value);
+                            preprocessorDefinitions = "_ARM_;" + preprocessorDefinitions;
+                            clRule.SetPropertyValue("PreprocessorDefinitions", preprocessorDefinitions);
 
                             if (isWindows) {
                                 // Enable compiling Win32 for ARM.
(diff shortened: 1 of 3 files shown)


new_platforms/vsextension/ProjectWizard/ImportEnclaveCommand.cs
```diff
@@ -393,11 +393,20 @@ private async void Execute(object sender, EventArgs e)
...
+                            preprocessorDefinitions = "_CRT_SECURE_NO_WARNINGS;" + preprocessorDefinitions;
+                            clRule.SetPropertyValue("PreprocessorDefinitions", preprocessorDefinitions);
+                        }
+
                         if (name.Contains("ARM"))
                         {
-                            var clRule = config.Rules.Item("CL") as IVCRulePropertyStorage;
-                            string value = clRule.GetUnevaluatedPropertyValue("PreprocessorDefinitions");
-                            clRule.SetPropertyValue("PreprocessorDefinitions", "_ARM_;" + value);
+                            preprocessorDefinitions = "_ARM_;" + preprocessorDefinitions;
```
HernanGatta: I hadn't noticed this before: GCC defines `__arm__` and MSVC defines `_M_ARM`. Why is `_ARM_` necessary?

[Reference](https://sourceforge.net/p/predef/wiki/Architectures/)

new_platforms/vsextension/ProjectWizard/ImportEnclaveCommand.cs
```diff
@@ -393,11 +393,20 @@ private async void Execute(object sender, EventArgs e)
                     {
                         var config3 = config as VCConfiguration3;
                         string name = config.Name;
+
+                        var clRule = config.Rules.Item("CL") as IVCRulePropertyStorage;
+                        string preprocessorDefinitions = clRule.GetUnevaluatedPropertyValue("PreprocessorDefinitions");
+                        if (isWindows)
+                        {
+                            // Allow calling getenv() in the generated host code.
+                            preprocessorDefinitions = "_CRT_SECURE_NO_WARNINGS;" + preprocessorDefinitions;
```
HernanGatta: Instead of defining `_CRT_SECURE_NO_WARNINGS` for the entire host, can you surround the call to `getenv` as:

```
#ifdef _MSC_VER
#pragma warning(push)
#pragma warning(disable:4996)
#endif

const char* simulation = getenv("OE_SIMULATION");

#ifdef _MSC_VER
#pragma warning(pop)
#endif
```

new_platforms/vsextension/ProjectWizard/ImportEnclaveCommand.cs
dthaler: not sure, but since it's not new, I'll investigate after this change

new_platforms/vsextension/ProjectWizard/ImportEnclaveCommand.cs
dthaler: Will do

dthaler: Done

new_platforms/vsextension/ProjectWizard/ImportEnclaveCommand.cs
dthaler: In the new_platforms branch, tcps.h has this:
tcps.h:#if defined(OE_SIMULATE_OPTEE) || defined(_ARM_)

This is not needed in the master branch.

PR #1939, opened 2019-06-13
Re-enable the crypto tests on Windows

ionutbalutoiu:
This change reverts PR https://github.com/microsoft/openenclave/pull/1926 where these tests were temporarily disabled.

The regression was fixed by reverting Git to `2.19.1` in order to have OpenSSL `1.0.2`.

Fixes https://github.com/microsoft/openenclave/issues/1927


ionutbalutoiu:
bors try

anitagov:
I am not so keen that these tests be enabled on Windows because the underlying issue where these tests FAIL with newer versions of openssl still exist. Would be nice to have the fixes in the core/tests in and then enable them with CI. My 2 cents.

PR #1948, opened 2019-06-14
Fix overlapped copy error in oe_iov_sync()

mikbras:
This PR fixes an overlapped copy error that was introduced when **memcpy** was replaced with **oe_memcpy_s**.


mikbras:
bors try

PR #1950, opened 2019-06-16
Fix VS include paths and compiler options for building enclaves on Linux

dthaler:
Use the compiler options per Anand's feedback

### DIFF
--- a/new_platforms/nuget/build/native/openenclave.targets
+++ b/new_platforms/nuget/build/native/openenclave.targets
@@ -231,7 +231,9 @@
 
     <ItemDefinitionGroup Condition="'$(OEType)' == 'Enclave' And $(OEIsMaster) == True">
         <ClCompile>
-            <AdditionalIncludeDirectories>/opt/openenclave/include/openenclave/3rdparty/libc;/opt/openenclave/include/openenclave/3rdparty/libcxx;/opt/openenclave/include/openenclave/3rdparty;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+            <!-- Compile as C++ as a workaround for not being able to have separate include paths for C vs C++ -->
+            <CompileAs>CompileAsCpp</CompileAs>
+            <AdditionalIncludeDirectories>/opt/openenclave/include/openenclave/3rdparty/libcxx;/opt/openenclave/include/openenclave/3rdparty/libc;/opt/openenclave/include/openenclave/3rdparty;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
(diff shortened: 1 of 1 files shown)


new_platforms/nuget/build/native/openenclave.targets
```diff
@@ -229,14 +229,22 @@
 
     <!-- Additional Include Directories for Enclaves only -->
 
-    <!-- *-SGX-Debug and *-SGX-Simulation-Debug -->
+    <ItemDefinitionGroup Condition="'$(OEType)' == 'Enclave' And $(OEIsMaster) == True">
+        <ClCompile>
+            <AdditionalIncludeDirectories>/opt/openenclave/include/openenclave/3rdparty/libc;/opt/openenclave/include/openenclave/3rdparty/libcxx;/opt/openenclave/include/openenclave/3rdparty;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+            <AdditionalOptions>-nostdinc -m64 -fPIE -ftls-model=local-exec -mllvm -x86-speculative-load-hardening %(AdditionalOptions)</AdditionalOptions>
```
anakrish: Is there some rules as to how these additional include directories would appear in the command line?

For cpp files, the libcxx include folder must appear before the libc include folders. 
The C++ headers often assume that the C headers are available to be included and use include_next to include their C equivalent. 
E.g https://github.com/microsoft/openenclave/blob/9e7cf8f1ebd1e37539a2fafaf577405288fbbebc/3rdparty/libcxx/libcxx/include/cstddef#L44

If the C include folder appears before the C++ include folder in the include path, then the include_next would fail.
Can you do:
```c
#include <cstddef>
#include <iostream>
```
and see if the compilation succeeds?


However when C files are being compiled, we need to make sure that the C version of an header (say stdio.h) is being picked up instead of the C++ version since the C++ version uses C++ specific syntax.
This can be achieved by not having the C++ headers in the include path or by having the C headers come before the C++ headers in the include path.

dthaler: That sounds like a fragile design.  I am not sure whether VS supports allowing the order to vary by file type within the project file.   Why can't the include files be fixed to allow using a consistent order?

dthaler: For example, the design could be fixed by having a C include file include the C++ file inside an #ifdef __cplusplus, or vice versa (with ifndef)

dthaler:
C source files don't get compilation errors due to the use of <CompileAs>CompileAsCpp</CompileAs> which translates to the -x c++ compiler argument, which means C++ headers in the include path are fine.

PR #1958, opened 2019-06-18
Support for deep copying out (and in/out) parameters

andschwa:
This PR adds experimental support for `out` and `in/out` parameters. A couple known limitations:

In the edge case of struct A having an array of struct B, and struct B has some array C with size D as a member of struct B, the current implementation will not generate correct code. Note that this does not affect struct A having an array E with size F as a member of struct A, even where the function declares an array of struct A, as the issues lies in expression generated to count host side pointers, and function argument pointers do not need to be counted.

Another limitation is that we do not pass the original direction of a pointer that is being deep copied, and so we assume all pointers underneath are `in/out` even if they were in fact only `in` or `out`. This must be fixed before this is moved out of the experimental stage.

### DIFF
--- a/include/openenclave/edger8r/common.h
+++ b/include/openenclave/edger8r/common.h
@@ -51,7 +51,7 @@ OE_INLINE oe_result_t oe_add_size(size_t* total, size_t size)
         goto done;
     }
 
-    // Add rounded-size and check for overlow.
+    // Add rounded-size and check for overflow.
     sum = *total + rsize;
     if (sum < *total)
     {
--- a/tests/oeedger8r/CMakeLists.txt
+++ b/tests/oeedger8r/CMakeLists.txt
@@ -10,4 +10,6 @@ endif()
 add_enclave_test(tests/oeedger8r edl_host edl_enc)
 add_enclave_test(tests/oeedger8r_other edl_host edl_other_enc)
 
+add_custom_target(edl_gen DEPENDS edl_enc_gen edl_host_gen)
+
 add_subdirectory(behavior)
--- a/tests/oeedger8r/edl/deepcopy.edl
+++ b/tests/oeedger8r/edl/deepcopy.edl
@@ -41,6 +41,11 @@ enclave {
     [count=3] CountStruct* array_of_struct;
   };
 
+  struct SuperNestedStruct {
+    // TODO: We should eventually support count being a member of the struct.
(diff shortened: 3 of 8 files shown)


andschwa:
@gupta-ak Our available queue implementation is not very well suited to a FIFO use 😢 

@radhikaj I ended up doing your original suggestion to avoid having to deal with unused warnings.

@anakrish This does not support "unlimited" out parameters, just 512. That's technically changeable, but really we need something with the semantics of `std::vector`.

andschwa:
Part of #1839.

andschwa:
Hi @anakrish, I'd really appreciate a review from you as I'd like to know which direction to go with this.

andschwa:
Rebased.

bors try

andschwa:
> host/syscall_u.c: In function 'oe_syscall_stub_ecall':
> host/syscall_u.c:23:12: error: variable '_ptrs_index' set but not used [-Werror=unused-but-set-variable]

Easy to fix.

andschwa:
@radhikaj @anakrish Please review. Latest iteration produces an expression for the count of pointers we need to save and restore, and does not generate any array to store them if there are no pointers to deep copy. This should avoid any issues with non-SGX implementations.

There's still a lot of clean up to be done, but I'd much prefer to do that afterward, so this can start getting tested by the likes of @mikbras.

Thanks!

andschwa:
bors try

andschwa:
Rebased (again) and simplified the logic.

bors try

andschwa:
While writing more tests, I encountered a currently unsupported scenario:

An array of structs with a nested deep copy member where the count of the nested deep copy member is a member of the superstruct. In this case, the generated code does not yet produce the correct count, as it should be the sum of the count members from each item in the array.

This is an extreme edge case, that can be supported in the future.

mikbras:
This test that I added crashes and I'm not sure why. It passes IOVEC struct where iov[0].base == NULL.

https://github.com/openenclave/openenclave/compare/andschwa/in-out...mikbras.in-out2

andschwa:
Tests fixed @mikbras (Thanks so much!)

andschwa:
Ah @mikbras I can't use your test for now. On Windows, it gets really mad at `void* _ptrs[n * ...]` where `n = 0`.

andschwa:
bors r+

(Did I ever mention how much I dislike Windows?)

PR #1962, opened 2019-06-18
[Ansible] Ignore preinstalled ocamlbuild

ionutbalutoiu:
The package `ocamlbuild` is required when installing `ocamlformat`
via `opam`.

However, if `ocamlbuild` is preinstalled, the `ocamlbuild` installation
from `opam` will fail unless forced to ignore the preinstalled
package.


ionutbalutoiu:
Dropped in favor of https://github.com/openenclave/openenclave/pull/2011

Issue #1834, opened 2019-05-28
Stdin/stdout/stderr dup in enclave


johnkord:
This is taken from #1822

There are some issues with opening /dev/[stdin/stdout/stderr], instead of duplicating 0,1,2. They would have seperate entries in the kernel file description table, so they would have different file poisition, etc. This means that all the reads / writes to stdin/stdout/stderr are completely unsynchronized between the host and the guest. For example, if you have something like ./host ./enc > log.txt, the enclave and host would overwrite the each other due to no file pointer synchronization. Further, /dev/std... are just symlinks to the process's 0,1,2 fds, so if the host closes or changes them before the enclave runs, the enclave would just pick up the changes. Personally, I think just doing dup on 0,1,2 would be a better alternative. There are some disadvantages of the dup approach though. For example, the enclave would pick up the fcntl settings and anything that changes the kernel file description object.

gupta-ak:
Here's a quick sample program that illustrates the issue with opening `/dev/stdout`: 

```C
#include <stdio.h>
#include <sys/fcntl.h>
#include <unistd.h>
#include <string.h>

int main(int argc, char** argv)
{
        int fd = -1;

        if (argc > 1 && strcmp(argv[1], "host") == 0)
                fd = 1;
        else if (argc > 1 && strcmp(argv[1], "dup") == 0)
                fd = dup(1);
        else
                fd =  open("/dev/stdout", O_WRONLY);

        if (fd == -1)
        {
                perror("Failed to open stdout");
                return -1;
        }

        write(1, "Hello ", sizeof("Hello ") - 1);
        write(fd, "World!\n", sizeof("World!\n") - 1);
        close(fd);
        return 0;
}
```

If stdout is the tty, you get the expected behavior:
``` bash
akashg@akash-client:~/test-posix$ ./a.out
Hello World!
```

On the other hand, if stdout is a file, you see that it overwrites the file:
```
akashg@akash-client:~/test-posix$ ./a.out > file.txt
akashg@akash-client:~/test-posix$ cat file.txt
World!
```

gupta-ak:
You can see this in hello world sample in master too:
```
akagup@akash-acc:~/openenclave/samples/helloworld$ ./host/helloworldhost ./enclave/helloworldenc.signed 
Hello world from the enclave
Enclave called into host to print: Hello World!
```
vs
```
[...]

PR #1966, opened 2019-06-18
Fix consolefs output when redirected to files.

gupta-ak:
This PR does two things:

1. Fixes the `print` test CMake so that the test actually runs. The test shows that the current master branch fails this test.
2. Fixes consolefs to duplicate the standard file descriptors instead of opening them directly, fixing the issues seen in the `print` test.


yakman2020:
Does this want to be in 0.6? It looks like it.

Does it satisfy the issue #1834?

gupta-ak:
bors try

anitagov:
Nice job @gupta-ak

PR #1968, opened 2019-06-19
Incorrect variable initialization in _do_eenter

anitagov:
Noticed the incorrect variable initialization in _do_eenter() function in host/sgx/calls.c a while back but never got around to issuing this PR.

-196,7 +196,7 @@ static oe_result_t _do_eenter(
        *func_out = 0;

    if (result_out)
        **func_out = 0;**
   This should be *result_out = 0;

    if (!code_out || !func_out || !result_out || !arg_out)

Reason we never had an issue with this is because *result_out is always NULL when we enter this routine.

### DIFF
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -196,7 +196,7 @@ static oe_result_t _do_eenter(
         *func_out = 0;
 
     if (result_out)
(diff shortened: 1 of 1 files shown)


anitagov:
bors try

PR #1972, opened 2019-06-19
Update all references to old location of repo

andschwa:
While these are redirected by GitHub, it is better that we explicitly update them.


andschwa:
Flaky CI grr...

bors r+

PR #1974, opened 2019-06-19
Implementation of synchronous switchless ecalls.

EMumau:
This is the implementation of synchronous switchless ecalls.  It includes every thing except for the changes to the edger8r.

* enclave/core/sgx/calls.c:  Has the enclave side worker thread management and the methods for marshaling calls to the user defined enclave methods.
* host/sgx/calls.c:  Has the host side worker thread management and the methods for marshaling calls from the host to the enclave.
* host/sgx/create.c:  Added a function call to shutdown the worker thread.
* host/sgx/enclave.h:  Added the switchless manager to the enclave object.
* host/switchless_manager.h:  Declarations for the host side switchless objects and enums.
* host/switchless_manager.c:  Implementations of the host side switchless functionality.
* include/openenclave/edger8r/host.h:  Expose the marshaling function to the edger8r generated code.
* include/openenclave/internal/calls.h:  Extended the enumeration to account for the enclave worker thread.
* include/openenclave/internal/switchless.h:  introduce a data structure with all of the members shared between the host and the enclave.

* tests/backtrace/host/host.cpp:  Changed the hard-coded stack trace to adjust for changes to enclave/core/sgx/calls.c.

* tests/switchless/*:  Tests for the switchless infrastructure.


anakrish:
@jxyang Can you review this? This is the switchless feature.

host/sgx/calls.c
```diff
@@ -898,6 +899,41 @@ oe_result_t oe_call_enclave_function(
...
+    /* start the switchless worker thread if needed */
+    oe_switchless_manager_startup(enclave);
+
+    /* set the synchrononization flag */
+    node->lock = 1;
+    oe_lockless_queue_push_back(
+        &(enclave->switchless_manager.switchless.ecall_queue), &(node->_node));
+
+    /* wait for the flag to clear */
+    while (__atomic_test_and_set(&(node->lock), __ATOMIC_ACQUIRE))
```
anakrish: Given that we are only waiting for **node->lock** to become 0, **__atomic_load** seems more appropriate here.

host/switchless_manager.c
```diff
@@ -0,0 +1,172 @@
...
+static int _thread_create(
+    oe_thread_t* thread,
+    _thread_op_t op,
+    _thread_arg_t arg)
+{
+    *thread = CreateThread(NULL, 0, op, arg, 0, NULL);
+    return NULL == *thread;
+} /* _thread_create */
+
+static void _thread_join(_thread_arg_t thread)
```
anakrish: These function would better fit in hostthread.h, windows/hostthread.c, linux/hostthread.c

host/switchless_manager.c
```diff
@@ -0,0 +1,172 @@
...
+        &(enclave->switchless_manager.switchless.state),
+        OE_SWITCHLESS_STATE_STARTING,
+        OE_SWITCHLESS_STATE_STOPPED);
+#elif defined __GNUC__
+    __atomic_compare_exchange_n(
+        &(enclave->switchless_manager.switchless.state),
+        &state,
+        OE_SWITCHLESS_STATE_STARTING,
+        1,
+        __ATOMIC_ACQ_REL,
```
anakrish: Can you describe why you prefer __ATOMIC_ACQ_REL instead of __ATOMIC_RELEASE for the success case?
We can aim for consistent use of acquire/release everywhere.

tests/switchless/enc/enc.c
```diff
@@ -0,0 +1,31 @@
...
+    {
+        args[i].sum = args[i].arg1 + args[i].arg2;
+    }
+}
+
+OE_SET_ENCLAVE_SGX(
+    1,    /* ProductID */
+    1,    /* SecurityVersion */
+    true, /* AllowDebug */
+    1024, /* HeapPageCount */
```
anakrish: We may want to reduce the page count to make the enclave run on Windows.

tests/switchless/host/switchless_args.h
```diff
@@ -0,0 +1,68 @@
...
+ */
+#ifndef EDGER8R_SWITCHLESS_ARGS_H
+#define EDGER8R_SWITCHLESS_ARGS_H
+
+#include <stdint.h>
+#include <stdlib.h> /* for wchar_t */
+
+/* #include <errno.h> - Errno propagation not enabled so not included. */
+
+#include <openenclave/bits/result.h>
```
anakrish: Can you create an issue to remove these auto-generated files in test

anakrish:
@andschwa @EMumau 
This PR has lots of auto-generated code. 
It is preferable to change the edger8r instead. Can you spin up another PR for it or add the edger8r changes to this PR.

enclave/core/sgx/calls.c
```diff
@@ -236,126 +243,236 @@ oe_result_t oe_register_ecall_function_table(
...
+{
+    oe_result_t result = OE_OK;
+    oe_switchless* switchless = (oe_switchless*)arg_in;
+
+    /* loop until the state changes to stopping */
+    while (OE_SWITCHLESS_STATE_STOPPING != _get_switchless_state(switchless))
+    {
+        /* note: this pop needs to be protected by a lock if there is more than
+         * one worker thread */
+        oe_switchless_synchronous_ecall* ecall_node =
```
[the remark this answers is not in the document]
EMumau: This PR only allows a single worker thread.  I made that note in case a future PR changes that.

host/sgx/calls.c
EMumau: This is using the atomic flag operations instead of the atomic value operations.  On some hardware, the flags are lighter weight, and they are also guaranteed to be atomic on all platforms where atomic values are not (but likely are).  For parity sake, __atomic_test_and_set is used on this thread while __atomic_clear is used on the other thread.

enclave/core/sgx/calls.c
```diff
@@ -236,126 +243,236 @@ oe_result_t oe_register_ecall_function_table(
...
+            (oe_switchless_synchronous_ecall*)oe_lockless_queue_pop_front(
```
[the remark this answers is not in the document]
jxyang: I think this is fine as long as we are careful. The queue is just a container to the parameters we pass into enclave. The parameters are used to be on the host side anyway and are validated extensively in `_handle_call_enclave_function` (now `_handle_ecall_function`). The added functionality is enqueue/dequeue which have low risk.

host/switchless_manager.c
EMumau: __atomic_compare_exchange_n is read-then-write operation.  In both the success and failure cases an __ATOMIC_ACQUIRE is performed to refresh the current value.  The success case writes to the value necessitating an __ATOMIC_RELEASE as well.  The failure case does not modify the value.  So the success case needs __ATOMIC_ACQ_REL where the failure case only needs __ATOMIC_ACQUIRE.

enclave/core/sgx/calls.c
```diff
@@ -236,126 +243,236 @@ oe_result_t oe_register_ecall_function_table(
...
+        ecall->input_buffer_size,
+        ecall->output_buffer,
+        ecall->output_buffer_size,
+        &(ecall->output_bytes_written),
+        &(ecall->result));
+
+#ifdef _MSC_VER
+    _bittestandreset(&(ecall->lock), 31);
+#elif defined __GNUC__
+    __atomic_clear(&(ecall->lock), __ATOMIC_RELEASE);
```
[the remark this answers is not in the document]
jxyang: NM, I saw it's used in oe_call_synchronous_switchless_enclave_function

tests/switchless/host/host.c
```diff
@@ -0,0 +1,234 @@
...
+    /* generate sample data */
+    for (size_t i = 0; i < SAMPLE_SIZE; ++i)
+    {
+        args[i].arg1 = generate_random_number();
+        args[i].arg2 = generate_random_number();
+        args[i].sum = 0;
+    }
+
+    /* explicitly start the enclave worker thread */
+    oe_switchless_manager_startup(enclave);
```
jxyang: I don't think this is necessary as the first switchless ecall will do it.

tests/switchless/host/host.c
```diff
@@ -0,0 +1,234 @@
...
+
+    /* record start time */
+    get_time(&start);
+
+    for (size_t i = 0; i < SAMPLE_SIZE; ++i)
+    {
+        OE_TEST(
+            OE_OK == synchronous_switchless_enc_sum(
```
jxyang: The way we are testing now is even though this call is in a loop, we would enqueue the job and wait until the dequeued by the enclave worker thread. Essentially, this means the length of the queue is always <= 1. We should try to use more than one thread to call into this function to stress testing the queue implementation.

jxyang:
I suggest we take a look of the following functions:  `oe_ecall`, `_do_eenter`, `oe_enter` (both host and enclave side), look for security measures in those function, and think about the effect of skipping them in switchless calls.

enclave/core/sgx/calls.c
```diff
@@ -236,126 +243,236 @@ oe_result_t oe_register_ecall_function_table(
...
-    size_t output_bytes_written = 0;
+    uint8_t* local_input_buffer = NULL;
+    uint8_t* local_output_buffer = NULL;
+    size_t local_output_bytes_written = 0;
     ecall_table_t ecall_table;
-
-    // Ensure that args lies outside the enclave.
-    if (!oe_is_outside_enclave(
-            (void*)arg_in, sizeof(oe_call_enclave_function_args_t)))
-        OE_RAISE(OE_INVALID_PARAMETER);
```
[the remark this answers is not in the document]
EMumau: These tests existed prior to my changes.  This question is outside of the scope of this PR.

host/switchless_manager.c
anakrish: > __ATOMIC_ACQUIRE is performed to refresh the current value
I don't quite follow. The current value of the variable being manipulated will always be refreshed due to the atomic operation, isn't it?

Couple of suggestions:
- It would be better to capture these rationales in code comments as opposed to PR comments so that a future reader can easily refer to them in source file without having to search for this PR in github.
- Is oe_switchless_manager_startup expected to be called only once? If so, it is better to wrap the call in oe_once

enclave/core/sgx/calls.c
```diff
@@ -236,126 +243,236 @@ oe_result_t oe_register_ecall_function_table(
...
         // Copy outputs to host memory.
-        memcpy(args.output_buffer, output_buffer, output_bytes_written);
+        memcpy(
+            src_output_buffer, local_output_buffer, local_output_bytes_written);
 
         // The ecall succeeded.
-        args_ptr->output_bytes_written = output_bytes_written;
-        args_ptr->result = OE_OK;
+        *src_output_bytes_written = local_output_bytes_written;
+        *src_result = OE_OK;
```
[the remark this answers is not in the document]
EMumau: I looked into this.  In the ecall path, this value is set prior to calling and only set to OK at this smae point.  I have added a change to the call path to the synchronous switchless path to mimic that same behavior.

enclave/core/sgx/calls.c
```diff
@@ -27,6 +28,12 @@
 #include "report.h"
 #include "td.h"
 
+#if _MSC_VER
+#include <intrin.h>
+#elif defined __GNUC__
+#include <pthread.h>
```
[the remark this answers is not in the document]
EMumau: removed.

enclave/core/sgx/calls.c
```diff
@@ -236,126 +243,236 @@ oe_result_t oe_register_ecall_function_table(
...
+    memset(local_output_buffer, 0, output_buffer_size);
 
     // Call the function.
     func(
-        input_buffer,
-        args.input_buffer_size,
-        output_buffer,
-        args.output_buffer_size,
-        &output_bytes_written);
+        local_input_buffer,
```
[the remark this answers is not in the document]
EMumau: updated.

enclave/core/sgx/calls.c
```diff
@@ -236,126 +243,236 @@ oe_result_t oe_register_ecall_function_table(
...
+        ecall->function_id,
+        ecall->input_buffer,
+        ecall->input_buffer_size,
+        ecall->output_buffer,
+        ecall->output_buffer_size,
+        &(ecall->output_bytes_written),
+        &(ecall->result));
+
+#ifdef _MSC_VER
+    _bittestandreset(&(ecall->lock), 31);
```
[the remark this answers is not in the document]
EMumau: removed.

enclave/core/sgx/calls.c
```diff
@@ -236,126 +243,236 @@ oe_result_t oe_register_ecall_function_table(
...
+#elif defined __GNUC__
+    __atomic_clear(&(ecall->lock), __ATOMIC_RELEASE);
+#endif /* _MSC_VER or __GNUC__ */
+
+done:
+    return result;
+} /* _handle_synchronous_switchless_ecall */
+
+static oe_switchless_state _get_switchless_state(oe_switchless* switchless)
```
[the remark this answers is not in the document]
EMumau: changed.

tests/switchless/host/host.c
EMumau: Yes, the first switchless ecall will do this.  I did it explicitly in this case to move the thread start process outside of the time gathering.

tests/switchless/host/host.c
EMumau: There are a suite of tests explicitly for the queue which provide multi-producers testing.  I will also add a test to use multiple producers to the switchless test.

host/sgx/calls.c
```diff
@@ -898,6 +899,41 @@ oe_result_t oe_call_enclave_function(
         output_bytes_written);
 }
 
+oe_result_t oe_call_synchronous_switchless_enclave_function(
+    oe_enclave_t* enclave,
+    oe_switchless_synchronous_ecall* node)
+{
+    oe_result_t result = OE_OK;
```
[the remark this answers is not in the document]
EMumau: changed.

enclave/core/sgx/calls.c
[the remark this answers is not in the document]
EMumau: I made the _t suffix change for the typedefs that I added.  There is not consistent enough use in the code base to call that a convention.

host/switchless_manager.c
```diff
@@ -0,0 +1,172 @@
...
+} /* oe_switchless_manager_init */
+
+oe_result_t oe_switchless_manager_startup(oe_enclave_t* enclave)
+{
+    oe_result_t result = OE_OK;
+    oe_switchless_state_t state = OE_SWITCHLESS_STATE_STOPPED;
+
+    /* check the current state:
+       - if it is stopped then change to starting */
+#ifdef _MSC_VER
```
jxyang: I'd define the platform dependent atomic functions in a separate file, say `common/atomic.c`, and hide the `#ifdef _MSC_VER` and `#elif defined __GNUC__` there. I think it's much cleaner that way.

tests/switchless/host/host.c
jxyang: That sounds good. But the relevant question is should we expose this function to the user. It seems to me too implementation dependent.

host/switchless_manager.c
EMumau: changed.

enclave/core/sgx/calls.c
```diff
@@ -236,126 +237,225 @@ oe_result_t oe_register_ecall_function_table(
...
+    memset(enc_output_buffer, 0, output_buffer_size);
 
     // Call the function.
     func(
-        input_buffer,
-        args.input_buffer_size,
-        output_buffer,
-        args.output_buffer_size,
-        &output_bytes_written);
+        enc_input_buffer,
```
anakrish: The rewrite using **enc**, **host** prefixes is more readable.

enclave/core/sgx/calls.c
```diff
@@ -236,126 +237,225 @@ oe_result_t oe_register_ecall_function_table(
...
+    oe_result_t result = OE_OK;
+
+    // Ensure that args lies outside the enclave.
+    if (!oe_is_outside_enclave(
+            (void*)arg_in, sizeof(oe_call_enclave_function_args_t)))
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+    // Copy args to enclave memory to avoid TOCTOU issues.
+    args_ptr = (oe_call_enclave_function_args_t*)arg_in;
+    args = *args_ptr;
```
anakrish: @jxyang With the new version, the call is forwarded to _handle_ecall_function. This implies that the arguments are copied. Can you chime in on whether  the explicit copying in enclave memory in line 385 is needed?
Note: Keep in mind that the compiler could inline any function in the file; but has to preserve C++ semantics.

enclave/core/sgx/calls.c
```diff
@@ -236,126 +237,225 @@ oe_result_t oe_register_ecall_function_table(
...
+        &(args_ptr->result));
+
+done:
     return result;
 }
 
+static oe_result_t _handle_synchronous_switchless_ecall(
+    oe_switchless_synchronous_ecall_t* ecall)
+{
+    oe_result_t result = OE_OK;
```
anakrish: Our pattern is to initialize result to OE_FAILURE or OE_UNEXPECTED and then to set it to OE_OK after all the operations have succeeded.

enclave/core/sgx/calls.c
```diff
@@ -236,126 +237,225 @@ oe_result_t oe_register_ecall_function_table(
...
+
+    // Ensure that args lies outside the enclave.
+    if (!oe_is_outside_enclave(
+            (void*)ecall, sizeof(oe_switchless_synchronous_ecall_t)))
+    {
+        OE_RAISE(OE_INVALID_PARAMETER);
+    }
+
+    result = _handle_ecall_function(
```
anakrish: @jxyang Can you confirm if we need an lfence here? In the boundary functions, we need to ensure that there is an lfence after all the early returns have been executed.

enclave/core/sgx/calls.c
```diff
@@ -236,126 +243,236 @@ oe_result_t oe_register_ecall_function_table(
...
+
+    /* loop until the state changes to stopping */
+    while (OE_SWITCHLESS_STATE_STOPPING != _get_switchless_state(switchless))
+    {
+        /* note: this pop needs to be protected by a lock if there is more than
+         * one worker thread */
+        oe_switchless_synchronous_ecall* ecall_node =
+            (oe_switchless_synchronous_ecall*)oe_lockless_queue_pop_front(
+                &(switchless->ecall_queue));
+        if (NULL != ecall_node)
```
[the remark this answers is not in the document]
anakrish: @EMumau @jxyang I see a security issue here. Consider the case where the host sets **switchless->ecall_queue** to point to a memory location within the enclave. In that case, we would end up modifying that memory location.

host/sgx/create.c
```diff
@@ -727,6 +727,9 @@ oe_result_t oe_terminate_enclave(oe_enclave_t* enclave)
     if (!enclave || enclave->magic != ENCLAVE_MAGIC)
         OE_RAISE(OE_INVALID_PARAMETER);
 
+    /* stop the switchless manager */
+    oe_switchless_manager_shutdown(enclave);
+
```
anakrish: Can the enclave destructor make switchless ecalls? If so, the switchless-manager must be shutdown later.

host/switchless_manager.c
```diff
@@ -0,0 +1,126 @@
...
+void oe_switchless_manager_init(oe_enclave_t* enclave)
+{
+    enclave->switchless_manager.switchless.state = OE_SWITCHLESS_STATE_STOPPED;
+    oe_lockless_queue_init(
+        &(enclave->switchless_manager.switchless.ecall_queue));
+} /* oe_switchless_manager_init */
+
+oe_result_t oe_switchless_manager_startup(oe_enclave_t* enclave)
+{
+    oe_result_t result = OE_OK;
```
anakrish: The pattern is to initialize result with OE_FAILURE or OE_UNEXPECTED and then to set it to OE_OK only when all the operations have succeeded.

host/switchless_manager.c
```diff
@@ -0,0 +1,126 @@
...
+        }
+        else
+        {
+            result = OE_FAILURE;
+        }
+    }
+    else
+    {
+        /* invalid state */
+        result = OE_FAILURE;
```
anakrish: If two threads simultaneously make a switchless ecall, then they might end up simultaneously calling  startup. As written currently, one of the threads will create the worker thread, whereas the other would return OE_FAILURE. This seems wrong.

enclave/core/sgx/calls.c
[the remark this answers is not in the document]
jxyang: Agreed. We should check whether the queue is outside of enclave.

enclave/core/sgx/calls.c
jxyang: I don't think the copying is necessary. Typically a TOCTOU attack happens in the following way:
```c
func(struct S* p) {
   if (p->f1 != NULL) {
       ... // executing some code, during which p->f1 could be set to NULL  
       a = *p->f1;
   }
}
```
In this case, `oe_is_outside_enclave` is checking `arg_in`, but not individual fields inside it. I don't believe there is risk of TOCTOU even in the original code.

enclave/core/sgx/calls.c
jxyang: I don't think we need it here. Typically Spectre-1 takes advantage of speculative execution on a privileged path, e.g., OOB access into kernel memory. Here, the path can be freely executed by giving a valid `ecall` address anyway.

host/switchless_manager.c
EMumau: moved.

host/switchless_manager.c
EMumau: changed.

enclave/core/sgx/calls.c
EMumau: changed.

enclave/core/sgx/calls.c
EMumau: I moved the __atomic_clear beyond the "done" label.  There was a potential that this would have deadlocked the host in the case that the ecall was not outside the enclave.

host/switchless_manager.c
EMumau: oe_switchless_manager_startup now returns OE_OK in all cases unless thread create fails.

host/switchless_manager.c
EMumau: - I've changed the operations to add more comments
- I've changed oe_switchless_manager_startup/shutdown to be re-entrant.  Neither function will return until the switchless manager is in the desired state (with the exception of thread creation failure).
- I've added comments to explain the acquire/release semantics.
- oe_switchless_manager_startup now returns OE_OK in all cases unless thread create fails.

- oe_once is not needed as re-entrant calls are handled.  Additionally, oe_once would not work with multiple enclaves, or if the enclave is terminated and then created.

host/switchless_manager.c
EMumau: I've added code to ensure that only one thread can perform the startup.  If more than one thread calls startup at the same time, the second thread will wait until the first thread completes the startup and then return success.

anitagov:
@EMumau - I am yet to review the code but applaud the great work that went into the design and the code. 

Do all the tests exercise this new flow automatically? If so, the following will help - 
From the testing standpoint, CI alone isn't enough as it will not catch race conditions. So, please regress using GCC and clang on your ACC VM
_ctest --repeat-until-fail <large-number> where large-number is at least 10000
To test with GCC,
CC=gcc CXX=g++ cmake ..; ctest --repeat-until-fail <large-number>_

enclave/core/sgx/calls.c
```diff
@@ -236,126 +237,222 @@ oe_result_t oe_register_ecall_function_table(
...
+    state = __atomic_load_n(&(switchless->state), __ATOMIC_ACQUIRE);
+    return state;
+} /* _get_switchless_state */
+
+static void _handle_launch_enclave_worker(uint64_t arg_in)
+{
+    oe_switchless_t* switchless = (oe_switchless_t*)arg_in;
+
+    /* loop until the state changes to stopping */
+    while (OE_SWITCHLESS_STATE_STOPPING != _get_switchless_state(switchless))
```
ihaller: Without pointer validation this loop condition can be used to leak information.

EMumau: Validation added 659cbd3

host/switchless_manager.c
anakrish: The post condition for this method is not clear. It does not guarantee that when this function returns, the state is 'running'.
Say, in line 37, the state is running and thus the control does not go into line 37. Meanwhile another thread calls shutdown and changes the state to stopped.

tests/switchless/host/host.c
anakrish: I'd prefer not to expose this function. Instead you could make one switchless ecall outside time gathering to initialize the system.

host/switchless_manager.c
EMumau: I agree that would be useful.  However, it would be a substantial amount of work to accomplish this.  I'm only using a few functions atomic functions (exchange, compare-exchange, load) and only for uint32_t types.  The Windows and GNU implementations for atomic intrinsic functions are implemented very differently from each other.  I would be nice if Windows supported stdatomic.h.  I do not believe that stdatomic.h is supported in enclaves at this point.
I have  encapsulated the usage of atomic functions into file scope functions to make the functional code more less distracting.

enclave/core/sgx/calls.c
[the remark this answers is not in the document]
EMumau: I have added testing to ensure that the queue (and the switchless_manager) lie outside of enclave memory.  There are also tests to ensure that the marshaled data is outside of the enclave memory.

I need to add tests to ensure that the nodes are outside of enclave memory as they are popped off of the queue.

enclave/core/sgx/calls.c
EMumau: removed this copy.

enclave/core/sgx/calls.c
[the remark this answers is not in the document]
EMumau: I have added testing to ensure that the queue (and the switchless_manager) lie outside of enclave memory. There are also tests to ensure that the marshaled data is outside of the enclave memory.

I need to add tests to ensure that the nodes are outside of enclave memory as they are popped off of the queue.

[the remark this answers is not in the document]
EMumau: I have added a new alternate function to the lockless_queue implementation that tests that nodes are in host memory before any writing is done during the pop operation.  I have added tests for this functionality as wee as negative tests to ensure that nodes created in the enclave memory are properly detected.

enclave/core/sgx/calls.c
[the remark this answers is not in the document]
EMumau: I have added a new alternate function to the lockless_queue implementation that tests that nodes are in host memory before any writing is done during the pop operation. I have added tests for this functionality as wee as negative tests to ensure that nodes created in the enclave memory are properly detected.

tests/switchless/enc/enc.c
radhikaj: @EMumau  Please reference below to see how @anitagov reduced page count for Windows 
https://github.com/openenclave/openenclave/commit/d6794b40b0229483ac8a800ea5b92ddea41fab5a

tests/switchless/host/host.c
radhikaj: @EMumau 
Instead of putting this in host/switchless_manager.h, can you put in an internal header? We would not want applications to call this API.

host/sgx/create.c
radhikaj: @EMumau 
Here is some information from anakrish
https://github.com/openenclave/openenclave/blob/1437683b73ac540d5467687e268f341485251d90/enclave/core/sgx/calls.c#L444-L460

An enclave has an init and a corresponding destructor
At the time of destruction it can make switchless ecalls. 
It is safest to shutdown the switchless manager after the destruction of the enclave.

enclave/core/sgx/calls.c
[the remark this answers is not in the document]
radhikaj: Added this to issue #2037

enclave/core/sgx/calls.c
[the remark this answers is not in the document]
radhikaj: @jxyang  Can this issue be closed?

tests/switchless/enc/enc.c
radhikaj: Added to issue #2037

tests/switchless/host/host.c
radhikaj: Captured in issue #2037

radhikaj:
@EMumau It looks like you have resolved a lot of comments. I have captured other issues in #2037 
Look forward to the generated code from you. 
After that, this pull request can be committed and we can move on to the next one.

enclave/core/sgx/calls.c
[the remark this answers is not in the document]
jxyang: I don't quite get the changes. The alternate function is `oe_lockless_queue_unsafe_pop_front`, right? Why do we need two versions? And seems even though the alternative version has `unsafe` in the name, it's actually safer.

tests/switchless/host/host.c
EMumau: These tests have been added.

tests/switchless/host/switchless_args.h
EMumau: The auto-generated files are now removed.

CodeMonkeyLeet:
Closing this stalled PR. @jxyang will revisit with a new PR in v0.8.

Issue #1794, opened 2019-05-16
InitializeEnclave on Windows occasionally fails due to SGX_UNMASKED_EVENT


CodeMonkeyLeet:
Every so often a test run on Windows will fail during the call to Windows' `InitializeEnclave()` system call with the underlying SGX `EINIT` error `SGX_UNMASKED_EVENT` (0x80). We need to either mitigate this with retry logic, or ensure that `oe_sgx_initialize_enclave()` or the underlying Windows `InitializeEnclave()` methods inhibit the unmasked events.

[The Intel SDM vol 3D](https://software.intel.com/sites/default/files/managed/7c/f1/332831-sdm-vol-3d.pdf) states:

> Periodically, EINIT polls for certain asynchronous events. If such an event is detected, it completes with failure code (ZF=1 and RAX = SGX_UNMASKED_EVENT), and RIP is incremented to point to the next instruction. These events includes external interrupts, non-maskable interrupts, system-management interrupts, machine checks, INIT signals, and the VMX-preemption timer. EINIT does not fail if the pending event is inhibited (e.g., external interrupts could be inhibited due to blocking by MOV SS blocking or by STI).

[OpenEnclave-Bors#865](https://oe-jenkins.eastus.cloudapp.azure.com/blue/organizations/jenkins/Bors/detail/staging/865/pipeline) has a repro of this:

```
test 73
        Start  73: tests/pthread

73: Test command: D:\Jenkins\workspace\Bors_staging\build\tests\thread\host\Release\thread_host.exe "D:/Jenkins/workspace/Bors_staging/build/tests/thread/host/pthread_enc"
73: Test timeout computed to be: 10000000
73: : d:\jenkins\workspace\bors_staging\tests\thread\host\host.cpp(306): error: oe_create_thread_enclave(): result=21
73: 18:09:31:000000 tid(0xbac) (H)[ERROR]InitializeEnclave failed (err=0x80) (oe_result_t=OE_PLATFORM_ERROR)[d:\jenkins\workspace\bors_staging\host\sgx\sgxload.c oe_sgx_initialize_enclave:929]
73: 18:09:31:000000 tid(0xbac) (H)[ERROR]:OE_PLATFORM_ERROR[d:\jenkins\workspace\bors_staging\host\sgx\create.c oe_sgx_build_enclave:577]
[...]

PR #1976, opened 2019-06-20
Retry InitializeEnclave on Windows on SGX EINIT SGX_UNMASKED_EVENT error

anitagov:
Fixes bug (Issue #1794 ) seen with Windows CI where a test fails during the call to Windows' InitializeEnclave() system call with the underlying SGX EINIT error SGX_UNMASKED_EVENT (0x80). Mitigating this with retry logic at the OE SDK level as inhibiting the unmasked events would require Windows kernel support.


anitagov:
Couple of important comments:
1) Did not add any delay or sleep as the context switch to kernel mode should suffice for the delay.
2) Set the MAX_RETRY_COUNT to 50.
3) Did not see any failure with 4 consecutive CI runs. So, would like this to get into master.
We an tweak 1) and/or 2) if needed if we see failures with Windows CI.

This change does not affect Linux.

host/sgx/sgxload.c
```diff
@@ -545,24 +546,37 @@ oe_result_t oe_sgx_create_enclave(
...
+                secs->size,
+                secs->size,
+                ENCLAVE_TYPE_SGX,
+                (const void*)secs,
+                sizeof(ENCLAVE_CREATE_INFO_SGX),
+                &enclave_error);
+
+            if (!base)
+            {
+                if (enclave_error == SGX_UNMASKED_EVENT)
```
jxyang: We are not going to see an error message in the worst case, we failed to create the enclave with after retrying MAX_EINIT_RETRY_COUNT times.

anitagov: Good point. Will incorporate

anitagov:
bors try

host/sgx/sgxload.c
CodeMonkeyLeet: Your fix still doesn't address the comment but instead spews an error message on every retry, which is not meaningful. I'm not sure why this loop needs to be so complicated even.

```c
void* base = NULL;
uint32_t enclave_error;

for (DWORD i = 0; i < MAX_EINIT_RETRY_COUNT; i++)
    base = enclave_create(
        NULL, /* Let OS choose the enclave base address */
        secs->size,
        secs->size,
        ENCLAVE_TYPE_SGX1,
        (const void*)secs,
        sizeof(sgx_secs_t),
        &enclave_error);
        
	if (enclave_error != SGX_UNMASKED_EVENT)
	    break;
}

if (!base)
{
    OE_RAISE_MSG(
        OE_PLATFORM_ERROR,
	"CreateEnclave failed (err=%#x)",
        enclave_error);
}

secs->base = (uint64_t)base;
```

host/sgx/sgxload.c
```diff
@@ -545,24 +546,46 @@ oe_result_t oe_sgx_create_enclave(
...
-            GetCurrentProcess(),
-            NULL, /* Let OS choose the enclave base address */
-            secs->size,
-            secs->size,
-            ENCLAVE_TYPE_SGX,
-            (const void*)secs,
-            sizeof(ENCLAVE_CREATE_INFO_SGX),
-            &enclave_error);
+        DWORD einit_retry_count = 0;
+        while (einit_retry_count < MAX_EINIT_RETRY_COUNT)
```
CodeMonkeyLeet: There should be a comment to indicate why we only do this retry loop on Windows and not Linux. You can lift text from #1794. You should also indicate why we think a sleep between loops is unnecessary (i.e. cost of kernel transition, as stated in the PR comments).

host/sgx/sgxload.c
anitagov: > Your fix still doesn't address the comment but instead spews an error message on every retry, which is not meaningful. I'm not sure why this loop needs to be so complicated even.
 
The error message is printed only when the retry max count is reached. Otherwise, it continues to the while loop:

                    if (einit_retry_count == MAX_EINIT_RETRY_COUNT)
                    {
                        OE_RAISE_MSG(
                            OE_PLATFORM_ERROR,
                            "CreateEnclave failed with SGX_UNMASKED_EVENT "
                            "after %d retries",
                            MAX_EINIT_RETRY_COUNT);
                    }
                    else
                        continue;
             }
OE_RAISE_MSG not only prints the error message but will jump to done and exit the routine.

I prefer for loops instead of while too.

host/sgx/sgxload.c
anitagov: ok, will incorporate.

host/sgx/sgxload.c
anitagov: > Your fix still doesn't address the comment but instead spews an error message on every retry, which is not meaningful. I'm not sure why this loop needs to be so complicated even.
> 
> ```c
> void* base = NULL;
> uint32_t enclave_error;
> 
> for (DWORD i = 0; i < MAX_EINIT_RETRY_COUNT; i++)
> {
>     base = enclave_create(
>         NULL, /* Let OS choose the enclave base address */
>         secs->size,
>         secs->size,
>         ENCLAVE_TYPE_SGX1,
>         (const void*)secs,
>         sizeof(sgx_secs_t),
>         &enclave_error);
>         
>     if (enclave_error != SGX_UNMASKED_EVENT)
>         break;
> }
> 
> if (!base)
> {
>     OE_RAISE_MSG(
>         OE_PLATFORM_ERROR,
> 	"CreateEnclave failed (err=%#x)",
>         enclave_error);
> }
> 
> secs->base = (uint64_t)base;
> ```

@CodeMonkeyLeet - I did not modify the section that you reference. I only modified the 
#elif defined(_WIN32) section with the create_enclave function. This is the section that gets run with the current AESM flow on Windows.

anitagov: We can add the retry if needed to this section once we have support for Windows CI with DCAP and hit this issue. This is the reason that I did not make any changes to this section.

jxyang:
@anitagov The failure happens at `InitializeEnclave`, not `CreateEnclave`. Sorry I may have misled you. It's in the same file, but in a different function `oe_sgx_initialize_enclave`.  I'd write the loop this way:
```c
        int i = 0;
        bool result = false;
        do
        {
            result = InitializeEnclave(
                    GetCurrentProcess(),
                    (LPVOID)addr,
                    &info,
                    sizeof(info),
                    &enclave_error);
        } while (i++ < MAX_EINIT_RETRY_COUNT &&
                 enclave_error == SGX_UNMASKED_EVENT);

        if (!result)
            OE_RAISE_MSG(
                OE_PLATFORM_ERROR,
                "InitializeEnclave failed (err=%#x)",
                enclave_error);
```

anitagov:
@jxyang - Yes, your comments misled me. I would have looked at the failing signature in the issue/failing pipelines to figure out the fix :-)

anitagov:
@jxyang @CodeMonkeyLeet - I fixed InitializeEnclave, the correct path right now. Added comments and have a cleaner flow. Please review.

Currently testing in sandbox mode. I will squash the commits before merge to master.

jxyang:
LGTM

anitagov:
Got 4 clean sandbox runs in a row. Waiting for approval.

host/sgx/sgxload.c
```diff
@@ -918,16 +919,39 @@ oe_result_t oe_sgx_initialize_enclave(
             (void*)&launch_token,
             sizeof(info.EInitToken)));
 
-        if (!InitializeEnclave(
+        bool result = false;
```
CodeMonkeyLeet: Check that compilation on Windows doesn't complain about the implicit cast from BOOL to bool. Probably just easier to use the Windows type. 

You're also shadowing the existing `result` variable in this function, please rename it:

```c
BOOL success = FALSE;
```

anitagov: ok

anitagov:
@yakman2020 - Nice if this gets into Release 0.6 although it only affects Windows (CI).

host/sgx/sgxload.c
```diff
@@ -918,16 +919,39 @@ oe_result_t oe_sgx_initialize_enclave(
             (void*)&launch_token,
             sizeof(info.EInitToken)));
 
-        if (!InitializeEnclave(
+        BOOL success = false;
```
CodeMonkeyLeet: You're still doing an implicit cast via the assignment of `false` to a BOOL type.

anitagov: Changed to FALSE.

Issue #1947, opened 2019-06-14
Add documentation for using Ninja as a build engine


johnkord:
Our getting started page should mention that it's possible to use Ninja as a build engine

yakman2020:
There is a PR on this (#1978) Will it suffice?

yakman2020:
@johnkord Can we close?

johnkord:
Yes once that PR gets merged, we can close this!

yakman2020:
Closed. Fixed by PR 1978

PR #1978, opened 2019-06-20
Add ninja install and usage to getting started docs

yakman2020:
@johnkord does this satisfy your concerns?

### DIFF
--- a/docs/GettingStartedDocs/Contributors/InstallInfo.md
+++ b/docs/GettingStartedDocs/Contributors/InstallInfo.md
@@ -7,8 +7,13 @@ not currently need to be installed system-wide, so you could choose to install i
 into your home directory. From the build subfolder in your source tree:
 
 ```bash
-cmake -DCMAKE_INSTALL_PREFIX=~/openenclave-install ..
+cmake -G "Unix Makefiles" -DCMAKE_INSTALL_PREFIX=~/openenclave-install ..
 make install
+```
+or
+```bash
+cmake -G Ninja  -DCMAKE_INSTALL_PREFIX=~/openenclave-install ..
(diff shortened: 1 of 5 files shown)


docs/GettingStartedDocs/Contributors/InstallInfo.md
```diff
@@ -7,9 +7,14 @@ not currently need to be installed system-wide, so you could choose to install i
 into your home directory. From the build subfolder in your source tree:
 
 ```bash
-cmake -DCMAKE_INSTALL_PREFIX=~/openenclave-install ..
+cmake -G "Unix Makefiles" -DCMAKE_INSTALL_PREFIX=~/openenclave-install ..
```
johnkord: Do you think we need to specify the "Unix Makefiles"?  I think by default it uses make in my experience, but I don't really know. @andschwa what do you think?

andschwa: No, this is the default on Linux.

yakman2020: @andschwa , adding "Unix Makefiles" is deliberate.  I'm trying to make a distinction between that and Ninja.  Besides, Unix Makefiles is not the default on windows (VC project is, I think) and that adds an additional degree of confusion.

yakman2020: @andschwa Could I get an approval or a change request?

jxyang: Yes, on windows the default target is Visual Studio

yakman2020:
fixes #1947

PR #1985, opened 2019-06-21
Attested TLS sample: fixed mrsigner error from using wrong header file when mixing CMake and Makefile building environment

soccerGB:
The changes in this PR is to address the following issues reported when building this sample in the repo tree.

All of our samples support both CMake and Makefile building methods

The build issue reported by anand only repos if building the same sample  in the following sequence. :  

1.Build the sample via CMake
2.Build the sample via Makefile
3.Then build the sample with CMake again— hit this issue

This was because both build system share the  same output directory, some of the CMake targets didn’t get run again in step 3 because output from step 2 confuses CMake. This explains how anand and Anita repros it.
This is a build issue not code code.

If a user installs the SDK sample builds only with Makefile or CMake, it works fine , no matter how many times it’s built. So, It wouldn’t impact normal SDK users, which we do not expect mixing both building methods like our building system does (ctest -V build all sample in CMake followed by Makefile)


soccerGB:
bors try

PR #1987, opened 2019-06-21
Fixes a crash in oedump

jxyang:
This fixes an oesign crash found during fuzzing.


jxyang:
I think you both are talking the same thing. I will take the advice and submit a new patch.

yakman2020:
Need to add this to master also

PR #1991, opened 2019-06-24
Attested TLS sample: added MRENCLAVE checking into this sample 

soccerGB:
Part of this change is to replace dynamically generated key pairs with static ones


soccerGB:
bors try

samples/attested_tls/non_enc_client/verify_callback.cpp
```diff
@@ -40,13 +41,27 @@ oe_result_t enclave_identity_verifier(oe_identity_t* identity, void* arg)
...
 
-    printf(TLS_CLIENT "identity->unique_id : ");
+    // the unique ID for the enclave, for SGX enclaves, this is the MRENCLAVE
+    // value
+    printf(TLS_CLIENT "Validating identity->unique_id(MRENCLAVE) :\n");
     for (int i = 0; i < OE_UNIQUE_ID_SIZE; i++)
+    {
         printf("0x%0x ", (uint8_t)identity->unique_id[i]);
+        if (SERVER_ENCLAVE_MRENCLAVE[i] != (uint8_t)identity->unique_id[i])
+        {
```
radhikaj: will you set an error here?

samples/attested_tls/scripts/gen_mrenclave_header.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
radhikaj: Can you please help me understand why this needs to be in a script?

samples/attested_tls/server/enc/identity_verifier.cpp
```diff
@@ -23,17 +23,17 @@ oe_result_t enclave_identity_verifier_callback(
...
         identity->security_version);
 
-    // the unique ID for the enclave, for SGX enclaves, this is the MRENCLAVE
-    // value
+    // the unique ID for the enclave, for SGX enclaves, this is client's
+    // MRENCLAVE
     printf(TLS_SERVER "identity->unique_id(MRENCLAVE) :\n");
     for (int i = 0; i < OE_UNIQUE_ID_SIZE; i++)
-        printf(TLS_SERVER "0x%0x ", (uint8_t)identity->unique_id[i]);
+        printf("0x%0x ", (uint8_t)identity->unique_id[i]);
```
radhikaj: But we are not doing a check to see that it is a known value, is that correct

samples/attested_tls/non_enc_client/verify_callback.cpp
soccerGB: the default result is OE_VERIFY_FAILED, so that should cover it.

samples/attested_tls/scripts/gen_mrenclave_header.sh
soccerGB: That's the only to make bash recognize the containing file is a bash script file
(Each script starts with a "shebang" and the path to the shell that you want the script to use)

samples/attested_tls/server/enc/identity_verifier.cpp
soccerGB: We are not checking for client enclave's MRENCLAVE because a TLS server enclave was built before an TLS client enclave. So, checking MRSIGNER is the most we could do unless we want to provision allowed client MRENCLAVE values after a server is running.

soccerGB:
The justification for marking this PR as 0.6.0 is there will be no real application that would want to dynamically generate key pair signing purpose, which creates so much trouble when an enclave developer want to adapt our sample into their own app. Also, doing so will make it easier to demonstrate on how to check MRENCLAVE in the sample, especially when building apps for both TLS client and server for running on different systems.

PR #1995, opened 2019-06-24
Correctly assert aesni instruction set.

anakrish:
- Avoid TOCTOU
- Make function strongly typed.

### DIFF
--- a/enclave/core/sgx/calls.c
+++ b/enclave/core/sgx/calls.c
@@ -142,6 +142,7 @@ static oe_result_t _handle_init_enclave(uint64_t arg_in)
     OE_ATOMIC_MEMORY_BARRIER_ACQUIRE();
     if (o == false)
     {
+        // Assert that arg_in is outside the enclave and is not null.
         if (!oe_is_outside_enclave(
                 (void*)arg_in, sizeof(oe_init_enclave_args_t)))
         {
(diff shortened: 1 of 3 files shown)


anakrish:
bors try

yakman2020:
bors try

enclave/core/sgx/cpuid.c
```diff
@@ -20,27 +20,26 @@ static uint32_t _cpuid_table[OE_CPUID_LEAF_COUNT][OE_CPUID_REG_COUNT];
...
 
-        OE_CHECK(oe_memcpy_s(
-            _cpuid_table,
-            OE_CPUID_LEAF_COUNT * OE_CPUID_REG_COUNT *
-                sizeof(_cpuid_table[0][0]),
-            args->cpuid_table,
-            OE_CPUID_LEAF_COUNT * OE_CPUID_REG_COUNT *
-                sizeof(args->cpuid_table[0][0])));
+    if (args == NULL || !oe_is_within_enclave(args, sizeof(*args)))
+        OE_RAISE(OE_FAILURE);
```
CodeMonkeyLeet: Use OE_INVALID_PARAMETER instead of the more generic OE_FAILURE

enclave/core/sgx/cpuid.c
```diff
@@ -20,27 +20,26 @@ static uint32_t _cpuid_table[OE_CPUID_LEAF_COUNT][OE_CPUID_REG_COUNT];
...
-            oe_abort();
 
-        OE_CHECK(oe_memcpy_s(
-            _cpuid_table,
-            OE_CPUID_LEAF_COUNT * OE_CPUID_REG_COUNT *
-                sizeof(_cpuid_table[0][0]),
-            args->cpuid_table,
-            OE_CPUID_LEAF_COUNT * OE_CPUID_REG_COUNT *
-                sizeof(args->cpuid_table[0][0])));
+    if (args == NULL || !oe_is_within_enclave(args, sizeof(*args)))
```
CodeMonkeyLeet: Side note: The null check on `args` is only necessary because of the `sizeof(*args)` construction here (`oe_is_within_enclave` already rejects null pointers) and it looks like `_handle_init_enclave` gets around this by using `sizeof(oe_init_enclave_args_t)` to elide the double check.

We should probably consider standardizing at some point.

yakman2020:
bors try passed. Let me know when you want bors r+
________________________________

enclave/core/sgx/calls.c
```diff
@@ -154,25 +155,22 @@ static oe_result_t _handle_init_enclave(uint64_t arg_in)
         if (_once == false)
         {
             /* Set the global enclave handle */
-            if (arg_in)
```
jxyang: Why we don't need to check whether `arg_in` is NULL anymore?

anakrish: @jxyang See my comment in line 145. oe_is_outside_enclave already checks for NULL parameters.

enclave/core/sgx/cpuid.c
anakrish: @CodeMonkeyLeet 
Though redundant, I added the pre-conditions to this function to keep this function safe from future modifications to the caller.

Note: In C/C++, the arguments to sizeof expressions is never evaluated. You could do sizeof(printf("Hello, World\n")); the language guarantees that the printf is used only for inferring the type. So there is no need to guard sizeof(*args) with a NULL check.

I personally prefer sizeof(expression) where possible instead of sizeof(type) since that is robust to expressions changing type in the future.

anakrish:
bors try

jxyang:
LGTM

johnkord:
I'm not too sure why you got a "Permission denied" @anakrish ! It shows you as a "reviewer" on the bors server, which gives you both "r+" and "try" permissions (and all other commands).  I can look into the logs tomorrow for you, if you're curious. My ONLY guess about this so far is that I don't see your @anakrish user on GitHub being in the Microsoft GitHub organization. Is your user in the org? Maybe I'm looking in the wrong place.

PR #2014, opened 2019-06-27
Use GCC (GNU) atomic builtins for acquire-release

anakrish:
This will emit the right instructions for all platforms.


anakrish:
bors try

anakrish:
> We also need definitions for these 2 commonly used memory orders: __ATOMIC_ACQ_REL and __ATOMIC_SEQ_CST

These are GCC/Clang builtins. We don't need to redefine them..

andschwa:
@anakrish This commit was authored by "Ubuntu"...

PR #2017, opened 2019-06-28
Enclave lifecycle and ECALL/OCALL support for AArch64

HernanGatta:
This is a follow-up to preview PR #1892, this time against `master`.

This PR contains:

1. Implementation of `oe_create_enclave`;
2. Implementation of `oe_terminate_enclave`;
3. Implementation of `oe_is_[within|outside]_enclave`;
4. Implementation of `oe_abort`;
5. ECALL support;
6. OCALL support;
7. Sharing of additional code between SGX and OP-TEE support.

This PR uses the following changes to OP-TEE OS and OP-TEE Client, in their respective repositories:

1. OP-TEE OS:
1.1. The Remote Procedure Call Pseudo-TA (RPC PTA);
1.2. The RPC PTA implements OCALL support in OP-TEE OS and exposes an interface to user-mode TAs;
1.3. [Diff](https://github.com/ms-iot/optee_os/compare/dd4c992df826610e3bfdd756cc7a1685e7b0495b...5461f173a5a4846396ccca3e866e1bea90c4e8f9#diff-ee061305f8535978e87ae5ebed0a7c8b).
2. OP-TEE Client:
2.1. The host-side RPC interface to the RPC PTA;
2.2. It receives RPC requests from the RPC PTA and lets TA hosts handle them;
2.3. [Diff](https://github.com/ms-iot/optee_client/compare/ms-optee-ocalls...ms-iot-openenclave).


yakman2020:
@mikebras pls review.

HernanGatta:
@mikbras @anakrish @andschwa 

When anyone gets a chance, do please take a look before this branch falls too far behind `master`; those merges sometimes get tricky.

enclave/core/sgx/linux/threadlocal.c
```diff
@@ -201,8 +203,8 @@ static volatile uint64_t _tbss_align = 1;
 static volatile bool _thread_locals_relocated = false;
 
 /* Thread local variables to track functions to call on thread exit */
-static __thread oe_tls_atexit_t* _tls_atexit_functions;
-static __thread uint64_t _num_tls_atexit_functions;
+static OE_THREAD_LOCAL_STORAGE oe_tls_atexit_t* _tls_atexit_functions;
```
anakrish: Why do we need to change SGX specific thread-local implementation if TrustZone does not support TLS?

include/openenclave/internal/calls.h
```diff
@@ -473,7 +480,13 @@ oe_result_t oe_ecall(
  * @retval OE_UNEXPECTED An unexpected error occurred.
  *
  */
-oe_result_t oe_ocall(uint16_t func, uint64_t arg_in, uint64_t* arg_out);
+oe_result_t oe_ocall(
```
anakrish: What is the rationale for this new signature?

enclave/core/sgx/linux/threadlocal.c
HernanGatta: Technically we don't. I figured that since I changed `__thread` to `OE_THREAD_LOCAL_STORAGE` everywhere else, for consistency I'd so here too.

Having said that, I see now that it could make the reader think that TLS is sometimes not supported or could be disabled on SGX, which is certainly not the case.

What do you think? I'll happily revert this change.

include/openenclave/internal/calls.h
HernanGatta: The SDK as is currently written passes OCALL parameters for built-in functions (e.g., ) in pointers via host memory.

Take `oe_host_malloc` as an example. This function performs an OCALL whose `IN` parameter represents the memory size to allocate, and an `OUT` parameter that is the host-side pointer. In SGX, passing these parameters does not require any specialized marshaling (i.e., a parameters structure), the host can read and write the parameters via registers.

Consider now `oe_host_realloc`. The OCALL this function performs takes in a pointer to structure as its `IN` parameter and its `OUT` parameter is a pointer. In SGX, the `IN` pointer refers to a location in host memory (previously allocated), and is passed by value in a register. The `OUT` parameter pointer works in the same way.

So the first invocation to `oe_ocall` passes a value to the host (`IN`), and a pointer to the enclave (`OUT`). The second invocation passes a pointer as a value to the host (`IN`), and a pointer to the enclave (`OUT`).

This discrepancy in semantics is irrelevant in SGX because the enclave can access host memory and the OE SDK has direct control over the registers used for parameter passing.

In TrustZone, those assumptions don't hold. Both values and pointers must be marshaled, wrapped in proper descriptor structures, much like `oeedger8r` does. Passing a value (the size in `oe_malloc`) means the SDK must know to tell the TEE API that the value in question is a value. If it is a pointer, passing the pointer as a value won't work, because the pointer points to enclave memory, and allocating and filling a host buffer ahead of time, like in the case of `oe_host_realloc` also doesn't work because the enclave cannot access host memory. So the pointer cannot be passed as is, and the SDK must know that it's a pointer to something of a known size, then marshal the entire structure.

enclave/core/sgx/linux/threadlocal.c
anakrish: I'd prefer reverting this change.

include/openenclave/internal/calls.h
anakrish: Thanks for the explanation. I now see why we need a signature similar to 
https://github.com/openenclave/openenclave/blob/87397fdb3948b3a60e2195cf9f28fa1fad8bc111/include/openenclave/edger8r/enclave.h#L68-L74

However is seems that the boolean parameter is redundant. The function-enum (OE_OCALL_MALLOC, OE_OCALL_REALLOC) itself indicates whether to interpret the value as a pointer or not.

Note @mikbras is working on moving most of the core hand-written functions over to EDL.

include/openenclave/bits/defs.h
```diff
@@ -149,4 +149,16 @@
...
 #endif
 
+/* OP-TEE provides single-threaded enclaves only, and its ELF loader does not
+ * support thread-local relocations. Hence, any enclave that includes a
+ * variable marked with __thread will not only not work, but it will fail to
+ * load altogether.
+ */
+#if defined(_ARM_) || defined(_M_ARM) || defined(__arm__) || \
+    defined(__thumb__) || defined(__aarch64__)
+#define OE_THREAD_LOCAL_STORAGE
```
anakrish: I think my previous comment got lost.
Can we instead just define __thread to empty string for ARM? That would avoid editing 3rd party code.

enclave/core/sgx/linux/threadlocal.c
HernanGatta: Will do.

include/openenclave/internal/calls.h
HernanGatta: On the redundancy question, I agree; however, `_handle_call_builtin_function` would have to deduce that information for all hand-written `OE_OCALL_*`. I'd rather the caller tell `oe_ocall` if the `IN` value is a value or a pointer, instead of encoding that information effectively twice.

Perhaps I'm missing something and you had a different idea, though.

As to your note, that'd be nice!

include/openenclave/bits/defs.h
HernanGatta: Hadn't thought of that! I'll try it.

enclave/core/calls.c
```diff
@@ -0,0 +1,43 @@
...
+    const oe_ecall_func_t* ecalls,
+    size_t num_ecalls)
+{
+    oe_result_t result = OE_UNEXPECTED;
+
+    if (table_id >= OE_MAX_ECALL_TABLES || !ecalls)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+    oe_spin_lock(&_ecall_tables_lock);
+    _ecall_tables[table_id].ecalls = ecalls;
```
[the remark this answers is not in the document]
HernanGatta: As it is, reassigning the table is not an error, per se. The code for `oe_create_enclave` for SGX does not put the table initialization in a `once` block. For each call, it installs the tables again.

Perhaps we can put the table installation into a `once` block in a different PR?

include/openenclave/bits/defs.h
HernanGatta: Removed `OE_THREAD_LOCAL_STORAGE` and did as you suggested. Thanks for the idea!

include/openenclave/internal/calls.h
HernanGatta: Thinking more about this, I'm leaning toward leaving this as-is for now, and waiting to see what Mike comes up with, then modifying this as required.

Thoughts?

host/CMakeLists.txt
```diff
@@ -137,8 +136,15 @@ if (OE_SGX)
...
 elseif(OE_TRUSTZONE)
   list(APPEND PLATFORM_SRC
-    optee/enclave.c
     optee/log.c)
+  
+  if (UNIX)
+    list(APPEND PLATFORM_SRC
+      optee/linux/enclave.c)
+  else()
+    message(STATUS "Add Windows OP-TEE files.")
```
andschwa: My only suggestion would be to swap this to an error.

enclave/core/calls.c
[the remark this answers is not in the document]
anakrish: Sounds good.

include/openenclave/internal/calls.h
anakrish: It would be good to keep oe_ecall and oe_ocall consistent.
Currently oe_ecall has the same signature as oe_ocall; but with an additional enclave parameter to indicate which enclave to invoke the method on:
https://github.com/openenclave/openenclave/blob/3dc3c9ff9e3336912cc88237f220f0326c6ce2e5/include/openenclave/internal/calls.h#L431-L435

HernanGatta: This is in fact a bug; the way I implemented `oe_ecall` only works for `oeedger8r` ECALL functions; all the others won't.

I'll fix it.

host/CMakeLists.txt
HernanGatta: Will do.

include/openenclave/bits/defs.h
```diff
@@ -149,4 +149,14 @@
 #define OE_DEPRECATED(FUNC, MSG) FUNC
 #endif
 
+/* OP-TEE provides single-threaded enclaves only, and its ELF loader does not
+ * support thread-local relocations. Hence, any enclave that includes a
+ * variable marked with __thread will not only not work, but it will fail to
+ * load altogether.
+ */
+#if defined(_ARM_) || defined(_M_ARM) || defined(__arm__) || \
```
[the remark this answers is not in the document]
HernanGatta: Good point, that was on oversight on my part. I'll look into isolating the redefinition of `__thread` to the enclave only.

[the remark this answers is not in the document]
HernanGatta: I've moved the redefinition to `enclave.h`. I had to have `enclave/core/errno.c` include `enclave.h`, which seems fine since it's enclave-only code.

include/openenclave/internal/calls.h
HernanGatta: Done.

enclave/core/CMakeLists.txt
```diff
@@ -3,24 +3,33 @@
...
 
-    set(EDL_DIR ${CMAKE_SOURCE_DIR}/common)
-    set(EDL_FILE ${EDL_DIR}/sgx/internal.edl)
+set(TEE_EDL_FILE ${EDL_DIR}/tee.edl)
 
-    add_custom_command(
-        OUTPUT internal_t.h internal_t.c internal_args.h
-        DEPENDS ${EDL_FILE} edger8r
-        COMMAND edger8r --search-path ${EDL_DIR} --trusted ${EDL_FILE})
+add_custom_command(
```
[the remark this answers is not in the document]
HernanGatta: Oh, perhaps the name isn't all that descriptive. What I meant by TEE is any trusted execution environment, SGX, TrustZone, what have you. So `tee.edl` is always processed regardless of the underlying hardware, and there is platform-specific functionality, such as in `sgx.edl`. There currently isn't a need for `trustzone.edl`, but there might be down the line.

Does that make sense?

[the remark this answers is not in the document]
anakrish: Sounds good.

HernanGatta:
bors try

HernanGatta:
bors try

mikbras:
Looks great Hernan. Thanks for all you hard work.

PR #2026, opened 2019-07-03
Bump LLVM libcxx to 8.0.0

aserdean:
Fixes part LLVM libcxx 8.0.0 of #1456

PR #2028, opened 2019-07-05
[CMAKE] Replace special characters in test-case name

aserdean:
During the bump from LLVM v7 to v8 a new test with a special was encountered:
https://github.com/openenclave/openenclave/pull/2026/commits/a457286615b53e8e2e1ef653fa98c492fd9de3db

This patch replaces "[" and "]" with "__".

Signed-off-by: Alin Gabriel Serdean <aserdean@ovn.org>


aserdean:
CC: @andschwa , @CodeMonkeyLeet

PR #2029, opened 2019-07-05
Update Linux docs and ansible to use new Intel SGX DCAP 1.2 driver.

johnkord:
This doesn't include updates to the Windows docs/ansible/script. I wasn't able to find the new DCAP release on Intel's SGX download page for Windows. Is that released to a different location?

After this merges into master, we should merge it to the v0.6.x release branch.

### DIFF
--- a/docs/GettingStartedDocs/Contributors/BuildingInADockerContainer.md
+++ b/docs/GettingStartedDocs/Contributors/BuildingInADockerContainer.md
@@ -23,6 +23,9 @@
 ```bash
 sudo docker run --device /dev/sgx:/dev/sgx -i -t oeciteam/oetools-full-18.04 bash
 ```
+  - If you're using the Intel SGX (non-DCAP) driver, you'll want to do two things:
+    - The device name is different (isgx as opposed to sgx), so you'll want this option instead:  `--device /dev/isgx:/dev/isgx`
+    - The aesm service will need to be running on the container host, and then its socket file directory will need to be exposed to the container as a volume by adding the `-v /var/run/aesmd:/var/run/aesmd` option.
 
 5. Clone the Open Enclave repository from within this container and run the build and tests. For example, if your system has the SGX DCAP driver installed and it has been made available to the container:
 ```bash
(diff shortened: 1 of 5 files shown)


docs/GettingStartedDocs/Contributors/BuildingInADockerContainer.md
```diff
@@ -9,11 +9,11 @@
 1. Install Docker CE by following these instructions: https://docs.docker.com/install/linux/docker-ce/ubuntu/#install-docker-ce
 
 2. Install the appropriate Intel SGX Driver for your platform. For example, if you're running on an SGX1 with FLC system, you'll probably want to install the Intel SGX DCAP Driver for your platfrom: https://01.org/intel-software-guard-extensions/downloads
-    - If you're running on an SGX with FLC system, you'll probably want to install the Intel SGX DCAP Driver for your platfrom: https://01.org/intel-softwareguard-extensions/downloads/intel-sgx-dcap-linux-1.1-release
+    - If you're running on an SGX with FLC system, you'll probably want to install the Intel SGX DCAP Driver for your platfrom: https://01.org/intel-softwareguard-extensions/downloads/intel-sgx-dcap-linux-1.2-release
     - If you're running on an SGX without FLC system, you'll want to install the Intel SGX Driver for your platform: https://01.org/intel-softwareguard-extensions/downloads/intel-sgx-linux-2.5-release
```
CodeMonkeyLeet: This should be bumped up to the 2.6 release ... we should also test that this works correctly in the manual scenario since our AESM prereqs script still pulls from the latest and we don't automate testing around that now.

johnkord: Thanks Simon, good advice.  I'll test that out right now on my local SGX enabled VM

johnkord: I've updated it accordingly! In fact, I had to add some more information to describe how to get the AESM scenario working in a container :)

Also, thanks for reminding me about that prereqs directory. We need to figure out what to do with that eventually. I'm going to file an issue for it

PR #2031, opened 2019-07-09
Edger8r changes for switchless ecalls and ocalls.

anakrish:
- Minimal change in generated code. For switchless ecalls, oe_switchless_call_enclave_function is called instead of oe_call_enclave_function. Similarly for switchless ocalls, oe_switchless_call_host_function is called.
- Switchless (transition_using_threads keyword) is supported only if --experimental option is passed in.
- Placeholder implementation. Always returns OE_UNSUPPORTED
- Behavior locked down with test.
- For switchless ocalls, oe_allocate_switchless_ocall_buffer and oe_free_switchless_ocall_buffer functions are used to allocate and free memory for making ocalls. These functions can use a host memory pool to allocate memory. Alternatively, these functions can allocate buffers in enclave memory and then use a ring buffer to transfer contents to the host.

### DIFF
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -870,6 +870,61 @@ done:
 /*
 **==============================================================================
 **
+** oe_switchless_call_enclave_function_by_table_id()
+**
+** Switchlessly call the enclave function specified by the given table-id and
+*function-id.
+**
+**==============================================================================
+*/
+
+static oe_result_t oe_switchless_call_enclave_function_by_table_id(
+    oe_enclave_t* enclave,
+    uint64_t table_id,
+    uint64_t function_id,
+    const void* input_buffer,
+    size_t input_buffer_size,
+    void* output_buffer,
+    size_t output_buffer_size,
+    size_t* output_bytes_written)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    oe_call_enclave_function_args_t args;
+
+    /* Reject invalid parameters */
+    if (!enclave)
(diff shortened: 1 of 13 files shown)


host/sgx/calls.c
```diff
@@ -898,6 +953,36 @@ oe_result_t oe_call_enclave_function(
...
+    uint32_t function_id,
+    const void* input_buffer,
+    size_t input_buffer_size,
+    void* output_buffer,
+    size_t output_buffer_size,
+    size_t* output_bytes_written)
+{
+    return oe_switchless_call_enclave_function_by_table_id(
+        enclave,
+        OE_UINT64_MAX,
```
radhikaj: For my understanding, Why is the table id OE_UINT64_MAX

anakrish: This is the convention used to identify the table generated by oeedger8r from an user EDL.

There are other internal tables (for sockets, file systems etc) that have predefined ids.

anakrish:
@radhikaj @jxyang Can you review and accept this PR? This implements the edger8r changes for both ecalls and ocalls. This will allow Xuejun to focus on the core SDK work and to fill in the placeholders.

anakrish:
bors try

host/sgx/calls.c
jxyang: `OE_UINT64_MAX` means we are searching all enclave function tables.

anakrish:
@andschwa The edger8r changes are very minimal and I am merging it to master. Kindly help out @jxyang if  tweaks are needed.

PR #2033, opened 2019-07-09
Debugger Contract: Notifications

anakrish:
- Send Enclave Creation/Termination notifications on Windows
- oe_debugger_contract version is initialized to 1 or the value
  of OE_DEBUGGER_CONTRACT_VERSION environment variable
  If version >= 1, then notifications are sent to the debugger.

### DIFF
--- a/debugger/debugrt/host/host.c
+++ b/debugger/debugrt/host/host.c
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 #include <openenclave/internal/debugrt/host.h>
+#include <stdio.h>
 #include <string.h>
 
 /**
@@ -14,7 +15,7 @@
 
 #include <Windows.h>
 
-static uint64_t _lock = 0;
+static volatile LONG _lock = 0;
 
 static void spin_lock()
 {
@@ -28,6 +29,63 @@ static void spin_lock()
 static void spin_unlock()
 {
     InterlockedExchange(&_lock, 0);
+}
+
+static bool raise_debugger_events()
+{
+    static bool initialized = false;
+
+    if (IsDebuggerPresent())
+    {
+        if (!initialized)
+        {
+            // Set oe_debugger_contract_version from the environment.
+            // Normally the debugger would set the value depending on
+            // the version it supports. Setting it from the environment
+            // allows using a specific version of the contract.
+            char* version = getenv("OE_DEBUGGER_CONTRACT_VERSION");
+            if (version != NULL)
+            {
+                int v = 0;
+                sscanf(version, "%d", &v);
+                oe_debugger_contract_version = (uint32_t)v;
+            }
+
+            initialized = true;
+        }
+        // Events are raised only if the contract is valid.
+        return (oe_debugger_contract_version >= 1);
+    }
+    else
+    {
+        return false;
+    }
+}
+
+void oe_notify_debugger_enclave_creation(const oe_debug_enclave_t* enclave)
+{
+    if (raise_debugger_events())
+    {
+        ULONG_PTR args[1] = {(ULONG_PTR)enclave};
+        RaiseException(
+            OE_DEBUGRT_ENCLAVE_CREATED_EVENT,
+            0, // dwFlags
+            1, // number of args
+            args);
+    }
+}
+
+void oe_notify_debugger_enclave_termination(const oe_debug_enclave_t* enclave)
+{
+    if (raise_debugger_events())
+    {
+        ULONG_PTR args[1] = {(ULONG_PTR)enclave};
+        RaiseException(
+            OE_DEBUGRT_ENCLAVE_TERMINATED_EVENT,
+            0, // dwFlags
+            1, // number of args
+            args);
+    }
 }
 
 #elif defined __GNUC__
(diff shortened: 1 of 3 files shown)


anakrish:
bors try

anakrish:
bors try

debugger/debugrt/host/host.c
```diff
@@ -30,6 +31,85 @@ static void spin_unlock()
...
+{
+    static bool initialized = false;
+
+    if (IsDebuggerPresent())
+    {
+        if (!initialized)
+        {
+            // Set oe_debugger_contract_version from the environment.
+            // Normally the debugger would set the value depending on
+            // the version it supports. Setting it from the environment
```
CodeMonkeyLeet: For my understanding, what does it mean to use a specific version of the contract? There's no example of how the debugrt negotiates whether it supports a given contract version specified by either the debugger or the environment setting. What does the "contract version" cover, for instance? Just the set of events fired but not the globals or data structures? 

Depending on how that's scoped, it seems like debugrt must only support the versions it understands (i.e. exactly `oe_debugger_contract_version == 1`). For example, if debugger uses contract version 2 because it expects a new global to be present for it to manipulate but it does not, should it fire the debug events? Or do you envision that this is piecemeal (i.e. firing these events will always continue to work for any version >= 1 and other features light up by version?

include/openenclave/internal/debugrt/host.h
```diff
@@ -56,11 +56,77 @@ OE_STATIC_ASSERT(OE_OFFSETOF(oe_debug_enclave_t, tcs_array) == 56);
...
+ * Upon attaching to an application, the debugger can scan this list
+ * and configure the enclaves for debugging.
+ */
 OE_EXPORT extern oe_debug_enclave_t* oe_debug_enclaves_list;
 
+////////////////////////////////////////////////////////////////////////////////
+/////////////// Events Raised for Windows Debuggers/////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+
+#if defined(WIN32) || defined(_WIN32)
```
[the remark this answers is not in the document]
anakrish: I don't remember where i picked this from.
Fixed to use `defined (_WIN32)`

debugger/pythonExtension/gdb_sgx_plugin.py
```diff
@@ -243,7 +243,7 @@ def update_untrusted_ocall_frame(frame_pointer, ocallcontext_tuple):
 
 class EnclaveCreationBreakpoint(gdb.Breakpoint):
     def __init__(self):
-        gdb.Breakpoint.__init__ (self, spec="oe_notify_gdb_enclave_creation", internal=1)
+        gdb.Breakpoint.__init__ (self, spec="oe_notify_debugger_enclave_creation", internal=1)
```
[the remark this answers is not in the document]
anakrish: Good catch. Fixed.

debugger/debugrt/host/host.c
anakrish: Thanks for the discussion. I have simplified the use of the variable.
The variable denotes the version of the contract supported by the SDK.
The debugger can introspect this variable to determine if it can debug an application; if it cannot, it is expected to advise the user to use a newer version of the debugger.

anakrish:
bors try

anakrish:
bors try

debugger/debugrt/host/host.c
```diff
@@ -30,6 +31,83 @@ static void spin_unlock()
...
+    {
+        if (!initialized)
+        {
+            // If specified, override oe_debugger_contract_version from the
+            // environment.
+            char* version = getenv("OE_DEBUGGER_CONTRACT_VERSION");
+            if (version != NULL)
+            {
+                int v = 0;
+                sscanf(version, "%d", &v);
```
jxyang: Check the return value of sscanf?

debugger/debugrt/host/host.c
```diff
@@ -77,6 +155,13 @@ OE_NO_OPTIMIZE_END
 
 #endif
 
+/**
+ * The version of the debugger contract supported by the runtime.
+ * For development purposes, this value can be overridden by setting
+ * the OE_DEBUGGER_CONTRACT_VERSION enviroment variable.
+ */
+uint32_t oe_debugger_contract_version = 1;
```
jxyang: This is different from the PR description, which says the version is 0 by default.

jxyang: If we have backward compatibility, is versioning still necessary?

debugger/debugrt/host/host.c
anakrish: Fixed. Added check to ensure that it is 1 before using the value.

debugger/debugrt/host/host.c
anakrish: > This is different from the PR description, which says the version is 0 by default.

Fixed the outdated PR description. Good catch.

anakrish: > If we have backward compatibility, is versioning still necessary?

Yes. This will allow an older debugger to determine that it cannot completely handle an application that supports a newer version of the contract.

Say in contract v2 we support SGX 2 which allows creating TCS dynamically.
A debugger that understands only contract v1 can detect this and inform the user to use a newer version of the debugger. The older debugger may still be able to partially debug the application.

anakrish:
bors try

PR #2043, opened 2019-07-18
Bump Mbed TLS to 2.16.1 LTS

### LINKED ISSUE #1457: Evaluate moving to mbedTLS v2.16 branch
mbedTLS has designated its latest update [v2.16.1](https://tls.mbed.org/download/start/mbedtls-2.16.1-apache.tgz) as of 2018-12-21 the new long-term stable production branch. We should evaluate the impact of moving to this new release:

- [ ] Reconcile breaking changes in the API surface with OE SDK use
- [ ] Evaluate whether this poses issues for link-time mbedTLS replacement for projects using mbedTLS updated with Everest TLS crypto.
- [ ] Evaluate cost-benefit of moving to 2.16.x
- [x] Determine if 2.7 branch is being EOL and timeframe for that.
- [ ] Request OSS SDL security review of new mbedTLS version.

EmilAlexandruStoica:
Fixes #1457 by upgrading Mbed TLS to 2.16 LTS branch


johnkord:
bors try

johnkord:
@EmilAlexandruStoica I'm going to look at the reasons a few of the OE tests fail that rely on mbedtls. I imagine we may be making some assumptions about mbedtls that have changed in the new version of mbedtls!

johnkord:
I've figured out why the tests fail, it's because there have been modifications to the mbedtls source to support Cheng-mean's attested TLS! I just ported those custom changes to your branch, but I don't have write permissions :) I've attached the patch here:

[emil_mbedtls.txt](https://github.com/openenclave/openenclave/files/3448396/emil_mbedtls.txt)

Making source code modifications of 3rdparty libraries in place is a bad idea in the long term (and unfortunately that's something we have to live with for now). When we eventually factor our 3rdparty source dependencies out and use submodules (for mbedtls and others), we should (if necessary) have a patch step that applies a patch on the source dependency.

@EmilAlexandruStoica Can you apply this patch to your branch and then add a commit with that patch on top of this PR? All tests (and bors/jenkins) should hopefully pass in that case!

EmilAlexandruStoica:
Thank you for checking this. I will change tomorrow before the meeting and run the CI to see if the tests pass.

johnkord:
@EmilAlexandruStoica After a bunch of really quite annoying debugging (mbedtls does some bizarre stuff with generating their tests before compiling them!), I figured out the main problem. We were forcing the mbedtls test to use the .data files, when in the new version it's expecting a .datax file apparently, which have types embedded before the values for each data element.

Also, there wasn't a handler for the lseek syscall, but that was a trivial thing to disagnose.

Anyways, try this out! I've attached another patch. This enables all the disabled tests and fixes the issues I described above. Can you try running it through Jenkins? :)

[2019-08-01_patch.txt](https://github.com/openenclave/openenclave/files/3459058/2019-08-01_patch.txt)

EmilAlexandruStoica:
Sure. I will do the changes today and see how it goes. Thanks again for checking this.

johnkord:
@EmilAlexandruStoica Looks like Windows fails with mbed tests because we reference some python scripts, and we don't install python as a dependency on Windows!

CodeMonkeyLeet:
Superseded by #2184. The changes here have been ported and updated into that PR.

PR #2045, opened 2019-07-18
Part II: Replace hand-written edge routines with EDL

mikbras:
This PR replaces all non-trivial hand-written edge routines with EDL. The exceptions are calls that EDL uses in its implementation, including:
-  **OE_ECALL_CALL_ENCLAVE_FUNCTION**
-  **OE_OCALL_CALL_HOST_FUNCTION**
- **OE_OCALL_MALLOC**
- **OE_OCALL_FREE**

So called "trivial" hand-written edge routines include calls that take a 64-bit input value and/or a 64-bit output value (where the values are not a complex types). These include:
- **OE_ECALL_DESTRUCTOR** - no arguments
- **OE_ECALL_INIT_ENCLAVE** - input(enclave handle), output(none)
- **OE_ECALL_VIRTUAL_EXCEPTION_HANDLER** - input(none), output(return value)
- **OE_OCALL_THREAD_WAKE** - input(TCS), output(none)
- **OE_OCALL_THREAD_WAIT** - input(TCS), output(none)
- **OE_OCALL_SLEEP** - input(milliseconds), output(none)
- **OE_OCALL_GET_TIME** - input(none), output(milliseconds)

Since EDL does not support arrays of pointers to blocks, this PR adds the **oe_vector_t** type for packing/unpacking vectors.


common/sgx/internal.edl
```diff
@@ -11,17 +11,126 @@
...
+            [out, size=strings_buf_size] void* strings_buf,
+            size_t strings_buf_size,
+            [out, count=1] size_t* strings_buf_size_out);
+
+        void oe_thread_wake_wait_ocall(
+            [user_check] oe_enclave_t* oe_enclave,
+            uint64_t waiter_tcs,
+            uint64_t self_tcs);
+
+        void* oe_realloc_ocall(
```
anakrish: It would be preferable to leave realloc as a core OCALL. 
For SGX, Ocalls require host memory allocation just to make ocalls and this would mean that in order to do reallocation, we end up doing an allocation of marshaling struct in host memory and then freeing it.

common/sgx/internal.edl
```diff
@@ -11,17 +11,126 @@
...
+
+        // Write a string to the console (device: (0=stdout, 1=stderr).
+        void oe_write_ocall(
+            int device,
+            [in, string] const char* str,
+            size_t len);
 
-        /* This function returns its value parameter. */
-        int oe_internal_ping_ocall(int value);
+        uint32_t oe_get_cpuid_table_ocall(
```
anakrish: I like this design of enclave pulling the CPUID table.

enclave/asym_keys.c
```diff
@@ -480,188 +480,110 @@ void oe_free_key(
...
+        OE_RAISE(OE_BUFFER_TOO_SMALL);
 
-done:
-    uarg->result = result;
+    if (key_info_size < arg.key_info_size)
+        OE_RAISE(OE_BUFFER_TOO_SMALL);
 
-    if (enclave_user_data != NULL)
-        free(enclave_user_data);
+    memcpy(key_buffer, arg.key_buffer, arg.key_buffer_size);
```
anakrish: Our functions rarely allocate and return buffers to the caller. oe_get_public_key_by_policy seems to be an exception.
Consider making oe_get_public_key_by_policy take an output buffer to write to rather than allocating and returning outputs.

enclave/asym_keys.c
```diff
@@ -480,188 +480,110 @@ void oe_free_key(
...
     {
-        *out = NULL;
-        result = OE_OK;
-        goto done;
-    }
+        uint8_t* key_buffer;
+        size_t key_buffer_size;
+        uint8_t* key_info;
+        size_t key_info_size;
+    } arg;
```
anakrish: Consider initializing the arg struct here. Otherwise if line 509 raises OE_INVALID_PARAMETER, the value of arg is junk.

enclave/asym_keys.c
```diff
@@ -480,188 +480,110 @@ void oe_free_key(
...
-        return;
-
-    arg = *uarg;
-    arg.key_buffer = NULL;
-    arg.key_buffer_size = 0;
+    struct
+    {
+        uint8_t* key_buffer;
+        size_t key_buffer_size;
+    } arg;
```
anakrish: Consider initializing arg here.

enclave/core/sgx/backtrace.c
```diff
@@ -83,3 +85,99 @@ int oe_backtrace(void** buffer, int size)
...
+    else if ((oe_result_t)retval != OE_OK)
+    {
+        goto done;
+    }
+
+    /* Convert vector to array of strings. */
+    {
+        oe_vector_t* vec;
+
+        if (!(vec = oe_vector_relocate(buf, (size_t)size)))
```
anakrish: It is not clear why we need a new concept of a vector. This vector does not seem to have the same semantics as std::vector.
What are the security guarantees of this vector if the host intentionally manipulates the vector's bytes?

Consider just writing strings one after the other onto a buffer and then reading them one after the other at the receiver with proper null checks.

enclave/core/sgx/report.c
```diff
@@ -393,36 +358,24 @@ void oe_free_report(uint8_t* report_buffer)
     oe_free(report_buffer);
 }
 
-oe_result_t _handle_get_sgx_report(uint64_t arg_in)
+uint32_t oe_get_sgx_report_ecall(
```
anakrish: Not clear why this can't return oe_result_t

enclave/core/sgx/tracee.c
```diff
@@ -123,21 +91,15 @@ oe_result_t _handle_oelog_init(uint64_t arg)
...
 
-oe_result_t oe_log(log_level_t level, const char* fmt, ...)
+oe_result_t oe_log(oe_log_level_t level, const char* fmt, ...)
 {
     oe_result_t result = OE_FAILURE;
-    oe_log_args_t* args = NULL;
     oe_va_list ap;
     int n = 0;
     int bytes_written = 0;
+    char message[OE_LOG_MESSAGE_LEN_MAX];
```
anakrish: This declares a 2048 byte variable in the stack. Is this safe?

enclave/sgx/qeidinfo.c
```diff
@@ -8,96 +8,103 @@
...
+    {
+        OE_RAISE(OE_FAILURE);
+    }
+
+    /* Subsequent calls to expand the buffers as needed. Note that the output
+     * buffer sizes returned from the first call might still be too small on
+     * the next call since the qe_id_info and issuer_chain could change
+     * between calls. To compensate, loop until success.
+     */
+    while ((oe_result_t)retval == OE_BUFFER_TOO_SMALL)
```
anakrish: Should we give up after n iterations?
What are the security implications of looping?

enclave/sgx/qeidinfo.c
```diff
@@ -8,96 +8,103 @@
...
 {
     oe_result_t result = OE_FAILURE;
-    size_t host_args_buffer_size = sizeof(*args);
-    uint8_t* host_args_buffer = NULL;
-    oe_get_qe_identity_info_args_t* host_args = NULL;
-    oe_get_qe_identity_info_args_t tmp_args = {0};
+    const size_t QE_ID_INFO_SIZE = OE_PAGE_SIZE;
+    const size_t ISSUER_CHAIN_SIZE = OE_PAGE_SIZE;
+    uint32_t retval;
+    oe_get_qe_identity_info_args_t args;
```
anakrish: Consider initializing the struct at definition.

enclave/sgx/revocationinfo.c
```diff
@@ -3,161 +3,170 @@
...
-    oe_get_revocation_info_args_t* host_args = NULL;
-    oe_get_revocation_info_args_t tmp_args = {0};
-    uint8_t* p = NULL;
-    size_t crl_url_sizes[2] = {0};
-
-    if (args == NULL || args->num_crl_urls != 2 || args->crl_urls[0] == NULL ||
-        args->crl_urls[1] == NULL)
-        OE_RAISE(OE_FAILURE);
+    oe_get_revocation_info_args_t in;
+    oe_get_revocation_info_args_t out;
```
anakrish: Consider initializing the structs at definition.

host/sgx/calls.c
```diff
@@ -367,22 +367,18 @@ static oe_result_t _handle_call_host_function(
...
-                                       "WRITE",
-                                       "SLEEP",
-                                       "GET_TIME",
-                                       "BACKTRACE_SYMBOLS",
-                                       "LOG"};
+    // clang-format off
+    static const char* func_names[] =
+    {
+        "CALL_HOST_FUNCTION",
+        "THREAD_WAKE",
```
anakrish: Thread wake and thread wait have been moved to EDL. Haven't they?

include/openenclave/internal/vector.h
```diff
@@ -0,0 +1,94 @@
...
+ *     [OFFSET0][4]                vector[0]
+ *     [OFFSET1][6]                vector[1]
+ *     [OOFSET2][5]                vector[2]
+ *     [red\0green\0blue\0]
+ *     ```
+ *
+ * Or more compactly:
+ *
+ *     ```
+ *     [3][OFFSET0][4][OFFSET1][6][OFFSET2][5][red\0green\0blue\0]
```
anakrish: The host can intentionally not null terminate the strings and serialize the buffer as:
`[3][OFFSET0][4][OFFSET1][6][OFFSET2][5][reddgreennbluee]`

enclave/asym_keys.c
dthaler: For things that need variable size buffers, functions should allocate internally rather than making the caller call twice (which also has problems if the size needed might change in between).   The Microsoft API guidance is to do allocation internally in such cases.

common/sgx/internal.edl
```diff
@@ -11,17 +11,126 @@
...
+            [in, size=report_size] const void* report,
+            size_t report_size);
+
+        public uint32_t oe_get_public_key_ecall(
+            [in, count=1] const oe_asymmetric_key_params_t* key_params,
+            [in, size=key_info_size] const void* key_info,
+            size_t key_info_size,
+            [out, size=key_buffer_size] void* key_buffer,
+            size_t key_buffer_size,
+            [out, count=1 ] size_t* key_buffer_size_out);
```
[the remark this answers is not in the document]
mikbras: Thanks! I didn't know that. Resolved.

common/sgx/internal.edl
```diff
@@ -11,17 +11,126 @@
...
+
+        public uint32_t oe_get_public_key_by_policy_ecall(
+            uint32_t seal_policy,
+            [in, count=1] const oe_asymmetric_key_params_t* key_params,
+            [out, size=key_buffer_size] void* key_buffer,
+            size_t key_buffer_size,
+            [out, count=1 ] size_t* key_buffer_size_out,
+            [out, size=key_info_size] void* key_info,
+            size_t key_info_size,
+            [out, count=1 ] size_t* key_info_size_out);
```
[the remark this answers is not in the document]
mikbras: Resolved.

common/sgx/internal.edl
```diff
@@ -11,17 +11,126 @@
...
+    untrusted
+    {
+        uint32_t oe_get_qetarget_info_ocall(
+            [out] sgx_target_info_t* target_info);
+
+        uint32_t oe_get_quote_ocall(
+            [in] const sgx_report_t* sgx_report,
+            [out, size=quote_size] void* quote,
+            size_t quote_size,
+            [out] size_t* quote_size_out);
```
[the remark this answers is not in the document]
mikbras: Resolved

common/sgx/internal.edl
mikbras: I thought so too until I realized that the original function has to allocate and free the following structure.

```c++
typedef struct _oe_realloc_args
{
    void* ptr;
    size_t size;
} oe_realloc_args_t;
```

The generated **oe_realloc_ocall** routines are just as efficient.

I added issue #2046 which proposes adding one or more additional arguments to **oe_ocall** for such situations. Also, the edger8r could use these additional arguments for optimization.

The **oe_ocall** and **oe_ecall** functions must carry simple 64-bit values (not pointers to host memory) because the TrustZone work needs to obtain host memory as well.

common/sgx/internal.edl
mikbras: Thanks! It worked around a chicken and egg problem. We cannot use edger8r function tables until we reach a certain level of initialization (which is why the init function is split into one ECALL and one OCALL).

enclave/asym_keys.c
mikbras: @anakrish: the function **oe_get_public_key_by_policy** is preexisting. There are many places where I could have rewritten the entire chain of requests emanating from the edge routine but this PR makes local changes. I noticed that the arguments structures are not confined to the edge routines. They leak into many other functions. This is not ideal but the current PR meets our security goals.

@dthaler: we are trading that ideal to minimize an attack vector. Any **OCALL** that receives memory becomes an attack target. Our general policy is not to pass memory from the host. Also, returning host memory does not play well with the TrustZone memory model nor with the **virtual enclave** memory model (where the enclave runs in a separate process).

enclave/asym_keys.c
mikbras: Line 509 does not examine the args structure. The args structure is cleared immediately afterwards with **memset**.

enclave/asym_keys.c
mikbras: Again, the agument structure is initialized with **memset** before first use.

enclave/asym_keys.c
anakrish: What I meant was that if line 509 raises OE_INVALID_PARAMETER, then control will jump to done label and we would execute `oe_free(arg.key_buffer)` on an uninitialized arg.

enclave/asym_keys.c
anakrish: Currently there is no guarantee that arg is always initialized when control reaches `oe_free(arg.key_buffer)`

enclave/core/sgx/backtrace.c
mikbras: I went with your suggested approach and backed out the vector type. We can discuss later, but some utilities are needed to compenstate for the lack of various features in EDL. I will expound on this in my memory model paper.

enclave/core/sgx/report.c
mikbras: I avoided the use of oe_result_t in EDL because I am not completely confident we can make size guarantees on all future platforms between host and enclave.

enclave/core/sgx/tracee.c
mikbras: I move this to heap memory for now. We really need to come up with a heuristic. Maybe the compiler has a max-stack-size option.

enclave/sgx/qeidinfo.c
mikbras: Yeah, I chose 8 for now. This is a very unlikely scenario but the certificate source could change at the exact time of call but it is a virtual impossibility that it will change twice in under a few milliseconds.

I don't think the security implementations are different than looping just once, since in any iteration the host has complete control over the returned data.

enclave/sgx/qeidinfo.c
mikbras: I want to see more evidence that initializing the first member zero-fills the whole structure for ancient compilers. Dave mentioned something about supporting C89?

@dthaler : Do you know whether initializing a structure with a single zero value clears the entire structure on old compilers?

enclave/sgx/revocationinfo.c
mikbras: Again, I want to see some documentaiton about this.

host/sgx/calls.c
mikbras: No, we retained them because they fall under the "trivial" categoery (single 64-bit input value).

include/openenclave/internal/vector.h
mikbras: This has been removed.

mikbras:
@anakrish: Thank you reviewing this. I have addressed pretty much everything as suggested with the following exceptions:
- I'd like to see some documenation about how the "{0}" sructure initializer correctly zero-fills the entire structure on older compilers like C89. Dave wants to support older compilers in the future.
- The new functions return uint32_t where applicable rather than oe_result_t. I was not confident about the size consistency between host and enclave. I added max members to enums to force this but I am not sure this will work consistently with future compilers.
- realloc remains an EDL call (because, like EDL, the old version obtained memory via oe_host_malloc too).

enclave/asym_keys.c
mikbras: Thanks for the explanation. I researched "{ 0 }" initializers and am switching to those. Thanks.

enclave/asym_keys.c
mikbras: Switched to struct initializer.

enclave/sgx/revocationinfo.c
mikbras: I researched this and I'm satisfied with "{ 0 }". Using that style.

common/argv.c
```diff
@@ -0,0 +1,167 @@
...
+
+            if (index == argc)
+                OE_RAISE(OE_FAILURE);
+
+            while (*p && p != end)
+                p++;
+
+            p++;
+
+            argv[index++] = q;
```
anakrish: There is a scenario to consider where argc != number of strings. Can we make sure this function is safe in that scenario.

common/sgx/internal.edl
anakrish: Thanks for creating the issue.

enclave/core/sgx/hostcalls.c
```diff
@@ -0,0 +1,25 @@
...
+#include "internal_t.h"
+
+void* oe_host_realloc(void* ptr, size_t size)
+{
+    void* retval = NULL;
+
+    if (!ptr)
+        return oe_host_malloc(size);
+
+    if (oe_realloc_ocall(&retval, ptr, size) != OE_OK)
```
anakrish: We need ensure that oe_is_outside_enclave(retval, size) is true. Otherwise the host can handback a pointer to enclave memory and get the enclave to overwrite its contents.

enclave/core/hostcalls.c
```diff
@@ -41,32 +41,6 @@ void* oe_host_calloc(size_t nmemb, size_t size)
...
-    arg_in->ptr = ptr;
-    arg_in->size = size;
-
-    if (oe_ocall(OE_OCALL_REALLOC, (uint64_t)arg_in, &arg_out) != OE_OK)
-    {
-        arg_out = 0;
-        goto done;
-    }
-
-    if (arg_out && !oe_is_outside_enclave((void*)arg_out, size))
```
anakrish: This is the check we need to incorporate in the new version.

enclave/sgx/revocationinfo.c
```diff
@@ -3,161 +3,168 @@
...
-    host_args = (oe_get_revocation_info_args_t*)p;
-    *host_args = *args;
-    p += sizeof(*host_args);
+    /* crl_urls */
+    memcpy(in.crl_urls, args->crl_urls, sizeof(in.crl_urls));
+    in.num_crl_urls = args->num_crl_urls;
 
-    // Copy input buffers.
-    for (uint32_t i = 0; i < args->num_crl_urls; ++i)
+    for (;;)
```
CodeMonkeyLeet: This should be iteration capped as in `qeidinfo.c`.

Also, have we considered mitigating the perf cost of this by pre-allocating a default size that should generally work? We know the general sizes of the collateral data we're pulling for the moment; @soccerGB should have the numbers.

common/argv.c
```diff
@@ -0,0 +1,167 @@
...
+    char** argv = NULL;
+    size_t argv_size;
+    size_t alloc_size = 0;
+
+    if (!buf || !argv_out || !malloc_func || !free_func)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+    /* Calculate the total size of argv. */
+    {
+        const char* p = (const char*)buf;
```
CodeMonkeyLeet: Please use the safemath.h methods for pointer arithmetic; there's not a strong binding of edger8r validated inputs with this method which could end up being used outside that context.

enclave/core/sgx/tracee.c
mikbras: Resolved. Thanks for catching this. I fixed the host side only. The same fix is needed for enclaves too.

enclave/core/sgx/report.c
mikbras: After thinking more about this I switched to using **oe_result_** in the EDL. If there is a problem, at least it will not be with internal APIs.

common/argv.c
mikbras: Resolved. Good catch.

enclave/core/sgx/hostcalls.c
mikbras: Good catch. Resolved.

enclave/core/hostcalls.c
mikbras: Got it. Thanks!

mikbras:
> Thanks for taking this on, there's quite a bit more complexity in this PR than I would have expected. A couple of general remarks:
> 
> Consider the perf impact of the changes, such as whether we are doing more data copies than needed, and especially if we incur more context switches.
> Are there features that are generally missing as part of the Edger8r that should be tracked as additional issues? For example, I'm unclear on what's driving the argv implementation, would we deprecate that in favor of something built into Edger8r?

The number of roundtrips has remained the same in all cases as far as I know.

I will be writing a document to share my experiences using EDL. I probably have the most experience so far. I already mentioned to Andy the limitation with argv-style arguments. I will create an issue to track others.

enclave/CMakeLists.txt
```diff
@@ -41,6 +41,9 @@ if(CMAKE_C_COMPILER_ID MATCHES GNU)
   endif()
 endif()
 
+# Add location of "internal_t.h"
+target_include_directories(oeenclave PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/core)
```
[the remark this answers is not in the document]
mikbras: I'll let Andy answer but the build will fail if this is removed.

tests/argv/enc/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+
+oeedl_file(../argv.edl enclave gen)
+
+add_enclave(TARGET argv_enc
+    UUID 71b0822f-42a3-4543-a97c-ca491f76b82c SOURCES enc.c ${gen})
```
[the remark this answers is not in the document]
mikbras: Good find. The use of UUID troubles me. I'm working with Hernan and will try to find a better solution so we don't have to maintain all these UUID's. This will continue to be a problem.

common/sgx/internal.edl
```diff
@@ -11,17 +11,126 @@
...
+            [in, size=report_size] const void* report,
+            size_t report_size);
+
+        public uint32_t oe_get_public_key_ecall(
+            [in] const oe_asymmetric_key_params_t* key_params,
+            [in, size=key_info_size] const void* key_info,
+            size_t key_info_size,
+            [out, size=key_buffer_size] void* key_buffer,
+            size_t key_buffer_size,
+            [out ] size_t* key_buffer_size_out);
```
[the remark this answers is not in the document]
mikbras: Resolved.

enclave/core/sgx/backtrace.c
```diff
@@ -83,3 +86,93 @@ int oe_backtrace(void** buffer, int size)
...
+    const size_t BUF_SIZE = 4096;
+    size_t buf_size = BUF_SIZE;
+    size_t buf_size_out;
+    uint32_t retval;
+    char** argv = NULL;
+
+    if (!buffer || size < 0)
+        goto done;
+
+    if (!(buf = dlmalloc(buf_size)))
```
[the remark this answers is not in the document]
mikbras: Resolved. Good disambiguation suggestion. This are the symbol names as specified by the **backtrace_symbols** Linux function. Using the name **symbols_buffer** now.

enclave/core/backtrace.c
```diff
@@ -2,46 +2,15 @@
...
-
-done:
-
-    if (args)
-        oe_host_free(args);
-
-    return ret;
-}
-
 oe_result_t oe_print_backtrace(void)
```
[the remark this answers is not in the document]
mikbras: I added issue #2048 because I believe **oe_assert()** should call this function before aborting.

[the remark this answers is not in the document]
mikbras: Also, the static size is necessary when calling the Linux **backtrace()** function whose signature is taken verbatim for **oe_backtrace()**. Avoiding the static buffer would require some rework and would be a deviation from the Linux signature. Also, since this is preexisting, I'd rather treat it separately from thei PR.

enclave/core/backtrace.c
```diff
@@ -2,46 +2,15 @@
 // Licensed under the MIT License.
 
 #include <openenclave/bits/safecrt.h>
+#include <openenclave/corelibc/stdio.h>
 #include <openenclave/enclave.h>
 #include <openenclave/internal/backtrace.h>
 #include <openenclave/internal/calls.h>
 #include <openenclave/internal/globals.h>
+#include <openenclave/internal/malloc.h>
```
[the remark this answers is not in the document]
mikbras: Removed.

include/openenclave/internal/report.h
```diff
@@ -102,10 +28,7 @@ typedef struct _oe_get_revocation_info_args
     size_t crl_size[3];              /* out */
     uint8_t* crl_issuer_chain[3];    /* out */
     size_t crl_issuer_chain_size[3]; /* out */
-
-    // Memory allocated by host to pass outputs back to the enclave. Enclave
-    // must free this memory via oe_host_free.
-    uint8_t* host_out_buffer; /* out */
+    uint8_t* buffer;
```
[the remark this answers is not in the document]
mikbras: Resolved.

host/sgx/ocalls.h
```diff
@@ -7,19 +7,11 @@
...
 void HandleThreadWait(oe_enclave_t* enclave, uint64_t arg);
 void HandleThreadWake(oe_enclave_t* enclave, uint64_t arg);
-void HandleThreadWakeWait(oe_enclave_t* enclave, uint64_t arg_in);
-
-void HandleGetQuote(uint64_t arg_in);
-void HandleGetQETargetInfo(uint64_t arg_in);
-void HandleGetQuoteRevocationInfo(uint64_t arg_in);
-void HandleGetQuoteEnclaveIdentityInfo(uint64_t arg_in);
 
 void oe_handle_backtrace_symbols(oe_enclave_t* enclave, uint64_t arg);
```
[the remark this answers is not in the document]
mikbras: Good catch. I though I deleted all these.

enclave/core/sgx/cpuid.c
```diff
@@ -20,24 +21,30 @@ static uint32_t _cpuid_table[OE_CPUID_LEAF_COUNT][OE_CPUID_REG_COUNT];
 **
 **==============================================================================
 */
-oe_result_t oe_initialize_cpuid(oe_init_enclave_args_t* args)
+oe_result_t oe_initialize_cpuid(void)
 {
     oe_result_t result = OE_UNEXPECTED;
+    uint32_t cpuid_table[OE_CPUID_LEAF_COUNT][OE_CPUID_REG_COUNT];
```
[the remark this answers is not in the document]
mikbras: Resolved.

enclave/core/sgx/cpuid.c
```diff
@@ -20,24 +21,30 @@ static uint32_t _cpuid_table[OE_CPUID_LEAF_COUNT][OE_CPUID_REG_COUNT];
...
+    uint32_t retval;
 
-    if (args == NULL || !oe_is_within_enclave(args, sizeof(*args)))
-        OE_RAISE(OE_INVALID_PARAMETER);
+    if (oe_get_cpuid_table_ocall(&retval, cpuid_table, sizeof(cpuid_table)) !=
+        OE_OK)
+    {
+        OE_RAISE(OE_FAILURE);
```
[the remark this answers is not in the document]
mikbras: Resolved.

enclave/core/sgx/calls.c
```diff
@@ -142,38 +142,23 @@ static oe_result_t _handle_init_enclave(uint64_t arg_in)
...
-            oe_init_enclave_args_t safe_args;
-
-            if (!oe_is_outside_enclave(args, sizeof(*args)))
-                OE_RAISE(OE_INVALID_PARAMETER);
-
-            /* Copy structure into enclave memory */
-            safe_args = *args;
-
-            if (!oe_is_outside_enclave(safe_args.enclave, 1))
+            if (!oe_is_outside_enclave(enclave, 1))
```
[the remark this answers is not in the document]
mikbras: Resolved. I added this to **enclave.h**:

```
 This pointer is opaque and is never dereferenced within the enclave. The corresponding
structure definition (struct _oe_enclave) is not visible from any enclave headers.
```

enclave/core/assert.c
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 #include <openenclave/enclave.h>
+#include <openenclave/internal/backtrace.h>
```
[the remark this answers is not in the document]
mikbras: Removed for now.  See issue #2048.

tests/argv/argv.edl
```diff
@@ -1,9 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#include "internal_u.h"
-
-int oe_internal_ping_ocall(int value)
+enclave
```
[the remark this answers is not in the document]
mikbras: I don't know how to fix this.

common/sgx/internal.edl
```diff
@@ -11,17 +11,126 @@
...
+
+        uint32_t oe_get_qe_identify_info_ocall(
+            [out, size=qe_id_info_size] void* qe_id_info,
+            size_t qe_id_info_size,
+            [out] size_t* qe_id_info_size_out,
+            [out, size=issuer_chain_size] void* issuer_chain,
+            size_t issuer_chain_size,
+            [out] size_t* issuer_chain_size_out);
+
+        uint32_t oe_get_revocation_info_ocall(
```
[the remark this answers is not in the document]
mikbras: No, arrays of pointers are not planned for deep copy. I mentioned this to Andy and he said it would be too hard to support. I'm planning a document on my experience with EDL.

common/sgx/internal.edl
```diff
@@ -11,17 +11,126 @@
...
+        void oe_thread_wake_wait_ocall(
+            [user_check] oe_enclave_t* oe_enclave,
+            uint64_t waiter_tcs,
+            uint64_t self_tcs);
+
+        void* oe_realloc_ocall(
+            [user_check] void* ptr,
+            size_t size);
+
+        // Write a string to the console (device: (0=stdout, 1=stderr).
```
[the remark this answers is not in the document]
mikbras: Resolved.

host/sgx/ocalls.c
```diff
@@ -251,22 +507,63 @@ static char** _backtrace_symbols(
...
+}
+
+void oe_write_ocall(int device, const char* str, size_t len)
+{
+    if (str && (device == 0 || device == 1))
     {
-        log_message(true, args);
+        FILE* stream = (device == 0) ? stdout : stderr;
+
+        if (len == (size_t)-1)
```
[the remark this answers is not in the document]
mikbras: Good idea. This expression is more succinct.

```
void oe_write_ocall(int device, const char* str, size_t maxlen)
{
    if (str && (device == 0 || device == 1))
    {
        FILE* stream = (device == 0) ? stdout : stderr;
        fprintf(stream, "%.*s", (int)strnlen(str, maxlen), str);
        fflush(stream);
    }
}
```

host/sgx/asym_keys.c
```diff
@@ -21,27 +21,114 @@ oe_result_t oe_get_public_key_by_policy(
     size_t* key_info_size)
 {
     oe_result_t result = OE_UNEXPECTED;
-    oe_get_public_key_by_policy_args_t args;
+    uint32_t retval;
+    const size_t KEY_BUFFER_SIZE = 1024;
+    const size_t KEY_INFO_SIZE = 1024;
```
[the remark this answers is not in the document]
mikbras: Thanks. Resolved.

host/sgx/asym_keys.c
```diff
@@ -21,27 +21,114 @@ oe_result_t oe_get_public_key_by_policy(
     size_t* key_info_size)
 {
     oe_result_t result = OE_UNEXPECTED;
-    oe_get_public_key_by_policy_args_t args;
+    uint32_t retval;
+    const size_t KEY_BUFFER_SIZE = 1024;
```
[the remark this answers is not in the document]
mikbras: Resolved. Thanks. I defined **DEFAULT_KEY_BUFFER_SIZE** on the host side and a comment on the enclave side to update that variable if keys sizes are increased.

tests/CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ if (UNIX OR ADD_WINDOWS_ENCLAVE_TESTS OR USE_CLANGW)
         add_subdirectory(sealKey)
         add_subdirectory(stdc)
         add_subdirectory(syscall)
+        add_subdirectory(argv)
```
HernanGatta: Consider building this test for TrustZone, too.

`argv/enc/enc.c` will need to use `OE_SET_ENCLAVE_OPTEE` like in `mixed_c_cpp/enc/enc.cpp`.

tests/argv/host/host.c
```diff
@@ -0,0 +1,36 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "argv_u.h"
+
+int main(int argc, const char* argv[])
+{
+    oe_result_t result;
+    oe_enclave_t* enclave = NULL;
+    const oe_enclave_type_t type = OE_ENCLAVE_TYPE_SGX;
```
HernanGatta: Consider changing `OE_ENCLAVE_TYPE_SGX` to `OE_ENCLAVE_TYPE_AUTO` so that the test may (eventually) run on TrustZone as well, seeing as the functionality is not SGX-specific.

common/argv.c
```diff
@@ -0,0 +1,171 @@
...
+        alloc_size += argv_size;
+
+        while (p != end)
+        {
+            const char* start = p;
+
+            while (*p && p != end)
+                p++;
+
+            if (*p != '\0')
```
anakrish: It would be preferable to do
```c
if (p == end)
    OE_RAISE(OE_FAILURE);
```

Consider a case where buf is not null terminated, but buf[buf_size] **happens** to be NULL.
In such a case, the loop
```c
while (*p && p != end)
   p++;
```
would terminate when p == end (given that *end happens to be NULL).

Since *p is NULL, we will do p++ and go beyond end. 
This will cause the check (p != end) to be never true (untill wrapping happens), leading to reading of all the memory to the right of buf until a NULL character is found.

host/sgx/create.c
```diff
@@ -332,6 +332,34 @@ static oe_result_t _add_data_pages(
...
+        oe_get_cpuid(
+            i,
+            subleaf,
+            &table[i][OE_CPUID_RAX],
+            &table[i][OE_CPUID_RBX],
+            &table[i][OE_CPUID_RCX],
+            &table[i][OE_CPUID_RDX]);
+    }
+
+    memcpy(cpuid_table, &table, cpuid_table_size);
```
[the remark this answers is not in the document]
mikbras: Resolved.

host/sgx/create.c
```diff
@@ -346,30 +374,12 @@ static oe_result_t _add_data_pages(
...
-    // Pass the enclave handle to the enclave.
-    args.enclave = enclave;
-
-    {
-        uint64_t arg_out = 0;
-        OE_CHECK(oe_ecall(
-            enclave, OE_ECALL_INIT_ENCLAVE, (uint64_t)&args, &arg_out));
-        OE_CHECK((oe_result_t)arg_out);
-    }
+    OE_CHECK((oe_result_t)result_out);
```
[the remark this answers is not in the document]
mikbras: Good point. I added **oe_is_valid_result()**. The checks for this function now look like this:

```
    if (result_out > OE_UINT32_MAX)
        OE_RAISE(OE_FAILURE);

    if (!oe_is_valid_result((uint32_t)result_out))
        OE_RAISE(OE_FAILURE);
```

enclave/sgx/revocationinfo.c
```diff
@@ -3,161 +3,168 @@
...
-    {
-        result =
-            oe_safe_add_sizet(strlen(args->crl_urls[i]), 1, &crl_url_sizes[i]);
-        if (result != OE_OK)
-            goto done;
-
-        result = oe_safe_add_sizet(
-            host_args_buffer_size, crl_url_sizes[i], &host_args_buffer_size);
+    if (!args)
+        OE_RAISE(OE_FAILURE);
```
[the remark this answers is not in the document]
mikbras: Thanks. Resolved.

enclave/sgx/qeidinfo.c
```diff
@@ -8,96 +8,110 @@
...
-    if (args == NULL)
+    const size_t QE_ID_INFO_SIZE = OE_PAGE_SIZE;
+    const size_t ISSUER_CHAIN_SIZE = OE_PAGE_SIZE;
+    uint32_t retval;
+    oe_get_qe_identity_info_args_t args = {0};
+    size_t iterations = 0;
+    const size_t MAX_ITERATIONS = 8;
+
+    if (args_out == NULL)
         OE_RAISE(OE_FAILURE);
```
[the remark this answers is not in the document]
mikbras: Resolved.

enclave/sgx/revocationinfo.c
mikbras: Created issue #2049 and will need Cheng-Ming's input.

common/argv.c
mikbras: Resolved. Thanks.

common/argv.c
mikbras: Resolved.

enclave/core/optee/hostcalls.c
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <openenclave/enclave.h>
+
+void* oe_host_realloc(void* ptr, size_t size)
```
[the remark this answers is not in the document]
mikbras: Thank you Hernan. I would like to help with that after this PR is in.

host/sgx/ocalls.c
```diff
@@ -105,67 +102,326 @@ void HandleThreadWake(oe_enclave_t* enclave, uint64_t arg_in)
...
+    if (!waiter_tcs || !self_tcs)
         return;
 
 #if defined(__linux__)
+    HandleThreadWake(enclave, waiter_tcs);
+    HandleThreadWait(enclave, self_tcs);
+#elif defined(_WIN32)
+    HandleThreadWake(enclave, waiter_tcs);
+    HandleThreadWait(enclave, self_tcs);
+#endif
```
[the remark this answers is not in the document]
mikbras: It will show up when we merge with the master branch.

common/sgx/internal.edl
```diff
@@ -11,17 +11,126 @@
 **==============================================================================
 */
 
-enclave {
+enclave
```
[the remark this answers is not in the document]
mikbras: Agreed. I'll have a follow on PR that separates out what you need.

tests/CMakeLists.txt
mikbras: THank you Hernan. We will work on this after we merge this PR.

tests/argv/host/host.c
mikbras: Thanks. Resolved.

common/result.c
```diff
@@ -111,3 +112,65 @@ const char* oe_result_str(oe_result_t result)
...
+        case OE_VERIFY_REVOKED:
+        case OE_CRYPTO_ERROR:
+        {
+            return true;
+        }
+        case __OE_RESULT_MAX:
+        {
+            return false;
+        }
+            /* Please do not add a default case! */
```
anakrish: Nice idea to let the compiler catch unhandled enums.

CodeMonkeyLeet: @anakrish For my understanding, how does this construction let the compiler catch unhandled enums?

enclave/sgx/revocationinfo.c
CodeMonkeyLeet: @mikbras Buffer sizing aside, why do we allow this to loop infinitely as opposed to capping it as in `qeidinfo.c`

host/sgx/asym_keys.c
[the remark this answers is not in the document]
CodeMonkeyLeet: @mikbras You've updated the `KEY_BUFFER_SIZE` but not the `KEY_INFO_SIZE`, were you intending to?

common/result.c
anakrish: @CodeMonkeyLeet 
It is along the lines of this example: https://godbolt.org/z/DQGCE8
The -Werror=switch would cause a warning to be emitted for unhandled enum values within a switch case.

mikbras:
> LGTM. Thanks for doing this!

Thank you Anand!

PR #2050, opened 2019-07-22
OE_AEP_ADDRESS variable to obtain precise location of ENCLU.

anakrish:
For single-step debugging to work, OE_AEP must exacly point to the
ENCLU instruction.
If used as a function with type void (void), the linker could emit
thunks to wrap the OE_AEP function (say for incremental linking).
To obtain the precise location of the ENCLU, the OE_AEP_ADDRESS
variable is used to store the computed address of the ENCLU instruction.

Note: OE_AEP is not a proper C function. It does not confirm to the
Linux/Windows ABI.

### DIFF
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -775,7 +775,7 @@ oe_result_t oe_ecall(
     OE_CHECK(_do_eenter(
         enclave,
         tcs,
-        OE_AEP,
+        (void (*)())OE_AEP_ADDRESS,
         code,
         func,
         arg,
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -104,7 +104,7 @@ ThreadBinding* GetThreadBinding()
 static oe_result_t _enter_sim(
     oe_enclave_t* enclave,
     void* tcs_,
-    void (*aep)(void),
+    uint64_t aep,
     uint64_t arg1,
     uint64_t arg2,
     uint64_t* arg3,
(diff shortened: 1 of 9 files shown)


anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

host/sgx/calls.c
```diff
@@ -775,7 +775,7 @@ oe_result_t oe_ecall(
     OE_CHECK(_do_eenter(
         enclave,
         tcs,
-        OE_AEP,
+        (void (*)())OE_AEP_ADDRESS,
```
jxyang: Maybe we should change the signature of `_do_eenter` instead of type-casting here?

jxyang:
LGTM except for a minor issue.

host/sgx/calls.c
anakrish: Great suggestion. I have changed the signature to take in uint64_t. This reflects AEP accurately since it is not a void (void) function.

host/sgx/linux/aep.S
```diff
@@ -34,3 +31,23 @@ OE_AEP:
...
+//==============================================================================
+//
+// uint64_t OE_AEP_ADDRESS
+//
+//     The address of the ENCLU instruction is stored in this variable.
+//     If the OE_AEP function were to be used in code, the linker could create
+//     thunks that wrap the function. For example, when incremental linking is
+//     enabled, the linker on windows creates an entry in the ILT table for
+//     each function and uses that wherever the function is referenced.
+//     Thus OE_AEP would endup pointing to the thunk in the ILT which is not
```
CodeMonkeyLeet: Typo: "end up", similarly in windows/aep.asm.

anakrish: Fixed.

Issue #2040, opened 2019-07-17
Remove weak alias for strcmp


mikbras:
LKL had to remove this alias because it is not exported.

OE_WEAK_ALIAS(oe_strcmp, strcmp);

johnkord:
We think we can remove this.

mikbras:
This weak alias was removed. All CI passes without it. Not sure of original intent.

PR #2052, opened 2019-07-23
Remove strcmp weak alias (one-line change)

mikbras:
This PR removes the following line from **enclave/core/string.c**.

```C++
OE_WEAK_ALIAS(oe_strcmp, strcmp);
```

This was reported by Peter from the **SGX-LKL** team (issue #2040). The intent of this line was probably to provide a **strcmp** definition when **MUSL** is not present to provide it.

### DIFF
--- a/enclave/core/string.c
+++ b/enclave/core/string.c
@@ -148,8 +148,6 @@ char* oe_strdup(const char* s)
     return p;
 }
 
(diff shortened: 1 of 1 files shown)


mikbras:
Thanks Anand and Xuejun.

Issue #2048, opened 2019-07-22
oe_assert() should call oe_print_backtrace()


mikbras:
The **oe_assert()** function should call **oe_print_backtrace()** before it aborts. This will provide the developer with the stack trace leading to the assert.

PR #2059, opened 2019-07-24
Extend assert() to print the call stack (backtrace).

mikbras:
This PR extends **assert()** to print the call stack in addition to the assertion message. This works for both **assert()** and **oe_assert()**.

See issue #2048.

### DIFF
--- a/enclave/core/assert.c
+++ b/enclave/core/assert.c
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 #include <openenclave/enclave.h>
+#include <openenclave/internal/backtrace.h>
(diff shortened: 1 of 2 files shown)


radhikaj:
@mikbras Can we rebase this and merge it?

jhand2:
@mikbras Please rebase and add commits to DCO

mikbras:
> Please don't introduce merge commits. Prefer rebasing instead.
> https://git-scm.com/book/en/v2/Git-Branching-Rebasing

Thank you Jordan. So do I need to backout the merge before rebasing? I'm not sure how to recover from this state.

PR #2062, opened 2019-07-26
Remove backward dependency from liboecore to liboeenclave

mikbras:
This PR removes the static backward dependency of **liboecore** on **liboeenclave**. This is urgently needed to resolve linking problems encountered by TrustZone. 

The PR introduces upcalls (callbacks) that are installed by **liboecore** from the **_start** function.


HernanGatta:
@mikbras: We're updating our fork of OP-TEE (3.3.0 to 3.6.0) and a new change is coming down the pipe that makes it so that the trick we are using to get enclaves to compile won't work anymore. Let's revive this.

jhand2:
I would prefer that we just add weak stubs in liboecore which always fail. Then if liboeenclave is linked, the weak symbols will be replaced.

kuanggang-git:

jhand2:
liboecore no longer depends on liboeenclave. Can this be closed?

PR #2063, opened 2019-07-30
Converge crypto implementations

CodeMonkeyLeet:
This changeset breaks out some of the refactor changes for supporting BCrypt on windows:

- Add host/crypto/magic.h for magic number reuse between openssl and bcrypt.
- Consolidate `oe_cert_chain_get_root_cert()` and `oe_cert_chain_get_leaf_cert()` implementations in common/cert.c.

### DIFF
--- a/common/cert.c
+++ b/common/cert.c
@@ -211,6 +211,33 @@ done:
     return result;
 }
 
+oe_result_t oe_cert_chain_get_root_cert(
+    const oe_cert_chain_t* chain,
+    oe_cert_t* cert)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    size_t length = 0;
+
+    OE_CHECK(oe_cert_chain_get_length(chain, &length));
+    if (length == 0)
+        OE_RAISE(OE_NOT_FOUND);
+    OE_CHECK(oe_cert_chain_get_cert(chain, length - 1, cert));
+    result = OE_OK;
+
+done:
+    return result;
+}
+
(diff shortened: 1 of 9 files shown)


common/cert.c
```diff
@@ -211,6 +211,36 @@ oe_result_t oe_get_crl_distribution_points(
...
+}
+
+oe_result_t oe_cert_chain_get_leaf_cert(
+    const oe_cert_chain_t* chain,
+    oe_cert_t* cert)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    size_t length;
+
+    OE_CHECK(oe_cert_chain_get_length(chain, &length));
```
soccerGB: Could the call to oe_cert_chain_get_length be skipped?

common/cert.c
```diff
@@ -211,6 +211,36 @@ oe_result_t oe_get_crl_distribution_points(
     return result;
 }
 
+oe_result_t oe_cert_chain_get_root_cert(
+    const oe_cert_chain_t* chain,
```
radhikaj: Can these functions be called from the enclave? Can a plugin just use headers under crypto. Just wondering why there was a change to include the internal cert.h rather than crypto cert.h

CodeMonkeyLeet: Yes, they can and they are defined in internal/crypto/cert.h. The change to the headers is unnecessary and will be reverted.

common/cert.c
CodeMonkeyLeet: Yes, this call is redundant in the original implementation. I will remove it.

PR #2064, opened 2019-07-30
Enable tests/crypto/host random tests on Windows

CodeMonkeyLeet:
- Enable `TestRandom()` and `TestRdrand()` on Windows host.
- Fix syntax error in common/sgx/rand.asm
- Fix missing byproduct declaration in tests/crypto/data/CMakelists.txt
- Rename random_tests.c constants for clarity.

### DIFF
--- a/common/sgx/rand.asm
+++ b/common/sgx/rand.asm
@@ -20,6 +20,6 @@ _rdrand_retry:
 
 	ret
 
-_rdrand ENDP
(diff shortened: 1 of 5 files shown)


CodeMonkeyLeet:
@radhikaj Yes, the remaining tests (except for TestASN1, which will stay unimplemented on Windows) will be enabled in an upcoming PR.

PR #2065, opened 2019-07-30
Clean up oe_memalign semantics in enclave.

### LINKED ISSUE #1444: oe_memalign use in host is not compatible with enclave use in common code
oehost implements `oe_memalign()` to split the implementation between Win32 & Linux implementations, but it expects that `oe_memalign()` is paired with `oe_memalign_free()` since it maps it to `_aligned_malloc()` and `_aligned_free()` on Win32.

This will be a problem for code in the common folder shared between host & enclave, since enclave usage is aligned with POSIX/MUSL expectation that memaligned pointers can simply be freed with `free()`. This is not yet a bug as of v0.41, since nothing in common uses oe_memalign currently, but the overlapping use of the same methods for different encapsulations should be resolved.

CodeMonkeyLeet:
- Add oe_memalign_free in enclave.
- Remove redundant `public_key_pem` and `private_key_pem` globals in ec_tests.c
- Update ec_tests.c to dynamically allocate aligned memory for `_test_crl_distribution_points()`

Fixes #1444

### DIFF
--- a/tests/crypto/ec_tests.c
+++ b/tests/crypto/ec_tests.c
@@ -49,10 +49,8 @@ static char _CHAIN[max_cert_chains_size];
 static char _PRIVATE_KEY[max_key_size];
 static char _PUBLIC_KEY[max_key_size];
 static uint8_t _SIGNATURE[max_sign_size];
-uint8_t private_key_pem[max_key_size];
-uint8_t public_key_pem[max_key_size];
-uint8_t x_data[max_coordinates_size];
-uint8_t y_data[max_coordinates_size];
+static uint8_t x_data[max_coordinates_size];
+static uint8_t y_data[max_coordinates_size];
 
 size_t private_key_size;
 size_t public_key_size;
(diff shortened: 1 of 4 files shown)


tests/crypto/ec_tests.c
```diff
@@ -878,7 +876,17 @@ static void _test_crl_distribution_points(void)
     OE_TEST(r == OE_BUFFER_TOO_SMALL);
 
     {
-        OE_ALIGNED(8) uint8_t buffer[buffer_size];
+        uint8_t* buffer = (uint8_t*)oe_memalign(8, buffer_size);
```
anakrish: Can we not use malloc here sicne malloc already guarantee alignment to sizeof(void*)?

CodeMonkeyLeet: Yes, you could. I'm effectively repurposing part of this test to validate that oe_memalign works correctly in host & enclave contexts is all.

PR #2068, opened 2019-07-31
Create oedebugrt.dll on Windows

anakrish:
Housing debugger metadata in a separate DLL is best practice on windows.

On Linux:
  oedebugrt is built as a static library. Therefore no symbols are
  exported. OE_DEBUGRT_EXPORT is null. Debugrt is subsumed into
  the oehost library which is expected to have debug symbols for
  debugging to work.

On Windows:
   oedebugrt is built as a separate DLL. oehost does not directly link
   against oedebugrt.dll. Instead it uses a bridge that dynamically
   load oedebugrt.dll via LoadLibraryEx and then calls into it via
   the bridge.
     a) Enclave applications can run even when oedebugrt.dll is not
        present in the system
     b) Release mode applications are completely decoupled from
        oedebugrt.dll
   The installation/packaging/signing/loading story for oedebugrt.dll
   will be similar to other DLLs that oehost will use.

Also change the order of fields for oe_debug_thread_binding_t.
The next field is grouped together with magic and version fields.
There are the common fields that all structures share.

### DIFF
--- a/host/sgx/windows/debugrtbridge.c
+++ b/host/sgx/windows/debugrtbridge.c
@@ -1,0 +1,132 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <Windows.h>
+#include <openenclave/internal/debugrt/host.h>
+#include <openenclave/internal/trace.h>
+#include "../../hostthread.h"
+
+static struct
+{
+    oe_once_type once;
+    HMODULE hmodule;
+    oe_result_t (*notify_enclave_created)(oe_debug_enclave_t* enclave);
+    oe_result_t (*notify_enclave_terminated)(oe_debug_enclave_t* enclave);
+    oe_result_t (*push_thread_binding)(
+        oe_debug_enclave_t* enclave,
+        struct _sgx_tcs* tcs);
+    oe_result_t (*pop_thread_binding)(void);
+} _oedebugrt;
+
+static void get_debugrt_function(const char* name, FARPROC* out)
+{
+    *out = GetProcAddress(_oedebugrt.hmodule, name);
+    if (*out == NULL)
+    {
+        OE_TRACE_FATAL("Could not find function %s in oedebugrt.dll", name);
+    }
+}
+
+static void load_oedebugrt(void)
+{
+    if (_oedebugrt.hmodule != NULL)
+    {
+        OE_TRACE_FATAL("oedebugrt.dll has already been loaded. Multiple load "
+                       "attempts detected.");
+    }
+
+#ifndef NDEBUG
+    /**
+     *  In debug mode, give preference to OE_DEBUGRT_PATH.
+     *  This is mainly used for OE SDK development.
+     */
+    {
+        char* debugrtpath = getenv("OE_DEBUGRT_PATH");
+        _oedebugrt.hmodule = LoadLibraryExA(
+            debugrtpath,
+            NULL, /* reserved */
+            /* Search only specified path. */
+            LOAD_LIBRARY_SEARCH_DEFAULT_DIRS);
+    }
+
+#endif
+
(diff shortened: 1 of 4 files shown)


host/sgx/windows/debugrtbridge.c
```diff
@@ -0,0 +1,132 @@
...
+        OE_TRACE_FATAL("Could not find function %s in oedebugrt.dll", name);
+    }
+}
+
+static void load_oedebugrt(void)
+{
+    if (_oedebugrt.hmodule != NULL)
+    {
+        OE_TRACE_FATAL("oedebugrt.dll has already been loaded. Multiple load "
+                       "attempts detected.");
```
radhikaj: This need not be a fatal error. Multiple calls to load library are ok.

host/sgx/windows/debugrtbridge.c
```diff
@@ -0,0 +1,132 @@
...
+#endif
+
+    /* Search for oedebugrt.dll only in the application folder. */
+    if (_oedebugrt.hmodule == NULL)
+    {
+        _oedebugrt.hmodule = LoadLibraryExA(
+            "oedebugrt.dll",
+            NULL, /* reserved */
+            LOAD_LIBRARY_SEARCH_APPLICATION_DIR);
+    }
```
radhikaj: Here we are automatically loading the debug extension? How does this relate with someone using the .load command to load the extension?

radhikaj: Also can we unload the dell we have loaded?

anakrish: > Also can we unload the dll we have loaded?

Done.

anakrish: > Here we are automatically loading the debug extension? How does this relate with someone using the .load command to load the extension?

We are loading the debugger runtime. The runtime will send notifications (enclave create/terminate etc) to the debugger (if one is attached).
If the debug extension has been loaded (via the .load command or by someother means), then the notifications will be handled.

anakrish:
bors try

host/sgx/windows/debugrtbridge.c
```diff
@@ -0,0 +1,138 @@
...
+
+#endif
+
+    /* Search for oedebugrt.dll only in the application folder. */
+    if (_oedebugrt.hmodule == NULL)
+    {
+        _oedebugrt.hmodule = LoadLibraryExA(
+            "oedebugrt.dll",
+            NULL, /* reserved */
+            LOAD_LIBRARY_SEARCH_APPLICATION_DIR);
```
radhikaj: Sorry, last question here. The way I normally stored my debugger extensions for windbg was under c:\debuggers (where windbg was installed). Using LOAD_LIBRARY_SEARCH_APPLICATION_DIR  would need app developers to keep copies of the extension with every app. Is there another way to do it? I dont know what the best practices from the debugging team are. If this is what is recommended, please ignore. Would using "LOAD_LIBRARY_SEARCH_DEFAULT_DIRS" be better?

anakrish: Just to be clear. This is oedebugrt.dll is **not** the debugger extension. It is part of the SDK runtime.

@CodeMonkeyLeet  has a better idea of how the OE DLLs (say for attestation, enclave platform, oedebugrt etc) will be distributed. When that is determined, we will have to change the way we search the debug rt dll.

radhikaj: Thanks for the explanation. I thought it was a debugger extension.

anakrish:
bors r+

Merging this since this is needed for future debugger work.

@jxyang Once this PR goes in, for debugging, you need to copy build\debugger\debugrt\host\oedebugrt.dll to the same folder as the host application (eg echo_host).

PR #2069, opened 2019-07-31
OpenSSL standalone install on Windows

### LINKED ISSUE #1941: Use standalone OpenSSL installation for Windows testing
Currently, we use `openssl` bundled with the Git Windows installation from:
`C:\Program Files\Git\mingw64\bin`

Recently, we installed a newer version of Git on our Jenkins Windows servers, and the `crypto`, `crypto_crls_cert_chains` tests started failing on Windows. This happened because `openssl` bundled with the old Git installation was version `1.0.2`, while the newer Git installation had `openssl 1.1.1`.

The `openssl` versions `1.0.2` and `1.1.1` differ in functionality, and for now our tests only work with `1.0.2`. See https://github.com/microsoft/openenclave/pull/1926#issuecomment-501516179 for more context about `openssl 1.1.1` in our tests.

Therefore, it feels cleaner to have a standalone [OpenSSL Windows](https://slproweb.com/products/Win32OpenSSL.html) installation, and do not depend on the one from Git. This way, we won't get regressions when upgrading Git, and we control easier the `openssl` version installed.

oprinmarius:
This PR enables OpenSSL standalone install on Windows instead of relying on openssl from Git.

fixes #1941

### DIFF
--- a/scripts/install-windows-prereqs.ps1
+++ b/scripts/install-windows-prereqs.ps1
@@ -5,6 +5,8 @@
 Param(
     [string]$GitURL = 'https://github.com/git-for-windows/git/releases/download/v2.19.1.windows.1/Git-2.19.1-64-bit.exe',
     [string]$GitHash = '5E11205840937DD4DFA4A2A7943D08DA7443FAA41D92CCC5DAFBB4F82E724793',
+    [string]$OpenSSLURL = 'https://slproweb.com/download/Win64OpenSSL-1_1_1d.exe',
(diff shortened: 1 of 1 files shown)


johnkord:
My only concern is that this is a third party distribution of OpenSSL executables, unrelated to the OpenSSL developers. There's no official build of it on Windows, from what I can tell. We only use this OpenSSL install for generating crypto/PKI data for tests from what I understand, but from my experience I don't think there are many alternatives to this on Windows, so we're probably stuck with it, unless we build/distribute it ourselves.

Issue #1593, opened 2019-03-21
Need to replace the cert re-ordering workaround with correct chain building logics  during the quote validation 


soccerGB:
There is a upcoming Intel root certificate update change that could create a condition that would break our current cert validation logics ( the cert root might new date than its descendant certs)

soccerGB:
[https://github.com/Microsoft/openenclave/issues/864](url)

johnkord:
@soccerGB Is there any more work to be done for this issue, or can we close this out?

soccerGB:
This is a nice to have feature - non-essential.

PR #2072, opened 2019-08-02
Implement host crypto functions on BCrypt for Windows

### LINKED ISSUE #375: Windows: Implement crypto functions for host on BCrypt
The host side implementation of crypto functions (other than hashing) are all not yet implemented. This means that features such as loading unsigned debug enclaves, which test sign during loading, will fail. 

This issue includes refactoring existing BCrypt implementations on the host side used by oesign:
- `host/crypto/bcrypt/rsa.c ` has copied structs and functions from `host/crypto/openssl/[rsa.c|keys.[c|h]]`, so they need to be consolidated.

This also includes enabling the following tests for enclaves on Windows:
- crypto
- crypto_crls_cert_chains
- mbed

This has a dependency on #839 as mbedtls has a dependency on CPUID in enclaves which OE provides through handling the illegal instruction exception.

This is related to the #1003 in terms of the broader crypto API refactor.

CodeMonkeyLeet:
This change set brings Windows `host/crypto/bcrypt` implementations up to parity with the `openssl` equivalents for Linux and includes the following:

- Implement missing cert.c, crl.c, ec.c functions.
- Refactor rsa.c to share code with new ec.c via key.c.
- Add PEM encoding functions for Windows crypto implementation.
    - Standardize constant names in internal/pem.h include header.
- Change common/cert.c to only provide `oe_get_crl_distribution_points()` impl on OpenSSL/mbedTLS Linux builds.
    - host/crypto/bcrypt/cert.c provides the Windows impl.
- Enable tests/crypto cert, CRL, EC and RSA tests on Windows.
    - ASN tests are not enabled or supported on Windows.
- Enable tests/crypto_crls_cert_chain on Windows.
    - Disable positive tests with extra certs on Windows, revisit this as part of #1593.

Fixes #375


host/crypto/bcrypt/rsa.c
```diff
@@ -9,446 +10,325 @@
...
-        /* Clear the fields of the implementation */
-        oe_secure_zero_fill(impl, sizeof(*impl));
     }
 
+    /*
+     * Note that we use the less secure PKCS1 signature padding
+     * because Intel requires it for SGX enclave signatures.
+     */
+    padding_info->type = BCRYPT_PAD_PKCS1;
+    padding_info->config = malloc(sizeof(BCRYPT_PKCS1_PADDING_INFO));
```
radhikaj: Check of out of memory here?

tests/crypto/tests.c
```diff
@@ -7,10 +7,10 @@ void TestAll()
 {
 #if !defined(_WIN32)
     TestASN1();
+#endif
     TestCRL();
```
radhikaj: Great that you are enabling many tests that were previously disabled!

host/crypto/bcrypt/cert.c
```diff
@@ -0,0 +1,1295 @@
...
+    DWORD cert_issuers_count,
+    HCERTSTORE cert_store)
+{
+    oe_result_t result = OE_UNEXPECTED;
+
+    bool is_validated = false;
+    DWORD flags = _OE_DEFAULT_GET_CRL_FLAGS;
+    PCCRL_CONTEXT prev_crl = NULL;
+    PCCRL_CONTEXT crl = CertGetCRLFromStore(
+        cert_store, cert_issuers[0]->pCertContext, prev_crl, &flags);
```
radhikaj: Can we add a check for crl == null here. I think it works correctly in your impl, since you have a while and then an if. So this is just a suggestion

host/crypto/bcrypt/cert.c
```diff
@@ -0,0 +1,1295 @@
...
+    oe_result_t result = OE_UNEXPECTED;
+    cert_chain_t* impl = (cert_chain_t*)chain;
+    PCCERT_CHAIN_CONTEXT cert_chain = NULL;
+    HCERTSTORE cert_store = NULL;
+    uint32_t cert_count = 0;
+    PCCERT_CONTEXT cert_context = NULL;
+
+    /* Zero-initialize the implementation */
+    if (impl)
+        memset(impl, 0, sizeof(cert_chain_t));
```
radhikaj: Would memset_s be appropriate here?

common/cert.c
```diff
@@ -2,12 +2,17 @@
 // Licensed under the MIT License.
 
 #include <openenclave/bits/safecrt.h>
-#include <openenclave/internal/crypto/asn1.h>
 #include <openenclave/internal/crypto/cert.h>
 #include <openenclave/internal/raise.h>
 #include <openenclave/internal/utils.h>
 #include "common.h"
 
+/* oe_get_crl_distribution_points is shared between host OpenSSL and
```
anakrish: Is this applicable for TrustZone?

host/crypto/bcrypt/bcrypt.h
```diff
@@ -6,10 +6,12 @@
  * are defined in both Windows.h and ntstatus.h. The WIN32_NO_STATUS will
  * prevent these redefinitions.
  */
-#include <ntstatus.h>
+#define NOCRYPT
```
anakrish: Can you update the comment to reflect why NOCRYPT is being defined?

host/crypto/bcrypt/cert.c
```diff
@@ -0,0 +1,1295 @@
...
+
+static const DWORD _OE_DEFAULT_CERT_CHAIN_FLAGS =
+    CERT_CHAIN_CACHE_END_CERT | CERT_CHAIN_REVOCATION_CHECK_CACHE_ONLY |
+    CERT_CHAIN_CACHE_ONLY_URL_RETRIEVAL | CERT_CHAIN_REVOCATION_CHECK_CHAIN;
+
+typedef struct _cert
+{
+    uint64_t magic;
+    PCCERT_CONTEXT cert;
+} cert_t;
```
anakrish: Elsewhere we assert that the sizeof(impl type) <= sizeof(container type). Is there some assertion we could add here?

host/crypto/bcrypt/cert.c
```diff
@@ -0,0 +1,1295 @@
...
+
+OE_INLINE void _cert_init(cert_t* impl, PCCERT_CONTEXT cert_context)
+{
+    if (impl)
```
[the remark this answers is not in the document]
CodeMonkeyLeet: It's not an expected condition, but I kept this for parity with the existing openSSL implementation. I'll take it out of both impls.

---
In reply to: [312187277](https://github.com/openenclave/openenclave/pull/2072#discussion_r312187277) [](ancestors = 312187277)

host/crypto/bcrypt/cert.c
CodeMonkeyLeet: Added checks for `cert_t` and `cert_chain_t` both in bcrypt and openssl implementations. Also renamed `Cert` and `CertChain` names in openssl impl to conform to snake casing.

host/crypto/bcrypt/cert.c
```diff
@@ -0,0 +1,1295 @@
...
+    }
+
+    /* Check that at least one valid URL was found */
+    if (found_urls_count == 0)
+        OE_RAISE(OE_NOT_FOUND);
+
+    *urls_total_length = found_urls_total_length;
+    *urls_count = found_urls_count;
+    *urls = found_urls;
+    found_urls = NULL;
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Yes, per the method comment, on successful execution, we relinquish ownership of the array to the caller via `*urls = found_urls`, who is expected to `_free_urls_array()` on it when done.

host/crypto/bcrypt/cert.c
```diff
@@ -0,0 +1,1295 @@
...
+    oe_result_t result = OE_UNEXPECTED;
+    const cert_chain_t* impl = (const cert_chain_t*)chain;
+    size_t length;
+    PCCERT_CONTEXT found_cert = NULL;
+
+    /* Clear the output certificate for all error pathways */
+    if (cert)
+        memset(cert, 0, sizeof(oe_cert_t));
+
+    /* Reject invalid parameters */
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Done

host/crypto/bcrypt/key.c
```diff
@@ -0,0 +1,566 @@
...
+    LPCWSTR key_blob_type,
+    BYTE** key_blob_data,
+    ULONG* key_blob_size)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    BYTE* export_data = NULL;
+    ULONG export_data_size = 0;
+    ULONG exported_size = 0;
+
+    if (key_blob_data)
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Added param checks for consistency.

host/crypto/bcrypt/rsa.c
CodeMonkeyLeet: Thanks for the catch, added allocation check.

host/crypto/bcrypt/cert.c
CodeMonkeyLeet: I'm not sure what the suggestion is ... is it to encapsulate the `while` loop in the `if (crl)` block? If so, that's not the intention, because the `while` loop can take non-null `crl` variable and set it to null if it cannot find a base CRL, so there still needs to be an `if (crl)` check after the `while` loop.

host/crypto/bcrypt/cert.c
CodeMonkeyLeet: Turns out `memset_s` is actually not one of MSVC's [Security-Enhanced Versions of CRT Functions](https://docs.microsoft.com/en-us/cpp/c-runtime-library/security-enhanced-versions-of-crt-functions?view=vs-2019) unlike `memcpy_s`

common/cert.c
CodeMonkeyLeet: We'll need to clarify with Hernan, but common/cert.c is currently used for shared code for SGX reporting between host and enclave. There needs to be a larger refactor that I would scope out of this PR if ARM needs to reuse this code.

host/crypto/bcrypt/bcrypt.h
CodeMonkeyLeet: Added a comment for this.

host/crypto/bcrypt/ec.c
```diff
@@ -0,0 +1,1091 @@
...
+done:
+    return result;
+}
+
+void oe_ec_public_key_init(
+    oe_ec_public_key_t* public_key,
+    BCRYPT_KEY_HANDLE* key_handle)
+{
+    oe_bcrypt_key_init(
+        (oe_bcrypt_key_t*)public_key, key_handle, OE_RSA_PUBLIC_KEY_MAGIC);
```
gupta-ak: RSA? I see all the ec public keys with this magic number.

CodeMonkeyLeet: Oof. Thanks for catching this!

host/crypto/bcrypt/cert.c
```diff
@@ -0,0 +1,1291 @@
...
+
+done:
+    if (buffer)
+        free(buffer);
+
+    return result;
+}
+
+/* Caller is expected to invoke _free_urls_array on output urls */
+oe_result_t _bcrypt_get_distribution_point_urls(
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Fixed

host/crypto/bcrypt/cert.c
```diff
@@ -0,0 +1,1291 @@
...
+            "CryptImportPublicKeyInfoEx2 failed, err=%#x\n",
+            status);
+
+    result = OE_OK;
+
+done:
+    return result;
+}
+
+oe_result_t _bcrypt_get_key_algorithm(
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Fixed

host/crypto/bcrypt/cert.c
```diff
@@ -0,0 +1,1291 @@
...
+        }
+    }
+
+    result = OE_OK;
+
+done:
+    return result;
+}
+
+oe_result_t _bcrypt_get_public_key_from_cert(
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Fixed

host/crypto/bcrypt/cert.c
```diff
@@ -0,0 +1,1295 @@
...
+done:
+    if (load_cert_store)
+        CertCloseStore(load_cert_store, 0);
+
+    if (pem_cert)
+        free(pem_cert);
+
+    if (der_blob.pbData)
+    {
+        der_blob.pbData = NULL;
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Good catch, fixed.

host/crypto/bcrypt/cert.c
```diff
@@ -0,0 +1,1291 @@
...
+
+        if (!success)
+            OE_RAISE_MSG(
+                OE_CRYPTO_ERROR,
+                "CryptVerifyCertificateSignatureEx failed, err=%#x\n",
+                GetLastError());
+
+        /* Perform CRL validations if certificate trust store provided
+         * contains any CRLs */
+        if (!isRootCert && cert_store)
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Good observation, moved the check for CRLs out of the loop.

host/crypto/bcrypt/cert.c
```diff
@@ -0,0 +1,1291 @@
...
+
+    /* Success on CertVerifyCertificateChainPolicy does not indicate
+     * successful validation, that's reflected in the policy_status */
+    if (policy_status.dwError != ERROR_SUCCESS)
+        OE_RAISE_MSG(
+            OE_VERIFY_FAILED,
+            "CertVerifyCertificateChainPolicy failed, err=%#x\n",
+            policy_status.dwError);
+
+    /* Explicitly walk the chain and validate the signatures */
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Not completely, my observations from the disassembly walk without debug symbols are recorded in the comment on prior to the `CertVerifyCertificateChainPolicy()` invocation on line 427, but it's not definitive. 

It looked like `CERT_TRUST_IS_UNTRUSTED_ROOT` was hit and causing a lot of other checks to be elided and immediately invalidate the signature, since we don't register the root cert trust in Windows root cert store explicitly but it's not clear if there's some way around that.

PR #2073, opened 2019-08-02
Memory model fix-ups and passing tests for ARM

HernanGatta:
This is the third PR in the ARM support series.

This PR:
1. Removes assumptions that an enclave can read host memory in:
1.1. `oe_host_calloc`;
1.2. `oe_host_memset` (new, see below);
1.3. `oe_host_strndup`.
2. Implements `oe_get_entropy` for OP-TEE;
3. Includes additional math function implementations to catch up with SGX;
4. Allows the following tests to pass on ARM:
4.1. `create-errors`
4.2. `crypto`
4.3. `hexdump`
4.4. `hostcalls`
4.5. `initializers`
4.6. `libc`
4.7. `mixed_c_cpp`
4.8. `pingpong`
4.9. `pingpong-shared`
4.10. `safecrt`

In the next PR, I'll submit changes to run these in an emulator during CI.

---

Why do we need 1.1 ... 1.3?

The first is only used for ECALLs in SGX*. Since SGX can read and write host memory, `oe_host_calloc` can be replaced with `oe_host_malloc` followed by `oe_memset_s`, which is how it is currently implemented. Then, `oe_host_calloc` can be removed entirely, including from tests and public headers; OP-TEE has no use for it.

The second I added because a test requires it (`hostcalls`), but it's asking for trouble; the test should be modified to not need to touch host memory at all.

As for the third, `oeedger8r` is capable of marshaling strings, and it is also not used anywhere but for in tests.

I'd strongly argue for nuking these unless there is a very good reason for keeping them.

\* See `enclave/core/sgx/calls.c:oe_call_host_function_by_table_id`.


mikbras:
Hi Hernan:

Thank you for your efforts. TrustZone is looking better all the time. 

I reviewed the PR and I just have one question. Does TrustZone/OP-TEE support shared host memory and if so, can you use that instead of adding oe_host_memset and oe_host_calloc? 

I implement shared host memory for **virtual enclave** and I just allocated a 1 megabyte chunk and managed it with dlmalloc, as shown here:

https://github.com/openenclave/openenclave/blob/mikbras.virtenc/venclave/host/hostmalloc.c

I had to implement a fake **sbrk** function (that allocates chunks from the shared memory chunk) and trick **dlmalloc** into using it.

include/openenclave/enclave.h
```diff
@@ -173,6 +173,21 @@ void* oe_host_realloc(void* ptr, size_t size);
  */
 void* oe_host_calloc(size_t nmemb, size_t size);
 
+/**
+ * Fill block of host memory.
+ *
+ * This function sets the first **num** bytes of the block of host memory
+ * pointed by **ptr** to value specified by **value**.
+ *
+ * @param ptr The host memory block to fill.
```
dthaler: we should be using [in,out] style doxygen annotations as well.  It's easiest to just start doing it for all new changes than to have to retroactively figure them out later when you actually know the answer for these lines now.

include/openenclave/enclave.h
```diff
@@ -173,6 +173,21 @@ void* oe_host_realloc(void* ptr, size_t size);
  */
 void* oe_host_calloc(size_t nmemb, size_t size);
 
+/**
+ * Fill block of host memory.
+ *
+ * This function sets the first **num** bytes of the block of host memory
```
dthaler: Mentions of parameter names should use /p instead of **

See https://stackoverflow.com/questions/15398711/whats-the-right-way-to-reference-a-parameter-in-doxygen

tests/crypto/tests.c
```diff
@@ -12,7 +12,9 @@ void TestAll()
     TestRSA();
 #endif
     TestRandom();
+#if defined(__x86_64__) || defined(__i386__)
```
dthaler: Why is this #if test needed?  At least add a comment into the code explaining it.

common/optee/optee.edl
```diff
@@ -0,0 +1,30 @@
...
+    include "openenclave/bits/types.h"
+
+    trusted
+    {
+        public void oe_optee_stub_ecall();
+    };
+
+    untrusted
+    {
+        void* oe_calloc_ocall(
```
jxyang: Does this function returns a pointer (to host heap) that cannot be used inside enclave? If so, what's the point?

To be clear, I was actually expecting this call to return a pointer to the shared memory pool on TrustZone.

HernanGatta: The returned pointer points to host-side memory. The issue is that this function is public, so it's expected to be implemented, even if it's useless on TrustZone. This is the reason why I argued for removing it from the public API surface altogether.

The shared memory pool (or dynamically shared memory buffers, which temporarily map host memory into the enclave) is only used in passing to marshal ECALLs/OCALLs, not for any storage longer than a function call boundary.

PR #2074, opened 2019-08-02
POC implementation of switchless ocalls

EMumau:
The branch name is a mistake.

This is an implementation of synchronous switchless ocalls.  The implementation uses a lockless ring buffer to marshal data between the host and the enclave.

This branch is forked off on my implementation of synchronous switchless ecalls.


CodeMonkeyLeet:
Superseded by #2104 and #2125

PR #2077, opened 2019-08-05
Added additional error codes to oe_verify_report()

### LINKED ISSUE #1044: Attestation: OE_VERIFY_FAILED was overloaded for many validation failures, need define new error codes for specific failure condition
The OE_VERIFY_FAILURE code is returned for any validation failures in the the code path of the oe_verify_report() API, which does not help narrow down the root causes, even thought the OE SDK has the exact error code for most the failure scenario. Unfortunately, the original failures were swallowed and replaced with generic  OE_VERIFY_FAILURE  code, which makes debugging attestation issues very difficult.

We need to map to original error codes to unique OE_ error code.

jazzybluesea:
Added the following error codes:
  OE_INCORRECT_REPORT_SIZE - OE report size does not match the expected size.
  OE_QUOTE_VERIFICATION_ERROR - Quote verification error.
  OE_QUOTE_ENCLAVE_IDENTIFY_VERIFICATION_FAILED - QE identify verification failed.
  OE_VERIFY_FAILED_OES_CMAC_MISMATCH - OES CMAC mismatch when verifying report.
  OE_QUOTE_ENCLAVE_IDENTITY_MRSIGNER_MISMATCH - QE identity mrsigner value mismatch.
  QE_QUOTE_ENCLAVE_IDENTITY_ISV_PROD_ID_MISMATCH - QE identity isv product id mismatch.
  QE_QUOTE_ENCLAVE_IDENTITY_ISVSVN_UNEXPECTED - QE identity isvsvn was unexpected.

Fixes #1044


common/sgx/qeidentity.c
```diff
@@ -57,18 +57,29 @@ oe_result_t oe_enforce_qe_identity(sgx_report_body_t* qe_report_body)
...
+            OE_RAISE_MSG(
+                OE_QUOTE_ENCLAVE_IDENTIFY_VERIFICATION_FAILED,
+                "isvprodid mismatch",
+                NULL);
 
         if (qe_report_body->isvsvn < g_qeisvsvn)
-            OE_RAISE_MSG(OE_VERIFY_FAILED, "isvsvn is out-of-date", NULL);
+            OE_RAISE_MSG(
+                OE_QUOTE_ENCLAVE_IDENTIFY_VERIFICATION_FAILED,
+                "isvsvn is out-of-date",
```
soccerGB: OE_QUOTE_ENCLAVE_IDENTIFY_VERIFICATION_FAILED error code is shared by above three error cases. 
It would be easier to root cause if distinct error codes are used.

achamayou: It would also probably help application operators/end-users to include the expected and found isvsvn and isvprodid values in the error messages.

jazzybluesea: Added 3 additional error codes.

include/openenclave/bits/result.h
```diff
@@ -284,6 +284,41 @@ typedef enum _oe_result
...
+
+    /**
+     * QE ISV SVN was not expected.
+     */
+    QE_QUOTE_ENCLAVE_IDENTITY_ISVSVN_UNEXPECTED,
+
+    /**
+     * OES CMAC mismatch when verifying report.
+     */
+    OE_VERIFY_FAILED_OES_CMAC_MISMATCH,
```
CodeMonkeyLeet: Do you mean AES_CMAC? What is OES_CMAC?

common/sgx/report.c
```diff
@@ -53,7 +53,8 @@ static oe_result_t _parse_sgx_report_body(
         sizeof(report_body->mrsigner)));
 
     if (report_body->isvprodid > OE_INT8_MAX)
-        goto done;
+        OE_RAISE(OE_QUOTE_VERIFICATION_ERROR);
```
CodeMonkeyLeet: Not your code, but I realized this check is wrong ... this should be a `OE_UINT16_MAX` check since the point of this is to break it into 2 bytes below (it assumes there's some value to shift in the upper bits). Also, it's unclear if this check is needed since `report_body->isvprodid` is already typed to be `uint16_t`

common/sgx/report.c
```diff
@@ -92,7 +93,7 @@ oe_result_t oe_parse_report(
         OE_RAISE(OE_INVALID_PARAMETER);
 
     if (header->report_size + sizeof(oe_report_header_t) != report_size)
-        OE_RAISE(OE_FAILURE);
+        OE_RAISE(OE_INCORRECT_REPORT_SIZE);
```
CodeMonkeyLeet: This seems extremely specific to warrant its own error code. I would suggest using `OE_QUOTE_VERIFICATION_ERROR` and performing an OE_RAISE_MSG instead to pinpoint the specific case. Similarly for the ISVPRODID check above and others.

common/sgx/qeidentity.c
```diff
@@ -117,34 +138,34 @@ oe_result_t oe_enforce_qe_identity(sgx_report_body_t* qe_report_body)
             sizeof(parsed_info.mrsigner)))
     {
         dump_info(
-            "parsed_info.mrsigner:",
+            "Expected mrsigner, parsed_info.mrsigner:",
```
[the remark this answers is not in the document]
jazzybluesea: Agree.

common/sgx/qeidentity.c
```diff
@@ -57,18 +57,39 @@ oe_result_t oe_enforce_qe_identity(sgx_report_body_t* qe_report_body)
...
-            OE_RAISE_MSG(OE_VERIFY_FAILED, "mrsigner mismatch", NULL);
+        {
+            dump_info(
+                "Expected mrsigner", g_qe_mrsigner, sizeof(g_qe_mrsigner));
+            dump_info(
+                "Actual mrsigner",
+                qe_report_body->mrsigner,
+                sizeof(qe_report_body->mrsigner));
+            OE_RAISE_MSG(
+                OE_QUOTE_ENCLAVE_IDENTITY_MRSIGNER_MISMATCH,
```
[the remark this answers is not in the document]
jazzybluesea: Agree.

include/openenclave/bits/result.h
jazzybluesea: Correct. Typo.

common/sgx/qeidentity.c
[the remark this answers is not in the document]
jazzybluesea: There are OE specific fields in oe_identity_t that I could use instead.  I could use the name of these instead to be more OE specific.

common/sgx/report.c
jazzybluesea: Removed check.

common/sgx/report.c
jazzybluesea: The problem with OE_RAISE_MSG is that the message will now show up in production (since it is not running in debug mode).

common/sgx/qeidentity.c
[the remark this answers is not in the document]
jazzybluesea: The problem with OE_RAISE_MSG is that the message will now show up in production (since it is not running in debug mode).

common/sgx/report.c
CodeMonkeyLeet: Error codes are generally a developer facing feature, so I'm not sure why the production distinction is important here? Under what conditions in production are you expecting to be able to retrieve the OE_RESULT value and interpret it, and who is your audience for this specific error code if not maintainers of OE? (i.e. there's nothing an app developer can do with this error state, is there?)

common/sgx/qeidentity.c
achamayou: @jazzybluesea thank you!

jazzybluesea:
Will create separate PR for the additional comments.

common/sgx/report.c
jazzybluesea: I think it will help faster diagnostics when debugging issues.

include/openenclave/bits/result.h
```diff
@@ -284,6 +284,41 @@ typedef enum _oe_result
...
 
+    /**
+     * OE report size does not match the expected size.
+     */
+    OE_INCORRECT_REPORT_SIZE,
+
+    /**
+     * Quote verification error.
+     */
+    OE_QUOTE_VERIFICATION_ERROR,
```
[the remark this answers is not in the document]
jazzybluesea: Removed.

PR #2084, opened 2019-08-08
Various readme updates

andschwa:
Just cleaned up some sentences and added direct links to building
against the SDK in simulation mode and the governance model.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,43 +7,56 @@ Introduction
 ------------
 
 Open Enclave (OE) is an SDK for building enclave applications in C and C++. An
-enclave application partitions itself into two components (1) An untrusted component (called the
-host) and (2) A trusted component (called the enclave). An enclave is a secure
-container whose memory is protected from entities outside the enclave. These
-protections allow enclaves to perform secure computations with assurances that
-secrets will not be compromised.
+enclave application partitions itself into two components:
+1. An untrusted component (called the host) and
+2. A trusted component (called the enclave).
 
-This SDK is a fully open-source and transparent project, which plans to generalize enclave application models across
-enclave implementations from different hardware vendors. It's a non-vendor specific solution that supports enclave applications both on
-Linux and Windows platforms.
+An _enclave_ is a protected memory region that provides confidentiality for data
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,43 +7,59 @@ Introduction
...
-protections allow enclaves to perform secure computations with assurances that
-secrets will not be compromised.
+enclave application partitions itself into two components:
+1. An untrusted component (called the host) and
+2. A trusted component (called the enclave).
 
-This SDK is a fully open-source and transparent project, which plans to generalize enclave application models across
-enclave implementations from different hardware vendors. It's a non-vendor specific solution that supports enclave applications both on
-Linux and Windows platforms.
+An enclave is a secure container whose memory is protected from entities outside
```
CodeMonkeyLeet: There was feedback from some folks in the past that "secure container" was potentially confusing (Anita & Dmitry, I think).

> An _enclave_ is a protected memory region that provides confidentiality for the data and code executing inside it.

README.md
```diff
@@ -7,43 +7,59 @@ Introduction
...
+the enclave. These protections allow enclaves to perform secure computations
+with assurances that secrets will not be compromised.
 
-The current implementation of Open Enclave is built on [Intel Software Guard Extensions (SGX)](https://software.intel.com/en-us/sgx), other enclave architectures (such as solutions from AMD or ARM) will be added in the future. This public preview focuses on the Linux platform.
+This SDK is a transparent and open source project. It plans to generalize
```
CodeMonkeyLeet: The flow is awkward ... combining with the above, consider:

> An _enclave_ is a protected memory region that provides confidentiality for data and code execution. It is an instance of a Trusted Execution Environment (TEE) which is usually secured by hardware, for example, [Intel Software Guard Extensions (SGX)](https://software.intel.com/en-us/sgx).

> This SDK aims to generalize the development of enclave applications across TEEs from different hardware vendors; while the current implementation is focused on Intel SGX, support for ARM TrustZone is already under development. As an open source project, this SDK also strives to provide a transparent solution that is agnostic to specific vendors, service providers and choice of operating systems.

Issue #1920, opened 2019-06-11
Improve "How to Submit a PR" / add "How to Review a PR"


andschwa:
# Improve "How to Submit a PR" / add "How to Review a PR"

While we have some "Do's and Do not's" for [PRs](https://github.com/microsoft/openenclave/blob/master/docs/Contributing.md#dos-and-donts-for-pull-requests), we should rewrite this section as a step-by-step guide on "How to Submit a PR" with a list of checkboxes (later we can even consider putting these in the PR template), and another section on "How to Review a PR" (again, with a checklist!)

Notes from @paulcallen, with my own commentary:

* feature work design meeting will be announced in teams under Open Enclave teams discussion

> Andy: is this an open channel? If not, we should use the existing Gitter so as to not exclude the community
>> Paul: This sounds fair. @andschwa is there somewhere specific on Gitter where we can do this? Can we create dedicated areas on there specifically for design or is it all dumped into one big open chat? If there is no organization then maybe something like Teams can be used, and we can create a public channel specific for design discussions... maybe even one per design discussion to keep things more organized.

* the design must be agreed upon before a code review, generally before major amount of code is even written

> Andy: This is pretty loose. What does it mean to submit a design? What does it mean for it to be agreed upon? Who are the decision makes for each design review? We have a little bit of this [here](https://github.com/microsoft/openenclave/blob/master/docs/GovernanceModel.md#accepting-contributions), and could learn from [PowerShell's RFC process](https://github.com/PowerShell/PowerShell-RFC)
>> Paul: @andschwa - You have already added a location in git for [design documents](https://github.com/openenclave/openenclave/blob/master/docs/Contributing.md#design-discussion). Thanks.

* each PRs need to get at least 2 reviewers in the related key areas to sign-off.

> Andy: This sounds reasonable to me, and we can even enforce a total of 2 reviewers using GitHub/Bors; however, what are the key areas and who are the owners? Sounds like we need a maintainers doc like [Apache Mesos](https://github.com/apache/mesos/blob/master/docs/committers.md#maintainers).
>> Paul: Yes!

* the reviewers should be chosen ahead of time.

> Andy: Again, sounds like we need a maintainers doc!
>> Paul: Yes!

[...]

PR #2088, opened 2019-08-12
Community Maintenance Committee

andschwa:
This PR defines our Community Maintenance Committee, its members, and its responsibilities. It also outlines the technical areas each committer (a superset of the Community Maintainers) is most interested in.

### DIFF
--- a/docs/GovernanceModel.md
+++ b/docs/GovernanceModel.md
@@ -6,16 +6,20 @@ is to enable a vibrant development community for the Open Enclave SDK. If you
 feel we should make any changes to our guidelines, please start a discussion
 with us.
 
-Our model is based on the [liberal contribution
-policy](https://opensource.guide/leadership-and-governance/). We make decisions
-through a consensus seeking process, rather than a formal voting process.
+Our model is based on the
+[liberal contribution policy](https://opensource.guide/leadership-and-governance/).
+See [below](#accepting-contributions) for more info.
 
 Code of Conduct
 ---------------
 
 In order to maintain a pleasant and welcoming environment, we want to reiterate
-that it is imperative that all community members adhere to our [Code of
-Conduct](Contributing.md#code-of-conduct).
+that it is imperative that all community members adhere to our
+[Code of Conduct](Contributing.md#code-of-conduct).
+Anyone failing to follow the Code of Conduct will be removed from the community
+by the [Community Maintenance Committee](Maintainers.md). If you are made to
+feel uncomfortable, or have any concerns about behavior within the community, we
+encourage you to reach out to members of the Community Maintenance Committee.
 
 Design and Development Discussions
 ----------------------------------
@@ -37,25 +41,40 @@ Remember that security issues should be reported through a separate channel, and
 will receive a response within 24 hours. See [Reporting Security
 Issues](Contributing.md#reporting-security-issues).
 
-Maintainers and Contributors
-----------------------------
+Maintainers, Committers, and Contributors
+-----------------------------------------
 
-We currently define "maintainer" as anyone with direct write access to the Open
-Enclave repository on GitHub, and "contributor" as anyone else. For now, the
-only maintainers are Microsoft employees, but we hope to expand this in the
-future to include consistent contributors. Our process for doing so is not yet
-defined, but if you are interested, let us know!
+We define "maintainer" as members of the Open Enclave "Community Maintenance
+Committee", as listed in the [maintainers document](Maintainers.md). A
+"committer" is anyone with direct write access to the Open Enclave repository on
+GitHub, as granted by the Committee. All maintainers are committers, but not all
+committers are maintainers. Finally, "contributor" is anyone else making
+contributions to the project, including: creating or commenting on issues,
+opening or reviewing pull requests, or other useful contributions such as
+providing support in forums or chats.
+
+See the [maintainers document](Maintainers.md) for the Community Maintenance
+Committee, our process for adding new committers and maintainers, as well the
+areas of expertise for each of the committers.
 
 Accepting Contributions
 -----------------------
 
-Project maintainers will merge changes that improve the product significantly
-and broadly and that align with the [Open Enclave roadmap](
-https://github.com/openenclave/openenclave/projects).
+Project committers will merge changes that improve the product significantly and
+broadly and that align with the
+[Open Enclave roadmap](https://github.com/openenclave/openenclave/projects).
+Contributions must also satisfy the other [published guidelines](Contributing.md).
+Committers may revert changes if they are found to be breaking.
 
-Contributions must also satisfy the other [published
-guidelines](Contributing.md). We may revert changes if they are found to be
-breaking.
+We make most decisions through a consensus seeking process, rather than a formal
+voting process. For example, committers can merge contributions that were
+reviewed without objections. If there are objections that cannot be resolved, an
+issue can be escalated to the Community Maintenance Committee to make a
+decision, which handles issues as discussed in the
+[maintainers document](Maintainers.md).
+
+See the [maintainers document](Maintainers.md) for the list of project
+committers, and how to become one.
 
 Community Approval of Releases
 ------------------------------
--- a/docs/Maintainers.md
+++ b/docs/Maintainers.md
@@ -1,0 +1,95 @@
+Community Maintenance Committee
+===============================
+
+This document describes the Community Maintenance Committee of Open Enclave. By
+our liberal contribution policy outlined in our
+[governance model](GovernanceModel.md), maintainers are committers that are
+trusted to grant new committer rights, and grant new membership into the
+Committee.
+
+When making decisions, the Community Maintenance Committee uses a "consensus
+seeking" process. This means that most decisions should be reached by consensus,
+but when that fails, the Committee calls for a vote where the super majority
+(two-thirds) wins. This is to prevent obstructionism by removing the possibility
+of a one person veto.
+
+Committee Members
+-----------------
+
+| Name                 | Company   | Email                         | GitHub Alias   |
+|----------------------|-----------|-------------------------------|----------------|
+| Akash Gupta          | Microsoft | akagup@microsoft.com          | gupta-ak       |
+| Amaury Chamayou      | Microsoft | amaury.chamayou@microsoft.com | achamayou      |
+| Andrew Schwartzmeyer | Microsoft | andschwa@microsoft.com        | andschwa       |
+| Dave Thaler          | Microsoft | dthaler@microsoft.com         | dthaler        |
+| John Kordich         | Microsoft | johnkord@microsoft.com        | johnkord       |
+| Mike Brasher         | Microsoft | mikbras@microsoft.com         | mikbras        |
+| Simon Leet           | Microsoft | simon.leet@microsoft.com      | CodeMonkeyLeet |
+
+Committee Responsibilities
+--------------------------
+
+The primary responsibility of the Committee is to grant new committer rights
+(that is, write access to the main Open Enclave SDK repository or related
+repositories), and to grant new membership into the committee. Conversely, the
+Committee must also remove committer rights and membership from those found to
+be violating the project's Code of Conduct or otherwise negatively affecting the
+project's community health.
+
+This Committee is not intended to make every technical decision, as those should
+generally be made by agreement among committers as PRs are reviewed and merged.
+Where disagreements take place and need further resolution, those can be brought
+up with the Committee as part of its responsibility to maintain the project's
+community health. Otherwise technical decisions are left to the active
+committers (by virtue of the liberal contribution policy).
+
+The Community Maintenance Committee should meet regularly, for example, once a
+month. This meeting is a private meeting among just the maintainers to nominate
+new committers and maintainer members. Priority consideration should be given to
+those actively contributing to the project. The Committee uses the consensus
+seeking process outlined above when making decisions, including adding or
+removing any members. The Committee should also discuss the community's health
+and work to resolve any negative issues.
+
+In order to maintain a healthy developer community, it is recommended that the
+Committee also host a regular public community meeting. This meeting should be
+open all members of the community, and start with an open forum to hear
+questions or concerns from the community. Any remaining time in the meeting
+should be used to discuss and review open pull requests or issues (especially
+including design documents).
+
+Project Committers
+==================
+
+The following people have been granted commit permissions (that is, write
+access) to the Open Enclave SDK by the Community Maintenance Committee. The area
+column describes which technical areas each committer is most interested in, and
+therefore should usually be consulted for changes relating to that area.
+However, it is up to each committer to determine who should review which PR, and
+when to merge it. Remember that a PR must not be merged if a committer objects;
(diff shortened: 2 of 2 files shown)


andschwa:
Just wanted to drop [this link](https://medium.com/the-node-js-collection/healthy-open-source-967fa8be7951) as a _very_ good read on how to approach this well.

docs/GovernanceModel.md
```diff
@@ -37,14 +37,20 @@ Remember that security issues should be reported through a separate channel, and
...
-defined, but if you are interested, let us know!
+Maintainers, Committers, and Contributors
+-----------------------------------------
+
+We define "maintainer" as members of the Open Enclave "Maintainers Committee",
+as listed in the [maintainers document](Maintainers.md). A "committer" is anyone
+with direct write access to the Open Enclave repository on GitHub, as granted by
+the maintainers committee. All maintainers are committers, but not all
+committers are maintainers. Finally, "contributor" is anyone else making
+contributions to the project, including: creating or comment on issues, opening
```
radhikaj: or "commenting"

docs/Maintainers.md
```diff
@@ -0,0 +1,43 @@
...
+committers and maintainer members. Priority consideration should be given to
+those actively contributing to the project. The committee uses the consensus
+seeking process outlined above when making decisions, including adding or
+removing any members. The committee should also discuss the community's health
+and work to resolve any negative issues.
+
+In order to maintain a healthy developer community, it is recommended that the
+Maintainers Committee also host a regular public community meeting. This meeting
+should be open all members of the community, and start with an open forum to
+hear address questions or concerns from the community. Any remaining time in the
```
radhikaj: hear and address

andschwa:
My paraphrased summary of a conversation with @dthaler:

Dave's main concern was the use of a the consensus-seeking policy to make decisions where those decisions were technical in nature. I did not include any technical responsibilities in this document, which is where I think the confusion came from. I intended to describe the Maintainers Committee as having the role of maintaining the _community_ (who can commit, and who can joint the committee that appoints new committers). My own thought was that the committers as a whole drove the technical direction of the project, by reviewing each others' PRs, but I did not record this in the document.

Dave thinks we should have some process laid out for making technical decisions, and while I don't disagree, it was not my goal in this document, so I am going to amend it to be more clear about that.

andschwa:
Updated with the following additional language, which I envisioned but did not write:

> This committee is not intended to make every technical decision, as those should generally be made by agreement among committers as PRs are reviewed and merged. Where disagreements take place and need further resolution, those can be brought up with the Maintainers Committee as part of its responsibility to maintain the project's community health. Otherwise technical decisions are left to the active committers (by virtue of the liberal contribution policy).

docs/GovernanceModel.md
```diff
@@ -37,14 +37,20 @@ Remember that security issues should be reported through a separate channel, and
...
+We define "maintainer" as members of the Open Enclave "Maintainers Committee",
+as listed in the [maintainers document](Maintainers.md). A "committer" is anyone
+with direct write access to the Open Enclave repository on GitHub, as granted by
+the maintainers committee. All maintainers are committers, but not all
+committers are maintainers. Finally, "contributor" is anyone else making
+contributions to the project, including: creating or commenting on issues,
+opening or reviewing pull requests, or other useful contributions such as
+providing support in forums or chats.
+
+See the [maintainers document](Maintainers.md) for the current Maintainers
```
CodeMonkeyLeet: Copyedit: Consider inlining this with the definition of the maintainer to preserve flow:

> We define a _maintainer_ as any member of the Open Enclave Maintainers Committee. The current members of this committee are listed in the [maintainers document](Maintainers.md), which also describes our process for adding new maintainers and committers. A _commiter_ is anyone granted direct write access to the Open Enclave repository on GitHub by the maintainers, so all maintainers are committers, but not all committers are maintainers. Finally, a _contributor_ is anyone engaged in improving the project, for example, by creating or commenting on issues, opening or reviewing pull requests, providing support in forums or chats, and so on.

docs/Maintainers.md
```diff
@@ -0,0 +1,50 @@
+Maintainers
+===========
+
+This document describes the Maintainers Committee of Open Enclave. By our
+liberal contribution policy outlined in our
+[governance model](GovernanceModel.md), maintainers are committers that are
+trusted to grant new committer rights, and grant new membership into the
+Maintainers Committee.
+
+When making decisions, the Maintainers Committee uses a "consensus seeking"
```
CodeMonkeyLeet: This process needs to be elaborated upon and it should be moved after the responsibilities of a maintainer are discussed to put the context of what decisions they should be making in context.

docs/Maintainers.md
```diff
@@ -0,0 +1,50 @@
...
+
+Maintainers Committee Members
+-----------------------------
+
+PENDING
+
+Responsibilities
+----------------
+
+The primary responsibility of the committee is to grant new committer rights
```
CodeMonkeyLeet: We should probably be more explicit about what the criteria is considered by the committee for adding committers.

docs/Maintainers.md
```diff
@@ -0,0 +1,50 @@
...
+(that is, write access to the main Open Enclave SDK repository or related
+repositories), and to grant new membership into the committee. Conversely, the
+committee must also remove committer rights and membership from those found to
+be violating the project's Code of Conduct or otherwise negatively affecting the
+project's community health.
+
+This committee is not intended to make every technical decision, as those should
+generally be made by agreement among committers as PRs are reviewed and merged.
+Where disagreements take place and need further resolution, those can be brought
+up with the Maintainers Committee as part of its responsibility to maintain the
```
CodeMonkeyLeet: There's a weird split here suggesting that the committee is not responsible for technical steering but is still required to be a tie breaker on technical issues. Is there a separate steering committee? If so, what's the relationship between the two committees? If we say that there is no steering committee and there is no technical direction from the maintainers, then what is expected of committers? Given that these docs define committers and maintainers in relation to each other, I don't think we can get away with not also detailing the responsibilities and expectations of committers.

docs/Maintainers.md
andschwa: Can you explain what part needs to be elaborated on?

docs/Maintainers.md
andschwa: Why? That's something the maintainers can discuss and agree upon when seeking consensus.

docs/Maintainers.md
andschwa: No, I was not defining a separate steering committee. Our current governance model is the liberal contribution policy, that is, technical decisions are made by collaboration among the most active committers. It's really only if they reach a deadlock that they might request a more formal intervention by the Maintainers Committee (and it happens at that point because it's become a community health issue).

docs/Maintainers.md
CodeMonkeyLeet: It's important for establishing transparency and trust in the process. How does this not come across as a kangaroo court if all of the initial maintainers are going to be from Microsoft?

docs/Maintainers.md
CodeMonkeyLeet: Is consensus being used as a synonym for unanimous agreement here? That's the implication given that the fallback is majority vote. It seems like a weirdly high initial bar.

andschwa: Here's the excerpt on the article I linked explaining why this bar/process works well:

> The TC uses a “consensus seeking” process for issues that are escalated to the TC. The group tries to find a resolution that has no open objections among TC members. If a consensus cannot be reached that has no objections then a majority wins vote is called. It is also expected that the majority of decisions made by the TC are via a consensus seeking process and that voting is only used as a last-resort.
The best solution tends to be the one everyone can agree to so you would think that consensus systems would be the norm. However, pure consensus systems incentivize obstructionism which we need to avoid.
In pure consensus everyone essentially has a veto. So, if I don’t want something to happen I’m in a strong position of power over everyone that wants something to happen. They have to convince me and I don’t have to convince anyone else of anything.
To avoid this we use a system called “consensus seeking” which has a long history outside of open source. It’s quite simple, just attempt to reach a consensus, if a consensus can’t be reached then call for a majority wins vote.
Just the fact that a vote is a possibility means that people can’t be obstructionists, whether someone favor a change or not, they have to convince their peers and if they aren’t willing to put in the work to convince their peers then they probably don’t involve themselves in that decision at all.
The way these incentives play out is pretty impressive. We started using this process in io.js and adopted it in Node.js when we merged into the foundation. In that entire time we’ve never actually had to call for a vote, just the fact that we could is enough to keep everyone working together to find a solution and move forward.

docs/Maintainers.md
andschwa: The key is that (hopefully) they won't be, but that's still under private discussion. Committers should be added with the least resistance possible. From the article:

> All contributors who land a non-trivial contribution should be on-boarded in a timely manner, and added as a committer, and be given write access to the repository.

This is explicitly to avoid the common open source issue of a project closely guarding its write access and causing the project to fail due to lack of influx of new committers. Fortunately, everyone I've reached out to become an official maintainer seems on-board with this philosophy (by virtue of that being our team and company's philosophy).

docs/Maintainers.md
CodeMonkeyLeet: You're using the phrase "liberal contribution policy" as if it is a widely understood phrase with a generally accepted set of conventions. Is it? Node.js has a set of principles that they follow, for example, but they quantify what it means to do this. For example, are we going to let every PR automatically merge by default if not reviewed? There needs to be some expectation that maintainers behave in a consistent fashion that is not going to be surprising to contributors submitting PRs in good faith (whether it's here or expanding the governance doc is an implementation detail).

andschwa: I'm sorry, I meant to be using it is a shared understanding of how we've already been developing: the active contributors in the project work together to merge PRs and so lead the technical direction of the project, without any formal voting or meeting to do so (that would be a "meritocracy."

This is a fairly common open source practice (to the point that it is detailed in [opensource.guide](https://opensource.guide/leadership-and-governance/#what-are-some-of-the-common-governance-structures-for-open-source-projects), which is what we gleaned our original docs from). Node.js has done an excellent job explaining how it works, and those principles are what we wrote as our intention to follow in our [governance docs](https://github.com/openenclave/openenclave/blob/master/docs/GovernanceModel.md). We explicitly chose _not_ to develop a formal voting process for (most) technical decisions.

andschwa:
I've updated the documentation and added the proposed list (all proposed members have consented to forming the initial committee).

I _would like_ sign-off by each of the proposed maintainers, and I would also like to expand on the name under "Company" to be team names instead, as we all come from a wide array across Microsoft.

If this document provides a sufficient framework, I suggest we merge it, and then I'll setup the first Community Maintenance Committee meeting, where further changes can be proposed, discussed, consensus reached, and changes made.

docs/GovernanceModel.md
```diff
@@ -6,16 +6,24 @@ is to enable a vibrant development community for the Open Enclave SDK. If you
...
 with us.
 
-Our model is based on the [liberal contribution
-policy](https://opensource.guide/leadership-and-governance/). We make decisions
-through a consensus seeking process, rather than a formal voting process.
+Our model is based on the
+[liberal contribution policy](https://opensource.guide/leadership-and-governance/).
+We make most decisions through a consensus seeking process, rather than a formal
+voting process. This means that contributions will be accepted once no
+committers object (and our PR guidelines have been met). If there is a major
```
dthaler: As phrased, this is problematic since it means you cannot accept a contribution unless you know that 0 objections are raised, so you have to slow down accepting contributions (e.g., PRs) until everyone has a chance to review, and have everyone as a reviewer on every PR, which I don't think it practical.

Also the "no committers object" contradicts Maintainers.md lines 10-14 which is better written.  That text is much better.  This text here is, in my opinion, in need of changes.

docs/Maintainers.md
```diff
@@ -0,0 +1,95 @@
...
+
+The primary responsibility of the Committee is to grant new committer rights
+(that is, write access to the main Open Enclave SDK repository or related
+repositories), and to grant new membership into the committee. Conversely, the
+Committee must also remove committer rights and membership from those found to
+be violating the project's Code of Conduct or otherwise negatively affecting the
+project's community health.
+
+This Committee is not intended to make every technical decision, as those should
+generally be made by agreement among committers as PRs are reviewed and merged.
```
dthaler: My experience with open source projects like Iotivity and AllJoyn is that a flat committee is a bad way to make technical decisions.  Instead, from experience it's better to assign a technical lead for each area, and they get to make the decision for that area, without having to wait for getting agreement from everyone else for every technical decision.   Policies/architecture for that area can be approved by the committee as the text states, but day to day technical decisions should be delegated to area leads if you want to maintain any degree of agility.

docs/Maintainers.md
```diff
@@ -0,0 +1,95 @@
...
+Where disagreements take place and need further resolution, those can be brought
+up with the Maintainers Committee as part of its responsibility to maintain the
```
dthaler: Lines 42 and 55 mention "Maintainers Committee".  What's the difference between that and the "Community Maintenance Committee"?

docs/Maintainers.md
```diff
@@ -0,0 +1,95 @@
...
+
+The following people have been granted commit permissions (that is, write
+access) to the Open Enclave SDK by the Community Maintenance Committee. The area
+column describes which technical areas each committer is most interested in, and
+therefore should usually be consulted for changes relating to that area.
+However, it is up to each committer to determine who should review which PR, and
+when to merge it. Remember that a PR must not be merged if a committer objects;
+instead, it should be brought up with the Community Maintenance Committee.
+
+| Name                  | Area                                      | GitHub Alias        |
```
dthaler: I think "General", "Core", and perhaps even "Security" are too broad and area not "Areas" per se.
Also from my experience with other open source projects, I would recommend having an area lead for "APIs and Language Bindings", and an area lead for "Platform Abstraction Layer (Host OS bindings and TEE bindings)".   And then allow them to designate a sub-maintainer per language binding, host OS binding, and TEE binding, if they want.

docs/Maintainers.md
andschwa: It should be "Community Maintenance Committee", good catch.

docs/Maintainers.md
andschwa: I agree, I just couldn't think of better names at the time 😂

docs/Maintainers.md
andschwa: That's why it's not making (most) technical decisions! We could later assign technical leads if we find it necessary, but we haven't had a need to do so yet.

docs/GovernanceModel.md
andschwa: I don't think this says or reads as "every committer needs to review every PR." I think our processes for properly reviewing PRs is (a) separate from this framework and (b) under improvement (see #1920). A committer objecting is a committer reviewing a PR and actually objecting; I don't think this reads a non-review as an objection.

How would you rephrase this to make that more clear?

andschwa: Just got it re-worded. Let me know what you think.

andschwa:
Thanks everyone!

bors r+

PR #2093, opened 2019-08-15
Add security policy and clean up license detection

CodeMonkeyLeet:
- Add `SECURITY.md` which is reflected in GitHub under the **Security** tab, **Policy** information.

- Move `.check-license.ignore` file under `/scripts` folder as the file was confusing [licensee](https://github.com/licensee/licensee) and causing GitHub not to display the MIT license information associated with the project.


CodeMonkeyLeet:
@andschwa Rephrased to take the feedback into account.

SECURITY.md
```diff
@@ -0,0 +1,25 @@
+# Security Policy
+
+## Supported Versions
+
+Open Enclave is still in the process of getting to a stable v1.0 release
+and so does not currently provide a long term support version. We make a
+good faith effort to address security issues on a regular cadence and
+release a minor version update approximately once every quarter. Users should
+expect to upgrade to the latest release version or master branch build to
+stay current on security updates.
```
johnkord: I think it's worth adding a sentence at the end of this paragraph saying something explicit about the table below, like: "Below you can find a table of the currently supported versions of Open Enclave"

CodeMonkeyLeet: Added a reference to the table.

PR #2096, opened 2019-08-15
Updating system PATH environment in registry.

### LINKED ISSUE #2095: Windows install script: System environment path is not updated in the registry
Currently the install-windows-preregs.ps1 updates the local PATH environment, therefore it is required to run this script every time in order to setup the correct PATH environment.  It would be nice to not have to do this every time since the script always downloads all the install binaries at the beginning.

jazzybluesea:
Previously when running the install-windows-prereqs.ps1 script, it would only update the system PATH environment of the local powershell session.  If you opened another x64 Native developer session the user would have to run the same script again in order to setup the PATH environment.

The install-windows-prereqs.ps1 script also downloads all the install binaries, which does not make sense to have it run every time.

Fixes #2095

### DIFF
--- a/scripts/install-windows-prereqs.ps1
+++ b/scripts/install-windows-prereqs.ps1
@@ -173,7 +173,11 @@ function Add-ToSystemPath {
     if(!$Path) {
         return
     }
-    $systemPath = [System.Environment]::GetEnvironmentVariable('Path', 'Machine').Split(';')
+    # Get system PATH environment variable from the registry
+    $regPathKey = "Registry::HKLM\System\CurrentControlSet\Control\Session Manager\Environment"
+    $regPathValue = (Get-ItemProperty -Path $regPathKey -Name PATH).Path
+
+    $systemPath = $regPathValue.Split(';')
     $currentPath = $env:PATH.Split(';')
     foreach($p in $Path) {
         if($p -notin $systemPath) {
@@ -184,6 +188,7 @@ function Add-ToSystemPath {
         }
     }
     $env:PATH = $currentPath -join ';'
+    Set-ItemProperty -Path $regPathKey -Name PATH -Value $systemPath
     setx.exe /M PATH ($systemPath -join ';')
     if($LASTEXITCODE) {
         Throw "Failed to set the new system path"


radhikaj:
> I have a couple of high level concerns with this:
> 
> * The Windows DLL search order prioritizes both the application folder as well as the _current_ folder over a folder in the system `PATH`. This makes it possible to perform DLL injection attacks purely in user mode even if the install location of the client DLL is locked down. The model of having the DLL be redistributed with the app (and thus loaded from the app folder which can be ACLed appropriately) is preferred.
> * We should not install the Azure DCAP client by default as part of setting up the OE SDK. This is an existing problem, but it is more invasive to modify the system configuration for it when it may not even be a desired component. This PR aside, we should make this an optional install anyway (i.e. default `AzureDCAPNupkgURL` to null and no-op if not provided a path) and update the Ansible script/manual install instructions to accommodate the option.
> * The Azure DCAP client package has its own install script, and behaves slightly differently from what we do here (it propagates the .dll to an install location of the user's choice, then paths that.) If we really want to "install" the client as part of this script, I would do so to that script instead of trying to keep this in sync with whatever the Azure DCAP client package might expect to do in the future.

On a system which has been configured to launch enclaves in FLC mode, the DCAP dll would be used for all enclave apps on the system. Therefore, putting it in the PATH would be useful to have since the DLL is not specific to the app. I am not sure how adding the DCAP dll to the system wide path, as mentioned below, makes the attack surface larger. Could you please elaborate?
> "This makes it possible to perform DLL injection attacks purely in user mode even if the install location of the client DLL is locked down."

I am in agreement with below. Changing the path in the DCAP client install script is a better solution since we change the path as we are installing DCAP client.
> The Azure DCAP client package has its own install script, and behaves slightly differently from what we do here (it propagates the .dll to an install location of the user's choice, then paths that.) If we really want to "install" the client as part of this script, I would do so to that script instead of trying to keep this in sync with whatever the Azure DCAP client package might expect to do in the future.

CodeMonkeyLeet:
In reply to @radhikaj:

> > "This makes it possible to perform DLL injection attacks purely in user mode even if the install location of the client DLL is locked down."
> 
> On a system which has been configured to launch enclaves in FLC mode, the DCAP dll would be used for all enclave apps on the system. Therefore, putting it in the PATH would be useful to have since the DLL is not specific to the app. I am not sure how adding the DCAP dll to the system wide path, as mentioned in your comment, makes the attack surface larger. Could you please elaborate?

A couple of baseline clarifications:
- This is regarding the Azure DCAP client, not the Intel DCAP library DLL.
- A Windows FLC system configured to use DCAP will have Intel's DCAP DLLs (PCE and QE enclaves) installed as part of their  _driver_ install. These are locked down with admin privileges.
- The Azure DCAP client DLL is distributed via nuget package without an installer.
    - It is not a requirement to use _Azure's_ DCAP client with Intel DCAP. Intel is expected to release their own DCAP client (as part of their PCK caching service package) which should be the baseline version that OE SDK defaults to. We need to keep the separation of OE SDK from Microsoft very clear. 
    - Even different apps on the same device may not use the same DCAP client, and conflicts will arise because Intel delay loads these by well-known filename. 
- Setting PATH to enable default pathing to the Azure DCAP client DLL is convenient, but not secure. 
    - The expectation is that the application has chosen a specific DCAP client implementation, which controls factors such as the caching lifetime of quote revocation collateral. If I can substitute the choice of your DCAP client, I can potentially make you think a quote is not revoked when it should be.
    - If relying on PATH for loading a shared system-wide DLL, I can trivially perform a replacement of the DCAP client DLL loaded by your app by launching it from a folder I control, and putting my malicious copy of a DCAP client DLL in that folder. The DLL search process will look in my launch folder, before it looks through all the folders in PATH.
    - To prevent this, an app should deploy its choice of its DCAP client in the same folder as the app, and lock that app folder down with admin permissions (e.g. installs to Program Files get this by default). The DLL search path prioritizes the local copy of the DLL preventing DLL injection, and the ACLs on the folder prevent someone from replacing the DLL without admin privileges.

radhikaj:
> In reply to @radhikaj:
> 
> > > "This makes it possible to perform DLL injection attacks purely in user mode even if the install location of the client DLL is locked down."
> > 
> > 
> > On a system which has been configured to launch enclaves in FLC mode, the DCAP dll would be used for all enclave apps on the system. Therefore, putting it in the PATH would be useful to have since the DLL is not specific to the app. I am not sure how adding the DCAP dll to the system wide path, as mentioned in your comment, makes the attack surface larger. Could you please elaborate?
> 
> A couple of baseline clarifications:
> 
> * This is regarding the Azure DCAP client, not the Intel DCAP library DLL.
> * A Windows FLC system configured to use DCAP will have Intel's DCAP DLLs (PCE and QE enclaves) installed as part of their  _driver_ install. These are locked down with admin privileges.
> * The Azure DCAP client DLL is distributed via nuget package without an installer.
>   
>   * It is not a requirement to use _Azure's_ DCAP client with Intel DCAP. Intel is expected to release their own DCAP client (as part of their PCK caching service package) which should be the baseline version that OE SDK defaults to. We need to keep the separation of OE SDK from Microsoft very clear.
>   * Even different apps on the same device may not use the same DCAP client, and conflicts will arise because Intel delay loads these by well-known filename.
> * Setting PATH to enable default pathing to the Azure DCAP client DLL is convenient, but not secure.
>   
>   * The expectation is that the application has chosen a specific DCAP client implementation, which controls factors such as the caching lifetime of quote revocation collateral. If I can substitute the choice of your DCAP client, I can potentially make you think a quote is not revoked when it should be.
>   * If relying on PATH for loading a shared system-wide DLL, I can trivially perform a replacement of the DCAP client DLL loaded by your app by launching it from a folder I control, and putting my malicious copy of a DCAP client DLL in that folder. The DLL search process will look in my launch folder, before it looks through all the folders in PATH.
>   * To prevent this, an app should deploy its choice of its DCAP client in the same folder as the app, and lock that app folder down with admin permissions (e.g. installs to Program Files get this by default). The DLL search path prioritizes the local copy of the DLL preventing DLL injection, and the ACLs on the folder prevent someone from replacing the DLL without admin privileges.



> In reply to @radhikaj:
> 
> > > "This makes it possible to perform DLL injection attacks purely in user mode even if the install location of the client DLL is locked down."
> > 
> > 
> > On a system which has been configured to launch enclaves in FLC mode, the DCAP dll would be used for all enclave apps on the system. Therefore, putting it in the PATH would be useful to have since the DLL is not specific to the app. I am not sure how adding the DCAP dll to the system wide path, as mentioned in your comment, makes the attack surface larger. Could you please elaborate?
> 
> A couple of baseline clarifications:
> 
> * This is regarding the Azure DCAP client, not the Intel DCAP library DLL.
> * A Windows FLC system configured to use DCAP will have Intel's DCAP DLLs (PCE and QE enclaves) installed as part of their  _driver_ install. These are locked down with admin privileges.
> * The Azure DCAP client DLL is distributed via nuget package without an installer.
>   
>   * It is not a requirement to use _Azure's_ DCAP client with Intel DCAP. Intel is expected to release their own DCAP client (as part of their PCK caching service package) which should be the baseline version that OE SDK defaults to. We need to keep the separation of OE SDK from Microsoft very clear.
>   * Even different apps on the same device may not use the same DCAP client, and conflicts will arise because Intel delay loads these by well-known filename.
> * Setting PATH to enable default pathing to the Azure DCAP client DLL is convenient, but not secure.
>   
>   * The expectation is that the application has chosen a specific DCAP client implementation, which controls factors such as the caching lifetime of quote revocation collateral. If I can substitute the choice of your DCAP client, I can potentially make you think a quote is not revoked when it should be.
>   * If relying on PATH for loading a shared system-wide DLL, I can trivially perform a replacement of the DCAP client DLL loaded by your app by launching it from a folder I control, and putting my malicious copy of a DCAP client DLL in that folder. The DLL search process will look in my launch folder, before it looks through all the folders in PATH.
>   * To prevent this, an app should deploy its choice of its DCAP client in the same folder as the app, and lock that app folder down with admin permissions (e.g. installs to Program Files get this by default). The DLL search path prioritizes the local copy of the DLL preventing DLL injection, and the ACLs on the folder prevent someone from replacing the DLL without admin privileges.

@CodeMonkeyLeet Thanks for the explanation. I can see the advantages of including it with the app.

jazzybluesea:
I think at the very least we should document that install-windows-prereqs.ps1 updates the shell's PATH environment and it will not persist.  In other words, the script has to be run every time in order to setup the correct PATH and be able to build.

dthaler:
Re "The Windows DLL search order prioritizes both the application folder as well as the current folder over a folder in the system PATH."   That hasn't been true since before XP SP2.  See SafeDllSearchMode at https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order

CodeMonkeyLeet:
@dthaler, thanks for the link, but I'm unclear on what the contradiction is? Per the article:

> If **SafeDllSearchMode** is enabled, the search order is as follows:
> 
> 1. The directory from which the application loaded.
> 2. The system directory. Use the [GetSystemDirectory](https://docs.microsoft.com/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getsystemdirectorya) function to get the path of this directory.
> 3. The 16-bit system directory. There is no function that obtains the path of this directory, but it is searched.
> 4. The Windows directory. Use the [GetWindowsDirectory](https://docs.microsoft.com/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getwindowsdirectorya) function to get the path of this directory.
> 5. The current directory.
> 6. The directories that are listed in the PATH environment variable. Note that this does not include the per-application path specified by the **App Paths** registry key. The **App Paths** key is not used when computing the DLL search path.

Folders in PATH have always been the lowest priority and remain so after the introduction of SafeDllSearchMode. The primary distinction is that the current folder was prioritized below the `System` and `Windows` folders. Is that not your understanding?

dthaler:
Sorry I misread "system directory" (bullet 2) instead of "system PATH" which you said.

dthaler:
I want to +1 something Simon said "We should not install the Azure DCAP client by default as part of setting up the OE SDK. "   I'll go even father and opine that I think that the OE project should not contain any Azure-specific code.  Instead, Azure-specific code should in my view be in a separate project that depends on the OE SDK.  So I would prefer deleting all Azure DCAP references from the main OE SDK.

jazzybluesea:
#2106 updates the install script to use setx to update the PATH environment.  Closing this PR.

PR #2097, opened 2019-08-15
Installing windows nuget prereqs under the repo directory

### LINKED ISSUE #2094: Windows install script and host/CMakeLists.txt SGX dependency mismatch
Currently the windows install script installs the nuget dependencies under system drive, but host/CMakeLists.txt expects these SGX dependencies to be under the CMake sourcedirectory.

install-windows-prereqs.ps1:
$OE_NUGET_DIR = Join-Path ${env:SystemDrive} "openenclave\prereqs\nuget"

host/CMakeLists.txt
if (WIN32)
    set(LIBPATHS
      ${CMAKE_SOURCE_DIR}/prereqs/nuget/EnclaveCommonAPI/lib/native/x64-Release
      ${CMAKE_SOURCE_DIR}/prereqs/nuget/DCAP_Components/build/lib/native/Libraries)
    set(INCPATHS
      "${CMAKE_SOURCE_DIR}/prereqs/nuget/EnclaveCommonAPI/Header Files"
      "${CMAKE_SOURCE_DIR}/prereqs/nuget/DCAP_Components/build/Header Files")
    set(WINSYSLOCATION
       $ENV{WINDIR}/System32)
  endif ()

jazzybluesea:
Previously, when building with -DUSE_LIBSGX=1 cmake would failed because it was not able to find the SGX library files.  The windows install script install-windows-prereqs.ps1 was installing the dependencies under <windows system drive>:\openenclave\prereqs.  This change resolves this issue.

Fixes #2094

### DIFF
--- a/scripts/install-windows-prereqs.ps1
+++ b/scripts/install-windows-prereqs.ps1
@@ -2,6 +2,7 @@
 # Licensed under the MIT License.
 
 Param(
+    [Parameter(Mandatory=$true)][string]$nugetInstallPath,
     [string]$GitURL = 'https://github.com/git-for-windows/git/releases/download/v2.19.1.windows.1/Git-2.19.1-64-bit.exe',
     [string]$SevenZipURL = 'https://www.7-zip.org/a/7z1806-x64.msi',
     [string]$VSBuildToolsURL = 'https://aka.ms/vs/15/release/vs_buildtools.exe',
@@ -19,7 +20,8 @@ Param(
 $ErrorActionPreference = "Stop"
 
 $PACKAGES_DIRECTORY = Join-Path $env:TEMP "packages"
-$OE_NUGET_DIR = Join-Path ${env:SystemDrive} "openenclave\prereqs\nuget"
+$OE_NUGET_DIR = $nugetInstallPath
+Write-Host "Installing nuget dependencies to $OE_NUGET_DIR"
 
 $PACKAGES = @{
     "git" = @{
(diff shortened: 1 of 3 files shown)


scripts/install-windows-prereqs.ps1
```diff
@@ -2,6 +2,7 @@
 # Licensed under the MIT License.
 
 Param(
+    [Parameter(Mandatory=$true)][string]$installPath,
```
CodeMonkeyLeet: This will break the Ansible openenclave/scripts/ansible/oe-windows-acc-setup.yml and that needs to be updated as well.

jazzybluesea: What should be used for Ansible?

CodeMonkeyLeet: Kind of, I would use the original hardcoded value: `${env:SystemDrive} "openenclave"`

jazzybluesea: I think is best to NOT make the parameter required in order to avoid the ansible specify the path.  I will instead default it in the script.

jazzybluesea:
I am going to create a different PR.  My current branch is was branched off the wrong branch.

scripts/install-windows-prereqs.ps1
jazzybluesea: Resolved.

jazzybluesea:
Closing. addressed by PR #2106

PR #2100, opened 2019-08-16
v0.6.x ansible changes

achamayou:
We find the playbooks bundled with OpenEnclave tremendously useful to get the right OpenEnclave dependencies without unnecessary duplication. We've been using them (along with playbooks of our own) to keep our development, test and CI VMs up to date.

As we're moving the last parts of our CI that weren't using containers to docker, we've noticed that there wasn't a variant of the playbooks targeted at containers (which typically won't have kernel headers and where the driver install does not make sense).

This PR add "no-driver" variants of the contributor and linux setups, which are aimed at containers but otherwise keep all other dependencies in sync. The validation task is split to allow that.

We'd like to get this into master as well if possible, but we're targeting 0.6 for now, and this seemed like a good place to start. What do you think?

### DIFF
--- a/scripts/ansible/oe-contributors-acc-setup-no-driver.yml
+++ b/scripts/ansible/oe-contributors-acc-setup-no-driver.yml
@@ -1,0 +1,19 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+---
+- hosts: localhost
+  any_errors_fatal: true
+  become: yes
+  tasks:
+    - import_role:
+        name: linux/openenclave
+        tasks_from: environment-setup.yml
+
+    - import_role:
+        name: linux/intel
+        tasks_from: sgx-packages.yml
+
+    - import_role:
+        name: linux/az-dcap-client
+        tasks_from: stable-install.yml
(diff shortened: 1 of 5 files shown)


andschwa:
My quick feedback from my phone would be that I’d want them in master first and foremost, and then if you really needed them backported to v0.6 we could do that if needed.

Also, thank you!

And I’ll get you whitelisted with the CLA not. I’m still waiting for an upstream fix to let me whitelist other organizations (e.g. Microsoft).

achamayou:
@andschwa thank you for the feedback, I'd signed the CLA before, but I've signed it again now.

I will open a separate PR against master with these changes. I think we would ultimately like to have them in v0.6.x, ideally in 0.6.1 or something similar. We ship our own setup playbooks, with OpenEnclave as a dependency, so users can setup their VMs and containers, and I would like to point to a stable release of OE there rather than master.

There may be other, less cumbersome ways to achieve this until 0.7 though. I'll try to think of something else :)

andschwa:
If you had signed it before we transitioned to the openenclave organization, and not opened a PR since, that would explain why it's asking again. However, you _are_ in the OE GitHub org...so you should have already been whitelisted, but CLA Assistant hasn't been perfect. 🤷‍♂

achamayou:
@andschwa it's quite possibly what happened, but in any case, it's signed again now!

I've opened https://github.com/openenclave/openenclave/pull/2103, which is the same changes against master.

achamayou:
@johnkord @andschwa thank you both for the super quick response, it's much appreciated! Since #2103 is now in master, I am merging this, which is effectively just backporting the changes from there.

PR #2103, opened 2019-08-16
Container-focused variants of existing playbooks

achamayou:
Spun out of https://github.com/openenclave/openenclave/pull/2100, against master this time.

We find the playbooks bundled with OpenEnclave tremendously useful to get the right OpenEnclave dependencies without unnecessary duplication. We've been using them (along with playbooks of our own) to keep our development, test and CI VMs up to date.

As we're moving the last parts of our CI that weren't using containers to docker, we've noticed that there wasn't a variant of the playbooks targeted at containers (which typically won't have kernel headers and where the driver install does not make sense).

This PR add "no-driver" variants of the contributor and linux setups, which are aimed at containers but otherwise keep all other dependencies in sync. The validation task is split to allow that.

### DIFF
--- a/scripts/ansible/oe-contributors-acc-setup-no-driver.yml
+++ b/scripts/ansible/oe-contributors-acc-setup-no-driver.yml
@@ -1,0 +1,19 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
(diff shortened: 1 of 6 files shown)


achamayou:
@johnkord thank you for the quick review!

Issue #2099, opened 2019-08-16
Shared Memory in OE


mikbras:
This issue proposes adding shared memory to Open Enclave. **Shared memory** is memory which is accessible to both the **host** and the **enclave** across all supported TEEs. 

In **SGX**, **all** host memory is shared **implicitly** with the enclave (stack, heap, global data, and code pages). In other words,  the host needn't do anything to share its memory with the enclave. 

In **TrustZone**, **designated** host memory is shared **explicitly** with the enclave (through an explicit function call). Some TrustZone TEEs may support shared memory, while others may not. On Windows, for example, the driver does not currently support shared memory. There may also be TrustZone devices which do not support shared memory, although we are uncertain.

Towards deriving a common memory model, we define three types of memory.
- **enclave memory** -- accessible only to the enclave.
- **shared memory** -- accessible to both the host and the enclave.
- **host memory** -- any non-shared host memory (stack, heap, data segments, code segments).

Host memory is problematic since it is accessible within SGX enclaves but inaccessible within TrustZone enclaves. To promote compatibility across SGX, TrustZone, and future TEEs, the concept of host memory should be abandoned in favor of shared memory. This implies formal mechanisms for designating, allocating, and freeing shared memory on both the host-side and enclave-side. In all, the following operations are needed with suggested function names.

1. Checking whether shared memory is supported (**oe_has_shared_memory**).
2. Registering a region of shared memory (**oe_register_shared_memory**).
3. Unregistering a region of shared memory (**oe_unregister_shared_memory**).
4. Allocating a block of shared memory (**oe_allocate_shared_memory**).
5. Freeing a block of shared memory (**oe_free_shared_memory**).

Again, these operations must be supported by the host and the enclave. The first parameter of all host functions will be an enclave instance. This implies that a host maintains a shared memory pool for each enclave instance.

Open Enclave supports the following enclave functions to determine whether a memory object falls entirely within trusted or untrusted memory.

- **oe_is_within_enclave** -- returns true if memory object is entirely within enclave memory.
- **oe_is_outside_enclave** -- returns true if the memory object is entirely outside enclave memory.

Note that both functions may return false for a given memory object (since the memory object may be partially within enclave memory). Note that in SGX, the enclave does not know the boundaries of host memory; it only knows its own boundaries.

The addition of shared memory leads to the controversial question of whether a third memory check is needed, perhaps something called **oe_is_within_shared_memory**. This function would check whether the entire memory objects falls within shared memory.

[...]

PR #2104, opened 2019-08-16
The first installment of switchless ocalls.

jxyang:
* The stack-based shared memory (shm) allocation
* For switchless ocalls, parameters are passed in a buffer allocated out of shm
* However they are still routed to regular ocalls (will fix in the 2nd installment) for testing purpose
* Added a test for multi-threaded switchless ocalls
* Remove/modify a couple of existing tests that assumed switchless ocalls are not supported
* Change the keyworld denotation for switchless functions from "transition_using_threads" to "switchless"


andschwa:
Haven't gotten to look at the rest, but please, fix your committership 😄 (Right now these are from "Ubuntu")

tests/oeedger8r/behavior/switchless_trusted.edl
```diff
@@ -4,6 +4,6 @@
 enclave {
     trusted {
         // Switchless functions not supported.
```
andschwa: But now maybe supported?!

tests/switchless/enc/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+
+oeedl_file(../switchless.edl enclave gen)
```
andschwa: Please don't use `oeedl_file` in new code, you'll run into dependency issues. See the deep-copy tests for up-to-date examples.

tools/oeedger8r/intel/Lexer.mll
```diff
@@ -88,7 +88,7 @@ rule tokenize = parse
   | "import"     { Timport }
   | "allow"      { Tallow }
   | "public"     { Tpublic }
-  | "transition_using_threads"       { Tswitchless }
+  | "switchless"       { Tswitchless }
```
andschwa: Hm...why are we diverging from Intel's EDL here?

tools/oeedger8r/Emitter.ml
```diff
@@ -493,11 +493,6 @@ let gen_enclave_code (ec : enclave_content) (ep : edger8r_params) =
         printf
           "Warning: Function '%s': Reentrant ocalls are not supported by Open \
            Enclave. Allow list ignored.\n"
-          f.uf_fdecl.fname ;
-      if f.uf_is_switchless && not ep.experimental then
-        failwithf
-          "Function '%s': switchless ecalls and ocalls are not yet supported \
-           by Open Enclave SDK."
           f.uf_fdecl.fname )
```
andschwa: Why don't I see any other edger8r changes...is there nothing to change in the generated code for supporting switchless? That would be...unexpected (to me anyway).

tools/oeedger8r/Emitter.ml
```diff
@@ -493,11 +493,6 @@ let gen_enclave_code (ec : enclave_content) (ep : edger8r_params) =
         printf
           "Warning: Function '%s': Reentrant ocalls are not supported by Open \
            Enclave. Allow list ignored.\n"
-          f.uf_fdecl.fname ;
-      if f.uf_is_switchless && not ep.experimental then
```
andschwa: I think we should probably keep switchless ocalls in the experimental stage, and emit a failure if not `--experimental` still.

tools/oeedger8r/Emitter.ml
jxyang: Most edger8r changes are already merged in this PR: https://github.com/openenclave/openenclave/pull/2031/files. Thanks to Anand.

tools/oeedger8r/intel/Lexer.mll
jxyang: There was a discussion during the design review meeting. People doesn't like the term `transition_using_threads`. To me personally, the term is confusing and too verbose. We think `switchless` is a more intuitive term.

Do we need to follow Intel's EDL literally?

tests/oeedger8r/behavior/switchless_trusted.edl
jxyang: Not yet. We are supporting switchless ocalls first.

tools/oeedger8r/Emitter.ml
jxyang: I agree. Though I think by the time switchless call is shipped (target 0.7), we should remove the experimental label.

tools/oeedger8r/intel/Lexer.mll
radhikaj: Based on the design review, lets go with "switchless". Thanks

enclave/core/shm.c
```diff
@@ -0,0 +1,103 @@
...
+
+    // check for capacity
+    size_t used_after;
+    OE_CHECK(oe_safe_add_sizet(shm.used, total_size, &used_after));
+
+    if (used_after <= shm.capacity)
+    {
+        shm.used = used_after;
+        return shm.buffer + shm.used - total_size;
+    }
```
radhikaj: else condition here?

enclave/core/shm.c
```diff
@@ -0,0 +1,103 @@
...
+    size_t total_size = oe_round_up_to_multiple(size, ALIGNMENT);
+
+    // check for overflow
+    OE_CHECK(total_size < size);
+
+    // check for capacity
+    size_t used_after;
+    OE_CHECK(oe_safe_add_sizet(shm.used, total_size, &used_after));
+
+    if (used_after <= shm.capacity)
```
radhikaj: Would it be possible to add  a comment to explain this logic? I am not sure I understand it.

enclave/core/shm.c
```diff
@@ -0,0 +1,103 @@
...
+    {
+        shm.used = used_after;
+        return shm.buffer + shm.used - total_size;
+    }
+
+done:
+    return NULL;
```
radhikaj: Do we always want to return NULL?

tests/switchless/enc/enc.c
```diff
@@ -0,0 +1,69 @@
...
+#include <openenclave/corelibc/string.h>
+#include <openenclave/enclave.h>
+#include <openenclave/internal/print.h>
+#include "switchless_t.h"
+
+char* oe_host_strdup(const char* str)
+{
+    size_t n = oe_strlen(str);
+
+    char* dup = (char*)oe_host_calloc(1, n + 1);
```
radhikaj: Since oe_shm_malloc alwaya returns null, dup is always null. Is that right?

tools/oeedger8r/Emitter.ml
andschwa: Oh yeah I remember that now :joy:

tools/oeedger8r/Emitter.ml
andschwa: @anakrish and I had wanted all edger8r features (and indeed, most features) to go through one release in `--experimental`, so that they're thoroughly tested. So we should target that in v0.8.

andschwa: @radhikaj What do you think?

tools/oeedger8r/intel/Lexer.mll
andschwa: @radhikaj Who all agreed on that? I think that design review happened while I was out of town for the FCA hackathon.

I would _highly recommend_ we not diverge from Intel's syntax, as our original intent was to be as compatible with Intel EDL as possible. This is an unnecessary divergence that's being made just for cosmetic reasons, so it doesn't seem worth it to me.

@mikbras were you able to attend that design meeting? I am looking for minutes right now...

jxyang: @andschwa Mike is on vacation. Let's continue the discussion through emails.

tools/oeedger8r/Emitter.ml
radhikaj: We will ship switchless calls in V0.7.   We should drop the experimental by then if it is sufficiently tested. We cannot target V0.8 .

tools/oeedger8r/intel/Lexer.mll
radhikaj: > @radhikaj Who all agreed on that? I think that design review happened while I was out of town for the FCA hackathon.
> 
> I would _highly recommend_ we not diverge from Intel's syntax, as our original intent was to be as compatible with Intel EDL as possible. This is an unnecessary divergence that's being made just for cosmetic reasons, so it doesn't seem worth it to me.
> 
> @mikbras were you able to attend that design meeting? I am looking for minutes right now...

@andschwa Attendees of the meeting agreed to it. @mikbras was in the meeting as well. Thanks @jxyang for starting a thread.

enclave/core/shm.c
jxyang: If we still has enough capacity as `if (used_after <= shm.capacity)`, it returns the pointer to the allocated memory.

tests/switchless/enc/enc.c
jxyang: Hmmm no. See above.

enclave/core/shm.c
```diff
@@ -0,0 +1,103 @@
...
+    return true;
+}
+
+void* oe_shm_malloc(size_t size)
+{
+    oe_result_t result = OE_UNEXPECTED;
+
+    if (shm.buffer == NULL)
+    {
+        shm.buffer = (uint8_t*)oe_reserve_shm(capacity);
```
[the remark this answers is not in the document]
jxyang: Good catch. Will do.

tests/switchless/enc/CMakeLists.txt
jxyang: Good suggestion. Will do.

tests/switchless/enc/enc.c
radhikaj: @jxyang thanks for the explanation

enclave/core/shm.c
radhikaj: @jxyang sorry I missed it. Thanks for pointing it out.

tools/oeedger8r/Emitter.ml
andschwa: That means we will skip a beta release cycle for the feature 😬 that scares me!

jxyang:
bors try

jxyang:
bors try

jxyang:
bors try

dthaler:
Issue #2099 tracks the discussion of shared memory in OE.

tools/oeedger8r/intel/Lexer.mll
dthaler: FWIW, I have no idea what "switchless" means and so don't think that term is intuitive.

jxyang:
Addressed concerns with the latest changes.

jxyang:
bors +

jxyang:
bors +

tests/switchless/enc/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+add_custom_command(
+  OUTPUT switchless_t.h switchless_t.c switchless_args.h
+  DEPENDS ../switchless.edl
```
andschwa: You'll want `edger8r` in the `DEPENDS` list otherwise when developing the edger8r you run into dependency issues 😅

jxyang: Noted.

PR #2106, opened 2019-08-16
Update install-windows-prereqs.ps1 to use Intel DCAP 1.2, PSW 2.4 & integrity checks for downloads

### LINKED ISSUE #2094: Windows install script and host/CMakeLists.txt SGX dependency mismatch

### LINKED ISSUE #2101: Perform integrity checks on downloaded blobs in install-windows-prereqs.ps1
We download executable and archive files in install-windows-prereqs.ps1 without validating their integrity, and a few of the URI endpoints are http. This needs to be changed so that we validate the integrity of these downloaded files against pre-computed measurements embedded in the script.

johnkord:
Fixes #2101
Fixes #2094


johnkord:
The order of installing the nuget pkgs from the new Intel DCAP drop with the Azure DCAP Client was a pain to get working. I think it might have to do with the EnclaveCommonAPI package having a different package version than the version in the package's filename, but honestly I'm not entirely sure :).

In any case, the new order of installing in this PR scripts works on a new Windows Server 2016 deployment.

This PR also adds integrity checking using Get-FileHash for each fo the downloaded packages. Fixes #2101

docs/GettingStartedDocs/GettingStarted.Windows.md
```diff
@@ -233,9 +233,10 @@ The following summary will assume that the contents were extracted to `C:\Intel
 4. Install the DCAP nuget packages:
     - The standalone `nuget.exe` [CLI tool](https://dist.nuget.org/win-x86-commandline/latest/nuget.exe) can be used to do this from the command prompt:
       ```cmd
-      nuget.exe install EnclaveCommonAPI -ExcludeVersion -Source "C:\Intel SGX DCAP for Windows v1.2.100.49925\nuget" -OutputDirectory C:\openenclave\prereqs\nuget
       nuget.exe install DCAP_Components -ExcludeVersion -Source "C:\Intel SGX DCAP for Windows v1.2.100.49925\nuget" -OutputDirectory C\openenclave\prereqs\nuget
+      nuget.exe install EnclaveCommonAPI -ExcludeVersion -Source "C:\Intel SGX DCAP for Windows v1.2.100.49925\nuget" -OutputDirectory C:\openenclave\prereqs\nuget
```
jazzybluesea: Could we please default the nuget output directory and have an option for the user to overwrite it?  cmake of host/CMakeLists.txt will fail because it expects the SGX libraries to be under <depot root>\prereqs\nuget

johnkord: Thanks for the comment @jazzybluesea ! What do you mean in this case by "default"?  As in, what do you suggest needs to change?

I see what you mean by host/CMakeLists.txt expecting these nuget packages to be in  ${CMAKE_SOURCE_DIR}/prereqs/nuget/… , but I'm not sure what you mean by default here!

docs/GettingStartedDocs/GettingStarted.Windows.md
```diff
@@ -233,9 +233,10 @@ The following summary will assume that the contents were extracted to `C:\Intel
...
       ```
+    - *Note:* EnclaveCommonAPI should be installed as the *very last* nuget package as a temporary workaround for a dependency issue.
 
 ##### [Azure DCAP client for Windows](https://github.com/Microsoft/Azure-DCAP-Client/tree/master/src/Windows) [optional]
```
CodeMonkeyLeet: I would move this entire optional section before the DCAP libs install section to preserve the order of nuget installations that work. We should probably also add a comment that it's a temporary workaround for the nuget package issue for EnclaveCommonAPI as in the DCAP section.

docs/GettingStartedDocs/GettingStarted.Windows.md
CodeMonkeyLeet: This is not a script but a set of manual instructions. Earlier in this doc, it has been established that `C:\openenclave` is the example root folder that the GitHub project was cloned into, I'm not sure anything needs to change here. 

If this comment should be tagged to install-windows-prereqs.ps1 file instead, then I would defer that to #2097 which focuses on the parameterization of this.

scripts/ansible/oe-windows-acc-setup.yml
```diff
@@ -8,32 +8,55 @@
...
-        vs_buildtools_2017: "https://oejenkins.blob.core.windows.net/oejenkins/vs_buildtools_2017.exe"
-        ocaml_url:          "https://oejenkins.blob.core.windows.net/oejenkins/ocpwin64-20160113-4.02.1+ocp1-mingw64.zip"
-        clang7_url:         "https://oejenkins.blob.core.windows.net/oejenkins/LLVM-7.0.1-win64.exe"
-        shellcheck_url:     "https://oejenkins.blob.core.windows.net/oejenkins/shellcheck-stable.exe"
-        nuget_url:          "https://oejenkins.blob.core.windows.net/oejenkins/nuget.exe"
-        devcon_package_url: "https://oejenkins.blob.core.windows.net/oejenkins/devcon_package.cab"
-        az_dcap_client_url: "https://oejenkins.blob.core.windows.net/oejenkins/Microsoft.Azure.DCAP.Client.1.0.0.nupkg"
+        intel_psw_2_2_url:   "https://oejenkins.blob.core.windows.net/oejenkins/intel_sgx_win_2.2.100.47975_PV.zip"
+        intel_psw_2_2_hash:  "EB479D1E029D51E48E534C284FCF5CCA3A937DA43052DCB2F4C71E5F354CA623"
+        intel_psw_2_3_url:   "https://oejenkins.blob.core.windows.net/oejenkins/Intel_SGX_PSW_for_Windows_v2.3.100.49777.exe"
```
CodeMonkeyLeet: This hasn't updated CI/CD to use PSW 2.4 and DCAP 1.2.

docs/GettingStartedDocs/GettingStarted.Windows.md
johnkord: We sorted it out, adding some changes to address this! Thanks Sergio! :)

scripts/install-windows-prereqs.ps1
```diff
@@ -3,19 +3,38 @@
...
+    [string]$IntelDCAPHash = 'F31E4451CA32E19CA3DCB0AFC49AFE9F4963C47BF62AAF24A8AE436BDA14FD8B',
     [string]$VCRuntime2012URL = 'https://download.microsoft.com/download/1/6/B/16B06F60-3B20-4FF2-B699-5E9B7962F9AE/VSU_4/vcredist_x64.exe',
-    [string]$AzureDCAPNupkgURL = 'https://oejenkins.blob.core.windows.net/oejenkins/Microsoft.Azure.DCAP.Client.1.0.0.nupkg' # TODO: Update this to official link once this is available
+    [string]$VCRuntime2012Hash = '681BE3E5BA9FD3DA02C09D7E565ADFA078640ED66A0D58583EFAD2C1E3CC4064',
+    [string]$AzureDCAPNupkgURL = 'https://oejenkins.blob.core.windows.net/oejenkins/Microsoft.Azure.DCAP.Client.1.0.0.nupkg', # TODO: Update this to official link once this is available
+    [string]$AzureDCAPNupkgHash = '152ACE956348E80E533E63E6CB1D3CA20E4CA7DC775FC0B9413F552368F971D6',
+    [switch]$WithoutFLC
 )
 
+if ($WithoutFLC)
```
CodeMonkeyLeet: It seems like this flag should also control skipping the installation of the Azure DCAP client package.

More generally, I'd like to ensure that the installation of Azure specific dependencies to be user-specified and not by default.

docs/GettingStartedDocs/GettingStarted.Windows.md
johnkord: (Also thanks Simon for your insight!)

scripts/install-windows-prereqs.ps1
```diff
@@ -449,26 +490,27 @@ function Install-DCAPDrivers {
     $TEMP_NUGET_DIR = "$PACKAGES_DIRECTORY\Azure_DCAP_Client_nupkg"
```
CodeMonkeyLeet: Probably useful to add a comment to point out that the ordering of nuget installs is important to preserve here until Intel fixes the package versioning.

docs/GettingStartedDocs/GettingStarted.Windows.md
jazzybluesea: Thank you John!

johnkord:
@jazzybluesea Do you have any other concerns with these changes? I'd be glad to merge them once I get a signoff :) 

Also @oprinmarius do you have any concerns about these changes? I need to test the Windows ansible changes, and I'm not exactly sure how.

docs/GettingStartedDocs/GettingStarted.Windows.md
```diff
@@ -33,7 +33,7 @@ To deploy all the prerequisities for building Open Enclave, you can run the ```s
 
 ```powershell
 cd scripts
-.\install-windows-prereqs.ps1
+.\install-windows-prereqs.ps1 -InstallPath YOUR_WORKSPACE_PATH_HERE -WithFLC $true -WithAzureDCAPClient $true
```
jazzybluesea: Suggestion, instead of -WithAzureDCAPClient perhaps something like: -WithDCAPClient <type>, where type could be: Azure, Intel, etc..?

scripts/ansible/oe-windows-acc-setup.yml
```diff
@@ -8,32 +8,58 @@
...
-        -GitURL            "{{ git_url }}"
-        -SevenZipURL       "{{ seven_zip_url }}"
-        -VSBuildToolsURL   "{{ vs_buildtools_2017 }}"
-        -OCamlURL          "{{ ocaml_url }}"
-        -Clang7URL         "{{ clang7_url }}"
-        -ShellCheckURL     "{{ shellcheck_url }}"
-        -NugetURL          "{{ nuget_url }}"
-        -DevconURL         "{{ devcon_package_url }}"
-        -AzureDCAPNupkgURL "{{ az_dcap_client_url }}"
+        -InstallPath         "C:\openenclave"
```
jazzybluesea: Is this a required option?

scripts/install-windows-prereqs.ps1
```diff
@@ -449,26 +497,31 @@ function Install-DCAPDrivers {
...
+        Throw "Failed to install nuget EnclaveCommonAPI"
     }
-    & "$PACKAGES_DIRECTORY\nuget.exe" install 'Microsoft.Azure.DCAP.Client' -Source "$TEMP_NUGET_DIR;nuget.org" -OutputDirectory "$OE_NUGET_DIR" -ExcludeVersion
+    & "$PACKAGES_DIRECTORY\nuget.exe" install 'EnclaveCommonAPI' -Source "$TEMP_NUGET_DIR;nuget.org" -OutputDirectory "$OE_NUGET_DIR" -ExcludeVersion
     if($LASTEXITCODE -ne 0) {
-        Throw "Failed to install nuget Microsoft.Azure.DCAP.Client"
+        Throw "Failed to install nuget EnclaveCommonAPI"
     }
 
     New-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\sgx_lc_msr\Parameters" -Name "SGX_Launch_Config_Optin" -Value 1 -PropertyType DWORD -Force
```
jazzybluesea: Could we please add a comment here?

scripts/install-windows-prereqs.ps1
```diff
@@ -3,71 +3,109 @@
 
 Param(
```
jazzybluesea: Could we please add a comment on how the hash values are generated so that someone else can update if needed? I am a little concerned on how maintainable is this.

docs/GettingStartedDocs/GettingStarted.Windows.md
johnkord: Well, there's only one right now :) If Intel or some other vendor produces one, then we can do this!

scripts/install-windows-prereqs.ps1
johnkord: The hash values are calculated using Get-FileHash, which is called on each of the downloaded packages below to verify the known hashes for each package. 

I think I made an invalid assumption here that each of these downloaded files are tied to static URLs with specific versions. That isn't the case for some of these, like shellcheck and nuget.exe. I need to fix that!

For each downloaded file tied to a specific version (like Git-2.19.1-64-bit.exe), we should expect the contents of that file to not change. If a file like this, which has a specific version, does change from what we expect, then it is probable that the file has been changed maliciously.

As for the maintenance overhead of updating these packages to new versions, it's not much cost to calculate the SHA256 hash of a downloaded file alongside it. One could simply run: "curl -o myfile URL" followed by a "Get-FileHash myfile" to get the expected hash of that file.

I am strongly in favor of maintaining integrity checks of automated downloads of expected packages. Otherwise, someone who owns one of these third party endpoints could inject malicious code onto a customer's system without them knowing what was downloaded/installed. Moreover, some of these endpoints are not SSL/TLS, so a simple MITM attack is possible for those. Integrity checks are a must, in my opinion.

scripts/install-windows-prereqs.ps1
johnkord: What kind of comment do you mean? Above (line 510) there is a comment that says: "# Note: the ordering of nuget installs below is important to preserve here until the issue with the EnclaveCommonAPI nuget package gets fixed."

scripts/install-windows-prereqs.ps1
jazzybluesea: I think the integrity checks make sense.  Maybe as a future improvement, as an idea, we could create a script that generates all these hash values to a version controlled yml file.  This yml file would be used as an input to both ansible file and this install script.  Then all we have to do is to update this new script.

scripts/ansible/oe-windows-acc-setup.yml
```diff
@@ -8,32 +8,58 @@
...
+        intel_psw_2_4_url:   "https://oejenkins.blob.core.windows.net/oejenkins/Intel%20SGX%20PSW%20for%20Windows%20v2.4.100.51291.exe"
+        intel_psw_2_4_hash:  "79AE32E984B5511CE4BF7568403333F837FBCE7E8D5730271C5D68F55BBF251D"
+        intel_dcap_url:      "https://oejenkins.blob.core.windows.net/oejenkins/Intel%20SGX%20DCAP%20for%20Windows%20v1.2.100.49925.exe"
+        intel_dcap_hash:     "F31E4451CA32E19CA3DCB0AFC49AFE9F4963C47BF62AAF24A8AE436BDA14FD8B"
+        git_url:             "https://oejenkins.blob.core.windows.net/oejenkins/Git-2.19.1-64-bit.exe"
+        git_hash:            "5E11205840937DD4DFA4A2A7943D08DA7443FAA41D92CCC5DAFBB4F82E724793"
+        seven_zip_url:       "https://oejenkins.blob.core.windows.net/oejenkins/7z1806-x64.msi"
+        seven_zip_hash:      "F00E1588ED54DDF633D8652EB89D0A8F95BD80CCCFC3EED362D81927BEC05AA5"
+        vs_buildtools_2017:  "https://oejenkins.blob.core.windows.net/oejenkins/vs_buildtools_2017.exe"
+        vs_buildtools_hash:  "6F49872B04A0EAEDF5ED96AB25F7697062D81A419D45D9970E41784F31165BF2"
```
oprinmarius: vs_buildtools_hash I think it is a typo here , this parameter is being called as vs_buildtools_2017_hash few lines down. 
It is missing 2017 from the name

scripts/install-windows-prereqs.ps1
jazzybluesea: This is line 527 where the registry is getting updated.  Not sure what setting a value of 1 to SGX_Launch_Config_Optin does.

johnkord: Oh I see what you mean! I'm not sure :) I think it's how we enable certain parts of the FLC driver. I should find out where this is documented and add a link to it!

CodeMonkeyLeet: It's necessary as an administrator action to enable the FLC driver to run on Windows. See https://download.01.org/intel-sgx/dcap-1.2/windows/docs/Intel_SGX_DCAP_Windows_SW_Installation_Guide.pdf

johnkord: I've included a link to this document in a comment in the script. Thanks @CodeMonkeyLeet !

johnkord:
I believe I've addressed all concerns in this PR. Please let me know if I've missed anything!

johnkord:
Thanks for the approval, @jazzybluesea !
bors r+

PR #2110, opened 2019-08-16
Cleaned up error codes based off feedback in PR #2077.

### LINKED ISSUE #1044: Attestation: OE_VERIFY_FAILED was overloaded for many validation failures, need define new error codes for specific failure condition

jazzybluesea:
Code review changes from previous PR #2077.
- Fixed typo: From OE_VERIFY_FAILED_OES_CMAC_MISMATCH to OE_VERIFY_FAILED_AES_CMAC_MISMATCH.
- Renamed OE_QUOTE_ENCLAVE_IDENTITY_MRSIGNER_MISMATCH  to OE_QUOTE_ENCLAVE_IDENTITY_UNIQUEID_MISMATCH.
- Renamed QE_QUOTE_ENCLAVE_IDENTITY_ISV_PROD_ID_MISMATCH to QE_QUOTE_ENCLAVE_IDENTITY_PRODUCTID_MISMATCH.
- Removed SGX specific error code QE_QUOTE_ENCLAVE_IDENTITY_ISVSVN_UNEXPECTED.
- Removed unnecessary error check.

Fixes #1044


include/openenclave/bits/result.h
```diff
@@ -300,24 +300,19 @@ typedef enum _oe_result
     OE_QUOTE_ENCLAVE_IDENTITY_VERIFICATION_FAILED,
 
     /**
-     * QE identify mrsigner do not match.
+     * QE identify unique Id do not match.
```
CodeMonkeyLeet: Copyedit: Clean up the grammar and spelling for the comments e.g.  "identify" should be "identity". I would suggest:

- Failed to verify the identity of the quoting enclave.
- Unique ID of the quoting enclave does not match the expected value.
- Product ID of the quoting enclave does not match the expected value.
- AES CMAC of the report does not match the expected value.

jazzybluesea: Done.

jazzybluesea:
bors +r

PR #2113, opened 2019-08-21
Check if targets for crypto and dlib are defined

### LINKED ISSUE #1730: find_package(openenclave) fails the second time
```
CMake Error at /opt/openenclave/lib/openenclave/cmake/openenclave-config.cmake:50 (add_executable):
  add_executable cannot create imported target "openenclave::oeedger8r"
  because another target with the same name already exists
```

Repro:
```
find_package(openenclave)
find_package(openenclave)
```

EmilAlexandruStoica:
Fixes #1730 by checking if the targets for crypto and dlib are defined


letmaik:
Where's the regression test? :)

EmilAlexandruStoica:
@andschwa Do you want me to create another pull request with the changes that you specified in the comment from above ?

andschwa:
That'd be great 😄 Thank you!

PR #2116, opened 2019-08-21
Update to OP-TEE 3.6.0

HernanGatta:
This PR updates the SDK to utilize OP-TEE 3.6.0, the latest release. Previously, OP-TEE version 3.3.0 was assumed. This update helps both keep up to date with the goings-on in OP-TEE, but also to eventually upstream our changes to it.

The biggest changes are:
1. OP-TEE now honours the ELF entry point (`_start`);
2. The mechanism to obtain the enclave load address has changed.

**Note:** This will require an update to the full build container (`oetools-full-*`) to include newer devkits. That must happen before this PR can be merged. I'll coordinate with John.

### DIFF
--- a/3rdparty/optee/libutee/entry.c
+++ b/3rdparty/optee/libutee/entry.c
@@ -1,0 +1,32 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <compiler.h>
+#include <tee_ta_api.h>
+#include <utee_syscalls.h>
+
+struct utee_params;
+
+TEE_Result __utee_entry(
+    unsigned long func,
+    unsigned long session_id,
+    struct utee_params* up,
+    unsigned long cmd_id);
+
+void __ta_entry(
+    unsigned long func,
+    unsigned long session_id,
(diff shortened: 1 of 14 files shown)


3rdparty/optee/libutee/entry.c
```diff
@@ -0,0 +1,32 @@
...
+
+struct utee_params;
+
+TEE_Result __utee_entry(
+    unsigned long func,
+    unsigned long session_id,
+    struct utee_params* up,
+    unsigned long cmd_id);
+
+void __ta_entry(
```
dthaler: why do you need 16-20 instead of just using 22-26?

HernanGatta: GCC specifies that attributes are to be attached to function declarations.

I've done some experimentation removing the declaration and leaving only the definition:
1. If the attribute is placed at the end of a function definition, GCC emits an error;
2. If the attribute is placed at the beginning of a function definition, GCC does not complain.

However, all documentation points to attributes being expected to be placed on function declarations and not definitions.

HernanGatta:
bors try

PR #2117, opened 2019-08-22
General ARM improvements

HernanGatta:
This is effectively #2073 but without changes to memory-related functions, pending a resolution of the memory model question.

This PR implements `oe_get_entropy` and `arc4random` for OP-TEE, fixes a bug in OCALL handling, changes the behavior of some OP-TEE-specific OE API implementations to match that of their SGX counterparts, includes additional math functions for parity with SGX, and enables the crypto and libc tests to build for ARM. It also modifies the hosts of simple tests to launch their enclaves with type AUTO instead of SGX.

**Note:** #2073 is now closed.


HernanGatta:
bors try

HernanGatta:
bors try

HernanGatta:
bors try-

PR #2120, opened 2019-08-22
Revisit OE SDK package settings

### LINKED ISSUE #2107: Revisit OE SDK package settings
While discussing the host verification package (PR #2092), the following questions came up about the OE SDK package:

1. Naming:
For the SDK file name, we are using CPack's default naming convention instead of the recommended Debian default naming convention. The default Debian package name will also include the architecture, which is useful when we start releasing ARM packages.

2. Dependency on `pkg-config`
The OE SDK package depends on `pkg-config`. I believe `pkg-config` is just for convenience instead of a hard dependency. So, it's more appropriate for it to be a Debian "recommendation" instead of a "dependency".

BRMcLaren:
Revisited OE SDK package settings, removal of unneeded dependency and revised Debian file naming scheme to reflect recommended default naming scheme which includes architecture version.

Fix #2107


johnkord:
Two comments by the way:

1. Can you add your full name to your github profile?
2. Can you change the grammar of the second commit to imperative form, and also add 'of the debian package' at the end?  So it should become something like: "Remove pkg-config dependency from the debian package"

gupta-ak:
Also, consider adding `pkg-config` as a recommendation (`CPACK_DEBIAN_PACKAGE_RECOMMENDS`) or suggestion (`CPACK_DEBIAN_PACKAGE_SUGGESTS`) to both the OE SDK package and the host verification package, since our samples do use `pkg-config` files in the Makefiles.

PR #2122, opened 2019-08-23
Modified install-windows-prereq.ps1 input parameters.

johnkord:
This will give an option to copy the Intel DCAP DLLs into windows\system32 instead of installing the drivers (which normally copy the DLLs into windows\system32 as well).

Since simulation mode doesn't work nicely on Windows, this isn't useful for most development scenarios. However it is useful if you want a Windows Docker container to run on a node without SGX and can build the Open Enclave SDK :)

### DIFF
--- a/scripts/ansible/oe-windows-acc-setup.yml
+++ b/scripts/ansible/oe-windows-acc-setup.yml
@@ -36,8 +36,8 @@
     - name: OE setup | Run the install-windows-prereqs.ps1 script (this may take a while)
       script: ../install-windows-prereqs.ps1
         -InstallPath         "C:\openenclave"
-        -WithFLC             "{{ 1 if dcap_testing_node is defined and dcap_testing_node == true else 0 }}"
-        -WithAzureDCAPClient "{{ 1 if dcap_testing_node is defined and dcap_testing_node == true else 0 }}"
+        -LaunchConfiguration "{{ SGX1FLC if dcap_testing_node is defined and dcap_testing_node == true else SGX1 }}"
+        -DCAPClientType      "{{ Azure }}"
         -IntelPSWURL         "{{ intel_psw_2_4_url if dcap_testing_node is defined and dcap_testing_node == true else intel_psw_2_2_url }}"
         -IntelPSWHash        "{{ intel_psw_2_4_hash if dcap_testing_node is defined and dcap_testing_node == true else intel_psw_2_2_hash }}"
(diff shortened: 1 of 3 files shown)


scripts/install-windows-prereqs.ps1
```diff
@@ -28,7 +29,8 @@ Param(
     [string]$AzureDCAPNupkgHash = '152ACE956348E80E533E63E6CB1D3CA20E4CA7DC775FC0B9413F552368F971D6',
     [Parameter(mandatory=$true)][string]$InstallPath,
     [Parameter(mandatory=$true)][bool]$WithFLC,
-    [Parameter(mandatory=$true)][bool]$WithAzureDCAPClient
+    [Parameter(mandatory=$true)][bool]$WithAzureDCAPClient,
+    [switch]$WithoutDriver
```
johnkord: Maybe, from a consistency perspective, I should make this a bool, like the others, rename it to "WithDriver", and default it to $true (without a mandatory flag).

CodeMonkeyLeet: Actually, I believe the standard preference is to use switches and structure bool defaults to `$false`.

Given the growing number of options, particularly given that some of them have interactions (i.e. Installing Azure DCAP client without FLC makes no sense), you might want to structure this as a single parameter taking an enum (or an array of enum values if you want to allow arbitrary combinations) instead.

johnkord: Good idea, thanks Simon.

scripts/ansible/oe-windows-acc-setup.yml
```diff
@@ -36,8 +36,8 @@
     - name: OE setup | Run the install-windows-prereqs.ps1 script (this may take a while)
       script: ../install-windows-prereqs.ps1
         -InstallPath         "C:\openenclave"
-        -WithFLC             "{{ 1 if dcap_testing_node is defined and dcap_testing_node == true else 0 }}"
-        -WithAzureDCAPClient "{{ 1 if dcap_testing_node is defined and dcap_testing_node == true else 0 }}"
+        -LaunchConfiguration "{{ SGX1FLC if dcap_testing_node is defined and dcap_testing_node == true else SGX1 }}"
+        -DCAPClientType      "{{ Azure }}"
```
johnkord: This will install the Azure DCAP client on all Windows CI systems, which will be useful for upcoming host verify tests.

johnkord:
Thanks for all the help, everyone.
bors r+

PR #2123, opened 2019-08-23
Remove root_index.txt and intermediate_index.txt from crypto tests on rebuild

### LINKED ISSUE #2076: Remove root_index.txt and intermediate_index.txt from crypto tests data on rebuild
OpenSSL used by the crypto tests to generate test certificates keeps track of the root and intermediate CA certs issued through the same configuration for revocation purposes in `root_index.txt` and `intermediat_index.txt`. 

These files are not overwritten on builds of the test data folders and instead are extended each time, and extend the corresponding `*.crl.der` test files each time. Given enough iterations, this causes the file sizes to exceed the fixed-buffer size limits for reading these files in the tests, which will then cause the tests to fail to read the der files.

This affects at least:
- tests/crypto
- tests/crypto_crls_cert_chains

This affects both Linux and Windows, although it is more visible on Windows as each execution through VS Test Explorer triggers a rebuild of test collateral, and so increments the size of these files.

oprinmarius:
Removes root_index.txt and intermediate_index.txt from crypto and crypto_crls_cert_chains tests on build
fixes #2076

### DIFF
--- a/tests/crypto/data/make-test-certs
+++ b/tests/crypto/data/make-test-certs
@@ -131,7 +131,9 @@ openssl x509 -req -in leaf2.csr -CA intermediate.cert.pem -CAkey intermediate.ke
 #  - intermediate_crl is issued by intermediate CA and revokes the leaf cert
 #  - root_crl is issued by root CA and also revokes the leaf cert
 
+rm -f intermediate_index.txt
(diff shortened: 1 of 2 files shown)


oprinmarius:
bors try

Issue #2098, opened 2019-08-15
Define a platform-agnostic layer of thread and atomic APIs


jxyang:
Related #850, #851 

On the host side, Windows and Linux have different APIs for thread creation, thread join, atomic cas, etc. For a few cross-platform tests that runs on both Windows and Linux, but have to rely on these APIs, the code tend to be messy with littered `#ifdef _MSC_VER` or `#ifdef __GNUC__` etc. This is also one of the reasons that some tests run on Linux only.  

For a clean design point of view, we should define a platform-agnostic layer of APIs that include but are not limited to:

* oe_thread_create
* oe_thread_join
* sleep
* atomic add
* atomic sub
* atomic cas

dthaler:
For threading, I think we should try to just expose POSIX APIs like pthread_create() and map them over native APIs on platforms that don't have it natively.

jxyang:
@dthaler That's likely to create confusions. Our convention is to use `oe_` prefix.

dthaler:
Actually we have a bunch of cases where the posix apis are defined to use oe_ prefixed versions implemented internally.   Callers use the posix one normally, unless it would cause problems (e.g., linker errors) in which case they have to use the raw oe_ ones.  But the signatures match.   That's all I'm saying here.

PR #2125, opened 2019-08-23
The 2nd half installment of switchless ocalls

jxyang:
* Create threads to service switchless ocalls
* Necessary sychronizations between threads
* Add 'oe_start_switchless_manager' to configure and start switchless manager
* Call the function in test cases that support switchless calls
* Add OS agnostic thread and atomic routines for better readability (#2098)

This also fixes issue 2098.


include/openenclave/host.h
```diff
@@ -317,6 +317,24 @@ void oe_free_key(
     uint8_t* key_info,
     size_t key_info_size);
 
+/**
+ * Start the switchless call manager for the enclave. This should be done
```
dthaler: This is not a useful description, since the term "switchless" is not defined and there is no pointer to any doc for more information, so the app developer has no idea what this is for (I have no clue myself).

include/openenclave/host.h
```diff
@@ -317,6 +317,24 @@ void oe_free_key(
...
+ * right after enclave creation for any enclave that has switchless ocalls.
+ *
+ * @param enclave The enclave handle.
```
dthaler: add doxygen [in,out] style annotations for all new public apis

include/openenclave/host.h
```diff
@@ -317,6 +317,24 @@ void oe_free_key(
...
+ * @param num_host_workers The number of host worker threads that are
+ * dedicated to servicing switchless ocalls.
+ *
+ * @retval OE_OK The switchless manager is successfully started.
+ * @retval OE_UNEXPECTED Something unexpected happened, e.g. the switchless
+ * manager is already started.
+ * @retval OE_INVALID_PARAMETER The enclave or num_host_workers is not valid.
+ * @retval OE_FAILURE The worker threads are not created properly.
+ */
+oe_result_t oe_start_switchless_manager(
```
dthaler: Why is this new API needed?   Where is the new sample code that shows how it is used?  Or is it just used by edger8r generated code?  (I don't see changes to edger8r).   We should not add new APIs that are public and have no sample code.

jxyang:
@dthaler My understanding of the paper is different from yours in two ways:

1) Intel SDK doesn't dynamically change a function from switchless to non-switchless, or vice versa, based on certain criterion of the function itself. Rather, the decision is made based on the availability of worker threads. If there is no free worker, the SDK will dispatch a function labeled `transition_using_threads` in traditional way.  That's the fallback mechanism we are considering supporting.

2) The paper is largely about dynamic adjustment of the number of worker threads based on work load. But the implementation is not in the official Intel SDK (yet). We can't pursue parity with a feature that is still private.

radhikaj:
> @dthaler My understanding of the paper is different from yours in two ways:
> 
> 1. Intel SDK doesn't dynamically change a function from switchless to non-switchless, or vice versa, based on certain criterion of the function itself. Rather, the decision is made based on the availability of worker threads. If there is no free worker, the SDK will dispatch a function labeled `transition_using_threads` in traditional way.  That's the fallback mechanism we are considering supporting.

I agree we cant make calls 'switchless' by default. Intel only falls back to 'traditional' calls when there are no worker threads free at the time of the call. It does not automatically decide to switch from a 'traditional' call to a 'switchless' call. Please let me know if there is a different understanding here. There are security implications with switchless calls and we want them to be 'opt-in' not automatic.

mikbras:
After thinking about this for awhile, I think we should hide **oe_start_switchless_manager()** and define a new flag for **oe_create_enclave()** as follows.

- OE_ENCLAVE_FLAG_ENABLE_SWITCHLESS

We might also want a flag to force switchless calls in all cases.

- OE_ENCLAVE_FLAG_DEFAULT_TO__SWITCHLESS

The number of threads can default to the number of enclave thread bindings at first. To provide finer grain control, we'd need **oe_create_enclave_ex()**, which would probably be necessary anyway for parity with Intel's SDK.

I think we should not allocate manager threads by default, since this will impose resource requirements in environments that don't want or can't support this feature. Also, we might eventually support UEFI which does not have thread support.

dthaler:
> After thinking about this for awhile, I think we should hide oe_start_switchless_manager() and define a new flag for oe_create_enclave() as follows.
> 
> OE_ENCLAVE_FLAG_SWITCHLESS

I would be ok with that, and would recommend adding "CONTEXT" for clarity, i.e., OE_ENCLAVE_FLAG_CONTEXT_SWITCHLESS

However, I still think this should be the default.   Radhika mentioned some security concerns, but I don't see any mention of such in the paper.  If there are security concerns then maybe we shouldn't have this at all, but I don't know of any such concerns so far.

> The number of threads can default to the number of enclave thread bindings at first. To provide finer grain control, we'd need oe_create_enclave_ex(), which would probably be necessary anyway for parity with Intel's SDK.

Sure, although I'm not yet convinced there's any useful scenario for using finer grain control.   My belief is that at runtime, the SDK can know better than the developer did at compile time especially since in many cases the developer may have no idea which actual machine will run the enclave and how many cores it has and what other enclaves will be on the same machine.

> I think we should not allocate manager threads by default, since this will impose resource requirements in environments that don't want or can't support this feature. For example, we might eventually support UEFI which does not have thread support.

If you don't have support at all, then of course you can't enable it.  But if the hardware has support, why would you know want to enable it?   Especially if there is runtime determination of whether a call should actually be made switchless or not.

jxyang:
@mikbras Our existing API `oe_create_enclave` already has parameters like `config` and `config_size`. Were they created with an expectation to be applied to situations like switchless? The description is follows:

```
 * @param config Additional enclave creation configuration data for the specific
 * enclave type. This parameter is reserved and must be NULL.
 *
```
We can use them to configure the number of worker threads. Specifically:

* define a structure containing # of work threads for switch OCALLs and ECALLs, and pass its pointer as `config` in `oe_create_enclave`
* This is only looked at when `flag` has `OE_ENCLAVE_FLAG_CONTEXT_SWITCHLESS` set. Maybe later we can add more capabilities similarly.
* if `OE_ENCLAVE_FLAG_CONTEXT_SWITCHLESS` is set in `flag` but `config` is NULL, we just default # of work threads to number of thread bindings.

mikbras:
> After thinking about this for awhile, I think we should hide oe_start_switchless_manager() and define a new flag for oe_create_enclave() as follows.
> OE_ENCLAVE_FLAG_SWITCHLESS
> 
> I would be ok with that, and would recommend adding "CONTEXT" for clarity, i.e., OE_ENCLAVE_FLAG_CONTEXT_SWITCHLESS
> However, I still think this should be the default.   Radhika mentioned some security concerns, but I don't see any mention of such in the paper.  If there are security concerns then maybe we shouldn't have this at all, but I don't know of any such concerns so far.
> 
> The number of threads can default to the number of enclave thread bindings at first. To provide finer grain control, we'd need oe_create_enclave_ex(), which would probably be necessary anyway for parity with Intel's SDK.
> 
> Sure, although I'm not yet convinced there's any useful scenario for using finer grain control.   My belief is that at runtime, the SDK can know better than the developer did at compile time especially since in many cases the developer may have no idea which actual machine will run the enclave and how many cores it has and what other enclaves will be on the same machine.
> 
> I think we should not allocate manager threads by default, since this will impose resource requirements in environments that don't want or can't support this feature. For example, we might eventually support UEFI which does not have thread support.
> 
> If you don't have support at all, then of course you can't enable it.  But if the hardware has support, why would you know want to enable it?   Especially if there is runtime determination of whether a call should actually be made switchless or not.

Good points Dave. I would say for now that we should always enable the switchless feature and we should set the number of threads to the number of configured enclave threads. That way every switchless ocall will have a thread to service it. I retract my earlier suggestion of adding a new flag. The need for finer grained control is yet to be seen.

mikbras:
Here is my proposal for introducing configuration settings with the existing **oe_create_enclave** interface.

```c++
/*
**
** enclave.h definitions:
**
*/

typedef enum _oe_config_type
{
    OE_CONFIG_TYPE_SWITCHLESS = 0x8582fab4,
    OE_CONFIG_RWX_PAGES = 0x864209e8,
}
oe_config_type_t;

/* Base structure type for configuration parameters. */
typedef struct _oe_config
{
    oe_config_type_t type;
}
oe_config_t;

/* The number of switchless threads. */
typedef struct _oe_config_switchless
{
    /* Extends oe_config_t. */
    oe_config_t base;
    size_t num_threads;
}
oe_config_switchless_t;

/* Extra read-write-execute pages (needed by SGX-LKL). */
typedef struct _oe_config_rwx_pages
{
    /* Extends oe_config_t. */
    oe_config_t base;
    size_t num_pages;
}
oe_config_rwx_pages_t;

oe_result_t oe_create_enclave(
    const char* enclave_path,
    oe_enclave_type_t enclave_type,
    uint32_t flags,
#if 0
    /* The original signature works. */
    const void* config,
#else
    /* This works and still accepts NULL and and void* pointers. */
    const oe_config_t* config[],
#endif
    uint32_t config_size,
    const oe_ocall_func_t* ocall_table,
    uint32_t ocall_table_size,
    oe_enclave_t** enclave_out);
```
/*
**
** Example:
**
*/

oe_result_t create_my_enclave(
    const char* enclave_path,
    oe_enclave_type_t enclave_type,
    uint32_t flags,
    const oe_ocall_func_t* ocall_table,
    uint32_t ocall_table_size,
    oe_enclave_t** enclave_out)
{
    oe_config_switchless_t config_switchless =
    {
        .base.type = OE_CONFIG_TYPE_SWITCHLESS,
        .num_threads = 16,
    };
    oe_config_rwx_pages_t config_rwx_pages =
    {
        .base.type = OE_CONFIG_RWX_PAGES,
        .num_pages = 4096,
    };
    const oe_config_t* config[] =
    {
        &config_switchless.base,
        &config_rwx_pages.base,
    };

    return oe_create_enclave(
        enclave_path,
        enclave_type,
        flags,
        config,
        OE_COUNTOF(config),
        ocall_table,
        ocall_table_size,
        enclave_out);
}

mikbras:
And here's and example which illustrates how to specify hardware specific (SGX) settings using the config parameter.

```c++
oe_result_t create_my_enclave(
    const char* enclave_path,
    oe_enclave_type_t enclave_type,
    uint32_t flags,
    const oe_ocall_func_t* ocall_table,
    uint32_t ocall_table_size,
    oe_enclave_t** enclave_out)
{
    oe_config_sgx_t config_sgx =
    {
        .product_id = 1,
        .security_version = 1,
        .allow_debug = true,
        .num_heap_pages = 1024,
        .num_stack_pages = 256,
        .num_tcs = 8,
    };
    const oe_config_t* config[] =
    {
        &config_sgx.base,
    };

    return oe_create_enclave(
        enclave_path,
        enclave_type,
        flags,
        config,
        OE_COUNTOF(config),
        ocall_table,
        ocall_table_size,
        enclave_out);
}
```

mikbras:
Also, we might want to change 'config' to 'setting'.

enclave/core/hostcalls.c
```diff
@@ -167,7 +167,7 @@ void* oe_allocate_switchless_ocall_buffer(size_t size)
 void oe_free_switchless_ocall_buffer(void* buffer)
 {
     OE_UNUSED(buffer);
-    /* Do nothing. Buffer will be freed on ECALL RETURN */
+    oe_shm_clear();
```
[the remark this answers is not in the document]
jxyang: @CodeMonkeyLeet `oe_shm_malloc` implicitly creates the arena if it's not there already. But your understanding of the semantic is correct.

Unfortunately we have to distinguish these functions from what is proposed in #2099. We can't do that without emphasizing the implementation is pool based, or arena based, or region based, or stack based... I have heard so many variants of the same concept :-). Naming them `oe_internal_shared_` or `oe_core_shared_*` would likely cause confusions when later #2099 is implemented.

What about `oe_shared_arena_malloc`?

enclave/core/sgx/calls.c
```diff
@@ -586,12 +593,19 @@ oe_result_t oe_call_host_function_by_table_id(
...
-    // TODO: for switchessless calls, push the job (wrapped in args) to an
-    // available worker thread, and wait for result
-    // if (!switchless)
+    if (!switchless)
     {
         OE_CHECK(oe_ocall(OE_OCALL_CALL_HOST_FUNCTION, (uint64_t)args, NULL));
     }
+    else
+    {
+        OE_CHECK(_oe_post_switchless_ocall(args));
```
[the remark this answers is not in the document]
jxyang: We may implement asynchronous switchless calls later. In that case, we are not supposed to block.

enclave/core/sgx/calls.c
```diff
@@ -586,12 +593,19 @@ oe_result_t oe_call_host_function_by_table_id(
...
+
+        while (args->result == __OE_RESULT_MAX)
```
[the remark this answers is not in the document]
jxyang: Atomic operations hurt the performance. I believe the fields in `args` are always aligned.

enclave/core/switchlesscalls.c
```diff
@@ -0,0 +1,122 @@
...
+    // Ensure the switchless manager and its arrays are outside of enclave
+    if (!oe_is_outside_enclave(manager, sizeof(oe_switchless_call_manager)) ||
+        !oe_is_outside_enclave(manager->host_worker_contexts, contexts_size) ||
+        !oe_is_outside_enclave(manager->host_worker_threads, threads_size))
+    {
+        OE_RAISE(OE_INVALID_PARAMETER);
+    }
+
+    // Copy the worker context array pointer and its size to avoid TOCTOU
+    _num_host_workers = manager->num_host_workers;
```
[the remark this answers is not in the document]
jxyang: Good catch. Yes the contents of these buffers are meant to be modified by the host any point of time. I think these calculated risks are inherent to switchless calls.

enclave/core/switchlesscalls.c
```diff
@@ -0,0 +1,122 @@
...
+**
+**  Post the function call (wrapped in args) to a free host worker thread
+**  by writing to its context.
+**
+**==============================================================================
+*/
+oe_result_t _oe_post_switchless_ocall(oe_call_host_function_args_t* args)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    args->result = __OE_RESULT_MAX; // Means the call hasn't been processed.
```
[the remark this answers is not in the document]
jxyang: Either way works. I am just planning for the future scenario that we may need to support asynchronous switchless calls.

enclave/core/switchlesscalls.c
```diff
@@ -0,0 +1,122 @@
...
+
+    if (_num_host_workers == 0)
```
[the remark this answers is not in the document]
jxyang: I meant to remind people who forget to initialize switchless before making swtchless calls. But I will take the suggestion considering Intel's SDK does this.

enclave/core/switchlesscalls.c
```diff
@@ -0,0 +1,122 @@
...
+                    (void* volatile*)&_host_worker_contexts[i].call_arg,
+                    NULL,
+                    args))
+            {
+                result = OE_OK;
+                break;
+            }
+        }
+    }
+
```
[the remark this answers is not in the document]
jxyang: I will make it cycle through all worker threads once and then give up and fall back to the regular ocall.

enclave/core/switchlesscalls.c
```diff
@@ -0,0 +1,122 @@
...
+done:
+    return result;
+}
+
+/*
+**==============================================================================
+**
+** oe_switchless_call_host_function()
+** This is the preferred way to call host functions switchlessly.
```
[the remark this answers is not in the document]
jxyang: I can't answer the question because the comment was made by the previous committer. I will remove it anyways.

enclave/core/switchlesscalls.c
```diff
@@ -0,0 +1,122 @@
...
+    return oe_call_host_function_by_table_id(
+        OE_UINT64_MAX,
+        function_id,
+        input_buffer,
+        input_buffer_size,
+        output_buffer,
+        output_buffer_size,
+        output_bytes_written,
+        true /* switchless */);
+}
```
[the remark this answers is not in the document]
jxyang: Must be something wrong with git "check out Windows style, check in Unix style". I will double check.

host/windows/hostthread.c
```diff
@@ -13,6 +13,22 @@
...
+}
+
+int oe_thread_join(oe_thread thread)
+{
+    if (WaitForSingleObject(thread, INFINITE) == WAIT_OBJECT_0)
+    {
+        CloseHandle(thread);
+        return 0;
+    }
+    return 1;
```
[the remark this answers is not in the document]
jxyang: Like Mike said in another thread (no pun intended), the error codes returned by the Windows implementation and Linux implementation are so different thus we can't make a cross-platform interpretation of them. However it may still be desirable to let the caller know whether this function succeeds or not.

I can make it return either `OE_OK` or `OE_THREAD_JOIN_ERROR`, does that address your concerns?

host/windows/hostthread.c
```diff
@@ -13,6 +13,22 @@
...
 
+int oe_thread_create(oe_thread* thread, void* (*func)(void*), void* arg)
+{
+    *thread = CreateThread(NULL, 0, func, arg, 0, NULL);
+    return *thread == NULL;
+}
+
+int oe_thread_join(oe_thread thread)
+{
+    if (WaitForSingleObject(thread, INFINITE) == WAIT_OBJECT_0)
```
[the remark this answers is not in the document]
jxyang: It does generate a warning. Thanks for the reminder.

host/windows/hostthread.c
```diff
@@ -13,6 +13,22 @@
 **==============================================================================
 */
 
+int oe_thread_create(oe_thread* thread, void* (*func)(void*), void* arg)
+{
+    *thread = CreateThread(NULL, 0, func, arg, 0, NULL);
```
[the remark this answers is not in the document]
jxyang: Yes, with a warning. I will add a cast.

host/hostthread.h
```diff
@@ -49,6 +49,25 @@ typedef DWORD oe_thread_key;
...
 
+/**
+ * Create a platform-specific thread.
+ *
+ * @param func The pointer to the start routine.
+ * @param arg The argument to the start routine.
+ *
+ * @returns Returns zero on success.
+ */
+int oe_thread_create(oe_thread* thread, void* (*func)(void*), void* arg);
```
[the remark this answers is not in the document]
jxyang: I have made it result `oe_result_t`. Let me know if that addressed your concern.

host/sgx/calls.c
```diff
@@ -358,6 +357,7 @@ static const char* oe_ecall_str(oe_func_t ecall)
     {
         "DESTRUCTOR",
         "INIT_ENCLAVE",
+        "INIT_SWITCHLESS",
```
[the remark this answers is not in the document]
jxyang: Can you clarify? Currently the enum is added as the 3rd one in the list. Do you mean move it further down just before `OE_ECALL_MAX`? How does that allow "separately hosts and enclaves to work together"?

host/sgx/enclave.h
```diff
@@ -124,6 +125,9 @@ struct _oe_enclave
 
     /* Meta-data needed by debugrt  */
     oe_debug_enclave_t* debug_enclave;
+
+    /* Manager for switchless calls */
+    oe_switchless_call_manager* switchless_manager;
```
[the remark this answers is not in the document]
jxyang: I looked at debugrt, and it seems there is no requirement for `debug_enclave` to be the last field. Correct me if I am wrong @anakrish .

host/sgx/switchless.c
```diff
@@ -0,0 +1,122 @@
...
+*/
+static void* _switchless_ocall_worker(void* arg)
+{
+    host_worker_thread_context* context = (host_worker_thread_context*)arg;
+
+    while (!context->stopping)
+    {
+        if (context->call_arg != NULL)
+        {
+            oe_handle_call_host_function(
```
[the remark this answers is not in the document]
jxyang: We want keep looping even though one switchless OCALL failed, so the next OCALLs can still be processed.

Generally the error code is in the output buffer, and it's the caller's responsibility to check it.

host/sgx/switchless.c
```diff
@@ -0,0 +1,122 @@
...
+        OE_RAISE_MSG(
+            OE_INVALID_PARAMETER,
+            "We should have at least one host worker thread");
+
+    if (enclave == NULL)
+        OE_RAISE_MSG(
+            OE_INVALID_PARAMETER, "The enclave is not created properly");
+
+    // Switchless manager can only be started once.
+    if (enclave->switchless_manager != NULL)
```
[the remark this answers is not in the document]
jxyang: We are moving this into enclave creation.

host/sgx/switchless.c
```diff
@@ -0,0 +1,122 @@
...
+        OE_RAISE_MSG(
+            OE_UNEXPECTED, "Switchless manager is already initialized");
+
+    // Limit the number of host workers to the number of thread bindings
+    // because the maximum parallelism is dictated by the latter for
+    // synchronous ocalls. We may need to revisit this for asynchronous
+    // calls later.
+    if (num_host_workers > enclave->num_bindings)
```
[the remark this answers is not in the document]
jxyang: Now with this function removed from the public interface, can you suggest a place to publish it?

host/sgx/switchless.c
```diff
@@ -0,0 +1,122 @@
...
+    // Start the worker threads, and assign each one a private context.
+    for (size_t i = 0; i < num_host_workers; i++)
+    {
+        manager->host_worker_contexts[i].enclave = enclave;
+        if (oe_thread_create(
+                &manager->host_worker_threads[i],
+                _switchless_ocall_worker,
+                &manager->host_worker_contexts[i]) != 0)
+        {
+            oe_stop_worker_threads(manager);
```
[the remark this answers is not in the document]
jxyang: It will. Because we pass in local variable `manager`, not `enclave->switchless_manager`.

include/openenclave/internal/atomic.h
```diff
@@ -38,4 +42,28 @@ OE_INLINE uint64_t oe_atomic_decrement(volatile uint64_t* x)
...
+    return __sync_bool_compare_and_swap(dest, old, newval);
+#elif defined(_MSC_VER)
+    return _InterlockedCompareExchange64(dest, newval, old) == old;
+#else
+#error "unsupported"
+#endif
+}
+
+OE_INLINE
+bool oe_atomic_cas_ptr(void* volatile* dest, void* old, void* newptr)
```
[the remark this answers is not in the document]
jxyang: Maybe to avoid type casting? or for backward compatibility with systems having 32-bit pointers? I am not sure why windows keep them separate. This is just to be safe.

include/openenclave/internal/switchless.h
```diff
@@ -0,0 +1,31 @@
...
+
+typedef uint64_t oe_thread_t;
+
+typedef struct _host_worker_thread_context
+{
+    oe_call_host_function_args_t* call_arg;
+    uint64_t sleep_time;
+    bool stopping;
+    oe_enclave_t* enclave;
+    oe_result_t ocall_result;
```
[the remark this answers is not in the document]
jxyang: Good catch. Removed.

tests/switchless/enc/enc.c
```diff
@@ -36,24 +36,72 @@ int enc_echo(char* in, char out[100])
...
+{
+    oe_result_t result;
+
+    if (oe_strcmp(in, "Hello World") != 0)
     {
         return -1;
     }
 
-    if (return_val != 0)
+    char* host_allocated_str = oe_host_strdup("oe_host_strdup2");
```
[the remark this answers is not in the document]
jxyang: It's not material to the perf testing. I will remove it from the test case.

tests/switchless/host/host.c
```diff
@@ -62,26 +64,38 @@ int main(int argc, const char* argv[])
...
-        else
-            printf("created thread %u\n", i);
-    }
+    oe_start_switchless_manager(enclave, 2);
 
-    for (int i = 0; i < NUM_HOST_THREADS; i++)
-    {
-        pthread_join(threads[i], NULL);
-    }
+    char out[100];
```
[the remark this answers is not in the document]
jxyang: They are inherited from test case `echo`. I will clean them up.

tests/switchless/host/host.c
```diff
@@ -62,26 +64,38 @@ int main(int argc, const char* argv[])
...
+    int return_val;
+
+    double switchless_microseconds = 0;
+    struct timespec start, end;
+
+    clock_gettime(CLOCK_REALTIME, &start);
```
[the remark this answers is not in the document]
jxyang: A million iteration takes 14 seconds in this case (1 for switchless calls, 13 for regular calls). I think it's OK considering many other test cases run significantly longer.

When we implement #2129, it would be a good idea to parameterize it.

host/windows/hostthread.c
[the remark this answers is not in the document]
dthaler: I've said this elsewhere but I'll repeat it again here.   I do not think we should have an oe_thread_join per se.   We should instead have oe_pthread_join that exposes the POSIX syntax (including on Windows) just prefixed with oe.  In that implementation it would return the POSIX error codes not oe_result_t error codes.

jxyang:
@mikbras It's a neat idea to define  `oe_config_t` and then have more specific configs `inherit` from it. I was thinking we can do the same thing to `oe_enclave_t` as well: making it a base struct, and having `oe_enclave_sgx_t` and `oe_enclave_optee_t` derived from it. 

For this PR, I just make the `switchless_settings` by itself though without inheritance from, `oe_config_t`.

jxyang:
The last commit should have addressed most of the comments. Consider the huge number of comments, it's understandable that I might miss some of them. Please raise the issues I missed or resolved by mistake again.

tests/switchless/host/host.c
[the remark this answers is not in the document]
CodeMonkeyLeet: Think of it this way, you're taxing every developer an extra 14 seconds when they run regression tests and that really adds up. Of the >100 tests only the crypto and thread tests take >10s to run and those really should be fixed. Just because we already have excessively long running tests doesn't mean we should add more of them. If we are not implementing #2129 until later, then don't put in the extra iterations right now, just take them out completely.

include/openenclave/host.h
```diff
@@ -45,11 +45,23 @@ OE_EXTERNC_BEGIN
...
+/**
+ *  Flag passed into oe_create_enclave to enable the enclave with switchless
+ *  calls. A switchless call is a function invocation from the trusted world to
+ *  the untrusted world, or vice versa, without context switches and transition
+ *  of the execution mode. Typically the caller posts the job to a shared
+ *  memory buffer accessible by both the trusted world and the untrusted world,
+ *  and a workers in the other world takes the job and posts the result, which
+ *  is retrieved by the caller.
+ */
+#define OE_ENCLAVE_FLAG_CONTEXT_SWITCHLESS 0x00000004u
```
CodeMonkeyLeet: Why is this flag necessary? There are already at least two mechanisms that do the same thing:

- A developer already has to explicitly decide whether to mark any methods in the EDL as `transition_using_threads` and it seems weirdly redundant to ask them to specify an additional flag on enclave creation if they've already expressed this intention. Particularly since we require that they call the EDL generated wrapper for enclave creation, it seems like edger8r should be able to already tell if switchless manager is necessary or not.

- There's an added option to pass the number of host worker threads as part of `config` to `oe_create_enclave()`. Why isn't this just equivalent to passing 0 to that function if you explicitly want to specify that no switchless calls should be used (and are not using the edger8r wrapper)?

host/windows/hostthread.c
```diff
@@ -13,25 +13,26 @@
 **==============================================================================
 */
 
-int oe_thread_create(oe_thread* thread, void* (*func)(void*), void* arg)
+oe_result_t oe_thread_create(oe_thread* thread, void* (*func)(void*), void* arg)
 {
-    *thread = CreateThread(NULL, 0, func, arg, 0, NULL);
-    return *thread == NULL;
+    DWORD (*start_routine)(void*) = (DWORD(*)(void*))func;
```
CodeMonkeyLeet: Consider a typedef for the `start_routine` function pointer type to make this more readable.

host/windows/hostthread.c
```diff
@@ -13,25 +13,26 @@
 **==============================================================================
 */
 
-int oe_thread_create(oe_thread* thread, void* (*func)(void*), void* arg)
+oe_result_t oe_thread_create(oe_thread* thread, void* (*func)(void*), void* arg)
```
CodeMonkeyLeet: I'm fine with using `oe_result_t` here but that makes it weirdly inconsistent with the rest of the functions in this file declared to mimic the POSIX function signatures. 

@mikbras Should there be a task to switch them all over to `oe_result_t` or do you prefer these to remain `int`?

host/windows/hostthread.c
```diff
@@ -13,25 +13,26 @@
...
+        return OE_OK;
     }
-    return 1;
+    return OE_THREAD_JOIN_ERROR;
 }
 
 oe_thread oe_thread_self(void)
 {
-    return GetCurrentThreadId();
+    return (oe_thread)(UINT_PTR)GetCurrentThreadId();
```
CodeMonkeyLeet: Why is the double cast necessary?

common/result.c
```diff
@@ -118,6 +118,12 @@ const char* oe_result_str(oe_result_t result)
             return "QE_QUOTE_ENCLAVE_IDENTITY_PRODUCTID_MISMATCH";
         case OE_VERIFY_FAILED_AES_CMAC_MISMATCH:
             return "OE_VERIFY_FAILED_AES_CMAC_MISMATCH";
+        case OE_SWITCHLESS_OCALL_MISSED:
```
dthaler: If this error code is needed at all, then rename to OE_CONTEXT_SWITCHLESS_OCALL_MISSED

include/openenclave/bits/result.h
```diff
@@ -314,6 +314,21 @@ typedef enum _oe_result
      */
     OE_VERIFY_FAILED_AES_CMAC_MISMATCH,
 
+    /**
+     * Failed to post a switchless call to host workers
```
dthaler: Don't use the term "switchless" by itself as it's ambiguous/confusing.  Say "context-switchless"

include/openenclave/host.h
dthaler: Completely agree with Simon, this flag should not be needed.

include/openenclave/internal/switchless.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <openenclave/bits/defs.h>
+#include <openenclave/bits/types.h>
+#include <openenclave/internal/calls.h>
+
+typedef uint64_t oe_thread_t;
+
+typedef struct _switchless_settings
+{
+    size_t num_host_workers;
```
dthaler: nit: bad coding style to use abbreviations like "num".  Suggest "host_worker_count"

include/openenclave/internal/switchless.h
```diff
@@ -10,22 +10,29 @@
 
 typedef uint64_t oe_thread_t;
 
+typedef struct _switchless_settings
```
CodeMonkeyLeet: This cannot be an internal struct if you expect that developers will pass this into the public `oe_create_enclave` call directly. 

Also missed this earlier: please make sure your typedefs are correctly suffixed by `_t`, so `oe_switchless_settings_t`, `oe_switchless_call_manager_t`

CodeMonkeyLeet: More generally, because we expect the `oe_create_enclave()` parameter `config` to be able to take a range of composable configuration structs, please make sure that the structs are designed as TLV (type-length-value) so that they can be walked by the `oe_create_enclave()` implementation, for example:

```c
typedef struct _oe_switchless_settings
{
    uint32_t magic;
    size_t size;
    size_t num_host_workers;
} oe_switchless_settings_t;
```

tests/switchless/host/host.c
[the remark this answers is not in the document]
jxyang: Make sense. Will do.

include/openenclave/host.h
jxyang: We need this flag so in later call to `oe_create_enclave`, we know the `void*` parameter is actually for configuring switchless.  

If we make the switchless setting a TLV type, then I agree we don't need this flag.

host/windows/hostthread.c
jxyang: Because `GetCurrentThreadId()` returns DWORD. That compiler complains if we cast it into `oe_thread` directly.

host/windows/hostthread.c
jxyang: I follow up with Mike. His opinion is that we should return int for those two function to be consistent. I will make them returning POSIX error code as much as possible to address Simon's original comment.

include/openenclave/internal/switchless.h
jxyang: The convention in OE seems to be using `num_`, such as `num_extensions`, `num_ocalls`.

host/windows/hostthread.c
jxyang: Also per Mike's suggestion, I have combined `oe_thread` and `oe_thread_t` into just `oe_thread_t`. Previously they are defined differently which is confusing.

include/openenclave/internal/switchless.h
dthaler: The Windows coding style conventions argue that it's a bad convention.  Specifically, using abbreviations in names is considered bad practice.

dthaler: I won't -1 a PR because of this, but I will continue to complain about instances of bad coding conventions.

include/openenclave/host.h
jxyang: The TLV type of configs is implemented. Removing the flag.

host/windows/hostthread.c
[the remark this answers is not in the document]
jxyang: `oe_thread_create` and `oe_thread_join` return POSIX error codes on Linux now. On Windows they are mapped to the POSIX error codes as much as possible.

jxyang:
@mikbras @CodeMonkeyLeet @dthaler Please review the latest changes that are made based on your comments.  

The only non-trivial change is the removal of the `OE_ENCLAVE_FLAG_CONTEXT_SWITCHLESS` flag, which becomes unnecessary after the introduction of TLV config types.  Those are exposed in `openenclave/host.h`.

enclave/core/switchlesscalls.c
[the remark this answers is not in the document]
mikbras: Seems okay now but you might want to have Anand stare at it for awhile. :)

host/hostthread.h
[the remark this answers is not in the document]
mikbras: My point is that we need a single thread routine signature:
```c++
typedef void (*oe_thread_proc_t)(void* arg);
```

On both Linux and Windows we need a static wrapper thread routine that calls this function. In this way the thread routine passed in the Linux and Windows case will be appropriate for the platform.

host/sgx/calls.c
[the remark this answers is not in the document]
mikbras: Oh sorry. Please see comment below.

```c++
typedef enum _oe_func
{
    OE_ECALL_DESTRUCTOR = OE_ECALL_BASE,
    OE_ECALL_INIT_ENCLAVE,
    OE_ECALL_INIT_SWITCHLESS,
    OE_ECALL_CALL_ENCLAVE_FUNCTION,
    OE_ECALL_VIRTUAL_EXCEPTION_HANDLER,
    /* Caution: always add new ECALL function numbers here */
    OE_ECALL_MAX,

    OE_OCALL_CALL_HOST_FUNCTION = OE_OCALL_BASE,
    OE_OCALL_THREAD_WAKE,
    OE_OCALL_THREAD_WAIT,
    OE_OCALL_MALLOC,
    OE_OCALL_FREE,
    OE_OCALL_SLEEP,
    OE_OCALL_GET_TIME,
    /* Caution: always add new OCALL function numbers here */
    OE_OCALL_MAX, /* This value is never used */

    __OE_FUNC_MAX = OE_ENUM_MAX,
} oe_func_t;
```

host/windows/hostthread.c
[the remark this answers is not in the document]
mikbras: Maybe add a wrapper function rather than a cast as described above.

host/windows/hostthread.c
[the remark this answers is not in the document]
mikbras: I would return -1 for failure here and in the Linux case. A return value of OE_EINVAL cannot be read by the caller unless they use the POSIX errnos.

include/openenclave/host.h
jxyang: We will use a structure (exposed publicly) `oe_config_context_switchless_t` for the configuration purpose.

enclave/core/hostcalls.c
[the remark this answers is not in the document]
jxyang: Changed to `oe_arena_malloc` etc.

host/windows/hostthread.c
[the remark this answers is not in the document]
jxyang: We actually defined `OE_EINVAL` in the cross-platform header `openenclave\corelibc\errno.h`.

host/hostthread.h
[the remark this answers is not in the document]
jxyang: @mikbras This PR has been dragged for too long. If it's Ok with you, I prefer to fix this in a separate PR. Issue filed: https://github.com/openenclave/openenclave/issues/2162

jxyang:
bors try

3rdparty/libcxx/libcxx/src/locale.cpp
```diff
@@ -5642,6 +5642,9 @@ __init_pat(money_base::pattern& pat, basic_string<charT>& __curr_symbol_,
                 return;
             case 1:  // Space between currency-and-sign or currency and value.
                 pat.field[1] = none;
+
```
dthaler: Why the change to this file?   Offhand, looks like it should be reverted

enclave/core/shm.h
```diff
@@ -6,23 +6,23 @@
...
+typedef struct _shared_memory_arena
 {
     /* Buffer holding the shared memory pool */
     uint8_t* buffer;
     size_t capacity;
     size_t used;
-    struct _shared_memory_pool* next;
-} Shared_memory_pool;
+    struct _shared_memory_arena* next;
+} Shared_memory_arena;
```
dthaler: seems like this should be shared_memory_arena_t

enclave/core/switchlesscalls.c
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "switchlesscalls.h"
+#include <openenclave/enclave.h>
+#include <openenclave/internal/atomic.h>
+#include <openenclave/internal/raise.h>
+
+// The number of host thread workers. Initialized by host through ECALL
+static size_t _num_host_workers = 0;
```
dthaler: _host_worker_count ?

enclave/core/switchlesscalls.c
```diff
@@ -0,0 +1,133 @@
...
+{
+    return oe_call_host_function_by_table_id(
+        OE_UINT64_MAX,
+        function_id,
+        input_buffer,
+        input_buffer_size,
+        output_buffer,
+        output_buffer_size,
+        output_bytes_written,
+        true /* switchless */);
```
dthaler: the fact that you need a comment here indicates that the type is probably wrong.
Instead of Boolean, consider an enum with two values.

host/hostthread.h
```diff
@@ -40,15 +40,32 @@ typedef pthread_key_t oe_thread_key;
...
 #define OE_H_MUTEX_INITIALIZER INVALID_HANDLE_VALUE
 
 typedef DWORD oe_thread_key;
 
 #endif
 
+/**
+ * Create a platform-specific thread.
+ *
+ * @param func The pointer to the start routine.
```
dthaler: public apis should use [in,out] style doxygen annotation

include/openenclave/internal/switchless.h
```diff
@@ -0,0 +1,32 @@
...
+    volatile oe_call_host_function_args_t* call_arg;
+    bool is_stopping;
+    oe_enclave_t* enclave;
+} host_worker_thread_context_t;
+
+typedef struct _oe_switchless_call_manager
+{
+    host_worker_thread_context_t* host_worker_contexts;
+    oe_thread_t* host_worker_threads;
+    size_t num_host_workers;
```
dthaler: nit: I'd recommend "host_worker_count", but this file isn't public so not as important as if it were public

tests/switchless/switchless.edl
```diff
@@ -3,18 +3,28 @@
 
 enclave {
     trusted {
-        public int enc_echo(
+        public int enc_echo_switchless(
```
dthaler: I thought there was an EDL keyword for switchless, why isn't that used?

include/openenclave/host.h
```diff
@@ -64,6 +65,38 @@ typedef void (*oe_ocall_func_t)(
     size_t output_buffer_size,
     size_t* output_bytes_written);
 
+/**
+ * Types of configurations passed into oe_create_enclave
+ */
+typedef enum _oe_enclave_config_type
+{
+    OE_ENCLAVE_CONFIG_CONTEXT_SWITCHLESS = 0xdc73a628,
```
dthaler: I will say the same thing I say to any PR touching public header files...
I do not believe any PR should be merged that adds/changes a public API unless that PR comes with sample code and tests, or the API is an addition that is explicitly marked as experimental so breaking changes can be made at any time.  (This one comes with tests but no sample code, and no marking of experimental.)   I believe the simplest way to address for now is to add some way to mark this as experimental, e.g., by putting this inside an #ifdef OE_EXPERIMENTAL or some other annotation, with a // comment saying it's experimental and breaking changes may occur in the future once more experience is gained.

3rdparty/libcxx/libcxx/src/locale.cpp
jxyang: This is probably done by format-code. I will revert.

tests/switchless/switchless.edl
jxyang: This is a regular ECALL. The `switchless` in the name simply means it will call into a switchless OCALL, in contrast to `enc_echo` which calls into a regular OCALL.

include/openenclave/host.h
jxyang: We have created an issue to add sample code for switchless. https://github.com/openenclave/openenclave/issues/2130

host/hostthread.h
jxyang: This function is not a public API, like all other functions in the same file.

enclave/core/switchlesscalls.c
jxyang: Will do.

enclave/core/shm.h
jxyang: Will do

enclave/core/switchlesscalls.c
jxyang: This is an internal function. I think boolean type is more lightweight than an enum, for the latter we need a header file and runtime memory to store it. I prefer not to waste enclave memory on this.

include/openenclave/host.h
dthaler: https://github.com/openenclave/openenclave/blob/master/docs/Contributing.md says:
"DON'T try to merge a giant feature in a single PR. Instead, break it into pieces with feature flags (pre-processor guards or --experimental flags), and submit multiple, smaller PRs to review each piece. Once the whole feature is in, agreed on, and tested, submit a PR to remove the guards."

I'm happy to +1 this PR once that's followed by adding a feature guard in the public header.  My OE_EXPERIMENTAL comment is about following the above requirement to have a guard since this is just a piece of a whole feature (sample being another piece).

host/hostthread.h
dthaler: Ah, agreed.   Non-public APIs really should too, but not a blocker.

host/sgx/create.c
```diff
@@ -401,6 +402,41 @@ static oe_result_t _initialize_enclave(oe_enclave_t* enclave)
     return result;
 }
 
+/*
+** _config_enclave()
+**
+** Config the enclave with an array of configurations.
+*/
+static oe_result_t _configure_enclave(
```
CodeMonkeyLeet: Future work: Consider minimizing the number of calls between host an enclave as these transitions are expensive, and sets up a pattern for host to call into the enclave for _each_ type of configuration separately. Switchless OCALLs is a bit of a special case since most of this is on the host side, but it might be worth breaking that out so the setup can be done before enclave initialization, and then the manager handle passed to enclave as part of the `_initialize_enclave` call. I'd also imagine that for switchless ECALLs, we'd pass the config struct into the enclave as part of `_initialize_enclave` instead of via yet another separate config ECALL.

enclave/core/shm.c
```diff
@@ -81,31 +81,31 @@ void* oe_shm_malloc(size_t size)
...
-        Shared_memory_pool* current = next;
+        oe_unreserve_arena(next->buffer);
+        Shared_memory_arena* current = next;
         next = next->next;
-        memset(current, 0, sizeof(shm));
+        memset(current, 0, sizeof(arena));
     }
-    _shm_list = NULL;
+    _arena_list = NULL;
 }
```
CodeMonkeyLeet: Missing LF for end of file.

jxyang:
bors try

enclave/core/shm.h
```diff
@@ -6,23 +6,23 @@
...
 
 bool oe_configure_shm_capacity(size_t cap);
 
-void* oe_shm_malloc(size_t size);
+void* oe_arena_malloc(size_t size);
```
CodeMonkeyLeet: >> Clarification: oe_shm_malloc() is responsible for allocation within a predefined arena, not for the arena allocation, correct? 
>>
> oe_shm_malloc implicitly creates the arena if it's not there already

I think part of my original confusion is that the methods don't do a good job of separating operations:
 - **on** the arena: `oe_reserve_arena()`, `oe_unreserve_arena()`, `oe_arena_destroy()`
 - **within** the arena: `oe_arena_malloc()`, `oe_arena_calloc()`, `oe_arena_clear()`

I would suggest the following to clarify:
- Functions managing an arena use the `verb-noun` pattern (standard):
    - `oe_allocate_arena()`/`oe_deallocate_arena()`: I think it should be made clear that the shared memory is not merely reserved as per VirtualAlloc/mmap semantics, but is actually committed for the arena, as evidenced by our use of `oe_host_malloc` in the SGX implementation. 
    - `oe_teardown_arena()`: `destroy` is usually paired with `create` and it's unclear what it means to destroy vs. deallocate an arena. I think `teardown` generally suggests additional work beyond the deallocation (compare `setup` or `initialize` vs. `create` ... `oe_cleanup_arena()` might also be appropriate for the same reasons) 
- Functions for alloc/dealloc in an arena use `arena` as a prefix modifier to the method (pseudoclass model used in the oecrypto methods):
    - `oe_arena_malloc()`, `oe_arena_calloc()` are fine, but the method names strongly suggest the need for an equivalent `oe_arena_free()` method and it's unclear what `oe_arena_clear()` does in relation to that paradigm. I'd suggest changing that to `oe_arena_free_all()` instead, as it expresses the relationship the deallocation relative to the standard malloc/free model. 
    - The lazy allocation of the arena itself via `oe_arena_malloc()` is fine, though it might be worth calling out the expected semantics in comments for the methods.

host/sgx/create.c
```diff
@@ -401,6 +402,41 @@ static oe_result_t _initialize_enclave(oe_enclave_t* enclave)
...
+
+    for (uint32_t i = 0; i < config_count; i++)
+    {
+        switch (configs[i].config_type)
+        {
+            // Configure the switchless calls, such as the number of workers.
+            case OE_ENCLAVE_CONFIG_CONTEXT_SWITCHLESS:
+            {
+                size_t max_host_workers =
+                    configs[i].u.context_switchless_config->max_enclave_workers;
```
CodeMonkeyLeet: Shouldn't this be 
> configs[i].u.context_switchless_config->max_**host**_workers;

Also, until we implement support for using the `max_enclave_workers` member, especially if we are shipping v0.7 without support for it, we should explicitly block `oe_create_enclave()` with non-zero `max_enclave_workers` and document it as such in the data structure doxygen.

jxyang:
bors try

enclave/core/shm.h
```diff
@@ -6,23 +6,23 @@
 
 #include <openenclave/bits/types.h>
 
-typedef struct _shared_memory_pool
+typedef struct _shared_memory_arena
```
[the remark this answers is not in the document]
jxyang: Nice catch, will change the file names.

This is a shared memory arena dedicated to parameter passing purpose. I can't rule out the possibility that in the future we have an arena for enclave memory or host memory. I think being specific doesn't hurt in this case.

enclave/core/shm.h
```diff
@@ -6,23 +6,23 @@
...
     /* Buffer holding the shared memory pool */
     uint8_t* buffer;
     size_t capacity;
     size_t used;
-    struct _shared_memory_pool* next;
-} Shared_memory_pool;
+    struct _shared_memory_arena* next;
+} Shared_memory_arena;
 
 bool oe_configure_shm_capacity(size_t cap);
```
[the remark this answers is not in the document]
jxyang: Will rename it. The reason this is not called anywhere currently is because the default arena size 1 mb is sufficient in most cases. But we do need this API to adjust the arena size for functions with gigantic sized parameters, say `char[1000000]`.

host/sgx/create.c
jxyang: Good point. I considered the option of using `_initialize_enclave` and its corresponding ECALL `OE_ECALL_INIT_ENCLAVE` to initialize switchless. But there are limitations for that:

* The function `_initialize_enclave` simply sets a opaque handle to the enclave created by the host. Therefore even if we set the switchess configurations, such as the switchless manager, in `oe_enclave_t` and pass the enclave instance to this function, it still can't access the switchless configurations.  
* We could combine the ECALLs that have to be made during `_initialize_enclave` and `configure_enclave`. This depends on that we have a full view of the `oe_enclave_t` instance inside the enclave. Not sure whether the opaque view is by design.

enclave/core/shm.h
jxyang: Nice. Will do.

host/sgx/create.c
jxyang: Ahh, we get away with this typo because we set both max_host_workers and max_enclave_workers to 2. Thanks for catching that.

include/openenclave/host.h
jxyang: @dthaler Just to clarify, the public API changes in host.h can be split into two types:

1. The added structs related to switchless configuartions, which could be changed frequently.
2. The change to `oe_create_enclave`, where we changed the type of one parameter `configs` from `void*` to `oe_enclave_config_t`, and renamed a couple of other parameter names. This seems to be a stable change.

Do you want `OE_EXPERIMENTAL` to be applied on both or just the former?  If you want both, that means we need two versions of `oe_create_enclave` in this header.

jxyang:
bors try

include/openenclave/host.h
dthaler: Only this file, and I would say yes to 1, and 2 is debatable... really oe_enclave_config_t needs a sample which is why I would say probably both.  But you don't need two full versions of oe_create_enclave per se.  For example,

oe_result_t oe_create_enclave( 
    const char* path,
    oe_enclave_type_t type,
    uint32_t flags,
#ifdef OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE
    const oe_enclave_config_t* configs,
    uint32_t config_count,
#else
    const void* config,
    uint32_t config_size,
#endif
    const oe_ocall_func_t* ocall_table,
    uint32_t ocall_table_size,
    oe_enclave_t** enclave);   

You then remove the #ifdefs in the PR that adds the sample (or later if you wish).

jxyang: If we keep two versions of oe_create_enclave, the tool oeedger8r has to generate two versions of calls into this function depending on whether `OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE` is on or off. How is that possible is beyond me since I am not familiar with the tool.

@anakrish @andschwa Could you help?

With the suggested changes, we need to modify a few places where oe_create_enclave does take an `oe_enclave_config_t*` pointer as parameter. Plus `host/sgx/create.c` in which we handle this parameter.

anakrish: @jxyang oeedger8r does not need to generate two versions; rather it just needs to generate a single version that compiles differently based on whether OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE is defined or not.

This PR already changes
https://github.com/openenclave/openenclave/blob/e7a8a23c136e88fb5960a6836afa43a6e19b1659/tools/oeedger8r/Emitter.ml#L1367-L1373

You just need to shove in the OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE as well:

```
    ; sprintf "oe_result_t oe_create_%s_enclave(" ec.enclave_name
    ; "    const char* path,"
    ; "    oe_enclave_type_t type,"
    ; "    uint32_t flags,"
    ; "#ifdef OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE"
    ; "    const oe_enclave_config_t* configs,"
    ; "    uint32_t config_count,"
    ; "#else"
    ; "    const void* config,"
    ; "    uint32_t config_size,"
    ; "#endif"
    ; "    oe_enclave_t** enclave);"
```

Similarly the signature at the function definition also needs to be changed.

jxyang: Cool. Thanks, Anand.

dthaler: great question.  Agree with Anand that's one way to do it (there may be others).   Since we're setting a precedent for other new features to follow, I'm in favor of variants that have the #ifdef in as few files as possible.  1 would be ideal.  Maybe 2 is the best we can do here.

enclave/core/arena.c
```diff
@@ -0,0 +1,115 @@
...
+void* oe_arena_calloc(size_t num, size_t size)
+{
+    size_t total = 0;
+    if (oe_safe_mul_sizet(num, size, &total) != OE_OK)
+        return NULL;
+
+    void* ptr = oe_arena_malloc(total);
+    if (ptr != NULL)
+    {
+        memset(ptr, 0, size);
```
anakrish: `memset(ptr, 0, total);`

enclave/core/arena.c
```diff
@@ -0,0 +1,115 @@
...
+
+// the global list of shared memory pools
+static shared_memory_arena_t* _arena_list = NULL;
+
+static oe_spinlock_t _arena_list_lock = OE_SPINLOCK_INITIALIZER;
+
+// default shared memory pool capacity is 1 mb
+size_t capacity = 1024 * 1024;
+
+size_t max_capacity = 1 << 30;
```
anakrish: This can be made a constant if it is not expected to change.

enclave/core/sgx/calls.c
```diff
@@ -421,6 +418,11 @@ static void _handle_ecall(
             arg_out = _handle_init_enclave(arg_in);
             break;
         }
+        case OE_ECALL_INIT_CONTEXT_SWITCHLESS:
+        {
+            arg_out = oe_handle_init_switchless(arg_in);
```
anakrish: Since switchless is initialized after enclave has been initialized, any switchless calls that are made by global constructors (constructors of global C++ objects or functions with __constructor__ attribute) within the enclave will not work.

jxyang: As agreed by Anand, we will track this in an issue and fix it later.

https://github.com/openenclave/openenclave/issues/2180

include/openenclave/host.h
jxyang: Thanks for the clarification. @dthaler.

Thanks for the suggestion, @anakrish.

Feature flag `OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE` is added.

jxyang:
bors try

jxyang:
Bors try

jxyang:
bors try

jxyang:
bors try

jxyang:
bors try

include/openenclave/bits/result.h
```diff
@@ -314,6 +314,21 @@ typedef enum _oe_result
...
+
+    /**
+     * Thread creation failed.
+     */
+    OE_THREAD_CREATE_ERROR,
+
+    /**
+     * Thread join failed.
+     */
+    OE_THREAD_JOIN_ERROR,
```
anakrish: This error does not seem to be raised anywhere

tests/create-errors/host/host.c
```diff
@@ -36,11 +45,25 @@ static void _test_invalid_param(const char* path, uint32_t flags)
...
+#ifdef OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE
+    /* Invalid configuration with incorrect **config_count** */
+    result = oe_create_create_errors_enclave(
+        path, OE_ENCLAVE_TYPE_SGX, flags, &invalid_config, 0, &enclave);
+
+    OE_TEST(result == OE_INVALID_PARAMETER);
+
+    /* Invalid configuration with correct **config_count** */
+    result = oe_create_create_errors_enclave(
+        path, OE_ENCLAVE_TYPE_SGX, flags, &invalid_config, 1, &enclave);
```
anakrish: invalid_config is not being initialized. Therefore this test is unreliable.

include/openenclave/host.h
```diff
@@ -64,6 +65,60 @@ typedef void (*oe_ocall_func_t)(
...
+    /**
+     * The type of the configuration in **u**
+     */
+    oe_enclave_config_type_t config_type;
+    /**
+     * The specific configuration for the enclave, such as configuring
+     * context-switchless calls.
+     */
+    union {
+        const oe_enclave_config_context_switchless_t* context_switchless_config;
```
anakrish: What is the rationale for a union of struct pointers instead of a union of structs?
My recollection of the discussions was that u would be a union of config structures.

@mikbras @dthaler

enclave/core/arena.c
```diff
@@ -0,0 +1,115 @@
...
+    return ptr;
+}
+
+void oe_arena_free_all()
+{
+    arena.used = 0;
+}
+
+// Free all shared memory pools in the global list
+void oe_teardown_arena()
```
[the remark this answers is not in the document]
jxyang: We need this to free the host memory, that happens when the ECALL returns.

enclave/core/arena.c
```diff
@@ -0,0 +1,115 @@
...
+    {
+        void* buffer = oe_allocate_arena(capacity);
+        if (buffer == NULL)
+            return NULL;
+        arena.buffer = (uint8_t*)buffer;
+        arena.capacity = capacity;
+        arena.used = 0;
+
+        // add the newly created pool to the global list
+        oe_spin_lock(&_arena_list_lock);
```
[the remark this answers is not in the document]
jxyang: That's right if we change the semantic of `oe_teardown_arena`. It used to be tearing down arenas on all threads. Now we need to limit to the arena of the current thread. Will do.

enclave/core/sgx/calls.c
```diff
@@ -585,9 +593,27 @@ oe_result_t oe_call_host_function_by_table_id(
...
+
+        // Fall back to regular OCALL if host worker threads are unavailable
+        if (post_result == OE_CONTEXT_SWITCHLESS_OCALL_MISSED)
+            OE_CHECK(
+                oe_ocall(OE_OCALL_CALL_HOST_FUNCTION, (uint64_t)args, NULL));
+        else
+        {
+            OE_CHECK(post_result);
+            // Wait until args.result is set by the host worker.
+            while (__atomic_load_n(&args->result, __ATOMIC_SEQ_CST) ==
```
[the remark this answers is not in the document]
jxyang: `args->result` is an integer. The atomic load should ensure we are not reading partial values.

Am I missing something?

include/openenclave/host.h
jxyang: If `u`  is a union of config structures, the struct `oe_enclave_config_t` doesn't have a fixed size, and we won't fit different types of configs into an array.

tests/create-errors/host/host.c
jxyang: Fixed

include/openenclave/bits/result.h
jxyang: Now captured in `oe_stop_worker_threads`

enclave/core/sgx/calls.c
```diff
@@ -585,9 +593,27 @@ oe_result_t oe_call_host_function_by_table_id(
...
+                   __OE_RESULT_MAX)
```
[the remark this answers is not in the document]
jxyang: Since this is a `post-poll` pattern, I expect the poster sets the flag, and the poller checks the flag.

host/sgx/switchless.c
```diff
@@ -0,0 +1,125 @@
...
+**
+*/
+static void* _switchless_ocall_worker(void* arg)
+{
+    oe_host_worker_context_t* context = (oe_host_worker_context_t*)arg;
+
+    while (!context->is_stopping)
+    {
+        volatile oe_call_host_function_args_t* local_call_arg = NULL;
+        if ((local_call_arg = context->call_arg) != NULL)
```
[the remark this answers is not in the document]
jxyang: We will use futex to address the synchronization in a follow-up PR.

enclave/core/switchlesscalls.c
```diff
@@ -0,0 +1,140 @@
...
+**
+**  Post the function call (wrapped in args) to a free host worker thread
+**  by writing to its context.
+**
+**==============================================================================
+*/
+oe_result_t oe_post_switchless_ocall(oe_call_host_function_args_t* args)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    args->result = __OE_RESULT_MAX; // Means the call hasn't been processed.
```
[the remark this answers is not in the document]
jxyang: Please see my question above about partial read.

enclave/core/arena.c
```diff
@@ -0,0 +1,115 @@
...
+// Licensed under the MIT License.
+
+#include "arena.h"
+#include <openenclave/bits/safemath.h>
+#include <openenclave/internal/raise.h>
+#include <openenclave/internal/thread.h>
+#include <openenclave/internal/utils.h>
+#include <string.h>
+
+#define ALIGNMENT (sizeof(uint64_t) * 2)
```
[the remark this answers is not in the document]
jxyang: Fixed

jxyang:
bors try

enclave/core/sgx/calls.c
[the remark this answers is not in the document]
anakrish: Yes. You are missing the acquire-release semantics.
You want the ocall to be completed and **then** set the result to a given value.

C/C++ don't provide statement level ordering guarantees; unless you use acquire-release semantics.

For example,
```c
struct args {
    int result;
    int x;
    int y;
};

void fcn(args* a)
{
    // Set fields
    a->x = 0;
    a->y = 133;

    // Then set the result
    a->result = 0;
}
```

The compiler is legally allowed to change the order of instructions, and rewrite them.
GCC for example does the following:
```c
void fcn(args* a)
{
    // Set result and x
    *(uint64_t*)a = 0;
    
    // Then set y
    a->y = 133;
}
``
Snippet: https://godbolt.org/z/6i3WzB

Even if the compiler preserves the order, the order may not be guaranteed across multiple cores unless you use synchronization instructions.

Switchless implementation requires precise ordering regarding when result is set/read. Thus it requires acquire-release semantics.

File: enclave/core/switchlesscalls.c (hunk 61 above)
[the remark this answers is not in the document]
Comment (anakrish) [reply enclave/core/switchlesscalls.c]: Here is a good link: https://preshing.com/20120913/acquire-and-release-semantics/

Also see use of OE_ATOMIC_MEMORY_BARRIER_ACQUIRE

File: include/openenclave/host.h (hunk 55 above)
Comment (anakrish) [reply include/openenclave/host.h]: If it is a union of config structures, the struct will still have a fixed size. Additionally one would be able to initialize the struct in place.

File: enclave/core/sgx/calls.c (hunk 58 above)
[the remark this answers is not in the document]
Comment (jxyang) [reply enclave/core/sgx/calls.c]: Oh you are talking about partial values in `args`, not `args-result` itself. Make sense. Will fix.

File: enclave/core/switchlesscalls.c (hunk 61 above)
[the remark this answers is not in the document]
Comment (jxyang) [reply enclave/core/switchlesscalls.c]: Thanks for the pointer. Fixed.

File: include/openenclave/host.h (hunk 55 above)
Comment (jxyang) [reply include/openenclave/host.h]: We will add more configuration types. Not all of them will have the same length as `oe_enclave_config_context_switchless_t`. Using pointers, not the structure itself, in the union, ensures the union size is fixed, hence the size of `oe_enclave_config_t` is fixed .

**Comment by jxyang** [comment]
bors try

File: include/openenclave/host.h (hunk 55 above)
Comment (jxyang) [reply include/openenclave/host.h]: The size of union will always be the size of its largest member. You are right that `oe_enclave_config_t` is size fixed even if we put config structures, instead of pointers, in it. Now I am not sure what was the reason we decided to put pointers in the union.

@dthaler @mikbras

**Comment by jxyang** [comment]
bors try

File: include/openenclave/host.h (hunk 55 above)
Comment (dthaler) [reply include/openenclave/host.h]: The only reason to use pointers is just memory. A union is the size of its largest member.  So if you have a small member, and a large member, then an large array of small members will be proportionately larger.  With pointers, you waste the size of a pointer per array element, but that's all.

File: enclave/core/sgx/calls.c [hunk 63]
```diff
@@ -585,9 +593,31 @@ oe_result_t oe_call_host_function_by_table_id(
...
+            OE_CHECK(
+                oe_ocall(OE_OCALL_CALL_HOST_FUNCTION, (uint64_t)args, NULL));
+        else
+        {
+            OE_CHECK(post_result);
+            // Wait until args.result is set by the host worker.
+            while (true)
+            {
+                OE_ATOMIC_MEMORY_BARRIER_ACQUIRE();
+                if (__atomic_load_n(&args->result, __ATOMIC_SEQ_CST) !=
```
anakrish: Do you think we should use the memorder parameter of __atomic_load_n instead of calling OE_ATOMIC_MEMORY_BARRIER_ACQUIRE explicitly?

enclave/core/switchlesscalls.c
```diff
@@ -0,0 +1,140 @@
...
+**==============================================================================
+*/
+oe_result_t oe_handle_init_switchless(uint64_t arg_in)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    oe_switchless_call_manager_t* manager = NULL;
+    oe_switchless_call_manager_t safe_manager;
+    size_t contexts_size, threads_size;
+
+    if (arg_in == 0)
```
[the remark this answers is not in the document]
anakrish: Can you address this too?

enclave/core/switchlesscalls.c
```diff
@@ -0,0 +1,143 @@
...
+**  Post the function call (wrapped in args) to a free host worker thread
+**  by writing to its context.
+**
+**==============================================================================
+*/
+oe_result_t oe_post_switchless_ocall(oe_call_host_function_args_t* args)
+{
+    oe_result_t result = OE_UNEXPECTED;
+
+    OE_ATOMIC_MEMORY_BARRIER_RELEASE();
```
anakrish: Great!

include/openenclave/host.h
anakrish: Sounds good.

include/openenclave/internal/atomic.h
```diff
@@ -38,4 +42,36 @@ OE_INLINE uint64_t oe_atomic_decrement(volatile uint64_t* x)
...
 
+OE_INLINE
+bool oe_atomic_compare_and_swap(
+    int64_t volatile* dest,
+    int64_t old,
+    int64_t newval)
+{
+#if defined(__GNUC__)
+    return __atomic_compare_exchange_n(
+        dest, &old, newval, 1, __ATOMIC_ACQ_REL, __ATOMIC_ACQUIRE);
```
anakrish: Maybe it is better to use __ATOMIC_ACQ_REL to provide both acquire and release semantics?
Otherwise from the function name `oe_atomic_compare_and_swap` it is not clear that it provides only acquire semantics.

include/openenclave/internal/atomic.h
```diff
@@ -38,4 +42,36 @@ OE_INLINE uint64_t oe_atomic_decrement(volatile uint64_t* x)
...
+
+OE_INLINE
+bool oe_atomic_compare_and_swap_ptr(
+    void* volatile* dest,
+    void* old,
+    void* newptr)
+{
+#if defined(__GNUC__)
+    return __atomic_compare_exchange_n(
+        dest, &old, newptr, 1, __ATOMIC_ACQ_REL, __ATOMIC_ACQUIRE);
```
anakrish: Same as above.

tools/oeedger8r/Emitter.ml
```diff
@@ -1427,16 +1432,26 @@ let gen_enclave_code (ec : enclave_content) (ep : edger8r_params) =
...
     ; "    const void* config,"
     ; "    uint32_t config_size,"
+    ; "#endif"
     ; "    oe_enclave_t** enclave)"
     ; "{"
     ; "    return oe_create_enclave("
     ; "               path,"
     ; "               type,"
     ; "               flags,"
+    ; "#ifdef OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE"
```
anakrish: We need to add the release barrier in emitted code when the arg struct's result field is set.

enclave/core/arena.c
```diff
@@ -0,0 +1,98 @@
...
+#include <openenclave/internal/raise.h>
+#include <openenclave/internal/thread.h>
+#include <openenclave/internal/utils.h>
+#include <string.h>
+
+// the per-thread shared memory pool
+static __thread shared_memory_arena_t arena = {0};
+
+// default shared memory pool capacity is 1 mb
+static size_t capacity = 1024 * 1024;
```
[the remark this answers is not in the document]
jxyang: Fixed

enclave/core/arena.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include <openenclave/bits/types.h>
+
+typedef struct _shared_memory_arena_t
+{
+    /* Buffer holding the shared memory pool */
+    uint8_t* buffer;
+    size_t capacity;
+    size_t used;
+    struct _shared_memory_arena_t* next;
```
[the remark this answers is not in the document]
jxyang: Fixed

enclave/core/sgx/calls.c
jxyang: They are functionally the same - probably translated into nop in both cases. I don't have preference one over the other.

include/openenclave/internal/atomic.h
jxyang: The memory model for the success case is indeed __ATOMIC_ACQ_REL.

The last parameter is for memory model of the failure case when the exchange doesn't happen. The gcc document says:

> This memory model cannot be __ATOMIC_RELEASE nor __ATOMIC_ACQ_REL. It also cannot be a stronger model than that specified by success_memmodel.

tools/oeedger8r/Emitter.ml
jxyang: There are two `result` fields in two separate structures. The emitted code sets the `result` field in `host_<ocall_function_name>_args_t`. But that's not the sync variable between the host thread and the enclave thread.  

I did add the write barrier in `oe_handle_call_host_function` before setting `oe_call_host_function_args_t.result`, the sync variable we do care about.

jxyang:
@anakrish I created an [issue](https://github.com/openenclave/openenclave/issues/2187) for initializing switchless more than once; fixed the naming issues; and confirmed we don't need to insert memory barriers in edger8r. Let me know if you still have concerns.

jxyang:
bors try

jxyang:
bors try

tools/oeedger8r/Emitter.ml
```diff
@@ -1368,8 +1368,13 @@ let gen_enclave_code (ec : enclave_content) (ep : edger8r_params) =
     ; "    const char* path,"
     ; "    oe_enclave_type_t type,"
     ; "    uint32_t flags,"
+    ; "#ifdef OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE"
```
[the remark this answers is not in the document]
jxyang: Yes, @dthaler proposed a snippet that has this definition in `oe_create_enclave`. It would have been cleaner if we have enclave configuration as a separate PR before this PR. But we have to deal with the fact that enclave configuration was not implemented. We have some thoughts, which were never turned into code.

PR #2142, opened 2019-08-30
Improve pre-commit hook

andschwa:
This PR fixes the pre-commit hook from failing due to `git rm`'d files, and `git commit --amend` with no edited files (or `git commit --allow-empty`).

### DIFF
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -3,6 +3,10 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+# TODO: Do not include deleted files
+# TODO: Skip hook if files is empty
+# TODO: Check authorship against blacklist (e.g. disallow "Ubuntu")
+
 set -o errexit
 
 scripts=$(git rev-parse --show-toplevel)/scripts
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -3,13 +3,9 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-# TODO: Skip hook if files is empty
 # TODO: Check authorship against blacklist (e.g. disallow "Ubuntu")
 
 set -o errexit
-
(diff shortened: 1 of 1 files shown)


andschwa:
I tested manually, and it seems to work properly. However, it's difficult to add automated tests for this. @mikbras can you check if it works for your use cases?

andschwa:
@johnkord I am still working on checking the Git authors against a blacklist. It's not trivial. If you do a `git shortlog --summary --numbered --email` you'll see that we have a lot of different bad user commits. I think most of them could be avoided with the heuristic of requiring the Git author name to be at least two words (as this would exclude usernames as author name as well as "root" and "Ubuntu").

However, there's still the caveat that you can edit and commit directly on GitHub, which can result in authorships that are the GitHub user name, and use the GitHub "noreply" email, which we also want to avoid. Checking that, however, will require a CI-time check against the _new_ commits in a PR (that is, not in this pre-commit hook, as GitHub doesn't run it).

The other issue is erroring and providing proper guidance...might be good enough to say "Please set your Git user to your full name" and point to the contributing guide.

Then an open question remains: what about emails?

andschwa:
Okay, at least the pre-commit hook now checks that the Git user name is set to at least two words. That should avoid "root" and "Ubuntu" and user names.

scripts/pre-commit
```diff
@@ -17,6 +14,29 @@ exit_() {
     exit 1
 }
 
+if [[ $(git config --get user.name | wc -w) -lt 2 ]]; then
+    # This heuristic avoids bad user names such as "root" or "Ubuntu"
+    # or a computer login name. A full name should (usually) have at
+    # least two words. We can change this if needed later.
```
johnkord: So far we don't have any single name contributors, so this is fine with me. We can change it if that situation changes, but given that single name individuals are rare we probably can just make exceptions for them individually.

andschwa:
bors r+

Thanks @johnkord!

PR #2149, opened 2019-09-05
Fix polymorphic exceptions caught by value

jhand2:
When compiling with gcc 8 (build defaults to gcc when clang is not installed) with -Wall, catching polymorphic exceptions by value is treated as a build error. This PR fixes all instances which cause a build using gcc 8 to fail.

In general it is good practice to catch C++ exceptions by reference, even if they don't fail the build. See this blog post for why: https://blog.knatten.org/2010/04/02/always-catch-exceptions-by-reference/

### DIFF
--- a/tests/cppException/enc/cppException.cpp
+++ b/tests/cppException/enc/cppException.cpp
@@ -177,7 +177,7 @@ bool BasicVerification()
     {
         return false;
     }
-    catch (BarClass01 ex_obj)
+    catch (BarClass01& ex_obj)
     {
         if (ex_obj.i != 0XFFFF)
         {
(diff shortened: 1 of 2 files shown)


tests/cppException/enc/cppException.cpp
```diff
@@ -247,7 +247,7 @@ bool BasicVerification()
     {
         return false;
     }
-    catch (BarClass02 ex_obj)
+    catch (BarClass02& ex_obj)
```
CodeMonkeyLeet: I think this was intentionally a catch by value based on the test comment, since it's looking for an exact class match.

Conversely, it seems like `BarClass01` should also be caught by reference for lines 283, 321 and 424 (for consistency with the updated catch on line 412), even though it doesn't matter for an exception class with no inheritance. i.e. the tests seem to checking for the catching of the correct type rather than mechanism of catching (as in lines 162/194) and catching consistently by reference eliminates an extra variable.

jhand2: Updated PR to address this but squashed into the same commit.

jhand2:
Thanks @CodeMonkeyLeet. Unless @mikbras has additional comments, can you please merge?

PR #2154, opened 2019-09-09
Add explicit OEHostVerify package dependencies

BRMcLaren:
Fixes a bug on Non-SGX enabled machines with dependency management, blocking the installation of the SDK.


BRMcLaren:
bors+

PR #2156, opened 2019-09-09
Update maintainers list

andschwa:
Per this morning's CMC meeting, we all agreed to remove Li Li and Pushkar Chitnis as committers, fix Brett McLaren's committer area, and add Anand Krishnamoorthi as a maintainer.


andschwa:
This is kind of an experiment, but I also made a @openenclave/maintainers GitHub (similar to @ openenclave/committers, which I don't want to @ mention and notify). Open question: should "maintainers" be a child team of "committers?" Here's [more about GitHub Teams](https://help.github.com/en/articles/about-teams).

PR #2163, opened 2019-09-11
Add proposed API guidelines

dthaler:
Adapted from the guidelines already used in IoTivity, another LF project

### DIFF
--- a/docs/ApiGuidelines.md
+++ b/docs/ApiGuidelines.md
@@ -24,9 +24,9 @@ Public API requirements:
   where shared libraries are build, public APIs cannot have binary-breaking changes across 
   Open Enclave versions either. (Source- and binary-breaking changes can still be made to a new 
   API before it is released for the first time.)
-* A public API must not be added to Iotivity if there is already a non-deprecated API that can just 
+* A public API must not be added to Open Enclave if there is already a non-deprecated API that can just 
   as easily be used (i.e., with approximately the same number of lines of code).
-* A public API must not be added to Iotivity if the API would provide generic functionality
+* A public API must not be added to Open Enclave if the API would provide generic functionality
   (i.e., not Open Enclave specific functionality) that could be provided by an external library.
 * All new Public C API names must start with oe_.
 * All public headers must follow Open Enclave [DevelopmentGuide.md](coding conventions).
@@ -69,7 +69,7 @@ Internal API requirements:
 * Public APIs cannot have source-breaking changes (and, for platforms where shared libraries or built,
   cannot have binary-breaking changes) across releases, except as covered by deprecation as explained
   below. API additions can be made at any time.
-*  xperimental and internal APIs can have breaking changes across releases.
+* Experimental and internal APIs can have breaking changes across releases.
 
 APIs can be deprecated by marking them as @deprecated. The associated text should explain what
 an application should do instead. Public APIs must be @deprecated in a release before they can be
--- a/docs/ApiGuidelines.md
+++ b/docs/ApiGuidelines.md
@@ -21,7 +21,7 @@ Public API requirements:
   should be one that is compiled as part of the build, and uses no experimental APIs. For rarely 
   used public APIs, the sample might simply be a snippet in the doxygen comments.
 * Public APIs cannot have source-breaking changes across Open Enclave versions. For platforms 
-  where shared libraries are build, public APIs cannot have binary-breaking changes across 
+  where shared libraries are built, public APIs cannot have binary-breaking changes across 
   Open Enclave versions either. (Source- and binary-breaking changes can still be made to a new 
   API before it is released for the first time.  See the Breaking Changes section below for further discussion.)
 * A public API must not be added to Open Enclave if there is already a non-deprecated API that can just 
@@ -45,7 +45,7 @@ Experimental API requirements:
   code size requirement can be minimized.
 * Experimental APIs must be in a header file that is published to the SDK output directory. They
   can be in their own header, or in a header file that is shared with public APIs but only
-  if surrounded by an appropriate ifdef (e.g., "#ifdef OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE").
+  if surrounded by an appropriate ifdef (e.g., "#ifdef OE\_CONTEXT\_SWITCHLESS\_EXPERIMENTAL\_FEATURE").
 * Experimental APIs should be fully documented using doxygen markup in the header file
 * Experimental APIs must have at least one test case.
 * Experimental APIs should ideally be used in some sample code (separate from samples that
--- a/docs/ApiGuidelines.md
+++ b/docs/ApiGuidelines.md
@@ -1,0 +1,76 @@
+# API Guidelines
+
+## APIs
+
+There are three types of APIs: 
+
+1. Public
+2. Experimental
+3. Internal
+
+### Public APIs
+
+Public APIs are the supported APIs that apps can use. 
+
+Public API requirements: 
+
+* All public APIs must have at least one test case.
+* All public APIs must be in a header file that is published to a location under the build directory.
+* All public APIs must be fully documented using doxygen markup in the public header files
+* All public APIs must be used in some sample code or snippet. Ideally, the sample
+  should be one that is compiled as part of the build, and uses no experimental APIs. For rarely 
+  used public APIs, the sample might simply be a snippet in the doxygen comments.
+* Public APIs cannot have source-breaking changes across Open Enclave versions. For platforms 
+  where shared libraries are build, public APIs cannot have binary-breaking changes across 
+  Open Enclave versions either. (Source- and binary-breaking changes can still be made to a new 
+  API before it is released for the first time.)
+* A public API must not be added to Iotivity if there is already a non-deprecated API that can just 
+  as easily be used (i.e., with approximately the same number of lines of code).
+* A public API must not be added to Iotivity if the API would provide generic functionality
+  (i.e., not Open Enclave specific functionality) that could be provided by an external library.
+* All new Public C API names must start with oe_.
+* All public headers must follow Open Enclave [DevelopmentGuide.md](coding conventions).
+
+### Experimental APIs
+
+Experimental APIs are considered to be potential future public APIs, but there is no guarantee 
+of support in future releases, nor is there any guarantee that breaking changes will not occur 
+across releases. 
+
+Experimental API requirements: 
+
+* Experimental APIs must be optional to compile in. That is, there must be some way to build
+  Open Enclave in a way that does not expose any experimental APIs, so that an application
+  can easily verify that it uses only fully-supported public APIs, and so that the Open Enclave
+  code size requirement can be minimized.
+* Experimental APIs must be in a header file that is published to the build directory. They
+  can be in their own header, or in a header file that is shared with public APIs but only
+  if surrounded by an appropriate ifdef.
+* Experimental APIs should be fully documented using doxygen markup in the header file
+* Experimental APIs should ideally be used in some sample code (separate from samples that
+  use only public APIs) or snippet. Ideally, the sample should be one that is compiled as part
+  of the build. For rarely used APIs, the sample might simply be a snippet in the doxygen
+  comments. Having working samples may help others evaluate and determine how ready it is for
+  prime time before the API moves from experimental to public.
+
+### Internal APIs
+
+Internal APIs are considered to be usable within Open Enclave but are not intended to be
+used by applications that use Open Enclave. 
+
+Internal API requirements: 
+
+* Internal APIs must not appear in any header file that is published to the build directory
+* Internal APIs should ideally be documented using doxygen markup in the private header files
+* Internal APIs must not be used in sample code
+
+## Breaking Changes
+
+* Public APIs cannot have source-breaking changes (and, for platforms where shared libraries or built,
+  cannot have binary-breaking changes) across releases, except as covered by deprecation as explained
+  below. API additions can be made at any time.
+*  xperimental and internal APIs can have breaking changes across releases.
+
+APIs can be deprecated by marking them as @deprecated. The associated text should explain what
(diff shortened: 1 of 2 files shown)


johnkord:
I will create an item for the next CGC agenda for this PR.

docs/ApiGuidelines.md
```diff
@@ -0,0 +1,76 @@
...
+* All public APIs must have at least one test case.
+* All public APIs must be in a header file that is published to a location under the build directory.
+* All public APIs must be fully documented using doxygen markup in the public header files
+* All public APIs must be used in some sample code or snippet. Ideally, the sample
+  should be one that is compiled as part of the build, and uses no experimental APIs. For rarely 
+  used public APIs, the sample might simply be a snippet in the doxygen comments.
+* Public APIs cannot have source-breaking changes across Open Enclave versions. For platforms 
+  where shared libraries are build, public APIs cannot have binary-breaking changes across 
+  Open Enclave versions either. (Source- and binary-breaking changes can still be made to a new 
+  API before it is released for the first time.)
```
achamayou: When you say versions, do you mean all versions, major revisions? Committing to eternal binary compatibility for .so sounds very onerous, I think at least major releases should allow this to be revisited (likewise for source compatibility, really).

achamayou: Nevermind, I should have read the bit about Breaking below. It might make sense to mention the exception here as well, in case someone reads just this bullet point and gets the wrong impression.

docs/ApiGuidelines.md
```diff
@@ -0,0 +1,76 @@
...
+* Internal APIs must not be used in sample code
+
+## Breaking Changes
+
+* Public APIs cannot have source-breaking changes (and, for platforms where shared libraries or built,
+  cannot have binary-breaking changes) across releases, except as covered by deprecation as explained
+  below. API additions can be made at any time.
+* Experimental and internal APIs can have breaking changes across releases.
+
+APIs can be deprecated by marking them as @deprecated. The associated text should explain what
```
achamayou: I think it would be nice to commit to some sort of minimal time duration for deprecation. In a situation where releases happen relatively rapidly in a row, users could find themselves on an now-removed API, and unable to upgrade to a version with security fixes on quite short notice.

Guaranteeing that deprecated public APIs stick around for at least X amount of time before going away is a good way to avoid that.

docs/ApiGuidelines.md
```diff
@@ -0,0 +1,76 @@
...
+
+* Internal APIs must not appear in any header file that is published to the build directory
+* Internal APIs should ideally be documented using doxygen markup in the private header files
+* Internal APIs must not be used in sample code
+
+## Breaking Changes
+
+* Public APIs cannot have source-breaking changes (and, for platforms where shared libraries or built,
+  cannot have binary-breaking changes) across releases, except as covered by deprecation as explained
+  below. API additions can be made at any time.
```
johnkord: Do we have a stated amount of time (or number of releases) that we will take to deprecate and then remove a public API? Of course, if a security issue appears with one of our public APIs we'll deprecate it and remove it immediately. Maybe we should we leave this unstated and leave it up to the committers to decide.

docs/ApiGuidelines.md
```diff
@@ -0,0 +1,76 @@
...
+* Experimental and internal APIs can have breaking changes across releases.
+
+APIs can be deprecated by marking them as @deprecated. The associated text should explain what
+an application should do instead. Public APIs must be @deprecated in a release before they can be
+removed.
```
johnkord: Does this mean that a public API can't be removed until after its deprecation has been announced first in a given release? Do patch releases count, or only minor/major version releases? Any exceptions to this rule, like for possible security concerns?

docs/ApiGuidelines.md
```diff
@@ -0,0 +1,76 @@
...
+3. Internal
+
+### Public APIs
+
+Public APIs are the supported APIs that apps can use. 
+
+Public API requirements: 
+
+* All public APIs must have at least one test case.
+* All public APIs must be in a header file that is published to a location under the build directory.
```
gupta-ak: It seems that the "build directory" term means something different here than what it does in CMake, so I think someone might get confused. In CMake, the intermediaries and results of the build go to the build directory, so headers generally won't be there. To add something to the SDK package, you would add it to the "install" directory, like here:
`install(FILES openenclave/enclave.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/openenclave/)`.

Maybe change "build directory" to something like "SDK output directory" and if you want to mention CMake specifically, you can also specify it must be under  `${CMAKE_INSTALL_INCLUDEDIR}/openenclave`.

docs/ApiGuidelines.md
```diff
@@ -0,0 +1,76 @@
...
+Experimental API requirements: 
+
+* Experimental APIs must be optional to compile in. That is, there must be some way to build
+  Open Enclave in a way that does not expose any experimental APIs, so that an application
+  can easily verify that it uses only fully-supported public APIs, and so that the Open Enclave
+  code size requirement can be minimized.
+* Experimental APIs must be in a header file that is published to the build directory. They
+  can be in their own header, or in a header file that is shared with public APIs but only
+  if surrounded by an appropriate ifdef.
+* Experimental APIs should be fully documented using doxygen markup in the header file
```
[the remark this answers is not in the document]
dthaler: Found an interesting comparable: https://github.com/blynkkk/blynk-library/commit/579545d1662c6285dc14cb145ce77b36ee9573fa

See how it uses @experimental in doxygen after putting into doxygen.config the following:

```
ALIASES                += experimental="\xrefitem experimental \"This feature is marked as experimental\" \"Experimental List\"\n"
ALIASES                += experimental{1}="\xrefitem experimental \"Experimental\" \"Experimental List\" \1\n"
```

This uses the custom tags mechanism explained in the stackoverflow thread: https://stackoverflow.com/questions/537043/custom-tags-with-doxygen

docs/ApiGuidelines.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Experimental APIs
+
+Experimental APIs are considered to be potential future public APIs, but there is no guarantee 
+of support in future releases, nor is there any guarantee that breaking changes will not occur 
+across releases. 
+
+Experimental API requirements: 
+
+* Experimental APIs must be optional to compile in. That is, there must be some way to build
```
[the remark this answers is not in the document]
dthaler: Updated per Andy's suggestion

docs/ApiGuidelines.md
```diff
@@ -0,0 +1,76 @@
...
+  prime time before the API moves from experimental to public.
+
+### Internal APIs
+
+Internal APIs are considered to be usable within Open Enclave but are not intended to be
+used by applications that use Open Enclave. 
+
+Internal API requirements: 
+
+* Internal APIs must not appear in any header file that is published to the build directory
```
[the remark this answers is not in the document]
dthaler: Updated

docs/ApiGuidelines.md
dthaler: Added forward reference

docs/ApiGuidelines.md
dthaler: Agree, but I think that time commitment is a separate issue from the API Guidelines per se.   Recommend tracking via a separate issue.

docs/ApiGuidelines.md
dthaler: > Does this mean that a public API can't be removed until after its deprecation has been announced first in a given release?

Yes, that is the proposal.

>  Do patch releases count, or only minor/major version releases? 

Proposal is only minor/major version releases.

> Any exceptions to this rule, like for possible security concerns?

Proposal is no.   Security concerns might motivate making an API have different behavior (e.g., silent success) in some or all conditions, but not removing the API in a way that would cause compiler errors.   Feel free to argue for an alternative proposal, I'm just answering what the proposal is, based on a precedent from another LF project.

docs/ApiGuidelines.md
```diff
@@ -0,0 +1,76 @@
+# API Guidelines
+
+## APIs
+
+There are three types of APIs: 
+
+1. Public
+2. Experimental
```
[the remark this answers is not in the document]
dthaler: Others argued during a meeting today that we don't need both Experimental and Preview.   That doesn't mean that question is closed, only that I'm not making any changes at this time since I saw no consensus on adding Preview as an api category.  We might consider it a separate issue to track though if you like.

docs/ApiGuidelines.md
andschwa: Yes, I agree with @gupta-ak. This may be slightly confusing as there is existing tech debt removing the manual placement of some libraries/binaries etc. in the build directory (which as Akash explained, is unnecessary as the build directory is _only_ intermediate artifacts, and should just be structured to the default locations used by CMake).

dthaler: Comment is outdated, as GitHub shows.  Current text says "SDK output directory" as suggested.

docs/ApiGuidelines.md
```diff
@@ -0,0 +1,77 @@
...
+Public API requirements: 
+
+* All public APIs must have at least one test case.
+* All public APIs must be in a header file that is published to a location under the SDK output directory.
+* All public APIs must be fully documented using doxygen markup in the public header files
+* All public APIs must be used in some sample code or snippet. Ideally, the sample
+  should be one that is compiled as part of the build, and uses no experimental APIs. For rarely 
+  used public APIs, the sample might simply be a snippet in the doxygen comments.
+* Public APIs cannot have source-breaking changes across Open Enclave versions. For platforms 
+  where shared libraries are build, public APIs cannot have binary-breaking changes across
```
radhikaj: typo, I think this should be "where shared libraries are built"

docs/ApiGuidelines.md
```diff
@@ -0,0 +1,77 @@
...
+
+Public API requirements: 
+
+* All public APIs must have at least one test case.
+* All public APIs must be in a header file that is published to a location under the SDK output directory.
+* All public APIs must be fully documented using doxygen markup in the public header files
+* All public APIs must be used in some sample code or snippet. Ideally, the sample
+  should be one that is compiled as part of the build, and uses no experimental APIs. For rarely 
+  used public APIs, the sample might simply be a snippet in the doxygen comments.
+* Public APIs cannot have source-breaking changes across Open Enclave versions. For platforms
```
radhikaj: Between major versions, I think it should be acceptable that we have public API source-breaking changes. It is not ideal, but is seen in many popular open source libraries.

radhikaj:
@dthaler Could you please take a look at this and decide if you would like to merge it? Thanks.

docs/ApiGuidelines.md
AevaOnline: Breaking public API changes damage the community's trust in a project, so I'd like us to follow a deprecation cycle that's consistent and clear -- and I like that this proposal makes it clear. I suggest that we adopt a strong stance on guaranteeing the stability of public APIs as early as is reasonably possible, but no sooner.

However, if we adhered to [SemVer](https://semver.org/), then a breaking API change -- and that includes the removal of a public stable API -- must always be accompanied by a **major** version change in the API. 

I like the current proposal for now (pre-1.0)

@dthaler could you add a section stating that, when we reach a v1.0, then the project will adopt SemVer?

dthaler: +1 to all of what Aeva said.  How about merging as is and adding such a statement in a separate PR?

docs/ApiGuidelines.md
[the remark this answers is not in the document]
jhand2: Trying to pair down the comments to make sure this is ok to merge. I'm closing this for now.

docs/ApiGuidelines.md
jhand2: @dthaler do we want to specify "between Open Enclave minor versions"?

jhand2: never-mind. Deferring discussion to below.

docs/ApiGuidelines.md
jhand2: I opened #2543 to track this

docs/ApiGuidelines.md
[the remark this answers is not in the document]
dthaler: PR #2364 added doxygen support

docs/ApiGuidelines.md
andschwa: I also agree with @AevaOnline and wanted to share that we already SemVer (with the caveat that we're not yet a 1.0 project): 
https://github.com/openenclave/openenclave/blob/master/CHANGELOG.md#changelog
https://github.com/openenclave/openenclave/blob/master/docs/Releasing.md#version-bump

dthaler:
Triage meeting 2/25/2020 on a different issue resulting in a discussion (not concluded) around the question of: when adding a public API, does it need to be functional on both Windows and Linux?  Or is it ok to only work in one of the two?  If a contributor submits code that only works on Windows, or only works on Linux, what should we do?

docs/ApiGuidelines.md
radhikaj: Can you deprecate in a minor release and then replace in the next major release?

docs/ApiGuidelines.md
radhikaj: I would suggest we document the breaking API change process in more detail
1. Submit a design PR and get it approved. The design PR can minimally state why the old API is problematic and why the new API is needed.
2. Mark API as deprecated in one release only after the design PR is approved.
3. Add new API in the next release.
4. Follow up questions: Can I deprecate in 0.7.2 and replace in 0.8? Can i deprecate in 0.8.2 and replace in 0.8.3?

docs/ApiGuidelines.md
dthaler: https://semver.org/ just says you have to bump the MAJOR version number to remove a public API, which seems fine to me.

dthaler: CGC discussion: when at 0.* then can break changes on MINOR version number, if deprecated in previous MINOR version number.

Issue #2151, opened 2019-09-09
SGX-LKL-PORT: conflict with definition of __cxa_atexit


prp:
`liboecore` exports `__cxa_atexit`, which is needed for OE's C++ support. This symbol, however, conflicts with the implementation in SGX-LKL. SGX-LKL cannot use the OE version of `__cxa_atexit` because it needs an implementation that handles the `vdso` argument correctly.

Ideally, the OE implementation of `__cxa_atexit` should be internal/part of the C++ support only.

mikbras:
PR #2164 addresses this issue.

wintersteiger:
PR #2164 was abandoned because it was redundant. Did that also fix this problem?

jhand2:
This is fixed by https://github.com/openenclave/openenclave/pull/2562

Thanks!

PR #2164, opened 2019-09-11
Relocate __cxa_atexit to oelibc (fixes issue #2151).

mikbras:
This PR fixes a symbol conflict described here (#2151).

It makes two simple changes:
- Renames the definition of **__cxa_atexit** in **oecore** to **oe_cxa_atexit**, and
- Adds **__cxa_atexit** function to **oelibc** (which calls **oe_cxa_atexit**).

This is needed by SGX-LKL which introduces its own C runtime. Any system that replaces **oelibc** would have encountered a similar conflict.

### DIFF
--- a/libc/__cxa_atexit.c
+++ b/libc/__cxa_atexit.c
@@ -4,9 +4,8 @@
 #include "../enclave/core/atexit.h"
 
 /*
- * Registers a function to be called by exit and by oe_call_atexit. This
- * function is used to implement atexit, which calls __cxa_atexit with the
- * following arguments: __cxa_atexit(func, NULL, NULL)
+ * Registers a function to be called by exit and by oe_call_atexit.
+ * This function wraps oe_cxa_atexit, which ignores dso_handle.
  */
 int __cxa_atexit(void (*func)(void*), void* arg, void* dso_handle)
 {
--- a/libc/__cxa_atexit.c
+++ b/libc/__cxa_atexit.c
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#include "../enclave/core/atexit.h"
+#include <openenclave/internal/cxa_atexit.h>
 
 /*
  * Registers a function to be called by exit and by oe_call_atexit.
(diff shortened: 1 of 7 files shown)


libc/__cxa_atexit.c
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "../enclave/core/atexit.h"
```
CodeMonkeyLeet: Do not include headers by relative traversal of the file hierarchy.

For libc implementations that need a specific OE runtime implementation, the corelibc headers already expose `oe_atexit()` which is a thin wrapper around `__cxa_atexit(),` which should have just been redefined to tuse `oe_cxa_atexit()`. If the `__cxa_atexit` symbol needs to be defined in libc outside of its use in `oe_atexit()`, then this implementation can just use the already exposed `oe_atexit()` with no loss of semantics.

If you still really want to expose an additional separate `oe_cxa_atexit()`, it should be broken out into a separate header under `include/openenclave/corelibc/bits` or `include/openenclave/internal`.

mikbras: Thank you for reviewing.

Note that **__cxa_atexit** takes an argument which is eventually passed to **func**, but **atexit** does not. So we definitely need **oe_cxa_atexit**.

I create **<openenclave/internal/cxa_atexit.h>** to house the function declaration.

CodeMonkeyLeet: Changes look good; clean up the PR with the squash and we can merge it.

anakrish:
I am not sure if __cxa_atexit can be replaced without replacing atexit as well.
Functions registered via calls to atexit and __cxa_atexit must be executed in reverse order of their registration and thus can be interspersed.

The following code
```c++
#include <cstdio>
#include <cstdlib>

void fun1()
{
    printf("fun1\n");
}

struct MyClass1 {
    MyClass1()
    {
	atexit(fun1);
    }

    ~MyClass1()
    {
	printf("~MyClass1\n");
    }
};

void fun2()
{
    printf("fun2\n");
}

static MyClass1 g_obj;

int main()
{
    atexit(fun2);
}
```

produces the following output:
```bash
fun2
~MyClass1
fun1
```

I see atexit being defined in corelibc; additionally there are explicit calls to oe_atexit.
Would this break the order of destructors if there are multiple definitions of atexit/oe_atexit?

mikbras:
Closing because this PR is redundant.

PR #2171, opened 2019-09-11
Add developer guidance around abbreviations and doxygen

dthaler:

### DIFF
--- a/docs/DevelopmentGuide.md
+++ b/docs/DevelopmentGuide.md
@@ -11,7 +11,17 @@ Coding Conventions
 * **DO** use `const` and `static` and visibility modifiers to scope exposure of
    variables and methods as much as possible.
 
+* **DO** use doxygen comments, with \[in,out\]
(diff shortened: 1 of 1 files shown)


docs/DevelopmentGuide.md
```diff
@@ -11,8 +11,18 @@ Coding Conventions
 * **DO** use `const` and `static` and visibility modifiers to scope exposure of
    variables and methods as much as possible.
 
+* **DO** use doxygen comments, with \[in,out\]
+  [direction annotation](http://www.doxygen.nl/manual/commands.html#cmdparam) in all public API
```
CodeMonkeyLeet: It seems like we might want to be more prescriptive about its use in more than just this context of in/out. Maybe we should link out to docs/refman/doxygen-howto.md instead of the general doxygen docs and expand that doc to be more prescriptive.

andschwa:
By the way, I think you can just merge this. The changes are reasonable, even if we want to expand the Doxygen how-to (which we could do in another PR).

Issue #2147, opened 2019-09-04
Logging format


olgavrou:
Hi!

In trying to make use of the logs (e.g. via uploading them to Azure Log Analytics) it is nice to be able to parse and display all of the available logs including the ones generated by OE.

I have played around with the formatting of the OE logs a bit and without too much code modification would like to propose the below changes to the logging

(1) User defined format

It would be very useful for users to be able to define their own format given the log line info. I propose doing this when logging to a user specified file via the existing `OE_LOG_DEVICE` environmental variable.

If the var is not specified then logging goes on as usual.

If the var is specified then we look for an extra environmental variable `OE_LOG_FORMAT`.

If `OE_LOG_FORMAT` is specified we call a method where we pass in the log line information and it is formatted by the `OE_LOG_FORMAT` string. If it is not then we use the default format string to log in the file.

When logging to a separate file we can decide if we want to keep the normal logs directed to stdout AND log with the user specified format to the file. This is useful if you want to keep the human readable logs, but also redirect a machine-readable format (e.g. JSON) to a file. The other option would be to turn off stdout logging when we redirect logs to a user defined file, as is done now.

Here is an example of the logs in json format defined by setting `export OE_LOG_FORMAT="{\"e_ts\":\"%s.%06ldZ\",\"level\":\"(%s)%s\",\"tid\":\"tid(0x%lx)\",\"msg\":\"%s\",\"file\":\"%s\",\"func\":\"%s\",\"number\":\"%s\"}"`

{
  "e_ts": "2019-09-04T13:24:17.562606Z",
  "level": "(E)ERROR",
  "tid": "tid(0x7efdeaffd700)",
  "msg": "switchless_enc:log message",
  "file": "../tests/switchless/enc/enc.c",
  "func": "enc_echo",
  "number": "57"
}

(2) OE logging format

I took the liberty of playing around with the OE logging format and brought it to this:

`2019-09-04T13:52:51.868071Z [(E)ERROR] tid(0x7eff067fc700) switchless_enc:log message [../tests/switchless/enc/enc.c:enc_echo:57]`

so `timestamp [(E|H)log_level] thread_id log_message [file:function:line]`

where timestamp is ISO 8601 UTC. This is not strictly needed for this issue but it would be nice to have at least the full timestamp.

I have made some code modifications so if there are no objections to this I could open a PR so that we can discuss further in detail.

Thanks!

andschwa:
[...]

PR #2172, opened 2019-09-12
Logging format

olgavrou:
Regarding  issue #2147

This PR does 2 things

(1) slightly changes OE logging format to look like this

`timestamp [(E|H)LogLevel] tid(0xtid) | message [file:function:number]`

example

`2019-09-12T16:25:30.267014Z [(H)ERROR] tid(0x7f31230c9700) | :OE_QUOTE_PROVIDER_CALL_ERROR […/openenclave/host/sgx/ocalls.c:oe_get_qe_identify_info_ocall:288]`

(2) Accepts a user defined logging format via environmental variable and uses it when logging to a file

I have taken the liberty to remove the logging header that was written to the log file as I think it disrupts it.

I have also modified the code to continue appending the logs in the default format to standard output while at the same time writing them to the user defined file. I am making the assumption that a user would still like the logs to be outputted as normal, but also optionally redirect them to a file in a machine-readable format to be picked up by a log analytics pipeline. Let me know if you think this is a bad assumption.

@andschwa @yakman2020


host/traceh.c
```diff
@@ -194,10 +163,27 @@ oe_result_t oe_log(oe_log_level_t level, const char* fmt, ...)
...
 {
+    // get timestamp for log
+#if defined(__linux__)
+    struct timeval time_now;
+    gettimeofday(&time_now, NULL);
+    struct tm* t = gmtime(&time_now.tv_sec);
+#else
+    time_t lt = time(NULL);
+    struct tm* t = localtime(&lt);
+#endif
```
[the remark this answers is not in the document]
olgavrou: I am not sure, I left this bit of code as it was originally. If there is no reason for it be localtime when not in Linux I can change it to UTC time here too :)

host/traceh.c
```diff
@@ -224,7 +209,31 @@ void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
                 _log_creation_failed_before = true;
                 return;
             }
-            _write_message_to_stream(log_file, is_enclave, level, message);
+
+            char* message_dup = strdup(message);
+            char* log_msg = strtok(message_dup, "[");
```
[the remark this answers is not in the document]
olgavrou: Thanks John I will change this :)

host/traceh.c
```diff
@@ -224,7 +209,31 @@ void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
...
+            size_t len = strlen(log_msg);
```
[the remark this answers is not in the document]
achamayou: I think it could still be a good idea to log something like "Failed to apply custom formatter to message" or something along those lines rather than hard shut down?

host/traceh.c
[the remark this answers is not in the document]
andschwa: @anakrish Any idea why we were doing this originally?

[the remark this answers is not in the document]
anakrish: @soccerGB might have more insights. But I agree, it is better to use consistent time zones on all platforms.

host/traceh.c
```diff
@@ -208,11 +194,10 @@ void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
     // Take the log file lock.
     if (oe_mutex_lock(&_log_lock) == OE_OK)
     {
-        if (!_log_file_name)
-        {
-            _write_message_to_stream(stdout, is_enclave, level, message);
-        }
-        else if (!_log_creation_failed_before)
+        _write_message_to_stream(
```
anakrish: What is the rationale for always printing the log to stdout?

host/traceh.c
```diff
@@ -224,7 +209,31 @@ void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
...
+            size_t len = strlen(log_msg);
+            if (log_msg[len - 1] == '\n')
+            {
+                log_msg[len - 1] = 0;
+            }
+            char* file_name = strtok(NULL, ":");
+            char* function = strtok(NULL, ":");
+            char* line_number = strtok(NULL, "]");
+
+            _write_custom_format_message_to_stream(
```
anakrish: Can we avoid parsing the message and instead create the message using the correct format in the first place?
As written currently, the log that gets printed in stdout has a different format from the log that gets printed in a file.

achamayou:
@anakrish there are two problems with users writing log parsers. Firstly the log format is not formally specified nor a public API, it could change any time. Secondly, every user has to replicate this log parsing logic, which could be non trivial if the log format isn't set carefully.

As Azure Blockchain Service (and other Azure services) start to deploy applications built on OE, they want to ingest logs and metrics for automatic monitoring and alerting. I'm sure you can appreciate that they aren't keen on writing log parsers for every library used by their service.

Since the vast majority of what they use is apparently JSON, ingested through fluentd, which are both very widespread, I think it makes sense to put in a little effort to support that flow, both for them and for users at large.

host/traceh.c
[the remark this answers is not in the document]
olgavrou: Done :)

host/traceh.c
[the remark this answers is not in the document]
olgavrou: Checking for NULL on each value returned by `strtok_r` and logging a message if that is the case

host/traceh.c
olgavrou: Yes thank you, I have changed it to log the message into the file in the same format as the default formatter if the custom one is missing.

host/traceh.c
olgavrou: I am making the assumption that a user would still like the logs to be outputted as normal, but also optionally redirect them to a file in a machine-readable format to be picked up by a log analytics pipeline. We can discuss this assumption further if people are uncomfortable with it.

host/traceh.c
```diff
@@ -67,88 +69,75 @@ void initialize_log_config()
         // inititalize if not already
         _log_level = _env2log_level();
         _log_file_name = getenv("OE_LOG_DEVICE");
+        _custom_log_format = getenv("OE_LOG_FORMAT");
```
olgavrou: Instead of printing a newline after each custom log message, I am enforcing that the custom log format ends with a newline. Alternatively I could create a new longer string and append a newline at the end of it, but this seems cleaner. 

Please let me know if there is a better way to exit from here.

host/traceh.c
```diff
@@ -70,10 +70,31 @@ void initialize_log_config()
...
+                stderr,
+                "%s\n",
+                "Custom log format does not end with a newline");
+            exit(1);
+        }
         _initialized = true;
     }
 }
 
+static void _replace_newline(char* log_msg)
```
achamayou: As discussed, this isn't enough to make it possible to write logs out as JSON event. Escaping messages is necessary.

olgavrou: I would like to raise a discussion on what I am doing here. 

A lot of log messages seem to contain newlines. They possibly could contain single or double quotes or other characters that would need to be escaped. Apart from the fact that a custom log format might want to have one line per log message, which will break if we allow newline characters, if we want to support e.g. JSON for machine-readable purposes then other characters might cause further issues.

It could be that the user could indicate that they want to escape any characters on a log message and we could use some string library to do this, or do so manually.

For now I am just replacing newline characters with spaces, but please let me know if you would like me to invest some time in proposing a solution, or if you have any string handling mechanism in the codebase already.

anakrish:
> @anakrish there are two problems with users writing log parsers. Firstly the log format is not formally specified nor a public API, it could change any time. 

On the other hand, what is the specification for OE_LOG_FORMAT? It is not clear what is supported or unsupported.

> As Azure Blockchain Service (and other Azure services) start to deploy applications built on OE, they want to ingest logs and metrics for automatic monitoring and alerting.

If this PR is to support that use case, which is valid, it would be preferable to propose a design document first. It is not clear to me that production enclaves would opt for plain-text logs rather than encrypted logs.

achamayou:
@anakrish the point of OE_LOG_FORMAT is that it lets the user specify & control their log format. I agree that documenting what is supported in OE_LOG_FORMAT is necessary.

This is not a standalone PR, a fairly detailed issue (linked at the top #2147) was filed, discussed and triaged first, but it can definitely be re-formatted using the Design Document template.

host/traceh.c
olgavrou: Replaced with a method called `_escape_characters` which does some basic control character escaping. I think this is sufficient for now but we can look into something more complete if we think it is necessary. I have made it so that it is called when the env variable `OE_JSON_ESCAPE` is set so that the user can choose if they want this functionality or not, depending on their log format

olgavrou:
I have added a design doc that covers the changes in this PR

host/traceh.c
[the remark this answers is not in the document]
soccerGB: Mixing of UTC time on Linux and local time was an oversight when I implemented logging. Feel free to make them consistent.

host/traceh.c
soccerGB: That takes away the option of logging to file only.
I recommend not to spew anything out when it's set to file

host/traceh.c
```diff
@@ -67,88 +71,152 @@ void initialize_log_config()
...
+                    break;
+                case '\"':
+                    log_msg_escaped[idx] = '\"';
+                    break;
+                case '\?':
+                    log_msg_escaped[idx] = '?';
+                    break;
+                default:
+                    sprintf(&log_msg_escaped[idx], "u%04x", log_msg[i]);
+                    idx += 3;
```
achamayou: 3?

host/traceh.c
```diff
@@ -67,88 +71,152 @@ void initialize_log_config()
...
+                    break;
+            }
+        }
+        else
+        {
+            log_msg_escaped[idx] = log_msg[i];
+        }
+        idx++;
+    }
+    if (idx < 2 * msg_size)
```
achamayou: That's a hardcoded asumption about the length of log_msg_escaped, which may or may not hold?

host/traceh.c
```diff
@@ -224,7 +315,70 @@ void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
...
+                        usecs,
+                        level,
+                        "Failed to apply custom formatter to message\n");
+                }
+                else
+                {
+                    if (_json_escape)
+                    {
+                        size_t msg_size = strlen(log_msg);
+                        char log_msg_escaped[2 * msg_size];
```
achamayou: Since unprintable values that aren't special cased effectively expand to 6 bytes, I think this may be too optimistic.

docs/DesignDocs/LoggingFormat.md
```diff
@@ -0,0 +1,79 @@
...
+If the var is not specified then logging goes on as usual.
+
+If the var is specified then we look for an extra environmental variable `OE_LOG_FORMAT`.
+If `OE_LOG_FORMAT` is specified we call a method where we pass in the log line information and it is formatted by the `OE_LOG_FORMAT` string. If it is set it needs to end with a newlone.
+
+If it is not then we use the default format string to log in the file.
+
+In addition to the `OE_LOG_FORMAT` environmental variable, if the `OE_JSON_ESCAPE` environmental variable is set, then the log message will be processed accordingly in order to be compatible with the JSON standard.
+
+When logging to a separate file the normal logs are directed to stdout AND the log with the user specified format is directed to the file. This is useful if you want to keep the human readable logs, but also redirect a machine-readable format (e.g. JSON) to a file.
```
soccerGB: There should be an option to have output go to file only.

host/traceh.c
```diff
@@ -67,88 +71,152 @@ void initialize_log_config()
         // inititalize if not already
         _log_level = _env2log_level();
         _log_file_name = getenv("OE_LOG_DEVICE");
+        _custom_log_format = getenv("OE_LOG_FORMAT");
+        _json_escape = getenv("OE_JSON_ESCAPE");
+        if (_custom_log_format)
```
soccerGB: I am wondering what it's called "OE_JSON_ESCAPE" instead of OE_LOG_ESCAPE.
I think escaping could apply to type pf message contents.
Why is JSON tied to it?

host/traceh.c
soccerGB: "enforcing that the custom log format ends with a newline" sounds good to me.

host/traceh.c
olgavrou: Changed to `OE_LOG_ESCAPE`. Originally tied to JSON as I was looking at the JSON standard to implement string escaping

host/traceh.c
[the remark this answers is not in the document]
olgavrou: using gmtime for all platforms

docs/DesignDocs/LoggingFormat.md
olgavrou: Changed it to default to file only when file is specified, but optionally we can log to both stdout and file.

host/traceh.c
olgavrou: Now by default when a file is specified it will log only to the file. Added another env. variable that if set enables the option to log to stdout and to the log file.

olgavrou:
@anakrish @soccerGB I have addressed all pending comments and added a design document for further reference. I have also added a testcase. Please let me know if there are any more concerns.

Thanks for all the feedback!

host/traceh.c
```diff
@@ -194,10 +163,27 @@ oe_result_t oe_log(oe_log_level_t level, const char* fmt, ...)
...
+    struct tm* t = localtime(&lt);
+#endif
+
+    char time[20];
+    strftime(time, sizeof(time), "%Y-%m-%dT%H:%M:%S", t);
+
+#if defined(__linux__)
+    long int usecs = time_now.tv_usec;
+#else
+    long int usecs = 0;
```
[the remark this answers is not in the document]
dthaler: Millisecond time is available via https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsystemtime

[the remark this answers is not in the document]
dthaler: Windows actually keep time in units of 100-ns intervals, readable via GetSystemTimeAsFileTime.   GetSystemTime just does a conversion to milliseconds based on that.  See https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsystemtimepreciseasfiletime?redirectedfrom=MSDN

host/traceh.c
soccerGB: Thanks for addressing my comment

docs/DesignDocs/LoggingFormat.md
soccerGB: Thanks for addressing my comment

docs/DesignDocs/LoggingFormat.md
```diff
@@ -0,0 +1,79 @@
...
+If the var is not specified then logging goes on as usual.
+
+If the var is specified then we look for an extra environment variable `OE_LOG_FORMAT`.
+If `OE_LOG_FORMAT` is specified we call a method where we pass in the log line information and it is formatted by the `OE_LOG_FORMAT` string. If it is set it needs to end with a newline.
+
+If it is not then we use the default format string to log in the file.
+
+In addition to the `OE_LOG_FORMAT` environment variable, if the `OE_LOG_ESCAPE` environment variable is set, then the log message will be processed accordingly in order to be compatible with the JSON standard.
+
+When logging to a separate file the defaut logs are not sent to stdout and the custom logs are directed to the file. If the user wants to keep the default logs directed to stdout AND the custom logs sent to the file, they can set the `OE_LOG_ALL_STREAMS` environment variable. This is useful if we want to keep the human readable logs, but also redirect a machine-readable format (e.g. JSON) to a file.
```
dthaler: nit: line length.  And insert comma after "e.g."

docs/DesignDocs/LoggingFormat.md
```diff
@@ -0,0 +1,79 @@
...
+
+User Experience
+---------------
+
+- The user can set the `OE_LOG_DEVICE` environment variable to specify the log file location
+
+- The user can set the `OE_LOG_FORMAT` environment variable to specify the log format. The user will be given the below information:
+
+        - time in string format
+        - microseconds
```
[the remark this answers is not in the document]
andschwa: It seems to me to mean microseconds of precision in the logging timestamp.

docs/DesignDocs/LoggingFormat.md
```diff
@@ -0,0 +1,79 @@
+Logging Format
+=====
+
+When using the OE SDK and in trying to make use of the logs (e.g. via uploading them to Azure Log Analytics) it is nice to be able to parse and display all of the available logs including the ones generated by OE.
```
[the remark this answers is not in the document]
andschwa: I don't' think "nits" are very useful review feedback, especially on a "request changes" review. We should not hold up PRs for "nits" when they can easily be fixed by the reviewer requesting them. Especially in a design doc.

docs/DesignDocs/LoggingFormat.md
```diff
@@ -0,0 +1,79 @@
...
+                "file": "switchless/host/host.c",
+                "func": "main",
+                "level": "(H)ERROR",
+                "msg": "Hello 'world'!",
+                "number": "79",
+                "tid": "tid(0x7fc2698a6d40)"
+        }
+        `
+
+- The user can set the `OE_LOG_ESCAPE` environment variable and if it is set then the log message will be escaped in order to be compatible with the JSON standard.
```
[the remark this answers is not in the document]
andschwa: I think the intent of this PR was, in fact, to enable JSON logging for CCF as they're already using it. Are you suggesting a complete change to the PR that we already accepted the proposal for?

andschwa:
And thank you @olgavrou for the thorough contribution!

docs/DesignDocs/LoggingFormat.md
```diff
@@ -0,0 +1,79 @@
...
+        - log level
+        - thread id
+        - log message
+        - file log initiated from
+        - function log initiated from
+        - file line number log initiated from
+
+        so for example a log format could be set like this:
+
+        `OE_LOG_FORMAT=$'{\"e_ts\":\"%s.%06ldZ\",\"level\":\"(%s)%s\",\"tid\":\"tid(0x%lx)\",\"msg\":\"%s\",\"file\":\"%s\",\"func\":\"%s\",\"number":\"%s\"}\n'`
```
anakrish: I have some questions:
- What would be another good example where the use can specify log format other than JSON? Adding a second non-JSON example would help users understand how to leverage this feature.
- Can the log format skip unnecessary fields? Say I don't want the thread-id since my application is single-threaded.
- Can the user specify fields in different order via the format?
- Not sure if it is a typo: The format above specified `e_ts`, `level`, `tid` ... but the output below has fields in different order.

docs/DesignDocs/LoggingFormat.md
```diff
@@ -0,0 +1,79 @@
...
+        }
+        `
+
+- The user can set the `OE_LOG_ESCAPE` environment variable and if it is set then the log message will be escaped in order to be compatible with the JSON standard.
+
+Specification
+-------------
+
+Users can specify their own format when logging to a user specified file via the existing `OE_LOG_DEVICE` environment variable.
+If the var is not specified then logging goes on as usual.
```
anakrish: Not clear what `as usual` means.

docs/DesignDocs/LoggingFormat.md
```diff
@@ -0,0 +1,79 @@
...
+
+If the var is specified then we look for an extra environment variable `OE_LOG_FORMAT`.
+If `OE_LOG_FORMAT` is specified we call a method where we pass in the log line information and it is formatted by the `OE_LOG_FORMAT` string. If it is set it needs to end with a newline.
```
anakrish: What happens if the OE_LOG_FORMAT does not end with a newline?

host/traceh.c
```diff
@@ -67,88 +75,159 @@ void initialize_log_config()
...
-        OE_REPO_BRANCH_NAME);
-    fprintf(stream, "Last commit:%s\n\n", OE_REPO_LAST_COMMIT);
+    size_t idx = 0;
+    for (size_t i = 0; i < msg_size; i++)
+    {
+        int c = log_msg[i];
+        if (log_msg[i] == '\"')
+        {
+            // single quotes are OK for JSON
+            log_msg_escaped[idx] = (char)putchar('\'');
```
anakrish: What does `putchar` do? Why not assign the the value directly?

host/traceh.c
```diff
@@ -67,88 +75,159 @@ void initialize_log_config()
...
+        else if (!isprint(c))
+        {
+            log_msg_escaped[idx] = '\\';
+            idx++;
+            log_msg_escaped[idx] = '\\';
+            idx++;
+            switch (log_msg[i])
+            {
+                case '\a':
+                    log_msg_escaped[idx] = 'a';
```
anakrish: \a is not listed in the JSON grammar here: https://json.org/
Can you double-check the list of characters to be escaped?

host/traceh.c
```diff
@@ -67,88 +75,159 @@ void initialize_log_config()
...
+        {
+            log_msg_escaped[idx] = '\\';
+            idx++;
+            log_msg_escaped[idx] = '\\';
+        }
+        else if (!isprint(c))
+        {
+            log_msg_escaped[idx] = '\\';
+            idx++;
+            log_msg_escaped[idx] = '\\';
```
anakrish: This seems like an extra `\`

host/traceh.c
anakrish: Why are single quotes OK for JSON?

docs/DesignDocs/LoggingFormat.md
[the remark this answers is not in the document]
dthaler: nits are never a reason to -1 a PR, but I believe nits are always useful feedback.   Is there a reason they're not?  I often give nits and then approve PRs.

docs/DesignDocs/LoggingFormat.md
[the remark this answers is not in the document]
dthaler: I thought the process we were trying to follow is to review design docs before PRs.   I'm just trying to follow the documented process.  I gather you're explaining that this work didn't actually follow the documented process?   Should we have a discussion about the "Design Discussion" section of Contributing.md?

docs/DesignDocs/LoggingFormat.md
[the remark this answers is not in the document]
andschwa: "Request changes" is a -1...

docs/DesignDocs/LoggingFormat.md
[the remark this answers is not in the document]
andschwa: Please see #2147 for the approved design review.

docs/DesignDocs/LoggingFormat.md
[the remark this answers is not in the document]
dthaler: Yes, but Request Changes was not due to any nits.  Those were just freebies :)

docs/DesignDocs/LoggingFormat.md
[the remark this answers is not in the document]
dthaler: (FYI, this comment is not a nit, it's a point of reader confusion where I still can't understand what was meant.  Hence the "Request Changes")

docs/DesignDocs/LoggingFormat.md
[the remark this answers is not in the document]
dthaler: Thanks for the pointer.   My comment here is not a blocker for this PR, it's just thoughts about future design extensions.   (Request Changes is not due to this comment.)

docs/DesignDocs/LoggingFormat.md
```diff
@@ -0,0 +1,79 @@
...
+
+If it is not then we use the default format string to log in the file.
+
+In addition to the `OE_LOG_FORMAT` environment variable, if the `OE_LOG_ESCAPE` environment variable is set, then the log message will be processed accordingly in order to be compatible with the JSON standard.
```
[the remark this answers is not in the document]
dthaler: The request here is specifically to either:
a) rename OE_LOG_ESCAPE to something with JSON in the name, since the type of escaping varies by log format.   (syslog escaping and JSON escaping are different.), or
b) leave the envar as is, but update the text to replace "the JSON standard" with "the standard for the indicated log format".

My preference is (b).

docs/DesignDocs/LoggingFormat.md
dthaler: Agree that a second non-JSON example would be helpful, but could be done in a future PR in my opinion.   If you want another standard format to use in examples, use syslog format.  There's examples in RFC 5424 and in various sources like the ones linked to as references from the wikipedia entry for syslog.

docs/DesignDocs/LoggingFormat.md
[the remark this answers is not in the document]
achamayou: @dthaler the full context for this is the CCF integration in Azure Blockchain Services. A lot of Azure services infrastructure (including ABS) uses fluentd, and expects processes to log in JSON format. Regardless of whether this is the best technical choice today, it is what we are targeting.

The statement above does not suggest JSON is a logging standard, just that this env var exists to allow emitting output that _complies_ with the JSON standard. This isn't exclusive with adding CBOR support in the future, which I'm sure will happen if/when there's a use case for it.

At the moment, there's a use-case for JSON, so we're contributing that :)

docs/DesignDocs/LoggingFormat.md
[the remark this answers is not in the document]
achamayou: This was previously called OE_JSON_ESCAPE and was changed on @soccerGB's request. I propose OE_LOG_JSON_ESCAPE. I don't think b) can be meaningfully implemented without another switch telling OE what log format the log format string is meant to be representing. There can be an OE_LOG_SYSLOG_ESCAPE in the future.

host/traceh.c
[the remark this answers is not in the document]
olgavrou: Thank you for the additional details, I think this can be done separately from this PR

host/traceh.c
olgavrou: The log messsage will be given to fprintf in the end, so if we want to print backslashes we have to escape them.

docs/DesignDocs/LoggingFormat.md
olgavrou: An error message is printed and the we exit: https://github.com/openenclave/openenclave/pull/2172/files#diff-c2dce26e07f538354d7842496dba322dR91

I will rephrase the documentation here.

host/traceh.c
olgavrou: Thanks for spotting that, putchar was mistakenly left there. Single quotes do not need to be escaped by JSON based on the RFC: https://tools.ietf.org/html/rfc8259#page-8

host/traceh.c
olgavrou: Thanks, I believe I escape only what is needed now.

docs/DesignDocs/LoggingFormat.md
olgavrou: Meaning same behaviour as before this PR. Removed in order to avoid confusion.

docs/DesignDocs/LoggingFormat.md
olgavrou: @anakrish 

- We don't have another example right now we were only interested in the JSON use case at this point
- This PR does not offer the ability to skip unnecessary fields
- This PR does not offer the ability to define a different order
-  This was a result of pretty printing JSON. Re-ordered to avoid confusion

docs/DesignDocs/LoggingFormat.md
olgavrou: Fixed

docs/DesignDocs/LoggingFormat.md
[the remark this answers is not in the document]
olgavrou: Renamed to `OE_LOG_JSON_ESCAPE`

docs/DesignDocs/LoggingFormat.md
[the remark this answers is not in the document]
olgavrou: Added more information on the fields. Hope it is clearer now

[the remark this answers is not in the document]
dthaler: Thanks!  Just a suggestion now that I understand it (I think), maybe "microseconds portion of the current time" would be slightly clearer.   If your clock has 10 microseconds of precision and the current time is 50 microseconds after the indicated second, then the value is 50, not 10, as I understand it.

dthaler:
Thanks for addressing my comments

tests/logging/main.c
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#define PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP \
```
anakrish: Is this used anywhere?

host/traceh.c
```diff
@@ -67,88 +75,138 @@ void initialize_log_config()
...
+                    break;
+                case '\t':
+                    log_msg_escaped[idx] = 't';
+                    break;
+                default:
+                    if ((uint8_t)log_msg[i] > 127 /* max ASCII value */)
+                    {
+                        return false;
+                    }
+                    sprintf(
```
anakrish: Is there a guarantee that this will not write beyond the output message buffer?

docs/DesignDocs/LoggingFormat.md
anakrish: @olgavrou Thanks for the answers.
@dthaler Thanks for the pointer to syslog.

andschwa:
Giving @olgavrou permissions to merge (this is the bors delegate feature):

bors d+

andschwa:
Oops I need to update bors I guess:

bors delegate+

host/traceh.c
olgavrou: I believe so. `log_msg_escaped` is allocated to be 7 times larger than `log_msg` so that we have enough space to escape without overflowing. In this case where we get for example `u0024`, we will then need to write to `log_msg_escaped` two backslashes, `u` and the encoding in hex which comes to a total of 7 characters.

tests/logging/main.c
olgavrou: host/traceh.c uses `OE_H_MUTEX_INITIALIZER` which is defined in `hostthread.h` and which in turn uses the definition of `PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP`. The latter I belive is defined in `pthread.h`. I thought it would be ok to define it here like this for the purposes of this test.

PR #2173, opened 2019-09-12
Setup CODEOWNERS

andschwa:
This will make PRs automatically ping people!

(Also testing the maintainers GitHub team here, which we might be renaming shortly.)

I'm sure I've missed stuff, and there's obviously folders we should break out more too, but it's a start!

See [GitHub's documentation](https://help.github.com/en/articles/about-code-owners).


andschwa:
As this will serve us better by being merged sooner (and that's also the only real way to test it), I'll go ahead and merge now. We can continue to iterate on it as people find they get too many (or too few, ha!) notifications.

bors r+

PR #2176, opened 2019-09-13
Proposal: Enforcing full EDL serialization

mikbras:
Based on Simon's suggestions, I rewrote the previous proposal, dropping the analysis between the SGX and TrustZone memory models (this will be covered by the shared memory model proposal).

This proposal focuses on warning when an EDL specification is not "fully serializable". This is really the core issue and it covers both security and portability concerns without having to talk about divergent shared memory models.

### DIFF
--- a/docs/DesignDocs/full_edl_serialization.md
+++ b/docs/DesignDocs/full_edl_serialization.md
@@ -1,0 +1,151 @@
+Enforcing full serialization in EDL
+===================================
+
+Motivation
+----------
+
+Currently EDL only provides partial support for parameter serialization. For
+example, the following EDL defines ECALLs whose parameters can only be
+partially serialized.
+
+```c++
+enclave
+{
+    /* Include struct iovec (foreign structure) */
+    #include <sys/uio.h>
+
+    trusted
+    {
+        public int send(
+            [in, count=iovcnt] const struct iovec* iov,
+            size_t iovcnt);
+
+        public int recv(
+            [out, count=iovcnt] struct iovec* iov,
+            size_t iovcnt);
+    };
+};
+```
+
+The generator produces code that serializes the array of **iovec**
+structures, but it cannot serialize the buffer referenced by the structure
+itself. Recall the definition of the **iovec** structure.
+
+```c++
+struct iovec
+{
+    void* iov_base;
+    size_t iov_len;
+};
+```
+
+Consequently, these ECALLs pass pointers to untrusted memory (**iov_base**) to
+the enclave developer's **send** and **recv** implementations.
+
+The exposure of untrusted pointers to the enclave implementation poses two main
+problems.
+
+- Implementations are vulnerable to coding errors that may introduce security
+  vulnerabilities.
+- The enclave application is not portable to trusted hardware environments
+  that do not support or have not enabled shared memory (e.g., TrustZone).
+
+Some potential security vulnerabilities include:
+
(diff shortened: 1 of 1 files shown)


docs/DesignDocs/full_edl_serialization.md
```diff
@@ -0,0 +1,151 @@
...
+
+The following EDL considers a second scenario that implicitly passes untrusted
+memory into the trusted implementation.
+
+```c++
+enclave
+{
+    untrusted
+    {
+        struct widget* get_widget();
```
dthaler: this might be a bad example, since you could potentially implement this by assuming [out] for all return pointers.  A stronger example might be a variable sized result like an array or buffer, e.g.

struct widget* get_widgets([out] int* widget_count);

docs/DesignDocs/full_edl_serialization.md
```diff
@@ -0,0 +1,151 @@
...
+- The enclave application is not portable to trusted hardware environments
+  that do not support or have not enabled shared memory (e.g., TrustZone).
+
+Some potential security vulnerabilities include:
+
+- Time-of-check, Time-of-use errors (TOC-TOU).
+- Unwittingly writing secrets onto host memory.
+- Overwrite attacks where **iov_base** points to enclave memory.
+
+EDL provides a *deep-copy* feature for fully serializing parameters as shown
```
radhikaj: @mikbras Could you clarify this for me? With deepcopy all structures should be completely serialized(Please correct me here if  I am wrong). Are there problems with the deepcopy implementation that would mean that structs would not be fully serialized? Please help me understand.

andschwa: This is the result of using `struct iovec` as-is (defined in the header, which does not have the deep-copy annotations on the struct). Contrast this to `struct local_iovec` which has the same internal structure, but with the necessary deep-copy annotations (`[size=iov_len]`).

docs/DesignDocs/full_edl_serialization.md
```diff
@@ -0,0 +1,151 @@
...
+- Non-serializable local structures (lacking the appropriate annotations)
+- Pointer return values
+
+The user may wish to treat these warnings as errors using an option similar to
+GCC's **-Werror** option.
+
+Specification
+-------------
+
+The generator should be modified to produce the warnings mentioned above.
```
andschwa: I think this specification is insufficient. What would you like the warnings to say, and what flags would you like the edger8r to have? I would suggest, following the three above:

* -Wforeign-struct
* -Wptr-in-local-struct
* -Wptr-return-value

ALong with `-Wno-` versions to disable them, `-Wall` to turn them all on (should this be default or not?) and `-Werror` as suggested.

andschwa:
@mikbras I updated the proposal with my extended specification, so I am going to dismiss my review and request that others sign-off.

andschwa:
Hey @mikbras how do you feel about the additions?

docs/DesignDocs/full_edl_serialization.md
mikbras: Hi Andy. If you want to take over the specification, that would probably be better since you have a deeper knowledge about these. The warnings you proposed make sense to me.

andschwa:
Hey all, should we merge this?

andschwa:
I'm going with yes 😁

bors merge

PR #2166, opened 2019-09-11
Rename Community Maintenance Committee

dthaler:
Given the confusion between Maintainer vs CMC member, Maintainers Group
for example might be clearer.  Mike and John indicated initial support
for this term.

Another term used by many projects is "Technical Steering Committee",
which is the term that also appears in Stephen Walli's doc.   I would
be ok with that term, just for consistency with other open source
projects, if the group prefers that term.  However, it has the same
confusion between Maintainer vs TSC member.   One way that could be
resolved, should the TSC name be used, would be to remove the term
"Maintainer" and only use "TSC member", "Committer", and "Contributor"
as roles, which is the approach Stephen's doc takes.

Signed-off-by: Dave Thaler <dthaler@microsoft.com>

PR #2186, opened 2019-09-17
Rename Maintainers.md and GovernanceModel.md

andschwa:
To Committers.md and Governance.md for consistency.

This spun out of #2166.

I am assigning @openenclave/committee but we don't need a full consensus for this, I'm just renaming some files as previously discussed.


andschwa:
I think it was the change to CODEOWNERS that pulled in the whole committers group as a code owner...which I'm fine with 😄

docs/CODEOWNERS
```diff
@@ -46,7 +46,7 @@ CMakeLists.txt @andschwa @BRMcLaren @EmilAlexandruStoica
 /docs/DesignDocs/ @openenclave/committers
 
 # And governance documents.
-/docs/Contributing.md @openenclave/maintainers
-/docs/GovernanceModel.md @openenclave/maintainers
-/docs/Maintainers.md @openenclave/maintainers
-/docs/Releasing.md @openenclave/maintainers
```
dthaler: can we create a /process subdirectory under /docs to group these and just put openenclave/committee on the whole subdirectory?  We may have other docs over time that get added so want to reduce opportunities for errors.

andschwa: I don't think that'll work for Contributing.md as GitHub specifically references it from docs.

andschwa: (And then I think it would be weird to have them in different places.)

andschwa: In my opinion...not necessary 😉

andschwa:
This "approved on behalf of" thing with GitHub is a little...weird.

Issue #2188, opened 2019-09-18
TestRdrand does not effectively test oe_rdrand handles saturation of the RDRAND interface.


CodeMonkeyLeet:
This test does not actually manage to exhaust the RDRAND entropy pool regardless of the number of iterations run since the operation of retrieving the RDRAND value through the bus architecture is slow enough that a single thread can't saturate the interface regardless of the number of iterations run. A slight modification of the code and use of debugger breakpoints reveals that the calls during the test execution never has to retry the RDRAND assembly call.

According to the [Intel DRNG implementation guide](https://software.intel.com/en-us/articles/intel-digital-random-number-generator-drng-software-implementation-guide) it looks like it will take about 10-12 concurrent threads to saturate the RDRAND interface, so the test needs to be rewritten to take that into account. 

In the meantime, this test also currently accounts for the bulk of the time it takes to run the tests/crypto suite (~12s) so it should disabled, or the number of iterations scaled way down since they don't achieve the intended goal anyway.

CodeMonkeyLeet:
Per triage, we're not going to modify the test to actually test RDRAND saturation and the workaround to reduce the overhead of this test by reducing the iterations run in PR #2189 is sufficient.

Issue #1766, opened 2019-05-07
Add support compiler runtimes in the enclave


CodeMonkeyLeet:
The enclave runtime currently avoids linking in the compiler runtimes (e.g. libgcc, ~~libclang~~ compiler-rt) and instead includes portable subsets of them (e.g. libunwind) to keep the TCB small and the portability issues manageable.

This creates compatibility issues that create developer friction where existing apps they need to avoid use of particular intrinsics (e.g. we don't support many math intrinsics) or provide missing linker dependencies themselves. For example:

- https://github.com/microsoft/openenclave/pull/531#issuecomment-415430684
- https://github.com/microsoft/openenclave/pull/603#issue-211443232
- #1729 

As OE SDK tries to support more transparent porting of apps into the enclave, this is something we should investigate if this is something that we should support to reduce the friction of dealing with missing compiler runtime functions & intrinsics. This can be relatively expensive as it would involve rationalizing libunwind against each compiler runtime we choose to support, but there may also be some things we can do in the tooling/dev environment to mitigate some of the developer pain in the meantime.

letmaik:
@CodeMonkeyLeet You mentioned libclang, which this is unrelated. You probably mean compiler-rt.

Good overview: https://clang.llvm.org/docs/Toolchain.html and https://compiler-rt.llvm.org/. Note in the latter link: "provides full support for the libgcc interfaces". Also, libunwind is not part of compiler-rt but separate to that. libgcc does have built-in unwind functionality on the other hand.

In my opinion, adding just the [`builtins` part](https://github.com/llvm-mirror/compiler-rt/tree/master/lib/builtins) of compiler-rt to OE would be a big step forward and deal with all relevant use cases that came up in the past. It's a fairly small math library and is compatible both with GCC and Clang. There would be no conflict with libunwind as well. Maybe someone from OE could just give it a try? I don't see much complexity here to be honest, but maybe I'm missing something.

letmaik:
I finally gave it a try and it's really too simple...

```cmake
enable_language(ASM)
set(COMPILER_RT_DEFAULT_TARGET_TRIPLE x86_64-pc-linux-unknown CACHE STRING "" FORCE)
add_subdirectory(external/compiler-rt EXCLUDE_FROM_ALL)
add_library(compiler-rt-builtins ALIAS clang_rt.builtins-x86_64)
```
[...]

Issue #216, opened 2018-04-26
Clarify the CPUID implication on mbedtls


lk2100:
The CPUID in enclave is fed from host, it means the host can control which feature is available for mbedtls. It might has security implication on mbedtls.

Had a quick chat with Simon. For the AESNI impact on mbedtls, we explicitly configure the mbedtls to use AESNI, the expectation is it will fail if AESNI is absent. In that case:
•	If host lies that AESNI is absent, mbedtls fails.
•	If host lies that AESNI is present, it’s not interesting because mbedtls attempting to use AESNI will hit SIGILL and fail out. 

If the mbedtls will not fail when AESNI is absent even after we explicitly configure it to use AESNI, we will do the same mitigation as Bcrypt (fail enclave initialization if AESNI is not present).

We need to clarify and test the behavior.

Issue #189, opened 2018-04-11
Test VectorException fails with optimized builds on SGX HW


anitagov:
VectorException fails with optimized builds on SGX HW (with workaround for the xstor64 fix #144)

1: Test command: /home/sgx-anita/relTesting/openenclave/build/tests/VectorException/host/VectorExceptionHost "./enc/VectorExceptionEnc.signed.so"
1: Test timeout computed to be: 9.99988e+06
1: === This program is used to test basic vector exception functionalities.
1: TestVectorException: will generate a hardware exception inside enclave!
1: TestVectorException: hardware exception is handled correctly!
1: Illegal GETSEC did not raise 2nd chance exception.
1: : /home/sgx-anita/relTesting/openenclave/tests/VectorException/host/host.c(41): error: ECALL TestSigillHandling failed args.result=-1
1/1 Test #1: tests/VectorException ............***Failed    0.28 sec

PR #2189, opened 2019-09-18
Update mbedtls_hardware_poll to use RDSEED

### LINKED ISSUE #242: Update mbedtls_hardware_poll to use RDSEED
OE currently provides entropy to mbedtls for seeding its CSPRNG using only requested length via the RDRAND instruction. This is suboptimal as RDRAND itself is implemented as CSPRNG to stretch hardware provided entropy. In order to provide the required true bits of entropy requested, we should:

- [ ] Check for the availability of RDSEED instruction using CPUID. 
    - [ ] If it is available, use RDSEED to provide the requested bits of entropy
    - [ ] If not, we should condense entropy to the required bits by hashing multiple RDRAND invocations.
        - Section 4.2.6 of https://software.intel.com/en-us/articles/intel-digital-random-number-generator-drng-software-implementation-guide has an example using AES CBC-MAC
        - Per the document above, reading 512 x 128-bit values (8192 bytes) of RDRAND forces it to reseed. Reading 3 windows of this (8192 * 3) provides enough entropy to be hashed down into a single 512-bit value (64 bytes). The hashing can also be done with SHA-512. 
        - This needs to be repeated for requests for > 64 bytes of entropy
    - [ ] We should fail up front if neither RDSEED nor RDRAND are supported, much like we do for AESNI (see #216)

CodeMonkeyLeet:
According to the [Intel DRNG implementation guide](
https://software.intel.com/en-us/articles/intel-digital-random-number-generator-drng-software-implementation-guide), Open Enclave can provide stronger prediction resistance by seeding the mbedTLS random number generator with RDSEED, or by condensing the entropy provided by RDRAND through several rounds of reseeding of its internal DRNG. This changeset applies this guidance:

- Update rand.asm/rand.S files to support RDSEED via `oe_rdseed`.
    - Also update the `oe_rdrand` instruction to PAUSE prior to each retry as
      suggested by the DRNG implementation guide.

- Change `oe_get_entropy` API to return the source of entropy provided.
    - This allows the entropy intialization function in the crypto library to
      compensate for the lower quality of RDRAND by condensing it.
    - This responsibility is delegated to the crypto library as condensing the
      entropy requires a cryptographic hashing function (e.g. AES-CBC-CMAC,
      SHA2) which is not available in `oecore` for `oe_get_entropy`.

- Update `mbedtls_hardware_poll` implementation to condense entropy from
  RDRAND as suggested by the Intel DRNG implementation guide when RDSEED is not
  available.
    - While most SGX-capable systems will have RDSEED available, this is not
      guaranteed for all Intel SKUs, or in simulation mode.

- Update tests/crypto to also test `oe_rdseed` implementation.
    - Also rename `TestRdrand` to `TestCpuEntropy` to reflect this.
    - Reduced the number of iterations for oe_rdrand as it does not manage to
      saturate the RDRAND interface single-threaded anyway. See #2188.

Fixes #242.


common/sgx/rand.S
```diff
@@ -3,15 +3,38 @@
 
 .text
 .globl  oe_rdrand
+.type oe_rdrand, @function
+.globl  oe_rdseed
+.type oe_rdseed, @function
 
 oe_rdrand:
+.cfi_startproc
```
anakrish: Since the function does not use the stack, there is no need for rbp, rsp manipulation as well as the leave instruction.

enclave/core/sgx/entropy.c
```diff
@@ -1,48 +1,82 @@
...
-    if (!output)
-        goto done;
+static oe_entropy_kind_t _get_entropy_kind()
+{
+    oe_entropy_kind_t result = OE_ENTROPY_KIND_NONE;
 
-    /* Copy 64-bit random integers to output */
-    {
-        size_t n = len / sizeof(uint64_t);
+    if (_has_cpuid_feature(1, OE_CPUID_RDRAND_FEATURE, OE_CPUID_RCX))
```
anakrish: What will happen if the host lies about rdrand/rdseed availability? Could it end up being emulated by the hypervisor?

enclave/core/sgx/entropy.c
```diff
@@ -1,48 +1,82 @@
...
+#include <openenclave/internal/raise.h>
 #include <openenclave/internal/rdrand.h>
+#include <openenclave/internal/rdseed.h>
+#include "cpuid.h"
 
-/* TODO: This should use RDSEED instead. See issue #242. */
-oe_result_t oe_get_entropy(void* output, size_t len)
+typedef uint64_t (*_entropy_function_t)(void);
+
+OE_INLINE bool _has_cpuid_feature(
```
anakrish: Does not seem like inline is needed here

common/sgx/rand.S
CodeMonkeyLeet: Removed from both versions.

enclave/core/sgx/entropy.c
CodeMonkeyLeet: Hypervisor doesn't intercept this AFAIK, but malicious host can set arbitrary CPUID values as part of the enclave initialization. The same logic as described in #216, applied to RDRAND instead of AESNI:

- If host lies that RDRAND is absent, enclave initialization stops the enclave from running.
- If host lies that RDRAND is present, it's not a security issue, the enclave simply fails later on first use of RDRAND and fail out with SIGILL.

We do not apply the same requirement to RDSEED because it's less common than RDRAND still and will block simulation mode in many more cases, so the worse lie the host can tell is that RDSEED is absent, but RDRAND is present. This is mitigated by exposing the type of entropy used the caller so that it can either fail out, or enhance the predictive resistance provided by RDRAND themselves as illustrated in this PR.

enclave/core/sgx/entropy.c
CodeMonkeyLeet: It's not strictly necessary, but it's a reasonable compiler hint since this method is broken out for readability and could be optimized by its inclusion into `oe_entropy_kind_t`. It also includes the `static` keyword which is convenient here.

3rdparty/mbedtls/mbedtls_hardware_poll.c
```diff
@@ -19,13 +93,48 @@ int mbedtls_hardware_poll(
...
+        /* If RDSEED is not supported, fallback to using RDRAND to obtain a
+         * seed for entropy. */
+        unsigned char* p = (unsigned char*)output;
+        size_t bytes_left = len;
+        while (bytes_left > 0)
+        {
+            uint8_t* seed_bytes = NULL;
+            size_t seed_size = 0;
+
+            if (_get_seed_from_rdrand(&seed_bytes, &seed_size) != OE_OK)
```
anakrish: It woulde be better to avoid the double call to `oe_get_entropy`. In line 102 and the other via __get_seed_from_rdrand

enclave/core/sgx/entropy.c
```diff
@@ -1,48 +1,86 @@
...
+        get_entropy = oe_rdrand;
+    else
+        OE_RAISE(OE_UNSUPPORTED);
+
+    while (bytes_left > 0)
+    {
+        uint64_t random = get_entropy();
+        size_t copy_size =
+            (sizeof(random) > bytes_left) ? bytes_left : sizeof(random);
+        memcpy(p, &random, copy_size);
```
anakrish: `memcpy` is a special function that is aggressively optimized by compilers.
If the size is a constant, memcpy is often emitted as mov instructions.

Therefore it is better to split this into:
```c
while (bytes_left > sizeof(uint64_t)
{
    // generate and copy random uint64_t value
   // memcpy will be optimized out
    bytes_left -= sizeof(uint64_t);
}
if (bytes_left > 0)
{
    // generate last 0-7 bytes.
    // memcpy will not be optimized out
}
```

common/sgx/rand.S
```diff
@@ -3,15 +3,28 @@
 
 .text
 .globl  oe_rdrand
+.type oe_rdrand, @function
```
anakrish: Consider replacing this external assembler code with GCC/clang inline assembly in Linux.
It will allow the compiler to perform better optimizations and would lead to code as close to performance as the GCC intrinsics themselves.

```c
OE_INLINE uint64_t oe_rdrand()
{
    uint64_t rv = 0;
    asm volatile(
	"_rdrand_retry: \n"
        "    rdrand %0 \n"
	"    jnc _rdrand_retry \n"
	: "=a"(rv)  // Outputs to rax 
	:           // No inputs
	: "cc"); // Clobbers flags
    return rv;
}
```

enclave/core/sgx/entropy.c
achamayou: That's disabled by -fno-builtin though: https://github.com/openenclave/openenclave/blob/de52efdfd7b15d2e04f280632f7b61d0e5d3553b/enclave/core/CMakeLists.txt#L228

https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html

anakrish: > That's disabled by -fno-builtin though: 

Good point. 

One thing to do to get the full benefit of the compiler optimizations would be to call `__builtin_memcpy` here. The compiler will perform the optimizations/inlining where possible, and resort to `memcpy` where it can't.

Note that OE does not have a requirement that memcpy should not be optimized; and it is not a given that we will continue to compile with -fno-builtin.

3rdparty/mbedtls/mbedtls_hardware_poll.c
CodeMonkeyLeet: This is by design. It produces the most straightforward separation of layering between a crypto library and the `oe_get_entropy` method in the absence of hash functions embedded in oecore itself to condense entropy itself.

common/sgx/rand.S
CodeMonkeyLeet: I would not optimize this specifically for `oe_rdrand()` since:

- The syntax for inline assembly has not been well-handled in the history of this project (e.g. #144, #189) In the area of security features, I would prioritize readability and maintainability relative to performance.

- There's an established pattern of splitting Linux/Windows implementations as .S/.asm pairs. If you think there's a valuable optimization that should be generally applied for how we deal with Linux assembly, (e.g. .c/.asm pairs) I think that should be a separate refactor, and we should establish that as a standard pattern of use with the committers.

enclave/core/sgx/entropy.c
CodeMonkeyLeet: I would explicitly not micro-optimize for performance here. The `oe_get_entropy` codepath should not be a hot spot in the code that requires this level of hand-tuning against compiler behaviors.

Empirically, the existing code vs. the proposed modification including both the loop restructure and use of `__builtin_memcpy` compiled with Release settings:
- At 1 mil iterations of `oe_get_entropy` averaged over 100x, both result in a runtime of 0.11s.
- The variance in both runs ranges between 0.10-0.12s which suggests that any effect of the optimization is well within the bounds of noise.

Again, here my preference is for less code than more optimized code so I will keep the simplified loop structure.

@anakrish, a couple of notes:
- The reasons for optimizations as in the original code need to be called out in the comments, otherwise it's possible they will be removed in future edits. In this case, I had actually done a previous perf comparison (albeit without `__builtin_memcpy` to circumvent `-fno-builtin`, unlike in this run) and concluded that there was no reason not to simplify the loop structure, but as the group of committers grows, we should be very conscientious in documenting intention behind special constructions of code.
- The broader issue of why we compile with `-fno-builtin` is more important to address rather than specific optimizations in an instance of code. If you are passionate about this, I'd direct your attention to #1766 instead, where Maik also suggested enabling broader support for just builtins.

3rdparty/mbedtls/mbedtls_hardware_poll.c
```diff
@@ -3,9 +3,83 @@
...
+    oe_entropy_kind_t kind = OE_ENTROPY_KIND_NONE;
+
+    /* Per Intel's DRNG software implementation guide we try to obtain an
+     * equivalent amount of entropy by condensing several reseed windows of
+     * RDRAND into a single value. Reading 3x of the RDRAND_RESEED_SIZE_BYTES
+     * window provides enough entropy to be hashed back down into a single
+     * 512-bit value (64 bytes), which is typically how much entropy is
+     * requested.
+     */
+    size_t rdrand_bytes_size = RDRAND_RESEED_SIZE_BYTES * 3;
```
gupta-ak: Shouldn't this be 4x to ensure future calls will also have enough entropy? For example, two consecutive calls to this function might only do 7 RDSEED intervals since 6 * 512 < 7 * 511 instead of the expected 8.

tests/crypto/cpu_entropy_test.c
```diff
@@ -0,0 +1,57 @@
...
+    for (uint64_t i = 0; i < MAX_LOOP_SIZE; i++)
+    {
+        rand_num = oe_rdrand();
+
+        /* 0 is a legal random value that could be returned, but the
+         * odds of this happening twice in a row are very unlikely
+         * unless we've run out of hardware entropy and are returning
+         * without retrying until we have sufficient entropy.
+         */
+        if (rand_num == 0)
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Anita added this test and I tried to reverse engineer the logic behind it, hence the added comment. My understanding is:
* The test is trying to saturate the underlying interface and force it to loop before returning.
* `rand_num == 0` does not necessarily indicate a failure of the `oe_rdrand` loop logic since success/failure is indicated by the CF 1/0, not the value of `rand_num` itself, so the test can't just assume `oe_rand` is failing to loop if it sees a random value of 0.
* If it does see two consecutive zero random values, then it assumes that the odds of that happening are so small, that it does suggest that `oe_rdrand` failed to block and retry. It's hokey, but I'm assuming it was an issue Anita encountered when this was running a 100 mil iterations at a time, and that this change made it more robust.
* I'm just keeping this as an automated sanity check that can be walked using a debugger at this point, see `TODO` regarding reducing the number of iterations it now runs.

3rdparty/mbedtls/mbedtls_hardware_poll.c
CodeMonkeyLeet: I'm not a cryptographer, so I went with the recommendation of Niels Ferguson who owns the https://github.com/microsoft/SymCrypt library. I'm not sure if it's necessary for the RDSEED intervals to be disjoint between consecutive calls as suggested.

johnkord:
Fixes #2188

Issue #2168, opened 2019-09-11
Inconsistent process for determining PR assignees


dthaler:
GovernanceModel.md explains:

> We define "maintainer" as members of the Open Enclave "Community Maintenance                                            Committee", as listed in the [maintainers document](Maintainers.md). A                                                  "committer" is anyone with direct write access to the Open Enclave repository on                                        GitHub, as granted by the Committee. All maintainers are committers, but not all                                        committers are maintainers.

Maintainers.md says regarding maintainers:

> maintainers are committers that are trusted to grant new committer rights

and regarding committers:

> The area column describes which technical areas each committer is most interested in, and                              therefore should usually be consulted for changes relating to that area.  However, it is up to each committer to determine who should review which PR, and when to merge it. Remember that a PR must not be merged if a committer objects; instead, it should be brought up with the Community Maintenance Committee.

Contributing.md on the other hand says:

> **DO** submit all code changes via pull requests (PRs) rather than through a                                            direct commit. PRs will be reviewed and potentially merged by the repo                                                  **maintainers** after a peer review that includes at least one **maintainer**.

Note the requirement about maintainers being here but not mentioned in the maintainer responsibilities on other pages.

> - **DO** tag any users that should know about and/or review the change.

The above does not say how to determine who that should be.  (The Maintainers.md page implies it should be all committers with an area related to the PR, but this isn't clear.)

and

> When a PR is ready to be merged, a maintainer will comment on it with `bors r+`.

So the above says that only maintainers (not committers in general) can merge PRs, which seems to contradict the other pages.

It would also streamline the process if we had git groups per area so one didn't need to look up the current list of relevant committers, one could just pick "security-committers" and/or "build-committers" or whatever other relevant groups exist.

dthaler:
It's also unclear whose responsibility it is to follow the "DO tag any users that should know about and/or review the change."   Is every contributor expected to know this (seems hard for new contributors)?  Is every committer responsible for reviewing the set of assignees?

andschwa:
Yeah that's exactly what happened.

andschwa:
I think we've actually resolved this between #2173 and #2166. The wording is now correct, and auto review tagging is happening.

dthaler:
Contributing.md still says 'DO tag any users that should know about and/or review the change."
If auto-review tagging is sufficient, then this DO can go away.
If it is not sufficient, then this DO is still insufficient in terms of being actionable.

dthaler:
[...]

PR #2190, opened 2019-09-18
Fix some language in contributing and governance

andschwa:
Just caught this while looking through the docs for #2168.

### DIFF
--- a/docs/Contributing.md
+++ b/docs/Contributing.md
@@ -82,15 +82,17 @@ Please do:
 - **DO** submit all code changes via pull requests (PRs) rather than through a
   direct commit. PRs will be reviewed and potentially merged by the repo
   Committers after a peer review that includes at least one Committer.
-- **DO** give PRs short-but-descriptive names (e.g. "Improve code coverage for
-  System.Console by 10%", not "Fix #1234").
+- **DO** give PRs short but descriptive names (e.g. "Improve code coverage for
+  edger8r", not "Fix #1234").
 - **DO** add breaking changes, new features, deprecations, and bug
-  fixes to the [unreleased section of the
-  changelog](../CHANGELOG.md#unreleased).
+  fixes to the [unreleased section of the changelog](../CHANGELOG.md#unreleased).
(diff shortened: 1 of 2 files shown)


andschwa:
Thanks @dthaler!

docs/Contributing.md
```diff
@@ -82,15 +82,17 @@ Please do:
 - **DO** submit all code changes via pull requests (PRs) rather than through a
   direct commit. PRs will be reviewed and potentially merged by the repo
   Committers after a peer review that includes at least one Committer.
-- **DO** give PRs short-but-descriptive names (e.g. "Improve code coverage for
-  System.Console by 10%", not "Fix #1234").
+- **DO** give PRs short, but descriptive names (e.g. "Improve code coverage for
```
CodeMonkeyLeet: I don't think you need the comma between the adjectives, it's only used before "but" when introducing an independent clause, which "but descriptive names" is not.

docs/Contributing.md
```diff
@@ -82,15 +82,17 @@ Please do:
...
+  edger8r", not "Fix #1234").
 - **DO** add breaking changes, new features, deprecations, and bug
-  fixes to the [unreleased section of the
-  changelog](../CHANGELOG.md#unreleased).
+  fixes to the [unreleased section of the changelog](../CHANGELOG.md#unreleased).
 - **DO** refer to any relevant issues and include [keywords](
   https://help.github.com/articles/closing-issues-via-commit-messages/) that
   automatically close issues when the PR is merged.
-- **DO** tag any users that should know about and/or review the change.
+- **DO** tag any users that should know about and/or review the change. While
```
CodeMonkeyLeet: Tangent: The use of CODEOWNERS.md actually pulls in a lot of reviewers today for some files and adding even more reviewers can cause them to be lost in the noise. We might want to consider guidance for targeting reviewers using the Assignee field as well.

andschwa: Yeah we need to watch this and identify which sets of files should be setup to who. The initial implementation was kind of just a best guess 😂

PR #2193, opened 2019-09-20
Prereqs folder deprecation

### LINKED ISSUE #2036: Determine the fate of the prereqs directory
There's a prereqs directory that hasn't been updated in some time, and it is not currently tested in our CI process. Here are my current thoughts on this:

Should we keep it around?
  - If not, then we should delete it and any references to it in our documentation.
    - Right now our ansible tasks only support SGX DCAP installation scenarios; we'll want to consider if the SGX (non-DCAP) scenario is something we should support with ansible too if we delete this directory. I imagine that wouldn't be a large amount of work.
  - If so, we should fix it up and add some kind of tests for it
  - Windows doesn't use this directory for anything, I believe, so we can ignore that platform. I believe there is a script for Windows that configures your system for either SGX DCAP or non-DCAP workflows depending on the option you specify.

johnkord:
For some reason, @vtikoo is unable to submit a PR from his fork to this repo. We need to fix that! I am submitting on his behalf.

Fix #2036


vtikoo:
### Summary
This PR deprecates the prereqs folder and all the Makefiles in it. The only usage for the Makefiles was in setup instructions for SGX1. An ansible script for SGX1 installation has been added at `scripts/ansible/oe-contributors-setup-sgx1.yml ` as the recommended setup.

scripts/ansible/roles/linux/intel/tasks/sgx-driver.yml
```diff
@@ -29,11 +29,21 @@
 
 - name: Download Intel SGX DCAP Driver
   get_url:
-    url: "{{ intel_sgx_driver_url }}"
+    url: "{{intel_sgx_w_flc_driver_url}}"
     dest: /tmp/sgx_linux_x64_driver_dcap.bin
     mode: 0755
     timeout: 120
   retries: 3
+  when: flc_enabled|bool
```
BRMcLaren: I don't believe there is a need to use the Jinja filter `|bool` here. 

I would expect the system to know that this is a boolean value, more can be read https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html.

BRMcLaren: Are we sure there is a need to use the Jinja filter, `|bool`?

I would expect the system to know that this is a boolean value, more can be read on Jinja filters here: https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html.

(occurs 3x times)

scripts/ansible/roles/linux/intel/tasks/sgx-driver.yml
```diff
@@ -29,11 +29,21 @@
...
+
+- name: Download Intel SGX1 Driver
+  get_url:
+    url: "{{intel_sgx1_driver_url}}"
+    dest: /tmp/sgx_linux_x64_driver_dcap.bin
```
BRMcLaren: Do we want to be storing both the sgx1 and dcap driver under the same destination?

If this is the case, a recommendation would be to abstract this out and to use a variable.

If this is not the case, a recommendation could be to refactor.

johnkord:
Actually, I have two suggestions for this PR, but they're unrelated to the files changed! :)

1. Can you clean up the commit messages a bit and make them more descriptive?
2. Can you read through the CLA (Contributor License Agreement) from the CLA Bot comment above and then go through the process to sign it (assuming you have no issues with the agreement)?

Thanks Vikas!

scripts/ansible/roles/linux/intel/tasks/sgx-driver.yml
vtikoo: I had initially tried without the `| bool` bit, but it was giving some warning message.

scripts/ansible/roles/linux/intel/tasks/sgx-driver.yml
vtikoo: @johnkord agree about having a generic name. Pushing a commit to fix this.

andschwa:
@johnkord Can't submit a PR from his fork?! That's a weird one...it looks like it wants to let me do it:

image

johnkord:
Updated commit messages look good, thanks so much for this @vtikoo ! I'm putting in my approval to merge this PR; I just can't officially sign off because I submitted the PR :)

johnkord:
Let's see if this README.md placeholder file fix solves this problem :)

bors r+

PR #2194, opened 2019-09-21
Remove vsbuildtools.exe hash check from install-windows-prereqs.ps1

johnkord:
This commit changes install-windows-prereqs.ps1 to now skip a hash check if an empty hash is specified. Unfortunately, the vsbuildtools.exe version can change without the URL changing, so we can't expect its hash to stay consistent, so we've disabled hash checking for vsbuildtools.exe.

Also the OpenSSLHash was missing from the PACKAGES variable, so I set it.

### DIFF
--- a/scripts/install-windows-prereqs.ps1
+++ b/scripts/install-windows-prereqs.ps1
@@ -29,8 +29,8 @@ Param(
     [string]$VCRuntime2012Hash = '681BE3E5BA9FD3DA02C09D7E565ADFA078640ED66A0D58583EFAD2C1E3CC4064',
     [string]$AzureDCAPNupkgURL = 'https://www.nuget.org/api/v2/package/Azure.DCAP.Windows/0.0.2',
     [string]$AzureDCAPNupkgHash = 'E319A6C2D136FE5EDB8799305F6151B71F4CE4E67D96CA74538D0AD5D2D793F1',
-    [string]$Python3URL = 'https://www.python.org/ftp/python/3.7.4/python-3.7.4.exe',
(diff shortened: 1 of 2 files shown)


scripts/install-windows-prereqs.ps1
```diff
@@ -127,7 +128,6 @@ function Write-Log {
     $msg = $Message | Timestamp
     Write-Output $msg
 }
-
 function New-Directory {
```
BRMcLaren: Could we revert the empty line change?

johnkord:
Thanks @BRMcLaren , those newline removals were accidental!

scripts/install-windows-prereqs.ps1
```diff
@@ -10,7 +10,7 @@ Param(
     [string]$SevenZipURL = 'https://www.7-zip.org/a/7z1806-x64.msi',
     [string]$SevenZipHash = 'F00E1588ED54DDF633D8652EB89D0A8F95BD80CCCFC3EED362D81927BEC05AA5',
     [string]$VSBuildToolsURL = 'https://aka.ms/vs/15/release/vs_buildtools.exe',
-    [string]$VSBuildToolsHash = '7D5B0220670BA1C174F0AE1FF2CE87D65A508E66C321431FBD4751F478037E12',
```
andschwa: Maybe a note in the script itself as to why we don't check it's hash? Or better, could we cache the tool? It's weird to be dependent on tools from elsewhere on the internet, especially if it means we cannot verify their integrity. I'm just thinking back to the day of downtime due to an externally hosted driver...

johnkord:
Rebased on master and addressed Andy's comment. Thanks all!

bors r+

PR #2195, opened 2019-09-22
Add a sample and a multi-threaded test for switchless

jxyang:
* Add a sample to illustrade switchless calls
* Explain the concept of switchless calls in the sameple readme
* Remove the experimental feature guard for switchless
* Add a multi-threaded switchless test case
* Enable both single-threaded and multi-threaded switchless tests on Windows


tests/switchless_threads/enc/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+  OUTPUT switchless_threads_t.h switchless_threads_t.c switchless_threads_args.h
+  DEPENDS
+  edger8r
+  ../switchless_threads.edl
+  COMMAND edger8r --experimental --trusted --search-path ${CMAKE_CURRENT_SOURCE_DIR}/.. switchless_threads.edl)
+
+# Dummy target used for generating from EDL on demand.
+add_custom_target(switchless_threads_enc_gen DEPENDS switchless_threads_t.h switchless_threads_t.c switchless_threads_args.h)
+
+add_enclave(TARGET switchless_threads_enc UUID 6e818629-0ce7-46cd-822a-6c7e081fc68b SOURCES enc.c switchless_threads_t.c)
```
BRMcLaren: `6e818629-0ce7-46cd-822a-6c7e081fc68b`

I am lost on where this value comes from and if it holds any significance.

```brmclare@BrettSGXDev:~/openenclave$ grep -rins 6e818629-0ce7-46cd-822a-6c7e081fc68b
tests/switchless_threads/enc/CMakeLists.txt:14:add_enclave(TARGET switchless_threads_enc UUID 6e818629-0ce7-46cd-822a-6c7e081fc68b SOURCES enc.c switchless_threads_t.c)```

It _seems_ to be completely random, looking at https://github.com/openenclave/openenclave/blob/master/cmake/add_enclave.cmake#L24 makes me wonder if the TA binaries will ever need a different UUID?

File: samples/switchless/README.md [hunk 2]
```diff
@@ -0,0 +1,133 @@
...
+- Demonstrate how to configure an enclave to enable switchless calls within it
+
+Prerequisite: you may want to read [Common Sample Information](../README.md#common-sample-information) before going further.
+
+## Switchless Calls
+
+In an enclave application, the host makes **ECALL**s into functions exposed by the enclaves it created. Likewise, the enclaves make **OCALL**s into functions exposed by the host that created them. In either case, the execution has to be transitioned from an untrusted environment to a trusted environment, or vice versa. Since the transition is costly due to heavy security checks, it might be more performance advantageous to make the calls **context-switchless**: the caller delegates the function call to a worker thread in the other environment, which does the real job of calling the function and post the result to the caller. Both the calling thread and the worker thread never leave their respective execution contexts during the perceived function call.
```
dthaler: "the enclaves make" is overstated.   It's perfectly fine to have an enclave application that makes no OCALLs (the same cannot be said about ECALLs).

dthaler: "might be more performance advantageous" is not very helpful.  How should the developer know whether to do this?  What sort of performance test should they do?   In this sample, how did you determine that the specific calls being done were more advantageous to do switchlessly?   Need more guidance for a developer to know whether/when to use this sample vs not use this sample.

dthaler: Split lines in the markdown into shorter lines to make it easier to review PR comments in context

samples/switchless/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+The calling thread and the worker thread need to exchange information twice during the call. When the switchless call is initiated, the caller needs to pass the `job` (representing the function call) to the worker thread. And when the call finishes, the worker thread needs to pass the result back to the caller. Both exchanges need to be synchronized.
```
dthaler: Need to better define the term "job".   What does "representing the function call" mean?

samples/switchless/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+## How does OE support switchless OCALLs
```
dthaler: I think "OE" should always be expanded in docs.

dthaler: Why is there a section called "How does OE support switchless OCALLs" but no equivalent title for ECALLs?

samples/switchless/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+A host worker thread checks and retrieves `job` from the shared memory region. It uses the untrusted OE runtime to process the `job` by unmarshaling the parameters, then dispatching to the callee function, and finally relaying the result back to the trusted OE runtime, which is further forwarded back to the caller.
+
+To support simultaneous switchless OCALLs made from enclaves, the host workers are multi-threaded. OE allows users to configure how many host worker threads are to be created for servicing switchless OCALLs. The following example illustrates how to do that. A word of caution is that too many host worker threads might increase competition of cores between threads and degrade the performance. Therefore, if a enclave has switchless calls enabled, OE caps the number of host worker threads for it to the number of enclave threads specified.
```
dthaler: Re "too many host worker threads might"... So, how is the developer supposed to come up with the number to use?   How did you come up with the number used in this sample?

samples/switchless/README.md
```diff
@@ -0,0 +1,133 @@
...
+```edl
+enclave {
+    trusted {
+        public void enclave_helloworld();
+
+    };
+
+    untrusted {
+        void host_helloworld();
+        void host_helloworld_switchless() transition_using_threads;
```
dthaler: This looks like the same API is exposed both ways.  Is that what the developer should learn from this: expose every API twice?   This looks confusing as presented.

samples/switchless/README.md
```diff
@@ -0,0 +1,133 @@
...
+    };
+};
+```
+
+Function `host_helloworld_switchless`'s declaration ends with keyword `transition_using_threads`, indicating it should be called switchlessly at run time. However, this a best-effort directive. OE runtime may still choose to fall back to a tradition OCALL if switchless call resources are unavailable, e.g., the enclave is not configured as switchless-capable, or the host worker threads are busy servicing other switchless OCALLs.
```
dthaler: If it's a best effort directive, then what's the point of keeping the non-switchless one in the EDL file?  I don't understand this sample.

samples/switchless/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+To generate the functions with the marshaling code, the `oeedger8r` tool is called in both the host and enclave directories from their Makefiles. For example:
+
+```bash
+cd host
+oeedger8r ../switchless.edl --untrusted --experimental
+```
+
+`oeedger8r` needs the command line flag `--experimental` to be able to recognize the keyword `transition_using_threads`.
```
dthaler: The sample README should call out that this sample is experimental then, and may be broken at any time.

samples/switchless/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+## About the host
+
+The host first defines a structure specifically for configuring switchless calls. In this case, we specify the first field `2` as the number of host worker threads for switchless OCALLs. The 2nd field specifies the number of enclave threads for switchless ECALLs. Since switchless ECALL is not yet implementated, we require the 2nd field to be `0`.
+
+```c
+oe_enclave_config_context_switchless_t config = {2, 0};
```
dthaler: I think hard coding the number is almost always guaranteed to be wrong.   I don't think a sample should recommend bad practice, i.e., hard coding numbers that are bad.   So either explain why "2" is the best number (in which case lots of apps can just copy this sample code and hard code the same number), or remove the number 2 from the sample.  E.g., make it be a command line argument.

samples/switchless/README.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+The host then puts the structure address and the configuration type in an array of configurations for the enclave to be created. Even though we only have one configuration (for switchless) for the enclave, we'd like the flexibility of adding more than one configurations (with different types) for an enclave in the future.
```
dthaler: Grammar "one configurations".   Delete the 's'

samples/switchless/README.md
```diff
@@ -0,0 +1,133 @@
...
+             OE_ENCLAVE_TYPE_SGX,
+             flags,
+             configs,
+             OE_COUNTOF(configs),
+             &enclave);
+```
+
+The host then makes an ECALL of `enclave_helloworld` to transition into the enclave. After the ECALL returns, the host terminates the enclave.
+
+As shown in the EDL file, the host exposes two host functions: `host_helloworld` and `host_helloworld_switchless`. The former prints "Hello world from regular OCALL", and the latter prints "Hello world from switchless OCALL".
```
dthaler: This sample demonstrates switchless in a case where switchless is inappropriate (i.e. a call that is not frequent).   If we're doing to demonstrate switchless, it should show a case where you'd actually want switchless, and then explain why this case is using switchless, i.e., why it's better than not using it.

samples/switchless/host/host.c
```diff
@@ -0,0 +1,77 @@
...
+
+    uint32_t flags = OE_ENCLAVE_FLAG_DEBUG;
+    if (check_simulate_opt(&argc, argv))
+    {
+        flags |= OE_ENCLAVE_FLAG_SIMULATE;
+    }
+
+    // Enable switchless and configure host worker number
+    oe_enclave_config_context_switchless_t config = {2, 0};
+    oe_enclave_config_t configs[] = {{
```
soccerGB: Is there any recommendation on how to configure (number threads) to get best performance from this switchless feature?

samples/switchless/switchless.edl
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+enclave {
+    trusted {
+        public void enclave_helloworld();
+    };
```
soccerGB: Does it make sense to also show transition_using_threads support on the ecall?

tests/switchless_threads/enc/CMakeLists.txt
jxyang: That's a good question. Looking at the same cmake file, the UUID is used for signing the enclave binary. But even if it's not provided, the `add_enclave` will generate one on the fly. 

I generated the UUID manually and passed in here, just to follow the tradition of other tests. Do we really need to do so? @andschwa

samples/switchless/README.md
jxyang: Fixed

samples/switchless/host/host.c
jxyang: Added recommendations in `How does OE support switchless OCALLs`

samples/switchless/switchless.edl
jxyang: Switchless ecall hasn't been implemented yet. We will update this sample once it's added.

samples/switchless/README.md
jxyang: Fixed

samples/switchless/README.md
jxyang: Fixed

samples/switchless/README.md
jxyang: Removed the regular ocall.

samples/switchless/README.md
jxyang: I think we will drop the `experimental` flag. Not sure it's in this PR or another PR. @radhikaj?

samples/switchless/README.md
jxyang: Added the logic behind '2' (it's 1 now actually).

samples/switchless/README.md
jxyang: Fixed with a loop calling the switchless function and explained the rational.

samples/switchless/README.md
jxyang: Provided guidance on when switchless calls are appropriate.

samples/switchless/README.md
soccerGB: It would be great to specifically call out that the current switchless call is only supported for OCALL, that is  "switchless ECALL is not yet implemented"

jxyang:
bors try

samples/switchless/host/host.c
soccerGB: Thanks for addressing my comment

samples/switchless/host/host.c
```diff
@@ -0,0 +1,75 @@
...
+             OE_ENCLAVE_TYPE_SGX,
+             flags,
+             configs,
+             OE_COUNTOF(configs),
+             &enclave)) != OE_OK)
+        fprintf(stderr, "oe_create_enclave(): result=%u", result);
+
+    // Call into the enclave
+    result = enclave_add_N(enclave, &m, n);
+    if (result != OE_OK)
```
soccerGB: Here is some of my thoughts about this sample:
The readme has good explanation on how to make best of this switchless feature for certain scenarios (such on frequent calls and short host operations).
To give a user better idea on the benefit, I suggest have two separate ocalls: one with switchless and one without , make them call back to the host for same numbers of times (such as 100 or 1000) on the same simple task (like add_N you are doing), then print out the time take it and compare them. I am pretty it would give developers a better sense on this feature's benefits on top of how to use it.

dthaler: Sounds good to me.

samples/README.md
```diff
@@ -147,3 +147,10 @@ The following samples demonstrate how to develop enclave applications using OE A
   - between two enclaves
   - between one non-enclave client and an enclave
 
+#### [Switchless Calls](switchless/README.md)
+
+- Explain the concept of switchless calls
+- Demonstrate how to enable switchless calls in an enclave application in steps:
```
dthaler: Make lines 152-156 match lines 6-10 of samples/switchless/README.md

samples/switchless/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+Prerequisite: you may want to read [Common Sample Information](../README.md#common-sample-information) before going further.
+
+## Switchless Calls
+
+In an enclave application, the host makes **ECALL**s into functions exposed by the enclaves it created. Likewise,
+the enclaves may make **OCALL**s into functions exposed by the host that created them. In either case, the
+execution has to be transitioned from an untrusted environment to a trusted environment, or vice versa. Since the
+transition is costly due to heavy security checks, it might be more performance advantageous to make the calls
+**context-switchless**: the caller delegates the function call to a worker thread in the other environment, which does the real job of calling the function and post the result to the caller. Both the calling thread and the
```
dthaler: nit: line 20 is too long, split into two lines in the source file

samples/switchless/README.md
```diff
@@ -0,0 +1,184 @@
...
+thread needs to pass the result back to the caller. Both exchanges need to be synchronized.
+
+While switchless calls save transition time, they require at least one additional thread to service the calls.
+More threads typically means more competition of the CPU cores and more context switches, hurting the performance.
+Whether to make a particular function switchless has to weigh the associated costs and savings. In general, the
+good candidates for switchless calls are functions that are: 1) short, thus the transition takes relatively high
+percentage of the overall execution time of the call; and 2) called frequently, so the savings in transition time
+add up.
+
+## How does OpenEnclave support switchless OCALLs
```
dthaler: "Open Enclave" (with a space).  Same in line 37, 38, etc.

samples/switchless/README.md
```diff
@@ -0,0 +1,184 @@
...
+OpenEnclave only supports synchronous switchless OCALLs currently. When the caller within an enclave makes a
+switchless OCALL, the trusted OpenEnclave runtime creates a `job` out of the function call. The `job` object
+includes information such as the function ID, the parameters marshaled into a buffer, and a buffer for holding the
+return value(s). The job is posted to a shared memory region which both the enclave and the host can access.
+
+A host worker thread checks and retrieves `job` from the shared memory region. It uses the untrusted OpenEnclave
+runtime to process the `job` by unmarshaling the parameters, then dispatching to the callee function, and finally
+relaying the result back to the trusted OpenEnclave runtime, which is further forwarded back to the caller.
+
+To support simultaneous switchless OCALLs made from enclaves, the host workers are multi-threaded. OpenEnclave
```
dthaler: "To support simultaneous switchless OCALLs made from enclaves" only applies if there can be multiple simultaneous ECALLs, which not all TEEs or enclaves support.   Maybe prefix this with something like "If an enclave supports multiple simultaneous ECALLs, each ECALL might attempt an OCALL, resulting in simultaneous OCALLs."

samples/switchless/README.md
```diff
@@ -0,0 +1,184 @@
...
+allows users to configure how many host worker threads are to be created for servicing switchless OCALLs. The
+following example illustrates how to do that. A word of caution is that too many host worker threads might increase
+competition of cores between threads and degrade the performance. Therefore, if a enclave has switchless calls
+enabled, OpenEnclave caps the number of host worker threads for it to the number of enclave threads specified.
+
+With the current implementation, we recommend users to avoid using more host worker threads than the minimum of:
```
dthaler: grammar: delete "to"

samples/switchless/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+1. the number of simultaneously active enclave threads, and
+2. the number of cores that are potentially available to host worker threads.
+
+For example, on a 4-core machine, if the number of the simultaneously active enclave threads is 2, and there is no
+host threads other than the two threads making ECALLs and the switchless worker threads, both 1) and 2) would be 2. So we recommend setting the number of host worker threads to 2.
```
dthaler: nit: split into 2 lines

samples/switchless/README.md
```diff
@@ -0,0 +1,184 @@
...
+following example illustrates how to do that. A word of caution is that too many host worker threads might increase
+competition of cores between threads and degrade the performance. Therefore, if a enclave has switchless calls
+enabled, OpenEnclave caps the number of host worker threads for it to the number of enclave threads specified.
+
+With the current implementation, we recommend users to avoid using more host worker threads than the minimum of:
+
+1. the number of simultaneously active enclave threads, and
+2. the number of cores that are potentially available to host worker threads.
+
+For example, on a 4-core machine, if the number of the simultaneously active enclave threads is 2, and there is no
```
dthaler: grammar: change "is no" to "are no"

samples/switchless/README.md
```diff
@@ -0,0 +1,184 @@
...
+**context-switchless**: the caller delegates the function call to a worker thread in the other environment, which does the real job of calling the function and post the result to the caller. Both the calling thread and the
+worker thread never leave their respective execution contexts during the perceived function call.
+
+The calling thread and the worker thread need to exchange information twice during the call. When the switchless
+call is initiated, the caller needs to pass the `job` (encapsulating information regarding the function call in a
+ single object, for details see the next section) to the worker thread. And when the call finishes, the worker
+thread needs to pass the result back to the caller. Both exchanges need to be synchronized.
+
+While switchless calls save transition time, they require at least one additional thread to service the calls.
+More threads typically means more competition of the CPU cores and more context switches, hurting the performance.
```
dthaler: Regarding competition of the CPU cores, I think somewhere in here should call out that the additional thread busy-waits and so consumes lots of CPU.  That should make the "competition" point more understandable.

samples/switchless/README.md
```diff
@@ -0,0 +1,184 @@
...
+        public void enclave_add_N([in, out] int* m, int n);
+    };
+
+    untrusted {
+        void host_increment([in, out] int* m) transition_using_threads;
+    };
+};
+```
+
+Function `host_increment`'s declaration ends with keyword `transition_using_threads`, indicating it should be called switchlessly at run time. However, this a best-effort directive. OpenEnclave runtime may still choose to fall back to a tradition OCALL if switchless call resources are unavailable, e.g., the enclave is not configured as switchless-capable, or the host worker threads are busy servicing other switchless OCALLs.
```
dthaler: nit: break into multiple lines

dthaler: grammar: insert "The" before "Open Enclave runtime".  Also insert "The" before "function"

samples/switchless/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+To generate the functions with the marshaling code, the `oeedger8r` tool is called in both the host and enclave directories from their Makefiles. For example:
+
+```bash
+cd host
+oeedger8r ../switchless.edl --untrusted --experimental
```
dthaler: Is "--experimental" still needed?   I thought the goal was to make this feature non-experimental?

samples/switchless/README.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+`oeedger8r` needs the command line flag `--experimental` to be able to recognize the keyword `transition_using_threads`.
+
+## About the host
+
+The host first defines a structure specifically for configuring switchless calls. In this case, we specify the
+first field `1` as the number of host worker threads for switchless OCALLs. In this example, 1) There is at most
+1 enclave thread all the time, and 2) The number of cores available to the host worker threads is  1 (assuming a
```
dthaler: I would change "2)" to something like "The number of cores available to the host worker threads is unknown, and so we use 1 as explained above."

Rationale: We should not have sample code that "assumes" a 2-core or a 4-core machine.

samples/switchless/README.md
```diff
@@ -0,0 +1,184 @@
...
+2-core machine) or 3 (assuming a 4-core machine). In any case, The minimum of 1) and 2) is 1, which we choose to
+be the number of host worker threads. The 2nd field specifies the number of enclave threads for switchless ECALLs.
+Since switchless ECALL is not yet implemented, we require the 2nd field to be `0`.
```
dthaler: By "require", do you mean that the API will fail if you pass non-zero?   Why?  Line 87 already explained it's just a hint. Suggest changing "we require the 2nd field to be 0" to something like "we can just pass 0".

However, on further thought, since the sample does not have transition_using_threads on any ECALL, shouldn't the explanation be that that's why we pass 0?  I.e., if OE did support switchless ECALLs, and we didn't use that marking on any ECALL, then passing a non-0 number would be useless, wouldn't it?

samples/switchless/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+```c
+oe_enclave_config_context_switchless_t config = {1, 0};
+```
+
+The host then puts the structure address and the configuration type in an array of configurations for the enclave to be created. Even though we only have one configuration (for switchless) for the enclave, we'd like the flexibility of adding more than one configuration (with different types) for an enclave in the future.
```
dthaler: line length

dthaler: I don't think "more than one configuration" is what we mean, since in normal English, configurations are alternatives, meaning you only use one configuration at a time (e.g., configurations of a project in Visual Studio, etc.).   Instead I think they are configuration options, or configuration elements, or configuration parameters.

This may mean the variable name and structure name are both probably needing to change for clarity.

tests/switchless_threads/host/host.c
```diff
@@ -0,0 +1,156 @@
...
+#include <stdlib.h>
+#include <string.h>
+#if __GNUC__
+#include <pthread.h>
+#elif _MSC_VER
+#include <windows.h>
+#endif
+#include "switchless_threads_u.h"
+
+#define NUM_HOST_THREADS 7
```
jazzybluesea: Maybe add a comment here on how you came up with 7 host threads.

samples/switchless/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+With the current implementation, we recommend users to avoid using more host worker threads than the minimum of:
+
+1. the number of simultaneously active enclave threads, and
+2. the number of cores that are potentially available to host worker threads.
+
+For example, on a 4-core machine, if the number of the simultaneously active enclave threads is 2, and there is no
+host threads other than the two threads making ECALLs and the switchless worker threads, both 1) and 2) would be 2. So we recommend setting the number of host worker threads to 2.
+
+The exception to the above rule happens when 1) or 2) is zero or negative. For example, if the host starts two more
```
jazzybluesea: Maybe adding a table would make it more clear?

jxyang: This is likely to change with a pending PR from Anand. I will just rephrase to make it more clear.

samples/switchless/README.md
jxyang: @mikbras once mentioned using `settings` instead of `configuration`. Inherently `settings` implies plurality. Does that sound a better choice? But I am not sure whether we call each structure in the array a `setting` or `settings`. It's for a single feature, but could have multiple knobs.

samples/switchless/README.md
jxyang: By "require", I mean the `oe_create_enclave` API will fail. This field is reserved for future versions when we enable switchless ECALLs, thus it makes sense to force it to be 0 now, following the examples of reserved bits in network protocols.

@CodeMonkeyLeet actually suggested this way. He can probably add more reasons behind it.

samples/switchless/README.md
jxyang: Fixed

samples/switchless/README.md
jxyang: Thanks for the suggestion.

samples/switchless/README.md
jxyang: Fixed.

samples/README.md
jxyang: Good catch. Fixed.

samples/switchless/host/host.c
jxyang: Good suggestion. In fact `tests/switchless` is implemented this way. I will port the code to this sample.

jxyang:
bors try

samples/switchless/README.md
dthaler: oe_enclave_setting_context_switchless_t setting = {1, 0};
would be fine with me.

radhikaj:
Bors try

tests/switchless_threads/host/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+add_custom_command(
+  OUTPUT switchless_threads_u.h switchless_threads_u.c switchless_threads_args.h
+  DEPENDS
+  edger8r
+  ../switchless_threads.edl
+  COMMAND edger8r --experimental --untrusted --search-path ${CMAKE_CURRENT_SOURCE_DIR}/.. switchless_threads.edl)
```
dthaler: Is --experimental still needed here?

tests/switchless_threads/host/host.c
```diff
@@ -0,0 +1,159 @@
...
+#include <stdlib.h>
+#include <string.h>
+#if __GNUC__
+#include <pthread.h>
+#elif _MSC_VER
+#include <windows.h>
+#endif
+#include "switchless_threads_u.h"
+
+// The enclave supports up to 8 concurrent threads in it. We have to reserve
```
dthaler: This statement isn't entirely true.  The number 8 in enc/enc.c is inside
OE_SET_ENCLAVE_SGX() which means it's only true for SGX.  The code here has no such limitation,
and we want switchless to apply to OP-TEE as well.  So minimally the comment should be fixed,
but ideally (not required for this PR), the code should be such that it won't accidentally compile for OP-TEE.
That is, compile time failures are better than runtime failures, until OP-TEE support is added.

tests/switchless_threads/host/host.c
```diff
@@ -0,0 +1,159 @@
...
+    if (argc != 2)
+    {
+        fprintf(stderr, "Usage: %s ENCLAVE_PATH\n", argv[0]);
+        return 1;
+    }
+
+    const uint32_t flags = oe_get_create_flags();
+
+    // Enable switchless and configure host worker number
+    oe_enclave_config_context_switchless_t config = {2, 0};
```
dthaler: Add comment here explaining why "2" is the right number to use.  Why isn't it 1?

samples/switchless/README.md
radhikaj: Configuration Parameters looks good to me. So "we'd like the flexibility of adding more than one configuration parameter".

tests/switchless_threads/host/host.c
jxyang: It should be 1. Thanks for catching it.

tests/switchless_threads/host/host.c
jxyang: Fixed the comments.

tests/switchless_threads/host/CMakeLists.txt
jxyang: We haven't decided whether to drop it altogether or use a `--preview` flag. @radhikaj

jxyang:
bors try

jxyang:
bors try

samples/switchless/README.md
andschwa: Maybe he's using Deep Copy...?

tests/switchless_threads/host/CMakeLists.txt
andschwa: We could rename experimental to preview, I'm fine with that. (In fact, I quite like it.)

dthaler: Regardless of the name (I have no preference between preview vs experimental), the edger8r support is a preview and the rest of the switchless feature has a hard dependency on it, then the whole switchless feature should be a preview in my view.

samples/switchless/README.md
radhikaj: @andschwa  No, he is not using deep copy.

tests/switchless_threads/host/CMakeLists.txt
radhikaj: I think we can go with --preview for switchless support. 
@dthaler : Do we have documentation that edger8r is in preview right now?

radhikaj: We can switch --experimental to --preview in a different PR.

samples/switchless/README.md
radhikaj: Opened issue #2202 to track @dthaler 's recommendation for the change in variable name

tests/switchless_threads/host/CMakeLists.txt
radhikaj: Opened Issue #2203 to track this.

anakrish:
bors try

include/openenclave/host.h
```diff
@@ -157,13 +150,8 @@ oe_result_t oe_create_enclave(
     const char* path,
     oe_enclave_type_t type,
     uint32_t flags,
-#ifdef OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE
     const oe_enclave_config_t* configs,
```
dthaler: For consistency with the other changes, I believe this should be something like:
oe_enclave_setting_t* settings,

host/optee/linux/enclave.c
```diff
@@ -399,13 +399,8 @@ oe_result_t oe_create_enclave(
     const char* enclave_path,
     oe_enclave_type_t enclave_type,
     uint32_t flags,
-#ifdef OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE
     const oe_enclave_config_t* configs,
```
dthaler: oe_enclave_setting_t* settings,
uint32_t setting_count,

anakrish: Fixed.

host/optee/linux/enclave.c
```diff
@@ -430,12 +425,8 @@ oe_result_t oe_create_enclave(
     if (!enclave_path || !enclave_out ||
         ((enclave_type != OE_ENCLAVE_TYPE_OPTEE) &&
          (enclave_type != OE_ENCLAVE_TYPE_AUTO)) ||
-#ifdef OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE
-        (config_count > 0 && configs == NULL) ||
-        (config_count == 0 && configs != NULL) ||
-#else
-        config || config_size > 0 ||
-#endif
+        (setting_count > 0 && configs == NULL) ||
```
dthaler: settings == NULL

samples/switchless/README.md
dthaler: Would still like this comment addressed by adding a phrase or sentence, but not a blocking issue.

samples/switchless/README.md
dthaler: Would still like this nit fixed, but not a blocking issue.

host/optee/linux/enclave.c
anakrish: Fixed. Thanks for catching this!

samples/switchless/README.md
anakrish: Added a sentence to indicate that currently the worker threads busy-wait.

samples/switchless/README.md
anakrish: Changes two occurrences.

PR #2198, opened 2019-09-25
Fixed issue with decrypting the encryption key with the password key

ichetty:
The current code was using the previous encryption key generated from previous encrypt operation which was cached in the global dispatcher object. The decrypt operation didn't alter the encryption key due to couple of reasons.

### DIFF
--- a/samples/file-encryptor/enclave/keys.cpp
+++ b/samples/file-encryptor/enclave/keys.cpp
@@ -226,7 +226,12 @@ int ecall_dispatcher::cipher_encryption_key(
     mbedtls_aes_init(&aescontext);
 
     // set aes key
-    ret = mbedtls_aes_setkey_enc(&aescontext, encrypt_key, ENCRYPTION_KEY_SIZE);
+    if (encrypt){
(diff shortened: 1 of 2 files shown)


andschwa:
Thanks for the PR @ichetty! @soccerGB can you take a look at this?

samples/file-encryptor/enclave/keys.cpp
```diff
@@ -226,7 +226,12 @@ int ecall_dispatcher::cipher_encryption_key(
     mbedtls_aes_init(&aescontext);
 
     // set aes key
-    ret = mbedtls_aes_setkey_enc(&aescontext, encrypt_key, ENCRYPTION_KEY_SIZE);
+    if (encrypt){
```
dthaler: style: { should be on new line, here and in 232, for consistency with rest of file

ichetty: Done!

ichetty:
Let me know if anything is needed from my side for this to be merged and closed. Thank you!

andschwa:
Thanks @ichetty, merging on behalf of @soccerGB 

bors r+

PR #2200, opened 2019-09-25
Switchless Ocall Worker Waiting Strategy Improvement

anakrish:
Each host worker checks for a message in its call_arg slot.
If there is a message, the message is immediately removed from
the slot and processed.

If there is no message, then a spin counter is incremented.
If the spin counter reaches a particular threshold, then it means
that the thread has been spinning for sometime waiting for messages
and now can be put to sleep.

A simple strategy of putting the thread to sleep is to sleep for
a predetermined amount of time say 1 microsecond. This has the
benefit of letting other threads run; however it does not give
optimal performance since the worker threads keep waking up,
spinning and going back to sleep for ever, consuming enough CPU
to affect normal ocall performance.

A better strategy is to use a futex to wait.
Each thread has an int32_t event field, initially zero.

When a host thread has spun till the threshold,
    if atomically sets the value of event to 0 if it was 1

    if the previous value (returned from atomic operation) was 0,
    it means that there has been no wake notifications from the enclave,
    and therefore the host thread does a futex_wait while event's value
    is still zero.

    if the previous value (returned from atomic operation) was 1,
    if means that the enclave has just posted a message and therefore
    the worker thread continues to fetch the message instead of sleeping.

When an enclave thread wants to post a switchless ocall message,
   it first finds a host worker thread with a free slot
       The host worker thread could be active or doing a futex_wait

       The enclave thread atomically sets the value of the host thread's
       event to 1 if it was 0

       if the previous value(returned from atomic operation) was 0,
       it means that the host thread is doing a futex_wait and therefore
       an core ocall (OE_OCALL_WAKE_HOST_WORKER) is made to futex_wake the thread.
       The number of such ocalls will be less in an application that makes switchless
       calls with enough frequency. (If switchless calls are rare, the application may
       not benefit from switchless calls).

       If the previous value(returned from atomic operation) was 1,
       it means that we have already woken up the host and therefore don't need to
       wake up the host again.

   if no host worker thread with free slot is found, the post fails and
   subsquently a normal ocall is made

On the enclave side, an atomic compare and exchange operation suffices to determine
if the host must be woken up via a core ocall or not.

On the host side, in linux futex syscalls (__NR_futex FUTEX_WAIT_PRIVATE and FUTEX_WAKE_PRIVATE)
are used whereas on Windows WaitOnAddress and WakeByAddressSingle are used.

### DIFF
--- a/enclave/core/switchlesscalls.c
+++ b/enclave/core/switchlesscalls.c
@@ -71,7 +71,6 @@ oe_result_t oe_handle_init_switchless(uint64_t arg_in)
     // Copy the worker context array pointer and its size to avoid TOCTOU
     _host_worker_count = safe_manager.num_host_workers;
     _host_worker_contexts = safe_manager.host_worker_contexts;
-
     result = OE_OK;
 
 done:
@@ -99,18 +98,49 @@ oe_result_t oe_post_switchless_ocall(oe_call_host_function_args_t* args)
     size_t tries = _host_worker_count;
     while (tries--)
     {
+        // Check if the worker's slot is free.
         if (_host_worker_contexts[tries].call_arg == NULL)
         {
+            // Try to atomically grab the slot by placing args in the slot.
+            // If the atomic operation was successful, then the worker thread
+            // will execute this switchless ocall. If the atomic operation
+            // failed, this means that the slot was grabbed by another
+            // switchless ocall and therefore, we must scan for another worker
+            // thread with a free slot.
             if (oe_atomic_compare_and_swap_ptr(
                     (void* volatile*)&_host_worker_contexts[tries].call_arg,
                     NULL,
                     args))
             {
-                // Set the worker's state to RUNNING.
-                if (__sync_val_compare_and_swap(
-                        &_host_worker_contexts[tries].event, 0, 1) == 0)
+                // The worker thread has been marked to execute this switchless
+                // call. Determine if it needs to be woken up or not.
+                //
+                // If event is 0, it means that it has gone to sleep. Wake it by
+                // making an ocall (OE_OCALL_WAKE_HOST_WORKER).
+                // Note: it is important to use an atomic cas operation to set
+                // the value to 1 before making the ocall. Setting the value to
+                // 1 prevents the host worker from simulataneously going to
+                // sleep. If instead, just a compare operation is used to
+                // determine if the host thread is sleeping or not, the host
+                // thread could go to sleep after the enclave has determined
+                // that the host is not sleeping, causing a deadlock.
+                //
+                // If event is 1, that indicates a pending wake notification.
+                int32_t oldval = 0;
+                int32_t newval = 1;
+                // Weak operation could sporadically fail.
+                // We need a strong operation.
+                bool weak = false;
+                if (__atomic_compare_exchange_n(
+                        &_host_worker_contexts[tries].event,
+                        &oldval,
+                        newval,
+                        weak,
+                        __ATOMIC_ACQ_REL,
+                        __ATOMIC_ACQUIRE))
                 {
-                    // The worker was previously sleeping.
+                    // The pevious value of the event was 0 which means that the
+                    // worker was previously sleeping.
                     // Wake it via an ocall.
                     oe_ocall(
                         OE_OCALL_WAKE_HOST_WORKER,
--- a/enclave/core/switchlesscalls.c
+++ b/enclave/core/switchlesscalls.c
@@ -71,6 +71,7 @@ oe_result_t oe_handle_init_switchless(uint64_t arg_in)
     // Copy the worker context array pointer and its size to avoid TOCTOU
     _host_worker_count = safe_manager.num_host_workers;
     _host_worker_contexts = safe_manager.host_worker_contexts;
+
     result = OE_OK;
 
 done:
@@ -105,6 +106,18 @@ oe_result_t oe_post_switchless_ocall(oe_call_host_function_args_t* args)
                     NULL,
                     args))
             {
+                // Set the worker's state to RUNNING.
+                if (__sync_val_compare_and_swap(
+                        &_host_worker_contexts[tries].event, 0, 1) == 0)
+                {
+                    // The worker was previously sleeping.
+                    // Wake it via an ocall.
+                    oe_ocall(
+                        OE_OCALL_WAKE_HOST_WORKER,
+                        (uint64_t)&_host_worker_contexts[tries],
+                        NULL);
+                }
+
                 return OE_OK;
             }
         }
--- a/host/CMakeLists.txt
+++ b/host/CMakeLists.txt
@@ -165,6 +165,7 @@ if (OE_SGX)
       sgx/linux/entersim.S
       sgx/linux/exception.c
       sgx/linux/sgxioctl.c
+      sgx/linux/switchless.c
       sgx/linux/xstate.c)
   else()
     list(APPEND PLATFORM_HOST_ONLY_SRC
@@ -176,6 +177,7 @@ if (OE_SGX)
       sgx/windows/enter.asm
       sgx/windows/entersim.asm
       sgx/windows/exception.c
+      sgx/windows/switchless.c
       sgx/windows/xstate.c)
   endif()
 
@@ -294,6 +296,8 @@ if (UNIX)
 elseif (WIN32)
   target_include_directories(oehost PRIVATE
     ${CMAKE_SOURCE_DIR}/3rdparty/mbedtls/mbedtls/include)
+  # Synchronization library is needed for WaitOnAddress/WakeByAddress functions
+  # used by switchless ocalls worker threads.
   target_link_libraries(oehost PRIVATE bcrypt Crypt32 Synchronization)
   target_include_directories(oehostverify PRIVATE
     ${CMAKE_SOURCE_DIR}/3rdparty/mbedtls/mbedtls/include)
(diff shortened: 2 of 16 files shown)


anakrish:
bors try

host/sgx/calls.c
```diff
@@ -339,6 +339,7 @@ static const char* oe_ocall_str(oe_func_t ocall)
         "FREE",
         "SLEEP",
         "GET_TIME",
+        "FUTEX_WAKE",
```
dthaler: Isn't "FUTEX_WAKE" Linux-specific?   I'd prefer we not have Linux-specific names in non-Linux-specific files.

host/sgx/switchless.c
```diff
@@ -10,6 +11,68 @@
 #include "../ocalls.h"
 #include "enclave.h"
 
+#if __linux__
```
dthaler: rather than having #if _linux__ around Linux-specific code, it should be moved to a linux-specific file.   Just like we have host/optee/linux/enclave.c for the same reason, this should be in host/sgx/linux/switchless.c

tests/CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ if (UNIX OR ADD_WINDOWS_ENCLAVE_TESTS OR USE_CLANGW)
         add_subdirectory(SampleAppCRT)
         add_subdirectory(sealKey)
         add_subdirectory(stdc)
+	add_subdirectory(switchless)
```
dthaler: nit: indentation looks off, maybe a tabs vs spaces issue?

tests/switchless/host/host.cpp
```diff
@@ -0,0 +1,197 @@
...
+    if (argc < 2)
+    {
+        fprintf(
+            stderr,
+            "Usage: %s ENCLAVE_PATH [num-host-threads] [num-enclave-threads]\n",
+            argv[0]);
+        return 1;
+    }
+
+    uint64_t num_host_threads = 2;
```
dthaler: why 2 instead of 1?    PR 2195 just changed from using 2 to 1.  Be consistent.

tests/switchless/host/host.cpp
```diff
@@ -0,0 +1,197 @@
...
+    // Enable switchless and configure host worker number
+    oe_enclave_config_context_switchless_t config = {num_host_threads, 0};
+    oe_enclave_config_t cfg;
+    cfg.config_type = OE_ENCLAVE_CONFIG_CONTEXT_SWITCHLESS;
+    cfg.u.context_switchless_config = &config;
+    oe_enclave_config_t configs[] = {cfg};
+
+    if ((result = oe_create_switchless_enclave(
+             argv[1],
+             OE_ENCLAVE_TYPE_SGX,
```
dthaler: Why "SGX" and not "AUTO"?  Same question for 138

host/sgx/switchless.c
```diff
@@ -10,6 +11,68 @@
...
+#include <sys/syscall.h>
+#include <unistd.h>
+
+static void _worker_wait(oe_host_worker_context_t* context)
+{
+    if (__sync_val_compare_and_swap(&context->event, 1, 0) == 0)
+    {
+        do
+        {
+            syscall(
```
radhikaj: Does syscall return an error in any scenario?

host/sgx/calls.c
mikbras: On Windows I suppose this will use **WaitOnAddress**. Maybe "WAIT_ON_ADDRESS" would be more descriptive?

host/sgx/switchless.c
mikbras: Hmm? It's already under the sgx directory, which limits us to just Linux and Windows? This means we would now have both "host/linux" and "host/sgx/linux" directories. I guess it makes sense.

host/CMakeLists.txt
```diff
@@ -351,6 +351,7 @@ target_compile_definitions(oehost
   # package.
   $<BUILD_INTERFACE:OE_API_VERSION=2>
   PRIVATE
+  OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE
```
mikbras: Did we decide to make this preview and if so should we remove this?

enclave/core/switchlesscalls.c
```diff
@@ -105,6 +106,18 @@ oe_result_t oe_post_switchless_ocall(oe_call_host_function_args_t* args)
                     NULL,
                     args))
             {
+                // Set the worker's state to RUNNING.
+                if (__sync_val_compare_and_swap(
+                        &_host_worker_contexts[tries].event, 0, 1) == 0)
```
mikbras: Related to the previous code: Why do we iterate backwards rather than forwards?

host/sgx/switchless.c
mikbras: Only if the input parameters are invalid, which we have ensured that they are not invalid. Besides it does not matter because we are looping until the event is non-zero.

host/sgx/switchless.c
```diff
@@ -37,6 +114,12 @@ static oe_result_t oe_stop_worker_threads(oe_switchless_call_manager_t* manager)
     for (size_t i = 0; i < manager->num_host_workers; i++)
     {
         manager->host_worker_contexts[i].is_stopping = true;
+        _worker_wake(&manager->host_worker_contexts[i]);
+
+        OE_TRACE_INFO(
+            "Switchless host worker thread %d spun for %lu times",
```
mikbras: %d does not match size_t. I wonder why the compiler did not catch this?

include/openenclave/internal/switchless.h
```diff
@@ -12,10 +12,21 @@
 typedef struct _host_worker_thread_context
 {
     volatile oe_call_host_function_args_t* call_arg;
-    volatile bool is_stopping;
+    uint64_t is_stopping;
```
mikbras: Did you make this uint64_t so it would alignment reasons?

host/sgx/switchless.c
dthaler: we already have both host/linux and host/optee/linux for exactly the same reason.

host/sgx/switchless.c
anakrish: @mikbras Thanks for the explanation.
Added comment indicating that it is OK to ignore the syscall return values since we are looping until event is non-zero.

host/sgx/switchless.c
anakrish: Fixed. Added host/sgx/inux/switchless.c and host/sgx/windows/switchless.c

enclave/core/switchlesscalls.c
anakrish: There is no technical reason. Just seems to be a coding style. Keeping it as is.

host/CMakeLists.txt
anakrish: Removed.

host/sgx/switchless.c
anakrish: Fixed.

include/openenclave/internal/switchless.h
anakrish: Yes. I have now locked the entire layout using static-asserts.

anakrish:
bors try

host/sgx/calls.c
anakrish: Fixed. Changed to `WAKE_HOST_WORKER` to match enum name.

tests/CMakeLists.txt
anakrish: Fixed.

tests/switchless/host/host.cpp
anakrish: Pre-existing code. Currently switchless is supported only on sgx.

include/openenclave/internal/switchless.h
```diff
@@ -12,10 +12,30 @@
...
+    // Number of times the worker spinned without seeing a message.
+    uint64_t spin_count;
+
+    // Statistics.
+    uint64_t total_spin_count;
 } oe_host_worker_context_t;
 
+/**
+ * oe_host_worker_context_t is used both by the host (windows/linux) and the
+ * enclave (ELF). Lockdown the layout.
```
dthaler: nit: "Lock down" (2 words) when used as a verb

tests/switchless/host/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@
 
 oeedl_file(../switchless.edl host gen)
 
+
```
dthaler: why the whitespace-only change?

tests/switchless/host/host.c
```diff
@@ -12,15 +12,46 @@
...
+
 #define STRING_LEN 100
 
 #if _MSC_VER
 static double frequency;
 #endif
 
+static int thread_create(oe_thread_t* thread, void* (*func)(void*), void* arg)
+{
+#if __GNUC__
```
dthaler: The #if tests look to me to be testing the wrong thing.   This looks more about whether it's Linux/posix or windows, rather than which compiler you're using.   So things like __attribute__ (which isn't used here) should be inside such checks, but Linux vs windows tests should not be based on any assumption about the compiler.

include/openenclave/internal/switchless.h
anakrish: Fixed.

tests/switchless/host/CMakeLists.txt
anakrish: Fixed.

tests/switchless/host/host.c
anakrish: Good point. I changed code to use __linux__/_WIN32 instead.

Additionally, I switched to using oe_thread_create/oe_thread_join which removes all this logic in the test. I also removed it from tests/switchless_threads/host/host.c where I had borrowed it from.

tests/switchless/host/host.cpp
anakrish: Changed it to 1.

anakrish:
bors try

Issue #2206, opened 2019-09-26
SGX: Add support for attestation collaterals.


jazzybluesea:

jazzybluesea:
Resolved by change #2298.  Closing this issue.

PR #2213, opened 2019-10-01
Add collateral to host remote report verification API

wintersteiger:
Is this roughly how remote quote verification collaterals will be exposed via the host verification library?


jazzybluesea:
> Is this roughly how remote quote verification collaterals will be exposed via the host verification library?

@wintersteiger Yes that's the plan.  However, we need to refactor the SGX collateral struct a little bit such that it is easily serializable/deserializable.  This will be addressed by #2206.

wintersteiger:
Great, thanks for the pointer @jazzybluesea!

Issue #1520, opened 2019-03-02
Standardize OE SDK ABI expectations


CodeMonkeyLeet:
As tests are starting to expose irregularities in the way we handled context preservation calling in and out of the enclave between host OS types, we need to more rigorously define and test the edge function ABI and other related call handling we manually author:

- [x] Evaluate and formalize context preserved on ecall from host based on registers preserved in [`__morestack`](https://github.com/Microsoft/openenclave/blob/0e58bcd77b0768543e54aeac6d0725b21684e45b/host/sgx/linux/enter.S#L9) and [`oe_snap_current_context`](https://github.com/Microsoft/openenclave/blob/1e22196bea525a7781883e6f33344c0440bf4529/include/openenclave/internal/context.inc#L26) _[Update: No additional reconciliation necessary between `__morestack` and `oe_snap_current_context` as they service two different scenarios. See comments below.]_
    - It looks like we could possibly optimize the set of registers currently being preserved so we can make a more straightforward statement around host edge function ABI being aligned with the platform its running on. 
- [x] Evaluate and formalize context preserved on ocall from enclave based on registers preserved/cleared/restored in [`oe_asm_exit`](https://github.com/Microsoft/openenclave/blob/1e22196bea525a7781883e6f33344c0440bf4529/enclave/core/sgx/exit.S#L51) and [`oe_enter`](https://github.com/Microsoft/openenclave/blob/1e22196bea525a7781883e6f33344c0440bf4529/enclave/core/sgx/enter.S#L38) _[Update: See documentation in comments below]_
    - It looks like enclave preserves a much smaller set of state on returning from ocall relative to the host, and this asymmetry should be rationalized. It would be ideal if the enclave ocall ABI can be understood in terms of the regular Linux x64 ABI as well.
- [x] Fix #1501 by having Windows host also preserve `mxcsr`
- [x] Merge fixes to preservation of host context on return from enclave for compatibility with Windows x64 ABI from [WinOE1_elibc branch](https://github.com/Microsoft/openenclave/compare/WinOE1_elibc)
    - Don't need to port .asm implementations of `oe_setjmp`/`oe_longjmp` since we're not compiling enclave as PE, but consider using the model to refactor existing Linux implementation to share a jump buffer struct definition for consistency between the calls.
- ~~Update enclave `oe_setjmp`/`oe_longjmp` to match Linux x64 ABI by preserving `mxcsr` & `fpcsr` (x87 control word) non-volatile bits as well.~~
[...]

PR #2223, opened 2019-10-02
Preserve Windows & Linux ABI expectations on ECALL

CodeMonkeyLeet:
This changeset addresses some of the issues in #1520 by reconciling some of the ABI inconsistencies when calling from a Windows host process into an Linux-based enclave. This includes:

- Initialize the MXCSR, RFLAGS and x87 FPU control word on ECALL to ensure that
  the enclave can expect the Linux x86_64 ABI-specified initial process state.

- Preserve the Windows x64 ABI-specified callee-preserved registers, x87 FPU,
  MMX, XMM and MXCSR states before each ECALL and restore them on return or
  OCALL.

- Update return_from_ecall to use r10 instead of callee-preserved rbx register
  as a temporary register.

### DIFF
--- a/enclave/core/sgx/enter.S
+++ b/enclave/core/sgx/enter.S
@@ -119,6 +119,23 @@ oe_enter:
     mov %rsp, %rbp
 
 .call_function:
+    // Set the MXCSR according to the Linux x86_64 ABI
+    mov $ABI_MXCSR_INIT, %r10
+    push %r10
+    ldmxcsr (%rsp)
+    pop %r10
+
+    // Set the FPU Control Word according to the Linux x86_64 ABI
+    mov $ABI_FPUCW_INIT, %r10
+    push %r10
+    fldcw (%rsp)
+    pop %r10
+
+    // Initialize the RFLAGS prior to calling enclave functions
+    // This only clears the DF and state flag bits since
+    // the system flags and reserved bits are not writable here
+    push $0
+    popfq
 
     // Get the host stack pointer.
     mov %gs:td_host_rsp, %r8
--- a/host/sgx/windows/enter.asm
+++ b/host/sgx/windows/enter.asm
@@ -4,6 +4,8 @@
 include ksamd64.inc
 
 extern __oe_dispatch_ocall:proc
+extern oe_save_host_context:proc
+extern oe_restore_host_context:proc
 
 ;;==============================================================================
 ;;
@@ -16,13 +18,14 @@ extern __oe_dispatch_ocall:proc
 ;;     [OUT] uint64_t* arg4,
 ;;     [OUT] oe_enclave_t* enclave);
 ;;
-;; Registers:
+;; Parameters passed on register and stack:
 ;;     RCX      - tcs: thread control structure (extended)
 ;;     RDX      - aep: asynchronous execution procedure
 ;;     R8       - arg1
 ;;     R9       - arg2
 ;;     [RBP+48] - arg3
 ;;     [RBP+56] - arg4
+;;     [RBP+64] - enclave
 ;;
 ;; These registers may be destroyed across function calls:
 ;;     RAX, RCX, RDX, R8, R9, R10, R11
@@ -33,7 +36,11 @@ extern __oe_dispatch_ocall:proc
 ;;==============================================================================
 
 ENCLU_EENTER    EQU 2
-PARAMS_SPACE    EQU 128
+
+ARG3_PARAM      EQU [rbp+48]
+ARG4_PARAM      EQU [rbp+56]
+ENCLAVE_PARAM   EQU [rbp+64]
+
 TCS             EQU [rbp-8]
 AEP             EQU [rbp-16]
 ARG1            EQU [rbp-24]
@@ -44,13 +51,21 @@ ENCLAVE         EQU [rbp-56]
 ARG1OUT         EQU [rbp-64]
 ARG2OUT         EQU [rbp-72]
 STACKPTR        EQU [rbp-80]
-MXCSR           EQU [rbp-88]
+HOST_CONTEXT    EQU [rbp-88]
+
+;; Reserve parameter space based on:
+;; + 88 bytes for 11*8-byte parameters TCS to HOST_CONTEXT
+;; HOST_CONTEXT points to the start of the context data consisting of:
+;; + 64 bytes for 8*8-byte callee-preserved registers
+;; + 512-byte OE_CONTEXT_FLOAT memory image used in fxsave/fxrstor
+;; + 8 bytes so that the stack remains 16-byte aligned.
+PARAMS_SPACE    EQU 672
 
 NESTED_ENTRY oe_enter, _TEXT$00
     END_PROLOGUE
 
     ;; Setup stack frame:
-    push rbp
+    push rbp ;; Stack is 16-byte aligned at this point
     mov rbp, rsp
 
     ;; Save parameters on stack for later reference:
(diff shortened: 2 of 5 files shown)


enclave/core/sgx/enter.S
```diff
@@ -50,6 +50,24 @@ oe_enter:
     mov %rsp, %gs:td_host_rsp
     mov %rbp, %gs:td_host_rbp
 
+    // Set the MXCSR according to the Linux x86_64 ABI
+    mov $ABI_MXCSR_INIT, %r10
```
anakrish: Great change that brings us closer to ABI compliance.

enclave/core/sgx/enter.S
```diff
@@ -50,6 +50,24 @@ oe_enter:
...
+    push %r10
+    ldmxcsr (%rsp)
+    pop %r10
+
+    // Set the FPU Control Word according to the Linux x86_64 ABI
+    mov $ABI_FPUCW_INIT, %r10
+    push %r10
+    fldcw (%rsp)
+    pop %r10
```
anakrish: Unfortunately, rsp still points to host stack at this time.
We could define a macro to initialize these values and then call the macro in exception_entry, nested_entry and clean_entry paths after the enclave rsp has been setup.

host/sgx/windows/enter.asm
```diff
@@ -74,22 +83,15 @@ NESTED_ENTRY oe_enter, _TEXT$00
...
-    ;; Save registers:
-    push rbx
-    push rdi
-    push rsi
-    push r12
-    push r13
-    push r14
-    push r15
+    ;; Store the host context on the host stack
+    mov HOST_CONTEXT, rsp
```
anakrish: This line is hard to read. 
The comment in line 53 says `64 bytes for all preserved registers in HOST_CONTEXT` but this line just stores the current rsp value in HOST_CONTEXT

enclave/core/sgx/enter.S
CodeMonkeyLeet: Good catch ... instead of putting it in 3 different places, could we just put it into the start of `call_function` instead? That's where all of the entries dovetail after restoring the stack pointer.

host/sgx/windows/host_context.asm
```diff
@@ -0,0 +1,78 @@
...
+.CODE
+
+PUBLIC oe_save_host_context
+oe_save_host_context PROC
+    ;; Subroutine prologue
+    push rbp 
+    mov rbp, rsp
+
+    ;; Save the flags to stack.
+    pushf
```
anakrish: Should this function also save MXCSR?

host/sgx/windows/host_context.asm
```diff
@@ -0,0 +1,78 @@
...
+    mov rbx, [rcx+OE_CONTEXT_RBX]
+    mov rdi, [rcx+OE_CONTEXT_RDI]
+    mov rsi, [rcx+OE_CONTEXT_RSI]
+    mov r12, [rcx+OE_CONTEXT_R12]
+    mov r13, [rcx+OE_CONTEXT_R13]
+    mov r14, [rcx+OE_CONTEXT_R14]
+    mov r15, [rcx+OE_CONTEXT_R15]
+
+    ;; Restore x87 and SSE values.
+    fxrstor [rcx+OE_CONTEXT_FLOAT]
```
anakrish: Should this function also restore MXCSR?

host/sgx/windows/enter.asm
```diff
@@ -149,29 +171,19 @@ return_from_ecall:
...
-    ;; Restore registers:
-    pop r15
-    pop r14
-    pop r13
-    pop r12
-    pop rsi
-    pop rdi
-    pop rbx
-
+    mov qword ptr [rax], r10
```
anakrish: Suggestion about pre-existing code.
Consider using a define instead of hard-coded offsets (e.g. rbp+48)

enclave/core/sgx/enter.S
anakrish: Nice idea. 
I checked that none of the instruction in different control paths leading to `call_function` can be influenced by host supplied values of these flags/registers, and therefore it is safe change.

host/sgx/windows/host_context.asm
CodeMonkeyLeet: MXCSR is preserved as part of the [fxsave instruction](https://www.felixcloutier.com/x86/fxsave#tbl-3-43), see comment on line 37

host/sgx/windows/host_context.asm
CodeMonkeyLeet: Similarly, fxrstor restores the MXCSR register

host/sgx/windows/enter.asm
CodeMonkeyLeet: Conceptually, the `HOST_CONTEXT` is just a pointer to the start of that block of information used in host_context.asm, which includes the 8 registers and fxsave memory image and is just sequentially pushed onto the stack. 

What would you suggest to clarify this? Is the confusion around "in" `HOST_CONTEXT`?

host/sgx/windows/host_context.asm
anakrish: Great.

host/sgx/windows/host_context.asm
anakrish: Ok. Would be good to tweak the comment in line 67 to mention that.

host/sgx/windows/enter.asm
CodeMonkeyLeet: Good idea, I've added definitions for these values for readability.

host/sgx/windows/enter.asm
CodeMonkeyLeet: Okay, I've updated the comments around the definition of `PARAMS_SPACE` and `HOST_CONTEXT`, let me know if that still doesn't help.

enclave/core/sgx/enter.S
```diff
@@ -119,6 +119,23 @@ oe_enter:
     mov %rsp, %rbp
 
 .call_function:
+    // Set the MXCSR according to the Linux x86_64 ABI
+    mov $ABI_MXCSR_INIT, %r10
+    push %r10
```
anakrish: Just occurred to me that push pop are not needed here and for fldcw
You could do
```asm
mov $ABI_MXCSR_INIT, (%rsp)
ldmxcsr (%rsp)
```

host/sgx/windows/enter.asm
anakrish: Ok. Thanks for clarifying.

host/sgx/windows/enter.asm
```diff
@@ -103,11 +111,22 @@ execute_eenter:
...
-	
+
+    ;; Restore the saved host context
+    mov rcx, HOST_CONTEXT
+    call oe_restore_host_context
+
 dispatch_ocall:
+
+    ;; Save callee-preserved registers
+    push rbx
```
anakrish: Do we need this? Technically the callee (`__oe_dispatch_ocall`) should preserve these registers right?

enclave/core/sgx/enter.S
```diff
@@ -119,6 +119,23 @@ oe_enter:
...
+    // Set the FPU Control Word according to the Linux x86_64 ABI
+    mov $ABI_FPUCW_INIT, %r10
+    push %r10
+    fldcw (%rsp)
+    pop %r10
+
+    // Initialize the RFLAGS prior to calling enclave functions
+    // This only clears the DF and state flag bits since
+    // the system flags and reserved bits are not writable here
+    push $0
```
jxyang: Do we need to save and restore RFLAGS register on the host side?

CodeMonkeyLeet: Yes, it's cleaned out as part of `oe_cleanup_registers` on the enclave side on enclave exit, called as a part of an ocall or return from ecall. AEX handles the context cleanup in the exception case.

CodeMonkeyLeet: To answer the "do we need to _save_" part of the question, no we don't. The RFLAGS are divided into:
* _Status flags_: regarded as volatile in both Linux and Windows ABIs
* _DF_: required to be cleared by both Linux and Windows ABIs
* _System Flags and IOPL Field_: privileged bits set by the OS, so we don't actually clear these when calling `popf` from user mode.

host/sgx/windows/enter.asm
CodeMonkeyLeet: No, you're right. It turns out the pushes were just masking a stack alignment bug prior to the `__oe_dispatch_ocall`.

I've clarified the alignment of the stack at the beginning of the function and corrected the offset for this pseudo-prologue for the call into `__oe_dispatch_ocall` to maintain that alignment.

enclave/core/sgx/enter.S
CodeMonkeyLeet: Thanks for the suggestion, but I think it's clearer to keep the two inits apart rather than save a push by reusing `(%rsp)`

host/sgx/windows/enter.asm
```diff
@@ -103,10 +111,11 @@ execute_eenter:
...
     mov ARG1OUT, rdi
     mov ARG2OUT, rsi
-	
-    ;; Restore the saved MXCSR
-    ldmxcsr MXCSR
-	
+
+    ;; Restore the saved host context
+    mov rcx, HOST_CONTEXT
+    call oe_restore_host_context
```
anakrish: Great work! Looks a lot cleaner.

Reading this new version, it looks like only the `execute_eenter` block needs to be in assembly.
The rest of this function can be written in C. Something we can attempt when there is bandwidth next release.

CodeMonkeyLeet: I agree, we should consider factoring out more of this into higher level code instead of assembly in the next milestone.

PR #1826, opened 2019-05-24
Enable tests/crypto/enclave on Windows

CodeMonkeyLeet:
- Enable tests/crypto/enclave on Windows

    - Enable building tests/crypto/enclave folder on Windows.

    - Remove f_readv in crypto.edl, which cannot compile on Windows.

    - Refactor SYS_readv syscall hook in crypto/enclave/enc/enc.c to
      call f_read instead of f_readv since it was already using a
      synchronous single-buffer design for the ocall.

    - Update f_* methods in crypto/enclave/host/host/c to open and
      read files as binary on Windows.

    - Fix tests/crypto/enclave/CMakeLists.txt to use add_enclave_test
      instead of add_test.

    - Fix add_enclave_test.cmake to support nested test paths when
      using the ADD_WINDOWS_ENCLAVE_TESTS flag.

- Refactor tests/crypto/data build for Windows

    - Move key and certificate generation actions into dedicated bash script
      in tests/crypto/data so that it can execute in the git bash environment
      during Windows build.

    - Build tests/crypto/data as a separate CMake step to remove implicit
      dependency of enclave tests on build of the host tests.

    - Standardize naming conventions for crypto test data.

- Refactor tests/crypto_crls_cert_chains/data build for Windows

    - Move key and certificate generation actions into dedicated bash script
      in tests/crypto_crls_cert_chains/data so that it can execute in the git
      bash environment during Windows build.

    - Build tests/crypto/data as a separate CMake step to remove implicit
      dependency of enclave tests on build of the host tests.

    - Standardize naming conventions for crypto_crls_cert_chains test data.

PR #2226, opened 2019-10-02
Fix incremental rebuild due to `add_custom_target`

andschwa:
No file dependencies were setup, so repeated invocations of the build
system would cause these keys to be regenerated each time.

This bug was introduced in #2064, which I found via a long `git bisect` session.

### DIFF
--- a/tests/crypto/data/CMakeLists.txt
+++ b/tests/crypto/data/CMakeLists.txt
@@ -10,33 +10,35 @@ else()
     message(FATAL_ERROR "Unknown OS. Only supported OSes are Linux and Windows")
 endif()
 
-add_custom_target(
-    crypto_test_data ALL
+# These are specifically the byproducts that will be directly consumed by crypto tests
+set(CRYPTO_TEST_DATA
+    asn1.cert.pem
+    coordinates.bin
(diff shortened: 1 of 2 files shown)


CodeMonkeyLeet:
@andschwa Thanks for the fix, reducing build times is always nice. I'm not sure how you concluded this was related to #2064 based on the bisect though? That PR doesn't change the behavior of the build process ... I think you meant #1826, which is the changeset that introduced the build of the tests/crypto/data folder as its own target.

Specifically in that PR, this was brought up in review and we didn't resolve the [question](https://github.com/openenclave/openenclave/pull/1826/files#r288801810) of replacing `add_custom_target` with `add_custom_command`. This changeset does a good job of illustrating how to compose the two, which is educational.

andschwa:
@CodeMonkeyLeet Oops GitHub lied to me. Specifically the commit that introduced the bug was a34d4771d9acb16cd1398707931d437fa79ce692, and I didn't double check which PR that came in through 😅 and yeah sorry I should have been more clear in that review, it was indeed about to introduce a bug 🐛 but we squashed it no worries.

PR #2229, opened 2019-10-03
Add a release manager selection process to Releasing.md

johnkord:
This adds a stop-gap process for nominating and selecting a release manager to unblock us for the upcoming release.

I chose 3 days for discussion as a minimum acceptable time period. We could increase that to 7 days (or more) if people wish!

### DIFF
--- a/docs/Releasing.md
+++ b/docs/Releasing.md
@@ -4,15 +4,19 @@ Open Enclave SDK Release Procedures
 This document covers how we handle creating a release. Let's look at it from an
 example:
 
+Selecting a Release Manager
+---------------------------
+
+The Community Governance Committee will select a release manager from one of the
+Committers by submitting a Pull Request that adds a "Release" tag to that
+Committer's "Area" in [Committers.md](Committers.md).
+
 Version Bump
 ------------
 
 > A note on naming: please prepend the `v` for consistency. Tags are `v0.7.0`
 > and branches are `v0.7.x`. We follow [Semantic
 > Versioning](https://semver.org/spec/v2.0.0.html).
-
-We want to release `v0.7.0` in about a week, so we choose a release manager from
-the committers.
 
 The initial announcement of the upcoming release will be a PR by the release
 manager to the `master` branch with the commit to bump the [VERSION


docs/Releasing.md
```diff
@@ -4,16 +4,38 @@ Open Enclave SDK Release Procedures
...
+---------------------------
+
+A committer can be nominated to become the release manager by drafting a pull
+request that sets the "Release" tag to the "Area" of one of the committers in
+[Committers.md](Committers.md). The submitter of that nomination should then
+send a link to that pull request through email to all of the committers for open
+discussion. This discussion & PR should be open for at least 3 days, and people
+can make their votes known in the pull request itself by either approving the
+pull request or requesting changes. If there is any opposition from the
+committers after 3 days, then the Committee Governance Committee should take up
```
andschwa: s/Committee Governance Committee/Community Governance Committee

docs/Releasing.md
```diff
@@ -4,16 +4,38 @@ Open Enclave SDK Release Procedures
...
+this issue as high priority over email and decide whether to accept or reject
```
andschwa: Email? Or the GitHub issue? I really don't want us splintering into yet more tools.

docs/Releasing.md
```diff
@@ -4,16 +4,38 @@ Open Enclave SDK Release Procedures
...
+tag to this committer gets merged to master. Only committers can nominate other
+committers to be the release manager. Only one committer can be tagged with the
+Release tag in [Committers.md](Committers.md).
+
 Version Bump
 ------------
 
+*Note*: The versioning information below is not fully applicable until it can be
+updated. There were a few issues with both Debian and NuGet packages regarding
+the v0.7.x naming scheme. This process will be changing soon. The release branch
```
andschwa: That naming scheme shouldn't ever make it into a package. The `x` explicitly denotes that it's an unreleased branch (i.e. _not_ a version). If a package is being generated, it should go with whatever the current patch is (I guess 0 if there is no v0.7 release) and then potentially iterate with RC suffixes.

andschwa: I personally don't like having a version set in the VERSION file before that version actually exists...but I'm not sure of another way to do this.

docs/Releasing.md
johnkord: Email is a private forum for the committee members to discuss an urgent issue. We could also call for a short-notice CGC meeting, or put it off until the next CGC meeting.

docs/Releasing.md
johnkord: The problem with using `x` will surface if you build a package on the master branch. It simply will not function with NuGet packages. Moreover, 0.7.x is considered a greater version than 0.7.0 for debian packages, and that is arguably even more broken because you won't even be aware that a package is available for upgrade when 0.7.0 is published to an apt repo. It's just not a good versioning scheme.

I proposed a possible different way of structuring the VERSION file in #2204: 

```
Just rebased this branch on master. I think I've addressed all comments in this PR. 
There are a few open comments to welcome discussion. Specifically, the VERSION file and how NuGet handles it is a little awkward.
I think we may wish to redesign the VERSION file to be something like:

OE_MAJOR=0
OE_MINOR=7
OE_PATCH=0
OE_PRERELEASE=alpha
OE_BUILD=1

as per https://semver.org/

Then, we can parse this file to produce a package's version number, depending on the type of package.
Different package formats (RPM, DEB, NuGet, etc) expect different version strings, and this kind of VERSION file makes it easier to parse for translation into a full version string for a given package format.
I'd like to address this in a future design doc.
```

johnkord: We probably don't want to get into the situation where we're updating this version file with every commit, because that produces a lot of commit message churn (and it was very annoying). But at least with this kind of situation we have a prerelease variable we can set to anything we want it to be. In the master branch it could be `dev`, and in the v0.7.x branch we could update that variable for `rc` or something, until its actually released, and then change it to `release` or something.

docs/Releasing.md
```diff
@@ -4,16 +4,38 @@ Open Enclave SDK Release Procedures
...
+send a link to that pull request through email to all of the committers for open
+discussion. This discussion & PR should be open for at least 3 days, and people
+can make their votes known in the pull request itself by either approving the
+pull request or requesting changes. If there is any opposition from the
+committers after 3 days, then the Committee Governance Committee should take up
+this issue as high priority over email and decide whether to accept or reject
+the new release manager based on feedback from the other committers. If there is
+no opposition from the committers after three days, the newly nominated
+committer becomes the release manager and the pull request to add the "Release"
+tag to this committer gets merged to master. Only committers can nominate other
```
BRMcLaren: This implies no self-nomination and if there is a voting process in place for opposition, it might be unnecessary.

johnkord: Ah thank you. Let me remove "other" here :)

docs/Releasing.md
```diff
@@ -4,16 +4,38 @@ Open Enclave SDK Release Procedures
...
+Selecting a Release Manager
+---------------------------
+
+A committer can be nominated to become the release manager by drafting a pull
+request that sets the "Release" tag to the "Area" of one of the committers in
+[Committers.md](Committers.md). The submitter of that nomination should then
+send a link to that pull request through email to all of the committers for open
+discussion. This discussion & PR should be open for at least 3 days, and people
+can make their votes known in the pull request itself by either approving the
+pull request or requesting changes. If there is any opposition from the
```
dthaler: I think there's one bug in the process proposed, which is that it doesn't talk about what happens if there's two proposals.   For example, what happens if there's two different PR's with different candidates, and no one opposes either one, just each has different people voting for them.   I think that's another case where the CGC should take up the issue.

And if there are multiple competing candidates (e.g., candidates from different companies), should the identity of who votes for whom be public or private?   For decisions that affect personnel, there are good arguments for keeping votes private.

I have no objection to the current text as interim process, but I think all of the above considerations need to be taken into account for the non-interim process.

johnkord: Yes, I agree. In the future, we may wish to call for an election among the committers with a deadline to get your votes in, or something along those lines.

johnkord:
Attention: @andschwa @dthaler @achamayou @mikbras @gupta-ak @CodeMonkeyLeet @anakrish
Could you please review this PR and either sign off or request changes? We need this process to be in place for me to merge the release manager assignment PR (which does have plenty of approvals), so I can draft a release branch and move forward with the release manager duties.

Thank you!

johnkord:
I have taken it upon myself to work as the acting-release-manager until this process PR gets merged. Please review when you have time, Committee Members ( @openenclave/committee ) !

docs/Releasing.md
CodeMonkeyLeet: I have a lot of questions as to whether this needs to be documented as a special process:
- Why is the process of selecting a Release Manager not just the standard process of the CGC making that an agenda item with the usual standard criteria for voting? 
    - I see the Release Manager as an operational role that the CGC should decide, not something that all committers elect.
    - As in the case of suggesting new committers, all committers are free to suggest a change in the Release Manager, and it becomes a standard order of business for the CGC to take up and decide on.
        - Is the codification of discussion duration etc. more broadly applicable to how the CGC makes decisions? Or is electing a Release Manager intended to be a higher bar for some reason?

Specifically, the heightened process seems to be an imply that the Release Manager serves some kind of elevated role, which is probably something we do not want to endow to an individual. I expect that a Release Manager is not a gatekeeper, just a coordinator for a release, with triage authority ultimately rooting to a technical steering committee of some kind.

I think it's detrimental to invest in a heavyweight process of electing a Release Manager such that it makes backups or handing off the role difficult. e.g. Release Manager gets run over by a bus the day on a day we announce a CVE and a patch update. I want the role to be sufficiently transferrable that someone can quickly step up, read the doc and get the release back on track.

johnkord: @dthaler @CodeMonkeyLeet I'm good with the idea that the CGC decides on the release manager, instead of calling for a vote of the committers. It really does simplify process, when it's not necessary to complicate it. I will update this document accordingly

johnkord:
I've changed the process here to fall in line with Simon's suggestion, to let the Committee Members select the release manager.

johnkord:
Thanks for the reviews, everyone. Friendly ping to @achamayou @mikbras @anakrish for reviews :)

johnkord:
Thanks everyone!

bors r+

PR #2230, opened 2019-10-03
Set John Kordich as the Release Manager.

johnkord:
This is the PR to nominate making myself (John Kordich) the release manager. It is based on the newly submitted guidelines for selecting a release manager in #2229


andschwa:
Added all committers as reviewers (I think this works?)

johnkord:
@andschwa I tried that too up above when I created this PR! Not sure if it worked or not :)

BRMcLaren:
Just here to say, that the adding Commitee as reviewers thing, did not work from the above attempts :) 

I added OpenEnclave/Committee and it seems to work on my end, as it shows it now?

andschwa:
I think it goes away when somebody from that group approves...GitHub is weird.

johnkord:
Thanks everyone!

bors r+

PR #2238, opened 2019-10-08
Add nuget and hostverify package builds to Packaging.Jenkinsfile.

johnkord:
This PR adds NuGet and hostverify packages to the nightly build.

Tested this in the sandbox for packaging job: https://oe-jenkins.eastus.cloudapp.azure.com/blue/organizations/jenkins/OpenEnclave-Sandbox-Packaging/detail/OpenEnclave-Sandbox-Packaging/8/pipeline

### DIFF
--- a/.jenkins/Packaging.Jenkinsfile
+++ b/.jenkins/Packaging.Jenkinsfile
@@ -14,7 +14,8 @@ def packageUpload(String version, String build_type) {
                            cmake ${WORKSPACE} -DCMAKE_BUILD_TYPE=${build_type} -DCMAKE_INSTALL_PREFIX:PATH='/opt/openenclave' -DCPACK_GENERATOR=DEB
                            make
                            ctest --output-on-failure --timeout ${CTEST_TIMEOUT_SECONDS}
-                           make package
+                           cpack
+                           cpack -D CPACK_DEB_COMPONENT_INSTALL=ON -DCPACK_COMPONENTS_ALL=OEHOSTVERIFY
                            """
                 oe.Run("clang-7", task)
                 azureUpload(storageCredentialId: 'oe_jenkins_storage_account', filesPath: 'build/*.deb', storageType: 'blobstorage', virtualPath: "master/${BUILD_NUMBER}/ubuntu/${version}/${build_type}/SGX1FLC/", containerName: 'oejenkins')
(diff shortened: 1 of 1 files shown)


.jenkins/Packaging.Jenkinsfile
```diff
@@ -14,7 +14,8 @@ def packageUpload(String version, String build_type) {
                            cmake ${WORKSPACE} -DCMAKE_BUILD_TYPE=${build_type} -DCMAKE_INSTALL_PREFIX:PATH='/opt/openenclave' -DCPACK_GENERATOR=DEB
                            make
                            ctest --output-on-failure --timeout ${CTEST_TIMEOUT_SECONDS}
-                           make package
+                           cpack
```
jazzybluesea: Is this function suppose to be for Linux? Maybe it will make it clearer if this function is renamed LinuxUpload?  It would also help if there is a comment for this function describing what it is doing.

.jenkins/Packaging.Jenkinsfile
```diff
@@ -26,17 +27,22 @@ def packageUpload(String version, String build_type) {
 
 def WindowsUpload() {
```
jazzybluesea: Please add a comment for this function.

.jenkins/Packaging.Jenkinsfile
```diff
@@ -14,7 +14,8 @@ def packageUpload(String version, String build_type) {
                            cmake ${WORKSPACE} -DCMAKE_BUILD_TYPE=${build_type} -DCMAKE_INSTALL_PREFIX:PATH='/opt/openenclave' -DCPACK_GENERATOR=DEB
                            make
                            ctest --output-on-failure --timeout ${CTEST_TIMEOUT_SECONDS}
-                           make package
+                           cpack
+                           cpack -D CPACK_DEB_COMPONENT_INSTALL=ON -DCPACK_COMPONENTS_ALL=OEHOSTVERIFY
                            """
                 oe.Run("clang-7", task)
                 azureUpload(storageCredentialId: 'oe_jenkins_storage_account', filesPath: 'build/*.deb', storageType: 'blobstorage', virtualPath: "master/${BUILD_NUMBER}/ubuntu/${version}/${build_type}/SGX1FLC/", containerName: 'oejenkins')
```
jazzybluesea: Why is this package uploaded under SGX1FLC?

.jenkins/Packaging.Jenkinsfile
```diff
@@ -26,17 +27,22 @@ def packageUpload(String version, String build_type) {
 
 def WindowsUpload() {
     stage('Windows Release') {
```
jazzybluesea: How about windows debug version?

.jenkins/Packaging.Jenkinsfile
johnkord: Yes totally agree, thanks @jazzybluesea!

.jenkins/Packaging.Jenkinsfile
johnkord: Technically, the package was built with USE_LIBSGX=ON, so we upload those under SGX1FLC build directories. It's a little confusing, and the USE_LIBSGX flag will probably be changing to something like USE_DCAP in the future, so we'll rename these directories to something like SGX1DCAP, maybe? A lot of the naming is awkward right now.

.jenkins/Packaging.Jenkinsfile
johnkord: Agree, should add!

.jenkins/Packaging.Jenkinsfile
johnkord: For now I will decline to add comments for functions, as we haven't done that for our Jenkinsfiles yet. I think we should add some documentation for all of our CI system test cases, though. I know @radhikaj is working on some documentation for supported platforms & configurations.

johnkord:
Thanks all!
bors r+

PR #2244, opened 2019-10-08
Enable the switchless sample on Windows

jxyang:


jxyang:
bors try

tests/switchless_threads/enc/enc.c
```diff
@@ -70,6 +70,6 @@ OE_SET_ENCLAVE_SGX(
     1,    /* ProductID */
     1,    /* SecurityVersion */
     true, /* AllowDebug */
-    1024, /* HeapPageCount */
-    1024, /* StackPageCount */
+    128,  /* HeapPageCount */
```
radhikaj: @jxyang , why did this need to be changed?

jxyang: This test fails on Windows 2-core machines because it goes beyond the EPC limit (62mb?)

PR #2221, opened 2019-10-01
Replace OCPWin and OCaml with esy

andschwa:
This replaces the CMake-driven OCaml build with [`esy`](https://esy.sh/) and `dune`, which means that our dependencies drop down to just installing `esy` via `npm`, and it builds a native `oeedger8r` executable across all our platforms.

Rough edges:
- [x] It builds in the source tree instead of the build tree, grr.
- [x] We need `npm install -g esy` done to our CI systems.
- [x] Update Ansible scripts.
- [x] Update Windows install script.
- [x] Still need to update the CI containers.
- [x] I need to setup the correct `.gitignore` file and commit the `esy.lock` directory to vendor our dependencies.

Gains: no more OCPWin! Modern OCaml! OCaml packages! Dune build! (And actually, no `npm` requirement on Linux, only on Windows, and only for now.)

PR #2246, opened 2019-10-09
Ansible and PowerShell changes to install esy

andschwa:
This was spun out from #2221 in order to make updating CI easier.


johnkord:
Actually Andy can we hold off on this for a bit? I realized that if we check in the .ps1 changes before the other esy changes to the build, people that use the ps1 for deploying a system will be broken because they won't have the ocpwin binaries.

bors r-

andschwa:
Will probably close this if my other attempt to fix the chicken egg problem works out...

scripts/ansible/roles/linux/openenclave/tasks/validation.yml
```diff
@@ -37,9 +37,9 @@
   register: clang_check_version
   failed_when: clang_check_version.stdout.find(clang_target_version) == -1
 
-- name: Ocaml version check
-  shell: "ocaml -version"
+- name: Esy version check
```
CodeMonkeyLeet: Should this continue to check for `ocaml` since tools/oeedger8r/CMakeLists.txt still calls directly into `ocamlopt` rather than esy? (i.e. is the implication that esy presence will always equate to ocaml presence moving forwards?)

scripts/ansible/oe-windows-acc-setup.yml
```diff
@@ -20,8 +20,8 @@
         seven_zip_hash:      "F00E1588ED54DDF633D8652EB89D0A8F95BD80CCCFC3EED362D81927BEC05AA5"
         vs_buildtools:       "https://oejenkins.blob.core.windows.net/oejenkins/vs_buildtools_2017.exe"
         vs_buildtools_hash:  "6F49872B04A0EAEDF5ED96AB25F7697062D81A419D45D9970E41784F31165BF2"
-        ocaml_url:           "https://oejenkins.blob.core.windows.net/oejenkins/ocpwin64-20160113-4.02.1+ocp1-mingw64.zip"
-        ocaml_hash:          "369F900F7CDA543ABF674520ED6004CC75008E10BEED0D34845E8A42866D0F3A"
+        node_url:            "https://nodejs.org/dist/v10.16.3/node-v10.16.3-x64.msi"
```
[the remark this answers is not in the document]
andschwa: We're tracking that in #2221.

scripts/ansible/roles/linux/openenclave/tasks/validation.yml
andschwa: No, as this is really just part of #2221 and was an attempt to solve a chicken/egg problem with updating dependencies. I am going to close this PR as John and I are working to solve updating the CI machines outside of this.

PR #2248, opened 2019-10-10
Update CMakeSettings.json with NUGET_PACKAGE_PATH

CodeMonkeyLeet:
Commit c3c6dc17 added a requirement to define the `NUGET_PACKAGE_PATH` variable
in CMake for WIN32 builds. This commit updates the CMakeSettings.json file for
Visual Studio users to provide the `NUGET_PACKAGE_PATH` parameter based on the
install path used when running install-windows-prereqs.ps1.

### DIFF
--- a/CMakeSettings.json
+++ b/CMakeSettings.json
@@ -31,7 +31,7 @@
       "inheritEnvironments": [ "msvc_x64_x64" ],
       "buildRoot": "${workspaceRoot}\\build\\x64-Debug",
       "installRoot": "${env.USERPROFILE}\\CMakeBuilds\\${workspaceHash}\\install\\${name}",
-      "cmakeCommandArgs": "-DBUILD_ENCLAVES=1 -DUSE_LIBSGX=1",
+      "cmakeCommandArgs": "-DBUILD_ENCLAVES=ON -DUSE_LIBSGX=ON -DNUGET_PACKAGE_PATH=${workspaceRoot}\\prereqs\\nuget",
       "buildCommandArgs": "-v",
       "ctestCommandArgs": ""
     },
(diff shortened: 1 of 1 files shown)


CMakeSettings.json
```diff
@@ -31,7 +31,7 @@
       "inheritEnvironments": [ "msvc_x64_x64" ],
       "buildRoot": "${workspaceRoot}\\build\\x64-Debug",
       "installRoot": "${env.USERPROFILE}\\CMakeBuilds\\${workspaceHash}\\install\\${name}",
-      "cmakeCommandArgs": "-DBUILD_ENCLAVES=1 -DUSE_LIBSGX=1",
+      "cmakeCommandArgs": "-DBUILD_ENCLAVES=1 -DUSE_LIBSGX=1 -DNUGET_PACKAGE_PATH=${workspaceRoot}\\prereqs\\nuget",
```
johnkord: This presumes the location of those installed nuget packages, which could be anywhere now! I'm okay with this as long as people understand they can change this field in the json file :)

Oh and I just noticed, you might want to change `-DUSE_LIBSGX=1` to `-DUSE_LIBSGX=ON`. It really doesn't matter much I don't think, does it @andschwa ?

jxyang: It's the same thing I think.

CodeMonkeyLeet: @johnkord Yes, the path can point anywhere and it will be explained in the documentation (WindowsSGX1GettingStarted.md/WindowsSGX1FLCGettingStarted.md) when we start documenting the Visual Studio integration use cases. This PR simply fixes the regression in existing behavior, selecting a reasonable default.

CMake [treats a whole bunch of constants as `TRUE`](https://cmake.org/cmake/help/v3.16/command/if.html), so functionally it doesn't matter. If we have a stylistic preference we should probably document it somewhere.

PR #2249, opened 2019-10-11
The Open Enclave Shared Memory Model

mikbras:
This document proposes a shared memory model for sharing untrusted memory with enclaves. The goal is build applications once that will run on both SGX and TrustZone (and other future platforms).

### DIFF
--- a/docs/DesignDocs/shared_memory.md
+++ b/docs/DesignDocs/shared_memory.md
@@ -18,7 +18,7 @@ is shared).
 This problem is partly addressed by the PR entitled [Enforcing full
 serialization in EDL](https://github.com/openenclave/openenclave/pull/2176),
 which proposes extending **edger8r** to print warnings for EDL specifications
-with parameters cannot be fully serialized. These specifications produce
+with parameters that cannot be fully serialized. Such specifications produce
 applications that implicitly share untrusted memory with the enclave. Aside
 from introducing potential security flaws, these applications are not portable
 to TrustZone and any future platforms that employ a similar model. By forcing
@@ -176,7 +176,7 @@ parameter.
 Assumptions:
 
 - All shared memory is read-write (non-executable). No mechanism is provided
-  for allocating read-only memory ore executable memory.
+  for allocating read-only memory or executable memory.
 
 - Shared TrustZone memory is visible at the same address on both the host and
   the enclave (hence all shared memory is zero-copy).
--- a/docs/DesignDocs/shared_memory.md
+++ b/docs/DesignDocs/shared_memory.md
@@ -1,0 +1,190 @@
+The Open Enclave shared memory model
+====================================
+
+This document proposes a *shared memory model* that will improve
+interoperability across different secure hardware platforms, initially
+including SGX and TrustZone.
+
+Motivation
+----------
+
+Ideally, applications developed with the Open Enclave API should run on all
+supported platforms. But unfortunately, the platforms themselves may have
+incompatible shared memory models. For example, SGX implicity shares all
+untrusted memory with the enclave, whereas TrustZone shares untrusted memory
+explicitly through programmatic measures.
+
+This problem is partly addressed by the PR entitled [Enforcing full
+serialization in EDL](https://github.com/openenclave/openenclave/pull/2176),
+which proposes **edger8r** options that reject EDL specifications that cannot
+be fully serialized. Such specifications produce applications that implicitly
+share untrusted memory with the enclave. Aside from introducing potential
+security flaws, these applications are not portable to TrustZone and any future
+platforms that employ a similar model. By forcing full serialization, the need
+for shared memory can be mostly eliminated. But sometimes applications need to
+pass untrusted shared memory to enclaves. A few of examples include:
+
+- Ring buffers
+- Context-switchless calls
+- Efficient transfer of large memory objects
+
+Whatever the reasons, a mechanism is needed for sharing untrusted memory with
+the enclave and that mechanism must be interoperable across secure hardware
+platforms. Further, the mechanism must be explicit and operations are needed
+for managing untrusted memory.
+
+User Experience
+---------------
+
+The host may share untrusted memory with the enclave through the use of the
+**user_check** EDL attribute. For example:
+
+```c++
+enclave
+{
+    trusted
+    {
+        void share_buffer_ecall(
+            [user_check] void* untrusted_memory,
+            size_t untrusted_memory_size);
+    };
+};
+```
+
+We assume in this example that the enclave needs to read or write the
+**untrusted_memory** buffer. Prior to this proposal, SGX applications
+could pass any pointer to untrusted memory. For example:
+
+```c++
+    static unsigned char buf[4096];
+
+    share_buffer_ecall(buf, sizeof(buf));
+```
+
+But this application is not portable to TrustZone since **buf** is not visible
+within the enclave. To make this work on TrustZone, the buffer should be
+explicitly allocated from a shared memory pool as follows.
+
+```c++
+    const size_t n = 4096;
+    unsigned char* buf;
+
+    if (oe_allocate_shared_memory(enclave, n, &buf) != OE_RESULT_OK)
+    {
+        /* Error */
+    }
+
+    memset(buf, 0, n);
+
+    share_buffer_ecall(buf, n);
+```
+
+This program fragment will work on both SGX and TrustZone. The application
+will eventually free this buffer by passing it to **oe_free_shared_memory()**.
+
+The enclave's implementation of **share_buffer_ecall()** should verify that
+the **untrusted_memory** is fully contained within shared memory as shown
+below.
+
+```c++
+void share_buffer_ecall(void* untrusted_memory, size_t untrusted_memory_size)
+{
+    if (!oe_is_shared_memory(untrusted_memory, untrusted_memory_size)
+    {
+        /* Error! */
+    }
+    ...
+}
+```
+
+Specification
+-------------
+
+This design introduces the following functions on the host side:
+
+```c++
+    oe_result_t oe_allocate_shared_memory(
+        oe_enclave_t* enclave,
+        size_t size,
+        void** ptr);
+
+    oe_result_t oe_free_shared_memory(
+        oe_enclave_t* enclave,
+        void* ptr);
+
+    bool oe_is_shared_memory(
+        oe_enclave_t* enclave,
+        const void* ptr,
+        size_t size);
+
+    typedef enum _oe_memory_type
+    {
+        /* !oe_is_trusted_memory() && !oe_is_shared_memory() */
+        OE_MEMORY_TYPE_UNKNOWN,
+
+        /* oe_is_trusted_memory() */
+        OE_MEMORY_TYPE_TRUSTED,
+
+        /* oe_is_shared_memory() */
+        OE_MEMORY_TYPE_SHARED,
+    }
+    oe_memory_type_t;
+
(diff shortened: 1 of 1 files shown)


docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+        /* Error */
+    }
+
+    memset(buf, 0, n);
+
+    share_buffer_ecall(buf, n);
+```
+
+This program fragment will work on both SGX and TrustZone. The application
+will eventually free this buffer by passing it to **oe_free_shared_memory()**.
```
johnkord: Is there an expectation that the host-side oe_allocate_shared_memory function is basically a malloc when compiled for SGX processors? Or does it do anything else on SGX processors? :)

docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+
+But this application is not portable to TrustZone since **buf** is not visible
+within the enclave. To make this work on TrustZone, the buffer should be
+explicitly allocated from a shared memory pool as follows.
+
+```c++
+    const size_t n = 4096;
+    unsigned char* buf;
+
+    if (oe_allocate_shared_memory(enclave, n, &buf) != OE_RESULT_OK)
```
jxyang: By passing an `enclave` parameter, do we intend to keep the allocated shared memory only accessible by host and **this** enclave?

docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+        const void* ptr,
+        size_t size);
+
+    typedef enum _oe_memory_type
+    {
+        /* !oe_is_within_enclave() && !oe_is_shared_memory() */
+        OE_MEMORY_TYPE_UNKNOWN,
+
+        /* oe_is_within_enclave() */
+        OE_MEMORY_TYPE_TRUSTED,
```
jxyang: Why do we need this type? I assume the memory we get back from `oe_allocate_shared_memory` is always of type `OE_MEMORY_TYPE_SHARED`?

docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+
+        /* oe_is_shared_memory() */
+        OE_MEMORY_TYPE_SHARED,
+    }
+    oe_memory_type_t;
+
+    oe_result_t oe_get_memory_type(
```
jxyang: What's the usage of this function?

docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+        oe_enclave_t* enclave,
+        const void* ptr,
+        size_t size,
+        oe_memory_type_t* type);
+```
+
+The same functions are available on the enclave side but without the enclave
+parameter.
```
jxyang: Probably worth mentioning the enclave side of the APIs are mostly for parameter passing of OCALLs.

docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+    oe_result_t oe_get_memory_type(
+        const void* ptr,
+        size_t size,
+        oe_memory_type_t* type);
+```
+
+Assumptions:
+
+- All shared memory is read-write (non-executable). No mechanism is provided
+  for allocating read-only memory or executable memory.
```
jxyang: This is probably a `nice to have` in the future. An enclave function that writes to the shared memory deserves more security scrutinies than a function that just read it.

docs/DesignDocs/shared_memory.md
jxyang: I think so. But on the enclave side, we probably need a custom allocator to improve performance (avoid ocall at least).

docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+incompatible shared memory models. For example, SGX implicit shares all
+untrusted memory with the enclave, whereas TrustZone shares untrusted memory
+explicitly through programmatic measures (by default, no untrusted memory
+is shared).
+
+This problem is partly addressed by the PR entitled [Enforcing full
+serialization in EDL](https://github.com/openenclave/openenclave/pull/2176),
+which proposes extending **edger8r** to print warnings for EDL specifications
+with parameters that cannot be fully serialized. Such specifications produce
+applications that implicitly share untrusted memory with the enclave. Aside
```
dthaler: "implicitly share untrusted memory" -> "assume that untrusted memory can be implicitly shared"

docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+from introducing potential security flaws, these applications are not portable
+to TrustZone and any future platforms that employ a similar model. By forcing
+full serialization, the need for shared memory can be mostly eliminated. But
+some applications need to pass untrusted shared memory to enclaves. Examples
+include:
+
+- Ring buffers
```
dthaler: Why do ring buffers need shared memory?   If they fall under "efficient transfer of large memory objects", then I would agree but that's already in the list and we don't need duplication so it could be removed.  If they don't, then having a slightly longer explanation would help.

docs/DesignDocs/shared_memory.md
dthaler: Is buf guaranteed to be initialized if failure is returned?  Since there's no return; in line 76, the sample code will fall through into the memset in line 79 causing a crash if not.   You probably want a return; after line 76.

docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+    oe_result_t oe_allocate_shared_memory(
+        oe_enclave_t* enclave,
+        size_t size,
+        void** ptr);
+
+    oe_result_t oe_free_shared_memory(
+        oe_enclave_t* enclave,
+        void* ptr);
+
+    bool oe_is_shared_memory(
```
dthaler: As discussed during a previous meeting, "is shared" is not sufficient for actual use, since it may be read-only or read-write.  If it's read-only then testing for "is shared" is not sufficient to tell whether it's writable.  Instead, we discussed that this should be replaced by an API that either tests for specific access rights (read, write, execute), or else returns the set of access rights available.

docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+        const void* ptr,
+        size_t size,
+        oe_memory_type_t* type);
+```
+
+The same functions are available on the enclave side but without the enclave
+parameter.
+
+```c++
+    oe_result_t oe_allocate_shared_memory(
```
jazzybluesea: Any reason why there is no realloc?

docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+applications that implicitly share untrusted memory with the enclave. Aside
+from introducing potential security flaws, these applications are not portable
+to TrustZone and any future platforms that employ a similar model. By forcing
+full serialization, the need for shared memory can be mostly eliminated. But
+some applications need to pass untrusted shared memory to enclaves. Examples
+include:
+
+- Ring buffers
+- Context-switchless calls
+- Efficient transfer of large memory objects
```
anakrish: Shared memory increases risk of [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) vulnerabilities. While the enclave is processing the shared the object, a malicious host could change the contents of the shared memory object, potentially causing security vulnerabilities. A common strategy to prevent TOCTOU issues is to copy the object to enclave memory, operate on it and then to copy it (results) out back to host memory.

It is not clear how shared-memory would help with efficient transfer of large memory objects since for secure coding, these objects must be copied to enclave memory.

docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+
+Whatever the reasons, a mechanism is needed for sharing untrusted memory with
+the enclave and that mechanism must be interoperable across secure hardware
+platforms. Further, the mechanism must be explicit, so operations are needed
```
anakrish: Worth noting that we don't yet guarantee that the mechanism is explicit for SGX enclaves.

docs/DesignDocs/shared_memory.md
anakrish: > By passing an enclave parameter, do we intend to keep the allocated shared memory only accessible by host and this enclave?

Great question.

docs/DesignDocs/shared_memory.md
anakrish: Worth defining the behavior when this buffer is passed to standard memory management functions`free`, `realloc` etc.

docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+will eventually free this buffer by passing it to **oe_free_shared_memory()**.
+
+The enclave's implementation of **share_buffer_ecall()** should verify that
+the **untrusted_memory** is fully contained within shared memory as shown
+below.
+
+```c++
+void share_buffer_ecall(void* untrusted_memory, size_t untrusted_memory_size)
+{
+    if (!oe_is_shared_memory(untrusted_memory, untrusted_memory_size)
```
anakrish: Worth defining the behavior if the enclave tries to access shared memory that has been freed by the host.

docs/DesignDocs/shared_memory.md
```diff
@@ -0,0 +1,192 @@
...
+    {
+        /* !oe_is_within_enclave() && !oe_is_shared_memory() */
+        OE_MEMORY_TYPE_UNKNOWN,
+
+        /* oe_is_within_enclave() */
+        OE_MEMORY_TYPE_TRUSTED,
+
+        /* oe_is_shared_memory() */
+        OE_MEMORY_TYPE_SHARED,
+    }
```
anakrish: Worth defining is a portion of memory will always be shared or not.
From what I understand, in SGX 2, an enclave can grow by adding TCS pages after creation. Does that mean that a memory that was once shared can become enclave-memory if it is freed and then rellocated to the enclave?

docs/DesignDocs/shared_memory.md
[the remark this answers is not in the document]
anakrish: @HernanGatta In GlobalPlatform, is it possible for two enclaves to share memory between each other?

docs/DesignDocs/shared_memory.md
dthaler: Some use cases are read-once use cases (no check) and so are not susceptible to TOCTOU.   Read-once use cases can be done without copying.   I agree it would be helpful to elaborate on this point with the issues you point out.

docs/DesignDocs/shared_memory.md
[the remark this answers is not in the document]
HernanGatta: Yes.

In GlobalPlatform terminology, a TA (enclave) can act as a client (host-like) to another TA. In this sense, a TA can cause another TA load and can drive it just like a REE host would. The interface is somewhat more limited, but a TA can, in fact, share memory with another TA.

This includes sharing memory that the client TA is sharing with its host in the REE. So the host can share memory with a TA, then that TA can re-share that memory with another TA.

radhikaj:
Thanks @mikbras for this well written design. I am thinking we would also need an API to say whether the platform supports shared memory.

kuanggang-git:

radhikaj:
Closing for now since this was last updated in Oct 2019.

PR #2253, opened 2019-10-12
Design review: Remote attestation collaterals API.

jazzybluesea:
Addresses #2206

### DIFF
--- a/docs/DesignDocs/RemoteAttestationCollaterals.md
+++ b/docs/DesignDocs/RemoteAttestationCollaterals.md
@@ -56,18 +56,18 @@ keys used to protect and authenticate Claims or Results.
 
 #### Attester:
 An Attestation Function that creates Evidence by
-collecting, formatting and protecting (e.g., signing) Claims.  It
+collecting, formatting and protecting (e.g., signing) Claims. It
 presents Evidence to a Verifier using a conveyance mechanism or
 protocol.
 
 #### Verifier:
 An Attestation Function that accepts Evidence from an Attester using
-a conveyance mechanism or protocol.  It also accepts Known-Good-Values
-and Endorsments from an Asserter using a conveyance mechanism or protocol.
+a conveyance mechanism or protocol. It also accepts Known-Good-Values
+and Endorsements from an Asserter using a conveyance mechanism or protocol.
 It verifies the protection mechanisms, parses and appraises Evidence
-according to good-known valid (or known-invalid) Claims and Endorsments.
+according to good-known valid (or known-invalid) Claims and Endorsements.
 It produces Attestation Results that are formatted and protected (e.g.,
-signed).  It presents Attestation Results to a Relying Party using
+signed). It presents Attestation Results to a Relying Party using
 a conveyance mechanism or protocol.
 
 Claims are statements about a particular subject. They consist of
@@ -78,11 +78,11 @@ claim value, which is arbitrary data. Example of claims could be
 #### Asserter:
 An Attestation Function that generates reference Claims
 about both the Attesting Computing Environment and the Attested
-Computing Environment.  The manufacturing and development
+Computing Environment. The manufacturing and development
 processes are presumed to be trustworthy processes.  In other
 words the Asserter is presumed, by a Verifier, to produce valid
-Claims.  The function collects, formats and protects (e.g. signs)
-valid Claims known as Endorsements and Known-Good-Values.  It
+Claims. The function collects, formats and protects (e.g. signs)
+valid Claims known as Endorsements and Known-Good-Values. It
 presents provable Claims to a Verifier using a conveyance
 mechanism or protocol.
 
@@ -90,7 +90,7 @@ mechanism or protocol.
 An Attestation Function that accepts Attestation
 Results from a Verifier using a conveyance mechanism or protocol.
 It assesses Attestation Results protections, parses and assesses
-Attestation Results according to an assessent context (Note:
+Attestation Results according to an assessment context (Note:
 definition of the assessment context is out-of-scope).
 
 
@@ -135,7 +135,7 @@ result = oe_get_evidence(
 ...
 ```
 
-##### Verifier verifies the evidence and endorsements (in the untrusted host or inside an enclave/TEE)
+##### Verifier verifies the evidence and endorsements (in an untrusted host or inside an enclave/TEE)
 ```C
 ...
 
@@ -256,7 +256,13 @@ typedef struct _oe_endorsements_t
     uint8_t buffer[];              ///< Buffer of offsets + data
 
 } oe_endorsements_t;
+```
 
+### Private SGX endorsement definitions
+
+`common/sgx/evidence.h`
+
+```C
 /*! Version of the supported SGX endorsement structures */
 #define OE_SGX_ENDORSEMENTS_VERSION     (1)
 
@@ -281,14 +287,8 @@ typedef enum _oe_sgx_endorsements_fields
     OE_SGX_ENDORSEMENT_FIELD_QE_ID_ISSUER_CHAIN,
     OE_SGX_ENDORSEMENT_FIELD_CREATION_DATETIME,
     OE_SGX_ENDORSEMENT_COUNT
-
 } oe_sgx_endorsements_fields;
-```
 
-### Private SGX endorsement definitions
-
-`common/sgx/evidence.h`
-```C
 /*! \struct oe_sgx_endorsements
  *
  * \brief SGX endorsements structure
@@ -345,10 +345,7 @@ typedef struct _oe_sgx_endorsements_t
      *     The size of creation_datetime.
      */
     oe_sgx_endorsement_item items[OE_SGX_ENDORSEMENT_COUNT];
-
 } oe_sgx_endorsements_t;
-
-
 ```
 
 ### New Public Attestation functions
@@ -475,7 +472,7 @@ Current set of claims definitions:
 ### OE Host Verify Library
 
 The OE Host Verify library is a standalone library used for verifying remote reports outside
-the TEE/enclave. The function `oe_verfiy_remote_report()` will be updated to support
+the TEE/enclave. The function `oe_verify_remote_report()` will be updated to support
 endorsements.
 
 Authors
--- a/docs/DesignDocs/RemoteAttestationCollaterals.md
+++ b/docs/DesignDocs/RemoteAttestationCollaterals.md
@@ -99,55 +99,22 @@ Currently, the existing endorsements used for Intel SGX quote verification are n
 User Experience
 ---------------
 
-There are 3 scenarios.
+There are 2 scenarios.
 
-### 1. Verifier does not specify the endorsements:
-In this scenario the verifier does not receive any endorsements from the attester (where the TEE resides).  The verifier also does not pass any endorsements to the verification function, instead the endorsements are fetched during verification by the OE SDK.
-
-##### Attester generates the evidence (inside an enclave/TEE)
-```C
-// Get OE report
-result = oe_get_report(
-    OE_REPORT_FLAGS_REMOTE_ATTESTATION,
-    NULL, // report_data
-    0,
-    NULL, // opt_params
-    0,
-    &report,
-    report_size);
-```
-
-##### Verifier verifies the evidence (in the untrusted host or inside an enclave/TEE)
-```C
-...
-
-// Verify report without endorsements
-result = oe_verify_remote_report_with_endorsements(
-            report,
-            report_size,
-            NULL, // endorsements
-            0,    // endorsements_size
-            NULL, // input_validation_time
-            parsed_report);
-...
-```
-
-### 2. Verifier is provided with endorsements:
+### 1. Verifier is provided with endorsements:
 In this scenario the attester/asserter provides the evidence and endorsements to the verifier.  The verifier is then free to use these to verify the TEE.
 
 ##### Attester generates the evidence and endorsements (inside an enclave/TEE)
 ```C
 ...
-result = oe_get_report(
-    OE_REPORT_FLAGS_REMOTE_ATTESTATION,
-    NULL, // report_data
+result = oe_get_evidence(
+    OE_EVIDENCE_FLAGS_REMOTE_ATTESTATION,
+    NULL, // custom_claims
     0,
     NULL, // opt_params
     0,
-    &report,
-    report_size);
-
-result = oe_get_endorsements(
+    &evidence,
+    &evidence_size,
     &endorsements,
     &endorsements_size);
 ...
@@ -158,30 +125,34 @@ result = oe_get_endorsements(
 ...
 
 // Verify report with endorsements
-result = oe_verify_remote_report_with_endorsements(
-            report,
-            report_size,
+result = oe_verify_evidence(
+            evidence,
+            evidence_size,
             endorsements,
             endorsements_size,
-            NULL, // input_validation_time
-            NULL, // parsed_report);
+            NULL,  // opt_params
+            0,
+            &claims,
+            &claims_size);
 ...
 ```
 
-### 3. Verifier specifies endorsements:
+### 2. Verifier specifies endorsements:
 In this scenario the attester only provides the evidence to the verifier.  The verifier then fetches the endorsements from a second source different than the OE SDK, and uses the evidence and endorsements to verify the TEE.
 
 ##### Attester generates the evidence (inside an enclave/TEE)
 ```C
 ...
-result = oe_get_report(
-    OE_REPORT_FLAGS_REMOTE_ATTESTATION,
-    NULL, // report_data
+result = oe_get_evidence(
+    OE_EVIDENCE_FLAGS_REMOTE_ATTESTATION,
+    NULL, // custom_claims
     0,
     NULL, // opt_params
     0,
-    &report,
-    report_size);
+    &evidence,
+    &evidence_size,
+    &endorsements,
+    &endorsements_size);
 ...
 ```
 
@@ -195,38 +166,49 @@ result = oe_get_report(
 //
 // Verifier builds **endorsements** structure
 //
+endorsements_external = ...
+endorsements_external_size = ...
 
-// Verify report with endorsements
-result = oe_verify_remote_report_with_endorsements(
+// Verify evidence with external endorsements
+result = oe_verify_evidence(
             report,
             report_size,
-            endorsements,
-            endorsements_size,
-            NULL, // input_validation_time
-            NULL, // parsed_report);
+            endorsements_external,
+            endorsements_external_size,
+            NULL,  // opt_params
+            0,
+            &claims,
+            &claims_size);
 ...
 ```
 
 Specification
 -------------
 
-### Endorsement structure
-Generic serializable structure that stores the endorsements in raw binary format.
+### <B>Public</B> type definitions
+Generic serializable public structure that stores the endorsements in raw binary format.
 
-`report.h`
+`attestation.h`
 ```C
+/**
+ * Flags passed to oe_get_evidence() function.
+ */
+#define OE_EVIDENCE_FLAGS_LOCAL_ATTESTATION  0x00000000
+#define OE_EVIDENCE_FLAGS_REMOTE_ATTESTATION 0x00000001
+
 /*! \struct oe_endorsements_t
  *
  * \brief OE endorsements
  *
- * Raw generic serializable structure that contains the endorsements.
+ * Raw generic serializable structure that contains the endorsements. All
+ * data should be in little endian format.
  *
  */
 typedef struct _oe_endorsements_t
 {
     uint32_t version;
     uint32_t enclave_type;        ///< The type of enclave
-    uint64_t buffer_size;         ///< Size of the buffer
+    uint64_t buffer_size;         ///< Size of the buffer  (oe_enclave_type_t)
     uint32_t num_elements;
 
     ///< Data buffer is made of an offset array of type uint32_t, followed by
@@ -238,13 +220,14 @@ typedef struct _oe_endorsements_t
     ///<  |-----------------------|
     ///<  |  enclave_type         |
     ///<  |-----------------------|
-    ///<  |  num_elements         |
-    ///<  |-----------------------|
     ///<  |  buffer_size          |
     ///<  |-----------------------|
-    ///<  |                       |
-    ///<  | array of uint32_t     |
-    ///<  | of length num_elements|
+    ///<  |  num_elements         |
+    ///<  |-----------------------|
+    ///<  |  offsets              |
+    ///<  |  (array of uint32_t   |
+    ///<  |  with length of       |
+    ///<  |  num_elements)        |
     ///<  |-----------------------|
     ///<  |  Data                 |
     ///<  |_______________________|
@@ -252,134 +235,202 @@ typedef struct _oe_endorsements_t
     uint8_t buffer[];              ///< Buffer of offsets + data
 
 } oe_endorsements_t;
+
+
+///< Number of CRLs in the SGX endorsements
+#define OE_SGX_ENDORSEMENTS_CRL_COUNT     (2)
+
+/*! \enum oe_sgx_endorsements_fields
+ *
+ * Specifies the order of the SGX endorsements fields stored in
+ * the oe_endorsements_t strcuture
+ */
+typedef enum _oe_sgx_endorsements_fields
+{
+    OE_SGX_ENDORSEMENT_FIELD_TCB_INFO = 0,
+    OE_SGX_ENDORSEMENT_FIELD_TCB_ISSUER_CHAIN = 1,
+    OE_SGX_ENDORSEMENT_FIELD_CRL_START_INDEX = 2,
+    OE_SGX_ENDORSEMENT_FIELD_CRL_END_INDEX =
+        OE_SGX_ENDORSEMENT_FIELD_CRL_START_INDEX + OE_SGX_ENDORSEMENTS_CRL_COUNT - 1,
+
+    OE_SGX_ENDORSEMENT_FIELD_CRL_ISSUER_CHAIN_START_INDEX =
+        OE_SGX_ENDORSEMENT_FIELD_CRL_END_INDEX + 1,
+
+    OE_SGX_ENDORSEMENT_FIELD_CRL_ISSUER_CHAIN_END_INDEX =
+        OE_SGX_ENDORSEMENT_FIELD_CRL_ISSUER_CHAIN_START_INDEX +
+        OE_SGX_ENDORSEMENTS_CRL_COUNT - 1,
+
+    OE_SGX_ENDORSEMENT_FIELD_QE_ID_INFO =
+        OE_SGX_ENDORSEMENT_FIELD_CRL_ISSUER_CHAIN_END_INDEX + 1,
+
+    OE_SGX_ENDORSEMENT_FIELD_QE_ID_ISSUER_CHAIN,
+    OE_SGX_ENDORSEMENT_FIELD_CREATION_DATETIME
+
+} oe_sgx_endorsements_fields;
 ```
 
-### New Endorsement functions
+### <B>Private</B> SGX endorsement definitions
 
-`endorsements.c`
+`common/sgx/evidence.h`
 ```C
-/*! \struct tee_endorsements
+/*! \struct oe_sgx_endorsements
  *
  * \brief SGX endorsements structure
  *
  * The generic oe_endorsements_t structure is parsed and converted into this
  * internal structure.  The order of the generic data elements should
  * coincide with the order of the fields in this structure.
+ *
+ * Data format: All data comes from the Data Center Attestation Primitives(DCAP)
+ * Client.
+ *
+ * For Azure DCAP Client
+ * (https://github.com/microsoft/Azure-DCAP-Client/blob/master/src/dcap_provider.h)
+ * see **sgx_ql_revocation_info_t** and sgx_qe_identity_info_t.
+ *
+ * For Intel DCAP Client
+ * (https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/README.md)
+ * see TBD.
+ *
  */
-typedef struct _tee_endorsements_t
+typedef struct _oe_sgx_endorsements_t
 {
-    uint8_t* tcb_info;                      ///< TCB info
-    size_t tcb_info_size;                   ///< TCB Info size
-    uint8_t* tcb_issuer_chain;              ///< PEM format
-    size_t tcb_issuer_chain_size;           ///< Size of the tcb_issuer_chain
-    uint8_t* crl[3];                        ///< CRLs
-    size_t crl_size[3];                     ///< CRLs sizes
-    uint8_t* crl_issuer_chain[3];           ///< PEM format
-    size_t crl_issuer_chain_size[3];        ///< Size of each crl_issuer_chain
+    uint8_t* tcb_info;                      ///< TCB info, null-terminated JSON string
+    uint32_t tcb_info_size;                 ///< TCB Info size
+    uint8_t* tcb_issuer_chain;              ///< PEM format, null-terminated string
+    uint32_t tcb_issuer_chain_size;         ///< Size of the tcb_issuer_chain
 
-    uint8_t* qe_id_info;                    ///< QE Identity info
-    size_t qe_id_info_size;                 ///< QE Identity size
-    uint8_t* qe_id_issuer_chain;            ///< PEM format
-    size_t qe_id_issuer_chain_size;         ///< Size of qe_id_issuer_chain
+    ///< CRLs in DER format, null-terminated
+    ///<     crl[0] = CRL for the SGX PCK Certificate
+    ///<     crl[1] = CRL for the SGX PCK Processor CA
+    uint8_t* crl[OE_SGX_ENDORSEMENTS_CRL_COUNT];
 
-    uint8_t* creation_datetime;             ///< Time the endorsements were generated
-    size_t create_datetime_size;            ///< The size of creation_datetime.  Should be 24.
+    ///< CRLs sizes
+    uint32_t crl_size[OE_SGX_ENDORSEMENTS_CRL_COUNT];
 
-} tee_endorsements_t;
+    ///< PEM format, null-terminated string
+    uint8_t* crl_issuer_chain[OE_SGX_ENDORSEMENTS_CRL_COUNT];
 
+    ///< Size of each crl_issuer_chain
+    uint32_t crl_issuer_chain_size[OE_SGX_ENDORSEMENTS_CRL_COUNT];
+
+    uint8_t* qe_id_info;                    ///< QE Identity info, null-terminated JSON string
+    uint32_t qe_id_info_size;               ///< QE Identity size
+    uint8_t* qe_id_issuer_chain;            ///< PEM format, null-terminated string
+    uint32_t qe_id_issuer_chain_size;       ///< Size of qe_id_issuer_chain
+
+    uint8_t* creation_datetime;             ///< Time the endorsements were generated,
+                                            ///< null-terminated string
+    uint32_t create_datetime_size;          ///< The size of creation_datetime.
+
+} oe_sgx_endorsements_t;
+```
+
+### New <B>Public</B> Attestation functions
+
+These functions supersede the existing functions:
+1. `oe_get_evidence()` supersedes `oe_verify_report()`
+2. `oe_verify_evidence()` supersedes `oe_verify_report()`
+
+These functions will use plug-in attestation framework.  For more information please
+see the design document `CustomAttestation.md`.  In short, the actual implementation
+of these functions will depend on which plug-in is registered.  By default there will
+be a built-in SGX plug-in.
+
+For more information on the parameters `custom_claims`, `claims` or `opt_params`
+please see design document `CustomAttestation.md`.
+
+`common/sgx/attestation.c`
+```C
 /**
- * Returns the endorsements needed for verification of a remote OE report.
+ * Get evidence signed by the enclave platform along with the corresponding
+ * endorsements for use in attestation.
  *
- * This function is only available inside an enclave.
+ * This function returns the evidence and endorsements used in **local** or
+ * **remote** attestation.
  *
- * @param[in] enclave The instance of the enclave that will be used.
- * @param[in] endorsements_buffer The buffer containing the endorsements to parse.
- * @param[in] endorsements_buffer_size The size of the **endorsements_buffer**.
+ * This function can only be called from a TEE/enclave.
  *
- * @retval OE_OK The endorsements were successfully retrieved.
+ * @param[in] flags Specifying default value (0) generates evidence for local
+ * attestation. Specifying OE_EVIDENCE_FLAGS_REMOTE_ATTESTATION generates
+ * evidence for remote attestation.
+ * @param[in] custom_claims A buffer to the optional custom claims.
+ * @param[in] custom_claims_size The size in bytes of custom_claims.
+ * @param[in] opt_params Optional additional parameters needed for the current
+ * enclave type.
+ *     For SGX:
+ *        This can be sgx_target_info_t for local attestation.
+ * @param[in] opt_params_size The size of the **opt_params** buffer.
+ * @param[out] evidence_buffer This points to the resulting evidence upon success.
+ * @param[out] evidence_buffer_size This is set to the size of the evidence buffer
+ * on success.
+ * @param[out] endorsements_buffer The buffer containing the endorsements to parse.
+ * @param[out] endorsements_buffer_size The size of the **endorsements_buffer**.
+ *
+ * @retval OE_OK The evidence and endorsements were successfully created.
+ * @retval OE_INVALID_PARAMETER At least one parameter is invalid.
+ * @retval OE_OUT_OF_MEMORY Failed to allocate memory.
+ *
  */
-oe_result_t oe_get_endorsements(
+oe_result_t oe_get_evidence(
+    uint32_t flags,
+    const uint8_t* custom_claims,
+    size_t custom_claims_size,
+    const void* opt_params,
+    size_t opt_params_size,
+    uint8_t** evidence_buffer,
+    size_t* evidence_buffer_size,
     uint8_t** endorsements_buffer,
     size_t* endorsements_buffer_size);
 
 /**
- * Free up any resources allocated by oe_get_endorsements()
+ * Free up any resources allocated by oe_get_evidence()
  *
+ * @param[in] evidence_buffer THe buffer containing the evidence.
  * @param[in] endorsements_buffer The buffer containing the endorsements.
  */
-void oe_free_endorsements(uint8_t* endorsements_buffer);
+void oe_free_evidence(
+    uint8_t* endorsements_buffer,
+    uint8_t* evidence_buffer);
 
 /**
- * Verify the integrity of the report and its signature,
- * with optional endorsements that are associated with the report. If
- * the endorsements are not specified, this function will fetch
- * the endorsements.  This only applies to remote reports.  For
- * local reports please use oe_verify_report().
- *
- * This function is similar to oe_verify_report() but it supports
- * endorsements.
+ * Verify the integrity of the evidence and its signature,
+ * with endorsements that are associated with the evidence.
+ * This function works for both local and remote attestation.
  *
  * This function is available in the enclave as well as in the host.
  *
- * @param[in] report The buffer containing the report to verify.
- * @param[in] report_size The size of the **report** buffer.
+ * @param[in] evidence_buffer The buffer containing the evidence to verify.
+ * @param[in] evidence_buffer_size The size of the **evidence** buffer.
  * @param[in] endorsements Optional The endorsement data that is associated with
- * the report (for remote reports only).
+ * the evidence.
  * @param[in] endorsements_size The size of the **endorsements** buffer.
- * @param[in] input_validation_time Optional datetime to use when validating
- * endorsements. If not specified, it will use the creation_datetime of the
+ * @param[in] input_validation_time Optional datetime to use when verifying
+ * evidence. If not specified, it will use the creation_datetime of the
  * endorsements (if any endorsements are provided).
- * @param[out] parsed_report Optional **oe_report_t** structure to populate with
- * the report properties in a standard format.
+ * @param[out] claims The list of claims.
+ * @param[out] claims_size The size of claims.
  *
- * @retval OE_OK The report was successfully created.
+ * @retval OE_OK The verification was successful.
  * @retval OE_INVALID_PARAMETER At least one parameter is invalid.
  *
  */
-oe_result_t oe_verify_report_with_endorsements(
-    const uint8_t* report,
-    size_t report_size,
+oe_result_t oe_verify_evidence(
+    const uint8_t* evidence_buffer,
+    size_t evidence_buffer_size,
     const uint8_t* endorsements,
     size_t endorsements_size,
     oe_datetime_t* input_validation_time,
-    oe_report_t* parsed_report);
+    uint8_t** claims,
+    size_t* claims_size);
 ```
 
-### Update existing functions
+### OE Host Verify Library
 
-In the host_verify library, this is the standalone library used for verifying remote reports outside
-the TEE/enclave. Update oe_verfiy_remote_report() to take endorsements as an input parameter.
+The OE Host Verify library is a standalone library used for verifying remote reports outside
+the TEE/enclave. The function `oe_verfiy_remote_report()` will be deprecated and should use the
+upcoming plug-in mode to do verification.
 
-`host/sgx/hostverify_report.c`
-```C
-/**
- * Verify the integrity of the report and its signature,
- * with optional endorsements that are associated with the report. If
- * the endorsements are not specified, this function will fetch
- * the endorsements.
- *
- * @param[in] report The buffer containing the report to verify.
- * @param[in] report_size The size of the **report** buffer.
- * @param[in] endorsements Optional The endorsement data that is associated with
- * the report (for remote reports only).
- * @param[in] endorsements_size The size of the **endorsements** buffer.
- * @param[in] input_validation_time Optional datetime to use when validating
- * endorsements. If not specified, it will use the creation_datetime of the
- * endorsements (if any endorsements are provided).
- * @param[out] parsed_report Optional **oe_report_t** structure to populate with
- * the report properties in a standard format.
- *
- * @retval OE_OK The report was successfully created.
- * @retval OE_INVALID_PARAMETER At least one parameter is invalid.
- *
- */
-oe_result_t oe_verify_remote_report(
-    const uint8_t* report,
-    size_t report_size,
-    const uint8_t* endorsements,
-    size_t endorsements_size,
-    oe_report_t* parsed_report);
-```
 
 Alternates
 ----------
@@ -392,9 +443,7 @@ user defined endorsements.  But we ended up deciding to have the user build the 
 structure instead, to avoid complexity.
 
 ### APIs:
-- To reduce complexity, `oe_get_endorsements()` is only available in the enclave.
-- To reduce complexity and break existing API users, `oe_verify_remote_report_with_endorsements()`
-only supports remote reports.  For local reports use `oe_verify_report()`
+- To reduce complexity, `oe_get_evidence()` is only available in the enclave.
 
 Authors
 -------
--- a/docs/DesignDocs/RemoteAttestationCollaterals.md
+++ b/docs/DesignDocs/RemoteAttestationCollaterals.md
@@ -1,0 +1,292 @@
+Remote Attestation Collaterals
+=====
+
+Adding attestation APIs to support collaterals for remote attestation.  Collaterals are information from a second source that the user/challenger can provide for attestation verification.
+
+Motivation
+----------
+
+Currently, the existing collaterals used for Intel SGX quote verification are not exposed to the user.  This makes it difficult for the user/challenger to specify his/her own set of policies.  Adding these new APIs allows the user to specify a validation policy of his/her choosing.  Possible policies:
+1. The user/challenger has the option to specify its set of collaterals during verification.
+2. The user/challenger has the option to provide a datetime to use during verification.  This datetime specified the date and time at which the user wants to do the verification.  If no datetime is provided, it uses the datetime when the collaterals were created.  The user can provide a datetime in the past, enabling auditing of the evidence and collaterals.
+
+
+User Experience
+---------------
+
+There are 3 scenarios.
+
+### 1. Challenger does not specify the collaterals:
+In this scenario the challenger does not receive any collaterals from the attester(where the TEE resides).  The challenger also does not pass any collaterals to the verification function, instead the collaterals are fetched during verification by the OE SDK.
+
+##### Attester generates the evidence (inside an enclave/TEE)
+```C
+// Get OE report
+result = oe_get_report(
+    OE_REPORT_FLAGS_REMOTE_ATTESTATION,
+    NULL, // report_data
+    0,
+    NULL, // opt_params
+    0,
+    &report,
+    report_size);
+```
+
+##### Challenger verifies the evidence (in the untrusted host)
+```C
+...
+
+// Verify report without collaterals
+result = oe_verify_remote_report_with_collaterals(
+            report,
+            report_size,
+            NULL, // collaterals
+            0,    // collaterals_size
+            NULL, // input_validation_time
+            parsed_report);
+...
+```
+
+### 2. Challenger is provided with collaterals:
+In this scenario the attester provides the evidence and collaterals to the challenger.  The challenger then uses these to verify the TEE.
+
+##### Attester generates the evidence and collaterals (inside an enclave/TEE)
+```C
+...
+result = oe_get_report(
+    OE_REPORT_FLAGS_REMOTE_ATTESTATION,
+    NULL, // report_data
+    0,
+    NULL, // opt_params
+    0,
+    &report,
+    report_size);
+
+result = oe_get_collaterals(
+    &collaterals,
+    &collaterals_size);
+...
+```
+
+##### Challenger verifies the evidence and collaterals (in the untrusted host)
+```C
+...
+
+// Verify report with collaterals
+result = oe_verify_remote_report_with_collaterals(
+            report,
+            report_size,
+            collaterals,
+            collaterals_size,
+            NULL, // input_validation_time
+            NULL, // parsed_report);
+...
+```
+
+### 3. Challenger specifies collaterals:
+In this scenario the attester only provides the evidence to the challenger.  The challenger then fetches the collaterals from a second source different than the OE SDK, and uses the evidence and collaterals to verify the TEE.
+
+##### Attester generates the evidence and collaterals (inside an enclave/TEE)
+```C
+...
+result = oe_get_report(
+    OE_REPORT_FLAGS_REMOTE_ATTESTATION,
+    NULL, // report_data
+    0,
+    NULL, // opt_params
+    0,
+    &report,
+    report_size);
+...
+```
+
+##### Challenger verifies the evidence with custom collaterals (in the untrusted host)
+```C
+...
+//
+// Challenger gets collaterals not using OE SDK
+//
+
+//
+// Challenger builds **collaterals** structure
+//
+
+// Verify report with collaterals
+result = oe_verify_remote_report_with_collaterals(
+            report,
+            report_size,
+            collaterals,
+            collaterals_size,
+            NULL, // input_validation_time
+            NULL, // parsed_report);
+...
+```
+
+Specification
+-------------
+
+### Collateral structure
+Generic serializable structure that stores the collaterals in raw binary format.
+
+`report.h`
+```C
+/*! \struct oe_collaterals_t
+ *
+ * \brief OE collaterals
+ *
+ * Raw generic serializable structure that contains the collaterals.
+ *
+ */
+typedef struct _oe_collaterals_t
+{
+    uint32_t version;
+    uint8_t enclave_type;         ///< The type of enclave
+    uint8_t reserved;
+    uint16_t reserved2;
+    uint32_t num_elements;
+    size_t buffer_size;           ///< Size of the buffer
+
+    ///< Data buffer is made of an offset array of type uint32_t, followed by
+    ///< the actual data.
+    ///< This array has the size of **num_elements** and stores the offset
+    ///< into the data section.
+    ///<  _________________________
+    ///<  |  version              |
+    ///<  |-----------------------|
+    ///<  |  reserved    | type   |
+    ///<  |-----------------------|
+    ///<  |  num_elements         |
+    ///<  |-----------------------|
+    ///<  |  buffer_size          |
+    ///<  |-----------------------|
+    ///<  |                       |
+    ///<  |  array(uint32_t)      |
+    ///<  |  size of num_elements |
+    ///<  |-----------------------|
+    ///<  |  Data                 |
+    ///<  |_______________________|
+    ///<
+    uint8_t buffer[];              ///< Buffer of offsets + data
+
+} oe_collaterals_t;
+```
+
+### New Collateral functions
+
+`collaterals.c`
+```C
+/*! \struct sgx_collaterals
+ *
+ * \brief SGX collateral structure
+ *
+ * The generic oe_collaterals_t structure is parsed and converted into this
+ * internal structure.  The order of the generic data elements should
+ * coincide with the order of the fields in this structure.
+ */
+typedef struct _sgx_collaterals_t
+{
+    uint8_t* tcb_info;                      ///< TCB info
+    size_t tcb_info_size;                   ///< TCB Info size
+    uint8_t* tcb_issuer_chain;              ///< PEM format
+    size_t tcb_issuer_chain_size;           ///< Size of the tcb_issuer_chain
+    uint8_t* crl[3];                        ///< CRLs
+    size_t crl_size[3];                     ///< CRLs sizes
+    uint8_t* crl_issuer_chain[3];           ///< PEM format
+    size_t crl_issuer_chain_size[3];        ///< Size of each crl_issuer_chain
+
+    uint8_t* qe_id_info;                    ///< QE Identity info
+    size_t qe_id_info_size;                 ///< QE Identity size
+    uint8_t* qe_id_issuer_chain;            ///< PEM format
+    size_t qe_id_issuer_chain_size;         ///< Size of qe_id_issuer_chain
+
+    uint8_t* creation_datetime;             ///< Time the collaterals were generated
+    size_t create_datetime_size;            ///< The size of creation_datetime.  Should be 24.
+
+} sgx_collaterals_t;
+
+/**
+ * Returns the collaterals needed for verification of a remote OE report.
+ *
+ * This function is only available inside an enclave.
+ *
+ * @param[in] enclave The instance of the enclave that will be used.
+ * @param[in] collaterals_buffer The buffer containing the collaterals to parse.
+ * @param[in] collaterals_buffer_size The size of the **collaterals_buffer**.
+ *
+ * @retval OE_OK The collaterals were successfully retrieved.
+ */
+oe_result_t oe_get_collaterals(
+    uint8_t** collaterals_buffer,
+    size_t* collaterals_buffer_size);
+
+/**
+ * Free up any resources allocated by oe_get_collateras()
+ *
+ * @param[in] collaterals_buffer The buffer containing the collaterals.
+ */
+void oe_free_collaterals(uint8_t* collaterals_buffer);
+
+/**
+ * Verify the integrity of the report and its signature,
+ * with optional collaterals that is associated with the report. If
+ * the collaterals are not specified, this function will fetch
+ * the collaterals.  This only applies to remote reports.  For
+ * local reports please use oe_verify_report().
+ *
+ * This function is similar to oe_verify_report() but it supports
+ * collaterals.
+ *
+ * This function is available in the enclave as well as in the host.
+ *
+ * @param[in] enclave The instance of the enclave that will be used.
+ * @param[in] report The buffer containing the report to verify.
+ * @param[in] report_size The size of the **report** buffer.
+ * @param[in] collaterals Optional The collateral data that is associated with
+ * the report (for remote reports only).
+ * @param[in] collaterals_size The size of the **collaterals** buffer.
+ * @param[in] input_validation_time Optional datetime to use when validating
+ * collaterals. If not specified, it will used the creation_datetime of the
+ * collaterals (if any collaterals are provided).
+ * @param[out] parsed_report Optional **oe_report_t** structure to populate with
+ * the report properties in a standard format.
+ *
+ * @retval OE_OK The report was successfully created.
+ * @retval OE_INVALID_PARAMETER At least one parameter is invalid.
+ *
+ */
+oe_result_t oe_verify_remote_report_with_collaterals(
+#ifndef OE_BUILD_ENCLAVE
+    oe_enclave_t* enclave,
+#endif
+    const uint8_t* report,
+    size_t report_size,
+    const uint8_t* collaterals,
+    size_t collaterals_size,
+    oe_datetime_t* input_validation_time,
+    oe_report_t* parsed_report);
+
+```
+
+Alternates
+----------
+
+### Collateral structure:
+- Multiple serializable structures were considered.  At the end the final design
+made sense given that it is generic and support serialization.  We considered
+supporting multiple flavors of the structure, a binary format and another for supporting
+user defined collaterals.  But we ended up deciding to have the user build the binary
+structure instead, to avoid complexity.
+
+### APIs:
+- To reduce complexity, `oe_get_collaterals()` is only available in the enclave.
+- To reduce complexity, `oe_verify_remote_report_with_collaterals()` only supports
+remote reports.  For local reports use `oe_verify_report()`
+
+Authors
+-------
+
+Name: Sergio Wong
+
+email: sewong@microsoft.com
+
+github username: jazzybluesea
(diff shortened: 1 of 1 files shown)


docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,285 @@
+Remote Attestation Collaterals
+=====
+
+Adding attestation APIs to support collaterals for remote attestation.  Collaterals are information from a second source that the user can provide for attestation verification.
+
+Motivation
+----------
+
+Currently, the existing collaterals used for Intel SGX quote verification are not exposed to the user.  Adding these new APIs allows the user to specify these collaterals instead of using the default collaterals used by OE SDK.
+
```
soccerGB: it would make a stronger case to describe a actual use case for it. -- such as some scenarios have a need to validate a previously generate quote with collaterals collected when the quote was generated for auditing purpose, instead of using the most up-to-date collaterals.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,285 @@
...
+    0,
+    NULL, // opt_params
+    0,
+    &report,
+    report_size);
+```
+
+##### Challenger verifies the evidence (in the host)
+```C
+...
```
soccerGB: you meant from an environment without any TEE support, right?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,285 @@
...
+            0,    // collaterals_size
+            NULL, // input_validation_time
+            parsed_report);
+...
+```
+
+### 2. User is provided with collaterals:
+
+##### Server generates the evidence and collaterals (inside an enclave)
+```C
```
soccerGB: What do you mean by "server" here?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,285 @@
...
+Currently, the existing collaterals used for Intel SGX quote verification are not exposed to the user.  Adding these new APIs allows the user to specify these collaterals instead of using the default collaterals used by OE SDK.
+
+
+User Experience
+---------------
+
+There are 3 user scenarios.
+
+### 1. User does not specify the collaterals:
+
```
soccerGB: Could you add some more description on each user scenario before getting to API?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,285 @@
...
+/*! \struct oe_collaterals
+ *
+ * \brief OE collaterals
+ *
+ * Raw generic serializable structure that contains the collaterals.
+ *
+ */
+typedef enum _oe_collaterals
+{
+    uint32_t version;
```
soccerGB: Is the input_validation_time supposed to be defined in this structure? Or it's one of the elements?

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Good point! I added this other scenario.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Yes, host implies the untrusted host.  The machine may or may not support a TEE.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: The server implies a machine that supports a TEE.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Done.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: it is one of the elements.  It is defined in the internal structure.  I have updated the internal structure.

docs/DesignDocs/RemoteAttestationCollaterals.md
soccerGB: I see. In that case, it’s better to use the attester instead of server.

docs/DesignDocs/RemoteAttestationCollaterals.md
soccerGB: Should it be in a public structure instead of an old one?

jazzybluesea: oe_collaterals_t is a public structure.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
+Remote Attestation Collaterals
+=====
+
+Adding attestation APIs to support collaterals for remote attestation.  Collaterals are information from a second source that the user/challenger can provide for attestation verification.
+
+Motivation
+----------
+
+Currently, the existing collaterals used for Intel SGX quote verification are not exposed to the user.  Adding these new APIs allows the user to specify these collaterals instead of using the default collaterals used by OE SDK.  This enables two possible user scenarios:
+1. The user/challenger has the option to specify its set of collaterals during verification.
```
soccerGB: Maybe we could describe in a way to convey the following idea (not exact words:

The existing SDK does not allow an enclave client to provide any policy input during the validation process. The policy was fixed in OE SDK's choosing. 

We are giving users the control of the quote validation policy by providing new APIs to provide collaterals (validation policy). They can control the policy (collaterals) used in the validation of their quote as long as they provide the collaterals in the expected format.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+    ///<  |  array(uint32_t)      |
+    ///<  |  size of num_elements |
+    ///<  |-----------------------|
+    ///<  |  Data                 |
+    ///<  |_______________________|
+    ///<
+    uint8_t buffer[];              ///< Buffer of offsets + data
+
+} oe_collaterals;
+```
```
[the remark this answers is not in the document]
jazzybluesea: Fixed typo.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Updated.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
+Remote Attestation Collaterals
+=====
+
+Adding attestation APIs to support collaterals for remote attestation.  Collaterals are information from a second source that the user/challenger can provide for attestation verification.
```
dthaler: FYI, the IETF RATS WG is now using the term "Endorsements" for things like PCK cert chain, CRL, etc.
Endorsements might be provided to a verifier by an Attester (along with the Evidence), or might be obtained separately by a verifier.   What's the relationship between Collaterals and Endorsements?

If they're the same, then I'd recommend changing to use IETF terminology.

dthaler: When you say "user/challenger", is that referring to the Relying Party or a Verifier, or what?   (I can't tell what "user" refers to here.)

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+
+Motivation
+----------
+
+Currently, the existing collaterals used for Intel SGX quote verification are not exposed to the user.  This makes it difficult for the user/challenger to specify his/her own set of policies.  Adding these new APIs allows the user to specify a validation policy of his/her choosing.  Possible policies:
+1. The user/challenger has the option to specify its set of collaterals during verification.
+2. The user/challenger has the option to provide a datetime to use during verification.  This datetime specified the date and time at which the user wants to do the verification.  If no datetime is provided, it uses the datetime when the collaterals were created.  The user can provide a datetime in the past, enabling auditing of the evidence and collaterals.
```
dthaler: Can't understand this.  Is the user/challenger the Verifier or the Relying Party or what?
And what is the "it" in "it uses the datetime"?  The Verifier?  The Attester?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+
+
+User Experience
+---------------
+
+There are 3 scenarios.
+
+### 1. Challenger does not specify the collaterals:
+In this scenario the challenger does not receive any collaterals from the attester(where the TEE resides).  The challenger also does not pass any collaterals to the verification function, instead the collaterals are fetched during verification by the OE SDK.
```
dthaler: As phrased I think the first sentence assumes the challenger is the Verifier not the Relying Party, since a Relying Party would receive info from a Verifier instead of an Attester.  Am I interpreting the intent correctly?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+    OE_REPORT_FLAGS_REMOTE_ATTESTATION,
+    NULL, // report_data
+    0,
+    NULL, // opt_params
+    0,
+    &report,
+    report_size);
+```
+
+##### Challenger verifies the evidence (in the untrusted host)
```
dthaler: Recommend replacing "Challenger" with "Verifier" to match IETF nomenclature.
However the "in the untrusted host" is confusing since that's only one of two cases,
and probably the less interesting (because less secure) of the two.   Can you add
"Verifier verifies the evidence (in another enclave)"?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+            report_size,
+            NULL, // collaterals
+            0,    // collaterals_size
+            NULL, // input_validation_time
+            parsed_report);
+...
+```
+
+### 2. Challenger is provided with collaterals:
+In this scenario the attester provides the evidence and collaterals to the challenger.  The challenger then uses these to verify the TEE.
```
dthaler: Do you assume that the collateral provided by the attester is necessarily identical to what the verifier would be able to get itself?  (Even if it is, there's a race condition where it may change right after the attester gets it and before the verifier would ask for it.)   But in general I suspect there might be cases where the verifier might want additional endorsements beyond ones the attester knows about, which scenario 3 admits.  But the point is that the Attester may not know and may send collaterals even though the Verifier is in scenario 3, unless you design a way for a Verifier to tell an Attester to not bother sending collaterals.

So "then uses" might be overstated, maybe you mean "is then free to use" would be more correct.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+    &report,
+    report_size);
+
+result = oe_get_collaterals(
+    &collaterals,
+    &collaterals_size);
+...
+```
+
+##### Challenger verifies the evidence and collaterals (in the untrusted host)
```
dthaler: See earlier comment, where there are also scenario requirements to do verification in an enclave.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+            collaterals_size,
+            NULL, // input_validation_time
+            NULL, // parsed_report);
+...
+```
+
+### 3. Challenger specifies collaterals:
+In this scenario the attester only provides the evidence to the challenger.  The challenger then fetches the collaterals from a second source different than the OE SDK, and uses the evidence and collaterals to verify the TEE.
+
+##### Attester generates the evidence and collaterals (inside an enclave/TEE)
```
dthaler: This says "and collaterals" but line 87 says only evidence.   I think "and collaterals" should be deleted here.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+    NULL, // report_data
+    0,
+    NULL, // opt_params
+    0,
+    &report,
+    report_size);
+...
+```
+
+##### Challenger verifies the evidence with custom collaterals (in the untrusted host)
```
dthaler: See earlier comments about also needing verification in an enclave.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+ * \brief OE collaterals
+ *
+ * Raw generic serializable structure that contains the collaterals.
+ *
+ */
+typedef struct _oe_collaterals_t
+{
+    uint32_t version;
+    uint8_t enclave_type;         ///< The type of enclave
+    uint8_t reserved;
```
dthaler: since you already have a version field above, why do you need any reserved fields?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+    ///<  |  buffer_size          |
+    ///<  |-----------------------|
+    ///<  |                       |
+    ///<  |  array(uint32_t)      |
+    ///<  |  size of num_elements |
+    ///<  |-----------------------|
+    ///<  |  Data                 |
+    ///<  |_______________________|
+    ///<
+    uint8_t buffer[];              ///< Buffer of offsets + data
```
dthaler: Line 163 says array of uint32_t but here it's uint8_t.   Can this be uint32_t?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+} oe_collaterals_t;
+```
+
+### New Collateral functions
+
+`collaterals.c`
+```C
+/*! \struct sgx_collaterals
+ *
+ * \brief SGX collateral structure
```
dthaler: Is collateral singular or plural?  It's plural in 178, 182, 186, etc.  But perhaps this line is right and those are wrong?

docs/DesignDocs/RemoteAttestationCollaterals.md
jiria: Imho we should state that it can run both in the untrusted code or in the TEE.

docs/DesignDocs/RemoteAttestationCollaterals.md
jiria: Nit: space missing before (.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+    uint8_t* tcb_info;                      ///< TCB info
+    size_t tcb_info_size;                   ///< TCB Info size
+    uint8_t* tcb_issuer_chain;              ///< PEM format
+    size_t tcb_issuer_chain_size;           ///< Size of the tcb_issuer_chain
+    uint8_t* crl[3];                        ///< CRLs
+    size_t crl_size[3];                     ///< CRLs sizes
+    uint8_t* crl_issuer_chain[3];           ///< PEM format
+    size_t crl_issuer_chain_size[3];        ///< Size of each crl_issuer_chain
+
+    uint8_t* qe_id_info;                    ///< QE Identity info
```
dthaler: Is the QE identity info specific to the machine or machine-independent?

I'd like to make sure that the same "collaterals" struct doesn't try to combine machine-specific and machine-independent info into the same struct, especially if this struct is sent across a network.  That's so (for example) machine specific information can be sent by an Attester's machine, whereas machine-independent info could be obtained separately (or cached across many machines) by a Verifier.   The "num_elements" design above should allow multiple independent structs to be used together in the same message just fine, if I understand correctly, so there's no need to use a combined struct.

docs/DesignDocs/RemoteAttestationCollaterals.md
jiria: Copy & paste? Attester does not generate collaterals in this case, right?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+ * @param[in] collaterals_buffer_size The size of the **collaterals_buffer**.
+ *
+ * @retval OE_OK The collaterals were successfully retrieved.
+ */
+oe_result_t oe_get_collaterals(
+    uint8_t** collaterals_buffer,
+    size_t* collaterals_buffer_size);
+
+/**
+ * Free up any resources allocated by oe_get_collateras()
```
dthaler: typo "collateras"

docs/DesignDocs/RemoteAttestationCollaterals.md
jiria: Similar to above, I think we should be explicit that this can run both in trusted and untrusted code.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+ *
+ * @param[in] collaterals_buffer The buffer containing the collaterals.
+ */
+void oe_free_collaterals(uint8_t* collaterals_buffer);
+
+/**
+ * Verify the integrity of the report and its signature,
+ * with optional collaterals that is associated with the report. If
```
dthaler: grammar: "collaterals … is" ?   (line 249 has "collaterals are")

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+ *
+ */
+typedef struct _oe_collaterals_t
+{
+    uint32_t version;
+    uint8_t enclave_type;         ///< The type of enclave
+    uint8_t reserved;
+    uint16_t reserved2;
+    uint32_t num_elements;
+    size_t buffer_size;           ///< Size of the buffer
```
jiria: Should we use a specific size type for the buffer_size, since these might be used across hosts?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+ * @param[in] collaterals_buffer The buffer containing the collaterals.
+ */
+void oe_free_collaterals(uint8_t* collaterals_buffer);
+
+/**
+ * Verify the integrity of the report and its signature,
+ * with optional collaterals that is associated with the report. If
+ * the collaterals are not specified, this function will fetch
+ * the collaterals.  This only applies to remote reports.  For
+ * local reports please use oe_verify_report().
```
dthaler: Why?   I find this unnecessarily complicated to force a difference on the app developer.
Can't we just have one combined API?   Currently this comes across as over-indexing on an Intel SGX-only design, rather than proposing a processor-agnostic API.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+ * This function is available in the enclave as well as in the host.
+ *
+ * @param[in] enclave The instance of the enclave that will be used.
+ * @param[in] report The buffer containing the report to verify.
+ * @param[in] report_size The size of the **report** buffer.
+ * @param[in] collaterals Optional The collateral data that is associated with
+ * the report (for remote reports only).
+ * @param[in] collaterals_size The size of the **collaterals** buffer.
+ * @param[in] input_validation_time Optional datetime to use when validating
+ * collaterals. If not specified, it will used the creation_datetime of the
```
dthaler: typo: "will used"

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+    ///<  |_______________________|
+    ///<
+    uint8_t buffer[];              ///< Buffer of offsets + data
+
+} oe_collaterals_t;
+```
+
+### New Collateral functions
+
+`collaterals.c`
```
jiria: Should this be `sgx_collaterals.c`? It seems that in the past we have used the TEE prefix where we had custom per TEE implementation.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+
+    uint8_t* qe_id_info;                    ///< QE Identity info
+    size_t qe_id_info_size;                 ///< QE Identity size
+    uint8_t* qe_id_issuer_chain;            ///< PEM format
+    size_t qe_id_issuer_chain_size;         ///< Size of qe_id_issuer_chain
+
+    uint8_t* creation_datetime;             ///< Time the collaterals were generated
+    size_t create_datetime_size;            ///< The size of creation_datetime.  Should be 24.
+
+} sgx_collaterals_t;
```
jiria: Should we have support here for user generated policy? Aka similar blob to the one that AAS would use?

docs/DesignDocs/RemoteAttestationCollaterals.md
jiria: Typo: `oe_get_collaterals()`.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+ *
+ * This function is only available inside an enclave.
+ *
+ * @param[in] enclave The instance of the enclave that will be used.
+ * @param[in] collaterals_buffer The buffer containing the collaterals to parse.
+ * @param[in] collaterals_buffer_size The size of the **collaterals_buffer**.
+ *
+ * @retval OE_OK The collaterals were successfully retrieved.
+ */
+oe_result_t oe_get_collaterals(
```
jiria: Would this function be used by the challenger as well for the Scenario 3? If so, wondering if it should have an input parameter where the policy can be specified.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+ *
+ * @param[in] collaterals_buffer The buffer containing the collaterals.
+ */
+void oe_free_collaterals(uint8_t* collaterals_buffer);
+
+/**
+ * Verify the integrity of the report and its signature,
+ * with optional collaterals that is associated with the report. If
+ * the collaterals are not specified, this function will fetch
+ * the collaterals.  This only applies to remote reports.  For
```
jiria: I wonder if this function should be useable for local reports as well or if `oe_verify_report()` should be extended to accept collaterals as well. I am thinking if the local enclave has an attestation policy to be used against a local report, how would that work?

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Will change to use IETF terminology.  Thank you Dave.  Just note that to get the evidence we will still be using oe_get_report().

jazzybluesea: Verifier.  Will add a terminology section at the top.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Will rephrase, the "it" is unclear in this case.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Yes.

jazzybluesea: Fixed.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: How about "Verifier verifies the evidence (in another enclave or untrusted host)"? Since the verification technically could happen in any untrusted host given that we have the host verifier library now available for SGX.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Correct forgot to remove "and collaterals".

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Agree.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: It should be plural.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Fixed.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Fixed.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: The alternative is to update the current function oe_verify_remote_report() to take collaterals as input parameters.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Fixed.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Correct.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: I changed it to uint64_t to be more generic.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Changed it to tee_*

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Fixed.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: collaterals apply only to remote attestation and not local attestation.

jiria: My impression was that collaterals would also include attestation policy. If collaterals are not to be used for local attestation, how will attestation policy be passed to local attestation?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+=====
+
+Adding attestation APIs to support collaterals for remote attestation.  Collaterals are information from a second source that the user/challenger can provide for attestation verification.
+
+Motivation
+----------
+
+Currently, the existing collaterals used for Intel SGX quote verification are not exposed to the user.  Adding these new APIs allows the user to specify these collaterals instead of using the default collaterals used by OE SDK.  This enables two possible user scenarios:
+1. The user/challenger has the option to specify its set of collaterals during verification.
+2. The user/challenger can perform the verification on evidence and collaterals that were created in the past.  The user can audit the verification results anytime he/she wants.  The user will have to provide a datetime to verify against.
```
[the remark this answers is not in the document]
jazzybluesea: I think scenario #1 covers this?

docs/DesignDocs/RemoteAttestationCollaterals.md
gupta-ak: good point it should be `uint64_t` or `uint32_t`, since it's being serialized.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+
+
+User Experience
+---------------
+
+There are 3 scenarios.
```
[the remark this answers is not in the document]
jazzybluesea: The APIs are spelled out as part of the specification.  The user experience section is supposed to give a feel of how the flow would look like.  If the reader wants more details he/she can look at the specification.  

I will be adding a terminology section at the beginning.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+    uint8_t enclave_type;         ///< The type of enclave
+    uint8_t reserved;
+    uint16_t reserved2;
+    uint32_t num_elements;
+    size_t buffer_size;           ///< Size of the buffer
+
+    ///< Data buffer is made of an offset array of type uint32_t, followed by
+    ///< the actual data.
+    ///< This array has the size of **num_elements** and stores the offset
+    ///< into the data section.
```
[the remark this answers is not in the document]
jazzybluesea: > I think this statement could be interpreted as `sizeof(num_elements)` instead of that the array is of length `num_elements`. Also, are the offsets in increasing order to make it easy to figure out the size of each element?
Yes, implementation wise it will be in increasing order.  Changed it to " array of uint32_t of length num_elements"

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+ * collaterals (if any collaterals are provided).
+ * @param[out] parsed_report Optional **oe_report_t** structure to populate with
+ * the report properties in a standard format.
+ *
+ * @retval OE_OK The report was successfully created.
+ * @retval OE_INVALID_PARAMETER At least one parameter is invalid.
+ *
+ */
+oe_result_t oe_verify_remote_report_with_collaterals(
+#ifndef OE_BUILD_ENCLAVE
```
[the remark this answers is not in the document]
jazzybluesea: I thought we agreed that we need to support verification functions inside an enclave as well.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+
+    uint8_t* qe_id_info;                    ///< QE Identity info
+    size_t qe_id_info_size;                 ///< QE Identity size
+    uint8_t* qe_id_issuer_chain;            ///< PEM format
+    size_t qe_id_issuer_chain_size;         ///< Size of qe_id_issuer_chain
+
+    uint8_t* creation_datetime;             ///< Time the collaterals were generated
+    size_t create_datetime_size;            ///< The size of creation_datetime.  Should be 24.
+
+} sgx_collaterals;
```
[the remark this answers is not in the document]
jazzybluesea: oe_collaterals_t is parsed and converted to the internal structure sgx_collaterals.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+ *
+ */
+typedef enum _oe_collaterals
+{
+    uint32_t version;
+    uint8_t enclave_type;         ///< The type of enclave
+    uint8_t reserved;
+    uint16_t reserved2;
+    uint32_t num_elements;
+    size_t buffer_size;           ///< Size of the buffer
```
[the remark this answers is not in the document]
jazzybluesea: Done.

docs/DesignDocs/RemoteAttestationCollaterals.md
dthaler: > collaterals apply only to remote attestation and not local attestation.

I don't think that's true for endorsements in general, especially for CyRes where there can be multiple layers, not just 2, in the attestation chain.

docs/DesignDocs/RemoteAttestationCollaterals.md
dthaler: WHy not just use the term Verifier then? :)

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+    OE_REPORT_FLAGS_REMOTE_ATTESTATION,
+    NULL, // report_data
+    0,
+    NULL, // opt_params
+    0,
+    &report,
+    report_size);
+```
+
+##### Challenger verifies the evidence (in the untrusted host)
```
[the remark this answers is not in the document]
jazzybluesea: Dave had the same comment. Already updated comment.

docs/DesignDocs/RemoteAttestationCollaterals.md
dthaler: Sounds good to me

docs/DesignDocs/RemoteAttestationCollaterals.md
dthaler: I like that idea, as long as those parameters are optional

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: I am not familiar with the "user generated policy".  Maybe we can handle that as a separate issue/PR?

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Same as above.  I am not familiar with the "user generated policy". Maybe we can handle that as a separate issue/PR?

docs/DesignDocs/RemoteAttestationCollaterals.md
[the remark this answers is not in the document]
jazzybluesea: Resolved offline.  No need to pass in enclave.

docs/DesignDocs/RemoteAttestationCollaterals.md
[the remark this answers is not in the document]
jazzybluesea: I will have a function that does the conversion.  Users will be able to use it as a reference.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: What you are asking can be supported using the upcoming plug-in model.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: What you are asking can be supported using the upcoming plug-in model.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Done.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Sounds good.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Fixed.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: I agree is redundant.  I was thinking from a hw point of view. :)

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: I am just trying to state is a raw buffer of bytes.  I rather deal with bytes then 4 bytes words.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Yes it should allow multiple independent structs.  With the plug-in model, you will be able to define your own set of attestation functions and define your endorsement data format.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Thanks.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,292 @@
...
+result = oe_get_report(
+    OE_REPORT_FLAGS_REMOTE_ATTESTATION,
+    NULL, // report_data
+    0,
+    NULL, // opt_params
+    0,
+    &report,
+    report_size);
+
+result = oe_get_collaterals(
```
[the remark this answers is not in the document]
jazzybluesea: We will need differentiate between collaterals that are from OE SDK and collaterals from a different source.  Will address this later.

docs/DesignDocs/RemoteAttestationCollaterals.md
soccerGB: "QE identity info" in the context of endorsement here is not machine dependent

docs/DesignDocs/RemoteAttestationCollaterals.md
soccerGB: @jiria 
We need to maintain backward compatible for oe_verify_report() API at least for one past public release. So changing the signature of oe_verify_report() is not part of this design plan.

At OE SDK, we don't directly support "Policy" information in our APIs, instead, we provide API channels for an involved entity to pass "blob" of data to another in the attestation stack.  In upcoming attestation plug-in model PR (with design doc, it will be posted after this one is in because it depends on collaterals APIs), it provides more customizable way for a plug-in to funnel additional data (such as policy data) to a verifier.

@dthaler

We could like to make it support both remote and local attestation. Could you share with us on the sample endorsement data in CyRes case?

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: My mistake. oe_verify_remote_report() is part of the recently added host_verify library.  A standalone library that can be used to verify OE reports.  The plan is to update this function regardless.  For the OE SDK library, users will need to use oe_verify_report_with_endorsements().  Updating oe_verify_report() is not an option since it will break existing users.

docs/DesignDocs/RemoteAttestationCollaterals.md
dthaler: I thought oe_verify_report() was considered "experimental" and so could be broken.  But if not, then agree we should just mark oe_verify_report as deprecated, and add a new function alongside it that's not deprecated.

Regarding cyres data, JiriA can probably provide more details than I can, but concepts like PCK cert chain (e.g., where the manufacturer vounches that it manufactured a device) and CRL would apply equally to cyres.

docs/DesignDocs/RemoteAttestationCollaterals.md
dthaler: even better in my view would be to deprecate (not remove) oe_verify_report and add a new function named something like oe_verify_evidence(), i.e., with that name instead of the name oe_verify_report_with_endorsements, since "report" is an Intel term not an IETF term so would be better to not appear in the API name or structure names going forward.

soccerGB: @dthaler
oe_verify_evidence() is a great idea.

Using standard naming is easier to explain their functionalities.

sylvanc:
Hi @soccerGB - based on our previous conversations, I would expect a few things to be different. Perhaps you can clarify why they are not accounted for here?

1. User experience 1 should, I think, not be present. We previously discussed removing this, and instead relying on user experience 3 plus the ability to explicitly fetch collateral from a specific source. Those fetch functions are not present in this PR.
2. Previously, we discussed that the quote generation call should also return the endorsements used to generate that quote, rather than having a separate call, to ensure that the precise endorsements used to generate a quote could always be presented. This PR appears to introduce a possible consistency problem.
3. This PR uses an input validation time. Previously, we discussed that the `oe_verify_report_with_endorsements` should return a date range for which the verification would be successful.

Thanks.

docs/DesignDocs/RemoteAttestationCollaterals.md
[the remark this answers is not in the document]
jazzybluesea: My assumption here was that the audience is familiar with attestation.  I will write a brief overview/introduction.

jazzybluesea:
> 3. This PR uses an input validation time. Previously, we discussed that the `oe_verify_report_with_endorsements` should return a date range for which the verification would be successful.
A date range will be returned as part of the claims output from doing the verification.  Please wait for more details once that design is out.

soccerGB:
@sylvanc  thanks for your feedback.

The updated design should already address your first and second comments. The endorsement is part of the return from oe_get_evidence(...). Thanks for this great feedback.

The valid date range will be returned as part of the claims in the oe_verify_evidence.

For the flexible runtime control of platform DCAP provider, it requires changes in Intel's DCAP library, we are actively working Intel to have them support it in their DCAP library. Once they deliver it, it should not take long to support that in OE SDK.

sylvanc:
Hi @soccerGB - the change to `oe_get_evidence` is good, and covers my second point, thanks. Removing the "autofetch" functionality from `oe_verify_evidence` is also good, and covers part of my first point.

The other part of my first point is that we will need a separate API for fetching endorsements from a specific source. For example, the verifier may choose to fetch current SGX collateral to compare the result of `oe_verify_evidence` when called with the endoresements provided and when called with external endorsements. This should also allow fetching, for example, cached SGX collateral from a non-Intel source.

For the date range, are there any notes on the design that encodes the range in the returned claims, for example what the API to examine that date range looks like from the caller's point of view? Will the `input_validation_time` parameter be removed from `oe_verify_evidence`?

docs/DesignDocs/RemoteAttestationCollaterals.md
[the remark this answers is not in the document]
jazzybluesea: As part of the verification, the identity of the enclave (in SGX this is the QE identity info) in the endorsement is used to verify the evidence. This already a check that is being done as part of verification.  Furthermore, in the SGX case you need to get the quote/evidence before you can get the collaterals/endorsements because the CRL depend on the PCK Cert that is embedded in the quote.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: I updated the design with oe_get_evidence() and a oe_verify_evidence() that will support both local and remote Attestation.

soccerGB:
@sylvanc 

"a separate API for fetching endorsements from a specific source"
It was part of our plan to have such an API, per our discussion before. However, it's blocked by Intel's changes to their SGX DCAP library for removing hardcoded file name. Once Intel delivers it, we will make necessary changes to expose this new API. 
 
"For the date range, are there any notes on the design that encodes the range in the returned claims, for example what the API to examine that date range looks like from the caller's point of view? Will the input_validation_time parameter be removed from oe_verify_evidence?"

the claims mentioned above are an array of (name, value) pairs.
Akash will post a design doc for the attestation plugin, which has more details information on it.

We have no plan to remove the  input_validation_time in our API ,we need this parameter for fully use of upcoming QvE feature in their DCAP library.

common/sgx/endorsements.c
```diff
@@ -0,0 +1,378 @@
...
+            OE_INVALID_PARAMETER, "Endorsement buffer size is invalid.");
+    data_size = endorsements->buffer_size - offsets_size;
+    offsets = (uint32_t*)endorsements->buffer;
+    data_ptr_start = (uint8_t*)(endorsements->buffer + offsets_size);
+
+    memset(sgx_endorsements, 0, sizeof(oe_sgx_endorsements_t));
+
+    for (int i = 0; i < OE_SGX_ENDORSEMENT_COUNT; i++)
+    {
+        uint8_t* item_ptr = data_ptr_start + offsets[i];
```
dthaler: Is this function only ever called by the original enclave?  Or can it be called by the relying party / verifier?  If the latter, this is not safe, since the receiver could be on a big-endian platform/architecture whereas the SGX enclave is little-endian.   So using offsets[i] is not safe without going through an ntohl() check.

Ok to handle this in a future PR if you file a bug to track.

common/sgx/endorsements.h
```diff
@@ -0,0 +1,114 @@
...
+
+/*! \struct oe_sgx_endorsements
+ *
+ * \brief SGX endorsements structure
+ *
+ * The generic oe_endorsements_t structure is parsed and converted into this
+ * internal structure.  The order of the generic data elements should
+ * coincide with the order of the fields in this structure.
+ *
+ * Data format: All data comes from the Data Center Attestation Primitives(DCAP)
```
dthaler: add space before (

common/sgx/endorsements.h
```diff
@@ -0,0 +1,114 @@
...
+ * Client.
+ *
+ * For Azure DCAP Client
+ * (https://github.com/microsoft/Azure-DCAP-Client/blob/master/src/dcap_provider.h)
+ * see **sgx_ql_revocation_info_t** and sgx_qe_identity_info_t.
+ *
+ * For Intel DCAP Client
+ * (https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/README.md)
+ * see TBD.
```
dthaler: Is there an issue # filed to track filling in this TBD?

common/sgx/quote.c
```diff
@@ -481,13 +481,21 @@ oe_result_t oe_verify_quote_internal_with_collaterals(
...
             OE_CHECK_MSG(
                 oe_datetime_from_string(
-                    collaterals_body->creation_datetime,
-                    sizeof(collaterals_body->creation_datetime),
+                    (const char*)sgx_endorsements
+                        ->items[OE_SGX_ENDORSEMENT_FIELD_CREATION_DATETIME]
+                        .data,
+                    sgx_endorsements
+                        ->items[OE_SGX_ENDORSEMENT_FIELD_CREATION_DATETIME]
+                        .size,
```
dthaler: If I understand right, since the verify function can be called on a machine that is not an SGX machine, this code is not safe... if it's run on a big endian machine, the .size value will be backwards since you didn't go through htonl (when sending) and ntohl (on receipt) or similar transform.   In short, when writing cross-platform code that parses messages sent across a network, it's never safe to simply cast a buffer to a struct that contains integers larger than 8 bits.

Ok to address in a separate PR if you file an issue to track this.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+
+In the current API, OE SDK has functions to get the evidence and to verify
+the evidence.  This document describes additional changes to the API to
+support the concept of endorsements. Endorsements are additional information
+from a second source that the verifier can use for attestation verification.
+
+Terminology
+----------
+From Remote Attestation Procedures Architecture
+(https://www.ietf.org/id/draft-birkholz-rats-architecture-02.txt).
```
dthaler: Should note that that draft is a Work in Progress and subject to change.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+### Claims
+As part of the claims form `oe_verify_evidence()`, there will be a validity datetime range that applies to the evidence and endorsements.  For more information see the design document `CustomAttestation.md`.
+
+### OE Host Verify Library
+The OE Host Verify library is a standalone library used for verifying remote reports outside
+the TEE/enclave. The function `oe_verfiy_remote_report()` will be deprecated and should use the
+upcoming plug-in mode to do verification.
+
+
+Alternates
```
dthaler: Did you mean "Alternatives"?  Or "Alternate Designs Considered"?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+    uint8_t** claims,
+    size_t* claims_size);
+```
+
+### Claims
+As part of the claims form `oe_verify_evidence()`, there will be a validity datetime range that applies to the evidence and endorsements.  For more information see the design document `CustomAttestation.md`.
+
+### OE Host Verify Library
+The OE Host Verify library is a standalone library used for verifying remote reports outside
+the TEE/enclave. The function `oe_verfiy_remote_report()` will be deprecated and should use the
```
dthaler: "and should" -> "and apps should"

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+upcoming plug-in mode to do verification.
```
dthaler: "mode" -> "model"

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+
+
+Alternates
+----------
+
+### Endorsement structure:
+- Multiple serializable structures were considered.  At the end the final design
+made sense given that it is generic and support serialization.  We considered
+supporting multiple flavors of the structure, a binary format and another for supporting
+user defined endorsements.  But we ended up deciding to have the user build the binary
```
dthaler: "user" -> "app"

tests/tools/oecert/host/host.cpp
```diff
@@ -195,53 +193,47 @@ static oe_result_t _gen_report(
...
-                }
-                fwrite(collaterals, collaterals_size, 1, col_fp);
-                fclose(col_fp);
-                printf("collaterals_size = %zu\n", collaterals_size);
+                printf("ERROR: Failed to get collaterals\n");
+                goto exit;
             }
-            else
+
+            // TODO: Current collateral structure is not self contained.
```
dthaler: Is there an Issue # filed to track this TODO?  (If so, would be nice to reference the # in the code comment)

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+    size_t evidence_buffer_size,
+    const uint8_t* endorsements,
+    size_t endorsements_size,
+    oe_datetime_t* input_validation_time,
+    uint8_t** claims,
+    size_t* claims_size);
+```
+
+### Claims
+As part of the claims form `oe_verify_evidence()`, there will be a validity datetime range that applies to the evidence and endorsements.  For more information see the design document `CustomAttestation.md`.
```
CodeMonkeyLeet: Claims are a key part of this API design, but is not included in this PR? And we're checking in an implementation of this at the same time as the design approval?

Please refactor this PR apart so the design can be evaluated as a whole, not as a partial design + implementation.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+
+### Endorsement structure:
+- Multiple serializable structures were considered.  At the end the final design
+made sense given that it is generic and support serialization.  We considered
+supporting multiple flavors of the structure, a binary format and another for supporting
+user defined endorsements.  But we ended up deciding to have the user build the binary
+structure instead, to avoid complexity.
+
+### APIs:
+- To reduce complexity, `oe_get_evidence()` is only available in the enclave.
```
CodeMonkeyLeet: Again, what complexity? Are we removing the existing `oe_get_report` from the host side if `oe_get_evidence()` is supposed to supersede it? What about trying to get local attestation targets?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+the TEE/enclave. The function `oe_verfiy_remote_report()` will be deprecated and should use the
+upcoming plug-in mode to do verification.
+
+
+Alternates
+----------
+
+### Endorsement structure:
+- Multiple serializable structures were considered.  At the end the final design
+made sense given that it is generic and support serialization.  We considered
```
CodeMonkeyLeet: Who's "we"? The point of this design process is to have the design for public scrutiny. This statement provides no information on the actual alternatives that were considered, nor the reasons for their rejection. A proposal is not required to present alternatives and if it's not going to actually provide actionable information, I would much rather this section be dropped as it reads like an attempt to forestall discussion of design alternatives.

docs/DesignDocs/RemoteAttestationCollaterals.md
CodeMonkeyLeet: Typo: `oe_verify_remote_report()`.

Again, alluding to a future design does not convince me of the necessity of the deprecation in this one.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+
+User Experience
+---------------
+
+There are 2 scenarios.
+
+### 1. Verifier is provided with endorsements:
+In this scenario the attester/asserter provides the evidence and endorsements to the verifier.  The verifier is then free to use these to verify the TEE.
+
+##### Attester generates the evidence and endorsements (inside an enclave/TEE)
```
CodeMonkeyLeet: This is not strictly true today; the DCAP client runs external to the enclave and the "endorsements" are not generated in that environment (nor required to be for confidentiality)

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+...
+result = oe_get_evidence(
+    OE_EVIDENCE_FLAGS_REMOTE_ATTESTATION,
+    NULL, // custom_claims
+    0,
+    NULL, // opt_params
+    0,
+    &evidence,
+    &evidence_size,
+    &endorsements,
```
CodeMonkeyLeet: (@dthaler for RATS insight)

Based on a reading of the RATS draft, the mapping of this to "endorsements" doesn't seem correct given the 4.3.2 section definition on Role Messages (as opposed to the intro section copied above):

> Endorsements:  Claims about immutable and implicit characteristics of
      the Attesting Computing Environment.  Typically, endorsement
      Claims are created by manufacturing or supply chain entities.

If the report/quote is the Evidence, then arguably, the only artifact in DCAP that could be construed to be the endorsement would be the PCK certificate, with the claim being "the attester with this key is recognized by the manufacturer", and that's not even part of the endorsements data structure defined here. Everything else here is actually the set of Known-Good-Values (literally, the QE ID and TCBInfo list the good values of claims like the SVN values in the evidence). 

I bring this up because if the mapping to endorsements is important, i.e. we want to support a model where the verifier can actually pull the PCK cert, then this iteration on E3 does not actually support that. It _could_ be supported in future versions of the SGX platform, and you should consider the versioning impact of this; in fact, is something we need to consider in the support of EPID attestation, where the actual endorsement comes sans all the KGV collateral on the verification side.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,406 @@
...
+Trusted Execution Environment (TEE) platform (HW + software).
+
+In the current API, OE SDK has functions to get the evidence and to verify
+the evidence.  This document describes additional changes to the API to
+support the concept of endorsements. Endorsements are additional information
+from a second source that the verifier can use for attestation verification.
+
+Terminology
+----------
+From Remote Attestation Procedures Architecture
```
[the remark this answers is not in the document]
jazzybluesea: done.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+    uint32_t crl_issuer_chain_size[OE_SGX_ENDORSEMENTS_CRL_COUNT];
+
+    uint8_t* qe_id_info;                    ///< QE Identity info, null-terminated JSON string
+    uint32_t qe_id_info_size;               ///< QE Identity size
+    uint8_t* qe_id_issuer_chain;            ///< PEM format, null-terminated string
+    uint32_t qe_id_issuer_chain_size;       ///< Size of qe_id_issuer_chain
+
+    uint8_t* creation_datetime;             ///< Time the endorsements were generated,
+                                            ///< null-terminated string
+    uint32_t create_datetime_size;          ///< The size of creation_datetime.
```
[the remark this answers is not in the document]
jazzybluesea: I actually implemented a little differently. Will update the doc.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+
+    ///< Size of each crl_issuer_chain
+    uint32_t crl_issuer_chain_size[OE_SGX_ENDORSEMENTS_CRL_COUNT];
+
+    uint8_t* qe_id_info;                    ///< QE Identity info, null-terminated JSON string
+    uint32_t qe_id_info_size;               ///< QE Identity size
+    uint8_t* qe_id_issuer_chain;            ///< PEM format, null-terminated string
+    uint32_t qe_id_issuer_chain_size;       ///< Size of qe_id_issuer_chain
+
+    uint8_t* creation_datetime;             ///< Time the endorsements were generated,
```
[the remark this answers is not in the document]
jazzybluesea: Yes this value is untrusted.  When doing oe_verify_evidence() the user has the choice of provide a timestamp to check validity.  If one is not provided, it will fall back to use this value.  This would allow the verifier to audit a quote + endorsements at some point in the past.  They have the choice to store the timestamp outside of the endorsements and pass it in when calling oe_verify_evidence().

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+ * Data format: All data comes from the Data Center Attestation Primitives(DCAP)
+ * Client.
+ *
+ * For Azure DCAP Client
+ * (https://github.com/microsoft/Azure-DCAP-Client/blob/master/src/dcap_provider.h)
+ * see **sgx_ql_revocation_info_t** and sgx_qe_identity_info_t.
+ *
+ * For Intel DCAP Client
+ * (https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/README.md)
+ * see TBD.
```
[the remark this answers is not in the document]
jazzybluesea: Will create a issue to track this.

include/openenclave/bits/attestation.h
```diff
@@ -0,0 +1,90 @@
...
+///< Version of the supported SGX endorsement structures
+#define OE_SGX_ENDORSEMENTS_VERSION (1)
+
+///< Number of CRLs in the SGX endorsements
+#define OE_SGX_ENDORSEMENTS_CRL_COUNT (2)
+
+/*! \enum oe_sgx_endorsements_fields
+ *
+ * Specifies the order of the SGX endorsements fields stored in
+ * the oe_endorsements_t strcuture
```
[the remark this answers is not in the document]
jazzybluesea: Thanks.

include/openenclave/bits/attestation.h
```diff
@@ -0,0 +1,90 @@
...
+    OE_SGX_ENDORSEMENT_FIELD_CRL_ISSUER_CHAIN_PCK_CERT,
+    OE_SGX_ENDORSEMENT_FIELD_CRL_ISSUER_CHAIN_PCK_PROC_CA,
+    OE_SGX_ENDORSEMENT_FIELD_QE_ID_INFO,
+    OE_SGX_ENDORSEMENT_FIELD_QE_ID_ISSUER_CHAIN,
+    OE_SGX_ENDORSEMENT_FIELD_CREATION_DATETIME,
+    OE_SGX_ENDORSEMENT_COUNT
+
+} oe_sgx_endorsements_fields;
+
+#endif /* _OE_BITS_ATTESTATION_H */
```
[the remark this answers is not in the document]
jazzybluesea: Fixed.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+ *
+ */
+typedef struct _oe_endorsements_t
+{
+    uint32_t version;
+    uint32_t enclave_type;        ///< The type of enclave
+    uint64_t buffer_size;         ///< Size of the buffer  (oe_enclave_type_t)
+    uint32_t num_elements;
+
+    ///< Data buffer is made of an offset array of type uint32_t, followed by
```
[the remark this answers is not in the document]
jazzybluesea: Fixed.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+            0,
+            &claims,
+            &claims_size);
+...
+```
+
+Specification
+-------------
+
+### <B>Public</B> type definitions
```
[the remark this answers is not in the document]
jazzybluesea: Removed.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+
+///< Number of CRLs in the SGX endorsements
+#define OE_SGX_ENDORSEMENTS_CRL_COUNT     (2)
+
+/*! \enum oe_sgx_endorsements_fields
+ *
+ * Specifies the order of the SGX endorsements fields stored in
+ * the oe_endorsements_t strcuture
+ */
+typedef enum _oe_sgx_endorsements_fields
```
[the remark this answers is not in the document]
jazzybluesea: This is public in the scenario where an API user wants to build his/her own version of the SGX endorsements.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,457 @@
...
+    uint8_t* creation_datetime;             ///< Time the endorsements were generated,
+                                            ///< null-terminated string
+    uint32_t create_datetime_size;          ///< The size of creation_datetime.
+
+} oe_sgx_endorsements_t;
+```
+
+### New <B>Public</B> Attestation functions
+
+These functions supersede the existing functions:
```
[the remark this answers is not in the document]
jazzybluesea: I added something like this: "Users should start using these new functions.  `oe_get_report()` and `oe_verify_report()` are deprecated and will be removed in future releases."

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: The ask was to only support oe_get_evidence() in the enclave.  If this function needs to be exposed to the host for local attestation, which as you pointed out it is required, then we will just have to expose the new function to the host as well.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: True.  I just meant to highlight that it was initiated inside the enclave.  In the future we will be able to do all this inside the enclave.  I can rephrase.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: I will leave it the other design to mention it since it is more appropriate.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Removed.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: My plan is to create a separate PR.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,406 @@
...
+the evidence.  This document describes additional changes to the API to
+support the concept of endorsements. Endorsements are additional information
+from a second source that the verifier can use for attestation verification.
+
+Terminology
+----------
+From Remote Attestation Procedures Architecture
+(https://www.ietf.org/id/draft-birkholz-rats-architecture-02.txt).
+
+- Evidence is provable Claims about a specific Computing Environment
```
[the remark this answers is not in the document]
jazzybluesea: I will just leave it as it is...

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: This section was removed...

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Added.

common/sgx/endorsements.h
jazzybluesea: Updated to have the intel structure name.  It was just updated yesterday!

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: I will leave it to the plug-in design doc to deprecate this function.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: Deprecation statement removed.

docs/DesignDocs/RemoteAttestationCollaterals.md
jazzybluesea: This section was removed.

tests/tools/oecert/host/host.cpp
jazzybluesea: Good idea.  Will keep that in mind.  This is still WIP.

common/sgx/endorsements.c
```diff
@@ -0,0 +1,378 @@
...
+        oe_free(endorsements);
+
+    OE_TRACE_INFO(
+        "Exit call %s: %d(%s)\n", __FUNCTION__, result, oe_result_str(result));
+
+    return result;
+}
+
+/**
+ * Convert a oe_endorsement_t structure to a SGX endorsement structure
```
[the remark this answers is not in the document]
jazzybluesea: Fixed.

common/sgx/endorsements.c
jazzybluesea: Is this also an issue with the current oe_report_t?

docs/DesignDocs/RemoteAttestationCollaterals.md
dthaler: The endorsement could be provided to the verifier in a way that is bundled with the evidence.   (E.g., if the PCK cert chain was provided with the quote, or inside the quote).   So just because it's in the quote/report doesn't mean it's not endorsement.   The question is whether the "it" is processor specific, or applies to all equivalent processors.  So a CRL in that sense could be in the endorsement category, as could anything else that is processor-independent.

common/sgx/endorsements.c
dthaler: Haven't looked but I would assume yes it's already a problem, which is why I'm fine with handling it in a separate issue/PR.

dthaler: I just verified via code inspection, and filed issue #2266 to track this.

docs/DesignDocs/RemoteAttestationCollaterals.md
CodeMonkeyLeet: @dthaler, I'm not suggesting that the PCK cert (chain) is not an endorsement. I'm saying that of all the SGX attestation artifacts, it's the one thing that _best fits_ that description. I'm pointing out that if the PCK cert (chain) is an Endorsement, then Scenario 2 proposed in this design is explicitly not supported in the current SGX implementation, since it cannot be correctly retrieved by the verifier. It is a limitation of the Coffeelake platform and this design needs to address how this inconsistency in the abstraction is conveyed to a developer.

I'm also asking if the distinction between KGV and Endorsements is meaningful in the context of this design (or even in the RATS draft). Particularly since the use of RATS terminology was suggested by you, I'm unclear on why the choice of "collaterals" as a term that could encompass both KGV and Endorsements is less preferred than a use of "endorsements" that overloads an existing meaning distinct from KGV.

dthaler: > @dthaler, I'm not suggesting that the PCK cert (chain) is not an endorsement. I'm saying that of all the SGX attestation artifacts, it's the one thing that best fits that description.

Why would a CRL not fit that description?  (please educate me)

> I'm pointing out that if the PCK cert (chain) is an Endorsement, then Scenario 2 proposed in this design is explicitly not supported in the current SGX implementation, since it cannot be correctly retrieved by the verifier. It is a limitation of the Coffeelake platform and this design needs to address how this inconsistency in the abstraction is conveyed to a developer.

Thanks for the info, I'm sure there's lots of platform limitations in different platforms.   Many of the assumptions SGX has will probably not apply in general, so it's not particularly surprising that a general assumption may not apply to current SGX either.

> I'm also asking if the distinction between KGV and Endorsements is meaningful in the context of this design (or even in the RATS draft). 

The RATS discussion is still in flux.  My opinion is that KGV as a term is not a helpful term, as it implies an assumption that there is such a thing, as opposed to "security policy" for which KGV is an example, but there may be more complex logic like known good ranges, or criteria like value of claim X must be > value of claim Y, etc.   The rats arch doc is not a WG item yet because all these sorts of discussions are going on.   https://tools.ietf.org/html/draft-thaler-rats-architecture-00 is my take at writing up a proposal that some believe might be easier to understand.

> Particularly since the use of RATS terminology was suggested by you, I'm unclear on why the choice of "collaterals" as a term that could encompass both KGV and Endorsements is less preferred than a use of "endorsements" that overloads an existing meaning distinct from KGV.

If you want, I can take the suggestion (or you can) to the RATS WG to use the term "collaterals" in the way you suggest.  I just want to align between OE and RATS, and either or both might change terms.

common/sgx/endorsements.c
jazzybluesea: Thank you Dave.

common/sgx/quote.c
jazzybluesea: See #2266

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,488 @@
...
+#### Attester:
+An Attestation Function that creates Evidence by
+collecting, formatting and protecting (e.g., signing) Claims.  It
+presents Evidence to a Verifier using a conveyance mechanism or
+protocol.
+
+#### Verifier:
+An Attestation Function that accepts Evidence from an Attester using
+a conveyance mechanism or protocol.  It also accepts Known-Good-Values
+and Endorsments from an Asserter using a conveyance mechanism or protocol.
```
jiria: typo Endorsements

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,488 @@
...
+It verifies the protection mechanisms, parses and appraises Evidence
+according to good-known valid (or known-invalid) Claims and Endorsments.
```
jiria: Endorsements

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,488 @@
...
+It produces Attestation Results that are formatted and protected (e.g.,
+signed).  It presents Attestation Results to a Relying Party using
```
jiria: double space

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,488 @@
...
+    NULL, // opt_params
+    0,
+    &evidence,
+    &evidence_size,
+    &endorsements,
+    &endorsements_size);
+...
+```
+
+##### Verifier verifies the evidence and endorsements (in the untrusted host or inside an enclave/TEE)
```
jiria: nit: in an untrusted host?

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,488 @@
...
+     * |-----------------------|
+     * |  buffer (data)        |
+     * |_______________________|
+     */
+    uint8_t buffer[];              ///< Buffer of offsets + data
+
+} oe_endorsements_t;
+
+/*! Version of the supported SGX endorsement structures */
+#define OE_SGX_ENDORSEMENTS_VERSION     (1)
```
jiria: Consider moving SGX specific definitions outside of the TEE agnostic header.

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,488 @@
...
+| signer_id        | uint8_t[32]        | The signer ID for the enclave (MRSIGNER for SGX).                    |
+| product_id       | uint8_t[32]        | The product ID for the enclave (ISVPRODID for SGX).                  |
+| validity_from    | oe_datetime_t      | Overall datetime from which the evidence and endorsements are valid. |
+| validity_until   | oe_datetime_t      | Overall datetime at which the evidence and endorsements expire.      |
+
+
+### OE Host Verify Library
+
+The OE Host Verify library is a standalone library used for verifying remote reports outside
+the TEE/enclave. The function `oe_verfiy_remote_report()` will be updated to support
```
jiria: typo: verify

docs/DesignDocs/RemoteAttestationCollaterals.md
```diff
@@ -0,0 +1,488 @@
...
+Current set of claims definitions:
+
+| Claim Name       | Claim Value Type   | Description                                                          |
+|:-----------------|:-------------------|:---------------------------------------------------------------------|
+| id_version       | uint32_t           | Claims version. Must be 0                                            |
+| security_version | uint32_t           | Security version of the enclave. (ISVN for SGX).                     |
+| attributes       | uint64_t           | Attributes flags for the evidence: <br/> `OE_REPORT_ATTRIBUTES_DEBUG`: The evidence is for a debug enclave.<br/> `OE_REPORT_ATTRIBUTES_REMOTE`: The evidence can be used for remote attestation.   |
+| unique_id        | uint8_t[32]        | The unique ID for the enclave (MRENCLAVE for SGX).                   |
+| signer_id        | uint8_t[32]        | The signer ID for the enclave (MRSIGNER for SGX).                    |
+| product_id       | uint8_t[32]        | The product ID for the enclave (ISVPRODID for SGX).                  |
```
jiria: Please have a look at https://github.com/openenclave/openenclave/pull/2262#discussion_r337815502 and at https://github.com/openenclave/openenclave/pull/2262#discussion_r337815883 and let me know what you think.

docs/DesignDocs/RemoteAttestationCollaterals.md
gupta-ak: fixed

docs/DesignDocs/RemoteAttestationCollaterals.md
gupta-ak: fixed

docs/DesignDocs/RemoteAttestationCollaterals.md
gupta-ak: i moved them

docs/DesignDocs/RemoteAttestationCollaterals.md
gupta-ak: I responded in the #2262 PR thread.

PR #2254, opened 2019-10-14
Remove lockless queue implementation

radhikaj:
This PR: https://github.com/openenclave/openenclave/pull/1838/commits added the lockless queue intended to be used for the switchless features. Our current switchless ocalls dont need this queue. Removing it for the time being. We can re-introduce it as needed later.


radhikaj:
> #1838 also added tests/lockless_queue/enc folder with two files that are not reverted as part of this changeset (although they are rendered inert). Can you make sure they are removed as well?

Thanks @CodeMonkeyLeet . I have removed tests/lockless_queue/enc

radhikaj:
Thanks. I squashed the commits.

PR #2261, opened 2019-10-17
Update packaging build for HAS_QUOTE_PROVIDER

johnkord:
We missed just one USE_LIBSGX reference :) This should finish them off! grep returns nothing else for USE_LIBSGX at least!


johnkord:
Thanks all!

bors r+

PR #2262, opened 2019-10-17
Attestation Plugin Design Doc

gupta-ak:


docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+attestation. The two key functions are `oe_get_report`, which produces an opaque
+blob that is signed by the enclave, and `oe_verify_report`, which can be used to
+verify the generated report. The original purpose of those two APIs were to
+provide a simple, cross-platform way to produce and verify attestation data.
+
+However, some developers need more flexibility for their attestation
+requirements. For example, one might want to extend Open Enclave's
+current attestation structures with extra information, such as geolocation
+or a timestamp. Another user might want their enclaves to generate attestation
+data that is in a compatible format with their existing authentication
```
dthaler: grammar: "compatible format with" -> "format compatible with"

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+    enclave. The enclave must attest to the relying party before the
+    relying party can trust it. The relying party can also play the role
+    of the verifier, but it does not necessarily have to.
+
+Specification
+-------------
+
+To support custom attestation formats, this document proposes adding a plugin
+model for attestation. The Open Enclave SDK will define a set of common APIs
+that each plugin must implement. Each plugin will define an UUID to distinguish
```
dthaler: grammar: "an UUID" -> "a UUID"

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+struct oe_claim_t
+{
+    char* name;
+    uint8_t* value;
+    size_t value_size;
+};
+
+/**
+ * Struct that defines the structure of each plugin. Each plugin must
+ * define an UUID for its format and implement the functions in this
```
dthaler: "an UUID" -> "a UUID"

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+the [endorsements design doc](RemoteAttestatonCollaterals.md). There will also
+be functions for registering and unregistering plugins called
+`oe_register_attestation_plugin` and `oe_unregister_attestation_plugin`. The user
+can statically link in their desired plugin and call the register plugin function.
+The attestation data can be retrieved from the "plugin aware" analogue of
+`oe_get_report` with the desired UUID. The generated data will have the UUID in
+its header. The user can call the analogue of `oe_verify_report` to verify the
+data and the Open Enclave runtime can use this UUID to determine what plugin
+verification routine to run.
+
```
dthaler: Explain whether the plugin is loaded into, and called by, the host side or the enclave side of the app.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+     * must be attached to the evidence and then cryptographically signed.
+     *
+     * @param[in] plugin_context A pointer to the attestation plugin struct.
+     * @param[in] flags Specifying default value (0) generates evidence for local
+     * attestation. Specifying OE_EVIDENCE_FLAGS_REMOTE_ATTESTATION generates
+     * evidence for remote attestation.
+     * @param[in] custom_claims The optional custom claims list.
+     * @param[in] custom_claims_length The number of custom claims.
+     * @param[in] opt_params The optional plugin-specific input parameters.
+     * @param[in] opt_params_size The size of opt_params.
```
dthaler: "… in bytes." (same for lines 147 and 151)  For consistency with line 195, 197, etc.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+        const oe_datetime_t* input_validation_time,
+        oe_claim_t** claims,
+        size_t* claims_length);
+};
+```
+
+Here is the rationale for each element in the plugin struct:
+
+- `format_id`
+  - Each plugin needs an unique identifier to distinguish itself.
```
dthaler: grammar: "an unique" -> "a unique"

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+  - A plugin might require some setup or teardown when it is registered or
+    unregistered, so these functions are required. Furthermore, a plugin
+    might require configuration, which is why there is a `config_data`
+    parameter. The configuration data can be plugin specific, so no format is
+    specified in this proposal.
+- `get_evidence`, `free_evidence`, and `free_endorsements`
+  - Producing evidence and endorsements is necessary for attestation.
+  - `flags` field to determine local vs. remote attestation.
+  - There is a `custom_claims` parameter because many attestation protocols
+    require the enclave to sign some claim from a relying party. For example,
```
dthaler: suggest "some claim" -> "some data", since a claim was defined as a name/value pair and it might not sign a name, just the value.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+    many protocols follow the "challenge response" architecture, which requires
+    the enclave to sign a nonce from the relying party.
+  - There an `opt_params` field because some plugins might require plugin
```
dthaler: grammar: "There an" -> "There is an"

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+    specific input. For example, the SGX local attestation needs the
+    other enclave's target info struct.
+  - There is an `endorsements` parameters to return the endorsements that are
```
dthaler: grammar: "an … parameters" -> "an … parameter"

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+    const oe_datetime_t* input_validation_time,
+    oe_claim_t** claims,
+    size_t* claims_length)
+{
+    OE_UNUSED(plugin_context);
+
+    /*
+     * Pseudocode description instead of actual C code:
+     *
+     * Call oe_verify_report will all the input parameters and get the oe_identity_t back.
```
dthaler: Can't parse "will all".  Seems like there's a word missing.  Maybe "will parse all"?

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+     * Look for the custom claims in the evidence header and extract them if found.
+     * Verify the hash of custom claims == report data field in evidence report.
+     * Convert oe_identity_t to the claims format.
+     */
```
dthaler: Another technical comment, which for this PR is just editorial, but may result in code changes elsewhere if needed:

Since the verify API may be running on a different machine with a different architecture than the one that generated the evidence, I think the /* comments */ should point out that the verify code needs to be smart about the byte order, so that if the SDK is compiled on a big-endian machine, it can still parse SGX evidence generated by a little-endian machine.  So one cannot, for example, simply cast to a struct and expect 32-bit int fields have the values you think they are.

I think this is important to call out here so that as people write more plugins, they don't make common mistakes.  (Either that or we have to redefine OE to not have a cross-plat goal and restrict it to only little-endian architectures, which I think runs counter to the goals of OE.)

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+    size_t value_size;
+};
+
+/**
+ * Struct that defines the structure of each plugin. Each plugin must
+ * define an UUID for its format and implement the functions in this
+ * struct. Ideally, each plugin should provide a helper function to
+ * create this struct on the behalf of the plugin users.
+ */
+struct oe_attestation_plugin_t
```
dthaler: Technical point (the only reason I'm not +1'ing the PR right now as the rest are comments that I would include with a +1):

Actually I'm just now realizing I don't understand why getting evidence and verifying evidence are in the same plugin.
An SGX plugin for getting evidence is only needed on an SGX machine (an SGX Attester).
An SGX plugin for verifying evidence is needed on every Verifier.
You can have SGX Attesters that aren't Verifiers, and you can have Verifiers that are not SGX Attesters.
By putting them in the same plugin, this increases the TCB unnecessarily.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+    return &sgx_attestation_plugin;
+}
+
+```
+
+### New Open Enclave APIs
+
+The functions are what the plugin user calls to use a plugin. They map almost
+exactly to the plugin API. The main difference is that `oe_get_evidence`
+require the UUID of the plugin as an input parameter.
```
dthaler: grammar: "require" -> "requires"

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+/**
+ * oe_register_attestation_plugin
+ *
+ * Registers a new attestation plugin and optionally configures it with plugin
+ * specific configuration data. The function will fail if the plugin UUID has
+ * already been registered.
+ * 
+ * This is available in the enclave and host.
+ *
+ * @param[in] plugin A pointer to the attestation plugin struct. Note that will
```
dthaler: can't parse grammar of "that will".  Maybe "that this will"?

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+ * @param[in] custom_claims_length The number of custom claims.
+ * @param[in] opt_params The optional plugin-specific input parameters.
+ * @param[in] opt_params_size The size of opt_params.
+ * @param[out] evidence_buffer An output pointer that will be assigned the
+ * address of the evidence buffer.
+ * @param[out] evidence_buffer_size A pointer that points to the size of the
+ * evidence buffer.
+ * @param[out] endorsements_buffer An output pointer that will be assigned the
+ * address of the endorsements buffer.
+ * @param[out] endorsements_buffer_size A pointer that points to the size of the
```
dthaler: "… in bytes" (same in 482 and 479), for consistency with rest of file

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+ *
+ * Verifies the attestation evidence and returns well known and custom claims.
+ * This is available in the enclave and host.
+ *
+ * @param[in] evidence_buffer The evidence buffer.
+ * @param[in] evidence_buffer_size The size of evidence_buffer in bytes.
+ * @param[in] endorsements_buffer The endorsements buffer.
+ * @param[in] endorsements_buffer_size The size of endorsements_buffer in bytes.
+ * @param[in] input_validation_time Optional datetime to use when verifying
+ * evidence. If not specified, it will use the creation_datetime of the
```
dthaler: For clarity suggest changing "If not specified, it" -> "If null, the function"

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,754 @@
...
+} oe_attestation_header_t;
+```
+
+### Backwards compatibility
+The new APIs should support verifying the old Open Enclave reports
+generated by `oe_get_report`. The `oe_attestation_header_t` structure
+shares the same 1st field (`uint32_t version`) as the old Open Enclave
+report header. Consequently, the `oe_verify_evidence` can use this
+information to decide if it needs to call a plugin or run the legacy
+verification routine (which is technically the same logic as the SGX plugin).
```
dthaler: lower case L here and in 584

docs/DesignDocs/CustomAttestation.md
soccerGB: This is a great point!

docs/DesignDocs/CustomAttestation.md
gupta-ak: Added an explanation.

docs/DesignDocs/CustomAttestation.md
gupta-ak: I added an explanation about the endianess for both the get_evidence and verify_evidence.

docs/DesignDocs/CustomAttestation.md
gupta-ak: I think it makes sense to separate them. For example, you could have a cluster with a bunch of TEEs producing their own evidence and you want your verifier service to handle all those types. In that case, the verifier would only care about the the `verify_evidence` part and the TEEs would only care about the `get_evidence` part.

We could have two plugins--one for getting evidence and one for verifier evidence. The UUID can be used to connect them.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,775 @@
...
+
+Terminology
+-----------
+
+This document uses the following terminology defined below. Note that
+these definitions are consistent with the terms defined in the
+[Remote Attestation Procedures (RATS)](https://datatracker.ietf.org/wg/rats/about/)
+working group.
+
+- Claims
```
letmaik: The term "claims" is not referenced in any of the other definitions. It feels like there is something missing to understand the full picture.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,775 @@
...
+
+To support custom attestation formats, this document proposes adding a plugin
+model for attestation. The Open Enclave SDK will define a set of common APIs
+that each plugin must implement. Each plugin will define a UUID to distinguish
+it from other plugins.
+
+Futhermore, there will be additional attestation "plugin aware" APIs that are
+analogous to `oe_get_report` and `oe_verify_report` called `oe_get_evidence`
+and `oe_verify_evidence` respectively. These align with the ones proposed in
+the [endorsements design doc](RemoteAttestatonCollaterals.md). There will also
```
anakrish: Typo  missing `i` in `RemoteAttestatonCollaterals.md`

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,775 @@
...
+be functions for registering and unregistering plugins called
+`oe_register_attestation_plugin` and `oe_unregister_attestation_plugin`. The user
+can statically link in their desired plugin and call the register plugin function.
```
anakrish: In the future, OE would be able to load modules dynamically. Would the requirement of `statically linking` the plugin work with dynamic loading?

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,775 @@
...
+  following claims (mapped from the `oe_identity_t`):
+  
+| Claim Name       | Claim Value Type   | Description                                                          |
+|:-----------------|:-------------------|:---------------------------------------------------------------------|
+| id_version       | uint32_t           | Claims version. Must be 0                                            |
+| security_version | uint32_t           | Security version of the enclave. (ISVN for SGX).                     |
+| attributes       | uint64_t           | Attributes flags for the evidence: <br/> `OE_REPORT_ATTRIBUTES_DEBUG`: The evidence is for a debug enclave.<br/> `OE_REPORT_ATTRIBUTES_REMOTE`: The evidence can be used for remote attestation.   |
+| unique_id        | uint8_t[32]        | The unique ID for the enclave (MRENCLAVE for SGX).                   |
+| signer_id        | uint8_t[32]        | The signer ID for the enclave (MRSIGNER for SGX).                    |
+| product_id       | uint8_t[32]        | The product ID for the enclave (ISVPRODID for SGX).                  |
```
jiria: For OPTEE, we also keep a note of trusted component public key [oe_identity_t.public_key](https://github.com/openenclave/openenclave/blob/feature.new_platforms/include/openenclave/bits/report.h). This change is not yet in `master` afaik, but wanted to start a discussion if there was something similar that would be interesting to map from SGX.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,775 @@
...
+| validity_from    | oe_datetime_t      | Overall datetime from which the evidence and endorsements are valid. |
+| validity_until   | oe_datetime_t      | Overall datetime at which the evidence and endorsements expire.      |
```
jiria: Another thing I would like to ask about. With SGX, there is only one TA to provide evidence for. With OPTEE, there are multiple levels of firmware, each with its own set of claims. We are currently expressing these as a [caller in oe_identity_t](https://github.com/openenclave/openenclave/blob/feature.new_platforms/include/openenclave/bits/report.h). Wondering if for example capturing claims about Quoting Enclave for SGX would be interesting.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,775 @@
...
+    const oe_datetime_t* input_validation_time,
+    oe_claim_t** claims,
+    size_t* claims_length)
+{
+    OE_UNUSED(plugin_context);
+
+    /*
+     * Pseudocode description instead of actual C code:
+     *
+     * Call oe_verify_report with all the input parameters and get the oe_identity_t back.
```
jiria: Should this be extended to support validating against Intel vs Azure baseline?

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,775 @@
...
+     * The function that gets executed when a plugin is registered.
+     *
+     * @param[in] plugin_context A pointer to the attestation plugin struct.
+     * @param[in] config_data An optional pointer to the configuration data.
+     * @param[in] config_data_size The size in bytes of config_data.
+     * @retval OE_OK on success.
+     */
+    oe_result_t (*on_register)(
+        oe_attestation_plugin_t* plugin_context,
+        const void* config_data,
```
jiria: Is `config_data` the field that would be used to pass in attestation policy? If so, how would I use different attestation policy per verification? Say, I have two different policies and which one is used is determined based on what operation is requested.

docs/DesignDocs/CustomAttestation.md
gupta-ak: We have `oe_get_public_key` that works for SGX, so I guess that could work for both. Is the public key distinct for each platform? In other words, if the enclave gets moved to a different machine, does the enclave's public key change? If so, then it's equivalent to SGX, since we derive the asymmetric keys from the sealing keys which are different for each platform.

docs/DesignDocs/CustomAttestation.md
gupta-ak: I could see how they align. The qeidentity field in the quote essentially all the `oe_identity_t` information of the quoting enclave. In both SGX and CyRes, it seems that the caller field can be the parent component of the TA in the trust chain.

docs/DesignDocs/CustomAttestation.md
gupta-ak: I'll change it to `linking`, I didn't mean to say that they could only be statically linked.

docs/DesignDocs/CustomAttestation.md
gupta-ak: thanks!

docs/DesignDocs/CustomAttestation.md
gupta-ak: I think `oe_verify_evidence` needs parameters for policy. What things are you thinking that would count as the policy? What format would they be in? Could the policies be validated in a plugin agnostic way?

docs/DesignDocs/CustomAttestation.md
gupta-ak: What do you mean Intel vs Azure baseline? The endorsements are part of the input parameters, so the caller can determine where to fetch it from

docs/DesignDocs/CustomAttestation.md
jiria: It would be awesome if we could align that. If you look at the `feature.new_platforms` branch, I have made some changes for OPTEE wrt the APIs as well (you need to free the stuff now). Please have a look if that is something that could be adopted.

docs/DesignDocs/CustomAttestation.md
jiria: I agree that `oe_verify_evidence` should have parameters for policy. @soccerGB earlier design was I believe passing policy through `on_register`, hence my question. Policy is something where caller can express additional assertions regarding the claims in the evidence. This should preferably be in the same format that AAS accepts. In my opinion policies should be plugin agnostic, there should be a common policy verification library shared with AAS. But I think @soccerGB had some other thoughts on this.

docs/DesignDocs/CustomAttestation.md
jiria: I was referring to `oe_verify_report`. But if I understand you correctly, `oe_verify_report` would not look at the endorsements, it would be job of the `sgx_attestation_plugin_verify_evidence`?

docs/DesignDocs/CustomAttestation.md
soccerGB: @jiria, thanks for the reminder.
I had an offline chat with gupta-ak and he will address your feedback.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,995 @@
...
+attester and another API for the verifier. Each plugin will define a UUID to
+distinguish plugins. An attester and verifier plugin sharing the same UUID
+indicates that that verifier is able to process the evidence format generated
+by the attester.
+
+Futhermore, there will be additional attestation "plugin aware" APIs that are
+analogous to `oe_get_report` and `oe_verify_report` called `oe_get_evidence`
+and `oe_verify_evidence` respectively. There will also
+be functions for registering and unregistering plugins called
+`oe_register_[attester|verifier]` and `oe_unregister_[attester|verfier]`. The user
```
jiria: verifier

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,995 @@
...
+    many protocols follow the "challenge response" architecture, which requires
+    the enclave to sign a nonce from the relying party.
+  - There is an `opt_params` field because some plugins might require plugin
+    specific input. For example, the SGX local attestation needs the
+    other enclave's target info struct.
+  - There is an `endorsements` parameter to return the endorsements that are
+    coupled with the evidence to ensure that the evidence and endorsements are
+    in sync.
+
+### Verfier Plugin API (Supported by host and enclave side)
```
jiria: Verifier

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,995 @@
...
+{
+    char* name;
+    uint8_t* value;
+    size_t value_size;
+};
+
+/**
+ * Supported policies for validation. Only time is supported for now.
+ */
+enum oe_policy_type_t
```
jiria: Why is this not a generic attestation policy that can target any combination of claims?

docs/DesignDocs/CustomAttestation.md
jiria: Thanks @soccerGB . @gupta-ak I had a quick look and it does not seem to be yet reflected. Can you please ping me when ready so I can have a look?

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,995 @@
...
+by the attester.
+
+Futhermore, there will be additional attestation "plugin aware" APIs that are
+analogous to `oe_get_report` and `oe_verify_report` called `oe_get_evidence`
+and `oe_verify_evidence` respectively. There will also
+be functions for registering and unregistering plugins called
+`oe_register_[attester|verifier]` and `oe_unregister_[attester|verfier]`. The user
+can link in their desired plugin and call the register plugin function.
+The attestation data can be retrieved from `oe_get_evidence` with the desired UUID.
+The generated data will have the UUID in its header. The user can `oe_verify_evidence`
```
dthaler: This line refers to the "UUID in its header".   Is this referring to some blob that needs to be passed over a network?  Or is it just internal to a machine?

If it needs to be passed over a network, then it presumably would be better to have a standard there rather than some OE-specific format, yes?   Would like to understand so I can help in bringing an ask to IETF or whatever other body if needed.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,995 @@
...
+        uint8_t* evidence_buffer);
+
+    /**
+     * Frees the generated attestation endorsements.
+     *
+     * @param[in] plugin_context A pointer to the attester plugin struct.
+     * @param[in] endorsements_buffer A pointer to the endorsements buffer.
+     * @retval OE_OK on success.
+     */
+    oe_result_t (*free_endorsements)(
```
dthaler: Since evidence and endorsements are both obtained from the same function (get_evidence), why do we need two separate free functions?   Is there a common case where you'd obtain them together but need to free them at different times?

docs/DesignDocs/CustomAttestation.md
gupta-ak: fixed.

docs/DesignDocs/CustomAttestation.md
gupta-ak: Fixed.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,995 @@
...
+[Remote Attestation Procedures (RATS)](https://datatracker.ietf.org/wg/rats/about/)
+working group.
+
+- Claims
+  - Claims are statements about a particular subject. They consist of
+    name-value pairs containing the claim name, which is a string, and
+    claim value, which is arbitrary data. Example of claims could be
+    [name="version", value=1] or [name="enclave_id", value=1111].
+- Evidence
+  - Evidence are claims about the enclave that are produced and signed by it.
```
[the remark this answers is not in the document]
gupta-ak: fixed.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -0,0 +1,995 @@
...
+    The SGX report would be an example of evidence.
+- Endorsements
+  - Endorsements are additional claims used in the evidence verification process,
+    but are not produced by the enclave. An example of an endorsement would be
```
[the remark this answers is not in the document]
gupta-ak: fixed.

docs/DesignDocs/CustomAttestation.md
gupta-ak: Yeah it will need to be passed over the network. The verifier gets the blob from the network and then calls `oe_verify_evidence`, which then parses the blob header to determine what verifier plugin to run using the UUID.

 It would be better if it's in a standard, but I didn't see something equivalent in the RATS standard. There's an UUID to uniquely identify the device, but I didn't see anything to distinguish different data formats in the paper.

docs/DesignDocs/CustomAttestation.md
gupta-ak: Good point. I removed `free_endorsements` and made `free_evidence` have the endorsements parameter to make it mirror `get_evidence`.

PR #2281, opened 2019-10-25
Merging latest master changes to v0.7.x

johnkord:
Note: This PR exists because it merges commits to the v0.7.x branch that have been rebased onto the v0.7.x branch. It is not targeting the master branch!


johnkord:
Thanks for the quick reviews!

PR #2284, opened 2019-10-30
Implement oe_enter, oe_enter_sim and __oe_host_stack_bridge in C

anakrish:
- Reduces the number of handwritten assembly code which is a source of
  bugs and CVEs.
- Most of the implementation is in C with minimal extended ASM.
- The C files can be used as templates for generating the assembly code if needed
- The assembly files can also be used to generate .asm for Windows via
  cross compilation since msvc doesn't support inline assembly in x64 mode.


anakrish:
bors try

anakrish:
Requires more work with GCC.

PR #2286, opened 2019-10-30
Implement oe_enter, oe_enter_sim and __oe_host_stack_bridge in C

anakrish:
- Reduces the number of handwritten assembly code which is a source of
  bugs and CVEs.
- Most of the implementation is in C with minimal extended ASM.
- The C files can be used as templates for generating the assembly code if needed
- The assembly files can also be used to generate .asm for Windows via
  cross compilation since msvc doesn't support inline assembly in x64 mode.


anakrish:
bors try

anakrish:
Messed up something

host/sgx/linux/enter.S
```diff
@@ -1,296 +0,0 @@
...
-#define ARG4            (-6*OE_WORDSIZE)(%rbp)
-#define ENCLAVE         (-7*OE_WORDSIZE)(%rbp)
-#define ARG1OUT         (-8*OE_WORDSIZE)(%rbp)
-#define ARG2OUT         (-9*OE_WORDSIZE)(%rbp)
-#define RSP             (-10*OE_WORDSIZE)(%rbp)
-#define HOST_CONTEXT    (-11*OE_WORDSIZE)(%rbp)
-// We need an extra OE_WORDSIZE to keep the stack aligned.
-#define PARAMS_SPACE    ((12*OE_WORDSIZE) + OE_CONTEXT_SIZE)
-
-.globl __morestack
```
radhikaj: Does removing this break the debugger?

PR #2288, opened 2019-10-31
Fixed report test case to update oe_datetime_t correctly.

### LINKED ISSUE #2287: Report unit test failure in tests/report/common/tests.cpp:1355  when valid_from.seconds happens to be zero
https://github.com/openenclave/openenclave/blob/a3ea8e5cea97a52ad0da9a8b3772ba0f1cd5a4de/tests/report/common/tests.cpp#L1355

valid_from.seconds -= 1;  <---when valid_from.seconds is 0, the subtraction will render 0xffffffff as result, which will fail VerifyReportWithCollaterals test.

Recommend: valid_from.seconds -> valid_from.year

jazzybluesea:
Previously, the report test case was incorrectly directly updating the seconds field of the oe_datetime_t. 
This issue is now resolved.  Fixes #2287.

### DIFF
--- a/tests/report/common/tests.cpp
+++ b/tests/report/common/tests.cpp
@@ -1352,7 +1352,7 @@ void test_verify_report_with_collaterals()
                 &valid_until,
                 NULL) == OE_OK);
 
-        valid_from.seconds -= 1;
+        valid_from.year -= 1;
         OE_TEST(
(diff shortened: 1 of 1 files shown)


jazzybluesea:
bors try

PR #2290, opened 2019-11-02
Fix the setx error:

nonpolarity:
ERROR: Invalid syntax. Default option is not allowed more than '2' time(s).
Type "SETX /?" for usage.

setx can only take 2 parameters. If there is space in the path, such as C:\Program Files\7-Zip, it means more than expected parameters.
The single quote make the whole string(with space(s)) as one parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,9 +40,6 @@ If you would like to run Ubuntu 16.04 or Ubuntu 18.04 in a Hyper-V VM on SGX
 capable hardware, see
 [Setting up a Linux Hyper-V VM on Windows with SGX Support](docs/GettingStartedDocs/HyperVLinuxVMSetup.md).
 
-If you would like to modify and build the Open Enclave SDK from sources, refer
-to the documents for [getting started](docs/GettingStartedDocs/Contributors/building_oe_sdk.md).
-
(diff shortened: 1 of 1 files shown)


scripts/install-windows-prereqs.ps1
```diff
@@ -240,7 +240,7 @@ function Add-ToSystemPath {
         }
     }
     $env:PATH = $currentPath -join ';'
-    setx.exe /M PATH ($systemPath -join ';')
+    setx.exe /M \'PATH\' \'($systemPath -join ';')\'
```
andschwa: To the question you asked me earlier: the `$env:PATH` line sets `PATH` to the scope of the currently executing script, while `setx.exe /M` updates the system-wide `PATH` variable. So I think we need both lines.

However, and especially to get around the max character limit of `setx`, I would instead replace this with in-line C#:

```powershell
[System.Environment]::SetEnvironmentVariable('PATH', $env:PATH, [System.EnvironmentVariableTarget]::Machine)
```

andschwa: FYI @BRMcLaren @johnkord

PR #2291, opened 2019-11-02
Simplify and clean up assembly functions

anakrish:
- oe_enter simplified and handle simulation mode too
- oe_enter_sim removed both on Linux and Windows
- __oe_host_stack_bridge (Linux) implemented in C
- Windows and Linux versions made as close as possible


anakrish:
bors try

host/sgx/windows/enter.asm
```diff
@@ -109,71 +84,66 @@ execute_eenter:
...
+    mov rdx, rsi
+    lea r8, ARG1
+    lea r9, ARG2
     mov rax, qword ptr TCS
     mov qword ptr [rsp+32], rax
-    mov rax, qword ptr ENCLAVE
-    mov qword ptr [rsp+40], rax
+    mov rax, ENCLAVE
+    mov [rsp+40], rax
     call __oe_dispatch_ocall ;; RAX contains return value
```
CodeMonkeyLeet: The comment about RAX is probably unnecessary at this point, and we should consider making `__oe_dispatch_ocall` a `void` return method since nothing actually makes use of the return value.

I would actually prefer the reverse of this change: update the linux version to invoke dispatch and let it return whether it was necessary or not, as it removes some brittle elements to maintain in assembly (e.g. OE_OCALL_CODE definitions being in sync).

host/sgx/windows/enter.asm
```diff
@@ -109,71 +84,66 @@ execute_eenter:
...
-    mov rsp, rbp
+    ;; Write results
+    ;; RSI contains arg1, RDI contains arg2
+    mov rbx, ARG3
+    mov [rbx], rdi
+    mov rcx, ARG4
+    mov [rcx], rsi
+
+    ;; Restore callee saved registers
+    fxrstor FX_SPACE
```
CodeMonkeyLeet: Again, this seems a little weird because you've separated out `fxrstor` from restoring registers in the context `execute_eenter` but we've effectively double-restored the FX state, once for on return from EENTER for the FX state, and here again for the XMM registers, even though each invocation restores everything. It deserves a comment at least, and from an encapsulation perspective, I'd say it's cleanest to always save and restore the full context as before rather than try to optimize partial contexts for save & restore. 

Alternatively, consider that restoring callee-saved registers could be done once on return from the EENTER instead of at the end of the function: we have to do it insofar as host doesn't trust enclave to preserve them, but within the context of this function, you should be able to definitively say whether any of these were modified (e.g. a call to dispatch should preserve callee-saved registers on your behalf) and we generally shouldn't have to restore them again.

host/sgx/windows/enter.asm
```diff
@@ -109,71 +84,66 @@ execute_eenter:
...
-    ;; Prepare to reenter the enclave, calling the entry point.
-    mov rax, ARG1OUT
-    mov ARG1, rax
-    mov rax, ARG2OUT
-    mov ARG2, rax
     jmp execute_eenter
 
 return_from_ecall:
-    ;; Stop speculative execution at target of conditional jump
-    lfence
```
jxyang: why we don't need lfence anymore?

host/sgx/windows/enter.asm
```diff
@@ -109,71 +84,66 @@ execute_eenter:
...
-    call oe_restore_host_context
+    ;; Restore flags and floating point state.
+    popfq
+    fxrstor FX_SPACE
+
+    ;; Check if an OCALL needs to be dispatched.
+    mov r10, rdi
+    shr r10, 48
+    cmp r10, OE_OCALL_CODE
+    jne return_from_ecall
```
jxyang: Need a lfence after the conditional jump?

host/sgx/linux/enter.S
```diff
@@ -84,213 +78,74 @@ __morestack:
...
-    mov ARG1OUT, %rax
-    mov %rax, ARG1
-    mov ARG2OUT, %rax
-    mov %rax, ARG2
     jmp .execute_eenter
 
 .return_from_ecall:
-    // Stop speculative execution at target of conditional jump
-    // ecall-return-check and ecall-return-check-1.
-    lfence
```
jxyang: Likewise, why remove lfence here?

host/sgx/linux/enter.S
```diff
@@ -84,213 +78,74 @@ __morestack:
...
+    // Check if an OCALL needs to be dispatched.
+    mov %rdi, %r10
     shr $48, %r10
     cmpq $OE_OCALL_CODE, %r10
     jne .return_from_ecall
 
 .dispatch_ocall:
-    // Stop speculative execution at fallthrough of conditional
-    // ecall-return-check.
-    lfence
```
jxyang: One more lfence removeval.

host/sgx/linux/enter.S
```diff
@@ -84,213 +78,74 @@ __morestack:
...
+    //      RBP and RSP are not changed,
     //      All other Registers are clobbered.
     mov TCS, %rbx
     mov AEP, %rcx
     mov ARG1, %rdi
     mov ARG2, %rsi
+
+    mov ENCLAVE, %r10
+    cmpb $0, OE_ENCLAVE_OFFSET_SIMULATE(%r10)
+    je .execute_enclu
```
jxyang: Need lfence after the conditional jump?

host/sgx/linux/enter.S
```diff
@@ -61,20 +57,18 @@ __morestack:
...
-
-    // Save the current context.
-    mov HOST_CONTEXT, %rdi
-    call oe_snap_current_context@PLT
-
-    // Save the stack pointer so enclave can use the stack.
-    mov %rsp, RSP
+    // Save flags and floating-point state.
+    fxsave FX_SPACE
+    pushfq
```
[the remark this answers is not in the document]
anakrish: [x64 prolog and epilog](https://docs.microsoft.com/en-us/cpp/build/prolog-and-epilog?view=vs-2019) must be written in a certain way to be x64 Windows ABI compliant.

I prolog, the non-volatile registers (RBX, RBP, RDI, RSI, RSP, R12, R13, R14, R15, and XMM6-15) must be pushed before allocating stack-space and establishing the frame pointer.

The epilog must first deallocate the fixed part of the stack, the nonvolatile registers are popped, and control is returned to the calling function.

Both the earlier and this PR's implementation are not x64 Win ABI compliant.
I will take another pass to make windows implementation compliant.
However, it means that register preservation must be separated from flags/FX state preservation.

anakrish:
Thanks everyone for the feedback.
Incorporate into a new PR https://github.com/openenclave/openenclave/pull/2301

PR #2293, opened 2019-11-06
Add support for V2 of SGX Collaterals/Endorsements.

jazzybluesea:
Parsing changes (#2308)
Validation changes (#TBD)

Signed-off-by: Sergio Wong sewong@microsoft.com

### DIFF
--- a/docs/DesignDocs/SGX_Endorsements_V2.md
+++ b/docs/DesignDocs/SGX_Endorsements_V2.md
@@ -1,0 +1,222 @@
+SGX Attestation Endorsements V2 Updates
+================
+
+Intel SGX released v2 of the Provision Certification Service for ECDSA Attestation.  The
+following APIs were affected:
+
+1. Get TCB Info.
+2. Get Quoting Enclave Identity.
+
+[API Version 1](https://api.portal.trustedservices.intel.com/documentation#pcs-certificate)
+
+[API Version 2](https://api.portal.trustedservices.intel.com/documentation#pcs-certificate-v2)
+
+This document describes the changes for supporting version 2 of the API as well as keeping
+the existing support for version 1 of the API.  This document assumes the user is familiar with
+the SGX attestation endorsements and how it is used during SGX quote verification.
+
+Motivation
+----------
+
+This change is required for customers who would like to use the v2 API web endpoints. Note that the Azure PCK
+Caching Service and the Azure DCAP Client will also need to be updated and it is outside of the this specification.
+
+User Experience
+---------------
+
+The user experience does not change, it should be seamless to the user.  The caller will be able to identify the version of the data by checking the existing `version` field.
+
+The quote verifier will specify whether to use the v1 or v2 Azure PCK Caching Service web API endpoints.  The
+quote verification logic will be able to determine the version of the API by checking the `version` field in endorsement data.
+
+Specification
+-------------
+
+## Differences between V1 and V2 APIs.
+
+The APIs to get the PCK Certificate and the revocation list are the same between the versions.  The APIs to
+get the TCB Info and get QE Identity are different.  The differences are highlighted below.
+
+### Get TCB Info
+1. tcbInfo:version changed from 1 to 2.
+2. New field tcbInfo:tcbType.
+3. New field tcbInfo:tcbEvaluationDataNumber.
+4. Field tcbInfo:tcbLevels:status was renamed to tcbInfo:tcbLevels:tcbStatus.
+5. New fields in tcbInfo:tcbLevels named tcbDate and advisoryIDs.
+
+### Get QE Identity Info
+1. The 'qeIdentity' block was renamed to 'enclaveIdentity'.
+2. A new field, enclaveIdentity:id was added to differentiate QE Id info from Quoting Enclave (QE) and Quoting Validation Enclave (QVE).
+3. enclaveIdentity:version changed from 1 to 2.
+4. New field enclaveIdentity:tcbEvaluationDataNumber.
+5. qeIdentity:isvsvn moved to enclaveIdentity:tcbLevels:tcb:isvsvn.
+6. New fields in enclaveIdentity:tcbLevels named tcbData, tcbStatus and advisoryIDs.
+
+### New field definitions
+
+#### tcbType:
+	type: integer
+	example: 0
+	description: >- Type of TCB level composition that determines TCB level comparison logic
+
+##### tcbEvaluationDataNumber:
+	type: integer
+	example: 2
+	description: >- A monotonically increasing sequence number changed when Intel updates the content of the TCB evaluation data set: TCB Info, QE Idenity and QVE Identity. The tcbEvaluationDataNumber update is synchronized across TCB Info for all flavors of SGX CPUs (Family-Model-Stepping-Platform-CustomSKU) and QE/QVE Identity. This sequence number allows users to easily determine when a particular TCB Info/QE Idenity/QVE Identiy superseedes another TCB Info/QE Identity/QVE Identity (value: current TCB Recovery event number stored in the database).
+
+##### tcbDate:
+	type: string
+	format: date-time
+	description: >- Representation of date and time when the TCB level was certified not to be vulnerable to any issues described in SAs that were published on or prior to this date. The time shall be in UTC and the encoding shall be compliant to ISO 8601 standard (YYYY-MM-DDThh:mm:ssZ).
+
+##### tcbStatus:
+	type: string
+	description: TCB level status
+
+##### advisoryIDs:
+	type: array
+	description: >- Array of Advisory IDs describing vulnerabilities that this TCB level is vulnerable to. Note: The value can be different for different FMSPCs. This field is optional. It will be present only if the list of Advisory IDs is not empty.
+	items:
+        type: string
+
+For more information on the new fields please see [API Version 2](https://api.portal.trustedservices.intel.com/documentation#pcs-certificate-v2).
+
+
+## Changes
+
+### Update current structs with new additional fields and create new type definitions.
+
+```C
+typedef enum _oe_tcb_level_status
+{
+    // Existing fields
+    OE_TCB_LEVEL_STATUS_UNKNOWN,
+    OE_TCB_LEVEL_STATUS_REVOKED,
+    OE_TCB_LEVEL_STATUS_OUT_OF_DATE,
+    OE_TCB_LEVEL_STATUS_CONFIGURATION_NEEDED,
+    OE_TCB_LEVEL_STATUS_UP_TO_DATE,
+
+    // New field for new value "OutOfDateConfigurationNeeded"
+    OE_TCB_LEVEL_STATUS_OUTOFDATE_CONFIGURATION_NEEDED,
+
+    __OE_TCB_LEVEL_MAX = OE_ENUM_MAX,
+
+} oe_tcb_level_status_t;
+
+// Existing TCB Level struct for TCB Info.  Will rename this
+// struct to oe_tcb_tcb_level given that
+// QE identity info also has its TCB level data field.
+typedef struct _oe_tcb_level
+{
+    // Existing fields
+    uint8_t sgx_tcb_comp_svn[16];
+    uint16_t pce_svn;
+    oe_tcb_level_status_t tcb_status;
+
+    // New fields
+    oe_datetime_t tcb_date;
+    uint8_t* advisory_ids_json;     // ["INTEL-SA-00079", "INTEL-SA-00076"]
+
+} oe_tcb_level_t;
+
+typedef struct _oe_parsed_tcb_info
+{
+    // Existing fields
+    uint32_t version;
+    oe_datetime_t issue_date;
+    oe_datetime_t next_update;
+    uint8_t fmspc[6];
+    uint8_t pceid[2];
+    uint8_t signature[64];
+    const uint8_t* tcb_info_start;
+    size_t tcb_info_size;
+
+    // New fields
+    uint32_t tcb_type;
+    uint32_t tcb_evaluation_data_number;
+
+} oe_parsed_tcb_info_t;
+
+/*!
+ * New enum for QE id field
+ */
+typedef enum _oe_qe_identity_id
+{
+    OE_IDENTITY_ID_QE,
+    QE_IDENTITY_ID_QVE
+} oe_qe_identity_id;
+
+/*!
+ * New TCB level for QE identity info.
+ */
+typedef struct _oe_qe_tcb_level
+{
+    uint32_t isvsvn;
+    oe_tcb_level_status_t tcb_status;
+    oe_datetime_t tcb_date;
+    uint8_t* advisory_ids_json;     // ["INTEL-SA-00079", "INTEL-SA-00076"]
+
+} oe_qe_tcb_level_t;
+
+typedef struct _oe_parsed_qe_identity_info
+{
+    // Existing fields
+    uint32_t version;
+    oe_datetime_t issue_date;
+    oe_datetime_t next_update;
+    uint32_t miscselect;         // The MISCSELECT that must be set
+    uint32_t miscselect_mask;    // Mask of MISCSELECT to enforce
+    sgx_attributes_t attributes; // flags and xfrm (XSAVE-Feature Request Mask)
+    uint64_t attributes_flags_mask;   // mask for attributes.flags
+    uint64_t attributes_xfrm_mask;    // mask for attributes.xfrm
+    uint8_t mrsigner[OE_SHA256_SIZE]; // MRSIGNER of the enclave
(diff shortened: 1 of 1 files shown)


docs/DesignDocs/SGX_Endorsements_V2.md
```diff
@@ -0,0 +1,222 @@
...
+This change is required for customers who would like to use the v2 API web endpoints. Note that the Azure PCK
+Caching Service and the Azure DCAP Client will also need to be updated and it is outside of the this specification.
+
+User Experience
+---------------
+
+The user experience does not change, it should be seamless to the user.  The caller will be able to identify the version of the data by checking the existing `version` field.
+
+The quote verifier will specify whether to use the v1 or v2 Azure PCK Caching Service web API endpoints.  The
+quote verification logic will be able to determine the version of the API by checking the `version` field in endorsement data.
```
soccerGB: How does a quote verifier provide this option?

docs/DesignDocs/SGX_Endorsements_V2.md
```diff
@@ -0,0 +1,222 @@
...
+
+Specification
+-------------
+
+## Differences between V1 and V2 APIs.
+
+The APIs to get the PCK Certificate and the revocation list are the same between the versions.  The APIs to
+get the TCB Info and get QE Identity are different.  The differences are highlighted below.
+
```
soccerGB: A brief description on what those new fields are for would be great,

docs/DesignDocs/SGX_Endorsements_V2.md
```diff
@@ -0,0 +1,222 @@
...
+5. New fields in tcbInfo:tcbLevels named tcbDate and advisoryIDs.
+
+### Get QE Identity Info
+1. The 'qeIdentity' block was renamed to 'enclaveIdentity'.
+2. A new field, enclaveIdentity:id was added to differentiate QE Id info from Quoting Enclave (QE) and Quoting Validation Enclave (QVE).
+3. enclaveIdentity:version changed from 1 to 2.
+4. New field enclaveIdentity:tcbEvaluationDataNumber.
+5. qeIdentity:isvsvn moved to enclaveIdentity:tcbLevels:tcb:isvsvn.
+6. New fields in enclaveIdentity:tcbLevels named tcbData, tcbStatus and advisoryIDs.
+
```
soccerGB: Do you want to mention QvE identity here as well?

docs/DesignDocs/SGX_Endorsements_V2.md
```diff
@@ -0,0 +1,222 @@
...
+    // Existing fields
+    OE_TCB_LEVEL_STATUS_UNKNOWN,
+    OE_TCB_LEVEL_STATUS_REVOKED,
+    OE_TCB_LEVEL_STATUS_OUT_OF_DATE,
+    OE_TCB_LEVEL_STATUS_CONFIGURATION_NEEDED,
+    OE_TCB_LEVEL_STATUS_UP_TO_DATE,
+
+    // New field for new value "OutOfDateConfigurationNeeded"
+    OE_TCB_LEVEL_STATUS_OUTOFDATE_CONFIGURATION_NEEDED,
+
```
soccerGB: This value is kind of confusing. 
I am wondering: would the status be considered as up-to-date if configured properly?

docs/DesignDocs/SGX_Endorsements_V2.md
```diff
@@ -0,0 +1,222 @@
...
+    __OE_TCB_LEVEL_MAX = OE_ENUM_MAX,
+
+} oe_tcb_level_status_t;
+
+// Existing TCB Level struct for TCB Info.  Will rename this
+// struct to oe_tcb_tcb_level given that
+// QE identity info also has its TCB level data field.
```
soccerGB: oe_tcb_tcb_level seems to be redudant...

docs/DesignDocs/SGX_Endorsements_V2.md
```diff
@@ -0,0 +1,222 @@
...
+/*!
+ * New TCB level for QE identity info.
+ */
+typedef struct _oe_qe_tcb_level
+{
+    uint32_t isvsvn;
+    oe_tcb_level_status_t tcb_status;
+    oe_datetime_t tcb_date;
+    uint8_t* advisory_ids_json;     // ["INTEL-SA-00079", "INTEL-SA-00076"]
+
```
soccerGB: what are ["INTEL-SA-00079", "INTEL-SA-00076"]?

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: This depends on how the version of the DCAP provider installed.

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: Sure.

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: I think the status "OutOfDateConfigurationNeeded" is a subset of "OutOfDate", so it will not be considered "UpToDate".

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: The additional "tcb_" prefix helps it differentiate between a TCB level in the TCB Info and a TCB level in the QE Identity Info, which will have the "qe_" prefix.

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: These are the values from the new advisoryIDs field.  This field is an array of strings that describe the vulnerabilities that the corresponding TCB level is vulnerable to.  The value can be different for each FMSPCs, and it is an optional field.

soccerGB: I see. Could add those information into the doc?

docs/DesignDocs/SGX_Endorsements_V2.md
```diff
@@ -0,0 +1,222 @@
...
+This document describes the changes for supporting version 2 of the API as well as keeping
+the existing support for version 1 of the API.  This document assumes the user is familiar with
+the SGX attestation endorsements and how it is used during SGX quote verification.
+
+Motivation
+----------
+
+This change is required for customers who would like to use the v2 API web endpoints. Note that the Azure PCK
+Caching Service and the Azure DCAP Client will also need to be updated and it is outside of the this specification.
```
dthaler: nit: "the this" -> "this"

docs/DesignDocs/SGX_Endorsements_V2.md
```diff
@@ -0,0 +1,222 @@
...
+
+User Experience
+---------------
+
+The user experience does not change, it should be seamless to the user.  The caller will be able to identify the version of the data by checking the existing `version` field.
+
+The quote verifier will specify whether to use the v1 or v2 Azure PCK Caching Service web API endpoints.  The
```
dthaler: Clarify: the text here mentions Azure... is this design doc only usable with Azure?  If so, I'd recommend changing the title.  If not, I'd recommend changing the text here to clarify.

docs/DesignDocs/SGX_Endorsements_V2.md
```diff
@@ -0,0 +1,222 @@
...
+
+##### tcbEvaluationDataNumber:
+	type: integer
+	example: 2
+	description: >- A monotonically increasing sequence number changed when Intel updates the content of the TCB evaluation data set: TCB Info, QE Idenity and QVE Identity. The tcbEvaluationDataNumber update is synchronized across TCB Info for all flavors of SGX CPUs (Family-Model-Stepping-Platform-CustomSKU) and QE/QVE Identity. This sequence number allows users to easily determine when a particular TCB Info/QE Idenity/QVE Identiy superseedes another TCB Info/QE Identity/QVE Identity (value: current TCB Recovery event number stored in the database).
+
+##### tcbDate:
+	type: string
+	format: date-time
+	description: >- Representation of date and time when the TCB level was certified not to be vulnerable to any issues described in SAs that were published on or prior to this date. The time shall be in UTC and the encoding shall be compliant to ISO 8601 standard (YYYY-MM-DDThh:mm:ssZ).
```
dthaler: nit: expand "SAs"  ("Security Advisories"?)

docs/DesignDocs/SGX_Endorsements_V2.md
dthaler: Agree that this value is confusing for the above reasons.  In my opinion, the names should be clear, and statuses should be disjoint, so that one status is not a subset of another status, if this is an enum.  My guess though is that you mean that OE_TCB_LEVEL_STATUS_CONFIGURATION_NEEDED and OE_TCB_LEVEL_STATUS_OUT_OF_DATE are both set, and so you need a separate value.   If so, then it seems to me that the use of an enum is itself problematic and replacing it with a bitfield would be better.

docs/DesignDocs/SGX_Endorsements_V2.md
```diff
@@ -0,0 +1,222 @@
...
+
+} oe_tcb_level_status_t;
+
+// Existing TCB Level struct for TCB Info.  Will rename this
+// struct to oe_tcb_tcb_level given that
+// QE identity info also has its TCB level data field.
+typedef struct _oe_tcb_level
+{
+    // Existing fields
+    uint8_t sgx_tcb_comp_svn[16];
```
dthaler: This looks like an sgx-specific field in a non-sgx-specific struct.  Either make the field name be generic, or make the struct name have sgx in the name.

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: There is already a section "New field definitions" in the doc that has a description of this new field.

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: Had a side discussion on this, will clarify the quote verifier as an entity that does the quote verification.

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: Fixed.

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: No it's not only usable with Azure.

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: This existing struct is defined in the common/sgx/tcbinfo.h file which implies sgx.  I think it makes sense to rename it sgx specific, but if we do this it should be part of a different issue given that all the types under common/sgx are sgx specific.

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: I found some logic in the Intel github that sheds some more information on this new status and looks like a combination of both the QE Identity Info status and the TCB Info status.  For now I will keep it as a separate status given that this is its own special case.

Status convergeTcbstatus(Status tcbInfoStatus, Status qeIdentityStatus) {
    if (qeIdentityStatus == STATUS_OK) {
        return tcbInfoStatus;
    }
    if (qeIdentityStatus == STATUS_SGX_ENCLAVE_IDENTITY_OUT_OF_DATE ||
        qeIdentityStatus == STATUS_SGX_ENCLAVE_REPORT_ISVSVN_OUT_OF_DATE ||
        qeIdentityStatus == STATUS_QE_IDENTITY_OUT_OF_DATE) {
        switch (tcbInfoStatus)
        {
        case STATUS_OK:
            return STATUS_TCB_OUT_OF_DATE;
        case STATUS_TCB_CONFIGURATION_NEEDED:
            return STATUS_TCB_OUT_OF_DATE_CONFIGURATION_NEEDED;
        default:
            return tcbInfoStatus;
        }
    }
    return STATUS_INVALID_PARAMETER;
}

dthaler: To me that code confirms that an enum is not appropriate since it's really an orthogonal property.  Do we really need to match a bad API design here?

jazzybluesea: I agree that using the bitfield is better, but I am not 100% sure what is the original intent of this new status.

docs/DesignDocs/SGX_Endorsements_V2.md
dthaler: The title still says "SGX Attestation Endorsements V2 Updates", nothing with Azure in the title.  Your response indicated that this doc was only relevant to people who use Azure, not on-prem SGX, so the title still needs updating unless I'm missing something.

dthaler:
I'm willing to listen to arguments that the API surface area for SGX needs to match what intel SDK has, even if the API is confusing/problematic.  Personally, I'd aspire to a more usable API.  Especially now that the Intel SGX SDK for Linux has been contributed to the CCC, we should be able to change that code too.

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: Ok, let's change it to a bitfield.  It is definitely less confusing.

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: Updated title. Thank you Dave.

jazzybluesea:
Cheng-mean/Dave, any other concerns you would like to address?  Thank you.

docs/DesignDocs/SGX_Endorsements_V2.md
```diff
@@ -0,0 +1,254 @@
...
+[API Version 2](https://api.portal.trustedservices.intel.com/documentation#pcs-certificate-v2)
+
+This document describes the changes for supporting version 2 of the API as well as keeping
+the existing support for version 1 of the API.  This document assumes the user is familiar with
+the SGX attestation endorsements and how it is used during SGX quote verification.
+
+Motivation
+----------
+
+This change is required for customers who would like to use the v2 API web endpoints. Note that the Azure PCK
```
dthaler: This paragraph is azure specific but line 1 is not Azure specific.  If the entire document is Azure-only and not on-prem then please update line 1.  However, I suspect part of this doc are not azure-only, and so maybe it's better to prefix such sentences as this with conditions like "If SGX is being used in an Azure data center context, ..."

docs/DesignDocs/SGX_Endorsements_V2.md
```diff
@@ -0,0 +1,254 @@
...
+#define OE_TCB_LEVEL_STATUS_QE_IDENTITY_OUT_OF_DATE (1 << 4)
+
+/*! \struct oe_tcb_level_status_t
+ */
+typedef union _oe_tcb_level_status
+{
+    struct {
+        uint32_t revoked : 1;                       //! "Revoked"
+        uint32_t outofdate : 1;                     //! "OutOfDate"
+        uint32_t configuration_needed :1;           //! "ConfigurationNeeded"
```
dthaler: nit: would be nice to use consistent spacing, i.e. space after : like in the rest of the lines

jazzybluesea: Good catch. Fixed typo.

docs/DesignDocs/SGX_Endorsements_V2.md
jazzybluesea: Resolved.  The changes should apply to either non-Azure DCAP clients as well.

dthaler: Agree, change looks good, thanks.

docs/DesignDocs/SGX_Endorsements_V2.md
```diff
@@ -0,0 +1,255 @@
...
+
+For more information on the new fields please see [API Version 2](https://api.portal.trustedservices.intel.com/documentation#pcs-certificate-v2).
+
+
+## Changes
+
+### Update current structs with new additional fields and create new type definitions.
+
+```C
+#define OE_TCB_LEVEL_STATUS_UNKNOWN                 (0)
```
dthaler: now that oe_tcp_level_status_t is a bitfield, can lines 91-96 be deleted or are they still used somewhere?

jazzybluesea: I actually already have a PR out with the code changes! Currently I am still using OE_TCB_LEVEL_STATUS_UNKNOWN.  The other ones are there just for reference as of now...

PR #2295, opened 2019-11-07
Update project to use DCO instead of CLA

CodeMonkeyLeet:
- Replace references to the CLA requirement from contribution documents
  with information about the new DCO requirements.

- Add a git hook to enforce that all commit messages must include a sign-off
  for the DCO.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,7 @@ if (IS_DIRECTORY "${PROJECT_SOURCE_DIR}/.git")
   # Install Git pre-commit hook
   if (NOT WIN32)
     file(
-      COPY scripts/pre-commit
+      COPY scripts/pre-commit scripts/commit-msg
       DESTINATION "${PROJECT_SOURCE_DIR}/.git/hooks")
   endif ()
 endif ()
--- a/scripts/commit-msg
+++ b/scripts/commit-msg
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+set -o errexit
+
+exit_() {
+    echo ""
+    echo "$1"
+    echo ""
+    echo "This hook can be skipped if needed with 'git commit --no-verify'"
+    echo "See '.git/hooks/commit-msg', installed from 'scripts/commit-msg'"
+    exit 1
+}
+
+sign_offs="$(grep '^Signed-off-by: ' "$1" || test $? = 1 )"
+
+if [[ -z $sign_offs ]]; then
+    exit_ "Commit failed: please sign-off on the DCO with 'git commit -s'"
+fi
+
+if [[ -n $(echo "$sign_offs" | sort | uniq -c | sed -e '/^[ 	]*1[ 	]/d') ]]; then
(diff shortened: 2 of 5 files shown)


CMakeLists.txt
```diff
@@ -53,6 +53,9 @@ if (IS_DIRECTORY "${PROJECT_SOURCE_DIR}/.git")
     file(
       COPY scripts/pre-commit
       DESTINATION "${PROJECT_SOURCE_DIR}/.git/hooks")
+    file(
+      COPY scripts/commit-msg
```
andschwa: `COPY` takes multiple files so this can just be `COPY scripts/pre-commit scripts/commit-msg`

scripts/commit-msg
```diff
@@ -0,0 +1,25 @@
...
+    exit 1
+}
+
+sign_offs="$(grep '^Signed-off-by: ' "$1" || test $? = 1 )"
+
+if [ -z "$sign_offs" ]; then
+    exit_ "Commit failed: please sign-off on the DCO with 'git commit -s'"
+fi
+
+if [ -n "$(echo "$sign_offs" | sort | uniq -c | sed -e '/^[ 	]*1[ 	]/d')" ]; then
```
andschwa: What is this black magic?

scripts/commit-msg
```diff
@@ -0,0 +1,25 @@
...
+    echo "$1"
+    echo ""
+    echo "This hook can be skipped if needed with 'git commit --no-verify'"
+    echo "See '.git/hooks/commit-msg', installed from 'scripts/commit-msg'"
+    exit 1
+}
+
+sign_offs="$(grep '^Signed-off-by: ' "$1" || test $? = 1 )"
+
+if [ -z "$sign_offs" ]; then
```
andschwa: Does this pass shellcheck? We usually prefer the Bashism [`[[`](http://mywiki.wooledge.org/BashFAQ/031).

dthaler:
Looks great to me.  Thanks!

scripts/commit-msg
CodeMonkeyLeet: Yes, I run shellcheck and it passes just fine even with the `-s bash` option. Is there a way to enforce stylistic preferences with it?

What does the "bashism" entail in this context? Using `[[` & `]]` and not quoting variables? Anything else?

CMakeLists.txt
CodeMonkeyLeet: Simplified this

scripts/commit-msg
CodeMonkeyLeet: It's part of the git commit-msg.sample that I carried forward. It takes all the found 'Signed-off-by:' lines found by the grep, sorts them, enumerates all unique values (`uniq -c`) and removes all lines with only a count of 1. The `sed` command just makes finding the lines prefixed by the count of `1` robust against differences in whitespaces.

scripts/commit-msg
andschwa: I guess not. But yeah `[` is an actual command (literally `which [` yields `/usr/bin/[`) but `[[` is just Bash syntax, so it's fewer process calls. Not that I really care about performance here.

jazzybluesea:
I just tried this.  This works great!

PR #2296, opened 2019-11-07
Add ocamlformat back to oeedger8r

andschwa:
PR https://github.com/ocaml-ppx/ocamlformat/pull/980 fixed the ocamlformat package to install on Windows, and the fix is in the latest release 0.12. This means that we can add it back, and make running ocamlformat part of the build, but it also requires auto-formatting the edger8r code again.


andschwa:
bors try

andschwa:
@anakrish Sorry, actually can't do it as part of the CMake build because we copy files into the build tree, and it doesn't make sense to format copied files. But mostly I've been using `esy` directly while developing, and then CMake at the end.

andschwa:
bors try

andschwa:
bors try

tools/oeedger8r/README.md
```diff
@@ -71,10 +71,17 @@ improved plugin model as it is a copy of Intel's code.
 
 ### Best Practices
 
-We use [ocamlformat](https://github.com/ocaml-ppx/ocamlformat) to format our
-code (such as `Emitter.ml`, but not Intel's code). It is the final say in
-formatting. This should be setup to run automatically in one's editor, as it has
-not yet been setup in CI.
+We use [ocamlformat v0.12](https://github.com/ocaml-ppx/ocamlformat) (bundled
```
anakrish: - Is ocamlformat run as part of commit-checks?

tools/oeedger8r/CMakeLists.txt
```diff
@@ -26,7 +26,8 @@ add_custom_command(
   COMMAND cmake -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR}/intel ${CMAKE_CURRENT_BINARY_DIR}/intel
   COMMAND cmake -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR}/src ${CMAKE_CURRENT_BINARY_DIR}/src
   COMMAND cmake -E copy_if_different ${CMAKE_CURRENT_SOURCE_DIR}/dune-project ${CMAKE_CURRENT_SOURCE_DIR}/package.json ${CMAKE_CURRENT_BINARY_DIR}
-  COMMAND esy
+  COMMAND esy install
```
anakrish: Why do we need to `esy install`?

tools/oeedger8r/README.md
andschwa: Not yet. I was looking to do that, but I'd want to filter it such that it's only run if the OCaml files are touched, which means more extensive changes to the hook than I expected.

tools/oeedger8r/CMakeLists.txt
andschwa: `esy` is equivalent to `esy install && esy build`. Running `esy install` is a necessary step to build; it resolves the `package.json` dependencies (such as OCaml etc.) and makes them available in the build environment. This is more again to `pyenv`, not `apt`: it's not a system-wide "install" it's an "install my dependencies for this environment" step (modeled after `npm` really).

andschwa:
@CodeMonkeyLeet There's really no "packages" being pulled in. The `esy.lock` folder is like any other lockfile, just a lot more verbose 😅 it's all just metadata recording the exact state of the package dependencies as of today (which is something we desire for reproducible builds). The actual packages (their sources and resulting artifacts) are pulled in at build-time to `~/.esy`, not any where in the repo. There is no OPAM maintenance to speak of, as we are not using the OCaml tools or OPAM tools directly, it's all bundled by `esy`.

PR #2299, opened 2019-11-08
Update 3rdparty/mbedtls/config.h with 2.16 options

CodeMonkeyLeet:
Merge new options from 3rdparty/mbedtls/mbedtls/include/mbedtls/config.h into
config.h used by OE SDK. These options were introduced with the upgrade to
v2.16.2.

This change also cleans up some Open Enclave configuration choices:

- Disable options previously deprecated in v0.7:
    - `MBEDTLS_TLS_DEFAULT_ALLOW_SHA1_IN_KEY_EXCHANGE`: Considerable advances
      have been made in breaking SHA1 since our original review and we would
      like to be more prescriptive in recommending the use of SHA256.
    - `MBEDTLS_KEY_EXCHANGE_RSA_ENABLED`: This option provides no perfect
      forward secrecy and is generally becoming less popular as this is
      recognized. The ECDHE variants are also more performant.

- Disable `MBEDTLS_SELF_TEST` by default.
    - This is a performance and TCB optimization as mbedtls is not currently
      compliance certified anyway.
    - The removal of this breaks the aarch64 build for OP-TEE, which had an
      incorrect link ordering for liboeutee and liboecore that was masked by
      libmbedcrypto pulling in the TEE_GenerateRandom symbol for self-tests
      first.
      - Update CMakeLists.txt for liboeutee to mark TEE_GenerateRandom with
        the `--undefined` linker flag to force this to be pulled in.

### DIFF
--- a/3rdparty/mbedtls/config.h
+++ b/3rdparty/mbedtls/config.h
@@ -8,7 +8,7 @@
  *  memory footprint.
  */
 /*
- *  Copyright (C) 2006-2015, ARM Limited, All Rights Reserved
+ *  Copyright (C) 2006-2018, ARM Limited, All Rights Reserved
  *  SPDX-License-Identifier: Apache-2.0
  *
  *  Licensed under the Apache License, Version 2.0 (the "License"); you may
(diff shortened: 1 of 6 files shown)


radhikaj:
@mikbras @anakrish @gupta-ak Could you please review this PR? Thanks.

Issue #1600, opened 2019-03-26
Build enclaves with more secure flags


CodeMonkeyLeet:
Per review of build options for ELF binaries, we should apply the recommended compiler flags for better security properties, and in particular, making them default in the pkg-config:

**Host:**
- [x] Use `-fstack-protector-strong` for stronger protections
- ~~Use `-fPIE` to take advantage of ASLR as a defense against ROP attacks~~
    - Per @anakrish, we will not enable this by default because the oehostapp.lib could be linked by shared libs and not just executable hosts.
- [x] Use `-z noexecstack` to mark the stack as non-executable and reduce exploitable code on the stack 
    - We may not be able to enable this if we are using trampolines.
- [x] Use `-D_FORTIFY_SOURCE=2` when `-O2` is enabled to replace some unsafe functions with checked versions where a static length can be determined.
    - Verify that this is properly supported in Clang
- ~~Use `-z now` to enable BIND_NOW so that the linker resolve all dynamically linked functions at the beginning of execution and make the global offset table (GOT) read-only. This prevents attacks from using GOT overwrites for ROP attacks.~~ 
    - There is significant startup cost overhead, so we defer this to the app dev to include at link time for their specific build instead of making it default in pkg-config or our cmake exports.
- We get partial RELRO by default with GCC/Clang. This ensures the GOT is before the BSS so overwriting global variables doesn't overwrite the GOT at least.
    - We don't need to explicitly add the linker flag for `-z relro`. 

**Enclave:**
- Use `-fstack-protector-strong` or `-fstack-protector-all`.
    - This is tracked separately with #81 which needs to be implemented for enclaves.
- [x] Use `-z noexecstack` to mark the stack as non-executable and reduce exploitable code on the stack.
- ~~Use `-D_FORTIFY_SOURCE=2` when `-O2` is enabled to replace some unsafe functions with checked versions where a static length can be determined.~~
    - This has no effect in our default compiler (Clang) and also does not work when not linked with custom string.h header as in corelibc. 
- [x] Use `-z now` to enable BIND_NOW so that the linker resolve all dynamically linked functions at the beginning of execution and make the global offset table (GOT) read-only. This prevents attacks from using GOT overwrites for ROP attacks. 
    - The switch to `-fPIE` in #1577 makes it so that enclaves generally don't emit a .rela.plt already, but we explicitly add the `-z now` linker flags to be clear and for future safety.
[...]

PR #2302, opened 2019-11-12
Fix vfprintf_chk and add other xprintf_chk functions

thomasten:
* Added missing parameter to __vfprintf_chk
* Added other ..printf_chk functions similar to the two existing ones

### DIFF
--- a/libc/__fprintf_chk.c
+++ b/libc/__fprintf_chk.c
@@ -7,18 +7,18 @@
 #include <stdarg.h>
 #include <stdio.h>
 
-// GCC sometimes replaces printf() calls with __printf_chk() calls. In glibc
+// GCC sometimes replaces fprintf() calls with __fprintf_chk() calls. In glibc
 // this function sets the output stream's _IO_FLAGS2_FORTIFY flag, which
 // causes glibc to perform various checks on the output stream. Since MUSL has
 // no equivalent flag, this implementation simply calls vfprintf().
-int __printf_chk(int flag, const char* format, ...)
+int __fprintf_chk(FILE* stream, int flag, const char* format, ...)
 {
     va_list ap;
 
     OE_UNUSED(flag);
 
     va_start(ap, format);
-    int ret = vfprintf(stdout, format, ap);
+    int ret = vfprintf(stream, format, ap);
     va_end(ap);
 
     return ret;
--- a/libc/CMakeLists.txt
+++ b/libc/CMakeLists.txt
@@ -81,7 +81,13 @@ add_library(oelibc STATIC
     regexec.c
     tre-mem.c
     __printf_chk.c
+    __fprintf_chk.c
+    __sprintf_chk.c
+    __snprintf_chk.c
+    __vprintf_chk.c
     __vfprintf_chk.c
+    __vsprintf_chk.c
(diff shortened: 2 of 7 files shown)


libc/__fprintf_chk.c
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/enclave.h>
+#include <stdarg.h>
+#include <stdio.h>
+
+// GCC sometimes replaces fprintf() calls with __fprintf_chk() calls. In glibc
```
dthaler: should this code be inside
#ifdef __GNUC__
?

radhikaj:
@thomasten. Thank you for this contribution. Can you please address feedback. We would love to be able to merge this after it gets approved.

thomasten:
> 
> 
> @thomasten. Thank you for this contribution. Can you please address feedback. We would love to be able to merge this after it gets approved.

@radhikaj I think I already addressed all feedback in 1b95835

libc/__fprintf_chk.c
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#ifdef __GNUC__
+
+#include <openenclave/enclave.h>
+#include <stdarg.h>
+#include <stdio.h>
+
+// GCC sometimes replaces fprintf() calls with __fprintf_chk() calls. In glibc
```
jhand2: What version of GCC is this and in what situations are you seeing these calls generated? If code is compiled with `-nostdlib`, calls to code in glibc shouldn't be linked, right?

thomasten: I think it happened when adding a static library to my enclave. I guess this is not supposed to be supported and one should recompile such dependencies?

jhand2: I could go either way. OE provides a standard C library, so I don't know that dependencies should all need to be compiled with nostdlib (or some other combination of compiler arguments). But supporting every internal function for every compiler could get a bit out of hand. I guess if someone wanted to target a different libc implementation in a non-enclave application they would have to recompile the dependencies in the same way.

@anakrish @CodeMonkeyLeet do either of you have an opinion here?

anakrish: GCC emits calls to these object size checking functions when compiled with _FORTIFY_SOURCE and optimizations https://gcc.gnu.org/onlinedocs/gcc/Object-Size-Checking.html

Secure coding practices recommend compiling with _FORTIFY_SOURCE (and also other flags like -fstack-protector) https://wiki.debian.org/Hardening

See https://github.com/openenclave/openenclave/issues/1600 for a list of secure flags that OE must be built with.

jhand2: Ok thats fine. For reference, BSD libc just aliases the functions (i.e. forwards __fprintf_chk to fprintf).

So I'm fine with supporting this in OE.

radhikaj:
@thomasten can you squash the commits? I can then merge this?

radhikaj:
Bors r+

jhand2:
Test failure looks like an infra issue unrelated to this PR

bors r+

PR #2303, opened 2019-11-12
Add malloc_usable_size

thomasten:


tests/memory/enc/basic.c
```diff
@@ -168,3 +169,31 @@ void test_posix_memalign(void)
...
+    if (p1 > p2)
+    {
+        int* const tmp = p1;
+        p1 = p2;
+        p2 = tmp;
+    }
+    OE_TEST(p1 < p2);
+
+    const size_t s1 = malloc_usable_size(p1);
+    OE_TEST(sizeof *p1 <= s1 && s1 <= (size_t)(p2 - p1) * sizeof *p1);
```
CodeMonkeyLeet: Isn't the upper bound just the difference of the two pointer addresses?

thomasten:
> 
> 
> @thomasten The change looks reasonable, but I'm curious about the use case you have in mind for this function. In general, `malloc_usable_size` is only intended for debugging of allocators and if you want to develop custom allocators for use inside an enclave, perhaps that's the discussion we should be having instead to better support your needs.

I've been experimenting with using RocksDB inside the enclave and it required malloc_usable_size. I didn't check what it is used for.

tests/memory/enc/basic.c
thomasten: As these are int pointers, the difference will be the number of ints in that range.

achamayou:
QuickJS needs this too.

PR #2305, opened 2019-11-12
Update dependency graph

andschwa:
I noticed that this had become out-of-date, so I updated it. It'd be
nice if we could automate this with a CI job at some point.

Signed-off-by: Andrew Schwartzmeyer <andrew@schwartzmeyer.com>


andschwa:
https://github.com/ocaml/merlin/issues/1028 is tracking Merlin's support of OCaml 4.9.

PR #2306, opened 2019-11-12
Fix TODOs that had been blocked by old OCaml

andschwa:
Starting to fix TODOs in the edger8r that were blocked by the old version of OCaml we were stuck with  due to OCPWin (before the switch to esy). Now that we're unblocked, we can clean up a lot of this.


andschwa:
bors try

andschwa:
https://github.com/ocaml/merlin/issues/1028 is tracking Merlin's support of OCaml 4.9.

PR #2309, opened 2019-11-13
Remove esy on-the-fly installation from CI containers

oprinmarius:
I built new docker images with the latest code from master, tagged them as latest and now are being used in the CI

Esy is already installed in Docker as part of the environment-setup
Removed redundant on-the-fly esy install from CI containers.


oprinmarius:
bors try

andschwa:
Also /cc @chrisrod-MSFT.

PR #2310, opened 2019-11-13
Added emailing of job status to JENKINSFILE, Packaging and libcxx_tests jenkins files

chrisrod-MSFT:
I added a global scoped variable to indicate failure - since a successful job run in the scope of finally will return null for currentBuild.result.  

That variable being null (or not) is used to determine the job's success. The email code was committed to the openenclave-ci, and for now it's configuration is global to anyone that uses that SCM library. Right now the email settings will CC: the project's default recipient list - configured in Jenkins, and send on To: the individual requesting the job. This can be changed later to include culprits or additional developers. [Extended Email Recipients](https://jenkins.io/doc/pipeline/steps/email-ext/#emailextrecipients-extended-email-recipients)

I'll need to make another pass for the remaining jobs' jenkinsfiles to add this capability once we triage what is needed.  For now this will cover release branches, bors, and sandbox jobs. See below for screenshots for examples on email:

image

image

image

### DIFF
--- a/.jenkins/Jenkinsfile
+++ b/.jenkins/Jenkinsfile
@@ -3,6 +3,7 @@ oe = new jenkins.common.Openenclave()
 
 GLOBAL_TIMEOUT_MINUTES = 240
 CTEST_TIMEOUT_SECONDS = 480
+GLOBAL_ERROR = null
 
 def ACCTest(String label, String compiler, String build_type) {
     stage("${label} ${compiler} SGX1FLC ${build_type}") {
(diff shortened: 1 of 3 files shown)


chrisrod-MSFT:
bors try

BRMcLaren:
bors try

.jenkins/Packaging.Jenkinsfile
```diff
@@ -48,11 +52,23 @@ def WindowsPackaging(String build_type) {
...
          "Windows Release" :                     { WindowsPackaging('RELEASE') }
+}catch(Exception e)
+{
+    println "Caught global pipeline exception :" + e
+    GLOBAL_ERROR = e
+    throw e   
+} finally {
+    currentBuild.result = (GLOBAL_ERROR != null) ? 'FAILURE' : "SUCCESS"    
+    oe.emailJobStatus(currentBuild.result)
+}
```
CodeMonkeyLeet: Please fix missing EOL here and in .jenkins/Jenkinsfile

.jenkins/Jenkinsfile
```diff
@@ -351,38 +355,50 @@ properties([buildDiscarder(logRotator(artifactDaysToKeepStr: '90',
...
-         "Win2016 Ubuntu1804 clang-7 Debug Linux-Elf-build" :   { win2016LinuxElfBuild('18.04', 'clang-7', 'Debug') },
-         "Win2016 Ubuntu1804 clang-7 Release Linux-Elf-build" : { win2016LinuxElfBuild('18.04', 'clang-7', 'Release') },
-         "Win2016 Ubuntu1804 gcc Debug Linux-Elf-build" :       { win2016LinuxElfBuild('18.04', 'gcc', 'Debug') },
-         "Win2016 Debug Cross Compile" :                        { win2016CrossCompile('Debug') },
-         "Win2016 Release Cross Compile" :                      { win2016CrossCompile('Release') },
-         "Win2016 Debug Cross Compile with DCAP libs" :         { win2016CrossCompile('Debug', 'ON') },
-         "Win2016 Release Cross Compile with DCAP libs" :       { win2016CrossCompile('Release', 'ON') },
-         "Host verification Debug" :                            { ACCHostVerificationTest('18.04', 'Debug') },
-         "Host verification Release" :                          { ACCHostVerificationTest('18.04', 'Release') }
+try{
```
BRMcLaren: What would happen if a single job failed?

Would it cause others to fail or would they be separate.?

.jenkins/Jenkinsfile
```diff
@@ -351,38 +355,50 @@ properties([buildDiscarder(logRotator(artifactDaysToKeepStr: '90',
...
+    oe.emailJobStatus('STARTED')
+
+    parallel "Check Developer Experience Ubuntu 16.04" :            { checkDevFlows('16.04') },
```
CodeMonkeyLeet: Nit: fix whitespace alignment

For my understanding, does a failure in a parallel job trigger a single exception that is mailed or a swath of all subsequent jobs being aborted create a race condition on the exception that actually gets captured by `GLOBAL_ERROR`?

.jenkins/Packaging.Jenkinsfile
```diff
@@ -48,11 +52,23 @@ def WindowsPackaging(String build_type) {
     }
 }
 
-parallel "1604 SGX1FLC Package Debug" :          { LinuxPackaging('1604', 'Debug') },
+try{
+    oe.emailJobStatus('STARTED')
+
+    parallel "1604 SGX1FLC Package Debug" :          { LinuxPackaging('1604', 'Debug') },
```
CodeMonkeyLeet: Nit: fix whitespace alignment

.jenkins/Packaging.Jenkinsfile
```diff
@@ -48,11 +52,23 @@ def WindowsPackaging(String build_type) {
...
          "1604 SGX1FLC Package Release" :        { LinuxPackaging('1604', 'Release') },
          "1604 SGX1FLC Package RelWithDebInfo" : { LinuxPackaging('1604', 'RelWithDebInfo') },
          "1804 SGX1FLC Package Debug" :          { LinuxPackaging('1804', 'Debug') },
          "1804 SGX1FLC Package Release" :        { LinuxPackaging('1804', 'Release') },
          "1804 SGX1FLC Package RelWithDebInfo" : { LinuxPackaging('1804', 'RelWithDebInfo') },
          "Windows Debug" :                       { WindowsPackaging('DEBUG') },
          "Windows Release" :                     { WindowsPackaging('RELEASE') }
+}catch(Exception e)
+{
```
BRMcLaren: I think we are doing the other bracket format, i.e. 

`}catch(Exception e) {
 `

.jenkins/libcxx_tests.Jenkinsfile
```diff
@@ -127,7 +133,13 @@ try {
              "libcxx ACC1804 gcc Debug" :              { ACClibcxxTest(BIONIC_LABEL, 'gcc', 'Debug') },
              "libcxx ACC1804 gcc Release" :            { ACClibcxxTest(BIONIC_LABEL, 'gcc', 'Release') },
              "libcxx ACC1804 gcc RelWithDebInfo" :     { ACClibcxxTest(BIONIC_LABEL, 'gcc', 'RelWithDebinfo') }
+}catch(Exception e){
+    println "Caught global pipeline exception :" + e
+    GLOBAL_ERROR = e
```
BRMcLaren: If we are setting Global Error and then throwing, is there a use of Global Error elsewhere?

.jenkins/Jenkinsfile
chrisrod-MSFT: By job, do you mean stage? In the case of a single stage fails - because it's part of the map in the parallel. It wouldn't impact other stages - just the status of the build, this was a run where I "simulated" a failure in multiple stages but not all of them.

image

image

BRMcLaren: Thank you for the clarification!

.jenkins/Packaging.Jenkinsfile
BRMcLaren: Yup yup

.jenkins/Jenkinsfile
BRMcLaren: Hey @CodeMonkeyLeet,

Checkout https://github.com/openenclave/openenclave/pull/2310#discussion_r345950041

I had the same questions and the job testing this scenario was posted.

johnkord:
This is neat, thanks @chrisrod-MSFT !  Do we always want email notifications for sandbox runs? We may want to give a user the ability to skip sending email notifications for sandbox runs, as a parameter.  What do you all think?

.jenkins/libcxx_tests.Jenkinsfile
chrisrod-MSFT: It's at used at the script scope level - it is only used again in the finally.

.jenkins/Packaging.Jenkinsfile
chrisrod-MSFT: Fixed, per my understanding. Is this convention/standard documented somewhere?

.jenkins/Packaging.Jenkinsfile
chrisrod-MSFT: Fixed? :)

.jenkins/Jenkinsfile
chrisrod-MSFT: The best example illustrating this behavior is from the below job sequence - which contains multiple stages run in parallel:

image

Multiple stages fail (by subscript execution in oe-ci library), while multiple stages succeed.  The email is not sent per stage failure, but only at the end.

image

image

image

.jenkins/Packaging.Jenkinsfile
CodeMonkeyLeet: @chrisrod-MSFT Usually we let clang-format take care of the formatting so we don't have to worry about it; the guideline for anything else not c/c++ is to conform to the existing formatting within the document for consistency.

.jenkins/Packaging.Jenkinsfile
chrisrod-MSFT: Thanks, should be done.

chrisrod-MSFT:
> This is neat, thanks @chrisrod-MSFT ! Do we always want email notifications for sandbox runs? We may want to give a user the ability to skip sending email notifications for sandbox runs, as a parameter. What do you all think?

Hey @johnkord it's funny I had thought about doing that as well using a pipeline parameter. I didn't want to introduce it unnecessarily, and I believe would require me to make changes to the job configuration first - then add support for it in Jenkins script.  We should look at doing this as part of a different body of work, in my opinion.

johnkord:
@chrisrod-MSFT Sounds good to me!

Issue #2314, opened 2019-11-13
Dead links in the Windows bootstrap script


jhand2:
The intel links `IntelPSWURL` and `IntelDCAPURL` in `install-windows-prereqs.ps1` are dead.

I'll handle updating them. But I am opening this issue because nobody else seemed to notice. Do we not use these scripts in any CI or anything like that?

jhand2:
Closed by https://github.com/openenclave/openenclave/pull/2316

PR #2316, opened 2019-11-13
Updated references (docs and scripts) to use new SGX version

chrisrod-MSFT:
1. Updated scripts and docs that referenced SGX versions to new versions in intel registration center
2. Uploaded and then referenced oejenkins Azure storage container files in scripts.
3. Updated changelog.


chrisrod-MSFT:
@johnkord does this look about right?

jhand2:
Looks good. We can close #2314 when this PR is merged

chrisrod-MSFT:
I believe we are holding on this PR as samples on Windows are not working. Will provide update as we have more information.

chrisrod-MSFT:
We should be good to go on this. Fresh installs using scripts on ACC VMs (Ubuntu 1604, 1804 and Windows 2016) with all tests passing.  Will need someone in @openenclave/committers to remove the label and merge.

PR #2317, opened 2019-11-14
Tell git to ignore ctags files

jhand2:
Tell git to ignore ctags files used by some editors for code navigation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 *.swp
 *.swo
 
+tags
+
 .vs/
 .vscode/
 graphviz/


andschwa:
@jhand2 would this be better suited to each developer's own global .gitignore? I feel like if we try to include every dev environment's gitignore in the repo it might get annoying.

Also, you may want to take a look at the generated CMake compile commands JSON file, it can be used the likes of LSP (with the clangd backend) pretty easily in just bout every editor. If you use Emacs or VS Code I'd be happy to show you how, and once upon a time I had it setup in Vim too.

jhand2:
> @jhand2 would this be better suited to each developer's own global .gitignore? I feel like if we try to include every dev environment's gitignore in the repo it might get annoying.

Good point, I'll abandon. Also I'll look into setting up lsp in vim. Looks pretty straightforward

andschwa:
Thanks @jhand2!

PR #2322, opened 2019-11-16
Port tls_e2e test host from pthread to std::thread

jhand2:
Port the tls_e2e test host side to use std::thread instead of pthread. This will allow the test to run on Windows, which does not officially support pthread.

Note that this change alone will not get the test running on Windows, but it is necessary groundwork.

### DIFF
--- a/tests/tls_e2e/host/host.cpp
+++ b/tests/tls_e2e/host/host.cpp
@@ -155,8 +155,6 @@ void run_server(void* arg)
 
     OE_TRACE_INFO("Leaving server thread...\n");
     fflush(stdout);
-done:
-    return NULL;
 }
 
 void run_client(void* arg)
@@ -222,8 +220,6 @@ void run_client(void* arg)
     }
 
     fflush(stdout);
-done:
-    return NULL;
 }
 
 int run_test_with_config(tls_test_configs_t* test_configs)
@@ -235,9 +231,8 @@ int run_test_with_config(tls_test_configs_t* test_configs)
 
     {
         // Release lock on scope exit
-        std::unique_lock<std::mutex> l(g_server_mutex);
-        while (!g_server_condition)
-            g_server_cond.wait(l);
+        std::unique_lock<std::mutex> server_lock(g_server_mutex);
+        g_server_cond.wait(server_lock, []{ return g_server_cond; });
     }
 
     fflush(stdout);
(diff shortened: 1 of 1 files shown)


radhikaj:
bors delegate=jhand2

radhikaj:
Bors try

tests/tls_e2e/host/host.cpp
```diff
@@ -60,20 +64,22 @@ typedef enum test_config_type
...
 oe_enclave_t* g_client_enclave = NULL;
+
 int g_server_thread_exit_code = 0;
 int g_client_thread_exit_code = 0;
-pthread_mutex_t server_mutex;
-pthread_cond_t server_cond;
 bool g_server_condition = false;
-pthread_t server_thread_id;
+
+std::mutex server_mutex;
```
anakrish: It is better to use a consistent naming convention `g_` as the variables above seem to be doing.

tests/tls_e2e/host/host.cpp
```diff
@@ -149,19 +155,17 @@ void* server_thread(void* arg)
 
     OE_TRACE_INFO("Leaving server thread...\n");
     fflush(stdout);
-    pthread_exit((void*)&g_server_thread_exit_code);
 done:
     return NULL;
 }
 
-void* client_thread(void* arg)
+void* run_client(void* arg)
```
anakrish: Do we need these `void*` returns now that we are using C++?

tests/tls_e2e/host/host.cpp
```diff
@@ -216,52 +220,32 @@ void* client_thread(void* arg)
...
-        oe_put_err("pthread_create(server): ret=%u", ret);
+    server_thread = std::thread(run_server, (void*)test_configs);
 
     OE_TRACE_INFO("wait until TLS server is ready to accept client request\n");
-    pthread_mutex_lock(&server_mutex);
+    std::unique_lock<std::mutex> l(server_mutex);
     while (!g_server_condition)
-        pthread_cond_wait(&server_cond, &server_mutex);
-    pthread_mutex_unlock(&server_mutex);
+        server_cond.wait(l);
```
anakrish: Can you double-check the logic here? Previously, upon return from pthread_cond_wait, the lock was immediately unlocked via a pthread_mutex_unlock.
In this new version, we don't seem to unlock the mutex immediately upon return from wait.

jhand2: Oh, you're right, I misread the `condition_variable::wait` and thought the lock was released when the condition was met, but thats not the case. I'll fix

jhand2:
bors try

jhand2:
bors try

jhand2:
bors try

jhand2:
Random test failed
```
The following tests FAILED:
	121 - tests/crypto/enclave (Failed)
	122 - tests/crypto/host (Child aborted)
```

Rebased and trying again

bors try

PR #2324, opened 2019-11-18
Optimization/Security: Reduce TCB via gc-sections.

anakrish:
Function and variables are put in their own sections (-ffunction-sections and -fdata-sections).
Linker is asked to garbage collect unused sections (-gc-sections).
Thus any function/variable that is not reachable from the enclave entry point is eliminated
from the enclave binary, reducing the footprint and tcb.

For example, release build of echo_enc previously took 970KB. Now it takes 650KB.


anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

PR #2335, opened 2019-11-20
Implement enclave exit in C

anakrish:
- oe_exit_enclave is now implemented in C with minimal inline assembly
- oe_exit_enclave is used to ecall return, abort as well as ocall.
  Previously the assembly code was duplicated in enter.S for ecall return.
- oe_exit_enclave is passed the information where to return to in host
  rather than fetching it from td. This allows it to be called in different
  places.
- oe_exit_enclave is complied with -fno-omit-frame-pointer to allow stitching
  the ocall stack to the oe_exit_enclave's frame.
- td does not clear the self-pointer. This allows oe_get_td to succeed.
- Added OE_NO_RETURN def
- Deleted lots of assembly code.


anakrish:
bors try

enclave/core/sgx/exit.c
```diff
@@ -0,0 +1,169 @@
...
+
+//==============================================================================
+//
+// This function is wrapper of _oe_exit_enclave. It is needed to stitch the host
+// stack and enclave stack together.
+//
+// oe_exit_enclave is #defined to __morestack.
+//
+// N.B: Don't change the function name, otherwise debugger can't work. GDB
+// depends on this hardcode function name when does stack walking for split
```
radhikaj: probable typo: hardcoded function name

radhikaj:
thanks @anakrish for making this code maintainable!

enclave/core/CMakeLists.txt
```diff
@@ -191,6 +191,12 @@ if (OE_SGX)
 
     set_source_files_properties(sgx/keys.c PROPERTIES COMPILE_FLAGS -Wno-type-limits)
 
+    # exit.c must be compiled with -fno-omit-frame-pointer so that the ocall stack
+    # can be stitched back to the frame of the _oe_exit_enclave function.
+    # Use -O2 to always generate optimal code irrespective of the build type.
```
jhand2: What are the consequences of not using `-O2` for all builds?

enclave/core/sgx/exit.c
```diff
@@ -0,0 +1,169 @@
...
+        "cmp $1, %%rax \n\t"
+        "je 2f \n\t"
+        "1: enclu \n\t"
+        "2: jmp *%%rbx \n\t"
+        "ud2 \n\t"
+        :
+        : "r"(rdi), "r"(rsi), "r"(rax), "r"(rbx), "r"(rcx), "r"(rdx));
+
+    // This function does not return.
+    while (1)
```
jhand2: What is the benefit of putting a while loop here if the code is supposed to be unreachable?

enclave/core/CMakeLists.txt
jhand2: What are the consequences of not using -O2 for all builds?

enclave/core/sgx/exit.c
```diff
@@ -0,0 +1,169 @@
...
+        "xor %%rcx, %%rcx \n\t"
+        "xor %%rdx, %%rdx \n\t"
+
+        // Execute either ENCLU or jmp based on RAX value
+        // which will be 1 for simulation mode. See above.
+        "cmp $1, %%rax \n\t"
+        "je 2f \n\t"
+        "1: enclu \n\t"
+        "2: jmp *%%rbx \n\t"
+        "ud2 \n\t"
```
jhand2: I don't think adding `ud2` here really adds much benefit. Maybe an assert at the end of the function for debug builds only would be sufficient.

enclave/core/sgx/exit.c
```diff
@@ -0,0 +1,169 @@
...
+    register uint64_t rax __asm__("rax") = td->simulate ? 1 : ENCLU_EEXIT;
+
+    // RSP and RBP must be restored to host values. The values are passed in
+    // to the assembly snippet in RCX and RDX.
+    register uint64_t rcx __asm__("rcx") = host_rsp;
+    register uint64_t rdx __asm__("rdx") = host_rbp;
+
+    asm volatile(
+        // Clear general purpose registers.
+        "xor %%r8, %%r8 \n\t"
```
jhand2: nit: Ending each instruction with `;` rather than ` \n\t` might be a little cleaner.

enclave/core/sgx/exit.c
jhand2: If the idea is to inform the compiler that the function doesn't return, you can just annotate it as OE_NO_RETURN

anakrish:
Closing this PR due to lack of traction.
Will recreate PR at a later date.

PR #2338, opened 2019-11-21
Remove workaround in dune build of Intel sources

andschwa:
We had set `wrapped` to `false` such that the libraries built by dune
from the Intel sources did not require the `Intel.` prefix. Now we can
remove this and fix up the few types we actually need to use manually.

Signed-off-by: Andrew Schwartzmeyer <andrew@schwartzmeyer.com>


andschwa:
bors merge

Thanks @anakrish!

PR #2339, opened 2019-11-22
Debuggers robustness: Rely on sgx_tcs_t rather than OE specific td_from_tcs

anakrish:
Also removed unused function oe_notify_ocall_end

### DIFF
--- a/debugger/pythonExtension/gdb_sgx_plugin.py
+++ b/debugger/pythonExtension/gdb_sgx_plugin.py
@@ -78,13 +78,13 @@ class oe_debug_enclave_t:
 
 
 # This constant definition must align with sgx_tcs_t
-TCS_GSBASE_OFFSET =  0X38
+TCS_GSBASE_OFFSET =  56
 
 # This constant definition must align with TD structure in internal\sgxtypes.h.
-TD_CALLSITE_OFFSET = 0XF0
+TD_CALLSITE_OFFSET = 240
 
 # This constant definition must align with Callsite structure in enclave\td.h.
-CALLSITE_OCALLCONTEXT_OFFSET = 0X40
+CALLSITE_OCALLCONTEXT_OFFSET = 64
 
 # These constant definitions must align with OCallContext structure in enclave\td.h.
 OCALLCONTEXT_LENGTH = 2 * 8
--- a/debugger/pythonExtension/gdb_sgx_plugin.py
+++ b/debugger/pythonExtension/gdb_sgx_plugin.py
@@ -77,8 +77,8 @@ class oe_debug_enclave_t:
         return self.magic == self.MAGIC_VALUE
 
 
-# This constant definition must align with the OE enclave layout.
-TD_OFFSET_FROM_TCS =  0X5000
+# This constant definition must align with sgx_tcs_t
+TCS_GSBASE_OFFSET =  0X38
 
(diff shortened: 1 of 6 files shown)


anakrish:
bors try

debugger/pythonExtension/gdb_sgx_plugin.py
```diff
@@ -77,8 +77,8 @@ def is_valid(self):
         return self.magic == self.MAGIC_VALUE
 
 
-# This constant definition must align with the OE enclave layout.
-TD_OFFSET_FROM_TCS =  0X4000
+# This constant definition must align with sgx_tcs_t
+TCS_GSBASE_OFFSET =  0X38
```
johnkord: Maybe we should change this to an integer since this represents a byte offset in the sgx_tcs_t struct. When I first went through this I was very confused because 38 didn't seem to add up to get to the `gsbase` element, and then I realized this number is hexadecimal for 56! It was my mistake of course, I just usually associate byte offsets with decimal numbers! 😊

debugger/pythonExtension/gdb_sgx_plugin.py
```diff
@@ -268,8 +268,11 @@ def stop(self):
         # Get untrusted stack frame pointer and corresponding TCS.
         frame_pointer = int(gdb.parse_and_eval("$rdi"))
         tcs_addr = int(gdb.parse_and_eval("$rsi"))
+        enclave_base_addr = int(gdb.parse_and_eval("$rdx"))
+        gs_base = read_int_from_memory(tcs_addr + TCS_GSBASE_OFFSET, POINTER_SIZE)
+
         # Get callsite of the TCS.
-        td_addr = tcs_addr + TD_OFFSET_FROM_TCS
+        td_addr = enclave_base_addr + gs_base
```
johnkord: Very clean change. Do you happen to know if there is some history as to why we used a different method originally?

anakrish:
bors try

debugger/pythonExtension/gdb_sgx_plugin.py
dthaler: C/C++ have things like the offsetof() macro (https://en.cppreference.com/w/cpp/types/offsetof).
Is there anything usable in Python rather than hardcoding a possibly fragile number?

anakrish:
bors try

debugger/pythonExtension/gdb_sgx_plugin.py
anakrish: @johnkord Nice suggestion. Done.

anakrish: @dthaler There is nothing readily available in python to calculate the offset of a C struct's field. 

Though the offset is hard-coded, the accompanying test asserts that the hard-coded offset is equal to the offset determined by the C compiler. Offset mismatch would fail the test.

In the future, these offsets will be part of the [debugger binary contract](https://github.com/openenclave/openenclave/blob/master/include/openenclave/internal/debugrt/host.h) which is a versioned contract between the debugger and the SDK. Most of the debugger has already been moved to the contract; just a few pending pieces we are trying to address as per scheduling.

debugger/pythonExtension/gdb_sgx_plugin.py
anakrish: @johnkord I am not sure about the history.

PR #2340, opened 2019-11-23
Refactor Emitter.ml into multiple modules

andschwa:
Our edger8r became a large, single file of over a thousand lines of OCaml code. This is a rough refactor into the modules `Common.ml`, `Header.ml`, and `Sources.ml`. It's not yet complete, and it introduces a few things to fix, but it's a working start.

- [x] Move `oe_gen_args_h` to `Header.ml`
- [x] Rename functions to make sense e.g. `Sources.generate_trusted` instead of `gen_u_c`
- [x] Factor out the `ep` and `ec` references, preferring to plumb through the necessary info as cleanly as possible (I think this might mean introducing an `oe_flags` type as part of the `edger8r_params`)
- [x] Try to shrink `Sources.ml` as it contains a lot of code still

/cc @anakrish


andschwa:
bors merge

PR #2342, opened 2019-11-25
Windows posix sockets

jhand2:
Implement socket syscalls for Windows to support a TLS between two enclaves on Windows. This is only implemented as much as is needed to run the TLS test at tests/tls_e2e. Additional testing is likely necessary.

Special thanks to @yakman2020 for implementing error mapping and @anakrish for implementing most of the ocalls.

My apologies that this PR has grown so much. Please let me know if you would prefer I split it out.


tests/tls_e2e/host/host.cpp
```diff
@@ -269,7 +269,9 @@ int run_scenarios_tests()
          {false, false, true},
          NEGATIVE_TEST}};
 
+#ifndef _WIN32
     signal(SIGPIPE, SIG_IGN);
+#endif
```
jhand2: Not sure why why this is needed in the linux test but the Windows test passes fine without it, Needs more investigation.

tests/syscall/socket/host/host.c
```diff
@@ -109,13 +115,18 @@ char* host_client(in_port_t port)
...
             sock_close(sockfd);
             return NULL;
         }
-        else
+#if _WIN32
+        else if (HOST_SOCKET_ERRNO == WSAEISCONN)
+        {
+            break;
+        }
+#endif
```
jhand2: This can likely be cleaned up now that errors are converted to BSD errors

syscall/devices/hostresolver/hostresolver.c
```diff
@@ -102,7 +102,7 @@ static int _hostresolver_getaddrinfo(
                 &retval, node, service, hints, &handle) != OE_OK)
         {
             ret = OE_EAI_SYSTEM;
-            OE_RAISE_ERRNO(OE_EINVAL);
+            OE_RAISE_ERRNO(oe_errno);
```
jhand2: There may have been a reason to swallow the error in this file. Need to look into this.

host/windows/syscall.c
```diff
@@ -20,16 +19,163 @@
...
+    {WSAELOOP, OE_ELOOP},
+    {WSAENAMETOOLONG, OE_ENAMETOOLONG},
+    {WSAEHOSTDOWN, OE_EHOSTDOWN},
+    {WSAEHOSTUNREACH, OE_EHOSTUNREACH},
+    {WSAENOTEMPTY, OE_ENOTEMPTY},
+    {WSAEUSERS, OE_EUSERS},
+    {WSAEDQUOT, OE_EDQUOT},
+    {WSAESTALE, OE_ESTALE},
+    {WSAEREMOTE, OE_EREMOTE},
+    {WSAEDISCON, 199},
```
dthaler: Why hard code numbers here instead of creating OE_* defines?

host/windows/syscall.c
```diff
@@ -20,16 +19,163 @@
...
+                                                    {7, SO_SNDBUF},
+                                                    {8, SO_RCVBUF},
+                                                    {9, SO_KEEPALIVE},
+                                                    {10, SO_OOBINLINE},
+                                                    {13, SO_LINGER},
+                                                    {18, SO_RCVLOWAT},
+                                                    {19, SO_SNDLOWAT}};
+
+// From netdb.h
+#define EAI_BADFLAGS -1
```
dthaler: These defines are already defined in ws2tcpip.h with different values, so this is confusing in a Windows file.
I don't think we should redefine standard values here in this file.  If you need to have different numerical values for EAI_MEMORY etc than Windows defines, then I think this should have a different name, like OE_EAI_MEMORY or something.

host/windows/syscall.c
```diff
@@ -20,16 +19,163 @@
...
+#define EAI_AGAIN -3
+#define EAI_FAIL -4
+#define EAI_FAMILY -6
+#define EAI_SOCKTYPE -7
+#define EAI_SERVICE -8
+#define EAI_MEMORY -10
+#define EAI_SYSTEM -11
+#define EAI_OVERFLOW -12
+
+static struct tab_entry wsa2eai[] = {{WSATRY_AGAIN, EAI_AGAIN},
```
dthaler: and here you can map the windows values to your custom values, like
{EAI_AGAIN, OE_EAI_AGAIN}
since EAI_AGAIN is already #defined as WSATRY_AGAIN.

tests/syscall/socket/host/host.c
```diff
@@ -16,6 +16,12 @@
 
 #define SERVER_PORT "12345"
 
+#if __linux__
+#define HOST_SOCKET_ERRNO errno
+#elif _WIN32
+#define HOST_SOCKET_ERRNO WSAGetLastError()
```
dthaler: Why can't you just
#define errno WSAGetLastError()
?

tests/syscall/socket/host/host.c
jhand2: Never mind, this is on the host side.

common/oe_host_socket.h
```diff
@@ -0,0 +1,126 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#ifndef _OE_HOST_SOCKET_H
+#define _OE_HOST_SOCKET_H
+
+#include <openenclave/corelibc/bits/types.h>
```
anakrish: Headers need to be wrapped within a `OE_EXTERNC_BEGIN .. OE_EXTERNC_END` pair so that the header can be inculded from C++ as well.

common/oe_host_socket.h
```diff
@@ -0,0 +1,126 @@
...
+#include <winsock2.h>
+#else
+#include <sys/socket.h>
+#endif
+
+#define GETADDRINFO_HANDLE_MAGIC 0xed11d13a
+
+typedef struct _getaddrinfo_handle
+{
+    uint32_t magic;
```
anakrish: Normally we use a uint64_t magic.

common/oe_host_socket.h
```diff
@@ -0,0 +1,126 @@
...
+
+    if (!err_no)
+    {
+        goto done;
+    }
+
+    if (!handle || !ai_flags || !ai_family || !ai_socktype || !ai_protocol ||
+        !ai_addrlen || !ai_canonnamelen)
+    {
+        *err_no = OE_EINVAL;
```
anakrish: Does posix mandate that in these scenarios, EINVAL code must be returned. It would be good to document why we are returning specific error codes.

common/oe_host_socket.h
```diff
@@ -0,0 +1,126 @@
...
+{
+    getaddrinfo_handle_t* handle = (getaddrinfo_handle_t*)handle_;
+
+    if (!handle || handle->magic != GETADDRINFO_HANDLE_MAGIC || !handle->res)
+        return NULL;
+
+    return handle;
+}
+
+OE_INLINE int _getaddrinfo_read(
```
anakrish: Why does this function need to be inline?

common/oe_host_socket.h
```diff
@@ -0,0 +1,126 @@
...
+            goto done;
+        }
+
+        if (*ai_canonnamelen > ai_canonnamelen_in)
+        {
+            *err_no = OE_ENAMETOOLONG;
+            goto done;
+        }
+
+        memcpy(ai_addr, p->ai_addr, *ai_addrlen);
```
radhikaj: Should we use memcpy_s here instead?

host/windows/syscall.c
```diff
@@ -312,9 +462,87 @@ int oe_syscall_rmdir_ocall(const char* pathname)
...
+
+static oe_socket_fd_t _invalid_socket = {OE_SOCKET_FD_MAGIC, INVALID_SOCKET};
+
+oe_host_fd_t _make_socket_fd(SOCKET sock)
+{
+    oe_host_fd_t fd = (oe_host_fd_t)&_invalid_socket;
+    if (sock != INVALID_SOCKET)
+    {
+        oe_socket_fd_t* socket_fd =
+            (oe_socket_fd_t*)malloc(sizeof(oe_socket_fd_t));
```
radhikaj: check for null here

host/windows/syscall.c
```diff
@@ -312,9 +462,87 @@ int oe_syscall_rmdir_ocall(const char* pathname)
...
+
+SOCKET _get_socket(oe_host_fd_t fd)
+{
+    oe_socket_fd_t* socket_fd = (oe_socket_fd_t*)fd;
+    if (socket_fd && socket_fd->magic == OE_SOCKET_FD_MAGIC)
+        return socket_fd->socket;
+    return INVALID_SOCKET;
+}
+
+static oe_host_fd_t _dup_socket(oe_host_fd_t oldfd)
```
anakrish: Is there a better way to clone a socket in Windows? I had put this in as a placeholder.

host/windows/syscall.c
```diff
@@ -567,7 +908,54 @@ int oe_syscall_getaddrinfo_open_ocall(
...
+        _set_errno(OE_EINVAL);
+        goto done;
+    }
+
+    _set_errno(0);
+
+    if (handle_out)
+        *handle_out = 0;
+
+    if (!handle_out)
```
radhikaj: wrap this into an else with the if above?

host/windows/syscall.c
```diff
@@ -723,5 +1144,25 @@ int oe_syscall_getgroups_ocall(size_t size, unsigned int* list)
 
 int oe_syscall_uname_ocall(struct oe_utsname* buf)
 {
-    PANIC;
+    // Based on
+    // https://docs.microsoft.com/en-us/windows/win32/sysinfo/getting-the-system-version
+    // OE SDK is supported only on WindowsServer and Win10
+    if (IsWindowsServer())
```
anakrish: Is there a better implementation of uname?

common/oe_host_socket.h
jhand2: getaddrinfo_read isn't a real syscall, its just used as a helper to allow the enclave to get the next record. I mostly just took the existing code from host/linux/syscall.c and moved it here so it could be reused.

In general, EINVAL means invalid argument. If the caller passes in null pointers, I think this is that right error to return.

common/oe_host_socket.h
jhand2: It doesn't. Will remove.

samples/attested_tls/server/enc/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) Open Enclave SDK contributors.
 # Licensed under the MIT License.
 
+find_package(PythonInterp 3 REQUIRED)
```
jhand2: Make sure python3 is included in ansible/windows ps scripts to bootstrap.

common/oe_host_socket.h
anakrish: Since there is the concept of `next record`, is this function thread-safe?

samples/attested_tls/scripts/gen_mrenclave_header.py
```diff
@@ -0,0 +1,60 @@
+# Copyright (c) Open Enclave SDK contributors.
```
jhand2: @radhikaj @anakrish this was originally a bash script and I ported it to python so it could run as part of both the Windows and Linux builds.

In general, is there a preference for what language should be used for cross-platform build scripts? Ideally we settle on just one so we don't have too many build-time dependencies.

tests/syscall/socket/enc/enc.c
```diff
@@ -172,7 +172,6 @@ int ecall_run_server()
             printf("enc: accept failed errno = %d \n", oe_errno);
         }
     }
-
```
jhand2: Add back newline

tests/syscall/platform/windows.h
```diff
@@ -20,6 +20,7 @@ typedef SOCKET socket_t;
 typedef int socklen_t;
 typedef int length_t;
 typedef void pthread_attr_t;
+typedef uint16_t in_port_t;
```
jhand2: not sure what this is. Remove

common/oe_host_socket.h
jhand2: It isn't. The handle passed in is retrieved with getaddrinfo_open ocall and a new structure is allocated for each open. So if two threads call open, their read calls will be on different structures.

If one thread calls open and then two threads call read on that same handle, then yes there is a race condition.

getaddrinfo is only exposed to the enclave as an atomic operation which calls open, read, and free, then returns the data to the caller. So I don't think it needs to be threadsafe.

host/windows/syscall.c
jhand2: Was there a reason you populated the information the way you did? For instance, if I populated it like this:

sysname: Windows
version: 10 (or Server2016, Server2019, etc.)
release: <windows build number>

Where is this information going to be used?

host/windows/syscall.c
jhand2: It sort of depends on what the expectation is. There is a function WSADuplicateSocket but it is designed for sharing sockets across processes. But that isn't the purpose of dup in Linux (I'm not really sure I understand the purpose of dup actually).

I'll do some more research.

dthaler: https://stackoverflow.com/questions/11635219/dup2-dup-why-would-i-need-to-duplicate-a-file-descriptor

host/windows/syscall.c
```diff
@@ -312,9 +450,94 @@ int oe_syscall_rmdir_ocall(const char* pathname)
...
+static oe_socket_fd_t _invalid_socket = {OE_SOCKET_FD_MAGIC, INVALID_SOCKET};
+
+oe_host_fd_t _make_socket_fd(SOCKET sock)
+{
+    oe_host_fd_t fd = (oe_host_fd_t)&_invalid_socket;
+    if (sock != INVALID_SOCKET)
+    {
+        oe_socket_fd_t* socket_fd =
+            (oe_socket_fd_t*)malloc(sizeof(oe_socket_fd_t));
+        socket_fd->magic = OE_SOCKET_FD_MAGIC;
```
dthaler: missing check for malloc failure

host/windows/syscall.c
```diff
@@ -312,9 +450,94 @@ int oe_syscall_rmdir_ocall(const char* pathname)
...
+    return -1;
+}
+
+static int _wsa_startup()
+{
+    static bool wsa_init_done = FALSE;
+    WSADATA wsaData;
+    int ret = 0;
+
+    if (!wsa_init_done)
```
dthaler: this code looks ok for TrustZone which is single-threaded, but does not look thread-safe for SGX

host/windows/syscall.c
```diff
@@ -723,5 +1181,25 @@ int oe_syscall_getgroups_ocall(size_t size, unsigned int* list)
 
 int oe_syscall_uname_ocall(struct oe_utsname* buf)
 {
-    PANIC;
+    // Based on
+    // https://docs.microsoft.com/en-us/windows/win32/sysinfo/getting-the-system-version
+    // OE SDK is supported only on WindowsServer and Win10
+    if (IsWindowsServer())
+    {
+        sprintf(buf->nodename, "(unknown)");
```
dthaler: nodename should be set to whatever gethostname() returns

tests/tls_e2e/host/host.cpp
jhand2: winsock treats a broken socket as a normal error and calls will just result in normal WSA errors.

host/windows/syscall.c
jhand2: Used WSADuplicateSocket with current process. As I understand it, the goal of dup is that if one fd is closed, the duplicate is not closed.

I need to make sure WSADuplicateSocket actually creates a reference to the same socket, rather than just making a separate socket with the same parameters.

host/windows/syscall.c
jhand2: Used oe_atomic_compare_and_swap

host/windows/syscall.c
jhand2: Used GetComputerNameEx

host/windows/syscall.c
andschwa: I don't know if it helps, but this was our implementation back in Apache Mesos: https://github.com/apache/mesos/blob/5c50403eb233991eaa8f83b5bc50eae2c33d5850/3rdparty/stout/include/stout/os/windows/dup.hpp#L25

jhand2: Ok cool, that is also what I did.

I found this in Win32 docs that confirms it does what I'm expecing:

> All of the state information associated with a socket is held in common across all the descriptors because the socket descriptors are duplicated and not the actual socket.

jhand2:
bors try

jhand2:
bors try

jhand2:
bors try

jhand2:
bors try

jhand2:
bors try

jhand2:
Ok hopefully last time

bors try

jhand2:
@radhikaj Could you merge? Thanks!

andschwa:
@jhand2 Try again! Sorry, I'd added you to the committers doc but forgot to update the teams.

jhand2:
@andschwa thanks!

bors r+

PR #2351, opened 2019-11-29
Fix oe_accept not returning peer addr

thomasten:
The posix accept() function takes a pointer to a sockaddr struct that it fills with the address of the peer socket. The oe_accept() function passes a buffer to the host call to receive the peer address, but on return the result is not used. This PR adds the code to copy the result to the user provided buffer so that oe_accept() behaves likes accept().

### DIFF
--- a/syscall/devices/hostsock/hostsock.c
+++ b/syscall/devices/hostsock/hostsock.c
@@ -273,10 +273,10 @@ static oe_fd_t* _hostsock_accept(
         new_sock->host_fd = retval;
 
         // copy peer addr to out buffer
-        if (addrlen && *addrlen <= addrlen_in)
+        if (addrlen)
         {
             oe_assert(addr);
-            if (oe_memcpy_s(addr, addrlen_in, &buf, *addrlen) != OE_OK)
+            if (oe_memcpy_s(addr, addrlen_in, &buf.addr, *addrlen) != OE_OK)
                 OE_RAISE_ERRNO(OE_EINVAL);
         }
     }
--- a/syscall/devices/hostsock/hostsock.c
+++ b/syscall/devices/hostsock/hostsock.c
@@ -271,6 +271,14 @@ static oe_fd_t* _hostsock_accept(
             OE_RAISE_ERRNO_MSG(oe_errno, "retval=%d", retval);
 
         new_sock->host_fd = retval;
+
+        // copy peer addr to out buffer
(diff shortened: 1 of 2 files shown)


jhand2:
Thanks for your PR! Please provide a PR description describing the issue you found and the approach you've taken to fix it.

syscall/devices/hostsock/hostsock.c
```diff
@@ -271,6 +271,14 @@ static oe_fd_t* _hostsock_accept(
             OE_RAISE_ERRNO_MSG(oe_errno, "retval=%d", retval);
 
         new_sock->host_fd = retval;
+
+        // copy peer addr to out buffer
+        if (addrlen && *addrlen <= addrlen_in)
+        {
+            oe_assert(addr);
+            if (oe_memcpy_s(addr, addrlen_in, &buf, *addrlen) != OE_OK)
```
jhand2: Use `&buf.addr` for clarity

syscall/devices/hostsock/hostsock.c
```diff
@@ -271,6 +271,14 @@ static oe_fd_t* _hostsock_accept(
             OE_RAISE_ERRNO_MSG(oe_errno, "retval=%d", retval);
 
         new_sock->host_fd = retval;
+
+        // copy peer addr to out buffer
+        if (addrlen && *addrlen <= addrlen_in)
```
jhand2: Isn't the `*addrlen <= addrlen_in` check redundant? memcpy_s already checks the sizes.

tests/syscall/socket/enc/enc.c
```diff
@@ -144,7 +144,16 @@ int ecall_run_server()
     {
         oe_sleep_msec(1);
         printf("enc: accepting\n");
-        connfd = oe_accept(listenfd, (struct oe_sockaddr*)NULL, NULL);
+
+        struct oe_sockaddr_in peer_addr = {0};
+        oe_socklen_t peer_addr_len = sizeof peer_addr;
```
jhand2: Prefer `sizeof` with parentheses for consistency with the rest of the codebase

syscall/devices/hostsock/hostsock.c
thomasten: Yes, that's right. I removed it.

jhand2:
Looks good, thanks!

bors r+

PR #2353, opened 2019-12-02
Fix race condition in epoll

thomasten:
One thread can add or delete file descriptors to/from an epoll instance
while another thread is waiting on it. This was not handled correctly.


andschwa:
@mikbras Can you look at this?

syscall/devices/hostepoll/hostepoll.c
```diff
@@ -457,26 +465,32 @@ static int _epoll_wait(
...
-
-            if (!mapping)
-                OE_RAISE_ERRNO(OE_ENOENT);
         }
     }
 
     ret = (int)retval;
 
 done:
+    if (locked)
```
radhikaj: Thanks for this fix. I am wondering why you decided to move from spinlock to mutex.

thomasten: I changed it because the lock is now acquired before calling oe_syscall_epoll_ctl_ocall(). That seemed to be a too expensive operation to me. However, I didn't do any measuring to justify this change. What do you think?

radhikaj: Thanks!

tests/syscall/epoll/enc/enc.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+extern "C" void trigger_and_delete_event()
+{
+    _send(1);
+
+    // delete fd from the epoll instance
+    OE_TEST(epoll_ctl(_epfd, EPOLL_CTL_DEL, _sockfd, nullptr) == 0);
+}
+
+extern "C" void cancel_wait()
```
jhand2: Why is this called `cancel_wait`? That doesn't look like what the function is doing.

syscall/devices/hostepoll/hostepoll.c
jhand2: I do agree that an ocall (and subsequent kernel call) seems like too expensive of an operation to hold a spinlock. I think this is the right approach.

@mikbras can you take a look at this as well? I'm a little worried about holding a lock and then making an ocall since threads can be re-entrant and cause deadlock. It makes me nervous but in this case epoll doesn't try to re-enter the enclave for any reason so my fear is probably unwarranted.

syscall/devices/hostepoll/hostepoll.c
```diff
@@ -457,26 +465,32 @@ static int _epoll_wait(
...
 
-            oe_spin_lock(&epoll->lock);
+            if (mapping)
+                event->data.u64 = mapping->event.data.u64;
+            else
             {
-                if ((mapping = _map_find(epoll, event->data.fd)))
-                    event->data.u64 = mapping->event.data.u64;
+                // fd has been deleted between the return of epoll_wait and the
+                // acquisition of the lock.
```
jhand2: I think in this case we should still return all the events. When the wait was triggered the event fd was still registered. Let me know your thoughts. Are there any detrimental effects of doing this?

tests/syscall/epoll/enc/enc.cpp
thomasten: How about `send_stop_request`?

syscall/devices/hostepoll/hostepoll.c
thomasten: Problem is that we cannot assign the real event data (`event->data.u64 = mapping->event.data.u64`) if the fd has been deleted.

jhand2: oh good point. I don't love shifting the whole array left for every deleted event. Instead can we replace the current element with the last element? Does ordering of the events matter to the caller?

syscall/devices/hostepoll/hostepoll.c
```diff
@@ -457,26 +465,32 @@ static int _epoll_wait(
...
+                --retval;
```
jhand2: Do we also need to decrement `i` here? I think if we don't we'll end up skipping the next event.

syscall/devices/hostepoll/hostepoll.c
jhand2: If order does matter its probably not a huge deal. This is pretty unlikely to happen and very unlikely to happen more than once. But for an application that calls epoll a lot on a lot of handles it could make a difference.

tests/syscall/epoll/enc/enc.cpp
jhand2: Sorry I should have been more clear. I don't understand how the below code would cancel the thread from waiting. Doesn't it just add a wait on a handle and then trigger it? Why would epoll stop waiting? `wait_for_events` has a `while (1)` loop so I don't see why it would ever stop unless one of the OE_TEST calls fails.

Excuse my likely ignorance of epoll

tests/syscall/epoll/enc/enc.cpp
```diff
@@ -0,0 +1,115 @@
...
+    uint8_t run = 1;
+
+    while (run)
+    {
+        epoll_event event{};
+
+        int n;
+        do
+        {
+            n = epoll_wait(_epfd, &event, 1, -1);
```
jhand2: How do you know that the `epoll_wait` ever returned? I think the test would be more complete if you kept a variable that tracks how many events were triggered in total check it at the end.

tests/syscall/epoll/enc/enc.cpp
thomasten: The code was a little cryptic, I admit. Improved.

syscall/devices/hostepoll/hostepoll.c
thomasten: AFAIK order does not matter. I agree replacing with last element is better.

syscall/devices/hostepoll/hostepoll.c
thomasten: Good catch! Fixed.

tests/syscall/epoll/enc/enc.cpp
thomasten: See changes of my last push: The wait thread must have read all the `action_t::run`s before it can read `action_t::stop` and end.

syscall/devices/hostepoll/hostepoll.c
jhand2: looks good, thanks

jhand2:
Looks like there was an issue with our build pipeline. Trying again.

bors r+

PR #2356, opened 2019-12-03
Fix opendir() does not return NULL on error

thomasten:
oe_opendir() always returned a valid pointer even if the host call
returned NULL.


jhand2:
bors try

jhand2:
Thanks!

bors r+

PR #2365, opened 2019-12-04
Add interfaces to edger8r

andschwa:
With the refactor into modules, this adds OCaml module interfaces, which explicitly define which functions are public in the modules (think C headers).


andschwa:
@radhikaj Could you review and sign-off?

andschwa:
Thanks @mikbras!

bors merge

PR #2369, opened 2019-12-05
Fix some of the memory safety issues reported by scan-build

jhand2:
In exploring static analysis tools, I ran Clang's scan-build static analysis tool on the codebase. I took a first pass at fixing some of the low-hanging fruit.

Specifically I took a look at:
* Use-after-free
* Double-free
* Dereference of null ptr
* A couple memory leaks

### DIFF
--- a/syscall/devices/hostfs/hostfs.c
+++ b/syscall/devices/hostfs/hostfs.c
@@ -433,8 +433,10 @@ static int _hostfs_dup(oe_fd_t* desc, oe_fd_t** new_file_out)
     file_t* file = _cast_file(desc);
     file_t* new_file = NULL;
 
-    if (new_file_out)
-        *new_file_out = NULL;
+    if (!new_file_out)
+        OE_RAISE_ERRNO(OE_EINVAL);
+
+    *new_file_out = NULL;
 
     /* Check parameters. */
     if (!file)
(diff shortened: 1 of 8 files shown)


jhand2:
bors try

jhand2:
Rebased against master to fix build error

bors try

syscall/devices/hostfs/hostfs.c
```diff
@@ -433,6 +433,9 @@ static int _hostfs_dup(oe_fd_t* desc, oe_fd_t** new_file_out)
     file_t* file = _cast_file(desc);
     file_t* new_file = NULL;
 
+    if (!new_file_out)
+        OE_RAISE_ERRNO(OE_EINVAL);
+
     if (new_file_out)
```
dthaler: I believe line 439 can be removed since with the new code above, this line cannot be reached unless new_file_out is non-null

syscall/fdtable.c
```diff
@@ -283,6 +283,9 @@ int oe_fdtable_reassign(int fd, oe_fd_t* new_desc, oe_fd_t** old_desc)
     int ret = -1;
     bool locked = false;
 
+    if (!new_desc || !old_desc)
+        OE_RAISE_ERRNO(OE_EINVAL);
+
 #if !defined(NDEBUG)
     _assert_fd(new_desc);
 #endif
```
dthaler: Line 293 can be removed since 286-287 mean you can't get here without the condition being true

PR #2370, opened 2019-12-06
change required approvers from 1 to 2

BRMcLaren:
Change the required reviewers from 1 to 2.

We should be requiring 2 approvers given cross-collaboration efforts.

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -1,6 +1,6 @@
 status = [ "continuous-integration/jenkins/branch" ]
 pr_status = [ "DCO" ]
-required_approvals = 1
+required_approvals = 2
 block_labels = [ "do not merge" ]
 delete_merged_branches = true
 timeout_sec = 7200 # two hours


BRMcLaren:
bors try

BRMcLaren:
@andschwa 

I agree with your statement. That being said with the upcoming cross-collaboration efforts, requiring two approvals may be better for stability purposes. Something reasonable would be 2 approvals, at least 1 being a Committer would be a good idea.

It's not a question of trust, in my opinion, it's a question of how fast we are growing and leveraging that extra experience to our advantage by promoting knowledge sharing.

Do you believe this should be brought up with the CGC? We can always revert if it is becoming a major hindrance, I believe the recommended approach right now is to get 2 approvers even if it is not forced by tooling.

jhand2:
I’m also on board with this. Plus the operational burden can be overcome by simply being more intentional about reviewing PRs. Especially small ones that don’t require a lot of time to look into.

andschwa:
> Do you believe this should be brought up with the CGC?

In my opinion, yes.

BRMcLaren:
Sure.

@andschwa @dthaler are both part of the committee and have different viewpoints, could you bring it up at the next meeting and get a vote?

andschwa:
Added to new business!

johnkord:
It may be worthwhile to have two approvals required before merge simply for account security reasons

BRMcLaren:
Thanks, @andschwa !

I had added the do-not-merge label to avoid accidental merging, which I guess we now know works!

Removing label and merging.

bors r+

PR #2371, opened 2019-12-06
Fix oegdb on Ubuntu with German locale

thomasten:
oegdb could not get symbols from enclave file on Ubuntu with a German
locale. This commit updates load_symbol_cmd.py to the current master of
intel/linux-sgx related to intel/linux-sgx#405


jhand2:
> Looks good, but why change the copyright for Intel?

This is just copied directly from intel/sgx-linux. See this issue that Thomas linked: https://github.com/intel/linux-sgx/issues/405

And link to the file in sgx-linux: https://github.com/intel/linux-sgx/blob/master/sdk/debugger_interface/linux/gdb-sgx-plugin/load_symbol_cmd.py

PR #2374, opened 2019-12-09
Fix Windows warnings exposed by /W1 accross the codebase

jhand2:
Also enables /W1 and /WX to ensure new W1 violations are not added. The
end goal is to enable /W3 /WX for all Windows builds.

### DIFF
--- a/cmake/compiler_settings.cmake
+++ b/cmake/compiler_settings.cmake
@@ -88,7 +88,10 @@ elseif (MSVC)
 
   # Explicitly set C/CXX flags rather than using the defaults. This uses the defaults
   # but removes /W3 from CMAKE_C(XX)_FLAGS. Using W3 and W1 together adds many warnings
-  # that W3 is being overwritten by W1
+  # that W3 is being overwritten by W1. W3 as a default flag is removed in cmake 3.15,
+  # so this behavior can be removed if/when cmake_minimum_required is raised to 3.15.
+  # ======= Default compiler flags for cmake version 3.12 can be found here: =======
+  # https://github.com/Kitware/CMake/blob/v3.12.0/Modules/Platform/Windows-MSVC.cmake
   set (CMAKE_C_FLAGS "/DWIN32 /D_WINDOWS")
   set (CMAKE_C_FLAGS_DEBUG "/MDd /Zi /Ob0 /Od /RTC1")
   set (CMAKE_C_FLAGS_RELEASE "/MD /O2 /Ob2 /DNDEBUG")
--- a/cmake/compiler_settings.cmake
+++ b/cmake/compiler_settings.cmake
@@ -16,6 +16,8 @@ if (NOT CMAKE_C_COMPILER_ID STREQUAL CMAKE_CXX_COMPILER_ID)
     "${CMAKE_C_COMPILER_ID} != ${CMAKE_CXX_COMPILER_ID}")
 endif ()
 
+set(CMAKE_C_STANDARD 11)
+
 # Set the default standard to C++14 for all targets.
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
@@ -72,7 +74,6 @@ if (CMAKE_CXX_COMPILER_ID MATCHES GNU OR CMAKE_CXX_COMPILER_ID MATCHES Clang)
   # and that are easy to avoid. Treat at warnings-as-errors, which forces developers
   # to fix warnings as they arise, so they don't accumulate "to be fixed later".
   add_compile_options(-Wall -Werror -Wpointer-arith -Wconversion -Wextra -Wno-missing-field-initializers)
-
   add_compile_options(-fno-strict-aliasing)
 
   # Enables XSAVE intrinsics
@@ -84,11 +85,36 @@ elseif (MSVC)
   if (MSVC_VERSION VERSION_LESS 1910)
     message(FATAL_ERROR "Only Visual Studio 2017 and above supported!")
   endif ()
+
+  # Explicitly set C/CXX flags rather than using the defaults. This uses the defaults
+  # but removes /W3 from CMAKE_C(XX)_FLAGS. Using W3 and W1 together adds many warnings
+  # that W3 is being overwritten by W1
+  set (CMAKE_C_FLAGS "/DWIN32 /D_WINDOWS")
+  set (CMAKE_C_FLAGS_DEBUG "/MDd /Zi /Ob0 /Od /RTC1")
+  set (CMAKE_C_FLAGS_RELEASE "/MD /O2 /Ob2 /DNDEBUG")
+
+  set (CMAKE_CXX_FLAGS "/DWIN32 /D_WINDOWS /GR /EHsc")
+  set (CMAKE_CXX_FLAGS_DEBUG "/DWIN32 /D_WINDOWS /MDd /Zi /Ob0 /Od /RTC1")
+  set (CMAKE_CXX_FLAGS_RELEASE "/MD /O2 /Ob2 /DNDEBUG")
+
+  # Can't use add_compile_options because it adds for all file types and ml64
+  # doesn't recognize /wd flags
+  # Turns off warnings for:
+  # * Unicode character cannot be represented by current code page
+  set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /wd4566")
+  set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4566")
+
+  # Add Flags we want to use for both C and CXX
+  add_compile_options(/WX)
+  add_compile_options(/W1)
+
+  # Ignore compiler warnings:
+  # * unicode character not supported
+
   if (CMAKE_MSVC_PARALLEL_ENABLE)
     add_compile_options(/MP)
     message(STATUS "Using parallel compiling (/MP)")
   endif()
-
 endif ()
 
 # Use ML64 as assembler on Windows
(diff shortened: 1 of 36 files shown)


jhand2:
bors try

jhand2:
Fixed some inconsistencies between uint64_t format strings on Windows and Linux

bors try

cmake/compiler_settings.cmake
```diff
@@ -16,6 +16,8 @@ if (NOT CMAKE_C_COMPILER_ID STREQUAL CMAKE_CXX_COMPILER_ID)
     "${CMAKE_C_COMPILER_ID} != ${CMAKE_CXX_COMPILER_ID}")
 endif ()
 
+set(CMAKE_CXX_STANDARD 99)
```
andschwa: Doesn't this change to 14 three lines further down?

cmake/compiler_settings.cmake
```diff
@@ -84,11 +85,33 @@ elseif (MSVC)
   if (MSVC_VERSION VERSION_LESS 1910)
     message(FATAL_ERROR "Only Visual Studio 2017 and above supported!")
   endif ()
+
+  # Explicitly set C/CXX flags rather than using the defaults. This uses the defaults
```
andschwa: We should link to the defaults.

host/CMakeLists.txt
```diff
@@ -223,6 +223,7 @@ list(APPEND PLATFORM_HOST_ONLY_SRC
   ../common/safecrt.c
   hexdump.c
   dupenv.c
+  fopen.c
```
andschwa: File getting adding?

cmake/compiler_settings.cmake
jhand2: whoops, this should be CMAKE_C_STANDARD

host/CMakeLists.txt
jhand2: on purpose

cmake/compiler_settings.cmake
jhand2: The comment explains which defaults I left off. I'll add why, but it is very much intentional.

jhand2: Added comment with justification and added back RTC1. All errors that triggered RTC failures have been fixed.

jhand2:
Fixed pragma that fails a couple of the Windows builds

bors try

jhand2:
Now that C99 is properly enforced, need to fix a couple instances where C11 features are used.

jhand2:
> Now that C99 is properly enforced, need to fix a couple instances where C11 features are used.

Getting OE to compile against C99 is non-trivial. Opened issue  #2376 and will not tackle in this PR.

bors try

cmake/compiler_settings.cmake
andschwa: I think we should link to wherever you found the defaults (I usually consult the CMake source code for the version we're tied to). But it's not an easy thing for most people to find.

jhand2: I just printed the values. I’ll look for a source file and link it

jhand2: Added more details and source file for reference.

host/crypto/bcrypt/pem.c
```diff
@@ -339,4 +337,4 @@ oe_result_t oe_get_next_pem_cert(
         free(found_pem);
 
     return result;
-}
```
jxyang: ditto

host/crypto/bcrypt/ec.c
```diff
@@ -96,8 +96,8 @@ static oe_result_t _fixup_public_key_info_for_import(
          * offsets from the original */
         new_key_info = (PCERT_PUBLIC_KEY_INFO)new_key_info_buffer;
         new_key_info->Algorithm.pszObjId =
-            new_key_info_buffer +
-            (key_info->Algorithm.pszObjId - key_info_base);
+            (LPSTR)new_key_info_buffer +
```
dthaler: I believe the LP types are obsolete and these should just be (PSTR).   (Longer explanation: as noted in https://stackoverflow.com/questions/2315311/what-is-a-long-pointer, the LP types are a holdover from Win16 days, and OE won't run on Win16 anyway, so we should use current types instead of obsolete ones.)

tools/oesign/oedump.c
```diff
@@ -179,7 +180,7 @@ void dump_enclave_properties(const oe_sgx_enclave_properties_t* props)
     bool debug = props->config.attributes & OE_SGX_FLAGS_DEBUG;
     printf("debug=%u\n", debug);
 
-    printf("xfrm=%lx\n", props->config.xfrm);
+    printf("xfrm=%" PRIx64 "\n", props->config.xfrm);
 
     printf(
         "num_heap_pages=%llu\n",
```
dthaler: shouldn't this use PRIu64 instead of %llu?

host/sgx/elf.c
```diff
@@ -284,7 +282,7 @@ int elf64_load(const char* path, elf64_t* elf)
     if (fd == -1 || _fstat64(fd, &statbuf) != 0)
         goto done;
 
-    if (!(statbuf.st_mode & _S_IFREG) != 0)
+    if (!(statbuf.st_mode & _S_IFREG))
```
jxyang: Isn't this equivalent to `if ((statbuf.st_mode & _S_IFREG) == 0)`? It's not encouraged to implicitly convert integer to boolean.

tools/oesign/oedump.c
jhand2: In this case, it is wrapped in the macro `OE_LLU` which does some weird cast that I don't quite understand (see include/openenclave/internal/types.h) and the warning doesn't get triggered.

In general it is better to just use PRIu64 and friends. But in the interest of keeping this PR focused I think its best to leave anything that doesn't trigger a warning.

host/sgx/elf.c
jhand2: Yes, it is equivalent. This triggers a warning due to double negation. I could change it to `if ((statbuf.st_mode & _S_IFREG) == 0)`.

For my knowledge, why is it not encouraged to use an int in an if condition?

common/sgx/revocation.c
```diff
@@ -496,4 +496,4 @@ oe_result_t oe_validate_revocation_list(
     oe_cert_free(&tcb_cert);
 
     return result;
-}
```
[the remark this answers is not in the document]
jhand2: This is just my editor adding a newline at the end of the file as is outlined in POSIX. Generally on Linux, editors will enforce this, but on Windows they sometimes don't. I looked at the hexdump of the file and it ends with a `0x0a` (newline) character and nothing else weird.

See: https://thoughtbot.com/blog/no-newline-at-end-of-file

host/crypto/bcrypt/pem.c
jhand2: See above

host/sgx/elf.c
jxyang: Some compilers with some flags warn about implicit boolean conversion. For example `-Wint-in-bool-context`.

host/crypto/bcrypt/ec.c
```diff
@@ -96,8 +96,8 @@ static oe_result_t _fixup_public_key_info_for_import(
          * offsets from the original */
         new_key_info = (PCERT_PUBLIC_KEY_INFO)new_key_info_buffer;
         new_key_info->Algorithm.pszObjId =
-            new_key_info_buffer +
-            (key_info->Algorithm.pszObjId - key_info_base);
+            (PSTR)new_key_info_buffer +
+            (key_info->Algorithm.pszObjId - (LPSTR)key_info_base);
```
dthaler: (PSTR) here too

tools/oesign/oedump.c
dthaler: Ah I see, that's fine.  On GNUC compilers, uint64_t is "unsigned long" and on MSVC it's "unsigned long long" (both defined in include/openenclave/bits/types.h).   So OE_LLU casts the unsigned long value to unsigned long long for GNUC and doesn't need anything for MSVC.   Using %llu allows the OE_LLU trick whereas PRIx64 requires no cast but instead uses %llu on MSVC and %I64u on GNUC to make the type always match.

cmake/compiler_settings.cmake
andschwa: Thanks!

host/sgx/elf.c
```diff
@@ -80,15 +80,13 @@ static bool _is_valid_elf64(const elf64_t* elf)
...
-            sizeof(header->e_phentsize) + sizeof(header->e_phnum),
-        "e_phentsize or e_phnum is too large");
+        sizeof(header->e_phentsize) + sizeof(header->e_phnum));
 
-    static_assert(
+    OE_STATIC_ASSERT(
         sizeof(uint64_t) >=
-            sizeof(header->e_shentsize) + sizeof(header->e_shnum),
-        "e_shentsize or e_shnum is too large");
+        sizeof(header->e_shentsize) + sizeof(header->e_shnum));
```
jxyang: Is there a reason you removed the message from the assertion?

jhand2: OE_STATIC_ASSERT doesn't support messages. But the file and line number from the failure should be sufficient to deduce what went wrong. It will also print the contents of the assertion in the  build error.

jhand2: I removed static_assert because Mike mentioned a desire to use C99 and this is the only place where static_assert is used (rather than OE_STATIC_ASSERT). Although technically OE_STATIC_ASSERT doesn't really work on C99 either.

jhand2:
Thanks all for the review!

bors r+

PR #2375, opened 2019-12-09
check for unsigned int added by Ryan Hsu 12/6/2019

ryanhsu19:
Added a check for unsigned int in the file samples/file-encryptor/enclave/keys.cpp

### DIFF
--- a/samples/file-encryptor/enclave/keys.cpp
+++ b/samples/file-encryptor/enclave/keys.cpp
@@ -290,9 +290,17 @@ int ecall_dispatcher::prepare_encryption_header(
     {
         TRACE_ENCLAVE("password_key");
         for (unsigned int i = 0; i < ENCRYPTION_KEY_SIZE_IN_BYTES; i++)
+        {
+            if (password_key[i] > UINT32_MAX)
+            {
+                TRACE_ENCLAVE(
+                    "password key is too large to fit into an unsigned int", 1);
+                goto exit;
+            }
             TRACE_ENCLAVE(
                 "password_key[%d] =0x%02x", i, (unsigned int)(password_key[i]));
-        goto exit;
+            goto exit;
+        }
     }
 
     // produce a encryption key


samples/file-encryptor/enclave/keys.cpp
```diff
@@ -290,9 +290,17 @@ int ecall_dispatcher::prepare_encryption_header(
     {
         TRACE_ENCLAVE("password_key");
         for (unsigned int i = 0; i < ENCRYPTION_KEY_SIZE_IN_BYTES; i++)
+        {
+            if (password_key[i] > UINT32_MAX)
```
jhand2: `password_key[i]` is a 1 byte unsigned char. It isn't possible to be larger than UINT32_MAX.

I am guessing you're meaning to check if all of password key (which is an array of bytes) can fit into a uint32?

Even then, isn't this check redundant? ENCRYPTION_KEY_SIZE_IN_BYTES is 32, so by definition, it can fit into a uint32

jhand2:
In general, `added by <name> <date>` in the commit message isn't necessary, git already keeps track of this information in the commit.

ryanhsu19:
Hi Jordan,

Thanks for the note. I will keep it in mind when I commit next time.
As for the cast for unsigned int cast, I think you are right that it is kind of redundant. Therefore, I don’t think I need this request anymore.

Ryan

samples/file-encryptor/enclave/keys.cpp
radhikaj: ENCRYPTION_KEY_SIZE_IN_BYTES  = 32 is the size of the buffer. Since we are casting password+key[i] to unsigned it, they check looks right to me. @jhand2 could you please elaborate here? Thanks.

jhand2:
Closing this PR based on

>As for the cast for unsigned int cast, I think you are right that it is kind of redundant. Therefore, I don’t think I need this request anymore.

jhand2:
Whoops, pr needs some changes but I was wrong about the PR not being needed

samples/file-encryptor/enclave/keys.cpp
jhand2: My mistake, the first thing I said was right but the part about it always fitting into a uint32 was wrong.

Still I’m a bit confused about what we’re checking here. As is this just checks that each byte in password_key fits into a uint32 which will always be true.

dthaler: I agree with Jordan, this change doesn't make sense to me either

ryanhsu19:
Hi Dave,

Yes, I would like to cancel the pull request, while I have reverted the commit to my repo. I was going to follow the unsigned int cast check on another file but found that this was redundant in this case.

Thank you!

Sincerely,
Ryan

jhand2:
> Yes, I would like to cancel the pull request

cc @radhikaj are you fine if I close this again? Is there an issue opened to help me understand what problem we're trying to solve?

PR #2377, opened 2019-12-10
Attempt to handle syscalls without ocall first

### LINKED ISSUE #2285: Spurious syscall warnings
Some syscalls trigger spurious warning messages. The warnings are emitted here: https://github.com/openenclave/openenclave/blob/0deb7c67911e22b3151997f4f631ed5cf4ad7e30/syscall/syscall.c#L815

However, some system calls _are_ handled, but elsewhere:
https://github.com/openenclave/openenclave/blob/0deb7c67911e22b3151997f4f631ed5cf4ad7e30/libc/syscalls.c#L257

In our case, this message is triggered rather frequently because of calls to `clock_gettime`. Further, the log level is set to `OE_LOG_LEVEL_ERROR` initially, until global initialization is done. Thus, the exact number and the timing of those warning messages varies and makes it a little hard to debug.

In the interest of wasted debugging time, I'd like to suggest to change this behavior. For example, those four syscalls could be handled before `_dispatch_oe_syscall` is called, or that function could have logging disabled temporarily, until it is known whether there is a real warning to report, e.g. if `errno != ENOSYS` at this point: https://github.com/openenclave/openenclave/blob/0deb7c67911e22b3151997f4f631ed5cf4ad7e30/libc/syscalls.c#L248

CC @letmaik

jhand2:
Only attempt to handle syscalls via ocall after an attempt has been made to handle it internally. This ordering change reduces warning messages indicating a syscall is unhandled when in fact it is handled, just later on.

Fixes #2285

### DIFF
--- a/libc/syscalls.c
+++ b/libc/syscalls.c
@@ -235,6 +235,22 @@ long __syscall(long n, long x1, long x2, long x3, long x4, long x5, long x6)
         /* The hook ignored the syscall so fall through */
     }
 
+    /* Handle syscall internally if possible. */
+    switch (n)
+    {
+        case SYS_nanosleep:
+            return _syscall_nanosleep(n, x1, x2);
+        case SYS_gettimeofday:
+            return _syscall_gettimeofday(n, x1, x2);
+        case SYS_clock_gettime:
+            return _syscall_clock_gettime(n, x1, x2);
+        case SYS_mmap:
+            return _syscall_mmap(n, x1, x2, x3, x4, x5, x6);
+        default:
+            /* Drop through and let the code below handle the syscall. */
+            break;
+    }
+
     /* Let liboesyscall handle select system calls. */
     {
         long ret;
@@ -247,31 +263,11 @@ long __syscall(long n, long x1, long x2, long x3, long x4, long x5, long x6)
         {
             return ret;
         }
-
-        /* Drop through and let the code below handle the syscall. */
-        errno = 0;
     }
 
-    switch (n)
-    {
-        case SYS_nanosleep:
-            return _syscall_nanosleep(n, x1, x2);
-        case SYS_gettimeofday:
-            return _syscall_gettimeofday(n, x1, x2);
-        case SYS_clock_gettime:
-            return _syscall_clock_gettime(n, x1, x2);
-        case SYS_mmap:
-            return _syscall_mmap(n, x1, x2, x3, x4, x5, x6);
-        default:
-        {
-            /* All other MUSL-initiated syscalls are aborted. */
-            fprintf(stderr, "error: __syscall(): n=%lu\n", n);
-            abort();
-            return 0;
-        }
-    }
-
-    return 0;
+    /* All other MUSL-initiated syscalls are aborted. */
+    fprintf(stderr, "error: unhandled syscall: n=%lu\n", n);
+    abort();
 }
 
 /* Intercept __syscalls_cp() from MUSL */


libc/syscalls.c
```diff
@@ -235,23 +235,7 @@ long __syscall(long n, long x1, long x2, long x3, long x4, long x5, long x6)
...
-        }
-
-        /* Drop through and let the code below handle the syscall. */
-        errno = 0;
-    }
-
+    /* Handle syscall internally if possible. */
     switch (n)
     {
         case SYS_nanosleep:
```
johnkord: Is there a reason these aren't included in _dispatch_oe_syscall() and separated by a switch here instead? (Not specifically directed at @jhand2 but a question for anyone!)

johnkord: The SYS_nanosleep case seems to make an OCALL, but the others seem to not. It makes reasonable sense to me to just bundle these cases all up with the _dispatch_oe_syscall() function, and that function is the one place in the code to refer to for how the enclave syscall handler behaves! What do you all think? :)

jhand2: I suppose the difference here is that all the syscalls in `_dispatch_oe_syscall` all eventually call `oe_syscall` and make an ocall.

You're right that `SYS_nanosleep` also makes an ocall, but it does it directly (with oe_ocall) rather than making it via `oe_syscall`. nanosleep is a little weird because it calls the `sleep` ocall, which is not actually the same as POSIX sleep (it takes in milliseconds rather than seconds), which in turn on linux calls nanosleep. `oe_sleep` just converts seconds to milliseconds and then calls this function. The Windows `Sleep` function also takes in milliseconds.

So one option for refactoring this code would be to have two syscalls handled by oe_syscall, sleep and nanosleep. On Linux they will call the correct respective syscalls. On Windows, they will both call Sleep, and convert to milliseconds when needed. I would leave `SYS_gettimeofday`, `SYS_clock_gettime`, `SYS_mmap` alone though since none of them make ocalls.

jhand2: I opened an issue here for this: #2405

I'd like to keep this PR targeted since it fixes a specific bug.

libc/syscalls.c
```diff
@@ -263,15 +247,27 @@ long __syscall(long n, long x1, long x2, long x3, long x4, long x5, long x6)
         case SYS_mmap:
             return _syscall_mmap(n, x1, x2, x3, x4, x5, x6);
         default:
+            /* Drop through and let the code below handle the syscall. */
```
mingweishih: s/handle/handles

libc/syscalls.c
```diff
@@ -263,15 +247,27 @@ long __syscall(long n, long x1, long x2, long x3, long x4, long x5, long x6)
...
+            break;
+    }
+
+    /* Let liboesyscall handle select system calls. */
```
mingweishih: s/handle select system calls/handles selected syscalls

libc/syscalls.c
```diff
@@ -235,23 +235,7 @@ long __syscall(long n, long x1, long x2, long x3, long x4, long x5, long x6)
...
-        if (!(ret == -1 && errno == ENOSYS))
-        {
-            return ret;
-        }
-
-        /* Drop through and let the code below handle the syscall. */
-        errno = 0;
-    }
-
+    /* Handle syscall internally if possible. */
```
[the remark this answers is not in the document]
jhand2: Singular is intentional. It means "Handle this particular syscall that is currently being dispatched"

libc/syscalls.c
jhand2: Singular is correct

libc/syscalls.c
jhand2: Singular is correct.

jhand2:
cc @CodeMonkeyLeet @anakrish @nonpolarity for review. Thanks!

libc/syscalls.c
CodeMonkeyLeet: @jhand2 I agree with you, though I think the inconsistencies in existing plurals on line 217 and 273 is creating the confusion. Consider fixing them up.

PR #2379, opened 2019-12-12
Add instructions for running scan-build on Linux

jhand2:
I have come up with some instructions for running scan-build on the OE codebase. I intend to start categorizing the bugs and opening issues for certain bug classes so others can help tackle them in smaller chunks.

The end goal once we get scan-build errors to zero is to run it in CI/CD and block PRs that introduce new bugs.

### DIFF
--- a/docs/GettingStartedDocs/Contributors/RunningStaticAnalysis.md
+++ b/docs/GettingStartedDocs/Contributors/RunningStaticAnalysis.md
@@ -1,0 +1,42 @@
+# Getting started with running static analysis tools
+
+## scan-build
+
+*Note: scan-build only works with clang and gcc and as such is only currently supported for linux builds*
+
+1. Install scan-build-7 (or whatever version matches your version of clang)
+
+    ```
+    $ apt install clang-tools-7
+    ```
+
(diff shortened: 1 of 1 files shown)


docs/GettingStartedDocs/Contributors/RunningStaticAnalysis.md
```diff
@@ -0,0 +1,42 @@
...
+3. Run scan-build for cmake and make build commands
+
+    For example:
+
+    ```{bash}
+    $ scan-build cmake .. -DHAS_QUOTE_PROVIDER=OFF
+    $ scan-build make -j`nproc`
+    ```
+
+*Note: scan-build also works fine with the cmake ninja generator*
```
andschwa: We generally prefer Ninja as it's required for Windows, and works much better on Linux than GNU Make. I'd recommend just updating the example to use Ninja.

ryanhsu19:
I just committed to my own repo and the number of bugs is now down to 150 on my repo. It also passed ctest. However Alvin and I had trouble looking for the format problem - git commit shows many format problems, while format-code shows nothing and, therefore, no change on git status. 

I will continue to investigate why, but shall I pull request from it?

I also found that most data-store and memory problems are false positives, due to scan-build not knowing what some OE_-macros are.

PR #2382, opened 2019-12-16
Remove path resolution for source in mounting external hostfs

jhand2:
The enclave's internal current working directory has no inherent relation to the external hostfs device. As such, resolving a relative source path using the internal cwd provides odd behavior.

To simplify mounting of hostfs, we will only allow absolute source paths.

### DIFF
--- a/syscall/mount.c
+++ b/syscall/mount.c
@@ -151,7 +151,7 @@ int oe_mount(
         target = target_path.buf;
     }
 
-    /* Note: Normialization of source path is left to the external device
+    /* Note: Normalization of source path is left to the external device
      * as it may not be a path internal to the enclave.
      */
 
--- a/syscall/mount.c
+++ b/syscall/mount.c
@@ -166,7 +166,12 @@ int oe_mount(
         struct oe_stat buf;
         int retval = -1;
 
-        if ((retval = oe_stat(target, &buf)) != 0)
+        /**
+         * oe_stat tries to do a mount resolution, but the directory is not yet
+         * mounted. As a result, we must call the filesystem's stat
+         * implementation directly.
+         */
+        if ((retval = device->ops.fs.stat(device, target, &buf)) != 0)
             OE_RAISE_ERRNO(oe_errno);
 
         if (!OE_S_ISDIR(buf.st_mode))
--- a/syscall/mount.c
+++ b/syscall/mount.c
@@ -137,7 +137,6 @@ int oe_mount(
     oe_device_t* device = NULL;
     oe_device_t* new_device = NULL;
     bool locked = false;
-    oe_syscall_path_t source_path;
     oe_syscall_path_t target_path;
     mount_point_t mount_point = {0};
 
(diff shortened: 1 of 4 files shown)


syscall/mount.c
```diff
@@ -152,14 +151,9 @@ int oe_mount(
...
 
-    /* Normalize the source path if any. */
-    if (source)
-    {
-        if (!oe_realpath(source, &source_path))
-            OE_RAISE_ERRNO(OE_EINVAL);
-
-        source = source_path.buf;
-    }
+    /* Note: Normialization of source path is left to the external device
```
radhikaj: Typo: normalization

tests/syscall/hostfs/enc/enc.c
```diff
@@ -18,6 +19,19 @@ void test_hostfs(const char* tmp_dir)
         exit(1);
     }
 
+    /* Mount with a relative source should fail */
+    if (mount(".", "/", OE_HOST_FILE_SYSTEM, 0, NULL) == 0)
```
radhikaj: Is there a test to ensure that after we mount an absolute path, fopen with relative path succeeds? I am just trying to figure out if we have tests to cover the recently reported issue with fopen

radhikaj:
@mikbras we would appreciate your comments on this PR.

tests/syscall/hostfs/enc/enc.c
jhand2: I added a test to `tests/syscall/fs` to check this

jhand2:
bors try

syscall/mount.c
```diff
@@ -172,7 +166,12 @@ int oe_mount(
         struct oe_stat buf;
         int retval = -1;
 
-        if ((retval = oe_stat(target, &buf)) != 0)
+        /**
+         * oe_stat tries to do a mount resolution, but the directory is not yet
+         * mounted. As a result, we must call the filesystem's stat
+         * implementation directly.
+         */
+        if ((retval = device->ops.fs.stat(device, target, &buf)) != 0)
```
jhand2: @johnkord @radhikaj My test caught a bug where oe_mount doesn't actually work with any target other than "/" because oe_stat requires the path the be mounted but paths in the process of being mounted are not yet mounted :smile:

Requesting new review in light of this new change.

johnkord: Oh that's a good catch!

tests/syscall/fs/enc/enc.cpp
```diff
@@ -637,17 +637,23 @@ void test_fs(const char* src_dir, const char* tmp_dir)
         OE_TEST(
             oe_mount("/", "/", OE_DEVICE_NAME_HOST_FILE_SYSTEM, 0, NULL) == 0);
         const int flags = OE_O_CREAT | OE_O_TRUNC | OE_O_WRONLY;
-        OE_TEST(oe_open(path, flags, MODE) == 0);
+        OE_TEST(oe_open(path, flags, MODE) != -1);
```
johnkord: Good change on checking return value of oe_open, I didn't even notice this the first time around!

johnkord: And good change to all the other expected return values too!
Hmm. We might want to add function headers to each of those functions that specify expected return types. (Not in this PR) I guess they're supposed to be equivalent to libc function calls, though.

jhand2:
Thanks!

bors r+

jhand2:
I guess debug builds can detect enclave memory leaks (very cool!)

Anyway, fixed the leaks.

bors r+

Issue #2373, opened 2019-12-09
Heap Data Measurement


eozturk1:
According to [Intel](https://software.intel.com/en-us/articles/innovative-technology-for-cpu-based-attestation-and-sealing), `MRENCLAVE` value consists of:

- The **contents** of the pages (**code, data, stack, heap**).
- The relative position of the pages in the enclave.
- Any security flags associated with the pages.

It seems OpenEnclave [does not include](https://github.com/openenclave/openenclave/blob/71edd2cf92f52017410d5e3262628b962067560a/host/sgx/create.c#L146) heap page contents in the `MRENCLAVE` value. What is the reason for this design choice?

mikbras:
Both Intel and Microsoft decided not to measure the heap pages in their respective SDKs. A correct, well-behaved program will never read from the initial heap page contents. However, programs that fail to properly initialize memory returned from **sbrk**, **malloc**, and **realloc** are potentially vulnerable to attack. So this attack vector should be closed. But how?

Measuring the pages on enclave load is a non-starter since the heap can be much larger than the EPC memory. For example, an enclave can have a 16GB heap while the total EPC memory might be 128MB. Measuring heap pages would result in extreme thrashing during loading due to swapping. That is probably why Intel decided not to measure heap pages.

One solution is to zero initialize all memory obtained from **sbrk** but only when it results in a new high. Since **dlmalloc** is based solely on **sbrk**, this will incrementally initialize heap memory on demand. This solution trades off performance for security.

mikbras:
I created the following PR to address this issue:

https://github.com/openenclave/openenclave/pull/2383

It guarantees that all memory obtained with **oe_sbrk()** and hencse **malloc** and **realloc** will be zero filled.

eozturk1:
Thank you for looking into this @mikbras! Interesting to see the documentation does not match the implementation on Intel's end. Can you please close the issue when the PR is merged?

jhand2:
We might consider turning the sbrk PR off by default to avoid perf penalty.

douglasism:
@mikbras Hi Mike, are you sure about your characterization in the initial reply? https://github.com/openenclave/openenclave/issues/2373#issuecomment-566306591

They way I understand the reason to not extend the heap page measurements is that the pages are set to 0 before they are added. This is a performance enhancement without a security risk given that the pages are filled with zero.

jhand2:
As @douglasism pointed out, the heap pages are all zero'd by the loader.

[...]

PR #2383, opened 2019-12-17
Modify sbrk to zero fill allocated space

mikbras:
This PR addresses issue #2373 which reports a security vulnerability resulting from a design decision not to initialize and measure heap pages during enclave loading. Please see the issue for background information.

This PR modifies **oe_sbrk()** to zero-fill memory incrementally as it is allocated. The PR also fixes the unhandled case where the increment parameter is negative.


enclave/core/sbrk.c
```diff
@@ -1,32 +1,89 @@
...
+        ret = _break;
+        _break += increment;
+
+        if (_break > _max)
         {
-            ptr = _heap_next;
-            _heap_next += increment;
+            /* Zero-initializes newly allocated memory. */
+            memset(_max, 0, (size_t)(_break - _max));
+            _max = _break;
```
radhikaj: Thanks a lot for this PR @mikbras
Would using memset_s be good?

mingweishih:
I'm still not sure we need to initialize memory in sbrk as the use of heap memory usually depends on the behavior of the program (e.g., malloc v.s. calloc). It would be great to firstly check potential attacks rooted from this issue. For example, one case I'm thinking of is that the attacker creates a new enclave, allocates the memory overlapped with the heap of the victim enclave, and reads the leftover content. I'm wondering would this be the case, or are there other potential attack vectors.

jhand2:
Do we (or sgx implementation) clear pages when they are released from EPC? If not that is probably worse that the vulnerabilities this protects against.

I talked with MingWei about this and he has a good point that this isn't really a TEE specific mitigation. Using uninitialized memory is just a general security bug, regardless of whether it runs in SGX. Maybe as OE we should help developers write secure code in general, but we should discuss if that is a goal or not.

We also talked about the perf penalty. We should probably come up with a way to profile this code to determine how large of a penalty we're adding for enclave applications.

CodeMonkeyLeet:
> Do we (or sgx implementation) clear pages when they are released from EPC? If not that is probably worse that the vulnerabilities this protects against.

I'm not quite sure I understand the question here: Currently in SGX, a page is either an EPC page or it is not as allocated based on the BIOS setting; it is not converted to/from the memory pool available to the OS. An EPC page can be added/removed from a specific enclave via EADD/EREMOVE (and eventually EAUG) but I imagine that it's ultimately the responsibility of the EPC manager (i.e. Windows OS or Linux SGX driver) to zero EPC pages reclaimed from an enclave.

Just as it's the case that the SDK doesn't invoke EREMOVE itself today (it relies on the virtual memory manager to do so when the enclave memory region is freed by the host process) it can't guarantee that EPC pages it used are cleaned up if the enclave terminates unexpectedly (or is killed by a malicious host).

jhand2:
> When an enclave is created, its heap pages are set to 0.

Looks like you're right.

@mikbras see the `host/sgx/create.c` function `_add_heap_pages`. It calls `_add_filled_pages` which zeros the page and then adds it.

Can you confirm this understanding? Then I agree, we can close this PR and issue.

jhand2:
I'm closing this PR. Please reopen if you feel it is still required

mikbras:
> > When an enclave is created, its heap pages are set to 0.
> 
> Looks like you're right.
> 
> @mikbras see the `host/sgx/create.c` function `_add_heap_pages`. It calls `_add_filled_pages` which zeros the page and then adds it.
> 
> Can you confirm this understanding? Then I agree, we can close this PR and issue.

Yes, the loader zero-initializes heap pages but a malicious loader may not. An alternative is for **sbrk()** to simply check whether extended regions are zero-initialized.

PR #2384, opened 2019-12-17
Modify oegdb to more relaibly find the oegdb libdir using readlink

jhand2:
On some systems (for me, fedora 31) `$BASH_SOURCE` reports a relative directory. When using oegdb from my path, that means that `dirname "${BASH_SOURCE[0]}"` is empty.

Instead use readlink to find the absolute path to the script and use that path for the lib directory.


jhand2:
> This looks good to me, but we should probably also do this for the other file in that directory named "oe-gdb"! (Or we should remove the `oe-gdb` file which has been deprecated)

Since it's just a debugging script and there is little-to-no risk of breaking someone's CI by removing it, I vote we just remove oe-gdb. I'll open a different PR to do this. Merging this one.

bors r+

PR #2385, opened 2019-12-17
Allow versions of clang-format other than 7, simply emit a warning

jhand2:
Allow `format-code` to run with versions of `clang-format` other than 7, even though the formatting rules could potentially be different in our remote builds. Emit a warning explaining that the caller should use `clang-format-7` if possible.

I am adding this change to make it possible (but not officially supported) to develop for OE on different Linux distros. For example, Fedora 31 only has packages for clang 8 and 9, not 7.

While clang 7 is the only officially supported version for OE development, 8 and 9 do work so I am attempting to lower the bar for entry for non-Ubuntu development.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -24,5 +24,7 @@ ObjCSpaceBeforeProtocolList: false
 PenaltyBreakBeforeFirstCallParameter: 1
 PenaltyReturnTypeOnItsOwnLine: 300
 PointerAlignment: Left
+SpacesInContainerLiterals: false
+SpacesInSquareBrackets: false
 TabWidth: 4
 ...
--- a/scripts/format-code
+++ b/scripts/format-code
@@ -140,8 +140,7 @@ check_version()
         # Because clang-format is not invariant across versions, this
         # is an explicit equivalence check.
         if [[ ${v1[$i]} -ne ${v2[$i]} ]]; then
-            echo "format-code requires clang-format version $1, installed version is $2"
-            exit 1
+            echo "Warning: format-code prefers clang-format version $1, installed version is $2"
         fi
     done
 }


jhand2:
bors try

andschwa:
Wouldn't using different versions of clang-format introduce formatting inconsistencies? I don't believe they're backwards (or forwards) compatible.

jhand2:
@andschwa  Not sure how that happened but I guess I edited your comment instead of adding my own?

anyway

> Wouldn't using different versions of clang-format introduce formatting inconsistencies? I don't believe they're backwards (or forwards) compatible.

clang-format follows a template file `.clang-format` so in general formatting rules should be the same. With that said, you're right that some rules may be enforce differently, primarily because we use "LLVM" as a base policy and that can change across versions.

I'll look more into this. I think confining people to always use clang-7 and related tools isn't really sustainable, we're going to need to support multiple compilers and toolsets. I'll see if there is a way to configure clang-format to behave the same across versions.

andschwa:
> in general formatting rules should be the same

Unfortunately, I don't believe this is true of clang-format. I do not believe it's backwards compatible. I may be misremembering, but `.clang-format` file or not, the tool itself behaves differently (it applies rules slightly differently, and it has different defaults for rules left unspecified) across versions.

We confine people to use the specific version of clang which we have validated and support, which is integral to maintaining the security of the project. To make this a non-issue, we have the validated Ansible setup scripts to setup a dev environment.

/cc @CodeMonkeyLeet

jhand2:
> Unfortunately, I don't believe this is true of clang-format. I do not believe it's backwards compatible. I may be misremembering, but .clang-format file or not, the tool itself behaves differently (it applies rules slightly differently, and it has different defaults for rules left unspecified) across versions.

Agreed, I believe this is true

> We confine people to use the specific version of clang which we have validated and support, which is integral to maintaining the security of the project. To make this a non-issue, we have the validated Ansible setup scripts to setup a dev environment.

I don't agree with this. I am fine with emitting a warning and I am fine with only *officially* supporting certain tool sets, but I don't think we should prevent people from using tooling that works, but doesn't run in the official OE CI. We can throw all the warnings we want, but I don't think we should fail the build (or fail a commit) just because a different tooling version is used. But that's just my opinion.

I'm not in a hurry to merge this, but I think our long term goal should be to support as many toolchains as possible (at least as many versions of clang and gcc). I want to lower the barrier to entry for pulling and building our code.

andschwa:
It's easy enough to skip the hook on the user's side (I believe when it fails it even provides instructions on how to do so), and then we run the checker with the correct version of clang-format in CI.

johnkord:
Nice discussion about clang-format! We have once experienced in the past a difference in output formatting between the default version of clang-format on Ubuntu 16.04 (clang-format-3.8) and when we upgraded to clang-format-7, but we never investigated why that was the case! Hopefully we don't hit any more issues with that.

From my own perspective, I think it's important we support reproducibility over having as much support for compilers+versions, considering how important a reproducible MRENCLAVE value is for a given enclave's source in attestation. Tagging @yakman2020 and @paulcallen who may be interested in this!

If people care about reproducibility of enclave builds they should be willing to build and install the entire toolchain (like a specific commit of LLVM/clang and its entire tool suite, etc). Making this the default way of building OE (even though it's annoying and tedious) is probably the right way forward. Unless you're just experimenting, evaluating or learning how to use OE. In that case someone should probably be using the `open-enclave` debian or NuGet package and they shouldn't have to worry about clang-format at all!

jhand2:
> Nice discussion about clang-format! We have once experienced in the past a difference in output formatting between the default version of clang-format on Ubuntu 16.04 (clang-format-3.8) and when we upgraded to clang-format-7, but we never investigated why that was the case! Hopefully we don't hit any more issues with that.

The reason they produce different output makes a lot of sense actually. From the clang mailing list:

*"If you want developers
to get the same layout, they have to use the same version of the tool. You
can assume that every single version change leads to a different layout
somewhere, or else we wouldn't be making the change. Often these changes
affect a rare corner case of the syntax, but in a reasonably large
codebase, they are still always triggered."*

> From my own perspective, I think it's important we support reproducibility over having as much support for compilers+versions, considering how important a reproducible MRENCLAVE value is for a given enclave's source in attestation. Tagging @yakman2020 and @paulcallen who may be interested in this!

clang-format doesn't really make a difference for reproducible builds. It just formats the code pre-checkin. It isn't going to change the code during a build.

> If people care about reproducibility of enclave builds they should be willing to build and install the entire toolchain (like a specific commit of LLVM/clang and its entire tool suite, etc).

If users care about reproducing the Azure build, then sure, they need to use the clang version that was used for that build. That's fine. But that isn't the only use case for this project. I don't think its reasonable to say you can only compile OE with one version of clang.

All this said, I agree with John and Andy, it is a problem that different versions of the tool produce different output. Some folks on the llvm mailing list have suggested just checking in clang-format-7 into out project's source tree. It is a small tool and we are asking all contibutors to use it for pre-checkin validation. So I think it does simplify things if people want to use some toolchain other than llvm-7 but still need use the same formatter.

jhand2:
By the way, here is the llvm mailing list thread I'm referencing: http://lists.llvm.org/pipermail/cfe-dev/2017-June/054351.html

johnkord:
> clang-format doesn't really make a difference for reproducible builds. It just formats the code pre-checkin. It isn't going to change the code during a build.

Although true, it's possible someone has installed a reproducible build toolchain that contains a different version of clang-format than the decided-upon version of clang-format, and that it ends up being invoked instead of the one we want to be invoked during commit-time. It's something to consider when setting everything up! We did have an issue before with clang-format-3.8 vs clang-format-7 both installed on the same systems.

> If users care about reproducing the Azure build, then sure, they need to use the clang version that was used for that build. That's fine. But that isn't the only use case for this project. I don't think its reasonable to say you can only compile OE with one version of clang.

Considering that attestation is fundamentally important to virtually any real-world non-malware use of SGX enclaves, and a reproducibly-built enclave image will possibly be different with even tiny differences in the compiler/linker, I think it could be a reasonable expectation that we have a list of supported dependency commit hash tuples. We could publish this list of our supported commit hash tuples of the OE project's dependencies, and then consumers (and auditors) of OE-built enclaves can use one of those commit hash tuples to verify their open source enclave images.

To me it makes a lot of sense to say "here's the set of dependencies of what we support" instead of saying we'll try to support every possible combination. Supporting everything is scary from a support standpoint!

In any case, I'm not too sure about what plan has been decided on with respect to this. @yakman2020 and @paulcallen might know! :)
 
> All this said, I agree with John and Andy, it is a problem that different versions of the tool produce different output. Some folks on the llvm mailing list have suggested just checking in clang-format-7 into your project's source tree. It is a small tool and we are asking all contibutors to use it for pre-checkin validation. So I think it does simplify things if people want to use some toolchain other than llvm-7 but still need use the same formatter.

I'm good with checking it in, as long as we have a version of it that will work on all Linux systems (or we check in multiple versions to work on our supported list of platforms). Does it have any shared library dependencies?

scripts/format-code
```diff
@@ -190,10 +189,10 @@ get_find_args()
     local defaultIncludeExts='h c cpp'
 
     if [[ -z ${userExcludeDirs} ]]; then
-        local excludeDirs=( ${defaultExcludeDirs} )
+        local excludeDirs=( "${defaultExcludeDirs}" )
```
andschwa: I’m no expert, but aren’t these breaking changes? Instead of an array of directories, the introduced quotes will make it it an array of one element, a string of space separated directory names, I think.

.clang-format
```diff
@@ -24,5 +24,7 @@ ObjCSpaceBeforeProtocolList: false
 PenaltyBreakBeforeFirstCallParameter: 1
 PenaltyReturnTypeOnItsOwnLine: 300
 PointerAlignment: Left
+SpacesInContainerLiterals: false
```
andschwa: This change doesn’t seem to have anything to do with the original PR but is instead changing our syntax style?

andschwa:
I’d be wary of checking in a binary. If we do that we need to make sure we get all the licensing correct. I also don’t think it’s necessary as we have a well documented and supported automatic installation of our dependencies.

.clang-format
jhand2: These rules are consistent with the llvm template for clang-7 but not for clang-9. So basically its overwriting some rules to make them behave like clang-7 for other versions.

jhand2:
> I’d be wary of checking in a binary. If we do that we need to make sure we get all the licensing correct.

Clang 7 license is as follows: https://github.com/llvm/llvm-project/blob/release/7.x/clang/LICENSE.TXT

Looks permissive, but I agree it is important to be careful.

> I also don’t think it’s necessary as we have a well documented and supported automatic installation of our dependencies.

Only if we plan to never support something other than Ubuntu.

scripts/format-code
jhand2: ah, you're right. I believe this was reported by the `check-linters` script. I'll revert back and take another look at why.

jhand2:
> > clang-format doesn't really make a difference for reproducible builds. It just formats the code pre-checkin. It isn't going to change the code during a build.
> 
> Although true, it's possible someone has installed a reproducible build toolchain that contains a different version of clang-format than the decided-upon version of clang-format, and that it ends up being invoked instead of the one we want to be invoked during commit-time. It's something to consider when setting everything up! We did have an issue before with clang-format-3.8 vs clang-format-7 both installed on the same systems.

clang-format should be run before checkin on dev machines and before merge  in CI. But in general it should not be run during a build (LLVM folks seem to agree with this guidance from what I've seen).

> 
> > If users care about reproducing the Azure build, then sure, they need to use the clang version that was used for that build. That's fine. But that isn't the only use case for this project. I don't think its reasonable to say you can only compile OE with one version of clang.
> 
> Considering that attestation is fundamentally important to virtually any real-world non-malware use of SGX enclaves, and a reproducibly-built enclave image will possibly be different with even tiny differences in the compiler/linker, I think it could be a reasonable expectation that we have a list of supported dependency commit hash tuples. We could publish this list of our supported commit hash tuples of the OE project's dependencies, and then consumers (and auditors) of OE-built enclaves can use one of those commit hash tuples to verify their open source enclave images.
> 
> To me it makes a lot of sense to say "here's the set of dependencies of what we support" instead of saying we'll try to support every possible combination. Supporting everything is scary from a support standpoint!
> 
> In any case, I'm not too sure about what plan has been decided on with respect to this. @yakman2020 and @paulcallen might know! :)
> 

Attestation is important, but a user of this project may not care about Azure's MRENCLAVE value, but rather some other party's value. Even if Azure doesn't support multiple MRENCLAVE values, OE should not lock users into using only the toolchain that Azure uses.

This is a little out of the scope of this PR but still an important discussion to have. I'll probably open a separate issue to track the discussion.

johnkord:
@jhand2 Sounds good yeah we got off topic! This change is good to merge imo, once other people sign off too! :)

By the way, Azure doesn't determine any MRENCLAVE value; you can think of MRENCLAVE as a hash of an executable! Two enclave images that are binary equivalent will always measure to have the same MRENCLAVE value on any SGX system

jhand2:
I feel comfortable merging this since with these changes against the OE codebase, clang-format-7, clang-format-8, and clang-format-9 all behave the same.

bors r+

PR #2388, opened 2019-12-18
Mark oe_setjmp and setjmp as returns_twice

### LINKED ISSUE #2386: oe_ocall segfault when compiling with clang 9
I know clang 9 is not a supported compiler for OE, but when it is used oe_ocall segfaults. The reason is that the stack pointer is set to a garbage value after returning from the ocall (via `oe_longjmp`) and the `pop %rbp` instruction reads from invalid memory.

This issue doesn't happen with clang 7 and gcc 8 (the other two compilers I tested) but it is mostly out of pure luck because those compilers generate much simpler assembly which does not depend as much on register state.

I believe I am close to root causing but I am opening this issue to take notes and track the issue for myself.

jhand2:
The function setjmp in libc has a behavior known as "returns twice". This means that setjmp returns from normal control flow, but can also return via another method. The `longjmp` function resets a saved register state and jumps to the instruction immediately after a call to `setjmp`, which is functionally equivalent to a return from `setjmp`.

clang 8 and 9 have a feature called Speculative Load Hardening, which is designed to mitigate some vulnerabilities in speculative execution.  One such mitigation is to check for a return address in the "red zone" of the stack (a range of the stack below %rsp) to ensure proper control flow. You can read more [here](https://llvm.org/docs/SpeculativeLoadHardening.html#indirect-call-branch-and-return-predicates). Clang cannot apply these mitigations to functions with nonstandard control flow (like returns_twice) but in OE clang does not know that setjmp and oe_setjmp have this havavior.

This PR adds annotations to these functions. Because it updates a 3rdparty library (musl) a .patch file is also included that can be applied to future versions of musl.

It also changes oe_setjmp and oe_longjmp to use straight assembly rather than inline assembly in C files. This ensures that the stack is not improperly modified by compiler generated instructions. It also removes the need to compile setjmp and longjmp with specific optimization.

Fixes #2386


enclave/core/sgx/setjmp.S
```diff
@@ -3,30 +3,30 @@
...
+    mov  %rsp, (%rdi)
+    mov  %rbp, 8(%rdi)
+    mov  %rsi, 16(%rdi) # Set rip to return address
+    mov  %rbx, 24(%rdi)
+    push %rsi # Put return address back on stack for return
+    mov  %r12, 32(%rdi)
+    mov  %r13, 40(%rdi)
+    mov  %r14, 48(%rdi)
+    mov  %r15, 56(%rdi)
+    xorl %eax, %eax # Set return value
```
jhand2: Be consistent, either use the width specifiers tor don't

enclave/core/sgx/longjmp.S
```diff
@@ -12,7 +12,7 @@
 //
 //==============================================================================
 
-/* ATTN: WINPORT */
+// Modified from linux/arch/x86/um/setjmp_64.S
```
jhand2: Specify that this is in the Linux kernel.

@mikbras @CodeMonkeyLeet @radhikaj Do we know if using linux kernel code runs us into any licensing issues?

The core is pretty simple, and not really specific to the linux kernel (it occurs in many other open source projects). Its also about 25% modified.

enclave/core/sgx/longjmp.S
```diff
@@ -22,16 +22,22 @@ oe_longjmp:
...
-    mov 8(%rdi), %rbp
-    mov 16(%rdi), %r12
-    mov 24(%rdi), %r13
-    mov 32(%rdi), %r14
-    mov 40(%rdi), %r15
-    mov 48(%rdi), %rdx
-    mov %rdx, %rsp
-    mov 56(%rdi), %rdx
-    jmp *%rdx
+    # Some versions of clang generate code that relies on the return address
```
anakrish: > Some versions of clang generate code that relies on the return address of this stack frame to be the address that was actually jumped to.  Set the stored %rip value as the return address for the stack frame directly above the one we're returning to.

Reading the code, we are not updating the `return address of this stack frame to the address that was actually jumped to`. There is no write to the memory locations addressed by %rsp or %rbp.

jhand2: ## Where are we setting the return address

`movq 16(%rdi),%rcx` retrieves the value of `%rip` stored in `env`. This is the return address that we're going to jump back to.

`movq (%rdi),%rdx` gets the stack pointer of the stack frame we're jumping back to

`movq %rcx,-0x8(%rdx)` sets the return address as the start of the stack frame directly above the stack frame we're returning to.

## As to your question about the code that necessitates this change

Here is roughly some code that clang-9 generates for `oe_ocall`. I have cut out some unnecessary stuff and tried to add comments to explain whats going on:
```
callq  135b70 <oe_setjmp> # setjmp will make longjmp return back to the following instruction
mov    -0x8(%rsp),%rcx # Get the first 8 bytes of the stack frame above this one (aka, the return address that was pushed onto the stack by the call instruction)
... {later on}
lea    -0x13(%rip),%rsi # get the address of the above mov instruction
cmp    %rsi,%rcx # check that the instruction we just retrieved is the same as the return address
mov    -0x40(%rbp),%rcx # Move 0xffffffffffffffff to rcx
cmovne %rcx,%rdx # if rcx and rdx are not equal, move -1 into %rdx
... {later on}
or     %rdx,%rsp # or rdx with rsp. If rdx was set to -1 earlier, rsp will now be a garbage value
pop    %rbp # pop causes rsp to be read from. If it is a garbage value, we will segfault.
retq
```

| Stack |
| ------ |
|     Frame 1   |
| {return address} |

jhand2: Basically we're tricking the calling function (oe_ocall) into thinking that it has arrived at the `mov    -0x8(%rsp),%rcx` instruction by returning from oe_setjmp, when in reality it has arrived there via a call to `oe_longjmp`.

enclave/core/sgx/longjmp.S
jhand2: Used musl-libc as a base instead so this is no longer relevant

jhand2:
After more investigation I found out why llvm generates the code it does and found a way to suppress the behavior using the returns_twice attribute.

Leaving this comment as a reminder to myself to do a more thorough write-up to close the loop here.

enclave/core/sgx/longjmp.S
jhand2: Found a cleaner way to fix the problem. Resolving this conversation.

jhand2:
After another look, other musl patches are applied at build time. I should do that here as well.

radhikaj:
Thanks @jhand2. Are all the right functions that return twice annotated now?

jhand2:
> Thanks @jhand2. Are all the right functions that return twice annotated now?

Yep!

jhand2:
Added copyright header

bors r+

PR #2389, opened 2019-12-19
Enable /W2 to treat W2 warnings as errors for Windows builds

jhand2:
Change windows builds to treat W1 and W2 warnings as errors, rather than just W1. Additionally, fix errors which arise from enabling this flag.

Warning for use of Flexible Array Members is ignored, although we should consider alternatives as these are a concerning in terms of memory safety.

### DIFF
--- a/include/openenclave/edger8r/common.h
+++ b/include/openenclave/edger8r/common.h
@@ -66,14 +66,24 @@ done:
     return result;
 }
 
-#define OE_ADD_SIZE(total, size)                \
-    do                                          \
-    {                                           \
-        if (oe_add_size(&total, size) != OE_OK) \
-        {                                       \
-            _result = OE_INTEGER_OVERFLOW;      \
-            goto done;                          \
-        }                                       \
+#define OE_ADD_SIZE(total, size)                                 \
+    do                                                           \
+    {                                                            \
+        if (sizeof(total) > sizeof(size_t) && total > SIZE_MAX)  \
+        {                                                        \
+            _result = OE_INVALID_PARAMETER;                      \
+            goto done;                                           \
+        }                                                        \
+        if (sizeof(size) > sizeof(size_t) && size > SIZE_MAX)    \
+        {                                                        \
+            _result = OE_INVALID_PARAMETER;                      \
+            goto done;                                           \
+        }                                                        \
+        if (oe_add_size((size_t*)&total, (size_t)size) != OE_OK) \
+        {                                                        \
+            _result = OE_INTEGER_OVERFLOW;                       \
+            goto done;                                           \
+        }                                                        \
     } while (0)
 
 /**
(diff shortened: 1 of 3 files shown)


tools/oeedger8r/src/Sources.ml
```diff
@@ -81,7 +81,9 @@ let _get_param_size (ptype, decl, argstruct) =
     and count = attr_value_to_string argstruct p.ps_count in
     match (size, count) with
     | Some s, None -> s
-    | None, Some c -> sprintf "(%s * sizeof(%s))" c type_expr
+    (* TODO: Check that c actually fits in size_t. Also check for overflow,
+     * similar to oe_add_size *)
```
jhand2: Implement this in this PR rather than waiting until later.

jhand2: Actually this is a bigger change than I want to do here. Opened an issue here: #2390

include/openenclave/edger8r/common.h
```diff
@@ -66,14 +66,24 @@ OE_INLINE oe_result_t oe_add_size(size_t* total, size_t size)
...
-        }                                       \
+#define OE_ADD_SIZE(total, size)                                 \
+    do                                                           \
+    {                                                            \
+        if (sizeof(total) > sizeof(size_t) && total > SIZE_MAX)  \
+        {                                                        \
+            _result = OE_INVALID_PARAMETER;                      \
+            goto done;                                           \
+        }                                                        \
+        if (sizeof(size) > sizeof(size_t) && size > SIZE_MAX)    \
```
johnkord: It seems very weird to me to compare both "size" and "sizeof(size)" here, so I've been staring at this for like a good 20 minutes thinking about what this means (it's the holidays... my brain still isn't working quite right yet).

But I think I know what we're doing here. Basically something like: "If someone has tried to invoke this macro with parameters that are larger SIZE_MAX, then error out".

I started thinking like "is that second condition even possible? how would someone trigger that second comparison to execute, considering the comparison `size > SIZE_MAX` needs to be evaluated on numeric values, are there any numeric types for `size` that could lead to this?" and yes, there's the notorious `long double`, our best friend.

This change looks reasonable to me!

tools/oeedger8r/src/Sources.ml
```diff
@@ -81,7 +81,9 @@ let _get_param_size (ptype, decl, argstruct) =
...
+    | None, Some c -> sprintf "((size_t)%s * sizeof(%s))" c type_expr
```
johnkord: Cool! I love when people contribute to the OCaml code :) Can you create an issue to track the newly added TODO above?

jhand2: Yep, I have this one to track https://github.com/openenclave/openenclave/issues/2390

include/openenclave/edger8r/common.h
jhand2: If sizeof(size) is larger than size_t but its value is still less than `SIZE_MAX` then we can still allow the cast.

On 64-bit machines size_t will be 64 bits and I don't know of any numeric types which are larger than that. And since SGX is only available on 64-bit processors (as far as I know) the case will probably never be hit.

But if we somehow end up running on a 32-bit CPU then and 64 bit type like `long long` would cause the second condition to be evaluated.

Issue #2394, opened 2019-12-23
Fail to create enclave with size more than 31 MB in windows


mingweishih:
A known issue is that windows does not support paging, therefore the size of enclave cannot exceed the hard limit of EPC size (approximately 90 MB). However, my test in ACC VM fails to create an enclave with size around only 31 MB.

This issue is also the cause to the failure of tests/memory on windows.

Steps to reproduce the issue:
- in helloworld sample, open enclave/helloworld.conf
- change the NumHeapPages to 7100 (alone with the default 1024 StackPages, setting the size of 31 MB in total)
- do `ninja run`

Results:
host.exe C:/workspace/samples/helloworld/build2/enclave/enclave.signed"
oe_create_helloworld_enclave(): result=21 (OE_PLATFORM_ERROR)
2019-12-23T18:05:32.000000Z [(H)ERROR] tid(0xd04) | enclave_create with ENCLAVE_TYPE_SGX1 type failed (err=0xb) (oe_result_t=OE_PLATFORM_ERROR) [..\..\host\sgx\sgxload.c:oe_sgx_create_enclave:463]

jhand2:
@mingweishih what are the specs of your VM (# of cores, total RAM).

I remember talking to Simon that some lower spec’d VMs only support 16MB of EPC per enclave, but I don’t remember the exact reason.

mingweishih:
Should be the default configuration --- 8 g ram and 2 vcpu. I'm just wondering why, as 31 MB is a bit far from 90 MB.

jhand2:
Where does the 90MB come from? Is that in the intel specification? I thought the memory limit was dependent on the hardware.

mingweishih:
Yes it depends on the bios configuration. However, in most case I would assume 128 MB (the maximal size in existing sgx machines, and only ~90 MB of them are available for enclaves) since all the VMs on the same hardware will share those EPC memory and there is currently no hardware support to reserve EPC memories per core / VM.

achamayou:
On ACC DC2s VMs the available EPC is about 28Mb: https://github.com/openenclave/openenclave/issues/542#issuecomment-413987655

mingweishih:
> On ACC DC2s VMs the available EPC is about 28Mb: #542 (comment)

This is exactly what I'm looking for. Thanks, @achamayou!

mingweishih:
Closing.

PR #2399, opened 2019-12-26
Enable create-rapid for sgx on windows

mingweishih:
This PR fixes minor issues of the create-rapid for sgx on windows and re-enables the test. The issue with the test is similar to #2394. More specifically, current window environment on ACC allows enclaves to take upto ~31 MB. As a workaround, this PR reduces the memory size per enclave and the number of simultaneous enclaves.

Note: According to my experiment, an enclave with small heap and stack size (4 pages each) takes around 1 MB of EPC. Therefore, the test fails with this configuration when the number of simultaneous enclaves is 32.


BRMcLaren:
bors try

mingweishih:
bors try

mingweishih:
bors try

Issue #1034, opened 2018-11-07
Support libunwind and libc tests for ELF enclaves on Windows


CodeMonkeyLeet:
The following tests relating to libc/libunwind are currently disabled on Windows because their functionality was not appropriately supported before. 
- backtrace
- libunwind
- stdc 

The causes for their failures on a Windows host need to be investigated; in particular, issues with libunwind/backtrace suggest issues that need to be fixed for the story around parity with Linux behaviors (#376)

Also need to investigate enabling the libcxx related tests on Windows:
- libcxx
- libcxxrt

anitagov:
Enabled both backtrace and stdc are enabled on Windows.

johnkord:
@anitagov Do you know what the status is for libunwind support in this issue?

anitagov:
@johnkord - No, I do not know the status of libunwind on Windows.

anitagov:
@johnkord - No change to the status of libunwind on Windows.

mingweishih:
#2400 for the follow-up on libunwind test support

mingweishih:
Update: libunwind is enabled with #2400. 
The PRs for enabling libc (#2403) and libcxxrt currently depend on #2434 and #2471, respectively.

PR #2400, opened 2019-12-26
Enable test\libunwind on Windows

mingweishih:
This PR enables the libunwind on Windows, which is the follow-up of #1034.


tests/libunwind/enc/enc.c
```diff
@@ -53,6 +53,8 @@ int t_printf(const char* s, ...)
 
 int t_setrlim(int r, int64_t lim)
 {
+    (void)r;
+    (void)lim;
```
jhand2: Use the `OE_UNUSED` macro for consistency. Is does essentially the same thing.

tests/libunwind/enc/CMakeLists.txt
```diff
@@ -52,6 +52,30 @@ function(add_libunwind_test_enc NAME TESTFILE)
     target_link_libraries(libunwindtest-${NAME}_enc libunwindtest-support)
     target_link_libraries(libunwindtest-${NAME}_enc -Wl,--undefined=pthread_mutex_lock)
     target_link_libraries(libunwindtest-${NAME}_enc -Wl,--undefined=_U_dyn_info_list_addr)
+
+    maybe_build_using_clangw(libunwindtest-${NAME}_enc)
+
+    # maybe_build_using_clangw populates variables in its parent scope (ie current scope)
```
jhand2: Is this done in other tests? I can look more into this when I have more time, it just seems a bit cumbersome.

johnkord: Hmm, I haven't seen this pattern before, but I think it's because `maybe_build_using_clangw` is being invoked in a function which is called in a loop in the normal part of this CMakeLists.txt file. And according to the cmake documentation, functions create new scopes (which makes sense). Invoking `maybe_build_using_clangw` here will populate those `CMAKE_*` variables in this function's scope, but it won't set those variables in the foreach loop that is calling this function.

If the `add_libunwind_test_enc` cmake function was removed and then all this code was embedded directly in the loop itself, the propagation code below probably wouldn't be necessary. That's just my guess though!

Honestly the `maybe_build_using_clangw` function is really awkward but I don't know of a better way to do what it does (uses a second compile/link toolchain to build separate executables within the same cmake build). Maybe someone can think of a better solution :)

tests/libunwind/host/host.cpp
```diff
@@ -6,11 +6,19 @@
...
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+
+#if defined(__linux__)
+#include <unistd.h>
+#elif defined(_WIN32)
+#include <process.h>
+#else
+#error "Unsupported OS platform"
```
johnkord: Should we put this somewhere else more global, maybe in some common header file? I don't think we support any platform but Linux & Windows right now.

mingweishih: The included header is specifically required by this .cpp file, so I think this should be fine. However, I agree on we don't need the if-elif-else check for now.

tests/libunwind/enc/CMakeLists.txt
jhand2: oh, I get it now. I didn't realize that this was being called inside a function. Thanks!

mingweishih:
> try
> Build failed
> 
> continuous-integration/jenkins/branch

@jhand2 @BRMcLaren could you run bors try for me once more? Thanks!

mingweishih:
bors try

jhand2:
bors try

mingweishih:
bors try

mingweishih:
@radhikaj can you merge this one? Thanks!

mingweishih:
bors try

PR #2403, opened 2020-01-04
Enable tests/libcxx on Windows

mingweishih:
This PR enables libcxx test on windows (as part of #1034). The modifications include lowering the EPC size of the enclave (see issue #2394) and using c++ thread APIs for the platform-independent, multi-threading support.

Note that three tests are moved to the broken list. The reason is that these tests are time-sensitive (i.e., depend on hard-coded wait time). Since the execution time varies from Linux to Windows, the hard-coded time may no longer be suitable after enabling the tests on Windows.


tests/libcxx/host/host.cpp
```diff
@@ -5,8 +5,15 @@
 #include <openenclave/internal/calls.h>
 #include <openenclave/internal/error.h>
 #include <openenclave/internal/tests.h>
+#include <inttypes.h>
+#if defined(_MSC_VER)
+#include <windows.h>
+typedef DWORD pthread_t;
+#else
 #include <pthread.h>
```
jhand2: can we just use C++ `std::thread` instead of pthread? It is already cross-platform so we don't need to have ifdefs

tests/libcxx/tests.broken
```diff
@@ -11,3 +11,7 @@
 ../../3rdparty/libcxx/libcxx/test/std/thread/futures/futures.async/async_race.38682.pass.cpp
 # The following test fails under clang-7 debug due to a memory leak.
 ../../3rdparty/libcxx/libcxx/test/libcxx/diagnostics/nodiscard_extensions.pass.cpp
+# The following three tests fails on ACC Windows VM.
```
jhand2: Do we know why these fail?

tests/libcxx/enc/CMakeLists.txt
```diff
@@ -68,6 +68,32 @@ function(add_libcxx_test_enc NAME CXXFILE)
 
     target_compile_definitions(libcxxtest-${NAME}_enc PRIVATE -DWITH_MAIN -D__TEST__="${CXXFILE}")
     target_link_libraries(libcxxtest-${NAME}_enc libcxxtest-support)
+
+    if (WIN32)
+        maybe_build_using_clangw(libcxxtest-${NAME}_enc)
+
+        # maybe_build_using_clangw populates variables in its parent scope (ie current scope)
+        # Propagate these variables back up to the caller.
```
jhand2: I still don't quite understand this. Can you add more to this comment to explain why the parent scope needs to be populated with these variables?

jhand2: Re-read John's comment on your other PR and I understand now. Thanks!

mingweishih: This is a workaround for tests (e.g., mbed, libunwind) that include source files from 3rdparty. To compile those files on Windows, we need to use clangw instead of cl. Currently we have maybe_build_clangw(target_file) script that updates all the compilation options for clangw. Since these tests tend to include multiple unit tests from the 3rdparty, the approach is to use a cmake function that iterates over all the tests. However, maybe_build_clangw can only set variables up to the scope of this function, so we need to populate those variables once more.
(See John's comment as well: https://github.com/openenclave/openenclave/pull/2400#discussion_r362641650)

mingweishih:
@jhand2 I've modified the test to use c++ thread. Please check. Thanks!

tests/libcxx/tests.broken
mingweishih: I briefly mentioned in the descrption of the PR. They fail because of execution time differs between Linux and Windows environments. The tests currently use fixed time duration to determine pass or fail. I think we can create another issue for the problem.

jhand2: Sounds good. Can you create an issue (if there isn't one already) and link it in this thread in case someone comes across this later?

mingweishih:
@dthaler I suggest adding the Linux label as we replace the use of pthread with c++ thread, which also affect the tests on Linux build.

tests/libcxx/tests.broken
mingweishih: Documented in #2409

jhand2:
> @jhand2 I've modified the test to use c++ thread. Please check. Thanks!

Thanks! I started looking into the changes but I'll need to take a little more time to go through it. I'll have it reviewed by tomorrow

tests/libcxx/host/host.cpp
```diff
@@ -1,22 +1,31 @@
...
 
 // Host maintains a map of enclave key to host thread ID
-static std::map<uint64_t, pthread_t> _enclave_host_id_map;
+static std::map<uint64_t, size_t> _enclave_host_id_map;
+// Since C++ thread does not allow for using the built-in
+// thread id in thread operations such as thread referencing
+// and joining, we use customized host_thread_id to keep track
+// of created thread objects
+static std::vector<std::thread> _host_thread_list;
+static std::map<std::thread::id, size_t> _thread_host_id_map;
```
jhand2: This is all pretty convoluted. Why can't you just have `std::map<uint64_t, std::thread> _enclave_host_id_map;`? You don't really need the thread's id for any operations you're doing, just a reference to the object.

jhand2: This is all pretty convoluted. Why can't you just have `std::map<uint64_t, std::thread> _enclave_host_id_map;`? You don't really need the thread's id for any operations you're doing, just a reference to the object.

mingweishih: The host thread routine (`host_enclave_thread`) need to retrieve the thread object of the thread itself. This can be done by both pthread (usiung `pthread_self())` and windows thread (get  `HANDLE` via thread id). However, it seems c++ thread does not allow that.

Using `map` may not be sufficient as we cannot obtain the `thread::id` before the thread creation.

mingweishih: @jhand2 check again?

jhand2: Thanks for cleaning this up.

Can you squash the PR into one commit?

mingweishih: done

jhand2:
bors try

mingweishih:
bors try

BRMcLaren:
bors try

mingweishih:
bors try

mingweishih:
> Please squash to one commit (in general, it would be good to do that before pushing).
> Feel free to wait to do this until bors finishes.

Will do that. I'm just trying to figure out why bors fail.

mingweishih:
> ## try
> # Build failed
> * [continuous-integration/jenkins/branch](https://oe-jenkins.eastus.cloudapp.azure.com/job/Bors/job/trying/1117/display/redirect)

Depends on #2434

mingweishih:
bors try

mingweishih:
Since #2434 is resolved, this one should be good to go.

Issue #2325, opened 2019-11-18
Stop using oeedl_file.cmake


andschwa:
It's an ancient script that we're trying to deprecate, but it's a lot of rote work to remove. It's not a great user experience because (a) it's not what we instruct consumers of the SDK to use and (b) it does a bunch of "magic" that is best done explicitly in each CMake file.

See the [CMake docs](https://github.com/openenclave/openenclave/blob/master/cmake/sdk_cmake_targets_readme.md#running-oeedger8r) on what to use instead (but use `edger8r` instead of `openenclave::edger8r` as that's the name used internally).

PR #2404, opened 2020-01-04
Add exception continuation handler

jhand2:
This takes the exception continuation feature added by @mikbras in the lsds fork of OE (https://github.com/lsds/openenclave) which allows exceptions thrown inside the enclave to be handled and potentially allow execution to continue. The enclave writer simple writes a hook function that will be called in the event of an exception.

The goal of this PR is to review the design and discuss alternatives.


jhand2:
In general I would prefer this behave more like the `signal` and `sigaction` Linux syscalls: http://man7.org/linux/man-pages/man2/sigaction.2.html

Given that we try to replicate a Linux syscall infrastructure inside the enclave, I think that interface would feel more ergonomic.

jhand2:
The cpuid test is a bit more involved than is actually needed to prove that the exception handler is working. Need to come up with a more minimal test.

tests/cpuid/enc/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+
+oeedl_file(../cpuid.edl enclave gen)
```
mingweishih: Note that `oeedl_file` is deprecated (see #2325).

tests/cpuid/host/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+
+oeedl_file(../cpuid.edl host gen)
```
mingweishih: oeedl_file is deprecated.

tests/cpuid/enc/enc.c
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+OE_SET_ENCLAVE_SGX(
+    1,    /* ProductID */
+    1,    /* SecurityVersion */
+    true, /* AllowDebug */
+    1024, /* HeapPageCount */
+    1024, /* StackPageCount */
+    2);   /* TCSCount */
```
mingweishih: Does the test require TCSCount =2?

jhand2:
After more investigation I'm not convinced this PR is necessary. Closing for now.

PR #2406, opened 2020-01-06
Avoid data race with variable _capacity

jxyang:
This fix copies the shared variable `_capacity` into a thread-local area atomically, and then operates on the thread-local copy thereafter. This ensures function `oe_arena_malloc` sees a consistent capacity value throughout its execution in the face of other threads modifying `_capacity`.

### DIFF
--- a/enclave/core/arena.c
+++ b/enclave/core/arena.c
@@ -39,11 +39,14 @@ void* oe_arena_malloc(size_t size)
     // Create the anera if it hasn't been created.
     if (_arena.buffer == NULL)
     {
-        void* buffer = oe_allocate_arena(_capacity);
+        _arena.capacity = __atomic_load_n(&_capacity, __ATOMIC_SEQ_CST);
+        void* buffer = oe_allocate_arena(_arena.capacity);
         if (buffer == NULL)
+        {
+            _arena.capacity = 0;
             return NULL;
+        }
         _arena.buffer = (uint8_t*)buffer;
-        _arena.capacity = _capacity;
         _arena.used = 0;
     }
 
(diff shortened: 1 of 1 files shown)


jxyang:
bors try

enclave/core/arena.c
```diff
@@ -39,11 +39,14 @@ void* oe_arena_malloc(size_t size)
     // Create the anera if it hasn't been created.
     if (_arena.buffer == NULL)
     {
-        void* buffer = oe_allocate_arena(_capacity);
+        _arena.capacity = __atomic_load_n(&_capacity, __ATOMIC_SEQ_CST);
```
jhand2: What is stopping simultaneous reads/writes to `_arena`?

jhand2: What is stopping simultaneous reads/writes to _arena?

jxyang: See my comment above.

jhand2: I suppose that's the problem, it can be modified by `oe_configure_arena_capacity` or `oe_arena_malloc` from multiple threads. But then that raises the question, does the assignment to _capacity in `oe_configure_arena_capacity` need to use `__atomic_store_n`?

mingweishih: According to the implementation, `oe_arena_malloc` only reads `_capacity`. So I assume the race occurs between `oe_arena_malloc` (read) and `oe_configure_arena_capacity` (write). This also requires `arena.buffer` is not initialized yet.

My follow-up on this is that the design of `oe_configure_arena_capacity` for one-time-use only? Once `arena.buffer` is set, `oe_configure_arena_capacity` seems to become ineffective.

jxyang:
bors try

PR #2407, opened 2020-01-06
Scan build warnings 2

ryanhsu19:
Third-party files have been removed from the list. See #2379 for details.

### DIFF
--- a/common/sgx/verifier.c
+++ b/common/sgx/verifier.c
@@ -120,6 +120,7 @@ static oe_result_t _add_claim(
     if (claim->value == NULL)
     {
         oe_free(claim->name);
+        claim->name = NULL;
         return OE_OUT_OF_MEMORY;
     }
     memcpy(claim->value, value, value_size);
--- a/enclave/core/sgx/report.c
+++ b/enclave/core/sgx/report.c
@@ -210,6 +210,9 @@ oe_result_t oe_get_remote_report(
 
     sgx_quote = (sgx_quote_t*)report_buffer;
 
+    if (sgx_quote == NULL)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
     // Ensure that report is within acceptable size.
     if (*report_buffer_size > OE_MAX_REPORT_SIZE)
         OE_RAISE(OE_UNEXPECTED);
--- a/enclave/crypto/cert.c
+++ b/enclave/crypto/cert.c
@@ -95,6 +95,9 @@ OE_INLINE void _referent_add_ref(Referent* referent)
 /* Decrease the reference count and release if count becomes zero */
 OE_INLINE void _referent_free(Referent* referent)
 {
+    if (!referent)
+        return;
+
     /* If this was the last reference, release the object */
     if (oe_atomic_decrement(&referent->refs) == 0)
     {
@@ -764,9 +767,7 @@ oe_result_t oe_cert_chain_read_pem(
     result = OE_OK;
 
 done:
-
     _referent_free(referent);
-
     return result;
 }
 
--- a/enclave/sgx/revocationinfo.c
+++ b/enclave/sgx/revocationinfo.c
@@ -82,7 +82,9 @@ oe_result_t oe_get_revocation_info(oe_get_revocation_info_args_t* args)
         if (in.tcb_info_size < out.tcb_info_size)
         {
             if (!(in.tcb_info = realloc(in.tcb_info, out.tcb_info_size)))
+            {
                 OE_RAISE(OE_OUT_OF_MEMORY);
+            }
 
             in.tcb_info_size = out.tcb_info_size;
         }
@@ -105,7 +107,9 @@ oe_result_t oe_get_revocation_info(oe_get_revocation_info_args_t* args)
             if (in.crl_size[i] < out.crl_size[i])
             {
                 if (!(in.crl[i] = realloc(in.crl[i], out.crl_size[i])))
+                {
                     OE_RAISE(OE_OUT_OF_MEMORY);
+                }
 
                 in.crl_size[i] = out.crl_size[i];
             }
@@ -136,17 +140,11 @@ oe_result_t oe_get_revocation_info(oe_get_revocation_info_args_t* args)
 
 done:
 
-    /* Free out buffer. */
+    /* Free buffers. */
     if (result != OE_OK)
     {
-        free(out.tcb_info);
-        free(out.tcb_issuer_chain);
-
-        for (size_t i = 0; i < OE_COUNTOF(out.crl); i++)
-            free(out.crl[i]);
-
-        for (size_t i = 0; i < OE_COUNTOF(out.crl_issuer_chain); i++)
-            free(out.crl_issuer_chain[i]);
+        oe_free_get_revocation_info_args(&in);
+        oe_free_get_revocation_info_args(&out);
     }
 
     return result;
--- a/host/sgx/elf.c
+++ b/host/sgx/elf.c
@@ -1556,7 +1556,7 @@ int elf64_add_section(
 {
     int rc = -1;
     size_t shstrndx;
-    mem_t mem;
+    mem_t mem = MEM_NULL_INIT;
     elf64_shdr_t sh;
 
     /* Reject invalid parameters */
@@ -1710,6 +1710,7 @@ int elf64_add_section(
     rc = 0;
 
 done:
+    mem_free(&mem);
     return rc;
 }
 
--- a/tests/crypto/asn1_tests.c
+++ b/tests/crypto/asn1_tests.c
@@ -266,6 +266,9 @@ static oe_result_t _parse(oe_asn1_t* asn1)
 
                 result = _parse(&sequence);
 
+                if (result != OE_OK)
+                    goto done;
+
                 break;
             }
             case OE_ASN1_TAG_INTEGER:
--- a/tests/libc/enc/enc.c
+++ b/tests/libc/enc/enc.c
@@ -111,7 +111,7 @@ int run_test(const char* name, int (*main)(int argc, const char* argv[]))
 
     /* Allocate an environment for invoking the test. */
     {
-        if (!(environ = (char**)calloc(1, sizeof(char**))))
+        if (!(environ = (char**)calloc(1, sizeof(char*))))
             goto done;
 
         memset(environ, 0, sizeof(char**));
--- a/tests/libunwind/enc/enc.c
+++ b/tests/libunwind/enc/enc.c
@@ -65,9 +65,9 @@ int test(char test_name[201], uint32_t pid)
     int rval = 1;
     g_pid = pid;
     printf("RUNNING: %s\n", __TEST__);
-    if (!(__environ = (char**)calloc(1, sizeof(char**))))
+    if (!(__environ = (char**)calloc(1, sizeof(char*))))
     {
-        rval = 1;
+        return rval;
     }
 
     static const char* argv[] = {
--- a/tests/stdc/enc/enc.cpp
+++ b/tests/stdc/enc/enc.cpp
@@ -274,6 +274,7 @@ int test(char buf1[BUFSIZE], char buf2[BUFSIZE])
     /* Cause malloc() to fail */
     void* p = malloc(1024 * 1024 * 1024);
     OE_TEST(p == NULL);
+
     OE_TEST(_called_allocation_failure_callback);
 
     return rval;
--- a/tests/syscall/cpio/commands.c
+++ b/tests/syscall/cpio/commands.c
@@ -72,7 +72,7 @@ done:
 
     oe_strarr_release(&dirs);
 
-    if (ret != 0)
+    if (ret != 0 && paths != NULL)
     {
         oe_strarr_release(paths);
         memset(paths, 0, sizeof(oe_strarr_t));
--- a/common/sgx/endorsements.c
+++ b/common/sgx/endorsements.c
@@ -80,8 +80,6 @@ static oe_result_t oe_create_sgx_endorsements(
             "Out of memory while creating endorsements.",
             NULL);
 
-    remaining_size = size;
-
     // Record creation datetime.
     {
         oe_datetime_t datetime_now = {0};
--- a/syscall/devices/hostepoll/hostepoll.c
+++ b/syscall/devices/hostepoll/hostepoll.c
@@ -757,10 +757,10 @@ static int _epoll_dup(oe_fd_t* epoll_, oe_fd_t** new_epoll_out)
 
     oe_errno = 0;
 
-    if (new_epoll_out)
-        *new_epoll_out = NULL;
-
     /* Check parameters. */
+    if (!new_epoll_out)
+        OE_RAISE_ERRNO(OE_EINVAL);
+
     if (!epoll)
         OE_RAISE_ERRNO(OE_EINVAL);
 
--- a/tests/SampleAppCRT/enc/SampleAppCRT.cpp
+++ b/tests/SampleAppCRT/enc/SampleAppCRT.cpp
@@ -37,15 +37,21 @@ int enc_test()
 #endif
 
     void* temp_region = malloc(1);
+    void* temp_region_2 = NULL;
     if (temp_region == NULL)
     {
         return -2;
     }
 
-    temp_region = realloc(temp_region, sizeof(asciistring));
-    if (temp_region == NULL)
+    temp_region_2 = realloc(temp_region, sizeof(asciistring));
+    if (temp_region_2 == NULL)
     {
+        free(temp_region);
         return -3;
+    }
+    else
+    {
+        temp_region = temp_region_2;
     }
 
     wcstombs((char*)temp_region, wcstring, wcslen(wcstring));
@@ -64,11 +70,15 @@ int enc_test()
         return -5;
     }
 
-    temp_region = realloc(temp_region, sizeof(wcstring));
-    if (temp_region == NULL)
+    temp_region_2 = realloc(temp_region, sizeof(wcstring));
+    if (temp_region_2 == NULL)
     {
         free(temp_region);
         return -6;
+    }
+    else
+    {
+        temp_region = temp_region_2;
     }
 
     mbstowcs((wchar_t*)temp_region, asciistring, strlen(asciistring));
@@ -84,7 +94,6 @@ int enc_test()
 #endif
 
     free(temp_region);
-
     return 0;
 }
 
--- a/tests/pingpong-shared/host/host.cpp
+++ b/tests/pingpong-shared/host/host.cpp
@@ -24,9 +24,8 @@ void Pong(const char* in, char* out)
         {
             got_pong = true;
         }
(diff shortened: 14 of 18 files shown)


samples/file-encryptor/enclave/keys.cpp
```diff
@@ -290,9 +290,17 @@ int ecall_dispatcher::prepare_encryption_header(
     {
         TRACE_ENCLAVE("password_key");
         for (unsigned int i = 0; i < ENCRYPTION_KEY_SIZE_IN_BYTES; i++)
+        {
+            if (password_key[i] > UINT32_MAX)
```
jhand2: Discussed previously, but it is not possible for this condition to be true. What warning is this addressing?

samples/file-encryptor/enclave/keys.cpp
```diff
@@ -290,9 +290,17 @@ int ecall_dispatcher::prepare_encryption_header(
...
+            {
+                TRACE_ENCLAVE(
+                    "password key is too large to fit into an unsigned int", 1);
+                goto exit;
+            }
             TRACE_ENCLAVE(
                 "password_key[%d] =0x%02x", i, (unsigned int)(password_key[i]));
-        goto exit;
+            goto exit;
```
jhand2: I think this goto should be outsize the loop

tests/crypto/asn1_tests.c
```diff
@@ -264,8 +264,6 @@ static oe_result_t _parse(oe_asn1_t* asn1)
                 if (result != OE_OK)
                     goto done;
 
-                result = _parse(&sequence);
```
jhand2: I think this line is supposed to be kept, it is trying to recursively parse sequence.

The missing thing here is that the response isn't used. So the whole case should be:
```
case OE_ASN1_TAG_SEQUENCE:
{
    oe_asn1_t sequence;

    result = oe_asn1_get_sequence(asn1, &sequence);
    if (result != OE_OK)
        goto done;

    result = _parse(&sequence);
    if (result != OE_OK)
        goto done;

    break;
}
```

tests/libunwind/enc/enc.c
```diff
@@ -65,10 +65,6 @@ int test(char test_name[201], uint32_t pid)
     int rval = 1;
     g_pid = pid;
     printf("RUNNING: %s\n", __TEST__);
-    if (!(__environ = (char**)calloc(1, sizeof(char**))))
-    {
-        rval = 1;
-    }
```
jhand2: This check should be kept. Just change `rval = 1;` to `return 1;`

ryanhsu19:
Thank you! I will keep working on other bugs.

radhikaj:
bors try

enclave/core/sgx/td.c
```diff
@@ -91,19 +91,23 @@ void td_push_callsite(td_t* td, Callsite* callsite)
...
 {
     if (!td || !td->callsites)
-        oe_abort();
-
-    if (td->depth == 1)
     {
-        // The outermost ecall is about to return.
-        // Clear the thread-local storage.
-        td_clear(td);
+        oe_abort();
```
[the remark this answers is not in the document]
jhand2: scan-build doesn't know that `oe_abort` stops execution. So it thinks td is potentially null and that we just ignore the result of the check.

jhand2:
Hey @ryanhsu19 can you clean up the commit history here? We shouldn't have the merge commit or the extra "check for unsigned int added by Ryan Hsu 12/6/2019" commit. The other two commits should just be squashed into one.

Let me know if you need help with how to do this.

enclave/core/sgx/td.c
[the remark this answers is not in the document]
mingweishih: I see. This is unfortunate as I like the original pattern better :)

enclave/core/sgx/report.c
```diff
@@ -210,6 +210,9 @@ oe_result_t oe_get_remote_report(
 
     sgx_quote = (sgx_quote_t*)report_buffer;
 
+    if (sgx_quote == NULL)
+        OE_RAISE(OE_UNEXPECTED);
```
jhand2: We should raise `OE_INVALID_PARAMETER` instead of unexpected

enclave/sgx/revocationinfo.c
```diff
@@ -81,20 +82,30 @@ oe_result_t oe_get_revocation_info(oe_get_revocation_info_args_t* args)
...
         {
-            if (!(in.tcb_info = realloc(in.tcb_info, out.tcb_info_size)))
+            if (!(temp = realloc(in.tcb_info, out.tcb_info_size)))
+            {
                 OE_RAISE(OE_OUT_OF_MEMORY);
+            }
+            else
+            {
+                in.tcb_info = temp;
+            }
```
jhand2: What is the reason for doing this? There is no memory safety issue in the original implementation that i can see.

host/sgx/elf.c
```diff
@@ -1674,7 +1677,10 @@ int elf64_add_section(
         /* Update number of sections */
         elf64_ehdr_t* ehdr = _get_header(elf);
         if (oe_safe_add_u16(ehdr->e_shnum, 1, &ehdr->e_shnum) != OE_OK)
+        {
+            mem_free(&mem);
```
jhand2: This should just be done once after the `done` label

tests/libc/enc/enc.c
```diff
@@ -111,7 +111,7 @@ int run_test(const char* name, int (*main)(int argc, const char* argv[]))
 
     /* Allocate an environment for invoking the test. */
     {
-        if (!(environ = (char**)calloc(1, sizeof(char**))))
+        if (!(environ = (char**)calloc(1, sizeof(char*))))
```
jhand2: why change this? They're functionally equivalent and its general practice to use the sizeof the type you're allocating

tests/libunwind/enc/enc.c
```diff
@@ -65,9 +65,9 @@ int test(char test_name[201], uint32_t pid)
     int rval = 1;
     g_pid = pid;
     printf("RUNNING: %s\n", __TEST__);
-    if (!(__environ = (char**)calloc(1, sizeof(char**))))
+    if (!(__environ = (char**)calloc(1, sizeof(char*))))
```
jhand2: same as above

tests/memory/enc/basic.c
```diff
@@ -38,6 +38,8 @@ void test_malloc(void)
     /* Ensure that malloc fails. */
     ptr = (int*)malloc(~((size_t)0));
     OE_TEST(ptr == NULL);
+    if (ptr)
+        free(ptr);
```
jhand2: This code is unreachable. Do the check/free before OE_TEST

tests/memory/enc/basic.c
```diff
@@ -102,11 +104,8 @@ void test_realloc(void)
 #pragma GCC diagnostic pop
 #endif
     OE_TEST(ptr2 == NULL);
-
-    /* realloc(X, 0) is implementation defined, but we can always free it. */
-    free(ptr);
-    ptr = realloc(NULL, 0);
-    free(ptr);
+    if (ptr2)
+        free(ptr2);
```
jhand2: Same as above

tests/stdc/enc/enc.cpp
```diff
@@ -274,6 +274,10 @@ int test(char buf1[BUFSIZE], char buf2[BUFSIZE])
     /* Cause malloc() to fail */
     void* p = malloc(1024 * 1024 * 1024);
     OE_TEST(p == NULL);
+
+    if (p)
+        free(p);
```
jhand2: unreachable

jhand2:
It would appear you accidentally pulled in some extra commits that pull a bunch of stuff into this PR

ryanhsu19:
Hi Jordan,

I will remove the last three ones for now. I changed it because it showed a scan-build error regarding the OE_TEST conditions, but again, OE_TEST runs oe_abort() once the condition is false.

As for the calloc, there is a type-mismatching error. What Alvin and I were thinking was, since char** is an array of char*, we discussed and changed sizeof(char**) to sizeof(char*), and the scan-build error has disappeared since then.

The revocation info part is for the memory leak issues.

I will change them back.

Ryan

common/sgx/verifier.c
```diff
@@ -39,8 +39,10 @@ static oe_result_t _on_unregister(oe_attestation_role_t* context)
 
 static void _free_claim(oe_claim_t* claim)
 {
-    oe_free(claim->name);
-    oe_free(claim->value);
+    if (claim->name != NULL)
```
dthaler: The standard free() is specified to be a no-op if NULL is passed, and so should oe_free.
Hence I think this change should not be needed and it should be annotated however free() is annotated so that no warnings are given when NULL is passed.

enclave/crypto/cert.c
```diff
@@ -764,8 +764,8 @@ oe_result_t oe_cert_chain_read_pem(
     result = OE_OK;
 
 done:
-
-    _referent_free(referent);
+    if (referent)
+        _referent_free(referent);
```
dthaler: Same comment here about _referent_free(), it should a no-op if NULL is passed so this change should not be needed.  Fix the header so no warning is generated.

enclave/sgx/revocationinfo.c
```diff
@@ -160,10 +158,12 @@ void oe_free_get_revocation_info_args(oe_get_revocation_info_args_t* args)
         free(args->tcb_issuer_chain);
 
         for (size_t i = 0; i < OE_COUNTOF(args->crl); i++)
-            free(args->crl[i]);
+            if (args->crl[i])
+                free(args->crl[i]);
```
dthaler: This was already safe per specification of standard free().  Do we know why it generated a warning?  Rather than changing all code that is already safe, it's better to have a fix that stops generating bad warnings.

common/sgx/verifier.c
jhand2: Indeed both `oe_free` and `free` check for null ptr.

@ryanhsu19 can you share the warning generated for these?

ryanhsu19:
As for enclave/sgx/revocationinfo.c, I made some changes due to previously many double-free warnings (there is only one such warning now and it is for a third-party file), and I included the change this file as I added the in-buffers into the /* free buffers */ section in lines 140-149. Same for common/sgx/verifier.c.

As for the enclave/crypto/cert.c, if you enter from mbedtls_x509_crt_free(referent->crt) and referent is a null pointer, then access to field crt in _referent_free() results in a dereference of a null pointer, therefore the warning. Since this is now the only one I can reconstruct, I will drop verifier.c and keep the in-buffers (to prevent possible memory leak) in revocationinfo.c.

jhand2:
@ryanhsu19 Can you reply to people's comments directly instead of making a new comment to address issues. Its too cumbersome to jump around the PR to find what you're referencing.

common/sgx/verifier.c
ryanhsu19: I included it because previously scan-build popped up some double-free warnings for many files, but I could not reconstruct it so I will remove it.

enclave/crypto/cert.c
ryanhsu19: If you enter from mbedtls_x509_crt_free(referent->crt) and referent is a null pointer, then access to field crt in _referent_free() results in a dereference of a null pointer, therefore the warning.

enclave/sgx/revocationinfo.c
ryanhsu19: I included it because scan-build popped up many double-free warnings. The main purpose for the file edit is to add in-buffers into the free buffer section in lines 140-149 to prevent potential memory leaks.

enclave/crypto/cert.c
jhand2: _referent_free does not check for a null pointer. We should fix the implementation to check there.

enclave/sgx/revocationinfo.c
jhand2: So if there are double free warnings, does this really solve them? `free` does not set the pointer to null, so even with this check the same double-free bug would persist.

common/sgx/verifier.c
jhand2: Fixing double free would require setting `claim->name` and `claim->value` to NULL after free. Not checking for NULL.

anakrish: Couple of questions:

1. Why does scanbuild not care that the `claim` variable is not null-checked before dereferencing (`claim-<name`)?

2. Why does scanbuild care that the parameter to oe_free be null checked. It is valid to pass NULL value to functions.

jhand2: 1. Scan build only catches bugs that can actually happen in the execution path. So if no code ever passes null to `_free_claim` then scan-build won't warn about it.
2. It doesn't. I don't think this fixes the reported warning.

enclave/crypto/cert.c
ryanhsu19: Okay. Working on it now.

dthaler: +1 to fixing _referent_free to check for a null pointer, for consistent with the rest of the *free() functions

samples/file-encryptor/enclave/keys.cpp
```diff
@@ -290,9 +290,17 @@ int ecall_dispatcher::prepare_encryption_header(
...
+        }
```
jhand2: Remove

tests/libc/enc/enc.c
jhand2: This is not resolved

tests/libunwind/enc/enc.c
jhand2: Also not resolved

tests/memory/enc/basic.c
```diff
@@ -102,11 +102,6 @@ void test_realloc(void)
 #pragma GCC diagnostic pop
 #endif
     OE_TEST(ptr2 == NULL);
-
-    /* realloc(X, 0) is implementation defined, but we can always free it. */
-    free(ptr);
-    ptr = realloc(NULL, 0);
-    free(ptr);
```
jhand2: Why was this removed?

tests/syscall/cpio/commands.c
```diff
@@ -72,7 +72,7 @@ int oe_lsr(const char* root, oe_strarr_t* paths)
 
     oe_strarr_release(&dirs);
 
-    if (ret != 0)
+    if (ret != 0 && paths != 0)
```
jhand2: should be `paths != NULL`

enclave/crypto/cert.c
```diff
@@ -96,15 +96,18 @@ OE_INLINE void _referent_add_ref(Referent* referent)
 OE_INLINE void _referent_free(Referent* referent)
 {
     /* If this was the last reference, release the object */
-    if (oe_atomic_decrement(&referent->refs) == 0)
+    if (referent)
```
mingweishih: For the code readability, consider do the check and early return instead of introducing nested brackets.
For example, `if (!referent) return;`

tests/libc/enc/enc.c
ryanhsu19: I know char* and char** are both pointers, but the meanings are different, and it does generate 10 related sizeof operand mismatch. Therefore I am wondering why char** is designed to point to an array of one char** element.

samples/file-encryptor/enclave/keys.cpp
ryanhsu19: Sorry I accidentally added it. I will remove it.

tests/syscall/cpio/commands.c
ryanhsu19: Thanks. Will show up in next update

tests/memory/enc/basic.c
ryanhsu19: Reverted it. There is still a problem when attempting to release already-freed ptr at line 106 (assume ptr != NULL and ptr is released at line 100), but I will try to deal with it after other warnings are settled.

enclave/crypto/cert.c
ryanhsu19: Got it. I will change it now.

common/sgx/verifier.c
```diff
@@ -118,10 +118,7 @@ static oe_result_t _add_claim(
 
     claim->value = (uint8_t*)oe_malloc(value_size);
     if (claim->value == NULL)
-    {
-        oe_free(claim->name);
```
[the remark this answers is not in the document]
ryanhsu19: I removed it because I could not reconstruct the situation and considered it unrelated. There is no scan-build error involved.

tests/libc/enc/enc.c
jhand2: oh, sorry, you're right. Re-resolving :smile:

tests/libunwind/enc/enc.c
jhand2: My mistake. Resolving

common/sgx/verifier.c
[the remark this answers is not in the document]
jhand2: But the oe_free line is in master. So if there is no associated warning you should leave it.

samples/file-encryptor/enclave/keys.cpp
```diff
@@ -290,9 +290,11 @@ int ecall_dispatcher::prepare_encryption_header(
     {
         TRACE_ENCLAVE("password_key");
         for (unsigned int i = 0; i < ENCRYPTION_KEY_SIZE_IN_BYTES; i++)
+        {
             TRACE_ENCLAVE(
                 "password_key[%d] =0x%02x", i, (unsigned int)(password_key[i]));
-        goto exit;
+            goto exit;
```
jhand2: Revert this back to what is in master.

ryanhsu19: Done

tests/memory/enc/basic.c
ryanhsu19: I thought about it for a while and reverted to the original version.

So there are 3 scan-build warnings, one for double-free and the others for memory leaks. Since the test is intended to ensure that memory allocation fails at the end of each unit test, it is not likely to change OE_TEST(ptr == NULL) and OE_TEST(ptr2 == NULL).

The double-free warning occurs because scan-build does not recognize oe_abort(), or it thinks that abort() is different than exit(). Because if ptr is really released at line 100, it means that ptr2 is successfully reallocated, therefore failing the test and leaving the rest of the code unreachable.

common/sgx/verifier.c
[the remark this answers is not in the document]
ryanhsu19: It does show "using freed memory" error in _free_claim() after I add it back, but I remember you said that oe_free(param) checks if param is null before it does anything right?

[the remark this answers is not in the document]
jhand2: It does, but claim->name is never set to null after the free. So before returning just do `claim->name = NULL;`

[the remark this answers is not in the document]
ryanhsu19: Oh, that's right. I just added it back.

jhand2:
Thanks! Please remove merge commits and squash commit history.

PR #2408, opened 2020-01-07
Add instructions for how to exclude files/directories from scan-build

jhand2:
Update scan-build instructions to explain how warnings should be treated and to show how to exclude paths/files with newer versions of scan-build.

End goal is to run scan-build in CI/CD, ignoring the 3rdparty directory. Even though clang-7 is still the recommended compiler for OE, people can use newer versions of scan-build to make sifting through these bugs a little easier.

### DIFF
--- a/docs/GettingStartedDocs/Contributors/RunningStaticAnalysis.md
+++ b/docs/GettingStartedDocs/Contributors/RunningStaticAnalysis.md
@@ -40,3 +40,29 @@
     ```
 
     Running `scan-view /tmp/scan-build-2019-12-12-102130-18694-1` will open results in a web browser.
+
+## Interpreting warnings
+
+Ideally, all warnings would be legitimate and CI/CD could reject PRs which introduce scan-build errors.
+Unfortunately there are currently many scan-build errors in the OE codebase, and many of them are in
+external code.
+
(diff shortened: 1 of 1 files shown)


docs/GettingStartedDocs/Contributors/RunningStaticAnalysis.md
```diff
@@ -40,3 +40,29 @@
...
     Running `scan-view /tmp/scan-build-2019-12-12-102130-18694-1` will open results in a web browser.
+
+## Interpreting warnings
+
+Ideally, all warnings would be legitimate and CI/CD could reject PRs which introduce scan-build errors.
+Unfortunately there are currently many scan-build errors in the OE codebase, and many of them are in
+external code.
+
+Specifically, errors reported from the 3rdparty directory should be ignored. Any other warnings are considered
+bugs and should be fixed or marked as false-positives.
```
dthaler: Can you be more specific on how they are "marked as false-positives"?

jhand2: I decided to remove this piece. Technically an ifdef can be added to keep the code from being analyzed, but I have yet to find a case where this is necessary. It is almost always correct to actually fix the issue and I don't want to encourage band-aid fixes by putting them in the documentation.

PR #2411, opened 2020-01-07
oe_enter_sim manages GS and FS registers

anakrish:
oe_enter_sim instruction simulates EENTER instruciton.
with this change, it manages GS and FS registers.
Before entering the enclave, the registers are set to the values
set by the EENTER instruction. Immediately upon return from the enclave,
the registers are reset to their host-side values.

- Localized register management.
- Avoids duplication of register management in many code paths.
- Removed unnecessary assertions.

### DIFF
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -115,7 +115,6 @@ static oe_result_t _enter_sim(
 {
     oe_result_t result = OE_UNEXPECTED;
     sgx_tcs_t* tcs = (sgx_tcs_t*)tcs_;
-    ThreadBinding* binding = GetThreadBinding();
     td_t* td = NULL;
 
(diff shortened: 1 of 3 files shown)


anakrish:
bors try

nonpolarity:
Perfect! Now the FS/GS pages have consistency in simulation mode.

host/sgx/calls.c
```diff
@@ -127,22 +126,8 @@ static oe_result_t _enter_sim(
     if (!tcs->u.entry)
         OE_RAISE(OE_NOT_FOUND);
 
-    /* Save old GS and FS register bases */
-    binding->host_gs = oe_get_gs_register_base();
```
radhikaj: Awesome, glad to see this simplified.

PR #2416, opened 2020-01-09
Fix oe_random for size > 1024

thomasten:
mbedtls_ctr_drbg_random() does not accept arbitrary large output sizes
and must be called repeatedly to fill large buffers.

### DIFF
--- a/tests/crypto/random_tests.c
+++ b/tests/crypto/random_tests.c
@@ -13,38 +13,48 @@
 #include "tests.h"
 
 #define SEQ_COUNT 64
-#define SEQ_LENGTH 19
 
-void TestRandom(void)
+static void _test_random(size_t seq_length)
 {
-    printf("=== begin %s()\n", __FUNCTION__);
+    printf("=== begin %s(%zu)\n", __FUNCTION__, seq_length);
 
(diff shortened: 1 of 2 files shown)


tests/crypto/random_tests.c
```diff
@@ -13,38 +13,48 @@
...
-#define SEQ_LENGTH 19
 
-void TestRandom(void)
+static void _test_random(size_t seq_length)
 {
-    printf("=== begin %s()\n", __FUNCTION__);
+    printf("=== begin %s(%zu)\n", __FUNCTION__, seq_length);
 
-    uint8_t buf[SEQ_COUNT][SEQ_LENGTH];
+    uint8_t buf[SEQ_COUNT][seq_length];
```
jhand2: Ah, didn't catch this the first time. We do not allow variable length arrays. You'll need to allocate `buf` on the heap or make buf sufficiently large for all calls to _test_random.

If you go for the second option, please create a macro with the max value of seq_length.

jhand2:
Thanks!

bors r+

jhand2:
Failure appears to be a infrastructure issue. Trying again

bors r+

PR #2417, opened 2020-01-11
Remove unused corelibc pthread feature

mikbras:
The pthread feature in liboecore is no longer used.


mikbras:
bors try

jhand2:
Looks good at first look. Can you update the PR description with a few things:

1. It looks like the feature is moved to libc, not just removed entirely. Description should reflect that.
2. Why was the feature needed before in corelibc and not needed there now?

Also please squash commits.

mikbras:
Thanks Jordan. I'm closing this for now because we might want to do more.

CodeMonkeyLeet:
@mikbras Also note that the pthread.h is documented as part of the public oelibc surface already. It's removal needs to go through the appropriate deprecation process.

Issue #565, opened 2018-08-21
Open Enclave does not call __init_libc which breaks several libc features


CodeMonkeyLeet:
Per @ihaller's observation:

> Looking into the TLS error, it seems to me that the function __init_libc is never called anywhere from OpenEnclave. And the only parts of its contents that do get replicated in the OpenEnclave initialization are the calls to the constructors. If this is the case, then the following are not initialized:
> * Static TLS of musl (__init_tls): this I understand is a design choice and understood by the team.
> * Stack protector cookie (__init_ssp): this will make the stack protector mostly meaningless and should be documented as an unsupported feature as security-conscious developers would be interested in using it.
> * The global musl configuration (_libc struct): some interesting things here, including a thread count which if left at 0 will break musl malloc once it is in use (will disable locking within the allocator).
> * Some other globals: no idea what these are.
> * It looked to me like the _[libc.secure]_ flag is related to the glibc secure execution mode, which is in general triggered for an SUID binary, but can also be set for other executables. I don’t know what happens, but it seems as if the polling is checking for redirection of stdin/stdout/stderr and exiting if it detects it?

@mikbras confirms that we currently only call global initializers, and this issue tracks investigating what is needed to incorporate the rest of these:
* No call to `__init_tls` is tracked as part of #538 for what would be needed to support `thread_local`
* No call to `__init_ssp` is tracked as part of #81 for what would be needed to support stack protector cookies.
* OE may not be using the thread count in libc since we use dlmalloc & oe_thread instead of MUSL malloc & pthreads, but we need to verify this.
    * We also need to investigate the impact of failing to check/set the various libc.auxv values and libc.secure.
* We need to further investigate whether the remaining globals not being initialized has an impact; current intuition is that there would not be legitimate use cases for them inside the enclave. Likely not a problem, but it is important to understand them because they are used internal to MUSL. We should make sure that code outside of it has no visibility to them:
    * __environ
    * __hwcap
    * __sysinfo
    * __progname (seems to be aliased on program_invocation_short_name for example, which is part of errno.h)
        * One option is to skip linking in libc.c and replace it with a stub in the enclave libc, but that will be musl-specific instead of being usable across different libc implementations.

jhand2:
Fully documented and implemented necessary libc initialization here: #2420

radhikaj:
@jhand2 Can I close this?

jhand2:
[...]

PR #2420, opened 2020-01-13
Add documentation for libc initialization

jhand2:
Fully document the libc initialization done by musl and explain which parts oe does and which it doesn't.

For the minimal initialization that oe should do for libc, add a global initialization constructor which will be called on enclave load.

See the documentation rendered in markdown here: https://github.com/jhand2/openenclave/blob/oe_init_libc/docs/DesignDocs/libc_initialization.md

### DIFF
--- a/docs/DesignDocs/libc_initialization.md
+++ b/docs/DesignDocs/libc_initialization.md
@@ -1,0 +1,176 @@
+# Musl libc Initialization
+
+## Normal musl libc initialization on Linux x86\_64
+
+A normal executable linux program which uses Musl libc will have the following initialization pattern,
+where `_start()` is the entry point of every program.
+
+
+```
+_start()
+    v
+    |
+    --> _start_c()
+              v
+              |
+              --> _libc_start_main()
+                            v
+                            |
+                            --> __init_libc()
+                            --> _libc_start_main_stage2()
+                                          v
+                                          |
+                                          --> __libc_start_init()
+                                          --> main()
+                                          --> exit()
+```
+
+### What do each of these functions do?
+#### \_start
+
+`_start` is the entry point for every userspace executable. Its only
+purpose is to initialize rbp to 0 and call `_start_c`.
+
+#### \_start\_c
+
+Sets some registers and calls `_libc_start_main`
+
+#### \_libc\_start\_main
+
+Calls `__init_libc` and `_libc_start_main_stage2`
+
+#### \_\_init\_libc
+
+Finally something interesting. `__init_libc` does many things:
+
+* Calls `__init_tls()` to initialize thread local storage
+* Calls `__init_ssp()` to initialize stack protector
+* Calls poll syscall to check if stdin, stdout, and stderr to be ready for I/O. If any are not available for I/O, it initializes them to /dev/null
+* sets the following global variables:
+    * `__environ`
+    * `__hwcap`
+    * `__sysinfo`
+    * `__progname`
+* Sets the following members of the global `libc` structure
+    * `auxv`
+    * `page_size`
+    * `secure`
+
+#### \_libc\_start\_main\_stage2
+
+Calls `__libc_start_init()`. Then calls `main()`. Then calls `exit()` with the return
+code of `main()`.
+
+#### \_\_libc\_start\_init:
+
+Calls `_init()` and loops through `__init_array()` and calls each of the
+compiler-generated init functions.
+
+This includes constructors of C++ global objects and functions marked with the
+constructor attribute.
+
+## OE Initialization
+
+### OE libc Initialization
+
+OE initializes libc in similar ways, but leaves out some of the initialization
+that is not needed in the enclave. Additional musl-specific initialization is
+done by a function `_oe_init_c()` which is marked with the constructor attribute.
+This ensures that enclaves which link libc will call `_oe_init_c()` during
+`oe_call_init_functions()`.
+
+#### \_\_init\_tls
+
+OE initializes thread local storage via a call to `oe_thread_local_init()` in
+enclave/core/linux/threadlocal.c which is called during `td_init()` which
+initializes a thread data structure.
+
+#### \_\_init\_ssp
+
+`td_init()` is also responsible for initializing the stack protector.
+
+#### Check for availability of stdin, stdout, and stderr
+
+OE enclaves do console I/O by making ocalls to the host. Because of this,
+the job of checking console I/O file descriptors is handled by the host's
+libc implementation.
+
+#### Initialize global variables
+
+* `__environ`
+
+    `__environ` is a pointer to the `envp` argument that can be passed to `main()`, which
+    contains the current environment variables. Because the enclave does not inherit an
+    initial set of environment variables, this is not initialized. Environment-based libc
+    functions such as `putenv`, `getenv`, etc. will still work as expected. No musl functions
+    assume __environ is non-null.
+
+* `__hwcap`
+
+    `__hwcap` is a bitmask of flags detailing whether various features are supported
+    by the system's CPU. In general, hwcaps are used by the kernel, but musl uses
+    a few of them for threading. It checks for `CPU_HAS_CAS_L`, `CPU_HAS_LLSC`, and
+    `HWCAP_TLS`.
+
+    Each of these are used for thread local storage. Because OE implements its own
+    TLS, these flags do not need to be initialized.
+
+* `__sysinfo`
+
+    On x86_64 this is not used by musl. OE does not need to initialize.
+
+* `__progname`
+
+    `__progname` contains the name of the executing program (`argv[0]`). It is used by
+    musl libc in `warn()` and `err()` functions to print a message to stderr prefixed by the
+    program name. Musl initializes `__progname` to null, so these calls will succeed and
+    simply have an empty prefix.
+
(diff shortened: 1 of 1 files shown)


enclave/core/CMakeLists.txt
```diff
@@ -172,6 +172,8 @@ if (OE_TRUSTZONE)
     target_link_libraries(oecore PUBLIC oelibutee_includes)
 endif ()
 
+target_link_libraries(oecore PUBLIC oelibc)
```
jhand2: Remove this, was just for testing

libc/CMakeLists.txt
```diff
@@ -843,9 +845,16 @@ target_link_libraries(oelibc
...
 
+# Ensure the libc global initialization function is linked
+target_link_libraries(oelibc
+    INTERFACE
+    -Wl,-u,__oe_init_c)
+
 target_include_directories(oelibc PRIVATE
-  ${PROJECT_BINARY_DIR}/3rdparty/musl/musl/src/include
-  ${PROJECT_BINARY_DIR}/3rdparty/musl/musl/src/internal)
+  ${PROJECT_SOURCE_DIR}/enclave/core
```
jhand2: Remove this, was for testing

jhand2:
@CodeMonkeyLeet @mikbras @anakrish @radhikaj We should be able to do global initialization for any module in the same way I have done libc initialization here. This could be used for hostfs and sockets for example.

3rdparty/musl/oe_init_c.c
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License.
+
+#include "libc.h"
+#include "unistd.h"
+
+#include <openenclave/internal/defs.h>
+
+static size_t __auxv;
+
+void OE_CONSTRUCTOR __oe_init_c(void)
```
anakrish: Marking this as a _GCC constructor_ would ensure that this function is called via [oe_call_init_functions](https://github.com/openenclave/openenclave/blob/71edd2cf92f52017410d5e3262628b962067560a/enclave/core/init_fini.c#L58).

Do we need additional guarantee that ___oe_init_c_ must be the first global constructor that would be called.

docs/DesignDocs/libc_initialization.md
```diff
@@ -0,0 +1,169 @@
...
+              v
+              |
+              --> _libc_start_main()
+                            v
+                            |
+                            --> __init_libc()
+                            --> _libc_start_main_stage2()
+                                          v
+                                          |
+                                          --> __libc_start_init()
```
anakrish: Can you expand this diagram to show when other GCC constructors and C++ constructors would be called.

3rdparty/musl/oe_init_c.c
jhand2: good point. A priority can be passed to `__attribute((constructor))__` to ensure __oe_init_c is called first. Currently I think this only matters for C++ enclaves.

docs/DesignDocs/libc_initialization.md
jhand2: Sure. So that I remember for later, it is the last thing before calling `main`

anakrish:
> 
> 
> @CodeMonkeyLeet @mikbras @anakrish @radhikaj We should be able to do global initialization for any module in the same way I have done libc initialization here. This could be used for hostfs and sockets for example.

Great idea! Can you see if you can express order of initialization of libraries as well? Say encryptedfs might depend on hostfs and therefore hostfs must be initialized before encryptedfs.

docs/DesignDocs/libc_initialization.md
```diff
@@ -0,0 +1,169 @@
+# Musl libc Initialization
+
+## Normal musl libc initialization on Linux x86\_64
+
+A normal executable linux program with uses Musl libc will have the following initialization pattern,
```
dthaler: I can't parse "with uses"

docs/DesignDocs/libc_initialization.md
```diff
@@ -0,0 +1,169 @@
+# Musl libc Initialization
+
+## Normal musl libc initialization on Linux x86\_64
```
dthaler: What about on Windows?  Should there be another section?  Or can this section be made OS-agnostic?

docs/DesignDocs/libc_initialization.md
```diff
@@ -0,0 +1,169 @@
...
+
+Calls `__init_libc` and `_libc_start_main_stage2`
+
+#### \_\_init\_libc
+
+Finally something interesting. `__init_libc` does many things:
+
+* Calls `__init_tls` to initialize thread local storage
+* Calls `__init_ssp` to initialize stack protector
+* Calls poll syscall to check if stdin, stdout, and stderr to be ready for I/O. If any are not available for IO, it initializes them to /dev/null
```
dthaler: punctuation nit: "I/O" or "IO"?  Be consistent.

docs/DesignDocs/libc_initialization.md
```diff
@@ -0,0 +1,169 @@
...
+enclave/core/linux/threadlocal.c which is called during `td_init()` which
+initializes a thread data structure.
+
+#### \_\_init\_ssp
+
+OE initializes the stack protector again during the `td_init()` function.
+
+#### Check for availability of stdin, stdout, and stderr
+
+OE enclaves do console IO (and all file IO as of this writing) by making
```
dthaler: I think the "and all file IO as of this writing" is likely to get out of date without noticing that this file and line needs to be updated.   The feature.new_platforms branch does have a file IO provider that does not make ocalls to the host, I don't know what the status is of porting it to integrate with musl libc and merging it into master, but the point is we know that this claim here is short-lived and easy to miss in the future.  Hence would be better to avoid making this claim here, especially since the section isn't about file IO per se but is just about stdin/stdout/stderr.

docs/DesignDocs/libc_initialization.md
jhand2: ah, *which uses :)

docs/DesignDocs/libc_initialization.md
jhand2: This is for in-enclave libc initialization and OE uses linux paradigms inside the enclave regardless of the host OS. I can specify this somewhere.

docs/DesignDocs/libc_initialization.md
jhand2: Actually, this is done in __libc_start_init

libc/CMakeLists.txt
```diff
@@ -843,9 +845,14 @@ target_link_libraries(oelibc
   PUBLIC oesyscall oecore oelibc_includes
   PRIVATE oelibasm)
 
+# Ensure the libc global initialization function is linked
+target_link_libraries(oelibc
+    INTERFACE
+    -Wl,-u,__oe_init_c)
```
jhand2: If a caller doesn't use cmake to build their enclave, they will likely not add this flag. Need to look for an in-source way to specify this. In theory the `used` attribute does this but for obvious reasons it is not used by the linker.

docs/DesignDocs/libc_initialization.md
```diff
@@ -0,0 +1,172 @@
...
+    * The `getauxval` call used for reading from the auxilary vector by userspace program.
+
+    OE initializes auxv as an array containing 1 zero entry. This denotes an empty vector.
+
+* page\_size
+
+    Musl libc uses page_size for dynamic linking, pthreads, memory allocation, and the
+    sysconf call.
+
+    In order to support calls to sysconf, OE initializes page\_size
```
CodeMonkeyLeet: Nit: Missing period at end of sentence. Probably also useful to indicate what we initialize it to (as done for `secure` below).

jhand2:
See #2428 for my implementation of general global module initialization functions (using libc as a template since it is a very small init function). I will likely scope this PR down to just documentation.

include/openenclave/internal/defs.h
```diff
@@ -71,4 +71,17 @@
 /* OE_FIELD_SIZE */
 #define OE_FIELD_SIZE(TYPE, FIELD) (sizeof(((TYPE*)0)->FIELD))
 
+#ifdef __GNUC__
+#define OE_CONSTRUCTOR __attribute__((used, constructor))
+#else
+#error "OE_EXPORT unimplemented"
+#endif
+
+#ifdef __GNUC__
```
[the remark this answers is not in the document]
jhand2: Moved implementation to my module initialization PR

CodeMonkeyLeet:
@jhand2 One additional thing: Can you move the "Fixes: #565" key phrase to PR #2428 instead?

jhand2:
Holding this PR as some of this information is only true after #2428 is merged

anakrish:
@jhand2 
I found some documentation online:

- [GCC: How Initialization Functions Are Handled](https://gcc.gnu.org/onlinedocs/gccint/Initialization.html)
- [OSDev: Creating a C Library](https://wiki.osdev.org/Creating_a_C_Library)
- [OSDev: https://wiki.osdev.org/Calling_Global_Constructors]

They describe how the compiler uses the following linker line to correctly initialize the C library and call global constructors:
`ld crt0.o crti.o crtbegin.o <objectfiles> crtend.o crtn.o`

Can you confirm if those documents are applicable for OE SDK as well?

Issue #2412, opened 2020-01-08
Fix support for just "out" parameters in deep copy


andschwa:
I discovered while adding further tests that when using deep copy for parameters that are only "out" parameters that it does not work. It works for "in" and "in/out" (the latter of which are both tested), but for "out" parameters there is a bit of logic to add to layout the host side pointers correctly (very similar to what's done for "in/out"), I think.

I had to prioritize in my remaining time working on this project, and so I prioritized making oeedger8r maintainable with the updated build system and isolated modules instead. Sorry for not being able to get to this part @anakrish, please let me know if you have any questions.

/cc @radhikaj

mingweishih:
Fixed by #2427.

PR #2427, opened 2020-01-14
Support the deep copy of out-only parameters

mingweishih:
This PR revolves #2412, which supports out only parameters with deep copy. Previously, the oeedger8r recursively generates checks for the existence of each pointer in the annotated parameter's data structure. This works for the cases of in and in/out as the content of the structure is copied to the enclave while this is not the case for the out-only parameter.

See the following example:
```
struct { 
   size_t size; 
   int *buf; 
} foo; 
```
Enclave checks: 
```
if (parags_in->s) 
    OE_SET_OUT_POINTER(s, 1 * sizeof(foo), foo*); 
If (parags_in->s && parags_in->s->buf) 
    OE_SET_OUT_POINTER(s->buf, 1 * sizeof(int), int*); 
```
-> This check fails because the `buf` is not initialized, preventing the setup of the data structure.

The patches modified the oeedger8r to ensure that for the case of out-only parameter, any nested pointer should be NULL and its parent should not before setting up the data structure.

Example:
```
if (parags_in->s) 
    OE_SET_OUT_POINTER(s, 1 * sizeof(foo), foo*); 
If (parags_in->s && !parags_in->s->buf) 
    OE_SET_OUT_POINTER(s->buf, 1 * sizeof(int), int*); 
```

### DIFF
--- a/include/openenclave/edger8r/common.h
+++ b/include/openenclave/edger8r/common.h
@@ -99,7 +99,7 @@ done:
  * buffer. Make sure that the buffer has enough space.
  */
 #define OE_SET_OUT_POINTER(argname, argsize, argtype)                        \
-    if (pargs_in->argname)                                                   \
+    do                                                                       \
     {                                                                        \
         pargs_in->argname = (argtype)(output_buffer + output_buffer_offset); \
         OE_ADD_SIZE(output_buffer_offset, (size_t)(argsize));                \
@@ -108,7 +108,7 @@ done:
             _result = OE_BUFFER_TOO_SMALL;                                   \
             goto done;                                                       \
         }                                                                    \
-    }
+    } while (0)
 
 /**
  * Compute and set the pointer value for the given parameter within the output
--- a/tests/oeedger8r/host/testdeepcopy.cpp
+++ b/tests/oeedger8r/host/testdeepcopy.cpp
@@ -177,12 +177,48 @@ void test_deepcopy_edl_ecalls(oe_enclave_t* enclave)
 
     {
         CountStruct s{};
+        uint64_t p[3] = {7, 7, 7};
+        s.count = 5;
+        s.size = 6;
+        s.ptr = p;
+        OE_TEST(deepcopy_inout_count(enclave, &s) == OE_OK);
+        OE_TEST(s.count == 7);
+        OE_TEST(s.size == 64);
+        test_struct(s, 3);
+    }
+
+    {
+        CountStruct s{};
         uint64_t p[3] = {0, 0, 0};
         s.ptr = p;
         OE_TEST(deepcopy_out_count(enclave, &s) == OE_OK);
         OE_TEST(s.count == 7);
         OE_TEST(s.size == 64);
         test_struct(s, 3);
+    }
+
+    {
+        auto s = init_struct<CountStruct>();
+        int ints[]{0, 1, 2, 3};
+        ShallowStruct shallow{1, 8, nullptr};
+        CountStruct counts[]{s, s, s};
+        NestedStruct n{13, ints, &shallow, counts};
+        OE_TEST(deepcopy_nested_out(enclave, &n) == OE_OK);
+
+        for (size_t i = 0; i < 4; ++i)
+            OE_TEST(n.array_of_int[i] == static_cast<int>(i));
+
+        // Out-only deepcopy does not support shallow pointer passing.
+        OE_TEST(n.shallow_struct != &shallow);
+        OE_TEST(n.array_of_struct == counts);
+
+        for (size_t i = 0; i < 3; ++i)
+            for (size_t j = 0; j < 8; ++j)
+            {
+                OE_TEST(n.array_of_struct[i].count == 7);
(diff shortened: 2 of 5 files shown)


mingweishih:
bors try

tools/oeedger8r/src/Sources.ml
```diff
@@ -178,7 +178,10 @@ let rec get_ptr_setter get_deepcopy args count setter (ptype, decl) =
         (* NOTE: This makes the embedded check in the `OE_` macro superfluous. *)
         [
           sprintf "if (pargs_in->%s)"
-            (String.concat " && pargs_in->" (List.rev (arg :: args)));
+            (if args = [] then arg
+            else ((String.concat " && pargs_in->" (List.rev(args))) ^
+              ((if is_out_setter setter then " && !" else " && ") ^
```
andschwa: I think you can probably just use the logic of `is_out_stter` inline instead of introducing a "common" function that is only used once. But otherwise this looks good to me.

mingweishih: Thanks for the suggestion, @andschwa. The patch is updated.

tests/oeedger8r/host/testdeepcopy.cpp
```diff
@@ -185,6 +197,30 @@ void test_deepcopy_edl_ecalls(oe_enclave_t* enclave)
...
+        int ints[]{0, 1, 2, 3};
+        ShallowStruct shallow{1, 8, nullptr};
+        CountStruct counts[]{s, s, s};
+        NestedStruct n{13, ints, &shallow, counts};
+        OE_TEST(deepcopy_nested_out(enclave, &n) == OE_OK);
+
+        for (size_t i = 0; i < 4; ++i)
+            OE_TEST(n.array_of_int[i] == static_cast<int>(i));
+
+        // Out-only deepcopy does not support shallow pointer passing.
```
radhikaj: Can you elaborate in this?

radhikaj:
@anakrish @mikbras please review

tests/oeedger8r/host/testdeepcopy.cpp
mingweishih: Nested shallow pointer passing relies on deep copying the address of the pointer and passing the address to the enclave (via the input buffer). Therefore, the enclave gets the address points to the pointer on the host side (i.e., pass-by-address). However, for an out-only parameter, the content of the nested structure will be zero-out (using the output buffer instead) and therefore not deep copied to the enclave side.

For example,
`struct ShallowStruct {
    uint64_t count;
    size_t size;
    uint64_t* ptr;
 };`
In the case of `[in, count=1] ShallowStruct* s`, the value of `ptr` will be copied to the enclave.
For the case of `[out, count=1] ShallowStruct* s`, the value of `ptr` will be zero-out.

Hope this makes sense!

include/openenclave/edger8r/common.h
```diff
@@ -99,7 +99,7 @@ OE_INLINE oe_result_t oe_add_size(size_t* total, size_t size)
  * buffer. Make sure that the buffer has enough space.
  */
 #define OE_SET_OUT_POINTER(argname, argsize, argtype)                        \
-    if (pargs_in->argname)                                                   \
+    do                                                                       \
```
anakrish: What is the need for removing this check?

Previously, `pargs_in->argname` is set to a non-null value only if the original value was a non-null value.
With this change, `pargs_in->argname` is always set to a non-null value.

tools/oeedger8r/src/Sources.ml
```diff
@@ -178,7 +178,10 @@ let rec get_ptr_setter get_deepcopy args count setter (ptype, decl) =
         (* NOTE: This makes the embedded check in the `OE_` macro superfluous. *)
         [
           sprintf "if (pargs_in->%s)"
-            (String.concat " && pargs_in->" (List.rev (arg :: args)));
+            (if args = [] then arg
```
anakrish: Do we support arbitrarily nested structures?

include/openenclave/edger8r/common.h
mingweishih: This is because only first-level arguments are non-null. This check would block the initialization for the nested arguments. Therefore, I move the check logic to https://github.com/openenclave/openenclave/pull/2427/files#diff-e7bae2bcdd6045dc447a7f504b89b903R18.

The current behavior works as follow:
For the first-level arg, the generated check will be:
`if (pargs_in->s) OE_SET_OUT_POINTER(s, ...);`
For the nested args, the checks become:
`if (pargs_in->s && !pargs_in->s->ptr) OE_SET_OUT_POINTER(s->ptr, ...);`
`if (pargs_in->s && pargs_in->s->ptr && !pargs_in->s->ptr->tail) OE_SET_OUT_POINTER(s->ptr->tail, ...);`

tools/oeedger8r/src/Sources.ml
mingweishih: Current deepcopy logic is that given a first-level pointer, try recursively to find all the nested pointers of that pointer.

PR #2428, opened 2020-01-15
Implement global module initialization functions for OE modules

### LINKED ISSUE #2425: Add global initialization functions for modules
Modules such as hostfs and sockets require an initialization function to be called. Add a feature to OE which will call an initialization function on enclave load for each module that is linked into the enclave binary.

This can potentially be implemented using global constructors, but there is an issue with global constructors not being linked into the final binary if they are not referenced.

Ideally, these callback functions will be called prior to C++ global constructors.

This same mechanism should be used for initializing switchless ocalls: #2180

jhand2:
Initial prototype for global module initialization functions to initialize OE modules.

OE modules are static libraries that provide discreet functionality (such as libc or a host filesystem) which are compiled as static libraries and linked into the enclave binary. Modules may have any number of initialization functions, which will be called on first enclave entry prior to global constructors.

Fixes #2425
Fixes: #565

### DIFF
--- a/docs/DesignDocs/openenclave_modules.md
+++ b/docs/DesignDocs/openenclave_modules.md
@@ -1,0 +1,110 @@
+# OpenEnclave Modules
+
+The Open Enclave SDK provides a core set of functionality for build enclave
+applications. Some of these features &2014; such as memory management and ecall/ocall
+marshalling &2014; are required to create an enclave. Other features such as
+libc functions, host filesystem, and network sockets can be optionally included
+inside the enclave by the enclave developer. These optional features are known
+as **Open Enclave Modules** and are static libraries that can be optionally
+linked into the enclave binary.
+
+## Currently supported modules
+
+Currently, Open Enclave only supports modules that are part of the OE SDK source tree.
+Today there are 5 optionally includable modules:
+
+* hostfs
+* libc
+* hostepoll
+* hostsock
+* hostresolver
+
+## Module initialization
+
+OE modules can each have an initialization function which will be called on
+first entry to the enclave. This ensures that a module are fully initialized
+prior to an application using it. There are two ways to demark initialization
+functions:
+
+1. `OE_MODULE_INIT Attribute`
+
+The `OE_MODULE_INIT` attribute can be applied to a function. This ensures that the
+function is called on enclave entry.
+
+```
+OE_MODULE_INIT
+void init_my_module(void)
+{
+    // Do stuff
+}
+```
+
+2. `OE_MODULE_INIT_PRIORITY` Attribute
+
+Optionally, modules can specify a priority for their initialization functions.
+Functions with a lower priority value will be called before functions with a
+higher priority value. This can be used if one module depends on another being
+initialized first.
+
+```
+// mod1.c
+OE_MODULE_INIT_PRIORITY(0)
+void mod1_init(void)
+{
+    // Initialize module 1
+}
+
+// mod2.c
+OE_MODULE_INIT_PRIORITY(1)
+void mod2_init(void)
+{
+    // Do stuff that depends on mod1. This function will
+    // always be called after mod1_init.
+}
+```
+
+### Linker disclaimer
+
+Because modules can be added and removed at compile-time without the core OE
+library knowing about them, initialization functions are not actually
+referenced by symbol. They are instead kept in their own section in the binary.
+To ensure a module's initialization function is included in the final enclave
+binary there are 3 options:
+
+1. Add the module initialization function in the same file as a symbol that is
+referenced in the enclave binary.
+
+```
+// mod_init.c
+
+int always_used_symbol = 1;
+
+OE_MODULE_INIT
+void mod_init(void)
+{
+    // Do initialization
+}
+
+// enclave_app.c
+int main_function()
+{
+    int val = always_used_symbol;
+    // Do stuff
+}
+```
+
+2. Pass a flag to tell the linker to always link the initialization function.
+
+```
+OE_MODULE_INIT
+void mod_init(void)
+{
+    // Do initialization
+}
+```
+
+Then during compilation `$CC -o enclave -lmymodule.a -W,l,-u mod_init`
+
+3. Use a custom GNU linker script and mark the `.init_array` option as KEEP.
+
+TODO: Link to documentation on GNU link scripts and KEEP directive.
--- a/include/openenclave/internal/defs.h
+++ b/include/openenclave/internal/defs.h
@@ -71,4 +71,12 @@
 /* OE_FIELD_SIZE */
 #define OE_FIELD_SIZE(TYPE, FIELD) (sizeof(((TYPE*)0)->FIELD))
 
+/* OE_MODULE_INIT */
+#ifdef __GNUC__
+#define OE_MODULE_INIT __attribute__((used, constructor))
+#define OE_MODULE_INIT_PRIORITY(__p) __attribute__((used, constructor(__p)))
+#else
+#error OE_MODULE_INIT not implemented
+#endif
+
 #endif /* _OE_INTERNAL_DEFS_H */
--- a/libc/CMakeLists.txt
+++ b/libc/CMakeLists.txt
@@ -2,7 +2,8 @@
 # Licensed under the MIT License.
 
 # Build the C library, using local sources and sources from musl.
-set(MUSLSRC ${PROJECT_SOURCE_DIR}/3rdparty/musl/musl/src)
+set(MUSLROOT ${PROJECT_SOURCE_DIR}/3rdparty/musl)
+set(MUSLSRC ${MUSLROOT}/musl/src)
 
 if (OE_SGX)
     add_library(oelibasm OBJECT
@@ -64,6 +65,7 @@ add_library(oelibc STATIC
     getaddrinfo.c
     getnameinfo.c
     kill.c
+    libc.c
     libunwind_stubs.c
     link.c
     locale.c
@@ -195,7 +197,6 @@ add_library(oelibc STATIC
     ${MUSLSRC}/fenv/__flt_rounds.c
     ${MUSLSRC}/internal/floatscan.c
     ${MUSLSRC}/internal/intscan.c
-    ${MUSLSRC}/internal/libc.c
     ${MUSLSRC}/internal/shgetc.c
     ${MUSLSRC}/legacy/getpagesize.c
     ${MUSLSRC}/locale/bind_textdomain_codeset.c
--- a/libc/libc.c
+++ b/libc/libc.c
@@ -1,0 +1,39 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include "libc.h"
+
+#include <openenclave/internal/defs.h>
+
+struct __libc __libc;
+
+size_t __hwcap;
+size_t __sysinfo;
+char *__progname = 0, *__progname_full = 0;
+
+weak_alias(__progname, program_invocation_short_name);
+weak_alias(__progname_full, program_invocation_name);
+
+// Modifications start here
+extern char** __environ;
+static size_t __auxv;
+
+/*
+ * Initialize Musl libc in a global constructor which will be called on enclave
+ * load. Ensure that the libc initialization is the first init function called
+ * by giving it a priority of 0.
+ *
+ * Module constructor functions must be in the same file as another symbol
+ * which is referenced from the main enclave bianry. This is why this function
+ * does not exist in its own file. Alternatively, the linker could be explicitly
+ * told to keep this symbol, by calling the linker with `-u oe_init_c`.
+ */
+OE_MODULE_INIT_PRIORITY(0)
+void oe_init_c(void)
+{
+    __environ = 0;
+    __auxv = 0;
+    libc.auxv = &__auxv;
+    libc.page_size = OE_PAGE_SIZE;
+    libc.secure = 1;
+}


enclave/core/CMakeLists.txt
```diff
@@ -264,8 +265,10 @@ if(OE_SGX)
         -Wl,--no-undefined,-Bstatic,-Bsymbolic,--export-dynamic,-pie,--build-id
         -Wl,-z,noexecstack
         -Wl,-z,now
-        -Wl,-gc-sections)
+        -Wl,-gc-sections
+        -Wl,-gc-keep-exported)
```
jhand2: `-gc-keep-exported` has no effect here. Remove..

enclave/core/CMakeLists.txt
```diff
@@ -264,8 +265,10 @@ if(OE_SGX)
...
 elseif(OE_TRUSTZONE)
+    # TODO: Figure out how to keep init sections for trustzone
```
jhand2: Outdated comment. Remove.

enclave/core/CMakeLists.txt
```diff
@@ -275,5 +278,10 @@ elseif(OE_TRUSTZONE)
         --as-needed)
 endif()
 
+# Ensure all moodule initialization functions are linked into the
+# enclave binary regardless of whether they are used explicitly
+target_link_libraries(oelibc INTERFACE
+    -Wl,-u,mod_init*)
```
jhand2: Consider documenting this somewhere as a required linker flag for enclave which use OE modules. Ideally there is a cleaner way to do this, but I haven't found one yet.

enclave/core/oeinit.c
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Open Enclave SDK contributors.
```
jhand2: Call this file `mod_init.c`

enclave/core/oeinit.c
```diff
@@ -0,0 +1,36 @@
...
+{
+}
+OE_MOD_INIT(__dummy_init);
+
+/*
+ **==============================================================================
+ **
+ ** oe_call_oeinit()
+ **
+ **      TODO: Document what this function does
```
jhand2: Document

enclave/core/oeinit.c
```diff
@@ -0,0 +1,36 @@
...
+ **
+ **==============================================================================
+ */
+void oe_call_oeinit(void)
```
jhand2: Call this function `oe_call_mod_init`

enclave/core/oeinit.h
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Open Enclave SDK contributors.
```
jhand2: Name this file `mod_init.h`

include/openenclave/bits/oeinit.h
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT License.
+
+#ifndef _OE_BITS_OEINIT_H
+#define _OE_BITS_OEINIT_H
+
+#ifdef __GNUC__
+typedef void (*oe_mod_init_t)();
+
+#define OE_MOD_INIT(__f) \
+    oe_mod_init_t mod_init_##__f __attribute__((used, section("oeinit"))) = __f
```
jhand2: Consider changing section name to `.oe_mod_init`

jhand2:
From @anakrish on #2420

> Can you see if you can express order of initialization of libraries as well? Say encryptedfs might depend on hostfs and therefore hostfs must be initialized before encryptedfs.

enclave/core/CMakeLists.txt
jhand2: This doesn't actually work. It only works if the full symbol name is given.

jhand2: @radhikaj @CodeMonkeyLeet @anakrish 

One way to force the module init function to link into the enclave binary is to use this linker hack (`-u <symbol_name`).

Another way is to reference the symbol in the main executable. This doesn't need to be via a function call. We could add a macro `OE_MODULE_INCLUDE(<mod_name>);` that can be added anywhere in the enclave code. This will ensure the initialization function is linked.

The only problem with this is that because modules are really just .a libraries, if the developer forgets to add this line, they can still call functions in the module (such as libc functions), the module will just not have been initialized yet. That said, the same is true of if the developer forgets to add the `-u` linker flag.

I don't really like the linker hack but it does work. Wanted to get some thoughts on which approach people prefer.

3rdparty/musl/oe_init_c.c
```diff
@@ -0,0 +1,25 @@
...
+#include <openenclave/bits/oeinit.h>
+
+static size_t __auxv;
+
+/*
+ * Initialize Musl libc in a global constructor which will be called on enclave
+ * load. Ensure that the libc initialization is the first init function called
+ * by giving it a priority of 0.
+ */
+static void oe_init_c(void)
```
anakrish: I thought you were planning to add `__attribute__((constructor(0))`  to oe_init_c. Don't constructors get preserved by the linker?

anakrish: Placing this in a separate file might be the reason why the linker is not picking this up. This object file might not even be linked in since no symbol within it is referenced outside.

enclave/core/CMakeLists.txt
anakrish: @jhand2 I'd prototyped a similar approach earlier: https://github.com/openenclave/openenclave/compare/anakrish/feature/initializers
I am not sure whether it would work given that we now aggressively remove sections.

I like your idea of using constructor attributes and priorities. If we can make that work, then the compiler/linker takes care of most of the work for us.

3rdparty/musl/oe_init_c.c
jhand2: Constructors aren't treated any differently. And different/same file doesn't seem to make a difference because each function is put in a different section anyway.

The reason I chose not to use a constructor is that in the specification, priority 0-100 is reserved for the compiler. So even if you use priority 0, the compiler is still allowed to also use 0 for some C++ constructor. Using a different section ensures we call all of these init functions prior to all global constructors.

jhand2: Constructors are only preserved if they are in a .so or are part of the main executable. Not if they are linked from static libraries.

enclave/core/CMakeLists.txt
jhand2: I think the problem is that the constructor attribute isn't designed to allow you to insert calls prior to C++ constructors. I think it is better to just implement priorities ourselves. Its pretty easy, you can just have initialization stages with a different section for each stage. So your sections would be: `.oeinit.early`, `.oeinit.default`, `.oeinit.late`, etc. Either way, we still run into the linker optimization problem.

The linux kernel actually faces a similar problem with kernel drivers. Drivers can either be dynamically loadable (.ko shared object files) or compiled into the kernel statically. They address this problem by using a linker script that marks their .init section with a `KEEP()`.

I think a linker script is actually the correct way to do this. The problem is that Linux can handle this a little more cleanly because they control the linking step. Technically we don't, so we would have to provide a linker script and get the enclave developer to pass the linker script to gcc/clang.

3rdparty/musl/oe_init_c.c
jhand2: I have modified the PR to go back to using constructors and written a design doc which discusses the linking options.

docs/DesignDocs/openenclave_modules.md
```diff
@@ -0,0 +1,105 @@
+# OpenEnclave Modules
+
+The Open Enclave SDK provides a core set of functionality for build enclave
+applications. Some of these features&2014;such as memory management and ecall/ocall
```
jhand2: This unicode character doesn't render properly

docs/DesignDocs/openenclave_modules.md
```diff
@@ -0,0 +1,105 @@
...
+void OE_MOD_INIT_PRI(1) mod2_init(void)
+{
+    // Do stuff that depends on mod1. This function will
+    // always be called after mod1_init.
+}
+```
+
+### Linker disclaimer
+
+Because modules can be added and removed at compile-time without the core OE
```
mingweishih: Can you explain a bit more on why/how modules can be added / removed at compile-time? Is there any case other than modifying the Makefiles?

docs/DesignDocs/openenclave_modules.md
```diff
@@ -0,0 +1,105 @@
...
+library knowing about them, initialization functions are not actually
+referenced by symbol. They are instead kept in their own section in the binary.
+To ensure a module's initialization function is included in the final enclave
```
mingweishih: Just to clarify: the goal here is to ensure the initialization function is included only when the module is added, is that correct?

docs/DesignDocs/openenclave_modules.md
```diff
@@ -0,0 +1,105 @@
...
+void OE_MOD_INIT mod_init(void)
+{
+    // Do initialization
+}
+```
+
+Then during compilation `$CC -o enclave -lmymodule.a -W,l,-u mod_init`
+
+3. Use a custom GNU linker script and mark the `.init_array` option as KEEP.
+
```
mingweishih: Consider adding test cases / templates for third-party modules that would be helpful for developers to adopt these options when writing their own module.

docs/DesignDocs/openenclave_modules.md
```diff
@@ -0,0 +1,105 @@
...
+    // Do stuff
+}
+```
+
+2. `OE_MOD_INIT_PRI` Attribute
+
+Optionally, modules can specify a priority for their initialization functions.
+Functions with a lower priority value will be called before functions with a
+higher priority value. This can be used if one module depends on another being
+initialized first.
```
mingweishih: What happen if functions are assigned with the same priority? I assume the order will just not be guaranteed, correct?

docs/DesignDocs/openenclave_modules.md
```diff
@@ -0,0 +1,105 @@
...
+
+* hostfs
+* libc
+* hostepoll
+* hostsock
+* hostresolver
+
+## Module initialization
+
+OE modules can each have an initialization function which will be called on
```
mingweishih: Just a random thought: would it be possible to support pass-by-value for the initialization functions? I imagine this would complicate the design (i.e., difficult to generalize). However, this may be needed for some use cases (e.g., larger number of functions with dependencies).

docs/DesignDocs/openenclave_modules.md
```diff
@@ -0,0 +1,105 @@
...
+first entry to the enclave. This ensures that a module are fully initialized
+prior to an application using it. There are two ways to demark initialization
+functions:
+
+1. `OE_MOD_INIT Attribute`
```
anakrish: Consider using full names for readability.

OE_MODULE_INITIALIZER and OE_MODULE_INITIALIZER_WITH_PRIORITY(p)

docs/DesignDocs/openenclave_modules.md
```diff
@@ -0,0 +1,105 @@
...
+
+The `OE_MOD_INIT` attribute can be applied to a function. This ensures that the
+function is called on enclave entry.
+
+```
+void OE_MOD_INIT init_my_module(void)
```
anakrish: Placing the attribute before the return type would make it consisted with other attributes like OE_NEVER_INLINE.

```c
OE_MODULE_INITIALIZER
void init_my_module(void)
{
    ...
}
```

docs/DesignDocs/openenclave_modules.md
```diff
@@ -0,0 +1,105 @@
...
+### Linker disclaimer
+
+Because modules can be added and removed at compile-time without the core OE
+library knowing about them, initialization functions are not actually
+referenced by symbol. They are instead kept in their own section in the binary.
+To ensure a module's initialization function is included in the final enclave
+binary there are 3 options:
+
+1. Add the module initialization function in the same file as a symbol that is
+referenced in the enclave binary.
```
anakrish: Would be good to explain a bit more about what `referenced in the enclave binary` means. Essentially a symbol references by any of the functions that are called directly or indirectly by `_start` function would be retained in the enclave binary. Additionally symbols referenced by constructors defined in files that define functions called directly or indirectly by _start would also be retained.

docs/DesignDocs/openenclave_modules.md
```diff
@@ -0,0 +1,105 @@
...
+2. Pass a flag to tell the linker to always link the initialization function.
+
+```
+void OE_MOD_INIT mod_init(void)
+{
+    // Do initialization
+}
+```
+
+Then during compilation `$CC -o enclave -lmymodule.a -W,l,-u mod_init`
```
anakrish: --require-defined would be a better option than -u since the former would error out if the symbol is not found whereas the latter won't.

docs/DesignDocs/openenclave_modules.md
anakrish: Consider adding test-case where one module depends on the other being initialized.

docs/DesignDocs/openenclave_modules.md
jhand2: Do you mean for specifying constructor order? Using the builtin compiler attribute mechanism I think the answer is no, although I may be able to come up with something with a little preprocessor magic

docs/DesignDocs/openenclave_modules.md
jhand2: Correct, order is up to the compiler in that case

docs/DesignDocs/openenclave_modules.md
jhand2: It is just by passing them to the linker or not passing them. So no mechanism other than modifying the build files.

docs/DesignDocs/openenclave_modules.md
jhand2: Correct. More specifically, only when it is linked

docs/DesignDocs/openenclave_modules.md
```diff
@@ -0,0 +1,110 @@
+# OpenEnclave Modules
+
+The Open Enclave SDK provides a core set of functionality for build enclave
+applications. Some of these features &2014; such as memory management and ecall/ocall
```
jhand2: Unicode still doesn't render. Just use normal dashes

include/openenclave/internal/defs.h
```diff
@@ -71,4 +71,12 @@
 /* OE_FIELD_SIZE */
 #define OE_FIELD_SIZE(TYPE, FIELD) (sizeof(((TYPE*)0)->FIELD))
 
+/* OE_MODULE_INIT */
+#ifdef __GNUC__
+#define OE_MODULE_INIT __attribute__((used, constructor))
+#define OE_MODULE_INIT_PRIORITY(__p) __attribute__((used, constructor(__p)))
```
jhand2: @anakrish one though. If we ever support dynamic loading, this doesn't really work. Constructor priority wouldn't be guaranteed accross .so files.

The staged model (of having something like early, default, late initialization stages) is a semantic we could extend to statically linked or dynamically linked module initialization.

jhand2:
> Nice idea and thanks for the writeup. I'm do have an overall concern that this seems to rely on a compiler specific feature (GCC support for `constructor` attribute with priorities) though. I assume that clang takes the attribute, but does it actually produce the semantics that you expect? It's not explicitly documented as supported in clang 

Interesting, I didn't realize the constructor attribute isn't documented for clang. It does work (I did my testing with clang) but its concerning that it isn't documented.

>and in previous cases (e.g. `section` attribute), there's no failure or warning on use of attributes where unsupported semantics are ignored.

What do you mean? As in clang won't say anything and will silently fail if you use an attribute that it doesn't support?

> It would be helpful to verify with a test that the behavior you want replicates with a test under either compiler.

In any case I agree a test is a good idea

shnwc:
Global module Initialization of host- and enclave-side modules is needed for attestation modules, for self-registration of attester and verifier plugins. Enclave-side module initialization function needs to be able to make OCALLs to the host.

docs/DesignDocs/openenclave_modules.md
```diff
@@ -0,0 +1,110 @@
+# OpenEnclave Modules
+
+The Open Enclave SDK provides a core set of functionality for build enclave
```
dthaler: grammar: "build" -> "building"?

docs/DesignDocs/openenclave_modules.md
```diff
@@ -0,0 +1,110 @@
...
+* hostfs
+* libc
+* hostepoll
+* hostsock
+* hostresolver
+
+## Module initialization
+
+OE modules can each have an initialization function which will be called on
+first entry to the enclave. This ensures that a module are fully initialized
```
dthaler: "a module are" is bad grammar

kuanggang-git:

radhikaj:
Closing this for now. Will reopen when this is prioritized.

PR #2429, opened 2020-01-15
Fix nanosleep() for req == rem

thomasten:
Calling nanosleep(&ts, &ts) resulted in a sleep of 0 because ts was
overwritten before milliseconds were calculated. Also fixes sleep() and
this_thread::sleep_for() which use nanosleep this way.

With this change the following libcxx tests fail:
165 - tests/libcxxtest-std_thread_thread.mutex_thread.lock_thread.lock.shared_thread.lock.shared.cons_mutex.pass (Child aborted)
166 - tests/libcxxtest-std_thread_thread.mutex_thread.lock_thread.lock.shared_thread.lock.shared.locking_lock.pass (Child aborted)
170 - tests/libcxxtest-read.mutex_thread.mutex.requirements_thread.sharedtimedmutex.requirements_thread.sharedtimedmutex.class_lock_shared.pass (Child aborted)

https://github.com/openenclave/openenclave/blob/8911dfdb304bacfcfcd37d1d92a52b3ac039a4c3/3rdparty/libcxx/libcxx/test/std/thread/thread.mutex/thread.mutex.requirements/thread.sharedtimedmutex.requirements/thread.sharedtimedmutex.class/lock_shared.pass.cpp#L38-L80

This is because these tests assert that the mutex operations are fast enough (l. 57). Previously, the sleep in line 77 did not sleep so that the condition was easily satisfied.

These tests still work if the tolerance is increased, e.g., like in line 46. How should I deal with them? Fix them (although it is 3rdparty)? Move to broken? Move to unsupported?

### DIFF
--- a/libc/syscalls.c
+++ b/libc/syscalls.c
@@ -107,20 +107,20 @@ static long _syscall_nanosleep(long n, long x1, long x2)
 
     OE_UNUSED(n);
 
+    if (req)
+    {
+        /* Convert timespec to milliseconds */
+        milliseconds += req->tv_sec * 1000UL;
+        milliseconds += req->tv_nsec / 1000000UL;
+
+        /* Perform OCALL */
+        ret = oe_sleep_msec(milliseconds);
+    }
+
+    /* Clear remaining time. Do this after reading from req because it may be
+     * that req == rem. */
     if (rem)
         memset(rem, 0, sizeof(*rem));
-
-    if (!req)
-        goto done;
-
-    /* Convert timespec to milliseconds */
-    milliseconds += req->tv_sec * 1000UL;
-    milliseconds += req->tv_nsec / 1000000UL;
-
-    /* Perform OCALL */
-    ret = oe_sleep_msec(milliseconds);
-
-done:
 
     return ret;
 }
--- a/tests/stdc/enc/enc.cpp
+++ b/tests/stdc/enc/enc.cpp
@@ -174,6 +174,7 @@ static void _test_time_functions(void)
     const uint64_t SEC_TO_USEC = 1000000UL;
     const uint64_t JAN_1_2018 = 1514786400UL * SEC_TO_USEC;
     const uint64_t JAN_1_2050 = 2524629600UL * SEC_TO_USEC;
+    const uint64_t SLEEP_SECS = 3;
     uint64_t now;
 
     /* Test time(): this test will fail if run after Jan 1, 2050 */
@@ -209,8 +210,6 @@ static void _test_time_functions(void)
 
     /* Test nanosleep() */
     {
-        const uint64_t SLEEP_SECS = 3;
-
         uint64_t before = oe_get_time();
 
         /* Sleep for SLEEP_SECS seconds */
@@ -219,6 +218,33 @@ static void _test_time_functions(void)
             timespec rem;
             OE_TEST(nanosleep(&req, &rem) == 0);
         }
+
+        uint64_t after = oe_get_time();
+
+        OE_TEST(after > before);
+    }
+
+    /* Test nanosleep() with req==rem */
+    {
+        uint64_t before = oe_get_time();
+
+        /* Sleep for SLEEP_SECS seconds */
+        {
+            timespec ts = {SLEEP_SECS, 0};
+            OE_TEST(nanosleep(&ts, &ts) == 0);
+        }
+
+        uint64_t after = oe_get_time();
+
+        OE_TEST(after > before);
+    }
+
+    /* Test sleep() */
+    {
+        uint64_t before = oe_get_time();
+
+        /* Sleep for SLEEP_SECS seconds */
+        OE_TEST(sleep(SLEEP_SECS) == 0);
 
         uint64_t after = oe_get_time();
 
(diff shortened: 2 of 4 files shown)


tests/stdc/enc/enc.cpp
```diff
@@ -224,6 +223,33 @@ static void _test_time_functions(void)
...
+
+        /* Sleep for SLEEP_SECS seconds */
+        {
+            timespec ts = {SLEEP_SECS, 0};
+            OE_TEST(nanosleep(&ts, &ts) == 0);
+        }
+
+        uint64_t after = oe_get_time();
+
+        OE_TEST(after > before);
```
jhand2: Won't this always be true, regardless of whether the sleep worked or not?

tests/stdc/enc/enc.cpp
```diff
@@ -224,6 +223,33 @@ static void _test_time_functions(void)
...
     }
+
+    /* Test nanosleep() with req==rem */
+    {
+        uint64_t before = oe_get_time();
+
+        /* Sleep for SLEEP_SECS seconds */
+        {
+            timespec ts = {SLEEP_SECS, 0};
+            OE_TEST(nanosleep(&ts, &ts) == 0);
```
jhand2: Should this also check if ts is zero'd out?

tests/stdc/enc/enc.cpp
```diff
@@ -224,6 +223,33 @@ static void _test_time_functions(void)
...
+    /* Test sleep() */
+    {
+        uint64_t before = oe_get_time();
+
+        /* Sleep for SLEEP_SECS seconds */
+        OE_TEST(sleep(SLEEP_SECS) == 0);
+
+        uint64_t after = oe_get_time();
+
+        OE_TEST(after > before);
```
jhand2: same as above

tests/stdc/enc/enc.cpp
```diff
@@ -174,6 +174,7 @@ static void _test_time_functions(void)
     const uint64_t SEC_TO_USEC = 1000000UL;
     const uint64_t JAN_1_2018 = 1514786400UL * SEC_TO_USEC;
     const uint64_t JAN_1_2050 = 2524629600UL * SEC_TO_USEC;
+    const uint64_t SLEEP_SECS = 3;
```
jhand2: 3 seconds is pretty long, given that we're calling sleep multiple times. Is the test equally effective if we reduce this to 1?

jhand2:
> This is because these tests assert that the mutex operations are fast enough (l. 57). Previously, the sleep in line 77 did not sleep so that the condition was easily satisfied.
>
> These tests still work if the tolerance is increased, e.g., like in line 46. How should I deal with them? Fix them (although it is 3rdparty)? Move to broken? Move to unsupported?

I'm not sure why that sleep is there in the first place. It would appear the test has always worked without it, maybe we should just remove it.

cc @anakrish @CodeMonkeyLeet

mingweishih:
May a unrelated question: why do we need to clear the `rem`? Is that the intended behavior of the nanosleep syscall?

jhand2:
> May a unrelated question: why do we need to clear the `rem`? Is that the intended behavior of the nanosleep syscall?

Yes, thats how nanosleep works. It can be interrupted, so if it returns and rem is non-zero, the caller the sleep the remaining time by passing rem as the first argument to nanosleep.

libc/syscalls.c
```diff
@@ -107,20 +107,20 @@ static long _syscall_nanosleep(long n, long x1, long x2)
...
-    /* Perform OCALL */
-    ret = oe_sleep_msec(milliseconds);
+        /* Perform OCALL */
+        ret = oe_sleep_msec(milliseconds);
+    }
 
-done:
+    /* Clear remaining time. Do this after reading from req because it may be
+     * that req == rem. */
+    if (rem)
```
mingweishih: Shouldn't we check the `ret` here as well? Or we just ignore the `rem` regardless what the `ret` will be.

CodeMonkeyLeet: +1.

I'm not sure this can be properly addressed by this PR though; the stack looks incorrectly implemented on the host side since we break the `nanosleep()` semantics by not marshalling the expected `rem` value back to the enclave, so the value is always all or nothing and not dependent on `EINTR`.

tests/stdc/enc/enc.cpp
CodeMonkeyLeet: Agreed, although this is an existing problem with the test that was just replicated for new cases in this PR.

thomasten: In fact it occasionally fails without the fix. (oe_get_time returns milliseconds.)

tests/stdc/enc/enc.cpp
thomasten: I'm not sure. The real nanosleep does not change rem if it has not been interrupted.

tests/stdc/enc/enc.cpp
thomasten: Yes, I changed that.

jhand2:
@thomasten thanks for your contribution (your commit for the libcxx tests was retained in my change to port nanosleep to a normal syscall)

Can you confirm that this is fixed and this PR can be closed?

thomasten:
Yes, this is fixed, thanks!

radhikaj:
Thanks @jhand2

PR #2430, opened 2020-01-15
Update libcxx_tests.Jenkinsfile

rbalteanu:


rbalteanu:
Updated libcxx_tests.Jenkinsfile so that it only runs tests on the pre-existing environment, it does not create the environment anymore.

jhand2:
Can you squash into one commit? It'll make things easier when/if this gets merged to master

BRMcLaren:
Applying do not merge label, this PR should to go in after the terraform work otherwise it will create a daily 4 hour downtime in the system.

jhand2:
> Applying do not merge label, this PR should to go in after the terraform work otherwise it will create a daily 4 hour downtime in the system.

@BRMcLaren the PR is into the branch `scalable_jenkins_terraform` rather than `master`. Isn't that correct?

BRMcLaren:
@jhand2 Good catch, not enough coffee this morning. You are correct.

.jenkins/libcxx_tests.Jenkinsfile
```diff
@@ -101,44 +25,27 @@ def ACClibcxxTest(String label, String compiler, String build_type) {
...
-             "libcxx ACC1604 gcc Debug" :              { ACClibcxxTest(XENIAL_LABEL, 'gcc', 'Debug') },
-             "libcxx ACC1604 gcc Release" :            { ACClibcxxTest(XENIAL_LABEL, 'gcc', 'Release') },
-             "libcxx ACC1604 gcc RelWithDebInfo" :     { ACClibcxxTest(XENIAL_LABEL, 'gcc', 'RelWithDebInfo') },
-             "libcxx ACC1804 clang-7 Debug" :          { ACClibcxxTest(BIONIC_LABEL, 'clang-7', 'Debug') },
-             "libcxx ACC1804 clang-7 Release" :        { ACClibcxxTest(BIONIC_LABEL, 'clang-7', 'Release') },
-             "libcxx ACC1804 clang-7 RelWithDebInfo" : { ACClibcxxTest(BIONIC_LABEL, 'clang-7', 'RelWithDebinfo') },
-             "libcxx ACC1804 gcc Debug" :              { ACClibcxxTest(BIONIC_LABEL, 'gcc', 'Debug') },
-             "libcxx ACC1804 gcc Release" :            { ACClibcxxTest(BIONIC_LABEL, 'gcc', 'Release') },
-             "libcxx ACC1804 gcc RelWithDebInfo" :     { ACClibcxxTest(BIONIC_LABEL, 'gcc', 'RelWithDebinfo') }
+    parallel "libcxx ACC1604 clang-7 Debug" :          { ACClibcxxTest("LIBCXX-1604", 'clang-7', 'Debug') },
```
BRMcLaren: What you could do is: 

Change the original from 

XENIAL_LABEL = "LIBCXX-${BUILD_NUMBER}-1604"
BIONIC_LABEL = "LIBCXX-${BUILD_NUMBER}-1804"

-> 

XENIAL_LABEL = "LIBCXX-1604"
BIONIC_LABEL = "LIBCXX-1804"

Then this section does not need to be changed and there's less pasting of the same variable. Which I think is preferrable.

.jenkins/libcxx_tests.Jenkinsfile
```diff
@@ -101,44 +25,27 @@ def ACClibcxxTest(String label, String compiler, String build_type) {
...
+             "libcxx ACC1804 clang-7 RelWithDebInfo" : { ACClibcxxTest("LIBCXX-1804", 'clang-7', 'RelWithDebinfo') },
+             "libcxx ACC1804 gcc Debug" :              { ACClibcxxTest("LIBCXX-1804", 'gcc', 'Debug') },
+             "libcxx ACC1804 gcc Release" :            { ACClibcxxTest("LIBCXX-1804", 'gcc', 'Release') },
+             "libcxx ACC1804 gcc RelWithDebInfo" :     { ACClibcxxTest("LIBCXX-1804", 'gcc', 'RelWithDebinfo') }
 } catch(Exception e) {
     println "Caught global pipeline exception :" + e
     GLOBAL_ERROR = e
     throw e   
 } finally {
-    cleanup()
```
BRMcLaren: We should definitely unregister the slaves as you have done, but I am unsure about deleting the cleanUp functionality. I think it should be modified to no longer delete the resource groups.

.jenkins/libcxx_tests.Jenkinsfile
```diff
@@ -101,44 +25,27 @@ def ACClibcxxTest(String label, String compiler, String build_type) {
     }
 }
 
-def cleanup(){
-    node("nonSGX") {
-        timeout(GLOBAL_TIMEOUT_MINUTES) {
-            cleanWs()
-            checkout scm
-            oe.deleteRG([XENIAL_RG, BIONIC_RG])
```
BRMcLaren: @oprinmarius Is it possible to do something malicious with this functionality?

I.e. could one do oe.delete(master_rg) or would it fail?

.jenkins/libcxx_tests.Jenkinsfile
oprinmarius: @BRMcLaren We do not need this function anymore. It was used to delete and unregister the agents spawned by the ACCDeployVM function , which we do not use anymore.
Nodes will be cleaned up by the Azure VM plugin

.jenkins/libcxx_tests.Jenkinsfile
oprinmarius: @BRMcLaren Master RG has delete lock enabled . But we won't be using this function in any of the jobs from now on.

.jenkins/libcxx_tests.Jenkinsfile
oprinmarius: I agree with this

PR #2433, opened 2020-01-15
Only allow oe_handle_init_switchless to be called once

### LINKED ISSUE #2187: oe_handle_init_switchless can be called multiple times
We should error out if that's called more than once.

jhand2:
Raise an error if `oe_handle_init_switchless` is called again after initialization. Even though is only ever called by the internal implementation, adding this check can't hurt. Especially if we later change how/when initialization happens.

Fixes #2187

### DIFF
--- a/enclave/core/switchlesscalls.c
+++ b/enclave/core/switchlesscalls.c
@@ -14,6 +14,11 @@ static size_t _host_worker_count = 0;
 // The array of host worker contexts. Initialized by host through ECALL
 static oe_host_worker_context_t* _host_worker_contexts = NULL;
 
+// Flag to denote if switchless calls have already been initialized.
+static bool _is_switchless_initialized = false;
+
+static bool _switchless_init_in_progress = false;
+
 /*
 **==============================================================================
 **
@@ -25,7 +30,12 @@ static oe_host_worker_context_t* _host_worker_contexts = NULL;
 */
 bool oe_is_switchless_initialized()
 {
-    return _host_worker_count != 0;
+    bool is_initialized;
+
+    is_initialized =
+        __atomic_load_n(&_is_switchless_initialized, __ATOMIC_SEQ_CST);
+
+    return is_initialized;
 }
 
 /*
@@ -46,6 +56,19 @@ oe_result_t oe_handle_init_switchless(uint64_t arg_in)
 
     if (arg_in == 0)
         OE_RAISE(OE_INVALID_PARAMETER);
+
+    if (!oe_atomic_compare_and_swap(
+            (volatile int64_t*)&_switchless_init_in_progress,
+            (int64_t) false,
+            (int64_t) true))
+    {
+        OE_RAISE(OE_BUSY);
+    }
+
+    if (oe_is_switchless_initialized())
+    {
+        OE_RAISE(OE_ALREADY_INITIALIZED);
+    }
 
     manager = (oe_switchless_call_manager_t*)arg_in;
     safe_manager = *manager;
@@ -71,9 +94,14 @@ oe_result_t oe_handle_init_switchless(uint64_t arg_in)
     // Copy the worker context array pointer and its size to avoid TOCTOU
     _host_worker_count = safe_manager.num_host_workers;
     _host_worker_contexts = safe_manager.host_worker_contexts;
+
+    __atomic_store_n(&_is_switchless_initialized, true, __ATOMIC_SEQ_CST);
+
     result = OE_OK;
 
 done:
+    __atomic_store_n(&_switchless_init_in_progress, false, __ATOMIC_SEQ_CST);
+
     return result;
 }
 
--- a/common/result.c
+++ b/common/result.c
@@ -128,6 +128,8 @@ const char* oe_result_str(oe_result_t result)
             return "OE_THREAD_JOIN_ERROR";
         case OE_ALREADY_EXISTS:
             return "OE_ALREADY_EXISTS";
+        case OE_ALREADY_INITIALIZED:
+            return "OE_ALREADY_INITIALIZED";
         case __OE_RESULT_MAX:
             break;
     }
@@ -195,6 +197,7 @@ bool oe_is_valid_result(uint32_t result)
         case OE_THREAD_CREATE_ERROR:
         case OE_THREAD_JOIN_ERROR:
         case OE_ALREADY_EXISTS:
+        case OE_ALREADY_INITIALIZED:
         {
             return true;
         }
--- a/include/openenclave/bits/result.h
+++ b/include/openenclave/bits/result.h
@@ -339,6 +339,11 @@ typedef enum _oe_result
      */
     OE_ALREADY_EXISTS,
 
+    /**
+     * The desired resource has already been initialized.
+     */
+    OE_ALREADY_INITIALIZED,
+
     __OE_RESULT_MAX = OE_ENUM_MAX,
 } oe_result_t;
 /**< typedef enum _oe_result oe_result_t*/


enclave/core/switchlesscalls.c
```diff
@@ -44,6 +44,9 @@ oe_result_t oe_handle_init_switchless(uint64_t arg_in)
     oe_switchless_call_manager_t safe_manager;
     size_t contexts_size, threads_size;
 
+    if (oe_is_switchless_initialized())
```
anakrish: Is this thread-safe?

jhand2: cc @jxyang Why is oe_handle_init_switchless exposed in a public header? From the host `oe_start_switchless_manager()` is only in an internal header.

And then from a philosophy perspective @anakrish, if a function is internal and we only call it once, can we safely ignore thread safety even though someone could technically include an internal header and call it?

anakrish: @jhand2 The reason for making it thread-safe is from a security perspective. Our philosophy is that enclaves don't trust the host; a host can intentionally trigger code paths that we have not anticipated.

Eventhough, oe_handle_init_switchless is an internal function, a host can launch two threads and make the OE_ECALL_INIT_CONTEXT_SWITCHLESS ecall. If we don't proactively guard against such malicious calls, the behavior of the enclave is undefined, potentially causing confidentiality vulnerabilities.

jhand2: That makes sense. I'll add proper logging :smile:

enclave/core/switchlesscalls.c
```diff
@@ -44,6 +53,12 @@ oe_result_t oe_handle_init_switchless(uint64_t arg_in)
     oe_switchless_call_manager_t safe_manager;
     size_t contexts_size, threads_size;
 
+    oe_spin_lock(&_lock);
+
+    // Check if already initialized
+    if (_host_worker_count != 0)
+        OE_RAISE(OE_ALREADY_EXISTS);
```
anakrish: OE_RAISE would bypass unlocking of the spinlock.
A coding pattern to handle this cleanly is:

```c
bool locked = oe_spin_lock(&_lock);
....
...

done:
    if(locked)
        oe_spin_unlock(&_lock);
```

enclave/core/switchlesscalls.c
```diff
@@ -25,7 +28,13 @@ static oe_host_worker_context_t* _host_worker_contexts = NULL;
 */
 bool oe_is_switchless_initialized()
 {
-    return _host_worker_count != 0;
+    bool is_initialized;
+
+    oe_spin_lock(&_lock);
+    is_initialized = _host_worker_count != 0;
```
anakrish: Suggestion: Consider using acquire-release on the _host_worker_count instead of a spinlock in this function.

enclave/core/switchlesscalls.c
jxyang: I think this error name `OE_ALREADY_EXISTS` is  ambiguous, probably don't fit into the scenario as well. Can we have a specific error code like `OE_INITIALIZED_MORE_THAN_ONCE`?

jhand2: I know that `OE_ALREADY_EXISTS` implies there was some resource being created but I think its clear given the context of the function being called (an init function) that OE_ALREADY_EXISTS means that switchless has already been initialized.

I understand your point, and we could add something like `OE_ALREADY_INITIALIZED`  but I don't want to create too many error codes that mean very similar things.

jxyang: Without reading the comments in result.h, I'd think `OE_ALREADY_EXIST` means the enclave already exists. I bet most people would think similarly.

jhand2: I think the error has to be taken into context of where it was thrown from (which verbose logging will tell you). Because even if we add `OE_ALREADY_INITIALIZED`, its not clear that it is referring to switchless calls unless you look at which function threw the error. You could similarly infer that the enclave itself has already been initialized.

jxyang: I am speaking from a readability perspective. Sorry for being picky. I just don't think `OE_ALREADY_EXIST` should be an error code name at the first place. `OE_ALREADY_INITIALIZED` is a much better name, and covers all scenarios that a singleton object is initialized twice.

jhand2: I agree that `OE_ALREADY_INITIALIZED` or even `OE_ALREADY` would be better names. The problem is that changing the name of an enum value is a breaking change in a public header and I don't know if this case warrants making that change.

At the very least if we are going to change the name of the enum I think it should be done in a different PR.

anakrish: I agree, it would be better not to make a breaking change.

enclave/core/switchlesscalls.c
```diff
@@ -25,7 +27,12 @@ static oe_host_worker_context_t* _host_worker_contexts = NULL;
 */
 bool oe_is_switchless_initialized()
 {
-    return _host_worker_count != 0;
+    bool is_initialized;
+
+    is_initialized =
```
anakrish: Nice.

enclave/core/switchlesscalls.c
```diff
@@ -14,6 +14,8 @@ static size_t _host_worker_count = 0;
 // The array of host worker contexts. Initialized by host through ECALL
 static oe_host_worker_context_t* _host_worker_contexts = NULL;
 
+static bool is_switchless_initialized = false;
```
anakrish: For consistency, follow the same naming convention as the variable above.

enclave/core/switchlesscalls.c
```diff
@@ -44,6 +51,16 @@ oe_result_t oe_handle_init_switchless(uint64_t arg_in)
     oe_switchless_call_manager_t safe_manager;
     size_t contexts_size, threads_size;
 
+    // Check if already initialized
+    if (!oe_atomic_compare_and_swap(
```
anakrish: Nice work using atomics instead of locking.

When two threads simultaneously call oe_handle_init_switchless

- one of them will atomically set is_switchless_initialized flag to true **and then** initialize the data structures.
- the other thread will see that the flag is true and error out

Another scenario to solve is 
- When one thread has set the flag to true, but **not yet completed initialization**
- Another enclave thread could check the flag (via oe_is_switchless_initialized) and incorrectly think that initialization is complete and go on to make a switchless call.

jhand2: Good point. I have updated the ordering of the checks and added a comment for the reasoning.

enclave/core/switchlesscalls.c
jxyang: Please create an issue for the name changes.

enclave/core/switchlesscalls.c
jxyang: @jhand2 Your comments are related for the 1st scenario only. 

For the 2nd scenario, depending on how much progress the initializing thread has made after setting `_is_switchless_initialized` to true. The 2nd thread may either:

* fall back to regular OCALL if the first thread has not set `_host_worker_count`
* crash if the first thread has set `_host_worker_count`, but not `_host_worker_contexts`

Ideally, we should make `oe_is_switchless_initialized` checking `_host_worker_contexts` instead of a flag doesn't really indicate the initialization has been completed or not.

jhand2: The only way to protect against that is to use a lock. Individual atomic operations are only effective if the variables can be updated independently. But if you want to update all 3 variables in one atomic operation I think a spin lock is required here.

cc @anakrish

enclave/core/switchlesscalls.c
jhand2: #2456

enclave/core/switchlesscalls.c
jhand2: Plus we would be locking a very small section of code so the perf hit of using a spinlock is very minimal

anakrish: There are a few ways to implement this correctly:

- Use a spinlock in both `oe_handle_init_switchless` and `oe_is_switchless_initialized`. Make sure the spinlock is released in error scenarios.

- Use one atomic-flag to make sure `oe_handle_init_switchless` performs initialization only once. Use another atomic-flag to check if initialization is complete or not. This second flag must follow acquire-release semantics or the compiler could reorder instructions and set this flag at a different location than written in code. This is in essense what Xuejun is suggesting.

There is also `oe_once` which allows initialization functions to be called only once. However, it does not support initialization functions with arguments.

jhand2: IMO spinlock is cleaner and probably more efficient (although I'm not that worried about perf in this case).

Using two atomically updated flags is a little hard to follow from a readability perspective.

Anyone have another opinion here?

radhikaj:
@anakrish , could you take another look please?

anakrish:
@radhikaj I'm fine with either implementation - spinlock or lockfree. The current changes are not thread-safe.

enclave/core/switchlesscalls.c
jhand2: @anakrish I took a shot at the second one. Can you take a look?

enclave/core/switchlesscalls.c
```diff
@@ -65,15 +83,30 @@ oe_result_t oe_handle_init_switchless(uint64_t arg_in)
         OE_RAISE(OE_INVALID_PARAMETER);
     }
 
+    // Check if already initialized. Doing this check here means that
+    // simultaneous calls to oe_handle_init_switchless may duplicate the above
+    // checks, however this is an acceptably small overhead to ensure that at
+    // least one initialization succeeds in a thread-safe way.
+    if (!oe_atomic_compare_and_swap(
```
anakrish: The initialized flag must be set *after* initialization has been done; otherwise another thread could incorrectly think that initialization has been done. This flag has to be set after line 103.

jhand2: Fixed

jhand2:
@anakrish could you take another look at this?

enclave/core/switchlesscalls.c
```diff
@@ -71,9 +89,21 @@ oe_result_t oe_handle_init_switchless(uint64_t arg_in)
...
+            (int64_t) false,
+            (int64_t) true))
+    {
+        OE_RAISE(OE_ALREADY_INITIALIZED);
+    }
+
     result = OE_OK;
 
 done:
+    __atomic_store_n(&_switchless_init_in_progress, false, __ATOMIC_SEQ_CST);
```
anakrish: @jhand2 First, thanks for taking the suggestions and improving the PR.

When __switchless_init_in_progress_ is reset, it will allow _oe_handle_init_switchless_ to be called again.

Consider renaming this flag to __switchless_init_called_ and not clear this flag. That will match the PR description in a thread-safe manner.

jhand2: I'll make a call to oe_is_switchless_initialized instead. This gives better return code semantics.

anakrish: That would work too.
Consider moving clearing of `in progress` flag to before done to better reflect the initialization semantics.

jhand2:
Test failures look like infra issue (platform errors). Trying again

bors r+

PR #2434, opened 2020-01-16
Allow windows build to get the compiler information of pre-built test

### LINKED ISSUE #2432: Windows build should know which compiler that pre-compiled test cases use
Some test cases depend on the compiler (e.g., https://github.com/openenclave/openenclave/blob/master/tests/libcxx/CMakeLists.txt#L28). However, current Windows build that uses pre-compiled test cases does not know which compiler these test cases use and therefore causes mismatch. This mismatch renders the compilation fails; i.e., the Windows build, without knowing the test cases are compiled with gcc, attempts to find the file, which does not exist.

More information
Config: Win2016 Ubuntu1804 gcc Debug Linux-Elf-build
log: `ninja: error: '../linuxbin/tests/libcxx/enc/libcxxtest-libcxx_containers_unord_next_pow2.pass_enc', needed by 'tests/libcxx_windows_include', missing and no known rule to make it`
Reason: the test case `libcxx_containers_unord_next_pow2.pass_enc` is not generated with gcc

Block #2403

mingweishih:
This PR resolves #2432, which provides a cmake script that allows the windows build to obtain compiler info from the pre-built cmake directories (only if LINUX_BIN_DIR is set). The script introduces two cmake variables `ENCLAVE_TESTS_CXX_COMPILER_ID` and `ENCLAVE_TESTS_CXX_COMPILER_VERSION` that store the corresponding information. Additional information can be exposed in the similar approach if needed.

### DIFF
--- a/tests/libcxx/CMakeLists.txt
+++ b/tests/libcxx/CMakeLists.txt
@@ -12,6 +12,23 @@ add_subdirectory(host)
 
 if (BUILD_ENCLAVES)
     add_subdirectory(enc)
+elseif (LINUX_BIN_DIR)
+    # Find the build information from pre-compiled tests. If found, `LINUX_CXX_COMPILER_ID`
+    # and `LINUX_CXX_COMPILER_VERSION` will hold the corresponding information.
+    list(APPEND CMAKE_MODULE_PATH "${LINUX_BIN_DIR}/libcxx/enc")
+    include(build_info OPTIONAL)
+endif()
+
+if (ADD_WINDOWS_ENCLAVE_TESTS)
+    if (NOT LINUX_CXX_COMPILER_ID OR NOT LINUX_CXX_COMPILER_VERSION)
+        message(FATAL_ERROR "Could not find the compiler information of linux tests!")
+    endif()
+    # Load target LINUX_CXX_COMPILER_ID and LINUX_CXX_COMPILER_VERSION values.
+    set(ENCLAVE_CXX_COMPILER_ID ${LINUX_CXX_COMPILER_ID})
+    set(ENCLAVE_CXX_COMPILER_VERSION ${LINUX_CXX_COMPILER_VERSION})
+else ()
+    set(ENCLAVE_CXX_COMPILER_ID ${CMAKE_CXX_COMPILER_ID})
+    set(ENCLAVE_CXX_COMPILER_VERSION ${CMAKE_CXX_COMPILER_VERSION})
 endif()
 
 foreach(testcase ${alltests})
@@ -20,7 +37,7 @@ foreach(testcase ${alltests})
 # the allocations are entirely optimized out by Clang in these tests and are excluded from Clang release builds -- Skip running these tests
     if ("${name}" MATCHES "cons_default_throws_bad_alloc.pass" OR "${name}" MATCHES "allocator_allocator.members_construct.pass")
         string(TOUPPER ${CMAKE_BUILD_TYPE} BUILD_TYPE_UPPER)
-        if (CMAKE_CXX_COMPILER_ID MATCHES Clang AND BUILD_TYPE_UPPER MATCHES REL)
+        if (ENCLAVE_CXX_COMPILER_ID MATCHES Clang AND BUILD_TYPE_UPPER MATCHES REL)
             continue()
         endif()
     endif()
@@ -30,23 +47,23 @@ foreach(testcase ${alltests})
             OR "${name}" MATCHES "sequences_array_at.pass" 
             OR "${name}" MATCHES "libcxx_containers_unord_next_pow2.pass" 
             OR "${name}" MATCHES "std_utilities_template.bitset_bitset.members_count.pass")
-        if (CMAKE_CXX_COMPILER_ID MATCHES GNU)
+        if (ENCLAVE_CXX_COMPILER_ID MATCHES GNU)
             continue()
         endif()
     endif()
 
 # the following test fails when built with clang, see #830 -- Skipping this test in clang
     if ("${name}" MATCHES "array_sized_delete_array_calls_unsized_delete_array.pass")
-        if (CMAKE_CXX_COMPILER_ID MATCHES Clang)
+        if (ENCLAVE_CXX_COMPILER_ID MATCHES Clang)
             continue()
         endif()
     endif()
 
 # the following test fails on GCC version 7, see  #1523 -- Skipping this test in GCC 7.0 -> 7.5
     if ("${name}" MATCHES "std_utilities_function.objects_comparisons_constexpr_init.pass")
-        if(CMAKE_CXX_COMPILER_ID MATCHES GNU
-               AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7
-               AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.5)
+        if(ENCLAVE_CXX_COMPILER_ID MATCHES GNU
+            AND ENCLAVE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7
+            AND ENCLAVE_CXX_COMPILER_VERSION VERSION_LESS 7.5)
             continue()
         endif()
     endif()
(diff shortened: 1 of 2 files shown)


mingweishih:
bors try

tests/libcxx/enc/CMakeLists.txt
```diff
@@ -82,7 +82,8 @@ foreach(testcase ${alltests})
 # the allocations are entirely optimized out by Clang in these tests and are excluded from Clang release builds
     if ("${name}" MATCHES "cons_default_throws_bad_alloc.pass" OR "${name}" MATCHES "allocator_allocator.members_construct.pass")
         string(TOUPPER ${CMAKE_BUILD_TYPE} BUILD_TYPE_UPPER)
-        if (CMAKE_CXX_COMPILER_ID MATCHES Clang AND BUILD_TYPE_UPPER MATCHES REL)
+        if ((CMAKE_CXX_COMPILER_ID MATCHES Clang OR ENCLAVE_TESTS_CXX_COMPILER_ID MATCHES Clang)
```
CodeMonkeyLeet: This file shouldn't require any changes since the enclaves are only built on Linux right now anyway, no? (`ENCLAVE_TESTS_CXX_COMPILER_ID` doesn't get set if `BUILD_ENCLAVES` is set, and `BUILD_ENCLAVES` is required to build this folder.)

CMakeLists.txt
```diff
@@ -181,6 +181,12 @@ option(ADD_WINDOWS_ENCLAVE_TESTS "Build Windows enclave tests" OFF)
 # Warning: turning on simulation mode on Windows may cause test failures and random crashes
 option(WIN32_SIMULATION "Windows Simulation Mode" OFF)
 
+# Obtain compiler information of the pre-built tests
```
CodeMonkeyLeet: Consider moving this logic into tests/CMakeLists.txt or even tests/libcxx/CMakeLists.txt right now since it's a fairly specific use case.

cmake/get_compiler_info.cmake
```diff
@@ -0,0 +1,27 @@
...
+# This is a helper function to obtain the compiler information
+# (e.g., the GNU or clang compiler and its version) that is used
+# to generate the pre-built enclave test cases.
+#
+function(get_compiler_info)
+    # Check if the path to the pre-built test cases is specified.
+    if (LINUX_BIN_DIR)
+        # Search the cmake file that includes the compiler infomation
+        # in the auto-generated file from pre-built cmake directories.
+        set(target_path ${LINUX_BIN_DIR}/../CMakeFiles)
```
CodeMonkeyLeet: This is an unsafe directory traversal; it's based on an assumption of what the CI/CD pipeline copies and sets the `LINUX_BIN_DIR` today that can easily be broken. 

Consider an explicit contract between the test enclave build process and the test host build process instead:
- tests/libcxx/enc/CMakeLists.txt generates a record of its compiler configuration to disk
- tests/libcxx/CMakeLists.txt probes for the record given its relative path under the `LINUX_BIN_DIR` and loads the values if found.

tests/libcxx/CMakeLists.txt
```diff
@@ -20,7 +20,8 @@ foreach(testcase ${alltests})
 # the allocations are entirely optimized out by Clang in these tests and are excluded from Clang release builds -- Skip running these tests
     if ("${name}" MATCHES "cons_default_throws_bad_alloc.pass" OR "${name}" MATCHES "allocator_allocator.members_construct.pass")
         string(TOUPPER ${CMAKE_BUILD_TYPE} BUILD_TYPE_UPPER)
-        if (CMAKE_CXX_COMPILER_ID MATCHES Clang AND BUILD_TYPE_UPPER MATCHES REL)
+        if ((CMAKE_CXX_COMPILER_ID MATCHES Clang OR ENCLAVE_TESTS_CXX_COMPILER_ID MATCHES Clang)
```
CodeMonkeyLeet: Consider consolidating the check into a single variable:
```cmake
if (ADD_WINDOWS_ENCLAVE_TESTS)
    # Load target LINUX_CXX_COMPILER_ID and LINUX_CXX_COMPILER_VERSION values
    set(ENCLAVE_CXX_COMPILER_ID ${LINUX_CXX_COMPILER_ID})
    set(ENCLAVE_CXX_COMPILER_VERSION ${LINUX_CXX_COMPILER_VERSION})
else()
    set(ENCLAVE_CXX_COMPILER_ID ${CMAKE_CXX_COMPILER_ID})
    set(ENCLAVE_CXX_COMPILER_VERSION ${CMAKE_CXX_COMPILER_VERSION})
endif()

...

if (ENCLAVE_CXX_COMPILER_ID MATCHES Clang)
...
```

tests/libcxx/enc/CMakeLists.txt
```diff
@@ -92,31 +93,35 @@ foreach(testcase ${alltests})
...
                AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7
                AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.5)
+               OR (ENCLAVE_TESTS_CXX_COMPILER_ID MATCHES GNU
+               AND ENCLAVE_TESTS_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7
+               AND ENCLAVE_TESTS_CXX_COMPILER_VERSION VERSION_LESS 7.5))
             continue()
         endif()
     endif()
 
 # the following test fails on GCC 7 or newer, see #1559 -- Skip this test on GCC 7.0 and above in non debug builds
```
nonpolarity: Do you know here whether "non debug builds" includes CMAKE_BUILD_TYPE="RelWithDebebInfo"?

mingweishih:
bors try

tests/libcxx/enc/CMakeLists.txt
mingweishih: You are correct, undo the modifications.

cmake/get_compiler_info.cmake
mingweishih: Updated based on your suggestion.

tests/libcxx/CMakeLists.txt
mingweishih: Update based on your suggestion.

tests/libcxx/CMakeLists.txt
```diff
@@ -12,6 +12,20 @@ add_subdirectory(host)
...
 if (BUILD_ENCLAVES)
     add_subdirectory(enc)
+elseif (LINUX_BIN_DIR)
+    # Find the build information from pre-compiled tests. If found, `LINUX_CXX_COMPILER_ID`
+    # and `LINUX_CXX_COMPILER_VERSION` will hold the corresponding information.
+    list(APPEND CMAKE_MODULE_PATH "${LINUX_BIN_DIR}/libcxx/enc")
+    include(build_info OPTIONAL)
+endif()
+
+if (ADD_WINDOWS_ENCLAVE_TESTS AND LINUX_CXX_COMPILER_ID AND LINUX_CXX_COMPILER_VERSION)
```
CodeMonkeyLeet: Consider nesting the `LINUX_CXX_COMPILER_ID` and `LINUX_CXX_COMPILER_VERSION` checks under the `if (ADD_WINDOWS_ENCLAVE_TESTS)` case and explicitly erroring out if either are unset. This allows better trapping and diagnosis of an unexpected case than letting it fall through to using the CMAKE compiler properties and failing that way.

tests/libcxx/enc/CMakeLists.txt
```diff
@@ -35,6 +35,12 @@ target_include_directories(libcxxtest-support PRIVATE
         ${CMAKE_CURRENT_BINARY_DIR}
         )
 
+# Log the build information.
+set(BUILD_INFO_FILE "${CMAKE_CURRENT_BINARY_DIR}/build_info.cmake")
+file(WRITE "${BUILD_INFO_FILE}" "")
+file(APPEND "${BUILD_INFO_FILE}" "set(LINUX_CXX_COMPILER_ID \"${CMAKE_CXX_COMPILER_ID}\")\n")
+file(APPEND "${BUILD_INFO_FILE}" "set(LINUX_CXX_COMPILER_VERSION \"${CMAKE_CXX_COMPILER_VERSION}\")")
```
CodeMonkeyLeet: Should probably add a `\n` newline character to conform to linux expectations that text files are terminated with a newline.

tests/libcxx/enc/CMakeLists.txt
```diff
@@ -35,6 +35,12 @@ target_include_directories(libcxxtest-support PRIVATE
         ${CMAKE_CURRENT_BINARY_DIR}
         )
 
+# Log the build information.
+set(BUILD_INFO_FILE "${CMAKE_CURRENT_BINARY_DIR}/build_info.cmake")
+file(WRITE "${BUILD_INFO_FILE}" "")
+file(APPEND "${BUILD_INFO_FILE}" "set(LINUX_CXX_COMPILER_ID \"${CMAKE_CXX_COMPILER_ID}\")\n")
```
CodeMonkeyLeet: Nonblocking: You could probably save yourself a line and `WRITE` this first line of content into the file instead of writing an empty string then appending.

PR #2436, opened 2020-01-16
Exposed backtrace via GNU functions (execinfo.h)

### LINKED ISSUE #961: Factor out ELF-specific backtrace handling
@CodeMonkeyLeet : Future work: We should file an issue around factoring out the ELF-specific handling of backtrace symbols into windows/Linux source files instead of inline in ocalls.c

Also we need to support backtraces on Windows.

### LINKED ISSUE #2421: Debugging: Expose backtrace functionality for users
OE currently supports oe_backtrace and oe_backtrace_symbols modeled after the GNU functions (backtrace and backtrace_symbols). oe_backtrace is used internally in debug builds to report memory leaks.

Backtrace functionality is in general useful feature for the end user to help debug their systems both in debug builds and production builds.

anakrish:
Compile with -fno-omit-frame-pointer. This allows using rbp
to walk up the stack.

Backtrace functionality is exposed via the GNU functions instead of
oe_ prefixed funtions.

A tricky aspect of the implementation is that backtraces are used
by debug malloc (which is available only in debug builds).
When debug malloc uses backtraces, it must not use (debug) malloc
to allocate memory; instead it must use the lower level dlmalloc.
This is enabled by passing memory management functions to
oe_back_trace_symbols_impl.

To avoid duplication of the backtrace gathering logic, the implementation
is moved into oe_backtrace_impl. It is expected to be called from
oe_backtrace and from backtrace. It removed the frame of the caller
in the gathered backtrace.

Fixes #2421
Fixes #961

### DIFF
--- a/3rdparty/musl/patches/execinfo.h
+++ b/3rdparty/musl/patches/execinfo.h
@@ -1,0 +1,23 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#ifndef _OE_MUSL_PATCHES_EXECINFO_H
+#define _OE_MUSL_PATCHES_EXECINFO_H
+
+#include <openenclave/bits/defs.h>
+
+OE_EXTERNC_BEGIN
+
+// See https://www.gnu.org/software/libc/manual/html_node/Backtraces.html
+// for a description of the GNU backtrace functions.
+
+int backtrace(void** buffer, int size);
+
+char** backtrace_symbols(void* const* buffer, int size);
+
+// This is not implemented yet.
+// void backtrace_symbols_fd(void *const *buffer, int size, int fd);
+
+OE_EXTERNC_END
+
+#endif /* _OE_MUSL_PATCHES_EXECINFO_H */
--- a/enclave/core/CMakeLists.txt
+++ b/enclave/core/CMakeLists.txt
@@ -212,6 +212,8 @@ if (OE_SGX)
         -nostdinc
         -fstack-protector-strong
         -fvisibility=hidden
+        # Preserve frame-pointer in Release mode to enable oe_backtrace.
+        -fno-omit-frame-pointer
         # Put each function or data in its own section.
         # This allows aggressively eliminating unused code.
         -ffunction-sections -fdata-sections
--- a/enclave/core/sgx/backtrace.c
+++ b/enclave/core/sgx/backtrace.c
@@ -8,6 +8,7 @@
 #include <openenclave/internal/backtrace.h>
 #include <openenclave/internal/calls.h>
 #include <openenclave/internal/globals.h>
+#include <openenclave/internal/malloc.h>
 #include <openenclave/internal/print.h>
 #include <openenclave/internal/raise.h>
 #include "sgx_t.h"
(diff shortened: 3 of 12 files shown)


anakrish:
bors try

enclave/core/CMakeLists.txt
```diff
@@ -212,6 +212,8 @@ if (OE_SGX)
         -nostdinc
         -fstack-protector-strong
         -fvisibility=hidden
+        # Preserve frame-pointer in Release mode to enable oe_backtrace.
+        -fno-omit-frame-pointer
```
jhand2: Doesn't all code need to be compiled with this option? Not just liboecore

anakrish:
Fixes #2421

enclave/core/sgx/backtrace.c
```diff
@@ -83,17 +73,20 @@ int oe_backtrace(void** buffer, int size)
...
     return n;
-#else
-    return 0;
-#endif
 }
 
-char** oe_backtrace_symbols(void* const* buffer, int size)
+int oe_backtrace(void** buffer, int size)
+{
+    return oe_backtrace_impl(__builtin_frame_address(0), buffer, size);
```
jhand2: Should this function be marked to always be inlined?

enclave/core/CMakeLists.txt
anakrish: Good point. 
Adding an option here in the PUBLIC compilation options for oecore ensures that all other source files including user written ones get the option if compiled via cmake.

I have updated pkgconfig as well so that the flag is propagated for make file based workflow.

enclave/core/sgx/backtrace.c
anakrish: Yes, this function ought to have the behavior of being inlined so that oe_backtrace itself does not show up in the stack frames. However we don't want to put the implementation in a header file and want to keep it internal.

Instead we pass the oe_backtrace function's frame to oe_backtrace_impl which has the same effect as if we are jumping to oe_backtrace_impl. The logic within oe_backtrace_impl starts the walk at the callee of the passed frame and thus oe_backtrace is excluded from the trace.

Any suggestions for better organizing these two functions' responsibilities welcome.

3rdparty/musl/patches/execinfo.h
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Open Enclave SDK contributors.
```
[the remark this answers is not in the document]
anakrish: No. This is an OE SGX specific implementation that relies on how we compile enclaves.
We rely on just walking up rbp to provide robust backtraces.

A more general implementation would be more complex, and would require dwarf interpreters.
E.g: https://github.com/ianlancetaylor/libbacktrace

anakrish:
bors try

anakrish:
bors try

3rdparty/musl/patches/execinfo.h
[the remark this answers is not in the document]
anakrish: @CodeMonkeyLeet I have updated CHANGELOG.md and LibcSupport.md. Thanks for pointing it out.

> I don't believe we have precedence for modifying the header directly. 

We are not modifying headers in this case; MUSL does not provide `execinfo.h`  (likely due to the complexity of providing a general purpose implementation).

anakrish:
bors try

anakrish:
bors try

PR #2439, opened 2020-01-17
Upgrade to mbedTLS v2.16.4

CodeMonkeyLeet:
Note for reviewers:
* dd2ba3a05f18de996dc102f7e65d58d660afbb22 is a straight clone of v2.16.4 and update to update.make
* 2003368498e520cd22da435e39cc6c98f57dc117 is what might require review and is simply a reapplication of the attested TLS patch and and update to CHANGELOG.md


jhand2:
Does the `0001-Patch-x509write_crt.c-for-attestedTLS.patch` file apply cleanly (with `git am`  or `git apply`) or did you have to do manual fixes?

In general I wonder if it is better to apply patches at build time rather than manually. Then doing updates should be more pain-free, especially if there is more than one patch file for a library. Not a question for this PR, just wanted to get your thoughts.

CodeMonkeyLeet:
@jhand2 
> Does the `0001-Patch-x509write_crt.c-for-attestedTLS.patch` file apply cleanly (with `git am` or `git apply`) or did you have to do manual fixes?

No, it doesn't; it always needs to be applied as a 3-way merge otherwise, and usually there's a manual conflict to resolve because the baseline is always effectively different from how the patch was created.

PR #2442, opened 2020-01-21
Add -fstack-protector-strong to the CFLAG for pkgconfig.

nonpolarity:
For SDK's end users, this option is already turned on for enclave through cmake configuration /opt/openenclave/lib/openenclave/cmake/openenclave-targets.cmake. This change will make -fstack-protector-strong as a default option in case end users use pkgconfig to setup the environment.


nonpolarity:
bors try

jhand2:
Thanks!

bors r+

PR #2451, opened 2020-01-23
Enable mbedtls tests on Windows

### LINKED ISSUE #1893: With MBEDTLS_NET_C component enabled, need to make sure all the relevant mbedtls test code are fully enabled.
mbedtls has several test programs or samples in various directories. The options enabled are listed in config.h. With MBEDTLS_NET_C enabled, networking I/O routines are enabled in enclaves and hence several tests can now be enabled.

jhand2:
Recently, POSIX sockets support was added to Open Enclave for Windows hosts.
Enable the mbedtls tests to be compiled and run on Windows.

Fixes #1893


jhand2:
bors try

tests/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ add_subdirectory(tools)
 if (OE_SGX)
     add_subdirectory(debugger)
     add_subdirectory(host_verify)
+    add_subdirectory(mbed)
```
mingweishih: Does this modification mean to enable tests in `WIN32_SIMULATION` mode? I think this is the only missing configuration before this change (not sure we have that config in CI/CD).

jhand2: I think this flag is left over and should probably be removed. There is no such thing as Windows simulation mode.

mingweishih: Then, I think the mbedtls tests are already enabled (based on the checks `ADD_WINDOWS_ENCLAVE_TESTS OR USE_CLANGW`). Can you double check? To avoid confusion, may we can have another PR to remove this flag and re-org the logic in this file.

jhand2: Oh you're right, they do already run. I'll close this. Thanks!

PR #2455, opened 2020-01-23
Add openssl engine support to oesign to allow signing keys via engines

### LINKED ISSUE #2090: Support crypto extensions in oesign tool
To accommodate a wider range of key management scenarios when using oesign, it would be valuable to take advantage of existing crypto library infrastructure for interfacing with Hardware Security Modules (HSM) and other custom crypto providers. For example:

- OpenSSL on Linux supports [ENGINE](https://github.com/openssl/openssl/blob/master/README.ENGINE) plug-ins 
- BCrypt on Windows supports custom [Cryptographic Providers](https://www.microsoft.com/en-us/download/details.aspx?id=30688) such as key storage providers (KSP)

In addition to enabling the use of custom hardware for acceleration and key storage, this mechanism can also be used to accommodate developers who want to do local secure caching of keys and signing operations inside an enclave (#2042), where the custom provider can be tailored to also account for the attestation endpoint, authentication, sealing and key escrow behaviors on a per enclave basis. 

This task requires a command line interface change to support pass-through of parameters to the engines (see OpenSSL's implementation).

yakman2020:
This is a requirement for CPDx integration, but a good idea generally.

If we use the esrp engine we can get keys via a protected service rather than pem files.  The private key provided by the engine does provide the necessary information to derive a public key, but does not compromise the private key bits themselves.

Fixes #2090


yakman2020:
addresses issue #2457

host/crypto/openssl/rsa.c
```diff
@@ -344,6 +359,7 @@ oe_result_t oe_rsa_public_key_equal(
         (oe_public_key_t*)public_key1, (oe_public_key_t*)public_key2, equal);
 }
 
+
```
jhand2: nit: extra newline

host/sgx/sgxsign.c
```diff
@@ -434,6 +434,8 @@ static oe_result_t _get_q1_and_q2(
 }
 #endif
 
+
+
```
jhand2: nit: extra newlines

include/openenclave/internal/rsa.h
```diff
@@ -117,6 +117,27 @@ oe_result_t oe_rsa_public_key_write_pem(
...
+ *
+ * @return OE_OK upon success
+ */
+oe_result_t oe_rsa_private_key_from_engine(
+    oe_rsa_private_key_t* private_key,
+    const char *engine_id,
+    const char *engine_load_path,
+    const char *key_id);
+
+
```
jhand2: nit: extra newline both above and below

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+## Engines 
+
+The concept of an engine was added to OpenSSL starting with 0.9.6. There are built in engine implementations
+implementations for
+- Microsoft CryptoAPI
+- VIA Padlock
+- nCipher CHIL
+and a api for controlling dynamically bound engines which are bound from shared libraries at runtime via either program logic
```
dthaler: s/a api/an API/

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,114 @@
...
+```
+# openssl engine
+(rdrand) Intel RDRAND engine
+(dynamic) Dynamic engine loading support
+
+```
+Of particular interest is the "dynamic" engine, which allow dynamic plugins to perform encryption duties. 
+These objects allow a plugin mechanism for implementations of cryptographic algorithms and can be statically linked in the
+OpenSSL library at compile-time or dynamically loaded at run-time in and out of the running application 
+with low overhead from external binary shared objects implementing the engine API using special built-in engine
```
dthaler: s/using special engine called "dynamic"/using the dynamic engine/

Rationale: (1) missing article.  (2) "dynamic" engine was already introduced in the first sentence of the paragraph

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,114 @@
...
+called “dynamic”. 
+
+Dynamic engines are particularly interesting due to the flexibility they provide:
```
dthaler: Here has dynamic "engines" plural but line 32 uses dynamic engine singular with dynamic "plugins" plural.  Should this say "Dynamic plugins are..."?

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,114 @@
...
+
+- They reduce the memory impact of OpenSSL, by avoiding statically linking support for unneeded hardware or features at
+  compile-time in favor of system configuration or automatically probing for supported devices at run-time and dynamically
+  loading only the required cryptographic modules.
+
+- They allow the developer or system administrator to replace compiled in OpenSSL functionality in case of bugs, 
+  known vulnerabilities or sub-optimal performances with more appropriate alternative implementations while maintaining compatibility
+  with existing applications.
+
+- They provide an alternative in case of issues with the OpenSSL core development team or distro vendor decision-making process, decoupling
```
dthaler: What's the relationship between openssl and mbedtls?  Do we really need to have the OE SDK depend on *both* for different purposes, instead of only one and making that one be replacable by the other?

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,114 @@
...
+  the decision to adopt the OpenSSL library from the choice of individual cryptosystem implementations, without incurring the 
+  costs of maintaining a fork of the OpenSSL project or patchsets, while also providing transparent binary compatibility with existing applications; 
+
+- They allow hardware vendors to release self-contained plugins to leverage advanced hardware concepts, such as GPU compute, FPLAs, or
+  quantum based encrytion which can work then with existing applications based on OpenSSL, keeping their software outside
+ of the main OpenSSL codebase.
+
+- They allow backporting newer crypto systems in previous versions of the OpenSSL library and existing applications based on it.
+
+- They allow a straightforward method of protyping and deploying new crypto systems or new implementations for already compiled-in
```
dthaler: typo "protyping"

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,114 @@
...
+  costs of maintaining a fork of the OpenSSL project or patchsets, while also providing transparent binary compatibility with existing applications; 
+
+- They allow hardware vendors to release self-contained plugins to leverage advanced hardware concepts, such as GPU compute, FPLAs, or
+  quantum based encrytion which can work then with existing applications based on OpenSSL, keeping their software outside
+ of the main OpenSSL codebase.
+
+- They allow backporting newer crypto systems in previous versions of the OpenSSL library and existing applications based on it.
+
+- They allow a straightforward method of protyping and deploying new crypto systems or new implementations for already compiled-in 
+  cryptosystems to the OpenSSL library, providing convenient way to test and benchmark new software implementations in a real-world context.
```
dthaler: grammar: s/providing convenient/providing a convenient/

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,114 @@
...
+  implementations or new functionality.
+
+- They offer flexibility to solve licensing issues. 
+
+  Currently the OpenSSL project is released under a "dual license" scheme, under the OpenSSL License,
+  (a derivative of the Apache License 1.0) and the SSLeay License (similar to a 4-clause BSD License). OpenSSL is in the process of 
+  transitioning to the Apache License 2.0. Contributors are thus forced to release their work under these licenses, which may conflict
+  with other applicable licenses especially when reusing code from projects released under a proprietary license or an incompatible copyleft license. 
+
+  Being objects dynamically loaded at runtime, engines can benefit from usually more flexible licensing requirements, providing bridge
```
dthaler: grammar: s/providing bridge/providing a bridge/

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,114 @@
...
+## Engine internal operation
+
+The cryptographic functionality that can be provided by an engine implementation includes the following abstractions:
+(1) RSA_METHOD, DSA_METHOD, DH_METHOD, EC_METHOD; providing alternative RSA/DSA/etc. implementations; 
+(2) RAND_METHOD: providing alternative (pseudo-)random number generation implementations;
+(3) EVP_CIPHER: providing alternative (symmetric) cipher algorithms;
+(4) EVP_MD: providing alternative message digest algorithms;
+(5) EVP_PKEY: providing alternative public key algorithms.
+
+At the highest level of abstraction, a dynamic engine can be split into two functional blocks.
```
dthaler: Should "dynamic engine" be "dynamic plugin"?

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,114 @@
...
+
+One block contains all the alternative implementations for the cryptosystems provided by the engine. This part mainly consists of a
+collection of structs for each cryptosystem, each linking to the actual functions implementing its operations.
+
+For example, an EVP_MD message digest struct would reference the actual init(), update(), and final() functions implementing
+the OpenSSL message digest streaming API, in addition to some utility functions allowing the OpenSSL library to cleanly handle, 
+clone and destroy instances of the provided message digest implementation. Every such struct would be individually registered against
+the OpenSSL library during the bind() process, and structs of the same kind (e.g. all the EVP_MD structs, all the EVP_PKEY_meth structs, etc.)
```
dthaler: grammar: always put a comma after "e.g."

host/crypto/bcrypt/rsa.c
```diff
@@ -175,6 +175,19 @@ void oe_rsa_public_key_init(
...
 
+
+oe_result_t oe_rsa_private_key_from_engine(
+    oe_rsa_private_key_t* private_key,
+    const char *engine_id,
+    const char *engine_load_path,
+    const char *key_id)
+{
+    /*
+     * bcrypt does not support engines, so nothing to do.
```
dthaler: should "engines" be "plugins"  (this goes back to the terminology question of whether it's 1 engine with multiple plugins, or multiple engines)

host/crypto/openssl/key.c
```diff
@@ -41,6 +42,85 @@ void oe_private_key_init(
...
+oe_result_t oe_private_key_from_engine(
+    const char *engine_id,
+    const char *engine_load_path,
+    const char *key_id,
+    oe_private_key_t* key,
+    int key_type,
+    uint64_t magic)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    oe_private_key_t* impl = (oe_private_key_t*)key;
```
dthaler: why is a cast here?  key and impl are both defined to be of the same type.

host/crypto/openssl/key.c
```diff
@@ -399,6 +480,125 @@ oe_result_t oe_private_key_sign(
...
+    const char *engine_load_path, // Location of the 
+    oe_hash_type_t hash_type,
+    const void* hash_data,
+    size_t hash_size,
+    uint8_t* signature,
+    size_t* signature_size)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    EVP_PKEY_CTX* ctx = NULL;
+    EVP_PKEY *pkey    = NULL;
```
dthaler: nit: inconsistent spacing before =

host/crypto/openssl/key.c
```diff
@@ -399,6 +480,125 @@ oe_result_t oe_private_key_sign(
...
+        OE_RAISE(OE_CRYPTO_ERROR);
+
+    if (!ENGINE_init(e))
+    {
+        /* if ENGINE_init failed there is no valid reference to e */
+        e = NULL;  
+        OE_RAISE(OE_CRYPTO_ERROR);
+    }
+
+    /* Then get the private key 2do: try to sign using the engine rather
```
dthaler: what does "2do" mean?   Is this a "TODO"?   Be explicit to make todos easier to find.
Even better would be to file a github issue and mention the # in the comment.

host/crypto/openssl/key.c
```diff
@@ -399,6 +480,125 @@ oe_result_t oe_private_key_sign(
...
+       than the private key, if possible. */
+
+    if (!(pkey = ENGINE_load_private_key(e, key_id, NULL, NULL)))
+        OE_RAISE(OE_CRYPTO_ERROR);
+
+    /* Create signing context */
+    if (!(ctx = EVP_PKEY_CTX_new(pkey, e)))
+    {
+        /* some engines are not capable of performing the signing operation, but some are. 
+           If we can, we should use them, but its not an error if we can't */
```
dthaler: s/its/it's/

host/crypto/openssl/key.c
```diff
@@ -399,6 +480,125 @@ oe_result_t oe_private_key_sign(
...
+        0)
+        OE_RAISE(OE_CRYPTO_ERROR);
+
+    result = OE_OK;
+
+done:
+    if (e)
+        ENGINE_finish(e);
+
+
```
dthaler: nit: can remove extra blank line.  one is sufficient.

host/sgx/sgxsign.c
```diff
@@ -585,6 +587,45 @@ static oe_result_t _init_sigstruct(
...
+    const char *engine_id,
+    const char *engine_load_path,
+    const char *key_id,
+    sgx_sigstruct_t* sigstruct)
+{
+    oe_rsa_private_key_t rsa;
+    bool rsa_initalized = false;
+    oe_result_t result = OE_UNEXPECTED;
+
+    if (sigstruct)
```
dthaler: You could move the memset down to line 610 and then you wouldn't need the IF test.  (The difference is just that if one of the other arguments is null and we fail, it won't be zeroed, but I think that's ok and more readable.)

include/openenclave/internal/sgxsign.h
```diff
@@ -35,6 +35,31 @@ oe_result_t oe_sgx_sign_enclave(
...
+ * Digitally sign the enclave with the given hash using an openssl engine
+ *
+ * This function digitally signs the enclave whose hash is given by the
+ * **mrenclave** parameter. The signing key is given by the **engine_id** and related **key_id**
+ * parameter. If successful, the function writes the signature into the
+ * **sigstruct** parameter (an SGX signature structure).
+ *
+ * @param mrenclave[in] hash of the enclave being signed
+ * @param engine_id[in] Text name of the engine to use
+ * @param key_id[in]    Integer handle for the key to use.
```
dthaler: nit: Inconsistent spacing and capitalization between 48 vs 49

include/openenclave/internal/sgxsign.h
```diff
@@ -35,6 +35,31 @@ oe_result_t oe_sgx_sign_enclave(
...
+ * @return OE_OK success
+ */
+oe_result_t oe_sgx_sign_enclave_from_engine(
+    const OE_SHA256* mrenclave,
+    uint64_t attributes,
+    uint16_t product_id,
+    uint16_t security_version,
+    const char *engine_id,
+    const char *engine_load_path,
+    const char * key_id,
```
dthaler: nit: inconsistant spacing around *.  Compare line 54, line 59, and 60, which all use different styles.  Pick one.   (Personally I would be in favor of using something like uncrustify to do such things automatically)

tests/tools/CMakeLists.txt
```diff
@@ -4,4 +4,7 @@
 if (OE_SGX)
     add_subdirectory(oecert)
     add_subdirectory(oecertdump)
+if (NOT WIN32)
```
dthaler: Can you add a # comment saying why this is in a NOT WIN32 condition?

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
```diff
@@ -0,0 +1,275 @@
...
+#include <openssl/engine.h>
+#include <openssl/types.h>
+#include <openssl/bio.h>
+#include <openssl/pem.h>
+#include <sys/stat.h>
+
+static const char *test_engine_id   = "oesign-test-engine";
+static const char *test_engine_name = "oesign test engine";
+
+#if 0
```
dthaler: why is there an #if 0?
Recommend deleting code if it's never used.

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
```diff
@@ -0,0 +1,275 @@
...
+    {
+        struct stat st;
+
+        if (stat(path, &st) != 0)
+            goto done;
+
+        *size = (size_t)st.st_size;
+    }
+
+    /* Allocate memory. We add 1 to null terimate the file since the crypto
```
dthaler: typo "terimate"

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
```diff
@@ -0,0 +1,275 @@
...
+ * We pass the name of the pem file as the keyid, so that we can actually use the signing key from the openenclave build 
+ * as a signing key for oesign from the test engine. We know the test went well when we can load a test enclave
+ */
+static EVP_PKEY *get_private_key_from_keyid(const char *keyid)
+{
+     uint8_t *pem_data = NULL;
+     size_t  pem_size  = 0;
+     EVP_PKEY* pkey = NULL; 
+
+     const char *path = keyid;  // Do we wanty to directly do that or have an environment variable to direct to a path for pem files?
```
dthaler: typo "wanty"

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
```diff
@@ -0,0 +1,275 @@
...
+{ 
+    (void)e;
+    printf("test_engine_destroy\n");
+    //2do:test_engine_param_free();
+    EVP_PKEY_free(test_engine_only_privkey);
+    EVP_PKEY_free(test_engine_only_pubkey);
+    return 1;
+}
+
+#if 0
```
dthaler: Why is there code inside a #if 0 ?
Can it be deleted?

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
```diff
@@ -0,0 +1,275 @@
...
+    if (!ENGINE_set_id(e, test_engine_id)) {
+        fprintf(stderr, "ENGINE_set_id failed\n");
+        goto done;
+    }
+    if (!ENGINE_set_name(e, test_engine_name)) {
+        printf("ENGINE_set_name failed\n");
+        goto done;
+    }
+
+#if 0
```
dthaler: Delete code?

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
```diff
@@ -0,0 +1,275 @@
...
+    }
+#endif
+
+    if (!ENGINE_set_destroy_function(e, test_engine_destroy)
+        || !ENGINE_set_init_function(e, test_engine_init)
+        || !ENGINE_set_finish_function(e, test_engine_finish)) {
+        goto done;
+    }
+
+
```
dthaler: nit: one blank line will suffice

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
```diff
@@ -0,0 +1,275 @@
...
+    ret = 1;
+
+    printf("zoot id=%p\n", id);
+    printf("zoot id=%s\n", id);
```
dthaler: output looks odd.  Two visually different values with same label?  If this is still needed, improve the labels.

tools/oesign/main.c
```diff
@@ -499,12 +508,15 @@ static const char _usage_sign[] =
...
     "\n"
-    "    The key is read from KEY_FILE and contains a private RSA key in PEM\n"
+    "    If specified, the key read from KEY_FILE and contains a private RSA key in PEM\n"
     "    format. The keyfile must contain the following header.\n"
     "\n"
     "        -----BEGIN RSA PRIVATE KEY-----\n"
     "\n"
     "    The resulting image is written to ENCLAVE_IMAGE.signed\n"
+    "\n"
+    " Keys may also be received from an openssl engine. specified by the string ENGINE-NAME\n"
```
dthaler: s/engine. specified/engine specified/

tools/oesign/main.c
```diff
@@ -731,6 +782,18 @@ int sign_parser(int argc, const char* argv[])
             case 'k':
                 keyfile = optarg;
                 break;
+            case 'n':
+printf("engine = %s\n", optarg);
```
dthaler: remove debugging printfs

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: There is no engine facility in mbedtls, so that reduces the question to do we need mbedtls. 

That seems an interesting question but outside the scope of this PR, no?

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: There is an engine named "dynamic" which connect to different provided dynamic engines.
Plugin is an adjetive to introduce the idea. Dynamic engine(s) are the term of art.

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: Dynamic engines is the term of art for .so engines which are dynalinked by the engine named "dynamic".
I'm not thrilled with the terminology either, but if I try to define new terms it will confuse folks who are used to the openssl documentation.

Proposed revision:
Of particular interest is the "dynamic" engine, which allow dynamic plugins (dynamic engines) to perform encryption duties.
These objects allow a plugin mechanism for implementations of cryptographic algorithms and can be statically linked in the
OpenSSL library at compile-time or dynamically loaded at run-time in and out of the running application
with low overhead from external binary shared objects implementing the engine API using special built-in engine
called “dynamic” to perform the loading.

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: fixed. Will resolve when I push.

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
yakman2020: done. Will resolve on push

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
yakman2020: done.  Will resolve on push

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
yakman2020: debug printfs got past. Sorry

tools/oesign/main.c
yakman2020: done. Will resolve when I push.

tools/oesign/main.c
yakman2020: done. Will resolve when I push.

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: done. Will resolve when I push.

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
dthaler: I agree the question is technically on the design doc for this and not the code PR per se, but I didn't see a design doc PR to comment on.  If there is one, can you point me to it?  (I looked in https://github.com/openenclave/openenclave/tree/master/docs/DesignDocs but nothing jumped out at me as being it.)

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
dthaler: I now understand you're using "dynamic engine" and "dynamic plugin" as synonyms, instead of plugins being things that plug into an engine.   As such, proposed revision:

Of particular interest are "dynamic" engines, which are plugins that perform encryption duties. These plugins for implementations of cryptographic algorithms can be statically linked in the
OpenSSL library at compile-time or dynamically loaded at run-time in and out of the running application with low overhead from external binary shared objects implementing the engine API.

host/crypto/openssl/key.c
yakman2020: I rephrased, though it seems a reasonable thing to say.
What do I use to indicate that something could be better but the surrounding environment does not yet enable it, so its not currently practical?

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: I do not believe dynamic engines are ever statically linked,

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: done

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: dynamic engine is the correct term and what is used in openssl documentation. Plugin is not a synonym but an adjective.

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: done

host/crypto/bcrypt/rsa.c
yakman2020: plugin is not the correct term. If we say plugin it will not correspond to the usage in openssl and would tend to confuse the target audience.  The doc on openssl engines is already garbled enough, so varying terminology is not good.

host/crypto/openssl/key.c
yakman2020: done

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: I don't understand what you are asking here.
Are you suggesting that oesign get completely rewritten to use mbedtls?

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: not, dynamic engine is a term of art, but dynamic plugin is a description, not a synonym. This is necessary because all openssl documentation uses that terminology.

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: done

host/crypto/openssl/key.c
yakman2020: done

host/sgx/sgxsign.c
yakman2020: yup. done

include/openenclave/internal/sgxsign.h
yakman2020: done

include/openenclave/internal/sgxsign.h
yakman2020: done

tests/tools/CMakeLists.txt
yakman2020: done

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
yakman2020: done

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
yakman2020: done

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
yakman2020: done

tests/tools/oesign-engine/test-engine/oesign-test-engine.c
yakman2020: done

cmake/add_enclave.cmake
```diff
@@ -132,10 +136,17 @@ function(add_enclave_sgx)
 
   # Sign the enclave using `oesign`.
   if(ENCLAVE_CONFIG)
-    add_custom_command(OUTPUT ${SIGNED_LOCATION}
-      COMMAND oesign sign -e $<TARGET_FILE:${ENCLAVE_TARGET}> -c ${ENCLAVE_CONFIG} -k ${ENCLAVE_KEY}
-      DEPENDS oesign ${ENCLAVE_TARGET} ${ENCLAVE_CONFIG} ${ENCLAVE_KEY}
-      WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+      if(SIGNING_ENGINE)
+          add_custom_command(OUTPUT ${SIGNED_LOCATION}
+          COMMAND oesign sign -e $<TARGET_FILE:${ENCLAVE_TARGET}> -c ${ENCLAVE_CONFIG} -n ${SIGNING_ENGINE} -p ${ENGINE_LOAD_PATH} -i ${ENGINE_KEY_ID}
```
mikbras: I wonder why they don't put full quotes around these arguments: like "${ENCLAVE_CONFIG}"?

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,111 @@
...
+- They allow the developer or system administrator to replace compiled in OpenSSL functionality in case of bugs,
+  known vulnerabilities or sub-optimal performances with more appropriate alternative implementations while maintaining compatibility
+  with existing applications.
+
+- They provide an alternative in case of issues with the OpenSSL core development team or distro vendor decision-making process, decoupling
+  the decision to adopt the OpenSSL library from the choice of individual cryptosystem implementations, without incurring the
+  costs of maintaining a fork of the OpenSSL project or patchsets, while also providing transparent binary compatibility with existing applications;
+
+- They allow hardware vendors to release self-contained plugins to leverage advanced hardware concepts, such as GPU compute, FPLAs, or
+  quantum based encrytion which can work then with existing applications based on OpenSSL, keeping their software outside
```
mikbras: Typo: encryption

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,111 @@
...
+are glued together by functions registered in the engine object that allow the OpenSSL library to query the engine for lists of
+provided algorithms or a specific algorithm indexed by NID.
+
+The other block contains the bind() method and the initialization and deinitialization functions.
+- The bind() method is called by the OpenSSL built-in dynamic engine upon load and is used to set the internal state of the
+engine object and allocate needed resources, to set its id and name, and the pointers to the init(), finish(), and destroy() functions.
+
+- The init() function is called to derive a fully initialized functional reference to the engine from a structural reference.
+
+- The finish() function is called when releasing an engine functional reference, to freeup any resource allocated to it
```
mikbras: Type: freeup => free up

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
paulcallen: host side uses OpenSSL by default (linux) and enclave uses mbedtls by default. OEsign is a host application and is already coded against OpenSSL. Nothing has changed in this pull request around this. Therefore if you have this question it seems like it should be in a separate issue?

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,111 @@
...
+
+The concept of an engine was added to OpenSSL starting with 0.9.6. There are built in engine implementations
+implementations for
+- Microsoft CryptoAPI
+- VIA Padlock
+- nCipher CHIL
+and a API for controlling dynamically bound engines which are bound from shared libraries at runtime via either program logic
+or configuration files.
+
+The openssl commnd line utility can show available engines via the "engine" command.
```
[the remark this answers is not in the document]
yakman2020: I see it!
Thanks.

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: done

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: done

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
dthaler: I am ok with Paul's answer (host side uses openssl and enclave uses mbedtls) for now, but the meta point is that we need something in the design docs that actually says/explains that, and I don't see it.  I am fine with a separate PR for that.

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: I rewrote the text to try to make it more clear.

tests/tools/oesign-engine/test-enclave/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+# Licensed under the MIT License.
+
+cmake_minimum_required(VERSION 3.11)
+
+project("oesign-test" LANGUAGES C CXX)
+
+set(CMAKE_CXX_STANDARD 11)
+
+add_subdirectory(enclave)
+add_subdirectory(host)
```
CodeMonkeyLeet: Can this be simplified greatly? All you really need for this test is to compile the enclave binary and you don't need any of the host pieces here at all, do you?

tests/tools/oesign-engine/test-enclave/oesign-test.py
```diff
@@ -0,0 +1,121 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+#!/usr/bin/env python3
+#  Run tests of oesign.
+#  We need this wrapper to return the correct return value from the tests, since
+#  several invocations of oesign are supposed to fail by design.
```
CodeMonkeyLeet: Not quite sure why this is necessary? You're already parameterizing the inputs to each of the tests, except that you're also choosing to hardcode a value in each one as well (e.g. `oesign_detects_failed_engine_id` when `engine_id` is already an input parameter that's ignored by the test)

It seems like you could create a single function that consolidates all of the variants and add a input parameter for expected success/failure, then it'd be fully parametrized and it wouldn't require parsing out separate definitions.

tests/tools/oesign-engine/test-engine/README.md
```diff
@@ -0,0 +1,9 @@
+# Oesign Test Engine
+
+The oesign test engine allows openenclave to validate openssl engine functionality
```
CodeMonkeyLeet: Branding: Open Enclave rather than "openenclave", similarly OpenSSL when referring to the product rather than `openssl` used on the command line.

tools/oesign/main.c
```diff
@@ -468,11 +474,16 @@ static const char _usage_gen[] =
 static const char _usage_sign[] =
     "Usage: %s sign {--enclave-image | -e} ENCLAVE_IMAGE "
     "{--config-file | -c} CONFIG_FILE {--key-file | -k} KEY_FILE\n"
+    "{--engine ENGINE-NAME -load-path ENGINE-LOAD-PATH-key KEY-ID }\n"
     "\n"
     "Where:\n"
     "    ENCLAVE_IMAGE -- path of an enclave image file\n"
     "    CONFIG_FILE -- configuration file containing enclave properties\n"
     "    KEY_FILE -- private key file used to digitally sign the image\n"
+    "    ENGINE-NAME -- text name of the engine to use, for example 'pkcs-11'"
```
CodeMonkeyLeet: `\n` needed here and at the end of each of the ENGINE-LOADPATH and KEY-ID descriptions.

tools/oesign/main.c
```diff
@@ -468,11 +474,16 @@ static const char _usage_gen[] =
 static const char _usage_sign[] =
     "Usage: %s sign {--enclave-image | -e} ENCLAVE_IMAGE "
     "{--config-file | -c} CONFIG_FILE {--key-file | -k} KEY_FILE\n"
+    "{--engine ENGINE-NAME -load-path ENGINE-LOAD-PATH-key KEY-ID }\n"
```
CodeMonkeyLeet: `--load-path` and `--key` for consistency with parameter naming. The parser also appears to support short versions of the parameter names (`-e`, `-l` and `-i` respectively) but that's not reflected here. 

I'm curious why we decided to accept the load path on the command line, it seems like the approach most users would be familiar with would be as done through the openssl commands (e.g. `openssl dgst -engine "myengine"` where `-inkey` is overloaded to accept a key ID).

CodeMonkeyLeet: Also consider if it might be cleaner if KEY_FILE could refer to a key configuration file that encapsulated the engine and key ID parameters. It avoids additional logic to manage the mutual exclusion checking and that multiple parameters must be specified with `--engine`. We already have a parser for name value pairs via CONFIG_FILE after all.

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,113 @@
+# Signing a Build With OpenSSL Engines
```
CodeMonkeyLeet: Thanks for writing this up, Bruce, it's pretty educational.

A lot of it is unrelated to OE SDK though (e.g. the discussion about the default set of engines could just be linked to OpenSSL's README.ENGINE, which looks like where you go it from anyway); it's unclear to me if we're trying to communicate broader support for OpenSSL Engines than we've actually decided to expose through the oesign tool. 

What are we trying to communicate here beyond the usage command?

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+## Engines
+
+The concept of an engine was added to OpenSSL starting with 0.9.6. There are built in engine implementations
+implementations for
+- Microsoft CryptoAPI
+- VIA Padlock
+- nCipher CHIL
+and a API for controlling dynamically bound engines which are bound from shared libraries at runtime via either program logic
```
dthaler: nit: still need to change "a API" to "an API"

include/openenclave/internal/sgxsign.h
```diff
@@ -35,6 +35,32 @@ oe_result_t oe_sgx_sign_enclave(
...
+ * This function digitally signs the enclave whose hash is given by the
+ * **mrenclave** parameter. The signing key is given by the **engine_id** and
+ * related **key_id** parameter. If successful, the function writes the
+ * signature into the
+ * **sigstruct** parameter (an SGX signature structure).
+ *
+ * @param mrenclave[in] hash of the enclave being signed
+ * @param engine_id[in] text name of the engine to use
+ * @param key_id[in]    integer handle for the key to use.
+ * @param sigstruct[out] the SGX signature
```
dthaler: nit: inconsistent spacing.  Either put one space after the ] in 49 or else make all the descriptions line up where line 50 doesn't currently.

tools/oesign/main.c
```diff
@@ -589,27 +611,49 @@ int oesign(const char* enclave, const char* conffile, const char* keyfile)
...
-            result);
-        goto done;
+        /* Load private key into memory */
+        if (_load_pem_file(keyfile, &pem_data, &pem_size) != 0)
+        {
+            Err("Failed to load file: %s", keyfile);
+            goto done;
+        }
+
+        /* Initialize the SigStruct object */
```
dthaler: nit: I think it's actually "sigstruct" and not "SigStruct" isn't it?

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: Shades of dickens!

host/crypto/openssl/key.c
```diff
@@ -399,6 +472,130 @@ oe_result_t oe_private_key_sign(
...
+    /* Load and init the engine. */
+    if (engine_load_path)
+    {
+        /* If the engine load path is set this is a dynamically loaded engine,
+         * not built in */
+        ENGINE_load_dynamic();
+        ENGINE* dyne = ENGINE_by_id("dynamic");
+
+        if (!ENGINE_ctrl_cmd_string(dyne, "ID", "esrp", 0))
```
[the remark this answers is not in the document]
yakman2020: Good catch. Checking it.

host/crypto/rsa.h
```diff
@@ -11,4 +11,10 @@ oe_result_t oe_rsa_get_public_key_from_private(
     const oe_rsa_private_key_t* private_key,
     oe_rsa_public_key_t* public_key);
 
+/* Caller is responsible for freeing public key. */
+oe_result_t oe_rsa_get_public_key_from_engine(
```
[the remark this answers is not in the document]
yakman2020: It is. I removed it.
At one time I was going to try to get the public key from the engine rather than private key, but that turned out not to be necessary as the private key from esrp contains adequate data to generate a public key.

tests/tools/oesign-engine/test-enclave/enclave/Makefile
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) Open Enclave SDK contributors.
```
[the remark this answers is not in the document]
yakman2020: done.

tests/tools/oesign-engine/test-enclave/CMakeLists.txt
yakman2020: We run the host side in the test in order to verify the signing process by loading the enclave.

tests/tools/oesign-engine/test-enclave/oesign-test.py
yakman2020: The standardised parameter list allows me to put those functions in a dict.  
If I modified the functions to take specific parameters or to set variables or whatever they would actually result in less declarative, longer code.

tests/tools/oesign-engine/test-engine/README.md
yakman2020: done

host/crypto/openssl/key.c
```diff
@@ -399,6 +472,130 @@ oe_result_t oe_private_key_sign(
     return result;
 }
 
+oe_result_t oe_sign_by_engine(
```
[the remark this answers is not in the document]
yakman2020: I'm not sure I agree that no example is better than a specific example, but I removed them.

tools/oesign/main.c
yakman2020: added

tools/oesign/main.c
yakman2020: Loadpath is a standard parameter for connecting to dynamic engines and required unless the engine is configured into openssl.conf.  The documentation of the subject shows the process using -pre argments. 

The key-id isn't a key, its a string with meaning only to the engine.

Overloading key-file to mean either a pem or a completely different kind of file seems to complicate the semantics.  

Wouldn't the idea of a configuration file to indicate an engine/loadpath/keyid triple would obfuscate what was going on on the command line, as you would not be able to see the engine parameters in the build. In fact you would not see that you were using an engine at all unless you happened to look at the (probably cmake-generated) configuration file.

docs/GettingStartedDocs/Contributors/SignUsingEngines.md
yakman2020: OpenSSL documentation for engines is epically bad.   It uses a lot of vague descriptions with words like "may" or "could" and  leaves one very little practical info to work with.  There are a couple of OK descriptions of engines if you do a sufficiently deep web search and read through all of the options, but basically there is very little good doc on the subject.

The doc was intended to communicate:
1. What is an engine, in specific terms, and what does it do.
2. If someone picking up the SDK wanted to use and engine, what would be realistic to expect it to do.
3. If someone picking up the SDK liked the idea of using an engine for their environment (which I believe is to be encouraged), how would they implement one, though clearly one could only sort of orient peoplem not provide complete directions.

To the end of (3), we also provide an example in the form of oesign-test-engine which doesn't do much but is a complete, functioning engine.

Again, there is no good doc on the subject on the net.

include/openenclave/internal/sgxsign.h
yakman2020: done.
At one point the text lined up but scripts/format-code "fixed" it. The only safe option seems to be one space, though I don't think that is very readable (IMO). But your point is taken that it needs to be one or the other.

tools/oesign/main.c
yakman2020: done

tools/oesign/main.c
yakman2020: @CodeMonkeyLeet  how would you deduce that the key-file parameter was in fact a key, or a config file containing the requisite triple of engineid/keyid/engineloadpath?  Would it be restricted to .pem files, with anything else treated as an engine config file?

I presume then that all oesign using cmake file would have to be modified to create the config file.

Wouldn't that impose a limitation on oesign that isn't there currently?  I do understand that the code as written prefers pem files, but they don't have to be named *.pem

mikbras: > Also consider if it might be cleaner if KEY_FILE could refer to a key configuration file that encapsulated the engine and key ID parameters. It avoids additional logic to manage the mutual exclusion checking and that multiple parameters must be specified with --engine. We already have a parser for name value pairs via CONFIG_FILE after all.

I think this makes embedding the tool into another script cumbersome. With engine parameters on the command line, the wrapper script can simply invoke **oesign** with the correct parameters. But with the engine configuration file, the wrapper script now needs to generate a file, manage the location of the file, invoke **oesign** and then delete the file. I really think it would be more convenient to achieve the same effect with command-line options if possible.

yakman2020: fixed help text as requested.

host/crypto/openssl/key.h
```diff
@@ -88,4 +96,15 @@ oe_result_t oe_public_key_verify(
     size_t signature_size,
     uint64_t magic);
 
+oe_result_t oe_sign_by_engine(
+    const char* engine_id, // string name of the engine, ie "esrp"
```
CodeMonkeyLeet: Also remove esrp and esrp-specific references from header file.

Comment is also incomplete on line 103

tests/tools/oesign-engine/test-enclave/host/Makefile
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) Open Enclave SDK contributors.
```
CodeMonkeyLeet: This file is completely unused. Please remove.

host/crypto/openssl/key.c
```diff
@@ -41,6 +42,83 @@ void oe_private_key_init(
...
+    /* Check parameters */
+    if (!engine_id || !engine_load_path || !key_id || !impl)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+    /* if the engine is pre-installed we will get a null */
+    ENGINE* e = NULL;
+    if (!(e = ENGINE_by_id(engine_id)))
+    {
+        /* if we got a null we can still load dynaimc. */
+        ENGINE_load_dynamic();
```
CodeMonkeyLeet: Consider forcing dynamic loading if `engine_load_path` is non-null. Dynamic loading as an OpenSSL feature is also used for creating a separate instance of the engine for the caller, and someone who's using this feature may not want to get the default registered instance when they've already explicitly specified the load path (and hence opt-in to dynamic loading)

host/crypto/openssl/key.c
```diff
@@ -41,6 +42,83 @@ void oe_private_key_init(
...
+
+    /* Check parameters */
+    if (!engine_id || !engine_load_path || !key_id || !impl)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+    /* if the engine is pre-installed we will get a null */
+    ENGINE* e = NULL;
+    if (!(e = ENGINE_by_id(engine_id)))
+    {
+        /* if we got a null we can still load dynaimc. */
```
CodeMonkeyLeet: Typo: dynamic

host/crypto/openssl/key.c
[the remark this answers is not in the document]
CodeMonkeyLeet: This entire function is still not invoked anywhere in this changeset. Please remove.

tools/oesign/main.c
```diff
@@ -743,23 +828,39 @@ int sign_parser(int argc, const char* argv[])
...
-    if (conffile == NULL)
+
+#if HAS_ENGINE_SUPPORT
+    if (keyfile && engine_id)
     {
-        Err("Config file flag is missing");
+        Err("only one of keyfile and engine flags are allowed");
+        ret = 1;
+    }
+    if (keyfile == NULL && engine_id == NULL)
```
CodeMonkeyLeet: Additional error cases to trap:
- `if (engine_id != NULL && key_id == NULL)` 
- `if (engine_id == NULL && (key_id != NULL || engine_load_path != NULL))`

It's better to trap these up front where useful messages can be displayed rather than letting them fall through to the methods that expect these conditions but only return `OE_INVALID_PARAMETER`

tools/oesign/main.c
```diff
@@ -499,12 +520,19 @@ static const char _usage_sign[] =
...
     "    format. The keyfile must contain the following header.\n"
     "\n"
     "        -----BEGIN RSA PRIVATE KEY-----\n"
     "\n"
     "    The resulting image is written to ENCLAVE_IMAGE.signed\n"
+    "\n"
+#if HAS_ENGINE_SUPPORT
+    " Keys may also be received from an openssl engine specified by the "
+    "string ENGINE-NAME\n"
+    " If they are received from an engine, KEY-FILE must not be specified. \n"
```
CodeMonkeyLeet: Probably want to specify that KEY_ID must be specified in this case.

tools/oesign/main.c
```diff
@@ -468,11 +480,20 @@ static const char _usage_gen[] =
 static const char _usage_sign[] =
     "Usage: %s sign {--enclave-image | -e} ENCLAVE_IMAGE "
     "{--config-file | -c} CONFIG_FILE {--key-file | -k} KEY_FILE\n"
+#if HAS_ENGINE_SUPPORT
+    "{--engine ENGINE-NAME --load-path ENGINE-LOAD-PATH --key-id KEY-ID }\n"
```
CodeMonkeyLeet: Snake cased argument names for consistency with existing examples: `ENGINE_NAME`, `ENGINE_LOAD_PATH`, `KEY_ID` here and in the rest of the help text.

tests/tools/oesign-engine/test-enclave/oesign-test.py
```diff
@@ -0,0 +1,57 @@
...
+       not args.engine_id or \
+       not args.engine_load_path or \
+       not args.key_id or \
+       not args.oesign_binary:
+
+        arg_parser.print_help()
+        sys.exit ( 1 )
+
+    test_rslt = TEST_PASSED
+    # The keyid will use a slightly bogus key provided by the environment. It should fail.
```
CodeMonkeyLeet: This comment looks out of place since it's just passing the `key_id` from the caller. Please correct or remove.

CodeMonkeyLeet:
bors try

host/crypto/bcrypt/rsa.c
```diff
@@ -175,15 +175,14 @@ void oe_rsa_public_key_init(
...
-    const char *key_id)
+    const char* engine_id,
+    const char* engine_load_path,
+    const char* key_id)
 {
     /*
-     * bcrypt does not support engines, so nothing to do. 
+     * bcrypt does not support engines, so nothing to do.
      */
     return OE_UNEXPECTED;
```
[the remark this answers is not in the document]
yakman2020: done

host/crypto/openssl/key.c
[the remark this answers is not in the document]
yakman2020: done

cmake/add_enclave.cmake
```diff
@@ -74,6 +77,7 @@ endmacro()
 function(add_enclave_sgx)
   set(options CXX)
   set(oneValueArgs TARGET CONFIG KEY)
```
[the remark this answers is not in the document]
yakman2020: done

host/crypto/bcrypt/key.c
```diff
@@ -12,6 +12,24 @@
...
+/* can't use an engine with bcrypt */
+oe_result_t oe_private_key_from_engine(
+    const char* engine_id,
+    const char* engine_load_path,
+    const char* key_id,
+    oe_private_key_t* key,
+    int key_type,
+    uint64_t magic)
+{
+    (void)engine_id;
```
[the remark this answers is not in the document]
yakman2020: done

host/crypto/openssl/key.c
```diff
@@ -399,6 +477,129 @@ oe_result_t oe_private_key_sign(
     return result;
 }
 
+oe_result_t oe_sign_by_engine(
+    const char* engine_id,        // string name of the engine
+    const char* key_id,           // string id of the desired signing key
+    const char* engine_load_path, // Location of the
```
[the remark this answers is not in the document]
yakman2020: function removed

host/crypto/openssl/key.h
yakman2020: removed function decl

tests/tools/oesign-engine/test-enclave/host/Makefile
yakman2020: removed already

tests/tools/oesign-engine/test-enclave/oesign-test.py
yakman2020: removed.

tools/oesign/main.c
yakman2020: done

tools/oesign/main.c
yakman2020: done

host/crypto/openssl/key.c
yakman2020: done

host/crypto/openssl/key.c
yakman2020: done

tools/oesign/main.c
```diff
@@ -743,23 +830,42 @@ int sign_parser(int argc, const char* argv[])
...
+
+#if HAS_ENGINE_SUPPORT
+    if (keyfile && engine_id)
     {
-        Err("Config file flag is missing");
+        Err("only one of keyfile and engine flags are allowed");
         ret = 1;
+        goto done;
+    }
+    if ((keyfile == NULL && engine_id == NULL) && engine_load_path != NULL)
```
CodeMonkeyLeet: The error message doesn't match the logic of this conditional; I'm not sure why you're checking that the `engine_load_path` is non-null in this case. This is also not logically equivalent to the checks I was asking for in the previous iteration:

> Additional error cases to trap:
> - if (engine_id != NULL && key_id == NULL) 
> - if (engine_id == NULL && (key_id != NULL || engine_load_path != NULL))

If you want to combine all the checks, they boil down to:

```c
if ((keyfile == NULL) == (engine_id == NULL))
{
    // You must specify either the keyfile or engine_id parameter, but not both.
}
if ((engine_id != NULL) && (key_id == NULL))
{
    // You must specify a key_id if providing an engine_id
}
if ((engine_id == NULL) && (key_id != NULL || engine_load_path != NULL))
{
    // You must not specify a key_id or engine_load_path if not using engine_id (i.e. using keyfile)
}
```

tools/oesign/main.c
```diff
@@ -468,11 +480,21 @@ static const char _usage_gen[] =
 static const char _usage_sign[] =
     "Usage: %s sign {--enclave-image | -e} ENCLAVE_IMAGE "
     "{--config-file | -c} CONFIG_FILE {--key-file | -k} KEY_FILE\n"
+#if HAS_ENGINE_SUPPORT
+    "{{--engine| -n} ENGINE-NAME {--load-path | -p } ENGINE-LOAD-PATH "
```
CodeMonkeyLeet: `ENGINE-NAME`, `ENGINE-LOAD-PATH` and `KEY-ID` still not using snake case as with `CONFIG_FILE` etc.

Also lines 493-496, 534 & 535-536.

oeciteam:
Bors try

yakman2020:
bors try

yakman2020:
bors try

yakman2020:
bors try

yakman2020:
bors try

yakman2020:
bors try

yakman2020:
bors try

yakman2020:
bors try

PR #2460, opened 2020-01-23
Fix SGX prereq installation instructions for Windows 10

### LINKED ISSUE #2449: Trouble executing sample projects
Hello,
I'm having trouble executing helloworld sample project in Windows 10. Whenever I run the ninja command in the build folder I get the following pop up and console error. I'll also leave the CMakeErrorLog, which I'm not sure if it's normal, because it seems build without problem.
Any recommendations on what I should do? Is it an installation problem? The Intel SGX SDK sample project is able to run normally, so maybe I'm missing something from OE?

image (1)
image

[CMakeError.log](https://github.com/openenclave/openenclave/files/4099949/CMakeError.log)

jhand2:
On Windows 10, the SGX PSW is installed by Windows update. Fix the installation instructions to skip PSW installation for Windows 10 and add a LaunchConfigutation option to the installer script for SGX1-NoDriver without FLC.

Fixes #2449

### DIFF
--- a/docs/GettingStartedDocs/Contributors/WindowsSGX1FLCGettingStarted.md
+++ b/docs/GettingStartedDocs/Contributors/WindowsSGX1FLCGettingStarted.md
@@ -38,6 +38,13 @@ Run the following from PowerShell to deploy all the prerequisites for building O
 ./scripts/install-windows-prereqs.ps1
 ```
 
+On Windows 10, the Intel PSW is delivered via Windows Update. Running the
+executable installer will fail on Windows 10 machines. To skip PSW installation:
+
+```powershell
+./scripts/install-windows-prereqs.ps1 -LaunchConfiguration SGX1FLC-NoDriver
+```
+
 To install the prerequisites along with the Azure DCAP Client, use the below
 command. The Azure DCAP Client is necessary to perform attestation on an Azure
 Confidential Computing VM. This command assumes that you would like the
--- a/docs/GettingStartedDocs/Contributors/WindowsSGX1GettingStarted.md
+++ b/docs/GettingStartedDocs/Contributors/WindowsSGX1GettingStarted.md
@@ -41,6 +41,13 @@ would like to install the packages to `C:/oe_prereqs`.
 ./scripts/install-windows-prereqs.ps1 -InstallPath C:/oe_prereqs -LaunchConfiguration SGX1 -DCAPClientType None
 ```
 
+On Windows 10, the Intel PSW is delivered via Windows Update. Running the
+executable installer will fail on Windows 10 machines. To skip PSW installation:
+
+```powershell
+./scripts/install-windows-prereqs.ps1 -InstallPath C:/oe_prereqs -LaunchConfiguration SGX1-NoDriver -DCAPClientType None
+```
+
 If you prefer to manually install prerequisites, please refer to this
 [document](WindowsManualInstallPrereqs.md).
 
--- a/scripts/install-windows-prereqs.ps1
+++ b/scripts/install-windows-prereqs.ps1
@@ -33,7 +33,7 @@ Param(
     [string]$Python3ZipURL = 'https://www.python.org/ftp/python/3.7.4/python-3.7.4-embed-amd64.zip',
     [string]$Python3ZipHash = 'FB65E5CD595AD01049F73B47BC0EE23FD03F0CBADC56CB318990CEE83B37761B',
     [Parameter(mandatory=$true)][string]$InstallPath,
-    [Parameter(mandatory=$true)][ValidateSet("SGX1FLC", "SGX1", "SGX1FLC-NoDriver")][string]$LaunchConfiguration,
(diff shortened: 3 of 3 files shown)


BRMcLaren:
Unsure if this is out of the scope of this PR, but attestation on Windows 10 will not work. It might be a good idea to add a disclaimer that certain tests not passing is expected behaviour and that it is not a sign that the user has set up incorrectly.

Which leads into the discussion on Windows 10 experimental support versus full support.

jhand2:
> Unsure if this is out of the scope of this PR, but attestation on Windows 10 will not work. It might be a good idea to add a disclaimer that certain tests not passing is expected behaviour and that it is not a sign that the user has set up incorrectly.
>
> Which leads into the discussion on Windows 10 experimental support versus full support.

Opened  #2461 to address this

docs/GettingStartedDocs/Contributors/WindowsSGX1FLCGettingStarted.md
```diff
@@ -38,6 +38,13 @@ Run the following from PowerShell to deploy all the prerequisites for building O
 ./scripts/install-windows-prereqs.ps1
 ```
 
+On Windows 10, the Intel PSW is delivered via Windows update. Running the
```
radhikaj: Will the right version of PSW be there even on older versions of Windows 10? I don’t know if we need a specific version of PSW in order for the sdk to work correctly

jhand2: As far as I know the PSW has a stable interface and new versions won't break that interface. That said I don't know for sure if it matters.

jhand2: @radhikaj Do you think its ok to merge this?

Unfortunately, we have no control over what PSW version is installed by Windows Update. So fair we have not seen the PSW make breaking changes, but I can't find Intel documentation that says they won't. In theory the SDK should also work with older versions of the PSW, but we only test against the latest.

dthaler: nit: Windows Update with a capital U

docs/GettingStartedDocs/Contributors/WindowsSGX1GettingStarted.md
```diff
@@ -41,6 +41,13 @@ would like to install the packages to `C:/oe_prereqs`.
 ./scripts/install-windows-prereqs.ps1 -InstallPath C:/oe_prereqs -LaunchConfiguration SGX1 -DCAPClientType None
 ```
 
+On Windows 10, the Intel PSW is delivered via Windows update. Running the
```
dthaler: nit: capitalize "Update"

docs/GettingStartedDocs/Contributors/WindowsSGX1GettingStarted.md
```diff
@@ -41,6 +41,13 @@ would like to install the packages to `C:/oe_prereqs`.
...
+executable installer will fail on Windows 10 machines. To skip PSW installation:
```
dthaler: Since Windows 10 ought to be the dominant case now, it seems like the common case should be covered first.  Line 41 then should be discussed as the exception, for legacy systems, rather than being the starting point for this discussion.  That is, reordering these paragraphs would be more helpful to readers.

jhand2: In that case `SGX1-NoDriver` should be the default.  But I don't know about breaking that behavior yet since I still think the majority of SGX users are still using Windows Server (obviously I could be wrong though).

I think I would rather keep this documentation wording until we confirm what the most common case is. Then we can update the tool defaults and the documentation together.

docs/GettingStartedDocs/Contributors/WindowsSGX1GettingStarted.md
```diff
@@ -41,6 +41,13 @@ would like to install the packages to `C:/oe_prereqs`.
 ./scripts/install-windows-prereqs.ps1 -InstallPath C:/oe_prereqs -LaunchConfiguration SGX1 -DCAPClientType None
 ```
 
+On Windows 10, the Intel PSW is delivered via Windows Update. Running the
+executable installer will fail on Windows 10 machines. To skip PSW installation:
+
+```powershell
+./scripts/install-windows-prereqs.ps1 -LaunchConfiguration SGX1-NoDriver
```
CodeMonkeyLeet: Consider including all the options specified in the previous statement (i.e. `-InstallPath C:/oe_prereqs -DCAPClientType None`) since invocations of install-windows-prereqs.ps1 are not going to be additive and folks tend to just copy & paste.

jhand2:
thanks!

bors r+

BRMcLaren:
@jhand2 this story will also apply for Windows 2019 as it relies on the same update mechanism

Thanks for looking into this.

BRMcLaren:
Bors r+

Issue #2464, opened 2020-01-24
Some scan-build errors are false positives


ryanhsu19:
[Update: this issue has been resolved in PR #2468 by setting certain functions as noreturn functions.]

This is related to PR #2379 and PR #2407. About 90% of the rest of the bugs are found in third-party files and those generated by third-party test functions, while the other 10%, in my opinion, are due to `OE_ABORT()` either not recognized or not considered normal exit by scan-build.

Scan-build classifies these errors into four categories: dereference of null pointer, Unix API, double free and memory leak.

**enclave/core/sgx/td.c** (lines 96 and 105):
In function `td_pop_callsite()`, scan-build assumes either condition of the first if-statement is true, yet continues to follow the statement after `oe_abort()`. Therefore, either `td` (line 96) or `td->callsites` (line 105) results in dereference of a null pointer.

**tests/attestation_plugin/plugin/tests.c** (lines 370, 380, 408 and 421):
In brief, in each of the four “dereference of null pointer” errors, scan-build tries to continue the program even though an `OE_TEST()` fails (each in line 556 and in line 559) somewhere before the bug arises.

**tests/oeedger8r/enc/testdeepcopy.cpp** (line 226):
The “dereference of null pointer” occurs when scan-build assumes that `n` is not equal to zero and `iov` is null in line 221. This mean that the condition of `OE_TEST()` is false. However, it continues until line 226 when `iov` gets dereferenced.

**tests/crypto/utils.c** (line 36):
Like above, it assumes `str` is not null and `buf` is equal to null in line 27 and keeps going until it reaches line 36 where `buf` gets dereferenced.

**tests/argv/enc/enc.c** (line 43):
**tests/crypto/asn1_tests.c** (line 362):
**tests/crypto/ec_tests.c** (line 913):
Same here, three Unix API errors found in these three files are due to scan-build not aborting when the condition of the first `OE_TEST()` is false yet attempting to execute the second `OE_TEST()`, which uses the false condition of the first `OE_TEST()`.

**tests/memory/enc/basic.c** (lines 40, 104 and 107):
**tests/stdc/enc/enc.cpp** (line 276):
[...]

PR #2468, opened 2020-01-25
Set oe_abort(), _handle_exit() and oe_exit_enclave() as noreturn functions

ryanhsu19:
Set oe_abort(), handle_exit() and oe_exit_enclave() as noreturn functions in order to stop scan-build from generating warnings associated with oe_abort() (see issue #2464 for details). Now the number of bugs has become 140 and all bugs are found in third-party files.

### DIFF
--- a/enclave/core/optee/gp.c
+++ b/enclave/core/optee/gp.c
@@ -384,6 +384,13 @@ void oe_abort(void)
 {
     /* No return */
     TEE_Panic(TEE_ERROR_GENERIC);
+
+    /**
+     * TEE_Panic() does not return, but it is not properly annotated.
+     * Ensure the compiler does not return from this function.
+     */
+    while (1)
+        ;
 }
 
 TEE_Result TA_CreateEntryPoint(void)
--- a/include/openenclave/bits/defs.h
+++ b/include/openenclave/bits/defs.h
@@ -13,6 +13,13 @@
 #define OE_API_VERSION 2
 #endif
 
+/* OE_NO_RETURN */
+#if defined(__GNUC__)
+#define OE_NO_RETURN __attribute__((__noreturn__))
+#else
+#define OE_NO_RETURN
+#endif
+
 /* OE_PRINTF_FORMAT */
 #if defined(__GNUC__) && (__GNUC__ >= 4)
 #define OE_PRINTF_FORMAT(N, M) __attribute__((format(printf, N, M)))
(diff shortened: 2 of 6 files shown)


include/openenclave/enclave.h
```diff
@@ -22,6 +22,7 @@
 #include "bits/report.h"
 #include "bits/result.h"
 #include "bits/types.h"
+#include "corelibc/bits/defs.h"
```
jhand2: Looking at the installed headers, this one is actually not installed. Can you move OE_NO_RETURN to `include/bits/defs.h` and then just include that header here instead?

ryanhsu19: Done with it.

include/openenclave/bits/defs.h
```diff
@@ -13,6 +13,13 @@
 #define OE_API_VERSION 2
 #endif
 
+/* OE_NO_RETURN */
+#if defined(__GNUC__)
```
mingweishih: Is this specifically for the GNUC? What about clang?

jhand2: > "`__GNUC__` doesn't mean GCC specifically. All compilers that support GNU C extensions define it, including clang and ICC."

mingweishih: Got it. Raising this just because I've seen some OE files use both `__GNUC__` and `__clang__`.

jhand2: Ya I think its just redundant

CodeMonkeyLeet:
bors retry

BRMcLaren:
bors try

ryanhsu19:
Just looked at the console log with Brett and it seems that there are still some issues. I will continue to look for the problem before I am ready to have bors try again.

enclave/core/optee/gp.c
```diff
@@ -384,6 +384,10 @@ void oe_abort(void)
 {
     /* No return */
     TEE_Panic(TEE_ERROR_GENERIC);
+
+    /* Ensure that it is unreachable on ARM machines */
```
jhand2: This whole file is arm only. Maybe something like `/* TEE_Panic() does not return, but is not properly annotated. Ensure the compiler does not return from this function */`

enclave/core/optee/gp.c
```diff
@@ -384,6 +384,10 @@ void oe_abort(void)
...
+    while (1)
+        ;
```
jhand2: nit: Put the semicolon on the same line as the while.

enclave/core/optee/gp.c
ryanhsu19: Now I've got it. I will use your words in a block comment then.

enclave/core/optee/gp.c
ryanhsu19: I tried but git would tell me to run format-code, which then placed the semicolon in another line.

jhand2: oh, ok. No problem

PR #2469, opened 2020-01-27
Support clang-9 and clang-8

### LINKED ISSUE #1688: Use latest version of clang if available
The default version of clang on ubuntu 18.04 is now 8.
OE SDK seems to hard-code the use of clang-7. This may not be desirable for users.

CC: @johnkord @andschwa

anakrish:
clang-8 and clang-9 have added improvements to x86-speculative-load-hardening.
Visual Studio's clang support now bundles clang-8.

clang-8 can be supported without any changes to master.

clang-9 generates stack checking for functions in sgx/calls.c as well as memset/memcpy.
Th former functions are called before stack checking is setup for an enclave thread,
whereas the latter are called during the setup/cleanup of stack checking.
Therefore, these files must be compiled without stack protectors.

Additional fixes:
- memset, memcpy and other C files were being compiled in oecore with different flags
and then in oelibc with different flags. Now they are compiled in oelibc with correct
flags
- Whole of oelibc was being compiled with no builtin memset or memcpy. Only those two
functions are now being compiled in that manner. This allows the compiler to use
the builtins as needed when compiling oelibc.

Fixes #1688

### DIFF
--- a/docs/GettingStartedDocs/install_oe_sdk-Ubuntu_18.04.md
+++ b/docs/GettingStartedDocs/install_oe_sdk-Ubuntu_18.04.md
@@ -44,6 +44,11 @@ sudo apt -y install clang-7 libssl-dev gdb libsgx-enclave-common libsgx-enclave-
 > package which is necessary for performing remote attestation in Azure. A general
 > implementation for using Intel DCAP outside the Azure environment is coming soon.
 
+If you wish to use clang-8 or clang-9, also install them
+```bash
+sudo apt -y install clang-8 clang-9
+```
+
 If you wish to use the Ninja build system rather than make, also install
 ```bash
 sudo apt -y install ninja-build
--- a/docs/GettingStartedDocs/install_oe_sdk-Windows.md
+++ b/docs/GettingStartedDocs/install_oe_sdk-Windows.md
@@ -41,9 +41,11 @@ C:\Program Files\Git\mingw64\bin\openssl.exe
 
 ### Clang
 
-Download [Clang/LLVM for Windows 64-bit](http://releases.llvm.org/7.0.1/LLVM-7.0.1-win64.exe).
+Download [Clang/LLVM 7 for Windows 64-bit](http://releases.llvm.org/7.0.1/LLVM-7.0.1-win64.exe).
 Install Clang 7.0.1 and add the LLVM folder (typically C:\Program Files\LLVM\bin)
 to PATH. Open Enclave SDK uses clang to build the enclave binaries.
+
+OE SDK supports versions of Clang/LLVM 7 or above. They can be downloaded from [LLVM Download Page](http://releases.llvm.org/download.html)
 
 Open up a command prompt and ensure that clang is added to PATH.
 
--- a/enclave/core/CMakeLists.txt
+++ b/enclave/core/CMakeLists.txt
@@ -78,7 +78,14 @@ if (OE_SGX)
 
     # Functions in td_basic.c will change the status of td and may trigger
     # stack check fail, thus it is necessary to turn off stack check.
-    set_source_files_properties(sgx/td_basic.c PROPERTIES
+    # Functions in calls (e.g __oe_handle_main, _handle_ecalls) are called
+    # before stack protector is setup up for the enclave thread. Therefore
+    # stack check must also be turned off for calls.c.
+    # Also refer to MUSL files included below.
+    set_source_files_properties(
+        sgx/td_basic.c
+        sgx/calls.c
+	PROPERTIES
         COMPILE_FLAGS -fno-stack-protector)
 
     # OS specific sources for SGX.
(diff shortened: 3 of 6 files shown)


anakrish:
bors try

anakrish:
@radhikaj If this PR passes all tests and is approved, do you think it would be good to incorporate it into v0.8?

enclave/core/CMakeLists.txt
```diff
@@ -78,7 +78,14 @@ if (OE_SGX)
 
     # Functions in td_basic.c will change the status of td and may trigger
     # stack check fail, thus it is necessary to turn off stack check.
-    set_source_files_properties(sgx/td_basic.c PROPERTIES
+    # Functions in calls (e.g __oe_handle_main, _handle_ecalls) are called
+    # before stack protector is setup up for the enclave thread. Therefore
```
dthaler: nit: s/setup up/set up/

libc/CMakeLists.txt
```diff
@@ -663,13 +663,10 @@ add_library(oelibc STATIC
...
     ${MUSLSRC}/string/memccpy.c
     ${MUSLSRC}/string/memchr.c
-    ${MUSLSRC}/string/memcmp.c
-    ${MUSLSRC}/string/memcpy.c
     ${MUSLSRC}/string/memmem.c
-    ${MUSLSRC}/string/memmove.c
     ${MUSLSRC}/string/mempcpy.c
     ${MUSLSRC}/string/memrchr.c
-    ${MUSLSRC}/string/memset.c
+    #${MUSLSRC}/string/memset.c
```
dthaler: Commented out files look wrong, remove the line if it's not needed

dthaler:
Also you might want to update https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/install_oe_sdk-Ubuntu_18.04.md etc which say to install clang-7

anakrish:
bors try

enclave/core/CMakeLists.txt
anakrish: Fixed.

libc/CMakeLists.txt
anakrish: Fixed. Good catch.

anakrish:
> 
> 
> Also you might want to update https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/install_oe_sdk-Ubuntu_18.04.md, https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/install_oe_sdk-Windows.md, etc which say to install clang-7

I have updated these locations to also mention support for clang-8, clang-9.

@radhikaj @CodeMonkeyLeet @jhand2 @dthaler @BRMcLaren 


1.  On Linux, our cmake samples will automatically pick up the latest version of clang or revert to clang-7
2. On Linux, or make samples hard code clang-7. Should we change them to check for later versions of clang first before falling back to clang-7?
3. Should we change our manual getting started instructions to tell the user to install latest version of clang instead of telling them to install clang-7?
4. Is it possible for the ansible  scripts to pick up latest version of clang instead of clang-7?
5. Can our Windows setup scripts pick up the latest version as well?

anakrish:
bors cancel

anakrish:
bors try-

anakrish:
bors try

docs/GettingStartedDocs/install_oe_sdk-Windows.md
```diff
@@ -41,10 +41,12 @@ C:\Program Files\Git\mingw64\bin\openssl.exe
 
 ### Clang
 
-Download [Clang/LLVM for Windows 64-bit](http://releases.llvm.org/7.0.1/LLVM-7.0.1-win64.exe).
+Download [Clang/LLVM 7 for Windows 64-bit](http://releases.llvm.org/7.0.1/LLVM-7.0.1-win64.exe).
```
dthaler: Line 48 says it supports "or above" and can be downloaded from the link there.  So is line 44 still required?  If so, why is installing 7 required when you want to use a later version?

docs/GettingStartedDocs/install_oe_sdk-Ubuntu_18.04.md
```diff
@@ -44,6 +44,11 @@ sudo apt -y install clang-7 libssl-dev gdb libsgx-enclave-common libsgx-enclave-
 > package which is necessary for performing remote attestation in Azure. A general
 > implementation for using Intel DCAP outside the Azure environment is coming soon.
 
+If you wish to use clang-8 or clang-9, also install them
```
dthaler: I see you updated this file but not https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/install_oe_sdk-Ubuntu_16.04.md

Do clang-8 and clang-9 not work for Ubuntu 16.04 or should that file be updated?

BRMcLaren:
@anakrish fixed the issue with Bors.

BRMcLaren:
bors try

radhikaj:
> @radhikaj If this PR passes all tests and is approved, do you think it would be good to incorporate it into v0.8?

@anakrish We will not take this for V0.8 since we already have an RC2 and are looking to release in the next couple of days. Thanks for checking

libc/CMakeLists.txt
```diff
@@ -663,13 +663,9 @@ add_library(oelibc STATIC
     ${MUSLSRC}/string/index.c
     ${MUSLSRC}/string/memccpy.c
     ${MUSLSRC}/string/memchr.c
-    ${MUSLSRC}/string/memcmp.c
-    ${MUSLSRC}/string/memcpy.c
     ${MUSLSRC}/string/memmem.c
-    ${MUSLSRC}/string/memmove.c
```
CodeMonkeyLeet: For my understanding, are memmove.c & memcmp.c being removed because they were never specified as `no-builtin` anyway? GCC actually [documents several other builtins](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html) if we want to clean that up (e.g. `mempcpy()`) it feels like we should be consistent about it. 

Otherwise, this list seems like a good reflection of the set of functionality _intended_ in the oelibc. I'd keep them around instead of trimming a specific subset if there's no real binary impact (since these get optimized out by the compiler anyway).

docs/GettingStartedDocs/install_oe_sdk-Ubuntu_18.04.md
CodeMonkeyLeet: I would actually not add this here as it complicates the flow. I think it's fair to document the set of clang versions we support via the CMakeLists.txt check. Everywhere else we should simplify by the providing the single standard path; previously we would test against the highest version supported and recommend that in all instructions.

enclave/core/CMakeLists.txt
```diff
@@ -154,13 +161,27 @@ add_library(oecore STATIC
     tracee.c
     ${PLATFORM_SRC})
 
+
+# The compiler can generate calls to memset and memcpy before stack checking has
```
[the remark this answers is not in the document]
anakrish: @CodeMonkeyLeet Good question. memset and memcpy are often emitted behind the scenes to initialize/copy POD types e.g structs/arrays. After combing through the GCC docs, [here ](https://gcc.gnu.org/onlinedocs/gcc/Standards.html)is the definitive answer: 
> GCC requires the freestanding environment provide memcpy, memmove, memset and memcmp

There are the 4 functions that GCC can emit even when compiling for a free-standing environment.

libc/CMakeLists.txt
anakrish: @CodeMonkeyLeet 
They were removed since they were being compiled already in oecore as well as here in oelibc, using different compiler flags. It is quite hard to tell which compiled version of these functions were being picked up.

We can retain these functions here but make sure that they are compiled using the same flags in oecore.

Alternatively we can remove them, since our free-standing environment (oecore) provides memcpy, memset, memmove and memcmp provides them

In any case, I will leave them as they are for now; but document why they are being compiled in oecore.

anakrish:
I am closing this PR since there needs to be a bigger discussion about CI/CD, default versions on different platforms, which versions the samples ought to use etc.

There are however some fixes related to compiler flags, documentation, stack protector. I will submit them in a separate PR.

Issue #1312, opened 2019-01-10
tests/print has issues on Windows


CodeMonkeyLeet:
This test is Linux-centric may need to be rewritten to run correctly on Windows:

- The test uses `sh` to pipe the output to fill then runs `diff` against oracle output. While the test executes in ctest, it always passes even when the oracle files have been modified to not match expected output, indicating that the `sh` errors are not correctly captured on Windows.
 
- It currently shows up in ctest on Windows but not in VS2017 Test Manager because the CMakeLists.txt declares the command with line breaks in the middle for formatting; removing the line breaks allows Test Manager to correctly pick up the test.

yakman2020:
Is this still true?

mingweishih:
Depend on #2471

mingweishih:
I confirm that the problem no longer exist on Windows. Suggest closing the issue.

radhikaj:
Thanks so much @mingweishih

PR #2471, opened 2020-01-27
Add file syscalls on Windows

nonpolarity:
Thanks @yakman2020 that most of implementation is imported from his feature.windows branch.

functions added:
oe_syscall_open_ocall(), oe_syscall_read_ocall(), oe_syscall_write_ocall(),
oe_syscall_lseek_ocall(), oe_syscall_close_ocall(), oe_syscall_opendir_ocall(),
oe_syscall_readdir_ocall(), oe_syscall_rewinddir_ocall(), oe_syscall_closedir_ocall(),
oe_syscall_stat_ocall(), oe_syscall_access_ocall(), oe_syscall_link_ocall(),
oe_syscall_unlink_ocall(), oe_syscall_rename_ocall(), oe_syscall_truncate_ocall(),
oe_syscall_mkdir_ocall(), oe_syscall_rmdir_ocall().
funtions modified:
oe_syscall_dup_ocall().

fs, hostfs and dup testes are turned on.
print test is fixed on windows and other OS.

Due to the behavior of Windows standard c write functions that adding extra '\r' for EOL, the implementation makes use of win32 APIs and now behavior of print from enclave side is identical with enclave on Linux hosts.


tests/syscall/fs/enc/CMakeLists.txt
```diff
@@ -22,5 +22,21 @@ else()
...
+    -Wno-unused-variable
+    -Wno-missing-braces
+    -Wno-unused-parameter
+    -Wno-float-conversion
+    -Wno-sign-conversion
+    -Wno-sign-compare
+    -Wno-conversion
+    -Wno-parentheses
+    -Wno-unused-value
+    -Wno-overflow)
```
jhand2: Why do we need to add all these options?

host/windows/syscall.c
```diff
@@ -1019,6 +1759,11 @@ int oe_syscall_fcntl_ocall(
     uint64_t argsize,
     void* argout)
 {
+    if (fd < 0)
+    {
+        return -1;
```
jhand2: set errno to `OE_EINVAL`

host/windows/syscall.c
```diff
@@ -692,6 +1420,18 @@ static oe_host_fd_t _dup_socket(oe_host_fd_t oldfd)
     return -1;
 }
 
+static bool _is_socket(oe_host_fd_t oldfd)
```
jhand2: nit: parameter should just be called `fd`

host/windows/syscall.c
```diff
@@ -692,6 +1420,18 @@ static oe_host_fd_t _dup_socket(oe_host_fd_t oldfd)
...
+{
+    if (!oldfd || oldfd >= 0 && oldfd <= 2) 
+    {
+        return false;
+    }
```
jhand2: _get_socket will check the magic number of on the file descriptor structure, do I don't think you need this if check.

tests/syscall/fs/enc/CMakeLists.txt
nonpolarity: @jhand2 this pull request is still under working. These options are added to make it more convenient to comment out some calls of functions.

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

jhand2:
@nonpolarity can you add "Fixes #1312" to your PR description?

host/windows/syscall.c
```diff
@@ -366,13 +650,140 @@ oe_host_fd_t oe_syscall_open_ocall(
...
-        PANIC;
+        DWORD desired_access = 0;
+        DWORD share_mode = 0;
+        DWORD create_dispos = OPEN_EXISTING;
+        DWORD file_flags = (FILE_ATTRIBUTE_NORMAL | FILE_FLAG_POSIX_SEMANTICS);
+        WCHAR* wpathname = oe_syscall_path_to_win(pathname, NULL);
+
+        if (strcmp(pathname, "/dev/null") == 0)
+        {
+            pathname = "nul";
```
radhikaj: *null probably?

.gitattributes
```diff
@@ -27,3 +27,7 @@
 /3rdparty/libunwind/libunwind** -text -eol
 /3rdparty/mbedtls/mbedtls/**    -text -eol
 /3rdparty/musl/musl/**          -text -eol
+
+# The result of printf on host side is based on the OS. Thus it is better to
+# use different bench files on different host OS and let git overmit it.
```
anakrish: typo: overmit

host/windows/syscall.c
```diff
@@ -324,6 +324,292 @@ __declspec(noreturn) static void _panic(
...
+        return NULL;
+    }
+    // Relative or incomplete path?
+
+    // absolute path with drive letter.
+    // we do not handle device type paths ("CON:) or double-letter paths in case
+    // of really large numbers of disks (>26). If you have those, mount on
+    // windows
+    //
+    if (isalpha(path[0]) && path[1] == ':')
```
anakrish: Is path[1] safe to access when path = ""

host/windows/syscall.c
```diff
@@ -324,6 +324,292 @@ __declspec(noreturn) static void _panic(
...
+        // Abosolute path is drive letter
+        required_size = strlen(path) + 1;
+    }
+    else if (path[0] == '/' || path[0] == '\\')
+    {
+        required_size = strlen(path) + 3; // Add a drive letter to the path
+    }
+    else
+    {
+        current_dir = _getcwd(NULL, 32767);
```
anakrish: Is there a better way to do this than allocating 32K buffer every time?

nonpolarity:
bors try

nonpolarity:
bors try

host/windows/syscall.c
```diff
@@ -324,6 +324,292 @@ __declspec(noreturn) static void _panic(
...
+char* oe_win_path_to_posix(const char* path)
+{
+    size_t required_size = 0;
+    size_t current_dir_len = 0;
+    char* current_dir = NULL;
+    char* enclave_path = NULL;
+
+    if (!path)
+    {
+        return NULL;
```
radhikaj: Can we OE_RAISE_ERROR here?

host/windows/syscall.c
```diff
@@ -324,6 +324,292 @@ __declspec(noreturn) static void _panic(
...
+    size_t postlen = 0;
+
+    if (!path)
+    {
+        return NULL;
+    }
+
+    if (strcmp(path, "/dev/null") == 0)
+    {
+        path = "NUL:";
```
radhikaj: Is this intentional? NUL:?

host/windows/syscall.c
```diff
@@ -324,6 +324,292 @@ __declspec(noreturn) static void _panic(
...
+    }
+
+    pathlen = MultiByteToWideChar(CP_UTF8, 0, path, -1, NULL, 0);
+    if (post)
+    {
+        postlen = MultiByteToWideChar(CP_UTF8, 0, post, -1, NULL, 0);
```
radhikaj: Check return value, if it did not succeed, it will be 0 and you can call GetLastError()

host/windows/syscall.c
```diff
@@ -324,6 +324,292 @@ __declspec(noreturn) static void _panic(
...
+    {
+        return NULL;
+    }
+
+    if (strcmp(path, "/dev/null") == 0)
+    {
+        path = "NUL:";
+    }
+
+    pathlen = MultiByteToWideChar(CP_UTF8, 0, path, -1, NULL, 0);
```
radhikaj: Check pathlen is not 0

host/windows/syscall.c
```diff
@@ -324,6 +324,292 @@ __declspec(noreturn) static void _panic(
...
+    if (post)
+    {
+        postlen = MultiByteToWideChar(CP_UTF8, 0, post, -1, NULL, 0);
+        // 0 is the error return.
+        if (postlen == 0)
+        {
+            fprintf(
+                stderr,
+                "oe_syscall_path_to_win string conversion failed winerr=%x\n",
+                GetLastError());
```
radhikaj: Raise error here

host/windows/syscall.c
```diff
@@ -324,6 +324,292 @@ __declspec(noreturn) static void _panic(
...
+
+    WCHAR* wpath = NULL;
+
+    if (path[0] == '/')
+    {
+        if (isalpha(path[1]) && path[2] == '/')
+        {
+            wpath =
+                (WCHAR*)(calloc((pathlen + postlen + 1) * sizeof(WCHAR), 1));
+            MultiByteToWideChar(CP_UTF8, 0, path, -1, wpath, (int)pathlen);
```
radhikaj: check return values on all calls to MultiByteToWideChar

tests/print/CMakeLists.txt
```diff
@@ -7,11 +7,17 @@ if (BUILD_ENCLAVES)
 	add_subdirectory(enc)
 
 	# note: this is Linux-specific
+    if (WIN32) 
+        set(BENCHDIR ${CMAKE_CURRENT_SOURCE_DIR}/bench/windows)
+    else()
+        set(BENCHDIR ${CMAKE_CURRENT_SOURCE_DIR}/bench/others)
```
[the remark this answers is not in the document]
nonpolarity: IMO, Windows is so outstanding compared to *nix that it deserves a name here.

host/windows/syscall.c
```diff
@@ -382,18 +793,75 @@ oe_host_fd_t oe_syscall_open_ocall(
...
+        case 1:
+            _set_errno(OE_EBADF);
+            goto done;
+
+        case 2:
+            _set_errno(OE_EBADF);
+            goto done;
+
+        default:
+            break;
```
radhikaj: Should this be "goto done" as well?

host/windows/syscall.c
```diff
@@ -382,18 +793,75 @@ oe_host_fd_t oe_syscall_open_ocall(
...
+    }
+
+    if (!ReadFile((HANDLE)fd, buf, (DWORD)count, &bytes_returned, NULL))
```
radhikaj: Suggestion: This logic can probably be moved to case 0, but I might be mistaken?

tests/syscall/CMakeLists.txt
```diff
@@ -2,14 +2,13 @@
...
-
-if(UNIX)
 add_subdirectory(cpio)
-add_subdirectory(datagram)
-add_subdirectory(dup)
-add_subdirectory(epoll)
 add_subdirectory(fs)
+add_subdirectory(dup)
 add_subdirectory(hostfs)
+if(UNIX)
```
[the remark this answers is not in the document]
nonpolarity: Thanks, some comment is added.

host/windows/syscall.c
radhikaj: Also 

> *null probably?

Also why is it being assigned to a string?

host/windows/syscall.c
```diff
@@ -324,6 +324,292 @@ __declspec(noreturn) static void _panic(
...
+        // Relative path
+        WCHAR* current_dir = _wgetcwd(NULL, 32767);
+        if (!current_dir)
+        {
+            _set_errno(OE_ENOMEM);
+            return NULL;
+        }
+        size_t current_dir_len = wcslen(current_dir);
+
+        wpath = (WCHAR*)(calloc(
```
radhikaj: After doing a calloc, check whether wpath is null.

host/windows/syscall.c
radhikaj: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/getcwd-wgetcwd?view=vs-2019
   char* buffer;

   // Get the current working directory:
   if ( (buffer = _getcwd( NULL, 0 )) == NULL )
      perror( "_getcwd error" );

host/windows/syscall.c
```diff
@@ -324,6 +324,292 @@ __declspec(noreturn) static void _panic(
...
+    if (current_dir)
+    {
+        free(current_dir);
+    }
+    return enclave_path;
+}
+
+// Allocates WCHAR* string which follows the expected rules for
+// enclaves comminication with the host file system API. Paths in the format
+// /<driveletter>/<item>/<item>  become <driveletter>:/<item>/<item>
```
radhikaj: <driveletter>/<item>/<item>  become <driveletter>:\<item>\<item>

host/windows/syscall.c
```diff
@@ -480,142 +948,469 @@ ssize_t oe_syscall_writev_ocall(
...
+    WCHAR* wpathname = pdir->pdirpath;
+    // Undo abosolute path forcing again. We do this over because we need to
+    // preserve the allocation address for free.
+    if (wpathname[0] == '/' && wpathname[2] == ':')
+    {
+        wpathname++;
+    }
 
-    PANIC;
+    FindClose(pdir->hFind);
```
radhikaj: Check return values here.
Also, use memset_s and memcpy_s instead of memset and memcpy on Windows

host/windows/syscall.c
```diff
@@ -480,142 +948,469 @@ ssize_t oe_syscall_writev_ocall(
...
+    ret = _wstat64(wpathname, &winstat);
+    if (ret < 0)
+    {
+        // How do we get to  wstat's error
+
+        _set_errno(_winerr_to_errno(GetLastError()));
+        goto done;
+    }
+
+#undef st_atime
```
radhikaj: why is this being undefed?

host/windows/syscall.c
```diff
@@ -480,142 +948,469 @@ ssize_t oe_syscall_writev_ocall(
...
 }
 
 int oe_syscall_unlink_ocall(const char* pathname)
 {
-    OE_UNUSED(pathname);
+    int ret = -1;
+    WCHAR* wpathname = oe_syscall_path_to_win(pathname, NULL);
 
-    PANIC;
+    ret = _wunlink(wpathname);
```
radhikaj: https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createhardlinkw
says use DeleteFile to delete hard links.

host/windows/syscall.c
```diff
@@ -480,142 +948,469 @@ ssize_t oe_syscall_writev_ocall(
...
+        _set_errno(_winerr_to_errno(GetLastError()));
+        goto done;
+    }
 
-    PANIC;
+    new_offset.QuadPart = length;
+    if (!SetFilePointerEx(
+            h, new_offset, (PLARGE_INTEGER)&new_offset, FILE_BEGIN))
+    {
+        _set_errno(_winerr_to_errno(GetLastError()));
```
radhikaj: It is possible that h would get leaked here, since on error the program jumps to done. You can probably move CloseHandle to done to solve this.

host/windows/syscall.c
```diff
@@ -480,142 +948,469 @@ ssize_t oe_syscall_writev_ocall(
...
 
 int oe_syscall_mkdir_ocall(const char* pathname, oe_mode_t mode)
 {
-    OE_UNUSED(pathname);
-    OE_UNUSED(mode);
+    int ret = -1;
+    WCHAR* wpathname = oe_syscall_path_to_win(pathname, NULL);
 
-    PANIC;
+    ret = _wmkdir(wpathname);
```
radhikaj: The mode is unused. Is this the intent?

host/windows/syscall.c
radhikaj: @nonpolarity I think you can make the second argument 0, based on https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/getcwd-wgetcwd?view=vs-2019

host/windows/syscall.c
```diff
@@ -324,6 +324,292 @@ __declspec(noreturn) static void _panic(
...
+        required_size = strlen(path) + 1;
+    }
+    else if (path[0] == '/' || path[0] == '\\')
+    {
+        required_size = strlen(path) + 3; // Add a drive letter to the path
+    }
+    else
+    {
+        current_dir = _getcwd(NULL, 32767);
+        current_dir_len = strlen(current_dir);
```
radhikaj: Use safe string APIs(strlen_s) instead

host/windows/syscall.c
```diff
@@ -324,6 +324,292 @@ __declspec(noreturn) static void _panic(
...
+
+            char drive_letter = *current_dir;
+            *current_dir = '/';
+            current_dir[1] = drive_letter;
+        }
+        // relative path. If the path starts with "." or ".." we accomodate
+        required_size = strlen(path) + current_dir_len + 1;
+    }
+
+    enclave_path = (char*)calloc(1, required_size);
```
radhikaj: Check that calloc returns a non null value in all places.

host/windows/syscall.c
nonpolarity: Although nul is not a file/path but a equivalent device of /dev/null on Linux, the behavior and usage is very close to a file/path. On windows while nul is used as a target file/path, the file system will not create a file name d "nul" but the output will be redirected to this nul device. Thus /dev/null and nul are treated as a special case on both enclave side and Windows side.

host/windows/syscall.c
nonpolarity: @radhikaj Beautiful solution! I thought I need the max path length 260.

host/windows/syscall.c
nonpolarity: @radhikaj This is a work around here. Due to the macro here https://github.com/openenclave/openenclave/blob/e83f101b2d0113680bc10359edc05ab4b92e942e/include/openenclave/internal/syscall/sys/stat.h#L53 the winstat.st_atime will be expanded as winstat.st_atim.tv_sec.

radhikaj: Will this macro over be used?  #define st_atime st_atim.tv_sec 
I am wondering if it is used, should it be redefed after?
If not, can we undef it in a more central location?

host/windows/syscall.c
```diff
@@ -324,6 +324,450 @@ __declspec(noreturn) static void _panic(
...
+// paths without drive letter are detected and the drive added
+// /<item>/<item> -> /<current driveletter>/<item>/item>
+// relative paths are translated to absolute with drive letter
+// returns null if the string is illegal
+//
+// The string  must be freed
+//
+// we don't handle paths which start with the "\\?\" thing. We never
+// use the 8 bit version of the win32 apis. That said, we get paths in 8 bit
+// characters from the enclave because there is no "wopen" in linux
```
dthaler: The term "8 bit characters" is ambiguous since there are many "8 bit character" encodings, and they are different.  It is not clear whether the implementation below has internationalization bugs.  Normally in Linux "char*" is UTF-8, whereas "char*" in Windows is ANSI code page (which is not interoperable).  So unless you convert the code page to UTF-8, I believe that non-ASCII characters in the path will be passed incorrectly.  I *think* what you mean is that the characters in each label in the path are left in the same format as the caller passed them in, so it's the responsibility of the caller to do any character set conversion.   Assuming that's what you mean, can you update the comment to say that (and not use "8 bit version" or "8 bit characters" as terms).

host/windows/syscall.c
```diff
@@ -324,6 +324,450 @@ __declspec(noreturn) static void _panic(
...
+    }
+    return enclave_path;
+}
+
+// Allocates WCHAR* string which follows the expected rules for
+// enclaves comminication with the host file system API. Paths in the format
+// /<driveletter>/<item>/<item>  become <driveletter>:/<item>/<item>
+//
+// The resulting string, especially with a relative path, will probably contain
+// mixed slashes. We beleive Windows handles this.
```
dthaler: typo "beleive"

host/windows/syscall.c
```diff
@@ -324,6 +324,450 @@ __declspec(noreturn) static void _panic(
...
+
+    /* remove trailing slashes, but keep the one at the start of the path */
+    while (w > path+1 && w[-1] == slash) {
+        w--;
+    }
+
+    *w = '\0';
+}
+
+char* oe_win_path_to_posix(const char* path)
```
dthaler: Is the input "path" guaranteed to be in UTF-8 or ACP (ANSI code page)?  (Not sure what calls this API.)

host/windows/syscall.c
```diff
@@ -324,6 +324,450 @@ __declspec(noreturn) static void _panic(
...
+done:
+    if (current_dir)
+    {
+        free(current_dir);
+    }
+    return wpath;
+}
+
+//
+// windows is much poorer in file bits than unix, but they reencoded the
```
dthaler: grammar: "they" -> "it"

host/windows/syscall.c
```diff
@@ -366,34 +809,213 @@ oe_host_fd_t oe_syscall_open_ocall(
...
+    DWORD bytes_written = 0;
+
+    HANDLE handle = (HANDLE)fd;
 
-    return _write((int)fd, buf, (unsigned int)count);
+    // Convert fd 0, 1, 2 as needed
+    switch (fd)
+    {
+        case 0:
+            // Error. You cant write to stdin
```
dthaler: nit: "cant" -> "can't"

host/windows/syscall.c
```diff
@@ -480,142 +1102,519 @@ ssize_t oe_syscall_writev_ocall(
...
+    return ret;
 }
 
 void oe_syscall_rewinddir_ocall(uint64_t dirp)
 {
-    OE_UNUSED(dirp);
+    DWORD err = 0;
+    struct WIN_DIR_DATA* pdir = (struct WIN_DIR_DATA*)dirp;
+    WCHAR* wpathname = pdir->pdirpath;
+    // Undo abosolute path forcing again. We do this over because we need to
```
dthaler: typo "abosolute"

host/windows/syscall.c
```diff
@@ -480,142 +1102,519 @@ ssize_t oe_syscall_writev_ocall(
...
+    // preserve the allocation address for free.
+    if (wcslen(wpathname) >= 3 && wpathname[0] == '/' && wpathname[2] == ':')
+    {
+        wpathname++;
+    }
 
-    PANIC;
+    if(!FindClose(pdir->hFind))
```
dthaler: nit: use consistent spacing.  (space after if)

host/windows/syscall.c
```diff
@@ -480,142 +1102,519 @@ ssize_t oe_syscall_writev_ocall(
...
+    _set_errno(0);
+
+    if (!dirp || !entry)
+    {
+        _set_errno(OE_EINVAL);
+        ret = -1;
+        goto done;
+    }
+
+    // Find file next doesn't return '.' because it shows up in opendir and we
```
dthaler: "Find file next" -> "FindNextFileW"

tests/print/bench/README
```diff
@@ -0,0 +1,2 @@
+On enclave side, print is based on the win32 api WriteFile(). Since WriteFile() will not write out the extra CR character as fwirte does, now the behavior of print is unified that the same code will produce exaclty the same output.
```
dthaler: typos: "fwirte", "exaclty"

host/windows/syscall.c
nonpolarity: It seems "NUL:" is still same as "nul". For compatibility on new Windows, "nul" will be used across this PR.

host/windows/syscall.c
nonpolarity: Yes for older version POSIX device this macro is necessary. At the boundary between Windows and POSIX the conflicts arise. I think undef here is the most convenient method, which is the same method for cygwin.

host/windows/syscall.c
nonpolarity: Thanks, I didn't realize a memory leaking.

host/windows/syscall.c
nonpolarity: Windows doesn't do mode in the same way as linux and can only take the first part which is the attribute for owner. But readable, writable and executable attributes are obvious necessary. Thus mode is useless as an input.

host/windows/syscall.c
```diff
@@ -324,6 +326,449 @@ __declspec(noreturn) static void _panic(
...
+
+    if (!path || strnlen_s(path, MAX_PATH) == 0 || strnlen_s(path, MAX_PATH) == MAX_PATH)
+    {
+        _set_errno(OE_EINVAL);
+        goto done;
+    }
+
+    if (strcmp(path, "nul") == 0)
+    {
+        enclave_path = calloc(strlen("/dev/null"), sizeof(char));
```
radhikaj: Check that enclave_path is not null after calloc

host/windows/syscall.c
```diff
@@ -324,6 +326,449 @@ __declspec(noreturn) static void _panic(
...
+
+        required_size = pathlen + current_dir_len + postlen + 1;
+        wpath = (WCHAR*)(calloc(required_size, sizeof(WCHAR)));
+        if (!wpath)
+        {
+            _set_errno(OE_ENOMEM);
+            goto done;
+        }
+
+        oe_memcpy_s(wpath, required_size * sizeof(WCHAR), current_dir, current_dir_len * sizeof(WCHAR));
```
radhikaj: oe_memcpy_s has a return value. Recommend using OE_CHECK macro to check the return value for each call.

host/windows/syscall.c
```diff
@@ -480,142 +1103,519 @@ ssize_t oe_syscall_writev_ocall(
...
+
 uint64_t oe_syscall_opendir_ocall(const char* pathname)
 {
-    OE_UNUSED(pathname);
+    struct WIN_DIR_DATA* pdir = NULL;
 
-    PANIC;
+    pdir = (struct WIN_DIR_DATA*)calloc(1, sizeof(struct WIN_DIR_DATA));
+    if (!pdir)
+    {
```
radhikaj: Set an out of memory error here before returning.

host/windows/syscall.c
```diff
@@ -480,142 +1103,519 @@ ssize_t oe_syscall_writev_ocall(
...
+        {
+            _set_errno(_winerr_to_errno(winerr));
+            ret = -1;
+        }
+        goto done;
+    }
+
+    nlen = WideCharToMultiByte(
+        CP_UTF8, 0, pdir->FindFileData.cFileName, -1, NULL, 0, NULL, NULL);
+    (void)WideCharToMultiByte(
```
radhikaj: why is the return voided out?

host/windows/syscall.c
```diff
@@ -480,142 +1103,519 @@ ssize_t oe_syscall_writev_ocall(
...
+    {
+        goto done;
+    }
+    if (!FindClose(pdir->hFind))
+    {
+        _set_errno(_winerr_to_errno(GetLastError()));
+        goto done;
+    }
+
+    free(pdir->pdirpath);
```
radhikaj: Is it guaranteed that pdir->pdirpath is not null?

host/windows/syscall.c
nonpolarity: Thanks, comment is updated. According to the implementation, now only ASCII is supported.

host/windows/syscall.c
nonpolarity: According to the implementation, only ASCII input is supported.

host/windows/syscall.c
nonpolarity: Now free() can deal with NULL.

host/windows/syscall.c
```diff
@@ -324,6 +326,449 @@ __declspec(noreturn) static void _panic(
...
+    }
+
+    // Clean up
+
+    // There are at least 2 chars are copied to enclave_path as disk:
+    // Check the length here and replace disk: at the first 2 position as /disk.
+    if (required_size >= 3 && isalpha(enclave_path[0]) && enclave_path[1] == ':')
+    {
+        enclave_path[1] = enclave_path[0];
+        enclave_path[0] = '/';
```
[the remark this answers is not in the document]
nonpolarity: noramlize_path() only handles extra unnecessary slashes and dots.

host/windows/syscall.c
```diff
@@ -324,6 +326,449 @@ __declspec(noreturn) static void _panic(
...
+    // we do not handle device type paths ("CON:) or double-letter paths in case
+    // of really large numbers of disks (>26). If you have those, mount on
+    // windows
+    //
+
+    int origin_len =  strnlen_s(path, MAX_PATH);
+    if (origin_len >= 2 && isalpha(path[0]) && path[1] == ':')
+    {
+        // Abosolute path, just replace c: to /c
+        required_size = origin_len + 1;
```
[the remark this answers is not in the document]
nonpolarity: The extra byte is for '\0' at the end since strlen() does not count the '\0'.

host/windows/syscall.c
```diff
@@ -324,6 +326,449 @@ __declspec(noreturn) static void _panic(
...
+//
+// The string  must be freed
+//
+// we don't handle paths which start with the "\\?\" thing. We never
+// use the 8 bit version of the win32 apis. That said, we get paths in 8 bit
+// characters from the enclave because there is no "wopen" in linux
+//
+// We always convert in the ocall function.
+//
+void normalize_path(char *path, size_t origlen, char slash)
```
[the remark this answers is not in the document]
nonpolarity: Sorry the comment is for oe_win_path_to_posix() then it make confusions here. This function have slash as parameter in case it be reused to deal with Windows path.

host/windows/syscall.c
```diff
@@ -324,6 +326,449 @@ __declspec(noreturn) static void _panic(
 **==============================================================================
 */
 
+// Allocates char* string which follows the expected rules for
```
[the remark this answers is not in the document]
nonpolarity: Sorry the comment is for oe_win_path_to_posix() then it make confusions here.

nonpolarity:
> @nonpolarity can you add "Fixes #1312" to your PR description?

1312 depends on this, sorry this PR does not fix it.

host/windows/syscall.c
```diff
@@ -366,34 +837,213 @@ oe_host_fd_t oe_syscall_open_ocall(
...
+            NULL);
+        if (h == INVALID_HANDLE_VALUE)
+        {
+            _set_errno(_winerr_to_errno(GetLastError()));
+            goto done;
+        }
+
+        ret = (oe_host_fd_t)h;
+
+        // Windows doesn't do mode in the same way as linux. We can set user
```
CodeMonkeyLeet: I don't think this is the right way to do this in Windows; notably `CreateFileW()` supports specifying `SECURITY_ATTRIBUTES` that can specify DACLs for file objects in Windows, and that's where you set the RWX as ACL associal with the file, not as file attributes.

This is non-trivial though, an a mapping of mode to Windows permissions should probably be abstracted out into its own helper functions and source files. For reference, see [Creating a security descriptor for a new object in C](https://docs.microsoft.com/en-us/windows/win32/secauthz/creating-a-security-descriptor-for-a-new-object-in-c--?redirectedfrom=MSDN)

CodeMonkeyLeet: @jhand2 @anakrish @nonpolarity 
Design question: Do we actually want to implement correct RWX semantics for syscalls on Windows? It looks like based on the old Windows POSIX wrapper functions that are being called in this implementation that they just did a crappy mapping to the Windows read-only file attribute. If not, this is another area that definitely needs public documentation to call out the unsupported behaviors.

host/windows/syscall.c
```diff
@@ -366,34 +837,213 @@ oe_host_fd_t oe_syscall_open_ocall(
...
+        case OE_STDOUT_FILENO:
+        case OE_STDERR_FILENO:
+            _set_errno(OE_EBADF);
+            goto done;
+
+        default:
+            break;
+    }
+
+    if (!ReadFile(handle, buf, (DWORD)count, &bytes_returned, NULL))
```
CodeMonkeyLeet: Again for `read()` and `write()` my concern is that Linux expects these to work for sockets as well as files, and we neither fail with a useful error, nor correctly redirect it to the socket implementation.

host/windows/syscall.c
```diff
@@ -324,6 +327,475 @@ __declspec(noreturn) static void _panic(
...
+    if (current_dir)
+    {
+        free(current_dir);
+    }
+    return enclave_path;
+}
+
+// Allocates WCHAR* string which follows the expected rules for
+// enclaves comminication with the host file system API. Paths in the format
+// /<driveletter>/<item>/<item>  become <driveletter>:\\<item>\<item>
```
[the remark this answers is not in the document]
jhand2: I think the best option is to sandbox the enclave's view of the file system. So on linux my host root might be `/tmp/oe12345` or `C:\Users\foo\Temp\oe12345`. Then from inside the enclave I just do `mkdir("/test")` and it could create `/tmp/oe12345/test` or `C:\Users\foo\Temp\oe12345\test`.

And then for specifying the root we make the application tell us the literal filepath from the OS with the correct slashes, drive letter, etc.

host/windows/syscall.c
jhand2: I think that supporting RWX is actually not too challenging if we just pass `SECURITY_ATTRIBUTES` to `CreateFileW` as you say. So unless there's something more complicated that I'm missing (which is very possible), I think we should try to implement file permissions properly.

tests/syscall/hostfs/host/host.c
```diff
@@ -38,7 +38,7 @@ int main(int argc, const char* argv[])
 
     printf("=== passed all tests (test_hostfs)\n");
 #if defined(_WIN32)
-    free(tmp_dir);
+    free((char*)tmp_dir);
```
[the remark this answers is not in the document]
nonpolarity: changed, thanks.

tests/print/bench/README
```diff
@@ -0,0 +1,2 @@
+On enclave side, print is based on the win32 api WriteFile(). Since WriteFile() will not write out the extra CR character as fwrite does, now the behavior of print is unified that the same code will produce exactly the same output.
```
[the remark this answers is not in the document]
nonpolarity: Updated. Your are always the best interpreter!

tests/print/CMakeLists.txt
```diff
@@ -4,17 +4,21 @@
...
-		COMMAND sh -c "OE_LOG_LEVEL=NONE host/print_host ./enc/print_enc >testout.stdout 2>testout.stderr &&
-			diff ${CMAKE_CURRENT_SOURCE_DIR}/printhost.stdout testout.stdout &&
-			diff ${CMAKE_CURRENT_SOURCE_DIR}/printhost.stderr testout.stderr"
-		)
-else()
-
-	add_enclave_test(tests/print print_host print_enc)
+    #note: this is Linux-specific
+    if (WIN32)
+        set(BENCHDIR ${CMAKE_CURRENT_SOURCE_DIR}/bench/windows)
```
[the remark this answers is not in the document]
nonpolarity: Updated to "results". I thought using "benchmark".

tests/syscall/CMakeLists.txt
```diff
@@ -2,14 +2,13 @@
...
 
 add_subdirectory(socket)
-
-if(UNIX)
 add_subdirectory(cpio)
-add_subdirectory(datagram)
-add_subdirectory(dup)
-add_subdirectory(epoll)
 add_subdirectory(fs)
+add_subdirectory(dup)
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thank.

CodeMonkeyLeet:
@nonpolarity 

I realized that I should probably provide some guidance as to what I feel is necessary to be fixed in this PR:

* Various one-off comments on correctness
* Refactor the path canonicalization code to use existing functions and move non-production code out of syscall.c
    * We don't really test a lot of the internal logic here, so I would strongly prefer not to maintain this as production code as much as possible. 
* Document in code why we are using the `WCHAR` versions of functions.
    * If it is only for supporting long path, add tests that demonstrate this actually works as expected (i.e. does not require configuration of the Windows host machine to opt into long paths support).
    * If it does require opt-in configuration, then we need to consider supporting the `\\?\` prefixed paths to support long paths on Windows, or don't claim that we support paths longer than `MAX_PATH` on Windows.
* Detect and return appropriate POSIX errors for functions that should support both sockets and files, but do not. Document the expected behavior.
    * You should probably at least implement a wrapper type for file handles like sockets already do so that each function has a consistent way to check this via `magic`. 
    * Track for follow-up: fixing it so that these methods do correctly inspect the provided pointer and call the appropriate underlying socket/file APIs.
* Document the expected behaviors for the incomplete handling of `mode` in the functions that handle it.
    * Track for follow-up: implementing a `mode` to Windows ACL mapping for these functions so they correctly handle `mode` flags on Windows.
    * This also potentially means reworking existing implementations to not use the Windows POSIX wrapper functions like `_wstat64()` which seems to return X mode not based on file ACLs. 
* Ensure that the support for mode and sockets is internally consistent across the functions
    * e.g. don't have `read()` support sockets if `close()` doesn't also support sockets, or `stat()` provide mode bits in a different way from how `open()` allows them to be used.

host/windows/syscall.c
```diff
@@ -324,6 +327,475 @@ __declspec(noreturn) static void _panic(
...
+// <driveletter>:\<item>\<item> -> /<driveletter>/<item>/item>
+// <driveletter>:/<item>/<item> -> /<driveletter>/<item>/item>
+// paths without drive letter are detected and the drive added
+// /<item>/<item> -> /<current driveletter>/<item>/item>
+// relative paths are translated to absolute with drive letter
+// returns null if the string is illegal
+//
+// The string  must be freed
+//
+// we don't handle paths which start with the "\\?\" thing. To support the
```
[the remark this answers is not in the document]
radhikaj: Can we scope out long paths to simplify this?

host/windows/syscall.c
```diff
@@ -324,6 +331,543 @@ __declspec(noreturn) static void _panic(
...
+        deny[2] |= GENERIC_WRITE;
+    }
+    if (rwx & S_IXOTH)
+    {
+        grant[2] |= GENERIC_EXECUTE;
+    }
+    else
+    {
+        deny[2] |= GENERIC_EXECUTE;
+    }
```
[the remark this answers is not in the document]
nonpolarity: Thanks, very beautiful solution!

.gitattributes
```diff
@@ -27,3 +27,7 @@
 /3rdparty/libunwind/libunwind** -text -eol
 /3rdparty/mbedtls/mbedtls/**    -text -eol
 /3rdparty/musl/musl/**          -text -eol
+
+# The result of printf on host side is based on the OS. Thus it is better to
+# use different bench files on different host OS and let git ignore it.
+/tests/print/bench/**           -text -eol
```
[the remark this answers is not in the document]
nonpolarity: Updated, thanks. It is necessary because the host side's print results are different.

host/windows/syscall.c
```diff
@@ -324,6 +331,543 @@ __declspec(noreturn) static void _panic(
...
+            goto done;
+        }
+        sprintf(enclave_path, "%s", "/dev/null");
+        enclave_path[dst_len - 1] = '\0';
+
+        goto done;
+    }
+
+    char buffer[MAX_PATH];
+    if(GetFullPathNameA(path, MAX_PATH, buffer, NULL) == 0)
```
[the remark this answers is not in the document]
nonpolarity: Thanks, this is a very neat and efficient solution!

host/windows/syscall.c
```diff
@@ -324,6 +327,475 @@ __declspec(noreturn) static void _panic(
...
+}
+
+//
+// windows is much poorer in file bits than unix, but it reencoded the
+// corresponding bits, so we have to translate
+static unsigned win_stat_to_stat(unsigned winstat)
+{
+    unsigned ret_stat = 0;
+
+    if (winstat & _S_IFDIR)
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks.

host/windows/syscall.c
```diff
@@ -324,6 +327,475 @@ __declspec(noreturn) static void _panic(
...
+    }
+    if (winstat & _S_IFIFO)
+    {
+        ret_stat |= OE_S_IFIFO;
+    }
+    if (winstat & _S_IFREG)
+    {
+        ret_stat |= OE_S_IFREG;
+    }
+    if (winstat & _S_IREAD)
```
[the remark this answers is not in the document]
nonpolarity: Yes the lowest 3 bits in octal are all set on Windows. Since these masks are not defined on Windows, can I just bool add these bites on OE side?

host/windows/syscall.c
```diff
@@ -344,7 +815,7 @@ oe_host_fd_t oe_syscall_open_ocall(
             goto done;
         }
 
-        ret = _dup(OE_STDIN_FILENO);
+        ret = (oe_host_fd_t)GetStdHandle(STD_INPUT_HANDLE);
```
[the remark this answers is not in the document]
nonpolarity: Yes _dup() works for these fileno. Since write syscall is implemented through writefile(), here we need to dup the handle to make print work.

host/windows/syscall.c
```diff
@@ -480,142 +1130,525 @@ ssize_t oe_syscall_writev_ocall(
...
-    OE_UNUSED(offset);
-    OE_UNUSED(whence);
+    ssize_t ret = -1;
+    DWORD sfp_rtn = 0;
+    LARGE_INTEGER new_offset = {0};
 
-    PANIC;
+    new_offset.QuadPart = offset;
+    if (!SetFilePointerEx(
+            (HANDLE)fd, new_offset, (PLARGE_INTEGER)&new_offset, whence))
```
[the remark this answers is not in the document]
nonpolarity: Thanks, fixed.

host/windows/syscall.c
nonpolarity: Fixed, thanks.

host/windows/syscall.c
```diff
@@ -480,142 +1130,525 @@ ssize_t oe_syscall_writev_ocall(
...
+        _set_errno(_winerr_to_errno(GetLastError()));
+        goto done;
+    }
+
+done:
+    if (h != INVALID_HANDLE_VALUE)
+    {
+        ret = !CloseHandle(h);
+        if (ret)
+        {
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks.

host/windows/syscall.c
```diff
@@ -480,142 +1130,525 @@ ssize_t oe_syscall_writev_ocall(
...
+    if (h == INVALID_HANDLE_VALUE)
+    {
+        _set_errno(_winerr_to_errno(GetLastError()));
+        goto done;
+    }
 
-    PANIC;
+    new_offset.QuadPart = length;
+    if (!SetFilePointerEx(
+            h, new_offset, (PLARGE_INTEGER)&new_offset, FILE_BEGIN))
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks.

host/windows/syscall.c
```diff
@@ -480,142 +1130,525 @@ ssize_t oe_syscall_writev_ocall(
...
-    OE_UNUSED(buf);
+    int ret = -1;
+    WCHAR* wpathname = oe_syscall_path_to_win(pathname, NULL);
+    struct _stat64 winstat = {0};
 
-    PANIC;
+    ret = _wstat64(wpathname, &winstat);
+    if (ret < 0)
+    {
+        // How do we get to  wstat's error
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks.

host/windows/syscall.c
nonpolarity: Fixed, thanks!

host/windows/syscall.c
```diff
@@ -324,6 +327,475 @@ __declspec(noreturn) static void _panic(
...
+            q = p + strnlen_s(p, OE_PATH_MAX);
+        len = q - p;
+        if (len < 0)
+        {
+            _set_errno(OE_EINVAL);
+        }
+
+        if (len == 1 && *p == '.') {
+            /* remove current component */
+        } else if (len == 2 && memcmp(p, "..", 2) == 0) {
```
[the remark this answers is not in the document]
nonpolarity: Fixed with GetFullPathNameA(), thanks.

host/windows/syscall.c
```diff
@@ -324,6 +327,475 @@ __declspec(noreturn) static void _panic(
...
+    char* current_dir = NULL;
+    char* enclave_path = NULL;
+
+    if (!path || strnlen_s(path, MAX_PATH) == 0 || strnlen_s(path, MAX_PATH) == MAX_PATH)
+    {
+        _set_errno(OE_EINVAL);
+        goto done;
+    }
+
+    if (strcmp(path, "nul") == 0)
```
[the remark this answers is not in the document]
nonpolarity: Fixed with _stricmp(), thanks.

host/windows/syscall.c
```diff
@@ -324,6 +345,617 @@ __declspec(noreturn) static void _panic(
...
+        _set_errno(GetLastError());
+        goto done;
+    }
+
+    // Deal with directory here.
+    // Compared to file, dir has much more aces as for CREATOR OWNER, GROUP
+    // It is much easy to let Windows finish the creation then apply the PACL.
+    if (dwDesiredAccess == FILE_DIRECTORY_FILE)
+    {
+        if (!CreateDirectoryA((LPTSTR) FileName, NULL))
```
dthaler, line 840:
```suggestion
        if (!CreateDirectoryA((PCSTR) FileName, NULL))
```

Explanation: T types would be used without the A, but the A pins it to the PCSTR type

host/windows/syscall.c
```diff
@@ -480,142 +1130,525 @@ ssize_t oe_syscall_writev_ocall(
...
+
+done:
+    return ret;
 }
 
 int oe_syscall_close_ocall(oe_host_fd_t fd)
 {
-    return _close((int)fd);
+    int ret = -1;
+    HANDLE handle = (HANDLE)fd;
```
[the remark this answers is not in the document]
dthaler: Right, if a socket fd is passed in, this needs to call closesocket rather than CloseHandle.   Line 1387 does a check that tries to determine whether a handle is a socket handle or file handle, you might as well do the same check in this function around line 1343.

host/windows/syscall.c
```diff
@@ -408,214 +1201,712 @@ ssize_t oe_syscall_readv_ocall(
...
+
+        case OE_STDERR_FILENO:
+            oldhandle = GetStdHandle(STD_ERROR_HANDLE);
+            break;
+
+        default:
+            break;
+    }
+
+    // Check if oldhanle is a valid file handle.
```
dthaler, line 1381:
```suggestion
    // Check if oldhandle is a valid file handle.
```

host/windows/syscall.c
dthaler: Line 1387 uses a heuristic to determine whether a handle is a socket handle or a file handle, and the same check could be done here.

host/windows/syscall.c
```diff
@@ -408,214 +1201,712 @@ ssize_t oe_syscall_readv_ocall(
...
     {
-        errno = EINVAL;
+        _set_errno(OE_EINVAL);
+        ret = -1;
         goto done;
     }
 
-    /* Handle zero data case. */
-    if (!iov || iovcnt == 0)
+    // oe_syscall_opendir_ocall already called FindFirstFileW which returned the '.' entry.
```
dthaler, line 1466:
```suggestion
    // oe_syscall_opendir_ocall already called FindFirstFileA which returned the '.' entry.
```

Rationale: line 1434 actually used the A api instead of the W api.   In the future it should be fixed to use the W api though but that can be done in a future PR.

host/windows/syscall.c
```diff
@@ -408,214 +1201,712 @@ ssize_t oe_syscall_readv_ocall(
...
+    // To preserve the readdir semantics, if oe_syscall_readdir_ocall is passed a valid dirp
+    // with the initial offset of 0, we return '.' directly instead of calling FindNextFileW.
```
dthaler, line 1468:
```suggestion
    // with the initial offset of 0, we return '.' directly instead of calling FindNextFileA.
```

Ditto, to match 1480

host/windows/syscall.c
```diff
@@ -324,6 +327,475 @@ __declspec(noreturn) static void _panic(
 **==============================================================================
 */
 
+void normalize_path(char *path, size_t origlen, char slash)
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks.

host/windows/syscall.c
```diff
@@ -480,142 +1130,525 @@ ssize_t oe_syscall_writev_ocall(
...
+        free(wpathname);
+    }
+    return ret;
 }
 
 int oe_syscall_mkdir_ocall(const char* pathname, oe_mode_t mode)
 {
-    OE_UNUSED(pathname);
-    OE_UNUSED(mode);
+    // Windows doesn't do mode in the same way as linux and can only take the
```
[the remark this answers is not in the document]
nonpolarity: Fixed with CreateDirectory(), thanks.

host/windows/syscall.c
```diff
@@ -324,6 +327,475 @@ __declspec(noreturn) static void _panic(
...
+
+    if (!path || strnlen_s(path, MAX_PATH) == 0 || strnlen_s(path, MAX_PATH) == MAX_PATH)
+    {
+        _set_errno(OE_EINVAL);
+        goto done;
+    }
+
+    if (strcmp(path, "nul") == 0)
+    {
+        required_size = strlen("/dev/null") + 1;
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks.

nonpolarity:
bors try

host/windows/syscall.c
nonpolarity: Fixed, thanks.

host/windows/syscall.c
nonpolarity: A helper function _is_valid_file_handle() is added, thanks.

host/windows/syscall.c
nonpolarity: Fixed, thanks.

host/windows/syscall.c
nonpolarity: Fixed. I even didn't notice the difference, thanks.

host/windows/syscall.c
```diff
@@ -324,6 +345,617 @@ __declspec(noreturn) static void _panic(
...
+
+    if (post && strnlen_s(post, OE_PATH_MAX) == OE_PATH_MAX)
+    {
+        _set_errno(OE_EINVAL);
+        goto done;
+    }
+
+    if (strcmp(path, "/dev/null") == 0)
+    {
+        // Just return "nul". On windows nul is the equivolent of /dev/null
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks.

host/windows/syscall.c
```diff
@@ -324,6 +331,543 @@ __declspec(noreturn) static void _panic(
...
+#define S_IWGRP 0x0020
+#define S_IXGRP 0x0010
+#define S_IROTH 0x0004
+#define S_IWOTH 0x0002
+#define S_IXOTH 0x0001
+
+static int set_security_info_from_rwx(HANDLE hFile, oe_mode_t rwx)
+{
+    size_t ret = -1;
+    PSID sid[3];
```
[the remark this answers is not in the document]
nonpolarity: Now the array is used to traverse the aces only, thanks.

host/windows/syscall.c
```diff
@@ -324,6 +331,543 @@ __declspec(noreturn) static void _panic(
...
+    if (rwx & S_IXOTH)
+    {
+        grant[2] |= GENERIC_EXECUTE;
+    }
+    else
+    {
+        deny[2] |= GENERIC_EXECUTE;
+    }
+
+    EXPLICIT_ACCESS ea[6];
```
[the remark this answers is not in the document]
nonpolarity: Now all the types are TRUSTEE_IS_WELL_KNOWN_GROUP identically, thanks.

host/windows/syscall.c
```diff
@@ -324,6 +331,543 @@ __declspec(noreturn) static void _panic(
...
+
+    // Create a new ACL that contains the new ACEs.
+    DWORD dwRes = SetEntriesInAcl(num_ea, ea, NULL, &pACL);
+    if (ERROR_SUCCESS != dwRes)
+    {
+        _set_errno(GetLastError());
+        goto done;
+    }
+
+    // Write the new ACL to file handle. Only Dacl but no Sacl.
```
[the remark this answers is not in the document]
nonpolarity: Fix, thanks.

host/windows/syscall.c
```diff
@@ -324,6 +331,543 @@ __declspec(noreturn) static void _panic(
...
+#define S_IWUSR 0x0200
+#define S_IXUSR 0x0100
+#define S_IRGRP 0x0040
+#define S_IWGRP 0x0020
+#define S_IXGRP 0x0010
+#define S_IROTH 0x0004
+#define S_IWOTH 0x0002
+#define S_IXOTH 0x0001
+
+static int set_security_info_from_rwx(HANDLE hFile, oe_mode_t rwx)
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks.

host/windows/syscall.c
```diff
@@ -31,10 +32,16 @@
...
 #include <openenclave/internal/syscall/dirent.h>
 #include <openenclave/internal/syscall/unistd.h>
+#include <openenclave/internal/safecrt.h>
+#include <openenclave/internal/raise.h>
+#include <openenclave/corelibc/limits.h>
 #include "../hostthread.h"
 #include "../../common/oe_host_socket.h"
 #include "syscall_u.h"
 
+//#pragma comment(lib, "advapi32.lib")
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks.

host/windows/syscall.c
nonpolarity: Fixed with a helper function _is_valid_file_handle(), thanks.

host/windows/syscall.c
nonpolarity: Directory is set through SetNamedSecurityInfo(), and file is set with SECURITY_ATTRIBUTES, thanks.

host/windows/syscall.c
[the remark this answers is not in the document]
nonpolarity: Fixed with a helper function _is_valid_file_handle(), thanks.

host/windows/syscall.c
```diff
@@ -480,142 +1130,525 @@ ssize_t oe_syscall_writev_ocall(
...
+    }
+    return ret;
 }
 
 int oe_syscall_rename_ocall(const char* oldpath, const char* newpath)
 {
-    OE_UNUSED(oldpath);
-    OE_UNUSED(newpath);
+    int ret = -1;
+    WCHAR* oldwpath = oe_syscall_path_to_win(oldpath, NULL);
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks.

host/windows/syscall.c
```diff
@@ -480,142 +1130,525 @@ ssize_t oe_syscall_writev_ocall(
...
 
 int oe_syscall_access_ocall(const char* pathname, int mode)
 {
-    OE_UNUSED(pathname);
-    OE_UNUSED(mode);
+    int ret = -1;
+    WCHAR* wpathname = oe_syscall_path_to_win(pathname, NULL);
 
-    PANIC;
+    int winmode = mode & ~1; // X_OK is a noop but makes access unhappy
```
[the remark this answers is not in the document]
nonpolarity: Fixed with GetEffectiveRightsFromAcl(), thanks.

nonpolarity:
bors try

nonpolarity:
bors try

dthaler:
@jhand2 @CodeMonkeyLeet @anakrish can this be merged as is now and handle additional changes in a future PR?  (I plan to contribute a PR that addresses the Unicode comments I had.)

jhand2:
> @jhand2 @CodeMonkeyLeet @anakrish can this be merged as is now and handle additional changes in a future PR? (I plan to contribute a PR that addresses the Unicode comments I had.)

I'm good with it as-is. But @CodeMonkeyLeet has been the most involved in reviewing this code (thanks!) so I'll let him weigh in first.

anakrish:
> @jhand2 @CodeMonkeyLeet @anakrish can this be merged as is now and handle additional changes in a future PR? (I plan to contribute a PR that addresses the Unicode comments I had.)

I don't have much expertise over the contents of this PR and have not provided any valuable review for this. I would like to see this PR merged too.

Thank you for volunteering to address the Unicode comments.

kuanggang-git:

anakrish:
@nonpolarity I appreciate you diligently and patiently working through all the feedback. Great to see this PR merged.

PR #2472, opened 2020-01-27
Implement nanosleep ocall in edl

### LINKED ISSUE #2405: Refactor sleep/nanosleep syscalls to be first-class syscalls
Currently sleep and nanosleep are treated as special case syscalls, even though they basically work as all the other syscalls do (by dispatching the call to the host via ocall). Refactor so that they are treated as normal syscalls instead of special-cased.

See discussion in #2377 for more information

jhand2:
**Disclaimer**: oeedger8r ocalls are much slower than hand-written ocalls. For this reason, nanosleep takes longer than before. This led to some libcxx test failures, which is why the tolerance is increased with `-DTEST_HAS_SANITIZERS`. We should consider holding this PR until the performance of this ocall can be improved.

Currently sleep and nanosleep are implemented as special-case ocalls and
not in EDL. Implement them in EDL.

Additionally, fix nanosleep to be properly implemented for resuming after
being interrupted for signal handling.

Note that in musl libc sleep is just implemented with nanosleep which is why the sleep ocall was removed entirely.

Fixes #2405

### DIFF
--- a/host/windows/syscall.c
+++ b/host/windows/syscall.c
@@ -1532,8 +1532,8 @@ int oe_syscall_nanosleep_ocall(struct oe_timespec* req, struct oe_timespec* rem)
 
     Sleep((DWORD)milliseconds);
 
-    // Windows sleep will always sleep is not interruptable by hardware
-    // exception handling. Just wait the whole time and zero rem.
+    // Windows sleep is not interruptable by hardware exception handling. Just
+    // wait the whole time and zero rem.
     if (rem)
         memset(rem, 0, sizeof(*rem));
 
--- a/host/windows/syscall.c
+++ b/host/windows/syscall.c
@@ -1502,3 +1502,38 @@ int oe_syscall_uname_ocall(struct oe_utsname* buf)
 done:
     return ret;
 }
+
+/*
+**==============================================================================
+**
+** nanosleep():
+**
+**==============================================================================
+*/
+
+int oe_syscall_nanosleep_ocall(struct oe_timespec* req, struct oe_timespec* rem)
+{
+    uint64_t milliseconds = 0;
+
+    if (!req)
+    {
+        _set_errno(OE_EINVAL);
+        return -1;
+    }
+
+    milliseconds += req->tv_sec * 1000UL;
+    milliseconds += req->tv_nsec / 1000000UL;
+
+    for (uint64_t i = 0; i < div; i++)
+        Sleep(UINT_MAX);
+
+    Sleep((DWORD)mod);
+
+    // Windows sleep will always sleep is not interruptable by hardware
+    // exception handling. Just wait the whole time and zero rem.
+    if (rem)
+        memset(rem, 0, sizeof(*rem));
+
+    _set_errno(0);
+    return 0;
+}
--- a/host/windows/syscall.c
+++ b/host/windows/syscall.c
@@ -1524,10 +1524,13 @@ int oe_syscall_nanosleep_ocall(struct oe_timespec* req, struct oe_timespec* rem)
     milliseconds += req->tv_sec * 1000UL;
     milliseconds += req->tv_nsec / 1000000UL;
 
-    for (uint64_t i = 0; i < div; i++)
+    while (milliseconds > UINT_MAX)
+    {
         Sleep(UINT_MAX);
+        milliseconds -= UINT_MAX;
+    }
 
-    Sleep((DWORD)mod);
+    Sleep((DWORD)milliseconds);
 
     // Windows sleep will always sleep is not interruptable by hardware
     // exception handling. Just wait the whole time and zero rem.
--- a/common/syscall.edl
+++ b/common/syscall.edl
@@ -21,6 +21,7 @@
 
 enclave {
 
+    include "openenclave/corelibc/time.h"
     include "openenclave/internal/syscall/netdb.h"
     include "openenclave/internal/syscall/sys/epoll.h"
     include "openenclave/internal/syscall/sys/poll.h"
@@ -407,5 +408,9 @@ enclave {
             [out, count=size] unsigned int* list)
             propagate_errno;
 
+        int oe_syscall_nanosleep_ocall(
+            [in] struct oe_timespec* req,
+            [in, out] struct oe_timespec* rem)
+            propagate_errno;
     };
 };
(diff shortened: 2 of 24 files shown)


thomasten:
Is oeedger8r ocalls being slower a known issue? Do you know why they are slower?

I think the libcxx tests do not fail because the ocall is slower, but because this_thread::sleep_for previously resulted in nanosleep of 0. So these two calls
https://github.com/openenclave/openenclave/blob/4e8a8780a495b3e9c2eb4fe2317e78f56cbfb912/3rdparty/libcxx/libcxx/test/std/thread/thread.mutex/thread.mutex.requirements/thread.sharedtimedmutex.requirements/thread.sharedtimedmutex.class/lock_shared.pass.cpp#L75-L78
https://github.com/openenclave/openenclave/blob/4e8a8780a495b3e9c2eb4fe2317e78f56cbfb912/tests/libcxx/enc/enc.cpp#L114-L125
are now taking much more time. The first one is required for the test being meaningful, but the second one may be improved.

Instead of `-DTEST_HAS_SANITIZERS` you may consider `cherry-pick`ing my fix
https://github.com/openenclave/openenclave/pull/2429/commits/12594df12a71a2932f28c6bc5297687a0530011f

jhand2:
> Is oeedger8r ocalls being slower a known issue? Do you know why they are slower?

They are, but I don't know how much slower. The reason is that oeedger8r allocates a parameter buffer on the host (which itself requires an ocall) regardless of how large the passed parameters actually are. So I know its slower but I don't know by how much.

We've been considering an optimization where we use a single static buffer per thread for any ocalls that have sufficiently small parameters.

> 
> I think the libcxx tests do not fail because the ocall is slower, but because this_thread::sleep_for previously resulted in nanosleep of 0.

You have a good point, they might just be failing because we're doing a sleep at all now, not necessarily because the previous implementation was that much lower overhead. I'll try cherry-picking your fix.

jhand2:
> 
> 
> @jhand2 Thanks for doing this.
> 
> [nanosleep](https://linux.die.net/man/2/nanosleep) is a high-resolution sleep. Do we fulfill this requirement most of the time? Do we really need nanosleep or could it just be a sleep with granularity of milli seconds?

Basically there's no such thing as a sleep with millisecond granularity. `usleep()` takes microseconds, but it is considered deprecated. In reality, in musl there's no such thing as sleep in musl.

This is the `sleep()` implementation:

```
unsigned sleep(unsigned seconds)
{
	struct timespec tv = { .tv_sec = seconds, .tv_nsec = 0 };
	if (nanosleep(&tv, &tv))
		return tv.tv_sec;
	return 0;
}
```

and here is `usleep()`

```
int usleep(unsigned useconds)
{
	struct timespec tv = {
		.tv_sec = useconds/1000000,
		.tv_nsec = (useconds%1000000)*1000
	};
	return nanosleep(&tv, &tv);
}
```

At the syscall layer there is only nanosleep. I'll look into profiling how much overhead there is for making a nanosleep call.

Although it is worth mentioning that even outside an an enclave, nanosleep is making a syscall and requires a context switch.

anakrish:
@jhand2 Thanks for the explanation. I was not concerned about the overhead of the context switch; but rather about the drastically different resolutions on Linux (nano) and Windows (milli).
In such a case, does it even make sense to support a nanosleep primitive?

jhand2:
> @jhand2 Thanks for the explanation. I was not concerned about the overhead of the context switch; but rather about the drastically different resolutions on Linux (nano) and Windows (milli).
> In such a case, does it even make sense to support a nanosleep primitive?

@anakrish Musl only uses nanosleep. So at the end of the day we have to support nanosleep (or replace more of musl which I don't really want to do). If the concern is that Windows doesn't have nanosecond resolution sleep, it is technically possible to implement, it is implemented in libcxx. It just isn't simple.

jhand2:
> I think the libcxx tests do not fail because the ocall is slower, but because this_thread::sleep_for previously resulted in nanosleep of 0. So these two calls

@thomasten good call. I cherry-picked your commit and it fixed the issue

jhand2:
bors try

jhand2:
bors try

jhand2:
bors try

jhand2:
@thomasten @CodeMonkeyLeet @anakrish 

So this implementation is correct and most of the C++ tests pass now. On Windows I do still see 1 test failure:
`3rdparty\libcxx\libcxx\test\std\thread\thread.mutex\thread.lock\thread.lock.unique\thread.lock.unique.locking\lock.pass.cpp`

tl;dr - can we remove this test and mark it as not supported?

This test fails because it tests thread contention where
1. thread 1 acquires lock
2. thread 2 tries to acquire lock
3. thread 1 sleeps for 250ms
4. thread 1 releases lock
5. thread 2 gets the lock

The test checks that the time that thread 2 was waiting to acquire the lock has overhead (non-sleep time) of less than 25ms.

From testing on my vm, the sleep call has an overhead of about 16ms and acquiring a lock has overhead of about 15ms. I don't really think there's anything we can do about this.

jhand2:
Even though bors succeeded I still have a concern about this test since under no circumstances can I get it to succeed locally. The latency is always too high.

thomasten:
> I think the only way around this is to disable the test.

I agree.

host/windows/syscall.c
```diff
@@ -1502,3 +1502,41 @@ int oe_syscall_uname_ocall(struct oe_utsname* buf)
...
+
+    while (milliseconds > UINT_MAX)
+    {
+        Sleep(UINT_MAX);
+        milliseconds -= UINT_MAX;
+    }
+
+    Sleep((DWORD)milliseconds);
+
+    // Windows sleep will always sleep is not interruptable by hardware
```
dthaler: Can't parse grammar of "will always sleep is not"

jhand2:
Fixed typo in comment.

bors r+

jhand2:
Test failed because I got unlucky and crypto test verified a cert that was generated on a minute boundary. We should consider loosening the requirements of that test but it's unrelated to this PR. Trying again

bors r+

jhand2:
bors got cancelled 😢 

bors r+

jhand2:
Resolved merge conflicts

bors r+

jhand2:
remote_attestation sample failed with infra issue. Recorded for aggregation.

bors r+

Issue #2220, opened 2019-10-01
Samples should not have unused parameters


johnkord:
Based on the comments from PR #2204, the samples should not require people to cast parameters to void because they are unused.  The functions should be designed to not take those unused parameters in the first place

radhikaj:
Investigate if we can change sealed_data_size param in edl to a constant. Or use the param and check whether it is what you would expect.

radhikaj:
@ryanhsu19 will investigate.

ryanhsu19:
In most samples, "level" remains unused in debug message printing functions such as my_debug() in samples/attested_tls/client/enc/client.cpp and samples/attested_tls/server/enc/server.cpp. 
However, when it gets to the bottom most related functions require the use of mbedtls_debug_print_msg(), which compares the debug threshold and the level before doing the rest. Therefore, I did not remove it.

ryanhsu19:
Currently I am trying to remove the "data" parameter from cert_verify_callback() because it remains voided and unused in all functions (4 in samples/attested_tls) in the sample/attested_tls. There is no problem compiling them, but it will be stuck with on tls_e2e with an error OE_PLATFORM_ERROR returned during ctest, and I am still trying to figure out why.

ryanhsu19:
Also, Alvin and I found that, by using oegdb to debug tests/tls_e2e/host/host.cpp, a signal SIGILL (illegal instruction) returns whenever it tries to run a server. By laying out the ASM, it seems to happen when it executes cpuid in mbedtls_aesni_has_support() in library/aesni.c. It also happened before I started modifying any file. Any ideas?

johnkord:
Hey @ryanhsu19 that's expected! The CPUID instruction is considered illegal inside an SGX enclave.  https://software.intel.com/en-us/node/703005
There's a "vectored exception handler" framework in Open Enclave that properly handles this situation and returns the CPUID values. In the end, CPUID ends up working through an OCALL. However from a debugger, you'll be notified that the illegal instruction signal was sent. You can ignore this signal inside GDB if you want by running `ignore SIGILL`

ryanhsu19:
@johnkord Got it, thanks! I will try again with the condition soon.

radhikaj:
@ryanhsu19 
In this function:
int ecall_dispatcher::unseal_data(
    sealed_data_t* sealed_data,
    size_t sealed_data_size,
    unsigned char** data,
    size_t* data_size)

[...]

Issue #2431, opened 2020-01-15
EDL does not support macros


ryanhsu19:
This is related to the comments in PR #2220. In samples/data-sealing/datasealing.edl, you can tell the edger8r the size of a pointer, but you cannot assign a macro to it because the edger8r treats it as a new member of the corresponding struct; that is,
 - it works by changing [in, size = sealed_data_size] to [in, size = 16], but
 - it does not work by changing [in, size = sealed_data_size] to [in, size = IV_SIZE], where IV_SIZE is defined as 16 in the header file included in the enclave.

anakrish:
CC: @mingweishih

mingweishih:
This feature is supported by the current oeedger8r implementation.

anakrish:
The recommended oeedger8r syntax is to use arrays with macro/enum dimension:
```c
  public void fun([in] int data[IV_SIZE]);
```

PR #2473, opened 2020-01-27
Make use of sealed_data_size in data-sealing samples to fix issue #2220

ryanhsu19:
This is related to issues #2220 and #2431. After discussion with @anakrish and @radhikaj, since sealed_data_size is always equal to IV_SIZE in the unseal_data function, we should use sealed_data_size instead of voiding it at the beginning. Therefore, I have added an if-statement to ensure sealed_data_size is IV_SIZE, and use it throughout the rest of the function.

### DIFF
--- a/samples/data-sealing/common/dispatcher.cpp
+++ b/samples/data-sealing/common/dispatcher.cpp
@@ -179,10 +179,17 @@ int ecall_dispatcher::unseal_data(
     size_t seal_key_size = 0;
     uint8_t* key_info = NULL;
     size_t key_info_size = 0;
-    (void)sealed_data_size;
 
     unsigned char* data_buf = NULL;
     int ret = 0;
+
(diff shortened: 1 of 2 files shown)


samples/data-sealing/common/dispatcher.cpp
```diff
@@ -173,17 +173,24 @@ int ecall_dispatcher::unseal_data(
     size_t* data_size)
 {
     oe_result_t result = OE_OK;
-    unsigned char iv[IV_SIZE];
+    unsigned char iv[sealed_data_size];
```
anakrish: Consider keeping the original line since non-constant array sizes are not supported by all compilers.

radhikaj:
Bors r+

ryanhsu19:
Changed both the title and the commit message.

Issue #2105, opened 2019-08-16
EDL enhancements


mikbras:
This issue requests various EDL enhancements to better guard against implicit transfer of host memory. EDL functions that implicitly pass host memory to the enclave pose two main problems.

- Potential leaking of secrets when enclaves inadvertently write to host memory (when the developer might have mistaken it for trusted memory).
- Inability to port applications to TrustZone and other TEEs that do not support shared memory.

The **deep-copy** work (currently underway) addresses these problems in part. Ideally, the **EDL generator** (**oeedger8r**) should error out when encountering dangerous EDL specifications. To provide compatibility with the **Intel(tm) EDL generator**, options should be provided to relax this checking (on a case-by-case basis). In the strictest mode, edge routines should serialize all parameters so that pointers are never passed between untrusted and trusted components. Each of the following sections describes an enhancement.

No foreign structures
------------------------

Issue #2008, opened 2019-06-27
Replace all hand-written edge functions with EDL functions


anakrish:
There are many core edge functions that are hand-written.

- These are a constant source of lingering security vulnerabilities.
- There is a tendency to add yet another hand-written edge function for a new feature
- There have been instances where a hand-written function was secure during review and check-in, but became unsecure due to subsequent code refactoring.

It is better to use auto-generation for all edge functions.

CC: @CodeMonkeyLeet  @ihaller

vtikoo:
Part II PR fixing this issue: https://github.com/openenclave/openenclave/pull/2045. There are handful of handwritten EDL's which have been verified as being safe.

BRMcLaren:
Closing on behalf of Vikas.

jhand2:
Reopening this issue because the following ocalls still potentially need to be converted to EDL:

OE_OCALL_WAKE_HOST_WORKER
OE_OCALL_THREAD_WAIT
OE_OCALL_THREAD_WAKE

jhand2:
Sleep is already converted here https://github.com/openenclave/openenclave/pull/2472

jhand2:
@ryanhsu19 please look into which of these ocalls can actually be ported to EDL

radhikaj:
https://github.com/openenclave/openenclave/pull/3347 fixes get time.
The others are not edlizable in the near future.

PR #2482, opened 2020-01-29
Document design for ocall opt-in mechanism

jhand2:
Currently many ocalls are compiled as part of the OE SDK and are included by default by any enclave that calls them. Add a design document to detail an explicit opt-in mechanism that would require developers to import all ecalls and ocalls defined in edl prior to their use.

A rendered version of this markdown can be found [here](https://github.com/jhand2/openenclave/blob/ocall_optin_design/docs/DesignDocs/system_ocall_opt_in.md)

### DIFF
--- a/docs/DesignDocs/system_ocall_opt_in.md
+++ b/docs/DesignDocs/system_ocall_opt_in.md
@@ -1,0 +1,139 @@
+# Explicit Enclave Opt-in to System OCalls
+
+System OCalls are OCalls that are part of Open Enclave core libraries. These are generally not
+called by the enclave developer but instead are called by other functions internal to OE SDK.
+
+Because information from inside the enclave may be transmitted to the untrusted host
+during an OCalls, an enclave developer should be able to opt-out of these OCalls
+being made. They should also have visibility into what OCalls might be called by OE SDK.
+
+## EDL installation
+
+In order to support ocall opt-in, system EDL files will now be installed as part of the SDK at
+`${install_path}/include/openenclave/edl/`. This allows them to be imported into user edl by the
+enclave developer.
+
+## Developer opt-in
+
+In order to use any system OCall, it must be imported into the enclave's EDL file. If it is not
+included, but the OCall is still called by the enclave application or some part of the SDK,
+enclave linking will fail.
+
+Users can import a system OCall the same way any other EDL file would be imported. For example,
+if an enclave with EDL file `sample.edl` wants to use the function `oe_log_ocall()` they can do
+one of the following:
+
+*Note: These examples assume that `${install_path}/include` is passed to edger8r as an EDL search path*
+
+1. Import just the required OCall
+
+```
+// sample.edl
+enclave
+{
+    from "openenclave/edl/logging.edl" import oe_log_ocall;
+
+    trusted
+    {
+        public void my_sample_ecall();
+    };
+
+    untrusted
+    {
+        public void my_sample_ocall();
+    };
+}
+```
+
+2. Import all OCalls from `tee.edl`
+
+```
+// sample.edl
+enclave
+{
+    from "openenclave/edl/logging.edl" import *;
+
+    trusted
+    {
+        public void my_sample_ecall();
+    };
+
+    untrusted
+    {
+        public void my_sample_ocall();
+    };
+}
+```
+
+In example (1), linking would fail if the enclave made a call to `oe_realloc_ocall()`
+(also part of tee.edl) unless the developer modified the import line to
+`from "openenclave/edl/tee.edl" import oe_log_ocall, oe_realloc_ocall;`. In this scenario,
+example (2) will work without modification.
+
+## EDL local structures
+
+Under this design, system EDL files will prefer using local structures; that is, structures
+that are defined in EDL. This is distinct from foreign structures, which are defined in
+C/C++ header files and included in the EDL file.
+
+In the case that a foreign structure must be used, system EDL files may only include headers
+that are installed with the SDK.
+
+Local EDL structures provide additional security benefits detailed in the
+[Full EDL Serialzation design doc](
+https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/full_edl_serialization.md#motivation)
+so foreign structures should be avoided whenever possible.
+
+## Developer opt-out
+
+Because system OCalls are generally called by the SDK implementation, developers
+may not have direct control over which OCalls will be called by their application.
+A developer can refuse to import a function in EDL which will cause a linker error
+if it is used, but they should also be able to take action to resolve this failure.
+
+To this end, any OE SDK feature which makes an OCall on behalf of an enclave
+should provide a mechanism to disable the behavior which makes an OCall.
+
+For example, the switchless calls feature makes an OCall
+`oe_wake_switchless_worker_ocall()`. If there is code compiled into the enclave
+which calls this function, the enclave developer will be forced to import that
+function into their own EDL. To allow the developer to opt out of this OCall,
+it should be possible for the enclave developer to turn off the switchless calls
+feature such that an enclave which does not use switchless calls will not require
+this OCall. This mechanism should not require the developer to recompile the OE
+SDK.
+
+## Exceptions to EDL import and opt-in/out mechanisms
+
+A small number of OCalls in OE SDK are not implemented in EDL. Until these are
+converted to EDL, these OCalls will always be included and the developer cannot
+opt out of them.
+
+The following 4 OCalls can be implemented in EDL, they just aren't today. Issue
+`#2008` tracks converting these to EDL.
+
+* `OE_OCALL_MALLOC`
+* `OE_OCALL_FREE`
+* `OE_OCALL_THREAD_WAIT`
+* `OE_OCALL_THREAD_WAKE`
+
+## List of system OCalls
+
+All system OCalls (other than the previously mentioned exceptions) are found
+in one of the following EDL files
+
+* `common/logging.edl`
+* `common/memory.edl`
+* `common/epoll.edl`
+* `common/fcntl.edl`
+* `common/ioctl.edl`
+* `common/poll.edl`
+* `common/signal.edl`
+* `common/socket.edl`
+* `common/time.edl`
+* `common/unistd.edl`
+* `common/utsname.edl`
+* `common/syscall.edl`
+* `common/syscall.edl`
+* `common/sgx/sgx.edl`
+* `common/sgx/switchless.edl`


achamayou:
That sounds good to me, we would love to use this in CCF.

mikbras:
Thank you Jordan for doing this. Looks great to me as is.

**For future consideration:**
Local structures are needed to support **fully serializable EDL**, which we can discuss separately. In short, OE must support serialization or flattening of all parameters as needed by TrustZone and possibly other TEEs (consider VSM and AMD). The SGX memory model makes all host memory visible to the enclave, but TrustZone does not. This means that deep structures (structure pointer fields) will are copied properly across ECALLs in TrustZone, where these pointer values would be unusable. But even with SGX, where the pointers can be read, they may lead to TOC-TOU vulnerabilities. There are other platforms as well like AMD Epyc where an AMD processor communicates with a ARM processor, where the byte orders are different. If we are to support this architecture, then the edge routines will have to handle byte ordering as well as flattening. Further, we have considered placing the enclave out of process and using EDL to support a wire protocol. All of these makes local structures the only realistic choice and foreign structures are unworkable. The layout of foreign structures is not known to EDL.

jhand2:
> All of these makes local structures the only realistic choice and foreign structures are unworkable. The layout of foreign structures is not known to EDL.

Thanks Mike. For this feature I plan to leave any structures that are in public headers just for simplicity of the change (its already going to be a big change). But ya I agree ultimately we will need to use all local EDL structures. This is tracked by #2105 and hopefully we can also do this for 0.9.

mikbras:
> All of these makes local structures the only realistic choice and foreign structures are unworkable. The layout of foreign structures is not known to EDL.
> 
> Thanks Mike. For this feature I plan to leave any structures that are in public headers just for simplicity of the change (its already going to be a big change). But ya I agree ultimately we will need to use all local EDL structures. This is tracked by #2105 and hopefully we can also do this for 0.9.

mikbras:
> All of these makes local structures the only realistic choice and foreign structures are unworkable. The layout of foreign structures is not known to EDL.
> 
> Thanks Mike. For this feature I plan to leave any structures that are in public headers just for simplicity of the change (its already going to be a big change). But ya I agree ultimately we will need to use all local EDL structures. This is tracked by #2105 and hopefully we can also do this for 0.9.

Thank you Jordan. The rules seem pretty simple:
- only flat structures can be foreign structures (but should ultimately be local structure to support byte ordering (which will be required by Epyc).
- all deep structures must be local structures.

Let me know if I can help with **syscall.edl**. I took a quick look and the changes seem pretty straightforward to me. Here's a diff of one change:

https://github.com/openenclave/openenclave/compare/mikbras.stat-edl

This is the easy way to do it. The harder change would be to use the interface structure throughout the syscall code. By copying at the callsite, the change seemed pretty simple to me.

anakrish:
@jhand2  Great writeup! Nice idea of using github PR to have this discussion.


> To this end, any OE SDK feature which makes an ocall on behalf of an enclave should provide a mechanism to disable the ocall at compile-time and fail gracefully.

Is this a must-have or a nice-to-have?

jhand2:
> 
> 
> @jhand2 Great writeup! Nice idea of using github PR to have this discussion.
> 
> > To this end, any OE SDK feature which makes an ocall on behalf of an enclave should provide a mechanism to disable the ocall at compile-time and fail gracefully.
> 
> Is this a must-have or a nice-to-have?

It is a nice to have but I think it makes the feature more useful if we do it where possible.

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
...
+    untrusted
+    {
+        public void my_sample_ocall();
+    };
+}
+```
+
+In example (1), linking would fail if the enclave made a call to `oe_realloc_ocall()`
+(also part of tee.edl) unless the developer modified the import line to
+`from "openenclave/edl/tee.edl" import oe_log_ocall, oe_realloc_ocall;`. In this scnario,
```
radhikaj: typo with scenario

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
+# Explicit Enclave Opt-in to system ocalls
+
+System ocalls are ocalls that are part of openenclave core libraries. These are generally not
+called by the enclave developer but instead are called by other functions internal to OE SDK.
+
+Because ocalls have potential security impact, and enclave developer should be able to opt
```
radhikaj: Probably the "and" is "any"?

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
...
+## Exceptions to edl import and opt-in/out mechanisms
+
+A small number of ocalls in OE SDK are not implemented in edl and cannot be for
+technical reasons. These ocalls will always be included and the developer cannot
+opt out of them because they support core SDK functionality.
+
+These calls are:
+* `OE_OCALL_MALLOC`
+* `OE_OCALL_FREE`
+* `OE_OCALL_THREAD_WAIT`
```
CodeMonkeyLeet: Do we really need to keep the threading ones? It seems like this is a side effect of some functions using oe_mutex_* rather than oe_spinlock_*, otherwise, these primarily service pthread functions that are opt-in.

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
...
+of OE SDK compilation.
+
+## Exceptions to edl import and opt-in/out mechanisms
+
+A small number of ocalls in OE SDK are not implemented in edl and cannot be for
+technical reasons. These ocalls will always be included and the developer cannot
+opt out of them because they support core SDK functionality.
+
+These calls are:
+* `OE_OCALL_MALLOC`
```
CodeMonkeyLeet: It's probably useful to explain the context in which these calls are used in the enclave runtime to assuage any concerns.

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
...
+* `OE_OCALL_FREE`
+* `OE_OCALL_THREAD_WAIT`
+* `OE_OCALL_THREAD_WAKE`
+* `OE_OCALL_WAKE_HOST_WORKER`
```
CodeMonkeyLeet: Similarly, is there a way we can factor apart the use of switchless calls? The EDL integration makes this tricky I suppose, but we should probably have documentation to indicate how to avoid hitting this ocall at least.

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
...
+
+## List of system ocalls
+
+All system ocalls (other than the previously mentioned exceptions) are found
+in one of the following edl files
+
+* `common/tee.edl`
+* `common/syscall.edl`
+* `common/sgx/sgx.edl`
```
CodeMonkeyLeet: This one is tricky because it's not really optional once you pull in oeenclave with its dependency on ocalls for sgx attestation quoting methods. `oe_get_cpuid_table_ocall()` is also always invoked on enclave creation, so you can't avoid including this.

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
+# Explicit Enclave Opt-in to system ocalls
```
CodeMonkeyLeet: Title casing consistency: "Explicit Enclave Opt-in to System OCalls"

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
...
+
+System ocalls are ocalls that are part of openenclave core libraries. These are generally not
+called by the enclave developer but instead are called by other functions internal to OE SDK.
+
+Because ocalls have potential security impact, and enclave developer should be able to opt
+out of these ocalls being made and should have visibility into what ocalls might be called
```
CodeMonkeyLeet: Typo: opt-out

docs/DesignDocs/system_ocall_opt_in.md
CodeMonkeyLeet: I would suggest consistency with [Intel documentation](https://software.intel.com/en-us/node/702976) casing on this term: OCall, ECall

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
...
+
+## EDL installation
+
+In order to support ocall opt-in, system EDL files will now be installed as part of the SDK at
+`${install_path}/include/openenclave/edl/`. This allows them to be imported into user edl by the
+enclave developer.
+
+## Developer opt-in
+
+In order to use any system ocall, it must be imported into the enclave's edl file. If it is not
```
CodeMonkeyLeet: Consistency: "EDL file" as a proper noun or ".edl file" as a file extension reference.

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
+# Explicit Enclave Opt-in to system ocalls
+
+System ocalls are ocalls that are part of openenclave core libraries. These are generally not
```
CodeMonkeyLeet: Branding: "Open Enclave"

docs/DesignDocs/system_ocall_opt_in.md
CodeMonkeyLeet: We should elaborate on the motivation for this proposal; referring to vague "potential security impact" tends to generate FUD.

There are two major concerns here:
* A developer needs to be aware of what information or functionality an enclave is reaching out to the untrusted host for, e.g. time syscalls, so that they can be evaluated in the app's security model. Having this be an explicit developer opt-in makes it much more obvious to a developer.
* A developer needs to be aware of when the enclave is calling out to the untrusted host. For example, they need to know if calling an OE function will result in an OCall, and understand that they must not hold enclave locks across such calls. The opt-in mechanism introducing link failures can help surface awareness of OE calls being used that rely on OCalls that would not otherwise be noticed.

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
...
+Because system ocalls are generally called by the SDK implementation, developers
+may not have direct control over which ocalls will be called by their application.
+A developer can refuse to import a function in EDL which will cause a linker error
+if it is used, but they should also be able to take action to resolve this failure.
+
+To this end, any OE SDK feature which makes an ocall on behalf of an enclave
+should provide a mechanism to disable the ocall at compile-time and fail
+gracefully.
+
+If this is not possible, the feature must be able to be disabled at the time
```
CodeMonkeyLeet: Did you mean runtime? The previous assertion was to provide a compile time option already.

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
...
+so foreign structures should be avoided whenever possible.
+
+## Developer opt-out
+
+Because system ocalls are generally called by the SDK implementation, developers
+may not have direct control over which ocalls will be called by their application.
+A developer can refuse to import a function in EDL which will cause a linker error
+if it is used, but they should also be able to take action to resolve this failure.
+
+To this end, any OE SDK feature which makes an ocall on behalf of an enclave
```
CodeMonkeyLeet: I'm not sure I understand what the opt-out option entails here. It seems like disabling a feature at compile time would be directly equivalent to refactoring the feature so that it could be explicitly opted into. Do you have a proposed example for such a mechanism to illustrate the difference you have in mind?

It seems like instead of introducing the opt-in/opt-out dichotomy, it would be more informative to propose how we would handle existing exceptions to this model, see other comments below.

docs/DesignDocs/system_ocall_opt_in.md
jhand2: I don't know, I added them with hopes that someone would tell me they can be in EDL :smile:

I can open a separate issue to move them to edl. I already did some work to move sleep and get_time to edl.

docs/DesignDocs/system_ocall_opt_in.md
jhand2: Technically attestation is optional though right? It might be a future work item but I think we could refactor to make attestation APIs opt in.

I'll look into the cpuid table, maybe this one actually is required. If it is the only one, we could move it to a separate EDL file and always include it (I don't want to force users to add an edl include line if they effectively have no choice).

docs/DesignDocs/system_ocall_opt_in.md
jhand2: I had in mind a couple scenarios:

1. I don't want to include a specific syscall, but I still want the rest of libc to work. There should be a mechanism for this.
2. Some syscall is so core that if it isn't included none of libc will work. In this case I should be able to just disable libc.

One of these types of opt-out (or both) should be available for any feature which uses ocalls.

jhand2: I guess these are just two ways of achieving the same goal. I'll deduplicate.

docs/DesignDocs/system_ocall_opt_in.md
jhand2: Good point, even if it isn't implemented in EDL, it may still be possible to opt out of this. Maybe providing a compile-time way to disable switchless ocalls would be sufficient.

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
...
+of OE SDK compilation.
+
+## Exceptions to edl import and opt-in/out mechanisms
+
+A small number of OCalls in OE SDK are not implemented in edl and cannot be for
+technical reasons. These OCalls will always be included and the developer cannot
+opt out of them because they support core SDK functionality.
+
+These calls are:
+* `OE_OCALL_MALLOC`
```
jhand2: If ocalls are disabled altoghther, OE_OCALL_MALLOC and OE_OCALL_FREE could maybe be opted out of.

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
...
+
+In order to support ocall opt-in, system EDL files will now be installed as part of the SDK at
+`${install_path}/include/openenclave/edl/`. This allows them to be imported into user edl by the
+enclave developer.
+
+## Developer opt-in
+
+In order to use any system OCall, it must be imported into the enclave's EDL file. If it is not
+included, but the OCall is still called by the enclave application or some part of the SDK,
+enclave linking will fail.
```
dthaler: Why linking rather than making compiling fail?  (When we did this in the new_platforms branch last year, I believe it was the compile step that would fail, because the prototype for the called API would not be found)

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,120 @@
...
+* `OE_OCALL_THREAD_WAKE`
+* `OE_OCALL_WAKE_HOST_WORKER`
+
+## List of system OCalls
+
+All system OCalls (other than the previously mentioned exceptions) are found
+in one of the following edl files
+
+* `common/tee.edl`
+* `common/syscall.edl`
```
dthaler: This coarse granularity means you might need a long list of function-level imports/exports if you want the granularity.
Why not split these into finer grained EDL files, e.g., separate out sockets, stdio, etc. (This is what we did in new_platforms)

docs/DesignDocs/system_ocall_opt_in.md
jhand2: Oh you're right, compiling would fail because the function wouldn't be present in the generated header.

I will update the doc.

docs/DesignDocs/system_ocall_opt_in.md
jhand2: Agreed, my goal is to split them out, but I haven't had a chance to do the work yet and I wanted to settle on a design first.

docs/DesignDocs/system_ocall_opt_in.md
jhand2: Actually let me take that back.

If the developer doesn't explicitly call any of these functions (as is the intention) then compilation will succeed just fine. The functions are just called by other OE libraries (such as oelibc). So in this example linking would fail when liboelibc.a is linked into the enclave binary if the proper edl file isn't imported.

dthaler: Got it, thanks

dthaler: Ok so the core issue then is that libc is monolithic (it wasn't in our new_platforms branch which is why we got compiler errors instead of linker errors).  So if you don't use any sockets APIs, it sounds like right now all the sockets apis are still in libc, making your TCB potentially larger with code that is never called by the enclave app, instead of having more granular lib files that reduce the TCB.   Are we assuming the linker will always optimize out un-called obj files out of the libc?

docs/DesignDocs/system_ocall_opt_in.md
jhand2: Opened #2585 to track this

docs/DesignDocs/system_ocall_opt_in.md
jhand2: Tracked by #2008

docs/DesignDocs/system_ocall_opt_in.md
jhand2: Tracked by #2008

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,127 @@
...
+A developer can refuse to import a function in EDL which will cause a linker error
+if it is used, but they should also be able to take action to resolve this failure.
+
+To this end, any OE SDK feature which makes an OCall on behalf of an enclave
+should provide a mechanism to disable the OCall at compile-time and fail
+gracefully.
+
+## Exceptions to edl import and opt-in/out mechanisms
+
+A small number of OCalls in OE SDK are not implemented in edl and cannot be for
```
anakrish: This is no longer true.
With #2551, a context parameter is passed to ecalls. ecall_context_t contains enough space (16 KB) to make all the core ocalls (and most user written ocalls too) without any additional ocalls for host memory allocation/deallocation.

With the above guarantee provided by the EDL mechanism,   malloc/free ocalls can be themselves be implemented in EDL since their tiny parameter size (maximum declared parameter size = sizeof(pointer) + sizeof(uint64_t) = 16 bytes) guarantees that they can be dispatched without any additional ocalls for host memory allocation.

We are pretty much at a point where *all* ocalls can be EDL generated.

jhand2: That's awesome!

CodeMonkeyLeet: We should update this doc to that effect and enumerate these as tasks rather than exceptions then.

As an extension of this, can we make a stronger statement about the developer opt-out? i.e. The design prescription is that OE SDK ocalls must be expressed as EDLs and the opt-out mechanism is always to not include the EDL, which renders this paragraph irrelevant, no?

> To this end, any OE SDK feature which makes an OCall on behalf of an enclave should provide a mechanism to disable the OCall at compile-time and fail gracefully.

The one remaining exception to this seems to be cpuid, and that's a case where there simply is no opt-out and no proposal for a mechanism to disable it with a graceful fallback mechanism. You would have to rewrite the initialization to pass the table in on initialization rather than have it call out to avoid this, and maybe that's a strategy that should be documented here.

docs/DesignDocs/system_ocall_opt_in.md
CodeMonkeyLeet: @jhand2 I don't believe this was addressed although a force push has unanchored this comment.

jhand2:
I would like to propose an amendment to the **EDL local structures** portion of this design.

cc @CodeMonkeyLeet @mikbras @anakrish @radhikaj 

# The poblem

After a couple different attempts to convert some of our EDL to "local structures only" I believe this is unnecessarily complicated/restrictive. Our EDL uses structures that are part of the public API surface and should remain in the published headers. To make this work, there needs to be a copy of each of these structures in edl and in C. This means converting back and forth between the two in code at the enclave boundary.

It also means that enclave writers cannot use structures that they've defined elsewhere in a header file as a parameter to their ecalls/ocalls. This means jumping through these same hoops for our developers.

# My proposal

An EDL annotation `[edl_struct=...]` which would denote a corresponding EDL structure to be used when marshalling/unmarshalling a C struct in EDL. So for example, take the `oe_asymmetric_key_params_t` struct.

```
// openenclave/bits/types.h
typedef struct _oe_asymmetric_key_params
{
    oe_asymmetric_key_type_t type;
    oe_asymmetric_key_format_t format;
    void* user_data;
    size_t user_data_size;
} oe_asymmetric_key_params_t;
```

```
// in foo.edl
struct oe_edl_asymmetric_key_params_t
{
    oe_asymmetric_key_type_t type;
    oe_asymmetric_key_format_t format;
    [size=user_data_size] void* user_data;
    size_t user_data_size;
};

enclave {
   trusted {
      void foo(
         [in, edl_struct=oe_edl_asymmetric_key_params_t]
            oe_asymmetric_key_params_t* params);
   }
}
```

Under the hood, the edger8r generated code would do a shallow copy between the C struct and the edl struct and then do a deep copy using the edl struct over the enclave boundary.

docs/DesignDocs/system_ocall_opt_in.md
anakrish: Good point about cpuid. With @jhand2 recent PR (215779f) to allow exception handler to make ocalls, cpuid table could be fetched  from host lazily only when a cpuid instruction is executed by an enclave.

anakrish:
@jhand2 Have you considered approaches for auto-generating EDL? With the existence of libclang and python bindings for it, it is much more easier to analyze and process C code.
@gupta-ak earlier worked on an edlgen tool where EDL was automatically generated from C header files.

I am very hesitant to expand EDL as a language; I'd rather prefer tooling atop it especially since it is quite easy to do so with LLVM/clang.

CodeMonkeyLeet:
@anakrish: 
> I am blocking this PR since it has got enough approvals to go in; but has an amendment proposal.
> 
> Should amendments, expansion of scope etc automatically cause approvals to be invalidated or at the least prevent check-in? What is the recommended open source practice in such cases?

I think it's fair to block or rescind approvals when there is significant change to a PR; there's a GitHub option to do this automatically in a PR, though I would leave this as a manual task (also an option to re-request reviews under the reviewer widget), since changing design or expected implementation should be infrequent and there are some cases of pre-approval pending trivial edits. We also have the option of undoing a merge if there are bad faith actors.

There's a broader question of what we should do about these design docs once approved, since they go out of sync with implementations or get redesigned (there's another example with #2579). This might be a topic of discussion with the CGC, and I personally would like some guidance about the transition of design proposals to a maintained developer knowledge base.

In this specific case, there's been no change made to the PR contents, but just a broader discussion, which sounds like it could be spun off into a different issue/design around an EDL feature.

CodeMonkeyLeet:
> Under the hood, the edger8r generated code would do a shallow copy between the C struct and the edl struct and then do a deep copy using the edl struct over the enclave boundary.

Wouldn't this just be an aliasing of the non-EDL type to the EDL one? It also requires a bit of thought as to when a user should use this proposed pattern (it's inefficient to do this for flat, foreign defined structs, as illustrated in the example where we don't expect to also map the nth order foreign structs `oe_asymmetric_key_type_t` etc.). 

> @jhand2 Have you considered approaches for auto-generating EDL? With the existence of libclang and python bindings for it, it is much more easier to analyze and process C code.
> @gupta-ak earlier worked on an edlgen tool where EDL was automatically generated from C header files.

From a dev experience perspective, I agree with @anakrish on this, since having a tool that can automatically generate the intermediate EDL local struct definition for marshalling is less friction than having to define structs twice and manually annotate the relationship between them. I'm unclear on the details of the proposal though, since EDL can't use the same named struct transparently.

I'll suggest a third option, which is to invert the paradigm: Use EDL to generate the headers for types that need to be marshalled, and use the generated headers elsewhere in your code. This is a common pattern in COM development, where you're going from a richer header definition (IDL in COM's case, EDL here) to a poorer one (C/C++ headers) rather than the other way around. This may require us to refactor apart some of the existing headers and the build process, but is relatively transparent to the developer otherwise. (i.e. they can already do this today)

jhand2:
> From a dev experience perspective, I agree with @anakrish on this, since having a tool that can automatically generate the intermediate EDL local struct definition for marshalling is less friction than having to define structs twice and manually annotate the relationship between them. I'm unclear on the details of the proposal though, since EDL can't use the same named struct transparently.
>

I agree with the dev experience point, but specifically using C to generate EDL doesn't really solve the problem as the main issue is lack of annotation in C structures.

> I'll suggest a third option, which is to invert the paradigm: Use EDL to generate the headers for types that need to be marshalled, and use the generated headers elsewhere in your code. This is a common pattern in COM development, where you're going from a richer header definition (IDL in COM's case, EDL here) to a poorer one (C/C++ headers) rather than the other way around. This may require us to refactor apart some of the existing headers and the build process, but is relatively transparent to the developer otherwise. (i.e. they can already do this today)

This would work and is a better experience than my initial proposal, I can't think of any objections.

For completeness, I did have one other idea: providing a way to annotate C structures with EDL markup. Something like:

```
// foo.h

struct foo {
   EDL_ANNOTATION("[in, count=data_size]") void* data;
   EDL_ANNOTATION("[in]") int data_size;
};
```

(a block comment would be ok too)

which oeedger8r could easily parse to make use of for deep copy. The major benefit of this is it allows developers to easily use their C structures in EDL without extra build configuration. But it does force the C code to be aware of EDL anyway so unless someone specifically prefers this, I think I prefer Simon's idea.

docs/DesignDocs/system_ocall_opt_in.md
dthaler: Unresolving just to get an answer to my question above "Are we assuming the linker will always optimize out un-called obj files out of the libc?"   If answer is Yes, ok to re-resolve.

jhand2: Correct, that is the assumption. Thanks

jhand2:
> I'll suggest a third option, which is to invert the paradigm: Use EDL to generate the headers for types that need to be marshalled, and use the generated headers elsewhere in your code. This is a common pattern in COM development, where you're going from a richer header definition (IDL in COM's case, EDL here) to a poorer one (C/C++ headers) rather than the other way around. This may require us to refactor apart some of the existing headers and the build process, but is relatively transparent to the developer otherwise. (i.e. they can already do this today)

To close the loop here, this is what we ended up going with. @mingweishih implemented this in https://github.com/openenclave/openenclave/pull/2691

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,152 @@
...
+C/C++ header files and included in the edl file.
+
+In the case that a foreign structure must be used, system edl files may only include headers
+that are installed with the SDK.
+
+Local EDL structures provide additional security benefits detailed [here](
+https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/full_edl_serialization.md#motivation)
+so foreign structures should be avoided whenever possible.
+
+## Developer opt-out
```
jhand2: @CodeMonkeyLeet and @jbdelcuv I have updated this section to make it more clear what it means to be able to opt-out of using these OCalls. Let me know if this makes sense

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,152 @@
...
+## EDL local structures
+
+Under this design, system edl files will prefer using local structures; that is, structures
+that are defined in edl. This is distinct from foreign structures, which are defined in
+C/C++ header files and included in the edl file.
+
+In the case that a foreign structure must be used, system edl files may only include headers
+that are installed with the SDK.
+
+Local EDL structures provide additional security benefits detailed [here](
```
dthaler: nit: avoid using "here" as link text.  See https://www.w3.org/TR/WCAG10-HTML-TECHS/#link-text or a web search on "click here style" for lots of discussion.  Short version: the main (but not only) reason is for accessibility where link readers will only read the link text and not the context and "here" is not useful for accessibility.

docs/DesignDocs/system_ocall_opt_in.md
jhand2: Updated the doc to be more clear about the concern:

> Because information from insdie the enclave may be transmitted to the untrusted host
during an OCalls, an enclave developer should be able to opt-out of these OCalls
being made. They should also have visibility into what OCalls might be called by OE SDK.

docs/DesignDocs/system_ocall_opt_in.md
jhand2: Reworked to hopefully be more clear

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,145 @@
...
+
+In order to use any system OCall, it must be imported into the enclave's EDL file. If it is not
+included, but the OCall is still called by the enclave application or some part of the SDK,
+enclave linking will fail.
+
+Users can import a system OCall the same way any other edl file would be imported. For example,
+if an enclave with edl file `sample.edl` wants to use the function `oe_log_ocall()` they can do
+one of the following:
+
+*Note: These examples assume that `${install_path}/include` is passed to edger8r as an edl search path*
```
CodeMonkeyLeet:
```suggestion
Users can import a system OCall the same way any other EDL file would be imported. For example,
if an enclave with EDL file `sample.edl` wants to use the function `oe_log_ocall()` they can do
one of the following:

*Note: These examples assume that `${install_path}/include` is passed to edger8r as an EDL search path*
```

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,145 @@
...
+`from "openenclave/edl/tee.edl" import oe_log_ocall, oe_realloc_ocall;`. In this scenario,
+example (2) will work without modification.
+
+## EDL local structures
+
+Under this design, system edl files will prefer using local structures; that is, structures
+that are defined in edl. This is distinct from foreign structures, which are defined in
+C/C++ header files and included in the edl file.
+
+In the case that a foreign structure must be used, system edl files may only include headers
```
CodeMonkeyLeet:
```suggestion
Under this design, system EDL files will prefer using local structures; that is, structures
that are defined in EDL. This is distinct from foreign structures, which are defined in
C/C++ header files and included in the EDL file.

In the case that a foreign structure must be used, system EDL files may only include headers
```

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,145 @@
...
+which calls this function, the enclave developer will be forced to import that
+function into their own EDL. To allow the developer to opt out of this OCall,
+it should be possible for the enclave developer to turn off the switchless calls
+feature such that en enclave which does not use switchless calls will not require
+this OCall. This mechanism should not require the developer to recompile the OE
+SDK.
+
+## Exceptions to edl import and opt-in/out mechanisms
+
+A small number of OCalls in OE SDK are not implemented in edl. Until these are
```
CodeMonkeyLeet:
```suggestion
## Exceptions to EDL import and opt-in/out mechanisms

A small number of OCalls in OE SDK are not implemented in EDL. Until these are
```

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,145 @@
...
+* `OE_OCALL_FREE`
+* `OE_OCALL_THREAD_WAIT`
+* `OE_OCALL_THREAD_WAKE`
+
+## List of system OCalls
+
+All system OCalls (other than the previously mentioned exceptions) are found
+in one of the following edl files
+
+TODO: Document what each edl file is for and what features may invoke them.
```
CodeMonkeyLeet:
```suggestion
All system OCalls (other than the previously mentioned exceptions) are found
in one of the following EDL files

TODO: Document what each EDL file is for and what features may invoke them.
```

CodeMonkeyLeet: If this TODO: is not going to be addressed for this design document, can we file an issue for it instead of tracking it in the design doc?

docs/DesignDocs/system_ocall_opt_in.md
CodeMonkeyLeet: `oe_get_cpuid_table_ocall()` is always invoked today as part of oecore, not oeenclave. It shouldn't block this design, but as a matter of implementation there will be methods like this that should be removed. In this specific case, we can refactor the enclave initialization function to provide the required information in the initial ECall without having to rely on an OCall, which is also preferred from a performance perspective by reducing the number of boundary crossings.

If we do identify a subset of OE runtime OCalls that must always be included, we should probably submit a separate addendum to this design regarding how it those headers are handled. For example, one way of reducing the impact of such functions is to enforce that they are only invoked in codepaths that do not interact with the application's codepaths (e.g. only during enclave init or teardown).

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,145 @@
...
+
+To this end, any OE SDK feature which makes an OCall on behalf of an enclave
+should provide a mechanism to disable the behavior which makes an OCall.
+
+For example, the switchless calls feature makes an OCall
+`oe_wake_switchless_worker_ocall()`. If there is code compiled into the enclave
+which calls this function, the enclave developer will be forced to import that
+function into their own EDL. To allow the developer to opt out of this OCall,
+it should be possible for the enclave developer to turn off the switchless calls
+feature such that en enclave which does not use switchless calls will not require
```
CodeMonkeyLeet:
```suggestion
feature such that an enclave which does not use switchless calls will not require
```

docs/DesignDocs/system_ocall_opt_in.md
jhand2: Tracked by #2719

jhand2:
@anakrish could you take another look at this when you get a chance?

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,141 @@
...
+
+## List of system OCalls
+
+All system OCalls (other than the previously mentioned exceptions) are found
+in one of the following EDL files
+
+* `common/attestation.edl`
+* `common/logging.edl`
+* `common/memory.edl`
+* `common/keys.edl`
```
mingweishih: This edl only defines ECalls, should it be removed from this list (or there may be more such cases)?

jhand2: Ah you're right. I'll go through and remove any with no ocalls

jhand2:
cc @anakrish The doc is now fully up to date with my current proposal. The extension I originally proposed is no longer valid.

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,139 @@
...
+    from "openenclave/edl/logging.edl" import oe_log_ocall;
+
+    trusted
+    {
+        public void my_sample_ecall();
+    };
+
+    untrusted
+    {
+        public void my_sample_ocall();
```
anakrish: The `public` modifier is not applicable for ocalls.

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,139 @@
...
+may not have direct control over which OCalls will be called by their application.
+A developer can refuse to import a function in EDL which will cause a linker error
+if it is used, but they should also be able to take action to resolve this failure.
+
+To this end, any OE SDK feature which makes an OCall on behalf of an enclave
+should provide a mechanism to disable the behavior which makes an OCall.
+
+For example, the switchless calls feature makes an OCall
+`oe_wake_switchless_worker_ocall()`. If there is code compiled into the enclave
+which calls this function, the enclave developer will be forced to import that
```
anakrish:
```suggestion
which calls this function whether switchless calls are used in the enclave or not, the enclave developer will be forced to import that
```

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -0,0 +1,139 @@
...
+* `common/fcntl.edl`
+* `common/ioctl.edl`
+* `common/poll.edl`
+* `common/signal.edl`
+* `common/socket.edl`
+* `common/time.edl`
+* `common/unistd.edl`
+* `common/utsname.edl`
+* `common/syscall.edl`
+* `common/syscall.edl`
```
anakrish: syscall.edl twice

PR #2484, opened 2020-01-30
Remove the oeedl_file cmake function

### LINKED ISSUE #2325: Stop using oeedl_file.cmake

jhand2:
The oeedl_file function is inflexible and is too large a function to justify how little it abstracts. Remove the function and instead use a cmake custom command to call edger8r for each enclave.

Fixes #2325


tests/CMakeLists.txt
```diff
@@ -1,7 +1,6 @@
 # Copyright (c) Open Enclave SDK contributors.
 # Licensed under the MIT License.
 
-include(oeedl_file)
```
radhikaj: Can this cmake module be deleted from the repo? Or does it continue to serve some purpose?

jhand2: Oh, will do. I intended to remove it but I guess I forgot :smile:

jhand2:
Got canceled for some reason

bors r+

nonpolarity:
This is a really huge project that almost all tests are related.

jhand2:
> This is a really huge project that almost all tests are related.

Not sure what you mean

nonpolarity:
> 
> 
> > This is a really huge project that almost all tests are related.
> 
> Not sure what you mean.
> 
> As for the motivation, this function has been considered deprecated for a while and people have been asked not to use it in new PRs. We also don't recommend that people use this externally.
> 
> The function is overly complicated, doesn't work in a number of cases, and hides what is actually happening for people who are new to the project. People generally look to test to figure out how they should write and compile their application. We shouldn't use internal only mechanisms for tests and the added code is very minimal.

Sorry I just mean a lot of work. Until oeedl_file is removed totally some developers may spread oeedl_file to new tests. Nice to see oeedl_file will become extinct.

jhand2:
bors cancel

jhand2:
@nonpolarity Ah I misunderstood 😄 Thanks! Not sure whats going on with bors but hopefully I can merge soon.

BRMcLaren:
bors try

BRMcLaren:
The B.O.R.S try was in regards to a potential issue with our CI/CD. Double checked, seemed to have been an infra timeout and nothing to be done. Please let me know if it happens again.

jhand2:
Thanks @BRMcLaren 

bors r+

andschwa:
Wow thanks so much @jhand2!

PR #2485, opened 2020-01-30
Complete the struct initializer to fix #2080

### LINKED ISSUE #2080: Uninitialized field warning with clang using OE_SET_ENCLAVE_SGX
```
props.cc:7:1: error: missing field 'oeinfo_size' initializer [-Werror,-Wmissing-field-initializers]
OE_SET_ENCLAVE_SGX(
^
/opt/openenclave/include/openenclave/bits/sgx/sgxproperties.h:132:9: note: expanded from macro 'OE_SET_ENCLAVE_SGX'
```
...when compiling with clang and `-Wall -Wextra -Werror`

Current work-around: `-Wno-error=missing-field-initializers`.

ryanhsu19:
Complete the struct initializer to remove the warning, `-Wmissing-field-initializers`, which is mentioned in #2080. Scan-build does not give such warnings afterwards.

### DIFF
--- a/include/openenclave/bits/sgx/sgxproperties.h
+++ b/include/openenclave/bits/sgx/sgxproperties.h
@@ -128,7 +128,7 @@ typedef struct _oe_sgx_enclave_properties
         },                                                                \
         .image_info =                                                     \
         {                                                                 \
-            0                                                             \
+            0, 0, 0, 0, 0, 0                                              \
         },                                                                \
         .sigstruct =                                                      \
         {                                                                 \


jhand2:
So basically this warning happens because clang/gcc aren't happy with the C standard.

From the C99 standard section "Initialization - Sematics":
> If there are fewer initializers in a brace-enclosed list than there are elements or members of an
aggregate, or fewer characters in a string literal used to initialize an array of known size than
there are elements in the array, the remainder of the aggregate shall be initialized implicitly
the same as objects that have static storage duration.

To resolve this warning, we just compile with `-Wno-missing-field-initializers`. This is a perfectly legitimate solution IMO, in which case would could just close 2080.

Now the other argument would be that we are an SDK and our goal should be to support the default behavior of most compilers as closely as possible. In this case, you could argue that we should do what gcc/clang say. If we go this route, we should open an issue to remove `-Wno-missing-field-initializers` and fix every case of this. The downside of doing this is that it's very tedious to initialize every field of a large struct to 0.

Honestly I'm not sure which I prefer.

jhand2:
Oh actually this issue is for something in a public header. In that case I agree that we should explicitly initialize all fields 😄

ryanhsu19:
I see, thanks!

PR #2490, opened 2020-01-31
Split out libc tests into one enclave binary per test

### LINKED ISSUE #2487: Run libc tests separately
Currently each libc test is run as a single test. This makes it very challenging to troubleshoot which specific test is failing. Additionally, if one test fails, the others are not even attempted.

Match the libcxx tests in putting each test in a separate enclave binary and running them as separate tests.

jhand2:
Previously, the libc tests were all run as part of a single enclave binary. Additionally, they would fail after a single test, meaning other tests were not run after the first failure.

Update the libc tests so that each test in libc-test is run as a separate test in cmake.

Note that this does have a perf impact (both at build time and test running time) but I still think it's the right thing to do.

Fixes #2487


tests/libc/CMakeLists.txt
```diff
@@ -7,4 +7,9 @@ if (BUILD_ENCLAVES)
     add_subdirectory(enc)
 endif()
 
-add_enclave_test(tests/libc libc_host libc_enc)
+include(./tests.cmake)
+
+foreach(testcase ${LIBC_TESTS})
+    get_testcase_name(${testcase} name "../../3rdparty/libcxx/libcxx/test/")
```
jhand2: this prefix should be to the musl tests

jhand2:
bors try

jhand2:
None of these build errors make any sense. Trying again

bors try

radhikaj:
@jhand2 When a developer executes ctest after building the SDK, how much longer does it take after you split tests up?

jhand2:
> @jhand2 When a developer executes ctest after building the SDK, how much longer does it take after you split tests up?

I ran just one pass for each of these benchmarks so there could be some variation but probably not much.

In simulation mode:
**Before**: 129.90 seconds
**After**: 149.96 seconds

On SGX hardware:
**Before**: 296.74
**After**: 308.41 seconds

So about 20 seconds slower on both. Not great (but you feel the difference more in simulation mode). But before this change debugging a libc test failure is very challenging

jhand2:
bors try

jhand2:
@radhikaj @CodeMonkeyLeet @anakrish Do you see the increase in elapsed time to run tests (+20 seconds) as to great to justify this change? Or do you think the debugging benefit is worth it?

For reference, each libc test takes 0.05 seconds in simulation mode, 0.1 seconds on SGX. It just adds so much time because there are so many of them.

cmake/add_enclave.cmake
```diff
@@ -164,9 +164,11 @@ macro(add_enclave_optee)
...
   set(CMAKE_EXE_EXPORTS_C_FLAG)
 
-  string(REPLACE "gcc" "ld" LINKER ${CMAKE_C_COMPILER})
-  set(CMAKE_C_LINK_EXECUTABLE "${LINKER} <CMAKE_CXX_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> <LINK_LIBRARIES> -lgcc -o <TARGET>")
-  set(CMAKE_CXX_LINK_EXECUTABLE "${LINKER} <CMAKE_CXX_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> <LINK_LIBRARIES> -lgcc -o <TARGET>")
+  string(REPLACE "gcc" "ld" C_LINKER ${CMAKE_C_COMPILER})
+  string(REPLACE "g++" "ld" CXX_LINKER ${CMAKE_CXX_COMPILER})
+
+  set(CMAKE_C_LINK_EXECUTABLE "${C_LINKER} <CMAKE_C_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> <LINK_LIBRARIES> -lgcc -o <TARGET>")
+  set(CMAKE_CXX_LINK_EXECUTABLE "${CXX_LINKER} <CMAKE_CXX_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> <LINK_LIBRARIES> -lgcc -o <TARGET>")
```
jhand2: Just realized this change is kinda pointless. The only necessary change was changing CMAKE_CXX_LINK_FLAGS to CMAKE_C_LINK_FLAGS on line 170.

I'll revert once bors finishes

tests/libc/enc/CMakeLists.txt
```diff
@@ -5,42 +5,31 @@
...
-    ${LIBC_TESTS}
-    ${gen})
+    ${gen}
+    ${test_extras})
 
-target_compile_options(libc_enc PRIVATE
+target_include_directories(
+    libctest-support PRIVATE
+    ${CMAKE_CURRENT_BINARY_DIR}
+    ../../../3rdparty/musl/libc-test/src/common)
```
CodeMonkeyLeet: Replace directory traversals with `${PROJECT_SOURCE_DIR}` here and line 93 as well?

anakrish:
@jhand2 Please do follow up with @mikbras. If memory serves right, libC tests used to take a lot of time due to there being a large number of tests, and Mike optimized the running time by combining them.

jhand2:
I think I actually fixed the build problem this time

bors try

tests/libc/enc/CMakeLists.txt
jhand2: Good point, a build toggle is probably the right way to go.

jhand2:
> Regarding the question of the additional runtime overhead, have we considered making a build/ctest toggle? There's precedence here for this in how we choose to run a subset of libcxx tests in the default case and a nightly run of CI/CD runs the full set. Similarly, if the expectation is that these usually pass and do not need devs to investigate failures in their inner loop regularly, it's useful to default to the previous behavior where it's treated as a single ctest, with the option to split them into separate ctests to enable deeper debugging.

Good point, a build toggle is probably the right way to go.

jhand2:
I have added a compile option. Now if you run cmake with `-DOE_SEPARATE_LIBC_TESTS`, each libc test will be run as a separate enclave.

bors try

jhand2:
Fixed merge conflict

bors try

jhand2:
Closing this PR in favor of  #2501

tests/libc/enc/CMakeLists.txt
```diff
@@ -62,51 +41,115 @@ target_compile_options(libc_enc PRIVATE
...
+        libc_t.c)
+
+    target_compile_options(libc_enc PRIVATE ${LIBC_COMPILE_OPTIONS})
+    target_include_directories(libc_enc PRIVATE ${LIBC_INCLUDE_DIRECTORIES})
+    target_link_libraries(libc_enc ${LIBC_ENC_REQUIRED_LIBRARIES})
+endif ()
+
+# On windows, build the enclave support library with clangw
+if (WIN32)
+    maybe_build_using_clangw(libunwindtest-${NAME}_enc)
```
mingweishih: I guess we can remove this if logic as `add_enclave` should handle the `clangw` configuration (also the `libunwindtest-${NAME}_enc` is incorrect).

tests/libc/enc/helpers.c
```diff
@@ -0,0 +1,164 @@
...
+
+// Space for global libc environment
+static char** environ = NULL;
+int t_status = 0;
+
+int device_init()
+{
+    OE_TEST(oe_load_module_host_file_system() == OE_OK);
+    OE_TEST(oe_load_module_host_socket_interface() == OE_OK);
+    // OE_TEST(oe_load_module_host_epoll() == OE_OK);
```
mingweishih: Should we just remove this commented check (or do we know why we need to keep that)?

jhand2:
> Thanks for making this! LGTM with only minor comments.

Sorry, I took a while to close this. https://github.com/openenclave/openenclave/pull/2501 replaces this PR

PR #2495, opened 2020-01-31
Optimize switchless ocall scheduling

anakrish:
Switchless worker thread used to clear its slot before handling the
switchless ocall. This caused another switchless ocall to be scheduled
on the worker thread, even before it could complete the current ocall.

In order to optimize scheduling, the work thread now first handles the
switchless ocall and then clears its slot. This causes the thread to
be considered "busy" till the ocall is handled. While the thread is
"busy", switchless ocalls are scheduled in other available worker threads.

Test cleanup:
tests/switchless has been tweaked to measure and display the speed up
factor of fastest and slowest enclave threads making switchless calls.
Previously, only the fastest time as displayed.

### DIFF
--- a/host/sgx/switchless.c
+++ b/host/sgx/switchless.c
@@ -29,10 +29,16 @@ static void* _switchless_ocall_worker(void* arg)
         volatile oe_call_host_function_args_t* local_call_arg = NULL;
         if ((local_call_arg = context->call_arg) != NULL)
         {
-            context->call_arg = NULL;
-
+            // Handle the switchless call, but do not clear the slot yet. Since
+            // the slot is not empty, any new incoming switchless call request
+            // will be scheduled in another available work thread and get
+            // handled immediately.
             oe_handle_call_host_function(
                 (uint64_t)local_call_arg, context->enclave);
+
+            // After handling the switchless call, mark this worker thread
+            // as free by clearing the slot.
+            context->call_arg = NULL;
 
             // Reset spin count for next message.
             context->total_spin_count += context->spin_count;
--- a/tests/switchless/host/host.c
+++ b/tests/switchless/host/host.c
@@ -135,6 +135,14 @@ int main(int argc, const char* argv[])
     if (argc == 4)
     {
         sscanf(argv[3], "%" SCNu64, &num_enclave_threads);
+        if (num_enclave_threads > NUM_TCS)
+        {
+            fprintf(
+                stderr,
+                "Number of enclave threads must be less than %d\n",
+                (int)NUM_TCS);
+            return 1;
+        }
     }
 
 #if defined(__WIN32)
@@ -162,7 +170,7 @@ int main(int argc, const char* argv[])
 
     // Measure switchless ocall performance.
     uint64_t num_extra_enc_threads = num_enclave_threads - 1;
-    thread_info_t tinfo[32];
+    thread_info_t tinfo[NUM_TCS];
     for (uint64_t i = 0; i < num_extra_enc_threads; ++i)
     {
         int ret = 0;
--- a/tests/switchless/host/host.c
+++ b/tests/switchless/host/host.c
@@ -92,12 +92,21 @@ double make_repeated_switchless_ocalls(oe_enclave_t* enclave)
         "%d switchless calls took %d msecs.\n",
         NUM_OCALLS,
         (int)(switchless_microseconds / 1000.0));
+
     return switchless_microseconds;
 }
 
-void* launch_enclave_thread(void* e)
+typedef struct
 {
-    make_repeated_switchless_ocalls((oe_enclave_t*)e);
+    oe_thread_t tid;
+    oe_enclave_t* enclave;
+    double elapsed;
+} thread_info_t;
+
+void* launch_enclave_thread(void* a)
+{
+    thread_info_t* info = (thread_info_t*)a;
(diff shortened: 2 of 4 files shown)


anakrish:
bors try

tests/switchless/host/host.c
```diff
@@ -151,26 +160,47 @@ int main(int argc, const char* argv[])
              &enclave)) != OE_OK)
         oe_put_err("oe_create_enclave(): result=%u", result);
 
-    char out[STRING_LEN];
-    int return_val;
-
+    // Measure switchless ocall performance.
     uint64_t num_extra_enc_threads = num_enclave_threads - 1;
-    oe_thread_t tid[32] = {0};
+    thread_info_t tinfo[32];
```
mingweishih: Do we check `num_extra_enc_threads` against `32` anywhere?

anakrish: I have defined NUM_TCS=32 in switchless.edl and used that in both host and enclave.

anakrish:
bors try

host/sgx/switchless.c
```diff
@@ -29,11 +29,17 @@ static void* _switchless_ocall_worker(void* arg)
         volatile oe_call_host_function_args_t* local_call_arg = NULL;
         if ((local_call_arg = context->call_arg) != NULL)
         {
-            context->call_arg = NULL;
```
jxyang: Help me understand this: the performance is improved because we avoid putting switchless calls on a thread that could stall on the current switchless call waiting to be finished, right? On the other hand, I image this would increase the chance that a switchless call has to fall back to the regular call. 

In this test case, there is only one work thread. So your change would guarantee the switchless call must fall back to regular call if the work thread is handling another switchless call. Are you suggesting the fall back actually improved performance?

anakrish: > 
> 
> Help me understand this: the performance is improved because we avoid putting switchless calls on a thread that could stall on the current switchless call waiting to be finished, right?
>

Yes. The non-optimality of the current scheduling is that no matter the number of host workers, enclave threads will stall since calls will be scheduled on a worker threads before it has finished handling the previous call.

**There is no way for the user to over come this and the user cannot achieve maximum throughput.**

>
>On the other hand, I image this would increase the chance that a switchless call has to fall back to the regular call.
> 

Since our switchless ocalls are blocking, **fallback will never happen when the number of host worker threads >= number of enclave threads making ecalls**. In this case, there will always be a host worker available.

With the new scheduling, in the case where number of worker threads < number of enclave threads making ecalls, fallback can happen. However, fallback still allows the system to make progress whereas stalling does not.

Note that if number of enclave threads > 2* worker threads, then fallback will happen in the current implementation as well.

>
>In this test case, there is only one work thread. 
>
This test allows passing in different number of host and enclave threads. 

>
> Are you suggesting the fall back actually improved performance?
>
No. Not stalling improved performance.

anakrish:
Merging this in to allow other optimization PRs.

bors r+

host/sgx/switchless.c
jxyang: That's my intuition: this PR will make fallback happens more often. Imaging a switchless call comes in, and find all work threads are occupied, it either 1) wait a while on the first thread it poked, or 2) fall back to the regular call. If I remember correctly, the regular call is 5-10 times slower than switchless call. So I'd not automatically conclude 2) is better than 1). 

I'd be more convinced with a print out of the speed up under difference scenarios, especially when enclave thread# > worker thread#.

jxyang: BTW, I think the problem with the non-optimal current scheduler is that we always check with the first work thread, if it's busy, then go on to the 2nd one, and so on... If we "load balance" somehow, for example, this time starts with #1 thread and continue forward, next time starts with the last thread and go backward, the performance should be better.

anakrish: @jxyang 
For my understanding:
- Is the current scheduling strategy of stalling instead of using another available worker explicitly chosen to minimize fallback? 
- Fallback will happen only if number of worker threads is less than number of enclave threads. Do we anticipate that for using the switchless calls feature, the enclave author will use less worker threads than enclave threads?

>
>  If we "load balance" somehow, for example, this time starts with #1 thread and continue forward, next time starts with the last thread and go backward, the performance should be better.
>

I am not sure I understand your proposal. Can you suggest a complete algorithm?

anakrish: Consider a simple scenario with 2 worker threads and 2 enclave threads

Here is what can happen with current scheduling:
1. enc-thread1 schedules call in worker1
2. enc-thread2 chooses worker1 and stalls waiting for worker1 to complete
3. worker1 completes and starts to process enc-thread2 call
4. enc-thread1 chooses worker 1 and stalls waiting for worker 1 to complete
5. worker1 completes and starts process enc-thread1 call
6. enc-thread2 chooses worker 1 and stalls waiting for worker 1 to complete
...

worker-2 is never used.

Is there are better way to leverage worker-2 than what is being done in this PR?

anakrish:
bors try

jhand2:
@BRMcLaren is there a way we can unstick a PR from being in the bors status "Pending - Waiting in queue"?

BRMcLaren:
Bors try

anakrish:
Resubmitted changes in PR #2584

BRMcLaren:
Bors merge

Issue #892, opened 2018-10-10
Add TEE-provided PRNG implementation to oe_random


CodeMonkeyLeet:
Per #888, there is a request for `sgx_read_rand` parity functionality provided inside an enclave for use by an enclave app. This issue tracks implementing the `oe_random` function to provide that directly instead of using an indirection to enclave crypto library as it does today.

dthaler:
To be clear, I think this issue does not track any change to the API surface, but rather changing the implementation under the public API surface tracked by issue #832

CodeMonkeyLeet:
Renamed for clarity.

radhikaj:
@mingweishih Please follow up with @CodeMonkeyLeet for this issue.

mingweishih:
Fixed by #2497.

PR #2497, opened 2020-02-03
Implement TEE-supported RNG for oe_random_internal

mingweishih:
This PR addresses #892; i.e., re-implementing the `oe_random_internal` based on the hardware-based source of RNG (e.g., `rdrand` in the case of sgx). Therefore, the `oe_random` no longer depends on the third-party library (`mbedtls`) as it used to do. 

Because of this, the PR also moves the `random.c` to `enclave/core`. In addition, the PR creates `random_internal.c` in both `enclave/core/sgx` and `enclave/core/optee`.

Update: This PR fixes both #892 and #2498


mingweishih:
bors try

jhand2:
Additionally, this will throw a hw exception on CPUs without the rdrand instruction. If we're ok with that, we should be explicit that we don't support CPUs with SGX but no rdrand

enclave/core/sgx/random_internal.c
```diff
@@ -0,0 +1,25 @@
...
+oe_result_t oe_random_internal(void* data, size_t size)
+{
+    for (size_t i = 0; i < size; i += RDRAND_BYTES)
+    {
+        size_t request_size = size - i;
+        if (request_size > RDRAND_BYTES)
+        {
+            request_size = RDRAND_BYTES;
+        }
+        uint64_t random_bytes = oe_rdrand();
```
anakrish: The previous implementation of oe_random_internal called `mbedtls_ctr_drbg_random`

That function has a more complex implementation: https://github.com/openenclave/openenclave/blob/040beb10dba521ee95af10b0898707e61cceb3b6/3rdparty/mbedtls/mbedtls/library/ctr_drbg.c#L444

Is there any cryptographic consequence of using one implementation vs the other?

enclave/core/sgx/random_internal.c
```diff
@@ -0,0 +1,25 @@
...
+        memcpy((void*)((uint8_t*)data + i), (void*)&random_bytes, request_size);
+    }
+
+    return OE_OK;
+}
```
CodeMonkeyLeet: Nit: Missing newline

enclave/core/sgx/random_internal.c
CodeMonkeyLeet: @anakrish Yes, but that is the intent of #892: to provide a public enclave API that gets access to trusted HW randomness that is not mediated by the enclave runtime. Internally, the oeenclave functionality still uses the mbedtls ctr_drbg.

enclave/core/optee/random_internal.c
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/enclave.h>
+
+oe_result_t oe_random_internal(void* data, size_t size)
+{
+    OE_UNUSED(data);
```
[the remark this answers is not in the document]
HernanGatta: That is correct. There are already calls to `TEE_GenerateRandom` in `enclave/core/optee/entropy.c` (`oe_get_entropy`) and `libc/optee/arc4random.c` (`arc4random`).

mingweishih:
> Same comment as Jordan.
> 
> All you need is a call to `TEE_GenerateRandom(data, size)` and to convert the resulting `TEE_Result` into an `oe_result_t`.

@HernanGatta I didn't include the PR for optee mainly because of I don't have the environment setup yet. Would you mind if you can help with that part (#2498)?

HernanGatta:
> @HernanGatta I didn't include the PR for optee mainly because of I don't have the environment setup yet.

Makes sense, I'm so used to using the emulator I didn't even think of it. Watch out for #2527, I'll be writing up some documentation as a follow-up to that PR so everyone can use an emulator to test on TrustZone locally.

> Would you mind if you can help with that part (#2498)?

Sure, I'll branch off your branch, implement and test the functionality, then create a PR into your branch. That way, SGX and TrustZone support for this change get merged at the same time.

mingweishih:
> @mingweishih
> 
> > I didn't include the PR for optee mainly because of I don't have the environment setup yet.
> 
> Makes sense, I'm so used to using the emulator I didn't even think of it. Watch out for #2527, I'll be writing up some documentation as a follow-up to that PR so everyone can use an emulator to test on TrustZone locally.
> 

That will be awesome. Looking forward to it!

> > Would you mind if you can help with that part (#2498)?
> 
> Sure, I'll branch off your branch, implement and test the functionality, then create a PR into your branch. That way, SGX and TrustZone support for this change get merged at the same time.

That should work. Thanks so much!

HernanGatta:
@mingweishih 
> That should work. Thanks so much!

[Here you go](https://github.com/mingweishih/openenclave/pull/1).

mingweishih:
> @mingweishih
> 
> > That should work. Thanks so much!
> 
> [Here you go](https://github.com/mingweishih/openenclave/pull/1).
> 
> Edit: I mentioned needing to check the result of `TEE_GenerateRandom`. Turns out, I was wrong, the function is `void`. In addition, its implementation is such that if reading random bytes fails, the enclave panics.

In fact, the case of sgx is similar (i.e., the rdtsc instruction does not report the case of failure).

mingweishih:
bors try

Issue #2174, opened 2019-09-12
Convert `oe_handle_init_switchless` into an EDL defined function


jxyang:
Per @anakrish's comments:

"Consider making this an EDL defined function. That would reduce the number of security checks needed to be coded explicitly.
We have actively tried to minimize the number of hand-written edge functions."

PR #2499, opened 2020-02-04
Convert `oe_handle_init_switchless` into an EDL defined function

mingweishih:
This PR addresses #2174, converting the hand-written ecall of `oe_handle_init_switchless` into EDL-based one.

### DIFF
--- a/common/tee.edl
+++ b/common/tee.edl
@@ -43,6 +43,10 @@ enclave
             [out, size=key_info_size] void* key_info,
             size_t key_info_size,
             [out] size_t* key_info_size_out);
+
+        public oe_result_t oe_handle_init_switchless(
+            [user_check] void* host_worker_contexts,
+            size_t num_host_workers);
     };
 
     untrusted
--- a/enclave/core/CMakeLists.txt
+++ b/enclave/core/CMakeLists.txt
@@ -66,6 +66,7 @@ if (OE_SGX)
         sgx/report.c
         sgx/sched_yield.c
         sgx/spinlock.c
+        sgx/switchlesscalls.c
         sgx/td.c
         sgx/td_basic.c
         sgx/thread.c
@@ -107,6 +108,7 @@ elseif (OE_TRUSTZONE)
         optee/sched_yield.c
         optee/spinlock.c
         optee/stubs.c
+        optee/switchlesscalls.c
         optee/thread.c
         optee/tracee.c)
 endif()
@@ -148,7 +150,6 @@ add_library(oecore STATIC
     string.c
     strtok_r.c
     strtoul.c
-    switchlesscalls.c
     tee_t_wrapper.c
     time.c
     tracee.c
--- a/enclave/core/optee/gp.c
+++ b/enclave/core/optee/gp.c
@@ -476,12 +476,6 @@ TEE_Result TA_InvokeCommandEntryPoint(
             result = TEE_ERROR_BAD_STATE;
             break;
         }
-        case OE_ECALL_INIT_CONTEXT_SWITCHLESS:
-        {
-            /* TODO: initialize switchless calls */
-            result = TEE_ERROR_NOT_IMPLEMENTED;
-            break;
-        }
         default:
         {
             /* No function found with the number */
--- a/enclave/core/optee/switchlesscalls.c
+++ b/enclave/core/optee/switchlesscalls.c
@@ -1,0 +1,47 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+// TODO: Switchless calls for op-tee
+
+#include "../switchlesscalls.h"
+#include <openenclave/edger8r/enclave.h>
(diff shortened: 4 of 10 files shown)


mingweishih:
bors try

common/tee.edl
```diff
@@ -15,6 +15,7 @@
 enclave
 {
     include "openenclave/bits/types.h"
+    include "openenclave/internal/switchless.h"
```
jhand2: If we're going to make this an edl-defined ecall then the function should be in a public header.

common/tee.edl
```diff
@@ -43,6 +44,9 @@ enclave
             [out, size=key_info_size] void* key_info,
             size_t key_info_size,
             [out] size_t* key_info_size_out);
+
+        public oe_result_t oe_handle_init_switchless(
+            [in] oe_switchless_call_manager_t* manager);
```
anakrish: Passing it by pointer would imply count=1, and the struct would be copied into the enclave.
Consider either passing by value or avoiding foreign structures and pass necessary values are parameters.

mingweishih:
bors try

common/tee.edl
mingweishih: Resolved.

common/tee.edl
mingweishih: Resolved.

common/tee.edl
```diff
@@ -43,6 +43,10 @@ enclave
             [out, size=key_info_size] void* key_info,
             size_t key_info_size,
             [out] size_t* key_info_size_out);
+
+        public oe_result_t oe_handle_init_switchless(
```
anakrish: Thanks for cleaning this up.

enclave/core/switchlesscalls.c
```diff
@@ -37,40 +38,28 @@ bool oe_is_switchless_initialized()
 **
 **==============================================================================
 */
-oe_result_t oe_handle_init_switchless(uint64_t arg_in)
+oe_result_t oe_handle_init_switchless(
```
HernanGatta: It seems to me that this function should be TEE-specific.

On OP-TEE, there is more work to do to ensure that the buffer used to signal OCALL threads is backed by the same physical memory. More specifically, checks must be performed to ensure that shared memory is available on the system and that *it* is used for the worker contexts instead of them being mediated by copies.

Also, by creating different implementations of the initialization routine, you can make the OP-TEE one return an error for now.

dthaler: +1

enclave/core/switchlesscalls.c
```diff
@@ -37,40 +38,28 @@ bool oe_is_switchless_initialized()
...
     }
 
-    /* lfence after checks. */
+    // lfence after checks.
     oe_lfence();
 
-    // Copy the worker context array pointer and its size to avoid TOCTOU
-    _host_worker_count = safe_manager.num_host_workers;
-    _host_worker_contexts = safe_manager.host_worker_contexts;
+    // Copy the worker context array pointer and its size to avoid TOCTOU.
```
CodeMonkeyLeet: This comment doesn't really make sense anymore; it's really just persisting the pointer value globally. TOCTOU should be handled by the copy into enclave during EDL generated stub already.

enclave/core/switchlesscalls.c
CodeMonkeyLeet: Sort of an inherent problem in the original implementation, but it's a little bit more obvious now that this is a function in the publicly inspectable EDL: this initialization is only expected to happen once per enclave initialization from the host internally as part of `oe_create_enclave()`  but there's nothing preventing it from being invoked elsewhere.

Particularly since there are no synchronization guards around the globals this method initializes, the enclave side of this needs to enforce `oe_once()` semantics around them, as happens in `oe_initialize_enclave()`

jhand2: I added synchronization to prevent this here: https://github.com/openenclave/openenclave/pull/2433

@mingweishih can you rebase against master?

mingweishih: Thanks @jhand2 for point that out. I'm also aware of that patch (which introduces conflict to this PR). I'll rebase the PR.

mingweishih:
bors try

jhand2:
Great work! Thanks

bors r+

jhand2:
@dthaler could you take another look when you get a chance? Thanks!

anakrish:
@mingweishih I had to convert all of switchless calling core functions to EDL in #2609. I believe I have addresses @HernanGatta's comments too by making switchless an SGX specific functionality (that can be revisited when OPTEE also adds switchless calling mechanism). Kindly take a look and see if this PR is still needed.

anakrish:
Since #2609 has been merged into master, I am closing this PR that has not had any activity.

mingweishih:
> Since #2609 has been merged into master, I am closing this PR that has not had any activity.

Thanks @anakrish !

PR #2501, opened 2020-02-04
Allow libc tests to be run individually

### LINKED ISSUE #2487: Run libc tests separately

jhand2:
The libc tests currently run all as a single test and if one fails then the tests stop running. Add the cmake option OE_SEPARATE_LIBC_TESTS which will lead ctest to run each libc test as a separate case.

Additionally, add a command-line option to the libc test host to allow a single test to be run at a time to support the cmake flag.

Fixes #2487

This is v2 of an earlier PR #2490


jhand2:
bors try

jhand2:
bors try

BRMcLaren:
bors try

jhand2:
try again to resolve platform error

bors try

jhand2:
Potential infra issues

bors try

jhand2:
trying on new infra

bors try

jhand2:
bors try

jhand2:
> LGTM
> 
> Before merging this in master, run the full C/C++ testsuite (we only run a small subset by default) and make sure that there is no build/test time regression.

I turned on ENABLE_FILL_LIBCXX_TESTS and confirmed no regressions.

bors r+

PR #2502, opened 2020-02-05
OP-TEE Build Fixes

HernanGatta:
Up to this point, no-one has attempted to build a DEB package from a build of the SDK targeting OP-TEE on ARM TrustZone. Seeing as I am working to enable the VS Code extension for Open Enclave to work against the `master` branch, using the CMake configuration files that the packaging process generates became desirable. This way, using CMake, one can target Intel SGX and both OP-TEE / ARM TrustZone platforms without needing to make any code changes anywhere.

This PR fixes a few mistakes in the files that configure the build for OP-TEE on ARM TrustZone. More specifically, it ensures that the resulting binary outputs and CMake configuration files are coherent.

### DIFF
--- a/3rdparty/optee/libutee/CMakeLists.txt
+++ b/3rdparty/optee/libutee/CMakeLists.txt
@@ -6,8 +6,7 @@ add_library(oeuteeasm
 set_property(TARGET oeuteeasm PROPERTY C_STANDARD 99)
 target_compile_options(oeuteeasm PRIVATE
   ${OE_TZ_TA_S_FLAGS}
-  $<BUILD_INTERFACE:-include ${OE_TZ_TA_DEV_KIT_CONF}>
-  $<INSTALL_INTERFACE:-include openenclave/3rdparty/optee/conf.h>)
+  -include ${OE_TZ_TA_DEV_KIT_CONF})
 target_include_directories(oeuteeasm PRIVATE
   ${OE_TZ_LIBUTEE_INC}
   ${OE_TZ_LIBUTILS_EXT_INC})
--- a/3rdparty/optee/libutee/CMakeLists.txt
+++ b/3rdparty/optee/libutee/CMakeLists.txt
@@ -6,7 +6,8 @@ add_library(oeuteeasm
 set_property(TARGET oeuteeasm PROPERTY C_STANDARD 99)
 target_compile_options(oeuteeasm PRIVATE
   ${OE_TZ_TA_S_FLAGS}
-  -include ${OE_TZ_TA_DEV_KIT_CONF})
+  $<BUILD_INTERFACE:-include ${OE_TZ_TA_DEV_KIT_CONF}>
+  $<INSTALL_INTERFACE:-include openenclave/3rdparty/optee/conf.h>)
 target_include_directories(oeuteeasm PRIVATE
   ${OE_TZ_LIBUTEE_INC}
   ${OE_TZ_LIBUTILS_EXT_INC})
@@ -32,7 +33,8 @@ add_library(oeutee
   ${OE_TZ_LIBUTILS_EXT_SRC}/consttime_memcmp.c)
 set_property(TARGET oeutee PROPERTY C_STANDARD 99)
 target_compile_options(oeutee PUBLIC
-  -include ${OE_TZ_TA_DEV_KIT_CONF})
+  $<BUILD_INTERFACE:-include ${OE_TZ_TA_DEV_KIT_CONF}>
+  $<INSTALL_INTERFACE:-include openenclave/3rdparty/optee/conf.h>)
 target_include_directories(oeutee PRIVATE
   ${OE_TZ_LIBUTEE_SRC}
   ${OE_TZ_LIBUTEE_INC}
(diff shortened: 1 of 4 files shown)


3rdparty/optee/libutee/CMakeLists.txt
```diff
@@ -6,7 +6,8 @@ add_library(oeuteeasm
 set_property(TARGET oeuteeasm PROPERTY C_STANDARD 99)
 target_compile_options(oeuteeasm PRIVATE
   ${OE_TZ_TA_S_FLAGS}
-  -include ${OE_TZ_TA_DEV_KIT_CONF})
+  $<BUILD_INTERFACE:-include ${OE_TZ_TA_DEV_KIT_CONF}>
+  $<INSTALL_INTERFACE:-include openenclave/3rdparty/optee/conf.h>)
```
jhand2: These are private compile flags and will only be used when compiling `oeuteeasm`. I don't understand why an INSTALL_INTERFACE is necessary

3rdparty/optee/libutee/CMakeLists.txt
```diff
@@ -32,7 +33,8 @@ add_library(oeutee
   ${OE_TZ_LIBUTILS_EXT_SRC}/consttime_memcmp.c)
 set_property(TARGET oeutee PROPERTY C_STANDARD 99)
 target_compile_options(oeutee PUBLIC
-  -include ${OE_TZ_TA_DEV_KIT_CONF})
+  $<BUILD_INTERFACE:-include ${OE_TZ_TA_DEV_KIT_CONF}>
+  $<INSTALL_INTERFACE:-include openenclave/3rdparty/optee/conf.h>)
```
jhand2: Can you help me understand what this is doing? Why is it being included at the compiler command-line by enclave developers?

3rdparty/optee/libutee/CMakeLists.txt
HernanGatta: Mm, good point. Perhaps only the interface at the `oeutee` level is enough.

3rdparty/optee/libutee/CMakeLists.txt
HernanGatta: Building enclaves for OP-TEE without the Open Enclave SDK works like this:

1. Choose the target platform (i.e., the physical board);
1. Choose OP-TEE features and settings:
    1. These are reflected as macros of the form `CFG_`.
1. Build OP-TEE:
    1. During the build process, a "TA Dev Kit" is generated;
    1. This TA Dev Kit includes a header file, `conf.h`, that includes the values of the configuration macros;
    1. Another file, `conf.mk`, contains the same information;
    1. The TA Dev Kit also includes a full Make-based build system for enclaves:
        1. The developer writes very simple `Makefile`s following specific rules and includes the TA Dev Kit's `Makefile`s;
        1. During the build, the TA Dev Kit turns the values in `conf.mk` into compiler options.

(TA = "Trusted Application", the term used by OP-TEE to refer to enclaves.)

Without these `CFG_` flags, it is not possible to include OP-TEE headers in Open Enclave enclaves. Generally, one wouldn't want to do that anyway, but such headers are necessary to set the OP-TEE header in an enclave via `OE_SET_ENCLAVE_OPTEE`, which is the OP-TEE version of `OE_SET_ENCLAVE_SGX`. This, in turn, requires that these `CFG_` flags be set.

Making every enclave `.c.` include this `host.h` header sets these configuration values properly every time, which is equivalent to doing so at the compiler command line, though easier.

I'm open to suggestions to doing this differently though.

jhand2: A quick grep shows that the only place an optee header is included is in `enclave/core/optee/header.c`. Can we just include conf.h there?

As for an enclave developer, I think calling optee APIs directly breaks OE encapsulation so if they want to do that they can include the header. But I don't know that it makes sense to do it for them in every file.

HernanGatta: If we were to do this, every enclave would have `OE_SET_ENCLAVE_SGX` available by virtue of including `enclave.h`, but not `OE_SET_ENCLAVE_OPTEE`.

To isolate OP-TEE headers, a developer would have to create a C file specifically to include the OP-TEE header that defines `OE_SET_ENCLAVE_OPTEE`.

I'm not opposed to your idea; I merely wonder whether we want the SDK to hide architectural differences between platforms, or sometimes pass that on to the developer.

jhand2: Oh I misread what you said before. You're right, `OE_SET_ENCLAVE_OPTEE` is something all developers need to be able to use (at least for now). If that depends on conf.h then I agree it makes sense to leave this.

3rdparty/optee/libutee/CMakeLists.txt
HernanGatta: Removed.

HernanGatta:
bors try

HernanGatta:
bors try-

HernanGatta:
@jhand2, how many reviewers are required to make Bors happy? According to the summary view at the bottom of the PR, it looks like all is green with a single approval.

jhand2:
@HernanGatta I've been seeing this same docker pull timeout issue with our CI/CD builds over the last day. I have reported it at #2503 and am running bors again.

bors r+

PR #2505, opened 2020-02-06
Ignore ~/.gdbinit when running oegdb tests

jhand2:
The oegdb tests fail when using some popular gdbinit plugins such as gdb-dashboard. Because this test does some python scripting in gdb, simply disable running with the user's .gdbinit file.

### DIFF
--- a/tests/debugger/oegdb/CMakeLists.txt
+++ b/tests/debugger/oegdb/CMakeLists.txt
@@ -8,6 +8,7 @@ add_test(
     NAME oegdb-test
     COMMAND
         ${OE_BINDIR}/oegdb --batch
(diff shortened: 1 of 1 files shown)


jhand2:
> LGTM, although I would also advise you document your experience in encountering this issue. Perhaps this is something that needs to be noted in the oegdb usage instructions as well.

I think this is more likely a bug in the .gdbinit script I use and doesn't appear to be specific to oegdb. oegdb works fine with the script I use, but the more advanced scripting the test does conflicts in some way. I didn't investigate much further than that but I don't really think its a limitation of oegdb.

jhand2:
Infra failure

bors r+

PR #2517, opened 2020-02-07
Add continuation hook mechanism to the oe vectored exception handler

jhand2:
Currently the OE vectored exception handler is called before the execution context is restored to the pre-exception state. Add a way to inject a callback which is called after the execution context has been restored. This enables more complicated operations such as making OCalls.

@mikbras This is a modified version of your initial implementation that attempts to unify the exception handling mechanisms so that the use case for each is clear to the developer.

### DIFF
--- a/enclave/core/sgx/exception.c
+++ b/enclave/core/sgx/exception.c
@@ -279,6 +279,9 @@ void oe_real_exception_dispatcher(oe_context_t* oe_context)
         td->host_rbp = td->host_previous_rbp;
         td->host_rsp = td->host_previous_rsp;
 
+        if (oe_exception_record.continuation_callback)
+            oe_exception_record.continuation_callback(&oe_exception_record);
+
         oe_continue_execution(oe_exception_record.context);
 
         // Code should never run to here.
--- a/tests/VectorException/enc/continuation_callback.c
+++ b/tests/VectorException/enc/continuation_callback.c
@@ -1,0 +1,45 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/enclave.h>
+#include <openenclave/internal/calls.h>
+
+#include "VectorException_t.h"
+
+#define OE_UD2_OPCODE 0x0b0f
+
+void call_illegal_instruction()
+{
+    asm volatile("ud2;");
+}
+
+uint64_t continuation_callback(oe_exception_record_t* record)
+{
+    record->context->rip += 2;
+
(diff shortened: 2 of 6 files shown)


tests/VectorException/enc/continuation_callback.c
```diff
@@ -0,0 +1,45 @@
...
+void call_illegal_instruction()
+{
+    asm volatile("ud2;");
+}
+
+uint64_t continuation_callback(oe_exception_record_t* record)
+{
+    record->context->rip += 2;
+
+    // OCalls can be make in the continuation callback
```
CodeMonkeyLeet: Typo: "... can be _made_ ..."

enclave/core/sgx/exception.c
```diff
@@ -279,6 +279,9 @@ void oe_real_exception_dispatcher(oe_context_t* oe_context)
         td->host_rbp = td->host_previous_rbp;
         td->host_rsp = td->host_previous_rsp;
 
+        if (oe_exception_record.continuation_callback)
```
CodeMonkeyLeet: I'm unclear on why the callback is necessary:
* This seems to be driven by the need to reset the `td->host_rbp`|`rsp` values so that the ocall doesn't end up in the wrong place on the host stack.
* Both handled and unhandled exceptions need to restore these values anyway. Why aren't they simply restored on entry into this method, in which case the exception handler implementation itself can ocall without needing to to defer that to a continuation callback.
* If there is some behavior that cannot be performed in exception handler after those values are restored, how do you message/prevent a continuation callback from trying to do these things and falling over?

jhand2: @mikbras mentioned that there are legitimate cases where one would want their exception handler before these registers are restored, although no concrete examples come to mind for me.

One thing that hangs me up from just doing as you suggest is that other exception handling mechanisms (Windows Vectored Exception Handling, POSIX signals) also call the exception handler before the execution state is restored. But I don't have a good idea of why.

I'll leave this PR open but I think I need to do some more research about why these other mechanisms behave the way they do. I would love if adding this extra handler weren't necessary.

CodeMonkeyLeet: @jhand2 Thanks for looking into this. If it turns out that this is necessary, then we should elaborate on the distinction between the two cases in documents and consider adding a usage sample to illustrate what to do in one versus the other. e.g. there's no explanation of the difference of the state of the registers in our public docs today.

jhand2:
This is not needed. See #2553

PR #2518, opened 2020-02-07
Publish return code defines for vectored exception handling

jhand2:
The vectored exception handling feature is published as a public API but the return code defines OE_EXCEPTION_CONTINUE_SEARCH and OE_EXCEPTION_CONTINUE_EXECUTION are in internal headers. Move these to enclave.h where oe_add_vectored_exception_handler resides.

### DIFF
--- a/include/openenclave/bits/exception.h
+++ b/include/openenclave/bits/exception.h
@@ -16,6 +16,14 @@
 
 OE_EXTERNC_BEGIN
 
+/** Return value used by an enclave vectored exception handler to indicate
+ *  to the dispatcher that it should continue searching for the next handler. */
+#define OE_EXCEPTION_CONTINUE_SEARCH 0x0
+
+/** Return value used by an enclave vectored exception handler to indicate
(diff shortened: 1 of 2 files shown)


include/openenclave/enclave.h
```diff
@@ -43,6 +43,13 @@ OE_EXTERNC_BEGIN
 #define __thread
 #endif
 
+/** Continue calling next vectored exception handler */
```
CodeMonkeyLeet: Consider being a bit more verbose here as these will end up in the doxygen comments for the enclave.h file on [openenclave.io](https://openenclave.io/apidocs/v0.8/enclave_8h.html). e.g. 
```
/**
 * Return value used by an enclave vectored exception handler to indicate
 * to the dispatcher that it should continue searching for the next handler.
 */
```

jhand2: Updated

jhand2:
Note that I moved these defines to `include/openenclave/bits/exception.h` which is included by enclave,h. I did this because the host-side exception handler also uses these.

PR #2522, opened 2020-02-07
Update libc-test to the latest version

jhand2:
Update libc-test to revision a51df71b050f3f9dfdc0a7d90978b57277b582ec. This revision contains many fixes for math tests which previously failed.

I will update the tests we run and the `LibcSupport.md` document in a separate PR to make it easier to review.

### DIFF
--- a/3rdparty/musl/README.md
+++ b/3rdparty/musl/README.md
@@ -1,12 +1,21 @@
-musl libc:
-==========
+# Musl libc and libc-test projects
 
-This directory contains **musl libc**, obtained from this URL.
+This directory contains the musl libc library and acompanying libc-test test
+suite.
 
-```
-https://www.musl-libc.org/releases/musl-1.1.19.tar.gz
-```
+See `./update.make` for specific information regarding the source and version
+of these projects.
+
+## musl libc:
+
+This directory contains **musl libc**
 
 Typing **make** installs the source tree under the build directory and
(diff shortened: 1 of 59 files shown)


3rdparty/musl/README.md
```diff
@@ -10,3 +10,15 @@ https://www.musl-libc.org/releases/musl-1.1.19.tar.gz
...
 applies patches to it. Nothing is built though, as building is performed
 from the **libc** directory.
+
+libc-test:
+==========
+
+libc-test is a generic libc testing framework, obtained from this git repo:
+
+```
+git://repo.or.cz/libc-test
```
CodeMonkeyLeet: A couple of suggestions:
* Can you clean up the text around the musl libc itself? It's not the right version anymore (OE SDK is on 1.1.21 now) 
* For both the libc and libc-test, can we instead redirect to the [update.make](https://github.com/openenclave/openenclave/blob/master/3rdparty/musl/update.make) file instead as the single source of truth so that we don't need to keep the README.md versioning in sync with the actual update.make? (i.e. put the latest libc-test commit ID in update.make)
* update.make uses the original repo `git://nsz.repo.hu:45100/repo/libc-test` rather than the mirror `git://repo.or.cz/libc-test` ... we should refer to one or the other consistently just to avoid confusion for folks who may not understand the relationship between the two.

jhand2: Done

PR #2526, opened 2020-02-09
Add ARM64 CI Prerequisites to Full Container

HernanGatta:
This PR downloads and extracts into the Full Docker container images an ARM TrustZone emulator and a runtime environment to allow for the execution of CI passes for ARM64. Additionally, `netcat` is added as a prerequisite for ARM.

I'll add the `ARMTZ_EMULATOR_URI` parameter to the corresponding Jenkins pipeline if and once this PR is approved, and before it is merged.

See PR #2527 for details.


HernanGatta:
@BRMcLaren, @jhand2, would you have any love to spare for this PR and @2527? They spent St. Valentine's alone :) .

BRMcLaren:
bors try

HernanGatta:
The last build failure was a fluke, a certificate validation test failed that has nothing to do with the contents of this PR. Trying again.

HernanGatta:
bors try

BRMcLaren:
bors try

HernanGatta:
@BRMcLaren, can I try steps 1&2 as outlined under the "Merging Strategy" header in #2527? They do not require merging anything, nor modifying any active CI artifacts.

radhikaj:
@ionutbalutoiu @BRMcLaren Please take a look at this one.

scripts/ansible/roles/linux/openenclave/vars/ubuntu/main.yml
```diff
@@ -45,6 +45,7 @@ apt_arm_packages:
   - "libc6-dev:arm64"
   - "libssl-dev:arm64"
   - "libfdt1"
+  - "netcat"
```
BRMcLaren: Is the netcat dependency only needed for ARM64 ?

BRMcLaren:
@HernanGatta Let's keep the conversation centered here: https://github.com/openenclave/openenclave/pull/2527#issuecomment-592204432

scripts/ansible/roles/linux/openenclave/vars/ubuntu/main.yml
CodeMonkeyLeet: On a related note, is this a hard dependency for QEMU or something else? This is still a smaller list than documented by docs/GettingStarted/OP-TEE/Debugging/QEMU.md so I'm just trying to get an intuition for why this addition is necessary & sufficient.

HernanGatta: The list in the document you referenced builds the emulator from source, whereas for CI, we use a pre-built one to avoid incurring the additional compile time (about 30' on a six-core system).

`netcat` (`nc`) is used in the Jenkins task as modified by #2527* to wait for the emulator to have created the TCP endpoint in the host used to forward the guest's port 22 for its SSH server. If we don't wait, even if it's less than a second, the script fails.

* `.jenkins/Jenkinsfile`, line 104.

CodeMonkeyLeet: @HernanGatta Got it, thanks for the explanation!

CodeMonkeyLeet: @HernanGatta:
> Having said that, it might be a good idea to include those dependencies in the Ansible role so that by the time anyone tries to build the emulator from scratch, the dependencies will already be installed. @CodeMonkeyLeet, what do you think?

I'll defer this to @BRMcLaren but my intuition is that we should treat them as separate configurations (i.e. use pre-built vs. build their own QEMU) and we should simply support the common path; we can always add them later and I imagine the share of folks wanting to take the extra time to build QEMU (did you mean seconds or minutes?) is probably lower in general.

radhikaj:
@BRMcLaren Can we get this in for 0.9?

BRMcLaren:
@radhikaj Today we set up HernanGatta with an environment to build according to the merging strategy outline here: https://github.com/openenclave/openenclave/pull/2527#issue-372884804

0.9, depends on them :)

kuanggang-git:

dthaler:
bors try

radhikaj:
@HernanGatta  - Please reopen when you can work on this again

PR #2534, opened 2020-02-10
Workaround for debugger dependency of SDK symbols

anakrish:
Currently, the Windows debuggers have the following *temporary* requirements.
In the next iterations, these requirments will be removed.

- WinDbg requires debug symbols oeedebugrt.dll. Therefore it must be built with
  debugging information (/Zi compiler flag, /debug linker flag). Additionally,
  oedebugrt.pdb must be distributed.
 - Visual Studio debugger requires debug symbol information for for __oe_dispatch_ocall
  function. Therefore this the file which contains this function (host/sgx/calls.c) is
  compiled with /Z7 flags which causes the object file to contain debug information.
  The object file becomes part of oehost.lib and when an enclave applicaiton is created,
  the linker extracts the debug information from this object file and puts it in the
  application's pdb.

### DIFF
--- a/cmake/copy_oedebugrt_target.cmake
+++ b/cmake/copy_oedebugrt_target.cmake
@@ -18,12 +18,19 @@ function(copy_oedebugrt_target TARGET_NAME)
     get_property(OEDEBUGRTLOCATION TARGET openenclave::oedebugrt PROPERTY LOCATION)
     # Add copy actions for the dependencies
     add_custom_command(
-        OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/oedebugrt.dll ${CMAKE_CURRENT_BINARY_DIR}/oedebugrt.pdb
+        OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/oedebugrt.dll
         DEPENDS ${OEDEBUGRTLOCATION}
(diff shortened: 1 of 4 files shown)


anakrish:
bors try

cmake/copy_oedebugrt_target.cmake
```diff
@@ -18,7 +18,7 @@ function(copy_oedebugrt_target TARGET_NAME)
     get_property(OEDEBUGRTLOCATION TARGET openenclave::oedebugrt PROPERTY LOCATION)
     # Add copy actions for the dependencies
     add_custom_command(
-        OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/oedebugrt.dll
+        OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/oedebugrt.dll ${CMAKE_CURRENT_BINARY_DIR}/oedebugrt.pdb
```
radhikaj: LGTM. Please verify that when you build samples(after installing sdk) oedebugrt.pdb and oedebugrt.dll are copied over to host exe location.

anakrish: Verified.

anakrish:
> 
> 
> LGTM ... can you also file an issue to track removing this workaround?

Created issue #2542

Issue #2089, opened 2019-08-13
Support digest signing workflow in oesign for production signing


CodeMonkeyLeet:
oesign takes the signing key and enclave to sign in a single step today, which is not amenable for production signing scenarios as the owner of the private key is usually in a locked down environment that is separate from the build pipeline.

Similar to Intel's sgx_sign tool, oesign needs to support _digest signing_, where: 
1. The lower trust build environment invokes oesign to generate a digest of the enclave
2. Only the digest is sent to the high trust production signing environment for signing
3. The resulting signature is returned to the build environment where oesign is invoked again to embed the signature to the enclave binary

PR #2545, opened 2020-02-12
Design proposal for digest signing support in oesign

CodeMonkeyLeet:
This is a proposal to address #2089

Signed-off-by: Simon Leet <simonle@microsoft.com>


docs/DesignDocs/oesign_digest_signing_support.md
```diff
@@ -0,0 +1,125 @@
...
+>
+> Similar to Intel's sgx_sign tool, oesign needs to support _digest signing_, where:
+>
+> 1. The lower trust build environment invokes oesign to generate a digest of the enclave
+> 2. Only the digest is sent to the high trust production signing environment for signing
+> 3. The resulting signature is returned to the build environment where oesign is invoked again to embed the signature to the enclave binary
+
+## User Experience
+
+To sign an OE enclave binary _my_enclave_ with the optional metadata specified in _my_enclave.conf_ and the _enclave_signing_key.pem_ owned by a different signing authority:
```
jazzybluesea: Maybe adding some diagrams here would help clarify the flow.

docs/DesignDocs/oesign_digest_signing_support.md
```diff
@@ -0,0 +1,125 @@
...
+- The resulting _my_enclave.dig_ is sent to a signing authority that owns the private key to sign the digest and return _my_enclave.dig.signed_ for the signature requestor to embed in _my_enclave_
+  - The example below assumes OpenSSL is used by the signing authority, but this can be any RSA signing interface used by the signing authority (e.g. hardware security modules).
+  - Note that the signing authority must distribute the public key in the form of _enclave_signing_cert.cer_ to the signature requestor for the final step, whether it's as a well-known public certificate or as additional collateral returned to the signature requestor with _my_enclave.dig.signed_.
+
+```bash
+oesign hash -e my_enclave -c my_enclave.conf -d my_enclave.dig
+
+openssl pkeyutl -sign -in my_enclave.dig -pkeyopt digest:sha256 -inkey enclave_signing_key.pem -keyform PEM -out my_enclave.dig.signed
+
+oesign sign -e my_enclave -c my_enclave.conf -x enclave_signing_cert.cer -d my_enclave.dig.signed
```
achamayou: sgx_sign seems to use `gendata` and `catsig` where the proposal is for `hash` and `sign`. I think `digest` would be better than `hash`, since that's what the doc calls it, and the files are `.dig`. `sign` seems a bit inaccurate since we're not specifically signing at this point, but I don't have a good suggestion either (pack, assemble, mkenc... ?).

MWShan: sgx_sign uses 'gendata' because it is generating the data to be signed.  I think that 'catsig' was used to 'concatenate the signature'; although 'addsig' might have been better.

CodeMonkeyLeet: I agree that 'digest' is a better name and will update to that.

I'm using 'sign' for the reason that Mark points out: oesign is basically doing the same thing as the single-signing step by creating a signature embedded in the enclave binary, even though part of it is that it uses a pre-signed digest value to populate the signature. Unlike the intel version, it's not just inserting the entire sign signature blob (which would be the result of the Intel signed gendata)

docs/DesignDocs/oesign_digest_signing_support.md
```diff
@@ -0,0 +1,133 @@
...
+
+And it should support a production signing workflow where keys are owned by a separate signing authority:
+
+![oesign digest signing workflow](images/oesign_digest_signing.svg)
+
+## User Experience
+
+To sign an OE enclave binary _my_enclave_ with the optional metadata specified in _my_enclave.conf_ and the _enclave_signing_key.pem_ owned by a different signing authority:
+- The oesign steps are done by the signature requestor (e.g. as part of a build pipeline)
+- The resulting _my_enclave.dig_ is sent to a signing authority that owns the private key to sign the digest and return _my_enclave.dig.signed_ for the signature requestor to embed in _my_enclave_
```
radhikaj: Add a `.` at the end of this sentence.

CodeMonkeyLeet: Fixed.

docs/DesignDocs/oesign_digest_signing_support.md
CodeMonkeyLeet: Added.

Issue #2046, opened 2019-07-19
Consider adding an additional input argument to oe_ocall and oe_ecall.


mikbras:
Primitive functions like **oe_host_realloc** perform the following steps.
- allocate host storage to hold the **ptr** and **size** arguments.
- invoke **oe_ocall** to perform the reallocation.
- free host storage.

This incurs 3 host transitions to perform a single realloc.

Adding one or more additional input argument would solve this problem. The arguments could be passed in registers between the enclave and the host.

Note that the edger8r could use these additional register arguments for optimization. For example, when it generates **oe_realloc_ocall**, it could use **oe_ocall** directly and pass both the **ptr** and **size** argument (which are passed in registers).

PR #2551, opened 2020-02-12
OE SGX: Add context parameter for ecalls.

anakrish:
The host passes in an instance of oe_ecall_context_t to the enclave
when making an ecall. This context can hold information that the enclave
can leverage to efficiently implement different functionality.

- Fast ocalls.
  Currently making an EDL generated ocall involves 4 core ocalls in addition
  to the core ocall to call the host implementation of the ocall.
  The 4 additional ocalls are needed to allocate/free the oe_call_host_function_args_t
  structure as well as to allocate/free the serialized ocall arguments.

  oe_ecall_context_t avoids all these additional ocalls to allocate/free memory.
  The context stores the oe_call_host_function_args_t as well has pointer to a host
  memory buffer for marshaling the serialized ocall inputs and outputs.
  The host memory buffer is allocated in thread local storage and can be large enough
  to accommodate most common ocalls.

- OCall stack stitching
  The context provides a place to store the enclave's exit frame (rbp, rsp, rip).
  A debug enclave can store its exit frame in the context prior to making the eexit
  for the ocall. Host side SDK can then use the exit frame information to stitch
  the ocall stack in place as well as pass it along to oedebugrt.
  This change is part of the debugger contract and will be done in another PR.

This PR addresses #2046. An additional argument (context) is added to the ecall and this in turn allows passing parameters efficiently to ocalls.

### DIFF
--- a/enclave/core/sgx/enter.S
+++ b/enclave/core/sgx/enter.S
@@ -40,7 +40,7 @@ oe_enter:
 .cfi_startproc
 
 .save_host_registers:
-    // Backup the current host rbp, rsp and context to previous.
+    // Backup the current host rbp, rsp, and context to previous.
     mov %fs:td_host_rbp, %r8
     mov %r8, %fs:td_host_previous_rbp
     mov %fs:td_host_rsp, %r8
--- a/enclave/core/sgx/enter.S
+++ b/enclave/core/sgx/enter.S
@@ -20,6 +20,7 @@
 //     Registers from host caller of EENTER:
 //         RDI - ARG1
 //         RSI - ARG2
+//         RDX - HOST ECALL CONTEXT
 //
 //     This function performs the following tasks:
 //
@@ -39,23 +40,26 @@ oe_enter:
 .cfi_startproc
 
 .save_host_registers:
-    // Backup the current host rbp and rsp to previous.
+    // Backup the current host rbp, rsp and context to previous.
     mov %fs:td_host_rbp, %r8
     mov %r8, %fs:td_host_previous_rbp
     mov %fs:td_host_rsp, %r8
     mov %r8, %fs:td_host_previous_rsp
+    mov %fs:td_host_ecall_context, %r8
+    mov %r8, %fs:td_host_previous_ecall_context
 
     // Save host registers (restored on EEXIT)
     mov %rcx, %fs:td_host_rcx // host return address here
     mov %rsp, %fs:td_host_rsp
     mov %rbp, %fs:td_host_rbp
+    mov %rdx, %fs:td_host_ecall_context
 
 .determine_entry_type:
     // Check if this is exception dispatching request.
     // exception-dispatching-request-check
     cmp $0, %rax
     jne .exception_entry
-    
+
     // Stop speculative execution at fallthrough of conditional
     // exception-dispatching-request-check.
     lfence
@@ -223,7 +227,7 @@ oe_enter:
     jmp .clear_enclave_registers
 
 .clean_exit:
-    // Stop speculative execution at target of conditional jump 
+    // Stop speculative execution at target of conditional jump
     // after exit-type-check.
     lfence
 
--- a/enclave/core/sgx/hostcalls.c
+++ b/enclave/core/sgx/hostcalls.c
@@ -3,18 +3,78 @@
 
 #include <openenclave/edger8r/enclave.h>
 #include <openenclave/enclave.h>
+#include <openenclave/internal/ecall_context.h>
+#include "td.h"
 
-// Function used by oeedger8r for allocating ocall buffers. This function can be
-// optimized by allocating a buffer for making ocalls and pass it in to the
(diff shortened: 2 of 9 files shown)


anakrish:
bors try

enclave/core/sgx/enter.S
```diff
@@ -39,23 +40,26 @@ oe_enter:
 .cfi_startproc
 
 .save_host_registers:
-    // Backup the current host rbp and rsp to previous.
+    // Backup the current host rbp, rsp and context to previous.
     mov %fs:td_host_rbp, %r8
     mov %r8, %fs:td_host_previous_rbp
     mov %fs:td_host_rsp, %r8
     mov %r8, %fs:td_host_previous_rsp
+    mov %fs:td_host_ecall_context, %r8
```
mikbras: It looks good the way it is, but I was wondering whether **td_host_ecall_context** could just be a **__thread** variable (TLS). Then I suppose this would become:

```
mov    %fs:td_host_previous_rsp@tpoff, %r8
```

And if so, should we transition to using thread local storage for other fields?

anakrish:
bors try

enclave/core/sgx/hostcalls.c
```diff
@@ -3,18 +3,78 @@
...
 
-// Function used by oeedger8r for allocating ocall buffers. This function can be
-// optimized by allocating a buffer for making ocalls and pass it in to the
-// ecall and making it available for use here.
+/**
+ * Validate and fetch this thread's ecall context.
+ */
+static oe_ecall_context_t* _get_ecall_context()
+{
+    td_t* td = oe_get_td();
```
mingweishih: Just out of curiosity, have OE checked the td is initialized anywhere? It seems OE assumes the `oe_get_id` always get initialized data.

include/openenclave/internal/sgxtypes.h
```diff
@@ -572,7 +572,7 @@ oe_thread_data_t* oe_get_thread_data(void);
 
 #define OE_THREAD_LOCAL_SPACE (OE_PAGE_SIZE)
 
-#define OE_THREAD_SPECIFIC_DATA_SIZE (3840)
+#define OE_THREAD_SPECIFIC_DATA_SIZE (3824)
```
mingweishih: Can we have more information regarding this number?

enclave/core/sgx/enter.S
anakrish: Yes. TLS in not initialized at this point. I have created an issue to remove use of td_t and use thread-local variables instead. #2554 
Based on initial experiments, it should be quite possible to switch to tls variables.

host/sgx/enter.c
```diff
@@ -57,20 +58,27 @@
...
 
 /**
  * oe_enter Executes the ENCLU instruction and transfers control to the enclave.
  *
  * The ENCLU instruction has the following contract:
- * EENTER(RBX=TCS, RCX=AEP, RDI=ARG1, RSI=ARG2) contract
+ * EENTER(RBX=TCS, RCX=AEP, RDI=ARG1, RSI=ARG2, RDX=ECALL_CONTEXT) contract
  * Input:
- *       RBX=TCS, RCX=AEP, RDI=ARG1, RSI=ARG2,
+ *       RBX=TCS, RCX=AEP, RDI=ARG1, RSI=ARG2, EDX=ECALL_CONTEXT
```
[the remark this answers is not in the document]
anakrish: Fixed.

include/openenclave/internal/ecall_context.h
```diff
@@ -0,0 +1,39 @@
...
+#include <openenclave/bits/result.h>
+#include <openenclave/bits/types.h>
+#include <openenclave/internal/calls.h>
+
+OE_EXTERNC_BEGIN
+
+typedef struct _oe_ecall_context
+{
+    // For debug enclaves, registers values for ocall stack stitching.
+    uint64_t eexit_rbp;
```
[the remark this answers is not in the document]
anakrish: Removed.

include/openenclave/internal/sgxtypes.h
anakrish: I documented how/why OE_THREAD_SPECIFIC_DATA_SIZE is computed the way it is. And also added an assertion.

anakrish:
> 
> 
> The title, description, and labels all do not mention this as being SGX-specific in any way.
> However when I look at the files changed, I only see SGX-specific code. Why is that?

Good point. Updated title and labels.

dthaler:
Actually my question was about why this is not relevant to optee also?

anakrish:
> 
> 
> Actually my question was about why this is not relevant to optee also?

Ah OK. This PR relies on SGX memory model where the enclave has access to host memory space.
From my understanding, OPTEE does not allow this access. I have created issue #2555 and assigned it to myself and @HernanGatta to investigate similar optimizations for OPTEE.

PR #2552, opened 2020-02-12
Enable tests for previously unsupported math functions

jhand2:
Many libc math tests which failed under the older version of libc-test now succeed. Enable the tests that pass and remove them from the list of unsupported functions in docs/LibcSupport.md.


jhand2:
bors try

jhand2:
The jnf function fails in Release mode (but not in debug mode). Just removing it for now.

bors try

PR #2553, opened 2020-02-12
Add test to make an OCall in the enclave Vectored Exception Handler

jhand2:
Add a test function to tests/VectorException to test the ability to make an OCall in the exception handler.


jhand2:
bors try

jhand2:
Infra failure

bors r+

Issue #2547, opened 2020-02-12
Add doxygen tag for advanced APIs and the ability to opt in to them


radhikaj:
From discussions related to #2328 it was determined that we should have a special category of APIs called "advanced" similar to "experimental". There should be a doxygen tag for these and a way for developers to opt into these APIs.

radhikaj:
@dthaler Can we close this?

dthaler:
@radhikaj Ok with me to close.

PR #2364, opened 2019-12-04
Add @experimental annotation support to doxygen

dthaler:
Addresses issue #2216

This uses the custom tags mechanism explained in the stackoverflow thread: https://stackoverflow.com/questions/537043/custom-tags-with-doxygen
Example use of @experimental in another GitHub project: blynkkk/blynk-library@579545d

Together, these cause a separate "experimental.html" page to be created that has a list of all experimental APIs.  Tagging an API with @experimental then causes the notice to show up on that API's own page, and the notice is hotlinked to the experimental.html page. 

@experimental
causes the following to show up (linked to the Experimental List page):

> This feature is marked as experimental:

@experimental FooBar
casues the following to show up (linked to the Experimental List page):

> This feature is marked as experimental:
>> FooBar

So you use no argument if the entire API is experimental.
You use an argument if some particular subset or variation is experimental.

See http://www.doxygen.nl/manual/custcmd.html#custcmd_complex and http://www.doxygen.nl/manual/commands.html#cmdxrefitem for more info.

Signed-off-by: Dave Thaler <dthaler@microsoft.com>

PR #2560, opened 2020-02-13
Add @advanced doxygen tag support

dthaler:
Similar to the support for @experimental added in PR #2364,
this PR adds support for @advanced in doxygen to help address
issue #2547.

This change causes a separate "advanced.html" page to be created that has
a list of all advanced APIs. Tagging an API with @advanced then causes
the following notice to show up on that API's own page, with the notice
hotlinked to the advanced.html page:

```
This feature is marked as advanced and is not intended for use by apps:
```

### DIFF
--- a/docs/refman/CMakeLists.txt
+++ b/docs/refman/CMakeLists.txt
@@ -19,13 +19,11 @@ if (ENABLE_REFMAN)
 
   # Allow @experimental to be used in headers to mark an API as experimental and add it to a list of
   # experimental APIs.
-  set(DOXYGEN_ALIASES
-    "experimental = \\xrefitem experimental \\\"This feature is marked as experimental\\\" \\\"Experimental List\\\""
-    "experimental{1} = \\xrefitem experimental \\\"Experimental\\\" \\\"Experimental List\\\" \\1")
-
-  # Allow @advanced to be used in headers to mark an API as advanced and add it to a list of
+  # Also allow @advanced to be used in headers to mark an API as advanced and add it to a list of
   # advanced APIs.
   set(DOXYGEN_ALIASES
+    "experimental = \\xrefitem experimental \\\"This feature is marked as experimental\\\" \\\"Experimental List\\\""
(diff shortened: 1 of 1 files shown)


docs/refman/CMakeLists.txt
```diff
@@ -19,9 +19,12 @@ if (ENABLE_REFMAN)
 
   # Allow @experimental to be used in headers to mark an API as experimental and add it to a list of
   # experimental APIs.
+  # Also allow @advanced to be used in headers to mark an API as advanced and add it to a list of
+  # advanced APIs.
   set(DOXYGEN_ALIASES
     "experimental = \\xrefitem experimental \\\"This feature is marked as experimental\\\" \\\"Experimental List\\\""
-    "experimental{1} = \\xrefitem experimental \\\"Experimental\\\" \\\"Experimental List\\\" \\1")
+    "experimental{1} = \\xrefitem experimental \\\"Experimental\\\" \\\"Experimental List\\\" \\1"
+    "advanced = \\xrefitem advanced \\\"This feature is marked as advanced and is not intended for use by apps\\\" \\\"Advanced List\\\"")
```
CodeMonkeyLeet: Don't we also need to add the per item alias? i.e.
```
"advanced{1} = \\xrefitem advanced \\\"Advanced\\\" \\\"Advanced List\\\""
```

dthaler: That would only be if we needed a 1-arg variation.  For comparison, the @experimental PR explained:

So you use no argument if the entire API is experimental.
You use an argument if some particular subset or variation is experimental.

Line 27 is sufficient if an entire API is advanced.  We would need the advanced{1} if we thought that some subset or variation (e.g., some argument being non-null) were advanced but the rest of the API were not.  I omitted it since I think we should rule out that case.

CodeMonkeyLeet: Thanks for the explanation, I agree we should probably keep it to the entire function. It can always be added later should the need arise.

PR #2562, opened 2020-02-13
Rename __cxa_atexit in oecore to oe_cxa_atexit

jhand2:
Symbols in oecore should not conflict with symbols defined in libc to allow alternate libc implementations to be used without issue.

Retain oe_cxa_atexit implementation so that it can be used by oecore. Modify oelibc to simply call oe_cxa_atexit from __cxa_atexit.

### DIFF
--- a/libc/atexit.c
+++ b/libc/atexit.c
@@ -16,3 +16,10 @@
 #undef OE_NEED_STDC_NAMES
 #undef __UNDEF_OE_NEED_STDC_NAMES
 #endif
+
+extern int oe_cxa_atexit(void (*func)(void*), void* arg, void* dso_handle);
+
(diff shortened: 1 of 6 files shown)


libc/atexit.c
```diff
@@ -16,3 +16,10 @@
 #undef OE_NEED_STDC_NAMES
 #undef __UNDEF_OE_NEED_STDC_NAMES
 #endif
+
+extern int oe_cxa_atexit(void (*func)(void*), void* arg, void* dso_handle);
+
+int __cxa_atexit(void (*func)(void*), void* arg, void* dso_handle)
```
Francis-Liu: I'm curious why wouldn't this conflict with alternate libc implementations?

jhand2: Because this is the libc implementation, which sgx-lkl doesn't use.

PR #2566, opened 2020-02-14
Modify the loader to page-align all source addresses

jhand2:
New revisions of the upstream Linux kernel SGX driver require all source and target addresses to be page-aligned for page loads into the enclave. Allocate the buffer for relocation pages on a page-aligned address to fulfill this requirement for source addresses.

This was tested with v26 of the currently-being-upstreamed driver.

### DIFF
--- a/host/sgx/elf.c
+++ b/host/sgx/elf.c
@@ -14,6 +14,7 @@
 #include <string.h>
 #include <sys/stat.h>
 #include "../fopen.h"
+#include "../memalign.h"
 #include "../strings.h"
 
 #define GOTO(LABEL)                                            \
@@ -1923,7 +1924,7 @@ oe_result_t elf64_load_relocations(
     {
         *size_out = oe_round_up_to_page_size(size);
 
-        if (!(*data_out = malloc(*size_out)))
+        if (!(*data_out = oe_memalign(OE_PAGE_SIZE, *size_out)))
         {
             *size_out = 0;
             goto done;
--- a/host/sgx/create.c
+++ b/host/sgx/create.c
@@ -87,9 +87,13 @@ static oe_result_t _add_filled_pages(
     uint32_t filler,
     bool extend)
 {
-    oe_page_t page;
     oe_result_t result = OE_UNEXPECTED;
+    oe_page_t* page = NULL;
     size_t i;
+
+    page = oe_memalign(OE_PAGE_SIZE, sizeof(oe_page_t));
+    if (!page)
+        OE_RAISE(OE_OUT_OF_MEMORY);
 
     /* Reject invalid parameters */
     if (!context || !enclave_addr || !vaddr)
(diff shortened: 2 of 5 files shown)


jhand2:
> @jhand2 Can you double check if this is the only change that's necessary?

My initial test pass succeeded but if I run multiple times, it looks like tests fail here and there, which would suggest you're right, there are some other places we use unaligned addresses. I'll dig a little more.

jhand2:
@CodeMonkeyLeet so option 1 is to try to find all these places and ensure they're page-aligned. This is a bit piece-meal but the change is relatively small.

Option 2 is to go with the intel approach and just mmap the whole image. This feels cleaner and will probably result in a bunch of churn in the loader. I'm up to give it a shot but I'm wondering if this is something we should tackle now or prioritize after other things.

CodeMonkeyLeet:
@jhand2 I don't think it's necessary to mmap the whole image for this fix, though there may be performance benefits we may want to investigate. The `src` alignment check added to `oe_sgx_load_enclave_data()` should be sufficient validation that all pages are loaded from page-aligned addresses.

I am surprised that calling `oe_sgx_load_enclave_data()` with a stack variable in `_add_control_pages()` and `_add_filled_pages()` still works though. It seems like a really bad idea in general to require 4KB page-aligned variable allocations on the stack and we should convert them into memaligned heap allocations anyway.

host/sgx/elf.c
```diff
@@ -1923,7 +1924,7 @@ oe_result_t elf64_load_relocations(
     {
         *size_out = oe_round_up_to_page_size(size);
 
-        if (!(*data_out = malloc(*size_out)))
+        if (!(*data_out = oe_memalign(OE_PAGE_SIZE, *size_out)))
```
anakrish: The line just immediately above rounds up to page size. Is this change redundant?

jhand2: The start address needs to be page aligned. The line above just ensures that the size is a multiple of OE_PAGE_SIZE.

jhand2:
> I am surprised that calling `oe_sgx_load_enclave_data()` with a stack variable in `_add_control_pages()` and `_add_filled_pages()` still works though. It seems like a really bad idea in general to require 4KB page-aligned variable allocations on the stack and we should convert them into memaligned heap allocations anyway.

@CodeMonkeyLeet I went ahead and fixed `_add_control_pages()` and `_add_filled_pages()` to use aligned src addresses for consistency.

I discussed with you previously that the stack protector test is failing when loading pages due to an mprotect. I spent some time with this today and I believe this is a bug in the upstream SGX driver. See my email on LKML for more context: https://lkml.org/lkml/2020/2/18/1477

jhand2:
> 
> 
> > I am surprised that calling `oe_sgx_load_enclave_data()` with a stack variable in `_add_control_pages()` and `_add_filled_pages()` still works though. It seems like a really bad idea in general to require 4KB page-aligned variable allocations on the stack and we should convert them into memaligned heap allocations anyway.
> 
> @CodeMonkeyLeet I went ahead and fixed `_add_control_pages()` and `_add_filled_pages()` to use aligned src addresses for consistency.
> 
> I discussed with you previously that the stack protector test is failing when loading pages due to an mprotect. I spent some time with this today and I believe this is a bug in the upstream SGX driver. See my email on LKML for more context: https://lkml.org/lkml/2020/2/18/1477

The above is mitigated by #2593

jhand2:
@anakrish and @CodeMonkeyLeet can you take another look? Thanks!

jhand2:
Thanks

bors r+

jhand2:
Fixed instances where memory allocated with `oe_memalign()` were being freed with `free` instead of `oe_memalign_free()`

jhand2:
Fixed one more instance of `oe_memalign_free` that I missed

bors r+

PR #2568, opened 2020-02-15
Fix VectorException test to actually test OCall

jhand2:
The VectorException test has code to test making an OCall from inside the exception handler but that test case is never called. Call the test case.

### DIFF
--- a/tests/VectorException/enc/enc.c
+++ b/tests/VectorException/enc/enc.c
@@ -285,19 +285,21 @@ int enc_test_vector_exception()
     return 0;
 }
 
-uint64_t test_divide_by_zero_handler_with_ocall(
-    oe_exception_record_t* exception_record)
+void call_invalid_instruction()
 {
-    if (exception_record->code != OE_EXCEPTION_DIVIDE_BY_ZERO)
+    asm volatile("ud2;");
+}
+
+uint64_t test_sigill_handler_with_ocall(oe_exception_record_t* exception_record)
+{
+    if (exception_record->code != OE_EXCEPTION_ILLEGAL_INSTRUCTION)
     {
         return OE_EXCEPTION_CONTINUE_SEARCH;
     }
 
     host_set_was_ocall_called();
 
-    // Skip the idiv instruction - 2 is tied to the size of the idiv instruction
-    // and can change with a different compiler/build. Minimizing this with the
-    // use of the inline assembly for integer division
+    // Skip the ud2 instruction
     exception_record->context->rip += 2;
     return OE_EXCEPTION_CONTINUE_EXECUTION;
 }
@@ -305,7 +307,7 @@ uint64_t test_divide_by_zero_handler_with_ocall(
 int enc_test_ocall_in_handler()
 {
     oe_result_t result = oe_add_vectored_exception_handler(
-        false, test_divide_by_zero_handler_with_ocall);
+        false, test_sigill_handler_with_ocall);
     if (result != OE_OK)
     {
         return -1;
@@ -315,16 +317,13 @@ int enc_test_ocall_in_handler()
         "enc_test_ocall_in_handler: will generate a hardware exception inside "
         "enclave!\n");
 
-    if (divide_by_zero_exception_function() != 0)
-    {
-        return -1;
-    }
+    call_invalid_instruction();
 
     oe_host_printf("enc_test_ocall_in_handler: hardware exception is handled "
                    "correctly!\n");
 
-    if (oe_remove_vectored_exception_handler(
-            test_divide_by_zero_handler_with_ocall) != OE_OK)
+    if (oe_remove_vectored_exception_handler(test_sigill_handler_with_ocall) !=
+        OE_OK)
     {
         return -1;
     }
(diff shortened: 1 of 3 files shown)


jhand2:
bors try

anakrish:
@jhand2 
Couple of suggestions regarding the release mode failures:
- The idiv instruction can take up 2 bytes (e.g idiv eax) or 3 bytes (e.g idiv r8) depending on which register is used. It could be possible that the compiler is using a register that triggers 3 byte instruction instead of 2 as expected by the test.
- To make ocalls work correctly, we would need to restore the host_rsp, host_rbp fields in td at the right time.
  ```c
      td->host_rbp = td->host_previous_rbp;
      td->host_rsp = td->host_previous_rsp;
      td->host_ecall_context = td->host_previous_ecall_context;
  ```
  Currently we do it after invoking the handler but before continuing the execution. Since handlers can also make ocalls, we would need to restore these fields before invoking the handler.

jhand2:
> @jhand2
> Couple of suggestions regarding the release mode failures:
> 
>     * The idiv instruction can take up 2 bytes (e.g idiv eax) or 3 bytes (e.g idiv r8) depending on which register is used. It could be possible that the compiler is using a register that triggers 3 byte instruction instead of 2 as expected by the test.

Ah I didn't know that. Thanks for the heads up :smile:

> 
>     * To make ocalls work correctly, we would need to restore the host_rsp, host_rbp fields in td at the right time.
>       ```c
>           td->host_rbp = td->host_previous_rbp;
>           td->host_rsp = td->host_previous_rsp;
>           td->host_ecall_context = td->host_previous_ecall_context;
>       ```
>       
>       
>       Currently we do it after invoking the handler but before continuing the execution. Since handlers can also make ocalls, we would need to restore these fields before invoking the handler.

I'm not sure I understand why they need to be restored in order to make an ocall. If that was the case wouldn't all tests fail rather than just having transient failures?

anakrish:
> I'm not sure I understand why they need to be restored in order to make an ocall. If that was the case wouldn't all tests fail rather than just having transient failures?

Only those tests that make an ocall in an exception handler would fail.

One interesting experiment to try would be to comment out the lines that restore those fields and run tests. Most tests that raise an exception (say use cpuid) and then make an ocall should fail.

jhand2:
> > I'm not sure I understand why they need to be restored in order to make an ocall. If that was the case wouldn't all tests fail rather than just having transient failures?
> 
> Only those tests that make an ocall in an exception handler would fail.
> 
> One interesting experiment to try would be to comment out the lines that restore those fields and run tests. Most tests that raise an exception (say use cpuid) and then make an ocall should fail.

You're right, setting host rbp and rsp does fix the problem due to the explicit contract that host rsp and rbp will not be clobbered by calling enclu. Thanks for the help!

bors try

jhand2:
@anakrish and @mikbras could you take another look at this? This should work for allowing ocalls from the enclave. Thanks!

jhand2:
> Can you also fix the existing zero division test so that the compiler always generates 2 byte instruction?
> It is an testing issue waiting to happen.

I can't find any documentation about the size of the idiv instruction being variable. Can you point me to anything? Otherwise I'm not sure how to ensure the instruction width.

anakrish:
> 
> 
> > Can you also fix the existing zero division test so that the compiler always generates 2 byte instruction?
> > It is an testing issue waiting to happen.
> 
> I can't find any documentation about the size of the idiv instruction being variable. Can you point me to anything? Otherwise I'm not sure how to ensure the instruction width.

You can try out the assembler here: https://defuse.ca/online-x86-assembler.htm

`idiv ecx` has machine code `f7 f9` wheras `idiv r8` has machine code `49 f7 f8`.

tests/VectorException/enc/enc.c
```diff
@@ -285,27 +285,29 @@ int enc_test_vector_exception()
...
-    oe_exception_record_t* exception_record)
+void call_invalid_instruction()
 {
-    if (exception_record->code != OE_EXCEPTION_DIVIDE_BY_ZERO)
+    asm volatile("ud2;");
+}
+
+uint64_t test_sigill_handler_with_ocall(oe_exception_record_t* exception_record)
+{
+    if (exception_record->code != OE_EXCEPTION_ILLEGAL_INSTRUCTION)
```
anakrish: One good way to lock down the behavior would be to take a snapshot of the registers before the ocall and then after return from the ocall. The non-volatile registers ought to be preserved.

jhand2: Added to the test and modified idiv to only generate a 2 byte instruction

jhand2:
bors try

jhand2:
bors try-

tests/VectorException/enc/enc.c
anakrish: Thanks.

jhand2:
Fixed release tests by marking clobbered registers in inline asm

bors r+

jhand2:
Use memory in inline asm to avoid clobbering registers before they are read on release builds.

bors r+

Issue #2569, opened 2020-02-15
Crash due to missing locks around enclave->bindings


mikbras:
A multi-threaded test encountered the following crash.

```
2020-02-15T16:53:13.731807Z [(H)ERROR] tid(0x7fa56f7fe700) | tcs=0x81260000
 [/root/openenclave/host/sgx/enclavemanager.c:oe_query_enclave_instance:199]
 ./run: line 1: 19776 Aborted                 (core dumped)
 ```

This due to missing locks here and in other places that should protect these fields:

```
oe_enclave_t.bindings
oe_enclave_t.num_bindings
```

mikbras:
Fixed by PR #2570

jhand2:
Can you give more details on which test was failing and under what circumstances?

jhand2:
We should consider adding better testing to stress test our thread-safety

achamayou:
@radhikaj can we please raise the priority of this item and get the fix merged?

It's a safety issue in multi-threaded applications, there is no user workaround at this time, it shouldn't take a month to make it to master.

jhand2:
It takes a month when the PR author doesn't bother to respond to feedback

PR #2570, opened 2020-02-15
Fix crash due to missing locks around enclave->bindings

mikbras:
This PR fixes issue #2569 (a crash due to missing locks around enclave->bindings).

The following crash occurred in a crash of a test that is part of a very large third-party application. I tried to reproduce it in a smaller test but was unable.

```
2020-02-15T16:53:13.731807Z [(H)ERROR] tid(0x7fa56f7fe700) | tcs=0x81260000
 [/root/openenclave/host/sgx/enclavemanager.c:oe_query_enclave_instance:199]
 ./run: line 1: 19776 Aborted                 (core dumped)
 ```

The crash is caused when one thread modifies **oe_enclave_t.bindings** while the another thread attempts to read it. All readers/writers of the bindings array must acquire a lock since the bindings change over the lifetime of the program.

### DIFF
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -477,8 +477,6 @@ int __oe_dispatch_ocall(
 
         if (enclave->simulate)
         {
-            oe_mutex_lock(&enclave->lock);
-
             /**
              * GetThreadBinding may not work since it uses pthread APIs.
              * pthread depends on FS register being set correctly, which
@@ -493,8 +491,6 @@ int __oe_dispatch_ocall(
                     break;
                 }
             }
-
-            oe_mutex_unlock(&enclave->lock);
         }
         else
         {
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -477,6 +477,8 @@ int __oe_dispatch_ocall(
 
         if (enclave->simulate)
         {
+            oe_mutex_lock(&enclave->lock);
+
             /**
              * GetThreadBinding may not work since it uses pthread APIs.
              * pthread depends on FS register being set correctly, which
@@ -491,6 +493,8 @@ int __oe_dispatch_ocall(
                     break;
                 }
             }
+
+            oe_mutex_unlock(&enclave->lock);
         }
         else
         {
--- a/host/sgx/enclavemanager.c
+++ b/host/sgx/enclavemanager.c
@@ -173,14 +173,20 @@ oe_enclave_t* oe_query_enclave_instance(void* tcs)
         OE_LIST_FOREACH(tmp, &oe_enclave_list_head, next_entry)
         {
             oe_enclave_t* enclave = tmp->enclave;
+
+            oe_mutex_lock(&enclave->lock);
+
             for (uint32_t i = 0; i < OE_COUNTOF(enclave->bindings); i++)
             {
                 if (enclave->bindings[i].tcs == (uint64_t)tcs)
                 {
                     ret = enclave;
+                    oe_mutex_unlock(&enclave->lock);
                     goto cleanup;
                 }
             }
+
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by mikbras** [comment]
bors try

File: host/sgx/switchless.c [hunk 1]
```diff
@@ -100,8 +100,12 @@ oe_result_t oe_start_switchless_manager(
     // because the maximum parallelism is dictated by the latter for
     // synchronous ocalls. We may need to revisit this for asynchronous
     // calls later.
-    if (num_host_workers > enclave->num_bindings)
-        num_host_workers = (uint32_t)enclave->num_bindings;
+    oe_mutex_lock(&enclave->lock);
+    {
+        if (num_host_workers > enclave->num_bindings)
```
anakrish: It is not clear why this code needs to be guarded. num_bindings is never modified after the enclave has been constructed.

host/sgx/calls.c
```diff
@@ -477,6 +477,8 @@ int __oe_dispatch_ocall(
 
         if (enclave->simulate)
         {
+            oe_mutex_lock(&enclave->lock);
```
anakrish: It is not clear why locking is needed here.
The ocall is dispatched in the same thread that made the ecall and no two threads concurrently use the same tcs.

anakrish: Note: With the newer implementation of oe_enter_sim, the[ GS/FS register restoration](https://github.com/openenclave/openenclave/blob/00bd2d52ea73fe6245e7164ecef6a2a704a9e345/host/sgx/enter.c#L226) is correctly done by the simulation of the ENCLU instruction. This special casing for simulation mode is no longer needed.

jhand2:
A agree with @anakrish's comments.

As an asside, "fix for issue 2569" isn't a very informative commit message/PR description. If this is to be merged, please update with more details. The commit message should explain what is fixed in a self contained way without referencing other issues.

See https://github.com/openenclave/openenclave/blob/master/docs/Contributing.md#commit-messages

host/sgx/switchless.c
mikbras: Thank you. Good observation. Fixed.

host/sgx/calls.c
mikbras: Good catch Anand. I removed conservative locking.

mikbras:
Thank you Anand for your review. I removed the two unnecessary locks.

host/sgx/enclavemanager.c
```diff
@@ -173,14 +173,20 @@ oe_enclave_t* oe_query_enclave_instance(void* tcs)
...
             oe_enclave_t* enclave = tmp->enclave;
+
+            oe_mutex_lock(&enclave->lock);
+
             for (uint32_t i = 0; i < OE_COUNTOF(enclave->bindings); i++)
             {
                 if (enclave->bindings[i].tcs == (uint64_t)tcs)
                 {
                     ret = enclave;
+                    oe_mutex_unlock(&enclave->lock);
```
CodeMonkeyLeet: Consider `break` for ending the loop instead for a single lock-unlock pair instead of an early return.

achamayou:
@mikbras we'd be really grateful if you could be merged soon, we think we're hitting this problem at the moment, and we'd really like to check if that fixes it.

jhand2:
I'm fine with merging this. @mikbras could you 

1. address Simon's comment
2. Squash commits
3. Add DCO

Thanks!

achamayou:
@jhand2 I have done all of this on https://github.com/openenclave/openenclave/pull/2656

jhand2:
Thanks @achamayou 

Closing this in favor of #2656

PR #2578, opened 2020-02-19
Cleanup simulation mode to use the GetThreadBinding function

anakrish:
This is now possible since oe_enter_sim correctly restores the GS/FS registers.


anakrish:
bors try

PR #2579, opened 2020-02-19
Attestation registered attester/verifier query and selection

yentsanglee:
Attester/verifier plugins can be registered and unregistered but functions are also needed to query registered attester/verifier plugins. 2 functions added:
1. oe_get_registered_attester_format_ids()
2. oe_get_registered_verifier_format_ids()

Helper function to select an evidence format from a format id list is added. The function is to help attester select one of the formats that it is able to generate evidence with:
1. oe_select_attester_evidence_format()

Some evidence formats do not have the concept of local or remote attestation. The second parameter 'flags' in oe_get_evidence() is removed. Local and remote attestations have their own format IDs. I.e.
1. OE_SGX_LOCAL_ATTESTATION_UUID
2. OE_SGX_ECDSA_P256_ATTESTATION_UUID


include/openenclave/attestation/plugin.h
```diff
@@ -293,16 +291,21 @@ oe_result_t oe_unregister_attester(oe_attester_t* plugin);
  */
 oe_result_t oe_unregister_verifier(oe_verifier_t* plugin);
 
+#if (OE_API_VERSION < 3)
+#define oe_get_evidence oe_get_evidence_v2
+#define oe_get_evidence_v2(p1, p2, p3, p4, p5, p6, p7, p8, p9, p10) \
+    oe_get_evidence_v3(p1, p3, p4, p5, p6, p7, p8, p9, p10)
+#else
+#define oe_get_evidence oe_get_evidence_v3
```
jhand2: @radhikaj Is there a place where we're queuing release notes for v0.9? We'll need to call out the incremented OE_API_VERSION in the notes. `#define OE_API_VERSION 2` will avoid the breakage.

shnwc:
Function oe_get_evidence_v3() shall provide the capability to allow its caller to pass the parameter pair (report_data, report_data_size) directly to in-enclave function oe_get_report_v2(), for inclusion in the output evidence. The current implementation of function _get_evidence() in enclave/sgx/attester.c can't meet this requirement.

include/openenclave/attestation/plugin.h
```diff
@@ -293,16 +291,21 @@ oe_result_t oe_unregister_attester(oe_attester_t* plugin);
  */
 oe_result_t oe_unregister_verifier(oe_verifier_t* plugin);
 
+#if (OE_API_VERSION < 3)
+#define oe_get_evidence oe_get_evidence_v2
+#define oe_get_evidence_v2(p1, p2, p3, p4, p5, p6, p7, p8, p9, p10) \
+    oe_get_evidence_v3(p1, p3, p4, p5, p6, p7, p8, p9, p10)
```
jhand2: @CodeMonkeyLeet this mechanism allows for calling `oe_get_evidence()` with the old parameters but it still breaks the semantics (flags will just be ignored). Is this considered ok?

include/openenclave/attestation/plugin.h
```diff
@@ -100,7 +97,6 @@ struct _oe_attester
      */
     oe_result_t (*get_evidence)(
         oe_attester_t* context,
-        uint32_t flags,
```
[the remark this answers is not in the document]
yentsanglee: This only breaks plugin authors which are a smaller group of people. One of the header installation I did in this PR is to fix plugin development. Without that fix, plugin will not build. Since we've not heard of any complaint about that, it's probably safe to assume there are very few that are actively developing plugin.
Plugin APIs were publish in the last release which was a month ago. The longer we delay the change, the harder it will be to introduce it.
If the consensus suggests the same API_VERSION should be done, we can add that. I just don't want to make it more complicated than necessary.

include/openenclave/attestation/plugin.h
yentsanglee: The removed flags is used to indicate if the attestation is for local or remote. When the flags parameter is ignored, default is remote.

---
In reply to: [383453867](https://github.com/openenclave/openenclave/pull/2579#discussion_r383453867) [](ancestors = 383453867)

include/openenclave/attestation/plugin.h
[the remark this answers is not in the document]
jhand2: @CodeMonkeyLeet any thoughts on versioning the plugin API? I think we should treat these APIs the same we do our normal public API surface.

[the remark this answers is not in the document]
yentsanglee: This only breaks plugin authors which are a smaller group of people. One of the header installation I did in this PR is to fix plugin development. Without that fix, plugin will not build. Since we've not heard of any complaint about that, it's probably safe to assume there are very few that are actively developing plugin.
Plugin APIs were publish in the last release which was a month ago. The longer we delay the change, the harder it will be to introduce it.
If the consensus suggests the same API_VERSION should be done, we can add that. I just don't want to make it more complicated than necessary.

---
In reply to: [383454992](https://github.com/openenclave/openenclave/pull/2579#discussion_r383454992) [](ancestors = 383454992)

include/openenclave/attestation/plugin.h
jhand2: This still changes how the function works. If I call oe_get_evidence with the local flag, it will just be ignored and I'll get the remote version. I consider that still broken even the API version define is used.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -838,6 +878,62 @@ oe_result_t oe_verify_evidence(
...
+ * oe_get_registered_attester_format_ids
+ *
+ * Get the evidence format identifiers of all registered attesters.
+ *
+ * @param[out] format_ids The list of the evidence format ids of the registered attesters.
+ * @param[out] format_ids_length The length of the evidence format ids list.
+ * @retval OE_OK on success.
+ * @retval Otherwise, returns the error code the plugin's function.
+ */
+oe_result_t oe_get_registered_attester_format_ids(
```
CodeMonkeyLeet: What's the use case for these new functions? Can you please illustrate them?

docs/DesignDocs/CustomAttestation.md
```diff
@@ -764,7 +804,7 @@ oe_result_t oe_unregister_verifier(
  * @retval OE_NOT_FOUND The plugin does not exist.
  */
 oe_result_t oe_get_evidence(
-    const uuid_t* evidence_format_uuid,
+    const uuid_t* evidence_format_id,
```
CodeMonkeyLeet: What is the relationship between the evidence_format_id and the format_id of the plugins? (i.e. why call them different things?)

docs/DesignDocs/CustomAttestation.md
```diff
@@ -646,9 +646,29 @@ typedef void* oe_sgx_plugin_opt_params;
  * Helper function that returns the SGX attester that can then be sent to
  * `oe_register_attester`.
  *
+ * @deprecated Instead, use oe_sgx_plugin_local_attester()
```
CodeMonkeyLeet: Please update this whole document, or better yet, submit a new design document that highlights the differences between the previous custom attestation model and the proposed changes. The API updates here are insufficient and confusing as they contradict things elsewhere in the doc (e.g. this doc should probably not have any SGX plugin information, that can just be looked up in code directly rather than keeping this in sync with actual implementation)

Things I'm looking to be explained:
* Why is there the shift to splitting local & remote attesters into separate implementations?
* What does it mean for a dev? Do I always register both? Why can I use attestation (e.g. samples) without explicitly registering either today? (i.e. is there a fallback to default sgx plugins even if I don't register them but register some other plugins?)
 
A sample is really necessary here and if there is not one already, please file an issue to add one.

docs/DesignDocs/CustomAttestation.md
```diff
@@ -646,9 +646,29 @@ typedef void* oe_sgx_plugin_opt_params;
...
+/**
+ * Helper function that returns the SGX local attester that can then be sent to
+ * `oe_register_attester`.
+ *
+ * @retval A pointer to the SGX attester. This function never fails.
+ */
+oe_attester_t* oe_sgx_plugin_local_attester(void);
+
+/**
+ * Helper function that returns the SGX ECDSA P256 attester that can then be sent to
```
CodeMonkeyLeet: Why call this the ECDSA P256 attester? Keep in mind that most developers will not be familiar with this as a reference to the DCAP-based remote attestation workflow. It's also not clear why the other plugin is called 'local_attester' if this one is referred to by the signature algorithm. 

In both cases, the doxygen comments need to be much more descriptive about what the attesters/verifiers actually do, which will help make up for only so much information you can put in a function name.

tests/attestation_plugin/plugin/tests.c
```diff
@@ -105,6 +105,191 @@ static void _test_and_unregister_verifier()
     OE_TEST(oe_unregister_verifier(&mock_verifier2) == OE_NOT_FOUND);
 }
 
+static bool _find_format_id(
```
CodeMonkeyLeet: Please consider update the attestation_plugin tests so that you can run both v2 and v3 configurations (i.e. update the makefile so it builds both variants) so that they will get tested through CI/CD until the older variant is removed, otherwise, it's likely CI/CD will not catch regressions in the v2 API behavior as it only runs with the default build configuration. The v2 configuration tests should also continue to test the old semantics (i.e. preserve the old tests so you know you have parity testing)

CodeMonkeyLeet:
@yentsanglee Please also ensure that all of your commits are signed off for [DCO](https://github.com/probot/dco)

include/openenclave/attestation/plugin.h
yentsanglee: Will ensure OE_EVIDENCE_FLAGS_LOCAL_ATTESTATION is honor when OE_API_VERSION is 2.

Plugin design requires the caller to manually register attester/verifier. This PR does not change the behavior.

---
In reply to: [383582536](https://github.com/openenclave/openenclave/pull/2579#discussion_r383582536) [](ancestors = 383582536)

docs/DesignDocs/CustomAttestation.md
dthaler: I think local vs remote should NOT be split into separate APIs that enclave apps call, as I think in general there is not a clean distinction between those terms.  We don't want SGX-specific things in a generic API.  (It's fine if it's in plugin-internal code.)

include/openenclave/attestation/plugin.h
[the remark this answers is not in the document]
dthaler: These appear to be public (or experimental) APIs and should follow the same API guidelines for public or experimental APIs. (See discussion in PR #2163)

include/openenclave/attestation/plugin.h
```diff
@@ -320,9 +323,8 @@ oe_result_t oe_unregister_verifier(oe_verifier_t* plugin);
  * @retval OE_NOT_FOUND The plugin does not exist or has not been registered.
  * @retval Otherwise, returns the error code the plugin's function.
  */
-oe_result_t oe_get_evidence(
-    const oe_uuid_t* evidence_format_uuid,
-    uint32_t flags,
+oe_result_t oe_get_evidence_v3(
```
dthaler: Why is this API *not* marked @experimental?   I think it (and its predecessors) should be.

include/openenclave/attestation/plugin.h
```diff
@@ -425,6 +427,64 @@ oe_result_t oe_verify_evidence(
...
+ *
+ * Get the unique identifiers of all registered attesters.
+ *
+ * @param[out] format_ids The list of the format ids of the registered
+ * attesters.
+ * @param[out] format_ids_length The length of the format ids list.
+ * @retval OE_OK on success.
+ * @retval Otherwise, returns the error code the plugin's function.
+ */
+oe_result_t oe_get_registered_attester_format_ids(
```
dthaler: Is this an API for apps to call?  Why would an app ever want to call this?   New APIs should have sample code not just tests.

docs/DesignDocs/CustomAttestation.md
dthaler: Thinking more, I think even OE_EVIDENCE_FLAGS_REMOTE_ATTESTATION is problematic, and I haven't seen a definition of "local" or "remote" in this document.   E.g., if you're attesting between an enclave and a USB-attached peripheral, is that local or remote?   I think most users would consider that local.

yentsanglee: There are 3 ways to get SGX evidences (reports) - local, dcap remote, and EPID remote. Each needs different sets of input for the report to be generated. Application has to be aware of the type of attestation and pass in parameters accordingly.

---
In reply to: [384113533](https://github.com/openenclave/openenclave/pull/2579#discussion_r384113533) [](ancestors = 384113533)

yentsanglee: SGX uses terms "report" and "quote" while OE uses "local report" and "remote report" in attestation API. To follow the same terminology, I think we should continue to use local. The remote will be replaced with "DCAP". More SGX remote attestation format such as EPID will join the plugin family.

---
In reply to: [384119063](https://github.com/openenclave/openenclave/pull/2579#discussion_r384119063) [](ancestors = 384119063)

dthaler: > There are 3 ways to get SGX evidences (reports) - local, dcap remote, and EPID remote. 
> Each needs different sets of input for the report to be generated. Application has to be aware of the type of attestation and pass in parameters accordingly.

Where is that documented in this document, and how is an application supposed to know which to use?   Where is the sample that works with all three that shows how it knows what to use?  (This seems like a bad model to me.)

include/openenclave/attestation/plugin.h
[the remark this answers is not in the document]
yentsanglee: 'flags' were designed specifically to indicate local or remote attestation for SGX. I reviewed other attestation designs and see no reason to keep the 'flags' parameter.

---
In reply to: [383578425](https://github.com/openenclave/openenclave/pull/2579#discussion_r383578425) [](ancestors = 383578425)

docs/DesignDocs/CustomAttestation.md
Yufei-Jiang-OpenSource: I have the similar concerns on the local/remote splitting and the naming here. I may need to think more to appraise the pros and cons of this local/remote splitting. But my feeling is that even if we choose to go with the local/remote splitting approach, we still need better names for those functions.

docs/DesignDocs/CustomAttestation.md
shnwc: I see the "flags" in oe_get_evidence() as a carry-over from oe_get_report(), which does not take in evidence format UUID. For SGX, oe_get_report() only generates two types of reports: local, or remote (which is ECDSA), as set by the “flags” parameter.

In oe_get_evidence(), UUID is used to set the format of the evidence to be generated. For SGX, UUID could be OE_SGX_LOCAL_ATTESTATION_UUID, 
OE_SGX_ECDSA_P256_ATTESTATION_UUID, and other formats as their support is added. With the UUID, the “flags” parameter becomes redundant and could even conflict with the UUID parameter. For example, in oe_get_evidence_v2(), what if the UUID is OE_SGX_ECDSA_P256_ATTESTATION_UUID while the “flags” is OE_EVIDENCE_FLAGS_LOCAL_ATTESTATION? It becomes impossible for oe_get_evidence_v2() to honor both parameters. Comments for this PR suggest that “flags” should take precedence.

New version oe_get_evidence() removes the carry-over and resolves the conflict by removing the redundant “flags” parameter.

shnwc:
> Please document the reasoning for the public API changes, and also how a developer should deal with the breaking API, migrate to the new APIs and how the additional non-breaking APIs should be used.

I see the "flags" in oe_get_evidence() as a carry-over from oe_get_report(), which does not take in evidence format UUID. For SGX, oe_get_report() only generates two types of reports: local, or remote (which is ECDSA), as set by the “flags” parameter.

In oe_get_evidence(), UUID is used to set the format of the evidence to be generated. For SGX, UUID could be OE_SGX_LOCAL_ATTESTATION_UUID, 
OE_SGX_ECDSA_P256_ATTESTATION_UUID, and other formats as their support is added. With the UUID, the “flags” parameter becomes redundant and could even conflict with the UUID parameter. For example, in oe_get_evidence_v2(), what if the UUID is OE_SGX_ECDSA_P256_ATTESTATION_UUID while the “flags” is OE_EVIDENCE_FLAGS_LOCAL_ATTESTATION? It becomes impossible for oe_get_evidence_v2() to honor both parameters. Comments for this PR suggest that “flags” should take precedence.

New version oe_get_evidence() removes the carry-over and resolves the conflict by removing the redundant “flags” parameter.

tests/attestation_plugin/plugin/tests.c
```diff
@@ -318,7 +566,11 @@ void test_runtime()
 
 void register_verifier()
 {
+#if (OE_API_VERSION < 3)
```
jazzybluesea: Why is this version specific? Can we test both verifiers in both API versions?

common/attest_plugin.c
```diff
@@ -453,3 +453,115 @@ oe_result_t oe_free_claims_list(oe_claim_t* claims, size_t claims_length)
...
+    {
+        tmp_format_ids = (oe_uuid_t*)oe_malloc(sizeof(oe_uuid_t) * node_count);
+        if (tmp_format_ids == NULL)
+            OE_RAISE(OE_OUT_OF_MEMORY);
+
+        node_index = 0;
+        cur = *list;
+        while (cur && node_index < node_count)
+        {
+            memcpy(
```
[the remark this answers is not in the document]
jhand2: UUIDs have the added benefit that their purpose is well understood and they are easy to generate (many tools can generate random UUIDs that you can just copy and paste).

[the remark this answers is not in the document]
jhand2: Agreed that collision is probably not a concern, but the semantics of UUIDs are simpler to work with.

kuanggang-git:

PR #2580, opened 2020-02-19
[RFC] Preview VS Code extension for Open Enclave

HernanGatta:
# Overview

A preview extension for Visual Studio Code for the Open Enclave SDK.

The extension offers a command titled "New Open Enclave Solution". When invoked, the developer can choose between two solution types: Standalone and Edge Container.

The Standalone solution bootstraps a CMake template that allows a developer to build and debug a host and enclave pair for Intel SGX and OP-TEE on ARM TrustZone, both for the Scalys Grapeboard and for an QEMU emulated board. VS Code Tasks are provided to automate the configuration and build steps for all platforms. Additionally, VS Code Launch Configurations are generated for debugging the host and enclave for each platform, too.

The Edge Container solution bootstraps a CMake template that allows a developer to build a host and enclave pair for OP-TEE on ARM TrustZone, both for the Scalys Grapeboard and for an emulated QEMU environment. In conjunction with the VS Code extension for Azure IoT Edge, an IoT Edge module can be built, published, and deployed to a fleet of devices that can leverage TEEs on the edge via Azure IoT Hub or Azure IoT Central.

For both solution types, no code changes are necessary to switch between platforms: all it takes is a rebuild.

**Note:** This extension used to live in the `feature.new_platforms` branch, and was recently modified to work with the `master` branch. The previous version of the extension was showcased on [Channel 9](https://youtu.be/5NbzIDThr1E) (43:41). This version includes Intel SGX support, too.

# Solution Types

Both solution types generate a CMake project. The resulting CMake project offers two options that the developer can change to reconfigure the project, in the CMake sense, for the different available TEEs:

* `OE_TEE`:
   This option can be either `SGX` or `OP-TEE`.
* `OE_OPTEE_PLATFORM`:
   If `OE_TEE` is `OP-TEE` this option can be `QEMU-ARMv8` or `Grapeboard`, and is unused otherwise.

If `OE_TEE` is `SGX`, the project searches for the Open Enclave CMake package in standard locations, and finds the installation of the Open Enclave DEB package.

If `OE_TEE` is `OP-TEE`, the project searches for the Open Enclave CMake package in a location provided by the extension as a function of the value of `OE_OPTEE_PLATFORM`. That is, the extension bundles two builds of the SDK: one for OP-TEE on ARM TrustZone on a QEMU emulated board, and one for OP-TEE on ARM TrustZone on the Scalys Grapeboard.

Once the appropriate Open Enclave CMake package is imported, the build toolchain is configured appropriately along with the rest of the project's targets.

Both solution types must build inside a Linux environment and generate Linux hosts only. Namely, the following resulting host/enclave combinations are supported:

* Linux Host / Intel SGX Enclave / FLC
* Linux Host / OP-TEE on ARM TZ Enclave / Scalys Grapeboard
* Linux Host / OP-TEE on ARM TZ Enclave / QEMU Emulated Board

## Standalone Solution

The Standalone solution type is so called because it builds directly on the same environment where the VS Code extension is installed. Additionally, the resulting host and enclave pair are not packaged in any way.

This solution type is only available when the extension is running on VS Code on Linux. Alternatively, if VS Code is running on Windows, but the Remote Development extension is installed and the developer has connected it to a remote Linux system, then the extension can be installed on that remote Linux system. If this is the case, then the Standalone solution type is available.

## Edge Container Solution

The Edge Container solution type generates a host and enclave pair just like the Standalone solution type. The difference is that the host includes code to communicate with Azure IoT Hub and Azure IoT Central (the two use the same protocol, so they are indistinguishable as far as the host is concerned).

Furthermore, this solution type also generates manifests that can be consumed by the VS Code extension for Azure IoT Edge. These manifests allow the solution to be built and packaged the same way IoT Edge modules are: as Docker containers.

Seeing as this solution type generates a project that builds inside of and produces Docker containers, that these Docker containers run Linux, and that Linux Docker containers run on Windows, if the extension is installed on Windows without a Remote Development system, this solution type remains available.

## Availability Summary

* Windows Development Host
   * No Remote Development connection
      * Edge Container solution only
   * Remote Development connection to Linux
      * Both solution types available
* Linux Development Host
   * Both solution types available

# Implementation

Below are described some of the details of the inner-workings of the extension.

## Debugging

A project generated from the Standalone solution type may be debugged. To this end, three VS Code Launch Configurations are generated with the project:

1. (oegdb) Launch Intel SGX (Hardware)
2. (oegdb) Launch Intel SGX (Simulation)
3. (gdb) Launch QEMU (AArch64/ARMv8-A)

### Intel SGX

For the first two Launch Configurations, VS Code's C/C++ debugger is attached to GDB via `oegdb`. A command line argument (`--debug`) is passed to the host so that it specify `OE_ENCLAVE_FLAG_DEBUG` as an enclave creation flag. Additionally, for the simulation Launch Configuration, another flag (`--simulate`) is also passed to the host so that it also include `OE_ENCLAVE_FLAG_SIMULATE` as another enclave creation flag.

### OP-TEE on ARM TrustZone

For the last Launch Configuration, first, VS Code launches QEMU, which exposes a GDB server on a known port. Then, VS Code's debugger is attached to this GDB server via the "Native Debug" extension.

In VS Code's built-in Terminal, the emulated board runs through its boot sequence. Once complete, the developer logs in at the prompt. By default, the directory where the build output is located is shared into and auto-mounted by the emulator. As such, the developer can navigate to this location and execute the host. The debugging symbols for both the host and the enclave will have been preloaded into GDB.

## Included Binaries

The extension includes a file named `devkit.tar.gz`. This tarball includes the two builds of the SDK for OP-TEE on ARM TrustZone (one for each target board), and the files necessary to emulate an ARM TrustZone-capable board.

When the extension first requires using the files in this tarball, it extracts them to the extension's global storage location. Then, it places a symlink to this location under the bootstrapped project's final directory. The project references the contents of the tarball for building and debugging.

# Future Work

The extension as-is is for preview purposes only, and to showcase the cross-platform capabilities of the SDK.

## Binary Distribution

Ideally, the extension would need not include builds of the SDK.

This poses a curious dilemma: OP-TEE is custom-built for every board. Hence, no two builds of OP-TEE are necessarily identical. As a result, the SDK must be built to target that build of OP-TEE, too. Hence, either the SDK is distributed as source only for OP-TEE on ARM TrustZone, or, with care, we might be able to produce portable binaries, though this is uncertain and much more research is required.

A third alternative is a middle ground where we distribute binaries of the SDK for select OP-TEE platforms for which we distribute the firmware, too. As long as developers use that firmware, then we can distribute SDK binaries that we know will work.

In all cases, more work is required to allow developers to bring their own OP-TEE, and build the SDK accordingly.

A side-effect of distributing binaries with the extension is that the resulting project depends on those binaries, too. This is troublesome for source control. Perhaps we could place the contents of the `devkit.tar.gz` in a Git repository and have developers consume that. Of course, that presumes developers will always use Git.

## Solution Types

Having two solution types is not ideal. Usually, a developer would start from a what is here dubbed a Standalone project, then create a wrapping project to package that into an IoT Edge solution, or into some other form of distribution, say CPack, or a NuGet package, or something else. As it is, the extension forces the developer to choose before starting.

# Known Issues

1. Debugging the host under the QEMU emulated board does not work:
   1. This is likely ASLR-related.
1. There is no debug Launch Configuration for Edge Container projects.


devex/vscode-extension/README.md
```diff
@@ -0,0 +1,383 @@
...
+
+For more information on the technology,
+[see here](https://github.com/openenclave/openenclave).
+
+Currently, this extension supports the following combinations of project types,
+and operating system and trusted execution environments:
+
+1. Linux Development Machine:
+   1. Standalone Project Targets:
+      1. Linux Host w/. Intel SGX Enclave
```
dthaler: nit: would be better to spell out "with" which is only 1 character longer than "w/." and is more understandable by non-native English speakers.

devex/vscode-extension/README.md
```diff
@@ -0,0 +1,383 @@
...
+  sudo apt install -y apt-transport-https az-dcap-client binfmt-support      \
+    build-essential clang-7 docker.io g++-aarch64-linux-gnu                  \
+    gcc-aarch64-linux-gnu gdb gdb-multiarch libc6 libc6-dev:arm64 libfdt1    \
+    libglib2.0-0 libpcre3 libpixman-1-0 libprotobuf10 libsgx-dcap-ql         \
+    libsgx-dcap-ql-dev libsgx-enclave-common libsgx-enclave-common-dev       \
+    libssl-dev libssl-dev:arm64 libstdc++6 open-enclave python python-crypto \
+    python-pip qemu-user-static zlib1g
+  ```
+
+* If you have Intel SGX-capable hardware, install the Intel SGX DCAP driver:
```
dthaler: Maybe hotlink "Intel SGX-capable hardware" to a section or page that explains how you can tell?
Also I thought DCAP was only needed on SGX1+FLC systems, not SGX1 systems, no?

devex/vscode-extension/README.md
```diff
@@ -0,0 +1,383 @@
...
+### Windows
+
+* [Install Docker](https://docs.docker.com/install/)
+
+* Ensure sure that Git long paths are enabled:
+  `git config --global core.longpaths true`
+
+* Ensure Docker is configured to use Linux containers.  See Docker help
+  regarding **Switch between Windows and Linux containers**
+  [here](https://docs.docker.com/docker-for-windows/).
```
dthaler: recommend replacing "[here]" with "the [Get started with Docker for Windows] page", for accessibility reasons.  (Short version: one example is when you tab through links and a screen reader for the blind will read the text.  Tabbing into this link will just say "here" which is useless.   Other examples exist and lots more discussion is available by searching on accessibility and "click here").

devex/vscode-extension/assets/baseSolutionTemplateFolder/[[project-name]].edl
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
Comment (dthaler) [inline devex/vscode-extension/assets/baseSolutionTemplateFolder/[[project-name]].edl]: If this is going into the master branch it should be:
Please remove all uses of copyright "Microsoft Corporation" for code intended to be in the main SDK repo.

devex/vscode-extension/assets/baseSolutionTemplateFolder/enc/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
dthaler: See earlier comment about replacing all occurrences of this line.

devex/vscode-extension/assets/baseSolutionTemplateFolder/enc/enc.c
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <stdio.h>
+
+// Include the trusted [[project-name]] header that is generated
+// during the build. This file is generated by calling the
+// sdk tool oe_edger8r against the [[project-name]].edl file.
```
dthaler: No underscore, it's just oeedger8r.  Same in other files, so search for all occurrences of oe_edger8r

devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c
```diff
@@ -0,0 +1,235 @@
...
+    return messageInstance;
+}
+
+static IOTHUBMESSAGE_DISPOSITION_RESULT SendEnclaveResponse(IOTHUB_MODULE_CLIENT_LL_HANDLE iotHubModuleClientHandle, char* messageBodyStr)
+{
+    IOTHUBMESSAGE_DISPOSITION_RESULT result;
+    IOTHUB_CLIENT_RESULT clientResult;
+
+    char* enclaveMessage = (char*)malloc(512 * sizeof(char));
+    int enclaveResult = call_enclave(messageBodyStr, enclaveMessage, 512);
```
Comment (dthaler) [inline devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c]: should this check for null enclaveMessage before doing call_enclave?

devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c
```diff
@@ -0,0 +1,235 @@
...
+            if (clientResult != IOTHUB_CLIENT_OK)
+            {
+                IoTHubMessage_Destroy(messageInstance->messageHandle);
+                free(messageInstance);
+                printf("IoTHubModuleClient_LL_SendEventToOutputAsync failed on sending msg#=%zu, err=%d\n", messagesReceivedByInput1Queue, clientResult);
+                result = IOTHUBMESSAGE_ABANDONED;
+            }
+            else
+            {
+                result = IOTHUBMESSAGE_ACCEPTED;
```
Comment (dthaler) [inline devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c]: where is messageInstance freed in this case?

devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c
```diff
@@ -0,0 +1,235 @@
...
+    unsigned char* messageBody;
+    size_t contentSize;
+
+    if (IoTHubMessage_GetByteArray(message, &messageBody, &contentSize) != IOTHUB_MESSAGE_OK)
+    {
+        messageBody = "<null>";
+    }
+
+    char* messageBodyStr = (char*)malloc(contentSize + 1);
+    memcpy(messageBodyStr, messageBody, contentSize);
```
Comment (dthaler) [inline devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c]: check for messageBodyStr == NULL

devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c
```diff
@@ -0,0 +1,235 @@
...
+    messageBodyStr[contentSize] = '\0';
+
+    printf("Received Message [%zu]\r\n Data: [%s]\r\n",
+            messagesReceivedByInput1Queue, messageBodyStr);
+
+    IOTHUBMESSAGE_DISPOSITION_RESULT result = SendEnclaveResponse(iotHubModuleClientHandle, messageBodyStr);
```
Comment (dthaler) [inline devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c]: Where is messageBodyStr freed?

devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c
```diff
@@ -0,0 +1,235 @@
...
+        printf("Failed to initialize the platform.\r\n");
+        iotHubModuleClientHandle = NULL;
+    }
+    else if ((iotHubModuleClientHandle = IoTHubModuleClient_LL_CreateFromEnvironment(MQTT_Protocol)) == NULL)
+    {
+        printf("ERROR: IoTHubModuleClient_LL_CreateFromEnvironment failed\r\n");
+    }
+    else
+    {
+        // Uncomment the following lines to enable verbose logging.
```
Comment (dthaler) [inline devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c]: Commented out code is not very maintainable... what if the signature changes?  You won't notice.
Replace with an ifdef, and make sure there's some test that verifies it builds with that define set.

devex/vscode-extension/assets/standaloneSolutionTemplateFolder/host/main.c
```diff
@@ -0,0 +1,27 @@
...
+int main(int argc, const char* argv[])
+{
+    int result = create_enclave(argc, argv);
+    if (result != 0)
+    {
+        return result;
+    }
+    char* enclaveMessage = (char*)malloc(512 * sizeof(char));
+    result = call_enclave("\"Process In Enclave\"", enclaveMessage, 512);
+    if (result != 0)
```
dthaler: add before line 20:
free(enclaveMessage);

devex/vscode-extension/marketplace-homepage.md
```diff
@@ -0,0 +1,383 @@
...
+
+For more information on the technology,
+[see here](https://github.com/openenclave/openenclave).
+
+Currently, this extension supports the following combinations of project types,
+and operating system and trusted execution environments:
+
+1. Linux Development Machine:
+   1. Standalone Project Targets:
+      1. Linux Host w/. Intel SGX Enclave
```
dthaler: Replace "w/." -> "with"

devex/vscode-extension/marketplace-homepage.md
```diff
@@ -0,0 +1,383 @@
...
+  sudo apt install -y apt-transport-https az-dcap-client binfmt-support      \
+    build-essential clang-7 docker.io g++-aarch64-linux-gnu                  \
+    gcc-aarch64-linux-gnu gdb gdb-multiarch libc6 libc6-dev:arm64 libfdt1    \
+    libglib2.0-0 libpcre3 libpixman-1-0 libprotobuf10 libsgx-dcap-ql         \
+    libsgx-dcap-ql-dev libsgx-enclave-common libsgx-enclave-common-dev       \
+    libssl-dev libssl-dev:arm64 libstdc++6 open-enclave python python-crypto \
+    python-pip qemu-user-static zlib1g
+  ```
+
+* If you have Intel SGX-capable hardware, install the Intel SGX DCAP driver:
```
dthaler: See earlier question about SGX1 (non-FLC) systems.

devex/vscode-extension/src/common/utility.ts
```diff
@@ -0,0 +1,180 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+"use strict";
+import * as fse from "fs-extra";
+import * as path from "path";
+import * as vscode from "vscode";
+import { AzureAccount, AzureSession } from "../typings/azure-account.api";
```
dthaler: Is this entire file Azure-specific?   Or does it combine Azure-specific and non-Azure-specific code?
If the former, then it would be better to rename the file, or put it under an azure subdirectory.
If the latter, then it would be better to split the file.  (Ok to do in a separate PR though.)

devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c
Comment (HernanGatta) [reply devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c]: The call to `IoTHubModuleClient_LL_SendEventToOutputAsync` queues the `messageInstance` to be sent. When that happens, or fails, `SendConfirmationCallback` is called by the IoT Hub SDK. That function frees the `messageInstance`

devex/vscode-extension/README.md
HernanGatta: Changed.

devex/vscode-extension/README.md
HernanGatta: Added link, and yes, you are correct. Added instructions for installing the non-FLC driver.

devex/vscode-extension/README.md
HernanGatta: Fixed.

devex/vscode-extension/assets/baseSolutionTemplateFolder/[[project-name]].edl
Comment (HernanGatta) [reply devex/vscode-extension/assets/baseSolutionTemplateFolder/[[project-name]].edl]: Fixed across all files.

devex/vscode-extension/assets/baseSolutionTemplateFolder/enc/CMakeLists.txt
HernanGatta: Idem.

devex/vscode-extension/assets/baseSolutionTemplateFolder/enc/enc.c
HernanGatta: Fixed.

devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c
Comment (HernanGatta) [reply devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c]: Fixed.

devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c
Comment (HernanGatta) [reply devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c]: Indeed, fixed.

devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c
Comment (HernanGatta) [reply devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c]: Fixed.

devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c
Comment (HernanGatta) [reply devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c]: Now compiled in unconditionally, the developer can switch the global flag to enable tracing instead.

devex/vscode-extension/assets/standaloneSolutionTemplateFolder/host/main.c
HernanGatta: Added at the end.

devex/vscode-extension/marketplace-homepage.md
HernanGatta: This file is a copy of `README`, so fixed here as well.

devex/vscode-extension/marketplace-homepage.md
HernanGatta: Idem.

devex/vscode-extension/src/common/utility.ts
HernanGatta: Turns out the import is unused, so I've removed it. This file has a utility function to iterate over an array, used elsewhere for ACR support.

devex/vscode-extension/README.md
dthaler: Thanks

devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c
Comment (dthaler) [reply devex/vscode-extension/assets/edgeSolutionTemplateFolder/modules/[[project-name]]/host/main.c]: Suggest adding a // comment explaining where it's freed

devex/vscode-extension/assets/standaloneSolutionTemplateFolder/host/main.c
HernanGatta: It's not; I've cleaned up all the sample code. Fixed.

HernanGatta:
@dthaler, @radhikaj, @anakrish:

I've been thinking about whether adding the VS and VS Code extensions to the main SDK tree is a good idea, and I'm currently leaning towards no.

The question I've been pondering is that of release cycles. A new release of the SDK does not necessarily mean that a new release of the extensions is warranted. It could very well happen that from one SDK release to another, nothing worthwhile happens to the extensions. Conversely, it could be the case that we'd like to release a fix or a new feature in the extensions, but we can't have a proper release until the SDK is next versioned. We certainly could release an update to them anyway, but then the extensions would have a separate version sequence from that of the SDK, which is the point I'm trying to make.

By having the extensions separate from the SDK's tree, we can decouple them from the SDK's release cycle and we can publish new versions when we can, and when we think that a new release is warranted.

In so few words, the SDK and the extensions seem to me like separate products; coupled, but distinct.

Thoughts?

HernanGatta:
@anakrish, @BRMcLaren, there is no place to list out files to exclude from the code format check, is there?

The template system in the extension uses `[[replace-me]]` strings to mark bits that need replacement during solution generation with well-known strings.

For instance, if the solution is called `MyTestProject`, then this bit in the extension's template:

```
    // Create the enclave.
    result = oe_create_[[project-name]]_enclave(
        argv[1], OE_ENCLAVE_TYPE_AUTO, flags, NULL, 0, &g_enclave);
```

turns into:

```
    // Create the enclave.
    result = oe_create_MyTestProject_enclave(
        argv[1], OE_ENCLAVE_TYPE_AUTO, flags, NULL, 0, &g_enclave);
```

When `clang-format` comes across this template file, it complains about missing spaces around and inside the brackets, it wants `oe_create_ [[project - name]] _enclave`, which messes the template up.

I could surround such problematic bits with `// clang-format off` and `// clang-format on`, but then those comments would make it into the user's own solution, which we don't want.

I know that the `format-code` script has an `--exclude-dirs` parameter, so we could have a new file at the root of the SDK, say `.formatignore`, listing directories and files to not bother checking. Then, we could feed that into `format-code` during CI in the `check-ci` script and in the pre-commit hook.

Thoughts?

anakrish:
@HernanGatta 
Can we have a different extension for the template files? Say `enc.c.in` so that they are not considered by clang-format? Is that a feasible approach?

I noticed that the template files have MIT license. Does that mean that we are forcing the MIT license on users that start off from a template generated project?

HernanGatta:
@anakrish, I hadn't considered having different extensions; it should be possible to do. Thank you for the idea.

And, no, we don't want to force the license on users. I think the original impetus for adding the license headers was because this is code being committed into the SDK. Being a template, however, the logic does not apply.

Curiously, the Azure IoT Edge extension for VS Code puts the copyright and license headers into the user's generated solution. Thinking about it now, it is odd. I don't recall seeing headers like that being added into template solutions generated by VS for .NET programs, for example.

I for one lean toward leaving the files header-less, and letting the user decide. Along with @dthaler, what are your thoughts?

HernanGatta:
bors try

dthaler:
> I for one lean toward leaving the files header-less, and letting the user decide. Along with @dthaler, what are your thoughts?

Yes I believe that generated files should be header-less.  That's what I have done in the VS extension.

HernanGatta:
bors try

HernanGatta:
bors try

HernanGatta:
bors try

HernanGatta:
bors cancel

HernanGatta:
bors try-

PR #2591, opened 2020-02-20
Refactor edl wrapper files

jhand2:
All internal edl files have wrappers in order to register ecall and ocall tables since out of the box edl only supports compiling a single edl file per enclave application.

Refactor to rely less on these wrappers with the goal of eventually removing these wrappers entirely.

These wrappers cannot be removed until we publish the edl files with the SDK and require enclave developers to import them into their own EDL.

### DIFF
--- a/tools/oeedger8r/src/Sources.ml
+++ b/tools/oeedger8r/src/Sources.ml
@@ -763,6 +763,8 @@ let generate_trusted (ec : enclave_content) (ep : Intel.Util.edger8r_params) =
   [
     sprintf "#include \"%s_t.h\"" ec.file_shortnm;
     "";
+    "#define OE_NEED_STDC_NAMES";
+    "";
     "#include <openenclave/edger8r/enclave.h>";
     "";
     "#include <stdlib.h>";
--- a/enclave/core/sgx/sgx_t_wrapper.c
+++ b/enclave/core/sgx/sgx_t_wrapper.c
@@ -5,9 +5,6 @@
 
 #include <openenclave/enclave.h>
 
-#include <openenclave/corelibc/stdio.h>
-#include <openenclave/corelibc/string.h>
-#include <openenclave/edger8r/enclave.h>
 #include <openenclave/internal/calls.h>
 #include <openenclave/internal/thread.h>
 
--- a/enclave/core/tee_t_wrapper.c
+++ b/enclave/core/tee_t_wrapper.c
@@ -1,13 +1,7 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
-#define OE_NEED_STDC_NAMES
-
 #include <openenclave/enclave.h>
-
-#include <openenclave/corelibc/stdio.h>
-#include <openenclave/corelibc/string.h>
-#include <openenclave/edger8r/enclave.h>
 #include <openenclave/internal/calls.h>
 #include <openenclave/internal/thread.h>
 
--- a/host/sgx/sgx_u_wrapper.c
+++ b/host/sgx/sgx_u_wrapper.c
@@ -3,7 +3,6 @@
 
 #include <openenclave/host.h>
 
-#include <openenclave/edger8r/host.h>
 #include <openenclave/internal/calls.h>
 #include <openenclave/internal/raise.h>
 
--- a/host/syscall_u_wrapper.c
+++ b/host/syscall_u_wrapper.c
@@ -3,7 +3,6 @@
 
 #include <openenclave/host.h>
 
-#include <openenclave/edger8r/host.h>
 #include <openenclave/internal/calls.h>
 #include "hostthread.h"
 
--- a/host/tee_u_wrapper.c
+++ b/host/tee_u_wrapper.c
@@ -3,7 +3,6 @@
 
 #include <openenclave/host.h>
 
-#include <openenclave/edger8r/host.h>
 #include <openenclave/internal/calls.h>
 #include <openenclave/internal/raise.h>
 
--- a/include/openenclave/edger8r/common.h
+++ b/include/openenclave/edger8r/common.h
@@ -66,24 +66,24 @@ done:
     return result;
 }
 
-#define OE_ADD_SIZE(total, size)                                 \
-    do                                                           \
-    {                                                            \
-        if (sizeof(total) > sizeof(size_t) && total > SIZE_MAX)  \
-        {                                                        \
-            _result = OE_INVALID_PARAMETER;                      \
-            goto done;                                           \
-        }                                                        \
-        if (sizeof(size) > sizeof(size_t) && size > SIZE_MAX)    \
-        {                                                        \
-            _result = OE_INVALID_PARAMETER;                      \
-            goto done;                                           \
-        }                                                        \
-        if (oe_add_size((size_t*)&total, (size_t)size) != OE_OK) \
-        {                                                        \
-            _result = OE_INTEGER_OVERFLOW;                       \
-            goto done;                                           \
-        }                                                        \
+#define OE_ADD_SIZE(total, size)                                   \
+    do                                                             \
+    {                                                              \
+        if (sizeof(total) > sizeof(size_t) && total > OE_SIZE_MAX) \
+        {                                                          \
+            _result = OE_INVALID_PARAMETER;                        \
+            goto done;                                             \
+        }                                                          \
+        if (sizeof(size) > sizeof(size_t) && size > OE_SIZE_MAX)   \
+        {                                                          \
+            _result = OE_INVALID_PARAMETER;                        \
+            goto done;                                             \
+        }                                                          \
+        if (oe_add_size((size_t*)&total, (size_t)size) != OE_OK)   \
+        {                                                          \
+            _result = OE_INTEGER_OVERFLOW;                         \
+            goto done;                                             \
+        }                                                          \
     } while (0)
 
 /**
--- a/syscall/syscall_t_wrapper.c
+++ b/syscall/syscall_t_wrapper.c
@@ -6,8 +6,6 @@
 #include <openenclave/enclave.h>
 
 #include <openenclave/corelibc/stdio.h>
-#include <openenclave/corelibc/string.h>
-#include <openenclave/edger8r/enclave.h>
 #include <openenclave/internal/calls.h>
 #include <openenclave/internal/thread.h>
 


tools/oeedger8r/src/Sources.ml
```diff
@@ -945,6 +944,10 @@ let generate_untrusted (ec : enclave_content) (ep : Intel.Util.edger8r_params) =
     "#include <string.h>";
     "#include <wchar.h>";
     "";
+    "#if defined(__GNUC__)";
```
anakrish: Why does oeedger8r need to generate this?

tools/oeedger8r/src/Sources.ml
```diff
@@ -762,12 +762,11 @@ let generate_trusted (ec : enclave_content) (ep : Intel.Util.edger8r_params) =
...
   [
     sprintf "#include \"%s_t.h\"" ec.file_shortnm;
+    "#define OE_NEED_STDC_NAMES";
     "";
     "#include <openenclave/edger8r/enclave.h>";
-    "";
-    "#include <stdlib.h>";
-    "#include <string.h>";
-    "#include <wchar.h>";
+    "#include <openenclave/corelibc/string.h>";
```
anakrish: Why do we want to introduce on dependency on oe specific headers rather than standard headers?

jhand2: Because edger8r-generated code shouldn't depend on libc.

tools/oeedger8r/src/Sources.ml
jhand2: In order to publish the internal edl files we will need to get rid of the _wrapper.c files. This means everything they do must now be done in the generated code.

The generated code doesn't forward-declare its functions which throws a warning on GCC which we treat as an error. So we can either forward declare the functions or add this. I'm fine with either.

tools/oeedger8r/src/Sources.ml
dthaler: > Because edger8r-generated code shouldn't depend on libc.

Can you point me to the design doc that explains why it shouldn't?

jhand2: No I don't believe one exists :smile: I'm happy to document it somewhere though.

jhand2: Note that what I meant (and have since edited) is that the generated code on the enclave side shouldn't depend on libc. On the host side its fine.

I'll detail why in documentation in a separate PR.

anakrish: edger8r generated code is intended primarily to be compiled in by an SDK user. 
There will be far more enclave applications using EDL than the few internal libraries that use EDL.

It should be fine for edger8r generated code to depend on standard headers; however if there is a compelling reason for it not to, we can evaluate it.

An aside, if oe provides headers that resemble standard C headers (e.g string.h, whcar.h) but do different things, then it becomes hard to reason about what edger8r generated code does.

tools/oeedger8r/src/Sources.ml
anakrish: Ah ok. Can we make that function static instead?

tools/oeedger8r/src/Sources.ml
jhand2: I don't really understand the objection.

> edger8r generated code is intended primarily to be compiled in by an SDK user.
There will be far more enclave applications using EDL than the few internal libraries that use EDL.

I am referring strictly to edger8r *generated* glue code. It doesn't matter what code enclave developers write in the ecall/ocall code. The problem is that using libc in generated code forces an enclave to depend on libc even if their application doesn't need it.

> It should be fine for edger8r generated code to depend on standard headers; however if there is a compelling reason for it not to, we can evaluate it.

Let me ask a different question. Is there a compelling reason the generated glue code needs to depend on libc?

> An aside, if oe provides headers that resemble standard C headers (e.g string.h, whcar.h) but do different things, then it becomes hard to reason about what edger8r generated code does.

Well we already do. For example `oe_strlen` is used all over the place in our codebase and is in openenclave/corelibc/string.h. The reality is we've re implemented small parts of libc for use in the enclave core so that not all enclave application need to depend on musl. We could rename all these headers to avoid confusion but its a big API breaking change so we'll need to do it more systematically.

CodeMonkeyLeet: @jhand2 I would prefer that edger8r always use the `oe_` prefixed names instead of relying on `#define OE_NEED_STDC_NAMES`. Even though it works in the wrapper today, that mechanism relies on inlining to eliminate the potential symbol collision (i.e. compiler needs to elide the corelibc passthrough implementation of `wcslen()` and only generate `oe_wcslen()`). 

It was primarily designed to enable oecore to use 3rdparty code with std libc dependencies without modification (e.g. dlmalloc) If the intention is that oeedger8r should use corelibc implementations, then let's use them explicitly and reduce the number of things that could go wrong.

tools/oeedger8r/src/Sources.ml
anakrish: In #2609 I have tweaked oeedger8r to generate 'static' keyword for local functions and removed these  workarounds in _wrapper.c files.

tools/oeedger8r/src/Sources.ml
anakrish: @jhand2 

oeedger8r generated code has been designed to be as much self-contained as possible. The reason is that edge functions are at the security boundary and we want them to be stable no matter how much rest of OE SDK changes.

oecore is NOT a stable interface; neither is it formally defined nor does it make any guarantees about stablity. Standard C library is well defined and does not change.

An example: Recently we decided to no longer publish safemath/safecrt headers (5b8ebe1). Though this PR changed the public API surface, it had zero impact on oeedger8r since the generated code had been designed to be self-contained and had its own safe math functions.

Yes. It is a good goal that edl/oeedger8r does not depend on standard C.
However it should not come at the expense of making edl/oeedger8r depend on unstable interfaces (oecore).

Currently, edger8r generated code has been designed to be self-contained with a minimal bit of dependency on standard C library. This is the reason the same edger8r generated code works on SGX as well as OPTEE. The generated code depends on
-  C header files
- functions that are part of free standing environment (e.g memcpy)
- string functions (strlen, wcslen etc)
- published oe headers

If the dependency (though tiny) with the C library has to be removed, it is preferable to continue to keep oeedger8r generated code self-contained. Instead of depending on OE core
- provide inline functions `oe_edl_strlen` that the generated code can call
- define types in EDL headers rather than oe core headers
- change edger8r to not  emit `#include <wchar.h>` if the EDL does not use wide characters.

Note: It is impossible to not depend on C library functions. Even for a `free standing` environment where no C library is available, GCC is free to generate memcpy, memset and other functions.

jhand2: Thanks for the background! I'll rework this to ensure the generated code stays self contained.

radhikaj: Thanks @jhand2 for incorporating comments.

tools/oeedger8r/src/Sources.ml
jhand2: Thanks. Will remove in the next iteration.

tools/oeedger8r/src/Sources.ml
jhand2: I didn't realize that the compiler would just fill in the implementations for stdc functions. Thanks for the heads up @anakrish. I've updated the PR to be a lot more scoped down and just address shrinking the need for the C wrappers.

jhand2:
bors try

jhand2:
@anakrish @mingweishih @radhikaj @anakrish @CodeMonkeyLeet Could I get a review on this?

Thanks!

tools/oeedger8r/src/Sources.ml
```diff
@@ -763,6 +763,8 @@ let generate_trusted (ec : enclave_content) (ep : Intel.Util.edger8r_params) =
   [
     sprintf "#include \"%s_t.h\"" ec.file_shortnm;
     "";
+    "#define OE_NEED_STDC_NAMES";
```
CodeMonkeyLeet: @jhand2 Where did we finally land on removing the use of this definition to require that generated stubs explicitly reference the `oe_*` prefixed versions of the functions? I see it's removed in enclave/core/tee_t_wrapper.c but it persists in the other wrappers.

jhand2: So @anakrish didn't want to use oe_* names because the interface isn't well-defined. I am looking into how I can just rely on the compiler to fill in all the functionality we need (which is already done today for wcslen for example).

I worry that this approach is too dependent on the compiler implementation though.

The other option is to define all the functions we need to include/openenclave/edger8r/common.h as inline functions. I don't love this either because code duplication but at least it has well-defined behavior.

jhand2: @anakrish  One thing about corelibc though. We actually do make guarantees about its stability. It is part of the public interface, so I actually think edger8r depending on corelibc functions seems ok from that perspective.

anakrish: Here is our  current documentation:
https://openenclave.io/apidocs/v0.8/files.html

That we don't even list any of the files in corelibc would indicate that it is more of an implementation detail rather than a  well-defined public API.

anakrish: My recommendation would be to add any functions that the edger8r needs to [include/edger8r/common.h](https://github.com/openenclave/openenclave/blob/71edd2cf92f52017410d5e3262628b962067560a/include/openenclave/edger8r/common.h)
and tweak the generator to use those functions. Say `oe_edl_malloc` instead of `malloc`, `oe_edl_strlen` instead of `strlen`.

This will make edger8r generated code more self-contained, while removing dependencies on libc. It also makes the requirements of the generated code very clear.

anakrish: @jhand2 Just now saw your earlier comment about adding functions to `edger8r/common.h`. So yes, I agree with you. Whether we inline or not is secondary; but the includes in `edger8r` present a well defined interface for what every TEE must implement to allow edger8r generated code to work.

For example, here is how OPTEE implements the edl functions [enclave/core/optee/hostcalls.c](https://github.com/openenclave/openenclave/blob/71edd2cf92f52017410d5e3262628b962067560a/enclave/core/optee/hostcalls.c)

CodeMonkeyLeet: I'm not sure I fully follow the discussion presented here, but I don't have specific objections to the various options presented here. It sounds like there's an agreement to use edger8r/common.h as the consolidation point for definitions that edger8r stubs have on the OE implemented libc implementations; that sounds like it probably deserves its own issue and PR.

jhand2: I guess I don't understand why `oe_edl_strlen` is better than `oe_strlen`. Do we expect code generated by EDL to work with all versions of OE no matter which version it was generated with? That seems like a big promise.

I think the freestanding libc is more complicated and less stable than we're letting on due to reasons I detailed here: https://github.com/openenclave/openenclave/pull/2767#issuecomment-606729916

jhand2: As an aside, the issue with OE_NEED_STDC_NAMES is name collision. If this is defined and you want to use full (musl) libc there are likely be name collision.

kuanggang-git:

jhand2:
This PR is superseded by a number of other PRs. Everything important done by this PR is already in master.

PR #2593, opened 2020-02-21
Link all host code with noexecstack

jhand2:
Compile all host code on linux with noexecstack to ensure the stack is not executable.

Not only is this good security practice as the Linux kernel invokes READ_IMPLIES_EXEC on mmap and mprotect for processes with executable stacks, which means read requests will be treated as read/execute requests. The SGX driver however ignores the READ_IMPLIES_EXEC flag when hooking mmap/mprotect calls which can causes spurious failures when attempting to mmap/mprotect enclave pages.

### DIFF
--- a/host/CMakeLists.txt
+++ b/host/CMakeLists.txt
@@ -452,12 +452,13 @@ add_library(oehostapp INTERFACE)
 target_link_libraries(oehostapp INTERFACE oehost)
 
 if (UNIX)
+  target_link_libraries(oehost INTERFACE
+      -Wl,-z,noexecstack)
   target_link_libraries(oehostapp INTERFACE
-      -rdynamic
-      -Wl,-z,noexecstack)
+      -rdynamic)
   target_link_libraries(oehostverify INTERFACE
-      -rdynamic
-      -Wl,-z,noexecstack)
+      -Wl,-z,noexecstack
+      -rdynamic)
 endif ()
 
 # Install targets


host/CMakeLists.txt
```diff
@@ -452,12 +452,12 @@ add_library(oehostapp INTERFACE)
...
-  target_link_libraries(oehostapp INTERFACE
-      -rdynamic
+  target_link_libraries(oehost INTERFACE
       -Wl,-z,noexecstack)
+  target_link_libraries(oehostapp INTERFACE
+      -rdynamic)
   target_link_libraries(oehostverify INTERFACE
-      -rdynamic
-      -Wl,-z,noexecstack)
+      -rdynamic)
```
CodeMonkeyLeet: `oehostverify` parallels `oehost` and does not link `oehost`, so it still needs its own copy of the `-Wl,-z,noexecstack` arguments.

Tangent: @anakrish I'm not sure why we decided to create a target library just for the purpose of tacking on the `-rdynamic` linker option. It seems like a weird indirection and it seems like an odd choice that we default all our samples to use it?

jhand2: Ah you're right, added the flag back to oehostverify.

Agreed that I don't see a reason for having both targets. If there isn't a good reason I can open an issue/pr to remove it (although it is a breaking change so maybe easier said than done). But I'll wait for Anand's response.

anakrish: @CodeMonkeyLeet  
oehostapp was done by jorkl 2 years ago https://github.com/openenclave/openenclave/commit/5f0cded986953dc765bf166b1f4f77c5d646ccf0

I am not sure why we would need the separation between oehost and oehostapp.

However, we do need the separation between oehostverify and oehostapp. The former allows attestation by a host that is not running on sgx hardware. The latter allows creating enclave applications.

host/CMakeLists.txt
```diff
@@ -452,12 +452,13 @@ add_library(oehostapp INTERFACE)
 target_link_libraries(oehostapp INTERFACE oehost)
 
 if (UNIX)
-  target_link_libraries(oehostapp INTERFACE
-      -rdynamic
+  target_link_libraries(oehost INTERFACE
```
anakrish: Thanks for this security oriented change. Always welcome.


This change is very subtle and I am not sure if I follow it.
Previously noexecstack was applied to oehostapp and thus all enclave hosts would inherit it.
With this PR, noexecstack is applied to oehost and thus oehostapp and in turn all enclave hosts would inherit it.

Can you add a description for what is the subtle difference?

jhand2: If you link against oehost (as any application is welcome to do, I found this out because the stack protector test links against oehost) you don't get the flag.

And more important than the security benefit, with the upstream kernel driver, many enclaves with an executable stack will simply fail to load due to the Intel PSW calling mprotect on every page that gets loaded.

Some more details in this thread https://lkml.org/lkml/2020/2/18/1477

anakrish: Thanks for the explanation and the fix.

The bigger issue seems to be that we expose oehost and oehostapp targets to the user while we advocate only the use of oehostapp. This definitely needs fixing. I would suggest creating a "good first issue" for this and schedule it for v0.9.

Keeping in mind that we consider the host to be malicious, the host could intentionally not honor the execstack bits  in the ELF. Are we safe in such a scenario?

jhand2: @CodeMonkeyLeet, can you confirm, are page protections measured as part of MRENCLAVE?

If so, then a malicious host is not a concern.

host/CMakeLists.txt
CodeMonkeyLeet: @anakrish let me refine the question: we currently expect everyone to link against oehostapp given the sample usage anyway, and so we ostensibly already recommend `-rdynamic` to be the default use case. We still make oehost available, presumably to provide an option without `-rdynamic` (and prior to this change, without `noexecstack`). 
- Do you think providing a library that doesn't pull in `-rdynamic` in CMake by default is valuable?
    - My understanding of `-rdynamic` is that it pulls in all symbols which is useful for backtracing and dynamic library load/link scenarios and is the preferable default.
- I assume this expands the size of the resulting binary; are there other potentially undesirable effects?
    - Assuming we only have the `-rdynamic` default, is there an easy way to override that in a user's build? (e.g. I couldn't find a `no-dynamic` option).

@jhand2 Yes, it would be a breaking change to remove one or the other library, but it's one of those things that's better to change sooner rather than later, and we would still want to mark deprecation in this release, then eliminate in the next one.

host/CMakeLists.txt
CodeMonkeyLeet: Yes, EADD measures the page protections for any page added to the enclave. 

The use of `noexecstack` is not a protection for the enclave threat model, but one for the general application model, which is still valuable, much like we enabled strong stack protector by default in the host lib link options.

host/CMakeLists.txt
jhand2: I opened #2595 to track. This issue aside, are we ok with this PR going in?

anakrish: @jhand2 Thanks for creating the issue #2595 . I have added the _first good issue_ label to it.
Though it is not readily obvious why SGX driver (which deals with creating enclaves) should be impacted by execstack bits of the host executable, it is good for this change to go in since we had intended noexecstack to be applied to hosts and we weren't ensuring it.

anakrish: @CodeMonkeyLeet I don't have answer to the significance of -rdynamic. Yes, it seems relevant only if a dynamically loaded (via dlopen) shared library were to depend on symbols in the main host executable. But I am not sure if it is applicable in some way specifically for enclave applications. Maybe the global enclave list needs to be visible to all modules even though it exists only in the main executable?
Not really sure. Hopefully, we can investigate it as part of #2595

jhand2:
Thanks folks!

bors r+

jhand2:
Docker infra failure

bors r+

jhand2:
Docker failed again. 3rd times the char?

bors r+

PR #2600, opened 2020-02-25
Simulation Mode in Windows

anakrish:
This is a long standing feature request.
Simulation mode reduces the barrier to start playing with the OE SDK.

We had held off on implementing the simulation mode earlier because
- Simulation mode requires ability to manipulate FS and GS registers.
  Though it is possible to change these registers, Windows thread-scheduler
  reverts them back in the middle of enclave executation.
- We didn't yet have thread-local storage and other features like stack protectors
  that depend on FS register
- The enter/exit functions were difficult to change.

With @nonpolarity's work to implement stack protectors 456f3a2, enclave's no longer
use the GS register and solely use FS register, conforming to Linux ABI.
Since Windows uses GS register to handle thread related data, there is no register conflict
between the enclave (FS) and host (GS).
In simulation mode, only FS register needs to be set before calling the enclave.
GS register can be left unchanged.

Windows thread-scheduler, unlike Linux, however restores the FS register to its original value
whenever the thread resumes execution. This can happen in the middle of an ecall and thus
and enclave could end up with the wrong value of FS register in the middle of an ecall.
FS as set by Windows points to memory not accessible to the application and this results in
access violations.
Simulation mode registers a signal handler (executed first) that determines whether an
exception occurent within a simulation mode enclave and whether FS was incorrect at the
time of exception. If so, the exception handler sets FS to the correct value and requests
the program to continue execution.

TODO:
- Add documention in another PR
- Update CI
- Update debuggers to handle simulation mode

Note:
- Debugging works out of the box under VS
- WinDbg has explict guards to not support simulation mode enclave.
  These guards need to be removed and any missing functionality added.
- When Windows restores the FS register, it often results in enclave code
  throwing an access violation. The debugger can be set up to ignore this
  exception the first time it is hit and the debugging experience is smooth
  afterwards.


anakrish:
bors try

host/sgx/enter.c
```diff
@@ -181,8 +181,7 @@ void oe_enter_sim(
     OE_ALIGNED(16)
     uint64_t fx_state[64];
 
-    // Backup host GS and FS registers.
-    void* host_gs = oe_get_gs_register_base();
```
CodeMonkeyLeet: Can you also clean up existing references to `oe_get_gs_register_base()` and possibly remove the method entirely if unused? It only seems to be called in a vestigial codepath in `GetThreadBinding()` when `USE_TLS_FOR_THREADING_BINDING`, which is always defined and could also probably be removed as a meaningless macro.

The comments in this file also still refer to storing thread data in the GS segment register, which is confusing, since it looks like it's all TLS-based now. 

On a related note, I'm not sure how `GetThreadBinding()` and `ThreadBinding` escaped the variable rename change to `oe_get_thread_binding()` and `oe_thread_binding_t` respectively, but we should probably track fixing that up as well if not addressed in this PR.

host/sgx/calls.c
```diff
@@ -545,6 +545,7 @@ static void* _assign_tcs(oe_enclave_t* enclave)
                     binding->flags |= _OE_THREAD_BUSY;
                     binding->thread = thread;
                     binding->count = 1;
+                    binding->enclave = enclave;
```
CodeMonkeyLeet: This works, but I don't think this should be a mutable field on binding; there's a tight coupling of TCS to the enclave associated with that TCS.  

It would be ideal if the TCS and its reverse mapping of the TCS to the host representation of the `oe_enclave_t` are defined together in the binding and immutable after enclave creation.

host/sgx/enclave.h
```diff
@@ -67,6 +67,10 @@ typedef struct _thread_binding
 
     /* Event signaling object for enclave threading implementation */
     EnclaveEvent event;
+
+    /* This field allows the simulation mode exception handler to read enclave
+     * properties of the current thread binding*/
```
dthaler: nit: add space after "binding"

host/sgx/windows/exception.c
```diff
@@ -40,4 +85,6 @@ _host_exception_handler(struct _EXCEPTION_POINTERS* exception_pointers)
 void oe_initialize_host_exception()
 {
     AddVectoredExceptionHandler(SET_AS_FIRST_HANDLER, _host_exception_handler);
+    AddVectoredExceptionHandler(
```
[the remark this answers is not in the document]
anakrish: Nice idea.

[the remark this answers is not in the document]
anakrish: Done.

host/sgx/calls.c
anakrish: Good point. Done.

host/sgx/enclave.h
anakrish: Done.

host/sgx/enter.c
```diff
@@ -191,9 +190,16 @@ void oe_enter_sim(
...
         // entry to the enclave.
-        oe_set_gs_register_base((void*)(enclave->addr + sgx_tcs->gsbase));
+        // In Linux, the new value of FS persists until it is explicitly
+        // restored below. Windows however restores FS to the original value
+        // unexpectedly (say when the thread is suspended/resumed).
+        // This leads to access violations since features like stack-protector
+        // and thread-local storage use the FS register; but its value has been
+        // restored by Windows. To let the enclave chug along in simulation
+        // mode, we add signal handler that resets the FS register to the
+        // desired value.
```
jhand2: signal handler implies POSIX signals. Can you change the comment to explicitly note that its a vectored exception handler? Also since it isn't done in this file, add `in host/sgx/windows/exception.c` to the end.

host/sgx/enter.c
anakrish: Good point. I have cleaned up the naming convention and comments,  and also removed `USE_TLS_FOR_THREAD_BINDING`. They don't add much to the bulk of the PR and are within its scope.

host/sgx/enter.c
anakrish: Nice suggestions. Done.

anakrish:
@CodeMonkeyLeet @jhand2 I have addressed all comments.

anakrish:
bors try

dthaler:
Can we get this merged soon?  FYI, this is blocking some of the work on the VS Extension.   Even if the debugger support isn't done yet, this is still useful.

jhand2:
thanks!

bors r+

PR #2601, opened 2020-02-25
[Docs] Update documentation for RHEL 8

ionutbalutoiu:
* Make sure that `install-ansible.sh` and `remove-ansible.sh` scripts work on `yum` based operating systems
* Adapt GettingStartedDocs/Contributors for RHEL 8:
  * Add RHEL 8 to the supported platforms
  * When running the Ansible playbooks `sudo` is not required since the playbooks use `become: yes` flag

### DIFF
--- a/docs/GettingStartedDocs/Contributors/SGX1FLCGettingStarted.md
+++ b/docs/GettingStartedDocs/Contributors/SGX1FLCGettingStarted.md
@@ -30,14 +30,16 @@ sudo scripts/ansible/install-ansible.sh
 If you are running in an Azure Confidential Compute (ACC) VM and would like to use the attestation features, you should also run the following command from the root of the source tree:
 
 ```bash
-sudo ansible-playbook scripts/ansible/oe-contributors-acc-setup.yml
+ansible-playbook scripts/ansible/oe-contributors-acc-setup.yml
 ```
 
 If you are not running in an ACC VM, you should instead run:
 
 ```bash
-sudo ansible-playbook scripts/ansible/oe-contributors-setup.yml
+ansible-playbook scripts/ansible/oe-contributors-setup.yml
 ```
+
+NOTE: The Ansible playbook commands from above will try and execute tasks with `sudo` rights. Make sure that the user running the playbooks has `sudo` rights, and if it uses a `sudo` password add the following extra parameter `--ask-become-pass`.
 
 ## Build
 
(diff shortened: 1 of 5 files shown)


ionutbalutoiu:
@BRMcLaren please take a look at this PR.

If these changes get merged, the documentation that we have for [GettingStartedDocs/Contributors/SGX1GettingStarted](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/SGX1GettingStarted.md) will work for RHEL 8 as well.

BRMcLaren:
bors try

docs/GettingStartedDocs/Contributors/SGX1FLCGettingStarted.md
```diff
@@ -30,13 +30,13 @@ sudo scripts/ansible/install-ansible.sh
 If you are running in an Azure Confidential Compute (ACC) VM and would like to use the attestation features, you should also run the following command from the root of the source tree:
 
 ```bash
-sudo ansible-playbook scripts/ansible/oe-contributors-acc-setup.yml
```
BRMcLaren: Are we sure sudo is not required? I was pretty sure it was.

ionutbalutoiu: Yes, I am sure.

The playbooks `oe-contributors-acc-setup.yml`,  `oe-contributors-setup.yml`, and `oe-contributors-setup-sgx1.yml` have the `become: yes` flag when executing their tasks:
* https://github.com/openenclave/openenclave/blob/574ba3b0f0ca9eddc3d45ed72a0df472bc1fd60b/scripts/ansible/oe-contributors-acc-setup.yml#L7
* https://github.com/openenclave/openenclave/blob/574ba3b0f0ca9eddc3d45ed72a0df472bc1fd60b/scripts/ansible/oe-contributors-setup.yml#L7
* https://github.com/openenclave/openenclave/blob/574ba3b0f0ca9eddc3d45ed72a0df472bc1fd60b/scripts/ansible/oe-contributors-setup-sgx1.yml#L7

Thus, the playbooks execute with `sudo` rights. However, if the user has a password for `sudo` (which doesn't happen on our testing VMs), the following extra parameter `--ask-become-pass` must be passed to the `ansible-playbook` command. I'll document this.

Also, I found out that if you execute the `ansible-playbook` command with `sudo` on RHEL 8 platform, by default, the `PATH` from the `root` user doesn't have the directory with the `ansible-playbook` binary, leading to `command not found` kind of error.

So, I think it's easier to have the documentation like this: execute `ansible-playbook` without `sudo`, and make sure you pass the `--ask-become-pass` flag if sudo has a password.

BRMcLaren: This makes a lot of sense, thanks for clarifying, I agree.

PR #2606, opened 2020-02-27
Fix gdb load symbols script to find .text section

### LINKED ISSUE #2056: oegdb: finding the text section.
The SGK-LKL had modified the oegdb plugin. Here is the note from Peter.

"The oegdb plug-in uses ‘.find(“.text”)’ to find the text section in the enclave image, which incorrectly matches “.init.text” in the SGX-LKL enclave image. We changed this to ‘== “.text”’. "

jhand2:
The gdb plugin looks for the .text section by just checking if the section contains .text. This means the plugin will treat any section which contains .text as the text section.

Change the script to look for a section which exactly matches with .text

Fixes #2056


anakrish:
I remember seeing this as a PR from @mikbras or someone else.

PR #2608, opened 2020-02-27
libmbedcrypto.a is missing symbols.

### LINKED ISSUE #2607: mbedcrypto.a library is missing mbedtls_hkdf()
When trying to link against the libmbedcrypto.a library that comes with OE sdk, the function mbedtls_hkdf() is missing.

jazzybluesea:
Updated cmake to include missing source files when compiling libmbedcrypto.a

Fix #2607

### DIFF
--- a/3rdparty/mbedtls/CMakeLists.txt
+++ b/3rdparty/mbedtls/CMakeLists.txt
@@ -63,6 +63,7 @@ add_library(mbedcrypto_static STATIC
   mbedtls/library/aes.c
   mbedtls/library/aesni.c
   mbedtls/library/arc4.c
+  mbedtls/library/aria.c
   mbedtls/library/asn1parse.c
   mbedtls/library/asn1write.c
   mbedtls/library/base64.c
@@ -70,6 +71,8 @@ add_library(mbedcrypto_static STATIC
   mbedtls/library/blowfish.c
   mbedtls/library/camellia.c
   mbedtls/library/ccm.c
+  mbedtls/library/chacha20.c
+  mbedtls/library/chachapoly.c
   mbedtls/library/cipher.c
   mbedtls/library/cipher_wrap.c
   mbedtls/library/cmac.c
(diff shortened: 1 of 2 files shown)


jhand2:
Can you add to the PR description a quick summary of what each of these functions do? I want to make sure there wasn't a specific reason they were excluded.

jazzybluesea:
The missing symbols are part of the original mbedtls library.  For more information you can find them here: https://tls.mbed.org/api/

The function I needed was mbedtls_hkdf() which is used for key derivation.

jhand2:
Basically my question is are these newly added functions in a recent release of mbedtls? If not then there may be a reason they weren't included

jazzybluesea:
I don't think so. Maybe @CodeMonkeyLeet can comment.

CodeMonkeyLeet:
> @jhand2: Basically my question is are these newly added functions in a recent release of mbedtls? If not then there may be a reason they weren't included
>
> @jazzybluesea: I don't think so. Maybe @CodeMonkeyLeet can comment.

No, this is unintentional. Prior to 8be9801f5f9f1c22e96238385aa3c6be6391eac6, we didn't need to worry about this because we just imported the mbedTLS libraries built from the upstream 3rdparty/mbedtls/mbedtls/library/CMakeLists.txt into the OE CMake project as externals. Now we actually compile the libraries using our own source list definition in 3rdparty/mbedtls/CMakeLists.txt, which means that we need to keep the sources list in sync with the upstream.

The disconnect observed in #2607 is due to the upgrade of mbedTLS from 2.7.11 to 2.16.2 (90b578f11553419731904dc3b61d8db41b45492f) introducing new upstream source files that were not then added to OE's 3rdparty/mbedtls/CMakeLists.txt.

**Moving forwards, we need to keep the two CMakeLists in sync.** We don't actually use all of the source files, but the idiomatic way to control this is via the 3rdparty/mbedtls/config.h options, not by manually omitting source files in our CMake.

3rdparty/mbedtls/CMakeLists.txt
```diff
@@ -70,6 +70,8 @@ add_library(mbedcrypto_static STATIC
   mbedtls/library/blowfish.c
```
CodeMonkeyLeet: Can you also add the missing mbedtls/library/aria.c file in its alphabetical position? I think that's the last missing file introduced in the 2.16.2 update.

jazzybluesea: Done.

PR #2609, opened 2020-02-27
OE SGX: Move switchless infrastructure to EDL

anakrish:
oecore:
- Switchless no longer uses core ecalls and ocalls.
  All the lower-level edge calls it uses are now expressed in EDL
- common/switchless.edl contains internal ecalls for switchless calls.
  It can be made user visible if needed.

oeedger8r
- Emit static keyword for generated local functions. This prevents compiler warnings
  about missing prototypes.
- Remove workarounds in wrapper files for internal EDLs.

OPTEE support
- Switchless calling is supported only in SGX. However the files were incorrectly put
  in core on the enclave side instead of in core/sgx. The files are now moved to correct
  locations.

### DIFF
--- a/common/sgx/switchless.edl
+++ b/common/sgx/switchless.edl
@@ -19,8 +19,8 @@ enclave
     trusted
     {
         public oe_result_t oe_init_context_switchless_ecall(
-	    [user_check] struct _host_worker_context* host_worker_contexts,
-	    uint64_t num_host_workers);
+            [user_check] struct _host_worker_context* host_worker_contexts,
+            uint64_t num_host_workers);
 
     };
 
@@ -28,6 +28,6 @@ enclave
     {
         // Wake up a host switchless ocall worker thread.
         void oe_wake_switchless_worker_ocall(
-	    [user_check] struct _host_worker_context* context);
+            [user_check] struct _host_worker_context* context);
     };
 };
--- a/common/sgx/switchless.edl
+++ b/common/sgx/switchless.edl
@@ -1,0 +1,33 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+/*
+**==============================================================================
+**
+** switchless.edl:
+**
+**     This file declares internal ECALLs/OCALLs used by switchless calling
+**     infrastructure.
+**
+**==============================================================================
+*/
+
+enclave
+{
+    include "openenclave/bits/types.h"
+
+    trusted
+    {
+        public oe_result_t oe_init_context_switchless_ecall(
+	    [user_check] struct _host_worker_context* host_worker_contexts,
+	    uint64_t num_host_workers);
+
+    };
+
+    untrusted
+    {
+        // Wake up a host switchless ocall worker thread.
+        void oe_wake_switchless_worker_ocall(
+	    [user_check] struct _host_worker_context* context);
+    };
+};
--- a/enclave/core/sgx/switchless_t_wrapper.c
+++ b/enclave/core/sgx/switchless_t_wrapper.c
@@ -38,6 +38,8 @@ static oe_result_t _call_host_function(
         false /* non-switchless */);
 }
 
+/* Include the oeedger8r generated C file. The macros defined above customize
+ * the generated code for internal use. */
 #include "switchless_t.c"
 
 /* Registers the switchless ECALL function table. */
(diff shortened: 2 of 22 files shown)


anakrish:
bors try

common/sgx/switchless.edl
```diff
@@ -0,0 +1,33 @@
...
+
+enclave
+{
+    include "openenclave/bits/types.h"
+
+    trusted
+    {
+        public oe_result_t oe_init_context_switchless_ecall(
+	    [user_check] struct _host_worker_context* host_worker_contexts,
+	    uint64_t num_host_workers);
```
jhand2: Can you tab these over so that they're indented further than the above line?

common/sgx/switchless.edl
```diff
@@ -0,0 +1,33 @@
...
+
+    };
+
+    untrusted
+    {
+        // Wake up a host switchless ocall worker thread.
+        void oe_wake_switchless_worker_ocall(
+	    [user_check] struct _host_worker_context* context);
```
jhand2: same as above

enclave/core/sgx/switchlesscalls.h
```diff
@@ -8,8 +8,6 @@
 
 bool oe_is_switchless_initialized();
 
-oe_result_t oe_handle_init_switchless(uint64_t arg_in);
```
jhand2: Isn't this an API breaking change?

host/tee_u_wrapper.c
```diff
@@ -31,14 +34,8 @@ static oe_result_t _call_tee_enclave_function(
         output_bytes_written);
 }
 
-/* Ignore missing edge-routine prototypes. */
-#if defined(__GNUC__)
-#pragma GCC diagnostic ignored "-Wmissing-prototypes"
-#endif
-
```
jhand2: I would expect removing this to cause a compiler error

common/sgx/switchless.edl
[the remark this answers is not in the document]
anakrish: Done.

common/sgx/switchless.edl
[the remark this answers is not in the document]
anakrish: Done.

enclave/core/sgx/switchlesscalls.h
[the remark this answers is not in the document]
anakrish: No. This is an internal function that is not visible to the user. Our public headers reside in include/openenclave/ ...

host/tee_u_wrapper.c
anakrish: The edger8r has been fixed to add the `static` keyword for local functions. edger8r already generates prototypes in header files for non local functions.

enclave/core/sgx/switchlesscalls.h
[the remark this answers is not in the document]
jhand2: ah you're right, my mistake

host/tee_u_wrapper.c
jhand2: ok. I would generally prefer this (and the modification to static) to be a separate change since they're pretty much completely separate from the core change.

jhand2: At least in a separate commit

enclave/core/sgx/switchless_t_wrapper.c
```diff
@@ -0,0 +1,51 @@
...
+        function_id,
+        input_buffer,
+        input_buffer_size,
+        output_buffer,
+        output_buffer_size,
+        output_bytes_written,
+        false /* non-switchless */);
+}
+
+#include "switchless_t.c"
```
dthaler: Including a .c file is very unusual. 
I'd recommend adding a /* comment */ explaining why you're doing it here.

host/syscall_u_wrapper.c
```diff
@@ -16,6 +16,9 @@
 /* Override oe_call_enclave_function() calls with _call_enclave_function(). */
 #define oe_call_enclave_function _call_enclave_function
 
+// Obscure the generated creation function by renaming it.
```
dthaler: nit: can we pick one comment style in this file and stick with it?  I.e., line 16, line 19, and line 22 currently alternate styles.

enclave/core/sgx/switchless_t_wrapper.c
anakrish: Done. Added a comment indicating that the file is being included so that the generated code is customized for internal use via the macros defined above.

host/syscall_u_wrapper.c
anakrish: Done.

PR #2611, opened 2020-02-28
 Extended Enclave Initialization Data

wintersteiger:
This adds support for Extended Enclave Initialization Data. 

Upon enclave creation (`oe_create_enclave_eeid`) the enclave image is extended with additional (instance-specific) data and re-signed with the debug key. The extended data also contains the original sigstruct of the re-signed enclave image, which is verified during extended quote verification (see `common/sgx/verify_eeid.c`). This allows us to specialize enclave image during enclave creation without the need for an entirely new image and a different signing key, while maintaining security/confidentiality guarantees.

The sample in `samples/remote_attestation_eeid` shows how to use this, although it should probably be merged with the non-eeid sample. 

CC @sylvanc @ad-l @letmaik @mikbras @jxyang


radhikaj:
@wintersteiger Thanks for the contribution. We would generally like to have a design discussion before submitting a PR. Also, we request that you submit an issue so we can discuss changing the public API surface:  Please see https://github.com/openenclave/openenclave/blob/master/docs/Contributing.md#design-discussion.  Also please see https://github.com/openenclave/openenclave/blob/master/docs/Contributing.md

include/openenclave/bits/eeid.h
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Open Enclave SDK contributors.
```
dthaler: The other files seem to be SGX-only but this file has a name and is in a directory that is not SGX-specific.
I can't tell if this is SGX only (in which case this file is in the wrong directory) or whether changes to other files are also needed to provide non-SGX support.

include/openenclave/host.h
```diff
@@ -156,6 +156,55 @@ oe_result_t oe_create_enclave(
...
+ *
+ * This function creates an enclave from an enclave image file. On successful
+ * return, the enclave is fully initialized and ready to use.
+ *
+ * @param path The path of an enclave image file in ELF-64 format. This
+ * file must have been linked with the **oecore** library and signed by the
+ * **oesign** tool.
+ *
+ * @param type The type of enclave supported by the enclave image file.
+ *     - OE_ENCLAVE_TYPE_SGX - An SGX enclave
```
dthaler: Why is only SGX listed here?

samples/remote_attestation_eeid/README.md
```diff
@@ -0,0 +1,233 @@
...
+- Written in C++
+- Demonstrates an implementation of remote attestation
+- Use of mbedTLS within the enclave
+- Use Asymmetric / Public-Key Encryption to establish secure communications between two attesting enclaves
+- Enclave APIs used:
+  - oe_get_report
+  - oe_verify_report,
+  - oe_is_within_enclave
+
+**Note: Currently this sample only works on SGX-FLC systems.** The underlying SGX library support for end-to-end remote attestation is available only on SGX-FLC system. There is no plan to back port those libraries to either SGX1 system or software emulator.
```
dthaler: Is this SGX-specific in some way or should the same sample apply to OP-TEE etc?

samples/remote_attestation_eeid/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+- This is useful to bootstrap a secure communication channel between the enclave and the challenger.
+
+  - In this sample, the enclave signs the hash of an ephemeral public key into its report, which the challenger can then use to encrypt a response to it.
+
+  - Other usage examples for `reportData` might be to include a nonce, or to initiate Diffie-Helman key exchange.
+
+##### 2) Verifying the integrity of the Enclave Report
+
+Once the report is generated and passed to the challenger, the challenger can call `oe_verify_report` to validate the report originated from an Trust Execution Environment (TEE, in the case it's a valid SGX platform).
```
dthaler: s/Trust/Trusted/

samples/remote_attestation_eeid/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+In the context of Open Enclave on Intel SGX platform, a remote report is verified using the certificate chain issued by Intel which is only valid for SGX platforms.
```
dthaler: We're trying to align terminology with the IETF, so "report" is not a term we should use in docs.   We use "evidence" for information sent to a verifier, and "attestation results" for information from a verifier to a relying party.

samples/remote_attestation_eeid/common/dispatcher.cpp
```diff
@@ -0,0 +1,205 @@
...
+    m_name = name;
+    m_crypto = new Crypto();
+    if (m_crypto == NULL)
+    {
+        goto exit;
+    }
+
+    // Extract modulus from raw PEM.
+    if (!m_crypto->get_rsa_modulus_from_pem(
+            // m_enclave_config->other_enclave_pubkey_pem,
```
dthaler: Why is there commented out code here?

samples/remote_attestation_eeid/host/host.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include "remoteattestation_u.h"
+
+oe_enclave_t* create_enclave(const char* enclave_path, oe_eeid_t* eeid)
+{
+    oe_enclave_t* enclave = NULL;
+
+    printf("Host: Enclave library %s\n", enclave_path);
+    oe_result_t result = oe_create_remoteattestation_enclave_eeid(
+        enclave_path,
+        OE_ENCLAVE_TYPE_SGX,
```
dthaler: why not OE_ENCLAVE_TYPE_AUTO?

include/openenclave/bits/eeid.h
```diff
@@ -0,0 +1,48 @@
...
+**==============================================================================
+**
+** oe_eeid_t
+**
+**==============================================================================
+*/
+typedef struct oe_eeid_t_
+{
+    uint32_t hash_state_H[8];
+    uint32_t hash_state_N[2];
```
dthaler: The above field names aren't self-explanatory.  Please add comments explaining what they are.

include/openenclave/enclave.h
```diff
@@ -386,6 +386,13 @@ oe_result_t oe_verify_report(
     size_t report_size,
     oe_report_t* parsed_report);
 
+struct oe_eeid_t_;
+oe_result_t oe_verify_report_eeid(
```
dthaler: Since this API name isn't aligned with the future direction of terminology (it uses "report" vs "evidence" or "attestation result"), it should be marked as @experimental so that we can make breaking changes at will.

include/openenclave/host.h
```diff
@@ -156,6 +156,55 @@ oe_result_t oe_create_enclave(
...
+ *
+ * @param eeid Optional user data for attestation.
+ *
+ * @param enclave This points to the enclave instance upon success.
+ *
+ * @returns Returns OE_OK on success.
+ *
+ */
+struct oe_eeid_t_;
+oe_result_t oe_create_enclave_eeid(
```
dthaler: Do we really need two different create_enclave APIs?

include/openenclave/host.h
```diff
@@ -156,6 +156,55 @@ oe_result_t oe_create_enclave(
...
+ * initialization data.
+ *
+ * This function creates an enclave from an enclave image file. On successful
+ * return, the enclave is fully initialized and ready to use.
+ *
+ * @param path The path of an enclave image file in ELF-64 format. This
+ * file must have been linked with the **oecore** library and signed by the
+ * **oesign** tool.
+ *
+ * @param type The type of enclave supported by the enclave image file.
```
dthaler: doxygen on public APIs should use [in,out] style annotations on all params

soccerGB:
Thanks for the effort in putting those changes together.

From a quick look at this PR, most of the changes are related to a customized quote format support. Have you guys thought about leveraging the Attestation Plugin model below for the quote operations proposed in the API changes?

https://github.com/openenclave/openenclave/blob/765e9d9262696137de6e8b3c8ed379aa0e9b6b42/docs/DesignDocs/CustomAttestation.md

include/openenclave/bits/eeid.h
wintersteiger: Conceptually this idea does not depend on SGX, but the implementation currently doesn't support anything else, mainly because I assume that the sigstruct is saved in the EEID data.

include/openenclave/host.h
wintersteiger: No; as an alternative we can add the additional argument to the existing create_enclave, or perhaps put it into the settings. I figured it's best to keep it separate for now.

samples/remote_attestation_eeid/README.md
wintersteiger: This is a copy of `samples/remote_attetsation/README.md`, I don't know whether this is still true.

kuanggang-git:

samples/remote_attestation_eeid/common/attestation.cpp
```diff
@@ -0,0 +1,167 @@
...
+ * 1) The remote report is first attested using the oe_verify_report API. This
+ * ensures the authenticity of the enclave that generated the remote report.
+ * 2) Next, to establish trust of the enclave that  generated the remote report,
+ * the mrsigner, product_id, isvsvn values are checked to  see if they are
+ * predefined trusted values.
+ * 3) Once the enclave's trust has been established, the validity of
+ * accompanying data is ensured by comparing its SHA256 digest against the
+ * report_data field.
+ */
+bool Attestation::attest_remote_report(
```
jxyang, line 77: We shouldn't provide EEID as an additional parameter. A better approach is to utilize the attestation plugin model, and turn EEID as one of the claims to be consumed by the verifier.

tests/eeid_plugin/plugin/eeid_verifier.c
```diff
@@ -0,0 +1,144 @@
+
```
jxyang, line 1: I believe this file should be in the same place with https://github.com/openenclave/openenclave/blob/aa1ebcb5a9d8db12a95932d20fa56f7159395d9a/common/sgx/verifier.c, but with the experimental tag. The users of EEID plugin shouldn't need to rewrite the functionalities here.

jxyang:
@gupta-ak This EEID plugin is similar to the SGX attestation plugin you implemented. Do you want to take this PR?

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,154 @@
+Extended Enclave Initialization Data
```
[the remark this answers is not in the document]
wintersteiger: Into master, yes (https://github.com/openenclave/openenclave/pull/2647). The final version will appear here the next time I rebase onto master.

common/sgx/sgxmeasure.c
```diff
@@ -3,7 +3,11 @@
 
 #include "sgxmeasure.h"
 #include <openenclave/bits/sgx/sgxtypes.h>
+#ifdef OE_BUILD_ENCLAVE
+#include <openenclave/enclave.h>
```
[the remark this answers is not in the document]
wintersteiger: Yes, I wanted to re-measure the everything in the enclave as a sanity check, so I needed `sgxmeasure.c` in both the enclave and the host build. I don't think this is strictly required though.

samples/remote_attestation_eeid/common/attestation.cpp
wintersteiger: Absolutely! I haven't ported this example yet, it still uses the older interface from before we had the plugin. Also, I'm not sure we want to keep this as a separate example.

tests/eeid_plugin/plugin/eeid_verifier.c
wintersteiger: That's right; I think the whole plugin should move out of the test directory. Would `enclave/sgx` be the right for the EEID attester, next to the SGX attester?

include/openenclave/bits/eeid.h
```diff
@@ -0,0 +1,68 @@
...
+    struct
+    {
+        uint32_t H[8];
+        uint32_t N[2];
+    } hash_state; /* internal state of the hash computation at the end of
+                           the enclave base image */
+    uint64_t signature_size; /* size of signature */
+    uint8_t signature[1808]; /* base-image signature and associated data (for
+                           SGX, the complete sigstruct of the base image) */
+    oe_enclave_size_settings_t
```
jxyang, line 37: In this structure, I'd like to see something like `ConfigID` to align with Icelake. It could be anything that uniquely identifies the enclave instance. In the case of SGX-LKL, it could be the root hash of the application. The verifier then just have to look at this data structure in most cases, and additionally verify the custom `data` section only when necessary. For that, we need the upper layer that constructed the `data` section to provide a customizable verifier.

samples/remote_attestation_eeid/common/attestation.cpp
jxyang: I think a sample would be nice to have. We just need to implement it with the plugin model, and have the attester and verifier in different host applications.

tests/eeid_plugin/plugin/eeid_verifier.c
jxyang: Maybe `openenclave/common` is a better place if we want to extend the concept of EEID to TrustZone. @dthaler ?

include/openenclave/bits/eeid.h
jxyang: If we move the application root hash into `data[]`, as we discussed offline, the data section should be application specific,  so the changes to SGX-LKL doesn't affect its verification. Make sense?

tests/eeid_plugin/plugin/eeid_verifier.c
dthaler: Only files that are truly SGX-specific should be under an sgx directory.   Files that are intended to be agnostic should not be (and should not have SGX-only code in such files).

include/openenclave/bits/eeid.h
wintersteiger: Sorry, doesn't make sense. What's the "application" here? The EEID data is instance-specific, i.e. specific to a particular enclave instance. If that instance happens to be an SGX-LKL enclave, then it is specific to SGX-LKL. In our current plans for SGX-LKL, the EEID data is a JSON file that contains the entire SGX-LKL configuration, including command line, environment variables, file system hashes, names of keys; essentially everything that ever will need to be attested. (But for OE, this data is opaque, so none of that needs to go into this particular header.)

jxyang: As we discussed offline, the `application-specific` bits should include the command line, the file system hash, and a few environment variables that have security impacts. The relying party has no interests in some of the SGX-LKL specific environment settings such as how many ethread we are running, stack size, TAP MTU, etc.

jxyang: Agree this is not OE specific. Move the discussion to another venue.

wintersteiger:
@jxyang and I decided to break this PR into smaller pieces to make it easier to review, so I'm closing this one. The first new one will follow shortly.

PR #2612, opened 2020-02-28
Append " -O0" compiling option in debug mode.

nonpolarity:
Append " -O0" to compiler options in debug mode.

For Clang " -O0" is added to CMAKE_C_FLAGS_DEBUG and CMAKE_CXX_FLAGS_DEBUG in debug mode.

For GNU compiler, according to https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html, the -g option already includes -O0 by default. -Og can disable optimization as -O0, but some symbol removing can cause severval tests fail. Here we just keep the default for debug mode.

### DIFF
--- a/cmake/compiler_settings.cmake
+++ b/cmake/compiler_settings.cmake
@@ -70,6 +70,16 @@ else ()
 endif ()
 
 if (CMAKE_CXX_COMPILER_ID MATCHES GNU OR CMAKE_CXX_COMPILER_ID MATCHES Clang)
+  # Enforce -O0 in debug mode.
+  string(TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_UPPER)
+  if (CMAKE_BUILD_TYPE_UPPER STREQUAL "DEBUG")
+    # For gcc the -g option alredy includes -O0.
+    # Here we just apply -O0 for Clang.
+    if (CMAKE_CXX_COMPILER_ID MATCHES Clang)
+      set (CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -O0")
+      set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0")
+    endif()
+  endif()
   # Enables all the warnings about constructions that some users consider questionable,
   # and that are easy to avoid. Treat at warnings-as-errors, which forces developers
   # to fix warnings as they arise, so they don't accumulate "to be fixed later".


cmake/compiler_settings.cmake
```diff
@@ -70,6 +70,13 @@ else ()
 endif ()
 
 if (CMAKE_CXX_COMPILER_ID MATCHES GNU OR CMAKE_CXX_COMPILER_ID MATCHES Clang)
+  # Enforce -O0 in debug mode.
+  string(TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_UPPER)
+  if (CMAKE_BUILD_TYPE_UPPER STREQUAL "DEBUG")
+    set (CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -O0")
```
jhand2: Should we also add /Od for Windows to disable optimization for Windows host code?

nonpolarity: Yes this option is already there. https://github.com/openenclave/openenclave/blob/d2050f5c2ac0caa7cf03eaf1b1adc3ddc9d6a5a3/cmake/compiler_settings.cmake#L100

anakrish:
@nonpolarity What is the reason for this change?
According to GCC docs, -O0 is the default https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

> -O0
> Reduce compilation time and make debugging produce the expected results. This is the default.

nonpolarity:
> 
> 
> @nonpolarity What is the reason for this change?
> According to GCC docs, -O0 is the default https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html
> 
> > -O0
> > Reduce compilation time and make debugging produce the expected results. This is the default.

@anakrish For clang there is still some optimizations while with "-g" option only. Any optimization will make debugger have some strange behavior such as optimizing out variable values or skipping over lines of source code. Actually for gcc I also prefer to add -O0 option that it also have this strange behavior occasionally.

anakrish:
> @anakrish For clang there is still some optimizations while with "-g" option only. Any optimization will make debugger have some strange behavior such as optimizing out variable values or skipping over lines of source code. Actually for gcc I also prefer to add -O0 option that it also have this strange behavior occasionally.

@nonpolarity Thanks for the explanation. It would be good to capture your explanation in comments since it not obvious otherwise why we are passing a default argument to the compiler.

Reading the above linked GCC doc,

> **-Og**
>    Optimize debugging experience. -Og should be the optimization level of choice for the standard edit-compile-debug cycle, offering a reasonable level of optimization while maintaining fast compilation and a good debugging experience. **It is a better choice than -O0 for producing debuggable code because some compiler passes that collect debug information are disabled at -O0**.
 > Like -O0, -Og completely disables a number of optimization passes so that individual options controlling them have no effect. Otherwise -Og enables all -O1 optimization flags except for those that may interfere with debugging:
>    -fbranch-count-reg  -fdelayed-branch 
    -fdse  -fif-conversion  -fif-conversion2  
    -finline-functions-called-once 
    -fmove-loop-invariants  -fssa-phiopt 
    -ftree-bit-ccp  -ftree-dse  -ftree-pta  -ftree-sra

Can you see if we should be passing **-Og** or **-O0** for best debugging experience?

nonpolarity:
> 
> 
> > @anakrish For clang there is still some optimizations while with "-g" option only. Any optimization will make debugger have some strange behavior such as optimizing out variable values or skipping over lines of source code. Actually for gcc I also prefer to add -O0 option that it also have this strange behavior occasionally.
> 
> @nonpolarity Thanks for the explanation. It would be good to capture your explanation in comments since it not obvious otherwise why we are passing a default argument to the compiler.
> 
> Reading the above linked GCC doc,
> 
> > **-Og**
> > Optimize debugging experience. -Og should be the optimization level of choice for the standard edit-compile-debug cycle, offering a reasonable level of optimization while maintaining fast compilation and a good debugging experience. **It is a better choice than -O0 for producing debuggable code because some compiler passes that collect debug information are disabled at -O0**.
> > Like -O0, -Og completely disables a number of optimization passes so that individual options controlling them have no effect. Otherwise -Og enables all -O1 optimization flags except for those that may interfere with debugging:
> > -fbranch-count-reg  -fdelayed-branch
> > -fdse  -fif-conversion  -fif-conversion2
> > -finline-functions-called-once
> > -fmove-loop-invariants  -fssa-phiopt
> > -ftree-bit-ccp  -ftree-dse  -ftree-pta  -ftree-sra
> 
> Can you see if we should be passing **-Og** or **-O0** for best debugging experience?

Thanks for your information. According to gcc's guide "Og" is a better choice. For clang "-Og" is still under development so right now "-O0" will be added.

nonpolarity:
bors try

cmake/compiler_settings.cmake
```diff
@@ -70,6 +70,18 @@ else ()
 endif ()
 
 if (CMAKE_CXX_COMPILER_ID MATCHES GNU OR CMAKE_CXX_COMPILER_ID MATCHES Clang)
+  # Enforce -O0 in debug mode.
+  string(TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_UPPER)
+  if (CMAKE_BUILD_TYPE_UPPER STREQUAL "DEBUG")
+    if (CMAKE_CXX_COMPILER_ID MATCHES GNU)
+      set (CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -Og")
```
andschwa: Isn't this already the default set by CMake? Perhaps `-g` is different from `-Og`? I'm looking at [GNU.cmake](https://github.com/Kitware/CMake/blob/07cfb18f9d29cfc0588ede928846a03ec5599c48/Modules/Compiler/GNU.cmake#L44).

nonpolarity:
bors try

cmake/compiler_settings.cmake
nonpolarity: Anand's reference https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html shows the detail. In short, -g turns on the debug info, and -O0 turns off optimization, which means 0 pass of optimization.

kuanggang-git:

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

PR #2618, opened 2020-03-02
Fix bug in add_enclave.cmake

mingweishih:
This PR fixes the bug in `add_enclave.cmake` (introduced by #2455), which does not properly add prefix to arguments obtained from `cmake_parse_arguments`.

### DIFF
--- a/cmake/add_enclave.cmake
+++ b/cmake/add_enclave.cmake
@@ -54,25 +54,15 @@ macro(add_enclave)
     "${multiValueArgs}"
     ${ARGN})
 
-  if (NOT SIGNING_ENGINE)
-     set( SIGNING_ENGINE "")
-  endif()
-  if (NOT ENGINE_LOAD_PATH)
-     set( ENGINE_LOAD_PATH "")
-  endif()
-  if (NOT ENGINE_KEY_ID)
-     set( ENGINE_KEY_ID "")
-  endif()
-
   if(OE_SGX)
     add_enclave_sgx(
       CXX ${ENCLAVE_CXX}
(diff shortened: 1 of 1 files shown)


cmake/add_enclave.cmake
```diff
@@ -54,25 +54,15 @@ macro(add_enclave)
...
       CXX ${ENCLAVE_CXX}
       TARGET ${ENCLAVE_TARGET}
       CONFIG ${ENCLAVE_CONFIG}
       KEY ${ENCLAVE_KEY}
-      SIGNING_ENGINE ${SIGNING_ENGINE}
-      ENGINE_LOAD_PATH ${ENGINE_LOAD_PATH}
-      ENGINE_KEY_ID ${ENGINE_KEY_ID}
+      SIGNING_ENGINE ${ENCLAVE_SIGNING_ENGINE}
+      ENGINE_LOAD_PATH ${ENCLAVE_ENGINE_LOAD_PATH}
+      ENGINE_KEY_ID ${ENCLAVE_ENGINE_KEY_ID}
```
jhand2: Are these to be passed in as environment variables? Where are they passed today?

mingweishih: By using `cmake_parse_arguments`, all the arguments will be added a prefix based on the first argument to the this function (in this case, `ENCLAVE`).

https://github.com/openenclave/openenclave/blob/c5ba7f3399335c035cc55e8b1c537c240879a450/cmake/add_enclave.cmake#L51

jhand2: gotcha, LGTM

mingweishih:
bors try

Issue #2543, opened 2020-02-11
Support semantic versioning by OE release 1.0


jhand2:
We should support (semantic versioning)[https://semver.org/] by the time we reach a v1.0 release of OE.

Reasoning from @AevaOnline
> Breaking public API changes damage the community's trust in a project, so I'd like us to follow a deprecation cycle that's consistent and clear -- and I like that this proposal makes it clear. I suggest that we adopt a strong stance on guaranteeing the stability of public APIs as early as is reasonably possible, but no sooner.
However, if we adhered to SemVer, then a breaking API change -- and that includes the removal of a public stable API -- must always be accompanied by a major version change in the API.

jhand2:
Also look into `Releasing.md` to figure out what is documented and what we are doing incorrectly as of now.

andschwa:
Just FYI:
- https://github.com/openenclave/openenclave/blob/master/CHANGELOG.md#changelog
- https://github.com/openenclave/openenclave/blob/master/docs/Releasing.md#version-bump

jhand2:
Ya it sounds like from triage we already document that we do this. But the deprecation model has maybe not been followed in the past.

I'll double check the documentation to make sure we aren't missing anything and then close this. Beyond that, its just a matter of following the rules we've outlined 😄

radhikaj:
Everything is pretty well documented now.

PR #2620, opened 2020-03-02
Update API guidelines with additional rules including SemVer

dthaler:
Addresses issue #2543
Also addresses remaining suggestions on PR #2163 as discussed by the
CGC.

### DIFF
--- a/docs/ApiGuidelines.md
+++ b/docs/ApiGuidelines.md
@@ -14,16 +14,25 @@ Public APIs are the supported APIs that apps can use.
 
 Public API requirements: 
 
-* All public APIs must have at least one test case.
+* All public APIs must have at least one test case. This means that, for any
+  pull request to be merged that adds a public API, it must also add at least
+  one test case for it.
 * All public APIs must be in a header file that is published to a location under the SDK output directory.
 * All public APIs must be fully documented using doxygen markup in the public header files
 * All public APIs must be used in some sample code or snippet. Ideally, the sample
   should be one that is compiled as part of the build, and uses no experimental APIs. For rarely 
   used public APIs, the sample might simply be a snippet in the doxygen comments.
-* Public APIs cannot have source-breaking changes across Open Enclave versions. For platforms 
-  where shared libraries are built, public APIs cannot have binary-breaking changes across 
-  Open Enclave versions either. (Source- and binary-breaking changes can still be made to a new 
-  API before it is released for the first time.  See the Breaking Changes section below for further discussion.)
+  This means that, for any pull request to be merged that adds a public API,
+  it must also add at least one sample or snippet showing how it is used.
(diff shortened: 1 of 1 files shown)


docs/ApiGuidelines.md
```diff
@@ -46,8 +55,10 @@ Experimental API requirements:
 * Experimental APIs must be in a header file that is published to the SDK output directory. They
   can be in their own header, or in a header file that is shared with public APIs but only
   if surrounded by an appropriate ifdef (e.g., "#ifdef OE\_CONTEXT\_SWITCHLESS\_EXPERIMENTAL\_FEATURE").
-* Experimental APIs should be fully documented using doxygen markup in the header file
-* Experimental APIs must have at least one test case.
+* Experimental APIs should be fully documented using doxygen markup in the header file, and include the @experimental markup.
```
CodeMonkeyLeet: Per the discussion in PR [#2579](https://github.com/openenclave/openenclave/pull/2579#discussion_r384115879), do we want to be more prescriptive in the use of the `@experimental` API tag? For example, do we want to recommend that all new APIs should be marked experimental APIs for at least a minor version release?

docs/ApiGuidelines.md
```diff
@@ -46,8 +55,10 @@ Experimental API requirements:
 * Experimental APIs must be in a header file that is published to the SDK output directory. They
   can be in their own header, or in a header file that is shared with public APIs but only
   if surrounded by an appropriate ifdef (e.g., "#ifdef OE\_CONTEXT\_SWITCHLESS\_EXPERIMENTAL\_FEATURE").
```
CodeMonkeyLeet: Nit: consider using code markdown for readability and not having to use character escapes (e.g. `#ifdef OE_CONTEXT_SWITCHLESS_EXPERIMENTAL_FEATURE`)

docs/ApiGuidelines.md
radhikaj: I think this is a good idea. Can we add it to the follow up PR that will address the outstanding comments here?

docs/ApiGuidelines.md
```diff
@@ -14,16 +14,25 @@ Public APIs are the supported APIs that apps can use.
 
 Public API requirements: 
 
-* All public APIs must have at least one test case.
```
radhikaj: I would like to add that any new public API should have a design PR that precedes it. Once it is approved, contributors may submit code to implement the new public API.

docs/ApiGuidelines.md
achamayou: I don't think that serves any purpose pre-1.0. There's no effective compatibility SLA, and staying away from `@experimental` APIs does nothing to change that. If OE decides to break compatibility, that's a minor release, and there's no break-free upgrade path from the previous release at all.

jhand2:
Merging this. Other discussions can continue in issues

bors r+

PR #2622, opened 2020-03-02
Remove unecessary thread local from oecore

jhand2:
The ocall hostbuffer code used an unnecessary thread local variable. Remove this usage to reduce oecore's dependance on threadlocal variables in general.

### DIFF
--- a/enclave/core/sgx/hostcalls.c
+++ b/enclave/core/sgx/hostcalls.c
@@ -46,10 +46,6 @@ void* oe_ecall_context_get_ocall_buffer(uint64_t size)
     return NULL;
 }
 
-// Thread local variable to keep track of whether ecall context's buffer
-// was used for making the current ocall.
-static __thread bool _thread_ecall_context_buffer_used;
-
 // Function used by oeedger8r for allocating ocall buffers.
 void* oe_allocate_ocall_buffer(size_t size)
 {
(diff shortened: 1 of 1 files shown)


enclave/core/sgx/hostcalls.c
```diff
@@ -58,21 +54,21 @@ void* oe_allocate_ocall_buffer(size_t size)
...
 
 // Function used by oeedger8r for freeing ocall buffers.
 void oe_free_ocall_buffer(void* buffer)
 {
+    oe_ecall_context_t* ecall_context = _get_ecall_context();
+
     // ecall context's buffer is managed by the host and does not have to be
     // freed.
-    if (_thread_ecall_context_buffer_used)
+    if (buffer == ecall_context->ocall_buffer)
```
anakrish: host can change the value of ecall_context->ocall_buffer. Can you make sure that we are still secure in that case and document the reasoning here?

jhand2: good question. I looked into it and here's what the host could do:

* Change `ocall_buffer` to null, cause a crash
* Change `ocall_buffer` from heap memory to the static buffer. This will cause the heap memory to not be free'd. That's fine as it's the host's memory anyway
* Change `ocall_buffer` from static buffer to heap memory. This will make us free some arbitrary heap memory. This could only make us free host memory (not enclave memory) so this is fine.
* Change `ocall_buffer` to enclave memory. Since we free it with oe_host_free, this will fault since host can't access enclave memory.

Overall I don't think this gives any extra privilege to the host.

anakrish: Great. It would be good to add a comment to the source file indicating that ecall_context->buffer is under hosts control and that we are robust to host manipulation due to the reasons you describe above.

The fourth case you mention is interesting. Currently oe_host_free just calls the host to free the memory. It would be better not to rely on the current behavior of oe_host_free. Say in the future it changes to do something to the memory (e.g clean it before returning to the host), in that case the host can use that fact to cause the enclave to overwrite its contents.

It would be better to detect ocall_buffer pointing to enclave memory and handle it here itself.

anakrish: I thought about it bit more. Actually there is a potential security vulnerability here as well as in _get_ecall_context and oe_ecall_context_get_ocall_buffer.

Any time there is a condition that involved host controlled data, we need to add an oe_lfence to prevent chance of speculative vulnerabilities.

Can you add oe_lfence appropriately to these 3 functions?

jhand2: I see how oe_lfence would benefit `oe_free_ocall_buffer`, but I don't understand the other two. Neither of those functions make decisions based on values controlled by the host.

td->host_ecall_context is a pointer to host memory but we only use the address, not the data it points to.

jhand2: @anakrish I added an lfence to `oe_free_ocall_buffer` but not the other two. Let me know your thoughts on the above comment.

anakrish: Both the comment and statement look good.
Are you planning to add lfences in the other two locations? If not, I can add them.

jhand2: See my first comment. I don't think the other two are necessary, unless I'm missing something.

jhand2: If we determine they are needed we can add then in a separate PR

anakrish: We do need lfence as a defense in depth strategy around boundary code - code that performs branches based on host controlled data. That is why oeedger8r generates lfence around boundary checks unconditionally irrespective of whether there is an exploitable gadget.

It is an oversight that the two other locations did not have lfence.

PR #2626, opened 2020-03-03
Get td_t structure based on SGX tcs rather than fs register

jhand2:
Currently in oe_enter, the thread data structure (td_t) is referenced as an offset of fs. Change this code to instead use an offset of the TCS structure; this offset is guaranteed by the loader.
    
This allows the td structure to be referenced throughout oe_enter even if %fs is modified during __oe_handle_main.

Additionally, Decouple td location from fs register in oe_asm_exit. Previously oe_asm_exit looked for the address of the td structure in the %fs register. Decouple this location by instead calling oe_get_td to get the address. This allows OE to more easily change where the td structure is kept in memory.

### DIFF
--- a/enclave/core/sgx/enter.S
+++ b/enclave/core/sgx/enter.S
@@ -39,20 +39,30 @@
 oe_enter:
 .cfi_startproc
 
+.get_td:
+
+    // Get the location of the td_t structure for this thread. This value is
+    // expected to be present in %r11 for the remainder of oe_enter.
+    //
+    // Upon first entry to the enclave, td->base.self in the td_t structure
+    // is not yet initialized. However, the loader in host/sgx/create.c places
+    // the td_t structure as a specific offset from TCS.
+    lea TD_FROM_TCS(%rbx), %r11
+
 .save_host_registers:
     // Backup the current host rbp, rsp, and context to previous.
-    mov %fs:td_host_rbp, %r8
-    mov %r8, %fs:td_host_previous_rbp
-    mov %fs:td_host_rsp, %r8
-    mov %r8, %fs:td_host_previous_rsp
-    mov %fs:td_host_ecall_context, %r8
-    mov %r8, %fs:td_host_previous_ecall_context
+    mov td_host_rbp(%r11), %r8
+    mov %r8, td_host_previous_rbp(%r11)
+    mov td_host_rsp(%r11), %r8
+    mov %r8, td_host_previous_rsp(%r11)
+    mov td_host_ecall_context(%r11), %r8
+    mov %r8, td_host_previous_ecall_context(%r11)
 
     // Save host registers (restored on EEXIT)
-    mov %rcx, %fs:td_host_rcx // host return address here
-    mov %rsp, %fs:td_host_rsp
-    mov %rbp, %fs:td_host_rbp
-    mov %rdx, %fs:td_host_ecall_context
+    mov %rcx, td_host_rcx(%r11) // host return address here
+    mov %rsp, td_host_rsp(%r11)
+    mov %rbp, td_host_rbp(%r11)
+    mov %rdx, td_host_ecall_context(%r11)
 
 .determine_entry_type:
     // Check if this is exception dispatching request.
@@ -66,7 +76,7 @@ oe_enter:
 
     // Check whether this is a clean entry or a nested entry
     // clean-entry-check.
-    mov %fs:td_depth, %r8
+    mov td_depth(%r11), %r8
     cmp $0, %r8
     je .clean_entry
     jmp .nested_entry
@@ -103,7 +113,7 @@ oe_enter:
     lfence
 
     // Restore stack pointer and enclave registers:
-    mov %fs:td_last_sp, %rsp
+    mov td_last_sp(%r11), %rsp
 
     // align the stack
     and $-16, %rsp
@@ -142,8 +152,8 @@ oe_enter:
     popfq
 
     // Get the host stack pointer.
-    mov %fs:td_host_rsp, %r8
-    mov %fs:td_host_rbp, %r9
+    mov td_host_rsp(%r11), %r8
+    mov td_host_rbp(%r11), %r9
 
     // Construct the frame and align the stack.
     pushq $0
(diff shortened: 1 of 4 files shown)


enclave/core/sgx/enter.S
```diff
@@ -209,12 +246,14 @@ oe_enter:
     lfence
 
     // Clear the td_t.last_sp field (force oe_enter to calculate stack pointer)
-    movq $0, %fs:td_last_sp
+    movq $0, td_last_sp(%r11)
 
 .clear_enclave_registers:
 
     // Clear these so information will not be leaked to host
+    push %r11
```
jhand2: Leave a comment about saving %r11 and that it will be cleared after it is needed

enclave/core/sgx/enter.S
```diff
@@ -39,20 +39,50 @@
...
+    push %rdx
+
+    // Get the location of the td_t structure for this thread. This value is
+    // expected to be present in %r11 for the remainder of oe_enter.
+    mov %rbx, %rdi
+    call td_from_tcs
+    mov %rax, %r11
+
+    // Restore register state
+    pop %rdx
```
anakrish: Keep in mind that upon entry, rsp still points to host memory. Therefore the pushes and pops actually operate on host memory which is unsafe.

enclave/core/sgx/enter.S
```diff
@@ -39,20 +39,50 @@
...
+    push %rbx
+    push %rcx
+    push %rdi
+    push %rsi
+    push %rdx
+
+    // Get the location of the td_t structure for this thread. This value is
+    // expected to be present in %r11 for the remainder of oe_enter.
+    mov %rbx, %rdi
+    call td_from_tcs
```
anakrish: td_from_tcs just adds a known offset to tcs to obtain td_t. Can we just do the addition here instead of calling a function?

anakrish:
>
>     * FS can be changed during enclave execution and oe_enter still references the same structure.
> 

This does not seem applicable. Isn't FS guaranteed to be enclave + tcs.fsbase upon enclave entry?

enclave/core/sgx/enter.S
jhand2: Ah good catch. I'll change it

jhand2:
> 
> 
> > ```
> > * FS can be changed during enclave execution and oe_enter still references the same structure.
> > ```
> 
> This does not seem applicable. Isn't FS guaranteed to be enclave + tcs.fsbase upon enclave entry?

Yes, but later in the function (after `__oe_handle_main()`) that same guarantee isn't true anymore

anakrish:
> 
> 
> > > ```
> > > * FS can be changed during enclave execution and oe_enter still references the same structure.
> > > ```
> > 
> > 
> > This does not seem applicable. Isn't FS guaranteed to be enclave + tcs.fsbase upon enclave entry?
> 
> Yes, but later in the function (after `__oe_handle_main()`) that same guarantee isn't true anymore

It is not clear how this changelist enables that. Irrespective of whether FS was changed by some code later in the enclave, the hardware provides  guarantees that in oe_enter, FS is enclave_base + tcs.fsbase.

jhand2:
> 
> 
> > > > ```
> > > > * FS can be changed during enclave execution and oe_enter still references the same structure.
> > > > ```
> > > 
> > > 
> > > This does not seem applicable. Isn't FS guaranteed to be enclave + tcs.fsbase upon enclave entry?
> > 
> > 
> > Yes, but later in the function (after `__oe_handle_main()`) that same guarantee isn't true anymore
> 
> It is not clear how this changelist enables that. Irrespective of whether FS was changed by some code later in the enclave, the hardware provides guarantees that in oe_enter, FS is enclave_base + tcs.fsbase.

We previously accessed fs during `oe_enter()` after `__oe_handle_main()`. After this PR we don't.

anakrish:
> 
> 
> > > > > ```
> > > > > * FS can be changed during enclave execution and oe_enter still references the same structure.
> > > > > ```
> > > > 
> > > > 
> > > > This does not seem applicable. Isn't FS guaranteed to be enclave + tcs.fsbase upon enclave entry?
> > > 
> > > 
> > > Yes, but later in the function (after `__oe_handle_main()`) that same guarantee isn't true anymore
> > 
> > 
> > It is not clear how this changelist enables that. Irrespective of whether FS was changed by some code later in the enclave, the hardware provides guarantees that in oe_enter, FS is enclave_base + tcs.fsbase.
> 
> We previously accessed fs during `oe_enter()` after `__oe_handle_main()`. After this PR we don't.

Even without this PR, the changes made to FS by any code after `__oe_handle_main` does not affect the value of FS in `oe_enter`. So it is not clear how this PR changes the enclave behavior in this context.

jhand2:
> Even without this PR, the changes made to FS by any code after `__oe_handle_main` does not affect the value of FS in `oe_enter`. So it is not clear how this PR changes the enclave behavior in this context.

__oe_handle main is called in oe_enter. There are parts of oe_enter after __oe_handle_main. We only leave oe_enter after __oe_handle_main returns.

anakrish:
> 
> 
> > Even without this PR, the changes made to FS by any code after `__oe_handle_main` does not affect the value of FS in `oe_enter`. So it is not clear how this PR changes the enclave behavior in this context.
> 
> __oe_handle main is called in oe_enter. There are parts of oe_enter after __oe_handle_main. We only leave oe_enter after __oe_handle_main returns.

Ah ok. It would be good to add a comment specifically around the call to `__oe_handle_main` stating that this is what we are specifically trying to do.

As Simon pointed out, we would need to address the oe_asm_exit (which is used for ocalls) as well to make this PR complete.

anakrish:
Upon return from `__oe_handle_main`, can you detect a changed FS and raise an `int 3`? That will help identifying code paths that change FS but not restore it.

jhand2:
> Upon return from `__oe_handle_main`, can you detect a changed FS and raise an `int 3`? That will help identifying code paths that change FS but not restore it.

Will that cause the debugger to always break if this condition happens?

jhand2:
@CodeMonkeyLeet and @anakrish I have updated the PR to abstract out %fs usage in oe_asm_exit and have changed the PR description and commit messages to be more clear.

I have not added the `int 3` instruction because I don't quite understand the implications yet. I'm reading up a little more.

anakrish:
The user level code is called via the sequence 

```
oe_enter
-> __oe_handle_main
    -> user-level code
```

Currently __oe_handle_main is compiled with stack protectors and therefore it is not robust to a change in FS. Both oe_enter and __oe_handle_main must provide the same guarantees.

anakrish:
@jhand2 Heads up, any changes to boundary functions get scrutinized heavily and justifiably take longer to pass review. Kindly plan your development around that.

jhand2:
> The user level code is called via the sequence
> 
> ```
> oe_enter
> -> __oe_handle_main
>     -> user-level code
> ```
> 
> Currently __oe_handle_main is compiled with stack protectors and therefore it is not robust to a change in FS. Both oe_enter and __oe_handle_main must provide the same guarantees.

An application which changes fs should be held responsible for retaining the stack protector if they use a version of OE compiled with stack protector enabled. I see no other way to guarantee robustness since changing fs is equivalent to changing the stack protector.

anakrish:
I spent quite a lot of time thinking about this PR; I like the code changes on their own (with potential changes), but I am hesitant about the contract with user-level code that we are implicitly signing up for with this PR.

Consider the snippet from the PR:
https://github.com/openenclave/openenclave/blob/83f6c5e8d428fe6ce2669cce03af9603209fef84/enclave/core/sgx/enter.S#L189-L204


```asm
   // Save reference to the td structure to enclave stack.
   mov OM_ENC_TD, %r11
   ...
   call __oe_handle_main
   ...
   // Restore td pointer
   mov OM_ENC_TD, %r11
```

We rely on the stack (RSP) to save and restore td (R11). **To be precise, we rely on the expectation that __oe_handle_main is free to change RSP, but must restore it back upon return.**

If user code is sophisticated enough to manipulate FS register, I would expect it to manipulate the  RSP register. Change to FS almost always goes hand in hand with change in stack RSP. FS points to the thread's local data, whereas RSP points to the thread's stack.

If user code is able to restore RSP before handing control back to OE SDK, it is fair to reason that it should be able to restore FS too.

So, by allowing user code to arbitrarily change FS and not restore it, while requiring user code to restore RSP, are we covering up for deficiencies in the user code that might make more sense fixed at the user code level?

anakrish:
@jhand2 Do you still need this PR?

jhand2:
> @jhand2 Do you still need this PR?

No, not for now at least. I'll solve the problem in a different way.

PR #2633, opened 2020-03-05
Skip the oesign-engine-runnable test in simulation mode

jhand2:
The oesign-engine-runnable test does not use the proper mechanism to allow it to be skipped in simulation mode this causes simulation mode test failures. Add a check to the flags to determine if the test should be skipped.


jhand2:
Failure due to flaky crypto test

bors r+

PR #2636, opened 2020-03-05
SGX: Switchless ECALLs implementation

anakrish:
This is for most part an exact mirror implementation of switchless OCALLs.

The host create switchless ecall worker threads. Each worker thread waits
on the host side for on an event that is triggered if the thread was waiting
at the time an ecall message was posted on the worker thread.
Upon waking up, the thread makes an ecall into the enclave to launch the
enclave portion of the worker thread. On the enclave side, the worker
looks at its context to see if there is an ecall scheduled on it. If so,
it processes the ecall. If not, it keeps spinning until a threshold is reached
at which time, it goes back into the host to wait for the event. This is because
there is no efficient way for an enclave to wait for an event; the OS threading
primitives are not available within the enclave and the wait must happen on the
host side.

- update changelog
- Make oe_handle_call_enclave_function non static to be reused by switchless ecalls
- Add switchless ecall worker ecall
- Update switchless sample to show switchless ecalls
- Update switchless tests to lock down switchless ecalls
- Remove oeedger8r negative test for switchless ecalls
- Remove oeedger8r error for transition_using_threads attribute on ecalls
- Added oe_enclave_worker_context_t type
- Refactored worker wait and wake functions on both Windows and Linux
- Update backtrace test with new internal function name
- Fix memory leak in switchless manager destruction (normal and error scenario)


CHANGELOG.md
```diff
@@ -16,7 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ---------------------
 
 ### Added
-- Support for Simulation Mode on Windows.
+- Support `transition_using_threads` EDL attribute for ecalls in oeedger8r.
+  OE SDK now supports both switchless OCALLs and ECALLs.
```
jhand2: keep "Support for Simulation Mode on Windows."

samples/switchless/README.md
```diff
@@ -34,9 +35,9 @@ the associated costs and savings. In general, the good candidates for switchless
 1) short, thus the transition takes relatively high percentage of the overall execution time of the call; and
 2) called frequently, so the savings in transition time add up.
 
-## How does Open Enclave support switchless OCALLs
+## How does Open Enclave support switchless CALLs
```
radhikaj: CALLS need not be capitalized.

samples/switchless/README.md
```diff
@@ -103,6 +119,8 @@ called switchlessly at run time. However, this a best-effort directive. Open Enc
 to fall back to a tradition OCALL if switchless call resources are unavailable, e.g., the enclave is not configured
```
radhikaj: Not related to your changes, but can you change this to a "traditional OCALL"

CHANGELOG.md
jhand2: Looks like this still needs to be fixed

samples/switchless/enclave/switchless.conf
```diff
@@ -4,7 +4,7 @@
 # Enclave settings:
 Debug=1
 NumHeapPages=1024
-NumStackPages=1024
-NumTCS=1
+NumStackPages=64
```
jhand2: Just out of curiosity, is it necessary for an enclave to specify both OE_SET_ENCLAVE_SGX and provide a conf file at signing time? Is just one or the other allowed?

CHANGELOG.md
anakrish: Good catch. Readded. The simulation mode line item had somehow appeared under v0.8.1 tag. I have moved it to the right place under Unreleased.

samples/switchless/README.md
anakrish: Fixed.

samples/switchless/README.md
anakrish: Fixed.

samples/switchless/enclave/switchless.conf
anakrish: Enclaves ideally should only provide a conf.The primary use of OE_SET_ENCLAVE_SGX is for tests where we have alternated between not signing and signing test enclaves.

If both are specified, the specifications in conf wins.

CHANGELOG.md
```diff
@@ -33,9 +39,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [v0.8.1][v0.8.1_log] - 2020-02-07
 ---------------------
 
-### Added
```
dthaler: Why is this part removed in this PR?

common/sgx/switchless.edl
```diff
@@ -22,6 +22,9 @@ enclave
             [user_check] struct _host_worker_context* host_worker_contexts,
             uint64_t num_host_workers);
 
+        public void oe_switchless_enclave_worker_thread_ecall(
```
dthaler: Why is there an API without _sgx_ in the name, in an SGX-specific file?
This at least requires a // comment.
(E.g., if this API is meant to be TEE-agnostic and should be moved out of this file in the future, then say so)

samples/switchless/README.md
```diff
@@ -1,13 +1,14 @@
 # Switchless Calls Sample
 
-This sample demonstrates how to make switchless calls to host from inside an enclave.
+This sample demonstrates how to make switchless ocalls to host from inside an enclave, and switchless ecalls to enclave
```
dthaler: grammar: "to host" -> "to a host application"
"to enclave" -> "to an enclave"

samples/switchless/README.md
```diff
@@ -34,9 +35,9 @@ the associated costs and savings. In general, the good candidates for switchless
 1) short, thus the transition takes relatively high percentage of the overall execution time of the call; and
 2) called frequently, so the savings in transition time add up.
 
-## How does Open Enclave support switchless OCALLs
+## How does Open Enclave support switchless calls
 
-Open Enclave only supports synchronous switchless OCALLs currently. When the caller within an enclave makes a
+Open Enclave supports both synchronous switchless OCALLs and ECALLs currently. When the caller within an enclave makes a
```
dthaler: grammar: remove "both".

Also I don't think the sentence is true as worded, since I believe switchless calls are currently only supported for SGX.   Furthermore, I think the design inherently only supports TEEs that permit shared memory.

samples/switchless/README.md
```diff
@@ -45,33 +46,38 @@ A host worker thread checks and retrieves `job` from the shared memory region. I
 runtime to process the `job` by unmarshaling the parameters, then dispatching to the callee function, and finally
 relaying the result back to the trusted Open Enclave runtime, which is further forwarded back to the caller.
 
+Switchless ECALLs work in a similar manner. The `job` is created by the untrusted Open Enclave runtime (host-side) and
+posted to a shared memory region. An enclave workter thread checks and retrieves the `job` from the shared region
```
dthaler: typo "workter"

samples/switchless/README.md
```diff
@@ -45,33 +46,38 @@ A host worker thread checks and retrieves `job` from the shared memory region. I
...
+and processes the `job` by unmarshaling the parameters, then dispatching to the callee function, and finally
+relaying the result back to the untrusted Open Enclave runtime, which is further forwarcded back to the caller.
```
dthaler: typo "forwarcded"

samples/switchless/README.md
```diff
@@ -45,33 +46,38 @@ A host worker thread checks and retrieves `job` from the shared memory region. I
...
+
 If an enclave supports multiple simultaneous ECALLs, multiple simultaneous switchless OCALLs could be made from the
 enclave. We use multi-threaded host workers in that scenario. Open Enclave
-allows users to configure how many host worker threads are to be created for servicing switchless OCALLs. The
-following example illustrates how to do that. A word of caution is that too many host worker threads might increase
+allows users to configure how many worker threads are to be created for servicing switchless OCALLs and ECALLs. The
+following example illustrates how to do that. A word of caution is that too many worker threads might increase
 competition of cores between threads and degrade the performance. Therefore, if a enclave has switchless calls
```
dthaler: grammar: "a enclave" -> "an enclave"

samples/switchless/README.md
```diff
@@ -45,33 +46,38 @@ A host worker thread checks and retrieves `job` from the shared memory region. I
...
-enabled, Open Enclave caps the number of host worker threads for it to the number of enclave threads specified.
+enabled, Open Enclave caps the number of worker threads for it to the number of enclave threads specified.
 
-With the current implementation, we recommend that users avoid using more host worker threads than the minimum of:
+With the current implementation, we recommend that users avoid using more worker threads than the minimum of:
 
-1. the number of simultaneously active enclave threads, and
-2. the number of cores that are potentially available to host worker threads.
+1. the number of simultaneously active caller threads, and
+2. the number of cores that are potentially available to worker threads.
```
dthaler: Is there a platform-agnostic API that an app can use to get the number of cores on the machine?

samples/switchless/README.md
```diff
@@ -45,33 +46,38 @@ A host worker thread checks and retrieves `job` from the shared memory region. I
...
 
-For example, on a 4-core machine, if the number of the simultaneously active enclave threads is 2, and there are no
-host threads other than the two threads making ECALLs and the switchless worker threads, both 1) and 2) would be 2.
-So we recommend setting the number of host worker threads to 2.
+For example, on a 4-core machine, if the number of the simultaneously active caller threads is 2, and there are no
+threads other than the two threads making switchless calls, and the switchless worker threads, both 1) and 2) would be 2.
+So we recommend setting the number host worker threads to 2.
```
dthaler: grammar: revert removal of "of"

samples/switchless/README.md
```diff
@@ -118,10 +136,10 @@ The host first defines a structure specifically for configuring switchless calls
 first field `1` as the number of host worker threads for switchless OCALLs. In this example, 1) There is at most
 1 enclave thread all the time, and 2) The number of cores available to the host worker threads is unknown, and
 so we use 1 as explained above. The 2nd field specifies the number of enclave threads for switchless ECALLs.
```
dthaler: Where was it explained above?  70-73 only talk about using 1 when the number is 0 or negative, not when it's unknown.  I suspect you need to fix lines 70-73.

samples/switchless/host/host.c
```diff
@@ -82,8 +82,8 @@ int main(int argc, const char* argv[])
         flags |= OE_ENCLAVE_FLAG_SIMULATE;
     }
 
-    // Enable switchless and configure host worker number
-    oe_enclave_setting_context_switchless_t switchless_setting = {1, 0};
+    // Enable switchless and configure 1 host and 1 enclave worker
```
dthaler: Expand comment... how were the numbers 1 and 1 chosen?

CHANGELOG.md
anakrish: Fixed. Note, to be consistent with changelog for earlier releases, I have moved the `Added` section before `Changed` section.

samples/switchless/README.md
anakrish: I am not aware of such an API. The closest is C++11 [std::thread::hardware_concurrency](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency). However, it is only a hint.

common/sgx/switchless.edl
anakrish: Good point. Added `oe_sgx_` prefix to all the functions in that EDL.

samples/switchless/README.md
anakrish: Fixed.

samples/switchless/README.md
anakrish: I have reworded the sentence. Let me know if it sounds OK.

samples/switchless/README.md
anakrish: Fixed.

samples/switchless/README.md
anakrish: Fixed.

samples/switchless/README.md
anakrish: I have completely rewritten the paragraph about configuring the enclave, providing the rationale/assumptions based on which the number of workers are chosen.

samples/switchless/host/host.c
anakrish: I have provided the rationale/assumptions based on which the sample was written.
Hopefully it is more understandable for the user now.

anakrish:
@dthaler I have addressed all you feedback. I have tried to reword the comments/readme to be more readable; provide more explanation and rationale. Can you do another review?

samples/switchless/README.md
```diff
@@ -114,14 +132,21 @@ oeedger8r ../switchless.edl --untrusted
...
-first field `1` as the number of host worker threads for switchless OCALLs. In this example, 1) There is at most
-1 enclave thread all the time, and 2) The number of cores available to the host worker threads is unknown, and
-so we use 1 as explained above. The 2nd field specifies the number of enclave threads for switchless ECALLs.
-Since switchless ECALL is not yet implemented, we require the 2nd field to be `0`.
+The host first defines a structure specifically for configuring switchless calls.
+
+We conservatively assume that we are executing on a machine with 2 cores. The main thread will keep posting
+switchless ocalls from within the enclave. To service these ocalls, we will create one host worker thread by setting the first field (`max_host_workers`)
+of the structure to `1`. Since there is only one thread making synchronous switchless ocalls,
+there is no benefit to creating more that one worker thread. By restricting the number of caller and worker threads both to 1 on a two core machine, we ensure that the caller and worker threads can potentially run uninterrupted on
```
dthaler: "more that one" -> "more than one"

samples/switchless/README.md
```diff
@@ -114,14 +132,21 @@ oeedger8r ../switchless.edl --untrusted
...
+each of the available cores, causing less contention, and leading to maximum speedup via switchless calling.
+
+Once all the switchless ocalls have been serviced, the main thread will return to the host and then keep posting switchless ecalls.
+Inorder to service these switchless ecalls, we will create one enclave worker thread by setting the second field (`max_enclave_workers`) of the structure to `1`. At this time the host worker thread that serviced the switchless ocalls will automatically go to sleep since there are no more ocalls for it to process. Just as in the case of ocalls, for ecalls, one enclave worker thread and one host caller thread would lead to maximum speedup via switchless calling in a two core machine. For a machine with more cores, the number of worker threads can be increased in accordance with the recommendations given earlier.
```
dthaler: Suggest splitting into multiple source lines to make diffs/reviewing easier

samples/switchless/README.md
```diff
@@ -114,14 +132,21 @@ oeedger8r ../switchless.edl --untrusted
...
 ## About the host
 
-The host first defines a structure specifically for configuring switchless calls. In this case, we specify the
-first field `1` as the number of host worker threads for switchless OCALLs. In this example, 1) There is at most
-1 enclave thread all the time, and 2) The number of cores available to the host worker threads is unknown, and
-so we use 1 as explained above. The 2nd field specifies the number of enclave threads for switchless ECALLs.
-Since switchless ECALL is not yet implemented, we require the 2nd field to be `0`.
+The host first defines a structure specifically for configuring switchless calls.
+
+We conservatively assume that we are executing on a machine with 2 cores. The main thread will keep posting
```
dthaler: I think the first sentence should say something like:
"We conservatively assume that we are executing on a machine with 1 available core."  (i.e., 1 not 2, to be conservative)

samples/switchless/host/host.c
```diff
@@ -82,8 +82,40 @@ int main(int argc, const char* argv[])
...
+    // and measure the time taken. Using the two measurements, we can arrive at
+    // the speedup factor that switchless ocalls provide over normal ocalls.
+    //
+    // Similarly, to demonstrate the benefit of using switchless ecalls, we will
+    // first make 1000000 switchless ecalls from the main thread and measure the
+    // time taken. Then we will make the same number of regular ecalls and
+    // measure the time taken. Using the two measurements, we can arrive at the
+    // speedup factor that switchless ecalls provide over normall ecalls.
+    //
+    // We will conservatively assume that we are running on a two core machine.
```
dthaler: I think the first sentence should say something like:
"We conservatively assume that we are executing on a machine with 1 available core."  (i.e., 1 not 2, to be conservative)

host/sgx/switchless.c
```diff
@@ -82,52 +121,80 @@ static oe_result_t oe_stop_worker_threads(oe_switchless_call_manager_t* manager)
...
     oe_switchless_call_manager_t* manager = NULL;
-    oe_host_worker_context_t* contexts = NULL;
-    oe_thread_t* threads = NULL;
+    oe_host_worker_context_t* host_contexts = NULL;
+    oe_thread_t* host_threads = NULL;
+    oe_enclave_worker_context_t* enclave_contexts = NULL;
+    oe_thread_t* enclave_threads = NULL;
 
-    if (num_host_workers < 1 || enclave == NULL)
+    if (num_host_workers < 0 || num_enclave_workers < 0 || enclave == NULL)
```
fchinchilla: These are size_t and can never be < 0, correct?

host/sgx/windows/switchless.c
```diff
@@ -13,18 +13,37 @@ void oe_host_worker_wait(oe_host_worker_context_t* context)
     int32_t oldval = 1;
     int32_t newval = 0;
 
-    if (_InterlockedCompareExchange((long*)&context->event, newval, oldval) ==
-        0)
+    if (_InterlockedCompareExchange(event, newval, oldval) == 0)
```
fchinchilla: Going through the call sequence, the "event" passed in to this function is allocated via calloc() - 
host_contexts = calloc(...) and enclave_contexts = calloc(...). Is that correct?

The reason I ask is because the documentation for the _Interlocked* APIs says the memory needs to be aligned (32bit aligned on x86 and 64bit aligned on x64).

host/sgx/switchless.c
anakrish: Good catch. I am surprised because I would expect the compiler to catch such checks.

We are compiling with `-Wextra` which detects and emits a `comparison of unsigned expression < 0 is always false` warning which we would have treated as error.

However, since December we are also compiling with `-Wno-type-limits` option which suppresses the above warning. I have created an issue #2766 to remove `-Wno-type-limits` option.

I will go ahead and remove these checks.

host/sgx/windows/switchless.c
anakrish: @fchinchilla

Thanks for reviewing.

Yes, we allocate host_contexts and enclave_contexts via calloc.

Due to malloc's [guarantee](https://en.cppreference.com/w/c/memory/malloc) to return an address suitably aligned for any fundamental type, implying at least 64 bit alignment (alignment of void*) if not 128-bit, and due to `int32_t event` [guaranteed](https://github.com/anakrish/openenclave/blob/f8973684d0799cdc34c39ba29cf3163213ecad7b/include/openenclave/internal/switchless.h#L35) to be at a 32-bit aligned-offset (20 bytes) within its calloc-ed enclosing structure, we can be sure that `event` is always 32-bit aligned.

Does x64 require that 32-bit values are also aligned to 64 bits for interlocked operations to work?

samples/switchless/README.md
anakrish: Fixed.

samples/switchless/README.md
anakrish: Good point. Done.

samples/switchless/README.md
anakrish: I agree. `Conservatively` does not sound right here. Will try to reword the paragraph minimally.

samples/switchless/host/host.c
anakrish: I have rephrased it to..

"In this sample, the main thread will first keep posting switchless ocalls from within the enclave. To service these ocalls, .... One a machine with atleast two cores, both the caller and worker threads can potentially run uninterrupted on dedicated cores, causing less contention and leading to maximum speedup via switchless calling. On a machine with more than two cores, ..."

samples/switchless/README.md
anakrish: I have rephrased it to..

"In this sample, the main thread will first keep posting switchless ocalls from within the enclave. To service these ocalls, .... One a machine with atleast two cores, both the caller and worker threads can potentially run uninterrupted on dedicated cores, causing less contention and leading to maximum speedup via switchless calling. On a machine with more than two cores, ..."

anakrish:
bors try

host/sgx/windows/switchless.c
fchinchilla: Looks like this is OK then. The _InterlockedCompareExchange64 needs 64bit aligned, but _InterlockedCompareExchange only needs 32bit aligned. The other usages of Interlocked APIs appear to be the generic and 32bit alignment should be ok.

anakrish: Thanks for confirming. I was worried for bit.

anakrish:
I believe I have addresses all the PR feedback. Everyone, thanks for reviewing.

anakrish:
Fixed windows compile error.
bors r+

anakrish:
Last run seems to be a sporadic failure with attested_tls sample on Windows.

PR #2637, opened 2020-03-06
POC: Ability to recover FS and GS

anakrish:
Upon entry to an enclave, FS and GS registers are always set based on the
fsbase and gsbase values in the tcs. This guarantee provided by the hardware
can be leveraged to recover the FS and GS values if some advanced user code
has changed FS and GS registers.

Currently FS points to td_t and td_t is obtained via oe_get_td function.
oe_get_td can determine if FS has been changed by checking for the validity
of the memory location at FS:0. FS:0 is a valid td_t if FS[0] points to itself
and if it has td_magic.

If FS[0] does not point to td_t, then td_t (FS) can be recovered via the following
strategy:

1) Execute the following code pattern that raises an exception
      jmp raise_exception
      .string "special-magic-string"
      raise_exception: ud2
      mov %rax, td_t_local_variable

2) This will trigger the oe exception handling framework and we will reenter
  the enclave and reach the first-pass-exception handler. Similar to what happens
  for any hardware exception.
  The exception handler has the original FS (td_t) and GS values.
  Exception handler can inspect the bytes preceeding the faulting instruction to
  check if it is "special-magic-string" and if so, emulate the instruction by
  loading the SSA's RAX with original FS value and then moving SSA's RIP past
  the ud2 instruction. This is similar to how cpuid is emulated.

3) Upon continuing execution via ERESUME, control will resume at the next instruction
  past ud2 and RAX will be loaded with correct td_t* that oe_get_td can use.

### DIFF
--- a/enclave/core/sgx/td.c
+++ b/enclave/core/sgx/td.c
@@ -150,6 +150,39 @@ void* td_to_tcs(const td_t* td)
 /*
 **==============================================================================
 **
+** oe_debug_sanitize_fs_gs()
+**
+**
+**==============================================================================
+*/
+
+void oe_fetch_core_fs_gs(uint64_t* fs, uint64_t* gs)
+{
+    // First check if current FS and GS are valid.
+    // They are valid if they point to a valid td_t.
+    td_t *fs_value, *gs_value;
+    asm("mov %%fs:0, %0" : "=r"(fs_value));
+    asm("mov %%gs:0, %0" : "=r"(gs_value));
+
+    if (!td_initialized(fs_value) || !td_initialized(gs_value))
+    {
+        // Raise a vectored exception.
+        // oe_core will register an int 3 handler that checks
+        // if the ud2 is preceeded by the special sequence of bytes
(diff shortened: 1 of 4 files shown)


enclave/core/sgx/td.c
```diff
@@ -147,6 +147,39 @@ void* td_to_tcs(const td_t* td)
...
+    if (!td_initialized(fs_value) || !td_initialized(gs_value))
+    {
+        // Raise a vectored exception.
+        // oe_core will register an int 3 handler that checks
+        // if the ud2 is preceeded by the special sequence of bytes
+        // to determine if it was a request for fetching GS and FS
+        asm volatile("jmp 1f \n\t"
+                     ".string \"oe_fetch_core_fs_gs_magic\" \n\t"
+                     "1: ud2 \n\t"
+                     : "=a"(fs_value), "=b"(fs_value));
```
CodeMonkeyLeet: I assume you mean `"=b"(gs_value)`

PR #2644, opened 2020-03-09
Remove oesign's dependency on libsgx_enclave_common and libsgx_dcap_ql

### LINKED ISSUE #2264: oesign is dynamically linked to sgx_enclave_common and other libs
oesign shouldn't need any of the symbols in sgx_enclave_common or other Intel libraries, so it shouldn't be dynamically linked against them

oesign being dynamically linked to some sgx_* libraries causes an issue on Windows, because the Intel SGX PSW cannot be installed on systems without SGX, and those libraries are provided by the Intel SGX PSW, but oesign.exe requires them to run. It is therefore impossible to use oesign.exe on a Windows system without SGX support without manually copying the sgx_* libraries from another system onto it.  (And because oesign.exe is used during the normal build of the Open Enclave SDK, one cannot build the Open Enclave SDK from source in this situation too)

### LINKED ISSUE #2544: oesign fails on non-FLC machine
The OE SDK can be used to build for either FLC or legacy SGX machines.  However, even if you built for legacy SGX, and want to sign for legacy SGX, the oesign tool will fail unless it's run on a machine with sgx_dcap_ql.dll installed, which is only required for FLC machines.

Expected behavior is that oesign works either way and only requires FLC support when signing an FLC enclave.   That is, a dev/build machine that is just doing legacy SGX enclaves should not need sgx_dcap_ql.dll installed.

In the future, if oesign is able to sign OP-TEE enclaves as well, it should also be runnable without installing any SGX prerequisites, unless you are explicitly trying to sign an SGX enclave.

nonpolarity:
A static library liboemr.a is introduced, which has no dependency on these shared libraries.
All the dependent sentences are stripped away through a macro "OEMR" during preprocessing.

Fixes #2544 
Fixes #2264

### DIFF
--- a/host/CMakeLists.txt
+++ b/host/CMakeLists.txt
@@ -36,7 +36,7 @@ if (OE_SGX)
 
     add_custom_target(switchless_untrusted_edl
         DEPENDS switchless_u.h switchless_u.c switchless_args.h)
-endif()
+endif ()
 
 ##==============================================================================
 ##
@@ -45,7 +45,7 @@ endif()
 ##
 ##==============================================================================
 
-if(OE_SGX)
+if (OE_SGX)
     set(SGX_EDL_FILE ${EDL_DIR}/sgx/sgx.edl)
 
     add_custom_command(
@@ -55,7 +55,7 @@ if(OE_SGX)
 
     add_custom_target(sgx_untrusted_edl
         DEPENDS sgx_u.h sgx_u.c sgx_args.h)
-endif()
+endif ()
 
 ##==============================================================================
 ##
@@ -131,9 +131,9 @@ elseif (WIN32)
     windows/hostthread.c
     windows/syscall.c
     windows/time.c)
-else()
-  message(FATAL_ERROR "Unknown OS. Only supported OSes are Linux and Windows")
-endif()
+else ()
+  message(FATAL_ERROR "Unknown OS. The only supported OSes are Linux and Windows")
+endif ()
 
 if (OE_SGX AND WIN32)
   # Use clang to compile the oe_enter function that calls ENCLU.
@@ -149,7 +149,7 @@ if (OE_SGX AND WIN32)
              -I${PROJECT_SOURCE_DIR}/include
              ${CMAKE_CURRENT_SOURCE_DIR}/sgx/enter.c
              -o enter.obj)
-endif()
+endif ()
 
 # SGX specific files.
 if (OE_SGX)
@@ -202,7 +202,7 @@ if (OE_SGX)
       sgx/linux/sgxioctl.c
       sgx/linux/switchless.c
       sgx/linux/xstate.c)
-  else()
+  else ()
     list(APPEND PLATFORM_HOST_ONLY_SRC
       sgx/windows/sgxquoteproviderloader.c)
 
@@ -212,22 +212,22 @@ if (OE_SGX)
       sgx/windows/exception.c
       sgx/windows/switchless.c
       sgx/windows/xstate.c)
-  endif()
+  endif ()
 
   set(PLATFORM_FLAGS "-m64")
-elseif(OE_TRUSTZONE)
+elseif (OE_TRUSTZONE)
   list(APPEND PLATFORM_SDK_ONLY_SRC
     optee/log.c)
 
   if (UNIX)
     list(APPEND PLATFORM_SDK_ONLY_SRC
       optee/linux/enclave.c)
-  else()
+  else ()
     message(FATAL_ERROR "OP-TEE is not yet supported on platforms other than Linux.")
-  endif()
+  endif ()
 
   set(PLATFORM_FLAGS "")
-endif()
+endif ()
 
 if (OE_SGX AND WIN32)
   # oedebugrt is accessed via a bridge on Win32 and need not be linked.
@@ -244,7 +244,7 @@ if (OE_SGX AND WIN32)
   set_source_files_properties(
     sgx/calls.c
     PROPERTIES COMPILE_FLAGS "/Z7")
-endif()
+endif ()
 
 # Common host verification files that work on any OS/architecture.
 list(APPEND PLATFORM_HOST_ONLY_SRC
@@ -286,9 +286,9 @@ add_library(oehostverify STATIC ${PLATFORM_HOST_ONLY_SRC})
 target_link_libraries(oehostverify PUBLIC oe_includes)
 target_link_libraries(oehost PUBLIC oe_includes)
 
-if(WIN32)
+if (WIN32)
   target_link_libraries(oehost PUBLIC ws2_32)
-endif()
+endif ()
 
 if (OE_SGX AND UNIX)
   # Link oedebugrt static library.
@@ -302,13 +302,13 @@ if (OE_SGX AND UNIX)
     sgx/enter.c
     PROPERTIES
     COMPILE_FLAGS "-O2 -fno-omit-frame-pointer")
-endif()
+endif ()
 
 add_dependencies(oehost syscall_untrusted_edl)
 add_dependencies(oehost tee_untrusted_edl)
-if(OE_SGX)
+if (OE_SGX)
   add_dependencies(oehost sgx_untrusted_edl switchless_untrusted_edl)
-endif()
+endif ()
 
 # TODO: Replace these with `find_package` and add as dependencies to
 # the CMake package.
@@ -488,3 +488,5 @@ install(TARGETS oehostapp EXPORT openenclave-targets)
 install(TARGETS oehostverify EXPORT openenclave-targets
   ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}/openenclave/host
   COMPONENT OEHOSTVERIFY)
+
+include(measure/CMakeLists.txt)
--- a/host/sgx/sgxload.c
+++ b/host/sgx/sgxload.c
@@ -2,8 +2,9 @@
 // Licensed under the MIT License.
 
 #include "sgxload.h"
+#if !defined(OEHOSTMR)
 #include <sgx_enclave_common.h>
-
+#endif // OEHOSTMR
 #if defined(__linux__)
 #include <sys/mman.h>
 #include <unistd.h>
@@ -27,6 +28,7 @@
 #include "sgxmeasure.h"
 #include "xstate.h"
 
+#if !defined(OEHOSTMR)
 static int _make_memory_protect_param(uint64_t inflags, bool simulate)
 {
     int outflags = 0;
@@ -106,6 +108,7 @@ static int _make_memory_protect_param(uint64_t inflags, bool simulate)
 
     return outflags;
 }
+#endif // OEHOSTMR
 
 /* Detect the XSave Feature Request Mask (XFRM) to set in the enclave */
 static uint64_t _detect_xfrm()
@@ -150,6 +153,7 @@ static sgx_secs_t* _new_secs(
     return secs;
 }
 
+#if !defined(OEHOSTMR)
 /*
 ** Allocate memory for an enclave so that it has the following layout:
 **
@@ -359,6 +363,7 @@ done:
 
     return result;
 }
+#endif // OEHOSTMR
 
 oe_result_t oe_sgx_initialize_load_context(
     oe_sgx_load_context_t* context,
@@ -418,12 +423,18 @@ oe_result_t oe_sgx_create_enclave(
      * mode or on Linux Kabylake machines. */
     if (context->type == OE_SGX_LOAD_TYPE_CREATE)
     {
+#if !defined(OEHOSTMR)
         if (oe_sgx_is_simulation_load_context(context))
         {
             /* Allocation memory-mapped region */
             if (!(base = _allocate_enclave_memory(enclave_size, context->dev)))
                 OE_RAISE(OE_OUT_OF_MEMORY);
         }
+#else
+        // Wrong code path
+        result = OE_UNSUPPORTED;
+        goto done;
+#endif // OEHOSTMR
     }
 
     /* Create SECS structure */
@@ -438,6 +449,7 @@ oe_result_t oe_sgx_create_enclave(
         /* Use this phony base address when signing enclaves */
         base = (void*)0x0000ffff00000000;
     }
+#if !defined(OEHOSTMR)
     else if (oe_sgx_is_simulation_load_context(context))
     {
         /* Simulate enclave creation */
@@ -464,13 +476,13 @@ oe_result_t oe_sgx_create_enclave(
 
         secs->base = (uint64_t)base;
     }
-
+#endif // OEHOSTMR
     *enclave_addr = base ? (uint64_t)base : secs->base;
     context->state = OE_SGX_LOAD_STATE_ENCLAVE_CREATED;
     result = OE_OK;
 
 done:
-
+#if !defined(OEHOSTMR)
     //  free enclave  memory
     if (result != OE_OK && context != NULL &&
         context->type == OE_SGX_LOAD_TYPE_CREATE && base != NULL)
@@ -478,6 +490,7 @@ done:
         _sgx_free_enclave_memory(
             base, enclave_size, oe_sgx_is_simulation_load_context(context));
     }
+#endif // OEHOSTMR
 
     if (secs)
         oe_memalign_free(secs);
@@ -625,6 +638,7 @@ oe_result_t oe_sgx_load_enclave_data(
         result = OE_OK;
         goto done;
     }
+#if !defined(OEHOSTMR)
     else if (oe_sgx_is_simulation_load_context(context))
     {
         /* Simulate enclave add page */
@@ -685,6 +699,7 @@ oe_result_t oe_sgx_load_enclave_data(
                 protect,
                 enclave_error);
     }
+#endif // OEHOSTMR
 
     result = OE_OK;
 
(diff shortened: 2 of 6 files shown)


nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

host/CMakeLists.txt
```diff
@@ -488,3 +488,321 @@ install(TARGETS oehostapp EXPORT openenclave-targets)
 install(TARGETS oehostverify EXPORT openenclave-targets
   ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}/openenclave/host
   COMPONENT OEHOSTVERIFY)
+
+# Add a library oemr independent of libsgx_enclave_common.so/libsgx_dcap_ql.so
+if (UNIX)
+	set(PLATFORM_OEMR_SRC
+		#    ../common/cert.c
+		#    crypto/openssl/asn1.c
```
radhikaj: Why are these commented out?

anakrish:
@nonpolarity @radhikaj @CodeMonkeyLeet 

My understanding was that we were planning to solve this via delayed loading of the intel shared libraries as opposed to the current static linking of the intel shared libraries.

Though this PR seems to address the specific issue of design, there are other related issues that a delayed loading approach would solve:

By delayed loading,
- oesign will not depend on intel sgx libraries
- Purely simulation workflow will not depend of sgx libraries being present in the system
- Avoids FLC vs NON FLC issues where the published OE SDK only works on SGX FLC machines

Here is an example of how we already delay load azure quote provider library:
https://github.com/openenclave/openenclave/blob/11090df10b820755eb710b08a473f96d5edb3282/host/sgx/windows/sgxquoteproviderloader.c

Here is another example of how we delay load oedebugrt.dll on Windows:
https://github.com/openenclave/openenclave/blob/11090df10b820755eb710b08a473f96d5edb3282/host/sgx/windows/debugrtbridge.c

radhikaj:
> @nonpolarity @radhikaj @CodeMonkeyLeet
> 
> My understanding was that we were planning to solve this via delayed loading of the intel shared libraries as opposed to the current static linking of the intel shared libraries.
> 
> Though this PR seems to address the specific issue of design, there are other related issues that a delayed loading approach would solve:
> 
> By delayed loading,
> 
> * oesign will not depend on intel sgx libraries
> * Purely simulation workflow will not depend of sgx libraries being present in the system
> * Avoids FLC vs NON FLC issues where the published OE SDK only works on SGX FLC machines
> 
> Here is an example of how we already delay load azure quote provider library:
> https://github.com/openenclave/openenclave/blob/11090df10b820755eb710b08a473f96d5edb3282/host/sgx/windows/sgxquoteproviderloader.c
> 
> Here is another example of how we delay load oedebugrt.dll on Windows:
> https://github.com/openenclave/openenclave/blob/11090df10b820755eb710b08a473f96d5edb3282/host/sgx/windows/debugrtbridge.c

@anakrish When we started off, we were trying to address these issues: https://github.com/openenclave/openenclave/issues/2264 and https://github.com/openenclave/openenclave/issues/2544

@nonpolarity has addressed them. We can take further action to remove dependencies on the host side completely, but can do that after.

host/CMakeLists.txt
anakrish: I agree with @radhikaj. These commented lines could be removed.

host/CMakeLists.txt
```diff
@@ -488,3 +488,321 @@ install(TARGETS oehostapp EXPORT openenclave-targets)
...
+    message(FATAL_ERROR "OP-TEE is not yet supported on platforms other than Linux.")
+  endif()
+
+  set(PLATFORM_FLAGS "")
+endif()
+
+if (OE_SGX AND WIN32)
+  # oedebugrt is accessed via a bridge on Win32 and need not be linked.
+  list(APPEND PLATFORM_OEMR_SRC
+    sgx/windows/debugrtbridge.c)
```
anakrish: This seems off. oesign is not concerned with debugging and would not need to do include debugrt stuff.

host/CMakeLists.txt
```diff
@@ -488,3 +488,321 @@ install(TARGETS oehostapp EXPORT openenclave-targets)
...
+#  set_source_files_properties(oemr
+#    sgx/enter.c
+#    PROPERTIES
+#    COMPILE_FLAGS "-O2 -fno-omit-frame-pointer")
+#endif()
+
+#add_dependencies(oemr syscall_untrusted_edl)
+#add_dependencies(oemr tee_untrusted_edl)
+if(OE_SGX)
+  add_dependencies(oemr sgx_untrusted_edl switchless_untrusted_edl)
```
anakrish: oesign most likely does not need any edl stuff either.

host/CMakeLists.txt
```diff
@@ -488,3 +488,321 @@ install(TARGETS oehostapp EXPORT openenclave-targets)
...
+set_property(TARGET oemr PROPERTY
+  ARCHIVE_OUTPUT_DIRECTORY ${OE_LIBDIR}/openenclave/host)
+
+if (UNIX)
+  target_link_libraries(oemr INTERFACE
+      -Wl,-z,noexecstack)
+endif ()
+
+# Install targets
+install(TARGETS oemr EXPORT openenclave-targets
```
anakrish: Since oems is used only to build oesign, and is not intended to be used by OE SDK user, we need not install it.

host/CMakeLists.txt
nonpolarity: Thanks, I will clean up when everything is done.

host/CMakeLists.txt
nonpolarity: Since sgx_untrusted_edl  depends on sgx_u.h, add_dependencies can ensure dependency on sgx_u.h. I will try to find if there is a straightforward way for dependency on file.

host/CMakeLists.txt
nonpolarity: Good point, this is just a internal and intermediate library.

nonpolarity:
bors try

nonpolarity:
bors try

host/CMakeLists.txt
```diff
@@ -307,7 +307,7 @@ endif()
 add_dependencies(oehost syscall_untrusted_edl)
 add_dependencies(oehost tee_untrusted_edl)
 if(OE_SGX)
-  add_dependencies(oehost sgx_untrusted_edl switchless_untrusted_edl)
+  add_dependencies(oehost sgx_untrusted_edl)
```
anakrish: Accidental removal of switchless_untrusted_edl?

radhikaj:
@anakrish Please discuss various approaches you have for this including delay loading and stubbing and removing host side dependencies.

nonpolarity:
bors try

host/CMakeLists.txt
nonpolarity: updated, thanks.

host/CMakeLists.txt
nonpolarity: Really bad accident, thanks.

nonpolarity:
> 
> 
> @nonpolarity Thanks for addressing the comments. Looks much simpler to understand.
> 
> Couple of suggestions:
> 
>     * It is not clear what `oemr` means. It would be good to use more descriptive names for the library as well as for the C preprocessor define
> 
>     * Since oemr and oehost do not seem to be related, it would be cleaner to create an oemr folder with just a CMakeLists.txt and a README.md. oemr/CMakeLists.txt could refer to files in host, and the README.md could describe the intent of the library.

Thanks for these suggestions. The library name is changed to liboehostmr, following naming convention of other host side libraries. The cmake config file is also moved to a subdir.

host/OEHOSTMR/CMakeLists.txt
```diff
@@ -0,0 +1,189 @@
...
+# Licensed under the MIT License.
+
+# Add a library oehostmr independent of libsgx_enclave_common/libsgx_dcap_ql
+if (UNIX)
+  set(PLATFORM_HOST_MR_SRC
+    crypto/openssl/init.c
+    crypto/openssl/key.c
+    crypto/openssl/rsa.c
+    crypto/openssl/sha.c
+	linux/hostthread.c
```
anakrish: Kindly fix the spacing issues in this file.

host/CMakeLists.txt
```diff
@@ -488,3 +488,5 @@ install(TARGETS oehostapp EXPORT openenclave-targets)
 install(TARGETS oehostverify EXPORT openenclave-targets
   ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}/openenclave/host
   COMPONENT OEHOSTVERIFY)
+
+include(OEHOSTMR/CMakeLists.txt)
```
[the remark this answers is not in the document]
anakrish: It would make sense to add a line `add_subdirectory(oehostmr)` here at the top-level CMakeLists.txt:
https://github.com/openenclave/openenclave/blob/99410b10b4782d33e76819bc681e651e96cd9739/CMakeLists.txt#L213

anakrish:
@CodeMonkeyLeet @nonpolarity 
>
> Thanks for tackling this! Since this is still in draft, have you considered making oehostmr a variant of the oehost build just as oehostverify is generated from the same CMakeLists.txt today?
>

Responding since I had asked Weichen to move the code out from host/CMakeLists.txt into a separate oehostmr/CMakeLists.txt. I prefer separate cmakefile since there is no change to oehost component other than conditional compilation in some source files. host/CMakeLists.txt is already a bit too unwieldy. This way, If it turns out that delay loading would solve the same problem, then it is easier to remove oehostmr component entirely; otherwise we may have to revert oehost's CMakeLists.txt to an earlier version.

host/OEHOSTMR/CMakeLists.txt
```diff
@@ -0,0 +1,189 @@
...
+    crypto/bcrypt/key.c
+    crypto/bcrypt/pem.c
+    crypto/bcrypt/random.c
+    crypto/bcrypt/rsa.c
+    crypto/bcrypt/sha.c
+    windows/hostthread.c
+    windows/syscall.c
+    windows/time.c)
+else()
+  message(FATAL_ERROR "Unknown OS. Only supported OSes are Linux and Windows")
```
dthaler: grammar: "Only" -> "The only"

host/OEHOSTMR/CMakeLists.txt
```diff
@@ -0,0 +1,189 @@
...
+    list(APPEND PLATFORM_HOST_MR_SRC
+      sgx/linux/xstate.c)
+  else()
+    list(APPEND PLATFORM_HOST_MR_SRC
+      sgx/windows/xstate.c)
+  endif()
+
+  set(PLATFORM_FLAGS "-m64")
+elseif(OE_TRUSTZONE)
+    list(APPEND PLATFORM_HOST_MR_SRC
```
dthaler: nit: indentation seems off

host/OEHOSTMR/CMakeLists.txt
```diff
@@ -0,0 +1,189 @@
...
+  if (UNIX)
+    list(APPEND PLATFORM_HOST_MR_SRC
+      sgx/linux/xstate.c)
+  else()
+    list(APPEND PLATFORM_HOST_MR_SRC
+      sgx/windows/xstate.c)
+  endif()
+
+  set(PLATFORM_FLAGS "-m64")
+elseif(OE_TRUSTZONE)
```
dthaler: nit: use consistent spacing.  Usually this file has a space before (like in line 13 etc) but here doesn't.  Similarly line 97 and 110 have different styles. Pick a style and be consistent throughout this file.

host/OEHOSTMR/README.md
```diff
@@ -0,0 +1,3 @@
+The `liboehostmr` is a subset of liboehost, and it's main functionality is to get the enclave's measurement. Following other host libraries' naming convention, this library starts with liboehost, and `mr` is short for `measurement report`.
```
dthaler: grammar: "it's" should be "its"

host/OEHOSTMR/README.md
```diff
@@ -0,0 +1,3 @@
...
+
+To reuse source code, a macro `OEHOSTMR` is used that some sentences refering to libsgx_enclave_common are stripped off. Thus this library has no dependency on libsgx_enclave_common(also libsgx_dcap_ql).
```
dthaler: grammar: can't parse "is used that some sentences … are stripped off".
Please reword.   Also add space before open paren.

CodeMonkeyLeet:
> @anakrish 
> 
> > Thanks for tackling this! Since this is still in draft, have you considered making oehostmr a variant of the oehost build just as oehostverify is generated from the same CMakeLists.txt today?
> 
> Responding since I had asked Weichen to move the code out from host/CMakeLists.txt into a separate oehostmr/CMakeLists.txt. I prefer separate cmakefile since there is no change to oehost component other than conditional compilation in some source files. host/CMakeLists.txt is already a bit too unwieldy. This way, If it turns out that delay loading would solve the same problem, then it is easier to remove oehostmr component entirely; otherwise we may have to revert oehost's CMakeLists.txt to an earlier version.

@nonpolarity Please feel free to disregard this comment then; I don't feel strongly about this and I didn't mean to randomize you.

host/OEHOSTMR/CMakeLists.txt
nonpolarity: Thanks, tab is replace with space.

host/CMakeLists.txt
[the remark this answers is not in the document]
nonpolarity: @anakrish since oehostmr is part of host as oehostverify, is it better to stay under host directory? Actually the working directory for oehostmr is also host not host/oehostmr, because change working directory requires lots of change of cmake config and even head file reference. So right now include(OEHOSTMR/CMakeLists.txt) is a easy way to separate the config file. Let me know if I have some misunderstanding.

host/OEHOSTMR/CMakeLists.txt
```diff
@@ -0,0 +1,189 @@
+# Copyright (c) Open Enclave SDK contributors.
```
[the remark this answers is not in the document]
nonpolarity: Yes there is a dilemma that the dir is not a  sub-dir of source code as sgx and crypto, neither a dependent cmake sub-module of host. Thus this dir is in upper case. I will change it to host/measure.

host/sgx/sgxload.c
```diff
@@ -418,12 +423,14 @@ oe_result_t oe_sgx_create_enclave(
      * mode or on Linux Kabylake machines. */
     if (context->type == OE_SGX_LOAD_TYPE_CREATE)
     {
+#if !defined(OEHOSTMR)
```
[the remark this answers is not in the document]
nonpolarity: Good suggestion, a return code OE_NOT_IMPLEMENTED is added.

host/OEHOSTMR/CMakeLists.txt
nonpolarity: Fixed, thanks.

host/OEHOSTMR/CMakeLists.txt
nonpolarity: Fixed, thanks.

host/OEHOSTMR/README.md
nonpolarity: Fixed, thanks.

host/OEHOSTMR/CMakeLists.txt
nonpolarity: Tab is fixed, thanks.

host/OEHOSTMR/README.md
nonpolarity: Fixed, thanks.

CodeMonkeyLeet:
```suggestion
The `liboehostmr` library reuses the source code from `liboehost` by using the preprocessor definition `OEHOSTMR`. When specified, `OEHOSTMR` removes any code references to libsgx_enclave_common or libsgx_dcap_ql so that `liboehostmr` has no dependencies on those libraries.
```

CodeMonkeyLeet: It looks like there's a merge conflict and both upper case and lower case versions of the file exist in this PR now?

common/result.c
```diff
@@ -130,6 +130,8 @@ const char* oe_result_str(oe_result_t result)
             return "OE_ALREADY_EXISTS";
         case OE_ALREADY_INITIALIZED:
             return "OE_ALREADY_INITIALIZED";
+        case OE_NOT_IMPLEMENTED:
```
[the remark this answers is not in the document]
nonpolarity: Got it, thanks.

host/OEHOSTMR/README.md
nonpolarity: Thanks for rewriting the whole paragraph! Yes I forget to use `git mv` but `mv` only, now it is fixed.

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

radhikaj:
@nonpolarity Thanks so much or this fix.

jhand2:
Awesome job getting this done! People have been complaining about this in gh issues for a while

PR #2647, opened 2020-03-10
Design of Extended Enclave Initialization Data

wintersteiger:
This PR is for the design discussion of Extended Enclave Initialization Data, see `docs/DesignDocs/ExtendedEnclaveInitializationData.md`.

Summary: Upon enclave creation (`oe_create_enclave_eeid`) the enclave image is extended with additional (instance-specific) data and re-signed with the debug key. The extended data also contains the original sigstruct of the re-signed enclave image, which is verified during extended quote verification. This allows us to specialize enclave image during enclave creation without the need for an entirely new image and a different signing key, while maintaining security/confidentiality guarantees.

CC @sylvanc @ad-l @letmaik @mikbras @jxyang


docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,154 @@
+Extended Enclave Initialization Data
+=====
+
+We present an extension of Intel SGX attestation in OpenEnclave that enables the attestation of enclave initialization parameters.
```
dthaler: Line 1 is not SGX specific, but line 4 is SGX specific.  Which is it?  Please correct one of the two lines.

Personally I think that it should NOT be SGX specific, and the design below is too SGX specific.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,154 @@
...
+Our extension uses MRENCLAVE as a platform configuration register to measure additional data besides the signed enclave image, such as instance-specific enclave configuration and the settings in the current enclave signing configuration.
+
+We introduce a new OE quote type for SGX quotes with extended enclave initialization data (EEID).
+Recipients of such quotes must be able to recover and verify the original enclave identity information (including the original MRENCLAVE and MRSIGNER of the base image), based on the knowledge of the initialization data, which is transmitted alongside the quote as a new collateral type.
+
+
+Motivations
+----------
+
+With current SGX quotes, applications can attest their configuration as part of the `reportData` argument of `oe_get_report`.
```
dthaler: We need to decouple discussion of evidence format from discussion of APIs.
APIs are supposed to be evidence format agnostic, not evidence format specific.
If you create a new evidence format, that's ok, but I think we need to find a way to have a generic API that is format agnostic and not add a new API per format.
(Note: oe_get_report is named poorly since it uses an SGX-specific name, rather than the IETF-agnostic name, and should really be oe_get_evidence.  Maybe we will rename it to that in the future?)

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,154 @@
...
+This mechanism works robustly when the application code is entirely and statically known to the validator of the quote.
+
+Things get more complicated if an enclave executes user code that is not part of the signed enclave image. For instance, consider an enclave containing a JavaScript interpreter that executes user scripts from the host and has access to the `oe_get_report` API. It is impossible to know which script has been executed by such an enclave based on a standard quote, even if the hash of the script is included in the `reportData`, because a malicious script can obtain a valid quote with `oe_get_report` pretending to be a honest script. Similarly, if an enclave loads and execute arbitrary assembly code from the host, this assembly code can use the `EGETQUOTE` instruction to create valid reports - hence, it is impossible to use attestation to tell which assembly code has been loaded by the enclave.
```
dthaler: This paragraph again seems to convolute evidence format and API.
The agnostic API supports (or needs to support) layered attestation, such as DICE cert chains, for TrustZone.
Once you have layered attestation, then I believe you already have a solution to the interpreter issue being discussed here.  See https://tools.ietf.org/html/draft-ietf-rats-architecture-02#section-4.2 for the IETF summary.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,154 @@
...
+```C
+oe_result_t oe_create_enclave_eeid(
+    const char* enclave_path,
+    oe_enclave_type_t enclave_type,
+    uint32_t flags,
+    const oe_enclave_setting_t* settings,
+    uint32_t setting_count,
+    const oe_ocall_func_t* ocall_table,
+    uint32_t ocall_count,
+    oe_eeid_t* eeid,
```
dthaler: I would instead recommend a design that makes layered attestation be a first class concept, and is integrated with the attestation plugin design.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
CodeMonkeyLeet: To extrapolate Dave's point a bit, I'm inferring that this design needs a layered _measurement_ framework to support the kind of layered attestation; the desire to bind the attestation promises to SGX hardware root of trust implies injecting custom measurements into the only avenue that SGX1 provides for this, which is at sign/load time.

Independent of the EEID proposal, it's worth considering whether the loader can be replaced in a customizable way to permit this, and also to tie that in to a contract with attester plugin registration (i.e. forms of attestation that are predicated on a specific load format must understand whether they can be appropriately supported and fail early if not.)

docs/DesignDocs/ExtendedEnclaveInitializationData.md
wintersteiger: I don't think the concept itself requires anything SGX-specific, so I agree that we should keep the design document platform independent. Currently, all we need is an SGX implementation though and I think that would be fine with the idea of a specific SGX attestation plugin.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
wintersteiger: Absolutely, the APIs should be as generic as possible. We can put the EEID into `oe_enclave_setting_t` so that it doesn't require a change to the API function itself. I think the verification can go into a verifier plugin, so that the API surface doesn't change.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
wintersteiger: A customizable loader is a great idea - I think that's the only bit that can't go into the current plugin infrastructure. It sounds like major surgery to me though. Are there other extensions that would benefit from that?

docs/DesignDocs/ExtendedEnclaveInitializationData.md
wintersteiger: I'm working on plugins at the moment, I think we can do most of what we want to do inside the plugin infrastructure, except for the bit that injects the additional attested data pages.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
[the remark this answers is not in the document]
wintersteiger: Yes, that should be `EREPORT`.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,154 @@
...
+
+An important instance of this problem is Azure Confidential Containers: the base enclave image contains the SGX-LKL runtime, which executes a user container. The container contains arbitrary code, which can use the `EGETQUOTE` instruction to obtain a valid quote for arbitrary `reportData` – potentially impersonating other containers.
+
+One way to solve this issue is to re-compile and re-sign a new enclave image for every container we execute (with the container data being measured together with the LKL base). However, this approach is cumbersome as we would like to be able to launch arbitrary containers on-demand without having to build and sign a new enclave image every time.
+
+Another class of problems we address is dynamic attestation of the memory and thread configuration. In current OE attestation, these settings (`NumStackPage`, `NumHeapPages`, `NumTCS`) are part of the configuration to sign an enclave. This means that it is not possible to deploy the same enclave image with different memory and core configurations.
```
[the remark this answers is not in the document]
wintersteiger: The re-signer may not be the same as the original signer, e.g. image/code creator may be different from the deployment infrastructure and recompiling/signing would remove the original signature. EEID allows us to keep the original image/code signature while allowing us to customize the memory settings at deployment time (signed with a different key). Both signatures can then be verified together or independently later.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,154 @@
...
+            context, enclave->addr, pages, num_pages, vaddr));
+
+        if (ee_sz < epg_sz)
+            free(pages);
+
+        OE_SHA256 ext_mrenclave;
+        oe_sha256_final(hctx, &ext_mrenclave);
+```
+
+Finally, since the changes invalidate the signature of the base image, we need to dynamically re-sign the enclave. For this, we use the `OE_DEBUG_SIGN_KEY`, which is well-known by all OE users. Note that applications should never rely on the apparent `MRSIGNER` of an EEID quote. Indeed, what we reflect to the application when we validate an EEID quote is the `MRSIGNER` of the base image.
```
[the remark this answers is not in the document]
wintersteiger: Zero requirement: I don't think so, we just chose to do that in our draft implementation because it was easy to do. Without that, we would have to re-hash the image up to the point before memory pages are added and during verification we would have to extend the image with the original pages that the signature was over.

Enforced: Yes, see `OE_RAISE(OE_INVALID_PARAMETER);` further above.

You're right, for the future we should not rely on those settings being zero, but everything would run a bit faster when they can be.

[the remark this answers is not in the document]
wintersteiger: There's nothing that stops us from using any other key, the debug key is just convenient because it comes with OE, so we don't need to worry about getting the key to the user.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,154 @@
...
+
+```C
```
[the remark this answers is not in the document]
wintersteiger: I think oesign does not need to be modified, but whoever verifies a quote needs to know the EEID to verify the quote from the extended enclave.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
jhand2: @CodeMonkeyLeet I suppose it depends on how its implemented. A lot of features on OE take hard dependencies on where/how things are laid out in memory. Did you have a thought about what exactly the developer would be able to customize?

i.e. are we talking wholesale replacement of the loader, or are we talking passed parameters to be able to inform the loader about abitrary things the application would like loaded/measured?

wintersteiger:
I've just pushed a couple minor fixes. What's the next step for this proposal to go forward?

I still like the idea of loader plugins - is this something that OE would like to add or is somebody already working on this? It feels like a major undertaking to me, but it may well be easier than I currently assume.

jhand2:
@wintersteiger Just wanted to make sure I didn't miss something, are the attestation/verification details for EEID attestation intentionally not included in this doc? I only see the loader proposal and want to make sure I understand the whole picture.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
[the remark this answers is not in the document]
wintersteiger: I was wrong, I did end up having to modify oesign to extract the base-image hash state, vaddr, and entry_point to make those values available at image build time instead of enclave creation time. It's fairly straight-forward though.

wintersteiger:
That's a fair point @jhand2. We didn't the details would be important here, but I'll add a paragraph on Monday anyways. The details are boring, but it works pretty much as you would expect: verify base image evidence, check options are the same, get base image hash state and hash the data pages (with new size settings), check that the final hashes match. (The unfinished plugin version may be the most readable: https://github.com/wintersteiger/openenclave/blob/7e6e008dcde4fedcf3ade3980bc9f2e5f4eb557c/tests/eeid_plugin/plugin/eeid_verifier.c#L30).

docs/DesignDocs/ExtendedEnclaveInitializationData.md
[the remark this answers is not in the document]
CodeMonkeyLeet: Could you include a brief paragraph on those changes as well here?

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,156 @@
...
+            &ext_mrenclave,
+            properties->config.attributes,
+            properties->config.product_id,
+            properties->config.security_version,
+            OE_DEBUG_SIGN_KEY,
+            OE_DEBUG_SIGN_KEY_SIZE,
+            sigstruct));
+```
+
+We also add a serialization function `oe_serialize_eeid` for the `oe_eeid_t`, so that EEID can be sent to verifiers as an endorsement for verification of the base image evidence together with the EEID configuration of the instance.
```
CodeMonkeyLeet, line 149: From an earlier unaddressed comment:
> This is problematic as it tightly couples a specific attester/verifier implementation to a general public API provided by the OE SDK. Do you expect that those are also ingested as part of the OE SDK?
>
> Specifically, it feels like the OE SDK is modified to privilege a specific use case as inherent in the OE API even though it's not generalizable beyond SGX1; In SGX2 with EDMM and KSS support, the problems that this design attempts to solve are natively supported. It's not clear to me how this design migrates forward to SGX2, or generalizes to other TEE, which is a core architectural premise of the OE SDK.

This still holds true for the updated design; I think that at a higher level, it's inevitable that OE will move a lot of the SGX1-specific semantics out of its API surface into a TEE extension model; at that point, this will likely be factored out into the TEE extension. In the meantime, it's useful to explore the use cases for something like this, and I would be in favor of having this as an [experimental API](https://github.com/openenclave/openenclave/blob/d7c8e614221a2ad586b2a4eae144b1672f636e9a/docs/ApiGuidelines.md#experimental-apis), with the expectation that it will likely evolve further into a different API.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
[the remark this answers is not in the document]
CodeMonkeyLeet: Can we remove this requirement then? Alternatively, have we considered adding a different metadata blob type and explicitly classifying this as a different enclave type?

docs/DesignDocs/ExtendedEnclaveInitializationData.md
CodeMonkeyLeet: @jhand2 Yes, depending on the TEE implementation, this could be a wholesale replacement of the loader (c.f. OP-TEE). We will likely set up a SIG to discuss in greater detail where to take this moving forwards.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
dthaler: I like the idea of "having this be an experimental API with the expectation that it will likely evolve further", and that would let this be merged now.

wintersteiger: I totally agree, this is clearly experimental, and I can see that changes will likely be necessary for non-SGX platforms (e.g. at least a replacement for sigstruct). I'll check with the team, but I don't expect any objections.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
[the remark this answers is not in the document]
wintersteiger: Sure, we can do that. I'd be happy to make it an entirely new enclave type (which could also be sgx-only for now?), although so far I thought of it more of an option to existing types rather than a different type altogether.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
wintersteiger: Forgot to address the SGX2 comment before: SGX2's config register does not solve the same problem, as it provides a slightly weaker security guarantee. As in our example of a script running in an interpreter within the enclave, suppose that interpreter and script come from different parties, but they are now running in the same (enclave) memory. The config register does not address the problem of malicious scripts that have access to `oe_get_report` (and similar) which they can use to fake evidence, even if the config register matches.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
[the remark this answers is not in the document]
wintersteiger: Done!

kuanggang-git:

CodeMonkeyLeet:
@wintersteiger:
> I totally agree, this is clearly experimental, and I can see that changes will likely be necessary for non-SGX platforms (e.g. at least a replacement for sigstruct). I'll check with the team, but I don't expect any objections.

Did you manage to get confirmation from your team? I think that's the last update folks are waiting for in terms of approving this PR.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,168 @@
...
+
+Verification of EEID reports
+----------------------------
+
+Verification of reports with EEID requires a fully populated `oe_eeid_t` and it performs the following steps (not necessarily in this order):
+
+- Create a SHA256 context and restore the internal state to `hash_state`,
+- re-measure the addition of heap, stack, and TCS pages with the settings from `size_settings` (starting at address `vaddr` and with `entry_point` in the TCS control pages),
```
jxyang, line 157: Just to be clear, the heap, stack and TCS pages are assumed to be 0-filled for measurement purpose, right?

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,168 @@
...
+- measure the additional page containing the `oe_eeid_t` (filled with the entire `oe_eeid_t`, including `data`),
+- check that the final hash matches the hash reported in the extended report (e.g. MRENCLAVE),
+- check the identity of the signer of the extended image (e.g. public key corresponding to `OE_DEBUG_SIGN_KEY`),
```
jxyang, line 160: It's not clear to me how could we prevent a malicious program from hacking the EEID structure and re-sign it, since `OE_DEBUG_SIGN_KEY` is known to every one.

wintersteiger:
@CodeMonkeyLeet thanks for checking, yes, everyone is happy with the experimental status!

docs/DesignDocs/ExtendedEnclaveInitializationData.md
wintersteiger: It looks like we'll go for an experimental API for now, but yes, we can also put the EEID into the settings.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,168 @@
...
+
+In today's OE evidence, applications can attest their configuration as part of the `report_data` argument of `oe_get_report`.
+This mechanism works robustly when the application code is entirely and statically known to the verifier of the evidence.
+
+Things get more complicated if an enclave executes user code that is not part of the signed enclave image. For instance, consider an enclave containing a JavaScript interpreter that executes user scripts that originate from the untrusted host in shared enclave memory, and which has access to the `oe_get_report` API. It is impossible to know which script has been executed by such an enclave based on traditional evidence, even if the hash of the script is included in the `report_data`, because a malicious script can obtain valid evidence via `oe_get_report`, while pretending to be an honest script. Similarly, if an enclave loads and executes arbitrary assembly code from the host, this assembly code can use an in-enclave CPU instruction, for instance the `EREPORT` instruction on SGX, to create valid evidence - hence, it is impossible to use traditional attestation to determine which assembly code was executed by the enclave.
+
```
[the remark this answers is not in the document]
wintersteiger: In theory a script could modify the `oe_eeid_t `, but attestation would fail because the corresponding EPC pages, and thus the enclave hash, are modified

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,168 @@
...
+The additional argument is of type `oe_eeid_t`, which contains the following information:
+
+```C
+typedef struct oe_eeid_t_
+{
+    uint32_t hash_state[10]; /* internal state of the hash at the end of the enclave base image */
+    uint8_t sigstruct[1808]; /* complete sigstruct (or similar object for other TEE types) computed for the base image */
+    oe_enclave_size_settings_t size_settings; /* heap, stack and thread configuration for this instance */
+    uint64_t vaddr;          /* location of application EEID in the image*/
```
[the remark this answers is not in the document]
wintersteiger: Sure, what would you suggest?

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,168 @@
...
+                (uint64_t*)(oeimage.image_base + sym_props.st_value);
+            oe_sgx_enclave_properties_t* p =
+                (oe_sgx_enclave_properties_t*)sym_props_addr;
+            p->header.size_settings = props.header.size_settings;
+        }
+    }
+```
+
+Then, after the enclave image is loaded but before it is initialized, we add the additional memory pages containing the contents of the `oe_eeid_t` struct:
+
```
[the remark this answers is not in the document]
wintersteiger: The new pages are measured and thus the final mrenclave is different, so there is no reason for `enclave_initialize` to fail.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
wintersteiger: No, they are filled with the expected contents, they mirror the addition of pages in the traditional setting.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
wintersteiger: It can attempt to do that, but remote attestation would fail. Since the EEID pages are measured, any change to them would make the enclave hash change. The debug key signature is not much more than an integrity check; it avoids distributing new keys. Note that the base image hash and signature (the whole sigstruct infact) are kept and verified, so nothing changes there.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
jxyang: I hope you understand the hassle of introducing a new OE public API and then later retracting it. Generally we'd try to avoid that.

wintersteiger: Yes, I do, and that's why I proposed new functions called `*_eeid` instead of modifying existing ones. Moving the extra argument into the settings of the existing functions effectively modifies the API too.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,168 @@
...
+    oe_enclave_t** enclave_out)
+```
+
+The additional argument is of type `oe_eeid_t`, which contains the following information:
+
+```C
+typedef struct oe_eeid_t_
+{
+    uint32_t hash_state[10]; /* internal state of the hash at the end of the enclave base image */
+    uint8_t sigstruct[1808]; /* complete sigstruct (or similar object for other TEE types) computed for the base image */
```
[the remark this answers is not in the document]
wintersteiger: That's the size of the Intel sigstruct, so that number is from the SGX spec. To support other TEEs with the same structure, this would have to be a variable size buffer. I'll change that.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,168 @@
...
+
+```C
+        oe_sha256_context_t* hctx = &context->hash_context;
+        sgx_sigstruct_t* sigstruct = (sgx_sigstruct_t*)properties->sigstruct;
+        memcpy(eeid->sigstruct, (uint8_t*)sigstruct, sizeof(sgx_sigstruct_t));
+        eeid->vaddr = *vaddr;
+        eeid->entry_point = oeimage.entry_rva;
+
+        uint64_t ee_sz = sizeof(oe_eeid_t) + eeid->data_size;
+        uint64_t epg_sz = eeid_pages_size(eeid);
```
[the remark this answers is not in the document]
wintersteiger: It's the size (in bytes) of the EEID pages. I'll rename it.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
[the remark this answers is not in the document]
jxyang: By reading the code, it seems `hash_state` is the hash arrived after adding TCS/stack/heap pages to base image MRENCLAVE, right? The name should reflect which hash value we are dealing with here.

The `EEID` is not in the elf image. So either the comment for `vaddr` or the variable name itself is confusing.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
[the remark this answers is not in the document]
jxyang: The more detailed answer probably should be "the extra pages are both measured in software (`oe_sgx_measure_load_enclave_data`)  and hardware (`enclave_load_data`), so `enclave_initialize` succeeds because both software and hardware agree on the same final mrenclave value."

docs/DesignDocs/ExtendedEnclaveInitializationData.md
jxyang: Looking at the implementation PR (#2611), it seems the heap/stack/tcs pages are not re-measured. Rather, they are included in `hash_state`. Is my understanding correct? If so, please correct either the implementation or the description here.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
jxyang: Please include the fact that the final mrenclave (and the whole sigstruct) is integrity protected by sgx hardware.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
wintersteiger: Sure, sounds good. I've changed the document accordingly.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,168 @@
...
+    uint8_t sigstruct[1808]; /* complete sigstruct (or similar object for other TEE types) computed for the base image */
+    oe_enclave_size_settings_t size_settings; /* heap, stack and thread configuration for this instance */
+    uint64_t vaddr;          /* location of application EEID in the image*/
+    uint64_t entry_point;    /* entry point of the image */
+    uint64_t data_size;      /* size of application EEID */
+    uint8_t data[];          /* actual application EEID */
+} oe_eeid_t;
+```
+
+Once an enclave has been started with EEID, it can use `oe_get_report` as usual to obtain evidence, except that this evidence will not be verifiable by enclaves built without EEID support, e.g. the Intel SGX SDK.
```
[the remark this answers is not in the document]
wintersteiger: Yeah, that was essentially a typo :-)

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,168 @@
...
+
+In today's OE evidence, applications can attest their configuration as part of the `report_data` argument of `oe_get_report`.
+This mechanism works robustly when the application code is entirely and statically known to the verifier of the evidence.
+
+Things get more complicated if an enclave executes user code that is not part of the signed enclave image. For instance, consider an enclave containing a JavaScript interpreter that executes user scripts that originate from the untrusted host in shared enclave memory, and which has access to the `oe_get_report` API. It is impossible to know which script has been executed by such an enclave based on traditional evidence, even if the hash of the script is included in the `report_data`, because a malicious script can obtain valid evidence via `oe_get_report`, while pretending to be an honest script. Similarly, if an enclave loads and executes arbitrary assembly code from the host, this assembly code can use an in-enclave CPU instruction, for instance the `EREPORT` instruction on SGX, to create valid evidence - hence, it is impossible to use traditional attestation to determine which assembly code was executed by the enclave.
```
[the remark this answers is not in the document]
wintersteiger: I've updated the motivation section accordingly, I hope this explains it better now. @ad-l: could you take a look at my changes too? Thanks!

docs/DesignDocs/ExtendedEnclaveInitializationData.md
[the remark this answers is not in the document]
wintersteiger: Yes, in the case of SGX that's the MRENCLAVE, I assumed "enclave hash" was the platform-independent term for this. Is there a better term?

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,168 @@
...
+
+An important instance of this problem is Azure Confidential Containers: the base enclave image contains the SGX-LKL runtime, which executes a user container. The container contains arbitrary code, including the `EREPORT` instruction to obtain valid evidence for arbitrary `report_data` – potentially impersonating other containers.
```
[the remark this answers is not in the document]
wintersteiger: Done

docs/DesignDocs/ExtendedEnclaveInitializationData.md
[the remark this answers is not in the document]
wintersteiger: No, the hash state is the _internal state_ of the hash computation _before_ the addition of data pages. The hash computation continues from that point onward, by incremental updates via `oe_sha256_update`, not by computing `hash(hash(...))`. The actual hash is computed only at the end via `oe_sha256_final`. We would have preferred `hash(hash(...))` too, as that would have simplified everything a great deal, but sadly that's not what the CPU (SGX) does.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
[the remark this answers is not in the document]
wintersteiger: Yes, they both measure the same pages.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
wintersteiger: Yes, the customizable loader has been proposed before, that's a good way to treat it, I think.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
wintersteiger: Yes, that was a bug in the implementation that's been fixed, but not pushed yet (we're still working out a tricky bit relating to the memory resizing; the updated page replay is here if you want to take a look https://github.com/wintersteiger/openenclave/commit/d292596a0cddba7f69a03b23dec7d031a0a5fe1c )

docs/DesignDocs/ExtendedEnclaveInitializationData.md
wintersteiger: Done!

docs/DesignDocs/ExtendedEnclaveInitializationData.md
ad-l: Despite the semantic overloading of `oe_enclave_setting_t` between pre-load and post-load configuration, I am OK with putting EEID there to minimize the impact on the public API.

docs/DesignDocs/ExtendedEnclaveInitializationData.md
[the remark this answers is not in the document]
ad-l: Sorry, that was my mistake. Mean `EREPORT` here

docs/DesignDocs/ExtendedEnclaveInitializationData.md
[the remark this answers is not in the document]
ad-l: @CodeMonkeyLeet the idea behind the zero requirement is to ensure that an enclave image meant to be used with EEID is not accidentally loaded without it. In other words, when the image is signed, the signer either decides that the heap & stack configuration is fixed, which makes it impossible to change at load time with EEID, or it set at load time, which then makes the image impossible to load without EEID. Does that make sense?

Now, the idea to sign an invalid configuration is a bit of a hack to implement this concept, and I understand your nervousness. Alternatively, we can add an explicit flag set at enclave signing time that indicates whether the image is meant to be loaded with EEID or not. It is also possible to make this more flexible by having 3 possible settings (EEID disabled, EEID optional, EEID mandatory) but I personally think it is dangerous to make EEID optional when we rely on it for secure key release policies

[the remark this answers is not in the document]
ad-l: @CodeMonkeyLeet I understand your objection. The downside of an ephemeral key is that it will negatively impact the enclave loading time - it takes a while to generate the primes for the ephemeral signing key...

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,154 @@
...
+            properties->config.attributes,
+            properties->config.product_id,
+            properties->config.security_version,
+            OE_DEBUG_SIGN_KEY,
+            OE_DEBUG_SIGN_KEY_SIZE,
+            sigstruct));
+```
+
+The changes to enclave initialization must be reflected for the validation of EEID quotes.
+We also introduce a serialization function `oe_serialize_eeid` for the `oe_eeid_t`, as it is required by the verifier to re-compute and validate both the `MRENCLAVE` and `MRSIGNER` of the base image, based on the configuration of the instance.
```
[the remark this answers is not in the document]
ad-l: It is an SGX1-specific implementation of the verifier, but it doesn't add any change to the generic attestation interface, which should generically support the concept of load-time configuration in addition to enclave held data. The implementation of load-time configuration attestation will be native in SGX2.

radhikaj:
> @CodeMonkeyLeet thanks for checking, yes, everyone is happy with the experimental status!

Thanks so much @wintersteiger. I might be missing something, but dont see the doc say that these APIs are experimental. Would it be possible for you to that?

wintersteiger:
@radhikaj of course, except I'm not sure what exactly you're asking for. Should I just add the word 'experimental' to the text, add a GitHub tag, or is there some other design document convention that I'm not aware of?

wintersteiger:
I have now updated the text and the code snippets to make it clear that this is experimental, please do let me know if it needs to be added elsewhere too.

jhand2:
Thanks for sticking with this! Could you squash you commits and apply DCO sign-off?

docs/DesignDocs/ExtendedEnclaveInitializationData.md
```diff
@@ -0,0 +1,168 @@
...
+
+One way to solve this issue is to re-compile and re-sign a new enclave image for every container we launch (with the container data being measured together with the LKL base image). However, this approach is cumbersome as we would like to be able to launch arbitrary containers on demand without having to build and sign a new enclave image every time.
+
+Another class of problems we address is dynamic attestation of the memory and threading configuration. In current OE attestation, these settings (`NumStackPage`, `NumHeapPages`, `NumTCS`) are fixed at enclave signinig time. This means that it is not possible to deploy the same enclave image with different memory and threading configurations.
+
+User Experience
+---------------
+
+The user decides to enable the use of EEID when the enclave base image is signed: while not strictly necessary, we require that EEID enclaves use `NumStackPages=0`, `NumHeapPages=0`, and `NumTCS=0`. This guarantees that enclave images meant to be used with EEID cannot be accidentally initialized with traditional attestation (the enclave would fail to load).
```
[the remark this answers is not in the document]
achamayou: Perhaps this can just be rephrased as "All parameters specified at enclave signature time should be set to a value that would prevent the enclave from loading, typically 0. In the case of SGX that's NumStackPages, NumHeapPages and NumTCS. This guarantees etc."

wintersteiger:
All done! Should I rebase it onto master too?

jhand2:
@wintersteiger there are no conflicts so it can go in as is. I just need @dthaler's approval and then I can merge.

jhand2:
Thanks everyone!

bors r+

PR #1301, opened 2019-01-08
Refactor dependencies on dlmalloc

mjp41:
I have been looking into using different allocators inside the enclave.  There are a few direct dependencies on dlmalloc in the OE code base that makes this difficult. The changes in this PR contain the use to dlmalloc to a single compilation unit, and refactor the code to cope with this.

* The PR introduces new symbols oe_internal_malloc, oe_internal_free, etc which are directly provided by dlmalloc (but potentially other allocators) and are used inside of directly calling dlmalloc routines. 
* The debug allocator is factor to call the new symbols

* A separate compilation unit is introduced for dlmalloc, so that the oe_malloc/oe_free code is separate from the code for compiling dlmalloc. 

* The PR also introduces some thread local stubs for allocators that use thread local (TCS) state.

* The PR also removes the dlmalloc.h, as nothing should depend on this, and dlmalloc is a single file. Including this would be taking a specific dependency, so I have removed it to stop that potentially happening.

Issue #2554, opened 2020-02-13
Remove td_t and use thread-local variables instead


anakrish:
[td_t ](https://github.com/openenclave/openenclave/blob/6b84df5855c190675a79095d958a9f42e14f1f6f/include/openenclave/internal/sgxtypes.h#L587) holds data specific to a thread. 

Eliminating td_t and using thread-local variables eliminates the need to maintain this special data structure and just using oe_thread_data_t (the first member of td_t). oe_thread_data_t can be made to confirm to MUSL/libC expectation and thus MUSL's pthread can ported more easily work within an enclave.

Initialization of thread-local variables must be ordered correctly to be able to remove td_t.

anakrish:
@nonpolarity Would you like to take on this interesting task?

nonpolarity:
@anakrish no problem, my pleasure. Please assign it to me.

anakrish:
Thanks!

PR #2649, opened 2020-03-10
Add support for a thread-aware allocator in OpenEnclave

achamayou:
As DC8 kit has become available in Azure, we have started scale-testing CCF (https://github.com/microsoft/CCF/) with higher numbers of threads. We have found to our disappointment that throughput decreased, because allocations and deallocations all contend a single lock in dlmalloc. OpenEnclave using dlmalloc (ie. the current code) is unable to effectively take advantage of multi-core hardware for workloads that involve dynamic allocation (ie. most general-purpose code).

We validated this result empirically on our "virtual" builds, by preloading dlmalloc and observing the same impact on throughput that we saw in the enclaves. The default system allocator, and other thread-aware allocators like jemalloc or snmalloc scaled as expected of course.

After adapting OpenEnclave to use snmalloc, we can run concurrent workloads in enclave without a bottleneck on memory allocation:

CCF SmallBank benchmark, 1m transactions, Standard_DC8 VM:

OpenEnclave with dlmalloc:

1 worker thread: 35k Tx/s
2 worker threads: 37k Tx/s
3 worker threads: 29k Tx/s
4 worker threads: 27k Tx/s

OpenEnclave with snmalloc:

1 worker thread: 39k Tx/s
2 worker threads: 77k Tx/s
3 worker threads: 110k Tx/s
4 worker threads: 115k Tx/s
5 worker threads: 143k Tx/s
6 worker threads: 156k Tx/s

There is still plenty of work to do improve performance in CCF, but removing this extreme contention on allocation helps quite a bit!

We have gone with snmalloc for the following reasons:

1. Its performance characteristics are excellent: https://github.com/microsoft/snmalloc/blob/master/snmalloc.pdf
2. It's actively maintained by Microsoft, extensively covered with tests and benchmarks, and registered with the MSRC, so we are confident any security issues will get addressed correctly.
3. It's easy to plug into OpenEnclave, thanks to its platform abstraction layer. It's so easy that it's been done twice before: #1301, #1420

I have not written a design document for this PR because it does not affect any public APIs, as far as I can tell.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -195,6 +195,14 @@ endif ()
 
 option(ADD_WINDOWS_ENCLAVE_TESTS "Build Windows enclave tests" OFF)
 
+# NOTE: Building OpenEnclave using snmalloc for memory allocation is an experimental option.
+# Longer term, there will be a pluggable allocator mechanism to allow users to choose from a
+# selection of allocators.
+option(USE_SNMALLOC "[EXPERIMENTAL] Build using snmalloc. If this is not set, dlmalloc is used." OFF)
+if (NOT USE_SNMALLOC)
+  set(USE_DLMALLOC true)
+endif ()
+
 find_program(VALGRIND "valgrind")
 if (VALGRIND)
   set(MEMORYCHECK_COMMAND_OPTIONS "--leak-check=full --error-exitcode=1")
--- a/enclave/core/CMakeLists.txt
+++ b/enclave/core/CMakeLists.txt
@@ -66,6 +66,18 @@ endif()
 set(PLATFORM_SRC "")
 set(MUSL_SRC_DIR ${PROJECT_SOURCE_DIR}/3rdparty/musl/musl/src)
 
+if (USE_SNMALLOC)
+    add_enclave_library(oeallocator OBJECT
+        sgx/snmalloc/snmalloc_wrapper.cpp
+    )
+    enclave_link_libraries(oeallocator PUBLIC oe_includes)
+    set_source_files_properties(sgx/snmalloc/snmalloc_wrapper.cpp PROPERTIES COMPILE_FLAGS -std=c++17\ -mcx16\ -fno-exceptions)
+    enclave_compile_options(oeallocator PUBLIC
+        -fPIC
+        -fvisibility=hidden)
+    install_enclaves(TARGETS oeallocator EXPORT openenclave-targets)
+endif ()
+
 if (OE_SGX)
     list(APPEND PLATFORM_SRC
         ../../common/sgx/endorsements.c
@@ -137,6 +149,14 @@ if (OE_TRUSTZONE OR (OE_SGX AND (UNIX OR USE_CLANGW)))
     list(APPEND PLATFORM_SRC init_fini.c)
 endif ()
 
+if (USE_DLMALLOC)
+    list(APPEND PLATFORM_SRC malloc.c)
+    # Unfortunately dlmalloc uses GNU extension that allows arithmetic
+    # null pointers.
+    set_source_files_properties(malloc.c
+        PROPERTIES COMPILE_FLAGS "-Wno-conversion -Wno-null-pointer-arithmetic")
+endif()
+
 add_enclave_library(oecore STATIC
     ../../common/safecrt.c
     ../../common/argv.c
@@ -159,7 +179,6 @@ add_enclave_library(oecore STATIC
     hexdump.c
     hostcalls.c
     intstr.c
-    malloc.c
     once.c
     printf.c
     pthread.c
@@ -199,6 +218,11 @@ if (OE_TRUSTZONE)
     enclave_link_libraries(oecore PUBLIC oelibutee_includes)
 endif ()
 
+enclave_link_libraries(oecore PUBLIC oe_includes)
+if (USE_SNMALLOC)
+    enclave_link_libraries(oecore PUBLIC oeallocator)
+endif()
+
 if (CMAKE_C_COMPILER_ID MATCHES GNU)
     enclave_compile_options(oecore PRIVATE -Wjump-misses-init)
 endif()
@@ -206,11 +230,6 @@ endif()
 # This directory contains enclave core libc headers:
 enclave_include_directories(oecore PRIVATE
     ${PROJECT_SOURCE_DIR}/include/openenclave/corelibc)
-
-# Unfortunately dlmalloc uses GNU extension that allows arithmetic
-# null pointers.
-set_source_files_properties(malloc.c
-    PROPERTIES COMPILE_FLAGS "-Wno-conversion -Wno-null-pointer-arithmetic")
 
 if (OE_SGX)
     set_source_files_properties(sgx/keys.c PROPERTIES COMPILE_FLAGS -Wno-type-limits)
--- a/enclave/core/atexit.c
+++ b/enclave/core/atexit.c
@@ -5,6 +5,7 @@
 #include <openenclave/enclave.h>
 #include <openenclave/internal/syscall/unistd.h>
 #include <openenclave/internal/thread.h>
+#include "oe_nodebug_alloc.h"
 
 /*
 **==============================================================================
@@ -34,7 +35,8 @@ static oe_spinlock_t _spin = OE_SPINLOCK_INITIALIZER;
 **
 ** _new_atexit_entry()
 **
-**     Allocate an oe_atexit_entry_t structure from the heap using sbrk().
+**     Allocate an oe_atexit_entry_t structure from the heap,
+**     using oe_nodebug_malloc().
 **
 **==============================================================================
 */
@@ -43,8 +45,8 @@ static oe_atexit_entry_t* _new_atexit_entry(void (*func)(void*), void* arg)
 {
     oe_atexit_entry_t* entry;
 
-    if ((entry = (oe_atexit_entry_t*)oe_sbrk(sizeof(oe_atexit_entry_t))) ==
-        (void*)-1)
+    if ((entry = (oe_atexit_entry_t*)oe_nodebug_malloc(
+             sizeof(oe_atexit_entry_t))) == (void*)-1)
         return NULL;
 
     entry->func = func;
--- a/enclave/core/malloc.c
+++ b/enclave/core/malloc.c
@@ -11,6 +11,8 @@
 #include <openenclave/internal/safecrt.h>
 #include <openenclave/internal/thread.h>
 #include "debugmalloc.h"
+#include "oe_alloc_thread.h"
+#include "oe_nodebug_alloc.h"
 
 /* The use of dlmalloc/malloc.c below requires stdc names from these headers */
 #define OE_NEED_STDC_NAMES
@@ -61,6 +63,23 @@ static int _dlmalloc_stats_fprintf(FILE* stream, const char* format, ...);
 #define FREE dlfree
 #define MALLOC_USABLE_SIZE dlmalloc_usable_size
 #endif
+
+void* oe_nodebug_malloc(size_t s)
+{
+    return dlmalloc(s);
+}
+void oe_nodebug_free(void* ptr)
+{
+    return dlfree(ptr);
+}
+void* oe_nodebug_realloc(void* ptr, size_t s)
+{
+    return dlrealloc(ptr, s);
+}
+void* oe_nodebug_memalign(size_t alignment, size_t size)
+{
+    return dlmemalign(alignment, size);
+}
 
 static oe_allocation_failure_callback_t _failure_callback;
 
@@ -247,4 +266,12 @@ oe_result_t oe_get_malloc_stats(oe_malloc_stats_t* stats)
 done:
     oe_mutex_unlock(&_mutex);
     return result;
+}
+
+void oe_alloc_thread_startup()
+{
+}
+
+void oe_alloc_thread_teardown()
+{
 }
--- a/enclave/core/oe_alloc_thread.h
+++ b/enclave/core/oe_alloc_thread.h
@@ -1,0 +1,17 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#ifndef _OE_ALLOC_THREAD_H
+#define _OE_ALLOC_THREAD_H
+
+#include <openenclave/bits/defs.h>
+
+OE_EXTERNC_BEGIN
+
+// Thread-specific startup and teardown calls for the memory allocator
+void oe_alloc_thread_startup();
+void oe_alloc_thread_teardown();
+
+OE_EXTERNC_END
+
+#endif /* _OE_ALLOC_THREAD_H */
--- a/enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
+++ b/enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
@@ -1,0 +1,119 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/internal/sgxtypes.h>
+#include "../../oe_alloc_thread.h"
+#include "../../oe_nodebug_alloc.h"
+
+namespace snmalloc
+{
+__thread void* allocator_local;
+
+class ThreadAllocUntyped
+{
+  public:
+    static void* get();
+};
+
+void* ThreadAllocUntyped::get()
+{
+    return allocator_local;
+}
+} // namespace snmalloc
+
+#define USE_RESERVE_MULTIPLE 1
+#define IS_ADDRESS_SPACE_CONSTRAINED
+
+// In standard configuration, snmalloc defines the oe_* symbols
+// directly. In debug configuration, snmalloc defines oe_nodebug_* symbols
+// instead, and oe_* symbols are directed to the debug shim.
+#if defined(OE_USE_DEBUG_MALLOC)
+#define SNMALLOC_NAME_MANGLE(a) oe_nodebug_##a
+
+extern "C" void* oe_malloc(size_t size)
+{
+    return oe_debug_malloc(size);
+}
+
+extern "C" void* oe_calloc(size_t nmemb, size_t size)
+{
+    return oe_debug_calloc(nmemb, size);
+}
+
+extern "C" void* oe_realloc(void* ptr, size_t size)
+{
+    return oe_debug_realloc(ptr, size);
+}
+
+extern "C" void* oe_memalign(size_t alignment, size_t size)
+{
+    return oe_debug_memalign(alignment, size);
+}
+
+extern "C" void* oe_free(void* ptr, size_t size)
+{
+    return oe_debug_free(ptr, size);
+}
+
+extern "C" size_t oe_malloc_usable_size(void* ptr)
+{
+    return oe_debug_malloc_usable_size(ptr);
+}
+#else
+#define SNMALLOC_NAME_MANGLE(a) oe_##a
+#endif
+// The Open Enclave runtime manages the thread local allocator pointer, see td_t
+#define SNMALLOC_EXTERNAL_THREAD_ALLOC
+// Enable the Open Enclave PAL(Platform Abstraction Layer) for Open Enclave,
+// see pal_open_enclave.h for details
+#define OPEN_ENCLAVE
+#include "../../../../3rdparty/snmalloc/src/override/malloc.cc"
+
+// A small number of places want to use the underlying allocator
+// without the debug shim on top, hence those explicit symbols.
+#ifndef OE_USE_DEBUG_MALLOC
+extern "C" void* oe_nodebug_malloc(size_t s)
+{
+    return oe_malloc(s);
+}
+extern "C" void oe_nodebug_free(void* ptr)
+{
+    return oe_free(ptr);
+}
+extern "C" void* oe_nodebug_realloc(void* ptr, size_t s)
+{
+    return oe_realloc(ptr, s);
+}
+extern "C" void* oe_nodebug_memalign(size_t alignment, size_t size)
+{
+    return oe_memalign(alignment, size);
+}
+#endif
+
+extern "C" void oe_alloc_thread_startup()
+{
+    allocator_local = current_alloc_pool()->acquire();
+}
+
+extern "C" void oe_alloc_thread_teardown()
+{
+    current_alloc_pool()->release(ThreadAlloc::get());
+    allocator_local = nullptr;
+}
+
+typedef void (*oe_allocation_failure_callback_t)(
+    const char* file,
+    size_t line,
+    const char* func,
+    size_t size);
+
+extern "C" void oe_set_allocation_failure_callback(
+    oe_allocation_failure_callback_t function)
+{
+    (void)function;
+}
+
+extern "C" void oe_memalign_free(void* ptr)
+{
+    oe_free(ptr);
+}
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,3 +6,6 @@
 	path = 3rdparty/optee/optee_client
 	url = https://github.com/ms-iot/optee_client
 	branch = ms-iot-openenclave
+[submodule "3rdparty/snmalloc"]
+	path = 3rdparty/snmalloc
+	url = https://github.com/Microsoft/snmalloc.git
--- a/3rdparty/libunwind/stubs.h
+++ b/3rdparty/libunwind/stubs.h
@@ -90,7 +90,7 @@ static __inline void* __libunwind_mmap(
     int fd,
     off_t offset)
 {
-    void* dlmemalign(size_t alignment, size_t size);
+    void* oe_nodebug_memalign(size_t alignment, size_t size);
     void* result = MAP_FAILED;
 
     if (addr || fd != -1 || offset)
@@ -102,7 +102,7 @@ static __inline void* __libunwind_mmap(
     if (flags != (MAP_PRIVATE | MAP_ANONYMOUS))
         goto done;
 
-    result = dlmemalign(4096, length);
+    result = oe_nodebug_memalign(4096, length);
 
 done:
 
@@ -111,13 +111,13 @@ done:
 
 static __inline int __libunwind_munmap(void* addr, size_t length)
 {
-    extern void dlfree(void* ptr);
+    extern void oe_nodebug_free(void* ptr);
 
     if (!addr)
         return -1;
 
     if (length)
-        dlfree(addr);
+        oe_nodebug_free(addr);
 
     return 0;
 }
--- a/enclave/core/debugmalloc.c
+++ b/enclave/core/debugmalloc.c
@@ -1,7 +1,6 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
-#define USE_DL_PREFIX
 #include "debugmalloc.h"
 #include <openenclave/corelibc/errno.h>
 #include <openenclave/enclave.h>
@@ -13,9 +12,10 @@
 #include <openenclave/internal/thread.h>
 #include <openenclave/internal/types.h>
 #include <openenclave/internal/utils.h>
-#include "../3rdparty/dlmalloc/dlmalloc/malloc.h"
 
 #if defined(OE_USE_DEBUG_MALLOC)
+
+#include "oe_nodebug_alloc.h"
 
 /*
 **==============================================================================
@@ -312,7 +312,7 @@ void* oe_debug_malloc(size_t size)
     void* block;
     const size_t block_size = _calculate_block_size(0, size);
 
-    if (!(block = dlmalloc(block_size)))
+    if (!(block = oe_nodebug_malloc(block_size)))
         return NULL;
 
     /* Fill block with 0xAA (Allocated) bytes */
@@ -339,7 +339,7 @@ void oe_debug_free(void* ptr)
         size_t block_size = _get_block_size(ptr);
         oe_memset_s(block, block_size, 0xDD, block_size);
 
-        dlfree(block);
+        oe_nodebug_free(block);
     }
 }
 
@@ -404,7 +404,7 @@ void* oe_debug_memalign(size_t alignment, size_t size)
     void* block;
     header_t* header;
 
-    if (!(block = dlmemalign(alignment, block_size)))
+    if (!(block = oe_nodebug_memalign(alignment, block_size)))
         return NULL;
 
     header = (header_t*)((uint8_t*)block + padding_size);
--- a/enclave/core/oe_nodebug_alloc.h
+++ b/enclave/core/oe_nodebug_alloc.h
@@ -1,0 +1,21 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#ifndef _OE_NODEBUG_ALLOC_H
+#define _OE_NODEBUG_ALLOC_H
+
+#include <openenclave/bits/defs.h>
+#include <stddef.h>
+
+OE_EXTERNC_BEGIN
+
+// A small number of places want to use the underlying allocator
+// without the debug shim on top, hence those explicit symbols.
+void* oe_nodebug_malloc(size_t s);
+void oe_nodebug_free(void* ptr);
+void* oe_nodebug_realloc(void* ptr, size_t s);
+void* oe_nodebug_memalign(size_t alignment, size_t size);
+
+OE_EXTERNC_END
+
+#endif /* _OE_NODEBUG_ALLOC_H */
--- a/enclave/core/sgx/backtrace.c
+++ b/enclave/core/sgx/backtrace.c
@@ -11,6 +11,7 @@
 #include <openenclave/internal/print.h>
 #include <openenclave/internal/raise.h>
 #include <openenclave/internal/safecrt.h>
+#include "../oe_nodebug_alloc.h"
 #include "sgx_t.h"
 #include "tee_t.h"
 
@@ -183,16 +184,12 @@ done:
 char** oe_backtrace_symbols(void* const* buffer, int size)
 {
     /* Backtrace must use the internal allocator to bypass debug-malloc. */
-    extern void* dlmalloc(size_t size);
-    extern void* dlrealloc(void* ptr, size_t size);
-    extern void dlfree(void* ptr);
-    return oe_backtrace_symbols_impl(buffer, size, dlmalloc, dlrealloc, dlfree);
+    return oe_backtrace_symbols_impl(
+        buffer, size, oe_nodebug_malloc, oe_nodebug_realloc, oe_nodebug_free);
 }
 
 void oe_backtrace_symbols_free(char** ptr)
 {
     /* Backtrace must use the internal allocator to bypass debug-malloc. */
-    extern void dlfree(void* ptr);
-
-    dlfree(ptr);
+    oe_nodebug_free(ptr);
 }
--- a/enclave/core/sgx/linux/threadlocal.c
+++ b/enclave/core/sgx/linux/threadlocal.c
@@ -9,6 +9,7 @@
 #include <openenclave/internal/raise.h>
 #include <openenclave/internal/safecrt.h>
 #include <openenclave/internal/utils.h>
+#include "../../oe_alloc_thread.h"
 #include "../td.h"
 
 /*
@@ -330,6 +331,9 @@ oe_result_t oe_thread_local_init(td_t* td)
 
             _thread_locals_relocated = true;
         }
+
+        // Must occur after thread local storage initialization
+        oe_alloc_thread_startup();
     }
 
     result = OE_OK;
@@ -381,6 +385,7 @@ oe_result_t oe_thread_local_cleanup(td_t* td)
     uint8_t* tls_start = _get_thread_local_data_start(td);
     if (tls_start)
     {
+        oe_alloc_thread_teardown();
         oe_memset_s(tls_start, (uint64_t)(fs - tls_start), 0, 0);
     }
 
--- a/libc/libunwind_stubs.c
+++ b/libc/libunwind_stubs.c
@@ -2,14 +2,12 @@
 // Licensed under the MIT License.
 
 #define _GNU_SOURCE
-#define USE_DL_PREFIX
 #include <assert.h>
 #include <openenclave/enclave.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
 #include <unistd.h>
-#include "../3rdparty/dlmalloc/dlmalloc/malloc.h"
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wmissing-prototypes"
--- a/scripts/.check-license.ignore
+++ b/scripts/.check-license.ignore
@@ -45,6 +45,7 @@ tools/oeedger8r/intel/.*
 3rdparty/musl/endian\.h\.suffix
 3rdparty/optee/optee_client
 3rdparty/optee/optee_os
+3rdparty/snmalloc
 cmake/NuGetDescription.txt
 docs/CODEOWNERS
 LICENSE
--- a/tests/bigmalloc/enc/enc.c
+++ b/tests/bigmalloc/enc/enc.c
@@ -12,8 +12,8 @@ oe_result_t test_malloc()
     const size_t GIGABYTE = 1024 * 1024 * 1024;
     size_t heap_remaining;
     uint8_t* ptr = NULL;
-    extern void* dlmalloc(size_t n);
-    extern void dlfree(void* ptr);
+    extern void* oe_malloc(size_t n);
+    extern void oe_free(void* ptr);
     oe_result_t return_value = OE_UNEXPECTED;
 
     /* Determine how much heap memory remains */
@@ -43,7 +43,7 @@ oe_result_t test_malloc()
     {
         const size_t allocation_size = (size_t)(0.99 * (double)heap_remaining);
 
-        if (!(ptr = (uint8_t*)dlmalloc(allocation_size)))
+        if (!(ptr = (uint8_t*)oe_malloc(allocation_size)))
         {
             return_value = OE_OUT_OF_MEMORY;
             goto done;
@@ -59,7 +59,7 @@ oe_result_t test_malloc()
 done:
 
     if (ptr)
-        dlfree(ptr);
+        oe_free(ptr);
 
     return return_value;
 }


jhand2:
A few initial thoughts:

1. Friendly reminder to clean up the commit history. The history should be free of "merge" commits and each commit should be targeted with a detailed description of what it does.
2. Is the recommendation to replace dlmalloc or support both via build configuration?
3. Most of our 3rd party dependencies are included as normal code rather than git submodules. I don't know why this is, but we should discuss with others to make sure we take whatever those reasons were into account

I'll try to review the substance of the PR this afternoon. I will bring up this PR in the triage meeting today if you are planning to join.

Thanks!

radhikaj:
We should support debug allocation. Recommend getting snmalloc working for debug allocations. we would like to enable snmalloc and in CI in mainline for a few weeks to verify that there are no stability issues. Then get rid of dlmalloc and replace with snmalloc.

achamayou:
@jhand2 I will squash and sign off, my initial purpose is to get this out for review. As discussed, short-term, the goal is to provide an option that unblocks the situation for multi-threaded code, with a view to eventually consolidating once there's consensus.

@radhikaj I'll have a look at getting the debug allocator to work with snmalloc.

@anakrish oeallocator is built without exceptions (see enclave/core/CMakeLists.txt), so I don't think it should cause material changes to oecore. snmalloc does not thread local storage, and the way it's hooked up to td_t is I believe the only way to make sure it's initialised there early enough. If there's an alternative solution that's better in some ways, I'm happy to look at it of course.

mjp41:
> Does snmalloc use thread-local variables? If so, it might make sense to use a thread-local variable to hold to to the thread's allocator instead of adding a member to td_t.

To do this, would include more C++ features, so would be harder to support inside oecore. For the current implementation for OE has no dependency on a C++ library at runtime. This was the main reason for modifying `td_t`. Outside the enclave, we have used C++ thread_static variables, but this requires runtime links to 

* `__cxa_thread_atexit`
* `__cxa_guard_acquire`
* `__cxa_guard_release`

This is not something you probably want to support in oecore.

enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
```diff
@@ -19,11 +19,72 @@ void* ThreadAllocUntyped::get()
 
 #define USE_RESERVE_MULTIPLE 1
 #define IS_ADDRESS_SPACE_CONSTRAINED
+
+// In standard configuration, snmalloc defines the oe_* symbols
+// directly. In debug configuration, snmalloc defines _oe_* symbols
+// instead, and oe_* symbols are directed to the debug shim instead.
+#if defined(OE_USE_DEBUG_MALLOC)
+#define SNMALLOC_NAME_MANGLE(a) _oe_##a
```
mjp41: Couldn't this line be:
```suggestion
#define SNMALLOC_NAME_MANGLE(a) oe_nodebug_##a
```

mjp41: Then you can remove the later definitions in this case.

achamayou: Absolutely, thank you!

achamayou:
bors try

CodeMonkeyLeet:
@mjp41 @achamayou @anakrish 
>> Does snmalloc use thread-local variables? If so, it might make sense to use a thread-local variable to hold to to the thread's allocator instead of adding a member to td_t.
>
> To do this, would include more C++ features, so would be harder to support inside oecore. For the current implementation for OE has no dependency on a C++ library at runtime. This was the main reason for modifying `td_t`. 

Does this apply equally to C11 `thread_local`/gcc `__thread`? There is already work in OE SDK to support that and there's also a `__cxa_thread_atexit` implemention in OE though not the guards.

anakrish:
> @mjp41 @achamayou @anakrish
> 
> > > Does snmalloc use thread-local variables? If so, it might make sense to use a thread-local variable to hold to to the thread's allocator instead of adding a member to td_t.
> > 
> > 
> > To do this, would include more C++ features, so would be harder to support inside oecore. For the current implementation for OE has no dependency on a C++ library at runtime. This was the main reason for modifying `td_t`.
> 
> Does this apply equally to C11 `thread_local`/gcc `__thread`? There is already work in OE SDK to support that and there's also a `__cxa_thread_atexit` implemention in OE though not the guards.



> > Does snmalloc use thread-local variables? If so, it might make sense to use a thread-local variable to hold to to the thread's allocator instead of adding a member to td_t.
> 
> To do this, would include more C++ features, so would be harder to support inside oecore. For the current implementation for OE has no dependency on a C++ library at runtime. This was the main reason for modifying `td_t`. Outside the enclave, we have used C++ thread_static variables, but this requires runtime links to
> 
>     * `__cxa_thread_atexit`
> 
>     * `__cxa_guard_acquire`
> 
>     * `__cxa_guard_release`
> 
> 
> This is not something you probably want to support in oecore.

I don't think C thread-locals (via the GNU __thread) or even POD C++ thread-locals would require guards. C++ thread-locals with non-trivial constructors and destructors would.

Note, I am not saying that we ought to use thread-locals, but am curious to know the rationale behind the decisions in this PR.

anakrish:
@achamayou 
With this PR, the choice of allocator seems to be made when the OE SDK itself is built. One the SDK is built with dlmalloc, it is not possible for an SDK user to then use snmalloc without rebuilding the SDK.

Can you see if it is possible to delay the decision to the SDK use time?

Say if the user links against `oe-snmalloc` libray in addition to `oe-enclave`, then snmalloc is used.
Otherwise, the default allocator (which happens to be dlmalloc for now) is used.

jhand2:
> @achamayou
> With this PR, the choice of allocator seems to be made when the OE SDK itself is built. One the SDK is built with dlmalloc, it is not possible for an SDK user to then use snmalloc without rebuilding the SDK.
> 
> Can you see if it is possible to delay the decision to the SDK use time?
> 
> Say if the user links against `oe-snmalloc` libray in addition to `oe-enclave`, then snmalloc is used.
> Otherwise, the default allocator (which happens to be dlmalloc for now) is used.

I would actually vote against this.

If the plan is to replace dlmalloc with snmalloc anyway, is there really a good reason to make this decision at runtime? I also think it complicates the allocator for a use-case that users/use-cases have asked for.

mjp41:
@anakrish @CodeMonkeyLeet 

> C++ thread-locals with non-trivial constructors and destructors would.

We have non-trivial destructors to handle returning the thread local allocator state to the global pool of allocators. This leads to the requirement on: `__cxa_thread_atexit`.

On other platforms (i.e. not OE), the thread local state for the allocator is trivially initialised to a `GlobalPlaceHolder`, which is a single global instance of the thread local allocator.  This unfortunately has a non-trivial constructor.  It is set up in a way, that guarantees any allocation request will hit a slow path (all structures are valid, but empty) , where we can then check if we have an actual thread local allocator, or this dummy one.  This means we can move a conditional branch off the fast allocation and deallocation paths.  This leads to the requirement on: `__cxa_guard_acquire` and  `__cxa_guard_release`.

On OE, because we have access to the thread startup and teardown point, we can avoid all these hoops and not have to worry about allocator initialisation. It is common for allocators tightly integrated into a runtime to take advantage in this way.

Hope this explanation helps.  I can make the allocator slower if you insist.

mjp41:
>Are there significant benefits over TCMalloc? Intel integrates it in their SDK and Google maintains a version of it, so it looks like it has greater traction?

TCMalloc has some use cases where its performance drops off considerably.  A particularly bad scenario is where a lot of allocations on one thread are deallocated on a different thread.  If you have a dedicated thread taking requests into the enclave, and a pool of threads handling processing and deallocating the requests, then you will hit the bad cases for TCMalloc involving a lot of lock contention.

This flow of data between threads was one of the primary design goals of snmalloc, and, we believe, will be a requirement for how to write multi-threaded software inside an enclave.

With snmalloc, we also have a well defined Platform Abstraction Layer, this makes it easier to port to new platforms.

achamayou:
> Nice to see a more performant handler implementation being contributed! For context, could you comment briefly on:
> 
> * Are there significant benefits over TCMalloc? Intel integrates it in their SDK and Google maintains a version of it, so it looks like it has greater traction?
> * What's the roadmap of snmalloc? Is it actively used in production elsewhere?

I'll let @mjp41 address the second point, I alluded to the first point in the PR description.

In a nutshell, snmalloc offers better performance (details in the paper), and looks healthier from a support perspective. It is registered with the Microsoft Security Response Center, and actively supported by Microsoft. tcmalloc is not officially supported by Google as far as I can tell:

https://google.github.io/tcmalloc/

> Disclaimer: This is not an officially supported Google product.

Intel does include a copy for optional use with their SDK, but switching isn't particularly straightforward nor user-friendly (https://github.com/intel/linux-sgx/tree/master/sdk/gperftools), and they have yet to make it default, so I don't know that it has significant traction. I certainly couldn't find any evidence of it.

achamayou:
@anakrish @jhand2 the existing situation with the debug allocator is also that it has to be selected at build time, this change is consistent with precedent.
If we think allocator switching should be available without re-building OE, I suggest this should be discussed as a separate new feature.

3rdparty/dlmalloc/dlmalloc/malloc.h
```diff
@@ -1,620 +0,0 @@
-/*
```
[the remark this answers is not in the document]
achamayou: Sure, I'll put it back.

enclave/core/malloc.c
```diff
@@ -248,3 +265,11 @@ oe_result_t oe_get_malloc_stats(oe_malloc_stats_t* stats)
     oe_mutex_unlock(&_mutex);
     return result;
 }
+
+void _oe_alloc_thread_startup()
```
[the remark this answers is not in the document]
achamayou: Will rename.

enclave/core/malloc.c
```diff
@@ -248,3 +265,11 @@ oe_result_t oe_get_malloc_stats(oe_malloc_stats_t* stats)
...
+{
+}
+
+void _oe_alloc_thread_teardown()
+{
+}
```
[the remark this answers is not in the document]
achamayou: Will add. If this important, I suggest changing the clang-format config to enforce it.

enclave/core/sgx/td_basic.c
```diff
@@ -18,6 +18,9 @@
 #include "linux/threadlocal.h"
 #endif
 
+void _oe_alloc_thread_teardown();
```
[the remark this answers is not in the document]
achamayou: There's just two, and they're only in two places now (malloc.c and snmalloc_wrapper.cpp), and a probably a single place eventually, so that's why I didn't do it. But I'm happy to do it.

CMakeLists.txt
```diff
@@ -195,6 +195,16 @@ endif ()
...
 
+option(USE_SNMALLOC "Build using snmalloc" OFF)
+if (USE_SNMALLOC)
+  add_definitions(-DOE_USE_SNMALLOC)
+else ()
+  set(USE_DLMALLOC true)
+endif ()
+
+# Warning: turning on simulation mode on Windows may cause test failures and random crashes
+option(WIN32_SIMULATION "Windows Simulation Mode" OFF)
```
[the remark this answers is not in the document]
achamayou: Sorry, I think this is a rebase accident, it shouldn't be here.

mjp41:
> >What's the roadmap of snmalloc? Is it actively used in production elsewhere?

> I'll let @mjp41 address the second point, I alluded to the first point in the PR description.

So we don't have a public roadmap.  We have been doing research around using the allocator to enable new security features.  In fact, @swamymsft was helping us with prototyping before he joined your team.  Happy to chat on Teams as we want to keep these features confidential for now.

include/openenclave/bits/properties.h
```diff
@@ -47,6 +47,18 @@ typedef struct _oe_enclave_properties_header
  * @endcond
  */
 
+// SNMALLOC Needs lots of address space.
+#ifdef OE_USE_SNMALLOC
+#define HMAX(a, b) ((a) < (b) ? (b) : (a))
+#define OE_HEAP_SIZE(HEAP_PAGE_COUNT, TCS_COUNT) \
+    (HMAX(HEAP_PAGE_COUNT, ((1ULL << 8) * (TCS_COUNT * 3 + 4))))
```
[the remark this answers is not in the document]
mjp41: `snmalloc` uses 1MiB chunks of memory for the heap space.  It requests a reasonable amount at each point to ensure it doesn't waste too much space for aligning memory.
```C++
   1ULL << 8
```
Is the 1MiB in pages.  Each thread will require a 1MiB chunk.  The `3` and `4` are used to deal with the cases where we are particularly unlucky in alignment requests.

This is from when I originally ported `snmalloc` to oe over a year ago.  At the time several tests failed due to starting enclaves with very small heap sizes. This was a way to get remove those failures.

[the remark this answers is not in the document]
achamayou: I don't think we need this anymore, so I'll remove it. In a nutshell, snmalloc wants 1Mb aligned per thread to start with, and this was trying to get it that.

enclave/core/sgx/td_basic.c
```diff
@@ -111,6 +114,8 @@ void td_init(td_t* td)
         /* List of callsites is initially empty */
         td->callsites = NULL;
 
+        _oe_alloc_thread_startup();
```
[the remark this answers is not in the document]
achamayou: That doesn't seem ideal, if a thread doesn't need dynamic allocation, we would cause it to crash for no reason.

include/openenclave/bits/properties.h
```diff
@@ -47,6 +47,18 @@ typedef struct _oe_enclave_properties_header
  * @endcond
  */
 
+// SNMALLOC Needs lots of address space.
+#ifdef OE_USE_SNMALLOC
+#define HMAX(a, b) ((a) < (b) ? (b) : (a))
+#define OE_HEAP_SIZE(HEAP_PAGE_COUNT, TCS_COUNT) \
```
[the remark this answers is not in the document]
achamayou: This is a leftover from previous integrations, I don't think we want it here. It's to do with making sure there's enough heap space for snmalloc to do its initial allocation, but I agree that this isn't currently the right way to go about it.

enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
```diff
@@ -0,0 +1,112 @@
...
+}
+
+extern "C" size_t oe_malloc_usable_size(void* ptr)
+{
+    return oe_debug_malloc_usable_size(ptr);
+}
+#else
+#define SNMALLOC_NAME_MANGLE(a) oe_##a
+#endif
+#define SNMALLOC_EXTERNAL_THREAD_ALLOC
```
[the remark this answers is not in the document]
achamayou: Done

achamayou:
bors try

achamayou:
@radhikaj @jhand2 @mikbras the debug allocator now works with snmalloc too.

achamayou:
bors try

jhand2:
@CodeMonkeyLeet @anakrish @radhikaj @mingweishih Does anyone have an opinion on using a git submodule to include snmalloc?

I'm not opposed to it, but I think it's confusing if some dependencies are brought in one way and some are brought in another? If we merge this as a submodule, I think our long-term goal should be to move all our 3rd party dependencies to submodules.

CodeMonkeyLeet:
> @CodeMonkeyLeet @anakrish @radhikaj @mingweishih Does anyone have an opinion on using a git submodule to include snmalloc? 

I'm not against it; we have precedence with the OP-TEE projects already. Specifically for this set of changes, we should consider revisiting all of the contributor getting started docs to ensure that they specify the need to `git submodule init` and `update`, since this is the first mainline submodule that affects the SGX setup (at least until the OP-TEE setup docs are more closely integrated with the SGX ones). 

Regarding moving all 3rdparty dependencies to submodules, I'm not against that either, although there were supply chain security concerns about that previously. I would suggest that we mitigate that via verified build environments published in container images rather than direct source ingestion moving forwards.

mingweishih:
I also think moving all 3rd party dependencies to submodules is the right direction to take. However, I think we need to consider whether we want to always pull the lastest code or stable versions (then we will update the pointers as the way we update 3rdparty code base now).

enclave/core/CMakeLists.txt
```diff
@@ -66,6 +66,18 @@ endif()
 set(PLATFORM_SRC "")
 set(MUSL_SRC_DIR ${PROJECT_SOURCE_DIR}/3rdparty/musl/musl/src)
 
+if (USE_SNMALLOC)
+    add_library(oeallocator OBJECT
```
mingweishih: Per recent changes, please use corresponding wrappers instead of built-in cmake functions. For example, using `add_enclave_library` instead of `add_library`. See https://github.com/openenclave/openenclave/blob/master/cmake/enclave_cmake_wrappers.cmake for more details.

enclave/core/CMakeLists.txt
```diff
@@ -66,6 +66,18 @@ endif()
...
+        sgx/snmalloc/snmalloc_wrapper.cpp
+    )
+    target_link_libraries(oeallocator PUBLIC oe_includes)
```
mingweishih: Ditto

enclave/core/CMakeLists.txt
```diff
@@ -66,6 +66,18 @@ endif()
...
+    set_source_files_properties(sgx/snmalloc/snmalloc_wrapper.cpp PROPERTIES COMPILE_FLAGS -std=c++17\ -mcx16\ -fno-exceptions)
+    target_compile_options(oeallocator PUBLIC
```
mingweishih: Ditto

enclave/core/CMakeLists.txt
```diff
@@ -66,6 +66,18 @@ endif()
...
+        -fPIC
+        -fvisibility=hidden)
+    install (TARGETS oeallocator EXPORT openenclave-targets)
```
mingweishih: Ditto

jhand2:
> I also think moving all 3rd party dependencies to submodules is the right direction to take. However, I think we need to consider whether we want to always pull the lastest code or stable versions (then we will update the pointers as the way we update 3rdparty code base now).

I would prefer the latter. Always pulling the latest code is undesirable from a security/stability/reproducability perspective.

CodeMonkeyLeet:
@mingweishih 
> I also think moving all 3rd party dependencies to submodules is the right direction to take. However, I think we need to consider whether we want to always pull the lastest code or stable versions (then we will update the pointers as the way we update 3rdparty code base now).

It should always take a dependency on a specific tag/commit on the submodule. My understanding is that git submodule natively supports that already, but we should be explicit about this workflow.

mingweishih:
Then, the question for this PR is that is there a stable branch of the snmalloc that submodules can point to (instead of always pulling from master). Sync with master should not be the default behavior.

anakrish:
> @CodeMonkeyLeet @anakrish @radhikaj @mingweishih Does anyone have an opinion on using a git submodule to include snmalloc?
> 
> I'm not opposed to it, but I think it's confusing if some dependencies are brought in one way and some are brought in another? If we merge this as a submodule, I think our long-term goal should be to move all our 3rd party dependencies to submodules.

I am not sure what the best practice here is. I would be fine with using submodules or using the current strategy of copying the 3rdparty sources locally or using a fork of the 3rdparty repo.
@andschwa Can you chime in?
I thought @paulcallen was driving this discussion some time ago.

anakrish:
> @anakrish @jhand2 the existing situation with the debug allocator is also that it has to be selected at build time, this change is consistent with precedent.
> If we think allocator switching should be available without re-building OE, I suggest this should be discussed as a separate new feature.

I am looking at this from an OE SDK user point of view.
SDK users are not expected to build the SDK (thought they may choose to); SDK contributors are expected to build the SDK to propose patches/PRs.

If the allocator choice is made at SDK build time then
- is snmalloc intended to be used only by an SDK contributor?
- if it is intended to be used by an SDK contributor, then what is the benefit we get out of adding it to the oerepo?

Keep in mind that we are creating a precedent here of adding functionality to OE SDK that with this PR serves mainly one advanced SDK contributor/user (CCF).

Don't get me wrong. I am not opposed to snmalloc or make OE more optimized.

I am concerned with the usability and how this functionality reaches the SDK user.

Have we thought of other alternatives like
- Switching over completely to snmalloc?
- Or provide both the user and let them choose at runtime?

anakrish:
> @anakrish @jhand2 the existing situation with the debug allocator is also that it has to be selected at build time, this change is consistent with precedent.
> If we think allocator switching should be available without re-building OE, I suggest this should be discussed as a separate new feature.

Also something to consider, that this is a good feature with potential benefits. 
It would be good to adhere to the [Contributor Guidelines](https://github.com/openenclave/openenclave/blob/master/docs/Contributing.md), and add tests that lock down the use of snmalloc.

anakrish:
@mjp41 
>
> We have non-trivial destructors to handle returning the thread local allocator state to the global pool of allocators. This leads to the requirement on: `__cxa_thread_atexit`.
>

Is this the case with snmalloc use in OE? Or the non OE case?

mjp41:
> >We have non-trivial destructors to handle returning the thread local allocator state to the global pool of allocators. This leads to the requirement on: __cxa_thread_atexit.

> Is this the case with snmalloc use in OE? Or the non OE case?

Non-oe case.

mjp41:
@anakrish , @CodeMonkeyLeet 

Re tag/release: for snmalloc.  We have been planning to make a release for snmalloc, but I have been holding off while the integration work with OE has been ongoing, so that I didn't have to do two releases in rapid succession.

My proposal would be 
1. You take this PR 
2. Run various tests and get your main partners to test this option.
3. If any issues arise, we fix them, and repeat from 2.
4. Once main partners happy this is stable feature,
  * switch snmalloc to default from dlmalloc
  * We'll make a formal release for snmalloc, which is what you will be using
5. After a couple of months, if still no issues remove dlmalloc.

Doing all this in one go seems risky, and doing nothing is unacceptable for performance of any multi-threaded workload in the enclave.

achamayou:
@CodeMonkeyLeet @jhand2 @mingweishih @jhand2 I've used a sub-module because I saw recent 3rdparty optee additions were done that way. If a full copy import is preferable, I'll do that of course.

The submodule as I've added it is absolutely pointing to a single commit, and requires a commit (and therefore a PR) to be changed.

@mingweishih thank you for pointing out the CMake changes, I will apply them.

@anakrish I think you may have missed the discussion that happened in the OE triage meeting. The intention is for this to be optionally enabled for evaluation, and to unblock users who write multi-threaded and performance-sensitive code, before replacing dlmalloc permanently. This isn't a CCF-specific improvement as you suggest, any application with multiple threads that perform memory allocation will benefit.

enclave/core/CMakeLists.txt
achamayou: Thank you for pointing this out. Fixed.

enclave/core/CMakeLists.txt
achamayou: Fixed.

enclave/core/CMakeLists.txt
achamayou: Fixed.

enclave/core/CMakeLists.txt
achamayou: Fixed.

dthaler:
> @CodeMonkeyLeet @anakrish @radhikaj @mingweishih Does anyone have an opinion on using a git submodule to include snmalloc?
> I'm not opposed to it, but I think it's confusing if some dependencies are brought in one way and some are brought in another? If we merge this as a submodule, I think our long-term goal should be to move all our 3rd party dependencies to submodules.

Definitely.  We previously discussed this in an OE design meeting and the long term goal is as Jordan says above: use submodules.

achamayou:
bors try

CMakeLists.txt
```diff
@@ -193,7 +193,13 @@ else ()
   option(USE_DEBUG_MALLOC "Build oeenclave with memory leak detection capability." OFF)
 endif ()
 
-option(ADD_WINDOWS_ENCLAVE_TESTS "Build Windows enclave tests" OFF)
+option(USE_SNMALLOC "Build using snmalloc. If this is not set, dlmalloc is used." OFF)
+if (NOT USE_SNMALLOC)
+  set(USE_DLMALLOC true)
+endif ()
+
+# Warning: turning on simulation mode on Windows may cause test failures and random crashes
```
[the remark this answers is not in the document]
achamayou: Sorry, this is still the same rebase accident that Simon had rightly pointed out, and for which I have not pushed the commit.

include/openenclave/bits/properties.h
```diff
@@ -47,6 +47,9 @@ typedef struct _oe_enclave_properties_header
  * @endcond
  */
 
+// This macro initializes and injects an oe_sgx_enclave_properties_t struct
```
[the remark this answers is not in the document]
achamayou: Same as above.

radhikaj:
@achamayou 
Are you blocked on making progress till this PR goes in?
Do you need it to go in now?
dlmalloc is preferable in certain scenarios such as Op-TEE. It is also versioned. The ideal direction for the SDK is to make it possible to plug in allocators. That way, developers can choose the best allocator for their specific scenarios.
Would you be able to help us move in this direction?

achamayou:
@radhikaj yes, we are unfortunately blocked. We cannot make progress on scaling CCF to more than one core and this is impacting deliverables. I have included some results from our performance testing at the beginning of this PR. We have known for a long time this would be a problem, hence the earlier PRs at the beginning of last year (#1301 and #1420).

This does not impact Op-TEE in any way, it neither removes dlmalloc nor changes the fact that it is the default allocator. If you make Op-TEE kit available, I'm sure @mjp41 will be happy to look at what it takes to support it. There's ongoing work on ARM support in snmalloc at the moment: https://github.com/microsoft/snmalloc/pull/142.

I am sure it would be nicer to have pluggable allocators, but I am afraid I can't spend any more time on this, I have a lot of work on my own project. I would like to insist again on the fact that any multi-threaded application is going to be seriously crippled by the current allocator situation. That may not be a concern for Op-TEE at the moment, but for SGX in cloud scenarios, it is a major problem.

achamayou:
@radhikaj I think the notion dlmalloc is versioned is questionable at best, this is the process through which it is imported. Unversioned and unauthenticated FTP downloads:

```
update-dlmalloc:
	rm -rf dlmalloc
	mkdir -p dlmalloc
	( cd dlmalloc; wget ftp://gee.cs.oswego.edu/pub/misc/malloc.h )
	( cd dlmalloc; wget ftp://gee.cs.oswego.edu/pub/misc/malloc.c )
```

There's certainly no way to derive the version from those links, and the code itself isn't all that precise either:

```
/*
  Default header file for malloc-2.8.x, written by Doug Lea
  and released to the public domain, as explained at
...
```

The documentation page was last updated in 2000: http://gee.cs.oswego.edu/dl/html/malloc.html, contains no release notes, no suggestion of which 2.8.x this is exactly nor how it differs from previous versions.

jhand2:
> @radhikaj yes, we are unfortunately blocked. We cannot make progress on scaling CCF to more than one core and this is impacting deliverables. I have included some results from our performance testing at the beginning of this PR. We have known for a long time this would be a problem, hence the earlier PRs at the beginning of last year (#1301 and #1420).
> 
> This does not impact Op-TEE in any way, it neither removes dlmalloc nor changes the fact that it is the default allocator. If you make Op-TEE kit available, I'm sure @mjp41 will be happy to look at what it takes to support it. There's ongoing work on ARM support in snmalloc at the moment: [microsoft/snmalloc#142](https://github.com/microsoft/snmalloc/pull/142).

I think Radhika's point was that OP-TEE is single threaded and wouldn't benefit.

> 
> I am sure it would be nicer to have pluggable allocators, but I am afraid I can't spend any more time on this, I have a lot of work on my own project. I would like to insist again on the fact that any multi-threaded application is going to be seriously crippled by the current allocator situation. That may not be a concern for Op-TEE at the moment, but for SGX in cloud scenarios, it is a major problem.

@anakrish did some work yesterday (Thanks Anand!) to prototype an interface for pluggable allocators: https://github.com/openenclave/openenclave/compare/master...anakrish:ak_allocator

I think we should seriously consider this path since as you've pointed out different allocators have different benefits and not all users of the SDK have the same needs.

jhand2:
> @radhikaj I think the notion dlmalloc is versioned is questionable at best, this is the process through which it is imported. Unversioned and unauthenticated FTP downloads:
> 
> ```
> update-dlmalloc:
> 	rm -rf dlmalloc
> 	mkdir -p dlmalloc
> 	( cd dlmalloc; wget ftp://gee.cs.oswego.edu/pub/misc/malloc.h )
> 	( cd dlmalloc; wget ftp://gee.cs.oswego.edu/pub/misc/malloc.c )
> ```
> 
> There's certainly no way to derive the version from those links, and the code itself isn't all that precise either:
> 
> ```
> /*
>   Default header file for malloc-2.8.x, written by Doug Lea
>   and released to the public domain, as explained at
> ...
> ```
> 
> The documentation page was last updated in 2000: http://gee.cs.oswego.edu/dl/html/malloc.html, contains no release notes, no suggestion of which 2.8.x this is exactly nor how it differs from previous versions.

This is more of a problem in that OE has done a poor job defining/documenting what version is being used, where is is coming from, what its expected hash is etc. This is a process fault not our part, not a point against dlmalloc

achamayou:
> I think Radhika's point was that OP-TEE is single threaded and wouldn't benefit.

I don't understand why a platform that's currently in preview not benefitting from this change should prevent it happening for supported platforms. Can you enlighten me?

> @anakrish did some work yesterday (Thanks Anand!) to prototype an interface for pluggable allocators: [master...anakrish:ak_allocator](https://github.com/openenclave/openenclave/compare/master...anakrish:ak_allocator)
> 
> I think we should seriously consider this path since as you've pointed out different allocators have different benefits and not all users of the SDK have the same needs.

I'm still very happy for this to happen (thanks @anakrish), but I'm entirely unhappy about it blocking this PR. It's a separate change, from a different author, it should happen in its own PR. Can you please provide a valid reason why these changes must be bundled or made dependent on each other?

> This is more of a problem in that OE has done a poor job defining/documenting what version is being used, where is is coming from, what its expected hash is etc. This is a process fault not our part, not a point against dlmalloc.

It does illustrate the effective lack of concern for allocator versioning so far, and raises the question of why this is being brought up as a blocker. malloc.c includes this statement:

```
  For convenience, an include file for code using this malloc is at:
     ftp://gee.cs.oswego.edu/pub/misc/malloc-2.8.6.h
```

This is not the path OpenEnclave uses, the contents may or may not match those of malloc.h.

By contrast, snmalloc is in git, does have release notes, release tarballs, an extensive (and public) CI: https://github.com/microsoft/snmalloc/releases

achamayou:
@dthaler @mingweishih I have addressed your requested changes, could you please review again?

CMakeLists.txt
```diff
@@ -193,7 +193,10 @@ else ()
   option(USE_DEBUG_MALLOC "Build oeenclave with memory leak detection capability." OFF)
 endif ()
 
-option(ADD_WINDOWS_ENCLAVE_TESTS "Build Windows enclave tests" OFF)
```
mingweishih: Why this is line is removed?

enclave/core/oe_alloc_thread.h
```diff
@@ -0,0 +1,15 @@
...
+{
+#endif
+
+    // Thread-specific startup and teardown calls for the memory allocator
+    void oe_alloc_thread_startup();
+    void oe_alloc_thread_teardown();
+
+#ifdef __cplusplus
+}
+#endif
```
mingweishih:
```suggestion
#endif

```

jhand2:
> @jhand2 I'm still very happy for this to happen (thanks @anakrish), but I'm entirely unhappy about it blocking this PR. It's a separate change, from a different author, it should happen in its own PR. Can you please provide a valid reason why these changes must be bundled or made dependent on each other?

I'll take another look at this PR in a little bit. My main concerns about about replacing dlmalloc, which of course this PR doesn't do. In principal I don't have a problem with a non-invasive option. I'll get to this as soon as I can.

3rdparty/dlmalloc/dlmalloc/malloc.h
```diff
@@ -1,12 +1,12 @@
 /*
```
mingweishih: Is there any modification to this file other than formatting? If not, I would prefer to keep the file as it is. I believe this file was directly copied from the 3rd party source. Please correct me if I'm wrong. Thanks!

CMakeLists.txt
mingweishih: Why this is line is removed?

jbdelcuv:
The SGX SDK added support for the TCMalloc library in the Linux 1.8 release (Mar. 2017).
Back then, we reviewed an number of open-source allocators (snmalloc wasn't available at that time) and decided that TCMalloc provided the scalability that customers were looking for.
We also made the decision to let developers choose the allocator to be linked into their enclaves at build time. Enclaves link with dlmalloc (included in the C standard library) by default. However, a developer can easily link TCMalloc providing the corresponding library and some flags in the linker command line.

As achamayou mentioned above, it appears that nowadays snmalloc is in a better position from a (security and functional) support standpoint.

With snmalloc, what's the default alignment for memory allocations? To store cryptographic keys it's convenient to support 16-byte alignment, for instance.
Does snmalloc call a global constructor during its initialization? 
This is the case with TCMalloc, which prevents using it in simulation mode, and the main reason dlmalloc remains the default allocator in the SGX SDK.

enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
```diff
@@ -0,0 +1,116 @@
...
+extern "C" void oe_set_allocation_failure_callback(
+    oe_allocation_failure_callback_t function)
+{
+    (void)function;
+}
+
+extern "C" void oe_memalign_free(void* ptr)
+{
+    oe_free(ptr);
+}
```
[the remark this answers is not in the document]
jhand2: The OE style guide says no newline at the end of a file

radhikaj:
@achamayou 
1) Please mark this feature as experimental. So when the SDK is built with snmalloc, it is built by opting into the experimental flag. Reasons for this are
  - CI does not build OE SDK with USE_SNMALLOC, so this PR is not tested in CI
  - A better direction for the OE SDK project is to have allocator plugins, So when we have the ability to plugin an allocator,  snmalloc can become a plugin
2) When we have a pluggable allocator design and implementation, the USE_SNMALLOC option will need to be removed. snmalloc will be a plugin.

include/openenclave/internal/sgxtypes.h
```diff
@@ -625,6 +625,8 @@ typedef struct _td
     struct _oe_ecall_context* host_ecall_context;
     struct _oe_ecall_context* host_previous_ecall_context;
 
+    void* allocator_local;
```
jhand2: Instead of adding a member to td_t, can this be a thread-local variable (__thread keyword) in the snmalloc wrapper file?

mjp41: I think that will work, but we still need the thread start-up and teardown calls.   That would also simplify the situation for when it becomes pluggable.

enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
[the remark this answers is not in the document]
achamayou: Earlier in the same PR:

image

The style guidelines do not say anything about this, as far as I can tell: https://github.com/openenclave/openenclave/blob/master/docs/DevelopmentGuide.md

The clang format config is based on the LLVM guidelines, which also do not mention this issue as far as I can tell (https://llvm.org/docs/CodingStandards.html). Disclaimer: I searched for keywords and scanned the index, I did not read the whole thing.

I'm going with: this passes the clang-format check, in local hooks and in CI. If someone feels strongly about line returns at the end of files, I suggest they tweak the clang-format configuration accordingly!

3rdparty/dlmalloc/dlmalloc/malloc.h
achamayou: Sure, me too. But the clang-format hook disagrees. So this won't commit without override, nor pass in the CI.

Maybe 3rdparty should be exempt from formatting? That's what we do in CCF. I do not feel strongly about this, I'm happy to go with anything.

mjp41:
@jbdelcuv 
> With snmalloc, what's the default alignment for memory allocations? To store some cryptographic keys it's convenient to support 16-byte alignment, for instance.

The implementation currently makes every block aligned to at least the least set bit of its size.  I.e. 16 byte allocation is 16 byte aligned, 32 byte request is 32 byte aligned, 48 is 16 byte aligned. This means we can handled aligned requests of any size with just two instructions additional instructions on the fast path.

> Does snmalloc call a global constructor during its initialization?

The implementation we have provided for open enclave does not.  We have a bunch of abstraction around the platform and thread local state.  For both standard Linux and Windows, we have a global initialiser as we can't hook into the thread startup earlier enough to guarantee no allocations have already happened without fast path cost.

enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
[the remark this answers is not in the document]
jhand2: Ah, yes, when I said style guide I meant clang-format. My point was that whatever passes clang-format is fine (and anything that doesn't pass clang-format will fail CI anyway)

CMakeLists.txt
achamayou: Because rebasing through dozens of conflicts is error prone. I will fix that, sorry.

include/openenclave/internal/sgxtypes.h
jhand2: I think a pluggable allocator could support plugin functions for enclave and thread lifecycle events so IMO, the startup and teardown calls are fine.

enclave/core/oe_alloc_thread.h
achamayou: I'm going to invoke https://github.com/openenclave/openenclave/blob/master/docs/DevelopmentGuide.md, specifically this bit:

```
The project Committers prefer to fix style issues in bulk using automation, so avoid submitting PRs intended to fix only a few instances of the inconsistent style.
```

If new lines at the end of files matter, then I suggest the clang format config be set accordingly for automated checking/conversion.

achamayou:
@radhikaj yes, I will mark this as experimental. I look forward to pluggable allocator support in OE, and I'll try to help with that as much as possible.

3rdparty/dlmalloc/dlmalloc/malloc.h
mingweishih: Yes you are correct --- 3rdparty should be exempt from formatting. However, we currently seems to do that manually.

In this case you need to do force push (I remember there is a option to ignore format checks on certain files).

CMakeLists.txt
mingweishih: Thanks!

enclave/core/oe_alloc_thread.h
mingweishih: I remember in some cases format checks do catch this. Given that the checks sill miss the other cases (e.g., like this one), we need to open an issue for that.

CMakeLists.txt
achamayou: Thanks for catching this, fixed now.

enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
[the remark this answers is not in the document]
achamayou:
```
amchamay@amchamay2:/data/src/openenclave$ scripts/format-code; echo $?
0
amchamay@amchamay2:/data/src/openenclave$ git diff
diff --git a/enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp b/enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
index c2e488141..cb7549491 100644
--- a/enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
+++ b/enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
@@ -113,4 +113,4 @@ extern "C" void oe_set_allocation_failure_callback(
 extern "C" void oe_memalign_free(void* ptr)
 {
     oe_free(ptr);
-}
\ No newline at end of file
+}
amchamay@amchamay2:/data/src/openenclave$ scripts/format-code; echo $?
0
```

As far as I can tell, clang-format as we have it configured does not care.

enclave/core/oe_alloc_thread.h
achamayou: There's no option for this in clang-format as far as I can tell. This also does not trigger a warning in clang for snmalloc_wrapper.cpp, because since C++11, there's literally no difference as far as the compiler is concerned:

> A source file that is not empty and that does not end in a new-line character, or that ends in a new-line character immediately preceded by a backslash character before any such splicing takes place, shall be processed as if an additional new-line character were appended to the file (C++11 §2.2/1).

I would have thought there would be one for oe_alloc_thread.h, but did not find one. A linux build of OE on Ubuntu prints out 304 warnings though, so I'm not certain it would have caught my eye.

CMakeLists.txt
```diff
@@ -195,6 +195,14 @@ endif ()
 
 option(ADD_WINDOWS_ENCLAVE_TESTS "Build Windows enclave tests" OFF)
 
+# NOTE: Building OpenEnclave using snmalloc for memory allocation is an experimental option.
+# Longer term, there will be a pluggable allocator mechanism allow users to chose from a
```
dthaler: grammar: "allow" -> "to allow"
"chose" -> "choose"

enclave/core/atexit.c
```diff
@@ -41,10 +41,11 @@ static oe_spinlock_t _spin = OE_SPINLOCK_INITIALIZER;
 
 static oe_atexit_entry_t* _new_atexit_entry(void (*func)(void*), void* arg)
 {
+    extern void* oe_nodebug_malloc(size_t);
```
dthaler: Rather than copying the same prototypes into multiple .c files (here, debugmalloc.c, sgx/backtrace.c, etc.), can you put them into a header file like oe_alloc_thread.h?   Want to avoid different files being out of sync if the prototype ever changes in the future.

enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
```diff
@@ -0,0 +1,116 @@
...
+extern "C" size_t oe_malloc_usable_size(void* ptr)
+{
+    return oe_debug_malloc_usable_size(ptr);
+}
+#else
+#define SNMALLOC_NAME_MANGLE(a) oe_##a
+#endif
+// The OpenEnclave runtime manages the thread local allocator pointer, see td_t
+#define SNMALLOC_EXTERNAL_THREAD_ALLOC
+// Enable the OpenEnclave PAL(Platform Abstraction Layer) for OpenEnclave,
```
dthaler: nit: "Open Enclave" should always have a space in the middle (see http://openenclave.io/sdk/).  2 occurrences needing a space in this line.

enclave/core/oe_alloc_thread.h
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#ifdef __cplusplus
```
jhand2: You can use `OE_EXTERNC_BEGIN` and `OE_EXTERNC_END` macros instead

anakrish:
@achamayou What are the instructions for building OE SDK with snmalloc turned on?

I did:
```bash
git submodule update --init --recursive
mkdir build
cd build
cmake .. -DUSE_SNMALLOC=on -DHAS_QUOTE_PROVIDER=off
make -j 16
```

Then all tests that create and load an enclave fail. I put a breakpoint on `oe_abort` and this is what I get:
```
Breakpoint 1, oe_abort () at /home/anakrish/work/pr/enclave/core/sgx/calls.c:987
987	    if (__oe_enclave_status < OE_ENCLAVE_ABORTING)
(gdb) bt
#0  oe_abort () at /home/anakrish/work/pr/enclave/core/sgx/calls.c:987
#1  0x00007ffff6121011 in snmalloc::PALOpenEnclave::error (str=0x7ffff61445f0 "Out of memory")
    at /home/anakrish/work/pr/enclave/core/sgx/snmalloc/../../../../3rdparty/snmalloc/src/override/../mem/../ds/../pal/pal_open_enclave.h:25
#2  0x00007ffff6122978 in snmalloc::PALOpenEnclave::reserve<false> (this=0x7ffff6970920, size=8388608)
    at /home/anakrish/work/pr/enclave/core/sgx/snmalloc/../../../../3rdparty/snmalloc/src/override/../mem/../ds/../pal/pal_open_enclave.h:47
#3  0x00007ffff61220c1 in snmalloc::MemoryProviderStateMixin<snmalloc::PALPlainMixin<snmalloc::PALOpenEnclave> >::reserve<false> (this=0x7ffff6970910, large_class=0)
    at /home/anakrish/work/pr/enclave/core/sgx/snmalloc/../../../../3rdparty/snmalloc/src/override/../mem/largealloc.h:275
#4  0x00007ffff6121de6 in snmalloc::MemoryProviderStateMixin<snmalloc::PALPlainMixin<snmalloc::PALOpenEnclave> >::new_block (this=0x7ffff6970910)
    at /home/anakrish/work/pr/enclave/core/sgx/snmalloc/../../../../3rdparty/snmalloc/src/override/../mem/largealloc.h:127
#5  0x00007ffff6125251 in snmalloc::MemoryProviderStateMixin<snmalloc::PALPlainMixin<snmalloc::PALOpenEnclave> >::alloc_chunk<snmalloc::MemoryProviderStateMixin<snmalloc::PALPlainMixin<snmalloc::PALOpenEnclave> >, 1ul> (
    this=0x7ffff6970910) at /home/anakrish/work/pr/enclave/core/sgx/snmalloc/../../../../3rdparty/snmalloc/src/override/../mem/largealloc.h:242
#6  0x00007ffff6124f4d in snmalloc::MemoryProviderStateMixin<snmalloc::PALPlainMixin<snmalloc::PALOpenEnclave> >::make ()
    at /home/anakrish/work/pr/enclave/core/sgx/snmalloc/../../../../3rdparty/snmalloc/src/override/../mem/largealloc.h:98
#7  0x00007ffff6124e3b in snmalloc::Singleton<snmalloc::MemoryProviderStateMixin<snmalloc::PALPlainMixin<snmalloc::PALOpenEnclave> >*, &snmalloc::MemoryProviderStateMixin<snmalloc::PALPlainMixin<snmalloc::PALOpenEnclave> >::make>::get (first=0x0) at /home/anakrish/work/pr/enclave/core/sgx/snmalloc/../../../../3rdparty/snmalloc/src/override/../mem/../ds/helpers.h:36
#8  0x00007ffff612199a in snmalloc::default_memory_provider () at /home/anakrish/work/pr/enclave/core/sgx/snmalloc/../../../../3rdparty/snmalloc/src/override/../mem/largealloc.h:429
#9  0x00007ffff61212f9 in snmalloc::AllocPool<snmalloc::MemoryProviderStateMixin<snmalloc::PALPlainMixin<snmalloc::PALOpenEnclave> > >::make ()
    at /home/anakrish/work/pr/enclave/core/sgx/snmalloc/../../../../3rdparty/snmalloc/src/override/../mem/globalalloc.h:41
#10 0x00007ffff612122b in snmalloc::Singleton<snmalloc::AllocPool<snmalloc::MemoryProviderStateMixin<snmalloc::PALPlainMixin<snmalloc::PALOpenEnclave> > >*, &snmalloc::AllocPool<snmalloc::MemoryProviderStateMixin<snmalloc::PALPlainMixin<snmalloc::PALOpenEnclave> > >::make>::get (first=0x0) at /home/anakrish/work/pr/enclave/core/sgx/snmalloc/../../../../3rdparty/snmalloc/src/override/../mem/../ds/helpers.h:36
#11 0x00007ffff6120f3d in snmalloc::current_alloc_pool () at /home/anakrish/work/pr/enclave/core/sgx/snmalloc/../../../../3rdparty/snmalloc/src/override/../mem/globalalloc.h:172
#12 0x00007ffff611ff36 in oe_alloc_thread_startup () at /home/anakrish/work/pr/enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp:92
#13 0x00007ffff611bd30 in td_init (td=0x7ffff6977000) at /home/anakrish/work/pr/enclave/core/sgx/td_basic.c:115
#14 0x00007ffff610b784 in _handle_ecall (td=0x7ffff6977000, func=1, arg_in=4804272, output_arg1=0x7ffff6970fc8, output_arg2=0x7ffff6970fc0) at /home/anakrish/work/pr/enclave/core/sgx/calls.c:360
#15 0x00007ffff610b3ef in __oe_handle_main (arg1=281479271677952, arg2=4804272, cssa=0, tcs=0x7ffff6972000, output_arg1=0x7ffff6970fc8, output_arg2=0x7ffff6970fc0) at /home/anakrish/work/pr/enclave/core/sgx/calls.c:955
#16 0x00007ffff610e33a in oe_enter () at /home/anakrish/work/pr/enclave/core/sgx/enter.S:183
#17 0x00000000004694b2 in __morestack (tcs=0x7ffff6972000, aep=4625144, arg1=281479271677952, arg2=4804272, arg3=0x7fffffffd6b8, arg4=0x7fffffffd6b0, enclave=0x494eb0) at /home/anakrish/work/pr/host/sgx/enter.c:163
#18 0x000000000045841f in _do_eenter (enclave=0x494eb0, tcs=0x7ffff6972000, aep=4625144, code_in=OE_CODE_ECALL, func_in=1, arg_in=4804272, code_out=0x7fffffffd6ac, func_out=0x7fffffffd6aa, result_out=0x7fffffffd6a8, 
    arg_out=0x7fffffffd6a0) at /home/anakrish/work/pr/host/sgx/calls.c:192
#19 oe_ecall (enclave=0x494eb0, func=1, arg=4804272, arg_out_ptr=0x7fffffffd820) at /home/anakrish/work/pr/host/sgx/calls.c:626
#20 0x000000000040df22 in _initialize_enclave (enclave=0x494eb0) at /home/anakrish/work/pr/host/sgx/create.c:411
#21 0x000000000040da2a in oe_create_enclave (enclave_path=0x7fffffffe2b2 "enc/echo_enc", enclave_type=OE_ENCLAVE_TYPE_SGX, flags=1, settings=0x0, setting_count=0, ocall_table=0x47d600 <__echo_ocall_function_table>, ocall_count=1, 
    enclave_out=0x7fffffffddc8) at /home/anakrish/work/pr/host/sgx/create.c:831
#22 0x0000000000409263 in oe_create_echo_enclave (path=0x7fffffffe2b2 "enc/echo_enc", type=OE_ENCLAVE_TYPE_SGX, flags=1, settings=0x0, setting_count=0, enclave=0x7fffffffddc8)
    at /home/anakrish/work/pr/build/tests/echo/host/echo_u.c:195
#23 0x0000000000408100 in main (argc=2, argv=0x7fffffffdf28) at /home/anakrish/work/pr/tests/echo/host/host.c:38
(gd
```

anakrish:
@achamayou I encountered compilation issues when trying to write a test in which enclaves are built using snmalloc. OE does not provide `cpuid.h` and the way in which I wrote the test, the enclave was using OE provided headers instead of the system include directory.

https://github.com/microsoft/snmalloc/blob/f4de00cbca6003c9a6d22ad0944465e855210ca6/src/aal/aal_x86.h#L3-L13

Does snmalloc rely on `cpuid` and `rdtsc`? Do they work reliably within enclaves?

enclave/core/oe_alloc_thread.h
achamayou: I've added final newlines here and elsewhere. I can't find a way to get clang format to do that, so I'm leaving that for another PR.

3rdparty/dlmalloc/dlmalloc/malloc.h
achamayou: I have done that now.

CMakeLists.txt
achamayou: Fixed

enclave/core/atexit.c
achamayou: Sure, the signatures are those of malloc/realloc/..., so I don't expect them to change any time soon, but that's fine.

enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
achamayou: Fixed

enclave/core/oe_alloc_thread.h
achamayou: Ah, I did not know about those, will do.

enclave/core/atexit.c
achamayou: Done

achamayou:
@jhand2 that's done now. Thank you for the helpful suggestions!

achamayou:
bors try

achamayou:
@anakrish that's the right way to build with snmalloc, the reason why those tests fail now is because they define a heap size that's too small for snmalloc's initial allocation. This was discussed earlier in the thread here: https://github.com/openenclave/openenclave/pull/2649#discussion_r391882371

You can check that by increasing the HEAP_SIZE in OE_SET_ENCLAVE_SGX(), or adding a factor in OE_SET_ENCLAVE_SGX() directly.

I think we can do a couple things about this, one is we can validate the size early on and give a friendlier error. The other is we can clamp the minimum size when using snmalloc. What do you prefer?

snmalloc does not rely on `cpuid` or `rdtsc`, they're only used in benchmarks. @mjp41 has added an SGX-specific PAL to avoid those includes altogether when using SGX.

achamayou:
bors try

enclave/core/atexit.c
```diff
@@ -43,8 +44,8 @@ static oe_atexit_entry_t* _new_atexit_entry(void (*func)(void*), void* arg)
 {
     oe_atexit_entry_t* entry;
 
-    if ((entry = (oe_atexit_entry_t*)oe_sbrk(sizeof(oe_atexit_entry_t))) ==
-        (void*)-1)
+    if ((entry = (oe_atexit_entry_t*)oe_nodebug_malloc(
```
CodeMonkeyLeet: Nit: can you correct the comment on line 38 since it no longer uses `sbrk()` for this purpose?

enclave/core/sgx/snmalloc/snmalloc_wrapper.cpp
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/internal/sgxtypes.h>
+#include "../../oe_alloc_thread.h"
+#include "../../oe_nodebug_alloc.h"
+
+namespace snmalloc
+{
+__thread void* allocator_local;
```
anakrish: Thanks for using a thread local as opposed to td_t. It allows us to explore removing td_t (#2554) without worrying that we may break clients like CCF.

achamayou:
@anakrish we look forward to snmalloc being used on other platforms, @mjp41 has added support for arm32(hf) and arm64 to snmalloc yesterday, and I've added the corresponding builds in the CI there: https://dev.azure.com/snmalloc/snmalloc/_build/results?buildId=817&view=results

enclave/core/atexit.c
achamayou: Sure, done.

achamayou:
bors try

achamayou:
bors try-

achamayou:
bors try

achamayou:
The Windows 2016 Sim Cross Compile has apparently failed on:

```
[76/2211] Performing download step for 'musl_includes'
ninja: build stopped: subcommand failed.
script returned exit code 1
```

I assume this is a transient error?

PR #2653, opened 2020-03-11
Fix backtrace test under gcc 8

anakrish:
GCC generates functions with .constprop, .clone and other suffixes
with optimizations turned on. In this test, GCC generates
func4.constprop.0 function that does a part of the logic in func4.
This generated function was part of the backtrace, and the
test failed since it was expecting func4. The fix is to ignore the
suffixes in compating expected and actual symbols.


anakrish:
bors try

Issue #2075, opened 2019-08-05
Enable /WX on Windows compilation and fix existing issues


CodeMonkeyLeet:
Builds of OE SDK on Windows today spew a lot of build warnings and these need to be cleaned up. At the same time, we need to stop introducing new build warnings on Windows by enabling the `/WX` flag (treat warnings as errors) to enforce this.

jhand2:
#2374 fixes /W1 warnings across the codebase and enables /WX for W1 warnings.

I'll consider this issue closed when we have /W3 /WX enabled but this is a good first step.

jhand2:
#2389 fixes W2 warnings

radhikaj:
@ryanhsu19 please continue looking  into this

ryanhsu19:
#2654 fixes /W3 warnings

Issue #2376, opened 2019-12-10
OE is not C99 compliant


jhand2:
From discussing with @mikbras, OE should compile against C99. Unfortunately, it does not.

In particular, the main thing I haven't been able to overcome is OE_STATIC_ASSERT:

```{C}
#define OE_STATIC_ASSERT(COND)       \
    typedef unsigned char OE_CONCAT( \
        __OE_STATIC_ASSERT, __LINE__)[(COND) ? 1 : -1] OE_UNUSED_ATTRIBUTE
```

It depends on type redefinition (`__OE_STATIC_ASSERT<line_number>` can be defined multiple times). Type redefinition is a C11 feature and using it fails the build when the C standard is set to C99.

We'll need to come up with something else if we want OE to be compilable with C99. If we don't care, we can just set the standard to C11 and document that.

jxyang:
Do we have customers working a platform that doesn't allow upgrade to a C11 compiler? @mikbras

dthaler:
My opinion: I believe that OE needs to support apps that need to consume the SDK, and need C99.  However, if they consume OE SDK in binary form, that does not mean that OE SDK needs C99 to compile.   Thus, I care that public headers and samples are C99 compliant, but internal code may not need to be.

jhand2:
>  I care that public headers and samples are C99 compliant, but internal code may not need to be.

Agreed

As for enforcing that this stays true.

Most of the samples are C++. `samples/helloworld` and `samples/switchless` are C, I'll check if those two can compile with C99 as-is.

As for public headers, I'm not totally sure how to enforce that via a static/build time check. Open to any options people know of.

PR #2654, opened 2020-03-11
Enable /W3 and fix /W3 warnings for Windows builds

ryanhsu19:
I change the compiler settings so that the compiler will treat /W3 warnings as errors. 

I fix the warnings as follows:

1. With `_CRT_SECURE_NO_WARNINGS` turned on, fix the rest of warnings.

2. Remove `CRT_SECURE_NO_WARNINGS` and add pragmas to suppress warnings. Then restore secure CRT functions and remove pragmas one by one.

3. Fix any other additional warnings (casting problems, pointer mismatches, etc.).

Most changes each have an `ifdef`-`else`-`endif` branch to deal with different functions used on different platforms. 

However, I need some advice on the following three things:

1. The compiler suggests changing `getenv()` to `_dupenv_s()`. However, I find that the buffer used by `_dupenv_s()` needs to be freed up manually. Therefore, code in the two associated files become a little messy. 

2. I change the `gmtime()` to `gmtime_s()` and `gmtime_r()` on Windows and Linux, respectively. However, I also change the declaration from `struct tm* tm;` to `struct tm tm;`. Here's why:
2a) If I use `struct tm* tm;` before calling them then there will be an uninitialized variable error, and 
2b) If I use `struct tm* tm = NULL;` both functions will not be happy with the NULL pointer and cause an error during ctest.

3. I only keep the pragma `#pragma warning(disable : 4996)` in the following files:
3a) `common/oe_host_string.h` for keeping both `oe_strerror()` and `oe_strerror_r()`.
3b)` tests/crypto/enclave/host/host.c` for using `_open()` instead of `_sopen_s()`, which I find difficult to change the original flags and modes to the corresponding flags and permissions.
3c) `tests/mbed/host/ocalls.c` for the reason mentioned above.

Fix part of #2075.

### DIFF
--- a/common/datetime.c
+++ b/common/datetime.c
@@ -229,20 +229,20 @@ oe_result_t oe_datetime_now(oe_datetime_t* value)
 {
     oe_result_t result = OE_UNEXPECTED;
     time_t now;
-    struct tm* timeinfo;
+    struct tm timeinfo;
 
     if (value == NULL)
         OE_RAISE(OE_INVALID_PARAMETER);
 
     time(&now);
-    timeinfo = gmtime(&now);
+    gmtime_r(&now, &timeinfo);
 
-    value->year = (uint32_t)timeinfo->tm_year + 1900;
-    value->month = (uint32_t)timeinfo->tm_mon + 1;
-    value->day = (uint32_t)timeinfo->tm_mday;
-    value->hours = (uint32_t)timeinfo->tm_hour;
-    value->minutes = (uint32_t)timeinfo->tm_min;
-    value->seconds = (uint32_t)timeinfo->tm_sec;
+    value->year = (uint32_t)timeinfo.tm_year + 1900;
+    value->month = (uint32_t)timeinfo.tm_mon + 1;
+    value->day = (uint32_t)timeinfo.tm_mday;
+    value->hours = (uint32_t)timeinfo.tm_hour;
+    value->minutes = (uint32_t)timeinfo.tm_min;
+    value->seconds = (uint32_t)timeinfo.tm_sec;
 
     result = OE_OK;
 done:
--- a/common/oe_host_string.h
+++ b/common/oe_host_string.h
@@ -38,10 +38,21 @@ int oe_strncmp(const char* s1, const char* s2, size_t n)
 
 /* host already has an oe_strlcat implementation */
 
+/*
+ * There are separate OE functions for strerror and strerror_s.
+ * Therefore we suppress the warning about not using more secure CRT functions.
+ */
 OE_INLINE
 char* oe_strerror(int errnum)
 {
+#if defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable : 4996)
+#endif
     return strerror(errnum);
+#if defined(_MSC_VER)
+#pragma warning(pop)
+#endif
 }
 
 OE_INLINE
--- a/debugger/debugrt/host/host.c
+++ b/debugger/debugrt/host/host.c
@@ -47,13 +47,26 @@ static bool raise_debugger_events()
         {
             // If specified, override oe_debugger_contract_version from the
             // environment.
-            char* version = getenv("OE_DEBUGGER_CONTRACT_VERSION");
-            if (version != NULL)
+            char* version;
+            size_t getenv_length;
+            getenv_s(&getenv_length, NULL, 0, "OE_DEBUGGER_CONTRACT_VERSION");
+
+            if (getenv_length > 0)
             {
-                int v = 0;
-                if (sscanf(version, "%d", &v) == 1)
+                version = (char*)malloc(sizeof(char) * getenv_length);
+                getenv_s(
+                    &getenv_length,
+                    version,
+                    sizeof(char) * getenv_length,
+                    "OE_DEBUGGER_CONTRACT_VERSION");
+
+                if (version != NULL)
                 {
-                    oe_debugger_contract_version = (uint32_t)v;
+                    int v = 0;
+                    if (sscanf_s(version, "%d", &v) == 1)
+                    {
+                        oe_debugger_contract_version = (uint32_t)v;
+                    }
                 }
             }
 
--- a/host/traceh.c
+++ b/host/traceh.c
@@ -3,6 +3,7 @@
 
 #include <openenclave/corelibc/limits.h>
 #include <openenclave/internal/calls.h>
+#include <openenclave/internal/datetime.h>
 #include <openenclave/internal/raise.h>
 #include <openenclave/internal/safecrt.h>
 #include <openenclave/internal/trace.h>
@@ -19,6 +20,11 @@
 #include "hostthread.h"
 
 #define LOGGING_FORMAT_STRING "%s.%06ldZ [(%s)%s] tid(0x%llx) | %s"
+
+#if defined(__linux__)
+#define sprintf_s(buffer, size, format, argument) \
+    sprintf(buffer, format, argument)
+#endif
 
 static char* _log_level_strings[OE_LOG_LEVEL_MAX] =
     {"NONE", "FATAL", "ERROR", "WARN", "INFO", "VERBOSE"};
@@ -201,8 +207,11 @@ static bool _escape_characters(
                         log_msg_escaped[idx] = '\0';
                         return false;
                     }
-                    sprintf(
-                        (char*)&log_msg_escaped[idx], "u%04hhx", log_msg[i]);
+                    sprintf_s(
+                        (char*)&log_msg_escaped[idx],
+                        msg_size - idx,
+                        "u%04hhx",
+                        log_msg[i]);
                     // idx is also incremented after switch case
                     idx += MAX_ESCAPED_CHAR_LEN - 1;
                     break;
@@ -310,23 +319,14 @@ done:
 void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
 {
     // get timestamp for log
-#if defined(__linux__)
-    struct timeval time_now;
-    gettimeofday(&time_now, NULL);
-    struct tm* t = gmtime(&time_now.tv_sec);
-#else
+    struct tm t;
     time_t lt = time(NULL);
-    struct tm* t = gmtime(&lt);
-#endif
+    gmtime_r(&lt, &t);
 
     char time[20];
-    strftime(time, sizeof(time), "%Y-%m-%dT%H:%M:%S", t);
-
-#if defined(__linux__)
-    long int usecs = time_now.tv_usec;
-#else
+    strftime(time, sizeof(time), "%Y-%m-%dT%H:%M:%S", &t);
     long int usecs = 0;
-#endif
+
     if (!_initialized)
     {
         initialize_log_config();
--- a/host/windows/syscall.c
+++ b/host/windows/syscall.c
@@ -17,12 +17,14 @@
 #include <io.h>
 #include <stdint.h>
 #include <sys/stat.h>
+#include <sys/types.h>
 
 // clang-format off
 
 #include <winsock2.h>
 #include <windows.h>
 #include <Ws2def.h>
+#include <Ws2tcpip.h>
 #include <VersionHelpers.h>
 // clang-format on
 
@@ -934,15 +936,21 @@ ssize_t oe_syscall_send_ocall(
     size_t len,
     int flags)
 {
-    ssize_t ret;
+    ssize_t ret = OE_EINVAL;
     _set_errno(0);
 
-    ret = send(_get_socket(sockfd), buf, len, flags);
+    if ((len & 0xFFFFFFFF) != len)
+    {
+        _set_errno(OE_EINVAL);
+        goto done;
+    }
+
+    ret = send(_get_socket(sockfd), buf, (int)len, flags);
     if (ret == SOCKET_ERROR)
     {
         _set_errno(_winsockerr_to_errno(WSAGetLastError()));
     }
-
+done:
     return ret;
 }
 
@@ -957,10 +965,16 @@ ssize_t oe_syscall_sendto_ocall(
     ssize_t ret;
     _set_errno(0);
 
+    if ((len & 0xFFFFFFFF) != len)
+    {
+        _set_errno(OE_EINVAL);
+        goto done;
+    }
+
     ret = sendto(
         _get_socket(sockfd),
         buf,
-        len,
+        (int)len,
         flags,
         (struct sockaddr*)src_addr,
         addrlen);
@@ -968,7 +982,7 @@ ssize_t oe_syscall_sendto_ocall(
     {
         _set_errno(_winsockerr_to_errno(WSAGetLastError()));
     }
-
+done:
     return ret;
 }
 
@@ -1506,13 +1520,13 @@ int oe_syscall_uname_ocall(struct oe_utsname* buf)
     // OE SDK is supported only on WindowsServer and Win10
     if (IsWindowsServer())
     {
-        sprintf(buf->sysname, "WindowsServer");
-        sprintf(buf->version, "2016OrAbove");
+        sprintf_s(buf->sysname, sizeof(buf->sysname), "WindowsServer");
+        sprintf_s(buf->version, sizeof(buf->version), "2016OrAbove");
     }
     else if (IsWindows10OrGreater())
     {
-        sprintf(buf->sysname, "Windows10OrGreater");
-        sprintf(buf->version, "10OrAbove");
+        sprintf_s(buf->sysname, sizeof(buf->sysname), "Windows10OrGreater");
+        sprintf_s(buf->version, sizeof(buf->version), "10OrAbove");
     }
 
     ret = 0;
--- a/include/openenclave/host.h
+++ b/include/openenclave/host.h
@@ -29,6 +29,24 @@
 
 OE_EXTERNC_BEGIN
 
+#ifndef _WIN32
+#define _getpid getpid
+#define sscanf_s sscanf
+#define sprintf_s(buffer, size, format, argument) \
+    sprintf(buffer, format, argument)
+#define strcat_s(destination, destination_size, source) \
+    strcat(destination, source)
+#define strcpy_s(destination, destination_size, source) \
+    ;                                                   \
+    {                                                   \
+        (void)(destination_size);                       \
+        strcpy(destination, source);                    \
+    }
+#define _strdup strdup
+#define strncat_s(destination, destination_size, source, source_size) \
+    strncat(destination, source, source_size)
+#endif
+
 /**
  *  Flag passed into oe_create_enclave to run the enclave in debug mode.
  *  The flag allows the enclave to be created without the enclave binary
--- a/tests/attestation_cert_apis/host/host.cpp
+++ b/tests/attestation_cert_apis/host/host.cpp
@@ -20,6 +20,8 @@
 #define TEST_RSA_KEY 1
 #define SKIP_RETURN_CODE 2
 
+#define FILENAME_LENGTH 80
+
 // This is the identity validation callback. An TLS connecting party (client or
 // server) can verify the passed in "identity" information to decide whether to
 // accept an connection reqest
@@ -92,16 +94,21 @@ void run_test(oe_enclave_t* enclave, int test_type)
 
     {
         // for testing purpose, output the whole cer in DER format
-        char filename[80];
+        char filename[FILENAME_LENGTH];
         FILE* file = NULL;
 
-        sprintf(
+        sprintf_s(
             filename,
+            sizeof(filename),
             "./cert_%s.der",
             test_type == TEST_RSA_KEY ? "rsa" : "ec");
         OE_TRACE_INFO(
             "Host: Log quote embedded certificate to file: [%s]\n", filename);
+#ifdef _WIN32
+        fopen_s(&file, filename, "wb");
+#else
         file = fopen(filename, "wb");
+#endif
         fwrite(cert, 1, cert_size, file);
         fclose(file);
     }
--- a/tests/crypto/enclave/host/host.c
+++ b/tests/crypto/enclave/host/host.c
@@ -38,7 +38,10 @@ int f_open(char* path, int flags, int mode)
      * also processed as binary (as they would be equivalently on Linux).
      */
     flags |= _O_BINARY;
+#pragma warning(push)
+#pragma warning(disable : 4996)
     return _open(path, flags, mode);
+#pragma warning(pop)
 #else
     return open(path, flags, mode);
 #endif
@@ -56,7 +59,7 @@ int f_openat(int dirfd, char* path, int flags, int mode)
 int f_read(int fd, char* ptr, size_t len)
 {
 #if defined(_WIN32)
-    return (int)_read(fd, ptr, len);
+    return (int)_read(fd, ptr, (int)len);
 #else
     return (int)read(fd, ptr, len);
 #endif
--- a/tests/crypto/read_file.c
+++ b/tests/crypto/read_file.c
@@ -5,10 +5,21 @@
 #include <openenclave/corelibc/string.h>
 #include "readfile.h"
 
+FILE* read_file(const char* filename, const char* mode)
+{
+    FILE* file;
+#ifdef _MSC_VER
+    fopen_s(&file, filename, mode);
+#else
+    file = fopen(filename, mode);
+#endif
+    return file;
+}
+
 oe_result_t read_cert(char* filename, char* cert)
 {
     size_t len_cert;
-    FILE* cfp = fopen(filename, "rb");
+    FILE* cfp = read_file(filename, "rb");
     if (cfp != NULL)
     {
         len_cert = fread(cert, sizeof(char), max_cert_size, cfp);
@@ -30,8 +41,8 @@ oe_result_t read_chain(
 {
     size_t len_cert1 = 0, len_cert2 = 0;
     char chain_temp[max_cert_size];
-    FILE* cfp1 = fopen(filename1, "rb");
-    FILE* cfp2 = fopen(filename2, "rb");
+    FILE* cfp1 = read_file(filename1, "rb");
+    FILE* cfp2 = read_file(filename2, "rb");
 
     if (cfp1 != NULL && cfp2 != NULL)
     {
@@ -61,9 +72,9 @@ oe_result_t read_chains(
     size_t len_cert1 = 0, len_cert2 = 0, len_cert3 = 0;
     char chain_temp1[max_cert_size];
     char chain_temp2[max_cert_size];
-    FILE* cfp1 = fopen(filename1, "rb");
-    FILE* cfp2 = fopen(filename2, "rb");
-    FILE* cfp3 = fopen(filename3, "rb");
+    FILE* cfp1 = read_file(filename1, "rb");
+    FILE* cfp2 = read_file(filename2, "rb");
+    FILE* cfp3 = read_file(filename3, "rb");
 
     if (cfp1 != NULL && cfp2 != NULL && cfp3 != NULL)
     {
@@ -89,7 +100,7 @@ oe_result_t read_chains(
 oe_result_t read_crl(char* filename, uint8_t* crl, size_t* crl_size)
 {
     size_t len_crl = 0;
-    FILE* cfp = fopen(filename, "rb");
+    FILE* cfp = read_file(filename, "rb");
 
     if (cfp != NULL)
     {
@@ -109,7 +120,7 @@ oe_result_t read_dates(char* filename, oe_datetime_t* time)
 {
     size_t len_date = 0;
     char buffer[max_date_size];
-    FILE* dfp = fopen(filename, "rb");
+    FILE* dfp = read_file(filename, "rb");
 
     if (dfp != NULL)
     {
@@ -121,7 +132,7 @@ oe_result_t read_dates(char* filename, oe_datetime_t* time)
     }
     buffer[len_date] = '\0';
 
-    sscanf(
+    sscanf_s(
         buffer,
         "%u :%u :%u :%u :%u :%u",
         &(time->year),
@@ -199,7 +210,7 @@ oe_result_t read_mod(char* filename, uint8_t* mod, size_t* mod_size)
     char buffer[(max_mod_size * 2) + 1];
     char* bufp = buffer;
 
-    FILE* mfp = fopen(filename, "rb");
+    FILE* mfp = read_file(filename, "rb");
     if (mfp != NULL)
     {
         numchars = fread(buffer, sizeof(char), max_mod_size * 2, mfp);
@@ -246,7 +257,7 @@ oe_result_t read_mixed_chain(
 oe_result_t read_sign(char* filename, uint8_t* sign, size_t* sign_size)
 {
     size_t len_sign;
-    FILE* sfp = fopen(filename, "rb");
+    FILE* sfp = read_file(filename, "rb");
     if (sfp != NULL)
     {
         len_sign = fread(sign, sizeof(char), max_sign_size, sfp);
@@ -280,7 +291,7 @@ oe_result_t read_pem_key(
     }
 
     /* Open file in binary mode. */
-    if (!(stream = fopen(filename, "rb")))
+    if (!(stream = read_file(filename, "rb")))
     {
         result = OE_FAILURE;
         goto done;
@@ -322,7 +333,7 @@ oe_result_t read_coordinates(
     size_t* y_size)
 {
     size_t len_x, len_y;
-    FILE* cfp = fopen(filename, "rb");
+    FILE* cfp = read_file(filename, "rb");
     if (cfp != NULL)
     {
         len_x = fread(x, sizeof(char), max_coordinates_size, cfp);
--- a/tests/crypto/readfile.h
+++ b/tests/crypto/readfile.h
@@ -22,6 +22,10 @@
 #define max_date_elements 6
 #define max_coordinates_size 32
 
+#ifndef _MSC_VER
+#define sscanf_s sscanf
+#endif
+
 oe_result_t read_cert(char* filename, char* cert);
 
 oe_result_t read_chain(
@@ -40,6 +44,8 @@ oe_result_t read_chains(
 oe_result_t read_crl(char* filename, uint8_t* crl, size_t* crl_size);
 
 oe_result_t read_dates(char* filename, oe_datetime_t* time);
+
+FILE* read_file(const char* filename, const char* mode);
 
 oe_result_t read_mod(char* filename, uint8_t* mod, size_t* mod_size);
 
--- a/tests/echo/enc/enc.c
+++ b/tests/echo/enc/enc.c
@@ -42,7 +42,8 @@ int enc_echo(char* in, char out[100])
         out,
         "oe_host_strdup1",
         host_allocated_str,
-        stack_allocated_str);
+        stack_allocated_str,
+        sizeof(stack_allocated_str));
     if (result != OE_OK)
     {
         return -1;
--- a/tests/echo/host/host.c
+++ b/tests/echo/host/host.c
@@ -11,13 +11,19 @@
 #include "../../../host/strings.h"
 #include "echo_u.h"
 
-int host_echo(char* in, char* out, char* str1, char* str2, char str3[100])
+int host_echo(
+    char* in,
+    char* out,
+    char* str1,
+    char* str2,
+    char* str3,
+    int out_length)
 {
     OE_TEST(strcmp(str1, "oe_host_strdup1") == 0);
     OE_TEST(strcmp(str2, "oe_host_strdup2") == 0);
     OE_TEST(strcmp(str3, "oe_host_strdup3") == 0);
 
-    strcpy(out, in);
+    strcpy_s(out, out_length, in);
 
     return 0;
 }
--- a/tests/libcxx/host/host.cpp
+++ b/tests/libcxx/host/host.cpp
@@ -280,7 +280,7 @@ int main(int argc, const char* argv[])
     }
 
     // Disable stdout buffering on host
-    setbuf(stdout, NULL);
+    setvbuf(stdout, NULL, _IONBF, 0);
 
     printf("=== %s: %s\n", argv[0], argv[1]);
 
--- a/tests/libunwind/host/host.cpp
+++ b/tests/libunwind/host/host.cpp
@@ -106,7 +106,7 @@ int main(int argc, const char* argv[])
     {
         oe_put_err("oe_create_libunwind_enclave(): result=%u", result);
     }
-    uint32_t pid = (uint32_t)getpid();
+    uint32_t pid = (uint32_t)_getpid();
 
     // Invoke "Test()" in the enclave.
     Test(enclave, pid);
--- a/tests/mbed/host/host.c
+++ b/tests/mbed/host/host.c
@@ -13,6 +13,9 @@
 
 #include "mbed_u.h"
 
+#define PATH_LENGTH 1024
+#define TEMP_LENGTH 500
+
 void remove_postfix(char* str, char* postfix)
 {
     char* match;
@@ -44,20 +47,20 @@ char* find_data_file(char* str, size_t size)
      * finds the data file. */
     remove_postfix(token, "-lvi-cfg");
 
-    strncat(str, tail, strlen(tail));
+    strncat_s(str, size, tail, strlen(tail));
     printf("######## data_file: %s ###### \n", token);
     return token;
 }
 
-void datafileloc(char* data_file_name, char* path)
+void datafileloc(char* data_file_name, char* path, int length)
 {
     char* tail = "../enc/";
 #ifdef PROJECT_DIR
-    strcpy(path, PROJECT_DIR);
+    strcpy_s(path, length, PROJECT_DIR);
 #else
     char* separator;
 
-    if (getcwd(path, 1024) != NULL)
+    if (getcwd(path, length) != NULL)
         fprintf(stdout, "Current working dir: %s\n", path);
     else
         perror("getcwd() error");
@@ -70,8 +73,8 @@ void datafileloc(char* data_file_name, char* path)
 
     *separator = '\0'; /* separating string */
 #endif
-    strcat(path, tail);
-    strcat(path, data_file_name);
+    strcat_s(path, length, tail);
+    strcat_s(path, length, data_file_name);
 
 #if defined(_WIN32)
     /* On Windows, replace forward slashes with backslashes in the path */
@@ -87,14 +90,14 @@ void datafileloc(char* data_file_name, char* path)
 
 void Test(oe_enclave_t* enclave, int selftest, char* data_file_name)
 {
-    char path[1024];
+    char path[PATH_LENGTH];
     int return_value = 1;
     char* in_testname = NULL;
     char out_testname[STRLEN];
     struct mbed_args args = {0};
     if (!selftest)
     {
-        datafileloc(data_file_name, path);
+        datafileloc(data_file_name, path, PATH_LENGTH);
         in_testname = path;
     }
 
@@ -127,7 +130,7 @@ void ocall_exit(int arg)
 int main(int argc, const char* argv[])
 {
     oe_result_t result;
-    char temp[500];
+    char temp[TEMP_LENGTH];
     oe_enclave_t* enclave = NULL;
     int selftest = 0;
     uint32_t flags = oe_get_create_flags();
@@ -142,7 +145,7 @@ int main(int argc, const char* argv[])
 
     printf("=== %s: %s\n", argv[0], argv[1]);
 
-    strcpy(temp, argv[1]);
+    strcpy_s(temp, TEMP_LENGTH, argv[1]);
 
     if (strstr(argv[1], "selftest"))
     {
--- a/tests/mbed/host/ocalls.c
+++ b/tests/mbed/host/ocalls.c
@@ -1,6 +1,10 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+#ifdef _WIN32
+#pragma warning(disable : 4996)
+#endif
+
 #include <assert.h>
 #include <fcntl.h>
 #include <openenclave/host.h>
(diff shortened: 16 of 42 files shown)


common/oe_host_string.h
```diff
@@ -1,6 +1,12 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+/*
+ * There are separate OE functions for strerror and strerror_s.
+ * Therefore we suppress the warning about not using more secure CRT functions.
+ */
+#pragma warning(disable : 4996)
```
jhand2: What will this pragma do on gcc/clang? Should this be ifdef'd to only win32?

debugger/debugrt/host/host.c
```diff
@@ -47,14 +47,28 @@ static bool raise_debugger_events()
         {
             // If specified, override oe_debugger_contract_version from the
             // environment.
+#ifdef _WIN32
+            char* version = NULL;
+            size_t size = 0;
+            _dupenv_s(&version, &size, "OE_DEBUGGER_CONTRACT_VERSION");
+#else
             char* version = getenv("OE_DEBUGGER_CONTRACT_VERSION");
+#endif
```
jhand2: Could we use `getenv_s` which is available in C11 on both Windows and Linux?

Alternatively you could use `oe_dupenv()`. In general I would prefer to avoid ifdefs when possible.

debugger/debugrt/host/host.c
```diff
@@ -47,14 +47,28 @@ static bool raise_debugger_events()
...
             if (version != NULL)
             {
                 int v = 0;
+#ifdef _WIN32
+                if (sscanf_s(version, "%d", &v) == 1)
```
jhand2: just use sscanf_s on both platforms

host/traceh.c
```diff
@@ -201,8 +201,16 @@ static bool _escape_characters(
                         log_msg_escaped[idx] = '\0';
                         return false;
                     }
+#ifdef _WIN32
+                    sprintf_s(
```
jhand2: Use sprintf_s on both platforms

host/windows/syscall.c
```diff
@@ -1488,13 +1490,16 @@ int oe_syscall_uname_ocall(struct oe_utsname* buf)
     // OE SDK is supported only on WindowsServer and Win10
     if (IsWindowsServer())
     {
-        sprintf(buf->sysname, "WindowsServer");
-        sprintf(buf->version, "2016OrAbove");
+        sprintf_s(buf->sysname, strlen("WindowsServer") + 1, "WindowsServer");
+        sprintf_s(buf->version, strlen("2016OrAbove") + 1, "2016OrAbove");
```
jhand2: `sizeof(<string>)` gives the same result (sizeof will include the null terminator) and it will be computed at compile time.

host/windows/syscall.c
```diff
@@ -1488,13 +1490,16 @@ int oe_syscall_uname_ocall(struct oe_utsname* buf)
...
     }
     else if (IsWindows10OrGreater())
     {
-        sprintf(buf->sysname, "Windows10OrGreater");
-        sprintf(buf->version, "10OrAbove");
+        sprintf_s(
+            buf->sysname,
+            strlen("Windows10OrGreater") + 1,
```
jhand2: sizeof

host/windows/syscall.c
```diff
@@ -1488,13 +1490,16 @@ int oe_syscall_uname_ocall(struct oe_utsname* buf)
...
+            "Windows10OrGreater");
+        sprintf_s(buf->version, strlen("10OrAbove") + 1, "10OrAbove");
```
jhand2: sizeof

tests/attestation_cert_apis/host/host.cpp
```diff
@@ -92,16 +94,28 @@ void run_test(oe_enclave_t* enclave, int test_type)
...
         FILE* file = NULL;
 
+#ifdef _WIN32
+        sprintf_s(
+            filename,
+            FILENAME_LEN,
+            "./cert_%s.der",
+            test_type == TEST_RSA_KEY ? "rsa" : "ec");
+#else
         sprintf(
```
jhand2: use sprintf_s on linux

tests/attestation_cert_apis/host/host.cpp
```diff
@@ -92,16 +94,28 @@ void run_test(oe_enclave_t* enclave, int test_type)
...
             filename,
             "./cert_%s.der",
             test_type == TEST_RSA_KEY ? "rsa" : "ec");
+#endif
         OE_TRACE_INFO(
             "Host: Log quote embedded certificate to file: [%s]\n", filename);
+#ifdef _WIN32
+        fopen_s(&file, filename, "wb");
+#else
         file = fopen(filename, "wb");
```
jhand2: same, just use fopen_s always

tests/crypto/enclave/host/host.c
```diff
@@ -1,6 +1,9 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+// Suppress warnings against using _open() instead of _sopen_s().
+#pragma warning(disable : 4996)
```
jhand2: May need win32 ifdef

common/oe_host_string.h
ryanhsu19: Yes, you're right. I'll change it so that it will only apply to Windows.

tests/attestation_cert_apis/host/host.cpp
ryanhsu19: Tried and found that fopen_s() does a good job anywhere except in another file, tests/crypto/read_file.c. It does not help with <stdio.h> explicitly included. I'm wondering: is there anything particular with this file? Here's an error message:
-----
[1/2] Building C object tests\crypto\enclave\enc\CMakeFiles\cryptoenc.dir\__\__\read_file.c.obj
FAILED: tests/crypto/enclave/enc/CMakeFiles/cryptoenc.dir/__/__/read_file.c.obj 
"C:/Program Files/Git/bin/bash.exe" "C:/Users/ryanhsu20/ver2/openenclave/scripts/clangw" "-DMBEDTLS_CONFIG_FILE="<config.h>" -DOE_API_VERSION=2 -DOE_BUILD_ENCLAVE -DOE_LINK_SGX_DCAP_QL -Itests\crypto\enclave\enc -I..\..\3rdparty\mbedtls\mbedtls\include -I3rdparty\mbedtls -I..\..\include -Ioutput\include\openenclave\libc /DWIN32 /D_WINDOWS /wd4566 /wd4200 /MD /O2 /Ob2 /DNDEBUG   /WX /W3 -Wall -Werror -Wpointer-arith -Wconversion -Wextra -Wno-missing-field-initializers -fno-strict-aliasing -mxsave -fno-builtin-malloc -fno-builtin-calloc -fno-builtin -mllvm -x86-speculative-load-hardening -m64 -fPIE -nostdinc -fstack-protector-strong -fvisibility=hidden -fno-omit-frame-pointer -ffunction-sections -fdata-sections -ftls-model=local-exec /showIncludes -o tests\crypto\enclave\enc\CMakeFiles\cryptoenc.dir\__\__\read_file.c.obj -c ..\..\tests\crypto\read_file.c"
..\..\tests\crypto\read_file.c:13:5: error: implicit declaration of function 'fopen_s' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    fopen_s(&cfp, filename, "rb");
    ^
..\..\tests\crypto\read_file.c:13:5: note: did you mean 'fopen'?
output\include\openenclave\libc\stdio.h:64:7: note: 'fopen' declared here

jhand2: ah, tests/crypto runs in both enclave and host, so the enclave version will fail to build. Go ahead and just use an ifdef here.

ryanhsu19: Got it, thanks!

host/traceh.c
```diff
@@ -313,14 +316,16 @@ void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
 #if defined(__linux__)
     struct timeval time_now;
     gettimeofday(&time_now, NULL);
-    struct tm* t = gmtime(&time_now.tv_sec);
+    struct tm t;
+    gmtime_r(&time_now.tv_sec, &t);
 #else
     time_t lt = time(NULL);
-    struct tm* t = gmtime(&lt);
+    struct tm t;
```
CodeMonkeyLeet: Nit: move common declarations outside of build-specific branches.

host/windows/syscall.c
```diff
@@ -919,7 +921,7 @@ ssize_t oe_syscall_send_ocall(
     ssize_t ret;
     _set_errno(0);
 
-    ret = send(_get_socket(sockfd), buf, len, flags);
+    ret = send(_get_socket(sockfd), buf, (int)len, flags);
```
CodeMonkeyLeet: This looks like it needs an arithmetic check that the downcast from unsigned 64-bit `size_t` to signed 32-bit `int` does not result in truncation.

host/windows/syscall.c
```diff
@@ -942,7 +944,7 @@ ssize_t oe_syscall_sendto_ocall(
     ret = sendto(
         _get_socket(sockfd),
         buf,
-        len,
+        (int)len,
```
CodeMonkeyLeet: This needs a downcast check

tests/attestation_cert_apis/host/host.cpp
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+#define FILENAME_LEN 80
```
CodeMonkeyLeet: Per dev guidelines, prefer verbose names over abbreviations and acronyms: `FILENAME_LENGTH`

tests/crypto/enclave/host/host.c
```diff
@@ -1,6 +1,11 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+// Suppress warnings against using _open() instead of _sopen_s().
```
CodeMonkeyLeet: Can you elaborate on the reason for this? Is this something that we're choosing not to fix or only deferring to fix?

tests/libcxx/host/host.cpp
```diff
@@ -280,7 +280,11 @@ int main(int argc, const char* argv[])
     }
 
     // Disable stdout buffering on host
+#ifdef _WIN32
+    setvbuf(stdout, NULL, _IONBF, 0);
```
CodeMonkeyLeet: Why not just use `setvbuf` in both cases here?

tests/mbed/host/ocalls.c
```diff
@@ -1,6 +1,11 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+// Suppress warnings against using _open() instead of _sopen_s().
+#ifdef _WIN32
+#pragma warning(disable : 4996)
```
CodeMonkeyLeet: If we're already disabling 4996, why bother with replacing the POSIX aliases with the Windows implementation-specific names (i.e. `_` prefixed names)?

tests/oeedger8r/host/teststring.cpp
```diff
@@ -226,7 +226,7 @@ void test_string_edl_ecalls(oe_enclave_t* enclave)
     const char* str_value = "Hello, World\n";
 
     char str[50];
```
CodeMonkeyLeet: Consider defining a constant for this magic number.

tests/pingpong-shared/host/host.cpp
```diff
@@ -24,7 +24,7 @@ void Pong(const char* in, char* out)
         {
             got_pong = true;
         }
-        strcpy(out, in);
+        strcpy_s(out, 128, in);
```
CodeMonkeyLeet: Where is this magic number 128 coming from? Consider adding a shared header for the enc & host and referencing constants from it, here and in the other files where this pattern occurs.

tools/oesign/getopt_long.c
```diff
@@ -48,7 +50,13 @@ char* optarg;     /* argument associated with option */
 #define IGNORE_FIRST (*options == '-' || *options == '+')
 #define PRINT_ERROR \
     ((opterr) && ((*options != ':') || (IGNORE_FIRST && options[1] != ':')))
+#ifdef _WIN32
+#define IS_POSIXLY_CORRECT                                       \
+    (!_dupenv_s(&dupenv_buf, &dupenv_size, "POSIXLY_CORRECT") && \
```
CodeMonkeyLeet: Same comment as debugrt/host, can we consolidate on `oe_dupenv()` instead of branching the code here?

host/windows/syscall.c
ryanhsu19: Yes, I added it because the compiler complained about this. I will see if I can add some checks before it.

tests/libcxx/host/host.cpp
ryanhsu19: Ah I missed this one. Will update it shortly.

CodeMonkeyLeet:
@jhand2 @ryanhsu19 
>@CodeMonkeyLeet I know you've mentioned before that we mostly don't care to use _s functions. I would prefer we pick one of the following:
>
> - Always use _s functions on the host for both Windows and Linux
> - Don't use _s functions on the host (when it is safe to use the alternative), simply suppress the warning.
>
> Do you have a preference between the two? Pretty much all _s functions are supported on Linux in C11.

For context, the OE SDK code has traditionally tried to use idiomatic C code rather than the MS extensions to C, which never really seemed to catch out outside of MS. I'm would prefer converged code rather than `#ifdef` littered through the code, so if `_s` versions are also supported on Intel, we can use them.

Elsewhere in the code, we've also introduced thin `oe_` wrappers to map to OS differences (most notably in the common code dealing with crypto) and we have examples like `oe_dupenv` ... we should probable consider a consistent strategy here; I'll note that it gets a bit confusing because we don't distinguish between the enclave & host versions of some of these, and this could be even more so with how the syscall headers on the enclave side already define `oe_open` for different reasons.

tests/pingpong-shared/host/host.cpp
ryanhsu19: Found the number in the declaration for the static buffer size when I was trying to determine the size for strcpy_s(). I will define a constant first.

tests/crypto/enclave/host/host.c
ryanhsu19: I have trouble locating the correct flags and modes for _sopen_s() but I will keep trying.

debugger/debugrt/host/host.c
ryanhsu19: getenv_s() works fine in this update, and if oe_dupenv() works I'll update it again.

tests/crypto/enclave/host/host.c
ryanhsu19: Temporarily restored to the original version with C4996 disabled because `_sopen_s()` did not seem to work well. In my experiment, I changed

 - `return open(path, flags, mode);`

to

 - `_sopen_s(&fd, path, flags, _SH_DENYNO, mode);`

where I declared `int fd;` beforehand and returned it afterwards.

tests/mbed/host/ocalls.c
ryanhsu19: I tried `_sopen_s()` and it did not seem to work well so I just restored to the original version.

debugger/debugrt/host/host.c
jhand2: getenv_s is fine. Thanks!

common/oe_host_string.h
```diff
@@ -1,6 +1,14 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+/*
+ * There are separate OE functions for strerror and strerror_s.
+ * Therefore we suppress the warning about not using more secure CRT functions.
+ */
+#ifdef _WIN32
+#pragma warning(disable : 4996)
```
dthaler: rather than disabling it all together, it's better to wrap the relevant parts (including this line) in
#pragma warning(push)
#pragma warning(pop)

common/datetime.c
```diff
@@ -229,20 +229,24 @@ oe_result_t oe_datetime_now(oe_datetime_t* value)
...
     time(&now);
-    timeinfo = gmtime(&now);
-
-    value->year = (uint32_t)timeinfo->tm_year + 1900;
-    value->month = (uint32_t)timeinfo->tm_mon + 1;
-    value->day = (uint32_t)timeinfo->tm_mday;
-    value->hours = (uint32_t)timeinfo->tm_hour;
-    value->minutes = (uint32_t)timeinfo->tm_min;
-    value->seconds = (uint32_t)timeinfo->tm_sec;
+#ifdef _WIN32
```
dthaler: this file uses ifdef _WIN32
traceh.c instead uses #if defined(__linux__)
Be consistent.  Even better would be to add something like:

#ifdef _WIN32
#define gmtime_r(now, timeinfo) gmtime_s(timeinfo, now)
#endif

to internal/datetime.h and include that from traceh.c (it's already included from this file),
and then just use gmtime_r in both files.

host/windows/syscall.c
```diff
@@ -1488,13 +1502,14 @@ int oe_syscall_uname_ocall(struct oe_utsname* buf)
     // OE SDK is supported only on WindowsServer and Win10
     if (IsWindowsServer())
     {
-        sprintf(buf->sysname, "WindowsServer");
-        sprintf(buf->version, "2016OrAbove");
+        sprintf_s(buf->sysname, sizeof("WindowsServer"), "WindowsServer");
```
dthaler: the second argument to sprintf_s should be the size of the buffer (buf->sysname), not the size of the 3rd argument ("WindowsServer").  Compare line 1485 and 1493, which are correct.

tests/attestation_cert_apis/host/host.cpp
```diff
@@ -92,16 +94,17 @@ void run_test(oe_enclave_t* enclave, int test_type)
...
     {
         // for testing purpose, output the whole cer in DER format
-        char filename[80];
+        char filename[FILENAME_LENGTH];
         FILE* file = NULL;
 
-        sprintf(
+        sprintf_s(
             filename,
+            FILENAME_LENGTH,
```
dthaler: safer to do sizeof(filename)

tests/attestation_cert_apis/host/host.cpp
```diff
@@ -92,16 +94,17 @@ void run_test(oe_enclave_t* enclave, int test_type)
...
             "./cert_%s.der",
             test_type == TEST_RSA_KEY ? "rsa" : "ec");
         OE_TRACE_INFO(
             "Host: Log quote embedded certificate to file: [%s]\n", filename);
-        file = fopen(filename, "wb");
+        fopen_s(&file, filename, "wb");
         fwrite(cert, 1, cert_size, file);
```
dthaler: A comment on pre-existing code: this isn't checking for fopen failure, which seems suspect.  This PR is ok though since it's no worse than before.

tests/crypto/read_file.c
```diff
@@ -8,7 +8,12 @@
 oe_result_t read_cert(char* filename, char* cert)
 {
     size_t len_cert;
+#ifdef _WIN32
```
dthaler: To Simon's point, I would prefer not seeing ifdefs scattered around every file and since fopen_s is part of the standard (see https://en.cppreference.com/w/c/io/fopen) we should just use fopen_s.
For any platform that doesn't support it, we should have in a .h file a macro that maps it to fopen.

tests/oeedger8r/host/teststring.cpp
```diff
@@ -225,8 +227,8 @@ void test_string_edl_ecalls(oe_enclave_t* enclave)
 {
     const char* str_value = "Hello, World\n";
 
-    char str[50];
-    sprintf(str, "%s", str_value);
+    char str[STR_LENGTH];
+    sprintf_s(str, STR_LENGTH, "%s", str_value);
```
dthaler: change STR_LENGTH to sizeof(str)

tests/pingpong-shared/host/host.cpp
```diff
@@ -51,7 +53,7 @@ OE_EXPORT int main_shared(int argc, const char* argv[])
         return 1;
     }
 
-    strcpy(buf, "String2");
+    strcpy_s(buf, BUF_LENGTH, "String2");
```
dthaler: change BUF_LENGTH to sizeof(buf)

tests/pingpong/host/host.cpp
```diff
@@ -24,11 +26,11 @@ void Pong(const char* in, char* out)
         {
             got_pong = true;
         }
-        strcpy(out, in);
+        strcpy_s(out, BUF_LENGTH, in);
```
dthaler: This looks dangerous since the API in line 19 does not guarantee that out is BUF_LENGTH.  Instead change the prototype so the length is a 3rd parameter in line 19.

dthaler:
Great work so far.  Just some comments to make it be even safer, but love to see this happening.

tests/crypto/read_file.c
ryanhsu19: It's not standard in C99, and both enclave and host use this. I'll try to integrate the ifdef _WIN32 stuff into the associated header file, readfile.h.

jhand2: Hmmm. From our discussion on https://github.com/openenclave/openenclave/issues/2376 it looks like we decided that samples and public headers should be C99 compliant. @dthaler thoughts on test code as well?

But to Ryan's point, inside the enclave _s functions are not supported and this test runs on both host and enclave, so in this particular instance we can't use just `fopen_s`

dthaler: Why is fopen_s not supported in the enclave?  Let's fix that.

Personally I don't think test code needs to be C99 compliant since test code shouldn't appear in arbitrary apps compiled with arbitrary compilers (that's what sample code is for).

jhand2: > Why is fopen_s not supported in the enclave? Let's fix that.

No _s functions are supported. I think it was a specific design decision, but I wasn't around at the time so I can really speak to the reasoning/point to a gh issue. I have heard something along the lines of that these APIs are mostly regarded in the Linux community as not very useful.

Is your thought that all the secure C11 functions should be available, or just specific ones?

> Personally I don't think test code needs to be C99 compliant since test code shouldn't appear in arbitrary apps compiled with arbitrary compilers (that's what sample code is for).

I agree. Thanks for the input.

tests/pingpong/host/host.cpp
ryanhsu19: Thanks! I've changed the one in tests/pingpong-shared as well!

tests/crypto/read_file.c
ryanhsu19: @jhand2 @dthaler I've changed the other things but this one proves to take more time than I'd thought. They take different arguments in different formats so I will need to write another wrapper to handle `fopen()`/`fopen_s()`, as well as `sscanf()/sscanf_s()` near line 125. The compiler complained even though I rewrote it using `__STDC_VERSION__` macros.
and the compiler still found a way to complain about all the `fopen_s()` and `sscanf_s()`.

On the other hand, I've revised the other issues accordingly, and it surely looks neat after all the changes. Are there any other concerns?

common/oe_host_string.h
```diff
@@ -1,6 +1,11 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+/*
+ * There are separate OE functions for strerror and strerror_s.
+ * Therefore we suppress the warning about not using more secure CRT functions.
+ */
```
jhand2: nit: Move this down to where oe_strerror is

Good change though, I didn't realize it was only one funciton causing the warning.

debugger/debugrt/host/host.c
```diff
@@ -47,13 +47,24 @@ static bool raise_debugger_events()
...
+            getenv_s(&getenv_length, NULL, 0, "OE_DEBUGGER_CONTRACT_VERSION");
+            if (getenv_length > 0)
             {
-                int v = 0;
-                if (sscanf(version, "%d", &v) == 1)
+                char* version = (char*)malloc(sizeof(char) * getenv_length);
+                getenv_s(
+                    &getenv_length,
+                    version,
+                    sizeof("OE_DEBUGGER_CONTRACT_VERSION"),
```
jhand2: This should be the size of `version` right? (`sizeof(char) * getenv_length`)

tests/crypto/enclave/host/host.c
jhand2: It looks like this wasn't fixed yet, reopening

dthaler:
> I would prefer we pick one of the following:
> Always use _s functions on the host for both Windows and Linux
> Don't use _s functions on the host (when it is safe to use the alternative), simply suppress the warning.
> 
> Do you have a preference between the two? Pretty much all _s functions are supported on Linux in C11.
> 
> For context, the OE SDK code has traditionally tried to use idiomatic C code rather than the MS extensions to C, which never really seemed to catch out outside of MS. I'm would prefer converged code rather than #ifdef littered through the code, so if _s versions are also supported on Intel, we can use them.

Any _s versions that are part of the C11 standard should be used (they're not "MS extensions to C").  fopen_s (https://en.cppreference.com/w/c/io/fopen), printf_s( https://en.cppreference.com/w/c/io/fprintf) etc are all part of the C11 standard.

tests/crypto/read_file.c
dthaler: @jhand2 I think all the secure standard C11 functions should be available in an enclave.  One has to be able to write apsp that compile with a C99 compliant compiler, but that doesn't mean C11 functions cannot be available.

common/oe_host_string.h
```diff
@@ -38,11 +43,18 @@ int oe_strncmp(const char* s1, const char* s2, size_t n)
 
 /* host already has an oe_strlcat implementation */
 
+#if defined(_WIN32)
```
dthaler: I think that this should be checking if defined(_MSC_VER) not _WIN32, since I think the warnings are compiler-specific, not platform-specific.

host/traceh.c
```diff
@@ -310,17 +314,18 @@ oe_result_t oe_log(oe_log_level_t level, const char* fmt, ...)
...
 #if defined(__linux__)
     struct timeval time_now;
     gettimeofday(&time_now, NULL);
-    struct tm* t = gmtime(&time_now.tv_sec);
+    gmtime_r(&time_now.tv_sec, &t);
 #else
     time_t lt = time(NULL);
-    struct tm* t = gmtime(&lt);
+    gmtime_s(&t, &lt);
```
dthaler: Should not need two #if cases here.  Include internal/datetime.h which games gmtime_r ubiquitous.

tests/crypto/read_file.c
```diff
@@ -121,7 +152,11 @@ oe_result_t read_dates(char* filename, oe_datetime_t* time)
     }
     buffer[len_date] = '\0';
 
+#ifdef _WIN32
```
dthaler: See comments on gmtime_r etc.  This should be able to use sscanf_s and on platforms where it's absent, a macro can map sscanf_s to sscanf.

tests/echo/host/host.c
```diff
@@ -17,7 +17,7 @@ int host_echo(char* in, char* out, char* str1, char* str2, char str3[100])
     OE_TEST(strcmp(str2, "oe_host_strdup2") == 0);
     OE_TEST(strcmp(str3, "oe_host_strdup3") == 0);
 
-    strcpy(out, in);
+    strcpy_s(out, 100, in);
```
dthaler: Hard coding 100 here is dangerous, since if someone later changes the caller's length there might be memory corruption.  Safer to have the caller pass in the length.

tests/libunwind/host/host.cpp
```diff
@@ -106,7 +106,11 @@ int main(int argc, const char* argv[])
     {
         oe_put_err("oe_create_libunwind_enclave(): result=%u", result);
     }
+#ifdef _WIN32
```
dthaler: I would like to not have any ifdef here.

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/posix-getpid?view=vs-2019 explains:
"The Microsoft-implemented POSIX function name getpid is a deprecated alias for the _getpid function. By default, it generates Compiler warning (level 3) C4996. The name is deprecated because it doesn't follow the Standard C rules for implementation-specific names. However, the function is still supported."

So one approach is to just violate posix and use getpid in both windows and Linux.
Another approach would be to follow posix guidelines and use _getpid in both cases, and if some platform (like Linux) violates posix guidelines then that platform can have a macro in a .h file that redefines _getpid to their posix-violating name (getpid).
Another approach would be to define oe_getpid (since there is no standard) and map it in a .h file to different names on different platforms.
But in any case the point is to define one common name in a .h file and make .c and .cpp files avoid platform specific ifdefs.

tests/mbed/host/host.c
```diff
@@ -31,7 +34,7 @@ char* find_data_file(char* str, size_t size)
         return token;
     }
 
-    strncat(str, tail, strlen(tail));
+    strncat_s(str, TEMP_LEN, tail, strlen(tail));
```
dthaler: Dangerous to hard code length here. Safer to make the caller pass in the size.

tests/mbed/host/host.c
```diff
@@ -129,7 +132,7 @@ int main(int argc, const char* argv[])
 
     printf("=== %s: %s\n", argv[0], argv[1]);
 
-    strcpy(temp, argv[1]);
+    strcpy_s(temp, TEMP_LEN, argv[1]);
```
dthaler: sizeof(temp) instead of TEMP_LEN.  Avoid using constants as the 2nd argument to strcpy_s throughout.

tests/ocall/host/host.cpp
```diff
@@ -77,7 +77,11 @@ int main(int argc, const char* argv[])
     /* Call enc_set_tsd */
     {
         int ret_val = -1;
+#ifdef _WIN32
+        result = enc_set_tsd(enclave, &ret_val, _strdup("TSD-DATA"));
```
dthaler: Same comment about this ifdef as for getpid, since this case is the same, where MSDN explains:
"The Microsoft-implemented POSIX function names strdup and wcsdup are deprecated aliases for the _strdup and _wcsdup functions. By default, they generate Compiler warning (level 3) C4996. The names are deprecated because they don't follow the Standard C rules for implementation-specific names. However, the functions are still supported."

tests/pingpong/host/host.cpp
```diff
@@ -51,7 +53,7 @@ int main(int argc, const char* argv[])
         return 1;
     }
 
-    strcpy(buf, "String2");
+    strcpy_s(buf, BUF_LENGTH, "String2");
```
dthaler: sizeof(buf)

tests/report/common/tests.cpp
```diff
@@ -1294,17 +1294,21 @@ void test_verify_report_with_collaterals()
 
         /* Test with time in the past */
         time_t t;
-        struct tm* timeinfo;
+        struct tm timeinfo;
         time(&t);
-        timeinfo = gmtime(&t);
+#ifdef _WIN32
+        gmtime_s(&timeinfo, &t);
```
dthaler: See earlier comments about gmtime_r

tests/echo/host/host.c
ryanhsu19: The 100 is hard-coded in the EDL file. I will change it to sizeof(out) first.

host/traceh.c
ryanhsu19: But in the `if` part, there is no `gettimeofday()` for Windows. Or should I just stick to the `else` part?

tests/crypto/enclave/host/host.c
ryanhsu19: Oops, just fixed it

tests/mbed/host/host.c
ryanhsu19: Just tried changing all the second argument to `sizeof(...)` but all the mbed tests failed. So I leave this as it is for now.

jhand2: huh. sizeof(temp) should be the same as TEMP_LEN. What exactly fails?

ryanhsu19: I just tried again. It seems that if I only change the one in main() it will be fine. Otherwise ctest will fail with an exception code 0xc0000409.

tests/mbed/host/host.c
jhand2: @ryanhsu19 you should use the `size` variable here

tests/mbed/host/host.c
```diff
@@ -40,11 +43,11 @@ void datafileloc(char* data_file_name, char* path)
 {
     char* tail = "../enc/";
 #ifdef PROJECT_DIR
-    strcpy(path, PROJECT_DIR);
+    strcpy_s(path, PATH_LEN, PROJECT_DIR);
```
jhand2: caller of datafileloc should pass in size of `path` buffer

tests/mbed/host/host.c
jhand2: This is the only place where the size of the buffer is known by the compiler. I commented elsewhere on what you should do instead

tests/mbed/host/host.c
ryanhsu19: Great, this is exactly what I need! I just updated and will mark this as resolved.

host/traceh.c
```diff
@@ -201,8 +202,11 @@ static bool _escape_characters(
                         log_msg_escaped[idx] = '\0';
                         return false;
                     }
-                    sprintf(
-                        (char*)&log_msg_escaped[idx], "u%04hhx", log_msg[i]);
+                    sprintf_s(
+                        (char*)&log_msg_escaped[idx],
+                        msg_size,
```
jhand2: I think this should be `msg_size - idx` right? Because that is whats left of the buffer

host/traceh.c
```diff
@@ -310,23 +314,14 @@ oe_result_t oe_log(oe_log_level_t level, const char* fmt, ...)
...
-#if defined(__linux__)
-    struct timeval time_now;
-    gettimeofday(&time_now, NULL);
-    struct tm* t = gmtime(&time_now.tv_sec);
-#else
+    struct tm t;
     time_t lt = time(NULL);
-    struct tm* t = gmtime(&lt);
-#endif
+    gmtime_r(&lt, &t);
```
jhand2: Doesn't gmtime_r not work on Windows? Or is this function only ever called from the enclave?

ryanhsu19: I created a macro in a header file so that it would change it to gmtime_s on Windows.

common/datetime.c
dthaler: Unresolving since not done yet in diffs.

host/traceh.c
jhand2: I would prefer a wrapper function instead (like oe_gmtime_r) and do an ifdef there. I think a macros hides too much of what is actually going on

tests/crypto/enclave/host/host.c
```diff
@@ -1,8 +1,13 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+#if defined(_WIN32)
+#pragma warning(disable : 4996)
```
dthaler: Can this be moved down to whatever it's trying to wrap, surrounded by a push/pop?

tests/crypto/read_file.c
```diff
@@ -5,10 +5,22 @@
 #include <openenclave/corelibc/string.h>
 #include "readfile.h"
 
+FILE* read_file(const char* filename, const char* mode)
+{
+    FILE* file;
+#ifdef _MSC_VER
+    fopen_s(&file, filename, mode);
+#else
+	file = fopen(filename, mode);
```
dthaler: nit: fix indentation here and in 16

tests/crypto/readfile.h
```diff
@@ -22,6 +22,10 @@
 #define max_date_elements 6
 #define max_coordinates_size 32
 
+#ifdef _MSC_VER
```
dthaler: I am not sure but I think this should check _WIN32 instead of _MSC_VER since I think sscanf_s is platform specific not compiler specific.

tests/echo/enc/enc.c
```diff
@@ -42,7 +42,8 @@ int enc_echo(char* in, char out[100])
         out,
         "oe_host_strdup1",
         host_allocated_str,
-        stack_allocated_str);
+        stack_allocated_str,
+		100);
```
dthaler: sizeof(stack_allocated_str)

tests/echo/host/host.c
```diff
@@ -11,13 +11,13 @@
 #include "../../../host/strings.h"
 #include "echo_u.h"
 
-int host_echo(char* in, char* out, char* str1, char* str2, char str3[100])
+int host_echo(char* in, char* out, char* str1, char* str2, char* str3, int length)
```
dthaler: Just a suggestion: rename length to out_length to make it clearer what it's the length of.

tests/mbed/host/ocalls.c
```diff
@@ -1,8 +1,11 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+#pragma warning(disable : 4996)
```
dthaler: Can this line be put inside a push/pop surrounding whatever code is affected?

tests/pingpong-shared/enc/enc.cpp
```diff
@@ -6,7 +6,7 @@
 
 void Ping(const char* in, char* out)
 {
-    Pong(in, out);
+    Pong(in, out, sizeof(out));
```
dthaler: BUG: this line is incorrect since sizeof(out) is sizeof(char*) not size of the buffer.  Instead, have the caller pass in the length as a third argument to Ping()

tests/pingpong/enc/enc.cpp
```diff
@@ -6,7 +6,7 @@
 
 void Ping(const char* in, char* out)
 {
-    Pong(in, out);
+    Pong(in, out, sizeof(out));
```
dthaler: BUG: this line is incorrect since sizeof(out) is sizeof(char*) not size of the buffer. Instead, have the caller pass in the length as a third argument to Ping()

tests/report/host/host.cpp
```diff
@@ -197,15 +198,19 @@ int main(int argc, const char* argv[])
 
     // Get current time and pass it to enclave.
     std::time_t t = std::time(0);
-    std::tm* tm = std::gmtime(&t);
-
+    std::tm tm;
+#ifdef _WIN32
+    gmtime_s(&tm, &t);
```
dthaler: Can this ifdef go away due to defining a macro?

tests/stack_smashing_protector/host/host.cpp
```diff
@@ -32,7 +32,12 @@ int main(int argc, const char* argv[])
     /* Call enc_set_thread_variable */
     {
         int ret_val = -1;
+#ifdef _WIN32
+        result =
+            enc_set_thread_variable(enclave, &ret_val, _strdup("TSD-DATA"));
```
dthaler: Can this ifdef go away due to defining a _strdup macro?

tests/tools/oecert/host/host.cpp
```diff
@@ -200,8 +200,7 @@ static oe_result_t _gen_report(
             uint8_t* collaterals = NULL;
             size_t collaterals_size = 0;
             oe_report_header_t* header = (oe_report_header_t*)remote_report;
-
-            sprintf(collateral_filename, "%s.col", report_filename);
+            sprintf_s(collateral_filename, 1024 + 1, "%s.col", report_filename);
```
dthaler: Change 1024+1 to sizeof(collateral_filename)

tests/crypto/readfile.h
ryanhsu19: If I change to _WIN32 then there will be a C99/C11 compilation problem.

host/traceh.c
dthaler: This is a matter of taste, but I have the opposite preference from @jhand2 for anything that's part of the standard.  https://en.cppreference.com/w/c/chrono/gmtime shows that gmtime_r is part of the C11 standard.

tests/crypto/readfile.h
dthaler: Ok thanks for checking.  Resolving this comment.

host/traceh.c
jhand2: Ah, didn't realize gmtime_r was a part of the new standard (not C11 though, it shows as C2x). I don't have a strong preference either way then

tests/pingpong/enc/enc.cpp
ryanhsu19: Done with the change for both pingpong and pingpong-shared.

host/traceh.c
jhand2: Resolving this, I think it's fine to leave as-is

tests/mbed/host/ocalls.c
```diff
@@ -13,20 +14,27 @@
...
 int mbed_test_close(int fd)
 {
-    return close(fd);
+    return _close(fd);
 }
 
 int mbed_test_lseek(int fd, int offset, int whence)
 {
-    return lseek(fd, offset, whence);
+    return _lseek(fd, offset, whence);
```
jhand2: Does this work on linux?

tests/pingpong-shared/pingpong.edl
```diff
@@ -1,18 +1,20 @@
...
+            [in, out, string] char* out,
+			int out_length);
     };
 
     untrusted {
         void Pong(
             [in, string] const char* in,
-            [in, out, string] char* out);
+            [in, out, string] char* out,
+			int out_length);
```
jhand2: Nit: use same indentation as the rest of the file (spaces, not tabs)

jhand2: this should be marked wit the [in] annotation

tests/pingpong-shared/pingpong.edl
```diff
@@ -1,18 +1,20 @@
...
-
+            [in, out, string] char* out,
+			int out_length);
     };
 
     untrusted {
         void Pong(
             [in, string] const char* in,
-            [in, out, string] char* out);
+            [in, out, string] char* out,
```
jhand2: this should have the count=out_length annotation

tests/pingpong-shared/pingpong.edl
```diff
@@ -1,18 +1,20 @@
...
 // Licensed under the MIT License.
 
 enclave {
     trusted {
         public void Ping(
             [in, string] const char* in,
-            [in, out, string] char* out);
-
+            [in, out, string] char* out,
+			int out_length);
```
jhand2: same as below

tests/pingpong-shared/pingpong.edl
ryanhsu19: Ah I see! No wonder the format-code does not pick out the problem. I'll change it as well as the settings in my editor.

jhand2:
bors try

tests/mbed/host/ocalls.c
ryanhsu19: Did a quick research and found it might not. I'll revert the change.

tests/pingpong-shared/pingpong.edl
ryanhsu19: There would be "unexpected pointer attributes for `int'" error after I added it.

jhand2: oh, actually that makes sense. An annotation wouldn't be needed for integral types. Go ahead and ignore :smile:

tests/mbed/host/ocalls.c
ryanhsu19: I think I'll just keep the #pragma stuff on the top because it applies to all the four functions in this file.

tests/mbed/host/ocalls.c
ryanhsu19: It applies to all the four functions so I'll just keep it here.

ryanhsu19:
I've decided to keep some #ifdefs for secure CRT functions after reading the previous bors log, and most of them are in the following header files:

 - include/openenclave/host.h

 - include/openenclave/internal/datetime.h

 - tests/crypto/readfile.h

ryanhsu19:
bors try

include/openenclave/host.h
```diff
@@ -29,6 +29,24 @@
...
+#define strcpy_s(destination, destination_size, source) \
+    ;                                                   \
+    {                                                   \
+        (void)(destination_size);                       \
+        strcpy(destination, source);                    \
+    }
+#define _strdup strdup
+#define strncat_s(destination, destination_size, source, source_size) \
+    strncat(destination, source, source_size)
+#endif
```
jhand2: Huh. This kinda concerns me since on Linux it looks like secure APIs are being used but they aren't.

jhand2:
> Blocking so we can discuss. Does anyone know why the _s functions wouldn't be defined in headers on the host? They should be part of the C11 standard.

Did some quick research and it turns out glibc doesn't actually support _s functions even though they're part of the C standard so that's fun.

@CodeMonkeyLeet @anakrish @mingweishih Does anyone have any thoughts here? I'm concerned about code obfuscation where on linux it looks like one function is being called when really it's another.

include/openenclave/host.h
ryanhsu19: Previously bors complained about some _s functions, so I started to try it on Ubuntu 18.04 and RHEL 8 afterwards and found that they did not accept them either.

jhand2:
Removing my blocker because I think merging as-is is better than what was there before and we can always tweak the approach later. Would still love any input others have though.

mingweishih:
> > Blocking so we can discuss. Does anyone know why the _s functions wouldn't be defined in headers on the host? They should be part of the C11 standard.
> 
> Did some quick research and it turns out glibc doesn't actually support _s functions even though they're part of the C standard so that's fun.
> 
> @CodeMonkeyLeet @anakrish @mingweishih Does anyone have any thoughts here? I'm concerned about code obfuscation where on linux it looks like one function is being called when really it's another.

(I always thought that _s functions are Windows specific.... )
One option to avoid this concern is something like
```
#ifndef _WIN32
#define oe_sscanf sscanf
#else
#define oe_sscanf sscanf_s
#endif
```

dthaler:
> I think merging as-is is better than what was there before and we can always tweak things incrementally if needed

+1

> One option to avoid this concern is something like
> #ifndef _WIN32
> #define oe_sscanf sscanf
> #else
> #define oe_sscanf sscanf_s
> #endif

That sscanf and sscanf_s have different prototypes so that isn't possible.

ryanhsu19:
bors try

PR #2659, opened 2020-03-12
Document for removing DCAP style sgx driver in favour of isgx

yakman2020:
Some customers wish to use acc vms with the non-dcap driver, but we are in the process of providing them with the dcap stack as part of the kernel. Instructions on reversing that process with persistence.


docs/GettingStartedDocs/UseNonDCAPDriverInAcc.md
```diff
@@ -0,0 +1,53 @@
+# Using the Non-DCAP Driver in an ACC VM
+
+ACC VMs come pre equipeed with the azure DCAP stack preinstalled. This can interfere with non-azure dcap clients,
```
dthaler: typo "equipeed"
Also "azure" should be capitalized
Also expand acronyms (ACC, DCAP) on first use in the file.

docs/GettingStartedDocs/UseNonDCAPDriverInAcc.md
```diff
@@ -0,0 +1,53 @@
+# Using the Non-DCAP Driver in an ACC VM
+
+ACC VMs come pre equipeed with the azure DCAP stack preinstalled. This can interfere with non-azure dcap clients, 
+so in this case we must disable the dcap driver.
+
+To do so, first lets get the non-dcap driver (if we haven't already).
```
dthaler: typo: "lets" -> "let's"

docs/GettingStartedDocs/UseNonDCAPDriverInAcc.md
```diff
@@ -0,0 +1,53 @@
...
+```
+curl https://download.01.org/intel-sgx/sgx-linux/2.9/distro/ubuntu16.04-server/sgx_linux_x64_driver_2.6.0_95eaa6f.bin --output sgx_linux_x64_driver_2.6.0_95eaa6f.bin
```
dthaler: The title of the doc says "ACC VM", but this line is Linux specific.  Does that mean there are no Windows ACC VM's?

docs/GettingStartedDocs/UseNonDCAPDriverInAcc.md
```diff
@@ -0,0 +1,53 @@
+# Using the Non-DCAP Driver in an ACC VM
```
paulcallen: Can we be more specific here. I think it should be 'Using Intel Legacy Launch SGX Driver on ACC Linux VMs'

docs/GettingStartedDocs/UseNonDCAPDriverInAcc.md
paulcallen: Then I would say something like "ACC VMs come with a pre-installed Intel SGX DCAP driver. If you need to use the Intel Legacy Launch SGX driver instead you will need to disable this SGX DCAP driver first."

docs/GettingStartedDocs/UseNonDCAPDriverInAcc.md
paulcallen: We have code that always gets the latest version of the Intel DCAP driver. We should grab that and adapt it for the legacy launch version of the driver.

docs/GettingStartedDocs/UseNonDCAPDriverInAcc.md
```diff
@@ -0,0 +1,53 @@
...
+Now we can install the intel non-dcap driver.
+```
+sudo chmod a+x sgx_linux_x64_driver_2.6.0_95eaa6f.bin
+sudo ./sgx_linux_x64_driver_2.6.0_95eaa6f.bin
+sudo modprobe -i isgx
+```
+Now if we look at /dev, the device /dev/sgx will be gone and /dev/isgx in its place.
+
+But if we reboot the situation will be reverted.
+To make it permanent, we need to edit a couple of config files, on in /etc/modules.d/ and one in /etc/modprobe.d.
```
paulcallen: "on in"->one in

docs/GettingStartedDocs/UseNonDCAPDriverInAcc.md
```diff
@@ -0,0 +1,53 @@
...
+# /etc/modules: kernel modules to load at boot time.
+#
+# This file contains the names of kernel modules that should be loaded
+# at boot time, one per line. Lines beginning with "#" are ignored.
+
+intel_sgx
+isgx
+```
+
+comment out intel_sgx, so the lines look like
```
paulcallen: edit this file as root and ...

docs/GettingStartedDocs/UseNonDCAPDriverInAcc.md
```diff
@@ -0,0 +1,53 @@
...
+isgx
+```
+
+This will prevent the kernel from loading the module, but in addition you want to ensure it is not loaded. 
+intel_sgx will preempt isgx and prevent it from loading. So we need to blacklist it.
+
+To do this, we add a conf file to /etc/modprobe.d. 
+As superuser, create the file '/etc/modprobe.d/blacklist-intel-sgx.conf' with the contents:
+```
+blacklist-intel-sgx.conf
```
paulcallen: shouldnt this be "intel_sgx"?

docs/GettingStartedDocs/UseNonDCAPDriverInAcc.md
dthaler: Meta-question: do we want to have the main repo cover docs for all product environments?  (ACC and any other cloud providers in the future)   If yes, that's fine just want to have this agreed on now.  (The alternative would be for ACC and other cloud provider specific stuff to be in one or more separate repos.)

paulcallen: if you want to split it out I would prefer to have an ACC directory and have it under there, then other cloud providers have their own directory. Putting in another repo will be harder to find.

paulcallen: @dthaler lets leave this in the current directory and create a new issue about the possibility of splitting platform specific documentation out to platform specific directories, such as Azure, On-prem, and any other in the future

dthaler: @paulcallen Sounds good to me.  Created issue #2759

kuanggang-git:

radhikaj:
Closing. Please reopen when ready to work on it.

PR #2663, opened 2020-03-12
add acc contributor template with no sgx package install

BRMcLaren:
Template for setting up automated Linux regression testing for Intel Packages. Allow installation of all dependencies but not including the SGX packages.

### DIFF
--- a/scripts/ansible/oe-contributors-acc-setup-no-psw.yml
+++ b/scripts/ansible/oe-contributors-acc-setup-no-psw.yml
@@ -1,0 +1,19 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+---
+    - hosts: localhost
+      any_errors_fatal: true
+      become: yes
+      tasks:
+        - import_role:
+            name: linux/openenclave
+            tasks_from: environment-setup.yml
+
+        - import_role:
+            name: linux/intel
+            tasks_from: sgx-driver.yml
+
+        - import_role:
+            name: linux/az-dcap-client
+            tasks_from: stable-install.yml


BRMcLaren:
bors try

jhand2:
I don't really understand what this is for. Does "no sgx" mean "no psw"? Because it looks like the sgx driver is still installed.

BRMcLaren:
@jhand2 this is correct. It's a template that installs everything except the SGX packages, but including the driver.

The use case is to allow us to test against the Intel release candidate packages by setting up a machine with all the dependencies but the SGX packages.  Open to other suggestions, but this is what we have been doing manually up until now.

jhand2:
> @jhand2 this is correct. It's a template that installs everything except the SGX packages, but including the driver.
>
> The use case is to allow us to test against the Intel release candidate packages by setting up a machine with all the dependencies but the SGX packages. Open to other suggestions, but this is what we have been doing manually up until now.

That seems fine to me. My only suggestion would be to change the name to `oe-contributors-acc-setup-no-psw.yml` for clarity

BRMcLaren:
Sure thing.

Changed the file name for clarity.

BRMcLaren:
bors try

scripts/ansible/oe-contributors-acc-setup-no-psw.yml
```diff
@@ -0,0 +1,19 @@
...
+            name: linux/openenclave
+            tasks_from: environment-setup.yml
+
+        - import_role:
+            name: linux/intel
+            tasks_from: sgx-driver.yml
+
+        - import_role:
+            name: linux/az-dcap-client
+            tasks_from: stable-install.yml
```
CodeMonkeyLeet: Nit: missing newline

BRMcLaren: Fixed.

PR #2671, opened 2020-03-13
Add more specific messages for OE_TCB_LEVEL_INVALID

larrywxyz:
Made the OE_TCB_LEVEL_INVALID message (e.g. "Platform TCB (2) is not up-to-date") more specific. 

- If PCE SVN was not up to date, then return "Platform TCB (2) PCE SVN is not up-to-date". 
- If Comp SVN was not up to date, then return "Platform TCB (2) Comp SVN is not up-to-date". 
- If both were not up to date, then return "Platform TCB (2) Comp SVN and PCE SVN are not up-to-date".

### DIFF
--- a/common/sgx/tcbinfo.c
+++ b/common/sgx/tcbinfo.c
@@ -327,6 +327,9 @@ static void _determine_platform_tcb_info_tcb_level(
     if (platform_tcb_level->status.AsUINT32 != OE_TCB_LEVEL_STATUS_UNKNOWN)
         return;
 
+    bool comp_svn_out_of_date = false;
+    bool pce_svn_out_of_date = false;
+
     // Compare all of the platform's comp svn values with the corresponding
     // values in the current tcb level.
     for (uint32_t i = 0; i < OE_COUNTOF(platform_tcb_level->sgx_tcb_comp_svn);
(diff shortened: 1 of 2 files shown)


jhand2:
Please add DCO signoff to your commit :smile:

`git commit --amend --signoff`

common/sgx/tcbinfo.c
```diff
@@ -327,21 +327,33 @@ static void _determine_platform_tcb_info_tcb_level(
...
     {
         if (platform_tcb_level->sgx_tcb_comp_svn[i] <
             tcb_level->sgx_tcb_comp_svn[i])
-            return;
+            comp_svn_out_of_date = true;
     }
     if (platform_tcb_level->pce_svn < tcb_level->pce_svn)
+        pce_svn_out_of_date = true;
+
+    if (tcb_level->status.fields.up_to_date == 1)
```
CodeMonkeyLeet: Can you add a comment explaining why this check is necessary?  It's not obvious given the assumption of `tcb_level.status` is monotonically increasing on iterative calls to this function (as stated in the function comment). If `platform_tcb_level` previously failed to to clear the lower status levels, why don't we set the additional debug bits at the lower tiers of failure as well? (i.e. what is the condition under which the comparison against a higher tier `tcb_level` conflicts with something previously set by a failure at a lower tier?)

common/sgx/tcbinfo.h
```diff
@@ -48,6 +48,9 @@ typedef struct _oe_tcb_info_tcb_level
     uint16_t pce_svn;
     oe_tcb_level_status_t status;
 
+    bool comp_svn_out_of_date;
+    bool pce_svn_out_of_date;
```
yentsanglee: oe_tcb_info_tcb_level_t and oe_parsed_tcb_info_t exactly map to the the response specified in "GET TCB INFO V2" section in [https://api.portal.trustedservices.intel.com/documentation#pcs-tcb-info-v2](https://api.portal.trustedservices.intel.com/documentation#pcs-tcb-info-v2).

comp_svn_out_of_date and pce_svn_out_of_date are not defined in the spec. They do not belong in this struct.

common/sgx/tcbinfo.c
```diff
@@ -327,21 +327,33 @@ static void _determine_platform_tcb_info_tcb_level(
...
+    {
+        platform_tcb_level->comp_svn_out_of_date = comp_svn_out_of_date;
+        platform_tcb_level->pce_svn_out_of_date = pce_svn_out_of_date;
```
yentsanglee: platform_tcb_level are specifically defined in [https://api.portal.trustedservices.intel.com/documentation#pcs-tcb-info-v2](https://api.portal.trustedservices.intel.com/documentation#pcs-tcb-info-v2) and should not be altered. I think a better approaches are:
1. For this function to return different oe_result_t from the values in comp_svn_out_of_date and pce_svn_out_of_date.
2. Add extra output parameter(s) to this function.
and then deliver the result all the way up to oe_parse_tcb_info_json() where error messages are printed.

jazzybluesea: One possibility is to create a "verification error type" enum that would uniquely identify the reason why the SGX quote verification failed.

yentsanglee:
comp_svn_out_of_date and pce_svn_out_of_date alone are not enough to determine the error cases. Too much extra code just to display the error for debugging purpose. We've instead created PR 2708 to print out related TCB Info in test tool (oecertdump).

PR #2674, opened 2020-03-13
Change switchless initialization flag to uint64_t

jhand2:
The switchless call initialization flag is passed to oe_compare_exchange
as a `uint64_t *` when it is just a `bool *`. This can cause initialization
failures since the bool is interpreted incorrectly on some systems.

Signed-off-by: Jordan Hand <jorhand@microsoft.com>

### DIFF
--- a/enclave/core/sgx/switchlesscalls.c
+++ b/enclave/core/sgx/switchlesscalls.c
@@ -18,7 +18,15 @@ static oe_host_worker_context_t* _host_worker_contexts = NULL;
 // Flag to denote if switchless calls have already been initialized.
 static bool _is_switchless_initialized = false;
 
-static bool _switchless_init_in_progress = false;
+/* Flag to denote if switchless init function is already in progress
+ * _switchless_init_in_progress is defined as int64_t and not bool because
+ * we use oe_atomic_compare_and_swap() to manipulate its value. It only takes
+ * int64_t* unlike the type-generic built-in atomics
+ * __atomic_load_n()/__atomic_save_n(), so if _switchless_init_in_progress is
+ * not defined as int64_t, we would need to type-pun through an incompatible
+ * type which results in undefined behavior per the C spec.
+ * */
+static int64_t _switchless_init_in_progress = 0;
 
 /*
 **==============================================================================
@@ -57,9 +65,7 @@ oe_result_t oe_init_context_switchless_ecall(
     uint64_t contexts_size = 0;
 
     if (!oe_atomic_compare_and_swap(
-            (volatile int64_t*)&_switchless_init_in_progress,
-            (int64_t) false,
-            (int64_t) true))
+            &_switchless_init_in_progress, (int64_t) false, (int64_t) true))
     {
         OE_RAISE(OE_BUSY);
     }


jhand2:
> Not sure I understand the proposed change:
> 
> * There's no `oe_compare_exchange()` ... did you mean `oe_atomic_compare_and_swap()`?

Ya I meant `oe_atomic_compare_and_swap()`

>   
>   * If so, that natively takes `int64_t` not `uint64_t`.

My mistake, I meant to use int64_t

> * Why does changing the type here matter?
>   
>   * Elsewhere in `oe_init_context_switchless_ecall()`, the variable is still being set in terms of casts from `true` & `false` values.
>   * The use of this variable is strictly local and it seems internally consistent; it's not exposed for others to use?
>     
>     * Is `_is_switchless_initialized` the problem instead?

See this sample I wrote up:
```
#include <stdio.h>
#include <stdint.h>
#include <stdbool.h>
#include <inttypes.h>

void print_uint64_t(uint64_t *val) {
    printf("Value is: %"PRIu64"\n", *val);
}

int main() {
    int64_t a = 0xFFFFFFFFFFFFFFFF;
    bool b = true;
    print_uint64_t((uint64_t*)&b);
}
```

When I run this, it prints -255. Casting a bool* to an int64_t* is unsafe. I'm not sure why this check works at all today.

CodeMonkeyLeet:
@jhand2 I see what you're saying now ... could you add comment to the `_switchless_init_in_progress` declaration explaining the issue so future maintainers don't try to flip this back?

> `_switchless_init_in_progress` is defined as `int64_t` and not `bool` because we use `oe_atomic_compare_and_swap()` to manipulate its value. It only takes `int64_t*` unlike the type-generic built-in atomics `__atomic_load_n()`/`__atomic_save_n()`, so if `_switchless_init_in_progress` is not defined as `int64_t`, we would need to type-pun through an _incompatible type_ which results in undefined behavior per the C spec.

It's a good catch, and may have resulted from the change to remove the `no-strict-aliasing` build options.

jhand2:
Updated with comment

enclave/core/sgx/switchlesscalls.c
```diff
@@ -18,7 +18,15 @@ static oe_host_worker_context_t* _host_worker_contexts = NULL;
...
-static bool _switchless_init_in_progress = false;
+/* Flag to denote if switchless init function is already in progress
+ * _switchless_init_in_progress is defined as int64_t and not bool because
+ * we use oe_atomic_compare_and_swap() to manipulate its value. It only takes
+ * int64_t* unlike the type-generic built-in atomics
+ * __atomic_load_n()/__atomic_save_n(), so if _switchless_init_in_progress is
+ * not defined as int64_t, we would need to type-pun through an incompatible
+ * type which results in undefined behavior per the C spec.
+ * */
+static int64_t _switchless_init_in_progress = 0;
```
anakrish: Good catch!
It would be good to remove unnecessary casts whether this is used.

jhand2: Done

PR #2677, opened 2020-03-13
Allocate and associate ocall buffers with TCS.

### LINKED ISSUE #2590: Support zig-zag ocalls: enclave1-ecall -> ocall1 -> enclave2 ecall -> ocall2
With #2551, a context parameter is passed to ecalls. The context parameter contains a buffer that the ecall can use to allocate ocall parameters. This provides significant performance benefits as well as allows all ocalls to be implemented in EDL (maintainability and security).

Currently, the buffer is allocated in host's thread-local memory:
https://github.com/openenclave/openenclave/blob/89dba4c34e47e02383828c093c29ec35ef021f04/host/sgx/enter.c#L73

This is very efficient. However in the following call pattern:
   `enclave1-ecall -> ocall1 -> enclave2 ecall -> ocall2`

would result in ocall2 and ocall1 using the same buffer which would lead to incorrect results.

We need an efficient fix to handle this scenario.

anakrish:
Each tcs identifies an enclave thread. oe_thread_binding_t houses data
associated with an enclave thread, that is accessible by both host and
enclave. Naturally, ocall buffers can also be housed in the thread-binding.

ecall_context_t houses data shared between enclave thread and host for
making a single ecall. ecall_context_t uses the ocall buffer from the
thread-binding. ecall_context_t itself lives only as long as the ecall.

Note: If ever OE goes back to supporting reentrant ecalls and ocalls,
a single ocall buffer per tcs will not suffice.

Closes #2590 

### DIFF
--- a/host/sgx/enter.c
+++ b/host/sgx/enter.c
@@ -97,10 +97,29 @@ int __oe_host_stack_bridge(
 }
 
 /**
- * Thread specific OCALL buffers. Large enough for most ocalls.
+ * Size of ocall buffers passed in ecall_contexts. Large enough for most ocalls.
+ * If an ocall requires more than this size, then the enclave will make an
+ * ocall to allocate the buffer instead of using the ecall_context's buffer.
  * Note: Currently, quotes are about 10KB.
  */
-static __thread uint8_t _thread_ocall_buffer[16 * 1024];
+#define OE_DEFAULT_OCALL_BUFFER_SIZE (16 * 1024)
+
+/**
+ * Setup the ecall_context.
+ */
+OE_INLINE void _setup_ecall_context(oe_ecall_context_t* ecall_context)
+{
+    oe_thread_binding_t* binding = oe_get_thread_binding();
+    if (binding->ocall_buffer == NULL)
+    {
+        // Lazily allocate buffer for making ocalls. Bound to the tcs.
+        // Will be cleaned up by enclave during termination.
+        binding->ocall_buffer = malloc(OE_DEFAULT_OCALL_BUFFER_SIZE);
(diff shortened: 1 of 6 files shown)


tests/ecall_ocall/host/host.cpp
```diff
@@ -256,6 +256,36 @@ static void test_cross_enclave_calls()
     printf("=== test_cross_enclave_calls passed\n");
 }
 
+// n is a pointer so that it will lie in the current ecall_context's ocall
+// buffer. If all enclaves in the same thread use the same ocall buffer, then
+// the value of n will get overwritten. If the ocall-buffers are not shared
+// by the enclaves even thought they are being used in the same thread,
```
CodeMonkeyLeet: Typo: even _though_

radhikaj:
thanks @anakrish

host/sgx/enter.c
```diff
@@ -216,8 +226,16 @@ void oe_enter_sim(
     void* host_fs = oe_get_fs_register_base();
     sgx_tcs_t* sgx_tcs = (sgx_tcs_t*)tcs;
     oe_ecall_context_t ecall_context = {{0}};
-    ecall_context.ocall_buffer = _thread_ocall_buffer;
-    ecall_context.ocall_buffer_size = sizeof(_thread_ocall_buffer);
+    oe_thread_binding_t* binding = oe_get_thread_binding();
```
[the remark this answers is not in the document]
anakrish: Good idea. Done.

tests/ecall_ocall/host/host.cpp
anakrish: Fixed.

Issue #2624, opened 2020-03-03
failed - ctest - oesign-engine and samples - SGX1 mode - Ubuntu 16.04 


mofanv:
Hi,

I'm following [SGX1GettingStarted.md](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/SGX1GettingStarted.md) to set up Openenclave. After building successfully, I run `ctest` but there are three tests failed, which are `oesign-engine-works`, `oesign-engine-sign`, and `samples`.

For the oesign-engine-works,
```
test 9
Start 9: tests/oesign-engine-works

9: Test command: /home/vincent/Documents/openenclave/tests/tools/oesign-engine/test-enclave/oesign-test.py "--oesign-binary" "/home/vincent/Documents/openenclave/build/output/bin/oesign" "--enclave-binary" "/home/vincent/Documents/openenclave/build/tests/tools/oesign-engine/test-enclave/enclave/enclave" "--config-file" "/home/vincent/Documents/openenclave/tests/tools/oesign-engine/test-enclave/enclave/test-enclave.conf" "--engine-id" "'oesign-test-engine'" "--engine-load-path" "/home/vincent/Documents/openenclave/build/tests/tools/oesign-engine/test-enclave/../test-engine/liboesign-test-engine.so" "--key-id" "/home/vincent/Documents/openenclave/build/tests/tools/oesign-engine/test-enclave/private.pem"
9: Test timeout computed to be: 1500
9: /home/vincent/Documents/openenclave/build/output/bin/oesign ERROR: failed to open: /home/vincent/Documents/openenclave/build/tests/tools/oesign-engine/test-enclave/enclave/enclave.signed
9: /home/vincent/Documents/openenclave/build/output/bin/oesign ERROR: _update_and_write_signed_exe(): result=OE_FAILURE (1)
9: test_engine_init
9: test_engine_load_privkey engine=0xe23210, keyid = /home/vincent/Documents/openenclave/build/tests/tools/oesign-engine/test-enclave/private.pem

[...]

PR #2604, opened 2020-02-26
Added libsgx-launch package to ansible

yakman2020:
Fixes #2599 

Adds the libsgx-launch and libsgx-aesm-launch-plugin to the ansible setup in order to allow non-dcap enclave launch. This reflects a change in the intel driver/package arrangement.

PR #2680, opened 2020-03-16
Ensure that aesm and launch are installed in non dcap

yakman2020:
This is not yet covered in the document and is non-obvious.  We have an example of the problem in #2624.

### DIFF
--- a/docs/GettingStartedDocs/Contributors/SGX1GettingStarted.md
+++ b/docs/GettingStartedDocs/Contributors/SGX1GettingStarted.md
@@ -7,6 +7,13 @@
 - Red Hat Enterprise Linux 8 64-bits
 - SGX1 capable system. Most likely this will be an Intel SkyLake or Intel KabyLake system
 
+## Prepare the machine for sgx 
+
+Ensure that the aesm service and lanuch enclave are properly installed.
+```bash
+sudo apt install sgx-aesm-service libsgx-launch libsgx-ae-le
+```
+
 ## Clone Open Enclave SDK repo from GitHub
 
 Use the following command to download the source code (make sure `git` is installed before doing this):


jhand2:
Why can't this just be done in the ansible script? Our setup docs don't specifically call out any other packages that should be installed, they just say to use ansible.

docs/GettingStartedDocs/Contributors/SGX1GettingStarted.md
```diff
@@ -7,6 +7,13 @@
 - Red Hat Enterprise Linux 8 64-bits
 - SGX1 capable system. Most likely this will be an Intel SkyLake or Intel KabyLake system
 
+## Prepare the machine for sgx 
+
+Ensure that the aesm service and lanuch enclave are properly installed.
```
dthaler: typo: "lanuch"

docs/GettingStartedDocs/Contributors/SGX1GettingStarted.md
```diff
@@ -7,6 +7,13 @@
 - Red Hat Enterprise Linux 8 64-bits
 - SGX1 capable system. Most likely this will be an Intel SkyLake or Intel KabyLake system
 
+## Prepare the machine for sgx
```
dthaler: nit: capitalize SGX

BRMcLaren:
bors try

BRMcLaren:
I am wondering if we need to add a test case for this to the nightly builds. I agree it is of value, but I am conflicted about merging anything if we don't have a test for it.

Id we were to add the ansible scripts like @jhand2 suggested and we use the principles of having multilevel testing I think that would be really strong.

jhand2:
> I am wondering if we need to add a test case for this to the nightly builds. I agree it is of value, but I am conflicted about merging anything if we don't have a test for it.
> 
> If we were to add the ansible scripts like @jhand2 suggested and we use the principles of having multilevel testing I think that would be really strong.

Doesn't this PR just change documentation?

BRMcLaren:
@jhand2 Maybe I misunderstand. Is this documentation that reflects the manual set up of [this](https://github.com/openenclave/openenclave/pull/2604) pull request?

jhand2:
> @jhand2 Maybe I misunderstand. Is this documentation that reflects the manual set up of [this](https://github.com/openenclave/openenclave/pull/2604) pull request?

Yes I believe that is correct. My point is I don't think it's necessary and is potentially confusing since we don't document this for any other packages

kuanggang-git:

PR #2683, opened 2020-03-16
Refactor common TEE EDL into separate files based on functionality

jhand2:
EDL functions which are common accross all TEEs were in a single tee.edl file. Refactor these to be split into different EDL files based on functionality.

### DIFF
--- a/common/edl/tee.edl
+++ b/common/edl/tee.edl
@@ -4,7 +4,7 @@
 /*
 **==============================================================================
 **
-** internal.edl:
+** tee.edl:
 **
 **     This file declares internal ECALLs/OCALLs used by liboehost/liboecore
 **     for usage with any TEE technology.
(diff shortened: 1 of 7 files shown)


common/memory.edl
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+/*
+**==============================================================================
+**
+** memory.edl:
+**
+**     This file declares internal ECALLs/OCALLs used by liboehost/liboecore
+**     for manipulating memory allocations accross the enclave boundary.
```
CodeMonkeyLeet: Typo: across

common/tee.edl
```diff
@@ -14,14 +14,19 @@
...
+
     include "openenclave/bits/types.h"
 
     trusted
     {
-        public void oe_log_init_ecall(
-            [in, string] const char* enclave_path,
-            uint32_t log_level);
-
         public oe_result_t oe_verify_report_ecall(
```
CodeMonkeyLeet: This should probably be spun out into an attestation.edl or similar at some point ...

common/memory.edl
```diff
@@ -0,0 +1,23 @@
...
+**
+**==============================================================================
+*/
+
+enclave
+{
+    untrusted
+    {
+        void* oe_realloc_ocall(
```
CodeMonkeyLeet: Not needed in this PR, but we really need to think about renaming this to `oe_host_realloc_ocall()` for parity with its usage as with the rest of the host memory manipulation functions.

common/logging.edl
```diff
@@ -0,0 +1,38 @@
...
+
+    untrusted
+    {
+        void oe_log_ocall(
+            uint32_t log_level,
+            [in, string] const char* message);
+
+        // Write a string to the console. Write to STDOUT if device=0. Write
+        // to STDERR if device=1. Write strnlen(str, maxlen) bytes.
+        void oe_write_ocall(
```
CodeMonkeyLeet: Not for this PR, but we should figure out if the log APIs now have dependencies on syscall APIs and just use those directly perhaps.

common/tee.edl
jhand2: I almost did that but then I wasn't sure which ones were related to attestation and which weren't. Would you say oe_verify_report_ecall is attestation-related but the get_public_key functions are not?

I can go ahead and do that in this PR

common/memory.edl
jhand2: Agreed. I also want to see if host malloc and host free can be implemented in EDL and put them here. Now that Anand implemented a small static buffer  for parameters I think they can.

common/logging.edl
jhand2: I think the reason oe_write_ocall exists is so that we can do logging without depending on the syscall APIs.

common/tee.edl
dthaler: > Would you say oe_verify_report_ecall is attestation-related but the get_public_key functions are not?

Yes, exactly.

CodeMonkeyLeet: I had equated oeenclave functionality with attestation, but you're right  that we should treat key handling functions as not being related to attestation. The desire of the comment was to have tee.edl be a wrapper that can just go away when all the nested functions are done, so we could split the resulting set into attestation.edl (verify report, which I suspect should get subsumed into the attestation plugin model) and tee_keys.edl.

jhand2: I split into `attestation.edl` and keys.edl`

common/attestation.edl
```diff
@@ -0,0 +1,25 @@
...
+**
+**     This file declares internal ECALLs used by liboehost/liboecore for
+**     gathering/verifying attestion information from the enclave.
+**
+**==============================================================================
+*/
+
+enclave
+{
+    include "openenclave/bits/types.h"
```
mingweishih: I think this header is not needed, can you double check?

jhand2: oh ya you're right. I got a little copy/paste happy here

jhand2:
@mingweishih I'll hold this PR until you merge https://github.com/openenclave/openenclave/pull/2691

common/edl/tee.edl
```diff
@@ -14,108 +14,14 @@
 
 enclave
 {
-    /**
-     * This enumeration defines the type of a asymmetric key.
-     * This definition is shared by the enclave and the host.
+    /* The end goal is to publish all EDL as part of the SDK. At that time
```
dthaler: grammar: "all EDL" -> "all EDL files"

BRMcLaren:
bor r-

PR #2685, opened 2020-03-16
Refactor syscall ocalls into smaller groups in multiple edl files

jhand2:
Currently all syscalls must be included as a monolith and are kept in a single EDL file. Refactor these into smaller groupings so they can be more easily included to provide discrete functionality when important by an enclave application.

The ocalls are organized by the POSIX header which contains the function. For example, if nanosleep() is in time.h, then oe_syscall_nanosleep_ocall is in time.edl.


common/fcntl.edl
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+/*
+**==============================================================================
+**
+** epoll.edl:
```
mingweishih: nit: wrong file name.

common/unistd.edl
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+/*
+**==============================================================================
+**
+** syscall.edl:
```
mingweishih: nit: wrong file name.

dthaler:
Is syscall.edl still needed?  Or can it be removed and just use the more specific files directly?

jhand2:
> Is syscall.edl still needed? Or can it be removed and just use the more specific files directly?

The goal is to eventually publish all the edl files as part of the SDK where enclave developers will import them into their own edl files. syscall.edl is just left here as a convenience to aggregate all these functions until we reach the end goal.

common/signal.edl
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+/*
+**==============================================================================
+**
+** signal.edl:
+**
+**     This file declares OCALLs and ECALLs needed by the enclave to implement
+**     functions in POSIX signal.h
```
mingweishih: nit: missing period.

common/time.edl
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+/*
+**==============================================================================
+**
+** time.edl:
+**
+**     This file declares OCALLs and ECALLs needed by the enclave to implement
+**     functions in POSIX time.h
```
mingweishih: nit: missing period.

common/unistd.edl
```diff
@@ -0,0 +1,43 @@
...
+**
+**     This file declares OCALLs and ECALLs needed by the enclave to implement
+**     functions in POSIX unistd.h
```
mingweishih: nit: missing period.

common/utsname.edl
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+/*
+**==============================================================================
+**
+** utsname.edl:
+**
+**     This file declares OCALLs and ECALLs needed by the enclave to implement
+**     functions in POSIX utsname.h
```
mingweishih: nit: missing period.

common/poll.edl
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+/*
+**==============================================================================
+**
+** poll.edl:
+**
+**     This file declares OCALLs and ECALLs needed by the enclave to implement
+**     POSIX poll APIs.
```
CodeMonkeyLeet: Nit: implement _the_ POSIX poll API. (there's only one in poll.h anyway)

common/epoll.edl
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+/*
+**==============================================================================
+**
+** epoll.edl:
+**
+**     This file declares OCALLs and ECALLs needed by the enclave to implement
```
CodeMonkeyLeet: Note that none of the new EDL files actually define ECALLS, and the one ECALL defined in syscalls.edl was to work around the problem that edger8r required at least one ECALL in the EDL file.

This also suggests that despite the refactoring, the only way to use the syscall EDLs is through syscall.edl only. Is that correct, and if so, is that intended?

common/ioctl.edl
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+/*
+**==============================================================================
+**
+** ioctl.edl:
+**
+**     This file declares OCALLs and ECALLs needed by the enclave to implement
+**     functions in POSIX ioctl.h
```
CodeMonkeyLeet: Nit: Isn't ioctl.h a Linux header, not a POSIX one?

common/socket.edl
```diff
@@ -0,0 +1,218 @@
...
+
+/*
+**==============================================================================
+**
+** socket.edl:
+**
+**     This file declares OCALLs and ECALLs needed by the enclave to implement
+**     socket I/O.
+**
+**     Note that some of the syscalls in file.edl can also be passed a socket
```
CodeMonkeyLeet: Did you mean fcntl.edl rather than file.edl?

common/epoll.edl
jhand2: Ah, I didn't realize an ecall is required.

The intent is that these can be important into the EDL file for an enclave application, so they aren't really intended to stand alone. The idea is to make it easy to say "I want to use sockets, all I have to do is `from "socket.edl" import *`". So in that case I think it's fine to have no ecalls.

On that note though, this seems like a bug in edger8r that we should fix. I don't think there's a reason to require an ecall in an enclave application.

common/ioctl.edl
jhand2: ah you're right :smile:

common/signal.edl
achamayou: @mingweishih for trivial changes, it's worth using suggestion blocks: 
image

The PR submitter can just "approve" and the change goes in straight away, it's often just as quick if not quicker than writing the comment :)

common/epoll.edl
CodeMonkeyLeet: I agree that this should be fixed in edger8r (maybe it has? @anakrish?). You should just strike the mention of ECALLs in the EDL headers that don't contain them. 

I noticed it because of the presence of the ECALL stub in syscalls.edl and the explanation for it. There needs to be follow-up work to remove that at least, and it would be useful to test that a sample can build against sockets.edl by itself without syscalls.edl before we start promoting that mode of usage.

PR #2634, opened 2020-03-05
Adding Open enclave SDL Bug Bar

swamymsft:
This pull request is to add a Open Enclave SDL bug bar to this project.

PR #2691, opened 2020-03-17
Publish EDL headers

### LINKED ISSUE #2666: Use EDL to generate headers structures that are used in EDL
# Current Behavior

Currently many of the structures passed as arguments in EDL functions are defined in C header and included into EDL. For example `oe_asymmetric_key_params_t` is used in `common/tee.edl` by including the bits/types.h header:

https://github.com/openenclave/openenclave/blob/99410b10b4782d33e76819bc681e651e96cd9739/common/tee.edl#L17

These structures are known as "foreign structures" as they are not defined in EDL.

# The problem

The problem with using foreign structures is that they cannot benefit from [EDL deep copy](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/full_edl_serialization.md). On SGX, using foreign structures can pose security issues, since the enclave will be reading host memory. For trustzone, since the enclave and host to not share an address space, passing pointers that are not deep copied can cause crashes and other undefined behavior since the enclave cannot dereference host memory.

# Requirements
* Structures used as arguments in EDL functions can be annotated for EDL deep copy
* Structures passed as arguments in EDL can also be used by the enclave C code.

# Nice to have
* Enclave developers can somehow use C structures in their EDL
    * This could be defining a structure in EDL and then mapping it to a corresponding C structure with generated code to convert between the two.

mingweishih:
This PR drafts the design of using EDL to generate data structures used by ECALLs/OCALLs (and other places). The detail of the design is as follows (fixes #2666).
    
 - Modify the `oeedger8er` to generate the publish candidate (avoid publishing ecall/ocall specific data structures)
      - ecall/ocall related data structures are generated under *_u.h / *_t.h
      - only user includes and user defined data structures in EDL are generated under *_args.h, which servers as the publish candidate.
      - Emit the preprocessor directives that avoid the re-definition of data structures.
- Define the data structure in the EDL directly.
- The generated *_args.h is renamed and installed to `{OUTPUTDIR}|{INSTALLDIR}/include/openenclave/bits/`.
- Use `#include<openenclave/bits/...>` to include the header file.
    
Other changes:
 - Move all the edl files to `common/edl` and `common/sgx/edl`
 - Update the CHANGELOG.md


common/edl/asym_keys.edl
```diff
@@ -0,0 +1,63 @@
...
+    enum oe_asymmetric_key_type_t {
+        /**
+         * A secp256r1/NIST P-256 elliptic curve key.
+         */
+        OE_ASYMMETRIC_KEY_EC_SECP256P1 = 1,
+
+        /**
+         * Unused.
+         */
+        _OE_ASYMMETRIC_KEY_TYPE_MAX = 4294967295
```
anakrish: Is it possible to use OE_ENUM_MAX or does edger8r parser have a limitation?
If it is a limitation, it would be good to capture it in an issue and add it to backlog.

include/openenclave/host.h
```diff
@@ -14,6 +14,7 @@
 #error "enclave.h and host.h must not be included in the same compilation unit."
 #endif
 
+#include <openenclave/bits/edl/asym_keys.h>
```
anakrish: Does the `bits/edl` folder have to be installed too?

mingweishih: Yes, it's installed by https://github.com/openenclave/openenclave/blob/74c7a6f9f93bff670c2e6d5ec0548552da5012c8/common/edl/CMakeLists.txt#L38

common/edl/asym_keys.edl
mingweishih: Good catch. Thanks!

common/edl/asym_keys.edl
```diff
@@ -0,0 +1,63 @@
...
+
+        /**
+         * The exported format of the key.
+         */
+        oe_asymmetric_key_format_t format;
+
+        /**
+         * Optional user data to add to the key derivation.
+         */
+        [size=user_data_size] void* user_data;
```
anakrish: Given that `user_data` is marshaled by oeedger8r, I suspect there will be an accompanying change to the code that was consuming `oe_assymmetric_key_params_t`. It would no longer need to do boundary validation on `user_data`.

jhand2:
I'll take a deeper look in a little bit but at first glance this looks fantastic. Thanks so much for digging into this! Did you run into any additional problems or does it just work as expected?

mingweishih:
> I'll take a deeper look in a little bit but at first glance this looks fantastic. Thanks so much for digging into this! Did you run into any additional problems or does it just work as expected?

It works well with your suggestion of using preprocessor directives to avoid redefinition conflicts (e.g., https://github.com/openenclave/openenclave/blob/74c7a6f9f93bff670c2e6d5ec0548552da5012c8/host/sgx/report.c includes both `openenclave/host.h` and `tee_u.h`). Besides that, the PR mostly plays around with CMake to make sure the the include path is updated correctly.

common/edl/asym_keys.edl
mingweishih: The annotation here is specifically for deep copy.
For example, https://github.com/openenclave/openenclave/blob/74c7a6f9f93bff670c2e6d5ec0548552da5012c8/tests/oeedger8r/edl/deepcopy.edl#L10 will result in shallow copy, while https://github.com/openenclave/openenclave/blob/74c7a6f9f93bff670c2e6d5ec0548552da5012c8/tests/oeedger8r/edl/deepcopy.edl#L22 will trigger the deep copy based on the `count`.

common/edl/asym_keys.edl
mingweishih: Turns out using `OE_ENUM_MAX` did cause problem. Need to investigate it.

common/edl/asym_keys.edl
anakrish: What I meant was that previously, the `user_data` field which is a nested pointer was being shallow-copied (equivalent t o `user_check` annotation). With this PR, it is being deep-copied.
Since `oe_assymmetric_key_params_t` was being passed as input to ecalls, then
- Either we were doing a check `oe_is_outside_enclave(params->user_data, params->user_data_size)`. If so this check can be removed now since edger8r does a deep copy.
- Or we weren't doing the above check, then it means that there was a potential security vulnarability since the host could set `params->user_data` to point to a location within the enclave. In that case, this PR fixes that vulnerability and it would be good to call that out in the PR.

common/edl/asym_keys.edl
mingweishih: Updated: should work now.

common/edl/asym_keys.edl
jhand2: @anakrish This enclave-side code should have been checking that the memory is all in the host, but it wasn't. So no corresponding code change is needed here.

But this is all the more reason to not use foreign structures.

mingweishih: Interestingly, we didn't have those checks. I'll look more into that and decide whether to address in this PR or to open another issue for that.

anakrish: @jhand2 Thanks for confirming that this is fixing a security vulnerability. It is important to be call that out in the PR comments. Interesting that the edge-function dealing with cryptography was not fully secure.

Yes, this strongly underlines the need to use fully marshalled EDL. I think in some places we use explicit `user_check` annotation. Hopefully once the clean up is done, there will just be a handful of `user_check` and everything else is fully marshaled.

anakrish: @swamymsft This PR fixes a security vulnerability. As @jhand2 mentioned, EDL specifications ought not to use foreign structures. Where would the vulnerability being fixed land in your bug bar #2634?

include/openenclave/host.h
jhand2:
```
install(DIRECTORY ${OE_EDL_INCLUDES}
  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/openenclave/bits)
```

Isn't it getting installed into `openenclave/bits/`? Should the above instead be:

```
install(DIRECTORY ${OE_EDL_INCLUDES}
  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/openenclave/bits/edl)
```

Either that or `set(OE_EDL_INCLUDES ${OE_INCDIR}/openenclave/bits/edl)` should be `set(OE_EDL_INCLUDES ${OE_INCDIR}/openenclave/bits)`

jhand2: Blocking on this because I think it will break when building a sample or something outside the OESDK tests.

mingweishih: The `DIRECTORY` keyword works as copying the entire directory to the destination.
In this case, copy the `edl` directory to `${CMAKE_INSTALL_INCLUDEDIR}/openenclave/bits`. The result is creating `${CMAKE_INSTALL_INCLUDEDIR}/openenclave/bits/edl`.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [Unreleased][Unreleased_log]
 ------------
 
+### Changed
```
jhand2: Should minor security fixes like the one in this PR also go into the changelog?

include/openenclave/host.h
jhand2: Oh ok, thanks for the clarification :smile:

mingweishih: I found that because I was specifying exactly `${CMAKE_INSTALL_INCLUDEDIR}/openenclave/bits/edl' and samples test failed.

CHANGELOG.md
mingweishih: I think the fix is not included yet, mainly because we didn't turn on the deep copy. As a result, using edl to generate data structures does not bring any changes. I need to check and manually insert checks either in this PR or creating another one.

mingweishih: FYI: deepcopy is still an experimental feature per #1839 (requiring `--experimental` flag).

common/edl/asym_keys.edl
swamymsft: @anakrish Reading the discussion it looks like the arbitrary enclave memory can be copied to host by specifying any address and size, is that correct?

CHANGELOG.md
CodeMonkeyLeet: @jhand2 I think this is appropriately under `Changed` since the existing bits headers are public headers and all public changes should be documented in the CHANGELOG.md.

jhand2: Oh good point. I would prefer checks in a different PR or changing deep copy to no longer be an experimental feature

common/edl/asym_keys.edl
mingweishih: Please see  the analysis as follows. The conclusion is that this seems no security vulnerability. However, we can perform the check much earlier.
 
`oe_asymmetric_key_params_t` is used by two ecalls: `oe_get_public_key_ecall` and `oe_get_public_key_by_policy_ecall`.
The current behavior of argument passing is shallow copy (with deepcopy turned off), meaning that the host pointer will be passed into the enclave.

The `user_data` and `user_data_size` is used as the following trace.
`oe_get_public_key_ecall`
-> `oe_get_public_key`
-> `_load_asymmetric_key`
-> `_derive_asymmetric_key`
-> `_create_asymmetric_keypair`
-> `oe_kdf_derive_key` (`user_data` and `user_data_size` are passed as arguments)
-> `kdf_hmac_sha256_ctr`
-> `oe_hmac_sha256_update`
At this point, the size will be checked against `OE_INT_MAX`:
```
if (!context || !data || size > OE_INT_MAX)
```

include/openenclave/host.h
CodeMonkeyLeet: Side note: Is there a reason for partitioning this out into the edl subfolder when publishing to  bits? i.e. at the point of consumption in the SDK package, whether this header was edl generated or not is irrelevant

common/edl/asym_keys.edl
jhand2: @swamymsft It is write to an arbitrary location in the enclave, but not arbitrary data. The host can make the enclave write a public key to any enclave address, but the host doesn't have full control over the contents of that key (only partial control).

common/edl/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+add_library(oeedl_includes INTERFACE)
```
CodeMonkeyLeet: Setting up a separate edl folder under common for just these types seems like it adds another way that EDL is handled differently in the CMake project and I would really like to minimize that where possible. A couple of observations:

* The common folder has EDL files at the top level and also in the sgx folder, so having an `edl` folder that doesn't contain all the EDLs seems misleading. There's a reason for this folder that's not reflected in the name, nor explained elsewhere.

   * If you're going to keep this folder, it might be useful to at least call out in a README.md for the folder that these type definitions are special because they are included by the dev-authored, public-facing host.h & enclave.h headers. (Otherwise they could just be consumed like other regular args.h from the EDL build process).

* It looks like the only reason this is in a separate directory to to have a separate CMake target for this file ... why?

   * For example, an alternative might be that these data types are declared in tee.edl, which is already built, and then the extra `oeedl_includes` library target and custom command to copy and rename the resulting `tee_args.h` to includes could be done in either the host or enclave builds.

mingweishih: The reason to have a separated edl with only data structures is trying make the published header as clean as possible. For example, the `tee_args.h` will contain several ECALL/OCALL marshalling structs, which I tried to avoid in the published header.

The end goal is to publishing all the data structures used by ECALLs/OCALLs in a similar way, which will expands the list. We can definitely move all the EDL files under `common` into the `common/edl` folder given that the list of EDL is increasing because of recent refactoring work.

include/openenclave/host.h
mingweishih: Agree. Will directly publish headers to bits instead.

common/edl/asym_keys.edl
swamymsft: thanks @jhand2 & @mingweishih, this isn't as worse as I initially thought. But I agree with @jhand2 that it can be used to infer the contents of the enclave memory depending on the key derivation.

If the key derivation deterministically yield to same value for a particular value in the memory, then an attacker can use it to actually know the arbitrary enclave memory value by repeatedly feeding different values through a host memory pointer that he control. In this case the bug can be converted to arbitrary enclave memory read. The severity depends on the key derivation function, any insights into this function?

common/edl/CMakeLists.txt
jhand2: I'm in favor of moving all the EDL files under this folder.

mingweishih: Related question to this decision.
If we consider allowing a developer to import any of internal edl files, then we need to expose those edl files as part of the installation package.

Take an example from this PR. If a developer wants to use `oe_asymmetric_key_params_t` as part of his ECALLs/OCALLs, he then needs to find the edl file somewhere and import.

Something like:

`from "{installdir}/include/openenclave/edl/asym_keys.edl" import *`

or

`from "{installdir}/edl/asym_keys.edl" import *`

or 

`from "{installdir}/common/edl/asym_keys.edl" import *`


@jhand2 What's the plan to opt-in OCALLs?

CodeMonkeyLeet: @jhand2 
> @CodeMonkeyLeet was there a desire to treat EDL the same as C files and not split it out in the directory structure? If so I would also be fine with keeping the directory structure as-is.

I think EDL files are already being treated in a special way regardless of where they are located based on the existing host/CMakeLists.txt and enclave/CMakeLists.txt, my concern is that we're introducing yet another special way to build EDL files and consume them in this special subdirectory. 

To be clear, I don't feel strongly that we need to resolve it in this PR, but I think it points to additional work we need to more broadly with regards to how OE consumes headers generated by EDL to enable deep-copy scenarios. Here's a couple of other pivots:

* I had a suggestion that these data structures should just be a part of tee.edl and then tee_args.h published directly to the bits folder. There could be concern that not all of the args in tee_args.h need to be published (e.g. `oe_log_ocall_args_t`; note that function headers are not published in *_args.h only the argument types). I think the way to address that should actually be to split the functions and its data types out into their own EDL, because it's really the functions themselves that create this dependency. 

* At a higher level, some of these methods are largely exposed through EDL just to provide convenience methods on the host side to invoke enclave functions, and it's becoming more clear that maybe that just complicates the API surface if the use cases are not actually in use (e.g. `oe_get_report` from the host side is such a candidate, given its different semantics from the enclave variant)

mingweishih: @jhand2 Do you expect the list of such data structure to grow (i.e., the one used by internal ECALLs needs to be published)? If this is a special case, then I agree not new creating this new cmake target for that.

@CodeMonkeyLeet Are you fine with the published headers containing ECALL/OCALL marshalling structs?

mingweishih: > @jhand2
> 
> > @CodeMonkeyLeet was there a desire to treat EDL the same as C files and not split it out in the directory structure? If so I would also be fine with keeping the directory structure as-is.
> 
> I think EDL files are already being treated in a special way regardless of where they are located based on the existing host/CMakeLists.txt and enclave/CMakeLists.txt, my concern is that we're introducing yet another special way to build EDL files and consume them in this special subdirectory.
> 
> To be clear, I don't feel strongly that we need to resolve it in this PR, but I think it points to additional work we need to more broadly with regards to how OE consumes headers generated by EDL to enable deep-copy scenarios. Here's a couple of other pivots:
> 
> * I had a suggestion that these data structures should just be a part of tee.edl and then tee_args.h published directly to the bits folder. There could be concern that not all of the args in tee_args.h need to be published (e.g. `oe_log_ocall_args_t`; note that function headers are not published in *_args.h only the argument types). I think the way to address that should actually be to split the functions and its data types out into their own EDL, because it's really the functions themselves that create this dependency.

This is the direction I thought initially, which cleanly addresses my concern. I'd like to go for it.

> * At a higher level, some of these methods are largely exposed through EDL just to provide convenience methods on the host side to invoke enclave functions, and it's becoming more clear that maybe that just complicates the API surface if the use cases are not actually in use (e.g. `oe_get_report` from the host side is such a candidate, given its different semantics from the enclave variant)

CodeMonkeyLeet: @mingweishih I'm okay with it as a tradeoff for a more consistent build flow. I agree with @jhand2 that having the data structures declared separately is cleaner in terms of public includes, and I will defer to him if he feels strongly about it. 

Regarding the list of data structures, I suspect it will continue to grow for any function that needs to marshal an OE data structure across the boundary, which may not be unique to this function once we start pushing oecore functions out of manual handling and into EDL.

jhand2: @mingweishih I would prefer that structs which will be built into headers and published as part of the public API surface be kept in separate EDL files. For internal structs I think it's fine to keep them in the same EDL file as the functions that use them.

common/edl/tee.edl
```diff
@@ -14,7 +14,63 @@
 
 enclave
 {
-    include "openenclave/bits/types.h"
+    /**
```
jhand2: Can we move these back to a separate EDL file and just be imported? tee.edl will be going away once all the edl files are published.

Specifically asking this because the header with these structures will be published. The issue is that the _args.h file may end up containing structures which don't need to be included by enclave.h/host.h if it is generated from a file with other structures as arguments. Plus if we want to also put internal structures here, I was the public ones to be separate.

jhand2:
Just a todo before this is merged to get rid of the merge commit :smile:

common/edl/tee.edl
mingweishih: I was thinking leaving this part to your PR  :smile:
By the way, the updated PR modifies the oeedge8er such that only user-included (C header) and user-defined structs will be included in _args.h.

jhand2: ah ok, thats fine then :smile:

jhand2:
Thanks @mingweishih!

bors r+

jhand2:
@mingweishih could you take a look at this https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/Bors/detail/staging/140/pipeline/1909

If looks like there is some issue on Windows with the header getting published

mingweishih:
> @mingweishih could you take a look at this https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/Bors/detail/staging/140/pipeline/1909
> 
> If looks like there is some issue on Windows with the header getting published

I'm looking into it. Thanks.

mingweishih:
> > @mingweishih could you take a look at this https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/Bors/detail/staging/140/pipeline/1909
> > If looks like there is some issue on Windows with the header getting published
> 
> I'm looking into it. Thanks.

Fixed. Try again?

mingweishih:
> > @mingweishih could you take a look at this https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/Bors/detail/staging/140/pipeline/1909
> > If looks like there is some issue on Windows with the header getting published
> 
> I'm looking into it. Thanks.

Should be good to go now.

jhand2:
thanks @mingweishih!

bors r+

mingweishih:
> thanks @mingweishih!
> 
> bors r+

Unfortunately, it seems I hit something else. Need to take another look.

mingweishih:
bors try

mingweishih:
bors try

PR #2702, opened 2020-03-19
Log TCB level variables for OE_TCB_LEVEL_INVALID

yentsanglee:
Log all SGX TCB level variables when the quote verification fails due to invalid TCB level status. Variables include:

- TCB Info
- FMSPC
- QE_ID
- CPU_SVN
- PCE_SVN
- PCE_ID

These variables help identify why quote verification fails.


yentsanglee:
Reverted the change. Will do this from test tool instead of production code.

PR #2705, opened 2020-03-19
Give warning when building with 32-bit targets

### LINKED ISSUE #2492: x86 compilation fails with cryptic errors
Here’s the errors:
1>c:\users\dthaler\source\repos\myenclave7\packages\open-enclave.0.7.0\include\openenclave\bits\types.h(39): error C2371: 'size_t': redefinition; different basic types
1>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.16.27023\include\vcruntime.h(184): note: see declaration of 'size_t'
1>c:\users\dthaler\source\repos\myenclave7\packages\open-enclave.0.7.0\include\openenclave\bits\types.h(48): error C2371: 'uintptr_t': redefinition; different basic types
1>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.16.27023\include\vadefs.h(30): note: see declaration of 'uintptr_t'
1>c:\users\dthaler\source\repos\myenclave7\packages\open-enclave.0.7.0\include\openenclave\bits\types.h(49): error C2371: 'ptrdiff_t': redefinition; different basic types
1>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.16.27023\include\vcruntime.h(185): note: see declaration of 'ptrdiff_t'
1>c:\users\dthaler\source\repos\myenclave7\packages\open-enclave.0.7.0\include\openenclave\bits\types.h(50): error C2371: 'intptr_t': redefinition; different basic types
1>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.16.27023\include\vcruntime.h(186): note: see declaration of 'intptr_t'

bits\types.h has a section specifically for MSVC host apps, which includes (among other definitions):

#elif defined(_MSC_VER)                                                                                        
typedef unsigned long long size_t;
typedef unsigned long long uintptr_t;                                                                                   
typedef long long ptrdiff_t;                                                                                            
typedef long long intptr_t;                                                                                            

However, Windows headers have things like this which define the types differently for 32-bit vs 64-bit:

#ifdef _WIN64
    typedef unsigned __int64 size_t;
    typedef __int64          ptrdiff_t;
    typedef __int64          intptr_t;
#else
    typedef unsigned int     size_t;
    typedef int              ptrdiff_t;
    typedef int              intptr_t;
#endif

So the definition matches for _WIN64 but not for 32-bit apps, which results in the errors quoted.

I searched through all the “Getting Started” docs in the github repo, and it contains no statement about 32-bit or 64-bit.  That is, there’s nothing saying 32-bit apps are NOT supported, nor is there any statement saying that 64-bit apps ARE supported.  So currently by lack of such a statement, OE SDK is assumed by the reader to have equal support.

Secondly, if you look at the new_platforms docs, which is what the IoT blog points to, and the VS extension used to use (before my changes),
https://github.com/openenclave/openenclave/blob/feature.new_platforms/new_platforms/docs/win_sgx_dev.md clearly says that x86 apps are supported by the extension, and so this is a regression for anyone already using the extension.

Now I’m ok with having a breaking change in the short term in the interest of using the master branch, but lack of 32-bit support needs to be clearly called out in the OE SDK docs and an intelligent error message given on compilation so the dev doesn’t spend hours figuring out what went wrong.   This also causes a problem for any apps that
also need to use (outside of an enclave) a third-party library that is 32-bit only, since they cannot use the OE SDK even if they are running on SGX hardware, and so that should also be called out as a limitation since many libraries only provide 32-bit binaries (some provide both 32 and 64-bit for performance, but many just provide 32-bit only since they run on both platforms).

mingweishih:
This PR adds compile-time warning when building 32-bit targets.
Also, the PR updates document that explicitly states that OE currently does not support 32-bit targets.

Fixes #2492

### DIFF
--- a/docs/GettingStartedDocs/Linux_using_oe_sdk.md
+++ b/docs/GettingStartedDocs/Linux_using_oe_sdk.md
@@ -3,6 +3,8 @@
 This document provides a brief overview of how to start exploring the Open Enclave SDK
 once you have it installed.
 
+**Note**: The SDK currently does not support 32-bit applications.
+
 ## Open Enclave SDK Layout
 
 On Linux, by default, the Open Enclave SDK is installed to `/opt/openenclave`.
--- a/docs/GettingStartedDocs/OP-TEE/Introduction.md
+++ b/docs/GettingStartedDocs/OP-TEE/Introduction.md
@@ -78,6 +78,10 @@ One of OP-TEE's test cases is known to fail on the TrustBox.
 
 [Tracking issue](https://github.com/openenclave/openenclave/issues/2275).
 
+### 32-bit targets are not supported
+
+[Tracking issue](https://github.com/openenclave/openenclave/issues/2493).
+
 ## Note on Forks
 
 While the SDK's support for Intel SGX is fairly self-contained, support for
--- a/docs/GettingStartedDocs/Windows_using_oe_sdk.md
+++ b/docs/GettingStartedDocs/Windows_using_oe_sdk.md
@@ -3,6 +3,8 @@
 This document provides a brief overview of how to start exploring the Open Enclave SDK
 once you have it installed.
 
+**Note**: The SDK currently does not support 32-bit applications.
+
 ## Open Enclave SDK Layout
 
 On Windows, if you installed the SDK using the NuGet package, it is by default installed to `%userprofile%\.nuget\packages`.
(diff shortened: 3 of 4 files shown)


radhikaj:
Thanks @mingweishih

PR #2650, opened 2020-03-10
Add changelog and change VERSION for 0.8.2

radhikaj:
Signed-off-by: Radhika Jandhyala <radhikaj@microsoft.com>

Will keep changelog updated as more PRs related to 0.8.2 are merged.

PR #2706, opened 2020-03-19
Add _verify_claims_hash before extract the claims

### LINKED ISSUE #2658: SGX verifier plugin shall verify the integrity of the serialized custom claims before extraction
In SGX verifier plugin code file common/sgx/verifier.c :

 Upon successful verification of the SGX report by oe_verify_quote_with_sgx_endorsements(), the _verify_evidence() function shall retrieve the hash value held in the report_data field (as defined in the oe_report_t structure), and use it to verify the integrity of the serialized claims. This verification shall be done before deserializing and extracting the claims.

qiucwang:
Fixes #2658
This pull request is to fix #2658.
One step missing from evidence verification is that the hash of the serialized claims should be verified against the report data, which is the hash of original claims, to guarantee the integrity of the claims, before deserializing the claims.
This pull request added a function _verify_claims_hash which performs such verification. The function is called in the _extract_claims so that claims will be deserialized only after it pass the claims hash verification.

### DIFF
--- a/common/sgx/verifier.c
+++ b/common/sgx/verifier.c
@@ -102,11 +102,13 @@ static oe_result_t _verify_local_report(
 }
 
 static oe_result_t _verify_claims_hash(
-    sgx_quote_t* quote,
+    const uint8_t* report,
+    oe_report_type_t report_type,
     const uint8_t* claims,
     size_t claims_size)
 {
     oe_result_t result = OE_UNEXPECTED;
+    int hash_cmp = 0;
     OE_SHA256 hash;
     oe_sha256_context_t hash_ctx = {0};
 
@@ -114,10 +116,22 @@ static oe_result_t _verify_claims_hash(
     OE_CHECK(oe_sha256_update(&hash_ctx, claims, claims_size));
     OE_CHECK(oe_sha256_final(&hash_ctx, &hash));
 
-    if (memcmp(&hash, &quote->report_body.report_data, OE_SHA256_SIZE) == 0)
-        result = OE_OK;
+    if (report_type == OE_REPORT_TYPE_SGX_REMOTE)
+    {
+        hash_cmp = memcmp(
+            &((sgx_quote_t*)report)->report_body.report_data,
+            &hash,
+            OE_SHA256_SIZE);
+    }
+    else if (report_type == OE_REPORT_TYPE_SGX_LOCAL)
+    {
+        hash_cmp = memcmp(
+            &((sgx_report_t*)report)->body.report_data, &hash, OE_SHA256_SIZE);
+    }
     else
-        result = QE_QUOTE_HASH_MISMATCH;
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+    result = (hash_cmp == 0) ? OE_OK : OE_QUOTE_HASH_MISMATCH;
 done:
     return result;
 }
@@ -349,13 +363,11 @@ static oe_result_t _extract_claims(
 
     // verify the integrity of the serialized claims with hash stored in
     // report_data.
-    if (header->report_type == OE_REPORT_TYPE_SGX_REMOTE)
-    {
-        OE_CHECK(_verify_claims_hash(
-            (sgx_quote_t*)header->report,
-            evidence + report_size,
-            evidence_size - report_size));
-    }
+    OE_CHECK(_verify_claims_hash(
+        header->report,
+        header->report_type,
+        evidence + report_size,
+        evidence_size - report_size));
 
     claims_header = (oe_sgx_plugin_claims_header_t*)(evidence + report_size);
 
--- a/common/sgx/verifier.c
+++ b/common/sgx/verifier.c
@@ -101,6 +101,25 @@ static oe_result_t _verify_local_report(
 #endif
 }
 
+static oe_result_t _verify_claims_hash(
+    sgx_quote_t* quote,
+    const uint8_t* claims,
+    size_t claims_size)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    OE_SHA256* hash = (OE_SHA256*)quote->report_body.report_data.field;
+    OE_SHA256 claims_hash;
+    oe_sha256_context_t hash_ctx = {0};
+
+    OE_CHECK(oe_sha256_init(&hash_ctx));
+    OE_CHECK(oe_sha256_update(&hash_ctx, claims, claims_size));
+    OE_CHECK(oe_sha256_final(&hash_ctx, &claims_hash));
+
+    result = memcmp(hash, &claims_hash, OE_SHA256_SIZE) ? OE_FAILURE : OE_OK;
+done:
+    return result;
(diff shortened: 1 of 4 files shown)


jhand2:
Can you update the PR description to detail what this change does and why it is needed?

Thanks!

CodeMonkeyLeet:
A couple of notes: 

* Please follow the [contribution guidelines](https://github.com/openenclave/openenclave/blob/master/docs/Contributing.md#commit-messages) for writing your commit messages.
    * If there's a single commit associated with the PR, it also becomes the PR description.
* If you want to reference an issue, use `#2658`, GitHub will automatically hotlink the issue like this: #2650 
   * If your PR fixes the issue, include the line `Fixes #2658` in your commit or PR description and GitHub will also automatically resolve the issue when the PR is merged.

qiucwang:
> A couple of notes:
> 
> * Please follow the [contribution guidelines](https://github.com/openenclave/openenclave/blob/master/docs/Contributing.md#commit-messages) for writing your commit messages.
>   
>   * If there's a single commit associated with the PR, it also becomes the PR description.
> * If you want to reference an issue, use `#2658`, GitHub will automatically hotlink the issue like this: #2650
>   
>   * If your PR fixes the issue, include the line `Fixes #2658` in your commit or PR description and GitHub will also automatically resolve the issue when the PR is merged.

Thanks!

qiucwang:
> Thanks for fixing this! Can you add a test for this case as well? I'm surprised this wasn't caught earlier.

The update passed current test case: ctest -R attestation_plugin
And sure, I'll add a new test which modify the claims after an evidence was generated and verify should failed. Sounds good?

CodeMonkeyLeet:
@qiucwang 
> The update passed current test case: ctest -R attestation_plugin
> And sure, I'll add a new test which will modify the claims after an evidence was generated and the verify should fail. Sounds good?

Yes, my concern is that the existing test did not catch the lack of the claims validation; adding the negative test case on tampered claims would help catch this in the future, thank you!

common/sgx/verifier.c
```diff
@@ -101,6 +101,25 @@ static oe_result_t _verify_local_report(
...
+    oe_result_t result = OE_UNEXPECTED;
+    OE_SHA256* hash = (OE_SHA256*)quote->report_body.report_data.field;
+    OE_SHA256 claims_hash;
+    oe_sha256_context_t hash_ctx = {0};
+
+    OE_CHECK(oe_sha256_init(&hash_ctx));
+    OE_CHECK(oe_sha256_update(&hash_ctx, claims, claims_size));
+    OE_CHECK(oe_sha256_final(&hash_ctx, &claims_hash));
+
+    result = memcmp(hash, &claims_hash, OE_SHA256_SIZE) ? OE_FAILURE : OE_OK;
```
yentsanglee: >memcmp(hash, &claims_hash, OE_SHA256_SIZE) [](start = 13, length = 42)

Make this a boolean. i.e. 
    (memcmp() == 0) ? good : bad;

yentsanglee: I'd suggest adding a new error code in oe_result_t to indicate the hash value mismatch. The failure most likely occurs when the claims attached to the evidence is modified.

qiucwang:
Add a new test case here to check tampered claims in tests/attestation_plugin/plugin/test.c.
At the end of `void verify_sgx_evidence()`, I manually tampered the last byte in header->data, where is the claims section. 
With unfixed verifier, the sgx_remote test 1 (just evidence) and 2 (with endorsement) will return OE_OUT_OF_MEMORY error because these two cases filled evidence with a NULL claim. Thus, tampering the last byte in claims section will actually tamper the claim size, which return a enormous huge number causing OUT_OF_MEMORY. The test 3 (with claims) will incorrectly return OE_OK, indicating current verifier misses the tampered claims risk.
With fixed verifier, the test cases will correctly return OE_FAILURE since claims section has been tampered.

common/sgx/verifier.c
qiucwang: Yes, I agree.
Any one else suggest doing so? And what kind of name we should use?

qiucwang:
> @qiucwang
> 
> > The update passed current test case: ctest -R attestation_plugin
> > And sure, I'll add a new test which will modify the claims after an evidence was generated and the verify should fail. Sounds good?
> 
> Yes, my concern is that the existing test did not catch the lack of the claims validation; adding the negative test case on tampered claims would help catch this in the future, thank you!

I've added on in the latest commit and make the comment above.

qiucwang:
> Thanks for fixing this! Can you add a test for this case as well? I'm surprised this wasn't caught earlier.

I've added on in the latest commit and make the comment above.

qiucwang:
>
> > The update passed current test case: ctest -R attestation_plugin
> > And sure, I'll add a new test which will modify the claims after an evidence was generated and the verify should fail. Sounds good?
> 
> Yes, my concern is that the existing test did not catch the lack of the claims validation; adding the negative test case on tampered claims would help catch this in the future, thank you!
Hi @CodeMonkeyLeet, I've added a new test case to catch that bug. And comment above.

common/sgx/verifier.c
qiucwang: How about OE_VERIFY_CLAIMS_TAMPERED? I saw several error codes in forms of OE_VERIFY_XXX

include/openenclave/bits/result.h
```diff
@@ -344,6 +344,12 @@ typedef enum _oe_result
      */
     OE_ALREADY_INITIALIZED,
 
+    /**
+     * The hash of claims section in the quote does not match the hash value
+     * held in the report_data field. This occurs when claims are tampered.
+     */
+    QE_QUOTE_HASH_MISMATCH,
+
```
soccerGB: I am curious why it's QE_QUOTE_HASH_MISMATCH, not OE_QUOTE_HASH_MISMATCH.
Any special reason?

yentsanglee: Right, it should be OE_

---
In reply to: [397695413](https://github.com/openenclave/openenclave/pull/2706#discussion_r397695413) [](ancestors = 397695413)

common/sgx/verifier.c
```diff
@@ -326,6 +347,16 @@ static oe_result_t _extract_claims(
     if (evidence_size - report_size < sizeof(*claims_header))
         OE_RAISE(OE_INVALID_PARAMETER);
 
+    // verify the integrity of the serialized claims with hash stored in
+    // report_data.
+    if (header->report_type == OE_REPORT_TYPE_SGX_REMOTE)
```
yentsanglee: >OE_REPORT_TYPE_SGX_REMOTE [](start = 31, length = 25)

Why is this remote only?

include/openenclave/bits/result.h
qiucwang: Yes...It should be OE. Sorry for the typo.

soccerGB:
Bors try

Issue #2714, opened 2020-03-20
Outdated buildandsign doc


dthaler:
GettingStartedDocs\buildandsign.md looks to be outdated.
* There is no mention of OE_SET_ENCLAVE_OPTEE, only SGX.
* The only discussion of .conf parameters is SGX-specific, without explaining why.

CodeMonkeyLeet:
Removed from SIG-Arch, will track the arch issue discussions in the future with #2844, this issue will track fixing up the documentation.

PR #2722, opened 2020-03-21
Design document for Pluggable Allocators

anakrish:
This pull request proposes the design for pluggable allocators.
Users will be able to provide replacement functions and plug-in their own high-performant
allocators into OE SDK.

### DIFF
--- a/docs/DesignDocs/PluggableAllocators.md
+++ b/docs/DesignDocs/PluggableAllocators.md
@@ -1,0 +1,149 @@
+Pluggable Allocators
+=====
+
+Open Enclave provides a default memory allocator for enclaves.
+This default allocator, [dlmalloc (Doug Lea Malloc)](http://gee.cs.oswego.edu/dl/html/malloc.html), is public domain, popular, and also serves as the basis for glibc's allocator. It is also easily modified to work within enclaves. However, it not performant in multi-threaded applications.
+
+There are various high-performance memory allocators that try to optimize, among other things, multi-threaded scenarios:
+- [The GNU Allocator](https://www.gnu.org/software/libc/manual/html_node/The-GNU-Allocator.html) which is based on `pthreads malloc`, which is in turn based on `dlmalloc`.
+- [Thread-Caching Malloc (tcmalloc)](https://gperftools.github.io/gperftools/tcmalloc.html).
+- [snmalloc](https://github.com/Microsoft/snmalloc).
+- Many others, such as application-specific allocators that often outperform general-purpose allocators.
+
+
+This document describes the design of a scheme to allow custom allocators to be used in enclaves.
+
+
+Precedents
+------
+
+- The GNU C Library provides a way to replace its malloc implementation, documented in [Replacing Malloc](https://www.gnu.org/software/libc/manual/html_node/Replacing-malloc.html#Replacing-malloc). </br>
+  To replace the allocator, the following replacement functions must be provided `malloc`, `free`, `calloc`, `realloc`, `aligned_alloc`, `malloc_usable_size`, `memalign`, `posix_memalign`, `palloc`, and `valloc`. </br>
+  Replacement happens by statically linking the allocator library before the GNU C Library.
+
+- This document covers compiling the Open Enclave SDK to use `snmalloc` (Pull Request #2649). Currently, users must recompile OE SDK with `USE_SNMALLOC` to leverage this.
+
+- Rust Language is [considering](https://doc.rust-lang.org/1.15.1/book/custom-allocators.html) capability to switch out the default allocators (`alloc_system` and `alloc_jemalloc`).
+
+
+Non-Requirements
+-----
+
+- This design does not aim to address what the ideal default allocator for Open Enclave should be.
+- Though the design could be extended to support multiple pluggable allocators, each managing a portion of the heap, it is currently a non-requirement.
+
+Requirements
+----
+
+The design for `Pluggable Allocators` additionally has the following requirements,
+
+- Allocators must be pluggable by an SDK user without having to recompile the OE SDK </br>
+  This is similar to the GNU C library in that one need not rebuild glibc to plug in a malloc replacement.
+
+- Must work on various TEE environments that the Open Enclave SDK will support. </br>
+  Note, it does not mean that a given allocator must work on all platforms; rather Open Enclave SDK must support plugging allocators on all platforms. </br>
+  From a design perspective, this precludes the use of architecture specific data structures in the plugin interface.
+
+- The design must support building existing functionality like `failure callbacks` and `debug allocators` atop custom allocators.
+
+- The design must allow allocators to perform thread-specific optimizations since multi-threaded applications benefit a lot from fine-tuned allocators.
+
+- Allocators must rely only on published interfaces and datastructures.
+
+- The design must allow for Open Enclave core (`liboecore`) to own a portion of the heap and have the allocator manage rest of the heap. </br>
+  This allows the SDK to manage a bit of memory for efficiently dispatching ecalls/ocalls, logging, and other core features, better than what a general-purpose
+  allocator would be capable of. Additionally, it also allows a clear memory demarcation between `core memory` and `application memory`.
+
+- Performance and safety. Plugin approaches often involve registering `hook` functions that will be called by the SDK via indirect calls.
+  It is desirable to avoid indirect calls since they are not performant. Additionally, indirect calls are subject to speculative execution side-channels.
+
+- Backwards-compatibility. No existing clients must require any change to get the default allocator.
+
+
+Design
+---
+
+A custom allocator must provide a set of replacement functions. They are declared in the published header `openenclave/advanced/allocator.h`. The set of replacement functions are mostly equivalents of the functions required by the GNU C Library for plugging-in custom allocators. Additionally, there are replacement functions that allow the allocator to work well with enclave threads.
+The SDK will provide default implementations of these functions that will result in the use of the default allocator.
+
+If no replacement functions are provided, then the default allocator is used. This retains the current behavior in which all enclaves get the default allocator without any additional linker options. Providing a partial set of replacement functions would result in linker error.
+
+The replacement functions can be provided
+- via object files. This happens when the allocator is not a separate static library.
+- via a separate allocator static library. In this case, the allocator library must be specified before `oelibc` in the linker command line. </br>
+
+
+Note,  ELF  `weak symbols` (see [Function Attributes](https://gcc.gnu.org/onlinedocs/gcc-3.2/gcc/Function-Attributes.html)) are designed to allow user code to override functions defined in libraries. Using `weak symbols` for default implementation of these function would cause the linker to look for custom replacement functions irrespective of the order. However, if only a partial set of replacement functions are provided, then weak symbols would result in the linker picking the default implementations for the rest of the functions. To avoid a mix of API from different allocators, the default implementations shall not be weak symbols.
+
+
+Replacement functions are not required to be reentrant.
+
+Undefined Behavior
+---
+
+- The replacement functions should not do much other than managing memory. Making ocalls, logging, printf etc may use dynamic memory and hence result in reentranct calls to the allocator.
+- All the replacement functions must be provided, or none. A partial set will result in both the default allocator and the custom allocator managing the heap.
+
+
+Replacement Functions
+---
+
+Published in `openenclave/advanced/allocator.h`
+
+- `void oe_allocator_init(void* heap_start_address, void* heap_end_address)` </br>
+  This function will be called by `oecore` at an appropriate time during enclave initialization. The total heap memory available to the enclave is specified in its configuration file or in a macro such as OE_SET_ENCLAVE_SGX or OE_SET_ENCLAVE_OPTEE
+  (e.g [hello_world.conf](https://github.com/openenclave/openenclave/blob/e2e440ec134d37e107b5a296fdc508a3a643598e/samples/helloworld/enclave/helloworld.conf#L6)) </br>
+  `oecore` may reserve a small portion for its own use and pass the rest to this replacement function. The allocator is expected to perform any global initialization in this function. </br>
+  This function is called prior to any global initialization.
+
+
+- `void oe_allocator_cleanup(void)` </br>
+  This function will be called by `oecore` at an appropriate time during enclave termination. `atexit` functions will be executed prior to `oe_allocator_cleanup`. </br>
+  `oe_allocator_cleanup` may not register functions with `atexit`.
+
+
+- `void oe_allocator_thread_init(void)` </br>
+  This function will be called by `oecore` during initialization of an enclave thread, after the thread-local variables for the thread have been initialized. </br>
+  In the future, if `oecore` needs to hold onto any thread-specific data on behalf of the allocator, it can be returned in this function. Currently the recommendation is to use `__thread` variables to store thread-specific allocator data.
+
+
+- `void oe_allocator_thread_cleanup(void)` </br>
+  This function will be called by `oecore` just prior to the termination of enclave thread.
+  Thread-specific exit functions (`__cxa_thread_atexit` functions) will be executed prior to `oe_allocator_thread_cleanup`. </br>
+  This allows the thread-specific exit functions to use memory management functions (e.g to free resources) before the allocator itself is cleaned up. </br>
+  It is recommended that allocators be judicious about using C++11 thread-specific objects in `oe_allocator_thread_cleanup` since
+  use of C++11 thread-local objects may trigger a call to (`__cxa_thread_atexit`) after all the thread-specific exit functions have executed.
+
+
+  It is important that the allocators be aware of the nature of enclave threads, and be designed accordingly. Enclave threads can be short-lived. For OE SGX, enclave threads last only for the duration of an ecall. </br>
+  Additionally, the same thread-control structure (e.g., `sgx_tcs_t` for SGX) can be bound to different host threads at different times during the lifespan of an enclave.
+
+
+- `void* oe_allocator_malloc(size_t size)` </br>
+  This function will be called by `oecore` to implement `malloc`. Depending upon the build settings, the allocated memory might be tracked by the `debug-malloc` feature. `oe_allocator_malloc` must provide the same semantics as [malloc](https://en.cppreference.com/w/c/memory/malloc).
+
+
+- `void oe_allocator_free(void* ptr)` </br>
+  This function will be called by `oecore` to implement `free`. `oe_allocator_free` must provide the same semantics as [free](https://en.cppreference.com/w/c/memory/free).
+
+
+- `void* oe_allocator_calloc(size_t nmemb, size_t size)` </br>
+  This function will be called by `oecore` to implement `calloc`. `oe_allocator_calloc` must provide the same semantics as [calloc](https://en.cppreference.com/w/c/memory/calloc).
+
+
+- `void* oe_allocator_realloc(void* ptr, size_t size)` </br>
+  This function will be called by `oecore` to implement `realloc`. `oe_allocator_realloc` must provide the same semantics as [realloc](https://en.cppreference.com/w/c/memory/realloc).
+
+- `void* oe_allocator_aligned_alloc(size_t alignment, size_t size)` </br>
+  This function will be called by `oecore` to implement [aligned_alloc](https://linux.die.net/man/3/aligned_alloc/man/3/posix_memalign). Memory allocated by this function can be freed via `free`.
+
+- `void oe_allocator_posix_memalign(void** memptr, size_t alignment, size_t size)` </br>
+  This function will be called by `oecore` to implement [posix_memalign](https://linux.die.net/man/3/posix_memalign). Memory allocated by this function can be freed via `free`.
+
+- `size_t oe_allocator_malloc_usable_size(void* ptr)` </br>
+  This function is called by `oecore` to implement [malloc_usable_size](http://man7.org/linux/man-pages/man3/malloc_usable_size.3.html). </br>
+  This function must be implemented.
+
+
+
+In comparison to GNU C Library, `oecore` does not require the allocator to implement `memalign`, `palloc` and `valloc`. These functions are obsolete and are will not be available.


docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
+Pluggable Allocators
+=====
+
+Open Enclave provides a default memory allocator for enclaves.
+This default allocator, [dlmalloc (Doug Lea Malloc)](http://gee.cs.oswego.edu/dl/html/malloc.html), is not performant in multi-thread applications.
+
+There are various high-performance memory allocators that try to optimize, among other things, multi-thread scenarios:
+- [The GNU Allocator](https://www.gnu.org/software/libc/manual/html_node/The-GNU-Allocator.html) which is based on `pthreads malloc`, which is in turn based on `dlmalloc`.
+- [Thread-Caching Malloc (tcmalloc)](https://gperftools.github.io/gperftools/tcmalloc.html) developed by Google.
+- [snmalloc](https://github.com/Microsoft/snmalloc) developed by Micrsoft Research.
```
achamayou: Typo: Microsoft

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+
+- This design does not aim to address what the ideal default allocator for Open Enclave should be.
+- Though the design could be extended to support multiple pluggable allocators, each managing a portion of the heap, it is currently a non-requirement.
+
+Requirements
+----
+
+The design for `Pluggable Allocators` additionally has the following requirements,
+
+- Allocators must be pluggable by an SDK user with having to recompile the OE SDK </br>
```
achamayou: s/with/without?

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+  This is similat to the GNU C library in that, one need not rebuild glibc to plugin a malloc replacement.
```
achamayou: Typo: similar

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+  From a design perspective, this precludes the use of architecture specific datastructures in the plugin interface.
+
+- The design must support allocators to play well with features that Open Enclave SDK already provides (`debug allocators`, `backtracing`, `stack unwinding`, `failure callbacks` etc)
+
+- The design must allow allocators to perform thread-specific optimizations since multi-threaded applications benefit a lot from fine-tuned allocators.
+
+- Allocators must rely only on published interfaces and datastructures.
+
+- The design must allow for Open Enclave core (`liboecore`) to own a portion of the heap and have the allocator manage rest of the heap. </br>
+  This allows the SDK to manage a bit ofmemory for efficiently dispatching ecalls/ocalls, logging and other core features, better than what a general purpose
```
achamayou: Missing space in ofmemory

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+  ELF  `weak symbols` (see [Function Attributes](https://gcc.gnu.org/onlinedocs/gcc-3.2/gcc/Function-Attributes.html)) are designed to allow user code to override functions defined in libraries.
+  Using `weak symbols` for default implementation of these function would cause the linker to look for custom replacement functions irrespective of the order.
+
+
+Replacement functions are not required to be reentrant.
+
+Undefined Behavior
+---
+
+- The replacement functions should not do much other than managing memory. Making ocalls, logging, printf etc may use dynamic memory and hence result in reentranct calls to the allocator.
```
achamayou: Typo: reentrant (or perhaps re-entrant?)

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+
+
+Precedents
+------
+
+- The GNU C Library provides a way to replace its malloc implementation, documented in [Replacing Malloc](https://www.gnu.org/software/libc/manual/html_node/Replacing-malloc.html#Replacing-malloc). </br>
+  To replace the allocator, the following replacement functions must be provided `malloc`, `free`, `calloc`, `realloc`, `aligned_alloc`, `malloc_usable_size`, `memalign`, `posix_memalign`, `palloc`, and `valloc`. </br>
+  Replacement happens by statically linking the allocator library before the GNU C Library.
+
+- The CCF team has implemented compiling the Open Enclave SDK to use `snmalloc` (Pull Request #2649). Currently, users must recompile OE SDK with `USE_SNMALLOC` to leverage this.
```
mjp41: Make this a link to the PR?

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+The design for `Pluggable Allocators` additionally has the following requirements,
+
+- Allocators must be pluggable by an SDK user with having to recompile the OE SDK </br>
+  This is similat to the GNU C library in that, one need not rebuild glibc to plugin a malloc replacement.
+
+- Must work on various TEE environments that the Open Enclave SDK will support. </br>
+  Note, it does not mean that a given allocator must work on all platforms; rather Open Enclave SDK must support plugging allocators on all platforms. </br>
+  From a design perspective, this precludes the use of architecture specific datastructures in the plugin interface.
+
+- The design must support allocators to play well with features that Open Enclave SDK already provides (`debug allocators`, `backtracing`, `stack unwinding`, `failure callbacks` etc)
```
mjp41: So these feel like two different things:

* debug allocators
* failure call backs

Are layered on top of the allocator.  While

* backtracing 
* stack unwinding 

Are core platform features that require allocation and often during failure so care is needed.

I would suggest that you actually make `debug allocators` and `failure callbacks` something that the allocator chooses to provide.  Many allocators have considerably more security features than your debug allocator, and by providing them as part of the allocator can do it considerably more efficiently.  For instance, `snmalloc` can detect leaks during teardown with no runtime overhead, by carefully checking each threads data-structures. Similarly for failure call-backs this can be done deep inside the allocator again taking checking off the critical path.

So I would be careful what you layer on top of a high-perf allocator:  `snmalloc`s fast path allocation is ~15 instructions with two not-taken conditional branches.

docs/DesignDocs/PluggableAllocators.md
anakrish: Ah. Looks like putting #2649 does not work when rendered as markdown. Will change it to full link to PR.

docs/DesignDocs/PluggableAllocators.md
anakrish: @mjp41 Good point about allocators themselves providing debug/leak detection capabilities. We can extend the contract in next iterations to allow allocators to expose more of their debugging capabilities.

anakrish: @mjp41 Couple of questions about snmalloc
- Does snmalloc pinpoint where the leaks originate by tracking the callstack of alloctions?
- How are you counting the instructions in fast-path? Do you enforce that all the function calls are inlined in the fast path?

Can you provide us with a small sample that can show case snmalloc? I can add that to our samples in one of the subsequent PRs.

anakrish:
@CodeMonkeyLeet What would be a good location to publish `allocator.h`?

@HernanGatta Can you see if this proposal makes sense for OPTEE?

docs/DesignDocs/PluggableAllocators.md
mjp41: So at the moment we only use the function to check everything has been deallocated in unit tests:
  https://github.com/microsoft/snmalloc/blob/51e919385fbb30a29e2ab815b9c122e0fbd7cf4b/src/test/func/memory/memory.cc#L115

There are also issues that function that do not deallocate until very late can cause this not to work.  For instance, `at_exit` allocates a block, which is not deallocated until after most stuff is torn down.  But in this case we could put it in the enclave tear down. 

We don't currently track the allocation sites.  ASAN's allocator has some nice features for doing that.  But we have not had the time to do that with `snmalloc`.

Re: Fast path.  `snmalloc` has two macros:
```
SNMALLOC_FAST_PATH
SNMALLOC_SLOW_PATH
```
https://github.com/microsoft/snmalloc/blob/0a081cacd649e138cb83edb56f773c7ecf6b6e2d/src/ds/defines.h#L8-L9

We use this to careful force inlining, and prevent other inlining. A lot of the code is written to cause the compiler to force tail call optimisation:
```C++
   if (likely(condition))
      return fast(size);
   return slow(size);
```
Where `fast` is declared with `SNMALLOC_FAST_PATH` and `slow` with `SNMALLOC_SLOW_PATH`. I am in the process of refactoring the slow paths to be of this form, so we get better codegen for those too.

This then gives a very nice allocation sequence:
```asm
0000000000028930 <malloc>:
   28930:	48 89 fe             	mov    %rdi,%rsi
   28933:	48 8b 05 ce 28 00 00 	mov    0x28ce(%rip),%rax        # 2b208 <_DYNAMIC+0x1e0>
   2893a:	64 48 8b 38          	mov    %fs:(%rax),%rdi
   2893e:	48 8d 46 ff          	lea    -0x1(%rsi),%rax
   28942:	48 3d ff ff 00 00    	cmp    $0xffff,%rax
   28948:	77 20                	ja     2896a <malloc+0x3a>
   2894a:	48 83 e0 f8          	and    $0xfffffffffffffff8,%rax
   2894e:	48 8d 0d 7b 86 fe ff 	lea    -0x17985(%rip),%rcx        # 10fd0 <_ZN8snmallocL18sizeclass_metadataE>
   28955:	48 8b 34 08          	mov    (%rax,%rcx,1),%rsi
   28959:	48 8b 04 f7          	mov    (%rdi,%rsi,8),%rax
   2895d:	48 85 c0             	test   %rax,%rax
   28960:	74 0d                	je     2896f <malloc+0x3f>
   28962:	48 8b 08             	mov    (%rax),%rcx
   28965:	48 89 0c f7          	mov    %rcx,(%rdi,%rsi,8)
   28969:	c3                   	retq   
   2896a:	e9 31 eb ff ff       	jmpq   274a0 <_ZN8snmalloc9AllocatorIXadL_ZNS_20needs_initialisationEPvEEXadL_ZNS_21init_thread_allocatorEvEENS_24MemoryProviderStateMixinINS_8PALLinuxEEENS_15DefaultChunkMapINS_21GlobalPagemapTemplateINS_11FlatPagemapILm24EhEEEEEELb1EE15alloc_not_smallILNS_7ZeroMemE0ELNS_12AllowReserveE1EEES1_m>
   2896f:	48 8b 87 c0 0d 00 00 	mov    0xdc0(%rdi),%rax
   28976:	48 39 87 c8 0d 00 00 	cmp    %rax,0xdc8(%rdi)
   2897d:	75 05                	jne    28984 <malloc+0x54>
   2897f:	e9 2c ec ff ff       	jmpq   275b0 <_ZN8snmalloc9AllocatorIXadL_ZNS_20needs_initialisationEPvEEXadL_ZNS_21init_thread_allocatorEvEENS_24MemoryProviderStateMixinINS_8PALLinuxEEENS_15DefaultChunkMapINS_21GlobalPagemapTemplateINS_11FlatPagemapILm24EhEEEEEELb1EE26small_alloc_next_free_listILNS_7ZeroMemE0ELNS_12AllowReserveE1EEES1_m>
   28984:	e9 c7 ec ff ff       	jmpq   27650 <_ZN8snmalloc9AllocatorIXadL_ZNS_20needs_initialisationEPvEEXadL_ZNS_21init_thread_allocatorEvEENS_24MemoryProviderStateMixinINS_8PALLinuxEEENS_15DefaultChunkMapINS_21GlobalPagemapTemplateINS_11FlatPagemapILm24EhEEEEEELb1EE19small_alloc_mq_slowILNS_7ZeroMemE0ELNS_12AllowReserveE1EEES1_m>
   28989:	0f 1f 80 00 00 00 00 	nopl   0x0(%rax)
```

> Can you provide us with a small sample that can show case snmalloc?

I would love to, but I am not entirely sure what you are after. Do you mean allocation patterns that work really well for `snmalloc`?

anakrish: Yes. Allocation patterns would be good.

It would be good to have a sample that demonstrates how an user could plug-in their own allocator. It makes sense to show case `snmalloc` as the custom allocator since it already has a PAL for Open Enclave.

The sample could be as simple as doing a bunch of allocations; however is there something about `snmalloc` that you would want to showcase via a sample?

anakrish: >
> We don't currently track the allocation sites. ASAN's allocator has some nice features for doing that. But we have not had the time to do that with snmalloc.
>

I think it is better that Open Enclave SDK stops at supporting pluggable allocators. It does not seem to add much value to funnel all the debugging capabilities of a particular allocator via some unified debug api in `allocator.h`. Once the application dependent allocator has been plugged in, the use can interface with the debugging capabilities of the allocator directly.

To reiterate, we can revisit user's interaction with debugging capabilities of custom allocators in future iterations.

Note: Currently `debug allocator` which tracks allocation sites is not a public feature. It is only used internally for detecting leaks in the SDK.

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+The replacement functions can be provided
+- via object files. This happens when the allocator is not a separate static library.
+- via a separate allocator static library. In this case, the allocator library must be specified before `oelibc` in the linker command line. </br>
+  ELF  `weak symbols` (see [Function Attributes](https://gcc.gnu.org/onlinedocs/gcc-3.2/gcc/Function-Attributes.html)) are designed to allow user code to override functions defined in libraries.
+  Using `weak symbols` for default implementation of these function would cause the linker to look for custom replacement functions irrespective of the order.
+
+
+Replacement functions are not required to be reentrant.
+
+Undefined Behavior
```
mingweishih: Can we rename this section is other like "Security disclaimer"? I think in somewhere we should emphasize that OE does not provide any security assurance of using a plugged allocators.

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
+Pluggable Allocators
+=====
+
+Open Enclave provides a default memory allocator for enclaves.
+This default allocator, [dlmalloc (Doug Lea Malloc)](http://gee.cs.oswego.edu/dl/html/malloc.html), is not performant in multi-thread applications.
+
+There are various high-performance memory allocators that try to optimize, among other things, multi-thread scenarios:
+- [The GNU Allocator](https://www.gnu.org/software/libc/manual/html_node/The-GNU-Allocator.html) which is based on `pthreads malloc`, which is in turn based on `dlmalloc`.
+- [Thread-Caching Malloc (tcmalloc)](https://gperftools.github.io/gperftools/tcmalloc.html) developed by Google.
```
dthaler: Why is who it's developed by relevant?  Line 8 didn't say who that one was developed by.  Recommend deleting the developed by in lines 9 and 10.  I can imagine that which LICENSE it's under might be relevant, but the developer should be immaterial here.

docs/DesignDocs/PluggableAllocators.md
dthaler: Same comment.  Why does it matter which "team" implemented it?   In a permanent design doc, I don't see why that should matter.

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+----
+
+The design for `Pluggable Allocators` additionally has the following requirements,
+
+- Allocators must be pluggable by an SDK user with having to recompile the OE SDK </br>
+  This is similat to the GNU C library in that, one need not rebuild glibc to plugin a malloc replacement.
+
+- Must work on various TEE environments that the Open Enclave SDK will support. </br>
+  Note, it does not mean that a given allocator must work on all platforms; rather Open Enclave SDK must support plugging allocators on all platforms. </br>
+  From a design perspective, this precludes the use of architecture specific datastructures in the plugin interface.
```
dthaler: nit: data structures should be two words

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+- All the replacement functions must be provided, or none. A partial set will result in both the default allocator and the custom allocator managing the heap.
+
+
+Replacement Functions
+---
+
+Published in `openenclave/enclave/allocator.h`
+
+- `void oe_allocator_init(void* heap_start_address, void* heap_end_address)` </br>
+  This function will be called by `oecore` at an appropriate time during enclave initialization. The total heap memory available to the enclave is specified in its configuration file
```
dthaler: The part about "in its configuration file" is not true, when OE_SET_ENCLAVE_SGX or OE_SET_ENCLAVE_OPTEE are used instead.  (See also issue #2714)

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+  In the future, if `oecore` needs to hold onto any thread-specific data on behalf of the allocator, it can be returned in this function. Currently the recommendation is to use `__thread` variables to store thread-specific allocator data.
+
+
+- `void oe_allocator_thread_cleanup(void)` </br>
+  This function will be called by `oecore` just prior to the termination of enclave thread. Thread-specific exit functions (`__cxa_thread_atexit` functions) will be executed prior to `oe_allocator_thread_cleanup`. </br>
+  `oe_allocator_thread_cleanup` may not use `__cxa_thread_atexit`; that is it cannot access C++ thread-local variables with non-trivial destructors for the first time. </br>
+  It is important that the allocators be aware of the nature of enclave threads, and be designed accordingly. Enclave threads can be short-lived. For OE SGX, enclave threads last only for the duration of an ecall. </br>
+  Additionally, the same thread-control structure (`sgx_tcs_t`) can be bound to different host threads at different times during the lifespan of an enclave.
```
dthaler: sgx_tcs_t is just an SGX-specific example.  Either insert "e.g." before it, or combine this into the previous sentence that already starts "For OE SGX" if you think this statement really is SGX specific.

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+
+Replacement Functions
+---
+
+Published in `openenclave/enclave/allocator.h`
+
+- `void oe_allocator_init(void* heap_start_address, void* heap_end_address)` </br>
+  This function will be called by `oecore` at an appropriate time during enclave initialization. The total heap memory available to the enclave is specified in its configuration file
+  (e.g [hello_worlf.conf](https://github.com/openenclave/openenclave/blob/e2e440ec134d37e107b5a296fdc508a3a643598e/samples/helloworld/enclave/helloworld.conf#L6)) </br>
+  `oecore` may reserve a small portion for its own use and pass the rest to this replacement function. The allocator is expected to perform any global initialization in this function. </br>
```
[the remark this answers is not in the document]
mjp41: I would be concerned about have a dedicated heap as you may introduce a concurrency bottleneck.  I was reading this as you allocate some static datastructures in this space, and pass the rest to the pluggable heap.

docs/DesignDocs/PluggableAllocators.md
mjp41: > Once the application dependent allocator has been plugged in, the use can interface with the debugging capabilities of the allocator directly.

Agreed, I think the point I should have made is: "don't force pluggable allocators to work with the debug allocator in OE, or the failure callbacks."  They can probably do something better if required.

CodeMonkeyLeet:
> @CodeMonkeyLeet What would be a good location to publish `allocator.h`?

I would spin up a new folder based on the new category for headers intended for enclave runtime library customization; something like `include/openenclave/addin` or `include/openenclave/advanced` for parity with the `@advanced` tag introduced recently.

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+  Using `weak symbols` for default implementation of these function would cause the linker to look for custom replacement functions irrespective of the order.
+
+
+Replacement functions are not required to be reentrant.
+
+Undefined Behavior
+---
+
+- The replacement functions should not do much other than managing memory. Making ocalls, logging, printf etc may use dynamic memory and hence result in reentranct calls to the allocator.
+- All the replacement functions must be provided, or none. A partial set will result in both the default allocator and the custom allocator managing the heap.
```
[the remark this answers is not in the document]
jbdelcuv: I agree, this requires more investigation.
The linker may not throw a multiple definition error because of the weak symbol and mix API from different libraries.

docs/DesignDocs/PluggableAllocators.md
anakrish: Fixed.

docs/DesignDocs/PluggableAllocators.md
anakrish: Fixed.

docs/DesignDocs/PluggableAllocators.md
anakrish: Fixed.

docs/DesignDocs/PluggableAllocators.md
anakrish: Fixed.

docs/DesignDocs/PluggableAllocators.md
anakrish: Fixed.

docs/DesignDocs/PluggableAllocators.md
anakrish: > For my understanding, is the conclusion of this discussion that pluggable allocators:
> 
>     * need to provide support for backtracing & stack unwinding
> 
>     * do not need to provide debug allocators and failure callbacks
> 
> 
> For the latter case, given that debug allocator use in OE SDK is internal and not compatible with pluggable allocators, is this an issue for CI/CD configuration in running tests against pluggable allocators?

Here are the answers:
- No allocators do not need to provide support for backtracing and stack unwinding. I will remove that confusing line from the doc.
- Allocators do not need to provide "debug alloctors" and "Failure callbacks". OE will provide these functionality atop the custom allocators, as it currently does atop dlmalloc.

- Debug allocators built atop custom allocators will not cause issues for CI/CD. We do need to lock down the custom allocator support in our CI/CD configuration.

docs/DesignDocs/PluggableAllocators.md
anakrish: Thanks for the suggestions. Auto incorporated.

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
+Pluggable Allocators
+=====
+
+Open Enclave provides a default memory allocator for enclaves.
+This default allocator, [dlmalloc (Doug Lea Malloc)](http://gee.cs.oswego.edu/dl/html/malloc.html), is not performant in multi-thread applications.
```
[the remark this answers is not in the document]
anakrish: Good suggestion. I have mentioned public domain, popularity, and easily modified to work within enclaves as reasons. The choice of dlmalloc however predates my association with Open Enclave.

anakrish:
> LGTM. Do you want to add the specifications for malloc/free/... to this document?

Good point. Added references to those functions.

anakrish:
> > @CodeMonkeyLeet What would be a good location to publish `allocator.h`?
> 
> I would spin up a new folder based on the new category for headers intended for enclave runtime library customization; something like `include/openenclave/addin` or `include/openenclave/advanced` for parity with the `@advanced` tag introduced recently.

Thanks for the suggestion. I like `advanced` more than `addin` since plugging in a custom allocator is an advanced use. 
However, we can always change the location later if someone feels strongly about it.

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+
+
+- `void* oe_allocator_realloc(void* ptr, size_t size)` </br>
+  This function will be called by `oecore` to implement `realloc`.
+
+
+- `void oe_allocator_posix_memalign(void** memptr, size_t alignment, size_t size)` </br>
+  This function will be called by `oecore` to implement [posix_memalign](https://linux.die.net/man/3/posix_memalign). Memory allocated by this function can be freed via `free`.
+
+- `void oe_allocator_reusable_size(void* ptr)` </br>
```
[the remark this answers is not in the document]
anakrish: > Can you elaborate on what this is supposed to do? Why "reusable size" instead of just having the allocator implement `oe_allocator_malloc_usable_size()` directly?
> 
That ought to have been `oe_allocator_malloc_size`. Fixed.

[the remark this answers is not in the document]
anakrish: I added the following lines to the doc. 
"The set of replacement functions are mostly equivalents of the functions required by the GNU C Library for plugging-in custom allocators. Additionally, there are replacement functions that allow the allocator to work well with enclave threads."

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+  This function is called by `oecore` to implement [malloc_usable_size](http://man7.org/linux/man-pages/man3/malloc_usable_size.3.html). </br>
+  This function must be implemented. However, allocators may choose to provide an empty implementation.
+
+
+
+In comparison to GNU C Library, `oecore` does not require the allocator to implement `memalign`, `palloc` and `valloc`. These functions are obsolete.
+`oecore` will implement `aligned_alloc` and `memalign` by incorporating additional constraints atop `oe_allocator_posix_memalign`.
```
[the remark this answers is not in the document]
anakrish: Both points are valid.
I have added `oe_allocator_aligned_alloc` to the set of replacement functions.
Removed statement that `oecore` will implement `memalign`. Instead mentioned that `memalign`, `palloc` and `valloc` will not be available.

Note: MUSL uses `memalign` as the basis for implementing `posix_memalign` and `aligned_alloc`. https://github.com/openenclave/openenclave/tree/6ee7306a84b722820cbbb82602e3ebd9d0ccc207/3rdparty/musl/musl/src/malloc

Once we plugin couple of allocators, we will know if our choice of replacement function was right.

docs/DesignDocs/PluggableAllocators.md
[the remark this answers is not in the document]
anakrish: @CodeMonkeyLeet the document intentionally does not specify how the memory will be used. `oecore` does not make a commitment that all of the heap will be managed by the allocator.
This gives `oecore` the flexibility to use a portion of the heap efficiently. It could keep data at fixed locations in its portion, or partition it per thread or use it to guarantee that `oecore` functions will never run out of memory. Or it may decide to pass all of the memory to the allocator.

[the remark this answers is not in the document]
anakrish: @mjp41 
I am resolving this conversation since I don't think it is a blocker. We would be discussing potential uses of `oecore`'s portion of the heap, that this proposal does not make.

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+  Note, it does not mean that a given allocator must work on all platforms; rather Open Enclave SDK must support plugging allocators on all platforms. </br>
+  From a design perspective, this precludes the use of architecture specific datastructures in the plugin interface.
+
+- The design must support allocators to play well with features that Open Enclave SDK already provides (`debug allocators`, `backtracing`, `stack unwinding`, `failure callbacks` etc)
+
+- The design must allow allocators to perform thread-specific optimizations since multi-threaded applications benefit a lot from fine-tuned allocators.
+
+- Allocators must rely only on published interfaces and datastructures.
+
+- The design must allow for Open Enclave core (`liboecore`) to own a portion of the heap and have the allocator manage rest of the heap. </br>
```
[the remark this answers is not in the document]
anakrish: This requirement is achieved via `oe_allocator_init`. I have tried to explain how it works there.

The requirement for a part of the heap to be owned by `oecore` allows us to experiment with some ideas: stronger memory guarantess in oecore (never out of memory); multiple allocators;
allow supporting sandboxed enclaves etc.

I agree, we need to extended the API surface and harden this requirement once we support SGX 2.


I am unclear the impact SGX 2 will have on this design. I do suspect SGX 2 will impact a lot more than just memory allocators. It would be pragmatic to extend the interface
Do let me know if more clarification is needed. We can tweak the wording in subsequent PRs.

I agree that this design does not cover SGX V2; and we would need to expand the interface to support SGX V2. I don't think,

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+If no replacement functions are provided, then the default allocator is used. This retains the current behavior in which all enclaves get the default allocator without any additional linker options.
+
+The replacement functions can be provided
+- via object files. This happens when the allocator is not a separate static library.
+- via a separate allocator static library. In this case, the allocator library must be specified before `oelibc` in the linker command line. </br>
+  ELF  `weak symbols` (see [Function Attributes](https://gcc.gnu.org/onlinedocs/gcc-3.2/gcc/Function-Attributes.html)) are designed to allow user code to override functions defined in libraries.
+  Using `weak symbols` for default implementation of these function would cause the linker to look for custom replacement functions irrespective of the order.
+
+
+Replacement functions are not required to be reentrant.
```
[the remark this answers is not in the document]
anakrish: The C standard does not require that any of these functions be re-entrant; nor does it require that the functions be non-reentrant.
I don't think we get any better semantics by requiring that the replacement functions be non-reentrant. We can definitely revisit this in the future.

docs/DesignDocs/PluggableAllocators.md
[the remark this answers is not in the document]
anakrish: @CodeMonkeyLeet @jbdelcuv
Valid concerns about weak symbols. The only thing weak symbols buy us is to avoid the user from having the specify the allocator library before the C library (oelibc) in the linker line.
GNU C library requires precise ordering and thus it is also reasonable for us to require ordering of the libraries in the linker line.

I have changed the document to say that we will not be using weak symbols since that might lead to a mix of APIs. Also mentioned that providing a partial set of replacement functions would result in linker error.

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,142 @@
...
+
+
+- `void oe_allocator_thread_init(void)` </br>
+  This function will be called by `oecore` during initialization of an enclave thread, after the thread-local variables for the thread have been initialized. </br>
+  In the future, if `oecore` needs to hold onto any thread-specific data on behalf of the allocator, it can be returned in this function. Currently the recommendation is to use `__thread` variables to store thread-specific allocator data.
+
+
+- `void oe_allocator_thread_cleanup(void)` </br>
+  This function will be called by `oecore` just prior to the termination of enclave thread. Thread-specific exit functions (`__cxa_thread_atexit` functions) will be executed prior to `oe_allocator_thread_cleanup`. </br>
+  `oe_allocator_thread_cleanup` may not use `__cxa_thread_atexit`; that is it cannot access C++ thread-local variables with non-trivial destructors for the first time. </br>
```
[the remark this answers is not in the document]
anakrish: After multiple attempts at wording which required explaining C++11's intricate details, here is what I came up with:

- `void oe_allocator_thread_cleanup(void)` </br>
This function will be called by `oecore` just prior to the termination of enclave thread. 
  Thread-specific exit functions (`__cxa_thread_atexit` functions) will be executed prior to `oe_allocator_thread_cleanup`. </br>
  This allows the thread-specific exit functions to use memory management functions (e.g to free resources) before the allocator itself is cleaned up. </br>
  It is recommended that allocators be judicious about using C++11 thread-specific objects in `oe_allocator_thread_cleanup` since 
  use of C++11 thread-local objects may trigger a call to (`__cxa_thread_atexit`) after all the thread-specific exit functions have executed.

docs/DesignDocs/PluggableAllocators.md
anakrish: While it is a valid suggestion, I don't know if OE's disclaimer about security of custom allocators would help much.Users that are plugging in some allocator or any other code for that matter into an enclave cannot be protected by OE. 

Did you have any specific scenario in mind that the disclaimer would help?

anakrish:
Folks, thanks for the review, suggestions and feedback.
I think I have addressed all of them. 

CC: @dthaler

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,149 @@
...
+
+- The design must allow allocators to perform thread-specific optimizations since multi-threaded applications benefit a lot from fine-tuned allocators.
+
+- Allocators must rely only on published interfaces and datastructures.
+
+- The design must allow for Open Enclave core (`liboecore`) to own a portion of the heap and have the allocator manage rest of the heap. </br>
+  This allows the SDK to manage a bit of memory for efficiently dispatching ecalls/ocalls, logging, and other core features, better than what a general-purpose
+  allocator would be capable of. Additionally, it also allows a clear memory demarcation between `core memory` and `application memory`.
+
+- Performance and safety. Plugin approaches often involve registering `hook` functions that will be called by the SDK via an indirect calls.
```
dthaler, line 57:
```suggestion
- Performance and safety. Plugin approaches often involve registering `hook` functions that will be called by the SDK via indirect calls.
```

docs/DesignDocs/PluggableAllocators.md
```diff
@@ -0,0 +1,149 @@
...
+
+- `void* oe_allocator_aligned_alloc(size_t alignment, size_t size)` </br>
+  This function will be called by `oecore` to implement [aligned_alloc](https://linux.die.net/man/3/aligned_alloc/man/3/posix_memalign). Memory allocated by this function can be freed via `free`.
+
+- `void oe_allocator_posix_memalign(void** memptr, size_t alignment, size_t size)` </br>
+  This function will be called by `oecore` to implement [posix_memalign](https://linux.die.net/man/3/posix_memalign). Memory allocated by this function can be freed via `free`.
+
+- `size_t oe_allocator_malloc_usable_size(void* ptr)` </br>
+  This function is called by `oecore` to implement [malloc_usable_size](http://man7.org/linux/man-pages/man3/malloc_usable_size.3.html). </br>
+  This function must be implemented. However, allocators may choose to provide an empty implementation.
```
dthaler, line 145: It can't really be empty since it has to return a value.  Maybe you mean "an implementation that always returns 0"?  Or what?   Simplest way to resolve is to delete the sentence about empty.
```suggestion
  This function must be implemented.
```

kuanggang-git:

docs/DesignDocs/PluggableAllocators.md
anakrish: Good point. Thanks for the suggestion.

anakrish:
Incorporated Dave's suggestion.

PR #2723, opened 2020-03-22
Prevent regeneration of asym_keys.h

anakrish:
asym_keys.h was getting regenerated every time you do a make,
causing recompilation of entire project.

### DIFF
--- a/host/CMakeLists.txt
+++ b/host/CMakeLists.txt
@@ -6,39 +6,36 @@ if (OE_SGX)
     set(SGX_EDL_DIR ${CMAKE_SOURCE_DIR}/common/sgx/edl)
 endif()
 
+set(OE_EDL_INCLUDES ${OE_INCDIR}/openenclave/bits)
+
 ##==============================================================================
 ##
 ## These rules generate the edge routines for the internal TEE-agnostic
 ## ECALLs/OCALLs used by liboehost/liboecore.
+##
+## The generated _args.h is published in ${OE_EDL_INCLUDES}.
 ##
 ##==============================================================================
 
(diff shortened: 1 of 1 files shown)


host/CMakeLists.txt
```diff
@@ -6,39 +6,37 @@ if (OE_SGX)
...
 
-set(OE_EDL_INCLUDES ${OE_INCDIR}/openenclave/bits)
 # Add dependencies to `oe_includes` so that the published header will be ready before
 # other files consume it.
 add_dependencies(oe_includes tee_untrusted_edl)
 
-add_custom_command(
-    TARGET tee_untrusted_edl
-    COMMAND ${CMAKE_COMMAND} -E copy tee_args.h ${OE_EDL_INCLUDES}/asym_keys.h)
+add_custom_target(tee_untrusted_edl
```
mingweishih: Shouldn't `add_custom_target` be putted before `add_dependencies(oe_includes tee_untrusted_edl)`?

mingweishih:
Just wondering would this fix still needed with Jordan's refactoring work on #2683.
@jhand2 @anakrish Can you double check?

host/CMakeLists.txt
jhand2: I think cmake will reorder them for you, but ya for readability thats probably best

jhand2:
> Just wondering would this fix still needed with Jordan's refactoring work on #2683.
> @jhand2 @anakrish Can you double check?

It will move but I think its still necessary. This should get merged first and then I'll rebase

host/CMakeLists.txt
anakrish: Good catch. Fixed.

PR #2725, opened 2020-03-23
Fix fd not being deleted from epoll on close

thomasten:
If an fd is closed, it is implicitly deleted from all epoll instances
by the system. The enclave epoll mapping must also be updated.

### DIFF
--- a/syscall/unistd.c
+++ b/syscall/unistd.c
@@ -261,7 +261,7 @@ static void _close_epoll_callback(oe_fd_t* desc, void* arg)
     oe_assert(desc);
     oe_assert(desc->type == OE_FD_TYPE_EPOLL);
 
-    const int fd = (int)arg;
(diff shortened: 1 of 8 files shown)


syscall/unistd.c
```diff
@@ -265,7 +276,13 @@ int oe_close(int fd)
         OE_RAISE_ERRNO(oe_errno);
 
     if ((ret = desc->ops.fd.close(desc)) == 0)
+    {
+        // Notify epoll instances that this fd has been closed.
+        oe_fdtable_foreach(
```
anakrish: Does it matter whether the notification happens after the close or whether it happens immediately prior to close?

radhikaj:
Thanks @thomasten

syscall/unistd.c
thomasten: I'm not sure. I've put it after the close because the close could fail.

PR #2726, opened 2020-03-23
Fix loop condition in _epoll_ctl_del

thomasten:
Add a missing comparison in a for loop condition. However, this "bug"
will probably never occur because the map element should always be
found.

### DIFF
--- a/syscall/devices/hostepoll/hostepoll.c
+++ b/syscall/devices/hostepoll/hostepoll.c
@@ -362,7 +362,7 @@ static int _epoll_ctl_del(epoll_t* epoll, int fd)
     {
         bool found = false;
 
-        for (size_t i = 0; epoll->map_size; i++)
+        for (size_t i = 0; i < epoll->map_size; i++)
         {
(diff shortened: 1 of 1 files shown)


jhand2:
Good catch, thanks!

syscall/devices/hostepoll/hostepoll.c
```diff
@@ -362,7 +362,7 @@ static int _epoll_ctl_del(epoll_t* epoll, int fd)
     {
         bool found = false;
 
-        for (size_t i = 0; epoll->map_size; i++)
+        for (size_t i = 0; i < epoll->map_size; i++)
```
anakrish: @mingweishih Is there some compiler flag that can detect such coding errors?

mingweishih: I don't this can be detected with compiler flags as this is legitimate in syntax-wise. However, some static analysis tools should help (maybe Clang already supports some).

PR #2733, opened 2020-03-24
add test cases about child thread

jinghe-INTC:


jinghe-INTC:
> Another great test!
> 
> On the host side, could you use C++ `std::thread` instead of `pthread`? We don't have Windows host support for pthread so using `std::thread` allows us to more easily run this test for Windows as well.



Thanks. I have updated the code.

jhand2:
CI is currently down

bors r-

jhand2:
@jinghe-INTC it looks like there are some failures for the new test you added. Could you take a look at the bors link when you get a chance?

jinghe-INTC:
> @jinghe-INTC it looks like there are some failures for the new test you added. Could you take a look at the bors link when you get a chance?

I have found the test failure in the bors log. It's interesting the result is different with that from my local machine. I have updated the case and verified on an Azure Linux VM.

radhikaj:
Bors r+

jinghe-INTC:
> Build failed
> 
> continuous-integration/jenkins/branch

I have no idea why the test failed, even after being verified on an Azure VM before. Some additional printings have been added for investigation.

jhand2:
bors try

jhand2:
@jinghe-INTC one thing i notices is on the host side you're including unistd.h which won't compile on Windows. Which function are you using from that header?

jhand2:
bors try-

jhand2:
> @jinghe-INTC one thing i notices is on the host side you're including unistd.h which won't compile on Windows. Which function are you using from that header?

Oh it's sleep, which won't work on Windows.

tests/child_thread/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+
+add_subdirectory(host)
+
+if (BUILD_ENCLAVES)
+	add_subdirectory(enc)
+endif()
+
+add_enclave_test(tests/child_thread_ecall child_thread_host child_thread_enc 0)
+add_enclave_test(tests/child_thread_destory child_thread_host child_thread_enc 1)
+add_enclave_test(tests/child_thread_destory_check child_thread_host child_thread_enc 2)
```
CodeMonkeyLeet:
```suggestion
add_enclave_test(tests/child_thread_destroy child_thread_host child_thread_enc 1)
add_enclave_test(tests/child_thread_destroy_check child_thread_host child_thread_enc 2)
```

tests/child_thread/host/host.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include <unistd.h>
+#include <cstdio>
+#include <cstdlib>
+#include <thread>
+
+#include "child_thread_u.h"
+
+#define ECALL_IN_CHILD_THREAD 0
+#define DESTORY_IN_CHILD_THREAD 1
+#define DESTORY_N_CHECK_CHILD_THREAD 2
```
CodeMonkeyLeet:
```suggestion
#define DESTROY_IN_CHILD_THREAD 1
#define DESTROY_AND_CHECK_CHILD_THREAD 2
```

tests/child_thread/host/host.cpp
```diff
@@ -0,0 +1,111 @@
...
+    enclave = (oe_enclave_t*)lpParam;
+    uint32_t magic;
+    oe_result_t result = get_magic_ecall(enclave, &magic);
+    OE_TEST(magic == 0x1234);
+    OE_TEST(result == OE_OK);
+    multi_thread_flag = true;
+    return NULL;
+}
+
+static void* child_thread_destory(void* lpParam, oe_result_t expect_result)
```
CodeMonkeyLeet:
```suggestion
static void* child_thread_destroy(void* lpParam, oe_result_t expect_result)
```

tests/child_thread/host/host.cpp
```diff
@@ -0,0 +1,111 @@
...
+        case DESTORY_IN_CHILD_THREAD:
+            child_thread = std::thread(child_thread_destory, enclave, OE_OK);
+            child_thread.join();
+            OE_TEST(multi_thread_flag == true);
+            break;
+        case DESTORY_N_CHECK_CHILD_THREAD:
+            child_thread = std::thread(child_thread_stay, enclave);
+            sleep(1);
+            another_child_thread =
+                std::thread(child_thread_destory, enclave, OE_OUT_OF_THREADS);
```
CodeMonkeyLeet:
```suggestion
        case DESTROY_IN_CHILD_THREAD:
            child_thread = std::thread(child_thread_destroy, enclave, OE_OK);
            child_thread.join();
            OE_TEST(multi_thread_flag == true);
            break;
        case DESTROY_AND_CHECK_CHILD_THREAD:
            child_thread = std::thread(child_thread_stay, enclave);
            sleep(1);
            another_child_thread =
                std::thread(child_thread_destroy, enclave, OE_OUT_OF_THREADS);
```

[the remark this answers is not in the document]
radhikaj: Thanks @CodeMonkeyLeet

radhikaj:
bors delegate=jinghe-INTC

radhikaj:
@jinghe-INTC You can run tests against this PR. Just add a comment against the PR with "bors try".

tests/child_thread/enc/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+# Licensed under the MIT License.
+
+set (EDL_FILE ../child_thread.edl)
+
+add_custom_command(
+    OUTPUT child_thread_t.h child_thread_t.c
+    DEPENDS ${EDL_FILE} edger8r
+    COMMAND edger8r --trusted ${EDL_FILE} --search-path ${CMAKE_CURRENT_SOURCE_DIR})
+
+add_enclave(TARGET child_thread_enc UUID 688ab13f-5bc0-40af-8dc6-01d007fd2210 SOURCES enc.cpp ${CMAKE_CURRENT_BINARY_DIR}/child_thread_t.c)
```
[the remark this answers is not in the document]
jinghe-INTC: Thanks for the help. I will also check for other enclave projects I am working on.

tests/child_thread/CMakeLists.txt
jinghe-INTC: typos are fixed

tests/child_thread/host/host.cpp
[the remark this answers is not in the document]
jinghe-INTC: TCS number is not set to 4. 
multi_thread_flag seems not necessary for child_thread_stay. I remove it in child_thread_stay.

jinghe-INTC:
> > @jinghe-INTC one thing i notices is on the host side you're including unistd.h which won't compile on Windows. Which function are you using from that header?
> 
> Oh it's sleep, which won't work on Windows. You can use C++ sleep functions instead

Changed to C++11 sleep

jinghe-INTC:
bors try

jinghe-INTC:
bors try

jinghe-INTC:
> ## try
> # Build failed
> * [continuous-integration/jenkins/branch](https://oe-jenkins-dev.westeurope.cloudapp.azure.com/job/Bors/job/trying/165/display/redirect)

The failure looks like a bug of OE: when one thread stays in ecall, and another thread destroys the enclave, the first ecall should fail and return an error code, instead of segmentation fault.

tests/child_thread/README.md
```diff
@@ -0,0 +1,4 @@
+This directory tests enclave with child thread:
+1. Do ecall in child thread.
+2. Destroy enclave in child thread
```
anakrish: @jinghe-INTC It would be good to elaborate what behavior we are trying to lock down by each of these tests.
For example, via *2*, are we just trying to check if we can destroy an enclave from a thread other than the main thread or is there more to this test?

tests/child_thread/host/host.cpp
```diff
@@ -0,0 +1,109 @@
...
+#define ECALL_IN_CHILD_THREAD 0
+#define DESTROY_IN_CHILD_THREAD 1
+#define DESTROY_AND_CHECK_CHILD_THREAD 2
+
+bool multi_thread_flag = true;
+
+void stay_ocall(void)
+{
+    // sleep for 5 seconds.
+    std::this_thread::sleep_for(std::chrono::milliseconds(5000));
```
anakrish: If we really want to be sure that this thread will stay in ocall, sleeping for 5 seconds may not be enough since the thread scheduler could suspend other threads too for a long time if the machine is under a heavy load.

 It would be more robust to use a std::condition_variable instead so that the main thread can precisely wake this thread when it wants to.

tests/child_thread/host/host.cpp
```diff
@@ -0,0 +1,109 @@
...
+            child_thread = std::thread(child_thread_destroy, enclave, OE_OK);
+            child_thread.join();
+            OE_TEST(multi_thread_flag == true);
+            break;
+        case DESTROY_AND_CHECK_CHILD_THREAD:
+            child_thread = std::thread(child_thread_stay, enclave);
+            std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+            another_child_thread =
+                std::thread(child_thread_destroy, enclave, OE_OK);
+            child_thread.join();
```
anakrish: If using a condition variable, we'd call `notify_one` here before going on to wait for the thread to join.

tests/child_thread/host/host.cpp
```diff
@@ -0,0 +1,109 @@
...
+    return NULL;
+}
+
+static void* child_thread_stay(void* lpParam)
+{
+    oe_enclave_t* enclave = NULL;
+    enclave = (oe_enclave_t*)lpParam;
+    int res;
+    oe_result_t result = stay_in_ocall_ecall(enclave, &res);
+    fprintf(stdout, "stay_in_ocall_ecall returned: 0x%x\n", result);
```
anakrish: I am not sure what the expected behavior is. If another thread terminates the enclave while this thread is in an ecall, do we expect this ecall to succeed?

jinghe-INTC: No. From my experience, the ecall should fail, and return some understandable error code.

jinghe-INTC:
> My comments have been addressed, so I'm removing the request for changes, but I agree that the test failure resulting in a segfault is not expected code behavior and should be addressed before this set of tests should be merged.

I agree. 
And I just added a new case: it  does sleeping in the ecall at the first thread, instead of in the ecall-ocall. It has been tested on my local platform, and the result is also failed with SEGFAULT. I will do a "bors try" to verify.

jinghe-INTC:
bors try

anakrish:
@jinghe-INTC, Can you disable the failing tests and update the PR? We will enable the disabled tests when we fix #2775

jinghe-INTC:
> @jinghe-INTC, Can you disable the failing tests and update the PR? We will enable the disabled tests when we fix #2775

Sure

jinghe-INTC:
bors try

kuanggang-git:

jhand2:
@anakrish there is a build error:

`undefined symbol: std::__1::this_thread::sleep_for(std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000000l> > const&)`

tests/child_thread/enc/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+# Licensed under the MIT License.
+
+set (EDL_FILE ../child_thread.edl)
+
+add_custom_command(
+    OUTPUT child_thread_t.h child_thread_t.c
+    DEPENDS ${EDL_FILE} edger8r
+    COMMAND edger8r --trusted ${EDL_FILE} --search-path ${CMAKE_CURRENT_SOURCE_DIR})
+
+add_enclave(TARGET child_thread_enc UUID 99313d27-e655-4505-903b-dda158c4ee3e SOURCES enc.cpp ${CMAKE_CURRENT_BINARY_DIR}/child_thread_t.c)
```
mingweishih, line 11:
```suggestion
add_enclave(TARGET child_thread_enc CXX UUID 99313d27-e655-4505-903b-dda158c4ee3e SOURCES enc.cpp ${CMAKE_CURRENT_BINARY_DIR}/child_thread_t.c)
```

mingweishih:
bors try

jinghe-INTC:
bors try

PR #2735, opened 2020-03-24
Update the libc++ support

### LINKED ISSUE #2646: C++ functions missing in OE SDK
Intel SGX-SDK provides these missing functions and OE can consider adding them.
1.	cinttypes 
•	SGX SDK supports imaxabs() and imaxdiv(). OE does not.
2.	cmath 
•	OE does not support: acosh(), asinh(), fmal(), lgamma(), lgammaf(), sinh(), sinhl(), tgamma()
3.	functional 
•	The header exists in OE, but the OE documentation says it is not supported.
4.	tuple 
•	OE doc says there are known issues with tuple.apply (but doesn’t say what they are)
5.	utility 
•	OE doc says “as_const” is not supported. SGX SDK supports this in Linux when compiler used is C++17

mingweishih:
This PR updates the test cases and documentation of the libc++ support. The summary of the modifications is as follows.
- Move test cases from unsupported list to supported list (ran by nightly build only).
- Add a list of test cases that depend on `std=c++17` (the default is `std=c++14`) that allows cmake to build those test cases correctly..
- Update the documentation.

Fixes #2646


tests/libcxx/tests.broken
```diff
@@ -11,8 +11,11 @@
...
 ../../3rdparty/libcxx/libcxx/test/libcxx/diagnostics/nodiscard_extensions.pass.cpp
-# The following three tests fails on ACC Windows VM.
+# The following three tests fail on ACC Windows VM.
 ../../3rdparty/libcxx/libcxx/test/std/thread/thread.mutex/thread.lock/thread.lock.shared/thread.lock.shared.cons/mutex.pass.cpp
 ../../3rdparty/libcxx/libcxx/test/std/thread/thread.mutex/thread.lock/thread.lock.shared/thread.lock.shared.locking/lock.pass.cpp
 ../../3rdparty/libcxx/libcxx/test/std/thread/thread.mutex/thread.lock/thread.lock.unique/thread.lock.unique.locking/lock.pass.cpp
 ../../3rdparty/libcxx/libcxx/test/std/thread/thread.mutex/thread.mutex.requirements/thread.sharedtimedmutex.requirements/thread.sharedtimedmutex.class/lock_shared.pass.cpp
+# The following C++17 tests fail
+../../3rdparty/libcxx/libcxx/test/std/language.support/support.dynamic/new.delete/new.delete.array/new_align_val_t_nothrow_replace.pass.cpp
+../../3rdparty/libcxx/libcxx/test/std/language.support/support.dynamic/new.delete/new.delete.single/new_align_val_t_nothrow_replace.pass.cpp
```
jhand2: Do we know why they fail/in what circumstances?

mingweishih:
> Awesome! Do the new tests mostly just run during the nightly builds? I can't remember which set of tests run on every build vs nightly. (I'm worried about the new tests taking too long if they're run every time)

I only enable those newly added tests in the nightly build (`tests.supported`), so shouldn't impact the normal build (`tests.supported.default`) that bors runs.

docs/LibcxxSupport.md
```diff
@@ -1,5 +1,8 @@
 # Open Enclave Support for libcxx
 
+The support of libcxx is based on [LLVM libc++](https://libcxx.llvm.org/),
```
anakrish: @mingweishih Can you also add couple of lines about what kind of C++ library features will likely not work within enclaves and why.  As are reader, I would find such a description very helpful in determining what parts of the library I can potentially use within an enclave.

docs/LibcxxSupport.md
```diff
@@ -109,7 +112,7 @@ Header | Supported | Comments |
 :---:|:---:|:---|
 bit | No | Header is not provided, C++20 is not yet supported. |
 cfenv | Yes | Supported as part of C++11. |
-cmath | Partial | **Unsupported functions:** acosh(), asinh(), fmal(), lgamma(), lgammaf(), sinh(), sinhl(), tgamma(). |
+cmath | Partial | **Unsupported functions:** fmal() and tgamma(). |
```
anakrish: Can we instead support them and say that they don't fully pass our test-suite for long double values.

tests/libcxx/tests.broken
mingweishih: They failed on the following check, but I have no clue why.

https://github.com/openenclave/openenclave/blob/8fe3697e7fe6d85019609d2a3852fe4bc7bb6bc2/3rdparty/libcxx/libcxx/test/std/language.support/support.dynamic/new.delete/new.delete.array/new_align_val_t_nothrow_replace.pass.cpp#L84

mingweishih:
bors try

tests/libcxx/CMakeLists.txt
```diff
@@ -1,13 +1,46 @@
...
-# read tests.supported, sanitize the cpp-file, and create the test-case
+# Read tests.supported, sanitize the cpp-file, and create the test-case.
 if (ENABLE_FULL_LIBCXX_TESTS)
     file(STRINGS "tests.supported" alltests)
 else()
     file(STRINGS "tests.supported.default" alltests)
 endif()
 
+# List of unsupported tests with GCC.
+set(GCC_UNSUPPORTED_LIST
```
CodeMonkeyLeet: This list is a little surprising to me ... do we know why these fail for GCC? I would expect that GCC7 implemented pretty much all of the C++17 features already.

tests/libcxx/enc/CMakeLists.txt
```diff
@@ -83,6 +85,10 @@ function(add_libcxx_test_enc NAME CXXFILE)
     enclave_compile_definitions(libcxxtest-${NAME}_enc PRIVATE -DWITH_MAIN -D__TEST__="${CXXFILE}")
     enclave_link_libraries(libcxxtest-${NAME}_enc libcxxtest-support)
 
+    if (CXXFILE IN_LIST cxx17tests)
+        set_enclave_property(TARGET libcxxtest-${NAME}_enc PROPERTY CXX_STANDARD 17)
```
CodeMonkeyLeet: Is this the only place we need to set the compiler option? I would expect that if we claim to support C++17 that we should set it globally for the OE SDK as part of cmake/compiler_settings.cmake ... maybe this could also result in the gcc side failures since oelibcxx is being compiled with C++14 standard still.

mingweishih: Update from my experiments
- Applying C++17 globally works both for clang and gcc (all default libcxx tests passed)
- Some tests (full libcxx tests) failed with C++17. I can compile a list of such tests and see how many are they.

tests/libcxx/CMakeLists.txt
mingweishih: All the tests associated with "deduction" failed with gcc during the compile-time. I don't know the root cause yet.

mingweishih: Update:
Those test cases require certain compiler support, see

https://github.com/openenclave/openenclave/blob/2536d0cc736d7cbbbb9ab4193d37ea983ecd39d9/3rdparty/libcxx/libcxx/utils/libcxx/test/config.py#L461

According to  https://gcc.gnu.org/projects/cxx-status.html#cxx17, it is supported by gcc 8.

tests/libcxx/enc/CMakeLists.txt
mingweishih: Update: There are at least 73 cases that are not supported by C++17. This means that we always need to keep such a list regardless using C++14 or updating to C++17. However, the list of C++17-dependent (291 cases) does less than C++14-dependent one.

docs/LibcxxSupport.md
```diff
@@ -23,30 +29,30 @@ cstddef | Yes | - |
 cstdlib | Partial | **Unsupported functions:** at_quick_exit(), quick_exit() |
 ctime | Partial | All time functions implicitly call out to untrusted host for time values. The resulting time values should not be used for security purposes. <br> **Supported functions:** time(), gettimeofday(), clock_gettime(), nanosleep(). _Please note that clock_gettime() only supports CLOCK_REALTIME_ |
 chrono | Partial | Supported as part of C++11. All time functions implicitly call out to untrusted host for time values. The resulting time values should not be used for security purposes. <br> **Supported classes:** system_clock, treat_as_floating_point, duration_values |
```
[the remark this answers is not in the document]
mingweishih: Updated. I marked those headers that are are affected by C++20 standard as "Supported up to C++17" based. For those introduced by C++11 and unchanged since then, I kept them as "Supported as part of C++11".

docs/LibcxxSupport.md
```diff
@@ -109,7 +115,7 @@ Header | Supported | Comments |
 :---:|:---:|:---|
 bit | No | Header is not provided, C++20 is not yet supported. |
 cfenv | Yes | Supported as part of C++11. |
-cmath | Partial | **Unsupported functions:** acosh(), asinh(), fmal(), lgamma(), lgammaf(), sinh(), sinhl(), tgamma(). |
+cmath | Partial | **Supported but not fully tested functions:** fmal() and tgamma(). |
```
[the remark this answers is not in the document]
mingweishih: Updated.

tests/libcxx/CMakeLists.txt
mingweishih: I re-organized that list based on this issue.

mingweishih: I reorganized that list based on this issue.

tests/libcxx/enc/CMakeLists.txt
radhikaj: @mingweishih Can you please help me understand this statement
> However, the list of C++17-dependent (291 cases) does less than C++14-dependent one.
Are there more tests for C++14 than there are for C++17?

radhikaj:
@mingweishih Thanks for looking through all these tests and fixing documentation for our C++ support.

tests/libcxx/enc/CMakeLists.txt
mingweishih: > @mingweishih Can you please help me understand this statement.Are there more tests for C++14 than there are for C++17?
> 
> > However, the list of C++17-dependent (291 cases) does less than C++14-dependent one.

Ah, my mistake. The statement should be "However, the list of C++17-dependent (291 cases) does more than C++14-dependent one." instead.

achamayou:
@mingweishi thank you for doing that, you’re going to make some users really happy over here!

tests/libcxx/enc/CMakeLists.txt
mingweishih: Yes, some test cases (updated number is 58) I mentioned are failing with c++17, but passing with c++14. The requirement of each test is well-documented in the comments of each test.

Those failing tests mostly depend on the API changes introduced in c++17.
For example, binary_function was removed in c++17:
https://github.com/openenclave/openenclave/blob/ee6782c3253fe941084185cb215ff48a122a6104/3rdparty/libcxx/libcxx/test/std/depr/depr.function.objects/depr.base/binary_function.pass.cpp#L12

Similarly, unary_function was also removed in c++17:
https://github.com/openenclave/openenclave/blob/ee6782c3253fe941084185cb215ff48a122a6104/3rdparty/libcxx/libcxx/test/std/depr/depr.function.objects/depr.adaptors/depr.member.pointer.adaptors/mem_fun1_t.pass.cpp#L11

It turns out that there are only two such cases between c++11 and c++14.
https://github.com/openenclave/openenclave/blob/ee6782c3253fe941084185cb215ff48a122a6104/3rdparty/libcxx/libcxx/test/std/language.support/support.dynamic/new.delete/new.delete.array/sized_delete_array11.pass.cpp#L15
and
https://github.com/openenclave/openenclave/blob/ee6782c3253fe941084185cb215ff48a122a6104/3rdparty/libcxx/libcxx/test/std/language.support/support.dynamic/new.delete/new.delete.single/sized_delete11.pass.cpp#L15

They are currently classified as unsupported.
https://github.com/openenclave/openenclave/blob/ee6782c3253fe941084185cb215ff48a122a6104/tests/libcxx/tests.unsupported#L298
and
https://github.com/openenclave/openenclave/blob/ee6782c3253fe941084185cb215ff48a122a6104/tests/libcxx/tests.unsupported#L290

radhikaj:
Thanks to @CodeMonkeyLeet @anakrish @achamayou @jhand2 for their in-depth reviews!

PR #2740, opened 2020-03-25
Fix inconsistent code generation in oeedger8r

### LINKED ISSUE #2661: edger8r generating unsafe result handling code
**Description**
Hello team,
I was reviewing the samples, and I noticed that the edger8r generated code was writing the result to a potentially unsafe memory location.  I don't think this is useful from an attack standpoint, or even exploitable for that matter, but I don't know enough about edger8r to make a broad, absolute statement. Anyway, this should be fixed.

For example, in `helloworld_t.c`:
```
void ecall_enclave_helloworld(
...
    uint8_t* output_buffer,
    size_t output_buffer_size,
    size_t* output_bytes_written)
{
...
    enclave_helloworld_args_t* pargs_out = (enclave_helloworld_args_t*)output_buffer;
...
    OE_ADD_SIZE(output_buffer_offset, sizeof(*pargs_out));

    /* Make sure input and output buffers lie within the enclave. */
...
    if (!output_buffer || !oe_is_within_enclave(output_buffer, output_buffer_size))
        goto done; <<<<<<<<<<<<<<<<<   case: executes due to bad output_buffer

...
done:
    if (pargs_out && output_buffer_size >= sizeof(*pargs_out))
        pargs_out->_result = _result; <<<<<<<<<<<<<<<<<<<<BAD: writes to output_buffer
}
```

mingweishih:
This PR fixes the inconsistency of the oeedger8r-generated code.
More specifically, removing redundant checks on the buffers (as `oe_is_within_enclave` already checks them) and adding the check of `oe_is_within_enclave` before writing the result upon `done`.

Fixes #2661 

### DIFF
--- a/tools/oeedger8r/src/Sources.ml
+++ b/tools/oeedger8r/src/Sources.ml
@@ -605,12 +605,11 @@ let get_ecall_function get_deepcopy (tf : trusted_func) =
     "";
     (* Buffer validation *)
     "    /* Make sure input and output buffers lie within the enclave. */";
-    "    if (!input_buffer || !oe_is_within_enclave(input_buffer, \
-     input_buffer_size))";
+    "    /* oe_is_within_enclave explicitly checks if buffers are null or not. */";
+    "    if (!oe_is_within_enclave(input_buffer, input_buffer_size))";
     "        goto done;";
     "";
-    "    if (!output_buffer || !oe_is_within_enclave(output_buffer, \
-     output_buffer_size))";
+    "    if (!oe_is_within_enclave(output_buffer, output_buffer_size))";
     "        goto done;";
     "";
     (* Prepare in and in-out parameters *)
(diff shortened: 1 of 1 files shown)


mingweishih:
> From a functionality perspective I think this is fine but on principal, I don't like having more than one cleanup label. goto's are constructs to have a healthy fear of in C (like malloc or string functions) and anything other than `goto done;` makes me uneasy.
> 
> In general goto as a general-purpose control flow mechanism can make code hard to read and hard to maintain. In this case it isn't really a big deal, I just don't want usage of goto to get away from us.
> 
> Maybe others feel differently, just wanted to give my general feel. Maybe I'm just a Dijkstra "goto considered harmful" purist 😄

Ah, I found that the additional goto is redundant. Just updated the PR.

tools/oeedger8r/src/Sources.ml
```diff
@@ -605,12 +605,12 @@ let get_ecall_function get_deepcopy (tf : trusted_func) =
     "";
     (* Buffer validation *)
     "    /* Make sure input and output buffers lie within the enclave. */";
-    "    if (!input_buffer || !oe_is_within_enclave(input_buffer, \
-     input_buffer_size))";
-    "        goto done;";
-    "";
     "    if (!output_buffer || !oe_is_within_enclave(output_buffer, \
      output_buffer_size))";
+    "        return;";
```
anakrish: While this is a perfectly good approach to the solution, can you see if we can preserve the `goto done` approach and yet be safe. So far we have consistently tried to have a single point of return for most of our code. There may be some exceptions and glaring omissions. But that has been the intent so far.

Also keep in mind that the macros used by edger8r (`OE_ADD_SIZE`, `OE_SET_IN_POINTER` etc) will try to `goto done` in case of error and might circumvent an early return.

mingweishih: Check you check again?

PR #2743, opened 2020-03-26
Update install_lvi_mitigation_bindir script

### LINKED ISSUE #2669: The `install_lvi_mitigation_bindir` should install the wrappers for clang, clang-8, and clang-9
#2651 supports using Make to build samples with clang, clang-8, or clang-9 (previously, only clang-7 available). Therefore, the `install_lvi_mitigation_bindir` script, which installs the compiler wrappers for supporting Load Value Injection(SGX) mitigation, should also add those clang wrappers to support this modification properly.

mingweishih:
The PR updates the script as follows.
- Support clang, clang-8, and clang-9 such that the users can now build the helloworld sample with LVI mitigation using multiple versions of clang.
- Handle ~ when specifying the installation path.

Also, fixing a bug in the helloworld sample that allows users to specify compiler with `CC=... make ...` when enabling LVI mitigation.

Fixes #2669 


mingweishih:
> This looks good to me. Is it possible to share the code between the wrappers?

One way I can think of is having all the wrappers call into a script. In that case, all the wrappers will have one line of code. Does that sound better?

mingweishih:
bors try

mingweishih:
> Can we consolidate all of the versions into a single parameterized script instead of of duplicating code between separate bash scripts? You've pretty much got the structure for it already.
> 
> If you want an example of a parameterized bash script, see scripts/format-code.

My understanding is for the use of parameterized script is, for example,
`compiler-wrapper --target clang-8 {ARGS to the compiler}`
However, the standard cmake cannot feed the parameter to the script. The expected behavior is
`clang-8 {ARGS to the compiler}`
To fill the gap, I think we still need to keep separated scripts for each of compilers.

Please correct me if I'm wrong.

scripts/lvi-mitigation/clang-7-lvi
```diff
@@ -2,33 +2,14 @@
 # Copyright (c) Open Enclave SDK contributors.
 # Licensed under the MIT License.
```
BRMcLaren, line 3: Part of these compiler changes will change the underlying expected environment, as these bits are linked in with the ansible on rebuilds. Can you ping me when you are ready to merge and we can run an e2e to ensure nothing has changed unexpectedly?

mingweishih: Sure. Also, I didn't include changes to ansible this PR, which may be required and  I'm thinking of opening another follow-up one after this one is merged. What do you think?

Note that this PR does not impact the build process of OE (and the tests), but only allow users to have more choices of compilers when building their enclave apps.

kuanggang-git:

mingweishih:
> Approving since it's functionally fine, but consider following up on this thread later:
> 
> > My understanding is for the use of parameterized script is, for example,
> > compiler-wrapper --target clang-8 {ARGS to the compiler}
> > However, the standard cmake cannot feed the parameter to the script. Instead, the expected behavior is invoking a executable with the exact name as a compiler. For example,
> > clang-8 {ARGS to the compiler}
> > To fill the gap, I think we still need to keep separated scripts for each of compilers.
> 
> I'm not quite sure what the difference you're expressing here is? My original comment is just based on the observation that all of the LVI scripts set up to be renamed as clang are structurally identical, and that suggests they can be consolidated into a single script, not that they needed to be invoked via a parameterized command line. Consider for example, how we generate the pkgconfig files directly through CMake parameters, rather than hardcoding individual source files.

I misunderstood your comments. Please check the updated PR. Thanks!

mingweishih:
bors try

Issue #2742, opened 2020-03-26
Add OESDK mailing lists to documentation


dthaler:
There are two relevant email lists today, neither of which is mentioned in any document in the repo:
oesdk@lists.confidentialcomputing.io
oesdkcgc@lists.confidentialcomputing.io 

I believe the oesdkcgc list should at least be referenced at Committers.md and
the oesdk list would be good to mention right in the main README file.

jhand2:
Currently only members of the GCG can send mail to the oesdkcgc mailing list. Do you think that is the correct behavior?

For the oesdk mailing list I agree we should publish it in the README

dthaler:
> Currently only members of the GCG can send mail to the oesdkcgc mailing list. Do you think that is the correct behavior?

My opinion: no.  I think anyone should be able to send email to it, e.g., to report a code of conduct issue or whatever else.

dthaler:
Between PR #2814 and PR #2744 this is now done.

PR #2744, opened 2020-03-26
Add CGC mailing list to documentation

dthaler:
Addresses the half of issue #2742 about the CGC.
The general oesdk mailing list is now covered in PR #2814

### DIFF
--- a/docs/CodeOfConduct.md
+++ b/docs/CodeOfConduct.md
@@ -55,8 +55,10 @@ a project may be further defined and clarified by project maintainers.
 ## Enforcement
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
-reported by contacting any member of the [Community Governance Committee](
-Committers.md#Committee_Members). All complaints will be reviewed and investigated
+reported by contacting the [Community Governance Committee](Committers.md)
+at oesdkcgc@lists.confidentialcomputing.io, or by contacting any individual
+[member of the Community Governance Committee](Committers.md#Committee_Members).
+All complaints will be reviewed and investigated
 and will result in a response that is deemed necessary and appropriate to the
 circumstances. The project team is obligated to maintain confidentiality with
 regard to the reporter of an incident.
--- a/docs/Committers.md
+++ b/docs/Committers.md
@@ -24,6 +24,9 @@ vote of all members of the Community Governance Committee.
 The running minutes for the Community Governance Community meetings are located
 at [aka.ms/openenclave/cgc](https://aka.ms/openenclave/cgc).
 
+The Committee can be reached by sending email to
+oesdkcgc@lists.confidentialcomputing.io.
+
 Committee Members
 -----------------
 


docs/Committers.md
```diff
@@ -24,6 +24,9 @@ vote of all members of the Community Governance Committee.
 The running minutes for the Community Governance Community meetings are located
 at [aka.ms/openenclave/cgc](https://aka.ms/openenclave/cgc).
 
+The Committee can be reached by sending email to
+oesdkcgc@lists.confidentialcomputing.io.
```
CodeMonkeyLeet: Folks can actually send a private mail to the administrators of the oesdk mailiing list, perhaps that's a better way to do this for anything that is not just sent to oesdk as a public mail?

docs/CodeOfConduct.md
```diff
@@ -55,8 +55,10 @@ a project may be further defined and clarified by project maintainers.
 ## Enforcement
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
-reported by contacting any member of the [Community Governance Committee](
-Committers.md#Committee_Members). All complaints will be reviewed and investigated
+reported by contacting the [Community Governance Committee](Committers.md)
+at oesdkcgc@lists.confidentialcomputing.io, or by contacting any individual
```
CodeMonkeyLeet: This doesn't actually work (similarly for the recommendation in docs/Committers.md), the list automatically rejects non-member mails as a tradeoff for not moderating mails to the list (which the oesdk list does) We should consider setting up a separate oesdkcoc mailing list or alias?

docs/Committers.md
AevaOnline: Are the administrators of that list enumerated somewhere? If that isn't a publicly-visible artefact of the list itself, are you concerned it may get out of date if the list membership is updated?

docs/CodeOfConduct.md
jhand2: Is there a reason people should not be able to directly send mail to the oesdkgcg list? Especially if oesdkcoc and oesdkgcg will go to the same people.

docs/Committers.md
AevaOnline: This raises a different question as I think about it -

is the intent of the oesdkcgc@ list to be a private discussion space for CGC members, or a public discussion space _about_ community governance?

If the former, it should remain private and moderate any non-member posts. If the latter, it should be publicly visible, and allow posts and subscriptions from any member of the oesdk@ community.

jhand2: It is intended to be the former.

But should others be able to send messages to the GCG that they wish to be private (for example CoC violations/concerns)? Or do you think it should just be a separate list as @CodeMonkeyLeet  suggests?

docs/CodeOfConduct.md
dthaler: My opinion: I think anyone should be able to send email to oesdkcgc.   Since all the CGC members currently have a public email address listed on the https://github.com/openenclave/openenclave/blob/master/docs/Committers.md page, I don't think there's significant benefit in preventing it or moderating it, given that anyone could just send email to the set of individual email addresses directly.

kuanggang-git:

dthaler:
Since the CGC has not come to a conclusion yet, I will separate out the oesdk list part from the oesdkcgc part, so the former can hopefully be merged, and will keep this PR as covering the oesdkcgc part since that's what the comments in this PR are all about

jhand2:
I have added this to the CGC project for discussion: https://github.com/openenclave/openenclave/projects/14

For this to be merged the mailing list behavior would have to be changed to allow non-members to send email to the cgc mailing list.

dthaler:
In the CGC meeting, I believe the agreement was to modify the oesdkcgc mailing list permissions and then merge this PR.

CodeMonkeyLeet:
@dthaler I've updated the permissions on the oesdkcgc mailing list, so you're good to go

dthaler:
@CodeMonkeyLeet thanks!   

Thanks Aeva for approving, can I get a 2nd approval from someone?

PR #2746, opened 2020-03-26
add spinlock/multithread/errno check in same/different enclave

yanxue-intel:

### DIFF
--- a/tests/thread/README.md
+++ b/tests/thread/README.md
@@ -16,4 +16,8 @@ Test various OE synchronization primitives:
   **oe_rwlock_t**
   1. *TestReadersWriterLock* : Tests readers-writer lock invariants by launching multiple reader and writer threads racing against each other. Asserts that multiple/all readers can be simultaneously active, only one writer is active,  readers and writers are never simultaneously active.
 
+  **oe_spinlock_t**
+  1. *TestErrnoMultiThreadsSameenclave* : Tests errno can be set correctly in multi-threads for same enclaves.
+  2. *TestErrnoMultiThreadsDiffenclave* : Tests errno can be set correctly in multi-threads for different enclaves.
+
 This directory builds test enclaves for both OE threads and pthreads.
--- a/tests/thread/enc/errno_tests.cpp
+++ b/tests/thread/enc/errno_tests.cpp
@@ -1,0 +1,30 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#ifdef _PTHREAD_ENC_
+#include "thread.h"
+#endif
+
+#include <openenclave/enclave.h>
+#include <openenclave/internal/print.h>
+#include <openenclave/internal/thread.h>
+#include <openenclave/internal/types.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <algorithm>
+#include "errno.h"
+#include "thread_t.h"
+
+void* enc_malloc(size_t size, int* err)
+{
+    void* ret = malloc(size);
+    *err = errno;
+    return ret;
+}
+
+int64_t enc_strtol(const char* nptr, int base, int* err)
+{
+    long ret = strtol(nptr, NULL, base);
+    *err = errno;
+    return ret;
+}
--- a/tests/thread/host/host.cpp
+++ b/tests/thread/host/host.cpp
@@ -210,6 +210,10 @@ void test_thread_locking_patterns(oe_enclave_t* enclave)
 }
 
 void test_readers_writer_lock(oe_enclave_t* enclave);
+void test_errno_multi_threads_sameenclave(oe_enclave_t* enclave);
+void test_errno_multi_threads_diffenclave(
+    oe_enclave_t* enclave1,
+    oe_enclave_t* enclave2);
 
 // test_tcs_exhaustion
 static std::atomic<size_t> g_tcs_out_thread_count(0);
@@ -290,6 +294,7 @@ int main(int argc, const char* argv[])
 {
     oe_result_t result;
     oe_enclave_t* enclave = NULL;
+    oe_enclave_t* enclave2 = NULL;
 
     if (argc != 2)
     {
@@ -320,7 +325,19 @@ int main(int argc, const char* argv[])
 
     test_tcs_exhaustion(enclave);
 
-    if ((result = oe_terminate_enclave(enclave)) != OE_OK)
+    test_errno_multi_threads_sameenclave(enclave);
+
+    result = oe_create_thread_enclave(
+        argv[1], OE_ENCLAVE_TYPE_SGX, flags, NULL, 0, &enclave2);
+    if (result != OE_OK)
+    {
+        oe_put_err("oe_create_thread_enclave(): result=%u", result);
+    }
+
+    test_errno_multi_threads_diffenclave(enclave, enclave2);
+
+    if ((result = oe_terminate_enclave(enclave)) != OE_OK ||
+        (result = oe_terminate_enclave(enclave2)) != OE_OK)
     {
         oe_put_err("oe_terminate_enclave(): result=%u", result);
     }
--- a/tests/thread/thread.edl
+++ b/tests/thread/thread.edl
@@ -40,6 +40,15 @@ enclave {
             [out] size_t* max_readers,
             [out] size_t* max_writers,
             [out] bool* readers_and_writers);
+
+        public void* enc_malloc(
+            size_t size,
+            [out] int *err);
+
+        public int64_t enc_strtol(
+            [in, string] const char *nptr,
+            int base,
+            [out] int *err);
     };
 
     untrusted {
--- a/tests/thread/enc/CMakeLists.txt
+++ b/tests/thread/enc/CMakeLists.txt
@@ -13,6 +13,7 @@ add_enclave(TARGET pthread_enc UUID abe989c7-b167-4cc8-9653-c18a0bdc66da CXX
     enc.cpp
     cond_tests.cpp
     rwlock_tests.cpp
+    errno_tests.cpp
     thread_t.c)
 
 add_enclave(TARGET oethread_enc UUID 35c689f8-f752-4896-9c83-ec16dc7bd10e CXX
@@ -20,6 +21,7 @@ add_enclave(TARGET oethread_enc UUID 35c689f8-f752-4896-9c83-ec16dc7bd10e CXX
     enc.cpp
     cond_tests.cpp
     rwlock_tests.cpp
+    errno_tests.cpp
     thread_t.c)
 
 enclave_compile_definitions(pthread_enc PRIVATE -D_PTHREAD_ENC_)
--- a/tests/thread/host/CMakeLists.txt
+++ b/tests/thread/host/CMakeLists.txt
@@ -8,7 +8,7 @@ add_custom_command(
     DEPENDS ${EDL_FILE} edger8r
     COMMAND edger8r --untrusted ${EDL_FILE} --search-path ${CMAKE_CURRENT_SOURCE_DIR})
 
-add_executable(thread_host host.cpp rwlocks_test_host.cpp thread_u.c)
+add_executable(thread_host host.cpp rwlocks_test_host.cpp errno_test_host.cpp thread_u.c)
 
 target_include_directories(thread_host PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 
--- a/tests/thread/host/errno_test_host.cpp
+++ b/tests/thread/host/errno_test_host.cpp
@@ -1,0 +1,129 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <errno.h>
+#include <openenclave/host.h>
+#include <openenclave/internal/error.h>
+#include <openenclave/internal/tests.h>
+#include <signal.h>
+#include <stdio.h>
+#include <cassert>
+#include <chrono>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <thread>
+#include "thread_u.h"
+
+#define ERRNO_TEST_TIMES 10
+#define LARGE_ENOUGH_MEMORY_SIZE 1024 * 1024 * 1024
+#define NUM_ERRNO_TEST_THREADS 2
+
+static int g_error_num_from_thread1 = 0;
+static int g_error_num_from_thread2 = 0;
+
+void* errno_thread1(void* para)
+{
+    int i = 0;
+    int t_errno = 0;
+    void* buffer = NULL;
+    oe_enclave_t* enclave = (oe_enclave_t*)para;
+
+    for (i = 0; i < ERRNO_TEST_TIMES; i++)
+    {
+        enc_malloc(enclave, &buffer, LARGE_ENOUGH_MEMORY_SIZE, &t_errno);
+
+        if (ENOMEM != t_errno)
+        {
+            printf(
+                "[errno_thread1]Error check errno, expected %d, got %d.\n",
+                ENOMEM,
+                t_errno);
+            g_error_num_from_thread1 |= 0x8;
+        }
+    }
+
+    return NULL;
+}
+
+void* errno_thread2(void* para)
+{
+    int i = 0;
+    int64_t res = 0;
+    int t_errno = 0;
+    oe_enclave_t* enclave = (oe_enclave_t*)para;
+
+    for (i = 0; i < ERRNO_TEST_TIMES; i++)
+    {
+        char num_str[] = "123456789987654321123456789987654321";
+        enc_strtol(enclave, &res, num_str, 10, &t_errno);
+
+        if (ERANGE != t_errno)
+        {
+            printf("[errno_thread2] expected %d, errno %d.\n", ERANGE, t_errno);
+            g_error_num_from_thread2 |= 0x40;
+        }
+    }
+    return NULL;
+}
+
+void test_errno_multi_threads_sameenclave(oe_enclave_t* enclave)
+{
+    printf("test_errno_multi_threads_sameenclave Starting\n");
+
+    std::thread threads[NUM_ERRNO_TEST_THREADS];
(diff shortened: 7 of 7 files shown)


tests/thread/thread.edl
```diff
@@ -40,6 +40,12 @@ enclave {
             [out] size_t* max_readers,
             [out] size_t* max_writers,
             [out] bool* readers_and_writers);
+
+        public void enc_spin_lock( [user_check]void * lock);
+        public void enc_spin_unlock( [user_check]void * lock);
+	public void* enc_malloc(size_t size, [user_check]int *err);
+	public void enc_errno([user_check]int *err);
+	public long enc_strtol([user_check]char *nptr,[user_check]char **endptr, int base, [user_check]int *err);
```
jhand2: nit: Indent

tests/thread/host/host.cpp
```diff
@@ -286,10 +286,15 @@ size_t host_tcs_out_thread_count()
     return g_tcs_out_thread_count;
 }
 
+int test_errno_multi_threads_sameenclave(oe_enclave_t* enclave);
+int test_errno_multi_threads_diffenclave(
+    oe_enclave_t* enclave1,
+    oe_enclave_t* enclave2);
```
jhand2: Could you move these forward declarations to the top of the file? Along with `test_readers_writer_lock` which isn't part of this PR but I would prefer they all go together.

tests/thread/host/spinlock_test_host.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+#define ERRNO_TEST_TIMES 1
+#define LARGE_ENOUGH_MEMORY_SIZE 1024 * 1024 * 1024
+#define LONG_ENOUGH_TIME 40000
+
+int g_flag_withlock = 0;
+static uint32_t g_lock = 0;
+static int g_error_num = 0;
+
+int _wait_single(void* thread_id, int time_out)
```
jhand2: could you chance this name to "timeout_ms" to make it clear that the timeout is in milliseconds?

tests/thread/host/spinlock_test_host.cpp
```diff
@@ -0,0 +1,161 @@
...
+{
+    for (int i = 0; i < time_out; i++)
+    {
+        usleep(1000);
+        if (0 != pthread_kill(*(pthread_t*)thread_id, 0))
+        {
+            sleep(2);
```
jhand2: What is the reason for sleeping here?

tests/thread/host/spinlock_test_host.cpp
```diff
@@ -0,0 +1,158 @@
...
+            enc_spin_unlock(enclave, &g_lock);
+        }
+    }
+    return NULL;
+}
+
+void test_errno_multi_threads_sameenclave(oe_enclave_t* enclave)
+{
+    printf("test_errno_multi_threads_sameenclave Starting\n");
+    g_flag_withlock = 1;
```
radhikaj, line 102: Since errno is thread specific, is the lock and unlock required here?
I am thinking the host can just call enc_malloc on one thread and enc_strtol on another thread and the erronos should be preserved without acquiring a spinlock. Please correct me if I am wrong.

tests/thread/README.md
```diff
@@ -16,4 +16,8 @@ Test various OE synchronization primitives:
   **oe_rwlock_t**
   1. *TestReadersWriterLock* : Tests readers-writer lock invariants by launching multiple reader and writer threads racing against each other. Asserts that multiple/all readers can be simultaneously active, only one writer is active,  readers and writers are never simultaneously active.
 
+  **oe_spinlock_t**
+  1. *TestErrnoMultiThreadsSameenclave* : Tests spinlock protecting errno in multi-threads for same enclaves.
+  1. *TestErrnoMultiThreadsDiffenclave* : Tests spinlock protecting errno in multi-threads for different enclaves.
```
radhikaj, line 21:
```suggestion
  2. *TestErrnoMultiThreadsDiffenclave* : Tests spinlock protecting errno in multi-threads for different enclaves.
```

radhikaj, line 21: For TestErrnoMultiThreadsDiffenclave g_flag_withlock is set to 0. So, there is no spinlock protecting errno.

tests/thread/host/spinlock_test_host.cpp
```diff
@@ -0,0 +1,158 @@
...
+            threads[i] = std::thread(errno_thread2, enclave2);
+        }
+    }
+
+    for (size_t i = 0; i < NUM_SPINLOCK_TEST_THREADS; i++)
+    {
+        threads[i].join();
+    }
+
+    printf("test_errno_multi_threads_diffenclave Complete\n");
```
radhikaj, line 155: If my understanding is correct, you are testing if you can have different threads calling into different enclaves generating different errnos inside the two different enclaves. Is this test ensuring that the errnos inside different enclaves dont overwrite each other? If it is, since different enclaves have different thread structures, I am not sure I understand why this is needed.

tests/thread/host/spinlock_test_host.cpp
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        char num_str[] = "123456789987654321123456789987654321";
+        enc_strtol(enclave, &res, num_str, NULL, 10, &t_errno);
+
+        if (ERANGE != t_errno)
+        {
+            printf("[errno_thread2] expected %d, errno %d.\n", ERANGE, t_errno);
+            g_error_num |= 0x40;
+            if (g_flag_withlock)
```
radhikaj, line 85: I think this unlock is not needed since there is an unlock immediately after this if statement block.

tests/thread/host/spinlock_test_host.cpp
yanxue-intel: yes, you are right. These cases are testing errno is not overwritten by multithread in the same/different enclaves.  These Cases focus on errno testing, i will change file name from "spinlock_test_host.cpp" to "errno_test_host.cpp".

tests/thread/host/spinlock_test_host.cpp
anakrish: @radhikaj Github sometimes makes this hard to review. If you expand the code, this unlock is within a loop and is therefore needed so as to match the lock.

anakrish: Actually, I was the one that misread the code :D 
Yes, upon scrolling and rereading, this unlock is not needed.

tests/thread/host/spinlock_test_host.cpp
anakrish: The threads do `g_error_num |= 0x80` and `g_error_num | = 0x40`. So locking is needed.

tests/thread/host/spinlock_test_host.cpp
```diff
@@ -0,0 +1,158 @@
...
+        if (ENOMEM != t_errno)
+        {
+            printf(
+                "[errno_thread1]Error check errno, expected %d, got %d.\n",
+                ENOMEM,
+                t_errno);
+            g_error_num |= 0x8;
+            if (g_flag_withlock)
+            {
+                enc_spin_unlock(enclave, &g_lock);
```
anakrish, line 51: After this `enc_spin_unlock` executes, the unlock in line 57 would execute too leading to double unlock.

tests/thread/host/spinlock_test_host.cpp
```diff
@@ -0,0 +1,158 @@
...
+    int i = 0;
+    int t_errno = 0;
+    void* buffer = NULL;
+    oe_enclave_t* enclave = (oe_enclave_t*)para;
+
+    for (i = 0; i < ERRNO_TEST_TIMES; i++)
+    {
+        if (g_flag_withlock)
+        {
+            enc_spin_lock(enclave, &g_lock);
```
anakrish, line 37: We don't recommend that locks are beyond the duration of an ecall. The reason is that in Open Enclave SDK, the binding between host thread and the enclave thread lasts only for the duration of the ecall.
Another ecall from the same host thread may end up using a different enclave thread. 

As a result, we avoid holding enclave locks beyond the duration of the ecall.

tests/thread/thread.edl
```diff
@@ -40,6 +40,25 @@ enclave {
             [out] size_t* max_readers,
             [out] size_t* max_writers,
             [out] bool* readers_and_writers);
+
+        public void enc_spin_lock(
+            [user_check] void * lock);
```
anakrish, line 45: Instead of user check, you could use and `out` parameter which is a better practice that we want to encourage our users to adopt.

tests/thread/host/spinlock_test_host.cpp
radhikaj: Thanks @anakrish. So the lock is needed on the global variable but not the errno. The description suggests you need the lock to protect errno

tests/thread/host/spinlock_test_host.cpp
radhikaj: I would request the test be changed the following way: 
1. while calling into an enclave don’t acquire a spin lock. 
2. The errno is stored in an out arg in the ecall
3. On the host end after the ecall is complete, acquire a lock and manipulate the global var to store the errno you just received. This is not necessary and you can use two global vars as well.

tests/thread/thread.edl
yanxue-intel: ok, thanks

kuanggang-git:

tests/thread/host/spinlock_test_host.cpp
yanxue-intel: Thanks for your suggestion, i updated the code.

yanxue-intel:
> @yanxue-intel Did you have a particular scenario in mind for this test?
> As @radhikaj pointed out, errno is thread-specific. Would ensuring that `errno` in each thread has a unique address suffice?

yes, I have modify the test cases, removed the spinlock.
the test case checks errno can be set correctly in multi-threads for same/different enclave.

radhikaj:
> > @yanxue-intel Did you have a particular scenario in mind for this test?
> > As @radhikaj pointed out, errno is thread-specific. Would ensuring that `errno` in each thread has a unique address suffice?
> 
> yes, I have modify the test cases, removed the spinlock.
> the test case checks errno can be set correctly in multi-threads for same/different enclave.
Thanks @yanxue-intel for modifying the test cases. Much appreciated. Thanks for this contribution

tests/thread/enc/errno_tests.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <algorithm>
+#include "errno.h"
+#include "thread_t.h"
+
+void* enc_malloc(size_t size, int* err)
+{
+    void* ret = malloc(size);
+    *err = errno;
```
anakrish, line 21: Note, standard C does not require that malloc set errno:
https://en.cppreference.com/w/c/memory/malloc

POSIX and Linux require that malloc set errno:
https://pubs.opengroup.org/onlinepubs/009695399/functions/malloc.html

tests/thread/enc/errno_tests.cpp
```diff
@@ -0,0 +1,30 @@
...
+    return ret;
+}
+
+long enc_strtol(const char* nptr, int base, int* err)
+{
+    long ret = strtol(nptr, NULL, base);
```
anakrish, line 27: Both C standard and POSIX require errno to be set for strtol.

anakrish:
@yanxue-intel Thanks for addressing all the comments!

anakrish:
bors try

anakrish:
@yanxue-intel Can you double check if you have pushed all the changes?

The number of arguments to `enc_malloc` in edl is 2 whereas at the actual callsite it seems to be `enc_malloc(enclave, &buffer, LARGE_ENOUGH_MEMORY_SIZE, &t_errno);`.

anakrish:
> @yanxue-intel Can you double check if you have pushed all the changes?
> 
> The number of arguments to `enc_malloc` in edl is 2 whereas at the actual callsite it seems to be `enc_malloc(enclave, &buffer, LARGE_ENOUGH_MEMORY_SIZE, &t_errno);`.

You can ignore my previous comment. Radhika pointed out that it was an ecall and hence we get two extra parameters. Time for me to take a break :)

tests/thread/enc/errno_tests.cpp
yanxue-intel: The following code is from "openenclave\enclave\core\malloc.c",  Does the malloc implementation in OE set errno?
void* oe_malloc(size_t size)
{
    void* p = MALLOC(size);

    if (!p && size)
    {
        errno = ENOMEM;

        if (_failure_callback)
            _failure_callback(__FILE__, __LINE__, __FUNCTION__, size);
    }

    return p;
}

tests/thread/enc/errno_tests.cpp
yanxue-intel: @anakrish why the case don't get errno from oe strtol on windows?
The follow is the log of the cases running on the windows platform.
[errno_thread1]Error check errno, expected 12, got 12.( this is malloc errno)
[errno_thread2] expected 34, got errno 0.( this is strtol errno)

yanxue-intel:
OE enclave is not set errno for strtol on windows.  The behavior is different from C standard.
 Is it the design of windows OE?

[https://en.cppreference.com/w/c/string/byte/strtol](url)
> If the converted value falls out of range of corresponding return type, a range error occurs (setting errno to ERANGE) and LONG_MAX, LONG_MIN, LLONG_MAX or LLONG_MIN is returned.

radhikaj:
> OE enclave is not set errno for strtol on windows. The behavior is different from C standard.
> Is it the design of windows OE?
> 
> [https://en.cppreference.com/w/c/string/byte/strtol](url)
> 
> > If the converted value falls out of range of corresponding return type, a range error occurs (setting errno to ERANGE) and LONG_MAX, LONG_MIN, LLONG_MAX or LLONG_MIN is returned.

Thanks @yanxue-intel 
@nonpolarity @anakrish @jhand2 would one of you be able to investigate why errno is not set on windows?

tests/thread/enc/errno_tests.cpp
anakrish: @yanxue-intel Can you use int64_t instead of long? long has different size on Windows and Linux. OE uses ELF format (Linux) on both Linux and Windows. edger8r will generate a warning when these non portable types are used.

tests/thread/thread.edl
```diff
@@ -40,6 +40,15 @@ enclave {
...
             [out] bool* readers_and_writers);
+
+        public void* enc_malloc(
+            size_t size,
+            [out] int *err);
+
+        public long enc_strtol(
+            [in, string] const char *nptr,
+            int base,
+            [out] int *err);
```
jhand2, line 51: @anakrish is `propagate_errno` not needed for ecalls?

jhand2: Oh I see, it's just passed manually. edl isn't aware if it.

tests/thread/enc/errno_tests.cpp
anakrish: From what I understand, there is an inconsistency:

`oe_malloc` sets errno: https://github.com/openenclave/openenclave/blob/6bed4b5e0927746ca2e9f0c0c8b154e5dd66a204/enclave/core/malloc.c#L93-L106

`oe_debug_malloc` (which is used in debug builds) does not seem to set errno:
https://github.com/openenclave/openenclave/blob/6bed4b5e0927746ca2e9f0c0c8b154e5dd66a204/enclave/core/debugmalloc.c#L310-L327

anakrish: Let me know if you are running into failures due to this. We can provide a fix.

tests/thread/enc/errno_tests.cpp
yanxue-intel: @anakrish I don't think the errno is affected by output type cast.
I replaced long with int64_t, but the errno was not set either.

jinghe-INTC:
bors try

jinghe-INTC:
bors try

jhand2:
@jinghe-INTC we were having problems with our CI system for the last few days. It should be fine now

bors try

radhikaj:
Opened issue #2858. Please disable the test on Windows and merge the PR. Thanks!

tests/thread/enc/errno_tests.cpp
anakrish: @yanxue-intel The latest version where `long` has been changed to `int64_t`, the test now passes in Windows. The outstanding failures are in Linux due to timeout.

anakrish:
bors try

anakrish:
In the last, run the outstanding failures are in Linux LVI builds.

anakrish:
The outstanding failure seems nothing to do with this test:
```bash
 === begin _test_verify_with_crl()

2020-04-14T00:10:23.000000Z [(H)ERROR] tid(0x7fc425a04740) | X509_verify_cert failed!

 error: (23) certificate revoked

 (oe_result_t=OE_VERIFY_REVOKED) [../host/Test failed: ../tests/crypto/crl_tests.c(146): _test_get_dates last.minutes == _time.minutes

```

tests/thread/enc/errno_tests.cpp
yanxue-intel: linux oe_thread failure maybe caused by test_tcs_exhaustion, please help to check details in log file, Thanks. [](url)https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/rest/organizations/jenkins/pipelines/Bors/branches/staging/runs/273/nodes/71/steps/846/log/?start=0

yanxue-intel:
> Opened issue #2858. Please disable the test on Windows and merge the PR. Thanks!

strtol sets errno on windows,  I couldn't get errno from enclave before is due to enc_strtol is not called on windows.
When I replaced long with int64_t, this trust function(enc_strtol) can be called.

radhikaj:
bors try

radhikaj:
Bors r+

jhand2:
The pthread test timed out in one stage which I don't think is related here. Re-running

bors r+

PR #2748, opened 2020-03-26
Switch to edl internal structures for fcntl.edl and socket.edl

jhand2:
In order to support publishing EDL files and more robust deep copy, system EDL files should not depend on internal header files and should only depend on public header files for integral types. Start moving structures to be defined in EDL.

This commit also changes the name of the struct oe_stat to oe_stat_t to avoid name conflicts with the function oe_stat.

Note that there is still more work to be done in refactoring the syscall edl files. I stopped here because the PR was already getting quite large and I wanted to make it easier to review.


include/openenclave/bits/edl/syscall_types.h
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#ifndef _OE_EDL_SYSCALL_TYPES_H
+#define _OE_EDL_SYSCALL_TYPES_H
+
+/* DISCLAIMER:
+ * This header is published with no guarantees of stability and is not part
```
anakrish: Nice disclaimer.

include/openenclave/internal/syscall/sys/stat.h
```diff
@@ -45,9 +46,24 @@ OE_EXTERNC_BEGIN
 #define OE_S_IRWGRP (OE_S_IRGRP | OE_S_IWGRP)
 #define OE_S_IRWOTH (OE_S_IROTH | OE_S_IWOTH)
 
-#define __OE_STAT oe_stat
-#include <openenclave/internal/syscall/sys/bits/stat.h>
-#undef __OE_STAT
+OE_STATIC_ASSERT((sizeof(struct oe_stat_t) % 8) == 0);
```
anakrish: Python libclang bindings can help with auto-generating these asserts:
https://eli.thegreenplace.net/2011/07/03/parsing-c-in-python-with-clang

common/edl/fcntl.edl
```diff
@@ -15,9 +15,40 @@
...
-    include "openenclave/internal/syscall/types.h"
+    include "openenclave/bits/edl/syscall_types.h"
+
+    struct oe_dirent
+    {
+        uint64_t d_ino;
+        oe_off_t d_off;
+        uint16_t d_reclen;
+        uint8_t d_type;
+        char d_name[256]; // OE_NAME_MAX + 1 == 256. TODO: Math in EDL?
```
mingweishih: I assume deep copy should be able to handle fixed size array?

jhand2: I think just normal shallow copy will handle it. Deep copy is just for following pointers.

mingweishih: Do you mean that this array does not need to be deep copied? I'm just thinking whether deep copy can infer the size of the array.

jhand2: I mean normal EDL can get the size of the array. This will still get copied even if the deep copy feature is not enabled

mingweishih: Got it! Thanks for the clarification.

common/edl/socket.edl
```diff
@@ -19,9 +19,28 @@
...
+    };
+
+    struct oe_addrinfo
+    {
+        int ai_flags;
+        int ai_family;
+        int ai_socktype;
+        int ai_protocol;
+        oe_socklen_t ai_addrlen;
+        struct oe_sockaddr* ai_addr;
```
dthaler: can ai_addr should be annotated as having ai_addrlen be the size of the buffer?
ai_addr can point to a buffer of length up to sizeof(sockaddr_storage)

jhand2: You're right, it can. I opened an issue to evaluate internal EDL structures for what annotation can be added to them: https://github.com/openenclave/openenclave/issues/2760

jhand2: I had to fix a Windows build issue so I went ahead and fixed this in this PR

common/edl/socket.edl
```diff
@@ -19,9 +19,28 @@
...
+    };
+
+    struct oe_addrinfo
+    {
+        int ai_flags;
+        int ai_family;
+        int ai_socktype;
+        int ai_protocol;
+        oe_socklen_t ai_addrlen;
+        [size=ai_addrelen] struct oe_sockaddr* ai_addr;
```
dthaler: typo: ai_addrlen -> ai_addrlen

common/edl/socket.edl
dthaler: Except you have a typo :)

common/edl/socket.edl
jhand2: huh. Wonder why it compiles locally. Thanks

jhand2: Oh, it's because deep copy wasn't turned on

PR #2750, opened 2020-03-26
Add Aeva to committee

andschwa:
Note that in the last CGC meeting the committee chose to not require all
committee members to be committers, so we need to adjust the written
rules (and the name of this document).


andschwa:
@openenclave/committee I'd like to revisit the choice to not give all committee members committer privileges. I wasn't there for the discussion, but believe I have a dissenting opinion. I think it is _necessary_ for the execution of the duties of Committee members that they have committer privileges (and it's also the way the Open Enclave GitHub Teams are setup). I will add this to the agenda for the next meeting, no need to discuss here, just a heads up since I mentioned it in this PR already.

PR #2752, opened 2020-03-26
Modify OE_RAISE_MSG() to have different versions for different compilers

shnwc:
- The bug: current implementation of OE_RAISE_MSG() fails on Linux with clang / gcc when NULL is put at the end of the parameter list (as commented in the code) . (null) is printed in log for the oe_result_t entry. As described in the comment in the existing header file, the NULL parameter is needed to address some limitations in MSVC compilers.

- The fix: define MSVC-specific and non-MSVC versions of OE_RAISE_MSG() that handle cases with and without NULL, to work properly with both MSVC and non-MSVC compilers while avoiding the need to modify existing code, regardless whether a trailing NULL parameter is added or not.

### DIFF
--- a/include/openenclave/internal/raise.h
+++ b/include/openenclave/internal/raise.h
@@ -93,12 +93,13 @@ OE_EXTERNC_BEGIN
 // variadic paramter is empty in OE_RAISE_MSG(RESULT, fmt, ...)
 // eg : OE_RAISE_MSG(OE_FAILURE, "your message", NULL);
 
-/* Note: on Linux, the above example fails, as the NULL pointer is fed as the
- * value for oe_result_t in log, showing invalid (null) output.
+/* But with gcc or clang that support the ##__VA_ARGS__ extension, the above
+ * approach fails, as the NULL pointer is fed as the value for oe_result_t
+ * position in log line, resulting in (null) output.
  *
- * A solution is to define OS specfic versions of OE_RAISE_MSG() that
- * handle cases with and without NULL, to work properly on all OSes while
- * avoiding the need to modify existing code.
+ * A solution is to define two versions of OE_RAISE_MSG() that
+ * handle calls with and without NULL, to work properly on both compiler
+ * types while avoiding the need to modify existing code.
  */
 
 #if defined(_MSC_VER)
@@ -126,7 +127,7 @@ OE_EXTERNC_BEGIN
     } while (0)
 
 #else
-// For Linux etc.: remove NULL if present
+// For non-MSVC: remove NULL if present
 #define OE_RAISE_MSG(RESULT, fmt, ...)                               \
     do                                                               \
     {                                                                \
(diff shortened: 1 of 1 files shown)


include/openenclave/internal/raise.h
```diff
@@ -85,20 +93,64 @@ OE_EXTERNC_BEGIN
...
-        goto done;                       \
+/* Note: on Linux, the above example fails, as the NULL pointer is fed as the
+ * value for oe_result_t in log, showing invalid (null) output.
+ *
+ * A solution is to define OS specfic versions of OE_RAISE_MSG() that
+ * handle cases with and without NULL, to work properly on all OSes while
+ * avoiding the need to modify existing code.
+ */
+
+#if defined(_MSC_VER)
```
dthaler: Line 104 doesn't quite match what line 99-101 says.  _MSC_VER is a compiler-specific, not OS-specific, preprocessor symbol.  Either update 99-101 to talk about compiler variations (if it's a compiler issue), or (if it's really an OS issue) 104-105 should instead test the platform like if defined(_WIN64).

include/openenclave/internal/raise.h
```diff
@@ -85,20 +93,64 @@ OE_EXTERNC_BEGIN
...
-            OE_TRACE_ERROR(              \
-                fmt " (oe_result_t=%s)", \
-                ##__VA_ARGS__,           \
-                oe_result_str(result));  \
-        }                                \
-        goto done;                       \
+/* Note: on Linux, the above example fails, as the NULL pointer is fed as the
+ * value for oe_result_t in log, showing invalid (null) output.
+ *
+ * A solution is to define OS specfic versions of OE_RAISE_MSG() that
```
CodeMonkeyLeet: Is it a simpler solution to simply reverse the order of the strings to make it consistent with `OE_RAISE()`?

```c
#define OE_RAISE_MSG(RESULT, fmt, ...)                                     \
    do                                                                     \
    {                                                                      \
        result = (RESULT);                                                 \
        if (result != OE_OK)                                               \
        {                                                                  \                                                          \
            OE_TRACE_ERROR(                                                \
                ":%s | " fmt,                                              \
                oe_result_str(result),                                     \
                ##__VA_ARGS__);                                            \
        }                                                                  \
        goto done;                                                         \
    } while (0)
```

shnwc: I tested with reversing the order of the strings as you mentioned, and the macro still fails, though in a different way.

If you look at the definition of OE_TRACE_ERROR(), you would see that the NULL ##__VA_ARGS__ would be fed to the first "%s" of its format (fmt " [%s:%s:%d]\n"), since "fmt" does not need any input. This also result in __FILE__, __FUNCTION__, and __LINE__ to be misaligned in the log line.

include/openenclave/internal/raise.h
shnwc: Good point, this is compiler issue. Will fix.

dthaler: Great, make sure to update line 129 too then.

anakrish:
bors try

include/openenclave/internal/raise.h
shnwc: That has been updated too.

PR #2753, opened 2020-03-26
Ensure pem-formated cert chain buffer has trailing null character

shnwc:
- The bug: oe_cert_chain_read_pem() fails when the input pem-formated cert chain buffer does not contain a trailing null character, and in verification of SGX ECDSA quote, the input buffer does not have this null character.
  - This bug is present in both enclave-side and host-side implementations
  - On enclave side, mbedtls_x509_crt_parse() fails with rc = 0xffffde80
  - On host side, _read_cert_chain() fails, resulting in error OE_FAILURE
- The fix: when the input buffer does not have a trailing null character, oe_cert_chain_read_pem() creates a tmp buffer to add it.

### DIFF
--- a/enclave/crypto/cert.c
+++ b/enclave/crypto/cert.c
@@ -9,6 +9,7 @@
 #include <mbedtls/platform.h>
 #include <mbedtls/x509_crt.h>
 
+#include <openenclave/corelibc/stdlib.h>
 #include <openenclave/enclave.h>
 #include <openenclave/internal/atomic.h>
 #include <openenclave/internal/cert.h>
@@ -17,6 +18,7 @@
 #include <openenclave/internal/print.h>
 #include <openenclave/internal/raise.h>
 #include <openenclave/internal/safecrt.h>
+#include <openenclave/internal/trace.h>
 #include <openenclave/internal/utils.h>
 #include <string.h>
 #include "crl.h"
@@ -738,8 +740,24 @@ oe_result_t oe_cert_chain_read_pem(
         OE_RAISE(OE_INVALID_PARAMETER);
 
     /* Must have pem_size-1 non-zero characters followed by zero-terminator */
-    if (strnlen((const char*)pem_data, pem_size) != pem_size - 1)
-        OE_RAISE(OE_INVALID_PARAMETER);
+    // Note: the above comment is incorrect. There is no trailing zero,
+    // but mbedtls_x509_crt_parse() requires a trailing zero
+    OE_TRACE_INFO(
+        "strlen()=%d vs pem_size=%d",
+        strnlen((const char*)pem_data, pem_size),
+        pem_size);
+
+    uint8_t* tmp_pem_data = (uint8_t*)pem_data;
+    size_t tmp_pem_size = pem_size;
+    if (strnlen((const char*)pem_data, pem_size) == pem_size)
+    {
+        tmp_pem_size = pem_size + 1;
+        if (!(tmp_pem_data = oe_calloc(1, tmp_pem_size)))
+            OE_RAISE(OE_OUT_OF_MEMORY);
+
+        oe_memcpy_s(tmp_pem_data, tmp_pem_size, pem_data, pem_size);
+        tmp_pem_data[pem_size] = '\0';
+    }
 
     /* Create the referent */
     if (!(referent = _referent_new()))
@@ -747,7 +765,11 @@ oe_result_t oe_cert_chain_read_pem(
 
     /* Read the PEM buffer into DER format */
     rc = mbedtls_x509_crt_parse(
-        referent->crt, (const uint8_t*)pem_data, pem_size);
+        referent->crt, (const uint8_t*)tmp_pem_data, tmp_pem_size);
+
+    if (tmp_pem_data != pem_data)
+        oe_free(tmp_pem_data);
+
     if (rc != 0)
         OE_RAISE_MSG(OE_CRYPTO_ERROR, "mbedtls_x509_crt_parse rc = 0x%x\n", rc);
 
--- a/enclave/crypto/cert.c
+++ b/enclave/crypto/cert.c
@@ -730,6 +730,8 @@ oe_result_t oe_cert_chain_read_pem(
     CertChain* impl = (CertChain*)chain;
     Referent* referent = NULL;
     int rc = 0;
+    uint8_t* tmp_pem_data = (uint8_t*)pem_data;
+    size_t tmp_pem_size = pem_size;
 
     /* Clear the implementation (making it invalid) */
     if (impl)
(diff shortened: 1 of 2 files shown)


enclave/crypto/cert.c
```diff
@@ -738,16 +740,36 @@ oe_result_t oe_cert_chain_read_pem(
         OE_RAISE(OE_INVALID_PARAMETER);
 
     /* Must have pem_size-1 non-zero characters followed by zero-terminator */
-    if (strnlen((const char*)pem_data, pem_size) != pem_size - 1)
-        OE_RAISE(OE_INVALID_PARAMETER);
+    // Note: the above comment is incorrect. There is no trailing zero,
```
dthaler: I'd recommend correcting line 740 so you don't have to say it's incorrect.

host/crypto/openssl/cert.c
```diff
@@ -636,15 +637,33 @@ oe_result_t oe_cert_chain_read_pem(
 
     /* Must have pem_size-1 non-zero characters followed by zero-terminator */
```
dthaler: Should this have the same comments as crypto/cert.c 742-744?

enclave/crypto/cert.c
```diff
@@ -738,16 +740,36 @@ oe_result_t oe_cert_chain_read_pem(
...
+    // but mbedtls_x509_crt_parse() requires a trailing zero
```
jhand2: You can probably remove the above command and just say "mbedtls_x509_crt_parse() requires a trailing zero"

enclave/crypto/cert.c
```diff
@@ -738,16 +740,36 @@ oe_result_t oe_cert_chain_read_pem(
...
+    OE_TRACE_INFO(
+        "strlen()=%d vs pem_size=%d",
+        strnlen((const char*)pem_data, pem_size),
+        pem_size);
```
jhand2: Is printing this necessary? Doesn't the below code fix the problem if there is one?

enclave/crypto/cert.c
```diff
@@ -738,16 +740,36 @@ oe_result_t oe_cert_chain_read_pem(
...
+
+    uint8_t* tmp_pem_data = (uint8_t*)pem_data;
+    size_t tmp_pem_size = pem_size;
+    if (strnlen((const char*)pem_data, pem_size) == pem_size)
+    {
+        tmp_pem_size = pem_size + 1;
+        if (!(tmp_pem_data = oe_calloc(1, tmp_pem_size)))
```
jhand2: Do we need to do a calloc of it is immediately followed by a memcpy? I think oe_malloc is fine

enclave/crypto/cert.c
```diff
@@ -738,16 +740,36 @@ oe_result_t oe_cert_chain_read_pem(
...
     if (!(referent = _referent_new()))
         OE_RAISE(OE_OUT_OF_MEMORY);
 
     /* Read the PEM buffer into DER format */
     rc = mbedtls_x509_crt_parse(
-        referent->crt, (const uint8_t*)pem_data, pem_size);
+        referent->crt, (const uint8_t*)tmp_pem_data, tmp_pem_size);
+
+    if (tmp_pem_data != pem_data)
+        oe_free(tmp_pem_data);
```
jhand2: if the check on line 764 fails then this leaks. Move this after the done label and do `if (tmp_pem && tmp_pem_data != pem_data)`

host/crypto/openssl/cert.c
```diff
@@ -636,15 +637,33 @@ oe_result_t oe_cert_chain_read_pem(
...
     /* Initialize OpenSSL (if not already initialized) */
     oe_initialize_openssl();
 
     /* Read the certificate chain into memory */
-    if (!(sk = _read_cert_chain((const char*)pem_data)))
+    if (!(sk = _read_cert_chain((const char*)tmp_pem_data)))
         OE_RAISE(OE_FAILURE);
 
+    if (tmp_pem_data != pem_data)
+        free(tmp_pem_data);
```
jhand2: same comment as the enclave-side change

enclave/crypto/cert.c
shnwc: This is the log line that revealed the bug. Will remove.

enclave/crypto/cert.c
shnwc: Agreed

enclave/crypto/cert.c
shnwc: Sure

host/crypto/openssl/cert.c
shnwc: Sure. Will make them similar

enclave/crypto/cert.c
shnwc: Sure.

enclave/crypto/cert.c
shnwc: Good catch!

shnwc:
> On the host side does this only happen on Linux? I'm worried we would run into the same problem in `host/crypto/bcrypt/...`.
> 
> Also, can you add a test which uses a non-null-terminated pem to make sure we fixed the problem in all cases?

shnwc:
> On the host side does this only happen on Linux? I'm worried we would run into the same problem in `host/crypto/bcrypt/...`.
> 
> Also, can you add a test which uses a non-null-terminated pem to make sure we fixed the problem in all cases?

@jhand2 on your comment quoted here, I only have a Ubuntu 18.04 platform, so won't be able to test on other platforms. 

But if you look at the flow leading to call to oe_cert_chain_read_pem(), it's the function oe_get_quote_cert_chain_internal() in common/sgx/quote.c that parses the received quote and feed the extracted cert chain data to oe_cert_chain_read_pem(), so it's unlikely the flow would behave differently on different platforms.

Or are you concerned that `host/crypto/bcrypt/...` might require non-null terminated buffer?

Not sure yet  how to respond to your comment "... can you add a test which uses a non-null-terminated pem ...". This bug was revealed in my tests when oe_cert_chain_read_pem() failed with a input pem_data[] buffer without a trailing zero.

jhand2:
> > On the host side does this only happen on Linux? I'm worried we would run into the same problem in `host/crypto/bcrypt/...`.
> > Also, can you add a test which uses a non-null-terminated pem to make sure we fixed the problem in all cases?
> 
> @jhand2 on your comment quoted here, I only have a Ubuntu 18.04 platform, so won't be able to test on other platforms.
> 

That's fine. As long as we have a unit test for it that will catch it on Windows.

> Not sure yet how to respond to your comment "... can you add a test which uses a non-null-terminated pem ...". This bug was revealed in my tests when oe_cert_chain_read_pem() failed with a input pem_data[] buffer without a trailing zero.

I mean could you add a test case to `tests/crypto`. This would allow us to see if it fails on any other platform are would protect against regressions as we run these tests on every PR.

jhand2:
> I mean could you add a test case to tests/crypto. This would allow us to see if it fails on any other platform are would protect against regressions as we run these tests on every PR.

@shnwc if not don't worry about it. I can just open an issue to add a test for this bug

enclave/crypto/cert.c
```diff
@@ -737,17 +741,27 @@ oe_result_t oe_cert_chain_read_pem(
     if (!pem_data || !pem_size || !chain)
         OE_RAISE(OE_INVALID_PARAMETER);
 
-    /* Must have pem_size-1 non-zero characters followed by zero-terminator */
-    if (strnlen((const char*)pem_data, pem_size) != pem_size - 1)
-        OE_RAISE(OE_INVALID_PARAMETER);
+    // mbedtls_x509_crt_parse() requires a trailing zero in its input buffer.
+    // If the input pem_data buffer does not have a trailing zero,
+    // we allocate a tmp buffer to add it.
```
yufjiang: If I understand correctly, this PR tries to handle those no-trailing-null pem-formated cert chain gracefully (by adding one), instead of just throwing OE_INVALID_PARAMETER? 

As far as I know, correct me if am wrong, the pem-formated cert chain returned by Intel DCAP client and 1.2+ version Azure DCAP client does have the trailing null character. I am curious in which case you can trigger this bug?

shnwc: Yes your understanding is accurate.

The test platform that I use is ICL notebook running Ubuntu 18.04 (not an Azure VM). It is configured with libsgx-dcap-ql v1.5, the DCAP reference QPL (DCAP client), and the reference PCK cert caching service sgx-dcap-pccs v1.5

In this setup, the pem-formatted cert chain has no trailing null, and the existing implementation fails (both the host and enclave sides).

This PR has no effect if the buffer does have a trailing null.

yufjiang: Thank you very much for clarifying. It is great that we have the fault tolerance capability by this PR. If I understand correctly, based on the contract between qpl/ql and OE, it is qpl's responsibility to properly add trailing null character. Is that correct?

PR #2755, opened 2020-03-27
Remove loadpe.c from the SGX loader

jhand2:
In the SGX loader there is a file `loadpe.c` for loading Portable Executable format (PE) binaries. OE only supports ELF enclaves. While the loader can load PE enclaves, they won't actually work properly with the enclave runtime.

Remove this file. It obfuscates what the loader is actually doing for people reading the code.


jhand2:
cc @CodeMonkeyLeet to make sure this doesn't mess with any of the loader design work you're planning

PR #2757, opened 2020-03-27
Add flag to support optional LVI mitigation tests

mingweishih:
This PR adds a new cmake option `LVI_MITIGATION_SKIP_TESTS` for explicitly disabling tests with LVI mitigation.
Also, adding the `ADD_LVI_MITIGATION` option to internal `add_enclave` function to build the enclave even when `LVI_MITIGATION_SKIP_TEST=ON`, which allows for testing the linking against LVI mitigated libraries.
    
Other changes:
- Update Jenkinesfile to use set `LVI_MITIGATION_SKIP_TESTS=ON` as default.
- Revert the changes on nightly_build.jenkinesfile to reflect the changes of the PR.
- Modify tests to use `add_enclave` correctly.
- Rename internal cmake function `add_enclave_executable` to `add_lvi_enclave_executable`, which now only adds a enclave with LVI mitigation as used by `add_enclave`.


mingweishih:
bors try

cmake/add_enclave.cmake
```diff
@@ -121,15 +122,23 @@ endfunction()
...
     "${options}"
     "${oneValueArgs}"
     "${multiValueArgs}"
     ${ARGN})
 
-  add_enclave_executable(${ENCLAVE_TARGET} ${ENCLAVE_SOURCES})
+  # Add an enclave with LVI mitigation when the `ENABLE_LVI_MITIGATION` argument or the `ENABLE_LVI_MITIGATION_TESTS`
+  # global variable is explicitly set.
+  if (LVI_MITIGATION MATCHES ControlFlow AND
+      ENCLAVE_ENABLE_LVI_MITIGATION OR ENABLE_LVI_MITIGATION_TESTS)
```
jhand2: does cmake allow parentheses to make the condition a little more clear? If so, add

cmake/add_enclave.cmake
```diff
@@ -121,15 +122,23 @@ endfunction()
...
   cmake_parse_arguments(ENCLAVE
     "${options}"
     "${oneValueArgs}"
     "${multiValueArgs}"
     ${ARGN})
 
-  add_enclave_executable(${ENCLAVE_TARGET} ${ENCLAVE_SOURCES})
+  # Add an enclave with LVI mitigation when the `ENABLE_LVI_MITIGATION` argument or the `ENABLE_LVI_MITIGATION_TESTS`
+  # global variable is explicitly set.
+  if (LVI_MITIGATION MATCHES ControlFlow AND
```
jhand2: Should this logic be in `add_enclave_executable` instead?

mingweishih: It then requires additional layer of parameter passing (i.e., the `ENABLE_LVI_MITIGATION`) to `add_enclave_executable`. As a result, I would still prefer to put the logic here (just extending the argument list of `add_enclave`).

The reason for having this parameter is to allow building certain tests (currently, mbedtls, libc, and libcxx) against LVI-mitigated libraries (but not run) by default when building with `LVI_MITIGATION=ControlFlow`. So that the CI build will still test linking those libraries.

Any other suggestion for doing this is welcome.

@CodeMonkeyLeet Please also take a look.

jhand2: It seems a little confusing that `ENABLE_LVI_MITIGATION_TESTS` only builds the tests, they won't actually run with ctest, right?

I see what you're trying to do, I just worry that the logic is confusing.

mingweishih: `ENABLE_LVI_MITIGATION_TESTS` is a global variable set during cmake initialization.
`ENABLE_LVI_MITIGATION_TESTS=ON` does both (build can run with ctest), while
`ENABLE_LVI_MITIGATION_TESTS=OFF` (default value) does nothing.

`ENABLE_LVI_MITIGATION` here is an argument passing into `add_enclave`, which results in building the enclave with LVI mitigation when it's set to `ON`. This is used to complement the default case so that we can selectively build some tests (but not run).

jhand2: Oh ok, I misunderstood. Thanks for clarifying

radhikaj:
Bors r+

mingweishih:
> @mingweishih
> 
> How much does this buy us? If it’s not much, I favor keeping the LVI tests as they are and running with each bors job.

We can do a bors try and let @BRMcLaren compare. I think the difference should not be too small.

mingweishih:
bors try

cmake/add_enclave.cmake
```diff
@@ -121,15 +122,23 @@ endfunction()
...
+      ENCLAVE_ENABLE_LVI_MITIGATION OR ENABLE_LVI_MITIGATION_TESTS)
+    add_enclave_executable(${ENCLAVE_TARGET} ${ENCLAVE_SOURCES})
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Actually, it looks like these tests should be updated not to invoke the macro directly and should have been calling `add_enclave()` ... please update these as well:

- [ ] tests/syscall/epoll/enc/CMakeList.txt
- [ ] tests/syscall/poller/enc/CMakeList.txt
- [ ] tests/syscall/sendmsg/enc/CMakeList.txt

mingweishih:
> Nice to see the more optimization of the build options available, this work is greatly appreciated!
> 
> I echo Jordan's sentiment that it's really confusing to figure out just exactly what is happening in the build system by reading the code, and I attribute that to the complexity of interactions between the LVI parameters, and the overloading of responsibilities in each setting, which is undesirable:
> 
> * `LVI_MITIGATION` not only conveys the type of mitigation desired, but also indicates that custom enclave macros like `add_enclave_executable()` should additionally build an LVI version of the target. In the end user case, it is also a coupled with setting of `LVI_MITIGATION_BINDIR` in openenclave-lvi-mitigation-config.cmake.in.
> * `ENABLE_LVI_MITIGATION_TESTS` implies build everything, but only if you call `add_enclave()` and not `add_enclave_executable()` directly as some tests do, and it also means add all the additional LVI tests for this.
> 
> I would really like to split out the build implication from the `LVI_MITIGATION` property and it's coupling with other parameters like `LVI_MITIGATION_BINDIR` but it has a public implication now due to the cmake.in; we might revisit that at some point later. I would still like to simplify the interaction between the new flags introduced though and I would suggest instead:
> 
> * Add a global `LVI_MITIGATION_SKIP_TESTS` value instead of `ENABLE_LVI_MITIGATION_TESTS`.
>   
>   * Modify `add_enclave_test()` to ignore the `LVI_MITIGATION` if this is specified so that it does not add the tests.
>   * Modify `add_enclave_executable()` to ignore the `LVI_MITIGATION` if this is specified so that it does not produce the extra binaries.
>   * Rename the original `add_enclave_executable()` to a new `add_lvi_enclave_executable()` function.
> * Add a `ADD_LVI_MITIGATION` parameter to the `add_enclave()` function.
>   
>   * The test subset that needs this always sets the property to ON when calling `add_enclave()`. It is not contingent on `LVI_MITIGATION`.
>   * `add_enclave()` also does not check `LVI_MITIGATION` or `LVI_MITIGATION_SKIP_TESTS`. It only checks `ADD_LVI_MITIGATION` and calls either `add_enclave_executable()` or `add_lvi_enclave_executable()` depending on whether the value it off or on respectively.
> 
> This makes the configuration split:
> 
> * _CI:_ `LVI_MITIGATION := ControlFlow`, `LVI_MITIGATION_SKIP_TESTS := ON`, and `ADD_LVI_MITIGATION := ON` for each of the selected tests.
> * _Nightly:_ `LVI_MITIGATION := ControlFlow`, **`LVI_MITIGATION_SKIP_TESTS := OFF`**, and `ADD_LVI_MITIGATION := ON` for each of the selected tests.

The original behavior of `add_enclave_executable` is:
- `LVI_MITIGATION=OFF`: Add normal enclave target
- `LVI_MITIGATION=ON`: Add normal and lvi-mitigated enclave targets
This allows for having minimal impacts/changes to the original build flow.
This also means that `LVI_MITIGATION=ON` applies to all test cases that use  `add_enclave_executable`, and the LVI PR tries to apply all the test cases with this wrapper.

If I understand correctly, now the suggestion is to selectively adding lvi-mitigated enclave targets. This means that there will be no option to build and run the all the tests with both normal and LVI mitigation (unless setting `ADD_LVI_MITIGATION := ON` to all the tests, which I don't think that is desirable based on the description).  And then my question will be: how to select those tests?

CodeMonkeyLeet:
@mingweishih 
> The original behavior of `add_enclave_executable` is:
> - `LVI_MITIGATION=OFF`: Add normal enclave target
> - `LVI_MITIGATION=ON`: Add normal and lvi-mitigated enclave targets
> This allows for having minimal impacts/changes to the original build flow.
> This also means that `LVI_MITIGATION=ON` applies to all test cases that use  `add_enclave_executable`, and the LVI PR tries to apply all the test cases with this wrapper.
>
> If I understand correctly, now the suggestion is to selectively adding lvi-mitigated enclave targets. > This means that there will be no option to build and run the all the tests with both normal and LVI mitigation (unless setting `ADD_LVI_MITIGATION := ON` to all the tests, which I don't think that is desirable based on the description).  And then my question will be: how to select those tests?

No, in the proposed set of changes, setting `LVI_MITIGATION` to `ControlFlow` behaves _exactly_ as it does today: it builds all both versions of all targets (binaries, libs etc.) and adds all tests.

If you additionally specify `LVI_MITIGATION_SKIP_TESTS` to `ON`, it behaves similarly to a non-LVI mitigated build today, except that it:
- skips building the additional LVI binaries, which today are just test enclaves in OE.
- skips adding the LVI copies of the tests
- still builds libraries and other targets with LVI options
- still leaves intact any assumptions around `LVI_MITIGATION_BINDIR` being configured.

If you add `ADD_LVI_MITIGATION` on top of that, it re-enables the binary build of the additional LVI versions of specific enclaves. 

If you add `ADD_LVI_MITIGATION` when the global `LVI_MITIGATION_SKIP_TESTS` has not been set, it is a no-op. It functionally means "Ignore `LVI_MITIGATION_SKIP_TESTS` when building this enclave". Perhaps it could be renamed to be more faithful to that semantic as `ALWAYS_BUILD_LVI_MITIGATED_VERSION`. It has no effect on `LVI_MITIGATION_SKIP_TESTS` suppressing the addition of the test.

mingweishih:
bors try

mingweishih:
bors try

mingweishih:
@radhikaj a data point: 
- try without this PR: 1h 12m 40s
- try with this PR: 44m 12s
Note that the time can be further optimized (require another PR to openclave-ci after this PR) for Win 2016 builds.

mingweishih:
> @mingweishih
> 
> > The original behavior of `add_enclave_executable` is:
> > 
> > * `LVI_MITIGATION=OFF`: Add normal enclave target
> > * `LVI_MITIGATION=ON`: Add normal and lvi-mitigated enclave targets
> >   This allows for having minimal impacts/changes to the original build flow.
> >   This also means that `LVI_MITIGATION=ON` applies to all test cases that use  `add_enclave_executable`, and the LVI PR tries to apply all the test cases with this wrapper.
> > 
> > If I understand correctly, now the suggestion is to selectively adding lvi-mitigated enclave targets. > This means that there will be no option to build and run the all the tests with both normal and LVI mitigation (unless setting `ADD_LVI_MITIGATION := ON` to all the tests, which I don't think that is desirable based on the description).  And then my question will be: how to select those tests?
> 
> No, in the proposed set of changes, setting `LVI_MITIGATION` to `ControlFlow` behaves _exactly_ as it does today: it builds both versions of all targets (binaries, libs etc.) and adds all tests.
> 
> If you additionally specify `LVI_MITIGATION_SKIP_TESTS` to `ON`, it behaves similarly to a non-LVI mitigated build today, except that it:
> 
> * skips building the additional LVI binaries, which today are just test enclaves in OE.
> * skips adding the LVI copies of the tests
> * still builds libraries and other targets with LVI options
> * still leaves intact any assumptions around `LVI_MITIGATION_BINDIR` being configured.
> 
> If you add `ADD_LVI_MITIGATION` on top of that, it re-enables the binary build of the additional LVI versions of specific enclaves.
> 
> If you add `ADD_LVI_MITIGATION` when the global `LVI_MITIGATION_SKIP_TESTS` has not been set, it is a no-op. It functionally means "Ignore `LVI_MITIGATION_SKIP_TESTS` when building this enclave". Perhaps it could be renamed to be more faithful to that semantic as `ALWAYS_BUILD_LVI_MITIGATED_VERSION`. It has no effect on `LVI_MITIGATION_SKIP_TESTS` suppressing the addition of the test.

The PR is updated. Please take another look. Thanks!

cmake/add_enclave.cmake
```diff
@@ -121,15 +122,22 @@ endfunction()
...
   cmake_parse_arguments(ENCLAVE
     "${options}"
     "${oneValueArgs}"
     "${multiValueArgs}"
     ${ARGN})
 
-  add_enclave_executable(${ENCLAVE_TARGET} ${ENCLAVE_SOURCES})
+  add_executable(${ENCLAVE_TARGET} ${ENCLAVE_SOURCES})
+  # Add an enclave with LVI mitigation when the `ADD_LVI_MITIGATION` argument (with `ENCLAVE` prefix)
+  # is set or the `LVI_MITIGATION_SKIP_TESTS` global variable is not set.
```
CodeMonkeyLeet: Given the confusion in our conversation on the review thread, I suggest explaining the semantics rather that rephrasing the boolean logic that can be read from the code:
```suggestion
  # Add an enclave with LVI mitigation if LVI_MITIGATION is globally configured.
  #
  # If the LVI_MITIGATION_SKIP_TESTS global variable is set, then it takes
  # precedence and suppress the addition of LVI mitigated binaries (which are
  # primarily test binaries in the OE SDK). This variable also skips adding ctests
  # for the LVI mitigated binaries in add_enclave_test.cmake.
  #
  # The ADD_LVI_MITIGATION argument to add_enclave() can override LVI_MITIGATION_SKIP_TESTS
  # on a per enclave basis. This parameter has no effect if either LVI_MITIGATION or
  # LVI_MITIGATION_SKIP_TESTS is not specified.
  # It only re-enables the additional LVI-mitigated build of the specified enclave.
  # It does not enable the additional ctest against the LVI-mitigated version of
  # the enclave.
```

cmake/enclave_cmake_wrappers.cmake
```diff
@@ -9,19 +9,18 @@ if (LVI_MITIGATION MATCHES ControlFlow)
     include(apply_lvi_mitigation)
 endif()
 
-# Wrapper of `add_executable`
-macro(add_enclave_executable NAME)
-    add_executable(${NAME} ${ARGN})
-    if (LVI_MITIGATION MATCHES ControlFlow)
+# Wrapper of `add_executable` for adding an enclave with LVI mitigation.
```
CodeMonkeyLeet: Consider adding the suggestion that this should only be called from the `add_enclave` function, unlike the other macros.

cmake/enclave_cmake_wrappers.cmake
```diff
@@ -9,19 +9,18 @@ if (LVI_MITIGATION MATCHES ControlFlow)
...
         apply_lvi_mitigation(${NAME}-lvi-cfg)
     endif()
-endmacro(add_enclave_executable)
+endmacro(add_lvi_enclave_executable)
 
 # Wrapper of `add_dependencies`
 macro(add_enclave_dependencies NAME)
     add_dependencies(${NAME} ${ARGN})
-    if (LVI_MITIGATION MATCHES ControlFlow)
+    if (LVI_MITIGATION MATCHES ControlFlow AND TARGET ${NAME}-lvi-cfg)
```
CodeMonkeyLeet: Non-blocking: This is another example where it works, but intent is difficult to infer from the implementation, and it takes work on the part of the reader to reverse engineer:

* Why is it necessary to specify _both_ conditions? 
    * The presence of the LVI target should indicate that[ LVI_MITIGATION was specified, no?
    * Is there a condition we are concerned about when a previous build of that target is left over but the CMake configuration has changed? 
       * My understanding is that CMake does not permit this, but maybe this is a concern.
    * Alternatively, instead of using an artifact that implicitly encapsulates several other checks like the new flags introduced, is it a clearer expression of intent the wrap the multi-condition check in a macro/function and invoke that everywhere?
        * Another side effect to consider is this: is it desirable that a missing target causes the logic to be skipped with no errror (e.g. race condition in a build enviornment removes the lvi target while a build is running.)

mingweishih: The side effect is exact the reason of adding this additional check.
For example  when `LVI_MITIGATION=ControlFlow` but `LVI_MITIGATION_SKIP_TESTS=ON`,
https://github.com/openenclave/openenclave/blob/cb79162e270da55a1cfa266f083719ac5038bd11/tests/ecall/enc/CMakeLists.txt#L15

Since the test is skipped, the enclave with lvi mitigation is not added. When calling the wrapper, cmake raises error of `the target does not exist` after passing the `if (LVI_MITIGATION MATCHES ControlFlow)` check.

I think check only the existence of the target should be sufficient. I'll update the PR accordingly.

Regarding the potential concern, I don't think will be a problem because once the CMake configuration is changed, CMake will re-initialize the dependency tree instead of using the cached one.

.jenkins/Jenkinsfile
```diff
@@ -32,6 +32,7 @@ def ACCTest(String label, String compiler, String build_type, String lvi_mitigat
                                -DCMAKE_BUILD_TYPE=${build_type}                       \
                                -DLVI_MITIGATION=${lvi_mitigation}                     \
                                -DLVI_MITIGATION_BINDIR=/usr/local/lvi-mitigation/bin  \
+                               -DLVI_MITIGATION_SKIP_TESTS=ON                         \
```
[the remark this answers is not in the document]
mingweishih: This depends on the changes to the openenclave-ci (where implements the logic of `win2016CrossCompile` build). I'll submit a PR there first.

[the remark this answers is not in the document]
mingweishih: Addressed here: https://github.com/openenclave/openenclave-ci/pull/15
TODO: update the nightly_build.jenkinesfile

mingweishih:
bors try

mingweishih:
bors try

PR #2754, opened 2020-03-27
uncomment failing rhel tests

BRMcLaren:

PR #2758, opened 2020-03-27
CI/CD build Optimization

mingweishih:
The PR optimizes the CI/CD build by reducing the number of configurations specified in the Jenkinsfile.
Also, create a copy of the original Jenkinsfile as nightly_build.Jenkinsfile (plus minor modification to enable full tests with LVI mitigation), which we should consider run as the nighly build.

### DIFF
--- a/.jenkins/nightly_build.Jenkinsfile
+++ b/.jenkins/nightly_build.Jenkinsfile
@@ -20,7 +20,7 @@ AGENTS_LABELS = [
 ]
 
 
-def ACCTest(String label, String compiler, String build_type, String lvi_mitigation = 'None') {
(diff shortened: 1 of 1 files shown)


BRMcLaren:
bors try

BRMcLaren:
@mingweishih would you be so kind as to do the same changes outline here?

https://github.com/openenclave/openenclave/pull/2754

We are going to end up in a merge conflict otherwise as we are splitting the Jenkins files.

radhikaj:
> @mingweishih would you be so kind as to do the same changes outline here?
> 
> #2754
> 
> We are going to end up in a merge conflict otherwise as we are splitting the Jenkins files.

@BRMcLaren Can we get this change in and then resolve conflicts on the other please?

PR #2762, opened 2020-03-28
fix nightly builds syntax

BRMcLaren:
There is a missing quotation mark in the Nightly build, Jenkins's syntax is not valid and is leading to failures.


BRMcLaren:
I've hooked up the backend nightly build to use this file over the weekend if it remains stable, we know this is the fix needed.

BRMcLaren:
bors r+

@achamayou  Thanks for taking the time out of your weekend!

PR #2767, opened 2020-03-31
Remove accidental dependencies on libc headers from oecore

jhand2:
There are numerous places in oecore and edger8r-generated code where standard C headers are included as system headers (i.e. #include <stdlib.h>) when the intention is really to use the corresponding corelibc header. In those instances explicitly include the corelibc header.

In instances in which standard C names are needed by 3rd party code, include those names only when compiling those files rather than making it the default for all files in oecore.

### DIFF
--- a/enclave/core/CMakeLists.txt
+++ b/enclave/core/CMakeLists.txt
@@ -154,6 +154,9 @@ endif ()
 
 if (USE_DLMALLOC)
     list(APPEND PLATFORM_SRC malloc.c)
+
+    list (APPEND NEEDS_STDC_NAMES malloc.c)
+
     # Unfortunately dlmalloc uses GNU extension that allows arithmetic
     # null pointers.
     set_source_files_properties(malloc.c
@@ -198,15 +201,37 @@ add_enclave_library(oecore STATIC
     tracee.c
     ${PLATFORM_SRC})
 
-# Suppress type conversion warnings introduced by 3rdparty code
-set_source_files_properties(${MUSL_SRC_DIR}/prng/rand.c ROPERTIES
+# Some files in oecore come directly from the musl source. For these files,
+# corelibc functions with stdc names are required.
+# Additionally, suppress type conversion warnings introduced by 3rdparty code
+set (CORELIBC_INCLUDES ${PROJECT_SOURCE_DIR}/include/openenclave/corelibc)
+
+list (APPEND NEEDS_STDC_NAMES
+    ${MUSL_SRC_DIR}/prng/rand.c
+    ${MUSL_SRC_DIR}/string/memmove.c
+    ${MUSL_SRC_DIR}/string/memset.c
+    ${MUSL_SRC_DIR}/string/memcmp.c
+    ${MUSL_SRC_DIR}/string/memcpy.c
+    sgx/sgx_t_wrapper.c
+    sgx/switchless_t_wrapper.c
+    __secs_to_tm.c
+    debugmalloc.c
+    strtok_r.c
+    tee_t_wrapper.c)
+
+list (APPEND W_NO_CONVERSION
+    ${MUSL_SRC_DIR}/prng/rand.c
+    ${MUSL_SRC_DIR}/string/memmove.c
+    ${MUSL_SRC_DIR}/string/memset.c
+    __secs_to_tm.c)
+
+set_source_files_properties(${W_NO_CONVERSION} PROPERTIES
     COMPILE_FLAGS -Wno-conversion)
-set_source_files_properties(${MUSL_SRC_DIR}/string/memmove.c PROPERTIES
-    COMPILE_FLAGS -Wno-sign-conversion)
-set_source_files_properties(${MUSL_SRC_DIR}/string/memset.c PROPERTIES
-    COMPILE_FLAGS -Wno-conversion)
-set_source_files_properties(__secs_to_tm.c PROPERTIES
-    COMPILE_FLAGS -Wno-conversion)
+
+set_property(SOURCE ${NEEDS_STDC_NAMES} APPEND_STRING PROPERTY
+    COMPILE_FLAGS " -I${CORELIBC_INCLUDES}")
+set_property(SOURCE ${NEEDS_STDC_NAMES} APPEND PROPERTY
+    COMPILE_DEFINITIONS OE_NEED_STDC_NAMES)
 
 maybe_build_using_clangw(oecore)
 
@@ -229,10 +254,6 @@ endif()
 if (CMAKE_C_COMPILER_ID MATCHES GNU)
     enclave_compile_options(oecore PRIVATE -Wjump-misses-init)
 endif()
-
-# This directory contains enclave core libc headers:
-enclave_include_directories(oecore PRIVATE
-    ${PROJECT_SOURCE_DIR}/include/openenclave/corelibc)
 
 if (OE_SGX)
     set_source_files_properties(sgx/keys.c PROPERTIES COMPILE_FLAGS -Wno-type-limits)
--- a/enclave/core/arena.c
+++ b/enclave/core/arena.c
@@ -2,12 +2,12 @@
 // Licensed under the MIT License.
 
 #include "arena.h"
+#include <openenclave/corelibc/string.h>
 #include <openenclave/edger8r/common.h>
 #include <openenclave/internal/raise.h>
 #include <openenclave/internal/safemath.h>
 #include <openenclave/internal/thread.h>
 #include <openenclave/internal/utils.h>
-#include <string.h>
 
 // The per-thread shared memory arena
 static __thread shared_memory_arena_t _arena = {0};
--- a/enclave/core/ctype.c
+++ b/enclave/core/ctype.c
@@ -1,7 +1,7 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
-#include <ctype.h>
+#include <openenclave/corelibc/ctype.h>
 
 static const unsigned short _ctype_b[384] = {
     0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000,
--- a/include/openenclave/edger8r/common.h
+++ b/include/openenclave/edger8r/common.h
@@ -66,24 +66,24 @@ done:
     return result;
 }
 
-#define OE_ADD_SIZE(total, size)                                 \
-    do                                                           \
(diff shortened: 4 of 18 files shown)


enclave/core/CMakeLists.txt
```diff
@@ -198,16 +201,38 @@ add_enclave_library(oecore STATIC
...
-    COMPILE_FLAGS -Wno-sign-conversion)
-set_source_files_properties(${MUSL_SRC_DIR}/string/memset.c PROPERTIES
-    COMPILE_FLAGS -Wno-conversion)
-set_source_files_properties(__secs_to_tm.c PROPERTIES
+# Some files in oecore come directly from the musl source. For these files,
+# corelibc functions with stdc names are required.
+# Additionally, suppress type conversion warnings introduced by 3rdparty code
+set (CORELIBC_INCLUDES ${PROJECT_SOURCE_DIR}/include/openenclave/corelibc)
+
+list (APPEND NEEDS_STDC_NAMES
```
jhand2: Could probably add a comment to better explain what this is for

enclave/core/CMakeLists.txt
```diff
@@ -154,6 +154,9 @@ endif ()
 
 if (USE_DLMALLOC)
     list(APPEND PLATFORM_SRC malloc.c)
+
+    list (APPEND NEEDS_STDC_NAMES malloc.c)
```
dthaler: nit: line 156 has no space before ( and here does.  Be consistent.

include/openenclave/edger8r/common.h
```diff
@@ -66,24 +66,24 @@ OE_INLINE oe_result_t oe_add_size(size_t* total, size_t size)
...
-        }                                                        \
-        if (oe_add_size((size_t*)&total, (size_t)size) != OE_OK) \
-        {                                                        \
-            _result = OE_INTEGER_OVERFLOW;                       \
-            goto done;                                           \
-        }                                                        \
+#define OE_ADD_SIZE(total, size)                                   \
+    do                                                             \
+    {                                                              \
+        if (sizeof(total) > sizeof(size_t) && total > OE_SIZE_MAX) \
```
anakrish: It would be better to separate out the edger8r changes to a separate PR.

tests/oeedger8r/edl/array.edl
```diff
@@ -2,6 +2,8 @@
 // Licensed under the MIT License.
 
 enclave  {
+    include "wchar.h"
```
anakrish: This seems like a breaking change.
Do user written edl files that use wchart_t now have to explicitly do this include?

tools/oeedger8r/src/Headers.ml
```diff
@@ -173,9 +173,6 @@ let generate_args (ec : enclave_content) =
     "#ifndef " ^ guard_macro;
     "#define " ^ guard_macro;
     "";
-    "#include <stdint.h>";
-    "#include <stdlib.h> /* for wchar_t */";
-    "";
     include_errno;
```
anakrish: This seems like a breaking change.
Do user written edl files that use wchart_t now have to explicitly do this include?

enclave/core/CMakeLists.txt
```diff
@@ -198,16 +201,38 @@ add_enclave_library(oecore STATIC
...
-    COMPILE_FLAGS -Wno-conversion)
-set_source_files_properties(${MUSL_SRC_DIR}/string/memmove.c PROPERTIES
-    COMPILE_FLAGS -Wno-sign-conversion)
-set_source_files_properties(${MUSL_SRC_DIR}/string/memset.c PROPERTIES
-    COMPILE_FLAGS -Wno-conversion)
-set_source_files_properties(__secs_to_tm.c PROPERTIES
+# Some files in oecore come directly from the musl source. For these files,
+# corelibc functions with stdc names are required.
+# Additionally, suppress type conversion warnings introduced by 3rdparty code
+set (CORELIBC_INCLUDES ${PROJECT_SOURCE_DIR}/include/openenclave/corelibc)
```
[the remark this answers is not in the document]
jhand2: We already do this today, I'm just making it more explicit which files actually need it. This is mostly for 3rd party files that we use in oecore so the implementations don't need to be rewritten.

enclave/core/CMakeLists.txt
```diff
@@ -198,16 +201,38 @@ add_enclave_library(oecore STATIC
...
+    ${MUSL_SRC_DIR}/prng/rand.c
+    ${MUSL_SRC_DIR}/string/memmove.c
+    ${MUSL_SRC_DIR}/string/memset.c
+    __secs_to_tm.c)
+
+set_source_files_properties(${W_NO_CONVERSION} PROPERTIES
     COMPILE_FLAGS -Wno-conversion)
 
+set_property(SOURCE ${NEEDS_STDC_NAMES} APPEND_STRING PROPERTY
+    COMPILE_FLAGS " -I${CORELIBC_INCLUDES}")
```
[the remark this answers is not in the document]
jhand2: I don't think we should. It is an implementation detail of oecore that some of the files it uses internally are actually lifted from musl. No need to expose that fact by making a separate library.

enclave/core/arena.c
```diff
@@ -2,12 +2,12 @@
 // Licensed under the MIT License.
 
 #include "arena.h"
+#include <openenclave/corelibc/string.h>
```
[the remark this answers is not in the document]
jhand2: That include is only added to the path for 3rd party C files. For code that we write there is no need to obfuscate where the header is coming from.

enclave/core/ctype.c
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
-#include <ctype.h>
+#include <openenclave/corelibc/ctype.h>
```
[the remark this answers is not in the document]
jhand2: see above

include/openenclave/edger8r/common.h
jhand2: Sure, that seems reasonable

tests/oeedger8r/edl/array.edl
jhand2: It is, but it is intentional. wchar_t is not a builtin C type and anyone who would want to use it in normal C has to include it somehow. In the same way that you would need to include a header for a struct defined by libc if you want to use it.

It also isn't really an official contract. People were getting wchar_t for free by chance, it is not documented that wchar_t is a builtin type.

Even still, I do agree this has the potential to break existing code and should be done in a separate change with an addition to the changelog.

tools/oeedger8r/src/Headers.ml
jhand2: see above. I'll split it out.

jhand2:
> It would be better to decouple the edger8r changes from this PR. Those seem like breaking changes that would require users to explicitly include wchar.h in their EDL files.

Will do

> Additionally, the PR description is not accurate in regards to edger8r.

Fair

> > There are numerous places in oecore and edger8r-generated code where standard C headers are included as system headers (i.e. #include <stdlib.h>) when the intention is really to use the corresponding corelibc header. In those instances explicitly include the corelibc header.
> 
> oeedger8r has till now not had a reason to use corelibc. Its use of libc is not accidental.
> oeedger8r is based upon Intel SDK's sgx_edger8r which also includes C header files.
> 
> https://github.com/openenclave/openenclave/blob/cb79162e270da55a1cfa266f083719ac5038bd11/tools/oeedger8r/intel/CodeGen.ml#L647-L649
> 
> The goals of oeedger8r has been to generate portable code for multiple TEEs (SGX, OPTEE), be self-contained, and have most of the generated code be visible to the user so that they can figure out what is going on by reading the marshaling code.

The edger8r "free standing libc" is much more brittle than the above implies. It only works because corelibc headers are in the include path. If the corelibc headers weren't there, the compiler throws an error that it can't find any of the libc headers. 

But then when it goes to actually resolve symbols it uses the freestanding environment (I don't really know why/how this works, maybe you have a better understanding). Look at `include/openenclave/corelibc/wchar.h`. It is empty, but it has to be there for the compiler to find the header path. But then it doesn't actually use that header in compilation. Compiling with `-ffreestanding` doesn't seem to change the need for this file to be there.

Anyway, I'll back out the edger8r-specific stuff here. But I still think the reality of edger8r's dependence on libc is more complicated.

jhand2:
> The code changes LGTM; I think it would be helpful to rephrase the PR title and description to break down:
> 
>     * It would be useful to talk about the desired effects of this change (removes the function name collisions with app inclusion of std libc headers, clarity of intent in code etc.)
> 
>     * Summary of changes (e.g. consolidation of `OE_NEED_STDC_NAMES` usage from individual source files to enclave/core/CMakeLists.txt)
> 
>     * Where this fits in with related PRs towards the stated expectation or ideal design (e.g. #2769?)

That all sounds good. Currently my plan is to address the edger8r stuff in separate PRs and then come back to finish this one up.

tests/oeedger8r/edl/array.edl
anakrish: Just so I understand what you are saying,

edger8r has always had first class support for wchar, as seen in the Lexer,
https://github.com/openenclave/openenclave/blob/261947bce47abcbe195f39f32d595f7061ec18fb/tools/oeedger8r/intel/Lexer.mll#L69-L77

and Parser,
https://github.com/openenclave/openenclave/blob/261947bce47abcbe195f39f32d595f7061ec18fb/tools/oeedger8r/intel/Ast.ml#L49-L53

When you say `people were getting wchar_t for free by chance`, are your referring to edger8r generated code or something else?

Also note that edger8r supports a lot of other types like `uint64_t`, 'size_t` that are similar to `wchar_t` in that they are are not C builtin types, but rather typedefs of C builtin types. Can you clarify why the user would have to include "wchar.h" to access wchar_t, but would not need to include "stdint.h" to access 'uint64_t'?

radhikaj: @jhand2 I am concerned about this breaking change as it requires every developer using edl to change what they include. It also deviates from Intel's edger8r. Is there a way around this? Can we make it so devs are not required to #include wchar.h?

jhand2:
@radhikaj @CodeMonkeyLeet @anakrish 

As I said in my last comment, the edger8r parts of this PR are mostly replaced by #2769 and #2771

Please review those and ignore this one for now

tests/oeedger8r/edl/array.edl
jhand2: See comment at bottom of PR

PR #2769, opened 2020-03-31
Install headers needed by oeedger8r-generated code

jhand2:
In order to remove oeedger8r's dependance on stdc, a small number of corelibc headers must be published. Install these headers under openenclave/corelibc.

Additionally, add an implementation of wcslen for generated code to use.

### DIFF
--- a/enclave/core/wchar.c
+++ b/enclave/core/wchar.c
@@ -1,0 +1,12 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/corelibc/wchar.h>
+
+size_t oe_wcslen(const wchar_t* s)
+{
+    const wchar_t* a;
+    for (a = s; *s; s++)
+        ;
+    return (size_t)(s - a);
+}
--- a/include/openenclave/bits/types.h
+++ b/include/openenclave/bits/types.h
@@ -34,6 +34,15 @@ typedef long intptr_t;
 typedef long time_t;
 typedef long suseconds_t;
 
+#ifndef __cplusplus
+#if __WCHAR_MAX__ > 0x10000
+// honor -fshort-wchar
+typedef int wchar_t;
+#else
+typedef short wchar_t;
+#endif
+#endif
+
 #elif defined(_MSC_VER)
 typedef long long ssize_t;
 typedef unsigned long long size_t;
@@ -50,6 +59,11 @@ typedef long long ptrdiff_t;
 typedef long long intptr_t;
 typedef long long time_t;
 typedef long long suseconds_t;
+
+#ifndef __cplusplus
+typedef short wchar_t;
+#endif
+
 #else
 #error "unknown compiler - please adapt basic types"
 #endif
--- a/include/openenclave/bits/types.h
+++ b/include/openenclave/bits/types.h
@@ -35,12 +35,7 @@ typedef long time_t;
 typedef long suseconds_t;
 
 #ifndef __cplusplus
-#if __WCHAR_MAX__ > 0x10000
-// honor -fshort-wchar
-typedef int wchar_t;
-#else
-typedef short wchar_t;
-#endif
+typedef __WCHAR_TYPE__ wchar_t;
 #endif
 
 #elif defined(_MSC_VER)
@@ -59,11 +54,6 @@ typedef long long ptrdiff_t;
 typedef long long intptr_t;
 typedef long long time_t;
 typedef long long suseconds_t;
-
-#ifndef __cplusplus
-typedef short wchar_t;
-#endif
-
 #else
 #error "unknown compiler - please adapt basic types"
 #endif
--- a/include/openenclave/corelibc/wchar.h
+++ b/include/openenclave/corelibc/wchar.h
@@ -1,2 +1,11 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
+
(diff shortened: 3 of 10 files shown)


include/openenclave/bits/types.h
```diff
@@ -34,6 +34,15 @@ typedef long intptr_t;
 typedef long time_t;
 typedef long suseconds_t;
 
+#ifndef __cplusplus
+#if __WCHAR_MAX__ > 0x10000
+// honor -fshort-wchar
```
jhand2: This comment should be on line 42 instead

jhand2:
bors try

radhikaj:
@jhand2, trying to understand this-
1) previously did oeeger8r depend on stdc?
2) why is it preferable that it depends on core libc?

include/openenclave/bits/types.h
```diff
@@ -50,6 +59,11 @@ typedef long long ptrdiff_t;
 typedef long long intptr_t;
 typedef long long time_t;
 typedef long long suseconds_t;
+
+#ifndef __cplusplus
+typedef short wchar_t;
```
anakrish: Not clear why we need this redefinition in the same file.

include/openenclave/corelibc/wchar.h
```diff
@@ -1,2 +1,11 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
+
+#ifndef _OE_WCHAR_H
+#define _OE_WCHAR_H
+
+#include <openenclave/bits/types.h>
+
+size_t oe_wcslen(const wchar_t* s);
```
anakrish: This header and any other published header would likely need OE_EXTERNC_BEGIN, OE_EXTERNC_END closures to that they can be included in C++ files.

include/openenclave/corelibc/wchar.h
```diff
@@ -1,2 +1,11 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
+
+#ifndef _OE_WCHAR_H
+#define _OE_WCHAR_H
+
+#include <openenclave/bits/types.h>
```
anakrish: Does it make sense to add a disclaimer that these headers are intended only for internal use?

include/openenclave/bits/types.h
jhand2: For windows

include/openenclave/corelibc/wchar.h
jhand2: Ya good point. At least until we can formalize what we want to support

jhand2:
> @jhand2, trying to understand this-
> 
>     1. previously did oeeger8r depend on stdc?

Yes, for a small number of functions

> 
>     2. why is it preferable that it depends on core libc?

An enclave doesn't actually need to depend on stdc. This leaves us in a weird place where technically you don't need to link stdc but edger8r may still generate code which depends on stdc.

Also there are cases where we coerce edger8r generated code to use stdc names for corelibc functions using `OE_NEEDS_STDC_NAMES`. This can cause name collisions.

include/openenclave/bits/types.h
jhand2: Although it does seem like the windows types aren't really necessary now that I think about it. Special type definitions shouldn't be needed on the host side. I mostly just defined it for parity with the other types which are defined for both.

anakrish: Ah ok. I see the `#elfi defined(_MSC_VER)` above.

jhand2:
Looks like __WCHAR_TYPE__ is defined by gcc, hopefully that works for arm as well.

bors try

enclave/core/wchar.c
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Open Enclave SDK contributors.
```
CodeMonkeyLeet, line 1: Do we need to add an implementation here, or can we reuse the 3rdparty/musl/musl/src/string/wcslen.c implementation under the `NEEDS_STDC_NAMES` section introduced in #2767?

jhand2: The mechanism in #2767 is for compiling internal files which are part of oecore but use stdc names. Since wchar.h will be included by code compiled by users, I would prefer to use the oe_* name.

jhand2: I want to minimize the use of NEEDS_STDC_NAMES as much as possible

CodeMonkeyLeet: Side note: As you start making these distinctions in use cases, it would be useful to start documenting the guidance for this as well to contributors.

jhand2: Sounds good, I can do that as part of #2767

anakrish:
Have we got buy-in from @HernanGatta or @dthaler? 
It would be good to give them an opportunity to comment as well since OPTEE also relies on edger8r generated code.

jhand2:
sure

bors r-

jhand2:
I created a draft PR #2771 to show what it would look like for oeedger8r to make use of these newly published files. I think that might help provide more context for this PR

jhand2:
@HernanGatta @dthaler please let me know if you see any conceptual issues with this PR or #2771

IMO as long as the CI run for #2771 passes arm compilation, it should be fine from a technical perspective. These corelibc functions are the same on SGX and OP-TEE.

jhand2:
Closing in favor of #2771

PR #2774, opened 2020-04-01
add fragment tests

yanxue-intel:

### DIFF
--- a/tests/memory/README.md
+++ b/tests/memory/README.md
@@ -6,3 +6,4 @@ This directory tests enclave memory management with the following tests:
     and freeing.
   - Stress test the malloc family functions by rapid allocation and freeing
     in a multi-threaded context.
+  - Check for memory fragmentation inside an enclave after repeated mallocs and frees.
--- a/tests/memory/enc/CMakeLists.txt
+++ b/tests/memory/enc/CMakeLists.txt
@@ -14,7 +14,8 @@ add_enclave(TARGET memory_enc UUID 719ff522-610b-43bd-9991-c4d52a91a7e1
   boundaries.c
   enc.c
   stress.c
+  fragment.c
   memory_t.c)
 
 enclave_include_directories(memory_enc PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
-enclave_link_libraries(memory_enc oelibc)
+enclave_link_libraries(memory_enc oelibc oecore)
--- a/tests/memory/enc/fragment.c
+++ b/tests/memory/enc/fragment.c
@@ -1,0 +1,129 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/enclave.h>
+#include <openenclave/internal/globals.h>
+#include <openenclave/internal/tests.h>
+
+#include <malloc.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "memory_t.h"
+
+#define MALLOC_SIZE_SMALL 1024
+#define EFFICIENCY_TEST_TIMES 5000000
+
+struct mallinfo
+{
+    size_t arena;    /* non-mmapped space allocated from system */
+    size_t ordblks;  /* number of free chunks */
+    size_t smblks;   /* always 0 */
+    size_t hblks;    /* always 0 */
+    size_t hblkhd;   /* space in mmapped regions */
+    size_t usmblks;  /* maximum total allocated space */
+    size_t fsmblks;  /* always 0 */
+    size_t uordblks; /* total allocated space */
+    size_t fordblks; /* total free space */
+    size_t keepcost; /* releasable (via malloc_trim) space */
+};
+
+struct mallinfo dlmallinfo(void);
+
+static size_t _get_heap_size()
+{
+    // call malloc to update the global data.
+    free(malloc(1));
+    return dlmallinfo().keepcost;
+}
+
+static int _malloc_free_fixed_size(size_t size, int times)
+{
+    int i = 0;
+    int* buffer = NULL;
+
+    for (i = 0; i < times; i++)
+    {
+        buffer = (int*)malloc(size);
+        if (NULL == buffer)
+        {
+            return -1;
+        }
+        free(buffer);
+    }
+
+    oe_host_printf("_malloc_free_fixed_size malloc times = %d.\n", i);
+    return 0;
+}
+
+static inline size_t _randx(size_t x)
+{
+    /* Note that rand() % N is biased if RAND_MAX + 1 isn't divisible
+     * by N. But, slight probability bias doesn't really matter in these
+     * tests. */
+    return (size_t)rand() % x;
+}
+
+/*
+ * Test making `times` allocations of random sizes.
+ * If seed is < 0 just use time() as a seed. If seed >= 0, use the provided
+ * to seed the rng.
+ */
+static int _malloc_free_random_size(int times, unsigned int seed)
+{
+    int i = 0;
+    size_t size = 0;
+    void* buffer = NULL;
+
+    oe_host_printf("_malloc_free_random_size seed = %d\n", seed);
+    srand(seed);
+    for (i = 0; i < times; i++)
+    {
+        size = _randx(_get_heap_size());
+        buffer = malloc(size);
+        if (NULL == buffer)
+        {
+            return -1;
+        }
+
+        free(buffer);
+    }
+    oe_host_printf("_malloc_free_random_size malloc times = %d.\n", i);
+    return 0;
+}
+
+void test_malloc_fixed_size_fragment(void)
+{
+    // get heap size before tests
+    size_t heap_size_before_test = _get_heap_size();
+    oe_host_printf(
+        "[test_malloc_fixed_size_fragment]heap size before test : %zu.\n",
+        heap_size_before_test);
+
+    OE_TEST(
+        _malloc_free_fixed_size(MALLOC_SIZE_SMALL, EFFICIENCY_TEST_TIMES) == 0);
+
+    size_t heap_size_after_test = _get_heap_size();
+    oe_host_printf(
+        "[test_malloc_fixed_size_fragment]heap size after test : %zu.\n",
+        heap_size_after_test);
+    OE_TEST(heap_size_before_test == heap_size_after_test);
+}
+
+void test_malloc_random_size_fragment(unsigned int seed)
+{
+    // get heap size before tests
+    size_t heap_size_before_test = _get_heap_size();
+    oe_host_printf(
+        "[test_malloc_random_size_fragment]heap size before test : %zu.\n",
+        heap_size_before_test);
+
+    OE_TEST(_malloc_free_random_size(EFFICIENCY_TEST_TIMES, seed) == 0);
+
+    size_t heap_size_after_test = _get_heap_size();
+    oe_host_printf(
+        "[test_malloc_random_size_fragment]heap size after test : %zu.\n",
+        heap_size_after_test);
+    OE_TEST(heap_size_before_test == heap_size_after_test);
+}
--- a/tests/memory/host/host.cpp
+++ b/tests/memory/host/host.cpp
@@ -1,6 +1,7 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+#include <time.h>
 #include <cstdio>
 #include <thread>
 #include <vector>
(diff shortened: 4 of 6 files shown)


anakrish:
bors try

yanxue-intel:
bors try

anakrish:
bors try

jhand2:
Thanks @yanxue-intel! Could you update the PR description to say what exactly is being tested? In general it is good to provide a description with each PR to give the reviewers more context on the goals of the PR.

Thanks again!

yanxue-intel:
> Thanks @yanxue-intel! Could you update the PR description to say what exactly is being tested? In general it is good to provide a description with each PR to give the reviewers more context on the goals of the PR.
> 
> Thanks again!

This PR including two test for checking fragment in enclave.
one case is checking for 1K memory to be malloc/free in enclave 5000000 times.
the other case is checking for random size memory to be malloc/free in enclave 5000000 times.

tests/memory/README.md has a short description.

tests/memory/README.md
```diff
@@ -6,3 +6,4 @@ This directory tests enclave memory management with the following tests:
     and freeing.
   - Stress test the malloc family functions by rapid allocation and freeing
     in a multi-threaded context.
+  - Checking that fragment after mallocing fixed size and random size
```
radhikaj, line 9: Can you please elaborate on this statement? It is not clear to me exactly what is being checked.

tests/memory/enc/fragment.c
```diff
@@ -0,0 +1,121 @@
...
+    MALLINFO_FIELD_TYPE uordblks; /* total allocated space */
+    MALLINFO_FIELD_TYPE fordblks; /* total free space */
+    MALLINFO_FIELD_TYPE keepcost; /* releasable (via malloc_trim) space */
+};
+
+struct mallinfo dlmallinfo(void);
+
+static size_t _get_heap_size()
+{
+    // call malloc to update the globle data.
```
radhikaj, line 37:
```suggestion
    // call malloc to update the globlal data.
```

tests/memory/enc/fragment.c
```diff
@@ -0,0 +1,121 @@
...
+    size_t heap_size_before_test = _get_heap_size();
+    oe_host_printf(
+        "[test_malloc_fixed_size_fragment]heap size before test : %zu.\n",
+        heap_size_before_test);
+
+    OE_TEST(
+        _malloc_free_fixed_size(MALLOC_SIZE_SMALL, EFFICIENCY_TEST_TIMES) == 0);
+
+    size_t heap_size_after_test = _get_heap_size();
+    oe_host_printf(
```
radhikaj, line 102: Is there a check that can be made to the heap size after as well?

tests/memory/enc/fragment.c
```diff
@@ -0,0 +1,121 @@
...
+
+#include "memory_t.h"
+
+#define MALLOC_SIZE_SMALL 1024
+#define EFFICIENCY_TEST_TIMES 5000000
+
+#define MALLINFO_FIELD_TYPE size_t
+struct mallinfo
+{
+    MALLINFO_FIELD_TYPE arena;    /* non-mmapped space allocated from system */
```
radhikaj, line 21: I am wondering why these cant be just of type size_t instead of being of type MALLINFO_FIELD_TYPE

radhikaj:
yan.xue@intel.com Thanks for adding these tests. They are great additions to the OE test suite.

tests/memory/README.md
radhikaj, line 9:
```suggestion
  - Check for memory fragmentation inside an enclave after repeated mallocs and frees.
```

tests/memory/enc/fragment.c
yanxue-intel: Good suggestion! Thanks

anakrish:
@yanxue-intel Can you squash your commits and signoff?

yanxue-intel:
Since Linux standard header malloc.h uses int, I am not sure if OE will change the MALLINFO_FIELD_TYPE type for aligning with system malloc.h

tests/memory/enc/CMakeLists.txt
```diff
@@ -14,7 +14,8 @@ add_enclave(TARGET memory_enc UUID 719ff522-610b-43bd-9991-c4d52a91a7e1
   boundaries.c
   enc.c
   stress.c
+  fragment.c
   memory_t.c)
 
 enclave_include_directories(memory_enc PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
-enclave_link_libraries(memory_enc oelibc)
+enclave_link_libraries(memory_enc oelibc oecore)
```
jhand2, line 21: nit: I think linking oelibc should automatically link oecore

tests/memory/enc/fragment.c
```diff
@@ -0,0 +1,121 @@
...
+
+static int _malloc_free_random_size(int times)
+{
+    int i = 0;
+    size_t size = 0;
+    void* buffer = NULL;
+
+    for (i = 0; i < times; i++)
+    {
+        size = _randx(_get_heap_size());
```
jhand2, line 77: Having a test with random behaviour concerns me since regressions may not be reproducible. I understand you want to test many different values. Could we seed the RNG at the beginning of the test so that the test's behavior is deterministic?

yanxue-intel: The purpose of this case is to test random behavior. because the user cannot determine the memory usage behavior.

yanxue-intel:
> This test checks heap fragmentation at the beginning and end of the test and prints it, but it doesn't actually do anything with those values and it doesn't fail if the heap is fragmented after the test.
> 
> Do we expect the keepcost to be the same before and after the test? Within a certain range? We should test for what we expect, otherwise this test won't fail even if the heap is fragmented.

The case expects result is heap size did not change after repeated mallocs and frees.

tests/memory/enc/fragment.c
jhand2: I'm looking for deterministic randomness. The goal of the test to to try a bunch a different unexpected values. It doesn't really matter if those values change on each new run of the test.

If you seed the rng at the beginning of the test with a constant, you'll still get random value, but those random values won't change on each new run. This makes the test reproducible and regressions will also be reproducible.

kuanggang-git:

tests/memory/enc/fragment.c
yanxue-intel: Thanks jhand2 for your suggestion. this method is great for reproducible.  I concerned whether using fixed sequences may cover up the problem. maybe this sequences is fine ,but another is not.

jhand2: In my opinion having a reproducible test is better than greater coverage because having tests with large coverage that can't be reproduced makes it hard to fix the problems detected.

What do you think?

jhand2: @yanxue-intel another option would be to generate a random seed at the beginning of the test and print that seed. Then the test could accept a command-line parameter to use a passed seed instead of randomly generating one.

That way if we see a failure with a given seed, we can pass in that seed to repro the failure. This gives us both benefits.

yanxue-intel: > can't



> In my opinion having a reproducible test is better than greater coverage because having tests with large coverage that can't be reproduced makes it hard to fix the problems detected.
> 
> What do you think?

agree with you.

dthaler:
Does malloc have POSIX compliant behavior or normal Linux behavior which is that it never returns NULL.  The Linux man page https://linux.die.net/man/3/malloc explains: 

> By default, Linux follows an optimistic memory allocation strategy. This means that when **malloc() returns non-NULL there is no guarantee that the memory really is available**. In case it turns out that the system is out of memory, one or more processes will be killed by the OOM killer. For more information, see the description of /proc/sys/vm/overcommit_memory and /proc/sys/vm/oom_adj in proc(5), and the Linux kernel source file Documentation/vm/overcommit-accounting. 

Even when out of memory, Linux malloc will return non-NULL, so one cannot test for out of memory by looking for a NULL return from malloc.   To do that check, one needs to make sure malloc is posix compliant and does not have the Linux behavior.

tests/memory/host/host.cpp
```diff
@@ -143,6 +153,12 @@ int main(int argc, const char* argv[])
     printf("===Starting malloc boundary test.\n");
```
jhand2, line 153:
```suggestion
    int seed = -1;
    if (argc > 1) {
        // Try to parse seed passed from command-line
        int len = strlen(argv[2]);
        char* endptr = NULL;
        
        seed = strtol(argv[2], endptr, 10);
        if (endptr - argv[2] != len) {
            fprintf(stderr, "%s is not a valid seed\n", argv[2]);
            return 1;
        }
    }

    printf("===Starting malloc boundary test.\n");
```

jhand2: Note you'll have to include stdlib.h and string.h in this file now.

yanxue-intel: looks good, thanks

jhand2:
bors try

yanxue-intel:
> Does malloc have POSIX compliant behavior or normal Linux behavior which is that it never returns NULL. The Linux man page https://linux.die.net/man/3/malloc explains:
> 
> > By default, Linux follows an optimistic memory allocation strategy. This means that when **malloc() returns non-NULL there is no guarantee that the memory really is available**. In case it turns out that the system is out of memory, one or more processes will be killed by the OOM killer. For more information, see the description of /proc/sys/vm/overcommit_memory and /proc/sys/vm/oom_adj in proc(5), and the Linux kernel source file Documentation/vm/overcommit-accounting.
> 
> Even when out of memory, Linux malloc will return non-NULL, so one cannot test for out of memory by looking for a NULL return from malloc. To do that check, one needs to make sure malloc is posix compliant and does not have the Linux behavior.

The Linux man page explains malloc return error:

> On error, these functions return NULL. NULL may also be returned by a successful call to malloc() with a size of zero, or by a successful call to calloc() with nmemb or size equal to zero. 

Just avoid use malloc(0),  it is ok to check that malloc return NULL when tests for out of memory.
if i am wrong, please correct me.

jhand2:
@dthaler @yanxue-intel 

Let's say we call `malloc()` and it fails and returns a garbage pointer and does not set errno. When we call `free()`, it should segfault, correct?

In that case the test will still fail, which is what we want to happen. I think this test still does everything we want with no false negatives. Do you agree?

dthaler:
> @dthaler @yanxue-intel
> Let's say we call malloc() and it fails and returns a garbage pointer and does not set errno. When we call free(), it should segfault, correct?

I don't believe that is correct, no.  It is not a "garbage" pointer per se, it is a pointer with no physical pages available, so trying to dereference the pointer will crash, but freeing it will not necessarily crash.

> In that case the test will still fail, which is what we want to happen. I think this test still does everything we want with no false negatives. Do you agree?

Only if the test *dereferences* the pointer to read or write a byte there.

radhikaj:
> > @dthaler @yanxue-intel
> > Let's say we call malloc() and it fails and returns a garbage pointer and does not set errno. When we call free(), it should segfault, correct?
> 
> I don't believe that is correct, no. It is not a "garbage" pointer per se, it is a pointer with no physical pages available, so trying to dereference the pointer will crash, but freeing it will not necessarily crash.
> 
> > In that case the test will still fail, which is what we want to happen. I think this test still does everything we want with no false negatives. Do you agree?
> 
> Only if the test _dereferences_ the pointer to read or write a byte there.

We need to investigate what the behavior is with dlmalloc inside an enclave. Does it return null when it runs out of memory?

radhikaj:
@yanxue-intel Can you please investigate why the test failed.

yanxue-intel:
> test_typeinfo

Thanks for the reminder. This problem cannot be reproduced on my local machine, I'm looking for an environment where I can reproduce the problem.

radhikaj:
bors try

anakrish:
bors try

anakrish:
@yanxue-intel Can you signoff your PR?

yanxue-intel:
> @yanxue-intel Can you signoff your PR?

OK

jhand2:
bors try

anakrish:
bors try

radhikaj:
bors try

radhikaj:
bors try

anakrish:
The memory test is timing out on Debug builds because, in debug builds, the `debug malloc` allocator is turned on. This allocator sets each byte in the allocated memory to 0xAA:
https://github.com/openenclave/openenclave/blob/704b40733d29b8d97d195fe732fabdcd65a39ec9/enclave/core/debugmalloc.c#L319

Since the test is performing a large number of allocations, debug malloc causes memset to be called a large number of times to set the allocated memory to 0xAA. This doesn't happen with non debug builds. In addition to the memory write overhead, debug malloc is likely triggering a large number of EPC page caching/flushing.

The fix is to disable the test if `USE_DEBUG_MALLOC` is true/on.

yanxue-intel:
> The memory test is timing out on Debug builds because, in debug builds, the `debug malloc` allocator is turned on. This allocator sets each byte in the allocated memory to 0xAA:
> https://github.com/openenclave/openenclave/blob/704b40733d29b8d97d195fe732fabdcd65a39ec9/enclave/core/debugmalloc.c#L319
> 
> Since the test is performing a large number of allocations, debug malloc causes memset to be called a large number of times to set the allocated memory to 0xAA. This doesn't happen with non debug builds. In addition to the memory write overhead, debug malloc is likely triggering a large number of EPC page caching/flushing.
> 
> The fix is to disable the test if `USE_DEBUG_MALLOC` is true/on.

I have disabled fragment tests when `USE_DEBUG_MALLOC` is true/on.

radhikaj:
bors try

PR #2779, opened 2020-04-01
Remove OE_SET_ENCLAVE_SGX; Explain why enclave has two threads.

### LINKED ISSUE #2715: Switchless sample unexpectedly has multiple values of NumTCS
https://github.com/openenclave/openenclave/blob/master/samples/switchless/enclave/enc.c line 39 sets NumTCS to 2.
https://github.com/openenclave/openenclave/blob/master/samples/switchless/enclave/switchless.conf line 8 sets NumTCS to 1.

It's unclear why the sample uses both a .conf file and a OE_SET_ENCLAVE_SGX() macro, let alone why they have different values. It's also unclear which one takes precedence, which overlaps with issue #2714

anakrish:
To be consistent with other samples, OE_SET_ENCLAVE_SGX is removed.

The sample originally had one enclave thread to demonstrate switchless ocalls.
The sample has been updated to demonstrate switchless ecalls using a single
enclave worker thread to service switchless ecalls. This makes the number of
threads (NumTCS) 2.

Fixes #2715

### DIFF
--- a/samples/switchless/enclave/enc.c
+++ b/samples/switchless/enclave/enc.c
@@ -39,11 +39,3 @@ void enclave_decrement_regular(int* n)
 {
     *n = *n - 1;
 }
-
-OE_SET_ENCLAVE_SGX(
-    1,    /* ProductID */
-    1,    /* SecurityVersion */
-    true, /* AllowDebug */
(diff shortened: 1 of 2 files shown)


anakrish:
bors try

PR #2781, opened 2020-04-01
Refactor additional syscall EDL files to use internal structures

jhand2:
Refactor the remaining syscall EDL files to use internal structures and types from published headers.

A couple notes:
1. This PR doesn't include epoll.edl. That case is more complicated because the structure is packed on x86_64. This will be handled separately.
2. Changes in this PR revealed a circular dependency bug which can cause build failures depending on the order of inclusion of corelibc headers. It is a small change to some internal headers so I went ahead and fixed it here, but in a separate commit.


common/edl/time.edl
```diff
@@ -14,8 +14,14 @@
 
 enclave
 {
-    include "openenclave/corelibc/time.h"
-    include "openenclave/internal/syscall/types.h"
+    // Needed for integral type time_t
+    include "openenclave/bits/types.h"
```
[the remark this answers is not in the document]
jhand2: Oh, I didn't see that, good point. I can remove it.

common/edl/poll.edl
```diff
@@ -14,9 +14,14 @@
 
 enclave {
 
-    include "openenclave/internal/syscall/sys/poll.h"
-    include "openenclave/internal/syscall/poll.h"
-    include "openenclave/internal/syscall/types.h"
+    include "openenclave/bits/edl/syscall_types.h"
+
+    struct oe_pollfd
```
anakrish, line 19: According to our naming convention guidelines, the structs ought to have an `_t` prefix.
https://github.com/openenclave/openenclave/blob/master/docs/DevelopmentGuide.md

common/edl/time.edl
```diff
@@ -14,8 +14,14 @@
 
 enclave
 {
-    include "openenclave/corelibc/time.h"
-    include "openenclave/internal/syscall/types.h"
+    // Needed for integral type time_t
```
anakrish, line 17: `time_t` is [implementation defined](https://en.cppreference.com/w/c/chrono/time_t) and has different underlying types in Windows and Linux. 
Use of non-portable types in EDL would lead to incorrect behavior.

jhand2: This doesn't change what type is used. If there is a bug now there was a bug before.

The reason this works is that on both Windows and Linux for x64 arch, time_t is defined as a 64 bit type.

common/edl/poll.edl
jhand2: These are existing types. If we want to change the names I would rather do it in a different PR.

common/edl/time.edl
radhikaj: @anakrish please open an issue and we can fix it in another Pr

common/edl/time.edl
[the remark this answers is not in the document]
jhand2: I'm going to leave this because time_t isn't considered a builtin type for EDL so getting time_t for free is just luck and could change with the implementation

PR #2782, opened 2020-04-02
oesign: fix a gcc9 build error

ipuustin:
With gcc9, build fails because `keyfile` argument can theoretically be NULL.  This is possibly because the value is checked for NULL later in `_load_pem_file()`, even though I can't see any call paths which lead to `keyfile` to actually being NULL at this point.

    /home/build/openenclave/tools/oesign/main.c: In function ‘oesign’:
    /home/build/openenclave/tools/oesign/main.c:664:13: error: ‘%s’ directive argument is null [-Werror=format-overflow=]
      664 |             Err("Failed to load file: %s", keyfile);
          |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Fix the build error by checking the value for NULL before using it.

### DIFF
--- a/tools/oesign/main.c
+++ b/tools/oesign/main.c
@@ -661,7 +661,7 @@ int oesign(
         /* Load private key into memory */
         if (_load_pem_file(keyfile, &pem_data, &pem_size) != 0)
         {
-            Err("Failed to load file: %s", keyfile);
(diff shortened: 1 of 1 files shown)


anakrish:
Thanks for fixing this!

anakrish:
bors try

PR #2784, opened 2020-04-02
Refactor SGX EDL files to be organized by feature groups

jhand2:
Refactor SGX-specific EDL file into smaller EDL files organized by feature.

In order to make attestation.edl publishable the sgxtypes.h file is also published with the SDK.

Note: I know this PR looks big but the majority of changes are just `#include <openenclave/internal/sgxtypes.h>` -> `#include <openenclave/bits/sgx/sgxtypes.h>`


common/sgx/edl/debug.edl
```diff
@@ -0,0 +1,36 @@
...
+
+/*
+**==============================================================================
+**
+** sgx/debug.edl:
+**
+**     Internal ECALLs/OCALLs to be used by liboehost/liboecore for operations
+**     related to debugging.
+**
+**     Note that this functionality is not inherently specific to SGX, it is
```
dthaler, line 12: Since this file is not inherently specific to SGX according to this line, should it be placed in a non-sgx-specific directory?

common/sgx/edl/thread.edl
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT License.
+
+/*
+**==============================================================================
+**
+** sgx/thread.edl:
+**
+**     Internal OCALLs to be used by liboehost/liboecore for thread operations.
+**
+**     Note that this functionality is not inherently specific to SGX, it is
```
dthaler, line 11: Should the EDL file be in a non-SGX-specific directory then?  (Analogy is that we have other header files with APIs that currently return "not implemented" on optee, this seems like it should be in the same category)
If this is punted to a future PR, then I'd recommend filing a GitHub issue, and optionally referencing that GitHub issue # in the comments in the edl.

common/sgx/edl/cpu.edl
```diff
@@ -0,0 +1,23 @@
...
+**     which are specific to the x86_64 architecture.
+**
+**==============================================================================
+*/
+
+enclave
+{
+    untrusted
+    {
+        oe_result_t oe_get_cpuid_table_ocall(
```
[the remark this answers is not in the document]
jhand2: CPUID is an x86 specific instruction but isn't necessarily SGX-specific. In theory a different TEE which also supports this instruction could use this ocall, its just that right now SGX is the only TEE we support for that arch.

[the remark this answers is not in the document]
jhand2: So this probably puts it into a similar category and debug.edl and thread.edl.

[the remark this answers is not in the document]
dthaler: Should the filename should be changed to be x86 specific then?

common/sgx/edl/attestation.edl
```diff
@@ -0,0 +1,84 @@
...
+enclave
+{
+    include "openenclave/bits/types.h"
+
+    // Most types in sgxtypes.h are packed to ensure the memory layout is
+    // the same as data retrieved from the CPU. Since EDL does not support
+    // packed structures, these types cannot be defined in EDL. They do not
+    // need the benefit of deep copy because they are all laid out in
+    // contiguous memory.
+    include "openenclave/bits/sgx/sgxtypes.h"
```
anakrish, line 24: Suggestion:
We have used opaque types (for lack of a better word) to prevent type information from leaking to code that does not need it. For example,`oe_cert_t` is defined to have enough space to store the certificate  of different crypto libraries.
https://github.com/openenclave/openenclave/blob/344cc4401bbc0abc85994ed16455a3f73516788a/include/openenclave/internal/crypto/cert.h#L17-L21
You can use `oe_cert_t` in code without knowing whether OpenSSLl or mbedTLS is used underneath.

Similarly the following type would avoid `sgxtypes.h` from having to be installed:
```
typedef _oe_sgx_report {
    uint64_t impl[64]; // 64 or whatever the accurate count is
} oe_sgx_report_t;
```

jhand2: I understand the sentiment, but these are structures that are defined by the SGX spec, not by us, so they won't change. In the Intel SDK these are published structures.

common/sgx/edl/attestation.edl
```diff
@@ -0,0 +1,84 @@
...
+    {
+        public oe_result_t oe_get_sgx_report_ecall(
+            [in, size=opt_params_size] const void* opt_params,
+            size_t opt_params_size,
+            [out] sgx_report_t* report);
+    };
+
+    untrusted
+    {
+        oe_result_t oe_get_qetarget_info_ocall(
```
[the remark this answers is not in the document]
jhand2: I agree that we should do this but I would rather do it in a separate PR (disambiguate all the SGX EDL function names) because it will touch a lot of code, some of which is in flux.

This PR doesn't actually publish the EDL files yet (just prepares them to be published) so we still have time to change before they're published.

common/sgx/edl/attestation.edl
anakrish: The rationale for the suggestion was not to safeguard against change, but to avoid publishing sgxtypes.h.
But I agree, it is not that important.

PR #2787, opened 2020-04-02
Fix bug with the CXX argument in add_enclave

### LINKED ISSUE #2786: CXX is always TRUE in add_enclave_sgx & add_enclave_optee cmake functions
`CXX` is an option to the `add_enclave` cmake macro (later forward to  `add_enclave_sgx` and `add_enclave_optee`) that is used to selectively link the enclave against `oelibcxx` library.
However, the current behavior is that `CXX` is always `TRUE` in  `add_enclave_sgx` and `add_enclave_optee` regardless it was specified in `add_enclave`.

 The bug is caused by the incorrect use of cmake argument parsing. More specially, `add_enclave`  parses `CXX` as an option, which has the following behavior:
- When `CXX` is specified (e.g., `add_enclave(... CXX ...)`), the value will be `TRUE`.
- When `CXX` is not specified, the value will be `FALSE`.

However, when `add_enclave` forwards the argument to `add_enclave_sgx|optee`, it uses the following code:
https://github.com/openenclave/openenclave/blob/344cc4401bbc0abc85994ed16455a3f73516788a/cmake/add_enclave.cmake#L59
And unfortunately, `add_enclave_sgx|optee` also parses CXX as an option.
https://github.com/openenclave/openenclave/blob/344cc4401bbc0abc85994ed16455a3f73516788a/cmake/add_enclave.cmake#L124
This means that `CXX` is always specified and therefore the value is always `TRUE`. The result is the following check is always passed.
https://github.com/openenclave/openenclave/blob/344cc4401bbc0abc85994ed16455a3f73516788a/cmake/add_enclave.cmake#L153

mingweishih:
This PR fixes the `CXX` argument in `add_enclave`, which was always `TRUE` after the argument was forwarded to `add_enclave_sgx` and `add_enclave_optee` (see more detail in #2786 ).

The fix is treating `CXX` as oneValueArg instead of option so that the argument can be correctly forwarded.

Fixes #2786 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fix #2607 so that libmbedcrypto now includes mbedtls_hkdf().
+- `CXX` is always `TRUE` in `add_enclave_sgx()` and `add_enclave_optee()`. Fixes #2786.
 
 ### Removed
 - Removed oe-gdb script which has been deprecated since v0.6. Use oegdb instead.
--- a/cmake/add_enclave.cmake
+++ b/cmake/add_enclave.cmake
@@ -9,6 +9,7 @@
 #  add_enclave(<TARGET target>
 #              [<UUID uuid>]
 #              [CXX]
+#              [ADD_LVI_MITIGATION ON/OFF]
 #              <SOURCES sources>
 #              [<CONFIG config>]
 #              [<KEY key>])
@@ -121,11 +122,10 @@ function(sign_enclave_sgx)
 endfunction()
 
 function(add_enclave_sgx)
-  set(options CXX)
-  set(oneValueArgs TARGET CONFIG KEY SIGNING_ENGINE ENGINE_LOAD_PATH ENGINE_KEY_ID ADD_LVI_MITIGATION)
+  set(oneValueArgs TARGET CONFIG KEY SIGNING_ENGINE ENGINE_LOAD_PATH ENGINE_KEY_ID CXX ADD_LVI_MITIGATION)
   set(multiValueArgs SOURCES)
   cmake_parse_arguments(ENCLAVE
-    "${options}"
+    ""
     "${oneValueArgs}"
     "${multiValueArgs}"
     ${ARGN})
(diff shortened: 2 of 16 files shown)


mingweishih:
> LGTM. If it a breaking change, accordingly update the change log.

After addressing Jordan's concern, there is no API changes with this PR.

jhand2:
@mingweishih I think we should call it out in the changelog. If someone's enclave needed the CXX runtime but forgot to add the flag, the old behaviour was that they got the runtime anyway, right?

mingweishih:
> @mingweishih I think we should call it out in the changelog. If someone's enclave needed the CXX runtime but forgot to add the flag, the old behaviour was that they got the runtime anyway, right?

Ah, good point. I'll update the changelog.

kuanggang-git:

mingweishih:
@BRMcLaren can you take a look? I didn't see errors in the log but it said build failed.

jhand2:
It still might be valuable for Brett to give any insight into why this might happen but anecdotally I see it happen from time to time. Rerunning.

bors r+

BRMcLaren:
@jhand2 

> It still might be valuable for Brett to give any insight into why this might happen but anecdotally I see it happen from time to time. Rerunning.

Would it help if I open an issue regarding CI failures in general? A lot of them are only heard about in passing, but if there is anything I can do to help, happy to.

BRMcLaren:
@mingweishih That.. looked like a graphical bug, and seems wrong. Let's see what happens with @jhand2

jhand2:
> @jhand2
> 
> > It still might be valuable for Brett to give any insight into why this might happen but anecdotally I see it happen from time to time. Rerunning.
> 
> Would it help if I open an issue regarding CI failures in general? A lot of them are only heard about in passing, but if there is anything I can do to help, happy to.

This is a particular issue I see recurring (maybe a handful of times a week). Either a couple stages will be marked as red even though they have no errors or bors will mark a run as failed even though all the stages in Jenkins are green. It isn't super disruptive but I do agree it would be good to track it with an issue. Then when I remember I can try to paste links when I see it happen

BRMcLaren:
@jhand2 thanks for the information, let's see if this is effective: https://github.com/openenclave/openenclave/issues/2819 

Please bear in mind, some issues the ci team might not be aware of.

PR #2789, opened 2020-04-03
Remove epoll.edl dependance on internal headers

jhand2:
Move the structures needed by by epoll.edl from an internal header to the syscall_types.h bits header. They are based on a public libc structures and will not change.

The epoll_event struct must be defined in C and not EDL because it is a packed structure.

### DIFF
--- a/include/openenclave/bits/edl/syscall_types.h
+++ b/include/openenclave/bits/edl/syscall_types.h
@@ -4,6 +4,9 @@
 #ifndef _OE_EDL_SYSCALL_TYPES_H
 #define _OE_EDL_SYSCALL_TYPES_H
 
+#include <openenclave/bits/defs.h>
+#include <openenclave/bits/types.h>
+
 /* DISCLAIMER:
  * This header is published with no guarantees of stability and is not part
  * of the Open Enclave public API surface. It is only intended to be used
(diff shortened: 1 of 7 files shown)


include/openenclave/bits/edl/syscall_types.h
```diff
@@ -4,11 +4,39 @@
...
+#include <openenclave/bits/defs.h>
+
 /* DISCLAIMER:
  * This header is published with no guarantees of stability and is not part
  * of the Open Enclave public API surface. It is only intended to be used
  * internally by the OE runtime. */
 
 typedef int64_t oe_host_fd_t;
 
+typedef union _oe_epoll_data_t {
```
mingweishih, line 16: Just found that we use this format only for union (although I personally like this better).

jhand2: I think some of these syscall types are intended to match the actual syscall struct name. For example `oe_epoll_event` matches the linux structure `epoll_event`.

But I do agree the _t format is preferrable. POSIX has some weird things because they don't use a type suffix. For example `stat` is the name of both a function and a struct.

include/openenclave/bits/edl/syscall_types.h
```diff
@@ -4,11 +4,39 @@
...
+    void* ptr;
+    int fd;
+    uint32_t u32;
+    uint64_t u64;
+} oe_epoll_data_t;
+
+/*
+ * The epoll events are packed in x86_64 but not on other architectures. This is
+ * probably because the x86 has no problem with unaligned accesses where as
```
dthaler, line 25:
```suggestion
 * probably because the x86 has no problem with unaligned accesses whereas
```

include/openenclave/bits/edl/syscall_types.h
mingweishih: I'm actually refer to
```
union _oe_epoll_data_t {
...
};
```
instead of
```
union _oe_epoll_data_t 
{
...
};
```

include/openenclave/bits/edl/syscall_types.h
```diff
@@ -4,11 +4,39 @@
...
+ * unaligned accesses are usually very inefficient, even if they are allowed, in
+ * other architectures. We maintain consistency with MUSL's and the linux
+ * kernel's handling on that architecture.
+ */
+#if !defined(__aarch64__)
+OE_PACK_BEGIN
+#endif
+struct oe_epoll_event
```
anakrish, line 33: Do these structs need to be packed at the boundary level? One could use unpacked EDL defined structs to marshal between host and enclave and transfer contents to packed structs only when needed.

jhand2: You're correct, they don't need to be packed at the boundary level.

For epoll in particular, whenever multiple events are passed, they're kept in contiguous memory so they can easily be memcpy'd all at the same time. The code to do the marshalling/unmarshalling is really ugly if you have to do it manually and it doesn't really provide any benefit in this case.

jhand2:
Bors run was cancelled

bors r+

include/openenclave/bits/edl/syscall_types.h
jhand2: oh, I didn't even notice that. We should probably fix that in clang-format :smile:

jhand2:
Fixed an issue with a couple missing includes

bors r+

Issue #2049, opened 2019-07-22
Update oe_get_revocation_info() to preallocate buffers to avoid double calls.


mikbras:
Consider updating **oe_get_revocation_info()** to pre-allocate reasonable buffer sizes to avoid double OCALL (that return **OE_BUFFER_TOO_SMALL** on first call).

yentsanglee:
Current revocation info size is 4402 bytes on coffeelake 6-core VM.

yentsanglee:
These warning are also annoying when running oecertdump tool:

2020-03-03T19:11:02.443703Z [(H)ERROR] tid(0x7f44999ca740) | :OE_BUFFER_TOO_SMALL [../host/sgx/report.c:_get_remote_report:113]
2020-03-03T19:11:02.443720Z [(H)ERROR] tid(0x7f44999ca740) | :OE_BUFFER_TOO_SMALL [../host/sgx/report.c:_oe_get_report_internal:174]

yentsanglee:
This not only makes double o-calls, it also make double PCK Cache service calls for PCK crl, qeid, and tcbinfo. Quite expensive.

yentsanglee:
Fixed.

PR #2790, opened 2020-04-03
Pre-allocating memory for the quote-verification-collateral arguments

shruti25ratnam:
Issue: Update oe_get_revocation_info() to preallocate buffers to avoid double calls. #2049


enclave/sgx/revocationinfo.c
```diff
@@ -25,6 +32,12 @@ oe_result_t oe_get_revocation_info(oe_get_revocation_info_args_t* args)
     if (!args)
         OE_RAISE(OE_INVALID_PARAMETER);
 
+    /*Allocate a default revocation-info-arg size*/
+    if(sizeof(out) < DEFAULT_ARGS_SIZE)
+    {
+        out = *((oe_get_revocation_info_args_t *)oe_malloc(DEFAULT_ARGS_SIZE));
```
jhand2, line 38: I don't follow. Why does changing the size of out change anything? The fields of out are passed to `oe_get_revocation_info_ocall` individually, the whole structure isn't used anywhere.

Isn't the issue that the individual buffers (such as `tcb_info` and `tcb_issuer_chain`) are null the first time around and so it takes an ocall just to get the sizes. So if they had preallocated buffers for each field then it could be done in one ocall instead of two.

shruti25ratnam: I agree. I'll make these changes. Thanks you.

kuanggang-git:

enclave/sgx/revocationinfo.c
shruti25ratnam: This file, enclave/sgx/revocationinfo.c will be removed from this PR.

common/sgx/collateral.h
```diff
@@ -66,6 +66,16 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
...
     oe_get_sgx_quote_verification_collateral_args_t* args);
 
+/**
+ * Pre-allocate memory for the resources that get passed into the 
+ * oe_get_quote_verification_collateral_ocall. To avoid extra passes needed
+ * to take care of OE_BUFFER_TOO_SMALL failures.
+ *
+ * @param[in] args The quote verification collateral.
+ */
+void oe_prealloc_quote_verification_collateral_args(
```
yentsanglee, line 76: >oe_prealloc_quote_verification_collateral_args [](start = 5, length = 46)

This function can be local. No need to declare in header file.

enclave/sgx/collateralinfo.c
```diff
@@ -169,6 +170,123 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
...
+oe_get_sgx_quote_verification_collateral_args_t* buf)
+{
+    /*These numbers are just estimates of buffer sizes*/
+    size_t prealloc_tcb_info_size = 5000;
+    size_t prealloc_tcb_info_issuer_chain_size = 3000;
+    size_t prealloc_pck_crl_size = 600;
+    size_t prealloc_root_ca_crl_size = 600;
+    size_t prealloc_pck_crl_issuer_chain_size = 3000;
+    size_t prealloc_qe_identity_issuer_chain_size = 3000;
+    size_t prealloc_qe_identity_size = 700;
```
yentsanglee, line 183: If these values are too big, we waste memory; if too small, we double call. According to current PCK cache, actual sizes are 2125, 1893, 460, 448, 1909, 1893, and 498. Your pre-set values are reasonable. However, the sizes differ by time and hardware. It's hard to get them right. I suggest we make the size values static (no pre-set) and remember the retrieved sizes. It pretty much guarantees we only double call once per execution.

shruti25ratnam:
Moved this PR here:  https://github.com/openenclave/openenclave/pull/2969
There was a need to change the downstream branch due to conflicts.

PR #2794, opened 2020-04-06
Remove td_t from sgxtypes.h

jhand2:
sgxtypes.h contains structures defined by the SGX specification. td_t is a thread data structure defined by OE and is an OE implementation detail, not an SGX one. Move td_t and related structures and functions to their own internal header.


include/openenclave/internal/sgx/td.h
```diff
@@ -0,0 +1,153 @@
...
+} td_t;
+OE_PACK_END
+
+OE_CHECK_SIZE(sizeof(td_t), 4096);
+OE_STATIC_ASSERT(
+    OE_THREAD_SPECIFIC_DATA_SIZE ==
+    sizeof(td_t) - OE_OFFSETOF(td_t, thread_specific_data));
+
+/* Get the thread data object for the current thread */
+td_t* oe_get_td(void);
```
dthaler, line 149: Question: the API oe_get_id() doesn't have sgx in the name.  Since this is under sgx/td.h do we think other TEEs in the future would need TD?  If so, should td_t be called sgx_td_t and oe_get_td be renaed to oe_get_sgx_td to avoid naming conflicts?

jhand2: td_t as it stands today is at least x86_64 specific (it saves rsp, rbp, etc.). @anakrish do you see any issue with changing the naming to be sgx-specific?

anakrish: I don't see any issue with renaming the function. 

However,  I do think that having platform specific prefixes first needs to be made a development guideline if it is already not so.

include/openenclave/internal/sgx/td.h
```diff
@@ -0,0 +1,153 @@
...
+**     the FS segment register (%fs:0) which contains
+**     oe_thread_data_t.self_addr.
+**
+**==============================================================================
+*/
+
+typedef struct _oe_thread_data oe_thread_data_t;
+
+/* Note: unused fields have a "__" prefix */
+struct _oe_thread_data
```
anakrish, line 30: Though this structure is named _oe_thread_data it, most if not all of its fields are mandated by the x64 Linux ABI.
That makes this an interesting structure; mostly ABI defined with couple of OE fields.

jhand2:
> Looks good to me.
> However, there is already a enclave/core/sgx/td.h. Can these structures be put there instead of a new header file?

td_t is also referenced in other places than oecore
* on the host for simulation mode
* in oelibc for pthread
* in libcxxrt for exception handling

I think it needs to stay be in a separate header

include/openenclave/internal/sgx/td.h
jhand2: Good point. I don't really know the answer so I opened https://github.com/openenclave/openenclave/issues/2796 to discuss

kuanggang-git:

jhand2:
@anakrish This pr now also changes `oe_get_td()` -> `oe_sgx_get_td()` and `td_t` -> `oe_sgx_td_t`.

Please confirm that this is ok. Thanks!

CodeMonkeyLeet:
> This pr now also changes `oe_get_td()` -> `oe_sgx_get_td()` and `td_t` -> `oe_sgx_td_t`.

Nonblocking observation: the description of this PR was to refactor out `td_t` from sgxtypes.h because it is OE-generic and not SGX-specific, but the naming change marks the type as specifically SGX-specific now ... can you clarify what the intention of this set of changes is? It feels like just rearranging furniture.

jhand2:
> > This pr now also changes `oe_get_td()` -> `oe_sgx_get_td()` and `td_t` -> `oe_sgx_td_t`.
> 
> Nonblocking observation: the description of this PR was to refactor out `td_t` from sgxtypes.h because it is OE-generic and not SGX-specific, but the naming change marks the type as specifically SGX-specific now ... can you clarify what the intention of this set of changes is? It feels like just rearranging furniture.

The reason I opened the PR is that sgxtypes.h was published under the pretense that all the structures in it are defined by the SGX spec and OE will not change them. But td_t was in that file and is not defined by SGX/Intel.

Now internally to OE, td_t is SGX (or at least x86) specific. It has a lot of fields that only make sense for SGX. So while it is an implementation detail of OE, not SGX, it is still only useful for SGX.

PR #2799, opened 2020-04-07
Remove accidental dependencies on libc headers from oecore

jhand2:
There are numerous places in oecore where standard C headers are included as system headers (i.e. #include <stdlib.h>) when the intention is really to use the corresponding corelibc header. In those instances explicitly include the corelibc header.

In instances in which standard C names are needed by 3rd party code, include those names only when compiling those files rather than making it the default for all files in oecore.

This PR is a revised version of #2767 with far fewer changes

### DIFF
--- a/enclave/core/malloc.c
+++ b/enclave/core/malloc.c
@@ -2,14 +2,13 @@
 // Licensed under the MIT License.
 
 /* The use of dlmalloc/malloc.c below requires stdc names from these headers.
- * Ensure that all corelibc headers are included with std names before
- * including other headers. */
-#define OE_NEED_STDC_NAMES
+ * OE_NEED_STDC_NAMES is set by enclave/core/CMakeLists.txt for this file
+ * to make the following corelibc headers include corresponding stdc names
+ * for types and functions. */
 #include <openenclave/corelibc/errno.h> // For errno & error defs
 #include <openenclave/corelibc/sched.h> // For sched_yield
 #include <openenclave/corelibc/stdio.h>
 #include <openenclave/corelibc/string.h>
-#undef OE_NEED_STDC_NAMES
 
 #include <openenclave/enclave.h>
 #include <openenclave/internal/fault.h>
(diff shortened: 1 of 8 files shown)


enclave/core/malloc.c
```diff
@@ -2,14 +2,11 @@
 // Licensed under the MIT License.
 
 /* The use of dlmalloc/malloc.c below requires stdc names from these headers.
- * Ensure that all corelibc headers are included with std names before
- * including other headers. */
-#define OE_NEED_STDC_NAMES
+ * This define is set by enclave/core/CMakeLists.txt */
```
dthaler, line 5: What is "This define" referring to?

syscall/poll.c
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+#include <openenclave/corelibc/stdlib.h>
```
dthaler, line 4: Duplicate of line 7

enclave/core/malloc.c
jhand2: Oh whoops. I removed the define and now the comment needs context

jhand2:
> It would be nice to have an automated check for this, maybe a script that scans includes î files under enclave/core for anything that exists under corelibc/ and is included un-prefixed?

The main reason this worked is that `include/openenclave/corelibc` was in the include path for every file in corelibc. I removed that so now if you try to reference a libc header without the prefix, the file will fail to compile. The only exceptions are for files which are part of 3rd party code (dlmalloc for example) and expect a libc headers to be available. Now that I think of it, this could impact snmalloc. I'll look into it.

jhand2:
> Before merging confirm whether `prng/rand.c` needs `-Wno-conversion`

Confirmed that line 14 of rand.c returns seed (which is a unit64_t) as an int. It is intentional and done safely but the cast is still implicit.

jhand2:
bors try

jhand2:
If this doesn't work I'll go ahead an get my OP-TEE build environment working again

bors try

jhand2:
bors try

PR #2801, opened 2020-04-07
Attestation: concepts and assumptions behind the OE SDK API

shnwc:
Add a design document with explanation of concepts and assumptions behind the design of the OE SDK Attestation API, as well as the design of SGX plugins


docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+    These security properties are required for attestation and secure
+    communication.
+  - They define which evidence formats attesters accept, and configure the
+  list to the verifiers.
+    - Every acceptable evidence format is TEE specific.
+- OE SDK provides means for the attesters and verifiers to be built and
+configured for specified TEEs and platforms, and to generate or verify
+the evidence formats according to their defined security policies.
+  - The OE SDK attestation API enables the attesters and verifiers to discover
+  and register supported plugins that are built, configured, and working
```
dthaler, line 85: That might be true for the current experimental API, but I would rather see a model where attesters and verifiers do not need to call an API to register supported plugins, but registration is handled via a link-time operation and/or a sign-time metadata operation.   My goal is to lower the bar to write an enclave application, or port an application to use an enclave.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+  properly. For a specific attester or verifier instance, the supported list
+  of evidence is not required to perfectly match the list defined in the
```
dthaler, line 87: "supported list of evidence" is confusing.  Do you mean list of evidence formats?  List of evidence plugins?  Something else?

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+  security policies.
+
+# Options for Enclave-side Plugin Library Initialization
+
+As described in the design document
+[Attestation: OE SDK Integration with Intel® SGX SDK quote-ex Library for Generation of Evidence in New Formats](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/SGX_QuoteEx_Integration.md),
+enclave-side plugin library initialization of attester plugins can be
+triggered by application enclave call of the OE SDK API
+`oe_get_attester_plugins()`.
```
dthaler, line 96: This is the API I would prefer to remove from public API surface area, if possible, or at minimum only be needed in extremely advanced cases and not needed in any common case.   (I'm fine for this to be an internal API invoked by the oehost library itself.)   I believe Option 3 below could be aligned with this direction.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+
+    The design document
+    [OpenEnclave Modules](https://github.com/jhand2/openenclave/blob/global_module_initialization/docs/DesignDocs/openenclave_modules.md)
+    describes how an enclave-side module can register its initialization
+    function for invocation on first entry to the enclave (upon enclave load).
+
+    With this option, the implementation of the OE SDK attestation framework
+    provides a hook for enclave-side plugin libraries to register their
+    initialization functions upon enclave load.
+    To support ordering in the initialization of multiple plugin libraries,
```
dthaler, line 140: Is ordering actually a requirement, or is this not needed?   (I ask because it wasn't mentioned anywhere earlier in the document, so if it is a requirement, please add a discussion of it earlier in the doc.)

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+verifier plugins.
+
+1. Every enclave-side plugin library implements the same OE SDK API
+`oe_get_attester_plugins()`.
+
+    With this option, a build of an application enclave project can only link
+    with a single attestation plugin library, to avoid function name collision.
+
+2. Compile-time macro to convert from TEE-agnostic API to TEE-specific
+implementation.
```
yentsanglee, line 114: This option requires the plugin author to define the macro and may not be scalable for unknown plugins.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+3. Every enclave-side plugin library registers its initialization function
+with the OE SDK framework.
+
+    The design document
+    [OpenEnclave Modules](https://github.com/jhand2/openenclave/blob/global_module_initialization/docs/DesignDocs/openenclave_modules.md)
+    describes how an enclave-side module can register its initialization
+    function for invocation on first entry to the enclave (upon enclave load).
+
+    With this option, the implementation of the OE SDK attestation framework
+    provides a hook for enclave-side plugin libraries to register their
```
yentsanglee, line 138: >enclave-side plugin libraries [](start = 24, length = 29)

The same initialization should also be supported on the host-side plugin libraries. We have to recommend a universal way (for all platforms) for libraries to initialize on load. Typical ways to do initialization on load are 1) global variables 2) .init_array. The order of such initialization is critical. If the plugin libraries try to register before OE SDK is ready, it's a potential issue.

I'd suggest OE SDK team to finalize the design in [https://github.com/jhand2/openenclave/blob/global_module_initialization/docs/DesignDocs/openenclave_modules.md](https://github.com/jhand2/openenclave/blob/global_module_initialization/docs/DesignDocs/openenclave_modules.md) and prototype this on Linux and Windows to ensure this is a viable option.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+    properties. They are created and maintained by the attester and not
+    received from a verifier
+    - Examples include a public key or a nonce generated by an attester
+    - These claims are passed to verifiers with integrity protection (as part
+    of the evidence), so verifiers can use the claims to learn the identity
+    and properties for the attester and for secure communication with the
+    latter
+  - They are always TEE agnostic
+  - Signed and included in evidence as clear text by an attester plugin,
+  and verified and reproduced by a verifier plugin, to ensure integrity
```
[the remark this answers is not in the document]
yentsanglee: Maybe "verified and restored"?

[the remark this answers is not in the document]
jhand2: My understanding of this passage is that the verifier will emulate whatever verification done in hardware (in SGX this could be recomputing the MRENCLAVE value or verifying the signature with a public key) to ensure the computed value matches the value reported by the attester.

I don't know how to say that in fewer words. Maybe that level of detail isn't even relevant for this document though. A verifier can verify evidence however it sees fit. Maybe that is computing a hash, maybe that is verifying a signature. The API probably doesn't need to be concerned with the how.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+usage, and the SGX plugin design.
+
+# Semantics of Custom Claims and Optional Parameters
+
+To help clarify what goes into custom claims and what goes to optional
+parameters in the `oe_get_evidence()` API function, it would be good to
+articulate the semantics of these two fields.
+
+- Custom claims have well-defined semantics:
+  - Can be any name-value pair
```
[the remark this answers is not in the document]
shnwc: Agreed that in general claims can have names / IDs of different formats. The "custom claims" here are specific for the `custom_claims` parameter of API function`oe_get_evidence()`. I'll add clarification around it.

To make it clear between the notes for the TEE-agnostic API and for the SGX plugin design, I'll organize the note into two chapters.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+  - Only meaningful to ISV attester / verifier applications
+    - Typically, custom claims are part of an attester instance identity or
+    properties. They are created and maintained by the attester and not
+    received from a verifier
+    - Examples include a public key or a nonce generated by an attester
```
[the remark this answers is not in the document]
shnwc: Good catch! nonce is generated by a verifier and can be used to ensure evidence freshness. It is still a custom claim. I'll clarify in the document.

docs/DesignDocs/NotesOnAttestationAPI.md
jhand2: I have the global module initialization work planned for v0.10 release. I will include you and @shnwc in the reviews.

Note that my current design is only for enclave-side initialization. Host-side initialization for Windows is more complicated.

kuanggang-git:

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+[CustomAttestation_V3.md](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/CustomAttestation_V3.md),
+as well as the design of SGX plugins as described in document
+[Attestation: OE SDK Integration with Intel® SGX SDK quote-ex Library for Generation of Evidence in New Formats](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/SGX_QuoteEx_Integration.md).
+
+Hopefully these notes can help the readers' understanding of the API and its
+usage, and the SGX plugin design.
+
+# Semantics of Custom Claims and Optional Parameters
+
+To help clarify what goes into custom claims and what goes to optional
```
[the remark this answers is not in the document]
shnwc: I see both the custom claims and optional parameters in general fit the term "Claims" in the RATS document, but in the specific context of the OE SDK API `oe_get_evidence()`, they have distinctively different semantics. This document tries to articulate those details. I'll update the notes for an clearer articulation.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+    received from a verifier
+    - Examples include a public key or a nonce generated by an attester
+    - These claims are passed to verifiers with integrity protection (as part
+    of the evidence), so verifiers can use the claims to learn the identity
+    and properties for the attester and for secure communication with the
+    latter
+  - They are always TEE agnostic
+  - Signed and included in evidence as clear text by an attester plugin,
+  and verified and reproduced by a verifier plugin, to ensure integrity
+  protection in transmission from an attester application to an verifier
```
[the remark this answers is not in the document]
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+  application
+    - Secrets shall never be included as a custom claim. Instead a hash of
+    a secret can be included by an attester as a claim in its evidence,
+    and passed to an verifier with integrity protection. The secret itself
+    has to be passed from the attester to the verifier with confidentiality
+    protection. Integrity of the secret can be verified by the verifier
+    using the hash in the evidence.
+  - Custom claims are NOT inspected or interpreted by OE SDK / plugins
```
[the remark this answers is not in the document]
shnwc: The API functions `oe_get_evidence()` and `oe_verify_evidence()` dictate that a custom claim is a name-value pair, with both the name and the value fields being null-terminated byte strings.

While a custom claim fits the definition of "Claims" in the RATS paper, I do not see the requirement that all different possible name / ID formats need have to be supported.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+- Optional parameters semantics are the opposite of that of custom claims:
+  - They are TEE and evidence format specific
+  - They are interpreted and processed by OE SDK attester and verifier plugins
+    - Typically, optional parameters are provided by a verifier to an attester,
+    so the attester can generate the correct evidence that can be successfully
+    verified by the verifier
```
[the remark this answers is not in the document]
shnwc: Assuming that the nonce is generated by an verifier application, it fits as a custom claim.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+  - They are not interpreted nor processed by attester / verifier
+  application software
+
+For SGX local attestation, the SGX target info structure generated by an
+verifier enclave for an attester enclave is a perfect fit for the optional
```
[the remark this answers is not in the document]
shnwc: `oe_get_verifier_settings()`

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+parameters in the attester enclave's call to `oe_get_evidence()`:
+- For SGX local attestation, the SGX target info is generated by a verifier
+enclave and passed to an attester enclave
+- With the received target info, the attester enclave can generate an SGX
+report targeted for the verifier enclave
+- This SGX report can only be verified by the verifier enclave that provided
+by target info. Any other enclaves, either running on the local or a remote
```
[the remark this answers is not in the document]
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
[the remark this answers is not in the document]
shnwc: Thanks! Will clarify in in the notes.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+- The target info is SGX specific, and is not parsed nor interpreted
+by the TEE-agnostic application code. It is handled by the SGX local
+attestation plugins underneath the `oe_get_evidence()` and
+`oe_verify_evidence()` API in the attester and the verifier respectively.
+
+# Attester – Verifier Security Model and TEE Agnostic Design
+
+- OE SDK is designed for attester and verifier software to be TEE agnostic,
+so that the same source code can be built for and run on different platforms
+- But owners of attesters and verifiers define their security policies,
```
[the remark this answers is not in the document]
shnwc: The RATS "appraisal policies" is focused on how a relying party uses the attestation results regarding an attester generated by a verifier. I think the "security policies" has a broader scope. I notice a note in the RATS document "Compare /security policy/ in [RFC4949]". Is that a TODO item?

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+and the policies are not TEE agnostic. For example:
+  - The policies define which TEEs, platforms and configurations are trusted
+  to run attesters and verifiers.
+    - They also define security properties of these entities, e.g. keypairs
+    and certs, service provider ID / SPID for SGX EPID evidence, etc.
+    These security properties are required for attestation and secure
+    communication.
+  - They define which evidence formats attesters accept, and configure the
```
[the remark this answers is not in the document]
shnwc: Will edit to clarify.

While an attester application software does not parse any evidence, the set of all the installed attester plugins, as checked by API `oe_select_attester_format()`, covers all the formats against which the attester can generate evidence. In the description. This list of formats is what the attester accepts.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,160 @@
...
+  list to the verifiers.
+    - Every acceptable evidence format is TEE specific.
```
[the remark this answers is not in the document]
shnwc: Agreed "all" could be overstatement. Will clarify.

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Totally agreed, and I have listed initialization options in the "Options for Enclave-side Plugin Library Initialization" section.

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Will clarify.

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Agreed. In my [fork](https://github.com/shnwc/openenclave), simpler public API `oe_initialize_attester_plugins(void)` and `oe_initialize_verifier_plugins(void)` have been implemented, that discover and register all working plugins.

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Agreed it has scalability issue.

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: See @yentsanglee comment on ordering. This document does not try to decide whether that's a requirement or not. It only discusses a possible solution if ordering is needed.

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: @yentsanglee agreed that host-side support is needed. Will edit the doc to reflect that. 

@jhand2 what is the plan to address the Linux and Windows host-side initialization?

docs/DesignDocs/NotesOnAttestationAPI.md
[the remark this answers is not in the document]
shnwc: @dthaler @yentsanglee yes "verified and reproduced" means "verified and restored", restoring to the original form of a list of name-value pairs for processing by verifier application. I'll edit this bullet.

@jhand2 SGX quote verification is quite complex, involving verification of quote and the cert chain that signs it. Though it does not involve re-computing MRENCLAVE. As you mentioned, how the evidence is verified is not the concern of this passage (which focuses on the parameters of the API.

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: @yentsanglee Can the function `void register_plugin_function(func)` be implemented to initialize its `_lock` and `_registered_function_list` upon the first call by an plugin library?

yentsanglee: If register_plugin_function is called in reaction to another global variable initialization, I am not sure how to universally guarantee _lock and _registered_function_list are initialized. Any suggestion?

dthaler: One example implementation would be for the metadata (.conf file) to contain the list of modules at sign time.   Then global initialization code enumerates the list from the metadata, and invokes the initialization functions in the order of that list.   Then the app writer does not need to be involved (only the signer), and you get ordering if it is needed.

yentsanglee: From what I understand, the developers have to:
1. Create a .conf file in a pre-defined syntax.
2. Links to the compiled metadata (.conf file)

OE SDK has to:
1. Have a signtool that compiles the .conf file
2. Invoke initialization functions.

Plugin authors have to:
1. Export an initialization function and avoid naming conflict

I think this deserves a separate design. This sounds like a plugin model for generic purpose (not just attestation).

shnwc: @jhand2 how about including this proposed approach in your [OpenEnclave Modules](https://github.com/jhand2/openenclave/blob/global_module_initialization/docs/DesignDocs/openenclave_modules.md) document?

shnwc: @yentsanglee upon enclave load, how does the OE SDK framework map an initialization function from its name listed in the .conf file to the entry point address  for invocation?

jhand2: The main barrier in my original design for modules what that there was no way to tell the linker to keep the initialization functions since we link with `--gc-sections` and there is no reference to the function (it's called by address). I see the same barrier in this design on the enclave side.

Also does this proposal have a host-side solution?

I also have the same question as @shnwc and I don't have a great answer right now.

shnwc: @jhand2 Just committed a new version that contains host-side options.

yentsanglee: I am not sure if .conf or metadata would improve the development experience but it has to work for enclave and host, Windows and Linux.

I suggest we redesign a plugin generically for OE SDK while the additional attestation formats continue to develop without introducing (much) new public APIs.

shnwc: @yentsanglee before option 3 support is ready for both enclave and host sides, we can readily support option 1, with the limitation that only one plugin library can be linked (which is currently the case for SGX). I'll update the document to capture that.

Can you elaborate on  second paragraph "I suggest we redesign a plugin ..."? Since the plugin APIs are all experimental, I'd like to see a new version of the API definition published soon, so the  SGX plugins implementation in my fork can be upstreamed.

yentsanglee: Since there are multiple instances of plugin design in OE SDK (for different TEEs, logging, etc), I think we should start a new overall design to set the plugin story straight for all purposes. For the EPID and Quote-ex work, I suggest we continue the work in your fork and integrate with minimum API changes.

I am fine with option 1 but that's where the disagreement started.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,359 @@
...
+
+    C/C++ global constructors can be used for this purpose.
+    An explanation of them can be found in
+    [this document](https://wiki.osdev.org/Calling_Global_Constructors).
+
+    Note: to enable this option, the OE SDK host-side framework has to be
+    initialized with the needed locks and list structure, before plugin
+    libraries can register their initialization functions.
+
+## Discussion
```
CodeMonkeyLeet, line 330: Great discussion on the topic, there's a lot to think about here. I think it's a good idea to break out the discussion into more contained pieces if we think it will help us make progress on this, but my main struggle in trying to wrap my head around this is not having a clear understanding of what the desired output of the library initialization discussion is:

- The comments between @shnwc and @yentsanglee seem to suggest that this is necessary for unblocking implementation of the quote-ex library integration. I'm not clear what agreement needs to happen to allow that to move forwards based on what I've read here, but if that is the goal, I would suggest explicitly scoping the discussion to that question and iterating on that, possibly just as part of the next SIG Arch/Attestation meeting. FWIW, I don't think there are blockers to the that work since it seems mostly internal; given the experimental labelling of the evidence APIs, it sounds like we would just go back and fix those implementations as needed once consensus is achieved on the broader attestation plugin model questions.

- Is the goal to arrive at a design proposal for initialization of attestation plug-ins? If so, that probably deserves to be spun off into its own design proposal that can be reviewed and approved as such. As it stands it's a confusing shift in context from explaining the overall attestation plugin model, to offering options and opinions on a specific engineering question.

- From the comments, it looks like we're still a step before breaking out into an engineering design for initialization; I read the discussion as trying to answer the unresolved question around what the _experience_ of initializing an attestation plugin should be, with the different engineering options creating different constraints on what we would offer.

In the context of that last interpretation, I would suggest using the constraints of the scenario itself to shape the design. For example, I think the relationship between attesters and verifiers is inherently asymmetric and that's not something that's really being taken into account when we assert that it is desirable for a common mechanism to apply to both the host and enclave. For example, consider what the user matrix for the initialization options looks like:

|             | Attester  | Verifier  |
|------------:|:---------:|:---------:|
| **Host**    | N/A       | Host dev  |
| **Enclave** | Enc dev   | Enc dev   |

- The fact that evidence creation has to occur inside an enclave is one of those fundamental asymmetries. While we could support attester plugins unrelated to TEEs at the API level, OE SDK does not provide a host attester lib today because that would be tangential to its core goals.

- The document implicitly sets up the dichotomy of attester plugin initialization in enclaves and verifier plugin initialization in host apps, even though that ends up comparing options that look structurally similar but apply to different users and use cases. For example, option 3 is expressed as a preference for both host and enclave and both involve manifests, but who uses them and why isn't directly addressed.

  - **Enclave dev**: This is the developer that makes the core of trust decisions as it pertains to the resulting measurement of the final enclave image, whether through code in terms of what initialization functions get called, or build time choices such as what to link and what configuration metadata to include.

  - **Host app dev**: We usually consider this to be the same developer as the enclave dev because the current OE dev model presents enclave app development as spanning both the host and enclave components. This has several relevant distinctions in this discussion though:

    - In the verifier case, the app can be a distinct app from the attester, and it may be developed as a normal, non-TEE app. A convergent development experience with enclave apps may not be relevant to all verifier app developers.
  
    - The host app dev controls runtime or enclave load time decisions, and these in turn can be exposed directly to the app user. The initialization controls potentially need to consider the value of a non-developer use case.

- The implicit case of verifier plugins in enclaves is really what feels to me like the scenario that's driving the desire to have a consistent initialization mechanism; while I agree with the intuition that verifier plugins at least should be handled in a consistent way, I believe that thinking through the scenario as distinct from the other two is valuable in identifying challenges that are not obvious when thinking about each of the dimensions in isolation. For example, in the enclave security model, the host app dev may be considered adversarial to the enclave, and if we choose a model where some options are then exposed to the app user, we would need to re-evaluate what that means in the enclave version of that experience.

To elaborate on each of the 3 scenarios and its constraints or requirements:

### 1. Attester plugins in enclaves

- Only static linking is supported
  - The cost to implement dynamic loading in enclaves is expensive, so for the moment, the choice of plugins must be decided at least once at build time by the enclave developer.
  - If we want to minimize the friction of using a feature, it's useful to reduce the number of places a dev has to specify intention to use a feature. The enclave dev already has a link time choice to include plugin libraries, and the manifest solution doesn't provide immediate value to the enclave developer in that sense.
    - Note that the manifest option can potentially provide value to host dev or app user, but that's a different value proposition that we haven't evaluated in this discussion yet.
- Enclaves are linked as executables with `--gc-sections`
  - Unused symbols will be trimmed, which means workarounds to keep symbols are necessary and may increased developer friction
  - This is true even for manifest-based solution (i.e. option 3).
  - If a reason for wanting converged host/enclave enclaves is the expectation that it may reduce overall costs for implementation, that may not be the case for reasons such as this.
- Attester plugins are limited at runtime to the TEE used by the enclave instance.

  - This is a constraint in the problem we don't seem to use today: an enclave dev needs to choose the attester specify an evidence format UUID to `oe_get_evidence()`.
    - That's not a problem and we will need to continue to support this; for instance, there will be TEEs that support multiple evidence formats (e.g. EPID vs. ECDSA on DCAP) where it remains the enclave dev's responsibility to choose which format to use.
    - This does assume that the enclave developer already knows which UUIDs are associated with which plugins, plugin libraries and supported TEEs. In this scenario, the existing initialization mechanism is fine: the dev explicitly calls initialization on each attester plugin by UUID, which could fail with an ignorable error when running on a TEE not supported by the plugin.

  - What is not called out clearly in this doc is that the current set of revisions is intended to support additional use cases where a developer does not need to know specific evidence format UUIDs. It would be useful to spell out the intended scenario instead, for example:
  
    - Enclave dev can write a _binary-compatible_ enclave that runs on multiple TEEs, but doesn't care about evidence formats beyond that (similar to the intent of the existing `oe_get_report()` API).
      - In this case, it would be ideal if every attester plugin the enclave dev links gets initialized without having to know UUIDs.
        - Option 4 on the enclave-side addresses this aspect better than option 3, all other things being equal.
      - Although the runtime choice of attester plugin will usually be constrained by the runtime TEE, in the case where a TEE supports more than one format, the attester plugins still need a mechanism for selecting a *default* format. That, in turn, raises other questions, like who gets to choose the default?
        - The plugin library? For example, an SGX one could detect the presence of DCAP vs. AESM dependencies in the runtime environment.
        - The enclave dev? This might defeat the point of the enclave dev not having to understand specific formats.
        - The host dev/app user? This might suggest that we want a manifest-based solution that is provided at runtime and not  build time.
      - Answering that question can help determine the relative value of options 3 and 4, or if they're both supported for different reasons. It will also help drive scoping of other questions not covered by this document, such as whether local/remote attestation should be handled by a setting in the attestation plugin model or represented as separate plugins.

### 2. Verifier plugins in host apps

- Plugin initialization mechanism must span both Windows and Linux.
  - This is something pointed out in Jordan's OE modules proposal, where a mechanism like `.init_array` in GCC for Linux doesn't have an exact equivalent in MSVC on Windows.
  - Although the OE SDK hostverify library is distributed as a static lib, verifier plugins could easily be implemented as shared libraries (DLL/SO).
    - Consider that changing the form of the plugin library produces options that are easier to abstract across Windows & Linux. In this case, there are fairly equivalent ways for doing dynamic loading that have already been encapsulated in the oehost lib, and that can form a simpler foundation for a manifest-based solution like option 3.
    - Also consider that on the host side, a dev can already use any existing mechanism to initialize the plugin; any additional formalism that OE imposes an overhead that should be justified.
      - e.g. unlike option 4 for host-side which suggests that the dev use C++ global constructors to register the initialization functions, the plugin dev could use the constructor to initialize the plugin and then register the UUID with the OE runtime. If so, what benefit that imposing the formalism of registering the initialization functions provide over the alternative?

- The OE runtime has no intrinsic criteria with which to decide if a verifier plugin is applicable at initialization time.
  - A single host app can require verifier plugins for different TEEs at the same time, and unlike the enclave, there's no way to scope this by the active TEE platform. In fact, there may be verifier plugins that don't even have corresponding attester plugins i.e. the evidence format is not generated by TEEs or is created by TEEs not yet supported by OE SDK.
  - The `oe_verify_evidence()` API already scopes at runtime which verifier plugin to use based on the provided evidence's format UUID, so any additional control mechanisms are really a refinement. For example, the additional control could be for:
  - Finer granularity of control. A host dev may omit a plugin library by default, but not individual plugins. This could be achieved in other ways, for example, by requiring that there is a 1:1 relationship between plugin and plugin library.
  - Exposing control of which verifier plugins to use to the app user. For example, if a goal is to support the use case where the host dev builds in all the plugin verifiers, but the app user gets to choose which are used, then that makes a case for implementing a manifest-based initialization scheme where a default manifest can simply be replaced by a user without rebuilding the app.
  
### 3. Verifier plugins in enclaves

- Same enclave build constraints as for attester plugins: Only static linking is supported and enclaves are linked as executables with `--gc-sections`.

- Unlike verifiers on the host side, some verifiers in enclaves may also have dependencies on TEEs like the attester plugins:
  - This specifically applies to local attestation verifiers.
  - This suggests that even if we have identical mechanisms for host and enclave verifier plugins, there will be different lists of of verifier plugins to initialize on each side.
  - Otherwise, the same verifier selection considerations apply: something other than the currently running TEE should decide what verifier plugins are initialized.
    - The additional consideration here is that allowing a host dev or app user to specify the choice of verifier plugins crosses enclave trust boundaries.
    - For example, let's say that in a peer-to-peer enclave attestation model, something horrible exploit is discovered that makes ECDSA quotes less trusted than EPID quotes on SGX. If I allow a user to only initialize the ECDSA format SGX verifier in the enclave, _and_ if the protocol between peers is that the verifier gets to say which format gets used, then this is a security issue.

To illustrate, consider this strawman design; it is not a recommendation, just an example to show how the design can change depending on what's important.

*On the host side:*

- I want to allow the _app user_ to configure verifier initialization.
  - As a host dev, I don't mind if I have to call a special OE function to initialize the verifier plugins to do so.
- I want to have a consistent experience on Windows and Linux.

In this case, it might make sense to define a new `oe_initialize_verifier_plugins()` method that takes a manifest file which lists the plugin modules (.dll/.so) and the initialization functions to invoke on each.

*On the enclave side:*

- I want the enclave dev choice to be simplified to just linking the correct plugin libraries.
- I don't want the enclave dev to decide defaults or relative priorities of evidence formats.
- I don't want the host dev or app user to control what attesters or verifiers the enclave uses.

In this case, it might make sense for the experience to be different from the host, where both attesters and verifiers on the enclave side adhere to the guidelines for OE modules, and the enclave dev has to link enclaves using a linker script that preserves the `.init_array` section for all the functions. For default/priority selections, I delegate that to the either the plugin's own detection logic (e.g. is AESM running for EPID?), or the initialization priority specified within plugin libraries.

If instead I decide that allowing the host dev or app user to configure either the attesters or verifiers used by the enclave is desirable and not a security risk:

- I would still have the constraint on the enclave dev to use the linker script to make sure all the initialization functions are availble, but they would not all be automatically invoked.
- Instead, I would also need to add an additional control point for the host dev at the API level. For example, we could modify `oe_create_enclave()` to take a new `oe_enclave_setting_type_t` that points to a manifest file that either the host dev can provide or expose to the app user to provide. We could try to make the manifest format consistent with the host, but we really wouldn't need the plugin module name per initialization function, so even there natural differences will occur.

For the discussion around choice of initialization options, I would like the end result to be an agreement on the explicit set of criteria that explains the chosen option(s).

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,359 @@
...
+header structure.
+
+In the OE SDK attestation plugin design, a plugin is a subsystem that exposes
+a set of API entry points for evidence generation or verification, as defined
+in this [header file](https://github.com/openenclave/openenclave/blob/master/include/openenclave/attestation/plugin.h).
+A plugin can either be an attester plugin, or a verifier plugin.
+Each plugin is associated with a single evidence format UUID. On the other hand,
+in theory multiple plugins can exist in a configuration for the same format,
+though in practice this is unlikely since there is no clear benefit for the
+added complexity in implementation.
```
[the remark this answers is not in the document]
yentsanglee: Agree with Simon. Today Azure has multiple services to verify SGX quote - PCK Cache and MAA service. The verifier plugins for both will have the same format ID but it's unlikely an application will have both registered at the same time. The statements look good to me. Perhaps minor changes to address Simon's concern.

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Let's start with option 1 as experimental in short-term, while working out a longer-term option.

docs/DesignDocs/NotesOnAttestationAPI.md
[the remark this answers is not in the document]
shnwc: @CodeMonkeyLeet very good points on the semantics of UUID. I'll update the document to capture them. The RE is an interesting idea. It is somewhat similar to the QE (Quoting Enclave) we are using for remote attestation, which works even if the attester and verifier enclave reside on the same platform.

@yentsanglee just curious what is MAA service? How does the PCK Cache verify quote? the name indicates it would simply retrieve and cache PCK certs. Do these services use the OE SDK plugins for quote verification or use some other implementations?

This document is growing to become much bigger than I originally expected :-).

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,359 @@
...
+
+In an implementation of the OE SDK design, the set of supported plugins
```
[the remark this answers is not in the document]
shnwc: "logical unit" is a good term. I see a plugin more like a C++ object. Will update the doc to capture the ideas.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,359 @@
...
+are implemented by a set of enclave-side and host-side plugin libraries and
+their dependencies. These plugin libraries can either be static or shared
+libraries, as supported by the build environment. There is no one-on-one
+mapping between plugins and plugin libraries.
+A single plugin may be implemented by multiple plugin libraries. For example,
+an enclave-side attester plugin typically need at least one enclave-side
+plugin library and a host-side plugin library. On the other hand,
+a single set of plugin libraries can implement multiple plugins.
+
+Note: an enclave-side plugin library can also be called an
```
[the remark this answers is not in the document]
shnwc: Agreed.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,359 @@
...
+
+As an example, here is a possible configuration for support of SGX attester and
+verifier plugins for ECDSA and EPID evidence formats:
+- ECDSA and EPID attester plugins can be implemented by a single set of plugin
+libraries and their dependencies, composed of:
+  - One enclave-side SGX evidence generation plugin library
+  - One host-side SGX evidence generation plugin library
+  - Dependencies: Intel SGX SDK DCAP quote generation library SGX quote-ex
+  library, SGX Quote Provider Library (QPL), as well as their dependencies
+- ECDSA verifier plugin can be implemented by another set of plugin libraries
```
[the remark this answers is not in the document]
shnwc: EPID evidence verification is not supported by Intel SGX SDK, and thus not implemented as an OE SDK verifier plugin. In existing SGX solutions, EPID quotes are sent to Relying Parties, which forward them to an Intel Attestation Service (IAS) for verification. Relying Parties and IAS are not necessarily SGX platforms and do not use Intel SGX SDK.

The set of  examples is supposed to show how plugins and plugin libraries are not mapped one-on-one. Agreed it's unclear and not needed. I'll remove this paragraph and expand the last part of the previous paragraph with some more examples.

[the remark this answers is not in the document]
dthaler: Thanks for the explanation of how EPID evidence verification is done using a Background-Check model.   Is there any way to do EPID evidence verification in a Passport model topology?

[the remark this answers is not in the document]
shnwc: For EPID evidence verification, the IAS is the Verifier, In a Passport model,  the Attester would need to implement the functionality of the Relying Party in the existing solutions to establish a trust relationship with the IAS, to submit EPID quotes and to receive signed Attestation Results. I don't have next-level details, but if there is serious requirement from application developers, we can look into it.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,359 @@
...
+  - Dependencies: Intel SGX DCAP quote verification library, QPL, as well as
+  their dependencies.
+
+## Options for Enclave-side Plugin Library Initialization
+
+As described in the latest version (as of March 2020) of the design document
+[Attestation: OE SDK Integration with Intel® SGX SDK quote-ex Library for Generation of Evidence in New Formats](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/SGX_QuoteEx_Integration.md),
+enclave-side plugin library initialization of attester plugins can be
+triggered by application enclave call of the OE SDK API
+`oe_get_attester_plugins()`.
```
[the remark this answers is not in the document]
shnwc: I've updated the document to use the `oe_attester_initialize()` described in PR #2949, and will bring the SGX_QuoteEx_Integration.md up to date to use the API in PR #2949 with a new PR.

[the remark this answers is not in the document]
shnwc: The new PR is #2985

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: @CodeMonkeyLeet thank you for the detailed and insightful comments!

The outcome of PR #2949 has unblocked quote-ex library integration.

The section on the options and proposal for plugin initialization kept growing in response to comments from the reviewers. Most of the options are “visions”, without companion in-depths design. I think we can evolve through many smaller steps / releases, instead of trying to reach a grand final design in a big step.

Yes, a verifier app can be a normal, non-TEE app. It can even be in languages other than C / C++ and run on many different platforms.

The “Attester – Verifier Security Model and TEE Agnostic Design” section is a start to try to raise and answer the questions about TEE-agnostic application implementation and TEE-aware policies.

Agreed the design proposal for initialization of attestation plugins, and of OE SDK plugins in general, should be covered by its own dedicated design document. It could be a continuation of PR #2428. For a comprehensive design, it would be good to document the personas, user experience, and requirements.

Wonder if you could create a design doc / PR to hold the detailed comments. That will enable commenting on specific sections or bullets, or even allow collaborative editing?

docs/DesignDocs/NotesOnAttestationAPI.md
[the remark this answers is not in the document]
dthaler: I don't see oe_get_verifier_settings in any header under include/openenclave today.   In #2959 its part of the abstract API referred to as "GetChallenge".   I'm not sure that "settings" is the right word to use in a concrete api for this.

docs/DesignDocs/NotesOnAttestationAPI.md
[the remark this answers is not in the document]
dthaler: It is not a TODO item.  It's saying that RFC 4949 uses a more general term and a security policy for attestation is called an appraisal policy.

docs/DesignDocs/NotesOnAttestationAPI.md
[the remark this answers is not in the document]
shnwc: It's named `oe_verifier_get_format_settings` in include/openenclave/attestation/verifier.h.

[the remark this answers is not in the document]
dthaler: Ok that one I can find.   But "settings" often implies it's persistent (e.g., saved to disk) vs "parameters" like other OE APIs use which has no such connotation

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+
+## Semantics of Custom Claims and Optional Parameters
+
+This section explains the custom claims vs optional parameters for the [OE SDK attestation API](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/Attestation_API_Proposal.md#proposed-oe-sdk-v010-public-api)
+function `oe_get_evidence()`.
+
+Among all the parameters for the OE SDK attestation API
+function `oe_get_evidence()`,
+two of them are a list of custom claims (`custom_claims`) and an optional
+parameters blob(`opt_params`). While in general the fields in these two
```
dthaler, line 23:
```suggestion
parameters blob (`opt_params`). While in general the fields in these two
```

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+`oe_get_evidence()`, it would be good to articulate the semantics of these
+two entities.
+
+- Custom claims have well-defined semantics:
+  - Custom claims can be any flat array of bytes supplied by attester
+  applications.
+  - Custom claims are only meaningful to attester / verifier applications.
+  They are not inspected or interpreted by the OE SDK or its plugins.
+  - They are always TEE agnostic
+  - Custom claims are included and signed in the evidence as clear text
```
dthaler, line 41:
```suggestion
  - Custom claims are included and signed in the evidence
```
Remove "as clear text", since some evidence formats do encryption or some other form of encoding for which they would not appear as clear text. Or would also be ok to do

```suggestion
  - Custom claims are included and signed in the evidence (as clear text, for the current SGX formats)
```

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+  application for processing.
+    - Secrets shall never be included as a custom claim. Instead a hash of
+    a secret can be included by an attester as a custom claim in its evidence,
+    and passed to a verifier with integrity protection. The secret itself
+    has to be passed from the attester to the verifier with confidentiality
+    protection. Integrity of the secret can be verified by the verifier
+    using the hash held as a custom claim in the evidence.
+  - Custom claims can cover attester instance identity or
+  properties.
+    - This type of custom claims are created and maintained by the attester
```
dthaler, line 54:
```suggestion
    - Custom claims of this type are created and maintained by the attester
```
grammar

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+    identity and properties for the attester and for secure communication
+    with the latter.
+    - The public key of an asymmetric key-pair generated by an attester,
+    serving as an unique public identity of the attester,
+    is a good example of a custom claim.
+  - Custom claims can also be created by a verifier application and passed to
+  an attester for inclusion in the evidence.
+    - This type of claims allows the attester to make assertions about
+    its evidence to the targeted verifier.
+    - A good example of this type of claims is a nonce generated by a
```
dthaler, line 67: FYI Text is ok, but I don't think it's actually true in general.   A nonce (e.g., the size in bytes of the nonce can be fixed and vary by format or by TEE, or one format or TEE might require a timestamp whereas another requires a random number) can be format-specific and TEE-specific.   So this text is just one way of doing a nonce, and in that sense is ok, but nonces can also be done as optional parameters.   Again it's ok to not change the text if this is understood.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+to generate an evidence containing the SGX report targeted for the
+verifier enclave.
+- This SGX report can only be verified by the verifier plugin of the verifier
+enclave that generated the SGX target info.
+Any other verifier enclaves, either running on the local or a remote platform,
+won’t be able to verify this SGX report.
+
+## Attester – Verifier Security Model and TEE Agnostic Design
+
+This section explains how the proposed [OE SDK attestation API](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/Attestation_API_Proposal.md#proposed-oe-sdk-v010-public-api)
```
dthaler, line 109:
```suggestion
This section explains how the proposed [OE SDK attestation API](Attestation_API_Proposal.md#proposed-oe-sdk-v010-public-api)
```
use relative path.  Consider doing the same for other links in the doc.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+- But owners of attesters and verifiers define their security appraisal policies,
+and the policies are not TEE agnostic. For example:
+  - The policies define which TEEs, platforms and configurations are trusted
+  to run the attesters and verifiers.
+    - They also define security properties of these entities, e.g. keypairs
+    and certs, service provider ID / SPID for SGX EPID evidence, etc.
+    These security properties are required for attestation and secure
+    communication.
+  - The policies define the list of evidence formats acceptable to the verifiers.
+    - Many acceptable evidence format are TEE specific. Examples include
```
dthaler, line 123:
```suggestion
    - Many acceptable evidence formats are TEE specific. Examples include
```
grammar

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+    SGX location attestation evidence formats, SGX ECDSA evidence formats,
+    and SGX EPID evidence formats.
+- The OE SDK provides means for the attesters and verifiers to be built and
+configured for specified TEEs and platforms, and to generate or verify evidences
```
dthaler, line 127:
```suggestion
configured for specified TEEs and platforms, and to generate or verify evidence
```
grammar

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+* Evidence format, as exposed by the [OE SDK attestation API](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/Attestation_API_Proposal.md).
+* Plugin, as a logical unit with a [well-defined set of entry points](https://github.com/openenclave/openenclave/blob/master/include/openenclave/internal/plugin.h),
+to support generation or verification of evidence identified by an evidence format.
+* Plugin library, for implementation of one or more plugins.
+
+In the proposed [OE SDK attestation API](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/Attestation_API_Proposal.md),
+every evidence format is uniquely identified by a UUID. Enclave and host
+applications can discover formats supported for evidence verification.
+Enclave applications can select a format for evidence generation. For evidence
+generation, API `oe_get_evidence()` takes in an evidence format UUID along with
```
dthaler, line 152:
```suggestion
generation, the `oe_get_evidence()` API takes in an evidence format UUID along with
```
grammar

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+other input parameters. Upon successful operation, `oe_get_evidence()` returns
+an evidence blob with the evidence format UUID included as a field in its
+header structure.
```
dthaler, line 155: This sentence is about to become false with PR #3198, so reword.
```suggestion
a blob of evidence blob in the indicated format.  If the evidence is wrapped with an OE evidence header, the evidence format UUID included as a field in the
header structure.
```

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+Each plugin is associated with a single evidence format UUID.
+
+In theory multiple plugins can exist in a configuration for the same format,
+though in practice this should not be implemented as there is no clear benefit
+for the added complexity in implementation.
+
+Please note that an evidence format does not just identify a syntactic data format
+contract as the word "format" may suggest. It further identifies a semantic or
+behavior contract. In the context of the OE SDK attestation plugin design,
+a primary use of the of the evidence format UUID is to identify which verifier
```
dthaler, line 172:
```suggestion
a primary use of the evidence format UUID is to identify which verifier
```

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+plugins work with which attester plugins. A verifier plugin associated with a
+specific UUID is guaranteed to be able to verify evidence generated by an
+attester plugin associated with the same UUID.
+
+In an implementation of the OE SDK design, the set of supported plugins
+are implemented by one or more enclave-side and / or host-side plugin libraries and
+their dependencies. These plugin libraries can either be static or shared
+libraries, as supported by the build environment. There is no one-on-one
```
dthaler, line 180:
```suggestion
libraries, as supported by the build environment. There is no one-to-one
```

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+a single set of plugin libraries can implement multiple plugins. For example,
+both the enclave-side SGX ECDSA and EPID attester plugins are implemented by
+the same set of SGX attester plugin libraries.
+
+## Options for Enclave-side Plugin Library Initialization
+
+As described in the design document
+[Attestation: OE SDK Integration with Intel® SGX SDK quote-ex Library for Generation of Evidence in New Formats](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/SGX_QuoteEx_Integration.md),
+enclave-side plugin library initialization of attester plugins can be
+triggered by application enclave call of the OE SDK API
```
dthaler, line 194:
```suggestion
triggered by an application enclave calling the OE SDK API
```
grammar

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+`oe_attester_initialize()`. The OE SDK framework and plugin library
+implementation enumerates and registers all attester plugins configured
+for the platform and the calling application.
+
+In general, there are several options for initialization of enclave-side
+plugin libraries while keeping application enclaves TEE agnostic.
+
+In this section, we focus on the discussion of the options for initialization
+of attester plugins. Same options are available for initialization of
```
dthaler, line 203:
```suggestion
of attester plugins. The same options are available for initialization of
```

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+1. Every enclave-side plugin library implements the same OE SDK API
+`oe_attester_initialize()`.
+
+    With this option, a build of an application enclave project can only be linked
+    with a single attestation plugin library, to avoid function name collision.
+
+2. Build-time macro to convert from TEE-agnostic API to TEE-specific
+implementation.
+
+    With this option, application enclaves still call TEE-agnostic OE SDK API
```
dthaler, line 215:
```suggestion
    With this option, application enclaves still call the TEE-agnostic OE SDK API
```

docs/DesignDocs/NotesOnAttestationAPI.md
dthaler: What about adding a phrase about Yen's point, maybe at the end of 226?  Up to you.

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+    function for invocation on first entry to the enclave (upon enclave load).
+
+    With this option, the implementation of the OE SDK attestation framework
+    provides a hook for enclave-side plugin libraries to register their
+    initialization functions upon enclave load.
+
+    If ordering between the initialization functions of multiple libraries is
+    required, it can be supported by assigning different priorities to the
+    initialization functions of these libraries when they are registered
+    upon load.
```
dthaler, line 261: If I understand right, this means that if you're writing a new library for which ordering would matter, the priority is hard coded right?  So the priorities must be known of any other libraries that you might care about ordering before or after.  If you care about one that isn't part of OE SDK and isn't open source, is there an API or log to figure out the priority?   (This is one reason I don't like this option, which applies to https://github.com/jhand2/openenclave/blob/global_module_initialization/docs/DesignDocs/openenclave_modules.md as well I think.)   But as this document is not proposing something new, the text is ok, and we can track this as part of issue #2903

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+initialization, with some differences as discussed below.
+
+1. Every host-side verifier plugin library implements the same OE SDK API
+`oe_verifier_initialize()`.
+
+    With this option, a build of an OE SDK host verifier application can only
+    link with a single verifier plugin library, to avoid function name collision.
+
+    But if the host-side verifier plugin libraries are dynamically loaded,
+    then the above limit no longer applies. Arbitrary number of shared
```
dthaler, line 293:
```suggestion
    then the above limit no longer applies. An arbitrary number of shared
```
grammar

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,366 @@
...
+is promising. It gives application enclave developers the most
+control. But this option requires the OE SDK build and signing tools to
+add support of manifests.
+
+For initialization of host-side plugin libraries, option #3 (load-time
+manifest to list plugin libraries and their initialization entry points)
+is also promising, allowing load-time control of the list of
+supported plugin libraries and the order in which these libraries are
+initialized. But support of this operation requires the OE SDK host-side
+framework to add capability to process manifests.
```
dthaler, line 351:
```suggestion
framework to add the capability to process manifests.
```

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: The nonce is used as example. I think the sub-bullets could be confusing, and I'll delete them.

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks! Added a sentence.

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Added note on lack of scalability.

Yes, agree with your comments about ordering.

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

docs/DesignDocs/NotesOnAttestationAPI.md
shnwc: Thanks!

shnwc:
@dthaler Thank you for your comments and approval! The latest commit has incorporated all the feedback, and is ready for merge.

dthaler:
bors try

docs/DesignDocs/NotesOnAttestationAPI.md
```diff
@@ -0,0 +1,362 @@
...
+have distinct semantics that might not be straightforward to application
+and plugin software developers.
+
+To help clarify what goes into the custom claims (`custom_claims`) and
+what goes into the optional parameters (`opt_params`) in the function
+`oe_get_evidence()`, it would be good to articulate the semantics of these
+two entities.
+
+- Custom claims have well-defined semantics:
+  - Custom claims can be any flat array of bytes supplied by attester
```
jxyang, line 36: I don't see the point of having plugin writers to implement their own serialization/deserialization routines for custom claims. The previous implementation defines an array of oe_claim_t objects, and provides serialization/deserialization for plugins. We should keep those, while at the same time, allow some plugins, to override the default serialization/deserialization routines if needed.

shnwc:
@jxyang please see my comments in PR #3264 on why the custom claims are changed to be held in a flat buffer. The application coders, instead of plugin writers, are responsible to put structure in their custom claims.

docs/DesignDocs/NotesOnAttestationAPI.md
dthaler: @jxyang I wouldn't assume that two plugins serialize custom claims the same way.   The only value of having some common implementation is if it's common for many plugins to do it the same way.  Not sure we have enough evidence of that yet.   Many other formats are standard, not OE specific, so overriding would be the norm I would think.

jxyang: @dthaler  This is a pain point from an user point of view. I have personally used SGX plugin and EEID plugin. And I think they share the same set of custom claims. In fact, I am implementing attestation TLS for SGX plugin first and then move to EEID plugin later. Both of them depend on custom claims.

yentsanglee:
This document helps conveying concepts behind design. Not meant to be consumed by OE SDK public audiences. Save to repo for now until we find a better repository (wiki, portal, etc) for it.

Issue #2796, opened 2020-04-06
Guidance for TEE specific naming of functions/types


jhand2:
There are many functions and types which are TEE specific but do not have names which denote that. For example the internal function `oe_get_td()` and the corresponding `td_t` structure are only used on SGX. Changing the naming to `oe_sgx_get_td()` and `sgx_td_t` would avoid naming conflicts (either at compile time or link time) in case another tee wants to internally introduce similar behavior with similar naming.

Why question is when should we provide guidance to name something with a TEE-specifc name? Is the desire that an application can link code for two TEEs and then decide which to use at runtime? Or that common internal code should be able to include headers for different TEE implementations in the same file without a compile-time switch?

In a world where TEE-specific code is split into different libraries, are these concerns still valid?

dthaler:
Some thoughts:
* My suggestion would be that all TEE-specific functions, even if internal, have the tee prefix, like oe_sgx_... or oe_optee… etc.   In my view, this makes backtraces, debugging symbols, etc. all be unambiguous and helps prevent later collisions.
* Having unique names might be especially useful for simulation tests that might be able to test multiple TEEs in simulation mode, so avoiding naming conflicts at link time I think would be very useful.

achamayou:
+1 for systematic prefixing.

CodeMonkeyLeet:
> My question is when should we provide guidance to name something with a TEE-specific name?

Yes, I think this is a good idea to standardize the naming convention around OE functions that are specific to a given TEE implementation. 

> Is the desire that an application can link code for two TEEs and then decide which to use at runtime? Or that common internal code should be able to include headers for different TEE implementations in the same file without a compile-time switch?

Yes. As a general design point, it is desirable that architecture compatible binaries (e.g. x86-64, ARM) that could span different TEE implementations can build the applications once, and instantiate the appropriate TEE at runtime instead of requiring separate builds. It may be that TEE differences will prevent this (e.g. OP-TEE and SGX today are on different architectures), but it would be nice not to introduce additional barriers that could be addressed early on with something simple like a naming convention.

> In a world where TEE-specific code is split into different libraries, are these concerns still valid?

Given the previous assertion, the same concerns continue to apply around symbol collisions unless we implement a fully dynamic load process for TEE-specific code factored out into shared libraries. Within the enclave, this implies implementing support for multi-module loading and it would be nice to not require that simply to avoid symbol collisions. 

[...]

PR #2804, opened 2020-04-07
Rename SGX-specific ocalls with SGX names

jhand2:
Rename ocalls in SGX-specific edl files to have SGX-specific names. This avoids potential naming conflicts with implementations from other TEEs.

Note that attestation-specific ocalls are left alone since they will hopefully be moved into plugins and these builtin EDL functions will be deprecated.

### DIFF
--- a/enclave/core/sgx/backtrace.c
+++ b/enclave/core/sgx/backtrace.c
@@ -112,7 +112,7 @@ char** oe_backtrace_symbols_impl(
         goto done;
 
     /* First call might return OE_BUFFER_TOO_SMALL. */
-    if (oe_backtrace_symbols_ocall(
+    if (oe_sgx_backtrace_symbols_ocall(
             &retval,
             oe_get_enclave(),
             (const uint64_t*)buffer,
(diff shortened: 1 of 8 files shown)


enclave/core/sgx/backtrace.c
```diff
@@ -112,7 +112,7 @@ char** oe_backtrace_symbols_impl(
         goto done;
 
     /* First call might return OE_BUFFER_TOO_SMALL. */
-    if (oe_backtrace_symbols_ocall(
+    if (oe_sgx_backtrace_symbols_ocall(
```
dthaler, line 115: Should also add `_sgx_` into lines 93, 184, and 191.  Similarly in other .c files in this directory.  Do you want to do that in this PR or in a separate PR?

jhand2: The goal of this PR was just to fix up the EDL files before I publish them. Still tracking the need to fix up naming across the board with https://github.com/openenclave/openenclave/issues/2796

PR #2806, opened 2020-04-08
oe-gdb: Automatically continue execution upon cpuid SIGILL

anakrish:
OE SDK's crypto library (mbedtls) uses cpuid instruction to check for
aesni capability. This causes a SIGILL exception that is eventually
handled by the enclave by emulating the cpuid exception.

This SIGILL however causes the debugger to stop. Anyone that debugs
a crypto application would be first greeted with this SIGILL as the
crypto library initialized.

This PR changes oegdb to continue executing without stopping when
SIGILLs are raised due to cpuid.

### DIFF
--- a/debugger/pythonExtension/signal_handler.py
+++ b/debugger/pythonExtension/signal_handler.py
@@ -1,0 +1,56 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+import gdb
+import sys
+
+# Instructions for which SIGILL is ignored.
+ignored_instructions = set()
+
+# Get mnemonic at pc.
+def get_sigill_mnemonic():
+    try:
+        frame = gdb.newest_frame()
+        # Disassemble to find out the instruction
+        # See: https://wiki.osdev.org/X86-64_Instruction_Encoding
+        max_instruction_length = 15
+        start_pc = frame.pc()
+        end_pc = start_pc + max_instruction_length
+        asm = frame.architecture().disassemble(start_pc= start_pc,
+                                               end_pc = end_pc)
+
+        insn = asm[0]['asm']
+        mnemonic = insn.split()[0]
+        return mnemonic
+    except:
+        return None
+
+def oe_stop_event_handler(event):
+    if isinstance(event, gdb.SignalEvent) and event.stop_signal == "SIGILL":
+        insn = get_sigill_mnemonic()
+        if insn in ignored_instructions:
+            print("oegdb: Ignoring sigill from " + insn +  ".")
+            gdb.execute("continue", from_tty=True, to_string=False)
+
+class IgnoreSigill(gdb.Command):
+    """ Ignore SIGILL raised by given instruction """
+    def __init__(self):
+        super(IgnoreSigill, self).__init__('oegdb-ignore-sigill', gdb.COMMAND_USER)
+
+    def invoke(self, arg, from_tty):
+        ignored_instructions.add(arg)
+        if from_tty:
+            print("oegdb: Added " + arg + " to SIGILL ignore list.")
+
+
+def register():
+    IgnoreSigill()
+    gdb.events.stop.connect(oe_stop_event_handler)
+    print("oegdb: Registered stop event handler to handle SIGILL.\n")
+    gdb.execute("oegdb-ignore-sigill cpuid", from_tty=True)
+
+def unregister():
+    try:
+        gdb.events.stop.disconnect(oe_stop_event_handler)
+    except:
+        pass
--- a/tests/debugger/oegdb/commands.gdb
+++ b/tests/debugger/oegdb/commands.gdb
@@ -1,8 +1,14 @@
 # Copyright (c) Open Enclave SDK contributors.
 # Licensed under the MIT License.
 
+# Pring gdb version
+python import os; os.system("gdb -v")
+
 # Enable pending breakpoints
 set breakpoint pending on
+
+# Add rdtsc to sigill ignore list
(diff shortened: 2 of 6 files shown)


anakrish:
bors try

debugger/pythonExtension/signal_handler.py
```diff
@@ -0,0 +1,29 @@
...
+# Licensed under the MIT License.
+
+import gdb
+import sys
+
+def oe_stop_event_handler(event):
+    if isinstance(event, gdb.SignalEvent) and event.stop_signal == "SIGILL":
+        try:
+            frame = gdb.newest_frame()
+            # Disassemble to find out the instructio
```
jhand2, line 11:
```suggestion
            # Disassemble to find out the instruction
```

debugger/pythonExtension/signal_handler.py
```diff
@@ -0,0 +1,29 @@
...
+                                                   end_pc = end_pc)
+
+            if asm[0]['asm'].startswith('cpuid'):
+                # Continue execution
+                print("oegdb: Ignoring emulated instruction cpuid.")
+                gdb.execute("continue", from_tty=True, to_string=False)
+        except:
+            print("Unexpected error:", sys.exc_info()[1])
+
+def register():
```
jhand2, line 26: Minor suggestion: `signal_handler.register` seems like it would take as a parameter the signal handler. Maybe `signal_handler.register_default()`

anakrish:
bors try

anakrish:
@achamayou @jhand2 @mingweishih Can you take another look?

I got a request to ignore `rdtsc` too and there might be other instructions the need to be ignored by an SGX enclave application. Hence created a GDB command `oegdb-ignore-sigill <instruction>`.
cpuid is ignored by default.

Though oegdb will continue execution for SIGILLs due to given instructions, the debugger will still print output that the signal was raised. The only way to suppress any output seems to be to do
`handle signal SIGILL noprint` upon which GDB will ignore all SIGILL signals.

So at best we can have GDB continue execution for certian SIGILLs and have printouts or
have GDB continue execution for all SIGILLs and no output :(

tests/debugger/oegdb/enc/enc.c
```diff
@@ -44,6 +60,9 @@ int square(int x)
         oe_is_within_enclave(&x, sizeof(x)) ? "within enclave"
                                             : "outside enclave");
     g_square_called = true;
+
+    // Raising signal in a function called by gdb crashes gdb 8.0 and above.
```
CodeMonkeyLeet, line 64: For my understanding, is this expected?

debugger/pythonExtension/signal_handler.py
```diff
@@ -0,0 +1,50 @@
...
+import sys
+
+# Instructions for which SIGILL is ignored.
+ignored_instructions = set()
+
+# Get mnemonic at pc.
+def get_sigill_mnemonic():
+    try:
+        frame = gdb.newest_frame()
+        # Disassemble to find out the instructio
```
CodeMonkeyLeet, line 14:
```suggestion
        # Disassemble to find out the instruction
```

tests/debugger/oegdb/enc/enc.c
anakrish: gdb 8.x seems like a downgrade from gdb 7.x. Lots of stability issue. It is the default in Ubuntu 19.10 which I'm running.

anakrish: From previous run on RHEL 8.
https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/rest/organizations/jenkins/pipelines/Bors/branches/trying/runs/209/nodes/66/steps/453/log/?start=0

```bash
Program received signal SIGILL, Illegal instruction.
0x00007ffff50034f6 in _raise_cpuid_exception () at ../tests/debugger/oegdb/enc/enc.c:39
39	    asm("movl  $1, %%eax   \n\t"
oegdb: Ignoring emulated instruction cpuid.
Continuing.
square called with 100
x lies within enclave
../../gdb/infcall.c:1188: internal-error: value* call_function_by_hand_dummy(value*, type*, int, value**, void (*)(void*, int), void*): Assertion `call_thread->thread_fsm == &sm->thread_fsm' failed.
A problem internal to GDB has been detected,
further debugging may prove unreliable.
Quit this debugging session? (y or n) [answered Y; input not from terminal]
```

debugger/pythonExtension/signal_handler.py
anakrish: Fixed.

anakrish:
Added log output to print gdb versions that trigger internal assertions.

anakrish:
bors try

anakrish:
bors try

achamayou:
@anakrish I think I would prefer to have the printouts and have this restricted to SIGILL/cpuid, I would worry about missing something otherwise.

kuanggang-git:

tests/debugger/oegdb/commands.gdb
```diff
@@ -1,9 +1,15 @@
...
 # Licensed under the MIT License.
 
+# Pring gdb version
+python import os; os.system("gdb -v")
+
 # Enable pending breakpoints
 set breakpoint pending on
 
+# Add rdtsc to sigill ignore list
+oegdb-ignore-sigill rdtsc
```
jxyang, line 11: Why `rdtsc` is not registered by default?

anakrish:
bors try

anakrish:
Closing this PR for now. Unfortunately, newer versions of GDB are unstable and crash with this change.

PR #2809, opened 2020-04-08
Make Windows file syscalls support Unicode paths

dthaler:
The full range of Unicode should now be supported for file syscalls with these changes.
Includes a test case to verify Unicode path conversion.
This also fixes a memory corruption bug at the end of _get_full_path_name which passed the wrong pointer to free().

**Background:** 
POSIX uses UTF-8 encoded in char* strings.  Windows instead treats char* strings as being in whatever the current ANSI code page is, and in general an ANSI code page cannot represent all Unicode characters so using any "char*" file API on windows inherently has problems of possibly failing to be able to represent actual file paths in use.   Instead, Windows uses UTF-16LE in wchar_t* strings (aka PWSTR) for the full range of Unicode.   Note that this also means that one can NOT reliably take a char* string buffer obtained from an enclave's file apis (which use UTF-8) and pass it to a Windows api like create_enclave, since the meaning of the bytes can be completely different between UTF-8 and the current ANSI code page.   Similarly, one cannot take char* argv[] from a console app and expect to pass it unmodified to an enclave that wants to pass it to a file API that supports UTF-8, since again the meaning of the bytes can be completely different.   As such, main() is problematic and instead wmain() is used for Unicode tests.   This PR is scoped to just the file syscalls, and leaves the following two issues unchanged as out of scope for this PR:
1. oe_create_enclave (and the underlying oe_fopen) use ANSI and so have problems with non-ASCII paths
2. The getaddrinfo and getnameinfo syscalls also blindly pass ANSI strings to UTF-8 apis and vice versa and so have problems with non-ASCII hostnames

### DIFF
--- a/host/windows/syscall.c
+++ b/host/windows/syscall.c
@@ -52,9 +52,9 @@
 struct WIN_DIR_DATA
 {
     HANDLE hFind;
-    WIN32_FIND_DATAA FindFileData;
+    WIN32_FIND_DATAW FindFileData;
     int dir_offs;
-    char* pdirpath;
+    PCWSTR pdirpath;
 };
 
 /*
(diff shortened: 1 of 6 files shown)


dthaler:
bors try

nonpolarity:
Thanks @dthaler for adding this important feature and also fixing my error!

dthaler:
bors try

dthaler:
bors try

dthaler:
bors try

PR #2810, opened 2020-04-08
Change default CMAKE_BUILD_TYPE to Release from Debug.

nonpolarity:
It is insecure for end users to use the SDK in debug mode, especially in a production environment.
The SDK developers now need to specify "-DCMAKE_BUILD_TYPE=Debug" to compile in debug mode.

### DIFF
--- a/cmake/compiler_settings.cmake
+++ b/cmake/compiler_settings.cmake
@@ -27,7 +27,7 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 # Set default build type and sanitize.
 # TODO: See #756: Fix this since debug is the default.
 if (NOT CMAKE_BUILD_TYPE)
-  set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Build type" FORCE)
+  set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Build type" FORCE)
 endif ()
 set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug;Release;RelWithDebInfo")
 
--- a/cmake/compiler_settings.cmake
+++ b/cmake/compiler_settings.cmake
@@ -70,6 +70,16 @@ else ()
 endif ()
 
 if (CMAKE_CXX_COMPILER_ID MATCHES GNU OR CMAKE_CXX_COMPILER_ID MATCHES Clang)
+  # Enforce -O0 in debug mode.
+  string(TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_UPPER)
+  if (CMAKE_BUILD_TYPE_UPPER STREQUAL "DEBUG")
+    # For gcc the -g option alredy includes -O0.
+    # Here we just apply -O0 for Clang.
+    if (CMAKE_CXX_COMPILER_ID MATCHES Clang)
+      set (CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -O0")
+      set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0")
+    endif()
+  endif()
   # Enables all the warnings about constructions that some users consider questionable,
   # and that are easy to avoid. Treat at warnings-as-errors, which forces developers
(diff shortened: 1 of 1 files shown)


nonpolarity:
bors try

jhand2:
Hmmm. Debug mode is just with symbols right? I don't think this flag makes an enclave a debug mode enclave.

nonpolarity:
@jhand2 we can have more discussion. Could you help me to hold this pr? I have no permission.

cmake/compiler_settings.cmake
```diff
@@ -27,7 +27,7 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 # Set default build type and sanitize.
 # TODO: See #756: Fix this since debug is the default.
 if (NOT CMAKE_BUILD_TYPE)
-  set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Build type" FORCE)
```
CodeMonkeyLeet, line 30: This change is not necessary given the reason in the PR description:
* This affects the default build option for contributors building the OE SDK directly, and in those cases, building Debug by default is more generally useful for the developers. The build system already builds RelWithDebInfo for OE SDK releases.
* The `CMAKE_BUILD_TYPE` only controls the compiler settings (optimization levels and inclusion of symbols), which is distinct from the enclave's `ALLOW_DEBUG` property set by `OE_SET_ENCLAVE_SGX` or by oesign. The latter flag is the security concern and is what allows the enclave to be created with the `OE_ENCLAVE_FLAG_DEBUG` flag that allows debuggers to attach to and introspect the enclave.
    * It is a specific use case to allow a Release compiled enclave binary set with the `ALLOW_DEBUG` flag for testing with a test signature, then re-sign the enclave with a different production key and the `ALLOW_DEBUG` flag disabled without having to recompile the enclave.

cmake/compiler_settings.cmake
```diff
@@ -70,6 +70,16 @@ else ()
 endif ()
 
 if (CMAKE_CXX_COMPILER_ID MATCHES GNU OR CMAKE_CXX_COMPILER_ID MATCHES Clang)
+  # Enforce -O0 in debug mode.
+  string(TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_UPPER)
+  if (CMAKE_BUILD_TYPE_UPPER STREQUAL "DEBUG")
+    # For gcc the -g option alredy includes -O0.
```
CodeMonkeyLeet, line 76: Is this change related to the changing of the default `CMAKE_BUILD_TYPE` in the PR description? It's seems like a good change to enforce consistency, but the PR should probably be updated to explain the impact of this fix instead.

@anakrish I wasn't aware that we default to `-O0` for debug builds; for my understanding, is there a reason for that instead of using `-Og`, or should we be using that instead for both gcc & clang?

nonpolarity: @CodeMonkeyLeet Thanks for your explanation! Anand has answered in another pr, just because that pr is not merge yet so the change also shows up here.
The gcc -g already include -O0 by default, but clang does not.
Anand also mentioned if -Og is a better choice. For clang -Og is still under development. for gcc -Og has some optimization that some symbols are removed and backtrace test can not pass.

cmake/compiler_settings.cmake
nonpolarity: Thanks! For the host side, most open source releases always turn off debug mode, since debug mode may have detailed log output. Since enclave side has more security check, this is not a security issue. I will withdraw this PR.

cmake/compiler_settings.cmake
CodeMonkeyLeet: @nonpolarity Thanks for the explanation, wasn't aware for the clang differences in behavior. Can we just drop commit dd9bfc13e054e49afc57ab7a741e8cffbcbf172f and just take this one then?

anakrish:
@nonpolarity I agree that it is good to have compiler appropriate flags for debug builds.

I agree with @CodeMonkeyLeet that we don't need to change the default build type.

nonpolarity:
Thanks @CodeMonkeyLeet @anakrish, the debug option PR is #2612 and this one is dropped.

PR #2814, opened 2020-04-08
Add OESDK mailing list to documentation

dthaler:
Addresses half of issue #2742
Discussion around the CGC mailing list is left to PR #2744

Factored out the non-CGC part in the hopes that this can be merged now


jhand2:
Thanks

bors r+

Issue #2797, opened 2020-04-07
Style enforcement and "linting" should be done by software, not human-ware


AevaOnline:
Other projects I have participated in prefer to automate the tasks of enforcing style guides, lint checking, and spelling/grammar checking (where appropriate). We seem to have some of that in place today, for example checking the formatting of .md files and the presence of license headers in code files, but many code reviewers still perform lint checking via github comments / suggestions.

I'd like to suggest this is a "bug" in our current process (the bug being that a lack of a lint-checker creates unnecessary work for humans), and that we fix this "bug".

I am not as familiar with the toolsets for C/C++ linting as I am with the Python, but I found the [cpplint tool](https://github.com/cpplint/cpplint) to quickly demonstrate the kind of thing I am familiar with. (just an example -- there may very well be other tools that do this, too)

```
(.venv) aeva@AEVA-SURFACE1:~/source/ccc/openenclave/openenclave$ cpplint --extensions=c,cc,h,cpp --recursive tests/ 2>&1 |head
tests/stdcxx/enc/enc.cpp:14:  Include the directory when naming .h files  [build/include_subdir] [4]
tests/stdcxx/enc/enc.cpp:18:  Do not use namespace using-directives.  Use using-declarations instead.  [build/namespaces] [5]
tests/stdcxx/enc/enc.cpp:23:  Static/global string variables are not permitted.  [runtime/string] [4]
tests/stdcxx/enc/enc.cpp:26:  { should almost always be at the end of the previous line  [whitespace/braces] [4]
tests/stdcxx/enc/enc.cpp:27:  public: should be indented +1 space inside class x  [whitespace/indent] [3]
tests/stdcxx/enc/enc.cpp:29:  { should almost always be at the end of the previous line  [whitespace/braces] [4]
tests/stdcxx/enc/enc.cpp:33:  { should almost always be at the end of the previous line  [whitespace/braces] [4]
tests/stdcxx/enc/enc.cpp:37:  { should almost always be at the end of the previous line  [whitespace/braces] [4]
```

jhand2:
For coding style we use clang-format. This is enforced by a commit hook as well as in CI.

We have also done some work with scan-build (an llvm-based static analyzer to catch bugs such as use-after-free, memory leaks, etc). We have fixed a lot of bugs by devs running it locally but there is still work to be done to get it running in CI.

AevaOnline:
[...]

Issue #2719, opened 2020-03-20
Document which features use which system ocalls


jhand2:
#2676 Tracks adding support to give developers more control over which ocalls the OE runtime may call on their behalf. Since developers will now have a choice to disable the OE runtime from making certain ocalls, we should document which features make which ocalls.

For example, if a developer wants to disable the syscall ocalls (by not including them in EDL), there should be enough documentation that the developer knows which posix functions they must not call in oelibc to resolve unresolved symbol linker errors.

PR #2816, opened 2020-04-08
Install system EDL files with the SDK

### LINKED ISSUE #2676: Require explicit developer opt-in for OCalls made by the OE runtime
OE has many OCalls which it will make during execution of the runtime, without an enclave application explicitly making these OCalls. We should make developers explicitly import these into their own EDL files so that an the OE runtime cannot make an OCall which was not opted into.

This is a meta issue to track smaller work items to support this feature

* [x] Design documentation: PR https://github.com/openenclave/openenclave/pull/2482
* [x] Refactor EDL into smaller pieces of functionality: Issue https://github.com/openenclave/openenclave/issues/2585
* [x] Generate structures from annotated EDL: Isse https://github.com/openenclave/openenclave/issues/2666
* [ ] Additional documentation to aid developers in opting-out of specific ocalls: Issue #2719
* [ ] Publish EDL files with the SDK: PR #2816

Known bugs:
* [ ] #2817
* [ ] #2818

jhand2:
Install system EDL files with the SDK and provide a build flag to OESDK to compile the SDK with/without builtin ecalls and ocalls. If this flag is disabled, each enclave will need to explicitly import the ecalls and ocalls used by their application in their own EDL file.

This also adds a test to show how ecalls and ocalls should be imported when the COMPILE_SYSTEM_EDL flag is disabled.

There are currently some limitationn as to the granularity at which ocalls can be opted out of.

1. Some features (switchless, cpuid, backtrace, logging) are built into oecore. These currently cannot be opted out of
2. The syscall feature uses a single large switch statement for making ocalls. Because of this, --gc-sections doesn't actually remove any extra ocalls. If you link liboesyscall, then all syscall ocalls will be included in the enclave binary.

Fixes #2676


jhand2:
TODO: Update changelog detailing deprecation notice of builtin ocalls/ecalls

kuanggang-git:

jhand2:
Holding off on merging this until #2837 is merged since they touch the same code

CHANGELOG.md
```diff
@@ -35,6 +39,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 https://tls.mbed.org/tech-updates/releases/mbedtls-2.16.5-and-2.7.14-released)
 release notes for the set of issues addressed.
 
+### Deprecation Notice
```
jhand2, line 42: @radhikaj let me know if you think the deprecation notice is fine or need more info/guidance

jhand2:
bors try

CHANGELOG.md
```diff
@@ -15,9 +15,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support `transition_using_threads` EDL attribute for ecalls in oeedger8r.
   OE SDK now supports both switchless OCALLs and ECALLs.
 - Published corelibc headers required by oeedger8r-generated code.
-  Disclaimer: these headers do not make any guarantees about stability. They
+  **Disclaimer:** these headers do not make any guarantees about stability. They
   are intended to be used by generated code and are not part of the OE public
   API surface.
+- `COMPILE_SYSTEM_EDL` build flag. This is on by default and will compile system
+  OCalls and ECalls into OE libraries as before. If it is set to off, each enclave
+  application must import the ECalls/OCalls their application needs into their
```
dthaler, line 23:
```suggestion
  application must import the ECalls/OCalls it needs into its```

File: enclave/core/CMakeLists.txt [hunk 3]
```diff
@@ -330,6 +359,10 @@ if(USE_DEBUG_MALLOC)
     message("USE_DEBUG_MALLOC is set, building oecore with memory leak detection.")
 endif()
 
+if (COMPILE_SYSTEM_EDL)
```
dthaler, line 362: Since 267 differs in style from 362, please run format-code on this file.  If format-code isn't processing CMakeLists.txt files, then please update issue #2797 with a comment about this being another gap.

tests/exclude_system_edl/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+    add_subdirectory(host)
+
+    if (BUILD_ENCLAVES)
+        add_subdirectory(enc)
+    endif()
+
+    add_enclave_test(tests/exclude_system_edl
+                     exclude_system_edl_host exclude_system_edl_enc)
+else ()
+    # Throw a helpful error here
```
dthaler, line 16: Is this a TODO or what?

tests/exclude_system_edl/enc/enc.c
```diff
@@ -0,0 +1,25 @@
...
+{
+    // Since nanosleep makes an ocall, we will need to include
+    // the proper ocall in our EDL file.
+    struct timespec t;
+    t.tv_sec = 0;
+    t.tv_nsec = 100;
+    nanosleep(&t, &t);
+}
+
+OE_SET_ENCLAVE_SGX(
```
dthaler, line 19: Question: Should this be using a .conf file instead?

tests/exclude_system_edl/host/host.c
```diff
@@ -0,0 +1,35 @@
...
+    if (argc != 2)
+    {
+        fprintf(stderr, "Usage: %s ENCLAVE_PATH\n", argv[0]);
+        return 1;
+    }
+
+    const uint32_t flags = oe_get_create_flags();
+
+    if ((result = oe_create_exclude_system_edl_enclave(
+             argv[1], OE_ENCLAVE_TYPE_SGX, flags, NULL, 0, &enclave)) != OE_OK)
```
dthaler, line 25: Why TYPE_SGX instead of TYPE_AUTO?   This test doesn't appear to be SGX-specific

enclave/core/CMakeLists.txt
jhand2: It only runs on .c, .cpp, and .h. I have updated the linked issue

tests/exclude_system_edl/CMakeLists.txt
jhand2: Oh I forgot to change this. I put the check at the root cmakelists.txt instead. Will remove

tests/exclude_system_edl/enc/enc.c
jhand2: Good question. Is that our guidance? I think it is.

Is there an example of a test that works for both SGX and trustzone without using the `OE_SET_ENCLAVE_*` macros that I can follow as a template?

tests/exclude_system_edl/host/host.c
jhand2: You're right, I mostly just copied another test to quickly make this test. I'll fix it up to not be SGX-specific

edl/sgx/platform.edl
```diff
@@ -18,4 +18,5 @@ enclave
     from "cpu.edl" import *;
     from "debug.edl" import *;
     from "thread.edl" import *;
+    from "switchless.edl" import *;
```
jhand2, line 21: Now that switchless is part of platform.edl for sgx, remove the switchless.edl wrapper

tests/exclude_system_edl/enc/enc.c
dthaler: Will defer to @CodeMonkeyLeet on what it should be doing, but I don't know of any good tests right now.   They all either use OE_SET_ENCLAVE_* for both SGX and OPTEE, or else they use .conf and don't work for OPTEE.   It's fine with me if this gets deferred to some item that updates tests as a whole to use .conf (if that's the recommendation), if there is some issue filed to do so.

jhand2: Ok so I looked a little more and while I think this is a good idea it looks like we don't actually have support for passing .conf files in creating optee enclaves. I have updated the test to use both OE_SET_ENCLAVE* macros and opened https://github.com/openenclave/openenclave/issues/2844 to track switching to using .conf files.

jhand2:
bors try

tests/exclude_system_edl/enc/enc.c
CodeMonkeyLeet: > Good question. Is that our guidance? I think it is.

There's no guidance around this currently. Samples do use .conf files, but the vast majority of tests do not, except for a handful that test signing specifically. Just in terms of consistency, I'd say use the macros, particularly since .conf files will definitely not work for OP-TEE at the moment.

> Is there an example of a test that works for both SGX and trustzone without using the `OE_SET_ENCLAVE_*` macros that I can follow as a template?

Hernan has added a set of tests that also do build in OP-TEE via the macro, and it coexists with the SGX version fine (they currently result in non-overlapping definitions anyway). See any tests in tests/CMakeLists.txt under the `if (UNIX ...)` block but not `if (OE_SGX)`

jhand2:
bors try

tests/exclude_system_edl/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) Open Enclave SDK contributors.
```
CodeMonkeyLeet, line 1: Could you add a README.md or some description as to what this test is supposed to illustrate? Is this just a temporary test that will be removed once all tests are updated to import system EDLs explicitly?

syscall/devices/hostepoll/CMakeLists.txt
```diff
@@ -5,8 +5,6 @@ add_enclave_library(oehostepoll STATIC hostepoll.c)
 
 maybe_build_using_clangw(oehostepoll)
 
-add_enclave_dependencies(oehostepoll syscall_trusted_edl)
```
CodeMonkeyLeet, line 8: For my understanding, these can all be removed because they were always getting this dependency transitively through the link dependency on oesyscall anyway, correct?

tests/exclude_system_edl/CMakeLists.txt
jhand2: This is just a temporary test to make sure everything works when COMPILE_SYSTEM_EDL=OFF. I can add a README.md to explain that

syscall/devices/hostepoll/CMakeLists.txt
jhand2: correct

jhand2:
> Just responding to some comment questions; I'm not sure I grok the scope of this change relative to #2676. Specifically in relation to the task:
> 
> > Additional documentation to aid developers in opting-out of specific ocalls: Issue #2719
> 
> #2719 is an issue deferred to v0.10 but that doesn't really touch on other documentation for SDK users of how they're supposed to manage the transition to not using `COMPILE_SYSTEM_EDL`; we have a test, but there's no sample or docs showing the `COMPILE_SYSTEM_EDL=OFF` workflow except for the one-line in the CHANGELOG.md.
>

Good point I'll add a section to the ocall opt-in doc with more details for 0.9.

> 
> Also, I'm not clear that the same guidance holds if we fix #2817/#2818 for v0.10? i.e. will there still be required system EDLs if they are fixed? Is that the intention?

So if #2817 is implemented then our guidance will change from "import syscall.edl" to "only import the syscalls you use". Even though the guidance changes, I see no reason to get rid of syscall.edl at least in the short term. So both would work.

#2818 is a little more complicated.
* Currently logging.edl is required. If we provide a mechanism to turn off logging to the host then the guidance would just be you don't have to import than anymore
* Currently all the files imported by sgx/platform.edl are required. Some of them (like switchless) are easy to fix. Others (like cpuid) might require a little more thought. The hard thing is we can't just remove switchless.edl from platform.edl because it will break people using platform.edl. In that case it's all or nothing. If you include sgx/platform.edl you get everything. If you don't then you have to import each individually. That's probably ok. So the guidance would be if you want to maintain parity, import platform.edl. If you want to be more choosey about which ecalls/ocalls you import you're more or less on your own.

tests/exclude_system_edl/CMakeLists.txt
jhand2: Added a README

CHANGELOG.md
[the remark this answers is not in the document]
jhand2: I hesitate to make this official guidance lest people mob me for making the deprecation process too cumbersome. I chose a more conservative model for this particular feature to give people more time to switch over to the new way. I don't really want to take the guidance further than that.

jhand2:
@CodeMonkeyLeet I added a section to [the design doc for this feature](docs/DesignDocs/system_ocall_opt_in.md) on porting guidance. Let me know if that clears anything up or if it is just more confusing.

jhand2:
bors try

CHANGELOG.md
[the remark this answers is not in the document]
radhikaj: I agree on
> it might be clearer to say that we are deprecating the automatic inclusion of oecore EDL by default, and that they should use the COMPILE_SYSTEM_EDL flag to test this behavior ahead of time.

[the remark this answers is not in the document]
jhand2: I updated the CHANGELOG and pointed to the design doc for further guidance on what to do. let me know if that's what your were looking for.

Also please let me know if there's anything I can change in the doc for clarity.

CHANGELOG.md
```diff
@@ -35,6 +39,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 https://tls.mbed.org/tech-updates/releases/mbedtls-2.16.5-and-2.7.14-released)
 release notes for the set of issues addressed.
 
+### Deprecated
```
radhikaj, line 42: Thinking about it a little more, I think calling it out in deprecated is wise since this is what developers will look at in order to plan for the next release.

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -138,3 +138,67 @@ in one of the following EDL files
...
+and SGX attestation. Until these features can be disabled, it is
+recommended that you simply import all calls from sgx/platform.edl.
+
+**syscall.edl**
+
+`syscall.edl` is only required if you link liboesyscall into your application.
+Currently syscalls cannot be chosen individually, so if an enclave depends on
+one syscall, all of the syscall OCalls will need to be imported.
+
+### Changing guidance
```
radhikaj, line 171: Thanks a lot for putting this together @jhand2. Much appreciated.

tests/exclude_system_edl/enc/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+set (EDL_FILE ../exclude_system_edl.edl)
+
+# Note: This pathing hack is to work around the fact that EDL/oeedger8r does
+# not support C preprocessor macros as suggested in #2833. If that is
+# implemented this will no longer be needed.
+if (OE_SGX)
```
radhikaj, line 9: I see the value in the macros! Thanks for adding this comment.

CHANGELOG.md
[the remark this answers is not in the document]
CodeMonkeyLeet: > I hesitate to make this official guidance lest people mob me for making the deprecation process too cumbersome. I chose a more conservative model for this particular feature to give people more time to switch over to the new way. I don't really want to take the guidance further than that.

To clarify, I'm not proposing that you make any kind of assertion about official guidance for the deprecation process in general, I'm asking how you envision the deprecation process of _this_ feature to proceed. The text asserts that the plan is:

- v0.9 does not break by default, supports `COMPILE_SYSTEM_EDL` for devs to experiment with importing EDLs manually.
- v0.10 breaks by default, removes `COMPILE_SYSTEM_EDL` entirely and devs must scramble to fix if they were not aware of the change in behavior previously.

I'm pointing out that is not an ideal developer experience. Many users do not read the CHANGELOG and will likely not be aware that there is any action needed if the builds transparently continue to work. This is why the existing two step deprecation plan breaks on the first edge:

- v0.9 breaks by default, but supports `COMPILE_SYSTEM_EDL` for devs to quickly get back to working builds while they deal with the long term switch over.
- v0.10 ends the grace period by removing the fallback `COMPILE_SYSTEM_EDL` mechanism entirely. 

I would suggest doing that if possible. If not, and it looks like internally we're not ready to make the default switch ourselves, I'm suggesting that this could benefit from the usual deprecation pattern:

- v0.9 does not break by default, supports `COMPILE_SYSTEM_EDL` for devs to experiment with importing EDLs manually.
- v0.10 breaks by default, but continues to support `COMPILE_SYSTEM_EDL` for devs to quickly get back to working builds while they deal with the long term switch over.
- v0.11 ends the grace period by removing the fallback `COMPILE_SYSTEM_EDL` mechanism entirely.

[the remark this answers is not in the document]
jhand2: Oh, that was not my intention. My intention for v0.10 was to do as you suggest and just change the default for `COMPILE_SYSTEM_EDL` from `ON` to `OFF`. I'll update the changelog to make that more clear.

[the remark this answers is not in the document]
jhand2: Updated CHANGELOG

CHANGELOG.md
```diff
@@ -35,6 +39,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- In the next release (v0.10), system EDL will no longer be compiled int OE
```
CodeMonkeyLeet, line 43:
```suggestion
- In the next release (v0.10), system EDL will no longer be compiled into OE
```

CHANGELOG.md
```diff
@@ -35,6 +39,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 https://tls.mbed.org/tech-updates/releases/mbedtls-2.16.5-and-2.7.14-released)
 release notes for the set of issues addressed.
 
+### Deprecated
+- In the next release (v0.10), system EDL will no longer be compiled into OE
+  libraries by default (COMPILE_SYSTEM_EDL will be OFF by default). See the
+  [system EDL opt-in document]
+  (docs/DesignDocs/system_ocall_opt_in.md#how-to-port-your-application) for
+  more details on how to rebuild the SDK to match this behavior and for
+  guidance on porting your application to the new model.
```
dthaler, line 48: Do we have a GitHub issue filed to track porting all samples over to the new model?

tests/exclude_system_edl/README.md
```diff
@@ -0,0 +1,7 @@
+# Exclude System EDL test
+
+This is a temporary test to ensure that an enclave application builds properly
+when the SDK is compiled with `-DCOMPILE_SYSTEM_EDL=OFF`.
+
+In the future all other tests should be updated to import system EDL in their own
```
dthaler, line 6: Is there a github issue # to track this?  Can we reference it in the comments?

CHANGELOG.md
jhand2: #2867

tests/exclude_system_edl/README.md
jhand2: #2866

PR #2820, opened 2020-04-09
Fix scripts/install-windows-prereqs.ps1 script

ionutbalutoiu:
* Fix scripts/install-windows-prereqs.ps1
  * Append the OpenSSL bin directory to the system PATH via the
    `-EnvironmentPath` parameter from the `Install-Tool` function.
    The logic behind `-EnvironmentPath` parameter makes sure that
    the PATH is not updated twice if the directory is already
    appended.
  * Remove redundant extra logic, from the `Install-OpenSSL`
    function, to append bin directory to system PATH. This also
    caused OpenSSL bin directory to be appended to the system PATH
    on every script execution. Therefore, it broke the idempotency
    of the entire script.
* Add missing error check for `.\InstallAzureDCAP.ps1`

Signed-off-by: Ionut Balutoiu ibalutoiu@cloudbasesolutions.com

### DIFF
--- a/scripts/install-windows-prereqs.ps1
+++ b/scripts/install-windows-prereqs.ps1
@@ -350,26 +350,11 @@ function Install-Git {
 }
 
 function Install-OpenSSL {
-    $installDir = $installDir = Join-Path $env:ProgramFiles "OpenSSL-Win64"
+    $installDir = Join-Path $env:ProgramFiles "OpenSSL-Win64"
     Install-Tool -InstallerPath $PACKAGES["openssl"]["local_file"] `
                  -InstallDirectory $installDir `
                  -ArgumentList @("/silent", "/eula=accept") `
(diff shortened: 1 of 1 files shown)


ionutbalutoiu:
While fixing the `install-windows-prereqs.ps1` script, I discovered that we are impacted by the following issues with the `InstallAzureDCAP.ps1`  script: https://github.com/microsoft/Azure-DCAP-Client/issues/95.

I already opened a PR to fix the `InstallAzure-DCAP.ps1` script: https://github.com/microsoft/Azure-DCAP-Client/pull/96.

kuanggang-git:

jhand2:
Failure due to failed docker login. Trying again

bors r+

jhand2:
Samples timed out on windows. Opening an issue for this and trying again

bors r+

PR #2825, opened 2020-04-09
Remove libsgx_urts as it is not needed and causes segfaults with clang-7

jhand2:
Remove libsgx_urts as it is not needed and causes segfaults with clang-7.


jhand2:
bors try

kuanggang-git:

PR #2830, opened 2020-04-10
Add Windows support for getnameinfo and getaddrinfo

dthaler:
getaddrinfo was partially supported already but did not correctly handle non-ASCII hostnames
(for background on Windows Unicode support, see the description of PR #2809).

Also fixes some bugs in the resolver test which previously assumed a machine was IPv4 only and so would fail on any machine that was IPv6-enabled, and also assumed that the oe_addrinfo layout in the OE EDL file matched the native addrinfo layout on the host OS and so would fail on any OS where they didn't exactly match.


dthaler:
bors try

dthaler:
bors try

kuanggang-git:

dthaler:
bors try

dthaler:
@BRMcLaren  Can you tell me why the samples test times out?  It looks like there's some 8-minute timeout (480 sec) somewhere that needs to be increased, but I'm not sure where.

dthaler:
bors retry

Issue #2802, opened 2020-04-07
snmalloc depends on symbols in libcxx


jhand2:
cc @achamayou 

When I compile OE with -DUSE_SNMALLOC=ON with clang-9, snmalloc appears to depend on `__cxa_guard_acquire` and `__cxa_guard_release` from libcxx. This leads to linker errors for enclaves which are not linked with libcxx. I am currently unable to test this with a different compiler to see if this issue is compiler-specific.

Has anyone else seen this/can anyone repro with a different compiler?

jhand2:
Confirmed also happens with gcc 9

jhand2:
Additional context @anakrish pointed me to a comment on the original snmalloc PR from @mjp41

>> Does snmalloc use thread-local variables? If so, it might make sense to use a thread-local variable to hold to to the thread's allocator instead of adding a member to td_t.
>
> To do this, would include more C++ features, so would be harder to support inside oecore. For the current implementation for OE has no dependency on a C++ library at runtime. This was the main reason for modifying td_t. Outside the enclave, we have used C++ thread_static variables, but this requires runtime links to
> __cxa_thread_atexit
> __cxa_guard_acquire
>  __cxa_guard_release
>
>This is not something you probably want to support in oecore.

anakrish:
@jhand2 Thanks for creating this. Fixing this will allow us to consider using `snmalloc` as potential default allocator.

It might make sense to create an issue in the snmalloc repo:
https://github.com/microsoft/snmalloc
I created couple of issues there and they got addressed pretty quickly.

jhand2:
Ok I can open an issue on the snmalloc repo. I do want to understand what exactly is causing the usage. First, I can't find anything in C++ called `thread_static`. Does this mean `static thread_local` instead?

anakrish:
I think these are the culprits:
https://github.com/microsoft/snmalloc/blob/74657d9dbc776a54f78b77896a9955ebce5c8076/src/ds/helpers.h#L24-L26

The compiler is required to make the construction of these static variables thread-safe.
```c
      static std::atomic_flag flag;
      static std::atomic<bool> initialised;
      static Object obj;
```

anakrish:
@jhand2 `thread_static` must have been a typo.

Yes, thread-local variables won't generate _cxa_guards.

mjp41:
[...]

PR #2831, opened 2020-04-10
Upgrade snmalloc to v0.4.0

achamayou:
This is primarily to pick up the fix for #2802, but v0.4.0 contains a number of other improvements as well: https://github.com/microsoft/snmalloc/releases/tag/v0.4.0

Checked builds passed with gcc-8, gcc-9, clang-7, clang-8, clang-9


achamayou:
bors try

anakrish:
@achamayou 
Thanks for fix the issue. We are getting very close to having snmalloc as a potential choice for default allocator.

There is just one more issue that was mentioned in comments in #2292 
snmalloc seems to be compiled using system-headers instead of oe library headers.
Can you see if that can be fixed too? Or would you rather that be addressed in a separate PR?

achamayou:
@anakrish ah I’d missed that, I’ll take a look tomorrow, it should be easy enough to fix.

achamayou:
@anakrish I think that's best done separately, because:

1. It isn't all that clear where the right place to get those includes is without adding a dependency I don't think we want. I've come up with this, but a review from someone who's more familiar with the build would be very helpful:

```
    set_source_files_properties(sgx/snmalloc/snmalloc_wrapper.cpp PROPERTIES COMPILE_FLAGS
        "-std=c++17 -mcx16 -fno-exceptions -nostdinc -nostdinc++ -I../3rdparty/libcxx/libcxx/include -Ioutput/include/openenclave/libc -I../include/openenclave/internal")
```

2. snmalloc's aal_x86.h includes some things that are missing from OE's libcxx headers, such as cpuid.h and emmintrin.h. They aren't used when snmalloc is configured to be used from OPEN_ENCLAVE, but the includes still happen, so I need to fix that.

So for now I suggest merging this to fix the gcc build at least, and sorting out the includes in a separate PR. Your input on 1. would be welcome for that.

achamayou:
@mikbras @CodeMonkeyLeet could you review please? The bors message is somewhat cryptic, but I guess it means at least one code owner should review?

PR #2836, opened 2020-04-10
Disable EC fixed key test to unblock WS2019 testing

CodeMonkeyLeet:
- The EC key with fixed value of (P256 group order - 1), which
  produces signatures that sometimes will not verify on WS2019.


CodeMonkeyLeet:
bors try

jhand2:
Appears there was trouble reaching github on the CI run.

bors try

PR #2837, opened 2020-04-10
Rename edl umbrella files to make more sense for public-facing imports

jhand2:
Now that OE EDL files will need to be imported by application EDL (pending #2816), there should be a clear path to enabling the same behavior of icluding all system EDL. This path is to include just the three ubrella EDL files (two for non-SGX).

Rename tee.edl -> core.edl and sgx/sgx.edl -> sgx/platform.edl. All the ecalls/ocalls included by sgx/platform.edl are needed for all SGX applications. Also add optee/platform.edl which contains no functions so that an application can support both optee and SGX by only changing build flags and no changes to the edl file.

A normal SGX application which wants to maintain parity with the old build system will now just add the following 3 lines to their edl file:

```
from "sgx/platform.edl" import *
from "core.edl" import *
from "syscall.edl" import *
```

Note that by default none of this will take effect until after v0.9. The -DCOMPILE_SYSTEM_EDL flag will be on by default in v0.9 and which means no changes will be needed to user EDL. This guidance becomes relevant for v0.10.


docs/DesignDocs/SGX_QuoteEx_Integration.md
```diff
@@ -65,7 +65,7 @@ The SGX ECDSA-p256 attester plugin is implemented in code file
 `enclave/sgx/attester.c` and other relevant enclave-side and host-side code files,
 called enclave-side and host-side plugin libraries in this document.
 he enclave-side plugin library interacts with the host-side plugin library
-via OCALLs defined in interface definition file `common/sgx/sgx.edl`.
+via OCALLs defined in interface definition file `common/sgx/platform.edl`.
```
jhand2, line 68: edl/sgx/platform.edl

docs/DesignDocs/SGX_QuoteEx_Integration.md
```diff
@@ -206,7 +206,7 @@ creates a list of attester plugins for them, and returns the created list
 to the caller.
 
 For SGX evidence formats enumeration, a new OCALL is added to interface
-definition file `common/sgx/sgx.edl` and implemented in the host-side
+definition file `common/sgx/platform.edl` and implemented in the host-side
```
jhand2, line 209: edl/sgx/platform.edl

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -135,5 +135,5 @@ in one of the following EDL files
 * `common/utsname.edl`
 * `common/syscall.edl`
 * `common/syscall.edl`
-* `common/sgx/sgx.edl`
+* `common/sgx/platform.edl`
```
jhand2, line 138: Change all the paths here

edl/oecore.edl
```diff
@@ -4,7 +4,7 @@
 /*
 **==============================================================================
 **
-** tee.edl:
+** oecore.edl:
```
jhand2, line 7: Maybe consider changing this to just core.edl. Then if the user's edl search path is <install_path>/include their import line would be: `from "openenclave/edl/core.edl" import *` which seems a little cleaner

anakrish:
Reading this PR in conjunction with #2186, I am a bit unclear about what the user experience is in v0.9.

Does the user have to make any changes to their existing EDL files or will they continue to work?

jhand2:
> Reading this PR in conjunction with #2186, I am a bit unclear about what the user experience is in v0.9.
> 
> Does the user have to make any changes to their existing EDL files or will they continue to work?

If the SDK is compiled with -DCOMPILE_SYSTEM_EDL=ON then no change for developers. This is the default and will be how the v0.9 package is built.

The changelog should be updated though to say that this flag will go away in v0.10 and then this guidance all becomes relevant.

anakrish:
Ok. In that case and given that today is the last day for code changes, does this PR have to go in for v0.9?
It seems there is no effect on user experience whether this PR goes in or not.

jhand2:
@radhikaj when are you planning to branch off for v0.9?

We at least need to make some sort of announcement for v0.9 that says system ecalls/ocalls will no longer be builtin to the SDK in v0.10.

Otherwise it doesn't really matter when this goes in from a release perspective since the package will be compiled the old way.

radhikaj:
> @radhikaj when are you planning to branch off for v0.9?
> 
> We at least need to make some sort of announcement for v0.9 that says system ecalls/ocalls will no longer be builtin to the SDK in v0.10.
> 
> Otherwise it doesn't really matter when this goes in from a release perspective since the package will be compiled the old way.

Planning to branch off on Tuesday. Would like to take this for v0.9..

jhand2:
As far as review burden, this PR is only cosmetic (I wrote entirely using find/replace). #2816 will need to change a little bit once this goes in. If this goes in today or tomorrow I can have #2816 ready for merging by Monday.

bors try

jhand2:
As an aside given the way we do release, there is effectively no such thing as code freeze until the release branch is created. That creates the least amount of friction for new changes going into master.

edl/sgx/sgx_attestation.edl
```diff
@@ -4,7 +4,7 @@
 /*
 **==============================================================================
 **
-** sgx/attestation.edl:
+** sgx/sgx_attestation.edl:
```
CodeMonkeyLeet, line 7: Does EDL actually care about this? we don't qualify other files in this folder the same way, and I suspect that the edl/attestation.edl file will be the one going away in the long run.

jhand2: It depends on how you set your edl include path in oeedger8r. If you pass both `edl` and `edl/sgx` as include paths, then there will be a conflict (attestation.edl exists twice) and I believe oeedger8r will use the second one. I figured it was easier to just rename it to avoid conflict. Especially since as you say this edl file is likely to go away in the future.

CodeMonkeyLeet: Actually this EDL file (edl/sgx/attestation.edl) is likely to hang around because of `oe_get_quote_verification_collateral_ocall()`. My comment is that edl/attestation.edl is likely to go away because it services the host-side call-in to `oe_get_report()`, which should ultimately go away in favor of the pluggable attestation APIs.

Issue #2668, opened 2020-03-12
fs and hostfs test cases rely on additional cmake tests to remove temporary directory


mingweishih:
Currently, both fs and hostfs tests (parts of syscall tests) rely on additional cmake tests to remove the temporary directory.
Take the `tests/syscall/hostfs` for example, currently it does:
```
set(TMP_DIR "${CMAKE_CURRENT_BINARY_DIR}/tmp")

add_test(tests/fs1 cmake -E remove_directory "${TMP_DIR}")

add_enclave_test(tests/fs2
    fs_host
    fs_enc
    "${PROJECT_SOURCE_DIR}"
    "${TMP_DIR}")
```

I suggest to removing such patterns and make the test self-contained.

Issue #2833, opened 2020-04-10
Allow simple preprocessor directives in  EDL


jhand2:
I mostly want to gauge if people think this is a good idea or not. Feel give reasons we shouldn't do this.

We could allow simple C preprocessor directives in EDL. In particular, I'm thinking `#ifdef` and `#ifndef`. This would also mean adding an option (such as `--define`) to oeedger8r. I mostly think this is useful for features which are only used on a single TEE. For example:

```
// myapp.edl
#ifdef PLATFORM_SGX
    from "switchless.edl" import *;
#endif
```

This is not currently possible. If your application wanted to use an SGX-specific feature only when on SGX, you would need a different EDL file for each TEE you want your app to run on.

dthaler:
I would love to see this, it was painful to use EDLs when we first started because of this lack of capability.

jhand2:
I see a couple ways to do this:
1. Re purpose the ocaml C preprocessor code into a library: https://github.com/ocaml-community/cppo
2. Invoke the C preprocessor via command-line

2 is probably easier but 1 may be easier in terms of dependency management.

@anakrish do you have an opinion of this?

jhand2:
@anakrish I have started implementing the 2nd option in oeedger8r-cpp, should have a PR ready soon. On Linux it will use the cpp cmdline tool, and on Windows it will use cl.exe

Do you have any concerns with this approach or do you think this is a fine way to handle this?

anakrish:
Thanks for getting a head start on this.
I don't see any concern. We just need to define the preprocessor capabilities.

Note: Intel edger8r does use the C compiler as an edl preprocessor:
https://github.com/intel/linux-sgx/blob/fdc9b33ee173a5e7743f20a2a9cb1dd8db114dca/sdk/edger8r/linux/Preprocessor.ml#L58
```ocaml
let command = sprintf "gcc -x c -E -P \"%s\" 2>/dev/null" full_path in
```
However, this results in incorrect locations for error messages, for which I created an issue:
https://github.com/intel/linux-sgx/issues/319

Essentially, we need to change the lexer to process the `#line` directives emitted by the preprocessor to preserve the original line numbers.

jhand2:
Support for this is added in https://github.com/openenclave/oeedger8r-cpp/pull/9

When this PR is merged and oeedger8r-cpp is made the default this issue will be resolved

mingweishih:
[...]

PR #2839, opened 2020-04-10
Reconfigure the fs test.

nonpolarity:
To enable unit test of cross-compiled enclave, the OS-specified tests must be
configurable at the ctest level. These sub-tests including test_pread_file()
and test_pwrite_file() are move to a separate host function test_fs_linux. 

#2668 is partially fixed.


nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

tests/syscall/fs_linux/fs.edl
```diff
@@ -3,6 +3,8 @@
 
 enclave {
 
+    //from "sgxfs.edl" import *;
```
mingweishih, line 6: Remove this commented line.

CodeMonkeyLeet, line 6: Please remove unnecessary import. 

Also, can this file be removed entirely? It's exactly the same signature as provided by the parent and I'm trying to figure out if there's a way to reduce code duplication.

tests/syscall/fs_linux/host/host.c
```diff
@@ -0,0 +1,103 @@
...
+
+    r = oe_terminate_enclave(enclave);
+    OE_TEST(r == OE_OK);
+
+    printf("=== passed all tests (hostfs)\n");
+
+    return 0;
+}
+
+#if defined(_WIN32)
```
dthaler, line 50: I don't follow why there is _WIN32 section in a file under fs_linux.   What does fs_linux mean?
(Answering Simon's question would likely answer mine as well.)

tests/syscall/fs_linux/enc/file_system.h
```diff
@@ -0,0 +1,578 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
```
dthaler, line 3: I can't tell what, if anything, in this file is actually Linux specific, nor if anything is SGX-specific, even though I see "linux" in the directory name this file is under, and SGX in the content.  I also agree with Simon's comments.

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

tests/syscall/fs_linux/host/host.c.bak
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) Open Enclave SDK contributors.
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks!

tests/syscall/fs_linux/fs.edl
nonpolarity: Fixed, thanks!

nonpolarity: Fixed, thanks!

tests/syscall/fs/enc/enc.cpp
```diff
@@ -0,0 +1,246 @@
...
+template <class FILE_SYSTEM>
+static void test_rename_file(FILE_SYSTEM& fs, const char* tmp_dir)
+{
+    char oldname[OE_PAGE_SIZE];
+    char newname[OE_PAGE_SIZE];
+    typename FILE_SYSTEM::stat_type buf;
+
+    printf("--- %s()\n", __FUNCTION__);
+
+    mkpath(oldname, tmp_dir, "alphabet.linked"),
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks!

tests/syscall/fs_linux/host/host.c
nonpolarity: Fixed with Simon's suggestion, thanks.

tests/syscall/fs_linux/enc/file_system.h
nonpolarity: Fixed with Simon's suggestion, thanks.

nonpolarity:
@CodeMonkeyLeet Thanks for your suggestion, now no extra test, and pread/pwrite test are in test_linux_fs.

tests/syscall/fs/host/host.c
```diff
@@ -3,59 +3,70 @@
...
 #define SKIP_RETURN_CODE 2
 
-int rmdir(const char* path);
-
-int test_fs_posix(
-    const char* enclave_path,
-    const char* src_dir,
-    const char* tmp_dir)
+#if defined(_WIN32)
+int rmdir(wchar_t* path)
```
CodeMonkeyLeet, line 21: Nice! It might be useful to name it something that indicates its function like `recursive_rmdir()` to differentiate its behavior from the standard `rmdir()`

We might also want to consider refactoring this out for other syscall tests to use (notably hostfs, which as the same structure) in a later PR.

tests/syscall/fs/enc/enc.cpp
```diff
@@ -418,6 +415,34 @@ static void test_invalid_path(FILE_SYSTEM& fs)
     OE_TEST(fs.truncate(path, 0) == -1);
 }
 
+template <class FILE_SYSTEM>
+void test_common(FILE_SYSTEM& fs, const char* tmp_dir)
+{
+    cleanup(fs, tmp_dir);
+    test_create_file(fs, tmp_dir);
+    test_read_file(fs, tmp_dir);
+    //    test_pread_file(fs, tmp_dir);
```
mingweishih, line 424: Should this line (and the following one) be removed?

tests/syscall/fs/host/host.c
```diff
@@ -3,59 +3,70 @@
...
+
+    doublenullpath = malloc((len + 2) * sizeof(wchar_t));
+    if (!doublenullpath)
     {
-        printf("=== Skipped unsupported test in simulation mode (sealKey)\n");
-        return SKIP_RETURN_CODE;
+        goto done;
     }
+    memcpy(doublenullpath, path, len * sizeof(wchar_t));
+    doublenullpath[len] = doublenullpath[len + 1] = L'\0';
```
CodeMonkeyLeet, line 34: Can you explain why this double-null path is needed? It has not been my experience that any of the Win32 wchar APIs require this and the earlier use of `wcslen` already assumes that the input `path` is null-terminated.

tests/syscall/fs/host/host.c
```diff
@@ -78,26 +89,105 @@ int wmain(int argc, wchar_t* argv[])
...
+            rmdir(abs_path);
+        }
+        else if ((file = fopen(abs_path, "r")))
+        {
+            fclose(file);
+            remove(abs_path);
+        }
+    }
+
+    result = remove(path);
```
CodeMonkeyLeet, line 141: Are we missing a `closedir()` for the initial open of the root `path` before removing it?

tests/syscall/fs/host/host.c
```diff
@@ -78,26 +89,105 @@ int wmain(int argc, wchar_t* argv[])
...
+    char abs_path[PATH_MAX];
+    int result = -1;
+    struct dirent* entry = NULL;
+    DIR* dir = NULL;
+    if (!(dir = opendir(path)))
+    {
+        goto done;
+    }
+
+    while ((entry = readdir(dir)))
```
CodeMonkeyLeet, line 118: Non-blocking:

@jhand2 Can you advise on an idiomatic way to do this in linux? A couple of suggestions based on my limited exposure:

* Should we be using `nftw()` here to skip navigating links during the directory walk?
* Is it a better practice to use `stat()` and `S_ISREG`/`S_ISDIR` to figure out the type of the item, instead of opening and closing it?

tests/syscall/fs/enc/enc.cpp
nonpolarity: Yes must be. Thanks!

tests/syscall/fs/CMakeLists.txt
```diff
@@ -9,18 +9,15 @@ endif()
...
 
-set_enclave_tests_properties(tests/fs2 PROPERTIES SKIP_RETURN_CODE 2)
+set_enclave_tests_properties(tests/fs PROPERTIES SKIP_RETURN_CODE 2)
 
 # Ensuring removing the tmp directory before the second test run.
 if (LVI_MITIGATION MATCHES ControlFlow AND NOT LVI_MITIGATION_SKIP_TESTS)
-    add_test(tests/fs1-lvi-cfg cmake -E remove_directory "${TMP_DIR}")
-    set_tests_properties(tests/fs2-lvi-cfg PROPERTIES DEPENDS tests/fs1-lvi-cfg)
+    set_tests_properties(tests/fs-lvi-cfg PROPERTIES DEPENDS tests/fs-lvi-cfg)
 endif()
```
jhand2, line 23: As Simon said above, I think you can get rid of this

tests/syscall/fs/CMakeLists.txt
```diff
@@ -9,18 +9,15 @@ endif()
 
 set(TMP_DIR "${CMAKE_CURRENT_BINARY_DIR}/tmp")
 
-add_test(tests/fs1 cmake -E remove_directory "${TMP_DIR}")
-
-add_enclave_test(tests/fs2
+add_enclave_test(tests/fs
```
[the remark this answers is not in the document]
mingweishih: Yes @CodeMonkeyLeet is correct. Thanks for catching this.
@nonpolarity please remove line 20 - 23 as we no longer need addition tests to delete directory.

tests/syscall/fs/enc/enc.cpp
```diff
@@ -775,6 +798,84 @@ void test_fs(const char* src_dir, const char* tmp_dir)
...
+
+    /* Test oe_set_thread_devid() */
+    {
+        printf("=== testing oe_set_thread_devid:\n");
+
+        fd_file_system fs;
+        device_registrant reg(OE_DEVID_HOST_FILE_SYSTEM);
+        test_pio(fs, tmp_dir);
+    }
+}
```
jhand2, line 878: nit: space after function

tests/syscall/fs/CMakeLists.txt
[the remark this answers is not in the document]
mingweishih: Can we do the same in the hostfs test case?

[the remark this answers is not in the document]
mingweishih: According to #2668, both fs and hostfs have the same issue. However, it seems the PR only addresses the issue in fs.

I'd suggest either extend the fix to hostfs or remove the fixes keyword and leave comment in the #2668 states that this PR partially fixes the issue.

radhikaj:
bors try

tests/syscall/fs/host/host.c
nonpolarity: Yes nftw is a better tool to avoid trivial file operations.

jhand2: I would use nftw and then call the remove syscall (https://linux.die.net/man/3/remove) in the callback which will call `unlink` for files and `rmdir` for directories.

nftw does a DFS when passed `FTW_DEPTH` so by the time you remove the directory it should be empty.

This is untested but it would be something like:

```
int nftw_callback(const char *path, const struct stat *sb, int flags, struct FTW *ftwbuf)
{
    int ret = remove(path);
    if (ret != 0)
        fprintf(stderr, "Could not remove file %s\n", path);

    return ret;
}

int rmdir_recusrive(char *path)
{
    // FTW_PHYS won't follow symlinks. This is mostly because following symlinks is
    // scary and you could end up destroying the filesystem
    return nftw(path, nftw_callback, 64 /* depth limit */, FTW_DEPTH | FTW_PHYS);
}
```

tests/syscall/fs/host/host.c
```diff
@@ -78,26 +88,105 @@ int wmain(int argc, wchar_t* argv[])
...
+
+    r = oe_terminate_enclave(enclave);
+    OE_TEST(r == OE_OK);
+
+    printf("=== passed all tests (hostfs)\n");
+
+    return 0;
 }
+#else
+int rmdir(const char* path)
```
jhand2, line 107: I would name this something else to avoid confusion with the linux syscall name

tests/syscall/fs/CMakeLists.txt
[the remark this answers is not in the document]
nonpolarity: Yes lines 20-23 are safe to remove now.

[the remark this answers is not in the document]
nonpolarity: @mingweishih you are right hostfs is not fixed yet.

tests/syscall/fs/enc/enc.cpp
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks!

tests/syscall/fs/host/host.c
nonpolarity: Yes double-null string is a requirement for this operation, someone explain here https://devblogs.microsoft.com/oldnewthing/20100218-00/?p=14893. But you are right wcslen already includes one null, I will check if one extra is good.

CodeMonkeyLeet: Thanks for the pointer! https://docs.microsoft.com/en-us/windows/win32/api/shellapi/ns-shellapi-shfileopstructw does indeed call out the need for a double null-terminated string

tests/syscall/fs/enc/enc.cpp
```diff
@@ -775,6 +798,84 @@ void test_fs(const char* src_dir, const char* tmp_dir)
...
+    (void)src_dir;
+
+    OE_TEST(oe_load_module_host_file_system() == OE_OK);
+#if defined(TEST_SGXFS)
+    OE_TEST(oe_load_module_sgx_file_system() == OE_OK);
+#endif
+
+    OE_TEST(oe_mkdir_d(OE_DEVID_HOST_FILE_SYSTEM, tmp_dir, 0777) == 0);
+
+    printf("=== running Linux-specified tests\n");
```
dthaler, line 812:
```suggestion
    printf("=== running Linux-specific tests\n");
```

tests/syscall/fs/host/host.c
```diff
@@ -14,48 +17,55 @@
 
 #define SKIP_RETURN_CODE 2
 
-int rmdir(const char* path);
-
-int test_fs_posix(
-    const char* enclave_path,
-    const char* src_dir,
-    const char* tmp_dir)
+#if defined(_WIN32)
```
dthaler, line 20: Since this file now has approximately 0 lines of code in common between Windows and Linux, I might suggest splitting it into two files, like host_win32.c and host_linux.c for example.  One benefit is that it then becomes somewhat easier to do a diff between the two files to see what the platform-specific differences are, and to help verify that they both do the same thing.

nonpolarity:
bors try

tests/syscall/fs/CMakeLists.txt
nonpolarity: Fixed, thanks!

tests/syscall/fs/enc/enc.cpp
nonpolarity: Fixed, thanks!

tests/syscall/fs/host/host.c
nonpolarity: Fixed, thanks!

tests/syscall/fs/enc/enc.cpp
nonpolarity: Fixed, thanks!

tests/syscall/fs/host/host.c
nonpolarity: Yes two files and no macros.

nonpolarity:
Thanks everyone! To merge this pr earlier for nightly test, the hostfs problem is left unfixed.

tests/syscall/fs/host/host.c
nonpolarity: Fixed with nftw.

tests/syscall/fs/host/host.c
nonpolarity: Fixed, thanks!

tests/syscall/fs/host/host.c
nonpolarity: Fixed, thanks!

BRMcLaren:
@nonpolarity thanks for uptaking the burden of helping re-stabilize the CI/CD infrastructure!

BRMcLaren:
@nonpolarity I don't want to disenfranchise anyone's opinions or step on their toes, however having the environment in an unstable state is a concern at this time. I am approving the PR, could we please open a GH issue for any remaining concerns?

nonpolarity:
@BRMcLaren Thanks! I will follow up the unsolved problem.

tests/syscall/fs/enc/enc.cpp
```diff
@@ -775,6 +798,84 @@ void test_fs(const char* src_dir, const char* tmp_dir)
     test_dup_case2(tmp_dir);
 }
 
+void test_fs_linux(const char* src_dir, const char* tmp_dir)
```
dthaler, line 801: Should this be inside an ifdef if it's not needed on WIN32?

tests/syscall/fs/fs.edl
```diff
@@ -3,12 +3,12 @@
...
 enclave {
 
-    //from "sgxfs.edl" import *;
-
     trusted {
         public void test_fs(
             [string, in] const char* src_dir,
             [string, in] const char* tmp_dir);
-
+        public void test_fs_linux(
```
dthaler, line 10: I see this was added to an EDL that was moved out of the linux subdirectory, but this is linux specific isn't it?
Until we have preprocessor directives as in issue #2833, we may want to keep separate EDL files.
@jhand2 do you have a recommendation here?

tests/syscall/fs/enc/enc.cpp
nonpolarity: edger8r does not yet support macro.

tests/syscall/fs/fs.edl
nonpolarity: Yes this is linux specific. To match the host side we should use two separate edl files, but enclave side it is hard to make corresponding functions linux specific. Maybe I can update in next pr.

tests/syscall/fs/enc/enc.cpp
dthaler: Right, so I would recommend the API to be in a Linux-only EDL so it can be ifdef'ed out on WIN32.

PR #2841, opened 2020-04-13
Emulate wrfsbase instruction

anakrish:
Emulate wrfsbase instruction.

### DIFF
--- a/enclave/core/sgx/exception.c
+++ b/enclave/core/sgx/exception.c
@@ -204,6 +204,61 @@ static struct
     {19, OE_EXCEPTION_SIMD_FLOAT_POINT},
 };
 
+static int _emulate_wrfsbase(sgx_ssa_gpr_t* ssa_gpr)
+{
+    // Emulate wrfsbase
+    const uint32_t OE_WRFSBASE_PREFIX_1 = 0xae0f48f3;
+    if (*((uint32_t*)ssa_gpr->rip) == OE_WRFSBASE_PREFIX_1)
+    {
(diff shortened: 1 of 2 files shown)


enclave/core/sgx/exception.c
```diff
@@ -219,8 +219,61 @@ int _emulate_illegal_instruction(sgx_ssa_gpr_t* ssa_gpr)
...
+        };
+
+        uint8_t last_byte = ((uint8_t*)ssa_gpr->rip)[4];
+        uint8_t idx = last_byte - 0xd0;
+        if (idx >= sizeof(regs) / sizeof(regs[0]))
+            return -1;
+
+        ssa_gpr->fs_base = regs[idx];
+        ssa_gpr->rip += 5;
+        return 0;
```
jhand2, line 276: Nit: this code is the same as the above code right? Can this be factored into `oe_emulate_wrfsbase()`. Also would rdfsbase also need to be emulated?

Feel free to ignore until we decide if the approach will be accepted or not.

anakrish: The registers are different. If found this to be more readable than if-else.

anakrish: We could emulate `rdfsbase` if needed. It is possible.

jhand2: Oh I see. That's fine then.

anakrish:
bors try

jhand2:
Test failures are expectd

Issue #2852, opened 2020-04-14
pre-commit hook causes problems on Windows


dthaler:
CMakeLists.txt currently installs the pre-commit hook if (NOT WIN32), and if installed anyway on Windows, it causes problems because check-license fails, preventing any commit from succeeding:
```
C:\git\dthaler\openenclave>git commit
head: cannot open 'C:/git/dthaler/openenclave//c/git/dthaler/openenclave/tests/syscall/socket/enc/enc.c' for reading: No such file or directory
```
That's because check-license assumes that `git rev-parse --show-toplevel' will report a prefix of 'realfile $f`, whereas on Windows the former is a windows style path like C:/git/dthaler/openenclave whereas the latter has a unix style path like /c/git/dthaler/openenclave.

So the two real problems this issue tracks is:
1) The checks aren't run on windows, leading to failures in bors try whenever someone doesn't manually run format-code, etc.   This results in taking longer to merge code due to multiple iterations to address, greater load on the CI/CD infrastructure, and hence longer queue times for other code too.
2) If you do a cmake in a WSL window, it breaks Windows because WSL and Windows share the same directory, so the pre-commit hook will be installed for Windows, which then breaks the ability to do any commits from Windows.  (I've been trying to track down what broke me for the last day or so, and finally got to the root cause.)

I think the proper fix is to first fix the check-license script, and then enable the pre-commit hook on Windows.

dthaler:
This is now fixed, so closing

PR #2854, opened 2020-04-14
Enable pre-commit checks on Windows

dthaler:
Addresses issue #2852

The path issue and solution are discussed at https://stackoverflow.com/questions/50579937/retrieve-the-path-of-a-repo-from-the-git-bash-from-windows-with-right-format/50580300#50580300

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,11 +62,9 @@ if (IS_DIRECTORY "${PROJECT_SOURCE_DIR}/.git")
     OUTPUT_STRIP_TRAILING_WHITESPACE)
 
   # Install Git pre-commit hook
-  if (NOT WIN32)
-    file(
-      COPY scripts/pre-commit scripts/commit-msg
-      DESTINATION "${PROJECT_SOURCE_DIR}/.git/hooks")
-  endif ()
+  file(
+    COPY scripts/pre-commit scripts/commit-msg
+    DESTINATION "${PROJECT_SOURCE_DIR}/.git/hooks")
 endif ()
 
 # Generates `compile_commands.json` used by some developers. Only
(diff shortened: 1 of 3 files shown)


HernanGatta:
If `shellcheck` is now a pre-requisite for Windows as well, it should be added to the [pre-requisites install script](https://github.com/dthaler/openenclave/blob/pre-commit/scripts/install-windows-prereqs.ps1), too.

CMakeLists.txt
```diff
@@ -62,11 +62,9 @@ if (IS_DIRECTORY "${PROJECT_SOURCE_DIR}/.git")
...
 
   # Install Git pre-commit hook
-  if (NOT WIN32)
-    file(
-      COPY scripts/pre-commit scripts/commit-msg
-      DESTINATION "${PROJECT_SOURCE_DIR}/.git/hooks")
-  endif ()
+  file(
+    COPY scripts/pre-commit scripts/commit-msg
+    DESTINATION "${PROJECT_SOURCE_DIR}/.git/hooks")
```
HernanGatta, line 67: It seems odd to me to do this in CMake. If the `pre-commit` hook now runs on Windows as well, why not install the hooks by default?

jhand2: Is there such thing as hooks which are installed by default? I thought they are local to the repository so they always need to be set up either manually or by some script.

HernanGatta: Looking carefully at the Git documentation, there's a note about this which I missed on my previous read:

> It’s important to note that client-side hooks are not copied when you clone a repository. If your intent with these scripts is to enforce a policy, you’ll probably want to do that on the server side [...]

[Source](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks)

So it looks like you are correct.

Issue #1881, opened 2019-06-06
Support (and test) dup'ing of directories.


mikbras:
Calling dup() on a directory descriptor does not seem to be supported in the POSIX extensions.

yakman2020:
Need a test to demonstrate the problem.

PR #2855, opened 2020-04-14
Add standard gai_strerror() API support

dthaler:
This is a standard resolver API that was previously missing from the OE SDK.

API references:
* https://pubs.opengroup.org/onlinepubs/009695399/functions/gai_strerror.html
* https://tools.ietf.org/html/rfc3493
* https://linux.die.net/man/3/gai_strerror
* https://docs.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-gai_strerrorw

### DIFF
--- a/include/openenclave/corelibc/netdb.h
+++ b/include/openenclave/corelibc/netdb.h
@@ -1,0 +1,46 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+/* DISCLAIMER:
+ * This header is published with no guarantees of stability and is not part
+ * of the Open Enclave public API surface. It is only intended to be used
+ * internally by the OE runtime. */
+
+#ifndef _OE_NETDB_H
+#define _OE_NETDB_H
+
+#include <openenclave/internal/syscall/netdb.h>
+
+OE_EXTERNC_BEGIN
+
+/*
+**==============================================================================
+**
+** OE names:
+**
+**==============================================================================
+*/
+
+const char* oe_gai_strerror(int errnum);
+
+/*
+**==============================================================================
+**
+** Standard-C names:
+**
+**==============================================================================
+*/
+
+#if defined(OE_NEED_STDC_NAMES)
+
+OE_INLINE
+const char* gai_strerror(int errnum)
+{
+    return oe_gai_strerror(errnum);
+}
+
+#endif /* defined(OE_NEED_STDC_NAMES) */
+
+OE_EXTERNC_END
+
+#endif /* _OE_NETDB_H */
--- a/enclave/core/CMakeLists.txt
+++ b/enclave/core/CMakeLists.txt
@@ -184,6 +184,7 @@ add_enclave_library(oecore STATIC
     ctype.c
     debugmalloc.c
     errno.c
+    gai_strerror.c
     gmtime.c
     hexdump.c
     hostcalls.c
--- a/enclave/core/gai_strerror.c
+++ b/enclave/core/gai_strerror.c
@@ -1,0 +1,48 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/corelibc/netdb.h>
+
+typedef struct _error_info
+{
+    int errnum;
+    const char* message;
+} error_info_t;
+
+// The following messages are copied from
+// ../../3rdparty/musl/musl/src/network/gai_strerror.c
+//
+// Unlike MUSL's implementation of strerror() which puts all strings
+// in a header file we can include, MUSL's implementation of gai_strerror()
+// instead puts the strings into the .c file itself along with code we
+// don't want.  As such, we have to copy the strings into this file rather
+// than being able to include them from MUSL.
+static error_info_t _errors[] = {
+    {OE_EAI_BADFLAGS, "Invalid flags"},
+    {OE_EAI_NONAME, "Name does not resolve"},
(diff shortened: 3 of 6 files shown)


jhand2:
Could this be implemented inside the enclave with `3rdparty/musl/musl/src/network/gai_strerror.c`?

dthaler:
> Could this be implemented inside the enclave with 3rdparty/musl/musl/src/network/gai_strerror.c?

That's a great question actually.   The messages are in the current locale of the user or machine, which doing an OCALL will get.  If the strings were compiled into the enclave, then we'd have to deal with localization inside the OE SDK and have multiple language packs, etc.  (On a German machine, the messages should be in German, on a Russian machine, they should be in Russian, etc.)  In my view, it's likely much cleaner to get the string from an OCALL and not deal with language issues in the enclave.   Let me know if you think otherwise.

jhand2:
> > Could this be implemented inside the enclave with 3rdparty/musl/musl/src/network/gai_strerror.c?
> 
> That's a great question actually. The messages are in the current locale of the user or machine, which doing an OCALL will get. If the strings were compiled into the enclave, then we'd have to deal with localization inside the OE SDK and have multiple language packs, etc. (On a German machine, the messages should be in German, on a Russian machine, they should be in Russian, etc.) In my view, it's likely much cleaner to get the string from an OCALL and not deal with language issues in the enclave. Let me know if you think otherwise.

@CodeMonkeyLeet do you know if we rely on localized strings anywhere else in oelibc? Or do we just avoid all those features on purpose?

CodeMonkeyLeet:
> @CodeMonkeyLeet do you know if we rely on localized strings anywhere else in oelibc? Or do we just avoid all those features on purpose?

OE SDK in the enclave libraries explicitly do not support locale handling today. For the subset of C/C++ functions supported in the enclave, they are documented as only supporting the C/POSIX locale.

dthaler:
bors try

dthaler:
bors try

jhand2:
@dthaler sorry about the build output. I still don't know why it's not showing up. It seems to show up for some runs but not others.

Anyway here is the error:
```
../host/linux/syscall.c:864:8: error: conflicting types for '_strcpy_to_utf8'

 size_t _strcpy_to_utf8(

        ^

In file included from ../host/linux/syscall.c:24:0:

../host/linux/../../common/oe_host_socket.h:46:8: note: previous declaration of '_strcpy_to_utf8' was here

 size_t _strcpy_to_utf8(

        ^

```

dthaler:
bors try

dthaler:
> OE SDK in the enclave libraries explicitly do not support locale handling today. For the subset of C/C++ functions supported in the enclave, they are documented as only supporting the C/POSIX locale.

I filed issue #1881 to track the issue of other parts of OE (strerror in particular) not being POSIX compliant today, since POSIX requires them to be returned in the _current_ locale, not the C/POSIX locale.

jhand2:
Note build failure is an issue with Windows Server 2019. We are investigating an looking for a fix.

radhikaj:
bors try

jhand2:
@dthaler it looks like the Jenkins UI bug is fixed. You can view failures by clicking the link under "Triggered builds"

dthaler:
The failure seems to be with attested_tls hitting OE_PLATFORM_ERROR generating the certificate and private key which seems to be unrelated to these changes.  Is that a separate known issue?

dthaler:
@radhikaj you were looking at attested_tls failures, do you recognize this error?

dthaler:
bors retry

jhand2:
I know that the decision to use C locale was intentional at the time, but I don't have a good handle on the pros and cons. I wonder if we should come to a consensus on https://github.com/openenclave/openenclave/issues/2881 before moving forward.

That said I would also be ok just merging this PR and coming back to #2881. Anyone else have an opinion?

dthaler:
bors try

dthaler:
@jhand2 @CodeMonkeyLeet, I redid this PR as discussed in meeting, can you re-review?

include/openenclave/corelibc/netdb.h
```diff
@@ -0,0 +1,46 @@
...
+**
+** Standard-C names:
+**
+**==============================================================================
+*/
+
+#if defined(OE_NEED_STDC_NAMES)
+
+OE_INLINE
+const char* gai_strerror(int errnum)
```
jhand2, line 37: Does this need to be part of corelibc? The other socket functionality it part of oelibc (see the `libc/` directory).

I would recommend adding the function at that level and just compiling in the source file from musl. You can follow the same model that is used for compiling socket.c in `libc/CMakeLists.txt`

dthaler: Is there a design doc somewhere that explains the difference between oelibc and corelibc?   (I can guess, but is it documented somewhere?)

jhand2: I don't think there is.

Basically there used to only be corelibc and the goal was to implement a minimal libc there. Then it it was decided that that was challenging, not worth the trouble, and using an existing libc implementation was better. So musl was added. But now oecore depends on some basic stdc functions (mostly string and memory related) so corelibc couldn't be entirely removed.

So for the most part new libc functionality all goes in oelibc. There is also probably a lot of things in corelibc that aren't needed anymore and need to be cleaned up.

dthaler: Filed issue #2907.  I did some more testing, and gai_strerror is already part of oelibc so I will abandon this PR.

PR #2826, opened 2020-04-09
Fix cpio cmake config that only to build if BUILD_ENCLAVES.

nonpolarity:
Change cpio cmake config that only to build if BUILD_ENCLAVES, to avoid unexpected cpio build error while enclaves are cross-compiled.

PR #2868, opened 2020-04-15
Fix socketpair/CMakeLists.txt and resolver/CMakeLists.txt.

nonpolarity:
A following up pr from #2826 
Change cmake config that only to build if BUILD_ENCLAVES, to avoid unexpected build error while enclaves are cross-compiled.


nonpolarity:
bors try

radhikaj:
Thanks @nonpolarity

PR #2870, opened 2020-04-15
Fix compile errors on gcc-8

anakrish:


anakrish:
These changes just missed the last merge.

jhand2:
FYI I'm going to squash this with the last commit. I want to keep commits in the branch self contained so they're easier to move around

PR #2871, opened 2020-04-15
Update mbedTLS to v2.16.6

CodeMonkeyLeet:


CodeMonkeyLeet:
bors try

jhand2:
bors try

jhand2:
Windows CI bug where all tests have return code 0xc0000139. I think this points to file corruption. Could be that VM was in a bad state? trying again

bors try

CodeMonkeyLeet:
@jhand2 The 0xc0000139 error in ctests is due to not having the appropriate version of the LC drivers we first saw on WS2019 VMs. The rerun is unlikely to resolve the issue until we root cause the driver problem.

BRMcLaren:
bors try

PR #2882, opened 2020-04-16
Dynamically load Intel SGX Enclave Common shared library

anakrish:
This eliminates that the host application had on Intel SGX Enclave Common Library.

This PR would allow the SDK to be built without the library being present in the
build system. Additionally, it will help make simulation mode work on non-sgx systems.

This based on the same approach we do for the DCAP Quote Provider shared library.
(libdcap_quoteprove.so). See host/sgx/linux/sgxquoteproviderloader.c.

We keep copy of sgx_enclave_common.h in 3rdparty/intel. This describes the
contract between OE SDK and the sgx_enclave_common shared library.

The shared library is lazily loaded via a oe_once function.
Necessary functions are looked up and cached.

NTH:
libsgx_enclave_common.so can provide a version information that the OE SDK
can use to determine if the library conforms to the contract at the time
the SDK was built.

Testing:
1) The existing CI tests will automatically lockdown the case where
   the library exists.
2) We need tests for
   - Case when the library is not present
   - Case when the library is present, but does not have all the
     functions.

### DIFF
--- a/host/CMakeLists.txt
+++ b/host/CMakeLists.txt
@@ -172,6 +172,7 @@ if (UNIX)
     crypto/openssl/key.c
     crypto/openssl/rsa.c
     crypto/openssl/sha.c
+    linux/dynload.c
     linux/hostthread.c)
 
   list(APPEND PLATFORM_SDK_ONLY_SRC
@@ -192,6 +193,7 @@ elseif (WIN32)
     crypto/bcrypt/sha.c
     crypto/bcrypt/pem.c
     crypto/bcrypt/util.c
+    windows/dynload.c
     windows/hostthread.c)
 
   list(APPEND PLATFORM_SDK_ONLY_SRC
@@ -260,6 +262,7 @@ if (OE_SGX)
     sgx/quote.c
     sgx/registers.c
     sgx/report.c
+    sgx/sgx_enclave_common.c
     sgx/sgxload.c
     sgx/sgxmeasure.c
     sgx/sgxquote.c
@@ -446,29 +449,13 @@ endif ()
 target_include_directories(oehost PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 
 if (OE_SGX)
-  # Always link with the EnclaveCommonAPI
-  if (WIN32)
-    set(LIBPATHS ${NUGET_PACKAGE_PATH}/EnclaveCommonAPI/lib/native/x64-Release)
-    set(INCPATHS "${NUGET_PACKAGE_PATH}/EnclaveCommonAPI/Header Files")
-  endif ()
+  # For sgx_enclave_common.h.
+  target_include_directories(oehost PRIVATE
+    ${CMAKE_SOURCE_DIR}/3rdparty/intel)
+
   if (NOT LIBPATHS)
     set(LIBPATHS "/usr")
   endif ()
-  find_library(LIBSGX_COMMON NAMES sgx_enclave_common HINTS ${LIBPATHS})
-  if (NOT LIBSGX_COMMON)
-    message(FATAL_ERROR "Intel SGX EnclaveCommonAPI library not found, aborting!")
-  endif ()
-  add_library(sgx_enclave_common SHARED IMPORTED)
-  if (WIN32)
-     set_target_properties(sgx_enclave_common PROPERTIES
-        INTERFACE_INCLUDE_DIRECTORIES "${INCPATHS}"
-        IMPORTED_LOCATION $ENV{WINDIR}/System32
-        IMPORTED_IMPLIB ${LIBSGX_COMMON})
-  elseif (UNIX)
-    set_target_properties(sgx_enclave_common PROPERTIES IMPORTED_LOCATION ${LIBSGX_COMMON})
-  endif ()
-  target_link_libraries(oehost PUBLIC $<BUILD_INTERFACE:sgx_enclave_common>)
-
   # Optionally link in DCAP library
   if (HAS_QUOTE_PROVIDER)
     if (WIN32)
--- a/host/linux/dynload.c
+++ b/host/linux/dynload.c
@@ -1,0 +1,20 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include "../dynload.h"
+#include <dlfcn.h>
+
+void* oe_shared_library_load(const char* name)
+{
+    return dlopen(name, RTLD_LAZY | RTLD_LOCAL);
+}
+
+void oe_shared_library_unload(void* handle)
+{
+    dlclose(handle);
+}
+
+void* oe_shared_library_lookup(void* handle, const char* function)
+{
+    return dlsym(handle, function);
+}
--- a/host/sgx/sgx_enclave_common.c
+++ b/host/sgx/sgx_enclave_common.c
@@ -1,0 +1,214 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include "sgx_enclave_common.h"
+#include <openenclave/internal/trace.h>
+#include <stdlib.h>
+#include "../dynload.h"
+#include "../hostthread.h"
+
+/****** Pointers to functions in libsgx_enclave_common.so **************/
+static void* COMM_API (*_enclave_create)(
+    COMM_IN_OPT void* base_address,
+    COMM_IN size_t virtual_size,
+    COMM_IN size_t initial_commit,
+    COMM_IN uint32_t type,
+    COMM_IN const void* info,
+    COMM_IN size_t info_size,
+    COMM_OUT_OPT uint32_t* enclave_error);
+
+static size_t COMM_API (*_enclave_load_data)(
+    COMM_IN void* target_address,
+    COMM_IN size_t target_size,
+    COMM_IN_OPT const void* source_buffer,
+    COMM_IN uint32_t data_properties,
+    COMM_OUT_OPT uint32_t* enclave_error);
+
+bool COMM_API (*_enclave_initialize)(
+    COMM_IN void* base_address,
+    COMM_IN const void* info,
+    COMM_IN size_t info_size,
+    COMM_OUT_OPT uint32_t* enclave_error);
+
+bool COMM_API (*_enclave_delete)(
+    COMM_IN void* base_address,
+    COMM_OUT_OPT uint32_t* enclave_error);
+
+static bool COMM_API (*_enclave_set_information)(
+    COMM_IN void* base_address,
+    COMM_IN uint32_t info_type,
+    COMM_IN void* input_info,
+    COMM_IN size_t input_info_size,
+    COMM_OUT_OPT uint32_t* enclave_error);
+
+/****** Dynamic loading of libsgx_enclave_common.so/.dll **************/
+static void* _module;
+
+static void _unload_libsgx_enclave_common(void)
+{
+    if (_module)
+        oe_shared_library_unload(_module);
+}
+
+static void _load_libsgx_enclave_common_impl(void)
+{
+    oe_result_t result = OE_FAILURE;
+    void* module = _module;
+    if (!module)
+    {
+#if _WIN32
+        const char* library_name = "sgx_enclave_common.dll";
+#else
+        const char* library_name = "libsgx_enclave_common.so";
+#endif
+        OE_TRACE_INFO("Loading %s\n", library_name);
+        module = oe_shared_library_load(library_name);
+        if (module)
+        {
+            *(void**)&_enclave_create =
+                oe_shared_library_lookup(module, "enclave_create");
+            if (!_enclave_create)
+            {
+                OE_TRACE_ERROR("enclave_create function not found.\n");
+                goto done;
+            }
+
+            *(void**)&_enclave_load_data =
+                oe_shared_library_lookup(module, "enclave_load_data");
+            if (!_enclave_load_data)
+            {
+                OE_TRACE_ERROR("enclave_load_data function not found.\n");
+                goto done;
+            }
+
+            *(void**)&_enclave_initialize =
+                oe_shared_library_lookup(module, "enclave_initialize");
+            if (!_enclave_initialize)
+            {
+                OE_TRACE_ERROR("enclave_initialize function not found.\n");
+                goto done;
+            }
+
+            *(void**)&_enclave_delete =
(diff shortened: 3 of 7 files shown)


anakrish:
bors try

host/sgx/sgx_enclave_common.c
```diff
@@ -0,0 +1,214 @@
...
+    COMM_OUT_OPT uint32_t* enclave_error);
+
+static bool COMM_API (*_enclave_set_information)(
+    COMM_IN void* base_address,
+    COMM_IN uint32_t info_type,
+    COMM_IN void* input_info,
+    COMM_IN size_t input_info_size,
+    COMM_OUT_OPT uint32_t* enclave_error);
+
+/****** Dynamic loading of libsgx_enclave_common.so/.dll **************/
```
mingweishih, line 44:
```suggestion
/****** Dynamic loading of the sgx_enclave_common library **************/
```

host/sgx/sgx_enclave_common.c
```diff
@@ -0,0 +1,214 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include "sgx_enclave_common.h"
+#include <openenclave/internal/trace.h>
+#include <stdlib.h>
+#include "../dynload.h"
+#include "../hostthread.h"
+
+/****** Pointers to functions in libsgx_enclave_common.so **************/
```
mingweishih, line 10:
```suggestion
/****** Pointers to functions in the sgx_enclave_common library **************/
```

host/sgx/sgx_enclave_common.c
```diff
@@ -0,0 +1,214 @@
...
+    COMM_IN size_t input_info_size,
+    COMM_OUT_OPT uint32_t* enclave_error);
+
+/****** Dynamic loading of libsgx_enclave_common.so/.dll **************/
+static void* _module;
+
+static void _unload_libsgx_enclave_common(void)
+{
+    if (_module)
+        oe_shared_library_unload(_module);
```
mingweishih, line 50: *Non-blocking* Although based on the flow, there should not be the case that calling `_load_libsgx_enclave_common` after this call. But may be explicitly set _module = `NULL` here could help with eliminating potential warning.

jhand2:
If we could convince Intel to produce dev nuget packages with the SGX libraries do you think that would also solve the problem? Are there still scenarios that would not be supported?

jhand2:
This doc details a nuget package for the EnclaveCommonAPI, it just seems to not be published yet: https://download.01.org/intel-sgx/sgx-dcap/1.6/windows/docs/Intel_SGX_DCAP_Windows_SW_Installation_Guide.pdf

host/CMakeLists.txt
```diff
@@ -260,6 +262,7 @@ if (OE_SGX)
     sgx/quote.c
     sgx/registers.c
     sgx/report.c
+    sgx/sgx_enclave_common.c
```
radhikaj, line 265: Is this a new file? I cant seem to find it in this PR.

radhikaj:
@anakrish Thanks for this. On the host end, the sgx_enclave_common lib is not unloaded until the process is terminated. Is that correct?

mingweishih:
> @anakrish Thanks for this. On the host end, the sgx_enclave_common lib is not unloaded until the process is terminated. Is that correct?

I think that's correct. The library is unloaded upon the enclave terminates.

https://github.com/openenclave/openenclave/blob/24ad84ef0faa9aa98ef8de508efd68432011df06/host/sgx/sgx_enclave_common.c#L109

anakrish:
bors try

host/windows/dynload.c
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include "../dynload.h"
+#include <Windows.h>
+
+void* oe_shared_library_load(const char* name)
+{
+    return (void*)LoadLibraryEx(name, NULL, 0);
```
CodeMonkeyLeet, line 9: Consider using the `LOAD_LIBRARY_SEARCH_SYSTEM32` flag since sgx_enclave_common.dll is part of the Intel driver components and should only be loaded from there.

host/sgx/sgx_enclave_common.c
```diff
@@ -0,0 +1,214 @@
...
+    COMM_IN size_t virtual_size,
+    COMM_IN size_t initial_commit,
+    COMM_IN uint32_t type,
+    COMM_IN const void* info,
+    COMM_IN size_t info_size,
+    COMM_OUT_OPT uint32_t* enclave_error)
+{
+    if (!_load_libsgx_enclave_common())
+    {
+        *enclave_error = ENCLAVE_NOT_SUPPORTED;
```
CodeMonkeyLeet, line 142: Can we make these wrappers part of the OE side instead? (i.e. `oe_` prefixed, return `oe_result_t` values etc.)

* I would prefer not to overload the meaning of SGX's error codes; the conditions for this are pretty specific on the SGX side today: SGX feature set or AESM doesn't exist. Here it's really an `OE_PLATFORM_ERROR` or a more unique error of your choice.
* I would prefer not to create the illusion this is actually a call into the sgx_enclave_common library already (i.e. folks may end up hunting for the function name in the SGX repo instead of OE based on an incorrect expectation of the error code).

host/linux/dynload.c
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include "../dynload.h"
+#include <dlfcn.h>
+
+void* oe_shared_library_load(const char* name)
+{
+    return dlopen(name, RTLD_LAZY | RTLD_LOCAL);
```
CodeMonkeyLeet, line 9: `RTLD_LOCAL` changes the semantics of this slightly for the sgx_common_library functions since previously based on static linkage, all the symbols were global. I don't know if it matters or if this is preferred?

More generally, this still makes me nervous as the security guidance on the Linux side has been towards `BIND_NOW`, and Linux doesn't have the same kinds of controls for shared library load pathing by the app? I don't know of a better solution though.

host/CMakeLists.txt
```diff
@@ -446,29 +449,13 @@ endif ()
 target_include_directories(oehost PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 
 if (OE_SGX)
-  # Always link with the EnclaveCommonAPI
-  if (WIN32)
-    set(LIBPATHS ${NUGET_PACKAGE_PATH}/EnclaveCommonAPI/lib/native/x64-Release)
-    set(INCPATHS "${NUGET_PACKAGE_PATH}/EnclaveCommonAPI/Header Files")
-  endif ()
+  # For sgx_enclave_common.h.
+  target_include_directories(oehost PRIVATE
```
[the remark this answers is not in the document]
jhand2: I would feel much better about this change if we could avoid checking in the PSW header

radhikaj:
@anakrish  When I test this on a non-SGX Ubuntu VM, I can build and run the tests on the VM. Samples fail though. The exported targets still as a dependency on sgx-enclave-common. See cmake/openenclave-config.cmake.in. 
if (OE_SGX)
  if (NOT TARGET openenclave::sgx_enclave_common)
    if (UNIX)
      find_library(SGX_ENCLAVE_COMMON_LIB NAMES sgx_enclave_common HINTS "/usr")
    elseif (WIN32)
      find_library(SGX_ENCLAVE_COMMON_LIB NAMES sgx_enclave_common
           HINTS ${NUGET_PACKAGE_PATH}/EnclaveCommonAPI/lib/native/x64-Release)
    endif

host/CMakeLists.txt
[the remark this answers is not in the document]
anakrish: @CodeMonkeyLeet 
We need to have a path for building the SDK for simulation mode, easily, on non-sgx machines. From a usability point of view, installing the Intel DCAP libraries following our documentation [here](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/WindowsManualSGX1Prereqs.md) or [here](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/WindowsManualSGX1FLCDCAPPrereqs.md) has been quite painful and error prone if you were to try to setup a Windows development machine to play with the SDK.

I would like to avoid that pain-point that is a barrier to contributing to the SDK.

What benefit do we get from having the user to go through the painful process of getting the enclave common API headers? 

At the very least, we ought to just pull the header from: 
https://github.com/intel/linux-sgx/blob/master/psw/enclave_common/sgx_enclave_common.h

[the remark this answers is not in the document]
jhand2: Intel plans to publish nuget packages soon right? When they do that can we remove this header and just have people download the nuget packages?

I understand the current process is error prone and painful so I'm with you on that. I would just prefer to move to using the packages when there's a better option available.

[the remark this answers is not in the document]
CodeMonkeyLeet: We already get the sgx_enclave_common.h header from a nuget package and that can be deployed on non-sgx machines, just like the .deb packages for development containing the same header are installed.

anakrish:
bors try

letmaik:
Is there a way to opt-out of this? dlopen`ing dependencies makes life harder if you want to analyze dependencies automatically (e.g. for auditing) and also if you want to bundle dependencies automatically into a package. Why is this change needed anyway?  You say "This eliminates that the host application had on Intel SGX Enclave Common Library." Well if it doesn't need it, why don't you remove the dependency in the build system? Can you expand a bit more on the cases for which the dependency is needed and for which it isn't?

host/CMakeLists.txt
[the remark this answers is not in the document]
letmaik: Why is OE including Intel headers if it doesn't need them for simulation mode builds? Wouldn't this PR be redundant if you just publish an additional Debian package for simulation mode without Intel dependencies?

anakrish:
> You say "This eliminates that the host application had on Intel SGX Enclave Common Library." Well if it doesn't need it, why don't you remove the dependency in the build system? 

This PR heads in the direction of removing build time dependency on Intel libraries. One will be able to build the SDK without the enclave common libraries.

anakrish:
> Can you expand a bit more on the cases for which the dependency is needed and for which it isn't?

To run enclaves in simulation mode, there is no need of any sgx libraries.
In a similar manner, to run enclaves without dcap attestation, there no need of dcap libraries.

anakrish:
> dlopen`ing dependencies makes life harder if you want to analyze dependencies automatically (e.g. for auditing) and also if you want to bundle depeendencies automatically into a package

Can dependencies be expressed at the (deb/rhel) package level rather than how shared libraries are loaded?

letmaik:
> > dlopen`ing dependencies makes life harder if you want to analyze dependencies automatically (e.g. for auditing) and also if you want to bundle depeendencies automatically into a package
> 
> Can dependencies be expressed at the (deb/rhel) package level rather than how shared libraries are loaded?

Currently the OE Debian package has this:
https://github.com/openenclave/openenclave/blob/0ac25ddab69745c66e182512974211f37d93a778/cmake/cpack_settings.cmake#L15

I'm assuming (it's missing in this PR) you would remove the dependencies from there so that it can be used on SGX and non-SGX systems (simulation mode). If you only do that and nothing else then I don't think that's a good idea because then you move the burden to the user on installing the right versions of the Intel libraries for SGX systems. You would at least have to create a new meta package that has a dependency on the existing OE package (now without Intel dependencies) and the Intel libraries.

anakrish:
@letmaik This was brought up in this weeks Sig-Arch meeting. The consensus was to dynamically load the SGX libraries but retain the package dependencies on SGX packages.

anakrish:
In the past 2 months since this PR, I've had to delete and recreate my fork and this PR currently shows "unknown repository". I'm closing this PR and have created another PR #3155 that is simpler and incorporates the feedback from this PR.

host/CMakeLists.txt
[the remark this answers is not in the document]
anakrish: Addressed in PR #3155 by using a wrapper header instead of the TEE header.

host/linux/dynload.c
anakrish: Addressed in PR #3155

host/sgx/sgx_enclave_common.c
anakrish: Addresses in PR #3155

host/sgx/sgx_enclave_common.c
anakrish: I have addressed this differently. See #3155 
The wrapper is thin and just forwards the call to SGX enclave common and does not overload the meaning of SGX error codes.
Failure to load sgx enclave common is treated as a catastrophic error and terminates the application. This is not that big of a deviation from the current behavior where the application won't load if sgx enclave common is not present in the system.

host/windows/dynload.c
anakrish: Addressed in #3155

letmaik:
> @letmaik This was brought up in this weeks Sig-Arch meeting. The consensus was to dynamically load the SGX libraries but retain the package dependencies on SGX packages.

@anakrish If you keep the SGX libraries as package dependency, what's the point then?

anakrish:
At the very least, it will allow building the SDK for simulation on non-sgx machines via HAS_QUOTE_PROVIDER=off. #3127
SGX packages can be installed on non-sgx machines; therefore it was deemed OK to retain the SGX dependencies in the OE SDK package. Note: The windows nuget package does not have the dependency.

letmaik:
Why is there so much resistence to offer a nice simulation experience? The Intel libraries are neither on the standard Ubuntu repositories nor on the Microsoft one, so the user manually has to add another repository to their system and install packages they will never use. It seems a shame that you went nearly all the way but then stop short of a complete solution.

anakrish:
@letmaik 
I see it as incremental changes. Folks have different perspectives,  different concerns, and that is why getting consensus is important. As evidenced by this PR, dynamically loading sgx-enclave-common - a strategy that we already use for other intel libraries - itself brought differing perspectives.

> The Intel libraries are neither on the standard Ubuntu repositories nor on the Microsoft one, so the user manually has to add another repository to their system and install packages they will never use 

It there a strong need for this requirement of not installing intel libraries? If so, we can bring it up in a Sig-Arch meeting as an amendment.
CC: @radhikaj

letmaik:
@anakrish There is no strong need.

PR #2883, opened 2020-04-16
Add option to format-code to only format staged changes

jhand2:
The git pre-commit hook only checks staged files, but then we advise people in the error message to run format-code on every file in the repo when this is not really necessary. Add a `--staged` flag to format code and suggest that people use that in the pre-commit hook error message.


radhikaj:
Thanks for this awesome fix!

radhikaj:
Bors r+

PR #2885, opened 2020-04-17
disable win 2019 tests from PR merges

BRMcLaren:
Disabling win2019 testing temporarily on pull requests due to discoveries of underlying issues with the Intel dependencies.


BRMcLaren:
Bors try

PR #2886, opened 2020-04-17
Change OE_TRACE_ERROR to OE_TRACE_WARNING when non-standard API is not available

yufjiang:
Function sgx_ql_set_logging_function is an Azure DCAP client specific API, which is not defined in Intel DCAP standard spec. To enable OE to be used in non-ACC environments, a warning, instead of an error, is given when this non-standard API is not available.


yufjiang:
bors try

Issue #2486, opened 2020-01-30
Ensure compiler portability in public headers


jhand2:
Compile all the samples and tests with `-Werror` and don't turn off any warnings. This will help ensure that our public headers don't trigger warnings regardless of how an enclave developer chooses to compile their enclave.

To support #2376 it would be great if we could also compile these things with C99

jhand2:
A more comprehensive solution would be to have a test that includes every public header and makes use of every macro but this seemed operationally prohibitive.

dthaler:
Every public API should have a test.   See PR #2163 and comments there from CGC members expressing support for that requirement.

mingweishih:
Add note: `-Werror` has already been a default option to all the tests. #2892 adds a test that includes all the public headers (excluding 3rd party ones) and compiles them against C99 standard.

radhikaj:
@mingweishih Can this be closed?

mingweishih:
Sure, closing.

PR #2892, opened 2020-04-18
Test public headers against C99 standard

### LINKED ISSUE #2376: OE is not C99 compliant

mingweishih:
This PR adds a test that compiles all the public headers (3rd party ones are not included) against C99 standard (using `std=gnu99`). Any violation should cause compile-time errors.

Also, cleaning up code that are unnecessary / inconsistent as follows.
* `host/sgx/enclave.h`
   - `typedef` the `struct _oe_enclave` as `oe_enclave_t` because it's used in the following `GetEnclaveEvent` function.
* `include/openenclave/corelibc/bits/types.h `
   - Remove `intptr_t`, `time_t`, and `suseconds_t`, which are already defined in `openenclave/bits/types.h`
* `include/openenclave/internal/crypto/crl.h`
   - Remove a redundant `typedef`

Fixes #2376 #2486

### DIFF
--- a/host/sgx/enclave.h
+++ b/host/sgx/enclave.h
@@ -90,7 +90,7 @@ oe_thread_binding_t* oe_get_thread_binding(void);
  *  This structure must be kept in sync with the defines in
  *  debugger/pythonExtension/gdb_sgx_plugin.py.
  */
-struct _oe_enclave
+typedef struct _oe_enclave
 {
     /* A "magic number" to validate structure */
     uint64_t magic;
@@ -130,7 +130,7 @@ struct _oe_enclave
 
     /* Manager for switchless calls */
     oe_switchless_call_manager_t* switchless_manager;
-};
+} oe_enclave_t;
 
 /* Get the event for the given TCS */
 EnclaveEvent* GetEnclaveEvent(oe_enclave_t* enclave, uint64_t tcs);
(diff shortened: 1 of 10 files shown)


mingweishih:
bors try

mingweishih:
bors try

jhand2:
Fails for arm64: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAgnostic-Linux/detail/Agnostic-Linux/81/pipeline

mingweishih:
> Fails for arm64: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAgnostic-Linux/detail/Agnostic-Linux/81/pipeline

Thanks! Investigating now.

mingweishih:
bors try

mingweishih:
bors try

PR #2893, opened 2020-04-18
Run cmake-format over cmake files

dthaler:
Issue #2797 called out a couple of gaps, of which one is that the cmake files (*.cmake, CMakeLists.txt) do not have a consistent style.   This PR addresses that specific gap by using the cmake-format tool that Amaury pointed to in that issue.

This PR has 2 commits which should be reviewed separately, for efficiency of review.  The first commit is the actual change, and is relatively small and easy to check each line of (especially since most of the code in the scripts is identical to the existing code-format scripts).   The second commit is the result of actually running the tool, and is large but dynamically generated so reviewers need not review everything, just spot check it.

Since there were previously a mix of styles in the cmake files, please look at the second commit to see if the style it's converting to looks ok, and if not then the options in the .cmake-format.json file in the first commit are what controls the style.


dthaler:
bors try

scripts/install-windows-prereqs.ps1
```diff
@@ -359,6 +359,10 @@ function Install-Python3 {
     New-Directory -Path $installDir -RemoveExisting
     Move-Item -Path "$tempInstallDir\*" -Destination $installDir
     Add-ToSystemPath -Path $installDir
+
+    Start-ExecuteWithRetry -ScriptBlock {
+        pip install cmake_format
```
jhand2, line 364: Does this script already install python?

dthaler: Yes, that's what line 354 does

dthaler:
Rebased.  I don't think there's any requirement to hold other PRs.  This creates a new gate.  If anything sneaks in after the rebase and before the merge, then it's easy to submit a follow-up PR to clean up any stragglers.   Based on 3 approvals, I think this can be merged now.

dthaler:
bors retry

PR #2897, opened 2020-04-20
Add build option to OE to disable building of tests

### LINKED ISSUE #2894: Add BUILD_TESTS as CMake build option
It would be nice if one could disable building of the tests.

jhand2:
Add build option to OE to disable building of tests

Fixes #2894


radhikaj:
Thanks a lot for taking this on @jhand2. to build tests, you have to opt in? Can the opt-in be to exclude tests?

jhand2:
> Thanks a lot for taking this on @jhand2. to build tests, you have to opt in? Can the opt-in be to exclude tests?

I set he default to `ON`. So if you don't pass any option to cmake, it will build the tests by default

PR #2914, opened 2020-04-22
add nested switchless calls, call ecall->ocall->ecall->ocall

yanxue-intel:

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -31,6 +31,7 @@ if (OE_SGX)
     add_subdirectory(host_verify)
     add_subdirectory(switchless)
     add_subdirectory(switchless_threads)
+    add_subdirectory(switchless_nestedcalls)
 endif()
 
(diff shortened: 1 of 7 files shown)


jinghe-INTC:
bors try

jhand2:
@anakrish is this expected to work?

anakrish:
> @anakrish is this expected to work?

For switchless calls it should work. For normal calls it won't work since re-entrant ecalls are not supported.

@yanxue-intel  This is a great test!

anakrish:
bors try

anakrish:
I couldn't get this test to fail, even after running it in a loop for 1000 times.

anakrish:
I am able to reproduce this with debug builds. USE_DEBUG_MALLOC is turned on only in debug builds.

anakrish:
This test and the associated failure due to memory leaks has revealed a long standing bug with switchless manager cleanup: enclave was being destructed before the switchless manager was cleaned up. This resulted in memory being reported as leaks.

Provided a fix in #2927

anakrish:
bors try

PR #2918, opened 2020-04-22
<Test> Attempting to narrow down pthread/othread hang

radhikaj:
Revert "add spinlock/multithread/errno check in same/different enclave"
This reverts commit cebdabc075568c640c4765de869c18174e2b0104.


jhand2:
bors try

PR #2919, opened 2020-04-22
Disable errno thread tests

jhand2:
Since the new errno oethread and pthread tests were added, we have seen multiple timeouts in CI. Disable these tests until the issue can be root-caused.


radhikaj:
Would like to take this for now to allow for PRs to merge.

PR #2921, opened 2020-04-22
Fix tcs exhaustion test to predictably terminate

anakrish:
Previously each thread would keep making ocalls to check if it can exit.
This could cause the thread to hog the CPU, preventing other threads
from making progress so that the termination condition for the test
is reached.

Wait on a condition variable instead. This allows other threads to run.

### DIFF
--- a/tests/thread/host/host.cpp
+++ b/tests/thread/host/host.cpp
@@ -7,6 +7,7 @@
 #include <atomic>
 #include <cassert>
 #include <chrono>
+#include <condition_variable>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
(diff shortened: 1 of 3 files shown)


anakrish:
bors try

tests/thread/host/host.cpp
```diff
@@ -241,53 +246,48 @@ void* tcs_thread(oe_enclave_t* enclave, size_t tcs_req_count)
...
         g_tcs_out_thread_count.load());
 
-    // Crux of the test is to get OE_OUT_OF_THREADS i.e. to exhaust the TCSes
-    OE_TEST(g_tcs_out_thread_count > 0);
-    // Verifying that everything adds up fine
-    OE_TEST(
-        tcs_used_thread_count + g_tcs_out_thread_count == test_tcs_req_count);
-    // Sanity test that we are not reusing the bindings
+    // Assert that expected thread exhaustion failures have been reached.
+    OE_TEST(g_tcs_out_thread_count == expected_out_of_threads);
```
CodeMonkeyLeet, line 282: This test is redundant since the test will just hang and not call `g_tcs_cv.notify_all()` if this condition were not already true by this point.

tests/thread/host/host.cpp
```diff
@@ -217,15 +218,19 @@ void test_errno_multi_threads_diffenclave(
...
+void* tcs_thread(oe_enclave_t* enclave, size_t expected_out_of_threads)
 {
-    oe_result_t result = enc_test_tcs_exhaustion(enclave, tcs_req_count);
+    oe_result_t result = enc_test_tcs_exhaustion(enclave);
     if (result == OE_OUT_OF_THREADS)
     {
-        // increment g_tcs_out_thread_count when thread exhaustion is reached
-        ++g_tcs_out_thread_count;
+        // Increment tcs count. Resume threads if expected tcs exhaustion
+        // failures has been reached.
```
mingweishih, line 231:
```suggestion
	  // failures have been reached.
```

tests/thread/host/host.cpp
anakrish: Redundancy is fine. I feel having this condition makes it clear what the test expectations are.

anakrish:
@CodeMonkeyLeet I am just going to merge this in. If the CI allows this to go through, it will either fix the hangs (if tcs exhaustion was the culprit) or have no impact on the hangs. Either ways I see nothing negative from pushing this test fix in.

anakrish:
Fixed format code. Accepting `suggestions` can mess up formatting.

anakrish:
The failure seems to be in tls:
> TLS server:failed
> ! mbedtls_ssl_handshake returned -0x7780
> TLS server:server communication error -30592
>TLS server:Last error was: -30592 - SSL - A fatal alert message was received from our peer

anakrish:
bors try

PR #2922, opened 2020-04-22
Fix the print test on cross-compilation configurations

mingweishih:
The print test does not correctly run on cross-compilation configurations (i.e., when `ADD_WINDOWS_ENCLAVE_TESTS=ON` and ` BUILD_ENCLAVES = OFF`).

This PR fixes the issue by explicitly adding `add_enclave_test`, which handles the test run with cross-compilation configurations, before the actual test (i.e., comparing the output of the test against the oracle output).

### DIFF
--- a/tests/print/CMakeLists.txt
+++ b/tests/print/CMakeLists.txt
@@ -5,21 +5,23 @@ add_subdirectory(host)
 
 if (BUILD_ENCLAVES)
     add_subdirectory(enc)
-
-    #note: this is Linux-specific
-    if (WIN32)
-        set(RESULTDIR ${CMAKE_CURRENT_SOURCE_DIR}/results/windows)
-    else()
-        set(RESULTDIR ${CMAKE_CURRENT_SOURCE_DIR}/results/others)
-    endif()
-
-    add_test(NAME tests/print
-        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
-        COMMAND sh -c "OE_LOG_LEVEL=NONE host/print_host ./enc/print_enc >testout.stdout 2>testout.stderr &&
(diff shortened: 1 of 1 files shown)


mingweishih:
bors try

tests/print/CMakeLists.txt
```diff
@@ -5,21 +5,23 @@ add_subdirectory(host)
...
+    set(RESULTDIR ${CMAKE_CURRENT_SOURCE_DIR}/results/others)
 endif()
+
+# Use add_enclave_test to ensure the enclave binaries are copied from the pre-built directory
+# when ADD_WINDOWS_ENCLAVE_TESTS is ON and BUILD_ENCLAVES is OFF.
+add_enclave_test(tests/print print_host print_enc)
+
+add_test(NAME tests/print-diff
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+    COMMAND sh -c "OE_LOG_LEVEL=NONE host/print_host ./enc/print_enc >testout.stdout 2>testout.stderr &&
```
anakrish, line 24: Is `sh` available on all platforms?

mingweishih: Git bash (required by OE Windows build) provides `sh` on Windows. Unless Git bash is not a requirement anymore, `sh` should be available on all the platforms that OE support.

anakrish: In that case do we need to access `sh` via `OE_BASH` as done here:
https://github.com/openenclave/openenclave/blob/71edd2cf92f52017410d5e3262628b962067560a/tests/crypto_crls_cert_chains/data/CMakeLists.txt#L25

mingweishih: Sure, I agree it's more reliable. Updated.

PR #2926, opened 2020-04-22
Remove redundant testing badge

BRMcLaren:
The current packaging regression testing has been augmented to include the testing of the installation candidates. This renders the "Nightly Package Testing" badge redundant and should be removed.


BRMcLaren:
Thanks for the reviews!

PR #2927, opened 2020-04-22
Stop switchless manager before invoking enclave destructor.

anakrish:
Otherwise the worker threads may not be able to clean up
correctly.


anakrish:
bors try

PR #2929, opened 2020-04-22
Retain ECALL worker thread's tcs binding.

anakrish:
Instead of returning when spin count threshold has been reached,
make an ocall to sleep in the host.
This ensures that the tcs is still bound to the host thread.
When the messages arrive, the worker thread can go back into the
enclave in the same tcs.
Without retaining the tcs binding, there may not be a tcs available
for the worker thread to go back into the enclave.

### DIFF
--- a/host/sgx/switchless.c
+++ b/host/sgx/switchless.c
@@ -70,6 +70,12 @@ static void* _switchless_ocall_worker(void* arg)
     return NULL;
 }
 
+void oe_sgx_sleep_switchless_worker_ocall(oe_enclave_worker_context_t* context)
+{
+    // Wait for messages.
+    oe_enclave_worker_wait(context);
+}
+
 /*
 ** The thread function that handles switchless ecalls
 **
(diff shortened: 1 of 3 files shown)


host/sgx/switchless.c
```diff
@@ -78,20 +84,13 @@ static void* _switchless_ecall_worker(void* arg)
 {
     oe_enclave_worker_context_t* context = (oe_enclave_worker_context_t*)arg;
 
-    // Loop until stop has been requested.
-    while (!context->is_stopping)
+    // Enter enclave to process ecall messages.
+    if (oe_sgx_switchless_enclave_worker_thread_ecall(
```
mingweishih, line 88: If I understand correctly, the change here is because the while loop is also done in the `oe_sgx_switchless_enclave_worker_thread_ecall`. Is that correct?

anakrish: Yes.

anakrish:
bors try

anakrish:
bors try

PR #2933, opened 2020-04-23
Infra fixes

oprinmarius:
- Sysprep script is failing with the following error:
`"Get-Service : Cannot find any service with service name 'WindowsAzureTelemetryService'"`
Thus, we remove this condition before running Sysprep

- Update Windows install script to copy SGX Dlls to System32 folder for Windows Server 2016 nonSGX

### DIFF
--- a/scripts/install-windows-prereqs.ps1
+++ b/scripts/install-windows-prereqs.ps1
@@ -602,6 +602,11 @@ function Install-DCAP-Dependencies {
                 }
                 Write-Output $install
             }
+            elseif (($LaunchConfiguration -eq "SGX1FLC-NoDriver") -and (${OS_VERSION} -eq "WinServer2016"))
+            {
+                 Write-Output "Copying Intel_SGX_DCAP dll files into $($env:SystemRoot)\system32"
+                 Copy-item -Path $PACKAGES_DIRECTORY\Intel_SGX_DCAP\$driver\drivers\*\*.dll $env:SystemRoot\system32\
+            }
         }
     }
 
@@ -645,7 +650,7 @@ function Install-DCAP-Dependencies {
         Throw "Failed to install nuget EnclaveCommonAPI"
     }
 
-    if (($LaunchConfiguration -eq "SGX1FLC") -or (${OS_VERSION} -eq "WinServer2019"))
+    if (($LaunchConfiguration -eq "SGX1FLC") -or (${OS_VERSION} -eq "WinServer2016"))
     {
         # Please refer to Intel's Windows DCAP documentation for this registry setting: https://download.01.org/intel-sgx/dcap-1.2/windows/docs/Intel_SGX_DCAP_Windows_SW_Installation_Guide.pdf
         New-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\sgx_lc_msr\Parameters" -Name "SGX_Launch_Config_Optin" -Value 1 -PropertyType DWORD -Force
--- a/.jenkins/infrastructure/provision/run-sysprep.ps1
+++ b/.jenkins/infrastructure/provision/run-sysprep.ps1
@@ -4,7 +4,6 @@
 $ErrorActionPreference = "Stop"
 
 while ((Get-Service RdAgent).Status -ne 'Running') { Start-Sleep -s 5 }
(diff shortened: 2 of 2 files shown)


scripts/install-windows-prereqs.ps1
```diff
@@ -602,6 +602,11 @@ function Install-DCAP-Dependencies {
                 }
                 Write-Output $install
             }
+            elseif (($LaunchConfiguration -eq "SGX1FLC-NoDriver") -and (${OS_VERSION} -eq "WinServer2016"))
```
jhand2, line 605: The PR description says to do this for non-SGX but this configuration is for an SGX machine, just with no driver. Is this intentional?

scripts/install-windows-prereqs.ps1
```diff
@@ -645,7 +650,7 @@ function Install-DCAP-Dependencies {
         Throw "Failed to install nuget EnclaveCommonAPI"
     }
 
-    if (($LaunchConfiguration -eq "SGX1FLC") -or (${OS_VERSION} -eq "WinServer2019"))
+    if (($LaunchConfiguration -eq "SGX1FLC") -or (${OS_VERSION} -eq "WinServer2016"))
```
jhand2, line 653: Why is this not needed for WS2019?

BRMcLaren:
https://oe-jenkins-tf.westeurope.cloudapp.azure.com/job/CI-CD_Infrastructure/job/OpenEnclave-Rebuild-Infrastructure/11/

Is a rebuild against the current master state(after the merge) let's see if there is any regression to start. The feedback time is very high.

scripts/install-windows-prereqs.ps1
BRMcLaren: I am running a regression with these lines reverted here:

https://oe-jenkins-tf.westeurope.cloudapp.azure.com/job/CI-CD_Infrastructure/job/OpenEnclave-E2E-Testing/181/

scripts/install-windows-prereqs.ps1
BRMcLaren: I am running a regression with these changes: https://github.com/BRMcLaren/openenclave/commit/c84e56054c164a1028f2ff40181676075165dcb9 

https://oe-jenkins-tf.westeurope.cloudapp.azure.com/job/CI-CD_Infrastructure/job/OpenEnclave-E2E-Testing/182/

(when it starts)

BRMcLaren:
I am also running an e2e https://oe-jenkins-tf.westeurope.cloudapp.azure.com/job/CI-CD_Infrastructure/job/OpenEnclave-Rebuild-Infrastructure/13/

On this commit change

https://github.com/BRMcLaren/openenclave/commit/384db752a9873bd990496a36b55151d272ef8efc

Specifically reverting the reg key on 653

scripts/install-windows-prereqs.ps1
ionutbalutoiu: @jhand2 

The `SGX1FLC-NoDriver` with `WinServer2016` is what we use for our machines in CI that run Windows simulation tests.

To be more specific, those are the Win Server 2016 machines what we don't install SGX driver on. And the SGX driver is not automatically setup via Windows Update on Win2016, so it doesn't matter at the end of the day if it has SGX or not since it's not used.

On these Windows Server 2016 machines, we run Windows simulation tests. And, the DLLs needs to be into system PATH. Otherwise, from what I know, the tests will fail.

However, this DLL copy is not needed on 2019, and that's why it was completely removed here: https://github.com/openenclave/openenclave/commit/0d74e2202fe8effce89aa635b6829f6875ce0b10#diff-846bcbf4c48fa80e76202a5e31d77476

scripts/install-windows-prereqs.ps1
ionutbalutoiu: @jhand2 @BRMcLaren 

This shouldn't be changed, and I'm really sorry for not seeing this change in my review.

Marius reverts this change with the PR https://github.com/openenclave/openenclave/pull/2935.

BRMcLaren:
Cancelling all e2e and rebuilds, there is a PR to fix regressions and this does not impact v0.9.

PR #2934, opened 2020-04-23
Fix the logic of detecting compilers when LVI mitigation is enabled

### LINKED ISSUE #2670: Implementing the logic of detecting clang in CMake for samples
Similar to the support of clang detection in Make for samples (#2651), we should support similar logic in CMake for the consistency.

### LINKED ISSUE #2930: Our gcc tests in CI use clang
See log for this GCC build `ACC1604 gcc Debug LVI`: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/rest/organizations/jenkins/pipelines/pipelines/pipelines/Azure-Linux/runs/148/nodes/30/steps/223/log/?start=0

```
-- The C compiler identification is Clang 7.1.0
-- The CXX compiler identification is Clang 7.1.0
-- The ASM compiler identification is Clang
-- Found assembler: /usr/local/lvi-mitigation/bin/clang-7
-- Check for working C compiler: /usr/local/lvi-mitigation/bin/clang-7
-- Check for working C compiler: /usr/local/lvi-mitigation/bin/clang-7 -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/local/lvi-mitigation/bin/clang++-7
-- Check for working CXX compiler: /usr/local/lvi-mitigation/bin/clang++-7 -- works
```

The builds for all GCC linux builds I've looked at look to be the same.

mingweishih:
This PR fixes the logic of detecting compilers when LVI mitigation is enabled. That is, the old logic always picks `clang-7` (if installed) regardless the environment variable `CC` is set to `gcc`. 
- Fixes #2930

This fix also allows users to specify the version of clang (other than the default version 7) when building the helloworld sample with LVI mitigation.
- Fixes #2670 

### DIFF
--- a/cmake/configure_lvi_mitigation_build.cmake
+++ b/cmake/configure_lvi_mitigation_build.cmake
@@ -1,25 +1,92 @@
 # Copyright (c) Open Enclave SDK contributors.
 # Licensed under the MIT License.
 
-# Adopt customized compiler wrappers for LVI mitigation.
-function (configure_lvi_mitigation_build LVI_MITIGATION_BINDIR)
-  if (NOT EXISTS "${LVI_MITIGATION_BINDIR}")
-    message(FATAL_ERROR "${LVI_MITIGATION_BINDIR} does not exist.")
+macro (detect_compiler BINDIR CC)
+  set(GCC gcc)
+  set(CLANG clang)
+  if (${CC} STREQUAL CXX)
+    set(GCC g++)
+    set(CLANG clang++)
   endif ()
-  # Perfer clang over gcc.
-  if (EXISTS "${LVI_MITIGATION_BINDIR}/clang-7")
+
+  # Set the COMPILER as the variable name (either C_COMPILER or CXX_COMPILER).
+  set(COMPILER "${CC}_COMPILER")
+
+  # Use the second level of unwrapping on COMPILER to get the compiler name.
+  # If the default compiler is found, return.
+  if (EXISTS "${BINDIR}/${${COMPILER}}")
+    return()
+  endif ()
+
+  if (NOT OE_IN_PACKAGE)
+    # Build OE. Fallback to gcc/g++.
+    set(${COMPILER} ${GCC})
+  else ()
+    # Build enclave applications. Try to search newer versions of clang/clang++.
+    # Be consistent to the logic implemented by samples/config.mk.
+    foreach (VERSION 9 8 7)
+      set(CLANG_VERSION "")
+      if (EXISTS "${BINDIR}/${${COMPILER}}-${VERSION}")
+        set(CLANG_VERSION ${VERSION})
+        break()
+      endif ()
+    endforeach ()
+    # Set the compiler if a version of clang/clang++ is found.
+    if (CLANG_VERSION)
+      set(${COMPILER} ${CLANG}-${CLANG_VERSION})
+    else ()
+      set(${COMPILER} ${GCC})
+    endif ()
+  endif ()
+endmacro ()
+
+# Adopt customized compiler wrappers for LVI mitigation.
+function (configure_lvi_mitigation_build)
+  cmake_parse_arguments(OE "IN_PACKAGE" "BINDIR" "" ${ARGN})
+
+  if (NOT EXISTS "${OE_BINDIR}")
+    message(FATAL_ERROR "${OE_BINDIR} does not exist.")
+  endif ()
+
+  if (NOT OE_IN_PACKAGE)
+    # Default to clang-7 when building SDK.
+    set(C_COMPILER clang-7)
+    set(CXX_COMPILER clang++-7)
+  else ()
+    # Default to clang when building enclave applications.
+    set(C_COMPILER clang)
+    set(CXX_COMPILER clang++)
+  endif ()
+
+  # Overwrite the default C compiler if CC is explicitly specified.
+  # Otherwise, select the compiler based on the detection logic.
+  if (DEFINED ENV{CC})
+    set(C_COMPILER $ENV{CC})
+  else ()
+    detect_compiler(${OE_BINDIR} C)
+  endif ()
+
+  if (EXISTS "${OE_BINDIR}/${C_COMPILER}")
     set(CMAKE_C_COMPILER
-        ${LVI_MITIGATION_BINDIR}/clang-7
-        PARENT_SCOPE)
-    set(CMAKE_CXX_COMPILER
-        ${LVI_MITIGATION_BINDIR}/clang++-7
+        ${OE_BINDIR}/${C_COMPILER}
         PARENT_SCOPE)
   else ()
-    set(CMAKE_C_COMPILER
-        ${LVI_MITIGATION_BINDIR}/gcc
-        PARENT_SCOPE)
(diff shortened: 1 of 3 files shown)


mingweishih:
bors try

mingweishih:
bors try

cmake/configure_lvi_mitigation_build.cmake
```diff
@@ -6,12 +6,29 @@ function(configure_lvi_mitigation_build LVI_MITIGATION_BINDIR)
     if (NOT EXISTS "${LVI_MITIGATION_BINDIR}")
         message(FATAL_ERROR "${LVI_MITIGATION_BINDIR} does not exist.")
     endif()
-    # Perfer clang over gcc.
-    if (EXISTS "${LVI_MITIGATION_BINDIR}/clang-7")
-        set(CMAKE_C_COMPILER ${LVI_MITIGATION_BINDIR}/clang-7 PARENT_SCOPE)
-        set(CMAKE_CXX_COMPILER ${LVI_MITIGATION_BINDIR}/clang++-7 PARENT_SCOPE)
+
+    # Default to clang-7 unless CC or CXX environment variable is explicitly set.
+    set(C_COMPILER clang-7)
```
anakrish, line 11: Can this be the latest version of installed clang rather than clang-7?

mingweishih: This script is used when building OE, which we explicitly set the requirement to clang-7. Do you think we need to add logic to differentiate the scenarios between building OE and apps?

jhand2:
Failure looks infra related

bors try

cmake/configure_lvi_mitigation_build.cmake
jhand2: OE emits a warning if the clang version is not 7, but will still build. We should probably consider doing the same here.

mingweishih: @anakrish  I've updated the PR such that when building the samples, the default will be clang instead of clang-7 (consistent to the make behavior). On the SDK build, clang-7 remains the default.

@jhand2 The warning is preserved with this PR.  Below is the example of specifying clang-8 when building OE.

`
CMake Warning at cmake/compiler_settings.cmake:8 (message):
  Open Enclave officially supports Clang 7 only, but your Clang version
  (8.0.0) is older or newer than that.  Build problems may occur.
`

cmake/configure_lvi_mitigation_build.cmake
```diff
@@ -1,25 +1,81 @@
 # Copyright (c) Open Enclave SDK contributors.
 # Licensed under the MIT License.
 
+macro(detect_compiler BINDIR CC)
+    # Compiler detection logic:
+    # 1. Use either the default compiler or the one specified via CC/CXX.
+    # 2. If the compiler in 1. does not exist and equals to clang/clang++,
+    #    the logic tries to pick the clang/clang++ with specific versions (order: 9, 8, 7).
```
jhand2, line 8: I think if someone specifies a compiler with CC that doesn't exist we should just not worry about it and fail. That is, blindly use CC/CXX no matter what they are.

cmake/configure_lvi_mitigation_build.cmake
```diff
@@ -1,25 +1,81 @@
...
+    # 3. If no compiler in 2. is found, the logic picks gcc/g++.
```
jhand2, line 9: So I would say the logic should be:

1. Use CC/CXX
2. If CC/CXX are not set, default to clang
3. If no clang, default to gcc

mingweishih: I agree on that if CC is set, the logic just picks whatever it is.
However, when CC is not set, the current logic of using make is trying to use the newer version of clang, see https://github.com/openenclave/openenclave/blob/master/samples/config.mk.

The main reason is that the default clang version is old (for example, it's version 6 on ubuntu 18.04 if install via `apt install`).

mingweishih: @anakrish can you add on this?

mingweishih: @jhand2 I've update the PR. Now it does
1. Use CC/CXX (and fail if the specified compiler is not found).
2. If CC/CXX is not set, default to clang/clang++
3. If no clang/clang++, try clang-9/8/7 (this is what samples/config.mk does)
4. If no compiler found in 3, fallback to gcc/g++.

jhand2: Looks good to me. Thanks!

cmake/configure_lvi_mitigation_build.cmake
```diff
@@ -1,25 +1,92 @@
...
-        ${LVI_MITIGATION_BINDIR}/clang++-7
-        PARENT_SCOPE)
+
+  if (NOT OE_IN_PACKAGE)
+    # Default to clang-7 when building SDK.
+    set(C_COMPILER clang-7)
+    set(CXX_COMPILER clang++-7)
+  else ()
+    # Default to clang when building enclave applications.
+    set(C_COMPILER clang)
```
anakrish, line 57: If clang was installed via apt and if a newer version of clang was also installed, then defaulting to clang would cause the older version (via apt) to be picked up. Can you confirm?

mingweishih: Yes, I can confirm that. And I think this is an expected behavior. If user does not specify a version of clang via `CC`, The version of clang associated to the `clang` binary in the path will be invoked.

anakrish:
@mingweishih Sorry, looks like you have pinged me many times on this PR, and I didn't get any notifications. Looks like I had not been "Watching" the openenclave repo. Maybe that should notify me of pings?

mingweishih:
> @mingweishih Sorry, looks like you have pinged me many times on this PR, and I didn't get any notifications. Looks like I had not been "Watching" the openenclave repo. Maybe that should notify me of pings?

I'm not sure about that. Even I'm not "watching" the repo 😄

anakrish:
> > @mingweishih Sorry, looks like you have pinged me many times on this PR, and I didn't get any notifications. Looks like I had not been "Watching" the openenclave repo. Maybe that should notify me of pings?
> 
> I'm not sure about that. Even I'm not "watching" the repo smile

Hmmm...looks like all the pings are buried in mail threads. Wonder if there is a better way to get notified

cmake/configure_lvi_mitigation_build.cmake
nonpolarity: @mingweishih I notice that on my Ubuntu there are binaries clang++-7 and clang-7 from clang 7, but no clang. Do we have "apt install clang" before this part?

mingweishih: This is intentional that makes the installed clang as a default compiler. If clang is not installed, the following logic will try to use clang-9/8/7.

mingweishih:
> LGTM
> 
> Is there a way to test the compiler-detection logic?

Currently, the the detection between gcc and clang will be tested on CI. The rest may defer to when OE supports multiple clang versions.

PR #2938, opened 2020-04-24
Ignore MSVC-compatible clang flags in clangw

### LINKED ISSUE #2857: SGX:Windows: clangw needs to handle -ms-compatibility arguments correctly
`clangw` (a wrapper over clang) is used to cross compile enclaves on Windows to Linux ELF binaries.
Its purpose is to convert MSVC flags to GCC ones, to ignore flags that have no mapping, and to pass in the `-target x86_64-pc-linux` flag to request cross-compilation.

Flags related to [MSVC compatibility](https://clang.llvm.org/docs/MSVCCompatibility.html), override the cross-compilation flag, and therefore must be ignored.

mingweishih:
This PR updates the clangw script to ignore specific flags for MSVC compatibility, which would otherwise cause the compilation to fail (e.g., suppressing the `-target x86_64-pc-linux` flag) or other unwanted effects. Refer to the [clang docs](http://clang.llvm.org/docs/UsersManual.html#microsoft-extensions) for the list of the flags.

Fixes #2857 

### DIFF
--- a/scripts/clangw
+++ b/scripts/clangw
@@ -49,6 +49,14 @@ function call_clang {
         [ "$a" == "/W4" ]           && continue
         [ "$a" == "/WX" ]           && continue
 
+        # Ignore specific clang flags for the MSVC compatibility.
+        # See http://clang.llvm.org/docs/UsersManual.html#microsoft-extensions for more detail.
+        [ "$a" == "-fms-compatibility" ]            && continue
+        [ "$a" == "-fms-extension" ]                && continue
+        [ "$a" == "-fdelayed-template-parsing" ]    && continue
+        # Use regex to match the value-passing flag.
+        [[ "$a" =~ "-fms-compatibility-version=" ]] && continue
+
         # Ignore warnings for specific error codes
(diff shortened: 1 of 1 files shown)


anakrish:
bors try

scripts/clangw
```diff
@@ -88,6 +96,7 @@ function call_clang {
 
     # Call clang with the transformed arguments arguments
     if [ $linking ]; then
+        # shellcheck disable=SC2230
```
CodeMonkeyLeet, line 99: Side note: why don't we just [fix this as recommended](https://github.com/koalaman/shellcheck/wiki/SC2230)? Alternatively, this check seems to have swapped to off by default in v0.7.1 and maybe we also want to consider updating our shellcheck version dependencies as a separate task (v0.4.7 on Linux and v0.7.0 on Windows currently).

jhand2: `command -v` doesn't work the save as `which` and does the wrong thing in many cases. In particular, it doesn't follow aliases. For example, on my fedora system `command -v grep` returns `alias grep='grep --color=auto'`.

I think ignoring the check is the right thing to do here.

PR #2939, opened 2020-04-24
one more child process case

jinghe-INTC:
add one case: create an enclave and then do fork(), create another enclave in the child process

### DIFF
--- a/tests/child_process/host/host.cpp
+++ b/tests/child_process/host/host.cpp
@@ -15,6 +15,7 @@
 
 #define ECALL_IN_CHILD_PROCESS 0
 #define DESTROY_IN_CHILD_PROCESS 1
+#define CREATE_IN_CHILD_PROCESS 2
 
 bool multi_process_flag = true;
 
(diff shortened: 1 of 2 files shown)


jinghe-INTC:
bors try

radhikaj:
bors try

BRMcLaren:
bors try

BRMcLaren:
bors try-

BRMcLaren:
bors try

anakrish:
@jinghe-INTC Can you signoff your PR?

tests/child_process/host/host.cpp
```diff
@@ -101,6 +102,35 @@ int main(int argc, const char* argv[])
...
+                    0,
+                    &enclave_in_child_process);
+                OE_TEST(result == OE_OK);
+                oe_terminate_enclave(enclave_in_child_process);
+                _exit(EXIT_SUCCESS);
+            }
+            else if (pid > 0) // parent process
+            {
+                fprintf(stdout, "parent pid = %d\n", getpid());
+                sleep(5);
```
mingweishih, line 124: Can you also add comment about the need of `sleep` here?

radhikaj:
Bors try

tests/child_process/host/host.cpp
jinghe-INTC: There used to be an issue in creating enclave in child process occasionally. It was fixed, so it doesn't need to repeat creating again and again any more. I can remove the sleeping in the parent process since it is really unnecessary.

mingweishih: Thanks!

mingweishih: It would be great if you can squash the comments into a single comment.

jinghe-INTC: Done.

mingweishih: Thanks so much!

PR #2942, opened 2020-04-24
Update docs of installation OE on windows

mingweishih:
This PR updates the documentation as follows.
- `WindowsManualSGX1FLCDCAPPrereqs.md`.
  - Fixing the indent --- Intel DCAP should not the sub-section of  Azure DCAP Windows
  - Re-organize---separating the nuget and driver installation

Update the installation path. The current behavior of installation via the nuget package is that if specifying the installation path to `C:\oe`, the corresponding path to the openenclave would be `C:\oe\open-enclave\openenclave`.

### DIFF
--- a/docs/GettingStartedDocs/Contributors/WindowsManualSGX1FLCDCAPPrereqs.md
+++ b/docs/GettingStartedDocs/Contributors/WindowsManualSGX1FLCDCAPPrereqs.md
@@ -17,18 +17,27 @@ C:\Intel SGX PSW for Windows v2.7.101.2\PSW_EXE_RS2_and_before\Intel(R)_SGX_Wind
 
 Note that this is optional since you can choose an alternate implementation of the DCAP client or create your own.
 The Azure DCAP client for Windows is necessary if you would like to perform enclave attestation on a Azure Confidential Computing VM. It is available from [nuget.org](https://www.nuget.org/packages/Microsoft.Azure.DCAP/) and can be installed directly via command below.
-This example assumes that `C:\oe_prereqs` is where you would like the prerequisites to be installed.
 
 ```cmd
 nuget.exe install Microsoft.Azure.DCAP -ExcludeVersion -Version 1.3.0 -OutputDirectory C:\oe_prereqs
 ```
 This example assumes you would like to install the package to `C:\oe_prereqs`.
 
-##### [Intel Data Center Attestation Primitives (DCAP) Libraries v1.6](http://registrationcenter-download.intel.com/akdlm/irc_nas/16620/Intel%20SGX%20DCAP%20for%20Windows%20v1.6.100.2.exe)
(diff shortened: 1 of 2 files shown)


samples/README_Windows.md
```diff
@@ -39,7 +39,7 @@ https://docs.microsoft.com/en-us/dotnet/framework/tools/developer-command-prompt
 As an example, if the Open Enclave SDK is installed to `C:\openenclave`, then you would set `CMAKE_PREFIX_PATH` as shown below
 
 ```cmd
-set CMAKE_PREFIX_PATH=C:\openenclave\lib\openenclave\cmake
+set CMAKE_PREFIX_PATH=C:\oe\open-enclave\openenclave\lib\openenclave\cmake
```
radhikaj, line 42: Does this match all the samples? https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/WindowsSGX1FLCGettingStarted.md#buildinginstallation-on-windows-using-developer-command-prompt for example, we set the install path as c:\openenclave. And that document follows into this. Can we have the same path everywhere?

dthaler: All the docs ought to match.  It looks like this PR updates some places and not other places, making the docs inconsistent.  For example, it looks like the following docs still use `C:\openenclave`:
* GettingStartedDocs\Windows_vscode.md
* GettingStartedDocs\Contributors\WindowsInstallInfo.md
* GettingStartedDocs\Contributors\building_oe_sdk.md
* GettingStartedDocs\Contributors\WindowsSGX1FLCGettingStarted.md
* GettingStartedDocs\Contributors\WindowsSGX1GettingStarted.md

I think all the docs should be kept consistent one way or the other before this is merged.

mingweishih: @dthaler GettingStartedDocs\Windows_vscode.md uses the path not only in text but also in screenshots. I'd prefer open another follow-up issue for updating the this specific documentation (i.e., requiring someone tests the vscode debugging while updating). What do you think?

dthaler: I think it's easier to standardize on c:\openenclave and not c:\oe in all examples, since most docs already use the former.   Then I don't think you need to modify the other files, just revert some changes in this PR.  What do you think?

mingweishih: Sure. I think the only inconsistency part is really the instruction on the installation via the nuget package ([doc](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/install_oe_sdk-Windows.md#download-and-install-the-open-enclave-sdk-nuget-package)).

Following this instruction, the openenclave is installed to the C:\oe\open-enclave\openenclave rather than C:\openenclave, which is currently the only place uses this path among the documentation.

The easy fix would be just adding an instruction/comment that resolves/explains the difference here. Does it sound good to you?

dthaler: I didn't follow, but if there's a difference then sure go ahead and propose text and maybe I'll understand then :)   Thanks for looking into the consistency issue.

mingweishih: Please check the updated version. Thanks!

PR #2947, opened 2020-04-27
Fix the thread event and wait/wake functions

### LINKED ISSUE #85: Windows: WaitForSingleObject() or WaitOnAddress()
Consider using WaitOnAddress rather than WaitForSingleObject.

This is in response to comments on PR #75. In the Linux implementation, enclave thread wake/wait events are implemented using a process private futex (`FUTEX_WAIT_PRIVATE`/`FUTEX_WAKE_PRIVATE` attributed syscalls), whereas the Windows implement uses Event objects and `SetEvent()`/`WaitForSingleObject()` methods.

This has semantic and perf implications as Windows Events are cross-process. The analogous method of implementing per process events on Windows is usually to use `WaitOnAddress()` instead which is much lighter weight although there's a little bit of extra work to handle event semantics (e.g. auto-reset)

_Edit: Fixed `WaitForSingleEvent()` to `WaitForSingleObject()`_

### LINKED ISSUE #2920: pthread and oethread timing out
The recent additions of TestErrnoMultiThreadsSameenclave and TestErrnoMultiThreadsDifferentenclave  are causing pthread and oethread tests to timeout from time to time causing discrepancies in CI. We are investigating this. In the mean time, we have turned off these tests. See PR #2919.

nonpolarity:
1. On Windows, replace WaitForSingleObject() with WaitOnAddress(), unify EnclaveEvent struct as on Linux.
Fixes #85

2. Let wake thread holds until wait thread starts.
There is a possible catch that if HandleThreadWake() is execute earlier than
HandleThreadWait(), HandleThreadWait() will miss the wake signal then
hang there forever. To avoid such situation HandleThreadWake() must
wait until HandleThreadWait() starts.
partially fix #2920

### DIFF
--- a/host/sgx/ocalls.c
+++ b/host/sgx/ocalls.c
@@ -33,6 +33,15 @@
 #include "quote.h"
 #include "sgxquoteprovider.h"
 
+// There is a possible catch that if HandleThreadWake() executes earlier than
+// HandleThreadWait(), HandleThreadWait() will miss the wake signal then
+// hangs there forever. To avoid such situation HandleThreadWake() must wait
+// until HandleThreadWait() starts.
+// Caller should still start HandleThreadWait() earlier than HandleThreadWake()
+// to make this case a small probability, then no extra performance loss.
+// Caller also must pair HandleThreadWake() with a calling of
+// HandleThreadWait(), in case HandleThreadWake() hang in a infinity loop.
+
 void HandleThreadWait(oe_enclave_t* enclave, uint64_t arg_in)
 {
     const uint64_t tcs = arg_in;
@@ -41,13 +50,15 @@ void HandleThreadWait(oe_enclave_t* enclave, uint64_t arg_in)
 
 #if defined(__linux__)
 
-    if (__sync_fetch_and_add(&event->value, (uint32_t)-1) == 0)
+    if (__sync_fetch_and_add(&event->lock, (uint32_t)-1) == 0)
     {
+        syscall(__NR_futex, &event->lock, FUTEX_WAKE_PRIVATE, 1, NULL, NULL, 0);
+
         do
         {
             syscall(
                 __NR_futex,
-                &event->value,
+                &event->lock,
                 FUTEX_WAIT_PRIVATE,
                 -1,
                 NULL,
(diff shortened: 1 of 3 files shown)


nonpolarity:
bors try

jhand2:
For my knowledge, what was wrong with WaitForSingleObject? Is the idea that the new code is simpler?

dthaler:
> For my knowledge, what was wrong with WaitForSingleObject? Is the idea that the new code is simpler?

See Issue #85

host/sgx/ocalls.c
```diff
@@ -41,28 +41,26 @@ void HandleThreadWait(oe_enclave_t* enclave, uint64_t arg_in)
...
     }
 
 #elif defined(_WIN32)
 
-    WaitForSingleObject(event->handle, INFINITE);
+    int32_t zero = 0;
+    int32_t one = 1;
+    if (_InterlockedCompareExchange(event, one, zero) == one)
+    {
+        WaitOnAddress(event, &one, sizeof(event), INFINITE);
```
anakrish, line 62: It has to be `sizeof(*event)` which would be 4 bytes instead of 8.

anakrish: Can you double check the logic?

As written it seems to be:

```c
    if (_InterlockedCompareExchange(event, one, zero) == one)  <- if event is one, then set it to zero
    {
        WaitOnAddress(event, &one, sizeof(event), INFINITE);     <- Wait until event is not one? This would return immediately?
```

nonpolarity:
> 
> 
> For my knowledge, what was wrong with WaitForSingleObject? Is the idea that the new code is simpler?

@jhand2 Actually WaitForSingleObject is simpler and error-proof, but WaitOnAddress is lightweight that no more transition to kernel mode.

host/sgx/ocalls.c
nonpolarity: > 
> 
> It has to be `sizeof(*event)` which would be 4 bytes instead of 8.

Thanks, yes sizeof(*event) is right.

nonpolarity: You are right again there is some error. Change to if (_InterlockedCompareExchange(event, one, zero) == zero).

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

BRMcLaren:
bors try-

BRMcLaren:
bors try

nonpolarity:
bors try-

nonpolarity:
bors try

nonpolarity:
bors try

host/sgx/ocalls.c
```diff
@@ -41,28 +41,32 @@ void HandleThreadWait(oe_enclave_t* enclave, uint64_t arg_in)
...
+    uint32_t LOCK_IS_FREE = 0;
+    uint32_t LOCK_IS_TAKEN = 1;
+    // Change the event to LOCK_IS_TAKEN from LOCK_IS_FREE.
+    if (_InterlockedCompareExchange(event, LOCK_IS_TAKEN, LOCK_IS_FREE) == LOCK_IS_FREE)
+    {
+        // Waiting while event is LOCK_IS_TAKEN.
+        // Add a loop in case of fake waking.
+        do
+        {
+            WaitOnAddress(event, &LOCK_IS_TAKEN, sizeof(*event), INFINITE);
```
radhikaj, line 67: @nonpolarity  I think this should be WaitOnAddress(event, &LOCK_IS_FREE, sizeof(*event), INFINITE). I might be wrong but https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitonaddress says the second argument should be "A pointer to the location of the previously observed value at Address". 


UINT32 myVar = TRUE;

void Wait()
{
        UINT32 unDesiredValue = FALSE;
        InterlockedExchange(&myVar, unDesiredValue);        
         WaitOnAddress(&myVar, &unDesiredValue, sizeof(unDesiredValue), INFINITE);


 }

void SetEvent()
{
    if (!myVar)
    {

        InterlockedExchange(&myVar, TRUE);
        WakeByAddressAll((PVOID)&myVar);
    }
}

nonpolarity: Thanks @radhikaj , I will check and test.

radhikaj: @nonpolarity, I read the code wrong. Before the wait the value on event is lock-taken. So your parameters are correct. Ignore previous comment

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

radhikaj:
Thanks @nonpolarity 
Lets reopen this with perf data,

Issue #2903, opened 2020-04-21
General design pattern for plugins


dthaler:
We have had a number of design discussions around different types of plugins, with some common items like how do plugins get loaded and initialized.  This issue tracks having a generic design pattern that different types of plugins can copy as appropriate.

Related issues on specific types of plugins:
* Attestation plugins: Issue #2729 PR #2801 
* Memory allocator plugins: Issue #2023 PR #1420
* Possibility of enclave loader plugins: #2844
* Possibility of logging plugins: Issue #2895 
* Possibility of crypto plugins: Issues #852 and #1003

Some of the above issues/PRs have discussion relevant to plugins in general that should be copied or moved to this issue.

dthaler:
The file https://github.com/jhand2/openenclave/blob/global_module_initialization/docs/DesignDocs/openenclave_modules.md is relevant, at least as a cross reference

PR #2949, opened 2020-04-27
Proposal of new OE SDK attestation public and plugin API

shnwc:
Proposal of a new OE SDK attestation public and plugin API for V0.10 release, as an evolution of the experimental attestation API implemented in the OE SDK V0.9 release.

### DIFF
--- a/docs/DesignDocs/Attestation_API_Proposal.md
+++ b/docs/DesignDocs/Attestation_API_Proposal.md
@@ -1,0 +1,608 @@
+Proposal of OE SDK Attestation Public and Plugin API for V0.10 Release
+====
+
+# Introduction
+
+This document describes the proposed OE SDK attestation public and plugin API
+for V0.10 release, as an evolution of the experimental attestation API
+implemented in the
+[OE SDK V0.9 release](https://github.com/openenclave/openenclave/tree/v0.9.x).
+
+This proposal is based on these efforts:
+- Design document [Attestation V3 Update]:
+[Custom Attestation Data Formats for Open Enclave V3 Updates](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/CustomAttestation_V3.md).
+- Discussion in [Issue #2729]: [Outstanding issues on attestation plugin design](https://github.com/openenclave/openenclave/issues/2729).
+- Design document [quote-ex Integration]:
+[Attestation: OE SDK Integration with Intel® SGX SDK quote-ex Library for Generation of Evidence in New Formats](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/SGX_QuoteEx_Integration.md).
+- Design document [Design Notes]:
+[Notes on OE SDK Attestation API and Plugin Library Design](https://github.com/shnwc/openenclave/blob/master/docs/DesignDocs/NotesOnAttestationAPI.md).
+  - Note: this document is in the process of being up-streamed as OE SDK [PR #2801](https://github.com/openenclave/openenclave/pull/2801).
+- IETF draft [RATS Arch]
+[Remote Attestation Procedures Architecture](https://tools.ietf.org/html/draft-ietf-rats-architecture)
+
+Note: the proposed API and plugin design will continue to be marked as
+"experimental", to have the flexibility for further improvements in future
+releases.
+
+# OE SDK Attestation Public API Proposal
+
+Note: in the RATS model as described in the
+[[RATS Arch]](https://tools.ietf.org/html/draft-ietf-rats-architecture)
+document, there are three roles: an Attester, a Verifier, and a Relying Party.
+The OE SDK attestation stack V0.10 does not try to provide a complete implementation of
+these roles, nor try to provide means for secure communication between them.
+The OE SDK attestation stack implements basic functionalities for evidence
+generation and verification, and exposes these functionalities to applications
+via a set of public API functions. Application software can use this public API
+to implement complete RATS Attester and Verifier roles. The application software
+is also responsible to convert the evidence verification results into signed
+Attestation Results for consumption by a Relying Party role, if this role is
+not in the same security domain as the Verifier role. But the proposed API
+readily supports applications functioning as a composite device implementing
+both the Verifier and Relying Party roles in the same security domain, since
+in this case the Attestation Results passed from the Verifier role to the
+Relying Party role do not need to be signed.
+
+## Existing OE SDK V0.9 Public API
+
+The OE SDK V0.9 experimental public API for attestation is composed of
+functions for a number of use cases to be implemented by applications.
+The use cases fall into two categories: evidence generation and verification.
+
+Use cases for evidence generation:
+
+- Retrieve a single SGX attester plugin.
+  - Function `oe_attester_t* oe_sgx_plugin_attester(void)`,
+  declared in header file [openenclave/attestation/sgx/attester.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/sgx/attester.h).
+  - It returns a single hardcoded SGX attester plugin for generation of
+  evidence in SGX ECDSA-p256 (simply called ECDSA) format.
+- Register an attester plugin.
+  - Function `oe_result_t oe_register_attester(oe_attester_t* plugin, const void* config_data, size_t config_data_size)`, declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
+  - The attester plugin returned by `oe_sgx_plugin_attester()` can be
+  registered with this function.
+  - Note: requirement for the application to supply a plugin-specific `config_data`
+  parameter is against the goal of a TEE-agnostic OE SDK API. More discussion on
+  TEE agnostic design can be found in the
+  [[Design Notes]](https://github.com/shnwc/openenclave/blob/master/docs/DesignDocs/NotesOnAttestationAPI.md#attester--verifier-security-model-and-tee-agnostic-design)
+  document.
+- Get evidence in a globally unique format, optionally along with a set of
+endorsements.
+  - Function `oe_result_t oe_get_evidence(const oe_uuid_t* evidence_format_uuid, uint32_t flags, const oe_claim_t* custom_claims, size_t custom_claims_length, const void* opt_params, size_t opt_params_size, uint8_t** evidence_buffer, size_t* vidence_buffer_size, uint8_t** endorsements_buffer, size_t* endorsements_buffer_size)`,
+    declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
+  - Upon successful evidence generation, `oe_get_evidence()` returns the
+  generated evidence in a dynamically allocated buffer. It also has the option
+  to return a set of endorsements in another dynamically allocated buffer.
+  After processing of the evidence and the endorsements, the caller is
+  responsible to free these buffers.
+  - The `flags` parameter is a legacy inherited from the legacy API
+  `oe_get_report()` to indicate SGX local or remote attestation.
+  It's redundant -- overlapping with the `evidence_format_uuid` parameter
+  which identifies evidence formats with globally unique UUIDs.
+  - Note: please see the "Semantics of Custom Claims and Optional Parameters"
+  section of the
+  [[Design Notes]](https://github.com/openenclave/openenclave/pull/2801)
+  document for a discussion of the semantics
+  of the `custom_claims` and `opt_params` parameters.
+- Free a dynamically allocated evidence buffer.
+  - Function `oe_result_t oe_free_evidence(uint8_t* evidence_buffer)`,
+  declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h)
+- Free a dynamically allocated endorsements buffer.
+  - Function `oe_result_t oe_free_endorsements(uint8_t* endorsements_buffer)`,
+  declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h)
+- Unregister an attester plugin.
+  - Function `oe_unregister_attester(oe_attester_t* plugin)`,
+  declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
+
+Use cases for evidence verification:
+
+- Retrieve a single SGX verifier plugin.
+  - Function `oe_verifier_t* oe_sgx_plugin_verifier(void)`, declared in header
+  file [openenclave/attestation/sgx/verifier.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/sgx/verifier.h).
+  - It returns a single hardcoded SGX verifier plugin for verification of
+  evidence in SGX ECDSA format.
+- Register a verifier plugin.
+  - Function `oe_result_t oe_register_verifier(oe_verifier_t* plugin, const void* config_data, size_t config_data_size)`, declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
+  - The verifier plugin returned by `oe_sgx_plugin_verifier()` can be
+  registered with this function.
+  - See notes for the `oe_register_attester()` API function for explanation of
+  the `config_data` parameter.
+- Verify evidence, optionally using a set of input endorsements and policies.
+  - Function `oe_result_t oe_verify_evidence(const uint8_t* evidence_buffer, size_t evidence_buffer_size, const uint8_t* endorsements_buffer, size_t endorsements_buffer_size, const oe_policy_t* policies, size_t policies_size, oe_claim_t** claims, size_t* claims_length)`,
+  declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
+  - Upon successful verification, `oe_verify_evidence` returns a list of claims
+  in a dynamically allocated buffer. This list includes all the custom claims
+  input to the call to `get_evidence()` that generated the verified evidence.
+    - Note: in case a custom claim has the same name as a well-known base claim,
+    the returned list does not have flags indicating which is base and which is
+    custom.
+- Free a dynamically allocated claims buffer.
+  - Function `oe_result_t oe_free_claims_list(oe_claim_t* claims, size_t claims_length)`,
+  declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h)
+- Unregister a verifier plugin.
+  - Function `oe_unregister_verifier(oe_verifier_t* plugin)`,
+  declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
+
+The single supported SGX evidence format UUID, OE_SGX_PLUGIN_UUID for SGX
+ECDSA format, is declared in header file
+[openenclave/internal/sgx/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/internal/sgx/plugin.h).
+
+In addition to the above plugin-based public API for attestation, the OE SDK
+also supports legacy API for generation and verification of SGX reports:
+`oe_get_report()` and `oe_verify_report()`.
+
+Areas of improvement in the existing attestation public API:
+- Attester and Verifier plugins are unnecessarily exposed to application
+software
+  - It's sufficient for application to use globally unique evidence format
+  UUID in generation and verification of evidence.
+  - Plugins are the artifact of the OE SDK attestation framework internal
+  design that application software does not need to care.
+- The API does not enable applications to discover which formats are supported
+for evidence generation or verification.
+- SGX specific functions are exposed as public API
+- The public and plugin API declarations are all contained in a
+single header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
+- For SGX, the globally unique UUIDs identify the evidence formats.
+There is no need of local / remote flag.
+
+## Proposed OE SDK V0.10 Public API
+
+Objectives: address the areas of improvement in the existing public API.
+
+- Separate public and plugin API declarations into different header files.
+- Only expose evidence formats for evidence generation and verification.
+There is no exposure of plugin design.
+- Remove TEE specific API elements, e.g. SGX-specific functions
+- Enable applications run-time discovery of supported formats for evidence
+generation and verification in
+[TEE-agnostic](https://github.com/shnwc/openenclave/blob/master/docs/DesignDocs/NotesOnAttestationAPI.md#attester--verifier-security-model-and-tee-agnostic-design)
+manner.
+- Remove legacy artifacts that have become redundant, e.g. the SGX
+local / remote flags.
+
+Proposed Headers files:
+
+- Attester public API to be declared in `openenclave/attestation/attester.h`
+- Verifier public API to be declared in `openenclave/attestation/verifier.h`
+- Plugin API (helper functions and plugin entry points) to be declared in
+`openenclave/internal/plugin.h`
+- SGX evidence format UUIDs (and any other SGX-specific plugin related names)
+to be declared in `openenclave/internal/sgx/plugin.h`
+
+Use cases for evidence generation:
+
+- Initialize attester environment.
+  - Function `oe_result_t oe_attester_initialize(void)`.
+  - Internally, the implementation enumerates and registers all
+  attester plugins configured for the platform and the calling
+  application.
+  - This function is idempotent and can be called multiple times without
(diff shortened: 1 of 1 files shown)


docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,534 @@
...
+## Existing OE SDK V0.9 Public API
+
+The OE SDK V0.9 experimental public API for attestation is composed of
+functions for a number of use cases to be implemented by applications.
+The use cases fall into two categories: evidence generation and verification.
+
+Use cases for evidence generation:
+
+- Retrieve a single SGX attester plugin.
+  - Function `oe_attester_t* oe_sgx_plugin_attester(void)`,
```
dthaler, line 32: I'd recommend that all APIs have a verb.  So oe_sgx_get_plugin_attester() or oe_get_sgx_plugin_attester(), if this API is needed.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,534 @@
...
+- Register an attester plugin.
+  - Function `oe_result_t oe_register_attester(oe_attester_t* plugin, const void* config_data, size_t config_data_size)`, declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
+  - The attester plugin returned by `oe_sgx_plugin_attester()` can be
+  registered with this function.
+  - Note: requirement for the application to supply a plugin-specific `config_data`
+  parameter is against the goal of a TEE-agnostic OE SDK API. More discussion on
+  TEE agnostic design can be found in the
+  [[Design Notes]](https://github.com/shnwc/openenclave/blob/master/docs/DesignDocs/NotesOnAttestationAPI.md#attester--verifier-security-model-and-tee-agnostic-design)
+  document.
+- Get an evidence in a globally unique format, optionally along with a set of
```
dthaler, line 45:
```suggestion
- Get evidence in a globally unique format, optionally along with a set of
```

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,534 @@
...
+  - Function `oe_result_t oe_get_evidence(const oe_uuid_t* evidence_format_uuid, uint32_t flags, const oe_claim_t* custom_claims, size_t custom_claims_length, const void* opt_params, size_t opt_params_size, uint8_t** evidence_buffer, size_t* vidence_buffer_size, uint8_t** endorsements_buffer, size_t* endorsements_buffer_size)`,
+    declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
+  - Upon successful evidence generation, `oe_get_evidence()` returns the
+  generated evidence in a dynamically allocated buffer. It also has the option
+  to return a set of endorsements in another dynamically allocated buffer.
+  After processing of the evidence and the endorsements, the caller is
+  responsible to free these buffers.
+  - The `flags` parameter is a legacy inherited from the legacy API
+  `oe_get_report()` to indicate SGX local or remote attestation.
+  It's redundant -- overlapping with the `evidence_format_uuid` parameter
```
dthaler, line 56: Is it possible to remove it, to get rid of redundancy?

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,534 @@
...
+  - Function `oe_result_t oe_free_endorsements(uint8_t* endorsements_buffer)`,
+  declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h)
+- Unregister an attester plugin.
+  - Function `oe_unregister_attester(oe_attester_t* plugin)`,
+  declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
+
+Use cases for evidence verification:
+
+- Retrieve a single SGX verifier plugin.
+  - Function `oe_verifier_t* oe_sgx_plugin_verifier(void)`, declared in header
```
dthaler, line 71: add _get_ into name

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,534 @@
...
+  file [openenclave/attestation/sgx/verifier.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/sgx/verifier.h).
+  - It returns a single hardcoded SGX verifier plugin for verification of
+  evidences in SGX ECDSA format.
+- Register a verifier plugin.
+  - Function `oe_result_t oe_register_verifier(oe_verifier_t* plugin, const void* config_data, size_t config_data_size)`, declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
+  - The verifier plugin returned by `oe_sgx_plugin_verifier()` can be
+  registered with this function.
+  - See notes for the `oe_register_attester()` API function for explanation of
+  the `config_data` parameter.
+- Verify an evidence, optionally using a set of input endorsements and policies.
```
dthaler, line 81:
```suggestion
- Verify evidence, optionally using a set of input endorsements and policies.
```

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,534 @@
...
+  - Function `oe_result_t oe_verify_evidence(const uint8_t* evidence_buffer, size_t evidence_buffer_size, const uint8_t* endorsements_buffer, size_t endorsements_buffer_size, const oe_policy_t* policies, size_t policies_size, oe_claim_t** claims, size_t* claims_length)`,
```
dthaler, line 82: Often endorsements will not be supplied across the same protocol as evidence, and the verifier will need to fetch them separately.   Is that intended to be done by the plugin itself underneath this api (i.e., when endorsements buffer is NULL)?

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,534 @@
...
+  declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
+  - Upon successful verification, `oe_verify_evidence` returns a list of claims
```
dthaler, line 84: This seems odd to me.  I would expect an API that would take in an Attestation Result format id, and return an Attestation Result buffer (in the requested format) that can be passed to a relying party.   I would think that the verifier doesn't care what the claims are, only the relying party does, so a separate API would take in an attestation result and be able to give information about claims.   Combining them only makes sense, I think, as an optimization for the special case when the Verifier and Relying Party are in the same entity.   See https://tools.ietf.org/html/draft-ietf-rats-architecture.   Alternatively, if your definition of "verifier" is really what the RATS architecture calls a "Relying Party" then I would suggest the APIs should be renamed to match the standard terminology.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,534 @@
...
+[Notes on OE SDK Attestation API and Plugin Library Design](https://github.com/shnwc/openenclave/blob/master/docs/DesignDocs/NotesOnAttestationAPI.md).
+  - Note: this document is in the process of being up-streamed as OE SDK [PR #2801](https://github.com/openenclave/openenclave/pull/2801).
+
+# OE SDK Attestation Public API Proposal
+
+## Existing OE SDK V0.9 Public API
+
+The OE SDK V0.9 experimental public API for attestation is composed of
+functions for a number of use cases to be implemented by applications.
+The use cases fall into two categories: evidence generation and verification.
```
dthaler, line 27: As I mention in a later comment, I think there should be 3 categories: Attester, Verifier, and Relying Party.   See https://tools.ietf.org/html/draft-ietf-rats-architecture (of which I would point out that Intel is a co-author and WG chair)

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,534 @@
...
+This proposal is based on these efforts:
+- Design document [Attestation V3 Update]:
+[Custom Attestation Data Formats for Open Enclave V3 Updates](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/CustomAttestation_V3.md).
+- Discussion in [Issue #2729]: [Outstanding issues on attestation plugin design](https://github.com/openenclave/openenclave/issues/2729).
+- Design document [quote-ex Integration]:
+[Attestation: OE SDK Integration with Intel® SGX SDK quote-ex Library for Generation of Evidence in New Formats](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/SGX_QuoteEx_Integration.md).
+- Design document [Design Notes]:
+[Notes on OE SDK Attestation API and Plugin Library Design](https://github.com/shnwc/openenclave/blob/master/docs/DesignDocs/NotesOnAttestationAPI.md).
+  - Note: this document is in the process of being up-streamed as OE SDK [PR #2801](https://github.com/openenclave/openenclave/pull/2801).
+
```
dthaler, line 20: Can you add https://tools.ietf.org/html/draft-ietf-rats-architecture to the list of efforts this should be based on?

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,534 @@
...
+  release.
+- Unregister an attester plugin.
+  - Helper function `oe_result_t oe_unregister_attester(oe_attester_t* plugin)`.
+  - The same definition as in the OE SDK V0.9 release, except that this function
+  is not part of the public API.
+
+In the OE SDK V0.10 release, attester plugins are managed internally between
+the OE SDK framework and plugin libraries, without exposing details to
+application software. As explained in the
+[[Design Notes]](https://github.com/shnwc/openenclave/blob/master/docs/DesignDocs/NotesOnAttestationAPI.md#options-for-enclave-side-plugin-library-initialization)
```
yentsanglee, line 399: In the NotesOnAttestationAPI.md, option 1, 2, and 4 have potential issue.  Option 1 and 2 do not support multiple external plugins. Option 4 is not guaranteed to work universally on all platforms and configurations. Option 3 is a potential solution which issue 2093 is trying to resolve.

Without a proper plugin initialization solution, I'd suggest we either don't change current initialization design or make it clear to the users that such design is purely 'experimental'.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Sure.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: My understanding is that the Relying Party role in the RATS model is played by verifier applications, and thus not in the scope of the public API, but a user of the API to consume evidence verification result. I'll add some notes for clarification.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: The proposed new API will do.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Thanks for the catch! While "evidence" is uncountable, what I see here is something countable: every call to `oe_get_evidence()` returns a different piece of evidence. Maybe should replace "an evidence" with "a piece of evidence"?

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: The proposed new API will do.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: The proposed new API will do.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Thanks! will replace with "a piece of evidence".

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Yes.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,534 @@
...
+Use cases for evidence verification:
+
+- Initialize verifier environment.
+  - Function `oe_result_t oe_verifier_initialize(void)`.
+  - Internally, the implementation enumerates and registers all verifier plugins
+  configured for the platform and the calling application.
+  - This function is idempotent and can be called multiple times without
+  adverse effect.
+- Enumerate all evidence formats that can be verified.
+  - Function `oe_result_t oe_verifier_enumerate_formats(oe_uuid_t** format_list, size_t* format_list_length)`.
```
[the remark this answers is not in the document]
shnwc: Will change.

docs/DesignDocs/Attestation_API_Proposal.md
dthaler: Verifier applications and Relying Party applications are two different things, not to be confused.

docs/DesignDocs/Attestation_API_Proposal.md
dthaler: Per the rats architecture, a piece of evidence would be a specific claim.   A set of claims is what gets signed (at least typically), and the signed set of claims is called evidence.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: As explained in an earlier comment, the OE SDK attestation API function `oe_verify_evidence()` is focused on verification of the input evidence. It outputs the verification result, as a returned error code plus a set of claims contained in the evidence if there is no error. The application that calls this API function implements the complete Verifier role, and is responsible to convert the result into a signed Attestation Result if the Relying Party is in a different security domain.

I do think the built-in base claims in the evidence should be separate from the custom claims. This is because there is no limitation on what names custom claims can have. It's possible for a custom claim to have name conflict a base claim. I'll update the proposal to separate these two types of claims.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Understood. Application developers can use the OE SDK attestation API to build Attester and Verifier roles / applications. This API does not provide functions to build Relying Party application to convert evidence verification results into signed Attestation Result.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Got it. So we stick with the uncountable "evidence".

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,534 @@
...
+- Verifier public API to be declared in `openenclave/attestation/verifiers.h`
+- SGX evidence format UUIDs to be declared in
+`openenclave/attestation/evidence_uuids_sgx.h`
+- Plugin API (helper functions and plugin entry points) to be declared in
+`openenclave/internal/plugin.h`
+
+Use cases for evidence generation:
+
+- Initialize attester environment.
+  - Function `oe_result_t oe_attester_initialize(void)`.
```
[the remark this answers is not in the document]
shnwc: From application developers' point of view, this function sets a requirement for an attester application to declare its intention to start using the plugin-based attester APIs.

Internally, the OE SDK framework / plugin implementation of this API function can do anything as needed to get the environment  set up properly. If resolution of issue #2903 leads to a good plugin design that performs all initialization internally, at that time this function can become NOP.

@yentsanglee on your comment "Such initialization function does not work for external plugins.", I'd like to learn about the cases you see that external plugins can'be be initialized by some sort of implementation of this function.

Legacy function `oe_get_report()` could be implemented by the OE SDK framework to call `oe_attester_initializat()` first, so applications don't need to call `oe_attester_initialize()` explicitly. Since `oe_attester_initialize()` is idempotent, there is no harm calling it multiple times.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Yes, let's mark the plugin design as "experimental", until issue #2903 is resolved.

shnwc: I'll add text to indicate that the proposed API and plugin design continue be marked as  "experimental"

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,590 @@
...
+
+Note: the proposed API and plugin design will continue to be marked as
+"experimental", to have the flexibility for further improvements in future
+releases.
+
+# OE SDK Attestation Public API Proposal
+
+Note: in the RATS model as described in the
+[[RATS Arch]](https://tools.ietf.org/html/draft-ietf-rats-architecture-02)
+document, there are three roles: an Attester, a Verifier, and a Replying Party.
```
radhikaj, line 31:
```suggestion
document, there are three roles: an Attester, a Verifier, and a Relying Party.
```

docs/DesignDocs/Attestation_API_Proposal.md
radhikaj: Today this API supports applications that are combined verifier+relying party. In the long term relying part and verifier could be different. Non blocker.

docs/DesignDocs/Attestation_API_Proposal.md
radhikaj: This is the v0.9 API

docs/DesignDocs/Attestation_API_Proposal.md
radhikaj: > Current SGX verifier plugin retrieves endorsements underneath the api if endorsements_buffer is null. We expect all verifier plugins to follow the same behavior.

This is in the v0.9 section.

docs/DesignDocs/Attestation_API_Proposal.md
radhikaj: This PR is experimental and short term. The long term requirements are expressed in PR#2959

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,590 @@
...
+- `OE_SGX_EPID_UNLINKABLE_PLUGIN_UUID`: for SGX unlinkable EPID evidence format.
+
+These plugins will be implemented by a single set of plugin libraries,
+composed of:
+- One enclave-side SGX evidence generation plugin library.
+  - It implements the OE SDK public API function
+  `oe_attester_initialize()` for attester plugins enumeration and
+  registration, and `oe_attester_shutdown()`.
+- One host-side SGX evidence generation plugin library.
+  - It implements OCALLs for the enclave-side library to access
```
jazzybluesea, line 530: Is doing OCALLs safe to do?  Will make this less secure given that the enclave will rely on the host to get the evidence?  This also adds overhead.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,590 @@
...
+library.
+
+The verifier plugin entry points are implemented in
+[common/sgx/verifier.c](https://github.com/openenclave/openenclave/blob/v0.9.x/common/sgx/verifier.c).
+The plugin entry points are returned in an `oe_verifier_t` structure by the
+implementation of the public API function `oe_sgx_plugin_verifier()`. This
+verifier plugin implementation is linked to both the OE SDK enclave-side and host-side
+static libraries.
+
+## Proposed V0.10 Implementation
```
jazzybluesea, line 512: In order for encourage adoption, I suggest creating samples that shows how to use this API and/or have an implementation guide.

docs/DesignDocs/Attestation_API_Proposal.md
radhikaj: The names for each of the claims would need to be globally unique. In various formats, names are ints, strings, void etc. 
There should never be a case where custom claims and base claims with the same name. The verifier should just fail if this happens.
For future intent: Instead of name value pairs, use a flat buffer. The relying party parses the buffer.
This is not a blocker for the v0.10 API.

docs/DesignDocs/Attestation_API_Proposal.md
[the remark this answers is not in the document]
radhikaj: For v0.10, we will proceed with this proposal. A longer term discussion around plugins needs to happen in SIG-Arch.

docs/DesignDocs/Attestation_API_Proposal.md
jazzybluesea: Resolving this given that there is already existing ocalls.  But I highly suggest avoiding ocalls given the added overhead and increased attack footprint.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: The design document has a section "Application Developer Experience" on a possible application scenario, and ctests will be updated to call the proposed APIs.

shnwc:
Squashed commits.

bors try

shnwc:
bors try

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Added the text between lines 40 - 44.

shnwc:
@radhikaj can you trigger a merge? Thanks!

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -0,0 +1,599 @@
...
+these entry points respectively.
+
+The Legacy functions `oe_get_report()` and `oe_verify_report()` should be
+implemented by the OE SDK framework to first call API functions
+`oe_attester_initializat()` and `oe_verifier_initialize()` respectively,
+since legacy applications don't call these API functions explicitly.
+ As `oe_attester_initializat()` and `oe_verifier_initialize()` are
+ idempotent, there is no harm calling them multiple times. For SGX plugins,
+ `oe_attester_shutdown()` and `oe_verifier_shutdown()` are NOP, so there
+ is no need to call them upon application close.
```
jhand2, line 593:
```suggestion
The Legacy functions `oe_get_report()` and `oe_verify_report()` should be
implemented by the OE SDK framework to first call API functions
`oe_attester_initializat()` and `oe_verifier_initialize()` respectively,
since legacy applications don't call these API functions explicitly.
 As `oe_attester_initializat()` and `oe_verifier_initialize()` are
 idempotent, there is no harm calling them multiple times. For SGX plugins,
 `oe_attester_shutdown()` and `oe_verifier_shutdown()` are NOP, so there
 is no need to call them upon application close.
```
```suggestion
The Legacy functions `oe_get_report()` and `oe_verify_report()` should be
implemented by the OE SDK framework to first call API functions
`oe_attester_initialize()` and `oe_verifier_initialize()` respectively,
since legacy applications don't call these API functions explicitly.
 As `oe_attester_initialize()` and `oe_verifier_initialize()` are
 idempotent, there is no harm calling them multiple times. For SGX plugins,
 `oe_attester_shutdown()` and `oe_verifier_shutdown()` are NOP, so there
 is no need to call them upon application close.
```

docs/DesignDocs/Attestation_API_Proposal.md
CodeMonkeyLeet: I'll note here that the design documents are not generally used as reference documents by end-users, nor are tests; the addition of samples is a requirement for public APIs per the [ApiGuidelines.md](https://github.com/openenclave/openenclave/blob/master/docs/ApiGuidelines.md) and are strongly recommended for experimental APIs. Particularly in this case, the guidance seems relevant:

>Having working samples may help others evaluate and determine how ready it is for prime time before the API moves from experimental to public.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: @jhand2 thank you for catching these strange typos! Just committed the fix.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: @CodeMonkeyLeet agreed! the proposed APIs will still be experimental, and the sample in https://github.com/openenclave/openenclave/tree/master/samples/remote_attestation will be adapted to use the new API. Maybe in this sample, we can have support of both the legacy (production) and new (experimental) APIs and use a macro to switch between them?

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: @jhand2 my last commit had the suggested edits, and was squashed and signed off. Let me know If I didn't do the squash in the correct way.

shnwc:
@jhand2 please help trigger the merge, or let me know if you have further comments. Thanks!

docs/DesignDocs/Attestation_API_Proposal.md
CodeMonkeyLeet: I think it's probably easier for the user of the samples to have them as separate samples instead of having to deal with even more instructions to build one version instead of the other. If the expectation is that the new samples will be direct replacements for the old ones, then I think it's fine to create a duplicate of the old samples and just use the new API calls, although that creates a bit more maintenance overhead. I would also suggest refactoring out shared code in that case that might be reusable as shared library components between samples.

shnwc: @CodeMonkeyLeet agreed that we can have new sample to showcase the new API, and refactor out reusable shared code to minimize maintenance overhead. The samples will be addressed in @yentsanglee's issue #2975

radhikaj:
Bors r+

radhikaj:
Bors r+

PR #2950, opened 2020-04-27
Enable the memory test on Windows

### LINKED ISSUE #2397: Enable test/memory on Windows
The test is currently disabled on the windows build. The cause of the issue is the hard limit of the EPC memory (see #2394). As a result, we may not be able to have the same test on Windows as on Linux.

mingweishih:
This PR enables the memory test on Windows. More detail of the PR is as follows.
- Set smaller heap size (~10MB) on Windows to accommodate to its lack support of paging.
- Update the logic to respect the size of the heap.
- Update the available heap in the enclave: subtracting the amount of heap used by the OE (obtained from manual tests) from the total heap size.

Fixes #2397

### DIFF
--- a/tests/memory/enc/boundaries.c
+++ b/tests/memory/enc/boundaries.c
@@ -9,6 +9,7 @@
 
 #define BUFSIZE 1024
 #define ITERS 1024
+#define OE_USED_HEAP 20584 /* Amount of heap consumed by OE. */
 
 void test_host_boundaries(buffer buf)
 {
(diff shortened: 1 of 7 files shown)


mingweishih:
bors try

tests/memory/enc/boundaries.c
```diff
@@ -17,16 +18,27 @@ void test_host_boundaries(buffer buf)
...
 {
-    void* array[ITERS];
+    void* array[ITERS] = {NULL};
+    /* Calculate the upper bound of the heap. */
+    size_t bound = __oe_get_heap_size() - OE_USED_HEAP;
+    size_t allocated = 0;
+
     for (int i = 0; i < ITERS; i++)
     {
+        if (allocated >= bound)
```
anakrish, line 28: Is this equivalent to just checking if malloc returns NULL and if so, break?

mingweishih: The test (below) uses the OE_TEST to test the return of malloc, so for the consistency, I add the check against the upper bound here.

mingweishih:
bors try

mingweishih:
bors try

PR #2953, opened 2020-04-28
Restore libsgx-urts install

eddyashton:
This reverts part of #2825. That change removed `libsgx-urts` from the list of required Intel SGX packages. However, this package is still transitively required by `libsgx-pce-logic` in order to build Open Enclave successfully. For example if `urts` is still on version 2.8 while the others have updated to 2.9, we get the following error:

```
Linking C executable tests/mem/mem
FAILED: tests/mem/mem 
: && /usr/bin/clang-7 -O2 -g -DNDEBUG   tests/mem/CMakeFiles/mem.dir/main.c.o  -o tests/mem/mem  output/lib/openenclave/host/liboehost.a /usr/lib/x86_64-linux-gnu/libcrypto.so /usr/lib/x86_64-linux-gnu/libdl.so -lpthread /usr/lib/x86_64-linux-gnu/libsgx_enclave_common.so /usr/lib/x86_64-linux-gnu/libsgx_dcap_ql.so -Wl,-z,noexecstack && :
//usr/lib/x86_64-linux-gnu/libsgx_pce_logic.so: undefined reference to `sgx_get_metadata'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.
```

This may not be a complete fix (do we need the same thing in the redhat packages? We didn't have one previously...), but it should be sufficient to fix ansible-configured Ubuntu machines.

### DIFF
--- a/scripts/ansible/roles/linux/intel/vars/ubuntu/main.yml
+++ b/scripts/ansible/roles/linux/intel/vars/ubuntu/main.yml
@@ -14,6 +14,7 @@ intel_sgx_packages:
 intel_dcap_packages:
   - "libsgx-dcap-ql"
   - "libsgx-dcap-ql-dev"
+  - "libsgx-urts"
 
 packages_validation_distribution_files:
   - "/usr/lib/x86_64-linux-gnu/libsgx_enclave_common.so"


eddyashton:
Unrelated: The DCO requirement that all commits are signed off makes it harder to contribute small changes like this via the Edit button in GitHub. Do you know if its possible to add the `--signoff` to these browser-generated commits?

eddyashton:
> Unrelated: The DCO requirement that all commits are signed off makes it harder to contribute small changes like this via the Edit button in GitHub. Do you know if its possible to add the `--signoff` to these browser-generated commits?

Answering my own question: [Not supported natively](https://github.community/t5/How-to-use-Git-and-GitHub/Feature-request-DCO-signoff-through-Github-PR-s/td-p/24784), but there is a [browser extension](https://codeengineered.com/blog/2018/dco-signoff-github-ui/).

letmaik:
Does that mean Intel didn't declare dependencies right in their Debian packages?

jhand2:
@eddyashton Shouldn't the deb package for libsgx-enclave-common pick up these dependencies transitively? If I make a new Ubuntu 18.04 system and install just libsgx-enclave-common and libsgx-dcap-ql, I get the following packages:

```
libsgx-ae-le
libsgx-ae-pce
libsgx-ae-qe3
libsgx-ae-qve
libsgx-aesm-launch-plugin
libsgx-dcap-ql
libsgx-enclave-common
libsgx-launch
libsgx-pce-logic
libsgx-qe3-logic
libsgx-urts
```

eddyashton:
Huh, you're right, it looks like this is declared as a dependency, so will be installed fresh:

```
$ apt-cache depends
 libsgx-pce-logic
libsgx-pce-logic
  Depends: libsgx-urts
  Depends: libsgx-ae-pce
  Depends: libc6
  Depends: libstdc++6
```

However, this problem occurs when trying to use the ansible scripts on a machine that already has an old version installed. The ansible script installs the _latest_ version of the _listed_ packages, but their dependencies can still be met by an old version of `libsgx-urts`. That's where my apt-fu ends - I don't know if the Intel packages could/should be declaring a minimum version on their dependencies?

letmaik:
> I don't know if the Intel packages could/should be declaring a minimum version on their dependencies?

They *totally* should. If they don't declare exactly what they need then it's broken. Big time.

jhand2:
Which ansible playbook are you using?

eddyashton:
> Which ansible playbook are you using?

This was reported via the `oe-contributors-acc-setup-no-driver.yml` playbook, but it should be equally present in all of them (since they all appear to import `sgx-packages.yml`).

jhand2:
Looking at `sgx-packages-install.yml`, we set `install_recommends` to `no` (https://docs.ansible.com/ansible/latest/modules/apt_module.html). I believe this is the root of why the package was not updated.

@ionutbalutoiu @BRMcLaren do you know why we do that? For SGX packages can we changes that to yes?

BRMcLaren:
@jhand2 This was before my time on the project and I do not have insight into why this was done, unfortunately.

My instinct is to try it and see what happens, this was probably done to protect against breaking changes in an upgrade scenario, but that is a guess at best.

ionutbalutoiu:
> Looking at `sgx-packages-install.yml`, we set `install_recommends` to `no`. See the [ansible documentation for this field](https://docs.ansible.com/ansible/latest/modules/apt_module.html#parameter-install_recommends). I believe this is the root of why the package was not updated.
> 
> @ionutbalutoiu @BRMcLaren do you know why we do that? For SGX packages can we changes that to yes?

Yes, we can change it.

I don't recall a particular reason for the `install_recommends` to be set to `no`, and I believe we should have it enabled.

letmaik:
@ionutbalutoiu Could you check and dump the Debian metadata for the packages we're talking about here? Then we'll have a better idea on whether `install_recommends` should be enabled, and my gut feeling is it should never be but rather any dependency should be listed explicitly.

jhand2:
> my gut feeling is it should never be but rather any dependency should be listed explicitly.

Why is that? Intel's deb packages call out their dependencies.

Enabling install_recommends seems simpler and more closely mimics how a normal user would install SGX libs without these scripts. Disabling it seems like arbitrary extra work.

IMO, OE should only declare dependencies on libraries it directly uses symbols from.

eddyashton:
I don't think this specific issue relates to `install_recommends`. If I'm understanding the sources correctly, the libsgx-urts dependency is defined [here](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/installer/linux/deb/libsgx-pce-logic/libsgx-pce-logic-1.0/debian/control). This is a `Depends`, not a `Recommends`, so its always installed transitively. The problem is that this dependency (in master and the 2.9 release) is on 2.8, but actually requires 2.9. So there _is_ a bug in Intel's stated dependencies, but we can work around this by installing the latest `libsgx-urts` explicitly. Happy to be convinced that's not the _right_ solution, if we can push a fix and in future trust the implicit transitive dependencies.

jhand2:
> I don't think this specific issue relates to `install_recommends`. If I'm understanding the sources correctly, the libsgx-urts dependency is defined [here](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/installer/linux/deb/libsgx-pce-logic/libsgx-pce-logic-1.0/debian/control). This is a `Depends`, not a `Recommends`, so its always installed transitively. The problem is that this dependency (in master and the DCAP 1.6 release) is on 2.8, but actually requires 2.9.
> 
> So there _is_ a bug in Intel's stated dependencies, but we can work around this by installing the latest `libsgx-urts` explicitly. Happy to be convinced that's not the _right_ solution, if we can push a fix and in future trust the implicit transitive dependencies.

Ah I see, thanks for the clarification. Could you add it as a dependency under `intel_dcap_packages` instead of `intel_sgx_packages`? I'm happy to merge this.

Would you like to open an issue on the intel DCAP github repo? Otherwise I can.

eddyashton:
> Could you add it as a dependency under `intel_dcap_packages` instead of `intel_sgx_packages`? I'm happy to merge this.

Will do.

> Would you like to open an issue on the intel DCAP github repo?

I've opened [this issue](https://github.com/openenclave/openenclave/pull/2953). I realised while writing it that the issue may not be with DCAP - perhaps DCAP1.6 can build against _either_ PSW 2.8, or PSW 2.9, but not the frankensituation where only `urts` is 2.8. In this case the missing dependency would be in `intel/linux-sgx`, but I'm less familiar with that repo.

CodeMonkeyLeet:
Correction for:
> I've opened [this issue](https://github.com/openenclave/openenclave/pull/2953).

The link should be to https://github.com/intel/SGXDataCenterAttestationPrimitives/issues/105, it currently links back to this PR.

CodeMonkeyLeet:
@letmaik Is there still a separate issue around the setting for `install_recommends` that should be tracked? If so, can you open an issue for it?

PR #2954, opened 2020-04-28
Remove switchless.edl depenedency on internal header

jhand2:
switchless.edl depended on an internal header. Move specific structs to a bits header to solve this issue.

### DIFF
--- a/edl/sgx/switchless.edl
+++ b/edl/sgx/switchless.edl
@@ -16,14 +16,47 @@ enclave
 {
     include "openenclave/bits/types.h"
 
+    struct oe_host_worker_context_t
+    {
+        void* call_arg;
+        oe_enclave_t* enc;
+        bool is_stopping;
+
+        int32_t event;
+
+        // Number of times the worker spun without seeing a message.
+        uint64_t spin_count;
+
+        // Statistics.
+        uint64_t total_spin_count;
+    };
+
+    struct oe_enclave_worker_context_t
+    {
+        void* call_arg;
+        oe_enclave_t* enc;
+        bool is_stopping;
+
+        int32_t event;
+
+        // Number of times the worker spun without seeing a message.
+        uint64_t spin_count;
+
+        // The limit at which to stop spinning and return to host to sleep.
+        uint64_t spin_count_threshold;
+
+        // Statistics.
+        uint64_t total_spin_count;
+    };
+
     trusted
     {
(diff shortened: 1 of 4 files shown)


jhand2:
bors try

include/openenclave/bits/edl/sgx/switchless.h
```diff
@@ -0,0 +1,45 @@
...
+#define _OE_EDL_SWITCHLESS_H
+
+#include <openenclave/bits/defs.h>
+#include <openenclave/bits/types.h>
+
+// This function cannot be defined in EDL because it uses volatile
+// members
+typedef struct _host_worker_context
+{
+    volatile void* call_arg;
```
mingweishih, line 14: Just wondering why we need to use `volatile` here. @anakrish?

anakrish: @mingweishih 
volatile is applicable only to prevent compiler optimizations that may break multi-threaded code, as happens here:
https://godbolt.org/z/VtE-fh

Our loops call un-inlined functions and therefore volatile is unnecessary. (We don't build with link time optimization).

It is safe to remove volatile in these structs. Instead, volatile could be added via pointers to these structs in non EDL code: `volatile oe_host_worker_context_t* context`.

I prefer adding volatile (in this case via pointers) because:
- It gives the reader an indication that a particular value is going to be used across multiple threds
- I makes the code robust, no matter how the code changes in the future

dthaler: Might be nice to add a code comment explaining this

anakrish: I thought about it a bit more. Though `volatile` will prevent compiler optimizations, it does not guarantee that the writes made to a variable in one thread (core) are ever visible to another thread (core). In theory, though one thread could set `stopping = true`, this updated value may never leave the current core's cache and reach the other thread that is spinning on the volatile variable. Acquire-release semantics are needed to ensure that the changes indeed are visible across multiple threads/cores.

I have created a task to verify that we are correctly using acquire-release semantics for switchless calls context. #2963

@jhand2 You can consider the volatiles here as non essential and apt for removal.
@mingweishih Thanks for pinging me and getting me to think about this again.

jhand2: Ok I moved the structs to switchless.edl. Note I had to change the `enclave` member to `enc` because enclave is reserved keyword in EDL.

PR #2527, opened 2020-02-09
Enable Minimal ARM64 CI Tests

HernanGatta:
## Overview

During CI runs, the SDK is built for OP-TEE on ARM TrustZone, including a number of tests in its test suite. However, no tests are executed even though some are known to pass on this platform.

This PR enables 17 passing tests to execute during CI runs for ARM64.

### Test Listing

```
Test project /openenclave/build
      Start  1: tests/mem
 1/17 Test  #1: tests/mem ........................   Passed    0.04 sec
      Start  2: tests/safecrt
 2/17 Test  #2: tests/safecrt ....................   Passed    0.21 sec
      Start  3: tests/safemath
 3/17 Test  #3: tests/safemath ...................   Passed    0.09 sec
      Start  4: tests/str
 4/17 Test  #4: tests/str ........................   Passed    0.04 sec
      Start  5: tests/logging
 5/17 Test  #5: tests/logging ....................   Passed    0.03 sec
      Start  6: tests/create-errors
 6/17 Test  #6: tests/create-errors ..............   Passed    0.04 sec
      Start  7: tests/crypto/enclave
 7/17 Test  #7: tests/crypto/enclave .............   Passed   14.08 sec
      Start  8: tests/crypto/host
 8/17 Test  #8: tests/crypto/host ................   Passed    0.47 sec
      Start  9: tests/hexdump
 9/17 Test  #9: tests/hexdump ....................   Passed    0.47 sec
      Start 10: tests/initializers
10/17 Test #10: tests/initializers ...............   Passed    0.20 sec
      Start 11: tests/mixed_c_cpp
11/17 Test #11: tests/mixed_c_cpp ................   Passed    0.17 sec
      Start 12: tests/pingpong
12/17 Test #12: tests/pingpong ...................   Passed    0.17 sec
      Start 13: tests/pingpong-shared
13/17 Test #13: tests/pingpong-shared ............   Passed    0.19 sec
      Start 14: tests/SampleApp
14/17 Test #14: tests/SampleApp ..................   Passed    0.24 sec
      Start 15: tests/SampleAppCRT
15/17 Test #15: tests/SampleAppCRT ...............   Passed    0.19 sec
      Start 16: tests/stdc
16/17 Test #16: tests/stdc .......................   Passed    3.29 sec
      Start 17: tests/libc
17/17 Test #17: tests/libc .......................   Passed    7.07 sec

100% tests passed, 0 tests failed out of 17

Total Test time (real) =  28.11 sec
```

### Dependencies

[...]

PR #2961, opened 2020-04-29
Port samples and tests to use manually imported system EDL

### LINKED ISSUE #2866: Make COMPILE_SYSTEM_EDL default to OFF for v0.10
Currently COMPILE_SYSTEM_EDL defaults to ON. This will build all generated stubs for OE system EDL files into OE libraries. The future direction is to not build these stubs into the SDK libraries and instead have applications include the ecalls/ocalls their application needs.

For v0.9 the default is still the old behavior but after that release we should alter the default behavior to the new method.

### LINKED ISSUE #2867: Update tests/samples to import system EDL manually
When COMPILE_SYSTEM_EDL is off by default, all the samples and tests should be switched over to import the system EDL they need. See #2866 and #2816 for more details.

jhand2:
This is a very large PR that does a few things. I have split up the PR description by replicating the commit message of each commit. Feel free to look at each commit individually to review each part in isolation.

Fixes #2866
Fixes #2867

## Canges from another PR

This PR also includes the commit from https://github.com/openenclave/openenclave/pull/2954. This commit will be removed from this PR once #2954 is merged.

## Changes in this PR

* Prevent conflicts on the host from the same EDL file being imported
> OE supports the EDL stubs for two different enclaves being linked into the same host binary. To prevent collisions, mark all host functions as weak symbols. Additionally, add include guards around generated argument structs.

* Port tests to import system EDL
> Switch COMPILE_SYSTEM_EDL to default to OFF and port each test to manually import the EDL files it requires.

* Fix inclusion of internal headers in bits headers
> The epid.h header which is included by the SGX attestation edl file includes an internal defs.h header. Move defs to bits/defs.h now that they are needed by EDL files.

* Port samples to manually import system edl
> Manully import EDL files which were previously built into core OE libraries.


include/openenclave/edger8r/host.h
```diff
@@ -106,6 +106,14 @@ OE_INLINE size_t oe_wcslen(const wchar_t* s)
     return wcslen(s);
 }
 
+#if defined(__GNUC__)
+#define OE_WEAK __attribute__((weak))
+#elif _MSC_VER
+#define OE_WEAK __declspec(selectany)
```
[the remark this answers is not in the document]
jhand2: Looks like you're right. I guess this won't work then.

I tried to prefix symbol names with the file name and then just do include guards around in inline function in the header like this

```
OE_INLINE int my_ecall() {
   return filename_prefixed_my_ecall();
}
```

But that doesn't work because when you compile the header for use in the OE core libraries, the symbol `filename_prefixed_my_ecall` will not be in the binary (it will have some other prefix with a different filename).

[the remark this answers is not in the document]
anakrish: Very challenging problem. I just wish `selectany` would work.
@mingweishih Any ideas?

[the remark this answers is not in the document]
mingweishih: It may work based on the [source](https://github.com/microsoft/cppwinrt/blob/4d8e0e26ad9c8728560bf62840bce4f5364d8779/strings/base_extern.h#L2). Maybe worth verifying?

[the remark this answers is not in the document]
mingweishih: Well, it seems a function pointer 😢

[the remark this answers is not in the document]
jhand2: We could prefix the functions and then declare a function pointer selectany. So we would generate the following for each ecall:

```
int filename_prefixed_my_ecall() {
   // do stuff
}

__declspec(selectany) int (*my_ecall)() = filename_prefixed_my_ecall;
```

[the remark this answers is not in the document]
anakrish: Sounds like a plan. With function pointers, do we still need prefixing with filename?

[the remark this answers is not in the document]
anakrish: On another note, I'm fine with this going in just for Linux and another PR to tackle the windows specific problem.

The only outstanding items seems to be "weak symbols for windows"; and I think that can be tackled in another PR, if it makes development easy.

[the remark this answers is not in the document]
jhand2: The oeedger8r test fails if this isn't addressed, so I think I'll have to address this to merge this PR

jhand2:
Running CI just to see where we're at. I did all development and testing on Linux.

bors try

tests/syscall/socket/host/CMakeLists.txt
```diff
@@ -6,7 +6,10 @@ set (EDL_FILE ../socket_test.edl)
 add_custom_command(
     OUTPUT socket_test_u.h socket_test_u.c
     DEPENDS ${EDL_FILE} edger8r
-    COMMAND edger8r --untrusted ${EDL_FILE} --search-path ${CMAKE_CURRENT_SOURCE_DIR}
+    COMMAND edger8r --untrusted ${EDL_FILE}
+               --search-path ${PROJECT_SOURCE_DIR}/edl
+               --search-path ${PROJECT_SOURCE_DIR}/edl/sgx
```
dthaler, line 11: Same question about tests... how does this sgx search path affect running the tests for optee?

samples/helloworld/enclave/CMakeLists.txt
```diff
@@ -4,7 +4,9 @@
 # Use the edger8r to generate C bindings from the EDL file.
 add_custom_command(OUTPUT helloworld_t.h helloworld_t.c helloworld_args.h
   DEPENDS ${CMAKE_SOURCE_DIR}/helloworld.edl
-  COMMAND openenclave::oeedger8r --trusted ${CMAKE_SOURCE_DIR}/helloworld.edl)
+  COMMAND openenclave::oeedger8r --trusted ${CMAKE_SOURCE_DIR}/helloworld.edl
+                                 --search-path ${OE_INCLUDEDIR}/openenclave/edl
+                                 --search-path ${OE_INCLUDEDIR}/openenclave/edl/sgx)
```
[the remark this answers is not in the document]
jhand2: Ah, this is what I was hoping for :smile: I couldn't figure out which samples are supposed to work for OP-TEE so I didn't add OP-TEE support for any of them. I can update this to work for both for the samples that should work on OP-TEE.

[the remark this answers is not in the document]
HernanGatta: I have not tried the samples on OP-TEE (though I should...). I would expect `helloworld` and `file-encryptor` to work, and the others not to since the infrastructure for them is not implemented on OP-TEE (attestation, sealing, and switchless).

The tests known to work are listed in [ARM CI PR](https://github.com/openenclave/openenclave/pull/2527) (still pending; I've some work to do on it). See the _Further Work_ section, too.

include/openenclave/bits/defs.h
```diff
@@ -182,4 +182,54 @@
 #error "OE_PACK_BEGIN and OE_PACK_END not implemented"
 #endif
 
+/*
+ * Inteded for use by other bits headers, not a part of the public API surface.
```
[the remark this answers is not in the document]
jhand2: Ya "not intended to be called by apps" is better

include/openenclave/edger8r/host.h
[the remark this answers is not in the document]
jhand2: Ok I have implemented this here: https://github.com/openenclave/openenclave/pull/2961/commits/3ed98e3cff51fc11d708c59f57fd1cc7ae118091

samples/helloworld/enclave/CMakeLists.txt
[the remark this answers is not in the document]
jhand2: From a code perspective I would expect them to work, but from a build/signing perspective, all the samples are signed with oesign using .conf files which is why I assumed they didn't work on OP-TEE

jhand2:
bors try

tests/syscall/socket/host/CMakeLists.txt
jhand2: Fixed this for tests. Still looking into how I can differentiate in the samples. If they don't already run on OP-TEE I would probably defer to a PR that fixes those problems first.

samples/attested_tls/client/host/Makefile
```diff
@@ -6,12 +6,15 @@ include ../../../config.mk
...
 LDFLAGS=$(shell pkg-config oehost-$(CXX_COMPILER) --libs)
+INCDIR=$(shell pkg-config oehost-$(C_COMPILER) --variable=includedir)
 
 all: build
 
 build:
 	@ echo "Compilers used: $(CC), $(CXX)"
-	oeedger8r ../tls_client.edl --untrusted
+	oeedger8r ../tls_client.edl --untrusted \
+								--search-path $(INCDIR)/openenclave/edl \
```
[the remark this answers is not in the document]
jhand2: Must have just been an indentation error on my part.

stylistically, should `--untrusted` and `--search-path` be aligned or is it better to just do one level of indentation for the hanging line?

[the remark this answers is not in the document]
jhand2: I went ahead and changed it to the latter

jhand2:
bors try

samples/file-encryptor/enclave/CMakeLists.txt
```diff
@@ -4,7 +4,9 @@
 # Use the edger8r to generate C bindings from the EDL file.
 add_custom_command(OUTPUT fileencryptor_t.h fileencryptor_t.c fileencryptor_args.h
   DEPENDS ${CMAKE_SOURCE_DIR}/fileencryptor.edl
-  COMMAND openenclave::oeedger8r --trusted ${CMAKE_SOURCE_DIR}/fileencryptor.edl)
+  COMMAND openenclave::oeedger8r --trusted ${CMAKE_SOURCE_DIR}/fileencryptor.edl
+                                 --search-path ${OE_INCLUDEDIR}
+                                 --search-path ${OE_INCLUDEDIR}/openenclave/edl/sgx)
```
dthaler, line 9: Hernan mentioned that he would expect the file-encryptor sample to work for OP-TEE, but line 9 seems to add SGX-specific stuff into the makefile without being inside an SGX condition.

samples/file-encryptor/enclave/Makefile
```diff
@@ -14,7 +15,9 @@ all:
 
 build:
 	@ echo "Compilers used: $(CC), $(CXX)"
-	oeedger8r ../fileencryptor.edl --trusted
+	oeedger8r ../fileencryptor.edl --trusted \
+		--search-path $(INCDIR) \
+		--search-path $(INCDIR)/openenclave/edl/sgx
```
dthaler, line 20: put SGX under some SGX-specific condition

samples/file-encryptor/host/CMakeLists.txt
```diff
@@ -3,7 +3,9 @@
 
 add_custom_command(OUTPUT fileencryptor_u.h fileencryptor_u.c fileencryptor_args.h
   DEPENDS ${CMAKE_SOURCE_DIR}/fileencryptor.edl
-  COMMAND openenclave::oeedger8r --untrusted ${CMAKE_SOURCE_DIR}/fileencryptor.edl)
+  COMMAND openenclave::oeedger8r --untrusted ${CMAKE_SOURCE_DIR}/fileencryptor.edl
+                                 --search-path ${OE_INCLUDEDIR}
+                                 --search-path ${OE_INCLUDEDIR}/openenclave/edl/sgx)
```
dthaler, line 8: put SGX under some SGX-specific condition

samples/file-encryptor/host/Makefile
```diff
@@ -6,12 +6,15 @@ include ../../config.mk
...
+INCDIR=$(shell pkg-config oehost-$(C_COMPILER) --variable=includedir)
 
 all: build
 
 build:
 	@ echo "Compilers used: $(CC), $(CXX)"
-	oeedger8r ../fileencryptor.edl --untrusted
+	oeedger8r ../fileencryptor.edl --untrusted \
+		--search-path $(INCDIR) \
+		--search-path $(INCDIR)/openenclave/edl/sgx
```
dthaler, line 17: put SGX under some SGX-specific condition

samples/helloworld/enclave/Makefile
```diff
@@ -24,7 +25,9 @@ all:
 
 build:
 	@ echo "Compilers used: $(CC), $(CXX)"
-	oeedger8r ../helloworld.edl --trusted
+	oeedger8r ../helloworld.edl --trusted \
+		--search-path $(INCDIR) \
+		--search-path $(INCDIR)/openenclave/edl/sgx
```
dthaler, line 30: put sgx under some sgx-specific condition

samples/helloworld/host/CMakeLists.txt
```diff
@@ -2,7 +2,9 @@
 # Licensed under the MIT License.
 add_custom_command(OUTPUT helloworld_u.h helloworld_u.c helloworld_args.h
   DEPENDS ${CMAKE_SOURCE_DIR}/helloworld.edl
-  COMMAND openenclave::oeedger8r --untrusted ${CMAKE_SOURCE_DIR}/helloworld.edl)
+  COMMAND openenclave::oeedger8r --untrusted ${CMAKE_SOURCE_DIR}/helloworld.edl
+                                 --search-path ${OE_INCLUDEDIR}
+                                 --search-path ${OE_INCLUDEDIR}/openenclave/edl/sgx)
```
dthaler, line 7: put sgx under some sgx-specific condition

samples/helloworld/host/Makefile
```diff
@@ -5,10 +5,13 @@ include ../../config.mk
...
 CFLAGS=$(shell pkg-config oehost-$(COMPILER) --cflags)
 LDFLAGS=$(shell pkg-config oehost-$(COMPILER) --libs)
+INCDIR=$(shell pkg-config oehost-$(COMPILER) --variable=includedir)
 
 build:
 	@ echo "Compilers used: $(CC), $(CXX)"
-	oeedger8r ../helloworld.edl --untrusted
+	oeedger8r ../helloworld.edl --untrusted \
+		--search-path $(INCDIR) \
+		--search-path $(INCDIR)/openenclave/edl/sgx
```
dthaler, line 14: put sgx under some sgx-specific condition

samples/file-encryptor/enclave/CMakeLists.txt
jhand2: I asked previously but didn't get an answer. Doesn't line 21 of `samples/file-encryptor/CMakeLists.txt` already mean the sample can't be built for OP-TEE?

If OP-TEE wasn't already supported, I think adding an OP-TEE build flag for these samples is out of scope for this PR

HernanGatta: Sorry. I see what you mean, you are right. I was thinking that in terms of the code it should work, but it's true that the build steps currently only generate properly signed SGX enclaves, and not OP-TEE TAs.

I had brought up the idea at one point to export the `add_enclave` function from the build system to consumers (i.e., when consuming the CMake exports from a binary release), so this signing logic wouldn't have to be copied around by users.

I do agree that solving that problem is out of scope for this PR, and so is duplicating the appropriate signature commands for OP-TEE; they are far less trivial than for SGX.

jhand2: Ok sounds good. is there an issue to track enabling samples on OP-TEE? If not I can open one.

samples/file-encryptor/enclave/Makefile
jhand2: Closing this and others based on discussion with @HernanGatta

jhand2:
bors try

samples/helloworld/enclave/CMakeLists.txt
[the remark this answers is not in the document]
jhand2: Closing based discussion in other thread.

jhand2:
bors try

samples/file-encryptor/enclave/CMakeLists.txt
HernanGatta: I don't there is one.

jhand2: Opened https://github.com/openenclave/openenclave/issues/2997

Thanks!

jhand2:
bors try

jhand2:
@mingweishih @anakrish Do you guys have anything else to add?

One thing to note is that I dropped the edger8r changes and just disabled the relevant edger8r test that checks to see if two EDL host stubs can be linked (see most recent commit).

mingweishih:
Minor suggestion: you may want to squash the first commit (`Remove switchless.edl depenedency on internal header`) into the second one.

jhand2:
> Minor suggestion: you may want to squash the first commit (`Remove switchless.edl depenedency on internal header`) into the second one.

Oh good catch, that commit is actually already in master, I can just drop it.

anakrish:
@mingweishih @jhand2 Do you have any tips to reviewing this PR? With 324 files, I fear I might end up not doing a thorough review. Should I review each commit instead?

jhand2:
> @mingweishih @jhand2 Do you have any tips to reviewing this PR? With 324 files, I fear I might end up not doing a thorough review. Should I review each commit instead?

Yes I would recommend reviewing each commit. I tried to do the bulk of the repetitive work in two commits (the samples commit and the tests commit). For those two I basically just did the same thing for ever test and sample.

The other commits should be much smaller.

samples/attested_tls/client/enc/CMakeLists.txt
```diff
@@ -5,8 +5,10 @@
 add_custom_command(
   OUTPUT tls_client_t.h tls_client_t.c tls_client_args.h
   DEPENDS ${CMAKE_SOURCE_DIR}/client/tls_client.edl
-  COMMAND openenclave::oeedger8r --trusted
-          ${CMAKE_SOURCE_DIR}/client/tls_client.edl)
+  COMMAND
+    openenclave::oeedger8r --trusted ${CMAKE_SOURCE_DIR}/client/tls_client.edl
+    --search-path ${OE_INCLUDEDIR} --search-path
```
anakrish, line 10: NTH: Can we avoid requiring the user to specify `${OE_INCLUDEDIR}/openenclave/edl/sgx`?
Intel SDK aseems to require specifying only `$(SGX_SDK)/include`:
https://github.com/intel/linux-sgx/blob/fdc9b33ee173a5e7743f20a2a9cb1dd8db114dca/SampleCode/Switchless/Makefile#L199
```make
	@cd App && $(SGX_EDGER8R) --untrusted ../Enclave/Enclave.edl --search-path ../Enclave --search-path $(SGX_SDK)/include
```

samples/local_attestation/localattestation.edl
```diff
@@ -2,6 +2,11 @@
 // Licensed under the MIT License.
 
 enclave {
+    from "openenclave/edl/attestation.edl" import *;
+    from "openenclave/edl/logging.edl" import *;
+    from "openenclave/edl/syscall.edl" import *;
+    from "platform.edl" import *;
```
anakrish, line 8: It reads as if `platform.edl` is a user-written EDL, since for other system edl's we require `openenclave/edl/` qualification.

anakrish:
I removed import statements from tests/echo/echo.edl to see what the user experience would be in case the user forgets to import a necessary edl file.
For most part, it was easy for me to look at the missing symbol (e.g: oe_syscall_nanosleep_ocall) and figure out that I must import the `syscall.edl` file.
For the functions defined in platform edl files, it was not clear.
For example, it is not clear that oe_sgx_backtrace_symbols_ocall came from edl/sgx/debug.edl.

From a usability standpoint, most users are going to encounter link errors when they upgrade to v0.10. Can we make the upgrade process easier in regards to how they can fix the link errors they they might see?

samples/local_attestation/localattestation.edl
jhand2: This is to allow someone to use the same edl file for both SGX and OP-TEE by just changing build files.

We could get rid of this by implementing https://github.com/openenclave/openenclave/issues/2833

Now that edger8r is C++, implementing this might be a bit easier.

samples/attested_tls/client/enc/CMakeLists.txt
jhand2: See below comment.

Intel SDK can do that because they only support SGX

jhand2:
> I removed import statements from tests/echo/echo.edl to see what the user experience would be in case the user forgets to import a necessary edl file.
> For most part, it was easy for me to look at the missing symbol (e.g: oe_syscall_nanosleep_ocall) and figure out that I must import the `syscall.edl` file.
> For the functions defined in platform edl files, it was not clear.
> For example, it is not clear that oe_sgx_backtrace_symbols_ocall came from edl/sgx/debug.edl.
> 
> From a usability standpoint, most users are going to encounter link errors when they upgrade to v0.10. Can we make the upgrade process easier in regards to how they can fix the link errors they they might see?

I can clearly document a guide on converting an app to the new model, but I don't know what changes I can make to add better error messages to the build since those are up to the linker.

jhand2:
bors try

jhand2:
bors try

anakrish:
The failure was due to e2e tests not able to access fedora packages.

PR #2964, opened 2020-04-29
Propose OE Enclave Runtime Support for OpenSSL

CodeMonkeyLeet:


docs/DesignDocs/enclave_openssl_support.md
```diff
@@ -0,0 +1,89 @@
...
+  - remote_attestation
+- Document the support for OpenSSL in OE SDK, i.e. add documentation analogous to [MbedtlsSupport.md](https://github.com/openenclave/openenclave/blob/master/docs/MbedtlsSupport.md)
+
+Also, all prototypes have technical gaps that will need to be addressed for production purposes. A non-exhaustive list of such tasks would include:
+
+- [ ] Add ctr_drbg.c to the `oecryptoopenssl` library implementation for parity with `oecryptombedtls`.
+- [ ] Update OpenSSL code changes to use `oe_get_entropy()` for entropy pool seeding, and avoid modifying the OpenSSL drbg code.
+- [ ] Formalize additional syscall stubs needed for OpenSSL as needed for [ssl_stubs.c](https://github.com/openenclave/openenclave-openssl/blob/master/0001-Get-openssl-to-build-against-MUSL-headers.-Use-SGX-r.patch).
+- [ ] Update the OpenSSL target branch to the latest [OpenSSL_1_1_1-stable](https://github.com/openssl/openssl/tree/OpenSSL_1_1_1-stable) and resolve test coverage differences or breaking changes.
+- [ ] Evaluate the set of [broken OpenSSL tests](https://github.com/openenclave/openenclave-openssl/blob/master/test/tests.broken) when executed in an enclave.
```
radhikaj, line 68: tests.broken was not complete. I think a bulk of work will be getting the openssl tests running inside enclaves.

docs/DesignDocs/enclave_openssl_support.md
```diff
@@ -0,0 +1,90 @@
+# OE Enclave Runtime Support for OpenSSL
+
+## Motivation
+
+Per discussion in issue [#852](https://github.com/openenclave/openenclave/issues/852):
+
+- There are users of the OE SDK for which the existing mbedTLS library it ships with is insufficient to meet their needs.
+  - It is not a pervasively used crypto library, which increases the costs of porting existing code into the enclave.
```
anakrish, line 8: One question that comes up is why open enclave defaults to a `non pervasively used crypto library`?

docs/DesignDocs/enclave_openssl_support.md
```diff
@@ -0,0 +1,90 @@
...
+
+The OE SDK package will be expanded to include:
+
+- A version of the OpenSSL libraries adapted for use in an enclave (the `openssl` component in the diagram).
+- An OE-specific wrapper implementation to allow `oeenclave` to use OpenSSL instead of mbedTLS (the `oecryptoopenssl` component in the diagram).
+
+An enclave app developer will have the choice at build time for their app to link either the `oecryptombed` or `oecryptoopenssl` wrappers against `oeenclave`, which will decide which crypto library `oeenclave` uses internally.
+
+A developer can link both `mbedcrypto` and `openssl` components into their enclave app at the same time, even if `oeenclave` can only use a single wrapper at a time. For example, an app developer may choose to retain the dependency `oeenclave` has on `oecryptombed` but also additionally link in `openssl`. Doing so would allow the enclave app to use some functionality available in OpenSSL but not mbedTLS, even though `oeenclave` continues to use mbedTLS. This would not be a recommended use case because of the increase in the Trusted Computing Base (TCB) of the enclave app, but it is allowed.
```
anakrish, line 30: Today there are the following mbedtls libraries:
- libmbedcrypto.a
- libmbedtls.a
- liboecryptombed.a
- libmbedx509.a

Would linking against liboecryptoopenssl, still give users attested-tls?

docs/DesignDocs/enclave_openssl_support.md
```diff
@@ -0,0 +1,90 @@
...
+
+Also, all prototypes have technical gaps that will need to be addressed for production purposes. A non-exhaustive list of such tasks would include:
+
+- [ ] Add ctr_drbg.c to the `oecryptoopenssl` library implementation for parity with `oecryptombedtls`.
+- [ ] Update OpenSSL code changes to use `oe_get_entropy()` for entropy pool seeding, and avoid modifying the OpenSSL drbg code.
+- [ ] Formalize additional syscall stubs needed for OpenSSL as needed for [ssl_stubs.c](https://github.com/openenclave/openenclave-openssl/blob/master/0001-Get-openssl-to-build-against-MUSL-headers.-Use-SGX-r.patch).
+- [ ] Update the OpenSSL target branch to the latest [OpenSSL_1_1_1-stable](https://github.com/openssl/openssl/tree/OpenSSL_1_1_1-stable) and resolve test coverage differences or breaking changes.
+- [ ] Evaluate the set of [broken OpenSSL tests](https://github.com/openenclave/openenclave-openssl/blob/master/test/tests.broken) when executed in an enclave.
+
+## Alternates
```
dthaler, line 71:
```suggestion
## Alternatives
```

docs/DesignDocs/enclave_openssl_support.md
```diff
@@ -0,0 +1,90 @@
...
+An additional refinement of that approach is to eliminate the internal OE crypto API layer entirely:
+
+- OE SDK calls OpenSSL APIs directly instead of using its `oecrypto` indirection.
+- `oecryptoopenssl` becomes unnecessary, and `oecryptombedtls` would need to be refactored to be an mbedTLS to OpenSSL API adaptor instead.
+- Other crypto libraries that are already compatible or have existing shims with the OpenSSL API can be reused with minimal change.
+
+This option is not being pursued currently because:
+
+- As with the option to expose the existing `oecrypto` API for adaptation, it is additive to the proposed solution and can be incrementally pursued at a later date.
+  - Since it doesn't formalize the `oecrypto` API for crypto libraries to adapt against publicly at this time, the current proposal also avoids introducing a breaking change for enclave crypto libraries in the future.
```
dthaler, line 84: If we want to break them eventually, it's almost always better to break sooner, when there's less stuff to break.
However, another approach to avoid breaking, is to provide an optional shim of "deprecated" oecrypto APIs over the actual ones, but that just depends on how long we want to support the "deprecated" APIs.
I'd be in favor of saying the north star is to deprecate the oecrypto API, even if it's not the first step.

docs/DesignDocs/enclave_openssl_support.md
```diff
@@ -0,0 +1,90 @@
...
+  - The deeper refactor to use OpenSSL directly in OE is effectively an optimization at that point by removing a layer of indirection, but at the cost of a full rewrite of the `oecryptombedtls` wrapper to continue supporting it. The relative value of doing this work is currently lower than just allowing OpenSSL to be used by the OE runtime; it already allows other OpenSSL-compatible libraries to be ported more easily into OE enclaves, while not incurring the immediate `oecryptombedtls` refactor costs.
+- OpenSSL is also currently in the process of its transition to its next major release (3.0), and it makes sense to target that version intercept as the more broadly stable crypto API surface to support.
```
dthaler, line 86: I would agree with this statement IF we have evidence that other libraries (wolfssl, libressl, etc.) will target it as well.   If the current API stays defacto standard and openssl diverges and no one follows, then that would be a bad move to pick the one that's not the defacto standard.   Today, I only have evidence that the current openssl api is a de facto standard.

docs/DesignDocs/enclave_openssl_support.md
CodeMonkeyLeet: OE SDK used the mbedTLS library early on because:
- it has a smaller TCB than OpenSSL designed specifically for embedded systems with limited memory (as is the case with early SGX1 enclaves). It is designed to be statically linked and highly customizable to keep the TCB small by not including unused code, and had explicit configuration controls to limit its system dependencies on filesystem, sockets, memory allocation etc.
    - This is particularly relevant because OE only used the crypto library early with virtually no external dependencies by excluding the TLS library, whereas the equivalents in OpenSSL have a much larger dependency tree to port into the enclave runtime environment. Its use predates the introduction of the oesyscall compatibility layer, which the OpenSSL prototype support depends on heavily.
    - These are still relevant properties on the IoT side of the equation with the OP-TEE support, which is why the plan is not to deprecate support for mbedTLS. 
- at the point it was introduced, it also had a shorter and less frequent history of vulnerability disclosures.

docs/DesignDocs/enclave_openssl_support.md
CodeMonkeyLeet: The equivalent set of libraries for OpenSSL would be:
- liboecryptoopenssl.a
- libopensslcrypto.a (i.e. OpenSSL's libcrypto)
- libopenssltls.a (i.e OpenSSL's libssl)
(see next paragraph)

The attested TLS functionality internally uses the oecrypto abstractions and substituting the OpenSSL equivalents is required to support it.

docs/DesignDocs/enclave_openssl_support.md
CodeMonkeyLeet: The point is that the proposed design sidesteps the question of deprecating or breaking APIs entirely, by avoiding making the oecrypto API public to begin with. As it remains an internal API, whether it is eventually removed or not does not involve a public deprecation strategy.

docs/DesignDocs/enclave_openssl_support.md
CodeMonkeyLeet: That's already kind of untrue because there are already a range of incompatible OpenSSL APIs: the move from 1.0 to 1.1 was explicitly done by OpenSSL to introduce a set of breaking API changes, and the jump from 1.1 to 3.0 is also to adopt a versioning format more aligned with what folks understand with semantic versioning instead of OpenSSL's traditional use where minor versions are API breaking. 

In general, it's possible that other distributions may choose not to follow the changes that OpenSSL makes and such choices are likely to be idiosyncratic. Beyond the API question, OE would need to move to 3.0 itself at some point, and if the argument is that it needs to support as a adapter interface it does not directly use (i.e. 1.1) then that is actually a counterargument to removing the oecrypto abstraction. (i.e. it will need to keep the current OpenSSL 1.1 to oecrypto shim for purely API reasons, but it will be using a different OpenSSL 3.0 to oecrypto shim for its actual use of the latest patched OpenSSL version.)

PR #2965, opened 2020-04-30
add switchless worker threads sleep and resume test

yanxue-intel:

### DIFF
--- a/tests/switchless_worksleep/enc/enc.cpp
+++ b/tests/switchless_worksleep/enc/enc.cpp
@@ -1,0 +1,25 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/corelibc/string.h>
+#include <openenclave/enclave.h>
+#include <openenclave/internal/print.h>
+#include "switchless_worksleep_t.h"
+
+void enc_ecall1_switchless(void)
+{
+    host_ocall1_switchless();
+}
+
+void enc_ecall2_switchless(void)
+{
+    host_ocall2_switchless();
+}
+
+OE_SET_ENCLAVE_SGX(
+    1,        /* ProductID */
+    1,        /* SecurityVersion */
+    true,     /* AllowDebug */
+    64,       /* HeapPageCount */
+    64,       /* StackPageCount */
+    NUM_TCS); /* TCSCount */
--- a/tests/switchless_worksleep/host/host.cpp
+++ b/tests/switchless_worksleep/host/host.cpp
@@ -1,0 +1,183 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <inttypes.h>
+#include <limits.h>
+#include <openenclave/host.h>
+#include <openenclave/internal/error.h>
+#include <openenclave/internal/tests.h>
+#include <openenclave/internal/thread.h>
+#include <chrono>
+#include <cstdio>
+#include <cstdlib>
+#include <thread>
+#include <vector>
+#include "../../../host/hostthread.h"
+#include "../../../host/strings.h"
+#include "switchless_worksleep_u.h"
+using namespace std;
+
+// global counters to increment in ocalls
+uint32_t ocall1_counter = 0;
+uint32_t ocall2_counter = 0;
+
+/**
+ * ocall1 - increment counter and return
+ */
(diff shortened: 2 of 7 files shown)


anakrish:
bors try

tests/switchless_worksleep/host/host.cpp
```diff
@@ -0,0 +1,214 @@
...
+#include <Windows.h>
+#include <intrin.h>
+
+uint32_t get_nprocs(void)
+{
+    SYSTEM_INFO system_info;
+    ZeroMemory(&system_info, sizeof(system_info));
+
+    GetSystemInfo(&system_info);
+    return static_cast<size_t>(system_info.dwNumberOfProcessors);
```
jhand2, line 35: You can get this in a cross platform way in C++: `std::thread::hardware_concurrency();`

tests/switchless_worksleep/host/host.cpp
```diff
@@ -0,0 +1,214 @@
...
+    for (i = 0; i < app_threads.size(); i++)
+    {
+        oe_thread_join(app_threads[i]);
+    }
+
+    // remove threads from vector
+    app_threads.clear();
+
+    // sleep 2 seconds
+    this_thread::sleep_for(std::chrono::milliseconds(2000));
```
jhand2, line 188: I don't follow. What is the sleep for? A 2 second sleep in a test is very long when we run hundreds of tests.

tests/switchless_worksleep/host/host.cpp
```diff
@@ -0,0 +1,214 @@
...
+        ret = enc_ecall2_switchless(enclave);
+        if (ret != OE_OK)
+        {
+            fails++;
+        }
+    }
+    OE_TEST(fails == 0);
+
+    // sleep 1 second
+    this_thread::sleep_for(std::chrono::milliseconds(1000));
```
jhand2, line 97: Why 1 second? Is the length significant? I also don't really understand why sleeping the host thread would affect the enclave thread.

tests/switchless_worksleep/enc/enc.cpp
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT License.
+
+#include <openenclave/corelibc/string.h>
+#include <openenclave/enclave.h>
+#include <openenclave/internal/print.h>
+#include "switchless_worksleep_t.h"
+
+void enc_ecall1_switchless(void)
+{
+    host_ocall1_switchless();
```
jhand2, line 11: What is the reason for doing an ocall here? We know the host thread is awake here because it just make the ecall. I don't understand what is being tested.

tests/switchless_worksleep/host/host.cpp
yanxue-intel: When enclave worker thread is out of work it should enter sleep mode.The sleep function is used to stop posting tasks(switchless call) and make enclave worker thread enter sleep mode.
The following code comes from  switchless implementation.
           
           // If there is no message, increment spin count until threshold is
           // reached.

            if (++context->spin_count >= spin_count_threshold)
            {
                // Reset spin count and return to host to sleep.
                context->total_spin_count += context->spin_count;
                context->spin_count = 0;
                break;
            }
            // In Release builds, the following pause has been observed to be
            // essential. Without it, the worker thread seems to hog the CPU,
            // preventing host threads from posting switchless ecall messages.
            asm volatile("pause");

tests/switchless_worksleep/host/host.cpp
yanxue-intel: Just for debug purposes ,it can be safely removed. Thanks for your reminder.

tests/switchless_worksleep/enc/enc.cpp
yanxue-intel: The purpose of doing ocall here is to want the tasks to include the switchless ocall and the switchless ocall.

tests/switchless_worksleep/host/host.cpp
yanxue-intel: Great! Thanks

jinghe-INTC:
bors try

tests/switchless_worksleep/host/host.cpp
jhand2: So the goal is just to release the CPU to another thread? If so I would prefer pause over sleep

PR #2967, opened 2020-04-30
oecore internal changes required to support pluggable allocators

anakrish:
This PR contains only internal changes. No new functionality is exposed
to the user. All existing behavior in regards to allocation are retained
including experimental snmalloc support; while conforming to pluggable
allocators API. dlmalloc nolonger uses any oecore internal functions and
is fully adapted via the pluggable allocator interface.
snmalloc's PAL uses __oe_get_heap_base and __oe_get_heap_end functions;
that needs to be fixed in the snmalloc repo.

- internal/allocator.h that defines the pluggable allocator API.
  To be documented and publieshed in next iterations.
- dlmalloc and snmalloc compiled as pluggable allocator object libraries
  By default, oedlmalloc objects are included in oecore
- Retains USE_SNMALLOC experimental feature. In that case, oesnmalloc
  objects are included in oecore.
- Move oe_sbrk to dlmalloc allocator. sbrk is not general enough, and
  allocators like snmalloc prefer managing the heap directly rather
  than via sbrk.
- Remove unused dlmalloc stats functions. They were unused and untested.
- Removed errno setting from oecore for allocator functions.
  It is the responsibility of the allocators to be standards conformant
  and set errno appropriatly.

Tests:
- All existing tests cover dlmalloc plugged in as default allocator.
- Experimental snmalloc support remains untested.


anakrish:
bors try

mjp41:
@anakrish please raise either an issue or PR on snmalloc for

> snmalloc's PAL uses __oe_get_heap_base and __oe_get_heap_end functions;
that needs to be fixed in the snmalloc repo.

So we can get that fixed for you.

3rdparty/snmalloc/allocator.cpp
```diff
@@ -0,0 +1,92 @@
...
+    allocator_local = current_alloc_pool()->acquire();
+}
+
+void oe_allocator_thread_cleanup(void)
+{
+    current_alloc_pool()->release(ThreadAlloc::get());
+    allocator_local = nullptr;
+}
+
+void* oe_allocator_malloc(size_t size)
```
achamayou, line 59: Do we need these wrappers? We can just SNMALLOC_NAME_MANGLE with "oe_allocator_" and get them directly.

anakrish: Great suggestion! Done.

anakrish:
> @anakrish please raise either an issue or PR on snmalloc for
> 
> > snmalloc's PAL uses __oe_get_heap_base and __oe_get_heap_end functions;
> > that needs to be fixed in the snmalloc repo.
> 
> So we can get that fixed for you.

Sure. I was planning to create issues in snmalloc repo once this goes in so that you will have a version of OE to try out.

anakrish:
bors try

3rdparty/dlmalloc/allocator.c
```diff
@@ -0,0 +1,117 @@
...
+#define MORECORE dlmalloc_sbrk
+#define ABORT oe_abort()
+#define USE_DL_PREFIX
+#define LACKS_STDLIB_H
+#define LACKS_STRING_H
+#define USE_LOCKS 1
+#define fprintf _dlmalloc_stats_fprintf
+#define NO_MALLOC_STATS 1
+
+#pragma GCC diagnostic push
```
jhand2, line 20: Could you add a comment here for why this is needed?

3rdparty/dlmalloc/allocator.c
```diff
@@ -0,0 +1,117 @@
...
+#ifdef __clang__
+#pragma GCC diagnostic ignored "-Wparentheses-equality"
+#endif
+
+OE_INLINE
+int sched_yield(void)
+{
+#ifdef __x86__
+    asm volatile("pause");
```
jhand2, line 29: On non-x86 that means this function is broken right? Do OP-TEE enclaves use this allocator as well? The `yield` instruction on arm should do the same thing.

enclave/core/debugmalloc.c
```diff
@@ -399,34 +398,41 @@ void* oe_debug_realloc(void* ptr, size_t size)
...
 
-    header = (header_t*)((uint8_t*)block + padding_size);
-
-    INIT_BLOCK(header, alignment, size);
-    _check_block(header);
-    _list_insert(&_list, header);
+    // The only difference between posix_memalign and the obsolete memalign is
+    // that posix_memalign requires alignment to be a multiple of sizeof(void*).
+    // Adjust the alignment if needed.
+    if (alignment < sizeof(void*))
```
jhand2, line 406: Shouldn't this be `if (alignment % sizeof(void*) != 0) / * round up to aligned */`?

enclave/core/malloc.c
```diff
@@ -132,147 +50,45 @@ void* oe_calloc(size_t nmemb, size_t size)
...
-        errno = ENOMEM;
+    void* ptr = NULL;
 
-        if (_failure_callback)
-            _failure_callback(__FILE__, __LINE__, __FUNCTION__, size);
-    }
+    // The only difference between posix_memalign and the obsolete memalign is
+    // that posix_memalign requires alignment to be a multiple of sizeof(void*).
+    // Adjust the alignment if needed.
+    if (alignment < sizeof(void*))
```
jhand2, line 71: Same comment as above

anakrish:
bors try

3rdparty/dlmalloc/allocator.c
anakrish: This may not be needed. Good catch. Will try to remove it and see what happens.

3rdparty/dlmalloc/allocator.c
anakrish: This is the current behavior. When dlmalloc is compiled for OPTEE, sched yield is empty:
https://github.com/openenclave/openenclave/blob/71edd2cf92f52017410d5e3262628b962067560a/enclave/core/optee/sched_yield.c#L4-L7

@HernanGatta Do you know what the equivalent of `pause` assembly instruction is for OPTEE/ARM?

anakrish:
bors try

3rdparty/dlmalloc/allocator.c
anakrish: Removed diagnostic push

enclave/core/debugmalloc.c
anakrish: Made it more readable using round to multiple

enclave/core/malloc.c
anakrish: Made it more readable using round to multiple.

anakrish:
bors try

3rdparty/snmalloc/allocator.cpp
achamayou: Thank you!

3rdparty/dlmalloc/allocator.c
anakrish: Since OPTEE is currently `single-threaded` and since no-op is the current behavior, I'm resolving this for now.

jhand2:
btw, does removing sbrk.c cause problems? Doesn't the OE allocator manage the heap using oe_sbrk?

anakrish:
> btw, does removing sbrk.c cause problems? Doesn't the OE allocator manage the heap using oe_sbrk? In particular, SGX-LKL uses changes in oe_sbrk to partition the OE heap. Where would that logic be moved to?
> 
> See [f7eb8c2](https://github.com/openenclave/openenclave/commit/f7eb8c2189056324d65ffb4440d7a8addd7ad7ea)

 `oe_allocator_init` passes the heap range to an allocator (could be dlmalloc or snmalloc or something else). Instead of passing the whole heap to the allocator, only the part of partitioned heap needs to be passed to the allocator via `oe_allocator_init`. This will allow the partitioning to work not only for dlmalloc which uses sbrk, but for any other allocator OE core may use (snmalloc, tcmalloc etc) that may not use sbrk.

PR #2969, opened 2020-04-30
Pre-allocating memory for quote verification collaterals before an ocall

shruti25ratnam:
Issue: Update oe_get_sgx_quote_verification_collateral to preallocate buffers to avoid double calls. #2049
Fix: 1. Added a list of pre-estimated memory sizes for each arguments passed into the oe_get_quote_verification_collateral_ocall().
2. Allocate the above estimated memory-buffers to each of the arguments before passing them on to the ocall.

### DIFF
--- a/common/sgx/collateral.h
+++ b/common/sgx/collateral.h
@@ -66,6 +66,16 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
 void oe_free_sgx_quote_verification_collateral_args(
     oe_get_sgx_quote_verification_collateral_args_t* args);
 
+/**
+ * Pre-allocate memory for the resources that get passed into the 
+ * oe_get_quote_verification_collateral_ocall. To avoid extra passes needed
+ * to take care of OE_BUFFER_TOO_SMALL failures.
+ *
+ * @param[in] args The quote verification collateral.
+ */
+void oe_prealloc_quote_verification_collateral_args(
+    oe_get_sgx_quote_verification_collateral_args_t* buf);
+
 OE_EXTERNC_END
 
 #endif // _OE_COMMON_REVOCATION_H
--- a/common/sgx/collateral.h
+++ b/common/sgx/collateral.h
@@ -66,16 +66,6 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
 void oe_free_sgx_quote_verification_collateral_args(
     oe_get_sgx_quote_verification_collateral_args_t* args);
 
-/**
- * Pre-allocate memory for the resources that get passed into the 
- * oe_get_quote_verification_collateral_ocall. To avoid extra passes needed
- * to take care of OE_BUFFER_TOO_SMALL failures.
- *
- * @param[in] args The quote verification collateral.
- */
-void oe_prealloc_quote_verification_collateral_args(
-    oe_get_sgx_quote_verification_collateral_args_t* buf);
-
 OE_EXTERNC_END
 
 #endif // _OE_COMMON_REVOCATION_H
--- a/enclave/sgx/collateralinfo.c
+++ b/enclave/sgx/collateralinfo.c
@@ -13,6 +13,40 @@
 #include "platform_t.h"
 
 /**
+ * Update these default size values as needed.
+ * These represent the default buffer sizes that can store their
+ * corresponding quote_verification_collateral_args completely.
+ * Last updated as per arg-sizes observered as of May 2020.
+ */
+#define TCBINFO_DEFAULT_SIZE 5000
+#define PCK_CRL_DEFAULT_SIZE 600
+#define QE_IDENTITY_DEFAULT_SIZE 1500
+#define ROOT_CA_CRL_DEFAULT_SIZE 600
+#define ALL_ISSUER_CHAIN_DEFAULT_SIZE 3000
+
+/**
+ * Pre-allocate memory for the resources that get passed into the
+ * oe_get_quote_verification_collateral_ocall. To avoid extra passes
+ * needed to take care of OE_BUFFER_TOO_SMALL failures.
+ *
+ * @param[in] buf The quote verification collateral.
+ * [in] def_sizes The default quote verification collateral sizes.
+ */
+void oe_prealloc_quote_verification_collateral_args(
+    oe_get_sgx_quote_verification_collateral_args_t* buf,
+    oe_get_sgx_quote_verification_collateral_args_t* default_sizes);
+
+/**
+ * This function is called to update the default collateral arg sizes.
+ *
+ * @param[in] default_sizes The default quote verification collateral sizes.
+ * [in] src_args The quote verification collateral.
+ */
+void oe_update_default_collateral_arg_sizes(
+    oe_get_sgx_quote_verification_collateral_args_t* src_args,
+    oe_get_sgx_quote_verification_collateral_args_t* default_sizes);
+
+/**
  * Call into host to fetch collateral information.
  */
 oe_result_t oe_get_sgx_quote_verification_collateral(
@@ -21,6 +55,29 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
     oe_result_t result = OE_FAILURE;
     oe_get_sgx_quote_verification_collateral_args_t in = {0};
     oe_get_sgx_quote_verification_collateral_args_t out = {0};
+
+    /**
+     * This variable is used to store default collateral arg sizes.
+     */
+    static oe_get_sgx_quote_verification_collateral_args_t default_arg_size = {
+        0,
+        {0},
+        0,
+        TCBINFO_DEFAULT_SIZE,
+        0,
+        ALL_ISSUER_CHAIN_DEFAULT_SIZE,
+        0,
+        PCK_CRL_DEFAULT_SIZE,
+        0,
+        ALL_ISSUER_CHAIN_DEFAULT_SIZE,
+        0,
+        ROOT_CA_CRL_DEFAULT_SIZE,
+        0,
+        QE_IDENTITY_DEFAULT_SIZE,
+        0,
+        ALL_ISSUER_CHAIN_DEFAULT_SIZE,
+        0};
+
     uint32_t retval;
 
     if (!args)
@@ -28,11 +85,11 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
 
     /* fmspc */
     memcpy(in.fmspc, args->fmspc, sizeof(in.fmspc));
+    oe_prealloc_quote_verification_collateral_args(&in, &default_arg_size);
 
     for (;;)
     {
         memcpy(&out, &in, sizeof(out));
-        oe_prealloc_quote_verification_collateral_args(&out);
 
         if (oe_get_quote_verification_collateral_ocall(
                 &retval,
@@ -154,6 +211,7 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
 
     OE_CHECK((oe_result_t)retval);
 
+    oe_update_default_collateral_arg_sizes(&out, &default_arg_size);
     *args = out;
     memset(&out, 0, sizeof(out));
     result = OE_OK;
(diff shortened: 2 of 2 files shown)


common/sgx/collateral.h
```diff
@@ -66,6 +66,16 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
...
     oe_get_sgx_quote_verification_collateral_args_t* args);
 
+/**
+ * Pre-allocate memory for the resources that get passed into the 
+ * oe_get_quote_verification_collateral_ocall. To avoid extra passes needed
+ * to take care of OE_BUFFER_TOO_SMALL failures.
+ *
+ * @param[in] args The quote verification collateral.
+ */
+void oe_prealloc_quote_verification_collateral_args(
```
jhand2, line 76: You can just forward declare this in in the .c file, it doesn't need to be in the header

enclave/sgx/collateralinfo.c
```diff
@@ -169,6 +170,123 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
...
+oe_get_sgx_quote_verification_collateral_args_t* buf)
+{
+    /*These numbers are just estimates of buffer sizes*/
+    size_t prealloc_tcb_info_size = 5000;
+    size_t prealloc_tcb_info_issuer_chain_size = 3000;
+    size_t prealloc_pck_crl_size = 600;
+    size_t prealloc_root_ca_crl_size = 600;
+    size_t prealloc_pck_crl_issuer_chain_size = 3000;
+    size_t prealloc_qe_identity_issuer_chain_size = 3000;
+    size_t prealloc_qe_identity_size = 700;
```
jhand2, line 183: Can you `#define` these magic numbers at the top of the file? For example `TCBINFO_DEFAULT_SIZE` and the leave a detailed comment of why we're hard-coding these values.

enclave/sgx/collateralinfo.c
```diff
@@ -32,6 +32,7 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
     for (;;)
     {
         memcpy(&out, &in, sizeof(out));
+        oe_prealloc_quote_verification_collateral_args(&out);
```
jhand2, line 35: Just call prealloc once before entering the loop

enclave/sgx/collateralinfo.c
```diff
@@ -169,6 +170,123 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
...
+{
+    /*These numbers are just estimates of buffer sizes*/
+    size_t prealloc_tcb_info_size = 5000;
+    size_t prealloc_tcb_info_issuer_chain_size = 3000;
+    size_t prealloc_pck_crl_size = 600;
+    size_t prealloc_root_ca_crl_size = 600;
+    size_t prealloc_pck_crl_issuer_chain_size = 3000;
+    size_t prealloc_qe_identity_issuer_chain_size = 3000;
+    size_t prealloc_qe_identity_size = 700;
+    if(!buf->tcb_info_size)
```
jhand2, line 184: Given the suggestion at the callsite, I think we should just blindly pre-allocate the buffers without checking the size of each field. We're always going to pre-allocate once before the loop.

enclave/sgx/collateralinfo.c
```diff
@@ -169,6 +170,123 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
...
+    {
+        buf->tcb_info = (uint8_t *)malloc(prealloc_tcb_info_size);
+        if(buf->tcb_info != NULL)
+        {
+            buf->tcb_info_size = prealloc_tcb_info_size;
+            memset(buf->tcb_info,0, buf->tcb_info_size);
```
jhand2, line 190: Instead of malloc + memset, [use calloc](http://www.cplusplus.com/reference/cstdlib/calloc/)

enclave/sgx/collateralinfo.c
```diff
@@ -169,6 +170,123 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
     return result;
 }
 
+void oe_prealloc_quote_verification_collateral_args(
+oe_get_sgx_quote_verification_collateral_args_t* buf)
```
jhand2, line 174: nit: indent

enclave/sgx/collateralinfo.c
```diff
@@ -169,6 +170,123 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
...
+            buf->qe_identity_issuer_chain_size = prealloc_qe_identity_issuer_chain_size;
+            memset(buf->qe_identity_issuer_chain,0, buf->qe_identity_issuer_chain_size);
+        }
+        else
+        {
+            goto cleanup;
+        }
+    }
+    return;
+cleanup:
```
jhand2, line 284: Use `done:` label for consistency. OE has some macros that call `goto done;`

enclave/sgx/collateralinfo.c
```diff
@@ -169,6 +170,123 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
...
+        {
+            buf->qe_identity_issuer_chain_size = prealloc_qe_identity_issuer_chain_size;
+            memset(buf->qe_identity_issuer_chain,0, buf->qe_identity_issuer_chain_size);
+        }
+        else
+        {
+            goto cleanup;
+        }
+    }
+    return;
```
jhand2, line 283: Avoid early return when using labels/gotos. goto is generally scary so when it is being used it should be the only point of exit from the function.

`free` checks for null pointers anyway so it's safe to call free on a null pointer, this is ensured by POSIX.

enclave/sgx/collateralinfo.c
yentsanglee, line 183: If these values are too big, we waste memory; if too small, we double call. According to current PCK cache, actual sizes are 2125, 1893, 460, 448, 1909, 1893, and 498. Your pre-set values are reasonable. However, the sizes differ by time and hardware. It's hard to get them right. I suggest we at least make the size values static and remember the retrieved sizes. It guarantees we only double call once per execution if our initial guessed sizes are incorrect.

enclave/sgx/collateralinfo.c
shruti25ratnam: The memcpy in line 34 would probably erase the pre-allocation if I called it outside the for loop. The 'in' buffer has all its args except fmspc, pointing to NULL. I can add a condition to execute this function only once, if we think that can save us some runtime.

jhand2: I'm not too worried about the runtime overhead, I was just thinking it could simplify the code in `oe_prealloc_quote_verification_collateral_args`. This isnt too big of a deal though.

enclave/sgx/collateralinfo.c
shruti25ratnam: Yes, I can do that. My intention was to scope them locally so that they get cleaned up after the function ends. But I do agree they may need to be defined static on the top with some explanation.

common/sgx/collateral.h
shruti25ratnam: I agree, I'll move it to the .c file. 

My initial intent in adding it to the header was:
1. to maintain consistency with other functions defined here like:  oe_free_sgx_quote_verification_collateral_args() and oe_get_sgx_quote_verification_collateral()
2. If this function ever needs to be repurposed. Which currently doesn't look like the case.

enclave/sgx/collateralinfo.c
shruti25ratnam: Sure, I agree it's a great suggestion.

enclave/sgx/collateralinfo.c
yentsanglee: It's a valid concern. If the initial buffer is too small, line 34 will set the expected buffer sizes to 'out' variable. Calling line 35 again after the first iteration will wipe out the purpose of line 34. This may lead to infinite looping.

---
In reply to: [418278362](https://github.com/openenclave/openenclave/pull/2969#discussion_r418278362) [](ancestors = 418278362)

jhand2: Oh @shruti25ratnam now that I think about it, you could just call `oe_prealloc_quote_verification_collateral_args` on `&in` before the loop. Then the pointers will get copied to out

enclave/sgx/collateralinfo.c
shruti25ratnam: @jhand2 what do you think about making these variables static inside the function? This is in line with yentsanglee 's suggestion below on line 180. Do you still recommend I #define them? 
I can do both static inside the function and #define them in this file, if needed. Just thought it may be too much of an overhead.

enclave/sgx/collateralinfo.c
shruti25ratnam: Please correct me if I'm wrong, per convention I thought 'in' buffer represents the values coming in from the ocall and 'out' represented variables passed out to the ocall. I'll go per your recommendation if this is not the case.

shruti25ratnam: > It's a valid concern. If the initial buffer is too small, line 34 will set the expected buffer sizes to 'out' variable. Calling line 35 again after the first iteration will wipe out the purpose of line 34. This may lead to infinite looping.
> 
> In reply to: [418278362](https://github.com/openenclave/openenclave/pull/2969#discussion_r418278362) [](ancestors = 418278362)

In the function in line 35, I only pre-allocate if the sizes of the collateral args are zero, which is usually only in the first iteration. If sizes are non-zero this function just passes through. This saves it from infinite looping. 

Although, I'm currently adding some changes to use the second iteration to replace the estimated sizes with the actual ones and making them static. So that the eventual calls to this function could use the actual values for pre-allocation.

enclave/sgx/collateralinfo.c
jhand2: I prefer to always `#define` magic numbers to clearly call them out, explain their purpose, and be able to easily change them later. But because they're only used once I don' have a strong preference either way.

enclave/sgx/collateralinfo.c
jhand2: > per convention I thought 'in' buffer represents the values coming in from the ocall and 'out' represented variables passed out to the ocall

You're basically using the values returned from the ocall as the "input" the next ocall. It just so happens that you already have some input values you want to try first. I would say it's fine.

> In the function in line 35, I only pre-allocate if the sizes of the collateral args are zero, which is usually only in the first iteration. If sizes are non-zero this function just passes through. This saves it from infinite looping.

Yes that's true. I just think the code in the function could be a lot simpler if you don't have to check for pre-populated sizes.

shruti25ratnam: Sounds good to me. 
@yentsanglee do you agree with the suggestion of pulling the function outside the loop? Upside: code is much simpler. Downside: we may not be able to use static variables to store the actual collateral sizes retrieved from the first run, as you suggested earlier.

jhand2: You can still use static values right? Just preallocate outside the loop using the static values and set them inside the loop. The next call to this function will use the updated values

yentsanglee: I think Shruti's plan was to keep those static variables in oe_prealloc_quote_verification_collateral_args(). However, the function name oe_prealloc_quote_verification_collateral_args() implies that it is to be called once. In order for it to also store previous values, its purpose becomes ambiguous. Jordan is suggesting declaring static values in oe_get_sgx_quote_verification_collateral() and passing them in to 'pre-allocate' and 'store' functions separately and that makes sense to me.

shruti25ratnam: Sure I am making these changes. Thanks for the suggestions.

enclave/sgx/collateralinfo.c
```diff
@@ -169,6 +208,133 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
     return result;
 }
 
+void oe_prealloc_quote_verification_collateral_args(
+oe_get_sgx_quote_verification_collateral_args_t* buf, oe_get_sgx_quote_verification_collateral_args_t* default_sizes)
```
jhand2, line 212: nit: This doesn't seem to follow the style guide. Running scripts/format-code on each of these files should fix any issues.

enclave/sgx/collateralinfo.c
```diff
@@ -169,6 +208,133 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
...
+     * are passed into oe_get_quote_verification_
+     * collateral_ocall().
+     * */
+    default_sizes->tcb_info_size = TCBINFO_DEFAULT_SIZE;
+    default_sizes->tcb_info_issuer_chain_size = ALL_ISSUER_CHAIN_DEFAULT_SIZE;
+    default_sizes->pck_crl_size = PCK_CRL_DEFAULT_SIZE;
+    default_sizes->root_ca_crl_size = ROOT_CA_CRL_DEFAULT_SIZE;
+    default_sizes->pck_crl_issuer_chain_size = ALL_ISSUER_CHAIN_DEFAULT_SIZE;
+    default_sizes->qe_identity_issuer_chain_size = ALL_ISSUER_CHAIN_DEFAULT_SIZE;
+    default_sizes->qe_identity_size = QE_IDENTITY_DEFAULT_SIZE;
```
jhand2, line 225: This will always overwrite the cached values whenever this function is called. You can probably set these when initializing the struct in order to ensure they don't get updated on every call.

enclave/sgx/collateralinfo.c
shruti25ratnam: I had to use early return because the goto set of code is only supposed to execute for unsuccessful allocations. I agree about avoiding using goto statements all together. I can use a function if you suggest. I was trying to keep it short.

enclave/sgx/collateralinfo.c
```diff
@@ -169,6 +228,144 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
...
+void oe_prealloc_quote_verification_collateral_args(
+    oe_get_sgx_quote_verification_collateral_args_t* buf,
+    oe_get_sgx_quote_verification_collateral_args_t* default_sizes)
+{
+    /**These numbers are size estimates of various
+     * quote_verification_collateral_args that
+     * are passed into oe_get_quote_verification_
+     * collateral_ocall().
+     * */
+    default_sizes->tcb_info_size = TCBINFO_DEFAULT_SIZE;
```
shruti25ratnam, line 240: Please ignore this push this function didn't get updated.

enclave/sgx/collateralinfo.c
```diff
@@ -169,6 +228,130 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
...
+    {
+        buf->tcb_info_size = default_sizes->tcb_info_size;
+    }
+    else
+    {
+        goto done;
+    }
+
+    buf->tcb_info_issuer_chain =
+        (uint8_t*)calloc(1, default_sizes->tcb_info_size);
```
yentsanglee, line 249: >tcb_info_size [](start = 43, length = 13)

tcb_info_issuer_chain_size

jhand2:
Failure doesn't look related to this PR

bors r+

PR #2977, opened 2020-04-30
Add host_verify sample to host_verify lib

### LINKED ISSUE #2512: Host verification package does not include a sample
We would like samples for the host verification package to be written and installed. Jenkins, as part of package release, can install the host verification package and then build/run the sample.

We should also address the documentation in  https://github.com/openenclave/openenclave/issues/2513 at the same time.

ryanhsu19:
Modify a cmake file so that cpack adds the latest host_verify sample to the open-enclave-hostverify library.

Fix #2512.

### DIFF
--- a/samples/CMakeLists.txt
+++ b/samples/CMakeLists.txt
@@ -26,7 +26,8 @@ if (BUILD_ENCLAVES)
 endif ()
 
 install(DIRECTORY host_verify
-	DESTINATION ${CMAKE_INSTALL_DATADIR}/openenclave/samples)
+        DESTINATION ${CMAKE_INSTALL_DATADIR}/openenclave/samples
+        COMPONENT OEHOSTVERIFY)
 
 if (WIN32)
   install(FILES README_Windows.md


samples/CMakeLists.txt
```diff
@@ -26,7 +26,8 @@ if (BUILD_ENCLAVES)
 endif ()
 
 install(DIRECTORY host_verify
-	DESTINATION ${CMAKE_INSTALL_DATADIR}/openenclave/samples)
+	DESTINATION ${CMAKE_INSTALL_DATADIR}/openenclave/samples
+        COMPONENT OEHOSTVERIFY)
```
mingweishih, line 30:
```suggestion
    COMPONENT OEHOSTVERIFY)
```

samples/CMakeLists.txt
```diff
@@ -26,7 +26,8 @@ if (BUILD_ENCLAVES)
 endif ()
 
 install(DIRECTORY host_verify
-	DESTINATION ${CMAKE_INSTALL_DATADIR}/openenclave/samples)
+	DESTINATION ${CMAKE_INSTALL_DATADIR}/openenclave/samples
```
mingweishih, line 29: So I guess the ident of this also needs a fix 😄

PR #2966, opened 2020-04-30
Docs on setting up OpenEnclave SGX and Intel QPL in Linux on non-ACC machines

yufjiang:
This document is to provide a viable solution to enable openeclave to run SGX DCAP remote attestation on Linux in non ACC machines.

PR #2979, opened 2020-05-01
Reorganize docs to categorize by TEE

jhand2:
Note: this PR is almost entirely moving files around. You may have an easier time reviewing by just looking at [my branch](https://github.com/jhand2/openenclave/tree/gettingstarted_reorg) and looking around the new `docs/` folder structure.

In particular:
* Created docs/LibrarySupport
* Created docs/GettingStarted/SGX
* Created docs/GettingStarted/Contibutors/SGX
* Created docs/GettingStarted/Contributors/OP-TEE

I recently spent some time trying to get an OP-TEE environment set up and what I found is it is very confusing to find the right docs because they are not very well organized.

For many of the getting started docs, it is unclear what exactly they are for without opening up and reading the doc. Move all SGX-specific docs into SGX folders to clearly denote they are for SGX.

Additionally, group docs for support of various 3rd party libraries inside the enclave.

I am open to other suggestions if folks don't like this but overall I think our docs need a lot of cleaning up and de-duplication of information, this is just a starting point.


edl/optee/platform.edl
```diff
@@ -14,7 +14,12 @@
 
 enclave
 {
-    // There are currently no OP-TEE specific ocalls/ecalls which must be
-    // included. This file exists to allow an application to blindly include
-    // platform.edl and be compiled for any TEE.
+    // OP-TEE enclaves are not linked with --gc-sections, so functions that
```
jhand2, line 17: This was included accidentally. Will remove

yufjiang:
Thanks for reorganizing docs! My PR on non-ACC Intel QPL set up documentation was just checked in https://github.com/openenclave/openenclave/pull/2966. You might also want to update your PR to make sure that doc is also in the right directory.

jhand2:
Closing for now

Issue #1902, opened 2019-06-07
Drive options on how to include submodules for 3rd party code in Open Enclave


paulcallen:
Today we have several third party dependencies. Currently we pull in the source into the 3rd party directory.
Taking dependencies on code that is unchanged but not doing it as a submodule seems like more work than necessary, so we need. It also makes it harder to move to newer versions of the code, or makes it harder to see that there are newer updates to the code.
There are policies that govern how to safely take in 3rd party code as submodules within Microsoft. We can do this safely and securely.
This issue to investigate this, drive the team to a consensus and move forward on the decision.

anakrish:
Capturing my comments from #1820:

What I would prefer is the option where we can keep track of the changes we make to 3rd party code, and reapply them when we upgrade the 3rd party libraries.

The key point to keep in mind is that we don't use 3rd-party code as-is. Since the code must run within an enclave, we have to make changes to it.

Some 3rd party code like mbedtls have a nice config.h which can be edited to configure it they way we want.

Some times the changes involve directly editing code deep in the guts of the 3rd-party module (see 5e8ea11#diff-d45)

Not all changes can be upstreamed since these changes are specific to an SGX enclave.

Which of the two approaches (or some other approach) would allow us to keep track of the changes we have done to 3rd party code and automatically make sure the changes are reapplied when we upgrade the 3rd party modules?

The issue I see with the current approach is that when a 3rd party module is upgraded as in https://github.com/microsoft/openenclave/pull/1786/files, there are a large number of files to look at, it is not possible to figure out which of those files contain manual edits.
I would instead prefer the PR to just contain the set of our changes plus just a reference to the libunwind version we are upgrading to.

anakrish:
The recent PR https://github.com/microsoft/openenclave/commit/1cd2a42585037e28a497e402ea60ac576b9128ed is a good an example that shows that our current approach to 3rdparty code (copy them into our repo and edit them) does not scale well and is error prone. In a latest update of 3rd party edger8r code, we ended up losing a set of changes needed to make edger8r frontend (maintained by intel) avoid looking for GCC on windows.

The solution for 3rd party code needs to:
- make it possible for us to change 3rd party code where appropriate
- make it easy to incorporate the (may be modified) 3rd party code built and tested via our CI
- make it easier to upstream our changes to 3rd party code

andschwa:
[...]

PR #2984, opened 2020-05-01
Support code coverage tests

mingweishih:
This is the initial PR that supports code coverage testing on OE. The scope of the PR is as follows.
- Support clang-based gcov coverage test
- Support SGX ubuntu build only
- Focus on the non-third-party, enclave libraries

The PR consists of seven parts (in separate commits)
1. Add the libgcov source to the submodule (llvm compiler-rt)
2. Patches to the libgcov
   - Replace the unsupported file system calls (e.g., `mmap`)
   - Invoke hostfs initialization in the gcov initializer
3. Add the support of flock() used by the libgcov
4. Support the CODE_COVERAGE and BUILD_LIBGCOV cmake options
   - Build the oe with -DCODE_COVERAGE=ON to enable code coverage testing
   - After running ctest, use `make code_coverage` to generate the report
   - Update test cases to avoid failures when enabling code coverage tests
   - Build the oe with -DBUILD_LIBGCOV=ON to build and expose the libgcov to sdk users.
5. Add documentation
6. Jenkins integration
7. Ansible update to install dependencies

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,4 @@
 	path = 3rdparty/libgcov/compiler-rt
 	url = https://github.com/llvm-mirror/compiler-rt
 	branch = release_70
+	ignore = all
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,3 +12,7 @@
 [submodule "tools/oeedger8r-cpp"]
 	path = tools/oeedger8r-cpp
 	url = https://github.com/openenclave/oeedger8r-cpp
+[submodule "3rdparty/libgcov/compiler-rt"]
+	path = 3rdparty/libgcov/compiler-rt
+	url = https://github.com/llvm-mirror/compiler-rt
+	branch = release_70
--- a/.jenkins/pipelines/Azure/Linux/Jenkinsfile
+++ b/.jenkins/pipelines/Azure/Linux/Jenkinsfile
@@ -20,6 +20,46 @@ AGENTS_LABELS = [
     "acc-rhel-8-vanilla":       env.RHEL_8_VANILLA_CUSTOM_LABEL ?: "vanilla-rhel-8"
 ]
 
+def ACCCodeCoverageTest(String label, String compiler, String build_type) {
+    stage("${label} ${compiler} ${build_type} Code Coverage") {
+        node("${label}") {
+            timeout(GLOBAL_TIMEOUT_MINUTES) {
+                cleanWs()
+                checkout scm
+                def task = """
+                           cmake ${WORKSPACE}                                           \
+                               -G Ninja                                                 \
+                               -DCODE_COVERAGE=ON                                       \
+                               -DUSE_DEBUG_MALLOC=OFF                                   \
+                               -DCMAKE_BUILD_TYPE=${build_type}                         \
+                               -Wdev
+                           ninja -v
+                           ctest --output-on-failure --timeout ${CTEST_TIMEOUT_SECONDS}
+                           ninja code_coverage
+                           """
+                oe.Run(compiler, task)
+
+                // Publish the report via Cobertura Plugin.
+                cobertura coberturaReportFile: 'build/coverage/coverage.xml'
+
+                // Publish the result to the PR(s) via GitHub Coverage reporter Plugin.
+                // Workaround to obtain the PR id(s) as Bors does not us to grab them reliably.
+                def log = sh (script: "git log -1 | grep -Po '(Try #\\K|Merge #\\K)[^:]*'", returnStdout: true).trim()
+                def id_list = log.split(' #')
+                id_list.each {
+                    echo "PR ID: ${it}, REPOSITORY_NAME: ${REPOSITORY_NAME}"
+                    withEnv(["CHANGE_URL=https://github.com/${REPOSITORY_NAME}/pull/${it}"]) {
+                        publishCoverageGithub(filepath:'build/coverage/coverage.xml',
+                                              coverageXmlType: 'cobertura',
+                                              comparisonOption: [ value: 'optionFixedCoverage', fixedCoverage: '0.60' ],
+                                              coverageRateType: 'Line')
+                    }
+                }
+            }
+        }
+    }
+}
+
 def ACCTest(String label, String compiler, String build_type, List extra_cmake_args = [], List test_env = [], boolean fresh_install = false) {
     stage("${label} ${compiler} ${build_type}, extra_cmake_args: ${extra_cmake_args}, test_env: ${test_env}${fresh_install ? ", e2e" : ""}") {
         node(label) {
@@ -320,6 +360,7 @@ try{
                 "ACC1804 clang-7 Release LVI FULL Tests": { ACCTest(AGENTS_LABELS["acc-ubuntu-18.04"], 'clang-7', 'Release', ['-DLVI_MITIGATION=ControlFlow', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
                 "ACC1804 gcc Debug LVI":                  { ACCTest(AGENTS_LABELS["acc-ubuntu-18.04"], 'gcc',     'Debug',   ['-DLVI_MITIGATION=ControlFlow', '-DLVI_MITIGATION_SKIP_TESTS=ON']) },
                 "ACC1804 gcc Release LVI":                { ACCTest(AGENTS_LABELS["acc-ubuntu-18.04"], 'gcc',     'Release', ['-DLVI_MITIGATION=ControlFlow', '-DLVI_MITIGATION_SKIP_TESTS=ON']) },
+                "ACC1804 Code Coverage Test" :            { ACCCodeCoverageTest(AGENTS_LABELS["acc-ubuntu-18.04"], 'clang-7', 'Debug') },
 
                 "ACC1604 clang-7 Debug LVI e2e":          { ACCTest(AGENTS_LABELS["acc-ubuntu-16.04-vanilla"], 'clang-7', 'Debug',   ['-DLVI_MITIGATION=ControlFlow', '-DLVI_MITIGATION_SKIP_TESTS=ON'], [], true) },
                 "ACC1604 gcc Release LVI e2e":            { ACCTest(AGENTS_LABELS["acc-ubuntu-16.04-vanilla"], 'gcc',     'Release', ['-DLVI_MITIGATION=ControlFlow', '-DLVI_MITIGATION_SKIP_TESTS=ON'], [], true) },
--- a/3rdparty/libgcov/CMakeLists.txt
+++ b/3rdparty/libgcov/CMakeLists.txt
@@ -1,0 +1,31 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+check_submodule_not_empty(compiler-rt)
+
+apply_patches(patches compiler-rt)
+
+add_enclave_library(
+  gcov
+  STATIC
+  # Only the following three files are required to support GCOV-based code coverage analysis.
+  compiler-rt/lib/profile/GCDAProfiling.c
+  compiler-rt/lib/profile/InstrProfiling.c
+  compiler-rt/lib/profile/InstrProfilingUtil.c)
+
+maybe_build_using_clangw(gcov)
+
+enclave_compile_options(gcov PRIVATE -Wno-error)
+
+enclave_link_libraries(gcov PRIVATE oelibc oehostfs)
+
+enclave_include_directories(gcov PRIVATE src)
+
+install_enclaves(
+  TARGETS
+  gcov
+  EXPORT
+  openenclave-targets
+  ARCHIVE
+  DESTINATION
+  ${CMAKE_INSTALL_LIBDIR}/openenclave/enclave)
--- a/3rdparty/libgcov/README.md
+++ b/3rdparty/libgcov/README.md
@@ -1,0 +1,13 @@
+# libgcov
+
+The source is based on the profile library as part of [LLVM compiler-rt libraries](http://compiler-rt.llvm.org/).
+The current version is based on the release_70 branch (clang-7).
+
+## Modifications
+
+The following changes are made to the source to accommodate TEEs (see patches/ for more detail):
+- Initialize the `hostfs` during the GCOV initialization (`llvm_gcov_init()`).
+  - Ensure that the following filesystem operations work correctly.
+- Replace `mmap` and `ummap` (in `map_file()` and `unmap_file()`)
+  - `mmap` is not supported by OE. Use `fread()` instead.
+  - Alternatively, OE can support `mmap` only for code coverage testing (if there is a performance gap). Note that `mmap` implies using the host memory.
--- a/3rdparty/libgcov/patches/GCDAProfiling.c.patch
+++ b/3rdparty/libgcov/patches/GCDAProfiling.c.patch
@@ -1,0 +1,91 @@
+diff --git a/3rdparty/libgcov/compiler-rt/lib/profile/GCDAProfiling.c b/3rdparty/libgcov/compiler-rt/lib/profile/GCDAProfiling.c
+index cbca36551..3c42b1008 100644
+--- a/3rdparty/libgcov/compiler-rt/lib/profile/GCDAProfiling.c
++++ b/3rdparty/libgcov/compiler-rt/lib/profile/GCDAProfiling.c
+@@ -33,6 +33,7 @@
+ #else
+ #include <sys/mman.h>
+ #include <sys/file.h>
++#include <sys/mount.h>
+ #endif
+ 
+ #if defined(__FreeBSD__) && defined(__i386__)
+@@ -58,6 +59,11 @@ typedef unsigned int uint32_t;
+ typedef unsigned long long uint64_t;
+ #endif
+ 
++/* Beginning of the OE modfication */
++#include "openenclave/bits/module.h"
++#include "openenclave/bits/fs.h"
++/* End of the OE modification */
++
+ #include "InstrProfiling.h"
+ #include "InstrProfilingUtil.h"
+ 
+@@ -255,28 +261,42 @@ static int map_file() {
+   if (file_size == 0)
+     return -1;
+ 
++#if defined(COMPILER_RT_HAS_MMAP)
+   write_buffer = mmap(0, file_size, PROT_READ | PROT_WRITE,
+                       MAP_FILE | MAP_SHARED, fd, 0);
++#endif
++
++  write_buffer = malloc(file_size);
+   if (write_buffer == (void *)-1) {
+     int errnum = errno;
+     fprintf(stderr, "profiling: %s: cannot map: %s\n", filename,
+             strerror(errnum));
+     return -1;
+   }
++  fseek(output_file, 0L, SEEK_SET);
++  fread(write_buffer, file_size, 1, output_file);
+   return 0;
+ }
+ 
+ static void unmap_file() {
++#if defined(COMPILER_RT_HAS_MSYNC)
+   if (msync(write_buffer, file_size, MS_SYNC) == -1) {
+     int errnum = errno;
+     fprintf(stderr, "profiling: %s: cannot msync: %s\n", filename,
+             strerror(errnum));
+   }
++#endif
++
++  fseek(output_file, 0L, SEEK_SET);
++  fwrite(write_buffer, file_size, 1, output_file);
+ 
+   /* We explicitly ignore errors from unmapping because at this point the data
+    * is written and we don't care.
+    */
++#if defined(COMPILER_RT_HAS_MUNMAP)
+   (void)munmap(write_buffer, file_size);
++#endif
++  free(write_buffer);
+   write_buffer = NULL;
+   file_size = 0;
+ }
+@@ -598,6 +618,23 @@ void llvm_gcov_init(fn_ptr wfn, fn_ptr ffn) {
+   if (atexit_ran == 0) {
+     atexit_ran = 1;
+ 
++    /* Beginning of the OE modfication */
++    if (oe_load_module_host_file_system() != OE_OK) {
++      fprintf(stderr, "oe_load_module_host_file_system() failed\n");
++      exit(1);
++    }
++
++    if (mount("/", "/", OE_HOST_FILE_SYSTEM, 0, NULL) != 0) {
++      fprintf(stderr, "[llvm_gcov_init] mount() failed\n");
++      exit(1);
++
++    }
++    // Workaround: Force to install the atexit handler of fdtable
++    //             prior to the llvm handler.
++    FILE *f = fopen("/dev/null", "r");
++    fclose(f);
++    /* End of the OE modification */
++
+     /* Make sure we write out the data and delete the data structures. */
+     atexit(llvm_delete_flush_function_list);
+     atexit(llvm_delete_writeout_function_list);
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,6 +271,12 @@ if (CODE_COVERAGE)
       FATAL_ERROR
         "The CODE_COVERAGE option currently is not supported on Windows.")
   endif ()
+  if (USE_DEBUG_MALLOC)
+    message(
+      FATAL_ERROR
+        "The CODE_COVERAGE option currently is not supported when the USE_DEBUG_MALLOC option is ON."
+    )
+  endif ()
   message("-- CODE_COVERAGE set - enable code coverage tests")
 
   if (CMAKE_C_COMPILER_ID MATCHES GNU)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,6 +260,74 @@ elseif (WITH_EEID AND OE_TRUSTZONE)
   message(FATAL_ERROR "WITH_EEID is not supported on ARM yet.")
 endif ()
 
+# Option to build and expose the libgcov such that enclave applications can use.
+# Note that this option is only effective when CODE_COVERAGE is not set.
+option(BUILD_LIBGCOV "Build the code coverage library" OFF)
+
+option(CODE_COVERAGE "Enable code coverage testing" OFF)
+if (CODE_COVERAGE)
+  if (WIN32)
+    message(
+      FATAL_ERROR
+        "The CODE_COVERAGE option currently is not supported on Windows.")
+  endif ()
+  message("-- CODE_COVERAGE set - enable code coverage tests")
+
+  if (CMAKE_C_COMPILER_ID MATCHES GNU)
+    message(FATAL_ERROR "Code coverage is currently only supported by clang.")
+  else ()
+    if (CMAKE_C_COMPILER_VERSION VERSION_LESS 7 OR CMAKE_C_COMPILER_VERSION
+                                                   VERSION_GREATER 7.99)
+      message(
+        WARNING "Code coverage may not work with clang versions other than 7.")
+    endif ()
+  endif ()
+
+  find_program(LCOV "lcov")
+  if (LCOV)
+    message("-- LCOV is found")
+  else ()
+    message(FATAL_ERROR "LCOV is not found.")
+  endif ()
+
+  add_custom_target(
+    code_coverage
+    COMMAND mkdir -p coverage
+    COMMAND lcov -c -i -d ${CMAKE_BINARY_DIR} --gcov-tool
+            ${OE_SCRIPTSDIR}/code-coverage/llvm-gcov -o coverage/base.info
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+
+  set(FILTERED_LIST
+      "'${PROJECT_SOURCE_DIR}/tests/*'" "'${PROJECT_SOURCE_DIR}/3rdparty/*'"
+      "'${CMAKE_BINARY_DIR}/tests/*'" "'${CMAKE_BINARY_DIR}/3rdparty/*'")
+
+  add_custom_command(
+    TARGET code_coverage
+    COMMAND
+      lcov -d ${CMAKE_BINARY_DIR} --gcov-tool
+      ${OE_SCRIPTSDIR}/code-coverage/llvm-gcov -c -o cov.info --rc
+      lcov_branch_coverage=1
+    COMMAND lcov -a base.info -a cov.info -o cov_total.info --rc
+            lcov_branch_coverage=1
+    COMMAND lcov --remove cov_total.info ${FILTERED_LIST} -o cov_filtered.info
+            --rc lcov_branch_coverage=1
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/coverage)
+
+  add_custom_command(
+    TARGET code_coverage
+    DEPENDS ${CMAKE_BINARY_DIR}/coverage/cov_filtered.info
+    COMMAND
+      ${OE_SCRIPTSDIR}/code-coverage/lcov_cobertura
+      ${CMAKE_BINARY_DIR}/coverage/cov_filtered.info -o
+      ${CMAKE_BINARY_DIR}/coverage/coverage.xml
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
+
+  add_custom_target(
+    code_coverage_clean
+    COMMAND find ${CMAKE_BINARY_DIR} -type f -name '*.gcda' -delete
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+endif ()
+
 find_program(VALGRIND "valgrind")
 if (VALGRIND)
   set(MEMORYCHECK_COMMAND_OPTIONS "--leak-check=full --error-exitcode=1")
@@ -285,7 +353,9 @@ if (NOT COMPILE_SYSTEM_EDL)
     add_subdirectory(tests)
   endif ()
 
-  if (BUILD_ENCLAVES)
+  # Skip samples when enabling the code coverage test.
+  # Otherwise, all the samples need to explicitly link against the libgcov library.
+  if (BUILD_ENCLAVES AND (NOT CODE_COVERAGE))
     add_subdirectory(samples)
   endif ()
 endif ()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,18 +7,10 @@
 # https://rix0r.nl/blog/2015/08/13/cmake-guide/
 cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
 
-# Utility to ensure specific submodules are present.
-function (check_submodule_not_empty path)
-  file(GLOB submodule_files "${path}/*")
-  list(LENGTH submodule_files number)
-  if (${number} EQUAL 0)
-    message(
-      FATAL_ERROR
-        "Submodule ${path} is empty. You can initialize the submodule now with \
-         'git submodule update --recursive --init',\
-         or during 'git clone' by passing '--recursive'.")
-  endif ()
-endfunction ()
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+
+# Include utilities such as `check_submodule_not_empty`.
+include(utilities)
 
 # Read version from "VERSION" file.
 file(STRINGS "VERSION" OE_VERSION_WITH_V)
@@ -43,8 +35,6 @@ endif ()
 # http://cmake.3232098.n2.nabble.com/Prefer-clang-over-gcc-td7597742.html
 set(CMAKE_C_COMPILER_NAMES clang-7 cc)
 set(CMAKE_CXX_COMPILER_NAMES clang++-7 c++)
-
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 # Will be overwritten during the cmake initialization.
 set(LVI_MITIGATION
--- a/cmake/enclave_cmake_wrappers.cmake
+++ b/cmake/enclave_cmake_wrappers.cmake
@@ -144,3 +144,14 @@ function (install_enclaves)
     endif ()
   endforeach (target)
 endfunction (install_enclaves)
+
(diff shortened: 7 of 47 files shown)


mingweishih:
bors try

mingweishih:
bors try

mingweishih:
bors try-

mingweishih:
bors try

mingweishih:
bors try

mingweishih:
bors try

mingweishih:
bors try

tests/CMakeLists.txt
```diff
@@ -81,6 +80,11 @@ if (UNIX OR ADD_WINDOWS_ENCLAVE_TESTS OR USE_CLANGW)
         add_subdirectory(thread_local_no_tdata)
         add_subdirectory(VectorException)
         add_subdirectory(stack_smashing_protector)
+
+        # Skip the following tests when enabling code coverage tests.
```
anakrish, line 84: Why do we need to skip this directory?

3rdparty/libgcov/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+
+add_enclave_library(gcov STATIC
+  # Only the following three files are required to support GCOV-based code coverage analysis.
+  src/GCDAProfiling.c
+  src/InstrProfiling.c
+  src/InstrProfilingUtil.c)
+
+maybe_build_using_clangw(gcov)
+
+enclave_compile_options(gcov PRIVATE -Wno-error)
```
radhikaj, line 12: Would you have to say do not use system headers?

3rdparty/libgcov/README.md
```diff
@@ -0,0 +1,13 @@
+# libgcov
```
radhikaj, line 1: Should you include the license at the top of this file?

3rdparty/libgcov/src/GCDAProfiling.c
```diff
@@ -0,0 +1,645 @@
...
+|* the -insert-gcov-profiling LLVM pass.
+|*
+|* We emit files in a corrupt version of GCOV's "gcda" file format. These files
+|* are only close enough that LCOV will happily parse them. Anything that lcov
+|* ignores is missing.
+|*
+|* TODO: gcov is multi-process safe by having each exit open the existing file
+|* and append to it. We'd like to achieve that and be thread-safe too.
+|*
+\*===----------------------------------------------------------------------===*/
```
radhikaj, line 21: Should you include the license at the top of this file?

radhikaj:
This looks great! Thanks @mingweishih

3rdparty/libgcov/README.md
mingweishih: I guess README.md file is an exception?

3rdparty/libgcov/src/GCDAProfiling.c
mingweishih: All the files (including this one) in the directory are directly copied from [LLVM compiler-rt](https://github.com/llvm-mirror/compiler-rt/tree/master/lib/profile). Not sure if I need to add license if they didn't come with it. In this PR, I put the LICENSE file of the compiler-rt under 3rdparty/libgcov/.

mingweishih:
> This looks great!
> 
> Suggestion:
> Can we use a submodule for libgcov and update it only when building for code coverage?

I have thought about this suggestion. The only problem is that the source of libgcov is really a subset of  [LLVM compiler-rt](https://github.com/llvm-mirror/compiler-rt/tree/master/lib/profile). Git currently does not well support the submodule on only a sub-directory (opposed to the entire directory). If we don't mind pulling the entire directory with a submodule, then we can use it. What do you think?

mingweishih:
bors try

mingweishih:
bors try

mingweishih:
bors try

mingweishih:
bors try

BRMcLaren:
This is an excellent addition!

3rdparty/libgcov/LICENSE.TXT
```diff
@@ -0,0 +1,91 @@
+==============================================================================
+compiler_rt License
```
dthaler, line 2: Can this be added as a submodule instead of forking the code?
I would really like us to be using submodules for all additions so we don't keep adding to the debt we already have.

mingweishih: Sure. This PR is still considered as a draft and mainly for testing the CI integration. I will address the comments in the next iteration.

BRMcLaren:
bor try

BRMcLaren:
bors try-

BRMcLaren:
bors try

BRMcLaren:
bors try

mingweishih:
bors try

BRMcLaren:
bors try

mingweishih:
bors try

mingweishih:
bors try

BRMcLaren:
bors try

mingweishih:
bors try

mingweishih:
bors try

mingweishih:
bors try

3rdparty/libgcov/README.md
```diff
@@ -0,0 +1,13 @@
...
+The source is based on the profile library as part of [LLVM compiler-rt libraries](http://compiler-rt.llvm.org/).
+The current version is based on the release_70 branch (clang-7).
+
+## Modifications
+
+Few changes are made to the source to accommodate TEE (See patches/ for more detail).
+- Initialize the `hostfs` during the GCOV initialization (`llvm_gcov_init()`).
+  - Ensure that the following filesystem operations work correctly.
+- Replace `mmap` and `ummap` (in `map_file()` and `unmap_file()`)
+  - `mmap` is not supported by OE. USe `fread()` instead.
```
dthaler, line 12:
```suggestion
  - `mmap` is not supported by OE. Use `fread()` instead.
```

3rdparty/libgcov/patches/GCDAProfiling.c.patch
```diff
@@ -0,0 +1,91 @@
...
++      exit(1);
++    }
++
++    if (mount("/", "/", OE_HOST_FILE_SYSTEM, 0, NULL) != 0) {
++      fprintf(stderr, "[llvm_gcov_init] mount() failed\n");
++      exit(1);
++
++    }
++    // Workaround: Force to install the atexit handler of fdtable
++    //             prioir to the llvm handler.
```
dthaler, line 84:
```suggestion
+    //             prior to the llvm handler.
```

3rdparty/libgcov/patches/GCDAProfiling.c.patch
```diff
@@ -0,0 +1,91 @@
...
++/* End of the OE modification */
++
+ #include "InstrProfiling.h"
+ #include "InstrProfilingUtil.h"
+ 
+@@ -255,28 +261,42 @@ static int map_file() {
+   if (file_size == 0)
+     return -1;
+ 
++#if 0  // mmap is not supported by OE.
```
dthaler, line 29: I'd recommend following the convention used by the upstream project like:
#if defined(COMPILER_RT_HAS_MMAP)

Same in 47, 61, etc.  It would be good to upstream these changes so we don't need to keep patching.

host/windows/syscall.c
```diff
@@ -1426,6 +1426,14 @@ int oe_syscall_close_ocall(oe_host_fd_t fd)
     return ret;
 }
 
+// Unsupported yet.
```
dthaler, line 1429: is there an Issue # open to track this?  Or will it be done in this PR?

tests/atexit/enc/enc.cpp
```diff
@@ -51,7 +51,7 @@ void atexit_with_ecall_ecall(void)
...
     1,    /* ProductID */
     1,    /* SecurityVersion */
-    true, /* Debug */
-    8,    /* NumHeapPages */
-    8,    /* NumStackPages */
-    1);   /* NumTCS */
+    true, /* AllowDebug */
+    1024, /* HeapPageCount */
+    1024, /* StackPageCount */
+    1);   /* TCSCount */
```
dthaler, line 57: Revert the labels in comments.  They should match the labels that are recognized in .conf files so shouldn't change unless the .conf file parser is changed to recognize the new names.

tests/child_process/enc/enc.cpp
```diff
@@ -18,7 +18,8 @@ void get_magic_ecall(void* pdata)
...
     1,    /* ProductID */
     1,    /* SecurityVersion */
-    true, /* Debug */
-    8,    /* NumHeapPages */
-    8,    /* NumStackPages */
-    1);   /* NumTCS */
+    true, /* AllowDebug */
+    1024, /* HeapPageCount */
+    1024, /* StackPageCount */
+    1);   /* TCSCount */
```
dthaler, line 24: Revert the labels in comments. They should match the labels that are recognized in .conf files so shouldn't change unless the .conf file parser is changed to recognize the new names.

tests/child_thread/enc/enc.cpp
```diff
@@ -32,7 +32,7 @@ int stay_in_ocall_ecall()
...
     1,    /* ProductID */
     1,    /* SecurityVersion */
-    true, /* Debug */
-    8,    /* NumHeapPages */
-    8,    /* NumStackPages */
-    4);   /* NumTCS */
+    true, /* AllowDebug */
+    1024, /* HeapPageCount */
+    1024, /* StackPageCount */
+    4);   /* TCSCount */
```
dthaler, line 38: Revert the labels in comments. They should match the labels that are recognized in .conf files so shouldn't change unless the .conf file parser is changed to recognize the new names.

mingweishih:
bors try

tests/CMakeLists.txt
mingweishih: The reason is that the libgcov relies on host fs to write runtime data to the file upon the enclave termination. Tests in this directory (and also child_process) have the pattern of multiple invocations of `oe_terminate_enclave`. The first invocation will free the data structures associated to host fs and the writeout operations in the subsequent invocations will fail.

I've added comments to explain the reason.

host/windows/syscall.c
mingweishih: I'll create one. Probably will leave that to the Windows support (in a separate PR).

3rdparty/libgcov/patches/GCDAProfiling.c.patch
mingweishih: Thanks for the suggestion. Since we currently use older version of the library (llvm-7 in order to match clang-7). I think it would make for sense to the upstream the changes once we start supporting the latest version of clang/llvm.

3rdparty/libgcov/CMakeLists.txt
mingweishih: Are you referring to the potential conflicts?

mingweishih:
> Thanks for doing this. I have some questions/comments, but nothing blocking.

Thanks for the comments! Since you've requested changes earlier (turning into a blocker), can you take another look and see if you agree to remove the blocker?

.gitmodules
```diff
@@ -12,3 +12,8 @@
 [submodule "tools/oeedger8r-cpp"]
 	path = tools/oeedger8r-cpp
 	url = https://github.com/openenclave/oeedger8r-cpp
+[submodule "3rdparty/libgcov/compiler-rt"]
+	path = 3rdparty/libgcov/compiler-rt
+	url = https://github.com/llvm-mirror/compiler-rt
```
anakrish, line 17: +1 for submodule use.

.jenkins/pipelines/Azure/Linux/Jenkinsfile
```diff
@@ -20,6 +20,48 @@ AGENTS_LABELS = [
...
 ]
 
+def ACCCodeCoverageTest(String label, String compiler, String build_type) {
+    stage("${label} ${compiler} SGX1FLC ${build_type} Code Coverage") {
+        node("${label}") {
+            timeout(GLOBAL_TIMEOUT_MINUTES) {
+                cleanWs()
+                checkout scm
+                def task = """
+                           sudo apt install -y lcov python3 python3-pip llvm-7
```
anakrish, line 30: Should these be moved to prereqs?

3rdparty/libgcov/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+file(GLOB PATCHES patches/*.patch)
```
anakrish, line 4: add `check_submodule_not_empty`
E.g: https://github.com/openenclave/openenclave/blob/24a9b211e4ad7d4be46bf43297699459998a674f/tools/CMakeLists.txt#L5

3rdparty/libgcov/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+if (PATCHES)
+  foreach (PATCH ${PATCHES})
```
anakrish, line 6: Can this be refactored into an utility that we could use to apply patches to future submodules?
E.g: We could make MUSL a submodule and apply patches.

CMakeLists.txt
```diff
@@ -254,6 +254,69 @@ elseif (WITH_EEID AND OE_TRUSTZONE)
...
+    message(FATAL_ERROR "Code coverage is currently only supported by clang.")
+  else ()
+    if (CMAKE_C_COMPILER_VERSION VERSION_LESS 7 OR CMAKE_C_COMPILER_VERSION
+                                                   VERSION_GREATER 7.99)
+      message(
+        WARNING "Code coverage may not work with clang versions other than 7.")
+    endif ()
+  endif ()
+
+  find_program(LCOV "lcov")
```
anakrish, line 275: We need to add lcov to prereqs. What about lcov on Windows?

3rdparty/libgcov/README.md
```diff
@@ -0,0 +1,13 @@
+# libgcov
+
+The source is based on the profile library as part of [LLVM compiler-rt libraries](http://compiler-rt.llvm.org/).
+The current version is based on the release_70 branch (clang-7).
+
+## Modifications
+
+Few changes are made to the source to accommodate TEE (See patches/ for more detail).
```
dthaler, line 8:
```suggestion
The following changes are made to the source to accommodate TEEs (see patches/ for more detail):
```

CMakeLists.txt
```diff
@@ -254,6 +254,69 @@ elseif (WITH_EEID AND OE_TRUSTZONE)
...
+      lcov -d ${CMAKE_BINARY_DIR} --gcov-tool
+      ${OE_SCRIPTSDIR}/code-coverage/llvm-gcov -c -o cov.info --rc
+      lcov_branch_coverage=1
+    COMMAND lcov -a base.info -a cov.info -o cov_total.info --rc
+            lcov_branch_coverage=1
+    COMMAND lcov --remove cov_total.info ${FILTERED_LIST} -o cov_filtered.info
+            --rc lcov_branch_coverage=1
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/coverage)
+
+  add_custom_command(
```
anakrish, line 305: Would cmake guarantee that these custom commands added to `code_coverage` target are executed in order?

3rdparty/libgcov/patches/GCDAProfiling.c.patch
dthaler, line 29:
```suggestion
+#if defined(COMPILER_RT_HAS_MMAP)
```

3rdparty/libgcov/patches/GCDAProfiling.c.patch
```diff
@@ -0,0 +1,91 @@
...
+             strerror(errnum));
+     return -1;
+   }
++  fseek(output_file, 0L, SEEK_SET);
++  fread(write_buffer, file_size, 1, output_file);
+   return 0;
+ }
+ 
+ static void unmap_file() {
++#if 0 // msync is not supported by OE.
```
dthaler, line 47:
```suggestion
+#if defined(COMPILER_RT_HAS_MSYNC)
```

3rdparty/libgcov/patches/GCDAProfiling.c.patch
```diff
@@ -0,0 +1,91 @@
...
+   }
++#endif
++
++  fseek(output_file, 0L, SEEK_SET);
++  fwrite(write_buffer, file_size, 1, output_file);
+ 
+   /* We explicitly ignore errors from unmapping because at this point the data
+    * is written and we don't care.
+    */
++#if 0 // munmap is not supported by OE.
```
dthaler, line 61:
```suggestion
+#if defined(COMPILER_RT_HAS_MUNMAP)
```

enclave/CMakeLists.txt
```diff
@@ -52,6 +52,14 @@ if (CMAKE_C_COMPILER_ID MATCHES GNU)
   endif ()
 endif ()
 
+if (CODE_COVERAGE)
```
anakrish, line 55: Can we make `add_enclave_library` do the check for CODE_COVERAGE instead of each enclave library having to check for CODE_COVERAGE.

tests/CMakeLists.txt
```diff
@@ -117,7 +124,13 @@ endif ()
 
 if (UNIX)
   if (OE_SGX)
-    add_subdirectory(child_process)
+    # The tests in child_process fail when enabling code coverage analysis.
+    # The failure is due to the child process may terminate after the parent
```
dthaler, line 128:
```suggestion
    # The failure happens if the child process terminates after the parent
```

.jenkins/pipelines/Azure/Linux/Jenkinsfile
mingweishih: I'll leave the decision to @BRMcLaren.  I think doing that requires updating the images.

CMakeLists.txt
mingweishih: This command depends on the `cov_filtered.info` from the previous command so the order is guaranteed.

enclave/CMakeLists.txt
mingweishih: That would apply the option to every libraries. Currently the scope is only non-3rdpart ones. However, I can add a cmake function to do this logic.

3rdparty/libgcov/CMakeLists.txt
mingweishih: Done. Please check `cmake/utilities.cmake`

mingweishih:
bors try

CMakeLists.txt
mingweishih: Added a check for the OS type.

enclave/CMakeLists.txt
mingweishih: Added a wrapper `enclave_enable_code_coverage`

.jenkins/pipelines/Azure/Linux/Jenkinsfile
BRMcLaren: It would be better if we did move to prereqs, it would require an ansible change.

mingweishih: I have the necessary changes for ansible already, do you want me to include them in this PR?
I would assume this also means you're going to update the image, and after that we should validate that with another bors try, correct?

mingweishih:
@BRMcLaren Added ansible changes, please take a look.

scripts/ansible/roles/linux/openenclave/vars/ubuntu/main.yml
```diff
@@ -22,9 +22,11 @@ apt_packages:
   - "gcc"
```
BRMcLaren, line 22: Hey @mingweishih , are these components needed for anything other than the code coverage components you have added?

If not, we should probably not add them here as they would be added by every user using the ansible scripts.

If we take a look [here](https://github.com/openenclave/openenclave/tree/master/scripts/ansible) this [yaml](https://github.com/openenclave/openenclave/blob/master/scripts/ansible/oe-contributors-acc-setup.yml) and [this](https://github.com/openenclave/openenclave/blob/master/scripts/ansible/oe-contributors-acc-setup-no-driver.yml) differ slightly because one doesn't install the driver. I think you want something like this whereas we have a template for constructing code coverage bits that includes all the user experience, but then also adds

```
    - import_role:
        name: linux/cicd
        tasks_from: code-coverage.yml
```

then maybe your code coverage.yml could contain something like
```

apt_packages:
  - "lcov"
  - "python3-pip"
``` 

and then if we add it [here](https://github.com/openenclave/openenclave/blob/master/scripts/ansible/oe-linux-acc-setup.yml) is probably the best way to do it.

Your previous observation that we would need to rebuild the images is, unfortunately, true. But if you check it works locally and then do something like [this](https://github.com/openenclave/openenclave/blob/master/.jenkins/pipelines/Azure/Linux/Jenkinsfile#L32) line you will be 99% sure it works and then we can rebuild everything and get you merged in.

mingweishih:
bors try

CMakeLists.txt
```diff
@@ -254,6 +244,74 @@ elseif (WITH_EEID AND OE_TRUSTZONE)
...
+  endif ()
+
+  find_program(LCOV "lcov")
+  if (LCOV)
+    message("-- LCOV is found")
+  else ()
+    message(FATAL_ERROR "LCOV is not found.")
+  endif ()
+
+  add_custom_target(
```
anakrish, line 277: @mingweishih 
I referred to this PR while implementing Code Coverage for oeedger8r. https://github.com/openenclave/oeedger8r-cpp/pull/16
Some obervations:
- Moving the cmake commands to the code-coverage script simplifies cmake
- The coverage workflow seems to be to
   - Generate base info
   - Run program/tests
   - Generate coverage info
  Here we don't seem to be running the program/tests between the baseline info and coverage info.
  Can you double check what is the correct order of steps?
- GCC docs say that -coverage or -fprofile-arcs must also be passed to the linker.

cmake/enclave_cmake_wrappers.cmake
```diff
@@ -144,3 +144,14 @@ function (install_enclaves)
...
+
+function (enclave_enable_code_coverage NAME)
+  if (NOT CODE_COVERAGE)
+    return()
+  endif ()
+
+  # Enable code coverage.
+  enclave_compile_options(${NAME} PRIVATE -g -O0 -fprofile-arcs -ftest-coverage)
+  # Link against libgcov.
+  enclave_link_libraries(${NAME} PRIVATE gcov)
```
anakrish, line 156: GCC docs say that -fprofile-arcs must also be passed to the linker. Is that equivalent to just linking with gcov?

mingweishih: Yes, that's correct (we need to link against our customized libgcov). The `-fprofile-arcs` option to the linker will pick the library installed in the system.

CMakeLists.txt
mingweishih: My original thought is simply that we have `make test` or `ctest` to run the test already, so I `make code-coverage` as complementary to that.

Note that even generating base info (with `lcov -i`) can be independent from the test run.

anakrish: But the lcov docs don't recommend running `lcov -i` after the test:
https://linux.die.net/man/1/lcov

**--initial**
Capture initial zero coverage data.
Run lcov with -c and this option on the directories containing .bb, .bbg or .gcno files **before running any test case.** The result is a "baseline" coverage data file that contains zero coverage for every instrumented line.  **Combine this data file (using lcov -a) with coverage data files captured after a test run to ensure that the percentage of total lines covered is correct even when not all source code files were loaded during the test.**

    Recommended procedure when capturing data for a test case:

    1. create baseline coverage data file

        # lcov -c -i -d appdir -o app_base.info

    2. perform test

        # appdir/test

    3. create test coverage data file

        # lcov -c -d appdir -o app_test.info

    4. combine baseline and test coverage data

        # lcov -a app_base.info -a app_test.info -o app_total.info

mingweishih: Sure, I'll take the recommendation then.

anakrish: I don't think it is a must have for this PR. You can investigate and make changes in next PR.

mingweishih: That also sounds good to me! I'm waiting for @BRMcLaren to unblock this PR from the CI/CI infrastructure so that we can merge it.

I'll create an issue/PR for the follow-up.

.jenkins/infrastructure/build_azure_managed_images.Jenkinsfile
```diff
@@ -7,7 +7,7 @@ import java.time.format.DateTimeFormatter
 OECI_LIB_VERSION = env.OECI_LIB_VERSION ?: "master"
 oe = library("OpenEnclaveCommon@${OECI_LIB_VERSION}").jenkins.common.Openenclave.new()
 
-GLOBAL_TIMEOUT_MINUTES = 240
+GLOBAL_TIMEOUT_MINUTES = 480
```
dthaler, line 10: I don't see this change mentioned in the PR description.   Is this intentional?  Is it intended to be permanent?  Or a workaround for some other issue #?

3rdparty/libgcov/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+check_submodule_not_empty(compiler-rt)
+
+apply_patches(patches compiler-rt)
```
dthaler, line 6: Why do you need to do this at all?  For other submodules we depend on, we depend on a fork that has the patches already applied.   This PR seems to take a different approach that has to re-apply the patches every time.

.jenkins/infrastructure/build_azure_managed_images.Jenkinsfile
mingweishih: I was asked by @BRMcLaren to carry this changes as we need to build a new image (based on this PR) that includes the newly introduced dependencies.

It would be good for Brett to comment on this.

mingweishih:
> I think the approach of using the upstream and having to apply a patch as part of the build needs to be discussed more, since this diverges from how the other submodules work, which depend on a fork that has the patches already applied. We shouldn't break from precedent without an explicit discussion, since I don't see any advantage in this approach that consumes more cycles at build time.

I agree that we can discuss about this, maybe on the sig-arch meeting next week.  In fact, we also did similar thing to the mbedtls (see the [patch](https://github.com/openenclave/openenclave/blob/master/3rdparty/mbedtls/0001-Patch-x509write_crt.c-for-attestedTLS.patch).

anakrish:
> > I think the approach of using the upstream and having to apply a patch as part of the build needs to be discussed more, since this diverges from how the other submodules work, which depend on a fork that has the patches already applied. We shouldn't break from precedent without an explicit discussion, since I don't see any advantage in this approach that consumes more cycles at build time.
> 
> I agree that we can discuss about this, maybe on the sig-arch meeting next week. In fact, we also did similar thing to the mbedtls (see the [patch](https://github.com/openenclave/openenclave/blob/master/3rdparty/mbedtls/0001-Patch-x509write_crt.c-for-attestedTLS.patch).

I agree that we need to formalize whether to apply patches to a submodule during configure time, or whether to have the submodule point to a patched fork repository.

Currently, we seem to be favoring applying patches during configure time to in repo copies of 3rdparty repositories. 
Another example: [MUSL](https://github.com/openenclave/openenclave/tree/master/3rdparty/musl/patches)

dthaler:
> Currently, we seem to be favoring applying patches during configure time to in repo copies of 3rdparty repositories.
> Another example: [MUSL](https://github.com/openenclave/openenclave/tree/master/3rdparty/musl/patches)

But I believe we already agreed that repo copies were bad.   I suspect the "patch" mechanism was done specific to the "copy" version since otherwise you can't easily do diffs.  The submodule mechanism is much cleaner and so I argue we should not pollute the submodule mechanism with the inferior "patch" mechanism.

anakrish:
> > Currently, we seem to be favoring applying patches during configure time to in repo copies of 3rdparty repositories.
> > Another example: [MUSL](https://github.com/openenclave/openenclave/tree/master/3rdparty/musl/patches)
> 
> But I believe we already agreed that repo copies were bad. I suspect the "patch" mechanism was done specific to the "copy" version since otherwise you can't easily do diffs. The submodule mechanism is much cleaner and so I argue we should not pollute the submodule mechanism with the inferior "patch" mechanism.

I don't think anyone is arguing against that. I also think a forked repo is the right way to go.

> For other submodules we depend on, we depend on a fork that has the patches already applied. This PR seems to take a different approach that has to re-apply the patches every time.

Do we already have  patched forked repositories to which submodules point to?

mingweishih:
> > > Currently, we seem to be favoring applying patches during configure time to in repo copies of 3rdparty repositories.
> > > Another example: [MUSL](https://github.com/openenclave/openenclave/tree/master/3rdparty/musl/patches)
> > 
> > 
> > But I believe we already agreed that repo copies were bad. I suspect the "patch" mechanism was done specific to the "copy" version since otherwise you can't easily do diffs. The submodule mechanism is much cleaner and so I argue we should not pollute the submodule mechanism with the inferior "patch" mechanism.
> 
> I don't think anyone is arguing against that. I also think a forked repo is the right way to go.
> 
> > For other submodules we depend on, we depend on a fork that has the patches already applied. This PR seems to take a different approach that has to re-apply the patches every time.
> 
> Do we already have patched forked repositories to which submodules point to?

So I assume this direction means that eventually we will have
openenclave/musl
openenclave/mbedtls 
...
?

anakrish:
> > > > Currently, we seem to be favoring applying patches during configure time to in repo copies of 3rdparty repositories.
> > > > Another example: [MUSL](https://github.com/openenclave/openenclave/tree/master/3rdparty/musl/patches)
> > > 
> > > 
> > > But I believe we already agreed that repo copies were bad. I suspect the "patch" mechanism was done specific to the "copy" version since otherwise you can't easily do diffs. The submodule mechanism is much cleaner and so I argue we should not pollute the submodule mechanism with the inferior "patch" mechanism.
> > 
> > 
> > I don't think anyone is arguing against that. I also think a forked repo is the right way to go.
> > > For other submodules we depend on, we depend on a fork that has the patches already applied. This PR seems to take a different approach that has to re-apply the patches every time.
> > 
> > 
> > Do we already have patched forked repositories to which submodules point to?
> 
> So I assume this direction means that eventually we will have
> openenclave/musl
> openenclave/mbedtls
> ...
> ?

I'd love it if the existing 3rparty code were moved to submodules. Currently it is hard to figure out what the patches are.

anakrish:
> > > > > Currently, we seem to be favoring applying patches during configure time to in repo copies of 3rdparty repositories.
> > > > > Another example: [MUSL](https://github.com/openenclave/openenclave/tree/master/3rdparty/musl/patches)
> > > > 
> > > > 
> > > > But I believe we already agreed that repo copies were bad. I suspect the "patch" mechanism was done specific to the "copy" version since otherwise you can't easily do diffs. The submodule mechanism is much cleaner and so I argue we should not pollute the submodule mechanism with the inferior "patch" mechanism.
> > > 
> > > 
> > > I don't think anyone is arguing against that. I also think a forked repo is the right way to go.
> > > > For other submodules we depend on, we depend on a fork that has the patches already applied. This PR seems to take a different approach that has to re-apply the patches every time.
> > > 
> > > 
> > > Do we already have patched forked repositories to which submodules point to?
> > 
> > 
> > So I assume this direction means that eventually we will have
> > openenclave/musl
> > openenclave/mbedtls
> > ...
> > ?
> 
> I'd love it if the existing 3rparty code were moved to submodules. Currently it is hard to figure out what the patches are.

anakrish:
@mingweishih Sorry for closing an reopening. I pressed the wrong button.

mingweishih:
> > > > > > Currently, we seem to be favoring applying patches during configure time to in repo copies of 3rdparty repositories.
> > > > > > Another example: [MUSL](https://github.com/openenclave/openenclave/tree/master/3rdparty/musl/patches)
> > > > > 
> > > > > 
> > > > > But I believe we already agreed that repo copies were bad. I suspect the "patch" mechanism was done specific to the "copy" version since otherwise you can't easily do diffs. The submodule mechanism is much cleaner and so I argue we should not pollute the submodule mechanism with the inferior "patch" mechanism.
> > > > 
> > > > 
> > > > I don't think anyone is arguing against that. I also think a forked repo is the right way to go.
> > > > > For other submodules we depend on, we depend on a fork that has the patches already applied. This PR seems to take a different approach that has to re-apply the patches every time.
> > > > 
> > > > 
> > > > Do we already have patched forked repositories to which submodules point to?
> > > 
> > > 
> > > So I assume this direction means that eventually we will have
> > > openenclave/musl
> > > openenclave/mbedtls
> > > ...
> > > ?
> > 
> > 
> > I'd love it if the existing 3rparty code were moved to submodules. Currently it is hard to figure out what the patches are.

I do like the idea of using submodules. However, I think maintaining our own fork seems a different thing. I'll leave the discussion to the sig-arch meeting.

dthaler:
Per previous OE SDK discussion (when the submodules dependencies on optee were merged into master), forks should be temporary until the patches get merged upstream.

mingweishih:
> Per previous OE SDK discussion (when the submodules dependencies on optee were merged into master), forks should be temporary until the patches get merged upstream.

What if the patches are too oe-specific so that unlikely go upstream? I think that may be the case for our existing patches.

anakrish:
> > Per previous OE SDK discussion (when the submodules dependencies on optee were merged into master), forks should be temporary until the patches get merged upstream.
> 
> What if the patches are too oe-specific so that unlikely go upstream? I think that may be the case for our existing patches.

In any case it is easier to manage patches if they are applied onto a fork. You get the benefit of git, ability to easily diff, pull requests feedback, and can also make sure you don't lose patches when you pull an updated version of the upstream repo.

Admittedly all the above are possible by maintaining patches as diff files; but it is demonstrably harder, IMHO.

mingweishih:
> > > Per previous OE SDK discussion (when the submodules dependencies on optee were merged into master), forks should be temporary until the patches get merged upstream.
> > 
> > 
> > What if the patches are too oe-specific so that unlikely go upstream? I think that may be the case for our existing patches.
> 
> In any case it is easier to manage patches if they are applied onto a fork. You get the benefit of git, ability to easily diff, pull requests feedback, and can also make sure you don't lose patches when you pull an updated version of the upstream repo.
> 
> Admittedly all the above are possible by maintaining patches as diff files; but it is demonstrably harder, IMHO.

I would say the assumption behind these benefits are the patches can eventually go upstream. If not, it's also difficult to synchronize the polluted forked repo (when there is a conflict).

anakrish:
> > > > Per previous OE SDK discussion (when the submodules dependencies on optee were merged into master), forks should be temporary until the patches get merged upstream.
> > > 
> > > 
> > > What if the patches are too oe-specific so that unlikely go upstream? I think that may be the case for our existing patches.
> > 
> > 
> > In any case it is easier to manage patches if they are applied onto a fork. You get the benefit of git, ability to easily diff, pull requests feedback, and can also make sure you don't lose patches when you pull an updated version of the upstream repo.
> > Admittedly all the above are possible by maintaining patches as diff files; but it is demonstrably harder, IMHO.
> 
> I would say the assumption behind these benefits are the patches can eventually go upstream. If not, it's also difficult to synchronize the polluted forked repo (when there is a conflict).

Actually you can find counter examples in CMakeLists.txt of libunwind, MUSL where just maintaining a patched fork would be beneficial; even if the patches may not be appropriate for upstreaming.

Curious about this: currently when we apply patches, we copy the files to build folder and apply the patches. If we apply patches in place in the source folder, won't git status report the submodule as containing local modifications?

mingweishih:
> > > > > Per previous OE SDK discussion (when the submodules dependencies on optee were merged into master), forks should be temporary until the patches get merged upstream.
> > > > 
> > > > 
> > > > What if the patches are too oe-specific so that unlikely go upstream? I think that may be the case for our existing patches.
> > > 
> > > 
> > > In any case it is easier to manage patches if they are applied onto a fork. You get the benefit of git, ability to easily diff, pull requests feedback, and can also make sure you don't lose patches when you pull an updated version of the upstream repo.
> > > Admittedly all the above are possible by maintaining patches as diff files; but it is demonstrably harder, IMHO.
> > 
> > 
> > I would say the assumption behind these benefits are the patches can eventually go upstream. If not, it's also difficult to synchronize the polluted forked repo (when there is a conflict).
> 
> Actually you can find counter examples in CMakeLists.txt of libunwind, MUSL where just maintaining a patched fork would be beneficial; even if the patches may not be appropriate for upstreaming.
> 
> Curious about this: currently when we apply patches, we copy the files to build folder and apply the patches. If we apply patches in place in the source folder, won't git status report the submodule as containing local modifications?

I'm fine if people think maintaining the patches fork is beneficial. It's just not that clear to me. But since this PR hits this issue, I agree we need a consistent way for doing that.

In this PR, I simply set the ignore = all in the submodules configuration.

anakrish:
Previous discussion related to 3rdparty code/submodules: https://github.com/openenclave/openenclave/issues/1902

dthaler:
> Previous discussion related to 3rdparty code/submodules: #1902

And more of the discussion is in 
https://github.com/openenclave/openenclave/pull/1820/files/f387980d1b8b55619bf7574f897148d78154bde8#diff-8903239df476d7401cf9e76af0252622 (although some was in a meeting to resolve those comments)

mingweishih:
> Previous discussion related to 3rdparty code/submodules: #1902

Thanks for the references. The discussion there (pointed out by @anakrish ) captured most of my thoughts. In addition, I'm also leaning toward the similar approach as Asylo, mentioned in the thread; i.e., like this PR currently does, make the original 3rdparty library as submodule and then apply patches.

dthaler:
Per SIG-arch meeting, please file an issue to convert to a submodule.

mingweishih:
> Per SIG-arch meeting, need to file an issue to convert libgcov to a fork.

 I don't think we have reach of the conclusion of using patched fork. The follow-up on this discussion is tracked in #3120.

mingweishih:
bors try

mingweishih:
bors try

mingweishih:
bors try

PR #2985, opened 2020-05-02
Update SGX_QuoteEx_Integration.md to use API in PR 2949

shnwc:
Update SGX_QuoteEx_Integration.md to use the new API proposed in PR #2949 and to add a Proposal section.

### DIFF
--- a/docs/DesignDocs/SGX_QuoteEx_Integration.md
+++ b/docs/DesignDocs/SGX_QuoteEx_Integration.md
@@ -27,10 +27,10 @@ and outside of data centers.
 
 The proposed extension only changes the internal implementation of the OE SDK
 attestation software stack. It does not impact the
-[OE SDK attestation API](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/CustomAttestation_V3.md).
+[OE SDK attestation API](https://github.com/openenclave/openenclave/pull/2949).
 With the integration of the quote-ex library, an attester application enclave's
-call to OE SDK API `oe_get_attester_plugins()` returns the list of all the
-SGX evidence attester plugins available to the calling enclave instance.
+call to OE SDK API `oe_attester_initialize()` triggers enumeration and
+registration of all supported attester plugins
 
 Integration of the quote-ex library depends on the installation of the
 Intel® SGX SDK quote-ex library package and its dependencies,
@@ -65,7 +65,7 @@ The SGX ECDSA-p256 attester plugin is implemented in code file
 `enclave/sgx/attester.c` and other relevant enclave-side and host-side code files,
 called enclave-side and host-side plugin libraries in this document.
 he enclave-side plugin library interacts with the host-side plugin library
-via OCALLs defined in interface definition file `common/sgx/sgx.edl`.
+via OCALLs defined in interface definition file `edl/sgx/platform.edl`.
 For SGX ECDSA-p256 evidence generation, there are 2 OCALLs:
 
 - `oe_get_qetarget_info_ocall(sgx_target_info_t* target_info)`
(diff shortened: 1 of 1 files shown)


docs/DesignDocs/SGX_QuoteEx_Integration.md
```diff
@@ -139,10 +139,19 @@ function takes an input attestation key ID in its parameter list.
 With DCAP, the quote generation can be done either in-process,
 or out-of-process by working with a background service (called AESM) running
 on the same platform.
+  - For in-process quote generation, the process that calls the DCAP library
+  is required to run in an account added to a special group `sgx_prv`,
+  as documented in the DCAP library [readme](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/README.md)
+  and the DCAP driver [readme](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/driver/linux/README.md).
+    - Note: `sgx_prv` is an SGX provisioning access control mechanism implemented
+    in the SGX Linux driver since version V1.22, and in the
+    upstream SGX Linux driver.
```
dthaler, line 148: Please add the equivalent statements about Windows, since the rest of this document does not seem to be specific to Linux

shnwc: Good catch! will add Windows-specific content.

docs/DesignDocs/SGX_QuoteEx_Integration.md
```diff
@@ -197,16 +206,28 @@ It first calls into the quote-ex library to check if the dependent background
 service is available.
 If so, the quote-ex library is used, otherwise the DCAP library is used.
 
+#### Proposal: Built-time Link with the quote-ex Library
+
+The proposal is to implement option 3, to support build-time link with only
```
CodeMonkeyLeet, line 211: Per the [SIG-Attestation meeting](https://hackmd.io/Xj6GpDSKSwuz5cZgQ0yg1A#13-May-2020), we shouldn't break the existing in-proc loading behavior of the existing `oe_get_report()` function until we deprecate that in favor of the extensible `oe_get_evidence()` APIs. This would be pending Intel's release of documentation for dealing with the `sgx_prv` security group with the upstream kernel driver as well.

In the meantime, we should support both and verify that:
- We can link both DCAP and quote-ex libraries to by default in the OE SDK cmake export at the same time.
- Callers of `oe_get_report()` continue to be able to load the QE in-proc, while the callers of `oe_get_evidence()` get the new behavior through loading quote-ex and using the out-of-proc AESM hosted QE when both libraries are linked. 

As a follow-up during implementation, can you also work with SIG-testing to ensure that we have the new dependencies available in the testing pipeline?

shnwc: Agreed that we should keep the existing in-proc behavior. Will update the proposal section. Will work with SIG-testing when implementation is getting started.

docs/DesignDocs/SGX_QuoteEx_Integration.md
```diff
@@ -139,10 +139,24 @@ function takes an input attestation key ID in its parameter list.
...
+    Enclave (QE) is required to run in an account added to a special group `sgx_prv`,
+    as documented in the DCAP library [readme](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/README.md) and the DCAP driver [readme](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/driver/linux/README.md).
+      - Note: `sgx_prv` is an SGX provisioning access control mechanism implemented
+      in the DCAP driver since version V1.22, and in the new upstream SGX Linux driver.
+    - Impact to DCAP library usage:
+      - For in-process quote generation, every process that calls the DCAP library
+      is required to run in an account that belongs to the special `sgx_prv` group.
+      - On the other hand, for out-of-process quote generation, only the AESM
+      service process account needs to be added to the special group `sgx_prv`.
+  - Note: on Windows platforms quote generation access control stays the same.
```
dthaler, line 154: "Stays the same" as what?   This is the "Background" section so it should explain what the current state is.

docs/DesignDocs/SGX_QuoteEx_Integration.md
```diff
@@ -139,10 +139,24 @@ function takes an input attestation key ID in its parameter list.
...
+  QE's are required to be white-listed.
```
dthaler, line 155: Whitelisted how?  By what?  Can you provide the same level of detail for Windows as for Linux?

docs/DesignDocs/SGX_QuoteEx_Integration.md
shnwc: Thanks! Committed edit to update the description on Windows quote generation access control.

docs/DesignDocs/SGX_QuoteEx_Integration.md
shnwc: Committed a new version with more details.

docs/DesignDocs/SGX_QuoteEx_Integration.md
dthaler: Thanks!

docs/DesignDocs/SGX_QuoteEx_Integration.md
dthaler: Thanks!

docs/DesignDocs/SGX_QuoteEx_Integration.md
```diff
@@ -197,16 +218,37 @@ It first calls into the quote-ex library to check if the dependent background
...
+
+#### Proposal: Built-time Link with the quote-ex Library
+
+The proposal is to start by implementing option 4, so that the existing OE SDK
+behavior for in-process quote generation stays the same, while enabling SGX
+EPID quote generation on properly configured platforms.
+
+At a future point when new quote-generation access control mechanisms are adopted
+on all platforms and the existing OE SDK behavior is deprecated, we will switch
+to implement option 2. With this option, OE SDK applications only link with the
```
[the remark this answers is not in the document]
dthaler: Meeting confirmed it is for both

shnwc:
bors try

dthaler:
Failure message seems to be:
> TASK [linux/openenclave : Install all the Open Enclave prerequisites YUM packages for development] ***
> fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to synchronize cache for repo 'epel'", "rc": 1, "results": []}

Issue #2993, opened 2020-05-05
Error The term 'pip' is not recognized as the name of a cmdle when running install-windows-prerqs.ps1 


jianlia:
On brand new VM, run  install-windows-prerqs.ps1 and failed with message "Error The term 'pip' is not recognized as the name of a cmdle "

It is regression from
https://github.com/openenclave/openenclave/commit/53e41c3dd88d58c0d7d58bad53d0980e5a2cd75d

[2020-05-05T04:12:37.1093282+00:00] Start-ExecuteWithRetry attempt 0
[2020-05-05T04:12:39.0317513+00:00] Start-ExecuteWithRetry RetryMessage: Failed to install cmake_format. Retrying
[2020-05-05T04:12:42.0354311+00:00] Start-ExecuteWithRetry attempt 1
[2020-05-05T04:12:42.0510139+00:00] Start-ExecuteWithRetry RetryMessage: Failed to install cmake_format. Retrying
[2020-05-05T04:12:45.0536078+00:00] Start-ExecuteWithRetry attempt 2
[2020-05-05T04:12:45.0536078+00:00] Start-ExecuteWithRetry RetryMessage: Failed to install cmake_format. Retrying
[2020-05-05T04:12:48.0563770+00:00] Start-ExecuteWithRetry attempt 3
[2020-05-05T04:12:48.0563770+00:00] Start-ExecuteWithRetry RetryMessage: Failed to install cmake_format. Retrying
[2020-05-05T04:12:51.0613506+00:00] Start-ExecuteWithRetry attempt 4
[2020-05-05T04:12:51.0613506+00:00] Start-ExecuteWithRetry RetryMessage: Failed to install cmake_format. Retrying
[2020-05-05T04:12:54.0677441+00:00] Start-ExecuteWithRetry attempt 5
[2020-05-05T04:12:54.0832347+00:00] Start-ExecuteWithRetry RetryMessage: Failed to install cmake_format. Retrying
[2020-05-05T04:12:57.0896717+00:00] Start-ExecuteWithRetry attempt 6
[2020-05-05T04:12:57.0896717+00:00] Start-ExecuteWithRetry RetryMessage: Failed to install cmake_format. Retrying
[2020-05-05T04:13:00.0942188+00:00] Start-ExecuteWithRetry attempt 7
[2020-05-05T04:13:00.0942188+00:00] Start-ExecuteWithRetry RetryMessage: Failed to install cmake_format. Retrying
[...]

PR #2998, opened 2020-05-05
Install PIP on Windows

dthaler:
Addresses issue #2993

### DIFF
--- a/scripts/install-windows-prereqs.ps1
+++ b/scripts/install-windows-prereqs.ps1
@@ -34,6 +34,8 @@ Param(
     [string]$Python3ZipHash = 'FB65E5CD595AD01049F73B47BC0EE23FD03F0CBADC56CB318990CEE83B37761B',
     [string]$NSISURL = 'https://oejenkins.blob.core.windows.net/oejenkins/nsis-3.05-setup.exe',
     [string]$NSISHash = '1A3CC9401667547B9B9327A177B13485F7C59C2303D4B6183E7BC9E6C8D6BFDB',
+    [string]$GetPipURL = 'https://bootstrap.pypa.io/3.4/get-pip.py',
+    [string]$GetPipHash = '564FABC2FBABD9085A71F4A5E43DBF06D5CCEA9AB833E260F30EE38E8CE63A69',
     [Parameter(mandatory=$true)][string]$InstallPath,
     [Parameter(mandatory=$true)][ValidateSet("SGX1FLC", "SGX1", "SGX1FLC-NoDriver", "SGX1-NoDriver")][string]$LaunchConfiguration,
     [Parameter(mandatory=$true)][ValidateSet("None", "Azure")][string]$DCAPClientType
@@ -114,6 +116,11 @@ $PACKAGES = @{
         "url" = $Python3ZipURL
         "hash" = $Python3ZipHash
         "local_file" = Join-Path $PACKAGES_DIRECTORY "Python3.zip"
+    }
+    "get-pip" = @{
+        "url" = $GetPipURL
+        "hash" = $GetPipHash
+        "local_file" = Join-Path $PACKAGES_DIRECTORY "get-pip.py"
     }
     "nsis" = @{
         "url" = $NSISURL
@@ -359,6 +366,17 @@ function Install-Python3 {
     New-Directory -Path $installDir -RemoveExisting
     Move-Item -Path "$tempInstallDir\*" -Destination $installDir
     Add-ToSystemPath -Path $installDir
+
+    Start-ExecuteWithRetry -ScriptBlock {
+        # Install PIP
+        python $PACKAGES["get-pip"]["local_file"]
+        $Scripts = Join-Path $installDir "Scripts"
+        Add-ToSystemPath -Path $Scripts
+
+        # Enable site packages so that PIP will run, by uncommenting out 'import site'
+        $configFile = Join-Path $installdir "python37._pth"
+        Set-Content -Path $configFile -Value "python37.zip`n.`n`nimport site"
+    } -MaxRetryCount $RetryCount -RetryInterval 3 -RetryMessage "Failed to install PIP. Retrying"
 
     Start-ExecuteWithRetry -ScriptBlock {
         pip install cmake_format
--- a/docs/GettingStartedDocs/Contributors/WindowsManualInstallPrereqs.md
+++ b/docs/GettingStartedDocs/Contributors/WindowsManualInstallPrereqs.md
@@ -87,8 +87,9 @@ C:\Users\test\AppData\Roaming\npm\esy
 ## Python 3
 
 Install [Python 3 for Windows](https://www.python.org/downloads/windows/) and ensure that python.exe is available in your PATH.
+Make sure the checkbox for PIP is checked when installing.
 
-Python 3 is used as part of the mbedtls tests.
+Python 3 is used as part of the mbedtls tests and for cmake-format.
 
 ## ShellCheck
 


scripts/install-windows-prereqs.ps1
```diff
@@ -360,6 +360,24 @@ function Install-Python3 {
     Move-Item -Path "$tempInstallDir\*" -Destination $installDir
     Add-ToSystemPath -Path $installDir
 
+    Start-ExecuteWithRetry -ScriptBlock {
+        # Install the PIP installer
+        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
```
jianlia, line 365: Can use Start-FileDownload to reduce some common code

dthaler: Right.  I started with that but was having problems with it at first and once it was working, didn't want to spend time potentially breaking it since the fix might be blocking folks.   Did it work for you as is.

dthaler: I updated it to use Start-FileDownload and also made it check the hash like it does for other files downloaded

jianlia: pip issue resolve in my local machine being tested. Thanks

dthaler:
bors try

scripts/install-windows-prereqs.ps1
```diff
@@ -360,6 +362,31 @@ function Install-Python3 {
     Move-Item -Path "$tempInstallDir\*" -Destination $installDir
     Add-ToSystemPath -Path $installDir
 
+    Start-ExecuteWithRetry -ScriptBlock {
+        # Install the PIP installer
+        $Destination = Join-Path $installDir "get-pip.py"
+        Start-FileDownload -URL $GetPipURL `
```
CodeMonkeyLeet, line 368: Why not just use the existing infrastructure for this by adding the PIP package to `$PACKAGES` where it will get downloaded and hash checked as part of the existing `Start-LocalPackagesDownload` function?

scripts/install-windows-prereqs.ps1
```diff
@@ -360,6 +362,31 @@ function Install-Python3 {
...
+        }
+        else
+        {
+            Write-Output "Computed hash ($downloaded_hash) matches expected hash ($GetPipHash)"
+        }
+
+        # Install PIP
+        python $Destination
+        $Scripts = Join-Path $installDir "Scripts"
+        $env:PATH += ";" + $Scripts
```
CodeMonkeyLeet, line 383: This only updates the `PATH` for this instance of the shell; it's sufficient for allowing the next script block to succeed, Shouldn't this just call `Add-ToSystemPath` instead, which updates the current shell $env:PATH as well as the global `PATH`? (The existing function also doesn't append duplicates)

scripts/install-windows-prereqs.ps1
```diff
@@ -34,6 +34,8 @@ Param(
     [string]$Python3ZipHash = 'FB65E5CD595AD01049F73B47BC0EE23FD03F0CBADC56CB318990CEE83B37761B',
     [string]$NSISURL = 'https://oejenkins.blob.core.windows.net/oejenkins/nsis-3.05-setup.exe',
     [string]$NSISHash = '1A3CC9401667547B9B9327A177B13485F7C59C2303D4B6183E7BC9E6C8D6BFDB',
+    [string]$GetPipURL = 'https://bootstrap.pypa.io/get-pip.py',
+    [string]$GetPipHash = 'CE486CDDAC44E99496A702AA5C06C5028414EF48FDFD5242CD2FE559B13D4348',
```
CodeMonkeyLeet, line 38: Non-blocking: This is going to break us at some point; the URL points to the latest version of get-pip.py instead of a fixed version, which means we're going to get hash check failures; it's not a unique problem to this package, but have we considered targeting a fixed version, even if is not the latest version? (e.g. https://bootstrap.pypa.io/3.4/get-pip.py)

CodeMonkeyLeet:
@dthaler Can you also `--amend` your commit 9a387f8f826b3f5f41302471ce8677d1768cc999 so that the DCO signoff matches your GitHub email account? The DCO bot doesn't like the ntdev.microsoft.com email.

scripts/install-windows-prereqs.ps1
dthaler: Great question, let me see if I can do that.

dthaler: The problem with that is that PIP and cmake_format have to go inside the python directory, which is not created until well after Start-LocalPackagesDownload completes, and then Install-Python3 gets called which unzips the python download into that directory.  So at most it could be downloaded and then copied around later, but that seems more work than just delaying the download until it can be downloaded in place.   (That's similar to how esy gets installed by waiting until node is installed and downloading it then so the precedent already exists.)

dthaler: Done

scripts/install-windows-prereqs.ps1
dthaler: Done

dthaler:
> @dthaler Can you also `--amend` your commit [9a387f8](https://github.com/openenclave/openenclave/commit/9a387f8f826b3f5f41302471ce8677d1768cc999) so that the DCO signoff matches your GitHub email account? The DCO bot doesn't like the ntdev.microsoft.com email.

It was actually the other way around.  It didn't like having it match my github email account, and required it to be ntdev.microsoft.com which one of the two wasn't.  After squashing, it now matches.

scripts/install-windows-prereqs.ps1
CodeMonkeyLeet: I'm not sure I understand why get-pip.py needs to be copied anywhere after it's downloaded? In my experience, get-pip.py installs to the path of the python.exe instance that's executing it, so the only change required here is to add the get-pip.py entries to `$PACKAGES`, then after python is installed, call `python $PACKAGES["get-pip"]["local_file"]`.

scripts/install-windows-prereqs.ps1
dthaler: Done

scripts/install-windows-prereqs.ps1
dthaler: Already done :)

anakrish:
python3's pip is called pip3. Consider using pip3 instead of pip to avoid issues where both python2 and python3 are installed on the user's machine.

dthaler:
bors retry

dthaler:
> python3's pip is called pip3. Consider using pip3 instead of pip to avoid issues where both python2 and python3 are installed on the user's machine.

As far as I know, it's already using pip3.  This is what is installed:
```
C:\Users\Administrator>where pip
C:\Program Files\python-3.7.4\Scripts\pip.exe

C:\Users\Administrator>where pip3
C:\Program Files\python-3.7.4\Scripts\pip3.exe

C:\Users\Administrator>pip --version
pip 19.1.1 from c:\program files\python-3.7.4\lib\site-packages\pip (python 3.7)

C:\Users\Administrator>pip3 --version
pip 19.1.1 from c:\program files\python-3.7.4\lib\site-packages\pip (python 3.7)
```

BRMcLaren:
[This](https://oe-jenkins-tf.westeurope.cloudapp.azure.com/blue/organizations/jenkins/CI-CD_Infrastructure%2FOpenEnclave-E2E-Testing/detail/OpenEnclave-E2E-Testing/183/pipeline/) is a link to a full rebuild + regression of this PR, merging now.

Thanks @dthaler !

Issue #2041, opened 2019-07-18
oesign code is too SGX specific


dthaler:
The tool is "oesign" not "oesgxsign" and so should be usable for TrustZone etc.   However, currently SGX-specific code is sprinkled liberally throughout the codebase.  It is not cleanly factored into generic functionality vs SGX specific functionality to make it easy to add support for other types of TEEs like TrustZone.

HernanGatta:
OP-TEE ships with a tool aptly named `sign.py` that does what `oesign` does for OP-TEE TA's. That is, it signs TA's in a way that OP-TEE can understand and validate.

In light of #2042, the Python script should be re-written in C as a first step. If #2042 is solved ahead of @2041, then the rewrite may be done so as for it to execute in an enclave. If not, then care should be taken to rewrite the logic in a way that may easily be moved into an enclave when we @2042 is solved.

anakrish:
CC @radhikaj

johnkord:
As per our triage meeting, we should talk about converging our signing formats between TEEs.
We should schedule a meeting to sync up on this! @radhikaj will be scheduling a meeting sometime soon to discuss this (and other) collaborations.

PR #3004, opened 2020-05-07
Move debug signkey to SGX-specific directory

jhand2:
OE uses a well-known RSA signing key to sign debug SGX enclaves. This is only used by SGX, move it to an SGX-specific directory and only build it into SGX hosts.

I don't know for a fact that this is unused by OP-TEE so please correct me if I am wrong.


dthaler:
I agree it isn't used for OP-TEE, but my understanding is that it _should_ be in the future.

HernanGatta:
As it is, OP-TEE checks signatures on TAs on the basis of a public key that's baked into OP-TEE. The same key is exported during OP-TEE's build process and consumed by the SDK's build to sign TAs. This key is different from the key used for SGX debug signing, as Jordan's discovered.

Forgive me if the question is silly, but why bother signing debug enclaves? Is it just to test the signing mechanism? I was under the impression that for SGX, the CPU checks the signature of the enclave, and that wouldn't matter in debug mode. Or perhaps the enclave must be signed anyway, as per a requirement of SGX?

In OP-TEE, there isn't a concept of debug signing per-se. OP-TEE carries a baked-in public key that it uses to verify TA signatures. While one is debugging, one bakes that key in and signs all TAs with that key. As far as OP-TEE is concerned, it checks the signature on TAs the same way it would if the key wasn't a debugging key (so in some sense, OP-TEE is always in "release mode", if you will).

There is no mechanism in OP-TEE to update the signing key once OP-TEE is built. As such, one would never ship OP-TEE in firmware without baking in the proper public key.

As to having a mechanism to update the signing key out in the field à la FLC, that's a different discussion.

HernanGatta:
Addendum: `oesign` should be able to [sign OP-TEE TAs](https://github.com/openenclave/openenclave/issues/2041) in the future, and it could use the same key for OP-TEE TAs as for SGX enclaves both for debug and release signing (assuming the key formats are compatible; I'll add a comment about that in the issue I linked above).

jhand2:
Thanks for the background Hernan!

> why bother signing debug enclaves? Is it just to test the signing mechanism? I was under the impression that for SGX, the CPU checks the signature of the enclave, and that wouldn't matter in debug mode. Or perhaps the enclave must be signed anyway, as per a requirement of SGX?

Yes, I beleive SGX won't load anything that is unsigned.

> Addendum: `oesign` should be able to [sign OP-TEE TAs](https://github.com/openenclave/openenclave/issues/2041) in the future, and it could use the same key for OP-TEE TAs as for SGX enclaves both for debug and release signing (~assuming the key formats are compatible; I'll add a comment about that in the issue I linked above~ John mentioned key format convergence already).
>
> As to whether it's desirable to ensure that the debug key used for SGX is the same as the debug key used for OP-TEE, the only benefit I see would be to avoid confusing developers.
> 

The reason I'm doing this now is that I'm trying to parse out what is core to OE which all TEEs will use, and what is SGX-specific. This will allow us to more clearly define an interface which TEEs have to implement. To me, it felt cleaner to let a TEE manage it's own debug keys in whatever format it needs.

I don;t feel super strongly about this, so if you think this should stay in the common directory that is ok with me. It just looked like from code something that only SGX cared about.

dthaler:
Reading Hernan's response I guess I'd prefer making it be common across TEEs unless that's later proven to be impractical.  So I'd keep it in the common directory, and we can then make it start being used for OP-TEE debug signing too.

CodeMonkeyLeet:
@dthaler I disagree that there needs to be an OE SDK sanctioned debug key that spans TEEs.

- The key is only used by the SGX enclave loading stack when dealing with unsigned SGX enclaves. It is an implementation detail to provide a convenience developer feature for running debug SGX enclaves without signing them first.
- The key, as it exists, is encoded as a binary blob in a source file, and is not consumable in any generic way (e.g. we don't a public key certificate to verify the debug key MRSIGNER with).
- The key is not used by oesign. Even if/when additional work is done to support signing OP-TEE enclaves with oesign, debug signing would still be done with developer-owned debug keys, not this specific key, as is the case with SGX today.

It's an interesting question to discuss for SIG-arch as to whether a common debug key makes sense, but for the purposes of this PR, I think it's a much more accurate reflection of the implementation as it stands, absent any plan of record to promote this specific key value to one that OE SDK might want to maintain for all TEEs.

PR #3006, opened 2020-05-07
Enable 3 more tests in simulation mode

thomasten:
These tests don't need to be skipped in simulation mode:
- host_verify
- syscall/fs
- oesign-engine-sign-runnable


jhand2:
bors try

PR #3007, opened 2020-05-07
Add experimental plugins for EEID attestation.

wintersteiger:
This PR contributes support for launching enclaves with [EEID](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/ExtendedEnclaveInitializationData.md) and corresponding (SGX-only) attestation and verification plugins. This follows the discussion in #2647; most suggestions have been implemented as per the discussion. 

There are two points that I want to discuss, which I will highlight in comments.

CC @sylvanc @ad-l @letmaik @mikbras @prp

PR #3008, opened 2020-05-07
EEID plugins for SGX-LKL

wintersteiger:
This adds the [EEID ](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/ExtendedEnclaveInitializationData.md) attestation plugins to the `feature/sgx-lkl-support` branch so that they can be used in SGX-LKL enclaves. 

CC @jxyang @jhand2 @prp


wintersteiger:
@jhand2 `feature/sgx-lkl-support` seems to be a few commits behind master, would be nice if we could rebase it before/while merging this.

jhand2:
Closing in favor of  #3007

PR #3012, opened 2020-05-09
[CI/CD] Enable RHEL 8 attestation

ionutbalutoiu:
* Bump Intel SGX driver URLs for Ubuntu. This makes sure that we use the same version on both Ubuntu / RHEL in our Jenkins CI/CD machines.
* Enable attestation for RHEL 8:
  * Remove `flc_enabled=false` when building RHEL 8 Azure image.
  * Do no fail if `flc_enabled` is enabled on RHEL 8.
  * Create symlink for `libdcap_quoteprov.so` into `/usr/lib64`.
  * Add Intel SGX DCAP driver setup on RHEL 8.
  * Install Azure-DCAP-Client from stable branches on RHEL 8.
* Re-enable RHEL 8 testing stages. The linked issues are already solved.
* Refactor pipelines:
  * Fix spacing.
  * Remove code duplication for stages used with / without `FULL_TEST_SUITE`.
* Add RHEL 8 testing with HAS_QUOTE_PROVIDER=ON:
  * This is a validation that attestation is working as expected.
* Integrate RHEL8Test function into ACCTest
  * This commit also fixes space formating for ACCHostVerificationTest function.
* Cleanup Agnostic/Linux/Jenkinsfile
  * Rename `simulationTest` to `simulationContainerTest` since it runs testing into a container.
  * Change `simulationContainerTest` to accept `extra_cmake_args` instead of individual parameters for the cmake flags.

Signed-off-by: Ionut Balutoiu <ibalutoiu@cloudbasesolutions.com>

### DIFF
--- a/.jenkins/pipelines/Azure/Linux/Jenkinsfile
+++ b/.jenkins/pipelines/Azure/Linux/Jenkinsfile
@@ -237,75 +237,72 @@ properties([buildDiscarder(logRotator(artifactDaysToKeepStr: '90',
 
 try{
     oe.emailJobStatus('STARTED')
+    def testing_stages = [
+        "Host verification 1604 Release":         { ACCHostVerificationTest('16.04', 'Release') },
+        "Host verification 1804 Release":         { ACCHostVerificationTest('18.04', 'Release') },
+        "ACC1804 GNU gcc SGX1FLC":                { ACCGNUTest() },
+        "RHEL-8 simulation clang-8 SGX1 Release": { RHEL8Test('clang', 'Release', ["OE_SIMULATION=1"]) },
+        "RHEL-8 simulation clang-8 SGX1 Debug":   { RHEL8Test('clang', 'Debug',   ["OE_SIMULATION=1"]) },
+        "RHEL-8 simulation gcc-8 Release":        { RHEL8Test('gcc',   'Release', ["OE_SIMULATION=1"]) },
+        "RHEL-8 simulation gcc-8 SGX1 Debug":     { RHEL8Test('gcc',   'Debug',   ["OE_SIMULATION=1"]) },
+        "RHEL-8 ACC clang-8 Release":             { RHEL8Test('clang', 'Release') },
+        "RHEL-8 ACC clang-8 Debug":               { RHEL8Test('clang', 'Debug') },
+        "RHEL-8 ACC gcc-8 Release":               { RHEL8Test('gcc',   'Release') },
+        "RHEL-8 ACC gcc-8 Debug":                 { RHEL8Test('gcc',   'Debug') }
+    ]
     if(FULL_TEST_SUITE == "true") {
-      stage("Full Test Suite") {
-        parallel "Host verification 1604 Debug" :                           { ACCHostVerificationTest('16.04', 'Debug') },
-                 "Host verification 1604 Release" :                         { ACCHostVerificationTest('16.04', 'Release') },
-                 "Host verification 1804 Debug" :                           { ACCHostVerificationTest('18.04', 'Debug') },
-                 "Host verification 1804 Release" :                         { ACCHostVerificationTest('18.04', 'Release') },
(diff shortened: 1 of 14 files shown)


ionutbalutoiu:
I validated this PR via E2E and everything passed:
https://oe-jenkins-tf.westeurope.cloudapp.azure.com/blue/organizations/jenkins/CI-CD_Infrastructure%2FOpenEnclave-E2E-Testing/detail/OpenEnclave-E2E-Testing/188/pipeline

.jenkins/pipelines/Azure/Linux/Jenkinsfile
```diff
@@ -237,75 +237,76 @@ properties([buildDiscarder(logRotator(artifactDaysToKeepStr: '90',
 
 try{
     oe.emailJobStatus('STARTED')
+    def testing_stages = [
+        "Host verification 1604 Release":         { ACCHostVerificationTest('16.04', 'Release') },
+        "Host verification 1804 Release":         { ACCHostVerificationTest('18.04', 'Release') },
+        "ACC1804 GNU gcc SGX1FLC":                { ACCGNUTest() },
+        "RHEL-8 simulation clang-8 SGX1 Release": { RHEL8Test('clang', 'Release', ["OE_SIMULATION=1"], ["-DHAS_QUOTE_PROVIDER=OFF"]) },
```
BRMcLaren, line 244: This is a good implementation and improvement of the current implementation of passing multiple configurations via multiple String parameters. It, however, does change the standardization from what is currently implemented and can be confusing.

I am in favor of the switch, would it be possible to alter the ACCTest configuration to be configured in the same way?

BRMcLaren:
bors try

.jenkins/pipelines/Azure/Linux/Jenkinsfile
ionutbalutoiu: That's a good point. I believe that we can do this.

I'll look into it.

nonpolarity:
Thanks @ionutbalutoiu take care of RHEL!

.jenkins/pipelines/Azure/Linux/Jenkinsfile
BRMcLaren: Apart from this, LGTM.

ionutbalutoiu: I've made this change for `ACCTest` and the testing stages look cleaner now:
https://github.com/ionutbalutoiu/openenclave/blob/31ee859724c40a8a4ac9aac8cf7fc53ceda64900/.jenkins/pipelines/Azure/Linux/Jenkinsfile#L232-L269

I also updated the rest of the functions (`ACCContainerTest`, `ACCPackageTest`) from `.jenkins/pipelines/Azure/Linux/Jenkinsfile` to use the same approach: use a `List` parameter called `extra_cmake_args` instead of separate `String` parameters for each CMake flag.

The advantage of this approach is: whenever we need to add more CMake flags, we don't need to change the function implementation again.

Besides `.jenkins/pipelines/Azure/Linux/Jenkinsfile`, I also updated `.jenkins/pipelines/Agnostic/Linux/Jenkinsfile` with the same changes.

The remaining `.jenkins/pipelines/Azure/Windows/Jenkinsfile` is not updated yet, since it uses the function `WinCompilePackageTest` from the shared library that has the same old interface (multiple `String` parameters for the CMake flags).

If the above changes look good to you, we can merge this. 

I can follow up with a separate PR to update `.jenkins/pipelines/Azure/Windows/Jenkinsfile` as well, after changing the shared library method first (which is also used in the `.jenkins/pipelines/Azure/Nightly/Packaging.Jenkinsfile`). I want to be careful to not break master branch while doing it.

ionutbalutoiu: Waiting for a successful E2E right now.

ionutbalutoiu:
bors try

ionutbalutoiu:
bors try

ionutbalutoiu:
bors try

Issue #2724, opened 2020-03-23
Verify remote report from CLI


jumaffre:
In each Open Enclave install comes the `liboehostverify.a` library that can be linked to a binary if one needs to verify the remote report of an enclave (e.g. a client talking to a remote enclave running in ACC). 

However, this is impractical for clients that want to verify a remote report without having to build their own C++ binaries. For example, clients may communicate via off-the-shelf `curl` with a remote enclave running a HTTP server. 

I would be awesome if Open Enclave exported a binary to wrap the existing [`oe_verify_remote_report`](https://github.com/openenclave/openenclave/blob/master/include/openenclave/host_verify.h#L44) API so that clients can verify remote reports from the command line, e.g.:

```
$ oeverifyreport --report /path/to/remote/report
Report successfully verified
* OR *
Report verification failed: <OE_ERROR_CODE>
```

It is something that the CCF team has identified before (see https://github.com/microsoft/CCF/issues/683) but we strongly believe that contributing it directly to Open Enclave would also benefit other users of the SDK. 

We would be happy to help building this feature into Open Enclave if you're happy for us to do so. If yes, what would be the best place to build this feature (e.g. as a new [tool](https://github.com/openenclave/openenclave/tree/master/tools) or as an extension to `oecertdump` or `oesgx`)?

jhand2:
I agree, this seems like a useful tool to have. Thanks for the suggestion!

I'm not sure I know the right place to put it. Is liboehostverify sgx-specific? If not, we could probably make a new tool.

jazzybluesea:
How about creating general attestation CLI?  This cli would support generating evidence for any TEE and verify evidence + endorsements for any TEE?

jhand2:
The reason I like a tool which just uses oehostverify is that it can be run on any machine, not just SGX. This is good for supporting the case @jumaffre mentioned. I just got remote evidence from some party and want to verify it with a tool. A CLI that also supports full attestation would require a hardware TEE on the machine.

Both may be useful but I think they serve different use-cases.

achamayou:
It may be useful to have a tool that generates attestations as well, but as @jhand2 correctly points out this requires TEE hardware. The purpose here is to enable very lightweight verification for end users who talk to an in-enclave service, but don't necessarily have that capability.

I don't believe oehostverify is SGX-specific now, although it used to be (eg. #2264).

jazzybluesea:
[...]

Issue #2878, opened 2020-04-16
Some jenkins builds don't have links to the build logs for each pipeline


jhand2:
Sometimes when Jenkins runs a trying or staging build in the bors pipeline, it doesn't properly show links to the sub-pipelines that it ran.

This is what I see:

![image](https://user-images.githubusercontent.com/11011105/79479872-3eab0600-7fc2-11ea-8bee-78dd1ae36c02.png)

This is what I expect to see:

![image](https://user-images.githubusercontent.com/11011105/79479983-6306e280-7fc2-11ea-8150-745d7a17847e.png)

Note the "Triggered builds" section which links to the job/logs for that pipeline

oprinmarius:
bors try-

douglasmaciver-su:
I am working on writing this up.

BRMcLaren:
Hello @jhand2 

There was a proposed fix for this but it does not seem to work 100% of the time, so let's also provide a workaround.

The bug in question seems to be one of Blue Ocean itself, as can be seen here: https://issues.jenkins-ci.org/browse/JENKINS-51111

1. If one looks at a job such as below and it did not contain a triggered jobs section, take note of the pipeline desired.

![image](https://user-images.githubusercontent.com/54328827/79670332-ef5a0680-8176-11ea-88d5-8057f0cd5230.png)

"Starting building: Pipelines » Azure-Linux #72" in this case.
2. Go to: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/job/pipelines/ and into the pipeline that is failing.
![image](https://user-images.githubusercontent.com/54328827/79670394-7313f300-8177-11ea-8029-3fed04b420f8.png)


3, Burrow into the job number:

![image](https://user-images.githubusercontent.com/54328827/79670402-8626c300-8177-11ea-98e8-55bb05003e74.png)

4. Open Blue Ocean
![image](https://user-images.githubusercontent.com/54328827/79670436-e0278880-8177-11ea-8d64-1010ccc9e697.png)

5. View logs
![image](https://user-images.githubusercontent.com/54328827/79670444-f9303980-8177-11ea-9a56-2422a6d1b596.png)



TL:DR.

Agnostic-Linux : https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAgnostic-Linux/detail/Agnostic-Linux/<TheNumberNotedInStep1>/pipeline

[...]

PR #3014, opened 2020-05-11
Install host_verify binary 

jumaffre:
Resolves https://github.com/openenclave/openenclave/issues/2724

This PR installs the new `host_verify` CLI in the openenclave installation and `open-enclave-hostverify` .deb package, as per the discussion on #2724. 

Suggestions/improvements most welcome.


ryanhsu19:
LGTM

samples/host_verify/CMakeLists.txt
```diff
@@ -5,15 +5,21 @@ cmake_minimum_required(VERSION 3.11)
...
 target_include_directories(host_verify PRIVATE)
 
-target_link_libraries(host_verify openenclave::oehostverify)
+if (NOT BUILD_HOSTVERIFY_INLINE)
+  target_link_libraries(host_verify openenclave::oehostverify)
+else ()
+  target_link_libraries(host_verify oehostverify)
+endif ()
 
 add_custom_target(run DEPENDS host_verify)
```
jhand2, line 24: Only add this target if `NOT BUILD_HOST_VERIFY_INLINE`

jhand2:
bors try

jhand2:
Here are the failed build logs

Linux: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAgnostic-Linux/detail/Agnostic-Linux/350/pipeline

For now you can just build/install this for SGX (call this out in the readme)

Windows: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAzure-Windows/detail/Azure-Windows/351/pipeline

Not sure why this failed actually. Let me know if you have any thoughts, otherwise I can look a little closer

jumaffre:
> Here are the failed build logs
> 
> Linux: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAgnostic-Linux/detail/Agnostic-Linux/350/pipeline
> 
> For now you can just build/install this for SGX (call this out in the readme)
> 
> Windows: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAzure-Windows/detail/Azure-Windows/351/pipeline
> 
> Not sure why this failed actually. Let me know if you have any thoughts, otherwise I can look a little closer

@jhand2 I have just managed to reproduce the issue locally using the `oetools-full-16.04` Docker image. I will report my findings and hopefully provide a fix for it.

jumaffre:
@jhand2, @yentsanglee, @jazzybluesea  I've investigated this today. 

My understanding is that `sgx/hostverify_report.c` and `sgx/sgxquoteprovider.c` are only included in `liboehostverify.a` on SGX platforms. Because of this, `host_verify`, which makes use of `liboehostverify.a`, does not link successfully on non-SGX platforms.

AIUI, it should be possible to verify the quote of an enclave on any platform. At the moment, this is not possible as `oe_verify_remote_report()` always calls into `oe_initialize_quote_provider()`. Correct me if I'm wrong but if endorsements are passed into `oe_verify_remote_report()`, then there shouldn't be any need to use a quote provider at all. 

I think the long term solution is to `#ifdef` out the calls to the quote provider functions if the platform is not SGX. What are your thoughts?

jumaffre:
FYI, I have managed to build `host_verify` on for ARM by commenting some code and modifying `host/CMakeLists.txt`. This is only intended for demonstrating what the current issue is and will break other things.

jhand2:
bors try

jumaffre:
Only installing `host_verify` on SGX platforms for now.

jhand2:
@jumaffre see instructions in this issue for how to find logs when they don't show up: https://github.com/openenclave/openenclave/issues/2878

jumaffre:
@jhand2 It seems that the Windows build is still failing with

```
CMake Error at samples/CMakeLists.txt:61 (install):
  install TARGETS given target "host_verify" which does not exist in this
  directory.
```

Would you be able to advise on this?

The `Azure Linux` build error seems unrelated to my changes?

jhand2:
> install TARGETS given target "host_verify" which does not exist in this
>   directory.

@jumaffre It looks like it depends on the version of cmake. Older versions of cmake (before 3.13) can only install targets in the same directory: https://stackoverflow.com/questions/34443128/cmake-install-targets-in-subdirectories

Windows is probably using a different cmake version, so you may need to do the install from the subdirectory.

jhand2:
bors try

jumaffre:
@jhand2 The `fopen` issue on Windows should now be fixed.

jhand2:
bors try

achamayou:
bors try

achamayou:
@jhand2 is there a known issue with the CI?

jhand2:
Yes sorry, there was an issue with the Jenkins servers going down that Brett fixed yesterday. I'm not sure why `bors` didn't pick up your try though.

bors try

jhand2:
Oh, it's running, it's just stuck: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/Bors/activity

I'll see if we can look into it

jhand2:
bors try-

jhand2:
bors try

jhand2:
Thanks! bors r+

jhand2:
Not sure why that didn't work.

bors r+

samples/host_verify/README.md
```diff
@@ -66,4 +66,6 @@ When you see the following message displayed on the screen, it means you have su
 Usage:
   ./host_verify -r <report_file>
   ./host_verify -c <certificate_file>
```
dthaler, line 68: I still don't understand this sample.  Line 11 says it can be used to verify a remote report.  But line 18 looks like it will only work for a local enclave not a remote one.  How does one get a <report_file> or a <certificate_file> for a remote enclave?  maybe line 11 is wrong and this is just for local testing?

samples/host_verify/host.c
```diff
@@ -161,6 +165,15 @@ int main(int argc, const char* argv[])
             "Usage:\n  %s -r <report_file>\n  %s -c <certificate_file>\n",
             argv[0],
             argv[0]);
+        fprintf(
+            stdout,
+            "Verify the integrity of enclave remote report or attestation "
```
dthaler, line 170: See earlier comment, the word "remote" here in this line looks like it may be wrong, since I don't see any discussion of how to do anything remotely.  (See also the question I posed on issue #2724)

samples/host_verify/README.md
jumaffre: > maybe line 11 is wrong and this is just for local testing?

My understanding is that it is just for local testing. The utility can be used to verify any remote reports.

samples/host_verify/host.c
jumaffre: By "remote", this means from an enclave that is not necessarily collocated on the same CPU, e.g. described [here](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/CustomAttestation.md) as "remote attestation".

dthaler: Correct, but this tool only accepts a local file, nothing that can be obtained from a remote enclave, hence my point about it being wrong since this tool cannot (at least using any currently  documented process) be used for doing anything about "remote" attestation.

jhand2: In SGX "remote" and "local" attestation collateral have different formats. So the report file was generated on some enclave machine in the "remote" format. Then the file made it to this machine through some means (it may have even been generated on this machine, though I know that's confusing), and then this tool can verify it. A "remote QUOTE" is basically just a "local REPORT" that is signed in a special way.

Ultimately it would be preferable that this tool can be used to verify any type of attestation collateral by loading a verifier plugin, I just think we aren't there yet in terms of what the host_verify library can do.

dthaler: > In SGX "remote" and "local" attestation collateral have different formats. So the report file was generated on some SGX machine in the "remote" format. Then the file made it to this machine through some means (it may have even been generated on this machine, though I know that's confusing), and then this tool can verify it. A "remote QUOTE" is basically just a "local REPORT" that is signed in a special way.

The above information is what I am saying is missing from any documentation.  I would expect this sample to say how to actually use it.  Can you point me to any document that says "to use this sample, run something like oecert on the remote machine, then copy the file, then run this sample"?   I can't find it, hence I believe this issue is not closed.

jhand2: Good point. I agree #2724 should be left open to capture that documentation.

PR #3016, opened 2020-05-11
Set gsbase prior to entering the enclave in SGX Simulation mode

jhand2:
On SGX hardware, fs and gs will hold the same value upon enclave entry (due to the values OE sets in the tcs). Modify the simulation mode entry code to ensure this is also true in simulation mode.

In particular, this is to support enclave apps which make may need to make use of fs and gs (via wr(f/g)sbase instruction) and to allow OE to use gs if fs is in use by the enclave app.

### DIFF
--- a/host/sgx/enter.c
+++ b/host/sgx/enter.c
@@ -230,15 +230,16 @@ void oe_enter_sim(
     OE_ALIGNED(16)
     uint64_t fx_state[64];
 
-    // Backup host FS register. Enclave does not use the GS register.
+    // Backup host FS and GS registers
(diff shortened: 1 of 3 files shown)


host/sgx/enter.c
```diff
@@ -232,6 +232,7 @@ void oe_enter_sim(
 
     // Backup host FS register. Enclave does not use the GS register.
```
jhand2, line 233: Update comments

jhand2:
Failure is due to known bug in crypto test

bors r+

PR #3017, opened 2020-05-11
Provide mechanism for opting out of switchless calls

jhand2:
Allow enclave developers to pass a define when compiling their app (host and enclave) to opt out of the ocalls/ecalls used by the switchless feature.

If this define is passed, then sgx/switchless.edl does not need to be imported by SGX apps.

In theory, this same mechanism can be used for any optional features that are built into oecore/oehost.

I am hoping to get some preliminary feedback to see if folks agree that this is a good/workable opt-out mechanism for developers. If so, I will add more documentation for developers to use.


jhand2:
> I don't see any changes to documentation, so who is this for? If it's for developers writing apps that use the OE SDK, then we would need documentation that they can discover.

Good point, I should have been more clear in the PR description. I was hoping to get feedback on the actual mechanism before spending time on the docs. I have updated the description. I will definitely add documentation before this is merged.

dthaler:
https://stackoverflow.com/questions/2290587/gcc-style-weak-linking-in-visual-studio

jhand2:
@CodeMonkeyLeet @anakrish @dthaler Here's another option:  #3023

CodeMonkeyLeet:
> @CodeMonkeyLeet @anakrish @dthaler Here's another option: #3023

Yeah, per the SIG-arch discussion I would prefer the newer #3023 as it doesn't require simultaneous use of both an opt-in (inclusion of EDL) and an opt-out (use of build flag) at the same time. 

In terms of the generalizability of the model proposed here, I would also say that the use of build flags is a less-preferred option in general, and I would prefer componentization boundaries (e.g. refactor so that the choice is an inclusion/exclusion of a library, not a build/link time flag) where possible. 

Another future consideration is that elimination of code is less important on the host side (i.e. TCB minimization is primarily an enclave side security benefit) and asymmetric solutions may also be appropriate in the general case.

jhand2:
Closing in favor of  #3023

PR #3019, opened 2020-05-12
Update attestation reference and terminology in design doc

dthaler:
This PR just updates the IETF reference to be the latest doc and
terminology.  (Previously the design doc referenced an old non-IETF
document from an individual, and used terminology different from
what was accepted by the IETF.)


dthaler:
@jazzybluesea @yentsanglee @shnwc @shruti25ratnam pinging again...  It's been >2 weeks with no reviews.

PR #3023, opened 2020-05-12
Add weak stubs to allow switchless to be opt-out

jhand2:
If switchless.edl is not imported, oecore and oehost still try to call into the switchless ecalls and ocalls. Provide stubs which always fail when switchless.edl is not opted into.

Note that the while the `alternatename` linker flag is supported by MSVC, it is not documented.


include/openenclave/internal/defs.h
```diff
@@ -7,8 +7,21 @@
...
+#define OE_WEAK_ALIAS __pragma(comment(linker, "/alternatename:" #NEW "=" #OLD))
+#else
+#error OE_WEAK_ALIAS not implemented
+#endif
+
+/* OE_WEAK */
+#ifdef __GNUC__
+#define OE_WEAK __attribute__((weak))
+#else
+#define OE_WEAK
```
CodeMonkeyLeet, line 23: This should probably error out as in the `OE_WEAK_ALIAS` case if used in a compiler context that does not support it.

jhand2: Honestly I just couldn't figure out how to do it. `#error` will always error when compiler is msvc. Is there some mechanism we use elsewhere that will throw an error when a macro is used?

jhand2:
> Tangent: there was a change in #3017 for CMakeLists.txt to exclude building tests/exclude_system_edl when not `OE_SGX`, is that no longer applicable here? Not sure I follow.

Oh ya, I just forgot to include that here. This is because in order to exercise that this actually works, I had to remove the switchless.edl import, so now I can't use `platform.edl` anymore. This test will be removed by #2961 anyway.

If I can make it so all edl is opt-in by the 0.10 release then it's not a problem anymore anyway.

include/openenclave/internal/defs.h
```diff
@@ -7,8 +7,21 @@
 #include <openenclave/bits/defs.h>
 
 /* OE_WEAK_ALIAS */
+#ifdef __GNUC__
 #define OE_WEAK_ALIAS(OLD, NEW) \
     extern __typeof(OLD) NEW __attribute__((__weak__, alias(#OLD)))
+#elif _MSC_VER
+#define OE_WEAK_ALIAS __pragma(comment(linker, "/alternatename:" #NEW "=" #OLD))
```
jhand2, line 14: This should be `OE_WEAK_ALIAS(OLD, NEW)`. This is updated in my local copy but I forgot to push the change

include/openenclave/internal/defs.h
dthaler: https://stackoverflow.com/questions/24225006/warning-and-error-as-macro

dthaler: _STATIC_ASSERT(FALSE)
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/static-assert-macro?view=vs-2019
https://docs.microsoft.com/en-us/cpp/cpp/static-assert?view=vs-2019

jhand2:
bors try

include/openenclave/internal/defs.h
```diff
@@ -7,8 +7,22 @@
 #include <openenclave/bits/defs.h>
 
 /* OE_WEAK_ALIAS */
+#ifdef __GNUC__
 #define OE_WEAK_ALIAS(OLD, NEW) \
     extern __typeof(OLD) NEW __attribute__((__weak__, alias(#OLD)))
+#elif _MSC_VER
+#define OE_WEAK_ALIAS(OLD, NEW) \
+    __pragma(comment(linker, "/alternatename:" #NEW "=" #OLD))
```
anakrish, line 15: I cannot find any documentation about `alternatename`. So I am a bit hesitant about its use on the host side.
On the enclave side, this PR looks great!

Do we even need a weak directive?
Given two libraries `a.lib` and `b.lib` both containing a function `foo`, would the linker pick up the definition from the first library in the command line or would it complain of a multiple definition error?

jhand2: The linker will complain about multiple definitions if this is not included.

In theory you could declare stubs in a library and require that library to be linked in order to opt out, but that felt a bit backwards.

jhand2:
> However, can you double-check the need for weak-symbol?

See above comment. I do think it is possible but I think this is cleaner.

> Also consider scenario where the host loads two enclaves, where one has opted-in and the other has opted-out of switchless calls. I cannot see anything that would be broken, but please do confirm.

Actually this is interesting because all host-side ecall functions are defined to be weak (to support this exact scenario) here: https://github.com/openenclave/openenclave/pull/2961

In the case of two weak functions with the same name, the one linked first will be chosen (based on my testing with GCC). This is a problem because depending on the link order, it is possible that the stubs are chosen.

I'll have to think about this.

> It would also be good to opt-in to switchless rather than opt-out. In EDL, one opts in to switchless.

In the mechanism in the PR opt-out just means you didn't opt in. So this PR is consistent with this goal.

include/openenclave/internal/defs.h
anakrish: My understanding is that the user opts in to switchless rather than opting out.
Is that not the case?

anakrish: I tested out MSVC, the linker prefers the definition in the first library, and does not emit a multiple definition error. This is consistent with the Linux linker behavior. Note: OE support for pluggable allocators is based on this linker behavior.

From my understanding of this PR,
- the empty stubs live in oecore
- the actual definitions live in oe_switchless.a/.lib that the enclave author has to link against.
Do we need a separate library just for the stubs?

jhand2: It is the case. I'm saying if you define a stub and then also import it in edl you will get a multiple definition error which is why weak symbols are necessary.

jhand2: > This is consistent with the Linux linker behavior

That's not the behavior I'm seeing. Where do pluggable allocators define a function twice? In pluggable allocators there is only ever one definition of the function, you just chose which one to use at link time, right?

> From my understanding of this PR,
>    the empty stubs live in oecore
>   the actual definitions live in oe_switchless.a/.lib that the enclave author has to link against.

This only works because the oecore version is weak. If I make both symbols strong I get linker errors.

anakrish: Interesting. It is not clear why an edl import would change the linker behavior in the above case.

anakrish: @mwshan Did the Intel SDK have to rely on weak symbols? From my understanding, Intel SDK supports opting out of "system edls".

jhand2: > Interesting. It is not clear why an edl import would change the linker behavior in the above case.

I don't understand. Can you point me to a place where a function is defined twice and doesn't cause a linker error? It is my understanding this will always cause a linker failure (at least in clang/gcc)

anakrish: > Interesting. It is not clear why an edl import would change the linker behavior in the above case.
> 
> I don't understand. Can you point me to a place where a function is defined twice and doesn't cause a linker error? It is my understanding this will always cause a linker failure (at least in clang/gcc)

1) The stubs are defined in `oecore` library. The actual implementations are in `oeswitchless` library. There are two definitions; one is intended to be replaced by the other. The linker picks the one in `oeswitchless`.

2) Similarly, `oecore` has implementations of allocator functions `oe_allocator_malloc`, `oe_allocator_free` etc which is implemented using `dlmalloc`. The library `oesnmalloc` implements the same functions, but using snmalloc. The linker picks up the definitions in `oesnmalloc` providing the ability to customize allocators.

3) Here is the GCC's published way to replace `malloc`. https://www.gnu.org/software/libc/manual/html_node/Replacing-malloc.html
There are multiple definitions of `malloc` - one provided by glibc and the one written by the user. The linker picks the one provided by the user, replacing GCC's malloc.

2) and 3) seem quite similar to 1). I am trying to understand what the difference is.

jhand2: 3 works because globe defines these functions as weak symbols. Search for weak_alias here: https://github.com/lattera/glibc/blob/master/malloc/malloc.c

2 works because oedlmalloc and oesnmalloc aren’t both linked at the same time. It’s always one or the other.

I don’t believe 1 works without weak symbols.

anakrish: > This is consistent with the Linux linker behavior
> 
> That's not the behavior I'm seeing. Where do pluggable allocators define a function twice? In pluggable allocators there is only ever one definition of the function, you just chose which one to use at link time, right?
> 

There are **two** definitions, in two different libraries. One of them is chosen at link time.

> From my understanding of this PR,
> the empty stubs live in oecore
> the actual definitions live in oe_switchless.a/.lib that the enclave author has to link against.
> 
> This only works because the oecore version is weak. If I make both symbols strong I get linker errors.

Are the stub functions in a separate file? One reason I can think of that might cause a multiple definition error is if the stubs are placed in a file containing other functions that the linker has pulled in.

jhand2: > Are the stub functions in a separate file? One reason I can think of that might cause a multiple definition error is if the stubs are placed in a file containing other functions that the linker has pulled in.

After some more investigation this does seem to be what is happening.

But since the edl generated functions are defined as weak (to avoid conflicting with each other in the event that stubs for two enclaves are linked into the same host), the stub defined in oehost must also be weak. If it is strong and the others are weak, it will replace them.

jhand2:
> > Also consider scenario where the host loads two enclaves, where one has opted-in and the other has opted-out of switchless calls. I cannot see anything that would be broken, but please do confirm.
> 
> Actually this is interesting because all host-side ecall functions are defined to be weak (to support this exact scenario) here: #2961
> 
> In the case of two weak functions with the same name, the one linked first will be chosen (based on my testing with GCC). That is, weak symbols don't override other weak symbols. This is a problem because depending on the link order, it is possible that the stubs are chosen.

This is addressed by the following fact about linking in gcc and msvc:

If a function is defined in both a linked object file and a static library, the version in the object file is used and the library version is ignored.

jhand2:
> In any case, /alternatename is an implementation detail that can be replaced at a later date without affecting the user experience in anyway.

I agree. I'm gonna go ahead and merge so that I can make progress on some other PRs but once everything lands, I'm happy to come back to this and see if there is a better way we can do it.

bors r+

jhand2:
Oh whoops, I forgot there's a build failure I need to address

bors r-

PR #3024, opened 2020-05-12
Remove OS-specific logic given that OE currently supports only Linux/ELF enclaves

mingweishih:
OE currently supports only SGX enclaves with Linux/ELF form. Therefore, this PR removes OS-specific logic in the enclave core library.

The changes are as follows.
- Remove source in enclave/core/sgx/windows
- Move sources from enclave/core/sgx/linux to enclave/core/sgx
- Remove the checks of the `__linux__`


mingweishih:
bors try

PR #3026, opened 2020-05-13
Update signing doc with OP-TEE info

dthaler:
Addresses issue #2714

### DIFF
--- a/docs/GettingStartedDocs/buildandsign.md
+++ b/docs/GettingStartedDocs/buildandsign.md
@@ -136,8 +136,8 @@ Debug=0
 ## Signing an OP-TEE Enclave
 
 The `oesign` tool is currently SGX-only.  For OP-TEE enclaves, signing
-is instead performed by the `sign.py` script that comes with OP-TEE, and the
-script uses a key that is integrated into OP-TEE itself.  The signing key
+is instead performed by the `sign.py` script that comes with OP-TEE.
+The signing key
 and script, among other artifacts, are exported to a "TA Dev Kit" during
 OP-TEE's build process. The Open Enclave SDK takes a `OE_TA_DEV_KIT_DIR`
 CMake parameter at
--- a/docs/GettingStartedDocs/buildandsign.md
+++ b/docs/GettingStartedDocs/buildandsign.md
@@ -139,13 +139,14 @@ The `oesign` tool is currently SGX-only.  For OP-TEE enclaves, signing
 is instead performed by the `sign.py` script that comes with OP-TEE, and the
 script uses a key that is integrated into OP-TEE itself.  The signing key
 and script, among other artifacts, are exported to a "TA Dev Kit" during
-OP-TEE's build process. The Open Enclave SDK takes a CMake parameter at
+OP-TEE's build process. The Open Enclave SDK takes a `OE_TA_DEV_KIT_DIR`
+CMake parameter at
 build time that specifies where to find the TA Dev Kit. This enables
 the `add_enclave` CMake function to locate sign.py and the TA signing key
 and apply them automatically to every TA built as part of the build process.
 
 There is currently no equivalent of a .conf file for OP-TEE, so all settings
-must be done using the `OE_SET_ENCLAVE_OPTEE` macro:
+must be specified using the `OE_SET_ENCLAVE_OPTEE` macro:
 
 ```c
 #define TA_UUID                                            \
@@ -158,8 +159,8 @@ must be done using the `OE_SET_ENCLAVE_OPTEE` macro:
 
 OE_SET_ENCLAVE_OPTEE(
     TA_UUID,               /* UUID */
-    2 * 1024 * 1024,       /* Heap size, in bytes */
-    24 * 1024,             /* Stack size, in bytes */
+    4 * 1024 * 1024,       /* Heap size, in bytes */
+    4 * 1024,              /* Stack size, in bytes */
     TA_FLAG_MULTI_SESSION, /* Flags */
     "1.0.0",               /* Version */
     "Sample enclave");     /* Description */
--- a/docs/GettingStartedDocs/buildandsign.md
+++ b/docs/GettingStartedDocs/buildandsign.md
@@ -122,9 +122,6 @@ Any properties set in the code also serve as default values when the enclave is
 signed using oesign, so the signing `CONFFILE` only needs to specify override
 parameters during signing.
 
-Using the `OE_SET_ENCLAVE_SGX` macro in code is not recommended and it
-may be removed in a future release.
-
 **Since enclaves that run in debug mode are not confidential, you should disable
 the ability to run the enclave in debug mode before deploying it into production.**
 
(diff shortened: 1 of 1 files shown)


docs/GettingStartedDocs/buildandsign.md
```diff
@@ -133,3 +135,50 @@ specify a `CONFFILE` that only changed that property:
...
+The flags property can contain any of the flags defined in the
+`include/user_ta_header.h` file in the TA Dev Kit:
+
+| Flag                        | Meaning                                     |
+| :-------------------------- | :------------------------------------------ |
+| TA_FLAG_SINGLE_INSTANCE     | all host apps use the same enclave instance |
+| TA_FLAG_MULTI_SESSION       | allow multiple sessions from host apps      |
+| TA_FLAG_INSTANCE_KEEP_ALIVE | keep enclave running after sessions end     |
+| TA_FLAG_SECURE_DATA_PATH    | accesses SDP memory                         |
+| TA_FLAG_REMAP_SUPPORT       | use map/unmap syscalls                      |
```
CodeMonkeyLeet, line 180: `TA_FLAG_REMAP_SUPPORT` is marked as deprecated in [user_ta_header.h](https://github.com/OP-TEE/optee_os/blob/master/lib/libutee/include/user_ta_header.h), should it still be mentioned here? 

@HernanGatta Similarly, everywhere this macro is used in OE SDK today, the flag `TA_FLAG_EXEC_DDR` is specified , which may be confusing since that is also marked as deprecated already (and may be confusing since it's not documented in this table).

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -123,6 +122,9 @@ Any properties set in the code also serve as default values when the enclave is
 signed using oesign, so the signing `CONFFILE` only needs to specify override
 parameters during signing.
 
+Using the `OE_SET_ENCLAVE_SGX` macro in code is not recommended and it
```
CodeMonkeyLeet, line 125: I would suggest removing this line since it's just muddies the waters. Until we have a firm deprecation plan for this feature, it just creates FUD, and when we do have a deprecation plan, we would update the docs around the associated scenarios around this macro (e.g. what is the new recommended way to easily run unsigned debug SGX enclaves?) It's also possible that we will retain backwards compatibility for this, even though it will become the less preferred option.

Right now it just reads like a block of "why are you telling me all this and then also telling me not to use the feature, and then also that I _have_ to use the equivalent for OP-TEE?"

docs/GettingStartedDocs/buildandsign.md
dthaler: I had noticed the DDR issue when generating the PR and omitted that.  Yes I think it should be removed from all of our code (which isn't a doc issue so didn't do it in this PR).   I hadn't noticed that REMAP_SUPPORT was also deprecated, I'll remove that from this PR accordingly.  Thanks for pointing this out.

docs/GettingStartedDocs/buildandsign.md
dthaler: Done

docs/GettingStartedDocs/buildandsign.md
dthaler: Done

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -133,3 +132,49 @@ specify a `CONFFILE` that only changed that property:
...
 Debug=0
 ```
+
+## Signing an OP-TEE Enclave
+
+The `oesign` tool is currently SGX-only.  For OP-TEE enclaves, signing
+is instead performed by the `sign.py` script that comes with OP-TEE, and the
+script uses a key that is integrated into OP-TEE itself.  The signing key
+and script, among other artifacts, are exported to a "TA Dev Kit" during
+OP-TEE's build process. The Open Enclave SDK takes a CMake parameter at
```
HernanGatta, line 142: If you'd like to make this sentence more precise, the CMAKE parameter is `OE_TA_DEV_KIT_DIR`.

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -133,3 +132,49 @@ specify a `CONFFILE` that only changed that property:
...
+build time that specifies where to find the TA Dev Kit. This enables
+the `add_enclave` CMake function to locate sign.py and the TA signing key
+and apply them automatically to every TA built as part of the build process.
+
+There is currently no equivalent of a .conf file for OP-TEE, so all settings
+must be done using the `OE_SET_ENCLAVE_OPTEE` macro:
```
HernanGatta, line 148: "so all settings must be _specified_"?

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -133,3 +132,49 @@ specify a `CONFFILE` that only changed that property:
...
+        0x126830b9, 0xeb9f, 0x412a,                        \
+        {                                                  \
+            0x89, 0xa7, 0xbc, 0xc8, 0xa5, 0x17, 0xc1, 0x2e \
+        }                                                  \
+    }
+
+OE_SET_ENCLAVE_OPTEE(
+    TA_UUID,               /* UUID */
+    2 * 1024 * 1024,       /* Heap size, in bytes */
+    24 * 1024,             /* Stack size, in bytes */
```
HernanGatta, line 162: Perhaps these values could be made to match those in the example for SGX, for mere consistency to show how to express the same configuration using both macros:

```
OE_SET_ENCLAVE_OPTEE(
    TA_UUID,               /* UUID */
    4 * 1024 * 1024,       /* Heap size, in bytes */
    4 * 1024 * 1024,             /* Stack size, in bytes */
...
```

Alternately, the SGX ones could be made to reflect the OP-TEE ones seeing as OP-TEE has stricter memory boundaries in practice.

dthaler: Done

docs/GettingStartedDocs/buildandsign.md
dthaler: Done

docs/GettingStartedDocs/buildandsign.md
dthaler: Done

docs/GettingStartedDocs/buildandsign.md
HernanGatta: I agree we should remove `TA_FLAG_EXEC_DDR` from the existing code; it used to be a mandatory flag.

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -133,3 +132,50 @@ specify a `CONFFILE` that only changed that property:
 # Enclave settings:
 Debug=0
 ```
+
+## Signing an OP-TEE Enclave
+
+The `oesign` tool is currently SGX-only.  For OP-TEE enclaves, signing
+is instead performed by the `sign.py` script that comes with OP-TEE, and the
+script uses a key that is integrated into OP-TEE itself.  The signing key
```
HernanGatta, line 140: Technically, the script doesn't automatically use the key in the TA Dev Kit. When building the SDK from source, it is the CMake `add_enclave` macro that picks up the key from the right place. If users wish to call `sign.py` manually, which they currently have to if consuming the SDK in binary form*, they must still pass the path to the key.

We still don't have official binary releases of the SDK for OP-TEE, though. So perhaps for this document, you can remove the words between the comma and the period:

For OP-TEE enclaves, signing is instead performed by the `sign.py` script that comes with OP-TEE. The signing key and script [...]

dthaler: Done

dthaler:
@CodeMonkeyLeet need one more approver, are you good with this now?   I am addressing the DDR flag removal in code in a separate forthcoming PR.

dthaler:
bors try

PR #3028, opened 2020-05-13
Prevent conflicts on the host from the same EDL file being imported

jhand2:
OE supports the EDL stubs for two different enclaves being linked into the same host binary. To prevent collisions, mark all host functions as weak symbols. Additionally, add include guards around generated argument structs.

### DIFF
--- a/include/openenclave/edger8r/host.h
+++ b/include/openenclave/edger8r/host.h
@@ -106,6 +106,26 @@ OE_INLINE size_t oe_wcslen(const wchar_t* s)
     return wcslen(s);
 }
 
+#if __GNUC__
+#define EDGER8R_WEAK_ALIAS(OLD, NEW) \
+    extern __typeof(OLD) NEW __attribute__((__weak__, alias(#OLD)))
+#elif _MSC_VER
+#define EDGER8R_WEAK_ALIAS(OLD, NEW) \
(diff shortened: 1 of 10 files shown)


BRMcLaren:
bors try

jhand2:
> Edit: Ignore me, I am checking something in the CI.

No problem. But no promises this actually compiles on all platforms

jhand2:
bors try

jhand2:
> This looks good to me.
> Do you anticipate more changes in edger8r?
> This one will not be that difficult for me to incorporate into the python/c++ edger8rs.

No, this is the last edger8r change I have planned for this feature

jhand2:
Feel free to ignore this PR until I address this build failures. It might take me a few days to get back to it.

jhand2:
Local test failure

bors r-

BRMcLaren:
bors try

jhand2:
It turns out this is not sufficient to fix the problem. ecalls are made by ID so the ID that is sent to the enclave may not match what the enclave expects. @mingweishih has some thoughts on how we can change this with per-EDL function tables to avoid conflicts. I'll leave this PR open for now but this likely needs more investigation.

mingweishih:
Closing, see #3086 for more detail.

PR #3034, opened 2020-05-13
Disable the child_process test

jhand2:
Disable the child_process test. Version 1.33 of the SGX DCAP driver introduces support accessing an enclave from a forked child process. This causes this test to fail when using that driver. Version 1.22 does not have fork support so this test will pass.

### DIFF
--- a/tests/child_process/host/host.cpp
+++ b/tests/child_process/host/host.cpp
@@ -73,7 +73,12 @@ int main(int argc, const char* argv[])
                 process_result = close(fpipe[0]);
                 OE_TEST(result == OE_OK);
                 oe_terminate_enclave(enclave);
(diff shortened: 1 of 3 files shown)


tests/child_process/host/host.cpp
```diff
@@ -73,7 +73,13 @@ int main(int argc, const char* argv[])
                 process_result = close(fpipe[0]);
                 OE_TEST(result == OE_OK);
                 oe_terminate_enclave(enclave);
-                wait(NULL);
+                j
+
+                    // Wait for child
+                    int child_status = -1;
+                while (wait(&child_status) > 0)
```
anakrish, line 80: The formatting seems off. Tab?

PR #3037, opened 2020-05-14
Reserve the beginning of the thread data on SGX

jhand2:
For the SGX runtime, reserve the beginning of the oe_thread_data_t structure in oecore to allow other libc implementations to be built on top of OE. oelibc and oelibcxx use this space, but an application
should be able to replace those implementations if they choose.


anakrish:
> The change looks fine to me. However, I'm just wondering why we put exception-related information into `thread_specific_data` before this PR. Would like to see @anakrish comments if there is any implication on the debugging or something else.

@mingweishih SGX exceptions happen on an enclave thread, and that's why the information is housed in a thread-specific data-structure like td_t.

PR #3038, opened 2020-05-14
Remove host library oehostapp.

### LINKED ISSUE #2595: oehostapp and oehost are potentially redundant
tl;dr We should remove either oehost or oehostapp cmake targets and just recommend one

Note, this would require a deprecation warning in v0.9 release

From @CodeMonkeyLeet in #2593:

> we currently expect everyone to link against oehostapp given the sample usage anyway, and so we ostensibly already recommend -rdynamic to be the default use case. We still make oehost available, presumably to provide an option without -rdynamic (and prior to this change, without noexecstack).
>
> Do you think providing a library that doesn't pull in -rdynamic in CMake by default is valuable? My understanding of -rdynamic is that it pulls in all symbols which is useful for backtracing and dynamic library load/link scenarios and is the preferable default. I assume this expands the size of the resulting binary; are there other potentially undesirable effects? Assuming we only have the -rdynamic default, is there an easy way to override that in a user's build? (e.g. I couldn't find a no-dynamic option).
>
> @jhand2 Yes, it would be a breaking change to remove one or the other library, but it's one of those things that's better to change sooner rather than later, and we would still want to mark deprecation in this release, then eliminate in the next one.

nonpolarity:
Due to oehostapp's redundancy to oehost, oehostapp is removed.
The appendent compiling option "-rdynamic" is also removed. To provide dynsym of oehost for dynamic loaded libraries, users could add this option back manually.

Fixes #2595

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Mark APIs in include/openenclave/attestation/sgx/attester.h and verifier.h as experimental.
 
+### Removed
+- Removed oehostapp and the appendent "-rdynamic" compiling option. Please use oehost instead and add the option back manually if necessary.
+
 [0.9.0][v0.9.0_log]
 ------------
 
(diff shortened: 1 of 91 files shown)


nonpolarity:
bors try

nonpolarity:
> LGTM! Can you also update the changelog?

Thanks! updated.

PR #3039, opened 2020-05-15
Fix compile errors with experimental USE_SNMALLOC

anakrish:
Compile errors were due to incorrect order of options in command-line.
Also dlmalloc specific test is not enabled for snmalloc.


anakrish:
bors try

anakrish:
> LGTM
> 
> One thought: now that snmalloc is built in a separate library, it would be possible to build the libs for both dlmalloc and snmalloc in CI and then just choose one at link time.
> 
> Not sure what that'll do to compile times though

There will be no noticeable effect on compile times (both dlmalloc and snmalloc are compiled as one source file each). I like the idea of compiling both. I will enable that with subsequent snmallocrelated PRs. Hopefully at the end of this release/start of next, we can revisit the choice of default allocator.

PR #3040, opened 2020-05-15
Fix libcxxrt test result on Windows.

nonpolarity:
The test result is OS-specific that log file has different line endings on Windows and other OS. Git always adjusts the line endings automatically. Thus this test has a prerequisite that git must execute on the same platform as on which platform the test runs. For example, git on cygwin changes line endings to lf and test on Windows expects crlf, then test will fail.
This change make the libcxxrt test stateless.


nonpolarity:
bors try

nonpolarity:
bors try

BRMcLaren:
bors try

PR #3042, opened 2020-05-15
Use platform's OCaml to build edger8r if available.

anakrish:
This restores the original way in which oeedger8r was built
on both platforms, if ocaml compiler is already available.
If not, it lets esy build the ocaml compiler and then build
oeedger8r.

Note: ocaml needs to be added back to prereqs on both platforms
to ensure a pain free/fast build of oeedger8r.

### DIFF
--- a/tools/oeedger8r/CMakeLists.txt
+++ b/tools/oeedger8r/CMakeLists.txt
@@ -1,6 +1,76 @@
 # Copyright (c) Open Enclave SDK contributors.
 # Licensed under the MIT License.
 
+find_program(OCAMLOPT ocamlopt)
+
+if (OCAMLOPT)
+  set(BINARY ${CMAKE_CURRENT_BINARY_DIR}/oeedger8r${CMAKE_EXECUTABLE_SUFFIX})
+  add_custom_command(
+    OUTPUT ${BINARY}
+    # Copy sources to build folder.
+    COMMAND cmake -E copy ${CMAKE_CURRENT_SOURCE_DIR}/intel/*
+            ${CMAKE_CURRENT_BINARY_DIR}
+    COMMAND cmake -E copy ${CMAKE_CURRENT_SOURCE_DIR}/src/*
+            ${CMAKE_CURRENT_BINARY_DIR}
+    # Generate lexer and parser.
+    COMMAND ocamllex Lexer.mll
+    COMMAND ocamlyacc Parser.mly
+    # Generate Intel library.
+    # Use -w -3 to suppress
+    # warning 3: deprecated: String.lowercase
+    COMMAND
+      ocamlopt -c -bin-annot -for-pack Intel -w -3 Ast.ml Util.ml SimpleStack.ml
+      Plugin.ml Preprocessor.ml Parser.mli Parser.ml Lexer.ml CodeGen.ml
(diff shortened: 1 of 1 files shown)


anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
@mingweishih @jhand2 Can you review this PR?

tools/oeedger8r/CMakeLists.txt
```diff
@@ -1,6 +1,76 @@
...
+    # Generate lexer and parser.
+    COMMAND ocamllex Lexer.mll
+    COMMAND ocamlyacc Parser.mly
+    # Generate Intel library.
+    # Use -w -3 to suppress
+    # warning 3: deprecated: String.lowercase
+    COMMAND
+      ocamlopt -c -bin-annot -for-pack Intel -w -3 Ast.ml Util.ml SimpleStack.ml
+      Plugin.ml Preprocessor.ml Parser.mli Parser.ml Lexer.ml CodeGen.ml
+    COMMAND ocamlopt -pack -o Intel.cmx Ast.cmx Util.cmx SimpleStack.cmx
```
jhand2, line 24: Can you split the source files into list variables? i.e. "INTEL_SOURCES" and "OE_SOURCES". Then you can use those in compilation and DEPENDS.

mingweishih: I second this. However, currently this command depends on the files and compiles their copies in `CMAKE_CURRENT_BINARY_DIR`. The prefix differences need to be handled if using list.

tools/oeedger8r/CMakeLists.txt
```diff
@@ -1,6 +1,76 @@
...
+    # Additionally suppress
+    # Warning 40: input_files was selected from type Intel.Util.edger8r_params.
+    # It is not visible in the current scope, and will not
+    # be selected if the type becomes unknown.
+    COMMAND
+      ocamlopt -bin-annot str.cmxa unix.cmxa Intel.cmx -w -3-40 Common.mli
+      Common.ml Headers.mli Headers.ml Sources.mli Sources.ml Emitter.ml main.ml
+      -o ${BINARY}
+    DEPENDS src/dune
+            src/Common.mli
```
mingweishih, line 36: Shouldn't this be `${CMAKE_CURRENT_SOURCE_DIR}/src/Common.mli` (same for the following lines)?

tools/oeedger8r/CMakeLists.txt
anakrish: I tried to create a list but it did not work out due to the reason @mingweishih points out.
- The sources have to be copied to bin dir; otherwise ocamlopt will pollute the source folder by generating artifacts there.
- Some sources (Parser.ml, Parser.mli, Lexer.ml) are generated on the fly from Parser.ml, Lexer.ml
- The DEPENDS declaration must have relative paths to the original source locations - they must have the intel/src prefix.

All this makes using a list of files a bit difficult.

tools/oeedger8r/CMakeLists.txt
anakrish: cmake by default will prefix relative paths with the current source directory if a file exists; if a valid file does not exist, then it will prefix it with current binary directory. I feel omitting the ${CMAKE_CURRENT_SOURCE_DIR} prefix is less verbose.

PR #3045, opened 2020-05-18
Handle case where a thread may miss the notify_all call.

anakrish:
When expected number of OUT_OF_THREADS are reached, the condition
variable is signalled via a notify_all to wake up all waiting threads.
However, not all threads may not yet have gone on to wait.
Use a thread-safe variable g_notify_called to handle this scenario.
Threads won't go on to wait if notify_all has already been called.


radhikaj:
Thanks @anakrish

anakrish:
bors try

anakrish:
bors single on

PR #3020, opened 2020-05-12
add 2 oesign cases

johnvenn:
add 2 new cases for oesign
1. sign the enclave with empty config
2. sign the enclave with config file containing negative values.

oesign-customize-cfg.py was a python script to generate cfg with the above purposes, and also,
in future, add more cases on this test point.

PR #3046, opened 2020-05-18
2 oesign cases: 1. sign w/ empty config 2. sign with config containing neg vals

johnvenn:
obselete pull request 3020, the same commit of 3020.


mingweishih:
bors try

johnvenn:
> Approve again given that I already did in the #3020
> 
> Just side note that instead of creating a new PR, the git issue could have been fixed in the #3020. However, to save time, I'm fine with this case.

yes, just to save time, I created a new PR. Thanks a lot for your comment.

BRMcLaren:
bors try

PR #3050, opened 2020-05-18
Refactoring code toward self-contained liboecore

mingweishih:
This PR refactors the code related to libcore with the goal of making it self-contained as follows.
- Remove the dependencies on libc from dlmalloc (as part of libcore).
- Remove the files that are part of of libcore from oelibc.

### DIFF
--- a/3rdparty/dlmalloc/CMakeLists.txt
+++ b/3rdparty/dlmalloc/CMakeLists.txt
@@ -3,16 +3,18 @@
 
 add_enclave_library(oedlmalloc OBJECT allocator.c)
 
-enclave_link_libraries(oedlmalloc PRIVATE oe_includes oelibc_includes)
+enclave_link_libraries(oedlmalloc PRIVATE oe_includes)
 
 if (OE_TRUSTZONE)
-    enclave_link_libraries(oedlmalloc PUBLIC oelibutee_includes)
-    set(TEE_C_FLAGS ${OE_TZ_TA_C_FLAGS})
-else()
-    set(TEE_C_FLAGS "")
-endif()
+  enclave_link_libraries(oedlmalloc PUBLIC oelibutee_includes)
+  set(TEE_C_FLAGS ${OE_TZ_TA_C_FLAGS})
+else ()
+  set(TEE_C_FLAGS "")
+endif ()
 
-enclave_compile_options(oedlmalloc PRIVATE
+enclave_compile_options(
+  oedlmalloc
+  PRIVATE
   -ftls-model=local-exec
   -nostdinc
   -fPIE
@@ -20,10 +22,17 @@ enclave_compile_options(oedlmalloc PRIVATE
   -fvisibility=hidden
   ${TEE_C_FLAGS})
 
+enclave_include_directories(oedlmalloc PRIVATE
+                            ${PROJECT_SOURCE_DIR}/include/openenclave/corelibc)
+
+# Allow dlmallc to find required stdc functions from corelibc.
+enclave_compile_definitions(oedlmalloc PRIVATE OE_NEED_STDC_NAMES)
+
 # Specify the warning options as source files properties so that
 # they will appear last in the compiler command line and supercede
 # other warning options.
-set_source_files_properties(allocator.c PROPERTIES
-  COMPILE_FLAGS "-Wno-conversion -Wno-null-pointer-arithmetic")
+set_source_files_properties(
+  allocator.c PROPERTIES COMPILE_FLAGS
+                         "-Wno-conversion -Wno-null-pointer-arithmetic")
 
 maybe_build_using_clangw(oedlmalloc)
--- a/3rdparty/dlmalloc/allocator.c
+++ b/3rdparty/dlmalloc/allocator.c
@@ -9,10 +9,7 @@
 #define LACKS_SYS_TYPES_H
 #define LACKS_TIME_H
 #define MORECORE dlmalloc_sbrk
-#define ABORT oe_abort()
 #define USE_DL_PREFIX
-#define LACKS_STDLIB_H
-#define LACKS_STRING_H
 #define USE_LOCKS 1
 #define fprintf _dlmalloc_stats_fprintf
 #define NO_MALLOC_STATS 1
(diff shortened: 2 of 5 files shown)


mingweishih:
bors try

mingweishih:
bors try

anakrish:
@mingweishih Can you please wait for https://github.com/openenclave/openenclave/pull/3039 to go in?

mingweishih:
> @mingweishih Can you please wait for #3039 to go in?

Sure. Feel free to cancel the try.

mingweishih:
bors try

3rdparty/dlmalloc/allocator.c
```diff
@@ -21,15 +18,6 @@
 #pragma GCC diagnostic ignored "-Wparentheses-equality"
 #endif
 
-OE_INLINE
-int sched_yield(void)
```
jhand2, line 25: Why was this removed? Is it not needed anymore?

mingweishih: It's not needed anymore because `corelibc/sched.h` defines it (previously dlmalloc did not use corelibc headers at all).

3rdparty/dlmalloc/CMakeLists.txt
```diff
@@ -3,27 +3,36 @@
...
+  PRIVATE
   -ftls-model=local-exec
   -nostdinc
   -fPIE
   -ffreestanding
   -fvisibility=hidden
   ${TEE_C_FLAGS})
 
+enclave_include_directories(oedlmalloc PRIVATE
+                            ${PROJECT_SOURCE_DIR}/include/openenclave/corelibc)
```
anakrish, line 26: What is the benefit of including corelibc with OE_NEED_STDC_NAMES rather than including oelibc_includes?

anakrish:
@mingweishih Can you also make sure to keep this PR consistent with https://github.com/openenclave/openenclave/pull/3039?

That PR contains compilation fixes for snmalloc.
snmalloc requires libcxx and I don't think libcxx can work with corelibc; it needs oelibc.
If we use corelibc for dlmalloc, we are making dlmalloc _special_ in some way.

Ideally, users need not need to know oe internals like corelibc to make an allocator work with oe.
It would be good for both dlmalloc and snmalloc, not to depend on oe internals.

3rdparty/dlmalloc/CMakeLists.txt
mingweishih: This allows oecore to be self-contained. Similar to we build memset/memmove/memcpy in oecore using corelibc with OE_NEED_STDC_NAMES.

anakrish: As far a I am aware, oelibc and corelibc with stdc names provide only function declarations right? 
If you compile allocator.c with oelibc and with corelibc would the generated object change?
Do you see any difference in the .o file?

mingweishih: Without the changes of the PR, the oecore has a dependency on oelibc. So, the question is more about whether we want to make oecore depend on oelibc, or make it self-contained. I feel the former is kind of conflicting with the reason for corelibc.

anakrish: > This allows oecore to be self-contained. Similar to we build memset/memmove/memcpy in oecore using corelibc with OE_NEED_STDC_NAMES.

These are 4 functions defined as part of a `freestanding` environment. Even `oecore` cannot use a `oe_` prefix for these funcitions.

https://github.com/openenclave/openenclave/blob/7bc2e60523313e360b557bf30fccd293f05563c1/include/openenclave/corelibc/string.h#L25-L29

anakrish: I am fine with these changes. I view an allocator as something that is not tightly coupled with oecore. An allocator can be written in C++ (like snmalloc, tcmalloc, jemalloc) and still be made to work within enclaves. I don't think the requirement `an allocator ought to use corelibc (oecore)`, though seems pure, is feasible in practice considering the allocators that are in existence.

mingweishih: Yes, this is something I was looking into as well (and confuses me a little bit). The current behavior is that `oecore` directly includes the sources of these four functions, and `oelibc` includes them as well (without this PR). Now when an enclave app use any of these functions, the app can link to either the object in `oecore` or `oelibc` (depend on their order during the link-time). This is kind of weird to me so in this PR I remove the duplicates from `oelibc`.

We can also go the other direction; i.e., make `oecore` depend on `oelibc` and remove the duplicates from `oecore`. But again, this makes me wonder why we need corelibc if we go for this direction.

mingweishih: I also agree that we don't need to make this as a hard requirement to any plugged allocator. But I think it makes sense for the default allocator (dlmalloc) to use corelibc instead of libc.

anakrish: Those 4 functions are special. GCC expects a `freestanding` environment to provide those functions; even when compiling the kernel GCC will still emit calls to those functions.

I'd like for us to be able to change the default allocator; say tomorrow we could choose to go with snmalloc or tcmalloc as the default allocator - couple of teams have already found the our default allocator (dlmalloc) to be non performant.
Thats why, I feel the default allocator itself ought to be compiled as if any other allocator is compiled. So that we may easily replace it at a future date.

Having said all that, I am fine with the changes in this PR. But it is important to be cognizant of the implications in regards to pluggable allocators, marshalling strategies etc.

mingweishih: Thanks for the inputs. I'll hold off this PR before having more clear idea of the benefits/high-level goals.

BRMcLaren:
bors try

3rdparty/dlmalloc/CMakeLists.txt
```diff
@@ -3,27 +3,36 @@
...
+
+# Allow dlmallc to find required stdc functions from corelibc.
```
dthaler, line 28:
```suggestion
# Allow dlmalloc to find required stdc functions from corelibc.
```

mingweishih:
Closing for now, will re-visit if necessary.

Issue #3044, opened 2020-05-17
install_windows_prereqs1.ps1:  Visual Studio Build Tools installation fails cryptically


anakrish:
On an ACC Win Server 2019 machine, the following failure is seem when installing Visual Studio 2019:
Installing C:\Users\radhikaj\AppData\Local\Temp\2\packages\vs_buildtools.exe with
-ArgumentList:
-q
--wait
--norestart
--add Microsoft.VisualStudio.Workload.VCTools
--add Microsoft.VisualStudio.Component.VC.CMake.Project
--add Microsoft.VisualStudio.Component.Windows10SDK.17134
--add Microsoft.VisualStudio.Component.VC.v141.ARM.Spectre
--add Microsoft.VisualStudio.Component.VC.v141.ARM64.Spectre
--includeRecommended
-FilePath:
C:\Users\radhikaj\AppData\Local\Temp\2\packages\vs_buildtools.exe
-PassThru:
True
-Wait:
True
Failed to install: C:\Users\radhikaj\AppData\Local\Temp\2\packages\vs_buildtools.exe
at Install-Tool, C:\openenclave\scripts\install-windows-prereqs.ps1: line 314
at Install-VisualStudio, C:\openenclave\scripts\install-windows-prereqs.ps1: line 448
at <ScriptBlock>, C:\openenclave\scripts\install-windows-prereqs.ps1: line 705
at <ScriptBlock>, <No file>: line 1

Issue #3043, opened 2020-05-17
install-windows-prereqs.ps1: Installation of Python adds partial path to system path


anakrish:
Running `./scripts/install-windows-prereqs.ps1 -InstallPath C:/oe_prereqs -LaunchConfiguration SGX1FLC -DCAPClientType Azure1` causes `C:\Program Files\python-3.7.4\Scr` to be added to system path.

Additionally it raises the following warnings, which indicates that the `install-windows-prereqs.ps1` may be corrupting the PATH.

SUCCESS: Specified value was saved.
[2020-05-17T17:17:45.3769286+00:00] Start-ExecuteWithRetry attempt 0
  **WARNING: The script wheel.exe is installed in '**C:\Program Files\python-3.7.4\Scripts**' which is not on PATH.**
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.

**WARNING: **The data being saved is truncated to 1024 characters**.**
[2020-05-17T17:17:56.3939980+00:00] Start-ExecuteWithRetry terminated
Collecting pip<19.2
  Using cached https://files.pythonhosted.org/packages/5c/e0/be401c003291b56efc55aeba6a80ab790d3d4cece2778288d65323009420/pip-19.1.1-py2.py3-none-any.whl
Collecting setuptools
  Using cached https://files.pythonhosted.org/packages/2c/c5/0d38afb961f83e0d51f319f7dc166195ebabc1ea3cb20a10a77f500f7156/setuptools-46.4.0-py3-none-any.whl
Collecting wheel
  Using cached https://files.pythonhosted.org/packages/8c/23/848298cccf8e40f5bbb59009b32848a4c38f4e7f3364297ab3c3e2e2cd14/wheel-0.34.2-py2.py3-none-any.whl
Installing collected packages: pip, setuptools, wheel
Successfully installed pip-19.1.1 setuptools-46.4.0 wheel-0.34.2

SUCCESS: Specified value was saved.
[2020-05-17T17:17:56.4096225+00:00] Start-ExecuteWithRetry attempt 0
**WARNING: **You are using pip version 19.1.1, however version 20.1 is available**.**
You should consider upgrading via the 'python -m pip install --upgrade pip' command.
[2020-05-17T17:17:58.0943130+00:00] Start-ExecuteWithRetry terminated
Collecting cmake_format
[...]

PR #3053, opened 2020-05-19
Run tests on fresh Linux environments

rs--:
Install system dependencies on Linux (RHEL 8, Ubuntu 16.04/18.04) before building, and testing OE. Windows will be added at a later date.


rs--:
bors try

rs--:
bors try

jhand2:
For posterity, can you add to the PR description/commit message what you mean by end-to-end tests (that is, explain that it also tests installation of packages)

anakrish:
@rs-- @BRMcLaren Are we doing e2e testing on every job/PR?

rs--:
@anakrish Yeah, the plan is to run it on a few configurations for every PR

anakrish:
Do you think it would be better to run e2e tests as periodically say twice a day instead of on every PR?

The reason I ask is that, the scripts change infrequently compared to code and the chance of a typical PR breaking the scripts is minimal, though not zero. The URLs that the script depends on however change occasionally, breaking the powershell script. So it seems periodic jobs would help identify a remote url change.

Some of the issues the scripts have occur due to reuse in an existing system:
E.g: #3043 #3044

Is there another strategy for catching those errors?

BRMcLaren:
> Do you think it would be better to run e2e tests as periodically say twice a day instead of on every PR?

FWIW there are e2e tests running in the background that are not reported up to the repo as badges. It's not 100% stable yet, but for the [curious](https://oe-jenkins-tf.westeurope.cloudapp.azure.com/job/CI-CD_Infrastructure/job/OpenEnclave-Rebuild-Infrastructure/)

Regarding your suggestion, I don't believe that is the [issue](https://github.com/openenclave/openenclave/issues/2993) @rs--  is trying to solve. Part of the current process of integration when an individual modifies scripts is sending them through a rebuild of the above Rebuild Infrastructure job... twice, which takes 16 or so hours and someone with Jenkins permissions to assist. It would be better for developers if there was an automated way for them to contribute to the project, add packages if they need to, fix broken URLs, or bugs to encourage independence.

I would add that having a subset of the testing configuration run on new machines to test the full Contributor experience seems nice and would guard against regressions. It would also allow developers to try new things and have them caught during integration time, things such as if someone wanted to use a new package manager instead of the ps1 script, the framework this PR implements would support making that extendable.

BRMcLaren:
> Is there another strategy for catching those errors?

This is an interesting proposition, maybe an e2e test that pulls from old machine images/definitions and runs as part of the maintenance testing instead of per PR?

anakrish:
> > Do you think it would be better to run e2e tests as periodically say twice a day instead of on every PR?
> 
> FWIW there are e2e tests running in the background that are not reported up to the repo as badges. It's not 100% stable yet, but for the [curious](https://oe-jenkins-tf.westeurope.cloudapp.azure.com/job/CI-CD_Infrastructure/job/OpenEnclave-Rebuild-Infrastructure/)
> 
That is great! Hopefully that will catch OpenSSL/Intel urls changing ahead of time.

anakrish:
>  It would be better for developers if there was an automated way for them to contribute to the project, add packages if they need to, fix broken URLs, or bugs to encourage independence.

This is good. I am planning to remove node/esy and add native ocaml install to both the windows and linux scripts. So it will be nice any mistakes are caught beforehand.

anakrish:
> > Is there another strategy for catching those errors?
> 
> This is an interesting proposition, maybe an e2e test that pulls from old machine images/definitions and runs as part of the maintenance testing instead of per PR?

That would work. That kind of test does not have to run every PR. The script for example does lots of things like uninstall old-driver, install new one, install visual studio etc.
I personally don't see value in testing all these scenarios on each PR - very few PRs have changes that will affect the script. However these are scenarios that ought to be locked down nonetheless.

.jenkins/pipelines/Azure/Linux/Jenkinsfile
```diff
@@ -16,14 +16,21 @@ AGENTS_LABELS = [
     "ubuntu-nonsgx":    env.UBUNTU_NONSGX_CUSTOM_LABEL ?: "nonSGX",
     "windows-nonsgx":   env.WINDOWS_NONSGX_CUSTOM_LABEL ?: "nonSGX-Windows"
 ]
+VANILLA_AGENTS_LABELS = [
```
BRMcLaren, line 19: I am wondering if it is best to have two different structures VANILLA and the other, as opposed to a single lookup map.

rs--: An option would be to have it in the same map, but just use `acc-rhel-8-vanilla`?

rs--:
bors try

.jenkins/pipelines/Azure/Linux/Jenkinsfile
BRMcLaren: Yes, this is what I was thinking.

BRMcLaren:
@rs--  Whenever we add complicated improvements like this, it's good to see stability. I have approved the change, but it would be good to see having 5 or so successful runs in a row to ensure this is not introducing too many failure points.

rs--:
bors try

rs--:
bors try

rs--:
bors try

rs--:
bors try

BRMcLaren:
bors try

BRMcLaren:
bors try

rs--:
bors try

BRMcLaren:
bors try

PR #3061, opened 2020-05-20
update project child process due the change of driver

jinghe-INTC:
verified with DCAP driver v1.33

### DIFF
--- a/tests/child_process/host/host.cpp
+++ b/tests/child_process/host/host.cpp
@@ -58,8 +58,8 @@ int main(int argc, const char* argv[])
                 fprintf(stdout, "child pid = %d\n", getpid());
                 uint32_t magic = 2;
                 result_child_process = get_magic_ecall(enclave, &magic);
-                OE_TEST(result_child_process != OE_OK);
-                OE_TEST(magic == 2);
+                OE_TEST(result_child_process == OE_OK);
+                OE_TEST(magic == 0x1234);
                 process_result =
                     write(fpipe[1], &result_child_process, sizeof(oe_result_t));
                 process_result = close(fpipe[1]);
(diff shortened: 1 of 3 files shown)


jinghe-INTC:
bors try

jhand2:
Failure log here: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAzure-Linux/detail/Azure-Linux/441/pipeline/

tests/child_process/host/host.cpp
```diff
@@ -90,7 +92,8 @@ int main(int argc, const char* argv[])
             if (pid == 0) // child process
             {
                 result_child_process = oe_terminate_enclave(enclave);
-                OE_TEST(result_child_process == OE_INVALID_PARAMETER);
+                OE_TEST(result_child_process == OE_OK);
```
jhand2, line 95: Given that the enclave is terminated in the parent and the child, I wonder if there is a race condition there. An OE enclave may not be able to be destroyed from both the parent and the child, I'm not sure.

That could be the cause of the build failure.

jinghe-INTC: I verified the case on VM or FLC platform and didn't see any error with destroying.
And from the bors log, it shows the error is at Test failed at child process status check: 
> ../tests/child_process/host/host.cpp(113): main child_status == 0

So I guess it needs an "_exit(EXIT_SUCCESS);"  for the child process. I have added it and will start a bors try.

jinghe-INTC:
bors try

tests/child_process/host/host.cpp
jhand2: oh, ya, that makes sense to me. Thanks!

PR #3062, opened 2020-05-20
SGX: Decouple the td structure from fs register on enter and exit

jhand2:
If the Linux kernel is configured to allow it, an enclave application can change the fs register during execution. Currently if an application does this, it will break assumptions OE makes about where the `oe_sgx_td_t` structure (stores thread specific information) is stored.

Decouple the exit and enter code to be decoupled from the fs register. This allows OE to cleanly change where it stores the td struct for certain applications.

In particular, this is to support a library OS such as SGX-LKL which uses modifies the fs register upon usermode thread context switches.

I know this touches some critical code paths and assembly is error prone so I am not in a rush to merge this, we can take our time discussing/validating.

### DIFF
--- a/enclave/core/sgx/enter.S
+++ b/enclave/core/sgx/enter.S
@@ -230,12 +230,7 @@ oe_enter:
 
 .clear_enclave_registers:
 
-    // Clear these so information will not be leaked to host
-    // Retain %r11 which is used to reference the td structure. It will be
-    // cleared immediately before EEXIT.
-    mov %r11, OM_ENC_TD
     oe_cleanup_registers
-    mov OM_ENC_TD, %r11
 
 .restore_host_registers:
 
--- a/enclave/core/sgx/enter.S
+++ b/enclave/core/sgx/enter.S
@@ -233,7 +233,6 @@ oe_enter:
     // Clear these so information will not be leaked to host
     // Retain %r11 which is used to reference the td structure. It will be
     // cleared immediately before EEXIT.
-    push %r11
     mov %r11, OM_ENC_TD
     oe_cleanup_registers
     mov OM_ENC_TD, %r11
--- a/enclave/core/sgx/enter.S
+++ b/enclave/core/sgx/enter.S
@@ -39,20 +39,30 @@
 oe_enter:
 .cfi_startproc
 
+.get_td:
+
+    // Get the location of the td_t structure for this thread. This value is
+    // expected to be present in %r11 for the remainder of oe_enter.
+    //
+    // Upon first entry to the enclave, td->base.self in the td_t structure
+    // is not yet initialized. However, the loader in host/sgx/create.c places
+    // the td_t structure as a specific offset from TCS.
+    lea TD_FROM_TCS(%rbx), %r11
+
 .save_host_registers:
     // Backup the current host rbp, rsp, and context to previous.
-    mov %fs:td_host_rbp, %r8
-    mov %r8, %fs:td_host_previous_rbp
-    mov %fs:td_host_rsp, %r8
-    mov %r8, %fs:td_host_previous_rsp
-    mov %fs:td_host_ecall_context, %r8
-    mov %r8, %fs:td_host_previous_ecall_context
+    mov td_host_rbp(%r11), %r8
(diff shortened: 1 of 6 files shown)


jhand2:
bors try

jhand2:
Fixed build issue

bors try

BRMcLaren:
bors try

enclave/core/sgx/exit.S
```diff
@@ -100,6 +119,9 @@ oe_asm_exit:
     // simulate-flag-check.
     lfence
 
+    // Clear %r9 which was being used to maintain td pointer
+    xor %r11, %r11
```
anakrish, line 123: This needs to be `%r9' I think. Can you please thoroughly recheck the register usage?

jhand2: Fixed this one and another mistake I found in enter.S. I'll go through again today to make sure everything looks right.

enclave/core/sgx/exit.S
```diff
@@ -51,11 +51,26 @@
...
 
+.get_td:
+
+    // Save argument registers
+    push %rdi
+    push %rsi
+
+    // Get a pointer to the td structure. This value is expected to be present
+    // in %r9 for the remainder of oe_asm_exit
+    call oe_sgx_get_td
```
anakrish, line 62: instead of this call we could pass td as a parameter to oe_asm_exit

enclave/core/sgx/exit.S
```diff
@@ -74,24 +89,28 @@ oe_asm_exit:
...
-    movq $0, %fs:td_last_sp
+    movq $0, td_last_sp(%r9)
 
 .clear_enclave_registers:
 
-    // Clear these so information will not be leaked to host
+    // Clear registers so information will not be leaked to host
+    // Retain %r9 which is used to reference the td structure. It will be
+    // cleared immediately before EEXIT.
+    push %r9
```
anakrish, line 99: This seems like a workaround for oe_cleanup_register's current behavior. That function can be modified instead.

jhand2: Good point. I could use r11 in both enter and exit and then just change the cleanup code to not clear r11.

jhand2: Done

enclave/core/sgx/exit.S
jhand2: Done

enclave/core/sgx/enter.S
```diff
@@ -160,21 +170,25 @@ oe_enter:
...
 
     // Save the host return address to enclave stack.
-    mov %fs:td_host_rcx, %r8
+    mov td_host_rcx(%r11), %r8
     mov %r8, OM_HOST_RETURN_ADDR
 
+    // Save reference to the td structure to enclave stack.
+    mov %r11, OM_ENC_TD
+
     // Call __oe_handle_main(ARG1=RDI, ARG2=RSI, CSSA=RDX, TCS=RCX, OUTPUTARG1=R8, OUTPUTARG2=R9)
```
jhand2, line 192: @anakrish doesn't `__oe_handle_main` never return? (Instead it calls oe_asm_exit eventually). I am wondering why this function has any code after calling `__oe_handle_main`. Am I misunderstanding something?

If this is the case, we could just clear all the registers and call `ud2` just as a precaution.

anakrish: `__oe_handle_main` will return when the ecall completes. `oe_asm_exit` is used for ocalls.
There is some code that is the same between the two and I might have a branch that tries to avoid the duplication.

enclave/core/sgx/exit.S
```diff
@@ -100,6 +104,9 @@ oe_asm_exit:
     // simulate-flag-check.
     lfence
 
+    // Clear %r9 which was being used to maintain td pointer
```
nonpolarity, line 107: I think you have tried to match comment with code, right?

jhand2: Oh good catch, I forgot to change this. Thanks

jhand2:
Issue with git submodule update

bors r+

jhand2:
more infra issues?

bors r+

PR #3063, opened 2020-05-20
Update snmalloc to latest and add test.

anakrish:
The test is a basic test to ensure that snmalloc is compiled
successfully as part of a build of the OE SDK.

This test will be extended to make sure that only snmalloc symbols
are present in the enclave.

Also make sure that the fragment tests which are dlmalloc specific become noops when compiled with USE_SNMALLOC


anakrish:
bors try

anakrish:
@jhand2 @mingweishih @dthaler @achamayou @CodeMonkeyLeet @nonpolarity @HernanGatta 

This PR requires `snmalloc` which is a submodule to checked out so that the snmalloc library and test can be built successfully. This is the first time the OE SDK will have a hard requirement on a submodule.

Given that submodules are a git preferred way to avoid duplicating 3rdparty code, it is likely that we will see more submodules in the future (eg: coverage library, oedebugrt etc).

Should we:
- require the contributor to run a `git submodule update --init --recursive` after syncing the repo
- or have the cmake build process automatically fetch the submodules
?

I have seen both the approaches being followed by different project.
Do you have a preference for either?

jhand2:
> Should we:
>    require the contributor to run a git submodule update --init --recursive after syncing the repo
>    or have the cmake build process automatically fetch the submodules

My vote is for #1 since it gives the developer a little more flexibility in what they have checked out in a submodule. But I don't feel strongly either way.

BRMcLaren:
bors try

mingweishih:
> > Should we:
> > (1) require the contributor to run a git submodule update --init --recursive after syncing the repo
> > (2) or have the cmake build process automatically fetch the submodules
> 
> My vote is for (1) since it gives the developer a little more flexibility in what they have checked out in a submodule. But I don't feel strongly either way.

In addition to (1), we can tie the submodule update to the git pull via hooks (e.g., [post-merge](https://git-scm.com/docs/githooks#_post_merge)). This ensures that when developer tries to synchronize their repo against master, they will also update the submodules.

achamayou:
@anakrish in the spirit of the principle of least surprise, I would prefer option 1. than to have CMake modify the source checkout (users can also do `git clone --recursive` to get everything in one go)

The flip side of that choice is that users do tend to forget to check out submodules, so we had this in our CMake when we used them:

```
function(check_submodule_not_empty path)
  file(GLOB submodule_files "${path}/*")
  list(LENGTH submodule_files number)
  if(${number} EQUAL 0)
    message(FATAL_ERROR "Submodule ${path} is empty. You can initialise submodules now with 'git submodule update --recursive --init', or during 'git clone' by passing '--recursive'.")
  endif()
endfunction()

# Give an early, helpful message when submodules are missing
check_submodule_not_empty(${CCF_DIR}/3rdparty/evercrypt-msr)
```

anakrish:
Thanks @jhand2 @mingweishih @achamayou I will go with option 1.
Additionally I will also add the submodule check that Amaury suggested. 

But first I need to fix ARM compile errors :)

tests/memory/enc/fragment.c
```diff
@@ -29,7 +29,16 @@ struct mallinfo
     size_t keepcost; /* releasable (via malloc_trim) space */
 };
 
-struct mallinfo dlmallinfo(void);
+static bool _dlmalloc_available = true;
+
+// Default implementation of dlmallinfo inddicates that dlmalloc is not
```
dthaler, line 34:
```suggestion
// Default implementation of dlmallinfo indicates that dlmalloc is not
```

tests/snmalloc/enc/enc.c
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/internal/tests.h>
+#include <stdlib.h>
+
+#include "snmalloc_t.h"
+
+// snmalloc requires atleast 4K heap pages.
```
dthaler, line 9:
```suggestion
// snmalloc requires at least 4K heap pages.
```

jhand2:
> > > Should we:
> > > (1) require the contributor to run a git submodule update --init --recursive after syncing the repo
> > > (2) or have the cmake build process automatically fetch the submodules
> > 
> > 
> > My vote is for (1) since it gives the developer a little more flexibility in what they have checked out in a submodule. But I don't feel strongly either way.
> 
> In addition to (1), we can tie the submodule update to the git pull via hooks (e.g., [post-merge](https://git-scm.com/docs/githooks#_post_merge)). This ensures that when developer tries to synchronize their repo against master, they will also update the submodules.

That's a great idea, I vote for that one now. Thanks!

tests/memory/enc/fragment.c
anakrish: Fixed.

tests/snmalloc/enc/enc.c
anakrish: Fixed.

anakrish:
bors try

anakrish:
bors try

enclave/core/sgx/threadlocal.c
```diff
@@ -332,6 +332,24 @@ oe_result_t oe_thread_local_init(oe_sgx_td_t* td)
             _thread_locals_relocated = true;
         }
 
+        {
+            // Currently this is the most appropriate spot to initialize the
```
CodeMonkeyLeet, line 336: This doesn't explain why it's a problem for `oe_allocator_init()` to be called during `_handle_init_enclave()` after `td_init` though. Consider:

```
// oe_allocator_init needs to be invoked once globally before oe_allocator_thread_init
// can be called. Since oe_allocator_thread_init needs to be invoked on each td_init,
// we call oe_allocator_init here with a run-once lock instead of during _handle_init_enclave,
// which is tool late since that only gets to run after td_init.
```
Also, why not just simplify this by calling it through `oe_once()`?

anakrish:
> LGTM. Is there going to be a second update revising the documentation around removing the experimental status of the feature?

Yes. This PR and the earlier one (#2967) do most of the heavy-lifting. All that is remaining once this goes in is to publish allocator.h (currently it is internal) and to create a sample.

enclave/core/sgx/threadlocal.c
anakrish: Updated comment. Nice suggestion about using oe_once. Done.

anakrish:
bors try

enclave/link.c
```diff
@@ -21,6 +22,10 @@ const void* oe_link_enclave(void)
         oe_verify_report_ecall,
         oe_get_public_key_by_policy_ecall,
         oe_get_public_key_ecall,
+        // Speciy oe_allocator_malloc so that there is a direct link from
```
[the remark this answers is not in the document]
anakrish: Fixed.

3rdparty/snmalloc/CMakeLists.txt
```diff
@@ -1,7 +1,12 @@
 # Copyright (c) Open Enclave SDK contributors.
 # Licensed under the MIT License.
 
+check_submodule_not_empty("${PROJECT_SOURCE_DIR}/3rdparty/snmalloc/snmalloc")
 add_enclave_library(oesnmalloc OBJECT allocator.cpp)
+# Add explicit dependency to musl_includes and libcxx includes.
```
mjp41, line 6: If you add the following, then it should pull in the required compiler settings from snmalloc:
```Cmake
# Include snmalloc library
set(SNMALLOC_ONLY_HEADER_LIBRARY ON)
add_subdirectory(snmalloc EXCLUDE_FROM_ALL)
target_link_libraries(oesnmalloc PRIVATE snmalloc_lib)
```
This has the advantage that if we alter our dependencies then it should get updated here automatically.  However, as you are building in a constrained setting, we may specify things that don't work here.

3rdparty/snmalloc/CMakeLists.txt
```diff
@@ -13,12 +18,15 @@ endif ()
 enclave_compile_options(
   oesnmalloc
   PRIVATE
+  -I${CMAKE_CURRENT_SOURCE_DIR}/snmalloc/src
```
mjp41, line 21: This line would go away with the above change.

3rdparty/snmalloc/CMakeLists.txt
```diff
@@ -1,7 +1,12 @@
 # Copyright (c) Open Enclave SDK contributors.
 # Licensed under the MIT License.
 
+check_submodule_not_empty("${PROJECT_SOURCE_DIR}/3rdparty/snmalloc/snmalloc")
 add_enclave_library(oesnmalloc OBJECT allocator.cpp)
+# Add explicit dependency to musl_includes and libcxx includes.
+# Specifying a dependency on oelibc_includes does not seem to work reliably
+# on all platforms.
+add_dependencies(oesnmalloc libcxx_includes musl_includes)
```
mjp41, line 10: If you add the following, then it should pull in the required compiler settings from snmalloc:
```suggestion

# Include snmalloc library
set(SNMALLOC_ONLY_HEADER_LIBRARY ON)
add_subdirectory(snmalloc EXCLUDE_FROM_ALL)
target_link_libraries(oesnmalloc PRIVATE snmalloc_lib)

```
This has the advantage that if we alter our dependencies then it should get updated here automatically.  However, as you are building in a constrained setting, we may specify things that don't work here.  Possible `target_link_library` should be `enclave_link_library` but I don't know you specifics well enough.

3rdparty/snmalloc/CMakeLists.txt
```diff
@@ -13,12 +18,15 @@ endif ()
...
   oesnmalloc
   PRIVATE
+  -I${CMAKE_CURRENT_SOURCE_DIR}/snmalloc/src
   -I${PROJECT_SOURCE_DIR}/include
-  -I${PROJECT_SOURCE_DIR}/3rdparty/libcxx/libcxx/include
+  # Explicitly specify include folders in correct order.
+  -I${LIBCXX_INCLUDES}
   -I${PROJECT_BINARY_DIR}/output/include/openenclave/libc
   -Wno-conversion
   -ftls-model=local-exec
```
mjp41, line 27: This line would go away with the above change.
```suggestion
```

3rdparty/snmalloc/allocator.cpp
```diff
@@ -29,16 +30,10 @@ void* ThreadAllocUntyped::get()
 // see pal_open_enclave.h for details
 #include "./snmalloc/src/override/malloc.cc"
```
mjp41, line 31:
```suggestion
#include "override/malloc.cc2
```
It is now in the include path.

anakrish:
bors try

anakrish:
bors try

anakrish:
Added file missed out in previous commit.
bors try

anakrish:
bors try

anakrish:
Yet another unrelated failure.

3rdparty/snmalloc/CMakeLists.txt
anakrish: Thanks for the suggestion. I have retained the current way since we are building in a constrained setting and would like to have precise control over the way snmalloc is built as part of the SDK.

3rdparty/snmalloc/allocator.cpp
anakrish: I will try to see if I can use malloc.cc2 in another PR. I don't think it is hight priority.

PR #2887, opened 2020-04-17
add switchless test for one TCS

yanxue-intel:
Signed-off-by: yan xue <yan.xue@intel.com>

PR #3065, opened 2020-05-21
Switchless one tcs test from Intel and accompanying code fix

anakrish:
Allow all the regular ecalls to first finish during enclave
initialization. Then launch ecall worker threads and wait
for the threads to enter the enclave and thus acquire a
dedicated tcs.

This allows creating enclaves where all ecalls after initialization
are switchless.

Intel test from #2887

### DIFF
--- a/host/sgx/switchless.c
+++ b/host/sgx/switchless.c
@@ -208,7 +208,20 @@ oe_result_t oe_start_switchless_manager(
         }
     }
 
+    // Inform the enclave about the switchless manager through an ECALL
+    if (num_host_workers > 0)
+    {
+        OE_CHECK(oe_sgx_init_context_switchless_ecall(
+            enclave,
+            &result_out,
+            manager->host_worker_contexts,
+            manager->num_host_workers));
+        OE_CHECK(result_out);
+    }
+
     // Start the enclave worker threads, and assign each one a private context.
+    // ecall worker threads are initialized after the regular ecall above to
+    // oe_sgx_init_context_switchless_ecall is complete.
     for (size_t i = 0; i < num_enclave_workers; i++)
     {
         OE_TRACE_INFO("Creating switchless enclave worker thread %d\n", (int)i);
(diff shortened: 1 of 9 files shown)


anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

host/sgx/switchless.c
```diff
@@ -222,22 +235,21 @@ oe_result_t oe_start_switchless_manager(
         {
             OE_RAISE(OE_THREAD_CREATE_ERROR);
         }
+
+        // Wait until the enclave worker thread has started.
+        // If so, spin_cout/total_spin_count will be non zero.
```
jhand2, line 240: spelling: spin_count

jhand2, line 240: I would change this to "spin_cout and total_spin_count". For I while I was confused because I thought it was saying "spin_count divided by total_spin_count"

tests/switchless_one_tcs/host/host.c
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    volatile status_e* p_switchless_func_status = &switchless_func_status;
+    while (*p_switchless_func_status != FUNC_WORKING)
+    {
+#if defined(__linux__)
+        usleep(100);
+#elif defined(_WIN32)
+        Sleep(0);
+#endif
```
jhand2, line 80: nit: if you compile this as C++ you can use `std::this_thread::yield` to make it OS independent. Not a huge deal though.

host/sgx/switchless.c
anakrish: Fixed.

anakrish: :) Fixed.

tests/switchless_one_tcs/host/host.c
anakrish: I'll leave this as it is for now. It's be useful to have a library of utilities for testing that provides a `yield` primitive so that we don't have to change languages just to use a function.

PR #3067, opened 2020-05-22
Update to use version 1.5.0 of Azure DCAP

deschuma:
Update OE SDK to use the latest version of Azure DCAP which includes changes to:
- Cache additional collateral on the client (not just certificates)
- Add environment variable to enable debug logging

Tested on Ubuntu 16, 18 and Windows Server 2019


anakrish:
@deschuma Is there no source code update needed on the Linux side ansible scripts?

deschuma:
@anakrish last time this was all that was needed -unless something has changed, I thought Linux side just picked up the latest version

jhand2:
@anakrish we don't specify a version for the package in ansible. The same is true for the libsgx packages. So whenever the apt repo gets updated, we automatically start using it.

Fixing the package versions is an option, but if users do an `apt upgrade` they will get the newest package and rolling back can sometimes be a pain depending on what other packages were also updated on the system.

anakrish:
@jhand2 @deschuma Then its fine.

@yentsanglee @soccerGB Can you please review/approve this PR since this is attestation related?

CodeMonkeyLeet:
@BRMcLaren Does this change need to be tested against the CI/CD infrastructure and the package cache updated?

PR #3076, opened 2020-05-26
Adding oe_buffer_too_small checks for all collateral buffers 

### LINKED ISSUE #2968: oe_get_quote_verification_collateral_ocall() may return success if buffer is too small
In host/sgx/ocalls.c function oe_get_quote_verification_collateral_ocall(), function may return success if all but the last buffer size are too small. This function should return as soon as _copy_output_buffer() encounters buffer too small error.

shruti25ratnam:
This change is to address the issue: oe_get_quote_verification_collateral_ocall() may return success if buffer is too small

Fixes #2968


jhand2:
Build failure was due to infra issue (docker login)

bors r+

PR #3077, opened 2020-05-26
Update ansible script to use Azrure dcap client version 1.5 for RHEL.

nonpolarity:
Update ansible script to use Azrure dcap client version 1.5 for RHEL, since the latest version 1.5 is released:

- Added caching of all collateral
- Added ability to enable debug logging independently of logging callback

### DIFF
--- a/scripts/ansible/roles/linux/az-dcap-client/vars/redhat/main.yml
+++ b/scripts/ansible/roles/linux/az-dcap-client/vars/redhat/main.yml
@@ -8,7 +8,7 @@ yum_packages:
   - "pkg-config"
 
 git_repo_url: "https://github.com/Microsoft/Azure-DCAP-Client.git"
-git_repo_branch: "1.4.0"
+git_repo_branch: "1.5"
 git_local_dir: "/tmp/Azure-DCAP-Client"
 
 validation_distribution_files:


nonpolarity:
bors try

anakrish:
This PR seems to contradict the following comment:
https://github.com/openenclave/openenclave/pull/3067#issuecomment-632919701
> we don't specify a version for the package in ansible. The same is true for the libsgx packages. So whenever the apt repo gets updated, we automatically start using it.

Do ansible scripts require versions or do we always pick the latest versions?

jhand2:
> This PR seems to contradict the following comment:
> [#3067 (comment)](https://github.com/openenclave/openenclave/pull/3067#issuecomment-632919701)
> 
> > we don't specify a version for the package in ansible. The same is true for the libsgx packages. So whenever the apt repo gets updated, we automatically start using it.
> 
> Do ansible scripts require versions or do we always pick the latest versions?

>This PR seems to contradict the following comment:
> https://github.com/openenclave/openenclave/pull/3067#issuecomment-632919701
> > we don't specify a version for the package in ansible. The same is true for the libsgx packages. So whenever the apt repo gets updated, we automatically start using it.

> Do ansible scripts require versions or do we always pick the latest versions?

What I said only applies to the ubuntu packages. The azure dcap client doesn't publish an RPM packages, so we have to go through some extra steps to pull the source and build it from the github releases. In that case you do have to specify which version you want to pull.

nonpolarity:
> 
> 
> > This PR seems to contradict the following comment:
> > [#3067 (comment)](https://github.com/openenclave/openenclave/pull/3067#issuecomment-632919701)
> > > we don't specify a version for the package in ansible. The same is true for the libsgx packages. So whenever the apt repo gets updated, we automatically start using it.
> > 
> > 
> > Do ansible scripts require versions or do we always pick the latest versions?
> 
> What I said only applies to the ubuntu packages. The azure dcap client doesn't publish an RPM package, so we have to go through some extra steps to pull the source and build it from the github releases. In that case you do have to specify which version you want to pull.

Yes on RHEL the situation is as @jhand2 described. The good thing is Azure dcap client owner are trying to release a rpm for RHEL. This pr is also used to verify the new version is good on RHEL.

PR #3080, opened 2020-05-28
Fix crypto crl test time diffing

### LINKED ISSUE #2617: tests/crypto tests fail transiently: oe_result_t=OE_VERIFY_REVOKED crl_tests.c_test_get_dates last.minutes == _time.minutes
The CRL test in tests/crypto fails (very rarely) because a cert generated on a minute boundary may have a different "cert issued" time and "file created" time.

We should create a tolerance function for minute/hour/day/month/second boundaries so that as long as the times are within a reasonable interval (30 sec?) the test will still pass.

This issue causes failures in CI which should be as reliable as possible.

jhand2:
The time diffing in the crypto CRL test is error prone to transient changes in the timestamp when the test was run. Modify the test to use a tolerance function to figure out the diff between two datetimes.

Fixes #2617

### DIFF
--- a/tests/crypto/crl_tests.c
+++ b/tests/crypto/crl_tests.c
@@ -127,6 +127,19 @@ static bool _year_is_a_leap_year(uint32_t year)
     }
 }
 
+static int32_t _diff_oe_datetime(oe_datetime_t* t1, oe_datetime_t* t2)
+{
+    int32_t dyear = (int32_t)t1->year - (int32_t)t2->year;
+    int32_t dmonth = (int32_t)t1->month - (int32_t)t2->month;
+    int32_t dday = (int32_t)t1->day - (int32_t)t2->day;
+    int32_t dhour = (int32_t)t1->hours - (int32_t)t2->hours;
+    int32_t dmin = (int32_t)t1->minutes - (int32_t)t2->minutes;
+    int32_t dsec = (int32_t)t1->seconds - (int32_t)t2->seconds;
+
+    return (dyear * 31556952) + (dmonth * 2592000) + (dday * 86400) +
+           (dhour * 3600) + (dmin * 60) + (dsec);
+}
+
 static void _test_get_dates(void)
 {
     printf("=== begin %s()\n", __FUNCTION__);
@@ -139,15 +152,9 @@ static void _test_get_dates(void)
     oe_datetime_t next;
     OE_TEST(oe_crl_get_update_dates(&crl, &last, &next) == OE_OK);
 
-    OE_TEST(last.year == _time.year);
-    OE_TEST(last.month == _time.month);
-    OE_TEST(last.day == _time.day);
-    OE_TEST(last.hours == _time.hours);
-    OE_TEST(last.minutes == _time.minutes);
-    // TEMPORARILY Disabled: _time is the certificate file's timestamp
-    // whereas last.seconds is the actual issue time of the certificate.
-    // The two values are not guaranteed to be equal.
-    // OE_TEST(last.seconds == _time.seconds);
+    int32_t diff = _diff_oe_datetime(&last, &_time);
+    printf("=== Diff between last and current time: %d\n", diff);
+    OE_TEST(abs(diff) < 60); // Ensure times are within 60 seconds
 
     oe_datetime_t next_expected;
     next_expected.year = _time.year + 1;
@@ -155,6 +162,7 @@ static void _test_get_dates(void)
     next_expected.day = _time.day;
     next_expected.hours = _time.hours;
     next_expected.minutes = _time.minutes;
+    next_expected.seconds = _time.seconds;
 
     // If a leap day occurs in the next time period, need to adjust
     // expected next date.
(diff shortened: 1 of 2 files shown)


jhand2:
bors try

jhand2:
Open to a smaller tolerance. 5 seconds is probably fine

jhand2:
Huh, either my diff function is broken or the certs really are far apart: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAgnostic-Linux/detail/Agnostic-Linux/527/pipeline/25

I'll look into it later

jhand2:
Huh, either my diff function is broken or the certs really are far apart: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAgnostic-Linux/detail/Agnostic-Linux/527/pipeline/25

I'll look into it later. Trying again for now

bors try

tests/crypto/crl_tests.c
```diff
@@ -139,22 +152,17 @@ static void _test_get_dates(void)
...
-    OE_TEST(last.day == _time.day);
-    OE_TEST(last.hours == _time.hours);
-    OE_TEST(last.minutes == _time.minutes);
-    // TEMPORARILY Disabled: _time is the certificate file's timestamp
-    // whereas last.seconds is the actual issue time of the certificate.
-    // The two values are not guaranteed to be equal.
-    // OE_TEST(last.seconds == _time.seconds);
+    int32_t diff = _diff_oe_datetime(&last, &_time);
+    printf("=== Diff between last and current time: %d\n", diff);
+    OE_TEST(abs(diff) < 60); // Ensure times are within 60 seconds
```
dthaler, line 157: Suggest creating a #define'd constant that you can use here and in line 216 rather than hard coding 60 in both places.   And is the number 60 arbitrary or is there a reason it's better than say 10 or 300?

jhand2: It's arbitrary. I'm open to changing it, I wanted to experiment with what the value normally is, although it my local testing it's pretty much always 0 or 1

tests/crypto/crl_tests.c
```diff
@@ -127,6 +127,19 @@ static bool _year_is_a_leap_year(uint32_t year)
     }
 }
 
+static int32_t _diff_oe_datetime(oe_datetime_t* t1, oe_datetime_t* t2)
+{
+    int32_t dyear = (int32_t)t1->year - (int32_t)t2->year;
+    int32_t dmonth = (int32_t)t1->month - (int32_t)t2->month;
```
anakrish, line 133: For a correct `diff` of date time won't there be logic to handle the varying lengths of different months, leap years etc?

tests/crypto/crl_tests.c
anakrish, line 157: If the tolerance needs to in minutes can we instead just add the tolerance to the check that compares the minutes?:

`OE_TEST( abs((int32_t)last.minutes -_(int32_t)time.minutes) <= 1)`;

tests/crypto/crl_tests.c
```diff
@@ -127,6 +127,19 @@ static bool _year_is_a_leap_year(uint32_t year)
...
+    int32_t dday = (int32_t)t1->day - (int32_t)t2->day;
+    int32_t dhour = (int32_t)t1->hours - (int32_t)t2->hours;
+    int32_t dmin = (int32_t)t1->minutes - (int32_t)t2->minutes;
+    int32_t dsec = (int32_t)t1->seconds - (int32_t)t2->seconds;
+
+    return (dyear * 31556952) + (dmonth * 2592000) + (dday * 86400) +
```
CodeMonkeyLeet, line 139: These are date-time values and not cumulative time values, so individual elements in t1 may be less that individual elements in t2, resulting in the addition of negative numbers. If you want to do this, you'd need to do the conversion to cumulative time first, and then the diff

It doesn't necessarily matter in the use case here, but really it seems like you should implement a specific boundary limit function instead of implementing a generic datetime diff function. i.e. in `_test_get_dates()` call `oe_datetime_compare()` and if they are not equal, then you only have to deal with checking that your min + sec difference is within your expected time bound.

tests/crypto/crl_tests.c
dthaler: Actually that's a great point.  A safer algorithm would be to convert to time_t (using mktime or whatever) and then take the difference in seconds.

tests/crypto/crl_tests.c
jhand2: minute 1, second 59
minute 2, second 1

How far apart are these?

jhand2: Tolerance is in seconds. 60 was a random choice, a much smaller number would be ideal

tests/crypto/crl_tests.c
jhand2: Dave's suggestion to convert to time_t is probably easier. I'll look into that.

johnkord:
Last night I was bored and I came across this PR, so I've added a friendly amendment to @jhand2 's branch that addresses some of these comments (this is, after all, related to my pet interview question). Feel free to accept the changes to your branch, @jhand2  or do what you wish with them! :)

https://github.com/jhand2/openenclave/pull/1

### Testing Notes
(Tested on Ubuntu 18.04 with no SGX as well as on Windows 10 with SGX1. I think there's a new Jenkins instance and I don't have access to the Sandbox there, so I can't give this branch a run through. I think this branch may need to be rebased on master to work on Windows, though)

jhand2:
bors try

jhand2:
Thanks @johnkord! I merged you PR into this branch.

CodeMonkeyLeet:
@dthaler I believe that your comment was addressed, can you review and remove your change request on this PR?

radhikaj:
Bors r+

PR #3069, opened 2020-05-24
EXPERIMENTAL: Use C++ implementation of oeedger8r

anakrish:
Pass USE_CPP_EDGER8R=on to use C++ oeedger8r instead of Ocaml edger8r.

C++ oeedger8r repository: https://github.com/openenclave/oeedger8r-cpp

The Ocaml implementation has the following drawbacks
- Ocaml is not that common a language and has proven to
  have a high barrier for entry to contributing to oeedger8r
- Ocaml compiler/tools are not readily available on Windows
  This has led to
  - Using not actively maintained build on Ocaml on Windows
  - Or using esy which builds Ocaml compiler on each Windows
    machine it is run on, adding upto 30 additional minutes
    to building the OE SDK on Windows

This C++ implementation is
- intended to improve the ease of contribution to the oeedger8r
  by use of a common language that is well supported on all
  platforms
- sticks to C++11 which is well supported
- is based on a Python rewrite of oeedger8r
  https://github.com/openenclave/oeedger8r-python

Signed-off-by: Anand Krishnamoorthi <anakrish@microsoft.com>

PR #3082, opened 2020-05-29
Switch to C++ edger8r.

anakrish:
Switch to C++ oeedger8r
Benefits
- Reduced barrier to contribution
- Avoids tooling issues; just needs C++ compiler
- Easier to add features due to use of a common programming language

Remove dependency on nodejs, esy
Benefits
- Significantly reduced SDK build times on fresh Windows machines.
  Previously took additional 20-30 mins to build Ocaml compiler and then oeedger8r
- No need for administrator prompt to build OE SDK
- Windows Pipeline reduces from 48 minutes to 24 minutes

Build oeedger8r as external project.

This ensures that ARM binary is not generated for OE_TrustZone
build of the SDK which requires the edger8r to be run on the
host x64 machine.

Latest


anakrish:
bors try

anakrish:
bors try

tools/CMakeLists.txt
```diff
@@ -1,18 +1,42 @@
...
-  add_subdirectory(oeedger8r-cpp)
-  add_executable(edger8r ALIAS oeedger8r)
-  install(
-    TARGETS oeedger8r
-    EXPORT openenclave-targets
-    DESTINATION ${CMAKE_INSTALL_BINDIR})
-else ()
-  add_subdirectory(oeedger8r)
-endif ()
+check_submodule_not_empty(oeedger8r-cpp)
```
radhikaj, line 4: Wouldnt this need to be added as a submodule to this project?

radhikaj:
Thanks @anakrish. The build time improvement is great! The removal of dependencies that need to be downloaded is also a big plus.

anakrish:
bors try

tools/CMakeLists.txt
anakrish: It is already a submodule. Added as part of #3069

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

andschwa:
@anakrish @radhikaj An alternative you could consider: decouple the build of the existing edger8r from the main Open Enclave repository, and instead just take it as a package dependency. Just move it into its own repo, set it up to build by itself and upload as a GitHub release, and configure the upstream Open Enclave build to download the release. Would probably be simpler than rewriting and maintaining a C++ implementation, and avoids the issue of building the edger8r for most folks.

anakrish:
@andschwa Thanks for the suggestions.

We are already decoupling the edger8r to a separate repository as part of the move to C++. That repository is self-contained and the edger8r can be built and tested in a matter of seconds.

For the current Ocaml based edger8r, the build times can be improved by reverting to building via native ocaml compiler instead of esy. It is possible to cache esy results and then restore performance; however that is not a a feasible solution for various build environments of the SDK; and that feels at best a hack to get around the fundamental problem that in order to compile around 12 ocaml files, esy builds the entire Ocaml compiler from scratch. Whatever workarounds we have for esy does not scale in practice and overall slows down the OE SDK CI and hurts developer experience.

The move to C++ is also intended to reduce the contribution barrier. Ocaml is a niche language and in practice has prevented more contribution to edger8r due to the steep learning curve. In other projects Ocaml might naturally fit in; but here it just seems to be more trouble than its worth.

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

andschwa:
> @andschwa Thanks for the suggestions.
> 
> We are already decoupling the edger8r to a separate repository as part of the move to C++. That repository is self-contained and the edger8r can be built and tested in a matter of seconds.
> 
> For the current Ocaml based edger8r, the build times can be improved by reverting to building via native ocaml compiler instead of esy. It is possible to cache esy results and then restore performance; however that is not a a feasible solution for various build environments of the SDK; and that feels at best a hack to get around the fundamental problem that in order to compile around 12 ocaml files, esy builds the entire Ocaml compiler from scratch. Whatever workarounds we have for esy does not scale in practice and overall slows down the OE SDK CI and hurts developer experience.
> 
> The move to C++ is also intended to reduce the contribution barrier. Ocaml is a niche language and in practice has prevented more contribution to edger8r due to the steep learning curve. In other projects Ocaml might naturally fit in; but here it just seems to be more trouble than its worth.

Did you try just using `dune` from OPAM as I suggested in the other PR, instead of writing CMake to drive the OCaml compilers? It'd be really straightforward, it just leaves you with esy as the Windows story.

anakrish:
> 
> Did you try just using `dune` from OPAM as I suggested in the other PR, instead of writing CMake to drive the OCaml compilers? It'd be really straightforward, it just leaves you with esy as the Windows story.
> 
What benefit would that give? It does not really solve the underlying issue with esy.

CHANGELOG.md
```diff
@@ -20,9 +20,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `COMPILE_SYSTEM_EDL` is now OFF by default, meaning system EDL must be imported by
   application EDL. See [system EDL opt-in document]
   (docs/DesignDocs/system_ocall_opt_in.md#how-to-port-your-application) for more information.
+- Switch to oeedger8r written in C++.
 
 ### Removed
 - Removed oehostapp and the appendent "-rdynamic" compiling option. Please use oehost instead and add the option back manually if necessary.
+- Removed dependency on nodejs and esy. Previously used to build Ocaml compiler and oeedger8r.
```
mingweishih, line 27:
```suggestion
- Removed dependencies on nodejs and esy, which were previously used to build Ocaml compiler and oeedger8r.
```

anakrish:
bors try

CHANGELOG.md
anakrish: Done.

Issue #24, opened 2017-11-21
Core Testing: Expand tests for thread binding behavior for enclaves


CodeMonkeyLeet:
We need to flush out the details for behaviors of the enclave APIs and validate them. This includes thread binding behavior and its interactions in multi-threaded scenarios:

- ~~Thread binding correctly avoids deadlocks from thread exhaustion in the enclave~~
    - [x] Thread binding now fails if we exhaust all enclave threads, we should simply verify that host handles this failure gracefully. (Addressed by PR #1023 )
- - [X] Validate implementations of all OE synchronization primitives (Covered with #476 )
- - [X] Stress and load tests for multi-threaded enclave applications (Missing coverage provided with #612 )

anitagov:
Need a test to verify that a reentrant ecall uses the same thread as the original ecall. => This is no longer applicable as reentrancy with ocalls is no longer supported (See Issue #474 ).

anitagov:
Thread binding now fails if we exhaust all enclave threads, we should simply verify that host handles this failure gracefully. => Adding a new test to thread tests to verify this.

anitagov:
PR #1023 adds a new test to verify TCS exhaustion returns OE_OUT_OF_THREADS has been merged to resolved.
The main 3 bullet items have been addressed. So, closing this issue.

PR #3086, opened 2020-06-01
Proposal of refactoring the calling mechanisms of TEE functions 

mingweishih:
Problem (based #3028):
OE supports the EDL stubs for two different enclaves being linked into the same host binary.
Using weak symbols to prevent the duplicate definitions of host-side stub is insufficient as ecalls are made by ID so the ID that is sent to the enclave may not match what the enclave expects. 

This PR adds a proposal of refactoring the calling mechanisms of TEE functions in order to address this problem.

### DIFF
--- a/docs/DesignDocs/TEECallsRefactoring.md
+++ b/docs/DesignDocs/TEECallsRefactoring.md
@@ -1,0 +1,442 @@
+Generalize the Calling Mechanisms of Enclave Functions
+=====
+
+Introduction
+------------
+
+Unlike typical function invocations via `call` instructions, TEE functions require
+specialized calling mechanisms that allow an enclave and a host to interact with each other.
+In the Open Enclave SDK, we use OCALLs to refer to the host-side functions for the
+enclave to use and use ECALLs to refer to the enclave-side functions for the host to use.
+The invocation of an OCALL/ECALL is analogous to message passing between two isolated entities via a channel.
+Regardless of the type of a channel, which depends on the implementation of TEEs, invoking an ECALL/OCALL requires
+a contract between the enclave and the host so that they know what messages to send and how to
+interpret the messages they receive and to take actions accordingly.
+
+This document describes the current implementation of ECALL invocations, its limitations, and the
+proposal of code refactoring.
+
+Current Implementation
+-----------------------
+
+**EDL**
+
+The Open Enclave SDK currently uses the oeedger8r tool to generate the stub code of ECALLs and OCALLs
+from an EDL file. The following example shows a file that defines an ECALL `sample_ecall()`
+and an OCALL `sample_ocall()`.
+
+```
+// sample.edl
+enclave
+{
+    trusted
+    {
+        public void sample_ecall();
+    };
+
+    untrusted
+    {
+        public void sample_ocall();
+    };
+}
+```
+
+Given the `sample.edl`, the oeedger8r generates `sample_u.c`/`sample_u.h` as the host-side stub
+and `sample_t.c`/`sample_t.h` as the enclave-side stub.
+
+**ECALL Invocation**
+
+To invoke an ECALL, the host simply calls the `sample_ecall()` function by feeding
+the pointer of the target `enclave` struct as the following code snippet.
+
+```
+result = sample_ecall(enclave);
+```
+
+The `sample_ecall()`, which is defined in `sample_u.c`, performs a series of checks,
+marshals the arguments, and invokes the `oe_call_enclave_function()` function
+as the following example.
+
+```
+if ((_result = oe_call_enclave_function(
+        enclave,
+        sample_fcn_id_sample_ecall,
+        _input_buffer,
+        _input_buffer_size,
+        _output_buffer,
+        _output_buffer_size,
+        &_output_bytes_written)) != OE_OK)
+    goto done;
+```
+
+The function, which is part of the `oehost` library, eventually dispatches the
+provided arguments to the target enclave (specified by the `enclave`).
+The argument `sample_fcn_id_sample_ecall` (with the `enum` type, defined in both
+`sample_u.h` and `sample_t.h`) is used as an identifier that allows the enclave to
+locate the corresponding ECALL. More specifically, `sample_t.c` defines
+`__oe_ecalls_table` (an array of function pointers) and `__oe_ecalls_table_size` as follows.
+
+```
+oe_ecall_func_t __oe_ecalls_table[] = {
+    (oe_ecall_func_t) ecall_sample_ecall
+};
+
+size_t __oe_ecalls_table_size = OE_COUNTOF(__oe_ecalls_table);
+```
+
+Both of them are global variables that are linked by the `oecore` library
+and are used in the `oe_handle_call_enclave_function()` function,
+which processes every request of ECALLs. The following code snippet shows
+how the function looks up the ECALL wrapper from the `__oe_ecalls_table`.
+
+```
+...
+ecall_table.ecalls = __oe_ecalls_table;
+ecall_table.num_ecalls = __oe_ecalls_table_size;
+
+if (args.function_id >= ecall_table.num_ecalls)
+    OE_RAISE(OE_NOT_FOUND);
+
+func = ecall_table.ecalls[args.function_id];
+...
+
+// Call the function.
+func(
+    input_buffer,
+    args.input_buffer_size,
+    output_buffer,
+    args.output_buffer_size,
+    &output_bytes_written);
+```
+
+After looking up based on the function id (`sample_fcn_id_sample_ecall`),
+`oe_handle_call_enclave_function()` calls into the wrapper function `ecall_sample_ecall()`(defined in `sample_t.c`).
+The wrapper function performs a series of checks, unmarshals the arguments, and invokes
+the ECALL implemented by the enclave with the following code snippet.
+
+```
+/* Call user function. */
+pargs_out->_retval = sample_ecall(
+);
+```
+
+At this point, an ECALL is successfully dispatched to the enclave.
+
+Limitation
+---------
+
+The current implementations pose a limitation to the scenario where the
+host instantiates different enclaves that use the same set of ECALLs
+(i.e., importing the same EDL files).
+Take the following case study for example.
+
+**Case Study 1: Importing the same EDLs with different order**
+
+Assume the OE SDK provides two EDL files that allow enclaves to opt-in as follows.
+
+```
+// common_1.edl
+enclave
+{
+    trusted
+    {
+        public void common_1_ecall();
+    };
+}
+```
+
+```
+// common_2.edl
+enclave
+{
+    trusted
+    {
+        public void common_2_ecall_1();
+        public void common_2_ecall_2();
+    };
+}
+```
+
+Consider two enclaves, `foo` and `bar`. Their EDL files are defined as follows.
+
+```
+// foo.edl
+enclave
+{
+    from "common_1.edl" import *;
+    from "common_2.edl" import *;
+
+    trusted
+    {
+        public void foo_ecall();
+    };
+}
+```
+
+```
+// bar.edl
+enclave
+{
+    from "common_2.edl" import *;
+    from "common_1.edl" import *;
+
+    trusted
+    {
+        public void bar_ecall();
+    };
+}
+```
+
+Note that both the EDL files import the two SDK-provided EDL files but in
+a different order. The resulting ECALL tables defined in the
+corresponding `_t.c` and `_u.c` files will be as follows.
+
+- `foo` ECALL table
+  ```
+  oe_ecall_func_t __oe_ecalls_table[] = {
+      (oe_ecall_func_t) ecall_common_1_ecall,
+      (oe_ecall_func_t) ecall_common_2_ecall_1,
+      (oe_ecall_func_t) ecall_common_2_ecall_2,
+      (oe_ecall_func_t) ecall_foo_ecall
+  };
+  ```
+- `bar` ECALL table
+  ```
+  oe_ecall_func_t __oe_ecalls_table[] = {
+      (oe_ecall_func_t) ecall_common_2_ecall_1,
+      (oe_ecall_func_t) ecall_common_2_ecall_2,
+      (oe_ecall_func_t) ecall_common_1_ecall,
+      (oe_ecall_func_t) ecall_bar_ecall
+  };
+  ```
+
+In addition, both `foo_u.c` and `bar_u.c` will implement same wrapper functions
+of the three imported functions. Assuming that we use *weak symbol* to avoid the
+duplication of the wrapper functions (i.e., only one implementation of the functions is picked),
+the invocation of these ECALLs to both enclaves ends up using the same function ids.
+
+See the following host-side code snippet for an example.
+```
+// ECALL to the enclave `foo`.
+common_1_ecall(enclave_foo);
+
+// ECALL to the enclave `bar`.
+common_1_ecall(enclave_bar);
+```
+
+Since the host can have only one implementation of `ecall_common_1_ecall` (the wrapper function),
+both invocations end up using the same function id, say `0`. However, this causes the mismatch
+on the call into the enclave `bar` where the expected id should be `2`.
+
+**Case Study 2: Selectively importing the ECALLs from same EDLs**
+
+Another scenario that both two enclaves selectively import ECALLs from the same EDLs but the
+selections are different.
+
+```
+// foo.edl
+enclave
+{
+    from "common_2.edl" import *;
+
+    trusted
+    {
+        public void foo_ecall();
+    };
+}
+```
+
+```
+// bar.edl
+enclave
+{
+    from "common_2.edl" import common_2_ecall_2;
+
+    trusted
+    {
+        public void bar_ecall();
+    };
+}
+```
+
+The resulting tables will be as follows.
+- `foo` ECALL table
+  ```
+  oe_ecall_func_t __oe_ecalls_table[] = {
+      (oe_ecall_func_t) ecall_common_2_ecall_1,
+      (oe_ecall_func_t) ecall_common_2_ecall_2,
+      (oe_ecall_func_t) ecall_foo_ecall
+  };
+  ```
+- `bar` ECALL table
+  ```
+  oe_ecall_func_t __oe_ecalls_table[] = {
+      (oe_ecall_func_t) ecall_common_2_ecall_2,
+      (oe_ecall_func_t) ecall_bar_ecall
+  };
+  ```
+
+Clearly, we can see the conflict on the function id.
+
+Proposed revision
+-----------
+
+The main goal of this proposal is to address the limitation that different enclaves on the
+same host cannot import the same EDL files (i.e., using the same set of ECALLs).
+To this end, this document proposes to refactor the ECALL calling mechanism as follows.
+
+**Maintain Local and Global ECALL tables on the Host**
+
+- Local Tables
+
+  Given that the rule of importing `EDL` files is flexible, the resulting
+  ECALL table is not deterministic; i.e., the same ECALL may have different ids
+  across enclaves. To solve the problem, this document uses an idea of storing the
+  list of ECALL id per enclave, which we refer to as local ECALL tables.
+  The local table is declared as follows.
+
+  ```
+  typedef struct _oe_ecall_id_t
+  {
+     uint64_t id;
+  } oe_ecall_id_t;
+
+  typedef struct _oe_enclave
+  {
+     ...
+     /* Per-edl ecall id table. */
+     oe_ecall_id_t ecall_id_table[OE_ECALL_MAX];
+     uint32_t ecall_id_table_size;
+     ...
+  } oe_enclave_t;
+  ```
+
+  The local table allows the host to retrieve the ECALL id per enclave instead
+  of using the hard-coded id in the same ECALL across different enclaves.
+
+- Global Table
+
+  To guarantee the constant-time look-up, the host needs to use the same id to retrieve
+  the ECALL id from the local table. To do so, the host additionally maintains a global
+  table that keeps such ids to which we refer as global ids. The detailed design is as follows.
+
+  - The global table is defined internally in the `oehost` as follows. The table stores the name
+    of each ECALL and the index to each name is used as the global id.
+    ```
+    const char* _ecall_table[OE_ECALL_MAX];
+    uint32_t _ecall_table_size;
+    ```
+  - The global table is updated during every initialization of a local table. More specifically,
+    the list of ECALL names is added to the `_u.c` and is passed to the `oe_create_enclave`.
+    ```
+    typedef struct _oe_ecall_info_t
+    {
+        const char* name;
+    } oe_ecall_info_t;
+
+    /**** Trusted function names. ****/
+    const oe_ecall_info_t __foo_ecall_info_table[] =
+    {
+        { "ecall_common_1_ecall" },
+        { "ecall_common_2_ecall_1" },
+        { "ecall_common_2_ecall_2" },
+        { "ecall_foo_ecall" },
+        { NULL }
+    };
+    ```
+    ```
+    oe_result_t oe_create_foo_enclave(
+      const char* path,
+      oe_enclave_type_t type,
+      uint32_t flags,
+      const oe_enclave_setting_t* settings,
+      uint32_t setting_count,
+      oe_enclave_t** enclave)
+    {
+        return oe_create_enclave(
+                 path,
+                 type,
+                 flags,
+                 settings,
+                 setting_count,
+                 __foo_ocall_function_table,
+                 3,
+                 __foo_ecall_info_table,
+                 3,
+                 enclave);
+    }
+    ```
+  - During the initialization of the local table, the host uses the following logic
+    to get the global id by name.
+    ```
+    uint32_t i, index;
+    for (i = 0; i < ecall_info_table_size; i++)
+    {
+      if found name in the global_table
+        index = the index of the name
+      else /* not found */
+        add the name to the table
+        index = the index of the name
+
+      /* Assign the ECALL id to the local table based on the global id. */
+      enclave->ecall_id_table[index] = i;
+    }
+    ```
+  - Modify the internal `oe_call_enclave_function` function to pass a global id
+    and the name string of the ECALL instead of hard-coded ECALL id.
+    The `global_id` is defined as a *static variable*. Given the same `global_id`
+    is used to retrieve the same ECALL across enclaves, the host initializes the
+    `global_id` (based on the name) in the first invocation and uses the cached value
+    for the subsequent ones.
+    ```
+    static uint64_t global_id = OE_GLOBAL_ECALL_ID_NULL;
+    ...
+    /* Call enclave function. */
+    if ((_result = oe_call_enclave_function(
+             enclave,
+             &global_id,
+             __foo_ecall_info_table[foo_fcn_id_ecall_common_1_ecall],
+             _input_buffer,
+             _input_buffer_size,
+             _output_buffer,
+             _output_buffer_size,
+             &_output_bytes_written)) != OE_OK)
+      goto done;
+    ```
+
+In our previous example, assuming that the host instantiates the `foo` before `bar`.
+By applying the above scheme, the host will maintain the following data structures.
+- `_ecall_table`
+  ```
+  _ecall_table[0]: { "ecall_common_1_ecall" }
+  _ecall_table[1]: { "ecall_common_2_ecall_1" }
+  _ecall_table[2]: { "ecall_common_2_ecall_2" }
+  _ecall_table[3]: { "ecall_foo_ecall" }
+  _ecall_table[4]: { "ecall_bar_ecall" }
+  ```
+- `foo` (at the time `bar` is not initialized yet)
+  ```
+  enclave->ecall_id_table[0]: { 0 /* ecall_common_1_ecall */ }
+  enclave->ecall_id_table[1]: { 1 /* ecall_common_2_ecall_1 */ }
+  enclave->ecall_id_table[2]: { 2 /* ecall_common_2_ecall_2 */ }
+  enclave->ecall_id_table[3]: { 3 /* ecall_foo_ecall */ }
+  ```
+- `bar`
+  ```
+  enclave->ecall_id_table[0]: { 2 /* ecall_common_1_ecall */ }
+  enclave->ecall_id_table[1]: { 0 /* ecall_common_2_ecall_1 */ }
+  enclave->ecall_id_table[2]: { 1 /* ecall_common_2_ecall_2 */ }
+  enclave->ecall_id_table[3]: { OE_ECALL_ID_NULL /* ecall_foo_ecall */ }
+  enclave->ecall_id_table[4]: { 3 /* ecall_bar_ecall */ }
+  ```
+
+In the oeedger8r-generated stub code of `ecall_common_1_ecall`, the local static variable
+`global_id` is initialized as `OE_GLOBAL_ECALL_ID_NULL`. After the first invocation of the ecall,
+the host looks up the global id by name and sets the `global_id` to `0`. For the subsequent invocations,
+the host directly uses the cached value `0` to look up the ecall id from the local table of each enclave
+(i.e., getting `0` from `foo` and `2` from `bar`) and therefore achieves the goal of constant-time look-up.
+
+Authors
+-------
+
+- Ming-Wei Shih <mishih@microsoft.com>


docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+    const void* report,
+    size_t report_size)
+  {
+      oe_result_t _result = OE_FAILURE;
+
+      /* CRC32 of oe_verify_report_ecall */
+      const uint32_t _function_hash = 3802944385;
+
+      /* Used to cache the index to the id caching table. */
+      static int _caching_table_id = OE_HOST_ECALL_ID_NULL;
```
anakrish, line 472: Using static variables introduces thread-safety concern in multi-threaded applications.
Can we avoid these statics?

mingweishih: These static variables are constant once set even in multi-threaded scenarios. Do you think that it still need to be avoided?

anakrish: They need to me made thread-safe if we plan to use them. While one thread is making a partial write to these variables, another thread could read the half-written value.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+---------
+
+The current implementation poses the following limitations.
+
+- Single enclave cannot include multiple EDL files.
+
+  According to the implementation, the `oecore` library
+directly links against the `__oe_ecalls_table` defined in a `_t.c` file.
+Therefore, the current logic cannot handle the case of including
+multiple `_t.c` files.
```
anakrish, line 194: Is there a need for one enclave including multiple `_t.c` files?

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+
+- Multiple enclaves on the same host cannot use/import the same EDL file.
```
anakrish, line 196: Consider rewording this more accurately. Multiple enclaves instances can today be created from the same enclave binary.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: Currently, this is only the internal use cases (e.g., `core.edl`, `syscall.edl`, etc.). Besides, I don't really see why we need to limit only a single external EDL file per enclave.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: In that case, can we add locks around the load/store?

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+typedef struct _oe_ocall_t {
+    oe_ocall_func_t ocall;
+    uint32_t hash;
+} oe_ocall_t;
+```
+
+**Maintain the List of Registration Functions**
+
+Each registration function in a EDL file is added to a list via the constructor.
+On Linux, we can do this via ` __attribute__((constructor))` as follows.
```
anakrish, line 382: One issue with `__attribute__((constructor))` is that we cannot guarantee the order in which these constructors will be called. What if the user code has an constructor that makes an ocall?

Can we ensure that we invoke these constructors at a well defined point in enclave initialization?

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+The list of registration functions can then be invoked during the host and enclave
+initialization.
+  
+**Cache the Function Id for the Const-Time Look-Up**
+
+One important property of current implementation is the const-time function look-up that
+this proposal would like to maintain. To this end, the design is obtaining the id by hash
+in the first invocation and caching the id for subsequent ones.
+
+- Caching an OCALL id.
```
anakrish, line 401: From what I understand, there are currently no issues with ocalls. If ocall_foo has id 1 in enclave_a and id 2 in enclave_b, enclave_a will pass 1 as function id when making the ocall whereas enclave_b will pass 2 as function id when making the ocall. Ocall are dispatched via specific ocall tables; one per edl file. These tables are guaranteed to have correct entries at index 1 and index 2 respectively.

docs/DesignDocs/TEECallsRefactoring.md
anakrish: If we are allowing multiple EDL files per enclave, then there it would make sense to revisit the `import` feature in EDL.

- When/Why would the user use the import statement ?
- When/Why would the user use multiple EDL files?

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: The would no longer be true after introducing opt-in system EDLs. I'd like @jhand2's input here. He mentioned a case that two enclave import the same EDL may result in a conflict.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: We don't do the initialization inside the constructor. Instead, it simply adds the registration functions to a linked list.  The list of the functions will be invoked at a well-defined point (the same as how we register the per-EDL table today). It will be clearer when I create the patch :)

docs/DesignDocs/TEECallsRefactoring.md
anakrish: Yes, either locks or lock-free atomics.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+
+      int _ecall_id = -1;
+      
+      ...
+      
+      /* Check if the index is cached. */
+      if (_index == OE_HOST_ECALL_ID_NULL)
+      {
+        _index = oe_host_get_ecall_id_by_hash(
+            enclave,
```
anakrish, line 482: Since
> As a result, the index of `sample_ecall` in
  `enclave_foo->ecall_id_table` and `enclave_bar->ecall_id_table` is the same, which allows for caching
  the ECALL id on the per enclave basis.

do we need the enclave as a parameter here?

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+  Because a single host may interact with multiple enclaves, we cannot directly cache the ECALL id inside an ECALL wrapper;
+  i.e., the host may use the same ECALL wrapper to dispatch the request to different enclaves. As a result, we need to
+  cache the id per enclave. More specifically, the host maintains an id caching table of ECALLs per `enclave` struct. Note that
+  this table includes all the ECALLs from multiple enclaves such that every `enclave` structs have the same table.
+  
+  Take a similar example of two enclaves: `enclave_foo` and `enclave_bar`. Both of the enclaves implement
+  the ECALL `sample_ecall`, but with different ids (e.g., `1` in `enclave_foo` and `2` in `enclave_bar`).
+  To cache the id per enclave, the `enclave` structs of both enclaves, `enclave_foo` and `_enclave_bar`,
+  on the host maintains the same table `ecall_id_table`. As a result, the index of `sample_ecall` in
+  `enclave_foo->ecall_id_table` and `enclave_bar->ecall_id_table` is the same, which allows for caching
```
anakrish, line 451: Great idea!

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+            _function_hash);
+      }
+
+      ...
+      
+      /* Obtain the ecall id from the caching table. */
+      _ecall_id = oe_get_ecall_function_id(enclave, _index);
+      if (_ecall_id == -1)
+      {
+          oe_call_enclave_get_function_id_by_hash(
```
anakrish, line 492: One interesting question is  how `oe_call_enclave_get_function_id_by_hash` itself will be dispatched? It seems to be a special ecall.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: Yes. This function call is essentially an ECALL. We need to specify the `enclave` here.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: This is true for the current implementation. But with the proposal, the local function id (defined in the `_t.c`) may not equal to the actual function id given that the ocall table can include OCALLs from multiple EDL files.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: Yes, we will need one special ecall and one special ocall for querying the id.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: I kind of thinking this as an option for users; i.e., they can choose either option based on their needs.

For example, users can import "string.edl" in "book.edl" and define a "math.edl" that implements an irrelevant set of ECALLs/OCALLs.

anakrish: But what would be the guideline for users? There does not seem to be a clear benefit that either approach has over the other. If I can specify multiple EDLs per enclave, then I don't see much value in allowing imports.

mingweishih: An internal use case of core.edl:
```
enclave
{
    /* The end goal is to publish all EDL files as part of the SDK. At that time
     * enclave developers will simply import the edl files/functions they
     * need and this file will no longer be required. These files are imported
     * as a convenience to allow refactoring of system EDL files without the need
     * to add a new build definition for each file.
     */
    from "attestation.edl" import *;
    from "keys.edl" import *;
    from "logging.edl" import *;
    from "memory.edl" import *;
};
```

docs/DesignDocs/TEECallsRefactoring.md
anakrish: Consider if the mapping of local ocall function id to global ocall function id can be done entirely on the host side.

docs/DesignDocs/TEECallsRefactoring.md
anakrish: But with this PR, we are also saying that the user can use 
"my_enclave.edl" and "attestation.edl" separately into their enclave without my_enclave.edl having to import "attestation.edl".
So it is not clear to my when I'd choose to import.

jhand2: @mingweishih core.edl and syscall.edl are helpers to make the transition to EDL import easier. We shouldn't really encourage using them.

I agree with Anand, I don't think supporting both mechanisms is beneficial.

Personally I prefer the import method because EDL already supports imports and it's easier to explain to users. From a technical standpoint, they mostly do that same thing.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+    oe_ocall_func_t ocall;
+    uint32_t hash;
+} oe_ocall_t;
+```
+
+**Maintain the List of Registration Functions**
+
+Each registration function in a EDL file is added to a list via the constructor.
+On Linux, we can do this via ` __attribute__((constructor))` as follows.
+We can use an equivalent approach for the constructor on Windows.
```
jhand2, line 383: Does an equivalent approach exist on Windows?

mingweishih: According to https://stackoverflow.com/questions/1113409/attribute-constructor-equivalent-in-vc, yes, but it needs some experiments.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
+Generalize the Calling Mechanisms of TEE Functions
+=====
+
+Introduction
+------------
+
+Unlike typical function invocations via `call` instructions, TEE functions require
+specialized calling mechanisms that allow an enclave and a host to interact with each other.
+In the Open Enclave SDK, we use OCALLs to refer to the host-side functions that for the
```
dthaler, line 9:
```suggestion
In the Open Enclave SDK, we use OCALLs to refer to the host-side functions for the
```

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+Considering an example with two enclaves, say `enclave_foo` and `enclave_bar`.
+Each enclave defines its own EDL file with a different number of ECALLs.
+In addition, both the EDL files import the same EDL file (the same `sample.edl` as above).
+As a result, the corresponding ECALL tables (defined in each `_t.c` file) include
+the wrapper function of `sample_ecall` but with different function id (because
+the number of ECALLs is different).
+Assuming that we use *weak symbol* to avoid the naming conflict of
+the wrapper function (i.e., only one implementation of the functions is picked),
+the invocation of `sample_ecall` to both enclaves ends up using the same
+table id and function id and potentially causing a mismatching;
```
dthaler, line 320:
```suggestion
table id and function id and potentially causing a mismatch;
```

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: I would think from the other way:  the SDK internal use of EDLs does not entirely use import. Instead, it currently uses the `_t_wrapper.c`/`_u_wrapper.c` (and includes a table registration function that requires the oecore to link against). This refactoring that gets rid of `_t_wrapper.c`/`_u_wrapper.c` naturally supports an enclave to include multiple EDL files. I don't really think there is a reason to limit that.

This is also necessary for the `oecore` refactoring. For example, moving the attestation-related OCALLs/OCALLs out.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+      __oe_ecalls_table,
+      __oe_ecalls_hash_list,
+      __oe_ecalls_table_size);
+}
+```
+
+Each table entry is then defined as the following struct.
+
+```
+typedef struct _oe_ocall_t {
```
dthaler, line 373: If this is defined, then I would rather see the generated code pass an array of these structs to oe_register_ecall_functions, rather than two parallel arrays.

docs/DesignDocs/TEECallsRefactoring.md
dthaler: The text is not correct.   `__attribute__((constructor))` is not specific to the OS (e.g., Linux), it's specific to the **compiler** (e.g., gcc).  As such you can do it on both Windows or Linux but only with specific compilers that support `__attribute__`.  For example it wouldn't work with MSVC on the host side.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+```
+
+**Maintain the List of Registration Functions**
+
+Each registration function in a EDL file is added to a list via the constructor.
+On Linux, we can do this via ` __attribute__((constructor))` as follows.
+We can use an equivalent approach for the constructor on Windows.
+```
+/* Constructor to register the callback. */
+ __attribute__((constructor)) static void core_ecall_functions_constructor()
```
dthaler, line 386: We should never have `__attribute__((constructor))` in any c/c++ file.  At most it should be in an OE_something macro because it varies by compiler (if it's supported at all, and I'm not yet convinced MSVC has a workable solution though I have not investigated the info on stackoverflow that you (MingWei) pointed to, so I don't know).

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+  
+**Cache the Function Id for the Const-Time Look-Up**
+
+One important property of current implementation is the const-time function look-up that
+this proposal would like to maintain. To this end, the design is obtaining the id by hash
+in the first invocation and caching the id for subsequent ones.
+
+- Caching an OCALL id.
+
+  Because an enclave only interacts with a single host, the case of OCALL wrapper is straightforward:
```
dthaler, line 403: Re "an enclave only interacts with a single host"... this is not true for OP-TEE in general which supports single-instanced enclaves.  As such, OCALLs are not as straightforward as this implies.   However, at present anyway, (per @HernanGatta) an OCALL is always dispatched to the host session that did the ECALL that the OCALL is in the context of, so it's straightforward to know _which_ host would apply for an OCALL, which is why it need not be an argument like the enclave is to an ECALL.  However, I think this means the OCALL logic has to be similar to the ECALL logic you show below.

docs/DesignDocs/TEECallsRefactoring.md
anakrish: @mingweishih Can you move the "multiple EDL files" per enclave requirement to next iteration of this feature? Allowing multiple EDL files per enclave makes the import feature redundant; and such a proposal would likely require multiple releases to accomplish.
If you remove that requirement, then the scope is reduced and I feel it would be easier to get consensus.

anakrish: > I would think from the other way: the SDK internal use of EDLs does not entirely use import.

@mingweishih The current plan is to require the user to explicitly import the system EDLs so that the "ghost ocalls" are explicitly opted-in.

This PR says that the user does not have to import system EDLs, rather the user can just use the system edls in the enclave without importing them. 
It is not clear what the user experience is when explicitly opting in to ghost ocalls.
Hence, my suggestion would be to amend the proposal to have minimal impact on existing features, arrive at consensus and propose a bigger change as a separate proposal.

anakrish: Here is the accepted system ocall opting in/out design doc:
https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/system_ocall_opt_in.md

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+              _function_hash,
+              &_ecall_id);
+          oe_set_ecall_function_id(enclave, _caching_table_id, _ecall_id);
+      }
+     
+      /* Call enclave function. */
+      if ((_result = oe_call_enclave_function(
+             enclave,
+            (uint64_t) _ecall_id,
+            (uint64_t) _function_hash,
```
CodeMonkeyLeet, line 503: I think I missed something: why does this take `_function_hash` if it already has the `_ecall_id`?

docs/DesignDocs/TEECallsRefactoring.md
CodeMonkeyLeet: I'm not quite sure why there needs to be separate calls for this. Minimizing roundtrips across the SGX context switch is always beneficial for performance, and it looks like the enclave has all the information it needs to make a function call based on the `_function_hash` already. Instead of making it a separate ECALL, why not take a pointer to `_ecall_id` in `oe_call_enclave_function()`: if provided, it can call the function directly, otherwise, if null, it returns the `_ecall_id` so that the host can cache it for next use.

docs/DesignDocs/TEECallsRefactoring.md
CodeMonkeyLeet: Using undocumented features makes me very nervous on the Windows side. Effectively, that linked solution needed to be tweaked to avoid linker optimizations from removing the direct addition to the MSVC owned constructor section. Can we find a more robust way to deal with this?

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,519 @@
...
+```
+static oe_ecall_func_t __oe_ecalls_table[] = {
+    (oe_ecall_func_t) ecall_oe_verify_report_ecall,
+    (oe_ecall_func_t) ecall_oe_get_public_key_ecall,
+    (oe_ecall_func_t) ecall_oe_get_public_key_by_policy_ecall,
+    (oe_ecall_func_t) ecall_oe_log_init_ecall
+};
+
+static uint32_t __oe_ecalls_hash_list[] = {
+    3802944385 /* CRC32 of oe_verify_report_ecall. */,
```
CodeMonkeyLeet, line 352: Why use CRC32 here? Is it just calculated over the function name?

Your odds of hitting a natural collision in a 32-bit space over just ~10k unique values is non-trivial, so if that is the intended mechanism for uniqueness in a consolidated single table, then you should probably use a 64-bit value. GUIDs have long been used for non-string identification purposes (see OP-TEE's TA identifiers or COM GUIDs) for global uniqueness.

docs/DesignDocs/TEECallsRefactoring.md
CodeMonkeyLeet: Does this proposal solve the problem for hosts with multiple enclaves using overlapping sets of EDL files? i.e. is the only blocker today the function tables or do duplicate struct and type definitions prevent this as well?

docs/DesignDocs/TEECallsRefactoring.md
anakrish: @CodeMonkeyLeet 
Yes, you can import the same edl multiple times and it behaves as if the imported item (the whole edl or a subset of it) has been imported only once.

mingweishih: Registering each function by hash (with the use of `weak symbol`) effectively avoids the duplications regardless how the same ECALL is included (e.g., directly defined in an EDL or imported).  That is, the function with the same hash can only be registered once. One issue remains is that two overlapped function with the same name but different implementation. However, I think this case does not make much sense in C as opposed C++.

The complication of whole problem actually comes from the fact that we support `import` in EDL. This means that how an EDL is used is not deterministic, which is also why using per EDL table is insufficient.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: As I mentioned in another thread, the problem of addressing overlapping sets of EDL files is the main reason for this proposal to registering each function by hash. And this is why the existing implementation of per-EDL table is insufficient.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: This is a great suggestion. Thanks.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: How about CRC64 or MD5 (truncated)? The hash needs to be deterministic so that oeedger8r can generate the same hash for a function in both `_t.c` and `_u.c` every time. I think GUIDs is randomly generated.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -300,218 +282,397 @@ else
...
+ECALL table is not determinstic (i.e., can be any order based on how
+EDL files are imported).
+
+To solve the problem, this phase proposes to using a single table
+that stores the function pointer along with its hash (e.g., the hash of the function name).
+Take the previous example with two enclaves, the ECALL tables now beome:
+
+- `foo` ECALL table
+  ```
+  oe_ecall_struct_t __oe_ecalls_table[] = {
```
anakrish, line 418: As stated above, the problem happens only on the host side. 
> Since the host can have only one implementation of `ecall_common_1_ecall` (the wrapper function),
initialization. both invocations end up using the same function id, say `0`. However, this causes the mismatch on the call into the enclave `bar` where the expected id should be `2`.

Therefore it is not fully justified why the enclave side table needs to change.

Consider this scheme  (not optimized; but a good starting point) where the mapping between the global ecall function identifier (say the function name) to the local function identifier (the ecall table index) is done entirey on the hsot side:
```
   1. Each oe_enclave_t maintains an array of ecall names.
   2. To invoke an ecall, the function name is passed in.
   3. Based on the function name, the appropriate index into the ecall table can be found out
   4. Use the index (local id) to dispatch the ecall.
```
This shows that the mapping can be done entirely on the host side. It then becomes a matter of doing the mapping efficiently.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -300,218 +282,397 @@ else
...
-             _input_buffer_size,
-             _output_buffer,
-             _output_buffer_size,
-             &_output_bytes_written)) != OE_OK)
-            goto done;
+**Register Functions**
 
-        ...
-     }
+In this phase, we keep the use of `_u_wrapper.c`/`_t_wrapper.c` that defines the registration
```
anakrish, line 440: Currently we don't have a way to ensure precise order of initialization. It is not possible with standard C/C++.
Hence if we can avoid relying on order of initialization, it would be better.

> these functnios will be invoked during the enclave initialization
(part of `_handle_init_enclave()`)

Are we going to define a new ELF section for these initialization functions?

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -300,218 +282,397 @@ else
...
+      void* output_buffer,
+      size_t output_buffer_size,
+      size_t* output_bytes_written)
+  {
+    ...
+    uint64_t function_id = OE_ECALL_ID_NULL;
+    static oe_mutex _ecall_lock = OE_H_MUTEX_INITIALIZER;
+
+    ...
+    oe_mutex_lock(&_ecall_lock);
```
anakrish, line 572: Consider if all these can be abstracted out into an infrastructure function `oe_get_ecall_function_id`
Something like...
```c
    static uint32_t global_function_id = (uint32_t) -1;
    uint32_t function_id = 0;

    if (oe_get_ecall_function_id(enclave, hash, &global_function_id, &function_id) != OE_OK)
         goto done;
```

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -300,218 +282,397 @@ else
...
+      (oe_ecall_func_t) ecall_common_2_ecall_1,
+      (oe_ecall_func_t) ecall_common_2_ecall_2,
+      (oe_ecall_func_t) ecall_common_1_ecall,
+      (oe_ecall_func_t) ecall_bar_ecall
+  };
+  ```
 
-**Maintain the List of Registration Functions**
+In addition, both `foo_u.c` and `bar_u.c` will implement same wrapper functions
+of the three imported functions. Assuming that we use *weak symbol* to avoid the
```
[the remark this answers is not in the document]
mingweishih: I'm not sure if `--use-prefix` solves the problem, unless we agrees that every ECALL ends up having prefix by default. Also, two different enclaves may not be able to use the same EDL name, which should be documented as well.

Just for the reference, I also found the the MSVC linker approach is used in [compile-rt](https://github.com/llvm-mirror/compiler-rt/blob/69445f095c22aac2388f939bedebf224a6efcdaf/lib/sanitizer_common/sanitizer_win_defs.h#L68).

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: Another goal of this design is ultimately to remove the needs of  `_t_wrapper.c`/`_u_wrapper.c`; that is, the enclave don't need to maintain multiple ECALL tables and to have host passing the hard-coded table id.  However, if the end goal is making all the internal EDL files optional (may the case for OCALLs, but how about ECALLs?), I'd agree that we only need host-side modifications.

Another advantage of using hash is its potential of extending it to do the runtime checks (e.g., type checks), which cannot be done by simply passing id.

Other than the those, the host side mapping can surely can done similarly to this proposal and provides the same efficiency. Another advantage is removing the need of the special ECALL for querying the ECALL id.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: No. As what we did today for registering table for internal EDLs, these newly added registration functions of internal EDLs will be invoked by the host.

For the external EDL (will have a single table), the table will be passed alone with the `oe_create_enclave`, like what we did for the ocall table today.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: I've updated the proposal to scope down and only make host-side modifications.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -300,218 +282,397 @@ else
...
+    trusted
+    {
+        public void bar_ecall();
+    };
 }
 ```
 
-Each table entry is then defined as the following struct.
+Note that both the EDL files import the two SDK-provided EDL files but in
+different order. The resulting ECALL tables defeind in the
```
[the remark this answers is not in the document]
mingweishih: Added.

mingweishih:
Link to the PoC: https://github.com/openenclave/oeedger8r-cpp/pull/24

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: Remove the use of constructor.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,596 @@
+Generalize the Calling Mechanisms of TEE Functions
```
anakrish, line 1: Does this title still apply? Also the PR title seems a bit broader than the new scope of this proposal.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,596 @@
...
+Unlike typical function invocations via `call` instructions, TEE functions require
+specialized calling mechanisms that allow an enclave and a host to interact with each other.
+In the Open Enclave SDK, we use OCALLs to refer to the host-side functions for the
+enclave to use and use ECALLs to refer to the enclave-side functions for the host to use.
+The invocation of an OCALL/ECALL is analogous to message passing between two isolated entities via a channel.
+Regardless of the type of a channel, which depends on the implementation of TEEs, invoking an ECALL/OCALL requires
+a contract between the enclave and the host so that they know what messages to send and how to
+interpret the messages they receive and to take actions accordingly.
+
+This document describes the current implementation of ECALL/OCALL invocations, its limitations, and the
```
anakrish, line 16: OCALL can be omitted?

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,596 @@
...
+    args_ptr->input_buffer,
+    args_ptr->input_buffer_size,
+    args_ptr->output_buffer,
+    args_ptr->output_buffer_size,
+    &args_ptr->output_bytes_written);
+```
+
+**Internal EDL**
+
+To use EDL files internally, the OE SDK has incorporated a workaround as follows.
```
anakrish, line 186: Not clear why a workaround is needed.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,596 @@
...
+    if (!ecall_table.ecalls)
+        OE_RAISE(OE_NOT_FOUND);
+}
+```
+
+Limitation
+---------
+
+The current implementations pose a limitation to the scenario where the
+host instantiates different enclaves that use the same set of ECALLs
```
anakrish, line 289: Given the problem, there does not seem to be much value in the description of the ocall dispatching mechanism.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,596 @@
...
+{
+    trusted
+    {
+        public void common_2_ecall_1();
+        public void common_2_ecall_2();
+    };
+}
+```
+
+Consider two enclaves, `boo` and `bar`. Their EDL files are defined as follows.
```
anakrish, line 320: typo 'foo' and 'bar'

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: Updated title.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: Removed OCALL.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: Removed the ocall section.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: Done. Good catch!

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: The use of __attribute__((constructor)) has been removed.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: The proposal is scoped down to solve the problem in the single-host-multiple-enclaves scenario.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: The proposal uses the name as an identifier now.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: Removed.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,436 @@
+Generalize the Calling Mechanisms of Enclavae Functions
```
dthaler, line 1:
```suggestion
Generalize the Calling Mechanisms of Enclave Functions
```

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,436 @@
...
+
+At this point, an ECALL is successfully dispatched to the enclave.
+
+Limitation
+---------
+
+The current implementations pose a limitation to the scenario where the
+host instantiates different enclaves that use the same set of ECALLs
+(i.e., importing the same EDL files).
+Taking the following case study for example.
```
dthaler, line 131:
```suggestion
Take the following case study for example.
```
grammar fix

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,436 @@
...
+
+  The local table allows the host to retrieve the ECALL id per enclave instead
+  of using the hard-coded id in the same ECALL across different enclaves.
+
+- Global Table
+  To guarantee the constant-time look-up, the host needs to use the same id to retrieve
+  the ECALL id from the local table. To do so, the host additionally maintains a global
+  table that keeps such ids to which we refer as global ids. The detailed design is as follows.
+
+  - The global table is defined internally in the `oehost` as follows. The table stores name
```
dthaler, line 321:
```suggestion
  - The global table is defined internally in the `oehost` as follows. The table stores the name
```

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,436 @@
...
+     uint32_t ecall_id_table_size;
+     ...
+  } oe_enclave_t;
+  ```
+
+  The local table allows the host to retrieve the ECALL id per enclave instead
+  of using the hard-coded id in the same ECALL across different enclaves.
+
+- Global Table
+  To guarantee the constant-time look-up, the host needs to use the same id to retrieve
```
dthaler, line 317: The local tables above make perfect sense to me.  I still don't follow why a global table or global ids are needed at all though.   You already have constant time lookup with the local table, right?

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,436 @@
...
+  ```
+- `bar`
+  ```
+  enclave->ecall_id_table[0]: { 2 /* ecall_common_1_ecall */ }
+  enclave->ecall_id_table[1]: { 0 /* ecall_common_2_ecall_1 */ }
+  enclave->ecall_id_table[2]: { 1 /* ecall_common_2_ecall_2 */ }
+  enclave->ecall_id_table[3]: { OE_ECALL_ID_NULL /* ecall_foo_ecall */ }
+  enclave->ecall_id_table[4]: { 3 /* ecall_bar_ecall */ }
+  ```
+As we can see, the scheme allows the host to use the same global id to look up different ECALL id
```
dthaler, line 430: I didn't see why we need a global id at all.  Above said "both `foo_u.c` and `bar_u.c` will implement same wrapper functions of the three imported functions" which seems to be the issue, but you can solve that by edger8r prefixing the wrappers with foo and bar respectively and then I think you don't need a global id and the solution is simpler.  Or am I missing something?

mingweishih: The downside of prefixing is that it would apply to every ecalls. This complicates the user experience; i.e., users need to add the prefix for both system ecalls (e.g., `foo_oe_log_init_ecall`, `bar_oe_log_init_ecall`) and their own ecalls (e.g., `foo_ecall_foo`, `bar_ecall_bar`).

This proposed design is to avoid making the prefix as a hard requirement when importing system edls.

@anakrish can you also comment on this?

anakrish: Yes, the user calls the enclave functions via the host-side wrapper functions. 
Given
```c
foo.edl

trusted {
   void ecall1();
};
```

Currently, the user invokes the ecall via:
```c
ecall1(enclave_handle);
```
Adding a prefix to the untrusted wrapper would change the user experience, and the user would have to make the ecall via:
```c
foo_ecall1(enclave_handle);
```

Intel SDK makes prefixing a user-choice via the `--use-prefix` edger8r option. Therefore the user can decide whether to use prefix or not (to avoid conflict), by passing --use-prefix to conflicting edls.

However the prefixing strategy does not work for system edls or user-written libraries.
Consider a reusable library (user_lib.a), that internally makes some ecalls. The library has an associated edl user_lib.edl:

```c
user_lib.edl:
trusted {
   void lib_ecall1();    // Function index is 0
};

user_lib.c :
void lib_function(enclave_handle h)
{
   ... do some stuff
   // call enclave function necessary for this library
   lib_ecall1(h);   <----- what prefix should this call use?
   // As written, the above wrapper will use index 0.
}
```
The redistributable library has to make the ecall at a specific location in code, and therefore it has to lockdown the prefix (which may be empty).

Now consider two enclaves that reuse this library:
```c

enclave1.edl
{
    import "user_lib.edl"; // lib_ecall1 has index zero
};

enclave2.edl
{
   import "something_else.edl";
   import "user_lib.edl";   <--- the function index for lib_ecall1 is different due to different imports.
}
```
Even if you make different prefixes for enclave1 and enclave2, the user_lib.a library has already chosen a specific prefix (in this case empty).
Therefore when the library is invoked via lib_function, passing in either enclave1 or enclave2 handles, the call to the wrapper `lib_ecall1` may fail since it may use a wrong index.

This is the scenario where prefixes wont work. If the ecall happens in a reusable library or the SDK itself, the call-site is bound to the prefix/index.

anakrish: @mingweishih @dthaler
I have created reproduction steps for the Intel SDK here:
https://github.com/anakrish/imported-ecall-issue

mingweishih: @dthaler Does the explanation above make sense to you?

mingweishih: @dthaler [logging.edl](https://github.com/openenclave/openenclave/blob/master/include/openenclave/edl/logging.edl) is another example that the prefix won't work.

 This system edl defines an ecall `oe_log_init_ecall`, which is used by the oehost library as follows.
https://github.com/openenclave/openenclave/blob/2a3da923e893e4263637f27983deb1016c7f8710/host/sgx/create.c#L1011

If user applies the prefix, say `foo`, then the oehost cannot correctly invoke the oeedger8r-generated implementation (in this case `foo_oe_log_init_ecall`). Instead, the default implementation will be used, which returns `OE_UNSUPPORTED`.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,436 @@
...
+  The local table allows the host to retrieve the ECALL id per enclave instead
+  of using the hard-coded id in the same ECALL across different enclaves.
+
+- Global Table
+  To guarantee the constant-time look-up, the host needs to use the same id to retrieve
+  the ECALL id from the local table. To do so, the host additionally maintains a global
+  table that keeps such ids to which we refer as global ids. The detailed design is as follows.
+
+  - The global table is defined internally in the `oehost` as follows. The table stores name
+    of each ECALL and the index to each name is used as the global id.
```
dthaler, line 322: Why do you need both an index and a name?   Seems like you should be able to use one or the other without needing both.  You just need one predictable (not dynamic) global identifier (which could just be the name itself) that you can use as a key to look up the local ID in the host's per-enclave table.  That global identifier can be hard-coded be in the edger8r generated code.  And you don't need a global table then either, at least that I understand so far.

docs/DesignDocs/TEECallsRefactoring.md
dthaler: Yes that makes sense to me, thanks!  See updated comment.

docs/DesignDocs/TEECallsRefactoring.md
mingweishih: The design is that the host stores the name only with the array
`const char* _ecall_table[OE_ECALL_MAX];`
and the index of each name in the array is used as global indices. 

Although the name (or any hard-coded global identifier) is sufficient to loop up the local id, we cannot guarantee the constant-time operations. Besides, there is no good data structures (e.g., hash map) in C unless we implement one.

The purpose of the global table is therefore keeping things simple: Keeping track of which ecall has been registered across multiple enclaves. Therefore, the host can ensure that the ecall with the same identifier will have the same index in the local table.

This can be illustrated by the example in this proposal:
	- `_ecall_table`
  ```
  _ecall_table[0]: { "ecall_common_1_ecall" }
  _ecall_table[1]: { "ecall_common_2_ecall_1" }
  _ecall_table[2]: { "ecall_common_2_ecall_2" }
  _ecall_table[3]: { "ecall_foo_ecall" }
  _ecall_table[4]: { "ecall_bar_ecall" }
  ```
- `foo` (at the time `bar` is not initialized yet)
  ```
  enclave->ecall_id_table[0]: { 0 /* ecall_common_1_ecall */ }
  enclave->ecall_id_table[1]: { 1 /* ecall_common_2_ecall_1 */ }
  enclave->ecall_id_table[2]: { 2 /* ecall_common_2_ecall_2 */ }
  enclave->ecall_id_table[3]: { 3 /* ecall_foo_ecall */ }
  ```
- `bar`
  ```
  enclave->ecall_id_table[0]: { 2 /* ecall_common_1_ecall */ }
  enclave->ecall_id_table[1]: { 0 /* ecall_common_2_ecall_1 */ }
  enclave->ecall_id_table[2]: { 1 /* ecall_common_2_ecall_2 */ }
  enclave->ecall_id_table[3]: { OE_ECALL_ID_NULL /* ecall_foo_ecall */ }
  enclave->ecall_id_table[4]: { 3 /* ecall_bar_ecall */ }
  ```

When host initializes the local table of `bar`, it basically replicates the global table such that the same global index can be used to locate the same ecall on different local tables. For example, the index of `ecall_common_1_ecall` is `0`. The global index `0` can be used to look up the local id on both `foo` (get `0`) and `bar` (get `2`).

CodeMonkeyLeet:
As discussed in the [SIG-Arch meeting](https://hackmd.io/KOeBDVlyQpOC-OqqHYvs0A#23-June-2020):

- @dthaler & @mingweishih to resolve the remaining comment by explaining the use of the static variable for caching the ecall IDs for subsequent invocations, and this design proposal should be good to go. 
- Review can move to implementation details in [oeedger8r-cpp PR #24 ](https://github.com/openenclave/oeedger8r-cpp/pull/24).

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,442 @@
...
+        index = the index of the name
+      else /* not found */
+        add the name to the table
+        index = the index of the name
+
+      /* Assign the ECALL id to the local table based on the global id. */
+      enclave->ecall_id_table[index] = i;
+    }
+    ```
+  - Modify the internal `oe_call_enclave_function` function to pass a global id
```
mingweishih, line 385: @dthaler  Please refer to here for the use of static variable to cache the global id after the first invocation of an ecall.

docs/DesignDocs/TEECallsRefactoring.md
```diff
@@ -0,0 +1,442 @@
...
+- `bar`
+  ```
+  enclave->ecall_id_table[0]: { 2 /* ecall_common_1_ecall */ }
+  enclave->ecall_id_table[1]: { 0 /* ecall_common_2_ecall_1 */ }
+  enclave->ecall_id_table[2]: { 1 /* ecall_common_2_ecall_2 */ }
+  enclave->ecall_id_table[3]: { OE_ECALL_ID_NULL /* ecall_foo_ecall */ }
+  enclave->ecall_id_table[4]: { 3 /* ecall_bar_ecall */ }
+  ```
+
+In the oeedger8r-generated stub code of `ecall_common_1_ecall`, the local static variable
```
mingweishih, line 433: @dthaler  Please refer to here for an example.

mingweishih:
@dthaler I've updated the document to explain the use of caching the global id (after the first invocation of an ecall) with a static variable. Please take a look.  Please free to comment on the implementation detail once we move to the code review stage.

Issue #2999, opened 2020-05-06
oe_verify_evidence should accept SGX quotes as input


LarryOsterman:
Currently oe_verify_evidence can only work with the output of oe_generate_evidence, which means it cannot be used to validate SGX collateral. oe_verify_evidence should be changed to accept collateral from known TEEs instead of must oe_generate_evidence.

yentsanglee:
Currently
- oe_verify_evidence() verifies evidence from oe_get_evidence()
- oe_verify_report() verifies report from oe_get_report()

It helps if the legacy format from oe_get_report() and the new format from oe_get_evidence() can be unified and oe_verify_evidence(), which is the promoted API, can verify both formats.

CodeMonkeyLeet:
@LarryOsterman, to clarify the requirement, are you asking for:
- a migration path between the `oe_get_report()` and `oe_get_evidence()` APIs, or
- `oe_verify_evidence()` to be able to handle raw TEE evidence not conforming to the data contract specified by the attestation plugin model?
    - For example, in the case of SGX, the desire is for `oe_verify_evidence()` to process a raw SGX quote produced by the Intel SGX SDK.

LarryOsterman:
I have two possible inputs to my service:

- An OpenEnclave report from an SGX enclave.

- An SGX quote generated by an Intel SGX quoting enclave

I would like to have access to the claims returned by oe_verify_evidence, specifically the claims which include the state of the quote and the validity lifetime of the quote.

In order to do that, there needs to be a mechanism where either (a) oe_verify_evidence can accept either an OpenEnclave report or an SgX quote (there is a trivial transformation between those two, so it doesn't matter which of the two it is) or (b) some other function that can accept an OpenEnclave report or sgx function which returns the information currently returned by oe_verify_evidence.

That API doesn't need to be oe_verify_evidence, I just need the information that oe_verify_evidence returns.

dthaler:
PR #2959 tracks requirements, and I believe it already covers all of Larry's requirements listed here.
It also describes proposal abstract APIs without dictating what the actual concrete APIs look like.

yentsanglee:
For backward compatibility, it will be great if oe_verify_evidence() or new verifying API can also verify the reports generated by oe_get_report().

radhikaj:
Bring up in sig-attestation. @radhikaj  to add to agenda

yentsanglee:
Resolved by PR #3441 and other PRs.

PR #3089, opened 2020-06-03
Add attestation evidence format support doc.

qiucwang:
Propose a new verify API `oe_verify_evidence_by_format_id` for attestation evidence format support requirement raised by issue #2999.


docs/DesignDocs/AttestationEvidenceFormatSupport.md
```diff
@@ -0,0 +1,91 @@
...
+2. Get an evidence and its format id from attester.
+3. Call the new API with the evidence and the format id to verify the evidence.
+4. The new API returns attestation results and claim set.
+
+
+Specification
+-------------
+
+### New API
+The new API is similar to current function oe_verify_evidence() except the extra `format_id` parameter.
```
dthaler, line 32: Can we just replace oe_verify_evidence and keep the same name?

docs/DesignDocs/AttestationEvidenceFormatSupport.md
```diff
@@ -0,0 +1,91 @@
...
+
+```
+#define OE_CLAIM_REPORT_DATA "report_data"
+```
+
+Alternates
+----------
+We can also develop APIs that convert a raw evidence to and from an oe_evidence.
+
+### convert_to_evidence().
```
dthaler, line 66: My opinion: I don't think we need this API, we can abandon the oe_evidence going forward.

docs/DesignDocs/AttestationEvidenceFormatSupport.md
```diff
@@ -0,0 +1,91 @@
...
+#### Outputs:
+- Evidence format ID
+- Raw evidence buffer
+- Raw evidence buffer size
+
+The converion itself needs extra memory allocation and copy.
+
+Authors
+-------
+Qiucheng Wang (qiucwang.microsoft.com)
```
dthaler, line 90:
```suggestion
Qiucheng Wang (qiucwang@microsoft.com)
```

docs/DesignDocs/AttestationEvidenceFormatSupport.md
```diff
@@ -0,0 +1,91 @@
+Attestation Evidence Format Support
+=====
+
+One of the goal of OE SDK attestation design is to be TEE-agnostic. At the same time, OE SDK should enable a verifier application to verify evidences from different sources and TEEs, if the format is known to the verifier.
+
+Motivation
+----------
+
+Current evidence verify API `oe_verify_evidence()` verifies the evidence gerenated by `oe_get_evidence()` and returns the claims contained in the evidence. Such evidence has a specific format including an `oe_attestation_header`.
```
dcmiddle, line 9: s/gerenated/generated

shnwc:
For SGX ECDSA quote (as in Larry’s original request), there are 3 types of evidence structures: 
- The raw SGX quote structure returned by the Intel SGX SDK
  - This quote structure is defined as type `sgx_quote_t`, in OE SDK in https://github.com/openenclave/openenclave/blob/master/include/openenclave/bits/sgx/sgxtypes.h
  - It is also defined as type `sgx_quote3_t` in Intel DCAP SDK in https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/quote_wrapper/common/inc/sgx_quote_3.h 
  - This structure is a fixed-length header followed by a variable-length signature field, with the size of this field contained in the header:
    - `uint16_t version;`
    - `uint16_t sign_type;` -- an enum type with values 0, 1, 2, or 3.
    - …
    - `uint32_t signature_len;`
    - `OE_ZERO_SIZED_ARRAY uint8_t signature[];`
- The OE report structure returned by `oe_get_report()`
  - This OE report structure is the raw SGX quote prefixed with an `oe_report_header_t` structure, defined in https://github.com/openenclave/openenclave/blob/master/include/openenclave/internal/report.h 
  - The header `oe_report_header_t` contains these 3 fields:
    - `uint32_t version;`
    - `oe_report_type_t report_type;` -- an enum type with values 1 or 2.
    - uint64_t report_size; -- number of bytes the raw SGX quote has.
- The OE evidence structure returned by `oe_get_evidence()`
  - This OE evidence structure is the report structure prefixed with an `oe_attestation_header_t` structure and suffixed with serialized custom claims, as defined in https://github.com/openenclave/openenclave/blob/master/common/attest_plugin.h
  - The header `oe_attestation_header_t` contains these 3 fields:
    - `uint32_t version;`
    - `oe_uuid_t format_id;` -- evidence format UUID
    - `uint64_t data_size;` -- number of bytes in the OE report + the serialized custom claims

Based on the above information, I think the implementation of `oe_verify_evidence_by_format_id()` can tell whether the input evidence buffer contains a raw SGX quote, an OE report structure, or an OE evidence structure. It can tell by checking if the evidence buffer has the right amount of data for one of the structure types.

If so, the proposed API `oe_verify_evidence_by_format_id()` can take one of the three evidence structures for SGX ECDSA quote in `evidence_buffer`, and does not need an additional flag other than the format_id to process data in `evidence_buffer`. In fact, the `format_id` parameter could be ‘NULL”, in that case, the `evidence_buffer` must contain an OE evidence structure. This is what the current API `oe_verify_evidence()` implements.

yentsanglee:
I think format ids for evidence buffer with sgx_quote_t and oe_report_header_t have to be defined. Callers to oe_verify_evidence_by_format_id() have to specify the format of the evidence buffer. We should not attempt to guess by scanning the evidence buffer. That's why in the "New format ID" section, we suggested

. OE_FORMAT_UUID_SGX_ECDSA_P256 for evidence format of sgx_quote_t
. OE_FORMAT_UUID_OE_REPORT for evidence format of oe_report_header_t
. No UUID for evidence format of oe_attestation_header_t because format id is already embedded in the buffer.

---
In reply to: [638478879](https://github.com/openenclave/openenclave/pull/3089#issuecomment-638478879) [](ancestors = 638478879)

shnwc:
Either way seem will work fine: to scan the evidence buffer to detect evidence structure type or to be informed by an input type / flag. I think it would be good to document the two options and the reasons why one option is chosen over the other.

When an input type / flag is used, as discussed in today's SIG-attestation meeting, instead of having multiple `format_id` values for the different evidence header structures containing the same SGX ECDSA quote, an alternative is to have an optional second input parameter like `header_type` to inform the evidence header structure type.

qiucwang:
Doing guess could be unreliable. So we proposed this format id to identify each evidence structure.
For users who full rely in OESDK ecosystem or want to add custom claims, they can use oe_generate_evidence to generate an oe_evidence structure and verify it by oe_verify_evidence. No format id needed.
For users who would provide their own evidence structures (like #2999, user hopes to verify a raw sgx quote), they should use this new API oe_verify_evidence_by_format_id and provide that id.
Since now raw_sgx_quote, oe_report, oe_evidence are all tied with OE_FORMAT_UUID_SGX_ECDSA_P256 (correct me if I'm wrong), we might need a new format id for oe_report. (In the future, we might no longer need oe_report struct at all). We don't need an id for current oe_evidence because that one should be verified by old oe_verify_evidence API. 
And again, we should avoid guessing, so new API oe_verify_evidence_by_format_id should not handle oe_evidence which is generated by oe_get_evidence (as in the future, oe_evidence might wrap other evidences with other format ids, make guessing even less reliable). That also avoid the situation that user provided format id might not be consistent with the format id in oe_evidence header. oe_evidence is only handled by old API oe_verify_evidence.
I will send another PR to add those details.

docs/DesignDocs/AttestationEvidenceFormatSupport.md
```diff
@@ -0,0 +1,91 @@
...
+
+```
+#define OE_FORMAT_UUID_OE_REPORT { _uuid_ }
+```
+
+### Additional Claims
+Without custom claims, a new `oe_claim` field is introduced to retrieve `report_data` field inside sgx_quote. Attestation service needs this field as a part of their validation process.
+
+```
+#define OE_CLAIM_REPORT_DATA "report_data"
```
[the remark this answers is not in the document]
dthaler: can the field name also be changed to "sgx_report_data"? or does it have to interoperate with a field that already exists with the "report_data" name?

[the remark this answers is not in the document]
qiucwang: Sure, `sgx_report_data` makes sense here.

docs/DesignDocs/AttestationEvidenceFormatSupport.md
```diff
@@ -0,0 +1,118 @@
...
+--------- 				        ----------
+1. SGX Quote 				        a. oe_verify_evidence(SGX_P256, evidence, ...)
+2. oe_get_report() 			        b. oe_verify_evidence(OE_REPORT_ID, evidence, ...)
+3. oe_get_evidence(SGX_P256) 	                c. oe_verify_evidence(null, evidence, ...)
+
+Evidence 1 should be verified by calling function a.
+Evidence 2 should be verified by calling function b.
+Evidence 3 should be verified by calling function c, not a.
+```
+In both scenarios, the plugin_verifier is located by format_id and then verifies the evidence. But the evidence in scenario 2 contains an extra claim buffer which is added by `oe_get_evidence()`. The plugin_verifier needs to do an extra size check to tell the difference. For example, `oe_verify_evidence(SGX_P256, evidence, ...)` will locate the verifier to verify the `evidence`. The verifier needs to check if the evidence size equals to the raw sgx quote size, or `if (evidence_size == ((sgx_quote) evidence).size)`. If it is, then there is no extra claim buffer. The verifier's claim extraction function should take care of that.
```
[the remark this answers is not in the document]
qiucwang: updated.

docs/DesignDocs/AttestationEvidenceFormatSupport.md
```diff
@@ -0,0 +1,118 @@
...
+Current evidence verify API `oe_verify_evidence()` verifies the evidences generated by `oe_get_evidence()` and returns the claims contained in the evidence. Such evidence has a specific format including an `oe_attestation_header`.
+As reported in issue #2999, attestation service has to verify evidences from different sources other than those generated by `oe_get_evidence()`. Such as evidence generated by legacy OE SDK API `oe_get_report()` and raw TEE evidence generation such as SGX quote generation mechanism. And like `oe_verify_evidence()` returns claims, attestation service is also required to return the supplemental information. To achieve that, a new verify API is needed to verify different evidence structures and return the required supplemental information.
+
+Goals:
+- Verifier can call OE SDK API to verify known evidence formats generated by either OE SDK API or other attesters.
+
+User Experience
+---------------
+
+New parameter `format_id` will be added to `oe_verify_evidence` to verify various evidences other than the ones generated by `oe_get_evidence()`, such as a raw SGX quote. To verify an evidence generated by `oe_get_evidence()`, `format_id` should be `null`.
```
achamayou, line 18: Is there a specific reason why the value is `null` for evidence created by `oe_get_evidence()`? It would seem more regular to assign that with its own (named) format_id, and to require format_id every time.

shnwc: I think that when the evidence data is from `oe_get_evidence()`, the `format_id` parameter is optional: can be NULL or a valid format ID (which must match the one in the evidence data).

docs/DesignDocs/AttestationEvidenceFormatSupport.md
```diff
@@ -0,0 +1,118 @@
...
+    size_t policies_size,
+    oe_claim_t** claims,
+    size_t* claims_length);
+```
+This API is used to verify both evidences whose format id is provided by verifier and who is generated by `oe_get_evidence()`. When actual format_id is provided, it decouples evidence format ID and raw evidence buffer. When format_id is `null`, the API acts the same as previous `oe_get_evidence()`. This new design will provide more flexibility.
+
+
+### New Format ID
+According to issue #2999 reported by attestation service, it has to support evidence generated by SGX quote generation and `oe_get_report()`. SGX quote format id is already defined as `OE_FORMAT_UUID_SGX_ECDSA_P256`. A new id needs to be defined for evidence generated by `oe_get_report()`.
+
```
shnwc, line 73: Let's keep the existing `OE_FORMAT_UUID_SGX_ECDSA_P256` for the evidence generated by `oe_get_evidence()`, and define two new ones `OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT` and `OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE`, for `oe_get_report()` output and raw SGX quote respectively.

I'm updating the design doc Attestation_API_Proposal.md and implementation to capture those changes.

docs/DesignDocs/AttestationEvidenceFormatSupport.md
qiucwang: > Is there a specific reason why the value is `null` for evidence created by `oe_get_evidence()`? It would seem more regular to assign that with its own (named) format_id, and to require format_id every time.

Because when verify API reads a `null` format_id, then it knows the evidence buffer is an evidence generated by `oe_get_evidence()` and will unpack the evidence first (remove oe_attestation header and extract the `format_id` in the header). Otherwise, when verify API reads a format_id, there is no solid way to tell whether the evidence buffer is a raw evidence, or an evidence generated by `oe_get_evidence`.

docs/DesignDocs/AttestationEvidenceFormatSupport.md
qiucwang: Will we have 3 different verifiers for these 3 evidences? And in the future, do we have to assigned two IDs for one evidence? One for raw evidence, another for the one generated by `oe_get_evidence()`.

shnwc: I see the need of 3 format IDs. The unpacked `oe_evidence` is `oe_report`, which in turn is SGX quote prefixed with a header.

I've just created PR #3136 with the changes in Attestation_API_Proposal.md.

docs/DesignDocs/AttestationEvidenceFormatSupport.md
```diff
@@ -0,0 +1,118 @@
...
+3. Call the new API with the evidence and the format id to verify the evidence.
+
+Scenario 2 - Verifier verifies an evidence generated by `oe_get_evidence()`:
+
+1. Call OE SDK public API to initialize verifier plugins.
+2. Call the new API to verify the evidence. The format id is not required because it is contained in the evidence buffer.
+
+Scenario 1 is for users who provide their own evidence structures (like #2999, user hopes to verify a raw sgx quote).
+
+Scenario 2 is for users who full rely in OESDK ecosystem. They can use oe_generate_evidence to generate an oe_evidence structure and verify it by oe_verify_evidence.
```
dthaler, line 33:
```suggestion
Scenario 2 is for users who use the existing OESDK everywhere. They can use oe_generate_evidence to generate an oe_evidence structure and verify it by oe_verify_evidence.
```
(Rationale: The OESDK "ecosystem" may include multiple formats, and in particular I hope it will include standards in the future, where the current format returned by oe_generate_evidence runs counter to emerging standards.  Hence softening wording in my suggestion.)

docs/DesignDocs/AttestationEvidenceFormatSupport.md
qiucwang: Well my previous design set same format id for a raw evidence and oe_evidence. But I think your design makes more sense.
Here is a summary:
| Format ID                            | Evidence                                                   |
|--------------------------------------|------------------------------------------------------------|
| `OE_FORMAT_UUID_SGX_ECDSA_P256`        | SGX ECDSA-p256 evidence   generated by `oe_get_evidence()` |
| `OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE`  | raw SGX ECDSA quote                                        |
| `OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT` | SGX ECDSA report generated by `oe_get_report()`            |

If we follow this way, in future, suppose we want to support new evidence formats, such as VBS, we will generate **a pair of ids**
| Format ID                | Evidence                                        |
|--------------------------|-------------------------------------------------|
| `OE_FORMAT_UUID_VBS`       | VBS evidence generated by   `oe_get_evidence()` |
| `OE_FORMAT_UUID_VBS_QUOTE` | raw VBS quote                                   |

Pros:
1. One evidence format strictly maps to one format id and verifier. It makes a cleaner design.
2. Claims extraction can be differentiated. For example, a SGX ECDSA-p256 evidence gets a `OE_CUSTOM_CLAIMS` claim entry, a raw SGX ECDSA quote gets a `OE_CLAIM_SGX_REPORT_DATA` claim entry. With different verifiers, we can call different extraction procedure for each evidence format.

Cons:
1. `oe_verify_evidence` select the plugin verifiers by `format_id` so we need to define two different verifiers for a raw evidence and the oe_evidence containing the raw evidence. We need to double the format id and verifiers even two verifiers do mostly the same thing.
2. Users might get confused about the pair of ids. We need to document it well.

docs/DesignDocs/AttestationEvidenceFormatSupport.md
achamayou: @qiucwang thank you, that makes complete sense, I'd misunderstood what happened in that case.

docs/DesignDocs/AttestationEvidenceFormatSupport.md
qiucwang: Thank, I will update that.

qiucwang:
Now we have 3 formats for SGX ECDSA evidence:
1. SGX ECDSA-p256 evidence generated by oe_get_evidence(); 2. raw SGX ECDSA quote; 3. SGX ECDSA report generated by oe_get_report();

Previously I misunderstood that we would have 3 different plugins to verify each format. That's why I agreed we could have 3 format ids.
Now if we use the same plugin to verify these 3 formats, I think there are 2 things we need to handle.

First, unpack the envelop (is the evidence has one) before pass the evidence to plugin. This one can be solved by checking `format_id == null`. If a `null` format_id is passed in, we assume it's the one with envelop.

Second, return specific claim structure. We need to return different claim structures for 3 different formats.  Quote from PR#3136:
> For evidence data generated by oe_get_evidence(), the list of claims returned by oe_verify_evidence() will contain one claim named OE_CLAIM_CUSTOM_CLAIMS_ID if the evidence data contains a non-empty custom claims buffer. For an SGX ECDSA report or a raw SGX ECDSA quote, the list of claims will contain one claim named OE_CLAIM_SGX_REPORT_DATA_ID.

Current API doesn't decouple custom claim from evidence buffer before passing an evidence to the plugin. The custom claim isn't treated as a part of envelop but a part of evidence. So after evidence is unpacked from the envelop, the plugin still cannot tell whether received evidence contains custom claim or not. But it needs to know. If evidence contains, it needs to return `OE_CLAIM_CUSTOM_CLAIMS_ID`, otherwise, it needs to return  `OE_CLAIM_SGX_REPORT_DATA_ID`.
I can think of 2 ways. 
Either plugin itself compares evidence_buffer_size (actual size of passed in evidence: raw_evidence_size + claim_size) and raw_evidence_size (fetching from parsed evidence header). If they are equal, then there is no custom claim and plugin needs to return OE_CLAIM_SGX_REPORT_DATA_ID. 
Or, we pass extra flags to plugin, telling it whether evidence contains custom claim or not. Then we need to update this API design.

qiucwang:
Besides, current SGX ECDSA-p256 evidence generated by oe_get_evidence() contains 4 layers: 1. oe_attestation_header, which is the envelop, 2.oe_report_header(), 3. sgx_quote; 4 custom claims
The 2nd layers seem serve no purpose here. Do you think we can remove this layer? If we remove oe_report_header from the enveloped evidence, I think we can use same format_id for raw sgx_quote and oe_evidence (the sgx_quote with envelop).
But as per issue #2999, we still need to deal with SGX ECDSA-p256 report generated by oe_get_report() as an evidence, we can have an extra format id to tell plugin, this is a special format and an oe_report_header needs to be unpacked before verify this evidence.

shnwc:
@qiucwang there will indeed be 3 SGX evidence verifier plugins, one for each of the 3 UUIDs. Implementation of `oe_verify_evidence()` will get a UUID from combination of `format_id` and `evidence_buffer`, use the UUID to search of a plugin, and invoke the chosen plugin to verify the evidence.

To get the UUID, `oe_verify_evidence()` will use the one in `format_id` if it's not NULL, otherwise, `evidence_buffer` must have an `oe_attestation_header` and the UUID from that header will be used.

yentsanglee:
> @qiucwang there will indeed be 3 SGX evidence verifier plugins, one for each of the 3 UUIDs.

Which 3 UUIDs did you refer to?

shnwc:
OE_FORMAT_UUID_SGX_ECDSA_P256, OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT, and OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE

yentsanglee:
According to [SIG-Attestation meeting](https://hackmd.io/Xj6GpDSKSwuz5cZgQ0yg1A?view) this morning:

"Three alternatives to get evidence with or without an envelope:
  1. Call a separate function to include the envelope
  2. Use a different UUID to state whether you get an envelop
  3. Pass in a flag to getevidence() to say whether it should include the envelope"

If we pick the 3rd option, we only need OE_FORMAT_UUID_SGX_ECDSA_P256 and OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT.

shnwc:
I prefer alternative 2 for evidence generation, and see that it allows attesters to stay TEE-agnostic. A verifier can request different envelop configurations by providing a different UUID list to an attester. 

As described in PR #3136, there are 3 UUIDs for verification SGX ECDSA evidence with different envelop configurations. Implementation of this approach has been straight forward. Support of these UUIDs can be added to `oe_get_evidence()` with a similarly easy implementation.

yentsanglee:
There is a problem with option 2. Let's say I develop a plugin which I define as ID_XYZ. Whose responsibility is it to define ID_XYZ_WITH_ENVELOPE? If I define both IDs, how do I handle ID_XYZ_WITH_ENVELOPE?

shnwc:
@yentsanglee in your example, there will be two plugins for evidence generation and two plugs for evidence verification: one for ID_XYZ and the other one for ID_XYZ_WITH_ENVELOPE. The plugins with be responsible to wrap / unwrap all the headers.

I've created a new PR #3168 for the changes in Attestation_API_Proposal.md as the implementation makes progress.

yentsanglee:
> @yentsanglee in your example, there will be two plugins for evidence generation and two plugs for evidence verification: one for ID_XYZ and the other one for ID_XYZ_WITH_ENVELOPE. The plugins with be responsible to wrap / unwrap all the headers.

The header wrap/unwrap code should be universal and all formats with envelope are defined the same way. We should not put the burden on plugin developers. In the SIG-attestation most people favored reducing format IDs (option 1 or 3).

shnwc:
I'm aware that multiple participants in the meeting favored reducing the number of format IDs, and agree that new IDs should not be introduced unless necessary. But when end-to-end use cases are considered, it could make more sense to add a new ID value instead of adding an new `has_header` parameter to `oe_get_evidence()` and `oe_verify_evidence()`.

This `has_header` parameter, when introduced, would need to be agreed / negotiated between a verifier and an attester. It effectively become integral part of the evidence format identity.

In addition, for SGX, the two format IDs, OE_FORMAT_UUID_SGX_ECDSA_P256 and OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT, do not just differ between whether a header (which contains a format ID) is wrapped or not around the evidence data, they also differ between whether the custom claims are hashed (so can have arbitrary size) or directly placed in the SGX report data field (so limited to no more than 64 bytes).

The complication of SGX evidence support is due to its support in legacy `oe_get_report()` / `oe_verify_reporet()`. Other TEEs don't have this legacy, and can avoid similar complication.

For header wrap/unwrap, agreed that ideally it should done by the plugin framework code. At the same time, we want to keep the framework code TEE-agnostic. For SGX, `oe_verify_evidence()` already has to take in 3 types of evidence data: with attestation + report headers, with only report header, or without any header. To keep the framework code TEE-agnostic, all the header handling will need to be done inside the plugin (unless a new `has_header` parameter is introduced).

One way to allow the framework code control how the header looks like while requiring plugins actually fill the header when needed is for the former to expose header-processing helper functions for invocation by plugins.

Since the attestation design is still experimental, and now we have format ID in both `oe_get_evidence()` and `oe_verify_evidence()`, I think we can safely drop the attestation header from the evidence data, and require `oe_verify_evidence()` to always get a valid `format_id` parameter instead.

Let's discuss in the SIG-attestation meeting.

docs/DesignDocs/AttestationEvidenceFormatSupport.md
qiucwang: I've updated based on our discussion. Two new format ids are introduced: "OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE" and "OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT".

PR #3090, opened 2020-06-03
Reduce committed SGX enclave size on Windows

CodeMonkeyLeet:
`oe_sgx_create_enclave()` currently only takes the enclave size as defined by its
ELRANGE, which can be significantly larger than the actual EPC memory that
needs to be committed to load and launch the enclave.

This change updates `oe_sgx_create_enclave()` to take both an initial commit size
in addition to the enclave ELRANGE size, which should expand the range of
enclave sizes that can successfully load on Windows (see previous issues such
as #2394 and #2397).

### DIFF
--- a/host/sgx/create.c
+++ b/host/sgx/create.c
@@ -805,7 +805,8 @@ oe_result_t oe_sgx_build_enclave(
         image_size, &props, &enclave_end, &enclave_size));
 
     /* Perform the ECREATE operation */
-    OE_CHECK(oe_sgx_create_enclave(context, enclave_size, &enclave_addr));
+    OE_CHECK(oe_sgx_create_enclave(
+        context, enclave_size, enclave_end, &enclave_addr));
 
     /* Save the enclave base address, size, and text address */
     enclave->addr = enclave_addr;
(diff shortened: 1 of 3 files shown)


CodeMonkeyLeet:
bors try

anakrish:
> This change updates oe_sgx_create_enclave() to take both an initial commit size
in addition to the enclave ELRANGE size, which should expand the range of
enclave sizes that can successfully load on Windows

Can this be locked down via a test?

CodeMonkeyLeet:
> Can the scenario be locked down via a test?

I'm open to suggestions here; I've done a manual verification in terms of figuring out a set of config values that don't work prior do the change, but do afterwards, the problem is that the threshold is dependent on the available EPC size of the target test device, so that makes it hard to automate.

anakrish:
> > Can the scenario be locked down via a test?
> 
> I'm open to suggestions here; I've done a manual verification in terms of figuring out a set of config values that don't work prior do the change, but do afterwards, the problem is that the threshold is dependent on the available EPC size of the target test device, so that makes it hard to automate.

One interesting scenario to lock down, irrespective of EPC size is what would happen if the enclave were to access a page that is outside the initial commit size.

Can we use oesgx to determine the amount of EPC available?

PR #3091, opened 2020-06-03
Fix `__oe_get_enclave_size()` to return ELRANGE size

### LINKED ISSUE #3056: __oe_get_enclave_size() does not reflect the ELRANGE
As observed in PR #3007:

- `_calculate_enclave_size()` in the enclave creation/measurement codepath produces 2 values:
    - `enclave_end`, the total size of the enclave loaded during enclave creation, including the enclave image and all the data pages such as the heap, stacks and TCS pages.
    - `enclave_size`, the size of the enclave as recorded by the SECS, which must be a power of 2 for SGX. This is the `enclave_end` value rounded up to the next power of 2, and reflects the ELRANGE of the enclave. 
- `oe_sgx_build_enclave()` calls `oe_sgx_create_enclave()` with `enclave_size` which sets that value in the SECS for the enclave. 
- `oe_sgx_build_enclave()` calls `patch()` on the enclave image with calculated `enclave_end` as the `image_info.enclave_size` global value before loading it into the enclave.
- After the enclave is initialized, the enclave internally can call `__oe_get_enclave_size()`, which returns the global `image_info.enclave_size` that was patched, but this value is potentially smaller than the size of the enclave as seen by SGX (which is the power of 2 `enclave_size` value).

This has some ambiguous semantics in the enclave, since `__oe_get_enclave_size()` is used by the bounds checks `oe_is_within_enclave()` and `oe_is_outside_enclave()`, which seem like they should reflect the SGX enclave boundary as determined by the enclave SECS rather than the limit of the pages initially provisioned in to the enclave. Its only other use seems to be in a vestigial helper function `oe_get_num_pages()` for a diagnostic code path for tests/ecall that is never built.

This may have more bearing once OE SDK supports SGX2, as there will likely be a need to figure out the end of the currently allocated pages in the enclave virtual address range for EDMM. At that point, we will need to track two separate variables: one for the currently allocated limit (`enclave_end`, which should be renamed, and likely be made mutable) and the actual size of the enclave (`enclave_size`). We should probably fix the semantics of the `__oe_get_enclave_size()` in the meantime though, by having `oe_sgx_build_enclave()` patch the image with `enclave_size` rather than `enclave_end`.

CodeMonkeyLeet:
`__oe_get_enclave_size()` is used in the `oe_is_within_enclave()` and `oe_is_outside_enclave()` checks, but currently only reflects the initially committed enclave pages during enclave creation, not the actual linear address range (ELRANGE) protected by SGX. This can create unexpected failures in SGX1 as a write to a unbacked page in ELRANGE as if it is outside the enclave will fail.

- Update `oe_sgx_build_enclave()` to patch oeimage with `enclave_size` instead of `enclave_end` (size of committed pages only).

- Remove extra EEID logic to provide the incorrect initial committed size as the enclave size. That code path behaves as in the normal case now.

Also updates tests/eeid_plugin to import the required OE system and platform EDLs;
PR #2961 refactored all tests to explicitly import their required EDLs, but did not update the EEID tests because they only build with the WITH_EEID experimental feature flag enabled.

Fixes #3056

### DIFF
--- a/enclave/core/sgx/calls.c
+++ b/enclave/core/sgx/calls.c
@@ -138,7 +138,6 @@ extern bool oe_disable_debug_malloc_check;
 #ifdef OE_WITH_EXPERIMENTAL_EEID
 extern volatile const oe_sgx_enclave_properties_t oe_enclave_properties_sgx;
 extern oe_eeid_t* oe_eeid;
-extern size_t oe_eeid_extended_size;
 
 int _is_eeid_base_image(const volatile oe_sgx_enclave_properties_t* properties)
 {
@@ -147,7 +146,7 @@ int _is_eeid_base_image(const volatile oe_sgx_enclave_properties_t* properties)
            properties->header.size_settings.num_tcs == 1;
 }
 
-static oe_result_t _eeid_patch_memory_sizes()
+static oe_result_t _eeid_patch_memory()
 {
     oe_result_t r = OE_OK;
 
@@ -157,7 +156,6 @@ static oe_result_t _eeid_patch_memory_sizes()
         uint8_t* heap_base = (uint8_t*)__oe_get_heap_base();
         oe_eeid_marker_t* marker = (oe_eeid_marker_t*)heap_base;
         oe_eeid = (oe_eeid_t*)(enclave_base + marker->offset);
-        oe_eeid_extended_size = marker->size;
 
         // Wipe the marker page
         memset(heap_base, 0, OE_PAGE_SIZE);
@@ -196,7 +194,7 @@ static oe_result_t _handle_init_enclave(uint64_t arg_in)
             oe_enclave_t* enclave = (oe_enclave_t*)arg_in;
 
 #ifdef OE_WITH_EXPERIMENTAL_EEID
-            OE_CHECK(_eeid_patch_memory_sizes());
+            OE_CHECK(_eeid_patch_memory());
 #endif
 
 #ifdef OE_USE_BUILTIN_EDL
--- a/enclave/core/sgx/globals.c
+++ b/enclave/core/sgx/globals.c
@@ -130,7 +130,6 @@ static volatile uint64_t _reloc_size;
 
 #ifdef OE_WITH_EXPERIMENTAL_EEID
 oe_eeid_t* oe_eeid = NULL;
-size_t oe_eeid_extended_size = 0;
 #endif
 
 /*
@@ -148,12 +147,7 @@ const void* __oe_get_enclave_base()
 
 size_t __oe_get_enclave_size()
 {
-#ifdef OE_WITH_EXPERIMENTAL_EEID
-    if (oe_eeid)
-        return oe_eeid_extended_size;
-    else
-#endif
-        return oe_enclave_properties_sgx.image_info.enclave_size;
+    return oe_enclave_properties_sgx.image_info.enclave_size;
 }
 
 const void* __oe_get_enclave_elf_header(void)
--- a/host/sgx/create.c
+++ b/host/sgx/create.c
@@ -813,7 +813,7 @@ oe_result_t oe_sgx_build_enclave(
     enclave->text = enclave_addr + oeimage.text_rva;
 
     /* Patch image */
-    OE_CHECK(oeimage.patch(&oeimage, enclave_end));
+    OE_CHECK(oeimage.patch(&oeimage, enclave_size));
 
     /* Add image to enclave */
     OE_CHECK(oeimage.add_pages(&oeimage, context, enclave, &vaddr));
--- a/host/sgx/create.c
+++ b/host/sgx/create.c
@@ -621,8 +621,11 @@ static oe_result_t _add_eeid_marker_page(
         memset(page, 0, sizeof(oe_page_t));
         oe_eeid_marker_t* marker = (oe_eeid_marker_t*)page;
 
+        /* The offset to the EEID in marker->offset is also the extended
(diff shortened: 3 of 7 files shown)


CodeMonkeyLeet:
@wintersteiger, note that I've made some fairly minor changes and fixups to the EEID stack.

CodeMonkeyLeet:
bors try

wintersteiger:
@CodeMonkeyLeet Thanks for that, now I understand what you meant! 

I think there's still a problem with this though: previously, `oe_eeid_extended_size` held the _extended_ size, computed by the second call to `_calculate_enclave_size` in `_add_eeid_marker_page`. Note that this uses different `props`. After your changes, `oe_enclave_properties_sgx.image_info.enclave_size` will hold the _base_ image sizes from the first call to `_calculate_enclave_size`, not the extended sizes. (Patching is the same for both, so as not to change the base image hash.) I think you're right in that we want `enclave_size` instead of `enclave_end`, but we still need another way to track the actual enclave size after extension. Since `oe_enclave_properties_sgx` is const, we can't overwrite it in there (I had a mutable version in an earlier draft), which is why I added `oe_eeid_extended_size`.

CodeMonkeyLeet:
> I think there's still a problem with this though: previously, `oe_eeid_extended_size` held the _extended_ size, computed by the second call to `_calculate_enclave_size` in `_add_eeid_marker_page`. Note that this uses different `props`. After your changes, `oe_enclave_properties_sgx.image_info.enclave_size` will hold the _base_ image sizes from the first call to `_calculate_enclave_size`, not the extended sizes. (Patching is the same for both, so as not to change the base image hash.) I think you're right in that we want `enclave_size` instead of `enclave_end`, but we still need another way to track the actual enclave size after extension. Since `oe_enclave_properties_sgx` is const, we can't overwrite it in there (I had a mutable version in an earlier draft), which is why I added `oe_eeid_extended_size`.

@wintersteiger thanks for pointing out the semantic inconsistency this introduced into `_add_eeid_marker_page()`, I've cleaned that up. 

To the broader point of whether the `oe_eeid_extended_size` is necessary, it's not actually used anywhere and my inclination is to remove it until it needs to be reintroduced.

In the master branch, it's semantically incorrect anyway because it was being assigned the `enclave_end` value that was calculated _before_ the dynamic size extension in the body of `_add_eeid_marker_page()`. The `enclave_end` value for the EEID extended size should be the offset to the EEID since it's now written after the extended data pages, so there's no need for a separate variable; we could reconstruct it from the existing `oe_eeid` pointer and enclave base address already available in the enclave.

wintersteiger:
Thanks for the fixes @CodeMonkeyLeet! 

You're right, we could re-compute the extended size in the enclave, but we would have to do so at every call to `__oe_get_enclave_size`, because we would not be able to store the result anywhere. It's not very expensive though. Further, we would need to copy the implementation of `_calculate_enclave_size` or move it to `common/*` to make it available in the enclave. I figured saving the size on the host and then writing it to a new global was the option with the least amount of downsides, but I may of course have been wrong. 

Another somewhat tempting option is to cast the `const` on `oe_enclave_properties_sgx` away for just one very brief moment... :-)

CodeMonkeyLeet:
@wintersteiger Yes, I agree that there's a slight arithmetic cost to deriving the extended base enclave size from existing variables. We could cache the value in the future if it turns out to be a performance hit, but nothing uses the value today, so I don't think it should be implemented until needed.

My intuition is that this discussion will be relevant again once OE SDK implements support for EDMM. Independent of EEID, the base OE enclave runtime would need to track the allocated enclave memory boundary, and while that value would start as the base image size, it would no longer be treated as a constant in the enclave.

wintersteiger:
@CodeMonkeyLeet That's a good point! Since this will have to be dynamic soon we may as well start making it dynamic now. We do currently use that in SGX-LKL, but only during startup and we cache the value, so at least in my current use-case it won't make a difference at all.

wintersteiger:
@CodeMonkeyLeet Just checking: do you want me to add the size calculation? Not sure I can push to your fork though.

CodeMonkeyLeet:
@wintersteiger No, thanks for the offer, but we should probably take that as a separate change; as noted previously, nothing currently requires it in the codebase, and I'd like to have a separate conversation with @jhand2 first around the form this dynamic sizing should take.

CodeMonkeyLeet:
Merge blocked by infrastructure failures to contact YUM servers. Retrying.

bors r+

wintersteiger:
@CodeMonkeyLeet  SGX-LKL uses it, i.e. it will report out-of-memory if we return the base image heap size.

CodeMonkeyLeet:
@wintersteiger I'm confused, what's the concern about getting base image heap size unexpectedly? How are you getting the base image heap size through OE today?

> i.e. the `oe_eeid_extended_size` never referred to the extended base enclave size to begin with. This PR just brings the normal codepath in line with the existing behavior of the EEID codepath.

The existing behavior of `__oe_get_enclave_size()` gives you the full virtual enclave size already and this PR doesn't change that.

wintersteiger:
In the case of EEID base images, the heap and stack size settings are 0, so 

```
return oe_enclave_properties_sgx.image_info.enclave_size;
```

will return a value that's too small. We need the extended size, which was previously saved in `oe_eeid_extended_size` (and on the marker page).

wintersteiger:
Darn, sorry about the noise! I thought that change was in `__oe_get_heap_size`. The one in `__oe_get_enclave_size` is perfectly fine, as you say, nobody uses it :-)

Issue #2776, opened 2020-04-01
Enclave has memory leaks or memory recovery delay


yanxue-intel:
A memory leak occurs in the enclave when malloc/free random size memory multiple times.

test cases in [https://github.com/openenclave/openenclave/pull/2774](url)
one case is checking for 1K memory to be  malloc/free in enclave 5000000 times.
the other case is checking for random size memory to be malloc/free in enclave 5000000 times.

tests ran on Ubuntu18.04, results as follow :
===Starting malloc fixed size fragment test.
[test_malloc_fixed_size_fragment]heap size before test : 1068928.
_malloc_free_fixed_size malloc times = 5000000.
[test_malloc_fixed_size_fragment]heap size after test : 1068928.
===Starting malloc random size fragment test.
[test_malloc_random_size_fragment]heap size before test : 1068928.
_malloc_free_random_size malloc times = 5000000.
[test_malloc_random_size_fragment]heap size after test : 1048448.

anakrish:
Thanks for identifying this!

achamayou:
@yanxue-intel there's one leak I know of, and have come across in #2649, it's the atexit handlers. I assumed it is deliberate because the code was directly calling oe_sbrk and bypassing the debug allocator. That's been modified to use oe_nodebug_malloc, which is more explicit, but the leak remains.

It's a very small amount of memory however, much less than your test suggests is leaking, but I thought I'd point it out nonetheless.

PR #3092, opened 2020-06-03
Fix malloc random size fragment test

larrywxyz:
#2776 

Seems like this issue got automatically closed when the PR closed, but it was never actually fixed. 

Basically, I think the problem with the test is that when random size memory being malloced is too large (in this case within 23 bytes of _get_heap_size()), dlmalloc will automatically increase the size of the heap. After the heap reaches its maximum size (2097024 bytes), the heap is shrunk back down (3968 bytes), before once again increasing in size if _randx(_get_heap_size()) is too large.

The simple fix is to restrict the max possible size being allocated in the random size fragment test.

### DIFF
--- a/tests/memory/enc/fragment.c
+++ b/tests/memory/enc/fragment.c
@@ -80,7 +80,7 @@ static int _malloc_free_random_size(int times, unsigned int seed)
     srand(seed);
     for (i = 0; i < times; i++)
     {
-        size = _randx(_get_heap_size());
+        size = _randx(_get_heap_size() - 23);
         buffer = malloc(size);
         if (NULL == buffer)
         {


tests/memory/enc/fragment.c
```diff
@@ -80,7 +80,7 @@ static int _malloc_free_random_size(int times, unsigned int seed)
     srand(seed);
     for (i = 0; i < times; i++)
     {
-        size = _randx(_get_heap_size());
+        size = _randx(_get_heap_size() - 23);
```
mingweishih, line 83: Thanks for investigate this! It seems that the problem is that the test hits a corner case (i.e., allocating certain large size heap). Rather than avoiding hitting the case, do you think we can fundamentally address the problem? I'd suggest as follows.
- Investigate if there is a fix (maybe certain configuration to the dlmalloc).
- Add tests to explicitly test this corner case. In this case, we should know what would be the expected behavior.

anakrish:
@larrywxyz 
>  dlmalloc will automatically increase the size of the heap

What does this mean within an enclave? The size of the enclave heap is a fixed-value.

tests/memory/enc/fragment.c
radhikaj: @larrywxyz  Can you address this comment?

radhikaj:
Thanks @larrywxyz Please reopen when you are able to work on this again.

PR #3104, opened 2020-06-04
Fix oesign tests introduced by #3046

### LINKED ISSUE #3103: oesign test cases added by #3046 are non-functional
PR #3046 introduced new tests for oesign that appear intended to assert the following:

1. oesign should fail when provided an RSA signing key that does not have the SGX-specified exponent 3.
2. oesign should fail when a negative value is provided for the configuration of heap pages
3. oesign should fail when provided an empty configuration file.

None of the tests actually test these conditions. The implementation is riddled with bugs:
- it tries to invoke a helper script oesign-customize-config.py, but the file is misnamed oesign-custmize-config.py script.
- the added tests reference the generated test config files from the source folder instead of the binary folder.
- the CMake changes do not actually trigger the creation of the new test data files because no explicit dependencies were expressed on them.

In all cases, this causes the invocation of oesign to fail, but for entirely incorrect reasons because the test collateral expected to be passed into oesign are simply not found.

If the tests did work, they would have caught that test assertions 1 & 2 are not enforced by oesign today and should be addressed.

Assertion 3 is up for debate. oesign doesn't require that a config file is always passed to it because the enclave may already be configured at build time by the developer using `OE_SET_ENCLAVE_SGX`. As long as all the required properties are already populated in the enclave metadata, `--config-file` is optional and an empty config file falls into the same category. The test for this case as implemented will continue to work though, since the oesign test enclave is explicitly implemented without the properties prepopulated.

CodeMonkeyLeet:
- Rename misspelled oesign-custmize-config.py to make-oesign-config.py.
- Fix config file tests to load generated .conf files from the correct
  CMAKE_CURRENT_BINARY_DIR location instead of CMAKE_CURRENT_SOURCE_DIR.
- Add `ALL` custom targets into test-enclave/CMakeLists.txt to as a global
  dependency to trigger generation of the test config files and signing keys.
- Rename test config files and signing keys to reflect their test intent.
- Add OE_INVALID_SGX_SIGNING_KEY error code and validation logic for expected
  exponent 3 value for SGX signing keys.
- Add checks for invalid negative numeric strings to _load_config_file()
  in oesign.c.
- Clarify oesign config file error messages.

Fixes #3103


CodeMonkeyLeet:
@anakrish Thanks for the info, it's something that I have other incoming PRs to handle, so it will not be done in this one. There are at least two more PRs to follow, depending on how I decide to group the changes:
- Refactor apart the oesign-engine tests so that new tests are not coupled to the testing of the specific oesign OpenSSL engine support feature.
- Refactor out the extraneous pass/fail handling wrapper oesign-test.py and refine the test pass/fail conditions in CMake directly.

BRMcLaren:
bors try

PR #3107, opened 2020-06-05
Refactor tests/tools/oesign-engine into tests/tools/oesign

CodeMonkeyLeet:
This rename reflects the expanded scope of testing under this folder. The tests/tools/oesign-engine tests were originally written specifically for only testing OpenSSL engine support in oesign, but new tests have since been added that apply more generally to oesign.

To cleanly encapsulate separate components that can be reused for other oesign tests, this changeset refactors the source tree so that the subfolders have clear boundaries of use:

- Move all build actions and sources specific to the oesign engine testing
  under `test-engine`:
  - Move the oesign engine test script oesign-test.py under test-engine.
  - Update oesign-test.py to use standard cmake targets where possible.
  - Move all engine ctests from test-enclave/CMakeLists.txt to
    test-engine/CMakeLists.txt.
- Add a new `test-inputs` folder to generate shared test collateral that
  can be reused by all oesign tests:
  - Move test-enclave/enclave/test-enclave.conf under test-inputs as
    valid.conf.
  - Move make-oesign-config.py under test-inputs.
  - Update make-oesign-config.py to consolidate both dev authored and
    generated config test files to the build output test-inputs folder.
- Update the target names for `test-enclave`
  - These are now `oesign_test_enc` and `oesign_test_host` instead of the
    generic `enclave` and `helloworld_host` to avoid collisions in the
    cmake dependency graph.
  - `test-enclave` is now only responsible for building the test enclave
    that can be reused by other tests.

The refactor of the test implementations all currently under test-engine will be done in a subsequent PR.


CodeMonkeyLeet:
bors try

CodeMonkeyLeet:
bors try

CodeMonkeyLeet:
bors try

CodeMonkeyLeet:
bors try

CodeMonkeyLeet:
bors try

tests/tools/oesign/test-inputs/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+
+# Generate a config file with negative value
+add_custom_command(OUTPUT negative_num_heap_pages.conf
+                   COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/make-oesign-config.py 1)
+
+add_custom_target(oesign_test_configs ALL DEPENDS valid.conf empty.conf
+                                                  negative_num_heap_pages.conf)
+
+# Generate a valid SGX signing key-pair
+add_custom_command(
```
radhikaj, line 22: I vaguely remember you should have a depends here. See https://github.com/openenclave/openenclave/pull/2226/files

CodeMonkeyLeet: @radhikaj Thanks for the review, Radhika! There is already a custom target that requires the build of the files generated by this custom command, see lines 34-36. They are being consolidated into a single target to make it easier to reference (i.e. downstream projects only need to specify a single target dependency)

tests/tools/oesign/test-enclave/enclave/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+# Use the edger8r to generate C bindings from the EDL file.
+add_custom_command(
+  OUTPUT helloworld_t.h helloworld_t.c helloworld_args.h
+  DEPENDS ${EDL_FILE} edger8r
+  COMMAND
+    edger8r --trusted ${EDL_FILE} --search-path ${PROJECT_SOURCE_DIR}/include
+    --search-path ${PLATFORM_EDL_DIR} --search-path ${CMAKE_CURRENT_SOURCE_DIR})
+
+add_executable(oesign_test_enc enc.c ${CMAKE_CURRENT_BINARY_DIR}/helloworld_t.c)
```
anakrish, line 14: I think this needs to be `add_enclave`

tests/tools/oesign/test-enclave/enclave/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+if (WIN32)
+  maybe_build_using_clangw(oesign_test_enc)
+endif ()
+
+target_compile_definitions(oesign_test_enc PUBLIC OE_API_VERSION=2)
+
+# Need for the generated file helloworld_t.h
+target_include_directories(oesign_test_enc PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
```
anakrish, line 23: `enclave_include_directories` and `enclave_link_libraries`

tests/tools/oesign/test-engine/CMakeLists.txt
```diff
@@ -0,0 +1,97 @@
...
+    ${OESIGN_TEST_INPUTS_DIR}/sign_key.private.pem --expect-fail)
+
+add_test(
+  NAME tests/oesign-engine-failed-key-id
+  COMMAND
+    ${CMAKE_CURRENT_SOURCE_DIR}/oesign-test.py --oesign-binary
+    ${OE_BINDIR}/oesign --enclave-binary $<TARGET_FILE:oesign_test_enc>
+    --config-file ${OESIGN_TEST_INPUTS_DIR}/valid.conf --engine-id
+    'oesign-test-engine' --engine-load-path $<TARGET_FILE:oesign_test_engine>
+    --key-id 'bogus\ key' --expect-fail)
```
anakrish, line 52: This is more a `non existent key` test rather than an `invalid key` test.

tests/tools/oesign/test-enclave/enclave/CMakeLists.txt
CodeMonkeyLeet: @anakrish No, this is specifically is `add_executable` because `add_enclave` attempts to sign the enclave at build time, which is specifically not necessary/desirable for this set of tests.  This set of tests also explicitly do not benefit from LVI mitigations, or ability to span OP-TEE currently.

tests/tools/oesign/test-enclave/enclave/CMakeLists.txt
CodeMonkeyLeet: For consistency with `add_executable`, this elides the additional overhead with the `enclave*` macros.

tests/tools/oesign/test-engine/CMakeLists.txt
CodeMonkeyLeet: Not sure what you mean; the `--key-id` is a value only interpreted by the engine, i.e. as far as oesign is concerned, it's just an arbitrary string. The engine in this case is the mock test-engine, which chooses to interpret the `--key-id` value as a file path. oesign doesn't care about the distinction of how the engine fails on this, only that it is eventually informed that it is an invalid parameter.

tests/tools/oesign/test-enclave/enclave/CMakeLists.txt
anakrish: I'd suggest adding a comment to note that difference.

tests/tools/oesign/test-engine/CMakeLists.txt
anakrish: Ah Ok. I thought `oesign` was validating that argument.

Looking at the code currently, we don't seem to be reporting an invalid `--key-id` in a manner that the user can act upon. The openssl code-path seems to do this:
https://github.com/openenclave/openenclave/blob/11eede869d823fab6b25052480a40c35e34e94ce/host/crypto/openssl/key.c#L109-L110

CodeMonkeyLeet: Yeah, I agree that the engine handling code path is not ideal in terms of its parameter handling; there should probably be a pass to standardize the parameter and error handling behavior in oesign (and more generally related tools like oesgx) but that's out of scope for this series of PRs.

CodeMonkeyLeet:
Backend out of space failure. Retrying.

bors r+

PR #3110, opened 2020-06-08
Fix regression on tests with LVI mitigation and a bug in the compiler detection

mingweishih:
This PR fixes the test regression with LVI mitigation. The cause is that we didn't correctly exclude the lvi target that causes the link-time errors.

Also, fixing a bug in the compiler detection when enabling LVI mitigation.
The bug causes the build system fallbacks to use the system installed compiler instead of wrappers when `CC` is not specified.


BRMcLaren:
https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/Sandbox/detail/Sandbox/49/pipeline/11

mingweishih:
bors try

BRMcLaren:
LGTM and confirming this fixes the issue

PR #3114, opened 2020-06-09
Add basic KSS support (ISVFAMILYID and ISVEXTPRODID) to OE SDK 

jianlia:

Enclave loader to detect whether the system support KSS, to parse the Enclave SIGSTRUCT.ATTRIBUTES and SIGSTRUCT.ATTRIBUTEMASK and decide whether to enable KSS


anakrish:
@jianlia Thanks for the contribution. How do we test this?

jianlia:
> @jianlia Thanks for the contribution. How do we test this?
Will update and add unit test :)

tools/oesign/oesign.c
```diff
@@ -300,6 +348,10 @@ void _merge_config_file_options(
     if (options->security_version != OE_UINT16_MAX)
         properties->config.security_version = options->security_version;
 
+    /* If SecurityVersion option is present */
```
CodeMonkeyLeet, line 351: Why is this duplicated from the above?

host/sgx/sgxsign.c
```diff
@@ -107,7 +107,6 @@ static oe_result_t _get_exponent(
     const oe_rsa_public_key_t* rsa,
     uint8_t exponent[OE_EXPONENT_SIZE])
 {
-    static const uint8_t OE_SGX_SIGNING_EXPONENT[] = {3, 0, 0, 0};
```
CodeMonkeyLeet, line 110: This PR has a bad merge that undoes previous fixes, please ensure that this and line 130-133 are not removed

include/openenclave/bits/sgx/sgxtypes.h
```diff
@@ -185,7 +186,10 @@ typedef struct _sgx_sigstruct
     uint32_t miscmask;
 
     /* (908) Reserved. Must be 0. */
-    uint8_t reserved2[20];
+    uint8_t reserved2[4];
```
CodeMonkeyLeet, line 189: Has it been validated that SGX1 devices that don't support KSS accept sigstructs where the reserved fields are non-zero?

tools/oesign/oesign.c
```diff
@@ -174,6 +178,50 @@ static int _load_config_file(const char* path, ConfigFileOptions* options)
 
             options->security_version = n;
         }
+        else if (strcmp(str_ptr(&lhs), "IsvFamilyID_H") == 0)
+        {
+            uint64_t n;
+
+            if (str_u64(&rhs, &n) != 0)
```
CodeMonkeyLeet, line 185: See updated numerical checks, these should check that the values provided are non-negative.

tools/oesign/oesign.c
```diff
@@ -174,6 +178,50 @@ static int _load_config_file(const char* path, ConfigFileOptions* options)
 
             options->security_version = n;
         }
+        else if (strcmp(str_ptr(&lhs), "IsvFamilyID_H") == 0)
```
CodeMonkeyLeet, line 181: This is not a good design for user-facing input configuration. Intel's initial design for this is awkward and we would not want developers to have to think about this value specified in the configuration as a pair of uint64 decimal numbers instead of just as a single 128-bit UUID.

We should restructure ISV Family ID and Extended Product ID to take a standard UUID/GUID formatted or hexadecimal strings instead.

tests/props/host/host.c
```diff
@@ -172,7 +172,7 @@ int main(int argc, const char* argv[])
             is_signed,
             1111,                                        /* product_id */
             2222,                                        /* security_version */
-            OE_SGX_FLAGS_DEBUG | OE_SGX_FLAGS_MODE64BIT, /* attributes */
+            OE_SGX_FLAGS_DEBUG | OE_SGX_FLAGS_MODE64BIT | OE_SGX_FLAGS_KSS, /* attributes */
```
CodeMonkeyLeet, line 175: Doesn't this break the tests? How is `OE_SGX_FLAGS_KSS` getting set on the enclaves?

tools/oesign/oesign.c
```diff
@@ -1,13 +1,15 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+#define __STDC_WANT_LIB_EXT1__ 1
```
CodeMonkeyLeet, line 4: Does GCC even support this? You're also not using the extended "safe" versions of the functions, so this seems unnecessary.

tools/oesign/oesign.c
jianlia: Will remove it.

host/sgx/sgxsign.c
jianlia: Sure

tests/props/host/host.c
jianlia: Yes actually. Will revert this one.

tools/oesign/oesign.c
jianlia: Remove it per comment.

tools/oesign/oesign.c
jianlia: Will update it with oe_uuid_t instead

tests/tools/oesign-engine/test-enclave/enclave/test-enclave-kss.conf
```diff
@@ -0,0 +1,13 @@
...
+# Enclave settings:
+Debug=1
+Kss=1
+NumHeapPages=1024
+NumStackPages=1024
+NumTCS=1
+ProductID=1
+SecurityVersion=1
+IsvFamilyID=7cc330ff-c8cf-49d4-92ef-e1674794f820
+IsvExtProductID=8cc330ff-c8cf-49d4-92ef-e1674794f820
```
CodeMonkeyLeet, line 13: Linux apps generally expect a newline character at the end of each text file for correct parsing, so GitHub flags the missing newline with the symbol you see here. Please try to add the required newlines back in when you see it in your PRs.

include/openenclave/internal/str.h
```diff
@@ -424,6 +424,78 @@ MEM_INLINE int str_split(str_t* str, const char* delim, str_t* lhs, str_t* rhs)
     return 0;
 }
 
+/**
+ * This type contains a Universally Unique Resource Identifier (UUID) type as
```
CodeMonkeyLeet, line 428: Why are you parsing to TA-specific data structure here? Elsewhere the functions that deal with the parsed value expect `oe_uuid_t` which is just a 128-bit buffer, and that may not line up with the memory layout of this definition because it's not a packed structure. You should probably simplify and parse directly to the byte-buffer representation that will avoid some of these errors, since none of this code ever expects to interpret the individual groups anyway.

tools/oesign/oesign.c
```diff
@@ -1,13 +1,14 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+#include <string.h>
```
CodeMonkeyLeet, line 4: I would avoid modifying the oesign tool right now since we don't intend to make the KSS functionality broadly exposed to users yet, and public changes to the tools generally go through a design review process. 

For the purposes of your investigations, it seems the addition of an internal `OE_SET_ENCLAVE_SGX_KSS` macro is sufficient for your testing needs.

tests/tools/oesign-engine/test-enclave/CMakeLists.txt
```diff
@@ -143,3 +143,11 @@ add_test(
```
CodeMonkeyLeet, line 143: I don't think this test runs or builds given the changes in master (oesign-engine has been removed as a folder and is now part of oesign/test-engine) See comments about rebasing and dropping the oesign changes to simplify.

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -31,17 +31,20 @@ typedef struct _oe_sgx_enclave_image_info_t
...
     uint16_t product_id;
     uint16_t security_version;
+    uint8_t isv_family_id[16];
+    uint8_t isv_ext_product_id[16];
 
     /* Padding to make packed and unpacked size the same */
     uint32_t padding;
 
-    /* (OE_SGX_FLAGS_DEBUG | OE_SGX_FLAGS_MODE64BIT) */
+    /* (OE_SGX_FLAGS_DEBUG | OE_SGX_FLAGS_MODE64BIT |OE_SGX_FLAGS_KSS) */
```
dthaler, line 47:
```suggestion
    /* (OE_SGX_FLAGS_DEBUG | OE_SGX_FLAGS_MODE64BIT | OE_SGX_FLAGS_KSS) */
```

host/sgx/sgxsign.c
```diff
@@ -521,6 +523,21 @@ static oe_result_t _init_sigstruct(
...
+            isvfamilyid,
+            sizeof(sigstruct->isvfamilyid)));
+        OE_CHECK(oe_memcpy_s(
+            sigstruct->isvextprodid,
+            sizeof(sigstruct->isvextprodid),
+            isvextprodid,
+            sizeof(sigstruct->isvextprodid)));
+    }
+
+
```
dthaler, line 540:
```suggestion
```

include/openenclave/internal/str.h
```diff
@@ -424,6 +424,78 @@ MEM_INLINE int str_split(str_t* str, const char* delim, str_t* lhs, str_t* rhs)
...
+    uint64_t uuid_parts[5];
+    char* id_copy;
+    char* posn;
+    char* current_token;
+
+#if defined(__linux__)
+    id_copy = strdup(str_ptr(str));
+#else
+    id_copy = _strdup(str_ptr(str));
+#endif
```
dthaler, line 452: These ifdefs are not needed, just use _strdup .  See host.h which #defines that as strdup on linux.  (Or in theory that could be swapped in host.h to just use strdup which is the [ISO standard](https://en.cppreference.com/w/c/experimental/dynamic/strdup) and define strdup as _strdup on Windows.)

include/openenclave/internal/str.h
```diff
@@ -424,6 +424,78 @@ MEM_INLINE int str_split(str_t* str, const char* delim, str_t* lhs, str_t* rhs)
...
+
+    if (strlen(id_copy) != 36)
+        return -1;
+
+    i = 5;
+#if defined(__linux__)
+    current_token = strtok_r(id_copy, "-", &posn);
+#else
+    current_token = strtok_s(id_copy, "-", &posn);
+#endif
```
dthaler, line 468: Similarly, should just use one or the other here.  strtok_s is the [ISO standard](https://en.cppreference.com/w/c/string/byte/strtok).  strtok_r is POSIX..  host.h can #define one to the other as needed, like it does for strdup today.

yentsanglee:
> typedef struct _sgx_report_body

isv_ext_prod_id and isv_family_id need to be defined in this struct

---
Refers to: include/openenclave/bits/sgx/sgxtypes.h:573 in 59bf81f. [](commit_id = 59bf81f6ea1ea7b1fa7baf47735f7dbc0d24ffde, deletion_comment = False)

tests/tools/oesign-engine/test-enclave/enclave/test-enclave-kss.conf
```diff
@@ -0,0 +1,13 @@
...
+
+# Enclave settings:
+Debug=1
+Kss=1
+NumHeapPages=1024
+NumStackPages=1024
+NumTCS=1
+ProductID=1
+SecurityVersion=1
+IsvFamilyID=7cc330ff-c8cf-49d4-92ef-e1674794f820
```
yentsanglee, line 12: It would be great to add more test cases to uuid_from_string() especially if IsvFamilyID or IsvExtProductID are not in correct UUID format. Since these values are specified externally, it's important that we can handle all error cases.

include/openenclave/internal/str.h
jianlia: Removed

include/openenclave/internal/str.h
jianlia: Good point. Updated.

include/openenclave/internal/str.h
jianlia: Update per comment. Thanks

tests/tools/oesign-engine/test-enclave/enclave/test-enclave-kss.conf
jianlia: Removed per refactoring.

tests/tools/oesign-engine/test-enclave/enclave/test-enclave-kss.conf
jianlia: Update by adding longer or shorted ID test cases

tests/tools/oesign-engine/test-enclave/CMakeLists.txt
jianlia: Update code on latest master to get refactor change

include/openenclave/bits/sgx/sgxproperties.h
jianlia: Done

radhikaj:
@jianlia Are you actively working on this? Can the comments be addressed and then this can be merged?

jianlia:
This was merged to a private branch as the supported hardware is not ready. Close for now.

PR #3119, opened 2020-06-09
Remove declaration for oe_strtol from corelibc/stdlib.h

### LINKED ISSUE #3087: Missing oe_strtol
I get undefined references when I try to use `oe_strtol`. It seems there is no implementation for that function, but there's a prototype in `corelibc/stdlib.h`  ([here](https://github.com/openenclave/openenclave/blob/3e15573418caed43f9094ff8aec36cdde4f278f7/include/openenclave/corelibc/stdlib.h#L49)).

Is this just an oversight or is there a reason for not including this function (and the prototype will be removed too)?

jhand2:
There is a declaration for oe_strtol in corelibc, but there is no actual implementation. Remove this function from the header.

Fixes #3087


jhand2:
bors try

PR #3122, opened 2020-06-09
If not specified, CMAKE_BUILD_TYPE is set to Debug

### LINKED ISSUE #2994: Make cmake and cmake -DCMAKE_BUILD_TYPE=Debug consistent
`cmake` and `cmake -DCMAKE_BUILD_TYPE=Debug` currently behave differently. 
The latter turns on DEBUG_MALLOC feature whereas the former doesn't.

anakrish:
This ensures that
  cmake
and
  cmake -DCMAKE_BUILD_TYPE=Debug
result in the same build configuration.

Previously, the former did not turn on DEBUG_MALLOC,
whereas the latter did.

Fixed #2994 

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,12 +94,18 @@ endif ()
 # ignored.
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+if ("${CMAKE_BUILD_TYPE}" STREQUAL "")
+  set(CMAKE_BUILD_TYPE "Debug")
+endif ()
+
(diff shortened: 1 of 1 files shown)


anakrish:
bors try

anakrish:
> Looks good to me!
> 
> It might be worth logging what the CMAKE_BUILD_TYPE is when generating the build with cmake. Maybe this value is printed somewhere else, though.
> 
> ```
> message(STATUS "Build type: ${CMAKE_BUILD_TYPE}") 
> ```

Done.

PR #3124, opened 2020-06-11
Remove __thread local storage modifier from liboecore

mikbras:
This PR removes all remaining uses of the **__thread** local storage modifier from **liboecore**. It relocates these variables to **td_t**.


anakrish:
@mikbras Can you sign off your PRs?

mikbras:
> @mikbras Can you sign off your PRs?

What does it mean for me to sign off my PR. I think I have fallen behind of the process terminology maybe. :)

mikbras:
> @mikbras Can you sign off your PRs?

Thank you Anand. Signed off.

mikbras:
bors try

anakrish:
@mikbras The failure seems to be in OP-TEE build which does not have the oe_sgx_get_td function,

enclave/core/arena.c
```diff
@@ -89,13 +94,16 @@ void* oe_arena_calloc(size_t num, size_t size)
 
 void oe_arena_free_all()
 {
-    _arena.used = 0;
+    oe_shared_memory_arena_t* arena = _get_arena();
+    arena->used = 0;
 }
 
 // Free the arena in the current thread.
 void oe_teardown_arena()
```
jhand2, line 102: Is any of this stuff called in OP-TEE? We should probably move arena.c and arena.h to enclave/core/sgx

jhand2:
> I'm a little more concerned with the implication that OE (and by extension, TEE provider implementations or other OE extension components) moving forwards cannot/should not use static thread locals. Is that correct?

oecore and below, so yes that would probably need to include TEE implementations. I do share your concern, but I can't think of a better way to allow an application to have full control over changes to the fs register.

The most correct solution would be if OE had kernel-like properties and could save and restore registers on context switches. But there are numerous reasons that's not possible (no access to ring-0 instructions, no single OE entry point).

anakrish:
> Tangent: @anakrish @mingweishih Per Ming-wei's analysis, switchless calls are an exclusively SGX feature today, and I don't believe we've decided whether it should be generalized across TEEs. This change pushes the coupling more deeply with the existing SGX-centric runtime implementation, which is probably fine if it stays that way.
> 

Yes. It is fine if switchless calls are deeply couple to sgx. When we need to implement switchless or a similar mechanism for other TEEs, it should be possible to decouple parts of the implementation.

> I'm a little more concerned with the implication that OE (and by extension, TEE provider implementations or other OE extension components) moving forwards cannot/should not use static thread locals. Is that correct?

Yes. Originally I was a bit concerned with this restriction too. But I think it is justified.
Given that we want oecore to not use libc, and instead use corelibc, it think this restriction is consistent and justified. Thread-local support is part of the C/C++ runtime since C/C++11, and therefore use of __thread/_Thread/thread_local within oecore implies a coupling with libc which we want to avoid.

enclave/core/arena.c
```diff
@@ -35,19 +39,20 @@ void* oe_arena_malloc(size_t size)
     oe_result_t result = OE_UNEXPECTED;
     size_t total_size = 0;
     const size_t align = OE_EDGER8R_BUFFER_ALIGNMENT;
+    oe_shared_memory_arena_t* arena = _get_arena();
 
     // Create the anera if it hasn't been created.
-    if (_arena.buffer == NULL)
```
jxyang, line 40: Where do we initialize `arena->buffer` to NULL?

HernanGatta:
Let me try this on an OP-TEE build. We need to make sure that no thread-local relocations end up in the ELF, otherwise TAs won't load.

mikbras:
> @mikbras The failure seems to be in OP-TEE build which does not have the oe_sgx_get_td function,

Thank you Anand. It is fixed now. I moved **arena.h** and **arena.c** to the sgx directory.

enclave/core/arena.c
mikbras: Good idea. Done.

mikbras:
> > Tangent: @anakrish @mingweishih Per Ming-wei's analysis, switchless calls are an exclusively SGX feature today, and I don't believe we've decided whether it should be generalized across TEEs. This change pushes the coupling more deeply with the existing SGX-centric runtime implementation, which is probably fine if it stays that way.
> 
> Yes. It is fine if switchless calls are deeply couple to sgx. When we need to implement switchless or a similar mechanism for other TEEs, it should be possible to decouple parts of the implementation.
> 
> > I'm a little more concerned with the implication that OE (and by extension, TEE provider implementations or other OE extension components) moving forwards cannot/should not use static thread locals. Is that correct?
> 
> Yes. Originally I was a bit concerned with this restriction too. But I think it is justified.
> Given that we want oecore to not use libc, and instead use corelibc, it think this restriction is consistent and justified. Thread-local support is part of the C/C++ runtime since C/C++11, and therefore use of __thread/_Thread/thread_local within oecore implies a coupling with libc which we want to avoid.

Great discussion. Thank you everyone. By avoiding use of the **__thread** modifier in **liboecore**, we allow applications to bring their own C runtime (which in most cases takes over the TLS region).

enclave/core/arena.c
mikbras: Thank you Xuejun. I added an oe_once function to zero-initialize the arena.

mikbras:
> Let me try this on an OP-TEE build. We need to make sure that no thread-local relocations end up in the ELF, otherwise TAs won't load.

Hello Hernan. I moved all **__thread** usages to the sgx directory except for one. There is now an SGX and OPTEE version of errno.c. The OPTEE version still uses the **__thread** modifier. Is there a way in OPTEE to put something into TLS without using the modifier?

mikbras:
bors try

mikbras:
Thank you everyone. I moved **arena.h** and **arena.c** to the sgx directory. I also created **errno.c** versions for both SGX and OPTEE. The OPTEE version is unmodified. I asked Hernan to look at this for OPTEE.

HernanGatta:
> Hello Hernan. I moved all **__thread** usages to the sgx directory except for one. There is now an SGX and OPTEE version of errno.c. The OPTEE version still uses the **__thread** modifier. Is there a way in OPTEE to put something into TLS without using the modifier?

On OP-TEE, `__thread` gets `#define`'d to nothing since the loader does not know what to do with thread-specific relocations, so you could remove the `__thread` modifier altogether from any OP-TEE code. As to your question, there is no TLS support at all, so no.

mikbras:
bors try

mikbras:
> > Hello Hernan. I moved all **__thread** usages to the sgx directory except for one. There is now an SGX and OPTEE version of errno.c. The OPTEE version still uses the **__thread** modifier. Is there a way in OPTEE to put something into TLS without using the modifier?
> 
> On OP-TEE, `__thread` gets `#define`'d to nothing since the loader does not know what to do with thread-specific relocations, so you could remove the `__thread` modifier altogether from any OP-TEE code. As to your question, there is no TLS support at all, so no, with or without the modifier.

Oh, great to hear Hernan. So **__thread** has no effect. Thank you for checking.

mikbras:
The build seems to be failing because the build infrastructure is messed up rather than the code.

mikbras:
bors try

mikbras:
bors try

jhand2:
Please remember to squash commits before merging. Thanks!

enclave/core/sgx/arena.c
```diff
@@ -20,6 +19,21 @@ static const size_t _max_capacity = 1 << 30;
...
+{
+    oe_sgx_td_t* td = oe_sgx_get_td();
+    memset(&td->arena, 0, sizeof(td->arena));
+}
+
+static oe_shared_memory_arena_t* _get_arena()
+{
+    static oe_once_t _once = OE_ONCE_INITIALIZER;
+
+    oe_once(&_once, _zero_initialize_arena);
```
anakrish, line 32: @jxyang We want the arena to be initialized every time an enclave thread is created, which currently means every time an ecall is made. `oe_once` means that  the action is executed only once; during the lifetime of the enclave.

As written oe_once will get executed for only one of the arenas; not for each arena. If an enclave ha 16 TCS, we want all the arenas to be initialized; not just one.

Note: We are not getting any runtime errors because, oe_teardown_arena zeroes out the memory, effectively doing a `_zero_initialize_arena`.

anakrish: @mikbras The use of `oe_once` to initialize per-thread arena introduces a bug. See my comment above.

mikbras: Oh yikes! You are right. I missed that. Let me rethink this.

mikbras: Okay, I backed out that change. I turns out that the TD is already zero-initialized by **td_init()** so there is no need to zero-initialize the arena.

mikbras:
bors try

anakrish:
@mikbras Can you please squash and r+ ?

anakrish:
Sporadic failure.

PR #3126, opened 2020-06-11
Configure memory allocator in OP-TEE-based enclaves

HernanGatta:
Recent changes to the enclave memory allocator require it to be
initialized during enclave creation and cleaned up during enclave
termination. This patch adds those function calls in the appropriate
lifecycle handlers for OP-TEE-based enclaves.

Signed-off-by: Hernan Gatta <hegatta@microsoft.com>


dthaler:
bors try

PR #3128, opened 2020-06-12
Add support for digest signing to oesign tool

### LINKED ISSUE #2089: Support digest signing workflow in oesign for production signing

CodeMonkeyLeet:
- Refactor `oesign()` internal implementation into submethods to support digest signing
- Add `digest` command to oesign
- Add digest signing options to `oesign sign` command
- Add tests/tools/oesign/test-digest to test `oesign digest` command and digest signing options for the sign command.
- Expand test-engine tests to test for parameter interactions with digest signing and engine signing parameter dependencies.
- Update `oe_public_key_verify()` in host/crypto/bcrypt to wrap `STATUS_INVALID_SIGNATURE` result from `BCryptVerifySignature()` as `OE_VERIFY_FAILED` for consistency with OpenSSL implementation on Linux.

Fixes #2089


CodeMonkeyLeet:
bors try

CodeMonkeyLeet:
bors try

CodeMonkeyLeet:
bors try

radhikaj:
Thanks @CodeMonkeyLeet !

CodeMonkeyLeet:
bors try

radhikaj:
@mingweishih  can you please review this

tests/tools/oesign/sign-and-verify.py
```diff
@@ -25,21 +35,18 @@
         arg_parser.print_help()
         sys.exit(1)
 
-    sign_cmd = "{} sign --enclave {} {}".format(args.oesign_path, args.enclave_path, args.oesign_args)
-    launch_cmd = "{} {}.signed".format(args.host_path, args.enclave_path)
+    if args.digest_args:
+        digest_cmd = "{} digest --enclave {} {}".format(args.oesign_path, args.enclave_path, args.digest_args)
```
achamayou, line 39: It's safer to assemble commands as lists, and to not pass shell=True to subprocess.call(). In particular, it avoids having to deal with paths that contain values that will break tokenisation such as spaces, which this doesn't do correctly.

host/sgx/sgxsign.c
```diff
@@ -664,3 +737,95 @@ oe_result_t oe_sgx_sign_enclave(
...
+    const OE_SHA256* mrenclave,
+    uint64_t attributes,
+    uint16_t product_id,
+    uint16_t security_version,
+    OE_SHA256* digest)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    sgx_sigstruct_t sigstruct;
+
+    if (digest)
```
mingweishih, line 751: Consider keep this check and remove the same check below (`!digest)` or remove this check and put the `memset`  after the checks of parameters.

host/sgx/sgxsign.c
```diff
@@ -664,3 +737,95 @@ oe_result_t oe_sgx_sign_enclave(
...
+    sgx_sigstruct_t* sigstruct)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    oe_cert_t cert = {0};
+    bool cert_initialized = false;
+    oe_rsa_public_key_t rsa = {0};
+    bool rsa_initalized = false;
+    OE_SHA256 digest = {0};
+
+    if (sigstruct)
```
mingweishih, line 787: Consider keep this check and remove the same check below (`!sigstruct)` or remove this check and put the `memset`  after the checks of parameters.

include/openenclave/internal/sgxsign.h
```diff
@@ -61,6 +60,51 @@ oe_result_t oe_sgx_sign_enclave_from_engine(
     const char* key_id,
     sgx_sigstruct_t* sigstruct);
 
+/**
+ * Get the digest of an unsigned sigstruct given the expected MRENCLAVE,
+ * ATTRIBUTES, ISVPRODID, and ISVSVN values.
+ *
+ * @param mrenclave[in] hash of the enclave to be signed
+ * @param digest[out] the digest of the sigstruct to be signed
```
dthaler, line 68: missing @param lines for attributes, product_id, and security_version

include/openenclave/internal/sgxsign.h
```diff
@@ -61,6 +60,51 @@ oe_result_t oe_sgx_sign_enclave_from_engine(
...
+ * **mrenclave** parameter with the provided **digest_signature**.
+ * The **digest_signature** is verified against the public signing
+ * certificate specified by the **cert_pem_data** parameter.
+ * If successful, the function writes the signature into the
+ * **sigstruct** parameter (an SGX signature structure).
+ *
+ * @param mrenclave[in] hash of the enclave to be signed
+ * @param cert_pem_data[in] PEM buffer containing the public signing cert
+ * @param cert_pem_size[in] size of the PEM buffer
+ * @param sigstruct[out] the SGX signature
```
dthaler, line 93: missing @param lines for other parameters

tools/oesign/oesign.c
```diff
@@ -346,43 +344,127 @@ int oesign(
...
 
+    result = OE_OK;
+
+done:
+    return result;
+}
+
+oe_result_t _get_sgx_enclave_hash(
+    const char* enclave,
+    oe_sgx_enclave_properties_t* props,
```
dthaler, line 373: Make const.  Also suggest not abbreviating "properties".

```suggestion
    const oe_sgx_enclave_properties_t* properties,
```

host/sgx/sgxsign.c
CodeMonkeyLeet: This replicates the pattern that Mike put in place to always return zeroed out buffers on failed function calls, which is why it precedes the invalid parameter check.

host/sgx/sgxsign.c
CodeMonkeyLeet: Same as above.

host/sgx/sgxsign.c
mingweishih: Thanks for the clarification!

tests/tools/oesign/sign-and-verify.py
CodeMonkeyLeet: Thanks for the info, I've made the changes but I'll submit it as a PR after this one to improve the python scripts as that probably deserves a separate review.

tests/tools/oesign/test-digest/CMakeLists.txt
```diff
@@ -0,0 +1,282 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+set(OESIGN_TEST_INPUTS_DIR ${CMAKE_CURRENT_BINARY_DIR}/../test-inputs)
+
+add_custom_command(
+  OUTPUT sign-and-verify.py
+  COMMAND cmake -E copy ${CMAKE_CURRENT_SOURCE_DIR}/../sign-and-verify.py
+          ${CMAKE_CURRENT_BINARY_DIR})
```
anakrish, line 9: In case CMAKE does not automatically do this, it would be good to add 
`DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/../sign-and-verify.py`

tests/tools/oesign/sign-and-verify.py
anakrish: Can we avoid the python scripts by just chaining the sign and run commands?
The idiomatic way to pass arguments would be to use `shlex.join`

tests/tools/oesign/test-engine/CMakeLists.txt
```diff
@@ -7,12 +7,20 @@ target_include_directories(oesign_test_engine PRIVATE include)
 
 set(OESIGN_TEST_INPUTS_DIR ${CMAKE_CURRENT_BINARY_DIR}/../test-inputs)
 
-configure_file(../sign-and-verify.py sign-and-verify.py COPYONLY)
+add_custom_command(
+  OUTPUT sign-and-verify.py
+  COMMAND cmake -E copy ${CMAKE_CURRENT_SOURCE_DIR}/../sign-and-verify.py
```
anakrish, line 12: `DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/../sign-and-verify.py`

tools/oesign/oesign.c
```diff
@@ -269,10 +262,24 @@ static int _load_pem_file(const char* path, void** data, size_t* size)
...
 }
 
+static int _load_pem_file(const char* path, void** data, size_t* size)
+{
+    int err = _load_file(path, data, size);
+    if (err == 0)
+    {
+        /* Zero terminate the PEM data. */
+        uint8_t* data_tmp = (uint8_t*)*data;
+        data_tmp[*size] = 0;
```
anakrish, line 272: Is this safe? Would *data have space for this null terminator?

tests/tools/oesign/test-digest/CMakeLists.txt
CodeMonkeyLeet: Added, but for my understanding, what's the purpose of adding source file dependencies for `add_custom_command`? It's still a build time error which will happen anyway when the command runs and the target source file doesn't exist in the repo.

host/sgx/sgxsign.c
```diff
@@ -439,17 +440,142 @@ static oe_result_t _get_q1_and_q2(
...
 
+static oe_result_t _hash_sigstruct(
+    const sgx_sigstruct_t* sigstruct,
+    OE_SHA256* hash)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    unsigned char buf[sizeof(sgx_sigstruct_t)];
+    size_t buf_size = 0;
+
+    OE_CHECK(oe_memcpy_s(
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Added. The buffer is being used to combined two non-contiguous segments of the sigstruct to be hashed as expected by SGX.

tests/tools/oesign/sign-and-verify.py
CodeMonkeyLeet: Strictly speaking, you probably could, but it was a real pain to do to get the bash command line to work correctly with the appropriate characters on both windows and linux. Just for readability purposes, I'd keep the script, for the same reason I'd rather have cmake invoke a .sh rather than copy and paste blobs of concatenated parameters in cmake directly.

Regarding idiomatic invocation of commands, Amaury is correct, the recommendation is to pass lists of args into `subprocess.call()` instead of stringifying them first anyway and the follow-up PR will deal with that.

tests/tools/oesign/test-engine/CMakeLists.txt
CodeMonkeyLeet: Added

tools/oesign/oesign.c
CodeMonkeyLeet: `_load_pem_file()` takes a dependency on `_load_file()` which always allocates the extra size needed for null-termination.

include/openenclave/internal/sgxsign.h
CodeMonkeyLeet: Fixed all comments in this file

tools/oesign/oesign.c
CodeMonkeyLeet: Made `const` and replaced all existing instances of `props` with `properties` that this instance was originally copied from

PR #3129, opened 2020-06-12
Fix spurious warning about a buffer being to small

wintersteiger:
In `oe_sgx_get_supported_attester_format_ids` OE complains that a buffer is too small in trace/debug output, while this is completely expected (the size being 0 is used to indicate that the required buffer size should be computed). This had some of us confused as it starting showing up in unrelated test runs.


wintersteiger:
Oh yeah, that's right, we need another branch here.

wintersteiger:
Fixed, thanks!

wintersteiger:
bors try

PR #3131, opened 2020-06-12
Move EEID attester plugin and install headers

wintersteiger:
For an enclave to get EEID attestation evidence, it explicitly needs to register the EEID attester and it needs to know the EEID plugin UUID. To expose those, I moved the corresponding header files and added them to the installation target. 

I'm not sure this is the intended way for enclaves to use an attestation plugin. I think we definitely need the plugin UUID in a public header, but should the other headers remain private/internal and the plugin infrastructure would automatically register the plugin corresponding to the UUID?

### DIFF
--- a/include/openenclave/attestation/sgx/eeid_plugin.h
+++ b/include/openenclave/attestation/sgx/eeid_plugin.h
@@ -1,13 +1,10 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
-// This file provides an implementation of EEID attester and verifier plugins.
+// This file provides the UUID of the EEID attester and verifier plugins.
 
 #ifndef _OE_EEID_UUID_H
 #define _OE_EEID_UUID_H
-
-#include <openenclave/internal/plugin.h>
-#include <openenclave/internal/sgx/plugin.h>
 
 #define OE_FORMAT_UUID_SGX_EEID_ECDSA_P256                                \
     {                                                                     \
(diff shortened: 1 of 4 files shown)


yentsanglee:
For an enclave to get EEID attestation evidence, it explicitly calls oe_sgx_eeid_verifier_initialize() or oe_sgx_eeid_attester_initialize(). However, the UUID needs to be exposed so app can explicitly specify the format when generating or verifying evidence.

Other UUIDs will need to move to public header.

wintersteiger:
+1 I think so too, we should just publish all UUIDs.

include/openenclave/attestation/sgx/eeid_plugin.h
```diff
@@ -1,14 +1,11 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
-// This file provides an implementation of EEID attester and verifier plugins.
+// This file provides the UUID oif the EEID attester and verifier plugins.
```
qiucwang, line 4:
```suggestion
// This file provides the UUID of the EEID attester and verifier plugins.
```

shnwc:
Agreed UUIDs should be made public. As documented in PR #3136, when support of SGX report and quote is added to `oe_verify_evidence()`, the UUIDs for those evidence data types need to be made available to applications.

radhikaj:
Bors r+

wintersteiger:
(Squashed and signed off)

PR #3136, opened 2020-06-16
Add support of SGX reports and quotes in oe_verify_evidence()

shnwc:
Add support of SGX reports and quotes in `oe_verify_evidence()` to incorporate PR 3089, change custom claims to be a flat buffer.

### DIFF
--- a/docs/DesignDocs/Attestation_API_Proposal.md
+++ b/docs/DesignDocs/Attestation_API_Proposal.md
@@ -12,6 +12,7 @@ This proposal is based on these efforts:
 - Design document [Attestation V3 Update]:
 [Custom Attestation Data Formats for Open Enclave V3 Updates](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/CustomAttestation_V3.md).
 - Discussion in [Issue #2729]: [Outstanding issues on attestation plugin design](https://github.com/openenclave/openenclave/issues/2729).
+- Proposal to add attestation evidence format support [[PR #3089](https://github.com/openenclave/openenclave/pull/3089)].
 - Design document [quote-ex Integration]:
 [Attestation: OE SDK Integration with Intel® SGX SDK quote-ex Library for Generation of Evidence in New Formats](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/SGX_QuoteEx_Integration.md).
 - Design document [Design Notes]:
@@ -67,7 +68,7 @@ Use cases for evidence generation:
   document.
 - Get evidence in a globally unique format, optionally along with a set of
 endorsements.
-  - Function `oe_result_t oe_get_evidence(const oe_uuid_t* evidence_format_uuid, uint32_t flags, const oe_claim_t* custom_claims, size_t custom_claims_length, const void* opt_params, size_t opt_params_size, uint8_t** evidence_buffer, size_t* vidence_buffer_size, uint8_t** endorsements_buffer, size_t* endorsements_buffer_size)`,
+  - Function `oe_result_t oe_get_evidence(const oe_uuid_t* format_id, uint32_t flags, const oe_claim_t* custom_claims, size_t custom_claims_length, const void* opt_params, size_t opt_params_size, uint8_t** evidence_buffer, size_t* vidence_buffer_size, uint8_t** endorsements_buffer, size_t* endorsements_buffer_size)`,
     declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
   - Upon successful evidence generation, `oe_get_evidence()` returns the
   generated evidence in a dynamically allocated buffer. It also has the option
@@ -76,7 +77,7 @@ endorsements.
   responsible to free these buffers.
   - The `flags` parameter is a legacy inherited from the legacy API
   `oe_get_report()` to indicate SGX local or remote attestation.
-  It's redundant -- overlapping with the `evidence_format_uuid` parameter
+  It's redundant -- overlapping with the `format_id` parameter
   which identifies evidence formats with globally unique UUIDs.
   - Note: please see the "Semantics of Custom Claims and Optional Parameters"
   section of the
@@ -159,15 +160,19 @@ generation and verification in
 manner.
 - Remove legacy artifacts that have become redundant, e.g. the SGX
 local / remote flags.
+- Incorporate changes proposed in [PR #3089](https://github.com/openenclave/openenclave/pull/3089).
 
 Proposed Headers files:
 
-- Attester public API to be declared in `openenclave/attestation/attester.h`
-- Verifier public API to be declared in `openenclave/attestation/verifier.h`
+- Attester public API to be declared in `<openenclave/attestation/attester.h>`
+- Verifier public API to be declared in `<openenclave/attestation/verifier.h>`
+- TEE-agnostic claim IDs to be declared in `<openenclave/bits/evidence.h>`
+- SGX-specific evidence format UUIDs and claim IDs to be declared in
+`<openenclave/attestation/sgx/evidence.h>`
 - Plugin API (helper functions and plugin entry points) to be declared in
-`openenclave/internal/plugin.h`
-- SGX evidence format UUIDs (and any other SGX-specific plugin related names)
-to be declared in `openenclave/internal/sgx/plugin.h`
+`<openenclave/internal/plugin.h>`
+- SGX-specific plugin related names to be declared in
+`<openenclave/internal/sgx/plugin.h>`
 
 Use cases for evidence generation:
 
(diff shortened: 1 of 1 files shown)


docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -159,15 +160,19 @@ generation and verification in
...
 local / remote flags.
+- Incorporate changes proposed in [PR #3089](https://github.com/openenclave/openenclave/pull/3089).
 
 Proposed Headers files:
 
-- Attester public API to be declared in `openenclave/attestation/attester.h`
-- Verifier public API to be declared in `openenclave/attestation/verifier.h`
+- Attester public API to be declared in `<openenclave/attestation/attester.h>`
+- Verifier public API to be declared in `<openenclave/attestation/verifier.h>`
+- TEE-agnostic claims IDs to be declared in `<openenclave/bits/evidence.h>`
```
dthaler, line 169:
```suggestion
- TEE-agnostic claim IDs to be declared in `<openenclave/bits/evidence.h>`
```

Also, evidence.h is fine for now but in the future, we might want claim ids (or at least some subset of them) to be reusable by attestation results too, in which case we may want a header like claims.h or something that can be included by both evidence.h and whatever is used for attestation results.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -235,8 +243,15 @@ Use cases for evidence verification:
   - As defined in document
   [[Attestation V3 Update]](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/CustomAttestation_V3.md)
 - Verify evidence, optionally with a set of endorsements and policies.
-  - Function `oe_result_t oe_verify_evidence(const uint8_t* evidence_buffer, size_t evidence_buffer_size, const uint8_t* endorsements_buffer, size_t endorsements_buffer_size, const oe_policy_t* policies, size_t policies_size, oe_claim_t** claims, size_t*claims_length)`.
-  - This function has the same definition as in the OE SDK V0.9 release.
+  - Function `oe_result_t oe_verify_evidence(const oe_uuid_t* format_id, const uint8_t* evidence_buffer, size_t evidence_buffer_size, const uint8_t* endorsements_buffer, size_t endorsements_buffer_size, const oe_policy_t* policies, size_t policies_size, oe_claim_t** claims, size_t*claims_length)`.
```
dthaler, line 246:
```suggestion
  - Function `oe_result_t oe_verify_evidence(const oe_uuid_t* format_id, const uint8_t* evidence_buffer, size_t evidence_buffer_size, const uint8_t* endorsements_buffer, size_t endorsements_buffer_size, const oe_policy_t* policies, size_t policies_size, oe_claim_t** claims, size_t* claims_length)`.
```
(add missing space)

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -235,8 +243,15 @@ Use cases for evidence verification:
...
+  - The optional parameter `format_id` is added.
+    - If it is `NULL`, the `evidence_buffer` must contain evidence data
+    generated by `oe_get_evidence()`. The header of of this type of evidence
```
dthaler, line 249:
```suggestion
    generated by `oe_get_evidence()`. The header of this type of evidence
```

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -565,6 +584,20 @@ plugin libraries, composed of:
   - It also implements OCALLs for the enclave-side library to access
   host-side services for SGX quote verification.
 
+The OE SDK V0.10 release will also have enclave-side and host-side plugins
+for verification of SGX ECDSA reports generated by legacy API function
```
dthaler, line 588:
```suggestion
for verification of SGX ECDSA reports generated by the legacy API function
```

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -565,6 +584,20 @@ plugin libraries, composed of:
...
+`oe_get_report()` and raw SGX ECDSA quotes generated by the Intel SGX SDK.
+These two evidence formats are identified by their UUIDs:
+- `OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT`: for SGX ECDSA report generated
+by `oe_get_report()`.
+- `OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE`: for raw SGX ECDSA quote generated
+by the Intel SGX SDK.
+
+For evidence data generated by `oe_get_evidence()`, the list of claims returned
+by `oe_verify_evidence()` will contain one claim named
```
dthaler, line 597:
```suggestion
by `oe_verify_evidence()` can contain one claim named
```
It would only have such a claim if the set of custom claims was not the empty set, correct?

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Agreed there is further clean up as the attestation API definition evolves. This header file `<openenclave/bits/evidence.h>` is existing, and this PR tried to avoid changes not directly to the proposed features.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Thanks!

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Thanks!

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Thanks!

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: What I assume is that `OE_CLAIM_CUSTOM_CLAIMS_ID` is always present in this case, but its value could be a zero-sized buffer.

In practice, either way would work, though the verifier application logic need to handle the error cases slightly differently.

dthaler: I would prefer a behavior (for ALL claims) that the claims are only returned if they're present in the evidence.   My second choice would be for the behavior to be the same for ALL claims, so that if you return claims with a 0-size buffer, then you do the same for ALL possible claims, but I like that behavior much less.   I would argue against having the behavior be different for different clams.

shnwc: I'm ok not to output an empty custom claims buffer. Let me update the document to reflect that.

The second choice would be hard to implement for those claims that are only applicable to a subset of formats. One example is `OE_CLAIM_SGX_REPORT_DATA_ID` claim.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -159,15 +160,19 @@ generation and verification in
...
+- TEE-agnostic claim IDs to be declared in `<openenclave/bits/evidence.h>`
+- SGX-specific evidence format UUIDs and claim IDs to be declared in
+`<openenclave/attestation/sgx/evidence.h>`
 - Plugin API (helper functions and plugin entry points) to be declared in
-`openenclave/internal/plugin.h`
-- SGX evidence format UUIDs (and any other SGX-specific plugin related names)
-to be declared in `openenclave/internal/sgx/plugin.h`
+`<openenclave/internal/plugin.h>`
+- SGX-specific plugin related names to be declared in
+`<openenclave/internal/sgx/plugin.h>`
```
yentsanglee, line 175: Format UUIDs need to be declared in public header because public functions oe_get_evidence() and oe_verify_evidence() both need them. See similar issue fixed in PR 3131

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -474,8 +490,11 @@ by the verifier plugin entry points defined in the `oe_verifier_t` structure:
     so the OE SDK framework can free it, and there is no need of a plugin entry
     point.
 - Verify evidence, optionally using a set of input endorsements and policies.
-  - Entry point `oe_result_t (*verify_evidence)(oe_verifier_t* context, const uint8_t* evidence_buffer, size_t evidence_buffer_size, const uint8_t* endorsements_buffer, size_t endorsements_buffer_size, const oe_policy_t* policies, size_t policies_size, oe_claim_t** claims, size_t* claims_length)`.
-  - This entry point has the same definition as in the OE SDK V0.9 release.
+  - Entry point `oe_result_t (*verify_evidence)(oe_verifier_t* context, const oe_uuid_t* format_id, const uint8_t* evidence_buffer, size_t evidence_buffer_size, const uint8_t* endorsements_buffer, size_t endorsements_buffer_size, const oe_policy_t* policies, size_t policies_size, oe_claim_t** claims, size_t* claims_length)`.
+  - The optional parameter `format_id` is added, to specify the
```
yentsanglee, line 494: Each verifier plugin implements this function so the 'format id' is already known (you use the format id to find this plugin). Why adding it?

docs/DesignDocs/Attestation_API_Proposal.md
yentsanglee: I'm ok with outputting an empty custom claim buffer if it was not specified when the evidence was generated.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Yes, `<openenclave/attestation/sgx/evidence.h>` for those UUIDs is a public header.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: An single entry point can be invoked by multiple plugins, so its implementation needs to know which plugin calls it. The `format_id` serves that purpose.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: A zero-sized custom claims buffer is still valid, and is hashed in the `oe_get_evidence()` implementation. As a result, it still make more sense for `oe_verify_evidence()` to return a zero_sized custom claims entry. This is a statement by `oe_verify_evidence()` that there is a zero-sized custom claims buffer, instead of it being absent in the quote.

dthaler: What purpose would inserting a zero-sized claims buffer into the serialized evidence buffer, as opposed to omitting it?   Can we just declare it an invalid parameter to pass in a zero-sized buffer?

shnwc: `oe_get_evidence()` should allow its caller to not pass any custom claim (i.e. to pass an empty buffer). In this case, if `oe_verify_evidence()` does not return a custom claims entry with an empty buffer, could there be a vulnerability for an attacker to inject its favorite custom claims? I don't think there is vulnerability here, but like to be sure.

PR #3137, opened 2020-06-16
Fix #3130

### LINKED ISSUE #3130: Use of OE_BUILD_ENCLAVE in headers
This is the only occurrence of the macro `OE_BUILD_ENCLAVE` in the public (installed) header files: https://github.com/openenclave/openenclave/blob/4cc53259c32f2c1a1bd321d85b37049c9fed7efd/include/openenclave/attestation/attester.h#L15

This means that all client code also needs to define this macro if they want to use an attester (plugin). Are there plans to make this the default behavior and to use this macro in other places, or is this an unintended "leak" of internals?

shnwc:
Fix #3130

### DIFF
--- a/include/openenclave/attestation/attester.h
+++ b/include/openenclave/attestation/attester.h
@@ -12,11 +12,11 @@
 #ifndef _OE_ATTESTATION_ATTESTER_H
 #define _OE_ATTESTATION_ATTESTER_H
 
-#ifndef OE_BUILD_ENCLAVE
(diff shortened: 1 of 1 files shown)


dthaler:
I'd recommend using more descriptive titles than just referencing an issue number (which should still be put into the description, this PR puts it in both places)

shnwc:
@dthaler thanks for the feedback. I'll do that in future issues / PRs.

PR #3140, opened 2020-06-17
Fix the bug of using weak_alias between static and public functions on windows

mingweishih:
On windows, the two functions being weak_aliased (i.e., using `/alternatename`) need to have the same visibility (i.e., both are static or public). However, we currently apply the weak_alias between a static function and a public function. This causes a link-time error when the public function is not defined.

To clarify, see the following examples:
```
// foo_internal.c
static _foo () {...}
OE_WEAK(_foo, foo);
```
```
// foo.c
foo() {...}
```
---> Work fine. The `foo` defined in foo.c will be used.

However, if only `foo_internal.c` is used, there will be a link-time error as the `_foo` can not be used as a public function.

### DIFF
--- a/host/sgx/switchless.c
+++ b/host/sgx/switchless.c
@@ -24,6 +24,20 @@
  */
 #define OE_ENCLAVE_WORKER_SPIN_COUNT_THRESHOLD (4096U)
 
+/**
+ * Declare the protoype of the following functions to avoid missing-prototypes
+ * warning.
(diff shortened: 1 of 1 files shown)


mingweishih:
Yes. I tested by manually opting out the switchless.edl.  I'll make follow-up PRs for that.

mingweishih:
bors try

PR #3142, opened 2020-06-17
Clean up oesign python test scripts

CodeMonkeyLeet:
- Per comments in PR #3128, update sign-and-verify.py to pass a list
  of arguments into `subprocess.call()` instead of a formatted string
  and remove the use of the `shell=True` option.
  - Update the callers of sign-and-verify.py to pass in argument lists.
- Mark appropriate arguments in sign-and-verify.py and
  make-oesign-config.py as `required=True` instead of manually checking
  for missing arguments and failing the scripts.


CodeMonkeyLeet:
bors try

PR #3144, opened 2020-06-17
Fix memory allocator intialization with EEID

wintersteiger:
This fixes the memory allocator heap size during enclave initialization. @anakrish: the recent sn/dlmalloc changes had the effect that dlmalloc always failed in EEID enclaves, because it thought there is 0 heap. This would not be necessary if the allocator would be initialized after `_eeid_patch_memory`, but I'm not sure whether that's feasible/desired.

CC @jhand2 

### DIFF
--- a/enclave/core/sgx/calls.c
+++ b/enclave/core/sgx/calls.c
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.
 
 #include "../calls.h"
-#include <openenclave/bits/eeid.h>
 #include <openenclave/bits/sgx/sgxtypes.h>
 #include <openenclave/corelibc/stdlib.h>
 #include <openenclave/corelibc/string.h>
@@ -11,7 +10,6 @@
 #include <openenclave/internal/allocator.h>
 #include <openenclave/internal/atomic.h>
 #include <openenclave/internal/calls.h>
-#include <openenclave/internal/eeid.h>
 #include <openenclave/internal/fault.h>
 #include <openenclave/internal/globals.h>
 #include <openenclave/internal/jump.h>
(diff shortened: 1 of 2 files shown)


wintersteiger:
bors try

wintersteiger:
(I've been granted bors try permission before, but something seems to be wrong with it.)

anakrish:
bors try

enclave/core/sgx/calls.c
```diff
@@ -159,6 +159,10 @@ static oe_result_t _eeid_patch_memory()
 
         // Wipe the marker page
         memset(heap_base, 0, OE_PAGE_SIZE);
+
+        // Re-initialize the memory allocator with updated heap size
+        oe_allocator_init(
```
anakrish, line 164: @wintersteiger Thanks for identifying the issue.

It is not safe to call `oe_allocator_init` here since we will end up calling it again when the enclave thread is initialized.
https://github.com/openenclave/openenclave/blob/90ab2aa8aec6768d9c9b0208073b8d9cdb8f6933/enclave/core/sgx/threadlocal.c#L377

It seems to me that patching the enclave memory is better done as one of the first things during enclave initialization.
A good place to do that would be here: https://github.com/openenclave/openenclave/blob/e90dd1268a8771c8ccc7de360d74d7b745f7d741/enclave/core/sgx/init.c#L47

wintersteiger: Yeah, it's intended to be done as the first thing. I'll take another look and move it to the place you suggested.

wintersteiger: Done! Thanks for the suggestion again, I think this is definitely a better place to do this.

anakrish:
bors try

PR #3146, opened 2020-06-17
Use the assembly versions of memcpy/memset/memmove on sgx build

mingweishih:
We originally use the platform-independent versions of memcpy/memset/memmove from MUSL (see [here](https://github.com/openenclave/openenclave/tree/master/3rdparty/musl/musl/src/string) for the files). For better performance, this PR switches the SGX build to use the assembly versions of them (under x86_64 folder). This also aligns with MUSL when it's built on x86_64 platform. Note that  the optee build will still use the platform-independent versions.

Some number for the reference:

Running an ecall that does memcpy on two 64000-byte buffers.
Using the platform-independent version of `memcpy` (w/ -O3): 37596.45 cycle
Using the assembly version of `memcpy`: 25541.41 cycle

### DIFF
--- a/enclave/core/CMakeLists.txt
+++ b/enclave/core/CMakeLists.txt
@@ -75,6 +75,9 @@ if (OE_SGX)
   list(
     APPEND
     PLATFORM_SRC
+    ${MUSL_SRC_DIR}/string/x86_64/memcpy.s
+    ${MUSL_SRC_DIR}/string/x86_64/memmove.s
+    ${MUSL_SRC_DIR}/string/x86_64/memset.s
     ../../common/sgx/endorsements.c
     ../../common/sgx/rand.S
     sgx/arena.c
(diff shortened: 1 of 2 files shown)


mingweishih:
bors try

mingweishih:
> LGTM.
> Thanks for identifying and fixing this!
> Can you take a complete pass and see if there are other functions that MUSL uses assembly for, but we have chosen to use C? If you don't have time, it would be good to capture the above in an issue.

I'll open an issue for the follow-up. Thanks!

nonpolarity:
Nice! For my knowledge, how can you measure the cycles? I thought that is a definite integer.

mingweishih:
> Nice! For my knowledge, how can you measure the cycles? I thought that is a definite integer.

Using rdtsc on the host. For example

a = rdtsc();
ecall();
b = rdtsc();
printf("%lu\n", b - a);

nonpolarity:
> 
> 
> > Nice! For my knowledge, how can you measure the cycles? I thought that is a definite integer.
> 
> Using rdtsc on the host. For example
> 
> a = rdtsc();
> ecall();
> b = rdtsc();
> printf("%lu\n", b - a);

Thanks! You are right, time can be a float.

CodeMonkeyLeet:
@HernanGatta You might want to take a look at this since this PR did not switch the OP-TEE builds to using the assembly implementation (MUSL does have ARM assembly as well).

PR #3150, opened 2020-06-17
Add vectored exception handling in simulation mode on Linux

### LINKED ISSUE #2124: SGX-LKL-PORT: OE exception handling in simulation mode
OE currently only supports exception handling in hardware mode (as the implementation is SGX specific).

During development/testing, we heavily rely on simulation mode, and therefore want to avoid big divergences between the two modes. Exception handling is an important enough feature that we require it in simulation mode.

If OE doesn't handle exceptions in simulation mode, we would have to implement our own mechanism for that, entirely bypassing the OE exception handling API when in simulation mode. (While the implementation of enclave exceptions in simulation mode would be quite different from hardware mode, it should also be simpler.)

nonpolarity:
By mock of aex and eresume in SGX mode, vectored exception handling is enabled.
partially fix #2124 since now it is enabled on Linux only.

### DIFF
--- a/host/CMakeLists.txt
+++ b/host/CMakeLists.txt
@@ -282,10 +282,17 @@ if (OE_SGX)
       PLATFORM_SDK_ONLY_SRC
       sgx/linux/aep.S
       sgx/enter.c
+      sgx/exception_sim.c
       sgx/linux/exception.c
       sgx/linux/sgxioctl.c
       sgx/linux/switchless.c
       sgx/linux/xstate.c)
+
+    # Functions in sgx/exception_sim.c will change the fs register and may trigger
+    # stack check fail, thus it is necessary to turn off stack check.
+    set_source_files_properties(sgx/exception_sim.c
+                                PROPERTIES COMPILE_FLAGS -fno-stack-protector)
+
   else ()
     list(APPEND PLATFORM_HOST_ONLY_SRC sgx/windows/sgxquoteproviderloader.c)
 
--- a/host/sgx/calls.c
+++ b/host/sgx/calls.c
@@ -114,6 +114,9 @@ static oe_result_t _enter_sim(
     td = (oe_sgx_td_t*)(enclave->addr + tcs->gsbase);
     td->simulate = true;
 
+    td->host_fsbase = (uint64_t)oe_get_fs_register_base();
+    td->host_gsbase = (uint64_t)oe_get_gs_register_base();
+
     /* Call into enclave */
     if (arg3)
         *arg3 = 0;
--- a/host/sgx/exception.h
+++ b/host/sgx/exception.h
@@ -6,6 +6,9 @@
 
 #include <openenclave/bits/types.h>
 #include <openenclave/internal/calls.h>
+#if defined(__linux__)
+#include <ucontext.h>
+#endif
 
 typedef struct _host_exception_context
 {
@@ -19,5 +22,13 @@ void oe_initialize_host_exception(void);
 
 /* Platform neutral exception handler */
 uint64_t oe_host_handle_exception(oe_host_exception_context_t* context);
+
+#if defined(__linux__)
+/* Check if the current enclave is in simulation mode. */
+bool is_simulation(void);
+
+/* Exception handler in simulation mode on Linux. */
+uint64_t oe_host_handle_exception_sim(ucontext_t* context, int sig_num);
+#endif
 
 #endif // _OE_HOST_EXCEPTION_H
--- a/host/sgx/exception_sim.c
+++ b/host/sgx/exception_sim.c
@@ -1,0 +1,220 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/host.h>
+#include <openenclave/internal/registers.h>
+#include <openenclave/internal/sgx/td.h>
+#include <signal.h>
+#include <ucontext.h>
+#include "enclave.h"
+#include "exception.h"
+
+bool is_simulation(void)
+{
+    // In simulation mode the fs still points to td.
+    oe_sgx_td_t* td = (oe_sgx_td_t*)oe_get_fs_register_base();
+    return (
+        td && td->magic == TD_MAGIC && td->base.self_addr == (uint64_t)td &&
+        td->simulate);
+}
+
+static void _oe_aex_sim(sgx_tcs_t* tcs)
+{
+    // Update cssa as AEX does in real mode.
+    tcs->cssa++;
+
+    // Change the FS/GS segment registers to host side.
+    oe_sgx_td_t* td =
+        (oe_sgx_td_t*)((unsigned char*)tcs + OE_TD_FROM_TCS_BYTE_OFFSET);
+    oe_set_fs_register_base((const void*)(td->host_fsbase));
+    oe_set_gs_register_base((const void*)(td->host_gsbase));
+}
+
+static void _oe_eresume_sim(sgx_tcs_t* tcs)
+{
+    // Update cssa as ERESUME does in real mode.
+    tcs->cssa--;
+
+    // Change the FS/GS segment registers to enclave side.
+    oe_enclave_t* enclave = oe_query_enclave_instance((void*)tcs);
+    uint64_t enclave_start = enclave->addr;
+    uint64_t enclave_fsbase = enclave_start + tcs->fsbase;
+    uint64_t enclave_gsbase = enclave_start + tcs->gsbase;
+    oe_set_fs_register_base((const void*)enclave_fsbase);
+    oe_set_gs_register_base((const void*)enclave_gsbase);
+}
+
+static sgx_ssa_gpr_t* _get_ssa_gpr(sgx_tcs_t* tcs)
+{
+    uint32_t cssa = tcs->cssa;
+    oe_sgx_td_t* td =
+        (oe_sgx_td_t*)((unsigned char*)tcs + OE_TD_FROM_TCS_BYTE_OFFSET);
+    uint64_t ssa_frame_size = td->base.__ssa_frame_size;
+    if (ssa_frame_size == 0)
+    {
+        ssa_frame_size = OE_DEFAULT_SSA_FRAME_SIZE;
+    }
+
+    unsigned char* ssa_base_address =
+        (unsigned char*)tcs + OE_SSA_FROM_TCS_BYTE_OFFSET;
+
+    // cssa always points to the unfilled ssa.
+    return (
+        sgx_ssa_gpr_t*)(ssa_base_address + cssa * ssa_frame_size * OE_PAGE_SIZE - OE_SGX_GPR_BYTE_SIZE);
+}
+
+static void _update_ssa_from_context(ucontext_t* context, sgx_tcs_t* tcs)
+{
+    sgx_ssa_gpr_t* ssa_gpr = _get_ssa_gpr(tcs);
+
+    // Update gpr.
+    ssa_gpr->rax = (uint64_t)(context->uc_mcontext.gregs[REG_RAX]);
+    ssa_gpr->rbx = (uint64_t)(context->uc_mcontext.gregs[REG_RBX]);
+    ssa_gpr->rcx = (uint64_t)(context->uc_mcontext.gregs[REG_RCX]);
+    ssa_gpr->rdx = (uint64_t)(context->uc_mcontext.gregs[REG_RDX]);
+    ssa_gpr->rsp = (uint64_t)(context->uc_mcontext.gregs[REG_RSP]);
+    ssa_gpr->rbp = (uint64_t)(context->uc_mcontext.gregs[REG_RBP]);
+    ssa_gpr->rsi = (uint64_t)(context->uc_mcontext.gregs[REG_RSI]);
+    ssa_gpr->rdi = (uint64_t)(context->uc_mcontext.gregs[REG_RDI]);
+    ssa_gpr->r8 = (uint64_t)(context->uc_mcontext.gregs[REG_R8]);
+    ssa_gpr->r9 = (uint64_t)(context->uc_mcontext.gregs[REG_R9]);
+    ssa_gpr->r10 = (uint64_t)(context->uc_mcontext.gregs[REG_R10]);
+    ssa_gpr->r11 = (uint64_t)(context->uc_mcontext.gregs[REG_R11]);
+    ssa_gpr->r12 = (uint64_t)(context->uc_mcontext.gregs[REG_R12]);
+    ssa_gpr->r13 = (uint64_t)(context->uc_mcontext.gregs[REG_R13]);
+    ssa_gpr->r14 = (uint64_t)(context->uc_mcontext.gregs[REG_R14]);
+    ssa_gpr->r15 = (uint64_t)(context->uc_mcontext.gregs[REG_R15]);
+    ssa_gpr->rip = (uint64_t)(context->uc_mcontext.gregs[REG_RIP]);
+    ssa_gpr->rflags = (uint64_t)(context->uc_mcontext.gregs[REG_EFL]);
+
+    // This flag is checked by virtual dispacher.
+    ssa_gpr->exit_info.as_fields.valid = true;
+}
+
+static void _update_sgx_vector(int sig_num, sgx_tcs_t* tcs)
+{
+    // Hardcode here must match g_vector_to_exception_code_mapping[] in
+    // enclave/core/sgx/exception.c
+    uint32_t sgx_vector = OE_EXCEPTION_UNKNOWN;
+    switch (sig_num)
+    {
+        case SIGFPE: // OE_EXCEPTION_DIVIDE_BY_ZERO
+            sgx_vector = 0;
+            break;
+        case SIGTRAP: // OE_EXCEPTION_BREAKPOIN
+            sgx_vector = 3;
+            break;
+        case SIGILL: // OE_EXCEPTION_ILLEGAL_INSTRUCTION
+            sgx_vector = 6;
+            break;
+        case SIGSEGV: // OE_EXCEPTION_ACCESS_VIOLATION
+            sgx_vector = 13;
+            break;
+        case SIGBUS: // OE_EXCEPTION_MISALIGNMENT
+            sgx_vector = 17;
+            break;
+    }
+
+    sgx_ssa_gpr_t* ssa_gpr = _get_ssa_gpr(tcs);
+    // Check struct sgx_exit_info for the detail that lhs is 8 bits only.
+    ssa_gpr->exit_info.as_fields.vector = (uint8_t)sgx_vector;
+}
+
+static void _update_context_from_ssa(ucontext_t* context, sgx_tcs_t* tcs)
+{
+    sgx_ssa_gpr_t* ssa_gpr = _get_ssa_gpr(tcs);
+
+    context->uc_mcontext.gregs[REG_RAX] = (greg_t)ssa_gpr->rax;
+    context->uc_mcontext.gregs[REG_RBX] = (greg_t)ssa_gpr->rbx;
+    context->uc_mcontext.gregs[REG_RCX] = (greg_t)ssa_gpr->rcx;
+    context->uc_mcontext.gregs[REG_RDX] = (greg_t)ssa_gpr->rdx;
+    context->uc_mcontext.gregs[REG_RSP] = (greg_t)ssa_gpr->rsp;
+    context->uc_mcontext.gregs[REG_RBP] = (greg_t)ssa_gpr->rbp;
+    context->uc_mcontext.gregs[REG_RSI] = (greg_t)ssa_gpr->rsi;
+    context->uc_mcontext.gregs[REG_RDI] = (greg_t)ssa_gpr->rdi;
+    context->uc_mcontext.gregs[REG_R8] = (greg_t)ssa_gpr->r8;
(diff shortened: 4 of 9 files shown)


host/sgx/exception_sim.c
```diff
@@ -0,0 +1,226 @@
...
+#endif
+#include "enclave.h"
+#include "exception.h"
+
+#if defined(_WIN32)
+#else
+bool is_simulate(oe_host_exception_context_t* context)
+{
+    uint64_t tcs_address = context->rbx;
+    oe_enclave_t* enclave = oe_query_enclave_instance((void*)tcs_address);
```
anakrish, line 20: Handle case where enclave could be NULL.

host/sgx/exception_sim.c
```diff
@@ -0,0 +1,226 @@
...
+}
+
+static void _oe_eresume_sim(ucontext_t* context, void* enclave_fs)
+{
+    // Update cssa as ERESUME does in real mode.
+    sgx_tcs_t* tcs = (sgx_tcs_t*)(context->uc_mcontext.gregs[REG_RBX]);
+    tcs->cssa--;
+
+    // Change the FS register to enclave side.
+    oe_set_fs_register_base(enclave_fs);
```
anakrish, line 146: Would be good to save and restore both fs and gs.

host/sgx/exception_sim.c
```diff
@@ -0,0 +1,226 @@
...
+    // Copy the data of context into ssa manually.
+    _oe_aex_sim(context, host_fs);
+    _update_ssa_from_context(context);
+    _update_sgx_vector(context, sig_num);
+
+    uint64_t ret = OE_EXCEPTION_CONTINUE_SEARCH;
+
+    // Check if the enclave exception happens inside the first pass
+    // exception handler.
+    oe_thread_binding_t* thread_data = oe_get_thread_binding();
```
anakrish, line 166: Instead of using td_t.simulate to store the host FSs, you could use oe_thread_binding_t to store both host fs and gs.

tests/VectorException/host/host.c
```diff
@@ -162,7 +156,13 @@ int main(int argc, const char* argv[])
     OE_TEST(enc_test_cpuid_in_global_constructors(enclave) == OE_OK);
 
     test_vector_exception(enclave);
-    test_sigill_handling(enclave);
+
+    // SGX specific test is not available in simulation.
```
anakrish, line 160: Expand the comment to explain why test is SGX specific. Looks like it uses CPUID.

anakrish:
@mingweishih This is a very interesting PR. Alvin has enabled signal handling to work in simulation mode. Can you also take a look and provide suggestions.

host/sgx/exception_sim.c
radhikaj: For my understanding, When is enclave NULL?

anakrish: It could happen in case where the signal is raised outside the enclave, even before an ecall is made. We register global signal handlers that will get called, and it is better to not make an assumption that an enclave exists.

host/sgx/exception_sim.c
```diff
@@ -0,0 +1,226 @@
...
+            break;
+        case SIGBUS: // OE_EXCEPTION_MISALIGNMENT
+            sgx_vector = 17;
+            break;
+        case SIGFPE: // OE_EXCEPTION_DIVIDE_BY_ZERO
+            sgx_vector = 0;
+            break;
+        case SIGSEGV: // OE_EXCEPTION_ACCESS_VIOLATION
+            sgx_vector = 13;
+            break;
```
radhikaj, line 106: Is there a default to handle here?

nonpolarity: The default option is no handler available(sgx_vector = -1). Only these signals and some other signals with handlers registered by user can be processed by OE, others are handed over to system.

host/sgx/exception_sim.c
```diff
@@ -0,0 +1,449 @@
...
+#include <signal.h>
+#if defined(_WIN32)
+#else
+#include <ucontext.h>
+#endif
+#include "enclave.h"
+#include "exception.h"
+
+#if defined(_WIN32)
+bool is_simulate(struct _EXCEPTION_POINTERS* exception_pointers)
```
dthaler, line 16: "is simulate" seems like bad grammar to me

```suggestion
bool is_simulation(struct _EXCEPTION_POINTERS* exception_pointers)
```

For comparison, sgx\create.c has: 
`enclave->simulate = oe_sgx_is_simulation_load_context(context);`
and sgx\sgxload.c has:
`bool is_simulation`
both of which use "is simulation" not "is simulate"

nonpolarity: Suggestion is applied, thanks!

host/sgx/exception_sim.c
```diff
@@ -0,0 +1,449 @@
...
+
+    // This flag is checked by virtual dispacher.
+    ssa_gpr->exit_info.as_fields.valid = true;
+}
+
+static void _update_sgx_vector(ucontext_t* context, int sig_num)
+{
+    // Hardcode here must match g_vector_to_exception_code_mapping[] in
+    // enclave/core/sgx/exception.c
+    uint32_t sgx_vector = (uint32_t)-1;
```
mingweishih, line 312: For readability, consider defining the `OE_EXCEPTION_UNKNOWN` in this file and add comments that it was defined the same as in exception.h.

host/sgx/exception_sim.c
```diff
@@ -0,0 +1,449 @@
...
+
+    // This flag is checked by virtual dispacher.
+    ssa_gpr->exit_info.as_fields.valid = true;
+}
+
+static void _update_sgx_vector(PCONTEXT context, PEXCEPTION_RECORD exceptionrecord)
+{
+    // Hardcode here must match g_vector_to_exception_code_mapping[] in
+    // enclave/core/sgx/exception.c
+    uint32_t sgx_vector = (uint32_t)-1;
```
mingweishih, line 89: For readability, consider defining the `OE_EXCEPTION_UNKNOWN` in this file and add comments that it was defined the same as in exception.h.

host/sgx/linux/exception.c
```diff
@@ -39,7 +39,15 @@ static void _host_signal_handler(
...
     // Call platform neutral handler.
-    uint64_t action = oe_host_handle_exception(&host_context);
+    uint64_t action;
+    if (!is_simulation(&host_context))
+    {
+        action = oe_host_handle_exception(&host_context);
+    }
+    else
+    {
+        action = oe_host_handle_exception_sim(context, sig_num);
```
mingweishih, line 49: What about windows?

host/sgx/exception_sim.c
```diff
@@ -0,0 +1,449 @@
...
+        // Reset the flag
+        thread_data->flags &= (~_OE_THREAD_HANDLING_EXCEPTION);
+        if (result == OE_OK && arg_out == OE_EXCEPTION_CONTINUE_EXECUTION)
+        {
+            // This exception has been handled by the enclave. Let's resume.
+            ret = OE_EXCEPTION_CONTINUE_EXECUTION;
+        }
+        else
+        {
+            // Un-handled enclave exception happened.
```
mingweishih, line 429:
```suggestion
            // Unhandled enclave exception happened.
```

host/sgx/exception.h
```diff
@@ -20,4 +24,18 @@ void oe_initialize_host_exception(void);
 /* Platform neutral exception handler */
 uint64_t oe_host_handle_exception(oe_host_exception_context_t* context);
 
+#if defined(_WIN32)
+/* Exception handler in simulation mode on Windows */
+uint64_t oe_host_handle_exception_sim(struct _EXCEPTION_POINTERS* exception_pointers);
+
+/* Check if the current enclave is in  simulation mode. */
```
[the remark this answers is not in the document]
nonpolarity: Suggestion is committed, thanks!

host/sgx/exception.h
```diff
@@ -20,4 +24,18 @@ void oe_initialize_host_exception(void);
...
+bool is_simulate(struct _EXCEPTION_POINTERS* exception_pointers);
+#else
+/* Exception handler in simulation mode on Linux */
+uint64_t oe_host_handle_exception_sim(ucontext_t* context, int sig_num);
+
+/* Check if the current enclave is in  simulation mode. */
```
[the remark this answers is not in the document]
nonpolarity: Suggestion is committed, thanks!

host/sgx/exception.h
```diff
@@ -20,4 +24,18 @@ void oe_initialize_host_exception(void);
...
+bool is_simulate(oe_host_exception_context_t* context);
```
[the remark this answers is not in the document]
nonpolarity: Updated, thanks!

host/sgx/exception.h
```diff
@@ -20,4 +24,18 @@ void oe_initialize_host_exception(void);
 /* Platform neutral exception handler */
 uint64_t oe_host_handle_exception(oe_host_exception_context_t* context);
 
+#if defined(_WIN32)
+/* Exception handler in simulation mode on Windows */
+uint64_t oe_host_handle_exception_sim(struct _EXCEPTION_POINTERS* exception_pointers);
+
+/* Check if the current enclave is in  simulation mode. */
+bool is_simulate(struct _EXCEPTION_POINTERS* exception_pointers);
```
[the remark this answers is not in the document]
nonpolarity: Updated, thanks!

host/sgx/exception_sim.c
```diff
@@ -0,0 +1,449 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/host.h>
+#include <openenclave/internal/registers.h>
+#include <openenclave/internal/sgx/td.h>
+#include <signal.h>
+#if defined(_WIN32)
```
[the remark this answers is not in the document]
nonpolarity: Fixed. I thought there may be some Windows specific header files. Thanks!

host/sgx/exception.h
```diff
@@ -20,4 +24,18 @@ void oe_initialize_host_exception(void);
 /* Platform neutral exception handler */
 uint64_t oe_host_handle_exception(oe_host_exception_context_t* context);
 
+#if defined(_WIN32)
+/* Exception handler in simulation mode on Windows */
```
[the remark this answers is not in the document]
nonpolarity: Suggestion is committed, thanks!

host/sgx/exception.h
```diff
@@ -6,6 +6,10 @@
 
 #include <openenclave/bits/types.h>
 #include <openenclave/internal/calls.h>
+#if defined(_WIN32)
```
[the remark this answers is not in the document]
nonpolarity: Fixed, I thought there may be some Windows specific header files. Thanks!

host/sgx/exception.h
```diff
@@ -20,4 +24,18 @@ void oe_initialize_host_exception(void);
...
 uint64_t oe_host_handle_exception(oe_host_exception_context_t* context);
 
+#if defined(_WIN32)
+/* Exception handler in simulation mode on Windows */
+uint64_t oe_host_handle_exception_sim(struct _EXCEPTION_POINTERS* exception_pointers);
+
+/* Check if the current enclave is in  simulation mode. */
+bool is_simulate(struct _EXCEPTION_POINTERS* exception_pointers);
+#else
+/* Exception handler in simulation mode on Linux */
```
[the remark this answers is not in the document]
nonpolarity: Suggestion is committed, thanks!

host/sgx/linux/exception.c
nonpolarity: The source file for Windows is another one host/sgx/windows/exception.c  , thanks!

host/sgx/exception_sim.c
nonpolarity: Fixed, thanks!

host/sgx/exception_sim.c
nonpolarity: Fixed, I did not notice there is OE_EXCEPTION_UNKNOWN. Thanks!

host/sgx/exception_sim.c
nonpolarity: Fixed, I did not notice there is OE_EXCEPTION_UNKNOWN. Thanks!

host/sgx/linux/exception.c
mingweishih: I didn't see any changes to the `host/sgx/windows/exception.c` in this PR. Is it intended?

host/sgx/exception_sim.c
nonpolarity: Fixed, thanks!

host/sgx/exception_sim.c
nonpolarity: Good suggestion, now both fs/gs are processed, just like oe_enter_sim.

host/sgx/exception_sim.c
nonpolarity: Fixed, thanks!

host/sgx/exception_sim.c
```diff
@@ -0,0 +1,449 @@
...
+#endif
+#include "enclave.h"
+#include "exception.h"
+
+#if defined(_WIN32)
+bool is_simulation(struct _EXCEPTION_POINTERS* exception_pointers)
+{
+    PCONTEXT context = exception_pointers->ContextRecord;
+    uint64_t tcs_address = context->Rbx;
+    oe_enclave_t* enclave = oe_query_enclave_instance((void*)tcs_address);
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks!

host/sgx/exception_sim.c
```diff
@@ -0,0 +1,449 @@
...
+        }
+    }
+    else
+    {
+        // Not an exclave exception.
+        // Continue searching for other handlers.
+        ret = OE_EXCEPTION_CONTINUE_SEARCH;
+    }
+
+    // Since aex was deferred, eresume must be advanced, to keep the status
```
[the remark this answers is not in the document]
nonpolarity: Fixed, thanks!

host/sgx/linux/exception.c
nonpolarity: Sorry it is still in progress, now it is pushed.

tests/VectorException/host/host.c
nonpolarity: Hi @anakrish I am not sure if this is the feature of SGX:
https://github.com/openenclave/openenclave/blob/f5952133bfe54bd97dfaa6af691ddfaea3c1f5a3/tests/VectorException/enc/sigill_handling.c#L73
while exception handling, in simulation mode rbx is not equal to tcs, but in SGX mode rbx is still tcs, although the above line succeed in both case. The different behaviors lead to different test result.

host/sgx/create.c
```diff
@@ -816,6 +817,10 @@ oe_result_t oe_sgx_build_enclave(
     enclave->size = enclave_size;
     enclave->text = enclave_addr + oeimage.text_rva;
 
+    /* Save the FS/GS segment registers of host side */
+    enclave->host_fsbase = (uint64_t)oe_get_fs_register_base();
```
anakrish, line 821: Each thread will have different FS, GS values. It seems one thread's FS, GS value will overwrite another threads.

nonpolarity: Thanks for your catch, now fs/gs are saved to td.

nonpolarity:
bors try

nonpolarity:
> `VectorException`

Yes this change is covered by VectorException. I am trying to add more sig test cases.

nonpolarity:
> 
> 
> Given the PR is scoped down to support Linux only, I think it's good to go. However, I echo @anakrish that do we have existing test (e.g., the `VectorException` test) that covers this changes? Or do we need new tests?

Yes this change is covered by VectorException. I am trying to add more sig test cases.

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

host/sgx/calls.c
```diff
@@ -114,6 +114,9 @@ static oe_result_t _enter_sim(
     td = (oe_sgx_td_t*)(enclave->addr + tcs->gsbase);
     td->simulate = true;
 
+    td->host_fsbase = (uint64_t)oe_get_fs_register_base();
```
anakrish, line 117: We still seem to be using td_t

nonpolarity:
> It would be better to avoid change to sgx_td_t since it is a data structure used by hardware.

Thanks, I am trying a separate data structure for simulation only.

host/sgx/exception_sim.c
```diff
@@ -0,0 +1,214 @@
...
+            sgx_vector = 13;
+            break;
+        case SIGBUS: // OE_EXCEPTION_MISALIGNMENT
+            sgx_vector = 17;
+            break;
+    }
+
+    sgx_tcs_t* tcs = (sgx_tcs_t*)(context->uc_mcontext.gregs[REG_RBX]);
+    sgx_ssa_gpr_t* ssa_gpr = _get_ssa_gpr(tcs);
+    ssa_gpr->exit_info.as_fields.vector = sgx_vector;
```
mingweishih, line 116: I though this need to changed; i.e., cast the `sgx_vector` to char/unsigned char.

nonpolarity:
> 
> 
> It would be better to avoid change to sgx_td_t since it is a data structure used by hardware.

@anakrish now the change to sgx_td_t is two uint64_t, and these value are only assigned while in simulation mode. So in SGX mode there is hardly any performance impact.

nonpolarity:
bors try

host/sgx/exception_sim.c
nonpolarity: Fixed, thanks!

host/sgx/exception_sim.c
```diff
@@ -0,0 +1,228 @@
...
+#include <openenclave/internal/sgx/td.h>
+#include <signal.h>
+#include <ucontext.h>
+#include "enclave.h"
+#include "exception.h"
+
+bool is_simulation(void)
+{
+    // In simulation mode the fs still points to td.
+    oe_sgx_td_t* td = (oe_sgx_td_t*)oe_get_fs_register_base();
```
anakrish, line 15: Can you add the following scenario to the test:
- Add a host side exception handler
- Introduce a host side exception that will be handled by the above handler.

I'm curious whether the expression `td->magic == TD_MAGIC` is safe when FS contains host-side FS value (ie when the exception happens in the host)

nonpolarity:
bors try

nonpolarity:
bors try

nonpolarity:
bors try

host/sgx/exception_sim.c
jhand2: It's probably safer to keep a host-side global for simulation mode right? This also avoids differences in how fs is managed on Linux and Windows.

jhand2: > I'm curious whether the expression td->magic == TD_MAGIC is safe when FS contains host-side FS value (ie when the exception happens in the host)

I agree that if we are worried about fs possibly being changed then this de-reference is unsafe.

anakrish: > It's probably safer to keep a host-side global for simulation mode right? This also avoids differences in how fs is managed on Linux and Windows.

Yes, a global (spin-lock guarded or lock-free) linked-list that gets populated only for simulation mode might be the most reliable option.

nonpolarity:
bors try

host/sgx/exception_sim.c
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    enclave_start = enclave->addr;
+    enclave_end = enclave->addr + enclave->size;
+    rip = (uint64_t)(context->uc_mcontext.gregs[REG_RIP]);
+    if (rip >= enclave_start && rip < enclave_end)
+    {
+        // Simulate the AEX in SGX hardware mode.
+        // Copy the data of context into ssa manually.
+        _oe_aex_sim(tcs);
```
CodeMonkeyLeet, line 175: There's actually a lot of stuff that AEX does which probably is not directly relevant to exception handling, but it makes me nervous because I don't understand the degree of accuracy it needs to be simulated at to not cause other kinds of issues downstream. A couple of things off the top of my head:

* AEX preserves XSAVE state in addition to GPRs. 
* AEX sets the context to a synthetic state that the host sees.

@mingweishih if he knows of any specific implications.

host/sgx/exception_sim.c
CodeMonkeyLeet: Dumb question: shouldn't only a single thread be accessing `td` at any given point? A second thread with a different FS encountering a host exception wouldn't reference the same FS base or TD, no?

host/CMakeLists.txt
```diff
@@ -282,10 +282,17 @@ if (OE_SGX)
...
       sgx/enter.c
+      sgx/exception_sim.c
       sgx/linux/exception.c
       sgx/linux/sgxioctl.c
       sgx/linux/switchless.c
       sgx/linux/xstate.c)
+
+    # Functions in sgx/exception_sim.c will change the fs register and may trigger
+    # stack check fail, thus it is necessary to turn off stack check.
+    set_source_files_properties(sgx/exception_sim.c
```
CodeMonkeyLeet, line 293: For my understanding, can you elaborate on the mechanism of failure? Is it because stack protection injects a cookie at the exception handler address, and so sees the enclave FS before we have the opportunity to fix it back up to the host value inside the exception handler itself?

Issue #2818, opened 2020-04-08
A few ocalls used by oecore cannot be opted out of


jhand2:
In particular, the following edl files have ocalls which are always linked into oecore and cannot be opted out of:

* sgx/attestation.edl
* sgx/cpu.edl
* sgx/debug.edl
* sgx/switchless.edl
* logging.edl

mingweishih:
With #3023, #3151 and #3203, now only sgx/attestation.edl, sgx/cpu.edl cannot be opt-out.
The opt-out of the sgx/cpu.edl is tracked by #3025.
Regarding sgx/attestation.edl, I'd like to get feedback to see whether it makes sense to be opt-out. More specifically, do we need to support an enclave without the attestation support?

cc @CodeMonkeyLeet @jhand2 @yentsanglee @radhikaj @anakrish

dthaler:
I think it should be supported, but it's not high priority.   Other libOS approaches like Enarx don't support attestation by the app (the deployment mechanism does the attestation), so I don't think it's a blocker in general.

mingweishih:
Close given #3214 is merged.

PR #3151, opened 2020-06-18
Opt-out logging.edl by default

mingweishih:
This PR solves part of #2818; i.e., removing the limitation that the `logging.edl` cannot be excluded.
The PR consists of four parts:
1. Add the default implementations of ecalls/ocalls (defined by `logging.edl`), and make these implementations weak. As a result, when the user does not opt in the edl, these implementations will be used. If the user opts in the edl, the implementations in the oeedger8r-generated source will be used instead.
2. Update the edls under `tests`; removing the import of  `logging.edl` if not necessary.
    Note that most of the test cases still need the `oe_write_ocall` in the edl, which is used by `oe_host_printf` or `OE_TEST`

3. Update the edls in the samples. Samples don't import logging.edl by default.

4. Update the docuementation.

### DIFF
--- a/enclave/core/hostcalls.c
+++ b/enclave/core/hostcalls.c
@@ -18,8 +18,8 @@
  * Declare the protoype of the following functions to avoid the
  * missing-prototypes warning.
  */
+oe_result_t _oe_log_is_supported_ocall();
 oe_result_t _oe_log_ocall(uint32_t log_level, const char* message);
-
 oe_result_t _oe_write_ocall(int device, const char* str, size_t maxlen);
 
 /**
@@ -29,6 +29,12 @@ oe_result_t _oe_write_ocall(int device, const char* str, size_t maxlen);
  * the implementions (which are strong) in the oeedger8r-generated code will be
  * used.
  */
+oe_result_t _oe_log_is_supported_ocall()
+{
+    return OE_UNSUPPORTED;
+}
+OE_WEAK_ALIAS(_oe_log_is_supported_ocall, oe_log_is_supported_ocall);
+
 oe_result_t _oe_log_ocall(uint32_t log_level, const char* message)
 {
     OE_UNUSED(log_level);
--- a/enclave/core/hostcalls.c
+++ b/enclave/core/hostcalls.c
@@ -13,6 +13,40 @@
 
 #include "core_t.h"
 
+#if !defined(OE_USE_BUILTIN_EDL)
+/**
+ * Declare the protoype of the following functions to avoid the
+ * missing-prototypes warning.
+ */
+oe_result_t _oe_log_ocall(uint32_t log_level, const char* message);
+
+oe_result_t _oe_write_ocall(int device, const char* str, size_t maxlen);
+
+/**
+ * Make the following OCALLs weak to support the system EDL opt-in.
+ * When the user does not opt in (import) the EDL, the linker will picked
+ * the following default implementations. If the user opts in the EDL,
+ * the implementions (which are strong) in the oeedger8r-generated code will be
+ * used.
+ */
+oe_result_t _oe_log_ocall(uint32_t log_level, const char* message)
+{
+    OE_UNUSED(log_level);
+    OE_UNUSED(message);
+    return OE_UNSUPPORTED;
+}
+OE_WEAK_ALIAS(_oe_log_ocall, oe_log_ocall);
+
+oe_result_t _oe_write_ocall(int device, const char* str, size_t maxlen)
+{
+    OE_UNUSED(device);
+    OE_UNUSED(str);
+    OE_UNUSED(maxlen);
+    return OE_UNSUPPORTED;
+}
+OE_WEAK_ALIAS(_oe_write_ocall, oe_write_ocall);
+#endif
+
 void* oe_host_malloc(size_t size)
 {
     uint64_t arg_in = size;
--- a/host/sgx/create.c
+++ b/host/sgx/create.c
@@ -1008,7 +1008,13 @@ oe_result_t oe_create_enclave(
     OE_CHECK(_initialize_enclave(enclave));
 
     /* Setup logging configuration */
-    oe_log_enclave_init(enclave);
+    if (oe_log_enclave_init(enclave) == OE_UNSUPPORTED)
+    {
+        oe_log(
+            OE_LOG_LEVEL_WARNING,
+            "The in-enclave logging is not supported. To enable, please import "
+            "the logging.edl.\n");
+    }
 
     /* Apply the list of settings to the enclave.
      * This may initialize switchless manager too.
--- a/host/traceh_enclave.c
+++ b/host/traceh_enclave.c
@@ -18,6 +18,38 @@
 
 #include "core_u.h"
 
+#if !defined(OE_USE_BUILTIN_EDL)
+/**
+ * Declare the protoype of the following functions to avoid the
+ * missing-prototypes warning.
+ */
+oe_result_t _oe_log_init_ecall(
+    oe_enclave_t* enclave,
+    const char* enclave_path,
+    uint32_t log_level);
+
+/**
+ * Make the following ECALL weak to support the system EDL opt-in.
+ * When the user does not opt in (import) the EDL, the linker will picked
+ * the following default implementation. If the user opts in the EDL,
+ * the implementions (which also weak) in the oeedger8r-generated code will be
+ * used. This behavior is guaranteed by the linker; i.e., the linker will pick
+ * the symbols defined in the object before those in the library.
+ */
+oe_result_t _oe_log_init_ecall(
+    oe_enclave_t* enclave,
+    const char* enclave_path,
+    uint32_t log_level)
+{
+    OE_UNUSED(enclave);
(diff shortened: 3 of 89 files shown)


host/traceh_enclave.c
```diff
@@ -18,6 +18,38 @@
...
+oe_result_t _oe_log_init_ecall(
+    oe_enclave_t* enclave,
+    const char* enclave_path,
+    uint32_t log_level);
+
+/**
+ * Make the following ECALL weak to support the system EDL opt-in.
+ * When the user does not opt in (import) the EDL, the linker will picked
+ * the following default implementation. If the user opts in the EDL,
+ * the implementions (which also weak) in the oeedger8r-generated code will be
```
mingweishih, line 35: Just note that we currently plan to make all the ecalls weak in order to realize the design described by #3086. See the PoC in https://github.com/openenclave/oeedger8r-cpp/pull/24.

mingweishih:
bors try

enclave/core/hostcalls.c
```diff
@@ -13,6 +13,40 @@
 
 #include "core_t.h"
 
+#if !defined(OE_USE_BUILTIN_EDL)
+/**
+ * Declare the protoype of the following functions to avoid the
+ * missing-prototypes warning.
+ */
+oe_result_t _oe_log_ocall(uint32_t log_level, const char* message);
```
anakrish, line 21: It would be better to generate these prototypes via the `--header-only` edger8r option.
Otherwise these prototypes here can go out of sync with logging.edl

enclave/core/hostcalls.c
```diff
@@ -13,6 +13,40 @@
...
+ * the implementions (which are strong) in the oeedger8r-generated code will be
+ * used.
+ */
+oe_result_t _oe_log_ocall(uint32_t log_level, const char* message)
+{
+    OE_UNUSED(log_level);
+    OE_UNUSED(message);
+    return OE_UNSUPPORTED;
+}
+OE_WEAK_ALIAS(_oe_log_ocall, oe_log_ocall);
```
anakrish, line 38: Is WEAK_ALIAS needed? The user has to pass liboe_logging_host.a or equivalent library to the linker line. Therefore those definitions would get picked up.
Can these stubs be entirely removed?

mingweishih:
Side note. There is no good way to inform users that certain internal functions that depdend on the opted-out edl do not work. It's particularly tricky for the logging.edl because we cannot use oe_host_printf.

host/traceh_enclave.c
```diff
@@ -18,6 +18,38 @@
...
+ * used. This behavior is guaranteed by the linker; i.e., the linker will pick
+ * the symbols defined in the object before those in the library.
+ */
+oe_result_t _oe_log_init_ecall(
```
anakrish, line 39: Are these stubs needed if an edl does not import logging.edl?

enclave/core/hostcalls.c
mingweishih: This forward declaration is for the internal implemention so that we can make it weak alias with the public symbol generated by  oeeger8r. This means that the name of this internal implementation needs to be different than what oeeger8r generates if we use the same edl.

enclave/core/hostcalls.c
mingweishih: What do you mean by liboe_logging_host.a? I may missed something. Are we going to create a new library?

host/traceh_enclave.c
mingweishih: Yes. If the edl does not import, the linker will use this implementation to resolve internal references.

jhand2:
I have concerns about merging then for basically the reason you mentioned above. If a user forgets to import logging.edl all logging calls will fail, and we can’t even warn them about it because we can’t log anything. 

I’m mostly ok with this approach for any other feature, but I don’t know if this is workable for this specific feature. Need to think some more

enclave/core/hostcalls.c
jhand2: @anakrish this pr is just to allow users to opt out of logging altogether. It doesn’t attempt to allow users to swap in their own logging framework (although you could technically do that with a similar mechanism)

jhand2: @anakrish this pr is just to allow users to opt out of logging altogether. It doesn’t attempt to allow users to swap in their own logging framework (although you could technically do that with a similar mechanism)

enclave/core/hostcalls.c
```diff
@@ -13,6 +13,40 @@
...
+
+oe_result_t _oe_write_ocall(int device, const char* str, size_t maxlen)
```
anakrish, line 40: One interesting scenario is what would happen if one enclave supports logging and another enclave does not support logging. Since the weak-symbol will be overwritten by the strong symbol, does this mean that both the enclaves will now have logging?

mingweishih: Good point. That may be the case for ecall and I need to think more about that.
For the ocall, this is fine because the enclave is compiled separately.
For example,

`enclave_foo` (does not import logging.edl) --- `_oe_write_ocall` will be picked by linker, which does nothing.
`enclave_bar` (imports the logging.edl) --- `oe_write_ocall` defined in the `_t.c` will be picked.

anakrish:
This is what I had in mind for a completely opt-in model:
https://github.com/mingweishih/openenclave/compare/opt-out-logging...anakrish:ak_logging_optin

In addition to importing logging.edl, the user must also link against a library say "oe_logging_enc.a" otherwise there will be link errors.

include/openenclave/edl/logging.edl
```diff
@@ -24,6 +24,8 @@ enclave
 
     untrusted
     {
+        void oe_log_is_supported_ocall();
```
anakrish, line 27: @mingweishih This is a nice idea. Consider if we can get rid of `oe_log_init_ecall` and make it an ocall that the enclave calls during initialization. That would change logging initialization from a "push" (host makes ecall) to a "pull" model (enclave makes ocall to initialize). That will get rid of ecalls that the SDK makes for logging.

host/sgx/create.c
```diff
@@ -1008,7 +1008,13 @@ oe_result_t oe_create_enclave(
     OE_CHECK(_initialize_enclave(enclave));
 
     /* Setup logging configuration */
-    oe_log_enclave_init(enclave);
+    if (oe_log_enclave_init(enclave) == OE_UNSUPPORTED)
+    {
+        oe_log(
+            OE_LOG_LEVEL_WARNING,
+            "The in-enclave logging is not supported. To enable, please import "
+            "the logging.edl.\n");
```
anakrish, line 1016: "please import logging.edl."

enclave/core/hostcalls.c
```diff
@@ -13,6 +13,46 @@
...
+oe_result_t _oe_write_ocall(int device, const char* str, size_t maxlen);
+
+/**
+ * Make the following OCALLs weak to support the system EDL opt-in.
+ * When the user does not opt in (import) the EDL, the linker will picked
+ * the following default implementations. If the user opts in the EDL,
+ * the implementions (which are strong) in the oeedger8r-generated code will be
+ * used.
+ */
+oe_result_t _oe_log_is_supported_ocall()
```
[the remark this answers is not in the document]
jhand2: All functions generated from EDL return oe_result_t. If you specify a return type then the return value is passed as an out parameter in the first argument.

so the edl signature
```int sample_ocall();```
would turn into 
```oe_result_t sample_ocall(int* ret);```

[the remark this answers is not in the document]
Comment (CodeMonkeyLeet) [reply enclave/core/hostcalls.c:32]: Got it, thanks for the reminder that the caller side stubs wrap and return the `oe_result_t` for `oe_call_enclave_function()`/`oe_call_host_function()` invocations.

**Comment by CodeMonkeyLeet** [comment]
###

File: enclave/core/hostcalls.c [hunk 9]
```diff
@@ -13,6 +13,46 @@
...
+ * missing-prototypes warning.
+ */
+oe_result_t _oe_log_is_supported_ocall();
+oe_result_t _oe_log_ocall(uint32_t log_level, const char* message);
+oe_result_t _oe_write_ocall(int device, const char* str, size_t maxlen);
+
+/**
+ * Make the following OCALLs weak to support the system EDL opt-in.
+ * When the user does not opt in (import) the EDL, the linker will picked
+ * the following default implementations. If the user opts in the EDL,
```
dthaler, line 28:
```suggestion
 * When the user does not opt in (import) the EDL, the linker will pick
 * the following default implementations. If the user opts into the EDL,
```

host/traceh_enclave.c
```diff
@@ -18,6 +18,38 @@
...
+    oe_enclave_t* enclave,
+    const char* enclave_path,
+    uint32_t log_level);
+
+/**
+ * Make the following ECALL weak to support the system EDL opt-in.
+ * When the user does not opt in (import) the EDL, the linker will picked
+ * the following default implementation. If the user opts in the EDL,
```
dthaler, line 34:
```suggestion
 * When the user does not opt in (import) the EDL, the linker will pick
 * the following default implementation. If the user opts into the EDL,
```

Issue #2941, opened 2020-04-24
Fix SGX DCAP delay-load in oehost


jhand2:
The DCAP library (sgx_dcap_ql) is intended to be delay-loaded on the host so that it is only needed when attempting to perform attestation. However, the current implementation links directly against the library, so the delay loading mechanism is redundant; the library is already loaded.

On Linux, ldd shows that libsgx_dcap_ql is loaded at the time the host executable is started (note that the same behavior also exhibits on Windows):
```
$ ldd ./host/echo_host 
        ...
        libsgx_enclave_common.so.1 => /usr/lib/x86_64-linux-gnu/libsgx_enclave_common.so.1 (0x00007f6d54357000)
        libsgx_dcap_ql.so.1 => /usr/lib/x86_64-linux-gnu/libsgx_dcap_ql.so.1 (0x00007f6d54352000)
        ...
```

This should be fixed so that OE applications which do not use attestation can be run without DCAP libraries installed.

jhand2:
At a minimum, it would be ideal if attestation was a separate .a file on the host and apps could decide not to link it in. In that case an enclave app could choose to not use attestation without needing to recompile the SDK.

This situation could improve with the advent of attestation plugins, but the above would be a good first step.

radhikaj:
The Intel SGX SDK delay loads this library.

anakrish:
This issue confuses libdcap_quoteprov.so which oehost dynamically loads    https://github.com/openenclave/openenclave/blob/463cf14493ab0ffeca275f3b4f5ab4edccbf5106/host/sgx/linux/sgxquoteproviderloader.c#L26-L27

with libsgx_dcap_ql.so which oehost links against
https://github.com/openenclave/openenclave/blob/463cf14493ab0ffeca275f3b4f5ab4edccbf5106/host/CMakeLists.txt#L495


> However, the current implementation links directly against the library, so the delay loading mechanism is redundant; the library is already loaded.

No delay-loading mechanism exists for sgx_dcap_ql.so; linking against it does not make a non-existent delay-loading redundant.

I'm closing this issue since two different libraries are being mixed up.

However it would be good to dynamically load sgx_dcap_ql.so like we do for libsgx_quoteprov.so.
I feel this is best done by attestation team.
CC: @radhikaj

anakrish:
Here are the set of link errors when you remove the link time dependency on sgx_dcap_ql.so
[...]

PR #3159, opened 2020-06-20
IGNORE - For testing purposes

anakrish:


anakrish:
bors try

anakrish:
Confirmed that oehost.a has link time dependency on sgx_dcap_ql.so.
#2941

PR #3161, opened 2020-06-20
Allow a plugin attester for generating a custom report - Fixes #3139

### LINKED ISSUE #3139: Enabling attested TLS based on oe_get_evidence and oe_verify_evidence
The current attested TLS sample relies on `oe_generate_attestation_certificate` (which calls `oe_get_report` internally) and `oe_verify_attestation_certificate` (which calls `oe_verify_remote_report` internally). With the introduction of attestation plugin model and EEID, we should replace the SGX-specific APIs (oe_get/verify_report) with more flexible  `oe_get_evidence` and `oe_verify_evidence`, so that attested TLS can be established on top of non-sgx-specific reports, such as reports with EEID.

jxyang:
This PR allows a plugin attester to be used for generating a custom report which is embedded in a self-signed certificate (required by attested TLS). Also a plugin verifier can be used to validate such a certificate accordingly.  Fixes #3139 

Signed-off-by: Xuejun Yang <xuejya@microsoft.com>


jxyang:
bors try

jxyang:
bors try

common/sgx/tlsverifier.c
```diff
@@ -172,3 +197,121 @@ oe_result_t oe_verify_attestation_certificate(
...
+        "oe_cert_verify failed with error = %s\n",
+        oe_result_str(result));
+
+    //------------------------------------------------------------------------
+    // Validate the report's trustworthiness
+    //------------------------------------------------------------------------
+
+    // determine the size of the extension
+    if (oe_cert_find_extension(
+            &cert, (const char*)oid_oe_report, NULL, &report_size) !=
```
yentsanglee, line 257: New oid needs to be defined for embedded evidence.

common/sgx/tlsverifier.c
```diff
@@ -172,3 +197,121 @@ oe_result_t oe_verify_attestation_certificate(
...
+ * @param[in] claim_verify_callback callback routine for custom claim checking
+ * @param[in] arg an optional context pointer argument specified by the caller
+ * when setting callback
+ * @retval OE_OK on a successful validation
+ * @retval OE_VERIFY_FAILED on quote failure
+ * @retval OE_INVALID_PARAMETER At least one parameter is invalid
+ * @retval OE_FAILURE general failure
+ * @retval other appropriate error code
+ */
+oe_result_t oe_verify_plugin_attestation_certificate(
```
yentsanglee, line 221: Suggest 'oe_verify_attestation_evidence_certificate'. Same in all occurrences.

Reason:
1. Embedded attestation evidence is generated by oe_get_evidence().

common/sgx/tlsverifier.c
```diff
@@ -56,6 +57,30 @@ static oe_result_t verify_report_user_data(
...
+static oe_result_t verify_public_key_claim(
+    oe_claim_t* claims,
+    size_t claims_length,
+    uint8_t* pub_key_buf,
+    size_t pub_key_buf_size)
+{
+    oe_result_t result = OE_FAILURE;
+    for (int i = (int)claims_length - 1; i >= 0; i--)
+    {
+        if (oe_strcmp(claims[i].name, "public_key") == 0)
```
yentsanglee, line 70: Even though "public key" is used internally, it's a good idea to define it in a header file and with unique naming (e.g. "_oe_internal_public_key") so it does not conflict with other claim names defined in include/openenclave/bits/evidence.h.

yentsanglee, line 70: Similar to the other comment.. the claim input is a clear buffer and output is a specific name (see PR 3168).

shnwc:
Is this part of the efforts to convert all attestation samples from oe_get_report() to oe_get_evidence()?

For better coordination of the attestation related efforts by multiple developers, I suggest that we bring our early proposals to the SIG-attestation meeting, before implementation if possible.

common/sgx/tlsverifier.c
jxyang: I think it's fine to multiplex the oid for both quote-based report and evidence-based report. The buffer contains a oe_report_header_t that identifies which type it is. The added test case uses that to correctly verify a certificate based evidences.

jxyang:
@shnwc This is a clear gap that I am trying to fill for a customer scenario. I have talked to members of SIG-attestation before the implementation. And I am happy to present to the whole group.

common/sgx/tlsverifier.c
jxyang: Thanks for catching that. Will do.

common/sgx/tlsverifier.c
jxyang: I think the goal of this API is to allow different plugins to generate their own certificates. Evidence is just the means to achieve a generic representation of proof provided by different plugins.

jxyang: As this will become a new public API to OE, I am open to suggestions of better names.

jxyang:
bors try

common/sgx/tlsverifier.c
dthaler: I don't understand this PR enough to comment, is there a design doc or diagram?   E.g., is the certificate part of the evidence?   (It certainly is in CyRes attestation.)   If its a public API without _sgx_ in the name, then it should be applicable to other TEEs too, and I can't tell yet whether this PR only makes sense for the SGX plugin.

common/sgx/tlsverifier.c
```diff
@@ -172,3 +197,121 @@ oe_result_t oe_verify_attestation_certificate(
...
+ * @param[in] arg an optional context pointer argument specified by the caller
+ * when setting callback
+ * @retval OE_OK on a successful validation
+ * @retval OE_VERIFY_FAILED on quote failure
+ * @retval OE_INVALID_PARAMETER At least one parameter is invalid
+ * @retval OE_FAILURE general failure
+ * @retval other appropriate error code
+ */
+oe_result_t oe_verify_plugin_attestation_certificate(
+    uint8_t* cert_in_der,
```
dthaler, line 222: Coding style nit: expand "cert" -> "certificate" throughout.  Similarly "buf" -> "buffer"

include/openenclave/bits/evidence.h
```diff
@@ -78,6 +78,11 @@ extern const char* OE_REQUIRED_CLAIMS[OE_REQUIRED_CLAIMS_COUNT];
 // This array is needed for tests
 extern const char* OE_OPTIONAL_CLAIMS[OE_OPTIONAL_CLAIMS_COUNT];
 
+/**
+ * Additional claims related to TLS certificate
+ */
+#define OE_CERTIFICATE_PUBLIC_KEY "public_key"
```
dthaler, line 84: If this is a claim name then say so in the #define name.  Other claim names all start with OE_CLAIM_ 
Also the others are in lines 61-77, not separated by what they're "related to".   Put here, it implies it's neither required nor optional.  Not sure what that would mean or why it's in neither array.

include/openenclave/host_verify.h
```diff
@@ -83,6 +84,43 @@ oe_result_t oe_verify_attestation_certificate(
...
+ */
+typedef oe_result_t (*oe_claim_verify_callback_t)(
+    oe_claim_t* claims,
+    size_t claims_length,
+    void* arg);
+
+/**
+ * oe_verify_plugin_attestation_certificate
+ *
+ * This function perform a custom validation on the input certificate. This
```
dthaler, line 101:
```suggestion
 * This function performs a custom validation on the input certificate. This
```

include/openenclave/enclave.h
```diff
@@ -677,6 +711,43 @@ oe_result_t oe_verify_attestation_certificate(
...
+ */
+typedef oe_result_t (*oe_claim_verify_callback_t)(
+    oe_claim_t* claims,
+    size_t claims_length,
+    void* arg);
+
+/**
+ * oe_verify_plugin_attestation_certificate
+ *
+ * This function perform a custom validation on the input certificate. This
```
dthaler, line 728:
```suggestion
 * This function performs a custom validation on the input certificate. This
```

tests/attestation_plugin_cert/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+* oe_generate_plugin_attestation_certificate
+* oe_free_attestation_certificate, and
+* oe_verify_attestation_certificate
+
+Test scenario:
+
+- **Host Side**
+  1. Create an enclave
+  2. Issue an ecall (get_tls_cert) into enclave for getting a self-signed certificate embedded with a plugin attestation report of the enclave
```
dthaler, line 14: The term "report" is SGX specific.  Is this meant to be SGX specific?

For example, in some evidence formats the evidence is already a certificate (and thus is usable with TLS).  It sounds like the "embedded" point might be SGX-specific?

Also, some authentication protocols use JWTs or CWTs or instead of certificates per se.  Would this require we have separate public APIs for certificates, JWTs, CWTs, etc.?

include/openenclave/bits/evidence.h
jxyang: This is what OE attestation plugin framework calls `custom claims`. In fact, it's a stationary claim in attested TLS in the way that:

1) the attester generates a public/private key pair
2) the attester generates a report, using the hash of the public key as the report data
3) the attester generates a certificate with the report embedded.

On the relying party side:
1) the verifier validates the certificate is signed properly
2) the verifier extracts the report and validates it is signed by a TEE (SGX in this case)
3) the verifier validates the hash of the public key in the certificate matches the report data

With all 3 steps passed, the public key is used for subsequent TLS communications.

In short, this claim neither required nor optional for remote attestation purpose. But it is required for attested TLS or any certificate based attestation.

common/sgx/tlsverifier.c
CodeMonkeyLeet: I agree with @yentsanglee's reasoning and left a similar comment in the header files before I saw this one; either suggested name seems fine to me, though Yen Lee's is shorter. I did make some other related suggestions there as well (e.g. `get` vs. `generate`)

include/openenclave/enclave.h
```diff
@@ -636,6 +636,40 @@ oe_result_t oe_generate_attestation_certificate(
...
+ * @param[in] private_key_size The size of the private_key buffer
+ * @param[in] public_key a public key used as the certificate's subject key
+ * @param[in] public_key_size The size of the public_key buffer.
+ *
+ * @param[out] output_cert a pointer to buffer pointer
+ * @param[out] output_cert_size size of the buffer above
+ *
+ * @return OE_OK on success
+ */
+oe_result_t oe_generate_plugin_attestation_certificate(
```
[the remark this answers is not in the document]
dthaler: +1 to preferring "get" instead of "generate"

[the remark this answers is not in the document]
CodeMonkeyLeet: One additional observation, this PR will conflict with #3198, which changes `evidence_format` to `format_id` and also brings in the API change to `oe_verify_evidence` to add the explicit `format_id` parameter.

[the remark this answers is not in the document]
shnwc: @CodeMonkeyLeet thanks for catching the conflict. A few files in the `include` and `tests` directories are also changed in PR #3198, and likely will have conflicts.

@jxyang since this PR will invoke the new APIs implemented in PR #3198, can you wait till that PR is merged, and rebase to it?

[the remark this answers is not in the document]
jxyang: @shnwc Sure. Looks like your PR has more than enough approvals. What's stopping it from being merged?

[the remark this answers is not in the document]
shnwc: PR #3198 is merged now. Changes in public API takes longer go reach agreement.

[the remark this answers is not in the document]
jxyang: Thanks. I will rebase.

common/sgx/tlsverifier.c
jxyang: Thanks @CodeMonkeyLeet for the context. Added another oid X509_OID_FOR_EVIDENCE_EXT in report.h.

common/sgx/tlsverifier.c
jxyang: Renamed to `oe_get_attestation_certificate_with_evidence` as @CodeMonkeyLeet suggested. Also moved the declaration to attester.h

common/sgx/tlsverifier.c
jxyang: The sister functions such as `oe_verify_attestation_certificate` use `cert` and `buf` throughout. I prefer to keep the terms consistent.

tests/attestation_plugin_cert/README.md
jxyang: I replaced `report` with `evidence` to be TEE agnostic in this file. These two functions `oe_get_attestation_certificate_with_evidence` and `oe_verify_attestation_certificate_with_evidence`  could be applicable to non-SGX TEEs, but the plugins/formats we have so far are SGX specific.

include/openenclave/bits/evidence.h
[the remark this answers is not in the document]
jxyang: Moved to `openenclave/internal/cert.h`.

include/openenclave/enclave.h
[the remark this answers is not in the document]
jxyang: Why does #3198 changes `evidence_format` to `format_id`? Should this API keep using `format_id`?

include/openenclave/enclave.h
```diff
@@ -636,6 +636,40 @@ oe_result_t oe_generate_attestation_certificate(
     uint8_t** output_cert,
     size_t* output_cert_size);
 
+/**
```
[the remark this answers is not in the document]
jxyang: Done.

include/openenclave/enclave.h
```diff
@@ -636,6 +636,40 @@ oe_result_t oe_generate_attestation_certificate(
...
+ * oe_generate_plugin_attestation_certificate.
+ *
+ * This function generates a self-signed x.509 certificate with an embedded
+ * report generated by a plugin attester from the underlying enclave.
+ *
+ * @param[in] format_id format id for the plugin attester
```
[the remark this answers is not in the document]
jxyang: Thanks for the suggestion. Updated accordingly.

include/openenclave/enclave.h
```diff
@@ -677,6 +711,43 @@ oe_result_t oe_verify_attestation_certificate(
...
+ * @param[in] claim_verify_callback callback routine for custom claim checking
+ * @param[in] arg an optional context pointer argument specified by the caller
+ * when setting callback
+ * @retval OE_OK on a successful validation
+ * @retval OE_VERIFY_FAILED on quote failure
+ * @retval OE_INVALID_PARAMETER At least one parameter is invalid
+ * @retval OE_FAILURE general failure
+ * @retval other appropriate error code
+ */
+oe_result_t oe_verify_plugin_attestation_certificate(
```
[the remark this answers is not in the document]
jxyang: Moved.

include/openenclave/enclave.h
```diff
@@ -677,6 +711,43 @@ oe_result_t oe_verify_attestation_certificate(
     oe_identity_verify_callback_t enclave_identity_callback,
     void* arg);
 
+/**
+ * claim validation callback type
+ * @param[in] claims a pointer to an array of claims
+ * @param[in] claims_length length of the claims array
+ * @param[in] arg caller defined context
+ */
+typedef oe_result_t (*oe_claim_verify_callback_t)(
```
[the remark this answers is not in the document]
jxyang: Renamed.

jxyang:
Note the added test is failing because of #3264.

include/openenclave/enclave.h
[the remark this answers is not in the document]
jxyang: Renamed to `oe_verify_claims_callback_t`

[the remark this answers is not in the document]
dthaler: My opinion is that oe_<verb>_<noun> or oe_<verb>_<type>_<noun> is better than oe_<type>_<verb>_<noun>.   (FYI, the guidance for new APIs that ship in Windows is the same.)   oe_verify_claims_callback_t is good.

common/sgx/tlsverifier.c
```diff
@@ -172,3 +198,131 @@ oe_result_t oe_verify_attestation_certificate(
...
+ * @retval other appropriate error code
+ */
+oe_result_t oe_verify_attestation_certificate_with_evidence(
+    uint8_t* cert_in_der,
+    size_t cert_in_der_len,
+    oe_verify_claims_callback_t claim_verify_callback,
+    void* arg)
+{
+    oe_result_t result = OE_FAILURE;
+    oe_cert_t cert = {0};
```
dthaler, line 229: https://github.com/openenclave/openenclave/blob/master/docs/Community/DevelopmentGuide.md says:

> DON'T use abbreviations unless they are already well-known terms known by users (e.g., "app", "info"), or are already required for use by developers (e.g, "min", "max", "args"). Examples of bad use would be num_widgets instead of widget_count, and opt_widgets instead of option_widgets or optional_widgets.

So: cert -> certificate, pub -> public, buf -> buffer

enclave/tls_cert.c
```diff
@@ -169,6 +175,93 @@ oe_result_t oe_generate_attestation_certificate(
...
 
+/**
+ * oe_get_attestation_certificate_with_evidence.
+ *
+ * This function generates a self-signed x.509 certificate with embedded
+ * evidence generated by an attester plugin for the enclave.
+ *
+ * @param[in] evidence_format The format UUID of the evidence to be generated.
+ *
+ * @param[in] subject_name a string contains an X.509 distinguished
```
dthaler, line 186:
```suggestion
 * @param[in] subject_name a string containing an X.509 distinguished
```

enclave/tls_cert.c
```diff
@@ -169,6 +175,93 @@ oe_result_t oe_generate_attestation_certificate(
...
+ * name (DN) for customizing the generated certificate. This name is also used
+ * as the issuer name because this is a self-signed certificate
+ * See RFC5280 (https://tools.ietf.org/html/rfc5280) specification for details
```
dthaler, line 189:
```suggestion
 * See RFC5280 (https://tools.ietf.org/html/rfc5280) for details
```

include/openenclave/attestation/attester.h
```diff
@@ -113,6 +113,40 @@ oe_result_t oe_get_evidence(
...
 
+/**
+ * oe_get_attestation_certificate_with_evidence
+ *
+ * This function generates a self-signed x.509 certificate with embedded
+ * evidence generated by an attester plugin for the enclave.
+ *
+ * @param[in] evidence_format The format UUID of the evidence to be generated.
+ *
+ * @param[in] subject_name a string contains an X.509 distinguished
```
dthaler, line 124:
```suggestion
 * @param[in] subject_name a string containing an X.509 distinguished
```

include/openenclave/attestation/attester.h
```diff
@@ -113,6 +113,40 @@ oe_result_t oe_get_evidence(
...
+ * name (DN) for customizing the generated certificate. This name is also used
+ * as the issuer name because this is a self-signed certificate
+ * See RFC5280 (https://tools.ietf.org/html/rfc5280) specification for details
```
dthaler, line 127:
```suggestion
 * See RFC5280 (https://tools.ietf.org/html/rfc5280) for details
```

include/openenclave/attestation/attester.h
```diff
@@ -113,6 +113,40 @@ oe_result_t oe_get_evidence(
     uint8_t** endorsements_buffer,
     size_t* endorsements_buffer_size);
 
+/**
+ * oe_get_attestation_certificate_with_evidence
```
dthaler, line 117: My opinion is that this is not an "attestation certificate".   An "attestation certificate" would be one like in DICE where the certificate itself is the evidence format.   This is a certificate with evidence embedded in it.   Suggest something like `oe_get_self_signed_certificate_with_evidence`.

include/openenclave/attestation/verifier.h
```diff
@@ -165,6 +165,44 @@ oe_result_t oe_verify_evidence(
...
+ * @param[in] claims_length length of the claims array
+ * @param[in] arg caller defined context
+ */
+typedef oe_result_t (*oe_verify_claims_callback_t)(
+    oe_claim_t* claims,
+    size_t claims_length,
+    void* arg);
+
+/**
+ * oe_verify_attestation_certificate_with_evidence
```
dthaler, line 181: As before, I think this name should change.  e.g., `oe_verify_self_signed_certificate_with_evidence` or something similar.

include/openenclave/internal/report.h
```diff
@@ -80,6 +80,13 @@ OE_STATIC_ASSERT(
     }
 #define X509_OID_FOR_QUOTE_STRING "1.2.840.113556.10.1.1"
 
+// ISO(1).ANSI(2).USA(840).Microsoft(113556).ACC(10).Classes(1).Subclass(2)
+#define X509_OID_FOR_EVIDENCE_EXT                            \
```
dthaler, line 84: My understanding is that evidence defines/typedefs go in evidence.h
Whereas report.h is for legacy report defines/typedefs

In addition, I think the define needs "OE" or "OPENENCLAVE" in the name somewhere since you use this to indicate that the evidence has an OE-specific header in it (that holds the UUID).   If standard OIDs are defined in the future for standard evidence formats, this name would be misleading.

tests/attestation_plugin_cert/enc/enc.cpp
```diff
@@ -0,0 +1,336 @@
...
+            if (security_version < 1)
+            {
+                OE_TRACE_ERROR(
+                    "identity->security_version checking failed (%d)\n",
+                    security_version);
+                goto done;
+            }
+        }
+        // Dump an enclave's unique ID, signer ID and Product ID. They are
+        // MRENCLAVE, MRSIGNER and ISVPRODID for SGX enclaves. In a real
```
dthaler, line 52: In my opinion, SGX specific info should only appear in a file where "sgx" is in the filename or path to the file.  Much of the rest of this file is agnostic.  I'd recommend moving this function (and any other sgx-only code in this file) to a separate file with sgx in the filename.

tests/attestation_plugin_cert/host/host.cpp
```diff
@@ -0,0 +1,219 @@
...
+                goto done;
+            }
+        }
+        // Dump an enclave's unique ID, signer ID and Product ID. They are
+        // MRENCLAVE, MRSIGNER and ISVPRODID for SGX enclaves. In a real
+        // scenario, custom id checking should be done here
+        else if (
+            strcmp(claim->name, OE_CLAIM_SIGNER_ID) == 0 ||
+            strcmp(claim->name, OE_CLAIM_UNIQUE_ID) == 0 ||
+            strcmp(claim->name, OE_CLAIM_PRODUCT_ID) == 0)
```
dthaler, line 60: Same comment here about separating SGX-specific code from TEE_agnostic code, by moving SGX-only code into a separate file with sgx in the filename or directory name.

tests/attestation_plugin_cert/host/host.cpp
```diff
@@ -0,0 +1,219 @@
...
+}
+
+int main(int argc, const char* argv[])
+{
+#ifdef OE_LINK_SGX_DCAP_QL
+
+#ifdef _WIN32
+    /* This is a workaround for running in Visual Studio 2017 Test Explorer
+     * where the environment variables are not correctly propagated to the
+     * test. This is resolved in Visual Studio 2019 */
```
dthaler, line 148: Should this be inside an _MSC_VER test then?

tests/attestation_plugin_cert/host/host.cpp
```diff
@@ -0,0 +1,219 @@
...
+        BOOL success = SetEnvironmentVariableW(L"LOCALAPPDATA", local_path);
+        CoTaskMemFree(local_path);
+
+        if (!success)
+            exit(1);
+    }
+#endif
+
+    oe_result_t result;
+    oe_enclave_t* enclave = NULL;
```
dthaler, line 185:
```suggestion
    oe_enclave_t* enclave = nullptr;
```
cpp files should use the C++ keyword `nullptr` instead of NULL

tests/attestation_plugin_cert/host/host.cpp
```diff
@@ -0,0 +1,219 @@
...
+    OE_TEST(result == OE_OK);
+
+    OE_TRACE_INFO("free cert 0xx%p\n", cert);
+    // oe_verifier_shutdown();
+    free(cert);
+}
+
+int main(int argc, const char* argv[])
+{
+#ifdef OE_LINK_SGX_DCAP_QL
```
dthaler, line 143: Other PRs remove this define.  See PR #3273 and #3270 so will need to update this as they are merged

dthaler:
FYI, the main comment I consider a blocker is the name of the OID.   Everything else is a strong suggestion not a blocker.

tests/attestation_plugin_cert/host/host.cpp
jxyang: Thanks for the info. Removed.

tests/attestation_plugin_cert/host/host.cpp
jxyang: Agree nullptr is better. Changed.

tests/attestation_plugin_cert/host/host.cpp
jxyang: Removed the workaround since we moved to VS 2019 already.

tests/attestation_plugin_cert/enc/enc.cpp
jxyang: Renamed the function name to `sgx_enclave_claims_verifier`.

common/sgx/tlsverifier.c
jxyang: IMHO, this guideline is arbitrary because each person may have difference list of `well-known terms`. In addition, the other functions in OE use the above mentioned abbreviations extensively.

jxyang:
> FYI, the main comment I consider a blocker is the name of the OID. Everything else is a strong suggestion not a blocker.

I renamed it to `X509_OID_FOR_OE_EVIDENCE_EXT` and moved to `evidence.h`.

common/sgx/tlsverifier.c
dthaler: Yes the only terms that should be considered well known are ones explicitly mentioned as exceptions in the coding style guidelines (by default, nothing).   I agree that other functions in OE are bad and should be cleaned up too, but not in this PR.

jxyang:
bors try

jxyang:
bors try

common/sgx/tlsverifier.c
yentsanglee: I do not see the change in report.h?

yentsanglee:
> > FYI, the main comment I consider a blocker is the name of the OID. Everything else is a strong suggestion not a blocker.
> 
> I renamed it to `X509_OID_FOR_OE_EVIDENCE_EXT` and moved to `evidence.h`.

Ideally, the 2 X509 OIDs are to be defined in the same header file. Probably in a small header file such as oe_oid.h.

jxyang:
> > > FYI, the main comment I consider a blocker is the name of the OID. Everything else is a strong suggestion not a blocker.
> > 
> > 
> > I renamed it to `X509_OID_FOR_OE_EVIDENCE_EXT` and moved to `evidence.h`.
> 
> Ideally, the 2 X509 OIDs are to be defined in the same header file. Probably in a small header file such as oe_oid.h.

I see both suggestions are reasonable.

jxyang:
bors try

common/sgx/tlsverifier.c
```diff
@@ -56,6 +59,33 @@ static oe_result_t verify_report_user_data(
     return result;
 }
 
+// Verify there is a matched claim for the public key */
+static oe_result_t verify_public_key_claim(
```
dthaler, line 63: [development guide](https://github.com/openenclave/openenclave/blob/master/docs/Community/DevelopmentGuide.md) says

> The single underscore (_ ) is reserved for local definitions (static, file-scope definitions). e.g. static oe_result_t _parse_sgx_report_body(..).


```suggestion
static oe_result_t _verify_public_key_claim(
```

common/sgx/tlsverifier.c
```diff
@@ -172,3 +202,136 @@ oe_result_t oe_verify_attestation_certificate(
     oe_free(report);
     return result;
 }
+
+/**
+ * oe_verify_attestation_certificate_with_evidence
```
dthaler, line 207: This is a non-SGX specific function, but is in an SGX-only file (unlike oe_get_attestation_certificate_with_evidence which is already in a non-SGX specific file).  Move this outside the sgx directory for symmetry.

common/sgx/tlsverifier.c
```diff
@@ -172,3 +202,136 @@ oe_result_t oe_verify_attestation_certificate(
...
+ *
+ * This function performs a custom validation on the input certificate. This
+ * validation includes extracting attestation evidence extension from the
+ * certificate before validating this evidence. An optional
+ * claim_verify_callback could be passed in for a calling client to further
+ * validate the claims of the enclave creating the certificate.
+ * @param[in] cert_in_der a pointer to buffer holding certificate contents
```
dthaler, line 214: Recommend spelling out "cert", "pub", "buf" throughout this file per style guidelines

common/sgx/tlsverifier.c
```diff
@@ -172,3 +202,136 @@ oe_result_t oe_verify_attestation_certificate(
...
+    oe_verify_claims_callback_t claim_verify_callback,
+    void* arg)
+{
+    oe_result_t result = OE_FAILURE;
+    oe_cert_t cert = {0};
+    uint8_t* report = NULL;
+    size_t report_size = 0;
+    oe_report_header_t* header = NULL;
+    uint8_t* pub_key_buf = NULL;
+    size_t pub_key_buf_size = KEY_BUFF_SIZE;
```
dthaler, line 238: "buf" vs "BUFF" is similarly inconsistent in terms of # of F's

include/openenclave/attestation/attester.h
dthaler: Xuejun pointed out in IM that this name is consistent with the existing API oe_verify_attestation_certificate which I claim has the same problem, but the consistency means that this PR is ok and can discuss separately.

tests/attestation_plugin_cert/tls.edl
```diff
@@ -0,0 +1,17 @@
...
+    from "openenclave/edl/fcntl.edl" import *;
+#ifdef OE_SGX
+    from "openenclave/edl/sgx/platform.edl" import *;
+#else
+    from "openenclave/edl/optee/platform.edl" import *;
+#endif
+
+    trusted {
+        public oe_result_t get_tls_cert_signed_with_ec_key([out] unsigned char** data, [out] size_t* data_size);
+        public oe_result_t get_tls_cert_signed_with_rsa_key([out] unsigned char** data, [out] size_t* data_size);
```
dthaler, line 15:
```suggestion
        public oe_result_t get_tls_certificate_signed_with_ec_key([out] unsigned char** data, [out] size_t* data_size);
        public oe_result_t get_tls_certificate_signed_with_rsa_key([out] unsigned char** data, [out] size_t* data_size);
```

common/sgx/tlsverifier.c
jxyang: made it two Fs

common/sgx/tlsverifier.c
jxyang: Will keep to be compatible with existing code base.

tests/attestation_plugin_cert/tls.edl
jxyang: Will keep to be compatible with existing code base.

common/sgx/tlsverifier.c
jxyang: Moved it under common (non-sgx specific)

common/sgx/tlsverifier.c
jxyang: Done.

jxyang:
bors try

jxyang:
bors try

jxyang:
bors try-

jxyang:
bors try

common/attest_plugin.c
```diff
@@ -273,6 +306,119 @@ oe_result_t oe_verify_evidence(
...
+    {
+        OE_TRACE_WARNING(
+            "No claim_verify_callback provided in "
+            "oe_verify_attestation_certificate_with_evidence call",
+            NULL);
+    }
+
+done:
+    oe_free(pub_key_buff);
+    for (size_t i = 0; i < claims_length; i++)
```
CodeMonkeyLeet, line 411: Consider using the `oe_free_claims` method instead

jxyang: Thanks. Done.

PR #3163, opened 2020-06-22
Tweak oeedger8r tests

anakrish:
oeedger8r will emit marshalling struct/enums etc only in the relevant cpp
files. Tests that make compile time assertions on the marshalling structs
need the struct definition. Such tests are grouped into _wrapper files that
include the corresponding all_u.c/all_t.c files.

### DIFF
--- a/tests/oeedger8r/enc/CMakeLists.txt
+++ b/tests/oeedger8r/enc/CMakeLists.txt
@@ -41,21 +41,19 @@ add_enclave(
   e71cbbea-a638-4653-b46e-2e58a2ca3408
   CXX
   SOURCES
-  all_t.c
+  all_t.h
+  all_t_wrapper.cpp
   bar_t.h
   bar.cpp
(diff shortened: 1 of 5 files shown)


anakrish:
bors try

anakrish:
bors try

mingweishih:
Does this also need to be tested with newer version of the oeedger8r? I think the submodule still pointes to older version (without the changes you mentioned).

anakrish:
We can update the submodule once your TEE refactoring PR is merged in.
This is intended to make that submodule update process easier.
This change per se does not depend on a particular version of edger8r and can be checked in now.

mingweishih:
> We can update the submodule once your TEE refactoring PR is merged in.
> This is intended to make that submodule update process easier.
> This change per se does not depend on a particular version of edger8r and can be checked in now.

Got it! Thanks for the clarification.

PR #3164, opened 2020-06-22
Add check for claims == NULL in oe_verify_evidence

wintersteiger:
I'm not sure whether we should reject `claims == NULL` in `oe_verify_evidence` outright, as it makes it impossible to report the required claims. I added the check just for `_check_claims` but I think it may be better to make that an `OE_INVALID_PARAMETER`. 

### DIFF
--- a/common/attest_plugin.c
+++ b/common/attest_plugin.c
@@ -224,7 +224,7 @@ oe_result_t oe_verify_evidence(
         claims,
         claims_length));
 
-    if (!_check_claims(*claims, *claims_length))
+    if (claims && claims_length && !_check_claims(*claims, *claims_length))
     {
         verifier->free_claims(verifier, *claims, *claims_length);
         *claims = NULL;


dthaler:
> I'm not sure whether we should reject `claims == NULL` in `oe_verify_evidence` outright, as it makes it impossible to report the required claims. I added the check just for `_check_claims` but I think it may be better to make that an `OE_INVALID_PARAMETER`.
> 
> Signed-off-by: Christoph M. Wintersteiger [cwinter@microsoft.com](mailto:cwinter@microsoft.com)

Current the doxygen comments in verifier.h and plugin.h show it as a required parameter.

yentsanglee:
> Currently the doxygen comments in verifier.h and plugin.h show it as a required parameter.

It makes sense for callers to pass in null if they don't care about claims. I suggest we modify the doxygen comments.

wintersteiger:
That's how I got there too; I didn't care about the claims for a quick test, but then two seconds later I realized that I do in fact need them :-)

wintersteiger:
I feel this one won't get a clear decision. The current fix allows both, so the comments can say that it's required, but it will behave gracefully if someone tries anyways, so I'm in favor of this version.

mingweishih:
@dthaler  @yentsanglee 
do you still have concern/suggestion on this PR?

yentsanglee:
Open issue 3231 to review attestation public APIs. This PR looks good to me.

mingweishih:
Thanks! bors r+

PR #3171, opened 2020-06-23
Add static links to dependencies of v0.9.x

BRMcLaren:
Intel and OpenSSL dependencies change upon each release, and then the CDN's expire. Store the current v0.9.x dependencies in Azure storage for permanent storage.


BRMcLaren:
@radhikaj This is going into v0.9.x and needs to be cherry picked as we moved away from esy.. CI machines no longer have esy and it causes failures with the old oeedgr8r

Issue #3099, opened 2020-06-04
SGX: Testing: child_process_destroy test occassionally fails in CI


anakrish:
https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAzure-Linux/detail/Azure-Linux/608/pipeline

anakrish:
CC @yanxue-intel

anakrish:
https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAzure-Linux/detail/Azure-Linux/615/pipeline/72

220/224 Test #220: tests/child_process_destroy ...........................................................................................................Child aborted***Exception:   0.41 sec

Test failed: ../tests/child_process/host/host.cpp(114): main child_status == 0

yanxue-intel:
@jinghe-INTC Can you take a look?

radhikaj:
Let's disable this test and re-enable later since this is blocking PRs as of now.

jinghe-INTC:
Is there anyway to reproduce the problem manually? How to setup an environment of "ACC1804 gcc Debug LVI e2e" which seems to be the only platform with the problem?

anakrish:
@mingweishih Can you help with setting up an "ACC 1804 gcc Debug LVI" environment?
Yes, it does seem that in both the links, it is similar kind of environment that has the problem.

@BRMcLaren Is there some way to grep the CI logs to indeed confirm that issue does indeed only happen on the above mentioned platform?

PR #3172, opened 2020-06-23
reopen test case child_process_destroy

jinghe-INTC:
update case by adding one second's sleep

update comment

Signed-off-by: Jing He <jing.j.he@intel.com>

### DIFF
--- a/tests/child_process/CMakeLists.txt
+++ b/tests/child_process/CMakeLists.txt
@@ -10,9 +10,9 @@ endif ()
 add_enclave_test(tests/child_process_ecall child_process_host child_process_enc
                  0)
 
-# Disabling until #3099 is addressed.
-#add_enclave_test(tests/child_process_destroy child_process_host
-#                 child_process_enc 1)
+# related issue #3099
+add_enclave_test(tests/child_process_destroy child_process_host
+                 child_process_enc 1)
 
 add_enclave_test(tests/child_process_create child_process_host
                  child_process_enc 2)
--- a/tests/child_process/host/host.cpp
+++ b/tests/child_process/host/host.cpp
@@ -91,6 +91,7 @@ int main(int argc, const char* argv[])
         case DESTROY_IN_CHILD_PROCESS:
             if (pid == 0) // child process
             {
+                sleep(1);
                 result_child_process = oe_terminate_enclave(enclave);
                 OE_TEST(result_child_process == OE_OK);
                 enclave = nullptr;


jinghe-INTC:
bors try

jinghe-INTC:
bors try

anakrish:
@jinghe-INTC Does adding 1 second sleep fix #3099 ?

BRMcLaren:
bors try

BRMcLaren:
bors try

jinghe-INTC:
> @jinghe-INTC Does adding 1 second sleep fix #3099 ?

Yes. It looks working around the unstable issue. I will try to root cause it.

jinghe-INTC:
It does not have to be a sleeping before the destroying in the child process, some extra printing like "printf("%d\n", __LINE__);" will also make it stable. However, if the printing is after the destroying, the case is unstable like before.

mingweishih:
> It does not have to be a sleeping before the destroying in the child process, some extra printing like "printf("%d\n", **LINE**);" will also make it stable. However, if the printing is after the destroying, the case is unstable like before.

Does it also imply that making any ocall before termination should work?
Also, instead of sleep, can you do some dummy operations (without making any ocall) and see if that work? Thanks!

jinghe-INTC:
> > It does not have to be a sleeping before the destroying in the child process, some extra printing like "printf("%d\n", **LINE**);" will also make it stable. However, if the printing is after the destroying, the case is unstable like before.
> 
> Does it also imply that making any ocall before termination should work?
> Also, instead of sleep, can you do some dummy operations (without making any ocall) and see if that work? Thanks!

1. I tried to copy the same ecall from parent process before termination and it didn't make any difference, still unstable.
2. A printing of PID of child process before termination made it more stable, although there is still about 1~2% occurrences of segmentation fault.
3. A sleep(0) also made it more stable, although there is still about 1~2% occurrences of segmentation fault. Is it one of the dummy operations you meant?

mingweishih:
> > > It does not have to be a sleeping before the destroying in the child process, some extra printing like "printf("%d\n", **LINE**);" will also make it stable. However, if the printing is after the destroying, the case is unstable like before.
> > 
> > 
> > Does it also imply that making any ocall before termination should work?
> > Also, instead of sleep, can you do some dummy operations (without making any ocall) and see if that work? Thanks!
> 
> 1. I tried to copy the same ecall from parent process before termination and it didn't make any difference, still unstable.
> 2. A printing of PID of child process before termination made it more stable, although there is still about 1~2% occurrences of segmentation fault.
> 3. A sleep(0) also made it more stable, although there is still about 1~2% occurrences of segmentation fault. Is it one of the dummy operations you meant?

Ah, I thought the operations were inside the enclave.

PR #3175, opened 2020-06-23
Update design doc to reflect Sig-Arch Decision.

### LINKED ISSUE #3154: SGX: Ghost OCalls/System Calls Opt In: Separate libraries for system edls
In order to use a system library that makes ocalls, has ecalls, in Intel SDK the user must
- import the system edl 
- link a system library into the enclave and the host.

E.g: to use switchless calling, the user must import switchless.edl:
https://github.com/intel/linux-sgx/blob/0e71c227ada2174c6dd7557e424815a3f953564a/SampleCode/Switchless/Enclave/Enclave.edl#L36
```c
from "sgx_tswitchless.edl" import *;
```

and link lsgx_tswitchless.a into the enclave
https://github.com/intel/linux-sgx/blob/0e71c227ada2174c6dd7557e424815a3f953564a/SampleCode/Switchless/Makefile#L154
```bash
	-Wl,--whole-archive  -lsgx_tswitchless -l$(Trts_Library_Name) -Wl,--no-whole-archive \
```

and link sgx_uswitchless.a into the host:
https://github.com/intel/linux-sgx/blob/0e71c227ada2174c6dd7557e424815a3f953564a/SampleCode/Switchless/Makefile#L104
```c
                  -Wl,--whole-archive  -lsgx_uswitchless -Wl,--no-whole-archive \
```

The approved design for [Explicit Enclave Opt-in to System OCalls](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/system_ocall_opt_in.md) does not address whether OE SDK should also require the user to explicitly explicitly link against libraries that accompany the system edls.

It is not clear if this was discussed and resolved or was never discussed.

One one hand not having the user link libraries reduces the burden on the user. The linker will anyways only pull in what is needed. (Note: We use -ffunction-section compiler option and gc-sections linker option).

On the other hand, requiring the user to link in a library explicitly makes it clear that the functionality in a library is being imported. Additionally it allows using --whole-archive to for include initializers in a library that the linker may otherwise not import.

anakrish:
Contrary to SGX SDK, OE SDK will not necessarily refactor code associated with
system edls into separate libraries that the user will have to link.

This means that for features like switchless, the user does not need to link in
additional libraries; only import switchless.edl.

However, there exists many libraries that accompany system edl:
liboehostepoll.a  liboehostresolver.a  liboesyscall.a
liboehostfs.a     liboehostsock.a

These will remain.

fixes #3154

### DIFF
--- a/docs/DesignDocs/system_ocall_opt_in.md
+++ b/docs/DesignDocs/system_ocall_opt_in.md
@@ -23,12 +23,14 @@ Users can import a system OCall the same way any other EDL file would be importe
 if an enclave with EDL file `sample.edl` wants to use the function `oe_log_ocall()` they can do
 one of the following:
 
-*Note: SGX SDK requires the user to link specific libraries in addition to importing the system edl.
-For example, in order to use the switchless calling feature, the user must import sgx_tswitchless.edl,and
-link sgx_tswitchless.a in enclave and sgx_uswitchless.a in the host. In contrasst, OE SDK will only require
-the user to import switchless.edl, and not require the user to link new libraries for system edls.
-However, The current libraries for system elds  will remain. (e.g: liboehostepoll.a,  liboehostresolver.a, liboelibc.a,
-liboesyscall.a). See [Sig-Arch](https://hackmd.io/@aeva/oesdk-sig-arch#23-June-2020)
+*Note: Both SGX SDK for Linux and Intel SGX SDK for Windows require the user to link specific
+libraries in addition to importing the system edl. For example, in order to use the switchless
+calling feature in SGX SDK for Linux, the user must import sgx_tswitchless.edl, and link
+sgx_tswitchless.a in the enclave and sgx_uswitchless.a in the host. In contrasst, OE SDK will
+only require the user to import switchless.edl, and not require the user to link new
+libraries for system edls. However, the current libraries for system EDLs will remain.
+(i.e.: liboehostepoll.a,  liboehostresolver.a, liboelibc.a, liboesyscall.a).
+See [Sig-Arch](https://hackmd.io/@aeva/oesdk-sig-arch#23-June-2020)
 
 
 *Note: These examples assume that `${install_path}/include` is passed to edger8r as an EDL search path*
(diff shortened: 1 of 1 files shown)


docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -23,6 +23,14 @@ Users can import a system OCall the same way any other EDL file would be importe
 if an enclave with EDL file `sample.edl` wants to use the function `oe_log_ocall()` they can do
 one of the following:
 
+*Note: SGX SDK requires the user to link specific libraries in addition to importing the system edl.
+For example, in order to use the switchless calling feature, the user must import sgx_tswitchless.edl,and
+link sgx_tswitchless.a in enclave and sgx_uswitchless.a in the host. In contrasst, OE SDK will only require
+the user to import switchless.edl, and not require the user to link new libraries for system edls.
+However, The current libraries for system elds  will remain. (e.g: liboehostepoll.a,  liboehostresolver.a, liboelibc.a,
```
dthaler, line 30:
```suggestion
However, the current libraries for system EDLs will remain. (i.e.: liboehostepoll.a,  liboehostresolver.a, liboelibc.a,
```

docs/DesignDocs/system_ocall_opt_in.md
```diff
@@ -23,6 +23,14 @@ Users can import a system OCall the same way any other EDL file would be importe
 if an enclave with EDL file `sample.edl` wants to use the function `oe_log_ocall()` they can do
 one of the following:
 
+*Note: SGX SDK requires the user to link specific libraries in addition to importing the system edl.
```
mingweishih, line 26: Should we say Intel SGX SDK?

anakrish: I thought I heard in Sig-Arch that Intel SGX SDK has been contributed to CCC and is just SGX SDK.
CC: @dthaler

dthaler: Technically there's two different SDK's right?   There's the "SGX SDK for Linux" that is now a CCC project with that label on the CCC site.  And then there's the "Intel SGX SDK for Windows".   If this sentence applies to both, then maybe it should explicitly mention both, rather than incorrectly implying there's only one.

anakrish: Mentioned both SDKs.

PR #3180, opened 2020-06-25
Usability oriented pointer checks.

anakrish:
When an enclave pointer is passed by the host back into the enclave
either as an in or in/out parameter detect it an raise OE_INVALID_PARAMETER.

Note: OE SDK marshalling strategy does not require these pointers to be passed in.
It is not a security issue. The pointers are now being passed in just for doing
these checks.

Note: Similar usability checks for out parameters are not implemented since
that may require changes to oeedger8r.

### DIFF
--- a/include/openenclave/edger8r/common.h
+++ b/include/openenclave/edger8r/common.h
@@ -86,6 +86,17 @@ done:
         }                                                          \
     } while (0)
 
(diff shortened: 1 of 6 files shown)


radhikaj:
@nonpolarity @mingweishih  Please review

radhikaj:
Closing!

Issue #3073, opened 2020-05-26
Logging: Provide a way to capture/redirect internal logs


anakrish:
OE SDK currently provides a way for users to turn on internal logging for debug-enabled enclaves by setting the **OE_LOG_LEVEL** and **OE_LOG_DEVICE** environment variables.

There is also support for custom-logging format: [LoggingFormat.md](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/LoggingFormat.md)

Three separate teams/users have expressed the need to be able to capture and redirect the internal logs generated by OE SDK:

1."Some applications will have their own mechanism for communication between enclave and host, and will want the log messages to use that/piggy back on existing traffic (we are in that situation), so something that would be beneficial is a logging mechanism that clearly separates message generation from their delivery, and allows user-provided delivery." @achamayou  (CCF)

2."This is mainly for enabling diagnosing issues from ACC customers running apps using OE SDK.Ideally, we want a want to **redirect all the existing logging** (hopefully we could expand it) via a plug-in so that it could redirect it to Kusto." @soccerGB (ACC)

3."If you just let me see your logging strings raw, I can take it from there. We just need to have a way for OE to take a c**allback for logging.**" - Internal Microsoft team

**The ask is specifically to be able to capture/redirect internal logs**. This is NOT an ask to add Azure specific code to the OE SDK, or to add telemetry.

anakrish:
One user remarked that just providing the ability to install a `logging-callback` would satisfy their needs:

```c
Typedef void(oe_logging_callback*)(void *context, char *message);

oe_result oe_set_logging_callback(void *context, oe_logging_callback* callback);

void oe_set_logging_callback(void *context, oe_logging_callback *callback)
{
    If (oe_mutex_lock(&_log_lock) == OE_OK)
    {
        _oe_logging_callback = callback;
        _oe_logging_context = context;
        oe_mutex_unlock(&_log_lock);
    }
}
```
"Then in oe_log_message, somewhere around line 347 or so, add:"

```c
if (!_oe_logging_callback)
{
    _oe_logging_callback(_oe_logging_context, message);
}
```

"I think that’s all that needs to happen to satisfy our requirements."

anakrish:
Though @soccerGB used the word "plug-in" in the issue @2895, he clarified that it could be something as simple as providing a custom Ocall implementation for the ocalls that print the messages:

[...]

PR #3182, opened 2020-06-25
Add callback feature to oe_log_message().

nonpolarity:
Base on #3073, this feature is added.
So far this is a internal api only, and it may be changed after discussion.

User can define a customized log function as following:
void oe_log_function(void* context,
    bool is_enclave,
    const char* time,
    long int usecs,
    oe_log_level_t level,
    const char* message)
{
    ...
}
To register this log function:
oe_set_log_callback(context, (oe_log_callback)(oe_log_function));
oe_log_message() then will use oe_log_function to process messages.

### DIFF
--- a/host/traceh.c
+++ b/host/traceh.c
@@ -314,6 +314,23 @@ done:
     return result;
 }
 
+void* oe_log_context = NULL;
+oe_log_callback_t oe_log_callback = NULL;
+
+oe_result_t oe_log_set_callback(void* context, oe_log_callback_t callback)
+{
+    if (oe_mutex_lock(&_log_lock) == OE_OK)
+    {
+        oe_log_context = context;
+        oe_log_callback = callback;
+        oe_mutex_unlock(&_log_lock);
+
+        return OE_OK;
+    }
+
+    return OE_UNEXPECTED;
+}
+
 // This is an expensive operation, it involves acquiring lock
 // and file operation.
 void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
@@ -331,15 +348,25 @@ void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
     {
         initialize_log_config();
     }
-    if (_initialized)
-    {
-        if (level > _log_level)
-            return;
-    }
 
     // Take the log file lock.
     if (oe_mutex_lock(&_log_lock) == OE_OK)
     {
+        if (oe_log_callback)
+        {
+            (oe_log_callback)(
+                oe_log_context, is_enclave, time, usecs, level, message);
+            goto done;
+        }
+
+        if (_initialized)
+        {
+            if (level > _log_level)
+            {
+                goto done;
+            }
+        }
+
         if (_log_all_streams || !_use_log_file)
         {
             _write_message_to_stream(
@@ -441,6 +468,7 @@ void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
             fflush(log_file);
             fclose(log_file);
         }
+    done:
         // Release the log file lock.
         oe_mutex_unlock(&_log_lock);
     }
--- a/include/openenclave/internal/trace.h
+++ b/include/openenclave/internal/trace.h
@@ -28,6 +28,17 @@ extern oe_log_level_t _log_level;
 #define OE_MAX_FILENAME_LEN 256U
 
 #if !defined(OE_BUILD_ENCLAVE)
+typedef void (*oe_log_callback_t)(
+    void* context,
+    bool is_enclave,
+    const char* time,
+    long int usecs,
+    oe_log_level_t level,
+    const char* message);
+oe_result_t oe_log_set_callback(void* context, oe_log_callback_t callback);
+extern void* oe_log_context;
+extern oe_log_callback_t oe_log_callback;
+
 oe_result_t oe_log_enclave_init(oe_enclave_t* enclave);
 void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message);
 #endif


nonpolarity:
bors try

achamayou:
@nonpolarity is this related to #3073?

I'm confused by the line that suggests this is internal only, followed by a reference to a user. Is this intended to be exposed at some point?

It also seems like the ability to set user data on the context is missing. It's extremely likely that a user wanting to re-direct logging is going to want a handle to whatever they're directing to on every call.

nonpolarity:
@achamayou Sorry to make you confused. Yes it is related to #3073, so far I just add it as internal api, and would make changes based on discussion through [sig-arch](https://hackmd.io/@aeva/oesdk-sig-arch) before make it an external api.
Yes the main purpose is to redirect the logs, the input context would grant user more opportunities to manipulate the messages. I will add a doc later.

host/traceh.c
```diff
@@ -314,6 +314,23 @@ oe_result_t oe_log(oe_log_level_t level, const char* fmt, ...)
     return result;
 }
 
+void* _oe_log_context = NULL;
```
mingweishih, line 317: nit: global variables should be initialized already.

include/openenclave/internal/trace.h
```diff
@@ -28,6 +28,17 @@ extern oe_log_level_t _log_level;
...
 #if !defined(OE_BUILD_ENCLAVE)
+typedef void (*oe_log_callback)(
+    void* context,
+    bool is_enclave,
+    const char* time,
+    long int usecs,
+    oe_log_level_t level,
+    const char* message);
+oe_result_t oe_set_log_callback(void* context, oe_log_callback callback);
+extern void* _oe_log_context;
```
mingweishih, line 39: what is the intention of making this `extern`?

host/traceh.c
mingweishih: why not making these variable `static`?

host/traceh.c
```diff
@@ -331,6 +348,18 @@ void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
     {
         initialize_log_config();
     }
+
+    if (_oe_log_callback)
```
mingweishih, line 352: This check seems to be redundant. Or do you mean `_oe_log_context` (can `_oe_log_context` be `NULL`)?

host/traceh.c
nonpolarity: At least these variable also need to be visible to host/host.c, so they are global.

include/openenclave/internal/trace.h
nonpolarity: Yes definition in traceh.c and declaration in trace.h can make these variables global and avoid potential redefinition error.

host/traceh.c
nonpolarity: _oe_log_context can be null. Only while _oe_log_callback is set the messages are redirected, or messages will be processed in oe_log_message as before(no redirect)

host/traceh.c
```diff
@@ -314,6 +314,23 @@ oe_result_t oe_log(oe_log_level_t level, const char* fmt, ...)
...
+void* _oe_log_context = NULL;
+oe_log_callback _oe_log_callback = NULL;
+
+oe_result_t oe_set_log_callback(void* context, oe_log_callback callback)
+{
+    if (oe_mutex_lock(&_log_lock) == OE_OK)
+    {
+        _oe_log_context = context;
+        _oe_log_callback = callback;
+        oe_mutex_unlock(&_log_lock);
```
anakrish, line 326: There is a subtle thread-safety issue here. The compiler or the hardware could reorder the instructions to:
```c
_oe_log_callback = callback;
_oe_log_context = context;
```
This means that when another thread checks
```c
if (_oe_log_callback)
   ...
```

`_oe_log_context` may be NULL or might be the incorrect one.

host/traceh.c
```diff
@@ -331,6 +348,18 @@ void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
     {
         initialize_log_config();
     }
+
+    if (_oe_log_callback)
+    {
+        if (oe_mutex_lock(&_log_lock) == OE_OK)
```
anakrish, line 354: Can we try avoiding the lock for each message that is printed?

host/traceh.c
nonpolarity: Can the mutex lock work as a barrier?
https://github.com/openenclave/openenclave/blob/2615de41ac44e47c0f683017135ebeccccb7fd23/host/traceh.c#L322

host/traceh.c
nonpolarity: I think the lock can guarantee the sequence of the messages in destination. Although there are timestamps, if the messages are messed up then the log is not readable. Yes we need a better method to avoid performance lost.

host/traceh.c
anakrish: It is a barrier; but the instructions between `oe_mutex_lock` and `oe_mutex_unlock` could be reordered arbitrarily.
Normally this doesn't matter; but elsewhere we are testing `_oe_log_callback` outside a lock.

anakrish: It would be good to define what the contract is. Is the user allowed to change the logging callback or set it only once?

host/traceh.c
anakrish: It would be good to define what the contract is. Are we using the lock to serialize all the log messages or it it being used for thread-safety?

host/traceh.c
nonpolarity: Yes you are right, I didn't realize the the if check is not safe. Now it is inside a lock. Thanks!

host/traceh.c
nonpolarity: Yes serialization may be a side effect of the lock for thread-safety.

include/openenclave/internal/trace.h
```diff
@@ -28,6 +28,17 @@ extern oe_log_level_t _log_level;
 #define OE_MAX_FILENAME_LEN 256U
 
 #if !defined(OE_BUILD_ENCLAVE)
+typedef void (*oe_log_callback)(
```
CodeMonkeyLeet, line 31: Naming convention to disambiguate:
- `oe_log_callback_t` since this is a function typedef
- Non-static functions and variables shouldn't use the `_` prefix, so the variables become `oe_log_callback` and `oe_log_context`

host/traceh.c
```diff
@@ -314,6 +314,23 @@ oe_result_t oe_log(oe_log_level_t level, const char* fmt, ...)
...
+    if (oe_mutex_lock(&_log_lock) == OE_OK)
+    {
+        _oe_log_context = context;
+        _oe_log_callback = callback;
+        oe_mutex_unlock(&_log_lock);
+
+        return OE_OK;
+    }
+
+    return OE_FAILURE;
```
CodeMonkeyLeet, line 331: This part of the code should be unreachable except for dev error (i.e. calling `oe_mutex_lock()` with an invalid lock pointer) In general, I would encourage everyone to avoid returning errors that are not meaningful in exchange for something that expresses intent or helpful for debugging.
```suggestion
    return OE_UNEXPECTED;
```

Another option here is to capture and return the actual result from the call to `oe_mutex_lock()`

CodeMonkeyLeet: 8b89a97 seems to have been dropped in your rebase, can you put it back in?

include/openenclave/internal/trace.h
nonpolarity: Fixed, thanks!

host/traceh.c
nonpolarity: Thanks for your notice, now it is back.

nonpolarity:
> 
> 
> LGTM except that one of the changes seemed be rebased out in the most recent push. Also, remember to update your commit with DCO.

Fixed, thanks!

PR #3183, opened 2020-06-25
Install eeid_verifier.h

wintersteiger:
Forgot to install a header file in #3131. 

In general: should the plugin initializers be called by the infrastructure, when it receives a request to verify evidence from a non-initialized plugin UUID, or will this stay on the client side?


mingweishih:
Do we have a test/sample for these installed headers?

wintersteiger:
Not a self-contained one. I'm integrating it with SGX-LKL at the moment, that how's I figure these things out.

PR #3187, opened 2020-06-25
Update oeedger8r to the latest version

### LINKED ISSUE #2693: Enable deep copy by default
The recent work on refactoring the use of EDL (#2676) could benefit from the deep-copy feature in terms of security (as described in #2666). We should consider enabling the deep copy by default.

This issue is part of #1839, which tracks the progress of deep-copy feature.

### LINKED ISSUE #2698: Function IDs and ECALL/OCALL marshalling structs should be generated under _t.c / _u.c
Currently, `oeedger8er` generates Function IDs and ECALL/OCALL marshalling structs under `_u.h` / `_t.h` (was under `_args.h` before #2691). However, these data types are only used by the bridging functions that `_t.c` / `_u.c` implements and therefore should not be visible to files that include `_u.h` / `_t.h`. As a result, we should consider moving the definition of those data types to `_t.c` / `_u.c`.

### LINKED ISSUE #2833: Allow simple preprocessor directives in  EDL

mingweishih:
This PR updates the oeedger8r to the latest version and also updates the changelog accordingly.

The changes include:
- The Deep-copy feature of oeedger8r is now enabled by default (fixes #2693)
- The oeedger8r-generated header files now contain only the function prototypes (fixes #2698).
- Support the --user-prefix feature
-  Oeedger8r now supports a subset of C-style preprocessor directives (#ifdef, #ifndef, #else, #endif) (fixes #2833).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   was available only on SGX1 FLC machines.
 - oesign tool supports the new `digest` command and options for [2-step signing using the digest](
   docs/DesignDocs/oesign_digest_signing_support.md).
+- Oeedger8r now supports the --use-prefix feature.
+- Oeedger8r now supports a subset of C-style preprocessor directives (#ifdef, #ifndef, #else, #endif).
 
 ### Changed
 - Mark APIs in include/openenclave/attestation/sgx/attester.h and verifier.h as experimental.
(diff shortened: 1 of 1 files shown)


mingweishih:
bors try

Issue #2713, opened 2020-03-20
Remove requirement for stub ecall in EDL files


jhand2:
Currently oeedger8r requires EDL for an enclave to have at least one public ecall. This requirement seems unnecessary, there is no reason that an enclave must define an ecall.

In particular, removing this requirement would allow for an EDL file containing only structs/enums to be used to generate header files.

dthaler:
Agree, I would love to see that requirement removed, as I've had to have dummy ecalls for years (going back to when just using the Intel SGX SDK) for exactly this reason.  I always found it annoying to have to add unused APIs just to make edger8r happy.

anakrish:
If I am not mistaken, Intel SDK initializes the enclave (e.g calling global constructors) when the first ecall is made, where as in OE SDK the initialization happens as part of enclave creation.

It would be good to talk to Intel to see if we can relax this restriction in edl.

radhikaj:
@jhand2 please update the issue.

radhikaj:
@mwshan request you follow up on @anakrish's question above.

MWShan:
The Intel SDK (actually the untrusted run-time system) makes a specific call into the enclave to perform initialization.  This is not an explicitly defined ECall in the EDL.  This call will return with the results of the initialization; thus, you would need an explicit ECall to get back into the enclave.

Does the OE SDK allow the thread to remain within the enclave on an implicit initialization ECall?  

Maybe we could define an EDL directive which overrides the requirement for an ECall.

@jbdelcuv - do you see any reason why we need an ECall?

jbdelcuv:
The Intel SGX SDK initializes the enclave as part of enclave creation. We use a reserved ECALL identifier (-1) for that purpose. Global constructors are executed when the first (ISV) ECALL is made.
The Edger8r expects an enclave EDL file to have at least a public ECALL function. An enclave without an ECALL isn't very useful. :)
However, I do agree that we can lift such restriction for library EDL files, i.e. EDL files that are used to generate header files only (using the --header-only option) or are meant to get imported into enclave EDL files.

jhand2:
> However, I do agree that we can lift such restriction for library EDL files, i.e. EDL files that are used to generate header files only (using the --header-only option) or are meant to get imported into enclave EDL files.

Just to clarify, I mis-spoke previously. When passed the --header-only option, edger8r does relax this restriction.

anakrish:
[...]

PR #3189, opened 2020-06-26
Remove the stub ecall from syscall.edl

mingweishih:
This PR removes the stub ecall in syscall.edl that is no longer needed (per #2713).


mingweishih:
bors try

PR #3191, opened 2020-06-26
Support syscall edl opt-in/out

### LINKED ISSUE #2817: Symbols for some syscall ocalls are linked even though these ocalls are unreachable
In the test for non-builtin ecalls/ocalls introduced in #2816, I discovered that references to many syscall ocalls are linked into the enclave binary even though they are not called by the enclave code and the enclave is linked with `--gc-sections`

Need to investigate why this is and untangle the dependencies.

TODO: detail which ocalls (they appear to be related to consolefs)

mingweishih:
This PR removes the limitation that an enclave always need to import the "syscall.edl". With the PR, the enclave now can selectively import syscall-related ocalls from the corresponding system edl files. For example, to support `printf` inside the enclave, the developer only need to do:
```
from "openenclave/edl/fcntl.edl" import oe_syscall_writev_ocall, oe_syscall_dup_ocall;
```

The PR consists of four parts:
1. Add the default implementations of all the syscall-related ocall wrappers.  These implementations are declared as weak so that the they can be replaced if the developer opt-in any of the ocalls via importing edls. Refer to the [comments](https://github.com/mingweishih/openenclave/blob/503831af4df567e0dcd7ca6626e83c1604e1090f/syscall/hostcalls.c#L15) for more details.
2. Update tests to remove the redundant import of syscall.edl
3. Update the design document

The follow-up of this PR includes:
- Inform the missing imports of edls/ocall via logging
- Update documentation and sample to opt-out unnecessary edls.

fixes #2817

### DIFF
--- a/syscall/hostcalls.c
+++ b/syscall/hostcalls.c
@@ -1,0 +1,1381 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include "syscall_t.h"
+
+/*
+ * This file implements the default implementations of syscall-related
+ * system ocall wrappers. Each implementation is simply an empty function
+ * that returns OE_UNSUPPORTED. If a user does not opt-into these ocalls
+ * (via importing the edls), the linker will pick the default implementions
+ * (which are weak). If the user opts-into any of the ocalls, the linker will
+ * pick the oeedger8r-generated wrapper of the corresponding ocall (which
+ * is strong) instead.
+ *
+ * Note that we need to make the default implementations weak to support
+ * selective ocall import. The reason for this is that if the linker picks
+ * one of the symbols from an object file, it also pulls the rest of the
+ * symbols in the same object file. This behavior causes multiple definition
+ * errors when the user wants to selectively import ocalls if the default
+ * implementations are strong. For example, the user imports one ocall from
+ * epoll.edl. The linker firstly picks the oeedger8r-generated implementation
+ * of the ocall. However, when the linker looks up the default implementations
+ * of the non-imported ocalls in this object file (hostcalls.o), it also pulls
+ * the default implementation of the imported ocall. Because both the default
+ * and the oeedger8r-generated implementations are strong, the linker raises the
+ * error.
+ */
+
+#if !defined(OE_USE_BUILTIN_EDL)
+
+/*
+**==============================================================================
+**
+** epoll.edl
+**
+**==============================================================================
+*/
+
+/**
+ * Declare the prototype of the following functions to avoid the
+ * missing-prototypes warning.
+ */
+oe_result_t _oe_syscall_epoll_create1_ocall(oe_host_fd_t* _retval, int flags);
+oe_result_t _oe_syscall_epoll_wait_ocall(
+    int* _retval,
+    int64_t epfd,
+    struct oe_epoll_event* events,
+    unsigned int maxevents,
+    int timeout);
+oe_result_t _oe_syscall_epoll_wake_ocall(int* _retval);
+oe_result_t _oe_syscall_epoll_ctl_ocall(
+    int* _retval,
+    int64_t epfd,
+    int op,
+    int64_t fd,
+    struct oe_epoll_event* event);
+oe_result_t _oe_syscall_epoll_close_ocall(int* _retval, oe_host_fd_t epfd);
+
+/**
+ * Implement the functions and make them as the weak aliases of
+ * the public ocall wrappers.
+ */
+oe_result_t _oe_syscall_epoll_create1_ocall(oe_host_fd_t* _retval, int flags)
+{
+    OE_UNUSED(_retval);
+    OE_UNUSED(flags);
+    return OE_UNSUPPORTED;
+}
+OE_WEAK_ALIAS(_oe_syscall_epoll_create1_ocall, oe_syscall_epoll_create1_ocall);
+
+oe_result_t _oe_syscall_epoll_wait_ocall(
+    int* _retval,
+    int64_t epfd,
+    struct oe_epoll_event* events,
+    unsigned int maxevents,
+    int timeout)
+{
+    OE_UNUSED(_retval);
+    OE_UNUSED(epfd);
+    OE_UNUSED(events);
+    OE_UNUSED(maxevents);
+    OE_UNUSED(timeout);
+    return OE_UNSUPPORTED;
+}
+OE_WEAK_ALIAS(_oe_syscall_epoll_wait_ocall, oe_syscall_epoll_wait_ocall);
+
+oe_result_t _oe_syscall_epoll_wake_ocall(int* _retval)
+{
+    OE_UNUSED(_retval);
+    return OE_UNSUPPORTED;
+}
+OE_WEAK_ALIAS(_oe_syscall_epoll_wake_ocall, oe_syscall_epoll_wake_ocall);
+
+oe_result_t _oe_syscall_epoll_ctl_ocall(
+    int* _retval,
+    int64_t epfd,
+    int op,
+    int64_t fd,
+    struct oe_epoll_event* event)
+{
+    OE_UNUSED(_retval);
+    OE_UNUSED(epfd);
+    OE_UNUSED(op);
+    OE_UNUSED(fd);
+    OE_UNUSED(event);
+    return OE_UNSUPPORTED;
+}
+OE_WEAK_ALIAS(_oe_syscall_epoll_ctl_ocall, oe_syscall_epoll_ctl_ocall);
+
+oe_result_t _oe_syscall_epoll_close_ocall(int* _retval, oe_host_fd_t epfd)
+{
+    OE_UNUSED(_retval);
+    OE_UNUSED(epfd);
+    return OE_UNSUPPORTED;
+}
+OE_WEAK_ALIAS(_oe_syscall_epoll_close_ocall, oe_syscall_epoll_close_ocall);
+
+/*
+**==============================================================================
+**
+** fcntl.edl
+**
+**==============================================================================
+*/
+
+/**
+ * Declare the prototype of the following functions to avoid the
+ * missing-prototypes warning.
+ */
+oe_result_t _oe_syscall_open_ocall(
+    oe_host_fd_t* _retval,
+    const char* pathname,
+    int flags,
+    oe_mode_t mode);
+oe_result_t _oe_syscall_read_ocall(
+    ssize_t* _retval,
+    oe_host_fd_t fd,
+    void* buf,
+    size_t count);
+oe_result_t _oe_syscall_write_ocall(
+    ssize_t* _retval,
+    oe_host_fd_t fd,
+    const void* buf,
+    size_t count);
+oe_result_t _oe_syscall_readv_ocall(
+    ssize_t* _retval,
+    oe_host_fd_t fd,
+    void* iov_buf,
+    int iovcnt,
+    size_t iov_buf_size);
+oe_result_t _oe_syscall_writev_ocall(
+    ssize_t* _retval,
+    oe_host_fd_t fd,
+    const void* iov_buf,
+    int iovcnt,
+    size_t iov_buf_size);
+oe_result_t _oe_syscall_lseek_ocall(
+    oe_off_t* _retval,
+    oe_host_fd_t fd,
+    oe_off_t offset,
+    int whence);
+oe_result_t _oe_syscall_pread_ocall(
+    ssize_t* _retval,
+    oe_host_fd_t fd,
+    void* buf,
+    size_t count,
+    oe_off_t offset);
+oe_result_t _oe_syscall_pwrite_ocall(
+    ssize_t* _retval,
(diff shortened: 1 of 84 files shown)


mingweishih:
bors try

samples/helloworld/helloworld.edl
```diff
@@ -2,7 +2,10 @@
 // Licensed under the MIT License.
 
 enclave {
-    from "openenclave/edl/syscall.edl" import *;
+    // Import necessary ocalls to support fprintf/printf.
```
dthaler, line 5: How does the developer know which syscalls to import?   Is this in the docs somewhere?

syscall/hostcalls.c
```diff
@@ -0,0 +1,1381 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include "syscall_t.h"
+
+/*
+ * This file implements the default implementations of syscall-related
+ * system ocall wrappers. Each implementation is simply an empty function
+ * that returns OE_UNSUPPORTED. If a user does not opt-in these ocalls
```
dthaler, line 9:
```suggestion
 * that returns OE_UNSUPPORTED. If a user does not opt-into these ocalls
```

syscall/hostcalls.c
```diff
@@ -0,0 +1,1381 @@
...
+ * (via importing the edls), the linker will pick the default implementions
+ * (which are weak). If the user opts-in any of the ocalls, the linker will
```
dthaler, line 11:
```suggestion
 * (which are weak). If the user opts-into any of the ocalls, the linker will
```

syscall/hostcalls.c
```diff
@@ -0,0 +1,1381 @@
...
+/*
+**==============================================================================
+**
+** epoll.edl
+**
+**==============================================================================
+*/
+
+/**
+ * Declare the protoype of the following functions to avoid the
```
dthaler, line 40:
```suggestion
 * Declare the prototype of the following functions to avoid the
```

syscall/hostcalls.c
```diff
@@ -0,0 +1,1381 @@
...
+/*
+**==============================================================================
+**
+** ioctl.edl
+**
+**==============================================================================
+*/
+
+/**
+ * Declare the protoype of the following functions to avoid the
```
dthaler, line 519:
```suggestion
 * Declare the prototype of the following functions to avoid the
```

syscall/hostcalls.c
```diff
@@ -0,0 +1,1381 @@
...
+/*
+**==============================================================================
+**
+** signal.edl
+**
+**==============================================================================
+*/
+
+/**
+ * Declare the protoype of the following functions to avoid the
```
dthaler, line 561:
```suggestion
 * Declare the prototype of the following functions to avoid the
```

syscall/hostcalls.c
```diff
@@ -0,0 +1,1381 @@
...
+/*
+**==============================================================================
+**
+** socket.edl
+**
+**==============================================================================
+*/
+
+/**
+ * Declare the protoype of the following functions to avoid the
```
dthaler, line 588:
```suggestion
 * Declare the prototype of the following functions to avoid the
```

syscall/hostcalls.c
```diff
@@ -0,0 +1,1381 @@
...
+/*
+**==============================================================================
+**
+** poll.edl
+**
+**==============================================================================
+*/
+
+/**
+ * Declare the protoype of the following functions to avoid the
```
dthaler, line 1197:
```suggestion
 * Declare the prototype of the following functions to avoid the
```

syscall/hostcalls.c
```diff
@@ -0,0 +1,1381 @@
...
+/*
+**==============================================================================
+**
+** time.edl
+**
+**==============================================================================
+*/
+
+/**
+ * Declare the protoype of the following functions to avoid the
```
dthaler, line 1233:
```suggestion
 * Declare the prototype of the following functions to avoid the
```

syscall/hostcalls.c
```diff
@@ -0,0 +1,1381 @@
...
+/*
+**==============================================================================
+**
+** unistd.edl
+**
+**==============================================================================
+*/
+
+/**
+ * Declare the protoype of the following functions to avoid the
```
dthaler, line 1266:
```suggestion
 * Declare the prototype of the following functions to avoid the
```

syscall/hostcalls.c
```diff
@@ -0,0 +1,1381 @@
...
+/*
+**==============================================================================
+**
+** utsname.edl
+**
+**==============================================================================
+*/
+
+/**
+ * Declare the protoype of the following functions to avoid the
```
dthaler, line 1364:
```suggestion
 * Declare the prototype of the following functions to avoid the
```

syscall/hostcalls.c
```diff
@@ -0,0 +1,1381 @@
...
+/*
+**==============================================================================
+**
+** fcntl.edl
+**
+**==============================================================================
+*/
+
+/**
+ * Declare the protoype of the following functions to avoid the
```
dthaler, line 127:
```suggestion
 * Declare the prototype of the following functions to avoid the
```

samples/attested_tls/client/tls_client.edl
```diff
@@ -2,8 +2,14 @@
...
 
 enclave {
-    from "openenclave/edl/logging.edl" import *;
-    from "openenclave/edl/syscall.edl" import *;
+    from "openenclave/edl/fcntl.edl" import
+        oe_syscall_read_ocall,
+        oe_syscall_write_ocall,
+        oe_syscall_close_ocall,
+        oe_syscall_writev_ocall,
+        oe_syscall_dup_ocall;
```
dthaler, line 10: Since this is a sample, we need to educate the developer.  How was this list obtained?   These aren't APIs that the sample directly calls, right?  So how does the developer figure out that this is the correct list?   Add a // comment explaining how this list was determined.   For example, do you take the APIs that the code calls and look them up in some doc that gives you things to put on this list?   If so, where is that doc?

tests/debugger/oegdb/enc/contract.c
```diff
@@ -20,7 +20,7 @@ void assert_debugger_binary_contract_enclave_side()
 {
     OE_TEST(TCS_GSBASE_OFFSET == OE_OFFSETOF(sgx_tcs_t, gsbase));
 
-    printf("Debugger contract validated on enclave side.\n");
+    oe_host_printf("Debugger contract validated on enclave side.\n");
```
anakrish, line 23: Why should `oe_host_printf` be used instead of `printf`?

mingweishih:
bors try

tests/debugger/oegdb/enc/contract.c
mingweishih: was trying to minimize the number of imported ocalls (`oe_host_printf` does not need `fcntl.edl`). But it turns out every tests need to import the `fcntl.edl` anyway to support code coverage analysis, I revert the changes.

mingweishih:
> Requesting changes to the samples, since one cannot expect an app developer to know what syscalls are needed for a given API unless there is a discoverable doc to look them up in. I don't see how we can expect the developer to selectively import syscalls as shown in the samples in this change.

I don't think we can document what syscalls are required for every libc/libc++ APIs.  As I mentioned in the description of the PR, I think we could report the lack of certain ocalls via logging. I can added that part in this PR if that can resolve your concern.

dthaler:
> I don't think we can document what syscalls are required for every libc/libc++ APIs. 
> As I mentioned in the description of the PR, I think we could report the lack of certain ocalls via logging. I can add that part in this PR if that resolves your concern.

Expecting developers to do undocumented steps, or to have to turn on logging and follow some other process that is hard, in order to write code is not (in my opinion) reasonable.   We need to lower the friction to write enclave apps, not make their life difficult, or people won't use the SDK.   So either the samples should just import all, or there should be a well documented process they can follow that doesn't involve lots of extra steps.   For example, the helloworld sample being the simplest sample should not involve lots of extra steps in order to author it from scratch.   If there is a well-documented process for those who want to trim down their TCB or something after getting something working, then it's fine if there is *a* sample that shows it and might explain how one got there by using logging.   But I think it should not be the norm in samples to expect jumping through a bunch of extra hoops.

mingweishih:
> > I don't think we can document what syscalls are required for every libc/libc++ APIs.
> > As I mentioned in the description of the PR, I think we could report the lack of certain ocalls via logging. I can add that part in this PR if that resolves your concern.
> 
> Expecting developers to do undocumented steps, or to have to turn on logging and follow some other process that is hard, in order to write code is not (in my opinion) reasonable. We need to lower the friction to write enclave apps, not make their life difficult, or people won't use the SDK. So either the samples should just import all, or there should be a well documented process they can follow that doesn't involve lots of extra steps. For example, the helloworld sample being the simplest sample should not involve lots of extra steps in order to author it from scratch. If there is a well-documented process for those who want to trim down their TCB or something after getting something working, then it's fine if there is _a_ sample that shows it and might explain how one got there by using logging. But I think it should not be the norm in samples to expect jumping through a bunch of extra hoops.

The work of system edl opt-in is essentially opposing to what you describe. We're making no or minimal number of imported edls by default; i.e., the developer starts from minimal TCB.  The opt-in mechanism allows the developer to be aware of when he increases the TCB of the enclave.

anakrish:
@mingweishih 
To help the reviewers, kindly add the link to the Intel SDK docs that describe the workflow for selectively importing functions from system EDLs.

I agree that documenting the system ocalls/ecalls that each C/C++ function needs would not be a worthwhile undertaking, given that the C/C++ library functions themselves do not document what system calls they might end up using.

However, I do think that it would be better to keep the current samples simple by importing all functions instead of selectively importing functions. A new sample could demonstrate the workflow for how the user can minimize TCB.

My comments in an earlier PR were along the same lines that it may not be obvious for an user which edl file to import to fix a particular link error: https://github.com/openenclave/openenclave/pull/2961#issuecomment-636225130

dthaler:
> The work of system edl opt-in is essentially opposing to what you describe. 

Yes I understand it's opposed to making the developer's job easy, which is the problem.   I have no problem with it as an option, but any "getting started" flow needs to be simple or no one will be able to get started. Again I think it's fine if there is a sample that shows it and might explain how one got there by using logging. But I think it should not be the norm in samples to expect jumping through a bunch of extra hoops.   And in my opinion, the logging support and docs on how to use it to figure out the right list, should be a prereq for this PR.

mingweishih:
> > The work of system edl opt-in is essentially opposing to what you describe.
> 
> Yes I understand it's opposed to making the developer's job easy, which is the problem. I have no problem with it as an option, but any "getting started" flow needs to be simple or no one will be able to get started. Again I think it's fine if there is a sample that shows it and might explain how one got there by using logging. But I think it should not be the norm in samples to expect jumping through a bunch of extra hoops. And in my opinion, the logging support and docs on how to use it to figure out the right list, should be a prereq for this PR.

I agree your (also @anakrish's) concerns around samples, which directly impact the user experience, and we need a smoother flow to this update. Given those, I revert the changes to samples and leave them as follow-up work. More specially, we need to discuss
- Whether to modify the existing samples or adding a new sample that demonstrates the selective import of syscall edls
- What's the flow of informing unimported edls to users via logging

samples/attested_tls/client/tls_client.edl
radhikaj: @dthaler , we have this issue to address this; https://github.com/openenclave/openenclave/issues/2719

radhikaj:
> > > The work of system edl opt-in is essentially opposing to what you describe.
> > 
> > 
> > Yes I understand it's opposed to making the developer's job easy, which is the problem. I have no problem with it as an option, but any "getting started" flow needs to be simple or no one will be able to get started. Again I think it's fine if there is a sample that shows it and might explain how one got there by using logging. But I think it should not be the norm in samples to expect jumping through a bunch of extra hoops. And in my opinion, the logging support and docs on how to use it to figure out the right list, should be a prereq for this PR.
> 
> I agree your (also @anakrish's) concerns around samples, which directly impact the user experience, and we need a smoother flow to this update. Given those, I revert the changes to samples and leave them as follow-up work. More specially, we need to discuss
> 
> * Whether to modify the existing samples or adding a new sample that demonstrates the selective import of syscall edls
> * What's the flow of informing unimported edls to users via logging

https://github.com/openenclave/openenclave/issues/2719 is the issue which tracks documentation. One of the existing samples can be modified to opt into a certain ocall. We can have a pointer to the docs and an explanation in that sample. We can put that  in a separate PR after  we discuss the best course of action.

mingweishih:
> LGTM.
> There are no public facing changes; this PR is good to go.
> 
> However, the user experience needs to be thought through when samples are changed.
> In all the test edl files, syscall.edl has been changed to fnctl.edl. It is not clear why fnctl is needed.
> If it enables printing to stdout, then it would be better to have edl names reflect the purpse (printing) rather than implementation detail (fnctl api).

Good suggestion. I'll make another follow-up PR for the better naming of syscall-related edls.

Issue #3181, opened 2020-06-25
Cross-compilation not supported on Linux


ryanhsu19:
This is a follow-up to pull request #3123.

After failing two "Linux nonSGX Verify Quote" instances (both followed by build success on "ACC-1804 Generate Quote" instances), I had a discussion with @mingweishih  and found that

1. In the current version, there is no logic to handle cross-compilation for Linux in `cmake/add_enclave_test.cmake`. As a result, it is supposed that cross-compilation is only done on a Windows instance followed by a Linux instance.

2. Also in the current version, even though we pass `BUILD_ENCLAVES=OFF` into the `cmake` command, `BUILD_ENCLAVES` would still be automatically switched to `ON` (see line 155 in `CMakeLists.txt`). Therefore, `cmake -DBUILD_ENCLAVES=OFF` becomes `cmake -DBUILD_ENCLAVES=ON` when `bors` runs on an "ACC-1804 Generate Quote" instance followed by a "Linux nonSGX Verify Quote" instance.

3. In my proposed version, in order to build the host-verify package, I do not need to build enclaves and would also need to turn off `BUILD_ENCLAVES`; otherwise, it would pack everything it needs to build enclaves, while the host-verify package does not require it. Therefore, I modify the logic so that `BUILD_ENCLAVES` is turned on by default (i.e., there is no `-DBUILD_ENCLAVES` in the `cmake` command) and when necessary. However, the problem arises. In my run,

   - `cmake -DBUILD_ENCLAVES=OFF` meant that it did not build enclaves on the first instance.

   - The other instance could not locate the enclaves generated by the first instance, therefore the build failure.

_Originally posted by @bors in https://github.com/openenclave/openenclave/pull/3123#issuecomment-648980263_

Issue #3160, opened 2020-06-20
Unable to separate oehost target while building host-verify package


ryanhsu19:
With only namespace `openenclave-targets` used to build both OE and host-verify packages, it is impossible to separate `openenclave::oehost` from all the necessary targets 

 - `openenclave::oehostverify`
 - `openenclave::host_verify`
 - `openenclave::oe_includes`

used by host-verify package; therefore,

 1. We are unable to build the host-verify sample without removing anything related to `openenclave::oehost` in the `openenclave-targets*.cmake` files installed by the host-verify package.

 2. We are unable to integrate this into CI/CD, which is related to #2519.

As for the structure of the minimal host-verify package, please refer to my original comment in #3123:

 - https://github.com/openenclave/openenclave/pull/3123#issuecomment-646300983_

radhikaj:
@ryanhsu19 Please add impact of this issue to the description.

ryanhsu19:
@radhikaj Just added the impact of this issue. Hopefully this will clarify the issue.

PR #3193, opened 2020-06-27
Minimize host-verify package without separating oehost

### LINKED ISSUE #2263: Allow building only hostverify library and with minimal build dependencies
OE provides roughly two parts: a) multiple libraries to develop and manage enclaves (target: platform that runs enclaves), and b) the hostverify library to ensure secure communication with remote enclaves (target: any platform, e.g. lightweight clients without enclave support).

Building clients that use the hostverify library currently is unnecessarily complicated. In many cases the provided Ubuntu packages are not enough, either because a different distribution needs to be supported or because there are policies to build all dependencies from source. Building from source is hard because OE's CMake scripts don't allow to build *only* the hostverify library and hence many unneeded build dependencies are required, including Intel's SGX libraries and an up-to-date version of the OCaml compiler.

IMO easily building the hostverify library from source should have first-class support.

ryanhsu19:
This is a workaround based on pull request #3123 and issues #3160 and #3181 so that we do not have to revise any logic or testing flow for now.

This change makes it possible to use `cpack` to create a usable `open-enclave-hostverify` package with only necessary header files and targets, including `openenclave::oehost` ~~and `openenclave::oesign`~~ which ~~are~~ is unable to separate; that is, we can build the `host-verify` sample without any modification of files installed from `open-enclave-hostverify`.

This pull request fixes #2263 partially, since it still includes targets `openenclave::oehost` ~~and `openenclave::oesign`~~.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -351,20 +351,22 @@ add_subdirectory(host)
 # Tests and samples have been ported to manually include system edl files.
 # They can no longer be used when system EDL is builtin.
 if (NOT COMPILE_SYSTEM_EDL)
-  if (BUILD_TESTS)
+  if (BUILD_TESTS AND (NOT COMPONENT MATCHES OEHOSTVERIFY))
     add_subdirectory(tests)
   endif ()
 
   # Skip samples when enabling the code coverage test.
   # Otherwise, all the samples need to explicitly link against the libgcov library.
-  if (BUILD_ENCLAVES AND (NOT CODE_COVERAGE))
+  if (NOT CODE_COVERAGE)
     add_subdirectory(samples)
   endif ()
 endif ()
 
-add_subdirectory(tools)
+if (NOT COMPONENT MATCHES OEHOSTVERFIY)
+  add_subdirectory(tools)
+endif ()
 
-if (BUILD_ENCLAVES)
+if (BUILD_ENCLAVES AND (NOT COMPONENT MATCHES OEHOSTVERIFY))
   add_subdirectory(enclave)
   add_subdirectory(3rdparty)
   add_subdirectory(libc)
@@ -372,7 +374,7 @@ if (BUILD_ENCLAVES)
   add_subdirectory(syscall)
 endif ()
 
-if (OE_SGX)
+if (OE_SGX AND (NOT COMPONENT MATCHES OEHOSTVERIFY))
   add_subdirectory(debugger)
 endif ()
 
--- a/samples/CMakeLists.txt
+++ b/samples/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Copyright (c) Open Enclave SDK contributors.
 # Licensed under the MIT License.
 
-if (HAS_QUOTE_PROVIDER)
+if (BUILD_ENCLAVES AND HAS_QUOTE_PROVIDER)
   install(
     DIRECTORY remote_attestation
     DESTINATION ${CMAKE_INSTALL_DATADIR}/openenclave/samples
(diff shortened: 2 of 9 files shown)


CMakeLists.txt
```diff
@@ -118,13 +118,7 @@ else ()
...
 
 if (OE_SGX)
-  if (WIN32)
-    # Building enclaves on windows is on by default but can be disabled for enclaves pre-compiled under linux
-    option(BUILD_ENCLAVES "Build ELF enclaves" ON)
-  else ()
-    set(BUILD_ENCLAVES ON)
-  endif ()
-
+  set(BUILD_ENCLAVES ON)
```
mingweishih, line 121: On Windows, we still want that `BUILD_ENCLAVES` can be overwritten. Can you revert the changes?

samples/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) Open Enclave SDK contributors.
 # Licensed under the MIT License.
 
-if (HAS_QUOTE_PROVIDER)
+if (BUILD_ENCLAVES AND HAS_QUOTE_PROVIDER)
```
mingweishih, line 4: Is there change in this PR that allows sample to be built when `BUILD_ENCLAVES` is `OFF`?

samples/test-samples.cmake
```diff
@@ -6,7 +6,12 @@
 #
 #     cmake -DHAS_QUOTE_PROVIDER=ON -DSOURCE_DIR=~/openenclave -DBUILD_DIR=~/openenclave/build -DPREFIX_DIR=/opt/openenclave -P ~/openenclave/samples/test-samples.cmake
 
-set(SAMPLES_LIST helloworld file-encryptor switchless host_verify)
+# Set SAMPLES_LIST so that helloworld becomes the first if BUILD_ENCLAVES=ON.
+if (BUILD_ENCLAVES)
```
mingweishih, line 10: Same here, see the question above.

samples/test-samples.cmake
```diff
@@ -21,13 +26,15 @@ else ()
...
   # they can run even if they weren't built against SGX, because in
   # that cause they directly interface with the AESM service.
-  list(APPEND SAMPLES_LIST data-sealing local_attestation)
-
-  # These tests can only run with SGX-FLC, meaning they were built
-  # against SGX.
-  if (HAS_QUOTE_PROVIDER)
-    list(APPEND SAMPLES_LIST remote_attestation)
-    list(APPEND SAMPLES_LIST attested_tls)
+  if (BUILD_ENCLAVES)
```
mingweishih, line 29: Same here, see the question above.

mingweishih:
bors try

samples/CMakeLists.txt
ryanhsu19: Just modified `CMakeLists.txt` in the root folder so that `host_verify` sample can always be built regardless of the value of `BUILD_ENCLAVES`.

ryanhsu19:
Update again to exclude `oesign` from host-verify package.
I use the following command to generate the host-verify package on Ubuntu 18.04 LTS:

 - `cmake -DCOMPONENT=OEHOSTVERIFY -DCMAKE_INSTALL_PREFIX=/opt/openenclave ..`

 - `cpack -G DEB -DCPACK_DEB_COMPONENT_INSTALL=ON -DCPACK_COMPONENTS_ALL=OEHOSTVERIFY`

ryanhsu19:
bors try

tools/oesign/CMakeLists.txt
```diff
@@ -22,25 +22,27 @@ if (NOT HAVE_GETOPT_LONG)
   list(APPEND SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/getopt_long.c)
 endif ()
 
-add_executable(oesign ${SOURCES})
+if (NOT COMPONENT MATCHES OEHOSTVERIFY)
```
mingweishih, line 25: Is this check needed? We already have
```
if (NOT COMPONENT MATCHES OEHOSTVERFIY)
  add_subdirectory(tools)
endif ()
```
in the `CMakeLists.txt`

ryanhsu19: Right, I just tried, but it seems that there are other places where `tools/oesign` can be accessed.

mingweishih:
Please sign off and squash the commits so we can merge. Thanks!

ryanhsu19:
bors try

PR #3194, opened 2020-06-27
Use recursive_mutex in threadcxx test

thomasten:
enc_lock_and_unlock_mutexes_cxx() locks mutexes recursively, so the
mutex objects should be of type recursive_mutex.

Signed-off-by: Thomas Tendyck <tt@edgeless.systems>


anakrish:
bors try

PR #3195, opened 2020-06-27
Fix multi-page EEID data measurement.

wintersteiger:
Large, multi-page EEID data did not get measured correctly. This fixes it and adapts the tests to exercise the fixed behavior. 

### DIFF
--- a/common/sgx/eeid.c
+++ b/common/sgx/eeid.c
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.
 
 #include <stdio.h>
-#include <stdlib.h>
 #include <string.h>
 
 #include <openenclave/bits/attestation.h>
@@ -11,7 +10,6 @@
 #include <openenclave/internal/crypto/sha.h>
 #include <openenclave/internal/eeid.h>
 #include <openenclave/internal/hexdump.h>
-#include <openenclave/internal/malloc.h>
 #include <openenclave/internal/raise.h>
 #include <openenclave/internal/safecrt.h>
 #include <openenclave/internal/types.h>
@@ -266,7 +264,7 @@ done:
     return result;
 }
 
-static oe_result_t _add_page(
+static oe_result_t _measure_page(
     oe_sha256_context_t* hctx,
     uint64_t base,
     void* page,
@@ -311,14 +309,14 @@ oe_result_t oe_remeasure_memory_pages(
     // page.
 
     for (size_t i = 0; i < eeid->size_settings.num_heap_pages; i++)
-        OE_CHECK(_add_page(&hctx, base, &blank_pg, &vaddr, false));
+        OE_CHECK(_measure_page(&hctx, base, &blank_pg, &vaddr, false));
 
     for (size_t i = 0; i < eeid->size_settings.num_tcs; i++)
     {
         vaddr += OE_PAGE_SIZE; /* guard page */
 
         for (size_t i = 0; i < eeid->size_settings.num_stack_pages; i++)
-            OE_CHECK(_add_page(&hctx, base, &stack_pg, &vaddr, true));
+            OE_CHECK(_measure_page(&hctx, base, &stack_pg, &vaddr, true));
 
         vaddr += OE_PAGE_SIZE; /* guard page */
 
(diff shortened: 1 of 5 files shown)


common/sgx/eeid.c
```diff
@@ -356,19 +356,21 @@ oe_result_t oe_remeasure_memory_pages(
...
+        char* eeid_bytes = (char*)eeid;
         size_t num_bytes = oe_eeid_byte_size(eeid);
         size_t num_pages =
             num_bytes / OE_PAGE_SIZE + ((num_bytes % OE_PAGE_SIZE) ? 1 : 0);
 
-        oe_page_t* pages = malloc(sizeof(oe_page_t) * num_pages);
-        memset(pages, 0, sizeof(oe_page_t) * num_pages);
-        memcpy(pages->data, eeid, num_bytes);
-
+        oe_page_t* page = malloc(sizeof(oe_page_t) * num_pages);
```
anakrish, line 364: Even though `num_pages` number of pages are allocated, only the first page is used.

wintersteiger: Yes, I re-use the same bit of memory (it's reset to 0 in the loop). I don't think it's necessary to allocate a new page for each iteration, is it?

wintersteiger: Sorry, misunderstood you - yes, that's far too much memory :-)

wintersteiger: (fixed)

common/sgx/eeid.c
```diff
@@ -356,19 +356,21 @@ oe_result_t oe_remeasure_memory_pages(
...
+        char* eeid_bytes = (char*)eeid;
         size_t num_bytes = oe_eeid_byte_size(eeid);
         size_t num_pages =
             num_bytes / OE_PAGE_SIZE + ((num_bytes % OE_PAGE_SIZE) ? 1 : 0);
 
-        oe_page_t* pages = malloc(sizeof(oe_page_t) * num_pages);
-        memset(pages, 0, sizeof(oe_page_t) * num_pages);
-        memcpy(pages->data, eeid, num_bytes);
-
+        oe_page_t* page = malloc(sizeof(oe_page_t));
```
CodeMonkeyLeet, line 364: This reminds me, for this change to be compatible with upcoming Linux kernel changes, all allocations we `_add_page` into enclave need to be page aligned, see #2566.
```suggestion
        oe_page_t* page = oe_memalign(OE_PAGE_SIZE, sizeof(oe_page_t));
        if (!page)
            OE_RAISE(OE_OUT_OF_MEMORY);
```

common/sgx/eeid.c
```diff
@@ -356,19 +356,21 @@ oe_result_t oe_remeasure_memory_pages(
...
-
-        free(pages);
+        {
+            memset(page->data, 0, sizeof(oe_page_t));
+            size_t n = (i != num_pages - 1) ? OE_PAGE_SIZE
+                                            : (num_bytes % OE_PAGE_SIZE);
+            memcpy(page->data, eeid_bytes + OE_PAGE_SIZE * i, n);
+            OE_CHECK(_add_page(&hctx, base, page->data, &vaddr, true));
+        }
+        free(page);
```
CodeMonkeyLeet, line 373: For symmetry with the use of `oe_memalign` (the special free is so that common code has the appropriate semantics on Windows as well, which requires `_aligned_free` for `_aligned_malloc`)
```suggestion
        oe_memalign_free(page);
```

common/sgx/eeid.c
wintersteiger: This is in the EEID re-measuring of memory pages, it doesn't actually add or remove anything, and it's in `common`. I renamed `_add_page` to `_remeasure_page` to make that clearer. The memory allocation was unnecessary, I replaced it with an `oe_page_t` on the stack.

CodeMonkeyLeet: Yes, the use of `_add_page` matching what we call in create.c was indeed the source of my confusion, thank you for clarifying and renaming!

wintersteiger:
(squashed and signed off)

wintersteiger:
bors try

achamayou:
bors delegate=wintersteiger

wintersteiger:
bors try

wintersteiger:
Just checking: does this fix need another review before it can be merged?

CodeMonkeyLeet:
Yes, we need a second approver for all PRs. @anakrish you've previously looked at this, can you verify that it has addressed your concerns?

@wintersteiger In the meantime, you might want to see if something about this PR drops the code coverage numbers that causes it to fail the check.

host/sgx/create.c
```diff
@@ -685,27 +685,27 @@ static oe_result_t _add_eeid_pages(
...
         for (size_t i = 0; i < num_pages; i++)
         {
+            memset(page->data, 0, sizeof(oe_page_t));
+            size_t n = (i != num_pages - 1) ? OE_PAGE_SIZE
+                                            : (num_bytes % OE_PAGE_SIZE);
+            memcpy(page->data, eeid_bytes + OE_PAGE_SIZE * i, n);
+
             uint64_t addr = enclave_addr + *vaddr;
-            uint64_t src = (uint64_t)(pages + i * OE_PAGE_SIZE);
+            uint64_t src = (uint64_t)(page->data);
```
anakrish, line 702: Not part of this PR, but do eeid pages have to have `SGX_SECINFO_W` flag?

wintersteiger:
@CodeMonkeyLeet I saw the drop in coverage, but ignored it. I don't know how the coverage checker works, but the code changes don't introduce any new branches, while the changes in the tests now test more branches than before. I'm sure I overlooked something, but I can't find a report that would show me what changed in this build.

host/sgx/create.c
wintersteiger: Good observation - I don't think it's needed anymore now that we have the marker page. I'll run some tests to make sure.

wintersteiger: Done, I've added it to this PR too.

wintersteiger:
Interesting! Making the EEID pages read-only raised the coverage by 4%! (I won't complain, but I can't explain it.)

PR #3197, opened 2020-06-29
Remove the import of logging from tls_client.edl

mingweishih:
The PR removes the import of logging.edl in the sample that was missed by #3151.


radhikaj:
Bors r+

anakrish:
One question: Does this mean that logging is now disabled for this sample?

mingweishih:
> One question: Does this mean that logging is now disabled for this sample?

Yes, we opt-out the `logging.edl` in all the samples in #3151.

PR #3198, opened 2020-06-29
Add support of OE report / SGX quote and flat buffer custom claims

shnwc:
Add support of OE reports and raw SGX quotes by `oe_verify_evidence()`. Change custom claims to a flat buffer.


shnwc:
@yentsanglee @qiucwang please review this PR. Thanks!

tests/eeid_plugin/host/host.c
```diff
@@ -112,8 +122,11 @@ void one_enclave_tests(const char* filename, uint32_t flags)
     oe_enclave_t* enclave = NULL;
     oe_result_t result = oe_create_eeid_plugin_enclave(
         filename, OE_ENCLAVE_TYPE_AUTO, flags, &setting, 1, &enclave);
+    // filename, OE_ENCLAVE_TYPE_AUTO, flags, NULL, 0, &enclave);
     OE_TEST(result == OE_OK);
 
+    // OE_TEST_CODE(OE_FAILURE, OE_OK);
```
dthaler, line 128: Delete commented out code
```suggestion
    OE_TEST(result == OE_OK);
```

common/sgx/verifier.c
```diff
@@ -302,98 +327,78 @@ static oe_result_t _fill_with_known_claims(
...
-    // report_data.
-    OE_CHECK(_verify_claims_hash(
-        header->report,
-        header->report_type,
-        evidence + report_size,
-        evidence_size - report_size));
-
-    claims_header = (oe_sgx_plugin_claims_header_t*)(evidence + report_size);
+    // verify the integrity of the custom_claims with hash stored in
+    // report_body, or retrieves the SGX report data
```
dthaler, line 379:
```suggestion
    // report_body, or retrieve the SGX report data
```

common/sgx/verifier.c
```diff
@@ -445,41 +472,168 @@ static oe_result_t _verify_evidence(
...
     oe_sgx_endorsements_t sgx_endorsements;
-    OE_UNUSED(context);
-
-    if (!evidence_buffer || !claims || !claims_length ||
-        evidence_buffer_size < sizeof(*header) ||
-        evidence_buffer_size - sizeof(*header) < header->report_size)
+    sgx_evidence_format_type_t format_type = SGX_FORMAT_TYPE_UNKNOWN;
+    // evidence_buffer can have oe_attestation_header_t and
+    // oe_report_header_t. Only report body is verified.
+    // Then custom claims buffer, if exist, is verified and extracted.
```
dthaler, line 482:
```suggestion
    // Then the custom claims buffer, if it exists, is verified and extracted.
```

tests/attestation_plugin/plugin/tests.c
```diff
@@ -602,34 +709,111 @@ void verify_sgx_evidence(
...
+
+    // Test SGX evidence with tampered custom claims.
+    // Doable only when the custom claims are hashed to SGX report data.
+    if (custom_claims && (format_type == SGX_FORMAT_TYPE_LOCAL ||
+                          format_type == SGX_FORMAT_TYPE_REMOTE))
+    {
+        printf(
+            "====== running verify_sgx_evidence failed with hampered claims\n");
+
+        // Tamper the last byte of the custom claims data.
```
dthaler, line 736:
```suggestion
        // Tamper with the last byte of the custom claims data.
```
(grammar nit)

tests/attestation_plugin/plugin/tests.c
```diff
@@ -602,34 +709,111 @@ void verify_sgx_evidence(
...
+            value = _find_claim(claims, claims_size, OE_CLAIM_SGX_REPORT_DATA);
+
+        OE_TEST(
+            value != NULL && !memcmp(custom_claims, value, custom_claims_size));
+    }
+    OE_TEST_CODE(oe_free_claims(claims, claims_size), OE_OK);
+    claims = NULL;
+    claims_size = 0;
+
+    // Test SGX evidence with tampered custom claims.
```
dthaler, line 728:
```suggestion
    // Test SGX evidence using tampered-with custom claims.
```

tests/attestation_plugin/plugin/tests.c
```diff
@@ -602,34 +709,111 @@ void verify_sgx_evidence(
...
-            endorsements_size,
-            NULL,
-            0,
-            &claims,
-            &claims_size) == OE_QUOTE_HASH_MISMATCH);
 
-    header->data[header->data_size - 1] ^= 1;
+    // Extract legacy OE report and SGX quote from ECDSA evidence
+    // and verify using the conrresponding legacy format IDs.
+    // These two types of evidence are not companied by endorsements.
```
dthaler, line 760:
```suggestion
    // These two types of evidence are not accompanied by endorsements.
```

common/attest_plugin.c
```diff
@@ -192,33 +193,29 @@ oe_result_t oe_verify_evidence(
...
-        (oe_attestation_header_t*)evidence_buffer;
-    oe_attestation_header_t* endorsements =
-        (oe_attestation_header_t*)endorsements_buffer;
-
-    if (!evidence_buffer || evidence_buffer_size < sizeof(*evidence) ||
-        (endorsements_buffer &&
-         endorsements_buffer_size < sizeof(*endorsements)))
+
+    if (!evidence_buffer || !evidence_buffer_size ||
+        (endorsements_buffer && !endorsements_buffer_size))
```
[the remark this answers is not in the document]
shnwc: Thanks!

common/sgx/eeid_verifier.c
```diff
@@ -248,10 +252,17 @@ static oe_result_t _eeid_verify_evidence(
     oe_eeid_t *attester_eeid = NULL, *verifier_eeid = NULL;
     oe_eeid_evidence_t* evidence = NULL;
 
-    if (!evidence_buffer || evidence_buffer_size == 0 ||
+    if (!evidence_buffer ||
+        evidence_buffer_size < sizeof(oe_attestation_header_t) ||
```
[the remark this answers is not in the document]
shnwc: Agreed.

common/sgx/eeid_verifier.c
```diff
@@ -248,10 +252,17 @@ static oe_result_t _eeid_verify_evidence(
...
         (endorsements_buffer && endorsements_buffer_size == 0))
```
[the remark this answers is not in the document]
shnwc: Agreed.

enclave/attest_plugin.c
```diff
@@ -105,57 +121,21 @@ oe_result_t oe_get_evidence(
...
+        endorsements_buffer_size));
 
-    if (endorsements_buffer)
-    {
-        *endorsements_buffer = total_endorsements_buf;
-        *endorsements_buffer_size = total_endorsements_size;
-        total_endorsements_buf = NULL;
-    }
+    // Note: plugins are responsible to wrap evidence and endorsements
+    // with attestation header.
```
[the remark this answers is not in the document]
shnwc: agreed.

include/openenclave/attestation/verifier.h
```diff
@@ -136,6 +136,11 @@ oe_result_t oe_verifier_free_format_settings(uint8_t* settings);
  *
  * @experimental
  *
+ * @param[in] format_id The optional format UUID of the evidence to be verified.
+ * If this parameter is NULL, the evidence_buffer must contain evidence data
+ * with a header that contains a valid format ID. Otherwise, the evidence_buffer
+ * must hold a valid format ID that identifies the type of evidence data held in
```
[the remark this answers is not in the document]
shnwc: Agreed. There is a typo in the comments. "Otherwise, the evidence_buffer ..." should be "Otherwise, this parameter ...". Will update.

include/openenclave/internal/plugin.h
```diff
@@ -23,6 +23,27 @@
 
 OE_EXTERNC_BEGIN
 
+/**
+ * Evidence header: the structure that the OE SDK runtime puts on top of
+ * evidence data.
```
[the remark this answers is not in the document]
shnwc: Thanks!

common/sgx/eeid_verifier.c
```diff
@@ -321,6 +332,22 @@ static oe_result_t _eeid_verify_evidence(
...
+        OE_CHECK(oe_verify_attestation_header(
+            endorsements_buffer, endorsements_buffer_size));
+
+        if (memcmp(
+                &context->base.format_id,
+                &header->format_id,
+                sizeof(oe_uuid_t)))
+            OE_RAISE(OE_CONSTRAINT_FAILED);
+
+        endorsements_buffer += sizeof(*header);
```
yentsanglee, line 348: In attest_plugin.c, endorsements_buffer is not treated as format of         oe_attestation_header_t. Why is it treated as such here?

include/openenclave/internal/sgx/plugin.h
```diff
@@ -5,87 +5,18 @@
...
-    const oe_claim_t* custom_claims,
-    size_t custom_claims_length,
-    uint8_t** claims_out,
-    size_t* claims_size_out,
-    struct _OE_SHA256* hash_out);
+    SGX_FORMAT_TYPE_LOCAL = 1,         // Evidence for SGX local attestation
+    SGX_FORMAT_TYPE_REMOTE = 2,        // Evidence for SGX remote attestation
+    SGX_FORMAT_TYPE_REMOTE_REPORT = 3, // Report for SGX remote attestation
+    SGX_FORMAT_TYPE_REMOTE_QUOTE = 4,  // Raw SGX quote
+    SGX_FORMAT_TYPE_UNKNOWN = 0
```
yentsanglee, line 18: Move enum 0 to the top so the list is numerically sorted.

include/openenclave/internal/sgx/plugin.h
```diff
@@ -5,87 +5,18 @@
...
- * @retval An appropriate error code on failure.
- */
-struct _OE_SHA256;
-oe_result_t oe_sgx_serialize_claims(
-    const oe_claim_t* custom_claims,
-    size_t custom_claims_length,
-    uint8_t** claims_out,
-    size_t* claims_size_out,
-    struct _OE_SHA256* hash_out);
+    SGX_FORMAT_TYPE_LOCAL = 1,         // Evidence for SGX local attestation
```
yentsanglee, line 14: My understanding is:
1 is generated by oe_get_evidence(LOCAL_UUID)
2 is generated by oe_get_evidence(REMOTE_UUID)
3 is generated by oe_get_report(REMOTE)
4 is generated by SGX quote generator

Is that correct?

include/openenclave/internal/plugin.h
```diff
@@ -261,6 +283,9 @@ struct _oe_verifier
      *
      * @param[in] context A pointer to the verifier plugin struct.
      * @param[in] evidence_buffer The evidence buffer.
+     * Note: for evidence data generated by oe_get_evidence(), this buffer
+     * contains all the data received by the public function
+     * oe_verify_evidence().
```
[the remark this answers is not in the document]
shnwc: I think the clarification is needed. Changing to "The evidence data in this buffer contains all headers (if any). The OE SDK framework implementation of oe_verify_evidence() (that calls this entry point) does not perform any unwrapping." Let me know if that's still not clear.

include/openenclave/internal/plugin.h
```diff
@@ -463,6 +488,67 @@ oe_result_t oe_find_verifier_plugin(
...
 
+/**
+ * Fill the supplied header with supplied value and the version number.
+ *
+ * Available only in enclave.
+ *
+ * @param[in] format_id The format ID for the input data.
+ * @param[in] data The input data for which the header is.
+ * @param[in] data_size The size of the input data.
+ * @param[in/out] header The header to be filled with the needed values.
```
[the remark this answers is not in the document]
shnwc: Thanks!

include/openenclave/internal/plugin.h
```diff
@@ -463,6 +488,67 @@ oe_result_t oe_find_verifier_plugin(
...
+ * put the resultant wrapped data in a dynamically allocated buffer.
+ *
+ * Available only in enclave.
+ *
+ * @param[in] format_id The format ID for the input data.
+ * @param[in] data The input data to be wrapped
+ * @param[in] data_size The size of the input data.
+ * @param[out] total_data Pointer a variable that points to the dynamically
+ * allocated buffer that holds the wrapped data.
+ * @param[out[ total_data_size Pointer to a variable that holds the size
```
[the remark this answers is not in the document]
shnwc: Thanks!

tests/eeid_plugin/host/host.c
shnwc: Thanks!

common/sgx/verifier.c
shnwc: Thanks!

common/sgx/verifier.c
shnwc: Thanks!

tests/attestation_plugin/plugin/tests.c
shnwc: Thanks!

tests/attestation_plugin/plugin/tests.c
shnwc: Thanks!

tests/attestation_plugin/plugin/tests.c
shnwc: Thanks!

common/attest_plugin.c
```diff
@@ -192,33 +193,29 @@ oe_result_t oe_verify_evidence(
...
         OE_RAISE(OE_INVALID_PARAMETER);
 
-    plugin_node = oe_attest_find_plugin(verifiers, &evidence->format_id, NULL);
+    if (!format_id)
+    {
+        oe_attestation_header_t* evidence =
```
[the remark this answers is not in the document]
shnwc: Thanks!

common/sgx/eeid_verifier.c
```diff
@@ -321,6 +332,22 @@ static oe_result_t _eeid_verify_evidence(
     /* EEID passed to the verifier */
     if (endorsements_buffer)
     {
+        // Verify and disard attestation header
```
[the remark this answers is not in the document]
shnwc: Thanks!

common/sgx/eeid_verifier.c
shnwc: For EEID, the attestation header was wrapped around the endorsements data in enclave/sgx/eeid_attester.c

include/openenclave/internal/sgx/plugin.h
shnwc: Thanks!

include/openenclave/internal/sgx/plugin.h
shnwc: Yes.

shnwc:
bors try

common/sgx/eeid_verifier.c
dthaler: Just a question: why does it need to be wrapped at all?

include/openenclave/internal/plugin.h
```diff
@@ -121,8 +148,8 @@ struct _oe_attester
         size_t* endorsements_buffer_size);
 
     /**
-     * Creates a legacy SGX report to be used in local or remote attestation.
-     * The report shall contain the data given by the **report_data** parameter.
+     * Creates a legacy OE report to be used in SGX local or remote attestation.
+     * The report shall contain the data given by the report_data parameter.
```
dthaler, line 152: FYI, technically I think the doxygen-recommended markup for report_data in this line is to prefix it with \p
as discussed at https://stackoverflow.com/questions/15398711/whats-the-right-way-to-reference-a-parameter-in-doxygen

include/openenclave/internal/plugin.h
```diff
@@ -463,6 +496,67 @@ oe_result_t oe_find_verifier_plugin(
...
+ *
+ * Available only in enclave.
+ *
+ * @param[in] format_id The format ID for the input data.
+ * @param[in] data The input data for which the header is.
+ * @param[in] data_size The size of the input data.
+ * @param[in,out] header The header to be filled with the needed values.
+ * @retval OE_OK on success.
+ * @retval OE_INVALID_PARAMETER At least one parameter is invalid.
+ * @retval OE_CONSTRAINT_FAILED If the header does not prefix the data
```
dthaler, line 510:
```suggestion
 * @retval OE_CONSTRAINT_FAILED If the header does not prefix the data.
```

include/openenclave/internal/plugin.h
```diff
@@ -463,6 +496,67 @@ oe_result_t oe_find_verifier_plugin(
...
+    oe_attestation_header_t* header);
+
+/**
+ * Wrap supplied evidence / endorsements data with an attestation header,
+ * put the resultant wrapped data in a dynamically allocated buffer.
+ *
+ * Available only in enclave.
+ *
+ * @param[in] format_id The format ID for the input data.
+ * @param[in] data The input data to be wrapped
```
dthaler, line 526:
```suggestion
 * @param[in] data The input data to be wrapped.
```

common/sgx/verifier.c
```diff
@@ -614,14 +783,20 @@ static oe_result_t _get_verifier_plugins(
...
+    uuids[1] = &_ecdsa_report_uuid;
+    uuids[2] = &_ecdsa_quote_uuid;
+
 #ifdef OE_BUILD_ENCLAVE
-    uuid_count = 2; // In enclave, only support local and ECDSA formats
+    uuids[3] = &_local_uuid;
+    uuid_count = 4; // In enclave, only support local and 3 ECDSA formats
 #else
-    uuid_count = 1; // In host, only support ECDSA format
+    uuid_count = 3; // In host, only support 3 ECDSA formats
```
dthaler, line 799: Just a suggestion... it would be nice to have a // comment explaining why local isn't supported in host.

enclave/sgx/attester.c
```diff
@@ -145,88 +67,198 @@ static oe_result_t _get_evidence(
...
+        // Combine the report and custom_claims to get the evidence.
+        // Format is attestation header || report || custom_claims.
+        tmp_buffer_size =
+            sizeof(oe_attestation_header_t) + report_size + custom_claims_size;
+        tmp_buffer = (uint8_t*)oe_malloc(tmp_buffer_size);
+        if (tmp_buffer == NULL)
+            OE_RAISE(OE_OUT_OF_MEMORY);
+
+        // Fill the evidence header
+        OE_CHECK(oe_fill_attestation_header(
```
qiucwang, line 142: So user will have two ways to generate a standard enveloped evidence: `oe_fill_attestation_header()` and `oe_wrap_with_attestation_header()` right?

enclave/sgx/attester.c
```diff
@@ -345,30 +378,49 @@ static oe_result_t _get_attester_plugins(
...
-    // Add one additional entry: the first one for local attestation
-    *attesters =
-        (oe_attester_t*)oe_malloc(sizeof(oe_attester_t) * (uuid_count + 1));
+    // For SGX ECDSA_p256, add support of OE report and SGX quote
+    for (size_t i = 0; i < uuid_count; i++)
+        if (!memcmp(uuid_list + i, &_ecdsa_uuid, sizeof(oe_uuid_t)))
+        {
+            legacy_uuid_count = 2;
+            break;
+        }
```
qiucwang, line 387: Sorry I just got confused here. So if _ecdsa_uuid is included, we add two more slots for OE report and SGX quote plugins. If not, we don't. In which case _ecdsa_uuid might not be included in sgx attester? Can't we just set `legacy_uuid_count = 2`?

include/openenclave/internal/plugin.h
shnwc: Thanks!

"Just a question: why does it need to be wrapped at all?" - good question. But this header is in the existing code, so it is preserved.

The "\p" prefix was not used in existing comments. Is that a style we want to adopt?

include/openenclave/internal/plugin.h
shnwc: Thanks!

include/openenclave/internal/plugin.h
shnwc: Thanks!

common/sgx/verifier.c
shnwc: Thanks!

common/sgx/verifier.c
```diff
@@ -445,41 +472,168 @@ static oe_result_t _verify_evidence(
...
+
+        if (report->version != OE_REPORT_HEADER_VERSION ||
+            report->report_type != OE_REPORT_TYPE_SGX_LOCAL)
+            OE_RAISE(OE_INVALID_PARAMETER);
+
+        format_type = SGX_FORMAT_TYPE_LOCAL;
+    }
+    else if (!memcmp(format_id, &_ecdsa_uuid, sizeof(oe_uuid_t)))
+    {
+        // evidence_buffer has oe_attestation_header_t and oe_report_header_t
```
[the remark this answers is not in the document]
shnwc: Thanks!

common/sgx/verifier.c
```diff
@@ -445,41 +472,168 @@ static oe_result_t _verify_evidence(
...
+
+        if (report->version != OE_REPORT_HEADER_VERSION ||
+            report->report_type != OE_REPORT_TYPE_SGX_REMOTE)
+            OE_RAISE(OE_INVALID_PARAMETER);
+
+        format_type = SGX_FORMAT_TYPE_REMOTE;
+    }
+    else if (!memcmp(format_id, &_ecdsa_report_uuid, sizeof(oe_uuid_t)))
+    {
+        // evidence_buffer has oe_report_header_t
```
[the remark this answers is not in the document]
shnwc: Thanks.

common/sgx/verifier.c
```diff
@@ -445,41 +472,168 @@ static oe_result_t _verify_evidence(
...
+
+        if (report->version != OE_REPORT_HEADER_VERSION ||
+            report->report_type != OE_REPORT_TYPE_SGX_REMOTE)
+            OE_RAISE(OE_INVALID_PARAMETER);
+
+        format_type = SGX_FORMAT_TYPE_REMOTE_REPORT;
+    }
+    else if (!memcmp(format_id, &_ecdsa_quote_uuid, sizeof(oe_uuid_t)))
+    {
+        // evidence_buffer has no header
```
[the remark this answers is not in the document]
shnwc: Thanks!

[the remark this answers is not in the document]
shnwc: Thanks!

common/sgx/verifier.c
```diff
@@ -118,45 +126,56 @@ static oe_result_t _verify_local_report(
...
-    oe_report_type_t report_type,
-    const uint8_t* claims,
-    size_t claims_size)
+// Either verify custom claims, or return SGX report data.
+static oe_result_t _process_claims_hash(
+    const sgx_evidence_format_type_t format_type,
+    const uint8_t* report_body, // Raw SGX quote or report
+    const uint8_t* custom_claims,
+    size_t custom_claims_size,
+    sgx_report_data_t* report_data)
```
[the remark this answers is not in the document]
shnwc: Thanks! Changing to `report_data_out`.

common/sgx/verifier.c
```diff
@@ -118,45 +126,56 @@ static oe_result_t _verify_local_report(
...
-    oe_sha256_context_t hash_ctx = {0};
 
-    OE_CHECK(oe_sha256_init(&hash_ctx));
-    OE_CHECK(oe_sha256_update(&hash_ctx, claims, claims_size));
-    OE_CHECK(oe_sha256_final(&hash_ctx, &hash));
+    if (!report_body)
+        OE_RAISE(OE_INVALID_PARAMETER);
 
-    if (report_type == OE_REPORT_TYPE_SGX_REMOTE)
+    if (format_type == SGX_FORMAT_TYPE_LOCAL ||
```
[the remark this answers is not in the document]
shnwc: Thanks!

common/sgx/verifier.c
```diff
@@ -118,45 +126,56 @@ static oe_result_t _verify_local_report(
...
+        else
+            report_data =
+                (uint8_t*)&((sgx_report_t*)report_body)->body.report_data;
+
+        result = !memcmp(report_data, &hash, OE_SHA256_SIZE)
+                     ? OE_OK
+                     : OE_QUOTE_HASH_MISMATCH;
     }
-    else if (report_type == OE_REPORT_TYPE_SGX_LOCAL)
+    else // SGX_FORMAT_TYPE_REMOTE_REPORT or _QUOTE
```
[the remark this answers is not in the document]
shnwc: Thanks!

common/sgx/verifier.c
```diff
@@ -118,45 +126,56 @@ static oe_result_t _verify_local_report(
 #endif
 }
 
-static oe_result_t _verify_claims_hash(
-    const uint8_t* report,
-    oe_report_type_t report_type,
-    const uint8_t* claims,
-    size_t claims_size)
+// Either verify custom claims, or return SGX report data.
```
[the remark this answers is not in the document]
shnwc: Thanks!

common/sgx/verifier.c
```diff
@@ -118,45 +126,56 @@ static oe_result_t _verify_local_report(
...
+            report_data =
+                (uint8_t*)&((sgx_quote_t*)report_body)->report_body.report_data;
+        else
+            report_data =
+                (uint8_t*)&((sgx_report_t*)report_body)->body.report_data;
+
+        result = !memcmp(report_data, &hash, OE_SHA256_SIZE)
+                     ? OE_OK
+                     : OE_QUOTE_HASH_MISMATCH;
     }
```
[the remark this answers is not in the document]
shnwc: Thanks!

enclave/sgx/attester.c
```diff
@@ -145,88 +67,198 @@ static oe_result_t _get_evidence(
...
+    size_t tmp_buffer_size = 0;
     uint8_t* endorsements = NULL;
     size_t endorsements_size = 0;
-    OE_UNUSED(context);
+    sgx_evidence_format_type_t format_type = SGX_FORMAT_TYPE_UNKNOWN;
+    oe_uuid_t* format_id = NULL;
 
-    if (!evidence_buffer || !evidence_buffer_size ||
+    if (!context || !evidence_buffer || !evidence_buffer_size ||
         (endorsements_buffer && !endorsements_buffer_size))
```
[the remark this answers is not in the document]
shnwc: Thanks!

enclave/sgx/attester.c
shnwc: Yes. `oe_wrap_with_attestation_header()` has the overhead to allocate a new buffer for the header plus the data. `oe_fill_attestation_header()` allows a plugin to avoid this overhead if its existing buffer has reserved space for the header.

enclave/sgx/attester.c
shnwc: If a platform has neither SGX DCAP nor quote-ex libraries installed, it would not have support of _ecdsa_uuid. Adding comments to clarify.

shnwc:
bors try

enclave/sgx/attester.c
qiucwang: Got it. Thanks!

enclave/attest_plugin.c
```diff
@@ -39,7 +39,32 @@ oe_result_t oe_unregister_attester_plugin(oe_attester_t* plugin)
...
+    size_t data_size,
+    oe_attestation_header_t* header)
+{
+    oe_result_t result = OE_UNEXPECTED;
+
+    if (!format_id || !data || !data_size || !header)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+    // The header must prefix the data
+    if ((uint8_t*)header + sizeof(*header) != data)
```
CodeMonkeyLeet, line 54: It seems like this method only exists to enforce this check, which requires that you've already laid out the buffer and calculated the offset to the data pointer, while reserving `sizeof(oe_attestation_header_t)` bytes at the start of it. Another way to look at it is that the check is a tautology: by definition of `oe_attestation_header_t` anything that comes after it is the `data`, why make someone pass you the pointer for it? 

Where it's currently being used outside of `oe_wrap_with_attestation_header` as in sgx/attester.c, it doesn't seem to do anything other than create a dev time failure path. It's probably easier to cast the buffer to `oe_attestation_header_t*` and just writing the `version`, `format_id` and `data_size` fields directly.

If the goal is to initialize constant fields to known good values (e.g. `magic` or `version`) consider defining a simple inline instead:
```c
inline void oe_set_attestation_header(
    uint8_t* buffer,
    size_t buffer_size,
    const oe_uuid_t* format_id,
    const size_t data_size);
```
That way you can check the minimum required size of `buffer_size` exceeds `data_size + sizeof(oe_attestation_header_t)` at least.

dthaler:
> * OE SDK does not provide a standard wrapper format for encapsulating the UUID.
>   
>   * The verifier must know the evidence format _a priori_ and pass the `format_id` to `oe_verify_evidence`. There is no support for accepting `format_id == NULL`.
>   * This is Dave's preferred approach.

Just a comment... Simon is correct this is my preferred approach, but it doesn't necessarily mean there's no support for null format_id.  Null format_id can be supported, for example, by having the core SDK recognized null and map it to some internal UUID that goes to the "plugin" (or whatever you want to call it) that handles the wrapper format.   In that sense, null is just a convenient alias for the internal UUID.

common/sgx/eeid_verifier.c
yentsanglee: eeid_attester.c wraps endorsements in its own structure oe_eeid_t. The original code seems fine.

shnwc:
Thank you @CodeMonkeyLeet for the insightful comments! 

> I do have one objection to taking this change, which is the requirement that both the verifier plugin as well as the OE SDK have to understand the same attestation header format for this to be useful, and the attestation header format is still internal to OE while the ability to pass a null `format_id` is supported on the public `oe_verify_evidence` API. I don't think that inconsistency is maintainable in the long run, and per the previous SIG-attestation meeting, the only consistent options I see here are:
> * (Option 1) OE SDK does not provide a standard wrapper format for encapsulating the UUID.
>    * …
> * (Option 2) OE SDK provides a standard wrapper format for encapsulating the UUID.
>    * …

Agree that the inconsistency is hard to maintain in the long run. The existing design (where an attestation header is always used and `oe_verify_evidence()` does not take in a separate `format_id` parameter) is more elegant.

The original goal was to meet Larry’s request to use the evidence verification code to verify SGX ECDSA quotes (either raw or in OE report format) and output the claims, and to avoid the need to copy the evidence data just to prefix it with a header. In the process, we somehow generalized the request, eventually to allow `oe_verify_evidence()` to take in optional `format_id` for all formats. This generalization reduced the elegancy of the overall design.

If Larry’s request is scoped to only support legacy SGX ECDSA quotes (or plus a limited number of other TEEs like VBS), then I see that the option 3 below can be more practical than options 1 and 2.

(Option 3) Add a new public API function to verify legacy evidence formats:

* Keep existing definition of existing API functions `oe_get_evidence()` and `oe_verify_evidence()`
  * Except that custom claims are held in an opaque flat buffer.
* Add new public API function `oe_verify_legacy_evidence([in] legacy_format_type, …)` to verify legacy evidence formats
  * Parameter `legacy_format_type` is not a UUID, but can be an enum type identifying SGX quote, OE report, or some other limited number of legacy formats
  * Note: this is different than `oe_verify_evidence_by_id()` that was proposed, that takes in a format UUID.
  * There is no input endorsements buffer
  * Otherwise the other parameters are that same as in `oe_verify_evidence()`
* Verifier plugin entry point `verify_evidence()` takes in an additional parameter `legacy_format_type`, indicating whether the input evidence is from `oe_verify_evidence()`, or one of the legacy formats.
  * It’s up to a plugin to decide which (if any) legacy formats to support.
* OE SDK framework implementation of `oe_verify_legacy_evidence()` is NOT TEE-agnostic. It needs to check the `legacy_format_type`, map it to a format UUID, find the corresponding plugin, and invoke the plugin’s `verify_evidence()` entry point.

shnwc:
@dthaler @yentsanglee @CodeMonkeyLeet Let me know your thoughts on the above option 3. If you guys are ok, I can update the design and implementation.

CodeMonkeyLeet:
@shnwc:
> The original goal was to meet Larry’s request to use the evidence verification code to verify SGX ECDSA quotes (either raw or in OE report format) and output the claims, and to avoid the need to copy the evidence data just to prefix it with a header. In the process, we somehow generalized the request, eventually to allow oe_verify_evidence() to take in optional format_id for all formats. This generalization reduced the elegancy of the overall design.

I'm not sure I follow ... my understanding is that option 1 by itself already directly addresses Larry's requirements, and option 2 layers on top of option 1 a convenience feature for OE app authors to pack the UUID with the evidence, to maintain parity with an existing OE SDK behavior. It's unclear to me what additional use case the proposed option 3 is intended to address, or how it relates to the existing PR.

It might be helpful to walk us through the scenario you have in mind using each of the options to show what your concerns are with the existing options, and how the new one addresses them.

dthaler:
> > The original goal was to meet Larry’s request to use the evidence verification code to verify SGX ECDSA quotes (either raw or in OE report format) and output the claims, and to avoid the need to copy the evidence data just to prefix it with a header. In the process, we somehow generalized the request, eventually to allow oe_verify_evidence() to take in optional format_id for all formats. This generalization reduced the elegancy of the overall design.
> 
> I'm not sure I follow ... my understanding is that option 1 by itself already directly addresses Larry's requirements, and option 2 layers on top of option 1 a convenience feature for OE app authors to pack the UUID with the evidence, to maintain parity with an existing OE SDK behavior. It's unclear to me what additional use case the proposed option 3 is intended to address, or how it relates to the existing PR.

Right.  The generalization provides an elegancy that works for standards and other non-SGX mechanisms.  The original design was not elegant nor interoperable in that respect.   The fact is that what you consider "legacy" for SGX is actually closer (from the core's perspective, not the plugin's) to what is needed for standards and interoperability.   So I do not like the option 3 as I think it won't work for anything but SGX.

shnwc:
Comparison among the three options, as well as the existing design:

* Option 1:
  * As compared to the existing design, the evidence generated by `oe_get_evidence()` is never self-contained. The `format_id` for the evidence always need to be communicated to verifiers separately (not as part of the evidence). `oe_verify_evidence()` always requires a non-NULL `format_id` parameter.
* Option 2:
  * As compared to the existing design, on the attester side, `oe_get_evidence()` takes an additional `bool` parameter `wrap_with_header`, to indicate whether the evidence (and endorsements buffer) will be wrapped with an attestation header to make them self-contained.
  * As compared to the existing design, on the verifier side, `oe_verify_evidence()` takes an additional parameter `format_id`, to specify evidence format and to indicate if the evidence is self-contained or not.
    * Evidence (and endorsements buffer) is self-contained if and only if `format_id` is NULL.
    *  If `format_id` is not NULL, the evidence buffer (and endorsements buffer) must not be self-contained.
  * As compared to the existing design, the attester and verifier application software not only need to negotiate a format ID, they also need to negotiate whether the evidence should be self-contained or not. The chosen format ID and wrapping flag will need to be communicated from the attester to the verifier separate from the evidence itself.
    * As a result, I have reservation about the value of wrapping evidence and endorsements.
  * Overall, option 2 is a layer on top of option, as commented by @CodeMonkeyLeet, so it's safe to implement option 2.
* Option 3:
  * This option tries to keep all the features of the existing design, without the additional complexities discussed above for option 1 and option 2.
  * But in order to keep all the existing features, legacy SGX quotes and reports are treated as legacy, and handled as such. As a result, `oe_verify_legacy_evidence()` won’t work easily for anything but SGX.

Based on all the discussion so far, I’m ok with option 2, and will update the design and implementation. Let me know if you think differently.

tests/eeid_plugin/enc/enc.c
```diff
@@ -9,6 +9,7 @@
 #include <openenclave/enclave.h>
 #include <openenclave/internal/report.h>
 #include <openenclave/internal/tests.h>
+#include <openenclave/internal/trace.h>
```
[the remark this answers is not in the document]
shnwc: Will delete. It was for the macros in it at some point during debugging.

include/openenclave/attestation/sgx/evidence.h
```diff
@@ -0,0 +1,78 @@
...
+#define _OE_ATTESTATION_SGX_EVIDENCE_H
+
+#include <openenclave/bits/result.h>
+
+OE_EXTERNC_BEGIN
+
+/**
+ * The SGX plugin UUID.
+ */
+#define OE_SGX_PLUGIN_UUID                                                \
```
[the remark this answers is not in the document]
shnwc: Will delete. `OE_SGX_PLUGIN_UUID` is a legacy macro that was used in multiple code files. But those files have been changed.

common/attest_plugin.c
```diff
@@ -192,33 +193,34 @@ oe_result_t oe_verify_evidence(
...
+
+    if (!evidence_buffer || !evidence_buffer_size ||
+        (!endorsements_buffer && endorsements_buffer_size) ||
+        (endorsements_buffer && !endorsements_buffer_size))
         OE_RAISE(OE_INVALID_PARAMETER);
 
-    plugin_node = oe_attest_find_plugin(verifiers, &evidence->format_id, NULL);
+    if (!format_id)
+    {
+        oe_attestation_header_t* evidence =
```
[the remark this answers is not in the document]
shnwc: With the latest design, this function will be removed.

common/attest_plugin.c
```diff
@@ -192,33 +193,34 @@ oe_result_t oe_verify_evidence(
...
-        OE_RAISE(OE_CONSTRAINT_FAILED);
-
     verifier = (oe_verifier_t*)plugin_node->plugin;
     OE_CHECK(verifier->verify_evidence(
         verifier,
-        evidence->data,
-        evidence->data_size,
-        endorsements ? endorsements->data : NULL,
-        endorsements ? endorsements->data_size : 0,
+        evidence_buffer,
```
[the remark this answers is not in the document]
shnwc: Latest design will fix this issue.

common/attest_plugin.c
```diff
@@ -403,3 +405,20 @@ oe_result_t oe_find_verifier_plugin(
...
+
+oe_result_t oe_verify_attestation_header(
+    const uint8_t* data_buffer,
+    size_t data_buffer_size)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    oe_attestation_header_t* header = (oe_attestation_header_t*)data_buffer;
+
+    if (!data_buffer || (data_buffer_size < sizeof(oe_attestation_header_t)) ||
+        header->version != OE_ATTESTATION_HEADER_VERSION)
```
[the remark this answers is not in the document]
shnwc: Latest design will fix this issue, when an attestation header is expected, it's format ID field will be check.

include/openenclave/internal/plugin.h
```diff
@@ -23,6 +23,27 @@
 
 OE_EXTERNC_BEGIN
 
+/**
+ * Evidence header: the structure that the OE SDK runtime puts on top of
+ * evidence data, for format IDs that require it.
+ */
+typedef struct _oe_attestation_header
```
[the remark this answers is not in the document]
shnwc: Good suggestion on the names. There is no need to bridge `oe_get_report(LOCAL)` to the new evidence API.

With the latest design, plugins will not see attestation headers.

enclave/attest_plugin.c
shnwc: Thanks! With the latest design, existing code will be used to handle the attestation headers.

shnwc:
bors try

shnwc:
@CodeMonkeyLeet the latest commit implemented option 2, and made the changes you requested. Please review. Thanks!

yentsanglee:
+1 for option 2 - adding "wrap_with_header" flag seems to be the cleanest way.

common/sgx/verifier.c
```diff
@@ -302,98 +339,78 @@ static oe_result_t _fill_with_known_claims(
...
-            entry->name_size,
-            entry->name + entry->name_size,
-            entry->value_size));
-
-        // Go to next entry.
-        claims_buf += size;
-        claims_buf_size -= size;
-    }
+    // Default hash for empty string, as described in the literature.
+    static uint8_t sha256_for_empty_string[] = {
```
CodeMonkeyLeet, line 349: Should declare this as `const` as well.

common/sgx/verifier.c
```diff
@@ -302,98 +339,78 @@ static oe_result_t _fill_with_known_claims(
...
+    OE_CHECK(_process_sgx_report_data(
+        format_type,
+        report_body,
+        custom_claims,
+        custom_claims_size,
+        &report_data));
+
+    // There is no "custom_claims" claim if custom_claims
+    // is expected but is empty. Otherwise there is either a "custom_claims"
+    // claim or an SGX report data claim
```
CodeMonkeyLeet, line 401: Consider making the text less confusing:
```suggestion
    // If provided, custom_claims can only be the OE_CLAIM_CUSTOM_CLAIMS
    // or OE_CLAIM_SGX_REPORT_DATA types. 
```

common/sgx/verifier.c
```diff
@@ -302,98 +339,78 @@ static oe_result_t _fill_with_known_claims(
...
     oe_claim_t* claims = NULL;
     uint64_t claims_length = 0;
     uint64_t claims_size = 0;
     size_t claims_added = 0;
+    size_t additional_claim = 0;
+    sgx_report_data_t report_data;
 
-    // Check if the buffer is the proper size.
-    if (evidence_size - report_size < sizeof(*claims_header))
+    if (!format_id || !report_body || !report_body_size)
```
CodeMonkeyLeet, line 387: Consider adding the buffer and size consistency check that was applied to `endorsements` elsewhere: `(!custom_claims == !custom_claims_size)`

enclave/attest_plugin.c
```diff
@@ -69,8 +69,9 @@ static oe_result_t _wrap_with_header(
 
 oe_result_t oe_get_evidence(
     const oe_uuid_t* format_id,
-    const oe_claim_t* custom_claims,
-    size_t custom_claims_length,
+    bool wrap_with_header,
```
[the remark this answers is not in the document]
yentsanglee: > @dthaler @yentsanglee From an API design perspective, should this be a bit flag array instead to accommodate future extensibility? Essentially, it's one of the few extensibility points for the OE SDK itself in the attestation plugin infrastructure.

I agree a 4-byte flag is more extensible.

common/attest_plugin.c
```diff
@@ -192,33 +193,67 @@ oe_result_t oe_verify_evidence(
...
+        oe_attestation_header_t* evidence =
+            (oe_attestation_header_t*)evidence_buffer;
+
+        if (evidence_buffer_size < sizeof(oe_attestation_header_t) ||
+            evidence->version != OE_ATTESTATION_HEADER_VERSION)
+            OE_RAISE(OE_INVALID_PARAMETER);
+
+        if (endorsements_buffer)
+        {
+            oe_attestation_header_t* endorsements =
```
yentsanglee, line 217: It is unnecessary for the endorsements be wrapped around oe_attestation_header_t.

enclave/attest_plugin.c
```diff
@@ -105,30 +106,46 @@ oe_result_t oe_get_evidence(
...
-            plugin_endorsements,
-            plugin_endorsements_size,
-            &total_endorsements_buf,
-            &total_endorsements_size));
+            plugin_evidence,
+            plugin_evidence_size,
+            &total_evidence_buf,
+            &total_evidence_size));
+
+        if (endorsements_buffer)
```
yentsanglee, line 127: It is unnecessary for the endorsements be wrapped around oe_attestation_header_t. 

oe_get_sgx_endorsements() and oe_parse_sgx_endorsements() both deal with raw endorsements without wrapping.

enclave/attest_plugin.c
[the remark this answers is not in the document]
shnwc: I'm ok either way. In the current design, there OE SDK itself does not do much. It basically just chooses a plugin for the format ID, and invokes plugin entry points.

I don't even see much value in the attestation header wrapped by the OE SDK. Since it is optional, the verifier can't inspect the evidence to tell its format or if a header is present. it must learn from the attester via some other means. The header used to be valuable when it was mandatory.

So I'm also OK if `oe_get_evidence()` is changed to never wrap an attestation header. With this change, there would be no need of the `wrap_with_header` parameter, and `oe_verify_evidence()` would always need a valid `format_id`.

@dthaler and @yentsanglee what are your recommendations?

include/openenclave/internal/plugin.h
```diff
@@ -94,25 +94,31 @@ struct _oe_attester
...
+     * @param[in] custom_claims The optional custom claims buffer.
+     * @param[in] custom_claims_size The number of bytes in the custom claims
+     * buffer.
      * @param[in] opt_params The optional plugin-specific input parameters.
      * @param[in] opt_params_size The size of opt_params in bytes.
      * @param[out] evidence_buffer An output pointer that will be assigned the
-     * address of the evidence buffer.
+     * address of the evidence buffer. The output evidence data contains all
+     * the headers (if any) needed. The implementation of oe_get_evidence()
+     * (that calls this entry point) does not perform further wrapping.
```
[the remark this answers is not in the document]
shnwc: @CodeMonkeyLeet thank you very much for catching this!

include/openenclave/internal/plugin.h
```diff
@@ -94,25 +94,31 @@ struct _oe_attester
...
      * @param[out] evidence_buffer_size A pointer that points to the size of the
      * evidence buffer in bytes.
      * @param[out] endorsements_buffer An output pointer that will be assigned
-     * the address of the endorsements buffer.
+     * the address of the endorsements buffer. The output endorsements data
```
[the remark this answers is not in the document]
shnwc: Thanks!

include/openenclave/internal/plugin.h
```diff
@@ -260,9 +266,15 @@ struct _oe_verifier
      * @experimental
      *
      * @param[in] context A pointer to the verifier plugin struct.
-     * @param[in] evidence_buffer The evidence buffer.
+     * @param[in] evidence_buffer The evidence buffer. The evidence data
```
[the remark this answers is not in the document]
shnwc: Thanks! these comments should have been removed, with the latest change.

common/sgx/verifier.c
shnwc: Agreed.

common/sgx/verifier.c
shnwc: Thank you!

common/sgx/verifier.c
shnwc: Thanks!

common/attest_plugin.c
shnwc: This is an existing behavior, and there is existing ctest code to specifically verify that the headers in the evidence and endorsements have matching format IDs.

The current proposal does not try to change this existing behavior.

enclave/attest_plugin.c
shnwc: Please see above comment.

shnwc:
bors try

shnwc:
@CodeMonkeyLeet the latest commit incorporated your comments. With the implementation of option 2, it also should have incorporated the changes you requested. Please confirm those requested changes, so that merging is not blocked on that.

common/attest_plugin.h
```diff
@@ -15,9 +18,11 @@
 
 #include <openenclave/internal/plugin.h>
 
+OE_EXTERNC_BEGIN
+
 /**
  * Evidence header: the structure that the OE SDK runtime puts on top of
- * evidence data.
+ * evidence data, for format IDs that require it.
```
CodeMonkeyLeet, line 25: This isn't associated with format IDs any more, consider:
```suggestion
 * evidence data when oe_get_evidence is asked to include the format ID
 * with the evidence.
```

common/sgx/verifier.c
CodeMonkeyLeet: That's an interesting choice, why do we allow that? There's no functional value in doing so, and providing a non-zero size to a null pointer to an array should still be trapped.

include/openenclave/internal/plugin.h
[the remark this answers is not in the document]
CodeMonkeyLeet: All the other instances were fixed but it looks like this still remains in the latest iteration.

common/attest_plugin.c
yentsanglee: Opened issue [3241](https://github.com/openenclave/openenclave/issues/3241) for further investigation.

enclave/attest_plugin.c
yentsanglee: Opened issue 3241 to track the issue.

common/attest_plugin.h
shnwc: Thanks!

enclave/attest_plugin.c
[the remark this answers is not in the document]
CodeMonkeyLeet: > I don't even see much value in the attestation header wrapped by the OE SDK. Since it is optional, the verifier can't inspect the evidence to tell its format or if a header is present. it must learn from the attester via some other means. The header used to be valuable when it was mandatory.

The value proposition of OE SDK offering a format-ID embedded wrapper hasn't changed between the iterations of this PR; even in the existing behavior of mandatory headers, the implicit contract between the attester and verifier was that both had to be using the OE SDK. The design choice to expose the wrapping option simply makes that contract more explicit for the parties that wish to continue to use that behavior, because OE SDK is expanding its interop with non-OE SDK attesters and verifiers.

common/sgx/verifier.c
shnwc: Since custom claims are attached at the end of the SGX quote, some callers can just get a pointer to right after the quote for the custom claims and calculate the size of the custom claims (which could be zero). In this case the pointer is not NULL, but the size is zero.

Yes, let me add the trap to null pointer with non-zero size.

shnwc:
> All the other instances were fixed but it looks like this still remains in the latest iteration.

Hmm interesting, I could not find this text in my copy. Wonder if you need to reload the page to get the latest version? I see "Outdated" right above your comment.

enclave/attest_plugin.c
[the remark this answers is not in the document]
shnwc: @CodeMonkeyLeet Thanks! Agreed that the value proposition stays as you described.

shnwc:
bors try

common/sgx/verifier.c
CodeMonkeyLeet: Thanks for the explanation, that would be useful to add as a comment here, since new contributors or maintainers may block this behavior in the future without understanding the context.

include/openenclave/internal/plugin.h
```diff
@@ -104,15 +105,17 @@ struct _oe_attester
      * evidence buffer in bytes.
      * @param[out] endorsements_buffer An output pointer that will be assigned
      * the address of the endorsements buffer.
+     * oe_get_evidence() (that calls this entry point) does not perform further
```
CodeMonkeyLeet, line 108: @shnwc This is the remaining outdated comment that still needs to be removed.

CodeMonkeyLeet:
@yentsanglee @dthaler @shnwc I'd like closure on the `oe_get_evidence` API question before I approve, although it's really a design discussion appropriate for #3233.

include/openenclave/internal/plugin.h
shnwc: Thank you!

shnwc:
> @yentsanglee @dthaler @shnwc I'd like closure on the `oe_get_evidence` API question before I approve, although it's really a design discussion appropriate for #3233.

Personally I'm ok either way: either keep the current definition and extend later when there is a need, or change it in this PR. The API is still experimental, and we will have some further clean up to make it production.

shnwc:
bors try

CodeMonkeyLeet:
@yentsanglee Is this set of changes intended for the v0.10 release? If not, I can approve and iterate on the API question in the design PR.

shnwc:
> @yentsanglee Is this set of changes intended for the v0.10 release? If not, I can approve and iterate on the API question in the design PR.

The main objective of this PR is to support the legacy OE report and raw quote formats for SGX ECDSA evidence that Larry requested. Not sure what Larry's expectation is, in terms of timeline and whether he needs it in v0.10. In any case, the API is still experimental, and will take a few more iterations to finalize.

shnwc:
> > @yentsanglee @dthaler @shnwc I'd like closure on the `oe_get_evidence` API question before I approve, although it's really a design discussion appropriate for #3233.
> 
> Personally I'm ok either way: either keep the current definition and extend later when there is a need, or change it in this PR. The API is still experimental, and we will have some further clean up to make it production.

Just had a discussion with Yen, and will make another commit and add the bit-wise flags to `oe_get_evidence()` shortly.

yentsanglee:
bit-wise flags sounds good. Let's try to get this merged once that's done.

shnwc:
bors try

shnwc:
The latest commit has the function `oe_get_evidence()` taking a bit-wise flags parameter, with one bit indicating if evidence will be wrapped with an attestation header.

include/openenclave/attestation/attester.h
```diff
@@ -73,17 +78,23 @@ oe_result_t oe_attester_select_format(
...
  *
- * @param[in] evidence_format The format UUID of the evidence to be generated.
- * @param[in] custom_claims The optional custom claims list.
- * @param[in] custom_claims_length The number of custom claims.
+ * @param[in] format_id The format ID of the evidence to be generated.
+ * @param[in] flags A bit-wise parameter. Currently there is one bit
+ * defined: OE_EVIDENCE_FLAGS_BIT_WRAP_HEADER. If this bit is set,
+ * the evidence and
+ * endorsements will be wrapped with an attestation header containing the
+ * format ID.
```
CodeMonkeyLeet, line 86: Fix up the line spacing.
```suggestion
 * defined: OE_EVIDENCE_FLAGS_BIT_WRAP_HEADER. If this bit is set,
 * the evidence and endorsements will be wrapped with a header
 * containing the format ID.
```

include/openenclave/attestation/attester.h
```diff
@@ -24,6 +24,11 @@
 
 OE_EXTERNC_BEGIN
 
+/**
+ * Bit-wise flags passed to oe_get_evidence() function.
+ */
+#define OE_EVIDENCE_FLAGS_BIT_WRAP_HEADER 0x00000001
```
CodeMonkeyLeet, line 30: Naming consistency with other flags: e.g. `OE_SGX_FLAGS_DEBUG`, so we should drop the extra `_BIT_`. 

I would also suggest choosing names that describe intent rather than implementation, so `OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID` rather than `WRAP_HEADER`; the same flag may be implemented differently in the future, and we may add more flags to indicate other information we might want to include in the existing header.

include/openenclave/attestation/attester.h
shnwc: Thanks!

include/openenclave/attestation/attester.h
shnwc: Agreed!

shnwc:
bors try

shnwc:
The latest commit incorporated all the comments, and is ready for merge.

yentsanglee:
bors retry

Issue #2958, opened 2020-04-28
Bug in edger8r import code


jhand2:
Let's say my enclave application has an edl file called `swithchless.edl` which imports some system edl

```
// switchless.edl
enclave
{
   from "sgx/switchless.edl" import *;

   trusted
   {
      int sample_ecall();
   };

   untrusted
   {
      int sample_ocall();
   }
};
```

Because the system EDL file `sgx/switchless.edl` has the same name as the enclave's EDL file, the code generator (`tools/oeedger8r/intel/CodeGen.ml`) will not import the system EDL file, thinking it has already seen it even though they're different files.

radhikaj:
@anakrish  will investigate and possibly submit a PR upstream to Intel SGX SDK

llly:
Importing EDL file with same name from different folder is supported by the code generator.
should be careful with  `--search-path` option in this scenario. 

I think the failure of sample `switchless.edl` is caused by syntax error.
`public` key word and semicolon after untrusted section are missed. Should be
```
// switchless.edl
enclave
{
   from "sgx/switchless.edl" import *;

   trusted
   {
      public int sample_ecall();
   };

   untrusted
   {
      int sample_ocall();
   };
};
```

jhand2:
I just wrote that free-hand as an example. In actual usage edger8r doesn't fail, it just silently only includes one of the files.

When checking for duplicates the parser strips out the path, so the path you import from doesn't seem to be relevant.

llly:
I cannot reproduce your failure. 
I have a sample EDL `sgx_tswitchless.edl`
```
enclave
{
   from "include/sgx_tswitchless.edl" import *;

   trusted
   {
     public int sample_ecall();
   };

   untrusted
   {
      int sample_ocall();
   };
};
```
 SGXSDK is installed in ~ and `~/sgxsdk/include/sgx_tswitchless.edl` is the imported EDL.
Run `oeedger8r sgx_tswitchless.edl --search-path ~/sgxsdk/`. 
Checking sgx_tswitchless_t.h I can see
```
 /**** ECALL prototypes. ****/
int sample_ecall(void);

sgx_status_t sl_init_switchless(void* sl_data);

sgx_status_t sl_run_switchless_tworker(void);

 /**** OCALL prototypes. ****/
oe_result_t sample_ocall(int* _retval);
```
Then I copy `~/sgxsdk/include/sgx_tswitchless.edl`  to `include/sgx_tswitchless.edl`.
[...]

PR #3200, opened 2020-06-30
Adopt oeedger8r preprocessor in tests edl to differentiate platforms

mingweishih:
This PR does the following things.
- Adopt the preprocessor feature of oeedger8r to differentiate the platform (sgx or optee).
- Rename the `sgx_attestation.edl` to `attestation` since there is no longer a limitation that two imported edls cannot have the same same (per #2958).

### DIFF
--- a/tests/cmake_name_conflict/name_conflict.edl
+++ b/tests/cmake_name_conflict/name_conflict.edl
@@ -4,7 +4,11 @@
 enclave {
     from "openenclave/edl/logging.edl" import oe_write_ocall;
     from "openenclave/edl/fcntl.edl" import *;
-    from "platform.edl" import *;
+#ifdef OE_SGX
+    from "openenclave/edl/sgx/platform.edl" import *;
+#else
+    from "openenclave/edl/optee/platform.edl" import *;
+#endif
 
     trusted {
         public void test_name_conflict();
--- a/docs/DesignDocs/system_ocall_opt_in.md
+++ b/docs/DesignDocs/system_ocall_opt_in.md
@@ -145,7 +145,7 @@ in one of the following EDL files
 * `edl/utsname.edl`
 * `edl/sgx/cpu.edl`
 * `edl/sgx/debug.edl`
-* `edl/sgx/sgx_attestation.edl`
+* `edl/sgx/attestation.edl`
 * `edl/sgx/switchless.edl`
 * `edl/sgx/thread.edl`
 
--- a/enclave/core/CMakeLists.txt
+++ b/enclave/core/CMakeLists.txt
@@ -3,9 +3,6 @@
 
 set(OE_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/include)
 set(EDL_DIR ${CMAKE_SOURCE_DIR}/include/openenclave/edl)
-if (OE_SGX)
-  set(SGX_EDL_DIR ${EDL_DIR}/sgx)
-endif ()
 
 ##==============================================================================
 ##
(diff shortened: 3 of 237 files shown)


mingweishih:
bors try

tests/cmake_name_conflict/name_conflict.edl
```diff
@@ -4,7 +4,11 @@
 enclave {
     from "openenclave/edl/logging.edl" import oe_write_ocall;
     from "openenclave/edl/fcntl.edl" import *;
-    from "platform.edl" import *;
+#ifdef OE_SGX
```
anakrish, line 7: It seems like the user now has to add 
```c
#ifdef OE_SGX
    from "openenclave/edl/sgx/platform.edl" import *;
#else
    from "openenclave/edl/optee/platform.edl" import *;
#endif
```

Can we avoid the user from having to do this?

mingweishih: After fully supporting system edl opt in, there will be more platform-dependent cases that can not simply rely on specifying path. For example,

#ifdef OE_SGX
    from "openenclave/edl/sgx/cpu.edl" import *;
     from "openenclave/edl/sgx/switchless.edl" import *;
#else
    from "openenclave/edl/optee/platform.edl" import *;
#endif

Given that, I think we need to use preprocessor (I thought this is the reason to support preprocessor, please correct me if I'm wrong).

mingweishih: Note that this PR currently modifies tests only.

anakrish: With preprocessor, there are multiple options. For example here is an alternative.

openenclave/edl/platform.edl
```
#ifdef OE_SGX_FEATURE_CPUID
from "openenclave/edl/sgx/cpu.edl" import *;
#endif 

#ifdef OE_SGX_FEATURE_SWITCHLESS
from "openenclave/edl/sgx/switchless.edl" import *;
#endif 

#ifdef OE_TRUSTZONE
from "openenclave/edl/optee/platform.edl" import *;
#endif 
```

Then the user edl can just import openenclave/edl/platform.edl :
user.edl
```c
from "openenclave/edl/platform.edl" import *;
``` 
and select various features by specifying the appropriate defines:
```bash
oeedger8r  user.edl --search-path path-to-oe -DOE_SGX_FEATURE_SWITCHLESS -DOE_SGX_FEATURE_CPUID.
```

I'm not saying that this is the way to go. But it seems there are different user experiences possible.

mingweishih: If we keep the import of `platform.edl` clean, we would cause the import of platform-dependent edls implicit. For example, in your example, it would be difficult for users to figure out the relationship between `-DOE_SGX_FEATURE_SWITCHLESS` and
```
#ifdef OE_SGX_FEATURE_SWITCHLESS
from "openenclave/edl/sgx/switchless.edl" import *;
#endif 
```
given that this import is hidden from users.

jhand2: Really my vision for this was that users would not import platform.edl. Instead they would just import the specific edl files they want for each platform they support.

I'm ok with leaving platform.edl as a convenience, but I agree with Ming-wei, I don't really want it to hide any more details than it already does.

anakrish: > If we keep the import of `platform.edl` clean, we would cause the import of platform-dependent edls implicit. For example, in your example, it would be difficult for users to figure out the relationship between `-DOE_SGX_FEATURE_SWITCHLESS` and
> 
> ```
> #ifdef OE_SGX_FEATURE_SWITCHLESS
> from "openenclave/edl/sgx/switchless.edl" import *;
> #endif 
> ```
> 
> given that this import is hidden from users.

In that case you'd have to document that flag `OE_SGX_FEATURE_SWITCHLESS` brings in switchless feature.
The intent of me bringing up that example was that there are multiple user experiences possible; not that I recommend one over the other.

So far I have not seen any documentation/discussion of the intended user experience. 
It is really not clear from this PR what the user experience is and I'd like our tests to mirror the user experience.
Hence I'm quite hesitant to approve this PR.

anakrish: > Really my vision for this was that users would not import platform.edl. Instead they would just import the specific edl files they want for each platform they support.
> 
> I'm ok with leaving platform.edl as a convenience, but I agree with Ming-wei, I don't really want it to hide any more details than it already does.

That's fair. But it is not clear how a user can figure out which edl files to import in the first place. It might be easy for us; but not quite obvious for the user.

CodeMonkeyLeet:
@mingweishih Sorry, late to the party on this PR, but I echo @anakrish's concerns about coming up with a coherent story for how a developer should understand EDL selection. I'd be interested in understanding how to consolidate the different axes (arch vs. feature set) of EDL selection, but I agree that it does not have to be in this PR. Please loop me in on where ever that discussion lands.

Issue #3201, opened 2020-06-30
Decide the precedent of how to import platform-dependent system edls.


mingweishih:
OE currently have umbrellas `platform.edl` for both sgx and optee. This provides two options for an platform-independent enclave to use.

- Option 1: Specify the search path to the oeedger8r
  In this option, an enclave imports the `platform.edl` in its edl as follows.
  ```
  from "platform.edl" import *;
  ```
  and specify the search path to the oeedger8r as follows.
  - `--search-path ${OE_INCLUDEDIR}/openenclave/edl/sgx` to use `sgx/platform.edl`
  - `--search-path ${OE_INCLUDEDIR}/openenclave/edl/optee` to use `optee/platform.edl`
  
- Option 2: Use the C-style preprocessor supported by oeedger8r
  In this option, the user passes the `-DOE_SGX` argument to the oeedger8r and import the `platform.edl` in its edl as follows.
  ```
  #ifdef OE_SGX
      from "openenclave/edl/sgx/platform.edl" import *;
  #else
      from "openenclave/edl/optee/platform.edl" import *;
  #endif
  ```

The option 2 also allows us to deprecate the `platform.edl` and to explicitly import platform-dependent edls.
For example,
 ```
#ifdef OE_SGX
    from "openenclave/edl/sgx/cpu.edl" import *;
    from "openenclave/edl/sgx/switchless.edl" import *;
#else
    from "openenclave/edl/optee/platform.edl" import *;
#endif
```
This, however, may come with the concern of imposing burden to the users. Alternatively, we could tweak the option 1.
For example,
```
/* sgx/platform.edl */
...
#ifdef SGX_SWITCHLESS
from "openenclave/edl/sgx/switchless.edl" import *;
#endif
```
In this case, the user only need to pass the `-DSGX_SWITCHLESS` argument to the oeedger8r when he or she wants to opt-into the switchless feature.

mingweishih:
cc @dthaler @anakrish @CodeMonkeyLeet @jhand2 @radhikaj

anakrish:
@mingweishih @dthaler @CodeMonkeyLeet @jhand2 

Do we all agree that it is better user experience to not have the user specify different search-paths for system edls based on platform? If so, I'd propose that we go further and propose that system-edls be available in the path by default without the user having to specify the search-path explicitly. This is analogous to the standard C headers that are available by default without the user having to specify -I/usr/include to the compiler.


One thing I'm not clear is this:
```user.edl

[...]

PR #3203, opened 2020-06-30
Support debug.edl opt-in/out

mingweishih:
This PR allows user to opt-out `debug.edl`, which was a hard requirement (part of #2818). If user does not import the debug.edl but invokes the `backtrace()`, he should be able to see the following log (if the logging feature is enabled)
```
backtrace_enc:In-enclave backtrace is not supported. To enable, please add 

from "openenclave/edl/sgx/debug.edl" import *;

in the edl file.
```

Note that the goal of this PR is having the implementation ready (only modify the edl in the backtrace test). Will wait for getting the consensus (per #3201) before generalizing the changes to both tests and samples.

### DIFF
--- a/enclave/core/sgx/backtrace.c
+++ b/enclave/core/sgx/backtrace.c
@@ -19,6 +19,51 @@
 #error "optimized __builtin_return_address() not supported by Intel compiler"
 #endif
 
+#if !defined(OE_USE_BUILTIN_EDL)
+/**
+ * Declare the prototype of the following function to avoid the
+ * missing-prototypes warning.
+ */
+oe_result_t _oe_sgx_backtrace_symbols_ocall(
+    oe_result_t* _retval,
+    oe_enclave_t* oe_enclave,
+    const uint64_t* buffer,
+    size_t size,
+    void* symbols_buffer,
+    size_t symbols_buffer_size,
+    size_t* symbols_buffer_size_out);
+
+/**
+ * Make the following OCALL weak to support the system EDL opt-in.
+ * When the user does not opt into (import) the EDL, the linker will pick
+ * the following default implementation. If the user opts into the EDL,
+ * the implemention (which is strong) in the oeedger8r-generated code will be
+ * used.
+ */
+oe_result_t _oe_sgx_backtrace_symbols_ocall(
+    oe_result_t* _retval,
+    oe_enclave_t* oe_enclave,
+    const uint64_t* buffer,
+    size_t size,
+    void* symbols_buffer,
+    size_t symbols_buffer_size,
+    size_t* symbols_buffer_size_out)
+{
+    OE_UNUSED(oe_enclave);
+    OE_UNUSED(buffer);
+    OE_UNUSED(size);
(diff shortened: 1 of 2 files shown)


enclave/core/sgx/backtrace.c
```diff
@@ -106,10 +151,16 @@ char** oe_backtrace_symbols_impl(
     char** argv = NULL;
 
     if (!buffer || size < 0)
+    {
+        retval = OE_INVALID_PARAMETER;
         goto done;
+    }
```
dthaler, line 157: OE_RAISE(OE_INVALID_PARAMETER);
?   And same question for line 160-163.
I think the difference is it's shorter, and it also logs the result.

enclave/core/sgx/backtrace.c
```diff
@@ -178,6 +229,11 @@ char** oe_backtrace_symbols_impl(
     if (argv)
         free_fcn(argv);
 
+    if ((oe_result_t)retval == OE_UNSUPPORTED)
+        OE_TRACE_WARNING(
+            "In-enclave backtrace is not supported. To enable, please import "
+            "debug.edl.\n");
```
dthaler, line 235: Shouldn't it say something like import "*" from debug.edl?   The developer can't just import the file, they have to specify which APIs to import from it.

mingweishih: My intention is informing the user to take the action rather than referring to the exact line of code they need to add. Do you think the latter is more appropriate?

By the way, oeedger8r actually supports the syntax of `import debug.edl`, which does the same as `from debug.edl import *`.  Refer to @anakrish's great [document](https://github.com/openenclave/oeedger8r-cpp/blob/master/docs/Grammar.md).

enclave/core/sgx/backtrace.c
mingweishih: Updated.

enclave/core/sgx/backtrace.c
mingweishih: Updated the message to be more specific

dthaler: Interesting, I didn't know you could just import debug.edl.  All the samples do the more verbose version, so I'm not sure how developers would discover the shorter version, as it's not referenced in any OE getting started doc that I can find (the doc you mention is in a different project).

dthaler: So yes, I think the latter is more appropriate since as is, it is not clear to the developer what to do.

mingweishih: Updated. Now the example of the log is
```
ssh_office@oe-mingwei-1804-8c:~/workspace/oe-ms2/build/tests/backtrace$ OE_LOG_LEVEL=WARNING ./host/backtrace_host enc/backtrace_enc 
=== test()
2020-07-01T18:19:25.000000Z [(E)WARN] tid(0x7f3dc9767740) | backtrace_enc:In-enclave backtrace is not supported. To enable, please add 

from "openenclave/edl/sgx/debug.edl" import *;

in the edl file.
```

PR #3206, opened 2020-07-01
create enclaves from multiple child and parent processes

jinghe-INTC:
It is a enhancement of the case of creating enclave in child process.

### DIFF
--- a/tests/child_process/CMakeLists.txt
+++ b/tests/child_process/CMakeLists.txt
@@ -16,3 +16,6 @@ add_enclave_test(tests/child_process_ecall child_process_host child_process_enc
 
 add_enclave_test(tests/child_process_create child_process_host
                  child_process_enc 2)
+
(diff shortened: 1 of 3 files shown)


jinghe-INTC:
bors try

jinghe-INTC:
> LGTM
> Thanks for locking down various fork scenarios.

No problem.
Unfortunately it didn't pass the code coverage check: "Coverage of 59.60% is lower than expected 60.00%." I am thinking about how to fix it.

mingweishih:
> > LGTM
> > Thanks for locking down various fork scenarios.
> 
> No problem.
> Unfortunately it didn't pass the code coverage check: "Coverage of 59.60% is lower than expected 60.00%." I am thinking about how to fix it.

You can ignore that for now. We currently didn't take the code coverage as hard requirement. It's more like an experimental feature.

PR #3207, opened 2020-07-01
Add check to ensure EEID data is within enclave memory.

wintersteiger:
This adds a check to ensure the EEID marker offset is in fact within the enclave boundaries. 

The host is in control of the offset value on the marker page. So, in theory, they could make this point into host memory, so that the host has control over the EEID content during attestation, which the hardware would not catch.

Thanks to @bodzhang for reporting this issue and suggesting a fix.

### DIFF
--- a/enclave/core/sgx/init.c
+++ b/enclave/core/sgx/init.c
@@ -73,7 +73,8 @@ static oe_result_t _eeid_patch_memory()
 
         uint8_t* heap_end = (uint8_t*)__oe_get_heap_end();
         uint8_t* tcs_end =
-            heap_end + (6 + 2 + eeid->size_settings.num_stack_pages) *
+            heap_end + (OE_SGX_TCS_CONTROL_PAGES + OE_SGX_TCS_GUARD_PAGES +
+                        eeid->size_settings.num_stack_pages) *
                            OE_PAGE_SIZE * eeid->size_settings.num_tcs;
 
         /* EEID must not overlap with tcs/stack/control pages */
--- a/enclave/core/sgx/init.c
+++ b/enclave/core/sgx/init.c
@@ -69,12 +69,14 @@ static oe_result_t _eeid_patch_memory()
         uint8_t* enclave_base = (uint8_t*)__oe_get_enclave_base();
         uint8_t* heap_base = (uint8_t*)__oe_get_heap_base();
         oe_eeid_marker_t* marker = (oe_eeid_marker_t*)heap_base;
-        uint8_t* eeid_ptr = enclave_base + marker->offset;
+        oe_eeid_t* eeid = (oe_eeid_t*)(enclave_base + marker->offset);
 
-        if (!_is_within_eeid_enclave(eeid_ptr))
+        if (!_is_within_eeid_enclave((uint8_t*)eeid) ||
+            !_is_within_eeid_enclave(eeid->data) ||
+            !_is_within_eeid_enclave(eeid->data + eeid->data_size))
             oe_abort();
 
-        oe_eeid = (oe_eeid_t*)eeid_ptr;
+        oe_eeid = eeid;
 
         // Wipe the marker page
         memset(heap_base, 0, OE_PAGE_SIZE);
--- a/enclave/core/sgx/init.c
+++ b/enclave/core/sgx/init.c
@@ -52,6 +52,14 @@ static int _is_eeid_base_image(
            properties->header.size_settings.num_tcs == 1;
 }
 
+static int _is_within_eeid_enclave(const uint8_t* ptr)
+{
+    /* For EEID enclaves, the SGX-reported enclave size is always
+     * EEID_SECS_SIZE. */
+    const uint8_t* base = (const uint8_t*)__oe_get_enclave_base();
+    return ptr >= base && ptr < base + EEID_SECS_SIZE;
+}
+
 static oe_result_t _eeid_patch_memory()
 {
     oe_result_t r = OE_OK;
@@ -61,7 +69,12 @@ static oe_result_t _eeid_patch_memory()
         uint8_t* enclave_base = (uint8_t*)__oe_get_enclave_base();
         uint8_t* heap_base = (uint8_t*)__oe_get_heap_base();
         oe_eeid_marker_t* marker = (oe_eeid_marker_t*)heap_base;
-        oe_eeid = (oe_eeid_t*)(enclave_base + marker->offset);
+        uint8_t* eeid_ptr = enclave_base + marker->offset;
+
+        if (!_is_within_eeid_enclave(eeid_ptr))
+            oe_abort();
+
+        oe_eeid = (oe_eeid_t*)eeid_ptr;
 
         // Wipe the marker page
         memset(heap_base, 0, OE_PAGE_SIZE);
--- a/enclave/core/sgx/init.c
+++ b/enclave/core/sgx/init.c
@@ -52,14 +52,6 @@ static int _is_eeid_base_image(
            properties->header.size_settings.num_tcs == 1;
 }
 
-static int _is_within_eeid_enclave(const uint8_t* ptr)
-{
-    /* For EEID enclaves, the SGX-reported enclave size is always
-     * EEID_SECS_SIZE. */
-    const uint8_t* base = (const uint8_t*)__oe_get_enclave_base();
-    return ptr >= base && ptr < base + EEID_SECS_SIZE;
-}
-
 static oe_result_t _eeid_patch_memory()
 {
     oe_result_t r = OE_OK;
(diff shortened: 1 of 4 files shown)


enclave/core/sgx/init.c
```diff
@@ -61,7 +69,12 @@ static oe_result_t _eeid_patch_memory()
         uint8_t* enclave_base = (uint8_t*)__oe_get_enclave_base();
         uint8_t* heap_base = (uint8_t*)__oe_get_heap_base();
         oe_eeid_marker_t* marker = (oe_eeid_marker_t*)heap_base;
-        oe_eeid = (oe_eeid_t*)(enclave_base + marker->offset);
+        uint8_t* eeid_ptr = enclave_base + marker->offset;
+
+        if (!_is_within_eeid_enclave(eeid_ptr))
```
bodzhang, line 74: Need to check the fixed size portion of eeid (sizeof(oe_eeid_t)) is within the enclave first,  then read the size of the variable length portion (eeid_ptr->data_size + eeid_ptr->signature_size), and check the whole data structure is within the enclave.

wintersteiger: Good point, we need to check the end as well.

wintersteiger: (Done)

enclave/core/sgx/init.c
```diff
@@ -52,6 +52,14 @@ static int _is_eeid_base_image(
            properties->header.size_settings.num_tcs == 1;
 }
 
+static int _is_within_eeid_enclave(const uint8_t* ptr)
```
bodzhang, line 55: OE SDK already has a _oe_is_within_enclave(const void* p, size_t n)_ API. No need to add a EEID-specific version. _oe_is_within_enclave(...)_ gets the enclave size from oe_enclave_properties_sgx.image_info.enclave_size, which should have been set to EEID_SECS_SIZE in EEID case.

wintersteiger: No, that's not the case. `oe_enclave_properties_sgx.image_info.enclave_size` holds the base image enclave size (i.e. heap/stack == 0). Note that there are two calls to `_calculate_enclave_size` and only the result of the first one is patched into the image. For EEID, we patch that same base-image value into that variable, because otherwise we would change the base image hash. I couldn't find a way to get the SECS data directly, but conveniently, we always set it to the maximum anyways.

bodzhang: In create.c:

    /* Calculate the size of this enclave in memory */
    OE_CHECK(_calculate_enclave_size(
        image_size, &props, &enclave_end, &enclave_size));

    /* Perform the ECREATE operation */
    OE_CHECK(oe_sgx_create_enclave(
        context, enclave_size, enclave_end, &enclave_addr));

    /* Save the enclave base address, size, and text address */
    enclave->addr = enclave_addr;
    enclave->size = enclave_size;
    enclave->text = enclave_addr + oeimage.text_rva;

    /* Patch image */
    OE_CHECK(oeimage.patch(&oeimage, enclave_size));

I thought _calculate_enclave_size() set enclave_size to EEID_SECS_SIZE (for EEID case) and the value is used byoeimage.patch to set oe_enclave_properties_sgx.image_info.enclave_size.

Anyway, if oe_enclave_properties_sgx.image_info.enclave_size indeed holds the base image enclave size (heap/stack=0), it needs to be fixed. The enclave code also uses it to check whether an output buffer that is supposed to be outside of enclave is indeed outside of enclave. If that check is not reliable, the enclave code can unintentionally overwrite its own heap or stack.

enclave/core/sgx/init.c
```diff
@@ -69,12 +69,14 @@ static oe_result_t _eeid_patch_memory()
         uint8_t* enclave_base = (uint8_t*)__oe_get_enclave_base();
         uint8_t* heap_base = (uint8_t*)__oe_get_heap_base();
         oe_eeid_marker_t* marker = (oe_eeid_marker_t*)heap_base;
-        uint8_t* eeid_ptr = enclave_base + marker->offset;
+        oe_eeid_t* eeid = (oe_eeid_t*)(enclave_base + marker->offset);
 
-        if (!_is_within_eeid_enclave(eeid_ptr))
+        if (!_is_within_eeid_enclave((uint8_t*)eeid) ||
+            !_is_within_eeid_enclave(eeid->data) ||
+            !_is_within_eeid_enclave(eeid->data + eeid->data_size))
```
bodzhang, line 76: I think eeid->data_size does not count the SIGSTRUCT at the end of the data[] buffer.

I just realized that the code should also check to make sure the eeid page location the marker->offset claims does not overlap with the heap or the TCS/Stack pages.

enclave/core/sgx/init.c
wintersteiger: You may be right, I have to read the code again as this change a few times.

enclave/core/sgx/init.c
wintersteiger: Would an overlap with the TCS/stack pages be a security problem? It's easy to add a check that it's within heap + eeid size, so I can add that anyways.

enclave/core/sgx/init.c
wintersteiger: You're right, it is in fact patched into the base image too now, so it should hold the right size during enclave entry.

enclave/core/sgx/init.c
wintersteiger: (Yes, done)

enclave/core/sgx/init.c
wintersteiger: Done!

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -28,6 +28,13 @@ typedef struct _oe_sgx_enclave_image_info_t
 /* Max number of threads in an enclave supported */
 #define OE_SGX_MAX_TCS 32
 
+#ifdef OE_WITH_EXPERIMENTAL_EEID
+/* This is the maximum size SGX (currently) allows. When signing EEID base
+ * images we don't know the size that the final image will have, so chose
+ * this maximum. */
+#define EEID_SECS_SIZE 68719476736
```
jhand2, line 35: A few things:

* Can you add to the comment (a) why this number and (b) what this is in GB
* Can you write this in hex instead? It looks like a random number in decimal

wintersteiger: Sure, done!

enclave/core/sgx/init.c
```diff
@@ -60,8 +60,25 @@ static oe_result_t _eeid_patch_memory()
     {
         uint8_t* enclave_base = (uint8_t*)__oe_get_enclave_base();
         uint8_t* heap_base = (uint8_t*)__oe_get_heap_base();
-        oe_eeid_marker_t* marker = (oe_eeid_marker_t*)heap_base;
-        oe_eeid = (oe_eeid_t*)(enclave_base + marker->offset);
+        const oe_eeid_marker_t* marker = (oe_eeid_marker_t*)heap_base;
```
anakrish, line 63: @wintersteiger @bodzhang 
We don't measure the contents of the heap pages currently. Therefore the host could initialize it to what-ever value it wants to. Do we have special logic to measure the contents of the marker page even though it is the first heap page?

wintersteiger: The marker page can't be measured at signing time as we don't know the offset, but I suppose we could measure it at enclave creation time and update the verifier accordingly. There are a number of checks in `init.c` to make sure it points somewhere sensible, but I don't think it makes sure that it's on a measured page. @bodzhang what's your opinion, should the marker be measured?

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -28,6 +28,13 @@ typedef struct _oe_sgx_enclave_image_info_t
 /* Max number of threads in an enclave supported */
 #define OE_SGX_MAX_TCS 32
 
+#ifdef OE_WITH_EXPERIMENTAL_EEID
+/* This is the maximum size SGX (currently) allows (64GB). When signing EEID
+ * base images we don't know the size that the final image will have, so we
+ * chose this maximum. */
+#define EEID_SECS_SIZE 0x1000000000
```
CodeMonkeyLeet, line 35: This doesn't need to be a publicly exposed definition, could you move it into the include/openenclave/internal/eeid.h header instead?

Also, consider `EEID_ELRANGE` instead; the `SECS` is a well-known data structure in SGX and the current name implies that it is somehow set to a 64GB size, which makes for a confusing read.

enclave/core/sgx/init.c
CodeMonkeyLeet: This PR is a stopgap fix before the suggestions proposed in #3269, correct? In that updated design, there is a fixed EEID page in the enclave layout, which precludes the retargeting attack. That would be my preference, but the current fix is fine in the meantime.

enclave/core/sgx/init.c
```diff
@@ -60,8 +60,25 @@ static oe_result_t _eeid_patch_memory()
...
+        if (!oe_is_within_enclave(eeid, sizeof(oe_eeid_t)) ||
+            !oe_is_within_enclave(
+                eeid->data, eeid->data_size + eeid->signature_size))
+            oe_abort();
+
+        oe_eeid = eeid;
+
+        uint8_t* heap_end = (uint8_t*)__oe_get_heap_end();
+        uint8_t* tcs_end =
+            heap_end + (6 + 2 + eeid->size_settings.num_stack_pages) *
```
CodeMonkeyLeet, line 76: Please add named definitions for the magic numbers 6 and 2; this source file is well removed from the context of where those numbers come from, e.g.:
```c
#define OE_SGX_TCS_CONTROL_PAGES 6
#define OE_SGX_TCS_GUARD_PAGES 2
```

enclave/core/sgx/init.c
wintersteiger: That's correct, this is just a fix for the current design, the new one won't won't use a marker page.

include/openenclave/bits/sgx/sgxproperties.h
wintersteiger: Sure, done! Just in case: Yes, this is indeed the size that is put into the SECS (`sgx_secs_t` via `enclave_size`) and it is set to 64GB.

enclave/core/sgx/init.c
wintersteiger: Done!

include/openenclave/internal/eeid.h
```diff
@@ -15,6 +15,13 @@
 
 #include <openenclave/bits/eeid.h>
 
+/* When signing EEID base images we don't know the size that the final image
+ * will have, so we chose a reasonably large size here (64GB). */
+#define EEID_ELRANGE 0x1000000000
```
dthaler, line 20: This number was previously in an SGX-only file and eeid.h is not in an SGX directory or with an sgx in the filename.  So this needs to have SGX in the macro name like lines 22 and 23 do.  Suggest OE_SGX_EEID_ELRANGE

wintersteiger: Good point, I'll add that!

CodeMonkeyLeet:
@bodzhang Did you have any remaining comments about this?

wintersteiger:
bors try

CodeMonkeyLeet:
bors try

CodeMonkeyLeet:
bors delegate+

PR #3208, opened 2020-07-01
Cleanup to conform to OE naming conventions

mikbras:
This PR:
- Adds oe prefixes to various symbols that were lacking them.
- Makes certain functions static (file scope).
- Renames a couple of functions from camel notation to snake notation.

### DIFF
--- a/common/sgx/collateral.c
+++ b/common/sgx/collateral.c
@@ -22,7 +22,7 @@
 #include "tcbinfo.h"
 
 // Defaults to Intel SGX 1.8 Release Date.
-oe_datetime_t _sgx_minimim_crl_tcb_issue_date = {2017, 3, 17};
+oe_datetime_t __oe_sgx_minimim_crl_tcb_issue_date = {2017, 3, 17};
 
 oe_result_t __oe_sgx_set_minimum_crl_tcb_issue_date(
     uint32_t year,
@@ -36,7 +36,7 @@ oe_result_t __oe_sgx_set_minimum_crl_tcb_issue_date(
     oe_datetime_t tmp = {year, month, day, hours, minutes, seconds};
 
     OE_CHECK(oe_datetime_is_valid(&tmp));
-    _sgx_minimim_crl_tcb_issue_date = tmp;
+    __oe_sgx_minimim_crl_tcb_issue_date = tmp;
 
     result = OE_OK;
 done:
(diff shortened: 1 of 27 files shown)


anakrish:
bors try

common/sgx/collateral.c
```diff
@@ -22,7 +22,7 @@
 #include "tcbinfo.h"
 
 // Defaults to Intel SGX 1.8 Release Date.
-oe_datetime_t _sgx_minimim_crl_tcb_issue_date = {2017, 3, 17};
+oe_datetime_t __oe_sgx_minimim_crl_tcb_issue_date = {2017, 3, 17};
```
anakrish, line 25: @mikbras Thanks for doing this.
The [DeveloperGuide](https://github.com/openenclave/openenclave/blob/master/docs/Community/DevelopmentGuide.md) says that
> Note that we no longer use CamelCase nor double underscores (__), but you may find remnants and so again should prefer the local style. This is especially the case for classes, which are still using PascalCase

Is the change to `__`  conformant?.

mikbras: Hmm? Good question. Our global variables use double underscores in many places. Our functions do not. Should we get rid of double underscore prefixes altogether? I can make that part of the PR.

dthaler: Good point Anand.  Yes, the C++ standard reserves double underscores at the start for language/compiler use, and should not be used by libraries.  https://en.cppreference.com/w/cpp/keyword says:

> Also, all identifiers that contain a double underscore __ in any position and each identifier that begins with an underscore followed by an uppercase letter is always reserved and all identifiers that begin with an underscore are reserved for use as names in the global namespace. See identifiers for more details.

Similarly it's discussed on forums like https://www.experts-exchange.com/questions/20552640/does-the-underscore-or-double-underscore-mean-something-in-c.html#:~:text=The%20C%20and%20C%2B%2B%20standards%20reserve%20symbols%20beginning,some%20assurance%20of%20not%20conflicting%20with%20users%27%20programs.

mikbras: Some of our 'classics' use double underscores:
```
0000000000000000 T __oe_get_heap_base
0000000000000000 T __oe_get_heap_end
0000000000000000 T __oe_get_heap_size
```
My opinion is that we should change all "__oe_" prefixed functions and variables to "oe_". 

What do you think?

anakrish: That sounds like a nice idea; consistent with our Developer Guidelines. 

You may run into difficulty renaming `__oe_ecalls_table` since it is generated by edger8r. But I or @mingweishih can fix that in a subsequent PR easily.

dthaler: Agreed, we should change those to be C++ compliant, and be compliant with the OE guideline Anand cited.

radhikaj:
@mikbras Can you address the comments please? I can then merge this.

radhikaj:
@mikbras How would you like to proceed on this PR?

radhikaj:
closing this PR for now. Please reopen when feedback can be addressed and the PR merged.

PR #3209, opened 2020-07-01
Fix typo

anakrish:


anakrish:
bors try

anakrish:
@mingweishih There seems to be a genuine problem with the tls_e2e test 

> 49: 2020-07-01T20:35:41.000000Z [(E)ERROR] tid(0x137c) | tls_server_enc:Last error was: -82 - NET - Failed to get an IP address for the given hostname

mingweishih:
I think it requires importing socket.edl. We didn't hit the failure when removing the import because of this typo (the test is not run)..

anakrish:
bors try

PR #3213, opened 2020-07-02
Add documentation for system edls

### LINKED ISSUE #2719: Document which features use which system ocalls

mingweishih:
This PR adds a document that enumerates the existing system EDLs and their dependent public APIs/syscalls.

fixes #2719 

### DIFF
--- a/docs/GettingStartedDocs/APIs_and_Libs.md
+++ b/docs/GettingStartedDocs/APIs_and_Libs.md
@@ -21,3 +21,7 @@ The subset of libcxx functionality provided by oelibcxx for use inside an enclav
 ## [mbedtls library](/docs/MbedtlsSupport.md)
 
 The subset of [mbedtls](https://tls.mbed.org/) functionality for use inside an enclave.
+
+## [System EDL files](/docs/SystemEdls.md)
+
+The list of system EDL files that allow for user opt-in.
--- a/docs/SystemEdls.md
+++ b/docs/SystemEdls.md
@@ -1,0 +1,202 @@
+# Open Enclave System EDL Files
+
+This document lists the system EDL files supported by Open Enclave. To opt-into these files, user can
+use `from path/to/edl import *` syntax to import all the ecalls/ocalls defined by the file.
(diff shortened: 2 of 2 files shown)


mingweishih:
> Looks like an ambitious exercise, and if we want to go this route, thanks for putting in this effort. We will have to make sure other PRs keep it up to date. Just one request: when adding documentation in a new file, we should update some existing file to point to it so it becomes discoverable.

I agree that if we take this route, we need to frequently update this document. However, I think it's always better than having no documentation especially we starts allowing these EDL files to be optional.  Would you recommend any approach that requires less maintenance effort?

radhikaj:
> Looks like an ambitious exercise, and if we want to go this route, thanks for putting in this effort. We will have to make sure other PRs keep it up to date. Just one request: when adding documentation in a new file, we should update some existing file to point to it so it becomes discoverable.

@mingweishih, can we add a link to this document at this page: https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/APIs_and_Libs.md

mingweishih:
> > Looks like an ambitious exercise, and if we want to go this route, thanks for putting in this effort. We will have to make sure other PRs keep it up to date. Just one request: when adding documentation in a new file, we should update some existing file to point to it so it becomes discoverable.
> 
> @mingweishih, can we add a link to this document at this page: https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/APIs_and_Libs.md

Sure, the change is already in this PR.

mingweishih:
> Looks good! Only minor nit is I would left align the text in each table cell.

Thanks for the suggestion. Updated!

PR #3214, opened 2020-07-02
Support sgx/attestation.edl opt-out

mingweishih:
This PR allows `sgx/attestation.edl` to be opt-out, which was a hard requirement (part of #2818).  This edl file includes two ecalls and four ocalls. When a user invokes those calls without importing the edls, he should be the corresponding log message (when the logging feature is enabled).

For example, the message of `oe_verify_local_report_ecall` is as follows
```
oe_verify_local_report_ecall is not supported. To enable, please add 

from "openenclave/edl/sgx/attestation.edl" import oe_verify_local_report_ecall;

in the edl file.
```

Given the `oe_get_report_v2_ecall` ecall and all the four ocalls are currently required by the attestation, the message for each of calls is the same as follows.
```
sgx remote attestation is not fully supported. To enable, please add 

from "openenclave/edl/sgx/attestation.edl" import *;

in the edl file.
```

Note that the goal of this PR is having the implementation ready (only modifies the edl in the report test). Will wait for getting the consensus (per #3201) before generalizing the changes to both tests and samples.

### DIFF
--- a/enclave/core/sgx/report.c
+++ b/enclave/core/sgx/report.c
@@ -16,6 +16,81 @@
 #include <openenclave/internal/utils.h>
 #include "platform_t.h"
 
+#if !defined(OE_USE_BUILTIN_EDL)
+/**
+ * Declare the prototypes of the following functions to avoid the
+ * missing-prototypes warning.
+ */
+oe_result_t _oe_get_qetarget_info_ocall(
+    oe_result_t* _retval,
+    const oe_uuid_t* format_id,
+    const void* opt_params,
+    size_t opt_params_size,
+    sgx_target_info_t* target_info);
+
+oe_result_t _oe_get_quote_ocall(
+    oe_result_t* _retval,
+    const oe_uuid_t* format_id,
+    const void* opt_params,
+    size_t opt_params_size,
+    const sgx_report_t* sgx_report,
+    void* quote,
+    size_t quote_size,
+    size_t* quote_size_out);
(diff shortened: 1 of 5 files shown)


mingweishih:
bors try

enclave/core/sgx/report.c
```diff
@@ -127,14 +202,21 @@ static oe_result_t _get_sgx_target_info(
...
-        return OE_FAILURE;
+    OE_CHECK(oe_get_qetarget_info_ocall(
+        &retval, format_id, opt_params, opt_params_size, target_info));
+    result = (oe_result_t)retval;
 
-    return (oe_result_t)retval;
+done:
+    if (result == OE_UNSUPPORTED)
+        OE_TRACE_WARNING(
+            "sgx remote attestation is not fully supported. To "
```
dthaler, line 215: should this say "SGX remote attestation is not enabled." instead of "fully supported"?  And same question on line 253 and in other files.  In other words, if OE_UNSUPPORTED is returned above, is there still some part of SGX remote attestation that is partially supported without adding the import in the EDL file?

mingweishih: Only one call (`oe_verify_local_report_ecall`), which is part of verifier, can be partially enabled. The rest calls are part of attester so should be enabled all together. Given that, I update the log message as you suggested and leave the message of `oe_verify_local_report_ecall` as the same.

PR #3217, opened 2020-07-03
direct [in] pointer as ECALL parameter, will not leak data from enclave

yanxue-intel:


anakrish:
bors try

PR #3218, opened 2020-07-03
fix oesign digest usage typo issue. add 3 oesign key test cases

johnvenn:
Signed-off-by: Xiangping Ji <xiangping.ji@intel.com>

1. fix oesign digest typo in usage
2. rename oesign test case of invalid-key-file test to not-exist-key-file
3. added 3 test cases : 
   3.1 key file end wrong -- PASS
   3.2 sign using public key -- PASS
   3.3 sign using key file format wrong -- FAIL


johnvenn:
not sure why there is a reference issue, I will reopen it

PR #3221, opened 2020-07-04
Replace CMAKE_SOURCE_DIR with PROJECT_SOURCE_DIR

### LINKED ISSUE #3220: Cannot include OE directly from another CMake project
The CMake files of OE have a few `CMAKE_SOURCE_DIR` in them which should all be replaced by [`PROJECT_SOURCE_DIR`](https://cmake.org/cmake/help/v3.18/variable/PROJECT_SOURCE_DIR.html) so that OE can be consumed with `add_subdirectory()`. `CMAKE_SOURCE_DIR` refers to the top-most folder (the consuming project) which is incorrect.

letmaik:
Fixes #3220.

### DIFF
--- a/enclave/core/CMakeLists.txt
+++ b/enclave/core/CMakeLists.txt
@@ -1,8 +1,8 @@
 # Copyright (c) Open Enclave SDK contributors.
 # Licensed under the MIT License.
 
-set(OE_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/include)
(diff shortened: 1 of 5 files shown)


radhikaj:
@letmaik There are still some places where CMAKE_SOURCE_DIR  gets used(samples, for instance).

letmaik:
@radhikaj Samples are stand-alone and declare their own CMake `project`. They don't require any change.

PR #3152, opened 2020-06-18
add experimental testing to CI

BRMcLaren:
Add experimental flag in CI to validate experimental features.

PR #3222, opened 2020-07-04
Add EEID CI job

letmaik:
Since this is an experimental feature I've only added a single new job, which is enough to catch accidental regressions.

@wintersteiger Does this make sense?


wintersteiger:
Sure that makes a lot of sense! I'm not sure what the policy for experimental feature CI jobs is, but it would be great to get at least some basic testing done automatically.

anakrish:
bors try

jxyang:
LGTM

@radhikaj @jhand2

letmaik:
@jxyang I can't find "ACC1804 clang-7 Debug EEID" in the jenkins jobs. Am I looking at the wrong thing?

mingweishih:
> @jxyang I can't find "ACC1804 clang-7 Debug EEID" in the jenkins jobs. Am I looking at the wrong thing?

@letmaik you're adding the job into nightly build, so it will not triggered in the bors try run. This in fact makes sense to me as this is an experimental feature.

I guess the question here is how do we test experimental features in CI/CD. Should we add a single configuration in the daily build, run in only in the nightly build, or have a separated pipeline for doing it?

cc @BRMcLaren

BRMcLaren:
> I guess the question here is how do we test experimental features in CI/CD. Should we add a single configuration in the daily build, run in only in the nightly build, or have a separated pipeline for doing it?
> 
> cc @BRMcLaren

I keep going back and forth on it honestly. What I think it should be is a non-gating optional pipeline as part of the PR process, but the infrastructure right now at this moment will not support that.

Let's add it to the PR build to block against accidental regression and open a bug to add the optional?

radhikaj:
@letmaik @jxyang @wintersteiger , please see previous PR for this: https://github.com/openenclave/openenclave/pull/3152
That PR adds a run with EEID on for every PR. If you are actively developing on this, I would suggest that is a better way to go.

letmaik:
> @letmaik @jxyang @wintersteiger , please see previous PR for this: #3152 that @BRMcLaren opened.
> That PR adds EEID run for every PR. If you are actively developing on this, I would suggest that is a better way to go.

Ah I didn't see that. Let's go with the other PR then.

PR #3224, opened 2020-07-06
fix oesign digest usage typo

johnvenn:
1. fix oesign digest usage typo.

### DIFF
--- a/tools/oesign/main.c
+++ b/tools/oesign/main.c
@@ -106,7 +106,7 @@ static const char _usage_sign[] =
     "\n";
 
 static const char _usage_digest[] =
-    "Usage: %s -e ENCLAVE_IMAGE [-c CONFIG_FILE] -d DIGEST_FILE\n"
+    "Usage: %s digest -e ENCLAVE_IMAGE [-c CONFIG_FILE] -d DIGEST_FILE\n"
     "\n"
     "Options:\n"
     "  -e, --enclave-image      path of an enclave image file.\n"


CodeMonkeyLeet:
Per today's [SIG-testing meeting](https://hackmd.io/9Ux-ivlGRvW8lmv5mVovfg#7-July-2020):
* It's not clear that the 3 added tests are actively testing code paths specific to oesign, for example, the handling of keys are primarily delegated to OpenSSL on the host side, and its behaviors drive what we see in related issue #3219. 
* As such, the tests are probably not interesting to add to oesign, unless there's an expectation that we are performing some special error handling for each of the test cases, which isn't reflected in the expected verification of the tests.
* The conclusion here is that this set of additional tests don't seem particularly valuable, and could be dropped entirely or consolidated into a single key format test (e.g. passing DER as opposed to PEM).

On a more general level, we would recommend that contributors approaching testing in a black-box manner to apply techniques such as [Equivalence Class Testing](https://www.professionalqa.com/equivalence-class-testing) to reduce the overall costs of running tests. We would also be happy to engage in test gap analysis as part of the SIG-testing forum to plan more involved test contributions up front so we're not necessarily limited to black box techniques as well.

johnvenn:
> Thanks for finding and fixing this as well!
> 
> It would probably also be helpful to edit the title of this PR to reflect its new contents.
thanks a lot for the recommendations. I have also re-edited the PR, but not sure why it's still the old one. now I changed it.

PR #3228, opened 2020-07-06
Fix typos in Jenkins script that cause libcxx tests fail

mingweishih:


radhikaj:
Thanks @mingweishih

PR #3229, opened 2020-07-06
System edl opt-in code clean up and add tests

mingweishih:
This PR cleans up the code related to system edl opt-in to make it consistent, which include
- Fix typos in the comments
- Use the same pattern for all the weak alias implementations of ecall/ocall

Also, adding a test to lock down the scenario with using the minimal set of system edls. The test invokes the unimported ecalls/ocalls and verifies the return value against `OE_UNSUPPORTED`.


mingweishih:
bors try

PR #3230, opened 2020-07-06
Update Window install script to support the SGX1FLC option on Win2019

### LINKED ISSUE #2960: install-windows-prereqs.ps1 should work on WS2019 with SGX1FLC option
Install-DCAP-dependencies still needs to be updated to be able to manually install the drivers over the existing ones via devcon; Install-DCAP-dependencies for WS2019 is just broken. (i.e. make SGX1FLC option just work on WS2019 instead of requiring SGX1FLC-nodriver option.)
See PR##2911 for manual instructions on updating drivers on WS2019.

### LINKED ISSUE #3010: install-windows-prereqs.ps1 can fail trying to set sgx_lc_msr regkey on WS2019
The install-windows-prereqs.ps1 script can fail with the error:

> Cannot find path 'HKLM:\SYSTEM\CurrentControlSet\Services\sgx_lc_msr\Parameters' because it does not exist.

This is because it always tries to set the sgx_lc_msr regkey assuming that the user may be using the `SGX1FLC-NoDriver` setting, but still wants the enable the default installed FLC driver. However, if the device is not SGX enabled or there WU has failed to setup the sgx_lc_msr regkey before the script is run, there's no regkey to create the value under.

As an independent fix, the script should:
- only attempt to write the regkey if both `($LaunchConfiguration -eq "SGX1FLC-NoDriver") -and (${OS_VERSION} -eq "WinServer2019")`
- probe the presence of the regkey and create it if not found.

This can also be addressed as part of fixing #2960: If the `SGX1FLC` option just works for WS2019 by always upgrading from local packages, then the script should only set the regkey in that scenario, and not try to set it in the `SGX1FLC-NoDriver` configuration.

rs--:
Update PSW and DCAP on Windows 2019 instead of attempting to install.

Fixes #2960 
Fixes #3010 

### DIFF
--- a/scripts/install-windows-prereqs.ps1
+++ b/scripts/install-windows-prereqs.ps1
@@ -16,6 +16,7 @@ Param(
     [string]$Clang7Hash = '672E4C420D6543A8A9F8EC5F1E5F283D88AC2155EF4C57232A399160A02BFF57',
     [string]$IntelPSWURL = 'http://registrationcenter-download.intel.com/akdlm/irc_nas/16607/Intel%20SGX%20PSW%20for%20Windows%20v2.7.101.2.exe',
     [string]$IntelPSWHash = 'AF669A4593411E9AABCE18838C91003866DDDEDAC5BEEC61DE160025008B0A19',
+    [string]$IntelPSWInfHash = '48e7c1e9-6de8-46ee-8ff9-46aa7b7ee5b9',
     [string]$ShellCheckURL = 'https://shellcheck.storage.googleapis.com/shellcheck-v0.7.0.zip',
     [string]$ShellCheckHash = '02CFA14220C8154BB7C97909E80E74D3A7FE2CBB7D80AC32ADCAC7988A95E387',
     [string]$NugetURL = 'https://www.nuget.org/api/v2/package/NuGet.exe/3.4.3',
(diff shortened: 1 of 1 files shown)


scripts/install-windows-prereqs.ps1
```diff
@@ -427,6 +427,30 @@ function Install-PSW {
...
+    }
+    Install-ZipTool -ZipPath $PACKAGES["psw"]["local_file"] `
+                    -InstallDirectory $tempInstallDir
+
+    $signedZip = Get-Item "$tempInstallDir\Intel*SGX*\PSW_INF_RS3_and_above\component\Signed_*.zip"
+    $infDir = "$tempInstallDir\PSW_INF"
+    Install-ZipTool -ZipPath $signedZip `
+                    -InstallDirectory $infDir
+
+    $infHash = "48e7c1e9-6de8-46ee-8ff9-46aa7b7ee5b9"
```
mingweishih, line 443: Would it be better to define this as part of `Param`, which is easier to track? It seems like it requires manually update every time  Intel updates the package.

mingweishih:
You can use the "fixes #issue" pattern in the description (e.g., fixes #issue1, fixes #issue2) . Once this PR is merged, those issues will be automatically closed.

radhikaj:
Thanks @rs-- for these fixes. Some of the docs also need to be updated to be inline with this: As an example. https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/WindowsSGX1FLCGettingStarted.md would need to be updated I think. Are there others?

rs--:
@radhikaj good point, I'll make sure the appropriate docs are updated

scripts/install-windows-prereqs.ps1
```diff
@@ -586,22 +611,39 @@ function Install-DCAP-Dependencies {
...
+                        Write-Output "Removing driver $driverLocation"
+                        Remove-DCAPDriver -Name $driverLocation
                     }
                 }
-                Write-Output "Installing driver $($drivers[${OS_VERSION}][$driver]['location'])"
-                $install = & $devConBinaryPath install "$($inf.FullName)" $drivers[${OS_VERSION}][$driver]['location']
-                if($LASTEXITCODE) {
-                    Throw "Failed to install $driver driver"
+
+                if ($OS_VERSION -eq "WinServer2019")
```
CodeMonkeyLeet, line 627: You should push this check up before the search for existing drivers, which also removes the WS2019 version check in the loop for the driver removal logic. I'd encourage you to minimize the number of branches due to the versioning logic to make updates more manageable. 

I also believe that this is not specific to WinServer2019; this code path probably should apply to Win10 and that should be tested and verified. Win10 should almost certainly not be hitting the else clause and running the Win2016 code, so I would invert the conditional check (i.e. Win2016 is the exceptional case, and Win10/WS2019 should be default/else)

scripts/install-windows-prereqs.ps1
```diff
@@ -16,6 +16,7 @@ Param(
     [string]$Clang7Hash = '672E4C420D6543A8A9F8EC5F1E5F283D88AC2155EF4C57232A399160A02BFF57',
     [string]$IntelPSWURL = 'http://registrationcenter-download.intel.com/akdlm/irc_nas/16607/Intel%20SGX%20PSW%20for%20Windows%20v2.7.101.2.exe',
     [string]$IntelPSWHash = 'AF669A4593411E9AABCE18838C91003866DDDEDAC5BEEC61DE160025008B0A19',
+    [string]$IntelPSWInfHash = '48e7c1e9-6de8-46ee-8ff9-46aa7b7ee5b9',
```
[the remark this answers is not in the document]
rs--: I have a draft PR because of the new directory structure (https://github.com/openenclave/openenclave/pull/3249). This line wouldn't be needed anymore due to that change.

rs--:
Closing this PR since the code added requires modification after Pull #3251 is merged; will re-open with all the modifications, and comments addressed.

PR #3236, opened 2020-07-08
EEID refinements

wintersteiger:
This are changes to align EEID with Icelake's config-id. The complete EEID data (the config) is still passed at enclave creation time, with only its hash being recorded on the (now single) EEID page and the enclave loads the actual config/data into its heap during initialization.

Further, support for EEID enclaves with static memory settings is added, for which the EEID verifier is greatly simplified as it doesn't need to re-measure the data pages.

Both, the config-id and complete config/data are now reported as claims, for downstream processing.

CC @ad-l @sylvanc @bodzhang

### DIFF
--- a/common/sgx/eeid.c
+++ b/common/sgx/eeid.c
@@ -604,7 +604,7 @@ static oe_result_t _hton_uint16(
         return OE_INVALID_PARAMETER;
 
     for (size_t i = 0; i < 2; i++)
-        (*position)[i] = (x >> (32 - ((i + 1) * 8))) & 0xFF;
+        (*position)[i] = (x >> (16 - ((i + 1) * 8))) & 0xFF;
 
     *position += 2;
     *remaining -= 2;
--- a/common/sgx/eeid.c
+++ b/common/sgx/eeid.c
@@ -595,7 +595,7 @@ done:
     return result;
 }
 
-static oe_result_t _hton_uint16_t(
+static oe_result_t _hton_uint16(
     uint16_t x,
     uint8_t** position,
     size_t* remaining)
@@ -612,7 +612,7 @@ static oe_result_t _hton_uint16_t(
     return OE_OK;
 }
 
-static oe_result_t _ntoh_uint16_t(
+static oe_result_t _ntoh_uint16(
     const uint8_t** position,
     size_t* remaining,
     uint16_t* x)
@@ -630,7 +630,7 @@ static oe_result_t _ntoh_uint16_t(
     return OE_OK;
 }
 
-static oe_result_t _hton_uint32_t(
+static oe_result_t _hton_uint32(
     uint32_t x,
     uint8_t** position,
     size_t* remaining)
@@ -647,7 +647,7 @@ static oe_result_t _hton_uint32_t(
     return OE_OK;
 }
 
-static oe_result_t _ntoh_uint32_t(
+static oe_result_t _ntoh_uint32(
     const uint8_t** position,
     size_t* remaining,
     uint32_t* x)
@@ -665,7 +665,7 @@ static oe_result_t _ntoh_uint32_t(
     return OE_OK;
 }
 
-oe_result_t hton_uint64_t(uint64_t x, uint8_t** position, size_t* remaining)
+oe_result_t oe_hton_uint64(uint64_t x, uint8_t** position, size_t* remaining)
 {
     if (*remaining < 8)
         return OE_INVALID_PARAMETER;
@@ -679,7 +679,7 @@ oe_result_t hton_uint64_t(uint64_t x, uint8_t** position, size_t* remaining)
     return OE_OK;
 }
 
-oe_result_t ntoh_uint64_t(
+oe_result_t oe_ntoh_uint64(
     const uint8_t** position,
     size_t* remaining,
     uint64_t* x)
@@ -740,9 +740,9 @@ oe_result_t oe_enclave_size_settings_hton(
     size_t* remaining)
 {
     oe_result_t result = OE_UNEXPECTED;
-    OE_CHECK(hton_uint64_t(sizes->num_heap_pages, position, remaining));
-    OE_CHECK(hton_uint64_t(sizes->num_stack_pages, position, remaining));
-    OE_CHECK(hton_uint64_t(sizes->num_tcs, position, remaining));
+    OE_CHECK(oe_hton_uint64(sizes->num_heap_pages, position, remaining));
+    OE_CHECK(oe_hton_uint64(sizes->num_stack_pages, position, remaining));
+    OE_CHECK(oe_hton_uint64(sizes->num_tcs, position, remaining));
     result = OE_OK;
 done:
     return result;
@@ -754,9 +754,9 @@ oe_result_t oe_enclave_size_settings_ntoh(
     oe_enclave_size_settings_t* sizes)
 {
     oe_result_t result = OE_UNEXPECTED;
-    OE_CHECK(ntoh_uint64_t(position, remaining, &sizes->num_heap_pages));
-    OE_CHECK(ntoh_uint64_t(position, remaining, &sizes->num_stack_pages));
-    OE_CHECK(ntoh_uint64_t(position, remaining, &sizes->num_tcs));
+    OE_CHECK(oe_ntoh_uint64(position, remaining, &sizes->num_heap_pages));
+    OE_CHECK(oe_ntoh_uint64(position, remaining, &sizes->num_stack_pages));
+    OE_CHECK(oe_ntoh_uint64(position, remaining, &sizes->num_tcs));
     result = OE_OK;
 done:
     return result;
@@ -776,24 +776,24 @@ oe_result_t oe_eeid_hton(
     uint8_t* position = buffer;
     size_t remaining = buffer_size;
 
-    OE_CHECK(_hton_uint32_t(eeid->version, &position, &remaining));
+    OE_CHECK(_hton_uint32(eeid->version, &position, &remaining));
 
     for (size_t i = 0; i < 8; i++)
-        OE_CHECK(_hton_uint32_t(eeid->hash_state.H[i], &position, &remaining));
+        OE_CHECK(_hton_uint32(eeid->hash_state.H[i], &position, &remaining));
     for (size_t i = 0; i < 2; i++)
-        OE_CHECK(_hton_uint32_t(eeid->hash_state.N[i], &position, &remaining));
+        OE_CHECK(_hton_uint32(eeid->hash_state.N[i], &position, &remaining));
 
     OE_CHECK(oe_enclave_size_settings_hton(
         &eeid->size_settings, &position, &remaining));
 
     OE_CHECK(_hton_buffer(
         eeid->config_id, sizeof(eeid->config_id), &position, &remaining));
-    OE_CHECK(_hton_uint16_t(eeid->config_svn, &position, &remaining));
+    OE_CHECK(_hton_uint16(eeid->config_svn, &position, &remaining));
 
-    OE_CHECK(hton_uint64_t(eeid->vaddr, &position, &remaining));
-    OE_CHECK(hton_uint64_t(eeid->entry_point, &position, &remaining));
+    OE_CHECK(oe_hton_uint64(eeid->vaddr, &position, &remaining));
+    OE_CHECK(oe_hton_uint64(eeid->entry_point, &position, &remaining));
 
-    OE_CHECK(hton_uint64_t(eeid->signature_size, &position, &remaining));
+    OE_CHECK(oe_hton_uint64(eeid->signature_size, &position, &remaining));
     OE_CHECK(_hton_buffer(
         eeid->signature, eeid->signature_size, &position, &remaining));
 
@@ -815,26 +815,26 @@ oe_result_t oe_eeid_ntoh(
     const uint8_t* position = buffer;
     size_t remaining = buffer_size;
 
-    OE_CHECK(_ntoh_uint32_t(&position, &remaining, &eeid->version));
+    OE_CHECK(_ntoh_uint32(&position, &remaining, &eeid->version));
 
     if (eeid->version != OE_EEID_VERSION)
         OE_RAISE(OE_INVALID_PARAMETER);
 
     for (size_t i = 0; i < 8; i++)
-        OE_CHECK(_ntoh_uint32_t(&position, &remaining, &eeid->hash_state.H[i]));
+        OE_CHECK(_ntoh_uint32(&position, &remaining, &eeid->hash_state.H[i]));
     for (size_t i = 0; i < 2; i++)
-        OE_CHECK(_ntoh_uint32_t(&position, &remaining, &eeid->hash_state.N[i]));
+        OE_CHECK(_ntoh_uint32(&position, &remaining, &eeid->hash_state.N[i]));
 
     oe_enclave_size_settings_ntoh(&position, &remaining, &eeid->size_settings);
 
     OE_CHECK(_ntoh_buffer(
         &position, &remaining, eeid->config_id, sizeof(eeid->config_id)));
-    OE_CHECK(_ntoh_uint16_t(&position, &remaining, &eeid->config_svn));
+    OE_CHECK(_ntoh_uint16(&position, &remaining, &eeid->config_svn));
 
-    OE_CHECK(ntoh_uint64_t(&position, &remaining, &eeid->vaddr));
-    OE_CHECK(ntoh_uint64_t(&position, &remaining, &eeid->entry_point));
+    OE_CHECK(oe_ntoh_uint64(&position, &remaining, &eeid->vaddr));
+    OE_CHECK(oe_ntoh_uint64(&position, &remaining, &eeid->entry_point));
 
-    OE_CHECK(ntoh_uint64_t(&position, &remaining, &eeid->signature_size));
+    OE_CHECK(oe_ntoh_uint64(&position, &remaining, &eeid->signature_size));
     OE_CHECK(_ntoh_buffer(
         &position, &remaining, eeid->signature, eeid->signature_size));
 
@@ -856,10 +856,11 @@ oe_result_t oe_eeid_evidence_hton(
     uint8_t* position = buffer;
     size_t remaining = buffer_size;
 
-    OE_CHECK(hton_uint64_t(evidence->sgx_evidence_size, &position, &remaining));
     OE_CHECK(
-        hton_uint64_t(evidence->sgx_endorsements_size, &position, &remaining));
-    OE_CHECK(hton_uint64_t(evidence->eeid_size, &position, &remaining));
+        oe_hton_uint64(evidence->sgx_evidence_size, &position, &remaining));
+    OE_CHECK(
+        oe_hton_uint64(evidence->sgx_endorsements_size, &position, &remaining));
+    OE_CHECK(oe_hton_uint64(evidence->eeid_size, &position, &remaining));
 
     size_t data_size = evidence->sgx_evidence_size +
                        evidence->sgx_endorsements_size + evidence->eeid_size;
@@ -885,10 +886,10 @@ oe_result_t oe_eeid_evidence_ntoh(
     size_t remaining = buffer_size;
 
     OE_CHECK(
-        ntoh_uint64_t(&position, &remaining, &evidence->sgx_evidence_size));
-    OE_CHECK(
-        ntoh_uint64_t(&position, &remaining, &evidence->sgx_endorsements_size));
-    OE_CHECK(ntoh_uint64_t(&position, &remaining, &evidence->eeid_size));
+        oe_ntoh_uint64(&position, &remaining, &evidence->sgx_evidence_size));
+    OE_CHECK(oe_ntoh_uint64(
+        &position, &remaining, &evidence->sgx_endorsements_size));
+    OE_CHECK(oe_ntoh_uint64(&position, &remaining, &evidence->eeid_size));
 
     size_t data_size = evidence->sgx_evidence_size +
                        evidence->sgx_endorsements_size + evidence->eeid_size;
--- a/common/sgx/eeid.c
+++ b/common/sgx/eeid.c
@@ -701,6 +701,34 @@ size_t oe_eeid_byte_size(const oe_eeid_t* eeid)
            eeid->signature_size;
 }
 
+oe_result_t oe_enclave_size_settings_hton(
+    const oe_enclave_size_settings_t* sizes,
+    uint8_t** position,
+    size_t* remaining)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    OE_CHECK(hton_uint64_t(sizes->num_heap_pages, position, remaining));
+    OE_CHECK(hton_uint64_t(sizes->num_stack_pages, position, remaining));
+    OE_CHECK(hton_uint64_t(sizes->num_tcs, position, remaining));
+    result = OE_OK;
+done:
+    return result;
+}
+
+oe_result_t oe_enclave_size_settings_ntoh(
+    const uint8_t** position,
+    size_t* remaining,
+    oe_enclave_size_settings_t* sizes)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    OE_CHECK(ntoh_uint64_t(position, remaining, &sizes->num_heap_pages));
+    OE_CHECK(ntoh_uint64_t(position, remaining, &sizes->num_stack_pages));
+    OE_CHECK(ntoh_uint64_t(position, remaining, &sizes->num_tcs));
+    result = OE_OK;
+done:
+    return result;
+}
+
 oe_result_t oe_eeid_hton(
     const oe_eeid_t* eeid,
     uint8_t* buffer,
@@ -722,11 +750,8 @@ oe_result_t oe_eeid_hton(
     for (size_t i = 0; i < 2; i++)
         OE_CHECK(_hton_uint32_t(eeid->hash_state.N[i], &position, &remaining));
 
-    OE_CHECK(hton_uint64_t(
-        eeid->size_settings.num_heap_pages, &position, &remaining));
-    OE_CHECK(hton_uint64_t(
-        eeid->size_settings.num_stack_pages, &position, &remaining));
-    OE_CHECK(hton_uint64_t(eeid->size_settings.num_tcs, &position, &remaining));
+    OE_CHECK(oe_enclave_size_settings_hton(
+        &eeid->size_settings, &position, &remaining));
 
     OE_CHECK(hton_uint64_t(eeid->vaddr, &position, &remaining));
     OE_CHECK(hton_uint64_t(eeid->entry_point, &position, &remaining));
@@ -763,12 +788,7 @@ oe_result_t oe_eeid_ntoh(
     for (size_t i = 0; i < 2; i++)
         OE_CHECK(_ntoh_uint32_t(&position, &remaining, &eeid->hash_state.N[i]));
 
-    OE_CHECK(ntoh_uint64_t(
-        &position, &remaining, &eeid->size_settings.num_heap_pages));
-    OE_CHECK(ntoh_uint64_t(
-        &position, &remaining, &eeid->size_settings.num_stack_pages));
-    OE_CHECK(
-        ntoh_uint64_t(&position, &remaining, &eeid->size_settings.num_tcs));
+    oe_enclave_size_settings_ntoh(&position, &remaining, &eeid->size_settings);
 
     OE_CHECK(ntoh_uint64_t(&position, &remaining, &eeid->vaddr));
     OE_CHECK(ntoh_uint64_t(&position, &remaining, &eeid->entry_point));
--- a/common/sgx/eeid.c
+++ b/common/sgx/eeid.c
@@ -5,7 +5,6 @@
 #include <string.h>
 
 #include <openenclave/bits/attestation.h>
-#include <openenclave/bits/eeid.h>
 #include <openenclave/bits/sgx/sgxtypes.h>
 #include <openenclave/internal/crypto/sha.h>
 #include <openenclave/internal/eeid.h>
--- a/common/sgx/eeid.c
+++ b/common/sgx/eeid.c
@@ -294,7 +294,7 @@ done:
     return result;
 }
 
-static oe_result_t _remeasure_eeid_page(
+static oe_result_t _measure_eeid_page(
     oe_sha256_context_t* hctx,
     uint64_t base,
     uint64_t* vaddr,
@@ -314,7 +314,7 @@ static oe_result_t _remeasure_eeid_page(
     memcpy(&eeid_page->eeid, eeid, sizeof(oe_eeid_t) + eeid->signature_size);
     memcpy(eeid_page->config_id.buf, config_id->buf, sizeof(config_id->buf));
 
-    OE_CHECK(_add_page(hctx, base, eeid_page, vaddr, true, true));
+    OE_CHECK(_measure_page(hctx, base, eeid_page, vaddr, true, true));
     free(eeid_page);
 
     result = OE_OK;
@@ -327,7 +327,8 @@ oe_result_t oe_remeasure_memory_pages(
     const oe_eeid_t* eeid,
     const OE_SHA256* config_id,
     OE_SHA256* computed_enclave_hash,
-    bool with_eeid_page)
+    bool with_eeid_page,
+    bool static_sizes)
 {
     oe_result_t result;
     oe_sha256_context_t hctx;
@@ -341,58 +342,64 @@ oe_result_t oe_remeasure_memory_pages(
     uint64_t vaddr = eeid->vaddr;
 
     if (with_eeid_page)
-        OE_CHECK(_remeasure_eeid_page(&hctx, base, &vaddr, eeid, config_id));
+        OE_CHECK(_measure_eeid_page(&hctx, base, &vaddr, eeid, config_id));
     else
-        OE_CHECK(_add_page(&hctx, base, blank_pg.data, &vaddr, true, true));
+        OE_CHECK(_measure_page(&hctx, base, blank_pg.data, &vaddr, true, true));
 
-    // This is where we replay the addition of memory pages, both, for
-    // verification of the extended image hash (with_eeid_pages=true) and
-    // the base image hash, for which there are no EEID pages, but one TCS
-    // page.
-
-    for (size_t i = 0; i < eeid->size_settings.num_heap_pages; i++)
-        OE_CHECK(_measure_page(&hctx, base, &blank_pg, &vaddr, false, false));
-
-    for (size_t i = 0; i < eeid->size_settings.num_tcs; i++)
+    if (!static_sizes)
     {
-        vaddr += OE_PAGE_SIZE; /* guard page */
+        // This is where we replay the addition of memory pages, both, for
+        // verification of the extended image hash (with_eeid_pages=true) and
+        // the base image hash, for which there are no EEID pages, but one TCS
+        // page. This is only required when using dynamic memory settings, as
+        // the hash state is saved after the data pages in the case of static
+        // sizes.
 
-        for (size_t i = 0; i < eeid->size_settings.num_stack_pages; i++)
+        for (size_t i = 0; i < eeid->size_settings.num_heap_pages; i++)
             OE_CHECK(
-                _measure_page(&hctx, base, &stack_pg, &vaddr, true, false));
+                _measure_page(&hctx, base, &blank_pg, &vaddr, false, false));
 
-        vaddr += OE_PAGE_SIZE; /* guard page */
+        for (size_t i = 0; i < eeid->size_settings.num_tcs; i++)
+        {
+            vaddr += OE_PAGE_SIZE; /* guard page */
 
-        sgx_tcs_t* tcs;
-        memset(&tcs_pg, 0, sizeof(tcs_pg));
-        tcs = (sgx_tcs_t*)&tcs_pg;
-        tcs->flags = 0;
-        tcs->ossa = vaddr + OE_PAGE_SIZE;
-        tcs->cssa = 0;
-        tcs->nssa = 2;
-        tcs->oentry = eeid->entry_point;
-        tcs->fsbase = vaddr + (5 * OE_PAGE_SIZE);
-        tcs->gsbase = tcs->fsbase;
-        tcs->fslimit = 0xFFFFFFFF;
-        tcs->gslimit = 0xFFFFFFFF;
+            for (size_t i = 0; i < eeid->size_settings.num_stack_pages; i++)
+                OE_CHECK(
+                    _measure_page(&hctx, base, &stack_pg, &vaddr, true, false));
 
-        OE_CHECK(oe_sgx_measure_load_enclave_data(
-            &hctx,
-            base,
-            base + vaddr,
-            (uint64_t)&tcs_pg,
-            SGX_SECINFO_TCS,
-            true));
+            vaddr += OE_PAGE_SIZE; /* guard page */
 
-        vaddr += OE_PAGE_SIZE;
+            sgx_tcs_t* tcs;
+            memset(&tcs_pg, 0, sizeof(tcs_pg));
+            tcs = (sgx_tcs_t*)&tcs_pg;
+            tcs->flags = 0;
+            tcs->ossa = vaddr + OE_PAGE_SIZE;
+            tcs->cssa = 0;
+            tcs->nssa = 2;
+            tcs->oentry = eeid->entry_point;
+            tcs->fsbase = vaddr + (5 * OE_PAGE_SIZE);
+            tcs->gsbase = tcs->fsbase;
+            tcs->fslimit = 0xFFFFFFFF;
+            tcs->gslimit = 0xFFFFFFFF;
 
-        for (size_t i = 0; i < 2; i++)
-            _measure_page(&hctx, base, &blank_pg, &vaddr, true, false);
+            OE_CHECK(oe_sgx_measure_load_enclave_data(
+                &hctx,
+                base,
+                base + vaddr,
+                (uint64_t)&tcs_pg,
+                SGX_SECINFO_TCS,
+                true));
 
-        vaddr += OE_PAGE_SIZE; /* guard page */
+            vaddr += OE_PAGE_SIZE;
 
-        for (size_t i = 0; i < 2; i++)
-            _measure_page(&hctx, base, &blank_pg, &vaddr, true, false);
+            for (size_t i = 0; i < 2; i++)
+                _measure_page(&hctx, base, &blank_pg, &vaddr, true, false);
+
+            vaddr += OE_PAGE_SIZE; /* guard page */
+
+            for (size_t i = 0; i < 2; i++)
+                _measure_page(&hctx, base, &blank_pg, &vaddr, true, false);
+        }
     }
 
     oe_sha256_final(&hctx, computed_enclave_hash);
@@ -531,9 +538,12 @@ oe_result_t verify_eeid(
     if (eeid->signature_size != 1808) // We only support SGX sigstructs for now.
         OE_RAISE(OE_UNSUPPORTED);
 
+    bool static_sizes = !is_eeid_base_image(base_image_sizes);
+
     // Compute expected enclave hash
     OE_SHA256 computed_enclave_hash;
-    oe_remeasure_memory_pages(eeid, config_id, &computed_enclave_hash, true);
+    oe_remeasure_memory_pages(
+        eeid, config_id, &computed_enclave_hash, true, static_sizes);
 
     // Check recomputed enclave hash against reported enclave hash
     if (memcmp(
@@ -555,7 +565,7 @@ oe_result_t verify_eeid(
     tmp_eeid.size_settings = *base_image_sizes;
 
     oe_remeasure_memory_pages(
-        &tmp_eeid, NULL, &computed_base_enclave_hash, false);
+        &tmp_eeid, NULL, &computed_base_enclave_hash, false, static_sizes);
 
     if (memcmp(
             computed_base_enclave_hash.buf,
--- a/common/sgx/eeid.c
+++ b/common/sgx/eeid.c
@@ -522,7 +522,7 @@ oe_result_t verify_eeid(
     uint16_t reported_product_id,
     uint32_t reported_security_version,
     uint64_t reported_attributes,
-    const sgx_attributes_t* r_sgx_attributes,
+    const sgx_attributes_t* reported_sgx_attributes,
     uint32_t reported_misc_select,
     const uint8_t** base_enclave_hash,
     const oe_eeid_t* eeid,
@@ -574,12 +574,23 @@ oe_result_t verify_eeid(
     // Check other image properties have not changed
     bool base_debug = sigstruct->attributes.flags & SGX_FLAGS_DEBUG;
     bool extended_debug = reported_attributes & OE_REPORT_ATTRIBUTES_DEBUG;
-    bool sgx_debug = r_sgx_attributes->flags & SGX_FLAGS_DEBUG;
+    bool sgx_debug = reported_sgx_attributes->flags & SGX_FLAGS_DEBUG;
+
+    uint64_t attributes_masked =
+        reported_sgx_attributes->flags & ~sigstruct->attributemask.flags;
+    uint64_t base_attributes_masked =
+        sigstruct->attributes.flags & ~sigstruct->attributemask.flags;
+    bool attributes_match = attributes_masked == base_attributes_masked;
+
+    uint64_t miscselect_masked = reported_misc_select & ~sigstruct->miscmask;
+    uint64_t base_miscselect_masked =
+        sigstruct->miscselect & ~sigstruct->miscmask;
+    bool miscselect_match = miscselect_masked == base_miscselect_masked;
 
     if (base_debug != extended_debug || base_debug != sgx_debug ||
-        sigstruct->miscselect != reported_misc_select ||
         sigstruct->isvprodid != reported_product_id ||
-        sigstruct->isvsvn != reported_security_version)
+        sigstruct->isvsvn != reported_security_version || !attributes_match ||
+        !miscselect_match)
         OE_RAISE(OE_VERIFY_FAILED);
 
     // Check old signature (new signature has been checked above)
--- a/common/sgx/eeid.c
+++ b/common/sgx/eeid.c
@@ -27,12 +27,6 @@
 #include "../../host/crypto/openssl/rsa.h"
 #endif
 
-int is_eeid_base_image(const oe_enclave_size_settings_t* sizes)
-{
-    return sizes->num_heap_pages == 0 && sizes->num_stack_pages == 0 &&
-           sizes->num_tcs == 1;
-}
-
 oe_result_t oe_create_eeid_sgx(oe_eeid_t** eeid)
 {
     oe_result_t result = OE_UNEXPECTED;
--- a/common/sgx/eeid.c
+++ b/common/sgx/eeid.c
@@ -577,14 +577,17 @@ oe_result_t verify_eeid(
     bool sgx_debug = reported_sgx_attributes->flags & SGX_FLAGS_DEBUG;
 
     uint64_t attributes_masked =
-        reported_sgx_attributes->flags & ~sigstruct->attributemask.flags;
+        reported_sgx_attributes->flags & sigstruct->attributemask.flags;
     uint64_t base_attributes_masked =
-        sigstruct->attributes.flags & ~sigstruct->attributemask.flags;
+        sigstruct->attributes.flags & sigstruct->attributemask.flags;
+    /*The base image sigstruct is missing the SGX_FLAGS_INITTED flag because it
+     * was never created or initialized */
+    base_attributes_masked |= SGX_FLAGS_INITTED;
     bool attributes_match = attributes_masked == base_attributes_masked;
 
-    uint64_t miscselect_masked = reported_misc_select & ~sigstruct->miscmask;
+    uint64_t miscselect_masked = reported_misc_select & sigstruct->miscmask;
     uint64_t base_miscselect_masked =
-        sigstruct->miscselect & ~sigstruct->miscmask;
+        sigstruct->miscselect & sigstruct->miscmask;
     bool miscselect_match = miscselect_masked == base_miscselect_masked;
 
     if (base_debug != extended_debug || base_debug != sgx_debug ||
--- a/common/sgx/eeid.c
+++ b/common/sgx/eeid.c
@@ -525,6 +525,8 @@ oe_result_t verify_eeid(
     uint16_t reported_product_id,
     uint32_t reported_security_version,
     uint64_t reported_attributes,
+    const sgx_attributes_t* r_sgx_attributes,
+    uint32_t reported_misc_select,
     const uint8_t** base_enclave_hash,
     const oe_eeid_t* eeid,
     const OE_SHA256* config_id,
@@ -576,8 +578,10 @@ oe_result_t verify_eeid(
     // Check other image properties have not changed
     bool base_debug = sigstruct->attributes.flags & SGX_FLAGS_DEBUG;
     bool extended_debug = reported_attributes & OE_REPORT_ATTRIBUTES_DEBUG;
+    bool sgx_debug = r_sgx_attributes->flags & SGX_FLAGS_DEBUG;
 
-    if (base_debug != extended_debug ||
+    if (base_debug != extended_debug || base_debug != sgx_debug ||
+        sigstruct->miscselect != reported_misc_select ||
         sigstruct->isvprodid != reported_product_id ||
         sigstruct->isvsvn != reported_security_version)
         OE_RAISE(OE_VERIFY_FAILED);
--- a/common/sgx/eeid.c
+++ b/common/sgx/eeid.c
@@ -597,10 +597,9 @@ oe_result_t verify_eeid(
         sigstruct->miscselect & sigstruct->miscmask;
     bool miscselect_match = miscselect_masked == base_miscselect_masked;
 
-    if (debug != oe_debug || !oe_remote ||
-        sigstruct->isvprodid != reported_product_id ||
-        sigstruct->isvsvn != reported_security_version || !flags_match ||
-        !xfrm_match || !miscselect_match)
+    if (debug != oe_debug || !oe_remote || !flags_match || !xfrm_match ||
+        !miscselect_match || sigstruct->isvprodid != reported_product_id ||
+        sigstruct->isvsvn != reported_security_version)
         OE_RAISE(OE_VERIFY_FAILED);
 
     // Check old signature (new signature has been checked above)
--- a/common/sgx/eeid.c
+++ b/common/sgx/eeid.c
@@ -521,7 +521,7 @@ oe_result_t verify_eeid(
     const uint8_t* reported_enclave_signer,
     uint16_t reported_product_id,
     uint32_t reported_security_version,
-    uint64_t reported_attributes,
+    uint64_t reported_oe_attributes,
     const sgx_attributes_t* reported_sgx_attributes,
     uint32_t reported_misc_select,
     const uint8_t** base_enclave_hash,
@@ -572,32 +572,39 @@ oe_result_t verify_eeid(
         OE_RAISE(OE_VERIFY_FAILED);
 
     // Check other image properties have not changed
-    bool base_debug = sigstruct->attributes.flags & SGX_FLAGS_DEBUG;
-    bool extended_debug = reported_attributes & OE_REPORT_ATTRIBUTES_DEBUG;
-    bool sgx_debug = reported_sgx_attributes->flags & SGX_FLAGS_DEBUG;
+    bool debug = reported_sgx_attributes->flags & SGX_FLAGS_DEBUG;
+    bool oe_debug = reported_oe_attributes & OE_REPORT_ATTRIBUTES_DEBUG;
+    bool oe_remote = reported_oe_attributes & OE_REPORT_ATTRIBUTES_REMOTE;
 
-    uint64_t attributes_masked =
+    uint64_t flags_masked =
         reported_sgx_attributes->flags & sigstruct->attributemask.flags;
-    uint64_t base_attributes_masked =
+    uint64_t base_flags_masked =
         sigstruct->attributes.flags & sigstruct->attributemask.flags;
+
     /*The base image sigstruct is missing the SGX_FLAGS_INITTED flag because it
      * was never created or initialized */
-    base_attributes_masked |= SGX_FLAGS_INITTED;
-    bool attributes_match = attributes_masked == base_attributes_masked;
+    base_flags_masked |= SGX_FLAGS_INITTED;
+    bool flags_match = flags_masked == base_flags_masked;
+
+    uint64_t xfrm_masked =
+        reported_sgx_attributes->xfrm & sigstruct->attributemask.xfrm;
+    uint64_t base_xfrm_masked =
+        sigstruct->attributes.xfrm & sigstruct->attributemask.xfrm;
+    bool xfrm_match = xfrm_masked == base_xfrm_masked;
 
     uint64_t miscselect_masked = reported_misc_select & sigstruct->miscmask;
     uint64_t base_miscselect_masked =
         sigstruct->miscselect & sigstruct->miscmask;
     bool miscselect_match = miscselect_masked == base_miscselect_masked;
 
-    if (base_debug != extended_debug || base_debug != sgx_debug ||
+    if (debug != oe_debug || !oe_remote ||
         sigstruct->isvprodid != reported_product_id ||
-        sigstruct->isvsvn != reported_security_version || !attributes_match ||
-        !miscselect_match)
+        sigstruct->isvsvn != reported_security_version || !flags_match ||
+        !xfrm_match || !miscselect_match)
         OE_RAISE(OE_VERIFY_FAILED);
 
     // Check old signature (new signature has been checked above)
-    if (base_debug && is_zero(sigstruct->signature, OE_KEY_SIZE))
+    if (debug && is_zero(sigstruct->signature, OE_KEY_SIZE))
         return OE_OK; // Unsigned debug image is ok.
     else
         OE_CHECK(verify_base_image_signature(sigstruct));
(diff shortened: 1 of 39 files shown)


enclave/core/sgx/calls.c
```diff
@@ -135,12 +136,58 @@ extern bool oe_disable_debug_malloc_check;
...
 
+#ifdef OE_WITH_EXPERIMENTAL_EEID
+extern oe_eeid_t* oe_eeid;
+extern const oe_enclave_config_t* oe_enclave_config;
+
+static oe_result_t _load_config(const oe_enclave_config_t* config)
+{
+    if (config && config->data && config->size)
+    {
+        if (oe_is_within_enclave(config->data, config->size) ||
```
anakrish, line 147: This check needs to be done on the enclave copy of config. Otherwise the host could exploit the TOCTOU vulnerability.

enclave/core/sgx/init.c
```diff
@@ -44,30 +44,16 @@ static void _check_memory_boundaries(void)
...
-        oe_eeid_marker_t* marker = (oe_eeid_marker_t*)heap_base;
-        oe_eeid = (oe_eeid_t*)(enclave_base + marker->offset);
-
-        // Wipe the marker page
-        memset(heap_base, 0, OE_PAGE_SIZE);
-    }
-
-    return r;
+    /* Get the last page before the heap. */
+    uint8_t* heap_base = (uint8_t*)__oe_get_heap_base();
```
anakrish, line 50: It would be better to update the heap pointers to exclude the eeid page. Otherwise, the last page might get allocated via malloc, wiping out the eeid.

enclave/core/sgx/calls.c
wintersteiger: Both are checked. This line of code checks that the incoming config is outside of the enclave (it aborts if within), a few lines below there's a check on the copy.

enclave/core/sgx/init.c
wintersteiger: Oh yes, that's there but hard to see! There is an extra page just before the heap starts, that's why in the next line I take `heap_base - OE_PAGE_SIZE`. That's a bit of a hack though, I should look for a cleaner solution to this, e.g. patching an rva into the image.

enclave/core/sgx/calls.c
anakrish: > Both are checked. This line of code checks that the incoming config is outside of the enclave (it aborts if within), a few lines below there's a check on the copy.

The issue is that after line `147` that checks `config->data` before the control reaches line `159`, the host can change the value of `config->data`. So the `memcpy(copy->data, config->data, config->size)` could be manipulated to copy data from `config->data` that has been changed to point to within the enclave.

anakrish: Note that `oe_is_within_enclave` is not the same thing as `!oe_is_outside_enclave`. If a part of `config->data` lies within the enclave while the rest lies outside, then `oe_is_within_enclave` will return false.

enclave/core/sgx/calls.c
```diff
@@ -135,12 +136,58 @@ extern bool oe_disable_debug_malloc_check;
...
+        oe_enclave_config_t* copy = oe_malloc(sizeof(oe_enclave_config_t));
+        if (!copy)
+            return OE_OUT_OF_MEMORY;
+        copy->size = config->size;
+        copy->data = oe_malloc(config->size);
+        if (!copy->data)
+            return OE_OUT_OF_MEMORY;
+        memcpy(copy->data, config->data, config->size);
+
+        if (!oe_is_within_enclave(copy->data, copy->size))
```
anakrish, line 161: `oe_is_within_enclave(copy->data, copy->size)` will always be true since `copy->data` is initialized via a malloc within the enclave `copy->data = oe_malloc(config->size)`

enclave/core/sgx/calls.c
```diff
@@ -135,12 +136,58 @@ extern bool oe_disable_debug_malloc_check;
...
+
+        /* Copy config into enclave heap */
+        oe_enclave_config_t* copy = oe_malloc(sizeof(oe_enclave_config_t));
+        if (!copy)
+            return OE_OUT_OF_MEMORY;
+        copy->size = config->size;
+        copy->data = oe_malloc(config->size);
+        if (!copy->data)
+            return OE_OUT_OF_MEMORY;
+        memcpy(copy->data, config->data, config->size);
```
anakrish, line 159: This is the problematic line since `config` is under host control.

mingweishih: This would allow host to control both `config->data` and `config->size` so that he can easily overwrite copy->data and trigger heap overflow.

enclave/core/sgx/calls.c
wintersteiger: Yes, I see what you mean now. I will remove this check as it's really not necessary at all. I don't think there's a security problem with the data already being in the enclave address space.

enclave/core/sgx/calls.c
wintersteiger: The host is in full control of `config` in all situations and at all times anyways, I don't think there's a problem with that here. I'll make a copy of `config` before running any checks, but the host can still pass in a `config->size` that's larger then the enclave memory space, but they might as well just refuse to load the enclave in the first place.

wintersteiger: I rewrote that bit now to take a copy of the host-provided values, so the size can't change anymore, but of course the host is still in control of the contents of `config->data`.

enclave/core/sgx/init.c
wintersteiger: (Done0

host/sgx/loadelf.c
```diff
@@ -692,8 +692,11 @@ static oe_result_t _patch(oe_enclave_image_t* image, size_t enclave_end)
     OE_CHECK(
         _set_uint64_t_symbol_value(image, "_reloc_size", image->reloc_size));
 
-    /* heap right after image */
-    oeprops->image_info.heap_rva = image->image_size + image->reloc_size;
+    size_t eeid_size = OE_PAGE_SIZE;
```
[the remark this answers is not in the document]
wintersteiger: This needs to be under the experimental macro for now, I fixed that slightly differently by introducing `_eeid_rva` to track the address of the EEID page.

host/sgx/create.c
```diff
@@ -305,6 +305,9 @@ static oe_result_t _calculate_enclave_size(
         *enclave_size = 0;
     *enclave_end = 0;
 
+    /* EEID takes one page */
+    size_t eeid_size = OE_PAGE_SIZE;
```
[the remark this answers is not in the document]
wintersteiger: (Same as above)

host/sgx/create.c
```diff
@@ -675,38 +713,34 @@ static oe_result_t _eeid_resign(
...
-
-            OE_CHECK(oe_sgx_load_enclave_data(
-                context, enclave_addr, addr, src, flags, true));
-            *vaddr += OE_PAGE_SIZE;
-        }
-
-        oe_memalign_free(pages);
+        /* Add data pages (dynamic memory settings) */
+        /* Save hash state before data pages. */
+        OE_CHECK(_add_eeid_page(context, enclave, entry, props, vaddr));
```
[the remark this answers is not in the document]
wintersteiger: This is split into two parts. For dynamic memory settings, the hash is taken before the memory pages are added and for static sizes, it's taken after adding them. It doesn't help to move the measurement of the EEID page until after the data pages in the case of dynamic settings, it just complicates the logic.

host/sgx/create.c
```diff
@@ -823,20 +857,13 @@ oe_result_t oe_sgx_build_enclave(
...
-        context, enclave, image_size, oeimage.entry_rva, &props, &vaddr));
-#endif
-
+    /* Add data and EEID pages */
+    OE_CHECK(_add_eeid_data_pages(
+        context, enclave, &props, oeimage.entry_rva, &vaddr));
+#else
     /* Add data pages */
     OE_CHECK(
         _add_data_pages(context, enclave, &props, oeimage.entry_rva, &vaddr));
```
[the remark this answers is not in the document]
wintersteiger: I'm not sure I understand this comment. This is already implemented in `_add_data_pages()` (see `*vaddr += OE_PAGE_SIZE;`).

host/sgx/create.c
```diff
@@ -588,58 +621,63 @@ oe_result_t oe_sgx_validate_enclave_properties(
...
+    else
+    {
+        /* Add an empty/guard page */
+    }
 
-        OE_CHECK(oe_sgx_load_enclave_data(
-            context, enclave->addr, addr, src, flags, false));
-        (*vaddr) += OE_PAGE_SIZE;
-        free(page);
+    uint64_t addr = enclave->addr + *vaddr;
```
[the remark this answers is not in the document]
wintersteiger: What do you mean by "committed"? It currently adds a measured page of all zeroes to non-EEID enclaves and the code relies on that to distinguish between non-EEID (0) and EEID (non-0).

enclave/core/sgx/init.c
```diff
@@ -80,10 +70,7 @@ static void _initialize_enclave_image()
     }
 
 #ifdef OE_WITH_EXPERIMENTAL_EEID
-    if (_eeid_patch_memory() != OE_OK)
-    {
-        oe_abort();
```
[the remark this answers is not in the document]
wintersteiger: That "flag" is `eeid_page->eeid.version == 0`. It's 0 for non-eeid enclaves, otherwise the version of the EEID datastructures.

enclave/core/sgx/calls.c
```diff
@@ -135,12 +136,57 @@ extern bool oe_disable_debug_malloc_check;
...
+    oe_enclave_config_t config = *config_from_host;
+
+    if (!config.data || !config.size)
+        return OE_INVALID_PARAMETER;
+
+    /* Copy config into enclave heap */
+    oe_enclave_config_t* copy = oe_malloc(sizeof(oe_enclave_config_t));
+    if (!copy)
+        return OE_OUT_OF_MEMORY;
+    copy->size = config.size;
```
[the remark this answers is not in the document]
wintersteiger: I removed that check a while ago, I think you may not have had the latest version of the code when you did your review.

enclave/core/sgx/calls.c
```diff
@@ -135,12 +136,57 @@ extern bool oe_disable_debug_malloc_check;
...
+    copy->data = oe_malloc(config.size);
+    if (!copy->data)
+        return OE_OUT_OF_MEMORY;
+    memcpy(copy->data, config.data, config.size);
+
+    oe_enclave_config = copy;
+    return OE_OK;
```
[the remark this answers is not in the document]
wintersteiger: Right, those are allowed now.

enclave/core/sgx/calls.c
```diff
@@ -135,12 +136,57 @@ extern bool oe_disable_debug_malloc_check;
...
+        return OE_INVALID_PARAMETER;
+
+    /* Copy config into enclave heap */
+    oe_enclave_config_t* copy = oe_malloc(sizeof(oe_enclave_config_t));
+    if (!copy)
+        return OE_OUT_OF_MEMORY;
+    copy->size = config.size;
+    copy->data = oe_malloc(config.size);
+    if (!copy->data)
+        return OE_OUT_OF_MEMORY;
```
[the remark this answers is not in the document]
wintersteiger: Yes, this i bit tricky, it seems the crypto code isn't linked to this but if code, so I can't easily hash the config. Is there an actual need to check this at this particular point in the code or is it ok to do this elsewhere?

enclave/sgx/eeid_attester.c
```diff
@@ -142,17 +162,37 @@ static oe_result_t _eeid_get_evidence(
...
         *endorsements_buffer_size = 0;
 
-    if (!eeid || eeid->data_size == 0 ||
-        eeid->signature_size != sizeof(sgx_sigstruct_t))
+    if (eeid->signature_size != sizeof(sgx_sigstruct_t))
         OE_RAISE(OE_FAILURE);
 
     size_t eeid_size = oe_eeid_byte_size(eeid);
 
+    sgx_custom_claims_size = custom_claims_size + 3;
```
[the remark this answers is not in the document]
wintersteiger: They are "reflected" but not explicit. I check the explicit values later to determine whether we're in a static or dynamic memory size scenario. Note that these claims are internal to the EEID attester/verifier, they are not in the user-facing claims (but we can add them there as well if we want to).

enclave/sgx/eeid_attester.c
```diff
@@ -194,12 +247,13 @@ static oe_result_t _eeid_get_evidence(
         evidence, *evidence_buffer, *evidence_buffer_size));
 
     // Write endorsements
-    if (endorsements_buffer)
+    if (endorsements_buffer && config && config->data && config->size)
```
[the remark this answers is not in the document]
wintersteiger: We want the full config in the endorsements for our use-cases. If the user doesn't need it, they can simply not pass an `endorsements_buffer`, or just discard it.

host/sgx/create.c
```diff
@@ -1004,8 +1033,16 @@ oe_result_t oe_create_enclave(
...
     enclave->num_ocalls = ocall_count;
 
-    /* Invoke enclave initialization. */
+/* Invoke enclave initialization. */
+#ifdef OE_WITH_EXPERIMENTAL_EEID
+    oe_enclave_with_config_t enclave_with_config;
+    enclave_with_config.enclave = enclave;
+    enclave_with_config.config.data = context.config.data;
+    enclave_with_config.config.size = context.config.size;
+    OE_CHECK(_initialize_enclave_with_config(&enclave_with_config));
```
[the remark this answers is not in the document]
wintersteiger: Adding a new ECALL makes the code more complex and harder to use, so for now I haven't added one. Since it would change the OE API/ECALL interface, this may also require a new design document and a public discussion.

enclave/core/sgx/calls.c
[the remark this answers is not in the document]
wintersteiger: Found a way, done!

enclave/sgx/eeid_attester.c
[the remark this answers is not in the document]
wintersteiger: Due to the recent changes to custom claims handling I had to change this around a bit, it now includes a properly serialized `oe_enclave_size_settings_t` as a custom claim.

host/sgx/create.c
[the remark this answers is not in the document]
bodzhang: "not committed" means the linear address does not map to any EPC page, and access to the linear address will cause page fault. See https://github.com/bodzhang/openenclave/blob/EEIDDesignChange/docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md for detail. 
For a regualr enclave, the implementaiton shoudl not add an all-0 page, instead, the _add_eeid_page(..) operation shoudl be skipped.

host/sgx/create.c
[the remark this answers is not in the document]
bodzhang: For the "static size" EEID enclave, adding the EEID page last, after the base enclave measurement is done, makes the design much easier to understand.

enclave/core/sgx/calls.c
```diff
@@ -135,12 +136,65 @@ extern bool oe_disable_debug_malloc_check;
 **==============================================================================
 */
 
+#ifdef OE_WITH_EXPERIMENTAL_EEID
+extern const oe_enclave_config_t* oe_enclave_config;
+
+static oe_result_t _load_config(const oe_enclave_config_t* config_from_host)
+{
+    const oe_eeid_page_t* eeid_page = __oe_get_eeid_page();
+    const oe_eeid_t* eeid = &eeid_page->eeid;
```
CodeMonkeyLeet, line 145: `__oe_get_eeid_page()` can return NULL

enclave/core/sgx/calls.c
```diff
@@ -135,12 +136,65 @@ extern bool oe_disable_debug_malloc_check;
...
+    oe_enclave_config_t config = *config_from_host;
+
+    if (!config.data || !config.size)
+        return OE_OK; /* no or zero-size configs are ok */
+
+    /* Check that the config hashes to the correct value */
+    OE_SHA256 config_hash;
+    oe_sha256(config.data, config.size, &config_hash);
+    if (memcmp(config_hash.buf, eeid_page->config_id.buf, OE_SHA256_SIZE) != 0)
+        return OE_INVALID_PARAMETER;
```
CodeMonkeyLeet, line 163: Consider a more descriptive error like `OE_VERIFY_FAILED`

enclave/core/sgx/globals.c
```diff
@@ -189,9 +190,29 @@ size_t __oe_get_reloc_size()
...
+    if (!_eeid_rva)
+        return NULL;
+
+    const oe_eeid_page_t* eeid_page = __oe_get_eeid_page();
+    return eeid_page && eeid_page->eeid.version == OE_EEID_VERSION
+               ? &eeid_page->eeid
+               : NULL;
+}
+
+const oe_enclave_config_t* __oe_get_enclave_config()
```
CodeMonkeyLeet, line 213: This seems to be another reason to carry the `config_id` within `oe_eeid_t` ... as far as I can tell, it's only used by the attester implementation:
- the attester plugin already has to `__oe_get_eeid()` today, so it makes sense for `config_id` to accompany it instead of carrying it as a new global.
- on KSS systems, it doesn't need to inject the custom claim, the base evidence has it already, so it also doesn't have a use for `__oe_get_enclave_config()`

enclave/core/sgx/init.c
```diff
@@ -41,33 +41,17 @@ static void _check_memory_boundaries(void)
...
+    /* Get the guard/EEID page */
+    const oe_eeid_page_t* eeid_page = (oe_eeid_page_t*)__oe_get_eeid_page();
+    if (!oe_is_within_enclave(eeid_page, OE_PAGE_SIZE))
+        oe_abort();
 
-    return r;
+    /* Without EEID, eeid_page page is all zero (including the version). */
+    if (eeid_page->eeid.version != 0 &&
+        eeid_page->eeid.version != OE_EEID_VERSION)
+        oe_abort();
```
CodeMonkeyLeet, line 54: Why does this abort if the page is not populated? From the comment in loadelf.c `_patch()`, it sounds like its valid for this page to exist but not be populated with eeid info if not used.

common/sgx/eeid_verifier.c
```diff
@@ -313,47 +386,65 @@ static oe_result_t _eeid_verify_evidence(
...
+        OE_RAISE(OE_VERIFY_FAILED);
+
+    const sgx_quote_t* sgx_quote = (const sgx_quote_t*)report_header->report;
+    const sgx_report_body_t* report_body = &sgx_quote->report_body;
+    const sgx_attributes_t* r_sgx_attributes = &report_body->attributes;
+    uint32_t r_misc_select = report_body->miscselect;
 
     const uint8_t* r_enclave_hash = parsed_report.identity.unique_id;
-    const uint8_t* r_signer_id = parsed_report.identity.signer_id;
+    const uint8_t* r_resigner_id = parsed_report.identity.signer_id;
```
CodeMonkeyLeet, line 403: Style: what is the `r_` prefix intended to convey here? Per coding guidelines, please avoid abbreviations; consider dropping the prefix or a meaningful word e.g. `in_`?

host/sgx/create.c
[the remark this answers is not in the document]
wintersteiger: Yes, that's exactly what it does, see 10 lines further down. This comment points to the dynamic case, not the static case.

host/sgx/create.c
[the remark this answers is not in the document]
wintersteiger: What would be the benefit of that? I currently use the fact that it's zero to recognize the fact that there is no EEID, which is very convenient. If we can't rely on this page being initialized, then I will need another new page that's initialized (to house that bit of information) or a custom flag in base images that's set if an image is intended only for EEID use.

enclave/core/sgx/init.c
wintersteiger: That's right, in that case the version is 0, because the whole page is zero. This check aborts if the version doesn't match 0 or the currently supported version (e.g. garbage).

common/sgx/eeid_verifier.c
wintersteiger: "reported", i.e. from the report (as opposed to computed, e.g. remeasured pages). All these names are way too long to be spelled out, especially at an 80 character line limit.

dthaler: I don't think expanding r to reported is a problem.  Line limit might result in a line break after the = but that would be ok.  "reported" is more readable.

common/sgx/eeid.c
```diff
@@ -614,7 +647,7 @@ static oe_result_t _hton_uint64_t(
     return OE_OK;
 }
 
-static oe_result_t _ntoh_uint64_t(
+oe_result_t ntoh_uint64_t(
```
dthaler, line 650: If it's not static, then it should have the oe prefix in the function name.
Also, the function name shouldn't end in _t which is reserved for types.
hence I'd recommend `oe_ntoh_uint64`

host/sgx/create.c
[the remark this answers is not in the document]
bodzhang: Adding a gaurd page instead of all-0 page between the Code and Initialized Data pages and the heap pages in the regualr enclave, the design causes minimum or argurably no impact to the regular enclave. Guard pages can't be accessed by enclave code, even if some code bug allows the attacker to trigger the enclave code to access memeory it's not suppoed to access, and are already used to seperate different sections of enclave pages in the existing regualr enclave design. With regard to the mechanism to distinguish EEID enclave for regualr enclave, I think an explict flag, for example, a EEID_SGX_Enclaev type in oe_enclave_type_t, is a cleaner design, with better readability and lesser security analysis need.

host/sgx/create.c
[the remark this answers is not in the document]
bodzhang: @wintersteiger, OK, I see. For the "static size" EEID enclave, the EEID page is added after heap and thread sections are added, but for the "dynamic size" EEID enclave, the EEID page is added before the heap and thread sections are added. I understand that the current implementation for "dynamic size" EEID enclave is simpler in this file, coding-wise, but the tradeoff is the enclave mesurement algorithm deviation from the "static size" EEID enclave. Perosnally, I prefer a consistent design over coding simplicty, but it's not a blocking issue for me.

include/openenclave/bits/eeid.h
```diff
@@ -17,7 +17,10 @@
 
 OE_EXTERNC_BEGIN
 
-#define OE_EEID_VERSION (1)
+#define OE_EEID_VERSION (2)
+
+#define OE_CLAIM_CONFIG_ID "config_id"
+#define OE_CLAIM_CONFIG "config"
```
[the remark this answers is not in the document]
bodzhang: To abstract out the low level implementaiton difference of EEID-based solution and KSS-based solution, the claims for the identity of the content loaded into the enclave post enclave initilization should use the same claim name.

host/sgx/create.c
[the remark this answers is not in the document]
bodzhang: The suggestion was to add a line: *vaddr += OE_PAGE_SIZE; inside _Add_data_pages(), at line 360. It shoudl simplify the code, if the EEID page is always added last, both in "static size" EEID enclave and in "dynamic size" EEID enclave. But if "dynamic size" EEID enclave implementaiton adds the EEID page before the heap and thread sections, the suggestion might not work well.

host/sgx/loadelf.c
```diff
@@ -692,8 +692,17 @@ static oe_result_t _patch(oe_enclave_image_t* image, size_t enclave_end)
...
         _set_uint64_t_symbol_value(image, "_reloc_size", image->reloc_size));
 
+#ifdef OE_WITH_EXPERIMENTAL_EEID
+    /* exactly one guard or EEID page */
+    uint64_t eeid_rva = image->image_size + image->reloc_size;
+    OE_CHECK(_set_uint64_t_symbol_value(image, "_eeid_rva", eeid_rva));
+
+    /* heap right after guard/EEID page */
+    oeprops->image_info.heap_rva = eeid_rva + OE_PAGE_SIZE;
+#else
```
bodzhang, line 702: Once the proposal of adding a guard page between the Code and Initialized Data pages and heap pages in regular enclaves is accepted, the code that adjusts `heap_rva` with OE_PAGE_SIZE should be moved out of the OE_WITH_EXPERIMENTAL_EEID compiling switch.

host/sgx/create.c
```diff
@@ -317,13 +317,19 @@ static oe_result_t _calculate_enclave_size(
...
                    (size_settings->num_tcs * (stack_size + control_size));
-
+#endif
     if (enclave_size)
     {
 #ifdef OE_WITH_EXPERIMENTAL_EEID
-        if (is_eeid_base_image(props))
+        if (is_eeid_base_image(&props->header.size_settings))
         {
             /* This is the maximum size SGX allows. When signing base images we
```
bodzhang, line 334: SGX allows much bigger enclave linear address range. The comment should be updated. For exmaple, "Maxinum size allowed for EEID SGX enclave with heap/stack size and NumTCS set by the enclave loader".

host/sgx/loadelf.c
bodzhang: In the updated EEID design proposal, `entry_rva` needs to be included in `oeprops`, for bothe EEID enclaves and regualr enclaves. Once the proposal is adopted, the code to set `entry_rva` probably can be added in this function, and outside of OE_WITH_EXPERIMENTAL_EEID compiling switch.

common/sgx/eeid.c
```diff
@@ -28,26 +28,24 @@
...
 #endif
 
-int is_eeid_base_image(const oe_sgx_enclave_properties_t* properties)
+int is_eeid_base_image(const oe_enclave_size_settings_t* sizes)
 {
-    return properties->header.size_settings.num_heap_pages == 0 &&
-           properties->header.size_settings.num_stack_pages == 0 &&
-           properties->header.size_settings.num_tcs == 1;
+    return sizes->num_heap_pages == 0 && sizes->num_stack_pages == 0 &&
+           sizes->num_tcs == 1;
```
bodzhang, line 34: In the updated EEID design, the "dynamic size" EEID enclave will have NumTCS=0.

enclave/core/sgx/globals.c
```diff
@@ -189,9 +190,29 @@ size_t __oe_get_reloc_size()
...
+    return (oe_eeid_page_t*)(base + _eeid_rva);
+}
+
+const oe_eeid_t* __oe_get_eeid()
+{
+    if (!_eeid_rva)
+        return NULL;
+
+    const oe_eeid_page_t* eeid_page = __oe_get_eeid_page();
+    return eeid_page && eeid_page->eeid.version == OE_EEID_VERSION
```
bodzhang, line 208: In the updated EEID design, the enclave code shoud check `entry_rva` in the eeid data matches the value in oeprops.

enclave/core/sgx/globals.c
```diff
@@ -213,8 +234,9 @@ const void* __oe_get_heap_base()
 size_t __oe_get_heap_size()
 {
 #ifdef OE_WITH_EXPERIMENTAL_EEID
-    if (oe_eeid)
-        return oe_eeid->size_settings.num_heap_pages * OE_PAGE_SIZE;
+    const oe_eeid_t* eeid = __oe_get_eeid();
+    if (eeid)
+        return eeid->size_settings.num_heap_pages * OE_PAGE_SIZE;
```
bodzhang, line 239: For "static size" EEID enclave, the heap size info is in `oe_enclave_properties_sgx`.

common/sgx/eeid.c
wintersteiger: Can you confirm that there is no code path that requires `NumTCS=1`, i.e. not for dynamic, static, or non-eeid?

host/sgx/create.c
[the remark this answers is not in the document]
wintersteiger: There is now a clean setting (i.e. `oe_enclave_setting_eeid_t`) that steers the logic and non-EEID images neither have a gap in the address space nor a measured page, so no impact on existing non-EEID images.

host/sgx/create.c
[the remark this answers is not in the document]
wintersteiger: Note that there is a difference between the time at which the measurement is taken and the space in which the EEID information resides. The EEID page was always at the same location and it's now also measured at the same time, but the code logic is significantly more complex now.

host/sgx/create.c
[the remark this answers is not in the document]
wintersteiger: For dynamic sizes we have to save the EEID related information before data pages are added, there is no way around that. There is now a dedicated function for this called `_save_eeid_info`.

enclave/core/sgx/init.c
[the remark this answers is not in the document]
wintersteiger: Regular (i.e. non-EEID, non-configid) enclaves have no EEID page or a gap in the address space in the current version. There is also a new global symbol called `_eeid_rva`, which is zero for regular enclaves.

include/openenclave/attestation/sgx/eeid_verifier.h
```diff
@@ -11,7 +11,8 @@
 
 OE_EXTERNC_BEGIN
 
-#define OE_CLAIM_EEID_BASE_ID "eeid_base_unique_id"
+#define OE_CLAIM_EEID_RESIGNED_UNIQUE_ID "eeid_resigned_unique_id"
```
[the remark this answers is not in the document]
wintersteiger: Quite right, done, thanks!

common/sgx/eeid_verifier.c
wintersteiger: Done, thanks!

include/openenclave/bits/types.h
```diff
@@ -180,4 +180,24 @@ typedef struct _oe_datetime
...
 
+#ifdef OE_WITH_EXPERIMENTAL_EEID
+/**
+ * This struct defines a buffer for enclave configuration (EEID or configid).
+ */
+typedef struct
+{
+    uint8_t* data;
+    size_t size;
+} oe_enclave_config_t;
```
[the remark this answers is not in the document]
wintersteiger: That's not the ID (32 byte) in 'config_id', it's the (larger) data that then gets hashed into a config_id. In line with the EEID acronym, I renamed it to `oe_enclave_initialization_data_t`, that should be much clearer.

include/openenclave/bits/types.h
```diff
@@ -180,4 +180,24 @@ typedef struct _oe_datetime
     uint32_t seconds; /* range: 0-59 */
 } oe_datetime_t;
 
+#ifdef OE_WITH_EXPERIMENTAL_EEID
```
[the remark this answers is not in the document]
wintersteiger: Yeah, you're right, we'll probably want new datastructures for that. Moved.

include/openenclave/bits/report.h
```diff
@@ -48,6 +48,13 @@ OE_EXTERNC_BEGIN
  */
 #define OE_PRODUCT_ID_SIZE 16
 
+#ifdef OE_WITH_EXPERIMENTAL_EEID
+/**
+ * Size of the enclave config hash in bytes.
```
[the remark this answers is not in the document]
wintersteiger: Fixed. Thanks for the heads up - I was expecting that to happen at some point, but wasn't sure whether it was on the table yet.

include/openenclave/internal/eeid.h
```diff
@@ -22,16 +24,23 @@ static const uint8_t OE_DEBUG_PUBLIC_KEY[] = {
     0xce, 0x73, 0xe4, 0x33, 0x63, 0x83, 0x77, 0xf1, 0x79, 0xab, 0x44,
     0x56, 0xb2, 0xfe, 0x23, 0x71, 0x93, 0x19, 0x3a, 0x8d, 0xa};
 
+/** Layout of the EEID memory page. */
+typedef struct
+{
+    OE_SHA256 config_id;
```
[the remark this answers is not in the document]
wintersteiger: Mainly because `oe_eeid_t` was a dirty hack :-) The config_id has now moved into `oe_eeid_t`, but the user now submits a `oe_enclave_setting_eeid_t`, which contains only the bits they should care about.

enclave/core/sgx/calls.c
wintersteiger: (Doesn't exist anymore)

enclave/core/sgx/calls.c
wintersteiger: Yeah, I wasn't sure which one to pick, there isn't any that perfectly describes the error. (Changed to `OE_VERIFY_FAILED` now.)

enclave/core/sgx/globals.c
wintersteiger: Yep, that's where the `config_id` is now.

enclave/core/sgx/init.c
wintersteiger: This has been refactored. There is a new global `_eeid_rva` now that indicates whether (and where) the EEID is.

enclave/core/sgx/calls.c
```diff
@@ -135,12 +136,65 @@ extern bool oe_disable_debug_malloc_check;
...
+#endif
+
 /*
 **==============================================================================
 **
 ** _handle_init_enclave()
 **
-**     Handle the OE_ECALL_INIT_ENCLAVE from host and ensures that each state
+**     Handle the OE_ECALL_INIT_ENCLAVE from host and ensures that each
+*state
```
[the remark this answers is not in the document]
wintersteiger: Yeah, auto-formatter, fixed.

enclave/core/sgx/calls.c
```diff
@@ -431,6 +497,11 @@ static void _handle_ecall(
             /* Cleanup verifiers */
             oe_verifier_shutdown();
 
+#ifdef OE_WITH_EXPERIMENTAL_EEID
+            /* Cleanup config */
```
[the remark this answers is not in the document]
wintersteiger: Fixed, thanks.

enclave/core/sgx/globals.c
```diff
@@ -189,9 +190,29 @@ size_t __oe_get_reloc_size()
 **==============================================================================
 */
 
-const void* __oe_get_eeid()
+const oe_eeid_page_t* __oe_get_eeid_page()
```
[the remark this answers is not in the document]
wintersteiger: Thanks for the pointer, I got rid of all the double underscores in this PR now.

include/openenclave/internal/eeid.h
```diff
@@ -22,16 +24,23 @@ static const uint8_t OE_DEBUG_PUBLIC_KEY[] = {
...
+    oe_eeid_t eeid;
+} oe_eeid_page_t;
+
 /**
  * Determine whether properties are those of a base image to be used with EEID
  *
- * @param[in] properties Properties of an SGX enclave.
+ * @param[in] sizes The size properties of an SGX enclave.
  *
  * @returns Returns true if **properties** are consistent with those of an EEID
```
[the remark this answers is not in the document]
wintersteiger: Fixed, thanks.

include/openenclave/internal/eeid.h
```diff
@@ -22,16 +24,23 @@ static const uint8_t OE_DEBUG_PUBLIC_KEY[] = {
...
  * base image.
  *
  */
-int is_eeid_base_image(const oe_sgx_enclave_properties_t* properties);
+int is_eeid_base_image(const oe_enclave_size_settings_t* sizes);
```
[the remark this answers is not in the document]
wintersteiger: "sizes" is just short for "some `oe_enclave_size_settings_t`". They are persistent in that they are (part of) the settings of an enclave, which appear in `oe_enclave_properties_header_t`.

include/openenclave/internal/eeid.h
```diff
@@ -85,6 +100,8 @@ typedef struct
  *
  * @param[in] eeid The EEID containing the required size settings.
  *
+ * @param[in] config_hash The SHA256 of the enclave config.
```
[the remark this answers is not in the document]
wintersteiger: Fixed, thanks!

include/openenclave/internal/eeid.h
```diff
@@ -199,9 +229,80 @@ typedef struct
...
+ * @returns Returns OE_OK on success.
+ *
+ */
+oe_result_t oe_enclave_size_settings_hton(
+    const oe_enclave_size_settings_t* sizes,
+    uint8_t** position,
+    size_t* remaining);
+
+/**
+ * Deserialize a oe_enclave_size_settings_t from network byte-order into a
```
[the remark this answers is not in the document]
wintersteiger: Fixed (see comment in the other thread).

include/openenclave/internal/globals.h
```diff
@@ -33,8 +37,14 @@ uint64_t oe_get_num_heap_pages(void);
...
-/* Extended enclave initialization data */
-const void* __oe_get_eeid(void);
+/* Extended enclave initialization meta-data page */
+const oe_eeid_page_t* __oe_get_eeid_page(void);
+
+/* Extended enclave initialization meta-data */
+const oe_eeid_t* __oe_get_eeid(void);
+
+/* Enclave config (EEID or configid) */
+const oe_enclave_config_t* __oe_get_enclave_config(void);
```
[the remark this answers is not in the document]
wintersteiger: I agree too, renamed!

tests/eeid_plugin/host/host.c
```diff
@@ -220,32 +229,54 @@ void start_enclave(const char* filename, uint32_t flags, enclave_stuff_t* stuff)
...
+            stuff->resigned_enclave_hash = stuff->claims[i].value;
+        else if (strcmp(name, OE_CLAIM_CONFIG_ID) == 0)
+            stuff->config_hash = stuff->claims[i].value;
+    }
 
     OE_TEST(stuff->enclave_hash != NULL);
+
+    /* Check that the claimed config_id matches */
+    OE_SHA256 config_hash;
+    oe_sha256(stuff->config.data, stuff->config.size, &config_hash);
```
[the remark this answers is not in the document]
wintersteiger: Couldn't agree more! But yes, this is not new here.

common/sgx/eeid.c
wintersteiger: Fixed, thanks!

host/sgx/create.c
wintersteiger: What does the enclave loader have to do with that? I found this number by trying to start enclaves with different sizes and this was the highest that my CPU didn't complain about. I'm sure newer CPUs will have other limits, but I haven't found any clear information about that yet.

wintersteiger:
Global comment: I had to add a flag (`-E`) to `oesign` to enable EEID during signing, because without this, I can't distinguish between non-EEID images and statically sized EEID images. For the same reason, there is also a new `bool eeid_enabled;` in `load.h`, which should really go elsewhere, but I couldn't find a better place. Ideally, I'd like to get rid of both of those flags, but I don't think that's possible if we want statically sized EEID images.

host/sgx/create.c
```diff
@@ -676,37 +738,52 @@ static oe_result_t _eeid_resign(
...
+    oe_enclave_t* enclave,
+    oe_sgx_enclave_properties_t* props,
+    uint64_t entry,
     uint64_t* vaddr)
 {
     oe_result_t result = OE_UNEXPECTED;
-    oe_eeid_t* eeid = context->eeid;
 
-    if (eeid)
+    /* Note: context->eeid_setting is also needed at signing time, for static
```
bodzhang, line 750: Why does the "static size" EEID enclave's base image needs context->eeid_setting?

wintersteiger: To distinguish it from non-EEID images. If EEID is enabled at signing time, it will leave a gap/guard page, while the "normal" non-EEID behavior is unchanged.

host/sgx/create.c
```diff
@@ -589,57 +624,85 @@ oe_result_t oe_sgx_validate_enclave_properties(
...
         eeid->vaddr = *vaddr;
-        eeid->signature_size = sizeof(sgx_sigstruct_t);
-        memcpy(
-            eeid->data + eeid->data_size,
-            (uint8_t*)&props->sigstruct,
-            sizeof(sgx_sigstruct_t));
+        eeid->signature_size = signature_size;
+        memcpy(eeid->signature, props->sigstruct, signature_size);
+
+        /* Compute config hash (all zero during signing) */
```
bodzhang, line 663: During the base image signing, EEID page should not be measured. Does the comment and implementaiton here mean EEID page with `config_id`=0 is part of the base image measurement?

wintersteiger: No, during signing this function is only used to record relevant information (like the hash state) so that `oesign dump-eeid` can dump it, but nothing is being measured. The signing case is here: https://github.com/openenclave/openenclave/blob/204830a5721b796684c62268474d8ce66ddc9e08/host/sgx/create.c#L769 and it doesn't call `_add_eeid_page` or `_eeid_resign`.

host/sgx/create.c
```diff
@@ -589,57 +624,85 @@ oe_result_t oe_sgx_validate_enclave_properties(
...
+        memset(eeid->config_id, 0, sizeof(eeid->config_id));
```
bodzhang, line 664: This implementation locks the definition of `config_id`, which should be applicaiton specific. A default definition of `conif_di`=SHA256(`eeid_setting.data`) is OK, but we should allow the enclave developer to override it. Including `config_id` (and `config_svn`) in `eeid_setting` would allow the developer to choose whether to use the default definition of `config_id`. If the developer sets `eeid_setting.data_size` as non zero, the default defintion is choosen. If the developer sets `eeid_setting.data_size`=0, the implementation should load `config_id`from `eeid_setting.config_id`as it is. The enclave developer can implement an explicit ECALL to add additional content post enclave initialization, and fully control the expected relationship between `config_id` (and `config_svn`) and the additional content.

common/sgx/eeid.c
bodzhang: I didn't see any check requiring NumTCS>0 in current OE code.

wintersteiger: There isn't; the last time I tried it ran just fine with NumTCS==0. My question was about your security analysis: is there any circumstance in which NumTCS==1 is required? (I don't think so, see also other thread about entry point, but the verifier should probably reject quotes from enclaves that claim they're running with NumTCS==0.)

common/sgx/eeid.c
```diff
@@ -255,8 +246,23 @@ oe_result_t oe_deserialize_eeid(
...
+    OE_CHECK(deserialize_element(
+        position, &remaining, (uint8_t*)data_size, sizeof(*data_size)));
+    if (data)
+    {
+        *data = malloc(*data_size);
+        if (!*data)
+            OE_RAISE(OE_OUT_OF_MEMORY);
+        OE_CHECK(deserialize_element(
+            position, &remaining, (uint8_t*)data, *data_size));
+    }
```
bodzhang, line 265: The SW convention the eeid enclave loader is suppoed to follow is to copy all fields of the base image `SIGSTRUCT` into the "dynamically" signed `SIGSTRUCT`, exepect `ENCLAVEHASH` and signing related fields. To verify that the eeid enclave loader indeed followed the SW convention, the verifier should perform checks between all relevant fileds in the base image `SIGSTRUCT` and the SGX `QUOTE` fileds, exactly as what the CPU would do. 
- reported_attribute(same as SECS.ATTRIBUTE) & baseSIGSTRUCT.ATTRIBUTEMASK = baseSIGSTRUCT.ATTRIBUTES & baseSIGSTRUCT.ATTRIBUTEMASK,
- reported_misc_select(same as SECS.MISCSELECT) & baseSIGSTRUCT.MISCMASK = baseSIGSTRUCT.MISCSELECT & baseSIGSTRUCT.MISCMASK,
- reported_product_id = baseSIGSTRUCT.ISVPRODID,
- reported_security_version = baseSIGSTRUCT.ISVSVN

If we don't prevent SGX2.0 enclave from being launched as EEID enclave in the future, the check should also inlcude ISVFAMILYID, ISVEXTPRODID

wintersteiger: Yes, that's exactly what the loader does. The verifier then performs most of those checks here: https://github.com/openenclave/openenclave/blob/c75ff164fda058f914782279dc12d4282223943a/common/sgx/eeid.c#L580 

The attributes change during enclave creation and the enclave doesn't have access to a pre-creation copy of the sigstruct. I'll have to run a few more checks to see if the masks actually do line up as you suggest.

wintersteiger: I add more masked checks which do pass the tests:  https://github.com/openenclave/openenclave/pull/3236/commits/abfdd68b050e70b3d86a960195f4899a14852d12

Note the negations on the masks; is that the correct masking?

wintersteiger: Something's off here. The Intel manual also says `&` without `~`, but then the default OE attribute mask doesn't make sense to me. Is the default right?  https://github.com/openenclave/openenclave/blob/05d109a5706fe4c439f64cd68ffee6e2430b9542/include/openenclave/bits/sgx/sgxtypes.h#L136

common/sgx/eeid.c
bodzhang: As long as the EEID init code inside the enclave checks that the entry point in EEID page matches the value in enlcave property (initalized data of the base image), "dyanmic size" EEID enclave is safe to use NumTCS=0. As a defense-in-depth implementation, the verifier can reject attestation with invalid size_info in the EEID page. But the measurement check should catch any manipulation of NumTCS on "dynamic size" EEID enclave.

host/sgx/create.c
bodzhang: The design goal is to add the guard page between the Code and Initialized Data pages and the heap pages in non-EEID enclave too. Once the design proposal is adopted, we can simplify the implementaiton with regard to some of the "OE_WITH_EXPERIMENTAL_EEID" compiling switches.

common/sgx/eeid.c
wintersteiger: Great, thanks for clarifying. I still think we don't need the entry point in the enclave properties, but I'll leave that discussion in the other thread.

common/sgx/eeid.c
wintersteiger: Got it now. It's really just about the "initialized" flag. The base image sigstruct was never created or initialized, so it's missing that flag. I updated the code accordingly.

host/sgx/create.c
wintersteiger: Right, I thought you wanted to keep the existing behavior unchanged. But, even then, the enclave has no way to tell whether it should expect a populated EEID page. In the past I experimented with a magic value like 0xEE1D on the unmeasured marker page, but that isn't great either. Currently, the `_eeid_rva` symbol is patched with the offset of the EEID page and a non-EEID enclave gets a `0` in there. For this to work, I have to know whether EEID is enabled at signing time though, otherwise the base image hash (and hash state) would change.

common/sgx/eeid.c
```diff
@@ -495,20 +522,25 @@ oe_result_t verify_eeid(
     uint16_t reported_product_id,
     uint32_t reported_security_version,
     uint64_t reported_attributes,
+    const sgx_attributes_t* reported_sgx_attributes,
+    uint32_t reported_misc_select,
     const uint8_t** base_enclave_hash,
-    const oe_eeid_t* eeid)
+    const oe_eeid_t* eeid,
+    const oe_enclave_size_settings_t* base_image_sizes)
```
bodzhang, line 529: Does the code extract `base_image_size` from the Attestation Evidence?

common/sgx/eeid.c
bodzhang: Should not use ~. If a bit in the mask is 0, the loader is free to set the corresponding bit in SECS. If a bit in the mask is 1, the loader must set the corresponding bit in SECS as same as the crespsonding bit in SIGSTRUCT.ATTRIBUTES (or MISCSELECT). So the default SGX_SIGSTRUCT_MISCSELECT=0x00000000,SGX_SIGSTRUCT_MISCMASK=0xffffffff mean the load must set SECS.MISCSELECT=0x00000000 (no opt-in for any MISCSELECT feature). The default of SGX_SIGSTRUCT_ATTRIBUTEMASK_FLAGS=0XfffffffffffffffbULL, SGX_SIGSTRUCT_ATTRIBUTEMASK_XFRM=0x0000000000000000ULL means the loader is free to enable/disable extended feature set in the enclave, but must make sure most bits in SECS.ATTRIBUTES.FLAGS matches SIGSTRUCT.ATTRIBUTES.FLAGS. I have some doubt about allowing the loader to enable MPX in enclave. I will analyze it later.

common/sgx/eeid.c
```diff
@@ -544,10 +574,23 @@ oe_result_t verify_eeid(
     // Check other image properties have not changed
     bool base_debug = sigstruct->attributes.flags & SGX_FLAGS_DEBUG;
     bool extended_debug = reported_attributes & OE_REPORT_ATTRIBUTES_DEBUG;
+    bool sgx_debug = reported_sgx_attributes->flags & SGX_FLAGS_DEBUG;
+
+    uint64_t attributes_masked =
+        reported_sgx_attributes->flags & ~sigstruct->attributemask.flags;
```
bodzhang: Shoud remove the "~"

bodzhang: Should also check attributes.xfrm.

common/sgx/eeid.c
```diff
@@ -544,10 +574,23 @@ oe_result_t verify_eeid(
     // Check other image properties have not changed
     bool base_debug = sigstruct->attributes.flags & SGX_FLAGS_DEBUG;
     bool extended_debug = reported_attributes & OE_REPORT_ATTRIBUTES_DEBUG;
+    bool sgx_debug = reported_sgx_attributes->flags & SGX_FLAGS_DEBUG;
```
bodzhang, line 577: Is `extended_debug` same as `sgx_debug`?

host/sgx/create.c
bodzhang: Based on the updated design, _eeid_rva (or the addr of the guard page before the heap, in the regualr enclaves) is fixed as (_heap_rva - 4KB). _heap_rva is already part of the intialized data of the enclave and can be accessed by the enclave code. For the "static size" EEID enclaves, the size of the enclave does not need to be patched ("dyanmic size" EEID enclave needs to patch the size as a fixed, large number), neither does the entry_point in enclave property (the updated deisgn calls for the regaulr enclave to include entry_point in enclave ptoperty too). The build tool does not need to distinuish "static size" EEID enclave vs. regualr enclave. "Dynamic size" EEID enlcave does have to be "patched" during build time by the build tool.

host/sgx/create.c
bodzhang: CPUID 12H sub-leaf0 (ECX=0) reports the max enclave size supported. Maybe some of the SGX CPUs do have 64GB as the max enclave size (64bit mode), but many other SGX CPUs have a larger limitation. I still think the comment should be changed. For exmaple "When signing base images we don't know the size that the final image will have, so we chose a size big enough to accomendate resonable heap/stack and #TCS selections, and supported by known SGX CPUs".

host/sgx/create.c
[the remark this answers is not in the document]
bodzhang: _Save_eeid_info() implementation looks good to me.

host/sgx/create.c
wintersteiger: We can't patch any base image at creation time, regardless of whether its statically or dynamically patched, because the verifier would have to re-patch it to verify the evidence, which in turn also means you have to have a copy of the image so you can patch it. Any patching must be done during signing and in exactly the same with (with the same numbers) during enclave creation. (Just in case: "patching" means modifying the base image in any way.)

You listed only the easy cases: what about non-EEID enclaves, how do they know not to expect an EEID page? Should they try to read from the page and then somehow catch the access violation?

I don't know what your "build tool" is, but `oesign` definitely needs to distinguish between EEID and non-EEID images, except if we find a way to distinguish them later from inside the enclave. This was easy in the past, because all images had their memory sizes set to 0/0/1, but that's not the case with statically sized images anymore.

host/sgx/create.c
wintersteiger: I didn't know the CPU can tell us the max size, we should definitely do that and remove this comment.

wintersteiger: Actually, I'm wrong. In many use-cases images will be built on one CPU but run on another.

common/sgx/eeid.c
bodzhang: For a base image with NumTCS=0, we do need the entry point as part of the base image measurement and require the EEID init code to check eeid.entry_point matches the entry point info in the base image.

enclave/core/sgx/init.c
```diff
@@ -41,33 +40,23 @@ static void _check_memory_boundaries(void)
...
+            oe_abort();
+
+        /* Check that the signature is within the EEID page */
+        if (sizeof(oe_eeid_t) + eeid->signature_size > OE_PAGE_SIZE)
+            oe_abort();
+
+        /* Check that the EEID version matches */
+        if (eeid->version != OE_EEID_VERSION)
+            oe_abort();
     }
```
bodzhang: Also need to check `eeid->entry_point`

enclave/core/sgx/globals.c
```diff
@@ -184,14 +184,23 @@ size_t __oe_get_reloc_size()
...
+** Extended enclave initialization data:
 **
 **==============================================================================
 */
 
-const void* __oe_get_eeid()
+bool oe_have_eeid(void)
 {
-    return oe_eeid;
+    return _eeid_rva != 0;
```
bodzhang: `_eeid_rva` is measured and indicats whether the base enclave is signed as an EEID enclave. Should add a comment to explain the design.

enclave/core/sgx/globals.c
```diff
@@ -213,8 +222,8 @@ const void* __oe_get_heap_base()
 size_t __oe_get_heap_size()
 {
 #ifdef OE_WITH_EXPERIMENTAL_EEID
-    if (oe_eeid)
-        return oe_eeid->size_settings.num_heap_pages * OE_PAGE_SIZE;
+    if (oe_have_eeid())
+        return oe_get_eeid()->size_settings.num_heap_pages * OE_PAGE_SIZE;
```
bodzhang: "Static size" EEID enclave's base image already specifies heap/stack size and number of threads in `oe_enclave_properties_sgx`. If an malicious loader sets `size_settings` in the eeid page differently, the enclave code might be fooled to read/write into unepxected memeory location. Whether the attack can be used to manipulate the eeid data inlcuded in the Attestation Evidence to still pass the verification requires more analysis. It's safer to use the heap size info in the base enclave image in the "static size" EEID enclave case. The 0-heap/0-stack/0-TCS flag of the "dynamic size" EEID enclave can be used to switch the code between `size_settings` in `oe_enclave_properties_sgx` and  `size_settings` in the eeid page.

host/sgx/create.c
bodzhang: I realized that `_eeid_rva` is measured as part of the base image and used to indicate whether the base image is signed as an EEID enclave. I'm OK with this design.

host/sgx/create.c
```diff
@@ -1005,8 +1096,19 @@ oe_result_t oe_create_enclave(
...
-    OE_CHECK(_initialize_enclave(enclave));
+/* Invoke enclave initialization. */
+#ifdef OE_WITH_EXPERIMENTAL_EEID
+    if (context.eeid_setting)
+    {
+        oe_enclave_with_config_t enclave_with_config = {0};
+        enclave_with_config.enclave = enclave;
+        enclave_with_config.config.data = (uint8_t*)context.eeid_setting->data;
+        enclave_with_config.config.size = context.eeid_setting->data_size;
+        OE_CHECK(_initialize_enclave_with_config(&enclave_with_config));
```
bodzhang: Instead of using a "config_data" specific `_initialize_enclave` call, can follow the design of the switchless call implementaiton, and use a dedicated `oe_load_config_data` ECALL, invoked within `_config_enclave` (line 1131), if the `settings` indicate the presence of config_data. This way, the enclave initialization code is not affected and `oe_load_config_data` ECALL happens after all necessary initalizations are done. Future changes to the enclave intialization design is much less likely to cause any compatibility issue with `oe_load_config_data` flow. If the developer of an enclave supporting loading additional content post initalization (EEID enclave or future SGX KSS-based implementation) does not provide config_data through `settings`, the devloper can implement their own ECALL to load additional content into the enclave, with thier own definition of config_id/config_svn (config_id is not necceary a hash of config_data). Note that not loading any additional data is a legit usage of Enclaves that support loading additional data.

common/sgx/eeid.c
bodzhang: It is from a customer claim in the Attestation Evidence.

enclave/core/sgx/init.c
wintersteiger: Outdated. No measured entry point yet, so nothing to check against.

enclave/core/sgx/globals.c
wintersteiger: Outdated.

enclave/core/sgx/globals.c
wintersteiger: Outdated.

host/sgx/create.c
wintersteiger: Outdated. Regardless of what we do, the developer can always add their own ecall and overwrite (or just ignore) the default config data. I won't touch this until the design discussion is finished and the new ecall is signed off on.

common/sgx/eeid.c
wintersteiger: Outdated. No, extended_debug is the debug flag that OE reports, not SGX.

common/sgx/eeid.c
wintersteiger: That's correct, it's from the base image and communicated via a custom claim.

common/sgx/eeid.c
wintersteiger: Outdated. I haven't thought about MPX yet, but in general I think it makes sense to review those defaults. The rest works since last week, with the minor annoyance of having to add the init-bit for the base image.

host/sgx/create.c
wintersteiger: Outdated.

enclave/core/sgx/calls.c
```diff
@@ -135,6 +136,160 @@ extern bool oe_disable_debug_malloc_check;
...
+    /* Check that the entry-point is as expected. */
+    if (eeid->entry_point != oe_enclave_properties_sgx.image_info.entry_rva)
+        oe_abort();
+
+    if (have_data)
+    {
+        /* Check that the data hashes to the correct value. */
+        OE_SHA256 config_hash;
+        oe_sha256(data.data, data.size, &config_hash);
+        if (memcmp(config_hash.buf, eeid->config_id, OE_SHA256_SIZE) != 0)
```
bodzhang, line 253: Use `oe_constant_time_mem_equal`, instead of `memcmp` for hash/MAC verification, to be consisitent with other OE code.

enclave/core/sgx/calls.c
```diff
@@ -135,6 +136,160 @@ extern bool oe_disable_debug_malloc_check;
...
+            OE_RAISE(OE_VERIFY_FAILED);
+
+        /* Copy the initialization data into enclave heap. */
+        _enclave_initialization_data =
+            oe_malloc(sizeof(oe_enclave_initialization_data_t));
+        if (!_enclave_initialization_data)
+            OE_RAISE(OE_OUT_OF_MEMORY);
+        _enclave_initialization_data->data = data.data;
```
bodzhang, line 261: Defining `_encalve_initialization_data` as a `oe_enclave_initialization_data_t` structure instead of a pointer will make the code simpler to understand.

enclave/core/sgx/calls.c
```diff
@@ -135,6 +136,160 @@ extern bool oe_disable_debug_malloc_check;
...
+    {
+        /* We have EEID with a config_id, but no data. It is up to the user to
+         * check that the config_id is as expected. */
+    }
+
+    eeid_checked = true;
+    result = OE_OK;
+
+done:
+    oe_free(data.data);
```
bodzhang, line 276: In several error paths, `data.data` is NULL. It's also explictly set to NULL on line 263.

enclave/core/sgx/calls.c
wintersteiger: There are exactly 4 uses of this function, while there are ~40 of uses of `memcmp` in `./enclave` and `./common`. Is there any benefit to this being constant time?

enclave/core/sgx/calls.c
wintersteiger: Then there would be no way to distinguish no-config from empty-config.

enclave/core/sgx/calls.c
wintersteiger: That's right and there's no problem with that. `free` or `oe_free` handle `NULL` as expected.

enclave/core/sgx/calls.c
bodzhang: I see. Once we have a flag for EEID enclave, as part of the base image, we won't need the "no-config" indication through `_enclave_initalization_data` pointer anymore.

enclave/core/sgx/calls.c
bodzhang: OK

enclave/core/sgx/calls.c
wintersteiger: Actually, can you add something about that flag to the design proposal? This would also mean that we need a (command-line) flag to `oesign`, because we need to patch that symbol to a fixed value at signing time. Some others may object to that and it may also not be in the "spirit" of the SGX config_id, so that users wouldn't expect that they need an additional flag.

enclave/core/sgx/calls.c
bodzhang: Non-constant-time memcmp might leak side channel info about partial match of the two string being compared. Known attack on HMAC was discovered, for example, on [openvpn](https://bugzilla.redhat.com/show_bug.cgi?id=960192), and the fix is to use constant-time-compare. I'm not sure whether SHA256 hash can be impacted the same way. But some community members for several crypto projects seem to advocate for simply replacing memcmp with the constant-time-compare. OE SDK seems to follow the same reasoning to use constant-time-compare in low level crypto related code.

wintersteiger: So I expect there will be a different PR that changes all/most of them? (There is no side-channel here because the config is public information anyways.)

enclave/sgx/eeid_attester.c
```diff
@@ -131,17 +139,30 @@ static oe_result_t _eeid_get_evidence(
...
         *endorsements_buffer_size = 0;
 
-    if (!eeid || eeid->data_size == 0 ||
-        eeid->signature_size != sizeof(sgx_sigstruct_t))
+    if (eeid->signature_size != sizeof(sgx_sigstruct_t))
         OE_RAISE(OE_FAILURE);
 
     size_t eeid_size = oe_eeid_byte_size(eeid);
 
+    // Serialize custom claims, including EEID base image sizes
```
bodzhang, line 147: A comment for the eeid evidence layout will be helpful to understand the code.

enclave/core/sgx/calls.c
```diff
@@ -161,7 +316,16 @@ static oe_result_t _handle_init_enclave(uint64_t arg_in)
...
         if (_once == false)
         {
-            oe_enclave_t* enclave = (oe_enclave_t*)arg_in;
+            oe_enclave_t* enclave;
+#ifdef OE_WITH_EXPERIMENTAL_EEID
+            oe_enclave_with_initialization_data_t* enclave_plus =
+                (oe_enclave_with_initialization_data_t*)arg_in;
+            enclave = enclave_plus->enclave;
+
+            OE_CHECK(_check_and_load_initialization_data(enclave_plus));
```
bodzhang, line 325: Would be safer to call this function after basic initalization functions are done, for example, after line 348.

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -23,6 +23,7 @@ typedef struct _oe_sgx_enclave_image_info_t
     uint64_t reloc_size;
     uint64_t heap_rva; /* heap size is in header.sizesettings */
     uint64_t enclave_size;
+    uint64_t entry_rva;
```
bodzhang, line 26: Although the latest EEID design proposal does not rely on `entry_rva` in _oe_sgx_enclave_image_info_t. I stil think this change is valuable for future EDMM work.

common/sgx/eeid.c
```diff
@@ -494,21 +515,26 @@ oe_result_t verify_eeid(
...
     oe_result_t result = OE_UNEXPECTED;
 
     if (!eeid)
         OE_RAISE(OE_INVALID_PARAMETER);
 
     if (eeid->signature_size != 1808) // We only support SGX sigstructs for now.
-        OE_RAISE(OE_VERIFY_FAILED);
+        OE_RAISE(OE_UNSUPPORTED);
+
+    bool static_sizes = !is_eeid_base_image(base_image_sizes);
```
bodzhang, line 533: Instead of using the base_image_sizes provided by the attestating enclave code, can use the `size_settings` in eeid data structure to drive the static vs. dynamic size logic. A "static size" EEID enclave is suppoed to have al-0 `size_setting` in eeid. The eeid data is measured, any spoofing or tampering of eeid data will be caught by mesurement checks.

enclave/sgx/eeid_attester.c
bodzhang, line 147: Base enclave size_setting info is not needed by the eeid enclave attestation verifier to do its verification work. See comments in eeid_verifier.c. A base enclave size_setting claim is not useful for the Relying Party either, as the potenitally malicious additional content loaded into an EEID enclave might be able to modify code logic to spoof base enclave size_setting info in the attestaiton evidence, if base enclave size_setting info is critical to detect the malicious cotent. I don't think it needs to be inlcuded in the attestation evidence.

enclave/core/sgx/calls.c
bodzhang: The flag is discussed in the latest version of the design proposal.

enclave/core/sgx/calls.c
wintersteiger: Why would that be safer?

bodzhang: `_check_and_load_initilization_data()` use malloc and other functions which need to be initialized. If OE enclave initialization implementation changes and moves things around, invoking malloc or other functions before a set of initialization functions might fail.

wintersteiger: The memory allocator is why we used to do part of this in `init.c`, but that's not necessary anymore. If things get moved around, this may also have to move, but I prefer to do this as early as possible to avoid any unexpected effects.

radhikaj:
@bodzhang @wintersteiger What is the intersection between this PR and  https://github.com/openenclave/openenclave/pull/3269? Can both go in?
If this is being replaced by 3269, can this be closed?

wintersteiger:
@radhikaj This is the draft implementation of the design changes proposed in #3269. Both will go in, but this PR lags behind the ongoing design changes in #3269.

wintersteiger:
Abandoned, closing!

PR #3238, opened 2020-07-08
Remove -readnow from symbol load line in oegdb

jhand2:
The using the -readnow flag can cause some performance issues when debugging very large enclaves with multiple symbol files. Remove the -readnow flag to allow lazy loading of symbols.


jhand2:
I am not able to repro these failures locally. Trying again

bors r+

radhikaj:
bors try

jhand2:
I guess the problem here is that due to lazy loading the exact stdout might not match what's in the test because the symbols might be loaded at a slightly different time. So I don't think an exact stdout match really works for these tests if we make this change.

anakrish:
I'll take a look and see what's going on.

jhand2:
@anakrish did you ever find what was going on here? Otherwise I'll close until we can investigate further.

PR #3244, opened 2020-07-10
Fix GCC errors about OE_RAISE jumping variable initialization

wintersteiger:
Fixes GCC errors about the `goto` in `OE_RAISE` jumping variable initializations (`-Werror=jump-misses-init`) in EEID code; no semantic changes.

### DIFF
--- a/common/sgx/eeid.c
+++ b/common/sgx/eeid.c
@@ -132,12 +132,11 @@ done:
 oe_result_t oe_serialize_eeid(const oe_eeid_t* eeid, char* buf, size_t buf_size)
 {
     oe_result_t result = OE_UNEXPECTED;
+    char** position = &buf;
+    size_t remaining = buf_size;
 
     if (!eeid || !buf || !buf_size)
         OE_RAISE(OE_INVALID_PARAMETER);
-
-    char** position = &buf;
-    size_t remaining = buf_size;
 
     OE_CHECK(serialize_element(
         "version",
@@ -209,12 +208,11 @@ oe_result_t oe_deserialize_eeid(
     oe_eeid_t* eeid)
 {
     oe_result_t result = OE_UNEXPECTED;
+    const char** position = &buf;
+    size_t remaining = buf_size;
 
     if (!buf || !buf_size || !eeid)
         OE_RAISE(OE_INVALID_PARAMETER);
-
-    const char** position = &buf;
-    size_t remaining = buf_size;
 
     memset(eeid, 0, sizeof(oe_eeid_t));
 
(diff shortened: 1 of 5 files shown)


enclave/crypto/sha.c
```diff
@@ -121,7 +121,7 @@ oe_result_t oe_sha256_restore(
 
     uint64_t NB = ((((uint64_t)num_hashed[1]) << 32) + num_hashed[0]) / 8;
     impl->ctx.total[0] = NB & 0xFFFFFFFF;
-    impl->ctx.total[1] = (NB >> 32) & 0xFFFFFFFF;
+    impl->ctx.total[1] = ((uint32_t)(NB >> 32)) & 0xFFFFFFFF;
```
CodeMonkeyLeet, line 124: For my understanding, why is this necessary, and also, why is it not necessary for the line above this? In either case, if we're cast truncating to 32-bits, it seems the extra bit masking operation is unnecessary.

common/sgx/eeid.c
```diff
@@ -132,13 +132,12 @@ static oe_result_t deserialize_element(
 oe_result_t oe_serialize_eeid(const oe_eeid_t* eeid, char* buf, size_t buf_size)
 {
     oe_result_t result = OE_UNEXPECTED;
+    char** position = &buf;
```
CodeMonkeyLeet, line 135: This is an unsafe pointer dereference because the null check happens after this line now.

common/sgx/eeid.c
```diff
@@ -209,13 +208,12 @@ oe_result_t oe_deserialize_eeid(
     oe_eeid_t* eeid)
 {
     oe_result_t result = OE_UNEXPECTED;
+    const char** position = &buf;
```
CodeMonkeyLeet, line 211: Similarly, unsafe pointer dereference.

common/sgx/eeid.c
```diff
@@ -794,21 +800,20 @@ oe_result_t oe_eeid_evidence_ntoh(
     oe_eeid_evidence_t* evidence)
 {
     oe_result_t result = OE_UNEXPECTED;
+    const uint8_t* position = buffer;
+    size_t remaining = buffer_size, data_size;
```
CodeMonkeyLeet, line 804: Style: please split variable declarations with initializations apart from those without for readability.

```suggestion
    size_t remaining = buffer_size;
    size_t data_size = 0;
```

common/sgx/eeid.c
```diff
@@ -765,21 +772,20 @@ oe_result_t oe_eeid_evidence_hton(
     size_t buffer_size)
 {
     oe_result_t result = OE_UNEXPECTED;
+    uint8_t* position = buffer;
+    size_t remaining = buffer_size, data_size;
```
CodeMonkeyLeet, line 776:
```suggestion
    size_t remaining = buffer_size;
    size_t data_size = 0;
```

common/sgx/eeid.c
```diff
@@ -390,6 +388,11 @@ static oe_result_t verify_base_image_signature(const sgx_sigstruct_t* sigstruct)
     oe_result_t result = OE_UNEXPECTED;
     unsigned char buf[sizeof(sgx_sigstruct_t)];
     size_t n = 0;
+#ifndef OE_BUILD_ENCLAVE
```
CodeMonkeyLeet, line 391: Consider splitting this function into dedicated subfunctions for mbedtls and OpenSSL handling; the interleaved `#ifdefs` make the code increasingly hard to follow.

Side note: it looks like the mbedtls section of the code should have the same compiler issue, but may not be flagged because it is not being built with GCC in your current environment?

enclave/crypto/sha.c
wintersteiger: It's necessary because GCC complains (clang doesn't). I think the issue is the automatic promotion of `(NB >> 32)` to a signed type, but I didn't investigate it any further.

common/sgx/eeid.c
wintersteiger: There's no dereference here, it just takes the address of the argument, which should always be safe, even if `buf` is `NULL`.

common/sgx/eeid.c
wintersteiger: (see above)

common/sgx/eeid.c
wintersteiger: This fixes both mbedTLS and OpenSSL, but  yeah, I think the time has come to split that up :-)

wintersteiger: (done)

wintersteiger:
bors try

CodeMonkeyLeet:
bors try

wintersteiger:
The bors build was successful, but GitHub says it's still waiting for the status to be reported. Is there something wrong with the infrastructure?

anakrish:
bors try

wintersteiger:
bors try

wintersteiger:
Bors was successful a second time, but the status is still "expected". Is there something wrong with bors or github?

radhikaj:
> Bors was successful a second time, but the status is still "expected". Is there something wrong with bors or github?

@BRMcLaren  @rs-- Do you have insights here?

radhikaj:
bors try

BRMcLaren:
We had unexpected down time and our monitoring didn't catch it over the weekend as it was primarily web server related versus functionality. Lots of unanswered web hooks.

Sorry about that.

wintersteiger:
I had to manually navigate to Blue Ocean to see the build logs and it looks like there is some sort of problem, but I have no idea what the hudson exception means. (Also, there seems to have been a problem with mail delivery.)

```
100% tests passed, 0 tests failed out of 265
...
Caught global pipeline exception: hudson.AbortException: script returned exit code 1
[Pipeline] emailext
Sending email to: cwinter@microsoft.com
Connection error sending email, retrying once more in 10 seconds...
Connection error sending email, retrying once more in 10 seconds...
Failed after second try sending email
[Pipeline] End of Pipeline
ERROR: script returned exit code 1
Finished: FAILURE
```

anakrish:
bors try

wintersteiger:
The latest build reports this:

```
fatal: [localhost]: FAILED! => {"changed": false, "dest": "/tmp/shellcheck.tar.xz", "elapsed": 0, "msg": "Request failed", "response": "HTTP Error 404: Not Found", "status_code": 404, "url": "https://storage.googleapis.com/shellcheck/shellcheck-v0.4.7.linux.x86_64.tar.xz"}
```

I'll try again later.

CodeMonkeyLeet:
@wintersteiger Should have been fixed by #3344.

bors try

Issue #3015, opened 2020-05-11
Update install-windows-prereqs.ps1 to pull Intel Windows SGX packages from cache


CodeMonkeyLeet:
One of the issues causing instability in the contributor/user setup instructions on a Windows device is the fact that the install-windows-prereqs.ps1 helper script targets the Intel Developer Zone download links to acquire the SGX dependencies on Windows. This has the following issues:

* Intel Developer Zone does not provide the ability to target a specific instance of the the target package by instance, so updating of a package breaks links to old instances.
* These links are usually CDN links and are brittle, breaking when Intel pushes updates to the file without re-versioning it (for example, to fix errors in the published package).

The mitigate these issues, we should host the specific version dependencies of the packages from a stable endpoint, which will also allow us to maintain multiple versions of the packages (e.g. one for latest v0.x.x release branch, and another for master branch).

* Create a public storage container in a cloud provider and cache the required versions of the Intel Windows SGX packages there.
* Per legal review, we will also need to update docs around use of the install-windows-prereqs.ps1 script and its help text to indicate that its use implies consent to the user agreement to the Intel EULA for those packages.

radhikaj:
Francisco will add links to WU Intel packages for WS2016 and WS2019/Windows 10.

fchinchilla:
# Server 2016
* Base INF
  * Generic Search https://www.catalog.update.microsoft.com/Search.aspx?q=SgxLCDevice . If you click on any of the downloads there is a link to download a specific file in the format GUID_SHA1HASH.cab
  * 2.7.100 http://download.windowsupdate.com/d/msdownload/update/driver/drvs/2020/04/8e78fd6b-efeb-4952-ab0d-945e61c164ba_f46d7480dc60858df4291eb20cc088183e6d09c7.cab
* PSW
  * Not currently available on WU
* DCAP
  * Generic Search https://www.catalog.update.microsoft.com/Search.aspx?q=SgxLCDevice_DCAP
  * 1.6 http://download.windowsupdate.com/d/msdownload/update/driver/drvs/2020/05/08cc8440-9f38-4635-9685-cdbf476666fa_92188002588d6b5fa58814bc92b78ad33ddf11af.cab 

# Server 2019
* Base INF
  * Generic Search https://www.catalog.update.microsoft.com/Search.aspx?q=ACPI\INT0E0C
[...]

Issue #3000, opened 2020-05-06
Consider replacing install-prereqs script with well-tested package managers


anakrish:
Currently, we have a power-shell script [install-windows-prereqs.ps1](https://github.com/openenclave/openenclave/blob/master/scripts/install-windows-prereqs.ps1) that installs the necessary components to set up a Windows development environment.

The script has the advantage that
- it checks the hash of any package/software it installs
- it fetches packages only from Azure endpoints and avoids 3rd party sites

There are some issues with this script that we have run into:
- The script can fail unexpectedly
- It redownloads packages 
- There is no easy way to install only specific packages
- It is not thoroughly tests like a dedicated package manager

Can we consider moving to other package managers like chocolately, scoop etc?

anakrish:
CC: @BRMcLaren @CodeMonkeyLeet @radhikaj @jianlia

dthaler:
> The script has the advantage that
> * it fetches packages only from Azure endpoints and avoids 3rd party sites

I'm not sure it's an "advantage" for an open source project like this to fetch packages only from Azure.  For an Azure script that's ok.   For folks who don't use Azure it might not be considered a feature.   If the intent is to only use Azure, then it should have azure in the filename and instructions to make it clear.

anakrish:
@dthaler That's a good point.

dthaler:
The phrase "Fetches packages from stable endpoints that we have control over" in the PR description still begs the question of who "we" refers to.   Do you mean 
a) that the person running the script has control over?  (No)
b) that the current OE SDK maintainers (which may change affiliations over time, and for whom Azure might or might not be desired) have control over?
c) that Azure has control over?
d) something else?

CodeMonkeyLeet:
It's not clear to me what action will resolve this issue: is it just an acknowledgement that an investigation has taken place? It suggests a task, but not the motivation or success criteria?

Based on the the title of the issue, I'd say that the premise of _replacing_ install-windows-prereqs.ps1 with the examples of chocolatey/scoop misunderstands the role of install-windows-prereqs.ps1. Consider the comparison to what is done on Linux:

* Apt/yum are package managers on Linux, and they are analogous to chocolatey/scoop on Windows.
* Ansible is the only environment configuration tool we use on Linux, but on Windows, that configuration setup is consolidated through the install-windows-prereqs.ps1 script (and simply invoked through Ansible from our automation). 
[...]

PR #3251, opened 2020-07-10
 Update Intel PSW/DCAP Installation

rs--:
Update Windows install script and docs to reflect Intel packaging update

Signed-off-by: Rob Sanchez rosan@microsoft.com

### DIFF
--- a/docs/GettingStartedDocs/Contributors/WindowsManualSGX1FLCDCAPPrereqs.md
+++ b/docs/GettingStartedDocs/Contributors/WindowsManualSGX1FLCDCAPPrereqs.md
@@ -1,6 +1,6 @@
 # SGX1 with Flexible Launch Control (FLC) Prerequisites on Windows
 
-## [Intel Platform Software for Windows (PSW) v2.7](http://registrationcenter-download.intel.com/akdlm/irc_nas/16607/Intel%20SGX%20PSW%20for%20Windows%20v2.7.101.2.exe)
+## [Intel Platform Software for Windows (PSW) v2.8](http://registrationcenter-download.intel.com/akdlm/irc_nas/16766/Intel%20SGX%20PSW%20for%20Windows%20v2.8.100.2.exe)
 
 The PSW only needs to be manually installed if you are running on Windows Server
 2016 or a version of Windows client lower than 1709. It should be installed automatically
@@ -12,30 +12,21 @@ To install the PSW on Windows Server 2016 and Windows client < 1709, unpack the 
 ZIP executable, and run the installer under `PSW_EXE_RS2_and_before`:
 
 ```cmd
-"C:\Intel SGX PSW for Windows v2.7.101.2\PSW_EXE_RS2_and_before\Intel(R)_SGX_Windows_x64_PSW_2.7.101.2.exe"
+"C:\Intel SGX PSW for Windows v2.8.100.2\PSW_EXE_RS2_and_before\Intel(R)_SGX_Windows_x64_PSW_2.8.100.2.exe"
 ```
 
 If you would like to manually update the PSW on Windows Server 2019 or Windows
-clients > 1709 without relying on Windows Update, you can do
-so as follows:
+clients > 1709 without relying on Windows Update, you can update the PSW components
+as follows:
 
-1. From the extracted PSW package, unzip the `PSW_INF_RS3_and_above\component\Signed_*.zip` file.
+1. Install `devcon.exe`, available as part of the [Windows Driver Kit for Windows 10](https://go.microsoft.com/fwlink/?linkid=2026156).
+   -  Note that `devcon.exe` is usually installed to `C:\Program Files (x86)\Windows Kits\10\tools\x64`
+   which is not in the `PATH` environment variable by default.
 
-   The following instructions will assume that it was unzipped to the `PSW_INF` folder.
-
-2. Install the PSW components:
-    - `devcon.exe` from the [Windows Driver Kit for Windows 10](
-      https://go.microsoft.com/fwlink/?linkid=2026156) can be used to update
-      the drivers from an elevated command prompt:
-
-    - Note that `devcon.exe` is usually installed to `C:\Program Files (x86)\Windows Kits\10\tools\x64`
-      which is not in the `PATH` environment variable by default.
-
-    - The hash values in the path may be different from the example command, please update as needed.
-
-    ```cmd
-    devcon.exe update PSW_INF\drivers\48e7c1e9-6de8-46ee-8ff9-46aa7b7ee5b9\sgx_psw.inf "SWC\VEN_INT&DEV_0E0C"
-    ```
+2. In an elevated command prompt, run the following command from the extracted PSW package under the `PSW_INF_RS3_and_above` folder:
+  ```cmd
+  devcon.exe update sgx_psw.inf "SWC\VEN_INT&DEV_0E0C"
+  ```
 
 You can verify that the correct version of Intel SGX PSW is installed by using
 Windows Explorer to open `C:\Windows\System32`. You should be able to find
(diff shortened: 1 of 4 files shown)


docs/GettingStartedDocs/Contributors/WindowsManualSGX1FLCDCAPPrereqs.md
```diff
@@ -12,30 +12,24 @@ To install the PSW on Windows Server 2016 and Windows client < 1709, unpack the
...
-
-    - The hash values in the path may be different from the example command, please update as needed.
-
-    ```cmd
-    devcon.exe update PSW_INF\drivers\48e7c1e9-6de8-46ee-8ff9-46aa7b7ee5b9\sgx_psw.inf "SWC\VEN_INT&DEV_0E0C"
-    ```
+- From the `PSW_INF_RS3_and_above` folder, run the command:
+```cmd
+devcon.exe update sgx_psw.inf "SWC\VEN_INT&DEV_0E0C"
+```
```
CodeMonkeyLeet, line 32:
````suggestion

1. Install `devcon.exe`, available as part of the [Windows Driver Kit for Windows 10](
  https://go.microsoft.com/fwlink/?linkid=2026156).
    -  Note that `devcon.exe` is usually installed to `C:\Program Files (x86)\Windows Kits\10\tools\x64`
    which is not in the `PATH` environment variable by default.
2. In an elevated command prompt, run the following command from the extracted PSW package under the `PSW_INF_RS3_and_above` folder:
  ```cmd
  devcon.exe update sgx_psw.inf "SWC\VEN_INT&DEV_0E0C"
  ```
````

Comment (CodeMonkeyLeet) [reply docs/GettingStartedDocs/Contributors/WindowsManualSGX1FLCDCAPPrereqs.md:32]: Note that these changes also need to be mirrored to the WindowsManualSGX1Prereqs.md file as well.

File: docs/GettingStartedDocs/Contributors/WindowsManualSGX1FLCDCAPPrereqs.md [hunk 2]
```diff
@@ -96,27 +80,28 @@ nuget.exe install EnclaveCommonAPI -ExcludeVersion -Source "C:\Intel SGX DCAP fo
 
     - If the driver is already installed and running, the machine will need to be rebooted for the change to take effect.
 
-3. Install or update the drivers:
+2. Install or update the drivers:
     - Refer to the PSW section above for notes on acquiring and using `devcon.exe`.
     - The hash values in the paths may be different from the example commands, please update them as needed.
```
CodeMonkeyLeet, line 85: This recommendation is no longer necessary, remove. 
```suggestion
```

scripts/install-windows-prereqs.ps1
```diff
@@ -14,16 +14,16 @@ Param(
     [string]$VSBuildToolsHash = '',
     [string]$Clang7URL = 'http://releases.llvm.org/7.0.1/LLVM-7.0.1-win64.exe',
     [string]$Clang7Hash = '672E4C420D6543A8A9F8EC5F1E5F283D88AC2155EF4C57232A399160A02BFF57',
-    [string]$IntelPSWURL = 'http://registrationcenter-download.intel.com/akdlm/irc_nas/16607/Intel%20SGX%20PSW%20for%20Windows%20v2.7.101.2.exe',
-    [string]$IntelPSWHash = 'AF669A4593411E9AABCE18838C91003866DDDEDAC5BEEC61DE160025008B0A19',
+    [string]$IntelPSWURL = 'http://registrationcenter-download.intel.com/akdlm/irc_nas/16766/Intel%20SGX%20PSW%20for%20Windows%20v2.8.100.2.exe',
```
CodeMonkeyLeet, line 17: Instead of continuing to use the Intel registration center links here, can we address #3015 as well by using the Microsoft Update links? We might want to reach out to Francisco about this since I don't know whether the latest packages have already been ingested to WU.

scripts/install-windows-prereqs.ps1
```diff
@@ -528,54 +528,45 @@ function Install-DCAP-Dependencies {
...
                     'description' = 'Intel(R) Software Guard Extensions Launch Configuration Service'
                 }
                 'sgx_dcap' = @{
-                    'zip_path'    = "$PACKAGES_DIRECTORY\Intel_SGX_DCAP\Intel SGX DCAP for Windows *\DCAP_INF\${OS_VERSION}\Signed_*.zip"
+                    'path'        = "$PACKAGES_DIRECTORY\Intel_SGX_DCAP\Intel*SGX*DCAP*\dcap\WindowsServer2019_Windows10"
                     'location'    = 'root\SgxLCDevice_DCAP'
                     'description' = 'Intel(R) Software Guard Extensions DCAP Components Device'
                 }
             }
             'Win10' = @{
```
CodeMonkeyLeet, line 553: Is the Win10 definition redundant now seeing as it's identical to the WS2019 one? I'm also a little nervous that the Win10 detection in Get-WindowsRelease only covers two specific versions and probably doesn't capture all the new versions since that logic was originally written.

scripts/install-windows-prereqs.ps1
CodeMonkeyLeet: Follow-up: I did a search of available WU packages and didn't see the updated package versions, feel free to ignore this suggestion unless you've managed to get the links from Francisco et. al.

rs--: Thanks for checking that out @CodeMonkeyLeet :)

scripts/install-windows-prereqs.ps1
rs--: Any suggestions on how we should address this?

BRMcLaren: We could just merge them into a single config `Win10_2019`.

BRMcLaren:
Set up 2016 and 2019 machines with the scripts, no issues

scripts/install-windows-prereqs.ps1
rs--: @BRMcLaren Yeah, I think that should be done, I was more curious about 

> I'm also a little nervous that the Win10 detection in Get-WindowsRelease only covers two specific versions

BRMcLaren: Maybe add this to the use case requirements of this [issue](https://github.com/openenclave/openenclave/issues/3000). Then when we tackle this in the next month or so completely refactor this script to treat less as a silver bullet.

The various windows versions include a server version, and I think we might wanna break that out into the CI use case versus a developer situation.

BRMcLaren:
bors try

scripts/install-windows-prereqs.ps1
CodeMonkeyLeet: To clarify the comment, I'm specifically referring to this code:

```powershell
function Get-WindowsRelease {
    $releases = @{
        18363 = "Win10"
        18362 = "Win10"
        17763 = "WinServer2019"
        14393 = "WinServer2016"
    }
...
```

It seems like our handling of Win10 is very narrow to those two versions for some reason. Realistically, the only thing that the script seems to care about is whether it's using the RS3 (16299) and up versions of Windows or not, so this script could be made less brittle in its current form by checking for the whether we are above or below that version instead of trying to split this into fixed versions (of which "Win10" spans a very wide range up to the current 19041).

Moving forwards in rethinking the script, I think it's an unscalable proposition to enumerate and test every version of Win10 that a dev could run on anyway, and we would likely simply ensure that it works with the latest Win10 client as a sanity test, with the expectation that the install instructions for WS2019 should apply to it.

radhikaj: What is the min version that should be checked?

CodeMonkeyLeet: @radhikaj We can probably keep 14393 as the min version; that's Win10 RS1 and also WS2016, IIRC.

PR #3257, opened 2020-07-13
Fine tune OCALL opt-out granularity.

### LINKED ISSUE #3254: Cannot opt-out of sgx/thread.edl if --gc-sections is not used
`oe_sgx_thread_wake_wait_ocall` does not have a weak implementation and the symbol gets pulled in during linking via `enclave/core/sgx/thread.c`, which contains `_thread_wake_wait` and which calls `oe_sgx_thread_wake_wait_ocall`.  `--gc-sections` would avoid this (as it would eliminate the pthread chain) but it shouldn't be a requirement for opt-out, since it is not a requirement for all the other ocalls.

### LINKED ISSUE #3255: Cannot opt-out of memory.edl if --gc-sections is not used
This issue is similar to https://github.com/openenclave/openenclave/issues/3254. Note that there are no other opt-out issues related to `--gc-sections` at the moment.

`oe_realloc_ocall` does not have a weak implementation and the symbol gets pulled in during linking via `enclave/core/hostcalls.c`, which contains `oe_host_realloc` and which calls `oe_realloc_ocall`. `--gc-sections` would avoid this but it shouldn't be a requirement for opt-out, since it is not a requirement for all the other ocalls.

anakrish:
Previously, even if an enclave does not use a particular ocall,
the ocall will still exist in the host binary since the ocall implementation
was placed in the same file as some other symbol that the application (host) needs.

To solve this, OCALLS for some core system edls are placed into separate
c files whereever possible. This ensures that if an enclave does not need an ocall,
the implementation of the ocall will not be picked up.

Note: Weak Symbols (OE_WEAK_ALIAS) is not a viable solution since the linker will
still link in the weak symbol if it is placed in a file that contains another symbol
needed by the application.

fixes #3254
fixes #3255

Note: Not all system ocall implementations have been moved to separate C files.


anakrish:
bors try

anakrish:
@letmaik Can your see if this PR fixes your issue?

letmaik:
The issues I was seeing were for the enclave side, which is what I'm mainly interested in. I think in this PR you touched the host side.

anakrish:
@mingweishih Can you see if this PR fits in which how you want to approach the future ghost ocalls work.

radhikaj:
@anakrish Please resolve conflict and then merge.

radhikaj:
bors try

PR #3258, opened 2020-07-13
Weak symbols for two system ocalls.

### LINKED ISSUE #3254: Cannot opt-out of sgx/thread.edl if --gc-sections is not used

### LINKED ISSUE #3255: Cannot opt-out of memory.edl if --gc-sections is not used

anakrish:
Weak implementations for
oe_sgx_thread_wake_wait_ocall and
oe_realloc_ocall

fixes #3254
fixes #3255

### DIFF
--- a/docs/SystemEdls.md
+++ b/docs/SystemEdls.md
@@ -37,7 +37,7 @@ oe_write_ocall | N/A | Required by internal APIs/macros such as `oe_host_printf`
 ### memory.edl
 Ocall | Dependent Public APIs | Comments |
 :---|:---:|:---|
-oe_realloc_ocall | oe_host_realloc | Required by OP-TEE. |
+oe_realloc_ocall | oe_host_realloc | _ |
 
 ## Syscall system EDLs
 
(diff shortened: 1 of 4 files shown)


anakrish:
@letmaik Can you review this one?

anakrish:
bors try

anakrish:
> Does this also remove the need for the comments in optee/platform.edl and SystemEdls.md regarding the OP-TEE hard dependency on memory.edl? It would be nice to clean those up as well as part of this change.

Done. Only tweaked comments for now. Next release the edls themselves will be refactored.

PR #3115, opened 2020-06-09
Add retry logic during dependencies installation

rs--:
Issue #3101

Automated builds were failing, among other reasons, due to dependencies installation. Based on investigations failures were not linked to one problem, which include, but are not limited to issues connecting to the repositories, or package installation lock issues.

Adding retry logic might not be the most elegant solution, but it should help mitigate false negative builds.

Signed-off-by: Rob Sanchez <rosan@microsoft.com>

PR #3259, opened 2020-07-13
Add retry loop to linux fresh install on jenkins

### LINKED ISSUE #3101: Testing:  Sporadic failures in E2E tests
Some instances:
https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAzure-Linux/detail/Azure-Linux/609/pipeline

```bash
TASK [linux/openenclave : Install all the Open Enclave prerequisites YUM packages for development] *** 
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to synchronize cache for repo 'epel'", "rc": 1, "results": []}
```

https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAzure-Linux/detail/Azure-Linux/570/pipeline
```bash
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to download packages: Curl error (28): Timeout was reached for https://mirrors.fedoraproject.org/metalink?repo=epel-8&arch=x86_64&infra=$infra&content=$contentdir [Connection timed out after 30000 milliseconds]", "results": []}
```
https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAzure-Linux/detail/Azure-Linux/603/pipeline
```bash
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?"], "stdout": "", "stdout_lines": []
```

BRMcLaren:
Fixes #3101 

We are seeing rhel package installation timeouts during the server flickes. Fix by adding a retry loop on installation.

Signed-off-by: Brett McLaren <brmclare@microsoft.com>

### DIFF
--- a/.jenkins/pipelines/Azure/Linux/Jenkinsfile
+++ b/.jenkins/pipelines/Azure/Linux/Jenkinsfile
@@ -68,8 +68,18 @@ def ACCTest(String label, String compiler, String build_type, List extra_cmake_a
                 checkout scm
                 if (fresh_install) {
                     sh  """
-                        sudo bash scripts/ansible/install-ansible.sh
-                        sudo \$(which ansible-playbook) scripts/ansible/oe-contributors-acc-setup.yml
+                        # Install ansible
+                        for i in 1 2 3 4 5
+                        do
+                            sudo scripts/ansible/install-ansible.sh && break
+                            sleep 15
+                        done
+                        # Run ACC Playbook
+                        for i in 1 2 3 4 5
+                        do
+                            sudo \$(which ansible-playbook) scripts/ansible/oe-contributors-acc-setup.yml && break
+                            sleep 15
+                        done
                         """
                 }
                 def task = """


BRMcLaren:
bors try

BRMcLaren:
bors try cancel

BRMcLaren:
bors try-

BRMcLaren:
bors try-

.jenkins/pipelines/Azure/Linux/Jenkinsfile
```diff
@@ -68,8 +68,18 @@ def ACCTest(String label, String compiler, String build_type, List extra_cmake_a
...
+                        for i in 1 2 3 4 5
+                        do
+                            sudo scripts/ansible/install-ansible.sh && break
+                            sleep 15
+                        done
+                        # Run ACC Playbook
+                        for i in 1 2 3 4 5
+                        do
+                            sudo \$(which ansible-playbook) scripts/ansible/oe-contributors-acc-setup.yml && break
+                            sleep 15
```
anakrish, line 81: This looks good to me. I'm however curious if there is a way to get ansible itself to keep retrying a task until it succeeds.
The failure seems not specific to OE project and hence I'm curious if ansible itself provides a solution.

BRMcLaren:
bors try

BRMcLaren:
bors try

BRMcLaren:
multiple retries to validate

BRMcLaren:
bors try

CodeMonkeyLeet (commented):
A couple of questions for my understanding:
* Is the ansible install script and playbook execution idempotent? The retries seem to assume that any failure in those operations don't leave any state on the target that prevent reruns from succeeding. 
* Related to Anand's question, can this be dealt with at the Ansible level? I ask because the manual setup instructions for contributors ask them to run the playbook, so they don't benefit from retries at the Jenkins level.

Side note: I also noted in ~~#3103~~ #3101 that the PR #3115 missed adding retry logic to /scripts/ansible/roles/linux/intel/tasks/redhat/sgx-driver-requirements.yml, which is still included via sgx-driver.yml in playbooks such as oe-contributors-acc-setup.yml? Given that the sgx-driver tasks are included before the sgx-package ones, it seems like we could be hitting the version of the task without retries first, so sgx-driver-requirements.yml needs to add the same retry parameters as sgx-packages-install.yml. 

I'll also note that this weird overlap also does not happen in the Ubuntu version of sgx-driver-requirements.yml, so this is potentially a bigger bug. It looks like the redhat sgx-driver-requirements.yml is setting up EPEL repo so that it can also pick up the dkms module for RHEL, like we do on Ubuntu, but for whatever reason, it shares the `intel_sgx_package_dependencies` with sgx-packages-install.yml, whereas in the Ubuntu version, they install different dependencies.

Edit: Fixed issue number.

.jenkins/pipelines/Azure/Linux/Jenkinsfile
achamayou: Ansible does have a mechanism for retries: https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#retrying-a-task-until-a-condition-is-met

BRMcLaren: Ansible does have a mechanism for retries, which we do [implement ](https://github.com/openenclave/openenclave/blob/4181f16725c02f525fd82e6a781146b02fdeeffa/scripts/ansible/roles/linux/intel/tasks/redhat/sgx-packages-install.yml#L9). and then exits when succeeds.

The prior comment can be interpreted 2 ways, one of which is the above and the other is to retry indefinitely. Retry forever? Nope, retry x amount of times, yes.

radhikaj:
@BRMcLaren Thanks a lot for this change. Really appreciate it. Can we address @CodeMonkeyLeet 's questions so this can be merged? Thanks!

BRMcLaren:
@CodeMonkeyLeet 

Sorry for the delay. I am reopening this change under another [PR](https://github.com/openenclave/openenclave/pull/3485) as I am looking flakey tests.

 >Is the ansible install script and playbook execution idempotent? The retries seem to assume that any failure in those operations don't leave any state on the target that prevent reruns from succeeding.

Playbook execution is, the install script upon testing was not. I have modified the commit to only do the ansible playbook execution as we are yet to see an ansible installation failure in Jenkins testing. 

> Related to Anand's question, can this be dealt with at the Ansible level? I ask because the manual setup instructions for contributors ask them to run the playbook, so they don't benefit from retries at the Jenkins level.

Yes. It actually is dealt with at the ansible level (I see your comment regarding the weird overlap, looking into it). The problem specifically in jenkins is for long timeouts such as package [repo](https://github.com/openenclave/openenclave/issues/3405) inconsistencies. Ansible will force install these so if there is an issue, the solution the many come to is "just retry forever" at the ansible level but it doesn't work out as well as one would hope as that feature doesn't exist. Having a wrapper in the Jenkins testing that pauses and retries a few times seems to be a good middle man for cutting down on the installation issues from a testing standpoint.

Given this, what we can do for users is add another script that is a wrapper around ansible and the installation script that encapsulates retries and that would be a reasonable enough solution. That being said, I haven't seen a user have issues and not instantly just retry.

Would you prefer having a separate script that Jenkins and users can use that encapsulates this functionality? Due to technical limitations, I am unsure of another solution.

CodeMonkeyLeet:
@BRMcLaren No, I would not suggest trying to solve this with yet another layer of indirection. I mentioned trying to solve this at the ansible level because:
* The recurrence of the flaky failures I've noticed were due to the same failures that #3115 was supposed to address (the recent #3405 notwithstanding).
* I observed that the fix #3115 did not fully apply ansible retry logic to all codepaths that could result in package download failures.
    * I am suggesting that we complete that fix in the ansible tasks and evaluate them, which would benefit general users of the ansible playbook that going directly to a jenkins level retrying the entire playbook again would not.
    * For example, we don't add a `delay` with the retries in the Ansible scripts, which reduces their value since we know that some of these sporadic failures can be avoided by waiting for the system (e.g. package server metadata mirroring) to settle before retrying.

BRMcLaren:
@CodeMonkeyLeet That's a valid point. My concern with it is that it imposes additional testing overhead and the sad reality is I am lacking the bandwidth. Would an acceptable solution be to add an issue for this (change from fixing at Jenkins level to Ansible level), finish up my flakey test investigation, and then try to get to it? 

My core priority is cutting down on the testing flakes not so much the general case so the quick fix seems appealing to me as moving forward the ideal user workflow is just to pull down a prebuilt container dev environment.

PR #3261, opened 2020-07-14
Update Changelog to add reference to issue#3250

radhikaj:


CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (docs/DesignDocs/system_ocall_opt_in.md#how-to-port-your-application) for more information.
```
CodeMonkeyLeet, line 39: Nit: GitHub markdown doesn't see this as a link because there is a line break between the `](` ... the line break can be inserted after the opening parenthesis on the previous line.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   - Note: SDK users would need to import logging.edl to enable logging. Logging is disabled by default.
   - See [System edls](docs/SystemEdls.md) for list of all edls and associated OCalls.
+  - See #3250 which is an issue in v0.10.0 due to which the same host cannot load enclaves created with different edls. A work around for this is to create a single edl and stub out appropriate calls for each enclave. This will be fixed in v0.11.0. See #3086 for more details.
```
CodeMonkeyLeet, line 42: I'd move this up the list of Changed items for visibility.
```suggestion
  - A known issue is that different enclaves importing functions from System EDLs cannot be loaded by the same host app unless all of the functions were imported with exactly the same ordinals. See #3250 for details. This will be addressed in the next release based on design proposal #3086. 
  - A workaround for this issue in the meantime is to define a standard import EDL for any enclaves that need to be loaded into the same host app. Ensuring this shared EDL is then the first import in each enclave's EDL will result in the common imports being assigned the same ordinals in each resulting enclave.
```

CHANGELOG.md
radhikaj: Thanks!

radhikaj:
Thanks @CodeMonkeyLeet @nonpolarity @BRMcLaren

Issue #3262, opened 2020-07-14
Fragment test


achamayou:
In relation to #3256 I've been looking at the memory fragment test, because it makes use of dlmallinfo(), which is dlmalloc-specific.

It defines __get_heap_size as dlmallinfo().keepcost: https://github.com/openenclave/openenclave/blob/master/tests/memory/enc/fragment.c#L34

That's:

>   keepcost:  the maximum number of bytes that could ideally be released
>                back to system via malloc_trim. ("ideally" means that
>                it ignores page restrictions etc.)

>   malloc_trim(size_t pad);
> 
>   If possible, gives memory back to the system (via negative arguments
>   to sbrk) if there is unused memory at the `high' end of the malloc
>   pool or in unused MMAP segments. You can call this after freeing
>   large blocks of memory to potentially reduce the system-level memory
>   requirements of a program. However, it cannot guarantee to reduce
>   memory. Under some allocation patterns, some large free blocks of
>   memory will be locked between two used chunks, so they cannot be
>   given back to the system.

So that's memory held by the allocator that isn't actively in use, and could conceivably be released. It's neither the whole available space for the heap, nor the actual heap size held by the allocator or handed out.

The test does use it as if it was though, eg. https://github.com/openenclave/openenclave/blob/master/tests/memory/enc/fragment.c#L107

So I think there are two questions:

1. Should the test be modified to use fordblks/uordblks instead?
2. Should there be at least an internal API exposing the total/in-use amounts, which this test can target, and which effectively becomes part of the pluggable allocator API?

CodeMonkeyLeet:
I'll also note that tests/bigmalloc has a similar issue where it cannot run with snmalloc because it assumes an `sbrk()` based allocator to infer remaining heap size. It may be worth thinking about a standardized `oe_allocator_mallinfo` contract with pluggable allocators to be able to standardize these tests.

achamayou:
Yes, good point. There has been some prior discussion about this, although in the context of exposing this to users instead https://github.com/openenclave/openenclave/issues/3097. May I propose an `oe_allocator_mallinfo`, to be exposed internally to tests only? Is the best way to do that a separate design document, or a PR against the existing pluggable allocator design document?

PR #3256, opened 2020-07-13
Update snmalloc to 0.5.0 release

achamayou:
This updates snmalloc to the latest release, which brings a number of improvements (listed under https://github.com/microsoft/snmalloc/releases/tag/0.5.0), and in particular lowers initial allocation.

I've run all the Open Enclave tests locally, and they pass provided the initial allocation is adjusted to be at least 64 + 64 * NumTCS (in pages), which isn't the case currently (there are several cases that set 128 pages of heap but have 16 TCS). I've also moved CCF over to use the latest OE with snmalloc 0.5 using the pluggable allocator, and all tests are passing there too (https://github.com/microsoft/CCF/pull/1400).

I am preparing a follow up PR that allows switching the Open Enclave tests more easily (I used the USE_SNMALLOC flag with some exceptions for uses of dlmallinfo), adjusts enclave sizes and removes the USE_SNMALLOC flag, but I would like to get this merged in time for 0.10 if possible, so I will file that separately if that's ok.

PR #3263, opened 2020-07-14
Enable tests with snmalloc

achamayou:
Follow-up to #3256 and to the previous effort #2730. This PR contains two sets of changes:

1. Enclave size adjustments, to fit snmalloc's initial memory requirements. These changes are much smaller than in #2730.
2. Enable a subset of CI configurations for Linux and Windows (PR and Full). I'd appreciate a review of these configuration choices, to ensure they are completely judicious.

There's an open question under #3262, which I am planning to address separately from this PR because it needs a design proposal.

### DIFF
--- a/.jenkins/pipelines/Agnostic/Linux/Jenkinsfile
+++ b/.jenkins/pipelines/Agnostic/Linux/Jenkinsfile
@@ -110,19 +110,22 @@ try{
     if(FULL_TEST_SUITE == "true") {
         stage("Full Test Suite") {
             testing_stages += [
-                "Sim 1804 clang-7 SGX1 Debug":      { simulationContainerTest('18.04', 'Debug',   ['-DHAS_QUOTE_PROVIDER=OFF', '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
-                "Sim 1804 clang-7 SGX1 Release":    { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF', '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
-                "Sim 1804 clang-7 SGX1FLC Debug":   { simulationContainerTest('18.04', 'Debug',   ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
-                "Sim 1804 clang-7 SGX1FLC Release": { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) }
+                "Sim 1804 clang-7 SGX1 Debug":               { simulationContainerTest('18.04', 'Debug',   ['-DHAS_QUOTE_PROVIDER=OFF', '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
+                "Sim 1804 clang-7 SGX1 Release":             { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF', '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
+                "Sim 1804 clang-7 SGX1FLC Debug":            { simulationContainerTest('18.04', 'Debug',   ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
+                "Sim 1804 clang-7 SGX1FLC Release":          { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
+                "Sim 1804 clang-7 SGX1FLC Debug snmalloc":   { simulationContainerTest('18.04', 'Debug',   ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF', '-DUSE_SNMALLOC=ON']) },
+                "Sim 1804 clang-7 SGX1FLC Release snmalloc": { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF', '-DUSE_SNMALLOC=ON']) },
             ]
             parallel testing_stages
         }
     } else {
         stage("PR Testing") {
             testing_stages += [
-                "Sim 1804 clang-7 SGX1 Release":    { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF', '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=ON']) },
-                "Sim 1804 clang-7 SGX1FLC Debug":   { simulationContainerTest('18.04', 'Debug',   ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=ON']) },
-                "Sim 1804 clang-7 SGX1FLC Release": { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=ON']) }
+                "Sim 1804 clang-7 SGX1 Release":             { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF', '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=ON']) },
+                "Sim 1804 clang-7 SGX1FLC Debug":            { simulationContainerTest('18.04', 'Debug',   ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=ON']) },
+                "Sim 1804 clang-7 SGX1FLC Release":          { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=ON']) },
+                "Sim 1804 clang-7 SGX1FLC Release snmalloc": { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=ON', '-DUSE_SNMALLOC=ON']) },
             ]
             parallel testing_stages
         }
--- a/.jenkins/pipelines/Azure/Linux/Jenkinsfile
+++ b/.jenkins/pipelines/Azure/Linux/Jenkinsfile
@@ -422,7 +422,10 @@ try{
         "ACC1804 gcc Debug LVI e2e":                     { ACCTest(AGENTS_LABELS["acc-ubuntu-18.04-vanilla"], 'gcc',     'Debug',   ['-DLVI_MITIGATION=ControlFlow', '-DLVI_MITIGATION_SKIP_TESTS=ON'], [], true) },
     
         "RHEL-8 clang-8 simulation Release e2e":         { ACCTest(AGENTS_LABELS['acc-rhel-8-vanilla'], 'clang', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF'], ['OE_SIMULATION=1'], true) },
-        "RHEL-8 ACC clang-8 SGX1 Release e2e":           { ACCTest(AGENTS_LABELS['acc-rhel-8-vanilla'], 'clang', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF'], [], true) }
+        "RHEL-8 ACC clang-8 SGX1 Release e2e":           { ACCTest(AGENTS_LABELS['acc-rhel-8-vanilla'], 'clang', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF'], [], true) },
+
+        "RHEL-8 ACC clang-8 SGX1 Release snmalloc":      { ACCTest(AGENTS_LABELS['acc-rhel-8'], 'clang', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF', '-DUSE_SNMALLOC=ON']) },
+        "RHEL-8 ACC gcc-8 SGX1 Release snmalloc":        { ACCTest(AGENTS_LABELS['acc-rhel-8'], 'gcc',   'Release', ['-DHAS_QUOTE_PROVIDER=OFF', '-DUSE_SNMALLOC=ON']) },
     ]
     if(FULL_TEST_SUITE == "true") {
         stage("Full Test Suite") {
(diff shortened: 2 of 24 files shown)


include/openenclave/internal/tests.h
```diff
@@ -66,6 +66,12 @@ OE_EXTERNC_BEGIN
 
 #define OE_TEST_IGNORE(COND) OE_TEST_IF(COND, false)
 
+/* Multi-threaded testcases may need a specific base allocation
+ * when using allocators that make use of thread-local storage, such
+ * as tcmalloc or snmalloc.
+ */
```
achamayou, line 72: This seems like a useful shortcut to have for all the tests that make use of threads, because other allocators also have requirements that are typically not satisfied by the existing values (eg. tcmalloc wants 3 * 1024).

I think it would be right to tie this to USE_SNMALLOC, and to default to the largest existing hardcoded value otherwise, but I am very open to suggestions.

tests/CMakeLists.txt
```diff
@@ -67,7 +67,10 @@ if (UNIX
     if (BUILD_ENCLAVES)
       # Disable the memory test for the scenario of building the enclave in Linux and running
       # it on Windows because of the mismatch on the enclave heap size configuration.
-      add_subdirectory(memory)
+      # Also disable the memory test with snmalloc for now because it calls dlmallinfo
+      if (NOT USE_SNMALLOC)
+        add_subdirectory(memory)
```
achamayou, line 72: Please see #3262

tests/attestation_cert_apis/enc/enc.cpp
```diff
@@ -308,6 +308,6 @@ OE_SET_ENCLAVE_SGX(
     1,    /* ProductID */
     1,    /* SecurityVersion */
     true, /* Debug */
-    128,  /* NumHeapPages */
+    192,  /* NumHeapPages */
```
achamayou, line 311: Although this isn't threaded, snmalloc's class size scheme means slightly higher memory usage in this case than with dlmalloc.

tests/attestation_plugin/enc/enc.c
```diff
@@ -296,6 +296,6 @@ OE_SET_ENCLAVE_SGX(
     1,    /* ProductID */
     1,    /* SecurityVersion */
     true, /* Debug */
-    128,  /* NumHeapPages */
+    192,  /* NumHeapPages */
```
achamayou, line 299: Although this isn't threaded, snmalloc's class size scheme means slightly higher memory usage in this case than with dlmalloc.

tests/create-rapid/enc/enc.cpp
```diff
@@ -13,7 +13,7 @@ OE_SET_ENCLAVE_SGX(
     1,    /* ProductID */
     1,    /* SecurityVersion */
     true, /* Debug */
-    8,    /* NumHeapPages */
+    128,  /* NumHeapPages */
```
achamayou, line 16: Although this isn't threaded, snmalloc's class size scheme means slightly higher memory usage in this case than with dlmalloc.

achamayou:
bors try

achamayou:
bors try

radhikaj:
Thanks for your work on this @achamayou .
Currently, the default allocator is dlmalloc and I would like CI to run thorough tests against dlmalloc in master and v0.10.0. Additional tests can be added for snmalloc.
In this PR,  it appears that you are switching all tests to use snmalloc on Linux platforms when clang is used.  Please correct me if I am wrong.
Then we will not be testing the default allocator. Given that dlmalloc is the default allocator, I would like to run CI using dlmalloc.

anakrish:
@achamayou 
>  the changes to Jenkinsfiles are intended to be rolled back before merging, the purpose is to enable the tests in CI.

It is not clear what this PR's goal is if the Jenkins files changes are going to be rolled back.

achamayou:
@anakrish the goal is to get all the tests passing in all the various configurations the CI targets, something I can’t do locally easily.

The changes other than to Jenkinsfile can be merged without impact on the current CI hopefully, and the tests with other allocators become easier to run either as additional config, or less frequently in the nightly for example.

I’m happy to change approach if we don’t think this is right of course.

achamayou:
@radhika, sorry, I think phrased the PR description particularly poorly. My goal is to make the smallest possible change to the tests that allow them to pass with snmalloc on all CI targets, and then to figure out the best way to enable that on an ongoing basis.

anakrish:
> @anakrish the goal is to get all the tests passing in all the various configurations the CI targets, something I can’t do locally easily.
> 
> The changes other than to Jenkinsfile can be merged without impact on the current CI hopefully, and the tests with other allocators become easier to run either as additional config, or less frequently in the nightly for example.
> 
> I’m happy to change approach if we don’t think this is right of course.

Ok. I see another approach that is being used for testing experimental features E.g: #3222

@BRMcLaren What would be the recommended approach for locking down experimental features like OE_USE_SNMALLOC/WITH_EEID?

I don't think this particular PR is needed for v0.10, since, even if it is merged in, snmalloc will still not be used in CI for all the tests.

This PR will become important when we resume the discussion about switching the default allocator. That is however a v0.11 task.

achamayou:
@anakrish absolutely, this is not intended to be merged in 0.10, I'm sorry for the confusion caused. This is just a step towards making it easier to test OE + snmalloc, and identifying sticky points like #3262

BRMcLaren:
> @BRMcLaren What would be the recommended approach for locking down experimental features like OE_USE_SNMALLOC/WITH_EEID?

We should add one smalloc/eeid configuration to each type of build target for minimal protection, we also need to add these to the full test suite and not just PR. @achamayou I understand your plan and logic here, but once complete it might be a good idea to keep the additions in without removing any of the prior tests. Maybe not a complete duplicate of testing but at lease 1 for each os/compiler/config

i.e.
```
 "ACC1604 clang-7 Release LVI":            { ACCTest(AGENTS_LABELS["acc-ubuntu-16.04"], 'clang-7', 'Release', ['-DLVI_MITIGATION=ControlFlow', '-DLVI_MITIGATION_SKIP_TESTS=ON', '-DUSE_SNMALLOC=ON']) },
 "ACC1804 gcc Debug LVI":            { ACCTest(AGENTS_LABELS["acc-ubuntu-16.04"], 'gcc', 'Debug ', ['-DLVI_MITIGATION=ControlFlow', '-DUSE_SNMALLOC=ON']) },
```
Would be your Linux ACC configs and while minimal it is a broad net for PR testing, then add every config permutation into the full test suite for nightly regression.

@anakrish  One of the things we should start thinking about is a matrix-based testing schema for v0.11, this is getting a little.. out of hand with all the flags 😄

achamayou:
@BRMcLaren yes, thank you for the hints. While iterating though, I think this is the set I need, but I'll retarget along the lines you suggest once every test break is resolved.

I've labeled the PR as "do not merge", for clarity. Sorry for the confusion caused.

achamayou:
bors try

tests/tls_e2e/client_enc/client.cpp
```diff
@@ -397,6 +397,6 @@ OE_SET_ENCLAVE_SGX(
     1,    /* ProductID */
     1,    /* SecurityVersion */
     true, /* Debug */
-    128,  /* NumHeapPages */
+    512,  /* NumHeapPages */
```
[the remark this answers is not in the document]
achamayou: Different tests had more or less headroom to begin with, and some fare better with the change than others, because snmalloc is size-class based. Other size-classed based allocators will exhibit similar behaviour, depending on the interaction between their size-class choices and the specific allocations made by the program.

I've tried to be precise and to raise the amount 64 pages at a time where needed, but if we are concerned about the amounts are still too large, we can review the details with @mjp41 and see what can be done.

As a datapoint, it may be worth noting that tcmalloc initial alloc is 3 * 1024.

achamayou:
> I don't think it's necessary to tie the heap sizing in the tests to the USE_SNMALLOC flag yet; until we have tests where running with a lower heap size than we would use for snmalloc becomes material, it's fine to use the larger snmalloc sizes as default. The goal really is that all the tests should run in call configurations with as little custom special casing as possible.

Ok, I'm happy to change that, I was just going for the least/no amount of change with the flag off.

> On OE_TEST_MT_HEAP_SIZE itself, can you include the comment the reasoning for the numbers? e.g. why the 64 pages baseline in addition to the 64 pages per thread. It would also be useful to have some guidance as to expected usage, for example as in tests/threadcxx where the macro value is additionally expanded by another 64 pages, presumably to accommodate the extra test needs (though it doesn't look like it relative to the older values).

I will add a comment, it's essentially because snmalloc needs an initial amount for meta-data plus a per-thread allocator, and as of 0.5.0, this can be set to 256Kb by defining SNMALLOC_USE_SMALL_CHUNKS.
Increases other than values set by this formula are made to meet the specific requirements of each test, where they increased. Because snmalloc uses size-classes, it may end up with a less dense heap that dlmalloc in some cases (although it will fare better in others, I have not tried to lower allocation to see if it did compared to dlmalloc in others).

achamayou:
bors try

achamayou:
bors try

achamayou:
@BRMcLaren is it ok if I add an extra argument to https://github.com/openenclave/openenclave-ci/blob/master/src/jenkins/common/Openenclave.groovy#L129, so I can pass extra cmake args like the Linux build configs do?

That would allow me to add some -DUSE_SNMALLOC=ON configs on Windows, is is the right way to do it?

BRMcLaren:
> @BRMcLaren is it ok if I add an extra argument to https://github.com/openenclave/openenclave-ci/blob/master/src/jenkins/common/Openenclave.groovy#L129, so I can pass extra cmake args like the Linux build configs do?
> 
> That would allow me to add some -DUSE_SNMALLOC=ON configs on Windows, is is the right way to do it?

Yes, that is the correct way to do this.

include/openenclave/internal/tests.h
```diff
@@ -66,6 +66,20 @@ OE_EXTERNC_BEGIN
 
 #define OE_TEST_IGNORE(COND) OE_TEST_IF(COND, false)
 
+/* Multi-threaded testcases may need a specific base allocation
```
dthaler, line 69:
```suggestion
/* Multi-threaded test cases may need a specific base allocation
```

include/openenclave/internal/tests.h
```diff
@@ -66,6 +66,20 @@ OE_EXTERNC_BEGIN
...
+ * when using allocators that make use of thread-local storage, such
+ * as tcmalloc or snmalloc.
+ *
+ * The formula below applies to snmalloc when SNMALLOC_USE_SMALL_CHUNKS
+ * is set, which it is in Open Enclave. It corresponds to 256Kb for meta-data,
+ * and an intial 256Kb allocation per thread. Not all of that memory may be used,
+ * in particular if fewer threads that TCS are actually spawned, but it is a safe
```
dthaler, line 76:
```suggestion
 * in particular if fewer threads than TCS are actually spawned, but it is a safe
```

achamayou:
@BRMcLaren Not sure what the procedure is to get https://github.com/openenclave/openenclave-ci/pull/21 in and deployed in Jenkins, additional info would be welcome.

.jenkins/pipelines/Azure/Windows/Jenkinsfile
```diff
@@ -122,21 +123,30 @@ try{
                 "Win2019 Debug Cross Compile with DCAP libs":             { windowsCrossCompile('acc-win2019', 'Debug', 'ON') },
                 "Win2019 Release Cross Compile with DCAP libs":           { windowsCrossCompile('acc-win2019', 'Release', 'ON') },
                 "Win2019 Debug Cross Compile DCAP LVI":                   { windowsCrossCompile('acc-win2019', 'Debug', 'ON', 'ControlFlow') },
-                "Win2019 Release Cross Compile DCAP LVI":                 { windowsCrossCompile('acc-win2019', 'Release', 'ON', 'ControlFlow') }
+                "Win2019 Release Cross Compile DCAP LVI":                 { windowsCrossCompile('acc-win2019', 'Release', 'ON', 'ControlFlow') },
+
```
achamayou, line 127: I don't have a good sense of whether that's an interesting set of builds for Windows, if I should pick some others instead/too. Some feedback from someone who understands what's most often used would be very welcome here.

cmake/package_settings.cmake
```diff
@@ -57,13 +57,9 @@ include(CMakePackageConfigHelpers)
 configure_package_config_file(
   ${PROJECT_SOURCE_DIR}/cmake/openenclave-config.cmake.in
   ${CMAKE_BINARY_DIR}/cmake/openenclave-config.cmake
-  INSTALL_DESTINATION
```
achamayou, line 60: Not sure why my cmake-format wants to do that and that hasn't happened upstream.

BRMcLaren:
> @BRMcLaren Not sure what the procedure is to get [openenclave/openenclave-ci#21](https://github.com/openenclave/openenclave-ci/pull/21) in and deployed in Jenkins, additional info would be welcome.

@achamayou It needs a backend library overload, I posted the jobs in the PR and merged. You should be good to go -- we will have a better process in the future.

achamayou:
@BRMcLaren thanks, that's super quick! I'll uncomment the Windows CI entries now.

achamayou:
bors try

achamayou:
bors try

achamayou:
bors try

achamayou:
I'm not sure how this change results in a code coverage decrease, since it effectively tests code that wasn't tested in the build before. I haven't found a way to display the delta to base/master.

If anyone has some insight into that, I would welcome pointers.

radhikaj:
@mingweishih Can you comment on Amaury's question related to code coverage? Thanks!

mingweishih:
After some investigation, the cause is that the following tests are skipped:
	 61 - tests/tls_e2e (Skipped)
	 68 - tests/attestation_cert_api (Skipped)
	 69 - tests/attestation_plugin (Skipped)

It should related to recent changes on dynamic loading dcap libaries. Can you rebase against master and try again?

achamayou:
@mingweishih thank you, will do once I’m back online I’m a week or so.

achamayou:
bors try

achamayou:
bors try

achamayou:
@BRMcLaren Jenkins seems to have run out of disk:

https://oe-jenkins-dev.westeurope.cloudapp.azure.com/job/pipelines/job/Azure-Linux/1260/console

> Cloning into 'pyca-cryptography'...
> fatal: write error: No space left on device
> fatal: index-pack failed

BRMcLaren:
@achamayou Thanks for the heads up. That's odd and should be impossible due to the agents being built from images...

I came in today and saw quite a few pipelines failing, I am going to run a test on all of them and if it fails, will open an issue to investigate.

BRMcLaren:
bors try

achamayou:
bors try

achamayou:
bors try

achamayou:
@BRMcLaren I've turned off the windowsLinuxElfBuild() after some confusion when I finally realised that the reason it did not work reliably is because it's stashing and popping a build directory with a label that does not include extra cmake args.

There seem to be much fewer windowsLinuxElfBuild() than windowsCrossCompile(), is one preferred over the other, is there a transition going on? Is it worth doing the extra changes to get windowsLinuxElfBuild() to test with snmalloc at all?

achamayou:
@radhikaj with the exception of the Windows Linux Elf build (pending question above), and of the tests that require dlmalloc-specific features (I've opened a design proposal under #3410 to solve this), this PR enables tests using snmalloc on what I think is a reasonably diverse selection of build configurations.

Could I have your opinion on whether this selection is appropriate, missing some important configurations, or perhaps too broad (and therefore too expensive in terms of executor usage)?

achamayou:
bors try

.jenkins/pipelines/Agnostic/Linux/Jenkinsfile
```diff
@@ -110,19 +110,22 @@ try{
     if(FULL_TEST_SUITE == "true") {
         stage("Full Test Suite") {
             testing_stages += [
-                "Sim 1804 clang-7 SGX1 Debug":      { simulationContainerTest('18.04', 'Debug',   ['-DHAS_QUOTE_PROVIDER=OFF', '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
-                "Sim 1804 clang-7 SGX1 Release":    { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF', '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
-                "Sim 1804 clang-7 SGX1FLC Debug":   { simulationContainerTest('18.04', 'Debug',   ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
-                "Sim 1804 clang-7 SGX1FLC Release": { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) }
+                "Sim 1804 clang-7 SGX1 Debug":               { simulationContainerTest('18.04', 'Debug',   ['-DHAS_QUOTE_PROVIDER=OFF', '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
```
radhikaj, line 113: Since these tests are run on every PR, I think we can add just one with SNMALLOC. 

Sim 1804 clang-7 SGX1FLC Release snmalloc": { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF', '-DUSE_SNMALLOC=ON'])

.jenkins/pipelines/Azure/Linux/Jenkinsfile
```diff
@@ -422,7 +422,10 @@ try{
         "ACC1804 gcc Debug LVI e2e":                     { ACCTest(AGENTS_LABELS["acc-ubuntu-18.04-vanilla"], 'gcc',     'Debug',   ['-DLVI_MITIGATION=ControlFlow', '-DLVI_MITIGATION_SKIP_TESTS=ON'], [], true) },
     
         "RHEL-8 clang-8 simulation Release e2e":         { ACCTest(AGENTS_LABELS['acc-rhel-8-vanilla'], 'clang', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF'], ['OE_SIMULATION=1'], true) },
-        "RHEL-8 ACC clang-8 SGX1 Release e2e":           { ACCTest(AGENTS_LABELS['acc-rhel-8-vanilla'], 'clang', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF'], [], true) }
+        "RHEL-8 ACC clang-8 SGX1 Release e2e":           { ACCTest(AGENTS_LABELS['acc-rhel-8-vanilla'], 'clang', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF'], [], true) },
+
+        "RHEL-8 ACC clang-8 SGX1 Release snmalloc":      { ACCTest(AGENTS_LABELS['acc-rhel-8'], 'clang', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF', '-DUSE_SNMALLOC=ON']) },
```
radhikaj, line 427: Lets just keep the gcc configuration here.

.jenkins/pipelines/Agnostic/Linux/Jenkinsfile
```diff
@@ -110,19 +110,22 @@ try{
...
+                "Sim 1804 clang-7 SGX1 Release":             { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=OFF', '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
+                "Sim 1804 clang-7 SGX1FLC Debug":            { simulationContainerTest('18.04', 'Debug',   ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
+                "Sim 1804 clang-7 SGX1FLC Release":          { simulationContainerTest('18.04', 'Release', ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
+                "Sim 1804 clang-7 SGX1FLC Debug snmalloc":   { simulationContainerTest('18.04', 'Debug',   ['-DHAS_QUOTE_PROVIDER=ON',  '-DLVI_MITIGATION=None', '-DLVI_MITIGATION_SKIP_TESTS=OFF', '-DUSE_SNMALLOC=ON']) },
```
radhikaj, line 117: This one can just have the release configuration

.jenkins/pipelines/Azure/Windows/Jenkinsfile
```diff
@@ -122,21 +123,29 @@ try{
...
+                "Win2016 Sim Release Cross Compile LVI snmalloc":         { windowsCrossCompile('acc-win2016', 'Release', 'OFF', 'ControlFlow', '1', 'OFF', ['-DUSE_SNMALLOC=ON']) },
+                "Win2016 Release Cross Compile DCAP LVI snmalloc":        { windowsCrossCompile('acc-win2016', 'Release', 'ON', 'ControlFlow', '0', 'OFF', ['-DUSE_SNMALLOC=ON']) },
```
radhikaj, line 129: We can get rid of the Cross Compile ones . We should add

.jenkins/pipelines/Azure/Windows/Jenkinsfile
radhikaj: We should probably add a DEBUG config as well
"Win2019 Debug Cross Compile DCAP LVI":                 { windowsCrossCompile('acc-win2019', 'Debug', 'ON', 'ControlFlow') },

.jenkins/pipelines/Azure/Windows/Jenkinsfile
```diff
@@ -122,21 +123,29 @@ try{
...
+                "Win2016 Debug Cross Compile DCAP LVI":                            { windowsCrossCompile('acc-win2016', 'Debug', 'ON', 'ControlFlow', '0', 'ON') },
+                "Win2016 Release Cross Compile DCAP LVI":                          { windowsCrossCompile('acc-win2016', 'Release', 'ON', 'ControlFlow', '0', 'ON') },
+                "Win2019 Ubuntu1804 clang-7 Release Linux-Elf-build LVI":          { windowsLinuxElfBuild('acc-win2019-dcap', '18.04', 'clang-7', 'Release', 'ControlFlow', 'ON') },
+                "Win2019 Sim Release Cross Compile LVI ":                          { windowsCrossCompile('acc-win2019', 'Release', 'OFF', 'ControlFlow', '1', 'ON') },
+                "Win2019 Debug Cross Compile DCAP LVI":                            { windowsCrossCompile('acc-win2019', 'Debug', 'ON', 'ControlFlow', '0', 'ON') },
+                "Win2019 Release Cross Compile DCAP LVI FULL Tests":               { windowsCrossCompile('acc-win2019', 'Release', 'ON', 'ControlFlow') },
+
+                "Win2016 Release Cross Compile DCAP LVI snmalloc":                 { windowsCrossCompile('acc-win2016', 'Release', 'ON', 'ControlFlow', '0', 'ON', ['-DUSE_SNMALLOC=ON']) },
```
radhikaj, line 147: We can remove "Win2016 Release Cross Compile DCAP LVI snmalloc":                 { windowsCrossCompile('acc-win2016', 'Release', 'ON', 'ControlFlow', '0', 'ON', ['-DUSE_SNMALLOC=ON']) },

.jenkins/pipelines/Azure/Windows/Jenkinsfile
radhikaj: I think the ones you have added are good. Can we add "Win2019 Sim Debug Cross Compile LVI snmalloc":         { windowsCrossCompile('acc-win2019', 'Debug', 'OFF', 'ControlFlow', '1', 'OFF', ['-DUSE_SNMALLOC=ON']) }, instead of the 2019 simulation mode release config below?

radhikaj:
Thanks @achamayou . I added a couple of comments around test configuration.

.jenkins/pipelines/Azure/Windows/Jenkinsfile
achamayou: Of course, done.

.jenkins/pipelines/Azure/Windows/Jenkinsfile
achamayou: Done.

achamayou:
@radhikaj thank you, I have made the necessary changes.

achamayou:
bors try

achamayou:
@CodeMonkeyLeet @BRMcLaren could you confirm whether this set of changes is acceptable or if anything still needs to be addressed?

.jenkins/pipelines/Agnostic/Linux/Jenkinsfile
BRMcLaren, line 113: `-DHAS_QUOTE_PROVIDER=
`
This flag has been deprecated, an alternative is not needed can we remove this from all files? It will simplify the test matrix here.

.jenkins/pipelines/Azure/Linux/Jenkinsfile
```diff
@@ -482,28 +485,43 @@ try{
...
-                "ACC1804 clang-7 Debug LVI":              { ACCTest(AGENTS_LABELS["acc-ubuntu-18.04"], 'clang-7', 'Debug',   ['-DLVI_MITIGATION=ControlFlow', '-DLVI_MITIGATION_SKIP_TESTS=ON']) },
-                "ACC1804 clang-7 Release LVI FULL Tests": { ACCTest(AGENTS_LABELS["acc-ubuntu-18.04"], 'clang-7', 'Release', ['-DLVI_MITIGATION=ControlFlow', '-DLVI_MITIGATION_SKIP_TESTS=OFF']) },
-                "ACC1804 gcc Debug LVI":                  { ACCTest(AGENTS_LABELS["acc-ubuntu-18.04"], 'gcc',     'Debug',   ['-DLVI_MITIGATION=ControlFlow', '-DLVI_MITIGATION_SKIP_TESTS=ON']) },
-                "ACC1804 gcc Release LVI":                { ACCTest(AGENTS_LABELS["acc-ubuntu-18.04"], 'gcc',     'Release', ['-DLVI_MITIGATION=ControlFlow', '-DLVI_MITIGATION_SKIP_TESTS=ON']) },
-                "ACC1804 Code Coverage Test" :            { ACCCodeCoverageTest(AGENTS_LABELS["acc-ubuntu-18.04"], 'clang-7', 'Debug') }
+                "ACC1604 Package RelWithDebInfo LVI":              { ACCPackageTest(AGENTS_LABELS["acc-ubuntu-16.04"], '16.04', ['-DLVI_MITIGATION=ControlFlow', '-DLVI_MITIGATION_SKIP_TESTS=ON']) },
```
BRMcLaren, line 506: This looks like just whitespace changes, there is nothing wrong with that just wanted to make sure I am not missing anything vital?

BRMcLaren (commented):
@achamayou, @CodeMonkeyLeet  will be able to answer the more low-level questions. I took a look through your testing files and I found no issue sub the quote provider flag being deprecated.

.jenkins/pipelines/Agnostic/Linux/Jenkinsfile
achamayou: Ok, I'll remove all the variants where it's off.

.jenkins/pipelines/Azure/Linux/Jenkinsfile
achamayou: I was hesitant to do it, this is just to keep them all aligned as they were before. If you select the side by side diff in github, you can chose "ignore whitespace" to avoid displaying those, but I agree it is distracting on the unified diff.

.jenkins/pipelines/Agnostic/Linux/Jenkinsfile
achamayou: @BRMcLaren sorry, looking at this again, it seems to be everywhere. I'm happy to remove it, but I think it'd make sense for that to be a separate PR.

anakrish (approved):
LGTM

PR #3265, opened 2020-07-15
Integrate with SGX quote-ex library for EPID evidence generation

shnwc:
The quote-ex library is detected and loaded at run-time.

Compilation of the host-side library depends on the installation of the Intel SGX SDK in its default directory.


shnwc:
bors try

PR #3267, opened 2020-07-15
Handle SGX1FLC launch config for Win2019

### LINKED ISSUE #2960: install-windows-prereqs.ps1 should work on WS2019 with SGX1FLC option

### LINKED ISSUE #3010: install-windows-prereqs.ps1 can fail trying to set sgx_lc_msr regkey on WS2019

rs--:
Update PSW and DCAP on Windows 2019 instead of attempting to install.

Fixes #2960
Fixes #3010

Signed-off-by: Rob Sanchez rosan@microsoft.com

### DIFF
--- a/scripts/install-windows-prereqs.ps1
+++ b/scripts/install-windows-prereqs.ps1
@@ -427,6 +427,24 @@ function Install-PSW {
     } -RetryMessage "Failed to start AESMService. Retrying"
 }
 
+function Update-PSW {
+    $tempInstallDir = "$PACKAGES_DIRECTORY\Intel_SGX_PSW"
+    if(Test-Path $tempInstallDir) {
+        Remove-Item -Recurse -Force $tempInstallDir
+    }
+    Install-ZipTool -ZipPath $PACKAGES["psw"]["local_file"] `
+                    -InstallDirectory $tempInstallDir
+
+    $inf = Get-Item "$tempInstallDir\Intel*SGX*\PSW_INF*\sgx_psw.inf"
+    $devConPath = Get-DevconBinary
+
+    $update = & $devConPath update $inf "SWC\VEN_INT&DEV_0E0C"
+    Write-Output update
+    if($LASTEXITCODE -ge 2) {
+        throw "Error: Failed to update PSW drivers - exit code $LASTEXITCODE"
+    }
+}
+
 function Install-VisualStudio {
     $installerArguments = @(
         "-q", "--wait", "--norestart",
@@ -543,11 +561,13 @@ function Install-DCAP-Dependencies {
                     'path'        = "$PACKAGES_DIRECTORY\Intel_SGX_DCAP\Intel*SGX*DCAP*\base\WindowsServer2019_Windows10"
                     'location'    = 'root\SgxLCDevice'
                     'description' = 'Intel(R) Software Guard Extensions Launch Configuration Service'
+                    'hardwareId'  = '*INT0E0C'
                 }
                 'sgx_dcap' = @{
                     'path'        = "$PACKAGES_DIRECTORY\Intel_SGX_DCAP\Intel*SGX*DCAP*\dcap\WindowsServer2019_Windows10"
                     'location'    = 'root\SgxLCDevice_DCAP'
                     'description' = 'Intel(R) Software Guard Extensions DCAP Components Device'
+                    'hardwareId'  = 'SWC\VEN_INT&DEV_0E0C_DCAP'
                 }
             }
             'Win10' = @{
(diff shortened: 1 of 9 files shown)


rs--:
Re-created PR #3230 . Builds passed locally on Windows Server 2016, and 2019 - waiting on successful E2E test runs before removing the draft tag, and merging.

scripts/install-windows-prereqs.ps1
```diff
@@ -543,11 +561,13 @@ function Install-DCAP-Dependencies {
                     'path'        = "$PACKAGES_DIRECTORY\Intel_SGX_DCAP\Intel*SGX*DCAP*\base\WindowsServer2019_Windows10"
                     'location'    = 'root\SgxLCDevice'
                     'description' = 'Intel(R) Software Guard Extensions Launch Configuration Service'
+                    'hardwareId'  = '*INT0E0C'
                 }
                 'sgx_dcap' = @{
                     'path'        = "$PACKAGES_DIRECTORY\Intel_SGX_DCAP\Intel*SGX*DCAP*\dcap\WindowsServer2019_Windows10"
                     'location'    = 'root\SgxLCDevice_DCAP'
                     'description' = 'Intel(R) Software Guard Extensions DCAP Components Device'
+                    'hardwareId'  = 'SWC\VEN_INT&DEV_0E0C_DCAP'
```
CodeMonkeyLeet, line 570: This change and the change on line 564 should be applied to the Win10 configuration from lines 573-584 as well.

scripts/install-windows-prereqs.ps1
```diff
@@ -582,17 +602,29 @@ function Install-DCAP-Dependencies {
                     Throw "Failed searching for $driver driver"
```
CodeMonkeyLeet, line 602: Can you verify if the conditional on line 544 is still correct? It looks like the nesting is incorrect now and the scope of that very large block is only concerned with driver install, in which case it should only encompass the `if($LaunchConfiguration -eq "SGX1FLC")` configuration. This would also make the nested check on line 597 redundant.

scripts/install-windows-prereqs.ps1
```diff
@@ -582,17 +602,29 @@ function Install-DCAP-Dependencies {
...
+                try
+                {
+                    Write-Output "Installing driver $($drivers[${OS_VERSION}][$driver]['location'])"
+                    $install = & $devConBinaryPath install "$($inf.FullName)" $drivers[${OS_VERSION}][$driver]['location']
+                    Write-Output $install
+                    if($LASTEXITCODE) {
+                        Throw "Failed to install $driver driver"
+                    }
+                }
+                catch
```
CodeMonkeyLeet, line 619: This catch fallback logic doesn't address my concern around the Win10 configuration mentioned in #3230 and it also obscures the intent of the code flow:

* The update logic still references the update target as indexed by `OS_VERSION` so it would fail on Win10 as that configuration did not specify the `hardwareId` values. Was this tested on Win10?
* This `catch` logic looks only ever intended to run for non-WS2016 code (the WS2016 configuration also doesn't specify `hardwareId` values for this codepath) but the way it does this is non-obvious because only WS2016 tries to uninstall a driver first, and is a based on the undocumented expectation that WS2016 does not have a driver installed by default but will uninstall & reinstall over one if it does (as opposed to upgrade).

Also, have we validated that the codepath for _installing_ the WS2019 driver works? I don't think it does because devcon.exe takes the hardware ID in for both the install and update commands; the distinction between the `location` and `hardwareId` parameters defined by this script seems to be spurious and they can probably be consolidated into the more accurately named `hardwareId`.

If uninstall/install is intended to only run for WS2016, can we explicitly break out that behavior by `OE_VERSION` check? Alternatively, if the desire is to always try the install and then try to upgrade, can this be tested to ensure that it works in all configurations (3 OS versions * with/without driver already installed)

scripts/install-windows-prereqs.ps1
```diff
@@ -582,17 +602,29 @@ function Install-DCAP-Dependencies {
...
+                {
+                    Write-Output "Updating driver $($drivers[${OS_VERSION}][$driver]['location'])"
+                    $update = & $devConBinaryPath update "$($inf.FullName)" $drivers[${OS_VERSION}][$driver]['hardwareId']
+                    Write-Output $update
+                    if($LASTEXITCODE -ge 2) {
```
CodeMonkeyLeet, line 624: Can you comment why >= 2 here?

scripts/install-windows-prereqs.ps1
```diff
@@ -582,17 +602,29 @@ function Install-DCAP-Dependencies {
...
-                $install = & $devConBinaryPath install "$($inf.FullName)" $drivers[${OS_VERSION}][$driver]['location']
-                if($LASTEXITCODE) {
-                    Throw "Failed to install $driver driver"
+                try
+                {
+                    Write-Output "Installing driver $($drivers[${OS_VERSION}][$driver]['location'])"
+                    $install = & $devConBinaryPath install "$($inf.FullName)" $drivers[${OS_VERSION}][$driver]['location']
+                    Write-Output $install
+                    if($LASTEXITCODE) {
+                        Throw "Failed to install $driver driver"
```
CodeMonkeyLeet, line 616: Can you do a specific check for the error code for when the driver is already installed and fall through to trying an install instead of using exception handling?

Issue #3178, opened 2020-06-24
FLC/Non-FLC in one build?


letmaik:
OE has `HAS_QUOTE_PROVIDER=ON|OFF` as build-time option. Since the Intel SGX libraries are now all dlopen'd I was wondering whether it makes sense to allow building a single OE variant that works across all SGX variants (currently FLC/non-FLC) with appropriate runtime introspection.

radhikaj:
This would be resolved if HAS_QUOTE_PROVIDE is deleted as an option.

anakrish:
I looked at this. We are quite close to being able to build OE on a non FLC machine and use it in an FLC machine (ie FLC/Non-FLC in one build).
In addition to removing HAS_QUOTE_PROVIDER, we need to 
- remove dependency on Intel SGX headers in host/sgx/quote.c
- Remove LING_SGX_DCAP_QL
- Figure out whether the platform has dcap and selectively enable tests.

This looks quite possible to do v0.11

anakrish:
This is fixed by the PRs currently checked in as part of #3272.

PR #3268, opened 2020-07-15
Integrate with SGX quote-ex library for EPID evidence generation

shnwc:
Implement the design proposal on [OE SDK integration with the Intel SGX SDK quote-ex library](../blob/master/docs/DesignDocs/SGX_QuoteEx_Integration.md), with update of the design doc in PR #3329.


shnwc:
bors try

host/sgx/sgxquote_ex.h
```diff
@@ -0,0 +1,108 @@
...
+    size_t key_id_count;
+    size_t mapped_key_id_count;
+    bool* mapped;
+    oe_uuid_t* uuid;
+    sgx_att_key_id_ext_t* sgx_key_id;
+} oe_sgx_quote_ex_library_t;
+
+oe_result_t oe_initialize_quote_ex_library(void);
+
+void oe_load_quote_ex_library(oe_sgx_quote_ex_library_t* library);
```
dthaler, line 95: Suggest adding sgx into the function names

host/sgx/windows/sgxquoteexloader.c
```diff
@@ -0,0 +1,146 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#if defined(OE_LINK_SGX_DCAP_QL)
+
+#if __has_include(<sgx_uae_quote_ex.h>) && __has_include(<sgx_urts.h>)
+#define OE_LINK_SGX_QUOTE_EX
```
CodeMonkeyLeet, line 7: I would discourage any additional build time dependency variant behavior since we want to minimize the variety of libraries that OE SDK has to carry/release. Be aware that the general direction is to move away from these kinds of build flags, see #3270 which also interacts with this PR in its use of `OE_LINK_SGX_DCAP_QL`

host/CMakeLists.txt
```diff
@@ -487,6 +489,21 @@ if (OE_SGX)
     endif ()
     # turn on 'OE_LINK_SGX_DCAP_QL' for the preprocessor
     target_compile_definitions(oehost PUBLIC OE_LINK_SGX_DCAP_QL)
+    # If Intel SGX SDK is installed in SGX_SDK_PATH, add it to include path
+    if (NOT SGX_SDK_PATH)
```
anakrish, line 493: It would be better to avoid this build time dependency on Intel SGX SDK.
We want to be able to build OE SDK package on non FLC and even non SGX machines. See #3178 

My recommendation would be have the necessary type definitions in [openenclave/bits/sgx/sgxtypes.h](https://github.com/openenclave/openenclave/blob/master/include/openenclave/bits/sgx/sgxtypes.h).

host/sgx/windows/sgxquoteexloader.c
anakrish: __has_include seems to be a C++17 feature. Does this work in MSVC?

anakrish:
bors try

enclave/sgx/attester.c
```diff
@@ -24,45 +24,6 @@
 #include "../core/sgx/report.h"
 #include "platform_t.h"
 
-#if !defined(OE_USE_BUILTIN_EDL)
-/**
- * Declare the prototype of the following function to avoid the
- * missing-prototypes warning.
- */
-oe_result_t _oe_get_supported_attester_format_ids_ocall(
```
mingweishih, line 32: why removing these code?

tests/edl_opt_out/edl/edl_opt_out.edl
```diff
@@ -13,6 +13,7 @@ enclave {
         oe_syscall_flock_ocall;
 #ifdef OE_SGX
     from "openenclave/edl/sgx/cpu.edl" import *;
+    from "openenclave/edl/sgx/platform.edl" import *;
```
CodeMonkeyLeet, line 16: This looks like a regression introduced by the change in enclave/sgx/attester.c. This test  explicitly checks that the attestation.edl functions do not need to be opted into by default and this PR effectively removes the weak links and forces the inclusion of  attestation.edl it seems.

enclave/sgx/attester.c
```diff
@@ -24,45 +24,6 @@
 #include "../core/sgx/report.h"
 #include "platform_t.h"
 
-#if !defined(OE_USE_BUILTIN_EDL)
```
mingweishih, line 27: Please revert the code changes here.

tests/edl_opt_out/edl/edl_opt_out.edl
mingweishih: @CodeMonkeyLeet is correct. Please revert the removal of weak links in attester.c and this line as well.

enclave/sgx/attester.c
shnwc: Got it.

tests/edl_opt_out/edl/edl_opt_out.edl
shnwc: Got it now. The relevant changes will be reverted.

enclave/sgx/attester.c
shnwc: Sure.

host/CMakeLists.txt
shnwc: The agreement from the SIG-attestation meeting was to have a 3rdparty directory into which the needed header files are copied. Will implement this scheme and remove the build-time dependency.

host/sgx/sgxquote_ex.h
shnwc: Got it, thanks!

host/sgx/windows/sgxquoteexloader.c
shnwc: Will fix. See [another comment](https://github.com/openenclave/openenclave/pull/3268/files/9010a324028d27c02fa0bab2361b695346842917#r455991059)

shnwc:
@CodeMonkeyLeet @yentsanglee I'm adding directory `3rdparty/sgxsdk/` and put an `update.make` to retrieve all the needed SGX SDK header files. Question: how to trigger execution of the new `update.make`?

mingweishih:
> @CodeMonkeyLeet @yentsanglee I'm adding directory `3rdparty/sgxsdk/` and put an `update.make` to retrieve all the needed SGX SDK header files. Question: how to trigger execution of the new `update.make`?

Does it mean that you add the sgxsdk as the git submodule?

mingweishih:
> > @CodeMonkeyLeet @yentsanglee I'm adding directory `3rdparty/sgxsdk/` and put an `update.make` to retrieve all the needed SGX SDK header files. Question: how to trigger execution of the new `update.make`?
> 
> Does it mean that you add the sgxsdk as the git submodule?

Given there is already an agreement from the sig meeting, I'm ok with it.

For the `update.make` approach, I would recommend not using it as Windows does not natively support `GNU make`. Instead, you can use `ExternalProject_Add` and cmake commands to do the file copy (see example https://github.com/openenclave/openenclave/blob/05d109a5706fe4c439f64cd68ffee6e2430b9542/3rdparty/musl/CMakeLists.txt#L43).

shnwc:
> > @CodeMonkeyLeet @yentsanglee I'm adding directory `3rdparty/sgxsdk/` and put an `update.make` to retrieve all the needed SGX SDK header files. Question: how to trigger execution of the new `update.make`?
> 
> Does it mean that you add the sgxsdk as the git submodule?

I'm trying to mimic the file `3rdparty/dlmalloc/update.make`, which contains `wget` to retrieve individual files. But I see that in https://github.com/openenclave/openenclave/blob/master/3rdparty/dlmalloc/, the retrieved files are already present. So they do not seem are retrieved on the fly when `openenclave` is cloned.

mingweishih:
> > > @CodeMonkeyLeet @yentsanglee I'm adding directory `3rdparty/sgxsdk/` and put an `update.make` to retrieve all the needed SGX SDK header files. Question: how to trigger execution of the new `update.make`?
> > 
> > 
> > Does it mean that you add the sgxsdk as the git submodule?
> 
> I'm trying to mimic the file `3rdparty/dlmalloc/update.make`, which contains `wget` to retrieve individual files. But I see that in https://github.com/openenclave/openenclave/blob/master/3rdparty/dlmalloc/, the retrieved files are already present. So they do not seem are retrieved on the fly when `openenclave` is cloned.

I doubt that that `update.make` is even used today...

mingweishih:
> > > > @CodeMonkeyLeet @yentsanglee I'm adding directory `3rdparty/sgxsdk/` and put an `update.make` to retrieve all the needed SGX SDK header files. Question: how to trigger execution of the new `update.make`?
> > > 
> > > 
> > > Does it mean that you add the sgxsdk as the git submodule?
> > 
> > 
> > I'm trying to mimic the file `3rdparty/dlmalloc/update.make`, which contains `wget` to retrieve individual files. But I see that in https://github.com/openenclave/openenclave/blob/master/3rdparty/dlmalloc/, the retrieved files are already present. So they do not seem are retrieved on the fly when `openenclave` is cloned.
> 
> I doubt that that `update.make` is even used today...

I felt it's a bad idea to retrieve files over internet at the build time, which would complicate the build process.
Is that already a consensus with this approach?

cc @dthaler

shnwc:
> > > > @CodeMonkeyLeet @yentsanglee I'm adding directory `3rdparty/sgxsdk/` and put an `update.make` to retrieve all the needed SGX SDK header files. Question: how to trigger execution of the new `update.make`?
> > > 
> > > 
> > > Does it mean that you add the sgxsdk as the git submodule?
> > 
> > 
> > I'm trying to mimic the file `3rdparty/dlmalloc/update.make`, which contains `wget` to retrieve individual files. But I see that in https://github.com/openenclave/openenclave/blob/master/3rdparty/dlmalloc/, the retrieved files are already present. So they do not seem are retrieved on the fly when `openenclave` is cloned.
> 
> I doubt that that `update.make` is even used today...

Actually `update.make` is executable and can run manually to retrieve all the listed header files. To mimic the `dlmalloc` behavior, I intend to run `update.make` manually and commit the retrieved header files to `git`.

As discussed in the SIG-attestation meeting, only a few header files are needed, so we do not want to go the submodule approach to retrieve the whole SGX SDK.

CodeMonkeyLeet:
> > > > > @CodeMonkeyLeet @yentsanglee I'm adding directory `3rdparty/sgxsdk/` and put an `update.make` to retrieve all the needed SGX SDK header files. Question: how to trigger execution of the new `update.make`?
> > > > 
> > > > 
> > > > Does it mean that you add the sgxsdk as the git submodule?
> > > 
> > > 
> > > I'm trying to mimic the file `3rdparty/dlmalloc/update.make`, which contains `wget` to retrieve individual files. But I see that in https://github.com/openenclave/openenclave/blob/master/3rdparty/dlmalloc/, the retrieved files are already present. So they do not seem are retrieved on the fly when `openenclave` is cloned.
> > 
> > 
> > I doubt that that `update.make` is even used today...
> 
> I felt it's a bad idea to retrieve files over internet at the build time, which would complicate the build process.
> Is that already a consensus with this approach?

@shnwc @mingweishih 

The update.make files are effectively the .gitmodules equivalent coupled with custom scripts for non-git sources. They are not intended to be automatically triggered and are automation scripts for use by the maintainers of the project to track and update the versions of cloned dependencies in the OE SDK repo itself.

They are make files, so can be executed on a lInux system via `make -f update.make`. The changes will then need to be git added and pushed as a PR.

Arguably, they should have been some kind of cross-platform script like python, or even bash that can run in git bash on Windows, but these are legacy at this point, and we are moving towards git submodules over time. 

**There should not be automatic updating of dependencies without maintainer commits to the project.** As with git submodules, the dependencies should be pinned to a known good version independent of individual contributors cloning it to their machines.

shnwc:
bors try

shnwc:
bors try

host/sgx/sgxquote.c
```diff
@@ -131,12 +213,67 @@ oe_result_t oe_sgx_qe_get_target_info(
...
+    oe_result_t result = OE_UNEXPECTED;
     quote3_error_t err = SGX_QL_ERROR_UNEXPECTED;
 
-    OE_UNUSED(format_id);
-    OE_UNUSED(opt_params);
-    OE_UNUSED(opt_params_size);
+    if (!format_id || !target_info)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+    if (oe_sgx_initialize_quote_ex_library() == OE_OK)
```
yentsanglee, line 222: Can you add comment to why getting target info from quote_ex is favor over _sgx_qe_get_target_info()?

shnwc: The two functions could be talking to different QEs. While the quote_ex always talks to the QE in AESM, _sgx_qe_get_target_info() could be talking to the DCAP in-proc QE.

shnwc:
bors try

shnwc:
bors try

host/sgx/sgxquote.c
CodeMonkeyLeet: @shnwc I don't think that speaks to @yentsanglee's suggestion, which isn't about why there are two methods of getting the target info in this function. I think he's pointing out that it's not obvious from reading the code why `_sgx_qe_get_target_info()` is only called if `oe_sgx_initialize_quote_ex_library()` fails entirely. i.e. it looks like an app expecting to talk to the DCAP in-proc QE would get short-circuited and never called if the quote_ex library is loadable.

It would be useful to explain as a code comment here (and elsewhere where this pattern is repeated, like in `oe_sgx_qe_get_quote_size()`) why this is either acceptable, or not the case (e.g. we expect that `oe_sgx_initialize_quote_ex_library()` is the function that checks the whether or not to use an internal provider and will fail so that this code then falls through to the `_sgx_qe_get_target_info()` call?).

shnwc:
> @shnwc
> Is the presence of `libsgx_quote_ex.so` optional on an FLC machine or is it required for attestation to work on an FLC machine?

The presence of `libsgx_quote_ex.so` is optional. If it is not present, the DCAP library will need to be present, and only ECDSA is supported.

host/sgx/sgxquote.c
shnwc: Got it. Will restructure the code and add relevant comments.

3rdparty/sgxsdk/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) Open Enclave SDK contributors.
```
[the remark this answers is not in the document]
shnwc: Make sense. Will  update.

3rdparty/sgxsdk/LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD License
```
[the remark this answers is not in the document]
shnwc: Make sense. Will remove this file.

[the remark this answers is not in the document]
anakrish: @shnwc Thanks for removing the dependency on Intel SDK and just depending on the published headers.
In the past we have been quite hesitant to include TEE headers directly into the OE repository See discussion this [PR](https://github.com/openenclave/openenclave/pull/2882/files).

Should we just have a repository `openenclave/sgxsdk` that holds the headers and then have a sub module dependency on that repository?
CC: @dthaler @mingweishih

host/sgx/linux/sgxquoteexloader.c
```diff
@@ -0,0 +1,129 @@
...
+}
+
+void oe_sgx_load_quote_ex_library(oe_sgx_quote_ex_library_t* library)
+{
+    oe_result_t result = OE_UNEXPECTED;
+
+    if (!library)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+    _quote_ex_library = library;
```
anakrish, line 50: Since a global variable `_quote_ex_library` is being set, it is not clear if we have made this code thread-safe.

3rdparty/sgxsdk/LICENSE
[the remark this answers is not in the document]
shnwc: @anakrish the current solution is only temporary. Eventually we expect to have a package that installs the needed headers on development platforms.

enclave/core/sgx/report.c
```diff
@@ -271,11 +271,6 @@ oe_result_t oe_get_remote_report(
     size_t sgx_report_size = sizeof(sgx_report);
     sgx_quote_t* sgx_quote = NULL;
 
-    // For remote attestation, the Quoting Enclave's target info is used.
```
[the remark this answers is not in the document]
shnwc: `oe_get_report()` stays the same and will only generate ECDSA reports. You've got a good catch in checking null `opt_params`. Will add the check to the `oe_get_report()` path, in `enclave/sgx/report.c`.

host/sgx/linux/sgxquoteexloader.c
shnwc: @anakrish the top-level function that set and manipulate this variable is `_load_quote_ex_library_once()` in `host/sgx/sgxquote.c`. This function is only run once, as guarded by a mutex in function `_use_quote_ex_library()`.

3rdparty/sgxsdk/LICENSE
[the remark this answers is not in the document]
anakrish: Thanks for the clarification.

shnwc:
bors try

shnwc:
bors try

shnwc:
@mingweishih the requested code reversal has been done in the latest commits. Please confirm and remove the block if you agree.

host/sgx/windows/sgxquoteexloader.c
```diff
@@ -0,0 +1,135 @@
...
+    oe_result_t result = OE_UNEXPECTED;
+
+    if (!library)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+    _quote_ex_library = library;
+
+    if (library->handle == 0)
+    {
+        HMODULE _handle = 0;
```
dthaler, line 54: https://github.com/openenclave/openenclave/blob/master/docs/Community/DevelopmentGuide.md says:
> The single underscore (_ ) is reserved for local definitions (static, file-scope definitions). e.g. static oe_result_t _parse_sgx_report_body(..).

So this variable should not have an underscore prefix since it is not static

host/sgx/sgxquote.c
```diff
@@ -125,18 +163,213 @@ static bool _load_sgx_dcap_ql(void)
     return (_module != NULL);
 }
 
+static void _load_quote_ex_library_once(void)
+{
+    bool* tmp_mapped = NULL;
+    oe_uuid_t* tmp_uuid = NULL;
```
dthaler, line 169: Spell out "tmp" or remove it, per guidance in https://github.com/openenclave/openenclave/blob/master/docs/Community/DevelopmentGuide.md to avoid abbreviations in names.

host/sgx/linux/sgxquoteexloader.c
shnwc: A better place for the comment is in the header file that declares this this function that requires platform specific implementation? I'll add comments there (`host/sgx/sgxquote_ex.h`)

host/sgx/windows/sgxquoteexloader.c
shnwc: The variable itself is indeed a static one. It is pointer assigned to the address passed the function as a parameter.

host/sgx/sgxquote.c
shnwc: Got it.

shnwc:
@dthaler the latest commit incorporated your comments 3 days ago. Thanks!

shnwc:
@mingweishih  the requested code reversal has been done in the latest commits. Please confirm and remove the block if you agree. Otherwise let me know if any requests were missed. Thanks!

shnwc:
bors try

host/sgx/sgxquote.c
```diff
@@ -125,18 +163,213 @@ static bool _load_sgx_dcap_ql(void)
...
+    if (!format_id || !target_info)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+    if (_use_quote_ex_library())
+    {
+        sgx_status_t status = SGX_ERROR_UNEXPECTED;
+        sgx_att_key_id_ext_t updated_key_id = {{0}};
+        size_t tmp_size = 0;
+        uint8_t* tmp_buffer = NULL;
+        sgx_target_info_t tmp_target_info;
```
dthaler, line 324: nit: you updated the variables above to remove the abbreviation, but not throughout the rest of this file, such as here.

host/sgx/linux/sgxquoteexloader.c
```diff
@@ -0,0 +1,129 @@
...
+
+    if (library->handle == 0)
+    {
+        void* _handle = 0;
```
dthaler, line 54: As noted in my previous comment, this variable is not static (it's "void*" not "static void*") and so should not start with _ per OE coding style guidelines.

3rdparty/sgxsdk/update.make
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/make -f
```
anakrish, line 1: Does this file need to be checked-in?  It is not used and the `wget` command would work only on Linux.

host/sgx/linux/sgxquoteexloader.c
shnwc: Oh got it. I mistook your reference as to `_quote_ex_library`. The variable name `_handle` was borrowed from existing file `host/sgx/windows/sgxquoteproviderloader.c`. Will fix the naming in both the Linux and Windows versions.

host/sgx/sgxquote.c
```diff
@@ -125,18 +163,213 @@ static bool _load_sgx_dcap_ql(void)
...
+            NULL);
+
+        if (status != SGX_SUCCESS)
+            OE_RAISE_MSG(
+                OE_PLATFORM_ERROR,
+                "sgx_init_quote_ex(NULL) returned 0x%x\n",
+                status);
+
+        tmp_buffer = (uint8_t*)oe_malloc(tmp_size);
+        OE_TEST(tmp_buffer);
```
anakrish, line 352: I don't think we use OE_TEST in SDK code; only in test code. It would be better to use OE_CHECK instead.

host/sgx/sgxquote.c
shnwc: Got it. Thanks!

3rdparty/sgxsdk/update.make
shnwc: I see that other 3rdparty directories have this file checked in, and some of them use `wget` too.

host/sgx/sgxquote.c
shnwc: Totally agree! Will fix.

shnwc:
bors try

shnwc:
@anakrish @dthaler Thank you very much for your comments! They should have been addressed in the latest commit.

3rdparty/sgxsdk/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
...
+# Licensed under the MIT License.
+
+if (OE_SGX)
+  # This 3rdpary directory only contains header files for compilation of
+  # SGX-specific host-side code.
+  # There is no new artifacts (e.g. new targets) to be added.
```
CodeMonkeyLeet, line 7: Consider changing this to a README.md file instead (see other 3rdparty folders) since this only contains a comment for the folder.

Also some typo and grammar corrections:
```suggestion
  # This 3rdparty directory only contains header files for compiling
  # SGX-specific host-side code.
  # There are no new artifacts (e.g. new targets) to be added.
```

enclave/sgx/attester.c
```diff
@@ -214,7 +229,8 @@ static oe_result_t _get_evidence(
             oe_result_str(result));
 
         // Get the endorsements from the report if needed.
-        if (endorsements_buffer)
+        // No support of endorsements for  EPID quotes
```
CodeMonkeyLeet, line 232:
```suggestion
        // No support of endorsements for EPID quotes
```

host/sgx/sgxquote.c
```diff
@@ -125,18 +163,214 @@ static bool _load_sgx_dcap_ql(void)
...
+        _quote_ex_library.load_result = result;
+
+    OE_TRACE_INFO(
+        "_load_quote_ex_library_once() result=%s\n",
+        oe_result_str(_quote_ex_library.load_result));
+
+    return;
+}
+
+// For choosing between the DCAP library and the quote-ex library for quote
```
CodeMonkeyLeet, line 294: (Quoting a dislocated comment for context)
> @shnwc I don't think that speaks to @yentsanglee's suggestion, which isn't about why there are two methods of getting the target info in this function. I think he's pointing out that it's not obvious from reading the code why _sgx_qe_get_target_info() is only called if oe_sgx_initialize_quote_ex_library() fails entirely. i.e. it looks like an app expecting to talk to the DCAP in-proc QE would get short-circuited and never called if the quote_ex library is loadable.
>
> It would be useful to explain as a code comment here (and elsewhere where this pattern is repeated, like in oe_sgx_qe_get_quote_size()) why this is either acceptable, or not the case (e.g. we expect that oe_sgx_initialize_quote_ex_library() is the function that checks the whether or not to use an internal provider and will fail so that this code then falls through to the _sgx_qe_get_target_info() call?).

Thanks for adding the comment, but it doesn't explain _why_ this choice exists to a new maintainer reading this code, or why this order of fallback is chosen. For example, why not try `_load_sgx_dcap_ql()` first and then fallback to `__use_quote_ex_library()` if that fails? They're both effectively just load library checks. Given the current logic, if I want to use the existing DCAP library, I cannot install the quote_ex library on the same device even though there may be two different host apps that may choose to use libraries; the host app doesn't seem to get a choice, only the runtime environment matters.

shnwc: @CodeMonkeyLeet this PR implements the design proposal on [OE SDK integration with the Intel SGX SDK quote-ex library](../blob/master/docs/DesignDocs/SGX_QuoteEx_Integration.md), with update of the design doc in PR #3329.

If a reference of the design doc is included in the proposed `3rdparty/sgxsdk/README.md`, will this be sufficient for clarification? Or do you see some other files where this design doc should be referenced?

3rdparty/sgxsdk/CMakeLists.txt
shnwc: Thanks!

enclave/sgx/attester.c
shnwc: Thanks!

host/sgx/sgxquote.c
CodeMonkeyLeet: The original design proposed a different linking and selection mechanism, which has since changed. Option 1 was only briefly documented and not deeply considered in the original design because it was not the recommended option and hence not the focus of that design.

Now that this PR is choosing to implement that option instead, I would like to ensure that I understand the developer use case. The question that I am asking here needs to be answered as part of the proposed update to that design #3329.

shnwc: Will add a reference to the design document as part of this comment.

shnwc:
bors try

shnwc:
@CodeMonkeyLeet please review the latest commit, to confirm that your requested changes have been incorporated. Thanks!

host/sgx/sgxquote.c
shnwc: > The original design proposed a different linking and selection mechanism, which has since changed. Option 1 was only briefly documented and not deeply considered in the original design because it was not the recommended option and hence not the focus of that design.
> 
> Now that this PR is choosing to implement that option instead, I would like to ensure that I understand the developer use case. The question that I am asking here needs to be answered as part of the proposed update to that design #3329.

The implementation moved from the original option 4 to the latest option 1 due to the merge of another PR #3273. The priority between the DCAP and the quote-ex libraries stays the same between the two options.

shnwc:
bors try

shnwc:
The latest commits in this PR and PR #3329 incorporated the agreement in this week's SIG-attestation meeting. Please review.

host/sgx/sgxquote.c
```diff
@@ -125,18 +165,231 @@ static bool _load_sgx_dcap_ql(void)
...
+// generation: the quote-ex library will be used if it can be loaded
+// and initialized successfully, otherwise the DCAP library is used.
```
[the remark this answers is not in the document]
shnwc: Thanks!

host/sgx/sgxquote.c
```diff
@@ -125,18 +165,231 @@ static bool _load_sgx_dcap_ql(void)
...
+                OE_PLATFORM_ERROR,
+                "sgx_init_quote_ex(local_buffer) returned 0x%x\n",
+                status);
+
+        memcpy(target_info, &local_target_info, sizeof(sgx_target_info_t));
+
+        result = OE_OK;
+        goto done;
+    }
+
```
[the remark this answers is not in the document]
shnwc: Sure, will replace all the goto's. By using goto's, my intention was to minimize changing the existing code.

shnwc:
bors try

PR #3269, opened 2020-07-15
Add ExtendedEnclaveInitializationData_V2.md 

bodzhang:
Add ExtendedEnclaveInitializationData_V2.md to docs/DesignDocs/, with the updated design for EEID.

Signed-off-by: Bo Zhang <zhanb@microsoft.com>


docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,182 @@
...
... the EEID support (URL), the variable length configuration data buffer is part of the EEID pages added into the enclave by the enclave loader, and the enclave code access the configuration data through the eeid data structure. Intel's 10th-gen Core processor supports the Key Separation and Sharing (KSS) feature, including `CONFIGID` and `CONFIGSVN`, which are new fields defined in `SECS`. `CONFIGID` is intended to allow enclave creator to indicate what additional content may be accepted by the enclave post enclave initialization, whose exact usage depends on the enclave implementation. `CONFIGSVN` can be used in case `CONFIGID` does not fully reflect the identity of the additional content that may be accepted by the enclave post enclave initialization, for example, `CONFIGID` as the hash of the signing key for verifying the additional content. The `CONFIGID` and `CONFIGSVN` are part of the enclave identity produced by the CPU, and reflects the identity of the additional code/data allowed to be loaded into the enclave, committed at the enclave initialization time. The CONFIGID/CONFIGSVN based solution on SGX CPUs would expose a API set different from the current EEID implementation.
```
dthaler, line 9: Please break source lines into short lines, say 80 characters or less, to stick with OE style and make github diffs easier to review.  As it is, this is hard to review (especially after various suggestions like below where it's really hard to see what's changed in the suggestions).

wintersteiger: +1. I hoped that GitHub would show my suggestion diffs in a better way, but it doesn't. Best to stick to a line limit to keep diffs readable.

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,182 @@
...
+
+As the enclave developer owns the definition of config_id and config_svn, the developer is responsible for providing the configuration data identified by config_id/config_svn to the Relying Party, if it's necessary for the Relying Party to analyze the configuration data. The Relaying Party should make sure the identity of the configuration data received matches config_id/config_svn claims produced by the enclave attestation verification process.
+
+For SGX KSS feature based solution, the attestation evidence verification implementation will be different, but the claims produced will also include config_id/config_svn as well as other enclave identity information, extracted from the HW generated enclave attestation data. The low level implementation difference between EEID based solution and SGX KSS feature based solution is not exposed to the Relaying Party.
```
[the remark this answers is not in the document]
ad-l: I am OK with the idea to use the same claim name implemented differently with or without KSS - however I want to make sure that we understand the impact on enclave runtime configuration. KSS configId is meant to be used for enclave-specific/application configuration so we need to make sure we have a clear plan on how we attest all runtime settings, including maximal heap, stack and TCS with EDMM. If it turns out that with KSS and EDMM we still have settings that can only be set at enclave signing time and that we want to change at enclave loading we may have to re-introduced a loader-specific hack on top of configId or mrenclave again...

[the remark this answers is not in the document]
wintersteiger: That's a good point @ad-l; if I understand correctly (and I may not), the memory settings are still fixed at signing time with EDMM, but now they allow a range of settings instead of one specific one. I think that may be fine for most applications, but there may be some that want tighter control than a large range of settings, e.g. for #TCS. @bodzhang: correct me if I'm wrong about that. Can you think of other new settings that we may want to change but still have attested?

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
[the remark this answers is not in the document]
bodzhang: Fixed

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,182 @@
...
+
+The initial EEID design only supports enclave load time selection of heap/stack size and #TCS, and places the EEID pages after the heap and thread sections. The enclave's linear address space size is set as a fixed value big enough to encompass all reasonable selections of heap/stack size and/or #TCS, with large range of linear address potentially not committed and not accessible by the enclave code. The variance of the location of the EEID pages and the variance of the end of the committed enclave pages, both under the potentially malicious loader's control, complicate the EEID implementation and require more security design/code review and validation. For usages that do not need to support enclave load time selection of heap/stack size and #TCS, an EEID enclave with enclave heap/stack size and #TCS specified at enclave signing time is more attractive.
+
+The update design aims to achieve the following goals:
```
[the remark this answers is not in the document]
bodzhang: Adopted the edit

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,182 @@
...
+
+```C
+#ifdef EXPERIMENTAL_EEID
+typedef struct oe_eeid_t_
+{
+    uint32_t version;        /* version number of the structure */
+    uint32_t hash_state[10]; /* internal state of the hash computation at the end of the enclave base image */
+    uint8_t config_id[32];   /* identity of the code/data allowed to be loaded into the enclave after enclave init */
+    unit16_t config_svn;     /* minimum SVN of the code/data allowed to be loaded into the enclave after enclave init */
+    oe_enclave_size_settings_t size_settings; /* heap, stack and thread configuration for this instance. if all-0, indicating heap/stack size and #TCS are specified at enclave signing time */
```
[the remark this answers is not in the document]
bodzhang: For "static sizing" EEID enclave, heap/stack/TCS info is part of the base enclave image and `size_setting` in eeid data structure is not used. If `size_setting` is all-0, it's an indication that the info is not used.

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,182 @@
...
+    uint64_t signature_size; /* size of signature, max size is 4096-sizeof(oe_eeid_t) bytes */
+    uint8_t signature[];     /* Buffer holding the signature */
+} oe_eeid_t;
+#endif
+```
+
+The enclave developer owns the definition of config_id and config_svn. The enclave developer is also responsible for the host side code that calculates the identity of the code/data to be loaded into enclave post enclave initialization and pass the proper values for config_id/config_svn to the enclave loader. The enclave loader sets the config_id/config_svn through the eeid page or in SGX `SECS` if SGX KSS is supported. The enclave developer should implement an explicit function (typically an ECALL) to load the code/data into the enclave memory post enclave initialization and to verify the identity and/or SVN of the loaded code/data against config_id/config_svn. On SGX CPUs supporting the KSS feature, config_id and config_svn are available in the SGX `REPORT`. The OE SDK libs will provide an API to retrieve config_id and config_svn. The low level implementation difference between EEID based solution and SGX KSS feature based solution is not exposed to the developer's code.
```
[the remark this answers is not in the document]
bodzhang: The relationship between config_id/config_svn and the extra content to be loaded into the enclave post enclave initialization is up to the enclave developer. For example, one developer might want to load a integrity protected disk's hash tree root hash into the enclave, so the enclave code can use it to verify the disk content when accessing the disk. In this example, the config_id might be simple the root hash itself. In another example, the developer might want to load a signed .so into the enclave, and can put the hash of the signing key cert in config_id, so the enclave code can verify the signing key cert provided to it together with the .so file in an ECALL matches the committed config_id at enclave load time. In this example, the developer can also commit the version of the signed .so through config_id at enclave load time (the enclave code will make sure the version of the .so file to be loaded through the ECALL matches the value in config_id). Implementing one fixed relationship between config_id/config_svn and the content allowed to be loaded into the enclave post initialization in a OE SDK lib can not meet all potential needs from the developers.

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,182 @@
...
+    uint8_t sigstruct[OE_SGX_SIGSTRUCT_SIZE];
+
+    /* (1904) from oeimage.entry_rva, must match TCS.OENTRY */
+    uint64_t entry_rva;
+} oe_sgx_enclave_properties_t;
+```
+
+The EEID initialization code inside the enclave makes sure the `entry_point` in the EEID data structure matches `entry_rva` in `oe_sgx_enclave_properties`.
+
+With this change, the base enclave of an EEID enclave with heap/stack size and #TCS selected by the loader can safely use  `NumStackPages=0`, `NumHeapPages=0`, and `NumTCS=0`. Recording the expected `TCS.OENTRY` value in `oe_sgx_enclave_properties` of a SGX enclave (regular enclave or EEID enclave) also simplifies the implementation of Enclave Dynamic Memory Management (EDMM) based thread context creation support in the future, where the enclave code needs to verify the content of a TCS page added by the OS.
```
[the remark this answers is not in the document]
bodzhang: If the base enclave has no TCS, without including the entry point in the initialized data (measured) and checking `entry_point` in EEID data against it, a malicious loader can set `entry_point` to a different value than the enclave developer intended, and add TCS page(s) with the spoofed value. The EEID attestation verifier will see all measurement checks pass.

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,182 @@
...
+
+The claims produced by the enclave attestation verification should include config_id and config_svn, as well as the base enclave's identity information from the base enclave signature. The EEID enclave attestation verification process confirms that the enclave loader follows the SW convention to add the EEID page (in EEID enclave with loader selected heap/stack size and #TCS, also the heap and thread sections) after the base image, by checking the expected measurements. So the base enclave's identity information recorded in the EEID page can be trusted to reflect the base enclave loaded.
+
+As the enclave developer owns the definition of config_id and config_svn, the developer is responsible for providing the configuration data identified by config_id/config_svn to the Relying Party, if it's necessary for the Relying Party to analyze the configuration data. The Relaying Party should make sure the identity of the configuration data received matches config_id/config_svn claims produced by the enclave attestation verification process.
```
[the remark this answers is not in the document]
bodzhang: fixed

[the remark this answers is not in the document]
bodzhang: What the enclave developer intended might not be hash(config_data)=`config_id`. If a developer wants to implement his/her own design of the relationship between config_id/config_svn and content (not necessarily limited to "config_data") to be loaded into the enclave post initialization, what's the mechanism to override the OE SDK lib or calling convention that forces hash(config_data)=`config_id`?

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
bodzhang: Addressed

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,182 @@
...
+| Stack pages                     |
+| Guard Page                      |
+| TCS Page                        |
+| SSA (State Save Area) 0         |
+| SSA (State Save Area) 1         |
+| Guard Page                      |
+| Thread local storage            |
+| FS/GS Page (oe_sgx_td_t + tsp)  |
+
+For a regular enclave, the enclave loader "adds" a Guard Page between the Code and Initialized Data pages and the Heap Pages (`EADD` the first heap page at the linear address = linear address of the last Code/Data page + 4KB).
```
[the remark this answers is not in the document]
bodzhang: "adds" highlighted in quotes emphasizes that a guard page is not actually added to the EPC. A guard page is basically created by advancing the linear address of the next page to be added to EPC.

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,182 @@
...
+
+For EEID enclaves with heap/stack size and #TCS selected by the loader at enclave loading time, the loader adds heap pages, thread sections and the EEID page after the base enclave's code and initialized data are measured by the CPU. The `hash_state` records the measurement before the heap and thread sections are added. This type of of EEID enclave's Enclave_Size in `SECS` is fixed as a large value, for example 64GB.
+
+As discussed in the original EEID design doc(URL), an EEID enclave with heap/stack size and #TCS selected by the loader at enclave load time needs to have one thread section in the base image to capture the expected `TCS.OENTRY` value in the base enclave measurement, which does complicate the base enclave measurement recreation algorithm used in EEID enclave attestation verification. This updated design addresses the need to capture the expected `TCS.OENTRY` value in a different way.
```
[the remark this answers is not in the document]
bodzhang: fixed

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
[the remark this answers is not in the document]
wintersteiger: I must be missing something here. With static memory sizes this is not a problem (at signing or creation time, as long as numTCS > 0). With dynamic memory settings, the final measurement (resigned, at creation time) does include the TCS's (and numTCS > 0). So in both settings, if the host puts a different entry point into the TCSs, attestation would fail, because the verifier expects the EEID entry point (see `oe_remeasure_memory_pages`). The base image hash would match (with numTCS=0), but the extended hash would not. numTCS is also in the EEID settings, maybe the verifier should check that it's not 0?

dthaler:
I believe the SIG-Attestation discussion concluded saying the content should be merged into the ExtendedEnclaveInitializationData.md file with the result being that that file contains only up-to-date information.

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,323 @@
+Extended Enclave Initialization Data V2
+=======================================
+
+In this paper, we present an updated design of the extended enclave
+initialization data (EEID).
```
dthaler, line 5: ... for SGX

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,323 @@
...
+                              end of the enclave base image */
+    uint64_t signature_size; /* size of signature */
+    oe_enclave_size_settings_t size_settings; /* heap, stack and thread
+                                                 configuration for this
+                                                 instance */
+    uint64_t vaddr;          /* location of the added data pages in enclave
+                                memory; EEID follows immediately thereafter */
+    uint64_t entry_point;    /* entry point of the image, for SGX, matches
+                                TCS.OENTRY */
+    uint64_t data_size;      /* size of application EEID */
```
dthaler, line 77: The comment on this line doesn't match the comment on data[] on the next line

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,323 @@
...
+#ifdef EXPERIMENTAL_EEID
+typedef struct oe_eeid_t_
+{
+    uint32_t version;        /* version number of the structure */
+    uint32_t hash_state[10]; /* internal state of the hash computation at the
+                                end of the enclave base image */
+    uint8_t config_id[32];   /* identity of the code/data allowed to be
+                                loaded into the enclave after enclave init */
+    unit16_t config_svn;     /* minimum SVN of the code/data allowed to be
+                                loaded into the enclave after enclave init */
```
dthaler, line 96: config_svn appears in sgx_key_request_t.  Does it really apply to all TEEs?  If so, should expand the "SVN" acronym in the comment.

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
bodzhang: fixed

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,323 @@
...
+and `NumTCS=0`. Recording the expected `TCS.OENTRY` value in
+`oe_sgx_enclave_properties` of a SGX enclave (regular enclave or EEID enclave)
+also simplifies the implementation of Enclave Dynamic Memory Management (EDMM)
+based thread context creation support in the future, where the enclave code
+needs to verify the content of a TCS page added by the OS.
+
+Preserving base enclave properties in the signature
+---------------------------------------------------
+
+When res-signing the image using the `OE_DEBUG_SIGN_KEY`, the dynamically
```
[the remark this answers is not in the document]
bodzhang: Typo. Should be "re-signing". Fixed.

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,323 @@
...
+- Create a SHA256 context and restore the internal state to `hash_state`,
+- re-create the base image hash from the SHA256 context
+- measure the additional pages added after the base image. For EEID enclave with
+  the size of heap/stack and #TCS specified at enclave signing time, the only
+  additional page is the `oe_eeid_t` page. For EEID enclave with the size of
+  heap/stack and #TCS selected by the enclave loader at enclave loading time,
+  the additional pages include the heap and thread sections with stack using the
+  `size_settings` in `oe_eeid_t`. Note that the number of TCS pages determines
+  the number of thread sections. The measurement of the heap and thread section
+  starts at address `vaddr`+4KB and includes `entry_point` in the TCS control
```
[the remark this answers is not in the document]
bodzhang: `oe_eeid_t` definition is SGX-specific.

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,348 @@
...
+/* Extends oe_enclave_properties_header_t base type */
+typedef struct _oe_sgx_enclave_image_info_t
+{
+    uint64_t oeinfo_rva;
+    uint64_t oeinfo_size;
+    uint64_t reloc_rva;
+    uint64_t reloc_size;
+    uint64_t heap_rva; /* heap size is in header.sizesettings */
+    uint64_t enclave_size;
+    uint64_t entry_rva; /*from oeimage.entry_rva, must match TCS.OENTRY */
```
CodeMonkeyLeet, line 261: I was thinking about this a bit more and I don't think this strategy works. The problem is that this implies that the enclave itself will validate that the measured `entry_rva` matches the TCS.OENTRY, but:
1. it requires a call-in to transfer control to the enclave, by which time a modified TCS.OENTRY would have already pre-empted the check. 
2. the enclave itself does not have read access to the contents of the TCS page, per PT_TCS definition. 

I'm not sure that you can get around having to measure and sign at least one TCS in the signing of the base enclave.

wintersteiger: Yeah, I also found that the TCS page isn't accessible, but I thought this would have to change for Icelake, because otherwise there's no way to TCSs that are dynamically added at runtime. Perhaps those pages would have to be added as a normal heap-type message and then the enclave changes it to `PT_TCS` from inside, a bit like EACCEPT'ing it but with the additional check?

I think it's fine if we keep the 0/0/1 settings as they are now, but it would of course be nice to change that to 0/0/0 if we find a way.

bodzhang: The enclave code checks the measured entry_rva in `_or_sgx_enclave_image_info_t` against the value set by the loader in the eeid page (not agianst the value in TCS), before any additional content is loaded into the enclave. But I do realize that it requires further security analysis to understand whether EENTERing the enclave at an unexpected entry point through a TCS added by a malicious loader can bypass the checking logic somehow. I aslo think we can start with 0/0/1 settings, if the security analysis shows that the same attack will be detected for sure with the 0/0/1 settings.

EDMM does suport coverting regualr page to TCS. That's exactly how a TCS page will be added/acceptted.

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
bodzhang: With the SGX KSS feature, EGETKEY keys can be bound to Config_id and config_svn, based on enclave code's selection (sgx_key_request_t). EEID's config_id/config_svn will not provide the same capability. The latest update to the design doc defines eeid.config_svn as "Optional subsidiary info about the identity of the additional content, for example, a version number".

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
bodzhang: I discovered more potential attack techniques. The latest update to the design doc introduces an "eeid measurement context" page to be added to the base enclave measurement. The context page include all loader controlled/provided setting, outside of the SW convention, needed by the eeid enclave attestation verifier to recreate the HW reported `MRENCLAVE` and the base enclave's `SIGSTRUCT.ENCLAVEHASH`. If a malicious loader spoofs those settings, the base enclave measurement recalculated based on the spoofed settings will not match `SIGSTRUCT.ENCLAVEHASH`.

docs/DesignDocs/ExtendedEnclaveInitializationData_V2.md
```diff
@@ -0,0 +1,424 @@
...
+|-----|-------|---------|-----------------------------------
+|  -  |   -   |   -     | On system where SGX-KSS feature is not available or disabled: N/A; On system with SGX-KSS enabled: loader sets SECS.config_id and SECS.config_svn as 0
+|  -  |   x   |   -     | On system where SGX-KSS feature is not available or disabled: Invalid;  On system with SGX-KSS enabled only: loader copies PREID to SECS.config_id and SECS.config_svn
+
+*Static sizing EEID SGX Enclave*
+
+| DYN | PREID | PREDATA | Behavior
+|-----|-------|---------|-----------------------------------
+|  -  |   -   |   x     | Static sizing EEID enclave only: loader sets eeid_page.config_id as PREDATA hash, and eeid_page.config_svn as 0, enclave runtime loads/verifies PREDATA
+|  -  |   x   |   -     | static sizing EEID enclave only: loader copies PREID to SECS.config_id and SECS.config_svn
```
CodeMonkeyLeet, line 233: Per SIG-Arch meeting, can you correct the wording? This doesn't write to SECS.config_* fields.

docs/DesignDocs/ConfigIDInterface.md
```diff
@@ -0,0 +1,210 @@
...
+typedef struct _oe_enclave_pre_measured_id
+{
+    /** The identity of additional content allowed to be loaded
+     *  into the enclave post enclave initialization. The identity is covered
+     *  by TEE generated Enclave Attestation Evidence.
+     */
+    uint8_t config_id[64];
+    uint16_t config_svn;
+}
+oe_enclave_pre_measured_id_t;
```
dthaler, line 60: What does "pre measured id" really mean?  Is the id itself before measurement?  Is the id an "identity" (like a key) or just an "identifier"?

docs/DesignDocs/ConfigIDInterface.md
```diff
@@ -0,0 +1,209 @@
...
+ * @retval OE_INVALID_PARAMETER At least one of the parameters is invalid.
+ * @retval OE_NOT_FOUND The input evidence format id is not supported.
+ * @retval other appropriate error code.
+ */
+oe_result_t oe_get_evidence(
+    const oe_uuid_t* format_id,
+    uint32_t flags,
+    const void* runtime_custom_claims_buffer,
+    size_t runtime_custom_claims_buffer_size,
+    const oe_inittime_custom_claim_buffer_t* inittime_custom_claims_buffer,
```
CodeMonkeyLeet, line 185: It's unclear to me why this needs to be a separate set of parameters in the `oe_get_evidence` function itself:

- From the attester side, if the desire is simply to bind an extra buffer to the resulting evidence blob, then including it as part of the existing `custom_claims_buffer` would be sufficient.
   - If syntactic sugar is desired for structuring the buffer from name value pairs, it would be preferable to shift that functionality into the helper functions like `oe_serialize_custom_claims`
- If the intent is to imply special tagging of the information for handling by an enlightened verifier plugin in the implementation of `oe_verify_evidence`, it seems like the verification of the integrity of a custom claim against information in the evidence can be better generalized across TEEs by pushing it into a separate function instead tightly coupling it to basic evidence structure verification performed by `oe_verify_evidence`. 
    - In the parlance of the IETF RATS aligned [AttestationApiRequirements.md](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/AttestationApiRequirements.md), this might be something in the `appraise_evidence`/`set_evidence_appraisal_policy` implementation on the verifier side.

yentsanglee:
If I understand the design correctly, the steps to specify inittime claims as an OE SDK consumer are:
   1) At enclave creation, set config_id as SHA256(inittime_claims)
   2) At each evidence generation during the life time of the enclave, pass in the exact initime_claims to oe_get_evidence().
   
Comments:
1. If the same inittime_claims has to be passed in to each oe_get_evidence(), why not adding a new API such as oe_set_inittime_claims() or making it an input of oe_create_enclave() similar to EEID? Calling it an "inittime" but passing it in at runtime seems inconsistent.
2. We can let OE SDK handle the integrity of inittime_claims (e.g. saving hash(inittime_claims+runtime_claims) to report_data). Free up config_id for other purpose.
3. At oe_verify_evidence(), inittime_claims is exposed as an entry in the claim array similar to OE_CLAIM_CUSTOM_CLAIMS_BUFFER.

docs/DesignDocs/ConfigIDInterface.md
bodzhang: I will create a new PR for config_id/config_svn later, after the initial discussion already started in this PR reaches a good stopping point.

docs/DesignDocs/ConfigIDInterface.md
bodzhang: The caller of the `oe_get_evidence` API has different expectations on how the attester plugin handles the run-time claims and  init-time claims. The plugin is supposed to include the run-time claim in the evidence in a way that provide integrity protection of the run-time claims. For example, the SGX Attester plugin sets the `reportdata` field of SGX REPORT/QUOTE as SHA256 hash of the run-time claims. The plugin is supposed to simply include the init-time claims in the evidence and leave the integrity protection responsibility to the underlining TEE and enclave SW outside of the plugin. For example, the SGX enclave developer implements enclave code that bind the init-time claim buffer to SGX ConfigID/ConfigSVN. To delineate the run-time claims vs. init-time claims, either the `oe_get_evidence` API supports different input parameters for run-time claims and init-time claims or the `oe_serialize_custom_claims` API does that. I think either approach can work. But if `oe_serialize_custom_claims` API syntax is considered TEE or plugin-implementation specific, the caller would have to write different calling code for different `oe_serialize_custom_claims` API syntax. As `oe_get_evidence` API syntax is considered TEE agonistic and plugin implementation agonistic, adding the dedicated init-time claims related input parameters in `oe_get_evidence` API enables the caller to have the same calling code for different TEEs and plugin implementations that support init-time claims concept.

The other part of the proposal in this PR - including an init-time claims integrity-protection algorithm identifier in the init-time claim buffer, is to enable enlightened verifier plugins to verify the integrity of the init-time claims included in the evidence. The goal of the API syntax design is again to enable the caller of the attester side APIs to have the same calling code for different TEES and plugin implementations. The proposal in this PR defines the init-time claims buffer passed to `oe_get_evidence` API as a structure containing the integrity-protection algorithm identifier. Alternatively, a dedicated init-time claims integrity-protection algorithm identifier input parameter can be added to `oe_get_evidence`, or to `oe_serialized_custom_claims`, if `oe_serialized_custom_claims` API is defined as TEE agonistic and plugin implementation agonistic. The verifier side policy about whether `oe_verify_evidence` should verify the integrity of the init-time claims is independent of the the attester side API syntax proposal. For example, instead of `oe_verify_evidence` verifies the integrity of the init-time claims, `oe_verifies_evidence` only outputs unverified init-time claims, and `appraise_evidence` verifies the init_time claims against the Config_ID base claim. But the requirement is that the init-time claims integrity algorithm identifier is provided to the verifier side, within the attestation evidence produced by `oe_get_evidence`.

docs/DesignDocs/ConfigIDInterface.md
```diff
@@ -0,0 +1,209 @@
...
+    OE_ENCLAVE_SETTING_CONTEXT_PRE_MEASURED_ID = 0x976a8f67,
+} oe_enclave_setting_type_t;
+
+typedef struct _oe_enclave_pre_measured_id
+{
+    /** The identity of additional content allowed to be loaded
+     *  into the enclave post enclave initialization. The identity is covered
+     *  by TEE generated Enclave Attestation Evidence.
+     */
+    uint8_t config_id[64];
```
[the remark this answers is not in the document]
bodzhang: @CodeMonkeyLeet , I like the idea of generalizing the data structure definition of "extra content" identifier. How about a simple variable length buffer?
```C
{
  uint8_t identifier_size;
  uint8_t identifier[];
}
```
We still need a better name for the structure though.

bodzhang:
@yentsanglee, the "extra content" associated with the init-time claims is loaded into the enclave post enclave initialization, by the "static" enclave code ("static" in the sense that its identity is captured in the enclave measurement established at enclave initialization, before any enclave code runs). The "extra content" might be code/data that influences enclave behavior, thus its identity need to be captured in attestation. In enclave TEEs that do not provide privilege isolation among code modules, any claim about the extra code's identity can't be part of the run-time claims, as the "extra content" has the ability to produce the run-time claims and spoof the identity value. SGX `REPORT.report_data` is set by the enclave code itself at run-time. It does not provide a mechanism to capture an immutable identity of the extra code loaded into the enclave after enclave initialization. In SGX, config_id/config_svn is used to capture the immutable identity of the "extra content", before the "extra content" is loaded into the enclave. The "static" code of the enclave is responsible to verify the "extra content" to be loaded matches the identity captured in config_id/config_svn. Enclave developer defines the binding between the "extra content" and config_id/config_svn. Developers might choose to use SHA256(extra code) as the identity and set config_id=SHA256(extra code), or SHA256(manifest of the extra code) as the identity, where the manifest might be the signature/cert of a signed code module. Another example, an enclave might be configured with the public key of an external service it will trust and use, at runtime (not hardcoded as part of the "static" code). In this example, the developer might set config_id as the public key or a hash of the public key. Developers can use many other type of schemes or crypto algorithms. Similar to run-time claims buffer bound to report_data,  init-time claims buffer bound to config_id/config_svn is implementation specific, and a Relying Party might analyze it against certain trust policy. A Verifier plugin can verify the integrity of the init-time claims buffer against config_id/config_svn and output the verified init-time claims, as long as the plugin can delineate the init-time claims buffer within the evidence buffer, and understand/support the binding algorithm between config_id/config_svn and the init-time claims buffer.

docs/DesignDocs/ConfigIDInterface.md
bodzhang: Based on the discussion in the Attestation arch SIG meeting, my understanding is that `oe_serialized_customer_claims` is supposed to be implemented by the enclave developer. In that case, the custom_claims_buffer's internal structure, for example, to delineate the init-time claims and run-time claims, is opaque to the Attester and Verifier Plugin. The Attester Plugin won't be able to locate the run-time claims portion of the custom_claims_buffer to create the integrity-protection metadata (in SGX Attester Plugin, it's SGX `REPORT.report_data`). If the Attester creates the integrity-protection metadata over the whole custom_claim_buffer, the operation only introduces unnecessary extra latency to operate on the init_time claims, with no security benefit.  More importantly, the Verifier Plugin will not be able to locate the run-time claims buffer to verify against the Config_ID.

Pushing the responsibility to verify init-time claims out of `oe_verifies_evidence`/`appraise_evidence` might be OK. But the upper level logic, for example, an Attestation Verification Service, such as [MAA](https://docs.microsoft.com/en-us/azure/attestation/overview) will need to handle that complexity.

docs/DesignDocs/ConfigIDInterface.md
bodzhang: Created [PR#3726](https://github.com/openenclave/openenclave/pull/3726) for config_id/config_svn.

docs/DesignDocs/ConfigIDInterface.md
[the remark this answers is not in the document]
bodzhang: In [PR#3726](https://github.com/openenclave/openenclave/pull/3726), name the structure as `oe_enclave_inittime_configuration_data`

docs/DesignDocs/ConfigIDInterface.md
bodzhang: [PR#3726](https://github.com/openenclave/openenclave/pull/3726) discussed the alternative of leaving the packaging and verification of the customer_claims_buffer to the upper level SW outside of the Attester/Verifier Plugin.

wintersteiger:
@bodzhang what's your position on this design document? Do you intend to go ahead with the plans for configid in a separate effort (#3735?) and remove those parts here, or do you still want both EEID and configid to be handled by the same (or very similar) interface?

The new [inittime design doc](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/InitTimeConfigurationInterface.md) suggests two implementation paths; is there any consensus on whether we should modify the plugin API to accommodate init-time claims yet?

Also, if you can recall, are there any major changes to this design that are not implemented in #3236 yet?

radhikaj:
@bodzhang , I am closing this and we can reopen when we refine the design.

PR #3253, opened 2020-07-10
<WIP>: Same build of OE SDK for both FLC/non FLC machines

anakrish:
This allows OE SDK packages to be built on non FLC machines.

Signed-off-by: Anand Krishnamoorthi <anakrish@microsoft.com>
fixes #3178

PR #3270, opened 2020-07-15
Cleanup: OE_LINK_SGX_DCAP_QL -> OE_HAS_SGX_DCAP_QL

anakrish:
The SDK no longer links against sgx_dcap_ql.so; instead
it is dynamically loaded.

Refer to #3272 for the bigger picture.

### DIFF
--- a/enclave/core/CMakeLists.txt
+++ b/enclave/core/CMakeLists.txt
@@ -321,7 +321,7 @@ enclave_compile_definitions(
   $<BUILD_INTERFACE:OE_API_VERSION=2>)
 
 if (HAS_QUOTE_PROVIDER)
-  enclave_compile_definitions(oecore PUBLIC OE_LINK_SGX_DCAP_QL)
+  enclave_compile_definitions(oecore PUBLIC OE_HAS_SGX_DCAP_QL)
 endif ()
 
 if (USE_DEBUG_MALLOC)
--- a/common/sgx/verifier.c
+++ b/common/sgx/verifier.c
@@ -16,7 +16,7 @@
 #include "quote.h"
 #include "report.h"
 
-#if defined(OE_LINK_SGX_DCAP_QL) && !defined(OE_BUILD_ENCLAVE)
+#if defined(OE_HAS_SGX_DCAP_QL) && !defined(OE_BUILD_ENCLAVE)
 #include "../../host/sgx/sgxquoteprovider.h"
 #endif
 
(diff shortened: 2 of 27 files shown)


anakrish:
bors try

enclave/core/CMakeLists.txt
```diff
@@ -321,7 +321,7 @@ enclave_compile_definitions(
   $<BUILD_INTERFACE:OE_API_VERSION=2>)
 
 if (HAS_QUOTE_PROVIDER)
-  enclave_compile_definitions(oecore PUBLIC OE_LINK_SGX_DCAP_QL)
+  enclave_compile_definitions(oecore PUBLIC OE_HAS_SGX_DCAP_QL)
```
mingweishih, line 324: Just curious, is there a way to determine whether the dcap library is loaded or not during the runtime? I'm asking because `OE_HAS_SGX_DCAP_QL` is still used at compile-time.

anakrish: We treat inability to load sgx_dcap_ql as a catastrophic error and terminate the application:
https://github.com/openenclave/openenclave/blob/ea85ec5403042e3937d2f531957e4e04a25bdb8a/host/sgx/sgxquote.c#L117

The expectation is that quoting is not requested in platforms that don't have sgx_dcap_ql.

mingweishih: Thanks for the clarification!

anakrish:
> This change LGTM, but is it necessary given #3253?
> 
> `HAS_QUOTE_PROVIDER` seems to serve a dual purpose of code elision (for setting `OE_HAS_SGX_DCAP_QL` in tests, existing host/common code) and cmake package/install configuration (hostverify vs. sdk as used in samples CMakeLists.txt). The latter still seems relevant in terms of build scoping, but it looks like we could remove `OE_HAS_SGX_DCAP_QL` entirely by simply not building the tests that need it (it looks like the else clause is to skip the whole test or indicate a tool is nonfunctional) instead of skipping lines of code.

I'm trying to check-in individual self-contained changes from #3253 .

In one of the subsequent PRs, we'll remove OE_HAS_SGX_DCAP_QL entirely from the libraries (oecore, oehost etc).
Yes, we could remove OE_HAS_SGX_DCAP_QL entirely; but based on initial investigation the tests might require a bit of work.
Some tests like tests/report selectively execute checks based on OE_HAS_SGX_DCAP_QL. I'm not yet sure whether we should instead convert it to a runtime check for availability of dcap shared libraries or retain it as a compile time check.

anakrish:
bors try

anakrish:
bors try

Issue #3272, opened 2020-07-15
Meta-Issue: Changes required to enable FLC/Non-FLC in one build


anakrish:
[FLC/Non-FLC in one build?](https://github.com/openenclave/openenclave/issues/3178) requires changes to the OE SDK to _allow building a single OE variant that works across all SGX variants (currently FLC/non-FLC) with appropriate runtime introspection._

This meta-issue captures that changes that would be needed to accomplish that goal.

- [x] **Necessary Cleanups** 
  Currently all the SGX libraries are dynamically loaded rather than being linked in, as used to be the case. The code base needs to be cleaned up to reflect this.
  - [x] Rename `OE_LINK_SGX_DCAP_QL`  to `OE_HAS_SGX_DCAP_QL` 
     PR #3270 
- [x] **Remove use of HAS_QUOTE_PROVIDER when building SDK libraries**
    This will ensure that the same libraries are produced whether SGX libraries are present in the build system or not.
    Attestation will be available only if the target system (the system in which enclaves are run) has the necessary quote provider libraries.
    PR #3273 
    Note: At this stage, HAS_QUOTE_PROVIDER will still be a cmake option. But internally it won't be used to build the libraries.
- [x] **Remove dependency on SGX headers when build SDK libraries**
  This will allow  the SDK libraries to be built on systems that don't have SGX headers installed.
  PR #3277 
- [x] **Remove use of OE_LINK_SGX_DCAP_QL** from tests.
  Some tests use this define to selectively perform assertions. Needs an investigation to determine if the tests can instead be converted to use a runtime query instead of a define to determine what assertions to perform.
PR #3282
- [x] **Remove use of HAS_QUOTE_PROVIDER** from package installation.
  PR #3288 
- [x] **Remove HAS_QUOTE_PROVIDER** as a cmake option.
  Tests will be enabled/disabled automatically based on the presence of quoting libraries on the system.
   PR #3288
- [ ] **Add non/SGX, non FLC machines to CI**
- [ ] **Update ansible and prereqs scripts**
   Update the scripts to remove dependencies if possible.
- [x] **Update docs** Remove references to HAS_QUOTE_PROVIDER.
  PR #3288

anakrish:
#3270 Renames `OE_LINK_SGX_DCAP_Q` to `OE_HAS_SGX_DCAP_QL`

anakrish:
[...]

PR #3277, opened 2020-07-16
Remove dependency on sgx_dcap_ql_wrapper.h for building SDK.

anakrish:
With this change, OE SDK with full attestation support,
can be built on a machine without any Intel SGX SDK headers.

This is a prerequisite for removing HAS_QUOTE_PROVIDER/OE_HAS_SGX_DCAP_QL
cmake option/define when building OE SDK.

See Meta-Issue: #3272 

### DIFF
--- a/host/sgx/sgxquote.c
+++ b/host/sgx/sgxquote.c
@@ -4,11 +4,11 @@
 
 #include "sgxquote.h"
 #include <openenclave/attestation/sgx/evidence.h>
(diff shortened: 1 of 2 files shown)


anakrish:
bors try

anakrish:
@letmaik FYI, with this PR the last remaining build-time dependency on SGX headers is removed.

Issue #3174, opened 2020-06-23
Can an app consume an OE SDK enclave and an Intel SDK enclave?


dthaler:
Can the same host application load an OE enclave and an SGX SDK enclave at the same time. That is,
Use Intel edger8r on enclaveA from some other ISV that provided an enclave in binary form
Use OE edger8r on enclaveB
Can you link the resulting untrusted stubs into the same host application, or are there any name collisions?

If there exist Intel SDK-generated enclaves out there today, this scenario can arise when someone wants to start migrating their own apps, or creating new apps, to the OE SDK.  If the same app calls multiple enclaves from third-parties, allowing them to move one-by-one rather than waiting until all the enclaves are OE SDK, would provide a faster transition path to OE SDK.

dthaler:
This question came up as part of the discussion around PR #3086 which solves a problem only if the problems are all OE SDK enclaves.

radhikaj:
Add that this is not supported in documentation.

radhikaj:
This can be covered by #3239.

PR #3280, opened 2020-07-17
Docs:App cannot consume an OE SDK and an Intel SDK enclave

radhikaj:
#3174 tracks validating that an application can consume both an Intel SGX SDK enclave and an OE SDK enclave. At this time, this scenario is untested. Adding a note to this effect per discussion in triage.


radhikaj:
The poring guide to port from Intel SGX SDK to OE SDK should address this.

PR #3282, opened 2020-07-17
Remove OE_HAS_SGX_DCAP_QL define.

anakrish:
Remove OE_HAS_SGX_DCAP_QL define

- Use runtime check for quote provider instead of compile time check.
- Remove use of HAS_QUOTE_PROVIDER and OE_HAS_SGX_DCAP_QL from tests.
- Remove use of OE_HAS_SGX_DCAP_QL from common/sgx/verifier.c
  that was missed in an earlier PR
- Remove OE_HAS_SGX_DCAP_QL from VS extension

### DIFF
--- a/host/sgx/tests.c
+++ b/host/sgx/tests.c
@@ -1,0 +1,21 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/host.h>
+#include <openenclave/internal/sgx/tests.h>
+#include "../hostthread.h"
+#include "sgxquoteprovider.h"
+
+static bool _has_quote_provider = false;
+
+static void _check_quote_provider(void)
+{
+    _has_quote_provider = (oe_initialize_quote_provider() == OE_OK);
+}
+
+bool oe_has_sgx_quote_provider(void)
+{
+    static oe_once_type once = OE_H_ONCE_INITIALIZER;
+    oe_once(&once, _check_quote_provider);
+    return _has_quote_provider;
+}
--- a/tests/debug-mode/host/host.c
+++ b/tests/debug-mode/host/host.c
@@ -4,6 +4,7 @@
 #include <openenclave/host.h>
 #include <openenclave/internal/calls.h>
 #include <openenclave/internal/error.h>
+#include <openenclave/internal/sgx/tests.h>
 #include <openenclave/internal/tests.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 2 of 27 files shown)


anakrish:
bors try

host/sgx/tests.c
```diff
@@ -0,0 +1,21 @@
...
+#include "sgxquoteprovider.h"
+
+static bool _has_quote_provider = false;
+
+static void _check_quote_provider(void)
+{
+    _has_quote_provider = oe_initialize_quote_provider() == OE_OK;
+}
+
+bool oe_has_quote_provider(void)
```
dthaler, line 16: Please add sgx into the function name
```suggestion
bool oe_has_sgx_quote_provider(void)
```

anakrish: Good point. Will do.

anakrish:
bors try

anakrish:
bors try

tests/debug-mode/host/host.c
```diff
@@ -68,10 +68,11 @@ static void _test_debug_signed(const char* path)
 {
     /* Signed debug mode should always pass. */
     _launch_enclave_success(path, _create_flags(true));
-#ifdef OE_HAS_SGX_DCAP_QL
-    /* Only works with FLC */
-    _launch_enclave_success(path, _create_flags(false));
-#endif
+    if (oe_has_sgx_quote_provider())
```
CodeMonkeyLeet, line 71: This is an unfortunate example of code semantics having shifted out from under this original implementation. This was originally intended as a compile time check for whether the project was compiled for SGX1 or SGX1FLC configurations, but the flag has since been repurposed after the loader stack stopped making that distinction.

What this test cares about is that the launch control stack, and whether the quote provider is installed or not is irrelevant. For CI/CD purposes, it just so happens they coincide, so this is a proxy signal for the whether it should run the production signed launch test or not. 

We should file a separate issue to have this fixed up as a CI/CD task:
- [ ] This test should take an explicit environment variable for `SKIP_NON_DEBUG_LAUNCH` and choose to execute these tests or not
- [ ] CI/CD infrastructure should set this environment variable for the SGX1 configuration pipelines.

include/openenclave/internal/tests.h
```diff
@@ -72,6 +72,11 @@ OE_EXTERNC_BEGIN
  */
 uint32_t oe_get_create_flags(void);
 
+/*
+ * Return whether SGX quote provider libraries are available in the system.
+ */
+bool oe_has_sgx_quote_provider(void);
```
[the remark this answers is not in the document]
anakrish: Good suggestion. I ended up creating internal/sgx/tests.h since there was no existing platform specific define I could readily use.

tests/debug-mode/host/host.c
anakrish: Good catch!
I have created issue #3287

PR #3285, opened 2020-07-20
Use relative paths to compilers when configuring LVI mitigations

eddyashton:
Trying to consume the LVI mitigated targets from an external project, I hit the following error:

```
CMake Error at /opt/openenclave/lib/openenclave/cmake/configure_lvi_mitigation_build.cmake:73 (message):
  -- /usr/bin/clang-8 is not found.
Call Stack (most recent call first):
  /opt/openenclave/lib/openenclave/cmake/openenclave-lvi-mitigation-config.cmake:43 (configure_lvi_mitigation_build)
  cmake/ccf_app.cmake:46 (find_package)
  cmake/common.cmake:108 (include)
  CMakeLists.txt:38 (include)


-- Configuring incomplete, errors occurred!
```

You will get the same error if `CC` or `CXX` are passed as environment variables when building OE with LVI, eg:

```
CC=/usr/bin/clang-8 cmake .. -GNinja -DUSE_SNMALLOC=ON -DLVI_MITIGATION=ControlFlow -DLVI_MITIGATION_BINDIR=/data/src/openenclave/build/lvi_mitigation_bin
```

The problem is this line in `configure_lvi_mitigation_build.cmake`:

```
  if (EXISTS "${OE_BINDIR}/${C_COMPILER}")
```

This path concatentation doesn't work if `C_COMPILER` is an absolute path, which it often is if its come from CC/CXX.

Not sure if this is the _right_ fix, but it seems to work.

Signed-off-by: Eddy Ashton <edashton@microsoft.com>

### DIFF
--- a/cmake/configure_lvi_mitigation_build.cmake
+++ b/cmake/configure_lvi_mitigation_build.cmake
@@ -86,6 +86,6 @@ function (configure_lvi_mitigation_build)
         ${OE_BINDIR}/${CXX_COMPILER}
         PARENT_SCOPE)
   else ()
-    message(FATAL_ERROR "-- ${CXX_COMPILER} is not supported.")
+    message(FATAL_ERROR "-- ${OE_BINDIR}/${CXX_COMPILER} is not supported.")
   endif ()
 endfunction ()
--- a/cmake/configure_lvi_mitigation_build.cmake
+++ b/cmake/configure_lvi_mitigation_build.cmake
@@ -60,7 +60,7 @@ function (configure_lvi_mitigation_build)
   # Overwrite the default C compiler if CC is explicitly specified.
   # Otherwise, select the compiler based on the detection logic.
   if (DEFINED ENV{CC})
-    set(C_COMPILER $ENV{CC})
+    get_filename_component(C_COMPILER $ENV{CC} NAME)
   else ()
     detect_compiler(${OE_BINDIR} C)
   endif ()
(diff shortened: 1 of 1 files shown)


cmake/configure_lvi_mitigation_build.cmake
```diff
@@ -70,13 +70,13 @@ function (configure_lvi_mitigation_build)
         ${OE_BINDIR}/${C_COMPILER}
         PARENT_SCOPE)
   else ()
-    message(FATAL_ERROR "-- ${C_COMPILER} is not found.")
+    message(FATAL_ERROR "-- ${OE_BINDIR}/${C_COMPILER} is not found.")
```
mingweishih, line 73: For the consistency, can you also update the message for the CXX compiler?

anakrish:
@mingweishih Can this be locked down with a test?

cmake/configure_lvi_mitigation_build.cmake
eddyashton: Sure, now added on line 89.

mingweishih:
> @mingweishih Can this be locked down with a test?

This cmake script is only executed when building the sdk. We currently don't have tests for such cases (i.e., testing the cmake functions).

cmake/configure_lvi_mitigation_build.cmake
mingweishih: Thanks!

anakrish:
> > @mingweishih Can this be locked down with a test?
> 
> This cmake script is only executed when building the sdk. We currently don't have tests for such cases (i.e., testing the cmake functions).

Could this then be locked down by exercising the scenario via the Jenkins scripts?

mingweishih:
> > > @mingweishih Can this be locked down with a test?
> > 
> > 
> > This cmake script is only executed when building the sdk. We currently don't have tests for such cases (i.e., testing the cmake functions).
> 
> Could this then be locked down by exercising the scenario via the Jenkins scripts?

This in fact can be exercised by our sample tests. Currently, we don't really test our samples with various compiler options although we support it (e.g., the compiler detection in Make you added).

Alternatively, if we want to test this specific scenario (i.e., `CC=/path/to/compiler`), I'd suggest adding one job to the nightly build.

radhikaj:
@mingweishih @anakrish Can we merge this and then add to Jenkins or is adding to Jenkins a prerequisite to merging?

radhikaj:
Thanks @eddyashton

Issue #3287, opened 2020-07-20
tests/debug-mode


anakrish:
The following issue was identified as part of the code review for PR #3282 (See [comment](https://github.com/openenclave/openenclave/pull/3282/files#r457628197)):
-----------------------------------------------------------------------------------------------------------------------------------------------

This is an unfortunate example of code semantics having shifted out from under this original implementation. This was originally intended as a compile time check for whether the project was compiled for SGX1 or SGX1FLC configurations, but the flag has since been repurposed after the loader stack stopped making that distinction.

What this test cares about is that the launch control stack, and whether the quote provider is installed or not is irrelevant. For CI/CD purposes, it just so happens they coincide, so this is a proxy signal for the whether it should run the production signed launch test or not. 

We should file a separate issue to have this fixed up as a CI/CD task:
- [ ] This test should take an explicit environment variable for `SKIP_NON_DEBUG_LAUNCH` and choose to execute these tests or not
- [ ] CI/CD infrastructure should set this environment variable for the SGX1 configuration pipelines.

PR #3288, opened 2020-07-20
Remove HAS_QUOTE_PROVIDER cmake option.

anakrish:
~Instead of removing HAS_QUOTE_PROVIDER completely, deprecate it.
It will be removed in a subsequent release.~

After discussion, HAS_QUOTE_PROVIDER will be rmeoved instead of being
deprecated. More folks were in favor of removing it altogether.
Rationale: It is a cmake option used by contributors and not users.

Remove use of HAS_QUOTE_PROVIDER in determining which sampes will
be installed. This implies that the SDK package will be irrespective
of which platform (SGX vs non SGX, FLC vs non FLC) the SDK is built on.

Samples testing will look for presence of sgx_dcap_ql on the target system
(instead of the build system) and
selectively enable testing of attestation dependent samples.

Remove references to HAS_QUOTE_PROVIDER from docs.

Updaed CHANGELOG.md.

Remove use of HAS_QUOTE_PROVIDER from CMakeSettings.json and VS Code Extension

Remove unused scripts/test-build-config

### Tests affected removal of HAS_QUOTE_PROVIDER
The follow tests used to depend on HAS_QUOTE_PROVIDER=on/off to determine what to test.
Their use of HAS_QUOTE_PROVIDER was removed in previous PR #3282.

Open Enclave CI does not have SGX1 non FLC machines. It  mimics SGX1 non FLC machines by building the OE SDK with HAS_QUOTE_PROVIDER=off. When HAS_QUOTE_PROVIDER option removed, this would imply that SGX1 non FLC scenario will not be mimiced in CI anymore. However as shown by the following table, test coverage is not lost. Only the following tests differed between FLC vs non FLC machines; and in non FLC machines test points were being turned off in these tests.

Test Name |  Effect of HAS_QUOTE_PROVIDER=off | Impact of removing HAS_QUOTE_PROVIDER   | Test Coverage Loss
---------------|-------------------------------------------------------|------------------------------------------------------------------|---
tests/attestation_cert_apis | Testing skipped via conditional compilation | Test always executed | None
tests/attestation_plugin | Testing skipped via conditional compilation | Test always executed | None
tests/attestation_plugin_cert | Testing skipped via conditional compilation | Test always executed | None
tests/debug_mode | Test to detect whether enclaves can be launched successfully is skipped. | `launch_enclave_success` checks (2 of them in total) always executed. | None (Note: The test ought not have used HAS_QUOTE_PROVIDER See #3287)
tests/eeid_plugin| Entire test is skipped via conditional compilation | Test always executed on all non simulation SGX configurations WITH_EEID=on | None
tests/qeidentity | `run_parse_advisoryids_json_test`, `run_qe_identity_test_cases` and `run_qe_identity_v2_test_cases` skipped. Note: rest of the test does nothing other than create/terminate enclave. | Those test cases always executed | None
tests/report | Remote attestation testing is skipped. | Remote attestation is always tests. | None
samples   |  **remote_attestation** and **attested_tls** samples won't be tested. |   These samples will be tested on all non simulation SGX configurations in CI. | None
tests/tls_e2e | Testing skipped via conditional compilation | Test always executed | None
tests/tools/oecert | Testing skipped via conditional compilation | Test always executed | None
tests/tools/oecertdump | Testing skipped via conditional compilation | Test always executed | None


anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [Unreleased][Unreleased_log]
 --------------
 
+### Added
+- OE SDK Release packages can now be built on non SGX and non FLC machines.
```
CodeMonkeyLeet, line 14:
```suggestion
- OE SDK Release packages can now be built on non-SGX and non-FLC machines.
```

scripts/test-build-config
```diff
@@ -1,178 +0,0 @@
-#!/usr/bin/env bash
```
CodeMonkeyLeet, line 1: Thanks for the clean up!

samples/CMakeLists.txt
```diff
@@ -80,23 +80,34 @@ else ()
     RENAME README.md)
 endif ()
 
+# Determine if SGX DCAP libraries are available in the test system.
```
CodeMonkeyLeet, line 83: This is still an awkward way to run the tests, because it has to assume that the system that builds the OE SDK project and ctests is also the same system that the test run on, which the removal of the HAS_QUOTE_PROVIDER flag elsewhere would otherwise help decouple. (i.e. it enables forward looking scenarios like allowing build on normal GP machine pools, then transfer to SGX & DCAP enabled machine to test).

Instead of making this a build variable, have we considered making this an environment variable checked directly in samples/test-samples.cmake? It's something that we already do with the `OE_SIMULATION` flag, so it seems appropriate that we should just move this one out to a run time setting we can tweak in CI/CD directly.

CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Changed
+
+### Removed
+- HAS_QUOTE_PROVIDER is deprecated as a cmake option. It has not effect on SDK build.
```
CodeMonkeyLeet, line 19: I'm not entirely sure this needs deprecation ... have we exposed this flag for public use by the developer? If it's just an internal build flag, we can just get rid of it.

Also typo "not" -> "no"

CMakeLists.txt
```diff
@@ -200,11 +200,13 @@ include(package_settings)
 # See `cmake/add_enclave.cmake` for enclave creation logic
 include(add_enclave)
 
-# User configurable options.
-option(
-  HAS_QUOTE_PROVIDER
-  "Take a build dependency on SGX DCAP, which requires FLC on target device to run."
-  ON)
+# HAS_QUOTE_PROVIDER option is deprecated.
```
achamayou, line 203: fwiw, cmake already gives a message about flags/options that aren't used, eg.

```
cmake -DFOO=BAR ..
...
  Manually-specified variables were not used by the project:

    FOO
```

Considering this was never officially supported, I think that would be acceptable too.

samples/CMakeLists.txt
achamayou: In principle, re-running cmake with the right flag for the target platform before triggering ctest ought to be enough to address those cases, since it acts incrementally, assuming no rebuild is needed (but then transferring the binaries wouldn't work). The nice thing about cmake flags compared to env vars is that they're more discoverable.

anakrish: I'd originally tried putting the logic for inferring HAS_QUOTE_PROVIDER in test-samples.cmake. That would've allowed the samples test to automatically select the appropriate list of samples to be tested. 
I gave up on that since `find_package` does not work in cmake scripts (the -P option).
I will try to find a work around.

Though an environment variable allows configuration at test execution time; it still has to be set up correctly which has to be done either by the user (which implies more responsibility on the user for testing attestaton samples) or by the cmake system (which implies a coupling between the build machine and the samples list).

CodeMonkeyLeet: Yeah, I think it's useful to consider the scenarios here:
1. **SDK user.** This CMakeLists.txt doesn't affect them, they always have to figure out their configuration and whether to build/run individual samples or not.
2. **OE SDK committer.** This PR preserves convenience for them where we consider the building and testing on the same device configuration as the primary use case.
   - I think I'm also much more willing to assume a default config in these scenarios, as we do with `OE_SIMULATION`. e.g we could define the environment var as the inverse `NO_DCAP_PROVIDER` so it doesn't have to explicitly set if we expect most committers to build and test with the provider. It's also a set-once property, which isn't significantly onerous in my mind and actually easier to toggle for experimental purposes than installing/uninstalling a DCAP provider lib.
   - I'll also note that the `find_package` check only indicates that a DCAP lib is installed, not that it is either compatible with OE SDK or will run, it's always the user's responsibility to understand that anyway. It's a reasonable assumption though that if you installed it, you've set up the system that it will run. 
      - A different approach may be for CMake to infer this and set the environmental variable on your behalf, if it hasn't been previously set, and to warn if it is set to a value counter to what cmake detects.  
3. **CI/CD**. This is more likely to use the build and test on different environments, and for that use case, I disagree with @achamayou's position that rerunning cmake is fine; that would require cloning both source and output trees to the test device which is largely unnecessary otherwise.

CHANGELOG.md
anakrish: Currently this flag is exposed to the SDK contributor and documented in all the SGX1 (non FLC) docs
E.g: https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/SGX1GettingStarted.md

Hence it seems appropriate to deprecate it.

CMakeLists.txt
anakrish: Yes, I noticed this too. I'm just making it clear that it is being deprecated, since it is mentioned in contributor documents for SGX1 (non FLC).

anakrish:
@CodeMonkeyLeet @achamayou @dthaler @mingweishih @radhikaj 

A quick poll.
`HAS_QUOTE_PROVIDER` is a cmake option that is currently documented in SGX1 workflow documents. This cmake option, however, is only used by SDK contributors rather than users.

Should we go ahead are remove it or should we deprecate it first and then remove it?

samples/CMakeLists.txt
anakrish: @CodeMonkeyLeet 
Can you look at the latest implementation?
- It uses cmake itself to determine whether the target system has DCAP libraries.
  As was the case with HAS_QUOTE_PROVIDER=on/off, no checks are made as to whether the DCAP libraries is compatible with the SDK or not.
- The build can be transferred to another system and the DCAP determination will be done on that system.

anakrish:
bors try

CodeMonkeyLeet:
> @CodeMonkeyLeet @achamayou @dthaler @mingweishih @radhikaj
> 
> A quick poll.
> `HAS_QUOTE_PROVIDER` is a cmake option that is currently documented in SGX1 workflow documents. This cmake option, however, is only used by SDK contributors rather than users.
> 
> Should we go ahead are remove it or should we deprecate it first and then remove it?

My preference is to just remove it. The basic expectation has been that a committer on master has always had to keep up with potentially breaking changes in the build/prerequisites configuration for building the project. 

Regardless of the outcome of this discussion, this PR should remove all existing examples specifying the `HAS_QUOTE_PROVIDER` option on the cmake commands; i.e. we shouldn't tell instruct users to use deprecated features anyway.

samples/CMakeLists.txt
anakrish: >     3. **CI/CD**. This is more likely to use the build and test on different environments, and for that use case, I disagree with @achamayou's position that rerunning cmake is fine; that would require cloning both source and output trees to the test device which is largely unnecessary otherwise.

cmake creates dependency rules between build artifacts and the sources for which the artifact is generated. This allows it to trigger the rebuild of an artifact if the source has been deemed to have changed.

What this implies is that we'd have to copy both the source and output trees from the build system on to the test system.
Additionally, the time stamps as well as the full file paths have to be preserved in order to be able to ctest on a different machine.

anakrish:
>
> Regardless of the outcome of this discussion, this PR should remove all existing examples specifying the `HAS_QUOTE_PROVIDER` option on the cmake commands; i.e. we shouldn't tell instruct users to use deprecated features anyway.

I was originally planning to update the docs in next PR; just to see if there was any opportunity to rewrite the docs. 
I will instead just remove HAS_QUOTE_PROVIDER from docs and update the PR.

anakrish:
> > Regardless of the outcome of this discussion, this PR should remove all existing examples specifying the `HAS_QUOTE_PROVIDER` option on the cmake commands; i.e. we shouldn't tell instruct users to use deprecated features anyway.
> 
> I was originally planning to update the docs in next PR; just to see if there was any opportunity to rewrite the docs.
> I will instead just remove HAS_QUOTE_PROVIDER from docs and update the PR.

Done.

CHANGELOG.md
anakrish: Fixed typo.

anakrish:
bors try

anakrish:
bors try

samples/test-samples.cmake
```diff
@@ -2,10 +2,41 @@
...
-# PREFIX_DIR to be defined:
+# COMPILER_SUPPORTS_SNMALLOC, USE_DEBUG_MALLOC and PREFIX_DIR to be defined:
 #
-#     cmake -DHAS_QUOTE_PROVIDER=ON -DSOURCE_DIR=~/openenclave -DBUILD_DIR=~/openenclave/build -DPREFIX_DIR=/opt/openenclave -P ~/openenclave/samples/test-samples.cmake
+#     cmake -DSOURCE_DIR=~/openenclave \
+#           -DBUILD_DIR=~/openenclave/build
+#           -DPREFIX_DIR=/opt/openenclave \
+#           -P ~/openenclave/samples/test-samples.cmake
+
+#Determine whether DCAP quote provider libraries are available in the system.
```
mingweishih, line 12:
```suggestion
#Determine whether DCAP quote provider libraries are available in the system.
# Determine whether DCAP quote provider libraries are available in the system.
```

samples/check_quote_provider.cmake
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+cmake_minimum_required(VERSION 3.13)
+project("Check Quote Provider")
+
+find_library(SGX_DCAP_QL NAMES sgx_dcap_ql REQUIRED)
+
+#Raise fatal error if sgx_dcap_ql library is not found.
```
CodeMonkeyLeet, line 8:
```suggestion
# Raise fatal error if sgx_dcap_ql library is not found.
```

CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [Unreleased][Unreleased_log]
 --------------
 
+### Added
+- OE SDK Release packages can now be built on non-SGX and non-FLC machines.
+
+### Changed
+
+### Removed
```
CodeMonkeyLeet, line 18: One additional thing: this probably doesn't need to be in the change log, I think the `Added` information is the important one. I'll leave it up to you.

anakrish:
> > @CodeMonkeyLeet @achamayou @dthaler @mingweishih @radhikaj
> > A quick poll.
> > `HAS_QUOTE_PROVIDER` is a cmake option that is currently documented in SGX1 workflow documents. This cmake option, however, is only used by SDK contributors rather than users.
> > Should we go ahead are remove it or should we deprecate it first and then remove it?
> 
> My preference is to just remove it. The basic expectation has been that a committer on master has always had to keep up with potentially breaking changes in the build/prerequisites configuration for building the project.
> 
> Regardless of the outcome of this discussion, this PR should remove all existing examples specifying the `HAS_QUOTE_PROVIDER` option on the cmake commands; i.e. we shouldn't tell instruct users to use deprecated features anyway.

I have gone ahead and removed it. Based on discussions in this PR, only 3 folks cared enough about it (@CodeMonkeyLeet - remove, @achamayou -remove, and @anakrish - deprecate) and `remove` was the majority opinion.

CHANGELOG.md
anakrish: Removed this note from the PR.

CHANGELOG.md
anakrish: Removed HAS_QUOTE_PROVIDER instead of deprecated based on discussion in this PR.

anakrish:
> Do we have plan for those tests have different code paths depending on the quote provider?

@mingweishih I have added a table to the PR description. The table lists the tests that relied on HAS_QUOTE_PROVIDER.
No testing coverage is lost. Can you take a look?

mingweishih:
> > Do we have plan for those tests have different code paths depending on the quote provider?
> 
> @mingweishih I have added a table to the PR description. The table lists the tests that relied on HAS_QUOTE_PROVIDER.
> No testing coverage is lost. Can you take a look?

Thanks for the great summary! The table looks good.

PR #3291, opened 2020-07-21
Reduce max threads to 8 for pluggable_allocator sample.

anakrish:
This value allows the sample to pass on all known ACC VM configurations.


samples/pluggable_allocator/README.md
```diff
@@ -145,36 +145,36 @@ Configuration:
         num-allocations (per-thread) = 100000
                  max-allocation-size = 16384 bytes
 num-threads = 1:
-      dlmalloc   (default allocator) =    9 milliseconds
-    oesnmalloc (pluggable allocator) =    6 milliseconds
+      dlmalloc   (default allocator) =   12 milliseconds
```
radhikaj, line 148: Maybe we need not have all this output(can be removed in another PR) since these numbers will change depending on the number of cores.

PR #3292, opened 2020-07-21
Reduce maximum threads to 8 for snmalloc sample.

anakrish:
This ensures that the sample works on all known ACC VM configurations.

The total heap requirement for the sample is estimated to be:
   max-num-threads * (snmalloc cost per thread + max-num-objects-per-thread * max-object-size)
   = 16 * (256 KB + 16 * 16KB)
   = 16 * 512KB
   = 16 * 512KB / 4096 pages
   = 2048 pages

Eventhough the memory (heap) requirement is 2048 pages, snmalloc raises
"Critical error: Out-of-memory during initialisation." even when the number of heap pages
are set to 4096 ~or 8192~.

This PR worksaround the above behavior.

### DIFF
--- a/samples/pluggable_allocator/README.md
+++ b/samples/pluggable_allocator/README.md
@@ -54,11 +54,11 @@ Plugging in a custom allocator is a straight-forward process that involves two s
     ```
     # snmalloc requires at least 256 KB per enclave thread.
     # Given 16 enclave threads (NumTCS), this implies
-    #    minimum heap size = (256 * 1024 * 16) / 4096 = 1044 pages.
-    # The heap size (8192 pages) is well above the minimum requirement,
+    #    minimum heap size = (256 * 1024 * 8) / 4096 = 512 pages.
+    # The heap size (4096 pages) is well above the minimum requirement,
     # and accounts for the large number of allocations performed by
     # each enclave thread in the sample.
-    NumHeapPages=8192
+    NumHeapPages=4096
     NumTCS=16
     ```
 
@@ -116,7 +116,7 @@ The elapsed times give an indication of how much speed up the custom allocator p
 the default allocator.
 
 The host then repeats the benchmarks on the two enclaves, but increases the number of threads
-each time, until the maximum number of threads (15) is reached.
+each time, until the maximum number of threads (8) is reached.
 
 It can be observed that the custom allocator scales nicely with multiple threads whereas the
 default allocator does not.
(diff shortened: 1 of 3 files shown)


anakrish:
bors try

anakrish:
@achamayou It seems that we can run into out of memory errors even with number of pages well above the expected memory requirement. Will create an issue and reproduction steps.

achamayou:
@anakrish thanks for the heads up, adding @mjp41. One thing that's worth noting is that the initial amount is 256Kb (once, for metadata), plus 256Kb per thread: https://github.com/openenclave/openenclave/blob/ebbf14511d5761cdaa539ec0091a70369d0a9f05/include/openenclave/internal/tests.h#L69

Another thing is that max-num-objects-per-thread * max-object-size may not be an accurate maximum value depending on the relationship between allocated sizes and size classes.

mjp41:
@anakrish So the allocation between 8KiB and 256KiB are consider "medium" allocations, and will result in a chunk per size class in this range.  I.e.  for your example
```
  10240, 12288, 14338, 16384
```
means it is likely that each allocation will be on its own slab, so they will use 256KiB each.  Given you are uniformly distributing based on size, on average half the allocation will be in this range, so it is highly likely all four size classes will be occupied. So the worst case is 
```
   5 * 256KiB
```
per thread.  If you drop the maximum allocation in the test to 8KiB.  It would run in much smaller memory. I can adjust this boundaries, this is just let to good performance in various benchmarks.

anakrish:
@achamayou @mjp41 
Thanks for the explanation regarding memory consumption. It makes sense.
Can you provide a small high-level write up of how to estimate the number of heap pages? That would be quite useful for the users in v0.11.

We'd encountered this late yesterday and as a quick lastt-minute fix for v0.10 reduced the thread count to 8 in the v0.10.x branch.
https://github.com/openenclave/openenclave/pull/3291
For consistency purposes, it is better to reduce the num threads to 8 in master as well.

mjp41:
@anakrish happy to write something up on this.  My main concern is the precise amount of memory is dependent on implementation details, so I should perhaps add it to snmalloc repo, and you can reference it?

anakrish:
> @anakrish happy to write something up on this. My main concern is the precise amount of memory is dependent on implementation details, so I should perhaps add it to snmalloc repo, and you can reference it?

I think referencing snmalloc repo works best. It would be better if the document was specifically tailored towards use within enclaves/constrained memory situations.

>> My main concern is the precise amount of memory is dependent on implementation details
Yes we understand this. Anything that would help the user experience is welcome. For example I wasn't aware of the consequence of 8-16KB allocations. I'd chosen it since certificate chains, quotes and other attestation collaters etc could be around 10KB.

mjp41:
@anakrish or @achamayou could you write up an issue on the snmalloc repo, and I will address when I get back to work.

PR #3293, opened 2020-07-21
Use enclave_* cmake macros for builds with EEID.

wintersteiger:
Fixes the build when both EEID and LVI mitigation are enabled. Fixes build failure in #3152.


wintersteiger:
bors try

anakrish:
bors try

PR #3294, opened 2020-07-21
Support request for multiple SGX quote collateral providers.

qiucwang:
In current implementation, [common/sgx/collateral.c](https://github.com/openenclave/openenclave/blob/master/common/sgx/collateral.c#L185-L212) calls into host to fetch collateral given the PCK certificate. There was only one provider(certificate issuer) "Intel SGX PCK Processor CA" so [host/sgx/sgxquoteprovider.c](https://github.com/openenclave/openenclave/blob/master/host/sgx/sgxquoteprovider.c#L16) hardcoded CA as "processor".
Now with the introducing of new platform like icelake, there could be multiple collateral providers. e.g., the provider name for icelake machine should be "platform". This PR is to support multiple collateral providers request.

Collateral is fetched by `fmpsc` extracted from PCK certificate. But as disscussed above, now we also need to know which CA issued that certificate (PCK Processor CA or PCK Platform CA). That's not provided in PCK certificate. But according to the [spec](https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_PCK_Certificate_CRL_Spec-1.4.pdf), we can tell it by "Platform Instance ID" field in SGX extension. It's a 16 bytes id that certificates issued by PCK Platform CA must contain this
field but certificates issued by PCK Processor CA do not contain this field. So that the `opt_platform_instance_id[16]` will be [an array of 0](https://github.com/openenclave/openenclave/blob/master/common/sgx/collateral.c#L193) when certificate is provided by PCK Processor CA, and a 16 bytes id when certificate is provided by PCK Platform CA. So we can use this information to detemine the collateral provider.

A new field `collateral_provider` is added in `oe_get_sgx_quote_verification_collateral_args_t` to indicate the collateral provider, which now is either CRL_CA_PROCESSOR(1) or CRL_CA_PLATFORM(2).

Tests have been run and passed in both coffeelake machine and icelake machine.

### DIFF
--- a/common/sgx/collateral.c
+++ b/common/sgx/collateral.c
@@ -43,6 +43,22 @@ done:
     return result;
 }
 
+// determine CA provider type by platform_instance_id from SGX PCK certificate
+// which is 16 bytes long
+static oe_result_t _get_crl_ca_type(
+    uint8_t* platform_instance_id,
+    oe_get_sgx_quote_verification_collateral_args_t* args)
+{
+    uint8_t null_platform_id[16] = {0};
+    if (memcmp(
+            platform_instance_id, null_platform_id, sizeof(null_platform_id)) ==
+        0)
+        args->collateral_provider = CRL_CA_PROCESSOR;
+    else
+        args->collateral_provider = CRL_CA_PLATFORM;
+    return OE_OK;
+}
+
 static oe_result_t _get_tcb_info_validity(
     const oe_parsed_tcb_info_t* parsed_tcb_info,
     oe_datetime_t* from,
@@ -203,6 +219,9 @@ oe_result_t oe_get_sgx_quote_verification_collateral_from_certs(
         parsed_extension_info.fmspc,
         sizeof(parsed_extension_info.fmspc)));
 
+    // Use platform instance id to determine the collateral provider (PCK CA)
+    OE_CHECK(
+        _get_crl_ca_type(parsed_extension_info.opt_platform_instance_id, args));
     OE_CHECK(oe_get_sgx_quote_verification_collateral(args));
 
     result = OE_OK;
--- a/enclave/sgx/collateralinfo.c
+++ b/enclave/sgx/collateralinfo.c
@@ -20,6 +20,7 @@
 oe_result_t _oe_get_quote_verification_collateral_ocall(
     oe_result_t* _retval,
     uint8_t fmspc[6],
+    uint8_t collateral_provider,
     void* tcb_info,
     size_t tcb_info_size,
     size_t* tcb_info_size_out,
@@ -52,6 +53,7 @@ oe_result_t _oe_get_quote_verification_collateral_ocall(
 oe_result_t _oe_get_quote_verification_collateral_ocall(
     oe_result_t* _retval,
     uint8_t fmspc[6],
+    uint8_t collateral_provider,
     void* tcb_info,
     size_t tcb_info_size,
     size_t* tcb_info_size_out,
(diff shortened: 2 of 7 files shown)


mingweishih:
Can you rename the title to be more specific?

qiucwang:
> Can you rename the title to be more specific?

Oh sorry for that. Fixed.

host/sgx/sgxquoteprovider.c
```diff
@@ -59,13 +56,20 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
...
     sgx_ql_qve_collateral_t* collateral = NULL;
+    char* ca_type = NULL;
     uint32_t host_buffer_size = 0;
     uint8_t* p = 0;
     uint8_t* p_end = 0;
     OE_TRACE_INFO("Calling %s\n", __FUNCTION__);
 
     uint8_t* fmspc = args->fmspc;
     uint16_t fmspc_size = sizeof(args->fmspc);
+    if (args->collateral_provider == CRL_CA_PROCESSOR)
```
yentsanglee, line 67: Add line separation between local variable declaration and code

enclave/sgx/collateralinfo.c
```diff
@@ -182,6 +186,10 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
 
     /* fmspc */
     memcpy(in.fmspc, args->fmspc, sizeof(in.fmspc));
+    /* collateral_provider */
+    in.collateral_provider = args->collateral_provider;
+    // memcpy(in.collateral_provider, args->collateral_provider,
```
jvega10, line 191: extra comments

include/openenclave/internal/report.h
```diff
@@ -40,6 +41,10 @@ typedef struct _oe_get_sgx_quote_verification_collateral_args
     uint8_t* host_out_buffer;             /* out */
 } oe_get_sgx_quote_verification_collateral_args_t;
 
+// Collateral provider for sgx quote verification
+#define CRL_CA_PROCESSOR (1)
```
yentsanglee, line 45: Suggest using enum

common/sgx/collateral.c
```diff
@@ -43,6 +43,20 @@ oe_result_t __oe_sgx_set_minimum_crl_tcb_issue_date(
     return result;
 }
 
+static oe_result_t _get_crl_ca_type(
+    uint8_t* platform_instance_id,
+    oe_get_sgx_quote_verification_collateral_args_t* args)
+{
+    uint8_t null_platform_id[16] = {0};
+    if (memcmp(
+            platform_instance_id, null_platform_id, sizeof(null_platform_id)) ==
```
achamayou, line 52: It's not immediately obvious locally that platform_instance_id is at least 16 bytes long.

enclave/sgx/collateralinfo.c
WangQiuc: Thanks

qiucwang: Thanks!

common/sgx/collateral.c
qiucwang: Thanks. I'll add a comment that platform_instance_id here is one of SGX PCK certificate extension fields which is 16 bytes long

include/openenclave/internal/report.h
qiucwang: If we use enum, we have to define a new struct type here. But the parameter marshalling makes it kind of tricky.
I've added a check to ensure the value should be either 1 or 2.

qiucwang:
bors try

yentsanglee:
Please squash all commits.

qiucwang:
bors try

qiucwang:
bors try

PR #3295, opened 2020-07-21
Fix memory leak in the host openssl wrapper

mingweishih:
This PR fixes couple memory leaks in the host-side openssl wrapper.

### DIFF
--- a/host/crypto/openssl/cert.c
+++ b/host/crypto/openssl/cert.c
@@ -669,6 +669,8 @@ oe_result_t oe_cert_chain_read_pem(
     result = OE_OK;
 
 done:
+    if (result != OE_OK && sk)
+        sk_X509_pop_free(sk, X509_free);
 
     if (tmp_pem_data && (tmp_pem_data != pem_data))
         oe_free(tmp_pem_data);
@@ -742,7 +744,6 @@ oe_result_t oe_cert_get_rsa_public_key(
     oe_result_t result = OE_UNEXPECTED;
     const cert_t* impl = (const cert_t*)cert;
     EVP_PKEY* pkey = NULL;
-    RSA* rsa = NULL;
 
     /* Clear public key for all error pathways */
     if (public_key)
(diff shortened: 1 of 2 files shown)


mingweishih:
bors try

host/crypto/openssl/cert.c
```diff
@@ -774,6 +776,9 @@ oe_result_t oe_cert_get_rsa_public_key(
         EVP_PKEY_free(pkey);
     }
 
+    if (rsa)
```
CodeMonkeyLeet, line 779: Is this even necessary? We don't actually use `rsa` anywhere and it looks like the variable can be eliminated in this function entirely.

mingweishih: Good point, let me try just remove it.

host/crypto/openssl/cert.c
```diff
@@ -669,6 +669,8 @@ oe_result_t oe_cert_chain_read_pem(
     result = OE_OK;
 
 done:
+    if (result != OE_OK && sk)
+        sk_X509_pop_free(sk, X509_free);
```
anakrish, line 673: Is there a reason why this leak had gone undetected? I'd suggest adding a test that would lock down the scenario.

mingweishih: We didn't enable debugmalloc or similar features on the host. The memory leaks in the PR are found because I was porting the openssl wrapper to the enclave where debugmalloc is enabled.

mingweishih:
bors try

host/crypto/openssl/cert.c
anakrish: Ah Ok. Thanks for the clarification. 
Do you know if it is feasible to enable valgrind or something else to detect host side memory issues?

mingweishih: I think either valgrind or clang ([LeakSanitizer](https://clang.llvm.org/docs/LeakSanitizer.html)) should just work.

anakrish: > I think either valgrind or clang ([LeakSanitizer](https://clang.llvm.org/docs/LeakSanitizer.html)) should just work.

Created issue #3298

PR #3297, opened 2020-07-21
Make install_lvi_mitigation_bindir exit on errors

eddyashton:
`install_lvi_mitigation_bindir` will print error messages but continue to execute, and doesn't set any exit codes to indicate errors. As such its awkward to use in any automation, or call from any script. This PR sets `-o errexit` (this seems to be the preferred style over `set -e`?) so errors are more easily detected.

Signed-off-by: Eddy Ashton <edashton@microsoft.com>


achamayou:
bors try

mingweishih:
> bors try

This script is not tested in our daily/nightly build.  It's only used/tested in our package release process.

PR #3301, opened 2020-07-22
Fix host-verify sample make run failure

yentsanglee:
Signed-off-by: Yen Lee <yenlee@microsoft.com>

Fixed host-verify sample 'make run' failure as described in issue 3300.


BRMcLaren:
Would it be possible to add a test along this fix to prevent a regression?

PR #3308, opened 2020-07-23
Pass segfault address from host to OE exception handler

prp:
This commit takes the address that caused a segfault, as reported in the
host signal handler, and passes it to the in-enclave OE exception
handler. This is needed so that an application signal handler can
obtain the faulting address inside the enclave.

The current implementation should be generalised to handle arbitrary
synchronous signals correctly. Note that SGX2 permits the exception
information to be obtained in-enclave, making this untrusted mechanism
unnecessary.

This is needed by PR https://github.com/lsds/sgx-lkl/pull/646 to fix https://github.com/lsds/sgx-lkl/issues/645.

### DIFF
--- a/host/sgx/linux/exception.c
+++ b/host/sgx/linux/exception.c
@@ -36,6 +36,7 @@ static void _host_signal_handler(
     oe_host_exception_context_t host_context = {0};
     host_context.rax = (uint64_t)context->uc_mcontext.gregs[REG_RAX];
     host_context.rbx = (uint64_t)context->uc_mcontext.gregs[REG_RBX];
+    host_context.address = (uint64_t) sig_info->si_addr;
     host_context.rip = (uint64_t)context->uc_mcontext.gregs[REG_RIP];
 
     // Call platform neutral handler.
--- a/enclave/core/sgx/exception.c
+++ b/enclave/core/sgx/exception.c
@@ -317,7 +317,7 @@ void oe_real_exception_dispatcher(oe_context_t* oe_context)
     oe_exception_record_t oe_exception_record = {0};
     oe_exception_record.code = td->exception_code;
     oe_exception_record.flags = td->exception_flags;
-    oe_exception_record.address = td->exception_address;
+    oe_exception_record.address = td->exception_host_address;
     oe_exception_record.context = oe_context;
 
     // Refer to oe_enter in host/sgx/enter.c. The contract we defined for EENTER
@@ -395,9 +395,11 @@ void oe_virtual_exception_dispatcher(
         {
             td->exception_address = ssa_gpr->rip;
 
-            /* TODO PRP: We need to read the exception information from
+            /* TODO We need to read the exception information from
              * oe_exception_record instead */
             td->exception_code = OE_EXCEPTION_PAGE_FAULT;
+            // Store the accessed address that caused the page fault from host
+            td->exception_host_address = oe_exception_record->address;
             td->exception_flags |= OE_EXCEPTION_FLAGS_HARDWARE;
         }
         else
--- a/host/sgx/exception.c
+++ b/host/sgx/exception.c
@@ -21,6 +21,7 @@ uint64_t oe_host_handle_exception(oe_host_exception_context_t* context)
 {
     uint64_t exit_code = context->rax;
     uint64_t tcs_address = context->rbx;
+    uint64_t host_address = context->address;
     uint64_t exit_address = context->rip;
 
     // Check if the signal happens inside the enclave.
@@ -46,8 +47,9 @@ uint64_t oe_host_handle_exception(oe_host_exception_context_t* context)
 
         oe_exception_record_t oe_exception_record = {0};
 
-        /* TODO PRP: We need to save the exception information in
+        /* TODO: We need to save the complete exception information in
          * oe_exception_record */
+        oe_exception_record.address = host_address;
 
         // Call into enclave first pass exception handler.
         uint64_t arg_out = 0;
--- a/host/sgx/exception.h
+++ b/host/sgx/exception.h
@@ -11,6 +11,7 @@ typedef struct _host_exception_context
 {
     uint64_t rax;
     uint64_t rbx;
+    uint64_t address;
     uint64_t rip;
 } oe_host_exception_context_t;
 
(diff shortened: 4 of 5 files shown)


host/sgx/linux/exception.c
```diff
@@ -36,6 +36,7 @@ static void _host_signal_handler(
     oe_host_exception_context_t host_context = {0};
     host_context.rax = (uint64_t)context->uc_mcontext.gregs[REG_RAX];
     host_context.rbx = (uint64_t)context->uc_mcontext.gregs[REG_RBX];
+    host_context.address = (uint64_t) sig_info->si_addr;
```
anakrish, line 39: Won't sig_info->si_addr be the same as REG_RIP which would be the AEP address?

prp: That's what the code originally assumed but no. `REG_RIP` will be the address of the instruction that caused the SIGSEGV but `sig_info->si_addr` contains the address being accessed, which is what we need inside the enclave.

anakrish: Ineresting. I'm surprised that the address of enclave memory location is revealed to the host.

jhand2:
I'm also pretty concerned about the enclave making decisions around exception handling based on information from the host. This seems like a pretty significant attack vector (the original page fault handling concerns me as well). I'm worried this is (a) something we can't ever merge to master and (b) something sgx-lkl probably shouldn't be doing.

host/sgx/linux/exception.c
prp: @anakrish You are right, `si_addr` with a hardware enclave always happens to be 0x0. (What threw me off is that this value happens to be what the JVM expects in its signal handler, so it makes Java work with SGX-LKL. Since the SGX driver on the host is aware of enclave page fault addresses, it didn't strike me as entirely odd that this would be exposed.)

@anakrish @jhand2 To provide some context, this is to fix the Java support for SGX-LKL (https://github.com/lsds/sgx-lkl/issues/645). The corresponding PR for SGX-LKL is https://github.com/lsds/sgx-lkl/pull/646. The Hotspot JVM (and probably other language runtimes) use page faults to acquiesce threads and make them reach safepoints. Apparently the handler checks the faulting address and expects to find the value 0x0.

Does that mean that there is no way under SGX1 for the enclave to find out the actual faulting address? If so, I will document this as a fundamental compatibility limitation of SGX-LKL here: https://github.com/lsds/sgx-lkl/blob/oe_port/docs/Incompatibilities.md (I also presume that this problem will go away under SGX2 -- can you confirm this?)

I will then close this PR and update PR https://github.com/lsds/sgx-lkl/pull/646 to always return `0x0` as the faulting address in hw mode. (We should decide if we want the correct semantics in sw mode.)

anakrish: Thanks for the explanation.

> Does that mean that there is no way under SGX1 for the enclave to find out the actual faulting address?

Disassembling the instruction that resulted in SEGV and exception context could be used to find out the actual faulting address. However this may not be easy to implement.

 @bodzhang @MWShan Please respond to @prp 's questions above regarding faulting address in SGX1 and SGX2.

jhand2: @prp My main concern is that if we expose this behavior to user applications it is not immediately clear that the exception record information should be considered untrusted. From the use-case you provided, I don't see any obvious security issues, but in order to expose it to the JVM, it would need to be exposed to any SGX-LKL app (via the POSIX signal handler), right?

prp: @jhand2 Right, for SGX1, there seems to be no trusted way to expose a page fault to the enclave. We could have an attested setting in the enclave_config called "permit_untrusted_host_traps" (with a default setting of false). This would permit particular applications (such as the JVM) to receive signals, but they would have to be aware of the security implications (which we already mention here: https://github.com/lsds/sgx-lkl/blob/oe_port/docs/Incompatibilities.md#summary-of-security-implications).

I believe that for SGX2, none of this will be necessary. For SGX1, it would be nice if the signal support was complete, but it seems that there is no practical way for the in-enclave signal handler to know the faulting address. It would be good to get confirmation on this.

jhand2: @prp An attested config setting seems like a good option. I think it would be good to submit this (and the other SIGSEGV commit) as a PR to master to make sure we design it in a way that is acceptable to other maintainers.

MWShan: In SGX1, there is not a direct way to determine the faulting address.  In addition, SGX1 will not put the #PF Exception Vector into the SSA.EXITINFO.VECTOR field, so the enclave cannot be certain that a #PF occurred.  With this feature, the untrusted code could potentially interrupt the enclave and then tell it that a #PF occurred at a specific address.

prp: > In SGX1, there is not a direct way to determine the faulting address. In addition, SGX1 will not put the #PF Exception Vector into the SSA.EXITINFO.VECTOR field, so the enclave cannot be certain that a #PF occurred. With this feature, the untrusted code could potentially interrupt the enclave and then tell it that a #PF occurred at a specific address.

Thanks @MWShan. Do you see an indirect way of obtaining the faulting address with SGX1? (Obviously with the security implications that we've discussed.)

MWShan: @prp - I cannot see a good indirect method (@anakrish mentioned disassembling the opcode, but seems very complex and open to errors).   You mentioned that the Hotspot JVM used a #PF to acquiesce threads - perhaps there is another method or fault that could be used.

prp: > @prp - I cannot see a good indirect method (@anakrish mentioned disassembling the opcode, but seems very complex and open to errors). You mentioned that the Hotspot JVM used a #PF to acquiesce threads - perhaps there is another method or fault that could be used.

@MWShan I don't think that there is an option in OpenJDK 8 to have the HotSpot JVM acquiesce threads through a different mechanism. As a workaround for SGX1, I will add a new attested option ("unsafe_host_signals"), which always returns the faulting address as zero. This will be enough to unblock Java.

prp:
Since this design does not work, I am closing the PR.

host/sgx/linux/exception.c
MWShan: @prp - in an earlier post, I stated that SGX1 did not support #PF being reported in EXITINFO.  A colleague has corrected me, #PF and #GP can be provided in EXITINFO if MISCSELECT supports them.  The allowed values of MISCSELECT (configured in the SECS) are enumerated via CPUID.(EAX=12H, ECX=0):EBX[31:0] (as stated in the SDM section 37.9.2).  

This fact may not be much help since I have not seen any SGX1 systems support the #PF/#GP in EXITINFO; though current SGX2 systems do have this feature.

PR #3320, opened 2020-07-27
Attestation format ID reduction/deprecation plan

yentsanglee:
Proposal of plan to:
1. Consolidate SGX_ECDSA_P256 format IDs.
2. Deprecate oe_report definitions.


docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -685,6 +685,109 @@ since legacy applications don't call these API functions explicitly.
...
+1. Consolidate SGX_ECDSA_P256 format IDs.
+2. Deprecate oe_report definitions.
+
+### Motivation
+The 2 extra SGX P256 evidence formats defined in V0.10 implementation can be reduced/deprecated to make the API surface cleaner:
+- `OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT`
+- `OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE`
+
+For every listed format, `oe_get_evidence()` can wrap with
+an attestation header when the `OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID` bit in its `flags` parameter is set. These are all possible SGX_ECDSA_P256 format structures that can be generated by oe_get_evidence():
```
qiucwang, line 700:
```suggestion
an attestation header when the `OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID` bit in its `flags` parameter is set. These are all possible SGX_ECDSA_P256 format structures that can be generated by `oe_get_evidence()`:
```

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -685,6 +685,109 @@ since legacy applications don't call these API functions explicitly.
...
+### Plan
+
+#### Phase 1 - V0.11
+- Deprecate `OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE` as proposed above.
+
+#### Phase 2 - V0.11
+- Remove `oe_report_header` from format 2. Increment `oe_attestation_header.version` to flag the change.
+- Modify `oe_verify_evidence()`, `oe_verify_report()`, and `oe_verify_attestation_certificate()` so they handle new format version.
+- Deprecate `oe_verify_attestation_certificate_with_evidence`.
+#### Phase 3 - V0.11
```
qiucwang, line 784:
```suggestion

#### Phase 3 - V0.11
```

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -685,6 +685,109 @@ since legacy applications don't call these API functions explicitly.
...
+
+### Plan
+
+#### Phase 1 - V0.11
+- Deprecate `OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE` as proposed above.
+
+#### Phase 2 - V0.11
+- Remove `oe_report_header` from format 2. Increment `oe_attestation_header.version` to flag the change.
+- Modify `oe_verify_evidence()`, `oe_verify_report()`, and `oe_verify_attestation_certificate()` so they handle new format version.
+- Deprecate `oe_verify_attestation_certificate_with_evidence`.
```
qiucwang, line 783:
```suggestion
- Deprecate `oe_verify_attestation_certificate_with_evidence`.

```

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -685,6 +685,109 @@ since legacy applications don't call these API functions explicitly.
...
+
+### Motivation
+The 2 extra SGX P256 evidence formats defined in V0.10 implementation can be reduced/deprecated to make the API surface cleaner:
+- `OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT`
+- `OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE`
+
+For every listed format, `oe_get_evidence()` can wrap with
+an attestation header when the `OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID` bit in its `flags` parameter is set. These are all possible SGX_ECDSA_P256 format structures that can be generated by oe_get_evidence():
+
+| Format ID | Evidence structure  |
```
shnwc, line 702: This table can't be displayed properly in github. For column divider, `\|\|` is needed instead of `||`.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -685,6 +685,109 @@ since legacy applications don't call these API functions explicitly.
...
+    custom_claims,
+    size_t custom_claims_size,  // Must be <= 64 bytes
+    ...);
+```
+
+- To generate format 5:
+```
+oe_get_evidence(
+    OE_FORMAT_UUID_SGX_ECDSA_P256,
+    0,
```
shnwc, line 747: With this proposal, the OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID flag is effectively extended / redefined so that the OE SDK framework not just wraps the plugin evidence with a header containing format ID, but also hashes the custom claims and passes the hash to the plugin. Missing of this flag requires a plugin to embed the custom claims in plugin-specific way (e.g. embedding it in  the 64-byte SGX report data field).

There has been discussions in SIG-attestation about some TEEs not having the custom claims size limit in the SGX report data field (64 bytes). For those plugins, forcing them to take a hash may not be the optimal solution.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -685,6 +685,109 @@ since legacy applications don't call these API functions explicitly.
...
+#### Phase 2 - V0.11
+- Remove `oe_report_header` from format 2. Increment `oe_attestation_header.version` to flag the change.
+- Modify `oe_verify_evidence()`, `oe_verify_report()`, and `oe_verify_attestation_certificate()` so they handle new format version.
+- Deprecate `oe_verify_attestation_certificate_with_evidence`.
+#### Phase 3 - V0.11
+- Deprecate `oe_get_report`.
+
+#### Phase 4 - Future
+- Deprecate OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT. Define it as OE_FORMAT_UUID_SGX_ECDSA_P256.
+
```
shnwc, line 789: How about adding a table with the eventual resultant SGX ECDSA format IDs / flags / evidence structures? How will all Larry's requested usages still be supported?

docs/DesignDocs/Attestation_API_Proposal.md
yentsanglee: I'd go for consistency. If OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID is set, custom claims is always appended to the evidence and report_data field contains hash value. This simplifies the code.

In the future when report_data field allows more than typical custom_claims size needed, we can make adjustment. From what I've read, 64 bytes report_data seems common.

docs/DesignDocs/Attestation_API_Proposal.md
yentsanglee: Until we are sure all have opt-out'ed of oe_report, phase 4 may never happen.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -685,6 +685,109 @@ since legacy applications don't call these API functions explicitly.
...
 # Authors
 
 - Shanwei Cen (@shnwc)
+- Yen Lee (yenlee@microsoft.com)
```
dthaler, line 793: Seems inconsistent for one author to be listed by github id and another author to be listed by email address.  Can we be consistent, one way or the other?

docs/DesignDocs/Attestation_API_Proposal.md
dthaler: We could define another flag to control the hashing behavior, but if we did that it should be done now instead of later.   My hope though is that TEEs other than SGX would never want to use EMBED_FORMAT_ID anyway and would instead use a standard format (for which EMBED_FORMAT_ID would break interop), so I'm hoping that over time EMBED_FORMAT_ID becomes a legacy thing.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Agreed. My preference is github ID, which is public and will not change. Email addresses can be private and can change.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: A new flag sounds like a good option. On the other hand, all flags are effectively part of the format ID, and will need to be communicated / agreed upon between attesters and verifiers, so the verifiers know how to process. So it's preferable to minimize or eliminate flags.

In the current implementation, `oe_verify_evidence()` can output either claim `OE_CLAIM_SGX_REPORT_DATA`, or claim `OE_CLAIM_CUSTOM_CLAIMS`, or neither. In the proposal, what claims will be output for each format?

shnwc: An alternative to the proposal could be for `oe_get_evidence()` to support `OE_FORMAT_UUID_SGX_ECDSA_P256` and `OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE`, and for `oe_verify_evidence()` to support all 3 format IDs. There is no change in the current API definition and implmeentation, except for removing support of `OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT` in the SGX ECDSA attester plugin library.

`oe_verify_evidence() support of all 3 format IDs is needed for backward compatibility, to meet Larry’s needs.

Support of generation of SGX ECDSA evidence in format 5 is needed for backward compatibility, to help porting of existing SGX SDK solutions to OE SDK, since those applications already expects raw SGX quotes.

Support of generation of SGX ECDSA evidence in formats 3 and 5 was added to also allow testing of `oe_verify_evidence()` for those formats. But as implemented in `tests/attestation_plugin/plugin/tests.c` lines 786 – 843, for testing purpose, those evidence can be constructed from evidence in format 1 or 2. So support of formats 3 is no longer needed.

As the need for backward compatibility reduces, `oe_get_evicence() support of format 3 and 4, and `oe_verify_evidence()` support of formats 3 to 6 can be phased out when possible.

The API documentation can be made to state clearly that applications should use `oe_get_evidence()`  only with `OE_FORMAT_UUID_SGX_ECDSA_P256` with `OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID` unless there is a need for backward compatibility.

qiucwang: > A new flag sounds like a good option. On the other hand, all flags are effectively part of the format ID, and will need to be communicated / agreed upon between attesters and verifiers, so the verifiers know how to process. So it's preferable to minimize or eliminate flags.
> 
> In the current implementation, `oe_verify_evidence()` can output either claim `OE_CLAIM_SGX_REPORT_DATA`, or claim `OE_CLAIM_CUSTOM_CLAIMS`, or neither. In the proposal, what claims will be output for each format?

Format 2 `oe_verify_evidence(NULL)` will contain `OE_CLAIM_CUSTOM_CLAIMS`. It's the same as current implementation.
Format 3 `oe_verify_evidence(OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT)` and 5 `oe_verify_evidence(OE_FORMAT_UUID_SGX_ECDSA_P256)` will contain `OE_CLAIM_SGX_REPORT_DATA`. Format 5 is basically trying to replace current `oe_verify_evidence(OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE)`. And we hope to deprecate the format 3 eventually.

yentsanglee: The proposal is to keep the public API simple and remove redundancy. Adding another flag doubles the evidence formats. I agree with @shnwc that it's preferable to minimize or eliminate flags.

Suggest rename OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID to something that reflects the fact.

To make the responsibility clear, `OE_CLAIM_SGX_REPORT_DATA` is to be added to the claims by SGX plugins; `OE_CLAIM_CUSTOM_CLAIMS` is added by OE SDK. In MAA's scenario, they get report_data from `OE_CLAIM_SGX_REPORT_DATA`.

Support selective formats is an alternative but it does not present clean and concise API. Many people don't read document.

docs/DesignDocs/Attestation_API_Proposal.md
yentsanglee: Use github ID.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: > Support selective formats is an alternative but it does not present clean and concise API. Many people don't read document.

Personally I think the current API definition is clean. In addition to clarification in API documentation, we could put the definitions of format IDs etc. for backward compatibility to dedicated header file(s). Those applications that want to use the definitions for backward compatibility would need to explicitly include these header files.

shnwc: > To make the responsibility clear, OE_CLAIM_SGX_REPORT_DATA is to be added to the claims by SGX plugins; OE_CLAIM_CUSTOM_CLAIMS is added by OE SDK. In MAA's scenario, they get report_data from OE_CLAIM_SGX_REPORT_DATA

With this design, `oe_verify_evidence()` for some SGX evidence formats return both  `OE_CLAIM_SGX_REPORT_DATA` and `OE_CLAIM_CUSTOM_CLAIMS`.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -685,6 +685,130 @@ since legacy applications don't call these API functions explicitly.
...
+Here 'Remove' means the API definition is removed from public header files. Referring to the removed API definition results in compilation error.
+'Deprecate' means the API definition is redefined in public header files. Referring to the deprecated API definition does not break the build but the behavior is changed internally to guarantee continuous support.
+
+#### Phase 1 - V0.11
+- Remove `OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE` as proposed above.
+
+#### Phase 2 - V0.11
+- Remove `oe_report_header` from format 2. Increment `oe_attestation_header.version` to flag the change.
+- Modify `oe_verify_evidence()`, `oe_verify_report()`, and `oe_verify_attestation_certificate()` so they handle new format version.
+- Deprecate `oe_verify_attestation_certificate_with_evidence`.
```
CodeMonkeyLeet, line 791: Can you elaborate on this deprecation and eventual removal? Nothing else in this PR seems like it explains the reasoning for this or its eventual removal. Given that this API was just added, it seems weird that we're already deprecating it. Or did you mean the older `oe_verify_attestation_certificate` API?

Similarly, should this also be deprecating the `oe_generate_attestation_certficate` at the same time?

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -685,6 +685,130 @@ since legacy applications don't call these API functions explicitly.
...
+oe_verify_evidence(
+    OE_FORMAT_UUID_SGX_ECDSA_P256,
+    ...);
+```
+
+### Plan
+Reduction/deprecation is planned in phases.
+
+Here 'Remove' means the API definition is removed from public header files. Referring to the removed API definition results in compilation error.
+'Deprecate' means the API definition is redefined in public header files. Referring to the deprecated API definition does not break the build but the behavior is changed internally to guarantee continuous support.
```
CodeMonkeyLeet, line 783: Can you elaborate on what it means for the API to be redefined in the public header files? It's a little different from my standard understanding of what it means for an API to be deprecated, which is that:
1. The API is marked as deprecated. This means that documentation (e.g. doxygen, change logs) assert that the use of the API is not recommended, and tools may start to warn if it used.
2. The API signature remains unchanged and behaves as it did previously. (i.e. it is still a breaking change and not deprecation if an API alters existing behavior).

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -685,6 +685,130 @@ since legacy applications don't call these API functions explicitly.
...
+- Modify `oe_verify_evidence()`, `oe_verify_report()`, and `oe_verify_attestation_certificate()` so they handle new format version.
+- Deprecate `oe_verify_attestation_certificate_with_evidence`.
+
+At the end of phase 2, oe_get_evidence() and oe_verify_evidence() support these evidence formats:
+
+| Format ID | Evidence structure  |
+| -- | - |
+| A. `OE_FORMAT_UUID_SGX_ECDSA_P256` | `SGX_ECDSA_quote(custom_claims)` |
+| B. `OE_FORMAT_UUID_SGX_ECDSA_P256 + flag` | `oe_attestation_header \|\| SGX_ECDSA_quote(hash) \|\| custom_claims` |
+| C. `OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT` | `oe_report_header \|\| SGX_ECDSA_quote(custom_claims)` |
```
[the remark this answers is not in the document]
yentsanglee: The proposal would block format D. When such combination is passed in to oe_get_evidence(), it'd return invalid param error.

docs/DesignDocs/Attestation_API_Proposal.md
yentsanglee: oe_verify_attestation_certificate_with_evidence() was added to differentiate the x.509 extension being an oe_report or oe_evidence. Since the OIDs are different, we did not need this additional API at the first place.

docs/DesignDocs/Attestation_API_Proposal.md
yentsanglee: Examples of 'redefine'
#define oe_get_report(x,y,z) oe_get_evidence(SGX_UUID, x, y, z, NULL, ...)

Code that call oe_get_report() still compiles but it's replaced by new API. Call stack is different but the outcome remain the same.

docs/DesignDocs/Attestation_API_Proposal.md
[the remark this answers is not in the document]
CodeMonkeyLeet: I think this pushes me further in the direction that `OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID` should not be redefined as proposed. The point of the original design is that it is entirely orthogonal to the definition of evidence formats and the plugin implementations. The proposal makes the flag effectively an extension of format-specific behaviors, in which case the presence/absence of an embedded attestation header should have just been defined as part of specific formats and not coupled into the general OE SDK API surface.

docs/DesignDocs/Attestation_API_Proposal.md
CodeMonkeyLeet: Thanks for the explanation, could you add that to the document for clarity?

Also, is there a plan for the existing versions of those functions that don't call into `oe_get_evidence`/`oe_verify_evidence` in the deprecation plan?

docs/DesignDocs/Attestation_API_Proposal.md
CodeMonkeyLeet: I think that description would be useful to include in the line items that talk about the deprecate `oe_get_report` and `oe_verify_report` functions, rather than have it be indirectly referenced in an atypical definition of "deprecate" (e.g. I imagine that it doesn't apply in the same way to deprecation of `OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT`)

yentsanglee:
Thanks for comments. Most of the reviewers prefer having 3 UUIDs for SGX_ECDSA_P256. Based on the reviewers comments, investigation, and PRO/CON analysis, I decide to stay with current design.

Issue #3338 is opened to track oe_verify_attestation_certificate_with_evidence() deprecation.

PR #3326, opened 2020-07-28
Update to lvi mitigation toolchain to 2.10 and fix a bug in the installation script

### LINKED ISSUE #3307: PSW version used by LVI
The `install_lvi_mitigation_bindir` script [installs version 2.9](https://github.com/openenclave/openenclave/blob/4bf2dca44b92a962c932f9277d6e488acdd13592/scripts/lvi-mitigation/install_lvi_mitigation_bindir#L58) of the Intel LVI mitigation tools. As far as I can tell, these tool versions are the same as the Intel PSW/SDK versions (they are downloaded from the same [versioned sources](https://download.01.org/intel-sgx/sgx-linux/)). The 0.10 release of Open Enclave claims it is tested against PSW v2.10, but it appears the LVI-mitigated binaries in the release will still have been built with PSW v2.9 LVI mitigation tools.

Is this the case?
Do you know what the implications are? Are there likely to be any version incompatibilities if we use v2.10 for everything else, but v2.9 of the LVI mitigation tools?
Can we ensure this script remains up-to-date in future?

I notice that the `as.ld.objdump` file has increased in size from [3.8MiB](https://download.01.org/intel-sgx/sgx-linux/2.9/) to [35.6Mib](https://download.01.org/intel-sgx/sgx-linux/2.10/), suggesting the mitigation logic is significantly more complex in v2.10, so OE is potentially missing important mitigations? Of course this is just conjecture since I don't know what's in these customised `as`/`ld` binaries - perhaps someone from Intel can shed more light on this?

mingweishih:
Update the toolchain of lvi mitigation to 2.10 that reflects the recent update from Intel.

Also, fixing a bug that causing the installation script exit if any of `clang`, `clang-7`, `clang-8`, `clang-9` is not installed.

Fixes #3307 

### DIFF
--- a/scripts/ansible/roles/linux/openenclave/tasks/ubuntu/lvi-setup.yml
+++ b/scripts/ansible/roles/linux/openenclave/tasks/ubuntu/lvi-setup.yml
@@ -41,25 +41,25 @@
 
 - name: Download as/ld tarball
   get_url:
-    url: 'https://download.01.org/intel-sgx/sgx-linux/2.9/as.ld.objdump.gold.r1.tar.gz'
-    dest: '/tmp/as.ld.objdump.gold.r1.tar.gz'
+    url: 'https://download.01.org/intel-sgx/sgx-linux/2.10/as.ld.objdump.gold.r2.tar.gz'
+    dest: '/tmp/as.ld.objdump.gold.r2.tar.gz'
 
 - name: Extract as and ld binaries
   unarchive:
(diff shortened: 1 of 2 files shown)


anakrish:
bors try

radhikaj:
Thanks @mingweishih 
Do we update this script for each release?

mingweishih:
bors try

mingweishih:
> Thanks @mingweishih
> Do we update this script for each release?

Yes as long as there is an update from intel.

PR #3327, opened 2020-07-28
Remove libsgx-enclave-common-dev

anakrish:
OE SDK no longer depends on libsgx-enclave-common-dev.
This package brought in sgx headers as well as the symbolic link
libsgx_enclave_common.so => libsgx_enclave_common.so.1

Since OE SDK dynamically loads SGX libraries, this dev package is no longer
needed.

This PR mainly involves removing references to libsgx-enclave-common-dev.
Additionally, ansible validation is changed to test existence of
libsgx_enclave_common.so.1.

### DIFF
--- a/cmake/cpack_settings.cmake
+++ b/cmake/cpack_settings.cmake
@@ -14,7 +14,7 @@ set(CPACK_PACKAGING_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX})
 
 # CPack variables for Debian packages
 set(CPACK_DEBIAN_PACKAGE_DEPENDS
-    "libsgx-enclave-common (>=2.3.100.46354-1), libsgx-enclave-common-dev (>=2.3.100.0-1), libsgx-dcap-ql (>=1.0.100.46460-1.0), libsgx-dcap-ql-dev (>=1.0.100.46460-1.0)"
+    "libsgx-enclave-common (>=2.3.100.46354-1), libsgx-dcap-ql (>=1.0.100.46460-1.0), libsgx-dcap-ql-dev (>=1.0.100.46460-1.0)"
 )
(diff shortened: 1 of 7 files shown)


anakrish:
bors try

BRMcLaren:
While running in a container can you please provide a POC that runs with exposed sgx device without the sgx libraries running the full sample suite.

jhand2:
@BRMcLaren libsgx-enclave-common is absolutely needed. But the dev libraries (with headers) are not used at build time and thus are not needed.

BRMcLaren:
@jhand2 What about run time?

anakrish:
> While running in a container can you please provide a POC that runs with exposed sgx device without the sgx libraries running the full sample suite.

Just so that we are on the same page
- This PR removes an unused SGX package (libsgx-enclave-common-dev); not all SGX packages.
- There is not difference whether it is running in a container or not.

Given the above, is there any value in running in a container without any SGX libraries? I don't think that captures what this PR enables.

jhand2:
> @jhand2 What about run time?

Headers are never needed at runtime. That is the only additional thing the -dev package includes.

anakrish:
Closing this PR and instead created issue #3328

BRMcLaren:
Reopening and mooching off work :) 

Thanks for the investigation @anakrish and thanks for waiting! #3328

BRMcLaren:
bors try

BRMcLaren (approved):

CodeMonkeyLeet:
To fix #3328, shouldn't this PR also remove libsgx-dcap-ql-dev as well?

BRMcLaren:
@CodeMonkeyLeet that's a good point, it's not my PR so I removed the "fixes" to ensure no accidental closure.

CodeMonkeyLeet (approved):

PR #3329, opened 2020-07-28
Sync SGX_QuoteEx_Integration.md with implementation in PR 3268

shnwc:
Update SGX_QuoteEx_Integration.md to reflect the implementation in PR 3268

### DIFF
--- a/docs/DesignDocs/SGX_QuoteEx_Integration.md
+++ b/docs/DesignDocs/SGX_QuoteEx_Integration.md
@@ -1,13 +1,13 @@
 Attestation: OE SDK Integration with Intel® SGX SDK quote-ex Library for Generation of Evidence in New Formats
 ====
 
-This design document proposes an extension of the OE SDK implementation
+This design document proposes an extension of the OE SDK V0.9 implementation
 for integration with the Intel® SGX SDK quote-ex library, for support of
 generation of evidence in new SGX formats such as Enhanced Privacy ID (EPID).
 
 # Motivation
 
-The existing implementation of OE SDK SGX attestation,
+The V0.9 implementation of the OE SDK SGX attestation,
 based on the Intel® SGX SDK Data Center Attestation Primitives (DCAP)
 quote generation library (simply called the DCAP library),
 only supports generation of evidence in a single SGX ECDSA-p256 format.
@@ -27,7 +27,7 @@ and outside of data centers.
 
 The proposed extension only changes the internal implementation of the OE SDK
 attestation software stack. It does not impact the
-[OE SDK attestation API](https://github.com/openenclave/openenclave/pull/2949).
+[OE SDK attestation API](Attestation_API_Proposal.md).
 With the integration of the quote-ex library, an attester application enclave's
 call to OE SDK API `oe_attester_initialize()` triggers enumeration and
 registration of all supported attester plugins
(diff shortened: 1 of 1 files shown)


shnwc:
The latest commits incorporated the agreement in this week's SIG-attestation meeting. Please review. @CodeMonkeyLeet @dthaler @yentsanglee

shnwc:
bors try

docs/DesignDocs/SGX_QuoteEx_Integration.md
```diff
@@ -165,79 +167,66 @@ on the same platform.
...
+An SGX platform can have either the DCAP library or the quote-ex library,
 or both of them installed.
-Installation and configuration of the AESM background service is independent
-of the installation of the two libraries.
 
-When both libraries and the background services are all installed and configured,
-the quote-ex should library takes precedence, as it supports more evidence
-formats.
+* When only the DCAP library is installed, it will be used.
+* When only the quote-ex library is installed, it will be used.
```
CodeMonkeyLeet, line 176: Thanks for laying out the logic here, it's useful. I just noticed that this isn't quite how the implementation in #3268 works though: by checking for the regkey setting first; if `SGX_AESM_ADDR` is not set, it defaults to DCAP, even if it only quote-ex is installed. 

I think the design here is fine, but the implementation needs to handle this extra bit of logic (use quote-ex if regkey not set and dcap is absent).

docs/DesignDocs/SGX_QuoteEx_Integration.md
```diff
@@ -165,79 +167,66 @@ on the same platform.
...
-
 As described previously, the quote-ex library supports a superset of formats
 as compared to the DCAP library, though it always depends on
 a background service for quote generation.
 
 If on SGX platforms the OE SDK always installs with the AESM background service
 (as a hard dependency), then it is possible for the host-side plugin library
 to be linked at build-time only with the quote-ex library. With this option,
-the dependency on the DCAP library is dropped.
+the dependency on the DCAP library can be dropped.
```
CodeMonkeyLeet, line 209: Just for my understanding, is there any reason to soften this statement? The previous version asserts that we would stop supporting in-proc QE/PCE loading i.e. removing DCAP is a _feature_ of this option. It technically doesn't matter since we're not taking this option, but reading it document as a whole, I think it muddles the distinction between the options unnecessarily and the previous version is clearer in intent.

docs/DesignDocs/SGX_QuoteEx_Integration.md
yentsanglee: @CodeMonkeyLeet, I think the description between line 134 and 183 states the implementation.

shnwc: After viewing the implementation in `host\sgx\sgxquote.c` line 175, I still see that the implementation is inline with the described logic. Before checking `SGX_AESM_ADDR`, the code tries to load the DCAP library. If the DCAP is missing, the code will not check `SGX_AESM_ADDR` and try quote-ex directly.

Let me know if I still missed something.

CodeMonkeyLeet: @shnwc No, you're right, the existing code does take care of it. I was confused by the fact that we have a redundant `_load_sgx_dcap_ql` call in the else clause for `_use_quote_ex_library` conditional checks (effectively, `_use_quote_ex_library` would have loaded one or the other for the lifetime of the process by the time it returns.)

Side note: Consider reversing the order of the conditional checks in `_load_quote_ex_library_once` so that the DCAP load is elided if it's going to fall through to quote_ex load anyway you don't end up with an unused library load. 
```c
if (SGX_DCAP_IN_PROCESS_QUOTING() && _load_sgx_dcap_ql())
```

shnwc: Good point! Will make the change and create a new minor PR for it.

docs/DesignDocs/SGX_QuoteEx_Integration.md
shnwc: I think what I wanted to mean is something like "the dependency on the DCAP library will be dropped", which would happen in the future if and when this option is taken.

CodeMonkeyLeet: Yeah, I think you should state that as "is/will be" for this option then. I'll approve with the expectation that this is fixed up.

shnwc: Thanks! This is fixed in the latest commit.

PR #3332, opened 2020-07-29
Remove suppressing of mscompatibility flags.

anakrish:
It turns out the user in fact needs the compatibility flags since
code written for windows is being compiled for enclaves, and uses msvc constructs.

The original scenario that caused us to add the supression is not relevant anymore.


anakrish:
bors try

PR #3339, opened 2020-08-01
Add support for large thread-local data

### LINKED ISSUE #1161: Thread Local Storage: Support unlimited thread-local data size
The current thread-local storage implementation uses the free space available in the thread-data (`td_t`) page for allocating thread-local variables. Currently this is approximately 4KB - size of other `td_t` members (see `OE_THREAD_LOCAL_SPACE`).

If an enclave binary contains thread-local data that cannot fit in this space, today we error out in ELF loading. We need to extend this scheme so that thread-local data of any size can be used within the enclave.

The key rules regarding the thread-local-data segment managed by the FS register are:
1) the thread-local variables are accessed via negative offsets to the FS register
2) the FS register must lie on a 4K boundary
3) FS[0] must contain the value of FS itself

Today data layout is as follows:
```
-------------------- 4KB page start
td_t              <-- GS register
[thread local
 data block ... ]
-------------------- 4KB page start
[explicit thread  <-- FS register
 local data set by
 by key ... ]
-------------------- 4KB page start
```
See enclave/core/thread.c `oe_thread_key_create()` for how the explicit thread local data is allocated.

Because the FS register used to index into thread local variables lies on the next page boundary, we always have to allocate at least another page, though we reuse the rest of that page for thread local variables allocated by key.

When adding support for expanding thread-local data size, we should also consider how to minimize wasted space.

CC: @mikbras @CodeMonkeyLeet @anitagov @lilyfang

thomasten:
Calculate number of pages required for all TLS of the enclave image and
add them during enclave creation.

The distance between TCS and TD now depends on TLS size, so macro
TD_FROM_TCS is replaced by variable td_from_tcs_offset whose value is
patched during enclave creation.

Fix #1161

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - OE SDK Release packages can now be built on non-SGX and non-FLC machines.
+- Support for arbitrarily large thread-local data for SGX machines.
 
 ### Changed
 
@@ -46,7 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   application EDL. See [system EDL opt-in document](docs/DesignDocs/system_ocall_opt_in.md#how-to-port-your-application) for more information.
   - Note: SDK users would need to import logging.edl to enable logging. Logging is disabled by default.
   - See [System edls](docs/SystemEdls.md) for list of all edls and associated OCalls.
-  - A known issue is that different enclaves importing functions from System EDLs cannot be loaded by the same host app unless all of the functions were imported with exactly the same ordinals. See #3250 for details. This will be addressed in the next release based on design proposal #3086. 
+  - A known issue is that different enclaves importing functions from System EDLs cannot be loaded by the same host app unless all of the functions were imported with exactly the same ordinals. See #3250 for details. This will be addressed in the next release based on design proposal #3086.
   - A workaround for this issue in the meantime is to define a standard import EDL for any enclaves that need to be loaded into the same host app. Ensuring this shared EDL is then the first import in each enclave's EDL will result in the common imports being assigned the same ordinals in each resulting enclave.
 - Mark APIs in include/openenclave/attestation/sgx/attester.h and verifier.h as experimental.
 - Remove CRL_ISSUER_CHAIN_PCK_PROC_CA field from endorsement struct define in include/openenclave/bits/attestation.h.
--- a/host/sgx/loadelf.c
+++ b/host/sgx/loadelf.c
@@ -622,7 +622,10 @@ done:
     return result;
 }
 
-static oe_result_t _patch(oe_enclave_image_t* image, size_t enclave_size)
+static oe_result_t _patch(
+    oe_enclave_image_t* image,
+    oe_sgx_load_context_t* context,
+    size_t enclave_size)
 {
     oe_result_t result = OE_UNEXPECTED;
     oe_sgx_enclave_properties_t* oeprops;
@@ -695,12 +698,12 @@ static oe_result_t _patch(oe_enclave_image_t* image, size_t enclave_size)
             oe_round_up_to_multiple(image->tbss_size, image->tbss_size);
     }
 
-    if (aligned_size > OE_THREAD_LOCAL_SPACE)
-    {
-        OE_TRACE_ERROR(
-            "Thread-local variables exceed available thread-local space.\n");
-        OE_RAISE(OE_INVALID_IMAGE);
-    }
+    aligned_size = oe_round_up_to_multiple(aligned_size, OE_PAGE_SIZE);
+    _set_uint64_t_dynamic_symbol_value(
+        image,
+        "td_from_tcs_offset",
+        aligned_size + OE_SGX_NUM_CONTROL_PAGES * OE_PAGE_SIZE);
+    context->num_tls_pages = aligned_size / OE_PAGE_SIZE;
 
     /* Clear the hash when taking the measure */
     memset(oeprops->sigstruct, 0, sizeof(oeprops->sigstruct));
@@ -781,7 +784,7 @@ oe_result_t oe_load_elf_enclave_image(
     image->type = OE_IMAGE_TYPE_ELF;
     image->calculate_size = _calculate_size;
     image->add_pages = _add_pages;
-    image->patch = _patch;
+    image->sgx_patch = _patch;
     image->sgx_load_enclave_properties = _sgx_load_enclave_properties;
     image->sgx_update_enclave_properties = _sgx_update_enclave_properties;
     image->unload = _unload;
--- a/include/openenclave/internal/load.h
+++ b/include/openenclave/internal/load.h
@@ -99,7 +99,10 @@ struct _oe_enclave_image
         oe_enclave_t* enclave,
         uint64_t* vaddr);
 
-    oe_result_t (*patch)(oe_enclave_image_t* image, size_t enclave_size);
+    oe_result_t (*sgx_patch)(
+        oe_enclave_image_t* image,
+        oe_sgx_load_context_t* context,
+        size_t enclave_size);
 
     oe_result_t (*sgx_load_enclave_properties)(
         const oe_enclave_image_t* image,
--- a/tests/thread_local_large/enc/enc.cpp
+++ b/tests/thread_local_large/enc/enc.cpp
@@ -1,0 +1,34 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/internal/tests.h>
+#include <array>
+#include <cstdint>
+#include "thread_local_large_t.h"
+
+using namespace std;
+
+void enc_test()
+{
+    static thread_local array<uint8_t, 9000> a;
+    for (auto& x : a)
+    {
+        OE_TEST(x == 0);
+        x = 2;
+    }
+
+    // Perform an ocall.
+    host_nop();
+
+    // Expect that TLS is the same as before the ocall.
+    for (const auto x : a)
+        OE_TEST(x == 2);
+}
+
+OE_SET_ENCLAVE_SGX(
+    0,    /* ProductID */
+    0,    /* SecurityVersion */
+    true, /* Debug */
+    64,   /* NumHeapPages */
+    16,   /* NumStackPages */
+    1);   /* NumTCS */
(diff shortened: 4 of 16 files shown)


anakrish:
bors try

tests/thread_local_large/enc/enc.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+void enc_test()
+{
+    static thread_local array<uint8_t, 9000> a;
+    for (auto& x : a)
+    {
+        OE_TEST(x == 0);
+        x = 2;
+    }
+    host_nop();
```
anakrish, line 19: What is the purpose of `host_nop`?

host/sgx/loadelf.c
```diff
@@ -713,12 +716,12 @@ static oe_result_t _patch(oe_enclave_image_t* image, size_t enclave_size)
...
     }
 
-    if (aligned_size > OE_THREAD_LOCAL_SPACE)
-    {
-        OE_TRACE_ERROR(
-            "Thread-local variables exceed available thread-local space.\n");
-        OE_RAISE(OE_FAILURE);
-    }
+    aligned_size = oe_round_up_to_multiple(aligned_size, OE_PAGE_SIZE);
+    _set_uint64_t_symbol_value(
```
jhand2, line 720: This depends on the enclave image including symbols, which I don't think we can rely on ( I know we do it in other places here but I don't know that we should). Is there a way to instead compute the value at enclave compile time so that the number can be built into the enclave image itself?

tests/thread_local_large/enc/enc.cpp
thomasten: This should make sure that TLS is fine after an ocall. I'll add a comment.

host/sgx/loadelf.c
thomasten: I guess the TLS size is only known after linking.
I can think of two ways:
* Put the variable at a known location within a specific section (e.g., .oeinfo or its own section) and patch it there.
* Let the enclave parse its ELF headers during TLS initialization to get the values.

thomasten:
> 
> 
> However, we may need more work to support the debugger since the ptrace library depends on a fixed `TD_FROM_TCS` offset:
> 
> https://github.com/openenclave/openenclave/blob/49d9b2d674d84ef2bffa5d54089b59de03c30fe8/debugger/ptraceLib/enclave_context.c#L174
> 
> @MWShan How would you recommend making `oeptrace` library support a variable TD_FROM_TCS offset?

_get_enclave_ssa_frame_size seems to always return OE_DEFAULT_SSA_FRAME_SIZE because __ssa_frame_size is unused. Can we thus remove this function or will it be needed in future?

The TD address might be calculated as in the Intel SDK: https://github.com/intel/linux-sgx/blob/master/sdk/debugger_interface/linux/se_ptrace.c#L194

anakrish:
bors try

host/sgx/loadelf.c
jhand2: I think both of those are decent solutions, although I'm leaning toward the second one. A 3rd option is to just pass in a value during an initialization ecall.

@CodeMonkeyLeet @anakrish @mingweishih do any of you have a preference?

CodeMonkeyLeet: Given that `td_from_tcs_offset` is just a derived value from constants and other values we already patched into the image as part of this function, we could just recalculate and cache it on enclave init. The tricky part is that it gets used very early in the `oe_enter` call in assembly, and I'm not sure it's worth the extra complexity in that codepath, so I think the current approach reasonable.

anakrish: > This depends on the enclave image including symbols, which I don't think we can rely on ( I know we do it in other places here but I don't know that we should). Is there a way to instead compute the value at enclave compile time so that the number can be built into the enclave image itself?

@jhand2 Great point.

We are incorrectly relying on regular symbols instead of dynamic symbols.

A loader does not rely on regular symbols to load an application/shared library. Instead it relies on dynamic symbols.
Dynamic symbols cannot be stripped away.

libcurl.so does not have any regular symbols whereas it has dynamic symbols necessarsy to perform loading/linking:
```bash
$objdump -t /usr/lib/x86_64-linux-gnu/libcurl.so

/usr/lib/x86_64-linux-gnu/libcurl.so:     file format elf64-x86-64

SYMBOL TABLE:
no symbols

$ objdump -T /usr/lib/x86_64-linux-gnu/libcurl.so
...
000000000002bff0 g    DF .text	000000000000000a  CURL_OPENSSL_4 curl_global_init
0000000000038820 g    DF .text	00000000000000b8  CURL_OPENSSL_4 curl_multi_strerror
000000000004e490 g    DF .text	00000000000000ef  CURL_OPENSSL_4 curl_pushheader_byname
000000000002ca60 g    DF .text	0000000000000083  CURL_OPENSSL_4 curl_easy_recv
0000000000036310 g    DF .text	0000000000000189  CURL_OPENSSL_4 curl_multi_perform
0000000000024a50 g    DF .text	0000000000000037  CURL_OPENSSL_4 curl_getenv
00000000000367f0 g    DF .text	0000000000000041  CURL_OPENSSL_4 curl_multi_socket
000000000002c460 g    DF .text	0000000000000055  CURL_OPENSSL_4 curl_easy_getinfo
000000000002c000 g    DF .text	0000000000000089  CURL_OPENSSL_4 curl_global_init_mem
000000000027e4c0 g    DO .bss	00000000000000c8  CURL_OPENSSL_4 curl_jmpenv
0000000000024bd0 g    DF .text	000000000000000c  CURL_OPENSSL_4 curl_escape
000000000000fde0 g    DF .text	0000000000000cf9  CURL_OPENSSL_4 curl_formadd
00000000000520e0 g    DF .text	0000000000000084  CURL_OPENSSL_4 curl_mime_encoder
```

When we strip out enclaves by specifying `-s` to the linker or by using `strip`, it no longer has any symbols and therefore loading fails as you pointed out:
```bash
$ strip echo_enc
$ objdump -t echo_enc

echo_enc:     file format elf64-x86-64

SYMBOL TABLE:
no symbols
```

However it still retains the symbols that we've specified as exported (ie needed for loading):
```bash
objdump -T echo_enc

echo_enc:     file format elf64-x86-64

DYNAMIC SYMBOL TABLE:
0000000000061edc g    DF .text	0000000000000000 memset
0000000000123fa0 g    DF .text	0000000000000000 oe_longjmp
0000000000145104 g    DF .text	0000000000000000 oe_egetkey
000000000006da10 g    DF .text	0000000000000000 abort
000000000035f9e0 g    D  .bss	0000000000000000 __bss_start
00000000001204bc g    DF .text	0000000000000000 oe_continue_execution
0000000000365730 g    D  .bss	0000000000000000 _end
000000000012046f g    DF .text	0000000000000000 oe_restore_partial_context
0000000000144b17 g    DF .text	0000000000000000 oe_rdseed
000000000035f9e0 g    D  .data	0000000000000000 _edata
0000000000016890 g    DF .text	0000000000000005 _start
0000000000016895 g    DF .text	0000000000000005 oe_call_link_enclave
0000000000061eb4 g    DF .text	0000000000000000 memmove
0000000000144b0c g    DF .text	0000000000000000 oe_rdrand
000000000012052b g    DF .text	00000000000001be oe_enter
00000000001203fc g    DF .text	0000000000000000 oe_snap_current_context
0000000000061e80 g    DF .text	0000000000000000 memcpy
0000000000127bb8 g    DF .text	0000000000000000 oe_setjmp
00000000001206e9 g    DF .text	000000000000004a oe_exception_dispatcher
```

To be consistent with the way loaders work, we'd need to
- export all the symbols necessary for loading
- Change the loader to use only dynamic symbols (they are meant for the loader)

I have a PR that does the above changes: https://github.com/openenclave/openenclave/pull/3359

I'd recommend that we go the above approach and make our loader consistent with the way loaders work.
That will allow enclaves to be loaded even if symbols are not available.

Of the two solutions proposed by @thomasten, (1) would  likely be easier to implement. (2) may not be feasible since the host-side loader needs to set the value of `_enclave_rva` so that the enclave can securely determine the enclave start address.

It would be preferable to revisit the above solutions if ever we need the enclave not have any dynamic symbols as well (that likely may not ever be possible since libcxx always exports symbols, and we'd like to pursure multi-module loading which needs dynamic symbols).

For the foreseeable future, changing the loader to rely only on dynamic symbols is the right approach.

anakrish:
bors try

anakrish:
bors try

include/openenclave/internal/load.h
```diff
@@ -105,7 +105,10 @@ struct _oe_enclave_image
         oe_enclave_t* enclave,
         uint64_t* vaddr);
 
-    oe_result_t (*patch)(oe_enclave_image_t* image, size_t enclave_size);
+    oe_result_t (*patch)(
```
dthaler, line 108: Since this API is changed in line 110 to be SGX-specific instead of TEE-agnostic, the function needs to be renamed to start with sgx, like line 113 already does

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - OE SDK Release packages can now be built on non-SGX and non-FLC machines.
+- Support for arbitrarily large thread-local data.
```
dthaler, line 15: Shouldn't this line say "for SGX machines"?   I don't see any changes for anything other than SGX.

host/sgx/loadelf.c
thomasten: Thanks @anakrish! I rebased to master and exported td_from_tcs_offset.

include/openenclave/internal/load.h
thomasten: renamed to sgx_patch

CHANGELOG.md
thomasten: changed as suggested

anakrish:
I've removed `oeptrace` library's use of hard-coded td_t offset in PR #3381

anakrish:
bors try

tests/thread_local_large/thread_local_large.edl
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+enclave {
+#ifdef OE_SGX
```
anakrish, line 5: To allow the enclave to work with code coverage, we'd need to do this currently:

```suggestion
#ifdef OE_SGX
        from "openenclave/edl/fcntl.edl" import *;
```

tests/thread_local_large/thread_local_large.edl
```diff
@@ -0,0 +1,19 @@
...
+        from "openenclave/edl/fcntl.edl" import *;
```
mingweishih, line 6:
```suggestion
    from "openenclave/edl/fcntl.edl" import *;
```

PR #3340, opened 2020-08-01
Add fsync and fdatasync

thomasten:
Added support for fsync and fdatasync syscalls.

On Windows, both invoke FlushFileBuffers which seems to be the closest available.


anakrish:
bors try

radhikaj:
@nonpolarity @jhand2  Can you review this one please

syscall/hostcalls.c
```diff
@@ -358,6 +358,15 @@ oe_result_t _oe_syscall_flock_ocall(
 }
 OE_WEAK_ALIAS(_oe_syscall_flock_ocall, oe_syscall_flock_ocall);
 
+oe_result_t _oe_syscall_fsync_ocall(int* _retval, oe_host_fd_t fd)
```
mingweishih, line 361: You need to have a forward declaration of this function like other functions do.

thomasten: Can you explain when this is needed and when it's not? For example, _oe_syscall_flock_ocall doesn't have a forward declaration.

mingweishih: In some cases it will throw the missing-prototypes warning (which we treat as an error). I haven't look into why flock is not the case, but just for consistency it would be good to add.

mingweishih: Besides, can you update the `edl_opt_out` test to ensure those default implementations are covered?

thomasten: done

PR #3341, opened 2020-08-03
Update VisualStudioWindows.md

JeromeCustodio:
i've updated the LLVM download link with version 10. I suppose it would be beneficial to get the latest one to get the updates and fixes. 
In my usecase, i need to use nlohmann/json. version 7 doesn't work (json::dump() ).  However when i upgraded to v10, it started working (as it is on the list of tested clang/llvm versions)

### DIFF
--- a/docs/GettingStartedDocs/VisualStudioWindows.md
+++ b/docs/GettingStartedDocs/VisualStudioWindows.md
@@ -24,7 +24,7 @@ You will need the following:
 - Intel's PSW2.6. This should be automatically installed on Windows 10 systems that support SGX.  If manual install is needed:
   - For systems with support for SGX1: [Intel's PSW 2.6,Intel Enclave Common API library](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/WindowsManualSGX1Prereqs.md)
   - For systems with support for SGX1+FLC: [Intel's PSW2.6, Intel's Data Center Attestation Primitives and related dependencies](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/WindowsManualSGX1FLCDCAPPrereqs.md)
-- [Clang/LLVM for Windows 64-bit](http://releases.llvm.org/7.0.1/LLVM-7.0.1-win64.exe).  The installer will ask whether LLVM should be added to the PATH, and this must be done at least for the current user, so Visual Studio can find it.
+- [Clang/LLVM for Windows 64-bit](https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/LLVM-10.0.0-win64.exe).  The installer will ask whether LLVM should be added to the PATH, and this must be done at least for the current user, so Visual Studio can find it.
 
 ## Walkthrough: Creating a C/C++ Enclave Application
 


radhikaj:
@anakrish  Can you comment on this?

docs/GettingStartedDocs/VisualStudioWindows.md
```diff
@@ -24,7 +24,7 @@ You will need the following:
 - Intel's PSW2.6. This should be automatically installed on Windows 10 systems that support SGX.  If manual install is needed:
   - For systems with support for SGX1: [Intel's PSW 2.6,Intel Enclave Common API library](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/WindowsManualSGX1Prereqs.md)
   - For systems with support for SGX1+FLC: [Intel's PSW2.6, Intel's Data Center Attestation Primitives and related dependencies](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/WindowsManualSGX1FLCDCAPPrereqs.md)
-- [Clang/LLVM for Windows 64-bit](http://releases.llvm.org/7.0.1/LLVM-7.0.1-win64.exe).  The installer will ask whether LLVM should be added to the PATH, and this must be done at least for the current user, so Visual Studio can find it.
+- [Clang/LLVM for Windows 64-bit](https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/LLVM-10.0.0-win64.exe).  The installer will ask whether LLVM should be added to the PATH, and this must be done at least for the current user, so Visual Studio can find it.
```
radhikaj, line 27: @JeromeCustodio Thanks for this contribution. I would like to say that Ci/Cd tests clang-7. So ideally, we would like to tell developers to install clang-7. 
Could you please provide details on what issues you ran into with clang-7.

JeromeCustodio: hi @radhikaj, actually there's, no issue with clang-7.  it was working fine,  It's sociability with some single header library files I am using (nlohmann::json) which works better on clang-10.

mingweishih: @JeromeCustodio In this case, is it necessary for this change? Given that our stabilized support is still based on clang-7, even the document change would create inconsistency to that.

JeromeCustodio: @mingweishih , I suppose the change is not required.  I haven't seen any issues using clang 10, will continue to use this for now. 
Thank you

mingweishih: Thanks again for using the OE SDK and the contribution. We're still in the process of evaluating/updating our compiler support (meaning that all our CI/CD should pass with the newer version of the compiler), which would take a while. Before the process is done, we cannot officially claim that the new version of compiler is supported (even though it's actually working as in your case).

radhikaj:
@JeromeCustodio  Thanks a lot for working on this. For now, can we close this?

JeromeCustodio:
will close pull request. resolution: no need to update documentation as ci/cd is set to llvm clang 7.

PR #3343, opened 2020-08-04
Fix oe_log_message() and time functions.

nonpolarity:
Fix oe_log_message().
1. Initialize usecs on Both Linux and Windows.
2. Add host thread id to callback function type.

Fix several time functions:
1. Implement equivalent gettimeofday() on Windows.
2. Add oe_localtime().

### DIFF
--- a/host/linux/time.c
+++ b/host/linux/time.c
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 #include <errno.h>
+#include <openenclave/internal/localtime.h>
 #include <openenclave/internal/time.h>
 #include <time.h>
 #include "../ocalls.h"
@@ -27,4 +28,9 @@ void oe_handle_get_time(uint64_t arg_in, uint64_t* arg_out)
 
     if (arg_out)
         *arg_out = _time();
+}
+
+int oe_localtime(time_t* timep, struct tm* result)
+{
+    return !localtime_r(timep, result);
 }
--- a/host/traceh.c
+++ b/host/traceh.c
@@ -4,8 +4,10 @@
 #include <openenclave/corelibc/limits.h>
 #include <openenclave/internal/calls.h>
 #include <openenclave/internal/datetime.h>
+#include <openenclave/internal/localtime.h>
 #include <openenclave/internal/raise.h>
 #include <openenclave/internal/safecrt.h>
+#include <openenclave/internal/time.h>
 #include <openenclave/internal/trace.h>
 #include <stdarg.h>
 #include <stdio.h>
@@ -338,11 +340,14 @@ void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
     // get timestamp for log
     struct tm t;
     time_t lt = time(NULL);
-    gmtime_r(&lt, &t);
+    struct timeval tv;
+    gettimeofday(&tv, NULL);
+    lt = tv.tv_sec;
+    oe_localtime(&lt, &t);
 
-    char time[20];
-    strftime(time, sizeof(time), "%Y-%m-%dT%H:%M:%S", &t);
-    long int usecs = 0;
+    char time[25];
+    strftime(time, sizeof(time), "%Y-%m-%dT%H:%M:%S%z", &t);
+    long int usecs = tv.tv_usec;
 
     if (!_initialized)
     {
@@ -355,7 +360,13 @@ void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
         if (oe_log_callback)
         {
             (oe_log_callback)(
-                oe_log_context, is_enclave, time, usecs, level, message);
+                oe_log_context,
+                is_enclave,
+                time,
+                usecs,
+                level,
+                (uint64_t)oe_thread_self(),
+                message);
             goto done;
         }
 
--- a/include/openenclave/internal/time.h
+++ b/include/openenclave/internal/time.h
@@ -22,6 +22,20 @@ OE_EXTERNC_BEGIN
 
 uint64_t oe_get_time(void);
 
+#ifdef _WIN32
+/*
+**==============================================================================
+**
+** gettimeofday()
+**
+**     Get seconds and useconds elapsed since the Epoch.
+**
+**==============================================================================
+*/
+
+int gettimeofday(struct timeval* tv, struct timezone* tz);
+#endif
+
 OE_EXTERNC_END
 
 #endif /* _OE_INCLUDE_TIME_H */
--- a/include/openenclave/internal/trace.h
+++ b/include/openenclave/internal/trace.h
@@ -34,6 +34,7 @@ typedef void (*oe_log_callback_t)(
     const char* time,
     long int usecs,
     oe_log_level_t level,
+    uint64_t host_thread_id,
     const char* message);
 oe_result_t oe_log_set_callback(void* context, oe_log_callback_t callback);
 extern void* oe_log_context;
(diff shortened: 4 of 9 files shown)


nonpolarity:
bors try

nonpolarity:
bors try

host/traceh.c
```diff
@@ -331,18 +331,43 @@ oe_result_t oe_log_set_callback(void* context, oe_log_callback_t callback)
     return OE_UNEXPECTED;
 }
 
+#ifdef _WIN32
+int gettimeofday(struct timeval* tv, struct timezone* tzp)
+{
+    static const uint64_t EPOCH = ((uint64_t)116444736000000000ULL);
```
anakrish, line 337: How was this EPOCH value chosen? Is this expected to be the same on all Windows Machines?

nonpolarity: EPOCH = 10000000 * 60 * 60 * 24 * 365 * (1970 - 1601) + 89. Yes this value is universal.

anakrish:
bors try

nonpolarity:
bors try

host/traceh.c
```diff
@@ -331,18 +331,46 @@ oe_result_t oe_log_set_callback(void* context, oe_log_callback_t callback)
...
+#ifdef _WIN32
+    localtime_s(&t, &lt);
+#else
+    localtime_r(&lt, &t);
+#endif
 
     char time[20];
-    strftime(time, sizeof(time), "%Y-%m-%dT%H:%M:%S", &t);
-    long int usecs = 0;
+    strftime(time, sizeof(time), "%Y-%m-%d %H:%M:%S", &t);
```
CodeMonkeyLeet, line 372: If we want to represent local time instead of UTC, I would suggest logging the time zone differential (i.e. use the format string `"%Y-%m-%dT%H:%M:%S%z"`)

The original format string looks like it was trying to be compliant with [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) but it really should have included the 'Z' at the end.

include/openenclave/internal/trace.h
```diff
@@ -34,6 +34,7 @@ typedef void (*oe_log_callback_t)(
     const char* time,
     long int usecs,
     oe_log_level_t level,
+    uint64_t host_thread_id,
```
CodeMonkeyLeet, line 37: Is this change supposed to be part of this PR? It doesn't look used anywhere.

nonpolarity:
> 
> 
> Thanks for fixing this up! See suggestions inline.

@CodeMonkeyLeet Thanks for your information! Yes much duplication. Let me figure out.

host/traceh.c
nonpolarity: Thanks, _time() is reused now.

host/traceh.c
```diff
@@ -331,18 +331,46 @@ oe_result_t oe_log_set_callback(void* context, oe_log_callback_t callback)
...
 void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
 {
     // get timestamp for log
     struct tm t;
     time_t lt = time(NULL);
-    gmtime_r(&lt, &t);
+    struct timeval tv;
+    gettimeofday(&tv, NULL);
+    lt = tv.tv_sec;
+#ifdef _WIN32
```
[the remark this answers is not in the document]
nonpolarity: oe_localtime() is added.

host/traceh.c
nonpolarity: Fixed. Thanks, I never noticed the detail.

include/openenclave/internal/trace.h
nonpolarity: Yes this change is part of the oe_log_message() callbak feature.

nonpolarity:
bors try

nonpolarity:
bors try

host/linux/time.c
```diff
@@ -28,3 +29,8 @@ void oe_handle_get_time(uint64_t arg_in, uint64_t* arg_out)
     if (arg_out)
         *arg_out = _time();
 }
+
+int oe_localtime(time_t* timep, struct tm* result)
+{
+    return !localtime_r((const time_t*)timep, result);
```
dthaler, line 35: Why is the cast needed here?   You should be able to pass a non-const ptr to a const argument without a cast.

host/traceh.c
```diff
@@ -24,6 +25,8 @@
 #if defined(__linux__)
 #define sprintf_s(buffer, size, format, argument) \
     sprintf(buffer, format, argument)
+#else
+#include <openenclave/internal/time.h>
```
dthaler, line 29: Why is this inside an else clause?  Would be more readable to unconditionally include it.

include/openenclave/internal/time.h
```diff
@@ -22,6 +22,20 @@ OE_EXTERNC_BEGIN
...
+**==============================================================================
+**
+** gettimeofday()
+**
+**     Get seconds and useconds elapsed since the Epoch.
+**
+**==============================================================================
+*/
+
+#ifdef _WIN32
```
dthaler, line 35: move ifdef above comment in 25

nonpolarity:
bors try

host/linux/time.c
nonpolarity: Fixed, thanks!

host/traceh.c
nonpolarity: Fixed, thanks!

include/openenclave/internal/time.h
nonpolarity: Fixed, thanks!

nonpolarity:
bors try

nonpolarity:
bors try

host/traceh.c
```diff
@@ -338,11 +340,19 @@ void oe_log_message(bool is_enclave, oe_log_level_t level, const char* message)
...
-    strftime(time, sizeof(time), "%Y-%m-%dT%H:%M:%S", &t);
-    long int usecs = 0;
+    struct timeval tv;
+    gettimeofday(&tv, NULL);
+    lt = tv.tv_sec;
+    oe_localtime(&lt, &t);
+
+#ifdef _WIN32
+    char time[25];
+    strftime(time, sizeof(time), "%Y-%m-%dT%H:%M:%S%z", &t);
```
CodeMonkeyLeet, line 350: Why are we using different format strings for Windows vs. Linux? It seems like Linux should also be using the lowercase `%z` format for the ISO standard offset (e.g. +100) rather than the timezone code? It also seems to account for the difference in the size of your time buffers. 

Also, it would be useful to document the time format in your design doc we produce by default; it would be helpful for developers using the new public API to understand what they get and if they want to change it in the callback they provide.

nonpolarity: Thanks! Now it is much better. I will add these useful info to the design doc.

nonpolarity:
bors try

host/traceh.c
dthaler: Per SIG-Arch meeting this morning, the callback should change to take &t directly and so there is no need to convert to a string here.

nonpolarity: @dthaler Thanks for the suggestion, yes both the public api and internal api are planned to take &t. 
For this pr, such change may introduce dependency on <time.h> which is prohibited by compiling option "-nostdinc".
For the coming pr for public api, since new separate header files are introduced, it is very simple to solve the dependency problem.
So my suggestion is can we postpone such change to the next coming pr?

dthaler: Ok to cover in next coming PR.

nonpolarity:
bors try

PR #3356, opened 2020-08-06
Make default SGX_SIGSTRUCT_ATTRIBUTEMASK_XFRM more restrictive

### LINKED ISSUE #3316: Default SGX SIGSTRUCT ATTRIBUTEMASK allows the loader to enable MPX inside the enclave
The [default SGX SIGSTRUCT ATTRIBUTEMASK](https://github.com/openenclave/openenclave/blob/05d109a5706fe4c439f64cd68ffee6e2430b9542/include/openenclave/bits/sgx/sgxtypes.h#L139) used by the signing tool allows the untrusted (outside of SGX enclave TCB) enclave loader to enable MPX (Intel [SDM](https://software.intel.com/content/www/us/en/develop/articles/intel-sdm.html#combined) chapter 17) inside the enclave.

To securely use MPX inside the enclave, the enclave developer must carefully design how the MPX Bound Directory and Bound Table are allocated and accessed. OE SDK does not provide any framework or library to accomplish that. If a developer unintentionally include MPX instructions in the enclave without the proper design (for example, by linking with a library compiled to use MPX features), and allows the untrusted loader to enable MPX inside the enclave, a malicious OS/loader might be able to set Bound Directory and/or Bound Table to overlap with enclave memory and trick the MPX instruction executed inside the Enclave to unintentionally overwrite Enclave code/data. Further analysis is required to understand whether this type of attack is realistic.

It's safer to set the SIGSTRUCT.ATTRIBUTEMASK default to disable MPX inside the enclave and require the enclave developers to explicitly override the default if they ant to support MPX inside the enclave.

CodeMonkeyLeet:
Fixes #3316.

### DIFF
--- a/host/sgx/sgxsign.c
+++ b/host/sgx/sgxsign.c
@@ -630,9 +630,7 @@ static oe_result_t _init_sigstruct(
 
     /* sgx_sigstruct_t.attributemask */
     sigstruct->attributemask.flags = SGX_SIGSTRUCT_ATTRIBUTEMASK_FLAGS;
-    sigstruct->attributemask.xfrm =
-        SGX_SIGSTRUCT_ATTRIBUTEMASK_XFRM; // Reason this mask is 0 is because we
-                                          // don't enforce XFRM in signature
+    sigstruct->attributemask.xfrm = SGX_SIGSTRUCT_ATTRIBUTEMASK_XFRM;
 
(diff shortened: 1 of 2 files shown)


CodeMonkeyLeet:
@bodzhang @MWShan Please take a look at the fix.

radhikaj:
@bodzhang  Can you please review? Thanks

PR #3357, opened 2020-08-06
Remove compaction bit from OE_XSAVE_INITIAL_STATE

### LINKED ISSUE #3350: Avoid compact form of xstate
OE assumes the compact form of xstate is always supported. This is true for machines with sgx capability, but not true for all intel CPUs. Since running the simulation mode could happen on non-sgx machines, we should avoid using the compact form to allow maximum compatibility.

https://github.com/openenclave/openenclave/blob/eb1f65bd181212d8c22afaec7050319b55ffe5de/enclave/core/sgx/asmdefs.c#L40

CodeMonkeyLeet:
Fixes #3350


jxyang:
LGTM

PR #3358, opened 2020-08-06
Add helper functions to serialize, deserialize custom claims, issue 3303

shants:
Added helper functions to serialize and deserialize custom claims

### DIFF
--- a/common/custom_claims.c
+++ b/common/custom_claims.c
@@ -1,0 +1,286 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/attestation/custom_claims.h>
+#include <openenclave/internal/raise.h>
+#include <openenclave/internal/safecrt.h>
+#include <openenclave/internal/safemath.h>
+
+#include "common.h"
+
+/**
+ * Serialized buffer will have a header with version and number of
+ * claims
+ */
+typedef struct _oe_custom_claims_header
+{
+    uint64_t version;
+    uint64_t num_claims;
+} oe_custom_claims_header_t;
+
+/**
+ * Serialized entry for custom claims. Each entry will have the name and value
+ * sizes and then the contents of the name and value respectively.
+ */
+typedef struct _oe_custom_claims_entry
+{
+    uint64_t name_size;
+    uint64_t value_size;
+    uint8_t name[];
+    // name_size bytes follow.
+    // value_size_bytes follow.
+} oe_custom_claims_entry_t;
+
+oe_result_t oe_free_serialized_custom_claims(uint8_t* custom_claims_buffer)
+{
+    oe_free(custom_claims_buffer);
+    return OE_OK;
+}
+
+static void _free_claim(oe_claim_t* claim)
+{
+    oe_free(claim->name);
+    oe_free(claim->value);
+}
+
+oe_result_t oe_free_custom_claims(oe_claim_t* claims, size_t claims_length)
+{
+    if (!claims)
+        return OE_OK;
+
+    for (size_t i = 0; i < claims_length; i++)
+        _free_claim(&claims[i]);
+    oe_free(claims);
+    return OE_OK;
+}
+
+static oe_result_t _add_claim(
+    oe_claim_t* claim,
+    const uint8_t* name,
+    size_t name_size, // Must cover the '\0' at end of string
+    const void* value,
+    size_t value_size)
+{
+    if (*(name + name_size - 1) != '\0')
+        return OE_CONSTRAINT_FAILED;
+
+    claim->name = (char*)oe_malloc(name_size);
+    if (claim->name == NULL)
+        return OE_OUT_OF_MEMORY;
+    oe_memcpy_s(claim->name, name_size, name, name_size);
+
+    claim->value = (uint8_t*)oe_malloc(value_size);
+    if (claim->value == NULL)
+    {
+        oe_free(claim->name);
+        claim->name = NULL;
+        return OE_OUT_OF_MEMORY;
+    }
+    oe_memcpy_s(claim->value, value_size, value, value_size);
+    claim->value_size = value_size;
+
+    return OE_OK;
+}
+
+static oe_result_t _fill_array_with_custom_claims(
+    const uint8_t* claims_buffer,
+    size_t claims_buffer_size,
+    oe_claim_t* claims)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    oe_custom_claims_header_t* header =
+        (oe_custom_claims_header_t*)claims_buffer;
+    size_t claims_index = 0;
+
+    if (claims_buffer_size < sizeof(*header))
+    {
+        OE_RAISE(OE_CONSTRAINT_FAILED);
+    }
+
+    claims_buffer += sizeof(*header);
+    claims_buffer_size -= sizeof(*header);
+    for (uint64_t i = 0; i < header->num_claims; i++)
+    {
+        oe_custom_claims_entry_t* entry =
+            (oe_custom_claims_entry_t*)claims_buffer;
+        uint64_t size = 0;
+
+        // Sanity check sizes.
+        if (claims_buffer_size < sizeof(*entry))
+            OE_RAISE(OE_CONSTRAINT_FAILED);
+
+        OE_CHECK(oe_safe_add_u64(sizeof(*entry), entry->name_size, &size));
+        OE_CHECK(oe_safe_add_u64(size, entry->value_size, &size));
+
+        if (claims_buffer_size < size)
+            OE_RAISE(OE_CONSTRAINT_FAILED);
+
+        // Add the claim.
+        OE_CHECK(_add_claim(
+            &claims[claims_index++],
+            entry->name,
+            entry->name_size,
+            entry->name + entry->name_size,
+            entry->value_size));
+
+        OE_CHECK(
+            oe_safe_sub_u64(claims_buffer_size, size, &claims_buffer_size));
+        claims_buffer += size;
+    }
+
+    result = OE_OK;
+
+done:
+    if (result != OE_OK)
+    {
+        for (size_t i = 0; i < claims_index; i++)
+            _free_claim(&claims[i]);
+    }
+    return result;
+}
+
+oe_result_t oe_deserialize_custom_claims(
+    const uint8_t* claims_buffer,
+    size_t claims_buffer_size,
+    oe_claim_t** claims_out,
+    size_t* claims_length_out)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    oe_custom_claims_header_t* claims_header = NULL;
+    oe_claim_t* claims = NULL;
+    uint64_t claims_length = 0;
+    uint64_t claims_size = 0;
+
+    if (claims_buffer_size < sizeof(oe_custom_claims_header_t))
+    {
+        OE_RAISE(OE_CONSTRAINT_FAILED);
+    }
+
+    claims_header = (oe_custom_claims_header_t*)(claims_buffer);
+
+    if (claims_header == NULL)
+    {
+        OE_RAISE(OE_INVALID_PARAMETER);
+    }
+
+    if (claims_header->version != OE_CUSTOM_CLAIMS_VERSION)
+    {
+        OE_RAISE(OE_CONSTRAINT_FAILED);
+    }
+
+    claims_length = claims_header->num_claims;
+
+    OE_CHECK(oe_safe_mul_u64(claims_length, sizeof(oe_claim_t), &claims_size));
+
+    claims = (oe_claim_t*)oe_malloc(claims_size);
+    if (claims == NULL)
+        OE_RAISE(OE_OUT_OF_MEMORY);
+
+    OE_CHECK(_fill_array_with_custom_claims(
+        claims_buffer, claims_buffer_size, claims));
+
+    *claims_out = claims;
+    *claims_length_out = claims_header->num_claims;
+    claims = NULL;
+    result = OE_OK;
+
+done:
+    if (claims)
+        oe_free_custom_claims(claims, claims_length);
+    return result;
+}
+
+static oe_result_t _get_claims_size(
+    const oe_claim_t* custom_claims,
+    size_t custom_claims_length,
+    size_t* size)
+{
+    oe_result_t result = OE_UNEXPECTED;
+
+    if (!custom_claims)
+        return OE_OK;
+
+    OE_CHECK(oe_safe_add_u64(*size, sizeof(oe_custom_claims_header_t), size));
+
+    for (size_t i = 0; i < custom_claims_length; i++)
+    {
+        OE_CHECK(
+            oe_safe_add_u64(*size, sizeof(oe_custom_claims_entry_t), size));
+        OE_CHECK(
+            oe_safe_add_u64(*size, oe_strlen(custom_claims[i].name) + 1, size));
+        OE_CHECK(oe_safe_add_u64(*size, custom_claims[i].value_size, size));
+    }
+    result = OE_OK;
+done:
+    return result;
+}
+
+static void _fill_buffer_with_custom_claims(
+    const oe_claim_t* custom_claims,
+    size_t custom_claims_length,
+    uint8_t* claims)
+{
+    // Custom claims structure would be:
+    //  - oe_custom_claims_header_t
+    //  - N claim entries of oe_sgx_custom_claims_entry_t
+    // claims buffer is previously allocated with appropriate size,
+    // copy corresponding values
+    oe_custom_claims_header_t* header = (oe_custom_claims_header_t*)claims;
+    header->version = OE_CUSTOM_CLAIMS_VERSION;
+    header->num_claims = custom_claims ? custom_claims_length : 0;
+    claims += sizeof(oe_custom_claims_header_t);
+
+    if (!custom_claims)
+        return;
+
+    for (size_t i = 0; i < custom_claims_length; i++)
+    {
+        oe_custom_claims_entry_t* entry = (oe_custom_claims_entry_t*)claims;
+        entry->name_size = oe_strlen(custom_claims[i].name) + 1;
+        entry->value_size = custom_claims[i].value_size;
+        oe_memcpy_s(
+            entry->name,
+            entry->name_size,
+            custom_claims[i].name,
+            entry->name_size);
+        oe_memcpy_s(
+            entry->name + entry->name_size,
+            entry->value_size,
+            custom_claims[i].value,
+            entry->value_size);
+        // move to next claim
+        claims += sizeof(*entry) + entry->name_size + entry->value_size;
+    }
+}
+
+oe_result_t oe_serialize_custom_claims(
+    const oe_claim_t* custom_claims,
+    size_t custom_claims_length,
+    uint8_t** claims_out,
+    size_t* claims_size_out)
+{
+    uint8_t* claims = NULL;
+    size_t claims_size = 0;
+    oe_result_t result = OE_UNEXPECTED;
+
+    // Get claims size.
+    OE_CHECK(
+        _get_claims_size(custom_claims, custom_claims_length, &claims_size));
+
+    // Allocate memory and set the claims.
+    claims = (uint8_t*)oe_malloc(claims_size);
+    if (claims == NULL)
+        OE_RAISE(OE_OUT_OF_MEMORY);
+    _fill_buffer_with_custom_claims(
+        custom_claims, custom_claims_length, claims);
+
+    *claims_out = claims;
+    *claims_size_out = claims_size;
+    claims = NULL;
+    result = OE_OK;
+
+done:
+    if (claims != NULL)
+        oe_free(claims);
+    return result;
+}
--- a/include/openenclave/attestation/custom_claims.h
+++ b/include/openenclave/attestation/custom_claims.h
@@ -1,0 +1,79 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#ifndef _OE_CUSTOM_CLAIMS
+#define _OE_CUSTOM_CLAIMS
+
+#include <openenclave/bits/evidence.h>
+#include <openenclave/bits/result.h>
+
+OE_EXTERNC_BEGIN
+#define OE_CUSTOM_CLAIMS_VERSION 1
+
+/**
+ * oe_free_serialized_custom_claims
+ *
+ * Free buffer of serialized custom claims.
+ *
+ * @param[in] custom_claims_buffer Serialized Custom claims
+ * buffer to free.
+ * @retval OE_OK on success.
+ */
+oe_result_t oe_free_serialized_custom_claims(uint8_t* custom_claims_buffer);
+
+/**
+ * oe_free_custom_claims
+ *
+ * Free list of custom claims.
+ *
+ * @param[in] custom_claims Custom claims array to free.
+ * @param[in] custom_claims_length Length of custom_claims.
+ * @retval OE_OK on success.
+ */
+oe_result_t oe_free_custom_claims(
+    oe_claim_t* custom_claims,
+    size_t custom_claims_length);
+
+/**
+ * oe_serialize_custom_claims
+ *
+ * Serializes a list of custom claims.
+ *
+ * @param[in] custom_claims Custom claims to serialize.
+ * @param[in] custom_claims_length Length of custom_claims.
+ * @param[out] claims_out Pointer to the address of a dynamically
+ * allocated buffer holding the serialized custom claims.
+ * @param[out] claims_size_out Size of the serialized custom claims.
+ * @retval OE_OK on success.
+ * @retval OE_INVALID_PARAMETER At least one parameter is invalid.
+ * @retval An appropriate error code on failure.
+ */
+oe_result_t oe_serialize_custom_claims(
+    const oe_claim_t* custom_claims,
+    size_t custom_claims_length,
+    uint8_t** claims_out,
+    size_t* claims_size_out);
+
+/**
+ * oe_deserialize_custom_claims
+ *
+ * Deserialize custom claim buffer
+ *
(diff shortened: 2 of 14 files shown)


tests/attestation_plugin/plugin/tests.c
```diff
@@ -83,9 +86,7 @@ static void _test_and_register_verifier()
         oe_register_verifier_plugin(&mock_verifier2, NULL, 0) ==
         OE_ALREADY_EXISTS);
 }
-
```
yentsanglee, line 86: Don't remove the line space between functions.

tests/attestation_plugin/plugin/tests.c
```diff
@@ -308,6 +304,62 @@ static void _test_verify_evidence_fail()
...
+    printf("====== _test_custom_claims_serde call oe_claims_serialize\n");
+    OE_TEST(
+        oe_serialize_custom_claims(
+            custom_claims, 2, &claims_out, &claims_size_out) == OE_OK);
+
+    printf("====== _test_custom_clims_serde call oe_claims_deserialize\n");
+    OE_TEST(
+        oe_deserialize_custom_claims(
+            claims_out, claims_size_out, &claims1, &claims_len) == OE_OK);
+
```
yentsanglee, line 331: Verify claims_len == 2

tests/attestation_plugin/plugin/tests.c
```diff
@@ -330,14 +382,15 @@ void test_runtime()
...
+    _test_custom_claims_serde();
+
     // Test unregister functions
     _test_and_unregister_attester();
     _test_and_unregister_verifier();
 #else
     printf("====== running test_runtime, host side, only verifier tests\n");
     // Test register functions.
     _test_and_register_verifier();
-
```
yentsanglee, line 340: Don't remove line space.

shnwc:
Agree with @yentsanglee that `oe_serialize_custom_claims()` and `oe_deserialize_custom_claims()` are public functions and not SGX specific.

I think it would be good to put their declaration and implementation in dedicated filess, with the implementation in e.g. `common/custom_claims.c`, and declaration in e.g. `include/openenclave/attestation/custom_claims.h`. Both functions can be supported in both the enclave and host sides.

shants:
Tests for the serialization/de-serialization were added in tests/attestation_plugin/plugin/tests.c, given that attestation_plugin take long time in execution,  should the new test case be added in separate file / directory ?
Current time of execution of tests in attestation_plugin ~ 85 sec ( with and without new test )

include/openenclave/internal/sgx/plugin.h
```diff
@@ -77,6 +78,65 @@ oe_result_t oe_sgx_hash_custom_claims(
...
+/**
+ * oe_serialize_custom_claims
+ *
+ * Serializes a list of custom claims.
+ *
+ * @param [in] custom_claims Custom claims to serialize.
+ * @param [in] custom_claims_length Length of custom_claims.
+ * @param [out] claims_out Pointer to the address of a dynamically
+ * allocated buffer holding the serialized custom claims.
+ * @param [out] claims_size_out Size of the serialized custom claims.
```
dthaler, line 109: nit: Remove space between "param" and "["

include/openenclave/internal/sgx/plugin.h
```diff
@@ -77,6 +78,65 @@ oe_result_t oe_sgx_hash_custom_claims(
...
+/**
+ * oe_deserialize_custom_claims
+ *
+ * Deserialize custom claim buffer
+ *
+ * @param[in] claims_buffer Pointer to the serialized custom claims buffer.
+ * @param[in] claims_buffer_size Size of the serialized custom claims buffer
+ * buffer.
+ * @param[out] claims_out Pointer to the address of a dynamically allocated
+ * buffer holding the list of custom claims).
```
dthaler, line 129:
```suggestion
 * buffer holding the list of custom claims.
```

common/sgx/verifier.c
```diff
@@ -870,3 +870,94 @@ oe_result_t oe_verifier_shutdown(void)
...
+        (oe_sgx_plugin_claims_header_t*)claims_buf;
+    size_t claims_index = 0;
+
+    claims_buf += sizeof(*header);
+    claims_buf_size -= sizeof(*header);
+    for (uint64_t i = 0; i < header->num_claims; i++)
+    {
+        oe_sgx_plugin_claims_entry_t* entry =
+            (oe_sgx_plugin_claims_entry_t*)claims_buf;
+        uint64_t size;
```
qiucwang, line 890:
```suggestion
        uint64_t size = 0;
```

qiucwang, line 890:
```suggestion
        uint64_t size = 0;
```

yentsanglee:
> Tests for the serialization/de-serialization were added in tests/attestation_plugin/plugin/tests.c, given that attestation_plugin take long time in execution, should the new test case be added in separate file / directory ?
> Current time of execution of tests in attestation_plugin ~ 85 sec ( with and without new test )

Please add a new test.

shnwc:
> Also please update the PR title to say SGX

The helper functions are TEE agnostic, can be used for all evidence formats, SGX or non-SGX.

shnwc:
> > Tests for the serialization/de-serialization were added in tests/attestation_plugin/plugin/tests.c, given that attestation_plugin take long time in execution, should the new test case be added in separate file / directory ?
> > Current time of execution of tests in attestation_plugin ~ 85 sec ( with and without new test )
> 
> Please add a new test.

Yes a separate new test would be good.

dthaler:
> > Also please update the PR title to say SGX
> 
> @dthaler The helper functions are TEE agnostic, can be used for all evidence formats, SGX or non-SGX. I don't think "SGX" should be added to the PR title or to the labels.

No they are SGX specific.  That is, they appear in include/openenclave/internal/sgx/plugin.h and that header is SGX-only.

shnwc:
> > > Also please update the PR title to say SGX
> > 
> > 
> > @dthaler The helper functions are TEE agnostic, can be used for all evidence formats, SGX or non-SGX. I don't think "SGX" should be added to the PR title or to the labels.
> 
> No they are SGX specific. That is, they appear in include/openenclave/internal/sgx/plugin.h and that header is SGX-only, and they are implemented in SGX-only .c files.

See my other [comment](https://github.com/openenclave/openenclave/pull/3358#issuecomment-670667554). Yes they are SGX-specific as currently implemented. But We should make the API and implementation TEE-agnostic.

dthaler:
> We should make the API and implementation TEE-agnostic.

Ok then assuming you mean "in this PR", I will re-review this PR again once that is done and will mark this as Changes Requested in the meantime.

tests/attestation_plugin/plugin/tests.c
```diff
@@ -308,6 +304,62 @@ static void _test_verify_evidence_fail()
     OE_TEST(oe_free_endorsements(endorsements) == OE_OK);
 }
 
+static void _test_custom_claims_serde()
```
jxyang, line 307: What's `serde`? Please spell out.

tests/attestation_plugin/plugin/tests.c
```diff
@@ -308,6 +304,62 @@ static void _test_verify_evidence_fail()
...
+                                    .value_size = sizeof(CLAIM2_VALUE)}};
+
+    printf("====== _test_custom_claims_serde call oe_claims_serialize\n");
+    OE_TEST(
+        oe_serialize_custom_claims(
+            custom_claims, 2, &claims_out, &claims_size_out) == OE_OK);
+
+    printf("====== _test_custom_clims_serde call oe_claims_deserialize\n");
+    OE_TEST(
+        oe_deserialize_custom_claims(
```
jxyang, line 329: I'd like to see a more realistic test: one function calls oe_get_evidence with a list of custom claims, and another function calls oe_verify_evidence and verifies the deserialized custom claims.

tests/CMakeLists.txt
```diff
@@ -92,6 +92,7 @@ if (UNIX
     add_subdirectory(enclaveparam)
     add_subdirectory(file)
     add_subdirectory(getenclave)
+    add_subdirectory(helper)
```
mingweishih, line 95: Can you rename the test to be more specific? "helper" is not intuitive to me what the test intends to do.

include/openenclave/internal/sgx/plugin.h
```diff
@@ -77,6 +78,65 @@ oe_result_t oe_sgx_hash_custom_claims(
...
+typedef struct _oe_sgx_plugin_claims_entry
+{
+    uint64_t name_size;
+    uint64_t value_size;
+    uint8_t name[];
+    // name_size bytes follow.
+    // value_size_bytes follow.
+} oe_sgx_plugin_claims_entry_t;
+
+/**
```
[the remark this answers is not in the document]
shants: added include/openenclave/attestation/custom_claims.h, /common/custom_claims.c and moved def and decl.

tests/attestation_plugin/plugin/tests.c
shants: added OE_TEST(claims_len == 2)

tests/attestation_plugin/plugin/tests.c
shants: renamed the test to _test_custom_claims_seriaize_deserialize()

tests/CMakeLists.txt
shants: @mingweishih I thought serialization/deserialization to be helper methods hence the name, I can think of  utils, common ( the files added are common/custom_claims.c ) or custom_claims. Any suggestions ?

common/custom_claims.c
```diff
@@ -0,0 +1,219 @@
...
+    for (size_t i = 0; i < claims_length; i++)
+        _free_claim(&claims[i]);
+    oe_free(claims);
+    return OE_OK;
+}
+
+static oe_result_t _add_claim(
+    oe_claim_t* claim,
+    const void* name,
+    size_t name_size, // Must cover the '\0' at end of string
```
dthaler, line 29: Does it need to be a string?  if so (and line 155 implies it has to be), then why isn't name a const char* and delete the name_size arg?
If it does not need to be a string (and I think there's a good argument for why requiring a string would be problematic for constrained devices, but maybe the answer for them is don't use these helper functions, write your own), then the comment is misleading and oe_claim_t would need to have a name_size field.  Point is that the type of the name argument here, and the type of the name field in oe_claim_t should be the same.  Currently they're different.

common/custom_claims.c
```diff
@@ -0,0 +1,219 @@
...
+        return OE_OUT_OF_MEMORY;
+    }
+    memcpy(claim->value, value, value_size);
+    claim->value_size = value_size;
+
+    return OE_OK;
+}
+
+static oe_result_t _fill_with_custom_claims(
+    const uint8_t* claims_buf,
```
dthaler, line 55: style: Avoid abbreviations, so expand "buf" to "buffer"

include/openenclave/attestation/custom_claims.h
```diff
@@ -0,0 +1,74 @@
...
+#define OE_PLUGIN_CLAIMS_VERSION 1
+
+typedef struct _oe_plugin_claims_header
+{
+    uint64_t version;
+    uint64_t num_claims;
+} oe_plugin_claims_header_t;
+
+/**
+ * Serialzied entry for custom claims. Each entry will have the name and value
```
dthaler, line 20:
```suggestion
 * Serialized entry for custom claims. Each entry will have the name and value
```

include/openenclave/attestation/custom_claims.h
```diff
@@ -0,0 +1,74 @@
...
+#ifndef _OE_CUSTOM_CLAIMS
+#define _OE_CUSTOM_CLAIMS
+
+#include <openenclave/bits/evidence.h>
+#include <openenclave/bits/result.h>
+
+OE_EXTERNC_BEGIN
+#define OE_PLUGIN_CLAIMS_VERSION 1
+
+typedef struct _oe_plugin_claims_header
```
dthaler, line 13: If I understand this right, the term "custom" should be in the type name

include/openenclave/internal/sgx/plugin.h
```diff
@@ -3,11 +3,12 @@
 
 #ifndef _OE_INTERNAL_SGX_PLUGIN
 #define _OE_INTERNAL_SGX_PLUGIN
-
+#include <openenclave/bits/evidence.h>
 #include <openenclave/bits/report.h>
 #include <openenclave/internal/crypto/sha.h>
 
 OE_EXTERNC_BEGIN
+//#define OE_SGX_PLUGIN_CLAIMS_VERSION 1
```
dthaler, line 11: Should not have commented out defines.  Just delete.

tests/CMakeLists.txt
dthaler: custom_claims is ok with me

tests/helper/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+add_subdirectory(host)
+
+if (BUILD_ENCLAVES)
+  add_subdirectory(enc)
+endif ()
+
+add_enclave_test(tests/helper helper_host helper_enc)
```
yentsanglee, line 10: I'd name the test case 'custom_claims'.

tests/attestation_plugin/plugin/tests.c
yentsanglee: I think the test case should focus on testing serialization/de-serialization. What happens inside oe_get_evidence() and oe_verify_evidence() is not to be covered in this test case.

tests/CMakeLists.txt
shants: renamed to custom_claims

include/openenclave/attestation/custom_claims.h
shants: changed to oe_custom_claims_header

common/custom_claims.c
shants: This is called from the deserialize function, we pass the name, name_size, value and value_size and then fill the oe_claim_t array ( line 36, 41)

tests/attestation_plugin/plugin/tests.c
jxyang: I disagree. The sole purpose of these helper functions is to pass/parse the buffer to/from oe_get_evidence/oe_verify_evidence. We should verify the helper functions end-to-end.

yentsanglee: The role of custom claims in the process of evidence validation:
[custom claims] -> (1. serialize) -> [flat custom claims buffer] -> (2. oe_get_evidence) -> [evidence] -> (3. oe_verify_evidence) -> [flag custom claim buffer] -> (4. de-serialize) -> [custom claims]

The test cases in this PR should focus on input/output of (1. serialize) and (4. de-serialize).

Issue #3389 will cover the test cases for input/output of (2. oe_get_evidence) and (3. oe_verify_evidence).

common/custom_claims.c
```diff
@@ -0,0 +1,222 @@
...
+    oe_claim_t* claims = NULL;
+    uint64_t claims_length = 0;
+    uint64_t claims_size = 0;
+
+    claims_header = (oe_custom_claims_header_t*)(claims_buffer);
+    if (claims_header == NULL)
+    {
+        OE_RAISE(OE_INVALID_PARAMETER);
+    }
+
```
shnwc, line 123: Should check that claim buffer size is at least sufficient for the header, and the version is expected.

include/openenclave/internal/sgx/plugin.h
```diff
@@ -3,7 +3,7 @@
 
 #ifndef _OE_INTERNAL_SGX_PLUGIN
 #define _OE_INTERNAL_SGX_PLUGIN
-
+#include <openenclave/bits/evidence.h>
```
shnwc, line 6: Wonder why this header file should be included?

tests/custom_claims/host/host.c
```diff
@@ -0,0 +1,47 @@
...
+
+    const uint32_t flags = oe_get_create_flags();
+
+    if ((result = oe_create_custom_claims_enclave(
+             argv[1], OE_ENCLAVE_TYPE_SGX, flags, NULL, 0, &enclave)) != OE_OK)
+        oe_put_err("oe_create_enclave(): result=%u", result);
+
+    int return_val;
+
+    result = enc_custom_claims(enclave, &return_val);
```
shnwc, line 33: Should also test host version of the functions, at least for de-serialization which is for verifiers.

common/custom_claims.c
dthaler: Ok, I guess since this encoding is already quite inefficient as noted above, having it use an extra null terminator in the buffer is ok.  Anyone worried about the overhead would never use these helper functions and would instead write their own more efficient ones I guess.

And since the name pointer is into the serialized buffer, it should probably be of type "const uint8_t*" then.

common/custom_claims.c
```diff
@@ -0,0 +1,222 @@
...
+        claim->name = NULL;
+        return OE_OUT_OF_MEMORY;
+    }
+    memcpy(claim->value, value, value_size);
+    claim->value_size = value_size;
+
+    return OE_OK;
+}
+
+static oe_result_t _fill_with_custom_claims(
```
dthaler, line 54: This name confused me at first since I couldn't tell if it was meant to fill the buffer or fill the claims.
Only the "const" in 55 helped.  Suggest: _fill_array_with_custom_claims
```suggestion
static oe_result_t _fill_array_with_custom_claims(
```

common/custom_claims.c
```diff
@@ -0,0 +1,222 @@
...
+    for (size_t i = 0; i < custom_claims_length; i++)
+    {
+        size += sizeof(oe_custom_claims_entry_t);
+        size += oe_strlen(custom_claims[i].name) + 1;
+        size += custom_claims[i].value_size;
+    }
+    return size;
+}
+
+static void _set_claims(
```
dthaler, line 164:
```suggestion
static void _fill_buffer_with_custom_claims(
```

common/custom_claims.c
shants: add the checks

tests/custom_claims/host/host.c
shants: added the tests for host version

include/openenclave/internal/sgx/plugin.h
shants: removed

yentsanglee:
bors try

yentsanglee:
bors retry

common/custom_claims.c
```diff
@@ -0,0 +1,235 @@
...
+
+    for (size_t i = 0; i < claims_length; i++)
+        _free_claim(&claims[i]);
+    oe_free(claims);
+    return OE_OK;
+}
+
+static oe_result_t _add_claim(
+    oe_claim_t* claim,
+    const void* name,
```
dthaler, line 29:
```suggestion
    const uint8_t* name,
```

common/custom_claims.c
```diff
@@ -0,0 +1,235 @@
...
+    size_t name_size, // Must cover the '\0' at end of string
+    const void* value,
+    size_t value_size)
+{
+    if (*((uint8_t*)name + name_size - 1) != '\0')
```
dthaler, line 34:
```suggestion
    if (*((name + name_size - 1) != '\0')
```

common/custom_claims.c
```diff
@@ -0,0 +1,235 @@
...
+    const void* value,
+    size_t value_size)
+{
+    if (*(name + name_size - 1) != '\0')
+        return OE_CONSTRAINT_FAILED;
+
+    claim->name = (char*)oe_malloc(name_size);
+    if (claim->name == NULL)
+        return OE_OUT_OF_MEMORY;
+    memcpy(claim->name, name, name_size);
```
CodeMonkeyLeet, line 40: There was an effort to switch to `oe_memcpy_s` for bounds checked memcpy as a result of static analysis warnings, could you please update these?

common/custom_claims.c
```diff
@@ -0,0 +1,235 @@
...
+    const uint8_t* claims_buffer,
+    size_t claims_buffer_size,
+    oe_claim_t* claims)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    oe_custom_claims_header_t* header =
+        (oe_custom_claims_header_t*)claims_buffer;
+    size_t claims_index = 0;
+
+    claims_buffer += sizeof(*header);
```
CodeMonkeyLeet, line 65: You have a good use of safe math elsewhere in this code, but can you please do a scrub and ensure that all inputs have size bounds either explicitly checked or use safe math as well?

For example, in this case, `claims_buffer` is a caller provided value, and the only validation `oe_deserialize_custom_claims` has done before passing it here is that it is non-null, so this addition itself may not be safe. Similarly, for `claims_buffer_size` and the increment/decrement math in the loop. 

I would encourage you to be extra careful with input validation, because serialization/deserialization code is very commonly exploited for arithmetic overflow and similar issues.

common/custom_claims.c
```diff
@@ -0,0 +1,235 @@
...
+    size_t custom_claims_length)
+{
+    size_t size = sizeof(oe_custom_claims_header_t);
+
+    if (!custom_claims)
+        return size;
+
+    for (size_t i = 0; i < custom_claims_length; i++)
+    {
+        size += sizeof(oe_custom_claims_entry_t);
```
CodeMonkeyLeet, line 169: This is another candidate for safe math. Something you may find helpful is to track in comments any preconditions/postconditions of buffers/sizes being manipulated, particularly for internal functions like this if you're splitting the validation with the caller.

include/openenclave/attestation/custom_claims.h
```diff
@@ -0,0 +1,74 @@
...
+ * @param[in] claims_buffer_size Size of the serialized custom claims buffer
+ * buffer.
+ * @param[out] claims_out Pointer to the address of a dynamically allocated
+ * buffer holding the list of custom claims.
+ * @param[out] claims_length_out The length of the claims_out list.
+ * @retval OE_OK on success.
+ * @retval OE_INVALID_PARAMETER At least one parameter is invalid.
+ * @retval An appropriate error code on failure.
+ */
+oe_result_t oe_deserialize_custom_claims(
```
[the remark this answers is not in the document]
dthaler: Very much agree.  Since you're hiding which allocator is used, we also need to hide which deallocator is used so the caller need not know whether free or something else is the right API.

include/openenclave/attestation/custom_claims.h
```diff
@@ -0,0 +1,74 @@
...
+ * sizes and then the contents of the name and value respectively.
+ */
+typedef struct _oe_custom_claims_entry
+{
+    uint64_t name_size;
+    uint64_t value_size;
+    uint8_t name[];
+    // name_size bytes follow.
+    // value_size_bytes follow.
+} oe_custom_claims_entry_t;
```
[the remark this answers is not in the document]
shants: These headers are being used for ser/de-ser purposes only, moved them to custom_claims.c

include/openenclave/attestation/custom_claims.h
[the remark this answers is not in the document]
shants: Added oe_free_serialized_custom_claims(), oe_free_custom_claims()

common/custom_claims.c
```diff
@@ -0,0 +1,289 @@
...
+    uint64_t value_size;
+    uint8_t name[];
+    // name_size bytes follow.
+    // value_size_bytes follow.
+} oe_custom_claims_entry_t;
+
+oe_result_t oe_free_serialized_custom_claims(uint8_t* custom_claims_buffer)
+{
+    if (!custom_claims_buffer)
+        return OE_OK;
```
dthaler, line 37: These lines should not be needed. oe_free(), like the POSIX free(), ought to be safe to pass a null pointer to.

include/openenclave/attestation/custom_claims.h
```diff
@@ -0,0 +1,79 @@
...
+ * @param[in] custom_claims_buffer Serialized Custom claims
+ * buffer to free.
+ * @retval OE_OK on success.
+ */
+oe_result_t oe_free_serialized_custom_claims(uint8_t* custom_claims_buffer);
+
+/**
+ * oe_free_custom_claims
+ *
+ * Free list of  custom claims.
```
dthaler, line 27:
```suggestion
 * Free list of custom claims.
```

tests/custom_claims/host/host.c
```diff
@@ -0,0 +1,45 @@
...
+    if (argc != 2)
+    {
+        fprintf(stderr, "Usage: %s ENCLAVE_PATH\n", argv[0]);
+        return 1;
+    }
+
+    const uint32_t flags = oe_get_create_flags();
+
+    if ((result = oe_create_custom_claims_enclave(
+             argv[1], OE_ENCLAVE_TYPE_SGX, flags, NULL, 0, &enclave)) != OE_OK)
```
dthaler, line 23: Unless the test is inherently SGX-only, it would be better to specify OE_ENCLAVE_TYPE_AUTO instead of OE_ENCLAVE_TYPE_SGX

tests/custom_claims/test_common/tests.c
```diff
@@ -0,0 +1,62 @@
...
+
+    for (size_t i = 0; i < claims_len; ++i)
+    {
+        printf(
+            "====== _test_custom_claims_seriaize_deserialize claim %s %s\n",
+            claims1[i].name,
+            claims1[i].value);
+    }
+
+    for (size_t i = 0; i < claims_len; ++i)
```
yentsanglee, line 46: Call oe_free_custom_claims() instead.

tests/custom_claims/test_common/tests.c
```diff
@@ -0,0 +1,62 @@
...
+        free(claims1[i].name);
+        free(claims1[i].value);
+    }
+
+    free(claims1);
+
+    claims_len = 0;
+    claims1 = NULL;
+
+    free(claims_out);
```
yentsanglee, line 57: Call oe_free_serialized_custom_claims() instead.

yentsanglee:
bors retry

PR #3359, opened 2020-08-06
Use dynamic symbols for loading enclaves

anakrish:
The following symbols are needed to load an enclave
- _start This is the enclave entry-point that calls __oe_handle_main
- oe_call_link_enclave This is used to force dependency on report/attestation functions
- _enclave_rva This is critical for the enclave to securely figure out the base-address
- _reloc_rva This points to the relocation entries of the enclave. This is critical for the
  enclave to function correctly. Relocations are performed as one of the first actions upon
  enclave initialization.
- _reloc_size This is the size in bytes of _reloc_rva entries.
- _tdata_rva This is the pointer to thread-local-data template.
- _tdata_size This is the size of the thread-local-data template.
- _tdata_align This is the alignment of thread-local-data template.
- _tbss_size This tis the size of zero initialized thread-local-data. (tbss section).
- _tbss_align This is the alignment of the tbss section.

All the above symbols must be visible to correctly load an enclave.

When an enclave is built using `-s` linker flag (as is often the case with release builds)
or if the enclave binary is stripped using the strip command, then these symbols will be
eliminated from the enclave binary and there would be no way to load them.

In order to perform dynamic linking/program loading, necessary symbols are always preserved
in the dynamic symbols section.

To make enclaves loading work without symbols, all of the above symbols must be exported
(earlier we were only exporting _start). Additionally the loader must rely on dynamic symbols.

### DIFF
--- a/enclave/core/sgx/globals.c
+++ b/enclave/core/sgx/globals.c
@@ -7,7 +7,8 @@
 
 /* Note: The variables below are initialized during enclave loading */
 
-extern volatile const oe_sgx_enclave_properties_t oe_enclave_properties_sgx;
+OE_EXPORT extern volatile const oe_sgx_enclave_properties_t
+    oe_enclave_properties_sgx;
 
 /**
  *****************************************************************************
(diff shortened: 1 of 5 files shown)


anakrish:
bors try

anakrish:
bors try

PR #3361, opened 2020-08-07
Add fstat

thomasten:
Added support for the fstat syscall.

### DIFF
--- a/host/linux/syscall.c
+++ b/host/linux/syscall.c
@@ -293,6 +293,26 @@ int oe_syscall_closedir_ocall(uint64_t dirp)
     return closedir((DIR*)dirp);
 }
 
+static void _stat_copy(const struct stat* st, struct oe_stat_t* buf)
+{
+    buf->st_dev = st->st_dev;
+    buf->st_ino = st->st_ino;
+    buf->st_nlink = st->st_nlink;
+    buf->st_mode = st->st_mode;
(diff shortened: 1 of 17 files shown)


host/linux/syscall.c
```diff
@@ -293,6 +293,27 @@ int oe_syscall_closedir_ocall(uint64_t dirp)
     return closedir((DIR*)dirp);
 }
 
+static void _stat_copy(const struct stat* st, struct oe_stat_t* buf)
+{
+    buf->st_dev = st->st_dev;
```
CodeMonkeyLeet, line 298: It's an existing redundancy that got copied and pasted, but could you remove it?
```suggestion
```

host/windows/syscall.c
```diff
@@ -1694,6 +1694,28 @@ int oe_syscall_closedir_ocall(uint64_t dirp)
     return ret;
 }
 
+static void _stat_copy(const struct _stat64* winstat, struct oe_stat_t* buf)
+{
+    // The macro #define st_atime st_atim.tv_sec
+    // provides backward compatibility for older version POSIX. Here we need
```
dthaler, line 1700: Please fix grammar
```suggestion
    // provides backward compatibility for older versions of POSIX. Here we need
```

thomasten: fixed and squashed

PR #3366, opened 2020-08-07
ECALL Implementation: Support global ids

anakrish:
ECALL Implementation: Support global ids
    
    Design doc: https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/TEECallsRefactoring.md
    
    Bring in @mingweishih's implementation of global ids for ECALLs.
    With this change, ECALL dispatching will now work in case two enclaves
    import system ECALLs in different order. Currently, ECALL dispatching would
    fail at runtime in such a scenario.
    
    Changes:
    - Update edger8r submodule
    - Implement functions to compute global ids for ECALLs based on name
    - Update enclave structure and enclave initialization for both SGX and OPTEE
    - Added test to lockdown global ids and local ids.
    

anakrish:
bors try

host/ecallids.c
```diff
@@ -0,0 +1,140 @@
...
+            goto done;
+    }
+
+    if (*global_id >= OE_MAX_ECALLS)
+    {
+        result = OE_OUT_OF_BOUNDS;
+        goto done;
+    }
+
+    OE_CHECK(oe_get_ecall_id_table(enclave, &ecall_id_table));
```
mingweishih, line 89: Consider adding comments here explaining why we need to do this (i.e.., we cannot directly obtain via `enclave->ecall_id_table` because of the `oe_enclave_t` is opaque here).

host/calls.c
```diff
@@ -114,14 +115,21 @@ oe_result_t oe_call_enclave_function_by_table_id(
...
     size_t input_buffer_size,
     void* output_buffer,
     size_t output_buffer_size,
     size_t* output_bytes_written)
 {
-    return oe_call_enclave_function_by_table_id(
+    oe_result_t result = OE_UNEXPECTED;
+    uint64_t function_id = OE_UINT64_MAX;
+
+    OE_CHECK(
```
mingweishih, line 129: I'd like to carry the following comments.

```suggestion
        /*
         * Look up the function id from the per-enclave table based on the
         * global id. The global id is defined as a static variable in the
         * oeedger8r-generated code. The function initializes the global id in
         * the first invocation and uses the cached global id for the subsequent
         * invocations.
         */
         OE_CHECK(
```

host/ecallids.c
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /*  Nothing to add when the table is empty, fall through. */
+    if (!ecall_info_table || !num_ecalls)
+    {
+        result = OE_OK;
+        goto done;
+    }
+
+    OE_CHECK(oe_get_ecall_id_table(enclave, &ecall_id_table));
```
mingweishih, line 120: Consider adding comments here as well.

mingweishih:
bors try

anakrish:
bors try

anakrish:
bors try

host/ecallids.c
```diff
@@ -0,0 +1,140 @@
...
+
+    if (!enclave || !global_id || !name || !id)
+    {
+        result = OE_INVALID_PARAMETER;
+        goto done;
+    }
+
+    if (*global_id == OE_GLOBAL_ECALL_ID_NULL)
+    {
+        if ((result = _get_global_ecall_id_by_name(name, global_id)) != OE_OK)
```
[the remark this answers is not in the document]
anakrish: Done

host/ecallids.c
anakrish: Done.

host/ecallids.c
anakrish: Done.

anakrish:
bors try

host/ecallids.c
```diff
@@ -0,0 +1,158 @@
...
+
+    /* Fetch ecall table using platform agnostic API */
+    OE_CHECK(oe_get_ecall_id_table(enclave, &ecall_id_table));
+
+    if (oe_mutex_lock(&_lock) != 0)
+        OE_RAISE(OE_FAILURE);
+    locked = true;
+
+    /* Iterate through the ecalls and assign ids. Doing this during enclave
+     * create avoid having the perform global id locks when making ecalls. */
```
dthaler, line 137:
```suggestion
     * creation avoids having to perform global id locks when making ecalls. */
```

host/ecallids.c
```diff
@@ -0,0 +1,158 @@
...
+    uint32_t i;
+    for (i = 0; i < num_ecalls; i++)
+    {
+        uint64_t global_id = OE_GLOBAL_ECALL_ID_NULL;
+        const char* name = ecall_info_table[i].name;
+        uint64_t local_id = i;
+
+        /* Assign a proper global id based on the global __ecall_table. */
```
dthaler, line 145:
```suggestion
        /* Assign a proper global id based on the global _ecall_table. */
```

host/ecallids.h
```diff
@@ -0,0 +1,60 @@
...
+// Licensed under the MIT License.
+
+#ifndef _OE_HOST_ECALLIDS_H
+#define _OE_HOST_ECALLIDS_H
+
+#include <openenclave/edger8r/host.h>
+#include <openenclave/host.h>
+
+#define OE_ECALL_ID_NULL OE_UINT64_MAX
+#define OE_MAX_ECALLS 256
```
dthaler, line 11: Why do you need a fixed maximum instead of being able to realloc dynamically?  256 seems like it runs the danger of a regression if there are complicated EDLs out there.

host/ecallids.c
```diff
@@ -0,0 +1,158 @@
...
+    {
+        if (strcmp(_ecall_table[i], name) == 0)
+        {
+            *global_id = i;
+            break;
+        }
+    }
+
+    /* If the name is not found, adding it to the table. */
+    if (*global_id == OE_GLOBAL_ECALL_ID_NULL)
```
dthaler, line 35: I think *global_id should be initialized to OE_GLOBAL_ECALL_ID_NULL around line 23

host/ecallids.c
```diff
@@ -0,0 +1,158 @@
...
+    {
+        if (_ecall_table_size >= OE_MAX_ECALLS)
+        {
+            *global_id = OE_MAX_ECALLS;
```
dthaler, line 39: seems safer to pass back OE_GLOBAL_ECALL_ID_NULL rather than OE_MAX_ECALLS

host/ecallids.c
```diff
@@ -0,0 +1,158 @@
...
+        /* The global_id variable is generated by edger8r, one global id
+         *  variable per ecall. The same variable will be reused every time that
+         * ecall is made. Therefore, for an ecall, the following code will be
+         * executed only once.
+         */
+        OE_CHECK(oe_get_global_id(name, global_id));
+    }
+
+    /* Look up function id based on global id. */
+    if (*global_id >= OE_MAX_ECALLS)
```
dthaler, line 101: per my earlier comment, seems safer to return and check against OE_GLOBAL_ECALL_ID_NULL here

tests/ecall_conflict/common.edl
```diff
@@ -0,0 +1,38 @@
...
+
+enclave {
+
+  // Define a struct to ensure that multiple definition error does not occur.
+  struct MyStruct {
+    int x;
+    int64_t y;
+  };
+
+  // Define an union to ensure that multiple definition error does not occur.
```
dthaler, line 12:
```suggestion
  // Define a union to ensure that multiple definition error does not occur.
```

tests/ecall_conflict/host/host.cpp
```diff
@@ -0,0 +1,236 @@
...
+    OE_TEST(global_id == 1);
+    OE_TEST(oe_get_global_id("enc_ecall2", &global_id) == OE_OK);
+    OE_TEST(global_id == 2);
+    OE_TEST(oe_get_global_id("enc_ecall3", &global_id) == OE_OK);
+    OE_TEST(global_id == 3);
+    OE_TEST(oe_get_global_id("enc_local_ecall2", &global_id) == OE_OK);
+    OE_TEST(global_id == 9);
+
+    /*
+     * Look up by gloal id. The name should not be NULL.
```
dthaler, line 158:
```suggestion
     * Look up by global id. The name should not be NULL.
```

host/ecallids.c
mingweishih: global_id is defined as static variable and initialalized in the oeedger8r generated code.

host/ecallids.c
mingweishih: ID_NULL indicates that the global_id has not been initialalized yet while ID_MAX indicates the global_id has been initialalized but failed. In this case, the corresponding ecall will always fail. We just reuse the ID_MAX as error code. Alternatively, we can introduce something like ID_ERROR.

dthaler: I don't see why we need two different values.
I also don't think having any hard coded max.

host/ecallids.c
dthaler: I think it's bad programming style (bug prone) to have an IN/OUT parameter when an OUT param will do and is more appropriate.

mingweishih: The reason for this is that we use the same call path that takes the global id as input. The goal is setting the global id once (non constant time lookup), and use the cached value for the subsequent invocations (constant time lookup). Not sure if OUT param can achieve this goal.

host/ecallids.c
anakrish: N/A

host/ecallids.h
anakrish: Good point. I have removed the maximum limit, and used realloc instead.

host/ecallids.c
anakrish: In the latest version, the function returns `OE_NOT_FOUND` instead of `OE_OK`. The is more accurate since no implementation of the given ecall was found for the given enclave.

host/ecallids.c
anakrish: In the latest version, the check reads better:
`if (*global_id >= ecall_id_table_size)`

anakrish:
bors try

host/ecallids.c
anakrish: Yes, as @mingweishih described, global_id is an in/out parameter so as to cache the result of the name to global identifier lookup.

For each ecall wrapper function in the host, edger8r generates a static `global_id` variable:
```c
oe_result_t ecall_foo_wrapper(...)
{
   static uint64_t global_id = OE_GLOBAL_ECALL_ID_NULL; // for caching the look up
   ...
   OE_CHECK(oe_get_ecall_ids(enclave, "foo", &global_id, &local_id); // look up global id from name and cache it.
   ...
   dispatch-ecall(enclave, local_id, args);
}
```

Alternatively, we could do this:
edger8r generates an array of global_id caching variables per enclave (user edl):

```
static uint64_t _global_ids[NUM_ECALLS_IN_ENCLAVE]; // 1 element per ecall.
```

`oe_register_ecalls` (which is called during enclave construction) is passed this array (as an out parameter) and it initializes it.

```
oe_result_t oe_register_ecalls(
    oe_enclave_t* enclave,
    const oe_ecall_info_t* ecall_info_table,
    uint32_t num_ecalls,
    uint64_t* global_ids) // out parameter
{
   ...
   for (uint32_t i=0; i < num_ecalls; ++i)
       oe_get_ecall_id(ecall_info_table[i].name, &global_ids[i]); // populate cache
   ...
}
```
Since global id has already been populated during `oe_register_ecalls`, ecalls can be dispatched as:
```c
oe_result_t ecall_foo_wrapper(...)
{
   static uint64_t global_id = OE_GLOBAL_ECALL_ID_NULL; // for caching the look up
   ...
   OE_CHECK(oe_get_ecall_function_id(
                  enclave, 
                  "foo", 
                  _global_id[fcn_idx], // pass in the global id for ecall to get the local id
                  &local_id); 
   ...
   dispatch-ecall(enclave, local_id, args);
}
```

anakrish: @dthaler @mingweishih 
Above, I've proposed an alternate strategy of initializing global ids.
I'm fine with the current approach in this PR or the new proposal.
Let me know what you think.

host/ecallids.c
```diff
@@ -0,0 +1,211 @@
...
+
+    /* Fill the ecall id table */
+    for (uint32_t i = 0; i < num_ecalls; i++)
+    {
+        uint64_t global_id = OE_GLOBAL_ECALL_ID_NULL;
+        const char* name = ecall_info_table[i].name;
+        uint64_t local_id = i;
+
+        /* Assign a proper global id based on the global _ecall_table. */
+        if ((_get_global_id(name, &global_id)) != OE_OK)
```
mingweishih, line 195: This seems will never be the case that returns value other than `OE_OK` now given that there is the same loop above. Just for the sake of checking, we can use `OE_CHECK` instead.

host/optee/linux/enclave.c
```diff
@@ -524,6 +569,9 @@ oe_result_t oe_terminate_enclave(oe_enclave_t* enclave)
     /* Finalize the context against OP-TEE */
     TEEC_FinalizeContext(&enclave->ctx);
 
+    /* Destroy the ecall id table */
+    free(enclave->ecall_id_table);
```
mingweishih, line 573: Should we check if pointer is NULL? Not sure if that's even possible tho.

host/sgx/create.c
```diff
@@ -1061,6 +1101,9 @@ oe_result_t oe_terminate_enclave(oe_enclave_t* enclave)
         free(enclave->debug_enclave);
     }
 
+    /* Destroy the ecall id table */
+    free(enclave->ecall_id_table);
```
mingweishih, line 1105: Should we check if pointer is NULL? Not sure if that's even possible tho.

host/ecallids.c
mingweishih: @anakrish Do you think there is other benefit of adopting the alternative approach (maybe related to sharing tables among the enclaves with the same instances)? I'm just a bit concerned because doing so requires changes to oeedger8r.

host/ecallids.c
```diff
@@ -0,0 +1,211 @@
...
+    oe_result_t result = OE_UNEXPECTED;
+    oe_ecall_id_t* ecall_id_table = NULL;
+    uint64_t max_global_id = 0;
+    uint64_t ecall_id_table_size = 0;
+    bool locked = false;
+
+    /* Validate parameters */
+    if (!enclave || !ecall_info_table || !num_ecalls)
+    {
+        result = OE_INVALID_PARAMETER;
```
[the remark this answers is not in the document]
anakrish: Fixed.

host/ecallids.c
anakrish: Fixed.

host/optee/linux/enclave.c
anakrish: free(NULL) is valid, and a noop. I have added an if check to be consistent with other uses of free.

host/sgx/create.c
anakrish: Fixed.

host/ecallids.c
anakrish: @mingweishih Yes, the alternative approach localizes all initialization to `oe_register_ecalls` function, and in future could help with sharing tables between enclaves created from the same edl/binary.

However, given the scope of the current PR, sharing tables is not a must have, and I'd be OK with either approach.

The current implementation does have the advantage that it doesn't need any edger8r changes.

mingweishih: Thanks for the clarification. In this case, I will prefer the current approach given the concern of IN/OUT would go away when we implement the table sharing.

host/optee/linux/enclave.c
dthaler: Should not have an if check, it just clutters up the code for no value, and (unless there is cross-function optimization) results in increased binary size and instruction branching.

host/optee/linux/enclave.c
```diff
@@ -524,6 +569,10 @@ oe_result_t oe_terminate_enclave(oe_enclave_t* enclave)
     /* Finalize the context against OP-TEE */
     TEEC_FinalizeContext(&enclave->ctx);
 
+    /* Destroy the ecall id table */
+    if (enclave->ecall_id_table)
```
dthaler, line 573: Remove if check

host/sgx/create.c
```diff
@@ -1061,6 +1101,10 @@ oe_result_t oe_terminate_enclave(oe_enclave_t* enclave)
         free(enclave->debug_enclave);
     }
 
+    /* Destroy the ecall id table */
+    if (enclave->ecall_id_tabl)
```
dthaler, line 1105: remove useless if checks

host/optee/linux/enclave.c
mingweishih: I think it's worth updating the our code base because we use different patterns in many places.

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try-

anakrish:
bors try

host/optee/linux/enclave.c
anakrish: I'm going to leave the check in since that is the majority coding pattern in use in the codebase.

host/sgx/create.c
anakrish: Same as above.

anakrish:
bors try-

anakrish:
bors try

host/ecallids.c
```diff
@@ -0,0 +1,206 @@
...
+    uint64_t max_global_id = 0;
+    uint64_t ecall_id_table_size = 0;
+    bool locked = false;
+
+    /* Validate parameters */
+    if (!enclave || !ecall_info_table || !num_ecalls)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+    if (oe_mutex_lock(&_lock) != 0)
+        OE_RAISE(OE_FAILURE);
```
CodeMonkeyLeet, line 159: Avoid using the uninformative `OE_FAILURE` where possible. `oe_mutex_lock` returns an `oe_result_t` so we should return the underlying error here. i.e. you could `OE_CHECK(oe_mutex_lock(&_lock));` here.

Also in `oe_get_global_id`

anakrish:
bors try

host/CMakeLists.txt
```diff
@@ -318,6 +318,7 @@ list(
   ../common/kdf.c
   ../common/argv.c
   asym_keys.c
+  ecallids.c
```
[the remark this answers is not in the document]
anakrish: Done.

host/ecallids.c
anakrish: The host side threading APIs return int.
https://github.com/openenclave/openenclave/blob/7060b014b6a7af86eb6329ca49e07296dda9fe3e/host/hostthread.h#L143

Created issue for the inconsistency : #3486

anakrish:
bors try-

anakrish:
bors try

PR #3370, opened 2020-08-10
Updating oecert to take a custom data file.

jazzybluesea:
Signed-off-by: Sergio Wong <sergio.wong@gmail.com>

### DIFF
--- a/tests/tools/oecert/enc/enc.cpp
+++ b/tests/tools/oecert/enc/enc.cpp
@@ -8,6 +8,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include "crypto.h"
 #include "oecert_t.h"
 
 // This is the identity validation callback. A TLS connecting party (client or
@@ -109,6 +110,37 @@ done:
 
     oe_free_attestation_certificate(output_cert);
 
+    return result;
+}
+
+oe_result_t get_remote_report_with_custom_data(
+    uint8_t* custom_data,
+    size_t custom_data_size,
+    uint8_t** report_buffer,
+    size_t* report_buffer_size)
+{
+    oe_result_t result = OE_FAILURE;
+
+    uint8_t sha256[32];
+
+    init_mbedtls();
+
+    if (Sha256(custom_data, custom_data_size, sha256) != 0)
+    {
+        goto done;
+    }
+
+    result = oe_get_report(
+        OE_REPORT_FLAGS_REMOTE_ATTESTATION,
+        sha256,
+        sizeof(sha256),
+        NULL,
+        0,
+        report_buffer,
+        report_buffer_size);
+
+done:
+    cleanup_mbedtls();
     return result;
 }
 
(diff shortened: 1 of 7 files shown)


tests/tools/oecert/host/host.cpp
```diff
@@ -147,17 +150,58 @@ static oe_result_t _gen_cert(
...
+        fclose(fp);
+
+        result = get_remote_report_with_custom_data(
+            enclave,
+            &ecall_result,
+            buffer,
+            size,
+            (uint8_t**)&remote_report,
+            &report_size);
+
```
yentsanglee, line 198: buffer needs to be freed.

tests/tools/oecert/enc/enc.cpp
```diff
@@ -112,6 +113,37 @@ oe_result_t get_tls_cert_signed_with_key(
...
+    uint8_t** report_buffer,
+    size_t* report_buffer_size)
+{
+    oe_result_t result = OE_FAILURE;
+
+    uint8_t sha256[32];
+
+    init_mbedtls();
+
+    if (Sha256(custom_data, custom_data_size, sha256) != 0)
```
yentsanglee, line 128: Hash of custom_data is embedded in the report but the custom_data itself is not carried over. What is the user scenario? Cannot the custom_data_file itself be the hash value?

tests/tools/oecert/oecert.edl
```diff
@@ -19,5 +19,11 @@ enclave {
             size_t public_key_size,
             [out] unsigned char **data, 
             [out] size_t *data_size);
+
+        public oe_result_t get_remote_report_with_custom_data(
```
dthaler, line 23: the term "report" is legacy, is this really needed or can we use "evidence" instead?

tests/tools/oecert/host/host.cpp
jazzybluesea: Lines 308-309 will free remote_report?

tests/tools/oecert/enc/enc.cpp
jazzybluesea: I wanted to generate a report with the sha256 of the custom data.  The custom data plus the report is then sent in the body for request to MAA.  I just needed a tool to do this for me.

tests/tools/oecert/enc/enc.cpp
```diff
@@ -112,6 +113,37 @@ oe_result_t get_tls_cert_signed_with_key(
...
+    {
+        goto done;
+    }
+
+    result = oe_get_report(
```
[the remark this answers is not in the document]
jazzybluesea: I see what you are saying, maybe for a different PR, my goals was to update the tool to support what I needed. A lot of the existing code and samples will need to migrated to use the new API...

tests/tools/oecert/enc/enc.cpp
yentsanglee: MAA takes evidence type, evidence, and enclave held data. In your case, evidence type is oe_report and enclave help data is the hash of a public key. So your custom_data is the public key, right?

jazzybluesea:
Closing for now.

PR #3373, opened 2020-08-11
Add CCC Technical Charter

dthaler:
The policy for all CCC projects
(found at
https://github.com/confidential-computing/governance/blob/master/project-progression-policy.md)
says under Project Acceptance Process:
"The proposal document will be finalized and a project charter prepared.
This charter document must be included in the project's main
repository."

This PR adds the approved charter in the project's main repository as
required.  I will note that the charter is not correct with respect to
OE governance, but that requires a CCC process to amend.  In the
meantime, we need to include this as is to be compliant.  (Today was
the first time I saw this charter, future charters will be visible to
the TAC.)


anakrish:
I second Simon's suggestion of using .md format instead of .pdf.

dthaler:
> I second Simon's suggestion of using .md format instead of .pdf.

I will ask in the CCC board meeting if this is ok.  (I think it should be but want CCC agreement that it's ok to put a format that is different from the one the board approved, as long as the content is the same.)

achamayou (approved):

PR #3381, opened 2020-08-12
Remove use of TD_FROM_TCS_BYTE_OFFSET

anakrish:
TD_FROM_TCS_BYTE_OFFSET will not valid thread-local data template can be of arbitrary size.

The debugger can figure out the enclave base address from tcs, and then figure out td_t
from enclave base address and tcs.

SSA (State Save Area) is added immediately after TCS.
Hence:
   SSA address = tcs address + 4096
SSA is also defined by adding tcs.ossa (SSA offset) to enclave base address.
Hence:
   SSA address = enclave base address + tcs.ossa

Using this, enclave base address can be computed:
   enclave base address + tcs.ossa = tcs address + 4096
   enclave base address = tcs address + 4096 - tcs.ossa

Upon entry to enclave, both GS and FS registers point to td_t
Since FS register can be changed by and enclave (e.g: SGX LKL), GS register is used.
Hence:
    td_t address = GS register

The value of GS register is defined by adding tcs.gsbase to enclave base address.
Hence:
    td_t address = enclave base address + tcs.gsbase
                 = tca address + 4096 - tcs.ossa + tcs.gsbase

The above formula does not use hard-coded td_t offset, and will work with arbitrary sized thread-local data.

The same strategy is used by the ptrace library in Intel SGX SDK.


anakrish:
bors try

debugger/ptraceLib/enclave_context.c
```diff
@@ -224,8 +223,25 @@ static int _get_enclave_thread_current_ssa_info(
...
+    //    encave-base-address + tcs.ossa = tcs_addr + 4096
+    //    enclave-base-address = tcs_addr + 4096 - tcs.ossa
+    enclave_base_address = (uint8_t*)tcs_addr + 4096 - tcs.ossa;
+
+    // GS register will point to td_t. GS register value can be computed by
+    // adding tcs.gsbase to enclave base address.
+    // Note: FS register will also point to td_t upon enclave entry, but
+    // enclaves can modify FS register to implement their own threading
+    // libraries.
+    td_t_addr = (uint8_t*)enclave_base_address + tcs.gsbase;
```
vtikoo, line 241: Is `tcs.gsbase` an offset relative of enclave_base_adress/TCS page?

anakrish: Yes. It is the offset of GS register value from the enclave base address.

debugger/ptraceLib/enclave_context.c
```diff
@@ -224,8 +223,25 @@ static int _get_enclave_thread_current_ssa_info(
...
 
+    // SSA is assigned to the page immediately after the tcs page.
+    // SSA address is also derived from adding tcs.ossa to enclave base address.
+    // See: host/sgx/create.c
+    // See:
+    // https://github.com/intel/linux-sgx/blob/62b116c502b09b125db9acc965694d3ecff8e698/sdk/debugger_interface/linux/se_ptrace.c#L194-L199
+    // This can be used to compute the enclave base address:
+    //    encave-base-address + tcs.ossa = tcs_addr + 4096
+    //    enclave-base-address = tcs_addr + 4096 - tcs.ossa
+    enclave_base_address = (uint8_t*)tcs_addr + 4096 - tcs.ossa;
```
CodeMonkeyLeet, line 234: Consider using `OE_PAGE_SIZE` in place of the magic number 4096. 

Similarly in the test code that tries to reconstruct this logic.

anakrish:
> LGTM, thanks for the explanation in comments, very helpful!
> 
> Can you also remove the `OE_TD_FROM_TCS_BYTE_OFFSET` definition from constants.x64.h to prevent its future use?

Great suggestion. Will wait for #3339 and then remove all references to the define as part of this PR.

debugger/ptraceLib/enclave_context.c
anakrish: Done.

debugger/ptraceLib/enclave_context.c
```diff
@@ -224,8 +223,25 @@ static int _get_enclave_thread_current_ssa_info(
         return -1;
     }
 
+    // SSA is assigned to the page immediately after the tcs page.
+    // SSA address is also derived from adding tcs.ossa to enclave base address.
+    // See: host/sgx/create.c
```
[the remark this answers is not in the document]
anakrish: + @radhikaj For scheduling/decision about contributing upstream vs separating it out.

To me separating it out would be a good first step nonetheless as it would avoid accidental dependencies with rest of OE.
Two pending tasks to make this module separable/upstreamable are:
- Remove dependence on OE headers
- Document the binary contract - especially the following two assumptions
  - The OSSA page must allocated immediately after TCS page.
  - The byte offset at which to fetch __ssa_frame_size from td_t structure.

@radhikaj Let me know what the scheduling is; I can take a stab at moving this to a separate repo that would help upstream later.

We can sanitize the comments as part of that effort.

PR #3383, opened 2020-08-12
Remove unused oe_sgx_load_context_t.dev field

CodeMonkeyLeet:


host/sgx/sgxload.c
```diff
@@ -180,36 +180,24 @@ static void* _allocate_enclave_memory(size_t enclave_size, int fd)
...
-        /* If no file descriptor, then perform anonymous mapping and double
-         * the allocation size, so that BASE can be aligned on the SIZE
-         * boundary. This isn't neccessary on hardware backed enclaves, since
-         * the driver will do the alignment. */
-        if (fd == -1)
+        /* Perform an anonymous mapping and double the allocation size,
+         * so that BASE can be aligned on the SIZE boundary. This isn't
+         * necessary on hardware backed enclaves, since the driver will
+         * do the alignment. */
+        mflags |= MAP_ANONYMOUS | MAP_NORESERVE;
```
dthaler, line 187: Since this line is not inside any condition anymore, can we combine this line into line 180?

host/sgx/sgxload.c
```diff
@@ -180,36 +180,24 @@ static void* _allocate_enclave_memory(size_t enclave_size, int fd)
...
+        if (oe_safe_mul_u64(mmap_size, 2, &mmap_size) != OE_OK)
```
anakrish, line 188: In simulation mode do we still need this logic for alignment on SIZE boundary?
The Windows branch of the code does have alignment this alignment logic.

CodeMonkeyLeet: @anakrish I think it was only for verisimilitude of the simulation. Removing it doesn't cause any issues, so I've simplified this down to an mmap of the required size as per the Windows implementation for simplicity.

host/sgx/sgxload.c
CodeMonkeyLeet: Refactored into a single declaration per Anand's suggestion.

CodeMonkeyLeet:
bors try

PR #3384, opened 2020-08-12
<WIP>: Add support for large thread-local data

### LINKED ISSUE #1161: Thread Local Storage: Support unlimited thread-local data size

anakrish:
Calculate number of pages required for all TLS of the enclave image and
add them during enclave creation.

The distance between TCS and TD now depends on TLS size, so macro
TD_FROM_TCS is replaced by variable td_from_tcs_offset whose value is
patched during enclave creation.

Fix #1161

Signed-off-by: Thomas Tendyck <tt@edgeless.systems>

Add fcntl.edl necessary for code coverage build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - OE SDK Release packages can now be built on non-SGX and non-FLC machines.
+- Support for arbitrarily large thread-local data for SGX machines.
 
 ### Changed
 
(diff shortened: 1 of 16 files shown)


anakrish:
bors try

anakrish:
bors try

Issue #3375, opened 2020-08-11
Support SGX TCB Info version 3


yentsanglee:
SGX TCB Info V3 has a new field 'id'. OE SDK attestation verification must handle this.

- [V2 spec](https://api.portal.trustedservices.intel.com/documentation#pcs-certificate)
- [V3 spec](https://sbx.api.portal.trustedservices.intel.com/documentation#pcs-certificate-v3)

Source file - _read_tcb_info at [tcbinfo.c](https://github.com/openenclave/openenclave/blob/master/common/sgx/tcbinfo.c)

yentsanglee:
TCB Info structure update (to v3) will be pushed to a future API revision.

radhikaj:
Will do when Intel releases the dependency

yentsanglee:
New field 'id' will be postponed to later Intel release.

PR #3386, opened 2020-08-13
Support V3 tcbinfo parsing.

qiucwang:
Fixes issue #3375 
SGX attestation V3 spec was released and [new property `id` was added to tcbinfo](https://sbx.api.portal.trustedservices.intel.com/documentation#pcs-tcb-info-v3).
This PR is to support V3 tcbinfo:
1. An enum type field `oe_tcb_info_id_t id` was added to `oe_parsed_tcb_info_t` struct, which to be consistent with the `id` field in `qe_identity_info`.
2. Logic check `(value > 2) != has_id)` was added, assuming the `tcbinfo` which has the property `id`, should have version higher than `2`, and that doesn't have the property `id` should have a version lower than `3`.
3. V3 tcbinfo test case files were added under `tests/report/data_v3`.
4. The negative tests were isolated from the positive test function so that those negative tests do not need to be repeated in the `tcbInfo_with_pceid.json` positive test after the `tcbInfo.json` positive test was tested.

### DIFF
--- a/common/sgx/tcbinfo.c
+++ b/common/sgx/tcbinfo.c
@@ -500,6 +500,32 @@ done:
 }
 
 /**
+ * Type: tcbLevel in TCB Info (V3)
+ * Schema: Identical to tcbLevel in V2 TCB Info
+ * {
+ *    "tcb" : object of type tcb (Note: QE Identity info has the same object,
+ * but with different set of values). "tcbDate" : oe_datetime_t when TCB level
+ * was certified not to be vulnerable. ISO 8601 standard(YYYY-MM-DDThh:mm:ssZ).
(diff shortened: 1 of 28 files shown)


common/sgx/tcbinfo.c
```diff
@@ -590,16 +659,59 @@ static oe_result_t _read_tcb_info(
         }
     }
 
-    if (parsed_info->version == 2)
+    if (parsed_info->version == 3)
```
yentsanglee, line 662: This block is 95% identical to the version 2 block. Can we consider merging the 2 blocks and leave the diffs in _read_tcb_info_tcb_level_v*()? My concern is that as more versions come in the future, we will have to maintain multiple copies of similar code.

Issue #1852, opened 2019-06-03
Future considerations for  work on crypto refactoring


radhikaj:
Currently we have openenclave/internal/crypto for any APIs that are called in enclave. This includes APIs that are common between host and enclave. In the future:
1) we can remove headers that are used only by tests
2) remove crypto APIs in enclave and host that are for tests only
3) Differentiate between enclave and host APIs by passing in a magic number to indicate if called from host or enclave.

radhikaj:
See PR https://github.com/microsoft/openenclave/pull/1773

radhikaj:
As part of V0.6 I am only thinking of moving the crypto headers to openenclave/crypto if @mikbras thinks that is needed

radhikaj:
@mikbras has confirmed that headers should live in openenclave/internal/crypto for now and can be moved to openenclave/crypto when we decide to make them (semi)public.

radhikaj:
Moving this to backlog

CodeMonkeyLeet:
Per [SIG-arch meeting](https://hackmd.io/KOeBDVlyQpOC-OqqHYvs0A?both#12-May-2020), deferring to next session to discussion actions expected from the SIG.

radhikaj:
This design doc addresses this issue: https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/enclave_openssl_support.md
We should keep the issue open till the implementation is there.

radhikaj:
This design doc addresses this issue: https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/enclave_openssl_support.md
We should keep the issue open till the implementation is there. PR #2694

radhikaj:
This is considered cleanup work and we will drop it from SIG-Arch.

CodeMonkeyLeet:
Note that #3395 partially addresses the differentiation of magic numbers: host and enclave implementation structures have different identifiers, as do structures belonging to different crypto implementations (OpenSSL, BCrypt, MbedTLS) in the same environment.

PR #3395, opened 2020-08-14
Support OpenSSL-based crypto wrapper

mingweishih:
This PR adds the support the OpenSSL-based crypto wrapper (in addition to the original mbedtls-based one).  The implementation is based on the [enclave_openssl_support proposal](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/enclave_openssl_support.md).

The detail of the PR are as follows.
- Move `enclave/crypto` to `enclave/crypto/mbedtls`, which is corresponding to the `oecryptombedtls` (formerly named `oecryptombed`) library.
  - `oecryptombedtls`  is always built by default.
- Add `enclave/crypto/openssl`, which is corresponding to the `oecryptoopenssl` library.
  - `oecryptoopenssl` is built only when the experimental option `BUILD_OPENSSL` is `ON`.
  - Most files in `host/crypto/openssl` is moved to `common/crypto/openssl`, which are shared by both `oehost` and `oecryptoopenssl`
- In addition to `oeenclave`, now an enclave needs to explicitly link against either `oecryptombedtls` or `oecryptoopenssl` to use the OE crypto APIs.
- Update samples's CMakeLists to link against `oecryptombedtls`.
- Update each test that uses OE crpyto APIs to use both versions of the crypto wrappers (resulting into two tests).


mingweishih:
bors try

mingweishih:
> @mingweishih Thanks for doing this.
> 
> A high-level observation before delving too deeply:
> 
> Our strategy has been to put code that is shared between host and enclave in a `common` folder.
> In this PR, there is duplication between `host/crypto/openssl` and `enclave/crypto/openssl`.
> 
> Would this duplication be avoided in subsequent PRs?


That makes sense to me. Do you think that refactoring should be as part of this PR or a separated one?

anakrish:
> > @mingweishih Thanks for doing this.
> > A high-level observation before delving too deeply:
> > Our strategy has been to put code that is shared between host and enclave in a `common` folder.
> > In this PR, there is duplication between `host/crypto/openssl` and `enclave/crypto/openssl`.
> > Would this duplication be avoided in subsequent PRs?
> 
> That makes sense to me. Do you think that refactoring should be as part of this PR or a separated one?

It makes sense to do it as part of this PR. When I look at each file under `enclave/crypto/openssl`, I don't know if each file differs from the corresponding file on the host side, and in how much detail I need to review the contents of the file.

If the existing files are being reused, after moving to a common folder, then I can pretty much skip these files with the confidence that they have already been reviewed and tested for the past 2 years.

@CodeMonkeyLeet what do you think?

enclave/crypto/openssl/asn1.c
```diff
@@ -0,0 +1,270 @@
...
+            &date->day,
+            &date->hours,
+            &date->minutes,
+            &date->seconds,
+            &date->year) != 6)
+    {
+        OE_RAISE(OE_FAILURE);
+    }
+
+    /* Convert the month string to integer */
```
dthaler, line 205: libc already has an implementation of this, can we just call strptime?  Do we really need to have two implementations of this?

enclave/crypto/openssl/crl.c
```diff
@@ -0,0 +1,277 @@
...
+            &date->minutes,
+            &date->seconds,
+            &date->year) != 6)
+    {
+        OE_RAISE(OE_FAILURE);
+    }
+
+    /* Convert the month string to integer */
+    {
+        static const char* _month[] = {"Jan",
```
dthaler, line 171: Again, can we just call strptime and not have to reimplement date/string conversion ourselves?

tests/attestation_cert_apis/enc/enc.cpp
```diff
@@ -55,6 +68,211 @@ oe_result_t enclave_identity_verifier(oe_identity_t* identity, void* arg)
...
 
+#ifdef OE_USE_OPENSSL
+static oe_result_t generate_rsa_pair(
+    uint8_t** public_key,
+    size_t* public_key_size,
+    uint8_t** private_key,
+    size_t* private_key_size)
+{
+    oe_result_t result = OE_FAILURE;
+    uint8_t* local_public_key = NULL;
```
dthaler, line 79: style nit: use the C++ keyword nullptr instead of NULL in .cpp files

tests/attestation_cert_apis/enc/enc.cpp
```diff
@@ -1,10 +1,23 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+#ifdef OE_USE_OPENSSL
```
dthaler, line 4: It looks like this test is written to require two separate build lines.  It would be more efficient in my opinion if the same test binary could test both openssl and mbedtls, thus not requiring CI/CD to build everything twice.

CodeMonkeyLeet:
@mingweishih @anakrish 
> It makes sense to do it as part of this PR. When I look at each file under enclave/crypto/openssl, I don't know if each file differs from the corresponding file on the host side, and in how much detail I need to review the contents of the file.
>
> If the existing files are being reused, after moving to a common folder, then I can pretty much skip these files with the confidence that they have already been reviewed and tested for the past 2 years.
>
> @CodeMonkeyLeet what do you think?

Yes, I just left a similar comment as part of my review.

tests/attestation_cert_apis/enc/enc.cpp
mingweishih: It's required to build twice as the OE library can link against one of the crypto wrapper implementations at one time.

dthaler: That's not very efficient.   Just like we have plugins/modules for other things, why can't we do the same for this?  At worst, I think it should be a link-time, not compile-time, choice, right?

mingweishih: Refer to the [design](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/enclave_openssl_support.md#user-experience). Yes, it's an link-time option, but per design, we cannot link against both implementation at the same time---both mbed- and openssl-based implementation are used for the same set of crypto APIs.

tests/attestation_cert_apis/enc/enc.cpp
mingweishih: Changed to .cpp to .c (this file don't need to be .cpp).

tests/attestation_cert_apis/enc/enc.cpp
dthaler: In the same CI/CD build, you could link it twice (with different resulting binary names) and not have to compile everything twice.  However, it would be better in the future for it to be a runtime choice not a link-time choice.
This is not a blocker, but currently this complicates CI/CD and slows down every bors run because it has to compile it both ways and test it both ways, so doubles the cost.

enclave/crypto/mbedtls/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+add_enclave_library(
+  oecryptombed
+  STATIC
+  ../../../common/asn1.c
```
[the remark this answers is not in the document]
mingweishih: Done.

samples/file-encryptor/enclave/CMakeLists.txt
```diff
@@ -23,4 +23,5 @@ target_include_directories(
   PRIVATE ${CMAKE_CURRENT_SOURCE_DIR} # Needed for #include "../shared.h"
           ${CMAKE_CURRENT_BINARY_DIR})
 
-target_link_libraries(enclave openenclave::oeenclave openenclave::oelibcxx)
+target_link_libraries(enclave openenclave::oeenclave openenclave::oecryptombed
```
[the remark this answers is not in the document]
mingweishih: Yes, I'll defer that part to another PR that ports the samples to use OpenSSL-based wrapper.

mingweishih:
bors try

cmake/add_enclave.cmake
```diff
@@ -182,6 +186,12 @@ function (add_enclave_sgx)
   endif ()
 
   enclave_link_libraries(${ENCLAVE_TARGET} oeenclave)
+  if (ENCLAVE_CRYPTO_LIB STREQUAL "openssl")
+    enclave_link_libraries(${ENCLAVE_TARGET} oecryptoopenssl)
+  else ()
```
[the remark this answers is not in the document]
mingweishih: I've updated the PR accordingly. However, I'd like to get some clarifications.
- My understanding is that the default setting only applies to the internal SDK (i.e., tests). For samples/apps, now developers have to explicitly link `oembedcrypto` or `oeopensslcrypto` regardless of this setting. Of course, we can update the cmake script in the samples to add an option. However, given we didn't do the similar thing for allocators, I'm sure if we need to do that for the crypto wrappers.
- What do we do for the tests when the default library is set to OpenSSL? In the current PR, I add tests based on OpenSSL crypto wrapper when `BUILD_OPENSSL=ON` while keeping the original tests based on Mbed (which we always enable by default).  To ensure those original tests work correctly when the default library is OpenSSL, I have to pass the `CRYPTO_LIB=Mbed` argument to `add_enclave`. Therefore, having a default setting does not provide much benefits.

mingweishih:
> @mingweishih @anakrish
> 
> > It makes sense to do it as part of this PR. When I look at each file under enclave/crypto/openssl, I don't know if each file differs from the corresponding file on the host side, and in how much detail I need to review the contents of the file.
> > If the existing files are being reused, after moving to a common folder, then I can pretty much skip these files with the confidence that they have already been reviewed and tested for the past 2 years.
> > @CodeMonkeyLeet what do you think?
> 
> Yes, I just left a similar comment as part of my review.

@anakrish  @CodeMonkeyLeet  I've updated the PR for this accordingly.

tests/sealKey/enc/enc.cpp
```diff
@@ -632,6 +632,6 @@ OE_SET_ENCLAVE_SGX(
     1,    /* ProductID */
     1,    /* SecurityVersion */
     true, /* Debug */
-    256,  /* NumHeapPages */
+    1024, /* NumHeapPages */
```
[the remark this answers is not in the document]
mingweishih: Sure, I'll follow up on this.

mingweishih:
bors try

mingweishih:
bors try

samples/data-sealing/common/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ target_link_libraries(
          # Alternatively we could use a CMake OBJECT library, but that
          # requires a newish version of CMake.
          $<$<PLATFORM_ID:Linux>:-Wl,--require-defined=__oe_ecalls_table>
-         openenclave::oeenclave openenclave::oelibcxx)
+         openenclave::oeenclave openenclave::oecryptombed openenclave::oelibcxx)
```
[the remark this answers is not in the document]
mingweishih: The only requirement is that oecrypto should come after oeenclave.

tests/attestation_cert_apis/enc/enc.cpp
mingweishih: I revert the change (from .cpp to .c) because so will make the entire file as a newly added file and reviewers cannot know what part of the file are introduced by this PR. We can make a follow-up PR to change the file extension.

CMakeLists.txt
```diff
@@ -164,6 +164,21 @@ if (BUILD_OPENSSL AND OE_TRUSTZONE)
...
 
+set(DEFAULT_ENCLAVE_CRYPTO_LIB
+    "Mbed"
+    CACHE STRING "Default crypto library used by the enclaves.")
+if ((NOT DEFAULT_ENCLAVE_CRYPTO_LIB STREQUAL "Mbed")
+    AND (NOT DEFAULT_ENCLAVE_CRYPTO_LIB STREQUAL "OpenSSL"))
+  message(
+    FATAL_ERROR "Unsupported crypto library: ${DEFAULT_ENCLAVE_CRYPTO_LIB}")
+endif ()
+if ((DEFAULT_ENCLAVE_CRYPTO_LIB STREQUAL "OpenSSL") AND (NOT BUILD_OPENSSL))
```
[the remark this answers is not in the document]
mingweishih: Yes both oecrypto libraries are tested. For example, previously we have `crypto_enc`. With this PR, we will have `crypto_mbed_enc` and `crypto_openssl_enc`, each of them uses the corresponding oecrypto library. However, the porting is not completed yet (with tls_e2e and samples remaining), which @manojrupireddy is working on.

cmake/add_enclave.cmake
```diff
@@ -182,6 +187,23 @@ function (add_enclave_sgx)
   endif ()
 
   enclave_link_libraries(${ENCLAVE_TARGET} oeenclave)
+
+  # If the CRYPTO_LIB argument to add_enclave() is not set, the following
+  # logic determines the default crypto library based on the value of the
```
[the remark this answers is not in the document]
mingweishih: Rename to the `DEFAULT_TEST_ENCLAVE_CRYPTO_LIB`

mingweishih:
> The PR looks good for most part.
> Outstanding items
> 
> * Remove the notion of a DEFAULT_ENCLAVE_CRYPTO_LIBRARY since this PR removes the notion of a default oecrypto library, instead making the user explicitly choose the library.

Yes, I agree. We probably don't need this option at all given that
- The option only affects test.
- We now need to explicitly specify the crypto library for an enclave.

> * It would be ideal if both oecryptombed and oecryptoopenssl could be tested as part of the same build of OE SDK.

Both version of oecrypto libraries have already been tested in the same build (when enabling BUILD_OPENSSL=ON). We now have both `mbed` and `openssl` version of tests that only used `mbed` before. For example, `tests/crypto/enclave` becomes `tests/crypto/enclave_mbed`. In addition, we have `tests/crypto/enclave_openssl`.

The remaining porting job are the tls_e2e test and samples that @manojrupireddy is working on.

> * The pkgconfig workflow is missing from this PR. The sample makefiles have not been updated.

As mentioned, this PR does not port the samples to use OpenSSL/oecryptoopenssl yet, so there is no pkgconfig update. The only changes it adding an explicit link against `oecryptombed` to reflect the breaking change and to retain the original behavior (i.e., using oecryptombed/mbedtls). Note that there is no modifications needed to the our pkgconfig because already links against `oecryptombed` by default. The follow-up PR that ports the samples to use OpenSSL/oecryptoopenssl should include the necessary change to the pkgconfig.

CHANGELOG.md
```diff
@@ -24,6 +24,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   https://github.com/ARMmbed/mbedtls/releases/tag/v2.16.7) for the set of
   issues addressed.
 
+### Breaking change
+- Now enclave applications need to explicitly link against oecryptombed (or oecryptoopenssl).
```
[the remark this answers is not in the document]
radhikaj: @mingweishih , adding a couple of lines around what @anakrish describes would be helpful. I would like to point to this in the email I send out to the community regarding the breaking change.

cmake/add_enclave.cmake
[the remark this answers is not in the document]
mingweishih: Updated the variable name to `DEFAULT_TEST_ENCLAVE_CRYPTO_LIB`. Also, this PR already test both crypto libs whenever they're available (build the SDK with `BUILD_OPENSSL=ON`)

CHANGELOG.md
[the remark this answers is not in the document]
mingweishih: Updated the changelog. Note that only building enclaves with cmake are affected by this changes. Our make + pkgconfig already does this (explicitly link oeecryptombed), so no modifications are required.

mingweishih:
bors try

common/crypto/openssl/init.c
```diff
@@ -0,0 +1,81 @@
...
+        OE_TRACE_ERROR("OpenSSL initialization failed");
+        _finialize();
+    }
+    return;
+}
+#endif
+
+static void _initialize(void)
+{
+    OpenSSL_add_all_algorithms();
```
anakrish, line 65: Can the user initialize OpenSSL differently that what we are doing?
- If yes, can OpenSSL handle double initialization?
- On the host side, presumably, the user may want to initialize OpenSSL differently. In that case, should we enforce a specific kind of initialization on the user?

common/crypto/openssl/init.c
```diff
@@ -0,0 +1,81 @@
...
+    ERR_load_BIO_strings();
+    ERR_load_crypto_strings();
+
+#ifdef OE_BUILD_ENCLAVE
+    _initialize_rdrand_engine();
+#endif
+}
+
+void oe_initialize_openssl(void)
```
anakrish, line 74: We explicitly call openssl in the following places:
common/crypto/openssl/key.c
159:5:    oe_initialize_openssl();
216:5:    oe_initialize_openssl();
447:5:    oe_initialize_openssl();
517:5:    oe_initialize_openssl();

common/crypto/openssl/ec.c
251:5:    oe_initialize_openssl();
357:5:    oe_initialize_openssl();

common/crypto/openssl/init.c
74:6:void oe_initialize_openssl(void)

common/crypto/openssl/init.h
8:6:void oe_initialize_openssl(void);

common/crypto/openssl/cert.c
517:5:    oe_initialize_openssl();
562:5:    oe_initialize_openssl();
670:5:    oe_initialize_openssl();
740:5:    oe_initialize_openssl();
12 matches

Is this sufficient? What if the user calls some openssl function that does not hit the above locations

anakrish (commented):
LGTM for most part.
I have a few questions about oe_initialize_openssl.

Also:
It seems OpenSSL is not tested in CI. I don't see BUILD_OPENSSL=on being passed to any Jenkins pipeline.

mingweishih:
> LGTM for most part.
> I have a few questions about oe_initialize_openssl.
> 
> Also:
> It seems OpenSSL is not tested in CI. I don't see BUILD_OPENSSL=on being passed to any Jenkins pipeline.

We have one configuration that does the OpenSSL experimental feature:https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAzure-Linux/detail/Azure-Linux/1627/pipeline/51

common/crypto/openssl/cert.c
```diff
@@ -1013,3 +1028,396 @@ oe_result_t oe_cert_get_validity_dates(
...
+        name = NULL;
+    }
+
+    return name;
+}
+
+/*
+ * Decode the OID from its BER format (a hex string) into a dot-notation string
+ */
+static char* _decode_oid_to_str(char* oid, int oid_size)
```
CodeMonkeyLeet, line 1087: This is problematic ... any time we introduce parsing of new arbitrary binary data structures, there's a potential for vulnerabilities. I definitely prefer if we could spend some more time exploring additional implementation options here, possibly as a follow-up PR. 

One option may be to reuse the ASN handling code from mbedtls; their source code is nicely self-contained and partitioned, and I believe we've done this on the Windows side to get some parity behavior with OpenSSL.

common/crypto/openssl/cert.c
```diff
@@ -1013,3 +1028,396 @@ oe_result_t oe_cert_get_validity_dates(
 
     return result;
 }
+
+#ifdef OE_BUILD_ENCLAVE
+/*
+ * Parse the name string into X509_NAME struct. The format of the string is
+ * "KEY1=VALUE1,KEY2=VALUE2,KEY3=VALUE3...".
+ */
+static X509_NAME* _X509_parse_name(const char* name_string)
```
CodeMonkeyLeet, line 1037: How locked down is this string format? As far as I can tell, the string comes directly from the user with no sanitation and no format checking ... does this need to be made more robust? (i.e. stripping spaces, checking for escape characters, etc.) It would be much preferable if we did not have to handle this directly.

CodeMonkeyLeet (commented):
Mostly looks good, see inline comments for details.

One additional consideration crossed my mind: if we're introducing a breaking change now, and requiring that devs be familiar with the `libcryptombed` library, can we please also rename it to `libcryptombedtls` so that it has naming consistency with `libcryptoopenssl`? It was not really a problem as an internal include, but will probably cause more confusion given the non-standard naming once folks have to specify it as a choice. (Similarly, it would be useful for the build option to be "MbedTLS" and not just "Mbed" for consistency).

tests/attestation_cert_apis/enc/enc.cpp
mingweishih: Done.

common/crypto/openssl/cert.c
mingweishih: Updated the implementation based on mbedtls_x509_string_to_names and added a test.

common/crypto/openssl/cert.c
mingweishih: Removed this function.

common/crypto/openssl/init.c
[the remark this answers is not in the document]
mingweishih: Updated to call these functions only on the host.

common/crypto/rsa.h
```diff
@@ -1,8 +1,8 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
-#ifndef _OE_HOST_CRYPTO_RSA_H
-#define _OE_HOST_CRYPTO_RSA_H
+#ifndef _OE_COMMON_CRYPTO_RSA_H
```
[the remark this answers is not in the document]
mingweishih: Removed.

samples/log_callback/enclave/CMakeLists.txt
```diff
@@ -25,8 +25,10 @@ if (LVI_MITIGATION MATCHES ControlFlow)
...
   # Link against LVI-mitigated libraries.
-  target_link_libraries(enclave openenclave::oeenclave-lvi-cfg
-                        openenclave::oelibc-lvi-cfg)
+  target_link_libraries(
+    enclave openenclave::oeenclave-lvi-cfg openenclave::oecryptombed-lvi-cfg
+    openenclave::oelibc-lvi-cfg)
 else ()
-  target_link_libraries(enclave openenclave::oeenclave openenclave::oelibc)
+  target_link_libraries(enclave openenclave::oeenclave
+                        openenclave::oecryptombed openenclave::oelibc)
```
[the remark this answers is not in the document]
mingweishih: Yes, that's correct.

common/crypto/openssl/cert.c
```diff
@@ -17,14 +17,29 @@
...
 #include "ec.h"
 #include "init.h"
 #include "rsa.h"
 
+#ifdef OE_BUILD_ENCLAVE
+/*
+ * Value used by the _decode_oid_to_str() function. Although the OID
+ * standard does not pose a limitaion on the depth of an OID (i.e., the
+ * number of arcs), our implementation only supports a simple decoding
+ * with the limited depth (i.e., decoding into a fixed size string).
```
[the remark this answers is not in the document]
mingweishih: Lines were removed.

common/crypto/openssl/cert.c
```diff
@@ -17,14 +17,29 @@
...
 #include "asn1.h"
 #include "crl.h"
 #include "ec.h"
 #include "init.h"
 #include "rsa.h"
 
+#ifdef OE_BUILD_ENCLAVE
+/*
+ * Value used by the _decode_oid_to_str() function. Although the OID
+ * standard does not pose a limitaion on the depth of an OID (i.e., the
```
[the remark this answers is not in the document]
mingweishih: Lines were removed.

cmake/add_enclave.cmake
```diff
@@ -315,7 +337,7 @@ macro (add_enclave_optee)
   add_dependencies(${ENCLAVE_TARGET} ${ENCLAVE_TARGET}.ld)
   target_include_directories(${ENCLAVE_TARGET}
                              PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/optee)
-  target_link_libraries(${ENCLAVE_TARGET} oeenclave)
+  target_link_libraries(${ENCLAVE_TARGET} oeenclave oecryptombed)
```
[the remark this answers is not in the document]
mingweishih: Given that we don't support OpenSSL on OP-TEE now, what would be the recommendation for the case if we set `DEFAULT_TEST_ENCLAVE_CRYPTO_LIB` to OpenSSL on OPTEE (if we don't do the hardcode here)?

common/crypto/openssl/cert.c
```diff
@@ -1019,3 +1033,333 @@ oe_result_t oe_cert_get_validity_dates(
 
     return result;
 }
+
+#ifdef OE_BUILD_ENCLAVE
+/*
+ * Parse the name string into X509_NAME struct. The format of the string is
```
dthaler, line 1039: It's bad form to put OE_BUILD_ENCLAVE-only functions in the common directory.
This should move to a file in the enclave directory outside of common.
(Ideally, one would never see OE_BUILD_ENCLAVE used anywhere in the common directory.)

common/crypto/openssl/cert.c
```diff
@@ -1019,3 +1033,333 @@ oe_result_t oe_cert_get_validity_dates(
...
+        BIO_free(bio);
+    if (obj)
+        ASN1_OBJECT_free(obj);
+    if (data)
+        ASN1_OCTET_STRING_free(data);
+    if (bc)
+        BASIC_CONSTRAINTS_free(bc);
+    if (subject_issuer_key_pair)
+        EVP_PKEY_free(subject_issuer_key_pair);
+    if (buff)
```
dthaler, line 1356: Suggestion: free() is safe to call with NULL.  Delete unnecessary if tests, to simplify code.

common/crypto/openssl/init.c
```diff
@@ -0,0 +1,89 @@
...
+#include <openssl/engine.h>
+static oe_once_t _openssl_initialize_once;
+static ENGINE* eng;
+#else
+#include <pthread.h>
+static pthread_once_t _once = PTHREAD_ONCE_INIT;
+#endif
+
+#ifdef OE_BUILD_ENCLAVE
+static void _finalize(void)
```
dthaler, line 20: Move OE_BUILD_ENCLAVE code to the enclave subdir instead of under the common subdir.

common/crypto/openssl/magic.h
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#ifndef _OE_COMMON_CRYPTO_OPENSSL_MAGIC_H
+#define _OE_COMMON_CRYPTO_OPENSSL_MAGIC_H
+
+#ifdef OE_BUILD_ENCLAVE
+/* oe_crypto magic numbers for enclave OpenSSL implementation */
```
dthaler, line 8: Move to separate files under host and enclave, rather than ifdefs in common.

common/crypto/openssl/rsa.c
```diff
@@ -365,6 +362,7 @@ oe_result_t oe_rsa_public_key_equal(
         (oe_public_key_t*)public_key1, (oe_public_key_t*)public_key2, equal);
 }
 
+#ifndef OE_BUILD_ENCLAVE
 oe_result_t oe_rsa_get_public_key_from_private(
```
dthaler, line 366: Move to a file under host instead of common

tests/attestation_cert_apis/enc/mbedtls_rsa.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+    OE_TRACE_INFO("public_key_size\n[%d]\n", *public_key_size);
+    OE_TRACE_INFO("public_key\n[%s]\n", *public_key);
+    result = OE_OK;
+
+done:
+    if (local_public_key)
+        free(local_public_key);
+    if (local_private_key)
+        free(local_private_key);
```
dthaler, line 98:
```suggestion
    free(local_public_key);
    free(local_private_key);
```
to simplify code.  free() is safe to call with null.

tests/attestation_cert_apis/enc/openssl_rsa.cpp
```diff
@@ -0,0 +1,123 @@
...
+    RSA* rsa;
+    BIO* bio = nullptr;
+    BIGNUM* e;
+    size_t local_public_key_size = 512;
+    size_t local_private_key_size = 2048;
+
+    // Generate RSA key
+    pkey = EVP_PKEY_new();
+    e = BN_new();
+    res = BN_set_word(e, (BN_ULONG)RSA_F4);
```
dthaler, line 34: BN_new does a malloc and so can fail with NULL, at which point this test will crash instead of failing gracefully.  Same in line 38 etc.

tests/attestation_cert_apis/enc/openssl_rsa.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+    OE_TRACE_INFO("public_key_size\n[%d]\n", *public_key_size);
+    OE_TRACE_INFO("public_key\n[%s]\n", *public_key);
+    result = OE_OK;
+
+done:
+    if (local_public_key)
+        free(local_public_key);
+    if (local_private_key)
+        free(local_private_key);
```
dthaler, line 114:
```suggestion
    free(local_public_key);
    free(local_private_key);
```

tests/attestation_plugin_cert/enc/enc.cpp
```diff
@@ -1,10 +1,15 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
-#include <mbedtls/ctr_drbg.h>
-#include <mbedtls/entropy.h>
+#ifdef OE_USE_OPENSSL
```
dthaler, line 4: To verify: is this built and tested in CI/CD both with and without this defined?

tests/attestation_plugin_cert/enc/mbedtls_rsa.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+    OE_TRACE_INFO("public_key_size\n[%d]\n", *public_key_size);
+    OE_TRACE_INFO("public_key\n[%s]\n", *public_key);
+    result = OE_OK;
+
+done:
+    if (local_public_key)
+        free(local_public_key);
+    if (local_private_key)
+        free(local_private_key);
```
dthaler, line 98:
```suggestion
    free(local_public_key);
    free(local_private_key);
```

tests/attestation_plugin_cert/enc/openssl_rsa.cpp
```diff
@@ -0,0 +1,123 @@
...
+    EVP_PKEY* pkey;
+    RSA* rsa;
+    BIO* bio = nullptr;
+    BIGNUM* e;
+    size_t local_public_key_size = 512;
+    size_t local_private_key_size = 2048;
+
+    // Generate RSA key
+    pkey = EVP_PKEY_new();
+    e = BN_new();
```
dthaler, line 33: need to test for null alloc or this test can crash.  Same in 38 etc.

tests/attestation_plugin_cert/enc/openssl_rsa.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+    OE_TRACE_INFO("public_key_size\n[%d]\n", *public_key_size);
+    OE_TRACE_INFO("public_key\n[%s]\n", *public_key);
+    result = OE_OK;
+
+done:
+    if (local_public_key)
+        free(local_public_key);
+    if (local_private_key)
+        free(local_private_key);
```
dthaler, line 114:
```suggestion
    free(local_public_key);
    free(local_private_key);
```

tests/crypto/tests.c
```diff
@@ -9,6 +9,9 @@ void TestAll()
     TestASN1();
 #endif
     TestCRL();
+#if defined(OE_BUILD_ENCLAVE)
```
dthaler, line 12: Why is this ifdef needed here but not in tests.h?

tests/attestation_plugin_cert/enc/enc.cpp
mingweishih: Yes, both cases are tested. There will be two tests: attestation_plugin_cert_mbedtls (without `OE_USE_OPENSSL` defined) and attestation_plugin_cert_openssl (with `OE_USE_OPENSSL` defined).

tests/attestation_cert_apis/enc/openssl_rsa.cpp
mingweishih: updated.

tests/attestation_plugin_cert/enc/openssl_rsa.cpp
mingweishih: updated

tests/crypto/tests.c
mingweishih: Updated

mingweishih:
bors try

common/sgx/eeid.c
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 #include <stdio.h>
+#include <stdlib.h>
```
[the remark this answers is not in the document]
mingweishih: This is caused by a side effect from not linking oeenclave against oembedtls by default.

mingweishih:
bors try

CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [Unreleased][Unreleased_log]
 --------------
 
+## Breaking Changes
+- liboembed is renamed to liboembedtls.
```
anakrish, line 14: An install of OE SDK built from current master has the following mbed libraries:
- libmbedcrypto.a 
- libmbedtls.a     
- liboecryptombed.a 
- libmbedx509.a   

liboembed is not in the list. Did you mean libmbedtls.a ?

It is not clear what the difference is between liboecryptombed.a and libmbedcrypto.a.

CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- The `openenclave-config.cmake` and `openenclave-lvi-mitigation-config.cmake` no longer include liboembedtls as `PUBLIC` link requirement for liboeenclave.
```
anakrish, line 15: There is no liboembedtls currently.

CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+    - Enclave apps that build with CMake and use the Open Enclave cmake configurations must now explicitly include `openenclave::oecryptombedtls` when linking `openenclave::oeenclave`.
```
anakrish, line 16: The name `oecryptombedtls` indicates that tls functionality is being added to the enclave, which not all enclaves might want to do.
Did you mean `oecryptombed` ?

CMakeLists.txt
```diff
@@ -164,6 +164,22 @@ if (BUILD_OPENSSL AND OE_TRUSTZONE)
   message(FATAL_ERROR "BUILD_OPENSSL is not supported on ARM yet.")
 endif ()
 
+set(DEFAULT_TEST_ENCLAVE_CRYPTO_LIB
+    "MbedTLS"
```
anakrish, line 168: It is better to 
- use  lower case characters
- perform case-insensitive compare

The mbed tls website (https://tls.mbed.org/about-us) itself does not follow a consistent capitalization scheme for the library,. It is not really obvious that `MbedTLS` is the preferred way to spell the option.

cmake/add_enclave.cmake
```diff
@@ -182,6 +187,23 @@ function (add_enclave_sgx)
...
+
+  # If the CRYPTO_LIB argument to add_enclave() is not set, the following
+  # logic determines the default crypto library based on the value of the
+  # DEFAULT_TEST_ENCLAVE_CRYPTO_LIB global variable (e.g., either "MbedTLS" or "OpenSSL").
+  # If the CRYPTO_LIB argument is set, it overrides the DEFAULT_TEST_ENCLAVE_CRYPTO_LIB.
+  if (NOT ENCLAVE_CRYPTO_LIB)
+    set(ENCLAVE_CRYPTO_LIB ${DEFAULT_TEST_ENCLAVE_CRYPTO_LIB})
+  endif ()
+
+  if (ENCLAVE_CRYPTO_LIB STREQUAL "MbedTLS")
```
anakrish, line 199: Prefer case-insensitive compare

common/crypto/openssl/cert.c
```diff
@@ -651,8 +657,10 @@ oe_result_t oe_cert_chain_read_pem(
         tmp_pem_data[pem_size] = '\0';
     }
 
+#if !defined(OE_BUILD_ENCLAVE)
     /* Initialize OpenSSL (if not already initialized) */
-    oe_initialize_openssl();
+    oe_crypto_initialize();
```
anakrish, line 662: How can we be sure that we are not missing any control path that requires initialization?

enclave/CMakeLists.txt
```diff
@@ -78,4 +79,11 @@ install_enclaves(
 
 if (WITH_EEID)
   enclave_compile_definitions(oeenclave PRIVATE OE_WITH_EXPERIMENTAL_EEID)
+  # oeenclave currently does not link against the oecrypto* library by default.
+  # Consequently, it does not inherit the include paths of the crypto (e.g., mbedtls)
+  # headers, which are needed by eeid.c. Therefore, we add the paths here.
+  enclave_include_directories(
```
anakrish, line 85: Shouldn't this be openssl and mbedtls headers?

enclave/crypto/mbedtls/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+add_enclave_library(
+  oecryptombedtls
```
anakrish, line 5: The list of files indicate that there is no TLS functionality in this library.
`oecryptombed` is a better name.

enclave/crypto/openssl/init.c
```diff
@@ -0,0 +1,65 @@
...
+
+static void _initialize(void)
+{
+    /*
+     * OpenSSL in the enclave requires to explicitly register the RDRAND engine.
+     */
+    _initialize_rdrand_engine();
+}
+
+void oe_crypto_initialize(void)
```
anakrish, line 62: Instead of calling oe_crypto_initialize explicitly in various control paths which raises possibility of missing some control path, it is better if the OE SDK runtime can call it deterministically at the most appropriate time during initialization.

To do that, we need to:
- provide a default empty implementation of oe_crypto_initialize in a separate file
- at oe_crypto_initialize to the list of required symbols in link.c
- call oe_crypto_initialize at the right place during enclave initialization

mingweishih: These suggested changes are already done as part of this PR. Enclave will already invoke oe_crypto_initialize as part of initialization.

enclave/CMakeLists.txt
mingweishih: EEID implementation currently is based on mbedtls only.

common/crypto/openssl/cert.c
mingweishih: Currently implementation retains the behavior on the host (lazy initialization), but explicitly does the initialization on the enclave side.

CHANGELOG.md
mingweishih: There is a typo, thanks for catching this.

enclave/crypto/mbedtls/CMakeLists.txt
mingweishih: This is for parity with the "OpenSSL" naming.

@CodeMonkeyLeet may add more comments on this.

cmake/add_enclave.cmake
mingweishih: Done.

CMakeLists.txt
mingweishih: Done.

mingweishih:
bors try

CMakeLists.txt
dthaler: Actually in prose it's always "Mbed TLS" with a space in it.  https://www.trustedfirmware.org/projects/mbed-tls/ is the official home (like OE is under CCC) and https://github.com/ARMmbed/mbedtls is the github project.  Both are consistent. (The tls.mbed.org site is obsolete and has the notice on it "The information on this website will be retained while we migrate but please note it may be out of date.")

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,350 @@
...
+            OE_CRYPTO_ERROR, "create basic constraint extension failed");
+
+    ret = X509_add_ext(x509cert, ext, -1);
+    if (!ret)
+        OE_RAISE_MSG(OE_CRYPTO_ERROR, "add basic constraint extension failed");
+
+    X509_EXTENSION_free(ext);
+    ext = NULL;
+
+    /* Set the subject key idenfitifer extension */
```
dthaler, line 254:
```suggestion
    /* Set the subject key identifier extension */
```

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,350 @@
...
+            OE_CRYPTO_ERROR, "create subject key identifier extension failed");
+
+    ret = X509_add_ext(x509cert, ext, -1);
+    if (!ret)
+        OE_RAISE_MSG(OE_CRYPTO_ERROR, "add basic constraint extension failed");
+
+    X509_EXTENSION_free(ext);
+    ext = NULL;
+
+    /* Set the authority key idenfitifer extension */
```
dthaler, line 267:
```suggestion
    /* Set the authority key identifier extension */
```

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,350 @@
...
+            OE_CRYPTO_ERROR, "create subject key identifier extension failed");
+
+    ret = X509_add_ext(x509cert, ext, -1);
+    if (!ret)
+        OE_RAISE_MSG(OE_CRYPTO_ERROR, "add basic constraint extension failed");
+    X509_EXTENSION_free(ext);
+    ext = NULL;
+
+    /* Set the custom extension */
+    data = ASN1_OCTET_STRING_new();
```
dthaler, line 281: missing check for alloc failure

enclave/crypto/openssl/init.c
```diff
@@ -0,0 +1,65 @@
...
+        OE_TRACE_ERROR("OpenSSL initialization failed");
+        _finalize();
+    }
+    return;
+}
+
+static void _initialize(void)
+{
+    /*
+     * OpenSSL in the enclave requires to explicitly register the RDRAND engine.
```
dthaler, line 57:
```suggestion
     * OpenSSL in the enclave requires us to explicitly register the RDRAND engine.
```

tests/tools/oesign/test-enclave/enclave/CMakeLists.txt
```diff
@@ -3,6 +3,16 @@
 
 set(EDL_FILE ../oesign_test.edl)
 
+if (DEFAULT_TEST_ENCLAVE_CRYPTO_LIB STREQUAL "MbedTLS")
```
dthaler, line 6: is this ok being case sensitive or is a TOLOWER operation needed here like I saw in another file in this PR?

mingweishih: Thanks for the catch. Updated.

common/crypto/openssl/cert.c
dthaler: In the future, I'd like to see this unified so the behavior is the same on the host and the enclave so all these ifdefs can go away.

common/crypto/openssl/rsa.c
```diff
@@ -0,0 +1,340 @@
...
+
+    /* Compare modulus and exponent */
+    if (BN_cmp(n1, n2) == 0 && BN_cmp(e1, e2) == 0)
+        *equal = true;
+
+    result = OE_OK;
+
+done:
+
+    if (rsa1)
```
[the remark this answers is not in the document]
mingweishih: Yes, I agree. My suggestion is keeping the consistency in this PR, and have another PR that does the check and update on such patterns.

common/crypto/openssl/cert.c
mingweishih: Once we drop the ubuntu 16.04 (which supports older version of OpenSSL), we will be able to remove these code.

mingweishih:
bors try

mingweishih:
bors try

CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [Unreleased][Unreleased_log]
 --------------
 
+## Breaking Changes
+- liboecryptombed is renamed to liboecryptombedtls.
+- The `openenclave-config.cmake` and `openenclave-lvi-mitigation-config.cmake` no longer include liboecryptombedtls as `PUBLIC` link requirement for liboeenclave.
```
CodeMonkeyLeet, line 15: It might help to link the two statements:
```suggestion
- liboecryptombed is now called liboecryptombedtls and will no longer be automatically included as a link dependency when linking liboeenclave in CMake.
   - The `openenclave-config.cmake` and `openenclave-lvi-mitigation-config.cmake` will not specify the renamed liboecryptombedtls as a `PUBLIC` link requirement for liboeenclave.
```

cmake/add_enclave.cmake
```diff
@@ -318,7 +341,7 @@ macro (add_enclave_optee)
   add_dependencies(${ENCLAVE_TARGET} ${ENCLAVE_TARGET}.ld)
   target_include_directories(${ENCLAVE_TARGET}
                              PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/optee)
-  target_link_libraries(${ENCLAVE_TARGET} oeenclave)
+  target_link_libraries(${ENCLAVE_TARGET} oeenclave oecryptombedtls)
```
CodeMonkeyLeet, line 344: Follow-up to previous comment that got dislocated:

>> Part of the suggestion to introduce the `DEFAULT_TEST_ENCLAVE_CRYPTO_LIB` is so that we have a consistent way to flip the default if we want to encourage OpenSSL as the primary enclave crypto library instead. To that end, we want to avoid hardcoding it in cases where the crypto library is not material to the test.
>> 
>> For OP-TEE, I would like to know that if we swapped the default, we would get testing on the aarch64 configs as to what breaks.
>>
> Given that we don't support OpenSSL on OP-TEE now, what would be the recommendation for the case if we set `DEFAULT_TEST_ENCLAVE_CRYPTO_LIB` to OpenSSL on OPTEE (if we don't do the hardcode here)?

I suggest making the logic from line 199-206 reusable here. My understanding is that you don't have to worry about the variable indicating OpenSSL when building OP-TEE because /CMakeLists.txt already implements the top level blocks for:
- `DEFAULT_TEST_ENCLAVE_CRYPTO_LIB` cannot be OpenSSL without `BUILD_OPENSSL`
- `BUILD_OPENSSL` cannot be specified with `OE_TRUSTZONE`

Having the logic in place here means that you only have to flip the switch on supporting OpenSSL in OP-TEE in one place in the build configuration (root CMakeLists.txt) instead of having to figure out all the other places lower in the tree where that decision was also made locally.

common/sgx/eeid.c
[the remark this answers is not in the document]
CodeMonkeyLeet: Non-blocking, but I'm not sure I follow ... the header inclusion suggests a missing compile time symbol, not a link time one. Are you referring to some CMake-specific interaction? How does that work?

common/crypto/magic.h
```diff
@@ -1,8 +1,8 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
-#ifndef _OE_HOST_CRYPTO_MAGIC_H
-#define _OE_HOST_CRYPTO_MAGIC_H
+#ifndef _OE_COMMON_CRYPTO_MAGIC_H
```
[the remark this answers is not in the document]
CodeMonkeyLeet: Thanks for splitting up all the magic numbers, @mingweishih! I've updated #1852 to note this.

enclave/crypto/mbedtls/CMakeLists.txt
CodeMonkeyLeet: @anakrish @mingweishih Yes, it's for parity with the OpenSSL naming, and while the contents of the resulting lib does not contain TLS functionality, it technically doesn't contain the crypto implementation either, just the wrappers over the MbedTLS impl; its role as illustrated in the dev sample using CMake is to pull in all the MbedTLS libs as well (`PUBLIC` link to mbedcrypto on line 27), which does include crypto, X509 and TLS.

tests/crypto/enclave/enc/CMakeLists.txt
```diff
@@ -26,21 +26,51 @@ set(SRCS
...
+  CRYPTO_LIB
+  mbedtls
   SOURCES
-  ${SRCS}
-  crypto_t.c)
+  enc.c
+  crypto_t.c
+  ${SRCS})
+
+#enclave_link_libraries(crypto_mbedtls_enc crypto_object)
```
CodeMonkeyLeet, line 45: Remove unused directives here and on line 67

CodeMonkeyLeet (commented):
Nice work, I still need to review the openSSL cert implementation and tests but I'm breaking here so that you can address the comment about the handling of OIDs without waiting for me to finish.

common/sgx/eeid.c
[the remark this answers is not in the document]
mingweishih: Here is more explanation on this:
When the `oeenclave` links against `oecryptombed` (or `oecrytombedtls` now), the `oeenclave` will inherit the compile define of `-DMBEDTLS_CONFIG_FILE="<config.h>"`. This ends up causing the `stdlib.h` be implicitly included along with the mbedtls headers (e.g., `mbedtls/pk.h`).

With this PR, the `oeenclave` no longer links against the `oecryptombedtls`, hence will not inherit the compile define. As a result, we need to manually include `stdlib.h` (for the use of `calloc`).

enclave/crypto/mbedtls/init.c
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/internal/crypto/init.h>
+
+void oe_crypto_initialize(void)
+{
```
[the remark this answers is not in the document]
mingweishih: Done.

CHANGELOG.md
mingweishih: Updated.

cmake/add_enclave.cmake
mingweishih: Updated as suggested.

common/crypto/openssl/crl.c
```diff
@@ -14,10 +14,15 @@
 #include <string.h>
 #include <time.h>
 
-#include "../magic.h"
 #include "asn1.h"
 #include "crl.h"
 
+#if !defined(OE_BUILD_ENCLAVE)
```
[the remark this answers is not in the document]
mingweishih: The change was suggested by @dthaler. I changed back to the original implementation which is simpler as suggested.

enclave/tls_cert.c
```diff
@@ -239,9 +263,23 @@ oe_result_t oe_get_attestation_certificate_with_evidence(
     OE_CHECK_MSG(
         result, "oe_get_evidence failed with %s\n", oe_result_str(result));
 
+    switch (oe_get_crypto_impl())
```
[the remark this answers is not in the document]
mingweishih: I bring back the implementation (to the first suggestion) and added the test accordingly.

tests/crypto/enclave/enc/CMakeLists.txt
mingweishih: Done

tests/debugger/oegdb/enc/CMakeLists.txt
```diff
@@ -15,6 +15,8 @@ add_enclave(
   oe_gdb_test_enc
   UUID
   ff375399-0b06-4189-9ddc-df5e38612e6d
+  CRYPTO_LIB
+  mbedtls
```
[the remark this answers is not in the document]
mingweishih: Added comments.

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,457 @@
...
+    {
+        X509_NAME_free(name);
+        name = NULL;
+    }
+
+    return name;
+}
+
+/*
+ * Decode the OID from its BER format (a hex string) into a dot-notation string
```
CodeMonkeyLeet, line 121: I would suggest referring to this as a _byte-encoding_ rather than a hex string as the latter can be misunderstood as "0xA2" rather than what you intend here.

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,457 @@
...
+    X509* x509cert = NULL;
+    X509V3_CTX ctx;
+    BIO* bio = NULL;
+    X509_NAME* name = NULL;
+    EVP_PKEY* subject_issuer_key_pair = NULL;
+    X509_EXTENSION* ext = NULL;
+    ASN1_OBJECT* obj = NULL;
+    ASN1_OCTET_STRING* data = NULL;
+    BASIC_CONSTRAINTS* bc = NULL;
+    unsigned char* buff = NULL;
```
CodeMonkeyLeet, line 225: Style: prefer verbose names over abbreviations (this is inconsistently abbreviated as `buf` elsewhere in the codebase for example)

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,457 @@
...
+    if (obj)
+        ASN1_OBJECT_free(obj);
+    if (data)
+        ASN1_OCTET_STRING_free(data);
+    if (bc)
+        BASIC_CONSTRAINTS_free(bc);
+    if (subject_issuer_key_pair)
+        EVP_PKEY_free(subject_issuer_key_pair);
+    if (buff)
+        oe_free(buff);
```
CodeMonkeyLeet, line 452: There was a comment in other discussions that we should be more consistent in nulling pointers after free to prevent use-after-free issues. Might be good to consider that here, and also to free the `p` pointer.

tests/crypto/cert_tests.c
```diff
@@ -0,0 +1,135 @@
...
+#include "../../../enclave/crypto/openssl/cert.h"
+#endif
+
+/* Test the internal API that is only avaiable on OpenSSL-based implementation.
+ */
+#if defined(OE_USE_OPENSSL)
+static void _test_X509_parse_name()
+{
+    X509_NAME* name = NULL;
+    char buf[256];
```
CodeMonkeyLeet, line 20: Use constants of defines instead of magic numbers where possible for readability - this value seems to be based off `OE_X509_MAX_NAME_SIZE` for example, and may warrant its inclusion in an internal header like enclave/crypto/openssl/cert.h

tests/crypto/cert_tests.c
```diff
@@ -0,0 +1,135 @@
...
+
+    printf("=== begin %s()\n", __FUNCTION__);
+
+    /* Basic case */
+    name = X509_parse_name("CN=Open Enclave SDK,O=OESDK TLS,C=US");
```
CodeMonkeyLeet, line 25: I'm always leery of adding new string parsers because of the testing surface involved. Consider expanding the test cases to consider or things we need to explicitly block to scope testing. Looking at [RFC 4514](https://tools.ietf.org/html/rfc4514)
- UTF-8 strings and canonicalization of name segments 
- Special character handling
- Escape handling of things other than commas

CodeMonkeyLeet (commented):
Okay, some of the previous fixes look good, see additional comments specifically around the cert functions.

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,457 @@
...
+
+/*
+ * Value used by the _decode_oid_to_str() function. Although the OID
+ * standard does not limit the depth of an OID definition tree (i.e., the
+ * number of arcs), our implementation only supports a simple decoding
+ * with a limited depth (i.e., decoding into a fixed size string).
+ */
+#define OE_OID_MAX_LENGTH 200
+
+#define OE_X509_MAX_NAME_SIZE 256
```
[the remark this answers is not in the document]
mingweishih: Added a comment.

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,457 @@
...
+         * decoded into two integers (x and y). The rule is as follows. z = 40 *
+         * x + y, where x = 0, 1, or 2.
+         */
+        if (strlen(oid_buf) == 0)
+        {
+            long first;
+            long second;
+
+            first = block / 40;
+            if (first > 2)
```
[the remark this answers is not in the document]
mingweishih: Done

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,457 @@
...
+        block <<= 7L;
+        block |= byte & 0x7f;
+
+        /*
+         * Zero length of oid_buf indicates that we're decoding the first block,
+         * which requires special handling. The value of the block (z) is
+         * decoded into two integers (x and y). The rule is as follows. z = 40 *
+         * x + y, where x = 0, 1, or 2.
+         */
+        if (strlen(oid_buf) == 0)
```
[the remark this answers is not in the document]
mingweishih: Updated.

enclave/crypto/openssl/cert.c
mingweishih: Updated.

enclave/crypto/openssl/cert.c
mingweishih: Updated.

enclave/crypto/openssl/cert.c
mingweishih: Updated.

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,457 @@
...
+
+    /*
+     * By default, the config->oid_ext stores the OID in the encoded form
+     * (BER-TLV) that can directly be consumed by Mbed TLS APIs. However,
+     * OpenSSL APIs require the OID in the decoded form. Therefore, we need to
+     * decode the OID first.
+     */
+    oid = decode_oid_to_str((char*)config->ext_oid, (int)config->ext_oid_size);
+    if (!oid)
+        OE_RAISE_MSG(OE_CRYPTO_ERROR, "decode oid failed");
```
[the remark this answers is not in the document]
mingweishih: Updated.

tests/attestation_plugin_cert/enc/openssl_rsa.cpp
```diff
@@ -0,0 +1,132 @@
...
+{
+    oe_result_t result = OE_FAILURE;
+    uint8_t* local_public_key = nullptr;
+    uint8_t* local_private_key = nullptr;
+    int res = -1;
+    EVP_PKEY* pkey = nullptr;
+    RSA* rsa = nullptr;
+    BIO* bio = nullptr;
+    BIGNUM* e = nullptr;
+    size_t local_public_key_size = 512;
```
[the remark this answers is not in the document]
mingweishih: Updated

tests/attestation_plugin_cert/enc/openssl_rsa.cpp
```diff
@@ -0,0 +1,132 @@
...
+    );
+
+    if (!res)
+        OE_RAISE_MSG(OE_FAILURE, "RSA_generate_key failed (%d)\n", res);
+
+    // Assign RSA key to EVP_PKEY structure
+    EVP_PKEY_assign_RSA(pkey, rsa);
+
+    // Allocate memory
+    local_public_key = (uint8_t*)malloc(local_public_key_size);
```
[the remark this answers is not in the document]
mingweishih: Updated.

enclave/crypto/openssl/cert.c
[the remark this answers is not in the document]
mingweishih: Reverted back the change and added comments. It turns out that the value first can only be `0`, `1`, and `2`. If the value of the first byte exceeds 80, `first` will be `2` and `second` can be arbitrarily large.

tests/crypto/cert_tests.c
```diff
@@ -0,0 +1,135 @@
...
+        name = X509_parse_name(test);
+        OE_TEST(name == NULL);
+    }
+
+    printf("=== passed %s()\n", __FUNCTION__);
+}
+
+static void _test_decode_oid_to_str()
+{
+    char* oid;
```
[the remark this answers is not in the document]
mingweishih: Added more test cases (both positive and negative)

tests/crypto/cert_tests.c
mingweishih: Current implementation is based on Mbed TLS's (https://github.com/openenclave/openenclave-mbedtls/blob/2b735b85590e661870bc8217cc09139ba17e9a6f/library/x509_create.c#L153). I believe the most of cases you described are not handled by this implementation. Would you suggest adding negative tests on them?

mingweishih:
bors try

mingweishih:
bors try

common/crypto/openssl/crl.c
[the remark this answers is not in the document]
CodeMonkeyLeet: @dthaler I found the comment and I think it's predicated on the assumption that the ifdef goes away entirely instead of getting dispersed through all the source files, is that correct? I'm very much of the opinion that build switch handling should obey the DRY principle. If the concern is not wanting host/enclave specific definitions as part of a common file, then it's still possible for the common/magic.h to do the build check once and #include the appropriate host/enclave copy of magic.h (i.e. common/magic.h becomes a redirect header, which is not uncommon in libraries like libc that need to aggregate across different arch implementations, for example).

enclave/crypto/openssl/cert.c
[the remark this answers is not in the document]
CodeMonkeyLeet: Thanks for the explanation and comments, that helps!

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,479 @@
...
+ * standard does not limit the depth of an OID definition tree (i.e., the
+ * number of arcs), our implementation only supports a simple decoding
+ * with a limited depth (i.e., decoding into a fixed size string).
+ */
+#define OE_OID_MAX_LENGTH 200
+
+/*
+ * Parse the name string into X509_NAME struct. The format of the string is
+ * "KEY1=VALUE1,KEY2=VALUE2,KEY3=VALUE3...". The implementation is based
+ * on the mbedtls_x509_string_to_names from Mbed TLS.
```
CodeMonkeyLeet, line 32: Instead of cloning an implementation, have we considered simply using the mbedtls implementation directly? It's a nice code base because many files are self contained, and we have precedence for this on the host side with the BCrypt wrappers using mbedtls' implementation of bignum because BCrypt doesn't expose its own. 

From a cursory glance, it looks like this would just pull in x509_create.c and asn1write.c, although the former does have a lot of symbols that are not easy to hide in the resulting .a file (shouldn't be a problem realistically since at link time we specify `-gc-sections` though).

I think that also simplifies the testing story in that we don't need to maintain parser specific tests for our fork of it.

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,479 @@
...
+        second = byte - 40 * first;
+        len = snprintf(tmp_buf, sizeof(tmp_buf), "%ld.%ld", first, second);
+        if (len <= 0 || len >= OE_OID_MAX_LENGTH)
+            goto done;
+
+        strncat(oid_buf, tmp_buf, (size_t)len);
+    }
+
+    /*
+     * Break the hex string into blocks, each of which represents a integer,
```
CodeMonkeyLeet, line 165: Same comment about "hex string" -> "byte-encoding"

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,479 @@
...
+        name = NULL;
+    }
+
+    return name;
+}
+
+/*
+ * Decode the OID from its BER format (byte-encoding) into a dot-notation string
+ */
+char* decode_oid_to_str(char* oid, int oid_size)
```
CodeMonkeyLeet, line 121: Updated implementation looks fine, but extending the line of thinking about using `mbedtls_x509_string_to_names`, if we go down that route, we should consider replacing this with `mbedtls_oid_get_numeric_string` directly, which seems only require oid.c ... again at that point, I think it reduces the test burden on us, with the expectation that the baseline function of just decoding OID to strings correctly is handled by mbedtls tests.

CodeMonkeyLeet (commented):
Pretty much looks ready to me, but the thought occurred to me that maybe we should just be using the MbedTLS implementations directly if we're re-implementing against their constraints anyway. Let me know what you think.

enclave/crypto/openssl/cert.c
mingweishih: Just checked the implementation of `mbedtls_x509_string_to_names`, which is defined as follows.
```
int mbedtls_x509_string_to_names(mbedtls_asn1_named_data **head, const char *name);
```

The case of using the function may not be straightforward given that the function not only parses the string but also sets the Mbed TLS-specific data structure (`mbedtls_asn1_named_data`).

enclave/crypto/openssl/cert.c
mingweishih: Code was removed.

enclave/crypto/openssl/cert.c
mingweishih: Update the code to use `mbedtls_oid_get_numeric_string`, which is more suitable than `mbedtls_x509_string_to_names`.

mingweishih:
@CodeMonkeyLeet 

> Pretty much looks ready to me, but the thought occurred to me that maybe we should just be using the MbedTLS implementations directly if we're re-implementing against their constraints anyway. Let me know what you think.

After some investigations, the conclusion is
- Keep the `X509_parse_name` (and the tests). This is because using `mbedtls_x509_string_to_names` is not straightforward.
- Update the `_decode_oid_to_str` to use `mbedtls_oid_get_numeric_string` in Mbed TLS. The tests are hence removed.

mingweishih:
bors try

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,405 @@
...
+             * The check of if(d - data == OE_X509_MAX_NAME_SIZE)
+             * below ensures that len should never exceed OE_X509_MAX_NAME_SIZE.
+             * We do not perform additional checks here.
+             */
+            size_t len = (size_t)(d - data) + 1;
+            if (oe_memcpy_s(value, OE_X509_MAX_NAME_SIZE, data, len) != OE_OK)
+                goto done;
+            value[len - 1] = '\0';
+            if (!X509_NAME_add_entry_by_txt(
+                    name, key, MBSTRING_ASC, (unsigned char*)value, -1, -1, 0))
```
CodeMonkeyLeet, line 87: Just in terms of parity with MbedTLS behavior, I don't think it handles multibyte ASCII strings. I can't find documentation on this, but minimally MbedTLS has to handle UTF-8 strings for ASN ... is there a reason we're using `MBSTRING_ASC` here instead of `MBSTRING_UTF8`?

tests/crypto/cert_tests.c
```diff
@@ -0,0 +1,93 @@
...
+    OE_TEST(name != NULL);
+    X509_NAME_get_text_by_NID(name, NID_commonName, buf, 256);
+    OE_TEST(strcmp(buf, "Open Enclave SDK,OE SDK") == 0);
+    X509_NAME_get_text_by_NID(name, NID_organizationName, buf, 256);
+    OE_TEST(strcmp(buf, "OESDK TLS") == 0);
+    X509_NAME_get_text_by_NID(name, NID_countryName, buf, 256);
+    OE_TEST(strcmp(buf, "US") == 0);
+    X509_NAME_free(name);
+
+    /* Spaces arround the key should not be allowed (i.e., invalid keys) */
```
CodeMonkeyLeet, line 57:
```suggestion
    /* Spaces around the key should not be allowed (i.e., invalid keys) */
```

tests/crypto/cert_tests.c
CodeMonkeyLeet: Yes, if we're keeping the OE specific implementation, we need to add tests to evaluate what we think we support. As mentioned in the above comments, OpenSSL CLI will support multiRDN names (e.g. handling "+"), but MbedTLS does not appear to. The UTF-8 vs MBCS handling behavior is also unclear to me.

enclave/crypto/openssl/cert.c
CodeMonkeyLeet: I think that's a fair assessment barring any deeper investigations of trying to map the linked list of `mbedtls_asn1_named_data` objects to `ASN1_OBJECT` for `X509_NAME_add_entry_by_OBJ` 

It would probably be helpful to add the implications of what this means for a user in the comments (e.g. unlike the parsing behavior users of OpenSSL CLI might be familiar with, we do not support multi-RDN names with "+" in them.) There's a broader compat question since OpenSSL seems to expect a different input format in their CLI app (e.g. `/type0=value0/type1=value1/type2=...`) and we might want to track that as an open and move on.

CodeMonkeyLeet (approved):
Thanks for all the work on this, Ming-wei, I'm still a bit wary of the extra DN name parsing and have some testing comments on that, but I'm approving with the expectation that we can clean that up more later.

enclave/crypto/openssl/cert.c
mingweishih: Added comments around the difference between OpenSSL CLI.

enclave/crypto/openssl/cert.c
mingweishih: Updated to use MBSTRING_UTF8 instead.

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,405 @@
...
+        }
+        else if (!in_tag && (*c == ',' || c == end))
+        {
+            /*
+             * The check of if(d - data == OE_X509_MAX_NAME_SIZE)
+             * below ensures that len should never exceed OE_X509_MAX_NAME_SIZE.
+             * We do not perform additional checks here.
+             */
+            size_t len = (size_t)(d - data) + 1;
+            if (oe_memcpy_s(value, OE_X509_MAX_NAME_SIZE, data, len) != OE_OK)
```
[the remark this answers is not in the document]
mingweishih: Remove the redundant code.

enclave/crypto/openssl/cert.c
```diff
@@ -0,0 +1,405 @@
...
+ * "KEY1=VALUE1,KEY2=VALUE2,KEY3=VALUE3...". The implementation is based
+ * on the mbedtls_x509_string_to_names from Mbed TLS.
+ */
+X509_NAME* X509_parse_name(const char* name_string)
+{
+    const char* s = name_string;
+    const char* c = s;
+    const char* end = s + strlen(s);
+    int in_tag = 1;
+    char key[OE_X509_MAX_NAME_SIZE];
```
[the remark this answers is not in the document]
mingweishih: Given that `X509_NAME_add_entry_by_txt` requires two input strings (key and value), keeping `key` and `data` separately should simplify the implementation (assume we follow the Mbed TLS's version).

tests/crypto/cert_tests.c
mingweishih: Added positive tests for the UTF-8 testing and no multiRDN support, and a negative test for escaping characters other than commas.

mingweishih:
bors try

CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [Unreleased][Unreleased_log]
 --------------
 
+## Breaking Changes
+- liboecryptombed is now called liboecryptombedtls and will no longer be automatically included as a link dependency when linking liboeenclave in CMake.
+     - The `openenclave-config.cmake` and `openenclave-lvi-mitigation-config.cmake` will not specify the renamed liboecryptombedtls as a `PUBLIC` link requirement for liboeenclave.
+     - Enclave apps that build with CMake and use the Open Enclave cmake configurations must now explicitly include `openenclave::oecryptombedtls` when linking `openenclave::oeenclave`.
+     - See the [CMakeLists.txt in the helloworld sample](samples/helloworld/enclave/CMakeLists.txt#L32) for an example.
+     - This change does not currently affect enclave apps relying on pkgconfig.
```
radhikaj, line 18: why does it not affect pkgconfig?

radhikaj (approved):
Thanks Ming-Wei! The samples and tests look good. Thanks for moving code common to the host and enclave into the common directory.

CHANGELOG.md
mingweishih: In pkgconfig, we already explicitly link the `oecryptombedtls` (https://github.com/openenclave/openenclave/blob/master/pkgconfig/CMakeLists.txt#L82). However, we will need to update the pkgconfig once we support samples with `oecryptoopenssl`.

PR #3400, opened 2020-08-17
Extend attested tls server for autoscaling deployments

vtikoo:
### Summary
The attested tls server if deployed as a ReplicaSet in kubernetes, treats load balancer pings as a failed TLS handshake and exits. The changes proposed here extend the tls server to be more suitable for such a deployment.

### DIFF
--- a/samples/attested_tls/README.md
+++ b/samples/attested_tls/README.md
@@ -93,3 +93,10 @@ mkdir build && cd build
 cmake ..
 make run
 ```
+### Running attested TLS server in loop
+By default the server exits after completing a TLS session with a client. `-server-in-loop` run-time option changes this behavior to allow the TLS server to handle multiple client requests.
+```bash
+./server/host/tls_server_host ./server/enc/tls_server_enc.signed -port:12341 -server-in-loop
+or
+make run-server-in-loop
+```
--- a/samples/attested_tls/server/enc/server.cpp
+++ b/samples/attested_tls/server/enc/server.cpp
@@ -20,7 +20,7 @@
 
 extern "C"
 {
-    int setup_tls_server(char* server_port);
+    int setup_tls_server(char* server_port, bool keep_server_up);
 };
 
 #define MAX_ERROR_BUFF_SIZE 256
@@ -123,14 +123,18 @@ exit:
 int handle_communication_until_done(
     mbedtls_ssl_context* ssl,
     mbedtls_net_context* listen_fd,
-    mbedtls_net_context* client_fd)
+    mbedtls_net_context* client_fd,
+    bool keep_server_up)
 {
     int ret = 0;
     int len = 0;
 
 waiting_for_connection_request:
 
-    if (ret != 0)
+    if (ret != 0 &&
+        // ignore EOF errors, which can be caused due to Load Balancers
+        // or health checks
+        ret != MBEDTLS_ERR_SSL_CONN_EOF)
     {
         mbedtls_strerror(ret, error_buf, MAX_ERROR_BUFF_SIZE);
         printf("Last error was: %d - %s\n", ret, error_buf);
@@ -140,7 +144,8 @@ waiting_for_connection_request:
     mbedtls_net_free(client_fd);
     mbedtls_ssl_session_reset(ssl);
 
-    printf(TLS_SERVER "Waiting for a client connection request...\n");
+    if (ret != MBEDTLS_ERR_SSL_CONN_EOF)
+        printf(TLS_SERVER "Waiting for a client connection request...\n");
     if ((ret = mbedtls_net_accept(listen_fd, client_fd, NULL, 0, NULL)) != 0)
     {
         char errbuf[512];
@@ -151,20 +156,19 @@ waiting_for_connection_request:
             errbuf);
         goto done;
     }
-    printf(
-        TLS_SERVER
-        "mbedtls_net_accept returned successfully.(listen_fd = %d) (client_fd "
-        "= %d) \n",
-        listen_fd->fd,
-        client_fd->fd);
 
     // set up bio callbacks
     mbedtls_ssl_set_bio(
         ssl, client_fd, mbedtls_net_send, mbedtls_net_recv, NULL);
 
-    printf(TLS_SERVER "Performing the SSL/TLS handshake...\n");
     while ((ret = mbedtls_ssl_handshake(ssl)) != 0)
     {
+        // Load balancer health-check pings can cause EOF errors
+        // Ignore the error, and wait for client to send request
+        if (ret == MBEDTLS_ERR_SSL_CONN_EOF)
+        {
+            goto waiting_for_connection_request;
+        }
         if (ret != MBEDTLS_ERR_SSL_WANT_READ &&
             ret != MBEDTLS_ERR_SSL_WANT_WRITE)
         {
@@ -269,14 +273,15 @@ waiting_for_connection_request:
     }
 
     ret = 0;
-    // comment out the following line if you want the server in a loop
-    // goto waiting_for_connection_request;
+
+    if (keep_server_up)
+        goto waiting_for_connection_request;
 
 done:
     return ret;
 }
 
-int setup_tls_server(char* server_port)
+int setup_tls_server(char* server_port, bool keep_server_up)
 {
     int ret = 0;
     oe_result_t result = OE_FAILURE;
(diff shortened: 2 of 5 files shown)


samples/attested_tls/README.md
```diff
@@ -93,3 +93,17 @@ mkdir build && cd build
...
 make run
 ```
+### Running attested tls sample on Kubernetes
+```bash
+make CLOUD_DEPLOYMENT=true
+make run
+```
+`CLOUD_DEPLOYMENT` flag modifies behavior of the test to make it
+suitable for scenario where the server is deployed in the form of a replicated set and is accessed via a load balancer.
+An example of this, would be as a Service on a Kubernetes cluster.
```
dthaler, line 103: Fix grammar
```suggestion
The `CLOUD_DEPLOYMENT` flag modifies the behavior of the test to make it
suitable for the scenario where the server is deployed in the form of a replicated set and is accessed via a load balancer.
An example of this would be as a Service on a Kubernetes cluster.
```

samples/attested_tls/README.md
```diff
@@ -93,3 +93,17 @@ mkdir build && cd build
...
+
+The flag modifies the behavior in two ways:
+1. Handle load balancer pings: load balancer pings fail the tls
```
dthaler, line 106:
```suggestion
1. Handle load balancer pings: load balancer pings fail the TLS
```

samples/attested_tls/README.md
```diff
@@ -93,3 +93,17 @@ mkdir build && cd build
...
+handshake with an EOF error. Enabling the flag ignores this error
+and also squelches logs if this has been
```
dthaler, line 108: ... if this has been what?

samples/attested_tls/server/enc/server.cpp
```diff
@@ -127,9 +127,16 @@ int handle_communication_until_done(
 {
     int ret = 0;
     int len = 0;
+# ifdef CLOUD_DEPLOYMENT
```
dthaler, line 130: ifdefs are not as good as runtime options.  Recommend changing to use a command line argument to the sample.

anakrish: I agree with Dave.
To go even further, can you change the sample so that it just works for both the cases.
Can we not handle `MBEDTLS_ERR_SSL_CONN_EOF`  always rather than only enabling it for the autoscaling case?

samples/attested_tls/README.md
vtikoo: Oops, thanks for catching this!

samples/attested_tls/server/enc/server.cpp
vtikoo: @anakrish  I hit this error when the TLS server received AKS load balancer pings. Not sure if there are other cases where this error is reported.

vtikoo: Maybe we can handle the error always, with comments explaining the reasons and assumptions. Would that be acceptable?

samples/attested_tls/README.md
vtikoo: Fixed.

samples/attested_tls/server/enc/server.cpp
anakrish: Yes. Handling that error always would be better.

samples/attested_tls/README.md
```diff
@@ -93,3 +93,17 @@ mkdir build && cd build
 cmake ..
 make run
 ```
+### Running attested tls sample on Kubernetes
```
CodeMonkeyLeet, line 96:
```suggestion
### Running the attested TLS sample on Kubernetes
```

samples/attested_tls/server/enc/Makefile
```diff
@@ -8,6 +8,11 @@ CXXFLAGS=$(shell pkg-config oeenclave-$(CXX_COMPILER) --cflags)
 LDFLAGS=$(shell pkg-config oeenclave-$(CXX_COMPILER) --libs)
 INCDIR=$(shell pkg-config oeenclave-$(C_COMPILER) --variable=includedir)
 
+CLOUD_DEPLOYMENT?=false
+ifeq ($(CLOUD_DEPLOYMENT), true)
+	CXXFLAGS      += -DCLOUD_DEPLOYMENT
```
CodeMonkeyLeet, line 13:
```suggestion
	CXXFLAGS += -DCLOUD_DEPLOYMENT
```

samples/attested_tls/server/enc/server.cpp
vtikoo: EOF errors are now always handled.

vtikoo:
@dthaler @anakrish based on your review suggestions:
- Have removed the ifdef toggle and EOF failures are handled always.
- Added a run-time option `-run-server-in-loop` to keep server up and waiting for more client requests.

samples/attested_tls/server/enc/Makefile
vtikoo: Not using compile time flags as per review suggestion.

samples/attested_tls/server/enc/server.cpp
```diff
@@ -151,20 +156,19 @@ int handle_communication_until_done(
             errbuf);
         goto done;
     }
-    printf(
```
vtikoo, line 154: Have removed this as this makes the logs chatty in case of frequent health checks.

samples/attested_tls/server/enc/server.cpp
```diff
@@ -151,20 +156,19 @@ int handle_communication_until_done(
...
-        "mbedtls_net_accept returned successfully.(listen_fd = %d) (client_fd "
-        "= %d) \n",
-        listen_fd->fd,
-        client_fd->fd);
 
     // set up bio callbacks
     mbedtls_ssl_set_bio(
         ssl, client_fd, mbedtls_net_send, mbedtls_net_recv, NULL);
 
-    printf(TLS_SERVER "Performing the SSL/TLS handshake...\n");
```
vtikoo, line 165: Same reason as above.

samples/attested_tls/server/host/host.cpp
```diff
@@ -43,15 +43,30 @@ int main(int argc, const char* argv[])
...
     int ret = 1;
     char* server_port = NULL;
+    bool keep_server_up = false;
 
     /* Check argument count */
     if (argc != 3)
     {
+        if (argc == 4)
+        {
+            char* loop_option = (char*)"-server-in-loop";
```
dthaler, line 53: why is this char* and not const char*?   strncmp should accept const char* on  Linux, Windows, and anything else that conforms to the posix standard.
```suggestion
            const char* loop_option = "-server-in-loop";
```

You could also just #define LOOP_OPTION "-server-in-loop" and use LOOP_OPTION in the code below.

samples/attested_tls/server/host/host.cpp
```diff
@@ -43,15 +43,30 @@ int main(int argc, const char* argv[])
...
+            size_t loop_option_len = 0;
+            loop_option_len = strlen(loop_option);
+            if (strncmp(argv[3], loop_option, loop_option_len) != 0) {
```
dthaler, line 56: why strncmp and not strcmp?  Won't this match things like -server-in-loophole?  I'm not sure that's desirable.
```suggestion
            if (strcmp(argv[3], loop_option) != 0) {
```

vtikoo: Thanks for catching this bug!

samples/attested_tls/server/host/host.cpp
```diff
@@ -43,15 +45,27 @@ int main(int argc, const char* argv[])
...
+                goto print_usage;
+            }
+            else {
+                keep_server_up = true;
+                goto read_port;
+            }
+        }
     print_usage:
-        printf("Usage: %s TLS_SERVER_ENCLAVE_PATH -port:<port>\n", argv[0]);
+        printf("Usage: %s TLS_SERVER_ENCLAVE_PATH -port:<port> [-server-in-loop]\n", argv[0]);
```
dthaler, line 64: Ok as is, but also consider a precedent where the option name is passed as an argument to guarantee it's identical.
```suggestion
        printf("Usage: %s TLS_SERVER_ENCLAVE_PATH -port:<port> [%s]\n", argv[0], LOOP_OPTION);
```

vtikoo: Thanks, thats a good point!

radhikaj:
Thanks @vtikoo

vtikoo:
bors try

radhikaj:
Bors r+

Issue #3379, opened 2020-08-12
Simplify SGX attestation UUIDs


yentsanglee:
Feedback from @LarryOsterman - Our verifier accepts SGX evidence type. Among all SGX attestation UUIDs defined in OE SDK, not sure which one to pick.

Current SGX types:
    * OE_FORMAT_UUID_SGX_ECDSA_P256
    * OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT
    * OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE
    * OE_FORMAT_UUID_SGX_ECDSA_P384
    * OE_FORMAT_UUID_SGX_EPID_LINKABLE
    * OE_FORMAT_UUID_SGX_EPID_UNLINKABLE

shnwc:
I’d assume @LarryOsterman's quotes are all in ECDSA-p256 format. DCAP only generates this type of quotes. ECDSA-Pp384 quote generation is not supported yet. EPID quote can only be generated by another library such as quote-ex, and before generating an EPID quote, a platform needs to be configured to join and EPID group.

To verify an ECDSA-p256 quotes, application software can call oe_verify_evidence() with format_id set to OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE.

LarryOsterman:
@shnwc How would I as a user of the API know which quote format I have? There are 6 different options, and to the best of my knowledge, the only way of determining which of the options is the correct one to use is to parse the evidence to determine what the format of the evidence is. But the whole point of this API is that I shouldn't have to parse the evidence.

If it is possible for the attestation API to distinguish between various options by looking at the evidence, then it should not be presented to the user as a choice. All it does is to introduce a failure point for the customer AND it becomes what the Windows team calls a "call generator" - in other words, an issue which is highly likely to cause the customer to request support.

It's probably safe to know that the caller knows that this evidence came from an SGX enclave (SGX). It's also probably safe to know that the caller knows that the evidence is of the form of an SGX Quote or an SGX Report. That reduces the set of 6 options to 4 (I'm not 100% sure what an SGX_EPID is - if it's an expression of a type 2 or 3 quote validation data, then that's another option that a caller shouldn't have to know.

However it is NOT reasonable to require the caller to know which encryption algorithm was used to sign the attestation evidence. That is a parameter which is not under the control of the caller, and thus they should not be required to know it.

yentsanglee:
[...]

PR #3402, opened 2020-08-17
Update SGX format IDs for support of OE reports and raw SGX quotes

shnwc:
* Update SGX format IDs to cover all OE reports and raw SGX quotes
* Remove OE report header from evidence generated by `oe_get_evidence()`
* Remove unsupported format ID `OE_FORMAT_UUID_SGX_ECDSA_P384`

### DIFF
--- a/docs/DesignDocs/Attestation_API_Proposal.md
+++ b/docs/DesignDocs/Attestation_API_Proposal.md
@@ -1,10 +1,10 @@
-Proposal of OE SDK Attestation Public and Plugin API for V0.10 Release
+Proposal of OE SDK Attestation Public and Plugin API for V0.1x Release
 ====
 
 # Introduction
 
 This document describes the proposed OE SDK attestation public and plugin API
-for V0.10 release, as an evolution of the experimental attestation API
+for V0.1x release, as an evolution of the experimental attestation API
 implemented in the
 [OE SDK V0.9 release](https://github.com/openenclave/openenclave/tree/v0.9.x).
 
@@ -30,7 +30,7 @@ releases.
 Note: in the RATS model as described in the
 [[RATS Arch]](https://tools.ietf.org/html/draft-ietf-rats-architecture)
 document, there are three roles: an Attester, a Verifier, and a Relying Party.
-The OE SDK attestation stack V0.10 does not try to provide a complete implementation of
+The OE SDK attestation stack V0.1x does not try to provide a complete implementation of
 these roles, nor try to provide means for secure communication between them.
 The OE SDK attestation stack implements basic functionalities for evidence
 generation and verification, and exposes these functionalities to applications
@@ -68,7 +68,7 @@ Use cases for evidence generation:
   document.
 - Get evidence in a globally unique format, optionally along with a set of
 endorsements.
-  - Function `oe_result_t oe_get_evidence(const oe_uuid_t* format_id, uint32_t flags, const oe_claim_t* custom_claims, size_t custom_claims_length, const void* opt_params, size_t opt_params_size, uint8_t** evidence_buffer, size_t* vidence_buffer_size, uint8_t** endorsements_buffer, size_t* endorsements_buffer_size)`,
+  - Function `oe_result_t oe_get_evidence(const oe_uuid_t* format_id, uint32_t flags, const oe_claim_t* custom_claims_buffer, size_t custom_claims_buffer_length, const void* opt_params, size_t opt_params_size, uint8_t** evidence_buffer, size_t* vidence_buffer_size, uint8_t** endorsements_buffer, size_t* endorsements_buffer_size)`,
     declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h).
   - Upon successful evidence generation, `oe_get_evidence()` returns the
   generated evidence in a dynamically allocated buffer. It also has the option
@@ -83,7 +83,7 @@ endorsements.
   section of the
   [[Design Notes]](https://github.com/openenclave/openenclave/pull/2801)
   document for a discussion of the semantics
-  of the `custom_claims` and `opt_params` parameters.
+  of the `custom_claims_buffer` and `opt_params` parameters.
 - Free a dynamically allocated evidence buffer.
   - Function `oe_result_t oe_free_evidence(uint8_t* evidence_buffer)`,
   declared in header file [openenclave/attestation/plugin.h](https://github.com/openenclave/openenclave/blob/v0.9.x/include/openenclave/attestation/plugin.h)
@@ -146,7 +146,7 @@ single header file [openenclave/attestation/plugin.h](https://github.com/openenc
 - For SGX, the globally unique UUIDs identify the evidence formats.
 There is no need of local / remote flag.
 
-## Proposed OE SDK V0.10 Public API
+## Proposed OE SDK V0.1x Public API
 
 Objectives: address the areas of improvement in the existing public API.
 
@@ -194,7 +194,7 @@ Use cases for evidence generation:
   [[Attestation V3 Update]](CustomAttestation_V3.md)
 - Get evidence in an globally unique format, optionally along with a set of
 endorsements.
-  - Function `oe_result_t oe_get_evidence(const oe_uuid_t* format_id, uint32_t flags, const void* custom_claims, size_t custom_claims_size, const void* opt_params, size_t opt_params_size, uint8_t** evidence_buffer, size_t* evidence_buffer_size, uint8_t** endorsements_buffer, size_t* endorsements_buffer_size)`.
+  - Function `oe_result_t oe_get_evidence(const oe_uuid_t* format_id, uint32_t flags, const void* custom_claims_buffer, size_t custom_claims_buffer_size, const void* opt_params, size_t opt_params_size, uint8_t** evidence_buffer, size_t* evidence_buffer_size, uint8_t** endorsements_buffer, size_t* endorsements_buffer_size)`.
   - The `flags` parameter in the OE SDK V0.9 API release is redefined
   to be a bit-wise parameter. In the current version, there is one bit
   defined:
@@ -368,7 +368,7 @@ These use cases are supported by the attester plugin entry points defined in the
   - The same entry point definition for both attester and verifier plugins.
 - Get evidence in a specified format, optionally along with a set of
 endorsements.
-  - Entry point `oe_result_t (*get_evidence)(oe_attester_t* context, uint32_t flags, const oe_claim_t* custom_claims, size_t custom_claims_length, const void* opt_params, size_t opt_params_size, uint8_t** evidence_buffer, size_t* evidence_buffer_size, uint8_t** endorsements_buffer, size_t* endorsements_buffer_size)`.
+  - Entry point `oe_result_t (*get_evidence)(oe_attester_t* context, uint32_t flags, const oe_claim_t* custom_claims_buffer, size_t custom_claims_buffer_length, const void* opt_params, size_t opt_params_size, uint8_t** evidence_buffer, size_t* evidence_buffer_size, uint8_t** endorsements_buffer, size_t* endorsements_buffer_size)`.
 - Free a dynamically allocated evidence buffer.
   - Entry point `oe_result_t (*free_evidence)(oe_attester_t* context, uint8_t* evidence_buffer)`.
 - Free a dynamically allocated endorsements buffer.
@@ -412,7 +412,7 @@ To setup this plugin, the application software invokes public API function
 `oe_sgx_plugin_verifier()` to get an `oe_verifier_t` structure, and calls public
 API function `oe_register_verifier()` with this structure.
 
-## Proposed OE SDK V0.10 Plugin API
+## Proposed OE SDK V0.1x Plugin API
 
 ### Objectives
 
@@ -438,9 +438,9 @@ by the attester plugin entry points defined in the `oe_attester_t` structure:
   - The same entry point definition as in the OE SDK V0.9 release.
 - Get evidence in an globally unique format, optionally along with a set of
 endorsements.
-  - Entry point `oe_result_t (*get_evidence)(oe_attester_t* context, const void* custom_claims, size_t custom_claims_size, const void* opt_params, size_t opt_params_size, uint8_t** evidence_buffer, size_t* evidence_buffer_size, uint8_t** endorsements_buffer, size_t* endorsements_buffer_size)`.
+  - Entry point `oe_result_t (*get_evidence)(oe_attester_t* context, const void* custom_claims_buffer, size_t custom_claims_buffer_size, const void* opt_params, size_t opt_params_size, uint8_t** evidence_buffer, size_t* evidence_buffer_size, uint8_t** endorsements_buffer, size_t* endorsements_buffer_size)`.
   - The legacy `flags` parameter in the OE SDK V0.9 release is removed.
-  - The optional custom claims are input as a flat buffer in `custom_claims`.
+  - The optional custom claims are input as a flat buffer in `custom_claims_buffer`.
   - The output evidence and endorsements data must be held in dynamically
   allocated buffers
   in trusted enclave memory. These buffers will be freed by the plugin framework
@@ -478,7 +478,7 @@ OE SDK framework:
   - The same definition as in the OE SDK V0.9 release, except that this function
   is not part of the public API, and the function name is changed.
 
-In the OE SDK V0.10 release, attester plugins are managed internally between
+In the OE SDK V0.1x release, attester plugins are managed internally between
 the OE SDK framework and plugin libraries, without exposing details to
 application software. As explained in the
 [[Design Notes]](https://github.com/shnwc/openenclave/blob/master/docs/DesignDocs/NotesOnAttestationAPI.md#options-for-enclave-side-plugin-library-initialization)
@@ -528,7 +528,7 @@ Use cases for a verifier plugin to interact with the OE SDK framework:
   - The same definition as in the OE SDK V0.9 release, except that this function
   is not part of the public API, and the function name is changed.
 
-In the OE SDK V0.10 release, initialization of verifier plugin libraries is
+In the OE SDK V0.1x release, initialization of verifier plugin libraries is
 triggered with the application call to `oe_verifier_initialize()`.
 The initialization function of every linked library enumerates its supported
 evidence formats and registers their plugins with the OE SDK framework via
@@ -555,18 +555,18 @@ implementation of the public API function `oe_sgx_plugin_verifier()`. This
 verifier plugin implementation is linked to both the OE SDK enclave-side and host-side
 static libraries.
 
-## Proposed V0.10 Implementation
+## Proposed V0.1x Implementation
 
-The OE SDK V0.10 release will have enclave-side SGX attester plugins for
+The OE SDK V0.1x release will have enclave-side SGX attester plugins for
 generation of SGX evidence in local, ECDSA, and EPID formats. These formats
 are defined as:
 - `OE_FORMAT_UUID_SGX_LOCAL_ATTESTATION`: for SGX local attestation.
-- `OE_FORMAT_UUID_SGX_ECDSA_P256`: for SGX ECDSA-p256 evidence format
-  - Note: this is the same as existing `OE_SGX_PLUGIN_UUID`.
+- `OE_FORMAT_UUID_SGX_ECDSA`: for SGX ECDSA-p256 evidence format
+  - Note: this is the same as `OE_SGX_PLUGIN_UUID` in V0.9.
 - `OE_FORMAT_UUID_SGX_EPID_LINKABLE`: for SGX linkable EPID evidence format.
 - `OE_FORMAT_UUID_SGX_EPID_UNLINKABLE`: for SGX unlinkable EPID evidence format.
 
-These plugins will be implemented by a single set of plugin libraries,
+These attester plugins will be implemented by a single set of plugin libraries,
 composed of:
 - One enclave-side SGX evidence generation plugin library.
   - It implements the OE SDK public API function
@@ -576,9 +576,12 @@ composed of:
   - It implements OCALLs for the enclave-side library to access
   host-side services for SGX quote generation.
 
-The OE SDK V0.10 release will have enclave-side plugins for verification of
-SGX local and ECDSA evidence, and host-side plugins for verification of
-SGX ECDSA evidence. These plugins will be implemented by a single set of
+The OE SDK V0.1x release will have enclave-side verifier plugins for
+verification of `OE_FORMAT_UUID_SGX_LOCAL_ATTESTATION` and
+`OE_FORMAT_UUID_SGX_ECDSA` evidence,
+and host-side verifier plugins for verification of
+`OE_FORMAT_UUID_SGX_ECDSA` evidence.
+These plugins will be implemented by a single set of
 plugin libraries, composed of:
 - One enclave-side SGX evidence verification plugin library.
   - It implements the enclave-side OE SDK public API function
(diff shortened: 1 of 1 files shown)


docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -592,62 +593,106 @@ plugin libraries, composed of:
...
-`oe_get_report()` and raw SGX ECDSA quotes generated by the Intel SGX SDK.
-These two evidence formats are identified by their UUIDs:
-- `OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT`: for SGX ECDSA report generated
-by `oe_get_report()`.
-- `OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE`: for raw SGX ECDSA quote generated
-by the Intel SGX SDK.
+for verification of OE reports for SGX generated by the legacy API function
+`oe_get_report()` and raw SGX quotes generated by the Intel SGX SDK
+DCAP library. These two evidence formats are identified by two UUIDs:
+- `OE_FORMAT_UUID_SGX_REPORT`: for all OE reports generated
```
yentsanglee, line 599: Can we name this OE_FORMAT_UUID_OE_REPORT? SGX_REPORT implies local report.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -592,62 +593,106 @@ plugin libraries, composed of:
...
+`OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID` bit in its `flags` parameter is set.
 * `oe_report_header` is the OE report header of type `oe_report_header_t`.
 * `hash` is the SHA256 hash of the custom claims held in a flat buffer
 `custom_claims`.
-* An SGX report (`SGX_report()`) or quote (`SGX_*_quote()`) contains a
+* An SGX report (`SGX_report()`) or quote (`SGX_*quote()`) embeds a
 flat buffer of 64 bytes for its SGX report data field. Depending on the format,
 this field holds either the `hash` of the custom claims, or the custom claims
 directly.
+* Note: for all evidence generated by API `oe_get_evidence()`, the
```
yentsanglee, line 685: This will not happen in release 0.10.1. The verification support of the new format should be implemented first before changing the evidence generation.

yentsanglee:
As mentioned in issue #3379, Doxygen-style UUID descriptions in evidence.h should be included in the fix of #3379.

docs/DesignDocs/Attestation_API_Proposal.md
dthaler: No, dropping SGX would imply it's TEE agnostic.

shnwc:
> As mentioned in issue #3379, Doxygen-style UUID descriptions in evidence.h should be included in the fix of #3379.

Do you mean only implementation of this PR will fix 3379? Can remove "Fixes issue #3379" in the PR text.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Sure, agreed.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: > This will not happen in release 0.10.1. The verification support of the new format should be implemented first before changing the evidence generation.

Do you mean we should not implement it in 0.10.1? After approval of this PR, there will be another code PR that implements all the proposed changes, targeted for 0.10.1.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: > No, dropping SGX would imply it's TEE agnostic.

@dthaler the legacy API `oe_get_report()` only supports two report types for SGX: SGX report for local attestation, and SGX ECDSA quote for remote attestation.

How about calling it `OE_FORMAT_UUID_SGX_OE_REPORT`?

dthaler: OE is already in the beginning, so it looks bad to see it twice in the same define.

dthaler: It sounds like you're saying that oe_get_report is a misnomer since it also supports "SGX ECDSA quotes for remote attestation" which are not usually called "reports", is that correct?   (If so I guess it's a good thing we can obsolete this name eventually.)   What about something like OE_FORMAT_SGX_REPORT_OR_QUOTE?

shnwc: > It sounds like you're saying that oe_get_report is a misnomer since it also supports "SGX ECDSA quotes for remote attestation" which are not usually called "reports".

In SGX terminology, "SGX report" is for local attestation, and "SGX quote" is for remote attestation. QE converts reports to quotes. So it becomes confusing when `oe_get_report()` produces "report" / evidence for both local and remote attestation.

I think `OE_FORMAT_SGX_REPORT_OR_QUOTE` could be confused with `OE_FORMAT_SGX_QUOTE`.

How about calling it `OE_FORMAT_UUID_LEGACY_REPORT`?

docs/DesignDocs/Attestation_API_Proposal.md
yentsanglee: If we change the evidence structure generated by oe_get_evidence(), apps that built with older OE SDK won't be able to call oe_verify_evidence() to verify evidence. I'd suggest we start by allowing oe_verify_evidence() to verify new evidence structure. oe_get_evidence() will follow in the next release.

shnwc: The attestation plugin API is still experimental, that should allow us more freedom to change the definition without having to maintain backward compatibility, especially when oe_attestation_header_t is not exposed as a public structure.

So I still think we can change both ends in a single step. Apps will need to be rebuilt when the new SDK release is used.

In any case, to remove oe_report_header, the version number `OE_ATTESTATION_HEADER_VERSION` will need to be bumped to 3. This allows `oe_verify_evidence()` to detect out-dated evidence and raise error. Or if backward compatibility is still desirable, `oe_verify_evidence()` can keep the logic for version 2 headers.

yentsanglee: Yes, the version number will need to be bumped to 3. I still prefer 2-phase plan but am open to your plan. Let's discuss in the meeting. There are only a few days left before cut-off time for release 0.10.1. Is your code PR almost ready?

yentsanglee:
> > As mentioned in issue #3379, Doxygen-style UUID descriptions in evidence.h should be included in the fix of #3379.
> 
> Do you mean only implementation of this PR will fix 3379? Can remove "Fixes issue #3379" in the PR text.

Code change is needed to address the concern in #3379. (Removed link to #3379)

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -592,62 +593,106 @@ plugin libraries, composed of:
...
+- `OE_FORMAT_UUID_SGX_QUOTE`: for all raw SGX ECDSA quote generated
+by the Intel SGX SDK DCAP library.
+  - Note: currently, the DCAP library only generates ECDSA-p256 quotes.
+  - In the future, if additional quote types are supported by the DCAP library,
+  the same UUID would be used for verification of these new types of quotes.
+
+There is no plan currently to implement verifier plugins for verification of
+EPID evidence. Verification of SGX EPID quotes is not supported by
+the Intel SGX SDK. In all existing SGX EPID solutions, EPID quotes are sent to
+Intel Attestation Service (IAS)) backend for verification.
```
dthaler, line 611:
```suggestion
Intel Attestation Service (IAS) backend for verification.
```

shnwc:
@dthaler @yentsanglee @CodeMonkeyLeet @MWShan please review the latest commit.

While in the SIG-attestation meeting we discussed about the need for multiple UUIDs for different types of legacy OE reports and SGX quotes, as explained in the latest commit, I see that our plan is for the attestation API function `oe_verify_evidence()` to only support a minimum set of legacy OE reports and SGX quotes as needed. It's not our intention for `oe_verify_evidence()` to support verification of all possible types of legacy OE reports and SGX quotes.

Based on this understanding, only two UUIDs for legacy OE report / SGX quote will be supported for verification: `OE_FORMAT_UUID_LEGACY_REPORT_REMOTE` and `OE_FORMAT_UUID_SGX_QUOTE_ECDSA_P256`. Let me know if you think more types should be supported.

yentsanglee:
I am good with `OE_FORMAT_UUID_LEGACY_REPORT_REMOTE` (assuming the local counterpart is named `OE_FORMAT_UUID_LEGACY_REPORT_LOCAL`).

For the _P256 suffix in SGX quote UUIDs, I think this is a decision that SGX team has to make. If the decision is that one verifier cannot verify all ECDSA quotes, then OE SDK defines them accordingly. For this coming release though, can we drop the _P256 suffix? We can add the suffix when _P384 is ready.

shnwc:
> I am good with `OE_FORMAT_UUID_LEGACY_REPORT_REMOTE` (assuming the local counterpart is named `OE_FORMAT_UUID_LEGACY_REPORT_LOCAL`).
> 
> For the _P256 suffix in SGX quote UUIDs, I think this is a decision that SGX team has to make. If the decision is that one verifier cannot verify all ECDSA quotes, then OE SDK defines them accordingly. For this coming release though, can we drop the _P256 suffix? We can add the suffix when _P384 is ready.

Agree it's fine to remove suffix `_P256` in the UUIDs, since only ECDSA_p256 is supported right now. When ECDSA_p384 is eventually added, we can add new UUID with a suffix for the new format.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -591,63 +594,119 @@ plugin libraries, composed of:
...
+the Intel SGX SDK DCAP and quote-ex libraries. Note: the plan is for
+the attestation API function `oe_verify_evidence()` to only support a minimum
+set of legacy OE reports and SGX quotes as needed. There is no intention for
+`oe_verify_evidence()` to support verification of all possible types of legacy
+OE reports and SGX quotes.
+
+- `OE_FORMAT_UUID_LEGACY_REPORT_REMOTE`
+  - For OE reports generated by the legacy API function `oe_get_report()`
+  with the `OE_REPORT_FLAGS_REMOTE_ATTESTATION` flag.
+  - Note: no plan to provide verifier plugin for verification of OE reports
```
CodeMonkeyLeet, line 614: I understand that the primary focus of the current work is to enable the attestation service scenarios, which are necessarily remote attestation. However, I would strongly suggest planning to implement verifier plugin for the legacy formats as part of the deprecation plan for the OE legacy APIs. 

i.e. it's probably cheaper to implement the verification plugin wrapper, deprecate the old `oe_get_report()` API and then removing the legacy verification plugin, than it is to continue to support the old reporting APIs.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -591,63 +594,119 @@ plugin libraries, composed of:
...
+  generated by the quote-ex library with other algorithm IDs (e.g. EPID quotes).
 
 For evidence data in format `OE_FORMAT_UUID_SGX_LOCAL_ATTESTATION`
-or `OE_FORMAT_UUID_SGX_ECDSA_P256`, custom claims buffer
-(if not empty) is attached to the end of the evidence, the list of claims returned
-by `oe_verify_evidence()` will contain one claim named
+or `OE_FORMAT_UUID_SGX_ECDSA_P256`, the custom claims buffer
+(if not empty) is attached to the end of the evidence. The list of claims
+returned by `oe_verify_evidence()` will contain one claim named
 `OE_CLAIM_CUSTOM_CLAIMS` if the evidence data contains a non-empty custom
```
CodeMonkeyLeet, line 630:
```suggestion
`OE_CLAIM_CUSTOM_CLAIMS_BUFFER` if the evidence data contains a non-empty custom
```

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -591,63 +594,119 @@ plugin libraries, composed of:
...
-| `OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT` | `oe_report_header \|\| SGX_ECDSA_quote(custom_claims)` |
-| `OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE` | `SGX_ECDSA_quote(custom_claims)` |
-| `OE_FORMAT_UUID_SGX_EPID_LINKABLE` | `SGX_EPID_linkable_quote(custom_claims)` |
-| `OE_FORMAT_UUID_SGX_EPID_UNLINKABLE` | `SGX_EPID_unlinkable_quote(custom_claims)` |
+| `OE_FORMAT_UUID_SGX_LOCAL_ATTESTATION` | `[ oe_attestation_header ] \|\| SGX_report(hash) \|\| custom_claims` |
+| `OE_FORMAT_UUID_SGX_ECDSA_P256` | `[ oe_attestation_header ] \|\| SGX_ECDSA_p256_quote(hash) \|\| custom_claims` |
+| `OE_FORMAT_UUID_SGX_EPID_LINKABLE` | `[ oe_attestation_header ] \|\| SGX_EPID_linkable_quote(custom_claims)` |
+| `OE_FORMAT_UUID_SGX_EPID_UNLINKABLE` | `[ oe_attestation_header ] \|\| SGX_EPID_unlinkable_quote(custom_claims)` |
+| `OE_FORMAT_UUID_LEGACY_REPORT_REMOTE` | `oe_report_header (for remote attestation) \|\| SGX_ECDSA_p256_quote(custom_claims)` |
+| `OE_FORMAT_UUID_SGX_QUOTE_ECDSA_P256` | `SGX_ECDSA_p256_quote(custom_claims)` |
```
CodeMonkeyLeet, line 684: Per the SIG attestation meeting, I thought this was going to be called 
`OE_FORMAT_UUID_RAW_SGX_ECDSA_P256` ... as it stands, I'm not sure that the "quote" modifier accurately distinguishes `OE_FORMAT_UUID_SGX_QUOTE_ECDSA_P256` from `OE_FORMAT_UUID_SGX_ECDSA_P256`

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -591,63 +594,119 @@ plugin libraries, composed of:
...
-| `OE_FORMAT_UUID_SGX_ECDSA_P256` | `oe_report_header \|\| SGX_ECDSA_quote(hash) \|\| custom_claims` |
-| `OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT` | `oe_report_header \|\| SGX_ECDSA_quote(custom_claims)` |
-| `OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE` | `SGX_ECDSA_quote(custom_claims)` |
-| `OE_FORMAT_UUID_SGX_EPID_LINKABLE` | `SGX_EPID_linkable_quote(custom_claims)` |
-| `OE_FORMAT_UUID_SGX_EPID_UNLINKABLE` | `SGX_EPID_unlinkable_quote(custom_claims)` |
+| `OE_FORMAT_UUID_SGX_LOCAL_ATTESTATION` | `[ oe_attestation_header ] \|\| SGX_report(hash) \|\| custom_claims` |
+| `OE_FORMAT_UUID_SGX_ECDSA_P256` | `[ oe_attestation_header ] \|\| SGX_ECDSA_p256_quote(hash) \|\| custom_claims` |
+| `OE_FORMAT_UUID_SGX_EPID_LINKABLE` | `[ oe_attestation_header ] \|\| SGX_EPID_linkable_quote(custom_claims)` |
+| `OE_FORMAT_UUID_SGX_EPID_UNLINKABLE` | `[ oe_attestation_header ] \|\| SGX_EPID_unlinkable_quote(custom_claims)` |
+| `OE_FORMAT_UUID_LEGACY_REPORT_REMOTE` | `oe_report_header (for remote attestation) \|\| SGX_ECDSA_p256_quote(custom_claims)` |
```
CodeMonkeyLeet, line 683: Per SIG attestation meeting, I thought we were also going to define:
* `OE_FORMAT_UUID_LEGACY_REPORT_LOCAL`
* `OE_FORMAT_UUID_SGX_QUOTE_EPID_LINKABLE` (would prefer `OE_FORMAT_UUID_RAW_SGX_EPID_LINKABLE`)
* `OE_FORMAT_UUID_SGX_QUOTE_EPID_UNLINKABLE` (would prefer `OE_FORMAT_UUID_RAW_SGX_EPID_UNLINKABLE`)

Independent of the presence of attester/verifier plugin implementation, it sounded from the meeting like these needed to be pre-defined (effectively UUIDs reserved and published by Intel) and should be documented as such.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: The only legacy OE report format is the output of `oe_get_report()` for SGX local attestation, and this will not one of the attestation service scenarios. We already have the legacy API `oe_verify_report()` for it.

I think that, when the plugin APIs are out of experimental, it's better to deprecate `oe_get_report()` and `oe_verify_report()` soon, instead of adding a plugin for OE report for SGX local attestaion.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Thanks!

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Agreed the two names could be hard to distinguish. How about `OE_FORMAT_UUID_RAW_SGX_QUOTE_ECDSA`? "quote" is a widely used SGX term.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Sorry I did not get as clear an agreement from the meeting. But I'm fine adding the names to the header file with comments that they are not currently supported and there is no plan to add their support. Along the same line, should `OE_FORMAT_UUID_SGX_ECDSA_P384` be listed too?

@dthaler @yentsanglee @MWShan what's your opinion?

dthaler: Only add UUIDs for which there exists code in OE to consume them.  So until there is code that uses P384, don't add the UUID to the header.

docs/DesignDocs/Attestation_API_Proposal.md
yentsanglee: I can accept `OE_FORMAT_UUID_RAW_SGX_QUOTE_ECDSA`. No P256 for now please.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: > Only add UUIDs for which there exists code in OE to consume them. So until there is code that uses P384, don't add the UUID to the header.

Following the same reasoning, `OE_FORMAT_UUID_LEGACY_REPORT_LOCAL`, `OE_FORMAT_UUID_SGX_QUOTE_EPID_LINKABLE`, and `OE_FORMAT_UUID_SGX_QUOTE_EPID_UNLINKABLE` will not be added since there is no code in OE SDK to consume them.

yentsanglee: > > Only add UUIDs for which there exists code in OE to consume them. So until there is code that uses P384, don't add the UUID to the header.
> 
> Following the same reasoning, `OE_FORMAT_UUID_LEGACY_REPORT_LOCAL`, `OE_FORMAT_UUID_SGX_QUOTE_EPID_LINKABLE`, and `OE_FORMAT_UUID_SGX_QUOTE_EPID_UNLINKABLE` will not be added since there is no code in OE SDK to consume them.

`OE_FORMAT_UUID_LEGACY_REPORT_LOCAL` (used to be `OE_FORMAT_UUID_SGX_LOCAL_ATTESTATION`) is used in the OE SDK code and[ local attestation sample](https://github.com/openenclave/openenclave/blob/master/samples/local_attestation/common/attestation.cpp)

shnwc: The proposed new `OE_FORMAT_UUID_LEGACY_REPORT_LOCAL` is different than the existing `OE_FORMAT_UUID_SGX_LOCAL_ATTESTATION`. The former format is generated by `oe_get_report()` for local attestation, and is prefixed with an legacy `oe_report_header_t`. It has no support in `oe_verify_evidence()`. The latter is generated by `oe_get_evidence()` and is consumed by `oe_verify_evidence()`.

The raw EPID quotes generated directly by the quote-ex library (not the EPID evidence generated by `oe_get_evidence()`) are not consumed by `oe_verify_evidence()`.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -591,63 +594,119 @@ plugin libraries, composed of:
...
-`oe_get_report()` and raw SGX ECDSA quotes generated by the Intel SGX SDK.
-These two evidence formats are identified by their UUIDs:
-- `OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT`: for SGX ECDSA report generated
-by `oe_get_report()`.
-- `OE_FORMAT_UUID_SGX_ECDSA_P256_QUOTE`: for raw SGX ECDSA quote generated
-by the Intel SGX SDK.
+Currently there is no plan to implement verifier plugins for verification of
+EPID evidence. Verification of SGX EPID quotes is not supported by
+the Intel SGX SDK. In all existing SGX EPID solutions, EPID quotes are sent to
+Intel Attestation Service (IAS) backend for verification.
```
dthaler, line 600: Can you point to an SDK or doc that explains how to send them to the IAS service?

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -591,63 +594,119 @@ plugin libraries, composed of:
...
+  * `oe_report_header` was defined for legacy API `oe_get_report()`.
+  Its `report_type` field is only intended to only indicate between SGX report
+  (for local attestation) and SGX ECDSA-p256 quote (for remote attestation).
+  * SGX Verifier plugins are able to identify the `custom_claims_buffer`
+  without the help of the `report_size` field in the `oe_report_header`.
+
+### Discussion
 
-Note: host-side SGX plugin library implementation of the public API functions
+Host-side SGX plugin library implementation of the public API functions
```
dthaler, line 709:
```suggestion
The host-side SGX plugin library implementation of the public API functions
```

docs/DesignDocs/Attestation_API_Proposal.md
dthaler: I tend to agree with Simon on this one.

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -591,63 +594,119 @@ plugin libraries, composed of:
...
+  generated by the legacy API function `oe_get_report()` without the
+  `OE_REPORT_FLAGS_REMOTE_ATTESTATION` flag (for local attestation).
+  These reports can be verified using legacy API function `oe_verify_report()`.
+- `OE_FORMAT_UUID_RAW_SGX_QUOTE_ECDSA`
+  - For SGX quotes in ECDSA_P256 format generated by the Intel SGX SDK DCAP
+  library, and the Intel SGX quote-ex library with algorithm ID
+  `SGX_QL_ALG_ECDSA_P256`.
+    - Note: the DCAP library only generates ECDSA-p256 quotes
+  - Note: no plan to provide verifier plugin for verification of SGX quotes
+  generated by the quote-ex library with other algorithm IDs (e.g. EPID quotes).
```
dthaler, line 624:
```suggestion
  generated by the quote-ex library with other algorithm IDs (e.g., EPID quotes).
```

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -591,63 +594,119 @@ plugin libraries, composed of:
...
+existing SGX SDK EPID-based solutions behavior, and will help maintain backward
 compatibility.
 
-The table below shows the structure of the evidence data for all the
-SGX format IDs, as generated by an attester plugin or verified by a verifier
-plugin. For every listed format, `oe_get_evidence()` can wrap with
-an attestation header when the `OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID` bit in its
+### Summary
+
+API function `oe_get_evidence()` supports the values listed below in its
```
dthaler, line 643:
```suggestion
The API function `oe_get_evidence()` supports the values listed below in its
```

docs/DesignDocs/Attestation_API_Proposal.md
```diff
@@ -591,63 +594,119 @@ plugin libraries, composed of:
...
+| `OE_FORMAT_UUID_SGX_ECDSA` | `[ oe_attestation_header ] \|\| SGX_ECDSA_p256_quote(hash) \|\| custom_claims_buffer` |
+| `OE_FORMAT_UUID_SGX_EPID_LINKABLE` | `[ oe_attestation_header ] \|\| SGX_EPID_linkable_quote(custom_claims_buffer)` |
+| `OE_FORMAT_UUID_SGX_EPID_UNLINKABLE` | `[ oe_attestation_header ] \|\| SGX_EPID_unlinkable_quote(custom_claims_buffer)` |
+| `OE_FORMAT_UUID_LEGACY_REPORT_REMOTE` | `oe_report_header (for remote attestation) \|\| SGX_ECDSA_p256_quote(custom_claims_buffer)` |
+| `OE_FORMAT_UUID_RAW_SGX_QUOTE_ECDSA` | `SGX_ECDSA_p256_quote(custom_claims_buffer)` |
 
 In the above table:
+* The optional header `oe_attestation_header` is a structure of type
+`oe_attestation_header_t`.
+* For every format supported by `oe_get_evidence()`, the evidence
```
dthaler, line 689: Can you state what structure type is SGX_report(), SGX_ECDSA_p256(quote), etc.?

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: I think we can explore various options when we are ready to deprecate `oe_get_report()`. As Simon mentioned, I'd like this PR to focus on enabling the attestation service scenarios.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Will add some links. Thanks!

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Thanks!

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: In response to
> I tend to agree with Simon on this one.

As commented previously:

> I think we can explore various options when we are ready to deprecate oe_get_report(). As Simon mentioned, I'd like this PR to focus on enabling the attestation service scenarios.

Let me know if this approach will not work.

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Thanks!

docs/DesignDocs/Attestation_API_Proposal.md
shnwc: Will add comment. Thanks!

shnwc:
bors try

shnwc:
bors retry

PR #3403, opened 2020-08-17
EXPERIMENT: Figure out enclave_base_address without help from loader.

anakrish:
Initialize _enclave_rva to its own address:
   OE_EXPORT volatile void* _enclave_rva = &_enclave_rva;

Due to this an X64_RELATIVE relocation entry will be created for _enclave_rva.
Additionally, in the image and in the enclave before relocation,
_enclave_rva will have its own RVA value. This is asserted by the loader.

Thus, before relocation,
   (uint8_t*) &_enclave_rva - (uint64_t) _enclave_rva
will give the enclave's base address which is stored in another variable for use.
After relocation, the above expression will evaluate to zero.

### DIFF
--- a/enclave/core/sgx/globals.c
+++ b/enclave/core/sgx/globals.c
@@ -125,7 +125,8 @@ OE_EXPORT extern volatile const oe_sgx_enclave_properties_t
  *
  **/
 
-OE_EXPORT volatile uint64_t _enclave_rva;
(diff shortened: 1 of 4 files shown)


anakrish:
bors try

anakrish:
bors try

anakrish:
Experiment complete.

PR #3408, opened 2020-08-18
Extend attestation certificate verifiers to support multiple OIDs

### LINKED ISSUE #3338: Extend oe_verify_attestation_certificate_with_evidence() to support multiple OIDs
`oe_verify_attestation_certificate()` is used to verify x.509 certificate with oe_report embedded in x.509 certificate extension ("1.2.840.113556.10.1.1").

`oe_verify_attestation_certificate_with_evidence()` was recently added to verify oe_evidence extension ("1.2.840.113556.10.1.2").

Since their OIDs are different, one function should be able to verify both extensions. The work here:
1. `oe_verify_attestation_certificate_with_evidence()` to verify both extension.
2. New OIDs were recently defined. Ensure `oe_verify_attestation_certificate()` and `oe_verify_attestation_certificate_with_evidence()` verify old and new OIDs:

OID | Purpose | Status | Assignment Date
-- | -- | -- | --
1.3.6.1.4.1.311.105.1 | OE SGX Report | Proposed | 8/8/2020
1.3.6.1.4.1.311.105.2 | OE SGX Evidence | Proposed | 8/8/2020

ryanhsu19:
Define new OIDs and enable both `oe_verify_attestation_certificate()` and `oe_verify_attestation_certificate_with_evidence()` to check for the latest extensions.

Add support for verification of OE reports to the function `oe_verify_attestation_certificate_with_evidence()`.

Fix #3338.


common/attest_plugin.c
```diff
@@ -321,6 +329,7 @@ oe_result_t oe_verify_attestation_certificate_with_evidence(
     size_t pub_key_buff_size = KEY_BUFF_SIZE;
     oe_claim_t* claims = NULL;
     size_t claims_length = 0;
+    uint8_t is_oe_evidence = 1;
```
yentsanglee, line 332: bool

include/openenclave/bits/evidence.h
```diff
@@ -25,6 +24,12 @@ OE_EXTERNC_BEGIN
     }
 #define X509_OID_FOR_OE_EVIDENCE_STRING "1.2.840.113556.10.1.2"
 
+#define X509_OID_FOR_NEW_OE_EVIDENCE_EXT                     \
+    {                                                        \
+        0x2A, 0x86, 0x48, 0x86, 0xF7, 0x14, 0x0A, 0x01, 0x02 \
```
yentsanglee, line 29: This OID does not match the string definition below

include/openenclave/internal/report.h
```diff
@@ -85,6 +85,12 @@ OE_STATIC_ASSERT(
     }
 #define X509_OID_FOR_QUOTE_STRING "1.2.840.113556.10.1.1"
 
+#define X509_OID_FOR_NEW_QUOTE_EXT                           \
+    {                                                        \
+        0x2A, 0x86, 0x48, 0x86, 0xF7, 0x14, 0x0A, 0x01, 0x01 \
```
yentsanglee, line 90: This OID does not match the string definition below

include/openenclave/bits/evidence.h
```diff
@@ -25,6 +24,12 @@ OE_EXTERNC_BEGIN
     }
 #define X509_OID_FOR_OE_EVIDENCE_STRING "1.2.840.113556.10.1.2"
 
+#define X509_OID_FOR_NEW_OE_EVIDENCE_EXT                     \
+    {                                                        \
+        0x2B, 0x06, 0x01, 0x04, 0x01, 0x82, 0x37, 0x69, 0x02 \
+    }
+#define X509_OID_FOR_NEW_OE_EVIDENCE_STRING "1.3.6.1.4.1.311.105.2"
```
mingweishih, line 31: Just a question, why do we need both format?

cc @yentsanglee

yentsanglee: The binary form to pass in to `mbedtls_x509write_crt_set_extension` to write to a certificate.
The string form to be compared with when parsing the certificate extension.

common/attest_plugin.c
```diff
@@ -338,40 +347,106 @@ oe_result_t oe_verify_attestation_certificate_with_evidence(
 
     //------------------------------------------------------------------------
     // Validate the report's trustworthiness
+    // Verify the remote report to ensure its authenticity.
+    // set enclave to NULL because we are dealing only with remote report now
     //------------------------------------------------------------------------
 
     // determine the size of the extension
     if (oe_cert_find_extension(
```
yentsanglee, line 355: The code in the four if blocks are almost identical. I'd suggest making an array of OIDs and perform in a loop:

`
oid_array = { oid_report_1, oid_report_2, oid_evidence_1, oid_evidence_2 };
for (n = 0; n < countof(oid_array); n++) {
    report_size = 0;
    if (oie_cert_find_extension(..., oid_array[n], ..., &report_size) == TOO_SMALL) {
        ....
        break;
    } 
}

if (n <2)
    // is report
else if (n < 4)
    // is evidence
else
    // no match
`

include/openenclave/bits/evidence.h
mingweishih: Is there any need for supporting such conversion? I'm asking because I implemented a simple oid parsing function (hex to string) when porting OpenSSL-based OE crypto APIs #3395  (https://github.com/openenclave/openenclave/pull/3395/files#diff-b1a9d4b221a2f293e70eee3421a9aa6dR1087).
This is just for supporting OpenSSL version of `oe_gen_custom_x509_cert`. Can you take a look?

yentsanglee: @mingweishih if oe_gen_custom_x509_cert() takes oid string as input, we don't need to provide the binary forms. That will be great.

mingweishih: Does that imply that the binary form is only needed by the mbedtls API (i.e., is the string format more common)? In that case, does it make sense to have `ext_oid` in the `oe_cert_config_t` to store the string format directly, and have the conversion function implemented in the mbedtls-based crypto layer instead of openssl-based one?

yentsanglee: The binary format is solely for the mbedtls API. The string format is more common and readable. However, the reason for having 2 forms is to eliminate the need for conversion. In your PR #3395, I see that you have to call _decode_oid_to_str() to convert to string. If that's the case for mbedtls too, then I'd suggest to have `ext_oid` to store the string form t avoid conversion.

common/attest_plugin.c
```diff
@@ -338,40 +351,78 @@ oe_result_t oe_verify_attestation_certificate_with_evidence(
...
-        NULL,
-        report,
-        report_size,
-        NULL,
-        0,
-        NULL,
-        0,
-        &claims,
-        &claims_length);
+    if (oid_array_index >= 2)
```
qiucwang, line 395:
```suggestion
    if (oid_array_index >= 2)    // oid_oe_evidence or oid_new_oe_evidence
```

common/attest_plugin.c
```diff
@@ -338,40 +351,78 @@ oe_result_t oe_verify_attestation_certificate_with_evidence(
...
+            report,
+            report_size,
+            NULL,
+            0,
+            NULL,
+            0,
+            &claims,
+            &claims_length);
+    }
+    else
```
qiucwang, line 410:
```suggestion
    else    oid_oe_report or oid_new_oe_report
```

common/attest_plugin.c
```diff
@@ -338,40 +351,78 @@ oe_result_t oe_verify_attestation_certificate_with_evidence(
...
-        0,
-        NULL,
-        0,
-        &claims,
-        &claims_length);
+    if (oid_array_index >= 2)
+    {
+        result = oe_verify_evidence(
+            // rely on the format UUID in the header. For attestation
+            // certificate, the report should always include the header.
```
qiucwang, line 399:
```suggestion
            // The format id parameter is NULL here as the format ID is embedded in 
            // the attestation header, which is always included in an attestation certificate.
```

common/attest_plugin.c
```diff
@@ -338,40 +351,78 @@ oe_result_t oe_verify_attestation_certificate_with_evidence(
...
+            0,
+            NULL,
+            0,
+            &claims,
+            &claims_length);
+    }
+    else
+    {
+        result = oe_verify_evidence(
+            // The UUID is OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT for all OE
```
qiucwang, line 413:
```suggestion
            // The format id is OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT for all OE
```

common/attest_plugin.c
```diff
@@ -322,6 +330,11 @@ oe_result_t oe_verify_attestation_certificate_with_evidence(
     oe_claim_t* claims = NULL;
     size_t claims_length = 0;
 
+    const char* oid_array[] = {
+        oid_oe_report, oid_new_oe_report, oid_oe_evidence, oid_new_oe_evidence};
+    size_t oid_array_index = 0;
+    size_t oid_array_count = sizeof(oid_array) / sizeof(oid_array[0]);
```
yentsanglee, line 336: OE_COUNTOF()

common/sgx/tlsverifier.c
```diff
@@ -115,15 +116,33 @@ oe_result_t oe_verify_attestation_certificate(
...
+    }
+    else
+    {
+        report = (uint8_t*)oe_malloc(report_size);
+        if (!report)
+            OE_RAISE(OE_OUT_OF_MEMORY);
 
-    report = (uint8_t*)oe_malloc(report_size);
-    if (!report)
-        OE_RAISE(OE_OUT_OF_MEMORY);
```
qiucwang, line 122: I saw there are only two oids here so you didn't use a loop like you did for `oe_verify_attestation_certificate_with_evidence()`. And I suggest to use structure like:
```
if(check(A))
{ process(A); }
else if(check(B))
{ process(B); }
else
{ raise fail; }
```
That might make other readers easier to follow the code.

tests/attestation_cert_apis/enc/enc.cpp
```diff
@@ -266,6 +316,14 @@ oe_result_t get_tls_cert_signed_with_key(
         "\nFrom inside enclave: verifying the certificate... %s\n",
         result == OE_OK ? "Success" : "Fail");
 
+    // validate cert with oe_verify_attestation_certificate_with_evidence() to
+    // ensure that the added report verifier part of the function works well
+    result = oe_verify_attestation_certificate_with_evidence(
+        output_cert, output_cert_size, sgx_enclave_claims_verifier, NULL);
+    OE_TRACE_INFO(
+        "\nFrom inside enclave: verifying the certificate... %s\n",
```
yentsanglee, line 324: Make this log different from line 316 so it's easier to tell which function fails.

common/attest_plugin.c
```diff
@@ -40,6 +44,50 @@ const char* OE_OPTIONAL_CLAIMS[OE_OPTIONAL_CLAIMS_COUNT] = {
 // Variables storing the verifier list.
 static oe_plugin_list_node_t* verifiers = NULL;
 
+// UUID for all OE reports generated by oe_get_report().
+static const oe_uuid_t _ecdsa_report_uuid = {
```
dthaler, line 48: add sgx to variable name

common/attest_plugin.c
```diff
@@ -40,6 +44,50 @@ const char* OE_OPTIONAL_CLAIMS[OE_OPTIONAL_CLAIMS_COUNT] = {
...
+    OE_FORMAT_UUID_SGX_ECDSA_P256_REPORT};
+
+// verify report user data against peer certificate
+static oe_result_t verify_report_user_data(
```
dthaler, line 52: Add sgx to function name.

Also this file is nominally a TEE-agnostic file, and should not normally have any SGX-only functions in this file.  Such code should be in a filename with "sgx" in the filename (or directory name)

dthaler: Specifically, I think this function should be down in the SGX plugin, rather than in the core plugin framework

include/openenclave/bits/evidence.h
```diff
@@ -25,6 +24,12 @@ OE_EXTERNC_BEGIN
     }
 #define X509_OID_FOR_OE_EVIDENCE_STRING "1.2.840.113556.10.1.2"
 
+#define X509_OID_FOR_NEW_OE_EVIDENCE_EXT                     \
```
mingweishih, line 27: I didn't see this define is used anywhere in the code. Am I missing something?

include/openenclave/internal/report.h
```diff
@@ -85,6 +85,12 @@ OE_STATIC_ASSERT(
     }
 #define X509_OID_FOR_QUOTE_STRING "1.2.840.113556.10.1.1"
 
+#define X509_OID_FOR_NEW_QUOTE_EXT                           \
```
mingweishih, line 88: I didn't see this define is used anywhere in the code. Am I missing something?

include/openenclave/bits/evidence.h
mingweishih: I think I'm referring to the consistency of our API regardless which crypto libraries is used. Do we have any API that takes oid as input? What I only know it's used internally in the `oe_generate_attestation_certificate` function, which I think it's fine to use hardcoded value there. I'm wondering if there is other case.

yentsanglee: There is no public APIs that take OID as input. All OID definitions are internal (either in `internal/evidence.h` or `internal/report.h`).

include/openenclave/bits/evidence.h
yentsanglee: X509_OID_FOR_NEW_OE_EVIDENCE_EXT is a recently agreed upon OID for evidence extension. It will be used for certificate generation in the next release. We want to ensure the verifiers can handle such OID before starting generating with the new OID.

include/openenclave/internal/report.h
yentsanglee: Same above - enable verifier scenario first.

include/openenclave/bits/evidence.h
mingweishih: Thanks for the clarification. Then, I think it makes sense to not doing the conversion by ourselves.

ryanhsu19:
bors try

PR #3414, opened 2020-08-19
Integrate SGX DCAP quote verification into OE SDK

hyjiang:
Proposal of integrating Intel(R) SGX DCAP quote verification into OE SDK.

Signed-off-by: hyjiang <hongyan.jiang@intel.com>

### DIFF
--- a/docs/DesignDocs/SGX_QuoteVerify_Integration.md
+++ b/docs/DesignDocs/SGX_QuoteVerify_Integration.md
@@ -1,0 +1,525 @@
+Attestation: OE SDK Integration with Intel® SGX DCAP Quote Verification Library for SGX Evidence Verification
+====
+
+This design document proposes an update of the OE SDK implementation
+for integration with the Intel® SGX Data Center Attestation Primitives (DCAP)
+Quote Verification Library (QVL), for support of evidence verification in
+ECDSA-p256 format.
+
+# Motivation
+
+The existing OE SDK implements its own SGX ECDSA quote verification.
+As implemented in code file `common/sgx/quote.c`, the verification code
+performs SGX cert chain verification and Enclave Identity
+verification, including X.509 parser, cert revocation checking, JSON parser,
+etc. Also it uses some 3rd-party libraries, such as mbedtls.
+
+However, there are a few areas that could be improved:
+- The current SGX ECDSA quote verification logic in `common/sgx/quote.c`
+  is not a complete implementation, e.g.:
+  - Only check quote version number in function `_validate_sgx_quote()`
+    in file `common/sgx/quote.c`
+  - Only allow TCB level `UpToDate`, all other TCB levels would be treated
+    as invalid. But in SGX design, some other TCB levels such as `OutOfDate`,
+    `SWConfigNeeded` should not be treated as critical error, it's up to
+    a verifier application to decide whether the platform TCB is valid or not
+    for its intended usages.
+- In the current implementation, all the quote verification logic is
+  built into verifier enclaves. This means that OE SDK greatly increases the TCB of
+  verifier enclaves. If security bugs in verification logic,
+  including 3rd-party component, are discovered, verifier enclaves need to be
+  upgraded and rebuilt.
+
+# User Experience
+
+The proposed update only changes the internal implementation of the OE SDK
+attestation software stack. It does not impact the
+[OE SDK attestation API](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/Attestation_API_Proposal.md).
+
+With the integration of Intel® SGX DCAP QVL (including an untrusted library
+(called QVL when there is no ambiguity)
+and a Quote Verification Enclave (QvE)), a verifier's call to OE SDK API
+`oe_verify_evidence()` for verification of SGX ECDSA evidence would trigger
+quote verification by invoking QVL or QvE, depending on whether the call is
+from the host side or the enclave side, respectively.
+
+Integration of the DCAP QVL depends on the installation of
+Intel® SGX DCAP packages `libsgx-dcap-quote-verify` and `libsgx-ae-qve` and
+their dependencies, as well as proper configuration of the components and their
+access to dependent backend services (e.g. Quote Provider Library and
+Provisioning Certificate Cashing Service (PCCS)). Details for the DCAP quote
+verification library installation and configuration are outside the scope of
+this document.
+
+# Specification
+
+## Existing OE SDK Implementation
+
+### Implementation of the SGX ECDSA-p256 verifier plugin
+
+The OE SDK framework implementation searches for a verifier plugin that
+supports the requested evidence format, and invokes the `verify_evidence()` or
+`verify_report()` entry point of the selected plugin.
+
+The SGX ECDSA-p256 verifier plugin is implemented in code files
+`enclave/sgx/report.c`, `common/sgx/verifier.c` and
+`common/sgx/quote.c`. The same source tree implements both the enclave-side and
+host-side verifier plugins.
+
+There are 2 different verifier application / enclave scenarios in current
+implementation for SGX ECDSA quote verification:
+- Scenario 1 - Call `oe_verify_report()` to verify SGX remote report (aka SGX quote)
+- Scenario 2 - Call `oe_verify_evidence()` to verify evidence in format
+  `OE_FORMAT_UUID_SGX_ECDSA_P256`.
+
+Based on design doc [Remote Attestation Collaterals](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/RemoteAttestationCollaterals.md),
+new API `oe_verify_evidence()` will supersede `oe_verify_report()`. In both
+scenarios, the same internal function `oe_verify_quote_with_sgx_endorsements()`
+is invoked for quote verification.
+
+For SGX ECDSA-p256 quote verification, in the enclave-side and host-side plugin
+implementation of function `_verify_evidence()` in `common/sgx/verifier.c`,
+several functions are called.
+- `oe_get_sgx_endorsements()` and `oe_parse_sgx_endorsements()`
+  - Get relevant endorsements, including SGX PCK cert CRL, TCB info, QE identity,
+    etc.
+- `oe_verify_quote_with_sgx_endorsements()`
+  - Verify quote with provided endorsements, including quote parsing, cert chain
+    veriifcation, TCB level matching, etc.
+- `oe_sgx_extract_claims()`
+  - Fill required and custom claims
+
+## Background: Intel® SGX DCAP QVL library and API
+
+For verification of SGX ECDSA quotes, the SGX DCAP QVL
+library has the following relevant API functions defined in its
+[header file](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteVerification/dcap_quoteverify/inc/sgx_dcap_quoteverify.h):
+
+```C
+/**
+ * Get supplemental data required size.
+ * @param p_data_size[OUT] - Pointer to hold the size of the buffer in bytes
+ *  required to contain all the supplemental data.
+ *
+ * @return Status code of the operation, one of:
+ *      - SGX_QL_SUCCESS
+ *      - SGX_QL_ERROR_INVALID_PARAMETER
+ *      - SGX_QL_ERROR_QVL_QVE_MISMATCH
+ *      - SGX_QL_ENCLAVE_LOAD_ERROR
+ **/
+quote3_error_t sgx_qv_get_quote_supplemental_data_size(uint32_t *p_data_size)
+
+/**
+ * Perform quote verification.
+ *
+ * @param p_quote[IN] - Pointer to SGX Quote.
+ * @param quote_size[IN] - Size of the buffer pointed to by p_quote (in bytes).
+ * @param p_quote_collateral[IN] - This is a pointer to the Quote Certification
+ *        Collateral provided by the caller.
+ * @param expiration_check_date[IN] - This is the date that the QvE will use to
+ *        determine if any of the inputted collateral have expired.
+ * @param p_collateral_expiration_status[OUT] - Address of the output expiration
+ *        status.  This input must not be NULL.
+ * @param p_quote_verification_result[OUT] - Address of the output quote
+ *  verification result.
+ * @param p_qve_report_info[IN/OUT] - This parameter can be used in 2 ways.
+ *        If p_qve_report_info is NOT NULL, the API will use Intel QvE to
+ *        perform quote verification, and QvE will generate a report   using the
+ *        target_info in sgx_ql_qe_report_info_t structure.
+ *        if p_qve_report_info is NULL, the API will use QVL library to perform
+ *        quote verification, not that the results can not be cryptographically
+ *        authenticated in this mode.
+ * @param supplemental_data_size[IN] - Size of the buffer pointed to by p_quote (in bytes).
+ * @param p_supplemental_data[OUT] - The parameter is optional.  If it is NULL,
+ *        supplemental_data_size must be 0.
+ *
+ * @return Status code of the operation, one of:
+ *      - SGX_QL_SUCCESS
+ *      - SGX_QL_ERROR_INVALID_PARAMETER
+ *      - SGX_QL_QUOTE_FORMAT_UNSUPPORTED
+ *      - SGX_QL_QUOTE_CERTIFICATION_DATA_UNSUPPORTED
+ *      - SGX_QL_UNABLE_TO_GENERATE_REPORT
+ *      - SGX_QL_ERROR_UNEXPECTED
+ **/
+quote3_error_t sgx_qv_verify_quote(
+    const uint8_t *p_quote,
+    uint32_t quote_size,
+    const struct _sgx_ql_qve_collateral_t *p_quote_collateral,
+    const time_t expiration_check_date,
+    uint32_t *p_collateral_expiration_status,
+    sgx_ql_qv_result_t *p_quote_verification_result,
+    sgx_ql_qe_report_info_t *p_qve_report_info,
+    uint32_t supplemental_data_size,
+    uint8_t *p_supplemental_data)
+
+```
+
+Notes on the API function `sgx_qv_verify_quote()`:
+- Verifier application (called the verifier) controls quote verification via trusted
+  QvE or untrusted QVL by specifying parameter `p_qve_report_info`
+  - Trusted QvE when this parameter is non-NULL, containing the `report`
+    for the verifier enclave that verifies the QvE security properties:
+    Quote verification would be done inside QvE, and QvE would return a report
+    targeting the verifier enclave, it means the verifier can verify QvE's
+    return report and identity
+  - Untrusted QVL when this parameter is NULL:
+    Quote verification would be done inside untrusted QVL library, the verifier can
+    use this way on a non-SGX capable system, but the result cannot be
+    cryptographically authenticated in this mode
+- If verifier doesn't provide quote verification collateral (aka Endorsement),
+  QVL will try to load the Quote Provide Library (QPL) and connect to the PCK
+  Cert Caching Server (PCCS) to get corresponding verification collaterals
+  automatically
+- This API can return a [supplemental data structure](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteVerification/QvE/Include/sgx_qve_header.h)
+to allow the verifier to have an alternative verification policy
+
+```C
+/** Contains data that will allow an alternative quote verification policy. */
+typedef struct _sgx_ql_qv_supplemental_t
+{
+    uint32_t version;               ///< Supplemental data version
+    time_t earliest_issue_date;     ///< Earliest issue date of all the collateral (UTC)
+    time_t latest_issue_date;       ///< Latest issue date of all the collateral (UTC)
+    time_t earliest_expiration_date;///< Earliest expiration date of all the collateral (UTC)
+    time_t tcb_level_date_tag;      ///< The SGX TCB of the platform that generated the quote is not vulnerable
+                                    ///< to any Security Advisory with an SGX TCB impact released on or before this date
+                                    ///< See [Intel Security Center Advisories](https://www.intel.com/content/www/us/en/security-center/default.html)
+    uint32_t pck_crl_num;           ///< CRL Num from PCK Cert CRL
+    uint32_t root_ca_crl_num;       ///< CRL Num from Root CA CRL
+    uint32_t tcb_eval_ref_num;      ///< Lower number of the TCBInfo and QEIdentity
+    uint8_t root_key_id[48];        ///< ID of the collateral's root signer (hash of Root CA's public key SHA-384)
+    sgx_key_128bit_t pck_ppid;      ///< PPID from remote platform.  Can be used for platform ownership checks
+    sgx_cpu_svn_t tcb_cpusvn;       ///< CPUSVN of the remote platform's PCK Cert
+    sgx_isv_svn_t tcb_pce_isvsvn;   ///< PCE_ISVNSVN of the remote platform's PCK Cert
+    uint16_t pce_id;                ///< PCE_ID of the remote platform
+    uint8_t sgx_type;               ///< Indicate the type of memory protection available on the platform, it should be
+                                    ///< one of Standard (0) and Scalable (1)
+
+    // Multi-Package PCK cert related flags, they are only relevant to PCK Certificates issued by PCK Platform CA
+    uint8_t platform_instance_id[16];///< Value of Platform Instance ID, 16 bytes
+    pck_cert_flag_enum_t dynamic_platform; ///< Indicate whether a platform can be extended with additional packages
+                                           ///< via Package Add calls to SGX Registration Backend
+    pck_cert_flag_enum_t cached_keys;      ///< Indicate whether platform root keys are cached by SGX Registration Backend
+    pck_cert_flag_enum_t smt_enabled;      ///< Indicate whether a plat form has SMT (simultaneous multithreading) enabled
+
+} sgx_ql_qv_supplemental_t;
+```
+
+## Proposed Changes
+
+In order to smoothly switch quote verification from OE to SGX DCAP QVL/QvE, we
+propose to implement the change in 2 steps as below:
+- Step 1: Implement QVL-based quote verification and make it co-exist with the
+  existing implementation
+- Step 2: Remove existing implementation and makes the DCAP QVL a hard dependency
+
+### Step 1: Implement QVL-based quote verification and make it co-exist with the existing implementation
+If SGX DCAP QVL/QvE are not installed, quote verification will continue to use existing
+logic. Otherwise, quote verification will switch to QVL/QvE automatically.
+
+#### Add functions in host-side plugin for SGX ECDSA-p256 quote verification
+To align with the current implementation of SGX quote-ex library integration,
+the OE SDK host-side verifier plugin library will dynamically detect the presence of
+QVL library and load it at runtime. In case the QVL library can't be found, a specfic
+error will be returned.
+
+- Add functions in file `host/sgx/sgxquote.c` and `host/sgx/quote.c`
+  - Use environment variable, e.g. “USE_SGX_QVL”. Only when this variable is set,
+    try to detect & load the QVL library, return specific error if environment variable
+    not set or QVL doesn't exist
+  - Call QVL API to verify the quote
+
+#### Add OCALL to support enclave-side SGX ECDSA-p256 quote verification
+As quote verification will be done by DCAP QVL/QvE, for enclave-side plugin,
+it requires OCALL to call host-side APIs.
+
+In this proposal, we will add one OCALL `oe_verify_quote_ocall()` in
+`edl/sgx/attestation.edl`.
+The OCALL is used for passing quote buffer and related parameters to host side.
+All other relevant logic will be implemented on the host side, in files
+`host/sgx/ocalls.c` and `host/sgx/quote.c`.
+If DCAP QVL cannot be found in host-side plugin, a specific error will be returned.
+
+Note that currently only ECDSA-p256 quotes are supported, but the OCALL will keep
+`format_id` and `opt_params` for forward compatibility.
+
+```C
+oe_result_t oe_verify_quote_ocall(
+    [in] const oe_uuid_t* format_id,
+    [in, size=opt_params_size] const void* opt_params,
+    size_t opt_params_size,
+    [in, size=quote_size] const void* p_quote,
+    size_t quote_size,
+    oe_datetime_t expiration_check_date,
+    [out] uint32_t *p_collateral_expiration_status,
+    [out] uint32_t *p_quote_verification_result,
+    [in, out, size=qve_report_size] void* p_qve_report_info,
+    size_t qve_report_size,
+    [out, size=supplemental_data_size] void* p_supplemental_data,
+    size_t supplemental_data_size,
+    [out] size_t* p_supplemental_data_size_out);
+```
+
+#### Update implementation of existing plugin functions
+These plugin functions include `_verify_evidence()`, `_verify_report()`
+and they both invoke function `oe_verify_quote_with_sgx_endorsements()`
+for quote verification.
+
+- Update API `oe_verify_quote_with_sgx_endorsements()` implementation as below:
+  - Try to call DCAP QVL to verify quote first
+    - In host-side plugin:
+      - Call host-side function `oe_sgx_verify_quote()` with SGX endorsements,
+        which will call SGX QVL library to verify quote.
+    - In enclave-side plugin:
+      - Construct structure `sgx_ql_qe_report_info_t` as below, `nonce` and
+        `app_enclave_target_info` are input, qve_report is QvE report which
+        target to application enclave.
+      ```C
+      typedef struct _sgx_ql_qe_report_info_t {
+        sgx_quote_nonce_t nonce;
+        sgx_target_info_t app_enclave_target_info;
+        sgx_report_t qve_report;
+      }sgx_ql_qe_report_info_t;
+      ```
+      - Call OCALL function `oe_verify_quote_ocall()` to host-side plugin library
+        to verify SGX quote.
+      - With all the QvE return values, call new API.
+        `oe_verify_qve_report_and_identity()` (described below) to
+        verify QvE report and identity
+  - If DCAP QVL returns error to indicate that QVL doesn't not exist on current
+    platform, then automatically fallback to the existing implementation
+  - Add log information to indicate whether the QVL-based or existing implementation
+    is used
+- If QVL-based quote verification is used, in order to exactly align with current
+  implementation, only SGX ECDSA quote with `UpToDate` TCB status would be accepted
+  in step 1.
+
+
+#### Add function in enclave-side plugin for QvE Identity verification
+For enclave-side verifier plugin, after quote verification, it needs to verify that the result
+comes from a trusted QvE.
+
+Intel® SGX SDK provides a library named `sgx_dcap_tvl` to help the verifier to
+verify QvE's identity, you can refer to [source file](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteVerification/dcap_tvl/sgx_dcap_tvl.cpp)
+
+This library uses hardcoded QvE identity values, because:
+- There is no need for x509 and JSON parser in verifier enclave's TCB.
+- Most of QvE's identity attributes only change infrequently.
+
+The only identity info may change more often is QvE's ISV SVN, so the verifier needs
+to provide a SVN number as threshold, only when current QvE's ISV SVN is equal to
+or larger than this threshold, the verifier can trust the QvE verification result.
+The verifier can get latest QvE ISV SVN from [QvE Identity at Intel® Provisioning
+Cert Server](https://api.portal.trustedservices.intel.com/documentation#pcs-qve-identity-v2)
+
+As OE SDK cannot use Intel® SGX SDK trusted library directly, we need to port
+this library into OE SDK as part of enclave-side plugin.
+
+```C
+oe_result_t oe_verify_qve_report_and_identity(
+        const uint8_t *p_quote,
+        uint32_t quote_size,
+        const sgx_ql_qe_report_info_t *p_qve_report_info,
+        time_t expiration_check_date,
+        uint32_t collateral_expiration_status,
+        sgx_ql_qv_result_t quote_verification_result,
+        const uint8_t *p_supplemental_data,
+        uint32_t supplemental_data_size,
+        sgx_isv_svn_t qve_isvsvn_threshold)
+```
+All the data returned by the QvE are included in the QvE report data field for
+integrity protection.
+```C
+QvE report_data = SHA256([nonce || quote || expiration_check_date ||
+expiration_status || verification_result || supplemental_data]) || 32 - 0x00
+```
+
+The flow of the API function `oe_verify_qve_report_and_identity()` is as below:
+- Verify QvE report.
+- Verify report data (report data described as above).
+- Check QvE Identity by comparing QvE report fields against hardcode values.
+  - Check that Report.MRSIGNER equals to Hardcoded QvE MRSIGNER.
+  - Check that Misc Select, Attribute and ProdID equal to the hardcoded values,
+    Misc select and Attribute need to apply Mask before comparison.
+- Check that Report.ISVSVN >= Hardcoded ISV SVN.
+
+Options to provide QvE ISV SVN threshold:
+- Option 1: Hardcode QvE ISV SVN in enclave-side plugin. Every time the QvE
+            ISV SVN is updated, the plugin library implementation will need to
+            be updated to stay in sync.
+- Option 2: Update existing enclave-side plugin API to allow the verifier to input
+            QvE ISV SVN threshold value. But it requires plugin API change, also
+            it's NOT TEE agnostic.
+- Option 3: Add custom policy for SGX, user can specify the QvE ISV SVN in SGX
+            custom policy. Then OE SDK doesn't need to change when SGX DCAP has
+            a new release.
+
+In order to align with existing implementation, we propose option 1 in step 1,
+and in step 2, we will try to change it to option 3.
+
+
+### Step 2: Remove existing implementation and makes the DCAP QVL a hard dependency
+In step 2, we propose to remove the existing implementation. The step 2 proposal
+is target for future OE release. 
+
+#### Update claims list to add SGX Quote Verification status and QVL/QvE returned supplemental data
+In SGX remote attestation, the verifier may want to provide a different quote
+verification policy than the one enforced by the sgx_qv_verify_quote() API.
+
+The proposal is to extend claims definition with new SGX related
+claims.
+
+SGX quote verification API `sgx_qv_verify_quote()` returns 2 types of errors,
+fatal errors or warning errors:
+- Fatal error: such as cert chain verification failed, cert revoked, API will
+  return an error code directly
+- Warning error, such as cert chain out-of-date, API will return successful, but
+  there is an output parameter to indicate the specific error code. In this case,
+  the verifier can decide whether the SGX quote is valid based on their own policy.
+
+We propose to introduce a new SGX-specific claim `OE_CLAIM_SGX_QUOTE_VERIFICATION_STATUS` to
+indicate QVL/QvE return status. The verifier can check the status to see if there
+is a warning error during quote verification. This claim can have one of three values:
+
+  - OE_OK - Quote verification passed, and TCB level is up to date
+  - Warning error code - API returns OE_OK,
+    but claim `OE_CLAIM_SGX_QUOTE_VERIFICATION_STATUS` in claims
+    returns an error code. The verifier can refer to this status in his/her
+    verification policy. Warning error can be one of below:
+
+```C
+   OE_SGX_CONFIG_NEEDED // Quote verification passed and the platform is patched
+                        // to latest TCB level, but additional configuration of
+                        // the SGX platform may be needed. For example, attester
+                        // may need to enable/disable SMT in BIOS
+   OE_SGX_OUT_OF_DATE  // The Quote is good but TCB level of the platform is
+                       // out of date, additional configuration of the SGX
+                       // Platform at its current patching level may be needed.
+                       // The platform needs patching to be at the latest TCB level
+   OE_SGX_SW_HARDENING_NEEDED  // The TCB level of the platform is up to date,
+                               // but SGX SW Hardening is needed. For example,
+                               // attester may need to apply LVI mitigation in
+                               // his/her enclave by linking LIV version SDK
+   OE_SGX_CONFIG_AND_SW_HARDENING_NEEDED  // The TCB level of the platform is up
+                                          // to date, but additional
+                                          // configuration of the platform at
+                                          // its current patching level may be
+                                          // needed. Moreover, SGX SW Hardening
+                                          // is also needed. For example,
+                                          // attester may need enable/disable
+                                          // SMT in BIOS, also he/she needs to
+                                          // apply LVI mitigation in enclave
+
+Note: SGX TCB = Platform CPU HW/FW TCB + SGX PCE TCB.
+```
+If the verifier needs to provide a different quote verification policy
+beyond the policy enforced by the sgx_qv_verify_quote() API, the verifier can
+request sgx_qv_verify_quote() API to return supplemental data. The detailed
+supplemental data definition is described in above `Background` section.
+
+So we also propose to add SGX supplemental data fields to the list of claims.
+
+[PR 3353](https://github.com/openenclave/openenclave/pull/3353) added SGX
+endorsements into SGX-specific claims, here we propose to use SGX supplemental
+data (returned by QVL/QvE) instead. Because user needs to parse the endorsements
+by himself if we only return the endorsements directly, but parsing x509 cert
+chain/CRL and JSON is quite a complex task.
+
+The new SGX-Specific claims definition (Pls refer to `Background` section for
+the description of each SGX specific claim):
+
+```C
+#define OE_CLAIM_SGX_QUOTE_VERIFICATION_STATUS "sgx_quote_verification_status"
+
+#define OE_CLAIM_SGX_TCB_LEVEL_DATE_TAG "sgx_tcb_level_date_tag"
+#define OE_CLAIM_SGX_PCK_CRL_NUM "sgx_pck_crl_num"
+#define OE_CLAIM_SGX_ROOT_CA_CRL_NUM "sgx_root_ca_crl_num"
+#define OE_CLAIM_SGX_TCB_EVAL_REF_NUM "sgx_tcb_eval_ref_num"
+#define OE_CLAIM_SGX_ROOT_KEY_ID "sgx_root_key_id"
+#define OE_CLAIM_SGX_PCK_PPID "sgx_pck_ppid"
+#define OE_CLAIM_SGX_TCB_CPUSVN "sgx_tcb_cpusvn"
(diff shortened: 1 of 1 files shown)


docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+====
+
+This design document proposes an update of the OE SDK implementation
+for integration with the Intel® SGX Data Center Attestation Primitives (DCAP)
+Quote Verification Library (QVL), for support of evidence verification in
+ECDSA-p256 format.
+
+# Motivation
+
+Existing OE SDK implements the functionality for SGX ECDSA quote verification.
```
dthaler, line 11: (grammar nit)
```suggestion
The existing OE SDK implements the functionality for SGX ECDSA quote verification.
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+As implemented in code file `common/sgx/quote.c`, the verification code
+performs SGX cert chain verification and Enclave Identity
+verification, including x509 parser, cert revocation checking, JSON parser and
```
dthaler, line 14:
```suggestion
verification, including X.509 parser, cert revocation checking, JSON parser,
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+etc. Also it uses some 3rd-party libraries, such as mbedtls.
+
+However, there are a few areas that could be improved:
+- Current SGX ECDSA quote verification logic in `common/sgx/quote.c`
+  is not a complete implementation, e.g.
```
dthaler, line 19:
```suggestion
- The current SGX ECDSA quote verification logic in `common/sgx/quote.c`
  is not a complete implementation, e.g.:
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+  - Only check quote version number in function `_validate_sgx_quote()`
+    in file `common/sgx/quote.c`
+  - Only allow TCB level `UpToDate`, all other TCB levels would be treated
+    as invalid. But in SGX design, some other TCB levels such as `OutOfDate`,
+    `SWConfigNeeded` should not be treated as critical error, it's up to
+    a verifier application to decide whehter the platform TCB is valid or not
```
dthaler, line 25:
```suggestion
    a verifier application to decide whether the platform TCB is valid or not
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+    for its intended usages.
+- In current implementaion, all the quote verification logic is
```
dthaler, line 27:
```suggestion
- In the current implementation, all the quote verification logic is
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+The proposed update only changes the internal implementation of the OE SDK
+attestation software stack. It does not impact the
+[OE SDK attestation API](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/Attestation_API_Proposal.md).
+
+With the integration of Intel® SGX DCAP QVL (including an untrusted library
+(called QVL when there is no ambiguity)
+and a Quote Verification Enclave (QvE)), a verifier's call to OE SDK API
+`oe_verify_evidence()` for verification of SGX ECDSA evidence would trigger
+quote verification by invoking QVL or QvE, depending on whether the call is
+from the enclave side or the host side.
```
dthaler, line 44:
```suggestion
from the host side or the enclave side, respectively.
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+this document.
+
+# Specification
+
+## Existing OE SDK Implementation
+
+### Implementation of the SGX ECDSA-p256 verifier plugin
+
+The OE SDK framework implementation searchs for a verifier plugin that
+supports the requested evidence format, and invoke the `verify_evidence()` or
```
dthaler, line 61:
```suggestion
The OE SDK framework implementation searches for a verifier plugin that
supports the requested evidence format, and invokes the `verify_evidence()` or
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+There are 2 different verifier application / enclave scenarios in current
+implementation for SGX ECDSA quote verification:
+- Scenario 1 - Call `oe_verify_report()` to verify SGX remote report (aka SGX quote)
+- Scenario 2 - Call `oe_verify_evidence()` to verify evidence in format
+  `OE_FORMAT_UUID_SGX_ECDSA_P256`.
+
+Based on design doc [Remote Attestation Collaterals](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/RemoteAttestationCollaterals.md),
+new API `oe_verify_evidence()` will supersede
+`oe_verify_report()`. So we only discuss scenario 2 here. But in both scenarios,
+the same function `oe_verify_quote_with_sgx_endorsements()` is invoked for
```
dthaler, line 78:
```suggestion
the same internal function `oe_verify_quote_with_sgx_endorsements()` is invoked for
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+quote verification.
+
+For SGX ECDSA-p256 quote verification, in the enclave-side and host-side plugin
+implementation of function `_verify_evidence()` in `common/sgx/verifier.c`,
+several functions are called.
+- `oe_get_sgx_endorsements()` and `oe_parse_sgx_endorsements()`
+  - Get relevant endorsements, including SGX PCK cert CRL, TCB info, QE identity
+    and etc
```
dthaler, line 86:
```suggestion
  - Get relevant endorsements, including SGX PCK cert CRL, TCB info, QE identity,
    etc.
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+- `oe_verify_quote_with_sgx_endorsements()`
+  - Verify quote with provided endorsements, including quote parsing, cert chain
+    veriifcation, TCB level matching and etc
```
dthaler, line 89:
```suggestion
    verification, TCB level matching, etc.
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+    Quote verification would be done inside QvE, and QvE would return a report
+    targeting the verifier enclave, it means the verifier can verify QvE's
+    reutrn report and identity
+  - Untrusted QVL when this parameter is NULL:
+    Quote verification would be done inside untrusted QVL library, the verifier can
+    use this way on a non-SGX capable system, but the result can not be
+    cryptographically authenticated in this mode
+- If verifier doesn't provide quote verification collateral (aka Endorsement),
+  QVL will try to load Quote Provide Library(QPL) and connect to 
+  PCK Cert Caching Sever(PCCS) to get corresponding verification collaterals
```
dthaler, line 172:
```suggestion
  QVL will try to load the Quote Provide Library (QPL) and connect to 
  the PCK Cert Caching Server (PCCS) to get corresponding verification collaterals
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+    pck_cert_flag_enum_t cached_keys;      ///< Indicate whether platform root keys are cached by SGX Registration Backend
+    pck_cert_flag_enum_t smt_enabled;      ///< Indicate whether a plat form has SMT (simultaneous multithreading) enabled
+
+} sgx_ql_qv_supplemental_t;
+```
+
+## Proposed Changes
+
+### Options for retrieving SGX endorsements
+There are two options for the OE SDK plugin library to retreive SGX endorsments.
```
dthaler, line 212:
```suggestion
There are two options for the OE SDK plugin library to retrieve SGX endorsements.
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+
+#### Option 1: Keep existing implemention in SGX verifier plugin
```
dthaler, line 214:
```suggestion
#### Option 1: Keep existing implementation in SGX verifier plugin
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+In the current plugin implementation, if the verifier doesn't provide SGX endoresements, it
+calls API `oe_get_sgx_endorsements()` to parse SGX ECDSA quote body to get PCK
+cert chain first, then calls OCALL API `oe_get_quote_verification_collateral_ocall()`
+to load Quote Provide Library(QPL) and connect to PCK Cert Caching Sever(PCCS)
+to get corresponding verification collaterals, including CRL, TCB and QE
+Identity and etc.
```
dthaler, line 220:
```suggestion
Identity, etc.
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+    collateral defintion is changed
+  - The OE SDK logic for retrieving and parsing endorsements increases
+    verifier enclave TCB.
+
+With option 2, OE SDK doesn't need to maintain the complex logic for SGX collateral
+retrieving & parsing, nor need to be updated in response to changes
+in SGX collateral definition.
+
+### Link with the SGX DCAP QVL Library
+To align with the current implementaion of SGX quote-ex library integration,
```
dthaler, line 246:
```suggestion
To align with the current implementation of SGX quote-ex library integration,
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+the OE SDK host-side plugin library will dynamically detect the presence of QVL
+library and loads it at runtime. In case the QVL library can't be found,
```
dthaler, line 248:
```suggestion
library and load it at runtime. In case the QVL library can't be found,
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+- Verify QvE report.
+- Verify report data (report data described as above).
+- Check QvE Identity by comparing QvE report fields against hardcode values.
+  - Check that Report.MRSIGNER equals to Hardcoded QvE MRSIGNER.
+  - Check that Misc Select, Attribute and ProdID equal to the hardcoded values,
+    Misc select and Attribute need to apply Mask before comparison.
+- Check that Report.ISVSVN >= Hardcoded ISV SVN.
+
+Options to provide QvE ISV SVN threshold:
+- Option 1: Hardcode QvE ISV SVN in enclave-side plugin. Everytime QvE
```
dthaler, line 364:
```suggestion
- Option 1: Hardcode QvE ISV SVN in enclave-side plugin. Every time the QvE
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+The proposal is to extend claims definition with new SGX related
+claims.
+
+SGX quote verification API `sgx_qv_verify_quote()` returns 2 types of errors,
+fatal errors or non-fatal errors:
+- Fatal error: such as cert chain verification failed, cert revoked, API will
+  return an error code directly
+- Non-fatal error, such as cert chain out-of-date, API will return successful, but
+  there is an output parameter to indicate the specific error code. In this case,
+  the verifier can decide whehter the SGX quote is valid based on their own policy.
```
dthaler, line 386:
```suggestion
  the verifier can decide whether the SGX quote is valid based on their own policy.
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+                                          // apply LVI mitgation in enclave
+
+Note: SGX TCB = Platform CPU HW/FW TCB + SGX PCE TCB.
+```
+If the verifier needs to provide a different quote verification policy
+beyond the policy enforced by the sgx_qv_verify_quote() API, the verifier can
+request sgx_qv_verify_quote() API to return supplemental data. The detailed
+supplemental data definition are described in above `Background` section.
+
+So we also propose to add SGX supplemental data fields the the list of claims.
```
dthaler, line 428:
```suggestion
So we also propose to add SGX supplemental data fields to the list of claims.
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+#define OE_CLAIM_SGX_PLATFORM_INSTANCE_ID "sgx_platform_instance_id"
+#define OE_CLAIM_SGX_DYNAMIC_PLATFORM "sgx_dynamic_platform"
+#define OE_CLAIM_SGX_CACHED_KEYS "sgx_cached_keys"
+#define OE_CLAIM_SGX_SMT_ENABLED "sgx_smt_enabled"
+
+#define OE_SGX_CLAIMS_COUNT 15
+```
+
+### Update `oe_sgx_extract_claims()` to remove SGX endorsement dependency
+As SGX endrosments will be retrieved by the SGX DCAP QVL library, the verifier
```
dthaler, line 461:
```suggestion
As SGX endorsements will be retrieved by the SGX DCAP QVL library, the verifier
```

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+      sgx_target_info_t app_enclave_target_info;
+      sgx_report_t qve_report;
+    }sgx_ql_qe_report_info_t;
+    ```
+    - Call OCALL function `oe_verify_quote_ocall()` to host-side plugin library
+      to verify SGX quote.
+    - With all the QvE return values, call new API.
+      `oe_verify_qve_report_and_identity()` (described below) to
+      verify QvE report and identity.
+- In API `_verify_evidence()`, remove function call `oe_get_sgx_endorsements()`,
```
yentsanglee, line 308: Current implementation, oe_get_sgx_endorsements() is **not** called, if caller provides an endorsement buffer.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+  and only call `oe_parse_sgx_endorsements()` if caller provides an endorsement buffer
+- In API `_verify_report()` and relevant internal APIs, such as
```
yentsanglee, line 310: Is this going to break oe_veriy_report()?

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+comes from a trusted QvE.
+
+Intel® SGX SDK provides a library named `sgx_dcap_tvl` to help the verifier to
+verify QvE's identity, you can refer to [source file](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteVerification/dcap_tvl/sgx_dcap_tvl.cpp)
+
+This library uses hardcoded QvE identity values, because:
+- There is no need for x509 and JSON parser in verifier enclave's TCB.
+- Most of QvE's identity attributes only change infrequently.
+
+The only identity info may change more often is QvE's ISV SVN, so the verifier needs
```
yentsanglee, line 326: Isn't it QVL's responsibility for checking current QvE's product id, ISV SVN, etc against QvE Identity data from [QvE Identity at Intel® Provisioning
Cert Server](https://api.portal.trustedservices.intel.com/documentation#pcs-qve-identity-v2)? I am afraid hardcoded QvE identity values do not scale.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: It's not the same report in oe_verify_report(). This is the Intel(R) QvE output report, which target to user's application enclave, only user's application enclave can verify this report.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Because QVL is an untrusted component, user cannot rely on QVL to verify QvE's identity. The good practice is to verify QvE's identity in user's enclave.

In the other word, an attacker can use faked QVL & QvE, if QvE's identity is checked by QVL, then user would have no way to find it's a faked QvE.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+Note: SGX TCB = Platform CPU HW/FW TCB + SGX PCE TCB.
+```
+If the verifier needs to provide a different quote verification policy
+beyond the policy enforced by the sgx_qv_verify_quote() API, the verifier can
+request sgx_qv_verify_quote() API to return supplemental data. The detailed
+supplemental data definition are described in above `Background` section.
+
+So we also propose to add SGX supplemental data fields the the list of claims.
+
+[PR 3353](https://github.com/openenclave/openenclave/pull/3353) added SGX
```
yentsanglee, line 430: There is at least one customer who is interested in raw TCB Info in JSON format. It may not be a hard requirement.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Yes. The proposal is about ask Intel(R) QVL to retrieve endorsements.
- If caller provides endorsements, then it will pass the endorsements to QVL, then QVL/QvE will use caller provided endorsements to verify quote
- If caller doesn't provide endorsements, we will set the endorsements buffer to NULL, then QVL will try to retrieve endorsements automatically

So we don't need function `oe_get_sgx_endorsements()` in both cases.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: May or May not?
If it's a hard requirement to return raw collateral data, then I need to consider the alternatives.

hyjiang:
@yentsanglee , my comments as below.

1. oe_get_evidence() may return endorsements. Internally, it calls oe_get_sgx_endorsements() and retrieves the endorsements from QPL directly. With the proposal, does OE call sgx_dcap_retrieve_verification_collateral() to retrieve the same information?

[HY] With the proposal, OE doesn't need to care about SGX endorsements. If caller provide endorsements, then pass it to QVL. If caller doesn't provide endorsements, then pass NULL to QVL, QVL will try to get endorsements in it's internal logic.
Then SGX QVL/QvE would return supplemental data (with integrity protection), the proposal is to add the supplemental data into custom claims, instead of return endorsements. I wrote some reasons in doc.

Also as you said, there is at least one customer to request raw TCB info. But I think it doesn't make sense to return the endorsements directly in claims.
- Endorsement itself is not from trusted source, we cannot return the retrieved buffer directly, otherwise, user may implement his    
  own policy based on hacked endorsements. So only verified endorsement can be  trusted, also we need to guarantee the return 
  buffer is same with the verified buffer (with integrity protection)
- Parsing endorsements is NOT a simple task for user. That's why SGX QVL/QvE returns supplemental data. 
  BTW, we can extend the supplemental data based on customer's request. There are some existing supplemental data are 
  requested by Microsoft in fact.   

2. Two optional claims "validity_from" and "validity_until" are calculated while parsing TCB Info. In the proposal, will these two claims still be provided?

[HY] My assumption is that these 2 claims are NOT optional, it's in claims definition. Pls correct me if my understanding is wrong.
In addition, the 2 claims are already in SGX supplemental data, so it's easy to add them into SGX claims.

3. When OE verifies SGX quote by consulting TCB Info, "issueDate" and "nextUpdate" checking is ignored (see function _read_tcb_info()). This is because customers do not update SGX software components (i.e. QE/PCE) as frequently. Data center has policies to enforce software update in its own cadence but not checking "issueDate" and "nextUpdate" is critical.

[HY] I don't think ignore "issueDate" and "NextUpdate" is a good practice, user should know the TCB status explicitly, up to date, out of date or something else.  It's up to user to ignore the error or not.
In SGX, we designed 2 types of errors, fatal error and non-fatal error. For example, the "out of date" error is non-fatal error, when it happened, verification API return success, but there is an output parameter to indicate the "out of date" error, then user can know current TCB is out of date, but it's up to him to decide whether current SGX platform and SW are trusted or not.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
CodeMonkeyLeet: I think @yentsanglee's point is a little more general than the specific `_verify_report()` function call; the policy enforcement was implicitly integrated into our quote verification logic in the old codepaths, but the delegation of that responsibility to the QVL/QVE means that there is potentially extra work to maintain parity behavior with the old `oe_get_report()`. 

For example, the `__oe_sgx_set_minimum_crl_tcb_issue_date()` is still a valid public API (in bits) even though we hide it from the doxygen now. It would be helpful to have a  section that briefly talks about what behaviors change with `oe_verify_report()` if it just ends up using this converged code path with the `oe_get_evidence()` APIs.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
CodeMonkeyLeet: Just to clarify, this is just a description of the Intel implementation of the enclave side trusted library for using the QVE, and we're just porting it into OE, correct? (i.e. we're not changing the internal logic of `sgx_verify_qve_report_and_identity()` when we import it as `oe_verify_qve_report_and_identity()`)

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+- Check QvE Identity by comparing QvE report fields against hardcode values.
+  - Check that Report.MRSIGNER equals to Hardcoded QvE MRSIGNER.
+  - Check that Misc Select, Attribute and ProdID equal to the hardcoded values,
+    Misc select and Attribute need to apply Mask before comparison.
+- Check that Report.ISVSVN >= Hardcoded ISV SVN.
+
+Options to provide QvE ISV SVN threshold:
+- Option 1: Hardcode QvE ISV SVN in enclave-side plugin. Everytime QvE
+            ISV SVN is updated, the plugin library implementation will need to
+            be updated to stay in sync.
```
CodeMonkeyLeet, line 366: Do we have a strategy for keeping this logic in sync with the QVL/QVE releases? The ship vehicle is not explicitly mentioned (I expect it's part of the dcap_ql package) so it might be worth pointing out whether the package versioning dependency is separate from the QPL. 

I'll note that this recommendation seems to contradict an earlier desire for the QVL to handle endorsements so that OE doesn't have to keep in sync with the Intel changes to endorsements (which seem less frequent that updates to the libraries?)

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+verify QvE's identity, you can refer to [source file](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteVerification/dcap_tvl/sgx_dcap_tvl.cpp)
+
+This library uses hardcoded QvE identity values, because:
+- There is no need for x509 and JSON parser in verifier enclave's TCB.
+- Most of QvE's identity attributes only change infrequently.
+
+The only identity info may change more often is QvE's ISV SVN, so the verifier needs
+to provide a SVN number as threshold, only when current QvE's ISV SVN is equal to
+or larger than this threshold, the verifier can trust the QvE verification result.
+The verifier can get latest QvE ISV SVN from [QvE Identity at Intel® Provisioning
```
CodeMonkeyLeet, line 329: This seems to be a direct parallel to the QEIdentity, which we explicitly treat as part of the collateral bundle handled during attestation. Is there a reason why this isn't treated the same way?

docs/DesignDocs/SGX_QuoteVerify_Integration.md
yentsanglee: It may not. Let me get back to you on this.

yentsanglee:
> * Endorsement itself is not from trusted source, we cannot return the retrieved buffer directly, otherwise, user may implement his
>   own policy based on hacked endorsements. So only verified endorsement can be  trusted, also we need to guarantee the return
>   buffer is same with the verified buffer (with integrity protection)

We have a first-party customer who relies on the endorsements output from calling oe_get_evidence(). They have special need to treat stale evidence as valid if it was valid when it was generated.

yentsanglee:
> [HY] I don't think ignore "issueDate" and "NextUpdate" is a good practice, user should know the TCB status explicitly, up to date, out of date or something else. It's up to user to ignore the error or not.
> In SGX, we designed 2 types of errors, fatal error and non-fatal error. For example, the "out of date" error is non-fatal error, when it happened, verification API return success, but there is an output parameter to indicate the "out of date" error, then user can know current TCB is out of date, but it's up to him to decide whether current SGX platform and SW are trusted or not.

oe_verify_evidence() returns success or failure. We can introduce warning results but there must be option to ignore TCB Info expiration. SGX Updates TCB Info every month. Getting security update early is a good thing but to some customers who don't care as much, it's too aggressive. Imaging being forced to reboot your computer for security update in the middle of something.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: @CodeMonkeyLeet , yes, the proposal is about porting the trusted library into OE, and the internal logic would not changed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+} sgx_ql_qv_supplemental_t;
+```
+
+## Proposed Changes
+
+### Options for retrieving SGX endorsements
+There are two options for the OE SDK plugin library to retreive SGX endorsments.
+
+#### Option 1: Keep existing implemention in SGX verifier plugin
+In the current plugin implementation, if the verifier doesn't provide SGX endoresements, it
```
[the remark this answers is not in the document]
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+calls API `oe_get_sgx_endorsements()` to parse SGX ECDSA quote body to get PCK
+cert chain first, then calls OCALL API `oe_get_quote_verification_collateral_ocall()`
+to load Quote Provide Library(QPL) and connect to PCK Cert Caching Sever(PCCS)
+to get corresponding verification collaterals, including CRL, TCB and QE
+Identity and etc.
+
+#### Option 2: Ask DCAP QVL to retrieve endorsements
+If the verifier doesn't provide endorsments when calling DCAP QVL API, the QVL API
```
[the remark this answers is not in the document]
hyjiang: Fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+        sgx_ql_qv_result_t quote_verification_result,
+        const uint8_t *p_supplemental_data,
+        uint32_t supplemental_data_size,
+        sgx_isv_svn_t qve_isvsvn_threshold)
+```
+All the data returned by the QvE are included in the QvE report data field for
+integrity protection.
+```C
+QvE report_data = SHA256([nonce || quote || expiration_check_date ||
+expiration_status || verification_result || supplemental_data] || 32 - 0x00)
```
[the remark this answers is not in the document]
hyjiang: - The hash in report data is used for guarantee the QvE output would NOT be changed during transition. All the items in 
   square brackets are QvE outputs, we will add the output hash into QvE report data. Then application enclave can 
   calculate a hash by using QvE outputs, then compare the hash with value in report data. In this way, application enclave  
   can make sure they are real QvE output. 
- SHA256 algorithm generates a fixed size 256-bit (32 bytes) hash, but SGX report data length is 64 bytes. So we will leave 
  the remaining 32 bytes as 0 in report data. 

I will add more description later.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+  - Check that Report.MRSIGNER equals to Hardcoded QvE MRSIGNER.
+  - Check that Misc Select, Attribute and ProdID equal to the hardcoded values,
+    Misc select and Attribute need to apply Mask before comparison.
+- Check that Report.ISVSVN >= Hardcoded ISV SVN.
+
+Options to provide QvE ISV SVN threshold:
+- Option 1: Hardcode QvE ISV SVN in enclave-side plugin. Everytime QvE
+            ISV SVN is updated, the plugin library implementation will need to
+            be updated to stay in sync.
+- Option 2: Update existing enclave-side plugin API to allow the verifier to input
```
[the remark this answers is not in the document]
hyjiang: Thanks. I think add an input in `oe_policy_t` is a possible solution, in this way, we don't need to change public API.
But it's only about Intel(R) QvE, so it's not TEE agnostic. 

Your option 3 still need to change public API or update `oe_policy_t`.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,488 @@
...
+- Non-fatal error, such as cert chain out-of-date, API will return successful, but
+  there is an output parameter to indicate the specific error code. In this case,
+  the verifier can decide whehter the SGX quote is valid based on their own policy.
+
+We propose to introduce a new SGX-specific claim `OE_CLAIM_SGX_QUOTE_VERIFICATION_STATUS` to
+indicate QVL/QvE return status. The verifier can check the status to see if there
+is a non-fatal error during quote verification. This claim can have one of three values:
+
+  - OE_OK - Quote verification passed, and TCB level is up to date
+  - Non-Fatal error code - API returns OE_OK,
```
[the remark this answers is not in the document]
hyjiang: I agree with you. 
But @yentsanglee pointed out that some customers may not care about the non-fatal (warning) errors, maybe return OE_OK is a good choice, then only the customer who want have a strict verification policy, he/she can check the additional status in SGX claims.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: For the QvE identity verification trusted library, basically we don't need to keep sync with QVL/QvE release. Because nearly all the info in QvE identity would NOT change, except there is a critical issue, such as QvE signing key is leaked.  
But note that QvE ISV SVN is the only exception, it may change from time to time. 

Below QvE identity for example, you will see that except `isvsvn`, all other items are about QvE attribute and mrsigner.
```C
"miscselect":"00000000","miscselectMask":"FFFFFFFF","attributes":"01000000000000000000000000000000","attributesMask":"FBFFFFFFFFFFFFFF0000000000000000","mrsigner":"8C4F5775D796503E96137F77C68A829A0056AC8DED70140B081B094490C57BFF","isvprodid":2,"tcbLevels":[{"tcb":{"isvsvn":3},"tcbDate":"2019-12-11T00:00:00Z","tcbStatus":"UpToDate...
```

In addition, all other SGX endorsements, such as PCK Cert, CRL, TCB info, etc. The internal fields may change from time to time, that's why I propose to get rid of the complex parsing tasks in OE.

hyjiang:
> > * Endorsement itself is not from trusted source, we cannot return the retrieved buffer directly, otherwise, user may implement his
> >   own policy based on hacked endorsements. So only verified endorsement can be  trusted, also we need to guarantee the return
> >   buffer is same with the verified buffer (with integrity protection)
> 
> We have a first-party customer who relies on the endorsements output from calling oe_get_evidence(). They have special need to treat stale evidence as valid if it was valid when it was generated.

Yes, I see. But my point is that if we really need to return SGX endorsements, we need to verify the quote with the endorsements first, then make sure the returned endorsements are same with the verified endorsements. In this way, we can guarantee the returned endorsements are valid.

hyjiang:
> > [HY] I don't think ignore "issueDate" and "NextUpdate" is a good practice, user should know the TCB status explicitly, up to date, out of date or something else. It's up to user to ignore the error or not.
> > In SGX, we designed 2 types of errors, fatal error and non-fatal error. For example, the "out of date" error is non-fatal error, when it happened, verification API return success, but there is an output parameter to indicate the "out of date" error, then user can know current TCB is out of date, but it's up to him to decide whether current SGX platform and SW are trusted or not.
> 
> oe_verify_evidence() returns success or failure. We can introduce warning results but there must be option to ignore TCB Info expiration. SGX Updates TCB Info every month. Getting security update early is a good thing but to some customers who don't care as much, it's too aggressive. Imaging being forced to reboot your computer for security update in the middle of something.

For warning results, this proposal is that `oe_verify_evidence()` will return OE_OK, and there will be an additional `verification_status` in custom claim, so only the customer who care about the warning results, he can handle the `verification_status` in claims.

hyjiang:
> Thanks for writing this up, Hongyan, a lot of good details to think about here!
> 
> To prioritize my feedback, the most important questions I'm looking to have answered are the difference in handling of the QVE identity vs. the existing QE identity handling, and some elaboration on potential impact to the legacy `oe_verify_report()` API.
> 
> Also, could you copy-paste this document into Word and just fix up all the minor spelling/grammar errors? I think it's distracting for some reviewers and may be preventing them from getting to the more important content to review.

- QE is the enclave which generate attestation key, and QE identity and attestation key (pub key and signature) are part of SGX  
  quote, so QE identity verification would be done during quote verification
- QvE is the enclave which helps user to do quote verification. But after quote verification, user need to make sure the QvE he 
   used is a Intel(R) released and latest QvE. Otherwise, attacker can use a faked QvE to do quote verification.
   
About the potential `oe_verify_report()` change. Maybe we have some misunderstand here.
In SGX, app enclave can sent it's own `target info` and a `nonce` to `verify_quote()` API, then QvE can use the `target info` to generate a QvE report, then app enclave can verify this report (local attestation), then verify the QvE identity. In this way, User can make sure the QvE he used is a trusted QvE.
You can see that we need to call `oe_verify_report()` to verify a local QvE report, it's a local attestation process.  
So my understanding is that we only need to update remote report (quote) verification part in API `oe_verify_report()`.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
shnwc: One question would be what this customer wants the raw TCB info blob for? If their real requirements are the individual fields in this blob that we will provide as a set of claims, then this proposal would meet their requirements better.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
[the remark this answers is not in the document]
shnwc: In current OE SDK API design, if a function returns some error code other than `OE_OK`, does it always abort and not return other results in output parameters?

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -208,55 +207,40 @@ typedef struct _sgx_ql_qv_supplemental_t
 
 ## Proposed Changes
 
-### Options for retrieving SGX endorsements
-There are two options for the OE SDK plugin library to retrieve SGX endorsements.
+In order to somoothly switch quote verification from OE to SGX DCAP QVL/QvE, we
```
achamayou, line 210:
```suggestion
In order to smoothly switch quote verification from OE to SGX DCAP QVL/QvE, we 
```

hyjiang: Thanks, fixed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,522 @@
...
+## Proposed Changes
+
+In order to smoothly switch quote verification from OE to SGX DCAP QVL/QvE, we
+propose to implement the change in 2 steps as below:
+- Step 1: Implement QVL-based quote verification and make it co-exist with the
+  existing implementation
+- Step 2: Remove existing implementation and makes the DCAP QVL a hard dependency
+
+### Step 1: Implement QVL-based quote verification and make it co-exist with the existing implementation
+If SGX DCAP QVL/QvE are not installed, quote verification will continue to use existing
```
yentsanglee, line 217: How do you uninstall quote verification on Windows?

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,522 @@
...
+      - Call OCALL function `oe_verify_quote_ocall()` to host-side plugin library
+        to verify SGX quote.
+      - With all the QvE return values, call new API.
+        `oe_verify_qve_report_and_identity()` (described below) to
+        verify QvE report and identity
+  - If DCAP QVL returns error to indicate that QVL doesn't not exist on current
+    platform, then automatically fallback to the existing implementation
+  - Add log information to indicate whether the QVL-based or existing implementation
+    is used
+- If QVL-based quote verification is used, in order to exactly align with current
```
yentsanglee, line 291: Current implementation also accepts "SWHardeningNeeded" as a success case. See [tcbinfo.c](https://github.com/openenclave/openenclave/blob/master/common/sgx/tcbinfo.c#L243).

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: QVL/QvE are part of DCAP QL installer in Windows, so QVL/QvE would be uninstalled when user uninstall DCAP QL.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Thanks, I will update the design doc to address this issue

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,522 @@
...
+to get corresponding verification collaterals, including CRL, TCB and QE
+Identity, etc.
+
+##### Option 2: Ask DCAP QVL to retrieve endorsements
+If the verifier doesn't provide endorsements when calling DCAP QVL API, the QVL API
+implementation will parse quote body to get PCK cert chain, then load QPL
+and connect to PCCS to get corresponding verification
+collaterals.
+
+##### Proposal: Ask DCAP QVL to retrieve endorsements
```
yentsanglee, line 466: If QVL/QvE code path is chosen, the endorsements are going to be impacted. So this really is a decision to be made in step 1. Right?

yentsanglee (commented):
I like the step approach. This has to proceed with care in order not to break Azure attestation. Mentioned before but re-iterating:
1. All current claim outputs have to be preserved. This includes "validity_from" and "validity_until".
2. Expiration date of TCB Info ("issueDate" and "nextUpdate") can be optionally ignored. This is critical.
3. Error logging in QVL/QvE must be made easy. 24/7 on-call supports must be able to diagnose by reading the logs.
4. Issue mitigation is critical. If a bug is identified in QVL/QvE, who will fix it and how will it be deployed?

docs/DesignDocs/SGX_QuoteVerify_Integration.md
yentsanglee: > QVL/QvE are part of DCAP QL installer in Windows, so QVL/QvE would be uninstalled when user uninstall DCAP QL.

Is DCAP QL also for quote generation? If so, the user cannot uninstall it.

shnwc:
> I like the step approach. This has to proceed with care in order not to break Azure attestation. Mentioned before but re-iterating:
> 
> 1. All current claim outputs have to be preserved. This includes "validity_from" and "validity_until".
> 2. Expiration date of TCB Info ("issueDate" and "nextUpdate") can be optionally ignored. This is critical.
> 3. Error logging in QVL/QvE must be made easy. 24/7 on-call supports must be able to diagnose by reading the logs.
> 4. Issue mitigation is critical. If a bug is identified in QVL/QvE, who will fix it and how will it be deployed?

I think items 1, 2 should be taken care in step 1, items 3 and 4 can wait for step 2 to be fully addressed. Step 1 can be seen as adding an experimental feature.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: The endorsements change will be in step 2. In steps, it will pass endorsements from OE to QVL directly.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: BTW, the comment in [tcbinfo.c](https://github.com/openenclave/openenclave/blob/master/common/sgx/tcbinfo.c#L243) is not accurate, only on the old platform which will be impacted by LVI, Intel will mark UpToDate tcb as SWHardeningNeeded

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: Yes, DCAP QL is for quote generation.

hyjiang:
> > I like the step approach. This has to proceed with care in order not to break Azure attestation. Mentioned before but re-iterating:
> > 
> > 1. All current claim outputs have to be preserved. This includes "validity_from" and "validity_until".
> > 2. Expiration date of TCB Info ("issueDate" and "nextUpdate") can be optionally ignored. This is critical.
> > 3. Error logging in QVL/QvE must be made easy. 24/7 on-call supports must be able to diagnose by reading the logs.
> > 4. Issue mitigation is critical. If a bug is identified in QVL/QvE, who will fix it and how will it be deployed?
> 
> I think items 1, 2 should be taken care in step 1, items 3 and 4 can wait for step 2 to be fully addressed. Step 1 can be seen as adding an experimental feature.

Yes, step 1 will keep current claims unchanged.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,525 @@
...
+
+```C
+oe_result_t oe_verify_quote_ocall(
+    [in] const oe_uuid_t* format_id,
+    [in, size=opt_params_size] const void* opt_params,
+    size_t opt_params_size,
+    [in, size=quote_size] const void* p_quote,
+    size_t quote_size,
+    oe_datetime_t expiration_check_date,
+    [out] uint32_t *p_collateral_expiration_status,
```
dthaler, line 254: This seems odd to have to do things in 2 passes, one to get untrusted values and then oe_verify_qve_report_and_identity() to verify those values are correct.  I didn't follow why this two-phase process was needed.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,525 @@
...
+to provide a SVN number as threshold, only when current QvE's ISV SVN is equal to
+or larger than this threshold, the verifier can trust the QvE verification result.
+The verifier can get latest QvE ISV SVN from [QvE Identity at Intel® Provisioning
+Cert Server](https://api.portal.trustedservices.intel.com/documentation#pcs-qve-identity-v2)
+
+As OE SDK cannot use Intel® SGX SDK trusted library directly, we need to port
+this library into OE SDK as part of enclave-side plugin.
+
+```C
+oe_result_t oe_verify_qve_report_and_identity(
```
dthaler, line 319: To confirm: are you proposing this as an _internal_ API?

docs/DesignDocs/SGX_QuoteVerify_Integration.md
```diff
@@ -0,0 +1,525 @@
...
+in SGX collateral definition.
+
+
+#### Update existing plugin functions to only use DCAP QVL/QvE
+- In API `_verify_evidence()`, remove function call `oe_get_sgx_endorsements()`,
+  and only call `oe_parse_sgx_endorsements()` if caller provides an endorsement buffer
+- In API `_verify_report()` and relevant internal APIs, such as
+  `oe_verify_report_internal()` and `oe_verify_sgx_quote()`, remove the code that
+  loads QPL and retrieves endorsements
+- There is an API `__oe_sgx_set_minimum_crl_tcb_issue_date()`in header file
```
dthaler, line 493: Just FYI, double-underscore shouldn't be used as a prefix (and this is noted in the OE developer guidelines), so at some point that function should be renamed.

hyjiang: > Just FYI, double-underscore shouldn't be used as a prefix (and this is noted in the OE developer guidelines), so at some point that function should be renamed.

Note that this is an existing API in current implementation

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: > To confirm: are you proposing this as an _internal_ API?

Yes, this is an internal API used by verifier plugin to verify QvE identity.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
hyjiang: > This seems odd to have to do things in 2 passes, one to get untrusted values and then oe_verify_qve_report_and_identity() to verify those values are correct. I didn't follow why this two-phase process was needed.

The `untrusted values` you mentioned are QvE output. When verifier uses QvE to verify the SGX quote, he/she needs to confirm 2 things.
- QvE is trusted and up to date
- The outputs are really come from QvE, and they have not been changed

So the 2nd phase is needed for above two items.

shnwc (approved):

docs/DesignDocs/SGX_QuoteVerify_Integration.md
yentsanglee: It's a hard requirement but I understand in step 1 the raw TCB Info will be kept.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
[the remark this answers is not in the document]
yentsanglee: Current OE SDK API design, OE_OK is the only success return value. Attestation APIs follow the same pattern. My only concern is backward compatibility. Otherwise, I generally agree with @CodeMonkeyLeet and @shnwc that the caller can make its own judgement by all the returned value.

yentsanglee (approved):

yentsanglee:
@hyjiang please squash all commits so it can be merged to main.

docs/DesignDocs/SGX_QuoteVerify_Integration.md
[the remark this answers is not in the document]
CodeMonkeyLeet: To recap a point in one of the SIG-attestation meetings: OE SDK currently makes specific policy decisions to ignore some verification issues on behalf of the caller. These decisions align with Azure expectations due to how it manages its SGX configuration, but should not be the default in a general SDK intended for multi-cloud/provider use like OE moving forwards. In that regard, a break with backwards compatibility is necessary here, and we should just manage it as we would other breaking changes to minimize the impact to the current users.

hyjiang:
> @hyjiang please squash all commits so it can be merged to main.

Done, pls help to merge.

qiucwang (approved):
LGTM

PR #3418, opened 2020-08-19
Add lfences to accompany oe_is_within_enclave, oe_is_outside_enclave calls.

### LINKED ISSUE #3401: Speculative execution mitigations: Calls to oe_is_outside_enclave should be followed with oe_lfence
Care must be taken with host<->enclave interfaces to protect against **speculative execution threats**.
Whenever an enclave bounds-checks memory to ensure it is _outside_ of the enclave with a call to `oe_is_outside_enclave()`, it should follow this validation with a call to `oe_lfence()` to stop speculative execution along that logic path. This guidance should be followed for both application and openenclave code, with exceptions expected in the low minority.

For example, with `oe_ecall_context_get_ocall_buffer()`, speculative execution may proceed past the bounds check of `oe_is_outside_enclave()`, as described below. 

```
void* oe_ecall_context_get_ocall_buffer(uint64_t size)
{
    oe_ecall_context_t* ecall_context = _get_ecall_context();
    if (ecall_context)
    {
        // Copy to volatile variables to prevent TOCTOU attacks.
        uint8_t* volatile ocall_buffer = ecall_context->ocall_buffer;
        volatile uint64_t ocall_buffer_size = ecall_context->ocall_buffer_size;
        if (ocall_buffer_size >= size &&
            oe_is_outside_enclave(ocall_buffer, ocall_buffer_size))
            return (void*)ocall_buffer;  <-----------------speculative execution may exercise this 
                                                           code path even if ocall_buffer is _inside_ the enclave,
                                                           allowing a potential side-channel leak in caller logic.
    }
    return NULL;
}
```

anakrish:
fixes #3401 

### DIFF
--- a/enclave/core/sgx/hostcalls.c
+++ b/enclave/core/sgx/hostcalls.c
@@ -14,9 +14,13 @@ static oe_ecall_context_t* _get_ecall_context()
 {
     oe_sgx_td_t* td = oe_sgx_get_td();
     oe_ecall_context_t* ecall_context = td->host_ecall_context;
-    return oe_is_outside_enclave(ecall_context, sizeof(*ecall_context))
-               ? ecall_context
-               : NULL;
+    if (!oe_is_outside_enclave(ecall_context, sizeof(*ecall_context)))
+        return NULL;
+
+    // Prevent speculative execution from accessing ecall_context that
+    // has failed the !oe_is_outside_enclave check.
+    oe_lfence();
+    return ecall_context;
 }
 
 /**
@@ -42,7 +46,13 @@ void* oe_ecall_context_get_ocall_buffer(uint64_t size)
         volatile uint64_t ocall_buffer_size = ecall_context->ocall_buffer_size;
         if (ocall_buffer_size >= size &&
             oe_is_outside_enclave(ocall_buffer, ocall_buffer_size))
+        {
+            // Prevent speculative execution from accessing ocall_buffer that
+            // has failed the oe_is_outside_enclave check.
+            oe_lfence();
+
             return (void*)ocall_buffer;
+        }
     }
     return NULL;
 }
--- a/3rdparty/libunwind/Gstep.c
+++ b/3rdparty/libunwind/Gstep.c
@@ -10,6 +10,7 @@
 #endif
 
 #include <libunwind.h>
+#include <openenclave/edger8r/enclave.h>
 #include <openenclave/enclave.h>
 #include "unwind_i.h"
 
@@ -36,6 +37,10 @@ int __libunwind_unw_step(unw_cursor_t* cursor)
         if (!oe_is_within_enclave((void*)c->cfa, 1024))
             return 0;
     }
+
+    // Prevent speculative execution from accessing cursor that
+    // has failed the oe_is_within_enclave checks.
+    oe_lfence();
 
     return unw_step(cursor);
 }
--- a/enclave/core/hostcalls.c
+++ b/enclave/core/hostcalls.c
@@ -64,6 +64,9 @@ void* oe_host_malloc(size_t size)
     if (arg_out && !oe_is_outside_enclave((void*)arg_out, size))
         oe_abort();
 
+    // Prevent speculative execution from accessing arg_out that
+    // has failed the oe_is_outside_enclave check.
+    oe_lfence();
     return (void*)arg_out;
 }
 
@@ -107,6 +110,9 @@ void* oe_host_realloc(void* ptr, size_t size)
         oe_abort();
     }
 
+    // Prevent speculative execution from accessing retval that
+    // has failed the oe_is_outside_enclave check.
+    oe_lfence();
     return retval;
 }
 
--- a/enclave/core/sgx/backtrace.c
+++ b/enclave/core/sgx/backtrace.c
@@ -3,6 +3,7 @@
 
 #include <openenclave/advanced/allocator.h>
 #include <openenclave/corelibc/stdio.h>
+#include <openenclave/edger8r/enclave.h>
 #include <openenclave/enclave.h>
 #include <openenclave/internal/argv.h>
 #include <openenclave/internal/backtrace.h>
@@ -67,6 +68,10 @@ const void* _check_address(const void* ptr)
 {
     if (!oe_is_within_enclave(ptr, sizeof(uint64_t)))
         return NULL;
+
+    // Prevent speculative execution from accessing ptr that
+    // has failed the oe_is_within_enclave check.
+    oe_lfence();
 
     return ptr;
 }
--- a/enclave/core/sgx/calls.c
+++ b/enclave/core/sgx/calls.c
@@ -166,6 +166,11 @@ static oe_result_t _handle_init_enclave(uint64_t arg_in)
             if (!oe_is_outside_enclave(enclave, 1))
                 OE_RAISE(OE_INVALID_PARAMETER);
 
+            // Eventhough oe_enclave_t* is an opaque type that is never
+            // dereferenced within the enclave, for consistency add an oe_lfence
+            // to accompany the oe_is_outside_enclave check.
+            oe_lfence();
+
             oe_enclave = enclave;
 
             /* Initialize the CPUID table before calling global constructors. */
@@ -224,6 +229,10 @@ oe_result_t oe_handle_call_enclave_function(uint64_t arg_in)
         args.output_buffer_size < sizeof(oe_result_t) ||
         !oe_is_outside_enclave(args.output_buffer, args.output_buffer_size))
         OE_RAISE(OE_INVALID_PARAMETER);
+
+    // Prevent speculative execution from accessing parameters that
+    // have failed the oe_is_outside_enclave check.
+    oe_lfence();
 
     // Validate output and input buffer sizes.
     // Buffer sizes must be correctly aligned.
@@ -576,6 +585,10 @@ static void _exit_enclave(uint64_t arg1, uint64_t arg2)
             oe_is_outside_enclave(
                 host_ecall_context, sizeof(*host_ecall_context)))
         {
+            // Prevent speculative execution from accessing context that
+            // has failed the oe_is_outside_enclave check.
+            oe_lfence();
+
             uint64_t* frame = (uint64_t*)__builtin_frame_address(0);
             host_ecall_context->debug_eexit_rbp = frame[0];
             // The caller's RSP is always given by this equation
--- a/enclave/core/sgx/exception.c
+++ b/enclave/core/sgx/exception.c
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 #include <openenclave/bits/sgx/sgxtypes.h>
+#include <openenclave/edger8r/enclave.h>
 #include <openenclave/enclave.h>
 #include <openenclave/internal/context.h>
 #include <openenclave/internal/cpuid.h>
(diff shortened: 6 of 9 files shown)


anakrish:
bors try

enclave/core/sgx/hostcalls.c
```diff
@@ -42,7 +42,16 @@ void* oe_ecall_context_get_ocall_buffer(uint64_t size)
         volatile uint64_t ocall_buffer_size = ecall_context->ocall_buffer_size;
         if (ocall_buffer_size >= size &&
             oe_is_outside_enclave(ocall_buffer, ocall_buffer_size))
+        {
+            // Prevent speculative execution from treating enclave memory as
+            // ocall_buffer. Without the lfence, even when oe_is_outside_enclave
+            // is false, speculative execution could return ocall_buffer
+            // (enclave memory) here and go on to access it later, potentially
+            // allowing side-channels.
+            oe_lfence();
```
mingweishih, line 51: I think this pattern has already been covered by speculative-load-hardening flag from clang according to https://llvm.org/docs/SpeculativeLoadHardening.html#accumulating-predicate-state.

I checked the assembly code of this function
```
  bed85:       e8 66 08 00 00          callq  bf5f0 <oe_is_outside_enclave>
   bed8a:       48 89 e1                mov    %rsp,%rcx
   bed8d:       48 c1 f9 3f             sar    $0x3f,%rcx
   bed91:       a8 01                   test   $0x1,%al
   bed93:       48 89 4d a8             mov    %rcx,-0x58(%rbp)
   bed97:       0f 85 1a 00 00 00       jne    bedb7 <oe_ecall_context_get_ocall_buffer+0x107>
   bed9d:       e9 00 00 00 00          jmpq   beda2 <oe_ecall_context_get_ocall_buffer+0xf2>
   beda2:       48 8b 45 a8             mov    -0x58(%rbp),%rax
   beda6:       48 8b 4d d0             mov    -0x30(%rbp),%rcx
   *bedaa:       48 0f 45 c1             cmovne %rcx,%rax                  ; path 1
   bedae:       48 89 45 b0             mov    %rax,-0x50(%rbp)
   bedb2:       e9 1d 00 00 00          jmpq   bedd4 <oe_ecall_context_get_ocall_buffer+0x124>
   bedb7:       48 8b 45 a8             mov    -0x58(%rbp),%rax
   bedbb:       48 8b 4d d0             mov    -0x30(%rbp),%rcx
   *bedbf:       48 0f 44 c1             cmove  %rcx,%rax                   ; path 2
   bedc3:       48 8b 55 e0             mov    -0x20(%rbp),%rdx
   bedc7:       48 89 55 f8             mov    %rdx,-0x8(%rbp)
   bedcb:       48 89 45 a0             mov    %rax,-0x60(%rbp)
   bedcf:       e9 1d 00 00 00          jmpq   bedf1 <oe_ecall_context_get_ocall_buffer+0x141>
   bedd4:       48 8b 45 b0             mov    -0x50(%rbp),%rax
   bedd8:       48 89 45 c0             mov    %rax,-0x40(%rbp)
```

As outlined in the code, both paths are guided by cmov* instructions already. Can you double check?

anakrish: @mingweishih 
Yes, when using clang the issue is already mitigated. 

However there are few advantages to explicitly hand mitigating it:
- We'd have mitigation for enclaves compiled using GCC. 
- It clearly tags those functions that deal with host controlled data.
- It is consistent with oeedger8r which also emits lfences.

What do you think?

mingweishih: Yes, it makes sense to cover the GCC case. However, given that this is the known gap between gcc and clang, do we want to manually analyze/fix all of potential cases (like @douglasmaciver mentioned below, which I haven't look into)?

mingweishih: Also note that, when enabling LVI mitigation, using gcc will also mitigate those vulnerable patterns (i.e., we could think the LVI mitigation as a superset of other speculation attack mitigation).

anakrish: Good point. Do we publish only LVI mitigated libraries?

mingweishih: We currently ship both (normal and mitigated) versions and make it optional for developer to link against.

douglasmaciver: @mingweishih I see broad benefit from compiler-injected hardening and I'm not suggesting that the approach doesn't provide protection. For example, compiler-injected hardening is good for large existing codebases where modifying the code with explicit protections (e.g. lfence) is impractical. But there are open concerns about the full efficacy of automated approaches and even the authors of the clang protections admit that the current approach's capabilities do not protect against all cases. Moreover, and this is especially true for applications with well-defined security boundaries, it may be more performant to target just the at-risk paths with explicit protections (e.g. lfence) than to incur the penalty of broad checks in code that is not at risk.

mingweishih: This does not reflect what we do today---we enable the compiler mitigation by default with clang. The example here is a typical pattern that should be covered by the instrumentation. Not using the compiler and only relying on manually injected protection does sound attractive form the runtime performance prospective. However, doing so requires a complete analysis on the entire codebase, which I don't think we're there yet. That's to be said, I don't think the current manual inspection is enough for us to argument that compiler-based mitigation is insufficient.

anakrish: Currently our philosophy has been this
- Use compiler mitigation where compiler supports it.
- Additionally, hand mitigate the boundary functions.
- Auto-mitigate boundary functions generated by edger8r (via lfence).

Since we don't disallow the SDK to be built using GCC, we have to continue to manually mitigate the boundary functions.
LVI mitigation has known to have a significant impact on performance. Therefore, we cannot rely on LVI mitigation to mitigate boundary functions.

>> However, doing so requires a complete analysis on the entire codebase, which I don't think we're there yet.
Yes, we need to audit the code regularly.  Keep in mind that, we are interested mainly in boundary functions where the host as control of data. The set of boundary functions in SDK are finite and minimal.

mingweishih: Yes, I agree the argument that there is a known gap between GCC and clang and we could manually patch the code like what this PR does. My question/concern is whether we're clear about the scope---but this should be a blocker for this PR. I'm totally fine with patching whenever we found new one.

I would still prefer using compiler mitigation when possible for those typical patterns. Unfortunately, I didn't see there is a plan for GCC to support the equivalent option. In addition, clang-10 is going to support more performant LVI mitigation. With that, we can set it by default and then load-hardening is no longer needed.

CodeMonkeyLeet: Just my 2¢: 
* I would suggest starting to think about the costs of these kinds of mitigations; LVI has demonstrated that there is definitely a point where we won't enable these kinds of systemic mitigations by default. The clang-mitigated versions of the libraries are what we ship and it seems like adding lfences on top of compiler mitigations only have the additional perf downsides. 
* I agree that the scoping to boundary functions is reasonable for any kind of manual intervention, but I also want to be very clear that it is not an effective mitigation from a security perspective of the application. The entire app needs to have the appropriate mitigations compiled in because spectre-2 is not only exploitable at the boundary. I don't think there's a realistic scenario where a dev cares about spectre-2 mitigation, and compiles OE SDK plus their app with GCC.
* I would love to see the numbers of the more performant LVI mitigation if you could point me at them. Unless it's incredibly trivial (which, let's be honest, even spectre-2 speculative load hardening was not entirely), I don't think we should bank on replacing one with the other.

radhikaj:
@anakrish Since we distribute libs compiled with clang, I think having these additional mitigations for gcc may not be necessary.  As Ming-Wei points out, keeping hand written mitigations for gcc in sync with those generated by clang is a significant cost.

markg-intel:
One thing that hasn't been explicitly pointed out is that adding mitigations by hand, in general, removes dependencies on tools (and on versions of tools that support mitigations, etc.). This tends to suggest that hand-mitigating boundary functions is good strategy - low performance impact and robust. 

Regarding gcc, I don't think it supports SLH. Very recent versions of the GNU tool chain do support LVI mitigations and as @mingweishih said, LVI mitigations are a superset of Spectre v1 mitigations. Having said this, the new tool chain supports "full" LVI mitigation as well as mitigation of only vulnerable control flow instructions. The superset statement above is only true when full mitigations are applied.

I'm not sure what is being assumed about gcc in some of the previous comments, which suggest that gcc (or GNU tool chain) can't mitigate Spectre v1. As stated above, the GNU tool chain can mitigate Spectre v1 by applying full LVI mitigations. I'm not necessarily advocating this, just clarifying. I think the direction laid out in the previous comments as well as in the PR itself is sound.

### More comments

clang's mitigation of C/C++ source code is more performant than GNU tool chain's. For assembly source, it's the same.

Focusing on boundary functions makes sense for Spectre v1, but not for LVI.

anakrish:
Thanks for you input @markg-intel. It clarifies where GCC and clang stand in regards to Spectre V1 and LVI mitigations, and also whether focusing on boundary functions are a good strategy for Spectre V1.

@mingweishih Do we use the appropriate mitigation flags for GCC?

anakrish:
@douglasmaciver @mingweishih 

I have taken a pass through our code base and added lfences to accompany the manual (as opposed to generated by edger8r) uses of oe_is_within_enclave and oe_is_outside_enclave checks.

anakrish:
bors try

anakrish:
bors try

mingweishih:
> Thanks for you input @markg-intel. It clarifies where GCC and clang stand in regards to Spectre V1 and LVI mitigations, and also whether focusing on boundary functions are a good strategy for Spectre V1.
> 
> @mingweishih Do we use the appropriate mitigation flags for GCC?

For LVI mitigation, we support the same for both clang and GCC.

mingweishih:
> @douglasmaciver @mingweishih
> 
> I have taken a pass through our code base and added lfences to accompany the manual (as opposed to generated by edger8r) uses of oe_is_within_enclave and oe_is_outside_enclave checks.

Sounds good. I'll take another pass.

douglasmaciver:
I recommend pausing on this PR.

The reason #3401 focuses on `oe_is_outside_enclave()` is that it just happened to be the function of the within/outside pair that I began my review with -- I had not yet begun my review of callers of `oe_is_within_enclave()`. The larger context of the review is that I'm examining the _intentional_ context of the use of those functions and why a developer would choose one over the other.  As a novice OE developer, its not clear if I should check that memory is !is_within() or is_outside(), or conversely, is_within() or !outside().  What may make these conundrums dangerous is if the developer does not grok the nuance of the inequality **is_within() != !is_outside()** due to the possibility of a buffer being partially in (and partially out) of the enclave memory space. 

What exacerbates this problem, is the additional cognitive load that speculative execution threats pose to the OE developer. In the discussion of this PR we have established that explicit calls to `oe_lfence()` are necessary along the host<-->enclave interface, but I wonder if the coding pattern that we are choosing is the best for developers, especially when one considers the importance of getting this validation right.

For example, given the security boundary of host<-->enclave, why would an enclave not abort when it is expecting a buffer to be inside but determines that it is not? This is clearly a case for PANIC. Furthermore, if our recommendation is to pair these calls, with `oe_lfence()`, might there be a meta/wrapper function that does it automatically?

douglasmaciver:
We should proceed with adding lfences at the locations identified. Addressing broader issues as I identified in https://github.com/openenclave/openenclave/pull/3418#issuecomment-683231371 should be moved to the backlog.

anakrish:
We will be deprecating support for GCC #3555. 
Since the lfences in this PR are in C code which is already mitigated by clang's SLH, they are no longer needed.

PR #3422, opened 2020-08-20
Initialize allocators when thread local is empty

mingweishih:
This PR ensures that OE always initializes/cleans up the allocator regardless the size of the thread-local data (even if the size is zero).


mingweishih:
> LGTM. Thanks for identifying and fixing the issue.
> Can you create an issue to create a test once we are indeed able to create enclaves with zero tbss and tdata?

Sure, that requires some refactoring to oeenclave (currently, the tdata/tbss is always non-zero when linking against oeenclave). I'm planning to submit the test along or after that PR.

mingweishih:
bors try

PR #3423, opened 2020-08-20
Print quote library load erorr

yentsanglee:
Issue - Attestation tests hit SEGFAULT when DCAP quote library is not installed.

The fix is to check function pointers before calling them.

### DIFF
--- a/host/sgx/sgxquote.c
+++ b/host/sgx/sgxquote.c
@@ -395,8 +395,14 @@ oe_result_t oe_sgx_qe_get_target_info(
     else
     {
         _load_sgx_dcap_ql();
-        err = _sgx_qe_get_target_info((sgx_target_info_t*)target_info);
+        if (!_sgx_qe_get_target_info)
+            OE_RAISE_MSG(
+                OE_QUOTE_LIBRARY_LOAD_ERROR,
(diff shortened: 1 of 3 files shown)


host/sgx/sgxquote.c
```diff
@@ -395,8 +395,13 @@ oe_result_t oe_sgx_qe_get_target_info(
     else
     {
         _load_sgx_dcap_ql();
-        err = _sgx_qe_get_target_info((sgx_target_info_t*)target_info);
+        if (!_sgx_qe_get_target_info)
```
CodeMonkeyLeet, line 398: Does this check need to be here? `_load_sgx_dcap_ql()` returns false on failure, which should be checked but isn't (i.e. `if (!_load_sgx_dcap_ql())` should be equivalent above no? 

Side note: alternatively, `_load_sgx_dcap_ql()` is an `oe_once` check that should have happened already in the previous `_use_quote_ex_library()` call, so it's just getting the cache result anyway. That could probably be dropped if you want to keep the check.

yentsanglee: Will just keep it simple by verifying pointer before accessing it.

yentsanglee:
bors try

PR #3425, opened 2020-08-21
Secondary ELF memory region (for feature/sgx-lkl/support branch only)

mikbras:
Add support for a secondary ELF memory region to the **feature/sgx-lkl-support** branch.

This PR extends Open Enclave  to handle two ELF images: (1) the enclave image, and (2) a secondary ELF image. The loader and signer now accept a path with two images of the form:

```
<enclave-image>[:<secondary-image>]
```

The secondary image is loaded right after the code and data pages of the enclave image. The secondary image becomes part of the enclave address space (and is included in the MRENCLAVE measurement).


radhikaj:
Closing this for now since this feature is not actively worked on.

PR #3426, opened 2020-08-21
Support full LVI mitigation on OpenSSL based on Intel-sgx-ssl

### LINKED ISSUE #3068: SGX: LVI mitigation tool limitation when processing assembly source with .byte code
The GNU Assembler and Python script based solution to parse assembly code and insert LFENCE or replacing RET/Indirect-branch instructions were adopted in Intel SGX SDK project and [Open Enclave project ](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/AdvancedBuildInfo.md#lvi-mitigation) to mitigate the [LVI issue](https://software.intel.com/security-software-guidance/insights/deep-dive-load-value-injection).

The GNU Assembler and the Python script search for relevant instruction mnemonic, for example, instructions reading from memory, or RET instruction, and perform the required insertion or replacement. This implementation is not able to handle .byte code instructions in the assembly code. I believe the GNU Assembler and the Python script will output a warning if it encounters .byte code. 

Though it should be an uncommon practice to write assembly source code using .byte code, and compilers typically won't generate assembly output with .byte code, some assembly source code does use .byte code. One prominent example is OpenSSL project. OpenSSL contains assembly implementation for certain crypto algorithm, to leverage specific CPU instructions for security or performance reason. OpenSSL source tree contains perl scripts that the build process uses to generate assembly code. The output assembly code contains .byte code for certain instructions. My guess is that the perl scripts were developed before those instructions were widely supported by the assemblers. So the developer(s) chose to use .byte code instead of mnemonic.

Due to the limitation of the GNU Assembler and the Python script implementation, the Intel [SGXSSL project](https://github.com/intel/intel-sgx-ssl) chose to [manually insert mitigation around .byte code the tool can't handle](https://github.com/intel/intel-sgx-ssl/blob/master/openssl_source/Linux/README.md). For example, in [aesni-x86_64.s](https://github.com/intel/intel-sgx-ssl/blob/master/openssl_source/Linux/aesni-x86_64.s), at [line 1847](https://github.com/intel/intel-sgx-ssl/blob/67e01f1ecc47d89e3ef6a73b0959ba5a6e61e27c/openssl_source/Linux/aesni-x86_64.s#L1847), the .byte code is [AESENCLAST](https://www.felixcloutier.com/x86/aesenclast) instruction with memory load. A LFENCE instruction was inserted after it.

The ongoing effort of supporting OpenSSL crypto in OE SDK as described in the [design doc](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/enclave_openssl_support.md) needs to keep the LVI mitigation tool limitation in mind.

mingweishih:
This PR addresses #3068 by using the existing fixes from Intel SGX SSL repo. That is, this PR adds Intel-sgx-ssl as a submodule and integrate the necessary changes/files from the repo into the existing build process of OpenSSL.

Fixes #3068

### DIFF
--- a/3rdparty/openssl/CMakeLists.txt
+++ b/3rdparty/openssl/CMakeLists.txt
@@ -2,9 +2,11 @@
 # Licensed under the MIT License.
 
 check_submodule_not_empty(openssl)
+check_submodule_not_empty(intel-sgx-ssl)
 
 include(ExternalProject)
 
+set(INTELSSL_DIR ${CMAKE_CURRENT_SOURCE_DIR}/intel-sgx-ssl/openssl_source)
 set(OPENSSL_DIR ${CMAKE_CURRENT_SOURCE_DIR}/openssl)
 set(OPENSSL_BUILD ${CMAKE_CURRENT_BINARY_DIR}/build)
 set(OPENSSL_BUILD_INCLUDES ${OPENSSL_BUILD}/include)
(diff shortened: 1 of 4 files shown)


3rdparty/openssl/CMakeLists.txt
```diff
@@ -891,8 +983,10 @@ add_enclave_library(
   openssl/crypto/x509v3/v3_sxnet.c
   openssl/crypto/x509v3/v3_tlsf.c
   openssl/crypto/x509v3/v3_utl.c
-  openssl/crypto/x509v3/v3err.c
-  ${PLATFORM_SRC})
+  openssl/crypto/x509v3/v3err.c)
+
+add_enclave_library(opensslcrypto_asm OBJECT ${GENERATED_ASM_SRC}
```
CodeMonkeyLeet, line 988: Is there are reason for creating a separate library for this that needs to be published instead of just linking the obj files directly into the existing opensslcrypto lib?

mingweishih: I simply want to call out that these assembly files   are not the same as what come from OpenSSL (either generated with the modified peel script or copied from sgxssl repo).

CodeMonkeyLeet: I guess I'm not sure I understand how putting it into a separate library helps? It's a non-optional library if you're using OpenSSL in enclaves so it always needs to be included anyway, and it creates an additional expansion of the matrix from 2 to four new libraries for OpenSSL (2 new LVI variant libraries) 

Also one additional clarification: the 3 manually edited .s files from Intel don't actually contain the LVI mitigation, correct? They just make it possible for the mitigating assembler to identify the location that needs to be patched by removing the use of constant encodings, correct?

mingweishih: Revert the changes of adding a new library as you suggest.

Regarding your question, it's partially correct. These 3 files contain lfence for indirect branch instructions (which are still constant encoded) but not for ret instructions (handled by the modified perl script). But for the simplicity of the build flow, I use them for both normal and LVI mitigated libraries.

mingweishih:
bors try

PR #3433, opened 2020-08-24
Add check_submodule_not_empty

anakrish:
Gives better error message if mbedtles submodule has not been checked out.
Consistent with other submodules.

Also run cmake-format on the file.


anakrish:
bors try

PR #3435, opened 2020-08-24
Update quorum and voting rules to be two-thirds in Technical Charter

dthaler:
To match
https://github.com/openenclave/openenclave/blob/master/docs/Community/governance/charter.md


achamayou (approved):

CodeMonkeyLeet:
Marking as `do not merge` until CGC confirms we're staying with existing quorum rules.

dthaler:
Including myself, that makes a total of 8 approvals which is >2/3 of the CGC, enough to pass.  Since the vote has been open for a week and has enough votes to pass, I am considering this one ready to merge.

bors r+

dthaler:
Removing "do not merge" label since has enough votes to pass.  We will only take it to LF once other votes are completed so can still change if we decide we need to.

PR #3445, opened 2020-08-25
Add oe_verify_report_internal() to liboehostverify.a

### LINKED ISSUE #3312: hostverify library: Cannot use oe_verify_attestation_certificate_with_evidence
This is with current `master`:
```
../openenclave/lib/openenclave/host/liboehostverify.a(verifier.c.o): In function `_verify_report':
.../common/sgx/verifier.c:742: undefined reference to `oe_verify_report_internal'
```
Looks like some sources are missing in the archive.

shants:
fixes #3312

Signed-off-by: Shantanu Sharma <shanshar@microsoft.com>

### DIFF
--- a/host/sgx/report.c
+++ b/host/sgx/report.c
@@ -2,168 +2,13 @@
 // Licensed under the MIT License.
 
 #include <openenclave/attestation/sgx/evidence.h>
-#include <openenclave/host.h>
-#include <openenclave/host_verify.h>
-#include <openenclave/internal/calls.h>
 #include <openenclave/internal/plugin.h>
 #include <openenclave/internal/raise.h>
 #include <openenclave/internal/report.h>
-#include <openenclave/internal/safecrt.h>
-#include <openenclave/internal/safemath.h>
-#include <openenclave/internal/sgx/plugin.h>
-#include <openenclave/internal/trace.h>
-#include <openenclave/internal/utils.h>
-#include "../common/sgx/quote.h"
-#include "core_u.h"
+
 #include "platform_u.h"
-#include "quote.h"
-
-#include "sgxquoteprovider.h"
-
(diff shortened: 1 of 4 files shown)


qiucwang:
bors try

host/sgx/report.c
```diff
@@ -130,6 +133,7 @@ oe_result_t oe_get_report_v2(
     if (!enclave || !report_buffer || !report_buffer_size)
         OE_RAISE(OE_INVALID_PARAMETER);
 
+#ifndef OE_BUILD_HOST_VERIFY
```
yentsanglee, line 136: Host_verify package never call oe_get_report*(). We can remove the entire function if possible.

host/sgx/report.c
```diff
@@ -249,13 +262,13 @@ oe_result_t oe_verify_report_internal(
     }
     else if (header->report_type == OE_REPORT_TYPE_SGX_LOCAL)
     {
-        oe_result_t retval;
-
         if (enclave == NULL)
             OE_RAISE(OE_INVALID_PARAMETER);
-
+#ifndef OE_BUILD_HOST_VERIFY
```
yentsanglee, line 267: Passing report_type OE_REPORT_TYPE_SGX_LOCAL to host_verify package should be treated as a error. We can exclude this `else if` block and let `else` handle it.

host/sgx/report.c
shants: host/sgx/report.c now has oe_get_report_v2(), moved rest to  host/sgx/report_common.c.

host/sgx/report.c
shants: Done.

yentsanglee:
bors try

PR #3452, opened 2020-08-27
Determine the certificate size for different reports dynamically

### LINKED ISSUE #3378: `OE_MAX_CERT_SIZE` is too small for EEID report
`OE_MAX_CERT_SIZE` is 8192, which should be enough for most cases. But with EEID plugin, we are creating a huge data blob that exceeds the limit. As my debugger shows, the data blob is 18671 bytes for an sample sgx-lkl application. We need to increase the size limit.

ryanhsu19:
Replace `OE_MAX_CERT_SIZE` with a combination of `OE_MIN_CERT_SIZE` and other inputs: `oid_size`, `public_key_buf_size` and `remote_report_buf_size`. A multiplier of one and a quarter is included to create room for paddings.

The fixed value of `OE_MIN_CERT_SIZE` (=256 bytes) is larger than the summation of sizes of the subject name (~64 bytes), issuer name (~64 bytes), first valid date (~16 bytes) and last valid date (~16 bytes) of a certificate.

In the `attestation_cert_api` `ctest`, 5KB of data are written into a 6KB-sized buffer in both test cases.

Fix #3378.

### DIFF
--- a/enclave/tls_cert.c
+++ b/enclave/tls_cert.c
@@ -48,6 +48,7 @@ static oe_result_t generate_x509_self_signed_certificate(
     size_t bytes_written = 0;
     uint8_t* cert_buf = NULL;
     oe_cert_config_t config = {0};
+    size_t oe_cert_size = 0;
 
     config.private_key_buf = private_key_buf;
     config.private_key_buf_size = private_key_buf_size;
@@ -64,13 +65,16 @@ static oe_result_t generate_x509_self_signed_certificate(
     config.ext_oid = (char*)oid;
     config.ext_oid_size = oid_size;
 
-    // allocate memory for cert output buffer
-    cert_buf = (uint8_t*)oe_malloc(OE_MAX_CERT_SIZE);
+    // allocate memory for cert output buffer and leave room for paddings
+    oe_cert_size =
+        remote_report_buf_size + public_key_buf_size + OE_MIN_CERT_SIZE;
+    cert_buf = (uint8_t*)oe_malloc(oe_cert_size);
     if (cert_buf == NULL)
         goto done;
 
     result = oe_gen_custom_x509_cert(
-        &config, cert_buf, OE_MAX_CERT_SIZE, &bytes_written);
+        &config, cert_buf, oe_cert_size, &bytes_written);
+
     OE_CHECK_MSG(
         result,
         "oe_gen_custom_x509_cert failed with %s",
--- a/include/openenclave/internal/crypto/cert.h
+++ b/include/openenclave/internal/crypto/cert.h
@@ -377,7 +377,9 @@ typedef struct _oe_cert_config
     size_t ext_oid_size;
 } oe_cert_config_t;
 
-#define OE_MAX_CERT_SIZE 8192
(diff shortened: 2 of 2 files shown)


enclave/tls_cert.c
```diff
@@ -64,13 +65,17 @@ static oe_result_t generate_x509_self_signed_certificate(
     config.ext_oid = (char*)oid;
     config.ext_oid_size = oid_size;
 
-    // allocate memory for cert output buffer
-    cert_buf = (uint8_t*)oe_malloc(OE_MAX_CERT_SIZE);
+    // allocate memory for cert output buffer and leave room for paddings
+    oe_cert_size =
```
mingweishih, line 69: Do we know where does the max value come from? If there is a reasonable max value, should we add a check against the size of certificate then?

include/openenclave/internal/crypto/cert.h
```diff
@@ -377,7 +377,8 @@ typedef struct _oe_cert_config
     size_t ext_oid_size;
 } oe_cert_config_t;
 
-#define OE_MAX_CERT_SIZE 8192
+/* includes subject name, issuer name and valid date range */
+#define OE_MIN_CERT_SIZE 256
```
mingweishih, line 381: Is 256 always sufficient to contain those fields? Would there be a corner case?

ryanhsu19: Although it is large enough for the two examples I've got, I'll discuss it with others in the attestation team.

enclave/tls_cert.c
yentsanglee: I think the max value were from previous estimation. The generated certificates are usually 5K - 7K. @ryanhsu19 what is a typical `oe_cert_size` value? If it's less than 8K, we can bump up OE_MIN_CERT_SIZE. It has to cover version, serial number, signature algorithm, validity, signature, etc.

yentsanglee: Padding with 1/4 of (oid_size + public_key_buf_size + remote_report_buf_size) means the padding a variable. For EEID, `remote_report_buf_size` can be 16K which means padding 4K. That seems too big.

ryanhsu19: IMO `remote_report_buf_size` is the most important factor, while others are near constant, so I reset `OE_MIN_CERT_SIZE` to `1024` and set `oe_cert_size` to `OE_MIN_CERT_SIZE + remote_report_buf_size*5/4`.

Both test cases in the `attestation_cert_api` test have a size of 5.1K and 5.5K, respectively, and the buffer size in my proposed version is around 6.8K in both cases.

A 4K padding for a 16K data is indeed large...let me try some other values.

ryanhsu19: After considering both `remote_report_buf_size` and `public_key_buf_size` as major dynamic factors and doing different experiments with the value of `OE_MIN_CERT_SIZE` being 512, 1024 and 2048, I have decided to set `OE_MIN_CERT_SIZE` to 2048. Since there is enough room for padding, I have removed the weight on `remote_report_buf_size`.

yentsanglee:
bors retry

PR #3453, opened 2020-08-28
Updated attested_tls with attestation plugin APIs.

qiucwang:
Updated attested_tls sample with attestation plugin APIs: 
`oe_get_attestation_certificate_with_evidence` and `oe_verify_attestation_certificate_with_evidence()`

The README file will be updated by @shants in another PR along with other sample updates.


qiucwang:
bors try

qiucwang:
bors try

samples/attested_tls/server/enc/cert_verifier.cpp
```diff
@@ -59,13 +60,14 @@ int cert_verify_callback(
     if (cert_size <= 0)
         goto exit;
 
-    result = oe_verify_attestation_certificate(
-        cert_buf, cert_size, enclave_identity_verifier_callback, NULL);
+    oe_verifier_initialize();
```
yentsanglee, line 63: Can we move oe_verifier_initialize() to somewhere else so it only gets called once?

samples/attested_tls/client/enc/cert_verifier.cpp
```diff
@@ -48,18 +48,20 @@ int cert_verify_callback(
     if (cert_size <= 0)
         goto exit;
 
-    result = oe_verify_attestation_certificate(
-        cert_buf, cert_size, enclave_identity_verifier_callback, NULL);
+    oe_verifier_initialize();
```
yentsanglee, line 51: Move to somewhere else so it only gets called once.

samples/attested_tls/server/enc/cert_verifier.cpp
qiucwang: Sure. Thanks for point that out.
Updated.

samples/attested_tls/client/enc/cert_verifier.cpp
qiucwang: Updated

samples/attested_tls/common/utility.cpp
```diff
@@ -163,6 +171,23 @@ bool verify_mrsigner(
...
+    size_t claims_size,
+    const char* name)
+{
+    for (size_t i = 0; i < claims_size; i++)
+    {
+        if (strcmp(claims[i].name, name) == 0)
+            return &(claims[i]);
+    }
+    return nullptr;
+}
```
mingweishih, line 193:
```suggestion
}

```

samples/attested_tls/non_enc_client/verify_callback.cpp
```diff
@@ -22,13 +23,31 @@
...
+    const oe_claim_t* claims,
+    size_t claims_size,
+    const char* name)
+{
+    for (size_t i = 0; i < claims_size; i++)
+    {
+        if (strcmp(claims[i].name, name) == 0)
+            return &(claims[i]);
+    }
+    return NULL;
```
mingweishih, line 41:
```suggestion
    return nullptr;
```

samples/attested_tls/client/enc/client.cpp
```diff
@@ -219,6 +219,7 @@ int launch_tls_client(char* server_name, char* server_port)
     mbedtls_ctr_drbg_init(&ctr_drbg);
     mbedtls_x509_crt_init(&client_cert);
     mbedtls_pk_init(&pkey);
+    oe_verifier_initialize();
```
mingweishih, line 222: Just want to call out that now the `oe_verifier_initialize` can still be implicitly called (e.g., when using `oe_verify_report`). Also, the `oe_verifier_shutdown` is always called at the enclave termination. These hold true for the `oe_attester_initialize`/`oe_attester_shutdown` as well. I'm wondering if we need to remove those implicit behaviors.

cc @dthaler @yentsanglee @CodeMonkeyLeet

samples/attested_tls/non_enc_client/verify_callback.cpp
qiucwang: `-std=c++11` flag was not set in non-enc Makefile so `nullptr` here raised the "error: ‘nullptr’ was not declared in this scope" for ACC1604 gcc Debug/Release LVI. So I kept it as `NULL` here.

samples/attested_tls/client/enc/client.cpp
dthaler: My opinion is that it is more dev friendly if the app does not need to explicitly initialize/shutdown the attester and so I would rather delete this line of code.

samples/attested_tls/common/utility.cpp
```diff
@@ -86,10 +92,12 @@ oe_result_t generate_certificate_and_pkey(
         goto exit;
     }
 
-    printf("public key used:\n[%s]", public_key_buf);
+    printf("public key used:\n%s\n", public_key_buf);
 
     // both ec key such ASYMMETRIC_KEY_EC_SECP256P1 or RSA key work
-    result = oe_generate_attestation_certificate(
+    oe_attester_initialize();
```
dthaler, line 98: See discussion in other file about this not being needed.

samples/attested_tls/non_enc_client/verify_callback.cpp
dthaler: Should look into why that failed and fix it.  Ok to do so in a separate PR.

samples/attested_tls/non_enc_client/verify_callback.cpp
```diff
@@ -55,71 +74,127 @@ bool verify_mrsigner(
...
     return true;
 }
 
-// This is the identity validation callback. An TLS connecting party (client or
-// server) can verify the passed in "identity" information to decide whether to
-// accept the connection reqest from an tls server running inside a specific
-// enclave In a real app, custom identity validation should be done inside this
-// routine
-oe_result_t enclave_identity_verifier(oe_identity_t* identity, void* arg)
+// This is the evidence claims validation callback. An TLS connecting party
```
dthaler, line 81:
```suggestion
// This is the evidence claims validation callback. A TLS connecting party
```

samples/attested_tls/non_enc_client/verify_callback.cpp
```diff
@@ -55,71 +74,127 @@ bool verify_mrsigner(
...
+// (client or server) can verify the passed in "identity" information to decide
+// whether to accept the connection reqest from an tls server running inside a
```
dthaler, line 83:
```suggestion
// whether to accept the connection request from a TLS server running inside a
```

samples/attested_tls/non_enc_client/verify_callback.cpp
```diff
@@ -55,71 +74,127 @@ bool verify_mrsigner(
...
+// specific enclave In a real app, custom identity validation should be done
+// inside this routine
```
dthaler, line 85:
```suggestion
// specific enclave. In a real app, custom identity validation should be done
// inside this routine.
```

samples/attested_tls/non_enc_client/verify_callback.cpp
```diff
@@ -199,8 +274,9 @@ int verify_callback(int preverify_ok, X509_STORE_CTX* ctx)
 #endif
 
     // verify tls certificate
-    result = oe_verify_attestation_certificate(
-        der, der_len, enclave_identity_verifier, NULL);
+    oe_verifier_initialize();
```
dthaler, line 277: See discussion in other files about this not being needed

samples/attested_tls/server/enc/server.cpp
```diff
@@ -316,6 +316,7 @@ int setup_tls_server(char* server_port)
     mbedtls_pk_init(&pkey);
     mbedtls_entropy_init(&entropy);
     mbedtls_ctr_drbg_init(&ctr_drbg);
+    oe_verifier_initialize();
```
dthaler, line 319: See discussion in other file about this not being needed

samples/attested_tls/client/enc/client.cpp
yentsanglee: Even though the initialize and the shutdown are implicitly called, I think it's a good practice to have them in the samples. Until we have an overall plugin design, external plugins have to provide similar initialization.

samples/attested_tls/non_enc_client/verify_callback.cpp
qiucwang: Thanks!

samples/attested_tls/non_enc_client/verify_callback.cpp
qiucwang: Thanks!

qiucwang:
bors try

samples/attested_tls/non_enc_client/verify_callback.cpp
qiucwang: > Should look into why that failed and fix it. Ok to do so in a separate PR.

I added -std=c++11 flag and now `nullptr` passed the bors try.

samples/attested_tls/common/utility.cpp
qiucwang: Thanks!

PR #3460, opened 2020-08-31
Fix for loop bugs in attesttls sample.

qiucwang:
PR#3453 introduced bugs in for loop in attesttls sample:
```
server\enc\identity_verifier.cpp:55:23: error: comparison of integers of different signs: 'int' and 'const size_t' (aka 'const unsigned long') [-Werror,-Wsign-compare]
    for (int i = 0; i < claim->value_size; i++)
```
This PR fixed those bugs.


qiucwang:
bors try

PR #3478, opened 2020-09-02
libsgx_dcap_quoteverify.so requirement

### LINKED ISSUE #3477: New DCAP release broke CI ansible scripts
With Intel's new DCAP release a change was noticed in the packaging and 
`/usr/lib/x86_64-linux-gnu/libsgx_dcap_quoteverify.so` is no longer installed.

This causes issues with the CI scripts as part of the validation checks we check the file paths. This package needs to be added to the scripts as a quick fix, the long term fix being to version the PSW/DCAP libraries to ensure this does not happen again.

CyanDevs:

Fixes #3477

### DIFF
--- a/scripts/ansible/roles/linux/intel/vars/ubuntu/main.yml
+++ b/scripts/ansible/roles/linux/intel/vars/ubuntu/main.yml
@@ -21,4 +21,3 @@ packages_validation_distribution_files:
   - "/usr/lib/x86_64-linux-gnu/libsgx_pce_logic.so"
   - "/usr/lib/x86_64-linux-gnu/libsgx_qe3_logic.so"
   - "/usr/lib/x86_64-linux-gnu/libsgx_dcap_ql.so"
-  - "/usr/lib/x86_64-linux-gnu/libsgx_dcap_quoteverify.so"


CyanDevs:
Fixes #3477

BRMcLaren:
bors try

scripts/ansible/roles/linux/intel/vars/ubuntu/main.yml
```diff
@@ -15,6 +15,7 @@ intel_dcap_packages:
   - "libsgx-dcap-ql"
   - "libsgx-dcap-ql-dev"
   - "libsgx-urts"
+  - "libsgx-dcap-quote-verify-dev"
```
radhikaj, line 18: Do we need libsgx-dcap-quote-verify lib?

radhikaj (commented):
Thanks @CyanDevs

scripts/ansible/roles/linux/intel/vars/ubuntu/main.yml
CyanDevs: These are the files provided by ligbsgx-dcap-quote-verify, though it is not apparent whether we need any of them. It does not seem like it since [libsgx_dcap_quoteverify.so.1 is mentioned anywhere](https://github.com/openenclave/openenclave/search?q=%22libsgx_dcap_quoteverify.so.1%22&unscoped_q=%22libsgx_dcap_quoteverify.so.1%22).

```
$ dpkg -c libsgx-dcap-quote-verify_1.8.100.2-bionic1_amd64.deb 
drwxr-xr-x root/root         0 2020-08-17 03:59 ./
drwxr-xr-x root/root         0 2020-08-17 03:59 ./usr/
drwxr-xr-x root/root         0 2020-08-17 03:59 ./usr/lib/
drwxr-xr-x root/root         0 2020-08-17 03:59 ./usr/lib/x86_64-linux-gnu/
-rw-r--r-- root/root   3020264 2020-08-17 03:59 ./usr/lib/x86_64-linux-gnu/libsgx_dcap_quoteverify.so.1.8.100.2
drwxr-xr-x root/root         0 2020-08-17 03:59 ./usr/share/
drwxr-xr-x root/root         0 2020-08-17 03:59 ./usr/share/doc/
drwxr-xr-x root/root         0 2020-08-17 03:59 ./usr/share/doc/libsgx-dcap-quote-verify/
-rw-r--r-- root/root       193 2020-08-17 03:59 ./usr/share/doc/libsgx-dcap-quote-verify/changelog.Debian.gz
-rw-r--r-- root/root      3800 2020-08-17 03:59 ./usr/share/doc/libsgx-dcap-quote-verify/copyright
lrwxrwxrwx root/root         0 2020-08-17 03:59 ./usr/lib/x86_64-linux-gnu/libsgx_dcap_quoteverify.so.1 -> libsgx_dcap_quoteverify.so.1.8.100.2
```

radhikaj: These are dynamically loaded dependencies. @CodeMonkeyLeet  can you chime in?

CodeMonkeyLeet: I would suggest pulling down libsgx-dcap-quote-verify as well; we don't explicitly use it right now, but attestation will be using it very soon, see #3414. 

@hyjiang will be integrating the DCAP quote verification into the OE SDK and he is also the owner of the both libsgx-dcap-quote-verify and its -dev package, I believe, so I'm tagging him in case he wants to chime in here.

yentsanglee: As Simon mentioned, we don't need libsgx-dcap-quote-verify until implementation for PR #3414 is in-place.

@hyjiang what is the quote verification package in Windows?

hyjiang: Yes, libsgx-dcap-quote-verify is needed for PR #3414 .
Quote verification libraries/head files are part of DCAP QL package before, we separate it into a standalone installer in latest release, the reason is that it allows user to only install quote verification libraries/header files on non-SGX system. 

But note that libsgx-dcap-quote-verify/dev only include one library and two header files.

CyanDevs: Since that package will be needed in the future, is the preference to add the libsgx-dcap-quote-verify in advance as part of this PR, as part of PR #3414, or sometime after that PR #3414 is merged?

BRMcLaren: The preference would be to handle it pre-emptively, to avoid the chicken and egg problem in merging #3414 .

CodeMonkeyLeet: I'll scope it this way:
* Is this PR intended to be in v0.11?
   * If so, keep in mind that the implementation of #3414 is not, and I would prefer not to introduce new dependency that users of the SDK have to deal with until it's used. 
    * CI/CD can still pre-emptively pull new packages down, but we just need to be careful that it does not inadvertently introduce a requirement for the end user ahead of time. (e.g. through shared scripts like install-windows-prereqs.ps1, although I don't think the bundling has changed on the Windows side)
   * Once v0.11 is locked in, master is fair game for this and #3414 and we probably want it to be available of folks having to implement the code change.
      * I would also suggest updating the user docs at the same time to reflect the new dependency.

anakrish: @CyanDevs @yentsanglee @hyjiang

OE SDK dynamically loads the necessary libraries as opposed to statically linking against them.
Why do we need the -dev package? Is it for prototypes of functions that are going to be looked up via dlsym?

BRMcLaren: The OE-CI needs the library in the path to be able to rebuild itself. 

If the package is not required in the latest dcap split, we need to delete the validation. 
If it is required add the package.

We are currently blocked on this, if we need any additional packages please file an issue and we can add them later.

hyjiang: Just want to highlight that libsgx-dcap-quote-verify is a dependency of libsgx-dcap-ql right now, it means if you want to install DCAP QL, it will try to install DCAP Quote Verify automatically. 
So there is no difference whether you add libsgx-dcap-quote-verify here.

Maybe we will change the dependency to `recommend` in future.

CyanDevs: @anakrish and @BRMcLaren 

I tested without installing libsgx-dcap-quote-verify-dev and without checking for /usr/lib/x86_64-linux-gnu/libsgx_dcap_quoteverify.so by removing lines 18 and 25. All tests passed.
```
100% tests passed, 0 tests failed out of 281

Total Test time (real) = 1838.88 sec
```
Assuming tests cover the executions as well, I suspect that we should be removing the package and library instead of adding it. Thoughts?

BRMcLaren: Sorry, could you clarify?

Was the total delta from the state of master to your current locally working fix just removing the validation?

CyanDevs: Yes, that's right. Here's the quick diff of the tested Ansible role vars for Ubuntu (based off of current commit):

```
@@ -15,11 +15,9 @@ intel_dcap_packages:
   - "libsgx-dcap-ql"
   - "libsgx-dcap-ql-dev"
   - "libsgx-urts"
-  - "libsgx-dcap-quote-verify-dev"
 
 packages_validation_distribution_files:
   - "/usr/lib/x86_64-linux-gnu/libsgx_enclave_common.so"
   - "/usr/lib/x86_64-linux-gnu/libsgx_pce_logic.so"
   - "/usr/lib/x86_64-linux-gnu/libsgx_qe3_logic.so"
   - "/usr/lib/x86_64-linux-gnu/libsgx_dcap_ql.so"
-  - "/usr/lib/x86_64-linux-gnu/libsgx_dcap_quoteverify.so"

```

Is it preferred to amend my last commit, or make a new commit to preserve commit history?

CyanDevs: New changes are pushed now.

BRMcLaren (approved):

rs-- (approved):

scripts/ansible/roles/linux/intel/vars/ubuntu/main.yml
anakrish: @CyanDevs
libsgx_dcap_quote_verify.so is a link to libsgx_dcap_quote_verify.so.1. It would be good to add a validation for libsgx_dcap_quote_verify.so.1.

E.g: libsgx_enclave_common.so:
```
lrwxrwxrwx 1 root root      26 Aug 16 20:56 /usr/lib/x86_64-linux-gnu/libsgx_enclave_common.so -> libsgx_enclave_common.so.1
lrwxrwxrwx 1 root root      34 Aug 16 20:56 /usr/lib/x86_64-linux-gnu/libsgx_enclave_common.so.1 -> libsgx_enclave_common.so.1.0.112.2
```
What do you see when you run `ls /usr/lib/x86_64-linux-gnu/libsgx* -la` on the machine?

We want to be sure that for running enclaves, the .so.1 versions are available.

PR #3482, opened 2020-09-03
Add mallinfo.h, implement it for dlmalloc, and convert testcases

achamayou:
Implements #3410, in order to solve #3262.

The implementation for snmalloc will follow in a separate PR, since it requires an update of the submodule.

PR #3410, opened 2020-08-18
Draft for oe_allocator_mallinfo proposal

achamayou:
Design proposal to solve #3262 by adding a new allocator interface, for internal testcases only.

PR #3484, opened 2020-09-03
Add mallinfo.h, implement it for dlmalloc, snmalloc, and convert testcases

achamayou:
Implements design document #3410, in order to solve #3262. Contains a minor amendment to the original design document, as proposed by @dthaler, to use `oe_result_t`.

Originally a follow up to #3482, which was the dlmalloc-only implementation, but now combined into a single PR implementing the interface for both dlmalloc and snmalloc, and updating the testcases to use it.


achamayou:
bors try

3rdparty/snmalloc/allocator.cpp
```diff
@@ -49,3 +58,14 @@ void oe_allocator_thread_cleanup(void)
     current_alloc_pool()->release(ThreadAlloc::get());
     allocator_local = nullptr;
 }
+
+int oe_allocator_mallinfo(oe_mallinfo_t* info)
+{
+    info->max_total_heap_size = _heap_end - _heap_start;
```
anakrish, line 64: Can _heap_end ever decrease under both snmalloc and dlmalloc?

anakrish (approved):
LGTM. Thanks for doing this!

include/openenclave/advanced/mallinfo.h
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Obtain current memory usage statistics.
+ *
+ * Users may make this call on allocators that support it, to find out how much
+ * memory can be allocated in total, how much is allocated at the moment, and
+ * the high watermark of allocation so far.
+ *
+ * @param[out] info An oe_mallinfo_t struct, to be populated by the allocator.
+ *
+ * @returns 0 on success.
```
dthaler, line 45: What does it return if the allocator doesn't support it?

tests/memory/enc/fragment.c
```diff
@@ -16,27 +17,12 @@
...
-};
-
-struct mallinfo dlmallinfo(void);
-
 static size_t _get_heap_size()
 {
-    // call malloc to update the global data.
-    free(malloc(1));
-    return dlmallinfo().keepcost;
+    static oe_mallinfo_t info;
```
dthaler, line 22: why does this need to be static?

3rdparty/snmalloc/allocator.cpp
achamayou: That's a good question, and I do not know what the future-proof answer is. Currently though, the allocators are given a fixed start and end via oe_allocator_init(), and there's no API/provision to change it, so the answer is no.

include/openenclave/advanced/mallinfo.h
achamayou: The design spec specifies non-zero, I can update this description. My intention was to leave it deliberately unspecified, and allow allocators to potentially define their own error codes if necessary.

In practice though, I believe the only realistic error value would be OE_NOT_IMPLEMENTED or something equivalent.

I've also wondered if it should be an oe_result_t, but that seemed overloaded enough already. If someone has a stronger opinion about this, I am happy to make the necessary adjustments.

tests/memory/enc/fragment.c
achamayou: It doesn't, I'll fix it.

3rdparty/snmalloc/allocator.cpp
anakrish: Sorry, I intended to ask if **_heap_start** would ever decrease.
My understanding is that both the allocators increment _heap_start to fetch memory from the heap. I don't know if either allocator ever gives back memory to the heap.
Speaking of which, `_heap_end - _heap_start` would give the amount of free memory no?

mjp41: `_heap_start` is not used by `snmalloc` the call to `setup_initial_range` modifies state inside the PAL to account for the range that is not visible to OE.

3rdparty/snmalloc/allocator.cpp
```diff
@@ -27,10 +28,18 @@ void* ThreadAllocUntyped::get()
 
 // Enable the Open Enclave PAL(Platform Abstraction Layer) for Open Enclave,
 // see pal_open_enclave.h for details
+#include "./snmalloc/src/override/malloc-extensions.cc"
 #include "./snmalloc/src/override/malloc.cc"
 
+static uint8_t* _heap_start;
```
mjp41, line 34: Would it make more sense to store `max_total_heap_size`?  The two variables are leading to confusion for their purpose.

3rdparty/snmalloc/allocator.cpp
achamayou: @anakrish I think you're thinking of `_heap_next` for dlmalloc, `_heap_start` is never modified there either. I'll change this to be captured as `total_heap_size` in `oe_allocator_init()` for both allocators, which will hopefully be clearer.

3rdparty/snmalloc/allocator.cpp
achamayou: Yes, agreed, I will make this change to both allocators.

3rdparty/snmalloc/allocator.cpp
achamayou: Did that on both PRs. The dlmalloc code in this one isn't updated, but will be once it's rebased on the other one of course.

anakrish: Yes,  `_heap_next` was what I was thinking of.

include/openenclave/advanced/mallinfo.h
dthaler: Yes I think it should be oe_result_t rather than a new error code space.

achamayou:
bors try

include/openenclave/advanced/mallinfo.h
achamayou: Ok, I've made the change on both PRs.

docs/DesignDocs/Mallinfo.md
```diff
@@ -48,9 +48,9 @@ struct oe_mallinfo_t {
...
 }
 
-int oe_allocator_mallinfo(struct oe_mallinfo_t * info);
+oe_result_t oe_allocator_mallinfo(struct oe_mallinfo_t * info);
 ```
 
 The allocator will set `max_total_heap_size` to the maximum number of bytes it can allocate in total, and `current_allocated_heap_size` to the number of bytes allocated at the moment. `peak_allocated_heap_size` will contain the highest value reached by `current_allocated_heap_size` during execution.
 
-Successful calls return 0, failures non-zero.
+Successful calls return OE_OK. The allocator may return OE_UNSUPPORTED if it not support the interface, or OE_FAILURE for other failures.
```
dthaler, line 56:
```suggestion
Successful calls return OE_OK. The allocator may return OE_UNSUPPORTED if it does not support the interface, or OE_FAILURE for other failures.
```

include/openenclave/advanced/mallinfo.h
```diff
@@ -0,0 +1,52 @@
...
+ * Obtain current memory usage statistics.
+ *
+ * Users may make this call on allocators that support it, to find out how much
+ * memory can be allocated in total, how much is allocated at the moment, and
+ * the high watermark of allocation so far.
+ *
+ * @param[out] info An oe_mallinfo_t struct, to be populated by the allocator.
+ *
+ * @returns OE_OK on success. OE_UNSUPPORTED if the allocator does not support
+ * the interface, or OE_FAILURE for other failures.
```
dthaler, line 46: Please use @retval instead of @returns so you get a nicely formatted table like is shown at https://openenclave.io/apidocs/v0.10/enclave_8h_a3401b750937f12b6b9f3911b5a39304f.html#a3401b750937f12b6b9f3911b5a39304f for example.   Also note that the text after the value itself should be phrased as a complete sentence in that style.

tests/bigmalloc/enc/enc.c
```diff
@@ -14,23 +15,19 @@ oe_result_t test_malloc()
...
     {
-        const uint8_t* base = (const uint8_t*)__oe_get_heap_base();
-        const uint8_t* brk = (const uint8_t*)dlmalloc_sbrk(0);
-        const uint8_t* end = (const uint8_t*)__oe_get_heap_end();
-
-        /* Sanity checks */
-        if (!(base <= brk && brk < end))
+        oe_mallinfo_t info;
+        oe_result_t rc = oe_allocator_mallinfo(&info);
+        if (rc != OE_OK)
```
dthaler, line 24: I'd recommend changing line 26 to return rc instead of OE_FAILURE so as to preserve the more granular error

include/openenclave/advanced/mallinfo.h
dthaler, line 46: Please use `@retval` instead of `@returns` for error code values so you get a nicely formatted table like is shown at https://openenclave.io/apidocs/v0.10/enclave_8h_a3401b750937f12b6b9f3911b5a39304f.html#a3401b750937f12b6b9f3911b5a39304f for example.   Also note that the text after the value itself should be phrased as a complete sentence in that style.

dthaler, line 46: Please use @retval instead of @returns so you get a nicely formatted table like is shown at https://openenclave.io/apidocs/v0.10/enclave_8h_a3401b750937f12b6b9f3911b5a39304f.html#a3401b750937f12b6b9f3911b5a39304f for example.   Also note that the text after the value itself should be phrased as a complete sentence in that style.

achamayou:
bors try

docs/DesignDocs/Mallinfo.md
achamayou: Done

include/openenclave/advanced/mallinfo.h
achamayou: Done

tests/bigmalloc/enc/enc.c
achamayou: Done

include/openenclave/advanced/mallinfo.h
achamayou: Done

achamayou: Done

achamayou:
bors try

achamayou:
@dthaler I believe I have addressed all feedback, but if I have missed anything, please let me know.

dthaler (approved):
Looks good now, thanks.

achamayou:
bors try

PR #3489, opened 2020-09-03
EXPERIMENTAL: WIP: Figure out elf information by parsing program headers.

anakrish:
Program headers are loaded as part of enclave image.
The enclave can iterate thorugh the program headers
- And find the header with PT_TLS to figure out
 tdata_rva, tdata_size, tdata_align, tbss_size and tbss_align.
 Note: Even though tdata_align and tbss_align can be different,
 their maximum is chosen and put in the program header.
 The maximum value is enough to align both .tdata and .tbss.

 The p_filesz gives tdata_size where as p_memsz - p_filesz gives
 the tbss size.

- Find out the ending offset of the loaded sections.
  This value rounded to OE_PAGE_SIZE gives the reloc_rva.
  Iterating through the contents of the relocations until
  an entry with zero r_offset, gives the number of actual
  relocations.

The enclave base address is computed by defining a variable

uint8_t* _base_address = (uint8_t*)&_base_address.

For such a variable, before relocations, its value will be equal
to its rva. Subtracting its value (rva) from its address gives
enclave base.

Assertions are added to make sure that the loader populated
values of correspoing symbols match the values we computed above.

### DIFF
--- a/enclave/core/CMakeLists.txt
+++ b/enclave/core/CMakeLists.txt
@@ -61,6 +61,7 @@ if (OE_SGX)
     sgx/backtrace.c
     sgx/calls.c
     sgx/cpuid.c
+    sgx/elf.c
     sgx/enter.S
     sgx/entropy.c
     sgx/errno.c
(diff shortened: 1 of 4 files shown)


anakrish:
bors try

anakrish:
bors try

radhikaj:
Closing this now as alternate approaches are being considered.

PR #3492, opened 2020-09-04
Disable unsafe OpenSSL APIs/Macros

mingweishih:
This PR disables a set of OpenSSL APIs/macros that are considered as unsafe based on OE's threat model. More specifically, those APIs allow users to configure an OpenSSL application to read certificates from the host filesystem, which is not trusted, and therefore are recommended not to use inside an enclave.

### DIFF
--- a/3rdparty/openssl/CMakeLists.txt
+++ b/3rdparty/openssl/CMakeLists.txt
@@ -10,6 +10,7 @@ set(INTELSSL_DIR ${CMAKE_CURRENT_SOURCE_DIR}/intel-sgx-ssl/openssl_source)
 set(OPENSSL_DIR ${CMAKE_CURRENT_SOURCE_DIR}/openssl)
 set(OPENSSL_BUILD ${CMAKE_CURRENT_BINARY_DIR}/build)
 set(OPENSSL_BUILD_INCLUDES ${OPENSSL_BUILD}/include)
+set(OPENSSL_INSTALL_INCLUDES ${OE_INCDIR}/openenclave/openssl)
 # CMake-generated path.
 set(OPENSSL_GENERATED_BUILD
     ${CMAKE_CURRENT_BINARY_DIR}/openssl_generated-prefix/src/openssl_generated-build
@@ -17,6 +18,10 @@ set(OPENSSL_GENERATED_BUILD
 
 set(OPENSSL_C_COMPILER ${CMAKE_C_COMPILER})
 set(OPENSSL_CXX_COMPILER ${CMAKE_CXX_COMPILER})
+
+set(OPENSSL_APPEND_UNSUPPORTED
+    "${CMAKE_CURRENT_SOURCE_DIR}/append-unsupported ${OPENSSL_INSTALL_INCLUDES}"
+)
 
 if (OE_SGX)
   set(ARCH "linux-x86_64")
@@ -77,6 +82,27 @@ ExternalProject_Add(
           ${OPENSSL_BUILD_INCLUDES}/crypto/dso_conf.h
   COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/opensslconf.h
           ${OPENSSL_BUILD_INCLUDES}/openssl/opensslconf.h
+  BUILD_COMMAND ""
+  INSTALL_COMMAND "")
+
+# The following commmand prepares the headers to be published. The headers
+# consist of those in the openssl/include/openssl folder, opensslconf.h,
+# and unsupported.h. The unsupported.h header file includes a list of APIs that are
+# disabled by OE for security concerns. A few headers are patched (using the
+# append-unsupported script) to include unsupported.h so that OE can enforce
+# compile-time errors when any of those unsafe APIs is used.
+ExternalProject_Add(
+  openssl_includes
+  SOURCE_DIR ${OPENSSL_DIR}
+  CONFIGURE_COMMAND
+  COMMAND ${CMAKE_COMMAND} -E copy_directory ${OPENSSL_DIR}/include/openssl
+          ${OPENSSL_INSTALL_INCLUDES}
+  COMMAND
+    ${CMAKE_COMMAND} -E copy ${OPENSSL_BUILD_INCLUDES}/openssl/opensslconf.h
+    ${OPENSSL_INSTALL_INCLUDES}/opensslconf.h
+  COMMAND ${OE_BASH} -c "${OPENSSL_APPEND_UNSUPPORTED}"
+  COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/unsupported.h
+          ${OPENSSL_INSTALL_INCLUDES}/unsupported.h
   BUILD_COMMAND ""
   INSTALL_COMMAND "")
 
@@ -250,6 +276,7 @@ ExternalProject_Add(
   BUILD_BYPRODUCTS ${PATCHED_ASM_SRC}
   INSTALL_COMMAND "")
 
+add_dependencies(openssl_includes openssl_configure)
 add_dependencies(openssl_generated openssl_configure)
 add_dependencies(openssl_generated openssl_copied)
 
@@ -868,12 +895,12 @@ add_enclave_library(
   openssl/ssl/tls_srp.c)
 
 enclave_include_directories(
-  opensslcrypto PUBLIC $<BUILD_INTERFACE:${OPENSSL_BUILD_INCLUDES}>
+  opensslcrypto PRIVATE $<BUILD_INTERFACE:${OPENSSL_BUILD_INCLUDES}>
   $<BUILD_INTERFACE:${OPENSSL_DIR}> $<BUILD_INTERFACE:${OPENSSL_DIR}/include>
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/openenclave/3rdparty>)
 
 enclave_include_directories(
-  opensslssl PUBLIC $<BUILD_INTERFACE:${OPENSSL_BUILD_INCLUDES}>
+  opensslssl PRIVATE $<BUILD_INTERFACE:${OPENSSL_BUILD_INCLUDES}>
   $<BUILD_INTERFACE:${OPENSSL_DIR}> $<BUILD_INTERFACE:${OPENSSL_DIR}/include>
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/openenclave/3rdparty>)
 
@@ -1034,13 +1061,11 @@ install_enclaves(
   DESTINATION
   ${CMAKE_INSTALL_LIBDIR}/openenclave/enclave)
 install(
-  DIRECTORY ${OPENSSL_DIR}/include/openssl
+  DIRECTORY ${OPENSSL_INSTALL_INCLUDES}
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/openenclave/3rdparty
   PATTERN "__DECC_INCLUDE_EPILOGUE.H" EXCLUDE
   PATTERN "__DECC_INCLUDE_PROLOGUE.H" EXCLUDE
   PATTERN "opensslconf.h.in" EXCLUDE)
-install(FILES ${OPENSSL_BUILD_INCLUDES}/openssl/opensslconf.h
-        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/openenclave/3rdparty/openssl)
 
 # Create openssl export target.
 add_enclave_library(openssl INTERFACE)
--- a/3rdparty/openssl/README.md
+++ b/3rdparty/openssl/README.md
@@ -27,3 +27,10 @@ that work with Open Enclave. The structure of the directory is as follows.
   the version of Perl that comes with the git bash does not support the Pod::Usage module required by the OpenSSL
   configuration scripts. See https://groups.google.com/g/git-for-windows/c/AQf2YbaxH6U/m/mwRScOGRCwAJ
   for the reference on the perl issue with git bash on Windows.
+
+- unsupported.h
+  The header lists the OpenSSL APIs, which are unsupported by OE for security concerns.
+
+- append-unsupported
+  Script to append the `#include<openssl/unsupported.h>` to the OpenSSL headers that include the
+  corresponding APIs. The patched headers will be installed as part of OE release packages.
--- a/3rdparty/openssl/unsupported.h
+++ b/3rdparty/openssl/unsupported.h
@@ -1,0 +1,158 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#ifndef _OE_OPENSSL_UNSUPPORTED_H
+#define _OE_OPENSSL_UNSUPPORTED_H
+
+#if !defined(OE_OPENSSL_SUPPRESS_UNSUPPORTED)
+
+#if defined(__cplusplus)
+#define OE_OPENSSL_EXTERN_C_BEGIN \
+    extern "C"                    \
+    {
+#define OE_OPENSSL_EXTERN_C_END }
+#else
+#define OE_OPENSSL_EXTERN_C_BEGIN
+#define OE_OPENSSL_EXTERN_C_END
+#endif
+
+#if defined(__GNUC__)
+#define OE_OPENSSL_UNSUPPORTED(MSG) __attribute__((deprecated(MSG)))
+#else
+#define OE_OPENSSL_UNSUPPORTED(MSG)
+#endif
+#define OE_OPENSSL_RAISE_ERROR(NAME) (OE_OPENSSL_UNSUPPORTED_ERROR, NAME)
+
+OE_OPENSSL_EXTERN_C_BEGIN
+
+#include <openssl/crypto.h>
+#include <openssl/ssl.h>
+#include <openssl/x509_vfy.h>
+
+// clang-format off
+#define OE_UNSUPPORTED_ENCLAVE_OPENSSL_FUNCTION \
+    "The function may be unsafe inside an enclave as it causes the enclave to read " \
+    "files from an untrusted host. Please refer to the following link for security guidance.\n\n" \
+    "https://github.com/openenclave/openenclave/blob/master/docs/OpenSSLSupport.md#security-guidance-for-using-openssl-apismacros" \
+    "\n\nTo use the function anyway, add the -DOE_OPENSSL_SUPPRESS_UNSUPPORTED option when compiling the program."
+#define OE_UNSUPPORTED_ENCLAVE_OPENSSL_MACRO \
+    "The macro may be unsafe inside an enclave as it causes the enclave to read " \
+    "files from an untrusted host. Please refer to the following link for security guidance.\n\n" \
+    "https://github.com/openenclave/openenclave/blob/master/docs/OpenSSLSupport.md#security-guidance-for-using-openssl-apismacros" \
+    "\n\nTo use the macro anyway, add the -DOE_OPENSSL_SUPPRESS_UNSUPPORTED option when compiling the program."
+// clang-format on
+
+/*
+**==============================================================================
+**
+** <openssl/crypto.h>
+**
+**==============================================================================
+*/
+
+/*
+ * The following macro allows users to configure an OpenSSL application (via
+ * OPENSSL_init_crypto or OPENSSL_init_ssl) to lookup a configuration file from
+ * the host filesystem, which is not trusted in OE's threat model, and therefore
+ * are recommended not to use inside an enclave.
+ */
+
+#ifdef OPENSSL_INIT_LOAD_CONFIG
+#undef OPENSSL_INIT_LOAD_CONFIG
+OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_MACRO)
+static inline unsigned long OPENSSL_INIT_LOAD_CONFIG()
+{
+    return 0x00000040L;
+}
+#define OPENSSL_INIT_LOAD_CONFIG \
+    OE_OPENSSL_RAISE_ERROR(OPENSSL_INIT_LOAD_CONFIG)
+#endif
+
+/*
+**==============================================================================
+**
+** <openssl/ssl.h>
+**
+**==============================================================================
+*/
+
+/*
+ * The following APIs allow users to configure an OpenSSL application to look up
+ * certificates from the host filesystem, which is not trusted in OE's
+ * threat model, and therefore are recommended not to use inside an enclave.
+ */
+
+OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_FUNCTION)
+int SSL_CTX_set_default_verify_paths(SSL_CTX* ctx);
+#define SSL_CTX_set_default_verify_paths \
+    OE_OPENSSL_RAISE_ERROR(SSL_CTX_set_default_verify_paths)
+
+OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_FUNCTION)
+int SSL_CTX_set_default_verify_dir(SSL_CTX* ctx);
+#define SSL_CTX_set_default_verify_dir \
+    OE_OPENSSL_RAISE_ERROR(SSL_CTX_set_default_verify_dir)
+
+OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_FUNCTION)
+int SSL_CTX_set_default_verify_file(SSL_CTX* ctx);
+#define SSL_CTX_set_default_verify_file \
+    OE_OPENSSL_RAISE_ERROR(SSL_CTX_set_default_verify_file)
+
+OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_FUNCTION)
+int SSL_CTX_load_verify_locations(
+    SSL_CTX* ctx,
+    const char* CAfile,
+    const char* CApath);
+#define SSL_CTX_load_verify_locations \
+    OE_OPENSSL_RAISE_ERROR(SSL_CTX_load_verify_locations)
+
+/*
+**==============================================================================
+**
+** <openssl/x509_vfy.h>
+**
+**==============================================================================
+*/
+
+/*
+ * The following APIs allow users to configure an OpenSSL application to look up
+ * certificates from the host filesystem, which is not trusted in OE's
+ * threat model, and therefore are recommended not to use inside an enclave.
+ */
+
+OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_FUNCTION)
+int X509_load_cert_file(X509_LOOKUP* ctx, const char* file, int type);
+#define X509_load_cert_file OE_OPENSSL_RAISE_ERROR(X509_load_cert_file)
+
+OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_FUNCTION)
+int X509_load_crl_file(X509_LOOKUP* ctx, const char* file, int type);
+#define X509_load_crl_file OE_OPENSSL_RAISE_ERROR(X509_load_crl_file)
+
+OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_FUNCTION)
+int X509_load_cert_crl_file(X509_LOOKUP* ctx, const char* file, int type);
+#define X509_load_cert_crl_file OE_OPENSSL_RAISE_ERROR(X509_load_cert_crl_file)
+
+OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_FUNCTION)
+X509_LOOKUP_METHOD* X509_LOOKUP_hash_dir(void);
+#define X509_LOOKUP_hash_dir OE_OPENSSL_RAISE_ERROR(X509_LOOKUP_hash_dir)
+
+OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_FUNCTION)
+X509_LOOKUP_METHOD* X509_LOOKUP_file(void);
+#define X509_LOOKUP_file OE_OPENSSL_RAISE_ERROR(X509_LOOKUP_file)
+
+OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_FUNCTION)
+int X509_STORE_load_locations(
+    X509_STORE* ctx,
+    const char* file,
+    const char* dir);
+#define X509_STORE_load_locations \
+    OE_OPENSSL_RAISE_ERROR(X509_STORE_load_locations)
+
+OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_FUNCTION)
+int X509_STORE_set_default_paths(X509_STORE* ctx);
+#define X509_STORE_set_default_paths \
+    OE_OPENSSL_RAISE_ERRO(X509_STORE_set_default_paths)
+
+OE_OPENSSL_EXTERN_C_END
+
+#endif /* !defined(OE_OPENSSL_SUPPRESS_UNSUPPORTED) */
+#endif /* _OE_OPENSSL_UNSUPPORTED_H */
--- a/docs/OpenSSLSupport.md
+++ b/docs/OpenSSLSupport.md
@@ -148,6 +148,31 @@ Note that the code snippet for the RDRAND engine opt-in is required to use not o
 but also other OpenSSL APIs that internally depend on the RAND APIs. Alternatively, developers
 can implement their own RAND method to replace the default method via `RAND_set_rand_method` API.
 
+## Security Guidance for using OpenSSL APIs/Macros
+
+OpenSSL provides APIs that allow users to configure sensitive settings like certificate trust and cipher suite preference from files.
+Because the host file system is considered untrusted in contexts such as SGX enclaves, OE SDK marks these APIs as unsupported to discourage their use.
+OE SDK does this by patching the OpenSSL headers that include such APIs or macros (i.e., appending the inclusion of an "unsupported.h" file to these headers).
+This ensures that when an enclave includes a patched OpenSSL header and uses the specific API or macro, user will receive compile-time errors.
+The errors can be disabled by specifying the `OE_OPENSSL_SUPPRESS_UNSUPPORTED` option to the compiler.
+
+See the following table for the detailed list of APIs/Macros.
+
+API / Macro | Original header | Comments | Guidance |
+:---:|:---:|:---|:---|
+OPENSSL_INIT_LOAD_CONFIG | crypto.h | The macro represents an option to the OPENSSL_init_ssl and OPENSSL_init_crypto APIs that initializes an application based on the openssl.cnf (loaded from the host filesystem). Therefore, the use of this option would allow an untrusted host to fully control the initialization of the application. | The recommendation is not to invoke initialization APIs with this option. Note that starting from v1.1.0, the explicit initialization is not required. |
+SSL_CTX_load_verify_locations | ssl.h | The API specifies the default locations from which an application looks up CA certificates for verification purposes. The API would allow an untrusted host to fully control what certificates the application will trust. | The recommendation is using the SSL_CTX_set_cert_store API that specifies an in-memory certificate verification storage (`X509_STORE`). Another alternative is to implement a customized verification callback and sets the callback via the SSL_CTX_set_verify API, which effectively bypasses the default implementation. |
+SSL_CTX_set_default_verify_paths | ssl.h | The API specifies the locations as part of arguments from which an application looks up CA certificates for verification purposes. The API would allow an untrusted host to fully control what certificates the application will trust. | The recommendation is using the SSL_CTX_set_cert_store API that specifies an in-memory certificate verification storage (`X509_STORE`). Another alternative is to implement a customized verification callback and sets the callback via the SSL_CTX_set_verify API, which effectively bypasses the default implementation. |
+SSL_CTX_set_default_verify_dir | ssl.h | Similar to SSL_CTX_set_default_verify_paths except for specifying dir only. | The recommendation is using the SSL_CTX_set_cert_store API that specifies an in-memory certificate verification storage (`X509_STORE`). Another alternative is to implement a customized verification callback and sets the callback via the SSL_CTX_set_verify API, which effectively bypasses the default implementation. |
+SSL_CTX_set_default_verify_file | ssl.h | Similar to SSL_CTX_set_default_verify_paths except for specifying file only. | The recommendation is using the SSL_CTX_set_cert_store API that specifies an in-memory certificate verification storage (`X509_STORE`). Another alternative is to implement a customized verification callback and sets the callback via the SSL_CTX_set_verify API, which effectively bypasses the default implementation. |
+X509_LOOKUP_hash_dir | x509_vfy.h | This API returns a `X509_LOOKUP_METHOD` method that loads files (certificates or CRLs) from the path specified by the `SSL_CERT_DIR` environment variable. This would allow an untrusted host to control what files the enclave will load. The API is used internally by SSL_CTX_set_default_verify_dir and SSL_CTX_set_default_verify_paths. | The recommendation is to implement a customized `X509_LOOKUP_METHOD` method based on in-memory operations. Note that to opt-in the new method, the user needs to explicitly register the method to a `X509_STORE` via the X509_STORE_add_lookup API and the set the `X509_STORE` to an `SSL_CTX` via SSL_CTX_set_cert_store. |
+X509_LOOKUP_file | x509_vfy.h | This API returns a `X509_LOOKUP_METHOD` method that loads files (certificates or CRLs) from the path specified by the `SSL_CERT_FILE` environment variable. This would allow an untrusted host to control what files the enclave will load. The API is used internally by SSL_CTX_set_default_verify_file and SSL_CTX_set_default_verify_paths. | The recommendation is to implement a customized `X509_LOOKUP_METHOD` method based on in-memory operations. Note that to opt-in the new method, the user needs to explicitly register the method to a `X509_STORE` via the X509_STORE_add_lookup API and the set the `X509_STORE` to an `SSL_CTX` via SSL_CTX_set_cert_store. |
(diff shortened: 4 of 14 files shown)


mingweishih:
bors try

dthaler (changes requested):
If I understand this PR right, I think this is going in the wrong direction.
The APIs are not unsafe.  Access to non-sealed files in the host is indeed unsafe,
but these APIs do not look like they are limited to that.

TrustZone and other TEEs do have secure storage (albeit not large, but large enough to store keys), such as RPMB.   

The OE FS APIs are pluggable (Mike's document at https://github.com/openenclave/openenclave/blob/master/docs/UsingTheIOSubsystem.md covers much of this) and the intent of OP-TEE devs is to be able to add secure FS's.  Deprecating these APIs would prevent this mechanism from being used to get keys from secure FS.

I have not seen any design doc covering this PR, so would recommend this be discussed in SIG-Arch before proceeding further.

mingweishih:
> If I understand this PR right, I think this is going in the wrong direction.
> The APIs are not unsafe. Access to non-sealed files in the host is indeed unsafe,
> but these APIs do not look like they are limited to that.
> 
> TrustZone and other TEEs do have secure storage (albeit not large, but large enough to store keys), such as RPMB.
> 
> The OE FS APIs are pluggable (Mike's document at https://github.com/openenclave/openenclave/blob/master/docs/UsingTheIOSubsystem.md covers much of this) and the intent of OP-TEE devs is to be able to add secure FS's. Deprecating these APIs would prevent this mechanism from being used to get keys from secure FS.
> 
> I have not seen any design doc covering this PR, so would recommend this be discussed in SIG-Arch before proceeding further.

We currently only plan to support OpenSSL for SGX. The suggested changes are part of conclusion from the threat model review of  OpenSSL support in OE.

I totally agree that these APIs are safe if we have secure FS for SGX. However, given that OE does not support it yet so the recommendation is deprecating these APIs for now.

I'll wait for @CodeMonkeyLeet's comments on this once he's back.

3rdparty/openssl/append-deprecations
```diff
@@ -0,0 +1,43 @@
...
+
+headers=`find ${directory} -name '*.h'`
+
+## Append the deprecation inclusion */
+for i in ${headers}
+do
+    if [[ "${i}" != *"ssl.h" ]] && [[ "${i}" != *"x509_vfy.h" ]]; then
+        continue
+    fi
+    echo "#include <openssl/deprecations.h>" >> ${i}
```
anakrish, line 38: Instead of modifying each file, an alternative is to add `-include openssl/deprecations.h` as a command line to the compiler via cmake/pkgconfig.

3rdparty/openssl/deprecations.h
```diff
@@ -0,0 +1,97 @@
...
+**==============================================================================
+*/
+
+/*
+ * The following APIs allow users to configure an OpenSSL application to lookup
+ * certificates from the host filesystem, which is not trusted in OE's
+ * threat model, and therefore are recommended not to use inside an enclave.
+ */
+
+OE_OPENSSL_DEPRECATED(OE_UNSUPPORTED_ENCLAVE_FUNCTION)
```
anakrish, line 47: Why are these functions only deprecated and not removed? 
Do we expect some users to use these functions?

dthaler:
> We currently only plan to support OpenSSL for SGX. 

https://github.com/openenclave/openenclave/blob/master/docs/Community/Governance.md#community-governance-structure
explains that technical decisions are made by SIGs.

I have seen no such decision by SIG-OPTEE, and indeed I suspect SIG-OPTEE would say the opposite.

> The suggested changes are part of conclusion from the threat model review of OpenSSL support in OE.

Where is that review documented?   When was it reviewed with SIG-Arch or any other SIG?

> I totally agree that these APIs are safe if we have secure FS for SGX. 

Great.

> However, given that OE does not support it yet so the recommendation is deprecating these APIs for now.

Recommendation from whom to whom?   API deprecation is a SIG API decision.

Furthermore, from an API SIG standpoint, not supporting something until the future is not a reason to have something be "deprecated".   It is debatably an argument to make something be experimental, which would be a separate discussion.

I have seen no such recommendation to SIG-API, we need to make sure we are following the OE governance process.

dthaler:
FYI, the OE [API process guidelines](https://github.com/openenclave/openenclave/blob/master/docs/ApiGuidelines.md) state:

> * Any breaking change to, or deprecation of, a released API must have a design reviewed and approved before the change is made, where the design minimally must state why a change is needed.

mingweishih:
A question before proceeding to sig discussion: Does the guideline apply to third party APIs?

dthaler:
> A question before proceeding to sig discussion: Does the guideline apply to third party APIs?

I believe the intent is all public APIs in the OE repo.  Normally deprecation of, or any other changes to, third party APIs is done in the third-party repo, not in the OE repo.  So any public API that is in the OE repo would fall under the purview of SIG API.

CodeMonkeyLeet:
@dthaler @mingweishih 

The discussion around this PR seems to have spiraled out beyond the scope of this code change into governance and process questions. I would suggest that more progress can be made by refocusing the discussion on specific design and implementation concerns.

I've added the topic to the next SIG-arch meeting for a more direct discussion, but I'll add a couple of observations and opinions here in the meantime:
- The OpenSSL feature is currently marked as experimental (gated by a build flag). It is not an existing or public feature set, nor has the expectation been set that it provides 100% of the functionality in normal mode OpenSSL 1.1.1 out of the box. See the [parity comparison with mbedTLS](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/CryptoLibConfiguration.md) for example. The attempt to use the API guidelines on deprecation and expanding its reading to include 3rd party APIs strikes me as disingenuous in this case. It is sufficient as a reviewer to indicate that a particular implementation and its design are not well-explained and request further elaboration and/or documentation. 
- That the deprecations should not be applied to OP-TEE is a reasonable point, but that does not invalidate or contradict the assertion that this particular contribution is only being implemented and tested for SGX. Asserting that SIG-OPTEE (whose charter remains undefined and whose members are not tagged as required reviewers in the current system for this PR) is somehow the relevant arbiter for this PR is unproductive at best. It would be much more helpful to point out specific changes in the review for the work to be more easily adapted for use in OP-TEE. 
- If there remain broader philosophical considerations around governance, SIG responsibilities and processes independent of the contents of this PR, we should expand on that discussion in the CGC. For example, one of the questions that seems to be glossed over in the discussion is how we should appropriately capture input from subject matter experts (e.g. cryptographers, pen-testers) not affiliated with the project.

dthaler:
I agree that there are both process and technical issues with this PR.
The specific issue with this PR is that I don't believe that "experimental" APIs/features should use the term "deprecate", you just break them.  So it is ok to not add APIs in an experimental feature, or remove APIs previously marked experimental, or change the prototype of experimental APIs, but I would never use the term "deprecate" to apply to anything in an "experimental" feature.   This is consistent with Anand's comment, where he said "Why are these functions only deprecated and not removed?"

We have a documented process for "deprecation", and I agree with those who have said that that process should not apply to experimental stuff and this PR is about an experimental feature.  Hence I maintain we should not use the word deprecate in this PR.

CodeMonkeyLeet:
@dthaler Regarding the use of the phrase "deprecation", I think you're taking that to be synonymous with a process of marking an API as obsolete. That's commonly the case, but I've seen the term applied to any APIs that are unrecommended or "disapproved" for use (in the specific case of the obsolescence process, because the API is being replaced/removed). For example, the C libraries and linux system libraries do this where the APIs may continue to exist indefinitely, but are flagged as unsuitable for use.

Feel free to suggest a different term that you feel better expresses the intent, but I don't find the usage of the term particularly problematic in this context; OE SDK itself has a precedence for this usage as it applies a deprecations.h header on top of musl headers, and those deprecations may themselves be removed in the future as broader functionality becomes supported. This PR looks like it was modeled directly off that existing example.

Removing the functions entirely is an option, but involves surgery in the OpenSSL source code and precludes the removed functions from being used in other contexts (such as OP-TEE, which you had expressed concern about). The inclusion of a deprecation header in default build configurations that can be overridden or contextually applied strikes me as a more flexible approach to providing guard rails on the use of those APIs.

anakrish:
The reason I'd prefer the functions to be removed is that we don't require that the user build their enclave with `-Werror` which would turn the deprecation warnings to errors. Therefore, there is a possibility that an user may end up calling these unsafe APIs.

Two approaches to ensure that the user can't use these unsafe APIs would be:
- Force the `-Werror` on the use if openssl is used. This can be accomplished via cmake/pkgconfig.
- Alternatively, use macros in deprecations.h so that an unsafe function `openssl_foo` becomes an undeclared identifier`openssl_foo_unsafe_removed`, resuling in compiler error if the unsafe function is invoked by the user.

CodeMonkeyLeet:
@anakrish I think those are both good suggestions, a couple of thoughts to add to that:
- `-Werror` is cheaper to implement though I think we backed away from that in the libc case. Setting it now would also affect those deprecations retroactively (and existing enclaves with ignored warnings). I think we would not want to do this given the broader user impact.
- Redirecting functions to undefined symbols is more targeted and preferable if we want to force a block. Is there a way to provide a custom error message instead of just hitting the missing symbol compile error though? The `deprecated` attribute/declspec has the benefit of attaching a custom message to the warning it generates and it would be helpful to be able to do the same here.

On a related note, if we are including the deprecation header by default (either via header injection or cmake/pkgconfig) and it will force breaks, we might also want to add a `#define` so that a dev can turn it off easily. For example, I believe the VC++ header libraries do this for their deprecations, where the warning message will include the macro to define to disable a specific subset of the deprecations.

anakrish:
>  The `deprecated` attribute/declspec has the benefit of attaching a custom message to the warning it generates and it would be helpful to be able to do the same here.

Yes, it is possible to generate a deprecation warning as well as a compile error by defining a macro as below:
```
__attribute__((deprecated("This API is unsafe")))
void foo(void);

// Define foo to be a macro that takes no arguments and evaluates to a comma separated expression that
// results in a compile error as well as the deprecation message.
#define foo    (foo_is_unsafe_for_use_within_enclaves, foo) 
```

Here is the user-experience: https://godbolt.org/z/hT8Mcn

3rdparty/openssl/deprecations.h
```diff
@@ -0,0 +1,97 @@
...
+int X509_load_cert_file(X509_LOOKUP* ctx, const char* file, int type);
```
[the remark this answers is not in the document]
mingweishih: I retain this API as part of deprecations since it's used internally by SSL_CTX_set_default_verify_* APIs.

mingweishih:
bors try

3rdparty/openssl/deprecations.h
```diff
@@ -0,0 +1,126 @@
...
+
+#ifndef OE_DISABLE_OPENSSL_DEPRECATIONS
+#define OE_OPENSSL_DEPRECATED(MSG) __attribute__((deprecated(MSG)))
+#else
+#define OE_OPENSSL_DEPRECATED(MSG)
+#endif
+
+OE_OPENSSL_EXTERN_C_BEGIN
+
+#include <openssl/crypto.h>
```
CodeMonkeyLeet, line 27: I'm guessing you're forcing the inclusion here so that the order in which the deprecations.h is included won't matter, but it does remind me of Anand's concern that the cmake/pkgconfig always pulls in the deprecation header for all object builds in the openSSL configuration, which means any module built targeting the openSSL configuration will include openSSL headers now (even if that particular component doesn't invoke openSSL).

I'm not sure there's a good way around that if we're not injecting this into the openSSL headers.

3rdparty/openssl/deprecations.h
```diff
@@ -0,0 +1,126 @@
...
+    extern "C"                    \
+    {
+#define OE_OPENSSL_EXTERN_C_END }
+#else
+#define OE_OPENSSL_EXTERN_C_BEGIN
+#define OE_OPENSSL_EXTERN_C_END
+#endif
+
+#ifndef OE_DISABLE_OPENSSL_DEPRECATIONS
+#define OE_OPENSSL_DEPRECATED(MSG) __attribute__((deprecated(MSG)))
```
CodeMonkeyLeet, line 20: From the discussions, it sounded like we wanted to not use the compiler deprecation mechanism and directly force compilation to break via macro redefinitions.

@anakrish?

docs/OpenSSLSupport.md
```diff
@@ -148,6 +148,31 @@ Note that the code snippet for the RDRAND engine opt-in is required to use not o
...
 can implement their own RAND method to replace the default method via `RAND_set_rand_method` API.
 
+## Security Guidance of using OpenSSL APIs
+
+Few OpenSSL APIs allow users to configure an application to load files from the host filesystem.
+Such behaviors may have security implications given that the host filesystem is untrusted under the
+threat model of OE. For these reasons, OE recommends users not to use those APIs and
+put them into `deprecations.h` as part of OpenSSL headers that are shipped along with the release package.
+OE's default compiler settings explicitly include the header such that users get warnings when an
+enclave uses any of these APIs.
```
CodeMonkeyLeet, line 158:
```suggestion
OpenSSL provides APIs that allow users to configure sensitive settings like certificate trust and cryptosuite preference from files.
Because the host file system is considered untrusted in contexts such as SGX enclaves, OE SDK marks these APIs as deprecated to discourage their use.
OE SDK does this by including a "deprecations.h" file as part of the default cmake/pkgconfig build configuration,
which will produce compiler warnings when any of these APIs are used in an enclave.
The warnings can be disabled by specifying the `OE_DISABLE_OPENSSL_DEPRECATIONS` option to the compiler.
```

docs/OpenSSLSupport.md
```diff
@@ -148,6 +148,31 @@ Note that the code snippet for the RDRAND engine opt-in is required to use not o
...
+
+See the following table for the detailed list of APIs.
+
+API | Original header | Comments
+:---:|:---:|:---|
+SSL_CTX_load_verify_locations | ssl.h | The API specifies the default locations from which an application looks up CA certificates for verification purposes. The API would allow an untrusted host to force the application to load fake certificates. |
```
CodeMonkeyLeet, line 164: We should probably also provide guidance as to what the alternative that a developer should use is.

tests/openssl/enc/CMakeLists.txt
```diff
@@ -95,9 +95,8 @@ function (add_openssl_test_enc NAME BUILDTEST)
     enc.c
     ${CMAKE_CURRENT_BINARY_DIR}/openssl_t.c)
 
-  enclave_include_directories(
-    openssl-${NAME}_enc PRIVATE ${CMAKE_CURRENT_BINARY_DIR}
-    ${OPENSSL_DIR}/include ${CMAKE_BINARY_DIR}/3rdparty/openssl/build/include)
+  enclave_include_directories(openssl-${NAME}_enc PRIVATE
```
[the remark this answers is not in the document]
mingweishih: The test will inherent the headers when linking against the OpenSSL libraries so I remove the redundant include here.

3rdparty/openssl/deprecations.h
mingweishih: Yes, the intention is forcing the order of inclusion because we're using the `-include` option and its order is not under our control. Is there any practice could walkaround this problem?

@anakrish

dthaler:
> Feel free to suggest a different term that you feel better expresses the intent

"Unsupported" is the precedent, which sounds fine to me.
E.g., in Mike's musl/deprecations.h we see:
`OE_LIBC_DEPRECATED(OE_UNSUPPORTED_ENCLAVE_FUNCTION)`

> OE SDK itself has a precedence for this usage as it applies a deprecations.h header on top of musl headers, and those deprecations may themselves be removed in the future as broader functionality becomes supported.

Yes (though the misuse of "deprecations" predates the API guidelines and my involvement in that area), though as noted above it implicitly admits that unsupported is a better term.

And yes removing the "deprecations" has already been done, e.g., strcpy was un-deprecated even though it is arguably just as dangerous as the openssl apis we're talking about here, so strcpy is a great point of comparison.

> Removing the functions entirely is an option, but involves surgery in the OpenSSL source code and precludes the removed functions from being used in other contexts (such as OP-TEE, which you had expressed concern about). 

Understood.  I would be happy with either of the following two approaches:
1) Don't add the APIs to headers/symbols until they are supported
2) Add them but mark them as UNSUPPORTED (instead of DEPRECATED)

radhikaj:
Per SIG-Arch discussion
1. Mark the APIs as unsupported using the # define mechanism
2.

radhikaj:
Action items from SIG-Arch:
1. Use UNSUPPORTED instead of DEPRECATED
2. Calling the APIs that are marked UNSUPPORTED - make it an error
3. The error should be a descriptive string that tells you what to instead. - WHY, alternative and how to turn it off.
4. Ubiquitous -independent of compiler(so not just pkgconfig and cmake) - If you add an openssl header first or the openenclave header first - both cases should result in the inclusion of deprecation headers. This PR already does this.

3rdparty/openssl/deprecations.h
mingweishih: This should be concern because we revert to the header patching approach.

tests/openssl_deprecated/enc/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+  enclave_compile_definitions(openssl_deprecated_${NAME}_enc PRIVATE ${DEFINE})
+
+  enclave_include_directories(openssl_deprecated_${NAME}_enc PRIVATE
+                              ${CMAKE_CURRENT_BINARY_DIR})
+
+  enclave_link_libraries(openssl_deprecated_${NAME}_enc openssl oelibc
+                         oehostsock oehostfs oehostresolver)
+
+  # Exclude the enclave from build.
+  # From: https://stackoverflow.com/questions/30155619/expected-build-failure-tests-in-cmake
```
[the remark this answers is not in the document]
mingweishih: Done.

tests/openssl/enc/CMakeLists.txt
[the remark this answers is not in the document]
mingweishih: The changes are outdated.

3rdparty/openssl/deprecations.h
```diff
@@ -0,0 +1,126 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#ifndef _OE_OPENSSL_DEPRECATIONS_H
+#define _OE_OPENSSL_DEPRECATIONS_H
+
+#if !defined(OE_OPENSSL_SUPPRESS_DEPRECATIONS)
```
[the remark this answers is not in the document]
mingweishih: Fixed.

3rdparty/openssl/deprecations.h
mingweishih: Changed to enforce errors instead of just warnings.

3rdparty/openssl/deprecations.h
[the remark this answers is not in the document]
mingweishih: `SSL_CTX_load_verify_store` is introduced in OpenSSL v3.0.0

https://github.com/openssl/openssl/blob/dd0164e7565bb14fac193aea4c2c37714bf66d56/util/libssl.num#L513

mingweishih:
@CodeMonkeyLeet  @dthaler  @anakrish 

I've addressed all the comments we discussed except for the samples/guidance that demonstrates alternatives to these unsupported APIs. Given the such alternatives are really implementation-specific (i.e., there is no single replacement to each API), so I'd like to put that to another PR. What do you think?

CodeMonkeyLeet (commented):
I think that the samples work is already planned as a separate set of PRs, but I would like to see a clear articulation of what the intended alternative is before we merged this; it seems like an incomplete design to go block a bunch of functions without being able to describe what the intended use case should do without those functions. 

In part, the design exercise I'm expecting here is at least documenting the assertion that there is no single replacement for each API. Ideally, identify the family of replacement functions as well. Note that there isn't an expectation that there needs to be a 1:1 API replacement, but that there is _at least_ one (preferably the next simplest) alternative we can point to. For example `SSL_CTX_set_cert_store` might cover the various `SSL_CTX_set_default*` we want to block, and it might require a few additional calls to have it set up, but we can identify the strategy around it. 

Where multiple strategies may exist (e.g. you mentioned callbacks), I would like to have some pros/cons discussion of which strategy we would recommend and have that be documented in the design.

3rdparty/openssl/unsupported.h
```diff
@@ -0,0 +1,161 @@
...
+/*
+ * The following macro allows users to configure an OpenSSL application (via
+ * OPENSSL_init_crypto or OPENSSL_init_ssl) to lookup a configuration file from
+ * the host filesystem, which is not trusted in OE's threat model, and therefore
+ * are recommended not to use inside an enclave.
+ */
+
+#ifdef OPENSSL_INIT_LOAD_CONFIG
+#undef OPENSSL_INIT_LOAD_CONFIG
+OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_MACRO)
```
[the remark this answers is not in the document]
mingweishih: This allows for consistent behaviors between functions and macros (i.e., throwing the message via `__attribute__((deprecated(MSG)))`). Would you suggest any alternative?

docs/OpenSSLSupport.md
```diff
@@ -148,6 +148,31 @@ Note that the code snippet for the RDRAND engine opt-in is required to use not o
...
+OE SDK does this by patching the OpenSSL headers that include such APIs or macros (i.e., appending the inclusion of an "unsupported.h" file to these headers).
+This ensures that when an enclave includes a patched OpenSSL header and uses the specific API or macro, user will receive compile-time errors.
+The errors can be disabled by specifying the `OE_OPENSSL_SUPPRESS_UNSUPPORTED` option to the compiler.
+
+See the following table for the detailed list of APIs/Macros.
+
+API / Macro | Original header | Comments
+:---:|:---:|:---|
+OPENSSL_INIT_LOAD_CONFIG | crypto.h | The macro represents an option to the OPENSSL_ssl_init and OPENSSL_crypto_init that initializes an application based on the openssl.cnf (loaded from the host filesystem). Therefore, the use of this option would allow an untrusted host to fully control the initialization of the application.  |
+SSL_CTX_load_verify_locations | ssl.h | The API specifies the default locations from which an application looks up CA certificates for verification purposes. The API would allow an untrusted host to force the application to load fake certificates. |
```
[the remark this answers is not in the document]
mingweishih: Updated the comments to be more accurate.

docs/OpenSSLSupport.md
```diff
@@ -148,6 +148,31 @@ Note that the code snippet for the RDRAND engine opt-in is required to use not o
...
+SSL_CTX_set_default_verify_paths | ssl.h | The API specifies the locations as part of arguments from which an application looks up CA certificates for verification purposes. The API would allow an untrusted host to force the application to load fake certificates. |
+SSL_CTX_set_default_verify_dir | ssl.h | Similar to SSL_CTX_set_default_verify_paths except for specifying dir only. |
+SSL_CTX_set_default_verify_file | ssl.h | Similar to SSL_CTX_set_default_verify_paths except for specifying file only. |
+X509_LOOKUP_hash_dir | x509_vfy.h | This API returns a `X509_LOOKUP_METHOD` method that loads files from the path specified by the `SSL_CERT_DIR` environment variable. The API is used by SSL_CTX_set_default_verify_dir and SSL_CTX_set_default_verify_paths internally.  |
```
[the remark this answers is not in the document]
mingweishih: I added a recommendation to implement a customized `X509_LOOPUP_METHOD`. Will that be sufficient?

3rdparty/openssl/CMakeLists.txt
```diff
@@ -80,6 +85,27 @@ ExternalProject_Add(
   BUILD_COMMAND ""
   INSTALL_COMMAND "")
 
+# The following commmand prepares the headers to be published. The headers
+# consist of those in the openssl/include/openssl folder, the opensslconf.h,
+# and the unsupported.h. The unsupported.h includes a list of APIs that are
+# disabled by OE for security concerns. Few headers are patched (using the
+# append-unsupported script) to include the unsupported.h so that OE can enforce
```
dthaler, line 92:
```suggestion
# consist of those in the openssl/include/openssl folder, opensslconf.h,
# and unsupported.h. The unsupported.h header file includes a list of APIs that are
# disabled by OE for security concerns. A few headers are patched (using the
# append-unsupported script) to include unsupported.h so that OE can enforce
```

3rdparty/openssl/README.md
```diff
@@ -27,3 +27,10 @@ that work with Open Enclave. The structure of the directory is as follows.
   the version of Perl that comes with the git bash does not support the Pod::Usage module required by the OpenSSL
   configuration scripts. See https://groups.google.com/g/git-for-windows/c/AQf2YbaxH6U/m/mwRScOGRCwAJ
   for the reference on the perl issue with git bash on Windows.
+
+- unsupported.h
+  The header lists the OpenSSL APIs, which are deprecated by OE for security concerns.
```
dthaler, line 32:
```suggestion
  The header lists the OpenSSL APIs, which are unsupported by OE for security concerns.
```

3rdparty/openssl/unsupported.h
```diff
@@ -0,0 +1,154 @@
...
+#define OE_OPENSSL_EXTERN_C_BEGIN \
+    extern "C"                    \
+    {
+#define OE_OPENSSL_EXTERN_C_END }
+#else
+#define OE_OPENSSL_EXTERN_C_BEGIN
+#define OE_OPENSSL_EXTERN_C_END
+#endif
+
+#define OE_OPENSSL_UNSUPPORTED(MSG) __attribute__((deprecated(MSG)))
```
dthaler, line 19: bits/defs.h has this:
```
#if defined(__GNUC__)
#define OE_DEPRECATED(FUNC, MSG) FUNC __attribute__((deprecated(MSG)))
#elif defined(_MSC_VER)
#define OE_DEPRECATED(FUNC, MSG) __declspec(deprecated(MSG)) FUNC
#else
#define OE_DEPRECATED(FUNC, MSG) FUNC
#endif 
```

But here only has the _GNUC_ line.  Can you confirm that either:
a) this was not intentional and update this PR to make it match, OR
b) this header will not appear in the OE SDK package?

3rdparty/openssl/unsupported.h
```diff
@@ -0,0 +1,154 @@
...
+/*
+**==============================================================================
+**
+** <openssl/ssl.h>
+**
+**==============================================================================
+*/
+
+/*
+ * The following APIs allow users to configure an OpenSSL application to lookup
```
dthaler, line 76:
```suggestion
 * The following APIs allow users to configure an OpenSSL application to look up
```

3rdparty/openssl/unsupported.h
```diff
@@ -0,0 +1,154 @@
...
+/*
+**==============================================================================
+**
+** <openssl/x509_vfy.h>
+**
+**==============================================================================
+*/
+
+/*
+ * The following APIs allow users to configure an OpenSSL application to lookup
```
dthaler, line 113:
```suggestion
 * The following APIs allow users to configure an OpenSSL application to look up
```

docs/OpenSSLSupport.md
```diff
@@ -148,6 +148,31 @@ Note that the code snippet for the RDRAND engine opt-in is required to use not o
 but also other OpenSSL APIs that internally depend on the RAND APIs. Alternatively, developers
 can implement their own RAND method to replace the default method via `RAND_set_rand_method` API.
 
+## Security Guidance of using OpenSSL APIs/Macros
```
dthaler, line 151:
```suggestion
## Security Guidance for using OpenSSL APIs/Macros
```
Note: assuming you accept this suggestion, you will need to update the cross-reference links to this section below.

docs/OpenSSLSupport.md
```diff
@@ -251,7 +276,7 @@ ui.h | No | Configured with no-ui-console. |
 uierr.h | Yes | - |
 whrlpool.h | No | Whirlpool is disabled by OE. |
 x509.h | Yes | - |
-x509_vfy.h | Yes | - |
+x509_vfy.h | Partial | Functions that are deprecated by OE for security concerns include: `X509_load_cert_file`, `X509_load_crl_file`, `X509_LOOKUP_hash_dir`, `X509_LOOKUP_file`, `X509_load_cert_crl_file`, `X509_STORE_load_locations`, `X509_STORE_set_default_paths`. Refer to [Security Guidance of using OpenSSL APIs/Macros](#security-guidance-of-using-openssl-apismacros) for more detail. |
```
dthaler, line 279:
```suggestion
x509_vfy.h | Partial | Functions that are unsupported by OE for security concerns include: `X509_load_cert_file`, `X509_load_crl_file`, `X509_LOOKUP_hash_dir`, `X509_LOOKUP_file`, `X509_load_cert_crl_file`, `X509_STORE_load_locations`, `X509_STORE_set_default_paths`. Refer to [Security Guidance of using OpenSSL APIs/Macros](#security-guidance-of-using-openssl-apismacros) for more detail. |
```

tests/openssl_unsupported/enc/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+                         EXCLUDE_FROM_ALL TRUE EXCLUDE_FROM_DEFAULT_BUILD TRUE)
+
+  add_test(
+    NAME openssl_unsupported_${NAME}
+    COMMAND ${CMAKE_COMMAND} --build . --target openssl_unsupported_${NAME}_enc
+            --config $<CONFIGURATION>
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+
+  # Instead of expecting the compilation to fail, checking whether the log includes
+  # the specific error message, which indicates the deprecation header is
```
dthaler, line 45:
```suggestion
  # the specific error message, which indicates the unsupported.h header is
```

tests/openssl_unsupported/enc/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+  # correctly included.
+  set_tests_properties(
+    openssl_unsupported_${NAME}
+    PROPERTIES
+      WILL_FAIL
+      TRUE
+      FAIL_REGULAR_EXPRESSION
+      "warning: .* is deprecated: The function|macro may be unsafe inside an enclave"
```
dthaler, line 53:
```suggestion
      "warning: .* is unsupported: The function|macro may be unsafe inside an enclave"
```

tests/openssl_unsupported/enc/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+  )
+
+endfunction (add_unsupported_test)
+
+set(DEPRECATED_LIST
```
dthaler, line 58:
```suggestion
set(UNSUPPORTED_LIST
```

tests/openssl_unsupported/enc/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+    SSL_CTX_load_verify_locations
+    X509_load_cert_file
+    X509_load_crl_file
+    X509_load_cert_crl_file
+    X509_LOOKUP_hash_dir
+    X509_LOOKUP_file
+    X509_STORE_load_locations
+    X509_STORE_set_default_paths)
+
+foreach (name ${DEPRECATED_LIST})
```
dthaler, line 72:
```suggestion
foreach (name ${UNSUPPORTED_LIST})
```

docs/OpenSSLSupport.md
```diff
@@ -235,7 +260,7 @@ seed.h | No | SEED is disabled by OE. |
 sha.h | Yes | - |
 srp.h | No | SRC is disabled by OE. |
 srtp.h | Yes | - |
-ssl.h | Partial | SSL2 and SSL3 methods are disabled. Heartbeats extension is disabled by defailt. |
+ssl.h | Partial | SSL2 and SSL3 methods are disabled. Heartbeats extension is disabled by default. Functions that are deprecated by OE for security concerns include: `SSL_CTX_set_default_verify_paths`, `SSL_CTX_set_default_verify_dir`, `SSL_CTX_set_default_verify_file`, `SSL_CTX_load_verify_locations`. Refer to [Security Guidance of using OpenSSL APIs/Macros](#security-guidance-of-using-openssl-apismacros) for more detail |
```
dthaler, line 263:
```suggestion
ssl.h | Partial | SSL2 and SSL3 methods are disabled. Heartbeats extension is disabled by default. Functions that are unsupported by OE for security concerns include: `SSL_CTX_set_default_verify_paths`, `SSL_CTX_set_default_verify_dir`, `SSL_CTX_set_default_verify_file`, `SSL_CTX_load_verify_locations`. Refer to [Security Guidance of using OpenSSL APIs/Macros](#security-guidance-of-using-openssl-apismacros) for more detail |
```

tests/openssl_unsupported/host/host.c
```diff
@@ -0,0 +1,42 @@
...
+    if (argc != 2)
+    {
+        fprintf(stderr, "Usage: %s ENCLAVE_PATH\n", argv[0]);
+        return 1;
+    }
+
+    const uint32_t flags = oe_get_create_flags();
+
+    if ((result = oe_create_openssl_unsupported_enclave(
+             argv[1], OE_ENCLAVE_TYPE_SGX, flags, NULL, 0, &enclave)) != OE_OK)
```
dthaler, line 28:
```suggestion
             argv[1], OE_ENCLAVE_TYPE_AUTO, flags, NULL, 0, &enclave)) != OE_OK)
```
I don't see anything specific to SGX in this sample, so avoid hard coding it specifically.

dthaler (commented):
Looking much better, thanks for the updates.  See remaining suggestions.

tests/openssl_unsupported/enc/CMakeLists.txt
mingweishih: The `warning: .* is deprecated:` part is generated by compiler and not under our control.

3rdparty/openssl/unsupported.h
mingweishih: Updated. Given that we don't support compiling enclaves with MSVC, I skipped the check.

3rdparty/openssl/unsupported.h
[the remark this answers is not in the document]
mingweishih: cc @anakrish

mingweishih:
bors try

tests/openssl_unsupported/enc/CMakeLists.txt
dthaler: Ah, got it, thanks.

dthaler (approved):
Looks good to me now, thanks for addressing the feedback

3rdparty/openssl/unsupported.h
```diff
@@ -0,0 +1,158 @@
...
+#define OE_OPENSSL_EXTERN_C_BEGIN
+#define OE_OPENSSL_EXTERN_C_END
+#endif
+
+#if defined(__GNUC__)
+#define OE_OPENSSL_UNSUPPORTED(MSG) __attribute__((deprecated(MSG)))
+#else
+#define OE_OPENSSL_UNSUPPORTED(MSG)
+#endif
+#define OE_OPENSSL_RAISE_ERROR(NAME) (OE_OPENSSL_UNSUPPORTED_ERROR, NAME)
```
anakrish, line 24: GCC now has a `error` attribute just like `deprecated` attribute. It will raise an error if only if the function is used.
clang does not yet support it. 
So our macro trick is good for now.

anakrish (approved):
LGTM!

3rdparty/openssl/unsupported.h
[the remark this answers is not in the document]
CodeMonkeyLeet: To clarify, what I mean is that with the use of `OE_OPENSSL_RAISE_ERROR` the use of the `deprecated` attribute on the function is unnecessary. Anand's example in the comment illustrates the ability of the macro approach to return a meaningful error in comparison to the deprecation error message, but combining them seems redundant.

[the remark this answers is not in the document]
mingweishih: My understanding is that the combination is necessary:
- deprecated attribute allows us to provide customized messages, but it generates warnings only.
- OE_OPENSSL_RAISE_ERROR enforces the compilation error (undefined error).

As @anakrish point out, the error attribute could does both, but it's currently not supported by clang.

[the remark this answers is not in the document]
mingweishih: For the reference, the current user experience is as follows: 

```
465: [100%] Building C object tests/openssl_unsupported/enc/CMakeFiles/openssl_unsupported_openssl_init_load_config_enc.dir/enc.c.o
465: /home/ssh_office/workspace/oe-ms2/tests/openssl_unsupported/enc/enc.c:18:25: error: use of undeclared identifier 'OE_OPENSSL_UNSUPPORTED_ERROR'
465:     unsigned long val = OPENSSL_INIT_LOAD_CONFIG;
465:                         ^
465: /home/ssh_office/workspace/oe-ms2/build/output/include/openenclave/openssl/unsupported.h:68:5: note: expanded from macro 'OPENSSL_INIT_LOAD_CONFIG'
465:     OE_OPENSSL_RAISE_ERROR(OPENSSL_INIT_LOAD_CONFIG)
465:     ^
465: /home/ssh_office/workspace/oe-ms2/build/output/include/openenclave/openssl/unsupported.h:24:39: note: expanded from macro 'OE_OPENSSL_RAISE_ERROR'
465: #define OE_OPENSSL_RAISE_ERROR(NAME) (OE_OPENSSL_UNSUPPORTED_ERROR, NAME)
465:                                       ^
465: /home/ssh_office/workspace/oe-ms2/tests/openssl_unsupported/enc/enc.c:18:25: warning: 'OPENSSL_INIT_LOAD_CONFIG' is deprecated: The macro may be unsafe inside an enclave as it causes the enclave to read files from an untrusted host. Please refer to the following link for security guidance.
465: 
465: https://github.com/openenclave/openenclave/blob/master/docs/OpenSSLSupport.md#security-guidance-for-using-openssl-apismacros
465: 
465: To use the macro anyway, add the -DOE_OPENSSL_SUPPRESS_UNSUPPORTED option when compiling the program. [-Wdeprecated-declarations]
465: /home/ssh_office/workspace/oe-ms2/build/output/include/openenclave/openssl/unsupported.h:68:28: note: expanded from macro 'OPENSSL_INIT_LOAD_CONFIG'
465:     OE_OPENSSL_RAISE_ERROR(OPENSSL_INIT_LOAD_CONFIG)
465:                            ^
465: /home/ssh_office/workspace/oe-ms2/build/output/include/openenclave/openssl/unsupported.h:62:1: note: 'OPENSSL_INIT_LOAD_CONFIG' has been explicitly marked deprecated here
465: OE_OPENSSL_UNSUPPORTED(OE_UNSUPPORTED_ENCLAVE_OPENSSL_MACRO)
465: ^
465: /home/ssh_office/workspace/oe-ms2/build/output/include/openenclave/openssl/unsupported.h:20:52: note: expanded from macro 'OE_OPENSSL_UNSUPPORTED'
465: #define OE_OPENSSL_UNSUPPORTED(MSG) __attribute__((deprecated(MSG)))
465:                                                    ^
465: 1 warning and 1 error generated.
```

docs/OpenSSLSupport.md
```diff
@@ -148,6 +148,31 @@ Note that the code snippet for the RDRAND engine opt-in is required to use not o
...
+SSL_CTX_set_default_verify_paths | ssl.h | The API specifies the locations as part of arguments from which an application looks up CA certificates for verification purposes. The API would allow an untrusted host to fully control what certificates the application will trust. | Same recommendation as SSL_CTX_load_verify_locations. |
+SSL_CTX_set_default_verify_dir | ssl.h | Similar to SSL_CTX_set_default_verify_paths except for specifying dir only. | Same recommendation as SSL_CTX_load_verify_locations. |
+SSL_CTX_set_default_verify_file | ssl.h | Similar to SSL_CTX_set_default_verify_paths except for specifying file only. | Same recommendation as SSL_CTX_load_verify_locations. |
+X509_LOOKUP_hash_dir | x509_vfy.h | This API returns a `X509_LOOKUP_METHOD` method that loads files from the path specified by the `SSL_CERT_DIR` environment variable. The API is used by SSL_CTX_set_default_verify_dir and SSL_CTX_set_default_verify_paths internally. | The recommendation is implementing a customized `X509_LOOKUP_METHOD` method based on in-memory operations. |
```
CodeMonkeyLeet, line 168: >> Might be useful to clarify that the X509_LOOKUP methods are not blocked because they are used SSL_CTX_set_default_verify_* methods; they're blocked because they provide analogous functionality to X509_load_*_file methods.
>
>I added a recommendation to implement a customized X509_LOOPUP_METHOD. Will that be sufficient?

My previous comment here was with regards to the Comment column, not the newly added Guidance column; i.e. the Comments should explain why the API is blocked (it provides way to load trust certificates and CRLs from untrusted host) rather than functional details (it is used in the SSL_CTX_* methods). 

Regarding the guidance, I think that's fine, though I would suggest adding a bit more context to the recommendation. For example:
```md
This method is used internally by `SSL_CTX_set_default_verify_dir` and `SSL_CTX_set_default_verify_paths`.
You can define your own enclave-specific [`X509_LOOKUP_METHOD`](
https://www.openssl.org/docs/man1.1.1/man3/X509_LOOKUP_meth_new.html)
but it will not be used by the SSL_CTX methods transparently. We recommend
explicitly setting certificate trust store witch `SSL_CTX_set_cert_store` instead.
```

docs/OpenSSLSupport.md
```diff
@@ -148,6 +148,31 @@ Note that the code snippet for the RDRAND engine opt-in is required to use not o
...
+X509_LOOKUP_file | x509_vfy.h | This API returns a `X509_LOOKUP_METHOD` method that loads files from the path specified by the `SSL_CERT_FILE` environment variable. The API is used by SSL_CTX_set_default_verify_file and SSL_CTX_set_default_verify_paths internally. | The recommendation is not to use this API (i.e., the lookup method associated with it). | Same recommendation as X509_LOOKUP_hash_dir. |
```
CodeMonkeyLeet, line 169: This row has an extra column, so the Guidance ends up displaying as

> The recommendation is not to use this API (i.e., the lookup method associated with it).

Instead of the intended:

> Same recommendation as X509_LOOKUP_hash_dir.

Also, for this and `X509_STORE_set_default_paths`, I would suggest just copy-pasting the guidance rather than having the reader go hunt in the table for the reference.

3rdparty/openssl/unsupported.h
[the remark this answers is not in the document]
anakrish: I like the user experience above. It is an error that the user cannot ignore.
It also prints out the link to the security guidance documentation.

docs/OpenSSLSupport.md
mingweishih: Thanks for the catch!

docs/OpenSSLSupport.md
mingweishih: Updated.

mingweishih: Note: the cross reference does not seem to work in the table.

CodeMonkeyLeet (approved):
LGTM, thanks for all the hard work you've put into this!

PR #3412, opened 2020-08-19
Update SIG Chairs

radhikaj:
Signed-off-by: Radhika Jandhyala <radhikaj@microsoft.com>

Updated owners

Signed-off-by: Radhika Jandhyala <radhikaj@microsoft.com>

PR #3493, opened 2020-09-04
README files for SIGs

dthaler:
A few SIGs already had README files and others did not.  This PR makes all SIGs have consistent README files.

PR #3412 updated the OWNERS files with chairs, but did not make corresponding changes to the README files which still said "TBD". This PR addresses that inconsistency.

Names simply match those already in the OWNERS files in the same directories. Any changes to names should be in a separate PR, since those are up to SIG chairs.

Any other content changes to the README files would be up to the relevant SIGs.

### DIFF
--- a/docs/Community/sig-website/README.md
+++ b/docs/Community/sig-website/README.md
@@ -1,0 +1,30 @@
+Website Special Interest Group
+=============================
+
+SIG-Website is responsible for ...
+
+Meetings
+--------
+
+TBD
+
+SIG Chair
+---------
+
+Radhika Jandhyala ([radhikaj](https://github.com/radhikaj))
+
+SIG Approvers
+-------------
+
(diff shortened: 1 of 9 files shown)


radhikaj (commented):
Thanks @dthaler

yentsanglee (approved):

docs/Community/sig-website/README.md
```diff
@@ -0,0 +1,30 @@
+Website Special Interest Group
+=============================
+
+SIG-Website is responsible for ...
```
achamayou, line 4: "updating and maintaining https://openenclave.io" presumably?

achamayou (approved):

docs/Community/sig-website/README.md
dthaler: I would presume so yes, but I'd prefer that the sig lead fill that in, in a PR, rather than me doing so.

PR #3499, opened 2020-09-07
Add proper EEID endorsements and move SGX endorsements there.

wintersteiger:
The EEID attester plugin incorrectly kept the underlying SGX endorsements in its evidence. This PR moves them into the right place and adds a test that illustrates how to use the EEID endorsements with and without the SGX endorsements.

### DIFF
--- a/common/sgx/eeid_verifier.c
+++ b/common/sgx/eeid_verifier.c
@@ -15,6 +15,7 @@
 
 #include <openenclave/attestation/sgx/eeid_plugin.h>
 #include <openenclave/attestation/sgx/eeid_verifier.h>
+#include <openenclave/attestation/sgx/evidence.h>
 #include <openenclave/bits/attestation.h>
 #include <openenclave/bits/eeid.h>
 #include <openenclave/bits/evidence.h>
@@ -29,6 +30,8 @@
 #include "../attest_plugin.h"
 #include "../common.h"
 #include "quote.h"
+
+static const oe_uuid_t _sgx_uuid = {OE_FORMAT_UUID_LEGACY_REPORT_REMOTE};
 
 static oe_result_t _eeid_verifier_on_register(
     oe_attestation_role_t* context,
@@ -47,54 +50,10 @@ static oe_result_t _eeid_verifier_on_unregister(oe_attestation_role_t* context)
     return OE_OK;
 }
 
-typedef struct _header
-{
-    uint32_t version;
-    oe_uuid_t format_id;
-    uint64_t data_size;
-    uint8_t data[];
-} header_t;
-
-static oe_result_t _add_claim(
-    oe_claim_t* claim,
-    void* name,
-    size_t name_size,
-    void* value,
-    size_t value_size)
-{
-    oe_result_t result = OE_UNEXPECTED;
-
-    if (*((uint8_t*)name + name_size - 1) != '\0')
-        return OE_CONSTRAINT_FAILED;
-
-    claim->name = (char*)oe_malloc(name_size);
-    if (claim->name == NULL)
-        return OE_OUT_OF_MEMORY;
-    OE_CHECK(oe_memcpy_s(claim->name, name_size, name, name_size));
-
-    claim->value = (uint8_t*)oe_malloc(value_size);
-    if (claim->value == NULL)
-    {
-        oe_free(claim->name);
-        claim->name = NULL;
-        return OE_OUT_OF_MEMORY;
-    }
-    OE_CHECK(oe_memcpy_s(claim->value, value_size, value, value_size));
-    claim->value_size = value_size;
-
-    result = OE_OK;
-done:
-    return result;
-}
-
 static oe_result_t _add_claims(
     oe_verifier_t* context,
-    const uint8_t* r_enclave_hash,
-    const uint8_t* r_signer_id,
-    uint16_t r_product_id,
-    uint32_t r_security_version,
-    uint64_t r_attributes,
-    uint32_t r_id_version,
+    const oe_claim_t* sgx_claims,
+    size_t sgx_claims_length,
     const uint8_t* r_enclave_base_hash,
     uint8_t* eeid_data,
     size_t eeid_data_size,
@@ -104,7 +63,7 @@ static oe_result_t _add_claims(
     oe_result_t result = OE_UNEXPECTED;
     size_t claims_index = 0;
     oe_claim_t* claims = NULL;
-    size_t num_claims = OE_REQUIRED_CLAIMS_COUNT + 1;
+    size_t num_claims = sgx_claims_length + 2;
 
     if (!claims_out || !claims_size_out)
         OE_RAISE(OE_INVALID_PARAMETER);
@@ -116,56 +75,28 @@ static oe_result_t _add_claims(
     if (claims == NULL)
         return OE_OUT_OF_MEMORY;
 
-    OE_CHECK(_add_claim(
-        &claims[claims_index++],
-        OE_CLAIM_ID_VERSION,
-        sizeof(OE_CLAIM_ID_VERSION),
-        &r_id_version,
-        sizeof(r_id_version)));
+    for (size_t i = 0; i < sgx_claims_length; i++)
+    {
+        const oe_claim_t* sgx_claim = &sgx_claims[i];
+        if (strcmp(sgx_claim->name, OE_CLAIM_FORMAT_UUID) != 0)
+        {
+            OE_CHECK(oe_sgx_add_claim(
+                &claims[claims_index++],
+                sgx_claim->name,
+                strlen(sgx_claim->name) + 1,
+                sgx_claim->value,
+                sgx_claim->value_size));
+        }
+    }
 
-    OE_CHECK(_add_claim(
-        &claims[claims_index++],
-        OE_CLAIM_SECURITY_VERSION,
-        sizeof(OE_CLAIM_SECURITY_VERSION),
-        &r_security_version,
-        sizeof(r_security_version)));
-
-    OE_CHECK(_add_claim(
-        &claims[claims_index++],
-        OE_CLAIM_ATTRIBUTES,
-        sizeof(OE_CLAIM_ATTRIBUTES),
-        &r_attributes,
-        sizeof(r_attributes)));
-
-    OE_CHECK(_add_claim(
-        &claims[claims_index++],
-        OE_CLAIM_UNIQUE_ID,
-        sizeof(OE_CLAIM_UNIQUE_ID),
-        (void*)r_enclave_hash,
-        OE_UNIQUE_ID_SIZE));
-
-    OE_CHECK(_add_claim(
-        &claims[claims_index++],
-        OE_CLAIM_SIGNER_ID,
-        sizeof(OE_CLAIM_SIGNER_ID),
-        (void*)r_signer_id,
-        OE_SIGNER_ID_SIZE));
-
-    OE_CHECK(_add_claim(
-        &claims[claims_index++],
-        OE_CLAIM_PRODUCT_ID,
-        sizeof(OE_CLAIM_PRODUCT_ID),
-        &r_product_id,
-        OE_PRODUCT_ID_SIZE));
-
-    OE_CHECK(_add_claim(
+    OE_CHECK(oe_sgx_add_claim(
         &claims[claims_index++],
         OE_CLAIM_FORMAT_UUID,
         sizeof(OE_CLAIM_FORMAT_UUID),
         &context->base.format_id,
         sizeof(oe_uuid_t)));
 
-    OE_CHECK(_add_claim(
+    OE_CHECK(oe_sgx_add_claim(
         &claims[claims_index++],
         OE_CLAIM_EEID_BASE_ID,
         sizeof(OE_CLAIM_EEID_BASE_ID),
@@ -174,7 +105,7 @@ static oe_result_t _add_claims(
 
     if (eeid_data && eeid_data_size)
     {
-        OE_CHECK(_add_claim(
+        OE_CHECK(oe_sgx_add_claim(
             &claims[claims_index++],
             OE_CLAIM_EEID_DATA,
             sizeof(OE_CLAIM_EEID_DATA),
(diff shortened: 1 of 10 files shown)


common/sgx/eeid_verifier.c
```diff
@@ -199,17 +207,29 @@ static oe_result_t _verify_sgx_report(
         }
     }
 
+    sgx_claims_size = sgx_evidence_buffer_size -
+                      (header->report_size + sizeof(oe_report_header_t));
+    *sgx_claims = oe_malloc(sgx_claims_size);
```
dthaler, line 212: need to check for allocation failure.   oe_sgx_extract_claims will crash if you pass it a NULL pointer.

In addition, the contract seems odd if we return a buffer to the caller to free even when we return failure from this function (e.g., if the checks fail at 216, 225, etc.), as it's a recipe for memory leak bugs.   I'd recommend changing this function to only return the buffer to the caller when returning OE_OK.

common/sgx/eeid_verifier.c
```diff
@@ -275,18 +298,45 @@ static oe_result_t _eeid_verify_evidence(
...
+    if (endorsements_buffer)
     {
-        if ((sgx_endorsements_buffer = oe_memalign(
-                 2 * sizeof(void*), sgx_endorsements_buffer_size)) == 0)
-            OE_RAISE(OE_OUT_OF_MEMORY);
-        memcpy(
-            sgx_endorsements_buffer,
-            evidence->data + evidence->sgx_evidence_size,
-            sgx_endorsements_buffer_size);
+        endorsements = oe_malloc(endorsements_buffer_size);
```
dthaler, line 310: check for allocation failure

include/openenclave/internal/eeid.h
```diff
@@ -243,6 +242,40 @@ oe_result_t oe_eeid_evidence_ntoh(
...
+ */
+oe_result_t oe_eeid_endorsements_hton(
+    const oe_eeid_endorsements_t* endorsements,
+    uint8_t* buffer,
+    size_t buffer_size);
+
+/**
+ * Read an oe_eeid_endorsements_t from a buffer using host byte-order.
+ *
+ * @param[in] buffer The buffer to read from
```
dthaler, line 265: All your other lines end in a period

tests/eeid_plugin/host/host.c
```diff
@@ -220,6 +222,26 @@ void start_enclave(const char* filename, uint32_t flags, enclave_stuff_t* stuff)
             &stuff->claims,
             &stuff->claims_length) == OE_OK);
 
+    /* Verify evidence with EEID endorsements, but without SGX endorsements */
+    size_t half_endorsements_size =
+        sizeof(oe_eeid_endorsements_t) + stuff->eeid->data_size;
+    oe_eeid_endorsements_t* half_endorsements = malloc(half_endorsements_size);
```
dthaler, line 228: check for out of memory

common/sgx/eeid_verifier.c
wintersteiger: Thanks for going through this and sorry for the excessive noise! Looks like a had a bad day when I wrote this, there were quite a few problems with that code.

wintersteiger:
bors try

common/sgx/eeid_verifier.c
```diff
@@ -317,36 +379,23 @@ static oe_result_t _eeid_verify_evidence(
             &sgx_claims_length,
             &parsed_report));
 
+        oe_free_claims(sgx_claims, sgx_claims_length);
```
yentsanglee, line 382: If you don't care about sgx_claims output, you may pass in NULL. `_verify_sgx_report()` may ignore extracting SGX claims if the caller is not interested it.

radhikaj:
@bodzhang Can you please review this?

common/sgx/eeid_verifier.c
wintersteiger: I thought so too, but that's not the case. `oe_sgx_extract_claims` assumes that `claims_out` is non-NULL and (my) `_verify_sgx_report` doesn't free its local claims, which would result in a memory leak. I'll move the `oe_free_claims` there. I'm happy to fix `oe_sgx_extract_claims` though, if that's welcome. I found it odd that in this particular case we're not allowed to pass NULL.

shnwc: I think that `oe_sgx_extract_claims()` should always expect a non-NULL `claims_out` parameter, since the purpose of this function is to extract claims. But `_verify_sgx_report()` should check its `sgx_claims` parameter, and skip calling `oe_sgx_extract_claims()` if this parameter is NULL.

Similarly, `_verify_evidence()` in `common/sgx/verifier.c` should check its `claims` parameter, and skip calling `oe_sgx_extract_claims()` if this parameter is NULL.

On the other hand, in `include/openenclave/attestation/verifier.h`, the `claims` parameter in API `oe_verify_evidence()` is not commented as optional, but the implementation of this API in `common/attest_plugin.c` allows this parameter to be NULL.

So looks like there is some clean up in comments and implementation. How about taking care of this in a separate issue / PR? @yentsanglee @wintersteiger

shnwc (commented):
`_verify_sgx_report()` should check its `sgx_claims` parameter, and skip calling `oe_sgx_extract_claims()` if this parameter is NULL..

common/sgx/eeid_verifier.c
wintersteiger: That's a very good point indeed! I just copied the structure from the SGX plugin, but I ended up using only the parsed report. I should really change and get the information from the claims instead, which fits in much better with the plugin infrastructure.

wintersteiger: (Done)

common/sgx/eeid_verifier.c
```diff
@@ -199,33 +202,144 @@ static oe_result_t _verify_sgx_report(
...
         header->report + header->report_size,
-        sgx_claims_size,
+        expected_sgx_claims_length,
         &sgx_endorsements,
-        sgx_claims,
-        sgx_claims_length);
+        &local_sgx_claims,
+        &local_sgx_claims_length));
+
+    *sgx_claims = local_sgx_claims;
```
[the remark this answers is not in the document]
wintersteiger: Sure, added. 

This is another instance where my copy of the SGX verification flow slightly differs and generally lags behind the changes in the SGX plugins. I really wanted to use the actual SGX plugin, but I was told that nobody ever considered the idea of plugins using other plugins, so I had to copy a bunch of code. I think we should reconsider this at some point; some users may want to write custom plugins for multi-platform/-tee attestation for which I'm sure they'll want to use the existing plugins.

shnwc:
bors try

wintersteiger:
bors try

include/openenclave/bits/eeid.h
```diff
@@ -59,6 +59,21 @@ typedef struct _oe_eeid
...
+OE_PACK_BEGIN
+/**
+ * Structure to keep EEID endorsements.
+ */
+typedef struct
+{
+    /** Size of the underlying SGX endorsements. */
+    size_t sgx_endorsements_size;
+    /** Size of EEID endorsements (EEID data). */
+    size_t eeid_endorsements_size;
```
bodzhang, line 71: My understanding is that the "eeid_endorsement" is the eeid configuration data. Comparing against the "customer claims", which can be considered "run time" claims (that can be produced by the any enclave code, including code downloaded into the enclave post enclave initialization), for example, a public key to be used to wrap secret to provision to the enclave, the eeid configuration data can be considered as "init time" claim that captures the identity of the code to be downloaded into the enclave post enclave initialization. The eeid configuration data seems to be fit better in the evidence buffer, together with the "customer claims", than in the endorsement buffer, which typically contains signature/cert-chains.

wintersteiger: There's already a copy in the evidence ([here](https://github.com/wintersteiger/openenclave/blob/ac364a2cbb60bbc4fb5b378dc8d11592652ab095/include/openenclave/internal/eeid.h#L204)). The second copy in the endorsements is just for convenience; the verifier will check equality with the copy in the evidence. This means that users can replace all or parts of the endorsements with their own, which we clearly need for SGX endorsements, so we might as well put the EEID data into the endorsements too, but users can replace/remove it if they don't trust them. It doesn't make a big difference though, we can remove it if you prefer.

wintersteiger:
bors try

wintersteiger:
bors try

wintersteiger:
bors try

wintersteiger:
bors try

wintersteiger:
bors try

wintersteiger:
bors try

common/sgx/eeid_verifier.c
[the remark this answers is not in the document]
shnwc: @wintersteiger tough there has not been been a design for plugins invoking other plugins, the EEID plugin code already are in the same directories as the regular SGX plugins (`common/sgx` and `enclave/sgx`), and call multiple global functions implemented for the regular SGX plugins. If you see that some other code can be shared, please let us know and we can see how those code can be made sharable.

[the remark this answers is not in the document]
wintersteiger: Thanks, that's a great observation! I'll have another look at it and see whether I can reuse more of that.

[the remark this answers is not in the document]
wintersteiger: Thanks again for that suggestion, by exporting a couple functions from the SGX plugin I was able to cut the code size of the EEID verifier plugin down by quite a bit! I added the `sgx_attestation_plugin_` prefix to those functions, but I'm happy to rename them if we should use a different pattern for those.

wintersteiger:
bors try

common/sgx/eeid_verifier.c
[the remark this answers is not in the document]
shnwc: @wintersteiger glad you were able to reuse more code. For function renaming, I think `oe_sgx_add_claim()`, `oe_sgx_verify_evidence()`, and `oe_sgx_free_claims_list()` are better aligned with the names of other global functions implemented in the `common/sgx/` directory. When possible, I'd prefer shorter names.

BRMcLaren:
There was a small fire in the CI department, re-running.

BRMcLaren:
bors try

wintersteiger:
bors try

wintersteiger:
bors try

common/sgx/eeid_verifier.c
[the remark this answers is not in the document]
wintersteiger: Thanks, done!

wintersteiger:
bors try

wintersteiger:
@dthaler @bodzhang @yentsanglee @shnwc I think there are no unresolved issues with this PR anymore, could I get you to approve it so it can get merged? Thanks!

include/openenclave/bits/eeid.h
bodzhang: OK to have two copies, as long as the Verifier plugin also output relevant claims based on the verified copy.

common/sgx/verifier.c
```diff
@@ -69,7 +69,7 @@ static void _free_claim(oe_claim_t* claim)
     oe_free(claim->value);
 }
 
-oe_result_t sgx_attestation_plugin_free_claims_list(
+oe_result_t oe_sgx_free_claims_list(
```
bodzhang, line 72: I'm not 100% sure about the naming convention, but my assumption is that "oe_sgx_..." is typically used for OE SDK core libraries for SGX. Here the function seems to be attestation plugin specific. Isn't the old name better?

common/sgx/verifier.c
```diff
@@ -185,7 +185,7 @@ static oe_result_t _process_sgx_report_data(
     return result;
 }
 
-static oe_result_t _add_claim(
+oe_result_t oe_sgx_add_claim(
```
bodzhang, line 188: Need to check the naming convention. Maybe it should be "sgx_attestation_plugin_add_Claim"

common/sgx/verifier.c
```diff
@@ -543,11 +543,11 @@ oe_result_t oe_sgx_extract_claims(
 
 done:
     if (claims)
-        sgx_attestation_plugin_free_claims_list(NULL, claims, claims_length);
+        oe_sgx_free_claims_list(NULL, claims, claims_length);
     return result;
 }
 
-static oe_result_t _verify_evidence(
+oe_result_t os_sgx_verify_evidence(
```
bodzhang, line 550: Typo? Is the intention "oe_sgx_verify_evidence"? Need to check naming convention. Maybe it should be "sgx_attestation_plugin_verify_evidence".

include/openenclave/internal/eeid.h
```diff
@@ -28,6 +28,19 @@ static const uint8_t OE_DEBUG_PUBLIC_KEY[] = {
...
+{
+    uint8_t* enclave_hash;
+    size_t enclave_hash_size;
+    uint8_t* signer_id;
+    size_t signer_id_size;
+    uint16_t product_id;
+    uint32_t security_version;
+    uint64_t attributes;
+    uint32_t id_version;
+} oe_eeid_relevant_sgx_claims_t;
```
dthaler, line 42: This header file is currently not in an sgx directory or having sgx in the filename, but this type has sgx in the name.  Can we keep sgx types in a file either with sgx in the name or under an sgx subdir?

include/openenclave/internal/eeid.h
```diff
@@ -136,11 +136,7 @@ oe_result_t oe_remeasure_memory_pages(
  *
  */
 oe_result_t verify_eeid(
-    const uint8_t* reported_enclave_hash,
-    const uint8_t* reported_enclave_signer,
-    uint16_t reported_product_id,
-    uint32_t reported_security_version,
-    uint64_t reported_attributes,
+    const oe_eeid_relevant_sgx_claims_t* relevant_claims,
```
dthaler, line 139: verify_eeid is a non-sgx specific name so if you use an sgx specific type, the function should be renamed in line 138 to put sgx in the name.  (And it should be in an sgx specific header file)

include/openenclave/internal/eeid.h
```diff
@@ -202,10 +198,9 @@ oe_result_t oe_eeid_ntoh(
...
 typedef struct
 {
-    size_t sgx_evidence_size;     /* Size of base-image evidence */
-    size_t sgx_endorsements_size; /* Size of base-image endorsements */
-    size_t eeid_size;             /* Size of EEID */
-    uint8_t data[];               /* Data (same order as the sizes) */
+    size_t sgx_evidence_size; /* Size of SGX evidence */
+    size_t eeid_size;         /* Size of EEID */
+    uint8_t data[];           /* Data (same order as the sizes) */
 } oe_eeid_evidence_t;
```
dthaler, line 204: looks like this name was already bad... sgx fields in a non-sgx struct.   give that this is in an internal header file, we should just fix it.

include/openenclave/internal/eeid.h
```diff
@@ -242,6 +237,40 @@ oe_result_t oe_eeid_evidence_ntoh(
...
+/**
+ * Convert an oe_eeid_endorsements_t into a buffer using network byte-order.
+ *
+ * @param[in] endorsements The oe_eeid_endorsements_t to convert.
+ *
+ * @param[in] buffer The buffer to write to.
+ *
+ * @param[in] buffer_size Size of **buffer**.
+ *
+ * @returns Returns OE_OK on success.
```
dthaler, line 249:
```suggestion
 * @retval OE_OK The operation was successful.
```
Same in 266 etc

include/openenclave/internal/sgx/plugin.h
```diff
@@ -89,11 +89,57 @@ oe_result_t oe_sgx_hash_custom_claims_buffer(
  * @param[in] claims_length The length of claims.
  * @retval OE_OK on success, otherwise an appropriate error code.
```
dthaler, line 90:
```suggestion
 * @retval OE_OK The operation was successful.
 * @retval other An appropriate error code.
```
Same in 107 etc.  Doxygen style with retval is that the description applies (just) to the first token after retval

common/sgx/verifier.c
wintersteiger: Yeah, that's a typo. I think the name is fine, but I'm also happy to rename this and other functions if we should apply a different convention.

include/openenclave/internal/eeid.h
wintersteiger: Fixed, the SGX in the name wasn't really necessary.

include/openenclave/internal/eeid.h
wintersteiger: Unnecessarily so; it's the base-image evidence size, Fixed.

wintersteiger:
bors try

include/openenclave/internal/eeid.h
```diff
@@ -161,7 +162,7 @@ oe_result_t oe_create_eeid_sgx(size_t data_size, oe_eeid_t** eeid);
  *
  * @param[in] eeid The oe_eeid_t
  *
- * @returns Returns OE_OK on success.
+ * @returns the size (in bytes) of the given  EEID structure.
```
dthaler, line 165:
```suggestion
 * @returns the size (in bytes) of the given EEID structure.
```

dthaler (approved):

yentsanglee (approved):

shnwc (approved):

wintersteiger:
bors try

yentsanglee:
If it's ready for merging, please squash all commits.

wintersteiger:
bors try

wintersteiger:
bors try

wintersteiger:
All done, rebased and squashed, ready for r+, thanks @yentsanglee!

PR #3503, opened 2020-09-08
Fix potential null pointer dereferences

### LINKED ISSUE #3456: Multiple potential null pointer derefs in hostcalls.c
There are two potential null pointer derefs.

Here with td:
https://github.com/openenclave/openenclave/blob/8e8c5cb91fdfef859bd20a8b996110cf5910616d/enclave/core/sgx/hostcalls.c#L13-L16

And here with ecall_context:
https://github.com/openenclave/openenclave/blob/8e8c5cb91fdfef859bd20a8b996110cf5910616d/enclave/core/sgx/hostcalls.c#L66-L72

larrywxyz:
Fixes #3456

### DIFF
--- a/enclave/core/sgx/hostcalls.c
+++ b/enclave/core/sgx/hostcalls.c
@@ -69,7 +69,7 @@ void oe_free_ocall_buffer(void* buffer)
 
     // ecall context's buffer is managed by the host and does not have to be
     // freed.
-    if (buffer == ecall_context->ocall_buffer)
+    if (ecall_context && buffer == ecall_context->ocall_buffer)
         return;
 
     // Even though ecall_context is memory controlled by the host, there


enclave/core/sgx/hostcalls.c
```diff
@@ -13,10 +13,14 @@
...
 {
     oe_sgx_td_t* td = oe_sgx_get_td();
-    oe_ecall_context_t* ecall_context = td->host_ecall_context;
-    return oe_is_outside_enclave(ecall_context, sizeof(*ecall_context))
-               ? ecall_context
-               : NULL;
+    if (td)
+    {
+        oe_ecall_context_t* ecall_context = td->host_ecall_context;
+        return oe_is_outside_enclave(ecall_context, sizeof(*ecall_context))
```
radhikaj, line 19: Add check to see if ecall_context is null as well.

dthaler: Should not be needed, oe_is_outside_enclave() already explicitly returns false if you pass in a null pointer

dthaler (approved):

radhikaj (approved):

enclave/core/sgx/hostcalls.c
```diff
@@ -13,10 +13,14 @@
 static oe_ecall_context_t* _get_ecall_context()
 {
     oe_sgx_td_t* td = oe_sgx_get_td();
-    oe_ecall_context_t* ecall_context = td->host_ecall_context;
-    return oe_is_outside_enclave(ecall_context, sizeof(*ecall_context))
-               ? ecall_context
-               : NULL;
+    if (td)
```
anakrish, line 16: There will always be `td`. It is `FS[0]`: 
https://github.com/openenclave/openenclave/blob/5d0ce69f19053a42c5feffec42eb6e3920bf1f00/enclave/core/sgx/td.c#L167-L174

If there is some kind of memory overwrite bug within the enclave, then there is a chance that FS[0] is NULL. In that case, all uses of oe_sgx_get_td ought to be guarded; not just this one.
Our threat model does not consider scenarios where bugs within the enclave cause core data structures to be overwritten. In such scenarios, oe_is_within_enclave check would be safe than a not-null check.

enclave/core/sgx/hostcalls.c
```diff
@@ -69,7 +73,7 @@ void oe_free_ocall_buffer(void* buffer)
 
     // ecall context's buffer is managed by the host and does not have to be
     // freed.
-    if (buffer == ecall_context->ocall_buffer)
+    if (ecall_context && buffer == ecall_context->ocall_buffer)
```
anakrish, line 76: This check is good.

PR #3505, opened 2020-09-08
Disable ubuntu 16.04 gcc w/ snmalloc configurations on nightly builds

mingweishih:
snmalloc does not work with the default gcc version on ubuntu 16.04. Disable such configurations (which cause failures) on nightly builds.


BRMcLaren (approved):

BRMcLaren:
CC @achamayou

achamayou:
@mingweishih @BRMcLaren my bad, I thought I'd run all the builds, but I clearly missed that one.

achamayou (approved):

PR #3510, opened 2020-09-09
Explicitly init OpenSSL to prevent the unsafe configuration

mingweishih:
This PR adds a constructor to the `oeenclave` that explicitly invokes the OpenSSL `OPENSSL_init_crypto` API with the `OE_OPENSSL_INIT_NO_LOAD_CONFIG` option when an enclave links against OpenSSL libraries. Doing so effectively prevents users to initialize an OpenSSL program with the `OE_OPENSSL_INIT_LOAD_CONFIG`, which would load the `openssl.cnf` from untrusted host filesystem (with hard-coded path determined at compile-time). Such behavior is not recommended by OE based on the threat model and the lack of secure FS support.

Note that this PR does not break the user experiences of using OpenSSL APIs as `OPENSSL_init_crypto` is allowed to be invoked multiple times.

### DIFF
--- a/enclave/sgx/openssl_init.c
+++ b/enclave/sgx/openssl_init.c
@@ -1,0 +1,47 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/bits/types.h>
+
+/* Refer to openssl/crypto.h for the original definitions */
+#define OE_OPENSSL_INIT_NO_LOAD_CONFIG 0x00000080L
+
+/* Forward declarion */
+typedef struct oe_openssl_init_settings OE_OPENSSL_INIT_SETTINGS;
+int OE_OPENSSL_init_crypto(
+    uint64_t opts,
+    const OE_OPENSSL_INIT_SETTINGS* settings);
+
+/*
+ * Weak implementation that prevents link-time errors when an enclave
+ * does not link against OpenSSL libraries.
+ */
+int OE_OPENSSL_init_crypto(
+    uint64_t opts,
+    const OE_OPENSSL_INIT_SETTINGS* settings)
+{
+    OE_UNUSED(opts);
+    OE_UNUSED(settings);
+    return 0;
+}
+OE_WEAK_ALIAS(OE_OPENSSL_init_crypto, OPENSSL_init_crypto);
+
+/*
+ * Define the function as a constructor that ensures it is invoked before
+ * any OpenSSL APIs. Note that the constructor is only included by the linker
+ * when an enclave links against OpenSSL libraries (i.e., the
+ * OPENSSL_init_crypto symbol is included in the binary); i.e., when the enclave
+ * does not link against OpenSSL libraries, the constructor is discarded by the
+ * linker.
+ */
+__attribute__((constructor)) void oe_openssl_init()
+{
+    /*
+     * Invoke the API with the OPENSSL_INIT_NO_LOAD_CONFIG option effectively
+     * nullifies the OPENSSL_INIT_LOAD_CONFIG option. This prevents users to
+     * explicitly configure an OpenSSL application to load openssl.cnf from
+     * the host filesystem, which is considered as untrusted under the threat
+     * model of TEE.
+     */
+    OPENSSL_init_crypto(OE_OPENSSL_INIT_NO_LOAD_CONFIG, NULL);
+}
--- a/enclave/CMakeLists.txt
+++ b/enclave/CMakeLists.txt
@@ -20,6 +20,7 @@ if (OE_SGX)
       sgx/attester.c
       sgx/report.c
       sgx/collateralinfo.c
+      sgx/openssl_init.c
       sgx/start.S)
   if (WITH_EEID)
     list(APPEND PLATFORM_SRC ../common/sgx/eeid_verifier.c ../common/sgx/eeid.c


mingweishih:
bors try

mingweishih:
@dthaler  This is another security recommendation, but with less impact to APIs (only an OpenSSL option will become invalid). However, we can have both this PR and #3492 for sig-review.

enclave/sgx/openssl_init.c
```diff
@@ -0,0 +1,47 @@
...
+__attribute__((constructor)) void oe_openssl_init()
+{
+    /*
+     * Invoke the API with the OPENSSL_INIT_NO_LOAD_CONFIG option effectively
+     * nullifies the OPENSSL_INIT_LOAD_CONFIG option. This prevents users to
+     * explicitly configure an OpenSSL application to load openssl.cnf from
+     * the host filesystem, which is considered as untrusted under the threat
+     * model of TEE.
+     */
+    OPENSSL_init_crypto(OE_OPENSSL_INIT_NO_LOAD_CONFIG, NULL);
```
CodeMonkeyLeet, line 46: Should this be invoking `OPENSSL_init_ssl` instead as that seems to be a superset of `OPENSSL_init_crypto`?

CodeMonkeyLeet, line 46: If we are going to force initialization, should we also include `OPENSSL_INIT_ENGINE_RDRAND` here?

CodeMonkeyLeet (commented):
Reading through the documentation for [`OPENSSL_init_crypto`](https://www.openssl.org/docs/man1.1.1/man3/OPENSSL_init_crypto.html), I do have some questions as to whether we should do this:

- The concern is that we want to block a user from explicitly specifying `OPENSSL_INIT_LOAD_CONFIG`, which is unsafe from untrusted host in SGX.
- We already invert the default in `OPENSSL_init_ssl` with the `OPENSSL_NO_AUTOLOAD_CONFIG` build option, so this change is an extra layer to block them from manually overriding it via `OPENSSL_init_crypto/ssl`. 
- I'm not sure if that extra level of protection is warranted, having read the docs, as it seems to have the undesirable side effect of preventing a dev from using it for setting other non-default values (e.g. enabling engines, or `OPENSSL_INIT_NO_ADD_ALL_CIPHERS` so that they can customize available ciphers).
- Given the approach in #3492, perhaps its better to provide a default header that flags the use of `OE_OPENSSL_INIT_LOAD_CONFIG` (which would cover both `OPENSSL_init_crypto` & `OPENSSL_init_ssl`) than it is to try and squat on the method invocation.

mingweishih:
> * `OPENSSL_INIT_NO_ADD_ALL_CIPHERS`

For clarification:

`OPENSSL_INIT_NO_LOAD_CONFIG` / `OPENSSL_INIT_LOAD_CONFIG` options are used by `OPENSSL_init_crypto`. The function ensures these two options are mutual exclusive. See

https://github.com/openssl/openssl/blob/526cf60408e1a356ec712b6c88a88864fdbe73af/crypto/init.c#L694

and

https://github.com/openssl/openssl/blob/526cf60408e1a356ec712b6c88a88864fdbe73af/crypto/init.c#L701

If passing the option to `OPENSSL_init_ssl`, it will forward the option to `OPENSSL_init_crypto` (https://github.com/openssl/openssl/blob/526cf60408e1a356ec712b6c88a88864fdbe73af/ssl/ssl_init.c#L205). 

If neither of these options are used, users can invoke the `OPENSSL_init_crypto`/`OPENSSL_ssl_crypto` with an option. This is the case with the `OPENSSL_NO_AUTOLOAD_CONFIG` build option, which only prevents the `OPENSSL_init_ssl` automatically sets the `OPENSSL_INIT_LOAD_CONFIG` before forwarding the settings to `OPENSSL_init_crypto` (https://github.com/openssl/openssl/blob/526cf60408e1a356ec712b6c88a88864fdbe73af/ssl/ssl_init.c#L202). See the following code snippet:

```
// This still work even if the OPENSSL_INIT_LOAD_CONFIG build option is set.
OPENSSL_init_crypto(OE_OPENSSL_INIT_LOAD_CONFIG, NULL);
```

```
// Ensure the subsequent invocation of OPENSSL_init_crypto with OE_OPENSSL_INIT_LOAD_CONFIG won't work
OPENSSL_init_crypto(OE_OPENSSL_INIT_NO_LOAD_CONFIG, NULL);
// No effects. But the invocation does not fail as well.
OPENSSL_init_crypto(OE_OPENSSL_INIT_LOAD_CONFIG, NULL);
```

Most similar options (e.g., `OPENSSL_init_crypto` / `OPENSSL_INIT_ADD_ALL_CIPHERS`) are guarded with the similar mechanisms and can be set with separated `OPENSSL_init_crypto` invocations.

enclave/sgx/openssl_init.c
```diff
@@ -0,0 +1,47 @@
...
+ */
+int OE_OPENSSL_init_crypto(
+    uint64_t opts,
+    const OE_OPENSSL_INIT_SETTINGS* settings)
+{
+    OE_UNUSED(opts);
+    OE_UNUSED(settings);
+    return 0;
+}
+OE_WEAK_ALIAS(OE_OPENSSL_init_crypto, OPENSSL_init_crypto);
```
anakrish, line 27: Consider renaming the function so that if an enclave does not use openssl,  then `nm -l enclave | grep openssl` return empty.

enclave/sgx/openssl_init.c
```diff
@@ -0,0 +1,47 @@
...
+
+/*
+ * Define the function as a constructor that ensures it is invoked before
+ * any OpenSSL APIs. Note that the constructor is only included by the linker
+ * when an enclave links against OpenSSL libraries (i.e., the
+ * OPENSSL_init_crypto symbol is included in the binary); i.e., when the enclave
+ * does not link against OpenSSL libraries, the constructor is discarded by the
+ * linker.
+ */
+__attribute__((constructor)) void oe_openssl_init()
```
anakrish, line 37: I don't think a constructor is the best solution here:
- Since no other compilation unit is calling any function defined in this file, the linker will not bring in any symbols from this file, including the constructor.
- Even if we force a dependency to this file, thereby retaining the constructor, the order of constructors is not predictable. We could work around it by explicitly specifying a priority for the constructor. But it just seems cleaner for oecore to call  `oe_crypto_init` (which could be a stub) at an appropriate point in initialization.

enclave/sgx/openssl_init.c
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/bits/types.h>
+
+/* Refer to openssl/crypto.h for the original definitions */
+#define OE_OPENSSL_INIT_NO_LOAD_CONFIG 0x00000080L
```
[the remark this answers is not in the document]
mingweishih: Yes, you are right. I define this under the assumption that the we don't ship OpenSSL headers today. But I realize that this file can be guarded by experimental flag and be moved out from the experimental altogether.

enclave/sgx/openssl_init.c
mingweishih: See my comments below.

mingweishih: We can do that but our current example of initializing RDRAND engine uses `ENGINE_load_rdrand()`, which effectively the same as that option.

https://github.com/openenclave/openenclave/blob/master/docs/OpenSSLSupport.md#how-to-use-rand-apis

enclave/sgx/openssl_init.c
mingweishih: For the current requirement, we don't need this constructor to be always linked; i.e., we only need the constructor when an enclave links against OpenSSL libraries. In this case, linker will always pick the constructor so we can ensure it runs before user's invocation.

We currently do need to enforce the execution order of this constructor and just need it to be present only an enclave links against OpenSSL. It does not depend on any constructor unless developer implements a constructor that uses OpenSSL APIs.

`oe_openssl_init` really indicates that it's specifically for OpenSSL, not Mbed. `oe_crypto_init` would be a bit confused given that we have `liboecrypto` already. However, the initialization for OpenSSL is needed even if the enclave is not using OpenSSL-based `liboecrypto`.

anakrish: >  we only need the constructor when an enclave links against OpenSSL libraries. In this case, linker will always pick the constructor so we can ensure it runs before user's invocation.

What causes the linker to pick up the constructor when linking against OpenSSL?

mingweishih: OpenSSL also does lazy initialization by calling `OPENSSL_init_crypto`/`OPENSSL_ssl_crypto`. Therefore, these symbols are always present even if an enclave linking against OpenSSL, uses any OpenSSL APIs, but not calls them explicitly.

mingweishih:
@anakrish @CodeMonkeyLeet If we want to simply deprecate the `OPENSSL_INIT_LOAD_CONFIG` option as part of #3492, what's the recommended approach that can catch the use of option during compile-time?

The `OPENSSL_INIT_LOAD_CONFIG` is defined as
```
# define OPENSSL_INIT_LOAD_CONFIG            0x00000040L
```

mingweishih:
The intention of this PR will be addressed as part of #3492. Closing this PR for now.

Issue #3516, opened 2020-09-10
time_t time(time_t*)


wintersteiger:
corelibc promises `time_t oe_time(time_t* tloc);` ([here](https://github.com/openenclave/openenclave/blob/03e07014e80d4894aee58d41216eeaa6d321a11d/include/openenclave/corelibc/time.h#L26)), but I can't find its implementation and there is no such symbol in the libraries. Is the implementation missing or should the prototype be removed from the headers?

radhikaj:
This can be removed, oe_time in the same file can also be removed.  
OE_INLINE
time_t time(time_t* tloc)
{
    return oe_time(tloc);
}
We should look through history and see if this was ever implemented.

anakrish:
Here is the callgraph for `time` function. It indicates that time function is being called by oecore (attestation) as well as be mbedtls. However, they seem to be using libc as opposed to corelibc (I'm not sure if that is recommended).

```
 time 0x6e150 /home/anakrish/work/fork2/3rdparty/musl/musl/src/time/time.c:4
  ├── oe_datetime_now 0x55e60 /home/anakrish/work/fork2/common/datetime.c:228
  │   └── oe_create_sgx_endorsements 0x5d2c0 /home/anakrish/work/fork2/common/sgx/endorsements.c:23
  │       └── oe_get_sgx_endorsements 0x5c0b0 /home/anakrish/work/fork2/common/sgx/endorsements.c:329
  │           ├── oe_verify_sgx_quote 0x19810 /home/anakrish/work/fork2/common/sgx/quote.c:428 ...
  │           ├── _verify_evidence 0x4b850 /home/anakrish/work/fork2/common/sgx/verifier.c:546 ...
  │           └── _get_evidence 0x4f610 /home/anakrish/work/fork2/enclave/sgx/attester.c:95 ...
  └── x509_get_current_time 0x91480 /home/anakrish/work/fork2/3rdparty/mbedtls/mbedtls/library/x509.c:904
      ├── mbedtls_x509_time_is_past 0x91370 /home/anakrish/work/fork2/3rdparty/mbedtls/mbedtls/library/x509.c:969
      │   ├── mbedtls_x509_crt_is_revoked 0x88420 /home/anakrish/work/fork2/3rdparty/mbedtls/mbedtls/library/x509_crt.c:1779
[...]

PR #3059, opened 2020-05-19
Implement new attestation API PR #2949 and quote-ex integration PR #2985

shnwc:
Implement new attestation API proposed in PR #2949, and the quote-ex library integration proposed in PR #2985, with the exception that the linking of the quote-ex library itself (which requies changes in cmake scripts) will be added in a separate PR.

PR #3517, opened 2020-09-10
Add gmtime.c to enclave/core

wintersteiger:
I needed `gmtime_r`, but `gmtime.c` was missing. I'm not sure whether it was intentionally excluded, but I think it should be included.

### DIFF
--- a/enclave/core/CMakeLists.txt
+++ b/enclave/core/CMakeLists.txt
@@ -161,6 +161,7 @@ add_enclave_library(
   backtrace.c
   ctype.c
   debugmalloc.c
+  gmtime.c
   hexdump.c
   hostcalls.c
   intstr.c


CodeMonkeyLeet:
The presence of gmtime.c and the include/openenclave/corelibc/time.h (see #3516) both look like regressions. PR #3057 removed these functions as they were not used by oecore, and they were reintroduced in PR #3059, likely as the result of a bad merge.

@wintersteiger oecore in general tries to avoid dealing with time (because currently pulling it from host is untrusted), higher level dependencies use the libc versions directly (mbedtls, attestation libs though there's an open question there as to whether they should) The transform functions are self-contained enough, but why do you need them?

@mingweishih & @anakrish Any objections to resurrecting these implementations?

mingweishih:
> The presence of gmtime.c and the include/openenclave/corelibc/time.h (see #3516) both look like regressions. PR #3057 removed these functions as they were not used by oecore, and they were reintroduced in PR #3059, likely as the result of a bad merge.
> 
> @wintersteiger oecore in general tries to avoid dealing with time (because currently pulling it from host is untrusted), higher level dependencies use the libc versions directly (mbedtls, attestation libs though there's an open question there as to whether they should) The transform functions are self-contained enough, but why do you need them?
> 
> @mingweishih & @anakrish Any objections to resurrecting these implementations?

Ah, thanks for the finding. That explains why #3516 looks familiar to me 🤣 
I think we should bring #3057 back. It's unclear to me why this is needed by oecore.

wintersteiger:
The endorsement extraction code uses `gmtime_r` ([here](https://github.com/openenclave/openenclave/blob/e506583006e4de5b9c7153b4b4ca069c4ecabeff/common/datetime.c#L238)) and because `MBEDTLS_HAVE_TIME_DATE` is enabled in the mbedTLS config ([here](https://github.com/openenclave/openenclave/blob/e506583006e4de5b9c7153b4b4ca069c4ecabeff/3rdparty/mbedtls/config.h#L160)), it assumes that there is an implementation of `mbedtls_platform_gmtime_r`, which I tried to provide via `oe_gmtime_r` (like [so](https://github.com/wintersteiger/sgx-lkl/blob/003e2b7737da2ba6b34b1069090d91ae2dd07c4a/src/enclave/enclave_oe.c#L339)).

To be honest, I think the certificate expiration checks are pretty much useless without a trusted source of time, we should just skip them in the enclave. I can successfully extract evidence/endorsements in SGX-LKL if after disabling `MBEDTLS_HAVE_TIME` and `MBEDTLS_HAVE_TIME_DATE` as well as the code block that records the endorsement creation time ([here](https://github.com/openenclave/openenclave/blob/7337126be9c2e8cb6725caeb26d387131501ed1a/common/sgx/endorsements.c#L74)). (I assume the verifier would fail if that date is 0, but that can be fixed too.)

mingweishih:
> The endorsement extraction code uses `gmtime_r` ([here](https://github.com/openenclave/openenclave/blob/e506583006e4de5b9c7153b4b4ca069c4ecabeff/common/datetime.c#L238)) and because `MBEDTLS_HAVE_TIME_DATE` is enabled in the mbedTLS config ([here](https://github.com/openenclave/openenclave/blob/e506583006e4de5b9c7153b4b4ca069c4ecabeff/3rdparty/mbedtls/config.h#L160)), it assumes that there is an implementation of `mbedtls_platform_gmtime_r`, which I tried to provide via `oe_gmtime_r` (like [so](https://github.com/wintersteiger/sgx-lkl/blob/003e2b7737da2ba6b34b1069090d91ae2dd07c4a/src/enclave/enclave_oe.c#L339)).
> 
> To be honest, I think the certificate expiration checks are pretty much useless without a trusted source of time, we should just skip them in the enclave. I can successfully extract evidence/endorsements in SGX-LKL after disabling `MBEDTLS_HAVE_TIME` and `MBEDTLS_HAVE_TIME_DATE` as well as the code block that records the endorsement creation time ([here](https://github.com/openenclave/openenclave/blob/7337126be9c2e8cb6725caeb26d387131501ed1a/common/sgx/endorsements.c#L74)). (I assume the verifier would fail if that date is 0, but that can be fixed too.)

Why not just using `gmtime_r` from libc which already includes it (https://github.com/openenclave/openenclave/blob/master/libc/CMakeLists.txt#L767) ?

wintersteiger:
SGX-LKL uses only selected parts of corelibc ([here](https://github.com/lsds/sgx-lkl/tree/oe_port/src/include/openenclave/corelibc)) to keep the dependencies as slim as possible and to avoid clashes with it's own libc (also musl).

dthaler:
> I think the certificate expiration checks are pretty much useless without a trusted source of time, we should just skip them in the enclave.

I agree they're pretty much useless without a trusted source of time.   And without certificate expiration checks, one could potentially argue that security is overall fairly weak.  The question is whether "pretty much useless" is any better than "useless".  I would probably argue it's better than nothing, since it can still defend against a remote attacker who hasn't compromised the host, and in that sense I am fine adding gmtime.c.

That said, I believe there is a strong need for a trusted source of absolute time (i.e., UTC time, to use in cert expiry checks and such).   I believe the best way to get that in the future would require two things:
1) A trusted source of relative time (e.g., time since boot).  On SGX I've been told that's only available via the Intel ME which deployments like Azure disable, AND
2) A Network Time Security for NTP client in an enclave, configured to talk to a highly trusted NTS/NTP server.

dthaler (approved):

wintersteiger:
Very good points @dthaler! A trusted source of time would be great, even if it is imprecise. A trusted NTP server may indeed be a practical solution that's not too much effort to implement. Although, I think that even a one-shot service that simply asks a trusted remote source during enclave initialization would suffice, especially for certificate expiry checks. Perhaps we should donate a couple enclave-enabled chips to the NPL or NIST? :-)

radhikaj (approved):

Issue #3523, opened 2020-09-10
[Bug] Windows PSW/DCAP Dependency Incompatibility with Windows Update on Windows 2019


BRMcLaren:
When launching Windows 2019 vm's a common practice is for Windows Update to run before a user is able to tunnel into the VM. Yesterday this was the output configuration of Windows Update

Intel DCAP v1.8.x
Intel PSW v2.7.x

Which is not compatible versions whereas the current dcap and psw need to be aligned in this case 1.8 and 2.9, otherwise compatibility issues will be seen. Our installation scripts for Windows only update the DCAP components and rely on Windows Update for the PSW, this leaves and edge case. We need to update the installation scripts to allow the user to do one of two things, either update to a compatible version (latest, usually) or allow users to not install any DCAP or PSW dependencies.

We should also add a document that outlines what that compatibility matrix is if it is not already done (I have not seen one).

radhikaj:
We will use this issue to track script problems and @BRMcLaren  will open another issue (tracked as External) to track the issue when mismatched PSW and DCAP are installed by WU.

BRMcLaren:
@radhikaj  we have been stable for 2 weeks after the fixes here: https://github.com/openenclave/openenclave/pull/3525

Closing as resolved, we have an external issue for the upstream windows update issues: https://github.com/openenclave/openenclave/issues/3532

PR #3525, opened 2020-09-10
Psw2019mismatch

BRMcLaren:
This PR:

* Changes the NoDriver Flag to NoPSWDCAP for the windows installation script. This is done as users either need to install Intel PSW/DCAP either both or neither.
* Some documentation changes to reflect the above
* Switches use of devcon to pnputil for windows 2019 as recommended by Intel
* Changes Jenkins image rebuild to use a new PSW and DCAP instead of just new DCAP
* Disable Win2019 Release Cross Compile DCAP LVI snmalloc test -- opening #3530 to track

I will add that as the deprecation of this script has been decided on, there is some refactoring either left out or issues tracked as getting it in to unblock the CI was more the priority.

### DIFF
--- a/.jenkins/pipelines/Azure/Windows/Jenkinsfile
+++ b/.jenkins/pipelines/Azure/Windows/Jenkinsfile
@@ -142,8 +142,8 @@ try{
                 "Win2019 Ubuntu1804 clang-7 Release Linux-Elf-build LVI":          { windowsLinuxElfBuild('acc-win2019-dcap', '18.04', 'clang-7', 'Release', 'ControlFlow', 'ON') },
                 "Win2019 Sim Release Cross Compile LVI ":                          { windowsCrossCompile('acc-win2019', 'Release', 'OFF', 'ControlFlow', '1', 'ON') },
                 "Win2019 Debug Cross Compile DCAP LVI":                            { windowsCrossCompile('acc-win2019', 'Debug', 'ON', 'ControlFlow', '0', 'ON') },
-                "Win2019 Release Cross Compile DCAP LVI FULL Tests":               { windowsCrossCompile('acc-win2019', 'Release', 'ON', 'ControlFlow') },
-                "Win2019 Release Cross Compile DCAP LVI FULL Tests snmalloc":      { windowsCrossCompile('acc-win2019', 'Release', 'ON', 'ControlFlow', '0', 'OFF', ['-DUSE_SNMALLOC=ON']) }
+                "Win2019 Release Cross Compile DCAP LVI FULL Tests":               { windowsCrossCompile('acc-win2019', 'Release', 'ON', 'ControlFlow') }
+                //"Win2019 Release Cross Compile DCAP LVI FULL Tests snmalloc":      { windowsCrossCompile('acc-win2019', 'Release', 'ON', 'ControlFlow', '0', 'OFF', ['-DUSE_SNMALLOC=ON']) }
             ]
             parallel testing_stages
         }
--- a/.jenkins/pipelines/Azure/Windows/Jenkinsfile
+++ b/.jenkins/pipelines/Azure/Windows/Jenkinsfile
@@ -127,8 +127,8 @@ try{
 
                 "Win2016 Sim Release Cross Compile LVI snmalloc":         { windowsCrossCompile('acc-win2016', 'Release', 'OFF', 'ControlFlow', '1', 'OFF', ['-DUSE_SNMALLOC=ON']) },
                 "Win2016 Release Cross Compile DCAP LVI snmalloc":        { windowsCrossCompile('acc-win2016', 'Release', 'ON', 'ControlFlow', '0', 'OFF', ['-DUSE_SNMALLOC=ON']) },
-                "Win2019 Sim Debug Cross Compile LVI snmalloc":           { windowsCrossCompile('acc-win2019', 'Debug', 'OFF', 'ControlFlow', '1', 'OFF', ['-DUSE_SNMALLOC=ON']) },
-                "Win2019 Release Cross Compile DCAP LVI snmalloc":        { windowsCrossCompile('acc-win2019', 'Release', 'ON', 'ControlFlow', '0', 'OFF', ['-DUSE_SNMALLOC=ON']) }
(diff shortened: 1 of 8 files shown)


radhikaj (approved):
Thanks for fixing this @BRMcLaren. Much appreciated!

scripts/install-windows-prereqs.ps1
```diff
@@ -35,7 +35,7 @@ Param(
     [string]$GetPipURL = 'https://bootstrap.pypa.io/3.4/get-pip.py',
     [string]$GetPipHash = '564FABC2FBABD9085A71F4A5E43DBF06D5CCEA9AB833E260F30EE38E8CE63A69',
     [Parameter(mandatory=$true)][string]$InstallPath,
-    [Parameter(mandatory=$true)][ValidateSet("SGX1FLC", "SGX1", "SGX1FLC-NoDriver", "SGX1-NoDriver")][string]$LaunchConfiguration,
+    [Parameter(mandatory=$true)][ValidateSet("SGX1FLC", "SGX1", "SGX1FLC-NoPSWDCAP", "SGX1-NoPSWDCAP")][string]$LaunchConfiguration,
```
rs--, line 38: Would something like `-NoIntelDrivers` be a better name? I initially thought `-NoPSWDCAP` meant installing DCAP but not PSW.

scripts/install-windows-prereqs.ps1
```diff
@@ -401,27 +401,32 @@ function Install-7Zip {
...
+        }
+        $unattendedParams = @('--s', '--a', 'install', "--output=$tempInstallDir\psw-installer.log", '--eula=accept', '--no-progress')
+        $p = Start-Process -Wait -NoNewWindow -FilePath $installer -ArgumentList $unattendedParams -PassThru
+        if($p.ExitCode -ne 0) {
+            Get-Content "$tempInstallDir\psw-installer.log"
+            Throw "Failed to install Intel PSW"
+        }
+    } else {
+        $psw_dir = Get-Item "$tempInstallDir\Intel*SGX*\PSW_INF*\"
+        pnputil /add-driver $psw_dir\sgx_psw.inf /install
```
rs--, line 428: Do we need to catch any errors from `pnputil`?

BRMcLaren:
Solves part of #3523

scripts/install-windows-prereqs.ps1
```diff
@@ -588,17 +593,17 @@ function Install-DCAP-Dependencies {
...
-                if($LASTEXITCODE) {
-                    Throw "Failed to install $driver driver"
+                if($OS_VERSION -eq "WinServer2016")
+                {
+                    $install = & $devConBinaryPath install "$($inf.FullName)" $drivers[${OS_VERSION}][$driver]['location']
+                    if($LASTEXITCODE) {
+                        Throw "Failed to install $driver driver"
+                    }
+                } else{
+                    $install = & pnputil /add-driver "$($inf.FullName)" /install
```
rs--, line 603: Same thing here wrt pnputil error

.jenkins/pipelines/Azure/Windows/Jenkinsfile
```diff
@@ -127,8 +127,8 @@ try{
 
                 "Win2016 Sim Release Cross Compile LVI snmalloc":         { windowsCrossCompile('acc-win2016', 'Release', 'OFF', 'ControlFlow', '1', 'OFF', ['-DUSE_SNMALLOC=ON']) },
                 "Win2016 Release Cross Compile DCAP LVI snmalloc":        { windowsCrossCompile('acc-win2016', 'Release', 'ON', 'ControlFlow', '0', 'OFF', ['-DUSE_SNMALLOC=ON']) },
-                "Win2019 Sim Debug Cross Compile LVI snmalloc":           { windowsCrossCompile('acc-win2019', 'Debug', 'OFF', 'ControlFlow', '1', 'OFF', ['-DUSE_SNMALLOC=ON']) },
-                "Win2019 Release Cross Compile DCAP LVI snmalloc":        { windowsCrossCompile('acc-win2019', 'Release', 'ON', 'ControlFlow', '0', 'OFF', ['-DUSE_SNMALLOC=ON']) }
+                "Win2019 Sim Debug Cross Compile LVI snmalloc":           { windowsCrossCompile('acc-win2019', 'Debug', 'OFF', 'ControlFlow', '1', 'OFF', ['-DUSE_SNMALLOC=ON']) }
+                //"Win2019 Release Cross Compile DCAP LVI snmalloc":        { windowsCrossCompile('acc-win2019', 'Release', 'ON', 'ControlFlow', '0', 'OFF', ['-DUSE_SNMALLOC=ON']) }
```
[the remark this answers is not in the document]
BRMcLaren: I opened this to track: https://github.com/openenclave/openenclave/issues/3530, we want to keep the testing just disable it until we can confirm if it is a code issue, a Jenkins issue, a server issue etc. I had 10 or failures in a row such as this: https://oe-jenkins-tf.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAzure-Windows/detail/Azure-Windows/250/pipeline where all the tests pass, the packaging passes and then the host just dies.

scripts/install-windows-prereqs.ps1
BRMcLaren: When it fails it will throw an error. I went down the path of error catching and retrying but as we are deprecating the scripts eventually just got it to work and made a note about it in the PR.

scripts/install-windows-prereqs.ps1
BRMcLaren: I have rebuilt the images to reference this flag, but this is a good point. Would you accept merging this in and then running a find/replace as a follow-up? If I change the flag now.. we delay the release by another day which is where my concern lies?

rs--: Works for me

rs-- (approved):

scripts/install-windows-prereqs.ps1
BRMcLaren: @rs-- I ended up needing to wait on automation, I pushed out the prod images with the underlying commit being before the refactor. I went ahead and added your changes requested. If e2e fails, I will remove the change and merge without. If it passes, merge with.

Issue #3430, opened 2020-08-24
OE SDK doesn't support signing and loading enclaves that need the provisioning key access


lzha101:
From the doc [build and sign](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/buildandsign.md#signing-an-sgx-enclave), current OE SDK supports below configuration fields: 
```
Debug
NumTCS
NumStackPages
NumHeapPages
ProductID
SecurityVersion
```
And the corresponding values could be also provided in the OE_SET_ENCLAVE_SGX  macro in another way for debug mode.

It doesn't provide any interface to configure other attribute flags, such as provisioning key, launch key access. Launch key may not be a big deal.  But the provisioning key access should be required in some Intel's AEs.  Need to update the oesign and loader to support this case.

anakrish:
Rationale: Needed for feature parity with SGX SDK.

MWShan:
@binxing - can you review this issue?

binxing:
The enclave configuration is embedded in the enclave image in the form of an oe_sgx_enclave_properties_t structure. oesign only overwrites those fields set explicitly in .conf file. My thought is, for enclaves that require setting attribute bits not provided/supported by oesign or OE_SET_ENCLAVE_SGX, the developer could define an oe_sgx_enclave_properties_t global constant explicitly. The code may look a bit ugly but that's needed for just a couple of architectural enclaves anyway.

Btw, I think OE_SET_ENCLAVE_SGX works in both debug and production code. Am I missing anything?

lzha101:
@binxing,  users may manually define the global constant, but current implementation restricts the attributes could never be other values rather than DEBUG flag and MODE64BIT flag.  See https://github.com/openenclave/openenclave/blob/master/include/openenclave/internal/sgx/sgxproperties.h#L35

Another thing is: when creating enclave, the attributes in secs are not copied from the oe_sgx_enclave_properties but set with (DEBUG and MODE64BIT) by the implementation each time, see https://github.com/openenclave/openenclave/blob/master/host/sgx/sgxload.c#L349

I suppose it is better to provide interfaces rather than asking users to define the global variable. As you said, the code looks ugly. And even though allow to use this variable,  OE also needs to consider how to correctly validate the input attributes.

binxing:
[...]

PR #3526, opened 2020-09-11
Add PROVISION_KEY/EINITTOKEN_KEY support to SGX enclave loader

binxing:
This PR adds `OE_SGX_FLAGS_PROVISION_KEY` and `OE_SGX_FLAGS_EINITTOKEN_KEY` to the list of valid enclave attributes so that SGX architectural enclaves (e.g. PCE and LE) can be built, signed and loaded by OE SDK.

Signed-off-by: Cedric Xing <cedric.xing@intel.com>


mingweishih (approved):
LGTM

CodeMonkeyLeet (commented):
Thanks for the contribution, Cedric! The changes look good though we usually also encourage the addition of tests along with the feature changes. In this case, it would be helpful to update tests/props to verify that the new options are correctly supported:

- [x] Add a third test enclave that uses a new `custom_props.c` that defines the `volatile const oe_sgx_enclave_properties_t oe_enclave_properties_sgx` directly to specify the new attributes for provision and EINIT token keys.
- [x] Add a new ctest and update the test host to verify the new attributes in the added `custom_unsigned` test case.

binxing:
Thanks @CodeMonkeyLeet for the comments! I added a test. However, it seems additional authorization (i.e. open()/ioctl() on /dev/sgx/provision) is required to set PROVISION_KEY attributes. Given FLC, I'm not sure the kernel behavior when EINITTOKEN_KEY is requested. Moreover, OE so far is unaware of SGX kernel interface so I don't think it appropriate to add such additional steps to OE. Maybe they should be part of the common enclave loader from Intel.

Btw, sorry for messing up the git branch that caused this PR to be closed accidentally. I'll have it fixed in no time.

tests/props/host/host.c
```diff
@@ -126,7 +126,7 @@ int main(int argc, const char* argv[])
 {
     oe_result_t result;
     oe_enclave_t* enclave = NULL;
-    bool is_signed = false;
+    int test_type = 0;
```
CodeMonkeyLeet, line 129: For clarity, can you use named enumerations here instead of raw ints?
```c
typedef enum
{
    TEST_PROPS_UNSIGNED,
    TEST_PROPS_SIGNED,
    TEST_PROPS_CUSTOM_SIGNED
} test_type_t;
```
```suggestion
    test_type_t test_type;
```

tests/props/host/host.c
```diff
@@ -165,29 +170,46 @@ int main(int argc, const char* argv[])
...
+            OE_SGX_FLAGS_DEBUG | OE_SGX_FLAGS_MODE64BIT, /* attributes */
+            512,                                         /* num_heap_pages  */
+            512,                                         /* num_stack_pages */
+            4);                                          /* num_tcs */
+        break;
+    case 1: /* signed */
         _check_properties(
             &properties,
-            is_signed,
+            !!test_type,
```
CodeMonkeyLeet, line 189: Similarly, this works, but the intent is unclear, consider instead:
```suggestion
            (test_type != TEST_PROPS_UNSIGNED),
```

also on line 200.

CodeMonkeyLeet (commented):
Thanks for adding the test! Just a couple of coding guidelines comments and we'll get this merged.

tests/props/host/host.c
```diff
@@ -165,29 +177,46 @@ int main(int argc, const char* argv[])
         oe_put_err("oe_create_enclave(): result=%u", result);
 
     /* Check expected enclave property values */
-    if (is_signed)
+    switch (test_type)
     {
+    case 0: /* unsigned */
```
CodeMonkeyLeet, line 182: Can you replace the case labels with the enum? That was part of the reason for the original suggestion to use enums.

lzha101:
Thanks Cedric for the patch to address the issue https://github.com/openenclave/openenclave/issues/3430.

I could be wrong. In my understanding, the patch in this PR allows PROVISION_KEY in the sigstruct by the self-defined the global variable `oe_enclave_properties_sgx`.  But if we want to do so, we also need to update the function `_new_secs()`，  right? 
https://github.com/openenclave/openenclave/blob/master/host/sgx/sgxload.c#L140. 
This `_new_secs()` function would construct a secs. Currently the flag setting is only about MOD64bit and DEBUG.  The attributes in this secs structure would be passed to ECREATE instruction. EINIT would check the corresponding attributes in secs and sigstruct. 
If we want to allow PROVISION_KEY bit, we also need to set the attributes in secs passed to ECREATE using the variable `oe_enclave_properties_sgx` to make sure the attributes are consistent. Otherwise, the attributes check in EINIT instruction would be failed.  Is that correct?

binxing:
Thanks @lzha101 for pointing out my mistake. The PR branch now should have that fixed. This version also has incorporated @CodeMonkeyLeet's comments on coding style.

tests/props/enc/custom_props.c
```diff
@@ -0,0 +1,21 @@
...
+OE_INFO_SECTION_BEGIN
+volatile const oe_sgx_enclave_properties_t oe_enclave_properties_sgx = {
+    .header = {.size = sizeof(oe_sgx_enclave_properties_t),
+               .enclave_type = OE_ENCLAVE_TYPE_SGX,
+               .size_settings = {.num_heap_pages = 512,
+                                 .num_stack_pages = 512,
+                                 .num_tcs = 4}},
+    .config = {.product_id = 1111,
+               .security_version = 2222,
+               .attributes = /* These 2 flags cannot be tested as SGX common
```
CodeMonkeyLeet, line 15: @binxing Why is this disabled (similarly for line 210 in host.c) i.e. what additional errors are you encountering given the changes to sgxload.c and sgxproperties.h already?

CodeMonkeyLeet: Copying context from #3430:

> One more problem that can't be addressed by #3526, is that the PROVISION_KEY attribute requires additional authorization by opening /dev/sgx/provision for the current in-tree driver. OE SDK is unaware of the kernel interface so I think it should be captured inside common enclave loader.

The default test configuration that we run in CI/CD still uses out-of-tree drivers, so these tests could still be enabled, no? Broadly speaking, once the OE SDK CI/CD starts to run against the in-tree SGX driver, we would have to configure permissions for all the test apps that used DCAP (since it would require process/user/group permissions) as well, and those changes would be needed for this new test as well, correct?

Is there a tracking item on the Intel side for this? I'm unclear on what change is expected in the enclave common loader for this to work.

binxing: The error I got was `ENCLAVE_NOT_AUTHORIZED` (`6`). It was due to `-EACCES` returned by the driver. However, I double-checked the common loader source today and figured that it did open `/dev/sgx/provision` whenever `PROVISION_KEY` is requested. What's missing was that my user account didn't belong to `sgx_prv` group. The test case then worked after the group membership was fixed. I have to apologize for the confusion.

I've updated the test case to request `PROVISION_KEY` on the PR branch and now everything is working. Please note that `EINITTOKEN_KEY` won't work (therefore not set/tested in the test case) because the in-tree driver blocks it. But that shouldn't be a practical problem as LE is no longer necessary.

lzha101: @CodeMonkeyLeet, which driver is using in CI/CD?  DCAP driver? If yes and OE has a test enclave requires to access provision key, the CI/CD should update configuration to grant the permission of loading a provision enclave following the DCAP driver [README ](https://github.com/intel/SGXDataCenterAttestationPrimitives/tree/master/driver/linux#launching-an-enclave-with-provision-bit-set).

CodeMonkeyLeet: @lzha101 Thanks for the pointer! OE CI/CD is currently still using the DCAP driver but doesn't use the provisioning key bit directly (it's only used through the PCE/QE as part of attestation). I believe we're still waiting for the availability of the in-tree driver to be available in the ubuntu image before testing the permission grants. It's largely a resource prioritization decision since we're also shifting the technology stack for our CI/CD right now, and we're keeping an eye on the progress of the in-tree driver upstreaming.

CodeMonkeyLeet (approved):
LGTM, thanks for all the fixups!

On additional question for my understanding: if the in-tree driver blocks enclaves with the `OE_SGX_FLAGS_EINITTOKEN_KEY` flag, this means that the LE is effectively deprecated moving forwards, correct? Does the AESM in PSW 2.10 already handle this case? (i.e. does not attempt to load the LE when in-tree driver is present).

include/openenclave/internal/sgx/sgxproperties.h
```diff
@@ -32,7 +32,8 @@ OE_INLINE bool oe_sgx_is_valid_num_tcs(uint64_t x)
 OE_INLINE bool oe_sgx_is_valid_attributes(uint64_t x)
 {
     /* Check for illegal bits */
-    if (x & ~(OE_SGX_FLAGS_DEBUG | OE_SGX_FLAGS_MODE64BIT))
+    if (x & ~(OE_SGX_FLAGS_DEBUG | OE_SGX_FLAGS_MODE64BIT |
+              OE_SGX_FLAGS_PROVISION_KEY | OE_SGX_FLAGS_EINITTOKEN_KEY))
```
CodeMonkeyLeet, line 36: Should we add support in `OE_SGX_FLAGS_EINITTOKEN_KEY` if it is being deprecated by Intel anyway?

binxing: @MWShan may correct me if I'm wrong. Intel doesn't deprecate LE or `EINITTOKEN_KEY`. It's the Linux community's choice to support FLC-enabled platforms only, while other OS's may continue to use/support LE. Technically, an SGX enclave is OS-agnostic and IMHO, as an SDK, OE should support all attributes documented in the SDM.

binxing:
bors try

binxing:
@radhikaj, I rebased the patch to the latest master, which should fix the build failure. Could you please try merging it again? Thanks!

binxing:
@mingweishih and @radhikaj, I tested the PR_support_provkey_and_einitkey_bits_in_loader branch locally but didn't encounter any build errors. How can I view the error messages of the build by bors? Or how to checkout the source that bors tried to build?

mingweishih:
> @mingweishih and @radhikaj, I tested the PR_support_provkey_and_einitkey_bits_in_loader branch locally but didn't encounter any build errors. How can I view the error messages of the build by bors? Or how to checkout the source that bors tried to build?

You should be able to see the log via the link associated with bors's comment.
Here are some failures I can see: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAgnostic-Linux/detail/Agnostic-Linux/1799/pipeline
```
CI failed: please run ./scripts/format-code



script returned exit code 1
```

It seems the code didn't pass the format check.

binxing:
@mingweishih, thanks and I've got the formatting fixed. Could you please try bors again?

mingweishih:
bors try

binxing:
Hi @mingweishih, @radhikaj and other reviewers, the build failed because tests/props_custom failed on both Linux and Windows. 

The reason it failed on Linux was because the user account that ran the test didn't belong to the `sgx_prv` group (the `PROVISION_KEY` attribute requires that group membership).  I think it could only be fixed by changing the bors test environment.

It probably failed on Windows for a similar reason but I'm not sure yet.

CodeMonkeyLeet:
@BRMcLaren Our CI image is still using the Intel driver download and not an ubuntu image with the built-in drivers right? It looks like we might need to start enabling the `sgx_prv` permission group in our config to enable this test per @binxing

BRMcLaren:
> @BRMcLaren Our CI image is still using the Intel driver download and not an ubuntu image with the built-in drivers right? It looks like we might need to start enabling the `sgx_prv` permission group in our config to enable this test per @binxing

Assuming machine image versus docker image in this context, yes that is correct.

BRMcLaren:
@binxing if we go change the environment, how do we test that this is done correctly, and is the fix? We will need to modify the ansible files to do this is it just a simple ctest of this branch or are there additional requirements?

binxing:
> @binxing if we go change the environment, how do we test that this is done correctly, and is the fix? We will need to modify the ansible files to do this is it just a simple ctest of this branch or are there additional requirements?

@BRMcLaren, yes, successful ctest of this branch should suffice to indicate a proper configuration. But a failed ctest may be because of something else. This branch tested successfully on ACC VM but I'm not sure if there are other differences between the envs that may affect the test results.

binxing:
The intention of this PR was to allow building architectural enclaves using OE SDK. But Intel gave up on that. So this PR is no longer needed.

PR #3529, opened 2020-09-12
Attested tls sample using openssl library - Non enclave client to enclaved server TLS channel

manojrupireddy:
* These changes add files to demonstrate TLS secure channel communication between an attested server running inside an enclave and an external client running outside an enclave. TLS channel is established using constructs from openssl library.
* This PR does not include secure channel communication where both server and clients are running inside an enclave. I will make a separate for that so that it will be bit easier for code review.
* Pending items in this PR
  * Yet to update Readme file 
  * After closing the sample application, the sockets opened for communication are not getting closed immediately. For 5-10 seconds they are hanging in TIME_WAIT state before going to close state. I will fix this in next revisions.


samples/attested_tls/common/utility.cpp
```diff
@@ -92,7 +89,7 @@ oe_result_t generate_certificate_and_pkey(
         goto exit;
     }
 
-    printf("public key used:\n%s\n", public_key_buf);
+    printf(" here public key used:\n%s\n", public_key_buf);
```
yentsanglee, line 92: " here " seems redundant

samples/attested_tls/server/enc/openssl/public.pem
```diff
@@ -0,0 +1,39 @@
+-----BEGIN RSA PRIVATE KEY-----
```
yentsanglee, line 1: This is identical to the private key.

samples/attested_tls/common/utility.cpp
```diff
@@ -110,13 +107,47 @@ oe_result_t generate_certificate_and_pkey(
...
+        goto exit;
+    }
+
+exit:
+    oe_attester_shutdown();
+    oe_free_key(public_key_buf, public_key_buf_size, NULL, 0);
+    return result;
+}
+
+// Consider to move this function into a shared directory
```
dthaler, line 124: I thought "common" was already a shared directory, no?

samples/attested_tls/non_enc_client/client.cpp
```diff
@@ -66,70 +67,24 @@ int parse_arguments(
...
 int communicate_with_server(SSL* ssl)
 {
-    unsigned char buf[200];
     int ret = 1;
     int error = 0;
-    int len = 0;
-    int bytes_written = 0;
-    int bytes_read = 0;
-
     // Write an GET request to the server
```
dthaler, line 72:
```suggestion
    // Write a GET request to the server
```

samples/attested_tls/server/enc/mbedtls/ecalls.cpp
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 #include <openenclave/enclave.h>
-#include "../../common/tls_server_enc_pubkey.h"
+#include "../../../common/tls_server_enc_pubkey.h"
```
dthaler, line 5: this ecalls.cpp file is another example of a file that doesn't seem to contain any mbedtls specific content.  Can this be left in enc?

samples/attested_tls/server/enc/openssl/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+# Licensed under the MIT License.
+
+# Use the edger8r to generate C bindings from the EDL file.
+add_custom_command(
+  OUTPUT tls_server_t.h tls_server_t.c tls_server_args.h
+  DEPENDS ${CMAKE_SOURCE_DIR}/server/tls_server.edl
+  COMMAND
+    openenclave::oeedger8r --trusted ${CMAKE_SOURCE_DIR}/server/tls_server.edl
+    --search-path ${OE_INCLUDEDIR} --search-path
+    ${OE_INCLUDEDIR}/openenclave/edl/sgx)
```
dthaler, line 11: Can we separate SGX-specific things from TEE-agnostic things?
For example you nicely have OE_CRYPTO_LIB in line 17 but here hard-codes SGX instead of using something like OE_TEE

samples/attested_tls/server/enc/openssl/server.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+    ret = 0;
+exit:
+    cert_buf_ptr = NULL;
+    oe_free_key(private_key_buf, private_key_buf_size, NULL, 0);
+    oe_free_attestation_certificate(output_cert);
+    return ret;
+}
+
+int create_listner_socket(int port, int& server_socket)
```
dthaler, line 191:
```suggestion
int create_listener_socket(int port, int& server_socket)
```

samples/attested_tls/server/enc/openssl/server.cpp
```diff
@@ -0,0 +1,377 @@
...
+{
+    int ret = -1;
+    struct sockaddr_in addr;
+    addr.sin_family = AF_INET;
```
dthaler, line 195: I'm sad to see IPv4-only samples.  
This is not a blocker, but I'd really like all samples to be IP-version agnostic.

manojrupireddy: I could not understand complete implementation details of `create_socket `in  [file](https://github.com/openenclave/openenclave/blob/98b71a493cbbadda84c72187e9b5a59a3a1d7dec/samples/attested_tls/non_enc_client/client.cpp#L139 ). Is it IP-version agnostic? @dthaler

samples/attested_tls/server/enc/openssl/public.pem
manojrupireddy: the earlier mbedtls implementation also have same pattern [here](https://github.com/openenclave/openenclave/tree/master/samples/attested_tls/server/enc). Could that be because of mistake? Do i need to replace with newly generated pubic , private keys. Also the alternative we can follow is generate them during build process, instead of checking in static files. @yentsanglee @qiucwang

samples/attested_tls/server/enc/openssl/CMakeLists.txt
manojrupireddy: Looks like hardcoded sgx value is repeated across all the sample files. Does that mean current samples support only sgx platform?

samples/attested_tls/server/enc/openssl/public.pem
yentsanglee: The public keys are used to compare against signer_id after verification but the comparisons use the public keys defined in `OTHER_ENCLAVE_PUBLIC_KEY`. Both public.pem files are not used except the contents were copied-n-pasted to tls_client_enc_pubkey.h and tls_server_enc_pubkey.h.

Even thought this public.pem is not used but it's clearly incorrect. You may generate new key pairs but remember to replace the content in `OTHER_ENCLAVE_PUBLIC_KEY`.

manojrupireddy:
Duplicate of https://github.com/openenclave/openenclave/pull/3750 PR

PR #3538, opened 2020-09-15
PR TO EXPERIMANTAL DEV BRANCH: Thread-local support for multiple modules.

anakrish:
The main module uses -ftls-model=local-exec flag.
This is the most optimal. The compiler/linker hard-codes the offsets from FS
since it knows that the main module is the first one to be loaded.

The secondary modules are compiles with -ftls-model=initial-exec flag.
This informs the compiler that the modules will be loaded along with the main module,
rather than being dynamically loaded via dlopen. In this case, the compiler
will generate R_X86_64_TPOFF relocations for thread-local variables.
Each thread-local variable has a corresponding offset variable. In order to
access a thread-local variable, the value of the corresponding offset variable is
first read and then added to FS. Since thread-locals have negative offsets, this
implies that the offset variables have negative values.

The loader places the thread local sections (tdata + tbss) of the modules one after
the other, starting with the main module. The linker initializes the offset variables
to offsets within the module's tls section. The loader converts these offsets to values
to values relative to the start of the aggregated tls section. Care is taken to account
for possible different alignments of the different tls sections.

### DIFF
--- a/enclave/core/sgx/threadlocal.c
+++ b/enclave/core/sgx/threadlocal.c
@@ -12,6 +12,7 @@
 #include <openenclave/internal/sgx/td.h>
 #include <openenclave/internal/thread.h>
 #include <openenclave/internal/utils.h>
+#include "openenclave/internal/link.h"
 #include "td.h"
 
 /*
@@ -190,20 +191,9 @@
  * value.
  *
  * To avoid looking up symbols within the enclave (symbols are not available)
- * the loader fetches the symbol's sh-value and stores it in the r_addend field
- * (r_added field is otherwise zero for R_X86_64_TPOFF64).
+ * the loader performs the relocations and patches the enclave image before
+ * measurement.
  */
-OE_EXPORT volatile uint64_t _tdata_rva = 0;
-OE_EXPORT volatile uint64_t _tdata_size = 0;
-OE_EXPORT volatile uint64_t _tdata_align = 1;
-
-OE_EXPORT volatile uint64_t _tbss_size = 0;
-OE_EXPORT volatile uint64_t _tbss_align = 1;
-
-static volatile bool _thread_locals_relocated = false;
-
-// TODO: Make this flexible in case more than one page of thread local storage
-// need to allocate.
 
 /* The thread data (td) object is always populated at the start of the
    FS segment, so this method just returns the address of the td.
@@ -212,14 +202,6 @@ static uint8_t* _get_fs_from_td(oe_sgx_td_t* td)
 {
     uint8_t* fs = (uint8_t*)td;
     return fs;
-}
-
-/**
- * Return aligned size.
- */
-static uint64_t _get_aligned_size(uint64_t size, uint64_t align)
-{
-    return align ? oe_round_up_to_multiple(size, align) : size;
 }
 
 /*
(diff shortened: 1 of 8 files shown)


include/openenclave/internal/link.h
```diff
@@ -37,6 +39,41 @@ typedef struct _oe_module_link_info
 
 } oe_module_link_info_t;
 
+/* thread-local management  function shared by host and enclave */
```
CodeMonkeyLeet, line 42:
```suggestion
/* thread-local management function shared by host and enclave */
```

host/sgx/loadelf.c
```diff
@@ -1206,43 +1237,24 @@ static oe_result_t _patch_elf_image(
...
     memset(oeprops->sigstruct, 0, sizeof(oeprops->sigstruct));
 
     /* Dynamically link the enclave and its needed modules */
-    OE_CHECK(_link_elf_images(image, &module_base));
+    {
+        int64_t tls_start_offset = 0;
+        OE_CHECK(_link_elf_images(image, &module_base, &tls_start_offset));
+
+        // Compute the offset of td from total thread local size.
+        uint64_t total_tls_size = (uint64_t)-tls_start_offset;
```
CodeMonkeyLeet, line 1249: We should probably check that `tls_start_offset <= 0` for this right?

enclave/core/sgx/threadlocal.c
```diff
@@ -412,14 +336,24 @@ oe_result_t oe_thread_local_cleanup(oe_sgx_td_t* td)
...
+                oe_get_module_tls_start_offset(link_info, tls_start_offset);
+    }
+
+    // Invoke the cleanup function even when the thread-local data is
+    // empty (i.e., tls_start is NULL when tdata and tbss are zero).
     oe_allocator_thread_cleanup();
-    if (tls_start)
-        oe_memset_s(tls_start, (uint64_t)(fs - tls_start), 0, 0);
+
+    uint64_t tls_size = (size_t)-tls_start_offset;
```
CodeMonkeyLeet, line 354: Probably want to abort if `tls_start_offset` is somehow positive. Also, if we're casting, it might as well be to `uint64_t` directly.

tests/ldd/enc/dep_1a/CMakeLists.txt
```diff
@@ -33,7 +33,7 @@ target_compile_options(
     # too conservative. Of the two efficient static thread-local
     # storage models, inital-exec and local-exec, we choose the most
     # optimal one.
-    -ftls-model=local-exec
+    -ftls-model=initial-exec
```
CodeMonkeyLeet, line 36: Could you update the comment here with the one you put in the PR, here and in dep_2a/CMakeLists.txt?

```c
# The secondary modules are compiles with -ftls-model=initial-exec flag.
# This informs the compiler that the modules will be loaded along with the main module,
# rather than being dynamically loaded via dlopen. In this case, the compiler
will generate R_X86_64_TPOFF relocations for thread-local variables.
```

CodeMonkeyLeet (approved):
Thanks for doing this, didn't notice that the linker options were different from the compiler options for PIC!

host/sgx/loadelf.c
anakrish: Done.

tests/ldd/enc/dep_1a/CMakeLists.txt
anakrish: Done.

include/openenclave/internal/link.h
anakrish: Done.

enclave/core/sgx/threadlocal.c
anakrish: Done. Also added oe_is_within_enclave checks just to be sure.

PR #3541, opened 2020-09-16
Update VS extension to OE 0.11

dthaler:
Changes:
* Add System EDLs (syscall.edl and platform.edl).  PR #2961 did this for samples but not for the VS extension
* Remove dependencies on sgx_enclave_common.lib and sgx_dcap_ql.lib.  (Anand, please let me know if this is incorrect.)
* Add dependency on shlwapi.lib since oehost.lib now depends on it for PathCombineW
* Update package dependencies to latest versions to pick up any bug fixes
* Update version numbers of OE and the extension

### DIFF
--- a/devex/vsextension/ItemTemplates/oehost/host.vstemplate
+++ b/devex/vsextension/ItemTemplates/oehost/host.vstemplate
@@ -17,7 +17,7 @@
   </WizardExtension>
   <WizardData>
     <packages repository="extension" repositoryId="OpenEnclaveVisualStudioExtension-1">
-      <package id="open-enclave-cross" version="0.8.1-c3b6262c-3" />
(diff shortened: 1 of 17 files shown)


anakrish (approved):
LGTM.

Does the extension framework allow specifying the oe version in a single place and referring to it everywhere else?

dthaler:
> LGTM.
> 
> Does the extension framework allow specifying the oe version in a single place and referring to it everywhere else?

Unfortunately no.  That's why I use a command to find and replace all instances (see step 3 at a [PR I am working on](https://github.com/dthaler/openenclave/blob/vsext-release-doc/devex/vsextension/vsix-release-process.md) to document what steps are done now... this needs to be improved in the future but the first step is just to document what we do now for the VS extension).

mingweishih (approved):

PR #3548, opened 2020-09-16
Reduce max threads to 8 for pluggable_allocator sample.

anakrish:
This value allows the sample to pass on all known ACC VM configurations.


BRMcLaren (approved):

anakrish:
bors try

radhikaj (approved):

dthaler (commented):
This looks fine, but how was the number determined?   Is the number the same number that would apply to any other sample?  Can we document the answer in a .md file somewhere?

PR #3554, opened 2020-09-17
Avoid overwriting CMAKE_PREFIX_PATH

### LINKED ISSUE #3543: Unnecessary CMAKE_PREFIX_PATH in openenclaverc
`CMAKE_PREFIX_PATH` env var is supposed to contain a colon separated list of paths for cmake functions (such as `find_package`) to search for external modules. `share/openenclave/openenclaverc` however overwrites (rather than appending to) `CMAKE_PREFIX_PATH`, which potentially disables other external modules and would be problematic should an enclave depend on more external modules than just `OpenEnclave`. In fact, `CMAKE_PREFIX_PATH` is unnecessary as long as /*path/to/OpenEnclave_InstallDir*/bin is listed in `PATH`. Shall we remove it from `openenclaverc`? Or did I miss anything?

mingweishih:
The PR updates openenclaverc file and documents on Windows to avoid overwrite CMAKE_PREFIX_PATH.

Note that when setting CMAKE_PREFIX_PATH as environment variable on Linux, multiple paths should be colon-separated instead of semicolon-separated.

Fixes #3543


mingweishih:
bors try

CodeMonkeyLeet (commented):
Changes look good, but what was the concern around appended paths not working with openenclaverc in #3543?

mingweishih:
> Changes look good, but what was the concern around appended paths not working with openenclaverc in #3543?

It turns out to be a confusion caused by the [cmake documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html). According to that, `CMAKE_PREFIX_PATH` should be a semicolon-separated list. However, this only applies to when defining the variable with `-D`. For example,

```
cmake .. -DCMAKE_PREFIX_PATH=path1;path2;path3
```

When `CMAKE_PREFIX_PATH` is defined as an environment variable on Linux (like in our case), it should be a colon-separated list instead. I used to follow the documented way and failed, so I thought it does not work 😭.

CodeMonkeyLeet (approved):

radhikaj (approved):
Thanks @mingweishih

PR #3564, opened 2020-09-19
Replace temporary_buffer_size with NULL in _get_attester_plugins()

### LINKED ISSUE #3500: _get_attester_plugins: buffer overread after ocall
Returned buffer size is not validated against supplied buffer size, resulting in a potential buffer overread into enclave heap.  Out-of-bounds memory is copied safely into enclave heap due to memory allocation being based on host-reported buffer size. Full exploitability undetermined, but risk thought to be low.

https://github.com/openenclave/openenclave/blob/5a85b33476f94eebd57a7abdd34bd7584efeb348/enclave/sgx/attester.c#L374-L384

ryanhsu19:
Replace `&temporary_buffer_size` in line 378 in `enclave/sgx/attester.c` with `NULL`.

The value of `temporary_buffer_size` is already validated near line 360 where `oe_get_supported_attester_format_ids_ocall` is first called, which eventually calls `oe_sgx_get_supported_attester_format_ids` in `host/sgx/sgxquote.c` and does check against supplied buffer size.

Fix #3500 .

### DIFF
--- a/enclave/sgx/attester.c
+++ b/enclave/sgx/attester.c
@@ -372,10 +372,7 @@ static oe_result_t _get_attester_plugins(
 
         // Get the format IDs
         result = oe_get_supported_attester_format_ids_ocall(
-            (uint32_t*)&retval,
-            temporary_buffer,
-            temporary_buffer_size,
-            &temporary_buffer_size);
+            (uint32_t*)&retval, temporary_buffer, temporary_buffer_size, NULL);
         OE_CHECK(result);
         OE_CHECK(retval);
     }


yentsanglee (approved):

douglasmaciver (approved):
While this change addresses the security concern, it does pose a functional incorrectness if the second call to `oe_get_supported_attester_format_ids_ocall` returns fewer format_ids than the first reported. But considering that the list of plugins may change after the second call in either case, this slight functional incorrectness does not introduce new functional risk, so I approve.

PR #3567, opened 2020-09-21
Remove oe_time/time from corelibc

wintersteiger:
Removes `oe_time` and `time` from corelibc. Proposed fix for #3516.

This changes `oe_datetime_now` to use `oe_get_time` instead of `time` in the enclave, so that it works in corelibc-only projects. However, it still calls `gmtime_r`, which may or may not end up being included in corelibc, see #3517.


mingweishih (approved):
LGTM

jhand2 (approved):

achamayou (approved):

wintersteiger:
bors try

Issue #3569, opened 2020-09-21
_get_attester_plugins: second ocall - should check output buffer size instead of ignoring it.


shnwc:
Follow up on issue #3500 and PR #3564 .

As @douglasmaciver [commented](https://github.com/openenclave/openenclave/pull/3564#pullrequestreview-492007388):

> While this change addresses the security concern, it does pose a functional incorrectness if the second call to oe_get_supported_attester_format_ids_ocall returns fewer format_ids than the first reported. ...

In case this happens, the list of plugins returned by `_get_attester_plugins()` in `enclave/sgx/attester.c` would countain plugins that have random format IDs.

A better fix is check the output buffer size by the second call to `oe_get_supported_attester_format_ids_ocall()`, to ensure it is no larger than the input buffer size. This fix will address the security concern, and be functionally correct.

douglasmaciver:
@shnwc Kudos for taking the initiative to make this better. While you are working on this, you might look at how the system handles changes to the plugins set, or if that's even possible.

radhikaj:
From triage - avoid double call patterns. Look into if there is a way around it here.

radhikaj:
Shan-wei will look into the short term solution.

shnwc:
> @shnwc Kudos for taking the initiative to make this better. While you are working on this, you might look at how the system handles changes to the plugins set, or if that's even possible.

With the current implementation, it's not possible for an attester application to discover changes in plugins set and re-register them after it calls `oe_attester_initialize()`, except restarting. But if a plugin can no longer function properly (e.g. if it requires a background service like AESM, but the service is no longer accessible), the plugin will return proper error code.

shnwc:
Closed by PR #3570

PR #3570, opened 2020-09-21
Check second ocall output buffer size in _get_attester_plugins

shnwc:
Fixes issue #3569

### DIFF
--- a/enclave/sgx/attester.c
+++ b/enclave/sgx/attester.c
@@ -365,6 +365,7 @@ static oe_result_t _get_attester_plugins(
     // It's possible that there is no supported format
     if (temporary_buffer_size >= sizeof(oe_uuid_t))
     {
+        size_t allocated_buffer_size = temporary_buffer_size;
         // Allocate buffer to held the format IDs
         temporary_buffer = (uint8_t*)oe_malloc(temporary_buffer_size);
         if (temporary_buffer == NULL)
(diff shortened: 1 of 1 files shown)


shnwc:
bors try

yentsanglee (approved):

enclave/sgx/attester.c
```diff
@@ -365,16 +365,22 @@ static oe_result_t _get_attester_plugins(
...
         result = oe_get_supported_attester_format_ids_ocall(
-            (uint32_t*)&retval, temporary_buffer, temporary_buffer_size, NULL);
+            (uint32_t*)&retval,
+            temporary_buffer,
+            allocated_buffer_size,
+            &temporary_buffer_size);
         OE_CHECK(result);
         OE_CHECK(retval);
+        if (temporary_buffer_size > allocated_buffer_size)
+            OE_RAISE(OE_UNEXPECTED);
```
CodeMonkeyLeet, line 383: Because this spans an OCALL, we've been pretty strict in checking that the returned size is exactly equal the originally specified size (i.e. also should not be smaller.)

shnwc: Agreed we can check to ensure the two values are the same. Will update the code.

shnwc:
bors try

CodeMonkeyLeet (approved):

PR #3578, opened 2020-09-23
change Linux Jenkins files to reference clang-8

BRMcLaren:
Signed-off-by: Brett McLaren <brmclare@microsoft.com>

### DIFF
--- a/.jenkins/pipelines/Azure/Nightly/Packaging.Jenkinsfile
+++ b/.jenkins/pipelines/Azure/Nightly/Packaging.Jenkinsfile
@@ -27,7 +27,7 @@ def LinuxPackaging(String version, String build_type, String lvi_mitigation = 'N
                            cpack
                            ctest --output-on-failure --timeout ${CTEST_TIMEOUT_SECONDS}
                            """
-                oe.Run("clang-7", task)
+                oe.Run("clang-8", task)
                 azureUpload(storageCredentialId: 'oe_jenkins_storage_account', filesPath: 'build/*.deb', storageType: 'blobstorage', virtualPath: "${BRANCH_NAME}/${BUILD_NUMBER}/ubuntu/${version}/${build_type}/lvi-mitigation-${lvi_mitigation}/SGX1FLC/", containerName: 'oejenkins')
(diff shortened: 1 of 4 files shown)


achamayou (approved):

.jenkins/pipelines/Azure/Nightly/Packaging.Jenkinsfile
```diff
@@ -27,7 +27,7 @@ def LinuxPackaging(String version, String build_type, String lvi_mitigation = 'N
                            cpack
                            ctest --output-on-failure --timeout ${CTEST_TIMEOUT_SECONDS}
                            """
-                oe.Run("clang-7", task)
```
achamayou, line 30: Not sure how easy it would be to parametrise this all the way through (or if it even makes sense, would two versions of clang ever be used at the same time), but it could help keep things in sync.

radhikaj (approved):

.jenkins/pipelines/Azure/Nightly/Packaging.Jenkinsfile
BRMcLaren: This is a good point, the changes if there were to be one would need to be here though: https://github.com/openenclave/openenclave-ci/pull/22/files.

To the best of my knowledge, we are planning to deprecate clang-7 and then moving to 10 in the new year. How about I parameterize and leave the default to clang-8?

achamayou: That sounds good to me, I can't think of any reason to use more than one variant of clang at once, and this will make future updates easier I would think.

BRMcLaren:
bors try

BRMcLaren:
bors try-

BRMcLaren:
bors try

dthaler:
@BRMcLaren I see this had 2 approvals (but bors failed) and this was closed without merging.  Is this abandoned or is there anotehr solution?

BRMcLaren:
Please follow: https://github.com/openenclave/openenclave/pull/3579 Changes need to go in at the same time.

BRMcLaren:
@dthaler Another solution, changes merged in here: https://github.com/openenclave/openenclave/pull/3579

PR #3582, opened 2020-09-24
pass check type of GLOB_DAT in rel.dyn in elf64_load_relocations

johnvenn:
Signed-off-by: Xiangping Ji <xiangping.ji@intel.com>
GLOB_DAT in rel.dyn check should pass in elf64_load_relocations in case where an extern func is declared as a weak symbol (__attribute__((weak))) and it's not actually defined.

regression test passed

### DIFF
--- a/host/sgx/elf.c
+++ b/host/sgx/elf.c
@@ -1909,7 +1909,11 @@ oe_result_t elf64_load_relocations(
     for (; p != end; p++)
     {
         uint64_t reloc_type = ELF64_R_TYPE(p->r_info);
-        if (reloc_type != R_X86_64_RELATIVE && reloc_type != R_X86_64_TPOFF64)
+        /* The enclave doesn't perform relocations on R_X86_64_GLOB_DAT, but
+         * we allow it for code that checks for the existence of weak symbols
+         * before using them. */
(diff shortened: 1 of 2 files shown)


host/sgx/elf.c
```diff
@@ -1909,7 +1909,8 @@ oe_result_t elf64_load_relocations(
     for (; p != end; p++)
     {
         uint64_t reloc_type = ELF64_R_TYPE(p->r_info);
-        if (reloc_type != R_X86_64_RELATIVE && reloc_type != R_X86_64_TPOFF64)
+        if (reloc_type != R_X86_64_RELATIVE && reloc_type != R_X86_64_TPOFF64
```
CodeMonkeyLeet, line 1912: Can you add a comment regarding the minimal support for `R_X86_64_GLOB_DAT` here? For example:
```
/* The enclave doesn't perform relocations on R_X86_64_GLOB_DAT, but we allow 
 * it for code that checks for the existence of weak symbols before using them. */
```

johnvenn: thanks for the suggestion. added in the code.

CodeMonkeyLeet (approved):
Thanks for submitting the fix for this, it's much appreciated!

radhikaj (approved):
Thanks!

PR #3583, opened 2020-09-24
Update CPUID exception handler to support more CPUID leaves.

lzha101:
Current default CPUID exception handler handles CPUID leaf 0, 1, 4, 7, which meets the requirement for OpenSSL, mbedTLS and others.  IPP crypto requires to access some of the extended CPUID leaves besides the supported leaves, such as 0x80000000, 0x80000001, 0x80000008. 
This patch adds three more CPUID leaves support in the default handler to make things easy for IPP crypto support (or other libraries that request CPUID extended leaves).

Signed-off-by: Zhang Lili <lili.z.zhang@intel.com>

### DIFF
--- a/enclave/core/sgx/cpuid.c
+++ b/enclave/core/sgx/cpuid.c
@@ -79,18 +79,18 @@ int oe_emulate_cpuid(uint64_t* rax, uint64_t* rbx, uint64_t* rcx, uint64_t* rdx)
     uint32_t cpuid_leaf = (*rax) & 0xFFFFFFFF;
     uint32_t cpuid_sub_leaf = (*rcx) & 0xFFFFFFFF;
 
-    if (cpuid_leaf < OE_CPUID_LEAF_COUNT &&
-        oe_is_emulated_cpuid_leaf(cpuid_leaf))
-    {
-        // For leaf 4 of cpuid, only subleaf of 0 is emulated
-        if ((cpuid_leaf == 4) && (cpuid_sub_leaf != 0))
-            return -1;
+    // emulate the cpuid leaves and find the corresponding index
+    uint32_t index = oe_get_emulated_cpuid_leaf_index(cpuid_leaf);
+    if (index == OE_UINT32_MAX)
+        return -1;
 
-        *rax = _cpuid_table[cpuid_leaf][OE_CPUID_RAX];
-        *rbx = _cpuid_table[cpuid_leaf][OE_CPUID_RBX];
-        *rcx = _cpuid_table[cpuid_leaf][OE_CPUID_RCX];
-        *rdx = _cpuid_table[cpuid_leaf][OE_CPUID_RDX];
-        return 0;
-    }
-    return -1;
+    // For leaf 4 of cpuid, only subleaf of 0 is emulated
+    if ((cpuid_leaf == 4) && (cpuid_sub_leaf != 0))
+        return -1;
+
+    *rax = _cpuid_table[index][OE_CPUID_RAX];
+    *rbx = _cpuid_table[index][OE_CPUID_RBX];
+    *rcx = _cpuid_table[index][OE_CPUID_RCX];
+    *rdx = _cpuid_table[index][OE_CPUID_RDX];
+    return 0;
 }
--- a/host/sgx/create.c
+++ b/host/sgx/create.c
@@ -392,7 +392,7 @@ oe_result_t oe_sgx_get_cpuid_table_ocall(
     for (unsigned int i = 0; i < OE_CPUID_LEAF_COUNT; i++)
     {
         oe_get_cpuid(
-            i,
+            supported_cpuid_leaves[i],
             subleaf,
             &leaf[OE_CPUID_RAX],
             &leaf[OE_CPUID_RBX],
--- a/include/openenclave/internal/cpuid.h
+++ b/include/openenclave/internal/cpuid.h
@@ -5,10 +5,10 @@
 #define _OE_CPUID_H
 
 #include <openenclave/bits/defs.h>
+#include <openenclave/bits/types.h>
 
 #define OE_CPUID_OPCODE 0xA20F
-#define OE_CPUID_LEAF_COUNT 8
-#define OE_CPUID_EXTENDED_CPUID_LEAF 0x80000000
+#define OE_CPUID_LEAF_COUNT 6 /* 0,1,4,7,0x80000000,0x80000001 */
 
 #define OE_CPUID_RAX 0
 #define OE_CPUID_RBX 1
@@ -20,17 +20,31 @@
 #define OE_CPUID_RDRAND_FEATURE 0x40000000u /* Leaf 1, subleaf 0, ECX */
 #define OE_CPUID_RDSEED_FEATURE 0x00040000u /* Leaf 7, subleaf 0, EBX */
 
+extern const uint32_t supported_cpuid_leaves[OE_CPUID_LEAF_COUNT];
+
 /**
- * The list of cpuid leafs that are emulated.
- * Currently 0, 1, 4, 7 leafs are emulated, consistent with Intel SDK.
- * Note: leaf 1 is used by mbedtls to determine aesni support.
- * The higher 8 bits of ebx for leaf 1 is the current processor id (initial APIC
+ * Get the leaf index for the emulated cpuid leaf
+ */
+OE_INLINE uint32_t oe_get_emulated_cpuid_leaf_index(uint32_t leaf)
+{
+    uint32_t i = 0;
+    while (i < OE_CPUID_LEAF_COUNT && leaf != supported_cpuid_leaves[i])
+        i++;
+    return i < OE_CPUID_LEAF_COUNT ? i : OE_UINT32_MAX;
+}
+
+/**
+ * Check if a cpuid leaf is emulated.
+ * Currently 0, 1, 4, 7, 0x80000000 and 0x80000001 leaves are
+ * emulated. Note: leaf 1 is used by mbedtls to determine aesni support. The
+ * higher 8 bits of ebx for leaf 1 is the current processor id (initial APIC
  * id). Since CPUID emulation returns cached values, this higher 8 bits of ebx
  * should not be relied upon for leaf 1.
  */
 OE_INLINE bool oe_is_emulated_cpuid_leaf(uint32_t leaf)
 {
-    return (leaf == 0) || (leaf == 1) || (leaf == 4) || (leaf == 7);
+    return (oe_get_emulated_cpuid_leaf_index(leaf) != OE_UINT32_MAX) ? true
+                                                                     : false;
 }
 
 #endif /* _OE_CPUID_H */
--- a/tests/VectorException/enc/sigill_handling.c
+++ b/tests/VectorException/enc/sigill_handling.c
@@ -6,6 +6,8 @@
 #include <openenclave/internal/print.h>
 #include "VectorException_t.h"
 
+#define OE_CPUID_TRACE_ENUM_LEAF 0x14
+
 // Wrapper over the CPUID instruction.
 void get_cpuid(
     unsigned int leaf,
@@ -164,12 +166,13 @@ int enc_test_sigill_handling(
     }
 
     // Test unsupported CPUID leaves
-    if (!test_unsupported_cpuid_leaf(OE_CPUID_LEAF_COUNT))
+    // Assume 0x14 is an unsupported CPUID leaf and oe_is_emulated_cpuid_leaf()
+    // returns false in this case
+    if (oe_is_emulated_cpuid_leaf(OE_CPUID_TRACE_ENUM_LEAF) != false)
     {
         return -1;
     }
-
-    if (!test_unsupported_cpuid_leaf(OE_CPUID_EXTENDED_CPUID_LEAF))
+    if (!test_unsupported_cpuid_leaf(OE_CPUID_TRACE_ENUM_LEAF))
     {
         return -1;
     }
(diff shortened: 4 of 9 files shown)


enclave/core/sgx/cpuid.c
```diff
@@ -10,6 +10,23 @@
 #include "platform_t.h"
 
 static uint32_t _cpuid_table[OE_CPUID_LEAF_COUNT][OE_CPUID_REG_COUNT];
+static const uint32_t _cpuid_leaves[OE_CPUID_LEAF_COUNT] = {0,
```
CodeMonkeyLeet, line 13: Since we are creating a mapping for supported leaves here, can this just be reduced to the set of emulated leaves only? It also simplifies the logic for `oe_is_emulated_cpuid_leaf` to have it simply iterate through the leaves in `_cpuid_leaves` instead of maintaining a separate bit of logic that can get out of sync.

host/sgx/create.c
```diff
@@ -374,6 +374,24 @@ static oe_result_t _add_data_pages(
 }
 
 #if !defined(OEHOSTMR)
+static const uint32_t _cpuid_leaves[OE_CPUID_LEAF_COUNT] = {0,
```
CodeMonkeyLeet, line 377: Since this is needed by host, enclave and tests, can you move this definition out to the common/cpuid.c source file as `oe_emulated_cpuid_leaves` instead of repeating the definition that needs to be kept in sync?

include/openenclave/internal/cpuid.h
```diff
@@ -7,7 +7,7 @@
 #include <openenclave/bits/defs.h>
 
 #define OE_CPUID_OPCODE 0xA20F
-#define OE_CPUID_LEAF_COUNT 8
+#define OE_CPUID_LEAF_COUNT 17 /* 0-7, 0x80000000-0x800000008*/
 #define OE_CPUID_EXTENDED_CPUID_LEAF 0x80000000
```
CodeMonkeyLeet, line 11: Remove this definition if it is no longer used (it's only reference seems to be in the test for sigill_handling.c)

tests/VectorException/enc/sigill_handling.c
```diff
@@ -169,18 +187,18 @@ int enc_test_sigill_handling(
         return -1;
     }
 
-    if (!test_unsupported_cpuid_leaf(OE_CPUID_EXTENDED_CPUID_LEAF))
+    if (!test_unsupported_cpuid_leaf(0x14))
```
CodeMonkeyLeet, line 190: Can you also add a sanity check that `oe_is_emulated_cpuid_leaf` returns false for `0x14`  and log that? It would help with debugging if the assumption that `0x14` is an unsupported leaf is violated in the future.

CodeMonkeyLeet (changes requested):
Thanks for the contribution! Could you elaborate more on what IPP needs the extended CPUID leaves introduced in this PR for? That would be valuable to document in the comments for the OE SDK emulated leaves in the header.

Also, specifically where crypto is taking a dependency on CPUID information from untrusted host, it's important to understand if there are security implications for this. For example, the OE SDK runtime enforces that AESNI and RDSEED/RDRAND must be provided or it fails enclave initialization to prevent the host from forcing the enclave to use less secure implementations in crypto.

host/sgx/create.c
lzha101: Thanks for providing the info. I thought about this when creating this PR, but not quite sure where I could place the variable.  Seems it should be more better to place under common/sgx/?

include/openenclave/internal/cpuid.h
dthaler: Unrelated question (ok to handle outside of this PR): is cpuid.h only for SGX? assuming so, seems like the file needs a rename or be moved under an sgx subdirectory.

enclave/core/sgx/cpuid.c
lzha101: Thanks for the comments. Fixed in the latest commit.

include/openenclave/internal/cpuid.h
lzha101: The unused definition has been removed. 
As to the cpuid.h,  based on current OE implementation, it seems the cpuid collection and usage are indeed for SGX only.  Not sure about your plan for OP-TEE,  so I keep the file as it is in the latest commit.

host/sgx/create.c
lzha101: Fixed in the latest commit by providing a cpuid.c under common/sgx folder.

enclave/core/sgx/cpuid.c
```diff
@@ -10,6 +10,7 @@
 #include "platform_t.h"
 
 static uint32_t _cpuid_table[OE_CPUID_LEAF_COUNT][OE_CPUID_REG_COUNT];
+extern const uint32_t supported_cpuid_leaves[OE_CPUID_LEAF_COUNT];
```
CodeMonkeyLeet, line 13: Is this extra declaration necessary? This file already includes openenclave/internal/cpuid.h which declares the same.

enclave/core/sgx/cpuid.c
```diff
@@ -78,19 +79,22 @@ int oe_emulate_cpuid(uint64_t* rax, uint64_t* rbx, uint64_t* rcx, uint64_t* rdx)
...
+    uint32_t index = 0;
 
-    if (cpuid_leaf < OE_CPUID_LEAF_COUNT &&
-        oe_is_emulated_cpuid_leaf(cpuid_leaf))
-    {
-        // For leaf 4 of cpuid, only subleaf of 0 is emulated
-        if ((cpuid_leaf == 4) && (cpuid_sub_leaf != 0))
-            return -1;
+    // emulate the cpuid leaves and find the corresponding index
+    while (index < OE_CPUID_LEAF_COUNT &&
```
CodeMonkeyLeet, line 85: Similarly, why reimplement `oe_is_emulated_cpuid_leaf` here instead of just using the inline from the cpuid.h header?

include/openenclave/internal/cpuid.h
```diff
@@ -21,16 +22,24 @@
 #define OE_CPUID_RDSEED_FEATURE 0x00040000u /* Leaf 7, subleaf 0, EBX */
 
 /**
- * The list of cpuid leafs that are emulated.
- * Currently 0, 1, 4, 7 leafs are emulated, consistent with Intel SDK.
- * Note: leaf 1 is used by mbedtls to determine aesni support.
- * The higher 8 bits of ebx for leaf 1 is the current processor id (initial APIC
+ * Checke if a cpuid leaf is emulated.
```
CodeMonkeyLeet, line 25:
```suggestion
 * Check if a cpuid leaf is emulated.
```

include/openenclave/internal/cpuid.h
```diff
@@ -21,16 +22,24 @@
...
  * id). Since CPUID emulation returns cached values, this higher 8 bits of ebx
  * should not be relied upon for leaf 1.
  */
+extern const uint32_t supported_cpuid_leaves[OE_CPUID_LEAF_COUNT];
 OE_INLINE bool oe_is_emulated_cpuid_leaf(uint32_t leaf)
 {
-    return (leaf == 0) || (leaf == 1) || (leaf == 4) || (leaf == 7);
+    // OE_CPUID_TRACE_ENUM_LEAF is used only in the exception test as an
+    // unsupported leaf
+    if (leaf == OE_CPUID_TRACE_ENUM_LEAF)
```
CodeMonkeyLeet, line 37: There should not need to be a special case to handle this. The point of testing an arbitrary leaf like 0x14 is to verify the negative test case, not that the implementation should implement special behavior for it. 

In fact, it would be better if `OE_CPUID_TRACE_ENUM_LEAF` is declared in the body of sigill_handling.c, which is the only place that it is used, instead of in the general header here.

tests/VectorException/enc/sigill_handling.c
```diff
@@ -6,6 +6,8 @@
 #include <openenclave/internal/print.h>
 #include "VectorException_t.h"
 
+extern const uint32_t supported_cpuid_leaves[OE_CPUID_LEAF_COUNT];
```
CodeMonkeyLeet, line 9: This file already includes openenclave/internal/cpuid.h that declares this.

tests/VectorException/host/host.c
```diff
@@ -13,6 +13,8 @@
 #define SKIP_RETURN_CODE 2
 
 static bool _was_ocall_called = false;
+extern const uint32_t supported_cpuid_leaves[OE_CPUID_LEAF_COUNT];
```
CodeMonkeyLeet, line 16: Again redundant because VectorException.edl includes openenclave/internal/cpuid.h which already declares this.

host/sgx/create.c
```diff
@@ -374,6 +374,8 @@ static oe_result_t _add_data_pages(
 }
 
 #if !defined(OEHOSTMR)
+extern const uint32_t supported_cpuid_leaves[OE_CPUID_LEAF_COUNT];
```
CodeMonkeyLeet, line 377: I'm not sure where this file is getting openenclave/internal/cpuid.h from, but `OE_CPUID_LEAF_COUNT` is defined there and so this file must already have included it for this definition to be valid, as well as the uses of `OE_CPUID_RAX` etc. below.

CodeMonkeyLeet (commented):
Thanks for the changes, they look good; just a couple of clean up comments to address.

include/openenclave/internal/cpuid.h
dthaler: Sounds like the simplest change (ok in a separate PR or in this PR, I have no preference) is to move cpuid.h down a level under internal/sgx

enclave/core/sgx/cpuid.c
lzha101: Oh, yes, I can remove the extern declaration in the C files. This reply is also suitable for the other comments below and related to this. I indeed firstly added the declarations in the C files and did some tests. And then refined the `oe_is_emulated_cpuid_leaf` function changing the cpuid.h, but forgot to remove the declarations in C files.  Thanks for catching this.

enclave/core/sgx/cpuid.c
lzha101: It is to get the index. `oe_is_emulated_cpuid_leaf` doesn't return the index. Or I can change the `oe_is_emulated_cpuid_leaf` interface to return the index.

CodeMonkeyLeet: I would suggest defining `oe_get_emulated_cpuid_leaf` which returns the index on success (and -1 or `OE_UINT32_MAX` on failure) in internal/cpuid.h, and then redefining `oe_is_emulated_cpuid_leaf` as an inline/macro wrapper around `oe_get_emulated_cpuid_leaf` that converts the result to a `bool`.

lzha101:
> Thanks for the contribution! Could you elaborate more on what IPP needs the extended CPUID leaves introduced in this PR for? That would be valuable to document in the comments for the OE SDK emulated leaves in the header.
> 
> Also, specifically where crypto is taking a dependency on CPUID information from untrusted host, it's important to understand if there are security implications for this. For example, the OE SDK runtime enforces that AESNI and RDSEED/RDRAND must be provided or it fails enclave initialization to prevent the host from forcing the enclave to use less secure implementations in crypto.

The CPUID leaves are required in [ipp-crypto](https://github.com/intel/ipp-crypto/), specifically in the initialization function [cpGetFeatures()](https://github.com/intel/ipp-crypto/blob/develop/sources/ippcp/cpinit.c#L104). You can see CPUID with extended leaves are called in lines: 
https://github.com/intel/ipp-crypto/blob/develop/sources/ippcp/cpinit.c#L114
https://github.com/intel/ipp-crypto/blob/develop/sources/ippcp/cpinit.c#L197

`ipp-crypto` is required in Intel(R) AEs.  So we need to support these leaves.  Actually SGX requires more CPUID leaves besides what I added here for the optimization initialization, if check the cpuid info initialization in SGX:
https://github.com/intel/linux-sgx/blob/master/psw/urts/cpu_features_ext.cpp#L51
As not all of these are required to ipp-crypto, I just cut down the unnecessary and only keep 0x80000000 and 0x80000001.
From Intel SDM:
INPUT EAX = 80000000H: Returns CPUID’s Highest Value for Extended Processor Information

INPUT EAX = 80000001H:  
        OUTPUT ECX  would indicates if LAHF/SAHF, LZCNT, PREFETCHW instructions are available. 
        OUTPUT EDX indicates if  SYSCALL/SYSRET, RDTSCP are available. 

As to security, these instructions are not related to crypto. And if any of the instructions is not supported by platform but the cpuid shows it is supported ( by hacking ), the execution would failed with an illegal instruction exception.

CodeMonkeyLeet:
@lzha101 Thanks for the links and the follow-up, it looks like that IPP:
- Only uses the 80000000H leaf to determine if the 80000001H extended leaf is supported
- Reads the 80000001H leaf to set the CPU feature mask as supporting PREFETCHW specifically. 

I think that's sufficient confirmation for the purposes of this PR, but I would like to have the Intel security folks (maybe @markg-intel) comment on whether the IPP library used in an enclave context needs to enforce the presence of additional CPUID features that OE SDK does not today:

- The `cpGetFeatures` method checks for a whole lot of CPUID features that may have crypto impact that OE does not enforce must be present for an enclave, for example, PCLMULQDQ. Are there any features where if absent, will cause the library to use software behavior that may be less secure? (e.g. falls back to a non-constant time/memory implementation vulnerable to side channels).
- A simple code search shows that the only direct call chain into `cpGetFeatures` is from the public `ippcpGetCpuFeatures` method, and this is determined by non-crypto related flags like PREFETCHW as well. As far as I can tell, IPP tries to group sets of these features into different base platforms, but it's unclear to me whether any of the crypto implementations modify their behavior based on these platform definitions. If they do, then are there any concerns that a malicious host can trick IPP in an enclave to use implementations intended for "older" platforms that are less secure?

anakrish (approved):
LGTM

CodeMonkeyLeet (approved):
LGTM, thanks for the contribution!

PR #3588, opened 2020-09-24
Make fine-grained OpenSSL unsupported headers to avoid inclusion errors

### LINKED ISSUE #3587: Unexpected compilation errors caused by OpenSSL unsupported.h
Currently we put OpenSSL APIs (across multiple headers) that we want to disable in the same header, `unsupported.h`.  The  `unsupported.h` explicitly includes the corresponding OpenSSL headers as follows to enforce the inclusion order.

```
#include <openssl/crypto.h>
#include <openssl/ssl.h>
#include <openssl/x509_vfy.h>
```

In addition, we enforce the these OpenSSL headers to include `unsupported.h` so that we can correctly trigger the compilation errors when users try to use any of the disabled APIs. However, the inclusion of multiple headers could cause unexpected compilation errors. See the following example:

If an application (indirectly or directly) includes `evp.h` before any of above three headers, then the following error will be triggered.

```
In file included from /home/ssh_office/workspace/oe-ms2/common/crypto/openssl/hmac.c:7:
In file included from /home/ssh_office/workspace/oe-ms2/build/output/include/openenclave/openssl/hmac.h:15:
In file included from /home/ssh_office/workspace/oe-ms2/build/output/include/openenclave/openssl/evp.h:16:
In file included from /home/ssh_office/workspace/oe-ms2/build/output/include/openenclave/openssl/bio.h:20:
In file included from /home/ssh_office/workspace/oe-ms2/build/output/include/openenclave/openssl/crypto.h:446:
In file included from /home/ssh_office/workspace/oe-ms2/build/output/include/openenclave/openssl/unsupported.h:29:
In file included from /home/ssh_office/workspace/oe-ms2/build/output/include/openenclave/openssl/ssl.h:25:
/home/ssh_office/workspace/oe-ms2/build/output/include/openenclave/openssl/pem.h:234:19: error: unknown type name 'EVP_CIPHER_INFO'; did you mean
      'EVP_CIPHER_CTX'?
int PEM_do_header(EVP_CIPHER_INFO *cipher, unsigned char *data, long *len,
                  ^~~~~~~~~~~~~~~
                  EVP_CIPHER_CTX
```

The reason for this is that `ssl.h` depends on some type definitions in `evp.h`. However, the `unsupported.h` forces the inclusion of `ssl.h` before the point of those type definition in the `evp.h`.

mingweishih:
This PR fixes #3587, which breaks the `unsupported.h` into the fine-grained headers, and appends the `#include <openssl/"header"_unsupported.h>` to the `header.h` only. This avoids the errors caused by that `unsupported.h` forces the inclusion of multiple headers.


CodeMonkeyLeet (commented):
Changes look fine, but I'm unclear on the explanation of the repro in #3587:
- unsupported.h is not included in evp.h
- evp.h is included before ssl.h by the app for the repro
- ssl.h depends on definitions in evp.h
- unsupported.h is appended to the end of ssl.h
- unsupported.h includes crypto.h, ssl.h and x509_vfy.h

How does that lead to the assertion:
> unsupported.h forces the inclusion of ssl.h before the point of those type definition in the evp.h

mingweishih:
> Changes look fine, but I'm unclear on the explanation of the repro in #3587:
> 
> * unsupported.h is not included in evp.h
> * evp.h is included before ssl.h by the app for the repro
> * ssl.h depends on definitions in evp.h
> * unsupported.h is appended to the end of ssl.h
> * unsupported.h includes crypto.h, ssl.h and x509_vfy.h
> 
> How does that lead to the assertion:
> 
> > unsupported.h forces the inclusion of ssl.h before the point of those type definition in the evp.h

The problem is that

evp.h
```
line 16 # include <openssl/bio.h>
...

line 396 typedef struct evp_cipher_info_st {
line 397    const EVP_CIPHER *cipher;
line 398    unsigned char iv[EVP_MAX_IV_LENGTH];
line 399} EVP_CIPHER_INFO;
```

Before reaching line 396, the inclusion of bio.h in line 16 will indirectly force the inclusion of ssl.h (bio.h -> crypto.h -> unsupported.h > ssl.h). This is where the error occurs.

CodeMonkeyLeet:
@mingweishih Thanks for the elaboration, that makes sense. Given that we are back to the model of injecting the unsupported.h header into the individual headers, does it still need to include the headers anymore? That would effectively break the loop that creates the issue described. 

I'm also trying to understand if there is an issue with the way the original musl/deprecations.h was implemented, and if that needs to be changed.

mingweishih:
> @mingweishih Thanks for the elaboration, that makes sense. Given that we are back to the model of injecting the unsupported.h header into the individual headers, does it still need to include the headers anymore? That would effectively break the loop that creates the issue described.
> 

With this PR, that is true that `*_unsupported.h` does not need to include headers anymore. However, in the original approach (i.e., having a single `unsupported.h`), the inclusion is still needed because it tries to disable APIs from multiple headers.

For example, `unsupported.h` disables `OPENSSL_INIT_LOAD_CONFIG` from `crypto.h` and `X509_LOOKUP_hash_dir` from `x509_vfy.h`. When a user includes `crypto.h` only, it will bring in `unsupported.h`. This will cause compilation errors because of the definition of `X509_LOOKUP_hash_dir` (used by ` unsupported.h`) is not found.

> I'm also trying to understand if there is an issue with the way the original musl/deprecations.h was implemented, and if that needs to be changed.

The current `musl/deprecations.h` only tries to disable APIs from a single header (pthread.h) so does not suffer from this issue.

3rdparty/openssl/include/crypto_unsupported.h
```diff
@@ -0,0 +1,42 @@
...
+** <openssl/crypto.h>
+**
+**==============================================================================
+*/
+
+/*
+ * The following macro allows users to configure an OpenSSL application (via
+ * OPENSSL_init_crypto or OPENSSL_init_ssl) to lookup a configuration file from
+ * the host filesystem, which is not trusted in OE's threat model, and therefore
+ * are recommended not to use inside an enclave.
```
radhikaj, line 25:
```suggestion
 * is recommended not to use inside an enclave.
```

3rdparty/openssl/include/crypto_unsupported.h
```diff
@@ -0,0 +1,42 @@
...
+    return 0x00000040L;
+}
+#define OPENSSL_INIT_LOAD_CONFIG \
+    OE_OPENSSL_RAISE_ERROR(OPENSSL_INIT_LOAD_CONFIG)
+#endif
+
+OE_OPENSSL_EXTERN_C_END
+
+#endif /* !defined(OE_OPENSSL_SUPPRESS_UNSUPPORTED) */
+#endif /* _OE_OPENSSL_CRYPTO_UNSUPPORTED_H */
```
radhikaj, line 42: >_OE_OPENSSL_CRYPTO_UNSUPPORTED_H [](start = 10, length = 32)

I think I am misunderstanding something. It appears there is a script append-unsupported to mark certain headers as unsupported. Why do we need additional macros for init_load in particular.

radhikaj (approved):

CodeMonkeyLeet (approved):
I think the approach is fine, although we may need to revisit MUSL approach since it's injecting a string header specific deprecations into apparently all of the MUSL headers.

PR #3596, opened 2020-09-27
Fix calculation error of sgxenclave TLS size

volcano0dr:
When the _calculate_enclave_size function calculates enclave_size, the tls size is calculated based on the default 1 page size, rather than the actual .tdata and .tbss size. This may cause the oe_sgx_enclave_load_data function to fail when the enclave is loaded, because the number of added pages exceeds the enclave_size size.

For example: modify samplew/helloworld code
helloworld.conf:

```
Debug=1
NumHeapPages=2360
NumStackPages=1024
NumTCS=1
ProductID=1
SecurityVersion=1
```

Declare tls data in enclave
enc.c:
```
#define TLS_SIZE (3848 + 1)
static __thread char tls[TLS_SIZE]
```

When loading enclave, calling oe_sgx_enclave_load_data failed. Error message:
enclave_load_data failed (addr=0x3c000000, prot=0x3 err=0xd) (oe_result_t=OE_PLATFORM_ERROR)

If TLS_SIZE = 3848, loading the enclave will succeed, and the TLS size is exactly 1 page size.

So the _calculate_enclave_size function should calculate enclave_size according to the actual TLS data size.

Signed-off-by: volcano <volcano_dr@163.com>

### DIFF
--- a/host/sgx/create.c
+++ b/host/sgx/create.c
@@ -38,6 +38,7 @@ static char* get_fullpath(const char* path)
 #include <openenclave/bits/sgx/sgxtypes.h>
 #include <openenclave/host.h>
 #include <openenclave/internal/calls.h>
+#include <openenclave/internal/constants_x64.h>
 #include <openenclave/internal/debugrt/host.h>
 #include <openenclave/internal/eeid.h>
 #include <openenclave/internal/load.h>
@@ -163,6 +164,7 @@ static oe_result_t _add_control_pages(
     uint64_t enclave_addr,
     uint64_t enclave_size,
     uint64_t entry,
+    size_t tls_page_count,
     uint64_t* vaddr,
     oe_enclave_t* enclave)
 {
@@ -227,8 +229,7 @@ static oe_result_t _add_control_pages(
          * segment.
          */
         tcs->fsbase =
-            *vaddr +
-            (context->num_tls_pages + OE_SGX_NUM_CONTROL_PAGES) * OE_PAGE_SIZE;
+            *vaddr + (tls_page_count + OE_SGX_TCS_CONTROL_PAGES) * OE_PAGE_SIZE;
 
         /* The existing Windows SGX enclave debugger finds the start of the
          * thread data by assuming that it is located at the start of the GS
@@ -268,8 +269,9 @@ static oe_result_t _add_control_pages(
     (*vaddr) += OE_PAGE_SIZE;
 
     /* Add blank pages (for either FS segment or GS segment) */
-    OE_CHECK(_add_filled_pages(
-        context, enclave_addr, vaddr, context->num_tls_pages, 0, true));
+    if (tls_page_count)
+        OE_CHECK(_add_filled_pages(
+            context, enclave_addr, vaddr, tls_page_count, 0, true));
 
     /* Add one page for thread-specific data (TSD) slots */
     OE_CHECK(_add_filled_pages(context, enclave_addr, vaddr, 1, 0, true));
@@ -285,6 +287,7 @@ done:
 
 static oe_result_t _calculate_enclave_size(
     size_t image_size,
+    size_t tls_page_count,
     const oe_sgx_enclave_properties_t* props,
     size_t* loaded_enclave_pages_size,
     size_t* enclave_size)
@@ -293,6 +296,7 @@ static oe_result_t _calculate_enclave_size(
     oe_result_t result = OE_UNEXPECTED;
     size_t heap_size;
     size_t stack_size;
+    size_t tls_size;
     size_t control_size;
     const oe_enclave_size_settings_t* size_settings;
 
@@ -310,13 +314,17 @@ static oe_result_t _calculate_enclave_size(
                  + (size_settings->num_stack_pages * OE_PAGE_SIZE) +
                  OE_PAGE_SIZE; // guard page
 
-    /* Compute the control size in bytes (6 pages total) */
-    control_size = 6 * OE_PAGE_SIZE;
+    /* Compute size of the TLS */
+    tls_size = tls_page_count * OE_PAGE_SIZE;
+
+    /* Compute the control size in bytes (5 pages total) */
+    control_size = (OE_SGX_TCS_CONTROL_PAGES + OE_SGX_TCS_THREAD_DATA_PAGES) *
+                   OE_PAGE_SIZE;
 
     /* Compute end of the enclave */
     *loaded_enclave_pages_size =
         image_size + heap_size +
-        (size_settings->num_tcs * (stack_size + control_size));
+        (size_settings->num_tcs * (stack_size + tls_size + control_size));
 
     if (enclave_size)
     {
@@ -338,6 +346,7 @@ static oe_result_t _add_data_pages(
     oe_enclave_t* enclave,
     const oe_sgx_enclave_properties_t* props,
     uint64_t entry,
+    size_t tls_page_count,
     uint64_t* vaddr)
 
 {
@@ -364,7 +373,13 @@ static oe_result_t _add_data_pages(
 
         /* Add the "control" pages */
         OE_CHECK(_add_control_pages(
-            context, enclave->addr, enclave->size, entry, vaddr, enclave));
+            context,
+            enclave->addr,
+            enclave->size,
+            entry,
+            tls_page_count,
+            vaddr,
+            enclave));
     }
 
     result = OE_OK;
@@ -585,6 +600,7 @@ static oe_result_t _add_eeid_marker_page(
     oe_sgx_load_context_t* context,
     oe_enclave_t* enclave,
     size_t image_size,
+    size_t tls_page_count,
     uint64_t entry_point,
     oe_sgx_enclave_properties_t* props,
     uint64_t* vaddr)
@@ -604,6 +620,7 @@ static oe_result_t _add_eeid_marker_page(
         oe_sha256_save(hctx, eeid->hash_state.H, eeid->hash_state.N);
         eeid->entry_point = entry_point;
         eeid->vaddr = *vaddr;
+        eeid->tls_page_count = tls_page_count;
         eeid->signature_size = sizeof(sgx_sigstruct_t);
         memcpy(
             eeid->data + eeid->data_size,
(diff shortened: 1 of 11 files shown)


volcano0dr:
The PR has not been revised completely.
There are two functions related to EEID: oe_remeasure_memory_pagehe, _eeid_patch_memory.
These two functions are also based on the size of 1 page to calculate the TLS size.

CodeMonkeyLeet (commented):
Thanks for finding this issue and submitting a fix for it! It is very much appreciated!

@anakrish It looks like #3339 didn't completely address the dynamic TLS sizing issue. In particular, the determination of `num_tls_pages` in `_patch_elf_image` should have instead happened earlier in `_calculate_enclave_size` and the results passed downstream to the image patching for consistency. Additionally, we should expand the test case to handle the regression testing for this issue. 

@wintersteiger For the EEID methods, can you take a look at #3339 and update the expectations of `oe_remeasure_memory_pages` and `_eeid_patch_memory` accordingly?

volcano0dr:
Fixed the `oe_remeasure_memory_pagehe` and `_eeid_patch_memory` functions related to TLS size.
330d9ba8b09f2c6f685f0de1ded08f76ef0f9c27

anakrish:
bors try

anakrish:
@volcano0dr Can you run `scripts/format-code` and push the changes. CI is failing due to unformatted file.

volcano0dr:
> @volcano0dr Can you run `scripts/format-code` and push the changes. CI is failing due to unformatted file.
@anakrish Format the code with the format-code script
8debe938f87e3a7e19debb3d562d67b8a950272b

wintersteiger:
I PR'd a couple fixes for the EEID-related code in https://github.com/volcano0dr/openenclave/pull/1.

include/openenclave/bits/eeid.h
```diff
@@ -51,6 +51,9 @@ typedef struct _oe_eeid
     /** Entry point of the image. */
     uint64_t entry_point;
 
+    /* Number of pages needed for thread-local data */
+    uint64_t num_tls_pages;
```
wintersteiger, line 55: I think `num_tls_pages` should go into `size_settings`, but there may have been previous discussions about that which I'm not aware of.

host/sgx/create.c
```diff
@@ -310,13 +312,26 @@ static oe_result_t _calculate_enclave_size(
...
+    if (image->elf.tbss_size)
+    {
+        tls_size += oe_round_up_to_multiple(
+            image->elf.tbss_size, image->elf.tbss_align);
+    }
+
+    tls_size = oe_round_up_to_multiple(tls_size, OE_PAGE_SIZE);
+
+    /* Compute the control size in bytes (5 pages total) */
+    control_size = 5 * OE_PAGE_SIZE;
```
wintersteiger, line 329: I'm not sure whether the `5` here is related to `OE_SGX_NUM_CONTROL_PAGES+1`; if it is, I think it would be nicer to use the macro.

wintersteiger (commented):
I added a couple non-blocking comments. The EEID changes should be merged into this one first to make sure the tests keep passing.

host/sgx/create.c
volcano0dr: Yes, it’s nicer to use macros.

sgxload.h:
OE_SGX_NUM_CONTROL_PAGES:  1 page TCS + 2 pages SSA + 1 page GAUARD
eeid.h:
OE_SGX_TCS_CONTROL_PAGES: 1 page TCS + 2 pages SSA + 1 page GAUARD + 1 page TD

volcano0dr: commit code: 3b14072cc1450cc9be56b86cfa38905caf43e12a

wintersteiger: Thanks for the fixes! I'm afraid I don't quite understand why those two are different, in my mind they should really be the same. From the places where they are used, it looks like `OE_SGX_TCS_CONTROL_PAGES == OE_SGX_NUM_CONTROL_PAGES + 1`? Now that there's `OE_SGX_NUM_CONTROL_PAGES` I think we should remove `OE_SGX_TCS_CONTROL_PAGES` from `eeid.h`.

include/openenclave/bits/eeid.h
volcano0dr: I think it is inappropriate to put num_tls_page into size_settings, because size_settings is set by the user when building an enclave or creating an enclave, but num_tls_page is not needed.

host/sgx/create.c
volcano0dr: Remove `OE_SGX_NUM_CONTROL_PAGES,` and `#define OE_SGX_TCS_CONTROL_PAGES 4 ` is defined in openenclave/internal/sgx/sgxproperties.h, and `#include <openenclave/internal/properties.h>` in eeid.c. What do you think of it.

CodeMonkeyLeet: I would suggest adding the definitions to include/openenclave/internal/constants_x64.h instead, under the category of _Open Enclave layout constant definitions_. 

Regarding why the values are different, the original `OE_SGX_NUM_CONTROL_PAGES` was defined to calculate how to get to the TD page offset in both it's use in `_add_control_pages` and `_patch`, whereas EEID is concerned about getting the total TCS size, which would need to include the TD page itself. 

I would suggest adding appropriate comments in the header definition to disambiguate this, for example:

```c
// SGX control pages, excluding thread data and local storage:
// 1 TCS page + 2 SSA pages + 1 guard page
#define OE_SGX_TCS_CONTROL_PAGES 4

#define OE_SGX_TCS_THREAD_DATA_PAGES 1
```

volcano0dr: Thank you for your comments, the code has been modified according to the above suggestions, please review again. 
2a2808148d257e127f48ee1a3d8be8730f593be8

CodeMonkeyLeet (commented):
The changes related to the OE_SGX_TCS_CONTROL_PAGES look good, I still have one outstanding comment though, see https://github.com/openenclave/openenclave/pull/3596#discussion_r498548076

host/sgx/create.c
```diff
@@ -284,6 +284,7 @@ static oe_result_t _add_control_pages(
 }
 
 static oe_result_t _calculate_enclave_size(
+    const oe_enclave_image_t* image,
```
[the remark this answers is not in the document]
volcano0dr: Thank you very much for your help, the code has been modified according to your suggestions, please review again, thank you.
b73333a5aa510c1d50f1f9a5afe1459ff7522e86

host/sgx/create.c
```diff
@@ -163,6 +164,7 @@ static oe_result_t _add_control_pages(
     uint64_t enclave_addr,
     uint64_t enclave_size,
     uint64_t entry,
+    size_t num_tls_pages,
```
dthaler, line 167: nit: style guidance is to avoid abbreviations
```suggestion
    size_t tls_page_count,
```

https://github.com/openenclave/openenclave/blob/master/docs/Community/DevelopmentGuide.md says:

> DON'T use abbreviations unless they are already well-known terms known by users (e.g., "app", "info"), or are already required for use by developers (e.g., "min", "max", "args"). Examples of bad use would be `num_widgets` instead of `widget_count`, and `opt_widgets` instead of `option_widgets` or `optional_widgets`.

host/sgx/loadelf.c
```diff
@@ -454,6 +455,29 @@ static oe_result_t _calculate_size(
     return OE_OK;
 }
 
+static oe_result_t _get_num_tls_pages(
+    const oe_enclave_image_t* image,
+    size_t* num_tls_pages)
```
dthaler, line 460:
```suggestion
    size_t* tls_page_count)
```

host/sgx/loadelf.c
```diff
@@ -677,12 +701,12 @@ static oe_result_t _set_uint64_t_dynamic_symbol_value(
 static oe_result_t _patch_elf_image(
     oe_enclave_elf_image_t* image,
     oe_sgx_load_context_t* context,
-    size_t enclave_size)
+    size_t enclave_size,
+    size_t num_tls_pages)
```
dthaler, line 705:
```suggestion
    size_t tls_page_count)
```

include/openenclave/bits/eeid.h
dthaler, line 55:
```suggestion
    uint64_t tls_page_count;
```

include/openenclave/internal/load.h
```diff
@@ -92,6 +92,10 @@ struct _oe_enclave_image
     oe_result_t (
         *calculate_size)(const oe_enclave_image_t* image, size_t* image_size);
 
+    oe_result_t (*get_num_tls_pages)(
+        const oe_enclave_image_t* image,
+        size_t* num_tls_pages);
```
dthaler, line 97:
```suggestion
    oe_result_t (*get_tls_page_count)(
        const oe_enclave_image_t* image,
        size_t* tls_page_count);
```

host/sgx/create.c
```diff
@@ -822,7 +848,13 @@ oe_result_t oe_sgx_build_enclave(
 
 #ifdef OE_WITH_EXPERIMENTAL_EEID
     OE_CHECK(_add_eeid_marker_page(
-        context, enclave, image_size, oeimage.elf.entry_rva, &props, &vaddr));
+        &oeimage,
+        context,
+        enclave,
+        image_size,
+        oeimage.elf.entry_rva,
```
CodeMonkeyLeet, line 855: @volcano0dr looks like a bit of leftover code from a previous iteration. Once you fix this and Dave's comments, I can kick off a CI run, thanks!

```suggestion
        context,
        enclave,
        image_size,
        oeimage.elf.entry_rva,
        num_tls_pages,
```

volcano0dr: This error has been corrected and the code has been modified according to Dave's comment.
53dd6448d1555101d2fd6a3ec0dc17e4c1fa78ef

dthaler (approved):

CodeMonkeyLeet:
bors try

CodeMonkeyLeet (commented):
There are still build breaks to be addressed, see comments inline.

When you are done with the fixes, could you also squash all the commits per the dev guidelines as we prepare to merge the changes?

host/sgx/create.c
```diff
@@ -822,7 +848,14 @@ oe_result_t oe_sgx_build_enclave(
 
 #ifdef OE_WITH_EXPERIMENTAL_EEID
     OE_CHECK(_add_eeid_marker_page(
-        context, enclave, image_size, oeimage.elf.entry_rva, &props, &vaddr));
+        &oeimage,
```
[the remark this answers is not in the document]
volcano0dr: Thank you again for your help. The code has been fixed and submitted as required.
3fddb63854dc16d605e0f3cd97960c6c6d0174e2

anakrish:
bors try

CodeMonkeyLeet (approved):
LGTM! Thanks again for the contribution, it's much appreciated!

PR #3599, opened 2020-09-28
Add design doc to enable debugmalloc for end users.

nonpolarity:
Proposal to make debugmalloc available to end users.

### DIFF
--- a/docs/DesignDocs/Enabledebugmalloc.md
+++ b/docs/DesignDocs/Enabledebugmalloc.md
@@ -1,0 +1,290 @@
+Proposal to enable debug malloc for end users
+===
+
+# Motivation
+
+Debug malloc is an internal tool that helps identifying memory leaks in
+enclaves. It is turned on by default in the debug builds of the OE SDK.
+
+The main logic of debug malloc is as follows.
+
+- Whenever memory is allocated, debug malloc captures the callstack and
+associates it with the allocated memory. Additionally, debug malloc sets the
+contents of the allocated memory to zero.
+
+- Whenever memory is freed, it clears the contents to a fixed pattern to help
+detect access of freed memory. It also frees the associated callstack
+information.
+
+- When an enclave terminates, debug malloc prints information about leaks---those objects that have been allocated,
+  but not freed.
+
+Debug malloc is performance intensive and is therefore not compiled into the SDK release.
+
+There have been requests from users that debug malloc be made available so that they can
+identify memory leaks within an enclave. Commonly used memory detection tools like Valgrind do not
+work within enclaves since they cannot access enclave memory.
+
+According to the user's developing experience, there are two major problems
+which the memory leaking detection can help to address:
+
+- A developer wants to track all memory allocations throughout the execution of an enclave
+  and to determine if any of these allocations were leaked.
+  We name it "global detection".
+
+- A developer wants to track memory allocations in a specific part of the enclave's
+  execution that he or she can specify at the source code level. We name it "local detection".
+
+Open Enclave already provides some support for "global detection", so this
+document will focus on the design for supporting "local detection."
+
+# User Experience
+
+## Global detection
+
+Debug malloc will be shipped to the users as a separate library
+oelibdebugmalloc.a, then there are two options to enable the global detection.
+
+### Global default option
+
+The "USE_DEBUG_MALLOC" define controls the global default. Now we already set "USE_DEBUG_MALLOC"
+to "ON" in debug mode by default. Users can modify this default option
+according to the need, such as changing it to "ON" in release mode or changing it
+to "OFF" in debug mode.
+
+Please note that this option only works if you are building OE from source
+code that this option is only valid in the OE project and not exported.
+Additional, as a global option it will add the "-loedebugmalloc" option to all
+the enclave targets in the OE project. If you are not building OE from source
+or just want to apply this library to specific enclaves, please check the next
+"custom option".
+
+### Custom option
+
+Since oelibdebugmalloc.a is a separate library, users can turn on debug malloc
+per enclave while the default option is "OFF".
+
+To use debug malloc, users will have to link oelibdebugmalloc.a into their
+enclaves explicitly. This can be done by
+```
+      cmake:
+      target_link_libraries(user-enclave oelibc oedebugmalloc oeenclave)
+
+      make:
+      LDFLAGS=$(shell pkg-config oeenclave-$(COMPILER) --libs)
+      LDFLAGS_CUSTOM=$(subst -loelibc, -loedebugmalloc -loelibc, $(LDFLAGS))
+      $(CC) -o user-enclave user-enc.o $(LDFLAGS_CUSTOM)
+
+```
+
+For the global detection, if there is any memory leak during the enclave's
+life cycle, an error code "OE_MEMORY_LEAK" will be raised during the enclave
+termination. This means it not only checks the users' enclave application,
+but also covers the SDK's allocations on the enclave side.
+
+## Local detection
+
+Currently, debug malloc prints leaks only during enclave termination.
+However, users would like to isolate and track specific memory allocations.
+To achieve this, the following APIs are provided in a new published header `debugmalloc.h`.
+
+- `oe_result_t oe_debug_malloc_tracking_start(void)`
+
+  Starts tracking allocations that will be displayed in report. Subsequent
+  calls to start without a preceding stop will be ignored and return value is
+  "OE_UNEXPECTED".
+
+- `oe_result_t oe_debug_malloc_tracking_stop(void)`
+
+  Stops tracking allocations that will be displayed in report. Multiple calls
+  to stop without another start are treated as no-ops and return value is
+  "OE_UNEXPECTED".
+
+- `    oe_result_t oe_debug_malloc_tracking_report(
+        uint64_t* out_object_count,
+        char** report)
+  `
+
+  Report the leaks, including the number and details of allocations that have
+  not been freed since start was called. If called after stop, it reports the
+  set of allocations that were not freed while tracking is turned on, even if
+  they might have been freed between the calls to stop and report. If called
+  before the earliest start, the report includes nothing.
+
+Please note that the timing and sequences of calling of these functions is
+important:
+
+- The starting and stopping tracking must cover the coupled malloc/free
+operations, which means start tracking before malloc and also stop tracking
+after free, or the report will definitely trigger false alarms that some
+reported objects actually are freed after stopping tracking.
+
+- oe_debug_malloc_tracking_stop() can only been called after calling
+oe_debug_malloc_tracking_start(), and oe_debug_malloc_tracking_start() can be
+called at the beginning or after calling oe_debug_malloc_tracking_stop(). While
+the calling sequence following this rule, the return values are OE_OK. Return
+value "OE_UNEXPECTED" means the sequence is not expected.
+
+Regardless the execution and output of these state switching functions,
+oe_debug_malloc_tracking_report() covers all the leaks happened during the
+tracking state cumulatively.
+
+```
+      stop()  |--------------|  start() |--------------|  start()
+   |--------->|              |--------->|              |----------|
+   |          | not tracking |          |   tracking   |          |
+   |----------|              |<---------|              |<---------|
+              |--------------|  stop()  |--------------|
+```
+
+- oe_debug_malloc_tracking_start() and oe_debug_malloc_tracking_stop() can be
+called many times:
+
+```
+  |----code1--|-----------|----code2--|--|----code3--|
+  start1->stop1->report1->start2->stop2->start3->stop3->report3
+```
+
+In this case report1 only covers code1, and report3 covers code1, code2 and
+code3.
+
+Please also note that these public APIs only affect the result of local
+detection. The result of global detection is independent of these APIs. While
+the calling of these APIs follow the rules, the result of local detection is
+supposed to be the subset of result of global detection.
+
+# Implementation
+
+The debug malloc employs a linked list to make a record of all memory operations.
+So, for the local detection, we can make use of the fulfilled record system.
+
+```c
+struct header
+{
+    /* Contains HEADER_MAGIC1 */
+    uint64_t magic1;
+
+    /* Headers are kept on a doubly-linked list */
+    header_t* next;
+    header_t* prev;
+
+    /* The alignment passed to memalign() or zero */
+    uint64_t alignment;
+
+    /* Size of user memory */
+    size_t size;
+
+    /* Return addresses obtained by oe_backtrace() */
+    void* addrs[OE_BACKTRACE_MAX];
+    uint64_t num_addrs;
+
+    /* Option if current object is tracked */
+    bool local_tracking;
+
+    /* Padding to make header a multiple of 16 */
+    uint8_t padding[7];
+
+    /* Contains HEADER_MAGIC2 */
+    uint64_t magic2;
+
+    /* User data */
+    uint8_t data[];
+};
+```
+
+We also add a variable
+```c
+bool oe_use_debug_malloc_tracking = false;
+```
+which will be changed by oe_debug_malloc_tracking_start() and
+oe_debug_malloc_tracking_stop(). While doing the allocation, the objective's
+`local_tracking` is marked as `oe_use_debug_malloc_tracking`, then
+oe_debug_malloc_tracking_report() can get the number of all the not yet freed
+objects.
+
+# Sample
+
+```c
+#include <openenclave/corelibc/stdlib.h>
+#include <openenclave/debugmalloc.h>
+#include <stdio.h>
+
+// Include the trusted debugmalloc header that is generated
+// during the build. This file is generated by calling the
+// sdk tool oeedger8r against the debugmalloc.edl file.
+#include "debugmalloc_t.h"
+
+// This is the function that the host calls. It prints
+// a message in the enclave before calling back out to
+// the host to print a message from there too.
+void enclave_hello()
+{
+    void* buf1, * buf2, * buf3;
+    uint64_t num;
+    char* report;
+
+    oe_assert(oe_debug_malloc_tracking_start() == OE_OK);
+    buf1 = oe_malloc(4096);
+    oe_assert(oe_debug_malloc_tracking_stop() == OE_OK);
+    if (oe_debug_malloc_tracking_report(&num, &report) == OE_OK)
+    {
+        printf("There are %d un-freed objects reported:\n%s\n", num, report);
+    }
+
+    oe_assert(oe_debug_malloc_tracking_start() == OE_OK);
+    buf2 = oe_malloc(4096);
+    oe_assert(oe_debug_malloc_tracking_stop() == OE_OK);
+
+    oe_assert(oe_debug_malloc_tracking_start() == OE_OK);
+    buf3 = oe_malloc(4096);
+    oe_assert(oe_debug_malloc_tracking_stop() == OE_OK);
+    if (oe_debug_malloc_tracking_report(&num, &report) == OE_OK)
+    {
+        printf("There are %d un-freed objects reported:\n%s\n", num, report);
+    }
+
+    oe_free(buf1);
+    oe_free(buf2);
+    oe_free(buf3);
+    oe_free(report);
+
+    // Call back into the host
+    oe_result_t result = host_hello();
+    if (result != OE_OK)
+    {
+        fprintf(
+            stderr,
+            "Call to host_hello failed: result=%u (%s)\n",
+            result,
+            oe_result_str(result));
+    }
+}
+```
+For the local detection, the expected output result are 1 and 3 leaks for each
+report, respectively.
+
+For the global detection, Since this sample has no memory leak at the point of
+enclave termination, the result is "OE_OK". If one of the `oe_free` calls is missing, the
+result should be "OE_MEMORY_LEAK".
+
+# Note
+
+To improve users' experience, we add the runtime functions that users can use
+to locate the memory leaks. However, the local detection implementation is not
+thread-safe.
+
+- For local detection, the tracking state is defined by the global variable
+`oe_use_debug_malloc_tracking`, which means if the new public functions
+execute on different threads at the same time, the results may be inaccurate.
+
+- The thread-safe concern is confined to local detection only. Debug malloc
+does not require thread-aware allocators to behave differently to work with
+debug malloc. The global detection is thread-safe and accurate.
+
+- Since debug malloc is a wrapper around the memory allocator, it is fully
+composable with a pluggable OE allocator, and it does not require any change to
+the OE allocators for the new features. Also, the underlying allocator
+execution on independent threads will not affect the result of debug malloc.
+
+Yet debug malloc is a useful tool for developers. Thus we hope the new feature
+can help the development of enclave applications.


docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,114 @@
...
+operations in enclave during the whole enclave lifetime. Based on the
+tracking record, if some memory is not freed after allocation, the tool can
+dectect the leaking and help to raise an error.
+
+As a wrapper of the lower memory allocator, the debugmalloc tool can help
+developers to track unintended memory leaking, no matter which pluggable
+memory allocator is employed at user's will.
+
+This proposal is suggesting to make debugmalloc available to end users, who
+may need the memory leaking dectetion during enclave application development.
```
dthaler, line 16:
```suggestion
may need the memory leaking detection during enclave application development.
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,114 @@
...
+## Local option
+
+To enable debugmalloc per enclave, the user just need to specify the link
+option "oedebugmalloc".
+Actually the global option is to append "oedebugmalloc" to the global link
+option list, then "oedebugmalloc" is included by default for all the enclave
+targets.
+
+Since the library takes effect through link time substitution, this link
+option must be placed befor "oecore", or there will be no debbugmalloc
```
dthaler, line 43:
```suggestion
option must be placed before "oecore", or there will be no debugmalloc
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,114 @@
...
+give the developers fine-grained control. It can only enable/disable debugmalloc
+per enclave, or from the view of building, per target.
+
+## Runtime option
+
+There is already some attempt to bypass debugmalloc(https://github.com/openenclave/openenclave/pull/3354).
+As a runtime option, the drawback is that it does not guarantee thread safety.
+Although for simple internal tests runtime option works, for user's multithread
+applications it does not work. Regular mutex lock does not help in such
+situation, since it requires the whole process from memomy allcation to memory
```
dthaler, line 56:
```suggestion
situation, since it requires the whole process from memory allocation to memory
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,154 @@
...
+{
+    oe_debug_malloc_start_tracking();
+}
+
+// This is the function that the host calls. It prints
+// a message in the enclave before calling back out to
+// the host to print a message from there too.
+void enclave_hello()
+{
+    oe_debug_malloc_init_callback();
```
dthaler, line 124: Delete this line since line 57 explains the callback is called automatically by the library so the app developer doesn't need to call it

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,154 @@
...
+
+# Design
+
+## User Experience
+- Debug malloc will be shipped to the users as a separate library oelibdebugmalloc.a
+- To use debug malloc, user will have to link oelibdebugmalloc.a into their enclaves.
+  This can be done by
+  ```
+      cmake:
+      target_link_libraries(user-enclave oelibc oedebugmalloc.a oeenclave)
```
dthaler, line 33: Why is the .a needed in this line?

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,154 @@
...
+- `void oe_debug_malloc_stop_tracking(void)`
+  Stop tracking memory allocations.
+
+- `void oe_debug_malloc_print_objects(void)`
+  Print tracked objects that alive, ie not yet freed.
+
+- `void oe_debug_malloc_init_callback(void)`
+  This function will be called by debug malloc during startup.
+  Users can implement this function to start a tracking session at the earliest possible
+  time.
```
dthaler, line 59: Since we have an init called automatically on enclave creation to (say) start tracking OE platform mallocs, should there be an automatically called uninit during enclave termination to (say) print unfreed objects?

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,154 @@
...
+
+Please note that the timing of calling of these functions is important. For
+example, the starting and stoping tracking must cover the coupled malloc/free
```
dthaler, line 62:
```suggestion
example, the starting and stopping tracking must cover the coupled malloc/free
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,154 @@
...
+operations, which means start tracking before malloc and also stop tracking
+after free, or debug malloc will definitely trigger false alarms. And by the same
+logic, oe_debug_malloc_print_objects() can only be called after tracking is
+stopped.
```
dthaler, line 66: In the meeting we suggested removing this restriction.

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,154 @@
...
+
+- `void oe_debug_malloc_init_callback(void)`
+  This function will be called by debug malloc during startup.
+  Users can implement this function to start a tracking session at the earliest possible
+  time.
+
+Please note that the timing of calling of these functions is important. For
+example, the starting and stoping tracking must cover the coupled malloc/free
+operations, which means start tracking before malloc and also stop tracking
+after free, or debug malloc will definitely trigger false alarms. And by the same
```
dthaler, line 64: Presumably the tracker deals with frees that weren't tracked (i.e., allocated before start tracking) when the encalve is doing start/stop itself, right?

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,154 @@
+Proposal to enable debug malloc for end users
+===
+
+# Motivation
```
CodeMonkeyLeet, line 4: Per the SIG arch meeting, I think it would be useful to discuss the use cases that we expect a developer to use this feature for, and to move the description of the intended logic into the Design section below. 

It sounds like the feature is designed to address at least two major use cases:

- A developer wants to track all allocations from the start of the enclave until it is terminated and determine if any allocations were leaked.
- A developer wants to track allocations in a specific region of the enclave app code that they can specify at dev time.

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,154 @@
...
+            result,
+            oe_result_str(result));
+    }
+}
+```
+While this sample has no memory leak, there is no expected extra output.
+However if there is memory leaking, oe_debug_malloc_print_objects() can report
+the leaks, and the enclave will raise an error "OE_MEMORY_LEAK".
+
+# Note
```
CodeMonkeyLeet, line 149: It would be useful to document how this new feature interacts with the pluggable allocators:

- It is fully composable with a pluggable OE allocator
- It does not require an OE allocator to implement new or different APIs to be used with the debug malloc.
- It does not require or guarantee any thread synchronization/serialization behaviors when used with thread-aware allocators.

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,154 @@
...
+However, users would like to isolate and track specific memory allocations.
+To achieve this, the following APIs are provided in a new published header `debugmalloc.h`.
+
+- `void oe_debug_malloc_start_tracking(void)`
+  Start tracking memory allocations.
+
+- `void oe_debug_malloc_stop_tracking(void)`
+  Stop tracking memory allocations.
+
+- `void oe_debug_malloc_print_objects(void)`
```
CodeMonkeyLeet, line 53: Per the SIG arch meeting, the suggestion is to return the string instead of logging directly to console, which would enable the caller to handle the output as they prefer (e.g. log to secure channel or output to file)

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,154 @@
...
+      $(CC) -o enclave enc.o $(LDFLAGS_CUSTOM)
+
+  ```
+
+## Fine-grained runtime control
+Currently, debug malloc prints leaks only during enclave termination.
+However, users would like to isolate and track specific memory allocations.
+To achieve this, the following APIs are provided in a new published header `debugmalloc.h`.
+
+- `void oe_debug_malloc_start_tracking(void)`
```
CodeMonkeyLeet, line 47: Consider separating out the use of the start/stop tracking methods specifically for only the use case where a developer wants to track allocations in a specific section of their code. (i.e. do not require a callback to invoke this method for tracking whole enclave app allocations, provide a different control mechanism for that instead).

Given that these are `void (void)` methods, it would also be useful to document the error handling or state machine when these functions are called. For example, what happens when start or stop calls are correctly paired or are nested?

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,154 @@
...
+to `off`.
+
+## User-defined callback function to subsitute the predefined function
+While user defines the callback function following the definition and matches
+the name exactly, the predefined function will be substituted by user's, thus
+user can decide the point to start tracking by calling the callback function.
+
+The expected implementation of the callback function is as following:
+  ```c
+  void oe_debug_malloc_init_callback(void)
```
CodeMonkeyLeet, line 95: As mentioned in the SIG arch meeting, I would prefer if we didn't offer an init callback method:

- It provides a lot of freedom for a dev to implement code this is unsafe or incorrect, given that the callback is invoked at a point that is heavily constrained during enclave init. 
- The primary use case for this seems to be to allow a developer to start tracking at the point of allocator init, and there are simpler mechanisms for that.

CodeMonkeyLeet (commented):
Thanks for writing this up, Alvin! I agree it's a useful feature for the devs and we should iterated on the details of this design.

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,154 @@
...
+  It also frees the associated callstack information.
+
+- When enclave terminates, it prints information about leaks - those objects that have been allocated,
+  but not feed.
+
+Debug malloc is performance intensive and is therefore not compiled into the SDK release.
+
+There have been requests from users that debug malloc be made available so that they can
+identify leaks within the enclave. Commonly used memory detection tools like Valgrind do not
+work within the enclaves since they cannot access enclave memory.
```
achamayou, line 23: Have we looked at enabling LeakSanitizer/AddressSanitizer in enclave?

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Yes exactly as you described. Thanks!

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Thanks, LeakSanitizer may be a good option. Now I just follow the fulfilled debug malloc and not much changes involved.

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Thanks, partially updated.

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Yes now this public function is removed.

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Fixed, thanks!

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Now the existing debug malloc is not changed, and it will raise an error if leak is detected during enclave termination.

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Fixed, thanks!

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: It is removed, thanks!

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Now the global detection is in charge of the whole life cycle from enclave creation to enclave termination, and the local detection just tracks the code from start to stop. Please check the updated doc, thanks!

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Thanks for the suggestion! According to the api of debug malloc https://github.com/openenclave/openenclave/blob/fbb477a36067b8fcbce2d78fe94e4f31a89e1442/enclave/core/debugmalloc.c#L454, maybe a int is more convenient for users. We can discuss about it.

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Thanks! the design is updated, and callback is removed.

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Thanks! callback is removed.

anakrish (approved):
LGTM.

@CodeMonkeyLeet's feedback needs to be addressed.

radhikaj:
@dthaler, can you please review this to see if you comments are addressed and if we can move forwards?

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+It is turned on by default in debug builds of the SDK, and helps detect leaks in the SDK.
+
+the main logic of Debug malloc is as following:
+- Whenever memory is allocated, it captures the callstack and associates it with the allocated memory.
+  Additionally it also sets the contents of the allocated memory to zero.
+
+- Whenever memory is freed, it clears the contents to a fixed pattern to help detect access of freed memory.
+  It also frees the associated callstack information.
+
+- When enclave terminates, it prints information about leaks - those objects that have been allocated,
```
mingweishih, line 16:
```suggestion
- When an enclave terminates, debug malloc prints information about leaks---those objects that have been allocated,
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+  but not feed.
```
mingweishih, line 17:
```suggestion
  but not freed.
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+
+Debug malloc is performance intensive and is therefore not compiled into the SDK release.
+
+There have been requests from users that debug malloc be made available so that they can
+identify leaks within the enclave. Commonly used memory detection tools like Valgrind do not
```
mingweishih, line 22:
```suggestion
identify memory leaks within an enclave. Commonly used memory detection tools like Valgrind do not
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+work within the enclaves since they cannot access enclave memory.
```
mingweishih, line 23:
```suggestion
work within enclaves since they cannot access enclave memory.
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
+Proposal to enable debug malloc for end users
+===
+
+# Motivation
+
+Debug malloc is an internal tool which can help identify memory leaks in enclaves.
+It is turned on by default in debug builds of the SDK, and helps detect leaks in the SDK.
+
+the main logic of Debug malloc is as following:
```
dthaler, line 9:
```suggestion
The main logic of Debug malloc is as follows:
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+Debug malloc is performance intensive and is therefore not compiled into the SDK release.
+
+There have been requests from users that debug malloc be made available so that they can
+identify leaks within the enclave. Commonly used memory detection tools like Valgrind do not
+work within the enclaves since they cannot access enclave memory.
+
+According to the user's developing experience, there are two major problems
+which the memory leaking detection can help to address:
+
+- A developer wants to track all allocations from the start of the enclave
```
mingweishih, line 28:
```suggestion
- A developer wants to track all memory allocations throughout the execution of an enclave
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+  until it is terminated and determine if any allocations were leaked.
```
mingweishih, line 29:
```suggestion
  and to determine if any of these allocations were leaked.
```

docs/DesignDocs/Enabledebugmalloc.md
dthaler, line 17:
```suggestion
  but not freed.
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+several pull requests, then this doc will emphasize the design of "local
+detection".
+
+# User Experience
+
+## Global detection
+
+### Default option
+
+The default option is "USE_DEBUG_MALLOC". Now we already set "USE_DEBUG_MALLOC"
```
dthaler, line 45:
```suggestion
The "USE_DEBUG_MALLOC" define controls the global default. Now we already set "USE_DEBUG_MALLOC"
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+to "ON" in debug mode by default. Users can modify this default option
+according to the need, such as change it to "ON" in release mode or change it
```
dthaler, line 47:
```suggestion
according to the need, such as changing it to "ON" in release mode or changing it
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+      make:
+      LDFLAGS=$(shell pkg-config oeenclave-$(COMPILER) --libs)
+      LDFLAGS_CUSTOM=$(subst -loelibc, -loedebugmalloc -loelibc, $(LDFLAGS))
+      $(CC) -o user-enclave user-enc.o $(LDFLAGS_CUSTOM)
+
+```
+
+For the global detection, if there is any memory leak during the enclave's
+life circle, an error code "OE_MEMORY_LEAK" will be raised during the enclave
+termination. This means it not only can check the users' enclave application,
```
dthaler, line 71:
```suggestion
termination. This means it not only can check the user's enclave application,
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+
+So far the "global detection" has already been fulfilled and improved across
+several pull requests, then this doc will emphasize the design of "local
+detection".
+
+# User Experience
+
+## Global detection
+
+### Default option
```
dthaler, line 43:
```suggestion
### Global default option
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+work within the enclaves since they cannot access enclave memory.
+
+According to the user's developing experience, there are two major problems
+which the memory leaking detection can help to address:
+
+- A developer wants to track all allocations from the start of the enclave
+  until it is terminated and determine if any allocations were leaked.
+  We name it "global detection".
+
+- A developer wants to track allocations in a specific region of the enclave
```
mingweishih, line 32:
```suggestion
- A developer wants to track memory allocations in a specific part of the enclave's
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+  app code that they can specify at dev time. We name it "local detection".
```
mingweishih, line 33:
```suggestion
  execution that he or she can specify at the source code level. We name it "local detection".
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+
+- The starting and stopping tracking must cover the coupled malloc/free
+operations, which means start tracking before malloc and also stop tracking
+after free, or the report will definitely trigger false alarms.
+
+- oe_debug_malloc_tracking_stop() can only been called after calling
+oe_debug_malloc_tracking_start(), and oe_debug_malloc_tracking_start() can be
+called at the beginning or after calling oe_debug_malloc_tracking_stop(). While
+the calling sequence following this rule, the return values are false. TRUE, or
+non-zero return value means the sequence is wrong and the report does not
```
dthaler, line 100: I'd recommend returning oe_result_t instead of bool, so the errors are more meaningful.
E.g. if you call start when it's already started, you might return OE_ALREADY_INITIALIZED or OE_UNEXPECTED.

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+
+      make:
+      LDFLAGS=$(shell pkg-config oeenclave-$(COMPILER) --libs)
+      LDFLAGS_CUSTOM=$(subst -loelibc, -loedebugmalloc -loelibc, $(LDFLAGS))
+      $(CC) -o user-enclave user-enc.o $(LDFLAGS_CUSTOM)
+
+```
+
+For the global detection, if there is any memory leak during the enclave's
+life circle, an error code "OE_MEMORY_LEAK" will be raised during the enclave
```
mingweishih, line 70:
```suggestion
life cycle, an error code "OE_MEMORY_LEAK" will be raised during the enclave
```

docs/DesignDocs/Enabledebugmalloc.md
mingweishih, line 71:
```suggestion
termination. This means it not only checks the users' enclave application,
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+termination. This means it not only can check the users' enclave application,
+but also can cover the SDK's allocation of the enclave side.
```
mingweishih, line 72:
```suggestion
but also covers the SDK's allocations on the enclave side.
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+after free, or the report will definitely trigger false alarms.
+
+- oe_debug_malloc_tracking_stop() can only been called after calling
+oe_debug_malloc_tracking_start(), and oe_debug_malloc_tracking_start() can be
+called at the beginning or after calling oe_debug_malloc_tracking_stop(). While
+the calling sequence following this rule, the return values are false. TRUE, or
+non-zero return value means the sequence is wrong and the report does not
+reflect the real situation.
+
+- oe_debug_malloc_tracking_start() and oe_debug_malloc_tracking_stop() can been
```
dthaler, line 103:
```suggestion
- oe_debug_malloc_tracking_start() and oe_debug_malloc_tracking_stop() can be
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+  |----code1--|-----------|----code2--|--|----code3--|
+  start1->stop1->report1->start2->stop2->start3->stop3->report3
+```
+
+In this case report1 only covers code1, and report3 covers code1, code2 and
+code3 cumulatively.
+
+# Implementation
+
+The debug malloc employees a linkedlist to make record of all memory operations.
```
dthaler, line 116:
```suggestion
The debug malloc employs a linked list to make a record of all memory operations.
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+
+    /* Contains HEADER_MAGIC2 */
+    uint64_t magic2;
+
+    /* User data */
+    uint8_t data[];
+};
+```
+
+We also add an variable
```
mingweishih, line 153:
```suggestion
We also add a variable
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+
+Currently, debug malloc prints leaks only during enclave termination.
+However, users would like to isolate and track specific memory allocations.
+To achieve this, the following APIs are provided in a new published header `debugmalloc.h`.
+
+- `bool oe_debug_malloc_tracking_start(void)`
+  Start tracking memory allocations.
+
+- `bool oe_debug_malloc_tracking_stop(void)`
+  Stop tracking memory allocations.
```
dthaler, line 84: What happens if you call oe_debug_malloc_tracking_stop() when USE_DEBUG_MALLOC is set?  Is that an error?

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+    void* buf2 = oe_malloc(4096);
+    oe_debug_malloc_tracking_stop();
+
+    oe_debug_malloc_tracking_start();
+    void* buf3 = oe_malloc(4096);
+    oe_debug_malloc_tracking_stop();
+    size2 = oe_debug_malloc_tracking_report();
+
+
+    printf("There are %d, %d un-freed objects reported", size1, size2);
```
dthaler, line 198: This line is confusing.  Put a separate printf for size1 up after line 186 and keep this one just about size2.

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+```
+For the local detection, the expected output result is
+`There are 1, 3 un-freed objects reported.`.
+For the global detection, Since this sample has no memory leak at the point of
+enclave termination, the result is 0. If one of the free is missing, the result
+should be "OE_MEMORY_LEAK".
+
+# Note
+- Since debug malloc is a wrapper of the lower memory allocator, It is fully
+composable with a pluggable OE allocator, and it does not require an change to
```
dthaler, line 224:
```suggestion
composable with a pluggable OE allocator, and it does not require any change to
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+the OE allocators for the new features.
+- To improve users' experience, we add the runtime functions that users can use
+to locate the memory leaks. However, the implementation is not thread-safe.
+Yet debug malloc is a useful tool for developers, Thus we hope the new feature
```
dthaler, line 228:
```suggestion
Yet debug malloc is a useful tool for developers. Thus we hope the new feature
```

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+For the global detection, Since this sample has no memory leak at the point of
+enclave termination, the result is 0. If one of the free is missing, the result
+should be "OE_MEMORY_LEAK".
+
+# Note
+- Since debug malloc is a wrapper of the lower memory allocator, It is fully
+composable with a pluggable OE allocator, and it does not require any change to
+the OE allocators for the new features.
+- To improve users' experience, we add the runtime functions that users can use
+to locate the memory leaks. However, the implementation is not thread-safe.
```
CodeMonkeyLeet, line 227: This is slightly different from the suggestion I made regarding its interaction with the pluggable allocation, which is only that debug malloc does not require thread-aware allocators to behave differently to work with debug malloc.

When we say that debug malloc itself is not thread-safe, we need to then explain what it means to be unsafe (e.g., does it crash? does it corrupt memory?) In this case, there are two aspects:
- The `start`/`stop`/`report` functions being called on multiple threads. e.g., if I call `start`, `stop` and `report` on different threads at the same time, I would expect the report may be inaccurate, but it shouldn't corrupt memory for example. 
- The underlying allocator tracking allocations separately on independent threads. e.g. if a piece of code that calls `report` after `stop` gets run on a different thread, I expect that it still returns me the same result and the underlying allocation doesn't have to do anything special to make this true.

CodeMonkeyLeet (commented):
Nice work, Alvin, thanks for cleaning up the doc from the SIG-arch review. I still have some questions about the details of the proposed state machine for the local allocation tracking methods though; see comments inline.

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+# Motivation
+
+Debug malloc is an internal tool which can help identify memory leaks in enclaves.
+It is turned on by default in debug builds of the SDK, and helps detect leaks in the SDK.
+
+the main logic of Debug malloc is as following:
+- Whenever memory is allocated, it captures the callstack and associates it with the allocated memory.
+  Additionally it also sets the contents of the allocated memory to zero.
+
+- Whenever memory is freed, it clears the contents to a fixed pattern to help detect access of freed memory.
```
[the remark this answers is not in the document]
nonpolarity: https://github.com/openenclave/openenclave/blob/fbb477a36067b8fcbce2d78fe94e4f31a89e1442/enclave/core/debugmalloc.c#L343

So the memory is not only freed, but also destroyed with pattern "0xDD". If a chunk is freed but we access it again and get "0xDD", we know the access is a mistake.

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: oe_debug_malloc_tracking_stop() is available when USE_DEBUG_MALLOC  is "on" or oedebugmalloc.a is linked explicitly by hand. So there is no error due to compiling option.

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Good suggestion, updated, thanks!

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Updated, thanks!

docs/DesignDocs/Enabledebugmalloc.md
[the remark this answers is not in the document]
nonpolarity: Good suggestion, now updated. Thanks!

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,229 @@
...
+
+- oe_debug_malloc_tracking_start() and oe_debug_malloc_tracking_stop() can been
+called many times:
+
+```
+  |----code1--|-----------|----code2--|--|----code3--|
+  start1->stop1->report1->start2->stop2->start3->stop3->report3
+```
+
+In this case report1 only covers code1, and report3 covers code1, code2 and
```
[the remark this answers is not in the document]
nonpolarity: Updated, thanks!

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,230 @@
...
+- `bool oe_debug_malloc_tracking_stop(void)`
+  Stop tracking memory allocations.
+
+- `int oe_debug_malloc_tracking_report(void)`
+  report the number of tracked objects that alive, ie not yet freed.
+
+Please note that the timing and sequences of calling of these functions is
+important:
+
+- The starting and stopping tracking must cover the coupled malloc/free
```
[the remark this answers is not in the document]
nonpolarity: Thanks, the descriptions are updated. Just to clarify, report is designed to covers all the leaks happened in the state of tracking, so the report is cumulative.

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Yes you are right, only local detection has thread safety concern. Please have a look again.

radhikaj (approved):

docs/DesignDocs/Enabledebugmalloc.md
dthaler: The "Global detection" section of this doc does not mention the stop API, and the Local detection section says "- oe_debug_malloc_tracking_stop() can only been called after calling
oe_debug_malloc_tracking_start()".   So if it can be called in Global detection mode, then the doc should be updated to say so.

nonpolarity: Thanks, it is updated to state global detection is independent of these APIs.

achamayou (approved):

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,289 @@
...
+  "OE_UNEXPECTED".
+
+- `oe_result_t oe_debug_malloc_tracking_stop(void)`
+
+  Stops tracking allocations that will be displayed in report. Multiple calls
+  to stop without another start are treated as no-ops and return value is
+  "OE_UNEXPECTED".
+
+- `    oe_result_t oe_debug_malloc_tracking_report(
+        uint64_t* out_num_objects,
```
dthaler, line 104:
```suggestion
        uint64_t* out_object_count,
```
avoid abbrevs

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,289 @@
...
+    oe_debug_malloc_tracking_report(&num, &report);
+    printf("There are %d un-freed objects reported:\n%s\n", num, report);
+
+    oe_debug_malloc_tracking_start();
+    void* buf2 = oe_malloc(4096);
+    oe_debug_malloc_tracking_stop();
+
+    oe_debug_malloc_tracking_start();
+    void* buf3 = oe_malloc(4096);
+    oe_debug_malloc_tracking_stop();
```
dthaler, line 238: SImon raises point about return value ignored

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,289 @@
...
+    oe_debug_malloc_tracking_report();
```
dthaler, line 239: Missing args

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,289 @@
...
+void enclave_hello()
+{
+    void* buf1, buf2, buf3;
+    uint64_t num;
+    char* report;
+
+    oe_debug_malloc_tracking_start();
+    void* buf1 = oe_malloc(4096);
+    oe_debug_malloc_tracking_stop();
+    oe_debug_malloc_tracking_report(&num, &report);
```
CodeMonkeyLeet, line 229: Caller should check the return for this since failed allocations can fail this function resulting in null report pointer.

docs/DesignDocs/Enabledebugmalloc.md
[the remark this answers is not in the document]
CodeMonkeyLeet: This cumulative tracking actually surprises me, and is not what I would expect looking at the state diagram that you added. Why are we tracking across `stop` boundaries? I would expect each start-stop pair to reset the state of local tracking. 

+@anakrish for his opinion on this.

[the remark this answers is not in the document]
nonpolarity: Considering tracking state and not tracking state as an entirety respectively, `stop` is not a boundary. Of course this is my subjective opinion, this design is based on such case:  since calling start() and stop() is easier than report(), the user can define many code segments as tracking state, then one time calling report() is enough to cover all the involved segments.

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Updated, thanks!

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Updated, thanks!

docs/DesignDocs/Enabledebugmalloc.md
nonpolarity: Updated, thanks!

docs/DesignDocs/Enabledebugmalloc.md
```diff
@@ -0,0 +1,294 @@
...
+            result,
+            oe_result_str(result));
+    }
+}
+```
+For the local detection, the expected output result are 1 and 3 leaks for each
+report.
+
+For the global detection, Since this sample has no memory leak at the point of
+enclave termination, the result is "OE_OK". If one of the free is missing, the
```
dthaler, line 271:
```suggestion
enclave termination, the result is "OE_OK". If one of the `oe_free` calls is missing, the
```

dthaler (approved):

nonpolarity:
bors try

nonpolarity:
bors try

PR #3612, opened 2020-09-29
Run attestation sample only if DCAP libraries exist

### LINKED ISSUE #3595: samples\attestation is still tested in ctest case "samples", even when DCAP libraries not installed on Windows
During OE Windows setup, I chose not to install DCAP driver by running:
`./scripts/install-windows-prereqs.ps1 -InstallPath C:/oe_prereqs -LaunchConfiguration SGX1 -DCAPClientType None
`

However later in ctest, case "samples" always failed. I looked into the detail, and found the sample project `attestation` had an execution failure. Although at the beginning of the samples' test,  it did say: 
```
System does not have DCAP libraries.  Attestation based samples will NOT be
  tested.
```

I am wondering if case `attestation` in 
https://github.com/openenclave/openenclave/blob/7201ae4aa35fb9ba644fdace465dd9b810e13f0f/samples/test-samples.cmake#L65
can be moved to 
https://github.com/openenclave/openenclave/blob/7201ae4aa35fb9ba644fdace465dd9b810e13f0f/samples/test-samples.cmake#L70

ryanhsu19:
Move `attestation` sample to line 70 so that, like `attested_tls` sample, it will be included in `ctest` only if DCAP libraries exist.

Fix #3595 .

### DIFF
--- a/samples/test-samples.cmake
+++ b/samples/test-samples.cmake
@@ -62,12 +62,12 @@ else ()
   # they can run even if they weren't built against SGX, because in
   # that cause they directly interface with the AESM service.
   if (BUILD_ENCLAVES)
-    list(APPEND SAMPLES_LIST data-sealing attestation)
+    list(APPEND SAMPLES_LIST data-sealing)
 
     # These tests can only run with SGX-FLC, meaning they were built
     # against SGX.
     if (HAS_QUOTE_PROVIDER)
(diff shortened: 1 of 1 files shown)


dthaler (approved):

BRMcLaren (approved):

BRMcLaren:
bors try

yentsanglee (approved):

PR #3613, opened 2020-09-29
Document and verify that claims in oe_verify_evidence() is optional

### LINKED ISSUE #3594: Confirm and check that the parameter `claims` in API `oe_verify_evidence()` is optional
In `include/openenclave/attestation/verifier.h`, the `claims` parameter in API `oe_verify_evidence()` is not documented as optional, but the implementation of this API in `common/attest_plugin.c` allows this parameter to be NULL. On the other hand, `_verify_evidence()` in `common/sgx/verifier.c` requires its `claims` parameter to be non-NULL.

Need to clarify doxygen documentation and update implementation to support the `claims` parameter in API `oe_verify_evidence()`as optional, and to skip calling `oe_sgx_extract_claims()` when this parameter is NULL.

This issue was brought up in this [comment thread](https://github.com/openenclave/openenclave/pull/3499#discussion_r494672966).

shnwc:
Fix #3594

### DIFF
--- a/common/sgx/verifier.c
+++ b/common/sgx/verifier.c
@@ -456,8 +456,9 @@ oe_result_t oe_sgx_extract_claims(
 
     // Note: some callers can have custom_claims_buffer pointing to a non-NULL
     // buffer containing a zero-sized array.
-    if (!format_id || !report_body || !report_body_size ||
-        (!custom_claims_buffer && custom_claims_buffer_size))
+    if (!format_id || !report_body || !report_body_size || !claims_out ||
+        !claims_length_out ||
+        (!custom_claims_buffer != !custom_claims_buffer_size))
         OE_RAISE(OE_INVALID_PARAMETER);
 
     // verify the integrity of the custom_claims_buffer with hash stored in
@@ -569,8 +570,9 @@ static oe_result_t _verify_evidence(
     size_t custom_claims_buffer_size = 0;
     oe_uuid_t* format_id = NULL;
 
-    if (!context || !evidence_buffer || !evidence_buffer_size || !claims ||
-        !claims_length)
+    if (!context || !evidence_buffer || !evidence_buffer_size ||
+        (!endorsements_buffer != !endorsements_buffer_size) ||
+        (!claims != !claims_length))
         OE_RAISE(OE_INVALID_PARAMETER);
 
     format_id = &context->base.format_id;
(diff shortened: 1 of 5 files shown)


shnwc:
bors try

common/sgx/verifier.c
```diff
@@ -456,8 +456,9 @@ oe_result_t oe_sgx_extract_claims(
 
     // Note: some callers can have custom_claims_buffer pointing to a non-NULL
     // buffer containing a zero-sized array.
-    if (!format_id || !report_body || !report_body_size ||
-        (!custom_claims_buffer && custom_claims_buffer_size))
+    if (!format_id || !report_body || !report_body_size || !claims_out ||
+        !claims_length_out ||
+        (!custom_claims_buffer != !custom_claims_buffer_size))
```
dthaler, line 461: This PR doesn't touch include\openenclave\internal\sgx\plugin.h but that's where the doxygen for oe_sgx_extract_claims() is and `custom_claims_buffer` and size are not stated as optional there, so I think we need to update that file too (I will also observe that the `sgx_endorsements` param is out of order in the param list in the doxygen so if you update the file maybe fix that too)

shnwc: Good catch! I'll fix. On the other hand, `include\openenclave\internal\sgx\plugin.h` is internal, so OE SDK application developers don't see this header file or its doxygen document.

shnwc:
bors try

include/openenclave/internal/sgx/plugin.h
```diff
@@ -33,11 +33,11 @@ typedef enum _sgx_evidence_format_type_t
...
  * @param[in] report_body Pointer to a buffer with raw SGX quote / report.
  * @param[in] report_body_size Size of the report_body buffer.
- * @param[in] sgx_endorsements Pointer to the endorsements buffer.
- * @param[in] custom_claims_buffer The data in a flat buffer to be packaged as a
- * claim of ID OE_CLAIM_CUSTOM_CLAIMS_BUFFER when format_type has the right
- * value.
+ * @param[in] custom_claims_buffer If not NULL, it holds the data in a flat
+ * buffer to be packaged as a claim of ID OE_CLAIM_CUSTOM_CLAIMS_BUFFER when
+ * format_type has the right value.
  * @param[in] custom_claims_buffer_size The size of the custom_claims buffer.
```
dthaler, line 39: Shouldn't the size param also say "If not NULL, ..."?

shnwc: `custom_claims_buffer_size` is of type `size_t`, so it can't be NULL.

dthaler (approved):

yentsanglee (approved):

qiucwang (approved):

PR #3614, opened 2020-09-29
Fix #3585, remove misleading buffer too small messages

### LINKED ISSUE #3585: Misleading OE_BUFFER_TOO_SMALL error messages
There are many benign OE_BUFFER_TOO_SMALL errors in traces from functions which return required buffer size. Such functions are typically called twice - first time to get required size; second time to retrieve data with allocated buffer. The first call unnecessarily prints OE_BUFFER_TOO_SMALL error messages that can be misleading when root causing issue from log.

Example:
```c++
    r = oe_get_crl_distribution_points(..., NULL, &buffer_size);
    if (r == OE_BUFFER_TOO_SMALL)
    {
        uint8_t* buffer = (uint8_t*)malloc(buffer_size);
        oe_get_crl_distribution_points(..., buffer, &buffer_size);
```

qiucwang:
Fixes #3585 

### DIFF
--- a/host/crypto/openssl/key.c
+++ b/host/crypto/openssl/key.c
@@ -347,7 +347,13 @@ oe_result_t oe_public_key_write_pem(
         if (*size < mem->length)
         {
             *size = mem->length;
-            OE_RAISE(OE_BUFFER_TOO_SMALL);
+
+            if (data)
+                OE_RAISE(OE_BUFFER_TOO_SMALL);
+            /* If data is null, this call is intented to get the correct
+             * size so no need to trace OE_BUFFER_TOO_SMALL */
+            else
+                OE_RAISE_NO_TRACE(OE_BUFFER_TOO_SMALL);
         }
 
         /* Copy result to output buffer */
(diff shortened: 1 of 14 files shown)


yentsanglee:
Log at info level rather than error level if the caller passes in a null buffer into these functions.

host/crypto/openssl/key.c
```diff
@@ -347,7 +347,7 @@ oe_result_t oe_public_key_write_pem(
         if (*size < mem->length)
         {
             *size = mem->length;
-            OE_RAISE(OE_BUFFER_TOO_SMALL);
+            OE_RAISE_NO_TRACE(OE_BUFFER_TOO_SMALL);
```
yentsanglee, line 350: I believe in this case you want trace.

yentsanglee (approved):

host/crypto/openssl/key.c
qiucwang: Thanks for catching that. Miss this one

qiucwang:
bors try

qiucwang:
bors try

qiucwang:
bors retry

ryanhsu19 (approved):

yentsanglee:
bors retry

PR #3620, opened 2020-09-30
Fallback to FXRSTOR in oe_enter if XSAVE is unsupported

### LINKED ISSUE #3360: Remove hard dependency on XRSTOR in OE SDK
OE SDK code currently assumes that support for XSAVE/XRSTOR is always available when using SGX enclaves. This can potentially cause runtime fails in other scenarios such as simulation mode, as seen with the related issue #3350 around assuming support for compacted representation of XSAVE data (XCOMP_BV[63]). 

The enclave side code should check that support for XSAVE/XRSTOR is supported at runtime, and fallback to the use of FXSAVE/FXRSTOR if not. As part of this work, we should also evaluation consolidating the codepaths which only use FXSAVE/FXRSTOR today to use the dynamic detection.

CodeMonkeyLeet:
- Check if XSAVE/XRSTOR is supported during enclave init and cache the state globally.
- Update `oe_enter` to use XRSTOR only if it is enabled by the OS, otherwise, fallback to FXRSTOR.
  - On first EENTER, this always uses FXRSTOR because the enclave hasn't checked feature support from the EREPORT yet.
  - This is acceptable as the first invocation is always the enclave init method, which doesn't use extended feature states.
  - Simulation mode only uses FXRSTOR as the use of EREPORT is not supported.
- Update `oe_enter` to always clear general purpose registers as well.
  - `oe_enter` now calls `oe_cleanup_registers` instead of just `oe_cleanup_xstates`, which also clears the RFLAGS that were previously handled separately at the beginning of `oe_enter`.

Fixes #3360

### DIFF
--- a/enclave/core/sgx/asmcommon.inc
+++ b/enclave/core/sgx/asmcommon.inc
@@ -28,13 +28,22 @@
     push %rax
     push %rdx
 
+    // Check if XSAVE/XRSTOR is supported
+    lea     oe_is_xsave_supported(%rip), %rax
+    movl    (%rax), %eax
+    cmpl    $0, %eax
+    jz      1f
+
     // Set the XSAVE_MASK
     mov $0xFFFFFFFF, %eax
     mov $0xFFFFFFFF, %edx
 
     // Restore initial enclave XSAVE state
     xrstor64 OE_XSAVE_INITIAL_STATE(%rip)
-
+    jmp     2f
+1:
+    fxrstor64 OE_XSAVE_INITIAL_STATE(%rip)
+2:
     // Restore the registers
     pop %rdx
     pop %rax
@@ -70,7 +79,9 @@
     xor %r14, %r14
     xor %r15, %r15
 
-    // zero out flags
+    // Zero out the RFLAGS
+    // This only clears the DF and state flag bits since
+    // the system flags and reserved bits are not writable here
     push %r15
     popfq
 .endm
--- a/enclave/core/sgx/asmcommon.inc
+++ b/enclave/core/sgx/asmcommon.inc
@@ -28,9 +28,10 @@
     push %rax
     push %rdx
 
-    // Check if XSAVE/XRSTOR is supported
-    lea     oe_is_xsave_supported(%rip), %rax
-    movl    (%rax), %eax
+    // Check if extended states are supported by the OS.
+    // If not, fallback to FXRSTOR to clear legacy SSE states only.
+    // Otherwise, clear both legacy SSE states and extended states with XRSTOR.
+    movl    oe_is_xsave_supported(%rip), %eax
     cmpl    $0, %eax
     jz      1f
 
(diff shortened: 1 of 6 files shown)


CodeMonkeyLeet:
bors try

enclave/core/sgx/asmcommon.inc
```diff
@@ -28,13 +28,22 @@
     push %rax
     push %rdx
 
+    // Check if XSAVE/XRSTOR is supported
+    lea     oe_is_xsave_supported(%rip), %rax
```
anakrish, line 32: Perhaps this can be `mov oe_is_xsave_supported(%rip),  %eax`

enclave/core/sgx/asmcommon.inc
```diff
@@ -28,13 +28,22 @@
...
     // Set the XSAVE_MASK
     mov $0xFFFFFFFF, %eax
     mov $0xFFFFFFFF, %edx
 
     // Restore initial enclave XSAVE state
     xrstor64 OE_XSAVE_INITIAL_STATE(%rip)
-
+    jmp     2f
+1:
+    fxrstor64 OE_XSAVE_INITIAL_STATE(%rip)
```
anakrish, line 45: Can you add a comment explaining why fxrstor64 is needed here, and not needed when the CPU does not support xsave.

enclave/core/sgx/calls.c
```diff
@@ -171,6 +172,10 @@ static oe_result_t _handle_init_enclave(uint64_t arg_in)
             /* Initialize the CPUID table before calling global constructors. */
             OE_CHECK(oe_initialize_cpuid());
 
+            /* Initialize the xstate settings
+             * Depends on TD and sgx_create_report, so can't happen earlier */
+            OE_CHECK(oe_set_is_xsave_supported());
```
anakrish, line 177: Can the host cause an exception in the enclave before xsave_supported is set?
Does our exception handling infrastructure need to do anything in regards to xsave supported?

CodeMonkeyLeet: The working assumption is the OS can always cause an enclave to fault at any time. However, AEX handles the XSAVE state area via the XFRM settings of the enclave and also provides synthetic state for extended features, which is unrelated to the flags that our runtime sets, so it is not my understanding that this needs to be mirrored in the exception handling code path. 

There's a different issue in the nested callback to the exception dispatcher where we are not preserving and restoring the xsave state (`oe_snap_current_context`/`oe_continue_execution`),but that's a separate work item to fix preservation of the XSAVE state everywhere (in host on ecall, in enclave on ocall, in exception dispatching).

enclave/core/sgx/asmcommon.inc
CodeMonkeyLeet: Thanks, removed the extra `lea`.

enclave/core/sgx/asmcommon.inc
CodeMonkeyLeet: Added more comments to the conditional and the branch.

anakrish (approved):
LGTM

bodzhang (approved):

PR #3622, opened 2020-10-01
Return OE_BUFFER_TOO_SMALL after initializing the collateral

### LINKED ISSUE #3566: Bug in collateral extraction?
The code following the call to `oe_get_quote_verification_collateral_ocall` [here](https://github.com/openenclave/openenclave/blob/98b71a493cbbadda84c72187e9b5a59a3a1d7dec/enclave/sgx/collateralinfo.c#L195) expects that all out-sizes have been filled in correctly if it returns `OE_BUFFER_TOO_SMALL`. That OCALL however, returns after the first element that doesn't fit into the buffer ([here](https://github.com/openenclave/openenclave/blob/36c0303cf92ff206f270ac89d7b3cd721d83fa76/host/sgx/ocalls/ocalls.c#L179)) and it won't initialize the remaining out-sizes.

That code has been there for months though, is there some hard-to-see detail that I'm missing? I think that either `oe_get_quote_verification_collateral_ocall` should produce all of the sizes the first time around, or `oe_get_sgx_quote_verification_collateral` needs to `continue;` the loop after each call to `oe_realloc`.

ryanhsu19:
Since `oe_get_quote_verification_collateral_ocall()` expects that all out-sizes have been filled correctly, the definition is revised so that OE_BUFFER_TOO_SMALL is returned after collateral initialization is done as much as possible.

Fix #3566 .

### DIFF
--- a/host/sgx/ocalls/ocalls.c
+++ b/host/sgx/ocalls/ocalls.c
@@ -158,6 +158,11 @@ oe_result_t oe_get_quote_verification_collateral_ocall(
     oe_get_sgx_quote_verification_collateral_args_t args = {0};
     bool buffer_too_small = false;
 
+    /* The value of any_buffer_too_small will be true if any buffer_too_small
+     * value becomes true. It is used to prevent terminating the collateral
+     * initialization process early. */
+    bool any_buffer_too_small = false;
+
     /* fmspc */
     memcpy(args.fmspc, fmspc, sizeof(args.fmspc));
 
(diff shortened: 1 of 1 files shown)


yentsanglee (approved):

qiucwang (approved):
LGTM. Just maybe naming of `result_oe_buffer_too_small` could be more straight here? Like `any_buffer_too_small` ? Or add a comment for this Boolean that it's true if any of the following buffer is too small. It's used here because we don't want to early stop collateral initialization? So that it could be more understandable for other readers.

ryanhsu19:
bors try

PR #3624, opened 2020-10-01
Refactor and cleanup SYSCALLs infrastructure

anakrish:
This PR refactors and cleans up the implementation of syscall dispatching.
It will be followed up with another PR that reduces the TCB.

### Description

Currently syscalls are dispatched via single ~800 line switch case in syscall/syscall.c.

In addition to resulting in a large function, and associated maintenance issues, a large switch case has the following drawback:

- **Larger TCB.** Enclaves that don't use a libc feature will still contain code for that feature.
  For example, echo enclave that just prints a hello world still has code for sockets, epoll and all the other
  syscalls it does not use.
  
- There is no easy way to figure out the **set of syscalls** supported by OE rather than perusing the syscall dispatch function.

- No **parameter checking** for syscalls.
  Different syscalls take different number of parameters. There are currently no checks to make sure that
  OE implementation of a syscall takes the necessary number of parameters.

Implementation:

- The list of supported syscalls are declared in syscall_decls.h

- Each syscall is declared via OE_DECLARE_SYSCALLn macro where n is the number of arguments the syscall expects.
  E.g:
     `OE_DECLARE_SYSCALL4(SYS_faccessat);`
     `OE_DECLARE_SYSCALL2(SYS_flock);`
  If the callsite (which is generally in MUSL) does not pass the expected number of parameters, there will be a compile error.

  Couple of syscalls may indeed expect varying number of parameters, these can be declared in their generic form.

- Each syscall is defined using the OE_DEFINE_SYSCALLn macro:
  E.g:
  ```c
  OE_DEFINE_SYSCALL5(SYS_epoll_pwait)
  {
     ....
  }
  ```

  Under the hood, this results in the function **oe_SYS_epoll_pwait_impl**

  This allows querying easily figuring out all the syscalls an enclave uses by searching for oe_SYS_ functions
  in the binary:
  ```bash
    objdump -t echo_enc | grep oe_SYS_
    00000000000fe950 l     F .text  00000000000000b0 oe_SYS_mount_impl
    00000000000fd7b0 l     F .text  0000000000000149 oe_SYS_faccessat_impl
    ...

- To retain current behavior, syscalls are still dispatched via a switchcase that is now much simpler.
   ```c
   static long _syscall(
    long num,
    long arg1,
    long arg2,
    long arg3,
    long arg4,
    long arg5,
    long arg6)
    {
       oe_errno = 0;

       switch (num)
       {
          OE_SYSCALL_DISPATCH(SYS_accept, arg1, arg2, arg3);
          OE_SYSCALL_DISPATCH(SYS_access, arg1, arg2);
  ```
  In the future the entire switch case will be avoided by the following strategy:

  A call to a specific syscall:
     `syscall(SYS_read, arg1, arg2)`

  will be transormed to:
     `oe_SYS_read_impl(arg1, arg2)`
     
  This completely avoids the switch case. As a result, only those syscalls that are
  actually used by a binary will get linked in.

### DIFF
--- a/syscall/CMakeLists.txt
+++ b/syscall/CMakeLists.txt
@@ -58,6 +58,10 @@ enclave_enable_code_coverage(oesyscall)
 
 add_enclave_dependencies(oesyscall syscall_trusted_edl)
 
+# Depend on oelibc in order to include bits/syscall.h which contains MUSL definitions of syscall enums.
+# Syscall infrastructure will assert that MUSL enums (e.g SYS_open) match the corresponding OE enums (e.g OE_SYS_open).
+enclave_link_libraries(oesyscall PRIVATE oelibc_includes)
+
 enclave_include_directories(
   oesyscall PRIVATE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR}
   ${PROJECT_SOURCE_DIR}/include/openenclave/corelibc)
--- a/syscall/syscall.c
+++ b/syscall/syscall.c
@@ -12,6 +12,7 @@
 #include <openenclave/internal/syscall/device.h>
 #include <openenclave/internal/syscall/dirent.h>
 #include <openenclave/internal/syscall/fcntl.h>
+#include <openenclave/internal/syscall/hook.h>
 #include <openenclave/internal/syscall/raise.h>
 #include <openenclave/internal/syscall/sys/ioctl.h>
 #include <openenclave/internal/syscall/sys/mount.h>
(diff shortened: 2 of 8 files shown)


anakrish:
bors try

CodeMonkeyLeet (commented):
Changes look good, see comment about the header inclusion though.

Thanks for the write up in the PR, it's very helpful. I think this PR is good, but I'm a little more hesitant about removing the centralized `_syscall` aggregation method in the future. That would imply having to recompile all libraries that make direct syscall invocations against the OE redefinitions, no? The impact of this is probably limited to specific kinds of apps (e.g. libOS that don't access syscalls through OE libc) but  I still wonder if there's value in keeping the `_syscall` shim around.

syscall/syscall.c
```diff
@@ -33,851 +34,1043 @@ typedef int (*ioctl_proc)(
...
-            {
-                oe_errno = OE_EINVAL;
-                goto done;
-            }
+OE_DEFINE_SYSCALL1(SYS_exit)
+{
+    oe_errno = 0;
+    int status = (int)arg1;
+    oe_exit(status);
+    // TODO: Is this correct return value.
```
jhand2, line 207: I checked the musl and glibc implementations, and they just don't use the return code of this syscall, so ultimately it probably doesn't matter. oe_exit doesn't return anyway.

syscall/syscall.c
```diff
@@ -33,851 +34,1043 @@ typedef int (*ioctl_proc)(
...
+        OE_SYSCALL_DISPATCH(SYS_mkdirat, arg1, arg2, arg3);
+        OE_SYSCALL_DISPATCH(SYS_mount, arg1, arg2, arg3, arg4, arg5);
+        OE_SYSCALL_DISPATCH(SYS_nanosleep, arg1, arg2);
+        OE_SYSCALL_DISPATCH(SYS_newfstatat, arg1, arg2, arg3, arg4);
+        OE_SYSCALL_DISPATCH(SYS_open, arg1, arg2, arg3);
+        OE_SYSCALL_DISPATCH(SYS_openat, arg1, arg2, arg3, arg4);
+        OE_SYSCALL_DISPATCH(SYS_poll, arg1, arg2, arg3);
+        OE_SYSCALL_DISPATCH(SYS_ppoll, arg1, arg2, arg3, arg4);
+        OE_SYSCALL_DISPATCH(SYS_pread64, arg1, arg2, arg3, arg4);
+        // TODO: 5 or 6 args?
```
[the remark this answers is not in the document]
jhand2: +1 to this suggestions

jhand2 (approved):
Looks great! Just a couple things + I agree with Simon's comments. Nothing blocking though.

syscall/syscall.c
```diff
@@ -33,851 +34,1043 @@ typedef int (*ioctl_proc)(
...
+static long _syscall(
+    long num,
+    long arg1,
+    long arg2,
+    long arg3,
+    long arg4,
+    long arg5,
+    long arg6)
+{
+    oe_errno = 0;
```
mingweishih, line 973: It seems now every syscall wrapper sets `oe_errno` to 0. I assume it's because those wrappers will be invoked directly in the following PRs and `_syscall` (and hence this line) can be bypassed. For the consistency, should be just remove this line?

mingweishih (approved):
The changes look good to me. Just a comment regarding the consistency.

include/openenclave/internal/syscall_decls.h
```diff
@@ -0,0 +1,181 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#ifndef _OE_INTERNAL_SYSCALL_DECLS_H
```
dthaler, line 4: https://github.com/openenclave/openenclave/blob/master/docs/Community/DevelopmentGuide.md says:

> DON'T use abbreviations unless they are already well-known terms known by users (e.g., "app", "info"), or are already required for use by developers (e.g., "min", "max", "args"). Examples of bad use would be `num_widgets` instead of `widget_count`, and `opt_widgets` instead of `option_widgets` or `optional_widgets`.

Any reason this file should say "decls" instead of "declarations"?  After all, the macro is OE_DECLARE_SYSCALL7 not OE_DECL_SYSCALL...

include/openenclave/internal/syscall_decls.h
```diff
@@ -0,0 +1,181 @@
...
+#endif
+
+OE_EXTERNC_BEGIN
+
+#define OE_SYSCALL_NAME(idx) oe##idx##_impl
+#define OE_CHECK_IDX(idx) OE_STATIC_ASSERT((idx + 1) > 0)
+
+#define OE_SYSCALL_DISPATCH(idx, ...) \
+    case OE_##idx:                    \
+        return OE_SYSCALL_NAME(_##idx)(__VA_ARGS__)
```
dthaler, line 29:
```suggestion
#define OE_SYSCALL_NAME(index) oe##index##_impl
#define OE_CHECK_IDX(index) OE_STATIC_ASSERT((index+ 1) > 0)

#define OE_SYSCALL_DISPATCH(index, ...) \
    case OE_##index:                    \
        return OE_SYSCALL_NAME(_##index)(__VA_ARGS__)
```
and same for line 40 etc.

Also I'd recommend `OE_CHECK_IDX` -> `OE_CHECK_INDEX`

syscall/syscall.c
```diff
@@ -33,851 +34,1043 @@ typedef int (*ioctl_proc)(
...
+
+    switch (num)
```
dthaler, line 975: Instead of a switch, could this use a function pointer table which I think would be more efficient, indexed by syscall #?  Probably all the functions would then take the same number of arguments so that a common prototype is used, but only use the args shown below.

syscall/syscall.c
```diff
@@ -33,851 +34,1043 @@ typedef int (*ioctl_proc)(
...
+    const char* target = (const char*)arg1;
+    int flags = (int)arg2;
+
+    (void)flags;
+
+    return oe_umount(target);
+}
+
+static long _syscall(
+    long num,
```
dthaler, line 965:
```suggestion
    long number,
```

Avoid abbreviations, and the Ubuntu man page calls it "number" anyway:

> SYSCALL(2)                                    Linux Programmer's Manual                                    SYSCALL(2)
> 
> NAME
>        syscall - indirect system call
> 
> SYNOPSIS
>        #define _GNU_SOURCE         /* See feature_test_macros(7) */
>        #include <unistd.h>
>        #include <sys/syscall.h>   /* For SYS_xxx definitions */
> 
>        long syscall(long number, ...);

anakrish:
> Changes look good, see comment about the header inclusion though.
> 
> Thanks for the write up in the PR, it's very helpful. I think this PR is good, but I'm a little more hesitant about removing the centralized `_syscall` aggregation method in the future. That would imply having to recompile all libraries that make direct syscall invocations against the OE redefinitions, no? The impact of this is probably limited to specific kinds of apps (e.g. libOS that don't access syscalls through OE libc) but I still wonder if there's value in keeping the `_syscall` shim around.

Libraries anyways ought to be recompiled with enclave appropriate flags to work within enclaves. Removing _syscall aggregation method does not incur an extra recompilation beyond that. Even libos applications ought to be compiled against OE specific MUSL to make sure that they work.

We can however retain the _syscall dispatcher as you suggest so that if some code does indeed call syscall with a non compile time known index, it will get routed to the _syscall dispatcher. However, the moment that happens, the TCB of the enclave would expand to include all the supported syscalls.

We can revisit this in the next PR.

syscall/CMakeLists.txt
```diff
@@ -58,6 +58,8 @@ enclave_enable_code_coverage(oesyscall)
 
 add_enclave_dependencies(oesyscall syscall_trusted_edl)
 
+enclave_link_libraries(oesyscall PRIVATE oelibc_includes)
```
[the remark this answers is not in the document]
anakrish: The syscall_decls.h asserts that the MUSL syscall enums (e.g SYS_open) match the corresponding OE enums (OE_SYS_open). MUSL invocations of syscalls use the MUSL enums whereas OE routing uses OE enums.
If both these enums don't match, then there will be silent runtime errors due to incorrect routing of calls.

To safeguard against these errors, syscall_decls.h uses compile time assertions to ensure that both the enums match.
The MUSL enums live in bits/syscall.h which is generated as part of oelibc_includes. Hence this dependency.

I have added comments explaining the above.

include/openenclave/internal/syscall.h
```diff
@@ -6,6 +6,7 @@
 
 #include <openenclave/bits/defs.h>
 #include <openenclave/bits/result.h>
+#include "syscall_decls.h"
```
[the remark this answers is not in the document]
anakrish: Good point. Added the include to internal/syscall/sys/syscall.h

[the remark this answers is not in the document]
anakrish: Renamed this syscall.h to syscall_hook.h. Additionally moved the files to internal/syscall folder.

include/openenclave/internal/syscall_decls.h
```diff
@@ -0,0 +1,181 @@
...
+#define OE_DEFINE_SYSCALL5 OE_DECLARE_SYSCALL5
+#define OE_DEFINE_SYSCALL6 OE_DECLARE_SYSCALL6
+#define OE_DEFINE_SYSCALL7 OE_DECLARE_SYSCALL7
+
+/** List of syscalls that are supported within enclaves.
+ ** In alphabetical order.
+ **/
+
+OE_DECLARE_SYSCALL3(SYS_accept);
+#if defined(OE_SYS_access)
```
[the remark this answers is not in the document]
anakrish: Yes, some methods seem to be not available on aarch64. The pattern to use `#if defined(OE_SYS_xxx)` was introduced in this changelist:
https://github.com/openenclave/openenclave/commit/916edc872a753195f529445c5d55a34e93ace614

I agree it is more readable to use the corresponding platform check ( `#if __x86_64__`) which documents also highlights if a syscall is only available in a certain platform(s).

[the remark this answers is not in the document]
anakrish: Removed `#if defined(OE_SYS_...)` and used the corresponding platform check instead.

syscall/syscall.c
anakrish: Good point.

syscall/syscall.c
anakrish: Removed. Added the comment explaning that the syscall implementation functions could be called directly, and hence they must each set oe_errno correctly.

syscall/syscall.c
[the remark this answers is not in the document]
anakrish: Created issue #3680

include/openenclave/internal/syscall_decls.h
anakrish: Done.

syscall/syscall.c
anakrish: Change num to number. But kept arg1, arg2, arg3 etc.

include/openenclave/internal/syscall_decls.h
anakrish: Moved/renamed the file to include/openenclave/internal/syscall/declarations.h
Updated included guard.

syscall/syscall.c
anakrish: This is a good suggestion. I'm planning to attempt this in a future PR.
Note: In subsequent PRs syscalls will be routed directly to the actually implementation instead of via the switch-case, thereby allowing the linker to eliminated syscalls that are not used by an enclave, thus reducing the TCB.

PR #3627, opened 2020-10-02
Make tcbinfo parsing more expandable to future versions.

qiucwang:
SGX V3 tcbinfo structure was revoked so PR #3386 is closed. But some of the changes are to make tcb parser more expandable to future versions so can be kept. This PR is to keep those changes, mostly

1. A while loop was implemented in `_read_tcb_info` to parse a list of tcbLevel entries. Now this while loop was moved into `_read_tcb_info_tcb_level` so that code can be cleaner in `_read_tcb_info`, considering more versions could be added to `_read_tcb_info` in the future.
2. The negative tests were isolated from the positive test function so that those negative tests do not need to be repeated in the `tcbInfo_with_pceid.json` positive test after the `tcbInfo.json` positive test was tested. So now all negative tests are tested once instead of twice.

### DIFF
--- a/common/sgx/tcbinfo.c
+++ b/common/sgx/tcbinfo.c
@@ -387,34 +387,47 @@ static void _move_to_end_of_tcb_levels(const uint8_t** itr, const uint8_t* end)
 static oe_result_t _read_tcb_info_tcb_level_v1(
     const uint8_t** itr,
     const uint8_t* end,
-    oe_tcb_info_tcb_level_t* platform_tcb_level)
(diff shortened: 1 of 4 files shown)


ryanhsu19 (approved):
LGTM. Thanks for making it easier to read and extend!

qiucwang:
bors try

qiucwang:
bors retry

qiucwang:
bors try

yentsanglee (approved):

PR #3639, opened 2020-10-05
SGX2 EDMM

wintersteiger:
A basic design proposal to start the discussion about support for EDMM in SGX2. 

CC @bodzhang @sylvanc @ad-l

### DIFF
--- a/docs/DesignDocs/SGX2-EDMM.md
+++ b/docs/DesignDocs/SGX2-EDMM.md
@@ -1,0 +1,105 @@
+# SGX2 EDMM in Open Enclave SDK
+
+Intel SGX2 on Icelake CPUs supports enclave dynamic memory management (EDMM),
+which many applications of the Open Enclave SDK (OE) will want to take advantage
+of. This document outlines the strategy for making these features available to
+applications.
+
+The key mechanism to support the dynamic addition of memory pages is the
+`ENCLS[EAUG]` instruction, which allocates and zeroes a new memory page via
+communication with the SGX driver. Before the enclave uses the new page, it
+acknowledges its addition by issuing an `ENCLU[EACCEPT]` instruction for the
+respective page address. Further, the `ENCLS[EMODPR]` and `ENCLU[EMODPE]`
+instructions restrict or extend page permissions, respectively. Page type
+changes (including page removal/trimming) are performed by the `ENCLS[EMODT]`
+instruction.
+
+## Heap allocation
+
+We propose to add a new OCALL to OE enclaves which requests new memory pages
+from the SGX driver (via `ioctl`), combined with an enclave function that, upon
+return from the OCALL, accepts the new pages. This does not require handling of
+signals or page faults and is easily integrated with memory managers like the
+default dlmalloc, by pre-mapping a configurable maximum number of pages and
+tracking the number of currently mapped pages. For performance reasons, we
+propose to add a switchless version of the OCALL whenever
+`OE_ENCLAVE_SETTING_CONTEXT_SWITCHLESS` is enabled:
+
+    trusted {
+      public oe_result_t oe_add_pages(size_t count);
+      public oe_result_t oe_add_pages_switchless(size_t count) transition_using_threads;
+    };
+
+The pluggable allocator API then uses `oe_add_pages`, before allocating
+additional pages, if necessary, before calling into `oe_allocator_*alloc`. The
+maximum number of heap pages (and perhaps other parameters like page
+permissions) are passed to `oe_allocator_init`.
+
+Applications that run a full OS kernel, for instance SGX-LKL, need to know about
+the size and layout of their address space and can therefore effectively use
+dynamic region allocation, which the SGX driver supports by tracking memory
+regions in the SGX driver. It automatically grows the heap by allocating a
+configurable number of pages upon request for a particular (unmapped) page.
+Dynamic regions are most conveniently declared as an initialization parameter to
+`oe_allocator_init` again.
+
+The permissions (R/W/X) of pages can be modified dynamically. Permissions of
+mapped pages are restricted by the `ENCLS[EMODPR]` instruction, which is a
+privileged instruction (ring 0) that the host executes. Similarly, permissions
+are extended by the `ENCLU[EMODPE]` instruction, which the enclave executes.
+OE's current pluggable allocators do not support permission changes. We propose
+to add a single, easy to use function to change (both restrict and extend) the
+page types of allocated memory, e.g.
+
+    oe_result_t oe_sgx_set_page_permissions(void* address, bool R, bool W, bool X);
+
+We expect that OE applications will usually want the R bit enabled by default,
+but some applications require the X bit to be set and others may chose to
+relinquish the W flag to avoid accidental modification of memory pages. (One of
+the reasons that SGX-LKL is forced to use a custom version of OE, is that the
+memory permissions are not configurable at signing, launching, or run time; see
+also [#3561](https://github.com/openenclave/openenclave/pull/3561).)
+
+## Extensions
+
+SGX2 supports "lazy" addition of heap and stack memory by triggering and
+catching page faults. However, we believe that most near-term applications will
+not have a strict requirement for this, so we propose not to implement this in
+OE for now. (Intel provides a sketch of a modification to dlmalloc's `sbrk`,
+which automates heap expansion [1].)
+
+Further, SGX2 supports the dynamic creation of thread control structures (TCS)
+by changing the type of allocated pages (via the `ENCLS[EMODT]` instruction).
+While fully dynamic threading is perhaps not a strict requirement for most
+applications, it is a worthwhile consideration to add a generic OE function that
+allows applications to easily change page types, should they have a need to do
+so.
+
+## Signing
+
+The new minimum and maximum memory size settings are signed and attested,
+therefore support for them is added to `oe_enclave_size_settings_t`:
+
+    typedef struct _oe_enclave_size_settings
+    {
+        uint64_t num_heap_pages;
+        uint64_t num_stack_pages;
+        uint64_t num_tcs;
+
+        uint64_t min_heap_pages, max_heap_pages;
+        uint64_t min_stack_pages, max_stack_pages;
+        uint64_t min_tcs, max_tcs;
+    } oe_enclave_size_settings_t;
+
+This automatically adds them to the enclave properties in the `.oeinfo` section.
+The parser for configuration files for the `oesign` tool is extended
+accordingly. While perhaps not necessary immediately, this should include
+minimum and maximum stack and TCS settings in preparation for a future extension
+to dynamic stack and TCS page addition.
+
+# References
+
+[1] Bin (Cedric) Xing, Mark Shanahan, Rebekah Leslie-Hurd: Intel Software Guard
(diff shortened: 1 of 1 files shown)


docs/DesignDocs/SGX2-EDMM.md
```diff
@@ -0,0 +1,105 @@
...
+We propose to add a new OCALL to OE enclaves which requests new memory pages
+from the SGX driver (via `ioctl`), combined with an enclave function that, upon
+return from the OCALL, accepts the new pages. This does not require handling of
+signals or page faults and is easily integrated with memory managers like the
+default dlmalloc, by pre-mapping a configurable maximum number of pages and
+tracking the number of currently mapped pages. For performance reasons, we
+propose to add a switchless version of the OCALL whenever
+`OE_ENCLAVE_SETTING_CONTEXT_SWITCHLESS` is enabled:
+
+    trusted {
```
dthaler, line 28: Line 25 says OCALL but line 28 says trusted which means ECALL

docs/DesignDocs/SGX2-EDMM.md
```diff
@@ -0,0 +1,105 @@
...
+configurable number of pages upon request for a particular (unmapped) page.
+Dynamic regions are most conveniently declared as an initialization parameter to
+`oe_allocator_init` again.
+
+The permissions (R/W/X) of pages can be modified dynamically. Permissions of
+mapped pages are restricted by the `ENCLS[EMODPR]` instruction, which is a
+privileged instruction (ring 0) that the host executes. Similarly, permissions
+are extended by the `ENCLU[EMODPE]` instruction, which the enclave executes.
+OE's current pluggable allocators do not support permission changes. We propose
+to add a single, easy to use function to change (both restrict and extend) the
```
dthaler, line 51: internal or public?  If public, can we make it more TEE-agnostic?

docs/DesignDocs/SGX2-EDMM.md
```diff
@@ -0,0 +1,105 @@
...
+communication with the SGX driver. Before the enclave uses the new page, it
+acknowledges its addition by issuing an `ENCLU[EACCEPT]` instruction for the
+respective page address. Further, the `ENCLS[EMODPR]` and `ENCLU[EMODPE]`
+instructions restrict or extend page permissions, respectively. Page type
+changes (including page removal/trimming) are performed by the `ENCLS[EMODT]`
+instruction.
+
+## Heap allocation
+
+We propose to add a new OCALL to OE enclaves which requests new memory pages
```
dthaler, line 19: internal or exposed to apps?

radhikaj:
SIG-Arch: 
Here are opens-
we need to discuss basic design and goals before we dive into APIs. 
1) How does the host support this feature, interactions between host and kernel. We need to discuss what to build on the kernel. Specific topics - primitives, on demand paging
2) Run time Memory management for EDMM, partitioning memory space

We will need to start discussions starting from building the APIs.

docs/DesignDocs/SGX2-EDMM.md
wintersteiger: Thanks, that was a typo, it should say `untrusted` of course.

docs/DesignDocs/SGX2-EDMM.md
wintersteiger: I'd love to! For now I kept this SGX-specific mainly because I don't have a good enough understanding of other TEE's memory management strategies, but I'm sure they also support at least r/w/x flags in some way. I'll do some reading.

docs/DesignDocs/SGX2-EDMM.md
wintersteiger: I think we should expose it too. We need a default (internal) implementation, but we also need to expose it to allow applications to implement their own strategies. This could conceivably be a part of the pluggable allocator API, but I think that would just end up being an exposed wrapper for an internal-only OCALL.

docs/DesignDocs/SGX2-EDMM.md
dthaler: @wintersteiger https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/OpteeIntegration.md has some good information about OP-TEE.

CodeMonkeyLeet (commented):
Thanks for writing this up, @wintersteiger, I think it's a great strawman for kicking off discussions around EDMM support in OE SDK, but I feel that it's skipping ahead to an API discussion when there are a couple of prerequisite discussions we need to resolve before we can evaluate an API design:

- What is the contract between the host and the enclave with regards to EDMM management?
  - This is currently in flux on the Linux side because Intel has tabled it for the upstreaming of their kernel driver. They have also stated that they expect that the behavior will not be the same as previously proposed/implemented in their SDK and out-of-tree driver. Windows has an implementation, but has not documented it AFAIK.
  - The kind of interaction with the host kernel, and how easy demand paging is going to be to implement on both Windows and Linux will change the calculus for what features OE SDK chooses to expose first. 
  - For example, it seems like implementing a switchless set of APIs may not be reusable in the demand paging scenario and depending on how the host decides to exposes the EDMM functionality, we may choose to prioritize that instead.
  - Per the SIG-Arch meeting, I think there's interest from Intel to start investigating this now as well.

- What is the virtual memory management model in the OE enclave runtime? 
  - Implicit in this design proposal for a public API is some expectation that the OE enclave runtime manages its internal representation of the enclave memory space in a way that permits the caller of the public API to participate in managing that representation. It might be simpler for an initial implementation not to have to deal with that extra complexity.
  - For example, the pluggable allocator design assumes that it owns and manages all heap memory in the enclave. Allocators are usually implemented on top of existing virtual memory primitives provided by the OS, but in this scenario, because the OE runtime itself depends on the allocator, the virtual memory management delegated entirely to the allocator plugin? Or does the allocator sit on top of virtual memory management functions provided by OE that still needs to maintain and extend its virtual address descriptor data structures without itself using malloc/free?
  - Extending that to an app or library OS that's calling virtual memory methods, is the enclave runtime responsible for minimizing virtual address space fragmentation? Does it attempt to reserve a portion of the virtual address space for its own use vs. the app's to avoid interleaving page allocations for different purposes?
  - There's also an expectation that whatever semantics we come up with for OE SDK in this regard would be extendable to an arbitrary TEE (minimally OP-TEE), and I'd like some confirmation on that front as well.
  - For the folks commenting on this PR, I think I would like to start the discussion around requirements for the virtual memory model provided by the OE SDK in each of these domains.

docs/DesignDocs/SGX2-EDMM.md
```diff
@@ -0,0 +1,105 @@
...
+but some applications require the X bit to be set and others may chose to
+relinquish the W flag to avoid accidental modification of memory pages. (One of
+the reasons that SGX-LKL is forced to use a custom version of OE, is that the
+memory permissions are not configurable at signing, launching, or run time; see
+also [#3561](https://github.com/openenclave/openenclave/pull/3561).)
+
+## Extensions
+
+SGX2 supports "lazy" addition of heap and stack memory by triggering and
+catching page faults. However, we believe that most near-term applications will
```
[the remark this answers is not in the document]
MWShan: @CodeMonkeyLeet - I agree with your assessment. I think that we can break the task down into a number of subtasks (which do focus on the virtual memory management).  Here is my list:

1. Kernel APIs
   1. Linux : not supported in kernel. Need to make a proposal for support (and work on the assumption that the final solution will be close to the proposal)
2. Enclave Common Loader: abstracts kernel APIs into a single set of APIs
3. Core Design - the basic memory management:
   1. Enclave Memory Mapping Structures: the enclave needs to know its own memory in order to know what it should and should not EACCEPT/EACCEPTCOPY
        1. Virtual Address Structures - enclave needs to know properties of each memory address region (similar to VMAs)
            1. RWX
            2. Page Type
            3. Properties (e.g GROWS_DOWN/GROWS_UP)
        2. Physical Page - enclave needs to keep track of information such as
            1. Pages which are EACCEPT'ed, TRIM'ed, etc.
    2. Low Level Functions (within the enclave):
        1. Primitives: Assembly Routines to implement SGX Instructions
            1. EACCEPT
            2. EACCEPTCOPY
            3. EMODPE
        2. Low Level: APIs that are presented to the enclave to manage memory. These will include OCalls to untrusted host side to make calls to the kernel functions.
            1. Allocate Pages - similar to mmap(.., MAP_ANONYMOUS, ..) and/or VirtualAlloc().
                1. Include APIs to free pages.
                2. Potentially need options to eagerly/lazily EACCEPT and an EACCEPTCOPY option
            2. Change Page Protection/Type - similar to mprotect() and/or VirtualProtect()
    3. Configuration: How is an enclave configured for core design
        1. Enclave range - the virtual size of the enclave.
4. EDMM Enabled Features - enabled by the Low Level Functions which is part of core design. This should also include configuration and design of high level functions. These do not need to be completed at the same time but should be considered to make sure that the Core Design can support these (and more features that EDMM can enable):
    1. Reserving/Committing Memory outside of the heap (provided in Lower Level APIs) - similar to mmap()/munmap()
    2. Changing page protections (provided in Lower Level APIs) - similar to mprotect()
    3. Heap Expansion/Contraction - may need sbrk() implementation (instead of mmap)
    4. Stack Expansion - may be complicated since this must be done without a stack
    5. Thread Creation/Deletion : somewhat complex because this must create/initialize many pages
        1. Must add pages, change page type (for TCS), etc.
    6. Dynamic Code Loading

haitaohuang:
Regarding #1 sub-task mentioned by @MWShan, created https://github.com/openenclave/openenclave/pull/3755 for a draft driver interface for EDMM support.

docs/DesignDocs/SGX2-EDMM.md
```diff
@@ -0,0 +1,105 @@
...
+
+    oe_result_t oe_sgx_set_page_permissions(void* address, bool R, bool W, bool X);
+
+We expect that OE applications will usually want the R bit enabled by default,
+but some applications require the X bit to be set and others may chose to
+relinquish the W flag to avoid accidental modification of memory pages. (One of
+the reasons that SGX-LKL is forced to use a custom version of OE, is that the
+memory permissions are not configurable at signing, launching, or run time; see
+also [#3561](https://github.com/openenclave/openenclave/pull/3561).)
+
```
haitaohuang, line 62: @wintersteiger could you elaborate on call flow in SGX-LKL for permission changes? Will it load code into a heap area, then change permissions? The referred PR is closed and you stated it won't be an issue with EDMM.
I'd like to confirm the APIs proposed in #3991 for enclave side would be sufficient.

wintersteiger: In the past, SGX-LKL simply added the X flag to the entire heap at enclave build time and at startup. This was done via fork of OE which had the extra `SGX_SECINFO_X` added to the pages set up in `_add_filled_pages` in `create.c`, so whatever it loaded into that memory was also executable. This is not necessary with EDMM because page permissions can be changed after enclave startup, as indeed the API in #3991 includes. I'm not sure whether there will be an EDMM version of SGX-LKL, but if so, it would be a trivial change, so I think we're fine there.

radhikaj:
@wintersteiger, closing this PR since we will follow the strategy laid out in https://github.com/openenclave/openenclave/pull/3991. Please comment on that PR.

Issue #3530, opened 2020-09-13
Win2019 Release Cross Compile DCAP LVI snmalloc unstable on new Intel release


BRMcLaren:
*Opening this to track and not lose, needs more initial investigation before triage, please skip on Monday's Triage.

It seems that the Win2019 Release Cross Compile DCAP LVI snmalloc is not stable on the latest Intel PSW release causing issues with the CI rebuild as 9/10 times it fails with 

```
[5/14] Building C object host\CMakeFiles\file-encryptor_host.dir\fileencryptor_u.c.obj
[6/14] Building CXX object enclave\CMakeFiles\enclave.dir\ecalls.cpp.obj
[7/14] Building CXX object enclave\CMakeFiles\enclave.dir\encryptor.cpp.obj
[8/14] Building C object enclave\CMakeFiles\enclave.dir\fileencryptor_t.c.obj
[9/14] Building CXX object enclave\CMakeFiles\enclave.dir\keys.cpp.obj
[10/14] Linking CXX executable enclave\enclave
[11/14] Building CXX object host\CMakeFiles\file-encryptor_host.dir\host.cpp.obj
..\host\host.cpp(314): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
..\host\host.cpp(147): warning C4996: 'fopen': This function or variable may be unsafe. Consider using fopen_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
C:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\ucrt\stdio.h(208): note: see declaration of 'fopen'
..\host\host.cpp(162): warning C4996: 'fopen': This function or variable may be unsafe. Consider using fopen_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
C:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\ucrt\stdio.h(208): note: see declaration of 'fopen'
[12/14] Generating private.pem, public.pem
Generating RSA private key, 3072 bit long modulus (2 primes)
..................++++
........++++*....................................................................++++*.............................++++
e is 3 (0x03)
writing RSA key
[13/14] Generating enclave/enclave.signed
Created C:/oe/open-enclave/openenclave/share/openenclave/samples/file-encryptor/build/enclave/enclave.signed
[14/14] Linking CXX executable host\file-encryptor_host.exe
[...]

PR #3641, opened 2020-10-06
Add support for pthread_spin_trylock()

### LINKED ISSUE #3483: Why is pthread_spin_trylock() not provided?
Open Enclave provides partial support for standard POSIX/LibC functions. The current support is listed [here](https://github.com/openenclave/openenclave/blob/master/docs/LibcSupport.md). This list does not include `pthread_spin_trylock()`. All other `pthread_spin_` functions _are_ supported, as far as I can tell only `trylock` is missing

Why is this? Is there a technical reason why `trylock` is not supportable? Or is this an accidental omission?

From a quick read of the code it looks like the `oe_spin_lock` implementation could be `oe_spin_trylock` with only slight modification, and this could be wrapped/exposed as `pthread_spin_trylock` in the same manner as the other `pthread_spin_` functions.

larrywxyz:
Closes #3483

### DIFF
--- a/enclave/core/optee/spinlock.c
+++ b/enclave/core/optee/spinlock.c
@@ -28,6 +28,14 @@ oe_result_t oe_spin_lock(oe_spinlock_t* spinlock)
     return OE_OK;
 }
 
+oe_result_t oe_spin_trylock(oe_spinlock_t* spinlock)
+{
+    if (!spinlock)
+        return OE_INVALID_PARAMETER;
+
+    return OE_OK;
+}
+
 oe_result_t oe_spin_unlock(oe_spinlock_t* spinlock)
 {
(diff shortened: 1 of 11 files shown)


CodeMonkeyLeet (commented):
Thanks for addressing this gap! Can you also please add a test for this method to tests/thread?

achamayou (approved):

achamayou:
bors try

CodeMonkeyLeet (approved):

achamayou:
Azure Linux failed with a timeout on tests/create-rapid:

image

The Windows build hit this:

image

which looks like #3530.

Neither of those is obviously related to the change, so running the CI again.

PR #3642, opened 2020-10-06
Fix improper size_t type usage

### LINKED ISSUE #3377: OE checks whether a size_t variable is negative
On this [line](https://github.com/openenclave/openenclave/blob/80309da445e768d05031b14ef0e33997662a7769/enclave/crypto/cert.c#L1180), we are checking if a size_t variable `bytes_written` is <= 0.  So even if the function `mbedtls_x509write_crt_der` returns a negative number to indicate error, we will never detect it.

We are lucky that `mbedtls_x509write_crt_der` never failed until today when I tried to generate an EEID based certificate and that triggered `MBEDTLS_ERR_ASN1_BUF_TOO_SMALL`.

larrywxyz:
Fixes #3377

### DIFF
--- a/enclave/crypto/cert.c
+++ b/enclave/crypto/cert.c
@@ -1175,10 +1175,15 @@ oe_result_t oe_gen_custom_x509_cert(
     // Write a built up certificate to a X509 DER structure Note: data
     // is written at the end of the buffer! Use the return value to
     // determine where you should start using the buffer.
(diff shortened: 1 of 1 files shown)


CodeMonkeyLeet (approved):

radhikaj (approved):
Thanks @larrywxyz

PR #3654, opened 2020-10-08
Update oeedger8r submodule

anakrish:
- Remove tests for deep-copy out parameters.
  These were never properly supported.
- Various fixes
  - Validations
  - Retain lhs size/count
  - String null checking


anakrish:
bors try

mingweishih (approved):
Thanks for doing this. Should we update the CHANGELOG as well? I'm fine if we update it in a separated PR.

CodeMonkeyLeet (approved):
Regarding the CHANGELOG update, I would suggest that any changes to the oeedger8r tool that is developer facing (e.g. if the validation fixes will now trap errors it did not previously) that a some description of what to expect and how the changes affect the dev experience be included.

anakrish:
Updated CHANGELOG.md

PR #3659, opened 2020-10-12
Add basic KSS support (ISVFAMILYID and ISVEXTPRODID) to OE SDK (Replaced by #3700) 

jianlia:

Enclave loader to detect whether the system support KSS, to parse the Enclave SIGSTRUCT.ATTRIBUTES and SIGSTRUCT.ATTRIBUTEMASK and decide whether to enable KSS


docs/GettingStartedDocs/buildandsign.md
```diff
@@ -97,6 +97,15 @@ NumHeapPages=1024
 NumStackPages=1024
 NumTCS=1
 ```
+Additionally, for platform that supports for Key sharing and separation (KSS),
```
dthaler, line 100:
```suggestion
Additionally, for platforms that support Key Sharing and Separation (KSS),
```

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -97,6 +97,15 @@ NumHeapPages=1024
...
+the following properties can be used:
```
dthaler, line 101:
```suggestion
the following additional properties can be used:
```

Also explain what happens if these are included and then the enclave is run on a non KSS machine.   Will it fail to load?  Will the properties simply be ignored?  I.e., does the developer need to compile the enclave twice, once for KSS and once without, to run on both types of platforms?

docs/GettingStartedDocs/buildandsign.md
dthaler: How does one know whether KSS is supported?  Maybe add a link to which document says. E.g., is it reported by the oesgx tool?

host/sgx/sgxsign.c
```diff
@@ -572,6 +572,8 @@ static oe_result_t _init_sigstruct(
     uint64_t attributes,
     uint16_t product_id,
     uint16_t security_version,
+    const uint8_t* family_id,
+    const uint8_t* ext_product_id,
```
dthaler, line 576: OE coding style is to avoid abbreviations.
```suggestion
    const uint8_t* extended_product_id,
```

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -31,17 +31,20 @@ typedef struct _oe_sgx_enclave_image_info_t
...
 #define OE_SGX_FLAGS_MODE64BIT 0x0000000000000004ULL
+#define OE_SGX_FLAGS_KSS 0x0000000000000080ULL
 #define OE_SGX_SIGSTRUCT_SIZE 1808
 
 typedef struct oe_sgx_enclave_config_t
 {
     uint16_t product_id;
     uint16_t security_version;
+    uint8_t family_id[16];
+    uint8_t ext_product_id[16];
```
dthaler, line 42:
```suggestion
    uint8_t extended_product_id[16];
```

include/openenclave/internal/sgxsign.h
```diff
@@ -36,6 +38,8 @@ oe_result_t oe_sgx_sign_enclave(
     uint16_t security_version,
     const uint8_t* pem_data,
     size_t pem_size,
+    const uint8_t* family_id,
+    const uint8_t* ext_product_id,
```
dthaler, line 42:
```suggestion
    const uint8_t* extended_product_id,
```

include/openenclave/internal/sgxsign.h
```diff
@@ -65,6 +71,8 @@ oe_result_t oe_sgx_sign_enclave_from_engine(
     const char* engine_id,
     const char* engine_load_path,
     const char* key_id,
+    const uint8_t* family_id,
+    const uint8_t* ext_product_id,
```
dthaler, line 75:
```suggestion
    const uint8_t* extended_product_id,
```

include/openenclave/internal/sgxsign.h
```diff
@@ -84,6 +94,8 @@ oe_result_t oe_sgx_get_sigstruct_digest(
     uint64_t attributes,
     uint16_t product_id,
     uint16_t security_version,
+    const uint8_t* family_id,
+    const uint8_t* ext_product_id,
```
dthaler, line 98:
```suggestion
    const uint8_t* extended_product_id,
```

tests/tools/oesign/test-enclave/enclave/enc.c
```diff
@@ -52,3 +53,61 @@ bool is_test_signed()
 
     return is_test_signed;
 }
+
+oe_result_t check_kss_extended_ids(
+    oe_uuid_t* family_id,
+    oe_uuid_t* ext_product_id)
```
dthaler, line 59:
```suggestion
    oe_uuid_t* extended_product_id)
```

tests/tools/oesign/test-enclave/host/host.c
```diff
@@ -19,6 +39,19 @@ int main(int argc, const char* argv[])
         oe_put_err("Usage: %s enclave_image_path\n", argv[0]);
     }
 
+    /* Load the ELF image */
+    if ((result = oe_load_enclave_image(argv[1], &oeimage)) != OE_OK)
+    {
+        oe_put_err("oe_load_enclave_image(): result=%u", result);
+    }
+
+    /* Load the SGX enclave properties */
```
dthaler, line 48: Prior to this PR, this file is TEE-agnostic.  I think we should not change it to be single TEE only.  At minimum, I think this needs to test whether it's being signed for SGX before introducing SGX-only code into this code path.

tests/tools/oesign/test-sign/CMakeLists.txt
```diff
@@ -203,3 +203,65 @@ add_test(
 
 set_tests_properties(tests/oesign-sign-lowercase-debug-property
                      PROPERTIES PASS_REGULAR_EXPRESSION "unknown setting")
+
+# Test valid kss properties argument
+set(OESIGN_SIGN_VALID_KSS
+    "[-c,${OESIGN_TEST_INPUTS_DIR}/kss_valid.conf,-k,${OESIGN_TEST_INPUTS_DIR}/sign_key.private.pem]"
+)
```
dthaler, line 210: Since this directory does not have 'sgx' in the title, I would expect there to be some "if sgx" check before settings such things.

tools/oesign/oesign.c
```diff
@@ -671,3 +784,76 @@ int oedigest(const char* enclave, const char* conffile, const char* digest_file)
...
+    if (ch >= '0' && ch <= '9')
+        return ch - '0';
+    if (ch >= 'a' && ch <= 'f')
+        return 10 + ch - 'a';
+    if (ch >= 'A' && ch <= 'F')
+        return 10 + ch - 'A';
+    return 0;
+}
+
+ bool is_hex(char ch)
```
dthaler, line 799: You shouldn't need this function, just use the standard isxdigit() from ctype.h

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -31,17 +31,20 @@ typedef struct _oe_sgx_enclave_image_info_t
...
 #define OE_SGX_FLAGS_DEBUG 0x0000000000000002ULL
 #define OE_SGX_FLAGS_MODE64BIT 0x0000000000000004ULL
+#define OE_SGX_FLAGS_KSS 0x0000000000000080ULL
 #define OE_SGX_SIGSTRUCT_SIZE 1808
 
 typedef struct oe_sgx_enclave_config_t
 {
     uint16_t product_id;
     uint16_t security_version;
+    uint8_t family_id[16];
```
yentsanglee, line 41: The `padding` at line 45 is a padding for `product_id` and `security_version` to ensure variables do not go across 64-byte boundary. I'd put `padding` immediately after `product_id` and `security_version`.

yentsanglee:
>     OE_CHECK(oe_sgx_validate_enclave_properties(&props, NULL));

Check family_id and ext_product_id

---
Refers to: host/sgx/create.c:794 in ff3d54c. [](commit_id = ff3d54cc9eb7ee7f2b47b1865125fc53b4f4cd77, deletion_comment = False)

host/sgx/create.c
```diff
@@ -832,6 +832,11 @@ oe_result_t oe_sgx_build_enclave(
         &loaded_enclave_pages_size,
         &enclave_size));
 
+    if (props.config.attributes & OE_SGX_FLAGS_KSS)
+    {
+        context->attributes.flags |= OE_SGX_FLAGS_KSS;
```
yentsanglee, line 837: Today the only possible bitmask flags for `context->attributes.flags` are `OE_ENCLAVE_FLAG_DEBUG (1)` and `OE_ENCLAVE_FLAG_SIMULATE (2)`. XOR it with OE_SGX_FLAGS_KSS (0x80) has no conflict today but it may in the future. I'd suggest we define another KSS flag value to go with `OE_ENCLAVE_FLAG_DEBUG (1)` and `OE_ENCLAVE_FLAG_SIMULATE (2)` but don't publish it.

CodeMonkeyLeet (commented):
Thanks for bringing this back! I carried over some unaddressed comments from the previous, and we need to close on the question of whether KSS support is only specifiable at sign-time, or whether the same sigstruct can support being run with and with KSS.

host/sgx/create.c
CodeMonkeyLeet: Another good catch @yentsanglee! I think this bug is not part of this code, but an existing one in [`oe_create_enclave`](https://github.com/openenclave/openenclave/blob/4ecb952ebca19932d28892ac2c84ba18a48574a2/host/sgx/create.c#L1018), where it's calling `oe_sgx_initialize_load_context` directly with flags using the `OE_ENCLAVE_FLAG_*` definitions, but the function (and the `context.attributes` member) expects flags defined using `OE_SGX_FLAGS_*`.

It should be the responsibility of the `oe_create_enclave` method to interpret the `OE_ENCLAVE_FLAG_*` into the appropriate `OE_SGX_FLAGS_DEBUG` (and whatever mechanism it will use to set the `OE_SGX_FLAGS_KSS`) before calling `oe_sgx_initialize_load_context` and the downstream methods should only have to deal with the `OE_SGX_FLAGS_*`.

CodeMonkeyLeet:
@jianlia One additional observation about the PR: can you rebase your branch against master instead of merging in changes from master? The commit 0a547bf for oeedger8r and other changes originating from master branch should not be part of the commit history for this PR.

include/openenclave/bits/sgx/sgxproperties.h
yentsanglee: @CodeMonkeyLeet Yeah, I meant 64-bit aligned. Thanks for the correction.

include/openenclave/internal/sgx/sgxproperties.h
```diff
@@ -32,7 +32,7 @@ OE_INLINE bool oe_sgx_is_valid_num_tcs(uint64_t x)
 OE_INLINE bool oe_sgx_is_valid_attributes(uint64_t x)
```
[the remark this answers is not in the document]
yentsanglee: I had a similar comment - "oe_sgx_validate_enclave_properties() should check family_id and ext_product_id too"

yentsanglee:
> > ```
> > OE_CHECK(oe_sgx_validate_enclave_properties(&props, NULL));
> > ```
> 
> oe_sgx_validate_enclave_properties() should check family_id and ext_product_id too
> 
> Refers to: host/sgx/create.c:794 in [ff3d54c](https://github.com/openenclave/openenclave/commit/ff3d54cc9eb7ee7f2b47b1865125fc53b4f4cd77). [](commit_id = [ff3d54c](https://github.com/openenclave/openenclave/commit/ff3d54cc9eb7ee7f2b47b1865125fc53b4f4cd77), deletion_comment = False)

family_id or ext_product_id can be any value. Nothing is considered invalid. The check is unnecessary.

include/openenclave/internal/sgx/sgxproperties.h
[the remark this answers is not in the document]
yentsanglee: family_id or ext_product_id can be any value. Nothing is considered invalid. The check is unnecessary.

tools/oesign/oesign.c
```diff
@@ -378,6 +467,24 @@ void _merge_config_file_options(
     if (options->security_version.has_value)
         properties->config.security_version = options->security_version.value;
 
+    bool kss_supported = _is_kss_supported();
```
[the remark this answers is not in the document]
bodzhang: With regard to Intel SDM vol. 3d description of EINIT, I think the documentation missed the ISVEXTPRODID check. Can someone from Intel confirm?

I think many developers won't be fully aware that using non-zero ISVFAMILYID/ISVEXTPRODID implies the SGX enclave will only run on SGX CPU with KSS support. But I think the tool chain (oesign and/or the loader) can and should own SIGSTRUCT.ATTRIBUTES.KSS, SIGSTRUCT.ATTRIBUTEMASK.KSS and SECS.ATTRIBUTE.KSS setting on behalf of the developer.

SGX CPU with KSS support enforces the following policy:

- If (SIGSTRUCT.ATTRIBUTEMASK.KSS ==0), loader can set SECS.ATTRIBUTES.KSS as 1 or 0. 

- If (SIGSTRUCT.ATTRIBUTES.KSS ==1 AND SIGSTRUCT.ATTRIBUTEMASK.KSS ==1), loader must set SECS.ATTRIBUTES.KSS as 1.

- If (SIGSTRUCT.ATTRIBUTES.KSS ==0 AND SIGSTRUCT.ATTRIBUTEMASK.KSS ==1), loader must set SECS.ATTRIBUTES.KSS as 0.  

- If SIGSTRUCT contains non-zero ISVFAMILYID/ISVEXTPRODID, loader must set SECS.ATTRIBUTES.KSS as 1

- If loader sets SECS with non-zero CONFIGID/CONFIGSVN, loader must also set SECS.ATTRIBUTES.KSS as 1

SGX CPU without KSS support enforces the following policy:

- SIGSTRUCT.ATTRIBUTES.KSS, SIGSTRUCT.ATTRIBUTEMASK.KSS must be 0

- SIGSTRUCT.ISVFAMILYID, SIGSTRUCT.ISVEXTPRODID must be 0

- SECS.ATTRIBUTES.KSS, SECS.CONFIGID, SECS.CONFIGSVN must be 0

OE loader should enumerates SGX CPU support for KSS, and set SECS according to the CPU capability, SIGSTRUCT and to not violate the CPU enforced policy listed above. With regard to SECS.ATTRIBUTES.KSS setting (on CPU with KSS support) the loader can follow a simple logic:

- Set SECS.ATTRIBUTES.KSS = 1, if the the CPU supports KSS, unless SIGSTRUCT. ATTRIBUTES.KSS=0 and SIGSTRUCT.ATTRIBUTEMASK.KSS=1.

If the developer does not intend to use any KSS related feature, the loader enabling KSS (on CPU with KSS support) and setting SECS.CONFIGID, SECS.CONFIGSVN as 0 does not violate any function or security policy of the enclave. (
The proposed [OE_ENCLAVE_SETTING_CONTEXT_PRE_MEASURED_ID data for CONFIGID/CONFIGSVN](https://github.com/openenclave/openenclave/pull/3269) decide what values to use for SECS.CONFIGID, SECS.CONFIGSVN when the developer intends to use the CONFIGID/CONFIGSVN feature of KSS. But it's out of scope of this PR).

Oesign should set SIGSTRUCT.ATTRIBUTES.KSS =0 and SIGSTRUCT.ATTRIBUTEMASK.KSS =0 by default, unless the developer sets non-zero ISVFAMILYID/ISVEXTPRODID. In that case, oesign sets SIGSTRUCT.ATTRIBUTES.KSS =1 and SIGSTRUCT.ATTRIBUTEMASK.KSS =1. The signing process should also allow the developer to use advanced interface to set SIGSTRUCT.ATTRIBUTES.KSS =0 and SIGSTRUCT.ATTRIBUTEMASK.KSS =1.

With SIGSTRUCT.ATTRIBUTES.KSS =0 and SIGSTRUCT.ATTRIBUTEMASK.KSS =0, developers not aware of KSS will be building enclaves that run on both SGX CPUs with and without KSS support.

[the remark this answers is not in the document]
CodeMonkeyLeet: Thanks for the write-up @bodzhang! To summarize my understanding, the difference from the previous 3 scenarios I laid out is that opting into KSS for ISVFAMILYID/ISVEXTPRODID constrains what the loader can do at runtime, and the latter can be implemented on top of of this. To phase this work out, I would suggest an ordering of implementation based on the semantics you've defined:

1. If a developer/signer uses ISVFAMILYID/ISVEXTPRODID, resulting enclave should have `SIGSTRUCT.ATTRIBUTES.KSS=1` and `SIGSTRUCT.ATTRIBUTEMASK.KSS=1`
   - This should be implemented as part of this PR, and should include both `OE_SET_ENCLAVE_SGX` and oesign codepaths.
   - OE SDK uses hardcoded `SGX_SIGSTRUCT_ATTRIBUTEMASK_FLAGS` today, where `SIGSTRUCT.ATTRIBUTEMASK.KSS=1` is true by default, so the PR only has to deal with setting `SIGSTRUCT.ATTRIBUTES.KSS=1` or not.
   - This PR needs to add an update to `OE_MAKE_ATTRIBUTES` or an equivalent implmentation for `OE_SET_ENCLAVE_SGX` to be consistent with the oesign.

2. By default, if a developer/signer does not specify ISVFAMILYID/ISVEXTPRODID, resulting enclave should have `SIGSTRUCT.ATTRIBUTES.KSS=0` and `SIGSTRUCT.ATTRIBUTEMASK.KSS=1`
   - Note that I have inverted the default proposed by Bo here, and I'm suggesting that we should not change the default in this PR.
   - The existing SDK still does not support CONFIGID/CONFIGSVN, so it should be treated as an unintended change in identity if the enclave is loaded by some loader that permits setting those properties at runtime.
   - The `SIGSTRUCT.ATTRIBUTES.KSS=0` and `SIGSTRUCT.ATTRIBUTEMASK.KSS=1` combination specifies intent of blocking that unintended scenario, and represents a safe default consistent with older versions of the SDK until CONFIGID/CONFIGSVN is supported. At that point, we can also revisit changing the default KSS attribute mask as a breaking change.

3. The developer/signer can specify that they do not care if the enclave is loaded with CONFIGID/CONFIGSVN and the resulting enclave should have `SIGSTRUCT.ATTRIBUTES.KSS=0` and `SIGSTRUCT.ATTRIBUTEMASK.KSS=0`.
   - Whether this becomes the default or not, it is an additional option we need to expose to a developer who doesn't use ISVFAMILYID/ISVEXTPRODID but may want their enclave to run with CONFIGID/CONFIGSVN.
   - I am in favor of making this an explicit opt-in because it is not obvious to a developer that their enclave may run with additional CONFIGID/CONFIGSVN properties.
      - On the flip side, this doesn't appear to be immediately harmful to the enclave since any attestation system that doesn't expect to use the CONFIGID/CONFIGSVN won't interact with it anyway. It may be undesirable for other reasons as it allows the host to inject data into the report which might be used for packet tagging perhaps.
   - In terms of code change, this requires changing the existing hardcoded `SGX_SIGSTRUCT_ATTRIBUTEMASK_FLAGS` so that the SIGSTRUCT.ATTRIBUTEMASK can be set dynamically by `OE_SET_ENCLAVE_SGX` and oesign.

4. There's an additional narrow case of allowing a developer/signer to specify that enclave _must_ run with CONFIGID/CONFIGSVN by specifying `SIGSTRUCT.ATTRIBUTES.KSS=1` and `SIGSTRUCT.ATTRIBUTEMASK.KSS=1` when ISVFAMILYID/ISVEXTPRODID is not provided. 
   - I think there's some additional design discussion for when CONFIGID/CONFIGSVN is added as to whether we need to support this. Off the top of my head, I can't think of a really strong reason to require this, but it's also not a lot of work. The real question is explaining to a dev why they would want to do this.
   - If we do support this case, `oe_create_enclave` should provide check and provide a specific error trace in this configuration and it detects none of the KSS properties are defined. 

5. For completeness, there should not be a case that `OE_SET_ENCLAVE_SGX` or oesign needs to support for setting `SIGSTRUCT.ATTRIBUTES.KSS=1` and `SIGSTRUCT.ATTRIBUTEMASK.KSS=0` at development/sign time.
   - This basically means "I want the loader to use KSS by default, but it can choose not to" and is no different in practical terms from setting `SIGSTRUCT.ATTRIBUTES.KSS=0` and `SIGSTRUCT.ATTRIBUTEMASK.KSS=0`, so we should standardize on that encoding instead.
      - The loader doesn't need `SIGSTRUCT.ATTRIBUTES.KSS=1` for it to set `SECS.ATTRIBUTES.KSS=1` in that case.
      - If the dev/signer uses `ISVFAMILYID/ISVEXTPRODID`, it becomes case (1) as it is not legal to set `SIGSTRUCT.ATTRIBUTEMASK.KSS=0`

host/sgx/create.c
jianlia: Added OE_SGX_FLAGS_KSS flag

host/sgx/sgxsign.c
jianlia: Updated to align with OE coding style

tools/oesign/oesign.c
```diff
@@ -671,3 +784,76 @@ int oedigest(const char* enclave, const char* conffile, const char* digest_file)
...
+            firstDigit = true;
+        }
+    }
+    if (index < expected_size)
+    {
+        free(id_copy);
+        return -1;
+    }
+    return 0;
+}
```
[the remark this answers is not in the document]
jianlia: Updated

tools/oesign/oesign.c
```diff
@@ -671,3 +784,76 @@ int oedigest(const char* enclave, const char* conffile, const char* digest_file)
 done:
     return ret;
 }
+
+char hexchar2int(char ch)
```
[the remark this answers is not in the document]
jianlia: Will address this in a separate PR

tests/tools/oesign/test-sign/CMakeLists.txt
jianlia: Related code move in block when KSS is enabled so setting will not applied if KSS not supported

tests/tools/oesign/test-sign/CMakeLists.txt
```diff
@@ -203,3 +203,65 @@ add_test(
 
 set_tests_properties(tests/oesign-sign-lowercase-debug-property
```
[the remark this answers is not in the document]
jianlia: Done

tests/tools/oesign/test-enclave/enclave/enc.c
jianlia: Updated per suggestion.

include/openenclave/internal/sgxsign.h
jianlia: Updated  per suggestion

host/sgx/sgxsign.c
```diff
@@ -640,6 +642,22 @@ static oe_result_t _init_sigstruct(
         mrenclave,
         sizeof(*mrenclave)));
 
+    if (attributes & SGX_FLAGS_KSS)
```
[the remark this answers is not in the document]
jianlia: Added zero check below

include/openenclave/internal/sgxsign.h
jianlia: Updated per comment

tools/oesign/oesign.c
```diff
@@ -378,6 +465,26 @@ void _merge_config_file_options(
...
+
+        if (options->extended_product_id.has_value)
+            memcpy(
+                properties->config.extended_product_id,
+                &options->extended_product_id.value,
+                sizeof(options->extended_product_id.value));
+
+        if (options->family_id.has_value ||
+            options->extended_product_id.has_value)
+            properties->config.attributes |= OE_SGX_FLAGS_KSS;
```
yentsanglee, line 485: Even though the 2 definitions are identical, you should use SGX_FLAGS_KSS instead of OE_SGX_FLAGS_KSS to be consistent with other attributes assignments in this file.

host/sgx/sgxsign.c
```diff
@@ -640,6 +642,31 @@ static oe_result_t _init_sigstruct(
...
+                sigstruct->isvfamilyid,
+                sizeof(sigstruct->isvfamilyid),
+                family_id,
+                sizeof(*family_id)));
+        if (ext_product_id)
+            OE_CHECK(oe_memcpy_s(
+                sigstruct->isvextprodid,
+                sizeof(sigstruct->isvextprodid),
+                ext_product_id,
+                sizeof(*ext_product_id)));
```
yentsanglee, line 658: Do we have to set `sigstruct->attributemask.flags` to SGX_FLAGS_KSS bit here?

host/sgx/create.c
yentsanglee: Possible bits for `context->attributes.flags` are:

```
#define OE_ENCLAVE_FLAG_DEBUG 0x00000001u
#define OE_ENCLAVE_FLAG_SIMULATE 0x00000002u
#define OE_ENCLAVE_FLAG_KSS 0x00000004u
```

Line 837 should set the bit of newly defined flag `OE_ENCLAVE_FLAG_KSS`. Without this change, the following statement will never be satisfied.

>> •	If SIGSTRUCT contains non-zero ISVFAMILYID/ISVEXTPRODID, loader must set SECS.ATTRIBUTES.KSS as 1

include/openenclave/host.h
```diff
@@ -65,6 +65,11 @@ OE_EXTERNC_BEGIN
  */
 #define OE_ENCLAVE_FLAG_SIMULATE 0x00000002u
 
+/**
+ *  Flag passed into oe_create_enclave to run the enclave with kss config.
+ */
+#define OE_ENCLAVE_FLAG_KSS 0x00000004u
```
yentsanglee, line 71: This flag is reserved for internal use. We do not allow oe_create_enclave() call to specifically set this KSS flag. 

Suggestion: Reword the comment and move it to line 78 (within @cond DEV) so DOXYGEN ignores it.

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -31,6 +31,7 @@ typedef struct _oe_sgx_enclave_image_info_t
 // oe_sgx_enclave_properties_t SGX enclave properties derived type
 #define OE_SGX_FLAGS_DEBUG 0x0000000000000002ULL
 #define OE_SGX_FLAGS_MODE64BIT 0x0000000000000004ULL
+#define OE_SGX_FLAGS_KSS 0x0000000000000008ULL
```
yentsanglee, line 34: This has to match SGX_FLAGS_KSS which is

`#define SGX_FLAGS_KSS            0x0000000000000080ULL     /* If set enclave uses KSS */`

host/sgx/sgxsign.c
jianlia: #define SGX_SIGSTRUCT_ATTRIBUTEMASK_FLAGS 0XfffffffffffffffbULL
Currently we mask all bits except bit 2 for MODE64BIT so I did not set it but it is a good call and I will set it explicitly.

include/openenclave/bits/sgx/sgxproperties.h
jianlia: Sure, I will reset it to SGX_FLAGS_KSS value. What is the context we have OE_SGX_FLAGS_xxx so have to maintain two set of flag values?

tools/oesign/oesign.c
jianlia: Updated

include/openenclave/host.h
jianlia: Updated

include/openenclave/bits/sgx/sgxproperties.h
yentsanglee: > Sure, I will reset it to SGX_FLAGS_KSS value. What is the context we have OE_SGX_FLAGS_xxx so have to maintain two set of flag values?

@CodeMonkeyLeet Why are we keeping 2 sets of similar definitions?

include/openenclave/host.h
```diff
@@ -68,6 +68,12 @@ OE_EXTERNC_BEGIN
 /**
  * @cond DEV
  */
+
+/**
+ *  Flag passed into oe_create_enclave to run the enclave with kss mode.
```
yentsanglee, line 73: Flag reserved for internal use to indicate the enclave was configured to run in kss mode

tools/oesign/oesign.c
```diff
@@ -378,6 +465,29 @@ void _merge_config_file_options(
     if (options->security_version.has_value)
         properties->config.security_version = options->security_version.value;
 
+    if (_is_kss_supported())
```
yentsanglee, line 468: Enclave can be signed on non-SGX machines. Don't check `_is_kss_supported()` here.

host/sgx/create.c
```diff
@@ -832,6 +832,11 @@ oe_result_t oe_sgx_build_enclave(
         &loaded_enclave_pages_size,
         &enclave_size));
 
+    if (props.config.attributes & OE_SGX_FLAGS_KSS)
```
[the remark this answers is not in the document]
CodeMonkeyLeet: @jianlia This comment hasn't been addressed.

include/openenclave/internal/sgxsign.h
CodeMonkeyLeet: Can you do a global find and replace for this? In this iteration, the only fixed value I see is on line 75 and there are many instances in the EDL, comments and functions that remain `ext_product_id`

tools/oesign/oesign.c
CodeMonkeyLeet: > The oesign tool can't rely on runtime environment checking to set this bit when setting the KSS support flag since enclaves can be signed on non-SGX machines, not just non-KSS supporting machines. oesign should just eliminate the check for _is_kss_supported. 

This still needs to be addressed.

tools/oesign/oesign.c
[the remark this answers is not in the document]
CodeMonkeyLeet: Please file an issue for tracking that if you do.

tools/oesign/oesign.c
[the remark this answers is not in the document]
CodeMonkeyLeet: Still not addressed in the latest iteration

tests/tools/oesign/test-enclave/host/host.c
```diff
@@ -49,6 +69,35 @@ int main(int argc, const char* argv[])
         }
     }
 
+    if (_is_kss_supported())
```
CodeMonkeyLeet, line 72: If this is passing in bors, I suspect that's a side effect of oesign only signing with KSS flag on KSS-enabled build machines, which is a bug. 

I think you're going to restructure this test, because once enclaves can be signed on non-KSS machines with KSS properties, the `oe_create_enclave` on line 45 will fail before it gets to this check, so you're going to have to introduce a mechanism to skip entire tests based on detection of this bit, and a parameter passed into the host to indicate whether a specific test expects KSS or not. 

See tests/debug-mode for an example.

host/sgx/sgxsign.c
```diff
@@ -640,6 +642,33 @@ static oe_result_t _init_sigstruct(
...
+        sigstruct->attributemask.flags |= SGX_FLAGS_KSS;
+    }
+    else
+    {
+        sigstruct->attributemask.flags &= ~SGX_FLAGS_KSS;
```
CodeMonkeyLeet, line 663: Do not unmask the `SGX_FLAGS_KSS` flag in the `attributemask` just because the enclave is not using KSS itself.

The `attributemask` determines whether two otherwise identical enclaves, one with and one without a particular attribute should be considered identical for attestation. Clearing this bit means that the dev wants them to be treated as the same (i.e. with or without the use of CONFIGID) Per [use case (3) in this comment](https://github.com/openenclave/openenclave/pull/3659#discussion_r508070776), this must be an explicit opt-in by the dev.

include/openenclave/internal/sgx/sgxproperties.h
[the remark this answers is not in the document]
CodeMonkeyLeet: @jianlia @yentsanglee 

I agree that in terms of the structure provide, any fixed 16-byte value of `family_id` and `extended_product_id` would be considered valid. The remaining concern still needs to be addressed:

> Specifically, this might need to cross-validate the attributes with the presence of the `family_id` or `ext_product_id` values, see comment in oesign.

Effectively, the check in oesign for `(attributes & `OE_SGX_FLAGS_KSS`) == 0` means that `family_id` and `extended_product_id`  must also be zero arrays needs to be enforced as inconsistent and rejected by the loader. That bit of logic is what needs to be reusable as a separate function.

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -101,7 +105,10 @@ typedef struct _oe_sgx_enclave_properties
 #define OE_SET_ENCLAVE_SGX(                                               \
     PRODUCT_ID,                                                           \
     SECURITY_VERSION,                                                     \
+    EXTENDED_PRODUCT_ID,                                                  \
```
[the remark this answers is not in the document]
bodzhang: Can [macro overload](https://stackoverflow.com/questions/3046889/optional-parameters-with-c-macros/28074198#28074198) be used to support optional `FAMILY_ID` and `EXTENDED_PRODUCT_ID` in `OE_SET_ENCLAVE_SGX' macro?

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -97,6 +97,18 @@ NumHeapPages=1024
...
+- **FamilyID**: The product family identity (ISVFAMILYID) a developer can specify to group
+different enclaves under a common identity. For example, an identifier for the application suite
+which includes several enclave apps.
+- **ExtendedProductID**: The extended product identity (ISVEXTPRODID) value that a developer
+can use as 128-bit globally unique identifier for the enclave where the 16-bit ProductID proves too restrictive.
+For example, in kss_valid.conf:
+```
+FamilyID=47183823-2574-4bfd-b411-99ed177d3e43
+ExtendedProductID=47183823-2574-4bfd-b411-99ed177d3e43
+```
```
bodzhang, line 111: The explanation of the updated `OE_SET_ENCLAVE_SGX` macro (or the suggested additional `OE_SET_ENCLAVE_SGX_KSS` macro) need to be updated in this doc.

host/sgx/create.c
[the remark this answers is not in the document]
bodzhang: Can the code also check for erroneous  props.config, ie. non-zero family_id/extende_product_id, but `OE_SGX_FLAG_KSS` bit is not set?

host/sgx/sgxsign.c
bodzhang: Clearing the `SGX_FLAGS_KSS` in `attributemask` means the dev wants the loader to decide whether to enable/disable KSS feature (EINIT instruction will pass either way, as long as other policy is not violated, for example, non-zero family_id/extended_product_id requiring KSS enabled). For family_id/extedned_product_id feature only, it makes sense to keep `SGX_FALGS_KSS` in `attributemask` as 1. SGX remote attestation will reflect KSS enable/disable state in SGX `REPORT`/`QUOTE` and it's up to the Relying Party to decide whether the security property of the same enclave in KSS enabled and KSS disabled mode is equivalent if the dev allows the loader to choose KSS enable/disable.

bodzhang (changes requested):

host/sgx/sgxsign.c
jianlia: Thanks. Lost change when I updated the PR. Updated.

tests/tools/oesign/test-enclave/host/host.c
jianlia: Related code move in block when KSS is enabled so setting will not applied if KSS not supported

tools/oesign/oesign.c
jianlia: Done

include/openenclave/bits/sgx/sgxproperties.h
jianlia: Good catch. Updated. Thanks

host/sgx/create.c
jianlia: Updated

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -88,7 +88,7 @@ In addition, the following two properties are defined by the developer and map d
   incremented whenever a security fix is made to the enclave code.
 
 Here is the example from helloworld.conf used in the helloworld sample:
-```
+
```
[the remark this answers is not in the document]
jianlia: Oops. Recovered.

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -97,6 +97,15 @@ NumHeapPages=1024
...
 ```
+Additionally, for platform that supports for Key sharing and separation (KSS),
+the following properties can be used:
+ **IsvFamilyID**: ISV assigned product family identity (ISVFAMILYID)
+ **IsvExtProductID**: ISV assigned extended product identity (ISVEXTPRODID)
+ ```
+ The example from kss_valid.conf
+ IsvFamilyID=47183823-2574-4bfd-b411-99ed177d3e43
+ IsvExtProductID=47183823-2574-4bfd-b411-99ed177d3e43
+ ```
```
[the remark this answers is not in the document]
jianlia: Updated

host/sgx/sgxload.c
```diff
@@ -349,6 +351,11 @@ oe_result_t oe_sgx_create_enclave(
     if (!(secs = _new_secs((uint64_t)base, enclave_size, context)))
         OE_RAISE(OE_OUT_OF_MEMORY);
 
+    if (context->attributes.flags & OE_SGX_FLAGS_KSS)
```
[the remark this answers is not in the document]
jianlia: Updated

tests/tools/oesign/test-sign/CMakeLists.txt
```diff
@@ -203,3 +203,65 @@ add_test(
...
+add_test(NAME tests/oesign-sign-familyid-toolong
+         COMMAND oesign sign -e $<TARGET_FILE:oesign_test_enc> -c
+                 ${OESIGN_TEST_INPUTS_DIR}/family_id_too_long.conf -k
+    ${OESIGN_TEST_INPUTS_DIR}/sign_key.private.pem)
+
+set_tests_properties(
+  tests/oesign-sign-familyid-toolong
+  PROPERTIES
+    PASS_REGULAR_EXPRESSION "bad value for 'IsvFamilyID'")
+
```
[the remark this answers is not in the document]
jianlia: Updated

host/sgx/sgxsign.c
jianlia: Will remove the unset/clearing. Thanks

tests/tools/oesign/test-enclave/host/host.c
jianlia: Removed _is_kss_supported from oesign to test code

jianlia: After I remove _is_kss_suuported from oesign, test will fail here so I will update the test here like the following flow
    if ((result = oe_create_oesign_test_enclave(
    {
        if (!(result == OE_PLATFORM_ERROR && is_kss_supported() &&  non-zero KSS properties))
            oe_put_err("oe_create_crypto_enclave(): result=%u", result);
        else
        {
            return 0;
        }
        
    }

tools/oesign/oesign.c
jianlia: Removed _is_kss_supported from Oesign

docs/GettingStartedDocs/buildandsign.md
jianlia: Updated in #3700

jianlia: Updated in #3700.

host/sgx/create.c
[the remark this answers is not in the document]
jianlia: Updated in #3700

include/openenclave/bits/sgx/sgxproperties.h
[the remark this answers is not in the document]
jianlia: Updated in #3700
Will look into Macro overload in future work.

include/openenclave/internal/sgx/sgxproperties.h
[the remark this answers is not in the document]
jianlia: Add  oe_sgx_is_unset_uuid function per comment.

include/openenclave/internal/sgxsign.h
jianlia: Thanks Replaced in #3700

tests/tools/oesign/test-enclave/host/host.c
jianlia: Updated in #3700

tools/oesign/oesign.c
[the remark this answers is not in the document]
jianlia: Updated in #3700

BRMcLaren:
@jianlia I noticed the "replaced by" in the title. Can this be safely closed?

jianlia:
> @jianlia I noticed the "replaced by" in the title. Can this be safely closed?

Yes.

PR #3464, opened 2020-09-01
EEID support for Windows

wintersteiger:
This adds EEID support for Windows.

There are a couple of bits that need discussion:

- [x] BCrypt doesn't give us access to the internal SHA256 hash state in a nice way; so this code contains magic offets. I'm getting in touch with the BCrypt team to see whether we can do something about that.
- [x] In the `clangw` script, I remove `-std:c11`, because MSVC claims that we would have to use `__asm` instead of `asm`. I'm not sure whether that's correct or not, but if so, we would have to change a few other macros. 
- [x] MSVC has a prototype for `memcpy` that does not contain the `restrict` keyword, so it reports a mismatch, see change in `corelibc/bits/defs.h`.

Also adds attestation plugin support to the host_verify sample.

Issue #70, opened 2018-01-06
Support SGX quote verification within an enclave


CodeMonkeyLeet:
This is a Coco requirement.

The existing Intel ECDSA-based quote creation and verification only happens on the host side as part of #44. Porting the extra crypto support and methods to be able to run the verification within an enclave is extra work that has a dependency on #68.

PR #3661, opened 2020-10-12
Fix custom claims problem in SGX verifier plugin

wintersteiger:
When there are no custom claims, `custom_claims_buffer` would point to after the report body and `custom_claims_buffer_size` would (correctly) be 0, but `oe_sgx_extract_claims` rejects that as invalid parameters ([here](https://github.com/openenclave/openenclave/blob/cf573c06adfbde75a3830a6dff119107b1ddee73/common/sgx/verifier.c#L461)).

### DIFF
--- a/common/sgx/verifier.c
+++ b/common/sgx/verifier.c
@@ -641,8 +641,11 @@ static oe_result_t _verify_evidence(
     {
         report_body = evidence_buffer;
         report_body_size = sizeof(sgx_report_t);
-        custom_claims_buffer = report_body + report_body_size;
-        custom_claims_buffer_size = evidence_buffer_size - report_body_size;
+        if (evidence_buffer_size > report_body_size)
+        {
+            custom_claims_buffer = report_body + report_body_size;
+            custom_claims_buffer_size = evidence_buffer_size - report_body_size;
+        }
 
         OE_CHECK(_verify_local_report(report_body, report_body_size));
     }
@@ -654,8 +657,12 @@ static oe_result_t _verify_evidence(
 
             report_body = evidence_buffer;
             report_body_size = sizeof(*quote) + quote->signature_len;
-            custom_claims_buffer = report_body + report_body_size;
-            custom_claims_buffer_size = evidence_buffer_size - report_body_size;
+            if (evidence_buffer_size > report_body_size)
+            {
+                custom_claims_buffer = report_body + report_body_size;
+                custom_claims_buffer_size =
+                    evidence_buffer_size - report_body_size;
+            }
         }
         else if (format_type == SGX_FORMAT_TYPE_LEGACY_REPORT)
         {
--- a/tests/attestation_plugin/host/host.c
+++ b/tests/attestation_plugin/host/host.c
@@ -111,11 +111,11 @@ int main(int argc, const char* argv[])
         argv[1], OE_ENCLAVE_TYPE_AUTO, flags, NULL, 0, &enclave);
     OE_TEST(result == OE_OK);
 
-    run_runtime_test(enclave);
-    register_sgx(enclave);
-    test_sgx(enclave);
-    unregister_sgx(enclave);
-    OE_TEST(oe_terminate_enclave(enclave) == OE_OK);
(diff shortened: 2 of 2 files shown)


wintersteiger:
@shnwc this is related to https://github.com/openenclave/openenclave/pull/3613. I have no idea how the tests passed without this fix; on the current master, the attestation_plugin tests fail, both on my machine and in some CI runs (e.g. those for https://github.com/openenclave/openenclave/pull/3464).

wintersteiger:
It looks like there's a problem with the tests. They are not being reported correctly when the enclave is aborted. This is what I get for the attestation plugin tests on a fresh master:

    ...
    70: running verify_sgx_evidence
    70: Test failed: /data/cwinter/openenclave-master/tests/attestation_plugin/plugin/tests.c(728): verify_sgx_evidence     OE_INVALID_PARAMETER!=OE_OK
    ...
    70: 2020-10-12T13:36:26+0000.604541Z [(H)ERROR] tid(0x7f56762bfb80) | :OE_ENCLAVE_ABORTING 
    ...
    1/2 Test #70: tests/attestation_plugin .........   Passed    0.58 sec

wintersteiger:
Fixed; the return values of the test functions were ignored in `tests/attestation_plugin`.

common/sgx/verifier.c
```diff
@@ -641,7 +641,9 @@ static oe_result_t _verify_evidence(
     {
         report_body = evidence_buffer;
         report_body_size = sizeof(sgx_report_t);
-        custom_claims_buffer = report_body + report_body_size;
+        custom_claims_buffer = report_body_size < evidence_buffer_size
+                                   ? report_body + report_body_size
+                                   : NULL;
         custom_claims_buffer_size = evidence_buffer_size - report_body_size;
```
yentsanglee, line 647: If evidence_buffer_size < report_body_size, custom_claims_buffer_size should be 0. Same below.

wintersteiger: That's correct, but the converse is not true: when `evidence_buffer_size - report_body_size` is 0, then `custom_claims_buffer` is not `NULL` (but `report_body`).

wintersteiger:
bors try

common/sgx/verifier.c
shnwc: How about checking custom data availability first then setting `custom_claims_buffer_size` and `custom_claims_buffer`? That feels more straightforward to me. Same below.

shnwc, line 647:
```suggestion
        if (evidence_buffer_size > report_body_size)
        {
            custom_claims_buffer = report_body + report_body_size;
            custom_claims_buffer_size = evidence_buffer_size - report_body_size;
        }
```

common/sgx/verifier.c
```diff
@@ -641,7 +641,9 @@ static oe_result_t _verify_evidence(
     {
         report_body = evidence_buffer;
         report_body_size = sizeof(sgx_report_t);
-        custom_claims_buffer = report_body + report_body_size;
+        custom_claims_buffer = report_body_size < evidence_buffer_size
```
shnwc, line 644:
```suggestion
```

common/sgx/verifier.c
```diff
@@ -641,7 +641,9 @@ static oe_result_t _verify_evidence(
...
+                                   ? report_body + report_body_size
```
shnwc, line 645:
```suggestion
```

common/sgx/verifier.c
```diff
@@ -641,7 +641,9 @@ static oe_result_t _verify_evidence(
...
+                                   : NULL;
```
shnwc, line 646:
```suggestion
```

common/sgx/verifier.c
wintersteiger: Sure, that's definitely easier to read!

wintersteiger:
bors try

shnwc (approved):

yentsanglee (approved):

yentsanglee:
Looks good. Please squash all commits so it can be merged.

wintersteiger:
Sure, all done!

PR #3675, opened 2020-10-15
Support V3 PCS DER encoded CRL

qiucwang:
SGX V3 PCS [spec](https://sbx.api.portal.trustedservices.intel.com/documentation#pcs-revocation-v3) indicates PckCrl can be retrieved in either PEM encoded or DER encoded format, depends on user provided parameter.
To comply with RFC5280 which suggest CRL should be DER encoded, this PR is to support DER encoded CRL in quote verification.


common/sgx/collateral.c
```diff
@@ -307,17 +309,41 @@ oe_result_t oe_validate_revocation_list(
     // will error out.
     for (uint32_t i = 0; i < OE_SGX_ENDORSEMENTS_CRL_COUNT; ++i)
     {
-        OE_CHECK_MSG(
-            oe_crl_read_pem(
-                &crls[i],
-                sgx_endorsements
+        // v1/v2 CRL is PEM encoded which starts with "-----BEGIN X509 CRL-----"
+        if (strncmp(
```
yentsanglee, line 313: I'd suggest using `memcmp()` and checking length of the buffer to ensure it's smaller than OE_PEM_BEGIN_CRL_LEN.

tests/tools/oecertdump/host/sgx_quote.cpp
```diff
@@ -443,26 +443,28 @@ oe_result_t generate_sgx_report(oe_enclave_t* enclave, bool verbose)
...
+            log("Endorsement: Revocation TCB Issuer Chain:\n%s\n\n",
+                tcb_issuer_chain.data);
             output_certificate_chain(
                 tcb_issuer_chain.data, tcb_issuer_chain.size, false);
 
             oe_sgx_endorsement_item crl_pck_cert =
                 endorsements.items[OE_SGX_ENDORSEMENT_FIELD_CRL_PCK_CERT];
-            log("Endorsement: CRL PCK Certificate:\n");
+            log("Endorsement: CRL PCK Certificate:\n%s\n\n", crl_pck_cert.data);
             decode_crl_pem(crl_pck_cert.data, crl_pck_cert.size);
```
yentsanglee, line 454: Is this going to break if the crl is der encoded?

qiucwang: No. If input buffer cannot be decoded as pem, it will print nothing. So we can tell whether input buffer is a pem or a der.

qiucwang:
bors try

common/sgx/collateral.c
yentsanglee: I'd suggest we also check endorsement length before memcmp(). I.e.

````
        if (sgx_endorsements
                        ->items[OE_SGX_ENDORSEMENT_FIELD_CRL_PCK_CERT + i]
                        .size >= OE_PEM_BEGIN_CRL_LEN &&
            memcmp(
                (const char*)sgx_endorsements
                    ->items[OE_SGX_ENDORSEMENT_FIELD_CRL_PCK_CERT + i]
                    .data,
                OE_PEM_BEGIN_CRL,
                OE_PEM_BEGIN_CRL_LEN) == 0)
````
It's impossible for a valid der-encoded crl to be that small, but we want to protect against malicious input.

File: common/sgx/collateral.c [hunk 3]
```diff
@@ -305,19 +306,37 @@ oe_result_t oe_validate_revocation_list(
...
-                    ->items[OE_SGX_ENDORSEMENT_FIELD_CRL_PCK_CERT + i]
-                    .data,
-                sgx_endorsements
-                    ->items[OE_SGX_ENDORSEMENT_FIELD_CRL_PCK_CERT + i]
-                    .size),
-            "Failed to read CRL. %s",
-            oe_result_str(result));
+        // v1/v2 CRL is PEM encoded which starts with "-----BEGIN X509 CRL-----"
+        if (sgx_endorsements->items[i].size >=
+                (OE_PEM_BEGIN_CRL_LEN + OE_PEM_END_CRL_LEN) &&
```
yentsanglee, line 315: Just ensure size is at least OE_PEM_BEGIN_CRL_LEN. This check is to safe-guard memcmp(), not a logic check.

yentsanglee (approved):

PR #3678, opened 2020-10-16
Display quote3_error_t enum values instead of error codes

ryanhsu19:
Display `quote3_error_t` enum values, instead of hexadecimal error codes, so that users can easily understand the errors defined in the Intel DCAP repository.

### DIFF
--- a/host/sgx/sgxquote.c
+++ b/host/sgx/sgxquote.c
@@ -32,6 +32,11 @@ OE_STATIC_ASSERT(sizeof(sgx_att_key_id_ext_t) == sizeof(sgx_att_key_id_t));
 #define SGX_QL_ALG_EPID_UNLINKABLE SGX_QL_ALG_EPID
 #define SGX_QL_ALG_EPID_LINKABLE SGX_QL_ALG_RESERVED_1
 
+// Facilitate writing switch statement for quote3_error_t values
+#define CASE_ERROR_RETURN_ERROR_STRING(x) \
+    case x:                               \
+        return #x;
+
 static oe_sgx_quote_ex_library_t _quote_ex_library = {0};
 static const oe_uuid_t _unknown_uuid = {OE_FORMAT_UUID_SGX_UNKNOWN};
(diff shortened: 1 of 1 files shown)


host/sgx/sgxquote.c
```diff
@@ -401,6 +406,99 @@ static bool _use_quote_ex_library(void)
...
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_PLATFORM_UNKNOWN);
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_QVEIDENTITY_MISMATCH);
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_QVE_OUT_OF_DATE);
+
+        // the rest of error codes will not be used, but we still have to
+        // enumerate them
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_ERROR_MIN);
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_ERROR_MAX);
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_SUCCESS);
+    }
```
yentsanglee, line 499: Here you should have a default case to catch errors defined in the future.

yentsanglee:
Next step is to turn all
`"quote3_error_t=0x%x\n", err);`
into
`"quote3_error_t=%s\n", get_quote3_error_t_string(err));`

host/sgx/sgxquote.c
ryanhsu19: Since the function would not be called when the error code == `SGX_QL_SUCCESS`, I replaced these lines with a default case.

yentsanglee: Please return the error number in the string.

yentsanglee (approved):

ryanhsu19:
bors try

shants (approved):
LGTM, approving the changes.

letmaik (approved):
Thanks!

ryanhsu19:
bors try

ryanhsu19:
bors try

PR #3684, opened 2020-10-20
add Proposal for Protected Code Loader support in OE

johnvenn:


docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+to user. 
+
+# Specification
+
+## Protected Code Loader Software Work Flow in Open Enclave
+### Build time encryption 
+    A seperate tool "oeencrypt" is provided at enclave build time to encrypt 
+	the enclave image right before enclave signing process.  
+### Load time decryption
+    On enclave initliazation operation(EINIT operation), right before relocation,
```
dthaler, line 39:
```suggestion
    On enclave initialization operation (EINIT operation), right before relocation, 
```

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+### Build time encryption 
+    A seperate tool "oeencrypt" is provided at enclave build time to encrypt 
+	the enclave image right before enclave signing process.  
+### Load time decryption
+    On enclave initliazation operation(EINIT operation), right before relocation, 
+    the encrypted enclave image need to be decrypted and than perform the relocation
+    operation.
+
+## Encryption Algorithm in Protected Code Loader
+Using openssl AES-256-GCM as the encryption/decryption algorithm.
```
dthaler, line 44: Can we support crypto-agility?

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+info on enclave loading and must remain plaintext.
+
+Definition of PCL entry(section ".pcltbl"):
+typedef struct pcl_table_t_
+{
+	/* Current state of PCL: initailized to PCL_PLAIN */
+    pcl_status_e pcl_state;                   
+    uint32_t     reserved1[3];                /* Must be 0 */
+	// GUID must match GUID in Sealed blob
+    uint8_t      pcl_guid[SGX_PCL_GUID_SIZE];
```
dthaler, line 59: This section is SGX-only but the notion of loading protected code is desired in other TEEs as well.   Can we cleanly separate TEE-agnostic functionality from TEE-specific functionality through some clean layer/interface?

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+Sections not mentioned above are sections containing IP information that need to be protected.
+Mainly those sections are:
+1. Code sections -- .text in elf file
+2. Data sections -- .data in elf file (initialized local/global variables)
+3. Read-only Data sections -- .rodata in elf file (const variables)
+4. sections containing relocation info related with the above items
+
+## Sealing/Unsealing the decryption key
+To deliver the decryption key in a secure way, the key used to encrypt the enclave should get sealed
+by key policy OE_SEAL_POLICY_PRODUCT on deliverting to Protected Code Loader for decryption in a
```
dthaler, line 100:
```suggestion
by key policy OE_SEAL_POLICY_PRODUCT on delivering to Protected Code Loader for decryption in a 
```

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+Mainly those sections are:
+1. Code sections -- .text in elf file
+2. Data sections -- .data in elf file (initialized local/global variables)
+3. Read-only Data sections -- .rodata in elf file (const variables)
+4. sections containing relocation info related with the above items
+
+## Sealing/Unsealing the decryption key
+To deliver the decryption key in a secure way, the key used to encrypt the enclave should get sealed
+by key policy OE_SEAL_POLICY_PRODUCT on deliverting to Protected Code Loader for decryption in a 
+sealed_blob. The protected code loader will unseal this sealed_blob to get the decryption key.
```
dthaler, line 101: FYI, in some other TEEs, the loader can not get the decryption key, it can _use_ the key to decrypt but can't _get_ the key.  E.g., the key is in a TPM.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+		uint64_t arg1;
+		uint64_t sealed_blob;
+	} oe_sgx_ecall_ms_t;
+4. new lib for section ".pcltbl" and decryption -- liboepcl.a: placed in OESDK installation lib folder
+
+### PCL Sample Code
+A Sample Code project will be provided in samples for how to use Protected Code Loader.
+
+## Debugging consideration when user launches an Encrypted Enclave
+Debugging with oegdb should work reguarly with a minor disclaimer: you can insert break points in
```
dthaler, line 146:
```suggestion
Debugging with oegdb should work regularly with a minor disclaimer: you can insert break points in
```

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+typedef struct _oe_enclave_setting
+{
+    oe_enclave_setting_type_t setting_type;
+    /**
+     * The specific setting for the enclave, such as for configuring
+     * context-switchless calls.
+     */
+    union {
+		/* for protected code loader */
+		uint8_t* sealed_blob;
```
dthaler, line 129: Where does the app get this blob from?   Add answer in the doc somewhere.

dthaler: Assuming the sealed_blob isn't the actual code to be decrypted, this seems to make an assumption that the enclave has 1 and only 1 possibility of what to decrypt.  I'd prefer a solution that does not force that assumption.
The encrypted code may be acquired at runtime (e.g., across a network), or there may be multiple possible ones compiled into the enclave, or whatever else.    Things get easier if the granularity is at the module level and not specific pages or segments, and it becomes more generally applicable across TEEs.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
bodzhang: It might be helpful to review the PCL design in Intel SGX SDK. https://download.01.org/intel-sgx/sgx-linux/2.11/docs/Intel_SGX_Developer_Reference_Linux_2.11_Open_Source.pdf , page 78-86 talks about PCL support. Intel SGX SDK's PCL sample code for the "utility enclave" seals a hardcoded "mock" key and call out "don't use in release code". https://github.com/intel/linux-sgx/blob/74b24b69f90dee2638ab646d9090cb286ec8abc9/SampleCode/SampleEnclavePCL/Seal/Seal.cpp#L37

dthaler: It would be good to add the references directly into the doc.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
johnvenn: ok, I will consider it.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
johnvenn: I think AES GCM algorithm is the most popular encrypt/decrypt algorithms used today. And for users wants to use this feature, they don't need to do more work on this encryption/decryption part. Also as mentioned in Cedric's PR 3652, the GCM is faster than other block cipher algorithms.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
johnvenn: pcl_guid is actually the aad for decryption key seal/unsealing functions. This is not SGX only, the MACRO SGX_PCL_GUID_SIZE is confusing here, I will change it in document. Also I will add more details on sealing/unsealing of the decryption key in this document.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
johnvenn: For use who's using PCL feature, they will not be aware of the encryption/decryption function of their enclave image - ELF file, which specific sections need to be encrypted/decrypted are done by pcl in oe sdk. The solution proposed here, we provided a simple way for user to use PCL, they don't need to do anything regarding the encryption/decryption algorithm, what they need to do is only to follow the PCL sample code to provide a raw enclave image, linking it to oepcl lib. Encryption is by oeencrypt, and decrypt is by oe sdk runtime.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
dthaler: I'm not an expert so just asking, but https://tools.ietf.org/html/rfc7696#section-3.1 says:

>    Sometimes more than one mandatory-to-implement algorithm is needed to
>    increase the likelihood of interoperability among a diverse
>    population.  For example, authenticated encryption is provided by
>    AES-CCM [RFC3610] and AES-GCM [GCM].  Both of these algorithms are
>    considered to be secure.  AES-CCM is available in hardware used by
>    many small devices, and AES-GCM is parallelizable and well suited to
>    high-speed devices.  Therefore, an application needing authenticated
>    encryption might specify one of these algorithms or both of these
>    algorithms, depending on the population.

Hence my question, since OE SDK is not just for high-speed devices like cloud servers.
If a crypto expert tells me that AES-256-GCM is sufficient for IoT, and will suffice for many years to come, then it's fine with me, but I have to ask.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+## Encryption Algorithm in Protected Code Loader
+Using openssl AES-256-GCM as the encryption/decryption algorithm.
+
+## ELF sections Left plaintext
+### PCL table entry 
+A PCL table entry (a section called ".pcltbl" in ELF file)is built into the Enclave 
+image by linking to PCL lib, i.e. liboepcl.a, this part contains enclave decryption
+info on enclave loading and must remain plaintext.
+
+Definition of PCL entry(section ".pcltbl"):
```
mingweishih, line 52: Add ``` before the line 53 and after line 71 so that the code syntax can be correctly rendered.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+8. sections holds the content provided by entries with index DT_SYMTAB, DT_STRTAB and DT_REL in
+   dyn_info (e.g. .dynsym, .dynstr, .rela.dyn)
+9. sections containing PCL code and data:
+   a. section ".pcltbl"  // Designated section for PCL table
+   b. .nipx, .nipd, .niprod, .nipd_rel, .nipd_rel_ro_local
+10. sections for debugging
+   .comment, .debug_abbrev, .debug_aranges, .debug_info, .debug_line, .debug_lc, .debug_ranges,
+   .debug_str
+
+## Elf sections to be encrypted
```
mingweishih, line 90: For consistency, capitalize the first character of words that are not articles/prepositions in title/sub-title.

```suggestion
## Elf Sections to Be Encrypted
```

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+Sections not mentioned above are sections containing IP information that need to be protected.
+Mainly those sections are:
+1. Code sections -- .text in elf file
+2. Data sections -- .data in elf file (initialized local/global variables)
+3. Read-only Data sections -- .rodata in elf file (const variables)
+4. sections containing relocation info related with the above items
+
+## Sealing/Unsealing the decryption key
```
mingweishih, line 98:
```suggestion
## Sealing/Unsealing the Decryption Key
```

docs/DesignDocs/Protected_Code_Loader_Proposal.md
CodeMonkeyLeet: Just as an aside: it is likely that AES-256-GCM will remain in use for many years; the CAESAR competition I think only announced the results early last year. If you compare that to when SHA-3 was decided (back in 2012) and the ramp for general adoption of it, you might have a sense for how long it might take the new authenticated encryption standard to become pervasive.

An interesting thing to note is that CAESAR breaks the choice of algorithms down between resource-constrained and high-performance app categories, so in the future, it may be that IoT may choose a different algorithm than something running on server.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+To deliver the decryption key in a secure way, the key used to encrypt the enclave should get sealed
+by key policy OE_SEAL_POLICY_PRODUCT on deliverting to Protected Code Loader for decryption in a 
+sealed_blob. The protected code loader will unseal this sealed_blob to get the decryption key. 
+
+## new APIs and libs
```
mingweishih, line 103:
```suggestion
## New APIs and Libraries
```

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+# User Experience
+## Enclave Build-time:
+### Linking pcl lib to the enclave 
+On building the enclave, a static lib "liboepcl.a" should be linked to the 
+enclave image with the compilation option "-Wl, whole-archive"
+### Encrypt the enclave image
+An encryption tool is provided as the pre-step to enclave signing process:
+
+```bash
+oeencrypt -i my_enclave -o my_enclave.enc -k keyfile
```
CodeMonkeyLeet, line 24: @HernanGatta Can you comment on whether it would be feasible/desirable for OE SDK to provide common tooling for TA encryption as well on OP-TEE?

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+4. sections containing relocation info related with the above items
+
+## Sealing/Unsealing the decryption key
+To deliver the decryption key in a secure way, the key used to encrypt the enclave should get sealed
+by key policy OE_SEAL_POLICY_PRODUCT on deliverting to Protected Code Loader for decryption in a 
+sealed_blob. The protected code loader will unseal this sealed_blob to get the decryption key. 
+
+## new APIs and libs
+### Encryption tool
+new tool for ELF image encryption - oeencrypt: placed in OESDK installation folder as the bin files
```
mingweishih, line 105:
```suggestion
- New tool for ELF image encryption --- oeencrypt: placed in OESDK installation folder as the bin files.
```

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+new lib for section ".pcltbl" and decryption -- liboepcl.a: placed in OESDK installation folderd
```
mingweishih, line 106:
```suggestion
- New library for section ".pcltbl" and decryption --- liboepcl.a: placed in OESDK installation folder
```

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+	the enclave image right before enclave signing process.  
+### Load time decryption
+    On enclave initliazation operation(EINIT operation), right before relocation, 
+    the encrypted enclave image need to be decrypted and than perform the relocation
+    operation.
+
+## Encryption Algorithm in Protected Code Loader
+Using openssl AES-256-GCM as the encryption/decryption algorithm.
+
+## ELF sections Left plaintext
```
CodeMonkeyLeet, line 46: Putting a comment I had from the previous SIG-Arch meeting in writing: Have we considered if this design can be simplified if the OE SDK loader simply supported dynamic binding on load? (i.e. an enclave app can consist of multiple shared objects and not only a statically linked binary)

It seems like that could simplify the separation of encrypted/unencrypted boundaries by simply marking individual binaries as encrypted or not, instead of having to redeclare custom sections of the ELF binary.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+by key policy OE_SEAL_POLICY_PRODUCT on deliverting to Protected Code Loader for decryption in a 
+sealed_blob. The protected code loader will unseal this sealed_blob to get the decryption key. 
+
+## new APIs and libs
+### Encryption tool
+new tool for ELF image encryption - oeencrypt: placed in OESDK installation folder as the bin files
+new lib for section ".pcltbl" and decryption -- liboepcl.a: placed in OESDK installation folderd
+as the lib files
+
+### Modifications to OE SDK runtime lib
```
mingweishih, line 109:
```suggestion
### Modifications to OE SDK Runtime Library
```

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+	{
+		uint64_t arg1;
+		uint64_t sealed_blob;
+	} oe_sgx_ecall_ms_t;
+4. new lib for section ".pcltbl" and decryption -- liboepcl.a: placed in OESDK installation lib folder
+
+### PCL Sample Code
+A Sample Code project will be provided in samples for how to use Protected Code Loader.
+
+## Debugging consideration when user launches an Encrypted Enclave
```
mingweishih, line 145:
```suggestion
## Debugging Consideration When User Launches an Encrypted Enclave
```

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
+Open Enclave Protected Code Loader
+====
+
+Enabling Enclave Code Confidentiality in Open Enclave.
+
+# Motivation
```
CodeMonkeyLeet, line 6: Is there a specific object to provide some kind of developer experience parity with the Intel SDK PCL feature? It would be useful to call them out as we consider the design. 

For example, is there a desire to preserve the mechanism for delivering PCL keys to the target device via key sealing (i.e. A dev can reuse a previously developed key sealing enclave and it would still work with the updated OE SDK based PCL enclave. That would also create a specific requirement on the Sealing API feature also under discussion.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
mingweishih: Can you make the work flow more clear in this document? How do user to generate a sealed blob is not clear in this document.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+as the lib files
+
+### Modifications to OE SDK runtime lib
+No new APIs exposed to user.
+1. A new setting is defined in include/openenclave/host.h as the argument to support
+protected code loader enclave loading in API oe_create_enclave.
+typedef enum _oe_enclave_setting_type
+{
+	// PCL setting type
+	OE_ENCLAVE_SETTING_PCL = 0xac120002,
```
CodeMonkeyLeet, line 116: I think my primary concern with this becoming part of the OE SDK public API surface is that it enshrines the existing PCL implementation as the standard model for key handling for PCL, which I don't think generalizes well. 

For example, the idea that the dev is responsible for passing to `oe_create_enclave` the sealed blob assumes that a different TEE can and should do the same, but it's not clear that handling the sealed key itself is ideal or necessary. Consider a strawman example instead, where instead of passing the blob, you pass a path/configuration to get the key in whatever form makes sense for the TEE; that provides more flexibility to generalize. For example, it could just be a file path to the sealed key blob for SGX, but it might be a device path to a trusted key store, or a network endpoint to fetch a key from.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+    On enclave initialization operation (EINIT operation), right before relocation, 
+    the encrypted enclave image need to be decrypted and than perform the relocation
+    operation.
+
+## Encryption Algorithm in Protected Code Loader
+Using openssl AES-256-GCM as the encryption/decryption algorithm.
+
+## ELF sections Left plaintext
+### PCL table entry 
+A PCL table entry (a section called ".pcltbl" in ELF file)is built into the Enclave
```
anakrish, line 48: If an enclave consists of multiple elf binaries, would each binary contain the PCL sections?

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+typedef struct _oe_enclave_setting
+{
+    oe_enclave_setting_type_t setting_type;
+    /**
+     * The specific setting for the enclave, such as for configuring
+     * context-switchless calls.
+     */
+    union {
+		/* for protected code loader */
+		uint8_t* sealed_blob;
```
anakrish, line 129: Doesn't the sealed blob require an accompanying blob size?

docs/DesignDocs/Protected_Code_Loader_Proposal.md
anakrish: I second @mingweishih. The purpose of sealed blob is not clear.

anakrish (commented):
I have a few questions:

1. One of the first things that an enclave does upon entry is to call oe_is_outside_enclave/oe_is_within_enclave to validate the inputs. Both these functions rely on knowing enclave boundaries which today can be inferred from `enclave_rva`, `oe_enclave_properties_sgx` and other variables.
  https://github.com/openenclave/openenclave/blob/8ea5970bc93d039eed5a5ecdd2c7bc1e0759690f/enclave/core/sgx/globals.c#L128-L130
  How does the design ensure that these variables themselves are not encrypted?

2. The purpose of sealed-blob is not clear. Can you clarify its usage?

3. What happens if the host intentionally passes incorrect decryption keys?

4. Can the design be extended to handle multi-module enclaves?

docs/DesignDocs/Protected_Code_Loader_Proposal.md
johnvenn: thanks, I will update a seperate title for the sealed_blob in PCL entry. The purpose of the sealed_blob is based on the thought of delivering the decryption key in a secure way.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+secret data embedded in the enclave image. 
+
+The Open Enclave Protected Code Loader shall enable providing confidentiality 
+and integrity to the IP sections in enclave image, i.e. , the ELF file.
+
+# User Experience
+## Enclave Build-time:
+### Linking pcl lib to the enclave 
+On building the enclave, a static lib "liboepcl.a" should be linked to the 
+enclave image with the compilation option "-Wl, whole-archive"
```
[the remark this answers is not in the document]
johnvenn: the "whole-archive" option is for section ".pcltbl", forcing this section to be included in the enclave executable.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,161 @@
...
+### Encrypt the enclave image
+An encryption tool is provided as a pre-step to the enclave signing process:
+
+```bash
+oeencrypt -i my_enclave -o my_enclave.enc -k keyfile
+```
+
+## Enclave Load-time:
+On Enclave Loading, if the image is encrypted, it needs to be decrypted
```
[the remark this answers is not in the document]
johnvenn: yes, decryption the enclave happens during enclave initialization after enclave loaded in the trusted environment. I will correct the statement here to make it clear.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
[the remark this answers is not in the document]
anakrish: The proposed alternatives are reasonable for now. I thought about this a bit more; we might be able to provide a seamless workflow by notifying the debugger about encrypted text sections only after they have been decrypted. We can attempt that once the feature is in.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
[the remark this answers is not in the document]
anakrish: Thanks. Please add the above explanation into the design doc.

docs/DesignDocs/Protected_Code_Loader_Proposal.md
johnvenn: updated

docs/DesignDocs/Protected_Code_Loader_Proposal.md
johnvenn: while considering the question, I checked dynamic linking branch and the test code, I didn't see any signing performed for the enclave or the dynamic lib linking to the enclave. How did you consider measurement on those multiple modules ? How did you manage heap/stack between those multiple modules? How is oesign used in dynamic linking?

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,189 @@
...
+Current Open Enclave SDK provides code integrity and data confidentiality at
+run-time but not enclave binary confidentiality on a disk. The enclave binary
+is in plaintext and can be reverse engineered to reveal code logic and
+secret data embedded in the enclave image. 
+
+The Open Enclave Protected Code Loader enables providing confidentiality 
+and integrity to the IP sections in the enclave image based on ELF format.
+
+The solution provided here is for enclaves built by statically linked libraries
+in OpenEnclave SDK. This is a parity of Protected Code Loader in Intel's SGX SDK.
```
dthaler, line 16:
```suggestion
in the Open Enclave SDK. This is a parity of Protected Code Loader in Intel's SGX SDK.
```

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,189 @@
...
+    the encrypted enclave image needs to be decrypted and then perform the relocation
+    operation.
+
+
+## Encryption Algorithm in Protected Code Loader
+Using openssl AES-256-GCM as the encryption/decryption algorithm.
+
+## ELF Sections Left Plaintext
+### PCL table entry 
+A PCL table entry (a section called ".pcltbl" in ELF file)is built into the Enclave
```
dthaler, line 53:
```suggestion
A PCL table entry (a section called ".pcltbl" in ELF file) is built into the Enclave 
```

docs/DesignDocs/Protected_Code_Loader_Proposal.md
```diff
@@ -0,0 +1,189 @@
...
+Solution: ISV should be able to choose when host attaches the debugger:
+1. Default: debugger shall be attached after PCL flow is done.
+2. For PCL and early trusted runtime development: debugger shall be attached before the first
+instruction inside an enclave
+
+# Alternative Architecture
+Current PCL architecture only supports section based encryption to ELFs with static link libraries.
+
+But if user wants to use dynamic linking for enclaves, there are one alternative:
+1. Encryption entire DSO(for algorithms, e.g.) linked to the enclave
```
dthaler, line 180:
```suggestion
1. Encryption entire DSO (for algorithms, e.g.) linked to the enclave
```

docs/DesignDocs/Protected_Code_Loader_Proposal.md
CodeMonkeyLeet: @johnvenn Dynamic linking still requires that all shared object files are loaded into the enclave memory space deterministically and measured before initialization, so it doesn't change the enclave measurement model. i.e. instead of loading a single binary, it simply extends that to dynamically loading and binding symbols from dependency binaries as part of the enclave app that is measured prior to initialization. It does not permit runtime loading of new shared objects into the enclave memory space, which would require an extension of the measurement model (such as the use of the experimental EEID feature). oesign behaves as it always does in this regard: it simply takes the enclave measurement, whether consisting of one module or multiple modules, and signs the entire enclave measurement.

The heap/stack model similarly doesn't change at all; those are process-level concepts, not specific to modules loaded for the app, enclave or otherwise. There is some additional C library initialization that needs to be extended when loading multiple modules (e.g. allocation of space for static thread-local variables from all loaded modules per TCS) which you will see extra code to handle in the feature/dynamic_binding branch.

radhikaj:
Closing for now. Please reopen when you work on it again.

PR #3697, opened 2020-10-30
Proposal for pthread library support in OE

xiangquanliu:


docs/DesignDocs/libc_pthread_proposal.md
```diff
@@ -0,0 +1,80 @@
+# To support pthread in Open Enclave SDK
+
+Currently, pthread management APIs are not implemented in OE SDK while they are
```
dthaler, line 3:
```suggestion
Currently, POSIX [pthread](https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_09) management APIs are not implemented in OE SDK while they are
```

docs/DesignDocs/libc_pthread_proposal.md
```diff
@@ -0,0 +1,80 @@
...
+implemented in SGX SDK. The pthread APIs are added when porting MKL-DNN to SGX SDK.
```
dthaler, line 4:
```suggestion
implemented in SGX SDK. The pthread APIs are added when porting the Intel(R) Math Kernel Library for Deep Neural Networks [MKL-DNN](https://oneapi-src.github.io/oneDNN/understanding_memory_formats.html) to SGX SDK.
```

docs/DesignDocs/libc_pthread_proposal.md
```diff
@@ -0,0 +1,80 @@
...
+To get better performance, MKL-DNN should be configured to build with OpenMP which
```
dthaler, line 5:
```suggestion
To get better performance, MKL-DNN should be configured to build with [OpenMP](https://www.openmp.org/) which
```

docs/DesignDocs/libc_pthread_proposal.md
```diff
@@ -0,0 +1,80 @@
...
+depends on pthread.
+
+## New APIs
```
dthaler, line 8:
```suggestion
## New Internal APIs
```
Important point is that these are not public APIs (the pthread* APIs are the public API) and can change at any time.

docs/DesignDocs/libc_pthread_proposal.md
```diff
@@ -0,0 +1,80 @@
...
+oe_thread_create is an internal function which will be wrapped to an external API pthread_create.
+
+The work sequence should be like,
+* Parent thread(Enclave): call pthread_create() inside the enclave codes written by users.
+* Parent thread(Enclave): initialize some structures and execute OCALL - OE_OCALL_THREAD_CREATE.
+* Parent thread(Host): allocate some resources and call oe_thread_create() to create a new thread.
+* Child thread(Host): execute ECALL - OE_ECALL_THREAD_CREATE_ROUTINE if new thread is created successfully.
+* Child thread(Enclave): get the user defined hook function and execute it.
+
+**NOTE**: Currently the parameter of attributes for the created thread is unused. So the created threads are
```
dthaler, line 22: The parameter should not be "unused".  You should explicitly check that it contains the behavior you support (e.g., == NULL) and fail with EINVAL if not.  Otherwise future additions can have app compact problems.

docs/DesignDocs/libc_pthread_proposal.md
```diff
@@ -0,0 +1,80 @@
...
+This function is used to wait for the specified thread to terminate.
+
+oe_thread_join is an internal function which will be wrapped to an external API pthread_join.
+
+The work sequence should be like,
+* Parent thread(Enclave): call pthread_create() to create a new thread and then return.
+* Other threads(Enclave/Host): call pthread_join() and then make an OCALL - OE_OCALL_THREAD_WAIT to wait for an event.
+* Child thread(Host): execute ECALL - OE_ECALL_THREAD_CREATE_ROUTINE.
+* Child thread(Host): wake up the waiting thread after OE_ECALL_THREAD_CREATE_ROUTINE return.
+* Other threads(Host): OE_OCALL_THREAD_WAIT return after being waked up.
```
dthaler, line 48:
```suggestion
* Other threads(Host): OE_OCALL_THREAD_WAIT return after being woken up.
```

docs/DesignDocs/libc_pthread_proposal.md
```diff
@@ -0,0 +1,80 @@
...
+    void* retval;  // Used by oe_thread_exit() and oe_thread_join() to pass the exit message
+    void* tid;
+} oe_thread_impl_t
+```
+
+## Support OSes
+To get better performance, the OS thread created in host should be detached. Currently the OS thread can be created 
+by oe_thread_create which is a wrapper of pthread_create in Linux and CreateThread in Windows. The attribute parameter 
+is not used, so the detached thread can't be created with default attribute. However, if only Linux is supported, there will be 
+no issue.
```
dthaler, line 78: It should work on Windows too.  https://sourceforge.net/projects/pthreads4w/ for example implements it on windows so it's certainly possible.  https://stackoverflow.com/questions/12744324/how-to-detach-a-thread-on-windows-c discusses how to create a detached thread on Windows.

CodeMonkeyLeet (commented):
Thanks for submitting the design proposal, XiangQuan, could you also address the questions as requested in the [DesignDocs/README.md](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/README.md)? In particular, it would be useful to understand the motivation and alternative implementations for comparison, as they help to guide the design discussion for the feature (e.g. how to scope the discussion. For example, whether we should also define support for `pthread_tryjoin_np`?)

In particular, I also have a range of additional questions that are not touched on in this design:

- What is the user experience in terms of how this feature is used?
    - The unstated assumption in this design seems to be that this is intended for SGX1. If so, then there's an expectation that a dev knows how to pre-allocate sufficient threads in the enclave configuration to service both normal ECalls, as well as any additional threads that their enclave app may end up creating.
    - It would be helpful to walk through the developer scenarios that we want to support.
    - Are the particular kinds of controls that would be useful to expose to the developer to improve the experience? e.g. would it be helpful to declare threads for thread creation and base threads available for host to call into enclaves for? 

- Is this feature specific to SGX1?
    - On SGX2 with EDMM support, we could which then dynamically allocate new thread data structures (notably the TCS) which could change the developer experience. If the assumption is that the user experience is unchanged moving from SGX1 to EDMM, that's also worth asserting in the design. 
   - OE SDK is in the process of breaking apart syscalls support into discrete modular libraries so that devs can explicitly include only what they use. For TEEs that do not currently support multiple threads (e.g. OP-TEE supported in OE), it would be useful to talk about the componentization of this feature and how the feature is targeted (e.g. not supported on OP-TEE, all features are going to be part of sgx/thread.edl that require dev to include) 

- How does thread creation interact with `oe_abort`?
   - `oe_abort` doesn't take down the enclave, only putting it in an `OE_ENCLAVE_ABORTING` state.
   - Entry into the enclave is blocked, but enclave threads continue to execute.
   - One implication is that a newly created host thread must fail gracefully and clean itself up if it encounters a failed enclave entry.
   - Another consideration is whether `pthread_cancel` semantics should be added and if `oe_abort` should cancel threads created in the enclave to force them to exit. 

Putting the call sequences together into a single view:

| Step | Host thread 1 | Host thread 2 | Enclave thread 1 | Enclave thread 2 |
|---:| --- | --- | --- | --- |
| 01 | *h1*: call `ecall_thread_example()` | | EENTER (ECall from *h1*) | |
| 02 | ... | | *e1*: handle `ecall_thread_example()` | |
| 03 | ... | | *e1*: call `oe_thread_create()` with `my_function()` | |
| 04 | ... | | *e1*: reserve TCS and set up `oe_thread_t` for *e2* | *e2*: reserved |
| 05 | EEXIT (OCall from *e1*) | | *e1*: `oe_thread_create()` calls `ocall_create_thread()` | ... |
| 06 | *h1*: handle `ocall_create_thread()` | | ... | ... |
| 07 | *h1*: call `pthread_create()` with `oe_run_thread()` to start *h2* | *h2*: started | ... | ... |
| 08 | *h1*: return | *h2*: call `oe_run_thread()` | EENTER (ORet from *h1*) | ... |
| 09 | ... | *h2*: call `ecall_run_thread()` | *e1*: return from `oe_thread_create()` | EENTER (ECall from *h2*) |
| 10 | EEXIT (OCall from *e1*) | ... | *e1*: call `oe_thread_join()` on *e2* | *e2*: handle `ecall_run_thread()` |
| 11 | *h1*: handle `ocall_join_thread()` | ... | *e1*: `oe_thread_join()` calls `ocall_join_thread()` | *e2*: call `my_function()` |
| 12 | *h1*: `ocall_join_thread()` calls `pthread_join()` on *h2* | ... | ... | *e2*: `my_function()` executes |
| 13 | *h1*: WAIT | ... | ... | *e2*: `my_function()` calls `oe_thread_exit()` |
| 14 | ... | ... | ... | *e2*: `oe_thread_exit()` marks `oe_thread_t` as exited |
| 15 | ... | *h2*: EEXIT (ERet from *e2*) | ... | *e2*: return |
| 16 | ... | *h2*: resume `oe_run_thread()` | ... | |
| 17 | *h1*: WAKE | *h2*: call `pthread_exit()` | ... | |
| 18 | *h1*: return | | EENTER (ORet from *h1*) | |
| 19 | ... | | *e1*: return from `oe_thread_join()` | |
| 20 | ... | | *e1*: resume rest of `ecall_thread_example()` | |
| 21 | EEXIT (ERet from *e1*) | | *e1*: return | |
| 22 | *h1*: resume remaining execution | | | |

There are a couple of design choices I would like clarity on:

- When is an SGX1 TCS reserved for the new thread and when is it released for reuse? 
   - For example, I assume that we need to reserve a TCS in the enclave during `oe_thread_create` because I would like the failure to be synchronous in the enclave instead of asynchronously (e.g. error in line 9 after `oe_thread_create` has already returned). That creates a different set of problems to solve though (e.g. keeping track of enclave threads both inside the enclave and in host, and keeping them in sync)
- It's also important to ensure that a created thread state in the enclave is cleaned out and not reusable by a malicious host. i.e. if `my_function()` exits, the enclave should clean all thread local state and the host cannot:
    - reuse state on that thread for a regular ECall invocation
    - reuse state on that thread and repeat another `ecall_run_thread()` invocation. 
 - An example of such a consideration is that it's probably not safe to defer thread state cleanup of the joined thread to `oe_thread_join()` when it is resumed after wake since we may want to support detached threads in the future, but that may introduce constraints on where data structures can be allocated (e.g. the `oe_thread_t` that `oe_thread_join()` needs to check to ensure that it is being correctly awakened probably should not be in the thread context being destroyed)

xiangquanliu:
Thanks! I will update the docs according to the comments.

docs/DesignDocs/libc_pthread_proposal.md
xiangquanliu: Seems the attr parameter of locks initialize function is "unused".  Should these pthread functions be aligned?

docs/DesignDocs/libc_pthread_proposal.md
xiangquanliu: OK. I will do a poc for this.

docs/DesignDocs/libc_pthread_proposal.md
```diff
@@ -0,0 +1,80 @@
...
+* Child thread(Host): wake up the waiting thread after OE_ECALL_THREAD_CREATE_ROUTINE return.
+* Other threads(Host): OE_OCALL_THREAD_WAIT return after being waked up.
+* Other threads(Enclave): in pthread_join() check the child thread status which is controlled in enclave before return.
+
+### oe_thread_self
+The function is used to get the calling thread’s ID. 
+
+oe_thread_self is an internal function which will be wrapped to an external API pthread_self.
+
+Currently it returns the td pointer. It will be changed to return an internal structure pointer which
```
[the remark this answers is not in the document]
xiangquanliu: Thread Local Storage. I will update the docs.

docs/DesignDocs/libc_pthread_proposal.md
```diff
@@ -0,0 +1,80 @@
...
+    void* (*start_routine)(void *);
+    void* arg;
+    void* retval;  // Used by oe_thread_exit() and oe_thread_join() to pass the exit message
+    void* tid;
+} oe_thread_impl_t
+```
+
+## Support OSes
+To get better performance, the OS thread created in host should be detached. Currently the OS thread can be created 
+by oe_thread_create which is a wrapper of pthread_create in Linux and CreateThread in Windows. The attribute parameter
```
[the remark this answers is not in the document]
xiangquanliu: I will try it and update the POC.

docs/DesignDocs/libc_pthread_proposal.md
xiangquanliu: Just checked https://sourceforge.net/projects/pthreads4w/, it is kind of complicated to implement the pthread_create. I think it is not suitable to include it in this patch.

PR #3699, opened 2020-10-30
Proposal for supporting deep copy of out parameters

mingweishih:
This PR adds the proposal of supporting deep copy of variable-length out parameters. The goal of the document is seeking the consensus on which design/direction that OE should take.

### DIFF
--- a/docs/DesignDocs/DeepCopyOutParameters.md
+++ b/docs/DesignDocs/DeepCopyOutParameters.md
@@ -1,0 +1,721 @@
+Support Deep Copy of Variable-Length Out Parameters
+=====
+
+This document provides a brief overview of the oeedger8r's deep copy feature and points out
+its missing piece: the support of the deep-copy `out` parameter. Next, the document describes a
+design that supports the deep-copy `out` parameter.
+
+Introduction
+------------
+
+The oeedger8r tool supports the feature of deep copy, which allows a user to pass a struct with nested pointers
+in ECALLs and OCALLs. To leverage this feature, the user should define the struct as part of an EDL file
+and use proper annotations on both the struct and the ECALL/OCALL, which intends to pass the struct.
+See also [Enforcing full serialization in EDL](/docs/DesignDocs/full_edl_serialization.md).
+
+An example of struct definitions is as follows.
+
+```
+struct Blob
+{
+  size_t len;
+  [size=len] char* buf;
+};
+
+struct NestedBlob
+{
+  size_t num;
+  [count=num] Blob* blob_array;
+};
+```
+
+The `num` variable specified to the `count` attribute represents the length (i.e., the number of elements)
+of the `blob_array`. Similarly, the `len` variable specified to the `size` attribute represents the size (in bytes) of the `buf`.
+Refer to the [Grammar documentation](https://github.com/openenclave/oeedger8r-cpp/blob/master/docs/Grammar.md) for
+more information regarding the syntax of oeedger8r.
+
+To pass the `NestedBlob` struct in an ECALL/OCALL, oeedger8r currently supports the following two cases:
+- deep-copy `in` parameter
+  ```
+  // EDL definition
+  void foo([in, count=1] NestedBlob *nb);
+  ```
+
+  The above code snippet shows an example of passing `nb` as an `in` parameter. In this case,
+  the caller is responsible for initializing `nb`, including `nb->num`, the memory pointed by `nb->blob_array`,
+  and each element of the array. The callee can then read the nested content of `nb`. However,
+  any callee's modifications to `nb` will not reflect on the caller site (i.e., `nb` is effectively read-only to the callee).
+
+- deep-copy `inout` parameter
+  ```
+  // EDL definition
+  void foo([in, out, count=1] NestedBlob *nb);
+  ```
+
+  The above definition is an example of passing `nb` as an `inout` parameter. Similar to the case
+  of the `in` parameter, the caller is responsible for initializing `nb` and the callee can read
+  the nested content of the `nb`. Besides, callee's modifications to `nb` will now reflect on the caller
+  site with a limitation that the callee cannot re-allocate or resize `nb->blob_array` and each `buf`
+  in the nested element (which could lead to undefined behaviors).
+
+The case that remains unsupported is the deep-copy `out` parameter as shown in the following example.
+```
+// EDL definition
+void foo([out, count=1] NestedBlob *nb);
+```
+The major difference between the `out` parameter and the other two is that the caller is no longer responsible
+for initializing `nb`.
+Instead, the callee takes charge of initializing `nb`, which implies that `b->blob_array` and each of the nested
+`buf` are variable-length. Such support is required by typical scenarios in which the caller cannot determine the size of
+nested buffers in a struct. Note that the caller is still responsible for allocating the `nb` itself. Please refer to the code snippet
+in the following section for an example.
+
+Proposed Design
+-----------
+
+This document proposes to add the support of deep-copy `out` parameter to OE SDK.
+
+### User Experience
+
+The following code snippet serves as an example that the proposal aims to support.
+
+- Caller
+  
+  The caller, either an enclave or a host, is required to prepare an instance of `NestedBlob`.
+  In this example, `nb` is the instance of `NestedBlob`. The caller then passes `nb` to
+  `foo` (OCALL or ECALL). After the execution resumes from `foo`, the caller should expect that `nb`
+  has been initialized by the callee.
+  Note that the caller should not initialize the content of `nb` before invoking `foo`. Any content
+  set before the invocation will not be passed to the callee (i.e., by the definition of an `out`
+  parameter) and the content will be overwritten after a successful invocation.
+
+  ```
+  NestedBlob nb;
+  memset(&nb, 0, sizeof(NestedBlob));
+  // nb.num and nb.blob_array have not been initialized yet.
+  foo(enclave, &nb); // The case of an ECALL,
+                     // or simply foo(&nb) for the case of an OCALL.
+  ...
+  /* Caller is responsible for freeing the buffers after use. */
+  if (nb.blob_array)
+  {
+      for (size_t i = 0; i < nb.num; i++)
+          free(nb.blob_array[i].buf);
+      free(nb.blob_array);
+  }
+  ```
+
+  - Memory management
+  
+    The caller is responsible for managing the content of `nb` after invoking `foo`.
+    Each nested buffer of `nb` is allocated separately and is therefore required to
+    releas each of the buffers via `free` as shown in the example.
+
+- Callee
+
+  The callee is responsible for initializing the `nb` if the `nb` is not `NULL`.
+  The `nb` being `NULL` indicates that the caller passes `NULL` to `foo`.
+  
+  ```
+  // User-defined code
+  void foo(NestedBlob *nb)
+  {
+      if (!nb)
+          return;
+      nb->num = 5;
+      nb->blob_array = (Blob*)malloc(5 * sizeof(Blob));
+      for (size_t i = 0; i < 5; i++)
+      {
+          Blob* b = &ptr->blob_array[i];
+          b->len = 10;
+          b->buf = (char*)malloc(10);
+          memset(b->buf, 'A', 10);
+      }
+  }
+  ```
+
+  - Memory management
+
+    The callee should **allocate each nested buffer separately on the heap via `malloc` ** as shown in the example.
+    These allocated buffers are freed automatically by the oeedger8r-generated code and therefore the
+    callee does not need to take further actions. 
+    
+Note that the proposal does not affect the behavior of the existing deep-copy `inout` parameter.
+
+### Design
+
+The idea of the design is that after the callee (either a host or an enclave) initializes `nb`,
+oeedger8r-generated code serializes the content of `nb` into a local buffer. Next, the content of the buffer
+is copied over to a caller's buffer. How the buffer is transmitted is TEE-specific;
+in the case of SGX in which the enclave has access to the memory on the host, the enclave, either
+being the caller or callee, is always responsible for copying the buffer.
+Finally, the caller unserializes the received content, which fills up the `nb`.
+The rest of the section walks through modified oeedger8r-generated code for both ECALL and OCALL that realizes the design.
+
+Note that this design is compatible with the switchless call feature, which also requires shared memory.
+
+- ECALL
+
+  The design adds `deepcopy_out_buffer` and `deepcopy_out_buffer_size` members to every internal struct of ECALLs.
+  The two members are used to hold and to obtain the enclave-set content by the enclave and the host, correspondingly.
+  How the content is transmitted across the boundary is TEE-specific, depending on the availability of the shared memory
+  between the host and the enclave. The values (zero or not) of the two members are also used by the
+  infrastructure to determine whether to transmit the buffer or not.
+  
+  ```diff
+  typedef struct _foo_args_t
+  {
+      oe_result_t _result;
+  +   uint8_t* deepcopy_out_buffer;
+  +   size_t deepcopy_out_buffer_size;
+      NestedBlob* ptr;
+  } foo_args_t;
+  ```
+  - Host
+
+    Similar to a normal ECALL, the host prepares parameters and invokes `oe_call_enclave_function`.
+    After returning from `oe_call_enclave_function`, the host obtains `_deepcopy_out_buffer` and
+    `_deepcopy_out_buffer_size` from `pargs_out`. `_deepcopy_out_buffer` holds a local copy of the enclave-set content.
+    Next, the host uses the `OE_SET_DEEPCOPY_OUT_PARAM` macro to unserialize the enclave-set content. After unserializing,
+    the host validates the `_deepcopy_out_buffer_offset` (increased during unserialization) against `_deepcopy_out_buffer_size`.
+    Before finishing the function, the host frees `_deepcopy_out_buffer`.
+    
+The following is an example of oeedger8r-generated code:
+    ```diff
+    oe_result_t foo(
+        oe_enclave_t* enclave,
+        NestedBlob* ptr)
+    {
+        oe_result_t _result = OE_FAILURE;
+
+        static uint64_t global_id = OE_GLOBAL_ECALL_ID_NULL;
+
+        /* Marshalling struct. */
+        foo_args_t _args, *_pargs_in = NULL, *_pargs_out = NULL;
+        /* Marshalling buffer and sizes. */
+        size_t _input_buffer_size = 0;
+        size_t _output_buffer_size = 0;
+        size_t _total_buffer_size = 0;
+        uint8_t* _buffer = NULL;
+        uint8_t* _input_buffer = NULL;
+        uint8_t* _output_buffer = NULL;
+        size_t _input_buffer_offset = 0;
+        size_t _output_buffer_offset = 0;
+        size_t _output_bytes_written = 0;
+    +   uint8_t* _deepcopy_out_buffer = NULL;
+    +   size_t _deepcopy_out_buffer_size = 0;
+    +   size_t _deepcopy_out_buffer_offset = 0;
+
+       /* Fill marshalling struct. */
+        memset(&_args, 0, sizeof(_args));
+        _args.ptr = (NestedBlob*)ptr;
+
+        /* Compute input buffer size. Include in and in-out parameters. */
+        OE_ADD_SIZE(_input_buffer_size, sizeof(foo_args_t));
+        /* There were no corresponding parameters. */
+    
+        /* Compute output buffer size. Include out and in-out parameters. */
+        OE_ADD_SIZE(_output_buffer_size, sizeof(foo_args_t));
+        if (ptr)
+            OE_ADD_SIZE(_output_buffer_size, ((size_t)1 * sizeof(NestedBlob)));
+    
+        /* Allocate marshalling buffer. */
+        _total_buffer_size = _input_buffer_size;
+        OE_ADD_SIZE(_total_buffer_size, _output_buffer_size);
+        _buffer = (uint8_t*)oe_malloc(_total_buffer_size);
+        _input_buffer = _buffer;
+        _output_buffer = _buffer + _input_buffer_size;
+        if (_buffer == NULL)
+        {
+            _result = OE_OUT_OF_MEMORY;
+            goto done;
+        }
+    
+        /* Serialize buffer inputs (in and in-out parameters). */
+        _pargs_in = (enc_foo_args_t*)_input_buffer;
+        OE_ADD_SIZE(_input_buffer_offset, sizeof(*_pargs_in));
+        /* There were no in nor in-out parameters. */
+    
+        /* Copy args structure (now filled) to input buffer. */
+        memcpy(_pargs_in, &_args, sizeof(*_pargs_in));
+
+        /* Call enclave function. */
+        if ((_result = oe_call_enclave_function(
+                 enclave,
+                 &global_id,
+                 __test_ecall_info_table[test_fcn_id_enc_foo].name,
+                 _input_buffer,
+                 _input_buffer_size,
+                 _output_buffer,
+                 _output_buffer_size,
+                 &_output_bytes_written)) != OE_OK)
+            goto done;
+
+        /* Setup output arg struct pointer. */
+        _pargs_out = (foo_args_t*)_output_buffer;
+        OE_ADD_SIZE(_output_buffer_offset, sizeof(*_pargs_out));
+    
+        /* Check if the call succeeded. */
+        if ((_result = _pargs_out->_result) != OE_OK)
+            goto done;
+
+        /* Currently exactly _output_buffer_size bytes must be written. */
+        if (_output_bytes_written != _output_buffer_size)
+        {
+            _result = OE_FAILURE;
+            goto done;
+        }
+
+    +   _deepcopy_out_buffer = _pargs_out->deepcopy_out_buffer;
+    +   _deepcopy_out_buffer_size = _pargs_out->deepcopy_out_buffer_size;
+
+        /* Unmarshal return value and out, in-out parameters. */
+        /* No return value. */
+
+    +   OE_READ_OUT_PARAM(ptr, (size_t)(((size_t)1 * sizeof(NestedBlob))));
+    +   if (ptr && ptr->blob_array)
+    +       OE_SET_DEEPCOPY_OUT_PARAM(ptr->blob_array, ((size_t)ptr->num * sizeof(Blob)), Blob*);
+    +   for (size_t _i_2 = 0; _i_2 < ptr->num; _i_2++)
+    +   {
+    +       if (ptr && ptr->blob_array && ptr->blob_array[_i_2].buf)
+    +           OE_SET_DEEPCOPY_OUT_PARAM(ptr->blob_array[_i_2].buf, ptr->blob_array[_i_2].len, char*);
+    +   }
+
+    +   if (_deepcopy_out_buffer_offset != _deepcopy_out_buffer_size)
+    +   {
+    +       _result = OE_FAILURE;
+    +       goto done;
+    +   }
+
+        _result = OE_OK;
+
+    done:
+        if (_buffer)
+            oe_free(_buffer);
+
+    +   if (_deepcopy_out_buffer)
+    +       oe_free(_deepcopy_out_buffer);
+
+        return _result;
+    }
+    ```
+
+  - Enclave
+
+    Upon the dispatching of the ECALL, the oeedger8r-generated
+    `ecall_foo` is invoked as usual. After the invocation of
+    the user-defined `foo` function, the enclave calculates the size
+    of content set by `foo`. Next, the enclave allocates an in-enclave
+    buffer, `_deepcopy_out_buffer`, and serializes the content into the buffer using the
+    `OE_WRITE_DEEPCOPY_OUT_PARAM` macro.
+    After serializing, the enclave allocates sets the `_deepcopy_out_buffer` and `_deepcopy_out_buffer_size`
+    to `pargs_out`. After the execution of the `ecall_foo`, the infrasturce will determine how
+    to transmit the buffer over to the host.
+    Before terminating the function, the enclave releases the content set by `foo`.
+    Therefore, only the host is responsible for managing the memory of `nb` set by the ECALL `foo`.
+
+    ```diff
+    static void ecall_foo(
+        uint8_t* input_buffer,
+        size_t input_buffer_size,
+        uint8_t* output_buffer,
+        size_t output_buffer_size,
+        size_t* output_bytes_written)
+    {
+        oe_result_t _result = OE_FAILURE;
+
+        /* Prepare parameters. */
+        foo_args_t* pargs_in = (foo_args_t*)input_buffer;
+        foo_args_t* pargs_out = (foo_args_t*)output_buffer;
+
+    +   uint8_t* _deepcopy_out_buffer = NULL;
+    +   size_t _deepcopy_out_buffer_offset = 0;
+    +   size_t _deepcopy_out_buffer_size = 0;
+
+        size_t input_buffer_offset = 0;
+        size_t output_buffer_offset = 0;
+        OE_ADD_SIZE(input_buffer_offset, sizeof(*pargs_in));
+        OE_ADD_SIZE(output_buffer_offset, sizeof(*pargs_out));
+
+        /* Make sure input and output buffers lie within the enclave. */
+        /* oe_is_within_enclave explicitly checks if buffers are null or not. */
+        if (!oe_is_within_enclave(input_buffer, input_buffer_size))
+            goto done;
+
+        if (!oe_is_within_enclave(output_buffer, output_buffer_size))
+            goto done;
+
+        /* Set in and in-out pointers. */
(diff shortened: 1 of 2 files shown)


docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,220 @@
...
+designs that support the `out` parameter based on the availability of shared memory between
+a host and an enclave.
+
+Introduction
+------------
+
+The oeedger8r tool supports the deep copy feature that allows a user to pass a struct with nested pointers
+in ECALLs and OCALLs. To leverage this feature, the user should define the struct as part of an EDL file
+and use proper annotations on both the struct and the ECALL/OCALL, which intends to pass the struct.
+An example of strcut definitions is as follows.
```
dthaler, line 15:
```suggestion
An example of struct definitions is as follows.
```

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,220 @@
...
+Such support is required by typical scenarios in which the caller cannot determine the size of `b->buf`.
+
+Proposed Design
+-----------
+
+An important constraint that affects the design of the `out` parameter support is the availability of
+shared memory between a host and an enclave. Given that this support may not be general across TEEs,
+we discuss two designs in this section.
+
+The following code snippet serves as a common example that both two design aims to support.
```
dthaler, line 71:
```suggestion
The following code snippet serves as a common example that both designs aim to support.
```

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,220 @@
...
+- Host
+  ```
+  Blob b;
+  // b->len and b->buf have not been initialized yet.
+  foo(enclave, &b);
+  // b->len and b->buf have been initialized.
+  // b->len == 10 and b->data equals to data.
+  ```
```
dthaler, line 79: Incomplete code.  How is b->data freed?   Please include in the sample what the developer needs to do to free the blob.  e.g., is it
free(b->data);
?

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,220 @@
...
+    b->buf = malloc(b->size);
+    memcpy(b->buf, data, b->size);
+  }
+  ```
+
+### Design with Shared Memory
+
+With the support of shared memory, the design is more straightforward.
+The core idea of the design is that after the callee initializes the `b`,
+it also allocates a host memory and copies the content `b` to the allocated
```
dthaler, line 96: Line 29 talks about both directions (ECALL and OCALL) but here only talks about one direction (ECALL).
What about out parameters of OCALLs?

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,220 @@
...
+      memcpy(lhs->buf, b.buf, lhs->len);
+    ...
+  }
+  ```
+
+### Design without Shared Memory
+
+The design without shared memory is more complicated in terms of implementation.
+The high-level idea of the design is that after the callee initializes `b`,
+explicitly invoking an OCALL that takes `b` as an `in` parameter.
```
dthaler, line 142: Again this only talks about out parameters of ECALLs, what about out parameters of OCALLs?

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,220 @@
...
+
+- Host
+
+  In addition to the usual oeedger8r-generated ECALL wrapper, the design
+  adds an internal OCALL definition and implementation that will be
+  automatically invoked by the enclave.
+  Note that goal of `b_out` being `user_check` is to forward the original value
```
dthaler, line 149:
```suggestion
  Note that the goal of `b_out` being `user_check` is to forward the original value
```

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,220 @@
...
+    ...
+    // Initialize the host memory
+    Blob *lhs = (Blob *) pargs->in;
+    lhs->len = b.len;
+    lhs->buf = oe_host_malloc(lhs->len);
+    if (lhs->buf && b.buf)
+      memcpy(lhs->buf, b.buf, lhs->len);
+    ...
+  }
+  ```
```
dthaler, line 136: Question: will the above also work for out parameters on _switchless_ calls?

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: Updated.

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: Updated.

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: Updated.

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: Good point. That's something we need to consider. Taking the switchless calls into account may further complicate the design without shared memory.

mingweishih:
@HernanGatta I have a question regarding the shared memory support on OP-TEE. I do see something in the document: https://optee.readthedocs.io/en/latest/architecture/core.html#shared-memory. Can you add more on this? If shared memory is available on OP-TEE as well, we could just go with the simpler design/implementation.

HernanGatta:
> @HernanGatta I have a question regarding the shared memory support on OP-TEE. I do see something in the document: https://optee.readthedocs.io/en/latest/architecture/core.html#shared-memory. Can you add more on this? If shared memory is available on OP-TEE as well, we could just go with the simpler design/implementation.

Hello Ming-Wei, thank you for tagging me in this. I'm thinking it through and trying to see what the best way to go about this is on OP-TEE.

docs/DesignDocs/DeepCopyOutParameters.md
radhikaj: switchless calls are not supported on OPTEE.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,266 @@
...
+
+    In the case of OCALL, the callee invokes the user-defined `foo` as usual.
+    ```
+    // Oeedger8r-generated code
+    void ocall_foo()
+    {
+      ...
+      Blob b;
+      foo(&b);
+      ...
```
dthaler, line 178: why is there no free(b->buf); ?

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,266 @@
...
+  // User-defined code
+  void foo(Blob *b)
+  {
+    b->len = 10;
+    b->buf = malloc(b->size);
+    memcpy(b->buf, data, b->size);
+  }
+  ```
+  Note that `b->buf` also needs to be freed on the callee side, which is left to the implementation
+  details.
```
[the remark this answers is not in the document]
mingweishih: We can free the allocated buffer automatically (in the oeedger8r-generated code). However, this may not work with 
 a corner case: instead of allocating memory, the callee assigns a global pointer to the `b->buf`. See the following example.
```
/* Managed somewhere else */
char *g_buf;
size_t g_buf_len;

void foo(Blob *b)
{
    b->len = g_buf_len;
    b->buf = g_buf;
}
```

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,266 @@
...
+      ...
+      Blob b;
+      foo(&b);
+      ...
+      // Initialize the host memory
+      Blob *lhs = (Blob *) pargs->in;
+      lhs->len = b.len;
+      lhs->buf = oe_host_malloc(lhs->len);
+      if (lhs->buf && b.buf)
+        memcpy(lhs->buf, b.buf, lhs->len);
```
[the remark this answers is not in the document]
mingweishih: Thanks for the suggestion. But it's unclear how memdup supports the copy of nested pointers, which I need to think.

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: See my previous comment regarding the special case.

docs/DesignDocs/DeepCopyOutParameters.md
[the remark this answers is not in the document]
dthaler: I don't understand what you're proposing as an answer.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,335 @@
...
+        b->buf = NULL;
+        if (_pargs_out->b->buf)
+        {
+          b->buf = oe_malloc(b->len);
+          if (!b->buf)
+          {
+            _result = OE_FAILURE;
+            goto done;
+          }
+          memcpy(b->buf, _pargs_out->b->buf, b->len);
```
bodzhang, line 211: If higher level code did not verify that `_pargs_out->b->buf`, with length of `b->len`, is fully outside of the enclave memory already, the code here must verify that before performing the copy.

mingweishih: `b->len` is inside the enclave while `_pargs_out->b->len` is on the hosot.

anakrish (commented):
Thanks for investigating this complex problem, writing up the design and seeking consensus.

It looks good to me. However, I'd like to see the following addressed:

1) One of the key concerns raised during security review of deep-copy was that objects ought to be transferred in a single memcpy across the host/enclave boundary. Copying an object tree, object by object is known to be error prone and hard to get right.
The current proposal seems to perform a malloc for every object, then copy it over, then malloc and copy the child objects. Can this be changed so that objects are copied across the host/enclave security boundary in a single memcpy and then reconstructed at the other side.
This is the approach edger8r currently takes prior to this proposal.


2) Variable length outputs can be encoded as two calls where the first call returns the size of the variable length output and the second call returns the object. 
    For example, the unsupported edge function
    ```c
    void get_data([out]size_t* size, [out] uint8_t** data);
    ```
    can be expressed as the following that works in both SGX and OPTEE
    ```c
     void get_data_size([out]size_t* size);
     void get_data(size_t size, [out,size=size] uint8_t* data);
    ```
One can envision a code generation scheme where the edger8r auto-generates and uses two calls in the same direction to implement deep-copy out parameters. 
The proposal does not clarify whether such an approach was considered or not.

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: The design has been modified and hence the comments are no longer applicable.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,344 @@
...
+      if (_pargs_out->b)
+      {
+        Blob *_rhs = (Blob *) _pargs_in->b;
+        _pargs_out->len = _rhs->len;
+        _pargs_out->b = NULL;
+        if (_rhs->b)
+        {
+          /* Use a local variable to prevent the host from
+             manipulating the address value used by memcpy below. */
+          char* tmp_buf = oe_host_malloc(_rhs->len);
```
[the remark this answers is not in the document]
mingweishih: Added `oe_allocate_host_buffer` and `oe_free_host_buffer`.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,344 @@
...
+  ```
+
+- Callee
+  ```
+  // User-defined code
+  void foo(Blob *b)
+  {
+    int result = 0;
+    b->len = 10;
+    b->buf = malloc(b->size);
```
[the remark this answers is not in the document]
mingweishih: Updated design.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,344 @@
...
+
+The following code snippet serves as a common example that both designs aim to support.
+- Caller
+  ```
+  Blob b;
+  // b->len and b->buf have not been initialized yet.
+  foo(enclave, &b);
+  // b->len and b->buf have been initialized.
+  // b->len == 10 and b->data equals to data.
+  free(b->buf);
```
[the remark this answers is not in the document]
mingweishih: Updated the example.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,344 @@
...
+  site with a limitation that the callee cannot re-allocate or resize `b->buf`
+  (which could lead to undefined behaviors).
+
+The case that remains unsupported is the `out` parameter as shown in the following example.
+```
+// EDL definition
+void foo[out, count=1] Blob *b);
+```
+The major difference between the `out` parameter and the other two is that caller is no longer responsible
+for initializing `b` (i.e., the caller can simply pass a `b` pointer with `b->buf` equals to `NULL`).
```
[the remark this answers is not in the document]
mingweishih: Updated the doc.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,344 @@
...
+        if (_pargs_out->b)
+        {
+          oe_host_free(_pargs_out->b->buf);
+          _pargs_out->b->buf = NULL;
+        }
+      }
+      /* Always free the intermediate enclave buffers. */
+      if (_pargs_in->b)
+      {
+        oe_free(_pargs_in->b->buf);
```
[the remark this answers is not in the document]
mingweishih: Updated.

mingweishih:
Thanks for the review, I've updated the both the design and the documentation.

> Thanks for investigating this complex problem, writing up the design and seeking consensus.
> 
> It looks good to me. However, I'd like to see the following addressed:
> 
> 1. One of the key concerns raised during security review of deep-copy was that objects ought to be transferred in a single memcpy across the host/enclave boundary. Copying an object tree, object by object is known to be error prone and hard to get right.
>    The current proposal seems to perform a malloc for every object, then copy it over, then malloc and copy the child objects. Can this be changed so that objects are copied across the host/enclave security boundary in a single memcpy and then reconstructed at the other side.
>    This is the approach edger8r currently takes prior to this proposal.

The updated design now uses single memcpy across the boundary. More specifically,
- The caller serializes the out parameter into a local buffer.
- The enclave performs a single memcpy that copies the buffer across the boundary.
- The callee re-constructs the out parameter.

> 2. Variable length outputs can be encoded as two calls where the first call returns the size of the variable length output and the second call returns the object.
>    For example, the unsupported edge function
>    ```c
>    void get_data([out]size_t* size, [out] uint8_t** data);
>    ```
>    
>    
>    can be expressed as the following that works in both SGX and OPTEE
>    ```c
>     void get_data_size([out]size_t* size);
>     void get_data(size_t size, [out,size=size] uint8_t* data);
>    ```
> 
> One can envision a code generation scheme where the edger8r auto-generates and uses two calls in the same direction to implement deep-copy out parameters.
> The proposal does not clarify whether such an approach was considered or not.

The new design uses the similar strategy but avoids using an additional ECALL/OCALL.
- The buffer size is calculated by the caller based on the `_output_bytes_written` that is set and returned by the callee.
- Introduce a `_host_buffer` to the internal `args_t` struct that is used to pass the host pointer.

I add the possible design for the double OCALL/ECALL scheme you suggested in the discussion section.
However, I would still prefer to implement the current design first and revisit this option later.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,838 @@
...
+is copied over to a caller's buffer. Given that only the enclave has access to the memory
+on the host, the enclave, either being the caller or callee, is always responsible for copying the buffer.
+Finally, the caller unserializes the received content, which fills up the `nb`.
+The rest of the section walks through modified oeedger8r-generated code for both ECALL and OCALL that realizes the design.
+
+Note that this design is compatible with switchelss call feature, which also requires shared memory.
+
+- ECALL
+
+  The design adds a `_host_buffer` member to the internal struct that holds the shared pointer.
```
anakrish, line 157: You could also add a `_host_buffer_size` member. The first 3 fields of any marshalling struct would then be:
```c
typedef struct _fcn_args_t
  {
      oe_result_t _result;
      uint8_t* _host_buffer;
      size_t _host_buffer_size;
      ... // other marshaling fields
  }
```

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,838 @@
...
+
+  - Enclave
+
+    Upon the dispatching of the ECALL, the oeedger8r-generated
+    `ecall_foo` is invoked as usual. After the invocation of
+    the user-defined `foo` function, the enclave calculates the size
+    of content set by `foo`. Next, the enclave allocates an in-enclave
+    buffer, `_deepcopy_out_buffer`, and serializes the content into the buffer using the
+    `OE_WRITE_DEEPCOPY_OUT_PARAM` macro.
+    After serializing, the enclave allocates a host buffer (held by the `pargs_out->_host_buffer`
```
anakrish, line 316: This does not have to be done by the generated code. 
The generated code just needs to do 
```c
    args->_host_buffer = _deepcopy_out_buffer;
    args->_host_buffer_size = _deepcopy_out_buffer_size;
```

The ECALL dispatching infrastructure can determine the best way to marshal _host_buffer over to the host, according to the platform (SGX/OPTEE).

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,838 @@
...
+    +   }
+
+    +   if (_deepcopy_out_buffer_offset != _deepcopy_out_buffer_size)
+    +   {
+    +       _result = OE_FAILURE;
+    +       goto done;
+    +   }
+    +   OE_ADD_SIZE(output_buffer_offset, _deepcopy_out_buffer_size);
+
+    +   /* Allocate a host buffer and copy the content over to the buffer. */
```
anakrish, line 407: Instead of this, we could generate:
```
pargs_out->_host_buffer = _deepcopy_out_buffer;
pargs_out->_host_buffer_size = deepcopy_out_buffer_size;
```

The infrastructure can marshal this over to the host as appropriate. The generated code doesn't have to deal with host memory.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,838 @@
...
+
+- OCALL
+
+  Similar to the case of ECALL, the design adds a `_host_buffer` member to the internal struct that holds the shared pointer.
+  The pointer holds the host-set content after `foo` successfully invokes `oe_call_host_function`.
+  ```diff
+  typedef struct _foo_args_t
+  {
+      oe_result_t _result;
+  +   uint8_t* _host_buffer;
```
anakrish, line 463: As for ecall, add `_host_buffer_size`.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,838 @@
...
+
+    +   if (_deepcopy_out_buffer_offset != _deepcopy_out_buffer_size)
+    +   {
+    +       _result = OE_FAILURE;
+    +       goto done;
+    +   }
+    +   OE_ADD_SIZE(output_buffer_offset, _deepcopy_out_buffer_size);
+
+    +   /* Pass the pointer of the deep-copied content to the enclave. */
+    +   pargs_out->_host_buffer = _deepcopy_out_buffer;
```
anakrish, line 699: Once the earlier suggestions for ecall are incorporated, the generated code for ecall/ocall deep-copy out parameters will look identical.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,838 @@
...
+
+        if (pargs_out && output_buffer_size >= sizeof(*pargs_out))
+            pargs_out->_result = _result;
+    }
+    ```
+
+Discussion
+-----------
+
+The proposed design relies on the shared memory between an enclave and a host and therefore does not apply
```
anakrish, line 742: By moving host buffer allocation to the infrastructure rather than the generated code, there is no longer any reliance on edger8r generated code on shared memory. It is the infrastructure that needs to marshal the deep-copy parameter appropriately on each platform (SGX, OPTEE).

anakrish (approved):
Looks great!

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: Updated.

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: Updated.

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: Updated. Currently, the enclave (OCALL wrapper) performs an additional check against the buffer to ensure it resides in enclave memory.

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: Removed the section.

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: Updated.

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: Updated.

radhikaj:
@dthaler, this is an experimental feature and follows the proposal we discussed in SIF-Arch last November. Would it be possible for you to take another look?

docs/DesignDocs/DeepCopyOutParameters.md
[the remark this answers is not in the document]
mingweishih: The design has been updated such that the memory transmission is now handled internally, not by the oeedger8r-generated code. Please take a look.

docs/DesignDocs/DeepCopyOutParameters.md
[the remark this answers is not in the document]
mingweishih: Now the callee-allocated memory will be freed automatically. Please check the updated document.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,266 @@
...
+  ```
+
+  The above definition is an example of passing `b` as an `inout` parameter. Similar to the case
+  of the `in` parameter, the caller is responsible for initializing `b` and the callee can read
+  the content of the `b`. Besides, callee's modifications to `b` will now reflect on the caller
+  site with a limitation that the callee cannot re-allocate or resize `b->buf`
+  (which could lead to undefined behaviors).
+
+
+The case that remains unsupported is the `out` parameter as shown in the following example.
```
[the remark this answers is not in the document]
mingweishih: Updated the proposal to mention that the design does not change the behavior of existing `inout` parameter.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,718 @@
...
+
+This document provides a brief overview of the oeedger8r's deep copy feature and points out
+its missing piece: the support of the deep-copy `out` parameter. Next, the document describes a
+design that supports the deep-copy `out` parameter.
+
+Introduction
+------------
+
+The oeedger8r tool supports the feature of deep copy, which allows a user to pass a struct with nested pointers
+in ECALLs and OCALLs. To leverage this feature, the user should define the struct as part of an EDL file
```
dthaler, line 12: Recommend adding a reference to full_edl_serialization.md, and updating full_edl_serialization.md with a reference to this design doc.  Also full_edl_serialization.md includes the statement:

> Although EDL provides sufficient mechanisms for performing full serialization, the generator does not currently enforce these mechanisms. By default, the generator produces edge routines that implicitly copy untrusted memory references into the enclave. This results in potential security flaws and non-portable code.
> 
> To overcome these problems, the generator can be modified to warn when an EDL specification cannot be fully serialized.

I'd suggest updating that text in this PR.

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,718 @@
...
+  [count=num] Blob* blob_array;
+};
+```
+
+The `num` variable specified to the `count` attribute represents the length (i.e., the number of elements)
+of the `blob_array`. Similarly, the `len` variable specified to the `size` attribute represents the size (in bytes) of the `buf`.
+Refer to the [Grammar documentation](https://github.com/openenclave/oeedger8r-cpp/blob/master/docs/Grammar.md) for
+more information regarding the syntax of oeedger8r.
+
+To pass the `NestedBlob` struct in an ECALL/OCALL, the oeedger8r currently supports the following two cases:
```
dthaler, line 35:
```suggestion
To pass the `NestedBlob` struct in an ECALL/OCALL, oeedger8r currently supports the following two cases:
```

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,718 @@
...
+This document proposes to add the support of deep-copy `out` parameter to OE SDK.
+
+### User Experience
+
+The following code snippet serves as an example that the proposal aims to support.
+
+- Caller
+  
+  The caller, either an enclave or a host, is required to prepare an instant of `NestedBlob`.
+  In this example, `nb` is the instant of `NestedBlob`. The caller then passes `nb` to
```
dthaler, line 83:
```suggestion
  The caller, either an enclave or a host, is required to prepare an instance of `NestedBlob`.
  In this example, `nb` is the instance of `NestedBlob`. The caller then passes `nb` to
```

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,718 @@
...
+    The callee should **allocate each nested buffer separately on the heap** as shown in the example.
+    These allocated buffers are freed automatically by the oeedger8r-generated code and therefore the
+    callee does not need to take further actions. 
+    
+Note that the proposal does not affect the behavior of the existing deep-copy `inout` parameter.
+
+### Design
+
+The idea of the design is that after the callee (either a host or an enclave) initializes `nb`,
+the callee serializes the content of `nb` into a local buffer. Next, the content of the buffer
```
dthaler, line 146:
```suggestion
oeedger8r-generated code serializes the content of `nb` into a local buffer. Next, the content of the buffer
```

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,718 @@
...
+is copied over to a caller's buffer. How the buffer is transmitted is TEE-specific;
+in the case of SGX in which the enclave has access to the memory on the host, the enclave, either
+being the caller or callee, is always responsible for copying the buffer.
+Finally, the caller unserializes the received content, which fills up the `nb`.
+The rest of the section walks through modified oeedger8r-generated code for both ECALL and OCALL that realizes the design.
+
+Note that this design is compatible with switchelss call feature, which also requires shared memory.
```
dthaler, line 153:
```suggestion
Note that this design is compatible with the switchless call feature, which also requires shared memory.
```

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,718 @@
...
+
+- ECALL
+
+  The design adds `deepcopy_out_buffer` and `deepcopy_out_buffer_size` members to every internal structs of ECALLs.
```
dthaler, line 157:
```suggestion
  The design adds `deepcopy_out_buffer` and `deepcopy_out_buffer_size` members to every internal struct of ECALLs.
```

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,718 @@
...
+  The two members are used to hold and to obtain the enclave-set content by the enclave and the host, correspondingly.
+  How the content is transmitted across the boundary is TEE-specific, depending on the availability of the shared memory
+  between the host and the enclave. The value (zero or not) of the two members are also used by the
```
dthaler, line 160:
```suggestion
  between the host and the enclave. The values (zero or not) of the two members are also used by the
```

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,718 @@
...
+  ```
+  - Host
+
+    Similar to a normal ECALL, the host prepares parameters and invokes `oe_call_enclave_function`.
+    After returning from `oe_call_enclave_function`, the host obtains `_deepcopy_out_buffer` and
+    `_deepcopy_out_buffer_size` from `pargs_out`. `_deepcopy_out_buffer` holds a local copy of the enclave-set content.
+    Next, the host uses the `OE_SET_DEEPCOPY_OUT_PARAM` macro to unserialize the enclave-set content. After unserializing,
+    the host validates the `_deepcopy_out_buffer_offset` (increased during unserialization) against `_deepcopy_out_buffer_size`.
+    Before finishing the function, the host frees `_deepcopy_out_buffer`.
+
```
dthaler, line 180:
```suggestion
    
The following is an example of oeedger8r-generated code:
```

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,718 @@
...
+        if (output_buffer_size >= sizeof(*pargs_out) &&
+            oe_is_within_enclave(pargs_out, output_buffer_size))
+            pargs_out->_result = _result;
+    }
+    ```
+
+- OCALL
+
+  Similar to the case of ECALL, the design adds the same modifications to every internal structs of OCALLs.
```
dthaler, line 439:
```suggestion
  Similar to the case of ECALL, the design adds the same modifications to every internal struct of OCALLs.
```

docs/DesignDocs/DeepCopyOutParameters.md
```diff
@@ -0,0 +1,718 @@
...
+    ```
+
+Conclusion and Plan
+-----------
+
+This document presents the design, which supports deep-copy `out` parameter.
+Moving forward, we plan to support the design as an experimental feature first.
+That is, users need to explicitly opt-in the feature (i.e., specifying `--experimental` flag in the
+command line when invoking the oeedger8r tool). Doing so allows OE to support the deep-copy `out` parameter
+in SGX while not introducing breaking changes to the OE's support on the OP-TEE.
```
dthaler, line 713:
```suggestion
in SGX while not introducing breaking changes to the OE's support on OP-TEE.
```

dthaler (approved):
Design looks good to me, just some comments.

docs/DesignDocs/DeepCopyOutParameters.md
dthaler, line 713:
```suggestion
in SGX while not introducing breaking changes to the OE's support on the OP-TEE.
```
@HernanGatta is there anything in here that would break something that works on OP-TEE?  If not, then this doesn't need to be experimental.

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: Updated. Please take a look.

docs/DesignDocs/DeepCopyOutParameters.md
mingweishih: The design won't break existing OP-TEE ECALLs/OCALLs that don't use the deep-copy out annotations (there shouldn't be any).
Even if developers try to use deep-copy out on OP-TEE, the compilation would go thru but the code just won't work unless we implement the OP-TEE support internally.

Given that we currently don't specify target TEE to the oeedger8r tool, we don't have a good way to tell developers that deep-copy out is not supported on OP-TEE. As a result, the current plan is guarding the feature with the `--experimental` flag, which makes the feature implicit. If the flag is not set, the developer will see an error message like:
```
`out' for user defined type `MyStruct\\*' requires the --experimental option`.
```

@dthaler @anakrish Do you think we can make this as non-experimental? We do have the switchless call feature that is similar to this case (only works for SGX).

Alternatively, we could introduce a TEE flag to oeedger8r or let the tool to detect the target based on the platform (x86 or arm). We could also add the OP-TEE support, which requires @HernanGatta's inputs. I think both of them could be future work.

anakrish: Let's first merge this in as experimental, and get feedback from the user that wants this.

dthaler: Unless we know of things that would break OP-TEE, I don't think OP-TEE should be any reason to make this be experimental.  If we think it needs to be experimental for SGX because SGX might break, then ok.  But I'd probably argue for non-experimental.

mingweishih: We're thinking of following the previous path on deep-copy support: making the feature as experimental first and then get feedback from users/ask for security reviews before making this as non-experimental.

radhikaj: I agree with what Ming-Wei said. Let’s get feedback and see if some folks can look at the security implications. Cc @mrragava

radhikaj:
I am merging this since it has 2 approvals. A move to non experimental can happen in a different PR

Issue #38, opened 2017-11-22
Document Windows enclave debugging instructions


CodeMonkeyLeet:
Follow up to #37, we need to provide clear guidance an instructions for debugging enclaves on Windows.

radhikaj:
This is now documented.

PR #3700, opened 2020-11-01
Add basic KSS support (ISVFAMILYID and ISVEXTPRODID) to OE SDK

jianlia:

Enclave loader to detect whether the system support KSS

### DIFF
--- a/docs/GettingStartedDocs/buildandsign.md
+++ b/docs/GettingStartedDocs/buildandsign.md
@@ -97,9 +97,40 @@ NumHeapPages=1024
 NumStackPages=1024
 NumTCS=1
 ```
+Additionally, a developer can specify additional Key Sharing and Separation (KSS) identity properties
+for use on the platforms that support it (for SGX enclave only):
+- **FamilyID**: The product family identity (ISVFAMILYID for SGX) a developer can specify to group
+different enclaves under a common identity such as an identifier for the application suite
+which includes several enclave apps.
+- **ExtendedProductID**: The extended product identity (ISVEXTPRODID for SGX) value that a developer
+can use as 128-bit globally unique identifier for the enclave where the 16-bit ProductID proves too restrictive.
+For more details, see Table 37-19 of the
+[Intel's Software Developers Manual](https://software.intel.com/content/www/us/en/develop/articles/intel-sdm.html#combined/).
 
-As a convenience, you can also specify the enclave properties in code using the
-`OE_SET_ENCLAVE_SGX` macro.  For example, the equivalent properties could be
+For example, in kss_valid.conf:
+```
+FamilyID=47183823-2574-4bfd-b411-99ed177d3e43
+ExtendedProductID=2768c720-1e28-11eb-adc1-0242ac120002
+```
+
+As a convenience, you can specify the enclave properties in code using the
+`OE_SET_ENCLAVE_SGX_KSS` macro if KSS properties are included.  For example, the equivalent properties could be
+defined in any .c or .cpp file compiled into the enclave:
+
+```c
+OE_SET_ENCLAVE_SGX_KSS(
+    1,    /* ProductID */
+    1,    /* SecurityVersion */
+    1,    /* Debug */
+    47183823-2574-4bfd-b411-99ed177d3e43, /* FamilyID */
+    2768c720-1e28-11eb-adc1-0242ac120002, /* ExtendedProductID */
+    1024, /* NumHeapPages: heap size in units of 4KB pages */
+    1024, /* NumStackPages: stack size, in units of 4KB pages */
+    1);   /* NumTCS */
+```
+
+You can also specify the enclave properties in code using the
+`OE_SET_ENCLAVE_SGX` macro if no KSS properties needed.  For example, the equivalent properties could be
 defined in any .c or .cpp file compiled into the enclave:
 
 ```c
--- a/host/sgx/create.c
+++ b/host/sgx/create.c
@@ -84,6 +84,21 @@ static void _initialize_enclave_host()
 }
 #endif // OEHOSTMR
 
+static bool _is_kss_supported()
+{
+    uint32_t eax, ebx, ecx, edx;
+    eax = ebx = ecx = edx = 0;
+
+    // Obtain feature information using CPUID
+    oe_get_cpuid(0x12, 0x1, &eax, &ebx, &ecx, &edx);
+
+    // Check if KSS (bit 7) is supported by the processor
+    if (!(eax & (1 << 7)))
+        return false;
+    else
+        return true;
+}
+
 static oe_result_t _add_filled_pages(
     oe_sgx_load_context_t* context,
     uint64_t enclave_addr,
@@ -589,6 +604,24 @@ oe_result_t oe_sgx_validate_enclave_properties(
         goto done;
     }
 
+    if (!(properties->config.attributes & OE_SGX_FLAGS_KSS))
+    {
+        if (!oe_sgx_is_unset_uuid(
+                (uint8_t*)properties->config.extended_product_id))
+        {
+            OE_TRACE_ERROR("oe_sgx_is_unset_uuid failed: extended_product_id "
+                           "should be empty");
+            result = OE_FAILURE;
+            goto done;
+        }
+        if (!oe_sgx_is_unset_uuid((uint8_t*)properties->config.family_id))
+        {
+            OE_TRACE_ERROR(
+                "oe_sgx_is_unset_uuid failed: family_id should be empty");
+            result = OE_FAILURE;
+            goto done;
+        }
+    }
     result = OE_OK;
 
 done:
@@ -679,6 +712,8 @@ static oe_result_t _eeid_resign(
             properties->config.security_version,
             OE_DEBUG_SIGN_KEY,
             OE_DEBUG_SIGN_KEY_SIZE,
+            properties->config.family_id,
+            properties->config.extended_product_id,
             sigstruct));
     }
 
@@ -831,6 +866,21 @@ oe_result_t oe_sgx_build_enclave(
         &props,
         &loaded_enclave_pages_size,
         &enclave_size));
+
+    if (props.config.attributes & OE_SGX_FLAGS_KSS)
+    {
+        if ((context->type == OE_SGX_LOAD_TYPE_CREATE) && !_is_kss_supported())
+        {
+            // Fail if the CPU does not support KSS and the enclave specifies
+            // the KSS flag
+            OE_RAISE_MSG(
+                OE_UNSUPPORTED,
+                "Enclave image was signed with kss flag but CPU doesn't "
+                "support KSS\n",
+                NULL);
+        }
+        context->attributes.flags |= OE_ENCLAVE_FLAG_SGX_KSS;
+    }
 
     /* Perform the ECREATE operation */
     OE_CHECK(oe_sgx_create_enclave(
--- a/host/sgx/sgxsign.c
+++ b/host/sgx/sgxsign.c
@@ -571,6 +571,8 @@ static oe_result_t _init_sigstruct(
     uint64_t attributes,
     uint16_t product_id,
     uint16_t security_version,
+    const uint8_t* family_id,
+    const uint8_t* extended_product_id,
     sgx_sigstruct_t* sigstruct)
 {
     oe_result_t result = OE_UNEXPECTED;
@@ -639,6 +641,23 @@ static oe_result_t _init_sigstruct(
         mrenclave,
         sizeof(*mrenclave)));
 
+    if (attributes & SGX_FLAGS_KSS)
+    {
+        if (family_id)
+            OE_CHECK(oe_memcpy_s(
+                sigstruct->isvfamilyid,
+                sizeof(sigstruct->isvfamilyid),
+                family_id,
+                sizeof(*family_id)));
+        if (extended_product_id)
+            OE_CHECK(oe_memcpy_s(
+                sigstruct->isvextprodid,
+                sizeof(sigstruct->isvextprodid),
+                extended_product_id,
+                sizeof(*extended_product_id)));
+        sigstruct->attributemask.flags |= SGX_FLAGS_KSS;
+    }
+
     /* sgx_sigstruct_t.isvprodid */
     sigstruct->isvprodid = product_id;
 
@@ -665,6 +684,8 @@ oe_result_t oe_sgx_sign_enclave_from_engine(
     const char* engine_id,
     const char* engine_load_path,
     const char* key_id,
+    const uint8_t* family_id,
+    const uint8_t* extended_product_id,
     sgx_sigstruct_t* sigstruct)
 {
     oe_rsa_private_key_t rsa;
@@ -684,7 +705,13 @@ oe_result_t oe_sgx_sign_enclave_from_engine(
 
     /* Initialize & sign the sigstruct */
     OE_CHECK(_init_sigstruct(
-        mrenclave, attributes, product_id, security_version, sigstruct));
+        mrenclave,
+        attributes,
+        product_id,
+        security_version,
+        family_id,
+        extended_product_id,
+        sigstruct));
     OE_CHECK(_sign_sigstruct(&rsa, sigstruct));
 
     result = OE_OK;
@@ -703,6 +730,8 @@ oe_result_t oe_sgx_sign_enclave(
     uint16_t security_version,
     const uint8_t* pem_data,
     size_t pem_size,
+    const uint8_t* family_id,
+    const uint8_t* extended_product_id,
     sgx_sigstruct_t* sigstruct)
 {
     oe_rsa_private_key_t rsa;
@@ -722,7 +751,13 @@ oe_result_t oe_sgx_sign_enclave(
 
     /* Initialize & sign the sigstruct */
     OE_CHECK(_init_sigstruct(
-        mrenclave, attributes, product_id, security_version, sigstruct));
+        mrenclave,
+        attributes,
+        product_id,
+        security_version,
+        family_id,
+        extended_product_id,
+        sigstruct));
     OE_CHECK(_sign_sigstruct(&rsa, sigstruct));
 
     result = OE_OK;
@@ -739,6 +774,8 @@ oe_result_t oe_sgx_get_sigstruct_digest(
     uint64_t attributes,
     uint16_t product_id,
     uint16_t security_version,
+    const uint8_t* family_id,
+    const uint8_t* extended_product_id,
     OE_SHA256* digest)
 {
     oe_result_t result = OE_UNEXPECTED;
@@ -753,7 +790,13 @@ oe_result_t oe_sgx_get_sigstruct_digest(
 
     /* Initialize & sign the sigstruct */
     OE_CHECK(_init_sigstruct(
-        mrenclave, attributes, product_id, security_version, &sigstruct));
+        mrenclave,
+        attributes,
+        product_id,
+        security_version,
+        family_id,
+        extended_product_id,
+        &sigstruct));
     OE_CHECK(_hash_sigstruct(&sigstruct, digest));
 
     result = OE_OK;
@@ -771,6 +814,8 @@ oe_result_t oe_sgx_digest_sign_enclave(
     size_t cert_pem_size,
     const uint8_t* digest_signature,
     size_t digest_signature_size,
+    const uint8_t* family_id,
+    const uint8_t* extended_product_id,
     sgx_sigstruct_t* sigstruct)
 {
     oe_result_t result = OE_UNEXPECTED;
@@ -796,7 +841,13 @@ oe_result_t oe_sgx_digest_sign_enclave(
 
     /* Initialize the sigstruct with the provided parameters */
     OE_CHECK(_init_sigstruct(
-        mrenclave, attributes, product_id, security_version, sigstruct));
+        mrenclave,
+        attributes,
+        product_id,
+        security_version,
+        family_id,
+        extended_product_id,
+        sigstruct));
 
     /* Verify that the digest of the resulting sigstruct still
      * matches the expected signature */
--- a/include/openenclave/bits/sgx/sgxproperties.h
+++ b/include/openenclave/bits/sgx/sgxproperties.h
@@ -31,6 +31,7 @@ typedef struct _oe_sgx_enclave_image_info_t
 // oe_sgx_enclave_properties_t SGX enclave properties derived type
 #define OE_SGX_FLAGS_DEBUG 0x0000000000000002ULL
 #define OE_SGX_FLAGS_MODE64BIT 0x0000000000000004ULL
+#define OE_SGX_FLAGS_KSS 0x0000000000000080ULL
 #define OE_SGX_SIGSTRUCT_SIZE 1808
 
 typedef struct oe_sgx_enclave_config_t
@@ -41,7 +42,9 @@ typedef struct oe_sgx_enclave_config_t
     /* Padding to make packed and unpacked size the same */
     uint32_t padding;
 
-    /* (OE_SGX_FLAGS_DEBUG | OE_SGX_FLAGS_MODE64BIT) */
+    uint8_t family_id[16];
+    uint8_t extended_product_id[16];
+    /* (OE_SGX_FLAGS_DEBUG | OE_SGX_FLAGS_MODE64BIT | OE_SGX_FLAGS_KSS) */
     uint64_t attributes;
 
     /* XSave Feature Request Mask */
@@ -71,8 +74,9 @@ typedef struct _oe_sgx_enclave_properties
     OE_EXTERNC __attribute__((section(OE_INFO_SECTION_NAME)))
 #define OE_INFO_SECTION_END
 
-#define OE_MAKE_ATTRIBUTES(ALLOW_DEBUG) \
-    (OE_SGX_FLAGS_MODE64BIT | (ALLOW_DEBUG ? OE_SGX_FLAGS_DEBUG : 0))
+#define OE_MAKE_ATTRIBUTES(ALLOW_DEBUG, REQUIRE_KSS)                   \
+    (OE_SGX_FLAGS_MODE64BIT | (ALLOW_DEBUG ? OE_SGX_FLAGS_DEBUG : 0) | \
+     (REQUIRE_KSS ? OE_SGX_FLAGS_KSS : 0))
 
 // This macro initializes and injects an oe_sgx_enclave_properties_t struct
 // into the .oeinfo section.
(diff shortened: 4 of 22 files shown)


host/sgx/create.c
```diff
@@ -832,6 +858,12 @@ oe_result_t oe_sgx_build_enclave(
         &loaded_enclave_pages_size,
         &enclave_size));
 
+    if (_is_kss_supported())
```
yentsanglee, line 861: This is my understanding from @bodzhang and @CodeMonkeyLeet's comments:

``c`
    if (_is_kss_supported())
    {
        // Set SECS.ATTRIBUTES.KSS bit if the CPU supports KSS
        context->attributes.flags |= OE_ENCLAVE_FLAG_KSS;
    }
    else if (props.config.attributes & OE_SGX_FLAGS_KSS)
	{
        // Fail if the CPU does not support KSS and the enclave specifies the KSS flag
        context->attributes.flags |= OE_ENCLAVE_FLAG_KSS;
		OE_RAISE_MSG(...);
    }
````

File: include/openenclave/host.h [hunk 2]
```diff
@@ -68,6 +68,12 @@ OE_EXTERNC_BEGIN
 /**
  * @cond DEV
  */
+
+/**
+ *  Flag passed into oe_create_enclave to run the enclave with kss mode.
```
yentsanglee, line 73: Please update the comment:

"Flag reserved for internal use to indicate the enclave was configured to run in kss mode"

include/openenclave/internal/sgxsign.h
```diff
@@ -25,6 +25,8 @@ OE_EXTERNC_BEGIN
  * @param security_version[in] ISVSVN value for the SGX sigstruct
  * @param pem_data[in] PEM buffer containing the signing key
  * @param pem_size[in] size of the PEM buffer
+ * @param family_id[in] ISVFAMILYID value for the SGX sigstruct
+ * @param ext_product_id[in] ISVEXTPRODID value for the SGX sigstruct
```
yentsanglee, line 29: Several "ext_product_id" need to be renamed

host/sgx/create.c
```diff
@@ -589,6 +604,17 @@ oe_result_t oe_sgx_validate_enclave_properties(
         goto done;
     }
 
+    if (!(properties->config.attributes & OE_SGX_FLAGS_KSS))
+    {
+        if (!oe_sgx_is_unset_uuid(
+                (uint8_t*)properties->config.extended_product_id))
```
yentsanglee, line 610: If either `extended_product_id` or `family_id` is not all zeros, we have to return failure besides logging.

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -97,9 +97,37 @@ NumHeapPages=1024
 NumStackPages=1024
 NumTCS=1
 ```
+Additionally, a developer can specify additional Key Sharing and Separation (KSS) identity properties
+for use on the platforms that support it:
+- **FamilyID**: The product family identity (ISVFAMILYID) a developer can specify to group
+different enclaves under a common identity. For example, an identifier for the application suite
+which includes several enclave apps.
+- **ExtendedProductID**: The extended product identity (ISVEXTPRODID) value that a developer
```
dthaler, line 105: Is the "extended product identity" concept really inherently specific to KSS?

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -97,9 +97,37 @@ NumHeapPages=1024
 NumStackPages=1024
 NumTCS=1
 ```
+Additionally, a developer can specify additional Key Sharing and Separation (KSS) identity properties
+for use on the platforms that support it:
+- **FamilyID**: The product family identity (ISVFAMILYID) a developer can specify to group
```
dthaler, line 102: I see terms like "ISVFAMILYID" which look to be SGX-only, or are those names intended to be TEE-agnostic?  I ask because elsewhere in this PR I see "OE_ENCLAVE_FLAG_KSS" (as opposed to "OE_ENCLAVE_FLAG_SGX_KSS") so assume KSS is being defined as a TEE-agnostic concept, which any TEE might or might not support.  If so, then the descriptions of each of these fields should not be SGX specific, but can use SGX in examples, like "(e.g., ISVFAMILYID for SGX)".

host/sgx/sgxsign.c
```diff
@@ -666,6 +685,8 @@ oe_result_t oe_sgx_sign_enclave_from_engine(
     const char* engine_id,
     const char* engine_load_path,
     const char* key_id,
+    const uint8_t* family_id,
+    const uint8_t* ext_product_id,
```
dthaler, line 689:
```suggestion
    const uint8_t* extended_product_id,
```
https://github.com/openenclave/openenclave/blob/master/docs/Community/DevelopmentGuide.md says:

> DON'T use abbreviations unless they are already well-known terms known by users (e.g., "app", "info"), or are already required for use by developers (e.g., "min", "max", "args"). Examples of bad use would be num_widgets instead of widget_count, and opt_widgets instead of option_widgets or optional_widgets.

Expand ext which is ambiguous ("external", "extend", "extended", ...)

host/sgx/sgxsign.c
```diff
@@ -704,6 +731,8 @@ oe_result_t oe_sgx_sign_enclave(
     uint16_t security_version,
     const uint8_t* pem_data,
     size_t pem_size,
+    const uint8_t* family_id,
+    const uint8_t* ext_product_id,
```
dthaler, line 735:
```suggestion
    const uint8_t* extended_product_id,
```

host/sgx/sgxsign.c
```diff
@@ -740,6 +775,8 @@ oe_result_t oe_sgx_get_sigstruct_digest(
     uint64_t attributes,
     uint16_t product_id,
     uint16_t security_version,
+    const uint8_t* family_id,
+    const uint8_t* ext_product_id,
```
dthaler, line 779:
```suggestion
    const uint8_t* extended_product_id,
```

host/sgx/sgxsign.c
```diff
@@ -772,6 +815,8 @@ oe_result_t oe_sgx_digest_sign_enclave(
     size_t cert_pem_size,
     const uint8_t* digest_signature,
     size_t digest_signature_size,
+    const uint8_t* family_id,
+    const uint8_t* ext_product_id,
```
dthaler, line 819:
```suggestion
    const uint8_t* extended_product_id,
```

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -138,6 +153,80 @@ typedef struct _oe_sgx_enclave_properties
...
+ * @param PRODUCT_ID ISV assigned Product ID (ISVPRODID) to use in the
+ * enclave signature
+ * @param SECURITY_VERSION ISV assigned Security Version number (ISVSVN)
+ * to use in the enclave signature
+ * @param ALLOW_DEBUG If true, allows the enclave to be created with
+ * OE_ENCLAVE_FLAG_DEBUG and debugged at runtime
+ * @param HEAP_PAGE_COUNT Number of heap pages to allocate in the enclave
+ * @param STACK_PAGE_COUNT Number of stack pages per thread to reserve in
+ * the enclave
+ * @param TCS_COUNT Number of concurrent threads in an enclave to support
```
dthaler, line 170:
```suggestion
 * @param[in] PRODUCT_ID ISV assigned Product ID (ISVPRODID) to use in the
 * enclave signature
 * @param[in] SECURITY_VERSION ISV assigned Security Version number (ISVSVN)
 * to use in the enclave signature
 * @param[in] ALLOW_DEBUG If true, allows the enclave to be created with
 * OE_ENCLAVE_FLAG_DEBUG and debugged at runtime
 * @param[in] HEAP_PAGE_COUNT Number of heap pages to allocate in the enclave
 * @param[in] STACK_PAGE_COUNT Number of stack pages per thread to reserve in
 * the enclave
 * @param[in] TCS_COUNT Number of concurrent threads in an enclave to support
```

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -138,6 +153,80 @@ typedef struct _oe_sgx_enclave_properties
...
+ * @param EXTENDED_PRODUCT_ID ISV assigned Extended Product ID (ISVEXTPRODID)
+ *  to use in the enclave signature
+ * @param FAMILY_ID ISV assigned Product Family ID (ISVFAMILYID)
+ * to use in the enclave signature
+ * @param ALLOW_DEBUG If true, allows the enclave to be created with
+ * OE_ENCLAVE_FLAG_DEBUG and debugged at runtime
+ * @param HEAP_PAGE_COUNT Number of heap pages to allocate in the enclave
+ * @param STACK_PAGE_COUNT Number of stack pages per thread to reserve in
+ * the enclave
+ * @param TCS_COUNT Number of concurrent threads in an enclave to support
```
dthaler, line 208:
```suggestion
 * @param[in] PRODUCT_ID ISV assigned Product ID (ISVPRODID) to use in the
 * enclave signature
 * @param[in] SECURITY_VERSION ISV assigned Security Version number (ISVSVN)
 * to use in the enclave signature
 * @param[in] EXTENDED_PRODUCT_ID ISV assigned Extended Product ID (ISVEXTPRODID)
 *  to use in the enclave signature
 * @param[in] FAMILY_ID ISV assigned Product Family ID (ISVFAMILYID)
 * to use in the enclave signature
 * @param[in] ALLOW_DEBUG If true, allows the enclave to be created with
 * OE_ENCLAVE_FLAG_DEBUG and debugged at runtime
 * @param[in] HEAP_PAGE_COUNT Number of heap pages to allocate in the enclave
 * @param[in] STACK_PAGE_COUNT Number of stack pages per thread to reserve in
 * the enclave
 * @param[in] TCS_COUNT Number of concurrent threads in an enclave to support
```

tests/tools/oesign/test-enclave/oesign_test.edl
```diff
@@ -12,5 +12,6 @@ enclave {
 
     trusted {
         public bool is_test_signed();
+        public oe_result_t check_kss_extended_ids(oe_uuid_t* family_id, oe_uuid_t* ext_product_id);
```
dthaler, line 15:
```suggestion
        public oe_result_t check_kss_extended_ids(oe_uuid_t* family_id, oe_uuid_t* extended_product_id);
```

tests/tools/oesign/test-sign/CMakeLists.txt
```diff
@@ -203,3 +203,67 @@ add_test(
...
+  COMMAND
+    oesign sign -e $<TARGET_FILE:oesign_test_enc> -c
+    ${OESIGN_TEST_INPUTS_DIR}/ext_prod_id_too_short.conf -k
+    ${OESIGN_TEST_INPUTS_DIR}/sign_key.private.pem)
+
+set_tests_properties(
+  tests/oesign-sign-extprodid-tooshort
+  PROPERTIES PASS_REGULAR_EXPRESSION "bad value for 'ExtendedProductID'")
+
+# Test invalid kss properties argument -ExtProdID too short
```
dthaler, line 247: nit: be consistent with line 233 (hyphen or not)
Also, is the argument -ExtProdID or -ExtendedProductID as implied in 245?

host/sgx/create.c
jianlia: Add it in oe_create_enclave, thanks

include/openenclave/host.h
jianlia: Updated.

include/openenclave/internal/sgxsign.h
jianlia: Renamed

host/sgx/create.c
jianlia: Updated

tests/tools/oesign/test-sign/CMakeLists.txt
jianlia: Remove the extra hyphen

host/sgx/sgxsign.c
jianlia: Updated

host/sgx/sgxsign.c
jianlia: Updated.

host/sgx/sgxsign.c
jianlia: Updated

include/openenclave/bits/sgx/sgxproperties.h
jianlia: Updated.

tests/tools/oesign/test-enclave/oesign_test.edl
jianlia: Updated

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -97,9 +97,37 @@ NumHeapPages=1024
 NumStackPages=1024
 NumTCS=1
 ```
+Additionally, a developer can specify additional Key Sharing and Separation (KSS) identity properties
```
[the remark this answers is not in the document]
jianlia: Updated

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -97,9 +97,37 @@ NumHeapPages=1024
...
+for use on the platforms that support it:
+- **FamilyID**: The product family identity (ISVFAMILYID) a developer can specify to group
+different enclaves under a common identity. For example, an identifier for the application suite
```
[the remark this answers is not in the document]
jianlia: Updated

docs/GettingStartedDocs/buildandsign.md
jianlia: Updated

include/openenclave/bits/sgx/sgxproperties.h
jianlia: Updated

host/sgx/create.c
```diff
@@ -966,6 +1003,15 @@ oe_result_t oe_create_enclave(
     oe_enclave_t* enclave = NULL;
     oe_sgx_load_context_t context;
 
+    if (flags & OE_SGX_FLAGS_KSS && !_is_kss_supported())
```
yentsanglee, line 1006: OE_SGX_FLAGS_KSS is not a valid flags here. See definition OE_ENCLAVE_FLAG_RESERVED.

Please move this checking to oe_sgx_build_enclave() as I commented in iteration 8.

host/sgx/create.c
yentsanglee: Thought about the code a little more, flipping the KSS support bit on _is_kss_supported() is problematic. It's going to generate different enclave measurement. Here's the revised version:

````
    if (props.config.attributes & OE_SGX_FLAGS_KSS)
    {
        if ((context.type == OE_SGX_LOAD_TYPE_CREATE) && !_is_kss_supported())
        {
            // Fail if the CPU does not support KSS and the enclave specifies the KSS flag
            OE_RAISE_MSG(...);
        }

        context->attributes.flags |= OE_ENCLAVE_FLAG_SGX_KSS;
    }
````

Comment (jianlia) [reply host/sgx/create.c]: Cool. Perfect way to use context!  Updated

**Review by yentsanglee — APPROVED** [review]

File: docs/GettingStartedDocs/buildandsign.md (hunk 15 above)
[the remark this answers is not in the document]
Comment (dthaler) [reply docs/GettingStartedDocs/buildandsign.md:100]: I don't see any change.  I see a link in line 108 but it's not a link that defines KSS or says whether it's SGX only or TEE agnostic concept, or says whether it's the same or different from a "security domain".

File: tests/tools/oesign/test-sign/CMakeLists.txt [hunk 18]
```diff
@@ -203,3 +203,67 @@ add_test(
...
+  COMMAND
+    oesign sign -e $<TARGET_FILE:oesign_test_enc> -c
+    ${OESIGN_TEST_INPUTS_DIR}/family_id_too_long.conf -k
+    ${OESIGN_TEST_INPUTS_DIR}/sign_key.private.pem)
+
+set_tests_properties(
+  tests/oesign-sign-familyid-toolong PROPERTIES PASS_REGULAR_EXPRESSION
+                                                "bad value for 'FamilyID'")
+
+# Test invalid kss properties argument -FamilyID too short
```
dthaler, line 259:
```suggestion
# Test invalid kss properties argument FamilyID too short
```

jianlia: Updated. Thanks!

vim.exe.stackdump
```diff
@@ -0,0 +1,19 @@
+Stack trace:
```
dthaler, line 1: Why is this file in the PR?

tools/oesign/oesign.c
```diff
@@ -671,3 +765,64 @@ int oedigest(const char* enclave, const char* conffile, const char* digest_file)
...
+}
+
+int uuid_from_string(str_t* str, uint8_t* uuid, size_t expected_size)
+{
+    int rc = -1;
+    size_t index = 0;
+    size_t size = 0;
+    char* id_copy;
+    char value = 0;
+    bool firstDigit = true;
```
CodeMonkeyLeet, line 792: Nit: use snake_casing for variable names per OE developer guidelines (i.e. `first_digit`)

docs/GettingStartedDocs/buildandsign.md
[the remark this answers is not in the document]
dthaler: There do not seem to be any good docs defining KSS from what I can tell (the Intel manuals don't).   As near as I can tell, KSS is just the ability to share keys across multiple related enclaves, and in that sense the concept is not intel sgx specific, it's a subset of what the GlobalPlatform standard "security domain" provides (see https://globalplatform.org/specs-library/tee-management-framework-including-asn1-profile/ for the detailed spec), where as "security domain" contains the related enclaves, and "store data" within a "security domain" is equivalent to sealing with a family key in SGX KSS.   I could be wrong due to lack of sufficient docs, but that's my reading so far.  As such, we should make sure that OE can support the concept in general, while still mapping to Intel's functionality.   Question is whether we can have agnostic names now, or whether we need to use TEE (i.e., SGX) specific names now and possibly deprecate them later if agnostic names are defined, similar to what we did for report/quote changing to evidence in APIs.

tests/tools/oesign/test-enclave/host/host.c
```diff
@@ -20,12 +40,34 @@ int main(int argc, const char* argv[])
...
+    }
+
+    /* Load the SGX enclave properties */
+    if ((result = oe_sgx_load_enclave_properties(
+             &oeimage, OE_INFO_SECTION_NAME, &properties)) != OE_OK)
+    {
+        oe_put_err("oe_sgx_load_enclave_properties(): result=%u", result);
+    }
 
     if ((result = oe_create_oesign_test_enclave(
```
CodeMonkeyLeet, line 59: Continuing discussion from #3659:
> After I remove `_is_kss_supported` from oesign, test will fail here so I will update the test here like the following flow:
> ```c
> if ((result = oe_create_oesign_test_enclave(
>    argv[1], OE_ENCLAVE_TYPE_AUTO, flags, NULL, 0, &enclave)) != OE_OK)
> {
>     if (!(result == OE_PLATFORM_ERROR && is_kss_supported() && non-zero KSS properties))
>         oe_put_err("oe_create_crypto_enclave(): result=%u", result);
>     else
>     {
>         return 0;
>     }
> }
> ```

1. The proposed change is not reflected in this PR, so how are the tests passing now?
2. The proposed change is incorrect. It will cause all oesign tests, even those that do not rely on KSS properties to skip the enclave creation step, which is one of the necessary validations as part of those tests. 

My recommendation here remains unchanged:
> I think you're going to restructure this test, because once enclaves can be signed on non-KSS machines with KSS properties, the oe_create_enclave on line 45 will fail before it gets to this check, so you're going to have to introduce a mechanism to skip entire tests based on detection of this bit, and a parameter passed into the host to indicate whether a specific test expects KSS or not.
>
> See tests/debug-mode for an example.

To elaborate:
- The test host needs to take a command line argument indicating whether the test is intended to use KSS properties. 
   - This means updating the sign-and-verify.py and the tests CMakeLists.txt to specify and pass through this option to the test host.
   - If the option is provided and KSS is not supported, then the test host will skip the `oe_create_enclave` and `check_kss_extended_ids` verifications. 
      - The host should explicitly print a message like "Skipping creation of enclave using KSS on platform without KSS support..." (see similar message for MRSIGNER check in simulation mode).
      - This message is needed as an alternative success indicator, since all tests will look for the "PASS: Launch of signed enclave succeeded" set as part of `set_test_properties` in the tests' CMakeLists.txt. (i.e. needs to be provided as a second value delimited by `;` for [PASS_REGULAR_EXPRESSION](https://cmake.org/cmake/help/latest/prop_test/PASS_REGULAR_EXPRESSION.html)

1
```diff
@@ -0,0 +1,21 @@
+Merge branch 'icx_kss_2' of https://github.com/jianlia/openenclave into icx_kss_2
```
CodeMonkeyLeet, line 1: This should not be included as part of the PR. Please resolve the merge conflict and remove this file.

CodeMonkeyLeet (changes requested):
Looks mostly good to me, see the comment on updates needed to the oesign tests though.

Also, you need to clean up the commit history of this PR before this will be allowed to merge into master; it still contains replica commits merged from master branch like the oeedger8r changes. I would also advise you to stop spinning new PR requests for the same history and body of work, as it makes it difficult to track the context of review comments across multiple PRs.

I suggest:
- address all the remaining PR comments and add commits as necessary
- once the reviewers are satisfied, refresh your local repo's view of the master branch and do a `git rebase` to master, squashing all of your commits.

tests/tools/oesign/test-enclave/host/host.c
jianlia: The new updated logic is as following:  Raise the exception in oe_sgx_build_enclave when context is OE_SGX_LOAD_TYPE_CREATE and platform doesn't support KSS and attribute has KSS enabled

    if (props.config.attributes & OE_SGX_FLAGS_KSS)
    {
        if ((context->type == OE_SGX_LOAD_TYPE_CREATE) && !_is_kss_supported())
        {
            // Fail if the CPU does not support KSS and the enclave specifies the KSS flag
            OE_RAISE_MSG(
                OE_UNSUPPORTED,
                "Enclave image was signed with kss flag but CPU doesn't support KSS\n",
                NULL);
        }
        context->attributes.flags |= OE_ENCLAVE_FLAG_SGX_KSS;
    }
At test tests\tools\oesign\test-enclave\host\host.c
Handle the OE_UNSUPPORTED error code and skip the test if attributes enable the KSS. The other tests will not be impacted. 
    if ((result = oe_create_oesign_test_enclave(
    {
        if (!is_kss_supported &&
            (result == OE_UNSUPPORTED &&
             properties.config.attributes & OE_SGX_FLAGS_KSS))
        {
            // Skip the test as it is not supported
            return 0;
        }
        else
            oe_put_err("oe_create_crypto_enclave(): result=%u", result);
    }
Result for platform doesn't support :
38: Created /home/acctestuser/openenclave/build/tests/tools/oesign/test-enclave/enclave/oesign_test_enc.signed
38: 2020-11-03T23:20:30+0000.526527Z [(H)ERROR] tid(0x7f07702bbb80) | Enclave image was signed with kss flag but CPU doesn't support KSS
38:  (oe_result_t=OE_UNSUPPORTED) [/home/acctestuser/openenclave/host/sgx/create.c:oe_sgx_build_enclave:879]
38: 2020-11-03T23:20:30+0000.526657Z [(H)ERROR] tid(0x7f07702bbb80) | :OE_UNSUPPORTED [/home/acctestuser/openenclave/host/sgx/create.c:oe_create_enclave:1090]
1/1 Test #38: tests/oesign-sign-valid-kss ......   Passed    0.05 sec

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -97,9 +97,40 @@ NumHeapPages=1024
...
+which includes several enclave apps.
+- **ExtendedProductID**: The extended product identity (ISVEXTPRODID for SGX) value that a developer
+can use as 128-bit globally unique identifier for the enclave where the 16-bit ProductID proves too restrictive.
+For more details, see Table 37-19 of the
+[Intel's Software Developers Manual](https://software.intel.com/content/www/us/en/develop/articles/intel-sdm.html#combined/).
+
+For example, in kss_valid.conf:
+```
+FamilyID=47183823-2574-4bfd-b411-99ed177d3e43
+ExtendedProductID=47183823-2574-4bfd-b411-99ed177d3e43
```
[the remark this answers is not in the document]
jianlia: Updated to new GUID

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -138,6 +153,80 @@ typedef struct _oe_sgx_enclave_properties
...
+ #define OE_SET_ENCLAVE_SGX_KSS(                                          \
+    PRODUCT_ID,                                                           \
+    SECURITY_VERSION,                                                     \
+    EXTENDED_PRODUCT_ID,                                                  \
+    FAMILY_ID,                                                            \
+    ALLOW_DEBUG,                                                          \
+    HEAP_PAGE_COUNT,                                                      \
+    STACK_PAGE_COUNT,                                                     \
+    TCS_COUNT)                                                            \
+ _OE_SET_ENCLAVE_SGX_IMPL                                                 \
```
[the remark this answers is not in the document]
jianlia: Add the missing bracket

tools/oesign/oesign.c
jianlia: Updated

tools/oesgx/oesgx.c
```diff
@@ -131,6 +132,22 @@ int main(int argc, const char* argv[])
...
+
+        result = _CPUID(&regs);
+        if (result)
+        {
+            printf("Read SGX_ATTRIBUTE_ENUMERATION failed:\n");
+            dump_regs(&regs);
+            return result;
+        }
+
+        printf("CPU supports KSS: %s\n", HAVE_KSS(regs) ? "true" : "false");
```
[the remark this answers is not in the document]
jianlia: Updated oesgx.c per comment.

tests/tools/oesign/test-enclave/host/host.c
CodeMonkeyLeet: Thanks for the clarification, I agree that it works, but it still creates misleading test logs. 

- sign-and-verify.py still expects that the enclave launch passes when it invokes the test host, and so continues to report with a fixed message that enclave launch succeeded, when that's not the case.
- the script believes that is actually the case because the test host just eats the error and reports that the test passed completely with no logging that it skipped the test. 

I would suggest at least fixing the following:
- Update the default success message for `launch_cmd` in sign-and-verify.py to reflect the new meaning of a successful return from test host (e.g. "Signed enclave test app succeeded")
   - This does mean updating the CMakeList.txt files still, but it's a straightfoward global find and replace now
- Update test host to `printf` a message indicating that a KSS signed enclave load was skipped as unsupported by the platform (i.e. the comment on line 66 is no enough, we want that information to be indicated in the test results)

docs/GettingStartedDocs/buildandsign.md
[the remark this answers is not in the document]
CodeMonkeyLeet: @dthaler No, KSS is not analogous to the functionality of Security Domains insofa; you should think of KSS as an opt-in extension of the existing SGX identity model by adding a couple of ID properties that will be included in the TEE-specific evidence. This includes:

- Two new sign-time authored identity values: FAMILYID and ISVEXTPRODID
- A new runtime identity value: CONFIGID, which is security versioned by CONFIGSVN.

If you squint at it enough, the comparison you're making kind of works, but really what this is about is providing a bigger identity namespace for SGX enclaves, and that doesn't intersect with the concept of permissions or key storage or association of keys to enclaves in a way that might relate to Security Domains and OTrP in GlobalPlatform.

All this does is allow a dev more options in providing differentiated identities for apps in a taxonomy of their choosing (which may not be hierarchical) Because this also translates into more ways to define an SGX sealing key based on which elements of the enclave identity are used in its derivation, it provides more flexibility for a dev to share/partition sealing keys, hence the name.

This example being SGX-specific I think is perfectly reasonable, and I don't believe this should be expected to generalize across TEEs, just as OP-TEE and SGX already do not share common identity metadata formats (i.e. OP-TEE basically only consisting of the TA's UUID)

tests/tools/oesign/test-enclave/host/host.c
jianlia: Thanks. Updated per comments (update the default success message and test host to add printf message. )

yentsanglee:
bors try

host/sgx/create.c
bodzhang: SECS.ATTRIBUTES is not part of the measurement, so slipping the OE_ENCLAVE_FLAG_SGX_KSS bit in `context->attributes.flags` should not cause problem with enclave measurement. But the code update in @yentsanglee suggestion matches the design better and is required for a different reason - when performing enclave signing, KSS support on the underlining system should not be required. (my assumption is that the code flow is used for both enclave loading and enclave signing).

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -84,24 +88,33 @@ typedef struct _oe_sgx_enclave_properties
...
 // clang-format off
 
-#define OE_SET_ENCLAVE_SGX(                                               \
+#define _OE_SET_ENCLAVE_SGX_IMPL(                                         \
     PRODUCT_ID,                                                           \
     SECURITY_VERSION,                                                     \
+    EXTENDED_PRODUCT_ID,                                                  \
+    FAMILY_ID,                                                            \
     ALLOW_DEBUG,                                                          \
+    ALLOW_KSS,                                                            \
```
bodzhang, line 117: By the current design, this flag is "REQUIRE_KSS" instead of "ALLOW_KSS". When this flag is set, if the underlining system does not support KSS, enclave loading will fail. Therefore, KSS is "REQUIRED". In future update, with CONFIGID/CONFIGSVN support, a developer who does not use non-zero FAMILY_ID.EXTENDED_PRODUCT_ID might implement an enclave that utilizes CONFIG_ID/CONFIG_SVN feature when KSS is available, but still functions without CONFIG_ID/CONFIG_SVN. In that case, the developer should indicate to the enclave loader that KSS is "DESIRED". I believe defining the flag as "REQUIRE_KSS" still works in that scenario. The zero value on this flag means "NOT_REQUIRED", no conflict with "DESIRED".

CodeMonkeyLeet (approved):
Thanks for all your hard work on this! LGTM

jianlia:
bors try

yentsanglee:
bors try

jianlia:
bors try

jianlia:
bors try

dthaler (commented):
See Bo's comment from 3 hours ago which, if correct, should be addressed before merging.  I don't see any other blockers so marking as Approve but please respond to Bo first.

dthaler (approved):
The lack of documentation about KSS is also unfortunate but not a blocker.  It just means that most people won't (or can't) use it, but that may be fine if the point is just to provide functionality parity with the Intel SDK.

include/openenclave/bits/sgx/sgxproperties.h
jianlia: Updated to REQUIRE_KSS, thanks

1
jianlia: Updated

yentsanglee:
bors retry

jianlia:
bors try

BRMcLaren:
bors try

jianlia:
bors try

PR #3704, opened 2020-11-03
Use newer OIDs for quote and evidence generation

### LINKED ISSUE #3703: Stop generating attestation certificates with old report OID
Per agreement with external teams, stop generating attestation certificates with old OIDs "1.2.840.113556.10.1.1" and "1.2.840.113556.10.1.2". 

New OIDs are "1.3.6.1.4.1.311.105.1" and "1.3.6.1.4.1.311.105.2".

ryanhsu19:
~~Remove old OIDs, and assign new OIDs to the variables that previously saved old OIDs.~~

Stop using old OIDs to generate attestation certificates.

Fix #3703 .

### DIFF
--- a/enclave/tls_cert.c
+++ b/enclave/tls_cert.c
@@ -25,8 +25,8 @@
 #define DATE_NOT_VALID_BEFORE "20190501000000"
 #define DATE_NOT_VALID_AFTER "20501231235959"
 
-static const unsigned char oid_oe_report[] = X509_OID_FOR_QUOTE_EXT;
-static const unsigned char oid_oe_evidence[] = X509_OID_FOR_OE_EVIDENCE_EXT;
+static const unsigned char oid_oe_report[] = X509_OID_FOR_NEW_QUOTE_EXT;
+static const unsigned char oid_oe_evidence[] = X509_OID_FOR_NEW_OE_EVIDENCE_EXT;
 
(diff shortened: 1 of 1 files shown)


yentsanglee:
The verifier should still recognize the old and new OIDs. Just change the evidence/certificate generation to use the new OIDs.

dthaler (approved):

dthaler:
If I understand right, this is introducing a change whereby older evidence (using the old OID) will no longer be able to be verified.   This should be called out in the changelog for the next release.

ryanhsu19:
After double-checking with @yentsanglee, I confirmed that the first step is to stop generating X509 certificates with the old OIDs, but end of supporting verification of those generated with the old ones will be the next step, hence the force-pushed change.

dthaler:
> After double-checking with @yentsanglee, I confirmed that the first step is to stop generating X509 certificates with the old OIDs, but end of supporting verification of those generated with the old ones will be the next step, hence the force-pushed change.

Thanks, I updated the PR title accordingly.

dthaler (approved):

ryanhsu19:
bors try

yentsanglee (approved):

ryanhsu19:
Seems to be a timeout error.
bors retry

PR #3721, opened 2020-11-09
Cleanup ABI state preservation on enclave transitions

CodeMonkeyLeet:
1. Standardize ABI state preservation between ECalls & OCalls
   - OE SDK preserves FXSAVE & RFLAGS state on the host side for ECalls, but not OCalls, which this commit corrects.
    - ECalls now rely on register preservation provided by the compiler for required XMM registers on Windows, and explicitly preserves MXCSR and FCW instead of using FXSAVE.
    - OCalls now also preserve MXCSR, FCW and RFLAGS state as part of the callsite context during oe_ocall for parity with ECalls.
    - Move enclave-specific context.inc/h files under the /enclave path

2. Clear AVX registers on ECall to avoid [AVX-SSE transition penalty](https://software.intel.com/content/www/us/en/develop/articles/avoiding-avx-sse-transition-penalties.html)

3. Clean up testing related to MXCSR and FXSAVE state:
    - Add ABI test for enclave transition functions `oe_enter` and `oe_ocall`
    - Fix tests/libc to reset FXSAVE state between libc tests.
      - The original libc tests for floating point math were compiled as individual executables and assume ABI-initialized floating point and MXCSR state. Since the enclave runs multiple tests in the same enclave consecutively, we reset the FXSAVE state on each run.
      - The test enclave was previously relying on an OCall (printf) before each each test that unintentionally reset FXSAVE state in the enclave prior to the introduction of MXCSR preservation on OCalls.
      - Also remove the host-side logging of MXCSR, which serves no current test purpose.
    - Remove tautological XFRM checks from tests/props
      - The additional checks for XFRM configuration on enclave creation added to tests/props only ends up verifying that a test shim for `oe_get_xfrm()` is called, and not that the enclave properties are set accordingly.


CodeMonkeyLeet:
bors try

dthaler (approved):
Looks ok to me but the build is hitting errors like this one:

```
../tests/abi/host/../abi_utils.h:166:23: error: always_inline function might not be inlinable [-Werror=attributes]
 OE_ALWAYS_INLINE bool is_same_abi_state(abi_state_t* a, abi_state_t* b)
                       ^
```

CodeMonkeyLeet:
bors try

enclave/core/sgx/calls.c
```diff
@@ -484,6 +484,17 @@ OE_INLINE void _handle_oret(
...
+    /* Restore the FXSTATE and flags */
+    asm volatile("pushq %[rflags] \n\t" // Restore flags.
+                 "popfq \n\t"
+                 "fldcw %[fcw] \n\t"     // Restore x87 control word
+                 "ldmxcsr %[mxcsr] \n\t" // Restore MXCSR
+                 : [mxcsr] "=m"(callsite->mxcsr),
+                   [fcw] "=m"(callsite->fcw),
+                   [rflags] "=m"(callsite->rflags)
+                 :
+                 :);
```
anakrish, line 496: Since flags register is being changed in the assembly snippet, "flags" or "cc" ought to be added to the clobber list for correctness though it may not affect the generated code in this case.

host/sgx/enter.c
```diff
@@ -44,14 +45,18 @@
...
-// Only rbp and rsp are preserved.
-#define OE_ENCLU_CLOBBERED_REGISTERS \
-    "r8", "r9", "r10", "r11", "r12", "r13", "r14", "r15"
+// Only rbp and rsp are preserved on return from ENCLU.
+// The XMM registers are listed as clobbered to signal to the compiler that
+// they need to be preserved when --target=x86_64-msvc-windows and are
+// ignored on Linux builds.
+#define OE_ENCLU_CLOBBERED_REGISTERS                                      \
+    "r8", "r9", "r10", "r11", "r12", "r13", "r14", "r15", "xmm6", "xmm7", \
+        "xmm8", "xmm9", "xmm10", "xmm11", "xmm12", "xmm13", "xmm14", "xmm15"
```
anakrish, line 59: Should we conservatively list "ymm" registers instead of "xmm" registers?

host/sgx/enter.c
```diff
@@ -156,14 +159,21 @@ void oe_enter(
...
+
     oe_ecall_context_t ecall_context = {{0}};
     _setup_ecall_context(&ecall_context);
 
+    // Compiler will usually handle this on exiting a function that uses AVX,
+    // but we need to avoid the AVX-SSE transition penalty here manually as
+    // part of the transition to enclave. See https://software.intel.com/content
+    // /www/us/en/develop/articles/avoiding-avx-sse-transition-penalties.html
+    if (oe_is_avx_enabled)
+        asm volatile("vzeroupper \n\t" : : :);
```
anakrish, line 175: Since vzeroupper wipes out the top 128 bits of the ymm registers, we'd need to specify the ymm registers in the clobber list to be correct.

host/sgx/enter.c
```diff
@@ -156,14 +159,21 @@ void oe_enter(
...
+    uint16_t fcw = 0;
+
     oe_ecall_context_t ecall_context = {{0}};
     _setup_ecall_context(&ecall_context);
 
+    // Compiler will usually handle this on exiting a function that uses AVX,
+    // but we need to avoid the AVX-SSE transition penalty here manually as
+    // part of the transition to enclave. See https://software.intel.com/content
+    // /www/us/en/develop/articles/avoiding-avx-sse-transition-penalties.html
+    if (oe_is_avx_enabled)
```
anakrish, line 174: Do we enforce that when oe_enter is called, oe_is_avx_enabled would have been correctly initialized? I see that we initialize it using oe_once. But I'm curious about the case where enclaves and called via global initializers.

anakrish (commented):
Thanks for doing this!
I have some questions. Looks good otherwise.

host/sgx/enter.c
CodeMonkeyLeet: No need, the Windows x64 ABI specifically states that:

> The x64 ABI considers the registers RAX, RCX, RDX, R8, R9, R10, R11, and XMM0-XMM5 volatile. When present, the upper portions of YMM0-YMM15 and ZMM0-ZMM15 are also volatile. On AVX512VL, the ZMM, YMM, and XMM registers 16-31 are also volatile.

This is also reflected in the standard prolog/epilog generated by the compiler on Windows.

enclave/core/sgx/calls.c
CodeMonkeyLeet: It doesn't seem to generate different assembly, but I added "cc" to the clobber list.

host/sgx/enter.c
CodeMonkeyLeet: Similar to the flags, this doesn't seem to change the generated assembly, but I've added the YMM registers to the clobber list.

host/sgx/enter.c
CodeMonkeyLeet: I'm not sure I understand the scenario you have in mind using global initializers. `oe_is_avx_enabled` is globally initialized for the host process once on first invocation of any `oe_create_enclave` call, which is a prerequisite for calling an enclave method and `oe_enter`. Prior to this changeset, `oe_create_enclave` already had a dependency on `oe_once`, and this just reuses that mechanism for that additional initialization of `oe_is_avx_enabled`, so any interactions with the timing of global initialization already exists.

CodeMonkeyLeet:
bors try

host/sgx/enter.c
anakrish: Given that `oe_is_avx_enabled` is initialized as part of `oe_create_enclave` and not via a global constructor in which case ordering of global constructors would matter, the scenario I was curios about is addressed.

host/sgx/enter.c
```diff
@@ -156,14 +159,39 @@ void oe_enter(
...
-    uint64_t fx_state[64];
+    // Additional control registers that need to be preserved as part of the
+    // Windows and Linux x64 ABIs
+    uint32_t mxcsr = 0;
+    uint16_t fcw = 0;
+
     oe_ecall_context_t ecall_context = {{0}};
     _setup_ecall_context(&ecall_context);
 
+    // Compiler will usually handle this on exiting a function that uses AVX,
```
anakrish, line 170: Thanks for the nice description. In simulation mode too the jump to the enclave code happens unbeknownst to the compiler. It would be good to refactor this into a separate function and call that in oe_enter_sim as well.

host/sgx/enter.c
```diff
@@ -156,14 +159,39 @@ void oe_enter(
...
+    // but we need to avoid the AVX-SSE transition penalty here manually as
+    // part of the transition to enclave. See https://software.intel.com/content
+    // /www/us/en/develop/articles/avoiding-avx-sse-transition-penalties.html
+    if (oe_is_avx_enabled)
+        asm volatile("vzeroupper \n\t"
```
anakrish, line 175: vzeroupper seems best placed within the while loop below.
Otherwise it won't be called when returning from an OCALL.

host/sgx/enter.c
CodeMonkeyLeet: Thanks for the catch, fixed.

host/sgx/enter.c
CodeMonkeyLeet: Fixed.

anakrish (approved):
LGTM.

Now that vzeroupper is so close to the assembly snippet for ENCLU, it is quite tempting to move vzeroupper into the ENCLU assembly template. That is however not a stopper and can be attempted later.

CodeMonkeyLeet:
@anakrish 
> Now that vzeroupper is so close to the assembly snippet for ENCLU, it is quite tempting to move vzeroupper into the ENCLU assembly template. That is however not a stopper and can be attempted later.

I would resist that temptation since unlike the ENCLU call, the call to VZEROUPPER is contingent on the presence of AVX features being enabled by the OS, and moving the conditional `JNZ` based on `oe_is_avx_enabled` into the ENCLU assembly block reduces the readability of that logic.

BRMcLaren:
bors try

PR #3723, opened 2020-11-09
Expose sgx report body fields by oe claims

### LINKED ISSUE #3689: Request to expose additional SGX report fields as claims
Summary:
•	All potentially-useful bits and fields in an SGX report should be exposed
•	Some of the missing fields are new with KSS/IceLake, but not all.

Here’s the definition of the SGX report structure, for IceLake onwards, with the fields where we need OE to emit new claims marked with >>:

```

    >> sgx_cpu_svn_t           cpu_svn;        /* (  0) Security Version of the CPU */
    >> sgx_misc_select_t       misc_select;    /* ( 16) Which fields defined in SSA.MISC */
    uint8_t                 reserved1[SGX_REPORT_BODY_RESERVED1_BYTES];  /* ( 20) */
    >> sgx_isvext_prod_id_t    isv_ext_prod_id;/* ( 32) ISV assigned Extended Product ID */
    >> sgx_attributes_t        attributes;     /* ( 48) Any special Capabilities the Enclave possess */**
    sgx_measurement_t       mr_enclave;     /* ( 64) The value of the enclave's ENCLAVE measurement */
    uint8_t                 reserved2[SGX_REPORT_BODY_RESERVED2_BYTES];  /* ( 96) */
    sgx_measurement_t       mr_signer;      /* (128) The value of the enclave's SIGNER measurement */
    uint8_t                 reserved3[SGX_REPORT_BODY_RESERVED3_BYTES];  /* (160) */
    >> sgx_config_id_t         config_id;      /* (192) CONFIGID */
    >> sgx_prod_id_t           isv_prod_id;    /* (256) Product ID of the Enclave */
    >> sgx_isv_svn_t           isv_svn;        /* (258) Security Version of the Enclave */
    >> sgx_config_svn_t        config_svn;     /* (260) CONFIGSVN */
    uint8_t                 reserved4[SGX_REPORT_BODY_RESERVED4_BYTES];  /* (262) */
    >> sgx_isvfamily_id_t      isv_family_id;  /* (304) ISV assigned Family ID */
    sgx_report_data_t       report_data;    /* (320) Data provided by the user */

```
Of these, two are bitfields: misc_select and attributes. The rest are uninterpreted integers or binary blobs:

cpu_svn is 16 bytes
isv_ext_prod_id is 16 bytes
config_id is 64 bytes
isv_prod_id is 16 bits
isv_svn is 16 bits
config_svn is 16 bits
isv_family_id is 16 bytes

misc_select is 32 bits, but only one of those has a defined meaning: extended exit info enabled (0x1). There should be a claim for that.

attributes is 128 bits, of which 64 are enclave flags and 64 are the xsave feature mask. I don’t think we should try to keep up with xsave feature bits, so those can probably just be an uninterpreted integer, but the flag bits are more significant:

#define SGX_FLAGS_INITTED        0x0000000000000001ULL     /* If set, then the enclave is initialized */
#define SGX_FLAGS_DEBUG          0x0000000000000002ULL     /* If set, then the enclave is debug */
#define SGX_FLAGS_MODE64BIT      0x0000000000000004ULL     /* If set, then the enclave is 64 bit */
#define SGX_FLAGS_PROVISION_KEY  0x0000000000000010ULL     /* If set, then the enclave has access to provision key */
#define SGX_FLAGS_EINITTOKEN_KEY 0x0000000000000020ULL     /* If set, then the enclave has access to EINITTOKEN key */
#define SGX_FLAGS_KSS            0x0000000000000080ULL     /* If set enclave uses KSS */

Inited is always 1, so we can ignore it. The others should all get a claim (KSS is new on IceLake).

qiucwang:
Fixes #3689 
Added two more fields in `sgx_report_body` struct that consistent with intel [SGX report structure](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/common/inc/internal/linux/arch.h#L55-L77)
```
uint8_t configid[64];    
uint16_t configsvn;
```
Added oe_claim entries to expose `sgx_report_body` fields. Flag fields as `misc_select` and `attributes` are converted to multiple booleans.

### DIFF
--- a/common/sgx/verifier.c
+++ b/common/sgx/verifier.c
@@ -226,24 +226,34 @@ static oe_result_t _fill_with_known_claims(
     oe_result_t result = OE_UNEXPECTED;
     oe_report_t parsed_report = {0};
     oe_identity_t* id = &parsed_report.identity;
+    const sgx_report_body_t* sgx_report_body = NULL;
     size_t claims_index = 0;
     oe_datetime_t valid_from = {0};
     oe_datetime_t valid_until = {0};
+    bool flag;
 
-    if (claims_length < OE_REQUIRED_CLAIMS_COUNT)
+    if (claims_length < OE_REQUIRED_CLAIMS_COUNT + OE_SGX_REQUIRED_CLAIMS_COUNT)
         OE_RAISE(OE_INVALID_PARAMETER);
 
     if (format_type == SGX_FORMAT_TYPE_LOCAL)
-        OE_CHECK(oe_parse_sgx_report_body(
-            &((sgx_report_t*)report_body)->body, false, &parsed_report));
+    {
+        sgx_report_body = &((sgx_report_t*)report_body)->body;
+        OE_CHECK(
+            oe_parse_sgx_report_body(sgx_report_body, false, &parsed_report));
+    }
+
     else
-        OE_CHECK(oe_parse_sgx_report_body(
-            &((sgx_quote_t*)report_body)->report_body, true, &parsed_report));
+    {
+        sgx_report_body = &((sgx_quote_t*)report_body)->report_body;
+        OE_CHECK(
+            oe_parse_sgx_report_body(sgx_report_body, true, &parsed_report));
+    }
 
     // Optional claims are needed for SGX quotes for remote attestation
     if (format_type != SGX_FORMAT_TYPE_LOCAL &&
         claims_length < OE_REQUIRED_CLAIMS_COUNT + OE_OPTIONAL_CLAIMS_COUNT +
-                            OE_SGX_CLAIMS_COUNT)
+                            OE_SGX_REQUIRED_CLAIMS_COUNT +
+                            OE_SGX_OPTIONAL_CLAIMS_COUNT)
         OE_RAISE(OE_INVALID_PARAMETER);
 
     // ID version.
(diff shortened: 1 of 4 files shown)


common/sgx/verifier.c
```diff
@@ -302,6 +312,91 @@ static oe_result_t _fill_with_known_claims(
...
+    flag = !!(sgx_report_body->attributes.flags & SGX_FLAGS_EINITTOKEN_KEY);
+    OE_CHECK(_add_claim(
+        &claims[claims_index++],
+        OE_CLAIM_SGX_HAS_EINITTOKEN_KEY,
+        sizeof(OE_CLAIM_SGX_HAS_EINITTOKEN_KEY),
+        &flag,
+        sizeof(bool)));
+
+    // SGX Use KSS or Not
+    flag = false;
```
yentsanglee, line 368: This should be the SGX_FLAGS_KSS bit.

tests/attestation_plugin/plugin/tests.c
```diff
@@ -790,9 +790,11 @@ void verify_sgx_evidence(
     // Check SGX endorsements related claims:
```
yentsanglee, line 790: Check the 10 new SGX claims here

common/sgx/verifier.c
qiucwang: Thanks!

tests/attestation_plugin/plugin/tests.c
qiucwang: Thanks!

yentsanglee (approved):

qiucwang:
bors try

ryanhsu19 (approved):
Great work!

ryanhsu19:
bors retry

common/sgx/verifier.c
```diff
@@ -302,6 +312,91 @@ static oe_result_t _fill_with_known_claims(
         format_id,
         sizeof(*format_id)));
 
+    // SGX Report CPU SVN
```
bodzhang, line 315: CPUSVN in SGX `QUOTE`/`REPORT` is a "hint" only. Security analysis should not be made based on the CPUSVN value reported by the SGX enclave SW, including the SGX PCE/QE. The SGX CPU's security property is endorsed by the PCK Cert Chain. The leaf cert of the PCK Cert Chain includes SVNs of CPU firmware/microcode modules that are inside SGX TCB.  Successful `QUOTE` verification, using the PCK Cert Chain, indicates the underlining SGX CPU's security property is equal to or higher than the SVNs in the PCK Cert Chain leaf Cert. The SVNs in the PCK Cert Chain leaf Cert might be associated with a SGX TCB that's out-of-date, ie, with known vulnerability. In that case, it's possible that the enclave SW that produced the SGX `QUOTE` has been compromised and might have spoofed the CPUSVN value in the SGX `QUOTE`. Only if the leaf Cert indicates the SGX CPU is free of any known security vulnerability, should the CPUSVN value in SGX `QUOTE` be trusted.

CPUSVN value in SGX `QUOTE`/`REPORT` is an opaque value. Two different value of CPUSVN can not be numerically compared for TCB level comparison purpose. 

Therefore, exposing the CPUSVN value in SGX `QUOTE` as a claim almost has no value. It should not be included in the outputted claims.

common/sgx/verifier.c
```diff
@@ -302,6 +312,91 @@ static oe_result_t _fill_with_known_claims(
...
+        OE_CLAIM_SGX_CPU_SVN,
+        sizeof(OE_CLAIM_SGX_CPU_SVN),
+        sgx_report_body->cpusvn,
+        sizeof(sgx_report_body->cpusvn)));
+
+    // SGX Exit Info Reported to an SSA Frame or Not
+    flag = !!(sgx_report_body->miscselect & SGX_MISC_FLAGS_EXIT_INFO);
+    OE_CHECK(_add_claim(
+        &claims[claims_index++],
+        OE_CLAIM_SGX_EXIT_INFO_ENABLED,
```
bodzhang, line 327: The claim name is not informative. The specific bit in miscselect indicates whether page fault or general protection exception occurred inside the SGX enclave is reported to the enclave SW. It's possible that future SGX version might add new bits also related to EXINFO. As the purpose of translating the SGX miscselect bitmap to SGX-specific claims is to simplify the upper level SW developer's work, it's better to make the claim names informative. Suggest to rename the claim as `OE_CLAIM_SGX_PF_GP_EXINFO_ENABLED`.

bodzhang (changes requested):

common/sgx/verifier.c
qiucwang: Thanks for the naming.
Renamed it.

common/sgx/verifier.c
qiucwang: Hided CPUSVN from `QUOTE` as we should use CPUSVN from the leaf certificate.

yentsanglee:
Looks good to me.

yentsanglee (approved):

yentsanglee:
bors retry

common/sgx/verifier.c
bodzhang: Looks good.

bodzhang (approved):

qiucwang:
bors retry

ryanhsu19 (approved):
Thanks for the great work!

qiucwang:
bors try

qiucwang:
bors try

qiucwang:
bors retry

PR #3724, opened 2020-11-10
Enforce attestation header in all evidence

yentsanglee:
To ensure evidence generated by OE SDK has the same header universally. With the header in all evidence, we can 
1.	Apply version control and maintain backward compatibility.
2.	Extend evidence format (e.g. for proposal to include inittime claims in the evidence buffer).
3.	Allow oe_verify_evidence() to detect format without format input. Attestation service may not want to deal with multiple OE evidence format IDs.

Changes:
1. Deprecate flag OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID. Attestation header with format ID is enforced.
2. oe_verify_evidence() with NULL format input verifies reports as well as all evidence formats.


dthaler:
If I understand this proposal (and maybe I don't), then I would object to it.
We need to be able to generate standard evidence (in addition to legacy evidence) and standard evidence (e.g., IETF, TCG, etc.) has no OE header.  Thus "To ensure evidence generated by OE SDK has the same header universally" is not only a non-goal, it's a requirement to NOT be the case.

shnwc:
> If I understand this proposal (and maybe I don't), then I would object to it.
> We need to be able to generate standard evidence (in addition to legacy evidence) and standard evidence (e.g., IETF, TCG, etc.) has no OE header. Thus "To ensure evidence generated by OE SDK has the same header universally" is not only a non-goal, it's a requirement to NOT be the case.

Good point. Do those standard evidence have their own header fields that allow oe_verify_evidence() to determine their format without needing a format ID input parameter?

dthaler:
> > If I understand this proposal (and maybe I don't), then I would object to it.
> > We need to be able to generate standard evidence (in addition to legacy evidence) and standard evidence (e.g., IETF, TCG, etc.) has no OE header. Thus "To ensure evidence generated by OE SDK has the same header universally" is not only a non-goal, it's a requirement to NOT be the case.
> 
> Good point. Do those standard evidence have their own header fields that allow oe_verify_evidence() to determine their format without needing a format ID input parameter?

We had this discussion before in SIG-Attestation.  Answer is no, you can never guess based on the contents of the buffer, you always need a separate discriminator like a UUID.  In many protocols, this is often a Media Type value, but could be anything else depending on the protocol carrying the evidence.

dthaler (changes requested):
This goes against previous SIG-Attestation agreements

yentsanglee:
Will make the change to call oe_verify_evidence() with NULL format input verifies reports as well as all evidence formats.

Other change requests were discussed and objected in SIG-Attestation meeting.

shnwc:
@yentsanglee will you create a new PR for the proposed change to `oe_verify_evidence()`?

yentsanglee:
The PR is #3732

PR #3741, opened 2020-11-18
Syscall TCB Reduction/Syscall Dispatching Rewrite

anakrish:
Syscalls are routed directly to their implementations rather than via a switch-case.

- Reduced TCB
  Switch-case causes all the syscalls to be retained in an enclave.
  Routing directly allows the linker to eliminate those syscalls that are not used by
  an enclave.
  For example echo enclave reports only the following syscalls:
  ❯ objdump -t echo_enc | grep oe_SYS_
  000000000003002b l     F .text  00000000000000bc oe_SYS_clock_gettime_impl
  00000000000300e7 l     F .text  00000000000000c6 oe_SYS_gettimeofday_impl
  0000000000067e10 l     F .text  000000000000005f oe_SYS_futex_impl

- Syscall callgraphs can be analyzed easily.
  Since syscall implementations are directly called rather than via a switch-case,
  the static callgraphs are simpler and more accurate.
  ❯ callgraph.py echo_enc oe_SYS_gettimeofday_impl
 oe_SYS_gettimeofday_impl 0x300e7 /home/anakrish/work/syscall/libc/syscalls.c:75
  └── __clock_gettime 0x67d97 /home/anakrish/work/syscall/3rdparty/musl/musl/src/time/clock_gettime.c:25
      └── time 0x67d3f /home/anakrish/work/syscall/3rdparty/musl/musl/src/time/time.c:4
          └── x509_get_current_time 0x38ff8 /home/anakrish/work/syscall/3rdparty/mbedtls/mbedtls/library/x509.c:929
              ├── mbedtls_x509_time_is_past 0x3927a /home/anakrish/work/syscall/3rdparty/mbedtls/mbedtls/library/x509.c:994
              │   ├── x509_crt_verifycrl 0x36d69 /home/anakrish/work/syscall/3rdparty/mbedtls/mbedtls/library/x509_crt.c:1832
              │   │   └── x509_crt_verify_chain 0x37491 /home/anakrish/work/syscall/3rdparty/mbedtls/mbedtls/library/x509_crt.c:2288
              │   │       └── mbedtls_x509_crt_verify_restartable 0x37a3f /home/anakrish/work/syscall/3rdparty/mbedtls/mbedtls/library/x509_crt.c:2562
              │   │           └── mbedtls_x509_crt_verify 0x379e4 /home/anakrish/work/syscall/3rdparty/mbedtls/mbedtls/library

- It is possible to figure out the list of syscalls retained in an enclave via:
  `objdump -t enclave-filename | grep oe_SYS_`

Details

- Syscalls are most often performed by libc (in our case MUSL) rather than the user code.
  A syscall in MUSL of the form
      syscall(SYS_open, a, b)
  is converted to
      oe_SYS_open_impl(a, b)
  avoiding the dispatch via switch-case.
  This is accomplished by patching musl/src/internal/syscall.h.

  Note: OE currently (prior to this PR) relies on SYSCALL_NO_INLINE define to get MUSL to
  not use SYSCALL instruction and instead call `syscall` function for dispatching syscalls.
  SYSCALL_NO_INLINE macro has been removed in newer versions of MUSL and MUSL no longer
  provides the above customization that we want. Therefore we'd anyways have to patch
  musl/src/internal/syscall.h when we upgrade MUSL.

- The existing switch-case based dispatching have been retained for
   - use by libOS. It may turn out that libOS does not need the switch-case based
     dispatching at which point we can remove it.
   - the case where user code directly makes a syscall instead relying of MUSL.
     Such cases ought to be quite rare.

- Syscall implementations are declared WEAK so that they can be overridden.
  This is intended for use by tests and not by users.
  This provides a (cleaner IMO) alternative to syscall hooks.
  The syscall function names have oe_SYS_ prefix and the chance of user accidentally
  overriding a syscall implementation is quite low. The chance can be further minimized
  by obfuscating the prefix.

- N_M (N or More) syscalls.
  Some syscalls take N or more parameters. Such syscalls are declared and defined using
  the N_M macros.
  E.g:
      OE_DEFINE_SYSCALL2_M(Sys_open)
  Sys_open is called with atleast 2 parameters; but sometimes with 3. There is no easy
  way to capture this in C. It C++ it is possible to provide overloads and capture all
  the calling patterns. As a reasonable alternative, we declare SYS_open to take 2 or
  more parameters.

  Tracking such syscalls will allows us in future to make sure that OE implementations
  of syscalls do indeed consume all the supplied parameters.

  MUSL sometimes calls certain syscalls with syscall_cp macro which always supplies
  6 parameters to the syscall - the original supplied parameters followed by zeros.
  It is not clear why MUSL does this. But this has the consequence that
  syscalls that are called thus must be declared using the N_M macros.

### DIFF
--- a/3rdparty/musl/CMakeLists.txt
+++ b/3rdparty/musl/CMakeLists.txt
@@ -6,9 +6,9 @@
 
 if (OE_SGX)
   set(ARCH "x86_64")
-else()
+else ()
   set(ARCH "aarch64")
-endif()
+endif ()
 
 # NOTE: MUSL is NOT built by the `ExternalProject_Add` command. These
 # C flags only apply to a configuration step that generates the MUSL
@@ -31,7 +31,7 @@ set(MUSL_DIR ${CMAKE_CURRENT_BINARY_DIR}/musl)
 set(MUSL_INCLUDES ${OE_INCDIR}/openenclave/libc)
 
 set(MUSL_APPEND_DEPRECATIONS
-  "${CMAKE_CURRENT_LIST_DIR}/append-deprecations ${MUSL_INCLUDES}")
+    "${CMAKE_CURRENT_LIST_DIR}/append-deprecations ${MUSL_INCLUDES}")
 if (USE_CLANGW)
   set(MUSL_CFLAGS "-target x86_64-pc-linux ${MUSL_CFLAGS}")
   set(MUSL_CC clang)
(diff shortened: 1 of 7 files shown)


anakrish:
bors try

anakrish:
bors try-

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

3rdparty/musl/CMakeLists.txt
```diff
@@ -31,80 +31,67 @@ set(MUSL_DIR ${CMAKE_CURRENT_BINARY_DIR}/musl)
...
+  COMMAND ${CMAKE_COMMAND} -E copy ${MUSL_DIR}/arch/${ARCH}/bits/syscall.h.in
+          ${MUSL_INCLUDES}/bits/syscall.h
+  # bash -c requires the command string to be in a single line
+  COMMAND
+    ${OE_BASH} -c
+    "sed -n -e s/__NR_/SYS_/p < ${MUSL_DIR}/arch/${ARCH}/bits/syscall.h.in >> ${MUSL_INCLUDES}/bits/syscall.h"
+  COMMAND ${CMAKE_COMMAND} -E copy ${MUSL_INCLUDES}/endian.h
+          ${MUSL_INCLUDES}/__endian.h
+  COMMAND ${CMAKE_COMMAND} -E copy ${PATCHES_DIR}/endian.h
+          ${MUSL_INCLUDES}/endian.h
```
dthaler, line 84: I don't understand why copies like this are needed at all.   Can't we just put ${PATCHES_DIR} in the include path before ${MUSL_INCLUDES} and avoid all this busy work?

include/openenclave/internal/syscall/declarations.h
```diff
@@ -96,85 +158,108 @@ OE_DECLARE_SYSCALL1(SYS_epoll_create);
...
 OE_DECLARE_SYSCALL0(SYS_getpid);
 OE_DECLARE_SYSCALL0(SYS_getppid);
-OE_DECLARE_SYSCALL3(SYS_getsockname);
-OE_DECLARE_SYSCALL5(SYS_getsockopt);
+OE_DECLARE_SYSCALL3_M(SYS_getsockname);
+OE_DECLARE_SYSCALL5_M(SYS_getsockopt);
+OE_DECLARE_SYSCALL2(SYS_gettimeofday);
 OE_DECLARE_SYSCALL0(SYS_getuid);
-OE_DECLARE_SYSCALL6(SYS_ioctl);
+// SYS_iocl is called with 3 or more args.
```
dthaler, line 200:
```suggestion
// SYS_ioctl is called with 3 or more args.
```

libc/syscalls.c
```diff
@@ -29,22 +30,25 @@ static const uint64_t _SEC_TO_MSEC = 1000UL;
...
 {
-    clockid_t clk_id = (clockid_t)x1;
-    struct timespec* tp = (struct timespec*)x2;
+    /* Always fail */
+    return EPERM;
+}
+
+OE_DEFINE_SYSCALL2(SYS_clock_gettime)
+{
+    clockid_t clk_id = (clockid_t)arg1;
```
dthaler, line 47:
```suggestion
    clockid_t clock_id = (clockid_t)arg1;
```
Avoid abbreviations, especially since the type spells it out

tests/crypto/enclave/enc/enc.c
```diff
@@ -37,149 +39,158 @@ char* oe_host_strdup(const char* str)
...
+{
+    /* Handle SYS_readv because fread invokes readv internally
+     * To avoid dealing with linux-specific readv semantics on Windows,
+     * marshal this as a synchronous C read() invocation.
+     */
+
+    struct iovec* iov = (struct iovec*)arg2;
+
+    // determine the total buffer size
+    size_t buf_size = sizeof(struct iovec) * (size_t)arg3;
```
dthaler, line 102:
```suggestion
    size_t buffer_size = sizeof(struct iovec) * (size_t)arg3;
```

tests/crypto/enclave/enc/enc.c
```diff
@@ -37,149 +39,158 @@ char* oe_host_strdup(const char* str)
...
     }
+    buf_size += data_size;
 
-    if (!ret)
+    // create the local buffer
+    struct iovec* iov_host = (struct iovec*)malloc(buf_size);
+    char* data_pos = (char*)iov_host + sizeof(struct iovec) * (size_t)arg3;
+
+    // initialize the buffers
+    char* buf_pos = data_pos;
```
dthaler, line 115:
```suggestion
    char* buffer_position = data_position;
```

tests/mbed/enc/enc.c
```diff
@@ -65,102 +66,137 @@ void test_checker(char* str)
...
+    return -1;
+}
+
+OE_DEFINE_SYSCALL3_M(SYS_read)
+{
+    errno = 0;
+    int ret = -1;
+    ssize_t rval = 0;
+    const size_t buf_len = (size_t)arg3;
+    char* host_buf = (char*)oe_host_malloc(buf_len);
```
dthaler, line 95:
```suggestion
    const size_t buffer_len = (size_t)arg3;
    char* host_buffer = (char*)oe_host_malloc(buffer_len);
```

syscall/syscall.c
```diff
@@ -267,23 +274,57 @@ OE_DEFINE_SYSCALL2(SYS_flock)
...
 }
 
-OE_DEFINE_SYSCALL1(SYS_fsync)
+OE_WEAK OE_DEFINE_SYSCALL4(SYS_fstatat)
+{
+    oe_errno = 0;
+    long ret = -1;
+    int dirfd = (int)arg1;
+    const char* pathname = (const char*)arg2;
+    struct oe_stat_t* buf = (struct oe_stat_t*)arg3;
```
dthaler, line 291:
```suggestion
    struct oe_stat_t* buffer = (struct oe_stat_t*)arg3;
```
Although some name closer to "stat" would be more clear

syscall/syscall.c
```diff
@@ -666,7 +734,18 @@ OE_DEFINE_SYSCALL4(SYS_pread64)
     return oe_pread(fd, buf, count, offset);
 }
 
-OE_DEFINE_SYSCALL5(SYS_pselect6)
+OE_WEAK OE_DEFINE_SYSCALL4(SYS_pread64)
+{
+    oe_errno = 0;
+    const int fd = (int)arg1;
+    void* const buf = (void*)arg2;
```
dthaler, line 741:
```suggestion
    void* const buffer = (void*)arg2;
```

syscall/syscall.c
```diff
@@ -687,7 +766,18 @@ OE_DEFINE_SYSCALL5(SYS_pselect6)
     return oe_select(nfds, readfds, writefds, exceptfds, tv);
 }
 
-OE_DEFINE_SYSCALL4(SYS_pwrite64)
+OE_WEAK OE_DEFINE_SYSCALL4_M(SYS_pwrite)
+{
+    oe_errno = 0;
+    const int fd = (int)arg1;
+    const void* const buf = (void*)arg2;
```
dthaler, line 773:
```suggestion
    const void* const buffer = (void*)arg2;
```

radhikaj (approved):
Thanks for looking into this and improving the experience

anakrish:
bors try

mingweishih (approved):
Looks good to me. Great work!

3rdparty/musl/CMakeLists.txt
anakrish: After the patches have been applied by copying, we run MUSL's _configure_ command which depends on files being in certain locations. Therefore, we have to apply the patches as we do now.
Post _configure_, we further copy files to the relevant _output_ folders for use by other modules as well as for packaging for installation. These too require files to be in specific places.

Note: MUSL itself does not come with cmake support and relies on linux specific commands to build. We have implemented this bit of CMake scripting to integrate it into OE's cmake based build system. This has the additional benefit that MUSL based enclaves can also be built on Windows.

3rdparty/musl/patches/syscall.h
```diff
@@ -0,0 +1,138 @@
...
+    __syscall_ret(OE_SYSCALL_NAME(_##index)(SYSCALL_ARGS(__VA_ARGS__)))
+
+#define socketcall __socketcall
+#define socketcall_cp __socketcall_cp
+
+#define SYSCALL_CP_ARGS_X(a, b, c, d, e, f, ...) SYSCALL_ARGS6(a, b, c, d, e, f)
+#define SYSCALL_CP_ARGS(...) SYSCALL_CP_ARGS_X(__VA_ARGS__, 0, 0, 0, 0, 0, 0)
+
+#define __syscall_cp(index, ...) \
+    OE_SYSCALL_NAME(_##index)(SYSCALL_CP_ARGS(__VA_ARGS__))
```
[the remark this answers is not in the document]
anakrish: Actually, the entire file has a lot of changes. Our version of the file is simpler than the original.
It is best to view the differences via a diff tool.

anakrish:
bors try

achamayou (approved):

Issue #3729, opened 2020-11-16
Highly auditble and reproducible build environment for openenclave using NIX should be added


yakman2020:
NIX is a package manager/build system which incorporates the principles of transparency and reproducibility at a very basic level. The result is that it is possible to provable trace the resulting .deb file through the dependent libraries and tools to the end result. This greatly increases the level of confidence a client project can have in the resulting package.

I propose to wrap the openenclave  build process in a nix expression, using a docker container specified by sha, to build the standard openenclave project in an isolated environment. A single command line argument would be enough to  cause the project and all of its dependencies to be built from known sources, or to build openenclave with cached dependencies in pretty much the usual time. 

NIX is an excellent system but it has some limitations.  I am not proposing that NIX become the mainline build process for production. 

- NIX supports linux only, and is unlikely to ever support Windows.
- IMO NIX package repos are not at industrial levels of quality and reliability.  To use the build in production would require someone to manage a curated package repo. The design of NIX facilitates that, but it is work that would be required for production.
-  Nix does support aarch64, but my submission would not be vetted on that platform.

With agreement in principle I can submit a detailed design doc.

CodeMonkeyLeet:
@yakman2020 Yes, please submit a design doc for this and also submit an additional issue to track how this proposal would be integrated with CI/CD. Does not have to be in v0.13, adding this to the v0.14 milestone which can be revisited in the next sprint.

@radhikaj FYI for SIG-Release tracking.

PR #3744, opened 2020-11-20
optional reproducible build process for openenclave on linux

yakman2020:
Present a nix based reproducible build environment for openenclave-sdk. 

Addresses issue #3729 for linux. Does not address windows, as that will require a completely different set of tools, plus changes in the Visual C environment.

Design document is located in docs/DesignDocs/ReproducibleBuildUsingNix.md


reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+If any changes are made to that package, any package related to the changed package will need to be custom
+built and cached in nix store.
+
+
+## A Nix build system for OpenEnclave SDK
+
+We propose to add an optional nix based build environment for OpenEnclave SDK using docker containers.
+
+This would consist of :
+- Dockerfile.nix and build-oe-nix-build.sh for building the oe nix docker image locally.
```
anakrish, line 232: I get `libsgx_enclave_common.so: no such file or directory` upon running `build-oe-nix-image.sh`:

```bash
$ sudo ./build-oe-nix-image.sh
...
...
+ nix-instantiate -I. -E '(import <nixpkgs> {}).dpkg'
warning: you did not specify '--add-root'; the result might be removed by the garbage collector
/nix/store/dfdvxn42glnb9xgy8yrkd70gjnvhx6y0-dpkg-1.19.7.drv
+ nix-env -I. -i /nix/store/dfdvxn42glnb9xgy8yrkd70gjnvhx6y0-dpkg-1.19.7.drv
installing 'dpkg-1.19.7.drv'
these paths will be fetched (9.35 MiB download, 58.04 MiB unpacked):
  /nix/store/sg42gk32rbx5b3klmw9ql2sy165rnq8v-perl-5.30.1
  /nix/store/xic28d0cmcxv08wmh539xqq09vnajdar-dpkg-1.19.7
copying path '/nix/store/sg42gk32rbx5b3klmw9ql2sy165rnq8v-perl-5.30.1' from 'https://cache.nixos.org'...
copying path '/nix/store/xic28d0cmcxv08wmh539xqq09vnajdar-dpkg-1.19.7' from 'https://cache.nixos.org'...
building '/nix/store/9lc3laqsk26d3c95d38l2cfb716npa2j-user-environment.drv'...
created 742 symlinks in user environment
+ nix-env -I. -i /nix/store/6a6ilqysgz1gwfs0ahriw94q35vj84sy-vim-8.2.1123
installing 'vim-8.2.1123'
warning: unable to download 'https://cache.nixos.org/6a6ilqysgz1gwfs0ahriw94q35vj84sy.narinfo': Couldn't resolve host name (6); retrying in 260 ms
these paths will be fetched (14.82 MiB download, 70.98 MiB unpacked):
  /nix/store/0hcbvd7wpsx0p4vjvn5b2jb04rf35mkc-gawk-5.1.0
  /nix/store/2pi6zgkwnr3zdslvlv16nixpzvbyjx1n-glibc-2.31
  /nix/store/6a6ilqysgz1gwfs0ahriw94q35vj84sy-vim-8.2.1123
  /nix/store/j206pskg2yzyla1cnfrb9kb5n5bfvjgm-bash-4.4-p23
  /nix/store/mphihm7gn8yb1svf8zgyn9qvl86zhy6m-ncurses-6.2
  /nix/store/y7ya11l8qy5z2zyhc8db7km7kykkl1sw-libunistring-0.9.10
  /nix/store/zzp30ib8rgm42d80gw3jl5cy5vgvaq9y-libidn2-2.3.0
copying path '/nix/store/y7ya11l8qy5z2zyhc8db7km7kykkl1sw-libunistring-0.9.10' from 'https://cache.nixos.org'...
copying path '/nix/store/zzp30ib8rgm42d80gw3jl5cy5vgvaq9y-libidn2-2.3.0' from 'https://cache.nixos.org'...
copying path '/nix/store/2pi6zgkwnr3zdslvlv16nixpzvbyjx1n-glibc-2.31' from 'https://cache.nixos.org'...
copying path '/nix/store/j206pskg2yzyla1cnfrb9kb5n5bfvjgm-bash-4.4-p23' from 'https://cache.nixos.org'...
copying path '/nix/store/0hcbvd7wpsx0p4vjvn5b2jb04rf35mkc-gawk-5.1.0' from 'https://cache.nixos.org'...
copying path '/nix/store/mphihm7gn8yb1svf8zgyn9qvl86zhy6m-ncurses-6.2' from 'https://cache.nixos.org'...
copying path '/nix/store/6a6ilqysgz1gwfs0ahriw94q35vj84sy-vim-8.2.1123' from 'https://cache.nixos.org'...
building '/nix/store/j7icczmyfwb65nh6a0l7zha9j68dfysc-user-environment.drv'...
created 782 symlinks in user environment
+ nix-store -I. --verify --check-contents
error: no arguments expected
Try 'nix-store --help' for more information.
+ set +x
Removing intermediate container 482622a5e9af
 ---> de39efe358c0
Step 29/38 : ADD ./sort_deb_sum.sh /home/$BUILD_USER
 ---> 20b6a44fa5f3
Step 30/38 : ADD ./nix-build.sh /home/$BUILD_USER
 ---> e7a6fb1f6d1e
Step 31/38 : ADD ./nix-shell.sh /home/$BUILD_USER
 ---> c640511c64ac
Step 32/38 : ADD ./build_deb_pkg.sh /home/$BUILD_USER
 ---> 6c457a9e8e6a
Step 33/38 : RUN mkdir -p /home/$BUILD_USER/.nix_libs
 ---> Running in f84b6e911dc9
Removing intermediate container f84b6e911dc9
 ---> 86b48a4e1e37
Step 34/38 : ADD ./nix-libs/libsgx_enclave_common.so /home/$BUILD_USER/.nix_libs
ADD failed: stat /var/lib/docker/tmp/docker-builder901887648/nix-libs/libsgx_enclave_common.so: no such file or directory
...

```

yakman2020: Investigating. I wonder if git has an issue with binaries

mingweishih: Can we avoid checking in binaries?

yakman2020: Just in case I'll put them into a tar file and copy from there.

mingweishih: Ideally, we should host the tar file somewhere else (i.e., azure storage) and just use a script to download it.

bodzhang: Is libsgx_enclave_common.so still needed to build the OE SDK binaries? Run time dependency on Intel SGX PSW components makes sense. Why do we have build time dependency?

anakrish: We don't need libsgx-enclave-common.so or any other sgx libraries to build the SDK.

reproducible/prep-nix-build.sh
```diff
@@ -0,0 +1,127 @@
...
+# and can instead run tests separately. This is desirable as many build machines don't support
+# SGX, or at least don't have it enabled. 
+#
+# prep-nix-build.sh $nixpkgs_dest
+#
+#
+# First clone the nixpkgs repo. With a little discipline we can 
+# keep all package accesses on this machine.
+#
+export NIX_PKGS_REPO=https://github.com/yakman2020/nixpkgs.git
```
bodzhang, line 16: https://github.com/yakman2020/nixpkgs.git acc-test2 branch is thousands of commits ahead of NixOS:master. What's the critical difference between https://github.com/yakman2020/nixpkgs.git acc-test2 and NixOS:master? How does the difference affect the reproducible build process?

reproducible/prep-nix-build.sh
```diff
@@ -0,0 +1,127 @@
...
+    nix-instantiate -I. -E '(import <nixpkgs> {}).dpkg'
+    nix-env -I. -i /nix/store/dfdvxn42glnb9xgy8yrkd70gjnvhx6y0-dpkg-1.19.7.drv
+
+    # needed for nix-shell debug
+    nix-env -I. -i /nix/store/z976i71y86b771hz14k62j5x8cifnppf-vim-8.2.1522.drv
+else
+    nix-instantiate -I. -E '(import <nixpkgs> {}).cmake'
+    nix-env -I. -i /nix/store/p34s8yj3ws6289xa30z4gf7yy5933g4l-cmake-3.16.3.drv
+    nix-instantiate -I. -E '(import <nixpkgs> {}).openssl'
+    nix-env -I. -i /nix/store/mddaj30wwqyfg3pvid3q7zmhsr3mjnp4-openssl-1.1.1g.drv!bin
```
bodzhang, line 100: This line looks different from other lines. It has "!bin" after "...drv". What's the importance?

reproducible/prep-nix-build.sh
```diff
@@ -0,0 +1,127 @@
...
+    nix-instantiate -I. -E '(import <nixpkgs> {}).gnumake'
+    nix-env -I. -i /nix/store/34k6saz5kl8j0945b28s00kxhyx3y10p-gnumake-4.2.1.drv
+    nix-instantiate -I. -E '(import <nixpkgs> {}).binutils'
+    nix-env -I. -i /nix/store/k2kiqgagj53mq9hv6m0rx9dihsdpyd0w-binutils-wrapper-2.31.1.drv
+    nix-env -I. --set-flag priority 5 binutils-2.31.1
```
bodzhang, line 105: What does "priority" impact?

reproducible/Dockerfile.nix
```diff
@@ -0,0 +1,185 @@
...
+
+
+
+RUN echo "User is $USER "
+#install the required software
+#RUN touch .bash_profile \
+#
+# We add the nix install and packages into the container rather than waiting for run time.
+# The packages are then located in the nix store until the next push of the container
+#RUN curl https://nixos.org/releases/nix/nix-2.3.7/install | /bin/bash
```
bodzhang, line 169: Is nix-2.3.7 installed or this line is really commented out?

reproducible/prep-nix-build.sh
```diff
@@ -0,0 +1,127 @@
...
+    nix-env -I. -i /nix/store/gbhhc8mxvw8hxx2k8r73jj5brbi7hk62-doxygen-1.8.17.drv
+    nix-instantiate -I. -E '(import <nixpkgs> {}).dpkg'
+    nix-env -I. -i /nix/store/dfdvxn42glnb9xgy8yrkd70gjnvhx6y0-dpkg-1.19.7.drv
+
+    # needed for nix-shell debug
+    nix-env -I. -i /nix/store/6a6ilqysgz1gwfs0ahriw94q35vj84sy-vim-8.2.1123 
+fi
+
+#
+# Make sure we don't go to cache by checking everything
```
bodzhang, line 124: Does this mean all the packages listed above are not pulled from https://cache.nixos.org, but instead rebuilt locally? If yes, what's the rationale to not use https://cache.nixos.org?

reproducible/nix-build.sh
```diff
@@ -0,0 +1,26 @@
...
+nix-build -I. shell.nix --substituters 'https://cache.nixos.org' \
+	    --argstr REV $BUILD_REV \
+	    --argstr SHA $BUILD_SHA \
+	    --arg DO_CHECK $DO_CHECK ${OE_SIM}
+#
+# We have built the nix environment, we need to export it 
+# so we can use it in other projects. 
+nix-store --dump-db >/output/oe-nix-store-$(date +%Y%j%H%M).nixdump
+
+nix-shell -I. shell.nix --substituters 'https://cache.nixos.org' \
```
bodzhang, line 19: What does this step do, beyond what's done on line 10?

docs/DesignDocs/ReproducibleBuildUsingNix.md
```diff
@@ -0,0 +1,397 @@
...
+
+- creates a nix expression for building openenclave via (shell.nix).
+
+- installs nix as the build user
+
+- calls the prep-nix-build.sh script which sets up the git clone of nix pkgs and preinstalls the package
+  derivations needed for building openenclave into the nix store. The search process for dependencies in
+  nix is to refer to the nix store first, before looking at nixpkgs or contacting the cache if enabled. While
+  we cannot select package versions from the expression, but must depend on the nixpkgs repo to provide the
+  correct versions, we can prime the nix store to provide the exact derivation we wish before referring to the repo.
```
bodzhang, line 345: Intel SGX SDK [reproducible build](https://github.com/intel/linux-sgx/blob/544159a94a59547a18b7bff1d28a94bb14053572/linux/reproducibility/Dockerfile) seems to relying on specific version of packages in nixpkgs repo/cache directly. What's the benefit of the "prime the nix store" process?

docs/DesignDocs/ReproducibleBuildUsingNix.md
```diff
@@ -0,0 +1,397 @@
...
+This process can be integrated with ADO or CPDX CI, but that is not part of this submission. We do not currently
+recommend depending exclusively on a nix environment for production packaging and CI.
+
+### Dockerfile.nix
+
+The dockerfile is passed up to four build args:
+
+ - BASE_IMAGE allows for different os base images, in particular it allows the substitution of a
+   CPDX base image rather than the generic ubuntu 20.04. As was previously mentioned, we are basing on 
+   ubuntu 20.04 because package versions align best with nixpkgs release-20.03.
```
bodzhang, line 318: Does this mean the OE SDK binaries produced might not work on Ubuntu 18.04 as the glibc version dynamically linked to the produced OE SDK binaries is v2.32, higher than the v2.27 available on Ubuntu 18.04?

reproducible/prep-nix-build.sh
yakman2020: So long as the branch is the same, the output is the same.

reproducible/prep-nix-build.sh
yakman2020: Subpackage.

reproducible/prep-nix-build.sh
yakman2020: Resolves where to get the dependency from. llvm_7 and clang_7 both contain overlapping libraries and binaries.

reproducible/Dockerfile.nix
yakman2020: 2.3.7 is installed, but rather than blindly getting a script from the net, we include it in the git repo.

reproducible/prep-nix-build.sh
yakman2020: The caching path goes local, substituter, build.  cache.nixos.org is the default substituter.
Only the derivations are built locally. From there, they are cached at cache.nixos.org.

reproducible/nix-build.sh
yakman2020: We use patchelf to remove the nix-oriented rpath and ld interpreter.  When we do that it makes it difficult to build further projects using the nix store.  We save the nix-store before beginning the patchelf process.

We do the ctests, patchelf and packaging because the very severe visibility restrictions in nix-build won't allow ctests to run.

docs/DesignDocs/ReproducibleBuildUsingNix.md
yakman2020: We lock the .drvs into the nix store in the docker build process. This preserves the relationship between the package and its dependencies, as well as preventing garbage collection of the drvs until the correct packages are installed.

docs/DesignDocs/ReproducibleBuildUsingNix.md
yakman2020: nixpkgs release-20.03 uses glibc v2.30, which is compatible with ubuntu but not ubuntu 18.04 which uses 2.27.

I tried overriding glibc version and was successful, but it meant removing all the patches the nixpkgs expressison contained, and eliminating all caching since glibc is a dependency for every package in the repo.  19.09 did contain an old enough glibc for ubuntu 18.04 but it is end of life and no longer supported. I was warned away from using it. Also, it contained a number of packages (openssl etc) that were very old.  I don't quite know why the major gap, but there it is.

This supports my impression that we need to manage our own fork of nixpkgs if we are to use this in production.

yakman2020: Quick background to clarify.  glibc is designed no to dynalink with platforms that have an older version of glibc than the client application. So a client that has been linked with glibc 2.27 will work transparently on a platform that has glibc 2.30, but a client that is linked with 2.30 will not work on a platform that has glibc 2.27.  It is not practical to upgrade the glibc for the platform.

The nixpkgs repo is  designed to implement a quasi-distro, nixos, which is based on the nix system.  The packages in the 
nixpkgs repo are intended to serve that purpose.  It is not a non goal that nixpkgs supports building for other distros, but the priority is ambiguous.  The straightforward fix is to build a fork of nixpkgs which contains the expressions for a set of vetted packages, but that would be a large job to do right, and an ongoing commitment of resources.

docs/DesignDocs/ReproducibleBuildUsingNix.md
yakman2020: Note the nix-env which retrieves or builds the build outputs from the .drv uses the expected sha digest of the packages, which will alert us if something else is coming up.

reproducible/README.md
yakman2020: Not to build the sdk, but to run ctests.

yakman2020: It would be good not to restrict visibility for now.  Its easy to retarget later.

yakman2020: FWIW, if I have the tar file in the docker build directory, docker unpacks the contents. If I get it from https:// or whatever, I have to do it. Not an issue, the main question is where to find a CDN that is both open to everybody and secure.

reproducible/Dockerfile.nix
```diff
@@ -0,0 +1,185 @@
...
+RUN /bin/bash ./prep-nix-build.sh /home/$BUILD_USER/nixpkgs
+
+ADD ./sort_deb_sum.sh /home/$BUILD_USER
+ADD ./nix-build.sh /home/$BUILD_USER
+ADD ./nix-shell.sh /home/$BUILD_USER
+ADD ./build_deb_pkg.sh /home/$BUILD_USER
+RUN mkdir -p /home/$BUILD_USER/.nix_libs
+ADD ./nix-libs/libsgx_enclave_common.so /home/$BUILD_USER/.nix_libs
+ADD ./nix-libs/libsgx_enclave_common.so.1 /home/$BUILD_USER/.nix_libs
+ADD ./nix-libs/libsgx_launch.so.1  /home/$BUILD_USER/.nix_libs
```
bodzhang, line 183: As [this feedback](https://github.com/openenclave/openenclave/pull/3744/files#r527924046) shows, libsgx_enclave_common is not required for OE SDK build. libsgx_launch, libprotobuf probably are not needed either. libstdc++ dependency is also unexpected.

yakman2020: They are required to run tests within the container.

They are not required for the build proper, but running ctests is a requirement if we are to use this in a pipeline.
nix restricts access to os libraries, so it may seem strange, but a binary running even in nix shell cannot see the native os libraries.

docs/DesignDocs/ReproducibleBuildUsingNix.md
```diff
@@ -0,0 +1,397 @@
...
+  modified by the addition of apt packages.
+
+The resulting image is named openenclave-build. It can (and should) be generated once and pushed into a common 
+repo.
+
+### build_openenclave.sh
+
+Runs the image produced by docker build, with arguments resolved.
+The output of the process is a .deb and a .nar file containing an exported nix store for use in developing 
+openenclave sdk apps using nix.
```
bodzhang, line 371: Is the instruction to use the OE SDK .deb and other tool chain to build an OE App captured in this design doc or some other doc?

reproducible/Dockerfile.nix
bodzhang: OK, I got it. This line is commented out and should be removed. The installation is handled in "install-nix.sh"

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+- build_openenclave.sh for building the oe deb package via a nix environment.
+- Dockerfile.test and test-oe-nix-build.sh for verifying that package is functional.
+
+The process builds a reproducible .deb package of openenclave-sdk. 
+
+This process can be integrated with ADO or CPDX CI, but that is not part of this submission. We do not currently
```
BRMcLaren, line 238: CDPx is a Microsoft thing, a better phrasing would be that it can be integrated with all CI build engines that support containers.

reproducible/README.md
BRMcLaren: @anakrish @bodzhang responding to you both above. I am wondering what the testing strategy here is. I do have some thoughts. The package itself once built in the Nix environment could be exported and installed on the host and then verified with the sample installation. The alternative would be to build in nix, run in nix (ctest) and then export. The problem I foresee with the later methodology however is that the psw/dcap libs could be out of date, and if they are, attestation will not work due to not being able to generate a valid quote. 

What are your thoughts on a valid testing strategy, this can be brought up with SIG-Testing if we want to dig into it a little more offline but the concerns stem from a CI and Release standpoint as that is the end result here.

docs/DesignDocs/ReproducibleBuildUsingNix.md
bodzhang: @yakman2020 , what does the term "packages" in  "This preserves the relationship between the package and its dependencies, as well as preventing garbage collection of the drvs until the correct packages are installed."  refer to? Does it refer to `nix-build` output, as the "qsgajw2yhrf33qgibsvd9y8zaw64w7j7-helloWorld" in the example? If that's the case, once the pairing between "znmj0axqf4v5mjwswd9bq3dvxf0jcsah-test-environment.drv" and "qsgajw2yhrf33qgibsvd9y8zaw64w7j7-helloWorld" is established and recorded for audit purpose, can the docker image directly use the "packages" in nixpkgs repo/cache and bypass the "prime the nix store" process?

docs/DesignDocs/ReproducibleBuildUsingNix.md
yakman2020: Both a directory readme.md and the end of the design doc.

yakman2020:
The docker image will directly use the packages in /nix/store. But they will definitely be the packages we think they are, or we will see an admittedly nonfatal error in the image build process.   I can make that error fatal if we think its a good idea.

nixpkgs branches get updated and backported, and once they do the resulting derviation and package shas are changed.  We would like to know this.

yakman2020:
But if we get the build products directly rather than the derivation, there will be no audit trail at all. The build products don't contain the derivation or a pointer to it.

reproducible/README.md
bodzhang: @BRMcLaren , I also prefer to see the NIX build process limited to the build itself, and the ctest can be run on the build output. In fact, I wonder whether it's just luck that only a handful of the Intel PSW/DCAP .so files need to be included to run the current ctest. I would assume many other files are required to run any test that involves SGX attestation flow.

reproducible/prep-nix-build.sh
bodzhang: @yakman2020 , what's the difference between https://github.com/yakman2020/nixpkgs.git acc-test2 and NixOS:master? If NixOS:master is used instead, what happens?

reproducible/prep-nix-build.sh
bodzhang: @yakman2020 , can you explain more? How was the specific "priority" value or dependency chosen?

reproducible/nix-build.sh
bodzhang: "difficult to build further projects using the nix store" - what are the "further projects"? The OE test apps run in the ctest stage?

docs/DesignDocs/ReproducibleBuildUsingNix.md
bodzhang: https://github.com/intel/linux-sgx/blob/544159a94a59547a18b7bff1d28a94bb14053572/linux/reproducibility/Dockerfile refers to a "/nix/store/681354n3k44r8z90m35hm8945vsp95h1-glibc-2.27". Is the problem with directly using the glibc-2.27 package that no pairing between the package and a .drv file can be established?

reproducible/README.md
BRMcLaren: @bodzhang I think we should be specific in 'build output' here, I believe we are both talking about the .deb produced, is something like this what you are imagining?

image

bodzhang: @BRMcLaren , based on this [comment](https://github.com/openenclave/openenclave/pull/3744#discussion_r527970744), ctest runs before the .deb is produced. So If the intermediate "build output" can be used to run ctest outside the container, the test requirement is met. Of course, changing the build-ctest process a bit to output .deb and then run ctest works too.

docs/DesignDocs/ReproducibleBuildUsingNix.md
bodzhang: I only found "Work Flow" section at the end of the document. My understand is that that section only talks about building the OE SDK, not about how a OE SDK user makes a reproducible build of an OE App. What I was referring to is an equivalent of https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/buildandsign.md, but with instruction to make a reproducible build. Probably also need to use a docker container with specific clang/gcc NIX package?

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+```
+will build the nix docker image. This can be pushed:
+```
+   docker push openenclave-build
+```
+and later pulled, or just used locally.
+
+To build the openenclave sdk 
+```
+   ./build_openenclave.sh -o ./result
```
vtikoo, line 316:
```suggestion
   ./build_openenclave.sh -o result
```
As the output directory is being used as a docker volume mount, the leading `.` would cause docker to freak out.

Seeing error similar to -
```
$ docker run -it -v ./result:/tmp/ hello-world
docker: Error response from daemon: create ./result: "./result" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+sha of the resulting package is specified, it will compare the .deb sha25sum to the specified value.
+At the end of the process the deb package will not contain any run time evidence of coming from 
+a nix environment. 
+
+Make sure your user is either in sudoers or the destination directory exists or you will
+be asked for a password.
+
+It is possible to manually execute nix-build within the container, in order to debug build problems.
+```
+./build_openenclave.sh -o ./result -i
```
vtikoo, line 328:
```suggestion
./build_openenclave.sh -o result -i
```
Same as above.

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+```
+   docker push openenclave-build
+```
+and later pulled, or just used locally.
+
+To build the openenclave sdk 
+```
+   ./build_openenclave.sh -o ./result
+```
+will build the oe deb and nix-store and place it in ./result. By default it will run ctests. If the
```
vtikoo, line 318: At the end of the build, couldn't find the debian package in `./result` on the host or in `/output` in the container. 
The debian package is created under the nix-store in the container, but probably isn't copied to `/output` ?

```
azureuser@62548e384ec9:~$ find / -type f -name "open-enclave_0.12.0_amd64.deb" 
/nix/store/0kjcfn87x6ai26nk8s7scq8wc5mqikcg-openenclave-sdk/open-enclave_0.12.0_amd64.deb
/nix/store/0kjcfn87x6ai26nk8s7scq8wc5mqikcg-openenclave-sdk/_CPack_Packages/Linux/DEB/open-enclave_0.12.0_amd64.deb
```

bodzhang: @yakman2020, can you clarify the expectation on the location of the Debian package?

yakman2020: The package goes into /output in the container, or the directory specified by -o on the host.

reproducible/README.md
yakman2020: Confirmed. Fix in a minute. I've been using ~/result without trouble, but that appears to be an absolute path.

yakman2020: fixed. Thank you.

reproducible/README.md
yakman2020: fixed

reproducible/README.md
yakman2020: brcamp@nixbed:~/Projects/openenclave/reproducible$ ls ~/results 
build  open-enclave_0.12.0_amd64.deb

yakman2020: Try 
```
git reset --hard HEAD~1
git pull
```

I've been force pushing into the PR commit.

yakman2020: And , by the way, thank you for your feedback.

reproducible/nix-build.sh
bodzhang: "use patchelf to remove the nix-oriented rpath and ld interpreter" discussion should be included included in the design doc or readme.

reproducible/README.md
BRMcLaren: @bodzhang [currently](https://github.com/openenclave/openenclave/blob/master/.jenkins/pipelines/Azure/Linux/Jenkinsfile#L144-L168) is what we do. Build - > Test - > Package - > Test.

Awhile back we did just Build - > Test but ended up running into packaging issues, since then we always have stuck with the model of testing the package as well. Considering the end product of this is to have a highly auditable package it might be best to also do "changing the build-ctest process a bit to output .deb and then run ctest works too." up front.

reproducible/prep-nix-build.sh
yakman2020: NixOS master doesn't work.  Everything after release-20.03 (which acc-test2 is based on) is based on glibc-2.32 which is only compatible with ubuntu 20.10 and fedora 33. 

Everything before 20.03 is EOL.

reproducible/README.md
yakman2020: The Ctests are based on the cmake heirarchy, which in turn is oriented to the nix store paths.  I'm unaware of any process we have in place that can get around that.   Do we have a system for just installing the deb and running the unit tests on it?

reproducible/prep-nix-build.sh
yakman2020: The value is selected manually.  In fact I used the value intel used for the same purpose.
llvm_7 and clang_7 (and many other package pairs) contain overlapping binaries. nix-env requires that in order to install,users specify which package they prefer to use when there is a conflict. Without it, the install fails..

reproducible/Dockerfile.nix
yakman2020: line removed

reproducible/nix-build.sh
yakman2020: I beleive it is.  I will reread and ensure it is sufficiently prominent.

docs/DesignDocs/ReproducibleBuildUsingNix.md
yakman2020: nix always uses any realised build output in nix store if it has them.  nix-env given a drv gets them from the prebuilt cache/channel.  But ther eis no auditability past that.  The drv points at the build output, the build output does not point at the drv.  A lone build output in nix is just as anonymous as a deb package.

docs/DesignDocs/ReproducibleBuildUsingNix.md
yakman2020: Correct, plus that derivation is coming from (I beleive) 19.09 or even 19.03, both EOL and deprecated.

yakman2020: IMO we should set up a common fork for nixpkgs that can be used by all of our partners and ourselves, which can be curated, and package versions set to reflect the OS versions we wish to support. Then the issue of packager version overrides etc disappears along with any possible reservations about trusting the packages

reproducible/README.md
yakman2020: Done.

docs/DesignDocs/ReproducibleBuildUsingNix.md
yakman2020: It has that.  
All the steps needed to create a reproducible package are done in the container.  There is no need for user intervention.

Building the build image is optional in the sense that if it was pushed somewhere one could pull it. Other than that, just run the build_openenclave.sh script and a suitable package will show up in the output directory.  In this, it is similar to linux-sgx/linux/reproducibility, which likewise just builds the package.  

Unlike the linux, it is quite practical to shift to a new version/tag and I probably should discuss that more.

reproducible/README.md
yakman2020: I've tried prototyping dumping the nix/store closure and reloading in another container with a similar nix store. That was able to run ctests directly. If the modle is we do the build, build the package then run ctests elsewhere, we can do that.

yakman2020: To be clear, are we saying that this is fundamentally a parallel process, therefore we can take the success of the unit tests  for granted, build the deb, then sanity check by building/running the samples?

We can do either. We can build/ctest at once, or build then dump the nix/store closure, then run another container and ctest.  In all cases we can and should sanity check the deb by installing into a container and validating, which infrastructure I believe already exists

In all cases we should dump the closure for the sake of anyone who wants to stay in nix world to build their downstream project.

docs/DesignDocs/ReproducibleBuildUsingNix.md
```diff
@@ -0,0 +1,397 @@
...
+be a singleton (for example integer), set (indicated by braces), or list (indicated by brackets []). Functions
+may include functions. They must be declared in the "let" section of the declaration.  Functions are called only 
+when referenced and reduced to values.
+
+When you see what appears to be regular code with braces and a series of semicolon separators is actually a set
+being used as an argument.
+
+There are no variables, just functions returning contants. Datatypes are a property of values, but once declared
+the function has a constant datatype. function results can be transformed but there is no casting as such. The let
+clause is where a number of functions can be declared. The in clause is where they are used.
```
dthaler, line 120:
```suggestion
the function has a constant datatype. function results can be transformed but there is no casting as such. The `let`
clause is where a number of functions can be declared. The `in` clause is where they are used.
```

docs/DesignDocs/ReproducibleBuildUsingNix.md
```diff
@@ -0,0 +1,397 @@
...
+```
+    arch = if platform ? arch then platform.arch else "x86_64-linux"
+```
+  
+There is no sequencing as such. Functions are composed through expression evaluation, so the expression:
+```
+    this that the_other 
+```
+will execute the function "the_other" and return its value as an argument to "that" which in turn returns its value to
+"this".  There are various caching facilities, so its possible that one or mor e of those arguments will be
```
dthaler, line 146:
```suggestion
"this".  There are various caching facilities, so it's possible that one or more of those arguments will be 
```

docs/DesignDocs/ReproducibleBuildUsingNix.md
```diff
@@ -0,0 +1,397 @@
...
+
+  installPhase =
+  ''
+       mkdir $out
+       cp -r ${myScript} $out/
+  '';
+}
+```
+  
+In the above, let is the initialisation/declaration part of the expression. The actual expression proper begins after
```
dthaler, line 189:
```suggestion
In the above, "let" is the initialisation/declaration part of the expression. The actual expression proper begins after
```

docs/DesignDocs/ReproducibleBuildUsingNix.md
```diff
@@ -0,0 +1,397 @@
...
+```
+   nix-build test-environment 
+```
+will perform both instantiate and realise. The result is a build output located in the nix-store.
+```
+/nix/store/syibqm72433pzxd46hyw3xbr5swq286j-test-environment
+```
+which contains:
+```
+$ ls /nix/store/syibqm72433pzxd46hyw3xbr5swq286j-test-environment
```
dthaler, line 260: Does nix work for Windows builds as well?  The title of this PR, the filename, and the intro in this doc are not Linux specific.  But various places herein only mention Linux.

docs/DesignDocs/ReproducibleBuildUsingNix.md
```diff
@@ -0,0 +1,397 @@
...
+  redhat, fedora, arch, gentoo) with easily auditable dependencies. In this case, it is necessary to
+  consider specific compatible versions of packages.  In particular, glibc must be an equal or older
+  revision than the glibc of the platform or the program will not dynamically link.  This is by design.
+  Ubuntu 18.04 is limited to glibc 2.27. The current lts release 20.04 usues glibc 2.31 and fedora 33 
+  uses libc-2.32.  The nixpkgs branch release-20.09 is based on glibc 2.32.  The somewhat older branch
+  20.03 uses libc-2.30 so can work on the current common distros, including the current rhel.  
+
+  While it is not difficult to change the version of glibc in a custom repo branch, since glibc is linked 
+  by every package, to do so would require a custom repo build. In that case, it would require a custom 
+  cache to provide prebuilding, or everything would require everything be built from source..
```
dthaler, line 288:
```suggestion
  cache to provide prebuilding, or everything would require everything be built from source.
``

File: reproducible/nix-build.sh (hunk 7 above)
Comment (yakman2020) [reply reproducible/nix-build.sh:21]: If there are downstream projects which would like to build in nix, we can accommodate them.

File: docs/DesignDocs/ReproducibleBuildUsingNix.md [hunk 21]
```diff
@@ -0,0 +1,397 @@
...
+
+### Caching and Hydra
+
+The official branches in nixpkgs are prebuilt and available at cache.nixos.org. Caching is keyed to the build outputs
+sha digest, so custom environments will be able to use those prebuilts so long as no changes have been made. 
+If any changes are made to that package, any package related to the changed package will need to be custom
+built and cached in nix store.
```
dthaler, line 295:
```suggestion
built and cached in the nix store.
```

docs/DesignDocs/ReproducibleBuildUsingNix.md
```diff
@@ -0,0 +1,397 @@
...
+
+
+## A Nix build system for OpenEnclave SDK
```
dthaler, line 298:
```suggestion
## A Nix build system for Open Enclave SDK
```

docs/DesignDocs/ReproducibleBuildUsingNix.md
```diff
@@ -0,0 +1,397 @@
...
+
+We propose to add an optional nix based build environment for OpenEnclave SDK using docker containers.
```
dthaler, line 300:
```suggestion
We propose to add an optional nix based build environment for Open Enclave SDK using docker containers.
```

docs/DesignDocs/ReproducibleBuildUsingNix.md
```diff
@@ -0,0 +1,397 @@
...
+
+This would consist of :
+- Dockerfile.nix and build-oe-nix-build.sh for building the oe nix docker image locally.
+- build_openenclave.sh for building the oe deb package via a nix environment.
+- Dockerfile.test and test-oe-nix-build.sh for verifying that package is functional.
+
+The process builds a reproducible .deb package of openenclave-sdk. 
+
+This process can be integrated with ADO or CPDX CI, but that is not part of this submission. We do not currently
```
dthaler, line 309: nit: expand ADO and CPDX acronyms on first use

docs/DesignDocs/ReproducibleBuildUsingNix.md
```diff
@@ -0,0 +1,397 @@
...
+   user and group information.  We default to "azureuser" "1000" "/home/azureuser"
+
+The script "build-oe-nix-image.sh" is a thin wrapper. It takes no arguments, and just passes the needed arguments
+to docker build.  If it needs to be modified, a new script should be created.
+
+Dockerfile.nix prepares the build image by:
+
+- adding via apt minimum packages for the outer image. This includes a text editor (vim) and ELF patch utility (patchelf).
+
+- sets up nit configuration.
```
dthaler, line 331: Should "nit" be "nix"?

reproducible/Dockerfile.nix
```diff
@@ -0,0 +1,186 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+# 
+ARG BASE_IMAGE="ubuntu@sha256:fff16eea1a8ae92867721d90c59a75652ea66d29c05294e6e2f898704bdb8cf1"
+FROM $BASE_IMAGE
+
+#
+# Build container to produce reproducible nix derivation and .deb package of the OpenEnclave SDK
```
dthaler, line 8:
```suggestion
# Build container to produce reproducible nix derivation and .deb package of the Open Enclave SDK
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
+## Nix Build system for OpenEnclave SDK
```
dthaler, line 1:
```suggestion
## Nix Build system for the Open Enclave SDK
```
"Open Enclave" is always 2 words

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+Nix expressions are kept in a git repo, github.com/nixos/nixpkgs. In principle a single version of each package
+is contained in the repo, so it is not necessary to specify package versions.  In practice these versions can be
+overridden, but that comes at a cost and is not recommended unless there is no option.
+
+The selection of packages (build outputs) in turn is defined by the branch and fork of the nix repo clone which
+you are using.
+
+Since the build outputs are measured, producing globally unique names, built versions can be cached. By default, 
+there is a global cache at https://cache.nixos.org which contains cached built derivations for the officially 
+released packages.  It is not necessary to use this cache. The cache protocol is simple http get/put protocol, so
```
dthaler, line 28:
```suggestion
released packages.  It is not necessary to use this cache. The cache protocol is a simple HTTP get/put protocol, so
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+any url that points to the correct heirarchy can be used. If no prebuilt version of a package is available at
+the cache, the package will be built. Since the package has a fixed output, there is no difference in output
+using the cached version or one custom built..
```
dthaler, line 31:
```suggestion
using the cached version or one custom built.
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+using the cached version or one custom built.. 
+
+#### Language:
+
+Nix is a language which looks much like Haskell. It is a Haskell derivitive in the same sense that
+Java is a C++ derivitive, which is to say it carries over syntactic conventions but not strictly or
+completely, and adds some features not in Haskell.
+
+The basis for the language is the function.  A function is an expression which returns a value, the value can
+be a singleton (for example integer), set (indicated by braces), or list (indicated by brackets []). Functions
```
dthaler, line 40:
```suggestion
be a singleton (for example integer), set (indicated by curly braces {}), or list (indicated by brackets []). Functions
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+may include functions. They must be declared in the "let" section of the declaration.  Functions are called only 
+when referenced and reduced to values.
+
+When you see what appears to be regular code with braces and a series of semicolon separators is actually a set
+being used as an argument.
+
+There are no variables, just functions returning contants. Datatypes are a property of values, but once declared
+the function has a constant datatype. function results can be transformed but there is no casting as such. The let
+clause is where a number of functions can be declared. The in clause is where they are used.
```
dthaler, line 49:
```suggestion
the function has a constant datatype. function results can be transformed but there is no casting as such. The "let"
clause is where a number of functions can be declared. The "in" clause is where they are used.
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+    # function returning a set. Set members are retrieved through the dot operator, so myset."one":
+    myset = { one = 1; two = 2; three = 3; }
+    #
+    mylist = [ "one", "two", "three" ];
+in 
+     
+```
+
+if then else exists, but the form is similar to the C conditional,  if <cond> then <alt1> else <alt2> where cond,
+alt1 and alt2 are all function calls.  if returns a value.
```
dthaler, line 64:
```suggestion
"if then else exists", but the form is similar to the C conditional,  `if <cond> then <alt1> else <alt2>` where `cond`,
`alt1` and `alt2` are all function calls.  "if" returns a value.
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+```
+    arch = if platform ? arch then platform.arch else "x86_64-linux"
+```
+  
+There is no sequencing as such. Functions are composed through expression evaluation, so the expression:
+```
+    this that the_other 
+```
+will execute the function "the_other" and return its value as an argument to "that" which in turn returns its value to
+"this".  There are various caching facilities, so its possible that one or mor e of those arguments will be
```
dthaler, line 75:
```suggestion
"this".  There are various caching facilities, so it's possible that one or more of those arguments will be 
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+
+  installPhase =
+  ''
+       mkdir $out
+       cp -r ${myScript} $out/
+  '';
+}
+```
+  
+In the above, let is the initialisation/declaration part of the expression. The actual expression proper begins after
```
dthaler, line 118:
```suggestion
In the above, "let" is the initialisation/declaration part of the expression. The actual expression proper begins after
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+  redhat, fedora, arch, gentoo) with easily auditable dependencies. In this case, it is necessary to
+  consider specific compatible versions of packages.  In particular, glibc must be an equal or older
+  revision than the glibc of the platform or the program will not dynamically link.  This is by design.
+  Ubuntu 18.04 is limited to glibc 2.27. The current lts release 20.04 usues glibc 2.31 and fedora 33 
+  uses libc-2.32.  The nixpkgs branch release-20.09 is based on glibc 2.32.  The somewhat older branch
+  20.03 uses libc-2.30 so can work on the current common distros, including the current rhel.  
+
+  While it is not difficult to change the version of glibc in a custom repo branch, since glibc is linked 
+  by every package, to do so would require a custom repo build. In that case, it would require a custom 
+  cache to provide prebuilding, or everything would require everything be built from source..
```
dthaler, line 217:
```suggestion
  cache to provide prebuilding, or everything would require everything be built from source.
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+
+### Caching and Hydra
+
+The official branches in nixpkgs are prebuilt and available at cache.nixos.org. Caching is keyed to the build outputs
+sha digest, so custom environments will be able to use those prebuilts so long as no changes have been made. 
+If any changes are made to that package, any package related to the changed package will need to be custom
+built and cached in nix store.
```
dthaler, line 224:
```suggestion
built and cached in the nix store.
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+
+
+## A Nix build system for OpenEnclave SDK
```
dthaler, line 227:
```suggestion
## A Nix build system for the Open Enclave SDK
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+
+We propose to add an optional nix based build environment for OpenEnclave SDK using docker containers.
```
dthaler, line 229:
```suggestion
We propose to add an optional nix based build environment for the Open Enclave SDK using docker containers.
```
Will this work for Windows builds too?

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+   user and group information.  We default to "azureuser" "1000" "/home/azureuser"
+
+The script "build-oe-nix-image.sh" is a thin wrapper. It takes no arguments, and just passes the needed arguments
+to docker build.  If it needs to be modified, a new script should be created.
+
+Dockerfile.nix prepares the build image by:
+
+- adding via apt minimum packages for the outer image. This includes a text editor (vim) and ELF patch utility (patchelf).
+
+- sets up nit configuration.
```
dthaler, line 260: Should "nit" be "nix"?

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+  modified by the addition of apt packages.
+
+The resulting image is named openenclave-build. It can (and should) be generated once and pushed into a common 
+repo.
+
+### build_openenclave.sh
+
+Runs the image produced by docker build, with arguments resolved.
+The output of the process is a .deb and a .nar file containing an exported nix store for use in developing 
+openenclave sdk apps using nix.
```
dthaler, line 300:
```suggestion
Open Enclave SDK apps using nix.
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+```
+   ./build-oe-nix-image.sh
+```
+will build the nix docker image. This can be pushed:
+```
+   docker push openenclave-build
+```
+and later pulled, or just used locally.
+
+To build the openenclave sdk
```
dthaler, line 314:
```suggestion
To build the Open Enclave SDK
```

reproducible/README.md
```diff
@@ -0,0 +1,343 @@
...
+repo.
+
+### build_openenclave.sh
+
+Runs the image produced by docker build, with arguments resolved.
+The output of the process is a .deb and a .nar file containing an exported nix store for use in developing 
+openenclave sdk apps using nix.
+
+
+### workflow
```
dthaler, line 303:
```suggestion
### Workflow
```

reproducible/build-oe-nix-image.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+# 
+# build-oe-nix-image.sh  
+# Builds the nix based build image for OpenEnclave SDK
```
dthaler, line 6:
```suggestion
# Builds the nix based build image for the Open Enclave SDK
```

reproducible/build_openenclave.sh
```diff
@@ -0,0 +1,97 @@
...
+while getopts "o:tic" opt; do
+  case ${opt} in
+    o ) # output dir
+        OUTPUT_DIR=$OPTARG
+        echo "output dir = ${OUTPUT_DIR}"
+      ;;
+    t ) # run tests with build
+        DO_TESTS=true
+      ;;
+    i ) # Dont execute nix-build.sh, just bash
```
dthaler, line 29:
```suggestion
    i ) # Don't execute nix-build.sh, just bash
```

dthaler (changes requested):
libprotobuf is included in binary form in this PR.  https://github.com/protocolbuffers/protobuf/blob/master/LICENSE says
"    * Redistributions in binary form must reproduce the above
copyright notice, this list of conditions and the following disclaimer
in the documentation and/or other materials provided with the
distribution."

Where is the above required info provided?

Also, the PR title, description, top of README, top of design doc, etc all talk about the "build process" for OE SDK, implying it is the only one.  But issue #3729 says:

> I am not proposing that NIX become the mainline build process for production.
> 
> NIX supports linux only, and is unlikely to ever support Windows.

Hence please update the PR title, description, README, and design doc to explain that this is a Linux build process, and explain whether we already have a reproducible build process for Windows or if this is a gap this PR does not address (in which case it will not completely resolve the issue).

reproducible/prep-nix-build.sh
```diff
@@ -0,0 +1,126 @@
+#!/bin/bash
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+# 
+
+#
+# Setup a host for docker builds using nix. This assumes nix has been installed
```
dthaler, line 7:
```suggestion
# Set up a host for docker builds using nix. This assumes nix has been installed.
```

reproducible/prep-nix-build.sh
```diff
@@ -0,0 +1,126 @@
...
+# We are currently installing the buildinputs for openenclave as derivations 
+# so we could audit the code. 
+#
+# Outside this script we are configuring nix to keep-derivations so we can inspect them as 
+# needed. With that, we can inspect source of all the build inputs if desired.
+#
+# We can also force a complete build from source by building openenclave via
+# nix-build --substituters '' .... . Since the build result is deterministic, 
+# we can verify we are getting the packages we think we are, as the output will have
+# the same sha256.
```
dthaler, line 70:
```suggestion
# the same sha256 hash.
```

mikbras (commented):
Thank you so much Bruce for doing all this. I see you have gotten pretty good coverage from others on the documentation and the scripts. I have some general remarks after trying to build.

- It might be nice to have a "quick start" guide, explaining how to kick off the process and where the output is placed. The README.md provides a lot of background that might not be interesting to the practitioner and maybe could be left for the design doc.
- The readme and the design document have identical content in several places. Maybe it would be better to remove the duplication and just use the readme as the quick-start guide.
-  Just checking but I thought I saw shared objects in the source tree in the PR, but they don't seem to be there when I checked out.
- My first attempt to perform a reproducible build was to examine the contents of the **reproducible** directory. I saw many scripts and was not sure which ones I could run directly. Some scripts are not meant to be run directly by the user while others are not. It might help to separate the internal scripts into their own directory. For example, maybe consider having a docker directory that contains everything comprising the Docker image.

I am still learning the system and trying to get some hands-on experience so I can provide more useful feedback. I'll have more questions and comments later.

reproducible/README.md
yakman2020: So, you can't run ctest without the rest of the nix-store when you build in nix land, but you can export the closure (the build output plus any references made from the build output) then import it into a new instance of the container. I have done that, so not you build in one container and can test in the other.

A side effect of this is that it is possible to use the closure for openenclave-sdk to build additional downstream projects using the sdk. I will be adding that in a little bit.

reproducible/nix-build.sh
yakman2020: Take a look at ctest_openenclave.sh for an example. We export the build closure before we adjust the ld interpreter and rpath to take the binaries out of nix-world.  The closures retain their nix features.

reproducible/Dockerfile.nix
yakman2020: the .nix_lib contents are removed from the build container altogether and only added in the test container.

docs/DesignDocs/ReproducibleBuildUsingNix.md
yakman2020: Nix is completely foreign to windows, and I am unaware of any move to port it.  Currently Visual C compiler and linker brand object files with time/date/user/permissions etc information that make it impossible to do a reproducible build.  A repro build could be done using clang, but I suspect that would be unacceptable to most windows developers.

yakman2020: I did change the PR title and comment.

reproducible/prep-nix-build.sh
yakman2020: The acc-test2 branch is based on release-20.03. There is an openenclave-sdk directory but I have not tried doing a full hydra build until the basic pr is done. The reason for the many commits is that I had to merge with 20.03 when 20.09 turned out to be unusable.  I could rebuild the branch from release-20.03 and eliminate the commits.

reproducible/prep-nix-build.sh
yakman2020: These packages are in fact pulled from cache. Note "substituters".

reproducible/README.md
yakman2020: changed to ~/result which is an absolute path. Relative paths are not allowed.

radhikaj:
Contributors can reopen when they are willing to actively work on this in the future. Closing for now.

PR #3746, opened 2020-11-20
Enforce the correct build dependency on oecryptopopenssl

### LINKED ISSUE #3745: cannot find a header file in building with OpenSSL
I tried to config with `cmake .. -DBUILD_OPENSSL=ON` and build, and got a failure:
```
openenclave/common/crypto/openssl/asn1.c:11:10: fatal error: 'openssl/asn1.h' file not found
#include <openssl/asn1.h>
         ^~~~~~~~~~~~~~~~
1 error generated.
```
I also tried the last `verified `commit `f83f25b81f7bb629e2fd14c212f077cbc2fb8e7f`, and result was the same as the current latest commit `700e999a144fa6de50fc6af65a74eafdd9b4e76e`.

There is at least one configuration of OpenSSL from the bors system, so is there anything I need to fix on my local machines? Actually the problem is also reproducible on an Azure VM.

mingweishih:
This PR fixes #3745, which ensures the OpenSSL is always built before the `oecryptoopenssl`.

### DIFF
--- a/enclave/crypto/openssl/CMakeLists.txt
+++ b/enclave/crypto/openssl/CMakeLists.txt
@@ -34,6 +34,9 @@ enclave_include_directories(
 
 enclave_link_libraries(oecryptoopenssl PUBLIC openssl)
 
+# Enforce the correct build dependency such that OpenSSL will be built first.
+add_enclave_dependencies(oecryptoopenssl opensslcrypto)
+
 set_enclave_property(TARGET oecryptoopenssl PROPERTY ARCHIVE_OUTPUT_DIRECTORY
                      ${OE_LIBDIR}/openenclave/enclave)
 


nonpolarity (approved):

anakrish (approved):
/lgtm

On a related note, it seems odd to see oe libraries without the `oe_` prefix.
Is there a reason why the library is called `opensslcrypto` and not `oeopensslcrypto`?

mingweishih:
> /lgtm
> 
> On a related note, it seems odd to see oe libraries without the `oe_` prefix.
> Is there a reason why the library is called `opensslcrypto` and not `oeopensslcrypto`?

This naming follows `mbedcrypto` that does not come with oe prefix as well. Maybe we used this naming to distinguish between `mbedcrypto` and `oecryptombed` back then.

PR #3748, opened 2020-11-23
Fix failure in nightly builds on Ubuntu 16.04 GCC

anakrish:
Add oelibcxx dependency. Without this, abi_enc build is launched befre libcxx headers are generated at correct location.

### DIFF
--- a/tests/abi/enc/CMakeLists.txt
+++ b/tests/abi/enc/CMakeLists.txt
@@ -15,10 +15,10 @@ add_enclave(
   abi_enc
   UUID
   7B7FB79E-2B73-46C5-8C30-3A71C014BD85
+  CXX
   SOURCES
   enc.cpp
   ${CMAKE_CURRENT_BINARY_DIR}/abi_t.c)
 
 enclave_include_directories(abi_enc PRIVATE ${CMAKE_CURRENT_BINARY_DIR}
                             ${CMAKE_CURRENT_SOURCE_DIR})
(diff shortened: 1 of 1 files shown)


anakrish:
bors try

tests/abi/enc/CMakeLists.txt
```diff
@@ -21,4 +21,4 @@ add_enclave(
 
 enclave_include_directories(abi_enc PRIVATE ${CMAKE_CURRENT_BINARY_DIR}
                             ${CMAKE_CURRENT_SOURCE_DIR})
-enclave_link_libraries(abi_enc oelibc)
+enclave_link_libraries(abi_enc oelibcxx)
```
mingweishih, line 24: If this works, it would be good to add the `CXX` to the `add_enclave` function above instead this.

anakrish: Sure. What is the difference between the two?

mingweishih: It does the same thing. The idea is that we don't need to manually specify the `oelibc` (always add) or `oelibcxx` (add when `CXX` is supplied) with `add_enclave`.

anakrish: Ah Ok. Done.

anakrish:
bors try

mingweishih (approved):
LGTM. Thanks for root causing the problem and the quick fix!

radhikaj (approved):

PR #3753, opened 2020-11-25
Mitigate indirect jumps in assembly code.

anakrish:

### DIFF
--- a/enclave/core/sgx/context.inc
+++ b/enclave/core/sgx/context.inc
@@ -143,9 +143,27 @@ oe_continue_execution:
     // Restore rdi.
     movq    OE_CONTEXT_RDI(%rdi), %rdi
 
-    // Recovery the correct rsp.
-    add     $(ABI_REDZONE_BYTE_SIZE + 8), %rsp
+    // LVI mitigation strategy:
+    // Since no register is free, we cannot load the target into a register and
+    // then issue an lfence. Instead, we use retq (return) instruction to carry
+    // out a jump. The return address is at 0(RSP). RSP is ABI_REDZONE_BYTE_SIZE
+    // + 8 bytes below the correct RSP which was fetched from the context.
+    // We issue a
+    //     retq $ABI_REDZONE_BYTE_SIZE
+    // This has the following effect:
+    //     - The return address is popped and control is transferred to
+    //       the target address.
+    //     - Additionally, ABI_REDZONE_BYTE_SIZE is subtracted from RSP,
+    //       thereby restoring RSP correctly.
+    //     - Ideally assembler's LVI mitigation would insert lfence as appropriate.
+    //       Since it does not seem to handle retq with an immediate value, the
+    //       LVI mitigation is manually inserted.
 
-    // Jump to the target.
-    jmp     *-(ABI_REDZONE_BYTE_SIZE + 8)(%rsp)
+    // Manual LVI mitigation since the assember does not handle the retq below.
+    notq (%rsp)
+    notq (%rsp)
+    lfence
+
+    // Jump to the target and restore RSP.
+    retq $ABI_REDZONE_BYTE_SIZE
 .cfi_endproc
--- a/enclave/core/sgx/longjmp.S
+++ b/enclave/core/sgx/longjmp.S
@@ -32,5 +32,10 @@ val_is_nonzero:
     mov 40(%rdi),%r13
     mov 48(%rdi),%r14
     mov 56(%rdi),%r15
-    jmp *16(%rdi)
+
+    // Load indirect jump address into register and then perform the jump.
+    // This pattern allows the assembler to detect and add lfence as
+    // appropriate.
+    mov 16(%rdi),%rdi
+    jmp *%rdi
 .cfi_endproc


anakrish:
bors try

enclave/core/sgx/context.inc
```diff
@@ -147,5 +147,12 @@ oe_continue_execution:
...
 
     // Jump to the target.
-    jmp     *-(ABI_REDZONE_BYTE_SIZE + 8)(%rsp)
+    // Since no register is free, we cannot load the target into a register and
+    // then issue an lfence. Instead, we use retq (return) instruction to carry
+    // out a jump. We first push the target address onto the stack, issue an
+    // lfence, and then do a retq. The retq will pop the target from the stack
+    // and then jump to it.
+    pushq     -(ABI_REDZONE_BYTE_SIZE + 8)(%rsp)
+    lfence
```
bodzhang, line 156: `ret` instruction loads (reads) from memory (stack), then jump. A `lfence` instruction inserted before `ret` might not "cut" the potential LVI flow targeting the top of the stack.  

The `lfence` instruction does prevent speculative execution of `ret` before `pushq` instruction on line 155 retires (any exception on that instruction will be delivered before execution of next instruction), thus "cuts" the potential LVI flow targeting `-(ABI_REDZONE_BYTE_SIZE + 8)(%rsp)` address.  The `pushq` instruction on line 155 also stores (writes) to the top of the stack. But a potential attack might be able to manipulate the r/w permission of top of the stack location and only cause an exception when load from that location. In that case, the load part of the `ret` instruction will trigger exception and lead to potential LVI attack.

If the assembler-based LVI mitigation is enabled, according to Intel's [LVI Deep Dive Whitepaper](https://software.intel.com/security-software-guidance/deep-dives/deep-dive-load-value-injection#specialinstructions ), the assembler is supposed to replace `ret` instruction with:

SHL QWORD PTR [rsp], 0 
LFENCE 
RET 

The instruction sequence loads from target address, lfence, then load/jump. According tot he white paper, it "cuts" the potential LVI flow targeting the top of the stack. 

Relying on the assembler-based mitigation, the following instruction sequence should be sufficient.

    pushq     -(ABI_REDZONE_BYTE_SIZE + 8)(%rsp)
    ret

The white paper also mentions similar technique for indirect branch through memory operand:

XOR QWORD PTR <some GPR> , [rsi] 
XOR QWORD PTR <some GPR> , [rsi] 
LFENCE
JMP QWORD PTR [rsi] 

I wonder instead of using XOR to load the target memory, but causing flag change, whether `SHL QWORD PTR [rsi] , 0` also works.

enclave/core/sgx/longjmp.S
```diff
@@ -32,5 +32,11 @@ val_is_nonzero:
     mov 40(%rdi),%r13
     mov 48(%rdi),%r14
     mov 56(%rdi),%r15
-    jmp *16(%rdi)
+
+    // Load indirect jump address into register and issue an lfence to mitigate
+    // LVI.
+    mov 16(%rdi),%rdi
+    lfence
```
bodzhang, line 39: Assembler-based LVI mitigation is suppose to insert `lfence` before the indirect branch with register operand. No need to manually insert one. For SGX apps targeting CPUs not affected by LVI only, Assembler-based LVI mitigation is not required, nor is manually inserted `lfence`. Stopping speculative execution when its not necessary for security reason does impact performance.

If the assembly sub-routine is not executed frequently, an extra `lfence` inserted manually probably will not cause too much performance degradation though.

bodzhang (changes requested):

enclave/core/sgx/context.inc
mingweishih: After converting the indirect `jmp` to `ret`, the assembler (when enabling LVI mitigation) should take care of what @bodzhang described. No need to manually insert `lfence`.

enclave/core/sgx/longjmp.S
mingweishih: Agree with @bodzhang. Similar to the case above, the assembler should be able to recognize the converted indirect `jmp` and correctly insert `lfence` for us.

anakrish: Inspecting the generated assembly code,
```bash
objdump -d echo_enc-lvi-cfg | grep -B 2 "jmpq   \*"

--
  129c9e:       4c 8b 77 30             mov    0x30(%rdi),%r14
  129ca2:       4c 8b 7f 38             mov    0x38(%rdi),%r15
  129ca6:       ff 67 10                jmpq   *0x10(%rdi)
```

the assembler fails to insert an lfence here. Therefore, we need to add an lfence here.
@mingweishih Can you please double-check if my analysis is correct?

enclave/core/sgx/context.inc
anakrish: Thanks for the explanation @bodzhang. I removed the extra lfence.
Note, the assembler transforms retq instruction as below:
```asm
    notq (%rsp)
    lfence
    retq
```

enclave/core/sgx/longjmp.S
anakrish: @mingweishih nvm. I double-checked and removed the extra lfence.

anakrish:
bors try

mingweishih (approved):
LGTM

anakrish:
bors try

enclave/core/sgx/context.inc
```diff
@@ -143,9 +143,27 @@ oe_continue_execution:
     // Restore rdi.
     movq    OE_CONTEXT_RDI(%rdi), %rdi
 
-    // Recovery the correct rsp.
-    add     $(ABI_REDZONE_BYTE_SIZE + 8), %rsp
-
-    // Jump to the target.
-    jmp     *-(ABI_REDZONE_BYTE_SIZE + 8)(%rsp)
+    // LVI mitigation strategy:
+    // Since no register is free, we cannot load the target into a register and
```
mingweishih, line 147:
```suggestion
    // Since no register is freed, we cannot load the target into a register and
```

enclave/core/sgx/context.inc
```diff
@@ -143,9 +143,27 @@ oe_continue_execution:
...
+    // then issue an lfence. Instead, we use retq (return) instruction to carry
+    // out a jump. The return address is at 0(RSP). RSP is ABI_REDZONE_BYTE_SIZE
+    // + 8 bytes below the correct RSP which was fetched from the context.
+    // We issue a
+    //     retq $ABI_REDZONE_BYTE_SIZE
+    // This has the followin effect:
```
mingweishih, line 153:
```suggestion
    // This has the following effect:
```

mingweishih (approved):
LGTM

enclave/core/sgx/context.inc
anakrish: Retaining original.

anakrish:
bors try

bodzhang (approved):

PR #3754, opened 2020-11-25
Update RHEL drivers

rs--:
Updated RHEL drivers to match Ubuntu Intel drivers:

PSW - 2.10.100
DCAP - 1.7.100

### DIFF
--- a/scripts/ansible/roles/linux/intel/tasks/redhat/sgx-packages-install.yml
+++ b/scripts/ansible/roles/linux/intel/tasks/redhat/sgx-packages-install.yml
@@ -44,10 +44,38 @@
   register: install_sgx
   until: install_sgx is success
 
-- name: Install the Intel DCAP packages
+- name: Download Intel SGX DCAP Packages
+  get_url:
+    url: "{{ intel_dcap_packages_url }}"
+    dest: "/tmp/sgx_linux_dcap_packages.tgz"
+    mode: 0755
+    timeout: 120
+  retries: 3
+  when: flc_enabled|bool
+
+- name: Create Intel DCAP directory
+  file:
+    path: "{{ intel_dcap_packages_directory }}"
+    state: directory
+
+- name: Unarchive Intel DCAP local rpm repo tgz
+  unarchive:
+    src: "/tmp/sgx_linux_dcap_packages.tgz"
+    dest: "{{ intel_dcap_packages_directory }}"
+    remote_src: yes
+
+- name: Enable local Intel DCAP repo
+  yum_repository:
+    name: intel-dcap-local-repo
+    description: Intel DCAP local RPM repository
+    baseurl: "file://{{ intel_dcap_packages_directory }}/sgx_rpm_local_repo"
+    enabled: yes
+    gpgcheck: no
+
+- name: Install Intel DCAP packages
   yum:
     name: "{{ intel_dcap_packages }}"
-    state: latest
+    state: present
   when: flc_enabled|bool
   retries: 10
   register: install_dcap
--- a/scripts/ansible/roles/linux/intel/tasks/redhat/sgx-packages-install.yml
+++ b/scripts/ansible/roles/linux/intel/tasks/redhat/sgx-packages-install.yml
@@ -47,35 +47,45 @@
 - name: Download Intel SGX DCAP Packages
   get_url:
     url: "{{ intel_dcap_packages_url }}"
-    dest: "/tmp/sgx_linux_dcap_packages.tgz"
+    dest: "/tmp/"
     mode: 0755
     timeout: 120
   retries: 3
   when: flc_enabled|bool
+  register: get_dcap_package
 
 - name: Create Intel DCAP directory
   file:
     path: "{{ intel_dcap_packages_directory }}"
     state: directory
 
+- name: Clean local rpm repo if it exists
+  file:
+    path: "{{ intel_dcap_packages_directory }}/sgx_rpm_local_repo"
+    state: absent
+
 - name: Unarchive Intel DCAP local rpm repo tgz
   unarchive:
-    src: "/tmp/sgx_linux_dcap_packages.tgz"
+    src: "{{ get_dcap_package.dest }}"
     dest: "{{ intel_dcap_packages_directory }}"
     remote_src: yes
+  when:
+    - get_dcap_package.changed
+    - get_dcap_package.dest
 
 - name: Enable local Intel DCAP repo
   yum_repository:
     name: intel-dcap-local-repo
     description: Intel DCAP local RPM repository
     baseurl: "file://{{ intel_dcap_packages_directory }}/sgx_rpm_local_repo"
-    enabled: yes
+    enabled: no
     gpgcheck: no
 
 - name: Install Intel DCAP packages
   yum:
     name: "{{ intel_dcap_packages }}"
     state: present
+    enablerepo: intel-dcap-local-repo
   when: flc_enabled|bool
   retries: 10
   register: install_dcap
@@ -83,5 +93,5 @@
 
 - name: Cleanup installation
   file:
-    path: "/tmp/sgx_rpm_local_repo.tgz"
+    path: "{{ get_dcap_package.dest }}"
     state: absent
--- a/scripts/ansible/roles/linux/intel/vars/redhat/ootpa.yml
+++ b/scripts/ansible/roles/linux/intel/vars/redhat/ootpa.yml
@@ -2,9 +2,9 @@
 # Licensed under the MIT License.
 
 ---
-intel_sgx_w_flc_driver_url: "https://download.01.org/intel-sgx/sgx-linux/2.12/distro/rhel8.2-server/sgx_linux_x64_driver_1.36.2.bin"
-intel_sgx1_driver_url: "https://download.01.org/intel-sgx/sgx-linux/2.12/distro/rhel8.2-server/sgx_linux_x64_driver_2.11.0_4505f07.bin"
-intel_sgx_rpm_repo_tgz_url: "https://download.01.org/intel-sgx/sgx-linux/2.12/distro/rhel8.2-server/sgx_rpm_local_repo.tgz"
+intel_sgx_w_flc_driver_url: "https://download.01.org/intel-sgx/sgx-linux/2.10/distro/rhel8.1-server/sgx_linux_x64_driver_1.35.bin"
+intel_sgx1_driver_url: "https://download.01.org/intel-sgx/sgx-linux/2.10/distro/rhel8.1-server/sgx_linux_x64_driver_2.6.0_602374c.bin"
+intel_sgx_rpm_repo_tgz_url: "https://download.01.org/intel-sgx/sgx-linux/2.10/distro/rhel8.1-server/sgx_rpm_local_repo.tgz"
 intel_sgx_package_dependencies:
   - "kernel-devel-uname-r == {{ ansible_kernel }}"
   - "elfutils-libelf-devel"
@@ -13,7 +13,7 @@ intel_sgx_package_dependencies:
   - "protobuf-devel"
   - "dkms"
 
-intel_dcap_packages_url: "https://download.01.org/intel-sgx/sgx-dcap/1.9/linux/distro/rhel8.2-server/sgx_linux_x64_driver_1.36.2.bin"
+intel_dcap_packages_url: "https://download.01.org/intel-sgx/sgx-dcap/1.7/linux/distro/rhel8.1-server/sgx_linux_x64_driver_1.35.bin"
 intel_dcap_packages_directory: "/tmp/sgx_linux_dcap_packages"
 intel_dcap_packages:
   - "libsgx-ae-qve"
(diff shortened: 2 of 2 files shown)


scripts/ansible/roles/linux/intel/vars/redhat/ootpa.yml
```diff
@@ -13,8 +13,10 @@ intel_sgx_package_dependencies:
...
   - "dkms"
 
+intel_dcap_packages_url: "https://download.01.org/intel-sgx/sgx-dcap/1.7/linux/distro/rhel8.1-server/sgx_linux_x64_driver_1.35.bin"
+intel_dcap_packages_directory: "/tmp/sgx_linux_dcap_packages"
 intel_dcap_packages:
-  - "https://download.01.org/intel-sgx/sgx-dcap/1.6/linux/distro/rhel8.0-Server/libsgx-ae-qve-1.6.100.2-1.el8.x86_64.rpm"
-  - "https://download.01.org/intel-sgx/sgx-dcap/1.6/linux/distro/rhel8.0-Server/libsgx-dcap-ql-1.6.100.2-1.el8.x86_64.rpm"
-  - "https://download.01.org/intel-sgx/sgx-dcap/1.6/linux/distro/rhel8.0-Server/libsgx-dcap-ql-devel-1.6.100.2-1.el8.x86_64.rpm"
-  - "https://download.01.org/intel-sgx/sgx-dcap/1.6/linux/distro/rhel8.0-Server/libsgx-urts-2.9.101.2-1.el8.x86_64.rpm"
+  - "libsgx-ae-qve"
```
radhikaj, line 19: How would you construct the full path to these with https://download.01.org/intel-sgx/sgx-dcap/1.6/linux/distro/rhel8.0-Server/

scripts/ansible/roles/linux/intel/vars/redhat/ootpa.yml
```diff
@@ -2,9 +2,9 @@
 # Licensed under the MIT License.
 
 ---
-intel_sgx_w_flc_driver_url: "https://download.01.org/intel-sgx/sgx-dcap/1.6/linux/distro/rhel8.0-Server/sgx_linux_x64_driver_1.33.bin"
-intel_sgx1_driver_url: "https://download.01.org/intel-sgx/sgx-linux/2.9.1/distro/rhel8.0-server/sgx_linux_x64_driver_2.6.0_95eaa6f.bin"
-intel_sgx_rpm_repo_tgz_url: "https://download.01.org/intel-sgx/sgx-linux/2.9.1/distro/rhel8.0-server/sgx_rpm_local_repo.tgz"
+intel_sgx_w_flc_driver_url: "https://download.01.org/intel-sgx/sgx-linux/2.10/distro/rhel8.1-server/sgx_linux_x64_driver_1.35.bin"
```
BRMcLaren, line 5: We actually want the sgx-dcap drivers, the ones from sgx-linux do not have flc IIRC.

https://download.01.org/intel-sgx/sgx-dcap/1.9/linux/distro/rhel8.2-server/
https://download.01.org/intel-sgx/sgx-linux/2.12/distro/rhel8.2-server/

Are the URL's I believe.

BRMcLaren:
bors try

I am curious about the flop from FQDN to direct package naming

scripts/ansible/roles/linux/intel/vars/redhat/ootpa.yml
CyanDevs: I believe the intention is to download the packages, add them to a local yum repo, and then install the same packages using the package name. 

Everything would be bundled into the SGX driver version (in this case v1.35) and there is no control over the versions of the individual packages themselves. Is this the desired behaviour?

scripts/ansible/roles/linux/intel/vars/redhat/ootpa.yml
rs--: I was matching it up with the drivers we have for Ubuntu. If we want the latest, should I bump the Ubuntu drivers as well?

scripts/ansible/roles/linux/intel/vars/redhat/ootpa.yml
rs--: Yes, @CyanDevs  that's correct. Since the packages are no longer uploaded individually, we download the package bundle, unarchive it, and add it to the local yum repo in `scripts/ansible/roles/linux/intel/tasks/redhat/sgx-packages-install.yml`

scripts/ansible/roles/linux/intel/vars/redhat/ootpa.yml
BRMcLaren: Gotcha. I'm sure the binaries are the same, so no action required.

intel_sgx_w_flc_driver_url should be using the sgx-dcap driver though, Ubuntu is the [same](https://github.com/openenclave/openenclave/blob/4d56a217912e44bd5c0379f6306bbba4b553824f/scripts/ansible/roles/linux/intel/vars/ubuntu/xenial.yml#L5)

scripts/ansible/roles/linux/intel/tasks/redhat/sgx-packages-install.yml
```diff
@@ -44,10 +44,38 @@
   register: install_sgx
   until: install_sgx is success
 
-- name: Install the Intel DCAP packages
+- name: Download Intel SGX DCAP Packages
+  get_url:
+    url: "{{ intel_dcap_packages_url }}"
+    dest: "/tmp/sgx_linux_dcap_packages.tgz"
```
CyanDevs, line 50: There's a trick that can be done to make get_url and subsequent actions idempotent and more reliable. Some advantages include:
* avoid renaming the package and losing the versioning information.
* skip tasks that don't need to be reran (for idempotency)

For the get_url module you can register the return value to be used. 
To get the value of wherever the file was downloaded, you can access the return value's `dest` key. For example if you registered the return value like below, then you can refer to it in as get_dcap_package.dest
```get_url:
    url: "{{ intel_dcap_packages_url }}"
    dest: "/tmp/"
    mode: 0755
    timeout: 120
  retries: 3
  when: flc_enabled|bool
  register: get_dcap_package
```

This is also useful when there is a subsequent step that uses the file, such as the unarchive step below. You can see if the download occurred, and if so, then run the unarchive step using get_dcap_package.dest as the src.
```
- name: Unarchive Intel DCAP local rpm repo tgz
  unarchive:
    src: "{{ get_dcap_package.dest }}"
    dest: "{{ intel_dcap_packages_directory }}"
    remote_src: yes
  when: 
    - get_dcap_package.changed
    - get_dcap_package.dest
```

So if the download didn't occur because the file existed, then it will skip the unarchiving step.

There is a caveat to this, that if the system is somehow in the state where the file is downloaded, but not extracted, then the unarchive step will be skipped. You can get around this by adding a playbook variable to force it. For example:
```
- name: Unarchive Intel DCAP local rpm repo tgz
  unarchive:
    src: "{{ get_dcap_package.dest }}"
    dest: "{{ intel_dcap_packages_directory }}"
    remote_src: yes
  when: get_dcap_package.changed and get_dcap_package.dest or
             force_dcap_package_unarchive | bool
```

scripts/ansible/roles/linux/intel/tasks/redhat/sgx-packages-install.yml
```diff
@@ -44,10 +44,38 @@
...
+  retries: 3
+  when: flc_enabled|bool
+
+- name: Create Intel DCAP directory
+  file:
+    path: "{{ intel_dcap_packages_directory }}"
+    state: directory
+
+- name: Unarchive Intel DCAP local rpm repo tgz
+  unarchive:
```
CyanDevs, line 62: See the unarchive module's `create` option, which is useful for determining if the step needs to be ran or not given the existence of a file path. https://docs.ansible.com/ansible/latest/collections/ansible/builtin/unarchive_module.html
For example:
```
- name: Unarchive Intel DCAP local rpm repo tgz
  unarchive:
    src: "/tmp/sgx_linux_dcap_packages.tgz"
    dest: "{{ intel_dcap_packages_directory }}"
    remote_src: yes
    creates: "{{ intel_dcap_packages_directory }}/sgx_rpm_local_repo"
```

scripts/ansible/roles/linux/intel/tasks/redhat/sgx-packages-install.yml
```diff
@@ -44,10 +44,38 @@
...
+    src: "/tmp/sgx_linux_dcap_packages.tgz"
+    dest: "{{ intel_dcap_packages_directory }}"
+    remote_src: yes
+
+- name: Enable local Intel DCAP repo
+  yum_repository:
+    name: intel-dcap-local-repo
+    description: Intel DCAP local RPM repository
+    baseurl: "file://{{ intel_dcap_packages_directory }}/sgx_rpm_local_repo"
```
CyanDevs, line 71: Since `intel_dcap_packages_directory: /tmp/sgx_linux_dcap_packages` this repository would be broken if /tmp is cleaned out, and would break all yum commands until the repo is fixed. I would suggest leaving this disabled and then in the yum module below use `enablerepo` to temporarily enable it for the installation. For example:
```
- name: Enable local Intel DCAP repo
  yum_repository:
    name: intel-dcap-local-repo
    description: Intel DCAP local RPM repository
    baseurl: "file://{{ intel_dcap_packages_directory }}/sgx_rpm_local_repo"
    enabled: no
    gpgcheck: no

- name: Install Intel DCAP packages
  yum:
    name: "{{ intel_dcap_packages }}"
    state: present
    enablerepo: intel-dcap-local-repo
  when: flc_enabled|bool
  retries: 10
  register: install_dcap
```

scripts/ansible/roles/linux/intel/vars/redhat/ootpa.yml
```diff
@@ -2,9 +2,9 @@
 # Licensed under the MIT License.
 
 ---
-intel_sgx_w_flc_driver_url: "https://download.01.org/intel-sgx/sgx-dcap/1.6/linux/distro/rhel8.0-Server/sgx_linux_x64_driver_1.33.bin"
```
CyanDevs, line 5: You can make these vars easier to handle by moving them to default variables and changing some parts to variables.

The advantage of having default vars over vars is the order of precedence of variables that Ansible uses, allowing some variables to overwrite others. E.g. Playbook vars will overwrite default vars, so one can quickly change the versions that are used by adding a playbook or runtime var without having to modify the role. To learn more, see https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

For example:
```
intel_driver_base_url: "https://download.01.org/intel-sgx/sgx-linux"
intel_driver_version: "2.10"
intel_driver_sgx_linux_version: "1.35"
intel_driver_sgx1_linux_version: "2.6.0_602374c"
intel_sgx_w_flc_driver_url: "{{ intel_driver_base_url }}/{{ intel_driver_version }}/distro/rhel8.1-server/sgx_linux_x64_driver_{{ intel_driver_sgx_linux_version }}.bin"
intel_sgx1_driver_url: "{{ intel_driver_base_url }}/{{ intel_driver_version }}/distro/rhel8.1-server/sgx_linux_x64_driver_{{ intel_driver_sgx1_linux_version }}.bin"
```
You could add more variables, but it depends on how you organized your roles and what tends to change or not.

scripts/ansible/roles/linux/intel/vars/redhat/ootpa.yml
radhikaj: Thanks for the explanation @rs-- and @CyanDevs

scripts/ansible/roles/linux/intel/tasks/redhat/sgx-packages-install.yml
rs--: Wow, very useful information, thanks! Each instance of the run should be on a clean system/the previous run should clean up, per the last step, so I think the 

```
  when: 
    - get_dcap_package.changed
    - get_dcap_package.dest
```

check should suffice.

scripts/ansible/roles/linux/intel/tasks/redhat/sgx-packages-install.yml
rs--: Instead of creating only if the file path doesn't exist, I've added another step to delete the directory if it does exist.

scripts/ansible/roles/linux/intel/vars/redhat/ootpa.yml
rs--: I like this solution, but given that we don't control the URLs, and the structure could change, I find that it's less error-prone to just copy and paste the url of the new drivers.

scripts/ansible/roles/linux/intel/tasks/redhat/sgx-packages-install.yml
CyanDevs: Make sure you access the key `dest` instead, so it's `path: {{ get_dcap_package.dest }}`
But yes the variable can be used by other tasks, so long as the tasks are after the registering of `get_dcap_package`

rs--: Yup, learnt this the hard way 😄

BRMcLaren (approved):

CyanDevs (approved):

PR #3755, opened 2020-11-25
Draft SGX2-EDMM driver interface

haitaohuang:


docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+## Basic EDMM flows
+
+SGX EDMM instructions support dynamic EPC page allocation/deallocation for enclaves and page property modification post-EINIT.  Following are the basic EDMM flows on which other more advanced usages of EDMM can be built.
+
+**Note:** The term "kernel" and "kernel space" are used in this document when general kernel space actions are described whether implemented in an OOT driver or in kernel tree. Kernel specific implementation details will be explicitly stated as "future kernel" or "kernel patches".  And implementation details such as OCalls issued by enclaves, ETRACK and IPI issued in kernel are generally omitted for brevity.
+
+- Allocate a new page at an address in ELRANGE of an enclave. 
+  - This can be an explicit syscall or triggered by #PF  when an unavailable page is accessed.
+  - Kernel issues EAUG for the page. All new pages should have RW permissions initially.
+  - The enclave then issues EACCEPT.
```
anakrish, line 19: How can the enclave be sure that 
- the kernel has indeed issues EAUG for the page?
- the a malicious kernel is not able trick the enclave in regards to the confidentiality or permissions of the new page?
- the contents of the page are what it wants to be?

MWShan: When the enclave issues EACCEPT or EACCEPTCOPY, and the instruction succeeds, then the enclave will have an assurance that the page has been EAUG'ed with the correct address and RW permission (EAUG always sets RW in EPCM).

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+Although interfaces described here are inspired to be as likely as possible a candidate for future kernel adoption, they are not intended to be a proposal for kernel implementation and are assumed to be implemented as an OOT driver. We hope from discussions enabled by this document, requirements and usage models can be identified to help shape future kernel interfaces. 
+
+## Basic EDMM flows
+
+SGX EDMM instructions support dynamic EPC page allocation/deallocation for enclaves and page property modification post-EINIT.  Following are the basic EDMM flows on which other more advanced usages of EDMM can be built.
+
+**Note:** The term "kernel" and "kernel space" are used in this document when general kernel space actions are described whether implemented in an OOT driver or in kernel tree. Kernel specific implementation details will be explicitly stated as "future kernel" or "kernel patches".  And implementation details such as OCalls issued by enclaves, ETRACK and IPI issued in kernel are generally omitted for brevity.
+
+- Allocate a new page at an address in ELRANGE of an enclave. 
+  - This can be an explicit syscall or triggered by #PF  when an unavailable page is accessed.
```
arai-fortanix, line 17: From the point of view of emulating a Linux environment, it would be helpful to have an option to only allocate pages on explicit syscalls and not implicitly when triggered by a page fault. A Linux application may have a SIGSEGV handler, and may handle a page fault in a way that doesn't result in a page mapping being added at that address.

haitaohuang: Can you specify the exact targeted call flow?
Note for large block of EPC allocation, mmap(..., MAP_POPULATE) proposed below can be used to instruct kernel to allocate EPC as immediately as possible. However, I think kernel should make the final decision on when exactly the physical EPC pages are committed.

jethrogb:
In the upstream kernel, you can't mmap() enclave memory until after a page is added to the enclave. The rest of the ELRANGE is generally mapped using anonymous non-fd-backed PROT_NONE memory. Given that, how are you planning to support the #PF flow?

arai-fortanix:
In case it's not clear, the scenario I care about is this:

- Some VA in the enclave is unmapped, and I never intend to create a mapping there. Let's call this address X.
- When code in the enclave faults at address X, the SIGSEGV handler in the user-space application makes an ecall to notify the enclave of the page fault.
- The enclave examines the faulting instruction and does some sort of emulation that doesn't involve mapping a page at address X. RIP is incremented past the faulting instruction. The enclave continues execution, without accessing address X.

The application might do the above multiple times, accessing address X repeatedly, and expecting to get page faults, but never actually maps anything at address X.

There may be other portions of the address space that we want backing added when referenced. Of course, the enclave needs some way to know which pages it should be calling EACCEPT for.

I don't have a preference for what the interfaces look like, but I think we need a way to mark portions of the address space as add-backing-pages-on-demand and other portions of the address space as do-not-add-backing-pages-on-demand. And we need the ability to change from one to the other. An example of this would be an application that lazily creates memory mappings. It may start with a region of memory mapped with PROT_NONE. In the enclave, this has to be a page that's not present in the enclave, because it's not possible to have pages that are in the enclave that are not accessible for reading. The application may take a fault on access to this page, and in the SIGSEGV handler, it may create a memory mapping at that address and proceed with execution. So we need to start by getting the SIGSEGV and not adding a backing page, because we have to deliver the SIGSEGV to the application and let it run its code that decides it's going to map something there. To perform the mapping, we need to add pages to the enclave at that address and fill them in with the mmap contents, before letting the applicaton proceed. If the application later munmaps that address, we need to be able to remove the pages from the enclave, and get SIGSEGVs for that VA again.

________________________________________

haitaohuang:
> In the upstream kernel, you can't mmap() enclave memory until after a page is added to the enclave. The rest of the ELRANGE is generally mapped using anonymous non-fd-backed PROT_NONE memory. Given that, how are you planning to support the #PF flow?

IIUC,  see change log of kernel patches,  v40 changed to allow mmap unpopulated area. Current kernel impl does not allow mmap or mprotect to change permissions beyond what's given in SecInfo during EADD IOCTCL. That obviously needs be extended for future EDMM usage. See a potential impl change described starting at line 58 in the draft.

haitaohuang:
@arai-fortanix, about this flow:
"
- Some VA in the enclave is unmapped, and I never intend to create a mapping there. Let's call this address X.
- When code in the enclave faults at address X, the SIGSEGV handler in the user-space application makes an ecall to notify the enclave of the page fault.
- The enclave examines the faulting instruction and does some sort of emulation that doesn't involve mapping a page at address X. RIP is incremented past the faulting instruction. The enclave continues execution, without accessing address X.
"
 Are you talking about libOS's implementation inside enclave?
If Linux mmap is not called to setup VMA and you only trigger SIGSEGV for LibOS-specific processing, then it would not trigger SGX subsystem in kernel to do anything.

arai-fortanix:
Yes, I'm talking about the libOS implementation inside of the enclave.

 If pages will only be added to the enclave for VA ranges that have been mmapped, then that will work fine for our use case. When we want to add mappings, we'll make an ocall to perform the host mmap, and then fault in an EACCEPT the pages. This seems fine for our purposes.

________________________________________

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * struct sgx_enclave_mprotect - parameter structure for the
+ *                                %SGX_IOC_ENCLAVE_MPROTECT ioctl
+ * @addr:	address of the memory to change protections
+ * @length:	length of the area.  This must be a multiple of the page size.
+ * @prot:	this must be or'ed of following:
+             PROT_READ
+             PROT_WRITE
+             PROT_EXEC
+             PROT_TRIM (new): change the page type to PT_TRIM, implies RW. User space should immediately EACCEPT, and then optionally call mprotect with PROT_EACCEPT.
```
raoulstrackx, line 92: What does "should immediately EACCEPT" mean here?

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+### Remarks
+
+Kernel should ensure that SGX instructions can succeed or catch and handle any fault.
+  - The kernel may maintain EPCM information on each page which includes access permission RWX, page types of PT_REG, PT_TRIM, PT_TCS.
+  - The kernel catches fault on EMODPR, EMODT and converts to error code returned to user space.
+
+The enclave run-time (or trusted run-time) may implement a parallel memory management structure which would provide information to the enclave on the enclave memory mappings.  The run-time can have a trusted API analogous to mmap which makes a call out of the enclave to issue the mmap  and then either perform EACCEPT on the pages and update the internal memory structures or configure the enclave to perform the EACCEPT when a #PF is delivered to the enclave.  With the outlined kernel interface, either implementation is possible.
+
```
raoulstrackx, line 110: Relying on page faults and error handling to add new pages to the enclave is ill-suited in some scenarios. The `mmap` approach can only provide a hint where a new page should be added, not enforce a specific location. A `syscall/ioctl` to issue an `eaug` instruction for a specific page would make much more sense.
Consider as an example an enclave written mostly in a memory safe language, with a small fraction of C/assembly. When such an enclave issues a page fault due to a page not being present, something bad must have happened, and the enclave should stop execution immediately. Letting it add new pages as default behavior may make successful exploitation of memory safety vulnerabilities much more likely.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
haitaohuang: Issue ENCLU.EACCEPT to finalize page type to TRIM

docs/DesignDocs/SGX2-EDMM-driver-interface.md
haitaohuang: First we are not proposing adding pages in error handling path. It is usual practice for kernel to allocate pages upon #PF. Kernel also can decide immediately to allocate EPC on mmap if it sees fit based on hint from user space and overall resource usage of the system. 

Second I don't see how direct EAUG would be better suited to mitigate the security problem you stated. You can use mmap/mprotect to specify  regions  allowed to add pages with specific access permissions allowed, which should effectively prevent ill-intended faults. Note, without EACCEPT, even if  a page is EAUG'd, app can't use the page. So user space can always control which page to be EACCEPT'd, further prevent the scenario you described.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+Future kernel may extend mmap and mprotect syscalls to support SGX EDMM usages. But we can't add/change syscall interfaces from an out-of-tree driver.  So, in this proposal for possible driver implementation, we reuse mmap for dynamic enclave memory mapping and expose a new IOCTL, sgx_enclave_mprotect, for enclave page modification. 
+ 
+## mmap
+
+After enclave is initialized (EINIT IOCTL done), the standard Linux mmap syscall can be used to create a new mapping configured for dynamically allocating enclave memory using EAUG. Following comments are specific to SGX EDMM usages, please refer to [mmap man page](https://man7.org/linux/man-pages/man2/mmap.2.html) for generic definitions.  
+
+### Remarks
+
+- To create a mapping for dynamic enclave memory allocation, mmap must be called with an open enclave file descriptor and with PROT_READ | PROT_WRITE for protection flags.
```
dimakuv, line 48: I think it's important to note here that only `PROT_READ | PROT_WRITE` combination is possible. Because this is the limitation of the Intel SGX architecture: the EAUG instruction adds a RW enclave page (and only RW).

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+Future kernel may extend mmap and mprotect syscalls to support SGX EDMM usages. But we can't add/change syscall interfaces from an out-of-tree driver.  So, in this proposal for possible driver implementation, we reuse mmap for dynamic enclave memory mapping and expose a new IOCTL, sgx_enclave_mprotect, for enclave page modification. 
+ 
+## mmap
+
+After enclave is initialized (EINIT IOCTL done), the standard Linux mmap syscall can be used to create a new mapping configured for dynamically allocating enclave memory using EAUG. Following comments are specific to SGX EDMM usages, please refer to [mmap man page](https://man7.org/linux/man-pages/man2/mmap.2.html) for generic definitions.  
+
+### Remarks
+
+- To create a mapping for dynamic enclave memory allocation, mmap must be called with an open enclave file descriptor and with PROT_READ | PROT_WRITE for protection flags. 
+  - Enclave must issue EACCEPT for the pages after mmap, then enclave may modify the content of the pages and extend/reduce permissions in secure way.
```
dimakuv, line 49: `must`? But this will fail if there was no previous `EAUG` from the SGX driver.

I think that you want to say is:
1. If the enclave asks for `mmap(MAP_POPULATE)`, then the enclave must issue EACCEPT after this mmap.
2. If the enclave asks for `mmap()`, then EACCEPT must be put in the SIGSEGV/SIGBUS sighandler of the trusted SGX runtime.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+    - enforces RW permissions for pages other than those loaded due to EADD or ECREAT.
+    - set up flags to track dynamic pages: type, permissions flag
+  - sgx_vma_mprotect
+    - Allow permissions changes to dynamic pages within limitations of OS policies, e.g.,
+	    - never allow WX
+	    - SELinux policy specific to SGX enclaves
+    - update flags for the dynamic pages
+
+## munmap 
+Calling munmap on an enclave page (dynamic allocated or not) has exactly the same effect of calling munmap on a regular RAM page. No sgx specific interface is needed. No behavior changes to current kernel space implementation.
```
dimakuv, line 69: I don't understand this part. The `munmap()` logic must be changed in two components:
1. The SGX driver must perform `PT_TRIM`.
2. The trusted SGX runtime (e.g. Graphene) must perform `EACCEPT` after the driver's `PT_TRIM`.

So what does it mean "no sgx specific interface is needed"? Not needed in which components?

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Remarks
+
+- Enclave memory mapings are shared (MAP_SHARED). The mappings in shared processes are kept alive and independently until the process exits
+    - munmap and closing file descriptors are not required for user space. A dead process automatically releases all mappings and file descriptors. 
+- Upon all enclave mappings are removed and file handles to the enclave are closed, either by explicit munmap/fclose syscalls or when all hosting apps exited:
+    - The kernel may mark its remaining pages are reclaimable and issue EREMOVE on them any time the kernel deems appropriate.
```
dimakuv, line 76: This whole remark is confusing. It is already done in the current SGX driver. There seems to be nothing new in the EDMM in this regard.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
dimakuv: @raoulstrackx Please see my comment under `Deallocate an existing page`. It is a similar flow.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+
+## mprotect IOCTL
+This IOCTL emulates the mprotect syscall with SGX specific extensions. In future kernel implementation, it could be mprotect or pkey_mprotect syscall with sgx extensions for the "prot" parameter.
```
dimakuv, line 79: > In future kernel implementation, it could be mprotect or pkey_mprotect syscall with sgx extensions for the "prot" parameter.

Is it really acceptable in the Linux kernel? Can we add new `prot` parameters to the mainline Linux?

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+};
+```
+
+### Remarks
+
+Kernel should ensure that SGX instructions can succeed or catch and handle any fault.
+  - The kernel may maintain EPCM information on each page which includes access permission RWX, page types of PT_REG, PT_TRIM, PT_TCS.
+  - The kernel catches fault on EMODPR, EMODT and converts to error code returned to user space.
+
+The enclave run-time (or trusted run-time) may implement a parallel memory management structure which would provide information to the enclave on the enclave memory mappings.  The run-time can have a trusted API analogous to mmap which makes a call out of the enclave to issue the mmap  and then either perform EACCEPT on the pages and update the internal memory structures or configure the enclave to perform the EACCEPT when a #PF is delivered to the enclave.  With the outlined kernel interface, either implementation is possible.
```
dimakuv, line 109: This is a good remark. To be honest, I think that the enclave run-time **must** implement some in-enclave memory manager. (For example, Graphene implements in-enclave VMA memory management.)

Also, I think the enclave run-time also **must** implement a rather complicated signal handling logic to correctly resolve on-demand enclave paging (to issue an `EACCEPT` instruction).

docs/DesignDocs/SGX2-EDMM-driver-interface.md
dimakuv: @raoulstrackx I would also recommend to check the original "Intel® Software Guard Extensions (Intel® SGX) Software
Support for Dynamic Memory Allocation inside an Enclave" paper. These flows are described very well there.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Example advanced flows
+
+More advanced flows can be implemented as combinations of the basic flows. Here we present a few examples.
```
dimakuv, line 113: I would also add an example flow on dynamically creating new threads. Then all interesting cases will be covered.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+1. Enclave calls mmap with MAP_GROWSDOWN for a stack region in enclave ELRANGE
+2. At some time later, enclave pushes to the top of the stack where no EPC page is populated yet, this results in #PF, causing enclave AEX.
+3. Kernel determines faulting address is in a stack region in enclave, EAUGs a page, injects a SIGBUS signal to enclave hosting process.
+4. User space untrusted side signal handler catches the SIGBUS and delivers it into enclave exception handler
+5. The enclave exception handler checks the faulting address against its record, determines the fault has happened in a stack area not yet EACCEPT'ed.
+6. Enclave issues EACCEPT, returns to untrusted signal handler of the hosting process, which returns to kernel fault handler.
+7. Kernel fault handler returns to enclave AEX address at which an ERESUME instruction is stored
+8. Enclave resumed and the original push instruction is retried and succeeds.
+
+
```
dimakuv, line 137: There is one corner-case that is missing from this document (but mentioned in the original paper).

Typically there are only two SSAs in SGX enclaves: one for normal enclave execution and one for exception handling. But with on-demand enclave pages (where #PF is transformed into a SIGSEGV/SIGBUS signal and sent to the enclave runtime), there can be a legit case of `in-enclave exception during first-exception handling`. E.g., an enclave may be inside the SIGILL signal handler (to e.g. emulate `rdtsc`), and it will try to access a lazily allocated page, which will trigger a #PF/SIGBUS. But since we are already in SSA=1, all SSAs are occupied, and the enclave cannot make progress, and the whole application hangs (or killed by the SGX driver).

So it looks like enclave runtime (SGX SDK, OpenEnclave, Graphene, ...) must adopt a model with 3 SSAs, where one SSA is just-in-case to resolve PFs. I would definitely add this into this document, it's a tricky issue (especially in multi-threaded apps).

dimakuv (changes requested):

docs/DesignDocs/SGX2-EDMM-driver-interface.md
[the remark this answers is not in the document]
haitaohuang: Thanks for feedback. This is in context of the other general PR for EDMM which pointed to that paper as reference. I'll add that too in this document and disclaimer that people should be familiar with it and related spec in the Intel SDM. This doc is to discuss kernel interface.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
haitaohuang: no, EACCEPT is required by design regardless and not necessarily done in sighandler (rather we should avoid that if possible). It is the basis of security for adding pages dynamically after EINIT.
1. If no EAUG by the time EACCEPT, #PF, kernel EAUG, EACCEPT resumes and reduces to the case 2
2. if EAUG happened previously, no #PF, EACCEPT just succeed if target page EPCM attributes matches what's given in EACCEPT operand for SECINFO in RBX, or causing #PF if any EPCM mismatch which will result in error returned from vDSO interfaces or signal raised if vDSO interface is not used (current Intel PSW impl)

docs/DesignDocs/SGX2-EDMM-driver-interface.md
haitaohuang: It's up to kernel community.  We don't know yet if it is acceptable. This document is to get feedback from user space implementers on possible design of the interface and identify functional requirements to kernel. Final signature/syntax of kernel interface may be different but hopeful is close and easy enough for us to adapt from user space side (similar to what we do now in enclave_common).

docs/DesignDocs/SGX2-EDMM-driver-interface.md
binxing: @dimakuv, in your case of SIGILL, the handler (SSA=1) would **not** emulate `rdtsc` directly. Rather, it pushes `SSA[0].rip` and then points it to the real handler for `rdtsc`, then it exits. The real handler will run in SSA=0.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+2. Enclave issues EACCEPT_COPY to copy trusted code from an existing EPC page to an unavailable page at an address inside that region
+   - Mechanism for enclave to establish the trustworthiness of new code is out-of-scope here.
+   - This results a #PF
+3. Kernel EAUG the page on #PF, set RW permissions in both EPCM and PTE
+4. The fault handler returns,  EACCEPT_COPY instruction retried, which sets RX permissions in EPCM specified in PageInfo operand.
+5. Enclave makes ocall which invokes mprotect syscall to change PTE permissions from RW to RX 
+
+### Lazy dynamic stack expansion
+An enclave can lazily expand its stacks as follows.
+1. Enclave calls mmap with MAP_GROWSDOWN for a stack region in enclave ELRANGE
```
binxing, line 128: IIUC, with MAP_GROWSDOWN set, the kernel expands the VMA downward upon a #PF within 64KB below its lower bound. In other words, MAP_GROWSDOWN takes effect only for #PFs outside of any VMA (i.e., #PF address must be unmapped). It doesn't sound like we are copying that behavior to SGX. What exactly does `MAP_GROWSDOWN` do in the context of SGX? Or what's the difference in behaviors between a VMA with `MAP_GROWSDOWN` set and a VMA without it?

docs/DesignDocs/SGX2-EDMM-driver-interface.md
haitaohuang: Regarding #of SSAs needed.
In our SDK, there is so called two-phase exception handling in TRTS side which uses only two SSAs.
First phase is the initial signal handler entry, always use SSA [1], do minimal to setup stack and entry point for second phase, and set IP to the second phase entry (https://github.com/intel/linux-sgx/blob/544159a94a59547a18b7bff1d28a94bb14053572/sdk/trts/trts_veh.cpp#L431), then returns to untrusted side sighandler, which immediately ERESUME. That would transfer control flow to second phase entry with SSA[0].
All exceptions only happen with SSA[0]. And SSA[1] is used only in phase 1. Exception/faults can happen recursively and stacked up exception->phase1->phase2->exception->phase1->...

At the end of second phase exception handling, control transferred to IP that caused the last exception (https://github.com/intel/linux-sgx/blob/544159a94a59547a18b7bff1d28a94bb14053572/sdk/trts/trts_veh.cpp#L278)

dimakuv: @binxing and @haitaohuang Thanks for reminding me this!

Yes, the same logic applies to Graphene. My bad, I forgot how exception handling works in SGX runtimes :)

So then there is no real issue. If the "SSA1" exception handler of the SGX runtime is correctly written, then it will **never** access a lazily allocated enclave page (in fact, I don't think the exception handler touches any memory at all except for SSA frames), then the issue I described in the previous comment does not exist.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
dimakuv: I think the Linux community won't add SGX-specific flags like `PROT_TRIM` and `PROT_TCS`.

Don't you want to untangle these features? What about this interface:
1. For RWX permission change, we re-use normal `mprotect()` syscall.
2. For `PROT_TRIM` change (unmap enclave page), we re-use normal `munmap()` syscall.
3. For `PROT_TSC` change (re-purpose an enclave page for threading), we add a new IOCTL `SGX_IOC_ENCLAVE_CREATETHREAD`.

And then we don't need this contrived `SGX_IOC_ENCLAVE_MPROTECT`.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+ *                                %SGX_IOC_ENCLAVE_MPROTECT ioctl
+ * @addr:	address of the memory to change protections
+ * @length:	length of the area.  This must be a multiple of the page size.
+ * @prot:	this must be or'ed of following:
+             PROT_READ
+             PROT_WRITE
+             PROT_EXEC
+             PROT_TRIM (new): change the page type to PT_TRIM, implies RW. User space should immediately EACCEPT, and then optionally call mprotect with PROT_EACCEPT.
+             PROT_TCS (new): change the page type to PT_TCS
+             PROT_EACCEPT (new): notify kernel EACCEPT is done for PT_TRIM type changes. This can be optional. Kernel will try EREMOVE on PT_TRIM pages in best effort on its own if user space does not notify EACCEPT.
```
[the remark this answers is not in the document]
haitaohuang: Yes. it's kind of a hint to kernel that the page is free. It might be useful for optimization purpose.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
haitaohuang: For all practical purposes, the page needs be RW so enclave can copy content into it before using it as EAUG always zero out the page. But I hesitant to say it's an arch limit because SDM does say EAUG can accept a PAGEINFO  to turn its EPCM permission to RX or RO immediately.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
haitaohuang: Robust Linux sighandler in a lib/runtime is challenging :-)
A few pointers:
1) Using vDSO interface for EENTER/EXIT/AEX handling would avoid some complexity in implementing sighandler. And the vDSO interface was proposed to mitigate exactly this issue.
2) In most usages you can avoid doing EACCEPT in signal handling path. 
3) In case you need do EACCEPT in trusted side signal handler, e.g., the lazy stack expansion, implementing something similar to Intel SDK "two-phase handler" would help.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
haitaohuang: @binxing and @dimakuv, those are good points. 
The desired behavior is for kernel to EAUG all pages between the current low end of the populated region and the faulting address if an VMA is marked with GROWSDOWN, and faulting address is below the low end of populated.
However, I also found that MAP_GROWSDOWN is not allowed for MAP_SHARED and/or fd backed mapping. So this may require some discussion in kernel community to change the current behavior, or we need come up with a different flag.

dimakuv: >However, I also found that MAP_GROWSDOWN is not allowed for MAP_SHARED and/or fd backed mapping. 

I didn't look at the current Linux code but it looks like:
1. Combination of `MAP_GROWSDOWN | MAP_SHARED` is legal (if there is no fd backing, i.e. if there is also `MAP_ANONYMOUS`).
2. Combination of `MAP_GROWSDOWN` and a backing fd is illegal. Intuitively it makes sense -- `MAP_GROWSDOWN` is purely for stacks, and a file-backed stack sounds like nonsense.

But going back to the original flow: does the Linux kernel even need to know about `MAP_GROWSDOWN`? This can be fully emulated by the enclave runtime + Intel SGX driver, I believe. Probably there will be a need for a new mprotect flag/ioctl to pass the enclave's intention of growing the enclave stack to the SGX driver... But we already do similar things to circumvent `munmap()` (I mean `PROT_TRIM`).

Also, I think this "lazy dynamic stack extension" is a very rare use case for SGX enclaves. We haven't observed any Graphene applications in the wild that wanted to extend the stack at runtime (we simply allocate a large-enough stack at runtime, and it always works).

docs/DesignDocs/SGX2-EDMM-driver-interface.md
haitaohuang: Can u elaborate the use case? Do it need EDMM to support?

docs/DesignDocs/SGX2-EDMM-driver-interface.md
haitaohuang: After reading more carefully the spec on TRIM and EACCEPT, I have come to conclusion user space has to signal EACCEPT is done to kernel before kernel can EREMOVE the page. Will update this part.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,137 @@
...
+**Note:** The term "kernel" and "kernel space" are used in this document when general kernel space actions are described whether implemented in an OOT driver or in kernel tree. Kernel specific implementation details will be explicitly stated as "future kernel" or "kernel patches".  And implementation details such as OCalls issued by enclaves, ETRACK and IPI issued in kernel are generally omitted for brevity.
+
+- Allocate a new page at an address in ELRANGE of an enclave. 
+  - This can be an explicit syscall or triggered by #PF  when an unavailable page is accessed.
+  - Kernel issues EAUG for the page. All new pages should have RW permissions initially.
+  - The enclave then issues EACCEPT.
+- Deallocate an existing page
+  - Enclave signals via a syscall to kernel that a page is no longer in use. 
+  - Kernel issues EMODT to change page type to PT_TRIM 
+  - The enclave issues EACCEPT
```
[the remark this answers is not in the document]
haitaohuang: This section is intended to serve a problems statement. I did not want to jump into implementation of particular flow. Understood the gap here needs be addressed. So added diagrams for basic flow after interface spec

docs/DesignDocs/SGX2-EDMM-driver-interface.md
haitaohuang: how exactly threads created is runtime thread library implementation. But I added a sequence diagram for TCS creation. Thread libraries need to associate the newly created TCS with OS thread and  use it for  ecalls in that thread

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,170 @@
+SGX EDMM Driver Interface Design
+=====================================
+
+## Motivation
+
+This document describes possible driver interfaces to facilitate discussions among SGX runtime implementors (e.g., https://github.com/openenclave/openenclave/pull/3639) on supporting different SGX EDMM flows.
+
+Although interfaces described here are inspired to be as likely as possible a candidate for future kernel adoption, they are not intended to be a proposal for kernel implementation and are assumed to be implemented as an OOT driver. We hope from discussions enabled by this document, requirements and usage models can be identified to help shape future kernel interfaces.
```
dthaler, line 8: Regarding "kernel" adoption, is this applicable to any OS (Linux, Windows, etc.) or only Linux?  Clarify.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,170 @@
...
+For design and implementation of current SGX1 support in upstream kernel (merged in 5.11RC), please refer to [this patch series](https://lwn.net/Articles/837121/)
+
+## Basic EDMM flows
+
+SGX EDMM instructions support dynamic EPC page allocation/deallocation for enclaves and page property modification post-EINIT.  Following are the basic EDMM flows on which other more advanced usages of EDMM can be built.
+
+**Note:** The term "kernel" and "kernel space" are used in this document when general kernel space actions are described whether implemented in an OOT driver or in kernel tree. Kernel specific implementation details will be explicitly stated as "future kernel" or "kernel patches".  And implementation details such as OCalls issued by enclaves, ETRACK and IPI issued in kernel are generally omitted for brevity.
+
+- Allocate a new page at an address in ELRANGE of an enclave. 
+  - This can be an explicit syscall or triggered by #PF  when an unavailable page is accessed.
```
dthaler, line 23: Editorial: expand "PF" on first use
```suggestion
  - This can be an explicit syscall or triggered by a page fault (#PF) when an unavailable page is accessed.
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,170 @@
...
+- [HASP@ISCA 2016: 11:1-11:9](https://caslab.csl.yale.edu/workshops/hasp2016/HASP16-17.pdf)
+- [Intel SDM Vol.4, Ch.36-42](https://software.intel.com/content/www/us/en/develop/articles/intel-sdm.html)
+
+For design and implementation of current SGX1 support in upstream kernel (merged in 5.11RC), please refer to [this patch series](https://lwn.net/Articles/837121/)
+
+## Basic EDMM flows
+
+SGX EDMM instructions support dynamic EPC page allocation/deallocation for enclaves and page property modification post-EINIT.  Following are the basic EDMM flows on which other more advanced usages of EDMM can be built.
+
+**Note:** The term "kernel" and "kernel space" are used in this document when general kernel space actions are described whether implemented in an OOT driver or in kernel tree. Kernel specific implementation details will be explicitly stated as "future kernel" or "kernel patches".  And implementation details such as OCalls issued by enclaves, ETRACK and IPI issued in kernel are generally omitted for brevity.
```
dthaler, line 20: expand IPI on first use.  "inter-process interrupts"?
```suggestion
**Note:** The term "kernel" and "kernel space" are used in this document when general kernel space actions are described whether implemented in an OOT driver or in kernel tree. Kernel specific implementation details will be explicitly stated as "future kernel" or "kernel patches".  And implementation details such as OCalls issued by enclaves, ETRACK and inter-process interrupts (IPIs) issued in kernel are generally omitted for brevity.
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,170 @@
...
+
+![SGX2 deallocation flow](images/SGX2-trim.svg)
+
+### Permission Changes
+
+![SGX2 permissions change flow](images/SGX2-perms.svg)
+
+**Notes:**
+- EACCEPT is needed for enclave to ensure untrusted runtime and OS indeed invoke EMODPR and EPCM permissions are set as expected.
+- It is assumed that both OS and enclave keep track of page permissions. However, it is possible to avoid that as the current [Intel SDK implementation](https://github.com/intel/linux-sgx/blob/master/sdk/trts/trts_ecall.cpp#L561) demostrates.
```
CodeMonkeyLeet, line 137: Can you elaborate on the premise on why both the OS and enclave would be keeping track of page permissions, as well as why the existing Intel SGX SDK implementation elides this? It seems to hinge on the partitioning of responsibilities between the OS for EMODPR for restrictions and enclave for EMODPE requiring EACCEPT, but it's not clear that the enclave is not tracking page permissions in that case.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
haitaohuang: clarified in title and introduction

docs/DesignDocs/SGX2-EDMM-driver-interface.md
haitaohuang: For OS to know when to do EMODPR, it needs to keep track of EPCM permissions to compare with target. But it is also OK to just issue EMODPR with the target permissions.
Similarly in enclave, to avoid extra overhead of EMODPR, ocall, it'd help to keep track of EPCM permissions of each page. 
I updated with pseudo code to show possible implementation in enclave without tracking permissions.

Regardless of implementation, the mprotect interface should have worked.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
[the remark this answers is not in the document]
haitaohuang: updated introduction to give more context and added sequence diagrams to clarify the call flows. Hope that addresses your concern.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,181 @@
...
+- [HASP@ISCA 2016: 11:1-11:9](https://caslab.csl.yale.edu/workshops/hasp2016/HASP16-17.pdf)
+- [Intel SDM Vol.4, Ch.36-42](https://software.intel.com/content/www/us/en/develop/articles/intel-sdm.html)
+
+For design and implementation of current SGX1 support in upstream Linux kernel (merged in 5.11RC), please refer to [this patch series](https://lwn.net/Articles/837121/)
+
+## Basic EDMM flows
+
+SGX EDMM instructions support dynamic EPC page allocation/deallocation for enclaves and page property modification post-EINIT.  Following are the basic EDMM flows on which other more advanced usages of EDMM can be built.
+
+**Note:** This document is Linux specific. The term "kernel" and "kernel space" are used in this document when general Linux kernel space actions are described whether implemented in an OOT driver or in kernel tree. Kernel specific implementation details will be explicitly stated as "future kernel" or "kernel patches".  And implementation details such as OCalls issued by enclaves, ETRACK and inter-process interrupts (IPIs) issued in kernel are generally omitted for brevity.
```
jethrogb, line 20:
```suggestion
**Note:** This document is Linux specific. The term "kernel" and "kernel space" are used in this document when general Linux kernel space actions are described whether implemented in an OOT driver or in kernel tree. Kernel specific implementation details will be explicitly stated as "future kernel" or "kernel patches".  And implementation details such as OCalls issued by enclaves, ETRACK and inter-processor interrupts (IPIs) issued in kernel are generally omitted for brevity.
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
dimakuv: The use case is just a normal `mmap(MAP_ANONYMOUS)` done by the enclavized application. The application expects normal Linux semantics which is to provide zeroed-out pages. Since the EAUG hardware instruction already does the zero-out for us, we don't need to "copy content" additionally -- we got it from the hardware for free.

But this is anyway a hypothetical discussion, since the hardware dictates the use of RW :)

docs/DesignDocs/SGX2-EDMM-driver-interface.md
dimakuv: OK.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
dimakuv: Thanks!

docs/DesignDocs/SGX2-EDMM-driver-interface.md
dimakuv: Agreed.

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,170 @@
...
+3. Kernel EAUG the page on #PF, set RW permissions in both EPCM and PTE
+4. The fault handler returns,  EACCEPT_COPY instruction retried, which sets RX permissions in EPCM specified in PageInfo operand.
+5. Enclave makes ocall which invokes mprotect syscall to change PTE permissions from RW to RX 
+
+### Lazy dynamic stack expansion
+An enclave can lazily expand its stacks as follows.
+1. Enclave calls mmap with MAP_GROWSDOWN for a stack region in enclave ELRANGE
+2. At some time later, enclave pushes to the top of the stack where no EPC page is populated yet, this results in #PF, causing enclave AEX.
+3. Kernel determines faulting address is in a stack region in enclave, EAUGs a page, injects a SIGBUS signal to enclave hosting process.
+4. User space untrusted side signal handler catches the SIGBUS and delivers it into enclave exception handler
```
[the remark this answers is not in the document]
haitaohuang: updated to include fault handling details and vDSO interface

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,256 @@
...
+| Page Protection mismatch in EPCM| #PF where PFEC.SGX=1  | invoke user handler   | Vector = #PF<br>MADDR = faulting address <br>ERRCD(PFEC).SGX=1<br> ERRCD(PFEC).W/R or I/D does not match protections    |  Check if EMODPE in progress and wait if so, <br>or call custom handler or abort  |
+| Access Page with EPCM.Pending  | #PF where PFEC.SGX=1   | invoke user handler   | Vector = #PF<br>MADDR = faulting address<br>ERRCD(PFEC).SGX=1  | Enclave is operating on the page, the handler should wait on the operation to complete. |
+| Access Page with EPCM.Modified | #PF where PFEC.SGX=1   | invoke user handler   | Vector = #PF<br>MADDR = faulting address<br>ERRCD(PFEC).SGX=1  | Enclave is operating on the page, the handler should wait on the operation to complete. |
+| Access Page with type PT_TRIM  | #PF where PFEC.SGX=1  | invoke user handler  | Vector = #PF<br>MADDR = faulting address<br>ERRCD(PFEC).SGX=1  | Memory Access Error - accessing a TRIM'ed page, call custom handler or abort   |
```
jethrogb, line 233: What about the condition where the enclave has disappeared (due to power state change)?

haitaohuang: When the enclave is lost, all ENCLS and ENCLU operations on the enclave will fault. And all of them will follow error paths to an eventual fault on  EENTER reported by kernel to user space. And app should cleanup and reload the enclave. Added a couple of cases into the table and a description of how kernel handle error cases in fault handling.

jethrogb (approved):

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
+SGX EDMM Linux Driver Interface Design
+=====================================
+
+## Motivation
+
+This document describes possible Linux driver interfaces to facilitate discussions among SGX runtime implementors (e.g., https://github.com/openenclave/openenclave/pull/3639) on supporting different SGX EDMM flows.
+
+Although interfaces described here are inspired to be as likely as possible a candidate for future Linux kernel adoption, they are not intended to be a proposal for kernel implementation and are assumed to be implemented as an OOT driver. We hope from discussions enabled by this document, requirements and usage models can be identified to help shape future kernel interfaces.
```
dthaler, line 8: split source lines (throughout file) to make git diffs easier.  Also, expand "OOT" acronym: out-of-tree (OOT)

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+
+Without losing generality, this document may describe how upper layer user space components would use the interfaces. However, details of design and implementation of those components are intentionally left out. The PR mentioned above would provide more contexts on other user space components and their relationships. Further, for those who may want to learn basic principles behind Intel(R) SGX EDMM instructions and how they are typically used, please refer to following references:
+- [HASP@ISCA 2016: 11:1-11:9](https://caslab.csl.yale.edu/workshops/hasp2016/HASP16-17.pdf)
+- [Intel SDM Vol.4, Ch.36-42](https://software.intel.com/content/www/us/en/develop/articles/intel-sdm.html)
+
+For design and implementation of current SGX1 support in upstream Linux kernel (merged in 5.11RC), please refer to [this patch series](https://lwn.net/Articles/837121/)
```
dthaler, line 14:
```suggestion
For design and implementation of current SGX1 support in the upstream Linux kernel (merged in 5.11RC), please refer to [this patch series](https://lwn.net/Articles/837121/)
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
+SGX EDMM Linux Driver Interface Design
```
dthaler, line 1: Since this is linux specific, please change the filename to have "linux" in the filename

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+- Extend EPCM permissions of a page, e.g., R->RW/RX
+  - Enclave issues EMODPE for the page
+  - Enclave requests via a syscall that the kernel update the page table permissions to match. 
+  - Kernel modifies permissions in PTE 
+- Reduce EPCM permissions of a page, e.g. RW/RX->R
+  - Enclave requests that the kernel restrict the permissions of an EPC page
+  - Kernel performs EMODPR,  updates page tables to match the new EPCM permissions, 
+  - Enclave issues EACCEPT
+  
+**Note:** Flows related to CET support inside enclave will be considered as a future enhancement.
```
dthaler, line 44:
```suggestion
**Note:** Flows related to Control-Flow Enforcement Technology (CET) support inside an enclave will be considered as a future enhancement.
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+
+Future kernel may extend mmap and mprotect syscalls to support SGX EDMM usages. But we can't add/change syscall interfaces from an out-of-tree driver.  So, in this proposal for possible driver implementation, we reuse mmap for dynamic enclave memory mapping and expose a new IOCTL, sgx_enclave_mprotect, for enclave page modification.
```
dthaler, line 46:
```suggestion
Future kernels may extend mmap and mprotect syscalls to support SGX EDMM usages. But we can't add/change syscall interfaces from an out-of-tree driver.  So, in this proposal for possible driver implementation, we reuse mmap for dynamic enclave memory mapping and expose a new IOCTL, sgx_enclave_mprotect, for enclave page modification. 
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+    - enforces RW permissions for pages other than those loaded due to EADD or ECREATE.
+    - set up flags to track dynamic pages: type, permissions flag
+  - sgx_vma_mprotect
+    - Allow permissions changes to dynamic pages within limitations of OS policies, e.g.,
+	    - never allow WX
+	    - SELinux policy specific to SGX enclaves
+    - update flags for the dynamic pages
+
+## munmap 
+Calling munmap on an enclave page (dynamic allocated or not) has exactly the same effect of calling munmap on a regular RAM page. No sgx specific interface is needed. No behavior changes to current kernel space implementation.
```
dthaler, line 75:
```suggestion
Calling munmap on an enclave page (dynamic allocated or not) has exactly the same effect as calling munmap on a regular RAM page. No SGX specific interface is needed. There are no behavior changes to the current kernel space implementation.
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+
+### Remarks
+
+- Enclave memory mapings are shared (MAP_SHARED). The mappings in shared processes are kept alive and independently until the process exits
+    - munmap and closing file descriptors are not required for user space. A dead process automatically releases all mappings and file descriptors. 
+- Upon all enclave mappings are removed and file handles to the enclave are closed, either by explicit munmap/fclose syscalls or when all hosting apps exited:
+    - The kernel may mark its remaining pages are reclaimable and issue EREMOVE on them any time the kernel deems appropriate.
+
+## mprotect IOCTL
+This IOCTL emulates the mprotect syscall with SGX specific extensions. In future kernel implementation, it could be mprotect or pkey_mprotect syscall with sgx extensions for the "prot" parameter.
```
dthaler, line 85:
```suggestion
This IOCTL emulates the mprotect syscall with SGX specific extensions. In a future kernel implementation, it could be mprotect or pkey_mprotect syscall with SGX extensions for the "prot" parameter.
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+
+```
+#define SGX_IOC_ENCLAVE_MPROTECT _IOW(SGX_MAGIC, 0x06, struct sgx_enclave_mprotect)
+/**
+ * struct sgx_enclave_mprotect - parameter structure for the
+ *                                %SGX_IOC_ENCLAVE_MPROTECT ioctl
+ * @addr:	address of the memory to change protections
+ * @length:	length of the area.  This must be a multiple of the page size.
+ * @prot:	this must be or'ed of following:
```
dthaler, line 94:
```suggestion
 * @prot:	this must be an or'ed set of any of the following bits:
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+struct sgx_enclave_mprotect {
+	__u64	addr;
+	__u64	length;
+	__u64	prot;
+};
+```
+
+### Remarks
+
+Kernel should ensure that SGX instructions can succeed or catch and handle any fault.
```
dthaler, line 111:
```suggestion
The kernel should ensure that SGX instructions can succeed or catch and handle any fault.
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+
+![SGX2 deallocation flow](images/SGX2-trim.svg)
+
+### Permission Changes
+
+![SGX2 permissions change flow](images/SGX2-perms.svg)
+
+**Notes:**
+- EACCEPT is needed for enclave to ensure untrusted runtime and OS indeed invoke EMODPR and EPCM permissions are set as expected.
+- It is assumed that both OS and enclave keep track of page permissions. However, it is possible for enclave to avoid that with implementation like this:
```
dthaler, line 137:
```suggestion
- EACCEPT is needed for an enclave to ensure that the untrusted runtime and OS indeed invoke EMODPR and EPCM permissions are set as expected.
- It is assumed that both the OS and enclave keep track of page permissions. However, it is possible for the enclave to avoid that with an implementation like this:
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+
+```
+//Change page permission to perms_target in EPCM without remember previous permissions.
+//The tradeoff here is possibly more ocall, emodpr, emodpe, eaccept than necessary.
```
dthaler, line 141:
```suggestion
// Change page permission to perms_target in EPCM without remembering previous permissions.
// The tradeoff here is possibly more ocall, emodpr, emodpe, eaccept than necessary.
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+![SGX2 TCS allocation flow](images/SGX2-tcs.svg)
+
+
+## Example advanced flows
+
+More advanced flows can be implemented as combinations of the basic flows. Here we present a few examples. 
+
+### Dynamic code loading
+
+To load dynamic code after EINIT, the enclave has to have verify the code to be trustworthy. The mechanism
```
dthaler, line 164:
```suggestion
To load dynamic code after EINIT, the enclave has to verify the code to be trustworthy. The mechanism
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+for an enclave to establish trustworthiness of the new code is out-of-scope of this document.
+
+Assuming a new code page is verified to be trusted and stored at an existing enclave page, then there could
+be many ways for an enclave to load the trusted code to a new executable page. For example, in SGX1 environment
```
dthaler, line 168:
```suggestion
be many ways for an enclave to load the trusted code to a new executable page. For example, in an SGX1 environment
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+without EDMM support, the enclave can reserve RWX regions and load trusted code directly into those regions.
+It is straightforward but not flexible or efficient use of EPC. Additionally the requirement of a RWX region
```
dthaler, line 170:
```suggestion
It is a straightforward but not flexible or efficient use of EPC. Additionally the requirement of an RWX region
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+goes against security policies of not running code in writable pages. With EDMM, the EACCEPTCOPY instruction
+allows an enclave copy code to a pending page and reset EPCM permissions to RX at the same time, thus provides
```
dthaler, line 172:
```suggestion
allows an enclave to copy code to a pending page and reset EPCM permissions to RX at the same time, thus provides
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+3. Kernel determines faulting address is in a stack region in enclave, EAUGs a page, invokes user space handler via vDSO.
+4. The user space handler delivers it into enclave exception handler
+5. The enclave exception handler checks the faulting address against its record, determines the fault has happened in a stack area not yet EACCEPT'ed.
+6. Enclave issues EACCEPT, returns to untrusted user handler of the hosting process, which returns to kernel fault handler.
+7. Kernel fault handler returns to enclave AEX address at which an ERESUME instruction is stored
+8. Enclave resumed and the original push instruction is retried and succeeds.
+
+## Exception Handling
+
+This section focuses on changes around \#PF handling which is affected by the new page states (i.e. states in EPCM) introduced by SGX/EDMM, along with the mechanisms for handling exceptions in enclaves.
```
dthaler, line 207:
```suggestion
This section focuses on changes around \#PF handling which is affected by the new page states (i.e., states in EPCM) introduced by SGX/EDMM, along with the mechanisms for handling exceptions in enclaves. 
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+
+As with all non-enclave exception scenarios, the kernel fault handler registered in the Interrupt Descriptor Table (IDT) would be the first in line to handle exceptions for AEX, and it needs to either handle it in kernel space, or if it can't handle, invoke user space exception handler. In both cases, after handlers return, control is tranferred to AEP trampoline, which enventually invokes ERESUME to reenter enclave. 
+
+Current kernel implementation (in release 5.11) can invoke user space exception handler in two ways depending on how EENTER and AEP trampoline are managed:
```
dthaler, line 213:
```suggestion
The current kernel implementation (in release 5.11) can invoke the user space exception handler in two ways depending on how EENTER and AEP trampoline are managed:
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+
+  1. Direct EENTER in runtime: the user space runtime manages EENTER, AEP trampoline directly and use Linux signal APIs to register and handle exceptions.
+  2. vDSO interface: the user space invokes [__vdso_sgx_enter_enclave](https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/tree/arch/x86/include/uapi/asm/sgx.h?h=x86/sgx#n124), passing in a callback for exception handling, and the vDSO implementation manages EENTER and AEP trampoline. 
+
+   The direct EENTER method requires signal handling in runtime library which is known to be challenging in Linux environment. Therefore, the vDSO interface is preferred and assumed in following discussion. (Runtime implementing direct EENTER method would have similar flow but the callbacks from vDSO are replaced with Linux signals.) For more details about the new SGX vDSO interface please refer to documentation in the [kernel header file](https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/tree/arch/x86/include/uapi/asm/sgx.h?h=x86/sgx#n124). General sequence is as follows:
```
dthaler, line 218:
```suggestion
   The direct EENTER method requires signal handling in runtime library which is known to be challenging in a Linux environment. Therefore, the vDSO interface is preferred and assumed in the following discussion. (Runtime implementing direct EENTER method would have similar flow but the callbacks from vDSO are replaced with Linux signals.) For more details about the new SGX vDSO interface please refer to documentation in the [kernel header file](https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/tree/arch/x86/include/uapi/asm/sgx.h?h=x86/sgx#n124). General sequence is as follows:
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+
+In addition, a running enclave can lose EPC context due to power events (S3/S4 transitions) or VM being suspended. A page fault on EENTER instruction (either at an initial ecall or at re-entering enclave for exception handling) results in those cases, and the user handler would receive a callback from vDSO with run.function = EENTER. 
+
+This table summarizes kernel, vDSO, user handler actions in different fault scenarios related to enclave operations. All exceptions considered here happen inside enclave causing AEX, or at EENTER/ERESUME, so the kernel will convert them to the synchronous callbacks thru vDSO interface as needed.
```
dthaler, line 240:
```suggestion
This table summarizes kernel, vDSO, and user handler actions in different fault scenarios related to enclave operations. All exceptions considered here happen inside enclave causing AEX, or at EENTER/ERESUME, so the kernel will convert them to the synchronous callbacks thru vDSO interface as needed. 
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+* Permissions: Combination of Read, Write, Execute
+* Page Type: SGX page type of pages in the region - PT_TCS, PT_REG, or PT_TRIM
+* State: the state of the region.  The state may indicate that the region is in transition.  For example is is changing page type or permissions.  
+* Table of information about the EACCEPT state of each page in the region.  This may be a temporary structure which keeps track of pages which are EACCEPTed for operations requiring EACCEPT.  This can ensure that the enclave does not EACCEPT a page twice.  For example, when a page is EAUG'ed to an enclave linear address, the enclave should only EACCEPT that page once.  If the enclave could be convinced to EACCEPT the page twice, then the OS can potentially EAUG two pages at the same enclave linear address and freely swap them by modifying PTEs.
+
+Enclaves should prevent two threads from simultaneously operating on the same region, e.g, trying to EMODPE on a page while permission change is in progress in another thread.  One way to ensure this is to use some lock/synchronization mechanism to protect the state of each region, have the second thread wait if page is in transition state.
```
dthaler, line 271:
```suggestion
Enclaves should prevent two threads from simultaneously operating on the same region, e.g., trying to EMODPE on a page while permission change is in progress in another thread.  One way to ensure this is to use some lock/synchronization mechanism to protect the state of each region, have the second thread wait if the page is in a transition state. 
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+| ERRCD(PFEC).SGX=1  | Not In-Transition | Error in run-time or kernel | Depending on run-time design, the enclave should not encounter this. |
+| ERRCD(PFEC).SGX=1  | In-Transition | Page is being accessed during transition | If future protections/page-type will allow access then pend on Lock/Mutex for region, else call exception handlers, abort if not handled |
+
+## Debugger Support
+
+The EDMM flows do not affect how debugger read/write enclave memory by EDBGRD/EDBGWR.  However, the DBGOPTIN (i.e. bit 0) of TCS.FLAGS must be set to enable hardware breakpoints and allow single-stepping inside the enclave thread entered with that TCS. Therefore, for TCS pages added dynamically using EDMM instructions, debuggers are required to trap TCS creation flow in order to set DBGOPTIN. For GDB on Linux, the debugger scripts can collaborate with runtime to set DBGOPTIN. One possible implementation could be described as below.
```
dthaler, line 285:
```suggestion
The EDMM flows do not affect how debugger read/write enclave memory by EDBGRD/EDBGWR.  However, the DBGOPTIN (i.e., bit 0) of TCS.FLAGS must be set to enable hardware breakpoints and allow single-stepping inside the enclave thread entered with that TCS. Therefore, for TCS pages added dynamically using EDMM instructions, debuggers are required to trap TCS creation flow in order to set DBGOPTIN. For GDB on Linux, the debugger scripts can collaborate with runtime to set DBGOPTIN. One possible implementation could be described as below.
```

docs/DesignDocs/SGX2-EDMM-driver-interface.md
```diff
@@ -0,0 +1,296 @@
...
+- GDB startup script sets a breakpoint on that empty function to receive the debug interrupt.
+- Once the breakpoint has been hit, GDB script extracts the address of the TCS page, and sets DBGOPTIN for that TCS.
+- GDB resumes the interrupted application.
+- From now on, hardware breakpoints and single stepping are allowed inside enclave threads  entered with the newly created TCS
+
+For Windows, similar collaboration between debugger and runtime can be implemented using exceptions instead of breakpoints on empty functions.
+
+For runtimes using EDDM to load dynamic modules into enclave after EINIT, the runtime needs to signal module loading events to the debugger so that the debugger can load additional symbols for those modules. That can also be implemented using exceptions or pre-defined breakpoints.
+
+**Note:** Kernel does not fixup Debug Exceptions (#DB) and Breakpoints (#BP).
```
dthaler, line 296:
```suggestion
**Note:** The kernel does not fix up Debug Exceptions (#DB) and Breakpoints (#BP). 
```

dthaler (commented):
see editorial nits

alexthomas1:
Were there any updates to this? Has adopting the EDMM been abandoned?

mingweishih:
> Were there any updates to this? Has adopting the EDMM been abandoned?

EDMM support in OE will be using https://github.com/intel/sgx-emm, where this design doc is hosted. However, the support is still a pending item.

alexthomas1:
> > Were there any updates to this? Has adopting the EDMM been abandoned?
> 
> EDMM support in OE will be using https://github.com/intel/sgx-emm, where this design doc is hosted. However, the support is still a pending item.

Thank you for the reply. Any update on when this will be supported? Is there a WIP branch available?

mingweishih:
> > > Were there any updates to this? Has adopting the EDMM been abandoned?
> > 
> > 
> > EDMM support in OE will be using https://github.com/intel/sgx-emm, where this design doc is hosted. However, the support is still a pending item.
> 
> Thank you for the reply. Any update on when this will be supported? Is there a WIP branch available?

There is a branch that I prototyped a while back: https://github.com/mingweishih/openenclave/tree/edmm. Just for reference, no guarantee for the branch to work.

radhikaj:
Thanks for the contribution. I am closing it and reopening it when someone will address the comments and can merge it.

PR #3758, opened 2020-12-01
Remove OpenSSL experimental flag

### LINKED ISSUE #3745: cannot find a header file in building with OpenSSL

mingweishih:
This PR moves the OpenSSL out of experimental features. The changes include
- Set the `BUILD_OPENSSL` to `ON` by default for SGX build (OP-TEE is not supported yet)
- Remove the dedicated OpenSSL configuration from Jenkins pipeline (now OpenSSL will be built on every SGX configurations).
- Fix tests to make them work on Windows build
- Fix the bug where build dependency is not correctly enforced, can be reproduced when building OE with single thread (Fixes #3745)
- Fix bugs in the tls_e2e tests that cause failure on the gcc release build.
- Update documents

### DIFF
--- a/3rdparty/openssl/CMakeLists.txt
+++ b/3rdparty/openssl/CMakeLists.txt
@@ -292,7 +292,8 @@ ExternalProject_Add(
   INSTALL_COMMAND "")
 
 add_dependencies(openssl_includes openssl_configure)
-add_dependencies(openssl_generated openssl_configure)
+add_dependencies(openssl_copied openssl_configure)
+add_dependencies(openssl_generated openssl_includes)
 add_dependencies(openssl_generated openssl_copied)
 
 # Add opensslcrypto (the libcrypto in openssl).
--- a/enclave/crypto/openssl/CMakeLists.txt
+++ b/enclave/crypto/openssl/CMakeLists.txt
@@ -24,9 +24,6 @@ maybe_build_using_clangw(oecryptoopenssl)
 
 enclave_enable_code_coverage(oecryptoopenssl)
 
-enclave_include_directories(oecryptoopenssl PUBLIC
-                            $<BUILD_INTERFACE:${OE_INCDIR}/openenclave>)
-
 enclave_include_directories(
   oecryptoopenssl PRIVATE
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/3rdparty/mbedtls/mbedtls/include>
@@ -35,7 +32,7 @@ enclave_include_directories(
 enclave_link_libraries(oecryptoopenssl PUBLIC openssl)
 
 # Enforce the correct build dependency such that OpenSSL will be built first.
-add_enclave_dependencies(oecryptoopenssl opensslcrypto)
+add_enclave_dependencies(oecryptoopenssl openssl)
 
 set_enclave_property(TARGET oecryptoopenssl PROPERTY ARCHIVE_OUTPUT_DIRECTORY
                      ${OE_LIBDIR}/openenclave/enclave)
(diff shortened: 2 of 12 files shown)


mingweishih:
bors try

mingweishih:
bors try

mingweishih:
bors try

radhikaj (approved):

radhikaj:
bors try

enclave/crypto/openssl/CMakeLists.txt
```diff
@@ -24,9 +24,6 @@ maybe_build_using_clangw(oecryptoopenssl)
 
 enclave_enable_code_coverage(oecryptoopenssl)
 
-enclave_include_directories(oecryptoopenssl PUBLIC
-                            $<BUILD_INTERFACE:${OE_INCDIR}/openenclave>)
-
 enclave_include_directories(
```
manojrupireddy, line 27: For my understanding purpose, can you explain what we are doing here earlier and now?

radhikaj:
bors try

mingweishih:
bors try

enclave/crypto/openssl/CMakeLists.txt
mingweishih: This is just a duplicated include. When linking against `openssl`, `oecryptoopenssl` will inhere the same include path.

mingweishih:
bors try

3rdparty/openssl/CMakeLists.txt
```diff
@@ -292,7 +292,7 @@ ExternalProject_Add(
   INSTALL_COMMAND "")
 
 add_dependencies(openssl_includes openssl_configure)
-add_dependencies(openssl_generated openssl_configure)
+add_dependencies(openssl_generated openssl_includes)
 add_dependencies(openssl_generated openssl_copied)
```
anakrish, line 296: There seems to be an ordering bug here. `openssl_copied` does not depend on anything else and therefore could be launched before openssl_configure.

anakrish (commented):
lgtm. I think there is a cmake module dependency issue.

3rdparty/openssl/CMakeLists.txt
mingweishih: It should be fine given that `openssl_copied` and `openssl_configure` are independent.  `opensslcrypto` depends on `openssl_generated`, which ensures both  `openssl_copied` and `openssl_configure` will be ready before building the library.

anakrish: From the code, `openssl_configure` creates `${OPENSSL_BUILD_INCLUDES}` and hence implicitly `${OPENSSL_BUILD}`
https://github.com/openenclave/openenclave/blob/b8451393224fa0ea18cd0607983c6ef4f4519ff9/3rdparty/openssl/CMakeLists.txt#L77

`openssl_copied` write to `${OPENSSL_BUILD}`
https://github.com/openenclave/openenclave/blob/b8451393224fa0ea18cd0607983c6ef4f4519ff9/3rdparty/openssl/CMakeLists.txt#L285-L286


Therefore there is a dependency.

mingweishih: Fixed

anakrish (approved):

PR #3763, opened 2020-12-03
Enable the Attested TLS sample with OpenSSL on CI and Update documentations

mingweishih:
This PR enables the attested TLS sample test on CI. In addition, the PR does the following
- Fix a bug that `OE_CRYPTO_LIB=openssl` does not work with cmake
- Update the README.md on the attested tls sample and comments CMakelists.txt/Makefile samples to reflect that setting `OE_CRYPTO_LIB` currently only works with the former.

### DIFF
--- a/samples/README.md
+++ b/samples/README.md
@@ -32,7 +32,8 @@ The following samples demonstrate how to develop enclave applications using OE A
 #### [File-Encryptor](file-encryptor/README.md)
 
 - Show how to encrypt and decrypt data inside an enclave
-- Use AES mbedTLS API to perform encryption and decryption
+- Use AES Mbed TLS API to perform encryption and decryption
+  - *Note:* this sample does not support OpenSSL yet.
 
 #### [Data-Sealing](data-sealing/README.md)
 
@@ -41,11 +42,15 @@ The following samples demonstrate how to develop enclave applications using OE A
 - Explore two supported seal polices
   - OE_SEAL_POLICY_UNIQUE
   - OE_SEAL_POLICY_PRODUCT
+- Use AES and MD Mbed TLS APIs to perform sealing and unsealing
+  - *Note:* this sample does not support OpenSSL yet.
 
 #### [Attestation](attestation/README.md)
 
 - Explain how OE attestation works
 - Demonstrate an implementation of attestation between two enclaves
+- Use SHA and PCKS1 Mbed TLS APIs to perform hashing and encryption
+  - *Note:* this sample does not support OpenSSL yet.
 
 #### [Attested TLS](attested_tls/README.md)
 
@@ -53,6 +58,7 @@ The following samples demonstrate how to develop enclave applications using OE A
 - Demonstrate an implementation for how to establish an Attested TLS channel
   - between two enclaves
   - between one non-enclave client and an enclave
+- Deomnstrate how to build the sample with either Mbed TLS or OpenSSL libraries
 
 #### [Switchless Calls](switchless/README.md)
 
--- a/samples/attested_tls/README.md
+++ b/samples/attested_tls/README.md
@@ -71,7 +71,57 @@ Note: Both of them can run on the same machine or separate machines.
 
 ## Build and run
 
-To build and run the samples, refer to documentation in the main [README file](../README.md#building-the-samples.md).
+### Linux
+
+#### CMake
+- Use Mbed TLS
+  ```bash
+  mkdir build
+  cd build
+  cmake -DOE_CRYPTO_LIB=mbedtls ..
+  make
+  make run
+  ```
+- Use OpenSSL
+  ```bash
+  mkdir build
+  cd build
+  cmake -DOE_CRYPTO_LIB=openssl ..
+  make
+  make run
+  ```
+
+#### GNU Make
+- Use Mbed TLS
+  ```bash
+  make OE_CRYPTO_LIB=mbedtls build
+  make run
+  ```
+- Use OpenSSL
+  ```bash
+  make OE_CRYPTO_LIB=openssl build
+  make run
+  ```
+
+### Windows
+
+#### CMake
+- Use Mbed TLS
+  ```bash
+  mkdir build
+  cd build
+  cmake -G Ninja -DOE_CRYPTO_LIB=mbedtls ..
+  ninja
+  ninja run
+  ```
+- Use OpenSSL
+  ```bash
+  mkdir build
+  cd build
+  cmake -G Ninja -DOE_CRYPTO_LIB=openssl ..
+  ninja
+  ninja run
+  ```
 
 Note: This sample has a dependency on the [socket support](../../docs/UsingTheIOSubsystem.md#a-socket-example) added in the OE SDK v0.6.0 release, so it needs to be linked against the liboehostsock and libhostresolver libraries. For more details, see [Using the Open Enclave I/O subsystem](../../docs/UsingTheIOSubsystem.md#opting-in).
 
--- a/samples/test-samples.cmake
+++ b/samples/test-samples.cmake
@@ -42,10 +42,12 @@ endif ()
 if (BUILD_ENCLAVES)
   set(SAMPLES_LIST debugmalloc file-encryptor helloworld log_callback
                    switchless)
+  set(CRYPTO_LIB_LIST mbedtls mbedtls mbedtls mbedtls mbedtls)
   # Debug malloc will set allocated memory to a fixed pattern.
   # Hence do not enable pluggable_allocator test under USE_DEBUG_MALLOC.
   if (COMPILER_SUPPORTS_SNMALLOC AND NOT USE_DEBUG_MALLOC)
     list(APPEND SAMPLES_LIST pluggable_allocator)
+    list(APPEND CRYPTO_LIB_LIST mbedtls)
   endif ()
 endif ()
 
@@ -64,11 +66,13 @@ else ()
   # that cause they directly interface with the AESM service.
   if (BUILD_ENCLAVES)
     list(APPEND SAMPLES_LIST data-sealing)
+    list(APPEND CRYPTO_LIB_LIST mbedtls)
 
     # These tests can only run with SGX-FLC, meaning they were built
     # against SGX.
     if (HAS_QUOTE_PROVIDER)
-      list(APPEND SAMPLES_LIST attested_tls attestation)
+      list(APPEND SAMPLES_LIST attested_tls attested_tls attestation)
+      list(APPEND CRYPTO_LIB_LIST mbedtls openssl mbedtls)
     endif ()
   endif ()
 endif ()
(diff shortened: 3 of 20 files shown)


mingweishih:
bors try

samples/test-samples.cmake
```diff
@@ -95,9 +99,22 @@ endif ()
 # A variable to know if all samples ran successfully
```
manojrupireddy, line 99: Can we add some check here to check the length of SAMPLES_LIST  equal to length of CRYPTO_LIB_LIST and generate fatal error if they don't match.

mingweishih: Given that we manually set the `CRYPTO_LIB_LIST` along with `SAMPLES_LIST`, their length should be match.
If they do not match, either line 105 or 106 will fail eventually.

samples/attested_tls/README.md
```diff
@@ -71,7 +71,55 @@ Note: Both of them can run on the same machine or separate machines.
...
+  mkdir build
+  cd build
+  cmake -DOE_CRYPTO_LIB=openssl ..
+  make run
+  ```
+
+#### GNU Make
+- Use Mbed TLS
+  ```bash
+  make OE_CRYPTO_LIB=mbedtls build
```
manojrupireddy, line 95: Will this override internal OE_CRYPTO_LIB setting?

mingweishih: Yes. But the behavior is the same as the default given that OE_CRYPTO_LIB is set to the same value.

manojrupireddy (approved):
Thank you for the fix. LGTM.

mingweishih:
bors try

samples/test-samples.cmake
```diff
@@ -42,10 +42,12 @@ endif ()
 if (BUILD_ENCLAVES)
   set(SAMPLES_LIST debugmalloc file-encryptor helloworld log_callback
                    switchless)
+  set(CRYPTO_LIB_LIST mbedtls mbedtls mbedtls mbedtls mbedtls)
   # Debug malloc will set allocated memory to a fixed pattern.
   # Hence do not enable pluggable_allocator test under USE_DEBUG_MALLOC.
   if (COMPILER_SUPPORTS_SNMALLOC AND NOT USE_DEBUG_MALLOC)
     list(APPEND SAMPLES_LIST pluggable_allocator)
+    list(APPEND CRYPTO_LIB_LIST mbedtls)
```
anakrish, line 50: What is the relation between snmalloc and mbedtls?

mingweishih: This means that the sample is linking against oecryptombedtls instead of oecryptoopenssl.

mingweishih: Basically, each sample in the list is paired with either mbedtls or openssl, which indicates with or crypoto library the sample will use. This provides the flexibility if later we want to test some samples with OpenSSL.

samples/attested_tls/README.md
```diff
@@ -71,7 +71,55 @@ Note: Both of them can run on the same machine or separate machines.
...
 
-To build and run the samples, refer to documentation in the main [README file](../README.md#building-the-samples.md).
+### Linux
+
+#### CMake
+- Use Mbed TLS
+  ```bash
+  mkdir build
+  cd build
+  cmake -DOE_CRYPTO_LIB=mbedtls ..
```
dthaler, line 81: Are these instructions really specific to this particular sample, or would they apply to many samples?
Why can't we just put this instruction into the ../README.md file?
All the other CMakeLists.txt files you touched here seem to already say:
```
# set OE_CRYPTO_LIB to either "mbedtls" or "openssl" based on the crypto wrapper to be used.
```
So based on that statement already being in the other samples I would say these instructions are in the wrong file and should be moved to ../README.md.

dthaler (commented):
Looks ok other than the change to the README file.  I would request that the other readme file be changed instead, per comments there.  Since the README.md file change is not really about enabling a sample "on CI", it is also fine with me to revert the README.md change in this PR and deal with it in a separate PR.

samples/attested_tls/README.md
mingweishih: Currently, these instructions are specific to the attested tls sample (the others either we don't support or they don't use crypoto APIs at all).  In this case, it's the text in the CMakeList.txt you mentioned is wrong. However, once all the samples support building with OpenSSL, we should update the text based on your suggestion.

Do you think we should fix the CMakelist.txt for now and revisit later?

dthaler: If you're saying the text in all the other CMakeLists.txt files is wrong then yes we should fix them now and revisit later.   But I recommend the README.md change here be in the same PR as the CMakeLists.txt fixes, and you're already touching the other CMakeLists.txt files in this PR so you might as well do it in this PR.

dthaler: The other possibility is to do the README and CMakeLists.txt changes in a separate PR, and remove all the CMakeLists.txt changes for the other samples from this PR.  E.g., if your point is that the helloworld sample, for example, doesn't depend on the crypto library at all, then it doesn't need to have the line you added:
```
set(OE_CRYPTO_LIB
    mbedtls
    CACHE STRING "Crypto library used by enclaves.")
```
You can just delete the OE_CRYPTO_LIB setting if it's not used, right?

mingweishih: Ok, I'll go ahead to update the CMakeLists in this PR.

manojrupireddy: I still think each sample application needs crypto library to be mentioned. Aren't all sample applications need OE crypto wrapper library for some internal functions that are not visible to developers?

mingweishih: Even if a sample does not use crypto APIs, we still need to link against either oecryptoopenssl or oecryptombedtls (based on the value of `OE_CRYPTO_LIB`) to fullfil the linker requirements. I would say once eventually we should be able to expose the OE_CRYPTO_LIB to user (i.e.., modify the ../README.md as you suggested)

mingweishih:
bors try

mingweishih:
bors try

mingweishih:
bors try

mingweishih:
@dthaler  Can you take a look if the updated version address your concern?

dthaler (approved):
> @dthaler Can you take a look if the updated version address your concern?

For this PR, yes I think so.  I would question why openssl works with attested_tls sample but not helloworld sample.  That seems really odd to me, can you elaborate?

Also, what's the issue # tracking updating the other samples? (if there isn't one, can you open one?)

mingweishih:
> > @dthaler Can you take a look if the updated version address your concern?
> 
> For this PR, yes I think so. I would question why openssl works with attested_tls sample but not helloworld sample. That seems really odd to me, can you elaborate?
> 
> Also, what's the issue # tracking updating the other samples? (if there isn't one, can you open one?)

It's just that we didn't test the helloworld sample on the CI yet. I'd like to have a separated PR to address the concerns all together and limit the support of openssl to the `attested_tls` in this release. #3766 tracks the issue.

dthaler:
> It's just that we didn't test the helloworld sample on the CI yet. I'd like to have a separated PR to address the concerns all together and limit the support of openssl to the `attested_tls` in this release. #3766 tracks the issue.

Re "in this release", in that case the CHANGELOG for this release should not imply that openssl is supported for enclaves.  It should only imply that this one sample works with openssl, and not that it is ready for general use.

samples/attested_tls/README.md
radhikaj: @mingweishih , please correct me if I am wrong, but in the helloworldsample: if we were to replace this line in CMakeLists.txt :     enclave openenclave::oeenclave openenclave::oecrypto${OE_CRYPTO_LIB} with openenclave::oecryptoopenssl, wouldnt that just mean the helloworldsample picks up the oecryptoopenssl library?
1) So it is not that the helloworld cannot work with oecryptoopenssl, it is just that we have shown an example of how it works in the attested_tls sample.
2) So if we just update the central readme to say the above, wouldnt that mean that the helloworld sample etc would work with openssl?

mingweishih:
> > It's just that we didn't test the helloworld sample on the CI yet. I'd like to have a separated PR to address the concerns all together and limit the support of openssl to the `attested_tls` in this release. #3766 tracks the issue.
> 
> Re "in this release", in that case the CHANGELOG for this release should not imply that openssl is supported for enclaves. It should only imply that this one sample works with openssl, and not that it is ready for general use.

To clarify: There are two parts of OpenSSL support: the OpenSSL libraries (`libssl` and `libcrypto`) and the OE crypto wrapper (`oecryptoopenssl`) for internal use. However, when an enclave links against `oecryptoopenssl` (using OE's cmake/make configuration), it automatically gets OpenSSL libraries.

I think we can still say that we support OpenSSL libraries for general use in this release and refer to the `attested_tls` sample for an example (i.e., developers have to link against `oecryptoopenssl` with OE's build configuration). For samples like helloworld, which do not use OpenSSL APIs at all, I think not supporting/testing them with `oecryptoopenssl` does not violate the claim.

samples/attested_tls/README.md
mingweishih: 1 is correct and 2 is mostly correct with two exceptions: data-sealing and file encryptor samples. 
The alternative is update the common README and mention that these two samples are not supported yet.
I don't have preferences between the two approaches. What do you think? @radhikaj  @dthaler

dthaler: My preference is to say that openssl can be used with arbitrary enclaves in this release (i.e., update the common README and mention that those two samples are not updated yet).   I have no preference on whether that's done in this PR or a separate PR, just that I prefer that it be done for this release.

mingweishih: Ok, I'll update them in this PR.

mingweishih:
@dthaler  Please take a look on the updated README. Thanks!

samples/README.md
```diff
@@ -41,18 +41,21 @@ The following samples demonstrate how to develop enclave applications using OE A
...
 - Demonstrate an implementation of attestation between two enclaves
+- *Note:* this sample does not support OpenSSL yet.
 
 #### [Attested TLS](attested_tls/README.md)
 
 - Explain what an Attested TLS channel is
 - Demonstrate an implementation for how to establish an Attested TLS channel
   - between two enclaves
   - between one non-enclave client and an enclave
+- Deomnstrate how to optionally build the sample with Mbed TLS or OpenSSL libraries.
```
radhikaj, line 58:
```suggestion
- Deomnstrate how to build the sample with either Mbed TLS or OpenSSL libraries.
```

radhikaj (approved):
The readme looks good! Thanks @mingweishih

anakrish (approved):
LGTM

samples/README.md
```diff
@@ -41,18 +41,21 @@ The following samples demonstrate how to develop enclave applications using OE A
 - Explore two supported seal polices
   - OE_SEAL_POLICY_UNIQUE
   - OE_SEAL_POLICY_PRODUCT
+- *Note:* this sample does not support OpenSSL yet.
```
dthaler, line 44:
```suggestion
- Use AES and MD Mbed TLS APIs to perform sealing and unsealing
  - *Note:* this sample does not support OpenSSL yet.
```

Also add the same Note after line 35 and fix mbedTLS -> Mbed TLS in line 35

samples/README.md
```diff
@@ -41,18 +41,21 @@ The following samples demonstrate how to develop enclave applications using OE A
...
 
 #### [Attestation](attestation/README.md)
 
 - Explain how OE attestation works
 - Demonstrate an implementation of attestation between two enclaves
+- *Note:* this sample does not support OpenSSL yet.
```
dthaler, line 50:
```suggestion
- Use SHA and PCKS1 Mbed TLS APIs to perform hashing and encryption
  - *Note:* this sample does not support OpenSSL yet.
```

radhikaj:
Bors r+

PR #3765, opened 2020-12-04
Add EXPORT openenclave-hostverify-targets and different target files

### LINKED ISSUE #2263: Allow building only hostverify library and with minimal build dependencies

### LINKED ISSUE #3160: Unable to separate oehost target while building host-verify package

### LINKED ISSUE #3181: Cross-compilation not supported on Linux

### LINKED ISSUE #3300: open-enclave host-verify sample build and run failures
1) open-enclave package,  host-verify sample , both cmake and gnu make can be used to build the sample. make run fails.
2) open-enclave-host-verify package, host-verify sample, gnu make can be used to build the sample. make run fails. cmake fails.

ryanhsu19:
Add a new export, `openenclave-hostverify-targets`, while using the same namespace, `openenclave::`, to deal with the building issue of the independent `open-enclave-hostverify` package.

Also remove unnecessary `COMPONENT NOT MATCHES OEHOSTVERIFY`s and only retain the one that switches between different target files generated by the new export.

Explicitly write out the compile options and definitions for `oehostverify` in `host/CMakeLists.txt`.

Resolve #2263, resolve #3160, resolve #3181, and resolve #3300.

### DIFF
--- a/host/CMakeLists.txt
+++ b/host/CMakeLists.txt
@@ -435,7 +435,7 @@ endif ()
 # For including edge routines.
 target_include_directories(oehost PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 
-# Compile definitions and options
+# Compile definitions and options for oehost and oehostverify
 target_compile_definitions(
   oehost
   PUBLIC # NOTE: This definition is public to the rest of our project's
@@ -444,13 +444,23 @@ target_compile_definitions(
          $<BUILD_INTERFACE:OE_API_VERSION=2>
   PRIVATE OE_BUILD_UNTRUSTED OE_REPO_BRANCH_NAME="${GIT_BRANCH}"
           OE_REPO_LAST_COMMIT="${GIT_COMMIT}")
+target_compile_definitions(
+  oehostverify
+  PUBLIC # NOTE: This definition is public to the rest of our project's
+         # targets, but should not be exposed to consumers of our
+         # package.
+         $<BUILD_INTERFACE:OE_API_VERSION=2>
+  PRIVATE OE_BUILD_UNTRUSTED OE_REPO_BRANCH_NAME="${GIT_BRANCH}"
+          OE_REPO_LAST_COMMIT="${GIT_COMMIT}")
 
 if (USE_DEBUG_MALLOC)
   target_compile_definitions(oehost PRIVATE OE_USE_DEBUG_MALLOC)
+  target_compile_definitions(oehostverify PRIVATE OE_USE_DEBUG_MALLOC)
 endif ()
 
 if (WITH_EEID)
   target_compile_definitions(oehost PRIVATE OE_WITH_EXPERIMENTAL_EEID)
+  target_compile_definitions(oehostverify PRIVATE OE_WITH_EXPERIMENTAL_EEID)
 endif ()
 
 if (UNIX)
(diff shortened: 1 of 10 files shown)


ryanhsu19:
bors try

host/CMakeLists.txt
```diff
@@ -469,15 +469,37 @@ if (CMAKE_C_COMPILER_ID MATCHES GNU)
...
-  oehostverify
-  PRIVATE $<TARGET_PROPERTY:oehost,COMPILE_OPTIONS>
-  INTERFACE $<TARGET_PROPERTY:oehost,INTERFACE_COMPILE_OPTIONS>)
-
 target_compile_definitions(
   oehostverify
-  PRIVATE $<TARGET_PROPERTY:oehost,COMPILE_DEFINITIONS>
-  INTERFACE $<TARGET_PROPERTY:oehost,INTERFACE_COMPILE_DEFINITIONS>)
+  PUBLIC # NOTE: This definition is public to the rest of our project's
+         # targets, but should not yet be exposed to consumers of our
```
mingweishih, line 475:
```suggestion
         # targets, but should not be exposed to consumers of our
```

samples/host_verify/CMakeLists.txt
```diff
@@ -63,3 +65,5 @@ else ()
     endif ()
   endif ()
 endif ()
+
+unset(COMPONENT)
```
mingweishih, line 69: Can you comment on why we need the unset?

host/CMakeLists.txt
```diff
@@ -494,6 +516,11 @@ endif ()
 install(
   TARGETS oehost
   EXPORT openenclave-targets
+  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}/openenclave/host)
+
+install(
+  TARGETS oehostverify
```
mingweishih, line 522: Question: Do we still need the openenclave-targets as part of OEHOSTVERIFY (line 527 - 531)?

mingweishih (commented):
The PR looks clean! Thanks for the work. Just has some comments.

mingweishih:
bors try

host/CMakeLists.txt
ryanhsu19: Good point. I'll need to test it out.

ryanhsu19:
The 1st `bors try` failure was due to `openenclave::oecore` target not found (in every failed case), but I could not reproduce it on my local Ubuntu 18.04 LTS machine. I'm still trying to figure it out.

host/CMakeLists.txt
```diff
@@ -469,15 +469,37 @@ if (CMAKE_C_COMPILER_ID MATCHES GNU)
 endif ()
 
 # Use the same the compile options and definitions from oehost.
-target_compile_options(
-  oehostverify
-  PRIVATE $<TARGET_PROPERTY:oehost,COMPILE_OPTIONS>
-  INTERFACE $<TARGET_PROPERTY:oehost,INTERFACE_COMPILE_OPTIONS>)
-
 target_compile_definitions(
```
mingweishih, line 472: One more comment: Instead of doing this (here and the other newly added code), can we just set those compile definitions/options at the same places as `oehost`? See what we did in line 345&347, 349&350 (it would be good to flip the order: oehost then oehostverify, making it consistent), 413&415 for examples.

ryanhsu19:
bors try

mingweishih (approved):
Thanks for addressing all the comments! Great work!

mingweishih:
Please squash the comments as well!

anakrish (approved):
LGTM

ryanhsu19:
> 
> 
> Please squash the comments as well!

Yeah, I'll do it after ensuring that everything's alright.

ryanhsu19:
bors try

PR #3774, opened 2020-12-09
Fix oesign test without quote provider

thomasten:


anakrish:
bors try

radhikaj:
@jianlia @manojrupireddy @mingweishih , can you please review?

tests/tools/oesign/test-enclave/host/host.c
```diff
@@ -93,7 +94,7 @@ int main(int argc, const char* argv[])
         }
     }
 
-    if (_is_kss_supported())
```
CodeMonkeyLeet, line 96: This works to unblock, but ideally test could work without the quote provider; `check_kss_extended_ids` is making the call with `OE_REPORT_FLAGS_REMOTE_ATTESTATION` but it could just get the local report and validate the IDs from that directly.

CodeMonkeyLeet: @anakrish We could approve this and follow up with a more comprehensive fix. Thoughts?

radhikaj: cc @mingweishih

radhikaj: cc @manojrupireddy

tests/tools/oesign/test-enclave/host/host.c
```diff
@@ -93,7 +94,7 @@ int main(int argc, const char* argv[])
...
+    if (_is_kss_supported() && oe_has_sgx_quote_provider())
```
mingweishih, line 97: Suggest adding comments to explain the workaround. We can revisit this later as @CodeMonkeyLeet  suggested.

```suggestion
    /* check_kss_extended_ids currently assumes the quote provider is available.
     * Skip if there is no quote provider for now.
     */
    if (_is_kss_supported() && oe_has_sgx_quote_provider())
```

mingweishih (approved):

yentsanglee (approved):

PR #3772, opened 2020-12-08
Correctly determine AVX support.

anakrish:
Based on https://software.intel.com/content/www/us/en/develop/blogs/is-avx-enabled.html,
return value of xgetbv must be first anded with the mask and then checked for equality
with the mask.

fixes #3762

Signed-off-by: Anand Krishnamoorthi <anakrish@microsoft.com>

PR #3779, opened 2020-12-10
Selectively use avx instructions

anakrish:

### DIFF
--- a/tests/abi/enc/enc.cpp
+++ b/tests/abi/enc/enc.cpp
@@ -9,6 +9,17 @@
 #include "../abi_utils.h"
 #include "abi_t.h"
 
+OE_EXTERNC_BEGIN
+// By default, assume avx is available. This is the case with most processors
+// with SGX support.
+bool oe_is_avx_enabled = true;
+OE_EXTERNC_END
+
+void enclave_set_oe_is_avx_enabled(bool enabled)
+{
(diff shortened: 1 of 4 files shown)


anakrish:
bors try

tests/abi/enc/enc.cpp
```diff
@@ -9,6 +9,17 @@
 #include "../abi_utils.h"
 #include "abi_t.h"
 
+OE_EXTERNC_BEGIN
+// By default, assume avx is available. This is the case with most processors
+// with SGX support.
+bool oe_is_avx_enabled = true;
```
mingweishih, line 15: Don't you need the similar check as #3772? If `oe_is_avx_enabled` is always set to `true`, then the test could still fail on the machine without AVX support.

mingweishih: Never mind. On the host, `oe_is_avx_enabled` is global and set by the check in #3772. I'm just confused by that both the enclave and the host use the same variable name.

anakrish: `oe_is_avx_enabled` defaults to true. It is set to the correct value by the host via `enclave_set_oe_is_avx_enabled`

mingweishih (approved):
LGTM

nonpolarity (approved):

radhikaj (approved):

PR #3792, opened 2021-01-08
Deprecate GCC support for compiling Open Enclave SDK for Intel SGX from source

CyanDevs:


BRMcLaren:
bors try

BRMcLaren:
This code does remove all the GCC components from the testing framework, but I believe we need to continue to test the packages themselves with GCC. Is it possible that doing something along the lines of refactoring this section of code would help accomplish this? https://github.com/openenclave/openenclave-ci/blob/master/src/jenkins/common/Openenclave.groovy#L33

.jenkins/pipelines/Agnostic/Linux/Jenkinsfile
```diff
@@ -38,27 +38,6 @@ def simulationContainerTest(String version, String build_type, List extra_cmake_
     }
 }
 
-def AArch64GNUTest(String version, String build_type) {
```
mingweishih, line 41: We need to keep this.

.jenkins/pipelines/Agnostic/Linux/Jenkinsfile
```diff
@@ -102,10 +81,6 @@ try{
         "Check CI":                                { checkCI() },
         "Check Developer Experience Ubuntu 16.04": { checkDevFlows('16.04') },
         "Check Developer Experience Ubuntu 18.04": { checkDevFlows('18.04') },
-        "AArch64 1604 GNU gcc Debug":              { AArch64GNUTest('16.04', 'Debug') },
-        "AArch64 1604 GNU gcc Release":            { AArch64GNUTest('16.04', 'Release') },
-        "AArch64 1804 GNU gcc Debug":              { AArch64GNUTest('18.04', 'Debug') },
```
mingweishih, line 107: We should retain the aarch64 configurations. 1604 ones can be dropped later along with deprecating 1604 support.

mingweishih (changes requested):

.jenkins/pipelines/Agnostic/Linux/Jenkinsfile
BRMcLaren: Should we maybe rename the 'gcc' part to not mislead?

I took a look at the code [here](https://github.com/openenclave/openenclave-ci/blob/master/src/jenkins/common/Openenclave.groovy#L66) which the tests rely on and it seems it doesn't have anything to do with gcc?

mingweishih: It looks like
https://github.com/openenclave/openenclave/blob/3b6f413686846ec3b47ef7e94824c4078c2546f6/.jenkins/pipelines/Agnostic/Linux/Jenkinsfile#L51
will default to aarch gcc tool chain.

BRMcLaren: I think your understanding of this matter may far exceed ours. If we are to deprecate gcc, that to me sounds like then this needs to be removed. Should the process instead be to deprecate gcc from docs and ci but keep these tests.

I looked into it a little, but it seems to me that GNU GCC != GCC which would align with what you are saying, whereas GNU GCC is for cross compilaiton.

CyanDevs: As I understand it, the reasoning is to keep GCC support for OPTEE (AArch64) even though we are deprecating it for SGX. I have pushed new changes that keeps this.

BRMcLaren:
bors try

radhikaj (approved):

mingweishih (approved):
LGTM

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 [Unreleased][Unreleased_log]
 --------------
+### Deprecated
+
+- Compiling Open Enclave SDK from source with GCC is no longer supported. The recommended compiler is Clang.
```
mingweishih, line 14:
```suggestion
- Compiling Open Enclave SDK for Intel SGX from source with GCC is no longer supported. The recommended compiler is Clang.
```

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -8,15 +8,15 @@ Also like the Intel SGX SDK, these binaries must be built with a specific set of
 To simplify the process of the specifying the correct build parameters, Open Enclave SDK provides
 a set of `pkg-config` settings that you can use in your build scripts.
 
-There are 8 configurations provided to handle all permutations of the following parameters:
```
mingweishih, line 11: This document is related to building samples and does not need to modify.

CHANGELOG.md
CyanDevs: Good suggestion for clarification. Thank you!

docs/GettingStartedDocs/buildandsign.md
CyanDevs: Thanks for the clarification, I have reverted the changes to this file.

docs/GettingStartedDocs/buildandsign.md
```diff
@@ -97,40 +97,9 @@ NumHeapPages=1024
 NumStackPages=1024
 NumTCS=1
 ```
-Additionally, a developer can specify additional Key Sharing and Separation (KSS) identity properties
```
BRMcLaren, line 100: Is this GCC specific? trying to understand the deletion.

CyanDevs: Good catch, this shouldn't have been changed.

PR #3798, opened 2021-01-13
Deprecate the support of building OE with GCC from source

mingweishih:
This PR adds warning messages when users try to build the SDK with GCC. Also, removing the test cases that explicitly check the GCC version.

### DIFF
--- a/cmake/compiler_settings.cmake
+++ b/cmake/compiler_settings.cmake
@@ -11,6 +11,11 @@ if (CMAKE_C_COMPILER_ID MATCHES Clang)
   endif ()
 endif ()
 
+if (OE_SGX AND CMAKE_C_COMPILER_ID MATCHES GNU)
+  message(WARNING "The GCC support on x86 platforms has been deprecated. "
+                  "Build problems may occur. Consider using Clang instead.")
+endif ()
+
 if (NOT CMAKE_C_COMPILER_ID STREQUAL CMAKE_CXX_COMPILER_ID)
   message(FATAL_ERROR "Your C and C++ compilers have different vendors: "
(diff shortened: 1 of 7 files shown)


mingweishih:
bors try

radhikaj (commented):
Can you also add a note to changelog?

radhikaj (approved):

cmake/compiler_settings.cmake
```diff
@@ -11,6 +11,13 @@ if (CMAKE_C_COMPILER_ID MATCHES Clang)
   endif ()
 endif ()
 
+if (OE_SGX AND CMAKE_C_COMPILER_ID MATCHES GNU)
+  message(
+    FATAL_ERROR
```
anakrish, line 16: Let's allow GCC compilation if the EXPERIMENTAL flag is also specified.

@mikbras @bodzhang @vtikoo 
Do you build oe from source? If so, what compiler do you use?

mingweishih: Now the PR is changed to give warnings instead of raising errors.

vtikoo: We are not specifying any compiler for the openenclave build, and CMake picks up Clang 7.1.0.

mingweishih:
> Can you also add a note to changelog?

Updated.

cmake/compiler_settings.cmake
mingweishih: Thanks for the information!

anakrish (commented):
LGTM. Will the CI support be dropped in another PR?

mingweishih:
> LGTM. Will the CI support be dropped in another PR?

It's done by #3792 already.

anakrish:
Great.

anakrish (approved):

radhikaj:
Bors r+

PR #3735, opened 2020-11-17
Add basic support of config_id/config_svn on Ice Lake platform.

nonpolarity:
Users can define such properties through config file while signing, and these properties will be included on the sgx report.

Signed-off-by: Alvin Chen <alvin@chen.asia>

PR #3799, opened 2021-01-14
provide settings configuration to set config_id/config_svn on ice lake platform enclaves

manojrupireddy:
This is in continuation to the work done by Alvin => https://github.com/openenclave/openenclave/pull/3735.
Made changes to the above draft PR , based on the code review comments there.


include/openenclave/host.h
```diff
@@ -142,6 +154,7 @@ typedef struct _oe_enclave_setting
 #ifdef OE_WITH_EXPERIMENTAL_EEID
         oe_eeid_t* eeid;
 #endif
+        const oe_enclave_setting_config_data* context_config_data;
```
CodeMonkeyLeet, line 157: Looks like a copy-paste issue, should just be:
```suggestion
        const oe_enclave_setting_config_data* config_data;
```

include/openenclave/host.h
```diff
@@ -122,6 +123,17 @@ typedef struct _oe_enclave_setting_context_switchless
...
 
+
+/**
+ * The setting for config_id/config_svn on Ice Lake platform.
+ */
+typedef struct _oe_enclave_setting_config_data
+{
+    uint8_t enclave_config_id[64];
+    uint16_t enclave_config_svn;
+    bool is_config_id_setting_mandatory;
```
CodeMonkeyLeet, line 134: I would invert the definition of this flag so the default is that it fails if the configuration is provided but is not supported at runtime by the current TEE e.g. `bool ignore_if_unsupported`

host/sgx/create.c
```diff
@@ -874,6 +874,27 @@ oe_result_t oe_sgx_build_enclave(
...
+        // Fail if the OE_SGX_FLAGS_KSS is not set when config_id/config_svn
+        // settings are mandatory
+        OE_RAISE_MSG(
+            OE_UNSUPPORTED,
+            "Enclave image requires config_id/config_svn settings but "
+            "OE_SGX_FLAGS_KSS is not set in properties\n",
+            NULL);
+    }
+
+    if(context->is_config_id_setting_mandatory && !_is_kss_supported())
```
CodeMonkeyLeet, line 888: Nit: Missing space between `if(` 
If you haven't already I suggest running it on the changelist to prevent CI from blocking it when you want to merge the changes.

host/sgx/create.c
```diff
@@ -1021,7 +1042,7 @@ oe_result_t oe_create_enclave(
 {
     oe_result_t result = OE_UNEXPECTED;
     oe_enclave_t* enclave = NULL;
-    oe_sgx_load_context_t context;
+    oe_sgx_load_context_t context = {0};
```
CodeMonkeyLeet, line 1045: Are you getting a compiler warning for this? `oe_sgx_load_context_t` has nested structures and arrays and IIRC, C would require that the initializer for the appropriate max depth e.g. `{{0}}` or however deeply the structure is nested.

host/sgx/create.c
```diff
@@ -874,6 +874,27 @@ oe_result_t oe_sgx_build_enclave(
         &loaded_enclave_pages_size,
         &enclave_size));
 
+    if (context->is_config_id_setting_mandatory && !(props.config.attributes & OE_SGX_FLAGS_KSS))
```
CodeMonkeyLeet, line 877: If we allow the developer to ignore this case (i.e. if _not_ mandatory _and_ KSS _not_ set) then OE should probably unset `is_config_id_setting_passed` (I would suggest renaming it to `use_config_id`) so downstream functions like `_new_secs()` behave consistently.

host/sgx/create.c
```diff
@@ -874,6 +874,27 @@ oe_result_t oe_sgx_build_enclave(
...
+        OE_RAISE_MSG(
+            OE_UNSUPPORTED,
+            "Enclave image requires config_id/config_svn settings but "
+            "OE_SGX_FLAGS_KSS is not set in properties\n",
+            NULL);
+    }
+
+    if(context->is_config_id_setting_mandatory && !_is_kss_supported())
+    {
+        // Fail if the CPU does not support KSS and cofig_id/config_svn setting are mandatory
```
CodeMonkeyLeet, line 890: Typo:
```suggestion
        // Fail if the CPU does not support KSS and config_id/config_svn setting are mandatory
```

If you're using VSCode, consider using extensions like the [Code Spell Checker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker) to catch minor things like this.

CodeMonkeyLeet (commented):
Thanks for taking the old PR forwards! See comments for suggestions and concerns.

host/sgx/create.c
mingweishih: Just checked, `{ 0 }` also initializes the nested struct. However, it seems unnecessary to initialize `context`.

include/openenclave/internal/sgxcreate.h
```diff
@@ -65,6 +65,11 @@ struct _oe_sgx_load_context
     /* EEID data needed during enclave creation */
     oe_eeid_t* eeid;
 #endif
+
+    uint8_t  enclave_config_id[64];
```
mingweishih, line 69: Consider creating a struct that matches the settings struct, and have a `bool` to check is the `config_data` is set.
```
    struct
    {
        uint8_t config_id[32];
        uint16_t config_svn;
        bool ignore_if_unsupported;
    } config_data;
    bool is_config_data_set;
```

Or even reuse the setting struct
```
typedef oe_enclave_setting_config_data oe_config_data_t;
...
struct _oe_sgx_load_context
{
   ...
   oe_config_data_t config_data;
   bool is_config_data_set;
}
```

include/openenclave/host.h
```diff
@@ -122,6 +123,17 @@ typedef struct _oe_enclave_setting_context_switchless
...
 } oe_enclave_setting_context_switchless_t;
 
+
+/**
+ * The setting for config_id/config_svn on Ice Lake platform.
+ */
+typedef struct _oe_enclave_setting_config_data
+{
+    uint8_t enclave_config_id[64];
+    uint16_t enclave_config_svn;
```
mingweishih, line 133:
```suggestion
    uint16_t config_svn;
```

host/sgx/create.c
mingweishih: Some obervations:
- `context->attributes.flags & OE_ENCLAVE_FLAG_SGX_KSS == true` should indicate that both KSS bit is set and the CPU supports the feature after the `if (props.config.attributes & OE_SGX_FLAGS_KSS)` statement.
- We should check only if the `is_config_id_setting_passed` (or `is_config_data_set` as suggested) is set.

Can the two if statements be merged into one as follows (also incorporate @CodeMonkeyLeet'` suggestion)?

```
    if (props.config.attributes & OE_SGX_FLAGS_KSS)
    {
        ...
    }

    if (context->is_config_data_set &&
        !(context->attributes.flags & OE_ENCLAVE_FLAG_SGX_KSS)
    {
        if (!context->config_data.ignore_if_unsupported)
        {
            OE_RAISE_MSG(...);
        }
        else
        {
            context->is_config_data_set = false;
        } 
    }
```

include/openenclave/host.h
```diff
@@ -122,6 +123,17 @@ typedef struct _oe_enclave_setting_context_switchless
     size_t max_enclave_workers;
 } oe_enclave_setting_context_switchless_t;
 
+
+/**
+ * The setting for config_id/config_svn on Ice Lake platform.
+ */
+typedef struct _oe_enclave_setting_config_data
+{
+    uint8_t enclave_config_id[64];
```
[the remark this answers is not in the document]
mingweishih: Add @bodzhang

include/openenclave/host.h
```diff
@@ -122,6 +123,17 @@ typedef struct _oe_enclave_setting_context_switchless
     size_t max_enclave_workers;
 } oe_enclave_setting_context_switchless_t;
 
+
+/**
+ * The setting for config_id/config_svn on Ice Lake platform.
+ */
+typedef struct _oe_enclave_setting_config_data
```
dthaler, line 130: Is this struct intended to be used with OE_SGX_ENCLAVE_CONFIG_ID as the setting type?  If so, the terms should match (i.e., SGX in the name, data vs id).   I see comment line 128 does imply SGX only so please rename this struct to be consistent with the setting type value.

include/openenclave/host.h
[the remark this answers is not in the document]
manojrupireddy: Hi @bodzhang , Can you confirm if this is the right data structure to use for SECS data structure. I found this in the intel codebase. I am planning to replicate the same in OE codebase and in intel codebase config_id is allocated 64 bytes. https://github.com/intel/linux-sgx/blob/d3bd1571240bcdf85734c232a4f0c86828443ebb/common/inc/internal/arch.h#L59 @bodzhang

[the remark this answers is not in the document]
mingweishih: There seem to be an inconsistent in the Intel manual. According to the table 37-2 (secs) and 37-23 (report), the configid field should be 64 bytes. However, the text in 38.4.1.3 mentions that "enclave is created the platform can additionally provide 32-byte configuration identifier (CONFIGID).".  I think it should be fine to keep 64 bytes for now.

host/sgx/create.c
```diff
@@ -1075,14 +1094,21 @@ oe_result_t oe_create_enclave(
     OE_CHECK(oe_sgx_initialize_load_context(
         &context, OE_SGX_LOAD_TYPE_CREATE, flags));
 
-#ifdef OE_WITH_EXPERIMENTAL_EEID
     for (size_t i = 0; i < setting_count; i++)
+    {
+        if (settings[i].setting_type == OE_SGX_ENCLAVE_CONFIG_DATA)
+        {
+            context.config_data = (oe_config_data_t*)settings[i].u.config_data;
```
manojrupireddy, line 1101: I am doubtful here as I am typecasting const to a non-const pointer. Can you suggest better alternatives.

CodeMonkeyLeet: I think you have a good intuition that casting to remove `const` is generally a bad practice; in this case, your code change defines both the new settings data structure (as const) and the context data structure member (as non-const), so I think the better question is why define those differently. It looks like context does not expect to modify the config data either, so it could be `const` as well.

As an extra consideration for more general cases it usually boils down to a couple of cases:
- The original intent was always that the parameter or member needs to be modified, in which case the source definition should just be non-const.
- The target does not need to be modified and so also can be marked const (this case)
- The source is intended to be const, but we need to modify the information for some reason. In this case, you should usually pass by value instead of reference (i.e. make a non-const copy of the data).

mingweishih (approved):
LGTM! Thanks for addressing all the feedbacks.

CodeMonkeyLeet (approved):

tests/icelake_config_id/icelake_config_id.edl
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+enclave {
+    from "openenclave/edl/syscall.edl" import *;
```
mingweishih, line 5: Instead of importing everything, please refer to https://github.com/openenclave/openenclave/blob/8f6691ef4927477f17ea7f338c57ab561fb7ffe5/tests/echo/echo.edl#L5 to import minimum list of ocall/ecalls, which supports `printf` or `oe_host_printf` used by tests.

tests/icelake_config_id/icelake_config_id.edl
```diff
@@ -0,0 +1,10 @@
...
+    from "openenclave/edl/sgx/platform.edl" import *;
+    trusted {
+        public int enclave_test_config_id();
+    };
+};
```
mingweishih, line 10: newline

tests/icelake_config_id/host/host.cpp
```diff
@@ -0,0 +1,82 @@
...
+        argv[1],
+        OE_ENCLAVE_TYPE_SGX,
+        flags,
+        mandatory_settings,
+        OE_COUNTOF(mandatory_settings),
+        &enclave);
+
+    // as enclave is not signed with the KSS Flags and config_id settings are
+    // mandatory enclave creation should fail
+    OE_TEST(result != OE_OK);
```
mingweishih, line 77: Can you check the value against `OE_UNSUPPORTED` instead?

tests/icelake_config_id/host/host.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+    const uint32_t flags = oe_get_create_flags();
+
+    oe_sgx_enclave_setting_config_data config_data_setting_optional = {
+        {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
+         0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
+         0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
+         0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15},
+        2,
+        true};
```
mingweishih, line 31: Consider adding comment to explain the Boolean value.
```suggestion
        true /* ignore_if_unsupported */};
```

tests/icelake_config_id/host/host.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+    oe_sgx_enclave_setting_config_data config_data_setting = {
+        {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
+         0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
+         0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
+         0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15},
+        2,
+        false};
```
mingweishih, line 39:
```suggestion
        false /* ignore_if_unsupported */};
```

mingweishih:
bors delegate+

PR #3800, opened 2021-01-14
compiler-rt for enclaves

anakrish:
GCC/clang generate calls to functions defined in "compiler-rt" when they encounter certain constructs
that don't have a direct mapping to a processor instruction.
See https://gcc.gnu.org/onlinedocs/gcc-4.9.4/gccint/Integer-library-routines.html
These functions are normally available in libgcc.a library.

Clang implements these functions in a "compiler-rt" library.
See: https://compiler-rt.llvm.org/
The files in this PR provide the builtins as described in the link above.

The code has been borrowed from Intel SDK.
https://github.com/intel/linux-sgx/tree/master/sdk/compiler-rt

Compiler-rt is one of the core libraries and when compiling a program normally,
the compiler would automatically add this library to the linker line.

The mirror the same behavior, compiler-rt is built as an object library which
is then added to oecore. This ensures that compiler-rt routines appear at the
appropriate position in the linker line - the last library. By adding the objects
to oecore, there is no change to the user experience, and thus this is not a
breaking change.

### DIFF
--- a/3rdparty/compiler-rt/compiler-rt/Makefile
+++ b/3rdparty/compiler-rt/compiler-rt/Makefile
@@ -1,0 +1,54 @@
+#
+# Copyright (C) 2011-2020 Intel Corporation. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#   * Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#   * Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in
+#     the documentation and/or other materials provided with the
+#     distribution.
+#   * Neither the name of Intel Corporation nor the names of its
+#     contributors may be used to endorse or promote products derived
+#     from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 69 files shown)


anakrish:
bors try

3rdparty/compiler-rt/compiler-rt/Makefile
```diff
@@ -0,0 +1,54 @@
+#
+# Copyright (C) 2011-2020 Intel Corporation. All rights reserved.
```
CodeMonkeyLeet, line 2: Is this file actually used? The CMakeLists.txt file looks like it doesn't depend on this at all.

CodeMonkeyLeet (commented):
Thanks for looking into this, I think it generally looks good. At a high level though, have we considered picking up the intrinsic implementations directly from [LLVM](https://github.com/llvm/llvm-project/tree/62ec4ac90738a5f2d209ed28c822223e58aaaeb7/compiler-rt/lib/builtins) instead of from the Intel clone of it? 

- The Intel copies of the files are > 5 years old and its unclear if there have been bug fixes or additional intrinsics that should be included as the ISAs have evolved. 
- The differences in licensing seems to be a function of the age of the files in Intel's repo; the current Apache 2.0 LLVM license references their legacy license which looks like what Intel picked up. It's still an OSI-approved license, so I would not use the Intel copies solely based on a preference for MIT dual-license vs. Apache 2.0. 
- There are additional arch-specific intrinsics that are not reflected in the Intel SDK set. This is actually true for x86_64 and ARM, but my primary concern is whether we are providing an appropriate set of builtins for ARM since that is not something Intel had to concern themselves with.

mrragava:
Could we use the files from compiler-rt which we already have in the repository for gcov runtime.

dthaler (commented):
Can we use a submodule rather than copying all the files?

PR #3801, opened 2021-01-14
Fix the switchless ECALL fallback path

mingweishih:
This PR fixes a bug such that a switchless ECALL will behave like a regular ECALL when the enclave is created without switchless settings. Also, adding tests to the existing switchless test case.


mingweishih:
bors try

anakrish (approved):
Thanks!

radhikaj (approved):
Thanks @mingweishih

radhikaj:
Bors r+

PR #3803, opened 2021-01-15
change tabs to spaces in log_callback makefile

BRMcLaren:
This PR addresses an issue in the log_callback make file where tabs instead of spaces are used. This was not caught before because it depends on an environment variable that is never set in CI and is not the general use case.

Signed-off-by: brmclare <brmclare@microsoft.com>

### DIFF
--- a/samples/debugmalloc/enclave/Makefile
+++ b/samples/debugmalloc/enclave/Makefile
@@ -6,15 +6,15 @@ include ../../config.mk
 CRYPTO_LDFLAGS := $(shell pkg-config oeenclave-$(COMPILER) --variable=${OE_CRYPTO_LIB}libs)
 
 ifeq ($(LVI_MITIGATION), ControlFlow)
-	ifeq ($(LVI_MITIGATION_BINDIR),)
-		$(error LVI_MITIGATION_BINDIR is not set)
-	endif
-	# Only run once.
-	ifeq (,$(findstring $(LVI_MITIGATION_BINDIR),$(CC)))
-		CC := $(LVI_MITIGATION_BINDIR)/$(CC)
-	endif
-	COMPILER := $(COMPILER)-lvi-cfg
-	CRYPTO_LDFLAGS := $(shell pkg-config oeenclave-$(COMPILER) --variable=${OE_CRYPTO_LIB}libslvicfg)
+    ifeq ($(LVI_MITIGATION_BINDIR),)
+        $(error LVI_MITIGATION_BINDIR is not set)
+    endif
+    # Only run once.
+    ifeq (,$(findstring $(LVI_MITIGATION_BINDIR),$(CC)))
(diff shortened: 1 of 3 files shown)


BRMcLaren:
bors try

mingweishih (approved):
LGTM

radhikaj (approved):

radhikaj:
Thanks @BRMcLaren

radhikaj:
Bors r+

CyanDevs:
bors try

anakrish (approved):

anakrish (changes requested):
GNU Make requires TABs for recipe lines
https://www.gnu.org/software/make/manual/html_node/Rule-Syntax.html

TABS are part of the Makefile syntax.
Why is this reported as an error?

CyanDevs:
This change was done due to a new error we found:

> ‘recipe commences before first target. Stop.’
> This means the first thing in the makefile seems to be part of a recipe: it begins with a recipe prefix character and doesn’t appear to be a legal make directive (such as a variable assignment). Recipes must always be associated with a target.
Source: https://www.gnu.org/software/make/manual/html_node/Error-Messages.html

In current tests (non-prow) LVI_MITIGATION is not set anywhere (or possibly set to something other than ControlFlow), so the makefile will always evaluate ifeq ($(LVI_MITIGATION), ControlFlow) to false.

```
ifeq ($(LVI_MITIGATION), ControlFlow)
	ifeq ($(LVI_MITIGATION_BINDIR),)
		$(error LVI_MITIGATION_BINDIR is not set)
	endif
	# Only run once.
	ifeq (,$(findstring $(LVI_MITIGATION_BINDIR),$(CC)))
		CC := $(LVI_MITIGATION_BINDIR)/$(CC)
	endif
	COMPILER := $(COMPILER)-lvi-cfg
	CRYPTO_LDFLAGS := $(shell pkg-config oeenclave-$(COMPILER) --variable=${OE_CRYPTO_LIB}libslvicfg)
endif
```
Example: https://github.com/openenclave/openenclave/blob/5495755d436524ee425276acfb602d454720ac6a/samples/log_callback/enclave/Makefile#L8

In Prow/Jenkins, LVI_MITIGATION is set in the [Jenkins jobs (example)](https://github.com/openenclave/test-infra/blob/b79301462fc3540209e41964c0e5a12b5bb27218/config/jenkins/configuration/jobs/openenclave/Ubuntu1604DebugClang-7.yml#L49-L50), so the above will be true and run the block. Since the lines have a tab prefix, which would be read as a recipe prefix character, it expects it to be part of a target when it is actually not, and therefore fail with the error `‘recipe commences before first target. Stop.’`

anakrish:
@CyanDevs Thanks for the explanation!

anakrish (approved):

PR #3809, opened 2021-01-27
Use clang by default with cmake in samples

### LINKED ISSUE #3179: Different behavior when samples are built using cmake vs make
Take samples/helloworld
When built using make
- OE_BUILD_ENCLAVE is not defined when building enclave. Not really sure if it is supposed to be defined.
- The compiler is chosen as clang-7 if available.

When built using cmake
- OE_BUILD_ENCLAVE is defined when building enclaves.
- The compiler is chosen as GCC while the compiler is invoked as /bin/cc.

This is an important difference to fix since
- There are security implications. Speculative load-hardening is available only in clang which is not chosen by default when using cmake. Users may create their own enclaves based on samples.
- OE_BUILD_ENCLAVE may cause different code to be compiled between make and cmake builds.

mingweishih:
This PR addresses the inconsistency between the behavior of building samples with make and cmake. Now the cmake will use clang by default unless the `CC` and `CXX` environment variables are specified to other compilers.

Fixes #3179.

### DIFF
--- a/samples/attestation/CMakeLists.txt
+++ b/samples/attestation/CMakeLists.txt
@@ -3,6 +3,19 @@
 
 cmake_minimum_required(VERSION 3.11)
 
+# If the CC environment variable has been specified or if the CMAKE_C_COMPILER
+# cmake variable has been passed to cmake, use the C compiler that has been
+# specified. Otherwise, prefer clang. Same for C++ compiler.
+# This must be done before the `project` command.
+if (UNIX)
+  if (NOT DEFINED ENV{CC} AND NOT DEFINED CMAKE_C_COMPILER)
+    find_program(CMAKE_C_COMPILER clang-8 clang)
+  endif ()
+  if (NOT DEFINED ENV{CXX} AND NOT DEFINED CMAKE_CXX_COMPILER)
+    find_program(CMAKE_CXX_COMPILER clang++-8 clang++)
+  endif ()
+endif ()
+
 project("Attestation Sample" LANGUAGES C CXX)
 
 find_package(OpenEnclave CONFIG REQUIRED)
--- a/samples/attested_tls/CMakeLists.txt
+++ b/samples/attested_tls/CMakeLists.txt
@@ -3,6 +3,19 @@
 
 cmake_minimum_required(VERSION 3.11)
 
+# If the CC environment variable has been specified or if the CMAKE_C_COMPILER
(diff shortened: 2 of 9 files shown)


mingweishih:
bors try

samples/attestation/CMakeLists.txt
```diff
@@ -3,6 +3,19 @@
 
 cmake_minimum_required(VERSION 3.11)
 
+# If the CC environment variable has been specified or if the CMAKE_C_COMPILER
```
anakrish, line 6: Can this be abstracted into a cmake utility provided by the SDK?
Note: We can to that in next release too.

samples/attested_tls/CMakeLists.txt
```diff
@@ -3,6 +3,19 @@
 
 cmake_minimum_required(VERSION 3.11)
 
+# If the CC environment variable has been specified or if the CMAKE_C_COMPILER
+# cmake variable has been passed to cmake, use the C compiler that has been
+# specified. Otherwise, prefer clang. Same for C++ compiler.
+# This must be done before the `project` command.
+if (UNIX)
```
anakrish, line 10: Do we need to do anything special for LVI?

mingweishih: Currently we only enable the LVI mitigation build for a few samples (e.g., helloworld).

samples/attestation/CMakeLists.txt
mingweishih: There is a trade-off; we could make the compiler selection transparent or provide a cmake utility that does this. However, the later may require additional `include()` for the cmake script. We can't make the include as part of `OpenEnclave` package because we invoke `find_package(OpenEnclave ...)` after `project`. I find that the current approach is simpler.  What do you think?

mingweishih: Note: we don't need this PR for this release as the clang-8 support is independent from the samples.

anakrish: Sounds good. We can address this the next time we upgrade compiler.

samples/attested_tls/CMakeLists.txt
anakrish: Ok.

anakrish (approved):
LGTM

dthaler (approved):

rs--:
bors try

PR #3797, opened 2021-01-12
Update compiler to clang-8

anakrish:
- Update docs
- LVI mitigation scripts
- Use clang-8's spectre mitigagtion flags (-mspeculative-load-hardening)

Signed-off-by: Anand Krishnamoorthi <anakrish@microsoft.com>

PR #3811, opened 2021-01-28
 Update docs and build scripts to reflect the clang-8 update

mingweishih:
The PR is based on #3797 without CI updates, which will be in a separated PR.

### DIFF
--- a/CMakeSettings.json
+++ b/CMakeSettings.json
@@ -6,7 +6,7 @@
   // Open Enclave SDK from Visual 2017. It is based on the default
   // CMakeSettings.json.
   // Prerequisites:
-  //  1. clang (preferable clang-7) must be installed on the machine
+  //  1. clang (preferable clang-8) must be installed on the machine
   //     The following programs must be available in the path:
   //         clang
   //         llvm-ar
--- a/cmake/compiler_settings.cmake
+++ b/cmake/compiler_settings.cmake
@@ -3,9 +3,9 @@
 
 # Check Clang version.
 if (CMAKE_C_COMPILER_ID MATCHES Clang)
-  if (CMAKE_C_COMPILER_VERSION VERSION_LESS 7 OR CMAKE_C_COMPILER_VERSION
-                                                 VERSION_GREATER 7.99)
-    message(WARNING "Open Enclave officially supports Clang 7 only, "
+  if (CMAKE_C_COMPILER_VERSION VERSION_LESS 8 OR CMAKE_C_COMPILER_VERSION
+                                                 VERSION_GREATER 8.99)
+    message(WARNING "Open Enclave officially supports Clang 8 only, "
                     "but your Clang version (${CMAKE_C_COMPILER_VERSION}) "
                     "is older or newer than that. Build problems may occur.")
   endif ()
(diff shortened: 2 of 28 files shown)


cmake/compiler_settings.cmake
```diff
@@ -60,24 +60,31 @@ include(ccache)
...
-  set(OE_SPECTRE_MITIGATION_FLAGS ${SPECTRE_MITIGATION_FLAGS})
-  # TODO: We really should specify this only on the `oecore` target;
-  # however, the third-party mbed TLS build needs it set to, so we
-  # have to keep this here for now.
-  add_compile_options(${OE_SPECTRE_MITIGATION_FLAGS})
-else ()
-  message(WARNING "Spectre 1 mitigations NOT supported")
+# Apply Spectre mitigation if available.
+# The flag requires clang 8 or above.
+set(SPECTRE_MITIGATION_FLAGS -mspeculative-load-hardening)
```
anakrish, line 65: My preference is for the newer flags. @radhikaj Please confirm if we want to switch to the newer flag.
We must eventually switch to the newer flag; but clang-7 doesn't understand it. Is now a good time for the switch?

tests/libcxx/README.md
```diff
@@ -17,7 +17,7 @@ The tests are currently ran with the following configurations.
 
 | Compiler    | Environment          | Test set |
 |-------------|----------------------|----------|
-| clang 7.1.0 | Ubuntu 16.04 & 18.04 | Full     |
+| clang 8.0.1 | Ubuntu 16.04 & 18.04 | Full     |
```
anakrish, line 20: @BRMcLaren @rs-- Can you confirm which version of clang CI has been updated to? Is it clang 8.0.0 or 8.0.1?

anakrish (commented):
@mingweishih Double-check to make sure all the places using spectre mitigation flags have been updated.

cmake/compiler_settings.cmake
mingweishih: Given that we're giving warning if users do not use clang-8, I think switching to the new flag along with the clang update is reasonable.  If users still want to use clang-7, the mitigation won't apply (`check_c_compiler_flag` will fail), which is fine as we no longer support clang-7.

tests/libcxx/README.md
mingweishih: On Linux, the default clang-8 should be 8.0.0 (installed via apt). However, on Windows, only the 8.0.1 binary is available.

mingweishih: Updated.

cmake/compiler_settings.cmake
radhikaj: For this release, I would prefer that developers  building their application code with clang-7 should not get errors. So would prefer keeping flags in the exported targets same as they were for clang-7. @mingweishih  @anakrish is this possible?

mingweishih: There is no error if developers use clang-7, but just the mitigation won't be applied.  This is due to that the build flags for application are inherent from the OE build. We can revisit this in the next release.

CMakeSettings.json
```diff
@@ -6,7 +6,7 @@
   // Open Enclave SDK from Visual 2017. It is based on the default
```
radhikaj, line 6: why is this rendered with red?

radhikaj:
bors try

radhikaj (approved):

CMakeSettings.json
mingweishih: I don't know neither.. maybe it's a json file?

cmake/compiler_settings.cmake
anakrish: Since clang-7 understands `-mllvm -x86...`, the mitigation will be applied as before.
We don't know if `-mllvm -x86..` and `-mspeculative-load-hardening` are equivalent.

anakrish (approved):

mingweishih:
bors delegate=rs--

rs--:
bors try

rs--:
bors try

rs--:
bors try

Issue #3733, opened 2020-11-17
Upgrade using ansible scripts failed to remove unnecessary files and caused random crashes


jxyang:
Related: https://github.com/openenclave/openenclave/issues/3328

We have removed the files for `libsgx-enclave-common-dev` from the setup scripts, but we don't explicitly remove them from an existing OE installation when we upgrade with `oe-contributors-acc-setup.yml`. With that we have both `libsgx-enclave-common` and `libsgx-enclave-common-dev` on the system. This causes random failures while calling `enclave_delete` in the enclave common library.

anakrish:
AFAIK `lib-sgx-enclave-common-dev` package installs `lib-sgx-enclave-common.so` which is a (symbolic?) link to the  `lib-sgx-enclave-common.X.so` shared library installed by `lib-sgx-enclave-common` package.

Having both `lib-sgx-enclave-common` and `lib-sgx-enclave-common-dev` packages installed ought not cause issues.

Can you let us know what `ldd libsgx-enclave-common.so` shows on the failing system?

radhikaj:
@jxyang Can you please respond to @anakrish's question above?

jxyang:
I have `cured` the system by manually removing the `lib-sgx-enclave-common-dev` package. My observation is that after I done that, the `enclave_delete` doesn't fail randomly anymore.

It's hard to go back to history and run the `ldd` command. But I think removing the redundant package in the ansible script is a prudent thing to do. Not urgent though, unless another user complaints the same issue.

anakrish:
@jxyang 
Good to hear that the problem is resolved. 
We are however curious about the root cause since `lib-sgx-enclave-common-dev` is designed to coexist with `lib-sgx-enclave-common`.

Can you please reinstall `lib-sgx-enclave-common-dev` package and see if the issue starts occuring?

PR #3816, opened 2021-02-02
Enforce Werror when building samples on Linux

### LINKED ISSUE #3773: Inconsistent use of the Werror flag for building samples between Linux and Windows
When building samples on Windows, the maybe_build_using_clangw adds the `Werror` flag, which results in treating warnings as errors.

https://github.com/openenclave/openenclave/blob/eaa360853145b92785ab52e609bc18d1db6d760e/samples/helloworld/enclave/CMakeLists.txt#L16

https://github.com/openenclave/openenclave/blob/eaa360853145b92785ab52e609bc18d1db6d760e/cmake/maybe_build_using_clangw.cmake#L57

This is inconsistent with Linux setting where we didn't add the `Werror` flag (we only enforce it for the building OE libraries and tests).

mingweishih:
This PR fixes #3773, which enforces `Werror` flag when building samples on Linux.

### DIFF
--- a/enclave/core/CMakeLists.txt
+++ b/enclave/core/CMakeLists.txt
@@ -275,7 +275,11 @@ if (OE_SGX)
     # user. There are valid reasons for an end user to use built-ins.
     $<BUILD_INTERFACE:-fno-builtin-malloc
     -fno-builtin-calloc
-    -fno-builtin>)
+    -fno-builtin>
+    # Treat warnings as errors by default. Application based on OE SDK
+    # will inhere this flag, which is consistent to the behavior of the
+    # clangw script.
+    -Werror)
 elseif (OE_TRUSTZONE)
   enclave_compile_options(oecore PUBLIC -fvisibility=hidden ${OE_TZ_TA_C_FLAGS})
 endif ()
--- a/pkgconfig/CMakeLists.txt
+++ b/pkgconfig/CMakeLists.txt
@@ -31,7 +31,10 @@ set(ENCLAVE_CFLAGS_LIST
     -fstack-protector-strong
     -fno-omit-frame-pointer
     -ffunction-sections
-    -fdata-sections)
+    -fdata-sections
+    # Treat warnings as errors by default, which is consistent to the behavior
+    # of the clangw script.
+    -Werror)
 
 set(ENCLAVE_CFLAGS_CLANG_LIST ${ENCLAVE_CFLAGS_LIST}
                               ${SPECTRE_MITIGATION_FLAGS})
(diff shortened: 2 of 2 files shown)


mingweishih:
bors try

enclave/core/CMakeLists.txt
```diff
@@ -275,7 +275,11 @@ if (OE_SGX)
     # user. There are valid reasons for an end user to use built-ins.
     $<BUILD_INTERFACE:-fno-builtin-malloc
     -fno-builtin-calloc
-    -fno-builtin>)
+    -fno-builtin>
+    # Treat warnings as errors by default. Application based on OE SDK
+    # will inhere this flag, which is consistent to the behavior of the
```
anakrish, line 280:
```suggestion
    # will inherit this flag, which is consistent to the behavior of the
```

anakrish (approved):
LGTM

enclave/core/CMakeLists.txt
```diff
@@ -275,7 +275,11 @@ if (OE_SGX)
     # user. There are valid reasons for an end user to use built-ins.
     $<BUILD_INTERFACE:-fno-builtin-malloc
     -fno-builtin-calloc
-    -fno-builtin>)
+    -fno-builtin>
+    # Treat warnings as errors by default. Application based on OE SDK
```
CodeMonkeyLeet, line 279: Is this desirable? This means that every application that uses liboecore via CMake import or pkgconfig will get `-Werror` by default, not just the samples. It's not generally my experience that SDKs set this on behalf of a dev (folks should feel to correct this if I am mistaken.)

This is also potentially a public build breaking change and should at least be documented in the CHANGELOG.md.

anakrish: Good point. I'm unapproving till we are sure that we don't force these flags on the users.

anakrish (changes requested):
We need to make sure that we are not forcing these flags on the users.

enclave/core/CMakeLists.txt
mingweishih: This PR is based on the conclusion from #3773. I'm fine to remove the enforcement o Windows side. Can we have a consensus? Cc @radhikaj @dthaler

anakrish: There is a distinction:
- We want the samples to be built using -Werror
- However, we don't want to force -Werror on the user.

My understanding is that, today, we don't force the flags on the user on both Linux and Windows.

mingweishih: No. -Werror is added in the clangw by default, so samples on Windows are always compiled with the flag, which is what the #3773 about.

anakrish: Thanks for the clarification. We were forcing the flag on Windows, whereas we weren't on Linux.
I'm removing my blocker. 

It is a matter of consistency then.

anakrish (approved):
Reapproving.

enclave/core/CMakeLists.txt
mingweishih: Exactly. Given that @CodeMonkeyLeet pointed out that we shouldn't enforce -Werror on behalf of users, I'm wondering how should we address the consistency? This PR currently follows the conclusion from triaging #3733.  We could also remove the enforcement on the Windows side. Thoughts?

anakrish: There are convincing arguments for both choices. Thinking about it bit more, my preference would be to enforce -Werror since it is generally considered a good programming practice to compile code with high warning levels.

dthaler: +1 to Anand's summary before about what is desired state:
* We want the samples to be built using -Werror
* However, we don't want to force -Werror on the user.

Above should be true for both Linux and Windows.

radhikaj:
bors try

radhikaj (changes requested):
Samples are failing. We should not change the flags that samples are built with since that could cause breaks in current code.

PR #3818, opened 2021-02-02
Update version to v0.14

radhikaj:
Signed-off-by: Radhika Jandhyala <radhikaj@microsoft.com>


rs-- (approved):

nonpolarity (approved):

mingweishih (approved):

dthaler:
What about updating the VS extension to 0.14 too?
Looks like it's still on 0.12, so are we missing a step in the overall release process?

rs--:
> What about updating the VS extension to 0.14 too?
> Looks like it's still on 0.12, so are we missing a step in the overall release process?

I ran into some issues on the v0.13 release, and got side tracked. I'll be sure that the v0.14 VS extension is published for this release.

Issue #3692, opened 2020-10-26
Dont use hardcoded salt values in file encryptor sample applications.


manojrupireddy:
This came as part of openssl file encryptor sample application [PR](https://github.com/openenclave/openenclave/pull/3446) . Currently we use hardcoded salt values for AES encryption and key generation using password related function(PBKDF2). Instead of hard coded salt values residing in the code, salt values should be randomly generated and the encrypted salt value should be part of `encryption_header_t` so that the salt value is available for entire encryption-decryption session.

PR #3826, opened 2021-02-05
replacing constant salt with per-file random salt

RRathna:
replacing constant salt, and generating it randomly per file to be encrypted.
Fixes  https://github.com/openenclave/openenclave/issues/3692

### DIFF
--- a/samples/file-encryptor/enclave/encryptor.cpp
+++ b/samples/file-encryptor/enclave/encryptor.cpp
@@ -7,23 +7,6 @@
 
 ecall_dispatcher::ecall_dispatcher() : m_encrypt(true), m_header(NULL)
 {
-    unsigned char iv[IV_SIZE] = {0xb2,
-                                 0x4b,
-                                 0xf2,
-                                 0xf7,
-                                 0x7a,
-                                 0xc5,
-                                 0xec,
-                                 0x0c,
-                                 0x5e,
-                                 0x1f,
-                                 0x4d,
-                                 0xc1,
-                                 0xae,
-                                 0x46,
-                                 0x5e,
-                                 0x75};
-    memcpy(m_original_iv, iv, IV_SIZE);
 }
 
 int ecall_dispatcher::initialize(
@@ -37,10 +20,20 @@ int ecall_dispatcher::initialize(
         "ecall_dispatcher::initialize : %s request",
         encrypt ? "encrypting" : "decrypting");
 
+    if (header == NULL)
+    {
+        TRACE_ENCLAVE("initialize() failed as NULL was passed in place of "
+                      "(encryption_header_t *)");
+        goto exit;
+    }
+
+    m_password = std::string(password, password + password_len);
     m_encrypt = encrypt;
+    m_header = header;
+
     memset((void*)m_encryption_key, 0, ENCRYPTION_KEY_SIZE_IN_BYTES);
 
-    ret = process_encryption_header(encrypt, password, password_len, header);
+    ret = process_encryption_header(header);
     if (ret != 0)
     {
         TRACE_ENCLAVE("process_encryption_header failed with %d", ret);
@@ -63,8 +56,9 @@ int ecall_dispatcher::initialize(
         TRACE_ENCLAVE("mbedtls_aes_setkey_dec failed with %d", ret);
         goto exit;
     }
+
     // init iv
-    memcpy(m_operating_iv, m_original_iv, IV_SIZE);
+    memcpy(m_operating_iv, m_header->salt, IV_SIZE);
 exit:
     return ret;
 }
--- a/samples/file-encryptor/enclave/encryptor.h
+++ b/samples/file-encryptor/enclave/encryptor.h
@@ -12,8 +12,6 @@
 
 using namespace std;
 
-#define IV_SIZE 16
-
 class ecall_dispatcher
 {
   private:
@@ -24,7 +22,6 @@ class ecall_dispatcher
     encryption_header_t* m_header;
 
     // initialization vector
-    unsigned char m_original_iv[IV_SIZE];
     unsigned char m_operating_iv[IV_SIZE];
 
     // key for encrypting  data
@@ -47,6 +44,7 @@ class ecall_dispatcher
   private:
     int generate_password_key(
         const char* password,
+        unsigned char* salt,
         unsigned char* key,
         unsigned int key_len);
     int generate_encryption_key(unsigned char* key, unsigned int key_len);
@@ -57,13 +55,10 @@ class ecall_dispatcher
         unsigned char* input_data,
         unsigned int input_data_size,
         unsigned char* encrypt_key,
+        unsigned char* salt,
         unsigned char* output_data,
         unsigned int output_data_size);
     int Sha256(const uint8_t* data, size_t data_size, uint8_t sha256[32]);
     void dump_data(const char* name, unsigned char* data, size_t data_size);
-    int process_encryption_header(
-        bool encrypt,
-        const char* password,
-        size_t password_len,
-        encryption_header_t* header);
+    int process_encryption_header(encryption_header_t* header);
 };
--- a/samples/file-encryptor/enclave/keys.cpp
+++ b/samples/file-encryptor/enclave/keys.cpp
@@ -17,7 +17,6 @@
 
 #define ENCRYPT_OPERATION true
 #define DECRYPT_OPERATION false
-#define SALT_SIZE_IN_BYTES 16 // Length of salt size
 
 void ecall_dispatcher::dump_data(
     const char* name,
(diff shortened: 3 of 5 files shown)


samples/file-encryptor/enclave/encryptor.h
```diff
@@ -12,8 +12,6 @@
 
 using namespace std;
 
-#define IV_SIZE 16
```
CodeMonkeyLeet, line 15: I would actually keep this as a separate definition for clarity: fundamentally, there's no reason why the salt size needs to be the same as the IV size, it's just a choice that this particular implementation makes to reuse the salt as the IV.

A little bit of background FYI:
* A _salt_ usually refers to a public, random value used in hashing to mitigate hash table/[rainbow table](https://en.wikipedia.org/wiki/Rainbow_table) attacks on the hashed output. It can be of arbitrary length, and usually chosen to be sufficiently long to make it expensive to generate a comprehensive rainbow table. 
* An _IV (initialization vector)_ is also a public value, used in encryption to produce different ciphertext give the same plaintext and key. As such, it is expected not to be reused across encryptions on the same key, and is sometimes chosen randomly for this reason. Its length is usually a fixed value based on the encryption algorithm and block chaining mode. 

In this case, the implementation is really doing this:

```c
#define IV_SIZE 16 // determined by AES256-CBC
#define SALT_SIZE_IN_BYTES IV_SIZE
```

Because this is a sample, I would choose to make that clear in the code as opposed to defining the IV buffers in terms of salt size.

Also just as FYI, while it's okay in this implementation to use the same value for salt and IV, usually it's clearer to keep them separate. A common construction of this when we don't want to store separate salt and IV values is to:
* take a salt value of chosen length that can be independent of IV length
* use the salt derive a longer value than needed when using the PBKDF (e.g. 512-bits instead of 256-bits)
* use half the bits as the key as originally intended (password key in this case) and the extra bits as the IV.

samples/file-encryptor/enclave/keys.cpp
```diff
@@ -247,7 +204,7 @@ int ecall_dispatcher::cipher_encryption_key(
         &aescontext,
         encrypt ? MBEDTLS_AES_ENCRYPT : MBEDTLS_AES_DECRYPT,
         input_data_size, // input data length in bytes,
-        iv,              // Initialization vector (updated after use)
+        iv,              // Initialization vector (updated afer use)
```
CodeMonkeyLeet, line 207: Typo
```suggestion
        iv,              // Initialization vector (updated after use)
```

samples/file-encryptor/shared.h
```diff
@@ -7,17 +7,20 @@
...
 
 // encryption_header_t contains encryption metadata used for decryption
 // file_data_size: this is the size of the data in an input file, excluding the
-// header digest: this field contains hash value of a password encrypted_key:
-// this is the encrypted version of the encryption key used for encrypting and
-// decrypting the data
+// header digest: this field contains hash value of a password
+// encrypted_key: this is the encrypted version of the encryption key used for
+//                encrypting and decrypting the data
+// salt: The iv/ salt used in PBKDF and AES encryption/ decryption
```
CodeMonkeyLeet, line 17: Wording for clarity:
```suggestion
// salt: The salt value used in deriving the password key.
//       It is also used as the IV for the encryption/decryption of the data.
```

```

File: samples/file-encryptor/enclave/encryptor.cpp [hunk 4]
```diff
@@ -64,7 +47,8 @@ int ecall_dispatcher::initialize(
         goto exit;
     }
     // init iv
-    memcpy(m_operating_iv, m_original_iv, IV_SIZE);
+    memcpy(m_original_iv, header->salt, SALT_SIZE_IN_BYTES);
```
CodeMonkeyLeet, line 50: Is it necessary to keep the extra `m_original_iv` buffer anymore? It looks like the encryptor is always re-initialized given the header anyway, the `m_original_iv` really isn't used anywhere.

CodeMonkeyLeet (commented):
Thanks for cleaning this up! Looks generally good, just a couple of clean-up and educational comments.

samples/file-encryptor/enclave/encryptor.cpp
RRathna: Removed m_original_iv

samples/file-encryptor/shared.h
RRathna: Incorporated suggestion.

samples/file-encryptor/enclave/encryptor.h
RRathna: Moved IV_SIZE to shared.h. Using IV_SIZE in encyptor.cpp, in place of SALT_SIZE_IN_BYTES.

samples/file-encryptor/enclave/keys.cpp
```diff
@@ -61,39 +60,16 @@ int ecall_dispatcher::Sha256(
 }
 
 // This routine uses the mbed_tls library to derive an AES key from the input
-// password and produce a password based key. Note : A set of hardcoded salt
-// values are used here for the purpose simplifying this sample, which caused
-// this routine to return the same key when taking the same password. This saves
-// the sample from having to write salt values to the encryption header. In a
-// real world application, randomly generated salt values are recommended.
+// password and produce a password based key.
```
mingweishih, line 63:
```suggestion
// password and produces a password-based key.
```

samples/file-encryptor/enclave/keys.cpp
```diff
@@ -276,16 +233,45 @@ int ecall_dispatcher::prepare_encryption_header(
...
+    }
+
+    // Generate random salt
+    ret = mbedtls_ctr_drbg_random(&ctr_drbg, salt, SALT_SIZE_IN_BYTES);
+    if (ret != 0)
+    {
+        TRACE_ENCLAVE("mbedtls_ctr_drbg_random() failed with -0x%04x", -ret);
+        goto exit;
+    }
+    memcpy(header->salt, salt, SALT_SIZE_IN_BYTES);
```
mingweishih, line 269: Side note: I just realize that we never check if `header` is null or not. Can you add the check?

samples/file-encryptor/enclave/keys.cpp
```diff
@@ -368,20 +359,25 @@ int ecall_dispatcher::parse_encryption_header(
         goto exit;
     }
 
+    memcpy(salt, header->salt, SALT_SIZE_IN_BYTES);
```
mingweishih, line 362: Same here. Can we check if the header is null in this function as well?

samples/file-encryptor/enclave/encryptor.cpp
```diff
@@ -64,7 +47,7 @@ int ecall_dispatcher::initialize(
         goto exit;
     }
     // init iv
-    memcpy(m_operating_iv, m_original_iv, IV_SIZE);
+    memcpy(m_operating_iv, header->salt, IV_SIZE);
```
mingweishih, line 50: Also check the header is null or not here.
Is there any case that we can't use `m_header`? 
This may be a bit off-topic, can be addressed in a separate PR:
There seems a bit inconsistency between `initialize` and `process_encryption_header`. For example, both of them set `m_encrypt`. Can we have the `initialize` function to initialize `m_passord`, `m_encrypt`, and `m_header` instead?

samples/file-encryptor/enclave/encryptor.h
```diff
@@ -24,8 +22,7 @@ class ecall_dispatcher
     encryption_header_t* m_header;
 
     // initialization vector
-    unsigned char m_original_iv[IV_SIZE];
-    unsigned char m_operating_iv[IV_SIZE];
+    unsigned char m_operating_iv[SALT_SIZE_IN_BYTES];
```
CodeMonkeyLeet, line 25:
```suggestion
    unsigned char m_operating_iv[IV_SIZE];
```

CodeMonkeyLeet (approved):
LGTM with the expectation that you'll clean up Ming-wei's and my remaining comment.

samples/file-encryptor/enclave/keys.cpp
RRathna: incorporated changes.

samples/file-encryptor/enclave/keys.cpp
RRathna: added null check

samples/file-encryptor/enclave/keys.cpp
RRathna: added null check.

samples/file-encryptor/enclave/encryptor.cpp
RRathna: Added null check and moved m_password and m_header initialization to initialize().

mingweishih: Should we use `m_header` instead of `header` here?

samples/file-encryptor/enclave/encryptor.h
RRathna: Thank you! Incorporated change suggested.

samples/file-encryptor/enclave/keys.cpp
```diff
@@ -336,21 +331,31 @@ int ecall_dispatcher::prepare_encryption_header(
...
-//  2)reproduce a encryption key from the password
+//  2)reproduce a password key from the password
 //  3)decrypt the encryption key with a password key
 int ecall_dispatcher::parse_encryption_header(
     encryption_header_t* header,
     string password)
 {
+    if (header == NULL)
+    {
+        TRACE_ENCLAVE("parse_encryption_header() failed with null argument"
```
mingweishih, line 350:
```suggestion
        TRACE_ENCLAVE("parse_encryption_header() failed with a null argument"
```

samples/file-encryptor/enclave/keys.cpp
```diff
@@ -402,11 +413,14 @@ int ecall_dispatcher::process_encryption_header(
     size_t password_len,
     encryption_header_t* header)
 {
-    int ret = 0;
+    if (header == NULL)
+    {
+        TRACE_ENCLAVE("process_encryption_header() failed with null argument"
```
mingweishih, line 418:
```suggestion
        TRACE_ENCLAVE("process_encryption_header() failed with a null argument"
```

samples/file-encryptor/enclave/encryptor.cpp
RRathna: replaced header with m_header.

mingweishih (approved):
LGTM. Thanks for addressing all the comments! Please squash the commits before we can merge the PR.

RRathna:
bors try

mingweishih:
bors try

RRathna:
bors try

RRathna:
bors try

dthaler (approved):

PR #3831, opened 2021-02-09
Update the oeedger8r submodule

### LINKED ISSUE #2390: Use safemath for all edger8r-generated operations
Some code generated by edger8r does potentially unsafe math operations with regards to overflow.

One such place is in tools/oeedger8r/src/Sources.ml:
```
match (size, count) 
| Some s, None -> s
(* TODO: Check that c actually fits in size_t. Also check for overflow,
 * similar to oe_add_size *)
| None, Some c -> sprintf "((size_t)%s * sizeof(%s))" c type_expr
(* TODO: Check that this is an even multiple of the size of type. *)
| Some s, Some c -> sprintf "(%s * %s)" s c
| None, None ->
    sprintf "sizeof(%s%s)" type_expr (get_array_dims decl.array_dims)
```

### LINKED ISSUE #3210: Ensure that oeedger8r generated names are conformant to Developer Guidelines
oeedger8r generates names using older naming convention. e.g: `__oe_ecalls_table`.
This issue is to keep track of the task of changing the emitted names to be conformant to [Developer Guidelines](https://github.com/openenclave/openenclave/blob/master/docs/Community/DevelopmentGuide.md)

### LINKED ISSUE #3513: oeedger8r should be modified to warn when an EDL specification cannot be fully serialized
This is a request that openenclave implements the proposal linked below.

There have been several cases where projects may have avoided edge security issues if there had been warnings as proposed in this specification:
[docs/DesignDocs/full_edl_serialization.md](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/full_edl_serialization.md)

>Although EDL provides sufficient mechanisms for performing full serialization, the generator does not currently enforce these mechanisms. By default, the generator produces edge routines that implicitly copy untrusted memory references into the enclave. This results in potential security flaws and non-portable code.
>
>To overcome these problems, the generator can be modified to warn when an EDL specification cannot be fully serialized.

### LINKED ISSUE #3557: oeedger-cpp ignores invalid syntax checks for "Pointer array not allowed - `p' is a pointer array"
my.edl
```
enclave {
    struct MyStruct {
        int       a1;
        short int a2;
        long int  a3;

        unsigned int       b1;
        unsigned short int b2;
        unsigned long  int b3;

        unsigned       c1;
        unsigned short c2;
        unsigned long  c3;

        short d1;
        long  d2;

        long long          e1;
        unsigned long long e2;

        char          f1;
        unsigned char f2;

        long double   g;
        float f[8];

        char *s;
        struct Foo foo;
    };


    trusted {
        public struct MyStruct ecall1([isptr, in, out] MyStruct s[8]);
        public struct MyStruct ecall2([in, out] MyStruct* p[8]);
    };
```

oeedger-cpp tool's result:
```
Generating edge routine, for the Open Enclave SDK.
Processing my.edl.
Success.
```

oeedger(OCAML) tool's result:
```
Generating edge routines for the Open Enclave SDK.
error: `ecall1': Pointer array not allowed - `s' is a pointer array.

Generating edge routines for the Open Enclave SDK.
error: `ecall2': Pointer array not allowed - `p' is a pointer array.
```

mingweishih:
This PR sync up the oeedger8r against the master branch, which introduces the following changes:
- Enforce the safe multiplication (see the [PR](https://github.com/openenclave/oeedger8r-cpp/pull/31)). Fix #2390
- Add checks against pointer array arguments (see the [PR](https://github.com/openenclave/oeedger8r-cpp/pull/65)). Fix #3557
- Make the oeedger8r-generated variables conform OE guidelines (see the [PR](https://github.com/openenclave/oeedger8r-cpp/pull/66)). Fix #3210
- Add new warning options for the non-serializable cases (see the [PR](https://github.com/openenclave/oeedger8r-cpp/pull/67)). Fix #3513


samples/attestation/common/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ target_link_libraries(
          #
          # Alternatively we could use a CMake OBJECT library, but that
          # requires a newish version of CMake.
-         $<$<PLATFORM_ID:Linux>:-Wl,--require-defined=__oe_ecalls_table>
+         $<$<PLATFORM_ID:Linux>:-Wl,--require-defined=oe_ecalls_table>
```
CodeMonkeyLeet, line 31: This is technically a breaking change if existing apps will not build successfully without changing their build definitions when updated to the next version of OE SDK. 

Consider adding an `__oe_ecalls_table` alias to the renamed `oe_ecalls_table` and add a deprecation warning this release, then dropping the alias in the next release.

include/openenclave/edger8r/common.h
```diff
@@ -66,20 +66,33 @@ OE_INLINE oe_result_t oe_add_size(size_t* total, size_t size)
...
+        }                                        \
+        c = _a * _b;                             \
+    } while (0);
+
+#define OE_ADD_SIZE(total, argcount, argsize)                      \
     do                                                             \
     {                                                              \
-        if (sizeof(total) > sizeof(size_t) && total > OE_SIZE_MAX) \
+        size_t _addend = (size_t)(argsize);                        \
+        if (argcount > 1)                                          \
```
CodeMonkeyLeet, line 86: Consider inverting the dependency instead of redefining the existing signature to include expanded use. It makes the semantics of the new use case clearer, and you don't need to update existing uses to a new macro signature:

```c
#define OE_ADD_ARG_SIZE(total, argcount, argsize) \
    do                                            \
    {                                             \
        size_t _total_argsize = 0;                \

        // Existing logic for OE_COMPUTE_SIZE(argcount, argsize, _total_argsize)

        OE_ADD_SIZE(total, _total_argsize);       \
     } while (0); 
```

include/openenclave/edger8r/common.h
```diff
@@ -90,16 +103,18 @@ OE_INLINE oe_result_t oe_add_size(size_t* total, size_t size)
...
-        if (input_buffer_offset > input_buffer_size)                       \
-        {                                                                  \
-            _result = OE_BUFFER_TOO_SMALL;                                 \
-            goto done;                                                     \
-        }                                                                  \
+#define OE_SET_IN_POINTER(argname, argcount, argsize, argtype)               \
+    if (_pargs_in->argname)                                                  \
+    {                                                                        \
+        size_t _size;                                                        \
+        OE_COMPUTE_SIZE(argcount, argsize, _size);                           \
```
CodeMonkeyLeet, line 110: See factoring comment for `OE_COMPUT_SIZE` and `OE_ADD_SIZE` above.

samples/attestation/common/CMakeLists.txt
anakrish: Good catch Simon. 
`__oe_ecalls_table` is not part of the public API, and therefore renaming it is not a breaking change.

The sample ought not use unpublished symbols in the first place. 
@mingweishih Can you find out why sample was written this way? The original reasons may no longer be valid; and there might be simpler ways to achieve whatever the sample is trying to do with the linker.

mingweishih: This is addressed by #3833

include/openenclave/edger8r/common.h
```diff
@@ -66,20 +66,33 @@ OE_INLINE oe_result_t oe_add_size(size_t* total, size_t size)
     return result;
 }
 
-#define OE_ADD_SIZE(total, size)                                   \
+#define OE_COMPUTE_SIZE(a, b, c)                 \
```
dthaler, line 69: It's not obvious from the signature that a and b are inputs and c is an output.
Consider: Can we either
a) somehow turn this into an inline function where it's more obvious, or 
b) somehow have the macro "return" of c, so you can do `c = OE_COMPUTE_SIZE(a, b);`
?

dthaler (approved):

include/openenclave/edger8r/common.h
mingweishih: Fixed

include/openenclave/edger8r/common.h
mingweishih: Fixed

include/openenclave/edger8r/common.h
mingweishih: The update should address this. Thanks!

anakrish (approved):
LGTM. Thanks for clearing off all these long standing issues!

PR #3833, opened 2021-02-11
Use OBJECT library in data-sealling and attestation samples

mingweishih:
OE has used CMake 3.12 for a while and the OBJECT library was introduced by CMake 2.8.8. Update the data-sealing and attestation samples to use the OBJECT library instead, which addresses the previous comments in the CMakeList.txt files.

### DIFF
--- a/samples/attestation/common/CMakeLists.txt
+++ b/samples/attestation/common/CMakeLists.txt
@@ -11,7 +11,7 @@ add_custom_command(
     ${OE_INCLUDEDIR}/openenclave/edl/sgx)
 
 # Create a library common to each of our two enclaves.
-add_library(common STATIC attestation.cpp crypto.cpp dispatcher.cpp
+add_library(common OBJECT attestation.cpp crypto.cpp dispatcher.cpp
                           ${CMAKE_CURRENT_BINARY_DIR}/attestation_t.c)
 
 if (WIN32)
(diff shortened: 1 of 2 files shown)


CodeMonkeyLeet (commented):
Nice, thanks for updating this!

I thought that the use of the [OBJECT](https://cmake.org/cmake/help/latest/command/add_library.html#id3) libraries in CMake requires a different syntax though i.e. the consuming component like the enclave CMakeLists.txt references this as a named set of obj files in `add_executable(my_exe $<TARGET_OBJECTS:common>)`, rather than as a link target as would be the case for a STATIC library. Can you double check from a clean CMake and rebuild that no other changes are needed?

mingweishih:
> Nice, thanks for updating this!
> 
> I thought that the use of the [OBJECT](https://cmake.org/cmake/help/latest/command/add_library.html#id3) libraries in CMake requires a different syntax though i.e. the consuming component like the enclave CMakeLists.txt references this as a named set of obj files in `add_executable(my_exe $<TARGET_OBJECTS:common>)`, rather than as a link target as would be the case for a STATIC library. Can you double check from a clean CMake and rebuild that no other changes are needed?

Thanks for pointing this out! Just checked, the objects are correctly pulled in. 
After some investigation, I found out that the direct use of OBJECT library in `target_link_libraries` (which is our use case) is supported by CMake from 3.12:

>The target_link_libraries() command now supports Object Libraries. Linking to an object library uses its object files in direct dependents and also propagates usage requirements.

--- from CMake 3.12 [release note](https://cmake.org/cmake/help/latest/release/3.12.html)

mingweishih:
bors try

CodeMonkeyLeet (approved):
Thanks for the explanation! LGTM

RRathna (approved):

PR #3834, opened 2021-02-12
Fixed OP-TEE standalone build

rs--:
Update OP-TEE build files, and documentation
- Updated documentation to include sample
- Fixed standalone build files to properly use compiler flags
- Fixed Linux OP-TEE release build flag
- Temporarily disable linter check; indirectly caused compiler flag issue - quotes around compiler flags cause the flags to not be used

### DIFF
--- a/scripts/check-linters
+++ b/scripts/check-linters
@@ -23,6 +23,11 @@ for i in $filelist; do
     if [[ $i =~ ^3rdparty/ ]]; then
         continue;
     fi
+    # Temporary, to remove later
+    # Quotes around compiler flags break OP-TEE builds
+    if [[ $i =~ ^devex/ ]]; then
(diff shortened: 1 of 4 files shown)


scripts/check-linters
```diff
@@ -23,6 +23,11 @@ for i in $filelist; do
     if [[ $i =~ ^3rdparty/ ]]; then
         continue;
     fi
+    # Temporary, to remove later
```
rs--, line 26: The linter forced quotes around bash variables, which caused the compiler flags to not be applied correctly. A subsequent PR will be created to remove the bypass once this PR is complete.

HernanGatta (approved):

dthaler (approved):

PR #3838, opened 2021-02-16
Design: Debugger Support for Multi-Module Enclaves

anakrish:
Signed-off-by: Anand Krishnamorthi <anakrish@microsoft.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,9 +21,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - -Werror: Turn warnings into errors.
   - -Werror=<option>: Turn the specified warning into an error.
 - oesign sign now allows option -o/--output-file, to specify location to write signature of enclave image.
+- Debugger Contract has been extended to support multiple modules.
+  - Refer to [design document](docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md) for details.
 
 ### Changed
 - oe_get_attestation_certificate_with_evidence() has been deprecated because it has been deemed insufficient for security. Use the new, experimental oe_get_attestation_certificate_with_evidence_v2() instead to generate a self-signed certificate for use in the TLS handshaking process.
+- Debugger Contract
+  - `path` fields in `oe_debug_enclave_t` and `oe_debug_module_t` are now defined to be in
+    UTF-8 encoding. Previously the encoding was undefined. To ensure smooth transition, debuggers
+	are required to try out both UTF-8 as well as the previous encoding and pick the one that works.
 
 ### Security
 - Update mbedTLS to version 2.16.9. Refer to the [2.16.9](https://github.com/ARMmbed/mbedtls/releases/tag/v2.16.9) release notes for the set of issues addressed.
--- a/docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
+++ b/docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
@@ -1,0 +1,461 @@
+Debugger Support for Multi-Module Enclaves
+====
+
+This document describes extensions to the
+[Debugger Contract](/include/openenclave/internal/debugrt/host.h) to support
+debugging enclaves comprised of multiple modules.
+
+Introduction
+------------
+
+Currently, enclaves are built as a single module - a special elf executable that
+is loaded by OE SDK's enclave loader. Since enclaves are loaded by a special
+loader, unlike normal executables and shared libraries that are loaded by the
+platform's loader (e.g /lib64/ld-linux-x86-64.so.2), the debugger does not
+natively know how to load symbols for enclaves. Therefore, the host side OE SDK
+runtime invokes following functions to enable the debugger to locate
+and debug an enclave. The following functions perform necessary house-keeping
+tasks and then call hook functions to interact with the debugger.
+
+- `oe_result_t oe_debug_notify_enclave_created(oe_debug_enclave_t* enclave)`
+
+  This function is called immediately after an enclave has been loaded into
+  memory, but before it has been initialized/begun execution. The debugger uses
+  the  `oe_debug_enclave_t` structure to load the symbols and mark a region of
+  program memory as being taken up by the enclave:
+
+  ```c
+
+    typedef struct _debug_enclave_t
+    {
+        uint64_t magic;
+
+        uint64_t version;
+
+        // Pointer to next enclave
+        // See "Attaching the debugger" section below
+        struct _debug_enclave_t* next;
+
+        const char* path;
+        uint64_t path_length;
+
+        const void* base_address;
+        uint64_t size;
+
+        // Enclave's TCS information.
+        struct _sgx_tcs** tcs_array;
+        uint64_t tcs_count;
+
+        uint64_t flags;
+    } oe_debug_enclave_t;
+
+  ```
+  The `path` and `path_length` fields tell the debugger the path to the
+  enclave's ELF binary. The elf binary is expected to contain the debugging
+  information, if any, for the enclave. `oegdb` currently does not support
+  debugging enclaves whose debugging information has been stripped and moved to
+  a separate .dbg file.
+
+  The `base_address` field tells the debugger the memory address at which the
+  enclave has been loaded. The `size` field is not that important since the
+  debugger can figure out the size of relevant sections fom the ELF binary.
+  The field is retained for making `oe_debug_enclave_t` complete for
+  describing an enclave. The debugger associates the memory region
+  `[base_address, base_address+size)` with the enclave.
+
+- `oe_result_t oe_debug_notify_enclave_terminated(oe_debug_enclave_t* enclave)`
+
+  This function is called immediately after an enclave has been terminated and
+  its memory has been deallocated. The debugger unloads the symbols for the enclave
+  and dissociates the memory region from the enclave.
+
+The debugger gains control and responds to the above notifications in the
+following ways:
+
+- Linux
+
+  The debugger sets hidden breakpoints in the hook functions associated with
+  the above notifications.
+  `oe_debug_notify_enclave_created` performs necessary house-keeping tasks
+  and then calls the following empty hook function:
+  ```c
+  OE_EXPORT void oe_debug_enclave_created_hook(
+      const oe_debug_enclave_t* enclave)
+  {
+  }
+  ```
+
+  Similary, `oe_debug_notify_enclave_terminated` performs necessary
+  house-keeping tasks and calls `oe_debug_notify_enclave_terminated_hook`:
+  ```c
+  OE_EXPORT void oe_debug_enclave_terminated_hook(
+      const oe_debug_enclave_t* enclave)
+  {
+  }
+  ```
+
+  These hook functions are exported so that the debugger can find them even
+  without debug symbols, and set hidden breakpoints. When these functions
+  are called, the breakpoints are hit and the debugger gets control.
+  The debugger then reads the `oe_debug_enclave_t` parameter to handle the
+  notification.
+
+- Windows
+
+  From Visual Studio Debugger team:
+  "In Windows, symbols are almost always found in separate PDB files instead of
+  being embedded within the DLL. And it is easy to get into situations where
+  symbols are not available. That makes relying on the debugger setting
+  breakpoints (via symbol data) a lot more fragile."
+  The debugger is instead notified of events by raising exceptions using the
+  `RaiseException` API.
+
+  `oe_debug_enclave_created_hook` raises the following exception:
+
+  ```c
+        ULONG_PTR args[1] = {(ULONG_PTR)enclave};
+        RaiseException(
+          OE_DEBUGRT_ENCLAVE_CREATED_EVENT,
+          0, // dwFlags
+          1, // number of args
+          args);
+  ```
+
+  whereas `oe_debug_enclave_terminated_hook` raises:
+
+  ```c
+        ULONG_PTR args[1] = {(ULONG_PTR)enclave};
+        RaiseException(
+          OE_DEBUGRT_ENCLAVE_TERMINATED_EVENT,
+          0, // dwFlags
+          1, // number of args
+          args);
+  ```
+
+  The Windows debuggers (Visual Studio and WinDbg) consume the
+  `OE_DEBUGRT_ENCLAVE_CREATED_EVENT` and `OE_DEBUGRT_ENCLAVE_TERMINATED_EVENT`
+  events. Upon receiving the events, the debuggers invoke the appropriate
+  infrastructure to enable debugging the module.
+
+### Attaching the debugger
+
+It is possible to attach the debugger to an application that has already created
+enclaves. In such a scenario, the debugger has missed the chance to be notified
+of enclave creation since it wasn't debugging the program when the enclave was
+created. To allow the debugger to load symbols for enclave after the fact, the
+host-side runtime maintains a global list of loaded enclaves:
+
+When an enclave is loaded, it is added to the list, and when an enclave is
+unloaded, it is removed from the list.
+
+```c
+OE_DEBUGRT_EXPORT extern oe_debug_enclave_t* oe_debug_enclaves_list;
+```
+
+The `oe_debug_enclave_list` is exported so that the debugger can find this list
+without symbols. Upon attaching to a program, the debugger is expected to iterate
+through this list and load symbols for the enclaves in the list. The debugger
+will process the list only once - when attaching to a program or when loading a
+memory dump (see "Analyzing crash-dumps" below).
+
+Adding an enclave to `oe_debug_enclaves_list` and notifying the debugger via the
+hook functions are not atomic operations, and the order in which they are
+performed lead to different edge-cases.
+
+- Order 1
+  1. Add enclave to oe_debug_enclaves_list
+  2. Call oe_debug_enclave_created_hook.
+
+  The debugger may attach to the running program between steps 1 and 2. In this
+  case, upon attaching, the debugger would process the list and load symbols for
+  the enclave. The debugger will be notified again about the same enclave when
+  the hook is called in step 2. This leads to double-notification.
+  
+- Order 2
+  1. Call oe_debug_enclave_created_hook.
+  2. Add enclave to oe_debug_enclaves_list
+
+  The debugger may attach to the running program between steps 1 and 2. In this
+  case, the debugger has missed the hook when it was invoked. The debugger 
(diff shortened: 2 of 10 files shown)


docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,268 @@
...
+natively know how to load symbols for enclaves. Therefore, the host side OE SDK
+runtime runtime invokes following functions to enable the debugger to locate
+and debug an enclave.
+
+- `oe_result_t oe_debug_notify_enclave_created(oe_debug_enclave_t* enclave)`
+
+  This function is called immediately after an enclave has been loaded into
+  memory, but before it has been initialized/begun execution. The debugger uses
+  the following fields in `oe_debug_enclave_t` structure to load the symbols
+  and mark a region of program memory as being take up by the enclave:
```
dthaler, line 24:
```suggestion
  and mark a region of program memory as being taken up by the enclave:
```

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,268 @@
...
+  The `base-address` field tells the debugger the memory address at which the
+  enclave has been loaded. The `size` field is not that important since the
+  debugger is interested only in the size of the `.text` section which it can
+  obtain from the ELF binary. The debugger associates the memory region
+  [base-address, base-address+size) with the enclave.
+
+- `oe_result_t oe_debug_notify_enclave_terminated(oe_debug_enclave_t* enclave)`
+
+  This function is called immediately after an enclave has been terminated and
+  it's memory has been deallocated. The debugger unloads the symbols for the enclave
```
dthaler, line 54:
```suggestion
  its memory has been deallocated. The debugger unloads the symbols for the enclave
```

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,268 @@
...
+- Linux
+
+  The debugger sets hidden breakpoints in the above notification functions. When
+  they are called by the SDK's host-side runtime, the breakpoints are hit and
+  the debugger gets control. The debugger then reads the `oe_debug_enclave_t`
+  parameter to handle the notification.
+
+- Windows
+ 
+  Use of breakpoints by the debugger is not recommended. Instead, the Windows
```
dthaler, line 69: can you say why it's not recommended?  (Recommended by whom?)

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,268 @@
...
+
+
+Use-cases for Multi-Module Enclaves
+-----------------------------------
+
+There are a few implementations of multi-module enclaves based on OE SDK:
+
+- [SGX LKL](https://github.com/lsds/sgx-lkl)
+
+  This project runs a modified linux kernel within an enclave and supports running
```
dthaler, line 107:
```suggestion
  This project runs a modified Linux kernel within an enclave and supports running
```

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,268 @@
...
+  enclaves comprised of executables and shared libraries. This project supports
+  debugging of multi-module enclaves.
+  https://github.com/lsds/sgx-lkl/tree/oe_port/tools/gdb
```
dthaler, line 110: Here and in 116 and 123, it's not clear what the purpose of the link is.  121-122 at least has some text preceding the link, but what are these?

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,268 @@
...
+
+The following `oe_debug_module_t` structure captures the information necessary
+to debug a module:
+
+```c
+
+    typedef struct _debug_module_t
+	{
+	    // Magic value at start of the structure.
+		uint64_t magic;
```
dthaler, line 141: nit: fix indentation to be consistent

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,268 @@
...
+where the enclave has been loaded.
+
+The following notification functions are used to notify the debugger about the
+loading and unloading of a module.
+
+- `oe_result_t oe_result_t oe_debug_notify_module_loaded(oe_debug_module_t* module)`
+
+   This function is called by the SDK's host-runtime after a module has been
+   loaded, but before it has been initialized. A module is initialized by invoking
+   it's constructor functions. It doesn not matter whether the module has been relocated
```
dthaler, line 167:
```suggestion
   its constructor functions. It does not matter whether the module has been relocated
```

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,268 @@
...
+	The enclave side runtime cannot call Windows API functions. Instead, it
+	relies on `int 3` exception to communicate with the debugger. The `int 3`
+	instruction is preceeded by special byte sequence `0x0edeb0gca11bac00` to
+	allow the debugger to distinguish between these special `int 3` exceptions
+	and any other `int 3` exceptions that may occur.
+	
+	`oe_debug_notify_module_loaded` uses the following instruction pattern:
+	```asm
+
+       mov module, %rax
```
dthaler, line 230: The same ELF file should run on both Windows and Linux, so you have to insert this at compile time without know whether it will run on Windows or Linux or both.

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
auott: It's a windows symbol issue. Since PDBs are almost always separate from the DLL it's fairly easy to get into situations where you do not have the symbols available. That makes relying on the debugger setting breakpoints (via symbol data) a lot more fragile. It's much simpler on the debugger side to just receive an event by an exception as that will always work, regardless of whether or not symbols for the oe runtime dlll are available.

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
auott: In the windows case we should be able to rely on symbols for the enclave portion (since they should be embedded in the ELF file). That way both the Linux and Windows in-enclave load can be handled the same way via internal breakpoints.

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,268 @@
...
+	immediately preceeding the program counter. If the bytes match the special
+	sequence, then the debugger reads `rax` register to obtain the pointer to
+	`oe_debug_module_t` structure, and `rbx` register to determine whether
+	the module was loaded or unloaded.
+
+
+The Debugger Contract version is incremented to `2` to reflect the addition of
+the notification functions.
+
+The host-side runtime will set [oe_debugger_contract_version](https://github.com/openenclave/openenclave/blob/6f94547dc920d70eb50ec8ca95ab65a17597364b/include/openenclave/internal/debugrt/host.h#L131)
```
auott, line 261: Note: I don't believe the Windows debuggers will currently do this, though this could definitely be a future behavior.

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,268 @@
...
+
+Debugger Contract Extensions to Support Multi-Module Enclaves
+-------------------------------------------------------------
+
+The following `oe_debug_module_t` structure captures the information necessary
+to debug a module:
+
+```c
+
+    typedef struct _debug_module_t
```
auott, line 138: Similar to the _debug_enclave_t structure, there will need to be a list of the modules that have already been loaded stored somewhere. That way a debugger that attaches can still determine the set of modules without having received the create events for them.

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,268 @@
...
+	instruction is preceeded by special byte sequence `0x0edeb0gca11bac00` to
+	allow the debugger to distinguish between these special `int 3` exceptions
+	and any other `int 3` exceptions that may occur.
+	
+	`oe_debug_notify_module_loaded` uses the following instruction pattern:
+	```asm
+
+       mov module, %rax
+	   mov $1, %rbx     // 1 => ModuleLoaded
+	   jmp do_int3
```
CodeMonkeyLeet, line 232: From the SIG-arch meeting: @binxing had a suggestion about using `nop` and having the debugger overwrite it with the `int 3` as needed, which would make the code path uniform on the enclave side regardless of whether the enclave is under debug. There was an open question as to whether the EDBGWR instruction would fault if the page itself is RX-only.

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,268 @@
...
+
+- `oe_result_t oe_debug_notify_module_unloaded(oe_debug_module_t* module)`
+
+   This function is called by the SDK runtimes (host-side or enclave-side), after a
+   module has been unloaded.
+
+Debugger Handling of Module Load/Unload Notifications
+
+- Linux
+  Similar to enclave load/unload, the debugger sets internal breakpoints in
```
CodeMonkeyLeet, line 183: From the SIG-arch meeting, there was a discussion as to whether this applied to all loaded modules, including the primary enclave app module, and the expectation was that it would exclude the primary enclave module (which would have the equivalent tasks handled as part of the enclave start notification).

The feedback was that it would be nice if the primary enclave could be handled uniformly, with whatever additional tasks specific to enclave start (e.g. TCS handling) being factored out into a separate implementation that is composable with the module load. There was some concern that this could break the existing contract with the windbg, and that would need to be investigated.

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
MWShan: EDBGWR ignores the EPCM RWX permissions (it does require that the enclave is loaded in debug mode).  Thus, the OS interfaces (WriteProcessMemory for Windows and ptrace for Linux) should work the same for when a debugger writes to non-enclave memory.

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
dthaler: @auott But enclaves are ELFs not DLLs so why are PDBs relevant for multi-module enclaves?

auott: It depends on where the load notification is coming from - not what is being loaded.
If the load notification comes from a DLL (similar to how the enclave load notification happens) then we have the PDB separation problem. If the notification is coming from inside an enclave, then we don't have the PDB problem.

My understanding was that both those scenarios are possible, though my previous response was about notifications coming from outside an enclave. For the latter case I don't see any issue with relying on symbols.

anakrish: One thing to keep in mind is that the contract supports Intel SDK's PE enclaves too. 
Instead of relying on debug symbols, would it be OK to rely on exported symbols that are guaranteed to be available even if debug symbols have been stripped off in a binary?
Then we could have both the ELF and PE enclaves export the functions and debugger ought to be able to put breakpoints on these even without debug symbols. However in that case, we'd also need the parameter to passed in a specific register so that the debugger can just read it without requiring debug symbols for the parameter.

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
anakrish: While, at first glance, it may seem nicer to treat primary and secondary modules the same, and factor out extra details of the primary module into a separate notification, it does not seem that much valuable to me for the following reasons:
- We are losing data in this abstraction. The fact that a module is a primary module now has to be figured out in another way. 
- An enclave that does not use multiple modules will now require extra notifications to enable debugging where previously it used to do that with fewer notifications.
- The information about the primary module will now be sent via multiple notifications instead of being atomically sent in a single notification. I suspect this could make it difficult for the debugger to correlate information if multiple threads were simultaneously creating enclaves.
- This change is a breaking change that will require a change to all the debuggers (oegdb, VS, WinDbg) to debug single module enclaves. Additionally, it might also require work in Intel SDK to preserve backwards compatibility. It is not clear to me what the bar is for justifying breaking changes in published contracts.

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
auott: I think that would be work as well.
The register problem would likely need to be solved anyways as putting a breakpoint based on a symbol would run into the same problem as the breakpoint would likely be at the beginning of the prolog.

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
anakrish: Austen and I discussed this. While this is an elegant proposal, it seems like it is quite close to what the debugger does underneath the hood for a normal breakpoint. Is there any benefit this approach would have in comparison to a hidden breakpoint?

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,379 @@
...
+
+        const char* path;
+        uint64_t path_length;
+
+        const void* base_address;
+        uint64_t size;
+
+        // Enclave's TCS information.
+        struct _sgx_tcs** tcs_array;
+        uint64_t num_tcs;
```
dthaler, line 46:
```suggestion
        uint64_t tcs_count;
```
Per OE style guidelines

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,379 @@
...
+    struct _debug_enclave_t* next;
+
+    const char* path;
+    uint64_t path_length;
+
+    const void* base_address;
+    uint64_t size;
+
+    struct _sgx_tcs** tcs_array;
+    uint64_t num_tcs;
```
dthaler, line 357:
```suggestion
    uint64_t tcs_count;
```
or tcs_array_size

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,379 @@
...
+
+    uint64_t flags;
+
+	// New field added to the end
+	oe_debug_module_t secondary_modules;
```
dthaler, line 362: nit: fix indentation

docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md
```diff
@@ -0,0 +1,366 @@
...
+        char* path;
+        uint64_t path_length;
+
+        // Memory region where the enclave has been loaded.
+        uint64_t base_address;
+        uint64_t size;
+
+        // Next item in list of module.
+        struct _debug_module_t* next;
+    } oe_debug_module_t;
```
MWShan, line 218: If an oe_debug_module_t structure is added to the linked list of modules before the oe_debug_notify_module_loaded() function is called and then removed after the oe_debug_notify_module_unloaded() function is called, then a debugger that attaches to a running process could get an event on a module that it already sees as loaded, or miss an event on a module that has been unloaded.

anakrish: Thanks for pointing this out Mark. As always, I appreciate your insights.

The following three operations are themselves not atomic, and also cannot be performed together atomically:
- Check if the debugger has been attached or not
- Notify the debugger
- Add an entry to the module list

Due to the lack of atomicity, the ordering of the operations can result in different edge-cases.
(The "check if debugger has been attached or not" is ignored from the scenarios below since debugger may immediately attach/detach after the check has been performed; making the check unreliable).

In one ordering, the debugger may get double notification:
```
    1. Runtime adds module to list
        <---- Debugger attaches to the program and processes the module
    2. Runtime notifies the debugger about the module <---- Debugger will have to avoid processing the module again
    
```

In another approach, the debugger may lose track of modules
```
    1. Runtime notifies the debugger  <-- notification ignored since debugger not attached
      <------ Debugger attaches to the program, processes the list and finds no module
    2. Runtime adds the module to the list  <--- list has already been processed; module is "lost"
```

I don't know if there is a way to solve the edge-case elegantly. We could go with the first ordering above and require that the debugger be prepared for double-notifications for parsed modules upon attach.

Note: The same ordering problem exists for the list of enclaves. I'd have to dig up my notes/mails, but IIRC, this edge-case was determined not too cost-effective to address given the timing constraints of the debugger contract.

anakrish: I discussed with @auott, the strategy that was used for handling the list of enclaves what that the debugger ought to handle potential double-notification during attach.

anakrish:
@MWShan @dthaler @CodeMonkeyLeet @mingweishih 
I have updated the PR and added more details. Can you take a look?

CHANGELOG.md
```diff
@@ -21,6 +21,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - -Werror: Turn warnings into errors.
   - -Werror=<option>: Turn the specified warning into an error.
 - oesign sign now allows option -o/--output-file, to specify location to write signature of enclave image.
+- Debugger Contract has been extended to support multiple modules.
+  - Refer to [design document](docs/DesignDocs/DebuggerSupportForMultiModuleEnclaves.md) for details.
+  - `path` fields in `oe_debug_enclave_t` and `oe_debug_module_t` are now defined to be in
```
CodeMonkeyLeet, line 26: Consider moving this under the `### Changed` header. While the feature is largely additional, the specific modification here that requires developer attention might be better surfaced there.

CodeMonkeyLeet (approved):

radhikaj (approved):

MWShan (approved):

PR #3855, opened 2021-02-25
Fix double free on OE_INVALID_IMAGE

thomasten:
If one tries to load an invalid ELF image, _unload_elf_image() is
called by _load_elf_image() and then again by
oe_load_elf_enclave_image(). Clear the struct in _unload_elf_image() so
that it does not call free on already freed pointers in the second
invocation.


anakrish (approved):
LGTM

@mrragava Can you also review this?

mingweishih (approved):
LGTM

mrragava:
LGTM

tests/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ if (OE_SGX)
   set(DEFINE_OE_SGX "-DOE_SGX")
 endif ()
 
+add_subdirectory(invalid_image)
```
mingweishih, line 32: It turns out the the test is unix-specific. You would need to move the test as part of https://github.com/openenclave/openenclave/blob/dbd3e7ad0c0e4b374b3c63e860ca2698d9984d52/tests/CMakeLists.txt#L151

thomasten: AArch64 fails, so I guess it's SGX-specific? I moved the test to OE_SGX and clarified the comment in the test. Please try again.

PR #3862, opened 2021-03-01
Guard d_reclen value in syscall readdir

manojrupireddy:
Host can set d_reclen to an arbitary value. Check the validity of d_reclen value after syscall.

### DIFF
--- a/syscall/devices/hostfs/hostfs.c
+++ b/syscall/devices/hostfs/hostfs.c
@@ -1132,6 +1132,11 @@ static struct oe_dirent* _hostfs_readdir(oe_fd_t* desc)
     if (retval != 0)
         OE_RAISE_ERRNO(OE_EINVAL);
 
+    /* Guard the special case that a host sets an arbitrarily value for
+     * d_reclen. */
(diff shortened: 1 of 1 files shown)


syscall/devices/hostfs/hostfs.c
```diff
@@ -1132,6 +1132,11 @@ static struct oe_dirent* _hostfs_readdir(oe_fd_t* desc)
     if (retval != 0)
         OE_RAISE_ERRNO(OE_EINVAL);
 
+    /* Guard the special case that a host sets an arbitrarily value for
```
mingweishih, line 1135: LGTM. It will be good if we can call out that this behavior is OE-specific; i.e., OE expects the `d_reclen` to always be `sizeof(struct oe_dirent)` while this may be true in Linux (https://www.man7.org/linux/man-pages/man3/readdir.3.html).

mrragava:
LGTM

anakrish (approved):
LGTM

mingweishih (approved):

PR #3863, opened 2021-03-01
[oesgx] epc subleaf memory calculation

RRathna:
correcting EPC memory calculation by iterating over ecx to find out if _CPUID() passes, and subleaf exists.

### DIFF
--- a/tools/oesgx/oesgx.c
+++ b/tools/oesgx/oesgx.c
@@ -65,7 +65,6 @@ static int _CPUID(Regs* regs)
     // Check if no sub-leaves are supported
     if (regs->eax == 0 && regs->ebx == 0 && regs->ecx == 0 && regs->edx == 0)
     {
-        printf("Error getting CPUID. Returned: %d", regs->eax);
         result = 1;
     }
     return result;
@@ -152,9 +151,9 @@ int main(int argc, const char* argv[])
 
     /* Enumeration of Intel SGX Capabilities: figure out EPC size */
     {
-        Regs regs = {SGX_CAPABILITY_ENUMERATION, 0, 0x2, 0};
+        unsigned int ecx = 0x2;
+        Regs regs = {SGX_CAPABILITY_ENUMERATION, 0, ecx, 0};
         uint64_t epc_size = 0;
-
         result = _CPUID(&regs);
         if (result)
         {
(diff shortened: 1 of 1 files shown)


tools/oesgx/oesgx.c
```diff
@@ -162,15 +163,24 @@ int main(int argc, const char* argv[])
...
+
+        while(!result && HAVE_EPC_SUBLEAF(regs))
+        {
+            epc_size +=
+                ((regs.ecx & 0x0fffff000) |
+                 ((uint64_t)(regs.edx & 0x0fffff) << 32));
+
+            ecx++;
+            regs.ecx = ecx;
+            result = _CPUID(&regs);
```
anakrish, line 182: We need to let the user know in case this _CPUID call fails.

RRathna: The _CPUID() call in the line 159 is being reported to the user. The _CPUID() call in line 182 is to check if ecx = 0x3 (or higher) correspond to a subleaf. Is there a way to find out the number of subleafs?

anakrish: Can the call in line 182 fail?

RRathna: yes, and the failure is the exit condition from the while loop..

RRathna:
Changes were tested on two devices as show below, with help from @tehsiuhuang.

daperoni@daperoni-t-n2:~/openenclave-install-2/bin$ ./oesgx
CPU supports SGX_FLC:Flexible Launch Control
CPU supports Software Guard Extensions:SGX1
MaxEnclaveSize_64: 2^(36)
CPU supports Key Sharing & Separation (KSS): false
EPC size on the platform: 58720256

acctestuser@teh-epc-yto:~/openenclave-install-2/bin$ ./oesgx
CPU supports SGX_FLC:Flexible Launch Control
CPU supports Software Guard Extensions:SGX1
MaxEnclaveSize_64: 2^(36)
CPU supports Key Sharing & Separation (KSS): false
EPC size on the platform: 117440512

tehsiuhuang (approved):

tools/oesgx/oesgx.c
RRathna: Added comments in src.

anakrish (approved):
LGTM

PR #3864, opened 2021-03-02
Separate image builds into stages

CyanDevs:
This separates out the Azure Managed Instance image builds into individual stages so that retries can occur at the stage level rather than repeating the entire process.

Minor changes:
* The retry sleep period has been decreased from 5 minutes to 1 minute each try.
* Ansible runs often fail waiting for apt lock from cache updates. The retry limit has been increased to 100 which almost guarantees the lock will be released sometime during the task.

This also sets the stage for some additional work for addition of Ubuntu 20.04 and deprecation of Windows Server 2016.

### DIFF
--- a/.jenkins/infrastructure/build_azure_managed_images.Jenkinsfile
+++ b/.jenkins/infrastructure/build_azure_managed_images.Jenkinsfile
@@ -23,7 +23,6 @@ AZURE_IMAGES_MAP = [
     ]
 ]
 
-
 def get_image_version() {
     def now = LocalDateTime.now()
     return (now.format(DateTimeFormatter.ofPattern("yyyy")) + "." + \
@@ -35,18 +34,21 @@ def get_image_id() {
     if (params.IMAGE_ID) {
         return params.IMAGE_ID
     }
+    checkout scm
     def last_commit_id = sh(script: "git rev-parse --short HEAD", returnStdout: true).tokenize().last()
     return (get_image_version() + "-" + last_commit_id)
 }
 
+
+
 def buildLinuxManagedImage(String os_type, String version) {
     node(params.AGENTS_LABEL) {
-        stage("${os_type}-${version}") {
-            timeout(GLOBAL_TIMEOUT_MINUTES) {
-                cleanWs()
-                checkout scm
-                def managed_image_name_id = get_image_id()
-                def gallery_image_version = get_image_version()
+        def managed_image_name_id = get_image_id()
+        def gallery_image_version = get_image_version()
+        withEnv(["DOCKER_REGISTRY=${OETOOLS_REPO}",
+            "MANAGED_IMAGE_NAME_ID=${managed_image_name_id}",
+            "GALLERY_IMAGE_VERSION=${gallery_image_version}"]) {
+            stage("${os_type}-${version}-cleanup") {
                 def az_cleanup_existing_image_version_script = """
                     az login --service-principal -u \$SERVICE_PRINCIPAL_ID -p \$SERVICE_PRINCIPAL_PASSWORD --tenant \$TENANT_ID
                     az account set -s \$SUBSCRIPTION_ID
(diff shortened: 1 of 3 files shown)


radhikaj:
@CyanDevs thanks for this improvement. Can the commits be squashed!

CyanDevs:
Done, thanks @radhikaj !

mingweishih (approved):

.jenkins/infrastructure/build_azure_managed_images.Jenkinsfile
```diff
@@ -84,36 +86,42 @@ def buildLinuxManagedImage(String os_type, String version) {
...
-                        az account set -s \$SUBSCRIPTION_ID
-                    """
-                    def az_build_managed_img_script = """
-                        source ${WORKSPACE}/.jenkins/infrastructure/provision/utils.sh
+                oe.azureEnvironment(az_rg_create_script, params.OE_DEPLOY_IMAGE)
+            }
 
+            try {
+                stage("${img_name_suffix}-provisioning") {
+                    def provision_script = """
```
rs--, line 116: Is it possible/worth it having these types of multi-line commands in a script instead? Having it in a script seems easier to maintain, unless that type of workflow doesn't work, or is too much trouble.

scripts/ansible/roles/linux/common/tasks/apt-repo.yml
```diff
@@ -2,11 +2,12 @@
 # Licensed under the MIT License.
 
 ---
+
 - name: Install apt-transport-https APT package
   apt:
     name: apt-transport-https
     state: latest
-  retries: 10
+  retries: 100
```
rs--, line 10: Does the 40th, 50th, etc, try eventually work or does this just delay failure?

.jenkins/infrastructure/build_azure_managed_images.Jenkinsfile
CyanDevs: Interesting idea to think about. Thanks @rs-- !

As we use a global library function to run commands/scripts inside an Azure environment and the command/script needs to be passed as a parameter each time, the consequences of splitting into individual stages is that the big script that we started with also has to be broke up. The advantage is that each part of the script can be ran separately and retried as needed.

It is then possible to move these functions into separate scripts by treating them like libraries similar to this: https://github.com/openenclave/openenclave-ci/blob/master/src/jenkins/common/Openenclave.groovy
The downside to that is that adds some overhead as the libraries need to then be loaded in with the Jenkins master and/or in the Jenkinsfile, which then becomes a trade off between which files/configuration do we want to maintain. I can't say for sure which one would be easier to maintain, but currently I think having the scripts in this single Jenkinsfile at least makes it easier to find and read. Would you agree with that?

scripts/ansible/roles/linux/common/tasks/apt-repo.yml
CyanDevs: From what I see in the logs, it usually succeeds before the 30th try, so it does eventually work. I have never seen it fail after setting it to 100 retries, which gives the tasks to wait up to 500 seconds for the apt lock to be released.

Example from a build:
```
    azure-arm: TASK [linux/common : Install apt-transport-https APT package] ******************

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (100 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (99 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (98 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (97 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (96 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (95 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (94 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (93 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (92 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (91 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (90 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (89 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (88 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (87 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (86 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (85 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (84 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (83 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (82 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (81 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (80 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (79 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (78 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (77 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (76 retries left).

    azure-arm: FAILED - RETRYING: Install apt-transport-https APT package (75 retries left).

    azure-arm: [WARNING]: Updating cache and auto-installing missing dependency: python-apt

    azure-arm: changed: [default]

    azure-arm:
    ```

.jenkins/infrastructure/build_azure_managed_images.Jenkinsfile
rs--: I'd say that if the functionality was generic enough like provisioning a VM from scratch, it would be a benefit to have it split out since we could use it in other places (like when we have to manually create a VM). But if we're only doing this here, we can just leave it, and split it out in the future when the functionality is needed elsewhere.

rs-- (approved):

.jenkins/infrastructure/build_azure_managed_images.Jenkinsfile
CyanDevs: A good plan. Thanks!

PR #3869, opened 2021-03-04
Wait for other apt processes first

CyanDevs:
install-ansible.sh can fail when there is an existing process using apt. This has been happening for a number of E2E tests.

E.g.
```
+ sudo bash scripts/ansible/install-ansible.sh
Hit:1 http://azure.archive.ubuntu.com/ubuntu bionic InRelease
Hit:2 http://azure.archive.ubuntu.com/ubuntu bionic-updates InRelease
Hit:3 http://azure.archive.ubuntu.com/ubuntu bionic-backports InRelease
Get:4 http://security.ubuntu.com/ubuntu bionic-security InRelease [88.7 kB]
Fetched 88.7 kB in 0s (257 kB/s)
Reading package lists...
E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
script returned exit code 100
```

### DIFF
--- a/scripts/ansible/install-ansible.sh
+++ b/scripts/ansible/install-ansible.sh
@@ -7,11 +7,30 @@ set -o errexit
 
 DIR=$(dirname "$0")
 
+function wait-apt-get {
+    i=0
+    echo "Checking for locks..."
+    # Check for locks
+    while fuser /var/lib/dpkg/lock > /dev/null 2>&1 ||
+          fuser /var/lib/dpkg/lock-frontend > /dev/null 2>&1 ||
+          fuser /var/lib/apt/lists/lock > /dev/null 2>&1; do
+        # Wait up to 600 seconds to lock to be released
+        if (( i > 600 )); then
+            echo "Timeout waiting for lock."
+            exit 1
+        fi
+        echo "Waiting for apt/dpkg locks..."
+        i=$((i++))
+        sleep 1
+    done
+    apt-get "${@}"
+}
+
 if which yum > /dev/null; then
     yum install git python3-pip -y
 elif which apt-get > /dev/null; then
-    apt-get update
-    apt-get install libssl-dev libffi-dev python3-pip -y
+    wait-apt-get update
+    wait-apt-get install libssl-dev libffi-dev python3-pip -y
 else
     echo "ERROR: Only these package managers are supported: yum, apt-get"
     exit 1


CyanDevs:
bors try

CyanDevs:
bors try

scripts/ansible/install-ansible.sh
```diff
@@ -10,6 +10,10 @@ DIR=$(dirname "$0")
 if which yum > /dev/null; then
     yum install git python3-pip -y
 elif which apt-get > /dev/null; then
+    echo "Checking for dpkg locks..."
+    while fuser /var/lib/dpkg/lock > /dev/null 2>&1 || fuser /var/lib/dpkg/lock-frontend > /dev/null 2>&1; do
```
rs--, line 14: Is there a chance this loop gets stuck without another exit condition/max retries?

CyanDevs: Given that it's a while loop, absolutely. But that would be due to a host issue where a process maintaining use of those lock files and the subsequent apt update command would fail.

On Jenkins we have timeouts for our builds so that would not result in a stuck build. But for users running this manually it is possible to get stuck.

If we want to do something about this situation, I could add a timeout and fail the entire script if the while loop never exits. What do you think of that?

CyanDevs:
bors try

anakrish (approved):
LGTM

scripts/ansible/install-ansible.sh
```diff
@@ -10,6 +10,10 @@ DIR=$(dirname "$0")
...
+        sleep 1
+    done
```
anakrish, line 16: There is a subtle issue here. Say the loop terminates and we finish line 16, before line 17 is executed, another process in theory could execute apt, and hold the lock, causing line 17 to fail.
Such situations are common in multi-threaded programs; not sure if it is something we need to guard against in the ansible's use of apt.

anakrish: One approach is to use _flock_ to decide which process gets to use apt. Lock is acquired, apt is used and then lock is released.

CyanDevs: This is true and it's a good point. The current mitigation for this scenario is that the Jenkins agent that this script runs on only allows one job at a time and then recycled afterwards, so there are no parallel install processes waiting in queue at the same time. Though I like this idea so I will try implementing it and see how it goes.

anakrish (approved):
LGTM. Nice solution

CyanDevs:
Unfortunately, this failed on CI as it did not bother see that there was a lock. Googling around it appears that apt and dpkg do not use flock to manage their file locks. Source: https://askubuntu.com/questions/960746/why-does-apt-not-display-an-error-message-even-if-flock-is-used

fuser seems to be an effective way to determine if apt-get is holding a lock so I will add that, and see if I can incorporate flock to maintain a lock between the time fuser runs and apt-get.

scripts/ansible/install-ansible.sh
CyanDevs: I couldn't solve this issue with flock as apt/dpkg doesn't use it. See https://github.com/openenclave/openenclave/pull/3869#issuecomment-791226643 for more info.

scripts/ansible/install-ansible.sh
CyanDevs: Implemented a 10 minute timeout now.

scripts/ansible/install-ansible.sh
```diff
@@ -7,11 +7,30 @@ set -o errexit
...
+        # Wait up to 600 seconds to lock to be released
+        if (( $i > 600 )); then
+            echo "Timeout waiting for lock."
+            exit 1
+        fi
+        echo "Waiting for apt/dpkg locks..."
+        ((i++))
+        sleep 1
+    done
+    apt-get ${1}
```
mingweishih, line 26: This will take only the first argument, should we use `$@` instead?

scripts/ansible/install-ansible.sh
rs--: Works for me, thanks!

scripts/ansible/install-ansible.sh
CyanDevs: Absolutely, thanks for the good catch!

mingweishih (approved):
LGTM. Thanks for the fix!

mingweishih:
bors delegate=CyanDevs

PR #3872, opened 2021-03-06
Added CodeQL Scans

rs--:
The scan is configured by a [config file](https://github.com/openenclave/openenclave-security/blob/main/src/static/codeql/queries/cpp/suites/oe-codeql-security-queries.qls) on the [openenclave-security repo](https://github.com/openenclave/openenclave-security).

The CodeQL scan is triggered:
- nightly on master
- PRs to master

To minimize the runtime of the Github Action, the Open Enclave SDK is built with the `-DBUILD_TESTS=OFF` flag.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,46 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+name: "CodeQL"
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-18.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'cpp' ]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: ${{ matrix.language }}
+        config-file: .github/codeql/codeql-config.yml
+
+    - name: Install Dependencies
+      run: |
+        sudo apt install -y doxygen
+        git submodule update --init --recursive
+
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,46 @@
...
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: ${{ matrix.language }}
+        config-file: openenclave/openenclave-security/src/static/codeql/queries/cpp/suites/oe-codeql-security-queries.qls@main
```
mrragava, line 31: Looks like the scan is using openenclave specific queries for some reason. Can we use codeql-config.yml instead.

rs--: Sure thing, has that config been checked into the oe-security repo? I'd prefer to have those files in that repo instead of the sdk repo

mrragava: > @mrragava Looking at the contents of the oe-codeql-security-queries.qls, it looks like the intention is that it should be running the OE-specific queries as well, so it doesn't seem like a targeting issue. Is the suggestion that this is misconfigured and should be using .yml files and not .qls files?

I'm not seeing OE-Queries being used in scan results in this PR checks tab.
oe-codeql-security-queries.qls is configured to scan both OE and CodeQL queries but the scan doesnt seems to pick it.
As per CodeQL we need to provide codeql-config.yml which points to query suites.
https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-codeql-code-scanning-in-your-ci-system

Refer:
https://github.com/mrragava/openenclave/blob/master/.github/codeql/codeql-config.yml

mingweishih:
> /lgtm
> /approved

Prow has not been deployed yet. You still need manually approve the PR 😄

mingweishih (approved):
LGTM

rs--:
bors try

anakrish (approved):
LGTM

PR #3874, opened 2021-03-09
Proposal for the FIPS crypto module loading for SGX

mingweishih:

### DIFF
--- a/docs/DesignDocs/CryptoModuleLoadingSupport.md
+++ b/docs/DesignDocs/CryptoModuleLoadingSupport.md
@@ -1,0 +1,372 @@
+Support for FIPS-Certified Cryptographic Module Loading in SGX Enclaves
+======
+
+Introduction
+------------
+
+An immediate requirement for developers to build production SGX enclaves with the OE SDK is
+making the enclaves conform to industry standards. One such standard is Federal Information
+Processing Standard 140 (FIPS 140). To be FIPS 140 compliant, an enclave must perform
+certain cryptographic operations through a FIPS-validated module. The module usually
+takes the form of a shared library. Once loaded, the module performs a series of self-tests
+(also known as power-up self-tests, or POST), which validates both the integrity of the module
+and the correctness of the cryptographic operations.
+
+Objective
+---------
+
+This proposal aims to add the support of loading a FIPS-validated module in the OE
+SDK in a timely manner, which allows an enclave to use the module and conforms to the FIPS requirement.
+To this end, the proposed design should introduce minimal, incremental changes to the OE loader
+that supports loading a single cryptographic module instead of general shared libraries.
+More specifically, the design targets the implementation of a module based on
+[SymCrypt](https://github.com/microsoft/SymCrypt).
+
+While multiple loader implementations exist---e.g., the [dynamic linker](https://git.musl-libc.org/cgit/musl/tree/ldso/dynlink.c)
+from musl---they may require more design considerations and code changes, which are not necessary for
+supporting only the cryptographic module loading. However, such implementation would still
+be ideal for the long term if we plan to generally support shared libraries loading in OE.
+Unless there are more use cases show up, we leave this as future work and discuss possible options in
+the [Longer-term Loader Design](#long-term-loader-design) section.
+
+Note that the proposal targets SGX only, which introduces modifications to the loader that
+OE has implemented. OP-TEE has its own loader, which is not part of OE, and therefore requires
+different design. We will leave this as future work.
+
+User Experience
+------
+
+- Module Usage
+
+  A developer can use the functions from the module in an enclave program as those from
+  static libraries. Also, the developer should expect the initialization (e.g., POST)
+  and termination functions of the module are automatically invoked by the enclave runtime.
+
+- Build
+
+  The developer should compile the enclave program as usual. During the link-time, the developer
+  should specify the module (e.g., `libsymcrypt.so`) in the command line as the following example.
+
+  ```
+  clang -o enclave enc.o libsymcrypt.so -loeenclave -loelibc -oecore
+  ```
+
+  Note that the module is required to be put under the same directory as the enclave binary.
+  Moreover, both the `rpath` and `runpath` linker options *should not* be used.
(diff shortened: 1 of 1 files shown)


docs/DesignDocs/CryptoModuleLoadingSupport.md
```diff
@@ -0,0 +1,348 @@
...
+[SymCrypt](https://github.com/microsoft/SymCrypt).
+
+While multiple loader implementations exist---e.g., the [dynamic linker](https://git.musl-libc.org/cgit/musl/tree/ldso/dynlink.c)
+from musl---they may require more design considerations and code changes, which are not necessary for
+supporting only the cryptographic module loading. However, such implementation would still
+be ideal for the long term if we plan to generally support shared libraries loading in OE.
+Unless there are more use cases show up, we leave this as future work and discuss possible options in
+the [Alternative Loader Implementations](#alternative-loader-implementations) section.
+
+Note that the proposal targets SGX only, which introduces modifications to the loader that
```
dthaler, line 32: Please update the title in line 1 to say SGX then

radhikaj (approved):

anakrish (approved):
LGTM. Thanks for the focused proposal addressing FIPS crypto needs and starting points for long-term loader plans.

CodeMonkeyLeet (commented):
Scope of the design change seems reasonable to me although you may want to consider expanding the discussion on the failure handling codepaths, particularly if they related to specific feature scoping decisions that need to be reflected to the developer (e.g. expected error codes/messages for unsupported relocations, test scenarios for them etc.)

On minor bug that needs to be additionally addressed is that `dl_iterate_phdr` in oelibc will be incorrect once there is more that one module in the enclave (it will only enumerate the primary enclave). It would be worth tracking as a separate issue since it's slightly orthogonal to the content of this design, and easy to overlook.

anakrish:
Good point @CodeMonkeyLeet. Thanks for pointing out `dl_iterate_phdr`. 

Due to the `dl` prefix, it seemed as if this function belongs in `libdl`, and I'd overlooked it since we don't support `libdl`.
But good that this belongs in `libc` instead, which implies we ought to be able to support it correctly.

This ought to be the recommended way for iterating over modules within an enclave.
@mingweishih Would `dl_iterate_phdr` provide access to all the information for FIPS CMAC computation?

docs/DesignDocs/CryptoModuleLoadingSupport.md
```diff
@@ -0,0 +1,372 @@
...
+certain cryptographic operations through a FIPS-validated module. The module usually
+takes the form of a shared library. Once loaded, the module performs a series of self-tests
+(also known as power-up self-tests, or POST), which validates both the integrity of the module
+and the correctness of the cryptographic operations.
+
+Objective
+---------
+
+This proposal aims to add the support of loading a FIPS-validated module in the OE
+SDK in a timely manner, which allows an enclave to use the module and conforms the FIPS requirement.
```
dthaler, line 19:
```suggestion
SDK in a timely manner, which allows an enclave to use the module and conform to the FIPS requirement.
```

docs/DesignDocs/CryptoModuleLoadingSupport.md
```diff
@@ -0,0 +1,372 @@
...
+
+  ```
+  clang -o enclave enc.o libsymcrypt.so -loeenclave -loelibc -oecore
+  ```
+
+  Note that the module is required to be put under the same directory as the enclave binary.
+  Moreover, both the `rpath` and `runpath` linker options *should not* be used.
+  See [Module Lookup](#module-lookup) for more detail.
+  The OE Make settings (i.e., `pkgconfig`) does not use these options by default and
+  CMake settings explicitly opts out these options by setting the `CMAKE_SKIP_RPATH` to `TRUE`.
```
dthaler, line 58:
```suggestion
  CMake settings explicitly opts out of these options by setting the `CMAKE_SKIP_RPATH` to `TRUE`.
```

docs/DesignDocs/CryptoModuleLoadingSupport.md
```diff
@@ -0,0 +1,372 @@
...
+
+  Again, this assumes that the `enclave/libsymcrypto.so` exists.
+  For cases such as the shared library is not found in the enclave directory or the library
+  is not supported (see the [Specification](#specification) for more detail), the loader
+  aborts the execution and prints proper messages.
+
+Specification
+------
+
+- The design supports a dynamically linked enclave binary against a *single* share library.
```
dthaler, line 85:
```suggestion
- The design supports a dynamically linked enclave binary against a *single* shared library.
```

docs/DesignDocs/CryptoModuleLoadingSupport.md
```diff
@@ -0,0 +1,372 @@
...
+
+  For a shared library with unsupported properties, the loader will explicitly error out.
+- The design retains the existing behavior if the enclave is statically linked.
+- The design targets SGX only.
+
+Design
+------
+
+Instead of supporting dynamic loading or `dlopen`, which requires Enclave Dynamic Memory Management
+(EDMM) feature in SGX2 that is not available yet, the idea of the design is supporting static loading
```
dthaler, line 102:
```suggestion
(EDMM) in SGX2 that is not available yet, the idea of the design is supporting static loading
```

dthaler (approved):

mingweishih:
> Good point @CodeMonkeyLeet. Thanks for pointing out `dl_iterate_phdr`.
> 
> Due to the `dl` prefix, it seemed as if this function belongs in `libdl`, and I'd overlooked it since we don't support `libdl`.
> But good that this belongs in `libc` instead, which implies we ought to be able to support it correctly.
> 
> This ought to be the recommended way for iterating over modules within an enclave.
> @mingweishih Would `dl_iterate_phdr` provide access to all the information for FIPS CMAC computation?

`dl_iterate_phdr` will be a callback to the module, so it may not be used to perform integrity test. However, I'll have another issue to track the problem or address it as part of actual implementation to the proposal.

PR #3886, opened 2021-03-16
Update VSIX include paths to put app's includes first

### LINKED ISSUE #3868: OpenSSL headers in OE package can break apps
Since openssl was not supported in earlier versions, applications that did the work themselves to use openssl within their application can break with OE v0.14 since the headers appear in the SDK and are thus in the path even when OE is using mbedtls.  That is, if the app uses one openssl version and the OE openssl package is a different version, the headers won't match.

This is particularly an issue with the VS extension which forces the OE path to be before any app-supplied include paths.

One possible fix might be to fix the VS extension to put the OE path at the end (instead of the beginning) of the include search path, so an app can easily override it.  Another possible approach would be to put the openssl headers somewhere you have to opt-in to get via a different path.

dthaler:
This allows the app to override the OE include files, e.g., to use their own version of OpenSSL.

Fixes #3868

### DIFF
--- a/devex/vsextension/msbuild/open-enclave-cross.targets
+++ b/devex/vsextension/msbuild/open-enclave-cross.targets
@@ -249,12 +249,12 @@
     <!-- Additional Include Directories for Enclaves and Hosts -->
     <ItemDefinitionGroup Condition="$(OEIsLinux) == False">
         <ClCompile>
-            <AdditionalIncludeDirectories>$(OEIncludePath);%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+            <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories);$(OEIncludePath)</AdditionalIncludeDirectories>
         </ClCompile>
     </ItemDefinitionGroup>
     <ItemDefinitionGroup Condition="$(OEIsLinux) == True">
         <ClCompile>
-            <AdditionalIncludeDirectories>$(OEIncludePath);/opt/openenclave/include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+            <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories);$(OEIncludePath);/opt/openenclave/include</AdditionalIncludeDirectories>
         </ClCompile>
     </ItemDefinitionGroup>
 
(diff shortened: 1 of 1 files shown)


anakrish:
While this code change looks trivial, there seems to be some fundamental problems with the implementation of the extension:

1. The mbedtls is hard-coded in the list of libraries. https://github.com/openenclave/openenclave/blob/82bff7df3d98b0ce89f3ae0f2eaa21e2fd5e9cfc/devex/vsextension/msbuild/open-enclave-cross.targets#L347
   - In [Release v0.13](https://github.com/openenclave/openenclave/releases/tag/v0.13.0) OpenSSL support was added, and the notion of a default crypto implementation was introduced. The VS extension hard-codes mbedtls, and it is not clear how a user could use openssl instead of mbedtls for OE's internal cryptographic operations.
   
 2. Minor NIT: I found the names confusing and hard to follow. For example, the element `OESgxLibsT` refers to Intel SGX trusted libs, and don't have anything to do with OE SDK's SGX libraries.
  https://github.com/openenclave/openenclave/blob/82bff7df3d98b0ce89f3ae0f2eaa21e2fd5e9cfc/devex/vsextension/msbuild/open-enclave-cross.targets#L200
  
    On the other hand, `OEIntelSgxLibsU` correctly indicates that it is a list of Intel untrusted libraries. However, the OE prefix does not seem to fit since these are not OE libraries.
 https://github.com/openenclave/openenclave/blob/82bff7df3d98b0ce89f3ae0f2eaa21e2fd5e9cfc/devex/vsextension/msbuild/open-enclave-cross.targets#L204

anakrish (changes requested):
I'm blocking this since there is a fundamental problem with the hard-coding of crypto libraries.
And it is not clear how the user can switch to OpenSSL throughout the enclave. OE's FIPS crypto is based on OpenSSL interface and the ability to switch to OpenSSL is thus important.

dthaler:
> I'm blocking this since there is a fundamental problem with the hard-coding of crypto libraries.
> And it is not clear how the user can switch to OpenSSL throughout the enclave. OE's FIPS crypto is based on OpenSSL interface and the ability to switch to OpenSSL is thus important.

I believe all of your comments relate to code before this change and so I think they should be treated as a separate issue, not a reason to block this PR.

anakrish:
Unblocking this.

anakrish (approved):
LGTM

mingweishih (approved):

anakrish:
@CyanDevs Seems to be another instance of the sporadic submodule update failure.

PR #3894, opened 2021-03-20
Module loading Support

mingweishih:
The PR implements the module loading based on the [Support for FIPS-Certified Cryptographic Module Loading in SGX Enclaves](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/CryptoModuleLoadingSupport.md).

The PR consists of two parts:
- Add the support to the loader and enclave runtime to support the module loading
- Add both positive and negative tests for the implementation


mingweishih:
bors try

enclave/core/init_fini.c
```diff
@@ -111,14 +133,35 @@ void oe_call_init_functions(void)
...
+
+void oe_call_fini_functions(void)
+{
     extern void (*__fini_array_start)(void);
     extern void (*__fini_array_end)(void);
+    const uint64_t baseaddr = (uint64_t)__oe_get_enclave_base();
+    const oe_enclave_module_info_t* module_info = oe_get_module_info();
 
-    for (fn = &__fini_array_end - 1; fn >= &__fini_array_start; fn--)
+    _call_fini_functions(&__fini_array_start, &__fini_array_end);
```
anakrish, line 155: This ordering is fine for FIPS. For general multi-module loading, we'd need to think about this carefully. Lot of the core functionality is provided in `oecore`, and we'd want to make sure that that is the root-most module so that it's destructors, if any, get called last.
Maybe that means that our current static libraries ought to be converted to shared libraries. Needs more thought.

enclave/core/sgx/globals.c
```diff
@@ -269,3 +269,17 @@ uint64_t oe_get_num_pages(void)
...
 }
+
+/*
+**==============================================================================
+**
+** Information for the module.
+**
+**==============================================================================
+*/
+OE_EXPORT const volatile oe_enclave_module_info_t _module_info;
```
anakrish, line 280: Good idea to make this const so that it gets assigned to ReadOnly pages.

enclave/core/sgx/reloc.c
```diff
@@ -40,7 +40,10 @@ bool oe_apply_relocations(void)
         /* Relocate the reference */
         if (reloc_type == R_X86_64_RELATIVE)
         {
-            *dest = (uint64_t)(baseaddr + p->r_addend);
+            int64_t addend = p->r_addend;
```
anakrish, line 43: FYI, one relocation type that I encountered when trying out many static libraries was
`R_X86_64_NONE`. This is a noop reloc. The linker was able to optimize away the relocation, but not the entry itself.
We need to implement this only if the final version of FIPS crypto has such a relocation.

enclave/core/sgx/reloc.c
```diff
@@ -40,7 +40,10 @@ bool oe_apply_relocations(void)
...
+            /* Skip if the symbol is undefined. */
```
anakrish, line 44:
```suggestion
            /* Process only if the symbol is defined. */
```

host/sgx/loadelf.c
```diff
@@ -50,6 +52,8 @@ static oe_result_t _unload_image(oe_enclave_image_t* image)
     if (image)
     {
         _unload_elf_image(&image->elf);
+        if (image->module)
```
anakrish, line 55: Should we unload the module first and then the image?

host/sgx/loadelf.c
```diff
@@ -834,6 +1019,155 @@ static oe_result_t _sgx_update_enclave_properties(
...
+    number_of_entries = section_size / sizeof(elf64_dyn_t);
+    for (uint64_t i = 0; i < number_of_entries; i++)
+    {
+        /* Explicitly prevent the use of DT_RPATH and DT_RUNPATH that affects
+         * the enclave measurement */
+        if (section_data[i].d_tag == DT_RPATH ||
+            section_data[i].d_tag == DT_RUNPATH)
+            OE_RAISE_MSG(
+                OE_UNEXPECTED,
+                "RPATH or RUNPATH should not be used in the enclave binary",
```
anakrish, line 1063: We can specify the compiler/linker/cmake flags to avoid RPATH/RUNPATH in the error message

anakrish (approved):
LGTM. Great work!

host/sgx/loadelf.c
mingweishih: This should be fine as the base image (referenced by image->elf) and the module (image->module) are loaded separately. We just need to make sure unloading both of them before memset the image itself (line 57).

mingweishih:
bors try

radhikaj (approved):

radhikaj (approved):

host/sgx/elf.c
```diff
@@ -1913,63 +1899,117 @@ oe_result_t elf64_load_relocations(
...
-        if (!(*data_out = oe_memalign(OE_PAGE_SIZE, *size_out)))
-        {
-            *size_out = 0;
-            goto done;
-        }
+    memset(*data_out, 0, *size_out);
+    OE_CHECK(oe_memcpy_s(*data_out, *size_out, dyn_data, dyn_size));
+    if (plt_data)
+        OE_CHECK(oe_memcpy_s(
+            (void*)((uint64_t)*data_out + dyn_size),
```
radhikaj, line 1972: Would it be better to use safe arithmetic here and at line 1983?

mingweishih:
bors try

host/sgx/loadelf.c
```diff
@@ -834,6 +1025,161 @@ static oe_result_t _sgx_update_enclave_properties(
...
+
+    number_of_entries = section_size / sizeof(elf64_dyn_t);
+    for (uint64_t i = 0; i < number_of_entries; i++)
+    {
+        /* Explicitly prevent the use of DT_RPATH and DT_RUNPATH that affects
+         * the enclave measurement */
+        if (section_data[i].d_tag == DT_RPATH ||
+            section_data[i].d_tag == DT_RUNPATH)
+            OE_RAISE_MSG(
+                OE_UNEXPECTED,
```
CodeMonkeyLeet, line 1069: Consider a specific  `OE_UNSUPPORTED_ENCLAVE_IMAGE` in cases where there are OE expectations being violated

host/sgx/loadelf.c
```diff
@@ -834,6 +1025,161 @@ static oe_result_t _sgx_update_enclave_properties(
...
+
+    if (elf64_find_section(
+            &image->elf.elf,
+            ".dynamic",
+            (uint8_t**)&section_data,
+            &section_size) != 0)
+        goto done;
+
+    if (!section_data || !section_size)
+        OE_RAISE(OE_UNEXPECTED);
```
CodeMonkeyLeet, line 1059: Here an elsewhere with ELF parsing errors (e.g. lines 1082, 1087), consider `OE_INVALID_IMAGE` instead of the more generic `OE_UNEXPECTED`.

include/openenclave/internal/load.h
```diff
@@ -89,6 +90,9 @@ struct _oe_enclave_image
      * other enclave binary formats are supported later */
     oe_enclave_elf_image_t elf;
 
+    /* Pointer to the image of a module. */
+    oe_enclave_elf_image_t* module;
```
CodeMonkeyLeet, line 94: Consider a more informative comment and name for `module`. For example
```c
/* Pointer to dependency image for the enclave
 * Only up to one such .so dependency is currently allowed */
oe_enclave_elf_image_t* submodule;
```

host/sgx/loadelf.c
```diff
@@ -695,6 +715,171 @@ static oe_result_t _set_uint64_t_dynamic_symbol_value(
...
+                symbol_definition.st_shndx != SHN_UNDEF)
+            {
+                p->r_addend =
+                    (int64_t)(
+                        dependency->image_rva + symbol_definition.st_value) +
+                    addend;
+            }
+            else
+            {
+                OE_TRACE_WARNING("symbol %s not found\n", name);
```
CodeMonkeyLeet, line 790: This should be a hard failure in your 2 binary model, no?

host/sgx/loadelf.c
```diff
@@ -834,6 +1025,161 @@ static oe_result_t _sgx_update_enclave_properties(
...
+    else /* Handle the case if the module path does not include directories */
+        snprintf(path_name, path_size, "%s", n);
+
+    *module_path = path_name;
+
+done:
+    return result;
+}
+
+static oe_result_t _load_module_image(
```
CodeMonkeyLeet, line 1122: Again, consider renaming to more descriptive names e.g. `_load_dependency_image`, `_find_dependency_image` etc.

host/sgx/loadelf.c
```diff
@@ -834,6 +1025,161 @@ static oe_result_t _sgx_update_enclave_properties(
...
+            section_data[i].d_tag == DT_RUNPATH)
+            OE_RAISE_MSG(
+                OE_UNEXPECTED,
+                "RPATH or RUNPATH should not be used in the enclave binary",
+                NULL);
+
+        if (section_data[i].d_tag == DT_STRTAB)
+            strtab_offset = section_data[i].d_un.d_ptr;
+        else if (section_data[i].d_tag == DT_NEEDED)
+            needed_offset = section_data[i].d_un.d_val;
```
CodeMonkeyLeet, line 1076: This logic doesn't trap the case where more than one dependency is specified and will only try to use the last found `DT_NEEDED` entry. It should fail with an explicit error if more than one dependency is needed.

host/sgx/loadelf.c
mingweishih: This allows weak symbol to be unresolved. For example
```
__attribute__((weak)) int sub(int a, int b);
```

tests/module/module_negative_path/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+add_enclave_library(module_negative_path SHARED ../module/module.c)
```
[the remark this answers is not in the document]
mingweishih: Added comments. The current loader logic only looks up the module in the same directory as the enclave binary. This test is intended to create the library under the different directory.

tests/CMakeLists.txt
```diff
@@ -108,6 +108,7 @@ if (UNIX
     add_subdirectory(libcxxrt)
     add_subdirectory(libunwind)
     add_subdirectory(mbed)
+    add_subdirectory(module)
```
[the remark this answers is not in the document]
mingweishih: Updated.

host/sgx/elf.c
```diff
@@ -1913,63 +1899,117 @@ oe_result_t elf64_load_relocations(
...
+        *size_out = 0;
+
+    if (!_is_valid_elf64(elf) || !data_out || !size_out)
+        goto done;
+
+    OE_CHECK(_elf64_load_relocations(elf, ".rela.dyn", &dyn_data, &dyn_size));
+    OE_CHECK(_elf64_load_relocations(elf, ".rela.plt", &plt_data, &plt_size));
+
+    /* Ensure the size is not zero and the content of .rela.dyn is not empty */
+    size = dyn_size + plt_size;
```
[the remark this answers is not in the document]
mingweishih: Done.

host/sgx/elf.c
```diff
@@ -1913,63 +1899,117 @@ oe_result_t elf64_load_relocations(
...
+        goto done;
+    }
+
     /* Make a copy of the relocation section (zero-padded to page size) */
+    *size_out = oe_round_up_to_page_size(size);
+
+    if (!(*data_out = oe_memalign(OE_PAGE_SIZE, *size_out)))
     {
-        *size_out = oe_round_up_to_page_size(size);
+        *size_out = 0;
```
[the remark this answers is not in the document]
mingweishih: Fixed.

host/sgx/loadelf.c
mingweishih: Updated.

include/openenclave/internal/load.h
mingweishih: Done

host/sgx/loadelf.c
mingweishih: Updated.

host/sgx/loadelf.c
mingweishih: Added logic to handle this case

host/sgx/loadelf.c
mingweishih: This allows for unresolved weak symbols. For example
```
__attribute__((weak)) int sub(int a, int b);
```

mingweishih:
bors try

host/sgx/loadelf.c
```diff
@@ -695,6 +715,171 @@ static oe_result_t _set_uint64_t_dynamic_symbol_value(
...
+        }
+    }
+
+    result = OE_OK;
+
+done:
+    return result;
+}
+
+static oe_result_t _patch_module_elf_image(
```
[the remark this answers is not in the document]
mingweishih: done

host/sgx/loadelf.c
mingweishih: done

mingweishih:
bors try

mingweishih:
bors try-

host/sgx/loadelf.c
CodeMonkeyLeet: I think you should be able to determine unresolved weak symbols (unused) at static link time and filter for those (they should have a zero value) instead of trying to find the symbol and allowing any unresolved symbol to be only a warning. Weak symbols should only be relevant at static link time, should be marked by the zero value if unused by runtime binding. 

For comparison, I would check that this code behaves the same when given a non-weak symbol that was dynamically linked with `LD_BIND_NOW` enabled in a non-enclave app.

host/sgx/loadelf.c
```diff
@@ -695,57 +728,170 @@ static oe_result_t _set_uint64_t_dynamic_symbol_value(
...
 
-    /* reloc right after image */
+    /* reloc right after image
+     * Note that the image_info holds the non-padded size. */
     oeprops->image_info.reloc_rva = image->image_size;
     oeprops->image_info.reloc_size = image->reloc_size;
+    if (module_image)
+    {
+        oeprops->image_info.reloc_rva += module_image->image_size;
+        oeprops->image_info.reloc_size += module_image->reloc_size;
```
CodeMonkeyLeet, line 850: This is the value prior to page alignment though, shouldn't this be patching to the page-aligned size?

host/sgx/loadelf.c
```diff
@@ -695,57 +728,170 @@ static oe_result_t _set_uint64_t_dynamic_symbol_value(
...
+    }
     OE_CHECK(_set_uint64_t_dynamic_symbol_value(
         image, "_reloc_rva", oeprops->image_info.reloc_rva));
     OE_CHECK(_set_uint64_t_dynamic_symbol_value(
         image, "_reloc_size", oeprops->image_info.reloc_size));
 
-    /* heap is right after all the relocs */
+    /* Merge the relocation data from both base and module images and
```
CodeMonkeyLeet, line 857: Consider breaking this reloc merge out into a separate function that can be run before the patching and `_patch_elf_image` could continue to enforces its original expectations.

Also, is the extra copy and sizing necessary in the case where there is no dependency image? I believe that the requirement is that the original implementation already requires that.

CodeMonkeyLeet (approved):
LGTM, just pay attention to the page-alignment ordering in `_patch_elf_image` and fix that if needed.

mingweishih:
bors try

tests/module_loading/module_negative_path/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+# Using the default setting such that the library will be generated
```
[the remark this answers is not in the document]
mingweishih: hmm.. I tried the option. It turns out that the copied library still has the same `SONAME` so that loader will look up the original library 😢 As a result, we still need to recompile the binary so the the name can be different.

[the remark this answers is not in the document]
CodeMonkeyLeet: Are you trying to run the test against the same enclave as the success case? i.e. you're binplacing the dependency twice, but not primary enclave bin? It's not really important to optimize this though.

host/sgx/loadelf.c
mingweishih: I updated the PR and make the relocation patching (merging + padding) right after loading the image. This allows for aligning more with existing behavior.

host/sgx/loadelf.c
mingweishih: The relocation  patching step now is separated from the original image patching.

mingweishih:
bors try

tests/module_loading/module_negative_path/CMakeLists.txt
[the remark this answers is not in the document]
mingweishih: nvm. Issue resolved 😄

host/sgx/loadelf.c
mingweishih: Now forcing the error is the unresolved symbol not a weak symbol.

PR #3898, opened 2021-03-22
Add 0-base SGX Enclave design doc

bodzhang:
Signed-off-by: Bo Zhang (ACC) <zhanb@microsoft.com>


docs/DesignDocs/SGX_0-base_Enclave.md
```diff
@@ -0,0 +1,228 @@
...
+to `oe_sgx_enclave_config` and `_oesgx_load_context`.
+
+```C
+typedef struct oe_sgx_enclave_config_t
+{
+    uint16_t product_id;
+    uint16_t security_version;
+
+    /* Flag to indicate 0-base enclave, keeping alignment */
+    uint32_t flag_0_base;
```
dthaler, line 151: why put the flag here instead of into the attributes field like other flags?

dthaler: Answered during the meeting that you need a separate set of flags in addition to the 64 bits consumed by hardware.  In that case, just make this one bit instead of 32:
```suggestion
    struct {
        int zero_base : 1;
    } flags;
```

docs/DesignDocs/SGX_0-base_Enclave.md
```diff
@@ -0,0 +1,228 @@
...
+
+    uint8_t family_id[16];
+    uint8_t extended_product_id[16];
+    /* (OE_SGX_FLAGS_DEBUG | OE_SGX_FLAGS_MODE64BIT | OE_SGX_FLAGS_KSS) */
+    uint64_t attributes;
```
dthaler, line 156: This is bad name design, since the field is called "attributes" but the defines are _FLAGS_ and so that communicates that the values do not belong in this field.   One of the two should be changed going forward.

docs/DesignDocs/SGX_0-base_Enclave.md
CodeMonkeyLeet, line 151: Follow up to SIG-arch discussion: given that `start_addr` must never be zero when `flag_0_base` is set, is there any scenario where we want to support a fixed `start_addr` without 0-base as measured in the properties?

If not, can this be inferred from `start_addr != 0`?

docs/DesignDocs/SGX_0-base_Enclave.md
```diff
@@ -0,0 +1,228 @@
...
+the Enclave's `SECS.SIZE` value.
+
+Running multiple instance of a single 0-base Enclave in a single process won't
+be possible, as each instance would has to be loaded at exactly the same
+[`ENCLAVE_START`, `ENCLAVE_END`] range to maintain the Enclave measurement,
+which is not possible within a single process.
+
+## Support SGX Enclave with SECS.BASEADDR set to 0
+
+To support "0-base" SGX Enclave with `SECS.BASEADDR`=0, an flag indicating
```
CodeMonkeyLeet, line 140: We should also document the expected design changes in the enclave runtime for enforcing the new `start_addr`

docs/DesignDocs/SGX_0-base_Enclave.md
```diff
@@ -0,0 +1,228 @@
...
+Linux, the threshold is typically 4KB or 64KB, and configurable by the system
+admin. For Windows, the threshold is 64KB. Selecting a `ENCLAVE_START` value
+much higher than the typical threshold, especially for Linux, makes sure the
+enclave can be loaded on most systems.  The Enclave Loader does not need to
+reserve the linear address range between `SECS.BASEADDR`=0 and `SECS.SIZE` for
+the Enclave. It should reserve the memory range starting from `ENCLAVE_START`,
+in this case, predefined at Enclave signing time, with a size smaller than
+`SECS.SIZE`.
+
+It's possible to support multiple 0-base Enclaves in a single process. The SGX
```
CodeMonkeyLeet, line 109: We should verify that Windows permits this and explicitly scope out supporting Windows if it does not.

docs/DesignDocs/SGX_0-base_Enclave.md
```diff
@@ -0,0 +1,228 @@
...
+    uint8_t family_id[16];
+    uint8_t extended_product_id[16];
+    /* (OE_SGX_FLAGS_DEBUG | OE_SGX_FLAGS_MODE64BIT | OE_SGX_FLAGS_KSS) */
+    uint64_t attributes;
+
+    /* XSave Feature Request Mask */
+    uint64_t xfrm;
+
+    /* 0-base enclave start addr */
+    uint64_t start_addr;
```
CodeMonkeyLeet, line 162: We should add guidance in documentation & samples about how devs are expected to set this value and the rationale for it (e.g. info from Anand that clang speculative load hardening expects that the first 2GB is reserved)

docs/DesignDocs/SGX_0-base_Enclave.md
```diff
@@ -0,0 +1,228 @@
...
+calculation for `EADD` instruction is handled in the `libsgx_enclave_common.so`
+layer or the SGX driver layer. In the OE SDK, only
+`oe_sgx_measure_load_enclave_data(...)` call flow used during enclave signing
+needs to be adjusted to address the offset calculation. In the SGX `TCS` page,
+several fields are defined as offset value from `SECS.BASEADDR`. OE SDK's
+`_add_control_pages(...)` function needs to take into account that 0-base
+enclave's `SECS.BASEADDR` is 0, not `start_addr`, when filling in the SGX `TCS`
+page.
+
+## Authors
```
CodeMonkeyLeet, line 226: The design proposal template suggests an Alternatives section, it might be worthwhile to add a paragraph just talking about other options in this space that were rejected (OS changes etc.)

docs/DesignDocs/SGX_0-base_Enclave.md
```diff
@@ -0,0 +1,228 @@
...
+Guaranteed Page Fault on NULL pointer dereference is also the expectation of many
+legacy applications.
+
+## SGX Enclave App Memory Layout and NULL Pointer Dereference Behavior
+
+In a typical SGX application, virtual address 0 is not part of the SGX Enclave
+linear address range of [`SECS.BASEADDR`, `SECS.BASEADDR` + `SECS.SIZE` -1]. The
+typical linear address memory layout of a SGX enclave is shown below. The blue
+pages are committed (through EADD or EAUG/EACCEPT instruction) Enclave pages.
+The red pages marked as Guard Page(s) are within the Enclave linear address
```
CodeMonkeyLeet, line 26: Consider adding a section that points out that the existing use of Guard Pages as unallocated pages in enclave to trigger #PF will need to be modified with the upcoming EDMM work. The added paging support would need to reject (i.e. not EACCEPT) page-ins of what should be guard page ranges by a malicious host OS.

CodeMonkeyLeet (commented):
Generally looks good, capturing some comments from SIG-arch

docs/DesignDocs/SGX_0-base_Enclave.md
bodzhang: Added "Implication for future sGX2.0 EDMM support" section in the updated PR.

docs/DesignDocs/SGX_0-base_Enclave.md
bodzhang: Added in the updated PR.

docs/DesignDocs/SGX_0-base_Enclave.md
bodzhang: Adopted the suggestion in the updated PR.

docs/DesignDocs/SGX_0-base_Enclave.md
bodzhang: This is legacy code, out of the scope of this design doc PR. Will address it separately

docs/DesignDocs/SGX_0-base_Enclave.md
bodzhang: Called out the need to provide instructions and samples to help the developer of 0-base Enclave to select the proper value for `ECLAVE_START` in the updated PR.

docs/DesignDocs/SGX_0-base_Enclave.md
bodzhang: Added a paragraph in the updated PR to clarify that no POC was implemented on Windows yet and Windows support would have to wait, if the existing Windows kernel APIs for SGX Enclave won't allow "0-base".

docs/DesignDocs/SGX_0-base_Enclave.md
bodzhang: A predetermined `start_addr` might have value outside of the "0-base" Enclave context, for example, to support loading legacy non-position-independent apps in the Enclave (libOS model) at specific linear address. A **preferred** `start_addr` might also make sense. I think having a flag and a `start_addr` allows more flexibility in the future.  The comments for `start_addr` were changed in the updated PR.

docs/DesignDocs/SGX_0-base_Enclave.md
bodzhang: Added a "Alternative" section in the updated PR.

anakrish (commented):
The proposal looks good to me. I'd like the following addressed before approving.

- The enclave start address is part of the measurement of a 0-base SGX enclave. This would incur some change to oesign tool or the enclave config schema. The natural way would be to add  `EnclaveBaseAddress`, `ZeroBased` fields to the config. Another approach would be to add command-line parameters to oesign. The design needs to recommend the preferred changes to enclave signing workflow.

- Even for a zero-based enclave, only the memory region [ENCLAVE-START, ENCLAVE-END] is reserved in the process' address space. It is not clear how the SDK ensures that the marshalling buffers are not allocated by the OS in the memory region [0, ENCLAVE-START] which become inaccessible to the 0-based enclave.

Future direction:
In the past, there have been efforts to "sandbox" an enclave to make sure that it does not accidentally access/write to host memory. Previous approaches involved inserting bounds-check instructions for memory accesses; resulting in performance costs not practical for real-world use. This proposal introduces an interesting side-effect that an enclave cannot accidentally access/write to memory below ENCLAVE-START; doing so results in a page fault.
Something to consider for the future would be to see if the PAGES above ENCLAVE-END, other than marshalling buffers, could be made into guard-pages.

anakrish (approved):
Thanks for addressing my feedback.

Nice section about two way sandboxing.

Regarding ensuring that marshalling buffers are allocated after the enclave end, Mike's approach of adding  "-fpie -Wl,-Ttext-segment,0x10000000" to the host seems reasonable.
I don't have an solid alternative suggestion right now; we could maybe have edger8r ensure that marshalling buffers lie after enclave end.

MWShan (approved):
Nice design.

I like Anand's suggestion about employing the Edger8r to ensure that the marshalling structure is declared outside of the enclave range, but the host may also need some facility to verify user-check pointer buffers, too.

mingweishih (approved):

PR #3900, opened 2021-03-22
Add supported v3 collateral version number

### LINKED ISSUE #3885: Ctest for the attestation failed
When I build with LVI mitigation, run:
```
cmake .. \
-DLVI_MITIGATION=ControlFlow \
-DLVI_MITIGATION_BINDIR=/usr/local/lvi-mitigation/bin
make
```
I got the result:
```
[ 15%] Linking CXX executable safecrt_enc-lvi-cfg
/usr/bin/ld: cannot find -link-lvi-mitigation
/usr/bin/ld: CMakeFiles/safecrt_enc-lvi-cfg.dir/__/common/test.cpp.o: unable to initialize decompress status for section .debug_str
/usr/bin/ld: CMakeFiles/safecrt_enc-lvi-cfg.dir/__/common/test.cpp.o: unable to initialize decompress status for section .debug_str
CMakeFiles/safecrt_enc-lvi-cfg.dir/__/common/test.cpp.o: file not recognized: File format not recognized
clang: error: linker command failed with exit code 1 (use -v to see invocation)
tests/safecrt/enc/CMakeFiles/safecrt_enc-lvi-cfg.dir/build.make:149: recipe for target 'tests/safecrt/enc/safecrt_enc-lvi-cfg' failed
make[2]: *** [tests/safecrt/enc/safecrt_enc-lvi-cfg] Error 1
CMakeFiles/Makefile2:8668: recipe for target 'tests/safecrt/enc/CMakeFiles/safecrt_enc-lvi-cfg.dir/all' failed
make[1]: *** [tests/safecrt/enc/CMakeFiles/safecrt_enc-lvi-cfg.dir/all] Error 2
Makefile:181: recipe for target 'all' failed
make: *** [all] Error 2
```
It works on Ubuntu18.04, SGX1+FLC model
My environment path is:
```
/home/dylan/.local/bin:/home/dylan/bin:.:/usr/local/jdk1.8/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
```
I have installed the  dependencies according to the guidance document, contents under the director `/usr/local/lvi-mitigation/bin` are:
```
drwxr-xr-x 2 root root  4096 Mar 15 14:51 ./
drwxr-xr-x 4 root root  4096 Mar 14 11:38 ../
lrwxrwxrwx 1 root root    57 Mar 14 11:38 as -> /usr/local/lvi-mitigation/external/toolset/ubuntu18.04/as*
-rwxr-xr-x 1 root root   231 Mar 15 19:49 clang-8*
-rwxr-xr-x 1 root root   233 Mar 15 19:49 clang++-8*
lrwxrwxrwx 1 root root    16 Mar 14 11:38 clang-8_symlink -> /usr/bin/clang-8*
lrwxrwxrwx 1 root root    18 Mar 14 11:38 clang++-8_symlink -> /usr/bin/clang++-8*
-rwxr-xr-x 1 root root   227 Mar 15 19:49 g++*
-rwxr-xr-x 1 root root   227 Mar 15 19:49 gcc*
lrwxrwxrwx 1 root root    12 Mar 14 11:38 gcc_symlink -> /usr/bin/gcc*
-r-xr-xr-x 1 root root   660 Mar 14 11:38 generate_wrapper*
lrwxrwxrwx 1 root root    12 Mar 14 11:39 g++_symlink -> /usr/bin/g++*
-r-xr-xr-x 1 root root  2649 Mar 14 11:38 install_lvi_mitigation_bindir*
-r-xr-xr-x 1 root root  1093 Mar 14 11:38 invoke_compiler*
lrwxrwxrwx 1 root root    57 Mar 14 11:38 ld -> /usr/local/lvi-mitigation/external/toolset/ubuntu18.04/ld*
lrwxrwxrwx 1 root root    62 Mar 15 14:50 ld.gold -> /usr/local/lvi-mitigation/external/toolset/ubuntu18.04/ld.gold*
-r-xr-xr-x 1 root root 11514 Mar 14 11:38 lvi-mitigation.py*
lrwxrwxrwx 1 root root    62 Mar 15 14:51 objdump -> /usr/local/lvi-mitigation/external/toolset/ubuntu18.04/objdump*
```
There is another error when I first run `make`
```
/usr/bin/as: unrecognized option '-mlfence-before-indirect-branch=register'
```
Then I create a soft link for `as` in the directory `/usr/bin` like, it works.
```
dylan@dylan:/usr/bin$ ll | grep as
lrwxrwxrwx  1 root  root          57 Mar 15 17:28 as -> /usr/local/lvi-mitigation/external/toolset/ubuntu18.04/as*
```
But  after `make` again, I got the result showed at the beginning: `[ 15%] Linking CXX executable safecrt_enc-lvi-cfg
/usr/bin/ld: cannot find -link-lvi-mitigation`

qiucwang:
According to Intel [SGX ECDSA QuoteLibReference page 20](https://download.01.org/intel-sgx/sgx-dcap/1.10/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf):
"See the sgx_ql_qve_collateral_t definition.
The ‘version’ field of the gx_ql_qve_collateral_t structure will reflect the version of the
PCCS/PCS API used to retrieve the collateral. For V1 and V2 APIs, the ‘version’ field with have
a value of 1. For V3 APIs, the ‘version’ field will have the value of 3."

Current OESDK support v3 collateral. But it only treat 1 as valid collateral version number. Need to be compatible with default SGX QPL.

### DIFF
--- a/host/sgx/sgxquoteprovider.c
+++ b/host/sgx/sgxquoteprovider.c
@@ -91,12 +91,12 @@ oe_result_t oe_get_sgx_quote_verification_collateral(
         OE_RAISE(OE_QUOTE_PROVIDER_CALL_ERROR);
     }
 
-    if (collateral->version != SGX_QL_QVE_COLLATERAL_VERSION)
+    if (collateral->version != SGX_QL_QVE_COLLATERAL_VERSION_1 &&
+        collateral->version != SGX_QL_QVE_COLLATERAL_VERSION_3)
     {
         OE_RAISE_MSG(
             OE_INVALID_ENDORSEMENT,
-            "Expected version to be %d, but got %d",
-            SGX_QL_QVE_COLLATERAL_VERSION,
+            "Invalid collateral version %d",
             collateral->version);
     }
 
--- a/host/sgx/sgxquoteprovider.h
+++ b/host/sgx/sgxquoteprovider.h
@@ -31,8 +31,14 @@ typedef struct _oe_sgx_quote_provider
 #define SGX_QL_FREE_QUOTE_VERIFICATION_COLLATERAL_NAME \
     "sgx_ql_free_quote_verification_collateral"
 
-/*! Version of the supported SGX quote verification collateral  */
-#define SGX_QL_QVE_COLLATERAL_VERSION (1)
+/**
+ * Version of the supported SGX quote verification collateral, which will
+ * reflect the version of the PCCS/PCS API used to retrieve the collateral.
+ * For V1 and V2 APIs, the ‘version’ field with have a value of 1.
+ * For V3 APIs, the ‘version’ field will have the value of 3."
+ */
+#define SGX_QL_QVE_COLLATERAL_VERSION_1 (1)
+#define SGX_QL_QVE_COLLATERAL_VERSION_3 (3)
 
 OE_EXTERNC_END
 


common/sgx/collateral.c
```diff
@@ -194,6 +194,32 @@ static oe_result_t _parse_sgx_extensions(
     return result;
 }
 
+static oe_result_t _hex_crl_to_byte(
```
qiucwang, line 197: That's a naive way for conversion for now. Cannot find a OE function to do that.
@mingweishih Should we make it a standard method and put it elsewhere?

mingweishih: If this is currently used here, I think it's fine.

common/sgx/collateral.c
```diff
@@ -194,6 +194,40 @@ static oe_result_t _parse_sgx_extensions(
...
+        int value = 0;
+        if (c >= '0' && c <= '9')
+            value = c - '0';
+        else if (c >= 'A' && c <= 'F')
+            value = 10 + c - 'A';
+        else if (c >= 'a' && c <= 'f')
+            value = 10 + c - 'a';
+        else
+            OE_RAISE_MSG(OE_INVALID_PARAMETER, "Invalid hex character: %c", c);
+        crl_byte[i >> 1] += value << (((i + 1) & 1) << 2);
```
mingweishih, line 222: Can you explain a bit more on this line (e.g., adding comments)?

common/sgx/collateral.c
```diff
@@ -194,6 +194,40 @@ static oe_result_t _parse_sgx_extensions(
     return result;
 }
 
+static oe_result_t _hex_crl_to_byte(
+    uint8_t* crl_hex,
+    size_t crl_hex_size,
+    uint8_t** crl_byte_buffer,
+    size_t* crl_byte_buffer_size)
+{
+    oe_result_t result = OE_OK;
```
mingweishih, line 203: Given this function is scoped to be used here, I think missing checks against inputs are fine. If later we want to make this shared by others, then we should add those checks.

mingweishih (commented):
The changes look good to me. However, I do have a general question: how do we test this code path? Is it already covered or do we have test that can be enabled in our CI?

common/sgx/collateral.c
qiucwang: Sure.

qiucwang:
> The changes look good to me. However, I do have a general question: how do we test this code path? Is it already covered or do we have test that can be enabled in our CI?

It's covered by attestation plugin tests. The CRL needs to be loaded to check against the certificate during the attestation. However, this code path is use to check DER format CRL, which only comes from SGX PCS V3 API. As long as there is a pipeline using SGX default QPL, this code path will be tested.

mingweishih:
> > The changes look good to me. However, I do have a general question: how do we test this code path? Is it already covered or do we have test that can be enabled in our CI?
> 
> CRL verification is covered by attestation plugin tests. The CRL needs to be loaded to check against the certificate during the attestation. However, this code path is use to check DER format CRL, which only comes from SGX PCS V3 API. As long as there is a pipeline using SGX default QPL, this code path will be tested.

But why the tests didn't fail on CI before?

qiucwang:
> > > The changes look good to me. However, I do have a general question: how do we test this code path? Is it already covered or do we have test that can be enabled in our CI?
> > 
> > 
> > CRL verification is covered by attestation plugin tests. The CRL needs to be loaded to check against the certificate during the attestation. However, this code path is use to check DER format CRL, which only comes from SGX PCS V3 API. As long as there is a pipeline using SGX default QPL, this code path will be tested.
> 
> But why the tests didn't fail on CI before?

Previous we were getting v1/v2 collateral with PEM CRL, which was handled properly with `oe_read_crl_pem()`. I think we only start to test v3 collateral recently in Cirrus test. But we get v3 CRL from azure dcap, who cached it in PEM format, so it's also not found. Azure dcap is looking into possibility to store DER CRL. Once they switch to DER, it will be tested.

mingweishih:
> > > > The changes look good to me. However, I do have a general question: how do we test this code path? Is it already covered or do we have test that can be enabled in our CI?
> > > 
> > > 
> > > CRL verification is covered by attestation plugin tests. The CRL needs to be loaded to check against the certificate during the attestation. However, this code path is use to check DER format CRL, which only comes from SGX PCS V3 API. As long as there is a pipeline using SGX default QPL, this code path will be tested.
> > 
> > 
> > But why the tests didn't fail on CI before?
> 
> Previous we were getting v1/v2 collateral with PEM CRL, which was handled properly with `oe_read_crl_pem()`. I think we only start to test v3 collateral recently in Cirrus test. But we get v3 CRL from azure dcap, who cached it in PEM format, so the bug was also not detected. Azure dcap is looking into possibility to store DER CRL. Once they switch to DER, it will be tested.

Is it possible to at least have a simple test against this code path now rather than leaving the code untested for a while? For example, caching a der crl in the `report` test.

qiucwang:
> Is it possible to at least have a simple test against this code path now rather than leaving the code untested for a while? For example, caching a der crl in the `report` test.
Test a sample der crl alone is tricky as `oe_validate_revocation_list()` requires an entire endorsement buffer instead of a crl buffer. We also need a sample PCK certificate to check against with. That's why we generate a real report/evidence and an endorsement each time we test verification function.
I think the best way to test it is to add a ctest pipeline using default SGX QPL as quote provider, which returns a DER crl hex string. That's exactly where #3885 failed.
@yentsanglee any suggestion?

qiucwang:
Hi @mingweishih , hex string helper change has been reverted as Intel SGX might change the DER CRL to binary format in the near future. So that hex string to byte conversion might not needed.

mingweishih (approved):
LGTM. My understanding is that this PR won't address #3885, which will be fixed in the future update from Intel, correct? If so, can you comment on the issue?

qiucwang:
> LGTM. My understanding is that this PR won't address #3885, which will be fixed in the future update from Intel, correct? If so, can you comment on the issue?

Sure.

mingweishih:
> > LGTM. My understanding is that this PR won't address #3885, which will be fixed in the future update from Intel, correct? If so, can you comment on the issue?
> 
> Sure.

Thanks! Also, please update the comments of this PR.

yentsanglee (approved):

qiucwang:
> > > LGTM. My understanding is that this PR won't address #3885, which will be fixed in the future update from Intel, correct? If so, can you comment on the issue?
> > 
> > 
> > Sure.
> 
> Thanks! Also, please update the comments of this PR.

Missed that. Updated.

PR #3907, opened 2021-03-26
upgrade musl to v1.1.23

RRathna:

### DIFF
--- a/3rdparty/musl/README.md
+++ b/3rdparty/musl/README.md
@@ -4,7 +4,7 @@ musl libc
 This directory contains the musl libc library and the accompanying libc-test test
 suite. The structure of the directory is as follows.
 
-- musl/ (tag: `v1.1.21`)
+- musl/ (tag: `v1.1.23`)
 
   The clone of musl libc that is included as a git submodule, which points to
   a recent release tag. To update the submodule, we use the following procedure:
--- a/3rdparty/musl/CMakeLists.txt
+++ b/3rdparty/musl/CMakeLists.txt
@@ -68,6 +68,8 @@ ExternalProject_Add(
           ${MUSL_INCLUDES}/bits
   COMMAND ${CMAKE_COMMAND} -E copy_directory ${MUSL_DIR}/arch/${ARCH}/bits
           ${MUSL_INCLUDES}/bits
+  COMMAND ${CMAKE_COMMAND} -E copy ${MUSL_DIR}/arch/generic/fp_arch.h
+          ${MUSL_INCLUDES}/fp_arch.h
   # bash -c requires the command string to be in a single line
   COMMAND
     ${OE_BASH} -c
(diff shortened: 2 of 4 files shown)


RRathna:
bors try

anakrish:
https://musl.libc.org/ says:
> SECURITY ADVISORY: All releases through 1.2.1 are affected by CVE-2020-28928 and should be patched or upgraded to 1.2.2.

Additionally the [release history](https://musl.libc.org/releases.html) says that 1.1 series is at EOL.


Therefore, instead of upgrading to 1.1.23, we need to upgrade MUSL to 1.2.2

CyanDevs:
bors single on

CyanDevs:
bors try

3rdparty/musl/README.md
```diff
@@ -4,7 +4,7 @@ musl libc
 This directory contains the musl libc library and the accompanying libc-test test
 suite. The structure of the directory is as follows.
 
-- musl/ (tag: `v1.1.21`)
+- musl/ (tag: `v1.1.23`)
```
CodeMonkeyLeet, line 7: Shouldn't we be upgrading to v1.2.2 given that v1.1.x is [end-of-life](https://musl.libc.org/releases.html)?

radhikaj: @CodeMonkeyLeet @anakrish do we expect breaking changes for v1.2.2? If you expect major differences over current version, would prefer to keep to this plan since we will release next week and have a short window to  test. We can move to 1.2.2 post release

CodeMonkeyLeet: Given that one of the reasons for the MUSL upgrade is to address known vulnerabilities before the release, I would suggest at least trying the upgrade to 1.2.2 directly. Based on the release history, the primary breaking change was the migration of 32-bit architectures to using 64-bit `time_t` which should not affect OE since we only build on 64-bit.

RRathna: Thank you @CodeMonkeyLeet, @anakrish . I am working on upgrading to v1.2.2. There are quite a few additions and file structure changes that I am addressing. I was thinking of adding that upgrade it as a separate PR, so that this version can get in for the release (at the least) and v1.2.2 in next couple days. Is that okay?

anakrish: I took a quick look at upgrading to v1.2.2. It breaks our assumptions around pthread, some match functions and many more incompatibilities.
I don't think we can upgrade to v1.2.2 in a week. Therefore I'm OK with this PR to upgrade to v1.1.23 and am removing my blocker.

However, I'd suggest that in future we do upgrades early in the release cycle so that we get enough time to upgrade to the correct version as well as do necessary testing.

RRathna:
bors try

RRathna:
There are known math errors in v1.1.23, that are fixed in later versions. MUSL will be upgraded to v1.2.2 directly.

PR #3914, opened 2021-03-30
Provide crypto choice from VS extension

### LINKED ISSUE #3887: VSIX Extention: Provide Crypto Choice
Since v0.13 OE allows choosing between mbedtls and OpenSSL. OpenSSL is important since FIPS module is going to be enabled via OpenSSL interface. 

In VSIX extension, the mbedtls is hard-coded in the list of libraries. https://github.com/openenclave/openenclave/blob/82bff7df3d98b0ce89f3ae0f2eaa21e2fd5e9cfc/devex/vsextension/msbuild/open-enclave-cross.targets#L347
   - In [Release v0.13](https://github.com/openenclave/openenclave/releases/tag/v0.13.0) OpenSSL support was added, and the notion of a default crypto implementation was introduced. The VS extension hard-codes mbedtls, and it is not clear how a user could use openssl instead of mbedtls for OE's internal cryptographic operations.
   
  OE no longer has a default crypto library and developer must explicitly choose the crypto library.
  
  This issue captures updating the VSIX extension to provide this choice to the user.
  
  Based on discussion around #3886, it is possible to provide this choice to the user.

dthaler:
Fixes #3887

Please check if the set of crypto libs is correct.  (This PR is scoped to only the crypto libs, but in general the intent is to move the details to open-enclave-cross.targets, which is updated with the nuget package, and out of OpenEnclave.Cpp.Clang.targets which is only updated with the VS extension.)

### DIFF
--- a/devex/vsextension/ProjectWizard/VCTargets/1033/openenclave-general.xml
+++ b/devex/vsextension/ProjectWizard/VCTargets/1033/openenclave-general.xml
@@ -35,6 +35,16 @@
                   Category="General">
   </StringProperty>
 
+  <EnumProperty Name="CryptoLib"
+                DisplayName="Crypto Library"
+                Description="Crypto Library to link with."
+                Category="General"
+                IsRequired="true"
+                Default="mbedtls">
+    <EnumValue Name="mbedtls" DisplayName="Mbed TLS" />
+    <EnumValue Name="openssl" DisplayName="OpenSSL" />
+  </EnumProperty>
+
   <BoolProperty Name="UseLldLink"
                 DisplayName="Use lld-link"
                 Description="Use lld-link for linking.  If this option is disabled, the Microsoft linker (link.exe) will be used instead."
--- a/devex/vsextension/ProjectWizard/VCTargets/OpenEnclave.Cpp.Clang.targets
+++ b/devex/vsextension/ProjectWizard/VCTargets/OpenEnclave.Cpp.Clang.targets
@@ -273,7 +273,7 @@ that use Clang Front-End.
       <OutDirTrimmed>$(OutDir.TrimEnd({'\\'))</OutDirTrimmed>
     </PropertyGroup>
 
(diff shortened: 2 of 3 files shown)


devex/vsextension/ProjectWizard/VCTargets/1033/openenclave-general.xml
```diff
@@ -35,6 +35,16 @@
                   Category="General">
   </StringProperty>
 
+  <EnumProperty Name="CryptoLib"
+                DisplayName="Crypto Library"
+                Description="Crypto Library to link with."
+                Category="General"
+                IsRequired="true"
+                Default="mbedtls">
+    <EnumValue Name="mbedtls" DisplayName="mbedTLS" />
```
CodeMonkeyLeet, line 44: FYI, since the donation to TrustedFirmware org, they've been more consistent in using the brand name as "Mbed TLS" in their docs:

```suggestion
    <EnumValue Name="mbedtls" DisplayName="Mbed TLS" />
```

CodeMonkeyLeet (approved):
LGTM, thanks for fixing this!

radhikaj (approved):

rs-- (approved):

devex/vsextension/ProjectWizard/VCTargets/1033/openenclave-general.xml
dthaler: Done

dthaler:
bors try

mingweishih (approved):

anakrish:
mbedtls is still hardcoded in open-enclave-cross.targets.

https://github.com/openenclave/openenclave/blob/6c665c038453cadda1229c1318084095fbd45e8f/devex/vsextension/msbuild/open-enclave-cross.targets#L361

This hard-coding is referred to in the problem statement for issue  #3887. If the above line is not used anywhere, then it ought to be removed.

dthaler:
@anakrish please re-review

anakrish (approved):

anakrish (approved):

PR #3905, opened 2021-03-25
Windows 2016 CI deprecation

BRMcLaren:
This PR deprecates Windows 2016 server support from open enclave SDK
   - Doc changes
   - Removal of CI testing logic
   - Update any non-sgx windows vm's to use 2019 (host verification tests) within ci
   - Removal of Windows server 2016 shared image gallery builders
   
Note: this PR does not deprecate the windows install-windows-prereqs.ps1 script, which will happen in a subsequent PR due to avoiding convolution.

PR #3916, opened 2021-03-30
Windows 2016 Developer set up deprecation

BRMcLaren:
This PR deprecates the 2016 installation setup components as the second half of #3905 . As windows, 2019 ~= for Windows 10 in regards to the installation of components with respect to Intel SGX, also tacked on some consolidation to simplify the script.

### DIFF
--- a/scripts/install-windows-prereqs.ps1
+++ b/scripts/install-windows-prereqs.ps1
@@ -174,26 +174,6 @@ function Start-LocalPackagesDownload {
     Write-Output "Finished downloading all the packages"
 }
 
-function Get-WindowsRelease {
-    $releases = @{
-        18363 = "Win10"
-        18362 = "Win10"
-        17763 = "WinServer2019"
-        14393 = "WinServer2016"
-    }
-    $osBuild = [System.Environment]::OSVersion.Version.Build
-    $osInfo = Get-CimInstance -ClassName Win32_OperatingSystem
-    $releaseName = $releases[$osBuild]
-    # ProductType: 1 - Work Station, 3 - Server
-    if (($osBuild -eq 17763) -and ($osInfo.ProductType -eq 1)) {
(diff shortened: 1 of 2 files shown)


scripts/install-windows-prereqs.ps1
```diff
@@ -648,7 +571,7 @@ function Install-DCAP-Dependencies {
         Throw "Failed to install nuget EnclaveCommonAPI"
     }
 
-    if (($LaunchConfiguration -eq "SGX1FLC") -or (${OS_VERSION} -eq "WinServer2019"))
+    if ($LaunchConfiguration -eq "SGX1FLC")
```
rs--, line 574: Since we ran this code path automatically when running on 2019, should this check be removed?

rs-- (approved):

BRMcLaren:
@rs--  Rebased and fixed the recommendation, added as a separate commit. Thanks for pointing it out.

scripts/install-windows-prereqs.ps1
BRMcLaren: You could think of this check as to see if the user is setting up on ACC hardware, as opposed to a non-ACC vm as line 575 will not run on a non-SGX VM.

We never encountered this edge case before due to us always using host verification 2016 boxes (in which is actually a testing hole on our part). So the above would need to run when running a SGX1FLC setup but not a SGX1FLC-NoIntelDrivers.

I can add a comment regarding this as it's tribal info? but I am of the opinion the above implementation is correct.

rs--: Works for me

CyanDevs (approved):
LGTM!

BRMcLaren:
Someone is decoupling their infra from this, they asked me to wait a day before merging so adding do not merge label for now.

PR #3924, opened 2021-04-01
Fix return value of SYS_getcwd

thomasten:
The syscall should return the number of bytes written.


mingweishih (approved):
LGTM. Thanks for the fix!

anakrish (approved):
Thanks for the fix!

PR #3928, opened 2021-04-01
Add getrandom support based on oe_random_internal

mingweishih:
This PR adds the support the `getrandom` (both libc API and the syscall) and the test. Although the syscall is Linux-specific, OE supports the API (and the syscall) based on the TEE-specific rand API (`oe_random_internal`) to get the random bytes. Therefore, the support of the API/syscall inside the enclave is platform-agnostic.

### DIFF
--- a/syscall/syscall.c
+++ b/syscall/syscall.c
@@ -8,6 +8,7 @@
 #include <openenclave/corelibc/stdlib.h>
 #include <openenclave/corelibc/string.h>
 #include <openenclave/internal/print.h>
+#include <openenclave/internal/random.h>
 #include <openenclave/internal/safemath.h>
 #include <openenclave/internal/syscall/device.h>
 #include <openenclave/internal/syscall/dirent.h>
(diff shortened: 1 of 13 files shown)


mingweishih:
bors try

anakrish (approved):
Very useful addition!

RRathna (approved):

syscall/syscall.c
```diff
@@ -1067,6 +1068,34 @@ OE_WEAK OE_DEFINE_SYSCALL2(SYS_umount2)
...
 
+OE_WEAK OE_DEFINE_SYSCALL3_M(SYS_getrandom)
+{
+    oe_errno = 0;
+    long ret = -1;
+    void* buf = (void*)arg1;
+    size_t buflen = (size_t)arg2;
+    unsigned int flags = (unsigned int)arg3;
+
+    (void)flags;
```
dthaler, line 1079: Would it be better to fail if flags != 0 instead of silently ignoring them?

tests/syscall/getrandom/enc/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+# Licensed under the MIT License.
+
+set(EDL_FILE ../getrandom.edl)
+
+add_custom_command(
+  OUTPUT getrandom_t.h getrandom_t.c
+  DEPENDS ${EDL_FILE} edger8r
+  COMMAND
+    edger8r --trusted ${EDL_FILE} --search-path ${PROJECT_SOURCE_DIR}/include
+    ${DEFINE_OE_SGX} --search-path ${CMAKE_CURRENT_SOURCE_DIR})
```
dthaler, line 11: Shouldn't this line be inside some sort of SGX-only check?

tests/syscall/getrandom/host/host.c
```diff
@@ -0,0 +1,38 @@
...
+    if (argc != 2)
+    {
+        fprintf(stderr, "Usage: %s ENCLAVE_PATH\n", argv[0]);
+        return 1;
+    }
+
+    const uint32_t flags = oe_get_create_flags();
+
+    result = oe_create_getrandom_enclave(
+        argv[1], OE_ENCLAVE_TYPE_SGX, flags, NULL, 0, &enclave);
```
dthaler, line 24:
```suggestion
        argv[1], OE_ENCLAVE_TYPE_AUTO, flags, NULL, 0, &enclave);
```

I don't see anything SGX-specific in this test

tests/syscall/getrandom/enc/CMakeLists.txt
mingweishih: `DEFINE_OE_SGX` is only set for SGX build: https://github.com/openenclave/openenclave/blob/55e5a29e858e8c6e46940c7d833db6c08046ef80/tests/CMakeLists.txt#L28

syscall/syscall.c
mingweishih: Just realize that the normal use of libc API may to set the flags. Error out would prevent such programs.

dthaler: actually I would claim that not implementing the behavior of the flags may prevent such programs.  Unless you have tested them, we should not claim to support the flags being set.

mingweishih: Fix by #3934

PR #3934, opened 2021-04-02
Explicitly disable the use of getrandom flags in OE

mingweishih:
The flags of getrandom are not implemented/supported in OE. Explicitly return failure when the flags are set. 


dthaler (approved):
Looks good.  (FYI, it's also fine if we support the flags, but then we better match the documented semantics.)

anakrish (approved):

PR #3940, opened 2021-04-06
Fix logging with module loading support

mingweishih:
Fix a bug in the module loading logging and add a log if a weak symbol is not resolved.

### DIFF
--- a/host/sgx/loadelf.c
+++ b/host/sgx/loadelf.c
@@ -797,10 +797,12 @@ static oe_result_t _link_elf_image(
             else
             {
                 if ((symbol->st_info >> 4) != STB_WEAK)
-                    OE_RAISE(
+                    OE_RAISE_MSG(
                         OE_UNSUPPORTED_ENCLAVE_IMAGE,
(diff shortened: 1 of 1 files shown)


host/sgx/loadelf.c
```diff
@@ -797,10 +797,12 @@ static oe_result_t _link_elf_image(
             else
             {
                 if ((symbol->st_info >> 4) != STB_WEAK)
-                    OE_RAISE(
+                    OE_RAISE_MSG(
```
anakrish, line 800: What is the difference between `OE_RAISE` and `OE_RAISE_MSG`?

mingweishih: `OE_RAISE` does not print message, just raises errors.

anakrish: Ah ok.

anakrish (approved):

manojrupireddy (approved):

BRMcLaren (approved):

PR #3946, opened 2021-04-08
Allow Simulation mode on Alpine Linux

anakrish:
Fix compile errors. With the PR and an updated oeedger8r submodule,
OE SDK can be compiled using clang-10 on Alpine Linux.
All tests ought to pass in Simulation Mode.
Ability to compile OE SDK in Apline Linux is important for investigating
dotnet debugger integration for Mystikos.

Description of Changes:
- OpenSSL
  Use basename + cut since basename -s is not available in busybox (Alpine)
- Prevent compiler warnings in clang-10
-host/linux/syscall.c
 Use correct include paths
- hostthread
  Use recursive mutex in Alpine. Use _NP (non portable) only if
  available
- ptracelib
  Signature of ptrace is different but equivalent on Ubuntu and Alpine.
  Ubuntu expects enum __ptrace_request as first parameter whereas alpine
  expects just an int.
  Tweak the library so that it compiles on both platforms. We just int
  as the first parameter for ptrace.
- aep
  OE_AEP_ADDRESS is moved out of .relro section. Defining it in
  .relro section causes a mutable relocation to be generated in
  a readonly section. This causes host applications to fail execution
  on Alpine which compiles executables as PIE by default.
  To fix this, OE_AEP_ADDRESS is defined in .data.rel.local section
  which the compiler uses by default for similar variable in C.
- tests/oeedger8r
  Prevent error from unused variable warning on host side.
- tests/crypto
  Prevent duplicate symbols link error
- host/CMakeLists.txt
  Prevent error due to warnings on Alpine. msghdr structure has
  different data types on Alpine.
- evidence.cpp
  Prevent error due to unused variable warning.

Signed-off-by: Anand Krishnamorthi <anakrish@microsoft.com>

### DIFF
--- a/3rdparty/openssl/append-unsupported
+++ b/3rdparty/openssl/append-unsupported
@@ -35,7 +35,9 @@ do
     if [[ "${i}" != *"crypto.h" ]] && [[ "${i}" != *"ssl.h" ]] && [[ "${i}" != *"x509_vfy.h" ]]; then
         continue
     fi
-    name="$(basename -s .h $i)"
+    # Use basename + cut which is portable.
+    # Alpine/busybox has very basic implementation of basename.
+    name=$(echo "$(basename $i)" | cut -f 1 -d '.')
     # We assume that every header ends with #endif, which is the case for OpenSSL.
     sed -i '$d' ${i}
     echo -e "#include <openssl/${name}_unsupported.h>\n#endif" >> ${i}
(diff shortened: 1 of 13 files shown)


anakrish:
bors try

mingweishih (approved):
LGTM

anakrish:
bors try

anakrish:
bors try

manojrupireddy (approved):

nonpolarity (approved):

Issue #2898, opened 2020-04-20
Publish oecert tool


jhand2:
The oecert tool is an internal command-line tool which uses the OE attestation code to output a report and certificate from SGX.

This tool may be valuable for people writing SGX verification code. We should consider publishing this tool with the SDK so developers can use it to generate test data. This would also allow users to generate files to pass to the new host_verify sample: https://github.com/openenclave/openenclave/pull/2865

sub task:
- [x] Option to skip evidence verification - #3912 
- [x] Option to print verbose loggings (similar to oecertdump tool) - #3889 
- [x] Hardcoded enclave filename - #3919
- [x] Option to specify in-proc or out-of-proc quote generation - #3949
- [x] Option to not generate report/evidence/certificate output file - #3939 
- [x] Enclave must be signed. Check mrsigner when verification option is set - #3955
- [x] Add legal printout - #3955
- [x] Publish the tool - #3947
- [x] Rename tool and parameters - #4067

yentsanglee:
oecert tool has to be run in SGX enclave-enabled environment. Users may build host verify sample from non-SGX VMs. One benefit for publishing oecert is so that the users can get a valid report from a SGX VM and copy the report file to a non-SGX VM for host verification.

If we decide to publish ocert, it should NOT be included in host verify package.

jhand2:
Sure, I think just publishing it with the open-enclave package is fine.

radhikaj:
Open another issue to make this tool TEE agile via plugins.

yentsanglee:
Still more work to do.

yentsanglee:
oeutil in tools folder.

PR #3947, opened 2021-04-08
Publish oecert and oecert_enc.signed to /opt/openenclave/bin

ryanhsu19:
Publish `oecert` and `oecert_enc.signed` to `/opt/openenclave/bin`.

This is related to #2898, which also covers other `oecert`-related works.


mingweishih (commented):
If we want to publish the tool, we should move the oecert under tools/ instead of current tests/tools/.

Also, we should update the changelog.

qiucwang:
> If we want to publish the tool, we should move the oecert under tools/ instead of current tests/tools/.
> 
> Also, we should update the changelog.

Yes, we will eventually move `oecert` to tools/ after all [expected features](https://github.com/openenclave/openenclave/issues/2898) have been added to `oecert`

yentsanglee:
@mingweishih oecert is also a test tool as some test cases depend on it. How would you recommend the source location to be?

mingweishih:
> @mingweishih oecert is also a test tool as some test cases depend on it. How would you recommend the source location to be?

Similar to existing tools such as oeedger8r and oesign that are put under tools/. We can directly invoke the executable via `add_custom_command` in the tests (and set the dependency on oecert). For example:

https://github.com/openenclave/openenclave/blob/295445e25aa72bca7c966b2e8b5f099ffb1b1723/tests/tools/oesign/test-digest/CMakeLists.txt#L27

mingweishih:
> > If we want to publish the tool, we should move the oecert under tools/ instead of current tests/tools/.
> > Also, we should update the changelog.
> 
> Yes, we will eventually move `oecert` to tools/ after all [expected features](https://github.com/openenclave/openenclave/issues/2898) have been added to `oecert`

Another issue with publishing the enclave binary is that it requires signing.  Current `oecert` is signed by a random key generated at the build time. Shouldn't the published enclave signed by more formal one?

qiucwang:
> Another issue with publishing the enclave binary is that it requires signing. Current `oecert` is signed by a random key generated at the build time. Shouldn't the published enclave signed by more formal one?
We plan to copy what remote attestation did. We will generate a key pair at run time and write public key in a header file. And later use that private key to sign the enclave and public key header file to verify against enclave's `signer_id`.

yentsanglee:
@mingweishih Do we have a formal key that's used to sign published enclave? We can bring this up in triage meeting.

mingweishih:
> @mingweishih Do we have a formal key that's used to sign published enclave? We can bring this up in triage meeting.

Not sure if we have a openenclave key. If not, we should bring this up for discussion. Add @radhikaj @dthaler @BRMcLaren

radhikaj:
Based on the SIG-Arch meeting, it is fine to publish the tool and sign it with a random key. The tool must print out text that it is not suitable for production use and it is purely a debugging utility.
moving it to tools directory is the right thing to do.

ryanhsu19:
@qiucwang I rebased the original commit to master branch, so you can check if it includes all the recent changes to `oecert`. `oecert`, `oecert_enc` and `oecert_enc.signed` will be installed to `/opt/openenclave/bin` after doing `make install`.

qiucwang:
> @qiucwang I rebased the original commit to master branch, so you can check if it includes all the recent changes to `oecert`. `oecert`, `oecert_enc` and `oecert_enc.signed` will be installed to `/opt/openenclave/bin` after `make install`.

Thanks! Could you wait for #3955 checking in and rebase on it? It includes some cmake changes.

ryanhsu19:
bors try

qiucwang:
@ryanhsu19 #3979 fix the oecert cmake issue and you might need to rebase on the master again. Thank you!

ryanhsu19:
Thanks!

ryanhsu19:
bors try

ryanhsu19:
bors try

ryanhsu19:
bors try

yentsanglee (approved):

tests/tools/CMakeLists.txt
```diff
@@ -2,7 +2,6 @@
 # Licensed under the MIT License.
 
 if (OE_SGX)
-  add_subdirectory(oecert)
```
dthaler, line 5: oesign has both the tool itself, and tests for the tool (which is why the oesign directory is added in line 6).
Are there no tests for oecert?  Should a separate issue be filed for that or is testing of it already covered by some other test?  (It's not good to have things published to /opt/openenclave/bin that don't have tests.)

ryanhsu19: There's a unit test `host_verify` that uses `oecert` as part of the test. There isn't a separate test for `oecert`. That's why `tests/tools/oecert` is removed and `tools/oecert` seems more appropriate.

dthaler: where does `host_verify` use `oecert`?  I can only find `_verify_cert()` in `host_verify\host\host.cpp` but I can't find any invocation of `oecert`

ryanhsu19: Ah, I was wrong. It was part of the ACC Host Verification Test in Jenkins:
https://github.com/openenclave/openenclave/blob/8238dd84a3b10ded58fa3bb7ddd3ca17b81de048/.jenkins/pipelines/Azure/Linux/Jenkinsfile#L158

tests/tools/oecert/README.md
```diff
@@ -1,59 +1,4 @@
...
-
-Generate an EC private key, of size 256, and output it to a file named key.pem
-=====
-openssl ecparam -name prime256v1 -genkey -noout -out keyec.pem
-
-
-Extract the public key from the key pair, which can be used in a certificate
-=====
-openssl ec -in keyec.pem -pubout -out publicec.pem
+This tool has been moved from the current directory, `/tests/tools/oecert`, to `/tools/oecert` recently. Please refer to [the README file](/tools/oecert/README.md) for more information.
```
dthaler, line 4:
```suggestion
This tool has been moved from the current directory, `/tests/tools/oecert`, to `/tools/oecert`. Please refer to [the README file](/tools/oecert/README.md) for more information.
```

"recently" will quickly get stale since it would need an update that might be forgotten about.

My preference though would be to just delete this file and leave it to the changelog to explain.

dthaler (approved):

qiucwang (approved):
LGTM. Thanks!

PR #3882, opened 2021-03-12
Install packages and update documentation for different call paths for SGX quote generation

ryanhsu19:
The [original PR](https://github.com/openenclave/openenclave/pull/3806) did not get updated after pushing commits either from the same branch or from another branch, so I have decided to open a new one here with the same description as follows:

Out-of-process quote generation is done by default on Linux (unless both `libsgx-quote-ex` is uninstalled and `SGX_AESM_ADDR` is unset) while in-process quote generation is always done on Windows.

Based on our discussion, we decide not to add another environment variable since it is prone to errors.

This also fixes #3843.

Signed-off-by: Ryan Hsu ryhsu@microsoft.com

PR #3949, opened 2021-04-08
Allow oecert for out-of-proc quoting for SGX

qiucwang:
Add quote-proc param to `oecert` so that `oecert` can specify using in-proc or out-of-proc quoting.

### DIFF
--- a/tests/tools/oecert/host/host.cpp
+++ b/tests/tools/oecert/host/host.cpp
@@ -27,11 +27,18 @@
 #define INPUT_PARAM_OPTION_REPORT "--report"
 #define INPUT_PARAM_OPTION_EVIDENCE "--evidence"
 #define INPUT_PARAM_OPTION_ENDORSEMENTS_FILENAME "--endorsements"
+#define INPUT_PARAM_OPTION_QUOTE_PROC "--quote-proc"
+#define INPUT_PARAM_QUOTE_IN_PROC "in"
+#define INPUT_PARAM_QUOTE_OUT_OF_PROC "out"
 #define INPUT_PARAM_OPTION_VERIFY "--verify"
 #define INPUT_PARAM_OPTION_OUT_FILE "--out"
 #define DEFAULT_LOG_FILE "oecert.log"
 #define INPUT_PARAM_OPTION_LOG_FILE "--log"
 #define INPUT_PARAM_OPTION_VERBOSE "--verbose"
+#define SGX_AESM_ADDR "SGX_AESM_ADDR"
+#if defined(_WIN32)
+#define SGX_AESM_ADDR_MAXSIZE 32
+#endif
 
 // Structure to store input parameters
 //
@@ -42,6 +49,7 @@ typedef struct _input_params
     const char* out_filename;
     const char* log_filename;
     const char* endorsements_filename;
+    const char* quote_proc;
     bool generate_certificate;
     bool generate_report;
     bool generate_evidence;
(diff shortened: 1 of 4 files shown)


tests/tools/oecert/host/host.cpp
```diff
@@ -27,11 +27,13 @@
 #define INPUT_PARAM_OPTION_REPORT "--report"
 #define INPUT_PARAM_OPTION_EVIDENCE "--evidence"
 #define INPUT_PARAM_OPTION_ENDORSEMENTS_FILENAME "--endorsements"
+#define INPUT_PARAM_OPTION_OUT_OF_PROC "--outofproc"
```
mingweishih, line 30: Consider using --out-of-proc instead.

tests/tools/oecert/host/host.cpp
```diff
@@ -418,6 +430,26 @@ static oe_result_t _read_key(const char* filename, uint8_t** data, size_t* size)
 static oe_result_t _process_params(oe_enclave_t* enclave)
 {
     oe_result_t result = OE_FAILURE;
+    bool sgx_asem_env_set = false;
+
+// read "SGX_AESM_ADDR" environment variable
+// if not set, unset it back after oecert function
+#if defined(__linux__)
+    sgx_asem_env_set = (getenv(SGX_AESM_ADDR) != NULL);
```
mingweishih, line 438: A high-level question: what happens if the out of proc is not supported by the environment? Is there a way to check such cases?

tests/tools/oecert/host/host.cpp
qiucwang: 👍

tests/tools/oecert/host/host.cpp
qiucwang: `out-of-proc` depends on `sgx-quote-ex` library. If it's not installed, it will return sgx-quote-ex not found and quote generation will fail. Ryan's PR #3882 has added those dependencies in ansible script.

mingweishih: What happens if users simply update the openenclave package without using ansible?

mingweishih: Ah, I guess sgx-quote-ex should be one of the dependencies then. Is that correct?

qiucwang: Yes, that's discussed at SIG-Attestation meeting. We will document this behavior so that user will know sgx-quote-ex is the dependencies for out-of-proc quoting.

tests/tools/oecert/host/host.cpp
yentsanglee: We need the parameter to force quote gen to be in-proc or out-of-proc (override users' setting). Something like --proc <in|out>

qiucwang: Make sense.

tests/tools/oecert/host/host.cpp
yentsanglee: Miscommunicated. In the future given proper configuration, out-of-proc quote gen should work in Windows. Enforcing in-proc/out-of-proc should apply to both Linux and Windows. For now we only test out-of-proc quote gen in Linux.

tests/tools/oecert/host/host.cpp
```diff
@@ -418,6 +430,26 @@ static oe_result_t _read_key(const char* filename, uint8_t** data, size_t* size)
 static oe_result_t _process_params(oe_enclave_t* enclave)
 {
     oe_result_t result = OE_FAILURE;
+    bool sgx_asem_env_set = false;
```
yentsanglee, line 433: We should restore whatever is set/unset to SGX_AESM_ADDR. I.e.

char* sgx_aesm_env = getenv(SGX_AESM_ADDR);
...
setenv(SGX_AESM_ADDR, agx_aesm_env, 1);  // Restore value

qiucwang: Updated.

yentsanglee (approved):

yentsanglee:
bors try

qiucwang:
Jenkis host-verify out-of-proc quoting testing failed due to sgx-quote-ex lib not installed. Please merge this PR after #3882 merged.

qiucwang:
bors try

qiucwang:
bors try

mingweishih (approved):

ryanhsu19 (approved):
Thanks for the work!

PR #3951, opened 2021-04-09
give NUGET_PACKAGE_PATH default option

BRMcLaren:
Signed-off-by: brmclare <brmclare@microsoft.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,10 +223,14 @@ if (WIN32)
   endif ()
 
   if (NOT NUGET_PACKAGE_PATH)
-    message(
-      FATAL_ERROR
-        "NUGET_PACKAGE_PATH not defined. Please define NUGET_PACKAGE_PATH as the path to the installed Intel and DCAP Client nuget packages."
-    )
+    # try with a default option if user has not specified a NUGET_PACKAGE_PATH
+    set(NUGET_PACKAGE_PATH "C:\\oe_prereqs")
+    if (NOT EXISTS "${NUGET_PACKAGE_PATH}")
+      message(
+        FATAL_ERROR
+          "NUGET_PACKAGE_PATH not defined. Please define NUGET_PACKAGE_PATH as the path to the installed Intel and DCAP Client nuget packages."
+      )
+    endif ()
   endif ()
 else ()
   find_program(OE_BASH bash)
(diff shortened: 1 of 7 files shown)


BRMcLaren:
bors try

anakrish (commented):
Thanks for taking this on. It definitely improves usability.
Can you see if any associated documentation needs to be updated?

BRMcLaren:
> Thanks for taking this on. It definitely improves usability.
> Can you see if any associated documentation needs to be updated?

From a user perspective, it makes sense we tell users exactly how to set the nuget_path without them needing to look into the code, then default if not set. But would you say it makes sense to not include the documentation for how to set the path? Curious if I should be wording the documentation as "optional" where it has a default or leaving it as is due to the documentation already have the verbosity to getting a new user up and started without bogging them down with our defaults. I sided with the latter, but can definitely do the former.

anakrish:
> > Thanks for taking this on. It definitely improves usability.
> > Can you see if any associated documentation needs to be updated?
> 
> From a user perspective, it makes sense we tell users exactly how to set the nuget_path without them needing to look into the code, then default if not set. But would you say it makes sense to not include the documentation for how to set the path? Curious if I should be wording the documentation as "optional" where it has a default or leaving it as is due to the documentation already have the verbosity to getting a new user up and started without bogging them down with our defaults. I sided with the latter, but can definitely do the former.

There are many places in the doc where we specify `NUGET_PACKAGE_PATH`. 
https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/WindowsSGX1FLCGettingStarted.md#buildinginstallation-on-windows-using-developer-command-prompt

```bash
cd C:/Users/test/openenclave
mkdir build/x64-Release
cd build/x64-Release
cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DNUGET_PACKAGE_PATH=C:/oe_prereqs -DCMAKE_INSTALL_PREFIX=C:/openenclave ../..
ninja
```
If we have a default location, then the user does not need to specify `NUGET_PACKAGE_PATH` explicitly.
Normally, prereqs are set up by our scripts to be located at `C:\oe_prereqs`. Therefore the default would work.

The only documentation that I could spot where we explain to the user how to manually install prereqs is here:
https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/WindowsManualSGX1FLCDCAPPrereqs.md#azure-dcap-client-for-windows-optional
Even in this doc, we describe manual installation to `C:\oe_prereqs`. Maybe we could add a line here to say that if the user chooses to install to a location different from the default, then they'd need to specify that via NUGET_PACKAGE_PATH.

It seems like rest of the references to NUGET_PACKAGE_PATH can be removed.

BRMcLaren:
Sure, that makes sense. Done the suggested changes and added a comment about what the user should do if not using the default location.

radhikaj:
Closing for now. Can be reopened when someone is willing to take this up again.

PR #3971, opened 2021-04-22
remove redundant oe_get_sgx_quote_validity()

qiucwang:
In `oe_sgx_verify_evidence()` code path, `oe_get_sgx_quote_validity()` is called twice. 1st call is to verify sgx quote's validity, including validity period checking and 2nd call is to extract validity period for required claims.
2nd call is redundant as validity period can be fetched in the 1st call and `oe_get_sgx_quote_validity()` is relatively time consuming in enclave. Multiple memory allocations are required in this call.

After the fix, `oe_sgx_verify_evidence()` enclave running time reduced from 125 ms to 80ms in Linux.

### DIFF
--- a/common/sgx/quote.c
+++ b/common/sgx/quote.c
@@ -482,7 +482,12 @@ oe_result_t oe_verify_sgx_quote(
 
     // Endorsements verification
     OE_CHECK(oe_verify_quote_with_sgx_endorsements(
-        quote, quote_size, &sgx_endorsements, input_validation_time));
+        quote,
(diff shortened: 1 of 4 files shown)


yentsanglee (approved):

dthaler (approved):
Looks ok since this is not a public API

mingweishih (approved):

ryanhsu19 (approved):
LGTM. The reduced running time of an enclave looks great!

PR #3979, opened 2021-04-27
Avoid always triggering re-compile for the oecert binary

mingweishih:
Fix the incorrect dependency that always triggers the re-compilation of the `oecert` binary for each build. More detail of the bug as follows:

The custom command under oecert/host/CMakeLists.txt depends on the `oecert_enclave_signed` defined in the oecert/enc/CMakeLists.txt, which is not in the same scope. Therefore, the custom command treats the `oecert_enclave_signed` as undefined and always gets re-executed (and hence the `oecert` binary that depends on the custom command) for each build.

### DIFF
--- a/tests/tools/oecert/CMakeLists.txt
+++ b/tests/tools/oecert/CMakeLists.txt
@@ -2,9 +2,24 @@
 # Licensed under the MIT License.
 
 if (BUILD_ENCLAVES)
-  add_subdirectory(enc)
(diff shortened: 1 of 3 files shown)


mingweishih:
bors try

qiucwang:
Thanks for fixing that.

qiucwang (approved):
Thanks for fixing it.

radhikaj (approved):

PR #3984, opened 2021-04-28
Support the CapturePFGPExceptions preference

### LINKED ISSUE #3923: Enable SIGSEGV handling inside SGX Enclave utilizing SGX MISCSELECT #PF/#GP support
Intel's Icelake CPU (10nm Gen 10 Core and Xeon-SP) supports delivering #PG and #GP exception info securely inside the enclave. For LibOS solutions that try to support high level programing language runtime inside an SGX Enclave, #PF exception handling inside the Enclave is required. Some high level programing language runtime expects the OS to deliver SIGSEGV signal triggered by #PF on NULL pointer dereference. SIGSEGV/#PF on stack guard page access is also commonly expected by many programing language runtime.

The feature is enumerated in CPUID(EAX=12H, ECX=0).EBX.bit0. If the CPU supports this feature (CPUID(EAX=12H, ECX=0).EBX.bit0=1), with SW opt-in (SECS.MISCSELECT.bit0=1), the CPU delivers #PF and #GP exception info in SSA. 

If supported and opted-in, when #PF/#GP occurs inside the enclave, the MISC region (right before GPR region) of the SSA frame contains exception info of #PF/#GP, for example, linear address of the page access that caused the #PF and exception error code for #PF/#GP, and the EXITINFO within the GPR region reports exception vector 13(#GP) or 14(#PF).

To support trusted #PF/#GP exception handler inside a OE-SGX Enclave using this feature, the exception info in the MISC region needs to be passed to the registered exception handler, and the enclave loader needs to support opt-in (to enable the feature) if the CPU enumerates the feature.

The current OE SDK implementation does not support SW opt-in to enable this feature. The signing tool sets SIGSTRUCT.MISCSELECT=0, and SIGSTRUCT.MISCMASK=0xffffffff, which means the enclave load must set SECS.MISCSELECT=0. The enclave loader does not enumerate CPU's support on this feature either.

mingweishih:
Expose the CapturePFGPExceptions configuration as part of SGX2 features, allowing the enclave to capture #PF and #GP exceptions. More detailed about the PR
- Add the CapturePFGPExceptions preference to the enclave config file. `oesign` now is able to parse the option (expected to be a binary value)
- oesign can sign the enclave with CapturePFGPExceptions=1 on both SGX1 and SGX2 machines.
- When  CapturePFGPExceptions=1, the OE loader will enable the feature when running on an SGX2-capable CPU.
- When feature is enabled on SGX2 machines, the `OE_EXCEPTION_PAGE_FAULT` and `OE_EXCEPTION_ACCESS_VIOLATION` code can be used to check if the exception is #PF or #GP, respectively.
- `faulting_address` and `error_code` of the `oe_exception_record_t` struct will contain the information of the exceptions.
- `include/openenclave/bits/exception.h` includes the information based on Intel SDM (Table 37.14) used to interpret the `error_code`

Fixes #3923

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [Unreleased][Unreleased_log]
 --------------
 
+### Added
+- Add the CapturePFGPExceptions preference for the SGX2 feature of capturing #PF and #GP exceptions inside an enclave.
+  - Developers can specify the CapturePFGPExceptions with a binary value in the enclave config file or set the value via the newly added OE_SET_ENCLAVE_SGX2 macro, which is used to set SGX2-specific properties.
+  - When setting CapturePFGPExceptions=1, the OE loader will enable the feature when running on an SGX2-capable CPU.
+  - Once enabled, the in-enclave exception handler can capture the #PF (with the OE_EXCEPTION_PAGE_FAULT code) and #GP (with the code OE_EXCEPTION_ACCESS_VIOLATION code) exceptions.
+  - More information about the exceptions can be found in the `faulting_address` and `error_code` members of the `oe_exception_record_t` structure passed into the handler.
+
 [v0.16.1][v0.16.1_log]
 --------------
 ### Added
--- a/enclave/core/sgx/exception.c
+++ b/enclave/core/sgx/exception.c
@@ -249,6 +249,8 @@ void oe_real_exception_dispatcher(oe_context_t* oe_context)
     oe_exception_record.code = td->exception_code;
     oe_exception_record.flags = td->exception_flags;
     oe_exception_record.address = td->exception_address;
+    oe_exception_record.faulting_address = td->faulting_address;
+    oe_exception_record.error_code = td->error_code;
     oe_exception_record.context = oe_context;
 
     // Refer to oe_enter in host/sgx/enter.c. The contract we defined for EENTER
@@ -271,6 +273,13 @@ void oe_real_exception_dispatcher(oe_context_t* oe_context)
             break;
         }
     }
+
+    // Clear information after all the handlers are done
+    td->exception_code = 0;
+    td->exception_flags = 0;
+    td->exception_address = 0;
+    td->faulting_address = 0;
+    td->error_code = 0;
 
     // Jump to the point where oe_context refers to and continue.
     if (handler_ret == OE_EXCEPTION_CONTINUE_EXECUTION)
@@ -317,8 +326,9 @@ void oe_virtual_exception_dispatcher(
         return;
     }
 
-    sgx_ssa_gpr_t* ssa_gpr =
-        (sgx_ssa_gpr_t*)(((uint8_t*)ssa_info.base_address) + ssa_info.frame_byte_size - OE_SGX_GPR_BYTE_SIZE);
+    uint64_t gprsgx_offset = (uint64_t)ssa_info.base_address +
+                             ssa_info.frame_byte_size - OE_SGX_GPR_BYTE_SIZE;
+    sgx_ssa_gpr_t* ssa_gpr = (sgx_ssa_gpr_t*)gprsgx_offset;
     if (!ssa_gpr->exit_info.as_fields.valid)
     {
         // Not a valid/expected enclave exception;
@@ -364,6 +374,16 @@ void oe_virtual_exception_dispatcher(
     }
     else
     {
+        /* The following codes can only be captured with SGX2 and the
+         * MISCSELECT[0] bit is set to 1. */
+        if (td->exception_code == OE_EXCEPTION_PAGE_FAULT ||
+            td->exception_code == OE_EXCEPTION_ACCESS_VIOLATION)
+        {
+            sgx_exinfo_t* exinfo =
+                (sgx_exinfo_t*)(gprsgx_offset - OE_SGX_MISC_BYTE_SIZE);
+            td->faulting_address = exinfo->maddr;
+            td->error_code = exinfo->errcd;
+        }
         // Modify the ssa_gpr so that e_resume will go to second pass exception
         // handler.
         ssa_gpr->rip = (uint64_t)oe_exception_dispatcher;
--- a/host/sgx/create.c
+++ b/host/sgx/create.c
@@ -97,10 +97,25 @@ static bool _is_kss_supported()
     eax = ebx = ecx = edx = 0;
 
     // Obtain feature information using CPUID
-    oe_get_cpuid(0x12, 0x1, &eax, &ebx, &ecx, &edx);
+    oe_get_cpuid(CPUID_SGX_LEAF, 0x1, &eax, &ebx, &ecx, &edx);
 
     // Check if KSS (bit 7) is supported by the processor
-    if (!(eax & (1 << 7)))
+    if (!(eax & CPUID_SGX_KSS_MASK))
+        return false;
+    else
+        return true;
+}
+
+static bool _is_misc_region_supported()
+{
+    uint32_t eax, ebx, ecx, edx;
+    eax = ebx = ecx = edx = 0;
+
+    // Obtain feature information using CPUID
+    oe_get_cpuid(CPUID_SGX_LEAF, 0x0, &eax, &ebx, &ecx, &edx);
+
+    // Check if EXINFO is supported by the processor
+    if (!(ebx & CPUID_SGX_MISC_EXINFO_MASK))
         return false;
     else
         return true;
@@ -721,6 +736,7 @@ static oe_result_t _eeid_resign(
             properties->config.attributes,
             properties->config.product_id,
             properties->config.security_version,
+            &properties->config.flags,
             OE_DEBUG_SIGN_KEY,
             OE_DEBUG_SIGN_KEY_SIZE,
             properties->config.family_id,
@@ -882,6 +898,14 @@ oe_result_t oe_sgx_build_enclave(
         &props,
         &loaded_enclave_pages_size,
         &enclave_size));
+
+    /* Check if the enclave is configured with CapturePFGPExceptions=1 */
+    if (props.config.flags.capture_pf_gp_exceptions)
+    {
+        /* Only opt into the feature if CPU (SGX2) supports the MISC region. */
+        if (_is_misc_region_supported())
+            context->capture_pf_gp_exceptions_enabled = 1;
+    }
 
     if (props.config.attributes & OE_SGX_FLAGS_KSS)
     {
--- a/host/sgx/sgxsign.c
+++ b/host/sgx/sgxsign.c
@@ -571,6 +571,7 @@ static oe_result_t _init_sigstruct(
     uint64_t attributes,
     uint16_t product_id,
     uint16_t security_version,
+    const oe_sgx_enclave_flags_t* flags,
     const uint8_t* family_id,
     const uint8_t* extended_product_id,
     sgx_sigstruct_t* sigstruct)
@@ -620,7 +621,10 @@ static oe_result_t _init_sigstruct(
     sigstruct->miscselect = SGX_SIGSTRUCT_MISCSELECT;
 
     /* sgx_sigstruct_t.miscmask */
-    sigstruct->miscmask = SGX_SIGSTRUCT_MISCMASK;
+    if (flags && flags->capture_pf_gp_exceptions)
+        sigstruct->miscmask = SGX_SIGSTRUCT_MISCMASK_EXINFO;
+    else
+        sigstruct->miscmask = SGX_SIGSTRUCT_MISCMASK;
 
     /* sgx_sigstruct_t.attributes */
     sigstruct->attributes.flags = attributes;
@@ -681,6 +685,7 @@ oe_result_t oe_sgx_sign_enclave_from_engine(
     uint64_t attributes,
     uint16_t product_id,
     uint16_t security_version,
+    const oe_sgx_enclave_flags_t* flags,
     const char* engine_id,
     const char* engine_load_path,
     const char* key_id,
@@ -709,6 +714,7 @@ oe_result_t oe_sgx_sign_enclave_from_engine(
         attributes,
         product_id,
         security_version,
+        flags,
         family_id,
         extended_product_id,
         sigstruct));
@@ -728,6 +734,7 @@ oe_result_t oe_sgx_sign_enclave(
     uint64_t attributes,
     uint16_t product_id,
     uint16_t security_version,
+    const oe_sgx_enclave_flags_t* flags,
     const uint8_t* pem_data,
     size_t pem_size,
     const uint8_t* family_id,
@@ -755,6 +762,7 @@ oe_result_t oe_sgx_sign_enclave(
         attributes,
         product_id,
         security_version,
+        flags,
         family_id,
         extended_product_id,
         sigstruct));
@@ -774,6 +782,7 @@ oe_result_t oe_sgx_get_sigstruct_digest(
     uint64_t attributes,
     uint16_t product_id,
     uint16_t security_version,
+    const oe_sgx_enclave_flags_t* flags,
     const uint8_t* family_id,
     const uint8_t* extended_product_id,
     OE_SHA256* digest)
@@ -794,6 +803,7 @@ oe_result_t oe_sgx_get_sigstruct_digest(
         attributes,
         product_id,
         security_version,
+        flags,
         family_id,
         extended_product_id,
         &sigstruct));
@@ -810,6 +820,7 @@ oe_result_t oe_sgx_digest_sign_enclave(
     uint64_t attributes,
     uint16_t product_id,
     uint16_t security_version,
+    const oe_sgx_enclave_flags_t* flags,
     const uint8_t* cert_pem_data,
     size_t cert_pem_size,
     const uint8_t* digest_signature,
@@ -845,6 +856,7 @@ oe_result_t oe_sgx_digest_sign_enclave(
         attributes,
         product_id,
         security_version,
+        flags,
         family_id,
         extended_product_id,
         sigstruct));
--- a/include/openenclave/bits/exception.h
+++ b/include/openenclave/bits/exception.h
@@ -77,6 +77,56 @@ OE_EXTERNC_BEGIN
 #define OE_EXCEPTION_FLAGS_SOFTWARE 0x2
 
 /**
+ * The following flags are used to interpret the error_code of the
+ * oe_exception_record_t struct when a PF or GP exception occurs. Note that
+ * these exceptions are captured only with SGX2 CPU and the MISCSELCT[0] is set
+ * to 1.
+ */
+
+/**
+ * Page-protection violation flag
+ * 0 - The fault was caused by a non-present page.
+ * 1 - The fault was caused by a page-protection violation.
+ */
+#define OE_SGX_PAGE_FAULT_P_FLAG 0x1
+/**
+ * Read/Write flag
+ * 0 - The fault was caused by a read.
+ * 1 - The fault was caused by a write.
+ */
+#define OE_SGX_PAGE_FAULT_WR_FLAG 0x2
+/**
+ * U/S flag
+ * 0 - The fault was caused by a supervisor-mode access.
+ * 1 - The fault was caused by a user-mode access.
+ */
+#define OE_SGX_PAGE_FAULT_US_FLAG 0x4
+/**
+ * RSVD flag
+ * 0 - The fault was not caused by a reserved bit violation.
+ * 1 - The fault was caused by a reserved bit set to 1
+ */
+#define OE_SGX_PAGE_FAULT_RSVD 0x8
+/**
+ * I/D flag
+ * 0 - The fault was not caused by an instruction fetch.
+ * 1 - The fault was caused by an instruction fetch.
+ */
+#define OE_SGX_PAGE_FAULT_ID_FLAG 0x10
+/**
+ * Protection Key flag
+ * 0 - The fault was not caused by protection keys.
+ * 1 - The fault was caused by protection-key violation.
+ */
+#define OE_SGX_PAGE_FAULT_PK_FLAG 0x20
+/**
+ * SGX flag
+ * 0 - The fault was not related to SGX.
+ * 1 - The fault is SGX-specific (e.g., access violation).
+ */
+#define OE_SGX_PAGE_FAULT_SGX_FLAG 0x8000
+
+/**
  * Blob that contains X87 and SSE data.
  */
 typedef struct _oe_basic_xstate
@@ -153,6 +203,12 @@ typedef struct _oe_exception_record
     uint32_t flags; /**< Exception flags */
 
     uint64_t address; /**< Exception address */
+
+    /* Information for PF/GP exceptions, which are only available
+     * for SGX2 and requires application opt-in (set CapturePFGPExceptions=1).
+     */
+    uint64_t faulting_address;
+    uint32_t error_code;
 
     oe_context_t* context; /**< Exception context */
 } oe_exception_record_t;
--- a/include/openenclave/bits/sgx/sgxproperties.h
+++ b/include/openenclave/bits/sgx/sgxproperties.h
@@ -34,13 +34,18 @@ typedef struct _oe_sgx_enclave_image_info_t
 #define OE_SGX_FLAGS_KSS 0x0000000000000080ULL
 #define OE_SGX_SIGSTRUCT_SIZE 1808
 
+typedef struct _oe_sgx_enclave_flags_t
+{
+    uint32_t capture_pf_gp_exceptions : 1;
+    uint32_t reserved : 31;
+} oe_sgx_enclave_flags_t;
+
 typedef struct oe_sgx_enclave_config_t
 {
     uint16_t product_id;
     uint16_t security_version;
 
-    /* Padding to make packed and unpacked size the same */
-    uint32_t padding;
+    oe_sgx_enclave_flags_t flags;
 
     uint8_t family_id[16];
     uint8_t extended_product_id[16];
@@ -99,7 +104,10 @@ typedef struct _oe_sgx_enclave_properties
  * @param[in] ALLOW_DEBUG If true, allows the enclave to be created with
  * OE_ENCLAVE_FLAG_DEBUG and debugged at runtime
  * @param[in] REQUIRE_KSS If true, allows the enclave to be created with
- * kss properties
+ * KSS properties
+ * @param[in] CAPTURE_PF_GP_EXCEPTIONS If true, allows the enclave to capture
+ * #PF and #GP exceptions if the CPU supports the feature. The setting is
+ * ignored otherwise
  * @param[in] HEAP_PAGE_COUNT Number of heap pages to allocate in the enclave
  * @param[in] STACK_PAGE_COUNT Number of stack pages per thread to reserve in
  * the enclave
@@ -114,7 +122,8 @@ typedef struct _oe_sgx_enclave_properties
     EXTENDED_PRODUCT_ID,                                                  \
     FAMILY_ID,                                                            \
     ALLOW_DEBUG,                                                          \
-    REQUIRE_KSS,                                                            \
+    REQUIRE_KSS,                                                          \
+    CAPTURE_PF_GP_EXCEPTIONS,                                             \
     HEAP_PAGE_COUNT,                                                      \
     STACK_PAGE_COUNT,                                                     \
     TCS_COUNT)                                                            \
@@ -136,7 +145,8 @@ typedef struct _oe_sgx_enclave_properties
         {                                                                 \
             .product_id = PRODUCT_ID,                                     \
             .security_version = SECURITY_VERSION,                         \
-            .padding = 0,                                                 \
+            .flags.capture_pf_gp_exceptions = CAPTURE_PF_GP_EXCEPTIONS,   \
+            .flags.reserved = 0,                                          \
             .family_id = FAMILY_ID,                                       \
             .extended_product_id = EXTENDED_PRODUCT_ID,                   \
             .attributes = OE_MAKE_ATTRIBUTES(ALLOW_DEBUG, REQUIRE_KSS)    \
@@ -184,6 +194,7 @@ typedef struct _oe_sgx_enclave_properties
     {0},                                                                  \
     ALLOW_DEBUG,                                                          \
     false,                                                                \
+    0,                                                                    \
     HEAP_PAGE_COUNT,                                                      \
     STACK_PAGE_COUNT,                                                     \
     TCS_COUNT)
@@ -223,6 +234,57 @@ typedef struct _oe_sgx_enclave_properties
     FAMILY_ID,                                                            \
     ALLOW_DEBUG,                                                          \
     true,                                                                 \
+    0,                                                                    \
+    HEAP_PAGE_COUNT,                                                      \
+    STACK_PAGE_COUNT,                                                     \
+    TCS_COUNT)
+
+/**
+ * Defines the SGX2 properties for an enclave
+ *
+ * The enclave properties should only be defined once for all code compiled into
+ * an enclave binary. These properties can be overwritten at sign time by
+ * the oesign tool.
+ *
+ * @param[in] PRODUCT_ID ISV assigned Product ID (ISVPRODID) to use in the
+ * enclave signature
+ * @param[in] SECURITY_VERSION ISV assigned Security Version number (ISVSVN)
+ * to use in the enclave signature
+ * @param[in] EXTENDED_PRODUCT_ID ISV assigned Extended Product ID (ISVEXTPRODID)
+ *  to use in the enclave signature (SGX2 feature)
+ * @param[in] FAMILY_ID ISV assigned Product Family ID (ISVFAMILYID)
+ * to use in the enclave signature (SGX2 feature)
+ * @param[in] ALLOW_DEBUG If true, allows the enclave to be created with
+ * OE_ENCLAVE_FLAG_DEBUG and debugged at runtime
+ * @param[in] REQUIRE_KSS If true, allows the enclave to be created with
+ * KSS properties (SGX2 feature)
+ * @param[in] CAPTURE_PF_GP_EXCEPTIONS If true, allows the enclave to capture
+ * #PF and #GP exceptions if the CPU supports the feature. The setting is
+ * ignored otherwise (SGX2 feature)
+ * @param[in] HEAP_PAGE_COUNT Number of heap pages to allocate in the enclave
+ * @param[in] STACK_PAGE_COUNT Number of stack pages per thread to reserve in
+ * the enclave
+ * @param[in] TCS_COUNT Number of concurrent threads in an enclave to support
+ */
+ #define OE_SET_ENCLAVE_SGX2(                                             \
+    PRODUCT_ID,                                                           \
+    SECURITY_VERSION,                                                     \
+    EXTENDED_PRODUCT_ID,                                                  \
+    FAMILY_ID,                                                            \
+    ALLOW_DEBUG,                                                          \
+    CAPTURE_PF_GP_EXCEPTIONS,                                             \
+    REQUIRE_KSS,                                                          \
+    HEAP_PAGE_COUNT,                                                      \
+    STACK_PAGE_COUNT,                                                     \
+    TCS_COUNT)                                                            \
+ _OE_SET_ENCLAVE_SGX_IMPL(                                                \
+    PRODUCT_ID,                                                           \
+    SECURITY_VERSION,                                                     \
+    EXTENDED_PRODUCT_ID,                                                  \
+    FAMILY_ID,                                                            \
+    ALLOW_DEBUG,                                                          \
+    REQUIRE_KSS,                                                          \
+    CAPTURE_PF_GP_EXCEPTIONS,                                             \
     HEAP_PAGE_COUNT,                                                      \
     STACK_PAGE_COUNT,                                                     \
     TCS_COUNT)
--- a/include/openenclave/bits/sgx/sgxtypes.h
+++ b/include/openenclave/bits/sgx/sgxtypes.h
@@ -47,6 +47,8 @@ OE_EXTERNC_BEGIN
 #define SGX_SECINFO_TCS 0x0000000000000000100
 #define SGX_SECINFO_REG 0x0000000000000000200
 
+#define SGX_SECS_MISCSELECT_EXINFO 0x0000000000000001UL
+
 #define SGX_SE_EPID_SIG_RL_VERSION 0x200
 #define SGX_SE_EPID_SIG_RL_ID 0xE00
 
@@ -132,6 +134,7 @@ OE_CHECK_SIZE(sizeof(sgx_attributes_t), 16);
 
 /* sgx_sigstruct_t.miscmask */
 #define SGX_SIGSTRUCT_MISCMASK 0xffffffff
+#define SGX_SIGSTRUCT_MISCMASK_EXINFO 0xfffffffe
 
 /* sgx_sigstruct_t.flags */
 /* Mask all bits except bit 2 for MODE64BIT */
@@ -367,6 +370,17 @@ typedef struct sgx_ssa_gpr_t
     uint64_t fs_base;
     uint64_t gs_base;
 } sgx_ssa_gpr_t, *PSGX_SsaGpr;
+
+typedef struct _sgx_exinfo_t
+{
+    /*
+     * The naming and size of the members follows the Table 37-13 in the Intel
+     * Software Developer's Manual Volume 3.
+     */
+    uint64_t maddr;
+    uint32_t errcd;
+    uint8_t reserved[4];
+} sgx_exinfo_t;
 
 /*
 **==============================================================================
@@ -1052,7 +1066,7 @@ typedef struct _sgx_key
 
 /* Enclave MISCSELECT Flags Bit Masks, additional information to an SSA frame */
 /* If set, then the enclave page fault and general protection exception are
- * reported*/
+ * reported */
 #define SGX_MISC_FLAGS_PF_GP_EXIT_INFO 0x0000000000000001ULL
 
 /* Enclave Flags Bit Masks */
--- a/include/openenclave/internal/sgx/td.h
+++ b/include/openenclave/internal/sgx/td.h
@@ -81,7 +81,7 @@ oe_thread_data_t* oe_get_thread_data(void);
  * Due to the inability to use OE_OFFSETOF on a struct while defining its
  * members, this value is computed and hard-coded.
  */
-#define OE_THREAD_SPECIFIC_DATA_SIZE (3744)
+#define OE_THREAD_SPECIFIC_DATA_SIZE (3732)
 
 typedef struct _oe_callsite oe_callsite_t;
 
@@ -162,6 +162,14 @@ typedef struct _td
     /* TLS atexit functions (see enclave/core/sgx/threadlocal.c) */
     oe_tls_atexit_t* tls_atexit_functions;
     uint64_t num_tls_atexit_functions;
+
+    /* The following information is only available for SGX2 and MISCSELECT is
+     * set to 1 */
+    /* The faulting address associated with the PF exception. Should be zero
+     * for other exception types. */
+    uint64_t faulting_address;
+    /* The error code for PF and GP exceptions. */
+    uint32_t error_code;
 
     /* Reserved for thread specific data. */
     uint8_t thread_specific_data[OE_THREAD_SPECIFIC_DATA_SIZE];
--- a/tests/pf_gp_exceptions/CMakeLists.txt
+++ b/tests/pf_gp_exceptions/CMakeLists.txt
@@ -1,0 +1,14 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+add_subdirectory(host)
+
+if (BUILD_ENCLAVES)
+  add_subdirectory(enc)
+endif ()
+
+add_enclave_test(tests/pf_gp_exceptions pf_gp_exceptions_host
+                 sgx_pf_gp_exceptions_enc)
+
+add_enclave_test(tests/pf_gp_exceptions_unsigned pf_gp_exceptions_host
+                 sgx_pf_gp_exceptions_enc_unsigned)
--- a/tests/pf_gp_exceptions/enc/enc.c
+++ b/tests/pf_gp_exceptions/enc/enc.c
@@ -1,0 +1,79 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/enclave.h>
+#include <openenclave/internal/print.h>
+#include <openenclave/internal/tests.h>
+#include "pf_gp_exceptions_t.h"
+
+static uint64_t faulting_address;
+static uint32_t error_code;
+
+uint64_t test_pfgp_handler(oe_exception_record_t* exception_record)
+{
+    if (exception_record->code == OE_EXCEPTION_PAGE_FAULT)
+    {
+        const int MOV_INSTRUCTION_BYTES = 3;
+        faulting_address = exception_record->faulting_address;
+        error_code = exception_record->error_code;
+        exception_record->context->rip += MOV_INSTRUCTION_BYTES;
+        oe_host_printf(
+            "#PF: addr: 0x%lx, code: %u\n",
+            exception_record->faulting_address,
+            exception_record->error_code);
+    }
+    else if (exception_record->code == OE_EXCEPTION_ACCESS_VIOLATION)
+    {
+        const int LGDT_INSTRUCTION_BYTES = 8;
+        faulting_address = exception_record->faulting_address;
+        error_code = exception_record->error_code;
+        exception_record->context->rip += LGDT_INSTRUCTION_BYTES;
+        oe_host_printf(
+            "#GP: addr: 0x%lx, code: %u\n",
+            exception_record->faulting_address,
+            exception_record->error_code);
+    }
+    else
+    {
+        oe_host_printf("Unhandled code: %u\n", exception_record->code);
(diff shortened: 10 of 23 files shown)


mingweishih:
bors try

mingweishih:
@yentsanglee  @qiucwang  please check oesign-related changes.

mingweishih:
bors try

tools/oesign/oesign.c
```diff
@@ -311,6 +312,32 @@ static int _load_config_file(const char* path, config_file_options_t* options)
...
+            if (options->pf_gp_exceptions.has_value)
+            {
+                oe_err(
+                    "%s(%zu): Duplicate 'PFGPExceptions' value provided",
+                    path,
+                    line);
+                goto done;
+            }
+
+            // Debug must be 0 or 1
```
bodzhang, line 328: "Debug" or "PFGPExceptions"?

include/openenclave/bits/sgx/sgxtypes.h
```diff
@@ -1086,7 +1095,9 @@ typedef struct _sgx_key
 
 /* OE seal key default flag masks*/
 #define OE_SEALKEY_DEFAULT_FLAGSMASK (~SGX_FLAGS_NON_SECURITY_BITS)
-#define OE_SEALKEY_DEFAULT_MISCMASK (~SGX_MISC_NON_SECURITY_BITS)
+/* Include the bit 0, which represents the EXINFO option for SGX2 */
+#define OE_SEALKEY_DEFAULT_MISCMASK \
+    ((~SGX_MISC_NON_SECURITY_BITS) | SGX_MISC_FLAGS_PF_GP_EXIT_INFO)
```
bodzhang, line 1100: It's arguable whether enabling #PF/#GP info delivery in Enclave might have security implication. But if the feature did have security implication with regard to sealed data migration, the implementation should change `SGX_MISC_NON_SECURITY_BITS` definition,  instead of overriding the relationship between `OE_SEALKEY_DEFAULT_MISCMASK` and `SGX_MISC_NON_SECURITY_BITS`

More importantly, we need to consider the impact of changing `OE_SEALKEY_DEFAULT_MISCMASK`. If an old version of OE app built with legacy OE SDK sealed data, and an updated version of the App was built with new definition of OE_SEALKEY_DEFAULT_MISCMASK, the newer version of the App won't be able to unseal the sealed data of the previous version of the App (running on the same platform), as the flagmask and miscmask are all part of the SEAL key derivation `KEYDEPEDENCIES`., which affects the derived SEAL key,  according to Intel's Software Developer's Manual.

bodzhang (changes requested):

radhikaj:
@MWShan, we would like to request reviews from you and Juan.

radhikaj:
@jbdelcuv, @haitaohuang Please review.

tools/oesign/oesign.c
```diff
@@ -738,19 +778,24 @@ int oedigest(const char* enclave, const char* conffile, const char* digest_file)
     oe_sgx_enclave_properties_t properties;
     OE_SHA256 mrenclave = {0};
     OE_SHA256 digest = {0};
+    uint32_t miscselect = 0;
```
yentsanglee, line 781: Original code initialize miscselect to SGX_SIGSTRUCT_MISCSELECT which is 0. If we not using SGX_SIGSTRUCT_MISCSELECT, please remove the definition.

include/openenclave/bits/sgx/sgxtypes.h
```diff
@@ -47,6 +47,8 @@ OE_EXTERNC_BEGIN
 #define SGX_SECINFO_TCS 0x0000000000000000100
 #define SGX_SECINFO_REG 0x0000000000000000200
 
+#define SGX_SECS_MISCSELECT_EXINFO 0x0000000000000001UL
```
yentsanglee, line 50: This should be moved to near line 134 where `SGX_SIGSTRUCT_MISCMASK` is defined.

host/sgx/sgxsign.c
```diff
@@ -617,7 +618,7 @@ static oe_result_t _init_sigstruct(
      */
 
     /* sgx_sigstruct_t.miscselect */
-    sigstruct->miscselect = SGX_SIGSTRUCT_MISCSELECT;
+    sigstruct->miscselect = miscselect;
```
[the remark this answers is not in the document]
mingweishih: My understanding based on your feedback is that it's ok to set
- `SECS.MISCSELECT.bit[0]` = 0
- `SIGSTRUCT.MISCSELECT.bit[0]` = 1
- `SIGSTRUCT.MISCMASK.bit[0]` = 0

I tried to create an enclave with this configuration but failed on the EINIT.
According to the SDM, I found

> * If CPUID.(EAX=12H, ECX=0):EBX[31:0] = 0, MISCSELECT must be 0.
If CPUID.(EAX=12H, ECX=0):EBX[31:0] !=0, enclave writers must specify MISCSELECT such that each cleared
bit in MISCMASK must also specify the corresponding bit as 0 in MISCSELECT.

It seems that the MISCSELECT.bin[0] should always be the same in both SECS and SIGSTRUCT. In this case, I think it makes sense to make it fail on SGX1 machine.

include/openenclave/bits/sgx/sgxtypes.h
mingweishih: Good point. Should we defer the changes to the OE_SEALKEY_DEFAULT_MISCMASK and have a separate discussion? I think it's important to maintain the backward compatibility while providing ways such that developers can derive the key based on MISCSELECT value.

host/sgx/sgxsign.c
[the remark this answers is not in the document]
lzha101: Seems the solution in this PR is to provide a configurable option to make the MISCSELECT required. If the option is set to 1, MISCSELECT feature will be required for the enclave loading. 
Why not just change the loader to detect the feature and just enable it if the CPU supports this? Similar as the AVX support in OE SDK. In this case,  you need to change default MISCMASK.bit0 to 0 during signing. That is:
Change [sgxtypes.h#L134](https://github.com/openenclave/openenclave/blob/master/include/openenclave/bits/sgx/sgxtypes.h#L134)  from:
`#define SGX_SIGSTRUCT_MISCMASK 0xffffffff`
to 
`#define SGX_SIGSTRUCT_MISCMASK 0xfffffffe`

and then detect the MISC feature during enclave loading.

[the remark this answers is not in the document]
bodzhang: The second comment in SDM's Table 37-21 seems wrong. In [the combined volume set](https://software.intel.com/content/dam/develop/external/us/en/documents-tps/325462-sdm-vol-1-2abcd-3abcd.pdf) volume 3, chapter 40 SGX Instruction Reference, section 40.3 ECREATE

image

My understanding is

-  If the CPU does NOT support MISCSELECT.bit0 feature (#PF/#GP info delivery in Enclave), `SIGSTRUCT.MISCSELECT.bit[0]` must be 0, 
-  If the CPU supports MISCSELECT.bit0 feature , `SIGSTRUCT.MISCSELECT.bit[0]`  and `SIGSTRUCT.MISCMASK.bit[0]` can be 0 or 1, and ECREATE instruction enforces the check shown in the picture above

I believe @lzha101 's proposal above should work.

include/openenclave/bits/sgx/sgxtypes.h
bodzhang: @mingweishih , yes, MISCSELECT implication to sealed data migration policy should be discussed separately. I prefer to keep `SGX_MISC_NON_SECURITY_BITS` and `OE_SEALKEY_DEFAULT_MISCMASK` definitions unchanged, for now.

CHANGELOG.md
```diff
@@ -15,6 +15,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Add the support of the OE_ENCLAVE_FLAG_DEBUG_AUTO flag to the oe_create_enclave API. When the flag is set and the OE_ENCLAVE_FLAG_DEBUG flag is cleared, the debug mode is automatically turned on/off based on the value of Debug specified in the enclave config file.
+- Add the PFGPExceptions configuration for SGX enclaves, which allows opting into the SGX2 feature of capturing #PF/#GP exceptions inside an enclave.
+  - Developers can specify the PFGPExceptions with binary value in the enclave config file.
+  - Once opt-in the feature (value=1), the in-enclave exception handler can capture the #PF (with the OE_EXCEPTION_PAGE_FAULT code) and #GP (with the code OE_EXCEPTION_ACCESS_VIOLATION) exceptions.
+  - More information about the exceptions can be found in the `faulting_address` and `error_code` members of the `oe_exception_record_t` structure passed into the handler.
+  - The feature is only available on the SGX2 CPUs. When enabling in SGX1 CPUs, `oe_create_enclave` will return the `OE_UNSUPPORTED` error.
```
dthaler, line 22:
```suggestion
  - The feature is only available on the SGX2 CPUs. When enabling PFGPExceptions in SGX1 CPUs, `oe_create_enclave` will return the `OE_UNSUPPORTED` error.
```

host/sgx/create.c
```diff
@@ -106,6 +106,21 @@ static bool _is_kss_supported()
         return true;
 }
 
+static bool _is_MISC_region_supported()
```
dthaler, line 109:
```suggestion
static bool _is_misc_region_supported()
```
https://github.com/openenclave/openenclave/blob/master/docs/Community/DevelopmentGuide.md#naming-conventions says:
> Use `lower_snake_case` for variable, member/field, and function names.

host/sgx/create.c
```diff
@@ -106,6 +106,21 @@ static bool _is_kss_supported()
...
+{
+    uint32_t eax, ebx, ecx, edx;
+    eax = ebx = ecx = edx = 0;
+
+    // Obtain feature information using CPUID
+    oe_get_cpuid(0x12, 0x0, &eax, &ebx, &ecx, &edx);
+
+    // Check if EXINFO is supported by the processor
+    if (!(ebx & 1))
```
dthaler, line 118: Avoid hard-coding magic numbers.  Use #defines instead of 0x12 and 1

include/openenclave/bits/sgx/sgxtypes.h
```diff
@@ -368,6 +370,13 @@ typedef struct sgx_ssa_gpr_t
     uint64_t gs_base;
 } sgx_ssa_gpr_t, *PSGX_SsaGpr;
 
+typedef struct _sgx_exinfo_t
+{
+    uint64_t maddr;
+    uint32_t errcd;
```
dthaler, line 376: Normal OE style guidance is to avoid abbreviations like "addr" and "err" (but acronyms are ok).  However using them is ok if they exactly match the terms in some official spec.   If these names literally come from an Intel spec, please include a comment above the typedef with the URI of the spec and section number.

tests/pf_gp_exceptions/enc/enc.c
```diff
@@ -0,0 +1,69 @@
...
+    OE_TEST(faulting_address == 0);
+
+    return 0;
+}
+
+OE_SET_ENCLAVE_SGX_PF_GP_EXCEPTIONS(
+    1,    /* ProductID */
+    1,    /* SecurityVersion */
+    true, /* Debug */
+    true, /* PFGPExcpeiotns */
```
dthaler, line 66:
```suggestion
    true, /* PFGPExceptions */
```

tests/pf_gp_exceptions/pf_gp_exceptions.edl
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT License.
+
+enclave {
+    from "openenclave/edl/logging.edl" import oe_write_ocall;
+    from "openenclave/edl/fcntl.edl" import *;
+#ifdef OE_SGX
+    from "openenclave/edl/sgx/platform.edl" import *;
+#else
+    from "openenclave/edl/optee/platform.edl" import *;
+#endif
```
dthaler, line 11: Isn't this test inherently SGX-only?  Assuming so, there should not be an ifdef

host/sgx/sgxsign.c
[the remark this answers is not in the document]
mingweishih: I guess the question is how should an enclave developer expect to handle the difference between SGX1 and SGX2.
There are two options
1. Developers should explicitly opt-in the feature. The makes sense to an enclave that relies on the capability of handling #PF/#GP inside the enclave (implemented by current PR).
2. Developers should implement the exception handler with logic for both the SGX1 (the feature is disabled) and SGX2 (the feature is enabled by default) (proposed in this thread).

Note that the enclave would behave differently based on whether the feature is enabled or not. The option 2 will cause problems if 
- the developer wants the enclave to behave the same in both SGX1 and SGX2 as there is no way for them to disable the feature on SGX2
- the developer does not handle #PF and #GP in the exception handler (which would become mandatory on SGX2).

I think we can bring up the discussion in the SIG-Arch meeting next week.

Other than that, I still cannot create an enclave on the SGX1 machine with either
`SIGSTRUCT.MISCSELECT[0]`=1, `SIGSTRUCT.MISCMASK`=0xFFFFFFFE, `SIGSTRUCT.MISCSELECT`=0  (EINIT failed)
or
`SIGSTRUCT.MISCSELECT[0]`=0, `SIGSTRUCT.MISCMASK`=0xFFFFFFFE, `SIGSTRUCT.MISCSELECT`=1 (ECREATE failed)

Can Intel folks help verifying? @MWShan @haitaohuang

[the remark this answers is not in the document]
bodzhang: @mingweishih , did you mean `SIGSTRUCT.MISCSELECT[0]`=0, `SIGSTRUCT.MISCMASK`=0xFFFFFFFE, `SECS.MISCSELECT`=1  failed ECREATE on SGX1.0? It's expected, as `SECS.MISCSELECT`=1 is not allowed on a CPU that does not support MISCSELECT feature. 

`SIGSTRUCT.MISCSELECT[0]`=1 is also expected to fail on SGX1.0, regardless of `SIGSTRUCT.MISCMASK` and 'SECS.MISCSELECT' value.

The interesting tests are:

`SIGSTRUCT.MISCSELECT[0]`=0, `SIGSTRUCT.MISCMASK`=0xFFFFFFFE, `SECS.MISCSELECT`=1, on icelake SGX2.0 -expect to pass
`SIGSTRUCT.MISCSELECT[0]`=0, `SIGSTRUCT.MISCMASK`=0xFFFFFFFF, `SECS.MISCSELECT`=1, on icelake SGX2.0  - expect to fail
`SIGSTRUCT.MISCSELECT[0]`=0, `SIGSTRUCT.MISCMASK`=0xFFFFFFFE, `SECS.MISCSELECT`=0, on SGX1.0 -expect to pass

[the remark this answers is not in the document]
mingweishih: @bodzhang I updated the PR based on your suggestion
- Now the `SIGSTRUCT.MISCSELECT` is always 0, and the `SIGSTRUCT.MISCMASK` is set to 0xFFFFFFFE or 0xFFFFFFFF based on the value of `PFGPExceptions` in the config file.
- `SECS.MISCSELECT` is set based on `_is_misc_region_supported()` && `PFGPExceptions`, which will always be 0 on SGX1. Therefore, the enclave creation won't fail on SGX1.

@lzha101 If we have the loader set the `SIGSTRUCT.MISCMASK` automictically, we then cannot sign a SGX2 enclave on SGX1. Given that the set `SIGSTRUCT.MISCMASK`=0xFFFFFFFE is fine on SGX1 (both `SIGSTRUCT.MISCSELECT[0]` and `SECS.MISCSELECT[0]` cannot be set to 1 anyways), I decide to set the mask value based on the `PFGPExceptions` value in the conf file. Can you double check?

include/openenclave/bits/sgx/sgxtypes.h
mingweishih: Reverted the change

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -223,6 +229,45 @@ typedef struct _oe_sgx_enclave_properties
...
+ * @param[in] ALLOW_DEBUG If true, allows the enclave to be created with
+ * OE_ENCLAVE_FLAG_DEBUG and debugged at runtime
+ * @param[in] PF_GP_EXCEPTIONS If true, allows the enclave to capture #PF/#GP
+ * exceptions if CPU supports the feature. The setting is ignored otherwise.
+ * @param[in] HEAP_PAGE_COUNT Number of heap pages to allocate in the enclave
+ * @param[in] STACK_PAGE_COUNT Number of stack pages per thread to reserve in
+ * the enclave
+ * @param[in] TCS_COUNT Number of concurrent threads in an enclave to support
+ */
+ #define OE_SET_ENCLAVE_SGX_PF_GP_EXCEPTIONS(                             \
```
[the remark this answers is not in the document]
mingweishih: Will come back to this part after the design is finalized.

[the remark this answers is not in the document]
mingweishih: updated

tools/oesign/oesign.c
mingweishih: The implementation now does not modify the SIGSTRUCT.MISCSELECT data, only the SECS.MISCSELECT, so I reserve the definition for now.

include/openenclave/bits/sgx/sgxtypes.h
mingweishih: See my comments above.

mingweishih:
bors try

CHANGELOG.md
```diff
@@ -15,6 +15,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add the support of getrandom libc API and syscall in enclaves.
 - Add `libsgx-quote-ex`, `sgx-aesm-service` and several SGX AESM plugins to Ansible scripts so that users will be able to select in-process or out-of-process call path for quote generation. Refer to the [attestation sample](samples/attestation/README.md#determining-call-path-for-sgx-quote-generation) for more information.
 - Add the support of the OE_ENCLAVE_FLAG_DEBUG_AUTO flag to the oe_create_enclave API. When the flag is set and the OE_ENCLAVE_FLAG_DEBUG flag is cleared, the debug mode is automatically turned on/off based on the value of Debug specified in the enclave config file.
+- Add the PFGPExceptions configuration for SGX enclaves, which allows opting into the SGX2 feature of capturing #PF/#GP exceptions inside an enclave.
+  - Developers can specify the PFGPExceptions with binary value in the enclave config file or set the value via the newly added OE_SET_ENCLAVE_SGX2 macro, which is used to set SGX2-specific properties.
```
dthaler, line 19:
```suggestion
  - Developers can specify the PFGPExceptions with a binary value in the enclave config file or set the value via the newly added OE_SET_ENCLAVE_SGX2 macro, which is used to set SGX2-specific properties.
```

CHANGELOG.md
```diff
@@ -15,6 +15,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Once opt-in the feature (value=1), the in-enclave exception handler can capture the #PF (with the OE_EXCEPTION_PAGE_FAULT code) and #GP (with the code OE_EXCEPTION_ACCESS_VIOLATION) exceptions.
```
dthaler, line 20:
```suggestion
  - After opting into the feature (value=1), the in-enclave exception handler can capture the #PF (with the OE_EXCEPTION_PAGE_FAULT code) and #GP (with the code OE_EXCEPTION_ACCESS_VIOLATION) exceptions.
```

host/sgx/create.c
```diff
@@ -883,6 +899,14 @@ oe_result_t oe_sgx_build_enclave(
         &loaded_enclave_pages_size,
         &enclave_size));
 
+    /* Check if the enclave is configured with PFGPExceptions=1 */
+    if (props.config.flags.pf_gp_exceptions)
+    {
+        /* Only opt into the feature if CPU (SGX2) supports the MISC region. */
+        if (_is_misc_region_supported())
+            context->pf_gp_exceptions_opt_in = 1;
```
dthaler, line 907:
```suggestion
            context->pf_gp_exceptions_enabled = 1;
```

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -34,13 +34,18 @@ typedef struct _oe_sgx_enclave_image_info_t
 #define OE_SGX_FLAGS_KSS 0x0000000000000080ULL
 #define OE_SGX_SIGSTRUCT_SIZE 1808
 
+typedef struct _oe_sgx_enclave_flags_t
+{
+    uint32_t pf_gp_exceptions : 1;
```
dthaler, line 39: Based on the comments I see in the code, it seems like this should be called
```suggestion
    uint32_t capture_pf_gp_exceptions : 1;
```

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -114,7 +119,8 @@ typedef struct _oe_sgx_enclave_properties
     EXTENDED_PRODUCT_ID,                                                  \
     FAMILY_ID,                                                            \
     ALLOW_DEBUG,                                                          \
```
dthaler, line 121: It looks like this code doesn't follow [OE style guidance](https://github.com/openenclave/openenclave/blob/master/docs/Community/DevelopmentGuide.md) which says:

> Use lower_snake_case for variable, member/field, and function names.

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -136,7 +142,8 @@ typedef struct _oe_sgx_enclave_properties
         {                                                                 \
             .product_id = PRODUCT_ID,                                     \
             .security_version = SECURITY_VERSION,                         \
-            .padding = 0,                                                 \
+            .flags.pf_gp_exceptions = PF_GP_EXCEPTIONS,                   \
```
dthaler, line 145:
```suggestion
            .flags.capture_pf_gp_exceptions = CAPTURE_PF_GP_EXCEPTIONS,   \
```

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -223,6 +231,54 @@ typedef struct _oe_sgx_enclave_properties
...
+ * @param[in] PF_GP_EXCEPTIONS If true, allows the enclave to capture #PF/#GP
+ * exceptions if CPU supports the feature. The setting is ignored otherwise
+ * (SGX2 feature)
+ * @param[in] HEAP_PAGE_COUNT Number of heap pages to allocate in the enclave
+ * @param[in] STACK_PAGE_COUNT Number of stack pages per thread to reserve in
+ * the enclave
+ * @param[in] TCS_COUNT Number of concurrent threads in an enclave to support
+ */
+ #define OE_SET_ENCLAVE_SGX2(                                             \
+    PRODUCT_ID,                                                           \
```
dthaler, line 265: follow OE style guidance for new macros.  use lower_snake_case for args

tests/pf_gp_exceptions/enc/enc.c
```diff
@@ -0,0 +1,71 @@
...
+uint64_t faulting_address;
+uint32_t error_code;
+
+uint64_t test_pfgp_handler(oe_exception_record_t* exception_record)
+{
+    if (exception_record->code == OE_EXCEPTION_PAGE_FAULT)
+    {
+        faulting_address = exception_record->faulting_address;
+        error_code = exception_record->error_code;
+        exception_record->context->rip += 3;
```
dthaler, line 18: 3 looks like a magic number here.  Either add a #define for it, or add a comment explaining where the number comes from.

tests/pf_gp_exceptions/host/host.c
```diff
@@ -0,0 +1,66 @@
...
+    if (_is_misc_region_supported())
+    {
+        result = enc_pf_gp_exceptions(enclave, &return_value);
+        if (result != OE_OK)
+            oe_put_err("oe_call_enclave() failed: result=%u", result);
+
+        OE_TEST(return_value == 0);
+    }
+    else
+        printf("CPU does not support the PFGPExceptions=1 configuration. Skip "
```
dthaler, line 57: As above, from my reading, this should be called the CapturePFGPExceptions=1 configuration.
```suggestion
        printf("CPU does not support the CapturePFGPExceptions=1 configuration. Skip "
```

tools/oesign/oesign.c
```diff
@@ -49,6 +49,7 @@ typedef struct _config_file_options
     optional_uint16_t security_version;
     optional_oe_uuid_t family_id;
     optional_oe_uuid_t extended_product_id;
+    optional_bool_t pf_gp_exceptions;
```
dthaler, line 52:
```suggestion
    optional_bool_t capture_pf_gp_exceptions;
```

include/openenclave/bits/exception.h
```diff
@@ -154,6 +204,11 @@ typedef struct _oe_exception_record
 
     uint64_t address; /**< Exception address */
 
+    /* Information for PF/GP exceptions, which are only available
+     * for SGX2 and requires user opt-in (set CapturePFGPExceptions=1). */
```
dthaler, line 208:
```suggestion
     * for SGX2 and requires application opt-in (set CapturePFGPExceptions=1). */
```

include/openenclave/bits/sgx/sgxproperties.h
mingweishih: It turns out changing the argument names to lower_snake_case that are the same as the name of any struct members lead to compilation errors.

For example, the following would fail
```
#define _OE_SET_ENCLAVE_SGX_IMPL(                                         \
    product_id,                                                           \
    ...
```
Unfortunately, most of the upper_snake_case arguments here match the name of struct members.  I'd prefer leave it unchanged for now. Any comments? @dthaler

mingweishih:
bors try

include/openenclave/bits/sgx/sgxproperties.h
dthaler: Ok to file a separate issue to track and deal with it in a future PR

mingweishih: https://github.com/openenclave/openenclave/issues/3998

include/openenclave/bits/sgx/sgxproperties.h
mingweishih: Filed as a separated issue: https://github.com/openenclave/openenclave/issues/3998

mingweishih:
@bodzhang can you take another look?

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -223,6 +231,54 @@ typedef struct _oe_sgx_enclave_properties
...
+ * OE_ENCLAVE_FLAG_DEBUG and debugged at runtime
+ * @param[in] CAPTURE_PF_GP_EXCEPTIONS If true, allows the enclave to capture #PF/#GP
+ * exceptions if CPU supports the feature. The setting is ignored otherwise
+ * (SGX2 feature)
+ * @param[in] HEAP_PAGE_COUNT Number of heap pages to allocate in the enclave
+ * @param[in] STACK_PAGE_COUNT Number of stack pages per thread to reserve in
+ * the enclave
+ * @param[in] TCS_COUNT Number of concurrent threads in an enclave to support
+ */
+ #define OE_SET_ENCLAVE_SGX2(                                             \
```
bodzhang, line 264: A developer might want use KSS and PF_GP_EXCEPTIONS. This macro does not support that.

mingweishih: `EXTENDED_PRODUCT_ID` and `FAMILY_ID` are KSS-specific properties. The KSS flag should be set if any of these properties is set. This is the case for oesign: https://github.com/openenclave/openenclave/blob/d5155f48fc9958b017654998e1adf1fdf629906b/tools/oesign/oesign.c#L465.
If this is not the case for this macro, we should have a separate PR to fix it.

host/sgx/sgxsign.c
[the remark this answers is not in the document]
lzha101: @mingweishih Hi Mingwei, what I mean is not to let the loader to set the SIGSTRUCT.MISCMASK, but let the loader to set the SECS.MISCSELECT dynamically. It should be the oesign to set the SIGSTRUCT.MISCSELECT[0]=0 and SIGSTRUCT.MISCMASK=0xFFFFFFFE.  That's why I said only to change the default setting of SIGSTRUCT.MISCMASK as the default value of SIGSTRUCT.MISCSELECT[0] is already 0. 
If my understanding for OE implementation is correct, seems OE implementation shares most the same code for measurement between oesign and loader.  Looks changing loader to dynamically set the SECS.MISCSELECT means oesign would set the SECS.MISCSELECT dynamically as well. However, the attributes setting in SECS during signing doesn't really impact the measurement. It is not a factor of the enclave hash calculation.  
https://github.com/openenclave/openenclave/blob/cd72fd7069488ba6f453c8f5f47bd9fd9a6e6c0d/host/sgx/sgxload.c#L363-L367
you can see in '[oe_sgx_measure_create_enclave()](https://github.com/openenclave/openenclave/blob/cd72fd7069488ba6f453c8f5f47bd9fd9a6e6c0d/common/sgx/sgxmeasure.c#L64)' function, it measures secs->ssaframesize and secs->size, no attributes. 
 
During loading,  SECS will be set to enclave_common_loader for ECREATE which means the attributes in SECS will take effect. 
https://github.com/openenclave/openenclave/blob/cd72fd7069488ba6f453c8f5f47bd9fd9a6e6c0d/host/sgx/sgxload.c#L384-L391
Then EINIT will check the attributes setting between SECS and SIGSTRUCT at loading time.

So I suppose it should be OK if we sign an enclave on either SGX1 or SGX2 platform and then load the enclave on either SGX1 platform without MISC feature or SGX2 platform with MISC enabled.

The above is just an explanation. I totally agree your current implementation (providing option to let users decide to opt-in the misc features) is the other option and workable . :)

include/openenclave/bits/sgx/sgxproperties.h
mingweishih: I've updated the PR to expose the `REQUIRE_KSS` option in the new macro. However, I found that there is an inconsistency between using oesign + config file versus using MACRO, which should require a fix.

In the oesign + config scenario, the oesign will patch the oeinfo section based on the config file. For the case of KSS, the `SGX_FLAG_KSS` flag will be set if `EXTENDED_PRODUCT_ID` and `FAMILY_ID` is specified in the config file.

For the macro scenario, the `SGX_FLAG_KSS` is explicitly set based on the `REQUIRE_KSS` flag. If the `REQUIRE_KSS` flag is set to `true`, the `SGX_FLAG_KSS` will be set in the oeinfo section. Now even if the enclave is signed by oesign with both `EXTENDED_PRODUCT_ID` and `FAMILY_ID` unspecified in the config file, the `SGX_FLAG_KSS` will remain `true` in the oeinfo section.

Regarding the macro design, based on the current implementation, it may not be possible to support what you suggested. An `OE_SET_ENCLAVE_SGX*` macro, internally invoked the `_OE_SET_ENCLAVE_SGX_IMPL` macro, which effectively initializes the data structure of the oeinfo section at compile time. The struct initialization cannot be done separately, unless we have newly added features defined in separated structs and append them to the oeinfo structure. This requires some work to the loader stack and will not be needed by the oesign + config scenario. Do you think it's worth reworking the marco? Also note that the use of marco should be mainly for debug purposes (e.g., we only use it in our tests) and we should discourage using it in production enclaves.

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -114,7 +119,8 @@ typedef struct _oe_sgx_enclave_properties
     EXTENDED_PRODUCT_ID,                                                  \
     FAMILY_ID,                                                            \
     ALLOW_DEBUG,                                                          \
-    REQUIRE_KSS,                                                            \
+    REQUIRE_KSS,                                                          \
+    CAPTURE_PF_GP_EXCEPTIONS,                                             \
```
lzha101, line 123: Seems the comments for this macro need to be updated to add PFGP description.

mingweishih: Thanks for the catch! Updated.

mingweishih:
bors try

mingweishih:
bors try

include/openenclave/bits/sgx/sgxproperties.h
bodzhang: @mingweishih, thanks for the explanation of the challenges to implement a scalable MACRO solution. I think it's OK to start with the current implementation in the CapturePFGPExceptions PR and explore better solutions in future PRs.

bodzhang (approved):

haitaohuang:
Changes all look good to me. I just wonder if it will cause issues we only require opt-out and by default runtime loader to enable MISC Select for exinfo if platform supports. The reason I'm asking is EDMM implementation would require this feature enabled set per #3755, so it may be more streamlined experience if user only need to ask for EDMM instead of asking for both.

mingweishih:
> Changes all look good to me. I just wonder if it will cause issues we only require opt-out and by default runtime loader to enable MISC Select for exinfo if platform supports. The reason I'm asking is EDMM implementation would require this feature enabled set per #3755, so it may be more streamlined experience if user only need to ask for EDMM instead of asking for both.

Good point. With this PR, the loader only sets the MISC exinfo bit when user opt-into the CapturePFGPExceptions feature. However, it seems the the EDMM requires the feature as well. We can have more discussion on the user experience when we add the EDMM support.

CHANGELOG.md
```diff
@@ -19,6 +19,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - The tool, currently under the `tools` directory, was originally named `oecert` under the `tests/tools` directory.
   - The tool can be used to generate certificates, reports, and evidence in various formats.
   - The tool is for debugging purposes and is not suitable for production use.
+- Add the CapturePFGPExceptions configuration for SGX enclaves, which allows opting into the SGX2 feature of capturing #PF/#GP exceptions inside an enclave.
```
dthaler, line 22:
```suggestion
- Add the CapturePFGPExceptions configuration for SGX enclaves, which allows opting into the SGX2 feature of capturing #PF and #GP exceptions inside an enclave.
```

CHANGELOG.md
```diff
@@ -19,6 +19,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Developers can specify the CapturePFGPExceptions with a binary value in the enclave config file or set the value via the newly added OE_SET_ENCLAVE_SGX2 m
+acro, which is used to set SGX2-specific properties.
```
dthaler, line 24:
```suggestion
  - Developers can specify the CapturePFGPExceptions with a binary value in the enclave config file or set the value via the newly added OE_SET_ENCLAVE_SGX2
  macro, which is used to set SGX2-specific properties.
```

CHANGELOG.md
```diff
@@ -19,6 +19,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - After opting into the feature (value=1), the in-enclave exception handler can capture the #PF (with the OE_EXCEPTION_PAGE_FAULT code) and #GP (with the co
+de OE_EXCEPTION_ACCESS_VIOLATION) exceptions.
```
dthaler, line 26:
```suggestion
  - After opting into the feature (value=1), the in-enclave exception handler can capture the #PF (with the OE_EXCEPTION_PAGE_FAULT code) and #GP (with the
OE_EXCEPTION_ACCESS_VIOLATION code) exceptions.
```

CHANGELOG.md
```diff
@@ -19,6 +19,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - More information about the exceptions can be found in the `faulting_address` and `error_code` members of the `oe_exception_record_t` structure passed into
+ the handler.
+  - The feature is only available on SGX2 CPUs and is ignored on SGX1 CPUs.
```
dthaler, line 29: What does it mean for a feature to be ignored? (as opposed to not available)

If you meant to say that the mechanism to _opt-in_ is ignored, then that's not a great developer experience to have a developer opt in and have nothing happen.   It's usually better to fail the opt-in than to not adhere to a contract.  The alternative would be to rename the opt-in to be a _preference_ rather than a directive.

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -100,6 +105,9 @@ typedef struct _oe_sgx_enclave_properties
  * OE_ENCLAVE_FLAG_DEBUG and debugged at runtime
  * @param[in] REQUIRE_KSS If true, allows the enclave to be created with
  * kss properties
+ * @param[in] CAPTURE_PF_GP_EXCEPTIONS If true, allows the enclave to capture
+ * #PF/#GP exceptions if CPU supports the feature. The setting is ignored
```
dthaler, line 109:
```suggestion
 * #PF and #GP exceptions if the CPU supports the feature. The setting is ignored
```

include/openenclave/bits/sgx/sgxproperties.h
dthaler: to answer the question, you just have to use names that don't collide, like prefix with underscore, or add some suffix

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -223,6 +234,57 @@ typedef struct _oe_sgx_enclave_properties
...
+ * @param[in] SECURITY_VERSION ISV assigned Security Version number (ISVSVN)
+ * to use in the enclave signature
+ * @param[in] EXTENDED_PRODUCT_ID ISV assigned Extended Product ID (ISVEXTPRODID)
+ *  to use in the enclave signature (SGX2 feature)
+ * @param[in] FAMILY_ID ISV assigned Product Family ID (ISVFAMILYID)
+ * to use in the enclave signature (SGX2 feature)
+ * @param[in] ALLOW_DEBUG If true, allows the enclave to be created with
+ * OE_ENCLAVE_FLAG_DEBUG and debugged at runtime
+ * @param[in] REQUIRE_KSS If true, allows the enclave to be created with
+ * kss properties (SGX2 feature)
```
dthaler, line 260:
```suggestion
 * KSS properties (SGX2 feature)
```

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -223,6 +234,57 @@ typedef struct _oe_sgx_enclave_properties
...
+ * @param[in] CAPTURE_PF_GP_EXCEPTIONS If true, allows the enclave to capture
+ * #PF/#GP exceptions if CPU supports the feature. The setting is ignored
```
dthaler, line 262:
```suggestion
 * #PF and #GP exceptions if the CPU supports the feature. The setting is ignored
```

tests/pf_gp_exceptions/enc/enc.c
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/enclave.h>
+#include <openenclave/internal/print.h>
+#include <openenclave/internal/tests.h>
+#include "pf_gp_exceptions_t.h"
+
+uint64_t faulting_address;
+uint32_t error_code;
```
dthaler, line 10: can these be static?

tests/pf_gp_exceptions/enc/enc.c
```diff
@@ -0,0 +1,74 @@
...
+    {
+        const int LGDT_INSTRUCTION_BYTES = 8;
+        faulting_address = exception_record->faulting_address;
+        error_code = exception_record->error_code;
+        exception_record->context->rip += LGDT_INSTRUCTION_BYTES;
+        oe_host_printf(
+            "#GP: addr: 0x%lx, code: %u\n",
+            exception_record->faulting_address,
+            exception_record->error_code);
+    }
```
dthaler, line 35: Line 25 isn't just "else" but rather an else if, so is there a third case?
Suggest adding another else that would (say) fail the test if it activates with any other code.

tests/pf_gp_exceptions/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+if (BUILD_ENCLAVES)
+  add_subdirectory(enc)
+endif ()
+
+add_enclave_test(tests/pf_gp_exceptions pf_gp_exceptions_host
+                 pf_gp_exceptions_enc)
+
+add_enclave_test(tests/pf_gp_exceptions_unsigned pf_gp_exceptions_host
+                 pf_gp_exceptions_enc_unsigned)
```
dthaler, line 14: Since this test is SGX specific, the binary should have sgx in the name

tests/pf_gp_exceptions/pf_gp_exceptions.edl
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+enclave {
+    from "openenclave/edl/logging.edl" import oe_write_ocall;
+    from "openenclave/edl/fcntl.edl" import *;
```
dthaler, line 6: can these be listed in alphabetical order?

CHANGELOG.md
mingweishih: Updated the text, can you check again?

include/openenclave/bits/sgx/sgxproperties.h
```diff
@@ -34,13 +34,18 @@ typedef struct _oe_sgx_enclave_image_info_t
 #define OE_SGX_FLAGS_KSS 0x0000000000000080ULL
 #define OE_SGX_SIGSTRUCT_SIZE 1808
 
+typedef struct _oe_sgx_enclave_flags_t
+{
+    uint32_t capture_pf_gp_exceptions : 1;
+    uint32_t reserved : 31;
+} oe_sgx_enclave_flags_t;
```
anakrish, line 41: Out of curiosity, why do we need a separate structure instead of fields in the parent struct?

include/openenclave/internal/sgx/td.h
```diff
@@ -163,6 +163,14 @@ typedef struct _td
     oe_tls_atexit_t* tls_atexit_functions;
     uint64_t num_tls_atexit_functions;
 
+    /* The following information is only available for SGX2 and MISCSELECT is
+     * set to 1 */
+    /* The faulting address associated with the PF exception. Should be zero
+     * for other exception types. */
+    uint64_t faulting_address;
```
anakrish, line 170: We need to ensure that these are cleared appropriately so that their values don't linger around.

anakrish, line 170: We need to ensure that these are cleared appropriately so that their values don't linger around.

tests/pf_gp_exceptions/enc/enc.c
```diff
@@ -0,0 +1,79 @@
...
+    oe_result_t result;
+
+    result = oe_add_vectored_exception_handler(false, test_pfgp_handler);
+    if (result != OE_OK)
+    {
+        return -1;
+    }
+
+    /* Trigger #PF */
+    uint64_t unmapped_address = 1;
```
anakrish, line 56: Can this be 0 instead of 1? One of the key uses of handling PF is to support language runtimes that rely on null-pointer dereference page faults.

tests/pf_gp_exceptions/host/host.c
```diff
@@ -0,0 +1,67 @@
...
+        fprintf(stderr, "Usage: %s ENCLAVE_PATH\n", argv[0]);
+        return 1;
+    }
+
+    const uint32_t flags = oe_get_create_flags();
+
+    result = oe_create_pf_gp_exceptions_enclave(
+        argv[1], OE_ENCLAVE_TYPE_SGX, flags, NULL, 0, &enclave);
+    /* The enclave creation should succeed on both SGX1 and SGX2 machines. */
+    if (result != OE_OK)
```
anakrish, line 45: Should enclave creation succeed on SGX1 machines? If the enclave relies on PFGP handling, then it cannot possibly work on SGX1 machines. Therefore isn't it better to fail early at enclave creation itself?

anakrish (commented):
The implementation looks good overall.
Couple of questions
- How will the implementation get locked down in CI?
- Is this an experimental feature?

tests/pf_gp_exceptions/host/host.c
bodzhang: Depending on the usage, it might be acceptable for the SGX enclave to abort operation when the trust-worthy #PF/#GP info is not available inside the Enclave (for example, on SGX1) and to attempt to handle #PF/#GP inside the enclave when the trust-worth info is available. For enclaves that can't function without trust-worthy #PF/#GP info, and the developer would rather to fail the enclave load on SGX1 CPU, the developer can set SIGSTRUCT.MISCSELECT and SIGSTRUCT.MICSMASK to reflect the policy of failing Enclave initialization if #FP/#GP MISCSELECT is not enabled, either because the untrusted loader on SGX2 CPU does not follow developer request to enable #FP/#GP MISCSELECT, or the CPU does not support #FP/#GP MISCSELECT.

CHANGELOG.md
```diff
@@ -19,6 +19,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - The tool, currently under the `tools` directory, was originally named `oecert` under the `tests/tools` directory.
   - The tool can be used to generate certificates, reports, and evidence in various formats.
   - The tool is for debugging purposes and is not suitable for production use.
+- Add the CapturePFGPExceptions preference for the SGX2 feature of capturing #PF and #GP exceptions inside an enclave.
+  - Developers can specify the CapturePFGPExceptions with a binary value in the enclave config file or set the value via the newly added OE_SET_ENCLAVE_SGX2 macro, which is used to set SGX2-specific properties.
+  - When setting CapturePFGPExceptions=1, the OE loader will enable the feature when running on an SGX2-capable CPU.
+  - Once enabled, the in-enclave exception handler can capture the #PF (with the OE_EXCEPTION_PAGE_FAULT code) and #GP (with the code OE_EXCEPTION_ACCESS_VIOLATION code) exceptions.
+  - More information about the exceptions can be found in the `faulting_address` and `error_code` members of the `oe_exception_record_t` structure passed into the handler.
```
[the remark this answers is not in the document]
mingweishih: The interpretation of the code is in the exception.h

include/openenclave/bits/sgx/sgxproperties.h
mingweishih: The oe_sgx_enclave_flag_t is passed as an argument such that we can pass down the information to enclave signing functions.

Current behavior is
- the oe_sgx_enclave_flags_t struct is initialized based on the _OE_SET_ENCLAVE_SGX macro. The oesign tool can overwrite the struct based on the conf file. After that, the struct is passing down to `oe_sgx_sign_enclave_*` functions as an argument,

enclave/core/sgx/exception.c
```diff
@@ -364,6 +367,16 @@ void oe_virtual_exception_dispatcher(
...
     else
     {
+        /* The following codes can only be captured with SGX2 and the
+         * MISCSELECT[0] bit is set to 1. */
+        if (td->exception_code == OE_EXCEPTION_PAGE_FAULT ||
+            td->exception_code == OE_EXCEPTION_ACCESS_VIOLATION)
+        {
+            sgx_exinfo_t* exinfo =
+                (sgx_exinfo_t*)(gprsgx_offset - OE_SGX_MISC_BYTE_SIZE);
+            td->faulting_address = exinfo->maddr;
```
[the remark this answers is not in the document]
mingweishih: Updated. Now they're cleared after all the registered exception handlers finish

tests/pf_gp_exceptions/enc/enc.c
mingweishih: Done

CHANGELOG.md
```diff
@@ -19,6 +19,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - The tool, currently under the `tools` directory, was originally named `oecert` under the `tests/tools` directory.
   - The tool can be used to generate certificates, reports, and evidence in various formats.
   - The tool is for debugging purposes and is not suitable for production use.
+- Add the CapturePFGPExceptions preference for the SGX2 feature of capturing #PF and #GP exceptions inside an enclave.
+  - Developers can specify the CapturePFGPExceptions with a binary value in the enclave config file or set the value via the newly added OE_SET_ENCLAVE_SGX2 macro, which is used to set SGX2-specific properties.
+  - When setting CapturePFGPExceptions=1, the OE loader will enable the feature when running on an SGX2-capable CPU.
```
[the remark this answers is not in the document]
mingweishih: No, currently the behavior is that it does not fail. This allows the enclave to run on both SGX1 and SGX2 platforms without re-compilation.

tests/pf_gp_exceptions/host/host.c
mingweishih: @anakrish  @bodzhang I can see the two options that we can provide to developers
1. once opt-in, the enclave can run on both SGX1 and SGX2 CPU, but the #PF/#GP info will only be available on the latter
2. once opt-in, the enclave is only allowed to run on SGX2 CPU

The PR currently implement 1 only. Do you think we need to support both?

mingweishih:
bors try

mingweishih:
> The implementation looks good overall.
> Couple of questions
> 
> * How will the implementation get locked down in CI?
> * Is this an experimental feature?

We cannot fully test the feature without SGX2 CPU on CI.  The PR currently only tests the behavior on SGX1 (enclave creation should succeed).

This feature is similar to KSS, which only adds SGX2-specific configuration/information. I'm in favor of not making it as experimental feature.

tests/pf_gp_exceptions/host/host.c
anakrish: Given, @bodzhang 's answer above, #1 is sufficient. The enclave author can get the behavior of #2 by using the SIGSTRUCT.MISCSELECT and SIGSTRUCT.MISCMASK fields.

anakrish (approved):
LGTM overall. I'm OK for this to go in even though we won't lock down SGX2 in CI for now.
Please do create issues to track adding tests for SGX2.

PR #3991, opened 2021-05-02
Add SGX Enclave  Memory Manager

haitaohuang:
Signed-off-by: Haitao Huang <4699115+haitaohuang@users.noreply.github.com>


docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
+SGX Enclave Memory Manager
+=================================
+
+## Motivation ##
+
+An enclave's memory is backed by special reserved region in RAM, called Enclave Page Cache(EPC). Enclave memory management tasks include allocating/reserving virtual address ranges, committing physical EPC pages, changing EPC page permissions or page type, or removing EPC pages. Those tasks require collaboration between trusted runtime, untrusted runtime and the OS. The SGX enclave memory manager (EMM) serves as a central component in the enclave trusted runtime that abstracts the interaction with untrusted runtime for all memory management flows and provides APIs for its clients to reserve virtual address ranges, commit EPC memory to the reserved address ranges, modify attributes of the reserved/committed pages.
```
dthaler, line 6: nit: please split lines into (say) 80 column width, which makes git diffs easier to follow.

dthaler, line 6:
```suggestion
An enclave's memory is backed by a special reserved region in RAM, called Enclave Page Cache (EPC). Enclave memory management tasks include allocating/reserving virtual address ranges, committing physical EPC pages, changing EPC page permissions or page type, and removing EPC pages. Those tasks require collaboration between the trusted runtime, the untrusted runtime, and the OS. The SGX enclave memory manager (EMM) serves as a central component in the enclave trusted runtime that abstracts the interaction with the untrusted runtime for all memory management flows and provides APIs for its clients to reserve virtual address ranges, commit EPC memory to the reserved address ranges, and modify attributes of the reserved/committed pages.   
```
fix grammar

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+
+An allocation, once created, will own its address range until the deallocation API, sgx_mm_dealloc, is called upon. No two active allocations can have overlapping address ranges.
+
+**Commit, Uncommit Enclave Memory**
+
+When a page in COMMIT_ON_DEMAND allocations is accessed, a page fault occurs if the page was not yet commited.  The EMM will perform EACCEPT to commit the EPC page on page fault after OS doing EAUG. 
+
+The clients can also call the EMM commit API, sgx_mm_commit, to proactively commit specific sub-regions in a COMMIT_ON_DEMAND allocation to avoid future page fault.
+
+Some EMM clients, <i>e.g.</i>, a dynamic code loader wishing to load code on page fault, can register a custom handler for page fault at the time of allocation request. In the custom page fault handler, it can invoke an API, sgx_mm_commit_data, to commit and load data to newly committed EPC page at the same time as supported by EACCEPTCOPY.
```
dthaler, line 29:
```suggestion
Some EMM clients, <i>e.g.</i>, a dynamic code loader wishing to load code on page faults, can register a custom handler for page faults at the time of allocation request. In the custom page fault handler, it can invoke an API, sgx_mm_commit_data, to commit and load data to newly committed EPC page at the same time as supported by EACCEPTCOPY. 
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+Committed pages will stay committed (regardless how they were committed) until the clients calls the uncommit API, sgx_mm_uncommit, on them or the allocation they belong to is deallocated by sgx_mm_dealloc.    
+
+**Modify Page Attributes**
+
+The EMM clients may call sgx_mm_modify/sgx_mm_modify_ex to request permissions and/or page type changes for pages in existing allocations.
+
+
+## Notes on Internal Design ##
+
+The enclave memory manager keeps track of memory allocation and layout info inside enclave address range (ELRANGE) using an internal structure called Enclave Memory Area (EMA) List. The EMA and EMA list are considered private data structure of memory manager, and their internals are not exposed in client-facing APIs.
```
dthaler, line 40:
```suggestion
The enclave memory manager keeps track of memory allocation and layout info inside enclave address range (ELRANGE) using an internal structure called the Enclave Memory Area (EMA) List. The EMA and EMA list are considered private data structure of memory manager, and their internals are not exposed in client-facing APIs.
```

Also I cannot parse the grammar of "EMA and EMA list are considered private data structure of memory manager".   structure is singular but 'are' is plural.  Are "EMA" and "EMA list" two different things?

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+- The EMA list tracks all memory regions in use (reserved, committed, commit-on-demand) in ELRANGE.
+- Ranges in ELRANGE not tracked by an EMA are considered free and ready for new allocations
+- The EMM labels certain EMAs reserved for runtime or its internal usage and make them not accessible from public APIs.
+- A thread calling an EMM API on an EMA with operation pending in another thread will wait until the pending operation is finished. 
+
+**Assumptions:**
+
+- When enclave is loaded, the OS has reserved the whole address range covered by ELRANGE. It is assumed the host app will not remap any part of this reserved range.
+  - In future, when the new feature that allows loading enclave with base address at zero, we may support partial ELRANGE reserved by the OS.
```
dthaler, line 49:
```suggestion
  - In the future, when the new feature that allows loading an enclave with base address at zero, we may support partial ELRANGE reserved by the OS.  
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+- The EMA list tracks all memory regions in use (reserved, committed, commit-on-demand) in ELRANGE.
+- Ranges in ELRANGE not tracked by an EMA are considered free and ready for new allocations
+- The EMM labels certain EMAs reserved for runtime or its internal usage and make them not accessible from public APIs.
+- A thread calling an EMM API on an EMA with operation pending in another thread will wait until the pending operation is finished. 
+
+**Assumptions:**
+
+- When enclave is loaded, the OS has reserved the whole address range covered by ELRANGE. It is assumed the host app will not remap any part of this reserved range.
+  - In future, when the new feature that allows loading enclave with base address at zero, we may support partial ELRANGE reserved by the OS.  
+- Memory manager does not check EPC pressure, or proactively trim pages when EPC runs low. The OS can reclaim EPC pages when EPC running low or cgroups threshold reached
```
dthaler, line 50:
```suggestion
- The memory manager does not check EPC pressure, or proactively trim pages when EPC runs low. The OS can reclaim EPC pages when EPC is running low or the cgroups threshold is reached
```

Also regarding "when cgroups threshold reached": how does this apply to Windows?

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+- Memory manager does not maintain and recycle committed then freed pages
```
dthaler, line 51:
```suggestion
- The memory manager does not maintain and recycle committed then freed pages
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+  - Whenever a page is freed (via dealloc or uncommit API), it is trimmed from enclave and need be re-allocated and committed before re-use.
```
dthaler, line 52:
```suggestion
  - Whenever a page is freed (via dealloc or uncommit API), it is trimmed from enclave and needs to be re-allocated and committed before re-use.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+  - Owner of a region can re-purpose a sub-region of it by calling sgx_mm_modify_type/permissions to split out the sub-region to be reused.
```
dthaler, line 53:
```suggestion
  - The owner of a region can re-purpose a sub-region of it by calling sgx_mm_modify_type/permissions to split out the sub-region to be reused.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+- Memory manager does not call back client for #GP handling. Memory manager code will ensure itself would not cause #GP, and only register a #PF handler with the enclave global exception handler registry. Clients wish to handle #GP can register its own exception handler with the global handler registry.
```
dthaler, line 54:
```suggestion
- The memory manager does not call back into the client for #GP handling. Memory manager code will ensure that it itself would not cause #GP, and only registers a #PF handler with the enclave global exception handler registry. A client wanting to handle #GP can register its own exception handler with the global handler registry.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+
+
+Public APIs
```
dthaler, line 57: In which library file?

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+/**
+ * Custom page fault (#PF) handler, do usage specific processing upon #PF, e.g. loading data
+ * and verify its trustworthiness, then call sgx_mm_commit_data to explicitly EACCEPTCOPY data.
+ * This custom handler is passed into sgx_mm_alloc, and associated with the newly allocated region.
+ * The memory manager calls the handler when a #PF happens in the associated region.
+ * 
+ * @param[in] pfinfo, info reported in the SSA MISC region for page fault
+ * @param[in] private_data, private data provided by handler in sgx_mm_alloc call.
+ * @return EXCEPTION_CONTINUE_EXECUTION on success handling the exception.
+ *    abort if internal corruption detected, Otherwise EXCEPTION_CONTINUE_SEARCH.
```
dthaler, line 93:
```suggestion
 *    abort if internal corruption detected. Otherwise EXCEPTION_CONTINUE_SEARCH.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+* @param[in] flags a bitwise OR of flags describing committing mode, committing order, address preference.
+*        Flags should include exactly one of following for committing mode:
+*            - EMA_RESERVE: just reserve an address range, no EPC commited. To allocate memory on a reserved range, 
+*                           call this function again with EMA_COMMIT_ON_DEMAND or EMA_COMMIT_NOW.
+*            - EMA_COMMIT_NOW: reserves memory range and commit EPC pages. EACCEPT for all allocated pages 
+*                              are done when this function returns.
+*            - EMA_COMMIT_ON_DEMAND: reserves memory range, EPC pages are committed (EACCEPT) on demand upon #PF.
+*        ORed with zero or one of the committing order flags:
+*            - EMA_GROWSDOWN: always commit pages from higher to lower addresses, no gaps in addresses above the last commited
+*            - EMA_GROWSUP:   always commit pages from lower to higher addresses, no gaps in addresses below the last commited.
```
dthaler, line 110:
```suggestion
*            - EMA_GROWSDOWN: always commit pages from higher to lower addresses, no gaps in addresses above the last committed.
*            - EMA_GROWSUP:   always commit pages from lower to higher addresses, no gaps in addresses below the last committed.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+
+/**
+ * Custom page fault (#PF) handler, do usage specific processing upon #PF, e.g. loading data
+ * and verify its trustworthiness, then call sgx_mm_commit_data to explicitly EACCEPTCOPY data.
+ * This custom handler is passed into sgx_mm_alloc, and associated with the newly allocated region.
+ * The memory manager calls the handler when a #PF happens in the associated region.
+ * 
+ * @param[in] pfinfo, info reported in the SSA MISC region for page fault
+ * @param[in] private_data, private data provided by handler in sgx_mm_alloc call.
+ * @return EXCEPTION_CONTINUE_EXECUTION on success handling the exception.
```
dthaler, line 92: use `retval` not `return` for discrete values

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+
+
+### sgx_mm_uncommit and sgx_mm_dealloc
+
+```
+/**
+* Uncommit (trim) physical EPC pages in a previously committed range. 
+* The pages in the allocation are freed, but the address range is still reserved.
+* @param[in] addr page aligned start address of the region to be trimmed
+* @param[in] length size in bytes of multiples of page size.
```
dthaler, line 137: nit: be consistent in terms of whether argument descriptions in in period or not

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+- Memory manager does not maintain and recycle committed then freed pages
+  - Whenever a page is freed (via dealloc or uncommit API), it is trimmed from enclave and need be re-allocated and committed before re-use.
+  - Owner of a region can re-purpose a sub-region of it by calling sgx_mm_modify_type/permissions to split out the sub-region to be reused.
+- Memory manager does not call back client for #GP handling. Memory manager code will ensure itself would not cause #GP, and only register a #PF handler with the enclave global exception handler registry. Clients wish to handle #GP can register its own exception handler with the global handler registry.
+
+
+Public APIs
+-----------------
+
+### sgx_mm_alloc
```
dthaler, line 60: Also, are these proposed as APIs to be added to OE?  (where APIs would be prefixed with `oe_`)

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
+SGX Enclave Memory Manager
+=================================
+
+## Motivation ##
+
+An enclave's memory is backed by a special reserved region in RAM, called Enclave Page Cache (EPC). Enclave memory management tasks include allocating/reserving virtual address ranges, committing physical EPC pages, changing EPC page permissions or page type, and removing EPC pages. Those tasks require collaboration between the trusted runtime, the untrusted runtime, and the OS. The SGX enclave memory manager (EMM) serves as a central component in the enclave trusted runtime that abstracts the interaction with the untrusted runtime for all memory management flows and provides APIs for its clients to reserve virtual address ranges, commit EPC memory to the reserved address ranges, and modify attributes of the reserved/committed pages.
```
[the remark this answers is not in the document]
mingweishih: Few comments/questions:
- The paragraph mentions "... collaboration among trusted runtime, untrusted runtime, and the OS." However, I don't see the OS is mentioned in the following sentences.
- Does EMM run in a dedicated enclave or part of each enclave application? That is, does a "client" refer to the enclave which talks to the EMM enclave or the enclave includes the EMM. Can you clarify a bit more?

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+ * and verify its trustworthiness, then call sgx_mm_commit_data to explicitly EACCEPTCOPY data.
+ * This custom handler is passed into sgx_mm_alloc, and associated with the newly allocated region.
+ * The memory manager calls the handler when a #PF happens in the associated region.
+ * 
+ * @param[in] pfinfo, info reported in the SSA MISC region for page fault
+ * @param[in] private_data, private data provided by handler in sgx_mm_alloc call.
+ * @return EXCEPTION_CONTINUE_EXECUTION on success handling the exception.
+ *    abort if internal corruption detected. Otherwise EXCEPTION_CONTINUE_SEARCH.
+ */
+typedef int (*enclave_fault_handler_t)(const sgx_pfinfo* pfinfo, void* private_data);
```
mingweishih, line 95: Note that if we want to implement this on OE, we should reuse OE's exception handler interfaces.

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+        struct {//PFEC bits
+            uint32_t p       : 1;    // P flag
+            uint32_t rw      : 1;    // RW access flag, 0 for read, 1 for write
+            uint32_t         : 13;   // U/S, I/O, PK and reserved bits not relevant for SGX PF
+            uint32_t sgx     : 1;    // SGX bit
+            uint32_t         : 16;   // reserved bits
+        };
+    } pfec;
+    uint32_t reserved; 
+} sgx_pfinfo;
```
mingweishih, line 82: Note from OE: the similar data structure will be supported by OE as part of #3984.

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+ *            PROT_EXEC
+ *            PROT_TRIM: change the page type to PT_TRIM, implies RW. 
+ *            PROT_TCS: change the page type to PT_TCS
+ *            PROT_SS_FIRST
+ *            PROT_SS_REST
+ *            PROT_NONE: Signal the kernel EACCEPT is done for PT_TRIM pages. 
+ * @return 0 on success, or error code similar to mprotect
+ */
+
+mprotect_ocall(void*addr, int flags)
```
mingweishih, line 309: missing function type and `;` at the end

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+ 
+
+EMM Initialization and Tracking Initially Committed (EADDed) Regions
+---------------------------------------------------------------------
+
+The memory manager must be initialized in first ECALL (ECMD_INIT_ENCLAVE in Intel SGX SDK) before any other clients can use it. Therefore, code and data of the memory manager will be part of initial enclave image that are loaded with EADD before EINIT, and as a part of the trusted runtime. 
+
+The trusted runtime should enumerate all initial committed regions (code, data, heap, stack, TCS, SSA), and call the EMM internal APIs to setup initial entries in the EMA list to track existing regions and mark some of them not modifiable by EMM public APIs. Runtime also ensures there is enough reserved space on heap for EMM to create the initial EMA list and the entries. Once initialized, the memory manager can reserve its own space for future expansion of the EMA list, and special EMAs to hold EMA objects. To keep it simple, the expansion can be done eagerly: commit more pages for EMA list once unused committed space in the EMA List Region below certain threshold.
```
mingweishih, line 325: Note for OE: This option indicates some work to the OE loader. Currently we don't explicitly pass the full enclave layout information (only necessary ones for relocations and heap allocators) to the enclave.

docs/DesignDocs/SGX_EMM.md
haitaohuang: The intent is to reuse implementation of these APIs between OE and Intel SGX sdk. I'm not very familiar with OE yet. From what I see, if we decide to expose these as common APIs across different technologies for OE, then we need add these APIs with oe_ prefix, and do weak alias for SGX impl?

docs/DesignDocs/SGX_EMM.md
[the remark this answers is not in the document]
haitaohuang: will do for next verion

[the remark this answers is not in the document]
haitaohuang: EMM does not directly work with OS, but rather thru interaction with  untrusted time. So no need to mention OS.

I mentioned typical clients below: The public EMM APIs defined here are most likely invoked by some intermediate runtime level components for specific usages, such as dynamic heap/stack, mmap, mprotect, higher level language JIT compiler, etc.

I think special reserved memory sounds better in this case or I should use a different term. I'll take other suggestions in next version

docs/DesignDocs/SGX_EMM.md
haitaohuang: This is a callback invoked by EMM, speciifc for PF. EMM implementation itself would need register runtime specific exception handler interface with the global registry

docs/DesignDocs/SGX_EMM.md
haitaohuang: will do for next version

docs/DesignDocs/SGX_EMM.md
haitaohuang: will do for next version

docs/DesignDocs/SGX_EMM.md
dthaler: Do you really need a weak alias?  I suspect a simple #define macro that defines one to be an alias for the other, would be sufficient.

mjp41:
`snmalloc` can be implemented easily against the API you are providing here.  The `snmalloc` requirements are explained here:

  https://github.com/microsoft/snmalloc/blob/master/docs/PORTING.md

The only addition from `snmalloc`s perspective that might be nice is a `MAP_ALIGNED(n)` that several of the BSDs support, and the equivalent supported by `VirtualAlloc2`.  `snmalloc` will allocate large ranges, and then manage that address space to find suitable alignments, if this is not provided.  This increases `snmalloc`s address space requirements without `MAP_ALIGNED` support.

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+*            - EMA_GROWSUP:   always commit pages from lower to higher addresses, no gaps in addresses below the last committed.
+*        Optionally ORed with EMA_FIXED to indicate allocation at fixed address. 
+*
+* @param[in] handler, custom handler for page faults in this region, NULL if no custom handling needed
+* @param[in] handler_private, private data for the @handler, which will be passed back when the handler is called.
+* @param[out] out_addr, pointer to store the start address of allocated range. Set to valid address by the function on success, NULL otherwise. 
+* @return 0 on success. EEXIST if range requested is overlapping a region in use and EMA_FIXED is requested,
+* ENOMEM for out of memory, EACCES if region is outside enclave address space.
+*/
+int sgx_mm_alloc(void *addr, size_t length, int flags, enclave_fault_handler_t handler, void* handler_private, void** out_addr)
```
dthaler, line 119:
```suggestion
int sgx_mm_alloc(void *address, size_t length, int flags, enclave_fault_handler_t handler, void* handler_private, void** out_address)
```
OE style guidelines say to avoid abbreviations

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+- The memory manager will track current permissions for each region, and can determine whether new permission requires OCALL for EMODPR, e.g RW<->RX, RW->R
+- This API should not be used to change EPC page type to PT_TRIM. Trimming pages are done by sgx_mm_uncommit and sgx_mm_dealloc only.
+
+
+### sgx_mm_commit
+```
+/**
+* Commit a partial or full range of memory allocated previously with EMA_COMMIT_ON_DEMAND.
+* The API will return 0 if all pages in the requested range are successfully committed. 
+* Calling this API on pages already committed has not effect.
```
dthaler, line 201:
```suggestion
* Calling this API on pages already committed has no effect. 
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,350 @@
...
+
+### sgx_mm_commit
+```
+/**
+* Commit a partial or full range of memory allocated previously with EMA_COMMIT_ON_DEMAND.
+* The API will return 0 if all pages in the requested range are successfully committed. 
+* Calling this API on pages already committed has not effect. 
+* @param[in] addr, page aligned starting address
+* @param[in] length, length of the region in bytes of multiples of page size.
+* @return 0 on success. EINVAL if any requested page is not in any previously allocated regions, -1 otherwise
```
dthaler, line 204: Why return -1 for other failures instead of a more specific error code from https://linux.die.net/man/3/errno or from `oe_result_t`?

docs/DesignDocs/SGX_EMM.md
dthaler: Also I don't see any text in this document about "reuse implementation".   Please explain the code relationship.  Does one depend on the other or should it be in a separate common library consumed by both?

dthaler:
What issue number is this for?   https://github.com/openenclave/openenclave/blob/master/docs/Community/Contributing.md#general-guidelines says

> * **DO** open an issue for design discussion before making any major changes.

haitaohuang:
> The only addition from `snmalloc`s perspective that might be nice is a `MAP_ALIGNED(n)` that several of the BSDs support, and the equivalent supported by `VirtualAlloc2`. `snmalloc` will allocate large ranges, and then manage that address space to find suitable alignments, if this is not provided. This increases `snmalloc`s address space requirements without `MAP_ALIGNED` support.

I'll add this support in next version. Thanks

haitaohuang:
will position EMM as a reusable library, and specify APIs of an abstraction layer for runtimes to implement. With that, coding style and naming convention of EMM would not be runtime specific.

docs/DesignDocs/SGX_EMM.md
haitaohuang: updated to return EFAULT for all other errors

docs/DesignDocs/SGX_EMM.md
haitaohuang: As we will position this as separate library, I'll keep abbreviations for now :-)

docs/DesignDocs/SGX_EMM.md
haitaohuang: Please see updated version with abstraction layer spec

docs/DesignDocs/SGX_EMM.md
haitaohuang: Per discussion in arch forum, we plan to reuse it as lib

radhikaj (approved):

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+changing EPC page permissions or page types, and removing EPC pages.
+Those tasks require collaboration between the trusted runtime, the untrusted
+runtime, and the OS. The SGX enclave memory manager (EMM) serves as a central
+component in the enclave trusted runtime that abstracts the interaction with
+the untrusted runtime for all memory management flows and provides APIs for
+its clients to reserve virtual address ranges, commit EPC memory to the reserved
+address ranges, and modify attributes of the reserved/committed pages.
+
+For details of specific memory management related flows, please refer to
+[the SGX EDMM driver API spec](https://github.com/openenclave/openenclave/pull/3755/files).
```
dthaler, line 18: design docs shouldn't refer to pull requests, only other docs

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+and uncommitting.
+
+An allocation, once created, will own its address range until the deallocation
+API, sgx_mm_dealloc, is called upon. No two active allocations can have
+overlapping address ranges.
+
+**Commit, Uncommit Enclave Memory**
+
+When a page in COMMIT_ON_DEMAND allocations is accessed, a page fault occurs if
+the page was not yet commited.  The EMM will perform EACCEPT to commit the EPC
```
dthaler, line 50:
```suggestion
the page was not yet committed.  The EMM will perform EACCEPT to commit the EPC
```
typo

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+
+**Modify Page Attributes**
+
+The EMM clients may call sgx_mm_modify/sgx_mm_modify_ex to request permissions
+and/or page type changes for pages in existing allocations.
+
+**Runtime Abstraction Layer**
+
+To make the EMM implementation portable across different SGX runtimes,
+e.g. OpenEnclave and Intel SGX SDKs, this document also proposes an abstraction
```
dthaler, line 75:
```suggestion
e.g., OpenEnclave and Intel SGX SDKs, this document also proposes an abstraction
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+#define EXCEPTION_CONTINUE_EXECUTION 0xFFFFFFFF
+
+/*
+ * The EMM page fault (#PF) handler
+ *
+ * @param[in] pfinfo, info reported in the SSA MISC region for page fault
+ * @retval EXCEPTION_CONTINUE_EXECUTION on success handling the exception.
+ * @retval EXCEPTION_CONTINUE_SEARCH if the EMM does not handle the exception.
+ */
+typedef int (*enclave_pfhandler_t)(const sgx_pfinfo *pfinfo);
```
dthaler, line 427:
```suggestion
typedef int (*sgx_enclave_pfhandler_t)(const sgx_pfinfo *pfinfo);
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+To support and use the EMM, an SGX trusted runtime shall implement following
+abstraction layer APIs.
+
+### Exception Handler Registration
+
+```
+/* Return value used by the EMM #PF handler to indicate
+ *  to the dispatcher that it should continue searching for the next handler.
+ */
+#define EXCEPTION_CONTINUE_SEARCH 0x0
```
dthaler, line 413:
```suggestion
#define SGX_EXCEPTION_CONTINUE_SEARCH 0x0
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+
+/* Return value used by the EMM #PF handler to indicate
+ *  to the dispatcher that it should stop searching and continue execution.
+ */
+#define EXCEPTION_CONTINUE_EXECUTION 0xFFFFFFFF
```
dthaler, line 418:
```suggestion
#define SGX_EXCEPTION_CONTINUE_EXECUTION 0xFFFFFFFF
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+
+/*
+ * Check whether the given buffer is strictly within the enclave.
+ *
+ * Check whether the buffer given by the **ptr** and **size** parameters is
+ * strictly within the enclave's memory. If so, return true. If any
+ * portion of the buffer lies outside the enclave's memory, return false.
+ *
+ * @param[in] ptr The pointer pointer to buffer.
+ * @param[in] size The size of buffer
```
dthaler, line 539:
```suggestion
 * @param[in] size The size of the buffer.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+    BitArray*           eaccept_map;    // bitmap for EACCEPT status, bit 0 in eaccept_map[0] for the page at start address
+                                        // bit i in eaccept_map[j] for page at start_address+(i+j<<3)<<12
+    sgx_mutex_t*        lock;           // lock to prevent concurrent modification
+    int                 transition;     // state to indicate whether a transition in progress, e.g page type/permission changes.
+    enclave_fault_handler_t
+                        h;              // custom PF handler  (for EACCEPTCOPY use)
+    void*               hprivate;       // private data for handler
+    ema_t*              next;           // next in doubly linked list
+    ema_t*              prev;           // prev in doubly linked list
+} ema_t;
```
dthaler, line 641:
```suggestion
} sgx_ema_t;
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+They can evolve over time, shown here for reference only.
+
+### Enclave Memory Area (EMA) struct
+
+Each enclave has a global doubly linked EMA list to keep track of all dynamically
+allocated regions in enclave address space (ELRANGE).
+
+```
+typedef BitArray;
+typedef struct _ema_t {
```
dthaler, line 625:
```suggestion
typedef struct _sgx_ema_t {
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+Following are internal functions and structures to be used by EMM implementation.
+They can evolve over time, shown here for reference only.
+
+### Enclave Memory Area (EMA) struct
+
+Each enclave has a global doubly linked EMA list to keep track of all dynamically
+allocated regions in enclave address space (ELRANGE).
+
+```
+typedef BitArray;
```
dthaler, line 624: what is BitArray?  should it be `sgx_bit_array_t`?

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+ * it maintains.
+ *
+ * @param[in] pfinfo, info reported in the SSA MISC region for page fault.
+ * @param[in] private_data, private data provided by handler in sgx_mm_alloc call.
+ * @retval EXCEPTION_CONTINUE_EXECUTION,  success on handling the exception.
+ * @retval EXCEPTION_CONTINUE_SEARCH, Exception not handled and should be passed
+ *         some other handler.
+ *
+ */
+typedef int (*enclave_fault_handler_t)(const sgx_pfinfo *pfinfo, void *private_data);
```
dthaler, line 162:
```suggestion
typedef int (*sgx_enclave_fault_handler_t)(const sgx_pfinfo *pfinfo, void *private_data);
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+
+/*
+ * Allocate a new memory region in enclave address space (ELRANGE).
+ * @param[in] addr, starting address of the region, page aligned. If NULL is provided,
+ *                  then the function will select the starting address.
+ * @param[in] length, size of the region in multiples of page size in bytes.
+ * @param[in] flags, a bitwise OR of flags describing committing mode, committing
+ * order, address preference, page type
+ *        Flags should include exactly one of following for committing mode:
+ *            - EMA_RESERVE: just reserve an address range, no EPC commited.
```
dthaler, line 229: prefix these defines with SGX_ if they are used with a function (like this one) that starts with sgx_
```suggestion
 *            - SGX_EMA_RESERVE: just reserve an address range, no EPC committed.
```
also note typo at end :)

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+
+/* Permissions flags */
+#define EMA_PROT_NOACCESS   0x0UL
+#define EMA_PROT_READ       0x1UL
+#define EMA_PROT_WRITE      0x2UL
+#define EMA_PROT_EXECUTE    0x4UL
+
+/*
+ * Allocate a new memory region in enclave address space (ELRANGE).
+ * @param[in] addr, starting address of the region, page aligned. If NULL is provided,
```
dthaler, line 223:
```suggestion
 * @param[in] addr starting address of the region, page aligned. If NULL is provided,
```
Don't put comma after param name in doxygen comments

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+ * This custom handler is passed into sgx_mm_alloc, and associated with the
+ * newly allocated region. The memory manager calls the handler when a #PF
+ * happens in the associated region. The handler may invoke abort() if it
+ * determines the exception is invalid based on certain internal states
+ * it maintains.
+ *
+ * @param[in] pfinfo, info reported in the SSA MISC region for page fault.
+ * @param[in] private_data, private data provided by handler in sgx_mm_alloc call.
+ * @retval EXCEPTION_CONTINUE_EXECUTION,  success on handling the exception.
+ * @retval EXCEPTION_CONTINUE_SEARCH, Exception not handled and should be passed
```
dthaler, line 158:
```suggestion
 * @retval EXCEPTION_CONTINUE_EXECUTION Success on handling the exception.
 * @retval EXCEPTION_CONTINUE_SEARCH Exception not handled and should be passed.
```

nit: Don't put comma after value name in doxygen comments, and be consistent in terms of whether the next word is capitalized or not, and whether the phrase ends in punctuation

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+ *                              to lower addresses, no gaps in addresses above the last committed.
+ *            - EMA_GROWSUP: if supported, a hint given for the kernel to EAUG pages from lower
+ *                              to higher addresses, no gaps in addresses below the last committed.
+ *        Optionally ORed with
+ *            -  EMA_ALIGNED(n):	Align the region on a requested	boundary.
+ *                           Fail if a suitable region cannot be found,
+ *                           The argument n specifies the binary logarithm of
+ *                           the desired alignmentand must be at least 12.
+ *        Optionally ORed with one of following page types:
+ *             - EMA_PAGE_TYPE_REG: regular page type.This is the default if not specified.
```
dthaler, line 479:
```suggestion
 *             - EMA_PAGE_TYPE_REG: regular page type. This is the default if not specified.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+ *             - EMA_PAGE_TYPE_SS_FIRST: the first page in shadow stack.
+ *             - EMA_PAGE_TYPE_SS_REST: the rest page in shadow stack. 
+ * @retval 0 on success.
+ * @retval EINVAL if any parameter passed in is not valid.
+ * @retval errno reported by dependent syscalls, e.g. mmap().
```
dthaler, line 484:
```suggestion
 * @retval errno reported by dependent syscalls, e.g., mmap().
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,679 @@
...
+ *        Optionally ORed with
+ *            -  EMA_ALIGNED(n):	Align the region on a requested	boundary.
+ *                           Fail if a suitable region cannot be found,
+ *                           The argument n specifies the binary logarithm of
+ *                           the desired alignmentand must be at least 12.
+ *        Optionally ORed with one of following page types:
+ *             - EMA_PAGE_TYPE_REG: regular page type.This is the default if not specified.
+ *             - EMA_PAGE_TYPE_SS_FIRST: the first page in shadow stack.
+ *             - EMA_PAGE_TYPE_SS_REST: the rest page in shadow stack. 
+ * @retval 0 on success.
```
dthaler, line 482:
```suggestion
 * @retval 0 The operation was successful.
```
Doxygen generates a table, so after the return value should be a statement to appear in the 2nd column of the table.  Recommend phrasing it like an error message, as a sentence.

anakrish:
@haitaohuang Thanks for this proposal. Looks very good.
My request would be to see if we can support SGX1 too using the same Memory Manager interface. That way OE could implement a mmap/munmap that just forwards to the Memory Manager and everything would automatically work without code changes when moving between SGX1 and SGX2. It is Ok that only a subset of the functionality works on SGX1 due to hardware limitation.

For reference, here is Mystikos' mmap implementation based on SGX1.
https://github.com/deislabs/mystikos/blob/main/kernel/mman.c

docs/DesignDocs/SGX_EMM.md
haitaohuang: will resolve this after driver interface merged

docs/DesignDocs/SGX_EMM.md
haitaohuang: will remove this definition. It is implementation detail and  the comments for eaccept_map should be enough for design doc.

haitaohuang:
@anakrish IIUC Mystikos mman implementation can still use the APIs here to reserve the whole range [base, end) with mm_alloc(..., SGX_EMA_COMMIT_ON_DEMAND), then reuse the same code to manage its BRK, MAP and VADs to track regions it allocates for user. The other differences on SGX2 now are following:
1. It needs to do sgx_mm_commit for each new allocation (triggered by sbrk, mmap, mremap).
2. sgx_mm_uncommit for deallocation (triggered by sbrk, munmap, mremap).
3. sgx_mm_modify_permission to change permissions in secure way.

For implementation on SGX1, it does not need to make any of those calls because the memory region is allocated and committed and no EPCM permissions change possible. So I expect those SGX2 code flow can be #ifdef out for SGX1 compilation.

On the other hand, for EMM to support both SGX1 and SGX2, we need add these limitations for SGX1:
- sgx_mm_allocate treats COMMIT_NOW and COMMIT_ON_DEMAND the same on SGX1.
- sgx_mm_commit/sgx_mm_uncommit has no effect.
- sgx_mm_commit_data on SGX1 platform will fail with an error code.
- sgx_mm_dealloc on SGX1 does not trim EPC and only marks pages as free forfuture allocations.
- sgx_mm_modify_page_type is not supported.
- sgx_mm_modify_ex/sgx_mm_modify_permissions will attempt to change OS PTE permissions only,
and can not enforce permissions in a trusted way on SGX1 platform.

It might save a few lines of code for upper layer by avoiding some #ifdefs. But I'm not sure it worth adding this extra layer for SGX1 runtime though.  
Your thoughts?

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+The public EMM APIs defined here are most likely invoked by some intermediate
+runtime level components for specific usages, such as dynamic heap/stack, mmap,
+mprotect, higher level language JIT compiler, etc.
+
+
+## User Experience ##
+
+**Allocate, Deallocate Enclave Memory**
+
+The EMM provides an API, sgx_mm_alloc, for clients to request enclave memory
```
dthaler, line 28: Clarify: when this says "the EMM provides an API", which SDK/library is this referring to?  PSW?  The Intel SGX SDK for Linux?  The Intel SGX SDK for Windows?  (Presumably it's not the OE SDK or the API name would start with oe_ instead.)  And what is the "client" here, is it an app or the OE SDK?  (I *think* this is answered down at line 78 but since it's way below, this text is confusing.  I might suggest moving the Runtime Abstraction Layer section starting at 72 up before this section.)

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+
+**Modify Page Attributes**
+
+The EMM clients may call sgx_mm_modify/sgx_mm_modify_ex to request permissions
+and/or page type changes for pages in existing allocations.
+
+**Runtime Abstraction Layer**
+
+To make the EMM implementation portable across different SGX runtimes,
+e.g., OpenEnclave and Intel SGX SDKs, this document also proposes an abstraction
```
dthaler, line 75:
```suggestion
e.g., Open Enclave and Intel SGX SDKs, this document also proposes an abstraction
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+to make necessary OCalls, register and receive callbacks on page faults. The EMM
+source code will be hosted and maintained in the [Intel SGX PSW and SDK repository](https://github.com/intel/linux-sgx)
+and can be built as a separate library to be linked into any runtime with the
+abstraction layer APIs implemented.
+
+## Notes on Internal Design ##
+
+The enclave memory manager keeps track of memory allocation and layout info inside
+enclave address range (ELRANGE) using an internal structure called the Enclave Memory
+Area (EMA) List. The EMA and EMA list are considered private data structure of memory
```
dthaler, line 86:
```suggestion
Area (EMA) List. The EMA and the EMA list are considered private data structures of the memory
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+commit-on-demand) in ELRANGE.
+- Ranges in ELRANGE not tracked by an EMA are considered free and ready for new allocations.
+- The EMM labels certain EMAs reserved for runtime or its internal usage and make them
+not accessible from public APIs.
+- A thread calling an EMM API on an EMA with an operation pending in another thread will wait
+until the pending operation is finished.
+
+**Assumptions:**
+
+- When enclave is loaded, the OS reserves the whole address range covered by ELRANGE.
```
dthaler, line 98:
```suggestion
- When an enclave is loaded, the OS reserves the whole address range covered by ELRANGE.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+It is assumed the host app will not remap any part of this reserved range.
+- When an enclave is loaded with base address at zero, only partial ELRANGE may be
```
dthaler, line 100:
```suggestion
- When an enclave is loaded with base address at zero, only a partial ELRANGE may be
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+  reserved by the OS. In that case, the EMM will assume the partial ELRANGE as valid reserved
```
dthaler, line 101:
```suggestion
  reserved by the OS. In that case, the EMM will assume the partial ELRANGE as a valid reserved
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+  range for use inside the enclave.
+- Memory manager does not check EPC pressure, or proactively trim pages when EPC runs low.
```
dthaler, line 103:
```suggestion
- The memory manager does not check EPC pressure, or proactively trim pages when EPC runs low.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+The OS can reclaim EPC pages when EPC running low or cgroups threshold reached
+- Memory manager does not maintain and recycle committed then freed pages
+  - Whenever a page is freed (via dealloc or uncommit API), it is trimmed from enclave
+  and need be re-allocated and committed before re-use.
```
dthaler, line 107:
```suggestion
  - Whenever a page is freed (via dealloc or uncommit API), it is trimmed from the enclave
  and needs to be re-allocated and committed before re-use.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+  - Owner of a region can re-purpose a sub-region of it by calling sgx_mm_modify_type/permissions
```
dthaler, line 108:
```suggestion
  - The owner of a region can re-purpose a sub-region of it by calling sgx_mm_modify_type/permissions
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+  to split out the sub-region to be reused.
+- Memory manager does not call back client for #GP handling. Memory manager code will ensure
```
dthaler, line 110:
```suggestion
- The memory manager does not call back into the client for #GP handling. Memory manager code will ensure that
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+itself would not cause #GP, and only register a #PF handler with the enclave global exception
+handler registry through the runtime abstraction layer. Clients wish to handle #GP can register
```
dthaler, line 112:
```suggestion
handler registry through the runtime abstraction layer. A client wishing to handle #GP can register
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+ * to explicitly EACCEPTCOPY data.
+ * This custom handler is passed into sgx_mm_alloc, and associated with the
+ * newly allocated region. The memory manager calls the handler when a #PF
+ * happens in the associated region. The handler may invoke abort() if it
+ * determines the exception is invalid based on certain internal states
+ * it maintains.
+ *
+ * @param[in] pfinfo info reported in the SSA MISC region for page fault.
+ * @param[in] private_data private data provided by handler in sgx_mm_alloc call.
+ * @retval SGX_EXCEPTION_CONTINUE_EXECUTION  Success on handling the exception.
```
dthaler, line 157:
```suggestion
 * @retval SGX_EXCEPTION_CONTINUE_EXECUTION Success on handling the exception.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+#define SGX_EMA_PROT_WRITE      0x2UL
+#define SGX_EMA_PROT_EXECUTE    0x4UL
+
+/*
+ * Allocate a new memory region in enclave address space (ELRANGE).
+ * @param[in] addr starting address of the region, page aligned. If NULL is provided,
+ *                  then the function will select the starting address.
+ * @param[in] length size of the region in multiples of page size in bytes.
+ * @param[in] flags a bitwise OR of flags describing committing mode, committing
+ * order, address preference, page type
```
dthaler, line 227:
```suggestion
 * order, address preference, and page type.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+int sgx_mm_alloc(void *addr, size_t length, int flags,
+                 sgx_enclave_fault_handler_t handler, void *handler_private,
+                 void **out_addr);
+
+```
+
+**Remarks:**
+- Permissions of newly allocated regions are always SGX_EMA_PROT_READ|SGX_EMA_PROT_WRITE and of page
+    type SGX_EMA_PAGE_TYPE_REG, except for SGX_EMA_RESERVE mode regions which will have SGX_EMA_PROT_NONE.
+- Once allocated by sgx_mm_alloc, a region will stay in allocated state and become
```
dthaler, line 274:
```suggestion
- Once allocated by sgx_mm_alloc, a region will stay in the allocated state and become
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+ * @retval 0 The operation was successful.
+ * @retval EINVAL The address range is not allocated or outside enclave.
+ */
+int sgx_mm_uncommit(void *addr, size_t length);
+
+/*
+ * Deallocate the address range.
+ * The pages in the allocation are freed and the address range is released for future allocation.
+ * @param[in] addr page aligned start address of the region to be freed and released
+ * @param[in] length size in bytes of multiples of page size.
```
dthaler, line 300:
```suggestion
 * @param[in] length size in bytes in multiples of page size.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+ * @retval 0 The operation was successful.
+ * @retval EINVAL The address range is not allocated or outside enclave.
+ */
+int sgx_mm_uncommit(void *addr, size_t length);
+
+/*
+ * Deallocate the address range.
+ * The pages in the allocation are freed and the address range is released for future allocation.
+ * @param[in] addr page aligned start address of the region to be freed and released
```
dthaler, line 299:
```suggestion
 * @param[in] addr page aligned start address of the region to be freed and released.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+ */
+int sgx_mm_dealloc(void *addr, size_t length);
+
+```
+
+### sgx_mm_modify_type, sgx_mm_modify_permissions and sgx_mm_modify_ex
+
+```
+/*
+ * Change permissions and/or page type of a previously allocated region
```
dthaler, line 312:
```suggestion
 * Change permissions and/or page type of a previously allocated region.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+/*
+ * Change permissions of an allocated region.
+ * Equivalent to sgx_mm_modify_ex(addr, length, -1, type).
+ */
+int sgx_mm_modify_type(void *addr, size_t length, int type);
+
+```
+**Remarks:**
+- The memory manager will track current permissions for each region, and can
+    determine whether new permissions require an OCall for EMODPR, e.g RW<->RX, RW->R.
```
dthaler, line 349:
```suggestion
    determine whether new permissions require an OCall for EMODPR, e.g., RW<->RX, RW->R.
```

docs/DesignDocs/SGX_EMM.md
dthaler: Some places in this doc have "of multiples" and others say "in multiples".  I am not sure if they are intended to be the same.  Does this mean that the length must be a multiple of page size, as opposed to being the number of multiples of page size?

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+ *                outside the enclave address range.
+ * @retval EPERM Any page in requested range is previously committed.
+ * @retval EPERM The target permissions are not allowed by OS security policy,
+ *                  e.g., SELinux rules.
+ */
+int sgx_mm_commit_data(void *addr, size_t length, uint8_t *data, int prot);
+
+```
+**Remarks:**
+- The memory manager decides whether OCalls are needed to request OS to make PTE
```
dthaler, line 398:
```suggestion
- The memory manager decides whether OCalls are needed to ask the OS to make Page Table Entry (PTE)
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+
+/* Return value used by the EMM #PF handler to indicate
+ *  to the dispatcher that it should stop searching and continue execution.
+ */
+#define SGX_EXCEPTION_CONTINUE_EXECUTION 0xFFFFFFFF
+
+/*
+ * The EMM page fault (#PF) handler
+ *
+ * @param[in] pfinfo info reported in the SSA MISC region for page fault
```
dthaler, line 423:
```suggestion
 * @param[in] pfinfo info reported in the SSA MISC region for page fault.
```
nit: Be consistent throughout this file in terms of whether param descriptions end in a period or not.

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+ * a #PF before all other handlers.
+ *
+ * @param[in] pfhandler the EMM page fault handler.
+ * @retval true Success.
+ * @retval false Failure.
+ */
+bool sgx_register_pfhandler(sgx_enclave_pfhandler_t pfhandler);
+
+/**
+ * Unregister the EMM handler with the global exception handler registry
```
dthaler, line 441:
```suggestion
 * Unregister the EMM handler with the global exception handler registry.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+ */
+bool sgx_unregister_pfhandler(sgx_enclave_pfhandler_t pfhandler);
+
+```
+
+### OCalls
+
+```
+/*
+ * Call OS to reserve region for EAUG, immediately or on-demand
```
dthaler, line 454:
```suggestion
 * Call OS to reserve region for EAUG, immediately or on-demand.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+ *
+ * @param[in] addr desired page aligned start address, NULL if no desired address
+ * @param[in] length size of the region in multiples of page size in bytes
+ * @param[in] flags  a bitwise OR of flags describing committing mode, committing
```
dthaler, line 458:
```suggestion
 * @param[in] flags a bitwise OR of flags describing committing mode, committing
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+ *                     order, address preference, page type. The untrusted side
+ *    implementation should always invoke mmap syscall with MAP_SHARED|MAP_FIXED_NOREPLACE, and
```
dthaler, line 460: nit: shouldn't these two lines have the same indentation?

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+ *             - SGX_EMA_PAGE_TYPE_SS_FIRST: the first page in shadow stack.
+ *             - SGX_EMA_PAGE_TYPE_SS_REST: the rest page in shadow stack.
+ * @retval 0 The operation was successful.
+ * @retval EINVAL Any parameter passed in is not valid.
+ * @retval errno Error as reported by dependent syscalls, e.g., mmap().
+ */
+int sgx_mmap_ocall(void *addr, size_t length, int prot, int flags);
+
+/*
+ * Call OS to change permissions, type, or notify EACCEPT done after TRIM
```
dthaler, line 489:
```suggestion
 * Call OS to change permissions, type, or notify EACCEPT done after TRIM.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+ * @param[in] length length of the area.  This must be a multiple of the page size.
+ * @param[in] flags this must be OR'ed of following:
+ *            SGX_EMA_PROT_READ
+ *            SGX_EMA_PROT_WRITE
+ *            SGX_EMA_PROT_EXEC
+ *            SGX_EMA_PAGE_TYPE_TRIM: change the page type to PT_TRIM, implies RW.
+ *            SGX_EMA_PAGE_TYPE_TCS: change the page type to PT_TCS
+ *            SGX_EMA_PROT_NOACCESS: Signal the kernel EACCEPT is done for PT_TRIM pages.
+ * @retval 0 The operation was successful.
+ * @retval EINVAL if any parameter passed in is not valid.
```
dthaler, line 501:
```suggestion
 * @retval EINVAL A parameter passed in is not valid.
```
retval generates a table where everything after the first token is in a separate column.  Use complete sentences line lines 500 and 502 do, no "if"

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+bool sgx_is_within_enclave(const void *ptr, size_t size);
+
+```
+
+### Support for EMM Initialization
+
+In addition to implement the abstraction layer APIs, a runtime shall provide
+iniitial enclave memory layout information to the EMM during early
+initialization phase of the enclave.
+The memory manager must be initialized in first ECALL (ECMD_INIT_ENCLAVE in
```
dthaler, line 556:
```suggestion
The memory manager must be initialized in the first ECALL (ECMD_INIT_ENCLAVE in
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+Intel SGX SDK) before any other clients can use it. Therefore, code and data
```
dthaler, line 557: When this document says "Intel SGX SDK" does it refer to both the Intel SGX SDK for Linux and the Intel SGX SDK for Windows, since those are different SDKs?  Clarify.  Avoid ever using the term "Intel SGX SDK" without "for ..." since the term is ambiguous/confusing.

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+of the memory manager will be part of initial enclave image that are loaded
+with EADD before EINIT, and as a part of the trusted runtime.
+
+The trusted runtime should enumerate all initial committed regions (code,
+data, heap, stack, TCS, SSA), and call the EMM internal APIs to setup
```
dthaler, line 562:
```suggestion
data, heap, stack, TCS, and SSA), and call the EMM internal APIs to set up
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+initial entries in the EMA list to track existing regions and mark some
+of them not modifiable by EMM public APIs. Runtime also ensures there is
```
dthaler, line 564:
```suggestion
of them as not modifiable by EMM public APIs. The runtime also ensures there is
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+enough reserved space on heap for EMM to create the initial EMA list and
```
dthaler, line 565:
```suggestion
enough reserved space on the heap for EMM to create the initial EMA list and
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+the entries. Once initialized, the memory manager can reserve its own
+space for future expansion of the EMA list, and special EMAs to hold
+EMA objects. To keep it simple, the expansion can be done eagerly: commit
+more pages for EMA list once unused committed space in the EMA List Region
+below certain threshold.
+
+Alternative option: At build time, the enclave signing tool can precalculate
+and fill in EMA entries that holds info of initial regions to be committed by
```
dthaler, line 573:
```suggestion
and fill in EMA entries that hold info on initial regions to be committed by
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+EADD during enclave load. The calculated start addresses in these EMAs can be
+relative to enclave secs->base. Runtime can patch those entries at
```
dthaler, line 575:
```suggestion
relative to enclave secs->base. The runtime can patch those entries at
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+int ema_commit_data(void *addr, size_t length, uint8_t *data, int prot);
+
+int ema_modify_ex(void *addr, size_t length, int prot, int type);
+
+```
+
+Internal APIs and Structures
+-------------------------------------
+
+Following are internal functions and structures to be used by EMM implementation.
```
dthaler, line 615:
```suggestion
The following are internal functions and structures to be used by the EMM implementation.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+They can evolve over time, shown here for reference only.
```
dthaler, line 616:
```suggestion
They can evolve over time, and are shown here for reference only.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+    sgx_enclave_fault_handler_t
+                        h;              // custom PF handler  (for EACCEPTCOPY use)
+    void*               hprivate;       // private data for handler
+    _ema_t*             next;           // next in doubly linked list
+    _ema_t*             prev;           // prev in doubly linked list
+} ema_t;
+
+```
+ **Remarks:**
+ - Access to the list (find, insert, remove EMAs) are synchronized for thread-safety.
```
dthaler, line 644:
```suggestion
 - Accesses to the list (find, insert, remove EMAs) are synchronized for thread-safety.
```

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+//EACCEPTCOPY
+int do_eacceptcopy(const sec_info_t* si, size_t dest, size_t src);
+
+```
+
+
+Metadata,  File format
+---------------------------------------
+
+Enclave metadata and file format are runtime specific. Detailed design is
```
dthaler, line 670:
```suggestion
The enclave metadata and file format are runtime specific. A detailed design is
```

docs/DesignDocs/SGX_EMM.md
haitaohuang: length should be in bytes but the number value should be multiples of page size, k*4096. So is `of multiples of page size` better?

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+layer for the runtimes. The runtime abstraction layer APIs are invoked by the EMM
+to make necessary OCalls, register and receive callbacks on page faults. The EMM
+source code will be hosted and maintained in the [Intel SGX PSW and SDK repository](https://github.com/intel/linux-sgx)
+and can be built as a separate library to be linked into any runtime with the
+abstraction layer APIs implemented.
+
+## Notes on Internal Design ##
+
+The enclave memory manager keeps track of memory allocation and layout info inside
+enclave address range (ELRANGE) using an internal structure called the Enclave Memory
```
mingweishih, line 85: Does this mean the enclave need to over request the ELRANGE during the creation?
Something like (where `ADDITIONAL REGION` is for the enclave to dynamically allocate via EMM):
```
[ENCLAVE LAYOUT][ADDITIONAL REGION]
<----          ELRANGE        ---->
```

And when 0-base enclave is enabled, the layout becomes

```
[REGION from ADDR 0][ENCLAVE LAYOUT][ADDITIONAL REGION]
<----                    ELRANGE                  ---->
```
I assume the EMM internally will ensure that `REGION from ADDR 0` should not be used?

haitaohuang: Runtime can return false from sgx_is_within_enclave or ema_alloc a RESERVED region for REGION from ADDR 0 with EMA_SYSTEM flag so no one else can allocate from it.

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,678 @@
...
+ * causes arithmetic operations to wrap.
+ *
+ */
+bool sgx_is_within_enclave(const void *ptr, size_t size);
+
+```
+
+### Support for EMM Initialization
+
+In addition to implement the abstraction layer APIs, a runtime shall provide
```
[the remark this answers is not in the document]
haitaohuang: The initial idea was for runtime to call ema_alloc with EMA_SYSTEM flag to set up those initial layouts. And other private APIs below can be used to adjust attributes and boundaries of initially committed pages

Now I am also thinking to add one more API specific for EADDed regions, some thing like this:
```
/*
 * Initialize an EMA. This can be used to setup EMAs to account regions that
 * are loaded and initialized with EADD before EINIT.  
 * @param[in] addr Starting address of the region, page aligned. If NULL is provided,
 *                  then the function will select the starting address.
 * @param[in] length Size of the region in multiples of page size in bytes.
 * @param[in] flags SGX_EMA_SYSTEM, or SGX_EMA_SYSTEM | SGX_EMA_RESERVE
 *           bitwise ORed with one of following page types:
 *             - SGX_EMA_PAGE_TYPE_REG: regular page type. This is the default if not specified.
 *             - SGX_EMA_PAGE_TYPE_TCS: TCS page.
 *             - SGX_EMA_PAGE_TYPE_SS_FIRST: the first page in shadow stack.
 *             - SGX_EMA_PAGE_TYPE_SS_REST: the rest page in shadow stack.
 * @param[in] prot permissions, either SGX_EMA_PROT_NONE or a bitwise OR of following with:
 *        - SGX_EMA_PROT_READ: Pages may be read.
 *        - SGX_EMA_PROT_WRITE: Pages may be written.
 *        - SGX_EMA_PROT_EXECUTE: Pages may be executed.
 * @param[in] handler A custom handler for page faults in this region, NULL if
 *                     no custom handling needed.
 * @param[in] handler_private Private data for the @handler, which will be passed
 *                     back when the handler is called.
 * @retval 0 The operation was successful.
 * @retval EACCES Region is outside enclave address space.
 * @retval EEXIST Any page in range requested is in use.
 * @retval EINVAL Invalid page type, flags, or addr and length are not page aligned.
 */ 
int ema_init(void *addr, size_t length, int flags, int prot,
              sgx_enclave_fault_handler_t handler,
              void *handler_private);
```

docs/DesignDocs/SGX_EMM.md
haitaohuang: will move up the abstraction layer section to give more context. thx

docs/DesignDocs/SGX_EMM.md
dthaler:
```suggestion
 * @param[in] length size in bytes.  The length must be a multiple of the page size.
```

dthaler (approved):
Understood that this document might move to an Intel repo and just be referenced from an OE document, since this is info about SGX not about the OE SDK per se.

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,712 @@
...
+```
+
+### Other Utilities
+
+```
+/*
+ * Define a mutex and create/lock/unlock/destroy functions.
+ */
+typedef struct _sgx_mm_mutex sgx_mm_mutex;
+sgx_mm_mutex *sgx_mm_mutex_create(void);
```
mingweishih, line 522: Does this assume creating a non-recursive mutex?

docs/DesignDocs/SGX_EMM.md
```diff
@@ -0,0 +1,712 @@
...
+/*
+ * Initialize the EMM internals and reserve the whole range available for user
+ * allocations via the public sgx_mm_alloc API. This should be called before
+ * any other APIs invoked. The runtime should not intend to allocate any subregion
+ * in [user_start, user_end) for system usage, i.e., the EMM will fail any allocation
+ * request with SGX_EMA_SYSTEM flag in this range and return an EINVAL error.
+ * @param[in] user_start The start of the user address range, page aligned.
+ * @param[in] user_end The end (exclusive) of the user address range, page aligned.
+ */
+void sgx_mm_init(size_t user_start, size_t user_end);
```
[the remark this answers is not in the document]
haitaohuang: This function only try to reserve an initial area (16 pages) for internal use and will call abort() if fails for some reason, e.g. there is not enough in [user_start, user_end).  Was the hang on this API or subsequent calls to other APIs?

haitaohuang:
Updated doc to reflect changes for ocalls to align with the new functions specified in Enclave Common, which are used for OCall implementation.
Also added note for runtime implemented mutex to be recursive.
Note as discussed in OE arch meeting, we moved EMM code to separate SGX EMM repo: https://github.com/intel/sgx-emm
Though review to continue on this PR and I will keep sync changes made in the EMM repo to this PR for review purpose, we won't need to merge this PR into OE.

docs/DesignDocs/SGX_EMM.md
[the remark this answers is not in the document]
haitaohuang: @mingweishih I will change to return ENOMEM if internal allocation failed for initial area. Let me know if more concerns.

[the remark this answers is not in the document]
mingweishih: Sounds good to me. Can you update this document as well?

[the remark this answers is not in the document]
haitaohuang: updated now.

mingweishih (approved):

haitaohuang:
Close without merge as commented earlier.
If any concerns in future, please open issues in https://github.com/intel/sgx-emm

PR #3994, opened 2021-05-04
Add OpenSSL path to Windows

rs--:


scripts/install-windows-prereqs.ps1
```diff
@@ -368,6 +368,7 @@ function Install-Git {
 function Install-OpenSSL {
     $installDir = Join-Path $InstallPath "OpenSSL"
     nuget.exe install openssl -Source $PACKAGES_DIRECTORY -OutputDirectory $InstallPath -ExcludeVersion
+    Add-ToSystemPath -Path @("$installDir\bin")
```
dthaler, line 371: I notice https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/install_oe_sdk-Windows.md doesn't say anything about installing openssl.  Should it?

rs--: I'm wondering if it was just missed in that doc that you linked, since OpenSSL is mentioned in this one: https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Contributors/WindowsManualInstallPrereqs.md#software-prerequisites

dthaler: Seems so, but even the doc you linked says nothing about adding it to the system path.  So if it needs to be in the system path as this PR implies, it seems like that should be mentioned in all three places as part of this PR.

BRMcLaren (approved):

BRMcLaren:
Thanks for catching this!

mingweishih (approved):

mingweishih:
Can you squash the comments before doing r+

dthaler (approved):

PR #3995, opened 2021-05-04
Replace unsecure string operation in oecert

qiucwang:
`strcpy` and `strcat` are not secure. Replaced it with snprintf to hardcode enclave filename.

### DIFF
--- a/tools/oecert/host/host.cpp
+++ b/tools/oecert/host/host.cpp
@@ -236,6 +236,7 @@ static char* _get_enclave_filename()
     char* enclave_filename = nullptr;
     char path[OE_PATH_MAX];
     size_t path_size = 0;
+    size_t enclave_filename_size = 0;
 
 #if defined(__linux__)
     path_size += (size_t)readlink("/proc/self/exe", path, OE_PATH_MAX);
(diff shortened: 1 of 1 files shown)


yentsanglee (approved):

tools/oecert/host/host.cpp
```diff
@@ -250,18 +251,20 @@ static char* _get_enclave_filename()
...
 
     if (enclave_filename == nullptr)
     {
         printf("Failed to malloc enclave_filename.\n");
         goto done;
     }
-
-    strcpy(enclave_filename, path);
-    strcat(enclave_filename, ENCLAVE_FILENAME_SUFFIX);
+    snprintf(
```
dthaler, line 262: This is still not safe, since `snprintf()` does not guarantee null termination.  
See https://en.cppreference.com/w/c/io/fprintf and https://stackoverflow.com/questions/1270387/are-snprintf-and-friends-safe-to-use
`snprintf_s()` on the other hand does safe truncation.

tools/oecert/host/evidence.cpp
```diff
@@ -131,30 +131,30 @@ void parse_certificate_extension(const uint8_t* data, size_t data_len)
     printf(
         "\n    parsed qe certificate extension (%s) {\n",
         SGX_EXTENSION_OID_STR);
-    printf("        ppid (hex): ");
+    printf("      ppid (hex): ");
```
dthaler, line 134: All the changes in this file look unrelated to the description of this PR.
https://github.com/openenclave/openenclave/blob/master/docs/Community/Contributing.md says:

> * **DON'T** mix independent, unrelated changes in one PR. Separate real project/test code changes from larger code formatting/dead code removal changes. Separate unrelated fixes into separate PRs, especially if they are in different libraries.

tools/oecert/host/host.cpp
qiucwang: 👍

tools/oecert/host/evidence.cpp
qiucwang: I'll put them in another PR.

tools/oecert/host/host.cpp
qiucwang: I checked again and `snprintf()` guarantees null termination. Microsoft's `_snprintf()` does not though.
https://stackoverflow.com/questions/7706936/is-snprintf-always-null-terminating
I also tested it on both linux and windows by truncating the buffer size, and the null terminator was included at both platforms.

qiucwang:
bors try

dthaler (approved):

PR #4001, opened 2021-05-07
Change output directory of oecert

ryanhsu19:
Change the output directory of `oecert` (since it has been moved from `tests` to `tools`) so that

1. `oecert` and `oecert_enc.signed` will show up in the same directory, `build/output/bin`, after doing `make`,
2. `oecert` and `oecert_enc.signed` will show up in `/opt/openenclave/bin` after doing `make install` (done in #3947 already), and
3. `build/tools/oecert` will only contain CMake files and Makefiles just like the other tools.


mingweishih:
bors try

yentsanglee (approved):

mingweishih (approved):

PR #4010, opened 2021-05-11
Fix Windows image builds

CyanDevs:
This fixes 2 issues:

1. When using a single node to build images in parallel, the windows-agents group in the Ansible host file will be overwritten every run or rerun.
2. Windows updates get stuck in a restart loop when the update item fails


mingweishih (approved):

rs-- (approved):

CyanDevs:
bors r+

Last run failed due to 1 node losing connection with master. Interestingly, relaunching the agent on the node does not resume the job even though max build durability should try to recover the build progress. Need to look into this more.

PR #4019, opened 2021-05-13
Rename oecert tool

ryanhsu19:
1. Propose `oegenerate` as the new name for the `oecert` tool (`oegen` was the original proposal but we think that "generate" sounds more complete than "gen"), while the other similar tool, `oecertdump`, is still `oecertdump`.
2. Fix some typos (rename `sgx_asem_env` to `sgx_aesm_env`) in `tools/oegenerate/host/host.cpp`.
3. Fix #4012.

Signed-off-by: Ryan Hsu <ryhsu@microsoft.com>

PR #4020, opened 2021-05-14
Adding an option to disable building oecert tool

mrragava:
oecert executable (normal) tool cannot link fuzzing enabled libraries, When building OE with ENABLE_FUZZING turned ON, OE libraries will be linked with fuzzing libraries. Executables has to be built as libfuzzer target to link OE fuzzing libraries.

Since oecert tool is not required for fuzzing, Added a flag to disable building oecert.

Signed-off-by: Ragavan Dasarathan <mrragava@gmail.com>


mingweishih:
Can you wait for #4019? The tool has been renamed,

PR #4029, opened 2021-05-19
Adding '/z7' flag to srcs in openenclave/host/ to incude debug symbols to OE SDK objects

RRathna:

### DIFF
--- a/host/CMakeLists.txt
+++ b/host/CMakeLists.txt
@@ -342,6 +342,21 @@ list(
   strings.c
   traceh_enclave.c)
 
+# Add MSVC compile flag /Z7 to enable debugging into OE SDK source from
+# applications that use release builds.
+if (OE_SGX AND WIN32)
+  foreach (file ${PLATFORM_HOST_ONLY_SRC})
+    if (${file} MATCHES "\.c$")
+      set_source_files_properties(${file} PROPERTIES COMPILE_FLAGS "/Z7")
+    endif ()
+  endforeach ()
+  foreach (file ${PLATFORM_SDK_ONLY_SRC})
+    if (${file} MATCHES "\.c$")
+      set_source_files_properties(${file} PROPERTIES COMPILE_FLAGS "/Z7")
+    endif ()
+  endforeach ()
+endif ()
+
 # Combine the following common code along with the platform specific code and
 # host verification code to get the full oehost target provided by the OE SDK.
 add_library(oehost STATIC ${PLATFORM_HOST_ONLY_SRC} ${PLATFORM_SDK_ONLY_SRC})


host/CMakeLists.txt
```diff
@@ -342,6 +342,21 @@ list(
   strings.c
   traceh_enclave.c)
 
+# Add MSVC compile flag /Z7 to enable debugging into OE SDK source from
```
mingweishih, line 345: Instead of doing this, you can just use `target_compile_options` against the `oehost` target.  See an example at: https://github.com/openenclave/openenclave/blob/e8f54a6b5c146eca81eb8c740f645674c6ec6a36/host/CMakeLists.txt#L485

Also, with these, we don't need the special handling for calls.c as it's part of `PLATFORM_SDK_ONLY_SRC`.

mingweishih (commented):
To be consistent with build type, should we set the flag only for Debug and RelWithDebInfo builds, not Release build?
On Linux, `-g` option is only added for those two types.

@anakrish thoughts?

host/CMakeLists.txt
RRathna: When I added /z7 to oehost, it was throwing invalid compile flag error for sgx/windows/aep.asm. Hence had to add flag to all .c srcs.

host/CMakeLists.txt
```diff
@@ -342,6 +342,21 @@ list(
...
+# applications that use release builds.
+if (OE_SGX AND WIN32)
+  foreach (file ${PLATFORM_HOST_ONLY_SRC})
```
anakrish, line 348: I'd prefer adding it to the target as Ming-Wei suggested instead of for each file.

Can you find out what compile options cmake adds by default for various  OE SDK build types (Release, RelWithDebInfo, Debug) on Windows? We need to make sure that /Z7 takes precedence over other debug flags cmake adds.

host/CMakeLists.txt
RRathna: Removed special assignment of \z7 flag to host/sgx/calls.c

RRathna:
There were several drawbacks observed when adding /z7 flag to all the sources in the host/sgx/.
1. The windows Debug and RelWithDebInfo package sizes grew almost 4 times.
2. The size of the applications built in Debug and RelWithDebInfo mode increases.
3. The /z7 flag introduces a dependency that the openenclave sources be placed in the exact path as that of the build machine even when a distribution of the same is being used in a different machine. This inconveniences the users greatly and makes setting up the environment for debugging more difficult.

Hence, it was decided to package the .pdb files that are generated as part of CMakeFiles/ during the build process (.pdbs are stored in PDB_OUTPUT_DIRECTORY only for shared libraries and exes, and not for static libraries). This is addressed in #4052

PR #4039, opened 2021-05-20
Create LVI wrapper for current and previous Clang versions

rs--:


mingweishih (approved):

CyanDevs (approved):
This approach is straightforward and it is possible to achieve the same outcome with less tasks and with loops and that could be an optimization to do in the future.

manojrupireddy (approved):

anakrish (approved):

PR #4040, opened 2021-05-20
increase heap size/stack size to fix snmalloc errors

manojrupireddy:
increase heap size/stack size to fix snmalloc errors

### DIFF
--- a/tests/tls_e2e/client_enc/mbedtls_client.cpp
+++ b/tests/tls_e2e/client_enc/mbedtls_client.cpp
@@ -348,6 +348,6 @@ OE_SET_ENCLAVE_SGX(
     1,    /* ProductID */
     1,    /* SecurityVersion */
     true, /* Debug */
-    512,  /* NumHeapPages */
-    128,  /* NumStackPages */
+    1024, /* NumHeapPages */
+    512,  /* NumStackPages */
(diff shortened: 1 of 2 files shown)


tests/tls_e2e/client_enc/mbedtls_client.cpp
```diff
@@ -348,6 +348,6 @@ OE_SET_ENCLAVE_SGX(
     1,    /* ProductID */
     1,    /* SecurityVersion */
     true, /* Debug */
-    512,  /* NumHeapPages */
-    128,  /* NumStackPages */
+    1024, /* NumHeapPages */
+    512,  /* NumStackPages */
```
mingweishih, line 352: Do we also need to increase the stack size?

manojrupireddy: the errors are due to Heapsize. Hence increasing heap size alone will fix the issue. The stack pages in sample applications are in the range of 1024. Hence increased it from 128 to 512 to keep it in the same order.

radhikaj (approved):
Thanks @manojrupireddy

anakrish (approved):

CyanDevs:
bors single on

PR #4068, opened 2021-06-03
[Mystikos] Support the CapturePFGPExceptions preference

mingweishih:
This PR does the following to the mystikos.v4 branch
- Cherry pick #3984 to the mystikos branch.
- Address the problem with exception simulation (https://github.com/deislabs/mystikos/issues/399)
- Update the pf_gp_exception test
- Fix a bug in `OE_ENCLAVE_FLAG_DEBUG_AUTO`

### DIFF
--- a/enclave/core/sgx/exception.c
+++ b/enclave/core/sgx/exception.c
@@ -3,16 +3,17 @@
 
 #include <openenclave/bits/sgx/sgxtypes.h>
 #include <openenclave/enclave.h>
-#include <openenclave/internal/cpuid.h>
-#include <openenclave/internal/sgx/td.h>
-#include <openenclave/internal/thread.h>
 #include <openenclave/internal/calls.h>
 #include <openenclave/internal/constants_x64.h>
+#include <openenclave/internal/cpuid.h>
 #include <openenclave/internal/fault.h>
 #include <openenclave/internal/globals.h>
 #include <openenclave/internal/jump.h>
 #include <openenclave/internal/safecrt.h>
+#include <openenclave/internal/sgx/td.h>
+#include <openenclave/internal/thread.h>
 #include <openenclave/internal/trace.h>
+#include "../tracee.h"
 #include "asmdefs.h"
 #include "context.h"
 #include "cpuid.h"
@@ -318,6 +319,8 @@ void oe_real_exception_dispatcher(oe_context_t* oe_context)
     oe_exception_record.code = td->exception_code;
     oe_exception_record.flags = td->exception_flags;
     oe_exception_record.address = td->exception_address;
+    oe_exception_record.faulting_address = td->faulting_address;
+    oe_exception_record.error_code = td->error_code;
     oe_exception_record.context = oe_context;
 
     // Refer to oe_enter in host/sgx/enter.c. The contract we defined for EENTER
@@ -340,6 +343,13 @@ void oe_real_exception_dispatcher(oe_context_t* oe_context)
             break;
         }
     }
+
+    // Clear information after all the handlers are done
+    td->exception_code = 0;
+    td->exception_flags = 0;
+    td->exception_address = 0;
+    td->faulting_address = 0;
+    td->error_code = 0;
 
     // Jump to the point where oe_context refers to and continue.
     if (handler_ret == OE_EXCEPTION_CONTINUE_EXECUTION)
@@ -385,20 +395,24 @@ void oe_virtual_exception_dispatcher(
         *arg_out = OE_EXCEPTION_CONTINUE_SEARCH;
         return;
     }
-    sgx_ssa_gpr_t* ssa_gpr =
-        (sgx_ssa_gpr_t*)(((uint8_t*)ssa_info.base_address) + ssa_info.frame_byte_size - OE_SGX_GPR_BYTE_SIZE);
+    uint64_t gprsgx_offset = (uint64_t)ssa_info.base_address +
+                             ssa_info.frame_byte_size - OE_SGX_GPR_BYTE_SIZE;
+    sgx_ssa_gpr_t* ssa_gpr = (sgx_ssa_gpr_t*)gprsgx_offset;
+    bool is_exit_info_valid = ssa_gpr->exit_info.as_fields.valid;
 
-    if (!ssa_gpr->exit_info.as_fields.valid)
+    if (!is_exit_info_valid)
     {
-        // Consider manually provided exception information for SGX1
-        if (oe_exception_record)
+        // Only allow the simulated #PF in debug mode
+        if (is_enclave_debug_allowed() && oe_exception_record)
         {
             td->exception_address = ssa_gpr->rip;
 
-            /* TODO PRP: We need to read the exception information from
-             * oe_exception_record instead */
             td->exception_code = OE_EXCEPTION_PAGE_FAULT;
             td->exception_flags |= OE_EXCEPTION_FLAGS_HARDWARE;
+
+            /* We expect the host to pass in the faulting address of #PF */
+            td->faulting_address = oe_exception_record->faulting_address;
+            td->error_code = OE_SGX_PAGE_FAULT_US_FLAG;
         }
         else
         {
@@ -446,6 +460,17 @@ void oe_virtual_exception_dispatcher(
     }
     else
     {
+        /* The following codes can only be captured with SGX2 and the
+         * MISCSELECT[0] bit is set to 1. */
+        if (is_exit_info_valid &&
+            (td->exception_code == OE_EXCEPTION_PAGE_FAULT ||
+             td->exception_code == OE_EXCEPTION_ACCESS_VIOLATION))
+        {
+            sgx_exinfo_t* exinfo =
+                (sgx_exinfo_t*)(gprsgx_offset - OE_SGX_MISC_BYTE_SIZE);
+            td->faulting_address = exinfo->maddr;
+            td->error_code = exinfo->errcd;
+        }
         // Modify the ssa_gpr so that e_resume will go to second pass exception
         // handler.
         ssa_gpr->rip = (uint64_t)oe_exception_dispatcher;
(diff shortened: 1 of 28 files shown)


enclave/core/sgx/exception.c
```diff
@@ -446,6 +457,17 @@ void oe_virtual_exception_dispatcher(
     }
     else
     {
+        /* The following codes can only be captured with SGX2 and the
+         * MISCSELECT[0] bit is set to 1. */
+        if (is_exit_info_valid &&
```
jxyang, line 462: Where do we check `MISCSELECT[0] == 1`?

tests/pf_gp_exceptions/host/host.c
```diff
@@ -0,0 +1,67 @@
...
+static bool _is_misc_region_supported()
+{
+    uint32_t eax, ebx, ecx, edx;
+    eax = ebx = ecx = edx = 0;
+
+    // Obtain feature information using CPUID
+    oe_get_cpuid(CPUID_SGX_LEAF, 0x0, &eax, &ebx, &ecx, &edx);
+
+    // Check if EXINFO is supported by the processor
+    if (!(ebx & CPUID_SGX_MISC_EXINFO_MASK))
```
jxyang, line 22: Just use `return (ebx & CPUID_SGX_MISC_EXINFO_MASK);`

tests/pf_gp_exceptions/enc/enc.c
```diff
@@ -0,0 +1,79 @@
...
+
+uint64_t test_pfgp_handler(oe_exception_record_t* exception_record)
+{
+    if (exception_record->code == OE_EXCEPTION_PAGE_FAULT)
+    {
+        const int MOV_INSTRUCTION_BYTES = 3;
+        faulting_address = exception_record->faulting_address;
+        error_code = exception_record->error_code;
+        exception_record->context->rip += MOV_INSTRUCTION_BYTES;
+        oe_host_printf(
```
jxyang, line 20: It's unsafe to print any message inside an exception handler.

If we really need it, save the variables and print them after the control is transferred from the exception handler to the regular code.

tests/pf_gp_exceptions/enc/enc.c
```diff
@@ -0,0 +1,79 @@
...
+    /* Trigger #PF */
+    faulting_address = 1;
+    asm volatile("xor %rax, %rax\n\t"
+                 "mov (%rax), %rax");
+    OE_TEST(faulting_address == 0);
+
+    /* Trigger #GP */
+    faulting_address = 1;
+    asm volatile("lgdt 0x00");
+    OE_TEST(faulting_address == 0);
```
jxyang, line 64: Verify the error code?

tests/pf_gp_exceptions/enc/enc.c
```diff
@@ -0,0 +1,79 @@
...
+    if (result != OE_OK)
+    {
+        return -1;
+    }
+
+    /* Trigger #PF */
+    faulting_address = 1;
+    asm volatile("xor %rax, %rax\n\t"
+                 "mov (%rax), %rax");
+    OE_TEST(faulting_address == 0);
```
jxyang, line 59: Is it possible to test #PF triggered by page permission?

enclave/core/sgx/exception.c
jxyang: I was curious if this branch should only be activated when `MISCSELECT[0] bit is set to 1`, why that is not in the if-condition of the branch?

enclave/core/sgx/exception.c
```diff
@@ -385,10 +394,12 @@ void oe_virtual_exception_dispatcher(
...
     }
-    sgx_ssa_gpr_t* ssa_gpr =
-        (sgx_ssa_gpr_t*)(((uint8_t*)ssa_info.base_address) + ssa_info.frame_byte_size - OE_SGX_GPR_BYTE_SIZE);
+    uint64_t gprsgx_offset = (uint64_t)ssa_info.base_address +
+                             ssa_info.frame_byte_size - OE_SGX_GPR_BYTE_SIZE;
+    sgx_ssa_gpr_t* ssa_gpr = (sgx_ssa_gpr_t*)gprsgx_offset;
+    bool is_exit_info_valid = ssa_gpr->exit_info.as_fields.valid;
 
-    if (!ssa_gpr->exit_info.as_fields.valid)
+    if (!is_exit_info_valid)
```
jxyang, line 402: The following code is unsafe. We should strength the condition to: if (oe_exception_record && debug_mode)...
```c
// Consider manually provided exception information for SGX1
        if (oe_exception_record)
        {
            td->exception_address = ssa_gpr->rip;

            /* TODO PRP: We need to read the exception information from
             * oe_exception_record instead */
            td->exception_code = OE_EXCEPTION_PAGE_FAULT;
            td->exception_flags |= OE_EXCEPTION_FLAGS_HARDWARE;
        }
```
This is issue https://github.com/deislabs/mystikos/issues/399

mingweishih: Thanks for the catch. I actually have another PR pending for this fix.
https://github.com/mingweishih/openenclave/commit/453da6c18a967e961c1e3d728eeb27a0e38e2953

I can add the fix to this PR as well.

enclave/core/sgx/exception.c
mingweishih: ah, I misunderstood your question. When `MISCSELECT[0] bit is set to 1`, it is SGX hardware that decides to capture #PF/#GP; i.e., if we configure the enclave with `MISCSELECT[0] bit set to 0`, the #PF/#GP will never be delivered to the enclave. Therefore, we don't need additional software check.

tests/pf_gp_exceptions/enc/enc.c
mingweishih: Will update on this.

tests/pf_gp_exceptions/enc/enc.c
mingweishih: Will update this, thanks.

enclave/core/sgx/exception.c
mingweishih: Updated, please check.

tests/pf_gp_exceptions/enc/enc.c
mingweishih: Updated.

tests/pf_gp_exceptions/enc/enc.c
mingweishih: For simplicity, I keep the use of address 0, which is unmapped or inaccessible on both Linux and Windows by default.

tests/pf_gp_exceptions/enc/enc.c
mingweishih: removed the print

tests/pf_gp_exceptions/host/host.c
mingweishih: done

jxyang (approved):

enclave/core/sgx/exception.c
```diff
@@ -385,20 +395,28 @@ void oe_virtual_exception_dispatcher(
...
             td->exception_address = ssa_gpr->rip;
 
             /* TODO PRP: We need to read the exception information from
              * oe_exception_record instead */
             td->exception_code = OE_EXCEPTION_PAGE_FAULT;
             td->exception_flags |= OE_EXCEPTION_FLAGS_HARDWARE;
+
+            /* We cannot know the faulting address that triggers #PF on SGX1
+             * unless the host passes in the information. */
+            td->faulting_address = 0;
```
bodzhang, line 418: Why can't the host pass in the faulting_address info?

enclave/core/sgx/exception.c
```diff
@@ -385,20 +395,28 @@ void oe_virtual_exception_dispatcher(
...
+    if (!is_exit_info_valid)
     {
         // Consider manually provided exception information for SGX1
-        if (oe_exception_record)
+        // Only allow the simulated #PF in debug mode
+        if (is_enclave_debug_allowed() && oe_exception_record)
         {
             td->exception_address = ssa_gpr->rip;
 
             /* TODO PRP: We need to read the exception information from
```
bodzhang, line 411: This TODO comment is obsolete, with the design of "only allow the simulated #PF in debug mode".

bodzhang (changes requested):

enclave/core/sgx/exception.c
mingweishih: removed the comment

enclave/core/sgx/exception.c
mingweishih: Update to read the information passed in by the host. However, it seems the kernel does not allow the host to obtain the faulting address inside a user-mode signal handler; i.e., the address obtained from the signal handler on the host is always zero.

host/sgx/exception.c
```diff
@@ -46,8 +47,10 @@ uint64_t oe_host_handle_exception(oe_host_exception_context_t* context)
 
         oe_exception_record_t oe_exception_record = {0};
 
-        /* TODO PRP: We need to save the exception information in
-         * oe_exception_record */
+        // ATTN: the faulting address is expected to be read from the
```
bodzhang, line 50: In this case, the simulated #FP will not be very useful. I can't think of any realistic exception handler of #PF that does not need to check the faulting address.

mingweishih: Should we remove the simulation code instead?

enclave/core/sgx/exception.c
jxyang: The kernel does pass a `mcontext` to the signal handler. In there the general purpose register  [` REG_CR2`](https://code.woboq.org/userspace/glibc/sysdeps/unix/sysv/linux/x86/sys/ucontext.h.html#REG_CR2) contains the faulty address of the sigsegv. We need to add plumbing to pass in this address from OE's signal handler on the host side to the signal handler in the enclave.

host/sgx/exception.c
jxyang: According to Peter, reporting faulty address at 0 is good enough for Java runtime. I think Java runtime depends on a SIGSEGV event, but not necessarily the address, to reach a safe point. In this regard, the simulated #FP event is still useful.

enclave/core/sgx/exception.c
jxyang: We have the dependency chain of:

 Release mode enclave -> openenclave.v4 -> this PR.

I think we should get this in quickly to move on to the other two issues.

mingweishih: Just tested, the results is the same as `si_addr`.
I print the value of `REG_CR2` in the handler
```
    uint64_t faulting_address = (uint64_t)context->uc_mcontext.gregs[REG_CR2];
    printf("faulting address: 0x%lx\n", faulting_address);
```
and add an ocall to access address 0x8 at the host alone with the current two tests (access 0x8 inside the enclave and run `lgdt` instruction).

Here is the results
```
faulting address: 0x0
Test #PF passed
faulting address: 0x0
Test #GP passed
faulting address: 0x8
```

jxyang: We need to verify when the faulty address is protected, i.e., inside the enclave, the kernel still report the correct address in si_addr or CR2.

mingweishih: Does you mean accessing the enclave memory by the host? it does not trigger a #PF.
- Read -> get dummy value (-1)
- Write -> no effects

jxyang: I mean changing the page permission on an enclave address to, for example, read only, and then writing to that address in the enclave code. Fundamentally this is the absolute dominate use case for #PF events in Mystikos.

mingweishih: I can confirm that CR2/si_addr is not cleared if #PF is on an enclave memory.
I test with writing to an enclave code page.
```
enclave code addr: 0x7fc603003500
faulting address: 0x7fc603003000
```

mingweishih: Let me update the comments and test cases accordingly.

host/sgx/exception.c
bodzhang: Even for Java/.Net runtime code that handles #PF on NULL reference, reporting faulting address of 0 when the faulting address is non-zero might be problematic. For other code, such as stack overflow detection, reporting incorrect faulting address will render the code logic useless, or even cause unexpected result.

bodzhang (approved):

PR #4072, opened 2021-06-04
Remove the double call pattern for oe_get_supported_attester_format_ids_ocall

mingweishih:
Remove the double call pattern for oe_get_supported_attester_format_ids_ocall with the use of deep-copy feature.

### DIFF
--- a/include/openenclave/edl/sgx/attestation.edl
+++ b/include/openenclave/edl/sgx/attestation.edl
@@ -23,6 +23,12 @@ enclave
     // contiguous memory.
     include "openenclave/bits/sgx/sgxtypes.h"
 
+    struct format_ids_t
+    {
+        [size=size] void* data;
+        size_t size;
+    };
+
     trusted
     {
         public oe_result_t oe_get_sgx_report_ecall(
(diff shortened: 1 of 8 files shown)


qiucwang (approved):
LGTM. Thanks for the optimization.

include/openenclave/edl/sgx/attestation.edl
```diff
@@ -23,6 +23,12 @@ enclave
     // contiguous memory.
     include "openenclave/bits/sgx/sgxtypes.h"
 
+    struct format_ids_t
+    {
+        [size=size] void* data;
```
anakrish, line 28: Given deep-copy support, is it possible to make the ocall strongly typed? 
The caller and callee cast seem to use `oe_uuid_t` for data.

mingweishih: Yes, but doing so requires the oeedger8r-generated files include the `evidence.h` where `oe_uuid_t` is defined. Keeping the use of void seems easier.

dthaler: FWIW, I'd prefer including evidence.h but don't feel strongly

dthaler (approved):

PR #4075, opened 2021-06-07
Fix compatibility issues between oeutil and oeverify

ryanhsu19:
Remove header information generated by `oe_get_evidence()` and pass the desired evidence format into `oe_verify_evidence()` in order to fix the compatibility issues between `oeutil` and `oeverify`.

Also, since evidence in `SGX_EPID_LINKABLE` or `SGX_EPID_UNLINKABLE` format can only be verified by Intel Attestation Service (IAS), modify `oeutil` and `oeverify` so that evidence in either EPID format will not be verified.

### DIFF
--- a/tools/oeutil/enc/generate_evidence_enc.cpp
+++ b/tools/oeutil/enc/generate_evidence_enc.cpp
@@ -134,7 +134,7 @@ oe_result_t get_plugin_evidence(
 
     OE_CHECK(oe_get_evidence(
         &evidence_format,
-        OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID,
+        0,
         NULL,
         0,
         NULL,
--- a/tools/oeutil/host/generate_evidence.cpp
+++ b/tools/oeutil/host/generate_evidence.cpp
@@ -294,7 +294,7 @@ static void _display_help(const char* command)
     printf(
         "\t%s, %s: verify the generated remote attestation certificate, "
         "report, "
-        "or evidence.\n",
+        "or evidence in ECDSA format.\n",
         SHORT_INPUT_PARAM_OPTION_VERIFY,
         INPUT_PARAM_OPTION_VERIFY);
     printf(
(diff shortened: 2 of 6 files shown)


mingweishih (approved):

tools/oeutil/enc/generate_evidence_enc.cpp
```diff
@@ -134,15 +134,15 @@ oe_result_t get_plugin_evidence(
...
+        0,
         NULL,
         0,
         NULL,
         0,
         &local_evidence,
         &local_evidence_size,
-        endorsements ? &local_endorsements : NULL,
-        endorsements ? &local_endorsements_size : 0));
+        &local_endorsements,
```
yentsanglee, line 144: These 2 lines are intentional. Passing in non-null endorsement unnecessarily make calls to retrieve endorsements which can be time consuming. Please restore.

yentsanglee:
We want change host_verify sample too now that the evidence format is different.

yentsanglee (changes requested):
We want change host_verify sample too now that the evidence format is different.

dthaler (approved):
Looks ok except for Yen's comment, which I agree with.  Once that is addressed it will look ok to me.

yentsanglee (approved):

yentsanglee:
bors try

qiucwang:
For oeutil/host/genereate_evidence.cpp, need to update `dump_oe_evidence()` as well.
After update, you can call `oeutil gen` with `--verbose` flag for verification.

ryanhsu19:
bors try

ryanhsu19:
bors try

ryanhsu19:
bors try

tools/oeutil/host/generate_evidence.cpp
```diff
@@ -668,23 +668,14 @@ oe_result_t dump_oe_report(const uint8_t* report, size_t report_size)
 oe_result_t dump_oe_evidence(const uint8_t* evidence, size_t evidence_size)
 {
     oe_result_t result = OE_UNEXPECTED;
```
yentsanglee, line 670: Ignore formats that are not SGX_ECDSA.

ryanhsu19:
bors try

PR #4082, opened 2021-06-09
Add specific error codes for quote ex failure.

### LINKED ISSUE #3902: More clear error value from quote library loading failure
`oe_get_evidence()` should return more clear error values when:

- SGX_AESM_ADDR is set and quote-ex library is not installed
- SGX_AESM_ADDR is set and AESM libraries are not installed
- SGX_AESM_ADDR is not set and DCAP QuoteLibrary is not installed

Functions to check:

- oe_sgx_load_quote_ex_library
- _load_quote_ex_library_once

Be sure to add test case for these error values.

Additional ask:
- Ensure `get_quote3_error_t_string` covers all SGX errors defined in [_quote3_error_t](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/94e327a8235bc7c85353dc5e0a6dc1564e6c96c6/QuoteGeneration/quote_wrapper/common/inc/sgx_ql_lib_common.h#L47).

qiucwang:
Fixes #3902 
`OE_QUOTE_LIBRARY_LOAD_ERROR` when SGX_AESM_ADDR is set and quote-ex library is not installed.
`OE_SERVICE_UNAVAILABLE` when SGX_AESM_ADDR is set and AESM service is not available.
`OE_QUOTE_LIBRARY_ERROR` when SGX_AESM_ADDR is set and quote-ex API failed.
`OE_QUOTE_LIBRARY_LOAD_ERROR` when SGX_AESM_ADDR is not set and DCAP QuoteLibrary loading failed.
Add more detail in error messages.
Add helper `get_sgx_status_t_string()`, similiar to `get_quote3_error_t_string`, to cover all `sgx_status_t` status codes.

### DIFF
--- a/host/sgx/sgxquote.c
+++ b/host/sgx/sgxquote.c
@@ -32,7 +32,8 @@ OE_STATIC_ASSERT(sizeof(sgx_att_key_id_ext_t) == sizeof(sgx_att_key_id_t));
 #define SGX_QL_ALG_EPID_UNLINKABLE SGX_QL_ALG_EPID
 #define SGX_QL_ALG_EPID_LINKABLE SGX_QL_ALG_RESERVED_1
 
-// Facilitate writing switch statement for quote3_error_t values
+// Facilitate writing switch statement for quote3_error_t and sgx_status_t
+// values
 #define CASE_ERROR_RETURN_ERROR_STRING(x) \
     case x:                               \
         return #x;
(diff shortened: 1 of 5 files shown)


host/sgx/sgxquote.c
```diff
@@ -135,6 +136,188 @@ static quote3_error_t (*_sgx_qv_verify_quote)(
...
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_NETWORK_FAILURE);
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_SERVICE_TIMEOUT);
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_ERROR_BUSY);
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_UNKNOWN_MESSAGE_RESPONSE);
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_PERSISTENT_STORAGE_ERROR);
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_ERROR_MESSAGE_PARSING_ERROR);
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_PLATFORM_UNKNOWN);
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_QVEIDENTITY_MISMATCH);
+        CASE_ERROR_RETURN_ERROR_STRING(SGX_QL_QVE_OUT_OF_DATE);
+
```
yentsanglee, line 227: 3 more error cases need to be added. See [sgx_ql_lib_common.h](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/quote_wrapper/common/inc/sgx_ql_lib_common.h):

```
    SGX_QL_UNKNOWN_API_VERSION  = SGX_QL_MK_ERROR(0x0048),           /// The current PCS API version configured is unknown
    SGX_QL_CERTS_UNAVAILABLE  = SGX_QL_MK_ERROR(0x0049),             /// Certificates are not available for this platform

    SGX_QL_PSW_NOT_AVAILABLE = SGX_QL_MK_ERROR(0x0052),             ///< SGX PSW library cannot be loaded, could be due to file I/O error
```

Also need to update 3rdparty\sgxsdk\include\sgx_ql_lib_common.h

qiucwang: Updated.

yentsanglee (approved):

ryanhsu19 (approved):
Looks good. Thanks for the work!

include/openenclave/bits/result.h
```diff
@@ -372,6 +372,23 @@ typedef enum _oe_result
...
+    /**
+     * The SGX AESM service used for quote generation and attestation is not
+     * available.
+     */
+    OE_SGX_AESM_UNAVAILABLE,
+
+    /**
+     * The SGX quote-ex library API failed.
+     */
+    OE_SGX_QUOTE_EX_LIBRARY_ERROR,
```
dthaler, line 390: Do we expect that every TEE will add its own error codes?

host/sgx/sgxquote.c
```diff
@@ -135,6 +136,191 @@ static quote3_error_t (*_sgx_qv_verify_quote)(
 static void* _ql_module;
 static void* _qvl_module;
 
+// This is a helper for getting "human readable error messages" - as mentioned
+// in issue 2945 - so that we do not have to return complex quote3_error_t codes
```
dthaler, line 140: I don't think we should cite issue #'s except in TODO's.  Since 2945 is closed, it would be more appropriate to copy any relevant information into the comments here.

host/sgx/sgxquote.c
```diff
@@ -135,6 +136,191 @@ static quote3_error_t (*_sgx_qv_verify_quote)(
...
+// in error messages.
+static const char* get_quote3_error_t_string(quote3_error_t err)
```
dthaler, line 142:
```suggestion
static const char* get_quote3_error_t_string(quote3_error_t error)
```
avoid abbreviations, per OE style guidelines

host/sgx/sgxquote.c
```diff
@@ -135,6 +136,191 @@ static quote3_error_t (*_sgx_qv_verify_quote)(
...
+{
+    static char quote3_error_t_number[16];
+    sprintf_s(
+        quote3_error_t_number, sizeof(quote3_error_t_number), "0x%x", err);
```
dthaler, line 146: Since quote3_error_t_number is only used in the default case if I understand right, move this down into the default case

host/sgx/sgxquote.c
```diff
@@ -135,6 +136,191 @@ static quote3_error_t (*_sgx_qv_verify_quote)(
...
+        default:
+            return quote3_error_t_number;
+    }
+}
+
+// This is a helper for getting "human readable sgx status".
+static const char* get_sgx_status_t_string(sgx_status_t status)
+{
+    static char sgx_status_t_number[16];
+    sprintf_s(sgx_status_t_number, sizeof(sgx_status_t_number), "0x%x", status);
```
dthaler, line 241: move down into default case

host/sgx/sgxquote.c
qiucwang: Updated.

host/sgx/sgxquote.c
qiucwang: Make sense. Updated.

include/openenclave/bits/result.h
qiucwang: Good point. Will combine them with generic error codes and raise more clear error messages.

yentsanglee (approved):

qiucwang:
bors try

ryanhsu19 (approved):

include/openenclave/bits/result.h
```diff
@@ -372,6 +372,11 @@ typedef enum _oe_result
      */
     OE_QUOTE_LIBRARY_LOAD_ERROR,
 
+    /**
+     * The library API used for quote generation and attestation failed.
+     */
+    OE_QUOTE_LIBRARY_ERROR,
```
dthaler, line 378: The term "quote" is SGX specific and so the error should have SGX in it.
Either that or replace "quote" with evidence if this is intended to be a TEE-agnostic result code.

```suggestion
     * The library API used for SGX quote generation and attestation failed.
     */
    OE_SGX_QUOTE_LIBRARY_ERROR,
```

yentsanglee: @dthaler  "QUOTE" has been widely used in the errors. If we add "SGX" or replace with "evidence", it looks inconsistent among other already defined errors. I suggest we leave it as `OE_QUOTE_LIBRARY_ERROR` and open a separate issue to fix the naming inconsistence without breaking customer's code. For example,

```
typedef enum _oe_result
{
    ....
    OE_QUOTE_PROVIDER_LOAD_ERROR,
    OE_SGX_QUOTE_PROVIDER_LOAD_ERROR = OE_QUOTE_PROVIDER_LOAD_ERROR,

    OE_QUOTE_VERIFICATION_ERROR,
    OE_SGX_QUOTE_VERIFICATION_ERROR = OE_QUOTE_VERIFICATION_ERROR,

    OE_QUOTE_LIBRARY_LOAD_ERROR,
    OE_SGX_QUOTE_LIBRARY_LOAD_ERROR = OE_QUOTE_LIBRARY_LOAD_ERROR,

    ....
}
```

dthaler: Agree that existing errors are similarly bad.  I agree with aliasing the existing ones, though the alias should be in the other direction, i.e.,
```
    OE_SGX_QUOTE_PROVIDER_LOAD_ERROR,
    OE_QUOTE_PROVIDER_LOAD_ERROR = OE_SGX_QUOTE_PROVIDER_LOAD_ERROR,
```
This is so that the alias could for example be put inside an ifdef as a way to someday remove support.

But I would recommend not adding to the debt and doing new ones with SGX (or EVIDENCE) now so that we don't introduce new app code that would need to change in the next release.

dthaler (approved):
Thanks for the quick response!

qiucwang:
bors retry

PR #4091, opened 2021-06-11
fix clang10 speculative load hardening errors on windows

manojrupireddy:
Disable speculative load hardening for selected files when using clang10 compiler 

### DIFF
--- a/cmake/maybe_build_using_clangw.cmake
+++ b/cmake/maybe_build_using_clangw.cmake
@@ -48,8 +48,7 @@ function (maybe_build_using_clangw OE_TARGET)
       -fno-builtin-malloc
       -fno-builtin-calloc
       -fno-builtin
-      -mllvm
-      -x86-speculative-load-hardening)
+      -mspeculative-load-hardening)
   else ()
     target_compile_options(
       ${OE_TARGET}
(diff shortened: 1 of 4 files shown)


mingweishih (approved):

cmake/maybe_build_using_clangw.cmake
```diff
@@ -48,8 +48,7 @@ function (maybe_build_using_clangw OE_TARGET)
       -fno-builtin-malloc
       -fno-builtin-calloc
       -fno-builtin
-      -mllvm
-      -x86-speculative-load-hardening)
+      -mno-speculative-load-hardening)
```
radhikaj, line 51: This appears to be turning off all speculative load hardening on Windows.

manojrupireddy: fixed. Thank you

cmake/maybe_build_using_clangw.cmake
```diff
@@ -48,8 +48,7 @@ function (maybe_build_using_clangw OE_TARGET)
       -fno-builtin-malloc
       -fno-builtin-calloc
       -fno-builtin
-      -mllvm
-      -x86-speculative-load-hardening)
+      -mspeculative-load-hardening)
```
radhikaj, line 51: Can we keep this as -mllvm  -x86-speculative-load-hardening?  Will we break code otherwise? If both flags are equivalent can we keep the old one?

radhikaj (approved):
bors r+

PR #4093, opened 2021-06-15
Expose necessary libc symbols for SymCrypt FIPS module

mingweishih:
Expose the necessary symbols required by the SymCrypt FIPS module based on its final implementation.

### DIFF
--- a/tests/module_loading/module/module.c
+++ b/tests/module_loading/module/module.c
@@ -1,6 +1,10 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+#include <pthread.h>
+#include <stdlib.h>
+#include <string.h>
+
 extern int debugger_test;
 extern int is_module_init;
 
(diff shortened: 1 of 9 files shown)


mingweishih:
bors try

tests/module_loading/module/module.c
```diff
@@ -37,3 +39,21 @@ int add_with_constant(volatile int a, volatile int b)
         t = -1;
     return t;
 }
+
+#define TEST_SYMBOL(name)        \
+    do                           \
+    {                            \
+        void* ptr = (void*)name; \
```
anakrish, line 46: Does this need to be volatile so that the compiler doesn't optimize away the assignment?

mingweishih: done

mingweishih:
bors try

mingweishih:
bors try

radhikaj (approved):

mingweishih:
bors try

mingweishih:
bors try

tests/module_loading/module/module.c
```diff
@@ -37,3 +39,25 @@ int add_with_constant(volatile int a, volatile int b)
...
+#define TEST_SYMBOL(name)                 \
+    do                                    \
+    {                                     \
+        volatile void* ptr = (void*)name; \
+        (void)ptr;                        \
+    } while (0)
+
+void test_libc_symbols()
+{
+    TEST_SYMBOL(memcmp);
```
anakrish, line 52: The contract with symcrypt also has the constraint that `sizeof(oe_mutex_t) <= sizeof(GLIBC's pthread_mutex_t)`.
It would be good to lock this down here so that
- We don't regress in the future
- The contract is complete

anakrish (commented):
LGTM.
You'd also need to update commands.py which is the equivalent oelldb multi-module test.

tests/module_loading/module/module.c
mingweishih: Update the PR to use musl headers directly and checks the size of `pthread_mutex_t`

anakrish (approved):

mingweishih:
Can't reproduce the failure on simulation mode. Try again.

PR #4094, opened 2021-06-15
Align the internal data type definitions with musl

mingweishih:
This PR updates few internal data types that are different from the musl definitions, which could lead to binary incompatibility. Also, add tests to compare the size of internal data types against musl's.


mingweishih:
bors try

radhikaj (approved):

anakrish (approved):
Thanks for locking this down!

PR #4109, opened 2021-06-22
Added upgrade test for Ubuntu

rs--:


rs--:
bors try

.jenkins/pipelines/Azure/Linux/Jenkinsfile
```diff
@@ -18,21 +18,86 @@ AGENTS_LABELS = [
     "acc-ubuntu-18.04-vanilla": env.UBUNTU_VANILLA_1804_CUSTOM_LABEL ?: "vanilla-ubuntu-1804"
 ]
 
+def CmakeArgs(String build_type = "RelWithDebInfo", String code_coverage = "OFF", String debug_malloc = "ON", String lvi_args="", String cmake_args = "") {
+    def args = "-G Ninja -DCMAKE_INSTALL_PREFIX:PATH='/opt/openenclave' -DCPACK_GENERATOR=DEB -DCODE_COVERAGE=${code_coverage} -DUSE_DEBUG_MALLOC=${debug_malloc} -DCMAKE_BUILD_TYPE=${build_type} ${lvi_args} ${cmake_args} -Wdev"
+    return args
+}
+
+def TestCommand() {
```
CyanDevs, line 26: With the exception of CmakeArgs() the other functions `TestCommand(), InstallBuildCommand(), InstallReleaseCommand(), and TestSamplesCommand()` can be simplfied to be strings, which will save you a few lines to maintain as you won't need to assign the function to a variable to use in your shell script.

For example,

```
def TestCommand = """
    echo "Running Test Command"
"""
```
which is then referenced like so
```
def ACCUpgradeTest(String label, String compiler, String version, List extra_cmake_args = [], List test_env = []) {
    stage("${label} Container ${version} RelWithDebInfo, extra_cmake_args: ${extra_cmake_args}") {
        node("${label}") {
            timeout(GLOBAL_TIMEOUT_MINUTES) {
                cleanWs()
                checkout scm
                def cmakeArgs = CmakeArgs("RelWithDebInfo","OFF","ON","-DLVI_MITIGATION_BINDIR=/usr/local/lvi-mitigation/bin",extra_cmake_args.join(' '))
                def task = """
                           \$InstallReleaseCommand
                           \$TestSamplesCommand
                           cmake ${WORKSPACE} ${cmakeArgs}
                           ninja -v
                           \$InstallBuildCommand
                           \$TestSamplesCommand
                           """
                withEnv(test_env) {
                    oe.Run(compiler, task)
                }
            }
        }
    }
}
```

.jenkins/pipelines/Azure/Linux/Jenkinsfile
```diff
@@ -18,21 +18,86 @@ AGENTS_LABELS = [
...
+        ctest --output-on-failure --timeout ${CTEST_TIMEOUT_SECONDS}
+    """
+    return buildAndTest
+}
+
+def InstallBuildCommand() {
+    def localPackage = """
+        echo "Running Install Build Command"
+        cpack -G DEB
+        sudo killall -9 apt && sudo killall -9 dpkg
```
CyanDevs, line 38: When we manually test, killing the apt/dpkg processes are usually fine. But for automated tests we want to make sure we know that the state of our environment is stable. Killing apt/dpkg processes may introduce flakiness by leaving the system in a state that has different package versions as the apt process is killed at different points in time across different builds, and therefore different from builds with the exact same parameter.

I don't see any apt lock issues with our existing Linux builds, but if you encounter this and wait to proceed past this error you should look into waiting for the lock to be released.  For example, I implemented a function for Ansible installations [here in #3869](https://github.com/openenclave/openenclave/pull/3869/files).

.jenkins/pipelines/Azure/Linux/Jenkinsfile
```diff
@@ -18,21 +18,86 @@ AGENTS_LABELS = [
...
+        sudo ninja -v install
+    """
+    return localPackage
+}
+
+def InstallReleaseCommand() {
```
CyanDevs, line 44: This seems to only be used for the upgrade tests, so defining it outside of that function seems to be moot for now. Do you think this will be used by any other test types? If not, it would probably be cleaner to move this directly into upgrade test.

CyanDevs (changes requested):

.jenkins/pipelines/Azure/Linux/Jenkinsfile
rs--: I don't think it'll be used for anything else right now, but a secondary goal with moving the commands into functions was to clean up the test declaration. In my opinion it was easier to understand what the tests were trying to accomplish by reading the function calls instead of big blocks of text.

.jenkins/pipelines/Azure/Linux/Jenkinsfile
CyanDevs: Sorry the above example is not correct. \$variable references an environment variable. The correct way to reference a groovy variable here is:
```
                def example = """
                           \"${InstallReleaseCommand}\"
                           """
```

CyanDevs (approved):
Approved as this is tested and working, and so I don't block this while I am away. 

Eventually we could look into templating such as moving some of those common functions into a groovy library, and/or remove the need to call a function that returns a string.

radhikaj (approved):

PR #4121, opened 2021-06-24
Use default Intel SGX drivers when present

CyanDevs:
This change will add the following behaviour to the Ansible setups:
* If an Intel SGX driver is already present, then it will not install a new driver
* If an Intel SGX driver is not present on Ubuntu 20.04 then it will install Intel SGX driver 1.41. This is necessary as 1.36 will not work on Linux kernel 5.7+
* Changed the method to validate the presence of the driver. It will now use `modprobe` to load the intel_sgx driver, rather than rely on specific files to be present. As such, the variables `intel_sgx_directory` and `driver_validation_directories` are no longer needed, and were the only two variables present in `roles/linux/intel/vars/main.yml `

### DIFF
--- a/scripts/ansible/roles/linux/intel/tasks/driver-validation.yml
+++ b/scripts/ansible/roles/linux/intel/tasks/driver-validation.yml
@@ -10,9 +10,7 @@
   include_vars:
     file: "{{ ansible_distribution | lower }}/{{ ansible_distribution_release | lower }}.yml"
 
-- name: Check for existing required directories
-  stat:
-    path: "{{ item }}"
-  with_items: "{{ driver_validation_directories }}"
-  register: directory
-  failed_when: directory.stat.isdir == False
+- name: Load default driver
+  modprobe:
+    name: intel_sgx
+    state: present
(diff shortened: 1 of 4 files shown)


CyanDevs:
Please do not merge yet. I have some testing left to do.

mingweishih (commented):
Can you rebase? The PR includes some of recent changes.

radhikaj (approved):
Approving only the yaml files

anakrish:
> * If an Intel SGX driver is already present, then it will not install a new driver

Is it intentional to not address the case where the existing driver is an older version?

CyanDevs:
> > * If an Intel SGX driver is already present, then it will not install a new driver
> 
> Is it intentional to not address the case where the existing driver is an older version?

That is actually the preferred path. We are preferring to use an existing older version rather than install a newer version. The default driver version on ACC VMs is v1.33. If one is not present (e.g. non-ACC, or image without a default driver, etc) then one has to be installed and Ubuntu 20.04 with kernel 5.8 seems to only be able to install v1.41 (or higher).

anakrish (approved):

CyanDevs:
In the latest commit I moved the service_facts module out of the block so that it can gather host service info.

PR #4126, opened 2021-06-29
change speculative load hardening flag to x86-*

manojrupireddy:


manojrupireddy:
bors try

radhikaj (approved):
Thanks @manojrupireddy

mingweishih (commented):
Please ad add following comments as part of git commit, explains the rationale behind this change.

The -mspeculative-load-hardening compiler flag bundles the retpoline-indirect-calls option by default, which introduces non-negligible runtime overhead compared to using the LLVM speculative load hardening scheme alone. Therefore, we decide to fallback to the -mllvm -x86-speculative-load-hardening flag that was previous OE's default. Security-conscious developers can choose to opt in the -mretpoline or -mspeculative-load-hardening flags by themselves.

mingweishih (approved):

PR #4138, opened 2021-07-01
Attested TLS APIs proposal

mingweishih:


docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,476 @@
...
+of verification APIs (e.g., `oe_verify_attestation_certificate_with_evidence_v2`
+and `oe_verify_attestation_certificate`).
+
+- oe_get_attestation_certificate_with_evidence_v3
+  ```c
+  /**
+   * oe_get_attestation_certificate_with_evidence_v3
+   *
+   * Similar to oe_get_attestation_certificate_with_evidence_2, this function
+   * generates a self-signed x.509 certificate with embedded evidence generated by
```
dthaler, line 48:
```suggestion
   * generates a self-signed X.509 certificate with embedded evidence generated by
```

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,476 @@
...
+   * an attester plugin for the enclave, but it also allows a user to pass in
+   * runtime and inittime claims.
+   *
+   * @experimental
+   *
+   * @param[in] format_id The format id of the evidence to be generated.
+   *
+   * @param[in] subject_name a string containing an X.509 distinguished
+   * name (DN) for customizing the generated certificate. This name is also used
+   * as the issuer name because this is a self-signed certificate
```
dthaler, line 58:
```suggestion
   * as the issuer name because this is a self-signed certificate.
```

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,476 @@
...
+   * See RFC5280 (https://tools.ietf.org/html/rfc5280) for details
```
dthaler, line 59:
```suggestion
   * See RFC5280 (https://tools.ietf.org/html/rfc5280) for details.
```

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,476 @@
...
+   * Example value "CN=Open Enclave SDK,O=OESDK TLS,C=US"
+   *
+   * @param[in] private_key a private key used to sign this certificate
```
dthaler, line 62: nit: half the lines have the param description start with a capital letter, and have with a lower case letter.
Similarly, half end in a period and half don't.  Be consistent.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,476 @@
...
+   * @retval OE_INVALID_PARAMETER One or more invalid parameters.
+   * @retval OE_FAILURE General failure.
+   * @retval Other appropriate error code.
+   */
+  oe_result_t oe_verify_attestation_certificate_with_evidence_v3(
+      uint8_t* certificate_in_der,
+      size_t certificate_in_der_size,
+      uint8_t* endorsements_buffer,
+      size_t endorsements_buffer_size,
+      oe_policy_t* policies,
```
dthaler, line 136: oe_verify_evidence() makes all the [in] params be const, this should too

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,476 @@
...
+   *
+   * This function is similar to
+   * oe_verify_attestation_certificate_with_evidence_v2 that has the same
+   * prototype but with the ability to verify a ceritifacte generated by
+   * oe_get_attestation_certificate_with_evidence_v3.
+   *
+   * @experimental
+   *
+   * @param[in] certificate_in_der A pointer to buffer holding certificate
+   * contents in DER format.
```
dthaler, line 117: question: I know this is experimental, but do we really want separate APIs for certificates vs (say) JWTs vs CWTs?
oe_verify_evidence in contrast has a format_id to allow flexibility of format.   I might ask the same about the format of the endorsements_buffer.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,476 @@
...
+   * @param[in] optional_parameters The optional format-specific input parameters.
+   * @param[in] optional_parameters_size The size of optional_parameters in bytes.
+   *
+   * @param[out] output_certificate a pointer to buffer pointer
+   * @param[out] output_certificate_size size of the buffer above
+   *
+   * @return OE_OK on success
+   */
+  oe_result_t oe_get_attestation_certificate_with_evidence_v3(
+      const oe_uuid_t* format_id,
```
dthaler, line 81: missing `flags` param that exists for oe_get_evidence().
I will also observe that there is no `endorsements_buffer` output like get_evidence has, why is that?

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,476 @@
...
+      const unsigned char* subject_name,
+      uint8_t* private_key,
+      size_t private_key_size,
+      uint8_t* public_key,
+      size_t public_key_size,
+      uint8_t* runtime_claims,
+      size_t runtime_claims_size,
+      uint8_t* inittime_claims,
+      size_t inittime_claims_size,
```
dthaler, line 90: InitTimeConfigurationInterface.md has:
```
    const void* runtime_custom_claims_buffer,
    size_t runtime_custom_claims_buffer_size,
    const oe_inittime_custom_claim_buffer_t* inittime_custom_claims_buffer,
    size_t inittime_custom_claims_buffer_size,
```
Be consistent.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
mingweishih: This sounds like supporting a format of "certificate with embedded evidence" for the endorsement. In this case, the `evidence_buffer` can be NULL.

```
        result = oe_verify_evidence(
            NULL,
            NULL,
            0,
            endorsements_buffer, /* include the certificate with the embedded evidence */
            endorsements_buffer_size,
            policies,
            policies_size,
            claims,
            claims_length);
``` 

But this requires significant changes to the  `oe_get_evidence` API to support this proposal unless we still keep the `oe_get_attestation_certificate_with_evidence_v3` but drop `oe_verify_attestation_certificate_with_evidence_v3`.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
mingweishih: The evidence embedded in the certificate uses the  `OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID` flag by default, which is the contract between the get and verify certificate APIs.

I'm not sure why endorsements is not supported (currently the get API passes endorsements  as `NULL`  to `oe_get_evidence`). I'm also unclear about the use case of supporting endorsements in these APIs.
@yentsanglee can you comment on this?

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,478 @@
...
+      const oe_policy_t* policies,
+      size_t policies_size,
+      oe_claim_t** claims,
+      size_t* claims_length);
+  ```
+  The API does not introduce changes to the function prototype. The purpose of
+  the API is to pair with `oe_get_attestation_certificate_with_evidence_v3`;
+  i.e., only the API is able to verify the certificated generated via
+  `oe_get_attestation_certificate_with_evidence_v3` with either `runtime_custom_claims_buffer`
+  or `inittime_custom_claims_buffer` supplied.
```
jxyang, line 144: Why can't we have both init claims and custom claims? And how do we verify runtime claims?

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,478 @@
...
++   custom_claims_buffer_size = public_key_size + runtime_custom_claims_buffer_size;
++   custom_claims_buffer = oe_malloc(custom_claims_buffer_size);
++   if (!custom_claims_buffer)
++       OE_RAISE(OE_OUT_OF_MEMORY);
+
++   memcpy(custom_claims_buffer, public_key, public_key_size);
++   memcpy(
++       custom_claims_buffer + public_key_size,
++       runtime_custom_claims_buffer,
++       runtime_custom_claims_buffer_size);
```
jxyang, line 201: The public key could have variable size. Once you copied it into the `custom_claims_buffer`, how do we extract it out?

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,478 @@
...
+        &evidence_buffer,
+        &evidence_buffer_size,
+        NULL,
+        0);
+    OE_CHECK_MSG(
+        result, "oe_get_evidence failed with %s\n", oe_result_str(result));
+
++   evidence_with_inittime_custom_claims_size =
++       evidence_buffer_size + inittime_custom_claims_buffer_size;
++   evidence_with_inittime_custom_claims =
```
jxyang, line 221: Same question about the init time claims. How can we extract it without the size info?

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,478 @@
...
++   memcpy(
++       evidence_with_inittime_custom_claims, evidence_buffer, evidence_buffer_size);
++   memcpy(
++       evidence_with_inittime_custom_claims + evidence_buffer_size,
++       inittime_custom_claims_buffer,
++       inittime_custom_claims_buffer_size);
+
+    result = generate_x509_self_signed_certificate(
+        oid_oe_evidence,
+        sizeof(oid_oe_evidence),
```
jxyang, line 235: Watch out for the length limit of the generated certificate. I think we already went to mbedtls once to ask them to increase the length limit with the self signed certificate generated by OE. Now adding inittime and runtime claims definitely will increase the length even further.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
mingweishih: Removed the word `either` that causes confusion.

The current design is that it's up to the verifier to implement the verification logic against the runtime claims output by the the `oe_verify_attestation_certificate_with_evidence_v3`.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
mingweishih: The design assumes the public_key_size is provided by the `oe_get_attestation_certificate_with_evidence_v3` API. The certificate generated by the API should include the public key with the given same size. The `oe_verify_attestation_certificate_with_evidence_v3` firstly gets the public key from the certificate and then extracts the runtime claim from the custom_claims_buffer.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
mingweishih: The design is based on the contract between `oe_get_attestation_certificate_with_evidence_v3` and `oe_verify_attestation_certificate_with_evidence_v3` APIs. Both APIs assume the evidence with the `OE_EVIDENCE_FLAGS_EMBED_FORMAT_ID` format instead of raw report. The evidence with this format comes with a header that includes the size of the evidence.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
yentsanglee: endorsements_buffer is a special contract between evidence generating and verifying. Verifier can use the same endorsement retrieved at the time of evidence generation to verify the evidence. This is useful for verifying possibly stale evidence. I don't see a use case for certificate.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,571 @@
...
+The Open Enclave SDK currently supports attestation APIs,
+`oe_get_attestation_certificate_with_evidence_v2` and
+`oe_verify_attestation_certificate_with_evidence_v2`, that
+allow developers to obtain a self-signed certificate with
+the enclave evidence embedded and to verify the certificate,
+respectively. The latter API also outputs an array of claims,
+allowing the developers to implement additional verification
+logic. However, these APIs have following limitations:
+
+- Lack of support of init-time custom claims
```
yentsanglee, line 16: Init-time claims can be entered at the enclave creation time `oe_create_enclave` and stays the same for the life time of the enclave. It is supported. No need to add extra parameters for it here.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,571 @@
...
+   * oe_verify_attestation_certificate_with_evidence_v2 that has the same
+   * prototype but with the ability to verify a ceritifacte generated by
+   * oe_get_attestation_certificate_with_evidence_v3.
+   *
+   * @experimental
+   *
+   * @param[in] certificate_in_der A pointer to buffer holding certificate
+   * contents in DER format.
+   * @param[in] certificate_in_der_size The size of the buffer above.
+   * @param[in] endorsements_buffer A pointer to buffer holding endorsements.
```
yentsanglee, line 118: The `endorsements_buffer` should either exist in both *get* and *verify* or none. Otherwise, where do you get the endorsements buffer from?

yentsanglee:
The danger of this design is that the certificate generated by _v3 cannot be verified by _v2. Backward compatibility is not possible. Version mismatch may happen.

Another option is to store the extra claims in certificate extension with pre-defined OIDs.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,571 @@
...
+
+Specification
+--------
+
+This document proposes new APIs,
+`oe_get_attestation_certificate_with_evidence_v3` and
+`oe_verify_attestation_certificate_with_evidence_v3`, that
+address the abovementioned limitations. Note that the certificate
+generated by `oe_get_attestation_certificate_with_evidence_v3`
+does not provide full backward compatibility to older version
```
dthaler, line 40:
```suggestion
does not provide full backward compatibility to older versions
```

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,571 @@
...
+and `oe_verify_attestation_certificate`).
+
+- oe_get_attestation_certificate_with_evidence_v3
+  ```c
+  /**
+   * oe_get_attestation_certificate_with_evidence_v3
+   *
+   * Similar to oe_get_attestation_certificate_with_evidence_2, this function
+   * generates a self-signed X.509 certificate with embedded evidence generated by
+   * an attester plugin for the enclave. In addition, it allows a user to pass in
```
dthaler, line 51:
```suggestion
   * an attester plugin for the enclave. In addition, it allows an app to pass in
```

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,571 @@
...
+   * @param[in] inittime_custom_claims_buffer_size The size of the buffer above in bytes.
+   * @param[in] runtime_custom_claims_buffer An optional flat buffer representing the run-time
+   * cutom claims.
+   * @param[in] runtime_custom_claims_buffer_size The size of the buffer above in bytes.
+   * @param[in] optional_parameters The optional format-specific input parameters.
+   * @param[in] optional_parameters_size The size of buffer above in bytes.
+   * @param[out] output_certificate A pointer to buffer pointer.
+   * @param[out] output_certificate_size The size of the buffer above.
+   *
+   * @return OE_OK on success
```
dthaler, line 77:
```suggestion
   * @retval OE_OK The operation was successful.
```
Use retval for discrete return values (like oe_result_t)

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
dthaler: Use case is the same as for evidence.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,571 @@
...
+  - `inittime_custom_claims_buffer_size`
+
+- oe_verify_attestation_certificate_with_evidence_v3
+  ```c
+  /**
+   * oe_verify_attestation_certificate_with_evidence_v3
+   *
+   * This function is similar to
+   * oe_verify_attestation_certificate_with_evidence_v2 that has the same
+   * prototype but with the ability to verify a ceritifacte generated by
```
dthaler, line 110:
```suggestion
   * prototype but with the ability to verify a certificate generated by
```

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,571 @@
...
+     sizeof(inittime_custom_claims),
+     NULL,
+     0,
+     &output_certificate,
+     &output_certificate_size);
+OE_CHECK(result);
+
+...
+
+Send output_certificate to verifier
```
dthaler, line 187: I'm having a hard time understanding the use case here.  I'd expect the certificate to go to the relying party, and only the evidence to go to the verifier.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,571 @@
...
+     input_certificate_size,
+     nullptr,
+     0,
+     nullptr,
+     0,
+     &claims,
+     &claims_length);
+OE_CHECK(result);
+
+// Implement veritication logic against claims
```
dthaler, line 206:
```suggestion
// Implement verification logic against claims
```

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,571 @@
...
+    result = oe_cert_verify(&cert, NULL, NULL, 0);
+    OE_CHECK_MSG(
+        result,
+        "oe_cert_verify failed with error = %s\n",
+        oe_result_str(result));
+
+    //------------------------------------------------------------------------
+    // Validate the report's trustworthiness
+    // Verify the remote report to ensure its authenticity.
+    // set enclave to NULL because we are dealing only with remote report now
```
dthaler, line 392: Why does this use the word "report"?   That sounds obsolete.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,571 @@
...
+        oid_array_index++;
+    }
+
+    // if there is no match
+    if (oid_array_index == oid_array_count)
+        OE_RAISE_MSG(
+            OE_FAILURE, "No expected certificate extension matched", NULL);
+
+    // find the extension
+    OE_TRACE_VERBOSE("extract_x509_report_extension() succeeded");
```
dthaler, line 426: oe_cert_find_extension?

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,571 @@
...
+            policies_size,
+            claims,
+            claims_length);
+    }
+
+    OE_CHECK(result);
+    OE_TRACE_VERBOSE("quote validation succeeded");
+
+    // verify report data: hash(public key)
+    // extract public key from the cert
```
dthaler, line 488:
```suggestion
    // extract public key from the certificate
```

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,571 @@
...
+
+  The current design does not implement the logic of verifying init-time custom claims.
+  This decision is letting the developers implement such logic. 
+  Alternatively, the implementation can enforce the check of hash(`inittime_custom_claims`) against
+  `CONFIGID` by default. This topic is also open for discussion.
+
+- Lack of integrity protection with init-time custom claims.
+
+  Based on the current design, the run-time custom claims are integrity-protected (i.e.,
+  its hash is part of the SGX report) while init-time custom claims rely on the
```
dthaler, line 558: Why is this specific to "SGX reports"?   This design doc should be support multiple evidence formats.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
mingweishih: This is inherent from oe_verify_attestation_certificate_with_evidence_v2, so this seems to be a problem for existing APIs.

https://github.com/openenclave/openenclave/blob/08ebe60c1d2e3ea24ac634c673a420296bff3352/include/openenclave/attestation/verifier.h#L249

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
mingweishih: CONFIG_DATA (or CONFIG_ID) is used to verify the integrity of the init-time claims. See example in
https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/InitTimeConfigurationInterface.md#plugin-api-change-with-explicit-init-time-custom-claim-buffer-support

```
    /// Verifier Plugin
    ///
    oe_result_t oe_verify_evidence(
    const oe_uuid_t* format_id,
    const uint8_t* evidence_buffer,
    size_t evidence_buffer_size,
    const uint8_t* endorsements_buffer,
    size_t endorsements_buffer_size,
    const oe_policy_t* policies,
    size_t policies_size,
    oe_claim_t** claims,
    size_t* claims_length)
    {
        ...
        //extract the init_time_claims_buffer and verify it
        inittime_claims_buffer = _find_init_time_claim_buffer(evidence_buffer,
                                   evidence_buffer_size);
        inittime_claims_buffer_size = _find_init_time_claim_buffer_size(
                                        evidence_buffer, evidence_buffer_size);
        config_id = _find_config_id(evidence_buffer, evidence_buffer_size);
        if (inittime_claim_buffer.integrity_algorithm_id ! = 0 )
            return error;
        if (SHA256(inittime_claim_buffer.buffer, inittime_claims_buffer_size)
              != config_id)
            return error;
        //output init_time claims
        ...
    }

```

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
mingweishih: This is a simplified end-to-end example (i.e., the model used in the attested-tls sample)

dthaler: The attested tls sample has a tls session between an attester and a relying party

mingweishih: I cannot find that there is a relying party involves in the attested tls sample. Can you point it out?

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,816 @@
...
+   * @param[in] private_key_size The size of the buffer above.
+   * @param[in] public_key A public key used as the certificate's subject key.
+   * @param[in] public_key_size The size of the buffer above.
+   * @param[in] inittime_custom_claims_buffer An optional flat buffer representing the init-time
+   * custom claims.
+   * @param[in] inittime_custom_claims_buffer_size The size of the buffer above in bytes.
+   * @param[in] runtime_custom_claims_buffer An optional flat buffer representing the run-time
+   * cutom claims.
+   * @param[in] runtime_custom_claims_buffer_size The size of the buffer above in bytes.
+   * @param[in] attestation_results_buffer An optional flat buffer representing the attestation
```
shnwc, line 111: Not sure how adding `attestatin_results_buffer` to this `oe_get_attestation_certificate_with_evidence_v3()` API is the best approach to add support of the passport mode. Does this API always include evidence in the output cert? But in passport mode, the attestation results is supposed to be carried instead of the evidence.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
```diff
@@ -0,0 +1,816 @@
...
+
+  - Background-Check Model
+
+    The attester generates the evidence and directly sends the evidence to the
+    relying party. The relying party then asks the verifier to assess the trustworthiness
+    of the evidence and then validates the attestation results based on customized policies.
+
+  The end-to-end use case supported by the APIs, however, assumes that the party
+  that sends the certificate is the attester and the receiver of the certificate plays
+  the roles of both the verifier and the relying party. More specifically, the
```
shnwc, line 44: The existing API has been designed to specifically support the background check mode, with a combined verifier + relying-party entity. The design in this document feels like a patch to stretch the existing API to support the new usages (especially the passport mode). I think it would be a better approach to start a with a top-down design to build a layer on top of the existing API. For passport mode support, I'd like to see a detailed description of the usages, the cert + attestation result design, and the end-to-end flows between the three parties (attester, verifier, relying party).

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
mingweishih: You're correct, the example of passport mode below does not require evidence to be embedded. The APIs probably need different names.

docs/DesignDocs/AttestationAPIsWithFlexibleClaims.md
mingweishih: I agree to have a different APIs other than extending the existing APIs. For example, simply `oe_get_attestation_certificate` and `oe_parse_attestation_certificate`. Please refer the user experience part for the example of work flow.

The biggest concern I have is how much backward compatibility the new APIs need to provide. Also, given that we have the `oe_create_attestation_ceritifcate` API already, we would need a better naming for the new get API.

shnwc: Note: this PR could be seen as an effort to design a higher-level API to is more aligned to Dave's document [Proposed Attestation API Requirements](https://github.com/openenclave/openenclave/blob/master/docs/DesignDocs/AttestationApiRequirements.md). The existing API can be seen as a lower-level one to build on top.

yentsanglee: First generation - `oe_generate_attestation_certificate` embeds OE_REPORT in OID 1.3.6.1.4.1.311.105.1
Second generation - `oe_get_attestation_certificate_with_evidence*` embeds OE_EVIDENCE in OID 1.3.6.1.4.1.311.105.2

Report and evidence have subtle difference.

We might consider this new thing embedded in certificate an extension of evidence and give it a new OID (e.g. 1.3.6.1.4.1.311.105.2.1). This will guarantee backward compatibility. We just have to be careful picking OID to avoid collision or misinterpretation.

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+   * @param[in] private_key A private key used to sign this certificate.
+   * @param[in] private_key_size The size of the buffer above.
+   * @param[in] public_key A public key used as the certificate's subject key.
+   * @param[in] public_key_size The size of the buffer above.
+   * @param[in] attestation_results_buffer An optional flat buffer representing the attestation
+   * results from a verifier.
+   * @param[in] attestation_results_buffer_size The size of the buffer above in bytes.
+   * @param[in] optional_parameters The optional format-specific input parameters.
+   * @param[in] optional_parameters_size The size of buffer above in bytes.
+   * @param[out] output_certificate A pointer to buffer pointer.
```
shnwc, line 119: Rename `output_certificate` to `output_certificate_in_der`?

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+   * @param[in] public_key_size The size of the buffer above.
+   * @param[in] inittime_custom_claims_buffer An optional flat buffer representing the init-time
+   * custom claims.
+   * @param[in] inittime_custom_claims_buffer_size The size of the buffer above in bytes.
+   * @param[in] runtime_custom_claims_buffer An optional flat buffer representing the run-time
+   * cutom claims.
+   * @param[in] runtime_custom_claims_buffer_size The size of the buffer above in bytes.
+   * @param[in] optional_parameters The optional format-specific input parameters.
+   * @param[in] optional_parameters_size The size of buffer above in bytes.
+   * @param[out] output_certificate A pointer to buffer pointer.
```
shnwc, line 171: Rename output_certificate to output_certificate_in_der?

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+(based on the DCAP client). In addition, the API outputs a
+list of claims, allowing the developer to implement additional
+verification logic (e.g., authenticate the enclave identity, or `MRENCLAVE`).
+After a successful validation, the receiver continues the handshake
+and finally establishes an *attested* secure channel with the sender.
+
+While the two APIs are useful in the above end-to-end scenario,
+they have the following limitations:
+
+- Fail to support general use cases
```
shnwc, line 34:
```suggestion
- Does not support general use cases
```

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+      const uint8_t* public_key,
+      size_t public_key_size,
+      const uint8_t* attestion_results_buffer,
+      size_t attestation_results_buffer_size,
+      uint8_t** output_certificate,
+      size_t* output_certificate_size);
+  ```
+  The API generates a certificate with an embedded attestation results that is specifically used
+  by an attested TLS with the passport model.
+  An important note is that the API does not enforce the binding between the
```
anakrish, line 138: Is there a reason for not enforcing the binding?

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+  public key and the attestation results as `oe_get_attestation_certificate_with_evidence_v2` does. The
+  attester and the relying party are responsible to enforce and verify the binding, respectively.
```
anakrish, line 140:
```suggestion
  attester and the relying party are responsible for enforcing and verifying the binding, respectively.
```

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+  existing `oe_verify_attestation_certificate_with_evidence_v2` API, the caller of the API needs to ensure
+  that the `runtime_custom_claims_buffer` includes exactly the public key.
+  
+- oe_parse_attestation_certificate
+  ```c
+  /**
+   * oe_parse_attestation_certificate
+   *
+   * This function parses the certificate generated by
+   * oe_get_attestation_certificate_with_evidence_v3 and
```
anakrish, line 216: There is no `oe_get_attestation_certificate_with_evidence_v3`

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+      0x08, 0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f };
+  ...
+
+  // Attester
+  uint8_t* runtime_custom_claims = public_key; /* Enforce the binding */
+  uint8_t* runtime_custom_claims_buffer_size = public_key_size;
+  uint8_t* attestation_results_buffer = NULL;
+  size_t attestation_results_buffer_size = 0;
+
+  // Send the evidence, init-time custom claims, and run-time custom claims
```
anakrish, line 280: Is there a recommended pattern for sending the information to the verifier?

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+  // to the verifier. The results are saved in the attestation_results_buffer.
```
anakrish, line 281: Would this involve a "non-attested" tls between the verifier and the attester?

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+       public_key_size,
+       attestation_results_buffer,
+       attestation_results_buffer_size,
+       &output_certificate,
+       &output_certificate_size);
+  OE_CHECK(result);
+
+  ...
+
+  Send output_certificate to the relying party during the TLS handshake
```
anakrish, line 304: How does this happen? What are the APIs in mbedtls/OpenSSL that would allow this?

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+
+  ...
+  // Relying party (receive input_certificate)
```
anakrish, line 307: What are the hook-points in mbedtls/OpenSSL that allow the relying party to extract the certificate?

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+  uint8_t* runtime_custom_claims = public_key; /* Enforce the binding */
+  uint8_t* runtime_custom_claims_buffer_size = public_key_size;
+  uint8_t* output_certificate = NULL;
+  size_t output_certificate_size = 0;
+
+  ...
+
+  result = oe_get_background_check_attestation_certificate(
+       &format,
+       (const unsigned char*)"CN=Open Enclave SDK,O=OESDK TLS,C=US",
```
anakrish, line 351: Does CN have to be this value?

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+    based on customized policies.
+
+  - Background-Check Model
+
+    The attester generates the evidence and directly sends the evidence to the
+    relying party. The relying party then asks the verifier to assess the trustworthiness
+    of the evidence and then validates the attestation results based on customized policies.
+
+  The APIs, however, only support the special case of the background-check model where
+  the sender of the certificate is the attester and the receiver of the certificate plays
```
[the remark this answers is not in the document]
mingweishih: No. An example of attestation results are MAA tokens, which will be embedded into the certificate.

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+  embed the attestation results into the certificate nor can the receiver extract
+  the raw evidence from the certificate and ask the verifier to assess the evidence,
+  thus failing to support the two abovementioned models.
+
+- Lack of support of init-time custom claims
+
+  According to [Open Enclave Init-time Configuration Interface](InitTimeConfigurationInterface.md),
+  the support of init-time custom claims is desirable when the TEE hardware has such
+  support (e.g., `CONFIGID` in SGX2). The `oe_get_attestation_certificate_with_evidence_v2`
+  API does not allow developers to provide an init-time custom claim buffer as input.
```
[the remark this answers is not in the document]
mingweishih: The [docs](https://github.com/ietf-rats-wg/architecture/blob/master/draft-ietf-rats-architecture.md) only mentions the term "claims".

docs/DesignDocs/AttestedTLSAPIs.md
mingweishih: In the passport model, the enforcement requires
- The attester enforces the binding into the evidence, sends to the verifier, and uses the returned attestation results to generate the certificate
- The relying party extracts the attestation results from the certificate, and validates the binding.

The above two steps require the attester and the relying party agree on a specific format of the attestation results (e.g., MAA otkens). OE currently does not have APIs support a specific verifier, which can surely be added in the future.

docs/DesignDocs/AttestedTLSAPIs.md
mingweishih: Will update the comments here. One should either uses the interface provided by a specific verifier or OE can provide API that supports it.

docs/DesignDocs/AttestedTLSAPIs.md
mingweishih: This depends on the requirement of the verifier. For example, MAA does not require the channel to be authenticated.

docs/DesignDocs/AttestedTLSAPIs.md
mingweishih: This is the same as how the attested tls works today: configuring the SSL/TLS data structure using the generated certificate.

docs/DesignDocs/AttestedTLSAPIs.md
mingweishih: Certificate can be obtained during the standard TLS handshake. Both Mbed TLS/OpenSSL allows server/client to implement a callback that verifies the certificate.

https://github.com/openenclave/openenclave/blob/641288ea0b837e55292a3c120302157aee47c603/samples/attested_tls/server/enc/mbedtls_server.cpp#L97

https://github.com/openenclave/openenclave/blob/641288ea0b837e55292a3c120302157aee47c603/samples/attested_tls/server/enc/openssl_server.cpp#L165

docs/DesignDocs/AttestedTLSAPIs.md
mingweishih: No. Just that existing verification APIs require it.

docs/DesignDocs/AttestedTLSAPIs.md
dthaler: In some cases the attester might need to "know" to ask the verifier for format in a particular format needed by a relying party (e.g., to ask for them in JWT vs CWT format), but it doesn't need to understand the format itself (in terms of ability to parse it) since it is an opaque buffer to the attester.

docs/DesignDocs/AttestedTLSAPIs.md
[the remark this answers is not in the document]
dthaler: Right because I think the notion of "init time custom claims" is really an SGX-specific artifact that RATS solves in a more general way, if I understand init time custom claims correctly.  In RATS layered attestation, configuration data can be another layer from the enclave itself.  SGX and OE historically has not supported layered attestation and instead had (has) an SGX-only model.  Once you support layered attestation, you have a much more general solution that does not need any special term for this.

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,758 @@
...
+`oe_verify_attestation_certificate_with_evidence_v2`
+API, which assesses the trustworthiness of the evidence
+(based on the DCAP client). In addition, the API outputs a
+list of claims, allowing the developer to implement additional
+verification logic (e.g., authenticate the enclave identity, or `MRENCLAVE`).
+After a successful validation, the receiver continues the handshake
+and finally establishes an *attested* secure channel with the sender.
+
+While the two APIs are useful in the above end-to-end scenario,
+they have the following limitations:
```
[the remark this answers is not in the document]
dthaler: Line 38 links to it now

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,623 @@
...
+
+- Do not to support general use cases
+
+  A typical attestation procedure includes three roles: an attester, a relying party,
+  and a verifier. Two general models are as follows
+  (refer to [Remote Attestation Procedures Architecture](https://github.com/ietf-rats-wg/architecture/blob/master/draft-ietf-rats-architecture.md)
```
dthaler, line 38: A better URI is probably https://datatracker.ietf.org/doc/html/draft-ietf-rats-architecture

mingweishih:
@dthaler @yentsanglee @shnwc @ryanhsu19 @qiucwang
I've updated the document based on the discussion from the sig meeting yesterday. Can you review again?

shnwc (approved):

docs/DesignDocs/AttestedTLSAPIs.md
```diff
@@ -0,0 +1,1004 @@
...
+`oe_verify_attestation_certificate_with_evidence_v2`.
+The `oe_get_attestation_certificate_with_evidence_v2` API allows
+a developer to obtain a self-signed certificate with the enclave
+evidence embedded, which then is used during the TLS handshake.
+More importantly, the API binds the public key of the certificate
+with evidence such that the TLS channel based on the public key
+can be *attested*. To validate the certificate, the receiver uses the
+`oe_verify_attestation_certificate_with_evidence_v2` API, which
+assesses the trustworthiness of the evidence (e.g., based on the DCAP client for SGX).
+In addition, the API outputs a list of claims, allowing the developer
```
dthaler, line 23: Since the receiver is a relying party, not a verifier, clarify that the list of claims returned is
NOT the list of claims in the evidence, it's a list of claims from the attestation result returned by the verifier.

PR #4140, opened 2021-07-03
VS Extension release 0.17.37

rs--:


anakrish (approved):

mingweishih (approved):
Why 0.17.37?

rs--:
> Why 0.17.37?

The version convention is <SDK version>.<VS Extension release>; it doesn't necessarily follow the same convention as the SDK releases.

PR #4143, opened 2021-07-06
Log all module symbols that are not found

thomasten:
We just switched from our own module loading implementation for [EGo](https://github.com/edgelesssys/ego) to the one recently introduced in OE. It would be helpful for us if the loader logs all symbols that are not found instead of just the first one.

### DIFF
--- a/host/sgx/loadelf.c
+++ b/host/sgx/loadelf.c
@@ -740,6 +740,7 @@ static oe_result_t _link_elf_image(
     uint64_t nrelocs = 0;
     const elf64_sym_t* symtab = NULL;
     size_t symtab_size = 0;
+    int missing_symbols_count = 0;
 
     if (elf64_get_dynamic_symbol_table(&image->elf, &symtab, &symtab_size) != 0)
         goto done;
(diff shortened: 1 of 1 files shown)


host/sgx/loadelf.c
```diff
@@ -747,6 +747,7 @@ static oe_result_t _link_elf_image(
     /* Iterate through relocation records in the target image */
     relocs = (elf64_rela_t*)image->reloc_data;
     nrelocs = image->reloc_size / sizeof(relocs[0]);
+    result = OE_OK;
```
anakrish, line 750: OE follow the coding pattern of initializing result to OE_UNEXPECTED, performing checks and raising errors, and then finally setting result to OE_OK.

It would be desirable to preserve that style of coding in this function.

We could maintain a count of missing symbols and then print that, and raise an error if count is non zero.

anakrish (commented):
LGTM.

host/sgx/loadelf.c
thomasten: Thanks for the suggestion. I updated it.

mingweishih (approved):
LGTM

anakrish (approved):

PR #4150, opened 2021-07-07
Support oe_sgx_get_additional_host_entropy for the FIPS module

mingweishih:
Add the internal `oe_sgx_get_additional_host_entropy` API, which is mandatory for the SymCrypt FIPS module to obtain extra entropy from the host. That is, the API is for SymCrypt FIPS module only instead of general use.

To enforce the hard requirement, the current design throws a runtime error if the `entropy.edl` is not imported.

### DIFF
--- a/enclave/crypto/openssl/entropy.c
+++ b/enclave/crypto/openssl/entropy.c
@@ -1,0 +1,55 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/internal/raise.h>
+#include "../../core/platform_t.h"
+
+/**
+ * Declare the prototype of the following function to avoid the
+ * missing-prototypes warning.
+ */
+oe_result_t _oe_sgx_get_additional_host_entropy_ocall(
+    oe_result_t* result,
+    uint8_t* data,
+    size_t size);
+
+oe_result_t _oe_sgx_get_additional_host_entropy_ocall(
+    oe_result_t* result,
+    uint8_t* data,
+    size_t size)
+{
+    OE_UNUSED(result);
+    OE_UNUSED(data);
+    OE_UNUSED(size);
+    return OE_UNSUPPORTED;
+}
+OE_WEAK_ALIAS(
+    _oe_sgx_get_additional_host_entropy_ocall,
+    oe_sgx_get_additional_host_entropy_ocall);
+
+OE_EXPORT
+int oe_sgx_get_additional_host_entropy(uint8_t* data, size_t size)
+{
+    oe_result_t result = OE_FAILURE;
+
+    if (!data || !size)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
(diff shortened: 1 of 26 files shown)


enclave/crypto/openssl/entropy.c
```diff
@@ -0,0 +1,55 @@
...
+
+    OE_CHECK(oe_sgx_get_additional_host_entropy_ocall(&result, data, size));
+    OE_CHECK(result);
+
+    result = OE_OK;
+
+done:
+    if (result == OE_UNSUPPORTED)
+    {
+        OE_TRACE_ERROR(
```
anakrish, line 46: `oe_sgx_get_additional_host_entropy` is called only after a counter reaches a predetermined value.
Depending on the predetermined value, there is a chance that the user's tests may not run long enough for this message to be printed.

It is important that when FIPS module is used, this message is eagerly printed so that the user is aware.

mingweishih: Will invoke the function as part of initialization process.

mingweishih:
bors try

host/sgx/ocalls/ocalls.c
```diff
@@ -334,3 +339,26 @@ oe_result_t oe_verify_quote_ocall(
...
+
+oe_result_t oe_sgx_get_additional_host_entropy_ocall(uint8_t* data, size_t size)
+{
+    oe_result_t result = OE_FAILURE;
+
+    if (!data || !size)
+        OE_RAISE(OE_INVALID_PARAMETER);
+
+#if defined(__linux__)
+    if (getrandom((void*)data, size, 0) == -1)
```
anakrish, line 351: There are other error scenarios to handle here. Sometimes getrandom may fill the buffer partially:
https://man7.org/linux/man-pages/man2/getrandom.2.html

anakrish (approved):
LGTM

mingweishih:
bors try

include/openenclave/edl/sgx/platform.edl
```diff
@@ -19,4 +19,7 @@ enclave
     from "openenclave/edl/sgx/debug.edl" import *;
     from "openenclave/edl/sgx/thread.edl" import *;
     from "openenclave/edl/sgx/switchless.edl" import *;
+#ifdef OE_SGX_ENTROPY
+    from "openenclave/edl/sgx/entropy.edl" import *;
+#endif
```
manojrupireddy, line 24: Isn't OE_SGX compiler flag accessible here. Trying to understand why we introduced new compiler variable OE_SGX_ENTROPY for sgx platform file.

mingweishih: The purpose of the new flag is preventing including the `entropy.edl` by default along with `platform.edl`. The behavior this PR wants to enforce is that the developer should explicitly opt into the `entropy.edl` to use the FIPS module. However, internally we can supply the `OE_SGX_ENTROPY` variable for forward declaration so that we don't need an additional invocation of the `oeedger8r`.

manojrupireddy (approved):

PR #4161, opened 2021-07-13
conf functions from MUSL

anakrish:
fpathconf, pathconf are handled within the enclave.

Previously, sysconf dummy unimplemented (ie returned EINVAL).
With this PR most of sysconf is handled within the enclave.
Some sysconf codepaths remain unimplemented (ie return ENOSYS)
as they call into unimplemented syscalls SYS_sysinfo, SYS_getrlimit,
SYS_prlimit64, SYS_sched_getaffinity.


anakrish:
bors try

mingweishih (approved):

PR #4162, opened 2021-07-13
Update 0-base tests

RRathna:

### DIFF
--- a/include/openenclave/host.h
+++ b/include/openenclave/host.h
@@ -50,6 +50,11 @@ OE_EXTERNC_BEGIN
 #endif
 
 /**
+ * NOTE: When introducing new OE_ENCAVE_FLAG_*, make sure to update
+ * tools/oesign.c:_map_attributes(), if relevant.
+ */
+
+/**
  *  Flag passed into oe_create_enclave to run the enclave in debug mode.
  *  The flag allows the enclave to be created without the enclave binary
  *  being signed. It also gives a developer permission to debug the process
--- a/tests/sgx_zerobase/enc/enc.cpp
+++ b/tests/sgx_zerobase/enc/enc.cpp
@@ -3,42 +3,84 @@
 
 #include <openenclave/enclave.h>
 #include <openenclave/internal/tests.h>
+#include <openenclave/internal/thread.h>
+#include <iostream>
+
 #include "sgx_zerobase_t.h"
 
-const char* protected_message = "Hello world from Enclave\n\0";
+static uint64_t faulting_address = sizeof(uint64_t) - 1;
+static uint32_t error_code;
+static uint32_t exception_code;
+static oe_once_t _exception_handler_init_once;
 
-int secure_string_patching(
-    const char* source,
-    char* output,
-    size_t output_length)
+uint64_t test_pfgp_handler(oe_exception_record_t* exception_record)
 {
-    const char* running_source = source;
-    size_t running_length = output_length;
-    while (running_length > 0 && *running_source != '\0')
+    if (exception_record->code == OE_EXCEPTION_PAGE_FAULT)
     {
-        *output = *running_source;
-        running_length--;
-        running_source++;
-        output++;
+        const int MOV_INSTRUCTION_BYTES = 3;
+        faulting_address = exception_record->faulting_address;
+        error_code = exception_record->error_code;
+        exception_record->context->rip += MOV_INSTRUCTION_BYTES;
+        exception_code = OE_EXCEPTION_PAGE_FAULT;
     }
-    const char* ptr = protected_message;
-    while (running_length > 0 && *ptr != '\0')
+    else
     {
-        *output = *ptr;
-        running_length--;
-        ptr++;
-        output++;
+        /* Unexpected code */
+        oe_abort();
     }
-    if (running_length < 1)
+
+    return OE_EXCEPTION_CONTINUE_EXECUTION;
+}
+
+void _initialize_exception_handler(void)
+{
+    oe_result_t result;
+    result = oe_add_vectored_exception_handler(false, test_pfgp_handler);
+}
+
+int test_enclave_memory_access(uint64_t address, bool* exception)
+{
+    oe_result_t result = OE_OK;
+
+    if (exception)
+    {
+        /* A handler should be added only once per program execution */
+        oe_once(&_exception_handler_init_once, _initialize_exception_handler);
+    }
+
+    if (result != OE_OK)
     {
         return -1;
     }
-    *output = '\0';
-    int rval = -1;
-    OE_TEST(
-        unsecure_string_patching(&rval, source, output, output_length) ==
-        OE_OK);
-    return rval;
+
+    /* Read value from memory address 'address'*/
+    asm volatile("mov (%0), %0\n\t" : : "r"(address) : "memory");
+
+    if (exception)
+    {
+        if (faulting_address == address)
+            *exception = true;
+        else
+            *exception = false;
+    }
+
+    return 0;
+}
+
+int test_ecall(const char* message)
+{
+    if (!message)
+        return -1;
+    else
+        fprintf(stdout, "[enclave] Message from host : %s\n", message);
+
+    int res = -1;
+    const char* input = "testing ocall\0";
+    OE_TEST(test_ocall(&res, input) == OE_OK);
+    if (res != 0)
+        fprintf(stderr, "[enclave] ocall failed %d\n", res);
+
+    return res;
 }
 
 OE_SET_ENCLAVE_SGX2(
@@ -47,7 +89,7 @@ OE_SET_ENCLAVE_SGX2(
     {0},     /* ExtendedProductID */
     {0},     /* FamilyID */
     true,    /* Debug */
-    false,   /* CapturePFGPExceptions */
+    true,    /* CapturePFGPExceptions */
     false,   /* RequireKSS */
     true,    /* CreateZeroBaseEnclave */
     0x21000, /* StartAddress */
--- a/tests/sgx_zerobase/host/host.cpp
+++ b/tests/sgx_zerobase/host/host.cpp
@@ -1,40 +1,37 @@
 // Copyright (c) Open Enclave SDK contributors.
 // Licensed under the MIT License.
 
+#include <openenclave/bits/exception.h>
 #include <openenclave/host.h>
+#include <openenclave/internal/error.h>
 #include <openenclave/internal/tests.h>
+#include <sys/mman.h>
 #include <iostream>
-#include <vector>
+
+#include "../host/sgx/cpuid.h"
 #include "sgx_zerobase_u.h"
 
 const char* message = "Hello world from Host\n\0";
 
-int unsecure_string_patching(
-    const char* source,
-    char* output,
-    size_t output_length)
+static bool _is_misc_region_supported()
 {
-    size_t running_length = output_length;
-    while (running_length > 0 && *source != '\0')
-    {
-        *output = *source;
-        running_length--;
-        source++;
-        output++;
-    }
-    const char* ptr = message;
-    while (running_length > 0 && *ptr != '\0')
-    {
-        *output = *ptr;
-        running_length--;
-        ptr++;
-        output++;
-    }
-    if (running_length < 1)
-    {
+    uint32_t eax, ebx, ecx, edx;
+    eax = ebx = ecx = edx = 0;
+
+    // Obtain feature information using CPUID
+    oe_get_cpuid(CPUID_SGX_LEAF, 0x0, &eax, &ebx, &ecx, &edx);
+
+    // Check if EXINFO is supported by the processor
+    return (ebx & CPUID_SGX_MISC_EXINFO_MASK);
+}
+
+int test_ocall(const char* message)
+{
+    if (!message)
         return -1;
-    }
-    *output = '\0';
+
+    fprintf(stdout, "[host] Message from enclave : %s\n", message);
+
     return 0;
 }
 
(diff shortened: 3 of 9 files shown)


radhikaj (approved):
NIT: put braces around all if and else blocks.

tests/sgx_zerobase/host/host.cpp
```diff
@@ -1,40 +1,54 @@
...
+        1,
+        PROT_READ | PROT_WRITE,
+        MAP_SHARED | MAP_ANONYMOUS,
+        -1,
+        0);
+
+    if (loc != (void*)addr)
+        return 1;
+
+    munmap(loc, 1);
```
mingweishih, line 51: why munmap here?

tests/sgx_zerobase/host/host.cpp
```diff
@@ -44,48 +58,142 @@ int main(int argc, const char* argv[])
...
+        return 1;
     }
 
-    const char expect[] = "My First App\n"
-                          "Hello world from Enclave\n"
-                          "My First App\n"
-                          "Hello world from Host\n";
+    fprintf(
+        stdout,
+        "\nTest 5: Check host access at address 0x0\n"
```
mingweishih, line 134: What's the intension of this test? mmap at address 0 and expect the mmap to fail?

tests/sgx_zerobase/host/host.cpp
```diff
@@ -44,48 +58,142 @@ int main(int argc, const char* argv[])
...
+
+    if (test_host_mmap(address) == 0)
     {
-        fprintf(stderr, "%s: returned string don't match\n", argv[0]);
+        fprintf(stderr, "host mmap at address %lx succeeded\n", address);
         return 1;
     }
 
+    /* Enclave memory access tests */
+    if (flags & OE_ENCLAVE_FLAG_SIMULATE)
```
mingweishih, line 146: Why doesn't Test 2 have this check?

tests/sgx_zerobase/host/host.cpp
```diff
@@ -44,48 +58,142 @@ int main(int argc, const char* argv[])
...
+        struct exceptions exception;
+
+        fprintf(
+            stdout,
+            "\nTest 6: Enclave memory access at 0x0\n"
+            "Expected result: page fault\n");
+
+        address = 0x0;
+
+        OE_TEST(test_enclave_memory_access(enclave, &res, address, &exception));
```
mingweishih, line 162: This test does not make sense if address 0x0 is not mapped.

tests/sgx_zerobase/host/host.cpp
```diff
@@ -44,48 +58,142 @@ int main(int argc, const char* argv[])
...
     {
-        fprintf(stderr, "%s: enclave called failed\n", argv[0]);
-        exit(1);
+        fprintf(stderr, "[host]: ecall/ocall failed %d\n", res);
+        return 1;
+    }
+
+    fprintf(
+        stdout,
+        "\nTest 4: Check host access to [/proc/sys/vm/mmap_min_addr, "
```
mingweishih, line 120: Can you also show the value of these two addresses?

tests/sgx_zerobase/host/host.cpp
```diff
@@ -44,48 +58,142 @@ int main(int argc, const char* argv[])
...
+        fprintf(stdout, "address %lx, exception %d\n", address, exception.code);
+        if (!exception.exception)
+            return 1;
+
+        fprintf(
+            stdout,
+            "\nTest 7: memory access at 0x22000 in a 0-base enclave\n"
+            "Expected result: no exceptions\n");
+
+        address = 0x30000;
```
mingweishih, line 178: Add comment or print the 0-base settings and explain why access to this address does not cause fail

tests/sgx_zerobase/enc/enc.cpp
```diff
@@ -3,42 +3,87 @@
...
-    while (running_length > 0 && *running_source != '\0')
+    if (exception_record->code == OE_EXCEPTION_PAGE_FAULT)
+    {
+        const int MOV_INSTRUCTION_BYTES = 3;
+        faulting_address = exception_record->faulting_address;
+        error_code = exception_record->error_code;
+        exception_record->context->rip += MOV_INSTRUCTION_BYTES;
+        exception_code = OE_EXCEPTION_PAGE_FAULT;
+    }
+    else if (exception_record->code == OE_EXCEPTION_ACCESS_VIOLATION)
```
mingweishih, line 24: Drop the else if block as it's not needed by this test.

tests/sgx_zerobase/enc/enc.cpp
```diff
@@ -3,42 +3,87 @@
...
     }
-    *output = '\0';
-    int rval = -1;
-    OE_TEST(
-        unsecure_string_patching(&rval, source, output, output_length) ==
-        OE_OK);
-    return rval;
+
+    /* Assign value to memory address 'address'*/
+    asm volatile("mov %0, %%r8\n\t"
```
anakrish, line 54: Since `r8` is being used, you need to mark it as clobbered in the clobber list of the inline assembly.
Alternatively the following might work, since %0 is a register.
```c
asm volatile("mov %0, (%0)\n\t"
                     : "memory"
                     : "r"(address));
```

tests/sgx_zerobase/enc/enc.cpp
```diff
@@ -3,42 +3,87 @@
 
 #include <openenclave/enclave.h>
 #include <openenclave/internal/tests.h>
+#include <iostream>
+
 #include "sgx_zerobase_t.h"
 
-const char* protected_message = "Hello world from Enclave\n\0";
+static uint64_t faulting_address;
```
anakrish, line 10: Since we are testing for address zero as well, it is better to initialize `faulting_address` to a non-zero invalid value like `(uint64_t) -1`.

tests/sgx_zerobase/host/host.cpp
```diff
@@ -1,40 +1,54 @@
...
-    *output = '\0';
+
+    fprintf(stdout, "[host] Message from enclave : %s\n", message);
+
+    return 0;
+}
+
+int test_host_mmap(uint64_t addr)
+{
+    void* loc = mmap(
```
anakrish, line 40: Why do we need to test host mmap?

RRathna: I am actually not testing the mmap function itself, using it to test access to different memory addresses. I'll rename it to host_mmap().

tests/sgx_zerobase/host/host.cpp
RRathna: yes, it should get an page fault - needed by mystikos.

mingweishih: Why is that needed? The use case of 0-base in mystikos should be
- The host mmap address 0; i.e., address 0 is accessible by the host
- Then the enclave attempts to access address 0 -> fault is raised

RRathna: Ah, okay. mmap will always fail for address lower than 0x1000 (the lower address range is only mappable using sudo). This test is not needed. *Enclave* page fault when accessing address zero is the requirement.

tests/sgx_zerobase/host/host.cpp
RRathna: It is mapped by sg_enclave_common.so.1.

tests/sgx_zerobase/host/host.cpp
RRathna: Wrong usage, removed now.

tests/sgx_zerobase/host/host.cpp
RRathna: Modified to skip tests after test 1 if in sim-mode.

tests/sgx_zerobase/host/host.cpp
RRathna: removed the function, since mmap is being used only once inside the host now.

tests/sgx_zerobase/host/host.cpp
```diff
@@ -44,55 +41,149 @@ int main(int argc, const char* argv[])
...
     }
 
+    /* Enclave memory access tests */
+    if (_is_misc_region_supported())
+    {
+        struct exceptions exception;
+
+        fprintf(
+            stdout,
+            "\nTest 5: Enclave memory access at 0x0\n"
```
mingweishih, line 137: Given that we use the default setting (i.e., the host cannot mmap at address 0), this test should access the address that can be mmapped and within the range (i.e., `0x15000` used by the previous test).

tests/sgx_zerobase/host/host.cpp
```diff
@@ -44,55 +41,149 @@ int main(int argc, const char* argv[])
...
+        }
+
+        fprintf(stdout, "address %lx, exception %d\n", address, exception.code);
+        if (!exception.exception)
+            return 1;
+
+        fprintf(
+            stdout,
+            "\nTest 6: memory access between enclave image start address "
+            "and the end of elrange - [0x21000, 0x40000)\n"
```
mingweishih, line 156: How do we get the elrange?

tests/sgx_zerobase/enc/enc.cpp
```diff
@@ -3,42 +3,76 @@
...
-    while (running_length > 0 && *running_source != '\0')
+    if (exception_record->code == OE_EXCEPTION_PAGE_FAULT)
     {
-        *output = *running_source;
-        running_length--;
-        running_source++;
-        output++;
+        const int MOV_INSTRUCTION_BYTES = 3;
+        faulting_address = exception_record->faulting_address;
+        error_code = exception_record->error_code;
```
mingweishih, line 20: `error_code` seems not needed and can be removed. But I'm curious what's the value of the error_code, can you show me?

RRathna: The exception error codes are followed as per definition in include/openenclave/bits/exception.h.

1: Test 2: Create 0-base enclave
1: Expected result : OE_OK
1: 
1: Test 3: Test ecall, ocall on 0-base enclave
1: Expected result : OE_OK
1: [host] Message from enclave : testing ocall
1: 
1: Test 4: Check host access between /proc/sys/vm/mmap_min_addr and enclave image start address - [0x1000, 0x21000)
1: Expected result : success
1: 
1: Test 5: Enclave memory access at 0x0
1: Expected result: page fault
1: address 0x0, exception 5
1: 
1: Test 6: memory access between enclave image start address and the end of elrange - [0x21000, 0x40000)
1: Expected result: no exceptions
1: address 0x2f000, exception 0
1: 
1: === passed all tests (sgx_zerobase)
1/1 Test #1: tests/sgx_zerobase ...............   Passed    0.12 sec

RRathna: Removed error_code and made argument 'exception' in ecall a bool.

tests/sgx_zerobase/host/host.cpp
RRathna: We need to test that accessing address 0x0 produces page-fault as it is required by mystikos for their node.js on OE project.. I have added another test to check access to address 0x15000 from enclave.

tests/sgx_zerobase/host/host.cpp
RRathna: Added a comment to explain calculation of address elrange_end.

tests/sgx_zerobase/host/host.cpp
mingweishih: Accessing to address 0 will raise a page fault already without 0-base by default, so I don't think the test currently is relevant to the feature (unless we can have host mmap at 0x0) . If it's required by other application based on OE, I think we don't need to it here.

RRathna: okay, understood. Removed access check at 0x0 from inside enclave.

mingweishih (approved):
Looks good to me. One minor last comment only to the edl file.

anakrish (approved):

tests/sgx_zerobase/sgx_zerobase.edl
```diff
@@ -6,17 +6,22 @@ enclave {
     from "openenclave/edl/fcntl.edl" import *;
     from "openenclave/edl/sgx/platform.edl" import *;
 
+    struct exceptions
```
mingweishih, line 9: Is this still needed?

tests/sgx_zerobase/enc/enc.cpp
```diff
@@ -3,42 +3,78 @@
...
-    const char* ptr = protected_message;
-    while (running_length > 0 && *ptr != '\0')
+
+    return OE_EXCEPTION_CONTINUE_EXECUTION;
+}
+
+int test_enclave_memory_access(uint64_t address, bool* exception)
+{
+    oe_result_t result = OE_OK;
+    static int once = 0;
```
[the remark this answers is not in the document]
RRathna: Thanks @anakrish. Modified to use oe_once.

radhikaj:
Bors r+

radhikaj:
Bors r+

RRathna:
bors try

include/openenclave/host.h
```diff
@@ -72,7 +72,7 @@ OE_EXTERNC_BEGIN
 /**
  *  Flag passed into oe_create_enclave to run the enclave in simulation mode.
  */
-#define OE_ENCLAVE_FLAG_SIMULATE 0x00000002u
+#define OE_ENCLAVE_FLAG_SIMULATE 0x00000020u
```
mingweishih, line 75: Can you explain on this change?

RRathna: When we use oesign and set debug=1, the properties.config.attribute gets set to |= SGX_FLAGS_DEBUG - https://github.com/openenclave/openenclave/blob/f8b79cf8eb64874ccd0cae951ff7f52672f1d126/tools/oesign/oesign.c#L518-L524.Later when the sgx_build_enclave() is called in host/sgx/create.c, the enclave->debug is set to 0 and enclave->simulate are both being set to 1. This is because we check to see if the value attributes has bit OE_ENCLAVE_FLAG_DEBUG set for debug and OE_ENCLAVE_FLAG_SIMULATION for simulate  - https://github.com/openenclave/openenclave/blob/f8b79cf8eb64874ccd0cae951ff7f52672f1d126/host/sgx/create.c#L827-L828.This is because OE_ENCLAVE_FLAG_SIMULATION = 0x2 and so is SGX_FLAGS_DEBUG = 0x2.

The problem was that when we created enclaves in debug mode, using oesign, they were actually being created in simulation mode.

RRathna: Added a function to map attributes in oesign.c from SGX_FLAGS_* to OE_ENCLAVE_FLAG_*.

RRathna:
bors try

tools/oesign/oesign.c
```diff
@@ -625,6 +626,29 @@ oe_result_t _initialize_enclave_properties(
 done:
     return result;
 }
+uint64_t _map_attributes(const oe_sgx_enclave_properties_t* properties)
```
mingweishih, line 629: Add a line above. Also, make the function static.

tools/oesign/oesign.c
```diff
@@ -625,6 +626,29 @@ oe_result_t _initialize_enclave_properties(
...
+{
+    /*
+     * This function maps the attributes set by oesign from
+     * SGX_FLAGS_* to OE_ENCLAVE_FLAG_* before calling into
+     * OE specific functions.
+     */
+    uint64_t attributes = properties->config.attributes;
```
mingweishih, line 636: Consider initialize this to 0.

tools/oesign/oesign.c
```diff
@@ -625,6 +626,29 @@ oe_result_t _initialize_enclave_properties(
...
+
+    if (properties->config.attributes & SGX_FLAGS_DEBUG)
+    {
+        attributes &= ~SGX_FLAGS_DEBUG;
```
mingweishih, line 640: Not needed if attributes is initialized as 0.

tools/oesign/oesign.c
```diff
@@ -625,6 +626,29 @@ oe_result_t _initialize_enclave_properties(
...
+        attributes |= OE_ENCLAVE_FLAG_DEBUG;
+    }
+
+    if (properties->config.attributes & SGX_FLAGS_KSS)
+    {
+        attributes &= ~SGX_FLAGS_KSS;
```
mingweishih, line 646: Ditto

PR #4170, opened 2021-07-17
mmap and munmap implementations

anakrish:
mmap and munmap are implemented to support their most common
use: to allocate and deallocate memory.

mmap works atop OE's memory allocator (dlmalloc/snmalloc) and
thus can leverage the allocator's performance.

A future PR will implement MAP_FIXED which supports mapping files.

### DIFF
--- a/include/openenclave/internal/safemath.h
+++ b/include/openenclave/internal/safemath.h
@@ -254,6 +254,17 @@ OE_INLINE oe_result_t oe_safe_mul_sizet(size_t a, size_t b, size_t* c)
     SAFE_MULTIPLY(a, b, c, 0, OE_SIZE_MAX);
 }
 
+OE_INLINE oe_result_t oe_safe_round_up_u64(uint64_t a, uint64_t b, uint64_t* c)
+{
+    if (!b)
+        return OE_INVALID_PARAMETER;
+    uint64_t value = 0;
+    oe_result_t result = oe_safe_add_u64(a, b - 1, &value);
+    if (result != OE_OK)
+        return result;
+    return oe_safe_mul_u64(value / b, b, c);
+}
+
 /* Re-enable GCC warnings for -Wtype-limits. */
 #ifdef __GNUC__
 #pragma GCC diagnostic pop
--- a/libc/mman.c
+++ b/libc/mman.c
@@ -1,0 +1,362 @@
+// Copyright (c) Open Enclave SDK contributors.
+// Licensed under the MIT License.
+
+#include <openenclave/corelibc/errno.h>
+#include <openenclave/internal/raise.h>
+#include <openenclave/internal/safemath.h>
+#include <openenclave/internal/thread.h>
+#include <openenclave/internal/utils.h>
+#include <stdlib.h>
+
+#include "mman.h"
+#include "openenclave/bits/defs.h"
+#include "openenclave/bits/result.h"
+#include "syscall.h"
+
+static oe_mapping_t* _mappings;
+static oe_spinlock_t _lock;
+
+static void _clear_mappings(void)
+{
+    oe_mapping_t* m = _mappings;
+    _mappings = 0;
+    while (m)
+    {
+        oe_mapping_t* next = m->next;
+        free((void*)m->start);
+        m->start = 0;
+        free(m->status_vector);
+        m->status_vector = NULL;
+        free(m);
+        m = next;
+    }
+}
+
+static void _call_atexit(void)
+{
+    atexit(_clear_mappings);
+}
+
+static void _register_atexit_callback(void)
+{
+    static oe_once_t once = OE_ONCE_INITIALIZER;
+    oe_once(&once, _call_atexit);
+}
+
+static oe_result_t _validate_mmap_parameters(
+    void* addr,
+    size_t length,
+    int prot,
+    int flags,
+    int fd,
+    off_t offset)
+{
+    oe_result_t result = OE_UNEXPECTED;
+    int flags_copy = flags;
+
+    // If addr is not NULL, then the kernel takes it as a hint about where to
+    // place the mapping; on Linux, the kernel will pick a nearby page boundary
+    // (but always above or equal to the value specified by
+    // /proc/sys/vm/mmap_min_addr) and attempt to create the mapping there.
+    // OE currently does not support this usage.
+    if (addr != NULL)
+        OE_RAISE_MSG(
+            OE_UNSUPPORTED, "non-zero `addr` parameter is not supported.");
+
+    // PROT_NONE and PROT_EXEC are not supported.
+    if (prot == PROT_NONE || (prot & PROT_EXEC))
+        OE_RAISE_MSG(OE_UNSUPPORTED, "unsupported `prot` value %d", prot);
+
+    if (length == 0)
+        OE_RAISE_MSG(OE_INVALID_PARAMETER, "length must be non zero");
+
+    enum
+    {
+        UNSUPPORTED,
+        IGNORED,
+        SUPPORTED
+    };
+    static struct support
+    {
+        int flag;
+        int support;
+    } flags_table[] = {
+        {MAP_SHARED, SUPPORTED},
+        {MAP_SHARED_VALIDATE, SUPPORTED},
+        {MAP_PRIVATE, SUPPORTED},
+#ifdef MAP_32BIT
+        {MAP_32BIT, UNSUPPORTED},
+#endif
+        {MAP_ANON, SUPPORTED},
+        {MAP_ANONYMOUS, SUPPORTED},
+        {MAP_DENYWRITE, IGNORED /* by spec */},
+        {MAP_EXECUTABLE, IGNORED /* by spec */},
+        {MAP_FILE, IGNORED /* by spec */},
+        {MAP_FIXED, UNSUPPORTED},
+        {MAP_FIXED_NOREPLACE, UNSUPPORTED},
+        {MAP_GROWSDOWN, UNSUPPORTED},
+        {MAP_HUGETLB, UNSUPPORTED},
+        {MAP_HUGE_2MB, UNSUPPORTED},
+        {MAP_HUGE_1GB, UNSUPPORTED},
+        {MAP_LOCKED, UNSUPPORTED},
+        {MAP_NONBLOCK, IGNORED /* no special handling by OE */},
+        {MAP_NORESERVE, IGNORED /* no special handling by OE */},
+        {MAP_POPULATE, IGNORED /* no special handling by OE */},
+        {MAP_STACK, IGNORED /* currently no-op on Linux */},
+        {MAP_SYNC, IGNORED /* no special handling needed for OE */},
+        // MUSL doesn't defined MAP_UNINITIALIZED
+        // { MAP_UNINITIALIZED, SUPPORTED }
+    };
+
+    // Of the above flags, only MAP_FIXED is specified in POSIX.1-2001 and
+    // POSIX.1-2008.  However, most systems also support MAP_ANONYMOUS.
+
+    for (int i = 0; i < OE_COUNTOF(flags_table); ++i)
+    {
+        if (flags_copy & flags_table[i].flag)
+        {
+            if (flags_table[i].support == UNSUPPORTED)
+                OE_RAISE_MSG(
+                    OE_UNSUPPORTED,
+                    "unsupported `flag` value %d",
+                    flags_table[i].flag);
+
+            // Remove flag.
+            flags_copy &= ~flags_table[i].flag;
+        }
+    }
+
+    if (flags_copy)
+        OE_RAISE_MSG(OE_INVALID_PARAMETER, "invalid flag supplied");
+
+    // MAP_SHARED, MAP_SHARED_VALIDATE, MAP_PRIVATE are all treated the same
+    // since an enclave is a single process. Exactly one of them must be
+    // specified. They occupy the lower two bits of flags.
+    OE_STATIC_ASSERT(MAP_SHARED == 0x01);
+    OE_STATIC_ASSERT(MAP_PRIVATE == 0x02);
+    OE_STATIC_ASSERT(MAP_SHARED_VALIDATE == 0x03);
+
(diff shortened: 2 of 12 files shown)


anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

radhikaj (approved):

libc/mman.c
```diff
@@ -0,0 +1,355 @@
...
+    // place the mapping; on     Linux, the kernel will pick a nearby page
+    // boundary (but always above or equal to the value specified by
+    //   /proc/sys/vm/mmap_min_addr) and attempt to create the mapping there.
+    // OE currently does not support this usage.
+    if (addr != NULL)
+        OE_RAISE_MSG(
+            OE_UNSUPPORTED, "non-zero `addr` parameter is not supported.");
+
+    // PROT_NONE and PROT_EXEC are not supported.
+    if (prot == PROT_NONE || (prot & PROT_EXEC))
```
mingweishih, line 65: How do we enforce `PROT_READ` or `PROT_WRITE` only?

libc/mman.c
```diff
@@ -0,0 +1,355 @@
...
+
+        break;
+    }
+}
+
+int oe_munmap(void* addr, uint64_t length)
+{
+    int ret = -1;
+    uint64_t start = (uint64_t)addr;
+    uint64_t end = oe_round_up_to_page_size(start + length);
```
mingweishih, line 319: consider using safe math

superajun-wsj:
Hi anakrish, I noticed that intel sgx sdk provide similar function to mmap and munmap, please refer to https://github.com/intel/linux-sgx/blob/master/common/inc/sgx_rsrv_mem_mngr.h, why not implement mmap and munmap in openenclave based on intel sgx-sdk please?

anakrish:
> Hi anakrish, I noticed that intel sgx sdk provide similar function to mmap and munmap, please refer to https://github.com/intel/linux-sgx/blob/master/common/inc/sgx_rsrv_mem_mngr.h, why not implement mmap and munmap in openenclave based on intel sgx-sdk please?

@superajun-wsj
The code you pointed to is SGX SDK specific, and will not work as-is with OE SDK. Moreover, the code seems more appropriate for EDMM/SGX2. Having said that, what are the specific features that you are interested in from the SGX SDK implementation?

superajun-wsj:
> > Hi anakrish, I noticed that intel sgx sdk provide similar function to mmap and munmap, please refer to https://github.com/intel/linux-sgx/blob/master/common/inc/sgx_rsrv_mem_mngr.h, why not implement mmap and munmap in openenclave based on intel sgx-sdk please?
> 
> @superajun-wsj
> The code you pointed to is SGX SDK specific, and will not work as-is with OE SDK. Moreover, the code seems more appropriate for EDMM/SGX2. Having said that, what are the specific features that you are interested in from the SGX SDK implementation?

we are interested in SGX2 more then SGX1, because SGX1 EPC is too small. Except that, it seems that sgx-sdk's implement could provide epc memory reserve feature, and it also support mmap a memory with a advise address, last it could privode munmap feature with an address and size. In one world, I think sgx-sdk's implement is very close to mmap and munmap except file fd mapping.

anakrish:
@superajun-wsj 
Thanks for your info.
This PR is more suited for SGX1. It will allow more code/libraries to work within enclaves.
We will keep your requirements in mind when we extend the mmap support for SGX2.
In this PR, mmap is implemented **atop** the memory allocators and thus it plays well with rest of OE SDK.
The allocator can be swapped; yet mmap would work. The caveats are that only a subset of mmap features are supported.

`sgx_rsrv_mem_mgr` seems to manage a separated "reserved memory". It is going to require  design consensus to determine whether OE also ought to use a separate reserved memory for mmap or whether the entire heap ought to be managed using mmap. Another challenge for OE is how to support custom alloctors and mmap. Normally, allocators work atop mmap.
In addition to SGX SDK, another reference implementation for OE  is Mystikos' mmap implementation:
https://github.com/deislabs/mystikos/blob/main/kernel/mman.c.

Having said that, let me see if it is possible to bring in SGX's `sgx_rsrv_mem_mgr` or an implementation similar to Mystikos to OE.

manojrupireddy (approved):

libc/mman.c
anakrish: Currently we don't enforce this by design.
In SGX1, we'd need to make an OCALL to call mprotect on the page to change the permissions.
However, the enclave has no way to enforce that the host has honored the mprotect request.
There this does not seem possible for SGX1.

There is also the added complexity that changing page permissions could corrupt the allocator since the allocator may choose to store house-keeping data on freed pages. To reliably support coexistence with the allocator, we would need to move mmap to live beneath the allocator.

libc/mman.c
```diff
@@ -0,0 +1,355 @@
...
+            {
+                // If pages lies in specified range, unset it's status.
+                if (start <= a && a < end)
+                    m->status_vector[bv_idx] &= ~bit_mask;
+
+                bit_mask <<= 1;
+                if (!bit_mask)
+                {
+                    // Retain the mapping if any bit in the vector is set.
+                    delete = delete &&!m->status_vector[bv_idx];
```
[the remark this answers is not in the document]
anakrish: Surprisingly, it doesn't.

anakrish:
bors try

include/openenclave/internal/safemath.h
```diff
@@ -254,6 +254,17 @@ OE_INLINE oe_result_t oe_safe_mul_sizet(size_t a, size_t b, size_t* c)
     SAFE_MULTIPLY(a, b, c, 0, OE_SIZE_MAX);
 }
 
+OE_INLINE oe_result_t oe_safe_round_up_u64(uint64_t a, uint64_t b, uint64_t* c)
+{
+    if (!b)
+        return OE_INVALID_PARAMETER;
+    uint64_t tmp = 0;
```
dthaler, line 261:
```suggestion
    uint64_t value = 0;
```
nit: avoid abbreviations per OE style guidance

libc/mman.c
```diff
@@ -0,0 +1,362 @@
...
+
+            // Remove flag.
+            flags_copy &= ~flags_table[i].flag;
+        }
+    }
+
+    if (flags_copy)
+        OE_RAISE_MSG(OE_INVALID_PARAMETER, "invalid flag supplied");
+
+    if (length == 0)
```
dthaler, line 131: suggest moving this check up to line 71.  No point in spending the cycles on 113-126 if length is 0

libc/mman.c
```diff
@@ -0,0 +1,362 @@
...
+        OE_RAISE_MSG(OE_INVALID_PARAMETER, "length must be non zero");
+
+    // MAP_SHARED, MAP_SHARED_VALIDATE, MAP_PRIVATE are all treated as same
+    // since enclave is single process. Exactly one of them must be specified.
```
dthaler, line 135:
```suggestion
    // MAP_SHARED, MAP_SHARED_VALIDATE, MAP_PRIVATE are all treated the same
    // since an enclave is a single process. Exactly one of them must be specified.
```

libc/mman.c
```diff
@@ -0,0 +1,362 @@
...
+    if ((ret = posix_memalign(&ptr, OE_PAGE_SIZE, length)) != 0)
+    {
+        // posix_memalign does not set errno (by spec).
+        // Set it ourselves.
+        oe_errno = ret;
+        OE_RAISE_MSG(
+            OE_OUT_OF_MEMORY, "posix_memalign failed with code %d", ret);
+    }
+
+    if (!vector || !ptr || !m)
```
dthaler, line 200: recommend moving this check up before line 191 so you don't call posix_memalign if out of memory

libc/mman.c
```diff
@@ -0,0 +1,362 @@
...
+    {
+        oe_errno = OE_ENOMEM;
+        OE_RAISE(OE_OUT_OF_MEMORY);
+    }
+
+    // Setup mapping.
```
dthaler, line 206:
```suggestion
    // Set up mapping.
```

libc/mman.c
```diff
@@ -0,0 +1,362 @@
...
+    OE_CHECK(oe_safe_add_u64((uint64_t)m->start, length, (uint64_t*)&m->end));
+    m->status_vector = vector;
+    memset(ptr, 0, length);
+
+    // Set relevant bits of status vector to 1.
+    {
+        int bv_idx = 0;
+        uint8_t bit_mask = 0x01;
+        // Since m->end has been rounded to OE_PAGE_SIZE and been validated via
+        // oe_safe_Add_u64, it is safe to add OE_PAGE_SIZE to a since it won't
```
dthaler, line 217:
```suggestion
        // oe_safe_Add_u64, it is safe to add OE_PAGE_SIZE to it since it won't
```
typo

libc/mman.c
```diff
@@ -0,0 +1,362 @@
...
+        if (start > m->start || end < m->end)
+        {
+            // Partial unmapping.
+            uint8_t bit_mask = 1;
+            int bv_idx = 0;
+
+            // Mark all pages in given address range as unmapped.
+            for (uint64_t a = m->start; a < m->end; a += OE_PAGE_SIZE)
+            {
+                // If pages lies in the specified range, unset it's status.
```
dthaler, line 289:
```suggestion
                // If pages lies in the specified range, unset its status.
```
typo

tests/mman/enc/enc.c
```diff
@@ -0,0 +1,317 @@
...
+        uint64_t t = p1_start;
+        p1_start = p2_start;
+        p2_start = t;
+
+        t = p1_end;
+        p1_end = p2_end;
+        p2_end = t;
+    }
+
+    // Do an unmap that starts within  p1 and ends within  p2.
```
dthaler, line 71:
```suggestion
    // Do an unmap that starts within p1 and ends within p2.
```

tests/mman/enc/enc.c
```diff
@@ -0,0 +1,317 @@
...
+        MAP_FAILED);
+    OE_TEST(errno == EINVAL);
+
+    // Zero length should fail.
+    OE_TEST(
+        mmap(NULL, 0, PROT_READ, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0) ==
+        MAP_FAILED);
+    OE_TEST(errno == EINVAL);
+
+    // Large mmap  should fail.
```
dthaler, line 143:
```suggestion
    // Large mmap should fail.
```

tests/mman/enc/enc.c
```diff
@@ -0,0 +1,317 @@
...
+        mmap(
+            NULL,
+            1024,
+            PROT_READ | PROT_WRITE,
+            MAP_ANONYMOUS | MAP_PRIVATE,
+            -1,
+            0) != MAP_FAILED);
+    OE_TEST(errno == 0);
+
+    // Test various flags
```
dthaler, line 184:
```suggestion
    // Test various flags.
```

tests/mman/enc/enc.c
```diff
@@ -0,0 +1,317 @@
...
+        MAP_FAILED);
+    OE_TEST(errno == EINVAL);
+
+    OE_TEST(
+        mmap(NULL, 1024, PROT_READ, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0) !=
+        MAP_FAILED);
+    OE_TEST(errno == 0);
+
+    OE_TEST(
+        mmap(NULL, 5024, PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0) !=
```
dthaler, line 170: why 5024 instead of 1024 like the other lines?

Maybe define and use something like
```
   const size_t chunk_size = 1024;
```
at the top for use in all of these?

tests/mman/host/host.c
```diff
@@ -0,0 +1,41 @@
...
+    if (argc != 2)
+    {
+        fprintf(stderr, "Usage: %s ENCLAVE_PATH\n", argv[0]);
+        return 1;
+    }
+
+    const uint32_t flags = oe_get_create_flags();
+
+    if ((result = oe_create_mman_enclave(
+             argv[1], OE_ENCLAVE_TYPE_SGX, flags, NULL, 0, &enclave)) != OE_OK)
```
dthaler, line 24: why TYPE_SGX instead of TYPE_AUTO?  I don't see any other SGX-only code in this file.
```suggestion
             argv[1], OE_ENCLAVE_TYPE_AUTO, flags, NULL, 0, &enclave)) != OE_OK)
```

dthaler (commented):
Overall looks good to me.  Just some nits.

libc/mman.c
anakrish: Done.

libc/mman.c
anakrish: Done.

libc/mman.c
anakrish: Done.

libc/mman.c
anakrish: Done.

libc/mman.c
anakrish: Fixed. Put a within backticks.

libc/mman.c
anakrish: Fixed.

tests/mman/enc/enc.c
anakrish: Done.

tests/mman/enc/enc.c
anakrish: Done.

tests/mman/enc/enc.c
anakrish: Done.

tests/mman/host/host.c
anakrish: Good point. Done.

tests/mman/enc/enc.c
anakrish: Good idea. Done.
This particular test is just to check for a multiple of chunk size.

Issue #2139, opened 2019-08-30
OP-TEE: oe_is_outside_enclave() should not check for !oe_is_within_enclave()


david-oswald:
In the current (non-functional) OP-TEE binding, the check for `oe_is_outside_enclave()` (i.e. "is the buffer completely outside the enclave) internally uses `!oe_is_within_enclave()`:

https://github.com/openenclave/openenclave/blob/master/enclave/core/optee/bounds.c#L35

However, this will fail if a buffer overlaps from untrusted into trusted memory, since `oe_is_within_enclave()` checks if the *whole* buffers falls into trusted memory (so will return `false` if the buffer only overlaps into trusted memory). 

While it might be difficult to achieve such an overlapping situation with the shared-mem constructs in Trustzone, I think this should be still addressed before the OP-TEE binding is marked as functional.

(Btw: reporting this as a normal issue rather than opening a case with MSRC, since OP-TEE support is not finished to the best of my knowledge)

yakman2020:
Pls issue a PR. Looks real.

HernanGatta:
I've been thinking about this for a while, and it's both a problem and not a problem, depending on how one looks at it. From a purely practical perspective, it does not appear to be a problem. From a theoretical perspective, however, these functions are, at least, incorrectly named. @mikbras and @dthaler also have their thoughts about them.

----

The following is an initial pass at an analysis of the practical aspect of this question; another post will follow regarding the theoretical perspective.

Let a buffer be defined as a tuple whose members are a pointer that references a contiguous block of virtual memory, and a length indicating how far past the pointer the buffer's owner may read and/or write. The functions in question, `oe_is_within_enclave` and `oe_is_outside_enclave`, accept one such buffer as their argument and provide a response after inspecting it.

A buffer so defined may be obtained by a host or an enclave by a call to `malloc`. In both cases, `malloc` hands out blocks of contiguous virtual memory within a known virtual address space. This virtual address space is defined by Linux in the host's case, and by OP-TEE in the enclave's case. In turn, this virtual address space is backed by one or more physical pages, which may or may not be contiguous.

In ARM TrustZone, OP-TEE programs the TrustZone Address Space Controller (TZASC) during boot to partition the physical address space into secure and non-secure memory regions. The information as to where in physical memory these regions start and end is included in per-platform configuration files and is baked into OP-TEE at compile-time. Additionally, the secure region of memory is hidden from Linux by altering the system memory map accordingly before the former has a chance to read it.

[...]

PR #4178, opened 2021-07-21
Introducing security best practices doc

douglasmaciver:
A small, one-file commit that introduces best practices for keeping
enclaves secure. Use this guide to apply secure patterns and
avoid common mistakes that put the security of enclave
applications at risk.

Credit for early draft work to mingweishih.

Fixes openenclave#4031

Signed-off-by: Douglas MacIver <domac@microsoft.com>

### DIFF
--- a/docs/GettingStartedDocs/best_practices_secure_enclaves.md
+++ b/docs/GettingStartedDocs/best_practices_secure_enclaves.md
@@ -1,0 +1,380 @@
+# Best practices for keeping enclaves secure
+Use this guide to apply secure patterns and avoid common mistakes that put the
+security of enclave applications at risk.
+
+1. [Best practices for interface custom marshaling](#1-best-practices-for-interface-custom-marshaling)
+
+    1.1 [Ensuring memory is where it should be](#11-ensuring-memory-is-where-it-should-be)
+
+    1.2 [Memory bounds: oe_is_within_enclave vs. oe_is_outside_enclave](#12-Memory-bounds-oe_is_within_enclave-vs-oe_is_outside_enclave)
+
+    1.3 [TOCTOU or double fetch vulnerabilities](#13-toctou-or-double-fetch-vulnerabilities)
+
+    1.4 [Ensure proper bounding of data buffers](#14-ensure-proper-bounding-of-data-buffers)
+
+    1.5 [Updated enclave ecall with custom marshaling](#15-updated-enclave-ecall-with-custom-marshaling)
+
+2. [Handling secrets in enclave applications](#2-handling-secrets-in-enclave-applications)
+
+    2.1 [How _not_ to handle application secrets](#21-how-not-to-handle-application-secrets)
+
+    2.2 [The new secure enclave way to handle application secrets](#22-the-new-secure-enclave-way-to-handle-application-secrets)
+
+<br />
+<br />
+<br />
+
+# 1. Best practices for interface custom marshaling 
+(For an overview of using the Enclave Definition Language (EDL) and the
+oeedger8r tool to produce enclave interface code, please refer to
+[Getting started with the Open Enclave edger8r](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Edger8rGettingStarted.md).)
+
+Calling into and out of enclaves uses special methods that orchestrate the
+context switch and marshal the function parameters. Much of the code necessary
+to manage these calls and parameter marshaling are common to most applications.
+The Open Enclave _oeedger8r_, using the interface definition defined by
+application developers in EDL files, generates boilerplate code for them.
+
+In some uncommon cases, developers may want to pass data types that are not
+defined at the interface level or handle marshaling differently from what the
+oeedger8r tool generates. Although this is not generally recommended, it is
+supported. To specify this, use `user_check` for the parameter
+constraint like the following example:
+
+```c++
+enclave {
+    trusted {
+        public void ecall_with_user_check(
+            [user_check] void* blob);
+    };
+};
+```
+In this simplified EDL example, the function takes one pointer-to-void
+parameter. Remember that the primary benefit of using supported types in the EDL
+is that the generated boilerplate code performs the necessary and secure
+marshaling for the developer. When developers specify that a parameter is
+`user_check`, they are signaling to oeedger8r that their application code will
+perform the marshaling. (Two examples of when custom marshaling is helpful
+are 1) when data formats are dynamic, and 2) when sharing large blocks of
+memory&mdash;custom marshaling avoids the boilerplate code's intermediate copying of data
+into safe enclave buffers.)
+
+> We'd like to stress that a secure best practice for enclave interfaces is to
+_avoid_ custom marshaling. But we understand that there may be cases where
+avoiding it is impractical, so in these sections we will help developers do it
+safely.
+
+The ecall function implementation below has several problems. We will be
+correcting this code as we move through the material. The code also assumes that the enclave
+can directly read the blob memory, which is true for SGX but may not be true for other TEEs.  The general impetus for
+custom marshaling in this sample code is to avoid expensive intermediate copying
+of shared memory introduced by the oeedger8r's generated code. (Data passed to
+`render_data()` is "flat", that is, the data format is non-referential and thus
+non-_self_-referencing. These attributes reduce threats to enclaves from
+untrusted data streams. More on this in the following sections.)
+
+```c++
+typedef struct _blob {
+    void* data;
+    size_t size;
+} blob_t;
+
+// WARNING: Portions of this code are intentionally flawed to demonstrate common
+// pitfalls.
+int ecall_with_user_check0(void* ptr) {
+
+    blob_t* blob;
+
+    if (ptr == nullptr)
+        return -1;
+
+    blob = (blob_t*)ptr;
+
+    // Blobs to be rendered are limited to 4K.
+    if (blob->size > 4096)
+        return -1;
+
+    return render_data(blob->data, blob->size);
+}
+```
+
+## 1.1 Ensuring memory is where it should be
+
+Remember that the code we are focused on is running _inside_ of the enclave.
+The caller invoked this ecall function from _outside_ the enclave, passing in a
+pointer to memory containing well-formed data (in the non-malicious case).
+But in the Open Enclave security model, the ecall invoker is untrusted&mdash;we must
+treat untrusted input with caution.
+
+One important security check is to ensure that memory buffers are on the correct
+side of the security boundary. The Open Enclave SDK provides two library
+functions for checking memory buffers:
+`oe_is_within_enclave()` and `oe_is_outside_enclave()`.
+
+Let's update the code to ensure that the memory buffer is at least the size of a
+`blob_t` and is located _strictly_ outside the secured enclave memory
+before operating on it by calling `oe_is_outside_enclave()`.
+
+```diff
+// WARNING: Portions of this code are intentionally flawed to demonstrate common
+// pitfalls.
+int ecall_with_user_check1(void* ptr) {
+
+    blob_t* blob;
+
+-   if (ptr == nullptr)
+-       return -1;
++   // Ensure passed-in pointer is not null and the memory buffer is located
++   // strictly outside of the secured enclave memory.
++   if (!oe_is_outside_enclave(ptr, sizeof(blob_t)))
++       return -1;
+
+    blob = (blob_t*)ptr;
+
+    // Blobs to be rendered are limited to 4K.
+    if (blob->size > 4096)
+        return -1;
+
+    return render_data(blob->data, blob->size);
+}
+```
+
+## 1.2 Memory bounds: oe_is_within_enclave vs. oe_is_outside_enclave
+
+Let's highlight a logical mistake that we've seen made in enclave applications.
+Some have regarded the memory validation functions as boolean opposites of each
+other, that is: `oe_is_within_enclave() == !oe_is_outside_enclave()`. This is
+incorrect for several reasons.
+
+To understand why these functions are _not_ boolean opposites, let's examine the
+Open Enclave implementation for SGX as an example:
+[memory.c](/enclave/core/sgx/memory.c).
+There are three conditions that the functions validate:
+1)	The pointer is not null.
+2)	The bounding arithmetic operations do not wrap (that is, numerical
+overflow).
+3)	The buffer lies strictly within or outside of the secured enclave memory.
+
+The first two conditions should make clear the pitfalls of considering the
+function pair to be boolean opposites: If the pointer is null or the bounding
+calculations overflow in a call to `oe_is_within_enclave()`, the function will
+return `false`, which should _not_ be interpreted as "the buffer is strictly outside the
+secured enclave memory" or `oe_is_outside_enclave() == true`.
+
+The last condition is subtle: In some potentially malicious cases, the buffer
+may be both partially within _and_ partially outside the secured enclave
+memory, spanning the boundary between the two. (Attacks against enclaves
+may use memory confusion to achieve overwrites of secured enclave memory, for
+example.) Both functions will return `false` when the memory range spans the
+boundary - another reason the functions are not boolean opposites.
+
+<br />
+<br />
+<br />
+
+## 1.3 TOCTOU or double fetch vulnerabilities
+
+One class of vulnerabilities that custom marshaling code should protect against
+is referred to as time-of-check/time-of-use, or TOCTOU. Another name for this
+problem is "double-fetch". The problem arises when memory is shared across
+security boundaries. As code validates input, it's critical that data is not
+fetched twice (or, strictly, more than once), allowing a malicious untrusted
+caller to change the data between the time-of-check and the time-of-use. Let's
+examine our function:
+
+```c++
+// WARNING: Portions of this code are intentionally flawed to demonstrate common
+// pitfalls.
+int ecall_with_user_check2(void* ptr) {
+
+    blob_t* blob;
+
+    // Ensure passed-in pointer is not null and the memory buffer is located
+    // strictly outside of the secured enclave memory.
+    if (!oe_is_outside_enclave(ptr, sizeof(blob_t)))
+        return -1;
+
+    blob = (blob_t*)ptr;
+
+    // Blobs to be rendered are limited to 4K.
+    if (blob->size > 4096) //TOCTOU: First fetch
+        return -1;
+
+    return render_data(blob->data, blob->size); //TOCTOU: Second fetch
+}
+```
+There's nothing wrong with the first fetch: The value is validated, and the
+appropriate logic branch is taken. The problem arises when the code needs the
+value again and reads it, once again, from the untrusted memory location outside
+of the secured enclave memory. Between the first and subsequent fetch the value may have been
+changed, neutralizing the size validation and possibly leading to some enclave
+memory corruption that is helpful to the attacker. Let's protect the enclave by
+"capturing" the values.
+
+> When "capturing" data values in this context, it's important to protect
+against the compiler's optimization, hence we use the `volatile` qualifier for
+our local captured variables. Otherwise, the compiler might optimize-away our
+local variables, removing the TOCTOU protection.
+
+```diff
+// WARNING: Portions of this code are intentionally flawed to demonstrate common
+// pitfalls.
+int ecall_with_user_check3(void* ptr) {
+
+    blob_t* blob;
++   void* volatile captured_data;
++   volatile size_t captured_size;
+
+    // Ensure passed-in pointer is not null and the memory buffer is located
+    // strictly outside of the secured enclave memory.
+    if (!oe_is_outside_enclave(ptr, sizeof(blob_t)))
+        return -1;
+
+    blob = (blob_t*)ptr;
+
++   // Capture blob descriptors to avoid TOCTOU problems.
++   captured_data = blob->data;
++   captured_size = blob->size;
+
+    // Blobs to be rendered are limited to 4K.
+-   if (blob->size > 4096) //TOCTOU: First fetch
++   if (captured_size > 4096)
+        return -1;
+
+-   return render_data(blob->data, blob->size); //TOCTOU: Second fetch
++   return render_data(captured_data, captured_size);
+}
+```
+
+## 1.4 Ensure proper bounding of data buffers
+
+Another element of marshaling code is ensuring, as complex structures are
+parsed, that nested structures are also within the proper bounds. This is
+especially important for internal functions like `render_data()` that may be
+unaware that the memory is untrusted. Use the same techniques on the inner
+structures that were used on the outer.
+
+> A warning about legacy data-parsers: Take care when passing "hot" data to code
+that may not have been written to parse maliciously crafted input. We have seen
+cases where legacy code that parses self-formatting or self-referencing data was
+used in new enclave applications. This can lead to significant vulnerabilities
+if the data being parsed is still controlled by the ecall invoker, as may be the case
+with custom marshaling. In the case of this sample code, as mentioned earlier,
+`render_data()` parses "flat" data that is not nested nor self-referencing, so
+it's safe (and performant) to pass it "hot" data that is controlled by the caller.
+Care must be taken though, to avoid data-consistency and other app-level
+integrity problems.
+
+```diff
+int ecall_with_user_check4(void* ptr) {
+
+    blob_t* blob;
+    void* volatile captured_data;
+    volatile size_t captured_size;
+
+    // Ensure passed-in pointer is not null and the memory buffer is located
+    // strictly outside of the secured enclave memory.
+    if (!oe_is_outside_enclave(ptr, sizeof(blob_t)))
+        return -1;
+
+    blob = (blob_t*)ptr;
+
+    // Capture blob descriptors to avoid TOCTOU problems.
+    captured_data = blob->data;
+    captured_size = blob->size;
+
+    // Blobs to be rendered are limited to 4K.
+    if (captured_size > 4096)
+        return -1;
+
++   // Ensure validity of nested structure
++   if (!oe_is_outside_enclave(captured_data, captured_size))
++       return -1;
+
++   // Data parsed by render_data is "flat", non-self-referencing.
+    return render_data(captured_data, captured_size);
+}
+```
+## 1.5 Updated enclave ecall with custom marshaling
+
+Thanks for following along with us. Here's our updated ecall function that
+marshals the parameter correctly for enclaves:
+```c++
+int ecall_with_user_check5(void* ptr) {
+
+    blob_t* blob;
+    void* volatile captured_data;
+    volatile size_t captured_size;
+
+    // Ensure passed-in pointer is not null and the memory buffer is located
+    // strictly outside of the secured enclave memory.
+    if (!oe_is_outside_enclave(ptr, sizeof(blob_t)))
+        return -1;
+
+    blob = (blob_t*)ptr;
+
+    // Capture blob descriptors to avoid TOCTOU problems.
+    captured_data = blob->data;
+    captured_size = blob->size;
+
+    // Blobs to be rendered are limited to 4K.
+    if (captured_size > 4096)
+        return -1;
+
+    // Ensure validity of nested structure
+    if (!oe_is_outside_enclave(captured_data, captured_size))
+        return -1;
+
+    // Data parsed by render_data is "flat", non-self-referencing.
+    return render_data(captured_data, captured_size);
+}
+```
+
+<br />
+<br />
+<br />
+
+# 2. Handling secrets in enclave applications
+
+Open Enclave is an SDK that helps developers build apps that will run inside a
+hardware-based Trusted Execution Environment (TEE). At their core, TEEs protect
+application code and data at runtime from the host environment. Without new,
+hardware-implemented protections, a malicious or compromised host
+would be able to modify code or read data. Any secrets managed by an
+application during runtime would be at risk of exposure. The Open Enclave SDK
+helps developers build applications that are protected from these threats.
+
+> It's important to note that enclaves do not generally provide
+_code confidentiality_. This is noteworthy in the context of this discussion
+because, as will be discussed, hard-coded credentials in application code are to
+be avoided&mdash;with or without enclave protections.
+
+Enclaves provide new trust boundary protections that address old threats and
+open new capabilities. But just as Open Enclave and TEEs shift the security
+model, application developers must shift how they handle data, especially
+secrets, like keys and passwords.
+<br />
+<br />
+<br />
+
+## 2.1 How _not_ to handle application secrets
+Let's remind ourselves of a way _not_ to handle application secrets. A common
+example of this well-discussed application weakness is described by
+[CWE-798: Use of Hard-coded Credentials (4.4) (mitre.org)](https://cwe.mitre.org/data/definitions/798.html).
+While application security is often a tradeoff between multiple factors and
+broad edicts are not always appropriate, it's not controversial to simply say
+that hard-coding credentials in application code _should not be done_.
+<br />
+<br />
+<br />
+
+## 2.2 The new, secure enclave way to handle application secrets
+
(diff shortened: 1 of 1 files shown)


docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,275 @@
...
+    return render_data(blob->data, blob->size);
+}
+```
+## 1.2 SGX enclaves: oe_is_within_enclave vs. oe_is_outside_enclave
+
+Let's highlight a logical mistake that we've seen made with SGX-based enclaves. Some have regarded the memory validation functions as boolean opposites of each other, that is: `oe_is_within_enclave() == !oe_is_outside_enclave()`. This is incorrect for several reasons.
+
+To understand why this function pair are _not_ boolean opposites, it helps to consider the SGX implementation: [memory.c](https://github.com/openenclave/openenclave/blob/08ebe60c1d2e3ea24ac634c673a420296bff3352/enclave/core/sgx/memory.c). There are three conditions that the functions validate:
+1)	The pointer is not null.
+2)	The bounding arithmetic operations do not wrap (i.e. numerical overflow).
```
dthaler, line 106:
```suggestion
2)	The bounding arithmetic operations do not wrap (i.e., numerical overflow).
```
nit: always put comma after "i.e." per Chicago Manual of Style

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,275 @@
...
+}
+```
+
+## 1.1 Ensuring memory is where it should be
+
+Remember that the code we are focused on is running _inside_ of the enclave. The host caller invoked this function from _outside_ the enclave, passing in a pointer to host memory containing well-formed data (in the non-malicious case). But in the Open Enclave security model, the host caller is untrusted - we must treat untrusted input with caution. 
+
+One important security check is to ensure that memory blocks are on the correct side of the security boundary. It's common for host applications to pass host memory blocks to the enclave, as is the case with this example code. The Open Enclave SDK provides two library functions for checking memory buffers: `oe_is_within_enclave()` and `oe_is_outside_enclave()`.
+
+Let's update the code to ensure that the memory block is at least the size of a `blob_t` and is located _strictly_ outside the secured enclave memory region before operating on it by calling `oe_is_outside_enclave()`.
```
dthaler, line 77: This section is not labeled as SGX specific, but the discussion appears to make an SGX-specific assumption that host memory can be directly read by the enclave.  This is not true in general, where line 94 below is unsafe because the memory pointed to by blob might not be readable from the encave.   If this intends to talk specifically about SGX again, then it should say so, like section 1.2 does.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,275 @@
...
++   return render_data(captured_data, captured_size);
+}
+```
+
+## 1.4 Ensure proper bounding of data buffers
+
+Another element of marshaling code is ensuring, as complex structures are parsed, that nested structures are also within the outer bounds. This is especially important for internal functions like `render_data()` that may be unaware that the memory is untrusted. Use the same techniques on the inner structures that were used on the outer.
+
+> A warning about legacy data-parsers: Take care when passing "hot" data to code that may not have been written to parse maliciously crafted input. We have seen cases where legacy code that parses self-formatting or self-refencing data was used in new enclave applications. This can lead to significant vulnerabilities if the data being parsed is still controlled by the host, as may be the case with custom marshaling. In the case of this sample code, as mentioned earlier, `render_data()` parses "flat" data that is not nested nor self-referencing, so it's safe (and performant) to pass it "hot" data that is controlled by the host. Care must be taken though, to avoid data-consistency and other app-level integrity problems.
```
dthaler, line 176:
```suggestion
> A warning about legacy data-parsers: Take care when passing "hot" data to code that may not have been written to parse maliciously crafted input. We have seen cases where legacy code that parses self-formatting or self-referencing data was used in new enclave applications. This can lead to significant vulnerabilities if the data being parsed is still controlled by the host, as may be the case with custom marshaling. In the case of this sample code, as mentioned earlier, `render_data()` parses "flat" data that is not nested nor self-referencing, so it's safe (and performant) to pass it "hot" data that is controlled by the host. Care must be taken though, to avoid data-consistency and other app-level integrity problems.
```

also, split source lines into shorter lines.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,275 @@
...
++   if (!oe_is_outside_enclave(captured_data, captured_size))
++       return -1;
+
++   // Data parsed by render_data is "flat", non-self-referencing.
+    return render_data(captured_data, captured_size);
+}
+```
+## 1.5 Corrected enclave ecall with custom marshaling
+
+Thanks for sticking with us. Here is our corrected ecall function that marshals the parameter correctly:
```
dthaler, line 209: It's only corrected for SGX, not for other TEEs because line 224 etc still assumes it can read host memory.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,275 @@
...
+}
+```
+
+<br />
+<br />
+<br />
+
+# 2. Handling Secrets in Enclave Applications
+
+Open Enclave is an SDK that helps developers build apps that will run inside a hardware-based Trusted Execution Environment (TEE). At their core, TEEs protect application code and data at runtime from the host environment. Without new, hardware-implemented protections, a malicious or compromised host operating system would be able to modify code or read data. Any secrets managed by an application during runtime would be at risk of exposure. Open Enclave SDK helps developers build applications that are protected from these threats.
```
dthaler, line 246:
```suggestion
Open Enclave is an SDK that helps developers build apps that will run inside a hardware-based Trusted Execution Environment (TEE). At their core, TEEs protect application code and data at runtime from the host environment. Without new, hardware-implemented protections, a malicious or compromised host operating system would be able to modify code or read data. Any secrets managed by an application during runtime would be at risk of exposure. The Open Enclave SDK helps developers build applications that are protected from these threats.
```

Grammar nit.  Also again, please split long lines into multiple lines in the source file.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,275 @@
...
+
+> It's important to note that enclaves do not provide _code confidentiality_. This is noteworthy in the context of this discussion because, as will be discussed, hard-coded credentials in application code are to be avoided -- with or without enclave protections.
```
dthaler, line 248:
```suggestion
> It's important to note that enclaves do not generally provide _code confidentiality_. This is noteworthy in the context of this discussion because, as will be discussed, hard-coded credentials in application code are to be avoided -- with or without enclave protections.
```

Add "generally" since there are ways to do it by loading an encrypted code module into an enclave.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Secure Key Release
+One of the core challenges customers have interacting with encrypted environments is how to ensure that they can securely and reliably communicate with the code running in the environment ("enclave code").
+
+One solution to this problem is what is known as "Secure Key Release", which is a pattern that enables this kind of communication with enclave code.
+
+To implement the "Secure Key Release" pattern, the enclave code generates an ephemeral asymmetric key. It then serializes the public portion of the key to some format (possibly a JSON Web Key, or PEM, or some other serialization format).
+The enclave code then calculates the SHA256 value of the public key and passes it as an input to code which generates an SGX Quote. For Open Enclave, this is done with `oe_get_evidence` or `oe_get_report`.
```
dthaler, line 271:
```suggestion
The enclave code then calculates a hash (e.g., SHA256) value of the public key and passes it as an input to code which generates attestation evidence. For Open Enclave, this can be done with `oe_get_evidence`.
```
Avoid SGX-specific terminology and references to deprecated/removed APIs.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,275 @@
...
+
+The client then sends the SGX quote and the serialized key to the attestation service. The attestation service will validate the quote and ensure that the hash of the key is present in the quote and will issue an "Attestation Token".
```
dthaler, line 273:
```suggestion
The client then sends the evidence and the serialized key to an attestation service. The attestation service will appraise the evidence and ensure that the hash of the key is present in the evidence and will issue an Attestation Result (sometimes called an "Attestation Token").
```

I'd prefer to delete "Attestation Token" as a capitalized term as Attestation Result is the standard term, and "Attestation Token" does not appear in any OE header or other document.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,275 @@
...
+
+The client can then send that Attestation Token (which contains the serialized key) to a 3rd party "relying party". The relying party then validates that the attestation token was created by the attestation service, and thus the serialized key can be used to encrypt some data held by the "relying party" to send to the service.
```
dthaler, line 275: This statement is confusing.  Why would the Attestation Result need to contain the serialized key?   The serialized key can be obtained or negotiated via the protocol between the attester and the relying party once the RP accepts the attestation token.  Why would you want to reveal the serialized key to the attestation service?   This seems less secure (doesn't follow the need to know principle.)   Given this is a security best practices document, I think any recommendation to reveal the key to the attestation service should be removed.

dthaler (dismissed):

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,275 @@
...
+
+```c++
+enclave{
+    trusted {
+        public void ecall_with_user_check(
+            [user_check] void* blob);
+    };
+};
+```
+In this simplified EDL example, the function takes one pointer-to-void parameter. Remember that the primary benefit of using supported types in the EDL is that the generated boiler-plate code performs the necessary and secure marshaling for the developer. When developers specify that a parameter is `user_check`, they are signaling to oeedger8r that their application code will perform the marshaling. (Two examples of when custom marshaling is helpful are 1) when data formats are dynamic, and 2) when sharing large blocks of memory -- custom marshaling avoids the boiler-plate code's intermediate copying of data into safe in-enclave buffers.)
```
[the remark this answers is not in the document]
douglasmaciver: Good catch on boilerplate. Updated.
Limited Markdown text lines to 80, with the exception of links.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: Yes! The Chicago Manual of Style.
The Microsoft Style Guide also states not to use "i.e.": https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/i/ie

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: It's better to generalize this section, so I removed host-specific references.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: Fixed.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: I added SGX to the description:
"Thanks for sticking with us. Here is our corrected ecall function that marshals
the parameter correctly for SGX-based enclaves:"

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: Fixed.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: Added "generally".

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: I took your edits, verbatim.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: I took your edits, verbatim.

douglasmaciver:
@dthaler Thank you for your careful review. 

(Apologies for Markdown text that exceeded the 80 column margin -- I should have caught that.  The downside of that mistake is that correcting it caused a lot of churn, clouding the other small edits that I made to address your minor comments. I just amended the commit instead of adding a new one, given that the delta was so great.)

I would like to continue discussing the two remaining unresolved topics.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: I'm beginning to waffle on this. Perhaps we should make the entire Custom Marshalling section explicitly SGX focused?

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: The high-level intent of this topic was to help Enclave developers envision a secure way to handle secrets and Secure Key Release was a prime model. We had not planned on covering the topic in detail here -- we thought we would simply reference existing Open Enclave SDK Secure Key Release documentation, but we could not find any. What we did find was SGX-focused documentation under the Azure Attestation docs: https://docs.microsoft.com/en-us/dotnet/api/overview/azure/security.attestation-readme#attest-sgx-enclave

The text in section 2.2 of our doc is almost verbatim from the above source, so corrections will have to be applied there also.  Would you help me rewrite this short introduction here? I'll submit a doc request for SKR against OE, and submit bugs against the multiple Azure Attestation pages that have this incorrect info.

dthaler: I'd prefer to keep the best practices doc generic (not specific to Azure) and cover "best" practices.  The azure doc you link just states it's current practice, it does not assert it is "best" practice as a model to follow, just a model.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,403 @@
...
+
+Calling into and out of enclaves uses special methods that orchestrate the
+context switch and marshal the function parameters. Much of the code necessary
+to manage these calls and parameter marshaling are common to most applications.
+The Open Enclave _oeedger8r_, using the interface definition defined by
+application developers in EDL files, generates boilerplate code for them.
+
+In some uncommon cases, developers may want to pass data types that are not
+defined at the interface level or handle marshaling differently from what the
+oeedger8r tool generates. To specify this, use `user_check` for the parameter
```
dthaler, line 40:
```suggestion
oeedger8r tool generates. Although this is not generally recommended, it is supported.
To specify this, use `user_check` for the parameter
```

Suggested text is my proposal to make it clear that doing so is not being stated to be a "best practice", but if you DO do it, then this doc covers the best practice on HOW.  Line 61-62 already does so, but this suggestion makes it clear up front for consistency in message.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,403 @@
...
+}
+```
+## 1.2 SGX enclaves: oe_is_within_enclave vs. oe_is_outside_enclave
+
+Let's highlight a logical mistake that we've seen made with SGX-based enclaves.
+Some have regarded the memory validation functions as boolean opposites of each
+other, that is: `oe_is_within_enclave() == !oe_is_outside_enclave()`. This is
+incorrect for several reasons.
+
+To understand why this function pair are _not_ boolean opposites, it helps to
```
dthaler, line 146:
```suggestion
To understand why these functions are _not_ boolean opposites, it helps to
```
(fix grammar mismatch in "this ... are")

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,403 @@
...
+calculations overflow in a call to `oe_is_within_enclave()`, the function will
+return `false`, which should _not_ be interpreted as "the memory is outside the
+enclave".
+
+The last condition is subtle: In some potentially malicious cases, the memory
+range may be both partially within and partially outside the enclave, spanning
+the boundary between the two. (Attacks against enclaves may use memory confusion
+to achieve overwrites of protected regions, for example.) Both functions will
+return `false` when the memory range spans the boundary - another reason the
+function pair are not boolean opposites.
```
dthaler, line 166:
```suggestion
functions are not boolean opposites.
```

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,403 @@
...
++       return -1;
+
++   // Data parsed by render_data is "flat", non-self-referencing.
+    return render_data(captured_data, captured_size);
+}
+```
+## 1.5 Corrected enclave ecall with custom marshaling
+
+Thanks for sticking with us. Here is our corrected ecall function that marshals
+the parameter correctly for SGX-based enclaves:
```
dthaler, line 298:
```suggestion
Thanks for sticking with us. Here is our updated ecall function that marshals
the parameter correctly for SGX-based enclaves (but is still incorrect for other TEEs
that do not allow enclave code to directly read host memory):
```

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,403 @@
...
+
++   // Ensure validity of nested structure
++   if (!oe_is_outside_enclave(captured_data, captured_size))
++       return -1;
+
++   // Data parsed by render_data is "flat", non-self-referencing.
+    return render_data(captured_data, captured_size);
+}
+```
+## 1.5 Corrected enclave ecall with custom marshaling
```
dthaler, line 295:
```suggestion
## 1.5 Updated enclave ecall with custom marshaling
```

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,403 @@
...
+
+    // Ensure passed-in pointer is not null and the memory block is located
+    // entirely outside of the enclave.
+    if (!oe_is_outside_enclave(ptr, sizeof(blob_t)))
+        return -1;
+
+    blob = (blob_t*)ptr;
+
+    // Capture blob descriptors to avoid TOCTOU problems.
+    captured_data = blob->data;
```
dthaler, line 314:
```suggestion
    // This line assumes enclave code can directly read host memory.
    captured_data = blob->data;
```

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,403 @@
...
+The client then sends the evidence and the serialized key to an attestation
+service. The attestation service will appraise the evidence and ensure that the
+hash of the key is present in the evidence and will issue an Attestation Result
+(sometimes called an "Attestation Token").
+
+The client can then send that Attestation Token (which contains the serialized
+key) to a 3rd party "relying party". The relying party then validates that the
+attestation token was created by the attestation service, and thus the
+serialized key can be used to encrypt some data held by the "relying party" to
+send to the service.
```
dthaler, line 403: Besides the other problems with this text that I raised, another problem is that the description doesn't match the name.
The name would imply that that a key is only released to an enclave after attestation completes, so it's known to be handed to a trusted client.   In contrast, lines 386-387 instead have the (not yet trusted) enclave _create_ the key and then pass it to the relying party which only _uses_ it if attestation passes.  Thus, it's more like a secure key use than a secure key release.   The only way secure key release would be an accurate name is if you're talking about releasing it to the relying party, but the relying party need not be running in an enclave, and so "secure" doesn't mean the same as what it means in the rest of this article.

rs--:
bors try

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: After some mulling, the proper way to take this guidance is to make it general and hw-agnostic, following the spirit of Open Enclave.  The challenge though, is the potential mixed semantics of the memory models: This is evident in the different memory models of SGX and OP-TEE. When both are considered, it's difficult to choose the proper abstraction: SGX is satisfied by "outside" or "within" the enclave, but with OP-TEE it's "secure" or "nonsecure" -- the memory buffer could be shared, and thus neither necessarily "outside" nor "within" the enclave. 

BTW, I've always been bothered by the OE OP-TEE implementation of `oe_is_outside_enclave()`, and now I'm convinced it's incorrect in ways similar to why the SGX implementations are not boolean opposites. @dthaler, what do you think?
https://github.com/openenclave/openenclave/blob/cd72fd7069488ba6f453c8f5f47bd9fd9a6e6c0d/enclave/core/optee/bounds.c#L33-L36

Let me attempt to generalize the text in light of the comments above.

dthaler: Most of the points in the doc apply equally to SGX, OP-TEE, and any other TEEs, so it's good to keep the doc general where possible and just use TEE-specific examples.  For OP_TEE in particular, the discussion in #2139 is relevant to this doc... is_within_enclave() still makes sense as being secure, non-shared.  is_outside_enclave() is more nuanced as discussed in that issue.

dthaler: My personal view is that `oe_is_outside_enclave()` is a confusingly named API to begin with (for both SGX and OP-TEE) for the reasons articulated in the PR and in #2139.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: For the scope of this best practice guidance, I will introduce the Secure Key Release design pattern and exclude implementation details. When SKR for Open Enclave is documented, a link can be included here.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: When I generalize the memory bounds discussion, I'm not going mention "host", which, as you have pointed out, is SGX-centric. Instead, memory buffers will be referred generically as either "trusted" or "untrusted" from the point of view of the enclave. Whether memory is strictly "within" or "outside" the enclave doesn't apply to all hardware implementations. (It is unfortunate that these function names don't map well to this generalized Open Enclave memory model.) Data in buffers that are "untrusted" -- or "outside" -- the enclave must be handled with care. The untrusted buffer contents may be maliciously constructed and may change from read to read.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: I will adopt "updated", but I have to change the table of contents too, so I'll handle it my broader update.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: I'm going to drop the "host" concept. See my response : https://github.com/openenclave/openenclave/pull/4178#discussion_r675225002

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: I agree. Trying to use existing "Secure Key Release" material did not work for this context. I removed the offending material and briefly sketched how attestation could be used with secure data services to manage secrets. It's difficult to go into detail and remain cloud service provider agnostic.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: Please see the related conversation here: https://github.com/openenclave/openenclave/pull/4178#discussion_r675261490

docs/GettingStartedDocs/best_practices_secure_enclaves.md
dthaler: Lines 387-404 are "the offending material".  Are you removing those lines in another commit?

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: 71be4bcb generalizes the text covering memory models.  I'd appreciate your continued review @dthaler.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: fixed in  71be4bcb,

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: Addressed in  71be4bcb.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: Addressed in 71be4bcb. I removed design details of the application secrets section. I can see the benefit of a detailed description of how to perform Secure Key Release (for example) in a cloud-agnostic way, but it's beyond the scope of this doc. I would appreciate your review @dthaler.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,380 @@
...
+1)	The pointer is not null.
+2)	The bounding arithmetic operations do not wrap (that is, numerical
+overflow).
+3)	The buffer lies completely within or outside of the secured enclave memory
+region.
+
+The first two conditions should make clear the pitfalls of considering the
+function pair to be boolean opposites: If the pointer is null or the bounding
+calculations overflow in a call to `oe_is_within_enclave()`, the function will
+return `false`, which should _not_ be interpreted as "the buffer is outside the
```
dthaler, line 161:
```suggestion
return `false`, which should _not_ be interpreted as "the buffer is strictly outside the
```
for consistency with line 114.  Or use "entirely" for consistency with line 229.  For bonus points, make 114 and 229 use the same word.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,380 @@
...
+custom marshaling avoids the boilerplate code's intermediate copying of data
+into safe enclave buffers.)
+
+> We'd like to stress that a secure best practice for enclave interfaces is to
+_avoid_ custom marshaling. But we understand that there may be cases where
+avoiding it is impractical, so in these sections we will help developers do it
+safely.
+
+The ecall function implementation below has several problems. We will be
+correcting this code as we move through the material. The general impetus for
```
dthaler, line 68:
```suggestion
correcting this code as we move through the material. The code also assumes that the enclave
can directly read the blob memory, which is true for SGX but may not be true for other TEEs.  The general impetus for
```

docs/GettingStartedDocs/best_practices_secure_enclaves.md
dthaler: The word was dropped but not the concept.  See suggested fix in line 68

douglasmaciver: The concept shifted from "host" memory to "shared" memory. OP-TEE uses "shared" memory also, it just happens to be mapped into its address space and marked as TEE_MEMORY_ACCESS_NONSECURE.  (As I understand it -- I'm not an OP-TEE authority.) For OP-TEE, the fact that it is "within" the enclave is what makes generically discussing the OE memory model difficult, especially with functions that focus on "within" and "outside" by name.

dthaler: Some TEEs do not support shared memory at all.   OP-TEE requires the host side to opt-in to sharing the memory with the enclave.   I believe for user-check this is not done for OP-TEE, so anything marked user-check cannot be dereferenced unless is_inside_enclave is true.

douglasmaciver: Okay, I will take the suggested edits. Thanks for your help Dave.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: Again, good catch.  I adopted "strictly" throughout the doc to align with the Open Enclave API documentation.

douglasmaciver: I also reduced "memory region" to "memory".

dthaler (approved):

douglasmaciver:
bors try

rs--:
@radhikaj can we get someone to have a look at this PR for the second review? Thanks

douglasmaciver:
@radhikaj I nominate @mingweishih as the second reviewer. He contributed to an early version of the doc.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,380 @@
+# Best practices for keeping enclaves secure
+Use this guide to apply secure patterns and avoid common mistakes that put the
+security of enclave applications at risk.
+
+1.  [Best practices for interface custom marshaling ](#1-best-practices-for-interface-custom-marshaling)
```
mingweishih, line 5:
```suggestion
1. [Best practices for interface custom marshaling](#1-best-practices-for-interface-custom-marshaling)
```

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,380 @@
...
+application developers in EDL files, generates boilerplate code for them.
+
+In some uncommon cases, developers may want to pass data types that are not
+defined at the interface level or handle marshaling differently from what the
+oeedger8r tool generates. Although this is not generally recommended, it is
+supported. To specify this, use `user_check` for the parameter
+constraint like the following example:
+
+```c++
+enclave{
```
mingweishih, line 45:
```suggestion
enclave {
```

docs/GettingStartedDocs/best_practices_secure_enclaves.md
```diff
@@ -0,0 +1,380 @@
...
+## 1.2 Memory bounds: oe_is_within_enclave vs. oe_is_outside_enclave
+
+Let's highlight a logical mistake that we've seen made in enclave applications.
+Some have regarded the memory validation functions as boolean opposites of each
+other, that is: `oe_is_within_enclave() == !oe_is_outside_enclave()`. This is
+incorrect for several reasons.
+
+To understand why these functions are _not_ boolean opposites, let's examine the
+Open Enclave implementation for SGX as an example:
+[memory.c](https://github.com/openenclave/openenclave/blob/08ebe60c1d2e3ea24ac634c673a420296bff3352/enclave/core/sgx/memory.c).
```
mingweishih, line 151:
```suggestion
[memory.c](/enclave/core/sgx/memory.c).
```

mingweishih (approved):
LGTM. Thanks for putting the information together. Approve with little suggestions on the format.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: Fixed in latest commit.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: Fixed in latest commit.

docs/GettingStartedDocs/best_practices_secure_enclaves.md
douglasmaciver: Fixed in latest commit.

PR #4179, opened 2021-07-21
Add missing MUSL search, string, stdlib functions.

anakrish:
Also fix bugs in malloc/debug-malloc implementations
- errno was not being set correctly in all cases.
- oe_posix_memalign could end up making a recursive call to itself with debug malloc turned on.

Also fix missing dependencies of report_enc and openssl-common targets.

### DIFF
--- a/libc/CMakeLists.txt
+++ b/libc/CMakeLists.txt
@@ -274,9 +274,9 @@ add_enclave_library(
   ${MUSLSRC}/locale/uselocale.c
   ${MUSLSRC}/locale/wcscoll.c
   ${MUSLSRC}/locale/wcsxfrm.c
-  ${MUSLSRC}/linux/getrandom.c
   ${MUSLSRC}/linux/epoll.c
   ${MUSLSRC}/linux/flock.c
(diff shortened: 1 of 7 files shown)


anakrish:
bors try

libc/CMakeLists.txt
```diff
@@ -600,16 +600,19 @@ add_enclave_library(
   ${MUSLSRC}/search/lsearch.c
   ${MUSLSRC}/select/select.c
   ${MUSLSRC}/select/poll.c
-  ${MUSLSRC}/stat/mkdir.c
+  ${MUSLSRC}/search/hsearch.c
```
mingweishih, line 603: The test is currently disabled: https://github.com/openenclave/openenclave/blob/82a698498910332eb637af80b542d2e061ee3a47/tests/libc/tests.cmake#L318

anakrish: Added test. Also fixed a bug with debugmalloc not setting errno that the test revealed.

mingweishih (approved):

anakrish:
bors try

radhikaj (approved):

PR #4195, opened 2021-07-27
added layered attesation test for enclave

pragyaaneja:


radhikaj:
@pragyaaneja  Can you please add a description of what this is doing. Is this intended to be run in CI?

PR #4196, opened 2021-07-28
Support compiler-rt

### LINKED ISSUE #1766: Add support compiler runtimes in the enclave

anakrish:
The object library is included in oelibc.
Use the compiler-rt sources published in a separate archive as part of LLVM 10.0.1 release.

fixes #1766 

### DIFF
--- a/3rdparty/compiler-rt/CMakeLists.txt
+++ b/3rdparty/compiler-rt/CMakeLists.txt
@@ -1,0 +1,248 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+include(FetchContent)
+FetchContent_Declare(
+  compiler-rt-sources
+  SOURCE_DIR
+  ${CMAKE_CURRENT_SOURCE_DIR}/compiler-rt
+  URL https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.1/compiler-rt-10.0.1.src.tar.xz
+  URL_HASH
+    SHA256=d90dc8e121ca0271f0fd3d639d135bfaa4b6ed41e67bd6eb77808f72629658fa)
+
+FetchContent_GetProperties(compiler-rt-sources)
+if (NOT compiler-rt-sources_POPULATED)
+  FetchContent_Populate(compiler-rt-sources)
+endif ()
+
+set(COMPILER_RT_SRC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/compiler-rt/lib/builtins/")
+
+# List based on compiler-rt/lib/builtins/CMakeLists.txt
+set(GENERIC_SOURCES
+    absvdi2.c
+    absvsi2.c
+    absvti2.c
+    adddf3.c
+    addsf3.c
+    addtf3.c
+    addvdi3.c
+    addvsi3.c
+    addvti3.c
+    ashldi3.c
+    ashlti3.c
+    ashrdi3.c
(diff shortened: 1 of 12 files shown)


anakrish:
bors try

3rdparty/compiler-rt/CMakeLists.txt
```diff
@@ -0,0 +1,202 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+check_submodule_not_empty(../libgcov/compiler-rt)
```
mingweishih, line 4: The compiler-rt is currently from llvm 7. While the compiler-rt is relative stable, do you see any potential issue if we need to update this in the future? One of the pain point is the LLVM does maintain each library as separated git repo anymore. Not sure if we can to use the approach as musl/mbedtls --- have a clone of the compiler-rt (sub-directory from LLVM repo) under the openenclave organization.

anakrish:
bors try

3rdparty/compiler-rt/CMakeLists.txt
anakrish: compiler-rt is relatively stable. However, LLVM-7's compiler-rt which we use might be quite old at this point.

Since we don't anticipate making changes to compiler-rt, and to avoid the maintenance overhead of having to maintain a clone of the compiler-rt sub-directory, how about we use the published sources from the LLVM release:

https://github.com/llvm/llvm-project/releases/tag/llvmorg-10.0.1

Sources of major components (llvm, clang, compiler-rt, libunwind etc) are published in separated archives too.
compiler-rt-10.0.1.src.tar.xz is only 2MB.

I've made change to download, verify and use the compiler-rt source archive.
What do you think?

anakrish:
bors try

anakrish:
bors try

mingweishih (approved):

3rdparty/compiler-rt/CMakeLists.txt
mingweishih: Sounds good to me.

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

anakrish:
bors try

radhikaj (approved):
bors r+

anakrish:
bors try

PR #4203, opened 2021-07-29
Check existence of built-in SGX driver

CyanDevs:
Kernel 5.11 of Ubuntu 20.04 will have Intel SGX driver built-in instead of a kernel module. This change is to preserve functionality of the Ansible role in the new kernel version.

This new method is not as reliable as a check as calling modprobe, since if any of the driver files exist in their expected paths, then it will consider the driver already installed. However, as far as I know, there is no straightforward way of checking for driver functionality/version when the SGX driver is built into the kernel.

### DIFF
--- a/scripts/ansible/roles/linux/intel/tasks/sgx-driver.yml
+++ b/scripts/ansible/roles/linux/intel/tasks/sgx-driver.yml
@@ -8,19 +8,27 @@
 - name: Populate service facts
   service_facts:
 
-- name: Load default driver
-  modprobe:
-    name: intel_sgx
-    state: present
+- name: Include distribution vars
+  include_vars:
+    file: "{{ ansible_distribution | lower }}/main.yml"
+
+- name: Check default driver files
+  stat:
+    path: "{{ item }}"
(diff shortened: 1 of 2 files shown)


scripts/ansible/roles/linux/intel/tasks/sgx-driver.yml
```diff
@@ -8,20 +8,28 @@
...
+
+- name: Check default driver files
+  stat:
+    path: "{{ item }}"
+  loop: "{{ intel_dcap_driver_files }}"
   ignore_errors: yes
-  register: intel_sgx_module
+  register: intel_sgx_driver
+
+- name:
```
rs--, line 22: nit pick - mind putting a name/description for this stage

rs-- (approved):

scripts/ansible/roles/linux/intel/tasks/sgx-driver.yml
CyanDevs: Thanks for catching this!

radhikaj (approved):
bors r+

PR #4206, opened 2021-07-30
Reduce Linux image size

CyanDevs:
This reduces the uncompressed image size from 3.86GB to 2.95GB without sacrificing functionality or significantly impacting image build speed

The strategies taken here are:
* Removing temporary files left behind by apt, pip, ansible, etc. and cleaning up unnecessary packages/files
* Squashing groups of temporary files and steps into the same Docker layer

### DIFF
--- a/.jenkins/infrastructure/docker/dockerfiles/linux/Dockerfile
+++ b/.jenkins/infrastructure/docker/dockerfiles/linux/Dockerfile
@@ -19,7 +19,7 @@
 #         --build-arg ubuntu_version=18.04 \
 #         --build-arg devkits_uri=https://tcpsbuild.blob.core.windows.net/tcsp-build/OE-CI-devkits-dd4c992d.tar.gz \
 #         -t oetools-18.04:1.x \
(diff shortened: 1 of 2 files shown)


radhikaj (approved):

anakrish (approved):

nonpolarity (approved):
Nice work!
One small question, do you think we could “rm /var/lib/apt/lists/* -rf” after apt operation?

manojrupireddy (approved):

CyanDevs:
> Nice work!
> One small question, do you think we could “rm /var/lib/apt/lists/* -rf” after apt operation?

Thanks, this is a great idea since that directory isn't cleaned up by apt-get clean and can contain > 100 MB of files.

CyanDevs:
Rerunning tests at: https://oe-jenkins-dev.westeurope.cloudapp.azure.com/blue/organizations/jenkins/pipelines%2FAzure-Linux/detail/Azure-Linux/3331/pipeline

PR #4207, opened 2021-08-01
Never set errno to zero in malloc and related functions

anakrish:
According to Linux man pages,
"The value of errno is never set to zero by any system call or library function"


anakrish:
bors try

mingweishih (approved):

dthaler (approved):

PR #4220, opened 2021-08-10
Add tests to validate enclave creation using conf files; fixed oesign bug when creating 0-base enclaves

RRathna:

### DIFF
--- a/tests/sgx_zerobase/CMakeLists.txt
+++ b/tests/sgx_zerobase/CMakeLists.txt
@@ -7,4 +7,21 @@ if (BUILD_ENCLAVES)
   add_subdirectory(enc)
 endif ()
 
-add_enclave_test(tests/sgx_zerobase sgx_zerobase_host sgx_zerobase_enc)
+add_enclave_test(tests/sgx_zerobase_macro_only sgx_zerobase_host
+                 sgx_zerobase_macro_only_enc)
+
+# Tests 0-base enclave creation when OE_SET_ENCLAVE_SGX*() macro is not
+# used and only oesign configuration file is used to set properties.
+add_enclave_test(tests/sgx_zerobase_conf_only sgx_zerobase_host
+                 sgx_zerobase_conf_only_enc_signed)
+
+# Tests 0-base enclave creation when OE_SET_ENCLAVE_SGX*() macro disables
+# 0-base enclave creation and oesign configuration file enables 0-base enclave creation.
+add_enclave_test(
+  tests/sgx_zerobase_prop_macro_disable_conf_enable sgx_zerobase_host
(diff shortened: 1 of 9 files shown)


radhikaj (approved):

tests/sgx_zerobase/enc/create_zerobase.conf
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Open Enclave SDK contributors.
+# Licensed under the MIT License.
+
+# Enclave settings:
+Debug=1
```
mingweishih, line 5: Shouldn't the conf file specifies the 0-base options?

mingweishih (commented):
Can you make sure the following cases are cover
- conf enables the 0-base feature regardless whether the macro sets or not.
- conf disable the 0-base feature (or simply does not specify the 0-base option) regardless whether the macro sets or not.

tests/sgx_zerobase/enc/create_zerobase.conf
RRathna: Modified the test cases to address these scenarios.

tests/sgx_zerobase/CMakeLists.txt
```diff
@@ -7,4 +7,21 @@ if (BUILD_ENCLAVES)
...
+
+# Tests 0-base enclave creation when OE_SET_ENCLAVE_SGX*() macro disables
+# 0-base enclave creation and oesign configuration file enables 0-base enclave creation.
+add_enclave_test(
+  tests/sgx_zerobase_prop_macro_disable_conf_enable sgx_zerobase_host
+  sgx_zerobase_macro_disable_conf_enable_enc_signed)
+
+# Tests 0-base enclave creation when OE_SET_ENCLAVE_SGX*() macro enables
+# 0-base enclave creation but oesign configuration file disables 0-base enclave creation.
+add_enclave_test(tests/sgx_zerobase_macro_enable_conf_disable sgx_zerobase_host
```
mingweishih, line 26: will the test be bypassed in this case given that the zero-base is not enabled? I didn't see relevant check on the host.

mingweishih (approved):
LGTM. Just have one question on the case of `sgx_zerobase_macro_enable_conf_disable`.

tests/sgx_zerobase/CMakeLists.txt
RRathna: > will the test be bypassed in this case given that the zero-base is not enabled? I didn't see relevant check on the host.

Just added a check in host/host.cpp to return 0 when it sees a "config_disable" and not run the 0-base specific tests.

anakrish (approved):

PR #4231, opened 2021-08-17
Allow switchless OCALLs and ECALLs (nested) in exit functions

mingweishih:
This PR allows the exit functions (e.g., registered by atexit or the C `destructor` attribute) to make switchless OCALLs and ECALLs (nested). To this end, the PR separates out the exit functions from the internal `DESTRUCTOR` ECALL into a new internal ECALL (`CALL_AT_EXIT_FUNCTIONS`). This allows `oe_terminate_enclave` to invokes the exit functions and stops the switchless manager before invoking the `DESTRUCTOR` ECALL.

Note that without the change to the `DESTRUCTOR` ECALL, simply stopping the switchless manager (`oe_stop_switchless_manager`) before the ECALL leads to a memory leak in the enclave worker thread (i.e., the `oe_sgx_switchless_enclave_worker_thread_ecall` will not return).


mingweishih:
bors try

mingweishih:
bors try-

mingweishih:
bors try

enclave/core/sgx/calls.c
```diff
@@ -460,14 +461,23 @@ static void _handle_ecall(
...
-            oe_call_fini_functions();
+                /* Call all finalization functions */
+                oe_call_fini_functions();
 
+                __oe_exit_functions_done = 1;
+            }
+            break;
+        }
+        case OE_ECALL_DESTRUCTOR:
+        {
```
anakrish, line 480: We can refactor the code to `_oe_call_at_exit_functions` and call it both in `OE_ECALL_CALL_EXIT_FUNCTIONS` as well as `OE_ECALL_DESTRUCTOR`. That way the current behavior will be retained if ever existing hosts load a newer enclave.

mingweishih:
bors try

anakrish (approved):
LGTM

RRathna (approved):

PR #4237, opened 2021-08-24
New Attested TLS APIs

mingweishih:
This PR implements the new attested TLS APIs proposed by #4138, which support the following two general attestation models
- passport
- background check


mingweishih:
bors try

mingweishih:
bors try

